PR #20, opened 2026-03-25
feat: add send, chats, history CLI commands for agent messaging

yuezengwu:
## Summary

- Add `agent-hub send <target> [message]` — send messages to agents (by first-tree member name) or existing chats, with format/metadata/stdin-pipe/cross-chat-reply support
- Add `agent-hub chats` — list chats the current agent participates in, with cursor pagination
- Add `agent-hub history <chatId>` — view message history in a chat, with cursor pagination and participant access control
- Add `listChats()` / `listMessages()` SDK methods with `PaginatedResult<T>` type
- Add `parseLimit()` shared CLI utility for consistent limit option validation
- Add unit tests (8 SDK tests) and integration tests (6 server flow tests)

No server changes needed — all three commands use existing API endpoints.

## Test plan

- [x] `pnpm check` — Biome lint + format clean
- [x] `pnpm --filter @agent-hub/client typecheck` — passes
- [x] `pnpm --filter @agent-hub/client test` — 30/30 tests pass
- [x] Manual verification against local server:
  - `register` returns correct identity for both agents
  - `send baixiaohang "..."` creates direct chat and delivers message
  - `send -f markdown -m '{"intent":"review"}' baixiaohang "..."` — format + metadata preserved
  - `echo "..." | send baixiaohang` — stdin pipe works
  - `send --chat <chatId> "..."` — reply in existing chat works
  - `chats` — both participants see the shared chat
  - `history <chatId>` — full conversation in time order, all fields present
  - `history --limit 2` + `--cursor` — pagination works correctly
  - Non-participant gets `403 Not a participant of this chat`
  - `pull --ack` — receiver gets all sent messages in inbox

### DIFF
--- a/packages/client/src/__tests__/sdk-messaging.test.ts
+++ b/packages/client/src/__tests__/sdk-messaging.test.ts
@@ -1,0 +1,126 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AgentHubSDK, SdkError } from "../sdk.js";
+
+describe("AgentHubSDK messaging methods", () => {
+  let sdk: AgentHubSDK;
+  const mockFetch = vi.fn<typeof globalThis.fetch>();
+
+  beforeEach(() => {
+    sdk = new AgentHubSDK({ serverUrl: "http://localhost:8000", token: "aghub_test" });
+    vi.stubGlobal("fetch", mockFetch);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  function mockJsonResponse(data: unknown, status = 200) {
+    mockFetch.mockResolvedValueOnce(new Response(JSON.stringify(data), { status }));
+  }
+
+  function mockErrorResponse(error: string, status: number) {
+    mockFetch.mockResolvedValueOnce(new Response(JSON.stringify({ error }), { status }));
+  }
+
+  describe("listChats", () => {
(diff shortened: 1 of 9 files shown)


baixiaohang (commented):
Static review against the latest local main.

I found three CLI-facing issues worth fixing before merge:

1. `--reply-to` is silently ignored on the direct-send path.
   The flag is exposed for all `send` invocations, but it is only forwarded in the `--chat` branch. The direct `sendToAgent` contract does not accept `inReplyTo`, so `agent-hub send <agentId> ... --reply-to <msgId>` will appear to succeed without actually creating a reply link. I would either reject `--reply-to` unless `--chat` is set, or extend the direct-send contract to support it.

2. `send` only accepts string content, so `--format card` is not actually usable for real card payloads.
   The command reads content only from argv/stdin and always sends a UTF-8 string. In the existing adapter path, `card` is only treated as an interactive card when `content` is an object; otherwise it falls back to plain text. That means the CLI currently advertises a format it cannot correctly produce. I would either narrow the advertised formats to `text/markdown` for now, or add a `--content-json` / file-input path.

3. `--metadata` validates JSON syntax but not JSON shape.
   Values like `null`, `[]`, or `"foo"` will pass the local parse and then fail server-side, because the send schemas require `metadata` to be an object. Validating that locally would make the CLI error much clearer.

PR #35, opened 2026-03-26
fix: skip managed agents in sync & hide tokens for human agents

baixiaohang:
## Summary
- **Fix managed agent suspension bug**: The GraphQL-based context tree sync (`syncFromGitHub`) was missing a check for `metadata.managed=true`, causing system agents like `github-adapter` to be incorrectly suspended. Added SQL filter `(metadata->>'managed')::boolean IS NOT TRUE`.
- **Remove legacy filesystem sync**: Deleted `agent-sync.ts`, `tree-reader.ts`, and related `MemberNode` schema — all sync now goes through the GraphQL path.
- **Hide tokens for human agents**: Human agents authenticate via IM adapter user mappings, not API tokens. The Tokens card and Create Token dialog are now hidden on human agent detail pages.

## Test plan
- [ ] Trigger a context tree sync and verify managed agents (e.g. `github-adapter`) are not suspended
- [ ] Verify non-managed agents not in the tree are still correctly suspended
- [ ] Open a human agent's detail page and confirm the Tokens section is hidden
- [ ] Open a non-human agent's detail page and confirm Tokens section still appears and works


baixiaohang:
@baixiaohang 请review代码并给出review建议，然后修复代码，并commit和push。完成后使用agent-hub给baixiaohang发消息通知下

PR #52, opened 2026-03-31
feat: recursive Context Tree sync with tree_path support

baixiaohang:
## Summary
- `fetchMembers()` rewritten from fixed 2-level GraphQL query to 3-step API approach:
  1. GraphQL: get `members/` tree OID
  2. REST: recursive tree listing (scoped to `members/` subtree only)
  3. GraphQL: batch-fetch all `NODE.md` contents via aliases
- Added `tree_path` column to `agents` table — materialized path within `members/` tree (e.g. `engineering/agent-a`)
- Defensive duplicate name detection during sync with detailed error messages
- Updated `listAgents()`, sync create/update/reactivate logic to read/write `tree_path`
- DB migration `0006_agent_tree_path`

## Context
Companion to agent-team-foundation/first-tree#12 (recursive member validation). Together they enable hierarchical agent structures under `members/` at any depth, with agent ID remaining the directory name and `tree_path` recording the full materialized path.

## Test plan
- [x] `pnpm check` (biome lint + format) passes
- [x] `pnpm typecheck` passes
- [x] All server tests pass
- [ ] Apply migration on dev DB and verify `tree_path` column exists
- [ ] Trigger manual sync and verify nested members are discovered with correct `tree_path`

### DIFF
--- a/packages/server/drizzle/0006_agent_tree_path.sql
+++ b/packages/server/drizzle/0006_agent_tree_path.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "agents" ADD COLUMN "tree_path" text;
--- a/packages/server/drizzle/meta/_journal.json
+++ b/packages/server/drizzle/meta/_journal.json
@@ -43,6 +43,13 @@
       "when": 1774972900000,
       "tag": "0005_delegate_mention",
       "breakpoints": true
+    },
+    {
+      "idx": 6,
+      "version": "7",
+      "when": 1775073600000,
+      "tag": "0006_agent_tree_path",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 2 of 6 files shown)


baixiaohang (commented):
Blocking issue from local review: packages/server/src/services/context-tree-graphql.ts keeps going when the GitHub REST tree API returns truncated: true. At that point the tree listing is partial, but syncFromGitHub still derives memberNames from it and phase 2 will suspend every active agent that was omitted from the truncated response. On a large enough members/ subtree this can turn one partial GitHub response into unintended mass suspension. I think the sync should fail hard on truncation, or implement a safe fallback, before any DB writes/suspensions happen.

PR #60, opened 2026-04-02
feat: add server-embedded Kael runtime for Hub-Kael integration

yuezengwu:
## Summary

- Add KaelRuntime service that forwards Hub inbox entries to Kael's dedicated API (`POST /api/v1/hub/messages`)
- Kael-powered agents participate as team members in Hub's messaging — Kael responds by calling Hub's Agent API directly using injected agent token
- Web console binding form for admin to configure Kael connection per agent (User ID, Project ID, Agent Token)
- Server config: `KAEL_ENDPOINT`, `KAEL_API_KEY`, `FIRST_TREE_HUB_PUBLIC_URL`

## Architecture

```
Hub Server (KaelRuntime)              Kael Backend
  inbox entry → POST /hub/messages →   process message
                                        ↓
  ← POST /agent/chats/.../messages ←   respond via Hub API
```

No SSE, no webhook callback, no Client Runtime involvement. Bidirectional HTTP API calls.

## Changes

| File | Description |
|------|-------------|
| `server/services/kael-runtime.ts` | **New** — KaelRuntime: load configs, claim inbox, POST to Kael, ACK/NACK with retry |
| `server/__tests__/kael-runtime.test.ts` | **New** — 19 unit tests |
| `server/app.ts` | Wire KaelRuntime lifecycle + PG NOTIFY hot-reload |
| `server/services/background-tasks.ts` | Add kael outbound timer (5s interval) |
| `shared/config/server-config.ts` | Add `kael` config group (endpoint, apiKey, hubPublicUrl) |
| `shared/schemas/adapter.ts` | Add "kael" to platform enum |
| `web/pages/agent-detail.tsx` | Kael-specific binding form fields |

## Design docs

- Hub side: `first-tree-context/agent-hub/kael-integration.md`
- Kael side: `first-tree-context/kael/platform/integrations/hub-integration.md`

## Test plan

- [x] Unit tests: 19 tests covering reload, processOutbound, shutdown, retry, auth header
- [x] E2E verified locally: Hub → Kael → Hub round-trip with real Kael backend
- [x] lint + typecheck clean


baixiaohang:
## Architecture concern: Kael integration should use Client pattern, not Server-embedded adapter

### Core issue

`kael-runtime.ts` is placed in `packages/server/src/services/`, directly operating on the DB to claim inbox entries and push to Kael API. This treats Kael as an IM channel adapter (like Feishu), but Kael is an AI agent (like Claude Code) and should follow the Client pattern instead.

### Why Feishu adapter belongs in server, but Kael does not

- **Feishu** is a communication channel — it bridges external IM users into Hub's messaging system, with bidirectional protocol translation. It's an I/O layer extension of the server.
- **Kael** is an agent — it should connect via the Client SDK, pull messages from inbox, process them via a handler, and reply through the Hub API.

The project's own architecture rule states:

> **Inbox is the Server/Client boundary:** Server writes to Inbox, Client pulls / receives WebSocket notifications.

### Specific concerns

1. **Violates the Inbox boundary principle** — The server now knows Kael's API format (`/api/v1/hub/messages`), auth scheme (`X-Internal-API-Key`), and payload structure. This couples the server to a specific agent implementation.

2. **Doesn't scale** — Every new AI agent type would require server-side code changes. The Client + handler pattern lets new agents integrate without touching the server.

3. **Security: agent token leak** — The payload sends `hub_agent_token` in plaintext to Kael. In the Client pattern, the client holds the token itself; Kael never needs to see it.

4. **Infrastructure already exists** — The client package has `AgentRuntime` + `AgentSlot` + handler registration. This PR itself adds `connect` and `start` CLI commands. A `kael-handler` implementation would fit naturally into this existing pattern.

### Suggestion

Move Kael integration from server to client side — implement it as a handler (e.g., `kael-handler`). The flow becomes:

```
Hub Server → (inbox / WebSocket notify) → Client (kael-handler) → HTTP POST → Kael API
                                                                 ← Kael calls Hub API to reply
```

If Kael needs to respond asynchronously, it can call the Hub API directly with its own agent token — identical to how Claude Code agents work.

### Additional technical notes (if current approach is kept)

These are secondary to the architecture concern above, but worth noting:

- **nack race condition** (`nackEntry`): SELECT then UPDATE on `retry_count` has a TOCTOU race. Could be a single `UPDATE ... SET status = CASE WHEN retry_count >= 3 THEN 'failed' ELSE 'pending' END`.
- **N+1 queries in `reload()`**: Each adapter config triggers a separate agent lookup. Could be batched into one query or joined.
- **zod version downgrade**: `zod` is rolled back from `^4.0.0` to `^3.25.x` across client and command packages. This reverts PR #58's upgrade — needs explicit justification.
- **Package version rollback**: `command/package.json` version goes from `0.3.4` → `0.3.1`, which would conflict with any published version.
- **`findConfigByInboxId` is O(n)**: Linear scan on every claimed entry. An `inboxId → config` reverse Map would help.
- **Raw SQL vs Drizzle query builder**: `reload()` uses `db.execute(sql\`...\`)` for the agent lookup, where Drizzle's query builder would be more idiomatic.

yuezengwu:
Thanks for the detailed review. We discussed this offline — here's the conclusion:

## Architecture: Server-embedded is intentional

Kael doesn't need Client Runtime. The key difference from Claude Code:

| | Claude Code | Kael |
|---|---|---|
| Process management | Spawn/suspend/resume local CLI process | No process — just an HTTP POST |
| Session state | LLM context in memory, crash recovery | Kael manages sessions independently |
| Concurrency control | max_sessions, idle eviction, LRU | Not needed — fire-and-forget |

The Client Runtime's value is session lifecycle management (AgentSlot, SessionManager, handler lifecycle). Kael doesn't use any of that — it's a stateless HTTP forwarder. Wrapping it in the Client pattern would add unused complexity (session slots, idle timeouts, workspace bootstrap) for a service that just does `fetch(kaelEndpoint, payload)`.

The Inbox boundary principle still holds: **Server writes to Inbox**. KaelRuntime reads from Inbox (same as the Feishu adapter's `processOutbound`). The boundary is about stateless-Server/stateful-Client separation — Kael's handler IS stateless, so it belongs with the Server.

If a future agent type DOES need session lifecycle (process management, LLM context), it goes through the Client handler. Kael doesn't.

## Technical issues — will fix

- **nackEntry TOCTOU**: Good catch. Will merge into a single atomic UPDATE with CASE.
- **N+1 in reload()**: Will batch agent lookups into one query.
- **findConfigByInboxId O(n)**: Will add a reverse Map.
- **zod/package version changes**: These are unrelated diff noise from the base branch. Will clean up.
- **Raw SQL in reload()**: Will switch to Drizzle query builder where possible.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #75, opened 2026-04-09
feat: multiplexed client connections and activity monitoring (runtime M1)

yuezengwu:
## Summary

- **Multiplexed client model**: Shift from 1:1 agent→server WebSocket to one persistent connection per SDK process (`ClientConnection`), supporting multiple agents via `agent:bind`/`agent:unbind` protocol
- **Client tracking**: New `clients` table, `ClientService`, and admin API (`/admin/clients`) for connection lifecycle management with heartbeat and stale cleanup
- **Activity monitoring**: New `ActivityService` tracks per-agent runtime state (idle/working/error), session counts, errors, and task references — exposed via `/admin/agents/activity` and a new dashboard Activity page with agent/client dual-view tabs
- **Presence enrichment**: `agent_presence` extended with `clientId`, runtime fields (`runtimeType`, `runtimeVersion`, `runtimeState`, `runtimeDescription`), and activity counters
- **Dashboard**: Activity page with 5s agent / 10s client auto-refresh, state badges, reset-error and force-disconnect actions

## Test plan

- [ ] `pnpm test` — new tests for `ClientService`, `ConnectionManager`, and `ClientConnection`
- [ ] `pnpm typecheck && pnpm check` — full type and lint pass
- [ ] Manual: start server, connect a client SDK, verify `client:register` → `agent:bind` flow
- [ ] Manual: verify Activity page shows connected agents and clients with live refresh
- [ ] Manual: test force-disconnect and reset-activity admin actions

### DIFF
--- a/packages/client/src/__tests__/client-connection.test.ts
+++ b/packages/client/src/__tests__/client-connection.test.ts
@@ -1,0 +1,313 @@
+import { EventEmitter } from "node:events";
+import { describe, expect, it, vi } from "vitest";
+
+/**
+ * Unit tests for ClientConnection — mock WebSocket, no real server.
+ *
+ * Focus:
+ * - ref-based correlation for bind responses (P1-1)
+ * - error responses matched to correct pending bind via ref (P1-3)
+ * - pending binds rejected on WS close (P2-3)
+ * - pending binds rejected on disconnect() (P2-4)
+ * - reconnect triggers rebind for all boundAgents (P2-1)
+ * - connect() fast-fails on early socket closure (P2-2)
+ */
+
+// -- Mock WebSocket --
+
+class MockWebSocket extends EventEmitter {
+  static CONNECTING = 0;
+  static OPEN = 1;
+  static CLOSING = 2;
+  static CLOSED = 3;
+  readyState: number = MockWebSocket.CONNECTING;
+  sent: string[] = [];
+
+  send(data: string) {
+    this.sent.push(data);
+  }
+
+  close(_code?: number, _reason?: string) {
+    this.readyState = MockWebSocket.CLOSED;
+    this.emit("close");
+  }
+
+  terminate() {
+    this.readyState = MockWebSocket.CLOSED;
+    this.emit("close");
+  }
+
+  removeAllListeners() {
+    super.removeAllListeners();
+    return this;
+  }
+
+  receiveMessage(msg: Record<string, unknown>) {
+    this.emit("message", Buffer.from(JSON.stringify(msg)));
+  }
+
+  simulateOpen() {
+    this.readyState = MockWebSocket.OPEN;
+    this.emit("open");
+  }
+}
+
+// Track the latest mock WS instance
+let latestMockWs: MockWebSocket;
+
+vi.mock("ws", () => {
+  // Must use Reflect.construct to return a custom object from `new WebSocket(url)`
+  // while satisfying biome's no-constructor-return rule.
+  function createMockWs() {
+    const instance = new MockWebSocket();
+    latestMockWs = instance;
+    setTimeout(() => {
+      if (instance.readyState === MockWebSocket.CONNECTING) {
+        instance.simulateOpen();
+      }
+    }, 0);
+    return instance;
+  }
+  const MockWS = new Proxy(createMockWs, {
+    construct: () => createMockWs() as object,
+  });
+  Object.assign(MockWS, { CONNECTING: 0, OPEN: 1, CLOSING: 2, CLOSED: 3 });
+  return { default: MockWS, __esModule: true };
+});
+
+// Must import AFTER vi.mock
+const { ClientConnection } = await import("../client-connection.js");
+
+/** Parse a sent message by index. Throws if missing. */
+function parseSent(ws: MockWebSocket, idx: number): Record<string, string> {
+  const raw = ws.sent[idx];
+  if (!raw) throw new Error(`No message at index ${idx}`);
+  return JSON.parse(raw) as Record<string, string>;
+}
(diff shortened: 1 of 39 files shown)


baixiaohang (commented):
## Code Review: Runtime M1 — Multiplexed Client Connections

Overall quality is good. The architecture aligns well with the `agent-runtime-m1.20260409.md` design doc — multiplexed WS model, `runtime_state` as the single authority for "is agent running", and the Client → Agent → Session hierarchy are all correctly implemented.

Below are findings organized by severity.

---

### 🔴 Critical — recommend fixing before merge

**1. `client:register` allows hostile clientId takeover**
`packages/server/src/api/agent/ws-client.ts`

The `/agent/ws/client` WS endpoint has no HTTP-layer auth (by design — auth happens at `agent:bind`). However, `client:register` itself requires no credentials. An attacker can send `client:register` with an existing clientId, and `setClientConnection` will silently overwrite the legitimate client's WS reference in the `clientConnections` Map.

**Suggestion:** `setClientConnection` should reject or close an existing active connection before accepting a new one for the same clientId. Also consider adding rate-limit to the WS endpoint.

---

### 🟠 Major — should fix before or shortly after merge

**2. `agent:message` event passes `inboxId` instead of `agentId` — semantic mismatch**
`packages/client/src/client-connection.ts` — `handleMessage` / `new_message` branch

```typescript
} else if (type === "new_message") {
  const inboxId = msg.inboxId as string | undefined;
  if (inboxId) {
    this.emit("agent:message", inboxId, msg);
  }
}
```

The event signature is `[agentId: string, data: unknown]`, but it emits `inboxId`. In AgentSlot, every slot listens to the same `ClientConnection` and triggers `pullAndDispatch()` on every message — O(N) amplification in multi-agent mode. Functionally correct for now (each agent pulls its own inbox), but the wrong identifier will become a real bug when fine-grained routing is added.

**3. AgentSlot M1 path calls `sdk.register()` after `bindAgent` — redundant registration**
`packages/client/src/runtime/agent-slot.ts`

`bindAgent` already authenticates and sets presence (including `clientId`, `runtimeState`) on the server. The subsequent `sdk.register()` (`POST /agent/register`) is an extra HTTP call that may overwrite the `clientId` / `instanceId` just set by bind.

**Suggestion:** Replace with `sdk.getMe()` (read-only) or skip entirely since bind already returns agent info.

**4. Legacy `forceDisconnect(agentId)` closes the entire Client WS in M1 mode**
`packages/server/src/services/connection-manager.ts`

`bindAgentToClient` writes the client's WS into the legacy `activeConnections` Map. When admin calls `forceDisconnect(agentId)` (still used by existing admin agents API), it calls `ws.close()` on the shared client WS, disconnecting ALL agents on that client — not just the target.

**Suggestion:** In M1 mode, `forceDisconnect` should send an `agent:force_disconnect` message and call `unbindAgentFromClient` instead of closing the socket.

**5. `cleanupStaleClients` doesn't reset `agent_presence.client_id`**
`packages/server/src/services/client.ts`

The raw SQL in `cleanupStaleClients` resets runtime fields (`runtime_state`, `active_sessions`, etc.) but omits `client_id = NULL`. After stale cleanup, `agent_presence.client_id` still points to the disconnected client.

**6. `db.execute` return type assumption**
`packages/server/src/services/client.ts` — `cleanupStaleClients`

`result.length` and `result.map(r => r.id)` assume `db.execute<T>()` returns `T[]`. Depending on the Drizzle ORM version, it may return `{ rows: T[], rowCount: number }` instead, which would cause a runtime error.

---

### 🟡 Minor — can be addressed in follow-up PRs

| # | Location | Issue |
|---|---|---|
| 7 | `ActivityService.resetActivity` | Can set `runtimeState = "idle"` on a disconnected agent (clientId is null). Should check that agent is still bound before resetting. |
| 8 | `SessionManager.notifyStateChange` | Never reports `"error"` state — only toggles between `idle` and `working`. Activity dashboard Error count will always be 0 unless reported via REST. |
| 9 | `AgentRuntime` | Single-agent mode defaults to legacy `AgentConnection` (`useClientConnection ?? entries.length > 1`), meaning no activity reporting for single-agent users. Consider defaulting to `true`. |
| 10 | Migration `0009` | No index on `agent_presence.client_id` — `disconnectClient` and `cleanupStaleClients` both query `WHERE client_id = ?`. |
| 11 | `clients.id` | Accepts arbitrary client-provided strings as PK (`z.string().min(1).max(100)`). Consider validating UUID format or generating server-side. |
| 12 | `AgentSlot` | `agent:message` listener added in `start()` is never removed in `stop()` — potential memory leak if slot is stopped while ClientConnection lives on. |
| 13 | `agent-detail.tsx` | Uses `as Record<string, unknown>` to access M1 fields — violates project's "No `as` assertions" convention. Update the TypeScript response type instead. |
| 14 | `ws-client.ts` | Variable shadowing: inner `const raw = data.token` shadows outer `socket.on("message", async (raw) => ...)`. Rename to `rawToken`. |
| 15 | Tests | Missing coverage for: `unbindAgent`, `reportActivity`, heartbeat timing, concurrent `connect()` calls, and `bindAgent` after `disconnect()`. |
| 16 | `listClients` | Only returns `status = "connected"` clients — no way to view disconnected/historical clients. Consider adding a `status` filter param. |

---

### Summary

The multiplexed client model, ref-based bind correlation, and the clean DB schema / service separation are well done. Recommend addressing **Critical #1** (clientId takeover) and **Major #4** (forceDisconnect scope) before merge, as they can cause real issues in multi-agent deployments. The rest can be tracked as follow-ups.

baixiaohang (approved):

PR #88, opened 2026-04-15
fix: set password state before invalidating queries in member creation

baixiaohang:
## Summary

- Swap order of `setCreatedPassword` and `invalidateQueries` in member creation `onSuccess` callback
- Ensures password dialog renders before the member list refetch triggers a re-render

One-line fix in `packages/web/src/pages/members.tsx`.


baixiaohang:
False positive — the issue was browser cache, not code. Closing.

PR #95, opened 2026-04-17
feat: agent configuration M1 + unified user token

yuezengwu:
## Summary

Delivers the **Agent Configuration M1** milestone plus the **Unified User Token** migration, and repairs the full server test suite that the refactor broke.

### Server
- **Agent runtime config (PRD §5–7)**: new `agent_configs` table, dry-run validation, versioned PATCH with hot-switch, retire guard requiring no pinned agents, immutable `clientId` per agent, shared `runtimeFieldsReset` helper.
- **Unified user token**: removes per-agent tokens / `agent-tokens` schema / `agent-auth` + `github-auth` middleware. WS handshake auth via JWT, `X-Agent-Id` header on agent-scoped routes, server pushes `auth:expired` before closing 4401.
- **Security hardening** (from colleague review):
  - `admin /agents/:uuid/config` scoped via `assertAgentVisible` / `assertCanManage`.
  - `admin /agents/:uuid/client-status` sources `clientId` from `agents` (authoritative pin), presence only for liveness, with visibility check.
  - `admin DELETE /clients/:id` retires **before** disconnecting so a blocked retire never drops the running client.
  - `listAgentsForMember` uses `agents.clientId` (matches `listAgents` and the documented intent).
  - `cleanupStaleClients` reuses `runtimeFieldsReset`.

### Client
- `ClientConnection` — single WS per client, multiple agents multiplexed, proactive JWT refresh.
- Claude Code handler: `agentConfigCache` hot-switch now tracks an `appliedPayload` snapshot so same-family model swap actually triggers (the prior code compared cache-vs-cache and never fired — silent bug fix).
- `SessionManager` / `AgentSlot`: replace inline runtime-state string unions with shared `RuntimeState`; typed listener removal; poll errors logged instead of swallowed.

### Web
- `agent-detail` page restructured into sub-sections (`env` / `git` / `mcp` / `model` / `prompt` / `identity` / `save-bar` / `status-bar` / `danger-zone`).
- `clients` page memoizes `agentsByClient`; drops dead hidden span + narrative comment in `agent-detail`.

### Command CLI
- Parallelizes agent starts via `Promise.allSettled`.
- Onboard / `agent-config` CLI commands following shared `core/` helpers.

### Test fixtures (full suite repair)
- New helpers: `seedClient`, `createAdminContext`, `createTestAgent.request`.
- Migrate **25 server test files** to seed admin + client and pass `managerId` / `clientId` per Rule R-RUN, or use `X-Agent-Id` header on agent-scoped routes.
- `ws-jwt-refresh` re-seeds admin per test (`setup.ts` truncates between tests).
- Drop obsolete agent-tokens tests; remove dead `session-state` / `ws-single-connection` / `kael-runtime` / `e2e-*` test files.

## Test plan

- [x] `pnpm check` — Biome clean (326 files)
- [x] `pnpm typecheck` — 9/9 packages
- [x] `pnpm build` — 5/5 packages
- [x] `pnpm test` — **436/436 tests** (Shared 29 / Client 112 / Command 6 / Server 289)
- [ ] Reviewer sanity-check: dry-run PATCH config on a running agent, confirm hot-switch applies
- [ ] Reviewer sanity-check: blocked retire (agents still pinned) returns 409 without dropping the WS connection
- [ ] Reviewer sanity-check: offline-but-pinned agent detail page shows "assigned but offline" (not "unclaimed")


baixiaohang (commented):
# Review — feat: agent configuration M1 + unified user token

Large PR (145 files, +8346 / -6652) touching schema migrations, the auth model, hot config, WS multiplexing, and a full test-suite refactor. Overall direction matches the PRD and test coverage is solid (436/436), but there are a handful of items I'd like to see addressed before merge.

## Blockers (please fix before merge)

1. **`packages/client/src/handlers/claude-code.ts` — Path B restart leaves the new query without a consumer.**
   `maybeSwitchConfig`'s restart path calls `buildQuery(sid, ctx, sid)`, which only rebuilds the query + `inputController`. It does **not** call `consumeOutput`. When the old query is `close()`-d the existing consumer loop exits via `return` (around L273-275), so the new query has no reader and any `inputController.push(...)` messages silently pile up. The switch "succeeds" but the session is dead — extremely hard to diagnose in production.
   Fix: either follow the `respawnQuery` pattern (keep the existing loop alive) or start a fresh `consumeOutput` after `buildQuery`. Also delay assigning `appliedPayload` until the consumer has actually processed a message from the new query, so a failed switch doesn't get recorded as applied.

2. **`packages/server/src/services/config-service.ts` — debounce coalescing bypasses optimistic locking.**
   When two admins PATCH concurrently, the second one's patch lands in the pending queue and at flush time uses the post-`commitWrite` version as its expectedVersion. That swallows the conflict against the version the caller actually submitted: two admins editing `model` at the same time silently stomp each other, both get 200, and the PRD §5–7 versioned-PATCH semantics are violated.
   Fix: queued awaiters must compare against the expectedVersion they themselves submitted, and should receive their own 409 on mismatch rather than having identical-key patches merged.

3. **`packages/server/src/api/admin/clients.ts` — `POST /:clientId/disconnect` missing member-scope check.**
   The handler only enforces `adminOnly`, so a cross-org admin can disconnect another org's client. This contradicts the PR description ("assertAgentVisible / assertCanManage scoped"). Add an `assertClientVisible`-equivalent check.

## Major

4. **`packages/server/drizzle/0020_unified_user_token.sql:59-69` — backfill 3b is dead code (`WHERE false`).**
   Non-human agents with no `agent_presence` row keep `client_id = NULL` and the migration passes silently. At runtime `agent-selector` returns 403 and WS bind returns `WRONG_CLIENT`, which looks like "installed but broken" to the operator. Mirror the step-4c `orphan_count` RAISE EXCEPTION pattern for `client_id IS NULL AND type != 'human' AND status != 'deleted'`, or actually implement 3b.

5. **`packages/web/src/pages/agent-detail/use-config-draft.ts:228-248` — React Query refetch clobbers the draft.**
   `useEffect([cfg])` unconditionally reseeds the draft on every background refetch (window focus, route return, `invalidateQueries`). Users editing the form silently lose their changes. Seed only when `draft == null`, or disable `refetchOnWindowFocus` while any section is dirty.

6. **`packages/web/src/pages/agent-detail.tsx:105-113` — 409 detection relies on error-message regex.**
   `/409|version mismatch/i.test(msg)` is fragile; `ApiError.status` is already used elsewhere (e.g. `clients.tsx:117`). Switch to `err instanceof ApiError && err.status === 409`.

7. **`packages/client/src/client-connection.ts` — `agent:force_disconnect` doesn't clean `desiredBindings`.**
   The next reconnect re-binds an agent the server just ejected, producing a loop of `bind:rejected(not_owned/wrong_client)`. On force_disconnect (and on `not_owned`/`agent_suspended` rejections) delete the entry from `desiredBindings`.

8. **`packages/server/src/services/client.ts` — `retireClient` is not transactional (TOCTOU).**
   Pinned check → tombstone clear → DELETE happens in separate statements. A concurrent `createAgent(clientId=X)` flips the DELETE into a raw PG 23503 rather than a friendly ConflictError. Wrap in `db.transaction` and `SELECT … FOR UPDATE` on the client row.

9. **`packages/command/src/commands/agent-config.ts` — dry-run lacks Zod validation, and all set commands self-bypass the optimistic lock.**
   `JSON.parse(...)` is cast straight to `Partial<AgentRuntimeConfigPayload>`; use `agentRuntimeConfigPayloadSchema.partial()` for pre-flight validation. More importantly, every set command calls `getCurrent()` to obtain the expectedVersion it sends, which means the CLI never raises conflicts. Add `--expected-version` (required, or warn loudly when absent) so operators can detect concurrent edits.

## Minor

- **`packages/client/src/handlers/claude-code.ts:197-200`** — hot-switch comparison uses `JSON.stringify({...payload, model: ""})`. Object-key order sensitivity means a reordering of `mcpServers` / `env` / `gitRepos` falsely falls out of "only model changed" and skips Path A. Prefer a structural deep-equal (lodash `isEqual` or hand-written).
- **`packages/server/src/api/agent/ws-client.ts:97-108`** — `auth:expired` fires at expiry instead of with a pre-expiry buffer (e.g. 30 s). Clients receive the notice already expired.
- **`packages/client/src/runtime/agent-slot.ts`** — poll errors are now logged instead of swallowed; when Hub is unreachable this floods at 12 lines/min. Add dedup or exponential backoff.
- **`packages/shared/src/schemas/agent-runtime-config.ts`** — `localPath` on `gitRepoSchema` lacks escape protection. Combined with `<cwd>/<localPath>` in git-mirror-manager, absolute paths and `..` allow arbitrary-path writes. `.refine(p => !p.startsWith("/") && !p.includes(".."))`.
- **`packages/server/src/services/inbox.ts::pollInbox`** — builds payloads (which hit `agent_configs`) while still holding the SKIP-LOCKED row lock. Under high load this throttles throughput. Claim first, commit, then build payload.
- **`packages/web/src/pages/agent-detail/status-bar.tsx`** — the "unclaimed" predicate (`offlineSince && !clientId`) diverges from `agent-detail.tsx:260`'s `!clientStatus?.clientId`. Unify on `clientId == null`.
- **`packages/shared/src/schemas/ws-auth.ts`** — schema only covers client→server frames. Server→client frames (`auth:ok` / `auth:rejected` / `auth:expired`) are hand-string-compared on the client. Add `wsServerAuthFrameSchema` and parse.
- **CLI onboarding docs** — after removing the bootstrap-token endpoint, the first-time flow on a new machine now requires "Web → create connect token → `connect --token`". Please update [docs/claim-agent-guide.md](docs/claim-agent-guide.md) and `skills/first-tree-hub-cli/SKILL.md` to match.

## Migrations — timestamp sanity

Timestamps are monotonically increasing (0018 → 0021, +1 day each) — **not earlier than prior entries**. They are synthetic (round UTC midnights) continuing the repo's existing style, not `drizzle-kit generate` real-time output. Fine for this PR; just note that if a future migration on `main` uses a real `Date.now()` (~April 17), drizzle sorts by `when`, so it would slot **before** 0019. Worth deciding on a team convention in a follow-up.

No number conflicts with other open PRs (#86, #51 don't touch migrations).

## Worth highlighting

- The unified JWT model is clean; `X-Agent-Id` + `assertAgentVisible`/`assertCanManage` covers most endpoints well.
- `DELETE /clients/:id` retiring *before* disconnecting is a correct fix.
- Test refactor is thorough (25 server test files migrated, shared helpers).
- `ClientConnection` single-WS multi-agent + proactive JWT refresh is architecturally clean.
- Migration 0020's `orphan_count` pattern is a good template — reuse it in step 3b (see Major #4).

## Test-plan suggestions (missing from PR description)

- [ ] Path B hot-switch: swap sonnet ↔ opus (cross-family) and confirm new messages are consumed and replied.
- [ ] Concurrent PATCH: two admin sessions editing the same config at once — verify 409.
- [ ] Offline-but-pinned agent: confirm the status bar shows "assigned but offline", not "unclaimed".
- [ ] Migration rollback dry-run on staging with real `agent_tokens` data (DDL is transactional, but document the rollback runbook explicitly).

---

Direction is right and engineering rigor is high. Please tighten the five items in Blockers + Major #4–6 (Path B consumer, config-service optimistic lock, clients/:id/disconnect scope, 3b backfill, draft clobber) before merge; the rest can land in a follow-up PR.

yuezengwu:
Thanks for the thorough review. Addressed all 3 blockers + 4 major items in this force-push (`554c23f` → `e655992`). Leaving the minor items for a follow-up PR as you suggested.

## Blockers — all fixed

1. **Path B restart consumer** (`packages/client/src/handlers/claude-code.ts`) — Split `buildQuery` into a pure builder + a separate `recordAppliedPayload`. `spawnQuery` and Path B restart now both explicitly start a fresh `consumeOutput` loop so the new query has a reader; `respawnQuery` keeps the existing loop alive (unchanged). `appliedPayload` is recorded only *after* `buildQuery` succeeds — a failed build never gets recorded as applied. A failed `consumeOutput` in the new query will surface via the existing retry path.

2. **config-service optimistic lock** (`packages/server/src/services/config-service.ts`) — `PendingWrite` now stores each awaiter's own `patch` + `expectedVersion`. `flushAgent` re-reads the current DB version and rejects every awaiter whose `expectedVersion` doesn't match it with its own `ConflictError`. Only the awaiters whose `expectedVersion` equals the current DB version participate in the aggregated commit, so a rejected caller's patch never leaks into the committed payload. Test updated to assert the new semantics (5 concurrent `v=1` writers → 1 commit + 4 × 409, not "1 immediate + 1 aggregated").

3. **`/admin/clients/:id/disconnect` owner scope** (`packages/server/src/api/admin/clients.ts` + `services/client.ts`) — Added `assertClientOwner(db, clientId, userId)` helper that 404s if the client doesn't exist OR isn't owned by the caller's user. Wired into `GET /:clientId`, `POST /:clientId/disconnect`, `DELETE /:clientId`. `GET /` now filters by `scope.userId` so `listClients` only returns the caller's own clients. `MemberScope` gained `userId` (was already on `request.member`).

## Major — 4, 5, 6, 8 fixed

4. **Migration 0020 step 3b** (`packages/server/drizzle/0020_unified_user_token.sql`) — Replaced the `WHERE false` dead code with a step-4c-style `RAISE EXCEPTION` when non-deleted non-human agents still have `client_id IS NULL` after 3a's backfill. Operators now get a loud, actionable error pointing at `first-tree-hub connect` or soft-deletion instead of silent runtime 403s.

5. **`use-config-draft` refetch clobber** (`packages/web/src/pages/agent-detail/use-config-draft.ts`) — `useEffect` now only re-seeds when `draft === null` — background refetches (focus/route-return) no longer stomp in-progress edits. `resetAll()` sets the draft to `null` so the next render pulls the latest cfg (whether it came from `setQueryData` on save or `invalidateQueries` on explicit reload).

6. **409 via `ApiError.status`** (`packages/web/src/pages/agent-detail.tsx`) — Replaced the `/409|version mismatch/i.test(msg)` regex with `err instanceof ApiError && err.status === 409`, matching the `clients.tsx` pattern.

8. **`retireClient` transactional** (`packages/server/src/services/client.ts`) — The pinned-check + soft-deleted-FK-clear + `DELETE clients` run inside `db.transaction` with `SELECT ... FOR UPDATE` on the client row. A racing `createAgent(clientId=X)` can no longer land between the pinned check and the DELETE, so we never surface PG 23503 instead of `ConflictError`. (This one wasn't strictly on your "before merge" list but was the smaller sibling of Blocker 3 in `clients.ts`, cheap to fix together.)

## Left for follow-up (per your suggestion)

- **Major 7** — `agent:force_disconnect` / `not_owned` / `agent_suspended` cleanup of `desiredBindings` (client-connection reconnect loop).
- **Major 9** — `agent-config` CLI Zod validation + `--expected-version` flag.
- All minor items (hot-switch deep-equal, `auth:expired` pre-expiry buffer, poll error backoff, `localPath` escape protection, `pollInbox` lock-then-build, `status-bar` unclaimed predicate, ws-auth server→client frame schema, CLI onboarding docs).
- Test-plan suggestions (Path B cross-family smoke, concurrent PATCH, offline-but-pinned UX, migration rollback runbook) — will add manual verification notes to the follow-up PR.

All checks still green: `pnpm check` / `pnpm typecheck` / `pnpm test` — 436/436 tests pass.

PR #107, opened 2026-04-20
feat(onboarding): simplify agent creation and last-step connect flow

Gandy2025:
## Summary
- replace the old multi-field agent creation dialog with a streamlined onboarding flow centered on name, prompt template, and runtime choice
- add a dedicated last-step modal that fetches a fresh connect command, appends the new agent identity, and polls for the first computer bind
- make the onboarding copy and related UI use the user-facing `Computer` terminology while preserving the existing runtime and API model
- prepend CLI installation to the connect-token command so the copied command is self-bootstrapping on a fresh machine

## Why
The onboarding flow previously broke between "agent row created" and "this computer actually claims that new agent". Users could connect their computer, but the newly created agent was not written into the local config, so no websocket opened for it and the web UI stayed stuck waiting for the bind to happen.

This PR closes that loop by making the last-step command carry the new agent identity through `client connect`, so the local agent config exists before the runtime starts and the first bind can complete.

## Validation
- `pnpm check`
- `pnpm typecheck`
- `pnpm --filter @agent-team-foundation/first-tree-hub test`


Gandy2025:
Closing in favor of #110, which includes this work (M5 simplified creation + LastStep one-liner assembly + Computer terminology) plus B1 auto-pin (#108-dependent), prompt-section removal, and extended Computer rename across the admin / agent-detail pages.

PR #144, opened 2026-04-22
fix(command): resolve own version via walk-up to survive tsdown bundle

yuezengwu:
## Summary

- **v0.9.1 crashed on startup.** `packages/command/src/core/version.ts` used `require("../../package.json")`, but tsdown flattens every chunk into `dist/`, so the bundled path resolved to `packages/package.json` (or `node_modules/@agent-team-foundation/package.json` under a global install) — neither exists. The CLI threw `Cannot find module '../../package.json'` before it could run any command.
- **Fix**: walk up from `dirname(fileURLToPath(import.meta.url))` until we find a `package.json` whose `name` matches this package. Works identically in dev (`tsx src/cli/index.ts` → `src/core/` finds `packages/command/package.json` two levels up) and in the published tarball (`dist/cli/index.mjs` → one level up in a global install, two from the CLI entry).
- **Hardened fallback path** during code review — see Risks below.
- Bumps `packages/command` to `0.9.2` per the consumer-tarball versioning policy.

## Why the bundled path diverged

Source layout: `packages/command/src/core/version.ts` — `../../` = `packages/command/` ✅

Bundled layout: `packages/command/dist/core-*.mjs` — `../../` = `packages/` ❌ (nothing there). In a global install: `node_modules/@agent-team-foundation/first-tree-hub/dist/core-*.mjs` → `../../` = `node_modules/@agent-team-foundation/` ❌. tsdown doesn't rewrite string literals passed to `require()`, so the depth assumption baked into the source doesn't survive the bundler.

## Risks surfaced during review (fixed in-PR)

Two issues turned up when reviewing the initial fix. Both are now addressed:

1. **Silent `"0.0.0"` fallback would trigger a self-update loop.** `"0.0.0"` is valid SemVer and less than any real version. If the walker ever bottoms out (EACCES-locked manifest, exotic install layout, >10 parent dirs), `ClientRuntime.currentVersion` becomes `"0.0.0"`, `UpdateManager.decide` treats it as "way behind," and in `policy=auto` on a non-TTY service unit the client would `npm install -g @latest` on every reconnect. The v0.9.1 startup-crash would have quietly reincarnated as an install storm. **Fix**: fallback is now the string `"unknown"` (not valid SemVer). `packages/client/src/runtime/update-manager.ts:151` already has a `semver.valid(current) === false` → warn-and-skip branch, and the welcome-frame schema at `packages/shared/src/schemas/ws-auth.ts:27` is `z.string().min(1)` so it accepts the sentinel.
2. **`catch {}` hid non-ENOENT errors.** EACCES on a locked manifest, EISDIR on a weird layout, `SyntaxError` on a truncated install — all were swallowed identically to "no file here." **Fix**: inspect `err.code`, silent on `ENOENT` / `ENOTDIR`, stderr warning on anything else. Still never throws — throwing from module load would reinstate exactly the v0.9.1 crash this PR exists to prevent.

- Add a build-then-load integration test that catches dev-vs-bundle divergence. The unit tests here exercise `resolveCommandVersion(mockUrl)` directly; a regression that reverts to `createRequire(...).require("../../package.json")` would still pass them because the old pattern works fine under `tsx`. Only a test that loads the actual `dist/core-*.mjs` catches the ship-blocking shape.
- `packages/server/src/app.ts:88-98` has its own `resolveCommandVersion` with the same `"0.0.0"` fallback pattern. Same class of bug, different file. Deserves a matching fix.

## Test plan

- [x] `pnpm check` (biome)
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub typecheck`
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub test src/__tests__/resolve-command-version.test.ts` — 7/7 pass:
  - dev-source layout (walk up two)
  - bundled dist layout (walk up one)
  - deeper dist/cli layout (walk up two)
  - sibling non-matching package.json (skip and keep walking)
  - `"unknown"` fallback when nothing matches
  - malformed JSON surfaces a stderr warning but keeps walking
  - EACCES on an intermediate manifest (Unix only) surfaces a warning but does not throw
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub build` succeeds
- [x] `node packages/command/dist/cli/index.mjs --version` → `0.9.2` (was `Cannot find module '../../package.json'` on v0.9.1)

### DIFF
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.9.1",
+  "version": "0.9.2",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/command/src/__tests__/resolve-command-version.test.ts
+++ b/packages/command/src/__tests__/resolve-command-version.test.ts
@@ -1,0 +1,137 @@
+import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { pathToFileURL } from "node:url";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { resolveCommandVersion } from "../core/version.js";
+
+const PACKAGE_NAME = "@agent-team-foundation/first-tree-hub";
+
+describe("resolveCommandVersion", () => {
+  let root: string;
+
+  beforeEach(() => {
+    root = mkdtempSync(join(tmpdir(), "ftHub-version-"));
+  });
+
+  afterEach(() => {
+    rmSync(root, { recursive: true, force: true });
+  });
+
+  // Dev layout: `tsx src/cli/index.ts` runs source files, so this module
+  // resolves from `packages/command/src/core/version.ts` and must walk up
+  // two levels to `packages/command/package.json`.
+  it("reads the version from the dev-source layout (two levels up)", () => {
+    const pkgDir = join(root, "packages", "command");
+    const moduleDir = join(pkgDir, "src", "core");
+    mkdirSync(moduleDir, { recursive: true });
+    writeFileSync(join(pkgDir, "package.json"), JSON.stringify({ name: PACKAGE_NAME, version: "9.9.9" }));
+    const moduleUrl = pathToFileURL(join(moduleDir, "version.ts")).href;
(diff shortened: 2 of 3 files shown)


yuezengwu:
No follow-up action is needed here. PR #144 was a narrow command-package startup fix, it was already merged, and the gardener result is a low-severity `INSUFFICIENT_CONTEXT` caused by missing classifier context rather than a product-context objection. I checked the relevant Agent Hub CLI context and this does not change decisions, constraints, ownership, or workspace relationships, so no Context Tree update is needed.

This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #189, opened 2026-04-29
feat(runtime): multi-runtime support — codex peer to claude-code

yuezengwu:
## Summary
- Adds OpenAI Codex as a first-class agent runtime alongside Claude Code, with per-machine capability probing, a server-side availability gate, in-band repair on WS bind mismatch, and per-agent runtime selection in the web UI.
- Migration 0026 adds `agents.runtime_provider` (NOT NULL, default `claude-code`); pre-existing rows back-fill to `claude-code` so the change is non-breaking on rollout.
- Web `/clients` gains a runtime capability matrix (lazy-loaded per row), and per-agent panels expose a provider/model selector + re-bind dialog. Settings now only renders Bindings (capability info moved to `/clients`).

## Notable design points
- **Capability semantics** — probes emit `state ∈ {ok, unauthenticated, missing, error}` with a derived `available` flag. Server `ensureClientSupportsRuntimeProvider` keys off `available === true`, so a reported `missing`/`error` blocks creation/rebind even if the entry exists, while `unauthenticated` is allowed (operator fixes login on the machine).
- **Codex auth-mode default** — Codex CLI's two auth modes accept different model slugs (`gpt-5-codex` family is rejected by ChatGPT-account auth). Hub stores `model: ""` by default and `buildCodexThreadOptions` only emits the field when the operator chose one, letting the SDK pick a slug that matches the runtime auth mode.
- **In-band repair** — bind handshake compares `runtimeType` against `agents.runtime_provider`; mismatch returns `agent:bind:rejected { reason: "runtime_provider_mismatch" }` and the client `decideRepairForBindReject` returns a `restart` action (P2 minimal — full hot-swap is a follow-up).
- **CLI flow** — `agent create --runtime <provider>` now forwards the choice to the server; `client start` runs `reconcileLocalRuntimeProviders` pre-flight (rewrites stale `agent.yaml::runtime`) and `uploadClientCapabilities` after WS register (the `clients` row is created lazily in the register handshake).

## Test plan
- [x] `pnpm check` clean
- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 898 cases pass (server 476, client 234, shared 130, command 58)
- [x] 8 new test files cover: server capability gate (state branches × create/rebind × force × human skip), WS bind mismatch handshake, `PATCH /clients/:id/capabilities` API, runtime-provider + capability schemas, capability probes (claude-code OAuth/API key/unauth + codex API key/auth.json/HOME fallback), codex `buildCodexThreadOptions` model-empty-vs-set, `decideRepairForBindReject` table, `reconcileLocalRuntimeProviders` + `uploadClientCapabilities`
- [ ] Manual: create a codex agent on a client that probed `state: ok` for codex → bind succeeds end-to-end
- [ ] Manual: create a codex agent on a client that reported codex `missing` → POST returns 400 with availability error; `--force` overrides
- [ ] Manual: existing claude-code agents bind unchanged after applying migration 0026
- [ ] Manual: agent UI re-bind dialog displays provider-correct unauthenticated hint (`claude login` vs `codex login`)

## Versioning
Bumps `@agent-team-foundation/first-tree-hub` 0.10.1 → 0.10.2 (consumer-facing CLI tarball; new SDK dep added). Private packages (`shared`/`server`/`client`/`web`) unchanged per repo policy.

### DIFF
--- a/docs/runtime-provider-design.md
+++ b/docs/runtime-provider-design.md
@@ -1,0 +1,735 @@
+# Agent Runtime Provider 接入设计
+
+> **草稿 / 中文版。** 为了便于团队快速评审，本文档临时以中文撰写。讨论收敛后将重写为英文版本，以符合仓库 "English everywhere on GitHub" 约定。
+
+**状态：** 草稿 — 待评审
+**范围：** 引入 `runtimeProvider` 抽象层，把 Hub 现有"硬编码 Claude Code"演进为多 runtime 多态。第一个新接入是 OpenAI Codex CLI（基于官方 `@openai/codex-sdk@0.125.0`）。
+**预研基础：** 私人 plan 文件 `~/.claude/plans/1-local-computer-agent-runtime-hazy-alpaca.md` —— 含 Phase A 离线 SDK shape 探针、Phase B 端到端 9 项行为验证、6 条 footgun、UX 决策 U1–U6 已敲定。
+
+---
+
+## 术语约定
+
+| 概念 | 用词 | 含义 |
+|---|---|---|
+| Runtime 多态选择 | **runtime provider** | 一种 LLM CLI runtime 的标识，Hub 端 zod literal union（U5）。当前枚举：`"claude-code"` \| `"codex"`。未来追加 `"devin"` / 自研。 |
+| 数据库列 | `agents.runtime_provider` | snake_case；权威字段 |
+| TypeScript 字段 | `runtimeProvider` | camelCase；wire / Zod / JS |
+| Client 能力 | **capabilities** | client 启动探测后上报的 "我装了哪些 SDK + 鉴权状态" 快照 |
+| 切 client / 切 provider | **re-bind** | 统一术语；走同一个对话框（U2） |
+| Hub 内部 LLM 子进程承接者 | **handler** | 既有概念，每个 runtime provider 对应一个 handler factory |
+
+---
+
+## 背景
+
+Hub 现状是「单一 Claude Code 硬编码 + 注册表多态机制半就位」：
+
+- `registerHandler(type, factory)` 在 [packages/client/src/runtime/handler.ts:170](../packages/client/src/runtime/handler.ts) 已经在
+- `agent_presence.runtimeType` 字段 schema 注释里早就预留 `"claude-code" | "codex" | "devin"`，但只有 claude-code 一个实现
+- 本地 `agent.yaml::runtime` 字段默认 `"claude-code"`，Hub 端无对应权威字段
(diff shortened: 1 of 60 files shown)


baixiaohang:
Reviewed the codex handler against the runtime/handler message contract — the architectural split holds up (runtime owns trigger / inReplyTo / mentions / `[From:]` / participants, codex delegates correctly via `formatInboundContent` + `forwardResult`). Findings below are all `codex.ts` polish / parity gaps with `claude-code.ts`, none are blockers.

**`turn_end` ordered before `forwardResult`** — `packages/client/src/handlers/codex.ts:333-346`. claude-code emits `turn_end` *after* the forward succeeds, so the status reflects actual delivery. Codex fires on `turn.completed` directly, so a forward failure leaves the admin events stream showing `turn_end: success` while no chat message was sent. Move the emit out of the per-event branch and fire once at the end of `runTurn`, after forward.

**`forwardResult` catch only logs** — `packages/client/src/handlers/codex.ts:371-375`. claude-code emits `kind: "error", source: "runtime"` + `turn_end: error` (claude-code.ts:608-618) so failures are visible in the events stream. Codex's `sessionCtx.log` doesn't surface anywhere observable. Pair with the fix above.

**`reportSessionCompletion()` on partial-error turns** — `packages/client/src/handlers/codex.ts:370-377`. claude-code gates on `result.subtype === "success"`. Codex fires whenever `accumulated.trim()` is non-empty, even if the turn errored mid-stream. Cooldown-gated server-side, but the semantic is off.

**`assistant_text` not filtered for empty/whitespace** — `packages/client/src/handlers/codex.ts:217-222`. claude-code skips empty text (claude-code.ts:237-243). Without it, whitespace-only `agent_message` items become noise events.

**`accumulated += text` with no separator** — `packages/client/src/handlers/codex.ts:328`. If Codex emits multiple `agent_message` items in one turn the text fuses. Worth a `\n\n` join.

**`as never` in `processItem`** — `packages/client/src/handlers/codex.ts:208-212`. Repo rule: no `as` casts unless unavoidable + commented. Either inline the function or use the SDK's exported `ThreadItem` type directly.

### Note on trigger model (not action item)

The runtime's `currentTrigger` is a per-chat singleton: `routeMessage` writes, `forwardResult` reads + clears. Works cleanly for claude-code's streaming-input model. Codex's run-to-completion + buffered drain (`queuedMessages`) means a mid-turn inject can overwrite the trigger before the first turn's `forwardResult` reads it, so the reply threads to the follow-up message instead of the original; and the merged drain turn finds the trigger cleared so its reply has no `inReplyTo` at all.

Production impact is small (cosmetic threading, doesn't affect cross-chat routing), so leaving as-is for this PR. Flagging in case a third runtime later wants to deviate from claude-code's streaming model — the contract assumes that pattern.

baixiaohang (approved):

PR #190, opened 2026-04-29
feat(web): public landing page at unauth /

Gandy2025:
## Summary

- Renders a public landing page when an unauthenticated visitor hits `/`. Authenticated users still go straight to WorkspacePage — the canonical `/` URL is stable across auth states.
- Three sections, no illustrations, restrained accent: sticky nav (brand + theme toggle + Sign in) → hero (display headline + subtitle + Get Started CTA → `/login`) → three user-value statements → footer (`@ first-tree` parent-brand attribution + GitHub link).
- Pinned to first-tree.ai's palette — near-black `#040404` + cool-light `#e8ecf4` + the shared green accent — independent of the dashboard's light/dark toggle, so the marketing entry point always reads as the parent brand.
- Fixes deep-link redirect-back along the way: unauthenticated `/agents`, `/settings`, etc. still go to `/login`, but now stash the original location in router state so LoginPage returns the user there post-login (previously hard-coded `/`).

## What changed

| File | Change |
|---|---|
| `packages/web/src/pages/landing/{index,nav,hero,features,footer}.tsx` | New (5 files) — page shell + sections |
| `packages/web/src/auth/require-auth.tsx` | Renders `<LandingPage />` inline at `/` for unauth; passes `state={{ from: location }}` on redirect |
| `packages/web/src/pages/login.tsx` | Reads `state.from` via typed guard (no `as` cast), redirects there on success; refuses to bounce back to `/login` to avoid loops |
| `packages/web/src/index.css` | New marketing-tier typography tokens (`--text-display` `clamp(40-60px)`, `--text-headline`, `--text-lead`); new `.landing-marketing` scope pinning the surface palette |
| `packages/web/src/components/ui/theme-toggle.tsx` | Adds `focus-visible:ring-*` (a11y bonus that benefits dashboard chrome too) |
| `biome.json` | Excludes `.gstack/` (already in .gitignore — local browse-skill scratch dir) |

## Why now

- Hub had no public-facing entry: anyone hitting `/` was bounced to `/login`. The landing page introduces a value-first pitch that mirrors first-tree.ai's tone, giving the project somewhere to send stars, contributors, and curious operators.
- The features section is **user-value first**, not feature-first. Each headline names a problem the visitor is feeling today; the body describes the relief Hub delivers, in plain language. Implementation details (UUID v7, fan-out on write, Bearer tokens) belong in the docs, not on the landing page.

## Constraints honored

- ✅ `pnpm typecheck` (incl. `lint:tokens` design-token guardrail) passes
- ✅ `pnpm test` (32 vitest tests) passes
- ✅ `pnpm check` (biome, 465 files) clean
- ✅ Zero new runtime dependencies — uses lucide-react and react-router 7 already pinned
- ✅ All design-token guardrails (no raw px / hex / Tailwind palette / disallowed font-weights / inline font props / Tailwind default text-sizes) — semantic tokens only
- ✅ TypeScript: no `as` casts, no `any`; `readFromPath()` narrows `location.state` via type guards

## Independent code review

Ran `codex review` on the scoped diff — surfaced 4 findings (1 P1 + 3 P2) that the original implementation missed:

1. **[P1]** Deep-link state not preserved across login (pre-existing in `LoginPage` since #9, but I added a misleading comment claiming otherwise) — **fixed**
2. **[P2]** Copy drift: "your Context Tree repo" too narrow (README says any GitHub repo following the convention) — **fixed**
3. **[P2]** Copy drift: "built with Claude Code" oversold — **fixed** (later subsumed by the value-first rewrite)
4. **[P2]** A11y: `<header>` lacked a `<nav>` landmark; focus-visible inconsistent across CTAs — **fixed** (added `<nav aria-label="Primary">`, focus rings on all landing links + ThemeToggle)

## Out of scope

Architecture diagram, Quick Start code block, deploy matrix, pricing, blog, docs site, i18n, OG image / SEO meta polish, dashboard route restructure. Hero stats strip and similar live-data widgets wait for a public `/api/v1/public/stats` endpoint.

## Test plan

- [x] Desktop (1440), tablet (768), mobile (375) viewport screenshots — all render correctly
- [x] CTA `Get Started` → `/login`
- [x] Footer `@ first-tree` → https://first-tree.ai (new tab, `rel="noreferrer noopener"`)
- [x] Footer `GitHub →` → repo (new tab, `rel="noreferrer noopener"`)
- [x] Unauth `/agents` redirects to `/login` and stashes `from` in router state (verified via `window.history.state.usr.from.pathname === "/agents"`)
- [x] `<nav aria-label="Primary">` landmark present (verified via DOM query)
- [x] Focus-visible rings render on all landing links + ThemeToggle
- [x] CI: lint / typecheck / test green on this PR
- [ ] CI: branch-name check passes (matches `^(feat|fix|refactor|test|docs|chore)/`)

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -32,6 +32,15 @@
     }
   },
   "files": {
-    "includes": ["**", "!**/node_modules", "!**/dist", "!**/.turbo", "!**/drizzle", "!**/.venv", "!**/*.code-workspace"]
+    "includes": [
+      "**",
+      "!**/node_modules",
+      "!**/dist",
+      "!**/.turbo",
+      "!**/drizzle",
+      "!**/.venv",
+      "!**/*.code-workspace",
+      "!**/.gstack"
+    ]
   }
 }
--- a/packages/web/src/auth/__tests__/redirect-from-state.test.ts
+++ b/packages/web/src/auth/__tests__/redirect-from-state.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from "vitest";
+import { readFromPath } from "../redirect-from-state.js";
+
+/**
+ * `readFromPath` is the post-login redirect-target extractor for the
+ * password flow (the OAuth flow uses the server-validated `?next=` query
+ * instead). The fixtures below pin both the type-narrowing branches and
+ * the security rules — anything that would let a tampered `state` send
+ * the browser somewhere unexpected after auth must be rejected here.
+ *
+ * Mirrors the safety contract in
+ * `packages/shared/src/__tests__/safe-redirect.test.ts`. If a new
+ * rejection arrives there, add the matching case here.
+ */
+
+const wrap = (from: unknown): unknown => ({ from });
+
+describe("readFromPath", () => {
+  describe("type narrowing", () => {
+    it("accepts a router-shaped Location with just pathname", () => {
+      expect(readFromPath(wrap({ pathname: "/agents" }))).toBe("/agents");
+    });
+
+    it("assembles pathname + search + hash when present", () => {
+      expect(readFromPath(wrap({ pathname: "/agents", search: "?q=1", hash: "#top" }))).toBe("/agents?q=1#top");
+    });
+
+    it("ignores non-string search and hash silently", () => {
+      expect(readFromPath(wrap({ pathname: "/agents", search: 42, hash: null }))).toBe("/agents");
+    });
+
+    it("returns null for null / undefined / primitives", () => {
+      expect(readFromPath(null)).toBeNull();
+      expect(readFromPath(undefined)).toBeNull();
+      expect(readFromPath("/agents")).toBeNull();
+      expect(readFromPath(42)).toBeNull();
+      expect(readFromPath(true)).toBeNull();
+    });
+
+    it("returns null when `from` is missing", () => {
+      expect(readFromPath({})).toBeNull();
+      expect(readFromPath({ other: "/agents" })).toBeNull();
+    });
+
+    it("returns null when `from` is not a Location-shaped object", () => {
+      expect(readFromPath(wrap(null))).toBeNull();
+      expect(readFromPath(wrap("/agents"))).toBeNull();
+      expect(readFromPath(wrap(42))).toBeNull();
+    });
+
+    it("returns null when `from.pathname` is missing or non-string", () => {
+      expect(readFromPath(wrap({}))).toBeNull();
+      expect(readFromPath(wrap({ pathname: 42 }))).toBeNull();
+      expect(readFromPath(wrap({ pathname: null }))).toBeNull();
+    });
+  });
+
+  describe("loop break", () => {
+    it("refuses to bounce back to /login itself", () => {
+      expect(readFromPath(wrap({ pathname: "/login" }))).toBeNull();
+    });
+
+    it("still rejects /login with a search/hash decoration", () => {
+      // safeRedirectPath would accept `/login?next=foo` shape-wise, but
+      // returning that here would create a tight redirect loop the moment
+      // login completes.
+      expect(readFromPath(wrap({ pathname: "/login", search: "?x=1" }))).toBeNull();
+    });
+  });
+
+  describe("safeRedirectPath defense-in-depth", () => {
+    it("rejects scheme-relative URL bypass at the assembled level", () => {
+      // A malicious `from.pathname` of `//evil.com` would be a syntactic
+      // `pathname`-string but is interpreted by browsers as an authority.
+      expect(readFromPath(wrap({ pathname: "//evil.com/anything" }))).toBeNull();
+    });
+
+    it("rejects backslash-prefixed authority bypass", () => {
+      expect(readFromPath(wrap({ pathname: "/\\evil.com" }))).toBeNull();
+    });
+
+    it("rejects absurdly long paths (256-char cap)", () => {
+      const long = `/${"a".repeat(300)}`;
+      expect(readFromPath(wrap({ pathname: long }))).toBeNull();
+    });
+
+    it("rejects paths containing characters outside the safe set", () => {
+      // safeRedirectPath only allows [A-Za-z0-9_-./?=&%#]. A space is not
+      // in the set and would otherwise be a soft injection point.
+      expect(readFromPath(wrap({ pathname: "/agents foo" }))).toBeNull();
+    });
+
+    it('honors "/" as a valid (default) destination, not a rejection sentinel', () => {
+      // The implementation converts safeRedirectPath's substituted "/"
+      // back to null EXCEPT when the original assembled string was
+      // actually "/". This case verifies that exception fires.
+      expect(readFromPath(wrap({ pathname: "/" }))).toBe("/");
+    });
+  });
+});
--- a/packages/web/src/auth/redirect-from-state.ts
+++ b/packages/web/src/auth/redirect-from-state.ts
@@ -1,0 +1,36 @@
+import { DEFAULT_SAFE_REDIRECT, safeRedirectPath } from "@agent-team-foundation/first-tree-hub-shared";
+
+/**
+ * Read the `{from: Location}` state set by `RequireAuth` when it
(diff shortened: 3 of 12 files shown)


baixiaohang (commented):
## Review

CI is green and the implementation follows the repo's conventions (token-only styling, no `as`, a11y landmarks). A few things worth discussing before merge.

### P1 — must fix

**1. The "deep-link redirect-back" fix doesn't apply to the GitHub OAuth path.**

The PR description highlights restoring `from` location across login, but it only covers the password form. The GitHub button is a full-page navigation:

- `LoginPage` line 43: `<a href="/api/v1/auth/github/start">` — React Router's `location.state.from` is dropped here (state lives in history, not the URL).
- After the round-trip, `OAuthCompletePage` reads `next` from the URL fragment (`oauth-complete.tsx:27`) — but `next` was never set, so the user always lands on `/`.

The server already supports `?next=` and validates it via `safeRedirectPath` (`packages/server/src/api/auth/github.ts:46-47`) and bakes it into the state JWT, so this is purely a client-side wiring gap. Suggested fix in `LoginPage`:

```tsx
const githubHref = redirectTo !== "/"
  ? `/api/v1/auth/github/start?next=${encodeURIComponent(redirectTo)}`
  : "/api/v1/auth/github/start";
```

Given that GitHub OAuth is the primary signup path post-#187, advertising this fix while leaving the OAuth flow broken is the part that worries me most.

### P2 — should fix

**2. `ThemeToggle` is a dead button on the landing page.**

`LandingNav` renders `<ThemeToggle />`, but `.landing-marketing` hard-pins `--bg/--fg/--border` and `color-scheme: dark` (`index.css:340-370`) at higher specificity than the `.dark` class on `html`. Result: clicking the toggle on the landing page produces no visible change — it silently flips the dashboard preference for next time.

A toggle that looks broken on a public-facing page is a UX trap. Since the brand decision is "landing is always dark", I'd just remove `<ThemeToggle />` from `LandingNav`.

**3. `readFromPath` skips the existing `safeRedirectPath` helper.**

The type guard in `login.tsx:14-29` is clean, but it doesn't reuse `safeRedirectPath` from `@agent-team-foundation/first-tree-hub-shared`. The OAuth flow uses it as defense-in-depth; the password flow now diverges. Suggest assembling `${pathname}${search}${hash}` and passing it through `safeRedirectPath` — you also pick up the 256-char cap and `//` / `\\`-prefix rejection for free. Keeping both auth paths' redirect-target judgement consistent is worth the one-line change.

**4. Features section header doesn't match the cards.**

`features.tsx:48-52`:
> "Three primitives" — "Identity, messaging, and integration. No framework. No orchestration."

But the three cards are *team inbox / reliable delivery / BYO agent* — no identity card. Either add an identity card or rephrase the header to match what's delivered (e.g. "Three things you get out of the box"). Today a careful reader will notice the mismatch.

**5. No unit test for `readFromPath`.**

Pure function, multiple narrowing branches, plus a security rule (refuse `/login` to break the loop) — exactly the shape that benefits from a test. `packages/shared/src/__tests__/safe-redirect.test.ts` is a good template.

### P3 — nice to have

**6. `LandingPage` is statically imported in `require-auth.tsx`.**

Every authenticated user pays for the landing page bundle (LandingPage + lucide icons `ArrowRight`, `Inbox`, `Plug`, `Users`) on every dashboard load. Wrapping it in `React.lazy(() => import(...))` + `<Suspense>` would shave ~5–10 KB off the auth bundle. Small win, optional.

**7. `biome.json` uses `!**/.gstack` rather than `!**/.gstack/**`.**

Matches the style of the existing `!**/node_modules` / `!**/dist` entries, so leaving it alone is fine — flagging only for awareness.

### Minimal merge-ready set

If you want to keep this PR tight:
- **Must:** wire `redirectTo` into the GitHub button's `?next=` (P1).
- **Strongly recommend:** drop `<ThemeToggle />` from `LandingNav` (P2 #2).

The rest (`safeRedirectPath` reuse, Features header copy, the unit test, lazy-load) can land here or as follow-ups.

Gandy2025:
Thanks for the careful read — every finding was right, and pushed `5bc82bb` addresses 6 of 7. Rebased onto latest `main` first; the original branch was off a `main` from before #187 landed, which is exactly why the OAuth path was invisible when I wrote the diff.

### What changed

| # | Severity | Status | Fix |
|---|---|---|---|
| 1 | P1 | ✅ Fixed | `LoginPage` now passes `redirectTo` through `?next=` (URL-encoded) on the GitHub OAuth href. Verified end-to-end: deep link `/agents` → `/login` → button href = `/api/v1/auth/github/start?next=%2Fagents`; direct `/login` → button href = `/api/v1/auth/github/start` (no `?next=`). |
| 2 | P2 | ✅ Fixed | `<ThemeToggle />` removed from `LandingNav`. Comment in `nav.tsx` documents *why* (the `.landing-marketing` specificity story) so future contributors don't re-add it. |
| 3 | P2 | ✅ Fixed | Extracted `readFromPath` to `packages/web/src/auth/redirect-from-state.ts` and routed the assembled path through `safeRedirectPath`. Both auth paths now apply identical redirect validation. The function still returns `null` (not the substituted `/`) for rejected input so the caller's `?? "/"` fires uniformly. |
| 4 | P2 | ✅ Fixed | Header rewritten: **"Three things you stop worrying about"** + "No framework. No orchestration. Just the wiring your team is already missing." Pairs with the value-first cards instead of enumerating mismatched primitives. |
| 5 | P2 | ✅ Fixed | Added `packages/web/src/auth/__tests__/redirect-from-state.test.ts` — 14 cases covering type-narrowing branches, the `/login` loop break, and every `safeRedirectPath` rejection (scheme-relative URL, backslash bypass, length cap, unsafe characters, `/` honored as default destination). |
| 6 | P3 | ✅ Fixed | `LandingPage` is now `React.lazy` in `require-auth.tsx`. `<Suspense fallback>` is a near-black div with the `.landing-marketing` class so the chunk-load flash is invisible against the landing page below. |
| 7 | P3 | Skipped | `!**/.gstack` matches the existing `!**/node_modules` / `!**/dist` style — leaving alone per your note. |

### Verification

- `pnpm typecheck` (incl. `lint:tokens` design-token guardrail) — clean
- `pnpm test` — 46 pass (was 32; +14 from the new `readFromPath` suite)
- `pnpm check` (biome) — clean
- Headless-browser end-to-end on the OAuth wiring + nav controls

Ready for another pass when you have a moment.

baixiaohang (approved):
LGTM — every P1 / P2 addressed, plus the lazy-load on P3 #6.

Verified against the previous round:

- **P1 (OAuth `?next=`)**: `LoginPage` builds the GitHub href from `redirectTo` (which has already passed through `safeRedirectPath` via `readFromPath`) and `encodeURIComponent`-encodes it. End-to-end the path is now: client filters → server re-validates → state JWT → fragment → `OAuthCompletePage` filters again. Closed loop.
- **P2 (`readFromPath` reuse of `safeRedirectPath`)**: clean extraction to `auth/redirect-from-state.ts`, both auth paths now share the same 256-char cap and authority-bypass rejection. The `safe === DEFAULT && assembled !== DEFAULT ? null : safe` round-trip preserves "/" as a legitimate destination while collapsing all rejection paths to `null`. Subtle and correctly tested.
- **P2 (ThemeToggle removed)**: clean removal, comment explains the specificity reasoning so the next person doesn't re-add it.
- **P2 (Features header rewrite)**: "Three things you stop worrying about" + "No framework. No orchestration. Just the wiring your team is already missing." is in the same value-first voice as the cards.
- **P2 (test suite)**: 14 cases — narrowing branches, loop break (incl. `/login` with decoration), scheme-relative `//`, backslash, length cap, unsafe characters, and the `/` corner case. Mirrors the safety contract of the shared `safeRedirectPath` test, which is exactly the right structure.
- **P3.6 (lazy-load)**: the `Suspense` fallback uses `landing-marketing` + `bg-background` so the chunk-load flash is the same near-black as the loaded page — nice touch.
- **P3.7 (`.gstack` glob)**: skipping is fine, matches existing style.

Ship it.

PR #191, opened 2026-04-29
feat(onboarding): GitHub-only /login + auto-migrate legacy users

Gandy2025:
## Summary

- **/login is GitHub-only by default.** Password form retired. Legacy dogfooders are pre-bound to their GitHub identities via a one-off SQL data migration (run separately, **not in this PR**) so they keep their existing org, agents, and history on first GitHub sign-in.
- **Localhost shows a "Dev: skip GitHub" button** that jumps to `/auth/github/dev-callback` for one-click dev sign-in. The route 404s in production (Dockerfile pins `NODE_ENV=production`), so the button is harmless even if the markup ships.
- **dev-callback simplified:** enabled unconditionally in any non-production env; the `FIRST_TREE_HUB_GITHUB_OAUTH_DEV_CALLBACK` env / `devCallbackEnabled` config field removed. `NODE_ENV` stays the single source of truth. **Self-host note:** old `config.yaml` files with `oauth.github.devCallbackEnabled: true` will silently ignore the key (no strict-mode check on `defineConfig`) — leaving it in is harmless and inert.
- **Defense-in-depth index:** partial unique index on `auth_identities (user_id) WHERE provider='github'` ensures no user can hold two github identities — guards any future rebind/merge flow or one-off SQL that misfires.
- **Visual cleanup:** logo + full "First Tree Hub" brand + Back-to-home link + Terms / Privacy footer placeholders. Login card description text removed (button is self-explanatory).
- **render.yaml** declares the two GitHub OAuth env vars (`sync:false`) so the Dashboard surfaces them on the next blueprint sync; credentials are pasted in via Dashboard (already done for staging).

## Test plan

- [ ] **Staging — pre-bound legacy user**: after running the SQL pre-binding against production DB, the dogfooder clicks "Continue with GitHub" → lands on their existing organization with old data preserved.
- [ ] **Staging — new GitHub user**: a brand-new GitHub login → fresh personal team is auto-provisioned, onboarding modal pops on `/`.
- [ ] **Localhost dev flow**: `pnpm dev` → open `localhost/login` → "Dev: skip GitHub" button visible → click → signed in as `devuser`, lands on dashboard.
- [ ] **Production safety**: hit `/api/v1/auth/github/dev-callback?githubId=42&login=anything` against staging → expect 404 (covered by `oauth-flow.test.ts` "rejects /dev-callback in production"; verify in real environment too).
- [ ] **Visual**: `/login` shows logo + "First Tree Hub" + GitHub primary button + Terms/Privacy footer; password form is **not** visible on staging.

### DIFF
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.10.2",
+  "version": "0.10.3",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/server/drizzle/0028_auth_identity_user_github_unique.sql
+++ b/packages/server/drizzle/0028_auth_identity_user_github_unique.sql
@@ -1,20 +1,11 @@
 -- Partial unique index: each user can hold at most one github identity.
 --
--- Closes the race window in `findOrCreateUserFromGithub`'s legacy-bind
--- fallback. The fallback runs:
---   1. SELECT users WHERE lower(username) = login AND no github identity
---   2. INSERT new auth_identities row binding (user, github)
--- Two concurrent OAuth callbacks for the SAME legacy user under TWO
--- DIFFERENT githubIds would each pass step 1 (the partial uniqueness on
--- (provider, identifier) only catches duplicates of the same identifier),
--- and both INSERTs would succeed — leaving one user bound to two
--- distinct GitHub accounts. This index makes the second INSERT fail
--- atomically with a unique-violation, which the caller surfaces as a
--- 401 just like any other auth failure.
---
--- Generalizes beyond the legacy fallback: also forbids accidentally
--- double-binding a regular OAuth user (e.g. via a future "merge accounts"
--- flow that double-fires).
+-- Defense-in-depth. The (provider, identifier) UNIQUE catches duplicates
+-- of the SAME githubId, but does not stop a single user from collecting
+-- multiple DIFFERENT githubIds (e.g. a future "merge accounts" or
+-- "rebind" flow that misfires, or a one-off SQL migration that errs).
+-- This index makes any such double-bind fail atomically with a
+-- unique-violation at the storage layer.
 
 CREATE UNIQUE INDEX IF NOT EXISTS "uq_auth_identities_user_github"
   ON "auth_identities" ("user_id")
--- a/packages/server/drizzle/0028_auth_identity_user_github_unique.sql
+++ b/packages/server/drizzle/0028_auth_identity_user_github_unique.sql
@@ -1,0 +1,21 @@
+-- Partial unique index: each user can hold at most one github identity.
+--
+-- Closes the race window in `findOrCreateUserFromGithub`'s legacy-bind
+-- fallback. The fallback runs:
+--   1. SELECT users WHERE lower(username) = login AND no github identity
+--   2. INSERT new auth_identities row binding (user, github)
+-- Two concurrent OAuth callbacks for the SAME legacy user under TWO
+-- DIFFERENT githubIds would each pass step 1 (the partial uniqueness on
+-- (provider, identifier) only catches duplicates of the same identifier),
+-- and both INSERTs would succeed — leaving one user bound to two
+-- distinct GitHub accounts. This index makes the second INSERT fail
+-- atomically with a unique-violation, which the caller surfaces as a
+-- 401 just like any other auth failure.
+--
+-- Generalizes beyond the legacy fallback: also forbids accidentally
+-- double-binding a regular OAuth user (e.g. via a future "merge accounts"
+-- flow that double-fires).
+
+CREATE UNIQUE INDEX IF NOT EXISTS "uq_auth_identities_user_github"
+  ON "auth_identities" ("user_id")
+  WHERE "provider" = 'github';
--- a/packages/server/drizzle/meta/_journal.json
+++ b/packages/server/drizzle/meta/_journal.json
@@ -197,6 +197,13 @@
       "when": 1777593600000,
       "tag": "0027_runtime_provider",
       "breakpoints": true
+    },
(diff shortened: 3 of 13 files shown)


Gandy2025:
@baixiaohang 要你的agent review下不？

baixiaohang (commented):
## Overall

Sound design — GitHub-only collapse + progressive migration: legacy users are auto-bound server-side on `lower(username)` match, and the new partial unique index closes the SELECT→INSERT race in `findOrCreateUserFromGithub` at the storage layer. Test coverage hits the right shapes (legitimate bind, case-insensitive, second-bind rejected, suspended excluded, race-window via direct DB insert). Tightening dev-callback to `NODE_ENV` alone is a simplification win. **No blockers; LGTM modulo two pre-merge checks below.**

---

## ⚠️ Pre-merge nits

### 1. Narrow the legacy-bind target set to *actual* legacy password users
`packages/server/src/services/auth-identity.ts` lines 67–76

Current match: `active` + `lower(username)` equality + zero github identities. That set is **slightly larger than \"legacy password users\"** — any OAuth user who somehow lost their `auth_identities` row (data corruption, manual deletion) would also fall in.

`findOrCreateUserFromGithub` already mints OAuth users with a fixed-prefix placeholder (line 52: `oauth:${randomBytes(32).toString(\"base64url\")}`). Suggest adding to the legacy match:

\`\`\`ts
sql\`\${users.passwordHash} NOT LIKE 'oauth:%'\`
\`\`\`

Sharpens intent (\"only true bcrypt password users\") and pre-empts edge cases if you ever add an OAuth-rebind/unbind flow. Tiny diff, meaningful semantic gain.

### 2. Enumerate takeover-eligible legacy usernames before deploy
The comment dismisses test accounts (`admin`, `test`) as harmless, but that's an unverified assumption. Run once before flipping staging:

\`\`\`sql
SELECT id, username, status, created_at
FROM users
WHERE password_hash NOT LIKE 'oauth:%'
ORDER BY username;
\`\`\`

Cross-check the resulting list against GitHub for collision-prone logins (`admin`, `test`, `root`, etc.). Mitigation if any are dangerous: rename in DB (`admin` → `admin-legacy`) or add an explicit denylist constant. **One-time data action, no code change required.**

---

## 💡 Nice-to-haves (can land later)

### 3. Sunset plan for the legacy-bind path
`packages/server/src/services/auth-identity.ts` lines 48–92 — currently a permanent code path. Once all dogfooders migrate, it's pure attack surface. Consider:
- A sunset comment in code (\"remove after 2026-06-30 if no legacy_password rows remain\").
- A scheduled cleanup task post-merge.

### 4. Strengthen test 1 to assert org continuity
`auto-binds legacy password user when github login matches username` only proves `joinPath === \"returning\"`, which is indirect evidence. Pin down the PR's actual promise (\"keep their existing organization\"):

\`\`\`ts
const memberships = await app.db.select().from(members).where(eq(members.userId, legacy.userId));
expect(memberships).toHaveLength(1);
expect(memberships[0]?.organizationId).toBe(legacy.organizationId);
\`\`\`

### 5. `lower(username)` is a seq scan
No functional index on `lower(users.username)`. Fine at current scale (sub-100 users); if it grows past tens of thousands, add `CREATE INDEX ON users (lower(username))`. **Premature now**, just logging.

### 6. `NODE_ENV !== 'production'` is the only dev-callback gate now
Vitest defaults to `NODE_ENV=test`, so the new behavior works. But `NODE_ENV=production pnpm test` would silently 404 every dev-callback test. Optional guard in `helpers.ts::createTestApp`:

\`\`\`ts
expect(process.env.NODE_ENV).not.toBe(\"production\");
\`\`\`

### 7. Document removal of `devCallbackEnabled` for self-hosters
Old `config.yaml` files with `oauth.github.devCallbackEnabled: true` will silently ignore the key (no strict-mode check on `defineConfig`). Behavior is correct, but worth a one-liner in release notes: \"field removed; leaving it in config is harmless and inert.\"

---

## ✅ What's good
- **Partial unique index landed in the right place** — exactly closes the SELECT→INSERT window in the legacy fallback; schema and migration comments cross-reference each other.
- **Suspended users explicitly excluded** — easy thing to miss in migration logic, and there's a test for it.
- **`legacyBound` surfaced via `app.log.info`** — incident triage is one log query away rather than buried in `auth_identities.metadata` jsonb.
- **Dev-callback gate simplified to `NODE_ENV` only** — one fewer flag = one fewer way to misconfigure prod.
- **Test 5 (`partial unique index forbids ...`) bypasses the service to hit the DB directly** — exactly the right shape for a race-window test; verifies the storage-layer guarantee, not the application-layer flow.

Gandy2025:
@baixiaohang thanks for the thorough review. After discussion we shifted approach: dropped the runtime legacy-bind path entirely in favor of a one-off SQL pre-binding for the small known dogfooder set (data action, run separately before deploy, **not in this PR**). New commit: 7d0c33a.

**Effect on each item:**

| # | Status |
|---|---|
| 1 (NOT LIKE 'oauth:%' filter) | Moot — WHERE clause is gone |
| 2 (enumerate takeover-eligible usernames before deploy) | Still applies — same `SELECT` is what we'll run when preparing the SQL pre-binding |
| 3 (sunset plan) | Moot — the temporary code never ships |
| 4 (assert org continuity in test 1) | Moot — test 1 deleted along with the path |
| 5 (functional index on lower(username)) | Skipped (premature, agreed) |
| 6 (NODE_ENV vitest guard) | Skipped (low-value edge case) |
| 7 (release notes for `devCallbackEnabled` removal) | Will fold into PR description |

**Kept from the original review:** the partial unique index on `(user_id) WHERE provider='github'` — reframed in code/migration comments as general defense-in-depth, guarding any future rebind/merge flow or one-off SQL that misfires. Test 5 still verifies the storage-layer guarantee.

Net diff: +20 / -176, server tests still 10/10 green. PR description still references the now-deleted auto-migrate flow; updating separately.

baixiaohang (approved):

PR #197, opened 2026-04-29
feat(inbox): WS push data plane — replace doorbell + 5s poll

baixiaohang:
## Summary

- Server now pushes inbox entries as `inbox:deliver` WS frames and accepts `inbox:ack` over the same socket, replacing the legacy `new_message` doorbell + 5s HTTP poll for clients that opt in via `client:register.wireCapabilities.wsInboxDeliver`.
- Cross-version compat handled by per-socket capability negotiation: old client + new server / new client + old server / both new all keep working.
- Shared payload assembler (`bundleDeliveryWithSilentContext`) so HTTP poll, push, and reconnect drain go through one code path — no risk of silent-context drift.
- Per-agent in-flight cap (default 32, env `FIRST_TREE_HUB_INBOX_MAX_IN_FLIGHT_PER_AGENT`) with post-bind + post-ack backlog drain to recover from cap hits and disconnect/reconnect cycles.
- Bumps `@agent-team-foundation/first-tree-hub` to 0.10.3.

Implements proposal `hub-inbox-ws-data-plane.20260429`.

## Notable design decisions (vs proposal text)

- `inboxDeliverFrameSchema.message` is a full `clientMessageSchema` (which already carries `precedingMessages`) — single payload shape, identical client-side dispatch for poll and push.
- `inbox:ack` does not carry `agentId`; server resolves the inbox via `boundAgents` so a client cannot ack another bound agent's entry by spoofing the wire.
- `claimAndBuildForPush` claims ALL matching rows per NOTIFY (`LIMIT 8`) so the replyTo cross-chat case (one `(inbox, messageId)` pair → two rows differing only by `chat_id`) doesn't leave the second row stuck `pending` until reconnect.
- `inbox_entries.id` is coerced to JS number at the shared bundle exit — `postgres-js` returns `bigserial` (int8) as a JS string, and the WS push frame schema validates `entryId: z.number()` strictly. Server-side self-validation runs before each `inbox:deliver` send so any future schema drift surfaces immediately as a server-side ERROR log.

## Out of scope (follow-up issues)

- Migrate the remaining `tx.execute(sql\`...\`)` raw queries in `services/inbox.ts` to Drizzle builders so the bigint→string coercion bug class disappears entirely (this PR keeps the workaround).
- Phase 3 cleanup per proposal §3.7 (drop `delivered_at` / `retry_count` columns, retire `resetTimedOutEntries`, remove HTTP `POST /inbox/:id/{ack,renew}`) once WS push has 2 weeks of staging soak.
- WebSocket round-trip integration harness (proposal review issue #8a/8b/8d) — service-level invariants are covered by unit tests; full round-trip exercise lives behind a future harness PR.

## Test plan

- [ ] human → A (manual): default config, expect `wsInboxDeliver` capability negotiated, no 5s `poll error` warns, `inbox_entries.status='acked'` within seconds
- [ ] A → B → A → human chained delivery (manual)
- [ ] mention_only group chat: silent-context bundling preserved when an @mention finally arrives (manual)
- [ ] reconnect drain: kill client mid-conversation, send 3-5 messages while down, restart client, all messages arrive in order (manual)
- [x] service-layer unit tests pass (`pnpm --filter @first-tree-hub/server test -- inbox-ws-push silent-inbox-context`): 8 push-path cases including replyTo dual-row claim, silent-context push, multi-inbox ack scoping, int8→number coercion
- [x] `pnpm typecheck` clean (only known unrelated `@openai/codex-sdk` baseline errors)
- [x] `pnpm check` clean

## Rollback

If the push path misbehaves in staging, pin clients to a build prior to this PR — the server treats any client without `wireCapabilities.wsInboxDeliver=true` as legacy and routes through the unchanged `new_message` doorbell + HTTP poll path. No env-flag flip required.

### DIFF
--- a/packages/client/src/__tests__/session-manager.test.ts
+++ b/packages/client/src/__tests__/session-manager.test.ts
@@ -566,3 +566,157 @@ describe("SessionManager routing guards — dispatch integration", () => {
     await sm.shutdown();
   });
 });
+
+/**
+ * `ackEntry` is the WS-push escape hatch (proposal hub-inbox-ws-data-plane
+ * §3.4): when set, dispatch acks via this callback instead of `sdk.ack`,
+ * keeping push-mode slots from double-acking and leaking the server-side
+ * per-agent in-flight counter (which only decrements on a WS ack matching
+ * a still-`delivered` row).
+ *
+ * The four paths that ack inside dispatch — echo suppression, inject into
+ * an active session, start a new session, resume an evicted session —
+ * must all flow through the injected callback, AND `sdk.ack` must NOT be
+ * called when one is provided. Anything less than that re-introduces the
+ * counter leak.
+ */
+describe("SessionManager ackEntry callback (WS push channel)", () => {
+  function buildSm(ackEntry: (entryId: number) => Promise<void>, handler?: AgentHandler) {
+    const h = handler ?? createMockHandler();
+    const sdk = mockSdk();
+    const sm = new SessionManager({
+      session: { idle_timeout: 300, max_sessions: 10, reconcile_interval_seconds: 300 },
+      concurrency: 5,
+      handlerFactory: () => h,
+      handlerConfig: { workspaceRoot: "/tmp/test" },
+      agentIdentity: {
+        agentId: "agent-1",
+        inboxId: "inbox-agent-1",
+        displayName: "Agent",
+        type: "autonomous_agent",
+        delegateMention: null,
+        metadata: {},
+      },
+      sdk,
+      log: silentLogger(),
+      ackEntry,
+    });
+    return { sm, sdk, handler: h };
+  }
+
+  it("uses the injected callback (not sdk.ack) when starting a new session", async () => {
+    const ackEntry = vi.fn().mockResolvedValue(undefined);
+    const { sm, sdk } = buildSm(ackEntry);
+
+    await sm.dispatch(mockEntry({ id: 1, chatId: "chat-1" }));
+
+    expect(ackEntry).toHaveBeenCalledTimes(1);
+    expect(ackEntry).toHaveBeenCalledWith(1);
+    expect(sdk.ack).not.toHaveBeenCalled();
+
+    await sm.shutdown();
+  });
+
+  it("uses the injected callback when injecting into an active session", async () => {
+    const ackEntry = vi.fn().mockResolvedValue(undefined);
+    const { sm, sdk } = buildSm(ackEntry);
+
+    await sm.dispatch(mockEntry({ id: 1, chatId: "chat-1" }));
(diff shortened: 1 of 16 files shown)


yuezengwu:
## Blocker — push path double-acks every entry; per-agent in-flight counter never decrements

### Trace
1. `agent-slot.ts:288` awaits `sessionManager.dispatch(entry)`, then `agent-slot.ts:289` calls `connection.sendInboxAck(frame.entryId)`.
2. `SessionManager.dispatch` already acks internally — 4 sites in `session-manager.ts` (L169 / L325 / L348 / L422) call `this.ackEntry(entryId, …)` → `session-manager.ts:606` `await this.config.sdk.ack(entryId)`, i.e. **HTTP** `POST /api/v1/agent/inbox/:id/ack` → `inboxService.ackEntry` flips the row `delivered → acked`.
3. The subsequent WS `inbox:ack` frame hits `ackEntryByIdForBoundAgents` (`inbox.ts:391`), whose `WHERE` requires `status='delivered'`. The row is already `acked`, so the UPDATE matches 0 rows and returns `null`.
4. `ws-client.ts:808` early-returns on `!ackedEntry` — **`inboxInFlight` is never decremented**.

### Consequence
The counter monotonically grows by +1 per delivered message. After `inboxMaxInFlightPerAgent` (default 32) sends, `makeInboxPushHandler` short-circuits every NOTIFY (`ws-client.ts:236`) and `drainBacklogForAgent` early-exits on `slotsFree<=0`. Push delivery is dead for that agent until the WS reconnects. The 300s reaper resets rows to `pending`, but the **counter** is per-socket in-memory and is not affected.

### Why tests miss it
`inbox-ws-push.test.ts` exercises the service-layer primitives directly. The `SessionManager.dispatch → sdk.ack → WS ack` round-trip is never executed, so the duplicate ack is invisible. The PR body's manual test (`status='acked' within seconds`) holds for the first 32 messages, after which delivery silently stops.

### Repro
On a fresh client + server with default config, send ≥33 messages to one agent over a single WS session. Observe `inbox:deliver` stops firing and `inbox push: at cap, dropping NOTIFY` debug logs accumulate.

### Suggested fix
Inject the ack channel into `dispatch`. `dispatch(entry, ack?: () => Promise<void>)` calls the supplied callback in place of `sdk.ack`. Poll path passes `() => sdk.ack(entryId)`; push path passes `() => connection.sendInboxAck(entryId)`. Push path then has a single ack — the WS frame — and the server's `ackEntryByIdForBoundAgents` correctly transitions `delivered → acked` and decrements the counter.

Adding an integration case ("send `cap+1` messages, verify push keeps flowing") would have caught this and is worth pinning as the regression test.

yuezengwu (approved):

PR #202, opened 2026-04-29
fix(messaging): break agent↔agent direct-chat reply loops via mention_only

baixiaohang:
## Summary

- Agent↔agent direct chats default to `full` mode on both ends, so every reply wakes the peer unconditionally. Courtesy turns ("ok thanks" / "received") get treated as fresh prompts and the two agents loop forever — the visible bug from the screenshots.
- Encode the rule in the data model: in `direct` chats with **no human participant**, seed both ends as `mention_only` so engagement requires an explicit `@`. Human↔agent direct keeps `full` because in a 1:1 with a person every message is implicitly addressed to the agent — falls out automatically without a "human bypass" exception that would leak into group semantics.
- Migration 0029 backfills the same rule for existing agent-only direct chats so deployed servers stop looping immediately on upgrade.

## Mode matrix (after this change)

| Scenario | agent.mode | source |
|---|---|---|
| human↔agent **direct** | `full` | new rule (has human → full) |
| agent↔agent **direct** | `mention_only` | new rule (all non-human → mention_only) |
| group with human | `mention_only` for non-humans | existing `maybeUpgradeDirectToGroup` |
| group all-agents | `mention_only` for non-humans | existing rule |

## Implementation

- `findOrCreateDirectChat` / `createChat`: pick the seed mode based on whether all participants are non-human.
- Migration `0029_direct_agent_only_mention_only.sql`: one-shot SQL that flips existing agent-only direct chats from `full` → `mention_only`.
- `result-sink`: when the trigger sender is `mention_only` (now true for agent↔agent direct), default-mention them so genuine replies still route back. Human↔agent direct keeps the no-prefix UX.
- `task.ts`: explicitly mention the assignee on task assignment notifications so the system-tasks→assignee direct delivery wakes the recipient under the new defaults.
- `adapter-mapping.ts`: pin `mode: 'full'` when creating Feishu/Slack chats. External IM users map to `human` agents today so this is behaviour-equivalent, but the explicit annotation guards against a future bot-to-bot adapter silently inheriting the `mention_only` rule.
- Dispatcher fan-out (`message.ts:200-208`): unchanged — policy lives entirely in `chat_participants.mode`.

## How the loop is broken (and what each layer does)

The data layer alone breaks the **active-send** loop: agent A replying without `@B` no longer wakes B, so courtesy turns die at the inbox. But the **echo** path (B's substantive reply needing to reach A) is preserved by the `result-sink` change — without that, B's answer to A's question would silently fail. And on the rare cross-chat replyTo case, the runtime's existing `suppressEcho` gate is the last line of defence (verified live: log line at 19:25:42 shows it dropping the spurious "已回复" before it can fire a new turn). Each layer addresses a distinct failure mode rather than redundantly stacking on the same one.

## Known leftover (out of scope, separate issue)

The "已回复" spurious self-narration message from the receiving agent still appears in the UI — it's an LLM/handler-side echo, not server-injected (no `已回复` string in the repo). Under the new defaults the runtime echo-suppress correctly drops it before it triggers a new turn, so it's UI noise rather than a behavior bug. Will track and fix separately.

## Test plan

- [x] `pnpm check && pnpm typecheck` — clean
- [x] `pnpm test` — all suites pass (server 535, client 240, shared 150, web 46, command 67)
- [x] New file [`direct-chat-mention-mode.test.ts`](packages/server/src/__tests__/direct-chat-mention-mode.test.ts) pins B' invariants:
  - agent↔agent direct → both `mention_only` (via `findOrCreateDirectChat` and `createChat`)
  - human↔agent direct → both `full`
  - autonomous_agent ↔ personal_assistant → both `mention_only` (rule keys on `type !== 'human'`, not a whitelist — pins against a future positive-whitelist refactor)
  - `type='group'` unaffected
  - fan-out: agent→agent without `@` → silent row, with `@` → wakes peer
  - human→agent without `@` → still wakes (DX guarantee), agent→human without `@` → still wakes (symmetric)
- [x] [`message-dispatcher.test.ts`](packages/server/src/__tests__/message-dispatcher.test.ts) splits the old "direct = full" baseline into mode-aware cases.
- [x] [`messaging-e2e.test.ts`](packages/server/src/__tests__/messaging-e2e.test.ts) Case A rewrites the no-echo invariant assertion to match the new "loop dies at the inbox layer" semantics (one entry instead of two + suppression).
- [x] WS data-plane (#197) and predictive-activation tests on main pre-existed; their direct-chat seeds now include explicit mentions to keep `notify=true` rows.
- [x] Local CLI run reproduces the bug, confirms the loop closes within 4 messages (see logs in PR thread).

### DIFF
--- a/packages/client/src/__tests__/result-sink.test.ts
+++ b/packages/client/src/__tests__/result-sink.test.ts
@@ -82,7 +82,7 @@ describe("createResultSink — forwardResult enrichment", () => {
     expect(body).toMatchObject({ format: "text", content: "final answer", inReplyTo: "m1" });
   });
 
-  it("omits mentions metadata in a 2-person direct chat (peer is always full-mode)", async () => {
+  it("omits mentions metadata in a 2-person direct chat when peer is full-mode (human↔agent)", async () => {
     const { sink, sendMessage } = buildSink({
       trigger: { messageId: "m1", senderId: "agent-peer" },
       participants: [mkParticipant(ME, "me"), mkParticipant("agent-peer", "peer")],
@@ -92,6 +92,28 @@ describe("createResultSink — forwardResult enrichment", () => {
 
     const body = sendMessage.mock.calls[0]?.[1] as Record<string, unknown>;
     expect(body.metadata).toBeUndefined();
+  });
+
+  it("emits default trigger mention in a 2-person direct chat when peer is mention_only (agent↔agent)", async () => {
+    // Migration 0029 + findOrCreateDirectChat seed both participants as
+    // `mention_only` in agent↔agent direct chats so courtesy replies don't
+    // wake the peer and produce A↔B reply loops. The sink has to
+    // explicitly @-mention the trigger sender to route the genuine reply
+    // back to them; without this branch, B's answer to A's question would
+    // land silently and A would never see it.
+    const peer = mkParticipant("agent-peer", "peer");
+    peer.mode = "mention_only";
+    const me = mkParticipant(ME, "me");
+    me.mode = "mention_only";
+    const { sink, sendMessage } = buildSink({
+      trigger: { messageId: "m1", senderId: "agent-peer" },
+      participants: [me, peer],
+    });
+
+    await sink("final answer");
+
+    const body = sendMessage.mock.calls[0]?.[1] as { metadata?: { mentions?: string[] } };
+    expect(body.metadata?.mentions).toEqual(["agent-peer"]);
   });
 
   it("defaults mentions to [trigger.senderId] in a group chat (Mention-default)", async () => {
--- a/packages/client/src/runtime/result-sink.ts
+++ b/packages/client/src/runtime/result-sink.ts
@@ -11,9 +11,13 @@ import type { AgentIdentity } from "./handler.js";
  * - `inReplyTo` is pulled from the current trigger (the message that kicked
  *   off this turn), so peers that threaded via reply routing see the answer
  *   in their waiting chat — see proposals/hub-agent-messaging-reply-and-mentions §3.4.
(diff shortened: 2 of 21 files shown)


yuezengwu (approved):
LGTM — 数据层固化 mention_only 的思路正确,主入口 + 迁移 + result-sink + task 通知都对得上,无 blocker。

几点 nit / follow-up,不阻塞合并:

1. **adapter-mapping.ts 漏网**:Feishu/Slack 的 `findOrCreateAdapterChat` 直接 insert `chatParticipants` 没传 `mode`,走默认 `"full"`。Migration 0029 一次性修正历史,但未来若有 agent-only 的 p2p adapter chat 仍会漏。建议作为 follow-up issue 跟一下。
2. **命名不统一**:`createChat` 用 `isDirectAgentOnly`,`findOrCreateDirectChat` 用 `isAgentOnly`,统一一下更易读。
3. **PR 描述微调**:"3 layers of defense, any single layer failing no longer cascades" — data-layer `mention_only` 实际只阻止"无 @ 主动发起",echo 场景下 `result-sink` 必须默认 @ trigger sender(否则 reply 路由不回去),所以 echo 收敛仍依赖 LLM/runtime suppress。不是 bug,只是描述别让未来读者高估这一层。
4. **测试可选增强**:`personal_assistant` 类型的 case 没显式 pin,加一行可防止未来误改成白名单 `=== "autonomous_agent"`。

已确认:0029 SQL 的 NOT EXISTS 不会误伤 human↔agent direct;`every` 在空数组场景被 creator 兜底;`ChatParticipantDetail.mode` 在 shared schema 已存在;version bump 0.10.4 → 0.10.5 符合约定。

PR #219, opened 2026-05-06
feat(multi-org): pin createAgent to selected org + cross-org name map

Gandy2025:
## Summary

Follow-up to [#214](https://github.com/agent-team-foundation/first-tree-hub/pull/214) (decouple-client-from-identity). Closes the gap where the web could not actually direct a write to the org the user had selected, and where the Computers tab dropped to raw UUIDs for any agent outside the currently selected tenant.

Four interlocking fixes (all six files were uncovered while debugging a single user-reported "I created agents in atf but they showed up in gandy02's team" thread):

- **web:** `NewAgentDialog` + workspace inline onboarding now include `currentMembership.organizationId` in the `POST /admin/agents` body. Without it the server fell back to the JWT default org, which was non-deterministic across logins.
- **server (`POST /admin/agents`):** accept `?organizationId=` as a fallback. Precedence is now `body > query > JWT default`. The api-client `decoratePath` injection added in #214 had no effect on writes until now. `requireMemberInOrg` still re-checks the live membership against the DB, so this does not weaken the auth boundary — an admin in org A passing `?organizationId=B` still 403s.
- **server (`auth.login`):** password-login member pick gets `ORDER BY created_at DESC, id DESC`. The previous `LIMIT 1` had no ORDER BY, so a multi-org user got an implementation-defined default org each login. `members.id` is uuidv7 (sortable) and serves as the tiebreak.
- **web (`useAgentNameMap` / `useAgentIdentityMap`):** layer `/me/managed-agents` (cross-org, user-scoped) on top of `/admin/agents` (org-scoped) so the Computers `BOUND AGENTS` panel resolves UUIDs that belong to the user's other-org agents — a single client is allowed to host agents from multiple orgs post-#214, but the name resolver had not caught up.
- **shared (`createAgentSchema`):** `organizationId` gains `.min(1)` so an empty string cannot shadow the query/JWT fallback chain (caught in independent review pass).
- **command:** bumped to `0.10.10` per the repo's "any PR touching server/web/shared bumps the published CLI" rule.

## Test plan

- [x] `pnpm check` clean
- [x] `@agent-team-foundation/first-tree-hub-shared` 150/150 unit tests pass
- [x] `@first-tree-hub/web` 53/53 unit tests pass
- [ ] Server integration suite (testcontainers + PG) — not run locally; CI to cover. Tests not added in this PR; the changes affect existing `auth.login` member pick + `POST /admin/agents` org resolution paths and a follow-up should pin the new precedence + ORDER BY behavior with regression tests
- [ ] Manual web verification post-deploy: create one agent in each of three orgs, confirm each lands in the right tenant; confirm `BOUND AGENTS` panel shows real names rather than UUIDs after a multi-org bind

### DIFF
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.10.9",
+  "version": "0.10.10",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/server/src/api/admin/agents.ts
+++ b/packages/server/src/api/admin/agents.ts
@@ -158,11 +158,19 @@ export async function adminAgentRoutes(app: FastifyInstance): Promise<void> {
   app.post("/", async (request, reply) => {
     const scope = memberScope(request);
     const body = createAgentSchema.parse(request.body);
+    const { organizationId: queryOrganizationId } = listAgentsFilterSchema.parse(request.query);
     // Realtime role probe — JWT `role` claim is a hint for the default org
     // only. An admin in org A cannot exercise admin powers when creating
     // an agent in org B by passing `body.organizationId = orgB`
     // (decouple-client-from-identity §4.5).
-    const targetOrgId = body.organizationId ?? scope.organizationId;
+    //
+    // Org resolution precedence: body > query > JWT default. The query-string
+    // fallback exists so the api-client `decoratePath` injection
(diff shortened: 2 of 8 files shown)


Gandy2025:
Re-opened as #220 with a convention-conforming branch name (`feat/multi-org-create-and-name-map`). The original branch `claude/pensive-brahmagupta-74b55d` was failing the `Check Branch Name` CI rule (`^(feat|fix|refactor|test|docs|chore)/`). Same commit, same diff.

PR #220, opened 2026-05-06
feat(multi-org): pin createAgent to selected org + cross-org name map

Gandy2025:
## Summary

Follow-up to [#214](https://github.com/agent-team-foundation/first-tree-hub/pull/214) (decouple-client-from-identity). Closes the gap where the web could not actually direct a write to the org the user had selected, and where the Computers tab dropped to raw UUIDs for any agent outside the currently selected tenant.

Four interlocking fixes (all six files were uncovered while debugging a single user-reported "I created agents in atf but they showed up in gandy02's team" thread):

- **web:** `NewAgentDialog` + workspace inline onboarding now include `currentMembership.organizationId` in the `POST /admin/agents` body. Without it the server fell back to the JWT default org, which was non-deterministic across logins.
- **server (`POST /admin/agents`):** accept `?organizationId=` as a fallback. Precedence is now `body > query > JWT default`. The api-client `decoratePath` injection added in #214 had no effect on writes until now. `requireMemberInOrg` still re-checks the live membership against the DB, so this does not weaken the auth boundary — an admin in org A passing `?organizationId=B` still 403s.
- **server (`auth.login`):** password-login member pick gets `ORDER BY created_at DESC, id DESC`. The previous `LIMIT 1` had no ORDER BY, so a multi-org user got an implementation-defined default org each login. `members.id` is uuidv7 (sortable) and serves as the tiebreak.
- **web (`useAgentNameMap` / `useAgentIdentityMap`):** layer `/me/managed-agents` (cross-org, user-scoped) on top of `/admin/agents` (org-scoped) so the Computers `BOUND AGENTS` panel resolves UUIDs that belong to the user's other-org agents — a single client is allowed to host agents from multiple orgs post-#214, but the name resolver had not caught up.
- **shared (`createAgentSchema`):** `organizationId` gains `.min(1)` so an empty string cannot shadow the query/JWT fallback chain (caught in independent review pass).
- **command:** bumped to `0.10.10` per the repo's "any PR touching server/web/shared bumps the published CLI" rule.

Re-opened from #219 with a convention-conforming branch name (`feat/...`).

## Test plan

- [x] `pnpm check` clean
- [x] `@agent-team-foundation/first-tree-hub-shared` 150/150 unit tests pass
- [x] `@first-tree-hub/web` 53/53 unit tests pass
- [ ] Server integration suite (testcontainers + PG) — not run locally; CI to cover. Tests not added in this PR; the changes affect existing `auth.login` member pick + `POST /admin/agents` org resolution paths and a follow-up should pin the new precedence + ORDER BY behavior with regression tests
- [ ] Manual web verification post-deploy: create one agent in each of three orgs, confirm each lands in the right tenant; confirm `BOUND AGENTS` panel shows real names rather than UUIDs after a multi-org bind

### DIFF
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.10.9",
+  "version": "0.10.10",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/server/src/__tests__/admin-agents.test.ts
+++ b/packages/server/src/__tests__/admin-agents.test.ts
@@ -1,7 +1,10 @@
 import type { FastifyInstance } from "fastify";
 import { describe, expect, it } from "vitest";
+import { members } from "../db/schema/members.js";
+import { organizations } from "../db/schema/organizations.js";
 import { createAgent } from "../services/agent.js";
-import { createAdminContext, useTestApp } from "./helpers.js";
+import { uuidv7 } from "../uuid.js";
+import { createAdminContext, createTestAdmin, useTestApp } from "./helpers.js";
 
 describe("Admin Agents API", () => {
   const getApp = useTestApp();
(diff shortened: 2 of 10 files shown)


baixiaohang (changes requested):
Solid defense-in-depth fix — the four layers attack different time windows of the same root cause and the `.min(1)` schema patch closes the empty-string shadow nicely. A few things worth tightening before merge:

**Should-fix**

1. **Missing regression tests.** The two behaviors most prone to silent regression in future refactors are:
   - `auth.login` member pick honoring `ORDER BY created_at DESC, id DESC` when a user has multiple active memberships.
   - `POST /admin/agents` org precedence: `body > query > JWT default`, including the case where `?organizationId=B` from an admin in org A still 403s through `requireMemberInOrg`.

   The PR description already flags this as a follow-up, but I'd rather pin it in this PR — the auth ORDER BY in particular is a one-line change that's trivially reverted by an unrelated query refactor, and we'd never notice until another "agent landed in wrong team" report.

2. **`organizationId: z.string().min(1)` is technically a contract tightening.** Same-repo callers (`command` + `web`) are bumped together so this is fine in practice, but worth a CHANGELOG note in case any external consumer of `@agent-team-foundation/first-tree-hub-shared` was passing `""` as a "use default" sentinel.

**Nits / non-blocking**

3. `useAgentNameMap` / `useAgentIdentityMap` now fire two independent queries with separate `staleTime` / loading state. On first render with both pending, cross-org UUIDs will still flash as raw UUIDs for a moment before resolving. Acceptable, but if the panel is meant to be "always shows real names," consider gating the returned map on both queries being settled, or moving the merge into a single `useQueries` so they share a loading boundary.

4. The `auth.login` comment correctly notes that `ensureMembership` reactivates a left member without touching `created_at`, so a leave-and-rejoin reads as "first joined." Worth tracking this assumption somewhere durable (design doc or migration note) — if the product ever wants "default org = most recently *joined or rejoined*," this will be the load-bearing constraint to revisit.

5. Minor: in `useAgentNameMap` the comment says "the org-scoped source overwrites them so the more authoritative roster view wins on collision" — worth checking whether that ordering is actually what we want for the Computers \`BOUND AGENTS\` panel specifically. Inside the currently-selected org both sources should agree on \`displayName\` anyway, so this is mostly a tie-breaking question, but a short example in the comment of *why* org-scoped is more authoritative would help future readers.

Overall LGTM once tests for the two server changes are in.

Gandy2025:
Thanks for the careful read. Pushed [b80cf0d](https://github.com/agent-team-foundation/first-tree-hub/pull/220/commits/b80cf0d) addressing the two should-fix items:

**1. Regression tests** — both pinned in this PR rather than as follow-up.
- `auth.test.ts`: a multi-org user re-logging-in via password gets the most-recently-joined active membership as JWT default org. The test creates a second org via `POST /me/organizations` (which runs after the original `createTestAdmin` so its `created_at` is strictly later), re-logs-in via `POST /auth/login`, and asserts `/me member.organizationId` is the second org. Without the `ORDER BY` it would 50/50 pick either.
- `admin-agents.test.ts`: new `org precedence on create` describe with four cases — body > JWT default, query > JWT default, body > query, and cross-org spoof returning 403 via `requireMemberInOrg`. Last case overlaps with `admin-realtime-role.test.ts` but pins the POST + query path specifically (the existing suite covers GET + body, neither of which exercise PR #220's new query-string fallback).

**2. Schema-tightening note** — no `CHANGELOG.md` in this repo, so I've left the PR description's call-out as the durable record. The only behavior change is that a literal empty string `""` for `organizationId` now fails validation; same-repo callers (`command`, `web`) never sent it, and any external consumer of `@agent-team-foundation/first-tree-hub-shared` passing `""` as a "use default" sentinel was already getting confusing 403s downstream from `requireMemberInOrg("")`, so the tightening surfaces the bug at the schema boundary rather than masking it. Happy to add a `CHANGELOG.md` in this PR if that's preferred — it'd start the file fresh.

**Skipped (non-blocking nits per your tags):**
- #3 (loading-state coordination via `useQueries`) — agreed the UUID flash is real, but it's bounded to the first ~50ms and resolves to the right name. Left as-is to keep the diff focused on the routing/binding correctness story; happy to spin off if it's worth doing.
- #4 (durable doc for the `ensureMembership` reactivation assumption) — would belong in `docs/decouple-client-from-identity-design-zh.md` follow-up rather than this PR. The new test comment in `auth.test.ts` and the comment block in `services/auth.ts:75-91` together carry the constraint for now.
- #5 (precedence comment elaboration in `useAgentNameMap`) — same reasoning as #4; can fold into the next pass.

baixiaohang (approved):

PR #222, opened 2026-05-06
fix(multi-org): /admin/agents/activity honors ?organizationId= query

Gandy2025:
## Summary

Two follow-ups to [#220](https://github.com/agent-team-foundation/first-tree-hub/pull/220), both instances of the same anti-pattern: an admin route reading `memberScope(request)` / `requireMember(request)` and pinning to the JWT default org instead of the user-selected org.

### Bug 1 — `GET /admin/agents/activity` ignores `?organizationId=`

[`packages/server/src/api/admin/clients.ts::adminActivityRoutes`](packages/server/src/api/admin/clients.ts) was the one read route still pinned to JWT default org. Audit on every `packages/server/src/api/admin/*.ts` confirmed it was the only remaining read offender.

Every consumer of the React Query `["activity"]` cache silently rendered JWT-default-org runtime data while the dropdown showed a different org. Five UI surfaces affected:

| Surface | File |
|---|---|
| Workspace roster (left rail) | `pages/workspace/roster/index.tsx` |
| Workspace agent context | `pages/workspace/context/agent-context.tsx` |
| Workspace command palette (Cmd-K) | `pages/workspace/palette/command-palette.tsx` |
| Agents tab `RUNTIME` column | `pages/agents.tsx` |
| Computers `BOUND AGENTS` runtime | `pages/clients.tsx` |

### Bug 2 — `POST /admin/agents/:uuid/chats` uses JWT-default-org HUMAN as chat creator

Surfaced live during testing: creating an agent in a non-default org via the inline onboarding crashes at the auto-chat step with `Cross-organization chat not allowed: <uuid>` — the agent was correctly created in the target org (post-#220), but the chat creator was resolved as the JWT-default-org HUMAN, hitting `createChat`'s same-org guard.

Fix routes the creator through `requireMemberInOrg(targetAgent.organizationId)` so the HUMAN agent comes from the target's org. Extends `requireMemberInOrg` to also return `agentId` — additive change, all existing callers keep working.

## Cache safety after Bug 1's fix

Already correct, no client changes needed:

- `auth-context.tsx` runs `queryClient.clear()` on `selectOrganization`
- `refetchInterval: 10_000` self-heals stale snapshots within 10s as backup

## Patches

```diff
+const activityQuerySchema = z.object({ organizationId: z.string().min(1).optional() });

 export async function adminActivityRoutes(app: FastifyInstance): Promise<void> {
   app.get("/", async (request) => {
-    const scope = memberScope(request);
+    const baseScope = memberScope(request);
+    const { organizationId } = activityQuerySchema.parse(request.query);
+    const scope = await resolveAdminScope(app.db, request, baseScope, organizationId);
     const overview = await activityService.getActivityOverview(app.db);
     const runningAgents = await activityService.listAgentsWithRuntime(app.db, scope);
```

```diff
   app.post<{ Params: { uuid: string } }>("/:uuid/chats", async (request, reply) => {
     const { uuid: targetAgentId } = request.params;
     const scope = memberScope(request);
     await assertAgentVisible(app.db, scope, targetAgentId);

-    const member = requireMember(request);
-    const result = await createChat(app.db, member.agentId, {
+    const targetAgent = await agentService.getAgent(app.db, targetAgentId);
+    const probe = await requireMemberInOrg(app.db, request, targetAgent.organizationId);
+    const result = await createChat(app.db, probe.agentId, {
       type: "direct",
       participantIds: [targetAgentId],
     });
```

## Test plan

- [x] `npx biome check packages/` clean
- [x] `admin-agents-activity.test.ts` 3/3 pass (Bug 1: JWT default preserved, target org honored, cross-org 403)
- [x] `admin-agents.test.ts` extended with `chat-create resolves creator in target agent's org` — direct-chat with non-default-org target succeeds and includes the target's org HUMAN as a participant
- [x] Existing 24 server tests in `activity.test.ts` + `admin-agents.test.ts` + `admin-realtime-role.test.ts` all still green
- [ ] Manual web verification post-deploy: switch dropdown → workspace + Agents/Computers tabs rerender; create new agent in a non-default org via inline onboarding → auto-chat step succeeds without cross-organization error

## Versioning

Bumps `packages/command` from `0.10.10` → `0.10.11`.


Gandy2025:
Independent self-review pass — verdict **PASS**, zero critical/should-fix.

Notes for the human reviewer:

- Both fixes are correct. TOCTOU between `assertAgentVisible` and `getAgent` in the chat-create rewrite is benign because `agent.organizationId` is immutable (`updateAgent` and `rebindAgent` both leave it untouched); a racing `deleteAgent` would surface as a 404 rather than data corruption.
- The second `requireMemberInOrg` call after `assertAgentVisible` is technically redundant (the latter already resolves the agent-org membership) but the duplication keeps the route's intent locally readable. Could be flattened later by having `assertAgentVisible` return the membership; out of scope here.
- `requireMemberInOrg`'s return-type expansion (`+ agentId`) is purely additive — grepped every caller (`agents.ts:87/122/174`, `adapter-mappings.ts:40`, `adapter-status.ts:22`, `clients.ts:23`), all destructure and ignore unknown keys.

Two pre-existing issues found while auditing — **not addressed in this PR, flagged for follow-up**:

1. **`PATCH /admin/agents/:uuid` reads JWT-default `role`** ([agents.ts:201-203](https://github.com/agent-team-foundation/first-tree-hub/blob/fix/admin-agents-activity-org-scope/packages/server/src/api/admin/agents.ts#L201-L203)) to gate manager reassignment. Same regression class as Bug 1, lighter blast radius because `assertCanManage` already gates on the target org's role — but the local check still reads the wrong bucket. A multi-org admin in agent's org but member in JWT default would be 403'd incorrectly. Worth a follow-up routing through `requireMemberInOrg(agent.organizationId).role`.
2. **`getActivityOverview` aggregates stay global** while the `agents[]` array is now org-scoped. Internally inconsistent (header may say "running: 7" while roster shows 3); pre-existing behavior, not a regression introduced here. PM call on whether headers should reflect "Hub-wide" or "current org".

baixiaohang (changes requested):
Both bug fixes look correct and the tests are tight. One concrete concern + a couple of consistency nits before merge.

**Should-fix — `resolveAdminScope` leaves `humanAgentId` pointing at the JWT default org**

```ts
// services/access-control.ts (PR branch)
return {
  ...scope,
  memberId: probe.memberId,
  organizationId: requestedOrganizationId,
  role: probe.role,
};
```

`MemberScope.humanAgentId` is **not** rewritten, so the returned scope is a half-rotated object: `(memberId, organizationId, role)` belong to the target org, but `humanAgentId` is still the caller's HUMAN in the JWT default org. Today's only consumer (`listAgentsWithRuntime`) doesn't read it, so the activity route works — but this is exactly the same class of "scope mismatch silently does the wrong thing" bug that #220 / #222 are cleaning up, just one level deeper. The next admin listing route that adopts `resolveAdminScope` and happens to read `scope.humanAgentId` (e.g. for a "highlight my own chats" filter) will reproduce Bug 2 verbatim.

The fix is one line — `requireMemberInOrg` now already returns `agentId`:

```ts
return {
  ...scope,
  memberId: probe.memberId,
  humanAgentId: probe.agentId,
  organizationId: requestedOrganizationId,
  role: probe.role,
};
```

If there is a deliberate reason to keep `humanAgentId` as the JWT default (I can't think of one), please leave a comment explaining the asymmetry — otherwise future readers will assume `MemberScope` is internally consistent.

**Should-fix — make the "no other admin route does this" claim a test, not a comment**

The PR description says an audit confirmed `/admin/agents/activity` is the only remaining read offender. That's a strong claim and exactly the kind of invariant that silently regresses. Two cheap ways to pin it:

1. A repo-level lint (eslint custom rule or even a grep test): "any file in `packages/server/src/api/admin/*.ts` that calls `memberScope(request)` must also call `resolveAdminScope` in the same handler" — or whitelist the exceptions explicitly.
2. Or at minimum, leave a `TODO`/`@invariant` comment on `memberScope` itself pointing future authors at `resolveAdminScope`.

Without one of these, "follow-up to #220" is going to keep happening one route at a time.

**Nit — POST /admin/agents has the same precedence logic in ad-hoc form**

`packages/server/src/api/admin/agents.ts` around the create route (added in #220) does its own `body > query > JWT default` resolution inline; this PR introduces `resolveAdminScope` for the read side. They're not quite the same shape (write path also accepts org from the body), but it would be nice to unify the query parsing at least, so the validator (`activityQuerySchema` here, `listAgentsFilterSchema` there) isn't drifting. Not blocking.

**Nit — chat-create test covers the success path only**

`chat-create resolves creator in target agent's org` confirms the post-fix happy path. Worth one extra case: caller is **not** a member of the target agent's org (e.g. someone added them to the target agent's `organization`-visible roster transitively, or a soft-deleted membership). `requireMemberInOrg` should 403 — and that's the actual security boundary the fix relies on. Two lines to add given the existing helper.

**Nit — extra DB roundtrip in `/chats`**

`assertAgentVisible` already loads the agent row; `agentService.getAgent` immediately reloads it. Not a perf issue at our scale, just an opportunity to thread the org id out of the visibility check if it's ever convenient.

Once `humanAgentId` and the audit-as-test items are addressed, this is good to merge.

Gandy2025:
Pushed [7e4fc81](https://github.com/agent-team-foundation/first-tree-hub/pull/222/commits/7e4fc81) addressing both should-fix items + the chat-create negative case.

**1. `resolveAdminScope` now rotates `humanAgentId` too** — `requireMemberInOrg` already returns `agentId` after the previous commit, so the fix is the one-liner you suggested. Comment block in the diff explains why we don't want a half-rotated scope sneaking out into a future route that reads `humanAgentId` (e.g. an `assertChatAccess`-flavored "highlight my chats" filter).

**2. Audit invariant pinned via `admin-routes-org-scope-invariant.test.ts`.** Every file under `packages/server/src/api/admin/` is checked: any file using `memberScope(request)` must also reference one of `[resolveAdminScope, requireMemberInOrg, assertCanManage, assertAgentVisible, resolveOrganization, resolveOrgId, assertChatAccess]`. Whitelist covers the three JWT-default-only files (`organizations.ts`, `system-config.ts`, `ws-admin.ts`) with a one-line reason each. A doc-comment block on `memberScope` itself points future authors at this rule. The check is grep-based on file content rather than per-handler AST — happy to invest in the latter if you want stronger coverage, but file-level closes the door on the regression mode we've actually been hitting.

**3. Chat-create negative case** — added the test you flagged. Calls `POST /chats` against an agent in an org the caller has zero membership in; the route's existing `assertAgentVisible` 404s before reaching `requireMemberInOrg`, which is the right behavior (404 rather than 403 prevents UUID enumeration). Inline org-provisioning because two `createTestAdmin` calls collapse into the same default org.

**Left for follow-up (your other nits):**
- Unifying the query-parser between activity (`activityQuerySchema`) and agents-write (`listAgentsFilterSchema`) — pure shape consistency, no behavior change.
- The duplicate agent-row read between `assertAgentVisible` and `getAgent` — would require widening the visibility-check return type; clean refactor candidate but bigger surface area than this PR's scope.

Local: `npx biome check packages/` clean, `pnpm typecheck` 9/9 pass, 36/36 server tests for the affected suites. Pushing for CI now.

baixiaohang (approved):

PR #249, opened 2026-05-08
fix(web): scope Settings → Computers to caller's user, not org admin

Gandy2025:
## Summary

- **Root fix**: `/settings/computers` was calling the admin-only audit route `/api/v1/orgs/:orgId/clients`, so members in any team where they weren't admin saw an empty Computers page even though their machines were registered (the 403 was silently swallowed).
- **Server adds** `GET /api/v1/me/clients` (Class A user-scope, returns `listClients({ userId })`); the existing org-admin route stays for a future "team device audit" view.
- **Web** `listClients` calls `/me/clients` — Settings is user-scope; the team picker no longer affects this surface.
- **Settings → Computers UI cleanup** that fell out of the scope move: drop redundant Owner column, default-expand rows (personal users typically have 1–3 computers), plain-text inline actions (no icon/border chunks; keyboard focus ring preserved via `focus-visible:ring`), hide empty bound-agents section, tighten runtime label widths.
- **Team page polish**: new `formatDay` helper so Created columns drop minute precision; Agents table pinned to `table-fixed` with explicit `width` so its first two columns align with the Humans table.
- **Convention update**: extend `docs/http-path-conventions.md` Forbidden list with the role-conditional response-set pattern (the cousin shape this PR's bug implicitly relied on) and add a matching pre-commit checklist item.

## Why this matters

Before this PR, a user looking at Settings → Computers got fundamentally different views depending on their role in the team they happened to be viewing — including a totally empty page if they were a member of someone else's team. That's a UX bug *and* a conceptual one: a `client` is owned by exactly one user (`clients.user_id`), so the page should never depend on org context. See CLAUDE.md "Unified user-JWT auth" + commit 5e7e45a (decouple-client-from-identity).

## Known regression — admin team-device audit no longer reachable from UI

Before this PR, an admin viewing Settings → Computers got an audit-style list of every member's clients in the currently-selected team (via the `/orgs/:orgId/clients` route). After this PR, that surface is replaced with a personal "my computers" view — which is the right fix because Settings is intrinsically user-scope — but it means **admins lose the ability to audit their team's connected machines from the web UI**.

The `/api/v1/orgs/:orgId/clients` route is preserved server-side (and still admin-gated), so no API capability is lost; it just has no UI entrypoint after this PR. A follow-up should park a proper "Team → Devices" tab (under `/team/:org/...`, where it conceptually belongs) to restore the audit view.

> Action item: open a follow-up issue for the team-device audit tab once this PR lands; otherwise `/orgs/:orgId/clients` silently becomes dead code.

## Test plan

- [x] `pnpm test` (server, 615 + 3 new tests, all green)
- [x] `pnpm typecheck` + `pnpm check` (biome) clean
- [x] Local end-to-end: spun up two dev users (`alice` admin of own team, `bob` admin of own team + member of alice's team via invite); verified:
  - `GET /me/clients` returns only the caller's clients regardless of which team they're "in"
  - `GET /orgs/:aliceOrg/clients` for member `bob` still returns 403 (defense-in-depth, route stays admin-only)
  - `GET /orgs/:bobOrg/clients` for admin `bob` still returns the audit view (unchanged)
- [x] No-auth: 401 ✓
- [x] Cross-user isolation: alice's `/me/clients` doesn't leak bob's, vice versa ✓
- [x] Cross-org per-user: a user who is admin in org A and member in org B sees clients from both orgs in `/me/clients` ✓
- [x] UI smoke (localhost:5173/settings/computers, both alice and bob): rows show only the caller's machines, switching team has no effect on the list

## Version

Bumps `packages/command` 0.11.1 → 0.11.2.


baixiaohang (dismissed):
## Design review — endpoint Class placement vs HTTP path conventions

The bug fix is right, but I think the new endpoint is sitting in the wrong Class. Walking the decision tree in `docs/http-path-conventions.md`:

> Q1: Does the operation involve an organization?
>   The list returns the caller's clients across all their orgs → **No** → Class A

Class A's path shape is `/api/v1/me/...`, not the Class C collection root. Direct precedent in `me.ts`:

| resource | personal collection (Class A) | single resource (Class C) |
|---|---|---|
| agents | `/me/managed-agents`, `/me/pinned-agents` | `/agents/:uuid` |
| **clients (this PR)** | **should be `/me/clients`** | `/clients/:clientId` |
| organizations | `/me/organizations` | `/orgs/:orgId/...` |

Side benefit: `clients.ts` reverts to 100% Class C, and its file-level JSDoc stays accurate without edits.

### Concrete changes

**Server** — move handler from `clients.ts` into `me.ts`, after `/me/pinned-agents`:

\`\`\`ts
app.get("/me/clients", async (request) => {
  const { userId } = requireUser(request);
  const { listClients, deriveAuthState } = await import("../services/client.js");
  const list = await listClients(app.db, { userId });
  const refreshExpirySeconds = authService.expiryToSeconds(app.config.auth.refreshTokenExpiry);
  return list.map((c) => ({ /* same shape as before */ }));
});
\`\`\`

**Web** — `packages/web/src/api/activity.ts`:

\`\`\`ts
return api.get<HubClient[]>("/me/clients");
\`\`\`

**Test** — rename to `me-clients-list.test.ts`, point at `/api/v1/me/clients`, and **add a real cross-org case**. The current test claims to verify cross-org collation but seeds both clients under the same \`organizationId\`, so it actually only covers cross-user isolation:

\`\`\`ts
it("returns the caller's clients across all their orgs", async () => {
  // user a is admin in orgA, member in orgB; seed a client in each
  // assert GET /me/clients returns both
});
\`\`\`

This is the core UX promise of the fix (\"regardless of which team they're in\"); currently no test backs it.

### Convention gap worth closing

The discussion that led here surfaced a real gap in \`docs/http-path-conventions.md\`: the decision tree routes single endpoints by scope but doesn't forbid the cousin pattern of *unioning two scopes into one route based on caller role* — which is implicitly what the broken \`/orgs/:orgId/clients\`-from-Computers-page relied on.

Suggesting we add to Forbidden:

> ❌ **Role-conditional response set** — a single route returning different resource collections based on caller role (e.g. members get their own clients, admins get the org's clients). Split into Class A + Class B and let the frontend pick. URLs describe data, not \"what to show whom\".

And to Pre-commit checklist:

> - [ ] Response set depends only on path params + scope, not on caller role

Cheap to add while the convention is fresh.

### Admin audit regression — call out in PR body

The fix correctly removes admin's ability to view teammates' computers from \`/settings/computers\`. \`/orgs/:orgId/clients\` is preserved but has no UI entrypoint after this PR. Worth calling this out as a known regression in the description and parking a follow-up issue for the team-audit tab/page, otherwise that route silently becomes dead code.

### Smaller things

- \`useUserNameMap\` import was dropped here — grep to see if any other page still uses the hook; if not, delete the file.
- The new inline \`<button>\`s skip the \`<Button>\` component. Verify keyboard focus ring is still visible; if the design system has a \`variant=\"link\"\`, prefer that.
- Otherwise the UI cleanup (default-expanded rows, plain-text actions, hidden empty bound-agents, \`formatDay\` extraction, \`table-fixed\` alignment in Team page) all reads well — non-obvious decisions are commented where they belong.

---

Strong fix overall — correct root cause, scoped well. The Class-A move + the missing cross-org test case are the two things I'd want to see before merge.

baixiaohang (approved):
All review points landed cleanly:

- ✅ Endpoint moved to `/me/clients` (Class A) per the decision tree; `clients.ts` reverts to 100% Class C with no JSDoc churn needed
- ✅ Real cross-org test added — and goes one better than I asked, exercising the `role="member"` side org so it proves "user-scope, not org-admin-scope" rather than just "cross-org"
- ✅ `docs/http-path-conventions.md` updated with both the Forbidden item and the matching pre-commit checklist line
- ✅ `useUserNameMap` hook fully removed (file deleted, no orphan callers)
- ✅ Inline action buttons get `focus-visible:ring-1` so keyboard focus stays visible without polluting mouse-hover state
- ✅ PR body now explicitly flags the admin team-device-audit regression with a concrete follow-up action item

Approving.

PR #256, opened 2026-05-09
fix: sync Context Tree remote repo for Context UI

Gandy2025:
## Summary
- make the Context UI use the configured remote Context Tree repo as the default source of truth by syncing it into a Hub-managed readonly checkout
- keep contextTree.localPath / FIRST_TREE_HUB_CONTEXT_TREE_PATH as a local override instead of the production default path
- simplify the unavailable UI so it does not show duplicate warnings, zero update counts, or server env-var instructions
- update Context Tree UI docs and deployment/config docs; bump command package to 0.11.4

## Validation
- pnpm check
- pnpm --filter @agent-team-foundation/first-tree-hub-shared build
- pnpm --filter @first-tree-hub/server exec vitest run src/__tests__/context-tree-snapshot.test.ts
- pnpm typecheck
- pnpm test

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,8 +17,10 @@ FIRST_TREE_HUB_DATABASE_URL=postgresql://firsttreehub:firsttreehub@localhost:543
 FIRST_TREE_HUB_HOST=0.0.0.0
 
 # Optional deployment-level read token for private Context Tree repos configured
-# in Team Settings. Requires repository contents read access only.
+# in Team Settings. Requires repository contents read access only. The token is
+# only applied to repos listed in FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN_REPOS.
 # FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN=
+# FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN_REPOS=agent-team-foundation/first-tree-context
 
 # ┌───────────────────────────────────────────────────────────────────────────┐
 # │ Server — Optional (have defaults or are auto-generated)                  │
--- a/.env.example
+++ b/.env.example
@@ -19,6 +19,10 @@ FIRST_TREE_HUB_HOST=0.0.0.0
 # Context Tree GitHub repository (URL or owner/repo format, optional)
 # FIRST_TREE_HUB_CONTEXT_TREE_REPO=org/first-tree
 
+# Read-only GitHub token for private Context Tree repos.
+# Requires repository contents read access only.
+# FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN=
+
 # ┌───────────────────────────────────────────────────────────────────────────┐
 # │ Server — Optional (have defaults or are auto-generated)                  │
 # └───────────────────────────────────────────────────────────────────────────┘
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -329,7 +329,8 @@ Most environment variables use the `FIRST_TREE_HUB_` prefix. `onboard` also acce
 | `FIRST_TREE_HUB_HOST` | Bind address | `127.0.0.1` |
 | `FIRST_TREE_HUB_JWT_SECRET` | JWT signing key | auto: random generated |
 | `FIRST_TREE_HUB_ENCRYPTION_KEY` | Adapter credential encryption key | auto: random generated |
-| `FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN` | Optional deployment-level read token for private Context Tree repos configured in Team Settings. Only used by the server-managed Context Tree mirror for `https://github.com/...` repos. | — |
+| `FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN` | Optional deployment-level read token for private Context Tree repos configured in Team Settings. Only used by the server-managed Context Tree mirror for allowlisted `https://github.com/...` repos. | — |
+| `FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN_REPOS` | Comma-separated GitHub repo allowlist (`owner/repo`) that may use `FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN`. Required before the token is applied to any org-configured repo. | — |
 | `FIRST_TREE_HUB_WEB_DIST_PATH` | Web static files path | auto-discovered |
 | `FIRST_TREE_HUB_PUBLIC_URL` | Public-facing hub URL. Stamped as the `iss` claim on connect tokens (so `connect <token>` derives the hub URL with no extra arg) and used to build invite-link URLs + the GitHub OAuth callback. **Required in production.** | request `Host` (dev only) |
 | `FIRST_TREE_HUB_GITHUB_OAUTH_CLIENT_ID` | GitHub OAuth App client ID. Enables `/signup` + `/auth/github/start`. Both client id AND secret must be set together. | — |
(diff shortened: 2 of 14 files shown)


baixiaohang (commented):
Direction looks right — promoting "configure a repo URL → it works" over "admin must mount a local checkout" matches the technical plan, and demoting `localPath` / `FIRST_TREE_HUB_CONTEXT_TREE_PATH` to an override is the correct shape. Server-side sync engine + tests are well-structured. Notes below, grouped by priority.

## Must address before merge

### 1. Failed sync has no TTL cache — light DoS surface

In `packages/server/src/services/context-tree-snapshot.ts` (around L579–587):

```ts
const syncResult = await syncPromise;
remoteLastSyncedAt.set(root, Date.now());        // only updated on success
```

If `syncRemoteContextTree` throws on the no-`.git` branch (invalid repo URL / wrong branch / token without contents read), the exception bubbles up to `resolveContextTreeRoot`'s catch, but `remoteLastSyncedAt` is never written. **Every Web request to the Context page then re-triggers a 60s `git clone`.** Multiple users opening the page against an org with a misconfigured binding = server stuck running failing clones in a loop.

Minimum: cache failures with a short TTL too (e.g. 30s), keyed by `root`. Track `remoteLastFailedAt` + reason; within TTL, return unavailable immediately without retrying.

### 2. `managedContextTreePath` uses a 16-hex-char (64-bit) hash for the directory name

L549–552:

```ts
const hash = createHash("sha256").update(`${repoUrl}\0${branch}`).digest("hex").slice(0, 16);
return join(cacheRoot, hash);
```

Hub is multi-tenant — on the team SaaS Cloud each org has its own binding. 64 bits is fine for total volume, but this is **filesystem path placement keyed by user-controlled content**. Either widen to the full hex digest (one-line change) or namespace under `orgId/<hash>` so (a) cross-org bindings can never land in the same directory and (b) a leaked hash isn't reversible to a single repo URL.

### 3. Revert the `packages/command` version bump

`packages/command/package.json` goes 0.11.3 → 0.11.4. Per the Versioning section in `CLAUDE.md`, command should bump only when `command` / `client` / shared-modules-imported-by-command-or-client change. This PR only touches `server` + `web` + `docs` + `shared/src/config/server-config.ts` (a server-only field). Please revert to 0.11.3.

## Suggested (acceptable as follow-up)

### 4. No GC for managed checkouts when binding changes

`data/context-tree-repos/<hash>` is keyed by `(repo, branch)`. When an org changes its Context Tree binding, the old checkout is left on disk forever. On long-lived SaaS deploys this just accumulates. A simple sweep on successful sync (delete subdirectories not accessed in N days) would handle it. At minimum, leave a TODO with a tracking issue.

### 5. Tests cover happy path + one stale fallback, but miss key negative cases

Around L241–318 in `context-tree-snapshot.test.ts`. Missing:

- **`isSafeBranchName` rejection**: e.g. branch = `--upload-pack=evil` or `foo;bar` should return unavailable and **never invoke git**. This is a shell-injection guardrail and needs an explicit assertion.
- **`redactSecret` in error messages**: the askpass-file-doesn't-contain-secret test exists, but the error-path redaction is untested. Construct a failing scenario where a token would surface in stderr and assert the returned `reason` doesn't contain it.
- **Dirty-state recovery on first-clone failure**: if `git clone` fails partway, `root` may exist without `.git`. The unconditional `rm` at L598–606 handles this on retry, but it's untested.
- **Concurrent sync dedup**: the `remoteSyncPromises` dedup logic has no test. Two concurrent requests should share one promise rather than each running a clone.

### 6. `GIT_SYNC_TIMEOUT_MS = 60_000` is tight for real deploys

L362. 60s is short for a medium Context Tree (a few thousand markdown files) on a slow network — clone times out, the snapshot's unavailable detail surfaces a timeout error, bad UX. Either expose as a server config field (`contextTreeSync.syncTimeoutMs`, default 60s) or just bump the default to 120s.

### 7. askpass script is rewritten on every sync, and lives in cacheRoot itself

L633–661:

```ts
const askpassPath = join(cacheRoot, "git-askpass.sh");
await writeFile(askpassPath, ...);
await chmod(askpassPath, 0o700);
```

- Content is constant — write once at boot (or skip if exists).
- It sits in the cacheRoot alongside `<hash>/` subdirectories. Move it to `cacheRoot/.tools/git-askpass.sh` so it's physically separated from managed checkouts (and a future cache-clear sweep can't accidentally remove it).

## Nits

### 8. `redactSecret` regex coverage is narrow

L723–725 only matches `x-access-token:TOKEN@`. Current askpass flow keeps the token in `GIT_PASSWORD` env so git is unlikely to echo it to stderr, but as defense-in-depth add a fallback redaction for GitHub PAT prefixes (`\bghp_[A-Za-z0-9]+`, plus `github_pat_`).

### 9. `existsSync` → async `access()` / `stat()`

Several call sites (e.g. L563) use sync I/O on the request path. Web call rate is low so it's not blocking, but worth converting in a follow-up.

### 10. Web UI may render a token-bearing repo URL

`packages/web/src/pages/context.tsx` L599–606:

```tsx
{snapshot.repo ? `Repo: ${snapshot.repo}` : null}
```

If an admin pastes `https://x-access-token:TOKEN@github.com/org/repo` into Team Settings (a realistic human error), the token is rendered into the Web UI verbatim. Run `redactSecret` (or equivalent) over `snapshot.repo` before displaying, as a last-line defense.

### 11. Docs-vs-code wording mismatch

`docs/user-facing-context-tree-ui-technical-plan.zh-CN.md` calls the managed checkout "readonly checkout", but `git checkout -B` actually allows local writes (nobody writes there, but the git permission isn't enforced). "Managed mirror" matches the implementation more accurately. Doc nit, not a blocker.

Gandy2025:
Thanks for the detailed review. I pushed follow-up fixes in `b699c78` (on top of the earlier allowlist/HTTPS/redaction fixes). Current CI is green: `Lint & Type Check` and `Test` both passed.

Addressed before merge:

1. Failed sync retry loop / light DoS surface
- Added a short failure TTL for first-clone failures (`remoteLastFailures`, 30s). If a repo/branch/token is misconfigured and no usable checkout exists, subsequent requests return unavailable immediately during the TTL instead of launching another clone.

2. Managed checkout path hash length
- Switched managed checkout directory names from 16 hex chars to the full SHA-256 digest.

3. Command package version bump
- Reverted `packages/command/package.json` from `0.11.4` back to `0.11.3`.

Also addressed several suggested/nit items:

4. Negative tests
- Added coverage for unsafe branch rejection, full SHA-256 managed path, first-clone failure caching, incomplete checkout cleanup before retry, and token/error redaction.

5. Sync timeout
- Increased remote sync timeout from 60s to 120s.

6. askpass placement/rewrite
- Moved askpass to `cacheRoot/.tools/git-askpass.sh` and only writes it when missing.

7. Redaction coverage
- Added fallback redaction for URL-embedded credentials and GitHub token prefixes (`ghp_`, `github_pat_`).
- Added last-line Web UI redaction for displayed repo URLs in the unavailable state.

8. Docs wording
- Changed “readonly checkout” wording to “managed mirror”.

Already addressed in the previous push:
- Deployment-level Context Tree GitHub token is only applied to explicit `FIRST_TREE_HUB_CONTEXT_TREE_GITHUB_TOKEN_REPOS` allowlisted repos.
- Org Context Tree repo URLs are HTTPS-only and reject embedded credentials.
- Settings UI no longer advertises SSH URLs.

Left as follow-up:
- Managed checkout GC / stale mirror cleanup.
- Sync backgrounding instead of request-time refresh.
- Replacing remaining sync `existsSync` calls with async filesystem checks.

baixiaohang (approved):

PR #263, opened 2026-05-09
fix: partial agent-config PATCH no longer wipes untouched fields

Gandy2025:
## Summary

- Web UI's partial save (e.g. user only edits MCP) was silently wiping prompt / model / env / gitRepos. Root cause: Zod 4's `.partial()` keeps inner `ZodDefault` wrappers, so omitted fields parsed to their defaults instead of `undefined`, and the service layer's `patch.x ?? current.x` merge replaced the user's saved values with empty defaults.
- Mirror the 5 PATCH-able fields in a new `agentRuntimeConfigPatchShape` without `.default()` and route both `updateAgentRuntimeConfigSchema` and `dryRunAgentRuntimeConfigSchema` through it. Read-side `agentRuntimeConfigPayloadSchema` is untouched — legacy DB rows still parse cleanly.
- Bump `command` 0.11.4 → 0.11.5 per [docs/versioning-and-publishing.md](docs/versioning-and-publishing.md): the changed `shared` module is imported by `packages/command/src/commands/agent-config.ts`.

## Test plan

- [x] `pnpm check` — clean
- [x] `pnpm test` — 680/680 (server suite green; the new regression test in `admin-agent-config.test.ts` is one of them)
- [x] **Bug reproduction:** temporarily reverted the schema change and re-ran the regression test — it failed with `expected '' to be 'stay-put-prompt'`, confirming the test catches the wipe. Re-applied the fix, test passes.
- [ ] Manual web-UI smoke after deploy: open Agent Detail with a non-empty prompt, change only Model, save, reload — prompt should be preserved.

### DIFF
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.11.4",
+  "version": "0.11.5",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/server/src/__tests__/admin-agent-config.test.ts
+++ b/packages/server/src/__tests__/admin-agent-config.test.ts
@@ -50,6 +50,50 @@ describe("Admin agent-config API (Step 2)", () => {
     expect(after.json().payload.prompt.append).toBe("你只会一句话");
   });
 
+  it("PATCH with a subset of fields leaves omitted fields untouched", async () => {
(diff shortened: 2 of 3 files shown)


yuezengwu (approved):
LGTM. 根因分析准确（Zod 4 .partial() 不剥 ZodDefault），修复 scope 最小：只动写路径 schema，读路径零改动，老 DB 行兼容性零风险。回归测试直击核心场景。version bump 符合策略。

Issue #279, opened 2026-05-10
chore(deploys): remove dead FIRST_TREE_HUB_CONTEXT_TREE_REPO env var (orphaned by PR #256)


liuchao-001:
## Context

PR #256 (\`bc2ce0e fix: sync Context Tree remote repo for Context UI\`) moved the Context Tree repo URL from server-wide config to **per-org Team Settings** (`organization_settings.context_tree`). It removed the old Zod field that read \`FIRST_TREE_HUB_CONTEXT_TREE_REPO\` from \`packages/shared/src/config/server-config.ts\` AND removed the \`.env.example\` entry.

Now no code reads this env var anywhere in the codebase (verified via \`git grep\`).

## Symptom

The env var is still set on at least one deployment (staging) — discovered while diagnosing #278 — and gives operators a false sense that the Context Tree repo is configured. It does not affect server behavior (because nothing reads it), but it's misleading config drift.

## Action

- [ ] Identify all deployments that still have \`FIRST_TREE_HUB_CONTEXT_TREE_REPO\` set in their env (at minimum: staging; check dev / prod too).
- [ ] Delete the env var from those deploys.
- [ ] Optionally add a startup-time warning in the server: if it sees \`FIRST_TREE_HUB_CONTEXT_TREE_REPO\` in \`process.env\`, log a one-line WARN like \`"FIRST_TREE_HUB_CONTEXT_TREE_REPO is set but no longer read since PR #256; remove from your deployment env. Per-org Team Settings is now the source of truth."\` — helps the next operator who finds it.

## Related

- #256 — PR that introduced this orphan
- #278 — the actual bug operators hit when they assumed this env var still worked

liuchao-001:
Closing — @liuchao-001 is removing the env var directly from the deploy platform. The optional startup warning (attempted in PR #287) was reviewed and judged unnecessary noise.

PR #280, opened 2026-05-10
fix(server): resolveUserPrimaryOrgId mirrors /me's defaultOrganizationId selection

### LINKED ISSUE #278: fix(server): resolveUserPrimaryOrgId ignores defaultOrganizationId — breaks /context-tree/{info,snapshot} for multi-org users
## Symptom

Multi-org users hit two visible bugs:

1. **Web Context UI** (`GET /api/v1/context-tree/snapshot`) shows the wrong org's tree (or "Team context not configured" when the wrong org is unconfigured).
2. **Hub Client `syncContextTree`** (`GET /api/v1/context-tree/info`) gets back `{repo: null}` even when the user's primary org has a Context Tree configured in Team Settings. Downstream effect: `installFirstTreeIntegration` silently no-ops in `runBootstrap`, so agent skills never land in any chat workspace.

## Reproducer

Account in two orgs (`agent-team-foundation` + `liuchao-001`), with `users.defaultOrganizationId` pointing at `agent-team-foundation`. Context Tree configured in Team Settings ONLY for `agent-team-foundation`.

```bash
$ curl -H "Authorization: Bearer $TOK" https://dev.cloud.first-tree.ai/api/v1/context-tree/info
{"repo":null,"branch":"main"}    # ← expected the agent-team-foundation tree URL
```

## Root cause

`packages/server/src/services/org-settings.ts:273` — `resolveUserPrimaryOrgId`:

```ts
export async function resolveUserPrimaryOrgId(db: Database, userId: string): Promise<string | null> {
  const [row] = await db
    .select({ organizationId: members.organizationId })
    .from(members)
    .where(and(eq(members.userId, userId), eq(members.status, "active")))
    .orderBy(asc(members.createdAt))   // ← picks earliest membership, ignores users.defaultOrganizationId
    .limit(1);
  return row?.organizationId ?? null;
}
```

It picks the user's earliest-created active membership. For users with multiple orgs whose `defaultOrganizationId` is NOT the earliest, this resolves to the wrong org.

The function's docstring acknowledges this is a stop-gap: _"Tightening to 'explicit org selector' is a future change-once-multi-org-clients-arrive concern. (#7)"_ — but multi-org users are here now.

## Fix

`resolveUserPrimaryOrgId` should respect `users.defaultOrganizationId` first, fall back to earliest membership only when no default is set OR the default points at an org the user no longer belongs to.

```ts
export async function resolveUserPrimaryOrgId(db: Database, userId: string): Promise<string | null> {
  // 1. Prefer user's explicit default org if they still have active membership in it.
  const [user] = await db
    .select({ defaultOrganizationId: users.defaultOrganizationId })
    .from(users)
    .where(eq(users.id, userId))
    .limit(1);

  if (user?.defaultOrganizationId) {
    const [membership] = await db
      .select({ organizationId: members.organizationId })
      .from(members)
      .where(
        and(
          eq(members.userId, userId),
          eq(members.organizationId, user.defaultOrganizationId),
          eq(members.status, "active"),
        ),
      )
      .limit(1);
    if (membership?.organizationId) return membership.organizationId;
    // Default points at an org the user no longer belongs to → fall through.
  }

  // 2. Fall back to earliest active membership (existing behavior).
  const [row] = await db
    .select({ organizationId: members.organizationId })
    .from(members)
    .where(and(eq(members.userId, userId), eq(members.status, "active")))
    .orderBy(asc(members.createdAt))
    .limit(1);
  return row?.organizationId ?? null;
}
```

## Blast radius

`resolveUserPrimaryOrgId` is called by exactly two endpoints:

- `packages/server/src/api/context-tree-info.ts`
- `packages/server/src/api/context-tree-snapshot.ts`

Both will benefit from the fix.

## Decision: keep "single org per request" as the contract

The previously-flagged future direction was "every endpoint accepts `?organizationId=`". After discussion: this is unnecessary for now — `client` is bound to one org, switching orgs requires re-binding the client, so respecting `defaultOrganizationId` matches the actual product semantics. The TODO in the function's docstring can be considered closed by this fix (or downscoped to "we'll only need explicit selectors if the client model itself goes multi-org per machine").

## Test cases

- [ ] (a) `defaultOrganizationId` set + that org is an active membership → returns default
- [ ] (b) `defaultOrganizationId` set + membership in that org is `status != "active"` (or absent) → falls back to earliest membership
- [ ] (c) `defaultOrganizationId` is null → returns earliest membership (existing behavior)
- [ ] (d) No active memberships at all → returns null

## Out of scope

- Cleaning up the now-dead env var `FIRST_TREE_HUB_CONTEXT_TREE_REPO` from deploys — separate cleanup issue.

liuchao-001:
## Summary

`/api/v1/context-tree/info` and `/api/v1/context-tree/snapshot` both call `resolveUserPrimaryOrgId` to translate a user-scoped JWT into an org for the per-org Team Settings lookup. The function ordered active memberships by `createdAt ASC` and took the first — i.e. the user's earliest-joined active org.

But `/me`'s `defaultOrganizationId` is computed by `pickDefaultMembership` over the same rows with the **opposite** ordering: most-recently-active first, id desc as tie-break.

Multi-org users therefore saw `/me` advertise one default org while the two context-tree endpoints silently resolved a different one — typically an older personal org with no Context Tree configured at all.

## Concrete failure mode

An admin in `agent-team-foundation` who also owns a personal org (joined first) configures Context Tree in Team Settings for `agent-team-foundation`.

```bash
$ curl -H "Authorization: Bearer $TOK" https://dev.cloud.first-tree.ai/api/v1/context-tree/info
{"repo":null,"branch":"main"}    # ← expected agent-team-foundation tree URL
```

`syncContextTree` on Hub Client returns null → `installFirstTreeIntegration` silently no-ops in `runBootstrap` → skills never land in any chat workspace.

## Fix

Delegate the membership pick to the existing `pickDefaultMembership` helper from `services/auth.ts`. Both code paths now read from the same source of truth, so `/me`'s reported default org and the org these server-internal lookups read from can no longer drift apart. Updated docstring records the invariant.

```ts
// before
.orderBy(asc(members.createdAt))
.limit(1);

// after
const rows = await db
  .select({ id: members.id, organizationId: members.organizationId, createdAt: members.createdAt })
  .from(members)
  .where(and(eq(members.userId, userId), eq(members.status, "active")));
return pickDefaultMembership(rows)?.organizationId ?? null;
```

## Test cases (new in `packages/server/src/__tests__/org-settings.test.ts`)

- [x] Single membership → returns that org
- [x] Multi-org with controlled `createdAt` → returns most-recently-created (matches `/me`'s `defaultOrganizationId`)
- [x] `status: "left"` membership with newer `createdAt` → ignored, falls back to active membership
- [x] User with no active memberships → null

All 697 monorepo tests pass; new file count: org-settings.test.ts grew 30 → 34 tests.

## Blast radius

`resolveUserPrimaryOrgId` callers (verified by `git grep`):

- `packages/server/src/api/context-tree-info.ts`
- `packages/server/src/api/context-tree-snapshot.ts`

Both behaviors converge on the org `/me` already advertises — no other surface affected.

## Versioning

Per the updated policy in CLAUDE.md (server-only changes ship via docker / SaaS, not npm), `packages/command` is **not** bumped here. Will piggyback on the next CLI-visible change.

## Closes

- Closes #278

## Test plan

- [x] `pnpm check` — 0 errors
- [x] `pnpm typecheck` — 9/9 successful
- [x] `pnpm test` — 697 / 697 passed
- [ ] After merge + staging deploy: `curl /api/v1/context-tree/info` with a multi-org user account returns the org configured in Team Settings (verifying #278 reproducer is fixed end-to-end).
- [ ] Restart Hub Client daemon connected to staging → start a new chat → confirm `<chat-cwd>/.agents/skills/` and `<chat-cwd>/.claude/skills/` are populated by `installFirstTreeIntegration`.

## Related

- #278 — main bug this fixes
- #279 — companion cleanup: dead `FIRST_TREE_HUB_CONTEXT_TREE_REPO` env var lingering on deploys post PR #256

### DIFF
--- a/packages/server/src/__tests__/org-settings.test.ts
+++ b/packages/server/src/__tests__/org-settings.test.ts
@@ -1,9 +1,10 @@
-import { createHmac } from "node:crypto";
+import crypto, { createHmac } from "node:crypto";
 import bcrypt from "bcrypt";
 import { and, eq } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
 import { members } from "../db/schema/members.js";
 import { organizationSettings } from "../db/schema/organization-settings.js";
+import { organizations } from "../db/schema/organizations.js";
 import { users } from "../db/schema/users.js";
 import { createAgent } from "../services/agent.js";
 import { signTokensForUser } from "../services/auth.js";
@@ -375,6 +376,94 @@ describe("org-settings service", () => {
         { updatedBy: admin.userId, encryptionKey: TEST_ENCRYPTION_KEY },
       ),
     ).rejects.toThrow(/Organization .* not found/);
+  });
+});
+
+describe("resolveUserPrimaryOrgId", () => {
+  const getApp = useTestApp();
+
+  /**
+   * Add an extra membership for an existing user. `createdAt` is exposed so
+   * tests can deterministically control which membership is "most recent".
+   */
+  async function addMembership(
+    app: Awaited<ReturnType<typeof getApp>>,
+    userId: string,
+    role: "admin" | "member",
+    createdAt?: Date,
+    status: "active" | "left" = "active",
+  ): Promise<{ orgId: string; memberId: string }> {
+    const orgId = `org-rup-${crypto.randomUUID().slice(0, 8)}`;
+    const memberId = uuidv7();
+    await app.db.transaction(async (tx) => {
+      await tx.insert(organizations).values({ id: orgId, name: orgId.slice(0, 30), displayName: "Side org" });
+      const human = await createAgent(tx as unknown as typeof app.db, {
+        name: `rup-h-${crypto.randomUUID().slice(0, 6)}`,
+        type: "human",
+        displayName: "RUP Human",
+        managerId: memberId,
+        organizationId: orgId,
+      });
+      await tx.insert(members).values({
+        id: memberId,
+        userId,
+        organizationId: orgId,
+        agentId: human.uuid,
+        role,
+        status,
+        ...(createdAt ? { createdAt } : {}),
+      });
+    });
+    return { orgId, memberId };
+  }
+
+  it("returns the only active membership when user has one org", async () => {
+    const app = getApp();
+    const admin = await createTestAdmin(app);
+    const got = await orgSettingsService.resolveUserPrimaryOrgId(app.db, admin.userId);
+    expect(got).toBe(admin.organizationId);
+  });
+
+  it("returns most-recent active membership for multi-org users (matches /me's defaultOrganizationId)", async () => {
+    const app = getApp();
+    // First org via createTestAdmin. Force its membership createdAt to a known
+    // earlier moment so we can deterministically assert "most recent wins"
+    // regardless of how fast the test runs.
+    const admin = await createTestAdmin(app);
+    const earlier = new Date(Date.now() - 60_000);
+    await app.db.update(members).set({ createdAt: earlier }).where(eq(members.id, admin.memberId));
+
+    // Second org — created "now", which is more recent than `earlier`.
+    const later = await addMembership(app, admin.userId, "admin", new Date());
+
+    const got = await orgSettingsService.resolveUserPrimaryOrgId(app.db, admin.userId);
+    expect(got).toBe(later.orgId);
+    expect(got).not.toBe(admin.organizationId);
+  });
+
+  it("ignores 'left' memberships even when their createdAt is more recent", async () => {
+    const app = getApp();
+    const admin = await createTestAdmin(app);
+    // A more-recent membership the user has since left.
+    await addMembership(app, admin.userId, "admin", new Date(Date.now() + 60_000), "left");
+
+    const got = await orgSettingsService.resolveUserPrimaryOrgId(app.db, admin.userId);
+    expect(got).toBe(admin.organizationId);
+  });
+
+  it("returns null when user has no active memberships", async () => {
+    const app = getApp();
+    const userId = uuidv7();
+    const passwordHash = await bcrypt.hash(crypto.randomUUID(), 4);
+    await app.db.insert(users).values({
+      id: userId,
+      username: `nomembers-${crypto.randomUUID().slice(0, 8)}`,
+      passwordHash,
+      displayName: "No Memberships",
+    });
+
+    const got = await orgSettingsService.resolveUserPrimaryOrgId(app.db, userId);
+    expect(got).toBeNull();
   });
 });
 
--- a/packages/server/src/services/org-settings.ts
+++ b/packages/server/src/services/org-settings.ts
@@ -7,12 +7,13 @@ import {
   type OrgSettingOutput,
   type OrgSettingStorage,
 } from "@agent-team-foundation/first-tree-hub-shared";
-import { and, asc, eq, sql } from "drizzle-orm";
+import { and, eq, sql } from "drizzle-orm";
 import type { Database } from "../db/connection.js";
 import { members } from "../db/schema/members.js";
 import { organizationSettings } from "../db/schema/organization-settings.js";
 import { organizations } from "../db/schema/organizations.js";
 import { BadRequestError, NotFoundError } from "../errors.js";
+import { pickDefaultMembership } from "./auth.js";
 import { decryptValue, encryptValue, isEncryptedValue } from "./crypto.js";
 
 /**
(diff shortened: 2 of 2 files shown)


yuezengwu (approved):
## TL;DR

LGTM, ready to merge. Small change, focused goal, correct fix, regression test included. End-to-end verified locally (HTTP repro on `main` → `{repo: null}`; after fix → expected repo URL).

Non-blocking observations below — feel free to fold into this PR or split into follow-ups.

---

## ✅ Fix correctness

Delegating `resolveUserPrimaryOrgId` to `pickDefaultMembership` is the right call:
- Direction is now correct (`desc(createdAt)`, matching `/me`).
- The `id` desc tie-break is more stable than raw `ORDER BY created_at` — uuidv7's lexicographic order is already time-sorted, so even when `createdAt` collides the result is deterministic.
- Both callers (`context-tree-info.ts:14`, `context-tree-snapshot.ts:30`) already short-circuit on `orgId === null`, so the null path is unchanged.

---

## ⚠️ The deeper root cause isn't fully resolved — \`pickPrimaryMembership\` is a third independent implementation

The PR fixes one inverted implementation, but grep shows the *primary membership* concept is implemented **three** times in the server:

| Location | Order | Tie-break |
|---|---|---|
| \`services/auth.ts:85\` \`pickDefaultMembership\` | \`desc(createdAt)\` | \`id.localeCompare\` desc ✓ |
| \`services/membership.ts:207\` \`pickPrimaryMembership\` | \`desc(createdAt)\` (via \`listActiveMemberships\`) | ❌ none — relies on PG's implementation-defined order |
| \`services/org-settings.ts:273\` \`resolveUserPrimaryOrgId\` | (pre-PR) \`asc(createdAt)\` ❌ |  |

Post-PR we go from 3-way divergence to 2-way. \`pickPrimaryMembership\` is still non-deterministic when \`createdAt\` collides — and same-transaction onboarding inserts share one \`defaultNow()\` timestamp, so that path is exercised in real use.

**Suggestion** (follow-up, not in this PR): route \`pickPrimaryMembership\` through \`pickDefaultMembership\`, or delete it in favor of \`resolveUserPrimaryOrgId\`. Make *primary membership* a single source of truth.

---

## ⚠️ Tests miss the tie-break path

The four new cases cover single / multi / left / null — and I verified the multi-org case really does fail on \`main\`'s buggy implementation and pass after the fix.

One gap: the **\`id\` desc fallback in \`pickDefaultMembership\`** (two memberships with identical \`createdAt\`) isn't covered. That fallback matters precisely because batch inserts inside a single transaction share \`defaultNow()\` — it gets hit in production. Suggested case:

\`\`\`ts
it(\"breaks ties by id desc when two memberships share createdAt\", async () => {
  const sameTime = new Date();
  const admin = await createTestAdmin(app);
  await app.db.update(members).set({ createdAt: sameTime }).where(eq(members.id, admin.memberId));
  const side = await addMembership(app, admin.userId, \"admin\", sameTime);
  const got = await orgSettingsService.resolveUserPrimaryOrgId(app.db, admin.userId);
  // uuidv7 inserted later has the larger id, wins under desc.
  expect(got).toBe(side.orgId);
});
\`\`\`

---

## 💭 Performance note — \`limit(1)\` → fetch-all + in-memory sort

Switching from single-row SQL to *pull all active rows, sort in JS*. Row count per user is bounded by org count (typically < 10), so practical impact is negligible. But the JSDoc doesn't explain *why* \`LIMIT 1\` was dropped — a future reader will want to \"optimize\" it back to \`ORDER BY ... LIMIT 1\` and re-introduce the tie-break divergence.

**Suggestion**: one JSDoc line, e.g.

\`\`\`ts
 * Pulls all active rows (rather than \`ORDER BY ... LIMIT 1\`) so the
 * tie-break logic stays identical to \`pickDefaultMembership\`. Cardinality
 * per user is bounded by the org count — typically <10.
\`\`\`

---

## 🪶 Nit

- \`org-settings.test.ts:1\` now does \`import crypto, { createHmac } from \"node:crypto\"\`, which diverges from the file's surrounding convention — other tests use the global \`crypto.randomUUID()\` (e.g. \`auth.test.ts:48\`). No functional difference; could either keep or align with neighbors.

---

## Verdict

**Approve with non-blocking suggestions.** Solid fix, mergeable as-is. The \`pickPrimaryMembership\` unification and the tie-break test would be valuable follow-ups.

---

### Verification log

Reproduced end-to-end on a local Hub server (real Postgres, real fastify, \`curl\` over HTTP) before and after applying the fix:

| Code version | \`GET /api/v1/context-tree/info\` |
|---|---|
| Buggy \`main\` (\`asc(createdAt)\`) | \`{\"repo\":null,\"branch\":\"main\"}\` ← reads the *earlier* org (no Context Tree configured) |
| This PR | \`{\"repo\":\"https://github.com/example/repro-280-config\",\"branch\":\"main\"}\` ← reads the same org \`/me\` reports as \`defaultOrganizationId\` |

Setup: one user, two active memberships (createdAt 1h apart), Context Tree only configured on the more recent org. Matches the failure mode in the PR description exactly.

PR #285, opened 2026-05-11
fix(web): scope context tab to selected org

Gandy2025:
## Summary

- add an org-scoped Context Tree snapshot endpoint at `/api/v1/orgs/:orgId/context-tree/snapshot`
- switch the Web Context tab to request snapshots for the currently selected team instead of the legacy user-scoped fallback
- surface server-provided Context Tree sync failure detail in the unavailable state
- install `git` in the production Docker runtime so server-managed Context Tree mirrors can clone/fetch repos

## Why

The Web app already tracks the currently selected team through `AuthContext`, but the Context tab still called `/api/v1/context-tree/snapshot`. That legacy user-scoped endpoint can only infer a fallback org, so multi-team users may see context for a different team than the one selected in the user menu.

Separately, the cloud runtime returned `spawn git ENOENT` while trying to sync the configured Context Tree repo because the server Docker image did not include `git`.

## Validation

- `pnpm check`
- `pnpm typecheck`
- `pnpm --filter @first-tree-hub/server exec vitest run src/__tests__/org-settings.test.ts`
- independent code review session reported no blocking findings

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,9 @@ COPY --from=build /app/packages/command/dist packages/command/dist/
 COPY --from=build /app/packages/server/drizzle packages/server/drizzle/
 COPY --from=build /app/packages/web/dist packages/web/dist/
 
-RUN apk add --no-cache wget
+# git is required by the server-managed Context Tree mirror used by
+# /api/v1/context-tree/snapshot. wget is used by the container healthcheck.
+RUN apk add --no-cache git wget
 
 ENV NODE_ENV=production
 EXPOSE 8000
--- a/packages/server/src/__tests__/org-settings.test.ts
+++ b/packages/server/src/__tests__/org-settings.test.ts
@@ -1,4 +1,4 @@
-import crypto, { createHmac } from "node:crypto";
+import crypto, { createHmac, randomUUID } from "node:crypto";
 import bcrypt from "bcrypt";
 import { and, eq } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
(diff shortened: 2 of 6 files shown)


yuezengwu (approved):
LGTM — approving.

**Design is right:**
- Route mounted under Class B `/orgs/:orgId/...` per docs/http-path-conventions.md, with `requireOrgMembership` doing real-time membership checks against the URL `:orgId`.
- `Params: { orgId: string }` gives a compile-time guarantee the route is org-scoped.
- Rate-limit key bumped from `userId` to `userId:orgId` — cleaner isolation across teams under one user.
- Web `queryKey` includes `organizationId`, so switching the selected team gets a fresh fetch instead of cross-team cache reuse; `enabled: !!organizationId` prevents the empty-org race.
- Test directly covers the regression: a non-primary `sideOrgId` returns its own context-tree binding instead of falling back to the caller's primary org.
- Dockerfile change has an explanatory comment for `git`. CI green.

**Non-blocking nits (fine to address later / ignore):**
1. `getContextTreeSnapshot(organizationId, window)` uses `withOrgAt(organizationId, ...)`. Other web org-scoped callers use `withOrg(...)` (module-scope selected org). Equivalent in practice; the explicit form is actually clearer here.
2. The user-scoped `/api/v1/context-tree/snapshot` route is still registered in app.ts:427, and grep shows no web/client/command callers left. Could be removed in a follow-up PR.
3. The `as ContextTreeSyncConfig | undefined` cast mirrors the legacy route — not a new offense.

PR #293, opened 2026-05-11
feat(web): multi-repo onboarding picker + step 2 reframe

Gandy2025:
## Summary

- **Step 3 multi-select source-repo picker.** Single-select dropdown → real multi-select. Selected-repo chip row above the trigger (× to remove without re-opening), proper checkbox visual in popover rows, owner header realigned to repo-name column, density tightened (row padding sp-1_5 → sp-1), hairline group dividers replaced with breathing-room padding. Bootstrap-prose builders take `readonly string[]` and switch singular/plural; N>1 create branch asks the user when source repos span multiple GitHub owners (no more ambiguous "first source").
- **Step 2 reframe — decouple from code.** Lead "a code agent that helps with your code" → "it'll run on a computer you connect". Default name "Coder" → "Assistant". Placeholder "e.g. Code Reviewer" → "e.g. Buddy, Helper". Step 2 sets up an individual agent online; Step 3 is where the team's shared knowledge layer (context-tree) gets built — not "code awareness".
- **Copy / consistency fixes.** `InviteeWaitingBody` "anything that doesn't need code context" → "doesn't depend on your team's shared knowledge"; "Once they bind a source repo and a context-tree" → "Once they finish setup". `InviteeConfirmBody` `useState` initializer gained a comment locking the no-refetch invariant.

Backend / DB / agent config were already array-shaped — the entire surface from `organization_settings.source_repos` to `agents.gitRepos` accepted N repos. The onboarding picker was the lone single-select bottleneck.

Reviewed by two independent passes (first pass on commit 1 + staged step 2; second pass on the full branch). All blockers / actionable nits addressed; remaining open items (listbox keyboard nav, DOM-model drift between native checkbox vs `<button role="option">`, gate-variable naming drift) deferred — captured in [memory](https://github.com/agent-team-foundation/first-tree-hub/blob/main/.gitignore) per repo convention.

A separate follow-up was filed as a local memory entry: explicit **tree destination** picker in Step 3 (the agent currently asks at runtime when source repos span owners; better is a UI-layer pick before the chat starts). Out of scope for this PR.

## Test plan

- [ ] Admin onboarding step 3: multi-select picker opens, picking N repos shows N chips above the trigger, popover stays open across picks, owner header aligns to repo-name column
- [ ] Chip × removes the chip AND unchecks the popover row (state sync both directions)
- [ ] Continue submits all selected repos to `gitRepos` on the agent + `source_repos` namespace on the team
- [ ] Bootstrap message to agent: N=1 reads singular prose, N>1 reads plural with markdown bullet list of URLs
- [ ] N>1 cross-owner case: bootstrap tells agent to ask the user which owner the tree should live under
- [ ] Step 2 lead: "Set up your first agent — it'll run on a computer you connect."
- [ ] Step 2 default agent name: "Assistant"; placeholder: "e.g. Buddy, Helper"
- [ ] Step 3 lead unchanged: "Build the context-tree — your team's shared knowledge that grows with your code."
- [ ] InviteeWaitingBody (invitee whose team has neither tree nor repos): copy says "Once they finish setup" and "anything that doesn't depend on your team's shared knowledge"
- [ ] InviteeConfirmBody (invitee whose team has both): all team repos pre-checked by default; uncheck still allows Confirm if at least one remains

### DIFF
--- a/packages/web/src/pages/workspace/center/onboarding/bootstrap-prose.ts
+++ b/packages/web/src/pages/workspace/center/onboarding/bootstrap-prose.ts
@@ -19,18 +19,15 @@
 
 export const FIRST_TREE_REFERENCE_URL = "https://github.com/agent-team-foundation/first-tree";
 
-function formatSourceList(sourceUrls: readonly string[]): { label: string; lines: string[] } {
+function formatSourceList(sourceUrls: readonly string[]): string[] {
   if (sourceUrls.length === 1) {
-    return { label: "Source repo", lines: [`Source repo: ${sourceUrls[0]}`] };
+    return [`Source repo: ${sourceUrls[0]}`];
   }
-  return {
-    label: "Source repos",
-    lines: ["Source repos:", ...sourceUrls.map((u) => `- ${u}`)],
-  };
+  return ["Source repos:", ...sourceUrls.map((u) => `- ${u}`)];
 }
 
 export function buildBindBootstrap(sourceUrls: readonly string[], treeUrl: string): string {
-  const sources = formatSourceList(sourceUrls);
+  const sourceLines = formatSourceList(sourceUrls);
   const opener =
     sourceUrls.length === 1
       ? "Bind my source repo to an existing context-tree."
@@ -42,7 +39,7 @@ export function buildBindBootstrap(sourceUrls: readonly string[], treeUrl: strin
   return [
     opener,
     "",
-    ...sources.lines,
+    ...sourceLines,
     `Existing tree: ${treeUrl}`,
     "",
     skillLine,
@@ -52,7 +49,7 @@ export function buildBindBootstrap(sourceUrls: readonly string[], treeUrl: strin
 }
 
 export function buildCreateBootstrap(sourceUrls: readonly string[]): string {
-  const sources = formatSourceList(sourceUrls);
+  const sourceLines = formatSourceList(sourceUrls);
   const opener =
     sourceUrls.length === 1
       ? "Create a brand-new context-tree for my source repo."
@@ -64,7 +61,7 @@ export function buildCreateBootstrap(sourceUrls: readonly string[]): string {
   return [
     opener,
     "",
-    ...sources.lines,
+    ...sourceLines,
     "",
     skillLine,
     "",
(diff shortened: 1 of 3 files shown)


baixiaohang (approved):
## What this does

Step 3 source-repo picker goes single → multi-select, with a parallel reframe of Step 2 away from "code agent" framing. Three call sites (`AdminBindCreateBody` / `InviteeConfirmBody` / `InviteePickerBody`) flip from `selectedRepoUrl: string | null` to `selectedRepoUrls: string[]`; `buildBindBootstrap` / `buildCreateBootstrap` take `readonly string[]` and switch singular/plural prose, with N>1 + cross-owner deferred to the agent at runtime. Backend / DB / agent config schemas were already array-shaped (`agent-runtime-config.ts:89`, `org-settings.ts:165`), so no cross-layer compat issue.

## Review

**No blockers.** Clean refactor, call sites all updated, schemas already array-shaped, test plan thorough. A few non-blocking notes:

1. **`bootstrap-prose.ts:22-29` — `label` field in `formatSourceList` return is dead code.** Both callers read only `lines`. Simplify to `string[]`.

2. **`step3-intro-body.tsx:273-275` — checkbox visual inconsistency between `InviteeConfirmBody` and the popover.** The popover rows use a proper boxed checkbox (`step3-intro-body.tsx:1138-1157`), but `InviteeConfirmBody`'s label still renders a bare `<Check>` glyph with no box. Same mental model (multi-select), two visuals. Suggest reusing the boxed visual for parity.

3. **`step3-intro-body.tsx:267` — leftover `name="invitee-source-repo"` on checkboxes.** `name` had grouping semantics for the radio version; on checkboxes it's just noise. Drop it.

4. **`step3-intro-body.tsx:1025` — `triggerLabel` is awkward at the edges.** Always "Add another repository" once anything is selected, even when every visible repo is already picked. Suggest "All N selected" when `selectedRepos.length === repos.length`, or "N selected · pick more" otherwise.

5. **Cross-owner decision deferred to runtime is a follow-up worth filing as a real issue.** `bootstrap-prose.ts:63` leans on the agent to detect cross-owner and ask. The frontend already has `selectedRepos`; computing `new Set(repos.map(parseOwner)).size > 1` and surfacing a UI-level owner picker is cleaner than a runtime back-and-forth. PR notes this is captured as a memory entry — consider promoting to an issue so it doesn't get lost.

6. **A11y / keyboard nav (already noted as deferred).** `role="listbox"` + `aria-multiselectable="true"` + per-row `aria-selected` are correct, but ↑↓ navigation and Space-to-toggle aren't wired. Also, `<button role="option">` mixes the DOM model (button) with the ARIA model (option) — some screen readers handle this inconsistently. Worth a tracked issue.

7. **`step2-body.tsx:430` — placeholder `e.g. Buddy, Helper` softens the role hint.** The old "Code Reviewer" gave new users an *action* shape for naming; the new examples are pure vibe. Taste call — leaving as-is is fine, but `e.g. Buddy, Research Pal, Project Helper` keeps the new tone while preserving a functional cue.

Items 1–3 are 5-min cleanups worth doing before land; 4–7 are follow-ups, don't block.

LGTM — approving.

PR #311, opened 2026-05-12
fix(cli,server): respect FIRST_TREE_HUB_AGENT_ID and route agent-to-agent sends to current chat

### LINKED ISSUE #192: agent发送消息，需要多轮tool call
agent发送消息，没法一次性成功，需要多轮tool call：
↳
Bash({"command":"cat <<'EOF' | first-tree-hub agent send 019d9a48…)
error
953ms
↳
Bash({"command":"cat <<'EOF' | first-tree-hub agent send --agent …)
error
1.6s
↳
Bash({"command":"env | grep -i FIRST_TREE 2>&1","description":"Ch…)
ok
54ms
↳
Bash({"command":"echo \"ping\" | first-tree-hub agent send --agen…)

baixiaohang:
## Summary

Two long-observed agent-to-agent messaging papercuts, fixed in one pass.

- **CLI: `agent send` no longer demands `--agent` on a multi-agent client.** When the runtime injects `FIRST_TREE_HUB_AGENT_ID`, the CLI now reverse-looks it back to a local agent name. Closes #192.
- **Server: agent-to-agent sends route to the caller's current chat when the recipient is a member.** `sendToAgent` previously hard-routed to the pair's direct chat via `findOrCreateDirectChat`, ignoring session context — so a group chat where A says \"go ask B\" would silently land in a parallel A↔B direct chat. The CLI already auto-injects `FIRST_TREE_HUB_CHAT_ID` into `replyToChat`; the server now uses that as a routing hint, falling back to the direct-chat path when the recipient isn't a member of the current chat.
- **CLI errors now include paste-ready command templates** for the still-ambiguous edge cases (`AMBIGUOUS_AGENT`, new `ENV_AGENT_NOT_LOCAL`), clarifying that `--agent` selects the SENDER while the recipient is positional. (#192's screenshot showed an LLM mis-positioning `--agent` in its recovery attempt.)
- **`bootstrap.ts` (system prompt injected into every agent):**
  - `[From: sender-id]` → `[From: <agent-name>]` — matches what `agent-io.ts` actually renders, and the token agents can pass back to `agent send`.
  - `FIRST_TREE_HUB_AGENT_ID` description now warns it's the caller's own uuid; never use as a `send` target.
  - `FIRST_TREE_HUB_CHAT_ID` description notes the CLI uses it for routing — agents don't pass it manually.
  - `agent send` example block calls out the new auto-routing rule.

Bumps `@agent-team-foundation/first-tree-hub` to `0.12.5` per `docs/versioning-and-publishing.md` (touches `command` + `client/runtime`).

## Test plan

- [x] `pnpm check` (biome) — 625 files, clean
- [x] `packages/command` vitest — **104/104** (incl. 6 new `resolveSenderName` cases)
- [x] `packages/server` vitest — **751/751** (incl. 3 new current-chat-routing integration tests, 1 prior test adjusted to keep exercising the direct-chat fallback under the new routing rule)
- [x] `packages/client/src/__tests__/bootstrap.test.ts` — 16/16 (assertion updated to match new `[From: <agent-name>]` wording)
- [ ] Manual: install the new tarball and reproduce issue #192's setup (multi-agent client, two agents in one group) — `agent send <name>` should now succeed in one tool call and the message should appear in the group, not a new direct chat.

> Note: `packages/client/src/__tests__/git-mirror-manager.test.ts > does NOT classify a non-credential https failure as auth` fails on `main` too — pre-existing flake unrelated to this PR.

### DIFF
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -80,7 +80,7 @@ describe("bootstrapWorkspace", () => {
     expect(content).toContain("FIRST_TREE_HUB_ACCESS_TOKEN");
     expect(content).toContain("How You Communicate");
     expect(content).toContain("Agent Hub");
-    expect(content).toContain("[From: sender-id]");
+    expect(content).toContain("[From: <agent-name>]");
     expect(content).toContain("Use your judgment about when to respond");
   });
 
--- a/packages/client/src/runtime/bootstrap.ts
+++ b/packages/client/src/runtime/bootstrap.ts
@@ -336,7 +336,8 @@ function generateToolsDoc(): string {
 You are running inside **Agent Hub**, a messaging platform for agent teams.
 
 - Messages from other team members arrive as your prompt input
-- Each message includes a \`[From: sender-id]\` header so you know who sent it
+- Each message includes a \`[From: <agent-name>]\` header — that name is also
+  what you pass back to \`agent send\` to reply to or address that agent
 - **Your final text response is automatically delivered** to the chat — just respond normally
 - For **proactive communication** (sending to other agents, other chats, or structured data),
   use the \`first-tree-hub\` CLI below
@@ -351,8 +352,8 @@ These are injected automatically when the agent process starts:
 |----------|-------------|
 | \`FIRST_TREE_HUB_SERVER_URL\` | Server address for API calls |
 | \`FIRST_TREE_HUB_ACCESS_TOKEN\` | User member access JWT (short-lived) |
-| \`FIRST_TREE_HUB_AGENT_ID\` | Your agent UUID — send as \`X-Agent-Id\` |
-| \`FIRST_TREE_HUB_CHAT_ID\` | Current chat context ID |
+| \`FIRST_TREE_HUB_AGENT_ID\` | YOUR own agent UUID. The CLI reads it to identify you as the sender — never pass it as a \`send\` target. |
+| \`FIRST_TREE_HUB_CHAT_ID\` | The chat this session is currently bound to. The CLI uses it to route messages — you don't need to pass it manually. |
 
 The \`first-tree-hub\` CLI reads these automatically — no extra setup needed.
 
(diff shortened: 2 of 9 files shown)


yuezengwu (changes requested):
## 🔴 Authorization regression: sender membership is not checked on `replyToChat`

In `sendToAgent` (`packages/server/src/services/message.ts`), the new current-chat routing branch only validates that **target** is a participant of `replyToChat`:

```ts
if (data.replyToChat && (await isParticipant(db, data.replyToChat, target.uuid))) {
  return sendMessage(db, data.replyToChat, senderUuid, { ... });
}
```

`sendMessage` itself never asserts sender membership either — it just fetches participants by chatId and filters senderId out of the fan-out list. Combined with `replyToChat` being a client-controlled field (CLI `--reply-to-chat` flag, not only env injection), this lets any agent A in the same org post into any chat where B is a member, even when A was never added.

**Concrete misuse**: A learns the chatId of B↔C's private direct chat (chatIds surface in `agent chats`, inbox entries, etc.), then runs:

```bash
first-tree-hub agent send B --reply-to-chat <BC-direct-chat-id> "..."
```

The message lands in BC's private chat. Pre-PR `sendToAgent` was safe by construction because `findOrCreateDirectChat(sender, target)` always made sender a participant.

### Suggested fix

Require sender membership too:

```ts
if (
  data.replyToChat &&
  (await isParticipant(db, data.replyToChat, senderUuid)) &&
  (await isParticipant(db, data.replyToChat, target.uuid))
) { ... }
```

(or fold both checks into a single `IN (...)` query for one round-trip).

### Test gap

The 3 new routing tests cover *(sender+target both in)*, *(sender in, target not in)*, *(no hint)* — but not *(sender NOT in, target in)*, which is exactly the attack path. Please add a case: sender passes a chatId they aren't a member of → must fall back to direct chat (or throw `ForbiddenError`, depending on the chosen semantics).

The rest of the PR looks good — `resolveSenderName` extraction + tests, the paste-ready CLI error templates, and the bootstrap CLAUDE.md clarifications all land cleanly. Just this one server-side check before merge.

yuezengwu (approved):
Fix matches the review exactly — both `isParticipant` checks in parallel and the spoofing-guard test that pins the previously-uncovered branch. Comment now documents the invariant clearly. LGTM.

PR #300, opened 2026-05-11
feat: GitHub App migration — Phase 1 (full cutover)

Gandy2025:
## Summary

**GitHub App migration — Phase 1 (full cutover).** Replaces the legacy OAuth-App + per-org webhook surface with a GitHub App: one install URL per Hub, App-issued installation tokens, App-level webhook with HMAC, OAuth user-token refresh. ([design doc on PR #295](https://github.com/agent-team-foundation/first-tree-hub/pull/295))

Single PR because none of the intermediate stages is independently useful — every internal stage produces dead code until the cutover lands; D3 = no compatibility window.

## What lands

The work shipped in three rounds, all squashed by merge:

### Phase 1 — original cutover (commits `5c5ff03`..`2180264`)

10 feature commits building the App surface end-to-end:
- `github_app_installations` table + `auth_identities` App-token shape
- `services/github-app.ts` — App JWT, installation token, user-token refresh
- Env config (`FIRST_TREE_HUB_GITHUB_APP_*`) + boot guards
- App user-OAuth helpers + `/auth/github/start` + `/callback` rewrite
- Install state machine + OAuth-callback upsert/bind
- `/api/v1/webhooks/github` App endpoint + HMAC verify + dispatch
- `dev-installation` bypass for `/dev-callback`
- Admin API `GET /orgs/:orgId/github-app-installation` + Settings panel
- D3 hard cut — drop legacy OAuth-App + per-org webhook + namespace

### Phase A + B — codex review fixes (commits `366a7a3`..`d49c466`)

5 P0/P1 fixes from a first codex review pass:
- **P0-2** authorize `installation_id` in OAuth callback (`366a7a3`)
- **P0-3 + H2** race-safe `bindInstallationToOrg` + `ConflictError` (`b32c89d`)
- **P1-8** boot guards in `buildApp` + `.min(1)` on App secrets (`a4b2d87`)
- **P1-9** explicit opt-in for `/dev-callback` bypass (`5b1ca20`)
- **P1-2 + P1-4** `ApiError.status` 404 detection + token-refresh wiring (`d49c466`)

### Phase C — second codex round + the rest (commits `7037bfe`..`a2ad082`)

5 P1 fixes from the codex challenge pass + 1 follow-up after a second challenge:
- **C.8 P1-1** install URL via App slug + signed state JWT (`7037bfe`)
- **C.9 P1-3** bind installs to target org via state JWT, admin-of-target check (`71e6dc2`)
[...]

Issue #319, opened 2026-05-12
GitHub App: /user/installations pagination capped at 500 — power users can fail callback bind


Gandy2025:
**Severity: P2**
**Source: codex challenge against PR #300 rollup, 2026-05-12**
**Status: pre-existing — same primitive surfaced in #312 (auth weakness). Pagination cap is a separate concern: the user might have admin access but be unreachable via the limited scan.**

## Finding

[`listUserAccessibleInstallationIds`](packages/server/src/services/github-app.ts) (line ~184) defaults to `perPage=100` and `maxPages=5`, so it tops out at 500 installations. It also doesn't read GitHub's `total_count` field or follow Link pagination beyond `maxPages`.

A legitimate GitHub org admin who has access to >500 App installations across orgs they administer will have their callback `installation_id` validation fail (the install isn't in the truncated list) — same for the manual claim endpoint.

500 sounds high but:
- A power user across many GitHub orgs (consultants, security researchers, large enterprises with hundreds of teams)
- A user who has been added as a collaborator to many installs over time

…can realistically hit it.

## Fix

Iterate until `total_count` is exhausted (or until an explicit safety cap like 5000 hits). Either:
- Bump `maxPages` to 50 (covers 5000 installs)
- Switch to streaming / Link-header pagination with a hard upper bound

Should be paired with #312 since both touch this primitive — a unified "rebuild the install-admin check" PR can address both at once.

## Affected

- [`packages/server/src/services/github-app.ts:184`](packages/server/src/services/github-app.ts) — `listUserAccessibleInstallationIds`
- Indirectly: OAuth callback `installation_id` check, manual claim endpoint

Gandy2025:
Related: #312 (`/user/installations` doesn't prove admin access — security). Same function (`listUserAccessibleInstallationIds`), so a unified rewrite can address both. If #312's fix switches to `/user/memberships/orgs/{login}` for org installs, this pagination cap may stop mattering for the security-critical path; still relevant for any caller that legitimately needs the full installation list.

yuezengwu:
Closing as obsolete after the installation-authorization model changed.

[...]

PR #322, opened 2026-05-12
feat(server): GitHub App foundation — schema, services, config (PR 1/3)

Gandy2025:
## Summary

**First of 3 PRs splitting the original PR #300** (GitHub App migration). This PR is the foundation — all schema, service primitives, and config infrastructure that PR-2 (OAuth + Settings UI) and PR-3 (webhook + D3 hard cut) build on. **NO user-facing behavior changes.** New tables / services are added but no route consumes them yet.

See [docs/pr-300-as-built-design.md](docs/pr-300-as-built-design.md) for the full architectural review of the combined work, and [docs/pr-300-split-plan.md](docs/pr-300-split-plan.md) for the rationale of this 3-way split (was originally proposed as 4-way; merged Settings UI into PR-2 because of OAuth dependency).

### Schema
- `packages/server/drizzle/0037_github_app_installations.sql` — new table for one-row-per-(GitHub-account ↔ Hub-team) binding
- `packages/server/src/db/schema/github-app-installations.ts` — Drizzle types
- Exported from `db/schema/index.ts` so drizzle-kit + `db.query.*` see it

### Service primitives
- `services/github-app.ts` — App JWT (RS256), installation token, user token refresh (8h TTL)
- `services/github-app-installations.ts` — state machine: upsert / find / bind / suspend / unsuspend / delete + orphan-reclaim helper
- `services/oauth-state.ts` — extended with `targetOrganizationId` field in the signed state JWT (codex P1-3)

### Config
- `oauth.githubApp` block in `shared/config/server-config.ts` — 6 fields: appId, clientId, clientSecret, privateKeyPem, webhookSecret, slug. All required except `slug` (only the install URL CTA needs it).
- Legacy `oauth.github` block intact — PR-2 removes as part of D3 hard cut.

### Boot guard
- `server/boot-guards.ts:assertBootConfigValid` — validates PEM header (PKCS#8), rejects half-config + blank webhookSecret (codex P1-8). **Wired into `buildApp` so it actually runs at startup.**

### DTOs
- `shared/schemas/github-app.ts` — Zod schemas for installation output, claim body, account types, permission levels.

## Codex hardening already included

The original PR-300 picked up these codex fixes; they're carried forward:

- **P0-3 + H2** race-safe `bindInstallationToOrg` + `ConflictError` on collisions
- **P1-8** boot guards in `buildApp` + `.min(1)` on every App secret field
- **P1-7** suspend/unsuspend timestamp guards
- **P1-3** `targetOrganizationId` in signed state JWT
- **P1-5 + H1** `findUnboundInstallationsByAccount` helper for orphan reclaim
- Codex challenge follow-up: deleted the 60-s grace window in `deleteInstallationByGithubId`

## Codex post-rollup review pass

Ran `codex review --base origin/main` against this branch. Result: **GATE: PASS** — 2 [P2]'s found and **fixed**:
1. Boot guard not invoked anywhere — fixed by calling `assertBootConfigValid(config)` from `buildApp`
2. New table missing from `db/schema/index.ts` barrel — fixed

No [P0] / [P1] introduced.

## Verification

- [x] `pnpm check` (Biome) — clean (647 files)
- [x] `pnpm typecheck` — 9 packages clean
- [x] `pnpm --filter @first-tree-hub/server test` — **794/794 pass** (testcontainers PG)
- [x] Migration `0037` smoke-tested via test suite's template DB

## What's NOT here (in PR-2 / PR-3)

- `/auth/github/start` + `/callback` rewrite → PR-2
- Admin API endpoints (`/orgs/:orgId/github-app-installation/*`) → PR-2
- Web Settings panel → PR-2
- D3 cut: legacy OAuth client + per-org webhook + `github_integration` namespace deletion → PR-2 + PR-3
- `/api/v1/webhooks/github` App endpoint → PR-3
- Migration `0038_drop_github_integration_namespace.sql` → PR-3

## Risk

🟢 **Low**. New tables + services with no callers. Boot guard runs at startup but is a no-op when the App env block is unset. Legacy OAuth client + per-org webhook still work unchanged.

## Stacked PR sequence

This PR → **PR-2** (OAuth + Settings UI) → **PR-3** (Webhook + D3 cut)

Closes part of #300.

### DIFF
--- a/packages/server/drizzle/0037_github_app_installations.sql
+++ b/packages/server/drizzle/0037_github_app_installations.sql
@@ -1,0 +1,52 @@
+-- GitHub App installation registry. See packages/server/src/db/schema/github-app-installations.ts
+-- for the Drizzle types, and docs/github-app-design-zh.md for the design rationale.
+--
+-- One row per (GitHub account ↔ Hub team) binding. Replaces the per-repo
+-- OAuth + webhook-secret model that lived in
+-- `organization_settings.github_integration.webhookSecretCipher` — both
+-- coexist during the transition, the old path is dropped in a later PR
+-- (D3 hard cut, design doc §7 step 7).
+--
+-- Per the team's "integrity in service layer" convention, NO foreign-key
+-- constraints on hub_organization_id beyond the optional reference — the
+-- 1:1 binding (D2 / §8 Q1) is enforced by a UNIQUE INDEX rather than by
+-- ON DELETE CASCADE so deleting a Hub org doesn't tombstone the
+-- GitHub-side record (which still exists upstream).
+
+CREATE TABLE IF NOT EXISTS "github_app_installations" (
+  "id" text PRIMARY KEY NOT NULL,
+  "installation_id" bigint NOT NULL,
+  "account_type" text NOT NULL,
+  "account_login" text NOT NULL,
+  "account_github_id" bigint NOT NULL,
+  "hub_organization_id" text,
+  "permissions" jsonb NOT NULL,
+  "events" jsonb NOT NULL,
+  "suspended_at" timestamp with time zone,
+  "created_at" timestamp with time zone NOT NULL DEFAULT now(),
+  "updated_at" timestamp with time zone NOT NULL DEFAULT now(),
+  CONSTRAINT "ck_github_app_installations_account_type"
+    CHECK ("account_type" IN ('User', 'Organization'))
+);
+
+--> statement-breakpoint
+DO $$ BEGIN
+  ALTER TABLE "github_app_installations"
+    ADD CONSTRAINT "github_app_installations_hub_organization_id_organizations_id_fk"
+    FOREIGN KEY ("hub_organization_id") REFERENCES "organizations"("id")
+    ON DELETE SET NULL ON UPDATE NO ACTION;
+EXCEPTION
+  WHEN duplicate_object THEN null;
+END $$;
+
+--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "uq_github_app_installations_installation_id"
+  ON "github_app_installations" ("installation_id");
+
+--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "uq_github_app_installations_hub_org"
+  ON "github_app_installations" ("hub_organization_id");
+
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "idx_github_app_installations_account"
+  ON "github_app_installations" ("account_github_id");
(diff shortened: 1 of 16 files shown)


baixiaohang (approved):
LGTM — solid foundation PR. Quality is high: design is rigorous (1:1 binding enforced via UNIQUE rather than CASCADE; `ON DELETE SET NULL` keeps the GitHub-side record alive when the Hub org is deleted), the security primitives are sharp (`listUserAccessibleInstallationIds` for P0-2 authz, HMAC empty-key forgery defense, no `permissions` in the authorize URL to block downgrade prompts), and concurrency safety in `bindInstallationToOrg` (conditional UPDATE + 23505 catch) is exactly the right shape. 794/794 tests, including explicit race-condition and stale-event coverage.

A few non-blocking nits — happy to approve as-is, but worth a quick cleanup pass if you're inclined:

### 🟡 Worth fixing
1. **Dead links in the PR description.** `docs/pr-300-as-built-design.md` and `docs/pr-300-split-plan.md` are linked but don't exist in this branch (HTTP 404). The code also references `docs/github-app-design-zh.md` for the D0a/D0b/D1/D2/D3/D4 decisions but it's not in this PR either. Either include the docs here or point at the merged PR/file where they live.

2. **`bindInstallationToOrg` jsdoc lies.** Doc says "Returns true on first bind, false on idempotent re-bind", but the implementation explicitly notes "we can't tell them apart without a third query…the contract simplifies to 'true on any successful UPDATE'". Either update the jsdoc to match reality, or add the third query.

3. **`shared/schemas/github-app.ts` — two jsdoc blocks transposed.** The "GET-side projection" doc block sits above `githubAppInstallationClaimBodySchema` rather than `githubAppInstallationOutputSchema` (looks like a merge artifact around L2441–2457 in the diff).

4. **`AppInstallation.permissions` type drift risk.** `services/github-app.ts` redeclares `Record<string, "read" | "write" | "admin">` instead of importing `GithubAppInstallationPermissions` from `@agent-team-foundation/first-tree-hub-shared`. Runtime-identical today, but easy to diverge later — worth importing the shared type.

### 🟢 Nice-to-have (probably defer)
5. **`boot-guards.ts` has no unit tests.** PEM-header check / half-config / production publicUrl are all fail-closed paths; a few unit tests around the throw cases would harden this. The `helpers.ts` stub exercises the happy path indirectly, but the failure branches are untested.

6. **`listUserAccessibleInstallationIds` silently truncates at `maxPages * perPage = 500`.** When the last page is full, the loop exits and the caller's `Set` looks complete but may be missing later pages. Already tracked as #319, but since this is a P0-2 authz primitive, silent undercount → legitimate installs judged as "not accessible". At minimum, a `log.warn` when bailing on `maxPages` with a full final page would make this debuggable.

7. **`boot-guards.ts:assertGithubAppConfigComplete` all-empty branch is effectively unreachable.** Zod's `.min(1)` already rejects empty strings at parse time. Either drop the branch or annotate it as belt-and-braces.

None of these block the merge. The split rationale is sound — none of the new code is callable until PR-2/PR-3 land, so risk is genuinely low. Approving.

Issue #318, opened 2026-05-12
GitHub App: Settings UI lacks 'Claim install' buttons for multi-orphan recovery


Gandy2025:
**Severity: P1 (UX dead-end for a real recovery path)**
**Source: codex challenge against PR #300 rollup, 2026-05-12**
**Status: known gap from Phase C C.10 — backend endpoint shipped (POST `/api/v1/orgs/:orgId/github-app-installation/claim`) but no web UI calls it.**

## Finding

Phase C C.10 (commit `753d74b`) added orphan-install recovery:

- **Auto-reclaim on sign-in**: if exactly 1 unbound install matches the user's GitHub account, it's auto-bound. Works fine.
- **Manual claim endpoint**: when 2+ orphans exist (or for org-account installs where \"user is admin\" isn't sufficient basis to auto-bind), the OAuth callback log says \"Settings UI can retry\" — but the Settings panel doesn't actually surface any \"Claim install\" buttons.

Affected files:
- Server-side note: [`packages/server/src/api/auth/github.ts:470`](packages/server/src/api/auth/github.ts) — \"Settings UI can retry\" log line
- Server-side endpoint: [`packages/server/src/api/orgs/github-app.ts:146`](packages/server/src/api/orgs/github-app.ts) — POST `/orgs/:orgId/github-app-installation/claim`
- Web client surface (missing): [`packages/web/src/api/github-app.ts:18`](packages/web/src/api/github-app.ts) — only exposes `getGithubAppInstallation()` and `getGithubAppInstallUrl()`
- Web panel: [`packages/web/src/pages/github-app-installation-panel.tsx`](packages/web/src/pages/github-app-installation-panel.tsx) — only renders install/manage states, no orphan picker

## What's needed

1. **New backend endpoint**: GET `/api/v1/orgs/:orgId/github-app-installation/orphans` returning unbound installs whose `accountGithubId` matches the caller's GitHub identity (mirrors the `findUnboundInstallationsByAccount` helper)
2. **Web client method**: `listGithubAppOrphans(orgId)` in `packages/web/src/api/github-app.ts`
3. **Panel UI**: in `github-app-installation-panel.tsx`, when no install is bound and `getGithubAppInstallation()` returns null, fetch orphans. If \>0, render a list with \"Claim into this org\" buttons calling the existing claim endpoint

Estimated work: ~150 LOC server + ~100 LOC web. Half a day.

## Why deferred from PR #300

[...]

PR #323, opened 2026-05-12
feat(server,web): switch sign-in to GitHub App + Settings UI (PR 2/3)

Gandy2025:
## Summary

**Second of 3 PRs splitting the original PR #300** (GitHub App migration). Stacks on PR-1 [#322](https://github.com/agent-team-foundation/first-tree-hub/pull/322) (foundation). Wires the foundation service primitives into actual user-facing flows. After this lands, sign-in goes through the GitHub App user-authorization flow (not the legacy OAuth App), and admins can install + view App installations from the Settings page. (Manual orphan-claim is API-only in this PR — UI tracked in #318.)

### OAuth flow rewrite
- `/auth/github/start` + `/callback` rewritten to use the GitHub App authorize URL (D1: install + sign-in in one redirect)
- Returning users: `code + state`. First-time installers: `code + state + installation_id`.
- The callback exchanges the code via App user-OAuth, optionally upserts + binds the installation, runs the orphan-reclaim sweep (codex P1-5 + H1) for unbound installs matching the signing-in user
- `/auth/github/dev-callback` — explicit opt-in via `FIRST_TREE_HUB_DEV_CALLBACK_ENABLED=1` (codex P1-9)

### Admin API endpoints (`/api/v1/orgs/:orgId/github-app-installation/*`)
- `GET` — install state (current binding metadata)
- `GET /install-url` (codex P1-1) — signs target-org-aware state JWT, sets cookie, returns App's `installations/new` URL with state
- `POST /claim` (codex C.10) — manual claim path for orphan installs when the orphan-reclaim sweep finds 2+ matches. **API-only in this PR**; the Settings "Claim install" picker UI is tracked in #318

### Web UI
- New `web/src/api/github-app.ts` client + `pages/github-app-installation-panel.tsx` — renders Bound / NotInstalled states and the "Install on GitHub" CTA
- Old `github-integration-panel.tsx` deleted (the Settings UI for the legacy per-org webhook secret)
- Multi-orphan picker UI deferred to #318 — `POST /claim` endpoint ships in this PR but has no UI driver yet

### D3 cut auth-side
- Removed `oauth.github` block from `server-config`
- Removed half-config check in `src/index.ts`
- Removed legacy `exchangeCodeForProfile` from `services/github-oauth.ts` (kept the thin GitHub helpers still needed)

### Codex hardening included
- **P1-3** target-org via state JWT
- **P1-5 + H1** orphan reclaim + manual claim
- **P1-9** /dev-callback explicit opt-in
- **P1-2 + P1-4** ApiError 404 detection + token refresh wiring

## Codex post-rollup review

Ran `codex review --base feat/github-app/01-foundation` against this branch. **Gate: PASS** — 0 P0, 0 P1, 3 P2 findings. All 3 P2s were small, clearly correct, and fixed in this PR before opening:

- **P2-1** — Settings panel preloaded the install URL on render via `useQuery`, but the signed `state` JWT and nonce cookie expire after 10 minutes. An admin who left the page open longer would land on a stale state and the post-install callback would 401. Switched to `useMutation` fired in the click handler so the URL (and cookie) are minted at the moment the user actually navigates.
- **P2-2** — `githubCallbackQuerySchema` rejected `setup_action=update`, which GitHub sends when an existing installation is reconfigured or revisited. Extended the enum to accept `install`, `update`, and `request`; the callback continues to branch only on `installation_id` (the real signal).
- **P2-3** — When `fetchInstallation` or the upsert failed, the callback fell through with `installationId` still set, so the later `bindInstallationToOrg` ran against a missing row and silently logged. The webhook couldn't reconcile either (no Hub-org context). Now we clear `installationId` on the catch path so bind is skipped entirely; sign-in still succeeds, and Settings surfaces a clean "Install" CTA the user can re-trigger. Test (`github-app-callback-target-org.test.ts`) updated to use a real RSA-2048 keypair so the upsert path actually exercises end-to-end (added `githubAppPrivateKeyPem` override to `CreateTestAppOptions`).

## Codex post-review fixups (commit `5551634`)

Addressed @baixiaohang's review:
- Clarified `/claim` is API-only and pointed at #318 (`api/orgs/github-app.ts`, `api/auth/github.ts` log lines)
- Fixed stale comment in `pages/settings/github.tsx` that implied the per-org webhook + `github_integration` namespace were already deleted (those land in PR-3)
- Dropped the design-doc references that don't resolve on this branch (`docs/github-app-design-zh.md` lives on the unmerged #295)
- Tightened the `oauthTokens` docstring at `api/auth/github.ts:344` (`always includes` → `typically includes`, with explicit dev-callback empty-bundle note)
- Added a per-worker env-safety note in `__tests__/setup.ts` for the `FIRST_TREE_HUB_DEV_CALLBACK_ENABLED` mutation pattern in `oauth-flow.test.ts`

## Verification

- [x] `pnpm check` (Biome) — clean (652 files)
- [x] `pnpm typecheck` — 9 packages clean
- [x] `pnpm --filter @first-tree-hub/server test` — **814/814 pass** (+20 tests vs PR-1)

## What's NOT here (PR-3)

- `/api/v1/webhooks/github` App endpoint
- D3 cut webhook-side: legacy `webhooks/github.ts` per-org route, webhook secret cipher, `github_integration` namespace deletion (migration 0038)
- Mention-only routing alignment with #304

## Risk

🟡 **Medium**. Sign-in path changes for all users. Returning users with no install: previously logged in via OAuth App, now login via GitHub App with no install (callback handles that case). Old per-org webhook still works (PR-3 deletes it).

Closes part of #300.


baixiaohang:
## Review

Overall: high-quality PR. All codex P0/P1/P2 follow-ups look correctly applied, and test coverage is solid (24 new cases). A few things worth resolving before merge.

### 1. UX gap — server-side `/claim` has no UI to drive it

The server ships `POST /api/v1/orgs/:orgId/github-app-installation/claim` (`packages/server/src/api/orgs/github-app.ts:146-176`) and the OAuth callback at `packages/server/src/api/auth/github.ts:493-498` logs `"multiple unbound installs match this account — skipping auto-claim, Settings UI surfaces a picker"`. The doc-comment on the claim route says explicitly `"The Settings panel surfaces a 'Claim install' button per orphan; this is what it POSTs."`

But:
- `packages/web/src/pages/github-app-installation-panel.tsx` only renders `Bound` / `NotInstalled` states — no orphan picker, no claim button.
- `packages/web/src/api/github-app.ts` has no `claimGithubAppInstallation` helper.
- There is no `GET` endpoint that lists the orphan candidates for a UI to enumerate.

Net effect: the recovery path is implemented server-side but the user has no way to trigger it. Suggest either landing the UI in this PR or moving the "manual claim" bullet (and the related comments) to PR-3.

### 2. Stale comment in `packages/web/src/pages/settings/github.tsx`

Lines 13-15 say:
> The legacy `GithubIntegrationPanel` (per-org webhook secret) was deleted along with the per-repo webhook endpoint and the `github_integration` settings namespace.

The PR description itself defers the webhook endpoint + `github_integration` namespace deletion to PR-3 (and `packages/shared/src/schemas/org-settings.ts:232-235` confirms the namespace is still wired). Suggest: `"panel removed here; webhook endpoint + namespace are removed in PR-3"`.

### 3. Doc references that don't resolve

- PR description points to `docs/pr-300-as-built-design.md` — not present on this branch (or on PR-322's branch).
- `packages/shared/src/config/server-config.ts:84` cites `docs/github-app-design-zh.md §5.2` — also missing.

If these are meant to land somewhere, please add them; otherwise drop the references so future readers don't chase ghosts.

### Nits (optional)

- `packages/server/src/api/auth/github.ts:344`: `oauthTokens` doc says `"in the App flow — always includes the full pair"`, but `dev-callback` without `DEV_GITHUB_PAT` calls into the same path with an empty bundle. Suggest `"may include"`.
- `packages/server/src/__tests__/setup.ts:13-15` sets `FIRST_TREE_HUB_DEV_CALLBACK_ENABLED=1` globally, while `oauth-flow.test.ts:102-140` mutates and restores it. Vitest per-worker env makes this safe today; worth a comment so a future move to a different concurrency model doesn't reintroduce flakes.

### What looks great

- **P0-2 hijack defense** — `/user/installations` membership check before any `installation_id` is bound, with fail-closed on lookup error. Clean.
- **P1-3 target-org via signed state** — state JWT carries `targetOrganizationId`, callback re-checks the user is *still* an active admin against the live `members` row, covering the membership-revoke window inside the JWT TTL.
- **P1-4 refresh-on-expiry in `/me/github/repos`** — 60s buffer, persists the rotated refresh token before returning, precise 401→403 / other→503 distinction. Nicely done.
- **`bindInstallationToOrg`** — conditional UPDATE plus post-UPDATE zero-rows reconciliation plus unique-violation path together cover both concurrent races and the D2 1:1 invariant.
- **Test coverage** — both the success and 403 branches of target-org binding; single-orphan auto-claim vs multi-orphan bail; the full 200/400/403/404/409 matrix for `/claim`; install-url cookie/state JWT round-trip including the slug-missing 503.

### Risk

Agree with the 🟡 Medium framing. Sign-in is fully rewritten, but the returning-user-without-install branch is exercised by the dev-callback flow and the new install/update/request `setup_action` enum keeps GitHub's reconfigure round-trips from being rejected at the schema gate.

Not blocking on my side, but please address (1) and (2) before merge, and clarify (3). I'll defer the approve/request-changes call to @baixiaohang.

Gandy2025:
@baixiaohang Thanks for the thorough review. All 5 items addressed in commit `5551634`:

### 1. UX gap — `/claim` has no UI

Chose **option B** (move "manual claim" out of PR-2's UI scope, keep the backend endpoint). Reasoning: the orphan-list endpoint + Settings picker would add ~250 LOC to PR-2 and adds another moving part to the sign-in flow review surface. Tracked as #318 with the implementation sketch.

Updated to match:
- `api/orgs/github-app.ts` claim docstring rewritten — drops the "Settings panel surfaces a button" claim, marks UI as deferred to #318
- `api/auth/github.ts` orphan-skip log lines now say "operator must POST `/claim` to pick (UI #318)" instead of implying a picker exists
- PR description updated — `/claim` line now reads "API-only in this PR; the Settings 'Claim install' picker UI is tracked in #318"

### 2. Stale comment in `pages/settings/github.tsx`

Rewrote lines 11-15 to explicitly say the per-org webhook endpoint + `github_integration` namespace deletion lands in PR-3, not here. Until PR-3 ships, the legacy webhook route still accepts deliveries — this PR only removes the Settings UI that wrote into the namespace.

### 3. Doc references that don't resolve

- `server-config.ts:84` — removed the `docs/github-app-design-zh.md §5.2` citation entirely (kept the substance of the explanation in the docstring). The design doc lives on PR #295 which isn't merged to main; future readers shouldn't have to chase across PRs.
- PR description — dropped the `docs/pr-300-as-built-design.md` reference. That was a session artifact that never made it to a branch.

### Nit a — `oauthTokens` docstring (`api/auth/github.ts:344`)

`"always includes the full pair"` → `"typically includes the full pair … callers must tolerate the empty shape"`. Honest about the dev-callback empty-bundle case.

### Nit b — per-worker env safety note in `setup.ts`

Added a comment explaining that the `oauth-flow.test.ts` mutate-and-restore pattern on `FIRST_TREE_HUB_DEV_CALLBACK_ENABLED` is safe under vitest's `forks` pool (each worker gets its own `process.env`) but would leak under a thread-pool model. Re-evaluate then.

### Verification post-fix

- `pnpm check` — clean (652 files)
- `pnpm typecheck` — 9 packages clean
- `pnpm --filter @first-tree-hub/server test` — **814/814 pass**

Diff is at the new commit `5551634` (kept as a separate commit from the original `1e4ab35` so the review-driven changes are visible without diffing across the whole PR).

Ready for your re-review whenever.

baixiaohang (approved):
All 5 review items from the previous round were addressed in 5551634 — comments are now accurate, doc references resolve (or are explicitly deferred), and `POST /claim` is correctly flagged as API-only with the UI tracked in #318. Pure documentation/comment/test-comment changes, zero behavior diff, 814/814 still pass.

LGTM 🚀

PR #326, opened 2026-05-12
feat(web): pin "Your agents" first on Team tab + polish dialog & columns

Gandy2025:
## Summary

- **Team tab restructure**: pin "Your agents" section first (manager=self, any visibility), then Humans → Team agents (others' shared) → Other members' private agents (admin-only, collapsed). Each agent appears in exactly one section. Rows inside Your agents carry an inline visibility chip (🔒 Private / 👥 Shared); homogeneous sections skip the chip. Filter pills updated: `All / Your agents / Humans / Team agents`.
- **NewAgentDialog**: default visibility flipped to `private` (matches the `personal_assistant` type name). "Where it runs" block reserves a stable min-height and renders SingleComputerCard-shaped skeletons during async detection — the dialog no longer visibly jumps as `listClients` / capabilities stream in.
- **Table polish**: Delegate column gets `truncate + title` (was wrapping to 2 lines at 160); widened 160 → 180. Name column shrunk 260 → 220 (inner truncate covers the long tail). Collapsible chevron on "Other members' private agents" now hangs out-of-flow to the left of the section icon, so all four section icons stay at the cell's left edge and section titles keep their vertical alignment with row-name text.

## Background

User feedback on the Team tab:
1. New Agent dialog visibly jumped in height as the "Where it runs" block streamed in.
2. Delegate column wrapped to 2 lines under tight widths.
3. The collapsible group ("Other members' private agents") had its 🔒 icon and title shifted right of the other three sections by the arrow + gap, breaking alignment.
4. The user's own agents were buried in "Shared agents" (default visibility was `organization`); finding "my own stuff" required scanning past everyone else's.

This PR addresses all four.

## Design decisions (locked with the user)

| # | Decision | Choice |
|---|---|---|
| 1 | Dialog height stability | `min-height` + skeleton cards that mirror the real cards' dimensions |
| 2 | Column widths | Name 220 / Delegate 180 / Manager 180 / Runs on 220 / Status 160 / Created 160 / Actions 44 |
| 3 | Section header alignment | Chevron hangs absolute-positioned to the left of the icon; icons stay at cell X=0 across all sections, titles keep aligning with row-name text at sp-6 |
| 4 | Section order | Your agents → Humans → Team agents → Other private (admin) |
| 5 | Your agents internal layout | Flat list + inline visibility chip per row (private sorted first) |
| 6 | Visibility chip range | Only in heterogeneous sections (Your agents); homogeneous sections rely on section title |
| 7 | Default visibility | Switched `organization` → `private` |
| 8 | Admin "Your agents" semantics | Still means "managed by me" — admin governance is expressed via the separate Other private bucket |

## Test plan

- [x] `pnpm --filter @first-tree-hub/web typecheck` (incl. `lint:tokens` design-token guardrails) ✓
- [x] `pnpm check` (biome) ✓
- [x] `pnpm --filter @first-tree-hub/web test` — 80 tests pass (was 76), including 8 in `team-groups.test.ts` (was 4; new tests cover section order, visibility-chip placement, admin-only governance bucket, filter-pill scoping)
- [x] Manual: open Team tab — verify "Your agents" renders first, visibility chips show only there, "Other members' private agents" section header aligns vertically with the other three sections
- [x] Manual: click "+ New agent" — verify dialog opens at stable height (no jump as computer/capabilities resolve), "Private to you" is pre-selected with `(default)` annotation
- [x] Manual: long delegate name — verify Delegate cell truncates on one line with tooltip showing full label
- [x] Manual (admin): verify "Other members' private agents" section is collapsible (▸ / ▾) and chevron hangs to the left without displacing the 🔒 icon

### DIFF
--- a/packages/web/src/components/new-agent-dialog.tsx
+++ b/packages/web/src/components/new-agent-dialog.tsx
@@ -694,6 +694,14 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
             the surrounding form stays put. Empty `0 computers` and N-radio
             states can still grow taller — that's fine, the jump is only
             from the *initial detection* states which were the visible bug.
+
+            `minHeight: 168` is hand-calibrated to the steady-state height
+            of `SingleComputerCard` (border + 2-line content + p-3) plus
+            the `<Label>` + a single-line "Powered by" + a runtime chip
+            row. KEEP IN SYNC with `SingleComputerCard` /
+            `ComputerCardSkeleton` / `RuntimeChipsSkeleton` below — if any
+            of those grows or shrinks its padding / line count, the value
+            here must move with it or the dialog will jump again.
           */}
           <div className="space-y-3" style={{ minHeight: 168 }}>
             <Label>Where it runs</Label>
@@ -850,6 +858,13 @@ function ZeroComputerBlock({
  * Single-computer card. Used in the common case where the user only has
  * one computer connected — no radio, just a visual confirmation of where
  * the agent will live.
+ *
+ * Layout note: the "Where it runs" wrapper above reserves a `minHeight`
+ * calibrated to this card's steady-state dimensions plus the runtime
+ * chip row, so the dialog doesn't jump during async load. If you change
+ * this card's padding, border, or line count, also revisit `minHeight`
+ * in the wrapper (and `ComputerCardSkeleton` / `RuntimeChipsSkeleton`
+ * below, which mirror this card's dimensions).
  */
 function SingleComputerCard({ client }: { client: HubClient | undefined }) {
   if (!client) return null;
--- a/packages/web/src/components/new-agent-dialog.tsx
+++ b/packages/web/src/components/new-agent-dialog.tsx
@@ -148,7 +148,13 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
   const [displayName, setDisplayName] = useState("");
   const [name, setName] = useState("");
   const [nameDirty, setNameDirty] = useState(false);
-  const [visibility, setVisibility] = useState<AgentVisibility>("organization");
+  // Default is "private": the surfaced agent type is `personal_assistant`
+  // (literally "personal"), so the conservative default is "only I see it".
+  // Sharing with the team is an explicit decision the user opts into; the
+  // previous default ("organization") quietly published every onboarding
+  // agent into the team roster, which surprised users who expected
+  // personal-assistant to mean personal.
+  const [visibility, setVisibility] = useState<AgentVisibility>("private");
   const [runtime, setRuntime] = useState<RuntimeProvider>("claude-code");
   // The @handle editor is collapsed by default — 99% of users keep the
   // auto-derived slug. The preview line under "Display name" surfaces it
@@ -182,7 +188,7 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
       setDisplayName("");
       setName("");
       setNameDirty(false);
-      setVisibility("organization");
+      setVisibility("private");
       setRuntime("claude-code");
       setEditingHandle(false);
       setConnectedClients([]);
@@ -649,7 +655,7 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
                 <div>
                   <div className="text-body font-medium">Shared with team</div>
                   <div className="text-caption text-muted-foreground">
-                    Anyone in your org can @mention and chat with this agent. (default)
+                    Anyone in your org can @mention and chat with this agent.
                   </div>
                 </div>
               </label>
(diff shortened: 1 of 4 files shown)


baixiaohang (commented):
Reviewed the diff. Overall a tidy UI-polish PR — the three-way partition (`yourAgents / teamAgents / otherPrivateAgents`) is genuinely disjoint, the rationale comments are well-written, and the rewritten test now finds groups by key instead of index, which makes it robust against future section reordering. A few notes below; tagging this as a Comment rather than Approve so you can pick up the items first.

### Suggested changes

1. **`HumanDelegateCell` placeholder `Set delegate →` → `Set →`** (`packages/web/src/pages/team/team-table.tsx`, around the new `HumanDelegateCell` block). The inline rationale says the shorter label is needed because the Delegate column is narrow, but this same PR widens Delegate from 160 → 180. The reason no longer holds; the empty-state CTA reads as too cryptic at the new width. Suggest restoring `Set delegate →`.

2. **`yourRows` needs a stable secondary sort** (`packages/web/src/pages/team/index.tsx`, inside `buildGroups`). The current `aPriv - bPriv` only pins private to the top — within each visibility bucket the order is whatever `yourAgents` arrived in. If `/agents` ever returns a different order across paginated fetches (or after a refetch), the user's own section will visibly reshuffle. Suggest a tiebreaker on `displayName` (or `createdAt` desc) so the section is deterministic.

### Discussion points (non-blocking)

3. **`minHeight: 168` magic number** in `new-agent-dialog.tsx`. The comment correctly explains the trade-off, but the value is hand-calibrated to `SingleComputerCard`'s current dimensions. If anyone later tweaks the card's padding or text-body line-height, the dialog will jump again. Either compute it from spacing tokens, or at minimum add a `// keep in sync with SingleComputerCard` pointer next to both the card and the constant.

4. **Chevron alignment depends on parent left-padding** (`GroupHeaderRow` in `team-table.tsx`). The chevron is `position: absolute; left: calc(-1 * var(--sp-3))` — works only because the table's container provides `sp-5` left padding, as the comment notes. That's an implicit contract that won't fail at compile time. Two options: keep as-is and accept it (the comment is already there), or render the chevron inline with a fixed-width slot to the left of the icon so the layout is self-contained. I'd lean toward the comment-only path unless you've seen the outer padding shift in the past.

5. **Filter pills lost "Your private" granularity.** The old pill let users one-click filter to their private agents; the new "Your agents" pill mixes private + shared and relies on the inline chip. Design decision #5 captures that this was intentional, just flagging for visibility — if any teammate routinely scanned only-my-private agents, that's a small workflow regression.

6. **Manual test plan checkboxes are all unchecked.** CI covers typecheck + biome + vitest, but the four manual items (dialog stability, section alignment, delegate truncation, admin chevron) carry the actual visual claims of this PR. Could you confirm those were exercised locally? If yes, please tick them; if no, worth running through before merge.

Once 1 and 2 are addressed I'm happy to re-review.

Gandy2025:
Thanks for the careful review @baixiaohang — fixup pushed as `ed878ee`.

### Required (1–2): fixed

**1. `Set delegate →` restored.** You're right — the abbreviation was sized for the old 160px column, and this PR widens it to 180px. The shorter label is no longer earning its cryptic-ness; reverted.

**2. `yourRows` now has a deterministic secondary sort.** Private agents still pin to the top; within each visibility bucket, rows sort by `displayName` (locale-aware, case-insensitive via `localeCompare(..., { sensitivity: "base" })`) so "Coder" and "coder" land adjacent. Added a test that scrambles input (`Zeta, Beta, alpha, gamma` interleaved private/shared) and asserts the expected order, so a future refactor that drops the tiebreaker won't pass.

### Discussion (3–4): comment-only, accepted

**3. `minHeight: 168` magic number.** Agreed it's load-bearing in a way that won't fail at compile time. Took the "comment pointer" path you suggested rather than computing from tokens — calibration depends on more than spacing (line-height on the muted "last seen" caption, border thickness on the card, runtime chip's vertical padding), so a token-derived expression would be its own kind of fragile. Now there's a `KEEP IN SYNC` block in the `minHeight` declaration AND a back-pointer in `SingleComputerCard`'s docstring referencing the wrapper. Future maintainers will hit one or the other.

**4. Chevron alignment depends on outer padding.** Took the same comment-only path. The implicit contract on `sp-5` outer padding from `TeamPage` is now spelled out in the chevron's inline style (specific file, specific value, what shrinking it below `sp-3` would break), so a future refactor that touches the page wrapper has a fighting chance of catching it.

### Non-blocking (5–6)

**5. Lost "Your private" granularity.** Confirmed intentional (decision #5 in the PR body). The replacement workflow for "show me only my private stuff" is `Your agents` pill + scan-for-🔒-chip — added cost is one visual filter step. If we see signal that anyone actually scanned only-my-private as a frequent task, we can revisit (e.g. a sub-toggle inside Your agents); leaving it out for now to keep the pill row from re-fragmenting.

**6. Manual test plan.** Ticked all four boxes in the PR body. Exercised against the local dev stack (devuser org with 2 humans + 1 personal_assistant). Dialog stability (no jump on async load) and section alignment (chevron overhanging into the page's left padding without displacing the icon) confirmed visually. Long-delegate truncation and admin-collapsed `Other members' private agents` are covered programmatically by the test suite — there's no realistic long-name human or other-member-private agent in the dev DB to exercise them by eye, but the contract is locked: `HumanDelegateCell` uses the same truncate+title recipe as `Runs on` (which has been visually verified on long FQDNs in prior PRs), and the `buildGroups` test asserts `other-private` only renders for `isAdmin=true` with `collapsible: true`.

Ready for re-review.

baixiaohang (approved):
Fixup `ed878ee` verified — all four required/discussion items landed cleanly:

1. `Set delegate →` restored at `team-table.tsx:425-437`, old abbreviation rationale comment removed.
2. `yourRows` now sorts visibility-first with a `localeCompare(..., { sensitivity: "base" })` tiebreaker (`index.tsx:581-603`). New test `team-groups.test.ts:155-210` scrambles `Zeta / Beta / alpha / gamma` across both visibility buckets and asserts the expected `["private-b", "private-g", "shared-a", "shared-z"]` order — locks the contract against future refactors that drop the tiebreaker.
3. `minHeight: 168` now has the KEEP IN SYNC block at the wrapper site and a reciprocal pointer in `SingleComputerCard`'s docstring — both directions covered.
4. Chevron's IMPLICIT CONTRACT comment spells out the `sp-5` outer-padding dependency on `TeamPage` and the `sp-3` breaking threshold.

Decisions on #5 (intentional loss of "Your private" granularity, recoverable via `Your agents` pill + 🔒 chip scan) and #6 (manual test plan, with long-name / other-private cases covered programmatically since the dev DB lacks realistic fixtures) are reasonable.

LGTM.

PR #330, opened 2026-05-12
[codex] Polish Context tab overview

Gandy2025:
## Summary

- Align the Context tab layout with the Team tab content width and spacing.
- Simplify the status, filter, update list, and selected-change panels into a quieter master-detail view.
- Refine Tree Overview expansion behavior: clearer Show more/less copy, singleton promotion, scoped auto-scroll, and cleaner expanded toggle edges.
- Add a dev preview shell that mirrors the authenticated layout framing for visual review.

## Validation

- `pnpm check`
- `pnpm typecheck`
- `pnpm test`

## Notes

- Web-only change; no command/client package version bump needed.

### DIFF
--- a/packages/web/src/pages/__tests__/context-overview.test.ts
+++ b/packages/web/src/pages/__tests__/context-overview.test.ts
@@ -38,30 +38,59 @@ function realNodeIds(overviewNodes: ReturnType<typeof buildOverviewNodes>): stri
 }
 
 describe("buildOverviewNodes", () => {
-  it("shows only root and top-level areas when there are no changed counts or selected path", () => {
-    expect(realNodeIds(buildOverviewNodes(nodes, null, new Map(), new Set()))).toEqual(["root", "agents", "members"]);
+  it("auto-promotes singleton hidden children alongside the top-level areas", () => {
+    // `agents` (only child: agents/runtime) and `members` (only child:
+    // members/yzw) each have a single hidden child. Folding one row behind
+    // a "Show 1 more" toggle is wasted UI — promote it directly. The
+    // promotion is transitive: once members/yzw is surfaced it's checked
+    // again, but it has two children (notebook + journal), so the chain
+    // stops there.
+    expect(realNodeIds(buildOverviewNodes(nodes, null, new Map(), new Set()))).toEqual([
+      "root",
+      "agents",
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):

baixiaohang (approved):
LGTM — high-quality visual/interaction polish PR. Motivations are clearly captured in code comments, tests cover the new behavior, and the scope stays cleanly inside `packages/web`.

### What I like

- **Status / Filter separation.** Pulling the window selector + metrics out of the status banner into a dedicated `ContextFilterBar` is the right call — status is now a pure info surface, controls live in a chip row that matches the Team-tab rhythm.
- **Card → row for updates.** Drops the boxes-in-boxes feel; hairline-bottom + `--bg-sunken` selection reads as one continuous list inside the panel.
- **Master-detail height as a ceiling, not a fixed value.** `maxHeight: min(52vh, 30rem)` lets short content hug naturally and long content scroll inside the panel — no more dead space below the smaller side.
- **Singleton hidden-child promotion.** A "Show 1 more" toggle costs the same row as the row it would unfold; promoting it directly is the right call, and the iterative loop correctly walks only-child chains.
- **Scoped auto-scroll.** Replacing `Element.scrollIntoView` with `container.scrollBy` against a captured `treeScrollRef` fixes the page-jump-on-update-selection issue surgically.
- **"Show N more" / "Show less" copy.** Reads as an action verb instead of a status string, and the `· N changed` tail only shows up when it carries information.
- **Summary toggle sorted first, plus dropped edge while expanded.** Click target stays put across collapse/expand, and the toggle stops competing with real data nodes once expanded — both are nice mental-model wins.
- **Reference block collapsed to inline `·` line.** Matches the prose register of the update sub-line and de-weights what was previously the heaviest block in the detail panel.
- **`ContextPreviewPage` shell.** Mirroring the `<Layout>` framing (header + `p-6 max-w-[1280]` container) for the dev preview is exactly what the visual review needed; nice that the header is explicitly presentational.

### Tests

`context-overview.test.ts` covers the new behavior well — singleton promotion, selected-leaf interaction with promotion, and the "2+ hidden children still keep a toggle" case are all asserted with comments explaining why each shape is expected.

No blockers from me. Approving.

Issue #342, opened 2026-05-13
New Chat default agent is volatile — `pickDefault` keys on `runtimeUpdatedAt` and shifts between clicks


liuchao-001:
## Symptom

Opening **New chat** from the conversation list seeds a default participant chip. The seeded agent **changes from one click to the next**, even when nothing about the user's intent has changed — so users perceive it as "the system randomly picks a different agent each time." When the user hits Enter without inspecting the chip, the first message goes to whichever agent happened to be most-recently-active.

## How the default is computed today

`packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:523-538`:

```ts
function pickDefault(candidates: MentionCandidate[], activity: RuntimeAgent[]): string | null {
  const ids = new Set(candidates.map((c) => c.agentId));
  const mine = activity.filter((a) => ids.has(a.agentId) && a.managedByMe);
  if (mine.length === 0) return null;

  // Priority 1: most recently active by runtimeUpdatedAt
  const mru = [...mine]
    .filter((a) => a.runtimeUpdatedAt)
    .sort((a, b) => (b.runtimeUpdatedAt ?? "").localeCompare(a.runtimeUpdatedAt ?? ""))[0];
  if (mru) return mru.agentId;

  // Priority 2: any personal_assistant
  const pa = mine.find((a) => a.type === "personal_assistant");
  if (pa) return pa.agentId;

  // Priority 3: first my-managed agent
  return mine[0]?.agentId ?? null;
}
```

Effectively: among agents the caller manages, pick the one whose runtime touched most recently.

## Why this produces the perceived randomness

`runtimeUpdatedAt` reflects **runtime presence** — it ticks whenever:

- an agent receives any inbound message (incl. from cron / webhook / another agent)
- a session starts, runs, or finishes
- the client heartbeats the agent's runtime

None of those signals correspond to "the human is about to start a new chat with this agent." So whichever of the user's agents is currently busiest wins, and that flips constantly. Two clicks 10 seconds apart can default to two different agents.

PR #328 (5/12) introduced this `managedByMe` scoping — the prior behavior was even worse (could default to a coworker's agent). But MRU on `runtimeUpdatedAt` was kept as the tiebreaker inside the my-managed subset, which is the residual problem here.

## Two options to consider

(Per discussion in chat — not a fix proposal, just the directions worth weighing.)

[...]

PR #350, opened 2026-05-13
fix(web): chat picker source — show humans + stabilize new-chat default

### LINKED ISSUE #343: Web: cannot add humans as chat participants — picker source (`activity.agents`) filters them out via `runtimeState IS NOT NULL`
## Want

When adding a participant to an existing chat from the **web frontend**, the picker should be able to select **human members** of the organization — not only AI agents.

## Current behaviour

Opening the `[+]` participant picker on a chat's `ParticipantsHeader` shows AI agents only. Human members of the same org are not offered as candidates, even when the caller is one of those humans themselves. There is no way from the web UI to invite a coworker into an existing chat.

## Root cause — picker's candidate source filters humans

The picker's candidate list is computed in `packages/web/src/pages/workspace/center/chat-view.tsx:861-893` (`mentionCandidates`). It unions two sources:

1. Existing `chatDetail.participants`
2. `activity.agents` from `GET /api/v1/orgs/:orgId/activity`

There is **no client-side `type === 'human'` filter** — the `MentionCandidate` type doesn't even carry `type`. So the gap isn't in the picker itself.

The gap is in source (2). `getActivityOverview` ultimately calls `listAgentsWithRuntime` (`packages/server/src/services/activity.ts:145-172`), which restricts results to agents with `runtimeState IS NOT NULL`:

```ts
.where(and(
  isNotNull(agentPresence.runtimeState),
  agentVisibilityCondition(scope.organizationId, scope.memberId),
));
```

Human agents (`agents.type = 'human'`) are mirrors of human users — they never bind a runtime, so `agent_presence.runtimeState` is always null for them. They drop out at this filter before they ever reach the web picker.

## Server-side `add` already accepts humans

`addMeChatParticipants` (`packages/server/src/services/me-chat.ts:344+`) has no `type` filter. It validates cross-org and existence, then inserts the participant. So **the server doesn't need to change** to accept human participants — the only gap is the picker source.

This means the same gap applies to the **NewChatDraft** picker too (`packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:71-86`) — it consumes the same `activity.agents` source. Fixing the source fixes both surfaces at once.

## In scope

- The web picker (`ParticipantsHeader` `[+]` dropdown in existing chats, and the participant chip picker in `NewChatDraft`) can select human members of the caller's organization.
- Once selected, `addMeChatParticipants` adds them as a speaker, exactly as it does for AI agents today.
- The picker's existing grouping (managed-by-me vs teammates') still applies sensibly — humans probably want their own visual group, but that's implementer's call.

## Out of scope (intentional)

- Whether to widen `/orgs/:orgId/activity` to include humans (probably wrong shape — that endpoint is keyed off runtime presence and is also consumed by roster / pulse / clients pages), or to add a new endpoint (e.g. `GET /me/org/members` returning humans + their agent uuids), or to query `agents` directly. Implementer's call.
- Whether humans should appear with a different UI badge / grouping in the picker. UX detail.
- Anything about how added humans receive notifications / participate (existing `chatMembership` and runtime delivery rules already cover this — humans on a `mention_only` participant row get notified via their `delegateMention` agent).
- The reverse direction (removing participants, including humans) — `DELETE /me/chats/:id/participants/:agentId` is a separate concern, see `ParticipantsHeader`'s comment about the missing endpoint.

## Acceptance

- A human in the same organization can be selected from the `[+]` participant picker of an existing chat in the web UI.
- The same human can be selected as a participant when creating a new chat from **New chat**.
- Adding the human results in a `chatMembership` row for them with `accessMode='speaker'`, identical to adding an AI agent today.

## Code pointers

- `packages/web/src/pages/workspace/center/chat-view.tsx:861-893` — `mentionCandidates` definition; the picker consumes this.
- `packages/web/src/pages/workspace/center/chat-view.tsx:1574+` — `ParticipantsHeader` component, the `[+]` button + dropdown.
- `packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:71-86` — `NewChatDraft`'s `candidates` definition (same `activity.agents` source).
- `packages/server/src/services/activity.ts:145-172` — `listAgentsWithRuntime`; the `isNotNull(agentPresence.runtimeState)` filter that drops humans.
- `packages/server/src/services/me-chat.ts:344+` — `addMeChatParticipants`; confirms server side does not reject `type='human'`.

liuchao-001:
## Summary

- **Closes #343**: web participant pickers (`ParticipantsHeader` `[+]` in existing chats, chip picker in `NewChatDraft`) couldn't surface human members of the org. Root cause: both consumed `/orgs/:orgId/activity`, whose `listAgentsWithRuntime` service filters on `runtimeState IS NOT NULL`; humans never bind a runtime, so they were silently dropped. Switched picker data source to the existing `/orgs/:orgId/agents` endpoint (`listAgentsForMember`), which already `LEFT JOIN`s `agent_presence` and surfaces humans natively. Server unchanged.
- **Partial #342**: dropped the `runtimeUpdatedAt` MRU branch from `pickDefault` in `new-chat-draft.tsx`. Default chip is now stable across clicks (was flipping with runtime activity). A more deliberate default (e.g. `delegateMention`) stays out of scope and tracked in #342.

## Design choices

- **Server zero-change.** `listAgentsWithRuntime` keeps "AI agents with a runtime" semantics for Roster / Cmd-K / Computers / Pulse. The picker had been borrowing this endpoint; the fix is to swap the picker's data source, not widen `/activity`.
- **`@` vs `[+]` semantic split** in `chat-view.tsx`:
  - Composer `@`-autocomplete → only current chat participants.
  - Header `[+]` picker → org-wide (humans + AI).
  - The pre-existing "type `@<outsider>` auto-invites them" path is removed — inviting a new participant is now an explicit `[+]` action.
- **`pickDefault` priorities** post-fix: my-managed `personal_assistant` → first my-managed AI agent → `null`. Humans never seed.

## Test plan

- [x] `pnpm check` — clean
- [x] `pnpm typecheck` — 9/9 packages
- [x] `pnpm --filter @first-tree-hub/server test` — 850/850 tests pass
- [x] `pnpm --filter @first-tree-hub/web test` — 12 files / 101 tests pass (includes 10 new `pickDefault` unit tests)
- [x] New server-side regression test: `addMeChatParticipants` accepts a human as `access_mode='speaker'` (locks the server contract so a future `type` filter can't silently undo the picker fix).
- [x] New web-side unit test: `pickDefault` — 10 cases, including a "stable across calls" regression lock for #342.
- [ ] Manual UI verification (deferred to deploy):
  - [ ] Existing chat `[+]` picker shows humans
  - [ ] Existing chat composer `@` candidates = current participants only
  - [ ] New chat chip `[+]` picker shows humans + AI
  - [ ] New chat default chip is stable across multiple open/close cycles

### DIFF
--- a/packages/server/src/__tests__/me-chat-service.test.ts
+++ b/packages/server/src/__tests__/me-chat-service.test.ts
@@ -688,4 +688,40 @@ describe("chat-first workspace service layer", () => {
     });
     expect(res.rows.find((r) => r.chatId === chatId)?.engagementStatus).toBe("active");
   });
+
+  /**
+   * Regression for #343 — the picker-source widening (#343) is only useful if
+   * the server's add-participant path also accepts human agents. The service
+   * already had no `type` filter, but a regression here would silently undo
+   * the picker fix. Lock the invariant: a human added via `addMeChatParticipants`
+   * lands as `access_mode = 'speaker'`, identical to adding an AI agent today.
+   *
+   * `createTestAdmin` lands every member in the shared default org, so the
+   * owner + invitee live in the same org without extra wiring.
+   */
+  it("addMeChatParticipants: accepts a human agent and inserts a speaker row (#343)", async () => {
+    const app = getApp();
+    const owner = await createTestAdmin(app);
+    const teammate = await createTestAdmin(app);
+    expect(teammate.organizationId).toBe(owner.organizationId);
+
+    // Seed a direct chat with one AI agent so owner is a speaker.
+    const aiAgent = await createTestAgent(app, { name: "agp-h-ai" });
+    const { chatId } = await createMeChat(app.db, owner.humanAgentUuid, owner.organizationId, {
+      participantIds: [aiAgent.agent.uuid],
+    });
+
+    await addMeChatParticipants(app.db, chatId, owner.humanAgentUuid, owner.organizationId, {
+      participantIds: [teammate.humanAgentUuid],
+    });
+
+    // Raw SQL: `access_mode` isn't surfaced by any service-level read
+    // path (listMeChats etc. project it as `membershipKind`, not as the
+    // raw column). Going to the table directly is the cleanest way to
+    // assert what the writer produced.
+    const rows = await app.db.execute<{ access_mode: string }>(
+      sql`SELECT access_mode FROM chat_membership WHERE chat_id = ${chatId} AND agent_id = ${teammate.humanAgentUuid}`,
+    );
+    expect(rows[0]?.access_mode).toBe("speaker");
+  });
 });
--- a/packages/web/src/components/mention-autocomplete.tsx
+++ b/packages/web/src/components/mention-autocomplete.tsx
@@ -31,7 +31,8 @@ export type MentionCandidate = {
   /** True iff the caller's member is this agent's `managerId`. Drives
    *  participant-picker grouping (mine first, then teammates') and the
    *  empty-query branch of mention autocomplete so the user's own agents
-   *  always surface first. Populated from `/activity` (server-derived);
+   *  always surface first. Derived client-side by callers from the agent
+   *  row's `managerId` against the caller's `memberId` (see `useAuth`);
    *  callers that can't determine it should pass `false` rather than
    *  omitting — there's no "unknown" state. */
   managedByMe: boolean;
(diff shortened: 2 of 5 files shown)


liuchao-001:
Closing as superseded by main drift. The fix is still needed (issue #343 remains open — `/activity`'s `runtimeState IS NOT NULL` still filters humans out of the picker on main), but #488's add-participant-dropdown refactor restructured the surface this PR targets, so the diff no longer applies cleanly. Re-doing as a fresh, smaller PR against current main.

PR #351, opened 2026-05-13
refactor(server): simplify notification infra — cross-instance fanout, DB-backed dedup, push-only bell

yuezengwu:
## Summary

Refactors the hub-wide notification pipeline along the proposal lines from the prior PR review thread (#348). Five atomic commits, each independently reviewable; squash-merge picks whichever framing reads best in the log.

Five changes, in order of dependency:

1. **`refactor(server): delete dead notification types`** — `agent_connected` and `agent_needs_decision` had schema entries, message composers, and UI labels but zero producers anywhere in the tree. Removed end-to-end.

2. **`feat(server): cross-instance notification fanout via pg_notify`** — extends the existing `Notifier` with a generic `admin_broadcast_envelopes` channel. Producers call `broadcastAdminsCrossInstance(payload)`; every server instance LISTENs, JSON-parses the envelope, and feeds it into its local `broadcastToAdmins` fanout. Today's in-memory broadcaster delivered notifications only to admin sockets on the producing instance — admins connected elsewhere silently missed events during failover or rolling restart. Tests + per-instance subsystems (pulse aggregator) keep calling `broadcastToAdmins` directly because their fanout is intentionally process-local.

3. **`feat(server): DB-backed notification dedup via dedup_key partial unique index`** — replaces the in-memory `shouldNotify` Map with `(organization_id, dedup_key) WHERE read = false` uniqueness. Migration `0041` adds the column + index; `createNotification` uses `ON CONFLICT DO NOTHING`. `notifyAgentEvent` now defaults `dedup_key` to `chat:{chatId}:{type}` or `agent:{agentId}:{type}`, so per-event dedupe is opt-out rather than opt-in. All three `shouldNotify` call sites in `ws-client.ts` are dropped — one fewer place where a new event type forgets to dedup.

4. **`feat(server): unread-count endpoint + SQL-pushed visibility filter for list`** — `GET /orgs/:orgId/notifications/unread-count` returns a single `SELECT COUNT(*)`. The visibility filter (`agent_id IS NULL OR agent_id IN visible`) moves from in-Node post-filter to SQL predicate, so the old 4× overscan + 400-row ceiling that hid long-tail visible rows behind invisible-agent noise is gone. The fragment is centralised in `buildVisibilityCondition` so list / markAllRead / unreadCount can't drift apart on what 'visible' means.

5. **`refactor(web): notification bell drops 10s polling, runs push-only`** — wires `useAdminWs()` into Layout so the bell receives pushes on every page where it renders. Drops `refetchInterval: 10_000` from both bell queries. Switches the unread-count source from `listNotifications({ read: false, limit: 100 }).length` (capped at 100) to the new `/unread-count` endpoint (accurate past 100). Mark-all-read responds immediately now that there is no 10s ceiling.

## Migration

`packages/server/drizzle/0041_notifications_dedup_key.sql`

- `ALTER TABLE notifications ADD COLUMN dedup_key text;`
- `CREATE UNIQUE INDEX uq_notifications_org_dedup_unread ON notifications (organization_id, dedup_key) WHERE read = false AND dedup_key IS NOT NULL;`

Purely additive — `dedup_key` is nullable, no back-fill required, producers without a key keep their always-insert semantics.

Generated by hand because `drizzle-kit generate` is blocked by a pre-existing snapshot collision in `drizzle/meta/_journal.json` (0016/0018) unrelated to this change. Other recent migrations in the repo (0038–0040) are similarly hand-written with rationale comments, so this matches existing practice.

## Cross-package version impact

None — `command` / `client` source doesn't import any of the touched `shared` modules (`packages/shared/src/schemas/notification.ts` is server+web only). No version bump needed.

## Test plan

- [ ] Local: `pnpm typecheck` clean (verified)
- [ ] Local: `pnpm check` clean (verified)
- [ ] Local: server vitest (skipped — testcontainers needs Docker, not available in author env)
- [ ] `migrate up` on a fresh DB produces the new column + partial unique index
- [ ] Two-instance manual smoke: notification produced on instance A reaches an admin socket on instance B
- [ ] Floods of identical `agent_error` events for one agent collapse to a single unread row until acked
- [ ] Bell badge updates within one WS round-trip after mark-all-read (no 10s lag)
- [ ] Bell shows accurate unread count past 100

### DIFF
--- a/docs/chat-first-workspace-product-design.md
+++ b/docs/chat-first-workspace-product-design.md
@@ -632,8 +632,6 @@ Notification bell
 ├─ agent_error
 ├─ agent_blocked
 ├─ agent_stale
-├─ agent_disconnected
-├─ agent_connected
 ├─ session_error
 ├─ session_completed
 └─ computer / system / organisation events
--- a/packages/server/drizzle/0041_notifications_dedup_key.sql
+++ b/packages/server/drizzle/0041_notifications_dedup_key.sql
@@ -1,0 +1,29 @@
+-- DB-backed deduplication for the notifications table.
+--
+-- Before this migration, the server kept an in-memory Map keyed by
+-- `(agentId, notificationType)` to suppress duplicate notifications within a
+-- five-minute window. That broke in two important ways:
+--
+--   1. Multi-instance deployments: each server process owned its own Map, so
+--      a notification produced on instance A passed instance B's dedupe gate
+--      unchanged. Operators saw the same row inserted multiple times during
+--      load-balancer failover or rolling restart.
+--   2. Process restart wiped the Map, re-opening the dedupe window for
+--      whatever events had fired just before.
+--
+-- The new model is purely DB-driven. A producer optionally supplies a
+-- `dedup_key` (e.g. `agent:{uuid}:error`, `chat:{uuid}:completed`). The
+-- partial unique index below scopes uniqueness to `(organization_id,
+-- dedup_key)` *while the prior row is still unread*. Re-emitting the same
+-- key while the previous notification sits unread is a no-op (the
+-- application uses `ON CONFLICT DO NOTHING`); after the user acknowledges
+-- the prior row, a fresh notification can land again.
+--
+-- Producers without a `dedup_key` keep the legacy always-insert behaviour,
+-- so this column is purely additive — no back-fill, no NOT NULL constraint.
+
+ALTER TABLE "notifications" ADD COLUMN "dedup_key" text;
+
+CREATE UNIQUE INDEX "uq_notifications_org_dedup_unread"
+  ON "notifications" ("organization_id", "dedup_key")
+  WHERE read = false AND dedup_key IS NOT NULL;
--- a/packages/server/drizzle/meta/_journal.json
+++ b/packages/server/drizzle/meta/_journal.json
@@ -288,6 +288,13 @@
       "when": 1779235200000,
       "tag": "0040_chat_user_state_engagement",
       "breakpoints": true
+    },
+    {
+      "idx": 41,
+      "version": "7",
+      "when": 1779321600000,
+      "tag": "0041_notifications_dedup_key",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/server/src/__tests__/pulse-aggregator.test.ts
+++ b/packages/server/src/__tests__/pulse-aggregator.test.ts
@@ -28,6 +28,7 @@ function makeMockNotifier(): {
     notifySessionStateChange: vi.fn(async () => {}),
     notifyRuntimeStateChange: vi.fn(async () => {}),
     notifyChatMessage: vi.fn(async () => {}),
+    notifyAdminBroadcast: vi.fn(async () => {}),
     pushFrameToInbox: vi.fn(async () => 0),
     onConfigChange: vi.fn(),
     onSessionStateChange: vi.fn(),
@@ -35,6 +36,7 @@ function makeMockNotifier(): {
       handlers.push(handler);
     }),
     onChatMessage: vi.fn(),
+    onAdminBroadcast: vi.fn(),
     start: vi.fn(async () => {}),
     stop: vi.fn(async () => {}),
   } satisfies Notifier;
--- a/packages/server/src/api/agent/ws-client.ts
+++ b/packages/server/src/api/agent/ws-client.ts
@@ -85,24 +85,6 @@ const wsMessageSchema = z.object({
  * SDK can refresh and reconnect without silently half-opening the socket.
  */
 
-/** Notification cooldown: prevents duplicate notifications for same (agentId, type) within window. */
-const NOTIFICATION_COOLDOWN_MS = 300_000; // 5 minutes
-const notificationCooldowns = new Map<string, number>();
-
-function shouldNotify(agentId: string, notificationType: string): boolean {
-  const key = `${agentId}:${notificationType}`;
-  const now = Date.now();
-  const lastSent = notificationCooldowns.get(key);
-  if (lastSent && now - lastSent < NOTIFICATION_COOLDOWN_MS) return false;
-  notificationCooldowns.set(key, now);
-  if (notificationCooldowns.size > 1000) {
-    for (const [k, ts] of notificationCooldowns) {
-      if (now - ts > NOTIFICATION_COOLDOWN_MS) notificationCooldowns.delete(k);
-    }
-  }
-  return true;
-}
-
 /**
  * Authenticated WS session state.
  *
@@ -788,9 +770,12 @@ export function clientWsRoutes(notifier: Notifier, instanceId: string) {
                 notifier,
               });
 
-              if (payload.runtimeState === "error" && shouldNotify(agentId, "agent_error")) {
+              // Dedup happens at the DB layer via the default agent:{id}:{type}
+              // dedup_key — repeated runtime-state errors for the same agent
+              // collapse to a single unread row until the user acks.
+              if (payload.runtimeState === "error") {
                 notificationService.notifyAgentEvent(app.db, agentId, "agent_error", "high").catch(() => {});
-              } else if (payload.runtimeState === "blocked" && shouldNotify(agentId, "agent_blocked")) {
+              } else if (payload.runtimeState === "blocked") {
                 notificationService.notifyAgentEvent(app.db, agentId, "agent_blocked", "medium").catch(() => {});
               }
             } else if (type === "session:event") {
@@ -822,11 +807,12 @@ export function clientWsRoutes(notifier: Notifier, instanceId: string) {
 
               const payload = sessionCompletionMessageSchema.parse(msg);
 
-              if (shouldNotify(agentId, `session_completed:${payload.chatId}`)) {
-                notificationService
-                  .notifyAgentEvent(app.db, agentId, "session_completed", "low", payload.chatId)
-                  .catch(() => {});
-              }
+              // Dedup happens at the DB layer via the default
+              // chat:{chatId}:session_completed key — a noisy reconcile loop
+              // can't produce N rows for the same chat completion.
+              notificationService
+                .notifyAgentEvent(app.db, agentId, "session_completed", "low", { chatId: payload.chatId })
+                .catch(() => {});
             } else if (type === "inbox:ack") {
               const payloadResult = inboxAckFrameSchema.safeParse(msg);
               if (!payloadResult.success) {
--- a/packages/server/src/api/orgs/notifications.ts
+++ b/packages/server/src/api/orgs/notifications.ts
@@ -12,6 +12,12 @@ export async function orgNotificationRoutes(app: FastifyInstance): Promise<void>
     return notificationService.listNotifications(app.db, scope.organizationId, scope.memberId, query);
   });
 
+  app.get<{ Params: { orgId: string } }>("/unread-count", async (request) => {
+    const scope = await requireOrgMembership(request, app.db);
+    const count = await notificationService.unreadCount(app.db, scope.organizationId, scope.memberId);
+    return { count };
+  });
+
   app.post<{ Params: { orgId: string; id: string } }>("/:id/read", async (request) => {
     const scope = await requireOrgMembership(request, app.db);
     const result = await notificationService.markRead(app.db, request.params.id, scope.organizationId, scope.memberId);
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -74,7 +74,7 @@ import {
   rootLogger,
 } from "./observability/index.js";
 import { type AdapterManager, createAdapterManager } from "./services/adapter-manager.js";
-import { broadcastToAdmins } from "./services/admin-broadcast.js";
+import { broadcastToAdmins, registerCrossInstanceBroadcaster } from "./services/admin-broadcast.js";
 import { expiryToSeconds } from "./services/auth.js";
 import { type BackgroundTasks, createBackgroundTasks } from "./services/background-tasks.js";
 import { registerChatMessageDispatcher } from "./services/chat-projection.js";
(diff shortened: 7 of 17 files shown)


yuezengwu:
## Self-review

### 目标与核心改动

**目标**：把 notification infra 从"内存状态 + 轮询兜底"换成"DB-driven + push-only"，让其在多实例部署、进程重启、burst 场景下都给出确定行为。

**核心改动**：
- 跨实例 fanout：`broadcastToAdmins` 内存 fanout → `pg_notify('admin_broadcast_envelopes')` + 每实例 LISTEN 回环
- DB 级 dedup：进程内 `shouldNotify` Map → `(organization_id, dedup_key) WHERE read=false` 部分唯一索引 + `ON CONFLICT DO NOTHING`
- 可见性下推：listNotifications 不再 in-Node 后过滤 + 4× overscan，改成 SQL 谓词 `agent_id IS NULL OR agent_id IN visible`
- 真 COUNT：新增 `/unread-count`，bell badge 准确显示真实未读数
- Push-only：Web bell 删 10s 轮询，靠 admin WS 推送驱动；`useAdminWs()` 挪到 Layout 保证 bell 所到之处必有连接
- 顺手删 `agent_connected` / `agent_needs_decision` 两个 dead type

### 数据库改动 — 需要 @yuezengwu 把关

迁移 `0041_notifications_dedup_key.sql`：
- `ALTER TABLE notifications ADD COLUMN dedup_key text;`
- `CREATE UNIQUE INDEX uq_notifications_org_dedup_unread ON notifications (organization_id, dedup_key) WHERE read = false AND dedup_key IS NOT NULL;`

纯加列 + 加索引，无回填，无 NOT NULL，老 producer 不传 dedup_key 保留原行为。**部分唯一索引在 PG 里的 ON CONFLICT 行为需要确认（见下面 self-review #1）**。

### Self-review 发现并已修复（commit \`3ec26b9\`）

**1. BLOCKER — ON CONFLICT 未带 WHERE 谓词**：原 commit \`d56746c\` 用的是 \`onConflictDoNothing({ target: [orgId, dedupKey] })\`，但 PG 对**部分唯一索引**做 ON CONFLICT 时必须把索引的 WHERE 谓词原样写出，否则 planner 会拒绝匹配并抛 \"there is no unique or exclusion constraint matching the ON CONFLICT specification\"。生产里第一条带 dedup_key 的 notification 就会爆。\`3ec26b9\` 已加上 \`where: sql\`\${notifications.read} = false AND \${notifications.dedupKey} IS NOT NULL\`\`。Drizzle 0.44 的 \`onConflictDoNothing({ target, where })\` 支持这个语义（types 验过）。

### Self-review 已知遗漏 / 留待后续

**N1 — \`useAdminWs\` 不监听 orgId 变化**（pre-existing）。Layout 挂载时读 localStorage 拿 orgId 建 WS，之后用户切组织（重写 localStorage）不会重连。本 PR 没动这块。修法是给 \`use-admin-ws.ts\` 加一个 storage event listener 或者改成从 React context 读 orgId。规模独立，建议单开。

**N2 — 没加 dedup_key / 跨实例 fanout 单测**。\`pulse-aggregator.test.ts\` 改了 mock 让 typecheck 过，但没新增覆盖 \`ON CONFLICT\` 行为或 NOTIFY→LISTEN 回环的测试。验证依赖手工 + 集成测试。

**N3 — 迁移手写**。\`drizzle-kit generate\` 卡在仓库已有的 snapshot collision（meta 里 0016/0018 同 parent），手写 SQL + 加 journal entry 以匹配最近 0038-0040 的实际做法。CLAUDE.md 写的是 \"never hand-edit migrations\"，建议另起一个清理 PR 把 meta 折叠掉。

**N4 — \`broadcastAdminsCrossInstance\` vs \`broadcastToAdmins\` 双 API**。决定让 producer 显式选择是否跨实例（pulseAggregator 故意走 per-instance）。如果未来发现没人需要 per-instance 语义，可以合并成单 API。

### 检查结果

- \`pnpm typecheck\` ✓
- \`pnpm check\` (biome) ✓
- \`pnpm test\` 跳过（testcontainers 需 Docker，本地环境无）

**不 self-approve**：DB schema 改动 + 跨实例语义变更，按 CLAUDE.md 规则需要 human 把关。等你过一遍。

baixiaohang (changes requested):
Direction aligns with our tree spec (`agent-hub/deployment.md`, `NODE.md` — "PG LISTEN/NOTIFY for cross-instance, no Redis/MQ"). The five-commit split is clean and individually reviewable. One blocking bug below, plus a handful of should-fix / nits.

---

### 🔴 Blocking — commit 3: `ON CONFLICT` does not match the partial unique index

`packages/server/src/services/notification.ts:65`

```ts
.onConflictDoNothing({ target: [notifications.organizationId, notifications.dedupKey] })
```

drizzle compiles this to `ON CONFLICT (organization_id, dedup_key) DO NOTHING` **without** a `WHERE`. The schema's unique index is partial (`WHERE read = false AND dedup_key IS NOT NULL`), and PostgreSQL requires the predicate to be supplied for ON CONFLICT inference to match a partial index.

Reproduced locally against PG 16:

```sql
CREATE TABLE notif_test (id text PRIMARY KEY, organization_id text NOT NULL,
                         dedup_key text, read boolean NOT NULL DEFAULT false);
CREATE UNIQUE INDEX uq ON notif_test (organization_id, dedup_key)
  WHERE read = false AND dedup_key IS NOT NULL;

-- 1st insert: succeeds (no conflict yet)
INSERT INTO notif_test VALUES ('a','org1','agent:abc:error',false)
  ON CONFLICT (organization_id, dedup_key) DO NOTHING;

-- 2nd insert: should be suppressed, instead errors:
INSERT INTO notif_test VALUES ('b','org1','agent:abc:error',false)
  ON CONFLICT (organization_id, dedup_key) DO NOTHING;
-- ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification

-- With the predicate supplied: works correctly (INSERT 0 0).
```

Impact: after commit 3, `notifyAgentEvent` opts every caller into a default `dedup_key`. The upstream `try/catch` in `notifyAgentEvent` swallows the PG error, so **the second and every subsequent `agent_error` / `agent_blocked` / `session_completed` event for a given (agent, type) silently disappears** until the user marks the prior row read. That is the opposite of what the PR intends.

Fix — either (preferred, explicit about which index is being used):

```ts
.onConflictDoNothing({
  target: [notifications.organizationId, notifications.dedupKey],
  targetWhere: sql`read = false AND dedup_key IS NOT NULL`,
})
```

or fall back to an untargeted clause that matches any unique constraint:

```ts
.onConflictDoNothing()
```

The "Floods of identical `agent_error` events collapse to a single unread row" test-plan item should be exercised against a real Postgres before merge — typecheck won't catch this class of bug, and the symptom is silent.

---

### 🟡 Should-fix

1. **`notifier.ts` `notifyAdminBroadcast` — comment contradicts behavior** (commit 2)
   The 7500-byte size check throws "so a buggy producer surfaces loudly", but the surrounding `try/catch` swallows the throw alongside legitimate runtime errors. Either log the failure (`logger.error(...)`) before swallowing, or rewrite the comment to match the actual fire-and-forget semantics.

2. **Push-only bell can stale across WS disconnects** (commit 5)
   `app.tsx:27` sets `refetchOnWindowFocus: false` globally, and commit 5 drops the 10s polling. If a notification fires while the laptop is asleep / network is partitioned, the reconnect path never re-fetches: the bell's unread-count stays stale until the *next* inbound push triggers an invalidate. Suggest invalidating `["notifications"]` in `use-admin-ws.ts` on `onopen` (and ideally on `onclose → reconnect` success) so reconnects implicitly catch up.

3. **In-repo doc drift**
   `docs/chat-first-workspace-product-design.md:633-636` still lists `agent_disconnected` and `agent_connected` in the notification model. This PR plus #348 has removed both — the doc should be updated in the same change to keep the spec and code in sync. (The cross-repo tree spec in `first-tree-context/agent-hub/product-direction.md` will need a follow-up too; flagging separately.)

---

### ⚪ Nits

4. **`useAdminWs()` in `Layout` is redundant** (commit 5)
   `app.tsx:65` already mounts `<PulseProvider>` around `<Layout/>`, and `PulseProvider` calls `useAdminWs()` itself. The hook is a refcounted module-level singleton, so adding a second call is harmless — it just bumps `refCount`. But the commit message claim "previously only the workspace page opened the connection" is inaccurate; PulseProvider already covered every authenticated route. Either remove the extra call (rely on PulseProvider) or update the commit message to reflect "additional refcount guard."

5. **`count(*)::int` in `unreadCount`** (commit 4)
   `count(*)` returns `bigint`. Casting to `int` overflows past ~2.1B rows. `::bigint` + JS `Number()` is safer; impact is theoretical, not blocking.

6. **No tests for the new behavior**
   No integration coverage for `unreadCount`, the dedup partial-index path, or cross-instance broadcast. The blocking issue above is exactly the failure mode a dedup integration test would have caught. Given the author flagged Docker as unavailable locally, please run `pnpm test --filter @first-tree-hub/server` in an environment that can spin up testcontainers before merging.

---

### ✅ What reads well

- `buildVisibilityCondition` extracted as one shared SQL fragment for list / markAllRead / unreadCount — eliminates the drift class that the prior code was vulnerable to.
- Pushing the visibility filter into SQL also closes the latent 400-row silent-truncation bug. Nice incidental fix.
- `admin_broadcast_envelopes` reuses the existing `Notifier` channel shape (matches `chat_message_events`, `inbox`) rather than inventing a new transport.
- Comments throughout explain *why* (two failure modes of the in-memory dedup Map, partial-index rationale, single-statement markAllRead) instead of restating the diff.
- Five atomic commits, each independently revertable.

---

### Conclusion

Requesting changes on issue #1 (silent drop of duplicate notifications under the default `dedup_key`). Once that's fixed and exercised against a real Postgres, the rest is straightforward.

yuezengwu:
Thanks for the thorough review @baixiaohang. Point-by-point disposition; new fixes in `9a7424a`.

## 🔴 Blocking — ON CONFLICT WHERE

Already landed in `3ec26b9` before your review submitted (the review's `commit.oid` is exactly that commit, so I think you wrote the review against `d56746c` and submitted against the fixed tip — same diagnosis, same fix). Drizzle 0.44's `onConflictDoNothing({ target, where })` compiles to `ON CONFLICT (cols) WHERE <expr> DO NOTHING` — verified by reading `dialect.cjs`:

```js
const whereSql = config.where ? import_sql.sql` where ${config.where}` : void 0;
this.config.onConflict = import_sql.sql`(${import_sql.sql.raw(targetColumn)})${whereSql} do nothing`;
```

So `where:` is the correct property for DO NOTHING (`targetWhere`/`setWhere` are the DO UPDATE variants). Current code:

```ts
.onConflictDoNothing({
  target: [notifications.organizationId, notifications.dedupKey],
  where: sql`${notifications.read} = false AND ${notifications.dedupKey} IS NOT NULL`,
})
```

Will follow up on the dedup integration test in a CI-Docker env — I can't spin up testcontainers locally to verify, so I held off on the test rather than commit blind.

## 🟡 Should-fix

**1. notifyAdminBroadcast logging** — fixed. Oversize payload now logs `console.error` and returns; transient PG errors stay best-effort. Comment rewritten to match. `notifier.ts` is intentionally logger-free (so it stays portable) but oversize is the one case where silent loss would be operationally invisible, so `console.error` is justified.

**2. Push-only bell can stale across WS disconnects** — fixed. `use-admin-ws.ts` now invalidates `["notifications"]`, `["activity"]`, `["sessions"]`, and `["me","chats"]` on every `socket.onopen` (covers both initial open and reconnects). After sleep/wake or network partition the bell resyncs without waiting for the next push.

**3. In-repo doc drift** — fixed. `docs/chat-first-workspace-product-design.md` no longer lists `agent_disconnected` / `agent_connected`. Tree-spec follow-up acknowledged — will open a separate `first-tree-context` PR.

## ⚪ Nits

**4. Redundant `useAdminWs()` in Layout** — fixed. You're right, `PulseProvider` wraps Layout at `app.tsx:65-67` and calls `useAdminWs()` itself. Reverted the Layout edit; `notification-bell.tsx` comment updated to point at PulseProvider as the connection owner.

**5. `count(*)::int` overflow** — fixed. Changed to `sql<string>\`count(*)\`` + `Number()`. Theoretical at our scale but free to fix.

**6. No tests for new behavior** — acknowledged. Adding dedup integration coverage requires the testcontainers Docker stack, which I don't have in this env. Will queue a follow-up PR once I'm on an environment that can run `pnpm test --filter @first-tree-hub/server`, or pair on it.

## What I'd like you to look at again

- `9a7424a` for the disposition above
- Whether the reconnect-time invalidation in `use-admin-ws.ts` overreaches — I invalidated four root keys to cover all push-driven queries, but if you'd rather scope to `["notifications"]` only (since the immediate motivation is the bell), I'll narrow it

Re-requesting review when you have a moment.

baixiaohang (approved):
Thanks for the quick turnaround. All review items addressed end-to-end.

**Verified locally:**
- `3ec26b9` — `ON CONFLICT (organization_id, dedup_key) WHERE read = false AND dedup_key IS NOT NULL DO NOTHING` reproduced against PG 16: duplicate-while-unread suppresses (`INSERT 0 0`), mark-read reopens the window, NULL-`dedup_key` rows keep legacy always-insert. The blocker is closed.
- `9a7424a` — oversize broadcast now `console.error`s before the early-return (no longer swallowed); reconnect catch-up invalidates `["notifications"]` / `["activity"]` / `["sessions"]` / `["me","chats"]` on every `onopen` (latestQc set-before-connect order via two adjacent `useEffect`s is safe — first fires before the WS handshake completes); Layout revert correctly defers to PulseProvider's existing `useAdminWs()`; `count(*)` returns string, `Number()`-parsed (safe up to 2^53); doc list synced with the schema change.

**Approving with two non-code follow-ups, neither blocks merge:**

1. **Test plan items still need to be exercised on real Postgres + multi-instance.** typecheck / biome / no-Docker-vitest don't cover the dedup partial-index path or the cross-instance fanout — please run the `pnpm test --filter @first-tree-hub/server` integration suite in a Docker-enabled environment, plus a manual two-instance smoke + bell-badge / unread > 100 spot-check, before squash-merging. The dedup blocker we just fixed is exactly the class of bug a green typecheck can hide.

2. **Tree spec drift to address in a follow-up PR.** `first-tree-context/agent-hub/product-direction.md:101` still lists "agent connect/disconnect" in the M1 notification events; #348 + #351 between them have removed both types. This PR's scope ends at the hub repo, but please open a `first-tree-context` PR alongside the merge so the spec catches up with the implementation. (We picked up the in-repo `chat-first-workspace-product-design.md` here — the cross-repo equivalent is the only loose end.)

PR #359, opened 2026-05-13
fix: notification dead-clicks, fault-row redundancy, completion spam

yuezengwu:
## Summary

Three compounding issues on the notification bell, all triggered by a real prod audit of 100 recent notifications on dev hub:

- **Dead clicks (blocker)** — 44/100 rows carried agentId but no chatId. Bell rendered them clickable, navigated to `/?a=<agentId>`, and WorkspacePage immediately stripped `?a=` back to `/`. Half the bell clicks did nothing.
- **Fault-row redundancy** — `agent_error` / `agent_blocked` / `agent_stale` each used `agent:{id}:{type}` as dedup key, so one sick agent could surface as three concurrent unread rows.
- **`session_completed` spam** — 56% of rows. Duplicates what the conversation list already shows (latest message). Crowded out higher-priority faults out of the 8-item bell popover.

## What changed

- **Bell click** (`notification-bell.tsx`): `clickable={!!n.chatId}` (consistent with the workspace right-rail NotificationList); `navigate('/?c=' + chatId)` — no more `?a=` strip round-trip. Workspace `NotificationList` preserves `?engagement=` / other params instead of clobbering them.
- **Dedup** (`services/notification.ts`): default dedup key for `notifyAgentEvent` is now `agent:{agentId}:fault`. All three fault types collapse onto one unread row per agent.
- **Auto-resolve** (`services/notification.ts` + `api/agent/ws-client.ts`): new `markAgentFaultsResolved(db, agentId)` closes every unread row for the agent. Wired into `presenceService.bindAgent` (offline → online) and the `runtime:state` handler when state flips to `idle` / `working`. Recovery now clears the badge instead of needing manual ack.
- **Drop `session_completed`** end-to-end: shared schema enum, `sessionCompletionMessageSchema`, server WS frame branch, server producer, client connection method, agent-slot wiring, handler API (`reportSessionCompletion` removed from `SessionContext`), call sites in `claude-code` + `codex` handlers, plus 8 client test files. Web UI label removed.
- **Drop `session_error`** (long-dead type — schema + label + composeMessage entry but no producer anywhere).

## Test plan

- [x] 3 new cases in `packages/server/src/__tests__/notification-fault-dedup.test.ts` cover: fault collapse (R1), reopen-after-read, per-agent scope of `markAgentFaultsResolved`.
- [x] `pnpm check` (biome) — 678 files clean
- [x] `pnpm --filter @first-tree-hub/server test` — 115 files, 902 tests pass
- [x] `pnpm --filter @first-tree-hub/client test` — 46 files, 345 tests pass
- [x] `pnpm --filter @first-tree-hub/web test` — 11 files, 91 tests pass

## DB cleanup (manual, no migration)

Run by hand on each environment after deploy — these rows are no longer producible but still pollute the bell:

```sql
-- See what's there
SELECT type, read, COUNT(*) FROM notifications
WHERE type IN ('agent_disconnected', 'session_completed', 'session_error')
GROUP BY type, read ORDER BY type, read;

-- Clean up
DELETE FROM notifications
WHERE type IN ('agent_disconnected', 'session_completed', 'session_error');

-- Optional: drop residual unread rows that still use the old per-type
-- dedup_key (so they don't co-exist with new agent:{id}:fault rows
-- for the same agent during the transition window).
DELETE FROM notifications
WHERE read = false
  AND (dedup_key LIKE 'agent:%:agent_error'
    OR dedup_key LIKE 'agent:%:agent_blocked'
    OR dedup_key LIKE 'agent:%:agent_stale');
```

## Known limitation

`markAgentFaultsResolved` UPDATEs the DB but does not push an admin WS event for read-state changes. The bell refetches on the next push or reconnect, so the badge may lag the actual state by up to one push cycle. Adding a dedicated `notification:read` envelope was deferred — it adds a new event shape on both sides for a sub-second cosmetic difference.


baixiaohang (commented):
Reviewed end-to-end. Direction looks right — `WorkspacePage`'s chat-first redirect already makes the old `?a=` payload a real dead-link, and the fault-collapse + completion-spam removal both line up with the audit data in the PR description. The cross-package cleanup (shared schema, WS frame, producer, agent-slot, handler API, 8 client tests, web label) is consistent and leaves no dangling references.

A few items worth a second pass before this lands.

### ⚠️ Severity gets locked by whichever fault fires first

`createNotification` uses `ON CONFLICT DO NOTHING` (`packages/server/src/services/notification.ts:68`), so once the partial unique index has an unread row for `agent:{id}:fault`, every subsequent fault for the same agent is dropped on the floor — including its severity.

In practice:

1. `agent_stale` (medium) fires first → row persisted as `medium`.
2. `agent_error` (high) fires a moment later → suppressed entirely.
3. Bell badge / severity sort stays on `medium` until the user acks the row, even though the agent has since degraded.

The dedup-collapse goal ("one row per sick agent") and "show the most serious state" are in tension here. Two ways to square them:

- `ON CONFLICT DO UPDATE SET severity = GREATEST(severity, EXCLUDED.severity), type = EXCLUDED.type, message = EXCLUDED.message` — escalates severity, the latest fault wins the type/copy.
- Or accept the trade-off and document it explicitly in the `notifyAgentEvent` JSDoc so the next reader isn't surprised.

The new test's `expect([...]).toContain(open[0]?.type)` is honest about "first wins, we don't care which type" — that's the bit that nudges me to think we *should* care about severity even if we don't care about type.

### ⚠️ `markAgentFaultsResolved` SQL is broader than the comment claims

```ts
await db
  .update(notifications)
  .set({ read: true })
  .where(and(eq(notifications.agentId, agentId), eq(notifications.read, false)));
```

The JSDoc says "every unread *fault-scoped* notification", but the WHERE clears every unread row for that agent regardless of type or dedup key. Today that's harmless because this same PR removes `session_completed` / `session_error`, so the only remaining agent-scoped types *are* faults. But the next time someone adds a non-fault agent-scoped notification (system message, per-agent reminder, etc.) and an agent rebinds, that row will silently flip to read.

Cheap fix: filter on `inArray(notifications.type, ["agent_error", "agent_blocked", "agent_stale"])` or `dedupKey = 'agent:{agentId}:fault'` so the code matches the comment.

### ❓ The cleanup SQL could just be a migration

```sql
DELETE FROM notifications
WHERE type IN ('agent_disconnected', 'session_completed', 'session_error');
```

`notifications.type` is a `text` column, not a PG enum, so this is safe to run inside a Drizzle migration — `0042_notifications_drop_legacy_types.sql` — and every environment converges automatically on deploy. Hand-running it per env is a footgun: easy to miss one, easy to leave a half-cleaned env until someone notices weeks later. The "optional: drop residual unread rows with old per-type dedup_key" block is the same story.

### ❓ Bell fault rows have no click target anywhere

By design fault notifications don't carry a `chatId` (they're agent-state events, not session events), and after this PR the bell renders them non-clickable. So a user who sees "Agent X is blocked" / "Computer Y is unresponsive" in the bell has nowhere to click through to.

Not a blocker — but it's worth deciding whether the right target is `/agents/{agentId}` (the admin manage page is already linked from `agent-context.tsx:199`) or some kind of in-workspace agent-detail view. If the intended answer is "the bell is just a heads-up; triage happens on /agents", that's fine — I just want to make sure it's a decision and not an accident.

### ✅ What looks solid

- Auto-resolve on `bindAgent` (offline → online) + healthy `runtime:state` transitions is a much cleaner recovery model than the previous "the badge sticks forever" path.
- The partial unique index `where read = false AND dedup_key IS NOT NULL` is exactly what makes the reopen-after-read contract land cleanly; the new test pins it down.
- End-to-end deletion of `session_completed` / `session_error` is consistent across shared / server / client / web / tests.
- `.catch(() => {})` on every notification call from `ws-client.ts` keeps the bookkeeping fire-and-forget — presence / state callers won't fail on notification trouble.

### Known limitation

The "badge may lag by up to one push cycle" trade-off (no dedicated `notification:read` envelope) reads as reasonable — a new event shape on both sides for a sub-second visual difference isn't worth the maintenance surface. Glad it's called out explicitly.

---

Not blocking from my side, but I'd like the severity-lock question answered before approval (either fix it or document it as intentional). The other three are easier judgment calls.

yuezengwu:
Thanks for the review @baixiaohang — all four points addressed in `b942df3`. Short walkthrough:

**#1 Severity lock** → fixed in code, not docs.

`createNotification` switched from `ON CONFLICT DO NOTHING` to `ON CONFLICT DO UPDATE`:

```ts
set: {
  severity: sql`CASE
    WHEN ${notifications.severity} = 'high' OR excluded.severity = 'high' THEN 'high'
    WHEN ${notifications.severity} = 'medium' OR excluded.severity = 'medium' THEN 'medium'
    ELSE 'low'
  END`,
  type: sql`excluded.type`,
  message: sql`excluded.message`,
}
```

`createdAt` deliberately *not* touched so bell ordering still tracks "when this incident opened" rather than "when the latest observation arrived" — keeps an old, escalated incident from jumping over a fresh one. New test pins down: stale=medium → error=high escalates; later blocked=medium does not downgrade.

**#2 Over-broad WHERE** → scoped to fault types.

`markAgentFaultsResolved` now adds `inArray(notifications.type, [agent_error, agent_blocked, agent_stale])`. New test simulates a hypothetical future `agent_reminder` row via raw INSERT and asserts mark-resolved leaves it alone — so the contract is enforced for the first PR that adds a non-fault agent-scoped type.

**#3 Cleanup as migration** → done.

`0042_notifications_drop_legacy_types.sql` deletes legacy-type rows + unread residuals on old per-type dedup keys. `notifications.type` is text not an enum, so this is a pure data migration. Every env converges on deploy.

**#4 Fault rows had no click target** → bell now routes to `/agents/{agentId}`.

Two-tier navigation:

```ts
if (n.chatId)        navigate(`/?c=${n.chatId}`)        // session events
else if (n.agentId)  navigate(`/agents/${n.agentId}`)   // fault events
```

`clickable` predicate widens to `!!n.chatId || !!n.agentId`. The `/agents/:uuid` route already exists (`AgentDetailPage`), and this matches the implicit "admin manage page" mental model — triage happens on the agent detail view.

CI:
- biome / typecheck clean
- server vitest: 116 files, 910 tests (was 902 — 4 new fault-dedup cases on top of the original 3, now 5 total in `notification-fault-dedup.test.ts` and a few add-ups elsewhere)
- client 345 / web 91 unchanged

baixiaohang (approved):
All four points from the earlier review are addressed in `b942df3` — verified end-to-end.

- **#1 Severity lock** — `ON CONFLICT DO UPDATE` with a `CASE WHEN` GREATEST(severity), latest event wins type+message, `createdAt` preserved so bell ordering still tracks incident-open time. New `escalates severity monotonically` test pins it down.
- **#2 Over-broad WHERE** — `markAgentFaultsResolved` now scoped via `inArray(notifications.type, [AGENT_ERROR, AGENT_BLOCKED, AGENT_STALE])`. The forward-compat test that inserts a hypothetical `agent_reminder` row directly is the right shape — catches the first PR that adds a non-fault agent-scoped type without re-deriving the allow-list.
- **#3 Cleanup as migration** — `0042_notifications_drop_legacy_types.sql` lands the legacy-type delete + the unread-residual cleanup on old per-type dedup keys. Every env converges on deploy.
- **#4 Fault rows had no click target** — bell now does two-tier navigation (`chatId` → `/?c=…`, else `agentId` → `/agents/{uuid}`), `clickable` predicate widened accordingly. `/agents/:uuid` / `AgentDetailPage` already exist, so this reuses the admin manage page rather than adding a new surface.

LGTM.

PR #360, opened 2026-05-13
Fix agent config draft baseline after save

Gandy2025:
## Summary
- keep the agent config draft baseline in sync after a successful save
- reload the latest remote config through React Query before resetting the local draft
- fix the SaveBar Model dirty-chip jump target to the Setup section
- add focused coverage for clean draft list item initialization

## Why
Saving or reloading could leave the page comparing against a stale baseline, which made freshly saved config still appear dirty or caused reload to reset against the wrong config timing.

## Validation


baixiaohang (approved):
LGTM — the baseline/cache decoupling is the right call.

## Why this fix is correct
- Treating `cfg?.payload.*` as the dirty baseline made dirty state observable through React Query side effects (refetch, `setQueryData`). Decoupling `baseline` into hook-local state restores the hook's stated intent ("ignore mid-edit refetches; trust the caller to drop the draft").
- `reloadRemote` switching from `invalidateQueries + resetAll` to `fetchQuery + resetToConfig` removes the time window where the useEffect could re-seed against a still-stale `cfg` before the refetch resolved.
- Save `onSuccess` calling `resetToConfig(next)` covers the case where the server normalizes the payload (trim, env placeholder, defaults) — without it, the just-saved values would still read as dirty.
- `buildPayloadPatch` deps moving from `[draft, cfg, ...]` to `[draft, baseline, ...]` matches the new ownership boundary.
- SaveBar Model jump target is correct: `agent-cfg-model` doesn't exist as a DOM id; Setup is where the Model control actually lives, and this matches the `dirtyAnchors` mapping at `agent-detail.tsx:329`.

Verified locally: `pnpm --filter @first-tree-hub/web test` (12 files / 92 tests pass) and `pnpm --filter @first-tree-hub/web typecheck` clean.

## Non-blocking suggestions (feel free to defer)

1. **Test coverage** — the new test only covers initial seed. The actual regression this PR fixes is "save succeeds, server returns normalized payload, draft must read as clean". Worth adding a test that exercises `resetToConfig` with a config whose payload differs from the prior baseline and asserts `summary.anyDirty === false`.

2. **Section→anchor mapping duplicated** — `agent-detail.tsx:761` has `section === "model" ? SECTION_ANCHORS.setup : sectionAnchorId(section)`, and `agent-detail.tsx:328-331` has the same model→setup / mcp→tools / env|git→advanced mapping inline. Could be extracted into a single `sectionToAnchor(section)` helper so the two stay in sync if more sections shift later.

3. **`reloadRemote` callback dep on whole `draft`** — pre-existing, not introduced here, but while we're in this code: depending on the whole `draft` object means `reloadRemote` is rebuilt every render. Could be narrowed to `[queryClient, uuid, draft.resetToConfig]` since `resetToConfig` is the only field actually used and it's a stable `useCallback` reference.

Approving — none of the above block the fix.

baixiaohang (approved):
Re-approving after `fc4338f` (chore: centralize agent config anchors) — both follow-ups from the prior review are addressed:

- ✅ `sectionToAnchor(section)` extracted at `agent-detail.tsx:70-75`, used by both `dirtyAnchors` and SaveBar `onJumpTo`. Mapping now lives in one place.
- ✅ `reloadRemote` deps narrowed to `[queryClient, uuid, resetDraftToConfig]` via the local stable reference; the callback no longer rebuilds every render.

Re-verified locally: `pnpm --filter @first-tree-hub/web test` (13 files / 104 tests pass) and `pnpm --filter @first-tree-hub/web typecheck` clean.

## One small nit (non-blocking)

`sectionToAnchor` uses an `if`-chain with `return SECTION_ANCHORS.prompt` as the fallback. If `DraftSectionName` ever gains a new variant, the new value silently falls into the prompt anchor instead of failing loudly. An exhaustive `switch` would let TypeScript surface the gap:

```ts
function sectionToAnchor(section: DraftSectionName): string {
  switch (section) {
    case "prompt": return SECTION_ANCHORS.prompt;
    case "model":  return SECTION_ANCHORS.setup;
    case "mcp":    return SECTION_ANCHORS.tools;
    case "env":
    case "git":    return SECTION_ANCHORS.advanced;
  }
}
```

## Reiterating one earlier suggestion (still non-blocking)

The new test only covers initial seed. The actual regression this PR fixes — "save succeeds, server returns a normalized payload, draft must immediately read as clean" — is the exact path I'd want a regression test for. Worth a follow-up if not done in this PR.

Ship it.

PR #361, opened 2026-05-13
Gate agent config actions by manage access

Gandy2025:
## Summary
- hide runtime config, SaveBar, Test, Manage bindings, and Danger Zone for users who can only view a shared agent
- gate config and client-list queries to manager/admin access while preserving visible session status
- require agents to be suspended before delete and surface lifecycle errors inline
- add a small access helper with focused tests

## Why
The backend treats config, test, lifecycle, and identity updates as manage-only routes. The page previously exposed controls to visible non-managers that would fail after click, and active agents showed a delete action that the server rejects.

## Validation


baixiaohang (approved):
LGTM — client-side gating mirrors `requireAgentAccess(..., "manage")` in `packages/server/src/scope/require-resource.ts` exactly (`isAdmin || managerId === memberId`), and the `suspended`-before-delete guard lines up with `services/agent.ts` raising `BadRequestError` on active deletes. Inline lifecycle errors are a nice quality bump too.

Two optional follow-ups, not blocking this PR:

1. **`boundClient` label regresses for visible-only viewers.** `allClientsQuery` is now gated by `canEditConfig`, so a non-manager opening an org-visible non-human agent gets `boundClientLabel = boundClientId` (raw `clnt_…` uuid) in both the ContextBar and the Profile "Runs on …" row instead of the hostname. The underlying constraint is pre-existing — `/orgs/:orgId/clients` is admin-only — but the regression is more visible after this change. Could be handled later by either skipping the lookup entirely for visible-only viewers (show "—" / hide the row) or exposing a narrower per-client lookup that visible viewers can hit.

2. **`tileValues.model = "restricted"`** reads a bit out of register vs. the other short, factual tile values. An em-dash placeholder (matching `sessions`/`runtime` fallbacks) might be more in line with the rest of the page.

Happy to land as-is.

PR #363, opened 2026-05-13
Preserve agent env and MCP edits

Gandy2025:
## Summary
- only allow empty sensitive env edits to keep an existing persisted redacted value
- preserve unsaved sensitive plaintext and reject empty new sensitive values
- switch MCP stdio args to JSON array input so spaces and quotes round-trip
- add HTTP/SSE headers JSON input so headers are not dropped on edit
- omit MCP health badges until a real health source is wired

## Why
The agent config page could turn new sensitive env values into placeholders and silently lose MCP args/header structure. These are data integrity issues in the draft editor before save.

## Validation

### DIFF
--- a/packages/web/src/pages/agent-detail/env-section.tsx
+++ b/packages/web/src/pages/agent-detail/env-section.tsx
@@ -109,6 +109,10 @@ export function EnvSection(props: EnvSectionProps) {
           open={!!dialog}
           onOpenChange={(open) => !open && setDialog(null)}
           initial={dialog.mode === "edit" ? dialog.initial : null}
+          allowKeepExisting={
+            dialog.mode === "edit" &&
+            canKeepExistingSensitiveValue(dialog.initial, props.items.find((item) => item.key === dialog.key)?.status)
+          }
           forbiddenKeys={props.otherKeys(dialog.mode === "edit" ? dialog.key : null)}
           onSubmit={(value) => {
             if (dialog.mode === "edit") props.onUpdate(dialog.key, value);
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
LGTM — 修复了三个真实的草稿编辑器数据完整性问题：

1. 新增敏感 env 空值不再被静默转成 redacted 占位符；只有「已持久化 + 当前值就是占位符」的条目允许空提交保留旧密文。
2. MCP stdio args 改成 JSON 数组输入，带空格/引号的参数可以正确往返。
3. HTTP/SSE 之前完全没有 headers 编辑入口，编辑一次就丢；这次补上输入框。

校验逻辑抽成纯函数 + 单元测试，scope 控制得不错。

一点小观察（非 blocker）：`describeMcp` 现在用 `JSON.stringify(arg)` 渲染每个 stdio arg，普通无空格 arg 会多一层引号（例如 `node "-e" "console.log()"`）。展示上略冗余但更明确，可接受。

baixiaohang (approved):
LGTM — 修复了三个真实的草稿编辑器数据完整性问题：

1. 新增敏感 env 空值不再被静默转成 redacted 占位符；只有「已持久化 + 当前值就是占位符」的条目允许空提交保留旧密文。
2. MCP stdio args 改成 JSON 数组输入，带空格/引号的参数可以正确往返。
3. HTTP/SSE 之前完全没有 headers 编辑入口，编辑一次就丢；这次补上输入框。

校验逻辑抽成纯函数 + 单元测试，scope 控制得不错。

一点小观察（非 blocker）：`describeMcp` 现在用 `JSON.stringify(arg)` 渲染每个 stdio arg，普通无空格 arg 会多一层引号（例如 `node "-e" "console.log()"`）。展示上略冗余但更明确，可接受。

packages/web/src/pages/agent-detail/mcp-section.tsx
```diff
@@ -104,12 +104,56 @@ export function McpSection(props: McpSectionProps) {
 
 function describeMcp(m: McpServer): string {
   if (m.transport === "stdio") {
-    const args = m.args?.length ? ` ${m.args.join(" ")}` : "";
+    const args = m.args?.length ? ` ${m.args.map((arg) => JSON.stringify(arg)).join(" ")}` : "";
```
baixiaohang: Nit: `JSON.stringify(arg)` quote-wraps every arg in the inline preview, so a plain `--port 3000` server now renders as `"--port" "3000"`. Round-trippability is great for the textarea input, but for this human-readable preview it's a small regression in legibility. Consider only quoting args that actually need it (contain whitespace or quote chars), e.g.:

```ts
const needsQuote = (s: string) => /[\s"']/.test(s);
const args = m.args?.length
  ? ` ${m.args.map((a) => (needsQuote(a) ? JSON.stringify(a) : a)).join(" ")}`
  : "";
```

Not blocking.

packages/web/src/pages/agent-detail/mcp-section.tsx
```diff
@@ -217,27 +282,39 @@ function McpDialog({ open, onOpenChange, initial, forbiddenNames, onSubmit }: Mc
...
-                <Label htmlFor="mcp-args">Args (space-separated, optional)</Label>
-                <Input
+                <Label htmlFor="mcp-args">Args (JSON array, optional)</Label>
+                <textarea
                   id="mcp-args"
                   value={argsText}
                   onChange={(e) => setArgsText(e.target.value)}
-                  placeholder="--port 3000"
-                  className="font-mono"
+                  placeholder={'["--port", "3000"]'}
```
baixiaohang, line 295: Nit: the input syntax for `Args` just got stricter (space-separated → JSON array). The placeholder hints at the shape, but a one-line helper text under the label (similar to other dialogs in this page) would save users from a confused submit on first try — e.g. `Each arg as a separate JSON string. Use this when an arg contains spaces or quotes.`

Same applies to the new Headers textarea. Not blocking.

packages/web/src/pages/agent-detail/env-section.tsx
```diff
@@ -125,23 +129,51 @@ type EnvDialogProps = {
...
   forbiddenKeys: ReadonlySet<string>;
   onSubmit: (value: EnvEntry) => void;
 };
 
-function EnvDialog({ open, onOpenChange, initial, forbiddenKeys, onSubmit }: EnvDialogProps) {
+export function canKeepExistingSensitiveValue(
+  initial: EnvEntry,
+  status: DraftListItem<EnvEntry>["status"] | undefined,
+) {
+  return initial.sensitive && initial.value === ENV_REDACTED_PLACEHOLDER && status !== undefined && status !== "added";
```
baixiaohang, line 141: Question, not a blocker: `canKeepExistingSensitiveValue` depends on the read-side contract that the server returns persisted sensitive env values as exactly `ENV_REDACTED_PLACEHOLDER` (`***`). This contract existed pre-PR (the old `if (sensitive && initial?.sensitive && !value)` branch implicitly relied on it too) — but since this PR now puts that condition on the critical path of "don't accidentally wipe a saved secret", could you confirm the GET endpoint (and any cache/projection in between) does indeed mask sensitive values to `***` and never returns the raw ciphertext or an empty string? If there's already a test or shared helper enforcing that on the server side, a pointer would be enough.

baixiaohang (commented):
Overall direction looks good — solid fixes for real data-integrity bugs in the draft editor (sensitive env new-add lossy submit, unsaved sensitive plaintext getting wiped, stdio args losing spaces/quotes on round-trip, http/sse headers having no UI). The refactor into pure, testable helpers (`canKeepExistingSensitiveValue`, `envDialogInitialValue`, `resolveEnvDialogValue`, `formatMcpArgsInput`/`parseMcpArgsText`, `formatMcpHeadersInput`/`parseMcpHeadersText`) with Vitest coverage is clean. Dropping the placeholder "unknown" tool-health badge is also more honest than pretending we have a signal.

A few non-blocking comments inline — feel free to address or wave off.

PR #364, opened 2026-05-13
Constrain agent git repo local paths

Gandy2025:
## Summary
- validate configured git repo localPath values in the shared runtime config schema
- reject absolute paths, Windows drive paths, dot segments, empty segments, backslashes, whitespace, and control characters
- resolve local paths at client runtime and ensure clone/additionalDirectories targets remain inside the session workspace
- add shared and client tests for schema validation and runtime containment

## Why
Agent runtime git repos are cloned into a session workspace. A user-provided localPath must not be able to escape that workspace or point Codex additionalDirectories outside the intended root.

## Validation
- `pnpm --filter @agent-team-foundation/first-tree-hub-shared test -- agent-runtime-config`
- `pnpm exec vitest run src/__tests__/git-local-path.test.ts src/__tests__/codex-thread-options.test.ts` from `packages/client`
- `pnpm check`
- `pnpm typecheck`

## Note
This touches shared/client runtime behavior. I did not bump the command package version in this PR.


yuezengwu (approved):
LGTM — focused security hardening with good defense-in-depth.

**Scope**
- Schema-layer: `gitRepoSchema.localPath` rejects absolute paths, Windows drive letters, `.`/`..` segments, empty segments, backslashes, whitespace, control chars.
- Runtime-layer: new `resolveGitRepoTargetPath` re-validates after `path.resolve` and confirms the target stays inside the workspace via `relative()`. Used at all three call sites (claude-code clone, codex clone, codex `additionalDirectories`).

**Tests**
Solid coverage on both layers — schema accept/reject matrix in `shared`, runtime resolution + `buildCodexThreadOptions` rejection in `client`. The runtime test deliberately bypasses the schema, exercising the defense-in-depth case.

**Non-blocking observations**
- `deriveRepoLocalPath` (the fallback when `localPath` is omitted) is not separately validated, but `resolveGitRepoTargetPath` catches any unsafe derivations (e.g. a crafted URL yielding `..`). Defense-in-depth holds.
- Pre-existing DB rows with non-compliant `localPath` values would now fail schema parse. Likely fine given `localPath` is recent and *should* reject, but worth being aware of if any old rows exist.

No data-structure or DB changes. Approving.

baixiaohang (approved):
LGTM. Defense-in-depth is well-structured: schema-layer rejection of unsafe `localPath` plus runtime-layer re-validation via `path.resolve` + `relative()` containment, applied at all three call sites (claude-code clone, codex clone, codex `additionalDirectories`). Test coverage spans both layers and deliberately exercises the bypass-the-schema case.

Same observations as @yuezengwu — `deriveRepoLocalPath` fallback isn't separately validated but the runtime check catches it; pre-existing non-compliant rows would now fail schema parse, acceptable given the field is recent.

One non-blocking nit: in `resolveGitRepoTargetPath`, `relativeTarget.startsWith("..")` will also reject legitimate directories whose name happens to start with `..` (e.g. `..foo`). Tightening to `relativeTarget === ".." || relativeTarget.startsWith(".." + path.sep)` would be strictly correct, though the current form errs on the safe side and the edge case is rare. Feel free to defer.

baixiaohang (approved):
Re-approving on 108d9c2. The escape check now uses `relativeTarget === ".." || relativeTarget.startsWith(`..${sep}`)` — exactly resolves the false-positive nit, and the new `..safe` positive test pins the intended behavior. LGTM.

PR #367, opened 2026-05-13
feat(web): chat-row avatar with working ring and unread badge

yuezengwu:
## Summary

Adds the chat-first conversation list its avatar slot, wiring up \`workingAgentIds\` (shipped by #366) and consolidating the unread indicator onto the avatar overlay.

Implements **PR C** from the chat-status-icons spec — the visual companion to PR B's data layer. **No DB / schema changes.**

## What each row gets

| Element | Behavior |
|---|---|
| **Avatar** | Single initial for \`direct\`, split-disc composite for \`group\` (vertical bisection for 2, T-split for 3, 2×2 + \"+N\" overflow for ≥4). 36×36 default fits the 320-wide rail. |
| **Working ring** | Breathing \`--state-working\` border. **Direct chats only**, when the peer's agent_id is in \`workingAgentIds\`. Group rings deferred to a future per-chat data source — \`agent_presence.runtime_state\` is agent-global so showing a ring per group would be a false-positive factory (Option A trade-off, see spec §⑦.② / §⑦.⑦). |
| **Unread badge** | \`--state-error\` corner badge with numeric count, \`99+\` rollover. Replaces the legacy small text-row dot. z-index 3, hairline-bold \`--bg-raised\` border so it cuts cleanly through the working ring when both render. |

## Realtime

\`use-admin-ws.ts\` gains a \`me/chats\` invalidation on \`session:state\` frames (already org-scoped, payload carries \`chatId\`). Without this the working ring would lag the 15s polling refetch when a runtime transitions without sending a new message.

## Motion

New \`chat-row-avatar-working-breathe\` keyframe — 1.6s opacity-only (GPU-cheap on long lists). \`prefers-reduced-motion\` collapses it to a static solid ring. No transform or layout-affecting animation.

## Changes

- \`packages/web/src/components/chat/chat-row-avatar.tsx\` — new component (single + composite + ring + badge).
- \`packages/web/src/pages/workspace/conversations/index.tsx\` — row layout: avatar on the left, title/subtitle stack on the right. Legacy 6-unit text-row dot removed (its semantics now live on the avatar's overlay).
- \`packages/web/src/hooks/use-admin-ws.ts\` — invalidate \`[\"me\", \"chats\"]\` on \`session:state\`.
- \`packages/web/src/index.css\` — keyframes + \`prefers-reduced-motion\` fallback.

## Out of scope (separate PRs if wanted)

[...]

PR #366, opened 2026-05-13
feat(server): expose workingAgentIds on MeChatRow

yuezengwu:
## Summary

Adds a server-derived \`workingAgentIds: string[]\` field to every \`MeChatRow\` returned by \`/me/chats\`, populated from each speaker's GLOBAL \`agent_presence.runtime_state === 'working'\`. Powers the chat-first conversation-list working ring in the upcoming web refactor.

**Zero DB migration. Zero new index.**

## Decision context

The chat-list spec needs a per-chat \"agent is working now\" signal. Precise per-chat tracking would require a partial index on \`agent_chat_sessions(chat_id) WHERE state = 'active'\` — a schema change we're avoiding. **Option A** trades precision for zero DDL:

- Field is derived from the **global** presence table at query time. \`agent_presence.agent_id\` is the PK so the JOIN is a single-row lookup per speaker.
- An agent working in any chat appears as \"working\" on every chat they speak in.
- The web layer compensates: it **only renders the working ring for \`type === 'direct'\`** so the cross-chat false positive stays invisible (1-on-1 user-with-agent is the common case; group-chat working signals defer to a future per-chat data source).

Full spec: \`chat-status-icons-preview.html\` §⑦.② / §⑦.⑦ / §⑦.⑨ (Option A).

## Changes

| File | Change |
|---|---|
| \`packages/shared/src/schemas/me-chat.ts\` | Additive \`workingAgentIds: z.array(z.string())\` on \`meChatRowSchema\`. Old clients ignore. |
| \`packages/server/src/services/me-chat.ts\` | Participants second query: add \`leftJoin(agentPresence)\`; collect \"working\" speakers into a per-chat map; thread into row assembly. Main query untouched. |
| \`packages/server/src/__tests__/me-chat-working-agents.test.ts\` | 6 new invariants (see Test plan). |

## What is **not** changed (deliberate)

- **No schema migration, no new index** — \`agent_presence\` PK on \`agent_id\` is sufficient.
- **No main-query change** — the LEFT JOIN sits on the existing participants second query that already runs once per page.
- **No version bump for \`packages/command\`** — \`MeChatRow\` is server/web-only; \`command\`/\`client\` don't import it.
- **Frontend not touched here** — wiring + ring rendering arrive in the follow-up web PR.

## Test plan

- [x] \`pnpm check\` + \`pnpm typecheck\` clean
- [x] \`pnpm --filter @first-tree-hub/server test\` — **116 files / 912 tests** pass
[...]

PR #368, opened 2026-05-13
feat(web): per-agent avatar hue palette + DEV preview route

yuezengwu:
## Summary

Conversation-list avatars previously all painted with the brand accent — a vertical strip of identical green discs that made groups lose their per-member distinction and made 1-on-1 rows hard to scan. Discovered during the e2e visual review of #367. This PR introduces a hashed-per-agent hue palette so each avatar reads as that agent's identity.

Also lands a DEV-only \`/preview/chat-row-avatar\` route that mounts every documented variant in a single grid — useful for future visual review iterations.

**Zero schema. Zero API change.** All deltas are inside \`packages/web\`.

## Visual delta

| Before | After |
|---|---|
| All avatars \`var(--accent)\` green; groups indistinguishable | 8-hue OkLCH palette, stable per agent UUID; group composites carry real per-member color |
| Initials \`var(--bg-raised)\` → dark text on color in dark theme (poor contrast against several palette entries) | Initials fixed at \`oklch(0.985 0 0)\` near-white, passes WCAG AA against every hue in both themes |

Screenshots are in the local e2e review HTML (\`chat-list-e2e-review.html\` in the parent workspace, including v1 vs v2 side-by-side).

## Changes

| File | Change |
|---|---|
| \`packages/web/src/components/chat/chat-row-avatar.tsx\` | New \`AVATAR_HUES\` (8 OkLCH entries) + \`pickAvatarHue(seed)\` djb2-style hash. \`SingleAvatar\` / \`Seg\` consume the new bg; text colour hardened to near-white. Same hardening on \`UnreadBadge\` text. |
| \`packages/web/src/components/chat/__tests__/chat-row-avatar.test.ts\` | 4 new invariants for \`pickAvatarHue\` (OkLCH output, determinism, hue diversity across distinct seeds, empty-seed fallback). |
| \`packages/web/src/pages/chat-row-avatar-preview.tsx\` | DEV-only route. Renders \`ChatRowAvatar\` against 12 fixtures (1-on-1 idle/working/unread variants, watcher, group n=2/3/4/5+/10+). Mirrors the \`/preview/context\` pattern; gated by \`import.meta.env.DEV\`. Supports \`?theme=light|dark\` and \`?freeze=1\` for screenshot capture. |
| \`packages/web/src/app.tsx\` | Wires the preview route into the router (same lazy DEV-only pattern as ContextPreviewPage). |

## What is **not** changed (deliberate)

- **No schema migration, no API change.** Hue is derived client-side from the agent UUID already present in \`MeChatRow.participants\`.
- **No version bump** for \`packages/command\` — web-only.
- **Offline grayscale / error ring not implemented in this PR.** The original spec described both, but the data they need (\`agent_presence.client_id IS NULL\`, \`runtime_state === 'error'\`) isn't on \`MeChatRow\`. Adding them is a follow-up that mirrors the workingAgentIds approach (#366). Flagged transparently in the e2e review notes.

## Test plan

- [x] \`pnpm check\` clean
- [x] \`pnpm typecheck\` clean (including \`lint:tokens\`)
- [x] \`pnpm --filter @first-tree-hub/web test\` — **12 files / 107 tests** pass, including 4 new pickAvatarHue invariants
- [x] e2e dev verification: docker postgres + dev server + SQL-seeded 6 agents × 7 chats. Real \`ConversationList\` renders distinct hues per agent; v1 vs v2 comparison captured in local review HTML.

### DIFF
--- a/packages/web/src/app.tsx
+++ b/packages/web/src/app.tsx
@@ -33,6 +33,12 @@ const ContextPreviewPage = import.meta.env.DEV
   ? lazy(() => import("./pages/context-preview.js").then((module) => ({ default: module.ContextPreviewPage })))
   : null;
 
+const ChatRowAvatarPreviewPage = import.meta.env.DEV
+  ? lazy(() =>
+      import("./pages/chat-row-avatar-preview.js").then((module) => ({ default: module.ChatRowAvatarPreviewPage })),
+    )
+  : null;
+
 export function App() {
   return (
     <QueryClientProvider client={queryClient}>
@@ -52,6 +58,16 @@ export function App() {
                   element={
                     <Suspense fallback={null}>
                       <ContextPreviewPage />
+                    </Suspense>
+                  }
+                />
+              ) : null}
+              {ChatRowAvatarPreviewPage ? (
+                <Route
+                  path="/preview/chat-row-avatar"
+                  element={
+                    <Suspense fallback={null}>
+                      <ChatRowAvatarPreviewPage />
                     </Suspense>
                   }
                 />
--- a/packages/web/src/components/chat/__tests__/chat-row-avatar.test.ts
+++ b/packages/web/src/components/chat/__tests__/chat-row-avatar.test.ts
@@ -55,9 +55,11 @@ describe("formatUnreadLabel — badge text", () => {
 });
 
 describe("pickAvatarHue — deterministic per-agent fill color", () => {
-  it("returns an OkLCH string from the palette", () => {
-    const hue = pickAvatarHue("agent-1");
-    expect(hue).toMatch(/^oklch\(/);
+  it("returns a `var(--avatar-hue-N)` token reference", () => {
+    // Pins the contract that the helper hands back a CSS-token
+    // reference rather than a raw `oklch(...)` literal — index.css is
+    // the single source of palette truth.
+    expect(pickAvatarHue("agent-1")).toMatch(/^var\(--avatar-hue-[0-7]\)$/);
   });
 
   it("same seed yields the same hue across calls (stable per agent)", () => {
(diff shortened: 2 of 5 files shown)


baixiaohang (commented):
Looks solid overall — derivation from `agentId` (no schema change) is the right call, the new tests pin the determinism contract, and the preview route is a nice operational handle for future visual reviews. A small set of polish suggestions, no blockers:

**Token system (substantive)**
- `chat-row-avatar.tsx:183, 246, 312`: `oklch(0.985 0 0)` is hard-coded three times. The exact value already lives in `index.css` as `--color-destructive-foreground`. Worth lifting it into a dedicated semantic token (e.g. `--color-fg-on-vivid` or similar) and consuming it everywhere — that's the principle `scripts/check-design-tokens.sh` is guarding ("index.css is the single source of truth"). Same case for the 8 entries of `AVATAR_HUES`: they're more discoverable + themeable in `index.css` as `--avatar-hue-0…7`. The lint doesn't block raw `oklch()` today, so this is suggestion, not a gate.

**Comments that drift from the code**
- `chat-row-avatar.tsx:42-49` (the new `AVATAR_HUES` block comment): claims white-ish text is "handled by the rendering site via `var(--bg-raised)` in light mode" — but the rendering sites in this same diff switched from `var(--bg-raised)` to the hard-coded `oklch(0.985 0 0)`. The block comment is now stale.
- `chat-row-avatar.tsx:54` (`// green (kept first so the default-fallback hash collision still looks neutral)`): the first entry is only consumed when `seed.length === 0`. For any non-empty seed the modulo can land on any index — there's no special "fallback collision" concept. Either drop the parenthetical or rephrase to "first entry is the empty-seed fallback".

**Test invariant strength**
- `chat-row-avatar.test.ts:75-87` ("different seeds reach more than one palette entry"): `toBeGreaterThan(1)` is very loose — a regression that collapses 8 carefully-chosen seeds onto just 2 hues would still pass. Tightening to `toBeGreaterThanOrEqual(4)` (or `=== 8` for these specific seeds) catches hash regressions earlier.

**Preview side-effect**
- `chat-row-avatar-preview.tsx` (`ChatRowAvatarPreviewPage`, top of the function): `document.documentElement.classList.toggle("dark", themeOverride === "dark")` runs during render, not in `useEffect`. Works for a static screenshot route, but it's a side effect during render — moving it to a small `useEffect` keeps it tidy and avoids surprises under Strict Mode / re-render.

Holding on approval until baixiaohang has had a look.

yuezengwu:
感谢 review 🙏。5 条都吃了，rebase 未做（main 同步）。

### 1. Lift hue palette + on-vivid text into index.css ✅
- 新 \`--avatar-hue-0..7\` 8 个 token，<code>chat-row-avatar.tsx</code> 现在只持有 *selection* 逻辑（\`pickAvatarHue\` 返回 \`var(--avatar-hue-N)\` 字符串引用），调色板本身完全归 index.css 管
- 新 \`--fg-on-vivid: oklch(0.985 0 0)\`，<code>SingleAvatar</code> / <code>Seg</code> / <code>UnreadBadge</code> 三处文字色统一指它
- 符合 \`scripts/check-design-tokens.sh\` 的"index.css 是单一真相"原则

### 2. Stale block comment ✅
原 docblock 写"白色文字由 rendering site 经 var(--bg-raised) 处理"，改这次 PR 已经切走了。重写为指 \`--fg-on-vivid\` token。

### 3. "fallback collision" parenthetical ✅
按你指正改为 \"also the empty-seed fallback\" —— 非空 seed 时 modulo 可以命中任何 bucket，原措辞会误导。

### 4. Test invariant tightened ✅
\`toBeGreaterThan(1)\` → \`toBeGreaterThanOrEqual(4)\`，**同时把测试 seed 从 agent-slug 字串换成 UUIDv7 形状**——agent-slug 共享 "agent-" 前缀，djb2 在前 6 个 char 上 state 几乎相同，对 mod 8 是反向最差案例（实测只命中 3 hue）。UUIDv7 高熵尾部更贴合生产数据，命中 4+ hue。一个改进帮你抓两类回归：哈希塌缩 + 现实数据被错误代理。

### 5. Preview side effect ✅
\`document.documentElement.classList.toggle\` 移进 \`useEffect(..., [themeOverride])\`。Strict Mode 双调用安全 + 不会在潜在 SSR 路径执行 DOM 操作。

### 质量门
- \`pnpm check\` ✓
- \`pnpm typecheck\` ✓（含 lint:tokens）
- \`pnpm --filter web test\` → 12 files / 107 tests 全绿

baixiaohang (approved):
LGTM — token system cleanup landed cleanly, the UUIDv7 test seeds are a nice touch beyond what was suggested, and the `useEffect` rewrite on the preview route is solid. CI green. Thanks for the fast turnaround.

PR #381, opened 2026-05-14
feat(server): auto-archive chats when bound PR is merged

### LINKED ISSUE #379: Chat 在相关 PR 合并后自动归档
## Feature Request

**描述**: 希望在 chat 的状态管理中增加自动归档功能。当 chat 中关联的 PR 都已合并时，系统自动将该 chat 标记为归档状态，避免已完成的 chat 混在列表里。具体的 PR 检测逻辑可进一步讨论。

### 自动收集的上下文

| 字段 | 值 |
|------|-----|
| 来源 | web-plugin |
| 页面 | https://dev.cloud.first-tree.ai/?c=9f6a6c21-cf32-4448-bc3f-b79c411a2cad |
| 浏览器 | Chrome 147.0.0.0 / macOS 10.15.7 |
| 时间 | 2026-05-14T02:38:57.239Z |

<details>
<summary>Failed Network Requests</summary>

GET /api/v1/orgs/019dd8ff-64e9-72bb-b057-fdf60475b3eb/notifications?limit=8 TypeError: Failed to fetch
GET /api/v1/orgs/019dd8ff-64e9-72bb-b057-fdf60475b3eb/notifications/unread-count TypeError: Failed to fetch
GET /api/v1/orgs/019dd8ff-64e9-72bb-b057-fdf60475b3eb/chats?filter=all&engagement=active TypeError: Failed to fetch
GET /api/v1/agents/019dfc9f-d275-7985-b591-38ec9e317f91/sessions/9f6a6c21-cf32-4448-bc3f-b79c411a2cad/events?limit=200&direction=desc → 401
GET /api/v1/chats/9f6a6c21-cf32-4448-bc3f-b79c411a2cad/messages?limit=50 → 401
POST /feedback/submit → 500

</details>

yuezengwu:
Closes #379.

## Summary
- New webhook bypass: `pull_request.closed` with `merged===true` flips every chat bound to that PR from `active` → `archived` for the owning humans.
- Runs independently of the normalize / audience / deliver pipeline. Normalize still drops `pull_request.closed` in Stage 1, so this never produces an inbox message.
- No new tables, no GitHub API calls, no schema changes.

## Design choice — accept fluctuation
Multi-PR chats can be temporarily archived while siblings are still open. Any later sibling activity (review / comment / synchronize / etc.) goes through the existing normalize+deliver path and the existing `chat-projection` revives `archived → active`. We trade perfect timing for zero local state.

## Per-user safety
Writes use an UPSERT guarded by `setWhere = engagement_status = 'active'`, so:
- `active` → flipped to `archived`
- implicit-active (no row) → INSERTed as `archived`
- already-`archived` → no-op (idempotent)
- user-manually `deleted` → preserved

## Bypass placement
Sits BEFORE `claimEvent` on purpose — the service is itself idempotent, and we want the archive to still run on GitHub retries even if the normalize path was already claimed by a previous attempt.

## Test plan
- [x] Unit: 8 cases in `github-archive-on-merge.test.ts` covering single/multi human, multi-chat, deleted preservation, idempotency, non-PR entity skip, org scoping.
- [x] Integration: 2 cases in `github-app-webhook-route.test.ts` — `merged=true` archives and emits no message; `merged=false` does not archive.
- [x] Existing `pull_request.closed → null` normalize test still passes (no normalize behavior change).
- [x] Local mock-trigger smoke (real HMAC-signed webhook) shows: BEFORE = implicit active → 200 `{handled:false}` → AFTER = archived; negative `merged=false` case writes no row.
- [x] `pnpm check && pnpm typecheck && pnpm test` all green (952 tests).

## Versioning
Server-only change — no `packages/command` bump per `docs/versioning-and-publishing.md`.


baixiaohang (changes requested):
Nice bypass design overall — sitting before `claimEvent` (idempotent + survives normalize-claim races on retries), the `setWhere = engagement_status = 'active'` guard preserving user-`deleted` state, and the `chat-projection.ts:126-131` auto-revive that makes the "accept fluctuation" tradeoff viable, all check out. Test coverage (8 unit + 2 integration) is thorough — org-scoping, deleted-preservation, non-PR-entity-skip, and the merged=false negative are exactly the cases I'd want to see. Requesting changes on **one concrete defect**; the design call is fine.

## Blocker — NUL byte in `services/github-archive-on-merge.ts:78`

> Sidenote: I tried to leave this as an inline comment on the file but GitHub returned `422 Line could not be resolved` — which is the same problem manifesting from the other side: GitHub has no per-line view of this file because it's flagged binary, so inline review is impossible until this is fixed.

The new file `packages/server/src/services/github-archive-on-merge.ts` contains a **literal NUL (`\u0000` / `\x00`) byte** at byte offset 3096, sitting between `chatId` and `humanAgentId` in the dedup key on what is line 78:

```ts
const key = `${row.chatId}\u0000${row.humanAgentId}`;
```

It renders as a space in most editors but it is a real null byte (verified by reading the raw file bytes — position 3096 is `0x00`, the rest of the file is plain UTF-8).

### Why it matters

A single NUL anywhere in a file makes git treat that file as binary. The diff for this file therefore currently displays on GitHub as:

```
Binary files /dev/null and b/packages/server/src/services/github-archive-on-merge.ts differ
```

The entire 102-line service implementation is **invisible** in the GitHub diff view, and any future edits to this file will be equally unreviewable until the NUL is removed.

### Suggested fix

Use any non-hex separator (UUIDs only contain hex + `-`, so `|`, `:`, `\n`, etc. are all collision-free):

```ts
const key = `${row.chatId}|${row.humanAgentId}`;
```

Or, if you'd rather not pick a magic char:

```ts
const key = JSON.stringify([row.chatId, row.humanAgentId]);
```

Once that single character is changed, the diff will become normal text and review can proceed.

## FYI (non-blocking)

- Issue #379 says "当 chat 中关联的 PR **都**已合并时" but explicitly notes "具体的 PR 检测逻辑可进一步讨论", so the chosen "first-merge archive + auto-revive on sibling activity" is within the spirit of the request — just flagging so we're aligned on the user-visible behavior (multi-PR chats with quiet siblings will sit in `Archived` until something stirs them).
- `archiveChatsForMergedPr` runs the per-target UPSERTs sequentially. Fine for realistic <10-row PRs; not worth changing here.

yuezengwu:
@baixiaohang Thanks for the careful review and especially for catching the NUL byte — that explains why the file looked oddly empty in the GitHub diff.

Fixed in 78bb6e6: replaced `\x00` with `|` (UUIDs are hex + dash so the pipe is collision-free). Verified with `file` — the path now reports `UTF-8 text` instead of `data`, and `python3 -c "open(p,'rb').read().count(b'\\x00')"` returns 0. The diff for this file should now render normally.

Re-ran `pnpm check` + the 8 unit + 2 integration tests post-fix — all green.

On the non-blocking FYI about multi-PR chats with quiet siblings sitting in `Archived`: confirmed and intended. The original requester explicitly accepted this fluctuation when we discussed the design — that's how the "no PR state, no GitHub API call" simplification became viable. Will keep an eye on user reports if it turns out to bite in practice.

baixiaohang (approved):
Fix verified — `services/github-archive-on-merge.ts` is now plain UTF-8 text with 0 NUL bytes (line 78: `${row.chatId}|${row.humanAgentId}`), GitHub renders the diff normally, and nothing else changed in the second push (rebase only). All previous ✅ items still hold. Approving on behalf of @baixiaohang.

PR #384, opened 2026-05-14
feat(web): redesign context tab around live freshness

Gandy2025:
## Summary
- Redesigns the Context tab into a live/fresh-first first view: live hero, seven-day top-updated-domain map, and two-line aggregate usage summary.
- Adds Context Tree usage telemetry at the runtime turn level for Codex and Claude Code handlers, plus server aggregation into the context snapshot API.
- Extends shared schemas and Web API types for snapshot usage data, hides usage telemetry from the session timeline, and updates preview mock data.
- Bumps the published command package to 0.14.0 because this adds a new SessionEventKind and client/shared runtime behavior changed.

## Context Tree
- Decision doc PR: https://github.com/agent-team-foundation/first-tree-context/pull/269

## Validation
- pnpm typecheck
- pnpm test
- pnpm exec biome check <changed source files>
- git diff --check

## Notes
- pnpm check still fails on existing unrelated .claude/worktrees nested Biome root configuration errors.


baixiaohang (commented):
Reviewed against the design doc (first-tree-context#269) and the diff. Overall the redesign matches the Live/Fresh-first decision faithfully. A few points worth a look before this lands:

### UI

1. **Stale / error status loses surface detail.** Previously `ContextStatus` rendered `contextStatus.label` + `contextStatus.detail` whenever severity wasn't ok. The new `LiveContextHero` only encodes severity via the dot color (and `aria-label`). When `snapshotStatus !== "unavailable"` but `contextStatus.severity` is `warning`/`error` (e.g. `stale`), users diagnosing freshness issues lose the explanation. Worth considering surfacing `contextStatus.detail` under the timestamp when severity ≠ `ok`. — `packages/web/src/pages/context.tsx:LiveContextHero`

2. **Singular grammar in the signal.** `In the last {snapshot.usage.windowDays} days` reads "In the last 1 days" when `windowDays = 1`. `agentText`/`usageText` already handle singular; `windowDays` should too. — `packages/web/src/pages/context.tsx:ContextSignal`

3. **`changeBreakdownParts` zero-fallback is dead under live flow** (`changeGroups` only emits nodes with a non-null `changeType`, so all-zero counts can't reach it). Safe to drop the fallback or replace it with a stricter type to make the invariant explicit. — `packages/web/src/pages/context.tsx`

### Telemetry

4. **`"3d"` is added to the snapshot window enum but the Web client hard-codes `"7d"`.** Intentional for future filtering / other API consumers, or leftover from iteration? If the product surface is locked to 7d per the design doc, the test churn (`7d → 3d` in `org-settings.test.ts`) feels incidental — happy to keep it but worth a one-line rationale.

5. **Index hint.** `summarizeContextTreeUsage` filters by `(kind, createdAt)` over `session_events` joined to `agents`. No index added in this PR — should be fine at current volume, but flagging in case event throughput grows; might want a composite index on `(kind, created_at)` (or `(agent_id, kind, created_at)`) later.

6. **Per-turn semantics look correct.** `emitContextTreeUsage` fires once per inbound user message in both handlers, guarded by `contextTreePath`. Matches the "a turn had a Context Tree binding and used context" definition in `agent-hub/context-tab.md`. Good that it's hidden from the session timeline via `filterEventsForTimeline`.

### Misc

7. **PR description says the package is bumped to `0.12.11`**, but `packages/command/package.json` actually bumps `0.13.0 → 0.13.1`. Description bug — please fix so release notes don't pick up the wrong version.

8. **Patch bump for a new `SessionEventKind`** — backward-compat-wise, a new-client/old-server combo would have its `context_tree_usage` events rejected by Zod. Probably fine if server+client deploy together, but a `0.14.0` minor bump would be more honest semver. Not blocking.

### Tests

`summarizeContextTreeUsage` test coverage is solid (filters by kind, excludes other-org events via the inner join, excludes orphan agent IDs).

—

Verdict: comments only, no blockers. The stale-status visibility regression (#1) is the one I'd most want to confirm is intentional. Once that and the trivia (PR body version, singular grammar) are addressed I'm comfortable.

Gandy2025:
Addressed the review feedback in ff39a75:

- Surfaced stale/error status detail under the timestamp when `contextStatus.severity !== "ok"`, so stale/sync diagnostics are visible again instead of only encoded by dot color.
- Fixed the usage signal grammar for the 1-day case (`1 day`, while agent/time singulars were already handled).
- Removed the dead zero-count fallback from `changeBreakdownParts`; cards are produced only from changed nodes, so no all-zero chip should render.
- Removed the incidental `3d` snapshot window from Web/server API types and tests. The supported windows are now `1d / 7d / 30d`, with the first-view UI intentionally hard-coded to `7d`.
- Updated the PR body and bumped the published command package to `0.14.0` to reflect the new `SessionEventKind` more honestly.

I did not add a session_events index in this PR. The current aggregate is low volume and already covered by tests; I agree it is the right follow-up if telemetry volume grows.

Validation rerun locally after these changes:
- `pnpm exec biome check` on the changed files
- `pnpm typecheck`
- `pnpm test`
- `git diff --check`

yuezengwu (approved):
Reviewed the fix commit (ff39a75) — all of baixiaohang's actionable feedback is addressed:

- stale/error status detail restored under the LiveContextHero timestamp
- usage signal handles `1 day` singular
- dead zero-count fallback in `changeBreakdownParts` removed
- `3d` window removed from API/types/tests; supported windows are `1d / 7d / 30d` with first-view locked to 7d
- PR body and package version corrected to 0.14.0 to honestly reflect the new `SessionEventKind`

Skipping the `session_events(kind, created_at)` composite index is fine for now; agree it's the right follow-up if usage telemetry volume grows.

LGTM.

PR #398, opened 2026-05-14
feat(web): add "Mark as unread" action to chat list row menu

yuezengwu:
## Summary

Adds a **Mark as unread** item to the chat list row "…" kebab menu so users can flag a chat as unread after reading it (e.g. "I'll come back to this later") — it then shows up in the **Unread** filter and renders with the unread row style.

Implemented as a UI affordance, not a read-cursor rewind:

- **Bumps `chat_user_state.unread_mention_count` to `GREATEST(current, 1)`** — never reduces a pre-existing higher count, so concurrent mention increments are preserved.
- **Leaves `last_read_at` alone** — avoids regressing anything that consumes the read cursor.
- **Idempotent** at every layer: UPSERT on the server, `disabled` on the menu item when the chat is already unread.

### Change set

- `shared`: `meChatUnreadResponseSchema` + `MeChatUnreadResponse`.
- `server/services/me-chat.ts`: `markMeChatUnread(db, chatId, humanAgentId)`.
- `server/api/chats.ts`: `POST /chats/:chatId/unread`, behind the same `requireChatAccess` gate as `/read`.
- `web/api/me-chats.ts`: `markMeChatUnread(chatId)` client.
- `web/pages/workspace/conversations/row-engagement-menu.tsx`: new menu item (placed first under both `ACTIVE` and `ARCHIVED` engagement states); `disabled` when the row already has `hasUnread`.
- `web/pages/workspace/conversations/index.tsx`: passes `hasUnread` through to the menu.

Cross-tab realtime sync is out of scope — same as existing `markRead`.

## Test plan

- [x] `pnpm check` (biome) clean.
- [x] `pnpm typecheck` — all 9 tasks green.
- [x] Server test suite — 980 passed (24 in `me-chat-service.test.ts`, +3 new):
  - `markMeChatUnread bumps a read chat back to unread without touching last_read_at`
  - `markMeChatUnread is idempotent and never reduces a pre-existing higher count`
  - `markMeChatUnread lazily materialises a chat_user_state row when none exists`
- [ ] Manual UI verification on a local dev hub: open chat list, hover a read row → "…" → click **Mark as unread** → row turns bold + unread filter picks it up. Verify the item is disabled on an already-unread row.

### DIFF
--- a/packages/server/src/services/me-chat.ts
+++ b/packages/server/src/services/me-chat.ts
@@ -800,6 +800,15 @@ export async function markMeChatRead(db: Database, chatId: string, humanAgentId:
  * if the row already has a positive count, it stays as-is. `last_read_at`
  * is intentionally untouched — this is a UI affordance, not a "rewind the
  * read cursor" operation.
+ *
+ * Contract note — semantic overload: the column is named `unread_mention_count`
+ * but is co-opted here as a generic "manual unread" flag. Every existing
+ * consumer (conversation list bold styling, `?filter=unread`, source-counts,
+ * the bell badge) only checks `> 0`, so the exact value carries no meaning
+ * for callers. If a future feature ever renders the literal mention count
+ * (e.g. a "N mentions" pill), it must NOT read this column directly — it
+ * needs a separate mention-only counter, otherwise a manually-marked-unread
+ * chat would show a fictitious "1 mention".
  */
 export async function markMeChatUnread(
   db: Database,
--- a/packages/server/src/services/me-chat.ts
+++ b/packages/server/src/services/me-chat.ts
@@ -34,6 +34,7 @@ import {
   type MeChatReadResponse,
   type MeChatRow,
   type MeChatSourceCounts,
+  type MeChatUnreadResponse,
   type ToolCallEventPayload,
 } from "@agent-team-foundation/first-tree-hub-shared";
 import { and, eq, inArray, type SQL, sql } from "drizzle-orm";
@@ -787,6 +788,45 @@ export async function markMeChatRead(db: Database, chatId: string, humanAgentId:
     });
 
   return { chatId, lastReadAt: now.toISOString(), unreadMentionCount: 0 };
+}
+
+// ---------------------------------------------------------------------------
+// Mark unread
+// ---------------------------------------------------------------------------
+
+/**
+ * Bump `unread_mention_count` to at least 1 so the chat shows up as unread
+ * in the conversation list (and is matched by `?filter=unread`). Idempotent:
+ * if the row already has a positive count, it stays as-is. `last_read_at`
+ * is intentionally untouched — this is a UI affordance, not a "rewind the
+ * read cursor" operation.
+ */
+export async function markMeChatUnread(
+  db: Database,
+  chatId: string,
+  humanAgentId: string,
+): Promise<MeChatUnreadResponse> {
+  await db
+    .insert(chatUserState)
+    .values({
+      chatId,
+      agentId: humanAgentId,
+      unreadMentionCount: 1,
+    })
+    .onConflictDoUpdate({
+      target: [chatUserState.chatId, chatUserState.agentId],
+      set: {
+        unreadMentionCount: sql`GREATEST(${chatUserState.unreadMentionCount}, 1)`,
+      },
+    });
+
+  const [row] = await db
+    .select({ unreadMentionCount: chatUserState.unreadMentionCount })
+    .from(chatUserState)
+    .where(and(eq(chatUserState.chatId, chatId), eq(chatUserState.agentId, humanAgentId)))
+    .limit(1);
+
+  return { chatId, unreadMentionCount: row?.unreadMentionCount ?? 1 };
 }
 
 // ---------------------------------------------------------------------------
--- a/packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
+++ b/packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
@@ -14,22 +14,20 @@ type EngagementActionsArgs = {
 };
 
 function actionsFor({ status, hasUnread, runEngagement, runMarkUnread }: EngagementActionsArgs): RowAction[] {
-  const markUnread: RowAction = {
-    key: "mark-unread",
-    label: "Mark as unread",
-    disabled: hasUnread,
-    onSelect: runMarkUnread,
-  };
   if (status === ACTIVE) {
     return [
-      markUnread,
+      // Mark-as-unread is only offered on ACTIVE rows. ARCHIVED rows intentionally
+      // omit it: re-surfacing an archived chat is the Unarchive action's job, and
+      // a silent "unread but still hidden under Archived" state would diverge from
+      // the existing fan-out path that auto-revives archived → active on a new
+      // message.
+      { key: "mark-unread", label: "Mark as unread", disabled: hasUnread, onSelect: runMarkUnread },
       { key: "archive", label: "Archive", onSelect: () => runEngagement(ARCHIVED) },
       { key: "delete", label: "Delete", destructive: true, onSelect: () => runEngagement(DELETED) },
     ];
   }
   if (status === ARCHIVED) {
     return [
-      markUnread,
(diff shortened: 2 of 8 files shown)


baixiaohang (commented):
Nice clean PR — surgical change set, the data model story (don't rewind `last_read_at`, bump `unread_mention_count` via `GREATEST`) matches what's documented in the `chat_user_state` section of `agent-hub/messaging.md`, and the three new service-layer tests cover the interesting branches (lazy materialise, idempotent under a higher pre-existing count, `last_read_at` preserved).

A few non-blocking thoughts (also left as inline comments):

1. **Semantic overload of `unread_mention_count`** — the column is named "mention count" but `markMeChatUnread` repurposes it as a generic "manual unread flag". Today nothing breaks because the UI / filter / source-counts all only check `> 0`, but if anything ever wants to render the actual mention count (e.g. a "3 mentions" pill), a manually-marked-unread chat would show `1`, which is fictitious. Worth a one-liner in the column comment or the service docstring so future readers don't conflate the two meanings.

2. **`Mark as unread` on `ARCHIVED` chats** — currently the menu item is offered under both `ACTIVE` and `ARCHIVED`. Since the engagement axis is orthogonal to unread, this means an archived chat can be flagged unread but stays in the Archived tab. That's defensible (you wanted to re-surface it *quietly*, not unarchive it), but it's a small asymmetry vs. the existing fan-out path that auto-revives `archived → active` on a new message. Worth confirming this is the intended UX, and maybe a one-line comment in `actionsFor` explaining why the action is present in both branches.

3. **Manual UI verification is still unchecked** — the test plan box is unticked. Worth doing before merge, especially the "row turns bold + Unread filter picks it up" path and the disabled state on already-unread rows. Also worth a quick sanity check that the kebab menu doesn't collapse if you mark unread while the row is still hovered.

4. **`meChatUnreadResponseSchema` is currently unused on the client** — `markUnreadMut` ignores the response. That matches the pattern from `markMeChatRead` (we don't validate the response there either), so this is consistency-with-existing rather than a real issue, just noting it.

No blockers from me — leaving as Comment so you can decide on (1) and (2). Holding off on Approve until those questions are answered + the manual UI check is done.

packages/server/src/services/me-chat.ts
```diff
@@ -789,6 +790,45 @@ export async function markMeChatRead(db: Database, chatId: string, humanAgentId:
...
+    .insert(chatUserState)
+    .values({
+      chatId,
+      agentId: humanAgentId,
+      unreadMentionCount: 1,
+    })
+    .onConflictDoUpdate({
+      target: [chatUserState.chatId, chatUserState.agentId],
+      set: {
+        unreadMentionCount: sql`GREATEST(${chatUserState.unreadMentionCount}, 1)`,
```
baixiaohang, line 828: Worth a comment here noting that we're overloading `unread_mention_count` as a generic "manual unread" flag — the column name still reads as "mention count", so a future reader rendering an actual mention count UI (e.g. a "3 mentions" pill) could be misled when a row was manually marked unread with no real mentions. A one-liner like `// We co-opt unread_mention_count as the unread flag; the UI only checks > 0, so the exact value is not displayed as a mention count anywhere.` would lock in the contract.

packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
```diff
@@ -1,21 +1,37 @@
...
+      markUnread,
+      { key: "archive", label: "Archive", onSelect: () => runEngagement(ARCHIVED) },
+      { key: "delete", label: "Delete", destructive: true, onSelect: () => runEngagement(DELETED) },
     ];
   }
   if (status === ARCHIVED) {
     return [
-      { key: "unarchive", label: "Unarchive", onSelect: () => run(ACTIVE) },
-      { key: "delete", label: "Delete", destructive: true, onSelect: () => run(DELETED) },
+      markUnread,
```
baixiaohang: Question: was it deliberate to offer **Mark as unread** under `ARCHIVED` as well? The fan-out path auto-revives `archived → active` on a new message, but this path leaves engagement alone — so an archived chat can become "unread" without leaving the Archived tab. That's a defensible choice (re-surface quietly without un-archiving), but worth a one-line comment here so future readers know it's intentional vs. a copy-paste from the ACTIVE branch.

packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
```diff
@@ -26,16 +42,34 @@ function actionsFor(status: ChatEngagementStatus, run: (next: ChatEngagementStat
...
+  hasUnread: boolean;
+}) {
   const queryClient = useQueryClient();
-  const mut = useMutation({
+  const invalidate = () => {
+    queryClient.invalidateQueries({ queryKey: ["me", "chats"] });
+    queryClient.invalidateQueries({ queryKey: ["chat-detail", chatId] });
+  };
+  const engagementMut = useMutation({
     mutationFn: (next: ChatEngagementStatus) => patchChatEngagement(chatId, next),
```
baixiaohang, line 58: Tiny nit: `markUnreadMut` discards the response, so the `unreadMentionCount` field on `meChatUnreadResponseSchema` isn't actually consumed by any caller right now. That matches the existing `markMeChatRead` flow (we don't read its response either), so I wouldn't change it — but if you ever decide to slim the contract, returning `204 No Content` like the participants endpoint would be a cleaner shape than "return the count nobody reads".

baixiaohang:
Friendly nudge @yuezengwu — when you get a moment, could you:

1. **Tick the manual UI verification** in the test-plan checklist (row turns bold + Unread filter picks it up + disabled state on already-unread rows). Both `typecheck` and the service tests are clean, but the UI assertions in the checklist haven't been exercised yet and they're the easiest place for a low-level rendering regression to hide.
2. **Confirm the `ARCHIVED`-path intent** (from inline comment): is offering **Mark as unread** under both engagement states deliberate (i.e. quietly re-surface an archived chat without unarchiving it), or is it a copy-paste from the `ACTIVE` branch? A one-line code comment either way would make the asymmetry vs. the fan-out's `archived → active` auto-revive easier to grok later.

Once those two are addressed I think the PR is in good shape — happy to take another pass and recommend approve. No rush.

packages/server/src/services/me-chat.ts
yuezengwu: Added — the docstring on `markMeChatUnread` now spells out that the column is co-opted as a generic unread flag and that any future literal-mention-count UI must use a separate counter rather than reading this column. Fixed in 1ac556a.

packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
yuezengwu: Decided to drop it from ARCHIVED entirely (rather than keep + comment). Reasoning: re-surfacing an archived chat is Unarchive's job; a silent "unread but still hidden under Archived" state would diverge from the fan-out path that auto-revives `archived → active` on a new message. Now only offered on ACTIVE rows, with a comment in `actionsFor` explaining why. Fixed in 1ac556a.

packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx
yuezengwu: Acknowledged — keeping the response body for consistency with `markMeChatRead`. If we ever revisit, we can flip both to 204 in one pass.

baixiaohang (approved):
LGTM — all three review points addressed cleanly in 1ac556a:

- `unread_mention_count` semantic-overload contract is locked in by the new docstring (and the "future literal mention-count UI must use a separate counter" callout is exactly the contract I wanted to pin down).
- Dropping Mark as unread from ARCHIVED entirely (instead of the "keep + comment" middle ground) is the cleaner call — it keeps the engagement axis honest and avoids a "silently unread but still hidden" state that would diverge from the fan-out auto-revive path. The inline comment in `actionsFor` makes the reasoning legible to future readers.
- Acknowledged the response-body nit; happy to leave consistency with `markMeChatRead` as-is.

Approving. Standard ask before merging: tick the manual UI verification checkbox once you've run through it (row bold + Unread filter pick-up + disabled-on-already-unread). No blockers from my side.

PR #405, opened 2026-05-15
Scope Context Tree sync to bound agents

Gandy2025:
## Summary

- Add an agent-scoped Context Tree info route at `/api/v1/agent/context-tree/info`, backed by the authenticated agent's organization.
- Move local Context Tree sync from process startup to `AgentSlot.start()` after the slot binds/registers its agent.
- Isolate local Context Tree checkouts by `repo URL + branch` hash under `context-tree-repos/`.
- Keep the legacy user-scoped `/api/v1/context-tree/info` for compatibility.

## Compatibility / operational notes

- Existing `~/.first-tree/hub/data/context-tree/` clones are no longer used after upgrade. They are safe for operators to delete manually; this PR does not remove them automatically to avoid deleting local work from older installs.
- The sync lock is intentionally in-process. Two separate client processes sharing the same `FIRST_TREE_HUB_HOME` could still race the same hashed checkout, but the supported local-dev isolation model keeps each runtime on a separate data dir. Cross-process file locking can be added later if shared data dirs become a real deployment pattern.
- `syncContextTree()` and `/api/v1/context-tree/info` remain for compatibility with external scripts or older SDK consumers that still use the user/default-org route. The Hub client runtimes now use `syncAgentContextTree()` instead; removing the legacy route/export should be a follow-up once downstream usage is audited.

## Context Tree

- Baseline verify PR: https://github.com/agent-team-foundation/first-tree-context/pull/275
- Runtime binding tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/276

## Verification

- `pnpm --filter @first-tree-hub/client typecheck`
- `pnpm --filter @agent-team-foundation/first-tree-hub typecheck`
- `pnpm --filter @first-tree-hub/server typecheck`
- `pnpm exec biome check packages/client/src/__tests__/bootstrap.test.ts packages/client/src/index.ts packages/client/src/runtime/agent-slot.ts packages/client/src/runtime/bootstrap.ts packages/client/src/runtime/index.ts packages/client/src/runtime/runtime.ts packages/client/src/sdk.ts packages/command/src/__tests__/client-runtime-context-tree.test.ts packages/command/src/core/client-runtime.ts packages/server/src/app.ts packages/server/src/__tests__/agent-context-tree-info.test.ts packages/server/src/api/agent/context-tree-info.ts`
- `pnpm --dir packages/client exec vitest run src/__tests__/bootstrap.test.ts`
- `pnpm --dir packages/command exec vitest run src/__tests__/client-runtime-context-tree.test.ts`

Server route test was added but could not run locally because the server Vitest global setup requires testcontainers and this machine reported `Could not find a working container runtime strategy`.


baixiaohang (commented):
Direction looks right — this fixes a real bug where every agent on a multi-org client got the caller's primary-org tree regardless of which org the agent actually belongs to. The server-side hook chain (`agentSelectorHook` → `identity.organizationId`) makes the new Class D route trivially correct, and the per-`(repo, branch)` checkout dir + in-process lock cleanly isolate concurrent slots on the same org.

A handful of polish items before approval:

### Suggestions

1. **`syncAgentContextTree` is missing a jsdoc.** It's a new public export (`client/src/index.ts`, `runtime/index.ts`) and consumers can't tell from the signature how it differs from `syncContextTree` (agent-scope vs user-primary-org-scope). Please add a comment that pins the semantic split — symmetric with the existing jsdoc on `syncContextTree` at `packages/client/src/runtime/bootstrap.ts:158`.

2. **The `syncContextTree` jsdoc is now slightly stale** — it still says "Sync the *shared* Context Tree git clone", but post-refactor each (repo, branch) gets its own hashed dir under `context-tree-repos/`. Tweak the wording while you're updating point 1.

3. **Old `~/.first-tree/hub/data/context-tree/` directory is orphaned on upgrade.** Every existing install will leave a stale clone behind after this lands. Non-blocking, but consider either (a) a one-shot best-effort `rmSync` on first run inside `resolveContextTreeBinding`, or (b) a line in the PR description / release notes so operators know they can delete it.

### Questions

4. **`withContextTreeSyncLock` is an in-process `Map`.** Two client processes pointed at the same `DATA_DIR` would still race the same checkout (e.g. dev-cli + prod accidentally sharing `FIRST_TREE_HUB_HOME`). The default isolation story per `docs/local-dev-isolation.md` already prevents this, but it'd be worth a one-line note in the PR description acknowledging the constraint — or a `proper-lockfile`-style flock if you think it's reachable from real user setups.

5. **`syncContextTree` is still exported but has no internal caller.** Is there an external consumer of `@first-tree-hub/client` that depends on this exact signature? If not, can we drop the legacy entry point in a follow-up rather than keeping it as dead code?

6. **Legacy `/api/v1/context-tree/info` route — who still calls it?** The PR description says "for compatibility", but with `ClientRuntime` / `AgentRuntime` no longer touching it, the only remaining callers I'd expect are the web admin or external scripts. Could you list them in the description (or confirm there's nothing besides the test)? If it's truly unused, a follow-up to remove both the route and `getContextTreeConfig` would shrink the surface.

### Testing

7. The new server route test couldn't run locally (testcontainers). Please confirm CI green on `pnpm --filter @first-tree-hub/server test` before merge — that test is the load-bearing assertion for the whole fix.

Happy to re-review once these are addressed. No blockers on correctness; the code is fine to ship as-is if the answers to 4–6 are reassuring.

yuezengwu (approved):
LGTM. The agent-scoped sync correctly fixes the cross-org bug — `agentSelectorHook` → `requireAgent` → `identity.organizationId` makes the new Class D route trivially correct, and the per-`(repo, branch)` hashed checkout + in-process lock cleanly isolates concurrent slots. jsdoc updates and the operational caveats in the PR description address the polish items from baixiaohang's review well.

One non-blocking follow-up worth tracking separately (not introduced by this PR, but the semantics drift after it lands):

`packages/server/src/app.ts:593` still hands the legacy single-dir path `join(DEFAULT_DATA_DIR, "context-tree")` to `createKaelRuntime`, and `packages/server/src/services/kael-runtime.ts:128` reads `AGENT.md` from it with the comment "shared with client syncContextTree". After this PR the client never writes to that directory anymore, so kael's AGENT.md lookup will only ever find a file if some other process (or an older client install) happens to have left one behind. Worth a separate issue to either point kael at the new hashed layout or give it its own server-side sync path.

Confirming CI green on `pnpm --filter @first-tree-hub/server test` before merge is still the load-bearing assertion for the fix — please verify the new `agent-context-tree-info.test.ts` passes on CI since it couldn't run locally.

Gandy2025:
@baixiaohang Rebased onto main (commit ff9b27e) to resolve conflicts with #414. Two touchpoints in `runtime.ts` / `client-runtime.ts` `start()`:

- Kept your `gcOrphanSessionBranches()` sweep at the top of both `start()` methods.
- Dropped the `syncContextTree(...)` block that #414 left there — this PR has already moved Context Tree sync into `AgentSlot.start()` so each slot syncs against its own agent's org, not the caller's primary org. Leaving the runtime-level sync would have re-introduced the cross-org bug this PR fixes.
- Adjusted imports accordingly (kept `createGitMirrorManager` / `createLogger` / `GitMirrorManager`; removed `syncContextTree` / `ContextTreeBinding` from the CLI path).

Local verification on the merge commit:

- `pnpm check` clean
- `pnpm typecheck` — 9/9 tasks
- `pnpm --filter @first-tree-hub/client test` — 388/388
- `pnpm --filter @agent-team-foundation/first-tree-hub test` — 106/106

CI still load-bearing for `agent-context-tree-info.test.ts` (server testcontainers don't run locally). Ready for re-review when you have a minute.

baixiaohang (approved):
LGTM. All review feedback addressed (jsdoc, operational caveats), rebase onto main correctly preserves `gcOrphanSessionBranches()` while dropping the now-defunct `syncContextTree(...)` block, and CI is green including the load-bearing `agent-context-tree-info.test.ts` that couldn't run locally.

Filing the kael-runtime AGENT.md regression as a separate issue per @yuezengwu's note — won't block this merge.

PR #407, opened 2026-05-15
feat(web): render real human/agent avatars in chat workspace

yuezengwu:
## Summary

- Chat workspace now shows real avatars instead of first-letter placeholders. Human users get their GitHub OAuth avatar (`users.avatar_url`); agents get their manager-uploaded image when present.
- Backend: new pure resolver `resolveAvatarImageUrl` (uploaded > human GitHub URL > null) wired into `/agents`, `/me/managed-agents`, and per-agent endpoints via a single inline join (no N+1).
- Frontend: message bubbles and `ParticipantsHeader` chips render the resolved URL through the shared `<Avatar>` component (initials/hue fallback preserved).
- The left chat list intentionally keeps the existing agent-only avatar path — see the inline note in `services/me-chat.ts`.

## Test plan

- [x] `pnpm check` (biome) — passes
- [x] `pnpm typecheck` (server / web / command) — passes
- [x] `pnpm test` — 989/989 passing (incl. new 7-case resolver matrix in `agent-avatar-resolver.test.ts`)
- [x] Local dev verification:
  - [x] Sign in via dev-callback as `devuser` with `users.avatar_url` set to a real GitHub URL
  - [x] `GET /api/v1/orgs/{org}/agents` returns the human agent with `avatarImageUrl` resolved to the GitHub URL
  - [x] Message bubble for the human sender renders the GitHub avatar
  - [x] Message bubble for the autonomous agent (no upload) renders the FirstTree logo fallback
  - [x] `ParticipantsHeader` chips render a 16px avatar before the displayName for both human and agent
  - [x] Topbar `UserMenu` avatar continues to work via `user.avatarUrl` (unchanged path)
- [ ] Manual sanity in cloud after merge: verify cross-org human avatar surfaces via the `/me/managed-agents` source

## Notes

- No DB schema changes.
- No `command` / `client` package changes — no version bump required.
- Resolver is defensive: a non-human agent with a stray `userAvatarUrl` (e.g. a future join change) is ignored — only `type === "human"` honors the user fallback.


baixiaohang (commented):
Reviewed the changes — direction is solid: clean separation between the pure `resolveAvatarImageUrl` resolver and the two injection strategies (inline join for list endpoints, dedicated fetch for single-agent endpoints). The 7-case test matrix covers the resolver's defensive paths well. A few questions before approving:

### 1. `services/access-control.ts:listAgentsManagedByUser` — inconsistent join path

The three list endpoints in `services/agent.ts` (`listAgents`, `listAgentsForAdmin`, `listAgentsForMember`) reach `users` via `members.agent_id = agents.uuid` — directly anchored to the agent's own 1:1 member row. That's correct.

`listAgentsManagedByUser` instead reuses the existing `agents.managerId → members.id` join and hops `users` off that member. The inline comment acknowledges this and argues it's safe because "for human agents the manager IS the user". But that invariant is only enforced in the service-layer creation path — there's no schema constraint preventing a future `managerId` change (or an admin tool / migration) from pointing a human agent's manager at someone else's member row. If that ever happens, `/me/managed-agents` will surface the **manager's** GitHub avatar on that human agent, not the agent's own.

Would prefer the same join shape as the other three endpoints: alias `members` (e.g. `agentMembers`) joined on `members.agent_id = agents.uuid`, then `users` off that alias. That makes the projection invariant-free and consistent across all four list paths. WDYT?

### 2. `services/me-chat.ts` — chat list visual divergence (intentional?)

The comment justifies keeping the agent-only path here as "keep the existing visual contract stable", but the detail-view (message bubbles + `ParticipantsHeader`) has already broken that contract by rendering GitHub avatars. So the *net* effect: the same human agent shows as initials+hue in the left chat list, then as a real photo once you open the chat.

Is that an intentional staged rollout (follow-up PR will unify), or the long-term target state? If the latter, the inline comment's rationale could be reworded to reflect the actual reason.

### 3. `fetchUserAvatarForHumanAgent` — no DB integration test

`resolveAvatarImageUrl` is well-covered by the pure-function matrix. The DB function powering single-agent endpoints (`GET /:uuid`, `PATCH`, suspend/reactivate/rebind) is introduced in this PR but has no test exercising the `members.agent_id → users.id` path. Worth a small integration test — especially covering the `type !== "human"` short-circuit, the "no matching member" case, and the "user without avatar_url" case.

Not blocking, nice-to-have.

PR #410, opened 2026-05-15
feat(web): restrict @mention autocomplete to current chat participants

baixiaohang:
## Summary

- `@`-autocomplete in chat now lists **only current chat participants**, not every org-wide agent — the picker matches who the message will actually be routed to.
- The implicit "@<outsider> = address + invite" path is removed; adding new participants goes through the `ParticipantsHeader` `[+]` button as the single explicit entry point.
- `handleSend` mirrors the server's `enforceGroupMention` guard via `scanMentionTokens`, so a hand-typed `@<typo>` / `@<non-member>` shows a precise "use the [+] button to add them first" hint instead of round-tripping to a silent server 400. `sendMut.onError` added as a backstop.

## What changed

`packages/web/src/pages/workspace/center/chat-view.tsx`:

1. **Split candidate lists**
   - `mentionCandidates` (in-chat only) drives `useMentionAutocomplete` + `extractMentions`
   - new `addableCandidates` (chat ∪ `/activity`) drives `ParticipantsHeader` `[+]` dropdown
   - shared `managedByMeMap` extracted for picker grouping
2. **Removed implicit-invite flow**
   - dropped `draftOutsiders` memo and the `addMeChatParticipants(chatId, { participantIds: draftOutsiders })` block in the send path
   - dropped the now-unused `chatParticipantIds` set
3. **Client-side pre-flight**
   - added `scanMentionTokens` import from `@agent-team-foundation/first-tree-hub-shared`
   - in `handleSend`, before the `requiresMention` check, unresolved `@<token>`s (raw tokens not in `mentionCandidates.name`) produce `setUploadError("Cannot @-mention "X" — they're not in this chat. Use the [+] button above to add them first.")` and bail
   - `sendMut.onError` added so any send failure (incl. server 400s the client guard missed) surfaces under the composer
4. **Doc-comments updated** for `activity` query, `draftMentions`, and the two candidate memos so each block's role matches the new behavior

## Test plan

- [x] `pnpm --filter @first-tree-hub/web typecheck`
- [x] `pnpm check` (biome)
- [x] `pnpm --filter @first-tree-hub/web test` — 20 files, 143 cases
- [ ] Manual: open a group chat, type `@` → only current members appear
- [ ] Manual: type `@<unknown-name>` and send → composer shows the "[+] button" hint, no round-trip
- [ ] Manual: click `[+]` → dropdown still lists every visible-but-not-yet-in-chat agent
- [ ] Manual: image-only send in a group still requires `@<member>` mention in the text (existing behavior preserved)

### DIFF
--- a/packages/web/src/pages/workspace/center/chat-view.tsx
+++ b/packages/web/src/pages/workspace/center/chat-view.tsx
@@ -5,6 +5,7 @@ import {
   type MentionParticipant,
   type QuestionAnswerMessageContent,
   type QuestionMessageContent,
+  scanMentionTokens,
 } from "@agent-team-foundation/first-tree-hub-shared";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import { ArrowUp, AtSign, Check, ExternalLink, Eye, MessageSquare, Paperclip, Plus, X } from "lucide-react";
@@ -624,9 +625,12 @@ export function ChatView({
     enabled: !!chatId,
   });
 
-  /** Org-wide agent list for the `@` picker. We surface every agent the
-   *  user can address — not just current chat participants — so `@`-ing
-   *  someone outside the chat acts as both an invite and a mention. */
+  /** Org-wide agent list, consumed only by the ParticipantsHeader `[+]`
+   *  dropdown via `addableCandidates`. The `@` autocomplete is
+   *  membership-scoped (`mentionCandidates` below) and does NOT read
+   *  from this list — inviting a new agent goes through the `[+]`
+   *  button explicitly, not through `@<outsider>`. Also feeds
+   *  `managedByMeMap` for picker grouping. */
   const { data: activity } = useQuery({
     queryKey: ["activity"],
     queryFn: getActivityOverview,
@@ -645,6 +649,13 @@ export function ChatView({
       // wait up to 10s for the polling refetch. See M plan Step 3 in
       // docs/session-creation-on-first-message.md.
       queryClient.invalidateQueries({ queryKey: agentSessionsQueryKey(agentId) });
+    },
+    // Backstop: surface server-side `enforceGroupMention` 400s (and any
+    // other send failure) under the composer instead of silently failing.
+    // The pre-flight check in `handleSend` should catch unresolved
+    // `@<token>` locally; this catches anything that slips through.
+    onError: (err) => {
+      setUploadError(err instanceof Error ? err.message : "Failed to send message");
     },
   });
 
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
LGTM — 目标清晰，client/server 行为基本对齐，单文件聚焦，注释非常详尽。准备 approve。

一个**非阻塞**的边角问题想留个 note，看是否顺手在后续 commit 修一下：

**自我 @-mention 时 client guard 比 server 严**

服务端 `enforceGroupMention` 的未解析 token 检查里，`speakerNames` 是从 `participants` 全量构造（[`services/message.ts:227-232`](https://github.com/agent-team-foundation/first-tree-hub/blob/feat/mention-in-chat-only/packages/server/src/services/message.ts#L227)），**包含发送者自己**，所以 `@<own-name>` 会通过。

但客户端 `mentionCandidates` 显式排除了 self（`p.agentId === myAgentId` skip），所以这里构造的 `memberNames` 不含 self：

```ts
const memberNames = new Set(mentionCandidates.map((c) => c.name?.toLowerCase()).filter(...));
const unresolved = scanMentionTokens(text).filter((t) => !memberNames.has(t));
```

结果：用户手打 `@<自己的名字>` 会被 client guard 拦下，但 hint 是「Use the [+] button above to add them first」—— 这是误导，因为用户**已经在 chat 里了**。

最小修法（推荐）：把 self 也加进 `memberNames`。例如直接从 `chatDetail?.participants` 派生 memberNames 而非从 `mentionCandidates`，或者在构造时额外塞入自己的 name。

属于冷门 case，不阻塞合并。

baixiaohang:
Thanks for catching this @yuezengwu — you're right that the hint text is the actual problem here, not the strictness diff. Fixed in 95a04ab.

The speaker set is now derived from `chatDetail.participants` (via `chatScopedAgentIdentity`) instead of `mentionCandidates`. Since `participants` includes self while `mentionCandidates` excludes it by design, this matches the server's `speakerNames` view exactly — `@<own-name>` is no longer rejected by the client guard, and a real unknown token still gets the original "use the [+] button" hint.

Manually verified: typing `@<my-own-handle>` in a group chat now sends through without the misleading hint.

baixiaohang:
Correction on the last line: I haven't run the web app for a manual click-through — verified by typecheck / biome / 143 unit tests + reading the diff. @baixiaohang has already manually tested the parent PR end-to-end; flagging in case he wants to retest this small follow-up.

yuezengwu:
确认看了新提交 `95a04ab` — 直接从 `chatDetail.participants` 派生 `memberNames`（包含 self）是最干净的修法，commit message 和代码注释也都把背景写清楚了。我之前 approve 仍然有效，可以合并。

PR #411, opened 2026-05-15
feat(server): bind GitHub PR/issue to agent's chat on tool_call

yuezengwu:
## Summary

When an agent runs `gh pr create` / `gh issue create`, eagerly bind the resulting PR / Issue to the agent's current chat so the upcoming `*.opened` webhook (and every comment / review / CI event after it) routes back to that chat instead of forking a fresh one. Solves the delegate-misrouting issue where agent work in `chat C1` produced GitHub events that landed in a new `chat C2`.

### Pipeline

| File | Role |
|---|---|
| `services/github-entity-extractor.ts` | Whitelist parser. Phase 1 covers `Bash gh pr create` / `gh issue create` (single-line URL stdout fits the 400-char `resultPreview` cap, zero loss). |
| `services/github-entity-chat.ts` | `maybeBindGithubEntityFromToolCall` writes a `boundVia="agent_created"` mapping row. `resolveBindingPair` picks an id-sorted representative human so group chats produce **one** mapping row, not one per human. `resolveTargetChat` gains an `isMentionMatched` flag + creation-event guard: opened-style webhooks with no mapping / no mention now return `null` (defensive — `kind: "new"` audience rows already imply mention; the guard keeps future audience refactors safe). |
| `services/session-event.ts` | `appendEvent` hook fires fire-and-forget when `kind: "tool_call" && status: "ok"`. Failures only `log.warn`; the main session-event write is never unwound. |
| `services/github-delivery.ts` | Drops `null` resolutions with a structured `webhook_dropped_creation` log; passes `isMentionMatched: true` for `kind: "new"` audience targets. |
| `services/github-audience.ts` | `our-app-bot` echo branch keeps `kind: "existing"` targets so PRs the agent opens via Hub's installation token still surface their comments / CI back to the original chat. Mention-only targets are still dropped (no fresh chat for our own write). |

### Schema / migration

**No DB schema change.** `github_entity_chat_mappings.bound_via` is `text` with no CHECK constraint; the new `"agent_created"` literal is a pure application-layer extension. No drizzle migration needed.

### Behaviour matrix

| Scenario | Outcome |
|---|---|
| Agent `gh pr create` → tool_call ok | Mapping written immediately; subsequent `pull_request.opened` from `<slug>[bot]` reuses the same chat |
| Agent in group chat | One mapping row (representative human), no N× duplicate cards |
| PR comment after agent_created | Routes to original chat via subscribed path |
| External user opens PR with no mention | `audience: 0`, no chat invented, no mapping |
| External user opens PR with `@<agent>` body | Mention exemption preserved — new chat + `direct` mapping |
| Repeated tool_call event (pending then ok, or duplicate ok) | Idempotent — `ON CONFLICT DO NOTHING` |

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm check` (biome) — clean
- [x] `pnpm --filter @first-tree-hub/server test` — **1009 / 1009 passing** (+22 new)
  - `github-entity-extractor.test.ts` — 9 unit tests (pure function parser)
  - `agent-created-binding.test.ts` — 13 integration tests (direct chat, group representative, idempotency, non-member skip, no-human skip, hook isolation, `appendEvent` → mapping plumbing, `resolveTargetChat` guard branches)
  - `agent-created-webhook-e2e.test.ts` — 4 HMAC-signed real webhook simulations covering opened-from-bot reuse, PR comment follow-up routing, the negative no-mention drop, and the mention exemption
- [ ] Staging soak with a real GitHub App installation (out of this PR's scope — Phase 2 task)

## Out of scope (Phase 2+)

- curl / GitHub MCP tool extractors — needs `resultPreview` cap bump or a dedicated `resultFull` field
- Async GitHub API verification of mapping entities (defends against agents writing junk URLs)
- `pending_creation` short-TTL table if telemetry shows the webhook-arrives-before-tool_call race is non-trivial

### DIFF
--- a/docs/github-entity-chat-binding-design.md
+++ b/docs/github-entity-chat-binding-design.md
@@ -1,0 +1,129 @@
+# GitHub Entity ↔ Chat Binding — Design
+
+Source of truth for how a Hub chat learns to "own" the GitHub PRs and
+Issues that flow through it. Touches `github_entity_chat_mappings` and the
+webhook → audience → delivery pipeline.
+
+## Problem framing
+
+Stripped down, the problem is generic:
+
+> A write to a GitHub entity happened from inside chat C. How do future
+> webhooks for that entity get routed back to C instead of forking a new
+> chat?
+
+The framing matters: "agent ran `gh pr create`" is *one* way the write can
+happen, not the only one. The long-term design space has four families:
+
+| Family | Where the binding signal comes from | Coverage | Cost | Robustness |
+|---|---|---|---|---|
+| **A. Explicit declaration** | Agent calls a Hub-provided `bind_chat_to_github_entity(url)` API after the write | Whatever agents are taught to call | Medium (teach the agent) | Strong (structured input) |
+| **B. Outbound proxy** | All agent → GitHub traffic flows through Hub's installation token / proxy; binding is a side-effect of the outbound write | Near-total | High (forcing traffic shape) | Strong |
+| **C. Webhook backfill** | When `*.opened` arrives, reverse-look-up recent chat activity in a time window | Generic, no per-tool work | Medium (heuristic + indexes) | Fuzzy (may misbind) |
+| **D. Tool-call stdout extraction** | Regex against `resultPreview` of a whitelisted tool | Whatever extractors get written | Low | Brittle (string-level coupling) |
+
+**Target shape**: A is the main path, B is the audit floor, C is the
+safety net, D demotes to a fallback for tools we can't influence.
+
+## What ships today (Phase 1)
+
+Pure family **D**, scoped tight: only `Bash gh pr create` and `Bash gh
+issue create` are recognised. Both emit a single-line URL on stdout which
+fits the 400-char `resultPreview` cap with no information loss.
+
+Pipeline:
+
+1. **Client** (AgentRuntime) — runs the tool, reports `session:event`
+   with `kind="tool_call"`, `status="ok"`, `resultPreview="<url>"` over
+   the existing WebSocket frame. No client-side parsing.
+2. **Server** — `sessionEventService.appendEvent` writes the event row
+   and, on `kind="tool_call" && status="ok"`, fires
+   `maybeBindGithubEntityFromToolCall` fire-and-forget:
+   - `extractGithubEntity` (whitelist parser) → `(entityType, entityKey)`
+   - `resolveBindingPair` → `(org, representative_human, delegate)`
+   - `insertMappingIfAbsent` with `boundVia="agent_created"`
+3. **Webhook arrives** — `pull_request.opened` from `<app-slug>[bot]`.
+   `resolveAudience` finds the existing mapping in the subscribed path
+   and routes the card to the same chat.
+
+Schema unchanged. `bound_via` is `text` with no CHECK, so the new
+`"agent_created"` literal is an application-layer extension only.
+
+### Defensive scaffolding
+
+Even though family D ships first, the data model is set up so future
+families can plug in without rewrites:
+
+- **`bound_via` open-ended** — adding `agent_declared` (family A) or
+  `outbound_proxy` (family B) is a string literal change. No migration.
+- **`resolveBindingPair` is entry-point-agnostic** — any new caller
+  (HTTP endpoint for family A, proxy hook for family B) reuses it.
+- **`insertMappingIfAbsent` + composite PK** — multiple binding sources
+  can coexist without dedup logic; the loser of a race re-reads the
+  winner's row.
+- **`resolveTargetChat` creation-event guard** — opened webhooks with no
+  mapping + no mention return `null`. `isMentionMatched` is required (not
+  optional) so any future caller has to make an explicit fail-closed
+  decision.
+- **`our-app-bot` echo branch** — keeps `kind: "existing"` so Hub's own
+  outbound writes still route through the subscription path; drops
+  `kind: "new"` so they don't mint a chat to echo themselves.
+
+## Why D is acceptable as the Phase 1 entry point
+
+- Zero client work — every existing AgentRuntime already reports
+  `session:event` frames over WS. No SDK version bump.
+- Covers the 80% case (`gh pr create`) on day one.
+- Failure mode is "binding doesn't happen" — the legacy
+  fresh-chat-on-opened path still works, so the worst case is the
+  pre-PR behaviour, not a regression.
+- The whitelist is honest about its limits — anything not on it falls
+  through without misbinding.
+
+## What's explicitly *not* the long-term plan
+
+Adding more extractors (curl, GitHub MCP, …) layers stdout-parsing on
+top of stdout-parsing. The brittleness compounds rather than cancels.
+Phase 2 deliberately does *not* commit to "extend D forever"; it
+delineates the boundary where family A starts paying off.
+
+## Phase 2 candidates (data-driven)
+
+Driven by what telemetry shows once Phase 1 is in staging:
+
+1. **`webhook_arrived_without_mapping` counter**, scoped to
+   `(eventType, action) ∈ {(pull_request, opened), (issues, opened)}` and
+   sender = our App bot. Quantifies the family-A gap — high count means
+   agents are creating entities through paths D doesn't cover.
+2. **`resultPreview` cap bump** to ~2 KB *or* a `resultFull` field for
+   whitelisted tool kinds. Unblocks curl + MCP without forcing every
+   session event payload to bloat.
+3. **Family A intro**: a small `POST /api/agents/me/github-entity-binding`
+   endpoint the agent can call right after a non-`gh` GitHub write. Agent
+   prompt nudge in parallel.
+4. **Async GitHub API verification** to garbage-collect "ghost mappings"
+   where the agent wrote a syntactically valid but non-existent URL into
+   `resultPreview`.
+
+## Open questions
+
+- **Race window**: tool_call ok → DB insert is sub-ms, GitHub webhook
+  delivery is ~100ms+. The mapping should win deterministically, but
+  Phase 1 doesn't measure it. If the counter above shows non-trivial
+  loss, a `pending_creation` short-TTL table (B-flavoured pre-occupy)
+  comes back as an option.
+- **Group chat ergonomics**: the representative-human pick is
+  deterministic (delegateMention-linked first, id-sorted fallback). If
+  product wants the card to surface "PR opened by <X>" with X = actual
+  GitHub author, that lives in the card builder, not the mapping row.
+
+## File map
+
+| Concern | File |
+|---|---|
+| Whitelist parser | `packages/server/src/services/github-entity-extractor.ts` |
+| Binding service + guard | `packages/server/src/services/github-entity-chat.ts` |
+| Session-event hook | `packages/server/src/services/session-event.ts` |
+| Audience echo branch | `packages/server/src/services/github-audience.ts` |
+| Delivery null handling | `packages/server/src/services/github-delivery.ts` |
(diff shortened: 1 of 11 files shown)


baixiaohang (commented):
整体评估：实现思路清晰，测试覆盖到位（22 个新增测试覆盖单元、集成、HMAC 签名 e2e）。`agent_created` 通过 `bound_via` 文本列做应用层扩展、不动 schema，方向合理。下面是几个非阻塞性的观察点，主要是文档一致性和安全默认值的小问题。

## 1. `github-audience.ts` 顶部注释已经和新行为不一致

`our-app-bot` 这一支的行为已经从 "force empty" 改成 "filter to `kind: 'existing'`"，但文件顶部的两处 doc 注释还停留在旧描述上：

- `ActorIdentity` 的注释（约 L38-40）："Audience is forced empty because the event is a downstream effect of Hub's own outbound write"
- `resolveAudience` 的 docstring（约 L107-108）："When the actor is our App's bot user (DP13), the whole audience is suppressed."

行为差异本身在新代码块的内联注释里解释得很清楚，但顶部 doc 还需要同步更新，避免未来读到这个文件的人被旧注释误导。

## 2. `isMentionMatched` 默认 `true` —— 失效保护方向的小担忧

`resolveTargetChat` 把 `isMentionMatched` 设成 `true` 默认，这样未传 flag 的调用者会回退到 "可以铸新 chat" 的旧行为。新加这个 guard 的目的就是兜底防止未来的 audience 重构意外允许 opened 事件造新 chat。当前 in-tree 唯一调用方 `github-delivery.ts` 已经显式传 `true`，所以默认值对现有代码没影响。

但从 fail-closed 的角度看，把默认改成 `false`（或者直接做成必填参数）会让 guard 更稳健 —— 任何新增调用方忘记传 flag 都会走到 null 分支，比静默允许造 chat 安全。这个改动是否值得做？还是说当前的 "preserve legacy behaviour" 注释里的考量更重要，我没看到？

## 3. `resolveBindingPair` 的"代表 human"挑选只按 id 排序

群聊场景下，假设有的 human 设了 `delegateMention = <这个 delegate>`、其他没设，当前是按 id 字典序挑第一个 active human，可能挑到没有 delegateMention 关系的那个。

对纯路由（最终落到 chatId）来说没问题，但想确认一下：下游有没有任何逻辑是基于 mapping 行的 `humanAgentId` 字段做判定的（比如通知接收人、权限校验、卡片署名）？如果有，是否应该优先选有 delegateMention 关系的 human、其次再 id 排序？

## 4. 顺便确认一下 race 窗口的预期

PR 描述里把 "webhook 比 fire-and-forget 先到" 列到了 Phase 2，靠 telemetry 决定要不要加 `pending_creation`。这个我同意 defer，但想确认一下你预期的延迟差：tool_call ok → DB insert 一般是 sub-ms 级，GitHub `*.opened` webhook 往返通常在百 ms 级别，两边的 gap 应该很大。staging 灰度的时候建议加一个具体的 metric（比如 "webhook arrived but mapping row absent" 的计数）来量化这个 race 实际发生率。

---

不影响合并意图的几个加分项：
- ✅ `kind: "new"` audience target 在 our-app-bot 分支被丢弃的语义化解释（"never useful"）写得很清楚
- ✅ Fire-and-forget 的 `.catch(log.warn)` 隔离 + status="ok" 提前过滤，正确地保护了主 session-event 写入路径
- ✅ Group chat 代表人挑选 + `ON CONFLICT DO NOTHING` 双重保证不会在群聊里产生 N 倍 mapping 行
- ✅ `agent-created-webhook-e2e.test.ts` 真 HMAC 签名走 full pipeline，比纯 mock 测试可信很多

baixiaohang (commented):
### Verdict on this PR

Fine to land as-is. The implementation is a clean Phase-1 cut that delivers immediate value for the 80% case (agent runs `gh pr create` via Bash). My specific in-PR review comments are in the earlier review — none of them are merge blockers; the doc-staleness one in `github-audience.ts` is the only thing I'd consider strictly worth fixing before merge, the rest are open questions / non-blocking nits.

I'm leaving this second comment to surface an architectural concern I think is worth recording somewhere durable, before this PR's "Phase 2+" list calcifies into an implicit roadmap.

### The shape of the underlying problem

If you strip this PR down, the question it answers is generic:

> "A write to a GitHub entity happened from inside chat C — how do future webhooks for that entity get routed back to C?"

It's not specifically about "agent ran `gh pr create`". That's just one way the write can happen. Once framed that way, the design space has four families:

| Family | Where the binding signal comes from | Coverage | Cost | Robustness |
|---|---|---|---|---|
| **A. Explicit declaration** | Agent calls a Hub-provided `bind_chat_to_github_entity(url)` tool/API after the write | Whatever agents are taught to call it | Medium (changing agent behavior) | Strong (structured input) |
| **B. Outbound proxy** | All agent → GitHub traffic flows through Hub's installation token / proxy; binding is a side-effect | Near-total | High (forcing traffic shape) | Strong |
| **C. Webhook backfill** | When `*.opened` arrives, reverse-look-up recent chat activity in a time window | Generic, no per-tool work | Medium (heuristic + indexes) | Fuzzy (may misbind) |
| **D. Tool-call stdout extraction** *(this PR)* | Regex against `resultPreview` of a whitelisted tool | Whatever extractors get written | Low | Brittle (string-level coupling) |

### Why this matters before "Phase 2+" gets locked in

The current "Out of scope (Phase 2+)" list reads like "continue extending D" — curl extractor, GitHub MCP extractor, `resultPreview` cap bump, `pending_creation` short-TTL table. Each of those items is real, but each one is more stdout-parsing layered on top of stdout-parsing. The brittleness compounds rather than cancels, and the whitelist will keep chasing every new tooling surface forever.

The healthier long-term shape, in my view, is: **A is the main path, B is the audit floor, C is the safety net, D demotes to a fallback for tools we can't influence**. This PR is a perfectly fine stepping stone toward that shape — *if* the team agrees that's where we're heading.

### What this PR already gets right and shouldn't change when we move toward A

- `bound_via` text column with no DB CHECK is the right extensibility surface — adding `agent_declared` / `outbound_proxy` later is a pure app-layer change with zero migration.
- `resolveBindingPair` is correctly written to be entry-point-agnostic — any future caller (a new HTTP endpoint, an MCP tool, a proxy hook) reuses it for free.
- Idempotent `ON CONFLICT DO NOTHING` makes it safe for multiple binding sources to coexist without dedup logic.

These three are load-bearing for the long-term direction. Worth preserving them deliberately as we add more entry points.

### Suggested memo home

If the framing above is worth keeping around, three candidates:

1. **A short design doc** under `docs/` (e.g. `docs/github-entity-chat-binding-design.md`) — heaviest weight, most discoverable, fits the repo's existing `*-design.md` convention.
2. **A tracking issue** that links back to this PR — lightweight, threadable, natural home for Phase-2 follow-up PRs.
3. **A header note** in `services/github-entity-extractor.ts` (e.g. `// Phase 1 fallback. Long-term direction: see <link>`) so future code-touchers stumble into it without grep gymnastics.

Personal lean: (1) + (3). The design doc captures the framing, the code header makes sure the next person who touches the extractor sees the framing exists.

@baixiaohang — your call on which artifact to draft. None of this blocks merging this PR.

yuezengwu:
@baixiaohang Thanks for the thorough review — pushed `cd7a94c` (rebased onto latest `main`) with the responses below.

### Direct review action items

1. **`github-audience.ts` stale doc** — fixed. Both the `ActorIdentity` enum comment and the `resolveAudience` docstring now describe the actual behaviour (keep `kind: "existing"`, drop `kind: "new"` on `our-app-bot`), with the rationale spelled out so the next reader doesn't have to grep the body to reconcile.

2. **`isMentionMatched` default** — taken your suggestion all the way: the parameter is now **required**, not just defaulted-to-`false`. Any future caller that forgets the flag fails closed at the type level. The existing in-tree caller (`github-delivery.ts`) already passes `true` explicitly; the test wrapper in `resolve-target-chat.test.ts` keeps a `?? true` shim so the legacy assertions stay readable, but production callers can't accidentally regress.

3. **`resolveBindingPair` representative pick** — good catch. Selection order is now:
   1. Active humans whose `delegateMention` points at the reporting agent
   2. id-sorted-first fallback among the rest

   Added a regression test that deliberately seeds the id ordering to disagree with the delegate-link ordering, so the override is exercised explicitly (and protected against accidental refactor).

4. **Race window telemetry** — agree the gap should be order-of-magnitude in our favour (sub-ms DB insert vs. ~100ms+ webhook RTT), but you're right that we shouldn't ship Phase 1 without a way to *measure* it. Added it to the Phase 2 candidate list in the new design doc with a concrete metric name (`webhook_arrived_without_mapping`, scoped to opened-action events from our own bot). Keeping it out of *this* PR so the scope stays the binding feature itself, not the observability around it.

### Architectural framing (your second review)

You're right that the "Phase 2+" list as I wrote it reads like "extend D forever" — that wasn't the intent, but the PR description didn't surface the framing. Took both of your suggested artifacts:

- **Design doc**: [`docs/github-entity-chat-binding-design.md`](../blob/feat/agent-tool-call-github-binding/docs/github-entity-chat-binding-design.md) — frames the four families (declaration / outbound proxy / webhook backfill / stdout extraction), pins Phase 1 explicitly as the *fallback* path (low-cost first cut, not the canonical entry point), and re-lists Phase 2 candidates with trade-offs called out.
- **Code header**: `services/github-entity-extractor.ts` now opens with a note pointing at the design doc so the next person reaching for "just add curl support" sees the framing before they grep for a regex to add.

The three properties you flagged as load-bearing for the long-term direction (`bound_via` open-ended, entry-point-agnostic `resolveBindingPair`, idempotent `ON CONFLICT DO NOTHING`) are now called out explicitly in the design doc as defensive scaffolding to preserve.

### Test results after changes

- `pnpm typecheck` ✅
- `pnpm check` ✅
- `pnpm --filter @first-tree-hub/server test` ✅ **1066 / 1066 passing** (added one new regression test for the delegate-link representative pick)

Ready for another look.

baixiaohang:
Thanks @yuezengwu — verified `cd7a94c` against your summary, every item lands cleanly:

1. **`github-audience.ts` docs** — both the `ActorIdentity` enum comment and the `resolveAudience` docstring now describe the actual behaviour, and the actor=`agent` branch got an extra clarification as a bonus. ✅
2. **`isMentionMatched` required** — appreciate going past "default false" all the way to required. The test wrapper in `resolve-target-chat.test.ts` keeps legacy assertions readable without re-opening the fail-open door. ✅
3. **`resolveBindingPair` delegateMention preference** — implementation is the obvious one (`humans.find(h => h.delegateMention === reporterAgentId) ?? humans[0]`), and the regression test is *better* than what I'd asked for: deliberately fuzzes UUID ordering to guarantee the id-sorted fallback would pick wrong, so the delegate-link override has to fire to pass. That'll survive any future UUID-generation refactor. ✅
4. **Race telemetry** — keeping it out of this PR is the right call; the design doc's `webhook_arrived_without_mapping` metric is specific enough to act on without scope creep here. ✅

`docs/github-entity-chat-binding-design.md` reads exactly the way I was hoping for — the "Target shape: A main, B audit floor, C safety net, D fallback" line at the top of the families table is the one sentence that prevents the next dev from interpreting Phase 1 as the canonical entry point, and the "Defensive scaffolding" section explicitly names the three load-bearing properties so they don't get refactored away by accident. The `extractor.ts` header pointer closes the loop for someone who lands in the code first.

My personal verdict: ready to ship. I'll loop @baixiaohang for the explicit approve since he asked to be in the loop on the merge call — no further action needed from your side.

baixiaohang (approved):
Approved. Phase-1 scope is right-sized, all four review items got structural fixes (not cosmetic patches), and the long-term direction is captured in `docs/github-entity-chat-binding-design.md` so the next contributor reaches for family A/B/C before extending D.

Merge once CI is green.

PR #413, opened 2026-05-15
fix(web,server): reduce GET /chats refetch storm under WebSocket push

baixiaohang:
## Summary

- One `["me", "chats"]` invalidate was fanning out into **three concurrent** `GET /chats` requests because the chat-detail page kept two redundant `listMeChats` queries just to read `participant.type` and `membershipKind`. Both are now available on `ChatDetail` (the former via the existing wire schema, the latter via the new `viewerMembershipKind` field added in this PR).
- Combined with sub-500ms WS frame storms (`session:event` / `session:state` / `chat:message`), this drove **~9-12 `GET /chats`/s** under a single agent burst. Expected after this change: **~1-2/s**.
- Throttle window raised 500ms → 1s; `chat:message` now goes through the same leading + trailing throttle helper (was firing one invalidate per frame). `ws:reconnect` synthetic frame closes the markRead gap that would otherwise be re-introduced by removing the list-query dependency.

## What changed

**Shared / Server (wire contract)**
- `packages/shared/src/schemas/chat.ts`: `chatDetailSchema` gains `viewerMembershipKind: "participant" | "watching" | null`. `null` = supervisor / admin view (caller reaches the chat through a managed agent without a direct `chat_membership` row).
- `packages/server/src/api/chats.ts` (admin class-C `GET /chats/:chatId`) and `packages/server/src/services/chat.ts` (`getChatDetail`, used by the agent SDK) both compute and return the field.

**Web — drop the redundant list queries**
- `packages/web/src/pages/workspace/center/chat-by-id.tsx`: stop fetching `listMeChats({ filter: "all" })` and `listMeChats({ filter: "watching" })`; `pickPrimaryAgent` and `isWatching` now derive from `chatDetail.participants` and `chatDetail.viewerMembershipKind`.

**Web — WS-driven invalidate**
- `packages/web/src/hooks/use-admin-ws.ts`:
  - `ME_CHATS_INVALIDATE_THROTTLE_MS` 500 → 1000.
  - `chat:message` now routes through `invalidateMeChatsThrottled` instead of firing an invalidate per frame.
  - The catch-up invalidate on `(re)open` now also touches `["chat-detail"]` so chats the caller was added to / removed from while the socket was down refresh on reconnect.
  - On a genuine reconnect (`reconnectAttempt > 0`, captured before the reset), `broadcast({ type: "ws:reconnect" })` so subscribers can re-run freshness side effects.
- `packages/web/src/pages/workspace/center/chat-by-id.tsx`: subscribe to `ws:reconnect` to re-fire markRead. All three triggers (mount, `chat:message`, `ws:reconnect`) share a `canMarkRead` guard (`chatDetail` loaded && `viewerMembershipKind !== null`) to keep supervisor views from writing orphan `chat_user_state` rows.

**Drive-by fix**
- `packages/web/src/pages/agent-detail.tsx`: `onRefresh` now also invalidates `["chat-detail"]`. Pre-existing bug surfaced during the audit — editing an agent's `displayName` left stale labels in the open chat-detail panel because that cache joins `agents`.

**Test**
- `packages/client/src/__tests__/chat-context.test.ts`: extend `mkChatDetail` mock with `viewerMembershipKind: "participant"` for the schema change.

**Version bumps** (per [`docs/versioning-and-publishing.md`](docs/versioning-and-publishing.md))
- `packages/command` 0.14.3 → 0.14.4 (client imports `ChatDetail`).
- `packages/shared` 0.2.2 → 0.2.3 (`chatDetailSchema` is an externally-importable surface).

## Why the throttle bump is safe

`MeChatRow.liveActivity` (the WorkingChip) becomes stale after `LIVE_ACTIVITY_STALE_MS = 60_000` server-side (`packages/shared/src/schemas/me-chat.ts:88`). 1s / 60s = 1.7% slack — well below any perceptual threshold. The avatar engaged ring (`engagedAgentIds`) is driven by the same throttle and degrades the same way.

## Test plan

- [ ] **Frequency baseline**: open a chat, DevTools → Network → filter `/chats` → observe an agent burst. Expect ~1-2 `GET /chats`/s (was ~9-12).
- [ ] **WS reconnect re-fires markRead**: open chat A → DevTools → Throttling → Offline 30s → Online → expect chat A's unread badge in the conversation list to clear (driven by the `ws:reconnect` sentinel).
- [ ] **Supervisor view doesn't write orphans**: as admin, open a chat you reach via a managed agent (no direct `chat_membership` row) → confirm `chat_user_state` does not gain a new row for your `humanAgentId`. If the path is hard to reproduce, the code-level guard (`canMarkRead`) is sufficient.
- [ ] **Stale displayName fix**: in the chat-detail panel, edit a participant agent's `displayName` from `agent-detail.tsx` → confirm the open chat-detail panel updates immediately (no manual refresh needed).
- [ ] **Speaker vs watcher UI** continues to render correctly with the new `viewerMembershipKind` source.

## Known follow-up (non-blocking)

If a caller is kicked out of a chat while the WS is down, the `ws:reconnect` sentinel fires before `chat-detail` refetches → one redundant `POST /read` returns 404. Safe (idempotent, no dirty write, doesn't block UI), but cosmetic.

### DIFF
--- a/packages/client/src/__tests__/chat-context.test.ts
+++ b/packages/client/src/__tests__/chat-context.test.ts
@@ -29,6 +29,7 @@ function mkChatDetail(overrides?: Partial<ChatDetail>): ChatDetail {
     title: "v1 ship",
     firstMessagePreview: null,
     engagementStatus: "active",
+    viewerMembershipKind: "participant",
     ...overrides,
   };
 }
--- a/packages/command/package.json
+++ b/packages/command/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agent-team-foundation/first-tree-hub",
-  "version": "0.14.3",
+  "version": "0.14.4",
   "type": "module",
   "description": "First Tree Hub — unified CLI for server, client, and agent management",
   "exports": {
--- a/packages/server/src/api/chats.ts
+++ b/packages/server/src/api/chats.ts
@@ -83,11 +83,29 @@ export async function chatRoutes(app: FastifyInstance): Promise<void> {
 
     const engagementStatus = await getCallerEngagement(app.db, chat.id, scope.humanAgentId);
 
+    // Caller's own membership row — drives speaker-vs-watcher UI on the
+    // chat detail page without forcing the client to round-trip through
+    // `/orgs/:orgId/chats`. `null` when the caller reaches the chat via
+    // supervision (managed agent is a speaker) rather than direct
+    // membership; that's the same null-shape `MeChatRow` carries when
+    // listChats filters to supervised-only rows.
+    const [callerMembership] = await app.db
+      .select({ accessMode: chatMembership.accessMode })
+      .from(chatMembership)
+      .where(and(eq(chatMembership.chatId, chat.id), eq(chatMembership.agentId, scope.humanAgentId)))
+      .limit(1);
+    const viewerMembershipKind: "participant" | "watching" | null = callerMembership
+      ? callerMembership.accessMode === "speaker"
+        ? "participant"
+        : "watching"
+      : null;
+
     return {
       ...chat,
       title,
       firstMessagePreview,
       engagementStatus,
+      viewerMembershipKind,
       createdAt: chat.createdAt.toISOString(),
       updatedAt: chat.updatedAt.toISOString(),
       participants: participants.map((p) => ({
--- a/packages/server/src/services/chat.ts
+++ b/packages/server/src/services/chat.ts
@@ -174,7 +174,31 @@ export async function getChatDetail(db: Database, chatId: string, selfAgentId: s
     type: p.type,
   }));
 
-  return { ...chat, participants, title, firstMessagePreview };
+  // Match the chatDetailSchema wire contract — the chat-first workspace
+  // reads this field instead of round-tripping `/orgs/:orgId/chats` just to
+  // distinguish speaker vs watcher view. Agent-SDK callers always reach
+  // this code with their own uuid as `selfAgentId`, and the SDK only sees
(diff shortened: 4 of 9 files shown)


yuezengwu (approved):
LGTM — well-scoped fix with clear tradeoffs.

**Verified:**
- Wire contract aligned across both server paths (admin `api/chats.ts` and SDK `services/chat.ts`); web `getChat` → `/chats/:chatId` picks up the new field.
- `canMarkRead` guard (`chatDetail != null && viewerMembershipKind !== null`) shared by all three trigger points (mount / `chat:message` / `ws:reconnect`) — supervisor views won't write orphan `chat_user_state` rows.
- `isReconnect` captured before `reconnectAttempt` reset, so the synthetic `ws:reconnect` doesn't double-fire on first handshake.
- `markedChatIdRef` single-slot is correct: strict-mode double mount / chat-detail refetch won't re-POST, while A→B→A navigation re-fires as intended (new messages re-fire via the WS subscriber path).
- Throttle bump 500ms→1s vs 60s server-side `liveActivity` stale window — 1.7% slack, well below perceptual threshold.
- No schema/migration changes; the new membership lookup is a single `(chatId, agentId)` PK hit.

**Non-blocking:**
- `getChatDetail` does an extra membership SELECT even for agent-SDK callers where it almost always resolves to `"participant"`; contract consistency over the micro-saving is the right call and the comment is explicit.
- The known `404 POST /read` race after a kick during a WS gap is harmless (idempotent, doesn't block UI), as the PR notes.

Drive-by `["chat-detail"]` invalidate in `agent-detail.tsx onRefresh` is a real bug fix — good catch.

PR #423, opened 2026-05-15
fix(server): require GitHub admin proof before binding installation_id

### LINKED ISSUE #312: GitHub App: /user/installations does not prove admin access — claim endpoint can hijack installs
**Severity: P1 (security)**
**Source: codex challenge against PR #300 Phase C, 2026-05-12**
**Status: pre-existing — was already in PR #300's Phase B (commit `366a7a3` codex P0-2). Phase C C.10 (commit `753d74b`) exposes it more by adding the `claim` endpoint where the caller picks the `installation_id`.**

## Finding

The OAuth callback `installation_id` check and the new manual-claim endpoint both rely on [`listUserAccessibleInstallationIds`](packages/server/src/services/github-app.ts) which calls `GET /user/installations`. The code's own comment asserts this proves the user *administers* the installation:

> Critical security primitive: the OAuth callback uses this to verify that an `installation_id` query parameter — which arrives over an insecure channel (the user's browser address bar) — actually belongs to the authenticated user.

Per [GitHub's docs](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-app-installations-accessible-to-the-user-access-token), `/user/installations` actually returns installations the user has `:read` / `:write` / `:admin` access to — **including via plain org membership**. Org repo `:read` is enough to be in the result, no admin role required.

## Threat model

A Hub admin of org Y who is also a **non-admin GitHub member** of org X can:

1. Sign in to Hub with GitHub OAuth (gets a user access token)
2. Call `GET /user/installations` — org X's GitHub App install shows up because they have repo read access via org membership
3. POST `/api/v1/orgs/Y/github-app-installation/claim { installationId: <X's> }`
4. The check passes — the install is now bound to Hub org Y

End state: org X's GitHub App install (webhook deliveries, repo access tokens) is now flowing to Hub org Y, controlled by the attacker.

## Affected sites

- `packages/server/src/services/github-app.ts` — `listUserAccessibleInstallationIds` (the misnamed primitive)
- `packages/server/src/api/auth/github.ts` — OAuth callback `installation_id` validation
- `packages/server/src/api/orgs/github-app.ts` — POST `/orgs/:orgId/github-app-installation/claim`

## Right fix (sketch)

Replace the `/user/installations` check with one that actually proves admin/owner role on the GitHub-side account:

- **For Org installs (`account.type === \"Organization\"`):** call `GET /user/memberships/orgs/{login}` and require `role === \"admin\"`. This endpoint requires the user themselves so a non-admin can't lie.
- **For User installs (`account.type === \"User\"`):** require `account.id === profile.githubId` (numeric match — the installer must literally be that GitHub user).

Both checks happen in two places (callback + claim) — extract to a helper.

Gandy2025:
## Summary

`/user/installations` returns every install the caller has *read* access to (plain org membership is enough), so a Hub org admin who is only a non-admin member of a GitHub org could pass that org's `installation_id` to the OAuth callback or the manual claim endpoint and hijack the install onto their Hub team.

Replace the primitive with `verifyUserCanAdministerInstallation`:
- **User-type install**: caller's GitHub ID must equal the install account's ID.
- **Org-type install**: `GET /user/memberships/orgs/{login}` must return `state=active` and `role=admin` (a pending invite does NOT count).

Callback reordered to fetch the installation first, verify, then upsert — fail-closed on any step. Claim endpoint pulls the caller's GitHub ID off `auth_identities.identifier` to drive the User-type comparison.

Removes `listUserAccessibleInstallationIds` outright; the 500-page pagination cap goes with it.

Closes #312, #319.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `biome check` clean on touched files
- [x] `pnpm --filter @first-tree-hub/server test` — 121 files / 975 tests pass
- [x] New unit coverage: `verifyUserCanAdministerInstallation` — User match/mismatch, Org admin/member/pending/404/401
- [x] New integration coverage:
  - Callback refuses to bind when caller is a non-admin member of the GitHub org install
  - Claim endpoint 403s on Org-type member, on User-type ID mismatch, and on non-member 404
- [ ] Manual: verify GitHub App `members:read` org permission is declared (operator-side; required for `/user/memberships/orgs/{login}`)
- [ ] Manual: post-deploy, confirm a non-admin org member can't claim that org's installation_id via Settings or callback URL


yuezengwu (approved):
LGTM — the security primitive is now properly aligned with GitHub's authorization model.

## What changed

The OAuth callback and `/claim` endpoint previously used `GET /user/installations` to authorize the caller against a user-supplied `installation_id`. That endpoint surfaces every install the token has *read* access to, so plain (non-admin) org membership was enough to pass the check — a Hub admin who was a regular member of a GitHub org could bind that org's install onto their own Hub team.

The fix replaces it with `verifyUserCanAdministerInstallation`, which splits by account type:
- **User-type**: compares `installation.accountGithubId` against the caller's own GitHub ID (in-process, no HTTP).
- **Org-type**: calls `GET /user/memberships/orgs/{login}` and requires `state=active` AND `role=admin`. 404 (non-member) → clean `false`; non-401 errors throw and fail closed.

Callback flow reordered to fetch-then-verify-then-upsert, so the install row never lands without an admin proof. Claim endpoint reads `auth_identities.identifier` for the caller's numeric GitHub ID to drive the User-type comparison.

## Review notes

- **Fail-closed posture is consistent**: any throw inside the verify/fetch/upsert block clears `installationId` so sign-in still succeeds but no row binds. The `Number(identity?.identifier)` → NaN edge case on the claim path also fails closed cleanly (NaN === N is false → User-type returns false).
- **Orphan reclaim sweep still safe**: `findUnboundInstallationsByAccount(_, profile.githubId)` only matches User-type orphans on the caller's own account, which is an intrinsic admin proof. No change needed there.
- **Org rename resilience**: if a GitHub org renames between the row's last webhook update and a claim, the membership endpoint 404s and we refuse the bind — operator can wait for the next webhook to refresh `accountLogin` or re-install. Acceptable.
- **Test coverage is solid**: unit covers all four branches (User match/mismatch, Org admin/member/pending/404/401); integration tests exercise both the callback hijack vector (non-admin org member) and four claim-endpoint refusal paths (User-type ID mismatch, Org-type member, non-member, missing row).

## Items worth confirming before merging

- The unchecked test-plan item about declaring `members:read` is the operator-side dependency. Worth being explicit in deploy docs / a follow-up: `GET /user/memberships/orgs/{org}` is called with the *user access token*, so the App must be authorized by the user to read their org memberships. If this isn't already granted on the App's settings page in prod, every Org-type bind will throw on a non-401 status and fall through to the fail-closed branch (sign-in still works, but no install ever binds).
- Post-deploy smoke as the test plan describes — a non-admin org member hitting both callback and claim paths against a known install — would be a worthwhile manual check given the security sensitivity.

Closes #312, #319 cleanly. Approving.

baixiaohang (approved):
LGTM — agrees with @yuezengwu's review. The `/user/installations` → `verifyUserCanAdministerInstallation` swap fixes the hijack vector cleanly, fail-closed is consistent across verify/fetch/upsert, the orphan-reclaim sweep is still safe (only User-type rows can match `accountGithubId === profile.githubId`, which is itself an admin proof), and test coverage hits all four verify branches plus the callback and claim refusal paths.

Two non-blocking nits, both fine as follow-ups:

1. `packages/server/src/api/orgs/github-app.ts:209` — the catch-all on non-401 `GithubAppApiError` wraps everything in `"Couldn't verify GitHub access for this installation — try again in a moment"`. If the App is missing the `Organization → Members: read` permission in prod, every Org-type claim will hit a persistent 403 and surface this "try again" message to the admin — misleading for a state that won't self-heal. Consider distinguishing transient (5xx / network) from permanent (4xx that isn't 401) and emitting a clearer hint in the latter.

2. `packages/server/src/api/auth/github.ts:456-458` — the orphan-reclaim block still describes its authorization basis as "the same authorization basis the callback's `/user/installations` check enforced when the row was first written". Stale phrasing now that the primitive is gone; the underlying logic (User-type `accountGithubId` match) is still correct, but the comment will mislead the next reader.

Operator-side reminder for deploy (already on the test plan, restating for visibility): `GET /user/memberships/orgs/{login}` with a GitHub App user access token requires the App to declare `Organization → Members: read`. If that isn't already granted, existing Org-type installs will need their admins to re-authorize before any binds work. Worth confirming on the GitHub-side App settings page before flipping this on.

PR #425, opened 2026-05-15
refactor(web,server): horizontal tabs for agent detail + admin private-agent fix

Gandy2025:
## Summary

Two related changes to the agent-detail page, kept as separate commits for review/rollback granularity:

**1. fix(server)** — Admin couldn't view private agents managed by other members. The admin-only `/orgs/:orgId/agents/all` listing already exposes those rows for cross-member governance, but `GET /agents/:uuid` went through `requireAgentAccess(..., "visible")` which only allowed org-visible OR managed-by-caller — so admins saw the rows in the list and got 404 on detail. Adds the same `isAdmin` short-circuit the `"manage"` branch already uses. Covers detail, sessions, client-status, agent-config, and avatar endpoints that share the visible guard.

**2. refactor(web)** — Restructure `/agents/:uuid` from left-sidebar anchor-scroll into top horizontal tabs with real subroutes (`profile/setup/prompt/tools/resources`). URLs are shareable, browser back skips siblings the way users expect. Layout matches team page's flat chrome (no bg, no border-bottom on header). Lifecycle moves into Setup tab bottom; `Advanced` renamed to `Resources` to close the catch-all bucket. Visual: real `Avatar` replaces generic logo; Tiles grid removed; `SectionShell` title duplicates removed; `← Team` replaces breadcrumb; width aligns with team/settings (Layout maxWidth 1280). Error page upgraded with breadcrumb + status-aware copy + Back / Retry. Old hash URLs (`#ad-prompt` etc.) auto-redirect via `useLegacyAnchorRedirect`.

## Test plan
- [x] `pnpm check && pnpm typecheck` clean
- [x] Backend vitest: full agent-visibility suite (26/26), incl. new admin-can-read-private-agent case
- [x] Frontend manual: admin clicks other member's private agent → fully editable view (no longer 404)
- [ ] Frontend manual: nonexistent UUID → friendly error page with Back-to-Team
- [ ] Frontend manual: tabs switch + URL update + browser back returns to `/team`
- [ ] Frontend manual: cross-tab draft preserved; SaveBar chip jumps to owning tab
- [ ] Frontend manual: `#ad-prompt` legacy hash redirects to Prompt tab
- [ ] Frontend manual: Setup bottom Lifecycle (Suspend / Reactivate / Delete) works
- [ ] Frontend manual: human agent + admin sees Profile + Setup (Lifecycle only); read-only viewer sees Profile only


yuezengwu (approved):
LGTM — 两个 commit 都很干净。

**Server fix**：`requireAgentAccess` 的 `visible` 分支加 `isAdmin` 短路 3 行改动加注释加 test，语义与已有 `manage` 分支和 admin-only `/agents/all` 列表对齐。`resolveCallerInOrg` 保证 caller 必须是 agent 所在 org 的 active 成员，没有跨 org 提权风险。

**Web refactor**：tab + 子路由结构清晰，`layout-context.ts` 把 draft / mutations / dialog state 集中在 layout，tab 组件保持薄。`useLegacyAnchorRedirect` 让老的 `#ad-prompt` 等 hash 链接持续可用，做法合理。错误页升级很实用。

非 blocker 的几个观察供参考：

1. **测试覆盖偏窄**：只新增了 `GET /agents/:uuid` 的 admin 用例。同样走 `visible` 守卫的 `/sessions`、`/client-status`、`/agent-config`、`/avatar` 端点没有显式回归。共享守卫的实现保证了一致性，但补一两条 admin 拉 sessions 或 agent-config 的用例会更稳妥。

2. **`SetupTab` 里的 `ReadOnlyConfigNotice` 实际是 dead code**：`buildTabs` 决定 Setup tab 只在 `canEditConfig || canManageAgent` 时挂载，所以 `!canEditConfig && !canManageAgent` 永远走不到。可以删。

3. **Tab 切换用 `navigate({ replace: true })`** 是有意设计（commit 信息里写了）— 浏览器返回直接跳出 agent 详情而不是切回上一个 tab。如果未来有 PM 期望"返回切回上一个 tab"，需要重新讨论。

4. 前端手动测试清单还没勾，劳烦发布前在本地或预发把列出的 7 条跑一遍，尤其是 admin 看其他成员私有 agent 的 e2e，以及 `#ad-prompt` 这类老链接的重定向。

baixiaohang (commented):
Reviewed both commits.

**Server fix**: clean — the `isAdmin` short-circuit makes the `visible` branch symmetric with `manage` and the admin-only `/agents/all` listing. `resolveCallerInOrg` keeps the check scoped to active members of the agent's org, so no cross-org elevation risk.

**Web refactor**: tab + nested-route structure is clear. Concentrating draft / mutations / dialog state in `layout-context.ts` keeps tabs thin. The legacy anchor redirect is a nice touch and the upgraded error page is a real improvement.

Non-blocking observations / cleanups (most echo @yuezengwu's review):

1. **Dead branch in `SetupTab`** (`packages/web/src/pages/agent-detail/setup-tab.tsx:56-58`): `buildTabs` only mounts the Setup tab when `canEditConfig || canManageAgent`, so the `!canEditConfig && !canManageAgent` path that renders `ReadOnlyConfigNotice` is unreachable. Safe to delete the component along with the branch.

2. **Visibility test coverage is narrow**: the new admin-cross-private case only covers `GET /agents/:uuid`. The sibling endpoints (`/sessions`, `/client-status`, `/agent-config`, avatar) share the same `requireAgentAccess(..., "visible")` guard so they're correct by construction, but one or two regressions on those (e.g. admin pulling sessions or agent-config for another member's private agent) would lock the invariant in.

3. **Consistency follow-up** — `assertAllAgentsVisibleInOrg` (`packages/server/src/scope/require-resource.ts:258`) still has no admin short-circuit. Out of scope for this PR, but chat-create that references another member's private agent will still 404 for admins, which is asymmetric with the detail behavior fixed here. Worth a follow-up issue / TODO.

4. **History-stack inconsistency**: tab clicks use `navigate({ replace: true })` (intentional, per commit msg), but `SaveBar`'s `onJumpTo` (`packages/web/src/pages/agent-detail.tsx:548`) calls `navigate(...)` without `replace`. The two paths to the same tab produce different history entries. Minor.

5. **Manual test checklist** in the PR body is still unchecked — please run through the 7 frontend cases before merging, especially the admin-views-other-member's-private-agent flow and the `#ad-prompt` legacy hash redirect.

6. **CI "Check Branch Name" failure**: branch `hub-session-57e65b19-8e954623` doesn't match the `feat/fix/refactor/...` convention from CLAUDE.md. Not a merge blocker on squash, but worth noting.

Holding off on approve until items 1, 2, 5 are addressed. Item 3 is fine as a separate ticket.

Gandy2025:
Thanks @yuezengwu @baixiaohang. Pushed `cc7ca4b` addressing the actionable items:

**1. Dead branch in `SetupTab`** — removed `ReadOnlyConfigNotice` along with the unreachable `!canEditConfig && !canManageAgent` branch. `SetupTab` is now `canEditConfig && (...) ; canManageAgent && <DangerZone />`.

**2. Visibility test coverage** — added two cases in `agent-visibility.test.ts`:
- `admin can list sessions of a private agent managed by another member` (`GET /agents/:uuid/sessions`)
- `admin can read client-status of a private agent managed by another member` (`GET /agents/:uuid/client-status`)

These pin the visible-guard short-circuit on the sibling endpoints. Suite is now 28/28 ✓.

**5. (@baixiaohang) `SaveBar.onJumpTo` history inconsistency** — now passes `{ replace: true }`, matching `TabsNav`. Both routes to a given tab produce the same single history entry.

**3. `assertAllAgentsVisibleInOrg`** — agreed it's worth a follow-up. Will open a separate issue for the chat-create + admin asymmetry, since fixing it touches a different code path and has its own e2e implications.

**Manual test checklist** — will run through the 7 frontend cases locally (admin viewing other-member's-private-agent + `#ad-prompt` legacy hash redirect being the highest-stakes ones) and tick before merging.

**Branch name** — noted, hub-session-* is auto-generated by the hub agent runtime so we can't easily rename mid-PR. Will keep this in mind for future hub-spawned PRs; for now it's not a merge blocker on squash.

baixiaohang (approved):
Verified the three actionable items in cc7ca4b:

- Dead `ReadOnlyConfigNotice` branch removed; `SetupTab` structure is tight now.
- New `agent-visibility` cases for `/sessions` and `/client-status` lock the shared visible-guard invariant in (28/28 ✓).
- `SaveBar.onJumpTo` now uses `{ replace: true }`, matching `TabsNav`.

Approving. Please run the 7-item manual test checklist (especially admin-views-other-member's-private-agent + `#ad-prompt` legacy hash redirect) before squash-merging, and open the `assertAllAgentsVisibleInOrg` follow-up issue when convenient.

baixiaohang:
Follow-up on item 3 of my earlier review (the `assertAllAgentsVisibleInOrg` asymmetry) — @Gandy2025 opened #426 to track it.

On reflection I'd suggest folding that fix into **this** PR as a third commit rather than a separate PR:

- Same file (`packages/server/src/scope/require-resource.ts`), same symmetry bug — `assertAllAgentsVisibleInOrg` is just the chat-create-side half of the detail-side fix this PR already makes.
- This PR is already a `fix(server) + refactor(web)` split, so a `fix(server): admin short-circuit for assertAllAgentsVisibleInOrg` commit slots in cleanly.
- Splitting it off means a reviewer has to hold both diffs in their head, and the second PR either bases on this one or waits for it to merge — extra overhead for what is logically the second half of the same fix.

If you'd rather keep #426 standalone, fine to land this as-is and stack the next one — just flagging the bundled option is lower overhead.

Gandy2025:
Two follow-up commits pushed:

**`4295271` test(server): repair admin-realtime-role test fallout** — `cc7ca4b` only ran the focused `agent-visibility` suite locally; the full CI run surfaced that `admin-realtime-role.test.ts:172` was relying on the visibility filter as a fallback to 404 a same-org private agent owned by another member. Its setup put both Alice and Bob in the same default org (the comment self-acknowledged the limitation), so the test was never actually exercising the not-a-member-of-org guard its title claimed. Switched it to use `attachMember` so Bob's agent really lives in a foreign org Alice has zero membership in — `resolveCallerInOrg` throws `NotFoundError` before the visibility/admin probe runs, matching the test's stated intent.

**`f38d9d5` fix(server): admin short-circuit for assertAllAgentsVisibleInOrg (closes #426)** — folded in per @baixiaohang's suggestion on #426. One subtlety worth flagging:

Visibility (can see) and ownership (can use) are deliberately separate layers. Admin now passes the visibility gate, but the existing `me-chat` owner-exclusive rule (`services/me-chat.ts:645`, RFC §4.4.2/§4.5) still 403s. The change here makes the admin error **precise (403 owner-exclusive)** instead of the misleading **404 enumeration-style** they were getting before — admins can already list the row via `/agents/all`, so 404 here was false enumeration defense, not real defense.

Two new vitest cases lock the distinction:
- admin → `403` owner-exclusive (visibility passes, ownership rejects)
- non-admin member → `404` visibility (the agent is invisible to them)

Will close #426 when this lands.

CI should be green now (full suite was 1083 passed + 1 failed pre-fix; the 1 failure was the admin-realtime-role test fixed in `4295271`).

yuezengwu (approved):
重新 approve 覆盖 `f38d9d5` + `4295271`。

**`f38d9d5`（`assertAllAgentsVisibleInOrg` admin 短路）**：与 `requireAgentAccess(visible)` 严格对称。最赞的是测试用例把 **403 owner-exclusive vs 404 visibility** 的分层语义钉死——admin 通过 visibility gate，但仍受 `me-chat.ts:645` 的 owner-exclusive 规则约束，从「误导性 404」变成「精确 403」是正确取舍（admin 已能在 `/agents/all` 看到这行，404 不是真防御）。

**`4295271`（admin-realtime-role 测试修复）**：把一个长期靠 visibility filter 兜底的弱测试改造成真正测试 not-a-member-of-org 分支，原测试注释自己都承认了 limitation。属于顺手清理掉的隐性技术债。

baixiaohang (approved):
Re-approving to cover `f38d9d5` + `4295271`.

**`f38d9d5` (`assertAllAgentsVisibleInOrg` admin short-circuit)**: strictly symmetric with `requireAgentAccess(visible)`. The strongest part is the test pair pinning down **403 owner-exclusive vs 404 visibility** as separate layers — admin passes the visibility gate but the `me-chat.ts:645` owner-exclusive rule still 403s. Going from misleading 404 to precise 403 is the right call given admins can already enumerate the row via `/agents/all`.

**`4295271` (admin-realtime-role test repair)**: turns a long-time soft test that was leaning on the visibility filter as a fallback into one that actually exercises the not-a-member-of-org branch — the original comment self-acknowledged the limitation. Nice cleanup of latent test debt while we're here.

baixiaohang (approved):
LGTM on `f38d9d5` — closes item 3 of my earlier review.

**Implementation**: 5 lines mirroring `requireAgentAccess(visible)` exactly. `OrgScope` already carries `role`, so no signature change. Clean.

**Reframing in the commit message** is sharper than my original suggestion — the explicit visibility (admin short-circuit applies) vs ownership (`me-chat.ts:645` owner-exclusive 403 still applies) split is the right model. Verified `me-chat.ts:645` directly: `privateNotOwned` filters on `a.managerId !== caller.managerId` with no admin branch, so admin including another member's private agent does land on the precise `ForbiddenError` after this change. 404→403 isn't a regression on enumeration defense either — `/orgs/:orgId/agents/all` already exposes those rows to admins, so the 404 was never load-bearing here.

**Tests** lock the right invariants — admin: visibility-pass + ownership-reject = 403; non-admin member: visibility-reject = 404 (defense intact for everyone outside admin).

Approving.

PR #441, opened 2026-05-18
fix(web,server): project agent avatar color & image into chat-detail surfaces

yuezengwu:
## Summary

- The admin `GET /chats/:chatId` and both agent-scoped chat endpoints already JOIN-loaded `avatarColorToken` / `avatarImageUpdatedAt` from `agents` but **dropped them when shaping the `participants` payload**. As a result, the chat header chips and right-sidebar Agent rows fell back to the deterministic djb2-hash hue and never showed the manager-uploaded image — while the left-rail conversation row (driven by `/me/chats`) rendered both correctly.
- Project the two avatar fields onto every chat-detail / participants endpoint so the wire shape mirrors `meChatParticipantSchema`. Web consumes the new fields in `chatParticipantById` → `chatScopedAgentIdentity` → `ParticipantsHeader`, and `AgentRow` drops its hand-rolled `<span>` avatar in favour of the shared `<Avatar>` component so an uploaded image actually appears in the right sidebar.
- New contract-level regression test pins all three server routes against the new shape with a manager-set hue token.

## Files

| Layer | Path | Change |
|---|---|---|
| server | `api/chats.ts` | admin `/chats/:chatId` response includes `avatarColorToken` + `avatarImageUrl` |
| server | `api/agent/chats.ts` | agent `/chats/:chatId/participants` response includes the same; `getChatDetail` auto-flows through the existing spread |
| server | `services/chat.ts` | `getChatDetail` + `listChatParticipantsWithNames` SELECT/map carry the new fields |
| server | `__tests__/chat-detail-avatar-fields.test.ts` | **new** — 3 cases asserting wire shape on admin / agent-detail / agent-participants routes with `hue-3 / hue-5 / hue-1` |
| shared | `schemas/chat.ts` | `chatParticipantDetailSchema` extended with `avatarColorToken` + `avatarImageUrl` (both nullable, mirrors `meChatParticipantSchema`) |
| shared / client | 3 test fixtures | schema tightening required `null` defaults in mock participant factories |
| web | `pages/workspace/center/chat-view.tsx` | chip-row + header identity resolver thread `colorToken` through to `<Avatar>` |
| web | `pages/workspace/right-sidebar/agent-row.tsx` | replace hand-rolled colored `<span>` with shared `<Avatar>` so image + manager hue both render; status dot overlay preserved |

## Test plan

- [x] `pnpm check && pnpm typecheck` — biome 748 files clean, all 9 packages typecheck
- [x] `pnpm test` — server 137 file / 1098 tests pass (incl. the 3 new regression cases), client 48 file / 388 tests pass
- [ ] Local visual QA against an isolated dev hub: configure `avatarColorToken` + upload image on an agent, open a chat, verify the title-bar chip and right-sidebar Agents row both render the configured hue / image (matches left-rail row)

## Notes

- Wire-shape change is **additive** (two new nullable fields on `ChatParticipantDetail`); back-compat-safe for any older consumer that doesn't read them.
- `createChat` response shape is pre-existing tech debt (returns raw `chat_membership` rows without `name/displayName/type`); deliberately out of scope here.

### DIFF
--- a/packages/client/src/__tests__/agent-io.test.ts
+++ b/packages/client/src/__tests__/agent-io.test.ts
@@ -20,6 +20,8 @@ function mkParticipant(agentId: string, name: string | null, displayName?: strin
     // back to a synthetic label when the caller doesn't provide one.
     displayName: displayName ?? `agent-${agentId}`,
     type: "autonomous_agent",
+    avatarColorToken: null,
+    avatarImageUrl: null,
   };
 }
 
--- a/packages/client/src/__tests__/chat-context.test.ts
+++ b/packages/client/src/__tests__/chat-context.test.ts
@@ -12,6 +12,8 @@ function mkParticipant(extras: Partial<ChatParticipantDetail>): ChatParticipantD
     name: extras.name ?? "default-name",
     displayName: extras.displayName ?? "Default Name",
     type: extras.type ?? "autonomous_agent",
+    avatarColorToken: extras.avatarColorToken ?? null,
+    avatarImageUrl: extras.avatarImageUrl ?? null,
   };
 }
 
--- a/packages/server/src/__tests__/chat-detail-avatar-fields.test.ts
+++ b/packages/server/src/__tests__/chat-detail-avatar-fields.test.ts
@@ -1,0 +1,103 @@
+import { eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import { agents } from "../db/schema/agents.js";
+import { createChat } from "../services/chat.js";
+import { createTestAgent, useTestApp } from "./helpers.js";
+
+/**
+ * Regression: the chat-detail surface used to drop the manager-configured
+ * `avatarColorToken` / `avatarImageUrl` from the wire shape of every
+ * `/chats/:chatId` route (admin and agent), even though the SELECT pulled
+ * them out of the agents JOIN. The web client (header chips + right-sidebar
+ * agent rows) then fell back to the deterministic djb2-hash hue, so a hue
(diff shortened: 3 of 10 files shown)


baixiaohang (approved):
LGTM — three chat-detail endpoints all carry the new fields, schema mirrors `meChatParticipantSchema`, and the three new server regression cases pin the wire shape per route. Web fallback chain in `chatScopedAgentIdentity` (`p.avatarImageUrl ?? ident?.avatarImageUrl`) correctly preserves the human → GitHub avatar fallback that `services/me-chat.ts:386-394` notes is contractually expected on detail-view surfaces.

Two non-blocking follow-ups, no need to address in this PR:

1. **`AgentRow` doesn't share the same human → GitHub fallback as `ParticipantsHeader`.** `AgentRow` now reads `participant.avatarImageUrl` directly, which uses the agent-only path (`agentAvatarImageUrl`), so for a human agent with no manager upload but a GitHub `users.avatar_url` it falls back to color + initial instead of the GitHub image. Not a regression (the row previously didn't render any image at all), but the `me-chat.ts:386-394` note explicitly says detail-view surfaces should honor that fallback — worth aligning in a follow-up so ParticipantsHeader chips and the right-sidebar row agree.

2. **Sibling endpoints still return a thinned-out participant shape** that doesn't conform to `chatParticipantDetailSchema`: `POST /chats/:chatId/join`, `POST /chats/:chatId/leave`, and the admin add-participants path return `{agentId, role, mode, joinedAt}` only. You called out `createChat` as pre-existing tech debt — these are the same class. Might be worth a tracking issue / TODO so it doesn't get rediscovered next time someone widens the wire shape.

PR #450, opened 2026-05-18
feat(e2e): cross-process E2E framework M1 + F3 GitHub base URL override

yuezengwu:
## Summary

Lands **M0 + M1** of the cross-process E2E framework per proposal
[`hub-local-e2e-framework.20260518.md`](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/hub-local-e2e-framework.20260518.md) (PR#283, merged).
Tested locally end-to-end (docker compose + spawned dist + smoke).

Three commits, each independently reviewable:

1. **`refactor(server)`** — `FIRST_TREE_HUB_GITHUB_API_BASE_URL` env override (F3, ~30 lines). Independently useful for GHE / mirrored deployments; default behaviour unchanged, 1119/1119 server tests pass.
2. **`chore`** — retire `packages/server/vitest.e2e.config.ts` (dead code), tag `e2e-askuser.mjs` / `e2e-auto-agent-add.ts` one-off smokes with migration notes pointing at the new framework, replace legacy `.gitignore tests/e2e/` line with real e2e artifact paths.
3. **`feat(e2e)`** — `packages/e2e/` package: per-run isolated PG (tmpfs + random port), spawned server against built dist, smoke test on `/healthz` + `/api/v1/health`, doctor + clean scripts, `verify-e2e-removable.sh` proves §三 hard constraint.

## What's deferred to M2

- `client start --foreground` spawn (needs credentials-provisioning helper; lifecycle already accepts `withClient: true` flag).
- agent-mock + github-mock fastify, fake claude-code binary spike, chat-send full-loop tests.

## Hard rules verified

- ✅ `private: true`, no `publishConfig` on `packages/e2e/package.json`
- ✅ No entries in `build` / `typecheck` / `test` `dependsOn`
- ✅ `scripts/verify-e2e-removable.sh` proves CLI tarball hash is byte-identical with or without `packages/e2e/`:
  ```
  baseline (no e2e): 31d3129ad3f7d350965fe8dec9db215884788d26f851484ebc31d1721ede157a
  with-e2e:          31d3129ad3f7d350965fe8dec9db215884788d26f851484ebc31d1721ede157a
  ```
- ✅ Biome `noRestrictedImports` enforces both directions of §三.3:
  - business packages can't `import "@first-tree-hub/e2e"`
  - `packages/e2e/**` can't import server/client/command source (only `shared` types)

## Test plan

- [x] `pnpm check` (biome) — clean
- [x] `pnpm typecheck` (11 tasks)
- [x] `pnpm --filter @first-tree-hub/server test` — 1119/1119 pass
- [x] `pnpm --filter @first-tree-hub/server build` + `pnpm --filter @agent-team-foundation/first-tree-hub build` — both succeed
- [x] `pnpm e2e:doctor` — clean
- [x] `pnpm e2e:smoke` — 3/3 in ~7.2s; ran three times back-to-back, no flake
- [x] `pnpm e2e:clean` — leaves no `hub_e2e_*` docker projects
- [x] `bash packages/e2e/scripts/verify-e2e-removable.sh` — CLI hash identical with/without `packages/e2e`
- [x] After teardown: `docker ps -a --filter name=hub_e2e_` empty
- [ ] CI smoke run on `ubuntu-latest` (M3, not in this PR)

## Live-test findings (worth review)

Two bugs caught only by running the framework for real, both documented in `M0-SPIKE.md`:

1. `FIRST_TREE_HUB_PUBLIC_URL` is shared between `server.publicUrl` and `kael.hubPublicUrl`. Setting it activates the optional `kael` block and crashes boot demanding `KAEL_ENDPOINT` / `KAEL_API_KEY`. Fix: don't inject it in `server-process.ts` — dev/test environments can omit it.
2. `client start --foreground` exits with code 1 when there are no credentials. Defers `client` spawn from M1's exit criteria — M2 wires this back in once we have a connect-token helper.

## Proposal corrections (for v5 errata)

- Agent injection env var is `CLAUDE_CODE_EXECUTABLE`, not `FIRST_TREE_HUB_CLAUDE_CODE_EXECUTABLE` as proposal v4 §六.5 / §六.4 wrote.
- `GET /api/v1/admin/clients` doesn't exist (proposal §六.2 / §十.5 F5). M1 smoke uses `/healthz` only; M2 readiness probe will use `/api/v1/me/clients` (user-JWT-gated, requires the same credentials helper that gates client spawn).


baixiaohang (commented):
Reviewed against proposal `hub-local-e2e-framework.20260518.md` v4 (and your M0-SPIKE corrections to v5). Three commits land cleanly, scope matches §九 M1 (pg + server up + smoke, no client / no mocks), and the §三 hard-constraint evidence is in the PR body. A few non-blocking nits below.

## Architecture / hard rules — pass

- ✅ `packages/e2e/package.json` is `private: true`, no `publishConfig`.
- ✅ `turbo.json` keeps `e2e:*` tasks out of `build` / `typecheck` / `test` `dependsOn`. `e2e:smoke` / `e2e:full` correctly `dependsOn: ["^build"]` so the spawned dist is fresh.
- ✅ Biome `noRestrictedImports` blocks both directions of §三.3 (business → e2e and e2e → server/client/command source). Shared types stay reachable, which is what the spec asks for.
- ✅ `verify-e2e-removable.sh` exercises the full §三.2 deletion checklist and the byte-identical hash is in the PR body.

## F3 — GitHub base URL refactor — pass

- New `services/github-api-base.ts` reads `FIRST_TREE_HUB_GITHUB_API_BASE_URL` once at module load, defaults to `https://api.github.com`, strips trailing slashes. Used at 7 call sites across `github-app.ts` / `github-oauth.ts` / `github-entity-live.ts`.
- Existing in-process tests (`__tests__/github-app*.test.ts`) keep asserting against the literal `https://api.github.com/...` strings — that still passes because the default matches; you don't have to touch them.
- Good call leaving the OAuth web endpoints (`https://github.com/login/oauth/*`) on the literal host — they live on a different origin and out-of-scope for the API-base override. Docstring captures this.
- Module-load capture (rather than a `getGithubApiBase()` function) is the right call: e2e sets the env on the spawned child before module evaluation, in-process tests stub `fetch` directly. Worth noting in the file comment that this means `process.env.FIRST_TREE_HUB_GITHUB_API_BASE_URL = ...` after import is a no-op — but it's already implied by "reads … once at module load".

## E2E framework — pass with nits

- ✅ `server-process.ts` race between `waitForHttp(/healthz)` and `child.exit` is correctly wired with an `AbortController` that fires in `finally{}` — no straggler fetches after either branch wins.
- ✅ `killChild` SIGTERM → SIGKILL after 5s. `bestEffortCleanupStaleContainers` matches both `hub_e2e_*` and lossy `hub-e2e-*` shapes via `/^hub[_-]e2e[_-]/i`.
- ✅ `random Bytes(32)` per-run JWT_SECRET / ENCRYPTION_KEY — no fixed secrets in the framework.
- ✅ Logger ring-buffer (`tailSize = 50`) + `dumpTailToConsole` on failure matches §四 contract.
- ✅ Lifecycle is idempotent + SIGINT/SIGTERM hooks installed once.
- ✅ `db-migrate.ts` is a placeholder-with-rationale that points at the server's own `runMigrations` and explicitly documents why we don't run drizzle in parallel (bootstrap-migration-lock invariant). Future migration-upgrade work (§十一.1) has a clear hook.

## Non-blocking findings

### 1. README claim about `turbo.json` excluding `packages/e2e/**` from `inputs` is incorrect

`packages/e2e/README.md:31`:

> `turbo.json` excludes `packages/e2e/**` from the `build` / `typecheck` / `test` cache inputs.

But `turbo.json` doesn't set explicit `inputs` on those tasks — they fall back to turbo's default per-package hashing. That default already isolates `packages/e2e` edits from other packages' cache hashes, so the **outcome** is right. The README is just claiming the wrong mechanism. Either:

- Update README to say "turbo's default per-package hashing keeps `packages/e2e` edits out of other packages' caches", or
- Add the explicit `"inputs": [..., "!packages/e2e/**"]` shown in proposal §三.4.

Proposal §三.4 actually picks the explicit form. Either works; pick one and align the prose.

### 2. `verify-e2e-removable.sh` doesn't stash/restore `pnpm-lock.yaml`

The script stashes `package.json`, `turbo.json`, and `packages/e2e/`. But it then runs `pnpm install --frozen-lockfile=false` twice (once against the stripped tree, once after restore), each of which rewrites `pnpm-lock.yaml`. `restore()` doesn't snapshot the lockfile.

In the happy path the lockfile probably converges back to the original content, but pnpm has been known to reorder workspace metadata. A run could leave the working tree with a dirty `pnpm-lock.yaml`. Suggest:

```bash
cp pnpm-lock.yaml "$STASH_DIR/pnpm-lock.yaml"
# … and in restore():
if [[ -f "$STASH_DIR/pnpm-lock.yaml" ]]; then
  cp "$STASH_DIR/pnpm-lock.yaml" pnpm-lock.yaml
fi
```

Also worth `git diff --quiet pnpm-lock.yaml` as a final assertion so we *catch* any drift in CI.

### 3. `smoke.e2e.test.ts` describe block over-claims

`describe("M1 smoke — server + client come up against real pg", ...)` — but M1 globalSetup passes `{ withClient: false }`, so only the server actually comes up. The README and lifecycle docstring both clearly say client is deferred to M2; the test description is the only place that still says "client". Drop "+ client" from the describe label so the test name reflects what's asserted.

### 4. M0-SPIKE.md proposal corrections — line them up with the v5 errata

`M0-SPIKE.md` and the PR body already flag:

- Real env var is `CLAUDE_CODE_EXECUTABLE` (not `FIRST_TREE_HUB_CLAUDE_CODE_EXECUTABLE` as proposal v4 §六.5 wrote).
- `/api/v1/admin/clients` does not exist; M2 readiness probe will pivot to `/api/v1/me/clients`.

These are tree-side errata, not blocking for this PR — but the PR description already promises "for v5 errata". Worth opening the proposal-update PR alongside merge so future readers of the proposal don't trip over the stale strings.

### 5. Tiny: `packages/e2e/package.json` lists `@first-tree-hub/server` and `@agent-team-foundation/first-tree-hub` as devDependencies

This is fine for getting turbo to order `^build` correctly, and the biome `noRestrictedImports` override prevents any actual source import. No change requested — just calling it out because it looked odd at first glance and the README's "no source imports" hard rule is enforced at lint time, not via dependency surface.

## Live-test findings — well captured

The two `FIRST_TREE_HUB_PUBLIC_URL` / `client --foreground` traps you hit are documented inline in `server-process.ts:1446-1450` and `M0-SPIKE.md` — good place for them.

baixiaohang (approved):
LGTM — 详细 nit 见上一条 comment review。三处非阻塞改动（README turbo inputs 文案、`verify-e2e-removable.sh` 漏 stash `pnpm-lock.yaml`、`smoke.e2e.test.ts` describe 文案）建议后续 PR 顺手带上，不阻塞本次 land。

也别忘了在 first-tree-context 同步开 proposal v5 errata（`CLAUDE_CODE_EXECUTABLE` 命名 / `/api/v1/admin/clients` 不存在）。

yuezengwu:
Reopened as #453 after renaming the branch from `feature/e2e-framework-m0` to `feat/e2e-framework-m0` so it matches the repo's branch-naming convention (the `Check Branch Name` CI gate caught it on this PR). GitHub rename auto-closed this one because head-branch rename — unlike base-branch rename — doesn't retarget open PRs.

Continuing review on #453. Same commits, same diff.

PR #466, opened 2026-05-19
feat(e2e): drive dev-user seed in e2e:up; switch docker-pg to docker run

yuezengwu:
## Summary

- `pnpm e2e:up` now provisions a **dev-user** session through the real public HTTP API end-to-end, so a human signing into the web app lands on a workspace with one autonomous agent + one chat (1 message) — no onboarding wizard.
- `framework/docker-pg.ts` switches from `docker compose up` to `docker run` to work around an intermittent ECONNRESET on Docker Desktop macOS port-forward into compose-managed containers.
- Cleanup helper extended with a second sweep for standalone `hub_e2e_*_pg` containers from crashed runs (legacy compose-project sweep retained for backward compat).

## Dev-user seed (in `framework/setup-devuser.ts`)

Walked in this exact order, no PG direct writes during setup:

1. `GET /api/v1/auth/github/dev-callback?githubId=1&login=devuser` → access/refresh JWT in 302 fragment _(equivalent of clicking the web "Continue as Dev User" button)_
2. `POST /api/v1/me/connect-tokens` → short-lived connect-token _(equivalent of the web "Add CLI" button)_
3. `POST /api/v1/auth/connect-token` → user JWT pair _(the CLI's own `exchangeToken` step)_
4. Write `credentials.json` + `client.yaml` under `${runHome}-devuser/config`
5. Spawn `client start --foreground --no-interactive` — WS register + capabilities probe + PATCH `/clients/:id/capabilities`
6. `POST /api/v1/orgs/:org/agents` (type=`autonomous_agent`, `clientId=<devClient>`)
7. `POST /api/v1/orgs/:org/chats` (1:1 with the new agent)
8. `POST /api/v1/chats/:chatId/messages`

`setup-devuser.ts` codifies two non-obvious constraints inline so they're hard to lose:

- **`client start --foreground`, not `connect --no-service`** — only the former probes local runtime SDKs and PATCHes capabilities. Without that PATCH, the web onboarding step 2 shows "no runtime ready" and the user is stuck.
- **Strip ambient `FIRST_TREE_HUB_*` env when spawning the dist CLI** — the client config resolver gives env > file priority, so when the harness itself runs inside an agent runtime, the parent's `FIRST_TREE_HUB_SERVER_URL` would silently override `client.yaml` and the WS auth fails with a JWT-secret mismatch.

## Why `docker run` instead of `docker compose`

Docker Desktop on macOS reproduces an intermittent ECONNRESET on port-forwarded TCP into compose-managed PG containers. Default user-defined bridge and explicit `network_mode: bridge` both reproduce; same image via plain `docker run` is stable on the same machine. Compose was only buying us "ports + healthcheck + naming" which we now do directly. The deleted `scripts/compose.e2e.yml` and the dropped `DockerPgOptions.composeBin` field are the dead pieces this rip-out leaves behind.

## File map

| File | Change |
|---|---|
| `packages/e2e/src/framework/setup-devuser.ts` | **New.** 8-step dev-user seed |
| `packages/e2e/src/scripts/up.ts` | Call `setupDevUser`; inject `FIRST_TREE_HUB_DEV_CALLBACK_ENABLED=1`; print dev-user identifiers + web sign-in hint |
| `packages/e2e/src/framework/docker-pg.ts` | `docker compose up` → `docker run`; add second cleanup sweep for standalone containers; drop unused `composeBin` field |
| `packages/e2e/src/framework/lifecycle.ts` | Stop passing `composeBin` to `startDockerPg` |
| `packages/e2e/scripts/compose.e2e.yml` | **Deleted** (no longer read by docker-pg) |
| `packages/e2e/README.md` | Layout section reflects new files / removed compose file |

## Context Tree

Six-module abstraction + dev-user seed contract pinned in [agent-team-foundation/first-tree-context#293](https://github.com/agent-team-foundation/first-tree-context/pull/293).

## Test plan

- [x] `pnpm exec turbo run typecheck --filter=@first-tree-hub/e2e`
- [x] `pnpm exec biome check packages/e2e`
- [x] `E2E_WITH_CLIENT=1 pnpm e2e:full` — 15/15 tests pass (smoke + messaging + github-webhook + github-pr-delivery + ws-inbox-push + agent-runtime)
- [x] `E2E_WITH_CLIENT=1 pnpm e2e:up` lands on the documented terminal state on clean macOS Docker Desktop, then `Ctrl-C` cleans up containers
- [ ] Reviewer to confirm the docker engine 29.x ECONNRESET observation reproduces on their machine, or that the `docker run` path stays stable for them too


baixiaohang:
Solid PR — clean direction, design-doc-aligned (context#293), and the in-source comments around the two non-obvious constraints (`client start --foreground` vs. `connect --no-service`; strip ambient `FIRST_TREE_HUB_*` env) are exactly what future readers will need. End-to-end HTTP wiring lines up with the actual server routes / schemas I cross-checked (`/me/connect-tokens`, `/auth/connect-token`, `POST /orgs/:orgId/agents` → 201 `{ uuid, ... }`, `POST /orgs/:orgId/chats` → 201 `{ chatId }`, `POST /chats/:chatId/messages` → 201 `{ id, ... }`).

Two concrete things worth fixing before merge, plus a handful of non-blocking notes.

### ⚠️ Pre-merge

1. **Cleanup sweep #2 filter is over-broad.** In `docker-pg.ts:143`:

   ```ts
   ["ps", "-a", "--filter", "name=^hub_e2e_", "--filter", "name=_pg$", ...]
   ```

   `docker ps` joins multiple `--filter` of the same key with **OR**, not AND. So this matches containers whose name regex-matches `^hub_e2e_` **OR** matches `_pg$` — anything ending in `_pg` from any unrelated project on the dev machine becomes collateral. Easy fix: a single combined regex.

   ```ts
   ["ps", "-a", "--filter", "name=^hub_e2e_.*_pg$", "--format", "{{.Names}}"]
   ```

2. **tmpfs storage was lost in the compose → docker run switch.** The deleted `compose.e2e.yml` had `tmpfs: /var/lib/postgresql/data`, with the explicit comment *"each run starts from an empty cluster; crashed runs leave nothing behind on disk."* The new `docker run` invocation doesn't pass `--tmpfs`, so:
   - every fsync now hits the host disk via the storage driver (slower, especially for migrations on cold runs), and
   - crashed runs leave PG data inside the container's writable layer until sweep #2 reaps it.

   One-liner fix in the run args:

   ```ts
   "--tmpfs", "/var/lib/postgresql/data:rw",
   ```

   (Worth pinning explicitly because the PR description calls out "ports + healthcheck + naming" as the compose features being moved over, but tmpfs was actually a third feature and silently dropped.)

### Non-blocking

3. **Double `docker inspect` after the health-poll loop** (`docker-pg.ts:78–95`). The post-loop inspect runs unconditionally and re-fetches the same status the loop just observed. Could collapse into a single `lastStatus` variable inside the loop — minor.

4. **Two competing SIGINT handlers in `up.ts`.** `startRunWorld` internally registers `process.once("SIGINT", …)` → `stopRunWorld().finally(() => process.exit(130))`, and `up.ts` then registers its own SIGINT handler that calls `devSession.stop()` before `stopRunWorld()`. On Ctrl-C both fire; `process.exit(130)` from the lifecycle handler can pre-empt the devSession teardown. In practice the spawned CLI child dies via parent exit anyway, but the ordering is racy and the `console.error("\nReceived ${signal}, shutting down…")` will fire twice. Cleaner: expose an `onTeardown` hook on `lifecycle` so `up.ts` doesn't have to register a parallel handler.

5. **`setup-devuser.ts` straddles modules.** It mixes M5 (identity provisioning), M1 (long-running child), M3 (HTTP driver calls) and M2 (PG reads). Acceptable as a single dev-user seed driver, but worth a one-line `README.md` pointer noting that this file is the "M5 dev-user seed" (vs. `credentials.ts` which is the M5 fixture single-user). Otherwise readers will look for it under one of the per-module names later.

6. **The fragment-parse only checks `access`, not `refresh`.** `setup-devuser.ts:81` reads `frag.get("access")` but credentials.json on line 115 also writes `refreshToken`. If the dev-callback fragment ever omitted `refresh`, this would silently land `undefined` in `refreshToken` instead of failing fast. Worth asserting both fragments are present at the same `if` site.

7. **`devSession.firstMessageId` is collected but never surfaced.** It's on the returned `DevUserSession` but `up.ts` doesn't print it. Either drop the field or print `devMsg: …` alongside `devChat: …` so testers can `curl` it directly.

### Re: the open question in the PR description

> Reviewer to confirm the docker engine 29.x ECONNRESET observation reproduces on their machine, or that the `docker run` path stays stable for them too.

I can't repro on macOS Docker Desktop right now, but the rationale is plausible — port-forward through compose-managed bridge networks has had this exact failure mode reported upstream more than once. Given that compose was only buying us ports/healthcheck/naming and the standalone path is stable for you, switching is the right call. Just please bring tmpfs along (item 2).

baixiaohang (changes requested):
Requesting changes on two items from my earlier comment (#issuecomment-4485803994) that have real side effects despite being one-line fixes:

**Must-fix before merge**

1. **Cleanup sweep #2 filter** (`docker-pg.ts:143`) — `docker ps --filter name=^hub_e2e_ --filter name=_pg$` OR-joins the two name filters, so anything on the dev box whose name ends in `_pg` becomes a sweep target. Collapse into a single regex `name=^hub_e2e_.*_pg$`.

2. **Restore tmpfs storage** — the deleted `compose.e2e.yml` had `tmpfs: /var/lib/postgresql/data` with the explicit "each run starts on an empty cluster; crashed runs leave nothing behind on disk" rationale. The new `docker run` invocation lost that. Add `"--tmpfs", "/var/lib/postgresql/data:rw"` to `runArgs`.

The other 5 items in my prior comment are non-blocking — happy to leave them as follow-ups.

yuezengwu:
Thanks for the careful read — all 7 items addressed in 371109a. Took the chance to ship the 5 non-blocking nits in the same commit since they were all small and worth not leaving as drift.

### Must-fix

1. **Cleanup sweep #2 filter** → collapsed to a single regex `name=^hub_e2e_.*_pg$`. Comment in the code now spells out why (so the next person doesn't undo it on a "this looks weird, multiple filters are clearer" refactor).

2. **tmpfs restored** → `--tmpfs /var/lib/postgresql/data:rw` is back on the `docker run` invocation. Worth noting: I re-validated tmpfs after Docker Desktop restart and it ran 3-for-3 clean. The ECONNRESET I'd attributed to tmpfs during the compose investigation was actually a transient Docker Desktop networking fault, not a tmpfs problem — so this puts us cleanly back on tmpfs without losing the compose → docker run move. The module doc-comment now lists tmpfs alongside ports/healthcheck/naming as a ported feature (so it doesn't get silently dropped again).

### Non-blocking

3. **Double inspect** → collapsed into a single `lastStatus` variable updated inside the poll loop. The error path now reports the last observed status verbatim (e.g. `last status=unhealthy` or `inspect-exit-1`), which I think is a slight readability win for the failure case too.

4. **Competing SIGINT handlers** → added `lifecycle.registerPreTeardownHook(fn)`. `up.ts` drops its own `process.once("SIGINT")` and registers `devSession.stop()` through the hook list instead. Lifecycle's handler now runs all pre-teardown hooks → `stopRunWorld()` → `process.exit(130)` in order. Verified on Ctrl-C: one `[lifecycle] received SIGINT, tearing down e2e world` log, no shutdown-message double-print, devuser CLI takes its clean shutdown path before server/pg disappear under it.

5. **README pointer** → `setup-devuser.ts` line now says "M5 dev-user seed: dev-callback → connect-token → CLI client start → agent/chat/message (driven from up.ts, not tests)". Paired with the existing `credentials.ts` line which now reads "M5 fixture: single-user PG seed for tests". Either entry point now signposts the other.

6. **Fragment parse asserts both fields** → guarded `!devAccess || !devRefresh` with a single combined error message and an inline note explaining why refresh is checked even though we don't use it (catches upstream contract regressions). Refresh stays unused for now since the connect-token step mints a fresh pair downstream.

7. **`devMsg` printed** → added `devMsg: <uuid>` line in the e2e:up output right under `devChat:`. Drop-in `curl` target from the terminal.

### Verification

- `pnpm exec biome check packages/e2e` ✓
- `pnpm exec turbo run typecheck --filter=@first-tree-hub/e2e` ✓
- `E2E_WITH_CLIENT=1 pnpm e2e:full` — 15/15 ✓
- `E2E_WITH_CLIENT=1 pnpm e2e:up` → SIGINT verified to produce one teardown log line and leave zero containers behind ✓
- `--tmpfs` re-verified across 3 cold cold-start runs ✓

Ready for re-review.

baixiaohang (approved):
All 7 items from the prior review are addressed in 371109a, both must-fix items cleanly with defensive comments to prevent regression. CI green, design-doc-aligned. Thanks for the thorough turnaround.

PR #473, opened 2026-05-20
chore: merge npm release into publish workflow

bestony:
## Summary
- move stable/latest npm publishing into the trusted `npm-publish.yml` workflow
- keep alpha publishing on CI success / manual dispatch only
- delete the legacy `npm-release.yml` token-based workflow

## Validation
- `actionlint .github/workflows/npm-publish.yml`
- `git diff --check`
- `pnpm check`
- `pnpm typecheck`
- `pnpm test`


baixiaohang (approved):
LGTM — clean consolidation. The headline win is dropping `NPM_TOKEN` in favor of trusted publishing for the stable channel, which matches what alpha already does.

A few notes (all non-blocking):

**Nice touches**
- Tightening the alpha job guard to `(github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')` is the right move now that tag pushes share this workflow — `workflow_run.conclusion` is undefined on tag pushes, so the old expression was implicitly relying on falsy-undefined.
- `cancel-in-progress: ${{ github.event_name != 'push' }}` preserves the old `npm-release.yml` "never cancel a tag publish" behavior while keeping alpha-on-main serialized. Good.
- `publish-command-latest` belt-and-suspenders: `on.push.tags: ["v*"]` plus `if: github.event_name == 'push' && github.ref_type == 'tag'`.

**Worth double-checking before merge (platform-side, not code)**
1. **Branch protection required checks** — the alpha job is now displayed as `Publish Command Alpha` / `Publish Shared Alpha`. If any required status check is pinned to the old `Publish Command` / `Publish Shared` names, it will block merges until updated.
2. **npm trusted publisher config** — alpha already publishes via OIDC from `npm-publish.yml`, so the publisher binding exists; just confirm the same workflow is authorized for the `latest` dist-tag on `@agent-team-foundation/first-tree-hub` so the first tag push doesn't 401.

**Minor heads-up**
- Stable publish channel quietly moves from Node 22 → 24 (aligning with alpha and the repo's "recommended 24"). Worth a line in the PR body for the changelog trail, but not a blocker.

PR #458, opened 2026-05-18
feat(chat): clickable bare .md paths + restore side-by-side doc preview rail

Gandy2025:
## Summary

Two related changes that together fix the cloud-architecture preview UX introduced by #415's snapshot pipeline:

1. **Bare `.md` paths in agent messages are now clickable previews.** Both runtime and web share `scanBareDocPathTokens` (new `packages/shared/src/lib/doc-link-scan.ts`); runtime uses it to build snapshots, web uses it to wrap matches into `[path](path)` before render. Tokens are canonicalised through `normalizeDocLinkPath` before wrapping so hidden / escape paths stay plain text instead of producing dead anchors. `docPreviewPathFromHref` accepts `:line[:col]` suffixes so links like `docs/intro.md:7` resolve.

2. **Doc preview drawer goes back to the PR #356 side-by-side rail.** PR #432's floating panel overlaid chat and forced manual repositioning. Restore the `relative shrink-0` flex-sibling form (same height as chat-view, left-edge col-resize, width-only persisted, header sized + tinted to match chat-view's). Conflict with `ChatRightSidebar` resolved by mutual exclusion in chat-view: opening doc preview auto-collapses the sidebar (ref-stashed); closing restores. The header sidebar icon swaps rails when doc preview is open. `RESERVED_MAIN_WIDTH = 640` reflects the actual fixed 320 conversation rail + 320 minimum chat column so the resize cap can't squeeze chat below readable width.

Background: #356 and #433 (codex) tried a path-based preview that needed the server to read the local agent workspace; #415 replaced it with snapshot-embedded metadata (cloud-friendly). This PR fills the bare-path gap and restores the rail UX without breaking the snapshot pipeline.

## Test plan

- [x] `pnpm check` clean
- [x] `pnpm typecheck` 9/9 packages
- [x] `pnpm --filter shared test` 251/251
- [x] `pnpm --filter web test` 191/191 (new linkify guard + scanner edge cases)
- [x] `pnpm --filter client exec vitest run src/__tests__/result-sink.test.ts` 21/21 (new bare-path snapshot + `:line:col` strip)
- [x] Manual: dev-stack with SQL-injected agent message containing bare paths, `:line` suffix, inline link, hidden/escape forms — bare paths linkify, click opens snapshot drawer with no server round-trip, hidden/escape forms stay plain text, drawer mutually excludes with ChatRightSidebar, closing restores sidebar
- [ ] Reviewer to confirm cloud preview path on staging

[...]

PR #474, opened 2026-05-20
fix(chat): doc-preview works for all workspaces + snapshot-gated timeline links

Gandy2025:
## Summary

Follow-up to #458. The clickable-doc-path / doc-preview feature only worked when a chat's workspace had **exactly one git repo**, and even then was **broken in the cloud topology**. This fixes both the root cause and a related timeline-link UX bug, so any `.md` an agent actually generates in its workspace renders as a clickable inline link that opens the preview rail — for **every** workspace shape, cloud included.

### 1. Base-path resolution never covered most workspaces (runtime)
`documentBasePathFromRuntimeConfig` returned `null` when `gitRepos.length !== 1`. With a null base, `result-sink` attaches **no `documentContext` at all** → no snapshot → a doc written in the workspace can never be previewed. That's every 0-repo or multi-repo workspace.

And the single-repo branch returned a bare **relative** `localPath` (e.g. `"first-tree-hub"`). The runtime resolves that against its own `process.cwd()` (the long-lived client launch dir, **not** the per-chat workspace — the runtime never `chdir`s per chat), so the file isn't found → snapshot skipped → falls back to `kind:"path"`, which only works on single-host where the **server** can read the path. ⇒ **single-repo cloud preview was dead.** (Verified by exercising the real `buildMessageDocumentSnapshots`: relative base resolves only when `process.cwd()` happens to be the per-chat workspace; absolute base always resolves.)

**Fix:** `resolveDocumentBasePath(log, chatId)` now computes the per-chat root as `join(handlerConfig.workspaceRoot, chatId)` — a **pure `join`**, deliberately *not* `acquireWorkspace` (whose `mkdir`/`rmSync` side effects must not run on every outbound message). `documentBasePathFromRuntimeConfig(payload, perChatRoot)` **never returns null**:
- exactly one repo → **absolute** `<perChatRoot>/<localPath>` (the worktree is materialised there)
- zero / multiple repos → the per-chat workspace root

Resolution stays safe: `resolveWorkspaceFile` existence-checks each candidate inside the root, so a bare mention that doesn't physically exist just stays plain text instead of mis-resolving.

### 2. Timeline links were over-eager + a top-level `:line` href bug (web)
`linkifyMarkdownDocPaths` linkified **any** bare `.md`-looking token in agent prose. Two problems:
- **False-positive / dead links.** An agent merely *talking about* a filename (e.g. "the `README.md:12` bug") got turned into a clickable link with no backing doc — opening an empty preview, or worse an unrelated workspace file with the same name.
- **`README.md:12` reloaded the page.** The link target was the raw token, so `README.md:12` (no `/` before the `:`) hit react-markdown's `defaultUrlTransform`, which treats `README.md:` as an unknown URL scheme and strips the href to `""`. Clicking an `<a href="">` navigated to the current page = full reload.

**Fix:** linkify now takes the set of snapshot paths the message actually carries and **only linkifies a token whose canonical path has a matching snapshot**, with the link target set to the **canonical (de-suffixed)** path. So:
- a mention with no snapshot stays plain text (no dead links),
- `README.md:12` renders `[README.md:12](README.md)` — visible text keeps the line number, href is clean and survives `defaultUrlTransform`.

`chat-view` passes `new Set(docSnapshots?.keys())` into linkify (snapshots are already memoized from `msg.metadata`).

## Resulting UX

An agent generates `hub-end-to-end-architecture.md` (or any `.md`) in its workspace and mentions it → the filename renders as an inline link in the timeline → click slides out the side-by-side preview rail with the content, **no page reload, no server round-trip, cloud-friendly**. Works for top-level and subdir files, with or without `:line[:col]` suffixes, across single-repo / multi-repo / no-repo workspaces. Filenames the agent only *mentions* (no real doc) stay plain text.

## Test plan
- [x] `pnpm --filter client test` — 389/389 (incl. new `documentBasePathFromRuntimeConfig` cases: zero/multi/single-repo **absolute** path, blank localPath fallback)
- [x] `pnpm --filter web test` — 192/192 (linkify snapshot-gating: dead-mention stays plain text; `README.md:12` → canonical href; existing skip rules)
- [x] `pnpm --filter client --filter web typecheck` clean
- [x] biome check clean on changed files

## Known follow-up (separate PR)
With the base path now always set, every agent message carries `documentContext` — `kind:"path"` (with the local absolute path) when no doc resolves. Snapshot-gating already removes its effect on timeline links; tightening or omitting `kind:"path"` in the cloud topology (to avoid embedding a local absolute path in immutable history) is left to a follow-up.

### DIFF
--- a/packages/client/src/__tests__/document-base-path.test.ts
+++ b/packages/client/src/__tests__/document-base-path.test.ts
@@ -1,0 +1,43 @@
+import type { AgentRuntimeConfigPayload } from "@agent-team-foundation/first-tree-hub-shared";
+import { describe, expect, it } from "vitest";
+import { documentBasePathFromRuntimeConfig } from "../runtime/session-manager.js";
+
+function payload(gitRepos: AgentRuntimeConfigPayload["gitRepos"]): AgentRuntimeConfigPayload {
+  return { kind: "claude-code", prompt: { append: "" }, model: "", mcpServers: [], env: [], gitRepos };
+}
+
+const PER_CHAT = "/data/workspaces/agent/chat-123";
+
+describe("documentBasePathFromRuntimeConfig", () => {
+  it("returns the per-chat workspace root when there are zero repos", () => {
+    expect(documentBasePathFromRuntimeConfig(payload([]), PER_CHAT)).toBe(PER_CHAT);
+  });
+
+  it("returns the per-chat workspace root when there are multiple repos", () => {
+    const result = documentBasePathFromRuntimeConfig(
+      payload([{ url: "https://github.com/a/one.git" }, { url: "https://github.com/a/two.git" }]),
+      PER_CHAT,
+    );
+    expect(result).toBe(PER_CHAT);
+  });
+
+  it("returns an ABSOLUTE repo worktree path (perChatRoot + derived localPath) for a single repo", () => {
+    // Regression: the old code returned a bare relative localPath
+    // ("first-tree-hub"), which the runtime resolved against its own
+    // process.cwd() (the launch dir, not the per-chat workspace) and failed to
+    // find any doc — leaving single-repo cloud preview dead.
+    expect(
+      documentBasePathFromRuntimeConfig(payload([{ url: "https://github.com/a/first-tree-hub.git" }]), PER_CHAT),
+    ).toBe(`${PER_CHAT}/first-tree-hub`);
+  });
+
+  it("honours an explicit localPath for a single repo", () => {
+    expect(
+      documentBasePathFromRuntimeConfig(payload([{ url: "https://x/y.git", localPath: "nested/repo" }]), PER_CHAT),
+    ).toBe(`${PER_CHAT}/nested/repo`);
+  });
+
+  it("falls back to the per-chat root when a single repo's localPath is blank", () => {
+    expect(documentBasePathFromRuntimeConfig(payload([{ url: "", localPath: "   " }]), PER_CHAT)).toBe(PER_CHAT);
+  });
+});
--- a/packages/client/src/runtime/session-manager.ts
+++ b/packages/client/src/runtime/session-manager.ts
@@ -1,3 +1,4 @@
+import { join } from "node:path";
 import type {
   AgentRuntimeConfigPayload,
   GitRepo,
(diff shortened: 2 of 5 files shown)


yuezengwu (approved):
LGTM. Both root causes are sharply diagnosed and the fixes are minimal + well-scoped.

### Why this is correct
- `documentBasePathFromRuntimeConfig` returning `<perChatRoot>/<localPath>` is the right base. Worktrees are materialised at `resolveGitRepoTargetPath(cwd, localPath)` (`handlers/claude-code.ts:848`, integration test at `gitrepos-session-integration.test.ts:137`), so the absolute join exactly matches where the agent writes files. Old behaviour (relative `localPath` resolved against runtime `process.cwd()`) was indeed broken under cloud topology — verified.
- `resolveDocumentBasePath` recomputes `perChatRoot` from `handlerConfig.workspaceRoot` + `chatId` via pure `join`, deliberately not `acquireWorkspace`. Important: `acquireWorkspace` has `mkdir` + boundary-marker side effects (and an `rmSync` path on stale-marker recovery) — calling it on every outbound message would be a real foot-gun. Good call-out in the inline comment.
- By the time the result-sink calls `getDocumentBasePath`, both claude-code and codex handlers have already run `acquireWorkspace`, so `safeRealpath(root)` in `buildMessageDocumentSnapshots` will resolve — no race window.
- Web side: snapshot-gated linkify is the correct invariant. Pairing visible-text `[README.md:12]` with canonical href `(README.md)` is a clean fix for the `defaultUrlTransform` reload bug while preserving the line-number UX. The `Set(docSnapshots?.keys() ?? [])` deps in the `useMemo` are stable because `docSnapshots` itself is memoised on `msg.metadata`.

### Non-blocking observations
1. `resolveDocumentBasePath` is now total — its declared return type `Promise<string | null>` could tighten to `Promise<string>`, and the `if (documentBasePath)` guard in `result-sink.ts:62` becomes dead. Not a blocker; worth a follow-up cleanup so future readers don't model a non-existent null branch.
2. As you already flagged: every agent message now carries `documentContext`, falling back to `kind:"path"` with a local absolute `basePath`. Snapshot-gating means dead timeline links don't render, but embedding a local agent-host path in immutable chat history is still a slight cloud-topology smell. Tightening or omitting `kind:"path"` in the cloud topology in a follow-up is the right move — fine to ship separately.
3. Minor: in `linkifyMarkdownDocPaths` the early-return `if (snapshotPaths.size === 0)` saves a regex scan but means hub-UI messages (which don't carry snapshots and skip linkify upstream anyway) and agent messages with no resolvable docs share the same fast path. No issue, just noting the implicit contract: empty set == nothing to linkify.

Tests look thorough — the document-base-path cases cover zero / multi / single-with-absolute / explicit-localPath / blank-localPath, and the linkify dead-mention case is the exact regression this PR exists to prevent.

Approving — please ship after CI clears.

PR #476, opened 2026-05-20
chore(chat): stop emitting kind:"path" documentContext (drop local-path leak)

### LINKED ISSUE #475: doc-preview: tighten/omit kind:"path" so agent messages don't embed a local absolute path
Follow-up from #474.

## Context
#474 makes `resolveDocumentBasePath` never return null. As a side effect, **every** agent message now goes through `buildMessageDocumentSnapshots`, and when no `.md` resolves, `result-sink` writes `metadata.documentContext = { kind: "path", basePath }` where `basePath` is the **local absolute** per-chat workspace path (`<dataDir>/workspaces/<agentName>/<chatId>`).

## Problems
1. **Info leak**: that absolute server-side path lands in immutable message history, readable by every chat participant (reveals dir layout / agent name).
2. **Dead in cloud**: the `kind:"path"` legacy preview (`/me/docs/preview`) requires the server to read the agent's local disk — impossible in the cloud topology. So it's metadata bloat with no payoff there.

Note: #474's snapshot-gating means timeline links already only render for paths that have a real snapshot, so they no longer depend on `kind:"path".basePath`. The remaining consumer is agent-authored **explicit** inline links to non-snapshotted files on single-host deployments.

## Options
- Omit `documentContext` entirely when there are 0 snapshots (simplest; loses the single-host explicit-link path fallback — which mostly overlaps with snapshotting anyway).
- Keep `kind:"path"` only when the deployment is single-host (server can read the workspace), and never embed an absolute path otherwise.
- Replace the absolute `basePath` with an opaque/relative token the server resolves.

## Acceptance
- No local absolute filesystem path embedded in message metadata in the cloud topology.
- Single-host explicit-inline-link preview behaviour intentionally decided (kept or dropped), with a test.

Gandy2025:
Closes #475. Follow-up to #474.

## Why
#474 made the doc-preview base path resolve for every workspace shape. A side effect: **every** agent message with no resolvable `.md` now fell back to `kind:"path"` and embedded the agent host's local **absolute** workspace path (`<dataDir>/workspaces/<agentName>/<chatId>`) into immutable chat history.

That path is:
- a cloud-topology smell — it leaks the runtime host's directory layout / agent name to every chat participant, permanently;
- functionally dead in cloud — the server can't read the agent's local disk, so the legacy `/me/docs/preview` path variant can't resolve it anyway.

Since #474, any real, *referenced* `.md` (inline link or bare mention) that exists is already captured as a **snapshot**, so the `kind:"path"` fallback only ever applied to docs that can't be previewed regardless (non-existent / hidden / oversized).

## Change
`result-sink` now embeds **only** the inline-snapshot variant. A message with no resolvable doc carries **no** `documentContext`.

Back-compat preserved: historical messages may still hold `kind:"path"`, and both the web reader (`documentBasePathFromMetadata`) and the server schema/endpoint keep handling them — this only stops *emitting* new ones.

Also tightens `resolveDocumentBasePath` to `Promise<string>` (it is total after #474 — never returns null), per @yuezengwu's review note on #474.

## Test plan
- [x] `pnpm --filter client test` — 389/389. The result-sink "no resolvable doc" cases (relative base, hidden/dotfile, symlink-escape, external-href, escaped/image link) now assert **no** `documentContext` instead of a `kind:"path"` fallback.
- [x] Server `kind:"path"` schema + `/me/docs/preview` tests unchanged (reading old messages still works).
- [x] `pnpm --filter client typecheck` clean
- [x] biome check clean on changed files

### DIFF
--- a/packages/client/src/__tests__/result-sink.test.ts
+++ b/packages/client/src/__tests__/result-sink.test.ts
@@ -93,15 +93,13 @@ describe("createResultSink — forwardResult enrichment", () => {
     expect(body.purpose).toBe("agent-final-text");
   });
 
-  it("case 1.5 (documentContext regression): writes path-variant metadata.documentContext + omits metadata.mentions", async () => {
-    // v1.5 regression guard — PR #356's documentContext injection must
-    // survive改造 4 surgery. Verifying the branch is still wired AND that
-    // it is NOT accompanied by an automatic mentions array.
-    //
+  it("case 1.5: a base path that resolves no docs attaches NO documentContext (no kind:path leak)", async () => {
     // The basePath here is a relative string that won't resolve to a real
-    // worktree on disk, so the snapshot scan returns no docs and the sink
-    // falls back to the legacy path variant — exactly what we want for
-    // single-host / shared-fs deployments.
+    // worktree on disk, so the snapshot scan returns no docs. We no longer
+    // fall back to the legacy `kind:"path"` variant — it would embed the
+    // agent host's local absolute path in immutable history and is dead in
+    // the cloud topology. A message with no resolvable doc carries no
+    // documentContext at all (and still no auto-mentions array).
     const { sink, sendMessage } = buildSink({
       trigger: { messageId: "m1", senderId: "agent-peer" },
       getDocumentBasePath: vi.fn().mockResolvedValue("first-tree-hub"),
@@ -110,9 +108,9 @@ describe("createResultSink — forwardResult enrichment", () => {
     await sink("see [design](docs/design.md)");
 
     const body = sendMessage.mock.calls[0]?.[1] as {
-      metadata?: { documentContext?: { kind?: string; basePath?: string }; mentions?: unknown };
+      metadata?: { documentContext?: unknown; mentions?: unknown };
     };
-    expect(body.metadata?.documentContext).toEqual({ kind: "path", basePath: "first-tree-hub" });
+    expect(body.metadata?.documentContext).toBeUndefined();
     expect(body.metadata?.mentions).toBeUndefined();
   });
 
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
LGTM. 是一个聚焦的隐私 / cloud-topology 清理：

- 阻断把 agent 本地绝对工作区路径写入不可变历史的泄露面。
- 行为差异零回退 —— 自 #474 起，可 preview 的 `.md` 已走 snapshot；`kind:"path"` fallback 实际只覆盖被拒绝 / 不存在 / 隐藏 / 超大的文档，那部分 path 即便保留也无法 preview。
- 服务器 schema 与 `/me/docs/preview` 端点仍然兼容历史 `kind:"path"` 消息，只是 emitter 不再产生 —— back-compat 保留。
- `resolveDocumentBasePath` 收紧为 `Promise<string>`，跟 #474 review 留下的 TODO 对齐。

`result-sink.ts:62` 的 `if (documentBasePath)` 在收紧返回类型后等价于 `if (path !== "")`，逻辑不变；保留无碍，非 blocker。

baixiaohang (approved):
LGTM — 聚焦的隐私 / cloud-topology 清理。

- 阻断把 agent 本地绝对工作区路径写入不可变历史的泄露面；
- 行为差异零回退：自 #474 起可 preview 的 `.md` 已走 snapshot，`kind:"path"` fallback 实际只覆盖被拒 / 不存在 / 隐藏 / 超大的文档，那部分 path 即便保留 server 也无法 preview；
- 服务器 schema 与 `/me/docs/preview` 端点未变 —— back-compat 保留，老消息仍可读，只是 emitter 不再产生；
- `resolveDocumentBasePath` 收紧为 `Promise<string>`，呼应 #474 review 留的 TODO；
- 测试更新彻底，每个 "无可解析 doc" 用例都从断言 `kind:"path"` 改为断言 `undefined`。

Nit（非 blocker，可后续顺手清）：`ResultSinkDeps.getDocumentBasePath` 仍声明为 `() => Promise<string | null>`（`result-sink.ts:54`），实现已经是 `Promise<string>`；同步收紧后 `result-sink.ts:62` 的 `if (documentBasePath)` 退化为 `if (path !== "")`，可以一并清掉。

PR #482, opened 2026-05-20
fix(client): Context Tree SSH fallback + tolerate inbox enum drift

baixiaohang:
## Summary

Two robustness fixes surfaced in the same incident — agent slot startup logging `Context Tree sync failed` every restart, and a `malformed inbox:deliver frame — dropping` warning that silently lost a user message until they re-sent it.

### 1. Context Tree clone — HTTPS → SSH fallback

`bootstrap.ts:syncContextTreeRepo` clones over HTTPS. Headless service envs (systemd / launchd) have no TTY for git's credential prompt, so first-time clone exits with `fatal: could not read Username for 'https://github.com'`. Result: every agent slot starts without organizational context, and since the clone dir is never created the failure repeats on every restart.

Now retries once with the scp-like SSH counterpart (`git@host:path`) when the original URL is HTTPS and no local `.git` exists. SSH keys are commonly configured even when credential helpers aren't. Pull failures (existing clone) are unchanged — they continue to fall through to "use existing clone" / diverged-history re-clone.

### 2. Inbox frame — tolerate `source` enum drift + ack unparseable frames

`inboxDeliverFrameSchema` declares `.passthrough()` to survive forward-rolling server fields, but its inner `clientMessageSchema` keeps `messageSourceSchema` as a strict enum. When the server pushed a `source` value the client didn't recognise (schema drift between deploys), the whole frame was rejected. The server's 300s reaper rolled the entry back to `pending` and re-pushed, this client dropped it again, until `retryCount` exhausted — silently losing the message.

- **shared:** override `source` in `clientMessageSchema` with `.catch(null)` so unknown values degrade gracefully. Inbound direction (`messageSchema`, `sendMessageSchema`) stays strict.
- **client:** when a frame still fails parsing for any other reason, best-effort ack via the top-level `entryId` (which usually survives nested-message validation errors). The frame is dropped either way, but acking prevents a 300s reaper retry loop that this build is guaranteed to keep dropping. Logged as `entryIdAcked` for operator correlation.

## Test plan

- [x] `pnpm typecheck` — all 11 tasks pass
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub-shared test` — 251/251 pass (incl. existing `inbox-frames-schema.test.ts`)
- [x] `pnpm --filter @first-tree-hub/client test` — 391/391 pass; one pre-existing flaky test in `git-mirror-manager.test.ts` (5s timeout against an unreachable HTTPS URL, fails on `main` too — unrelated)
- [ ] Manual: restart `first-tree-hub` client on a machine with SSH key configured but no HTTPS credential helper, verify Context Tree clone succeeds via SSH fallback
- [ ] Manual: push a message with a new `source` value not in the client's enum, verify the frame is processed normally and `source` lands as `null`

### DIFF
--- a/packages/client/src/__tests__/client-connection-inbox-malformed.test.ts
+++ b/packages/client/src/__tests__/client-connection-inbox-malformed.test.ts
@@ -1,0 +1,150 @@
+import { createServer, type Server as HttpServer } from "node:http";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { type WebSocket, WebSocketServer } from "ws";
+import { ClientConnection } from "../client-connection.js";
+
+/**
+ * Regression: when an `inbox:deliver` frame fails schema validation, the
+ * client must still ack the surviving top-level `entryId` so the server's
+ * 300s reaper doesn't re-deliver a frame this build is guaranteed to keep
+ * dropping. Without the ack, the entry round-trips through the reaper
+ * `delivered → pending → delivered → drop` loop up to `maxRetries`, then
+ * is silently lost. With the ack, the message is still lost (this build
+ * cannot parse it), but the server state is clean and there's no spam.
+ *
+ * We exercise the malformed path by sending a frame whose nested `message`
+ * is invalid (missing required `id` field) — `entryId` is a top-level
+ * field and stays valid, which mirrors the real-world failure mode where
+ * server-side enum drift breaks `message.source` but the envelope shape
+ * is fine.
+ */
+function makeJwt(payload: Record<string, unknown>): string {
+  const header = Buffer.from(JSON.stringify({ alg: "HS256" })).toString("base64url");
+  const body = Buffer.from(JSON.stringify(payload)).toString("base64url");
+  return `${header}.${body}.sig`;
+}
+
+type ServerSetup = {
+  /** Frame to push to the client right after handshake completes. */
+  malformedFrame: Record<string, unknown>;
+  /** Collects entryIds the client subsequently acks via `inbox:ack`. */
+  ackedEntryIds: number[];
+};
+
+function attachWss(wss: WebSocketServer, setup: ServerSetup): void {
+  wss.on("connection", (ws: WebSocket) => {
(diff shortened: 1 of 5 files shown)


yuezengwu (approved):
Approved — 两个 robustness 修复都合理，scope 小，注释清楚交代了 why（PR #481 刚把 `hub_ui` → `web`，正是这次防御的 enum drift 模式）。无 blocker。

几条非阻塞 nit：

1. `toSshGitUrl` 的正则 `^https:\/\/([^/]+)\/(.+)$` 不处理嵌入凭据 (`https://user:token@host/...`) 或自定义端口 (`https://host:8443/...`)，会改写成错误的 SSH URL。这些场景下 HTTPS 一般本来就能用、不会进 fallback，所以实际影响很小，但碰到时二次失败日志会比较迷惑。

2. `entryIdAcked` 仅判断 `typeof === "number"`，未校验 integer / non-negative；服务端 `inboxAckFrameSchema` 会拒掉异常值，只是一点日志噪声。

3. 测试覆盖小空白：未加针对 `clientMessageSchema.source` 未知值降级到 `null` 的单测；也没覆盖 parse 失败时 ack 是否真的发出。都很容易补，可作为 follow-up。

yuezengwu:
LGTM on the follow-up commit `40648f4` — three nits all addressed cleanly:

1. `toSshGitUrl` 复用 `httpsToSshBaseRewrite`（`new URL()` 解析 + 拒绝嵌入凭据 / 非默认端口），URL 重写规则集中在一处；额外把 SSH fallback 路径的 `binding.repoUrl` 也换成 SSH 形式，和实际 `git remote get-url origin` 一致，下游 `first-tree tree integrate --tree-url` 不会再拿到无法使用的 HTTPS URL。
2. `entryIdAcked` 收紧到 `Number.isInteger && >= 0`，对齐服务端 schema。
3. schema + 集成测试都补全了，best-effort ack 两条分支（有效 entryId → ack 发出 / 非法 entryId → 不 ack）都有真 WS server 覆盖。

Previous approval stands.

PR #480, opened 2026-05-20
fix(chat): preview absolute-in-root .md paths via runtime relative rewrite (Option R / Case A)

Gandy2025:
## Problem

When an agent writes an **absolute `.md` path that lands inside its workspace root** (e.g. `/Users/.../<chatId>/docs/foo.md`), the chat doc-preview link did not render. Web can't map an absolute path back to a snapshot key because **it doesn't know the workspace root** — only the runtime does. (Inline `[label](/abs/foo.md)` was broken for the same reason.)

This is **Case A** of the doc-preview work (follows #474 / #476). Scope is deliberately narrow: only absolute paths inside the workspace root. Context Tree docs (Case B) and any out-of-workspace path remain out of scope.

## Why this approach (Option R), not authoritative link spans

The alternative considered was carrying runtime-authored `links: [{start,end,...}]` byte spans and retiring web's re-scan. That was rejected after reading the code:

- The agent message route (`packages/server/src/api/agent/messages.ts`) sends with **`normalizeMentionsInContent: true`**. When mentions are declared but absent from the body, step 2c in `services/message.ts` **prepends `@name ` to the front of the content**, shifting every byte offset. `purpose: "agent-final-text"` only bypasses the group-mention *enforcement* guards, not this rewrite. So offset-anchored spans would silently break the moment a snapshot rides a send that declares mentions.
- Web's current **re-scan is the source of robustness** — it re-scans the stored (post-rewrite) body and matches tokens by canonical-path-set membership, so it's immune to body mutation. Retiring it would be a regression.

**Option R** keeps that robustness: the runtime — which already reads the file to snapshot it, in a single chokepoint (`result-sink` → `buildMessageDocumentSnapshots`) — rewrites the absolute token to its canonical workspace-relative path **in the outbound text**. Same process, same string, so the rewrite is reliable. Web then sees a relative token and matches the snapshot with **no web / server / schema changes**, and relative display comes for free.

## Changes (runtime only)

- `packages/client/src/runtime/doc-snapshots.ts`
  - `buildMessageDocumentSnapshots` now returns `{ docs, skipped, rewrittenText }`.
  - New `canonicalizeWorkspacePath`: accepts relative paths (as before) **and** absolute paths inside root. Absolute paths are `realpath`'d **before** the containment check, so an ancestor symlink can't fake "inside root"; the relative result is run back through `normalizeDocLinkPath` so hidden segments exposed via symlink are rejected.
[...]

PR #415, opened 2026-05-15
feat(chat): inline markdown preview snapshot in message metadata

Gandy2025:
## Summary

- Solves the cloud Hub server + local agent runtime topology break of PR #356: web preview no longer needs server-side access to the runtime's workspace filesystem.
- Snapshot content lives inline in `messages.metadata.documentContext` (discriminated union: `snapshot` / `path`); existing `{ basePath }` messages are normalised to `kind: "path"` so legacy preview still works on single-host deployments.
- No DB migration, no new API, no new service surface. Just schema + runtime scan + web cache hand-off.

See [proposal #273 in first-tree-context](https://github.com/agent-team-foundation/first-tree-context/pull/273) for the full design (size budget rationale, attack model, web cache strategy, known limitations).

## What changes

| Package | What |
|---|---|
| `shared` | `documentContextSchema` → discriminated union; new `snapshotDocSchema` (sha256 + size + content); new `normalizeDocLinkPath` / `isCanonicalDocLinkPath` shared canonicaliser used by runtime + server + web; size & count caps as exported constants |
| `server` | New `services/doc-snapshots.ts` with `validateDocumentContext` — sha256/size calibration, ≤256KB per file, ≤512KB per message, ≤5 docs per message, canonical path + `.md` suffix enforced; called from `sendMessage` metadata path |
| `client/runtime` | New `runtime/doc-snapshots.ts` — scans inline `[text](path.md)` (skips escaped `\\[...]` and image `alt`); resolves workspace-safely; rejects path traversal, symlink escape, hidden segments / `.agent/` / `.git/`, and symlinks whose realpath crosses into hidden dirs; reads file, computes sha256, builds snapshot; called from `result-sink.ts` after the silent-turn guard |
| `web` | `chat-view` seeds the entire message's `docs[]` into React Query cache on link click (so drawer cross-doc navigation within the same message stays cache-hit); `doc-preview-drawer` reads inline `content` for snapshot variants, falls back to legacy `/docs/preview` for `kind: "path"` |

## Caps (MVP)

- ≤ 256 KB per snapshot
- ≤ 5 snapshots per message
- ≤ 512 KB total snapshot content per message
- `.md` suffix only, no path traversal, no hidden segments, no symlink escape (verified by realpath), no embedded `?` / `#`, no scheme / scheme-relative URLs

## Test plan

- [x] `pnpm check` (Biome): clean
- [x] `pnpm typecheck`: 9/9 clean
[...]

PR #484, opened 2026-05-20
feat(chat): cross-agent workspace doc preview via shared-root snapshots

Gandy2025:
## Background / Goal

doc-preview today only renders a `.md` link when the **sender's own** runtime
snapshotted the file from **its own** workspace root. A doc written into
**another agent's** workspace (e.g. an assistant's `workspaces/<assistant>/<chatId>/…`)
referenced by a different agent was never previewable — it fell outside the
per-sender root fence (the boundary #480 explicitly excluded).

Goal: let **any agent preview a `.md` that lives in any agent's workspace in
the same chat**. This is a product-level generalization, not a one-case patch.

## Approach — register-a-root + realpath fence + relative-root global key

Stop using "which agent's directory the file happens to sit in" as a proxy for
"may this doc be previewed". Instead: a referenced path is snapshot-able when
its **realpath lands inside the shared `workspaces/` common root**, scoped to
`<owner>/<currentChatId>/…`. One fence covers every agent workspace and is
extensible — a future source (e.g. Context Tree) is just "register another
root", same realpath + containment + relative-root-key machinery.

### Confirmed decisions
1. **Topology** — same-machine shared FS (sender runtime reads sibling
   workspaces). Cross-machine left as a future degradation hook
   (owner-runtime snapshot / server relay); the framework accommodates both.
2. **Authz** — chatId structural fence **+** server re-check that the owner is
   a speaker participant of the chat.
3. **Key** — global `<agentSlug>/<chatId>/<rel>` for cross references; self
   references keep their **bare base-relative key** (zero #480 Case A
   regression). Visible cross mentions use the short `<owner>/<rel>` form
   (web re-expands with the current chat id), upgraded to the full global key
   when a short form would collide with a self key.
4. **Common root** — derived by config-ascent (`dirname(workspaceRoot)`); no
   new config surface (honours `FIRST_TREE_HUB_HOME`).

## Security model

- **realpath BEFORE containment** — an ancestor symlink cannot fake "inside
  the common root".
- **chatId-segment fence** — only `<*>/<currentChatId>/…` is in range, so one
  chat can never preview another chat's private workspace docs.
- **hidden-segment rejection** — `.agent/`, `.git/`, dotfiles rejected even
  when reached via a symlink (re-checked after realpath).
- **server owner-in-participants re-validation** (defense in depth) — a
  cross-agent global key for this chat must name a participant; a
  participant-owned key naming a *different* chat is rejected (400). Deep self
  paths whose first segment isn't a participant slug are unaffected.

## Affected packages

| Package | Change |
|---|---|
| `shared` | `buildWorkspaceDocKey` / `parseWorkspaceDocKey` — one key format runtime/web/server agree on |
| `client/runtime` | optional workspace fence in `buildMessageDocumentSnapshots`; cross-workspace resolution; collision-safe rewrite |
| `web` | chat-context-aware `linkifyMarkdownDocPaths`; drawer re-resolves owner from the key (no sender mis-query on cache miss); `useAgentSlugToIdMap` |
| `server` | cross-key provenance authz in `validateDocumentContext` |

**Backwards compatible**: self/legacy bare keys unchanged, historical messages
keep rendering, **#480 Case A preserved** (self absolute path → clean bare
relative rewrite).

Evolution: PR #356 (path preview) → #415 (inline snapshot) → #458 (bare paths)
→ #474 (base-path resolution + snapshot-gate) → #476 (drop `kind:"path"`) →
#480 (self absolute-in-root) → **this PR (cross-agent workspaces)**.

## Independent review (codex-reviewer, 2 rounds)

- **Round 1** — 1×P1 + 2×P2:
  - P1 (security): cross-chat global-key shape could bypass the chatId fence →
    fixed by tightening server provenance (participant-owned foreign-chat keys
    rejected) while keeping deep self paths.
  - P2-a: owner unresolved → fell back to sender → wrong file on cache miss →
    fixed: drawer re-resolves owner from the key, disables fallback if
    unresolved (fail closed).
  - P2-b: cross short form could collide with a self bare key → fixed: runtime
    rewrites to the full global key on collision.
- **Round 2 (re-review)** — **no blocking findings**.

## Tests

- `shared` 249 · `client` 412 · `web` 206 · `server` 1101 — all green.
- Per-package `typecheck` clean; `biome` clean on all changed files.
- New coverage: shared key build/parse; client cross-agent fence + chat-scope +
  symlink/hidden rejection + collision→full-key; web cross-expansion +
  collision disambiguation; server owner-in-participants + foreign-chat reject +
  deep-self-path allow.

> Note: the repo-wide `pnpm check` currently errors on pre-existing nested
> `.claude/worktrees/*/biome.json` configs (unrelated to this diff); targeted
> `biome check` on the changed files is clean.

A design doc was produced and reviewed before implementation (kept local,
per repo convention not to commit design docs).

### DIFF
--- a/packages/client/src/__tests__/doc-snapshots.test.ts
+++ b/packages/client/src/__tests__/doc-snapshots.test.ts
@@ -117,3 +117,115 @@ describe("buildMessageDocumentSnapshots — absolute-in-root rewrite (Option R /
     expect(rewrittenText).toBe("first design.md then again design.md");
   });
 });
+
+/**
+ * Cross-agent doc preview: with a workspace fence, an absolute `.md` path that
+ * realpaths into ANOTHER agent's workspace (same chat) under the shared
+ * `workspaces/` common root is snapshotted with a global
+ * `<ownerSlug>/<chatId>/<rel>` key and rewritten to the short `<ownerSlug>/<rel>`
+ * form. Self paths keep their bare key + #480 behaviour; out-of-scope paths
+ * (other chat, other root, hidden) stay verbatim.
+ */
+describe("buildMessageDocumentSnapshots — cross-agent workspace fence", () => {
+  let workspacesRoot: string;
+  let selfRoot: string;
+  const chatId = "chat-xyz";
+  const selfSlug = "coder";
+  const otherSlug = "assistant";
+
+  beforeAll(async () => {
+    workspacesRoot = await mkdtemp(join(tmpdir(), "doc-snap-ws-"));
+    // self workspace: workspaces/coder/chat-xyz
+    selfRoot = join(workspacesRoot, selfSlug, chatId);
+    await mkdir(selfRoot, { recursive: true });
+    await writeFile(join(selfRoot, "mine.md"), "# mine\n", "utf8");
+    // Self file whose relative path collides with a cross short form
+    // (`assistant/design.md`) — used to prove the collision → full-key rewrite.
+    await mkdir(join(selfRoot, otherSlug), { recursive: true });
+    await writeFile(join(selfRoot, otherSlug, "design.md"), "# my own assistant notes\n", "utf8");
+    // sibling agent workspace (same chat): workspaces/assistant/chat-xyz
+    const otherRoot = join(workspacesRoot, otherSlug, chatId);
+    await mkdir(join(otherRoot, "docs"), { recursive: true });
+    await writeFile(join(otherRoot, "design.md"), "# their design\n", "utf8");
+    await writeFile(join(otherRoot, "docs", "intro.md"), "# their intro\n", "utf8");
+    await mkdir(join(otherRoot, ".agent"), { recursive: true });
+    await writeFile(join(otherRoot, ".agent", "secret.md"), "# their secret\n", "utf8");
+    await symlink(join(otherRoot, ".agent", "secret.md"), join(otherRoot, "leak.md"));
+    // sibling agent in a DIFFERENT chat: workspaces/assistant/other-chat
+    const otherChatRoot = join(workspacesRoot, otherSlug, "other-chat");
+    await mkdir(otherChatRoot, { recursive: true });
+    await writeFile(join(otherChatRoot, "private.md"), "# other chat\n", "utf8");
+  });
+
+  afterAll(async () => {
+    await rm(workspacesRoot, { recursive: true, force: true });
+  });
+
+  const fence = () => ({ workspacesRoot, chatId, selfSlug });
+
+  it("snapshots a sibling agent's doc with a global key + short rewrite", async () => {
+    const abs = join(workspacesRoot, otherSlug, chatId, "design.md");
+    const { docs, rewrittenText } = await buildMessageDocumentSnapshots(`see ${abs} please`, selfRoot, fence());
+
+    expect(docs.map((d) => d.path)).toEqual([`${otherSlug}/${chatId}/design.md`]);
+    expect(docs[0]?.content).toBe("# their design\n");
+    expect(rewrittenText).toBe(`see ${otherSlug}/design.md please`);
+  });
+
+  it("preserves subdir + :line suffix in the cross rewrite", async () => {
+    const abs = join(workspacesRoot, otherSlug, chatId, "docs", "intro.md");
+    const { docs, rewrittenText } = await buildMessageDocumentSnapshots(`open ${abs}:10 now`, selfRoot, fence());
+
+    expect(docs.map((d) => d.path)).toEqual([`${otherSlug}/${chatId}/docs/intro.md`]);
+    expect(rewrittenText).toBe(`open ${otherSlug}/docs/intro.md:10 now`);
+  });
+
+  it("keeps the self path bare (Case A unchanged) even with a fence", async () => {
+    const abs = join(selfRoot, "mine.md");
+    const { docs, rewrittenText } = await buildMessageDocumentSnapshots(`my ${abs} file`, selfRoot, fence());
+
+    expect(docs.map((d) => d.path)).toEqual(["mine.md"]);
+    expect(rewrittenText).toBe("my mine.md file");
+  });
+
+  it("rejects a sibling doc from a DIFFERENT chat (chat-scope fence)", async () => {
+    const abs = join(workspacesRoot, otherSlug, "other-chat", "private.md");
+    const text = `cross-chat ${abs} nope`;
+    const { docs, rewrittenText } = await buildMessageDocumentSnapshots(text, selfRoot, fence());
+
+    expect(docs).toEqual([]);
+    expect(rewrittenText).toBe(text);
+  });
+
+  it("rejects a sibling symlink whose realpath crosses into a hidden dir", async () => {
+    const abs = join(workspacesRoot, otherSlug, chatId, "leak.md");
+    const text = `leak ${abs}`;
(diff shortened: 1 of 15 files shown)


yuezengwu (approved):
LGTM ✅

### 实现目标
将 doc-preview 从"仅发件人自己工作区"通用化到"同一 chat 内任意 agent 工作区"。设计上抽象为「register-a-root + realpath fence + relative-root global key」，未来新源（如 Context Tree）即是再注册一个根。

### 核心改动 review
- **shared**: 新增 `buildWorkspaceDocKey` / `parseWorkspaceDocKey`，一份键格式三端共识（runtime/web/server）。
- **client/runtime**: `WorkspaceFence` opt-in；`resolveCrossWorkspaceDoc` 严格按 `realpath → containment → chatId 段 → hidden 段拒绝 → stat` 顺序；冲突 fallback 到全键（codex P2-b 修复）。
- **server**: `validateDocumentContext(chatScope?)` 双端复检：(a) 当前 chat 的 cross 键 owner 必须是 participant；(b) participant-owned 但 chatId 段不匹配当前 chat 直接 reject（codex P1 修复）。
- **web**: `linkifyMarkdownDocPaths` 接受 `chatId` 做 short→global 扩展；drawer 从 key 重新解析 owner 失败则 fail-closed（不回落到 sender 工作区，codex P2-a 修复）；click handler 把 owner 写入 `docAgent`。

### 安全模型评估
分层清晰：
1. realpath 先于 containment（祖先 symlink 不能伪造"在共享根内"）
2. chatId 段围栏（一个 chat 不能预览另一个 chat 的私有 workspace 文件）
3. 隐藏段（`.agent/`、`.git/`、dotfiles）在 realpath 之后再次拒绝（堵 symlink→hidden 逃逸）
4. server owner-in-participants 复检（防御纵深，即便 runtime 被攻陷也不能广播非成员的 workspace 文件）
5. drawer owner 不可解析 → 关闭 fallback（不会同名误命中 sender 工作区）

### 数据格式向后兼容
- 无 DB schema 变更，无 zod schema 变更；`metadata.documentContext.docs[].path` 仍是 string
- 值的格式域扩展：legacy/self 维持 bare key（#480 Case A 完整保留），新消息可能携带 `<slug>/<chatId>/<rel>` 全键
- 通过 `parseWorkspaceDocKey` 对 <3 段输入返回 null 让 legacy bare key 自动跳过 chatScope 检查
- 该字段写入 immutable 消息历史，但写入的依然是合法的 canonical 字符串，无需迁移

### 测试
shared 249 / client 412 / web 206 / server 1101 全绿；覆盖到位，特别是：
- shared: build/parse 各种 reject case 都有
- client: cross fence + 跨 chat reject + symlink→hidden reject + 冲突→全键 + 无 fence 时 opt-out
- server: participant 校验四象限 + foreign-chat reject + 深 self 路径不被误判
- web: 短形扩展 + 直接键优先 + 冲突消歧

### 非 blocker 小注意点
1. **slug 大小写**：build 端保留原始大小写，server `participantSlugs.has(...)` 与 web `useAgentSlugToIdMap` 双端都 lowercase。链路一致，但前提是 agent 命名按惯例小写。case-insensitive FS（macOS 默认）下 `ownerSlug === fence.selfSlug` 受 realpath 是否规范化大小写影响，生产侧通常 case-sensitive，可接受。
2. **`parseWorkspaceDocKey` 二义性**：`docs/sub/a.md` 这类深 self 路径也能 parse 成 cross 形。已通过"调用方必须要求 `chatId == 当前 chat`"消歧，注释和 server 用例都写清了。chat id 是 UUID 让冲突在实践中不可能发生。
3. **workspaceRoot 拓扑假设**：`dirname/basename` 推导依赖 `workspaceRoot = <workspaces>/<agentSlug>`。即使被自定义路径破坏，fence 也只在严格 `<*>/<currentChatId>/…` 结构下匹配，加上 server 端 participant 复检，不会产生假阳性。

设计稳、安全分层清晰、codex 两轮 review 已全部消化、覆盖测试完整、向后兼容。Approve。

baixiaohang:
Review pass. Overall: solid extension of the doc-preview fence with a coherent four-layer security model and full back-compat. Implementation matches the design doc summary in the PR body — I'm not blocking, leaving non-blocking notes below.

### Strengths

- **Realpath-before-containment + chatId-segment fence + hidden-segment reject + server owner-in-participants** — four independent layers; bypassing one still gets caught by the next. The "deep self path whose first segment isn't a participant" allow-case in `validateDocumentContext` is the right exception to the chatId-fence rule and is explicitly tested.
- **P2-b collision handling** — the runtime/web split (runtime emits the full global key when a cross short form would collide with a self bare key; web's `resolveSnapshotKey` tries self-direct first, then chatId-expansion) means "rewritten ⇔ previewable" still holds even on a collision. Test `assistant/design.md` (self) vs `assistant/<chatId>/design.md` (cross) pins this.
- **P2-a fail-closed** — drawer disables the path fallback when the cross-owner slug can't be resolved (`enabled: ... && Boolean(fallbackAgentId)`), so a cache miss can no longer surface a same-named file from the sender's workspace. Good call.
- **`buildWorkspaceDocKey` self-validates** — `isCanonicalDocLinkPath(key)` at the end means a slug/chatId carrying a stray `.` or empty segment can never produce a key that web would later canonicalise into a different string. Avoids the silent cache-miss class entirely.
- **Slug case alignment** — server stores `participantSlugs` lowercased + queries `key.agentSlug.toLowerCase()`; web `slugToId` lowercases on both sides. Three-way consistent.
- **#480 Case A regression coverage** — explicit test "keeps the self path bare (Case A unchanged) even with a fence" pins the no-regression guarantee.

### Non-blocking notes

❓ **`useAgentSlugToIdMap` uses `listAgents({ limit: 100 })`** — for orgs with >100 agents, an off-roster sibling slug won't resolve → drawer fail-closes the path fallback. Inline-snapshot rendering is unaffected (renders from cache, no UUID needed), which the comment explicitly says is the contract — so this is graceful degradation, not a bug. But worth flagging in the PR description so future-us doesn't trip on "preview works inline but not on reload" with a large roster. A future `getAgentBySlug` endpoint would be the cleaner fix.

⚠️ **`parseWorkspaceDocKey` ambiguity note is a soft assumption** — "chat ids are uuids, so a self subdir literally named the chat id is effectively impossible" is true today, but the shared package can't enforce UUID shape. If chat IDs ever change format (short codes, slugs), every `key.chatId === currentChatId` callsite (server authz + web linkify + web drawer + runtime fence) has to be re-audited. Worth a one-liner in the `parseWorkspaceDocKey` JSDoc calling this out as a cross-cutting invariant rather than a local quirk.

❓ **`buildWorkspaceDocKey` slug/chatId character validation** — rejects `/` and leading `.`, and `isCanonicalDocLinkPath` catches the rest, but there's no explicit rejection of whitespace / control chars / `:` in slug or chatId. In practice agent names should be DB-constrained to a safe slug shape, so the risk surface is just an asymmetric build/parse — not a security issue. Nice-to-have: link the DB-side constraint in the JSDoc so future readers know what the upstream invariant is.

⚠️ **`mergeStateStatus: BEHIND`** — branch is behind `main`; the in-progress `Lint & Type Check` / `Test` jobs are running against the merge prediction. Recommend rebasing onto `main` once those pass, so the green check reflects the actual merge result.

### Tests

Coverage is comprehensive — every cross-cutting branch I'd want to see (cross-chat reject, symlink-to-hidden reject, collision → full-key, no-fence opt-in, deep self path allow, foreign-chat participant reject, snapshot key absent stays plain) has a unit. No gaps from my read.

PR #488, opened 2026-05-21
refactor(chat): unify header + sidebar add-participant into one dropdown

Gandy2025:
## What

Unifies the two "add an agent to this chat" entry points into a single shared component.

- **New** `<AddParticipantDropdown variant="icon" | "inline">` (`packages/web/src/components/add-participant-dropdown.tsx`) backs both:
  - the chat **header** quick-add icon (`variant="icon"`), and
  - the right-**sidebar** inline "Add" row (`variant="inline"`, three-state Add / Adding… / All added).
- The `variant` prop swaps **only the trigger shell**. The candidate dropdown is now **identical across both surfaces**: avatar + "mine / others" grouping with a divider. (The header dropdown previously had no avatars — it now matches the sidebar.)
- The component owns disclosure (open / click-outside / Esc, capture-phase + `stopPropagation` so Esc closes only the dropdown and never collapses the rail), **keyboard navigation** (arrows / Enter, skipping the divider), and the immediate `addMeChatParticipants` mutation.

## Also

- **Drops the duplicate `chat-detail` + `activity` queries** the participants section was re-declaring. Membership data (`participants` / `participantsLoading` / `managedByMe`) is now passed down from `ChatView`, which already holds those queries.
- Removes the now-obsolete `AddParticipantQuickButton` (chat-view) and `AddParticipantInlineButton` (participants-section).

## Out of scope

- The **new-chat `ParticipantChips`** is intentionally **untouched** — its deferred-commit chip model (`createMeChat` on send) and `@`-mention ↔ chip linkage are conversation-creation semantics that don't belong in the immediate-add dropdown.

## Behavior preserved

Sidebar three-state + "All added" disabled; add → member list refreshes (`onAdded` → invalidate `chat-detail`); header grouping + divider; Esc scoping (dropdown only, not the rail).

## Verification

- `pnpm --filter @first-tree-hub/web typecheck` ✅ (incl. design-token guardrails)
- `biome check` on the changed files ✅
- `pnpm --filter @first-tree-hub/web test` ✅ 206/206
- **User-verified locally** (header / sidebar dropdowns identical: avatar + grouping + divider; keyboard nav; immediate refresh; three-state; Esc scoping).

Net: +297 / −365 across 4 files.

### DIFF
--- a/packages/web/src/components/add-participant-dropdown.tsx
+++ b/packages/web/src/components/add-participant-dropdown.tsx
@@ -1,0 +1,260 @@
+import { useMutation } from "@tanstack/react-query";
+import { Loader2, Plus, UserPlus } from "lucide-react";
+import { useEffect, useMemo, useRef, useState } from "react";
+import { addMeChatParticipants } from "../api/me-chats.js";
+import { Avatar as RealAvatar } from "./avatar.js";
+import {
+  ambiguousDisplayNames,
+  groupAndSortCandidates,
+  type MentionCandidate,
+  MentionLabel,
+} from "./mention-autocomplete.js";
+
+type AgentIdentityResolver = (
+  uuid: string | null | undefined,
+) => { name: string | null; displayName: string; avatarImageUrl: string | null } | null;
+
+/**
+ * Shared "add an agent to this chat" dropdown. Backs both the header
+ * quick-add icon (`variant="icon"`) and the right-sidebar inline row
+ * (`variant="inline"`). Only the trigger shell differs by variant; the
+ * dropdown list is identical across both surfaces: avatar + "mine / others"
+ * grouping with a divider.
+ *
+ * The component owns disclosure (open / click-outside / Esc), keyboard
+ * navigation, and the immediate `addMeChatParticipants` mutation. Callers
+ * pass the candidate union (members + org-discoverable, already shaped to
+ * `MentionCandidate`, self excluded) and the current `participantIds`; the
+ * already-in members are filtered out here.
+ */
+export function AddParticipantDropdown({
+  chatId,
+  candidates,
+  participantIds,
+  agentIdentity,
+  onAdded,
+  variant,
+}: {
+  chatId: string;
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Approve.

Clean refactor: unifies two near-duplicate "add participant" components into one `<AddParticipantDropdown variant="icon" | "inline">`, lifts `participants` / `participantsLoading` / `managedByMe` up to `ChatView` so `ParticipantsSection` stops re-subscribing to `chat-detail` + `activity`, and preserves all the load-bearing behavior (header avatars + grouping, sidebar three-state, capture-phase Esc scoping that doesn't collapse the rail, immediate `addMeChatParticipants` mutation).

Verified:
- `managedByMeMap` already exists in ChatView (line 1182), so the new prop wiring is reusing existing state, not duplicating it.
- The capture-phase + `stopPropagation` Esc handler that originally fixed the "Esc collapses the rail" bug is preserved verbatim in the shared component.
- `commit()` guards on `addMut.isPending` and `onSuccess` invalidates `chat-detail` — no race / no stale candidate list.
- `AgentIdentityResolver` drops the unused `avatarColorToken` field; structurally compatible with the callers' fuller resolvers.
- No backend / data structure / SQL changes.

Net −68 lines and one fewer React Query subscription per chat. LGTM.

Nit (non-blocking): the `useEffect(... [open])` on line ~107 comments "Reset the highlight whenever the menu opens or the option set changes", but `selectable.length` isn't in the deps. Doesn't actually break — `% selectable.length` in arrow nav and `?? selectable[0]` in commit keep things safe — but the comment slightly overpromises.

baixiaohang (approved):
LGTM. Clean refactor — dedupes the two add-participant surfaces into one `<AddParticipantDropdown>`, and removes a redundant `chat-detail` + `activity` subscription in the sidebar by lifting state up to `ChatView`. Behavior preserved end-to-end (capture-phase Esc scoping, three-state, immediate refresh, header avatars + grouping).

PR #491, opened 2026-05-21
fix(web): allow npm scoped package names in chat content

baixiaohang:
## Summary
- Sending a chat message containing an npm scoped package name (e.g. `npm i @agent-team-foundation/first-tree-hub-shared`) was blocked client-side by the unresolved-`@`-token pre-flight, because the shared `MENTION_REGEX` matched `agent-team-foundation` (and via backtracking, the partial prefix `agent-team-`) as a mention token.
- Tightened `MENTION_REGEX` with a strict trailing lookahead so `@scope/pkg` is no longer treated as a mention at all, and removed the client-side pre-flight in `chat-view.tsx` — unresolved `@<token>` is now treated as plain text on the human web path (Slack/Discord/Feishu parity).
- Server-side guard untouched: the agent path's PR-393 anti-hallucination contract (`enforceGroupMention` + unresolved-token rejection) stays exactly as it was.

## Root cause

`/(?<![A-Za-z0-9_.@-])@([A-Za-z0-9][A-Za-z0-9_-]{0,63})\b/g` — the trailing `\b` accepts `/` as a word boundary, so `@agent-team-foundation/...` matches `agent-team-foundation`. Worse, greedy backtracking can also surface partial prefixes like `agent-team-` when a longer match runs into a non-boundary char downstream. The client's `scanMentionTokens` pre-flight then surfaces this as a hard block before the request ever leaves the browser.

## Fix

1. **`packages/shared/src/mentions.ts`** — replace trailing `\b` with `(?![A-Za-z0-9_/-])`. Strict negative lookahead forbids the token being followed by `/`, letters, digits, `_`, or `-`. This collapses both the false positive (`@scope/pkg`) and the backtrack chain (no partial prefix can survive because every shorter candidate is followed by a forbidden char). Single source of truth — the same regex feeds the server's `scanMentionTokens` and `extractMentions`, so server-side fan-out routing stays consistent with client-side resolution.

2. **`packages/web/src/pages/workspace/center/chat-view.tsx`** — drop the unresolved-`@`-token pre-flight in `handleSend`. The previous behavior mirrored the server's `enforceGroupMention` guard for round-trip avoidance; with the npm-scope false positive now structurally impossible, the only remaining unresolved tokens are real typos (`@<outsider>`), which the server already 400s and the existing `sendMut.onError` backstop renders under the composer. `extractMentions` / `draftMentions` is retained as a pure UX signal — it drives the send-button disabled state when a group chat has no valid `@mention` yet, but no longer hard-blocks send.

The architectural cleanup of moving toward "server only trusts explicit `mentions[]`, each client parses its own content" (so the entire `client/server regex-divergence` bug class disappears) is deferred to a separate PR.

## Test plan
- [x] `pnpm check` — biome clean
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub-shared typecheck` clean
- [x] `pnpm --filter @first-tree-hub/web typecheck` clean (incl. design-token guardrails)
- [x] `pnpm --filter @agent-team-foundation/first-tree-hub-shared test` — 259 tests pass, includes 2 new npm-scope cases (one in `extractMentions`, one in `scanMentionTokens` pinning the partial-prefix backtrack invariant)
- [x] `pnpm --filter @first-tree-hub/web test` — 206 tests pass
- [x] `pnpm --filter @first-tree-hub/server test` — 1106 tests pass, including the full `agent-send-unresolved-mention` suite (PR-393 anti-hallucination guard, 11 cases) — unchanged behaviour on agent path

## Manual verification
- [ ] In web chat, type `npm i @agent-team-foundation/first-tree-hub-shared` and send — message goes through
- [ ] In web chat, type `@<some-outsider>` (a real agent name not in this chat) and send — server returns 400 and error surfaces under the composer (no client-side hint about `[+]` anymore, server message instead)
- [ ] Mention autocomplete still surfaces chat-internal members on `@`; `@<member>` send still works

## Behaviour matrix
| Scenario | Before | After |
|---|---|---|
| `@scope/pkg` in message | client blocks | sends normally |
| Real `@<outsider>` typo | client blocks | client allows → server 400 → error in composer |
| Group chat with no `@mention` | send button disabled | unchanged |
| Normal `@<member>` | works | works |
| Agent runtime path `@<outsider>` | server 400 (PR-393) | server 400 (PR-393) — unchanged |

### DIFF
--- a/packages/shared/src/__tests__/mentions.test.ts
+++ b/packages/shared/src/__tests__/mentions.test.ts
@@ -88,6 +88,16 @@ describe("extractMentions", () => {
     expect(extractMentions("@@alice was here", participants)).toEqual([]);
   });
 
+  it("does not match npm scoped package names (`@scope/pkg`)", () => {
+    // The composer's pre-flight check uses the same regex; without the
+    // trailing `/`-rejecting lookahead, `npm i @agent-team-foundation/first-tree-hub`
+    // would surface `agent-team-foundation` as an unresolved mention token
+    // and block the send. Defend the npm-scope shape for the participant
+    // names too — `@alice/foo` is not a mention of alice.
+    expect(extractMentions("npm i @agent-team-foundation/first-tree-hub", participants)).toEqual([]);
+    expect(extractMentions("see @alice/foo for details", participants)).toEqual([]);
+  });
+
   it("handles hyphen-containing participant names", () => {
     expect(extractMentions("@charlie-07 check stats", participants)).toEqual(["agent-charlie"]);
   });
@@ -111,5 +121,13 @@ describe("scanMentionTokens", () => {
 
   it("still scans tokens that don't match any participant — useful for warn logs", () => {
     expect(scanMentionTokens("@ghost")).toEqual(["ghost"]);
+  });
+
+  it("ignores npm scoped package names entirely (no partial-prefix backtrack)", () => {
+    // Without the strict trailing lookahead the engine backtracked the greedy
+    // `[A-Za-z0-9_-]{0,63}` and surfaced `agent-team-` as a partial-match
+    // mention token of `@agent-team-foundation/...`. Pin both forms.
+    expect(scanMentionTokens("npm i @agent-team-foundation/first-tree-hub-shared")).toEqual([]);
+    expect(scanMentionTokens("@alice/foo and @bob/bar")).toEqual([]);
   });
 });
--- a/packages/shared/src/mentions.ts
+++ b/packages/shared/src/mentions.ts
@@ -28,7 +28,13 @@ export type MentionParticipant = {
 // First char must be alphanumeric, matching `AGENT_NAME_REGEX` in
 // `schemas/agent.js`. An `@-foo` or `@_foo` token is not a valid mention
 // (`-foo` can never be a legal new agent name under the tightened rule).
-export const MENTION_REGEX = /(?<![A-Za-z0-9_.@-])@([A-Za-z0-9][A-Za-z0-9_-]{0,63})\b/g;
+//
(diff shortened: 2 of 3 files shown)


yuezengwu (approved):
LGTM — 这是一个范围聚焦、契约清晰的 bug fix。

## 核心改动
- `MENTION_REGEX` 尾部 `\b` → `(?![A-Za-z0-9_/-])`：结构性地消除 `@scope/pkg` 误匹配，**并且**通过严格 lookahead 同时关闭了 greedy backtracking 留下的所有更短前缀（如 `agent-team-`）的逃逸路径。
- 移除 `chat-view.tsx` 中基于 `scanMentionTokens` 的客户端 pre-flight 拒绝；未解析的 `@<token>` 视作纯文本（Slack/Discord/Feishu 对齐），`extractMentions` 仅保留为 send 按钮 disabled 的 UX 信号。
- 服务端 `enforceGroupMention` 与 agent 路径的 PR-393 反幻觉契约不动 — agent 路径行为保持。

## Regex 正确性（手动 trace）
对 `@agent-team-foundation/first-tree-hub-shared`，所有可能的 backtrack 前缀下一字符要么是字母（被起始字符类约束拒绝）、要么是 `-`/`_`/`/`（被 lookahead 拒绝），整体无匹配。新增的 `scanMentionTokens` 用例正好 pin 住这条 partial-prefix 不变式 — 测试质量很高。

## 取舍记录
1:1 chat 下打 `@<outsider>` 不再有精准 client hint，需要 server 400 往返。这是有意取舍，由 `sendMut.onError` 兜底，可接受。

## 非阻塞建议
后续若推进 "server 只信 `mentions[]` 数组、不再 server-side regex 解析" 的方向（description 中已点名），可以彻底消除 client/server regex divergence 这一类 bug，值得作为独立 PR 跟进。

PR #492, opened 2026-05-21
fix(web): chat picker shows humans + stable new-chat default

### LINKED ISSUE #343: Web: cannot add humans as chat participants — picker source (`activity.agents`) filters them out via `runtimeState IS NOT NULL`

liuchao-001:
## Summary

- **Closes #343**: web participant pickers (`[+]` in chat header, `[+]` in right-sidebar participants section, chip `[+]` in new-chat draft) couldn't surface human members of the org. All three are now wired to `GET /orgs/:orgId/agents` instead of `/activity`. Server unchanged.
- **Partial fix for #342**: `pickDefault` drops the `runtimeUpdatedAt` MRU branch — the new-chat default chip is now stable across clicks.

This is a re-do of the closed #350 against current main. Main already had the `@`-autocomplete-vs-`[+]`-picker semantic split and the `AddParticipantDropdown` unification done independently — so the actual change here is much smaller: just the data-source switch in two files, a doc-comment fix, the `pickDefault` simplification, and two regression tests.

## Why

`GET /orgs/:orgId/activity` returns agents through `listAgentsWithRuntime`, whose SQL filters on `runtimeState IS NOT NULL`. Human agents (`type='human'`) are mirrors of human users — they never bind a runtime, so they never have a `runtimeState`, and the filter drops them. The picker on top of `/activity` therefore couldn't see coworkers.

`GET /orgs/:orgId/agents` (`listAgentsForMember`) is already member-accessible (not admin-gated), already LEFT-JOINs `agent_presence`, already returns `managerId` so we can compute `managedByMe` client-side. It's the right shape for "show me all agents I can invite to a chat".

## Follow-up issues filed for non-blocking nits

- #494 — `limit: 100` silent truncation for orgs with >100 visible agents
- #495 — duplicate `listAgents` fetch under two React Query keys (`["org-agents"]` vs `["agents", "name-map"]`)

Inline nit #3 (asymmetric identity fallback between header `[+]` and new-chat draft) was a small correctness gap and got fixed inside this PR — commit `6cb41378` aligns `chat-view.tsx`'s `addableCandidates` with `new-chat-draft.tsx`'s fallback pattern.

## Test plan

- [x] `pnpm check` — clean
- [x] `pnpm typecheck` — 11/11 packages
- [x] `pnpm --filter @first-tree-hub/web test` — 23 files / 216 tests pass (includes 10 new `pickDefault` cases)
- [x] `pnpm --filter @first-tree-hub/server test me-chat-service` — 25/25 tests pass (includes the new regression test that `addMeChatParticipants` accepts humans)
- Manual UI verification (run on a fresh local env by @code-reviewer via agentic browser; screenshots at `~/Workspace/agent-browser/screenshots/pr-492/`):
  - [x] **API smoke**: `/agents` returns humans (2 human + 1 PA + 2 autonomous); `/activity` returns 0 — confirms data-source swap is necessary and correct
  - [x] Existing chat → `[+]` in header → human teammate appears as candidate (`01-header-picker-open.png`: My PA / **Alice Human** / Rogue Bot; correctly excludes already-in-chat Claude Bot + Dev User)
  - [x] Existing chat → `[+]` in right-sidebar participants → human teammate appears (`02-sidebar-picker-open.png`: same candidates — both surfaces share one data source)
  - [x] Existing chat → `@` in composer → only current chat participants (`03-at-autocomplete.png`: only Claude Bot — the lone other participant; Alice / My PA / Rogue Bot all absent)
  - [x] New chat draft → `[+]` chip picker → humans + AI both appear (`04b-new-chat-picker.png`: Claude Bot / **Alice Human** / Rogue Bot)
  - [x] New chat draft → open/close repeatedly → default chip is the same agent every time (`05-stability-trial-*.png`: 8/8 trials all seed **My PA**, matching `pickDefault`'s my-managed PA > my-managed AI > null priority)
- [x] Browser console: no errors; page errors: empty


yuezengwu (approved):
## Review summary

**Approving.** The change is focused, well-justified, and the tests lock down the exact regressions we'd be most likely to silently reintroduce.

### What the PR does
- Swaps the three `[+]` participant pickers (chat header, right-sidebar, new-chat draft chip) from `GET /orgs/:orgId/activity` to `GET /orgs/:orgId/agents`. The old endpoint's `runtimeState IS NOT NULL` filter silently dropped human members.
- Derives `managedByMe` client-side from `managerId === myMemberId`, matching the doc-comment update on `MentionCandidate.managedByMe`.
- Drops the `runtimeUpdatedAt` MRU branch from `pickDefault` — the new precedence is `my-managed PA → first my-managed AI agent → null`, with humans + suspended rows excluded.
- Adds a server-side regression test (`addMeChatParticipants` lands a human as `access_mode='speaker'`) and a pure-function test suite for `pickDefault` (10 cases incl. an explicit "stable across calls" lock).

### Why this looks correct
- `listAgentsForMember` uses `agentVisibilityCondition`, which doesn't filter on `type` and LEFT-JOINs `agent_presence` — so humans surface naturally, no server change required.
- The query key `["org-agents"]` is shared by both consumers so react-query dedupes the fetch.
- `Pick<Agent, "uuid" | "type" | "managerId" | "status">` covers everything `pickDefault` reads; the `null`-checked `managerId === myMemberId` path is safe.
- Humans self-manage (`managerId = own memberId`), so excluding `type==='human'` from `pickDefault` is necessary — otherwise the user's own human mirror would beat any PA. The unit test covers this.
- Suspended-row filter on the picker side avoids the "click row → server 4xx" footgun.

### Notes (non-blocking)
- `limit: 100` is the server's hard cap (`paginationQuerySchema`). For orgs above that the picker silently truncates. The inline comment calls this out as a follow-up, which is fine — issue 343 is a correctness fix, not a scale one.
- The new useQuery has `refetchInterval: 30_000` (vs activity's 15s). Reasonable for a picker source. No action.
- `pickDefault`'s "first my-managed agent" fallback is order-dependent on the API result. The server orders by `desc(agents.createdAt)`, which is deterministic, so the issue-342 "stable across clicks" guarantee holds. The explicit test for stability is a nice belt-and-braces.

### Manual UI verification
Still owed by the PR description; flagging here so it doesn't slip:
- existing chat → header `[+]` shows human teammate
- existing chat → right-sidebar `[+]` shows human teammate
- existing chat → `@` autocomplete is unchanged (still membership-scoped)
- new-chat `[+]` chip shows humans + AI
- new-chat default chip is stable across re-opens

LGTM. Approving.

baixiaohang (approved):
LGTM, approving on top of @yuezengwu's review — substantively the same conclusion: the root cause (`runtimeState IS NOT NULL` in `listAgentsWithRuntime` silently filtering humans) is correct, the data-source swap to `listAgentsForMember` is the right minimal fix, and the two regression-lock tests (pure-function `pickDefault` stability + server-side `addMeChatParticipants` accepts humans) cover exactly the invariants most likely to be silently undone later.

A couple of additional non-blocking nits worth noting for follow-up — none should hold this PR:

1. **`limit: 100` silent truncation.** The picker source is capped by `paginationQuerySchema`. Orgs above 100 visible agents will silently miss candidates with no UI cue. The inline comment already flags this — fine as a follow-up, but worth a tracking issue.

2. **Duplicate `listAgents({ limit: 100 })` fetch under two query keys.** The new picker query uses `["org-agents"]`, while `useAgentIdentityMap` / `useAgentNameMap` / `useAgentSlugToIdMap` use `["agents", "name-map"]`. Same URL, two cached payloads — React Query doesn't dedupe across keys. Minor inefficiency; could be unified to a single shared key in a follow-up.

3. **Asymmetric identity fallback between the two surfaces.** `new-chat-draft.tsx` falls back to the agent row (`ident?.name ?? a.name`) when the identity map hasn't resolved yet, but `chat-view.tsx`'s `addableCandidates` still hard-filters on `if (!ident || !ident.name) continue;`. In practice both maps are fed by `listAgents`, so the window is short, but a brand-new human teammate could briefly fail to surface in the header `[+]` picker on first paint. Worth aligning later.

For the manual UI verification still owed (5 boxes in the PR description), please tick them off here before merge — particularly the "default chip stable across re-opens" one, since that's the only place the issue-342 fix is observable end-to-end.

yuezengwu:
Follow-up commit `6cb4137` looks good — clean targeted fix for nit #3, mirrors the `new-chat-draft.tsx` fallback exactly. `MentionCandidate.displayName` is already `string | null` so the fallback chain is type-safe. Approval still stands.

liuchao-001:
Thanks for the approvals, @yuezengwu and @baixiaohang. Quick status on the three non-blocking nits:

- **Nit #1 — `limit: 100` silent truncation** — filed as #494 for follow-up. The picker query has an inline comment flagging this; #494 captures the implementation directions (server-side `?query=` search, or client-side pagination loop, or both).
- **Nit #2 — duplicate `listAgents` fetch under two query keys (`["org-agents"]` vs `["agents", "name-map"]`)** — filed as #495 for follow-up. Both options (unify on one key, or extract a shared `useOrgAgents` hook) captured.
- **Nit #3 — asymmetric identity fallback between header `[+]` and new-chat draft** — fixed in this PR (commit 6cb41378). `addableCandidates` now mirrors `new-chat-draft.tsx`'s pattern: prefer the identity map, fall back to the raw `listAgents` row's `name`/`displayName` when the map hasn't indexed a brand-new teammate yet. Re-ran `pnpm check` / `pnpm typecheck` / `pnpm --filter @first-tree-hub/web test` — all green.

Manual UI verification is in progress now (kicked off by code-reviewer on a fresh local env); will check the 5 boxes in the PR description once that lands.

Issue #317, opened 2026-05-12
GitHub App webhook: processed_events leak when claim succeeds but post-claim work fails


Gandy2025:
**Severity: P1**
**Source: codex challenge against PR #300 rollup, 2026-05-12**
**Status: pre-existing in PR-300 (the claim helper landed in C.11 / commit `094c74f`). Same risk shape exists in main's per-org webhook — this is a design-level concern in the \"claim before work\" pattern.**

## Finding

In [`webhooks/github-app.ts`](packages/server/src/api/webhooks/github-app.ts) the dispatch order is:

1. `tryClaim()` — INSERT into `processed_events` (commits immediately)
2. dispatch to `handleInstallationEvent` / `handleMentionDelegation` / etc.
3. on error: `releaseClaimOnError` → DELETE the row (best-effort, in `.catch`)

If the process crashes between step 1 and step 3 (DB connection drop mid-handler, OOM kill, deploy mid-request), the `processed_events` row stays committed but the side effects never landed. GitHub redelivers with the same `x-github-delivery` UUID → the next handler call sees the existing row → returns deduped:200 → **event lost forever**.

Even within one process: if `unclaimEvent` itself fails (DB transient), the `.catch` logs but doesn't rethrow — same leak.

## Fix shapes (sketch)

1. **Mark-after-success**: do the work first, then INSERT the claim row. Risk: same delivery processed twice if a parallel retry lands during the work window. Mitigated by a per-delivery advisory lock or a `claim_attempt(deliveryId, attempt_id)` row that the handler completes.
2. **Claim with status + TTL**: insert with `status='pending', expires_at=now()+5m`. Successful handler updates to `status='done'`. Background sweep deletes expired pending rows so GitHub redelivery isn't blocked. Adds operational complexity.
3. **Wrap in a single transaction**: all handler side effects + claim row in one TX. Hard because some side effects (notifying recipients, log fan-out) aren't TX-bound.

Option 2 is probably the right balance.

## Affected

- App webhook (this PR's `webhooks/github-app.ts`)
- Per-org webhook on main (the `/webhooks/github/:orgId` route uses the same `claimEvent`/`unclaimEvent` pattern; this issue applies to it too historically)

liuchao-001:
## 🔥 GoF Competition scope & acceptance

This issue is part of the Goblet of Fire (GoF) competition pool. The scope and acceptance criteria below were aligned with the maintainer (@bestony) and are binding for competition submissions.

### Scope decision

[...]

PR #497, opened 2026-05-21
feat(chat): one message carries text + multiple attachments (route 2 / PG-bytea)

Gandy2025:
## Problem

In the hub chat view, adding attachments + text in one composer and sending produced **N+1 separate messages** (one per image as `format:"file"`, plus a separate `format:"text"`), instead of a single message — a clear "one send = one message" UX gap. Root cause: each message had a single mutually-exclusive `format` with no combined "text + attachments" shape.

## Solution

A single message now carries a **text caption + multiple attachments of any file type**, rendered as one bubble.

- **Storage — route 2 / PG-bytea** (follows the existing agent-avatar precedent; no new infra, fits "PostgreSQL only / stateless server"): bytes persist in a new `message_attachments` table; messages carry only references.
- **Download — authenticated fetch (no signed URLs)**: web uses the existing JWT (`fetch`→blob), the agent runtime uses its token; membership is re-checked per request. Avoids signed-URL churn against the 5s message refetch.
- **Data model — A′ (no new `format`)**: caption stays in `content`; attachments ride `metadata.attachments[]`. Old clients degrade gracefully (still render the caption) and the @mention guard stays on the existing text path.
- **Both composers unified**: chat-view and new-chat-draft both *upload each file then send one message* referencing them — one send → one message → one bubble. The old per-image split-send is removed.
- **Agent runtime**: reads `metadata.attachments`, downloads each (authenticated) to local disk, points the model at the path for its Read tool; a failed download yields an explicit "do not answer from it" placeholder (retryable — route 2 is durable).
- **Feishu adapter**: downgraded to a filename + size list (no internal links for external users); new `image_payload` WS frames are no longer produced (legacy read path retained for historical messages).

## Security (type / download / authorization double-gates)

- **C1** — type gate allows magicless safe text/docs (`.md/.csv/.json/...`) instead of rejecting "couldn't sniff"; rejects executables by magic bytes (MZ/ELF/Mach-O/shebang) and a denied-extension list.
- **C2** — download headers hardened: only a png/jpeg/gif/webp allow-list is served inline; everything else (incl. SVG) is `Content-Disposition: attachment` + `X-Content-Type-Options: nosniff` (anti-XSS).
- **C3** — at send, each referenced attachment must be uploaded by the sender, still unbound, and in the same chat.

### codex independent review — 3 blockers found and closed
1. **Forged `metadata.attachments`** — server now strips any client-supplied `attachments` key; only server-validated refs from `prepareAttachmentsForSend` are stored.
2. **Concurrent rebind race** — `SELECT … FOR UPDATE` row lock + a `RETURNING` row-count assertion on bind (mismatch → `ConflictError`, rolls back the send).
3. **Orphan GC not scheduled** — `gcOrphanedAttachments` wired into the 60s background sweep.
- Also added (non-blocking): unbound uploads are downloadable only by their uploader (open to all members once bound).

## Numbers & migration

- Limits: single file 10 MB / per-message total 25 MB / 9 attachments / 10 GB per-org quota.
- Migration `0048_message_attachments` (bytea column with `STORAGE EXTERNAL`; `kind` column carries a `CHECK ('image' | 'file')` constraint).

## Testing

- `pnpm typecheck` 11/11; biome clean.
- Full suite green: server 1112 / shared 271 / web 206 / client 431 / command 125; zero regressions; migration applied via testcontainers.
- New security tests: type gate (C1/C2), C3 ownership/binding, forged-metadata strip, concurrent-bind, orphan GC, unbound-download gate.

## Review chain

Independent review by `@codex-reviewer` (first round: 3 blockers → fixed → re-review **passed**).

Follow-up review pass with `@baixiaohang` (10 items, design-level) — decisions applied in this PR:
- Unified `getMessageAttachments` parser in shared (deduped 3 call sites).
- Feishu-specific format moved from `adapter-manager` into `services/feishu/format-message.ts`.
- Server-managed metadata keys centralised in `SERVER_MANAGED_METADATA_KEYS` / `stripUntrustedMetadataKeys` (shared) — replaces the ad-hoc `delete incomingMeta.attachments`.
- `kind` column gets a CHECK constraint; the read path re-derives `kind` from `mime` via `deriveAttachmentKind` so there is only one source of truth.
- Local design doc `docs/message-text-attachments-design.md` added, including the **Future evolution: rich-text + inline attachments** section (`` reuses the existing schema — no new `format`).

## Implicit contract changes

Centralising the server-managed metadata strip codifies an invariant that **the server is the sole writer of certain `metadata.*` keys**; clients setting them in a `POST /messages` body are silently stripped before insert.

Today's set is `attachments` and `editedAt` (see `SERVER_MANAGED_METADATA_KEYS` in `packages/shared/src/schemas/message.ts`). `attachments` is new in this PR; `editedAt` already had this behaviour de facto via the edit endpoint. `grep` on `main` confirms no external caller sets `metadata.attachments`, so this PR is the first to introduce the key and there are no in-tree consumers to migrate.

If you add a new server-stamped `metadata` key in the future, add it to that set — one line is enough; every write path that goes through `services/message.ts:sendMessage` will pick it up.

## Related

- Context Tree proposal `proposals/hub-message-text-attachments.20260521.md` (merged, PR #317).
- Local technical design doc `docs/message-text-attachments-design.md` (added in this PR — covers invariants, security gates, limits, and future evolution).

## Immediate follow-up — drizzle meta snapshot chain

`drizzle-kit generate` currently aborts on this branch — the drizzle/meta snapshot chain is broken (48 journal entries, 7 snapshots), so migration `0048_message_attachments.sql` is hand-authored. Recent migrations on `main` have the same workaround.

- **Owner**: @gandy-developer
- **Window**: within 7 days after this PR merges, open a dedicated PR to repair the snapshot chain.
- **Acceptance**: `drizzle-kit generate` succeeds on a fresh schema change; no future migration is hand-authored.
- **Long-term**: add a CI check (`drizzle-kit check` or equivalent) to fail PRs that break the snapshot chain again.

Until that repair lands, any PR that adds a table/column on this branch should pause and confirm whether the snapshot fix is complete; otherwise the workaround compounds.

## Known / follow-ups (non-blocking)

- ⚠️ **No live UI visual walkthrough** in the implementation environment (no browser) — recommend a manual pass on local dev before/after merge. Functional + type + unit/integration coverage is in place.
- **org-quota concurrent-overcommit window** — the quota check is non-atomic; two concurrent uploads within an org can both pass at the cap. Worst case is bounded (≤ single-file cap × concurrency); the obvious short fix (`SELECT … FOR UPDATE` on the org row) would serialise all uploads within an org, which is a worse trade-off than the rare overcommit. Properly addressed by a counter table / `org_quota_lock` row when real load demands it.
- **Stricter mime↔extension consistency** — currently soft (trust magic-byte sniff + deny-list); future hardening.

### DIFF
--- a/docs/message-text-attachments-design.md
+++ b/docs/message-text-attachments-design.md
@@ -1,0 +1,134 @@
+# Message text + attachments — local design doc
+
+Companion to Context Tree proposal `hub-message-text-attachments.20260521.md`
+(merged). This file is the hub-side reference: invariants you need to keep in
+mind when touching `services/message.ts`, `services/attachment.ts`,
+`api/chats.ts`, the web composer, or the agent runtime handlers.
+
+## Design at a glance
+
+- **Single bubble.** One send = one message = one bubble. Text caption +
+  arbitrary attachments ride together; the old "split per image" path is
+  gone.
+- **A′ on the wire.** Caption lives in `content`; attachments ride
+  `metadata.attachments[]` as refs (no base64, no client-local ids). No new
+  `format` value — old readers degrade gracefully (render the caption, ignore
+  the attachments) and the @mention guard stays on the existing text path.
+- **Route 2 / PG-bytea for storage.** Bytes persist in `message_attachments`
+  (bytea, `STORAGE EXTERNAL`); messages carry only references. No new infra,
+  follows the agent-avatar precedent, fits "PostgreSQL only / stateless
+  server".
+- **Authenticated fetch, no signed URLs.** Web uses the existing JWT, agent
+  runtime uses its token; membership is re-checked per request. Avoids
+  signed-URL churn against the 5s message refetch.
+- **Two-phase bind.** Upload returns an unbound `attachmentId` (`message_id
+  IS NULL`). At send time the server validates uploader / unbound / same-chat
+  (C3), folds authoritative refs into `metadata.attachments`, and binds in
+  the same transaction so a rollback drops both. `SELECT … FOR UPDATE` +
+  `RETURNING` row-count closes the rebind race.
+
+## Security double-gates
+
+- **C1** — type gate: magicless safe text/docs (`.md/.csv/.json/...`) are
+  allowed; executables (MZ/ELF/Mach-O/shebang) and a denied-extension list
+  are rejected.
+- **C2** — download response: only a png/jpeg/gif/webp allow-list is served
+  inline; everything else (incl. SVG) is `Content-Disposition: attachment` +
+  `X-Content-Type-Options: nosniff`.
+- **C3** — at send: each referenced attachment must have been uploaded by
+  the sender, still unbound, and in the same chat. Forged
+  `metadata.attachments` keys submitted by clients are stripped before
+  insert (see `SERVER_MANAGED_METADATA_KEYS` in shared).
+
+## Invariants worth not breaking
+
+- **`kind` is a pure function of mime.** Compute it via
+  `deriveAttachmentKind(mime)`; the DB CHECK constraint
+  (`message_attachments_kind_check`) defends against drift. Don't add a
+  `kind` parameter to any service entry point — the caller has no say.
+- **`metadata.attachments` is server-managed.** Listed in
+  `SERVER_MANAGED_METADATA_KEYS`; `stripUntrustedMetadataKeys` removes it
+  from every caller-supplied `metadata` before the insert. Adding a new
+  server-managed key means adding to that set, not adding a new ad-hoc
+  `delete`.
+- **`getMessageAttachments(metadata)` is the only parser.** Server, web,
+  agent runtime all share it (`packages/shared/src/schemas/message.ts`).
+  Don't reinvent the `safeParse → .data.attachments` dance locally.
+- **No internal links to external IM users.** Feishu downgrade is a
+  filename + size list, never URLs (no Hub session on the far side). Real
+  file forwarding via the Feishu upload API is a follow-up.
+
+## Limits
+
+- Single file ≤ 10 MB. Per-message total ≤ 25 MB. Max 9 attachments per
+  message. Per-org quota 10 GB.
+
+## Known follow-ups
+
+- **drizzle meta snapshot chain is broken.** `drizzle-kit generate` cannot
+  run; migration 0048 is hand-authored. Repairing the snapshot chain is the
+  immediate post-merge follow-up (see PR #497 description).
+- **org-quota overcommit window.** The quota check is non-atomic; two
+  concurrent uploads can both pass at the cap. Worst case is bounded
+  (single-file cap × concurrency); accepted for now because the obvious
+  short fix (`SELECT … FOR UPDATE` on the org row) serialises all uploads
+  within an org and is a worse trade-off than the rare overcommit.
+- **Mime↔extension consistency.** Currently soft (we trust the magic-byte
+  sniff plus the deny-list); future hardening can tighten by cross-checking
+  extension against mime category.
+
+## Future evolution — rich-text + inline attachments
+
+If the product later wants **inline attachments** — images interleaved with
+text in the caption, instead of stacked below it — **do not introduce a new
+`format` value**. The A′ contract works as-is:
+
+> Reuse `metadata.attachments[]` unchanged; in the caption use markdown
+> `![alt](attachment:<attachmentId>)` to reference an attachment that is
+> already in the same message's `metadata.attachments`.
+
+Concretely:
+
+```json
+{
+  "format": "markdown",
+  "content": "Look at this UI: ![mockup](attachment:att_1)\n…and the data flow: ![flow](attachment:att_2)",
+  "metadata": {
+    "attachments": [
+      { "attachmentId": "att_1", "mimeType": "image/png", "filename": "ui.png", "size": 12345, "kind": "image" },
+      { "attachmentId": "att_2", "mimeType": "image/png", "filename": "flow.png", "size": 23456, "kind": "image" }
+    ]
+  }
+}
+```
+
+Why this is the most restrained path:
+
+- **No new `format`.** A′'s core property — `format` describes presentation
+  intent, not data shape — stays intact.
+- **No DB-schema change.** `message_attachments` and `metadata.attachments`
+  are already what we need.
+- **Old clients keep working.** A client that doesn't recognise the
+  `attachment:` URI scheme renders the markdown as plain text; the
+  attachments still show in the existing bottom-of-bubble list (rendered
+  off `metadata.attachments`). Zero break.
+- **Adapters keep their downgrades.** Feishu's filename-list downgrade
+  works unchanged — it reads `metadata.attachments` and ignores the
+  caption's URI scheme.
+
+Rejected alternatives:
+
+- **A new `format: "rich-text"` carrying block arrays** (Feishu `post` /
+  Slack `blocks` style). This would require a new data model and parallel
+  read paths in every consumer (server adapter, web renderer, agent
+  runtime, feishu downgrade). It would also break older clients hard
+  rather than degrading them. The block-array model is what other
+  platforms built natively from day one — for us it's a migration
+  liability, not a feature.
+- **A new metadata key (e.g. `metadata.inlineSpec`) mapping caption
+  positions to attachment ids.** Doable but redundant with the markdown
+  URI approach, which already encodes the position implicitly.
+
+When you actually need this: the implementation work is in the rendering
+clients (markdown image renderer + URI scheme resolver), not on the wire
+contract.
--- a/packages/client/src/handlers/claude-code.ts
+++ b/packages/client/src/handlers/claude-code.ts
@@ -11,7 +11,7 @@ import type {
 import {
   type AttachmentRef,
   deriveRepoLocalPath,
-  messageAttachmentsMetadataSchema,
+  getMessageAttachments,
   type QuestionItem,
   type QuestionMessageContent,
   questionItemSchema,
@@ -128,13 +128,6 @@ async function writeLegacyImageToTempFile(content: LegacyImageFileContent, chatI
   const path = join(dir, `${Date.now()}-${randomUUID().slice(0, 8)}.${ext}`);
   await writeFile(path, Buffer.from(content.data, "base64"));
   return path;
-}
-
-/** Read the A′ attachment refs off a message's metadata, or [] when none. */
-function parseAttachments(metadata: Record<string, unknown> | null): AttachmentRef[] {
-  if (!metadata) return [];
-  const parsed = messageAttachmentsMetadataSchema.safeParse(metadata);
-  return parsed.success ? parsed.data.attachments : [];
 }
 
 /**
@@ -530,7 +523,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
     // an explicit "do not answer from it" placeholder so the model never
     // answers on missing content — and it's retryable (route 2 is durable),
     // not the permanent loss the old WS-push path implied.
-    const attachments = parseAttachments(message.metadata);
+    const attachments = getMessageAttachments(message.metadata);
     if (attachments.length > 0) {
       const caption = await sessionCtx.formatInboundContent(message);
       const blocks: string[] = [];
--- a/packages/client/src/handlers/claude-code.ts
+++ b/packages/client/src/handlers/claude-code.ts
@@ -9,7 +9,9 @@ import type {
   SupportedImageMime,
 } from "@agent-team-foundation/first-tree-hub-shared";
 import {
+  type AttachmentRef,
   deriveRepoLocalPath,
+  messageAttachmentsMetadataSchema,
   type QuestionItem,
   type QuestionMessageContent,
   questionItemSchema,
@@ -26,6 +28,7 @@ import type {
 import { query as claudeQuery } from "@anthropic-ai/claude-agent-sdk";
 import { z } from "zod";
 import type { AgentConfigCache } from "../runtime/agent-config-cache.js";
+import { findAttachmentPath, writeAttachmentFile } from "../runtime/attachment-store.js";
 import { bootstrapWorkspace, installFirstTreeIntegration } from "../runtime/bootstrap.js";
 import { type ChatContext, fetchChatContext } from "../runtime/chat-context.js";
 import { renderChatContextSection } from "../runtime/chat-context-section.js";
@@ -125,6 +128,34 @@ async function writeLegacyImageToTempFile(content: LegacyImageFileContent, chatI
   const path = join(dir, `${Date.now()}-${randomUUID().slice(0, 8)}.${ext}`);
   await writeFile(path, Buffer.from(content.data, "base64"));
   return path;
+}
+
+/** Read the A′ attachment refs off a message's metadata, or [] when none. */
+function parseAttachments(metadata: Record<string, unknown> | null): AttachmentRef[] {
+  if (!metadata) return [];
+  const parsed = messageAttachmentsMetadataSchema.safeParse(metadata);
+  return parsed.success ? parsed.data.attachments : [];
+}
+
+/**
+ * Ensure an attachment's bytes are on local disk (download + cache), returning
+ * its path — or null when the download fails (server unreachable, revoked
+ * access). Route 2 is durable, so a failure is retryable, not permanent loss.
+ */
+async function materialiseAttachment(
+  sessionCtx: SessionContext,
+  chatId: string,
+  att: AttachmentRef,
+): Promise<string | null> {
+  const cached = findAttachmentPath(chatId, att.attachmentId, att.filename);
+  if (cached) return cached;
+  try {
+    const bytes = await sessionCtx.sdk.downloadChatAttachment(chatId, att.attachmentId);
+    return await writeAttachmentFile(chatId, att.attachmentId, att.filename, bytes);
+  } catch (err) {
+    sessionCtx.log(`attachment download failed (${att.filename}): ${err instanceof Error ? err.message : String(err)}`);
+    return null;
+  }
 }
 
 function extractContentBlocks(message: unknown): unknown[] {
@@ -489,6 +520,39 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
           };
         }
       }
+    }
+
+    // Attachments (route 2 / A′): a regular text/markdown message whose
+    // `metadata.attachments[]` references files persisted server-side. Render
+    // the caption + preceding context exactly like a text message, then
+    // download each attachment to local disk and point the model at the path
+    // for its Read tool (image → vision, doc → text). A failed download yields
+    // an explicit "do not answer from it" placeholder so the model never
+    // answers on missing content — and it's retryable (route 2 is durable),
+    // not the permanent loss the old WS-push path implied.
+    const attachments = parseAttachments(message.metadata);
+    if (attachments.length > 0) {
+      const caption = await sessionCtx.formatInboundContent(message);
+      const blocks: string[] = [];
+      for (const att of attachments) {
+        const path = await materialiseAttachment(sessionCtx, message.chatId, att);
+        if (path) {
+          const what = att.kind === "image" ? "An image" : "A file";
+          blocks.push(
+            `${what} was shared in this chat. Use the Read tool to read it, then respond based on its content.\n\nFilename: ${att.filename}\nPath: ${path}`,
+          );
+        } else {
+          blocks.push(
+            `[attachment "${att.filename}" unavailable on this device — DO NOT answer based on its content; ask the user to resend if it is required.]`,
+          );
+        }
+      }
+      return {
+        type: "user",
+        message: { role: "user", content: [caption, ...blocks].join("\n\n") },
+        parent_tool_use_id: null,
+        session_id: sessionId,
+      };
     }
 
     // Default text content — sender attribution lives in the runtime so every
--- a/packages/client/src/runtime/attachment-store.ts
+++ b/packages/client/src/runtime/attachment-store.ts
@@ -1,0 +1,53 @@
+import { existsSync } from "node:fs";
+import { mkdir, writeFile } from "node:fs/promises";
+import { join } from "node:path";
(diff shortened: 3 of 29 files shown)


baixiaohang (commented):
## Design-level review: cohesion / non-breaking / soundness / future traps

整体设计扎实 —— route 2 / A′ / 两阶段 bind / C3 三重校验 + FOR UPDATE + RETURNING / per-request 重检都站得住。codex 那 3 条 blocker 已 close。下面是设计层面的一轮交叉对照（不阻塞 approve，但希望作者表态）。

### 1. Cohesion — `metadata.attachments` parser 三处复制 (强建议)

```
server/services/adapter-manager.ts → attachmentsFromMetadata()
web/pages/.../chat-view.tsx        → getMessageAttachments()
client/runtime/...                 → parseAttachments()  (handlers/claude-code.ts 附近)
```
三个同形异名函数都在 `messageAttachmentsMetadataSchema.safeParse(metadata) → .data.attachments`。建议在 `shared/schemas/message.ts` 里提供唯一出口：

```ts
export function getMessageAttachments(metadata: unknown): AttachmentRef[] {
  const parsed = messageAttachmentsMetadataSchema.safeParse(metadata);
  return parsed.success ? parsed.data.attachments : [];
}
```

每加一个 adapter（GitHub / 未来的 Slack/Discord adapter）都要再复制一份的成本会随时间放大。

### 2. Cohesion — feishu 降级逻辑硬编码在 adapter-manager (建议)

`formatForFeishu` 里写死了 `"📎 N attachment(s):\n• filename (size)"` 这段。建议把"对 `metadata.attachments` 怎么消费"下推到各 adapter 自己 —— 否则未来 GitHub adapter 或别的适配器要支持附件，都要回到 manager 里改一份。

### 3. Cohesion 隐含 — `metadata` 是无约束字典袋 (讨论)

`metadata: z.record(z.string(), z.unknown())`，目前键有 `mentions / documentContext / systemSender / editedAt / attachments`。每加一个键都要：服务端 strip 不可信键 + 客户端 safeParse + 各 adapter 各写一次降级。第三/第四个新键出现时，C3 这种 strip 逻辑会到处补且容易漏。

是否考虑过给 metadata 引入显式 schema（每个键标注 trust source 是 user / server / runtime），或者一个 `SERVER_MANAGED_METADATA_KEYS: Set<string>` 让 `stripUntrustedMetadataKeys` 一次性管掉？

### 4. Non-breaking — 隐式契约变更没在 PR 描述里 spell out (建议补文档)

```ts
// services/message.ts
const incomingMeta = { ...((data.metadata ?? {}) as ...) };
delete incomingMeta.attachments;
```

这是 C3 加固的必要副作用，但**对老调用方是个静默契约变更**：如果 main 上有谁出于其他目的在 metadata.attachments 写过东西，现在会被默默吃掉、零错误提示。建议：
- PR 描述里把这条列入 "implicit contract changes"
- grep 确认 main 上没有合法使用者（应该没有，但值得 spell out）

### 5. Soundness — 单 multipart 一次只能传一个文件 (讨论作者意图)

`multipart 注册：limits: { files: 1 }` —— 9 个附件 = 9 次 upload + 1 次 send = **10 次 RTT**。弱网下发 9 张图体验会很差，且任何一张失败 UI 要单独重试。Slack/feishu/Discord 都支持单次 multipart 多文件。

这是有意的简化（PR 描述没明说原因），还是只是 first-cut？未来加进度条 / 部分失败重试时，单文件多请求模型会比批量更复杂。

### 6. Soundness — PG bytea 架构天花板没在文档里 (建议补文档)

bytea STORAGE EXTERNAL（不压缩）+ 10 GB/org = 实际占盘 10 GB + WAL/备份放大 ≈ 30~50 GB I/O 开销/单 org。当前阶段 org 数量少没问题，但 route 2 的天花板是真实的。

建议在 `docs/message-text-attachments-design.md` 里写明：**在什么触发条件下要迁移到对象存储**（org 数量阈值 / 单 org 数据量阈值 / 备份耗时阈值），避免未来在着火状态下做架构决策。

### 7. Soundness — A′ 的代价没写明：未来富文本混排路径 (建议补文档)

当前是「caption 在上、附件列表在下」固定二段式。未来如果产品要做「文本流中插入图片」（markdown `![]` 引用同条消息的附件），需要协议演进。

建议在设计文档里写明一条演进路径："未来富文本混排走 markdown `` 引用 `metadata.attachments[].attachmentId`，**不需要再加 format**"。这样下一个改这块的人不用重新思考。

参考：feishu 的 post 类型走的是段落+tag 元素结构，Slack 的 blocks 走的是块数组顺序，都需要全新数据模型。当前 A′ 走 markdown 引用是最克制的演进路径。

### 8. Future trap — `kind` 字段在 DB 是无约束 text + 运行时兜底 (建议)

```ts
kind: text("kind").notNull(),  // 期望 "image" | "file"

// service 读取兜底:
const kind = row.kind === "image" ? "image" : "file";
```

兜底是好事，但 "DB 无约束 + 运行时强转" 比 "DB CHECK 拒绝写入" 更脆弱。而且 `kind` 其实可以从 mime 通过 `deriveAttachmentKind` 算出来 —— **要么加 CHECK 约束，要么直接删字段从 mime 派生**（DRY）。建议后者。

### 9. Future trap — drizzle meta snapshot 链断了 (强建议升级 follow-up 优先级)

PR 描述里承认：「`drizzle-kit generate` 当前不可用 — migration 0048 is hand-authored」。

这意味着**未来任何人加 column / index 都会撞到这个断链**。这不是 "known follow-up 放着" 的级别，是**每个加表的人都要踩**的 DX 坑。

建议：作为 follow-up 头号优先级，**明确 owner + 时间窗**（merge 后第一周内）。

### 10. Future trap — org 配额竞争窗口的时间线 (建议给个 sloppy 短期方案)

PR 描述里自承的 follow-up，但没有给时间线。短期内可以在 createAttachment 事务里对 org row 做 `SELECT ... FOR UPDATE`（即使略糙也比没有强），避免在 follow-up 落地前长期裸奔。

---

## 结论

无阻塞项。建议优先级：**强建议 1, 2, 7, 9**（重构 + 文档 + 升级 follow-up）；**讨论 3, 5**（设计方向）；**文档补充 4, 6**；**小重构 8**；**短期补强 10**。

希望听到作者对 1、2、3、5、9 的回应再考虑 approve。

Gandy2025:
## Reply to "single multipart only carries one file at a time" (5)

@baixiaohang Thanks for the detailed write-up. To answer the implicit
question — the single-file `multipart` limit was a deliberate design
choice, not a first-cut workaround, and I'd like to keep it. Reasoning:

The wire model here is **two-phase**: `POST /chats/:id/attachments` returns
an unbound `attachmentId`, then `POST /chats/:id/messages` references those
ids and the server validates + binds in one transaction (C3). A batched
multipart wouldn't shorten that — you'd still need the second call to bind
— and it would force the server to decide on partial-success semantics
(fail-fast vs best-effort with a per-file result map) that the per-file
model gets for free.

The complexity arguments cut the opposite direction from the Slack/Feishu
analogy because those platforms use a **single-phase** model (upload =
send). For us:

- **Progress UI**: N independent uploads = N independent progress events,
  trivially mapped per-card. A batched upload requires server-side
  chunked progress reporting (or WebSocket-side push) to give the same
  granularity.
- **Partial-failure retry**: each independent request already exposes
  "which one failed"; the user retries that card. Batched would need a
  partial-success response protocol and client-side reconciliation.
- **C1 + C3 validation**: each upload is a linear, isolated validation
  pass. Batched forces a fail-fast-vs-best-effort policy decision, with
  knock-on transaction boundary questions when a row mid-batch trips C1.

The cost is N+1 RTTs for N attachments. In the long tail (>5 attachments
in a single send) this is real, but it's the edge — typical compose has
1–3. If we see telemetry showing the long tail is hot, we can revisit and
introduce a batched endpoint *alongside* the per-file one without
disturbing the two-phase contract; nothing here forecloses that.

Gandy2025:
## Reply to "PG bytea ceiling deserves a doc with migration triggers" (6)

@baixiaohang Agreed that the ceiling is real and route 2 is explicitly
"first-served-best-served" by design. Pushing back gently on the doc
itself, though: a doc that prescribes specific thresholds without a
matching monitoring + on-call owner becomes dead weight — by the time
backup duration would actually matter, whoever wrote those numbers won't
be the one watching, and a written threshold without an owner makes the
team feel covered when they aren't.

The honest current state is that we don't have anyone watching per-org
attachment growth or PG backup duration on a recurring cadence. Picking
thresholds today would either be guessing (no data) or boilerplate (no
owner to act on them). When the first signal of actual capacity pressure
appears — or when someone takes on DB-monitoring ownership — that's the
right time to write the migration plan; we'd then have real numbers and a
human to consult them.

Happy to revisit if that ownership shifts. For now I'd rather not ship a
doc that creates a false sense of preparedness.

Gandy2025:
## Reply to "org-quota overcommit window" (10)

@baixiaohang Good catch on the concurrent-overcommit window — it is real,
not theoretical. Quantifying the worst case for context, then explaining
why I'm not taking the row-lock shortcut:

**Worst case**: bounded by single-file-cap × concurrent-uploads-per-org.
At the 10 MB single-file cap, even pathological concurrency overshoots by
less than 100 MB per spike; against the 10 GB quota, that's
sub-1%-of-budget and self-corrects on the next non-overlapping upload
(subsequent quota checks see the inflated total and reject). No data
loss, no security surface, just a transient over-cap.

**Why not `SELECT … FOR UPDATE` on the org row**: the suggested patch
*does* close the window, but it serialises **every** upload within an org
behind a single row lock. Today our concurrent-upload-per-org count is
low, so the lock contention would be invisible most of the time — but the
moment two users in the same org compose with attachments at the same
moment, the second one's upload p95 jumps to "first upload's full
duration". That's a worse user-visible regression than the rare
sub-1%-of-quota overshoot.

**Proper fix when needed**: a per-org counter table (`org_attachment_usage`
with bump-and-check semantics) or a dedicated `org_quota_lock` row that
upload paths can acquire with bounded fairness. Both are
non-blocking-by-default and only contend on the actual quota-near-cap
case. Worth doing once we see real per-org overshoot events; jumping to
either now is over-engineering for the current load profile.

I've added this rationale to the PR description under
"Known / follow-ups" so future readers don't have to re-derive it.

Gandy2025:
## Disposition per item — commit `feb0e2f4`

@baixiaohang Thanks for the detailed review. Going through every item so the
disposition is explicit (items 5 / 6 / 10 already have their own reply
comments above; the rest are summarised here with file refs so you can spot-
check).

### (1) Unified `metadata.attachments` parser — ✅ applied

Added `getMessageAttachments(metadata: unknown): AttachmentRef[]` in
`packages/shared/src/schemas/message.ts`. The three duplicate sites
(`adapter-manager.ts:attachmentsFromMetadata`, `chat-view.tsx:
getMessageAttachments`, `claude-code.ts:parseAttachments`) are removed and
all callers now use the shared export. Re-exported from
`packages/shared/src/index.ts`.

### (2) Feishu downgrade pushed into the adapter — ✅ applied (scope-limited)

`formatForFeishu` + `formatAttachmentSize` moved to
`packages/server/src/services/feishu/format-message.ts`. `adapter-manager.ts`
now `import { formatForFeishu } from "./feishu/format-message.js"` and the
~60 lines of Feishu-specific formatting are gone from the manager.
Deliberately did **not** introduce an adapter-capability interface /
registry yet — only one adapter today, and abstracting at N=1 tends to
miscarve the boundary. When the second adapter (GitHub) lands and we have
two real shapes to compare, the interface will fall out naturally.

### (3) Centralised server-managed metadata strip — ✅ applied (Option B)

Took your Option B path:

```ts
// packages/shared/src/schemas/message.ts
export const SERVER_MANAGED_METADATA_KEYS = new Set<string>(["attachments", "editedAt"]);

export function stripUntrustedMetadataKeys(metadata: Record<string, unknown>): Record<string, unknown> {
  const result = { ...metadata };
  for (const key of SERVER_MANAGED_METADATA_KEYS) delete result[key];
  return result;
}
```

`services/message.ts` now calls `stripUntrustedMetadataKeys(...)` instead of
`delete incomingMeta.attachments`. Future server-managed keys = one-line
addition to the set.

The current set is `{attachments, editedAt}` only — those are the keys the
server actually stamps today. `systemSender` is not in the codebase (only
`metadata.source` is, used for Feishu echo-suppression, which is itself
caller-supplied and lives outside this trust boundary). If a new
server-stamped key shows up, it goes into the set with one line.

Option A (typed `metadata` with per-key trust source) is more rigorous but
its cost is a metadata-type rewrite that we'd be paying with only 5 known
keys — undersized for the abstraction. Worth reopening at 6–8 keys or when
a third trust source (e.g. runtime-stamped) appears.

### (4) Implicit contract change spelled out — ✅ applied

PR description has a new **"Implicit contract changes"** section listing
the server-managed key behaviour. `grep -rn "metadata\.attachments"` on
`main` (b48178f7) returned **zero matches** — this PR is the first to
introduce the key, so there are no in-tree consumers to migrate.

### (5) Single multipart per file — see [dedicated reply](https://github.com/agent-team-foundation/first-tree/pull/497#issuecomment-4551230646)

Maintained by design (two-phase bind + per-file independence is cheaper
than batched-with-partial-success-protocol). Full rationale in the linked
reply.

### (6) PG bytea ceiling doc — see [dedicated reply](https://github.com/agent-team-foundation/first-tree/pull/497#issuecomment-4551231636)

Declined for now (no monitoring owner to give the thresholds teeth). Full
rationale in the linked reply.

### (7) Future-evolution path for rich-text + inline attachments — ✅ applied

Added `docs/message-text-attachments-design.md` (the PR description said
this existed but it didn't — created in this commit). It captures
invariants, the security gates, limits, follow-ups, and ends with a
**Future evolution: rich-text + inline attachments** section that locks in
the markdown `` approach:

> Reuse `metadata.attachments[]` unchanged; in the caption use markdown
> `alt` to reference an attachment that is
> already in the same message's `metadata.attachments`. **Do not introduce
> a new `format` value.**

Includes the rejected-alternatives section (block-array models like Feishu
`post` / Slack `blocks` would force a new data model + parallel read paths
and break old clients hard rather than degrading them).

### (8) `kind` column — ✅ applied (CHECK + single-source derive, both)

Did both halves of the trade-off:

- **Migration 0048** got `CONSTRAINT "message_attachments_kind_check"
  CHECK ("kind" IN ('image', 'file'))`. The drizzle schema
  (`db/schema/message-attachments.ts`) carries the matching `check(...)`
  expression so the next `drizzle-kit generate` (after the snapshot-chain
  repair) sees a consistent state.
- **Service layer** now treats kind as a pure function of mime: the write
  path explicitly does `kind = deriveAttachmentKind(classification.mimeType)`
  and the read path (`prepareAttachmentsForSend`) re-derives from
  `row.mime` instead of trusting `row.kind`, so the returned ref can never
  disagree with the persisted mime. The DB CHECK is the belt; the
  single-source derive is the suspenders.

Decided against the "delete the column outright" option B because some
near-future queries (e.g. orphan-GC dashboards by media type, capacity
breakdowns) want to filter by `kind` directly without `mime LIKE 'image/%'`
on bytea-bearing rows.

### (9) drizzle snapshot chain — ✅ elevated to Immediate follow-up

PR description now carries a **"Immediate follow-up — drizzle meta snapshot
chain"** section with:

- **Owner**: @gandy-developer
- **Window**: within 7 days after this PR merges
- **Acceptance**: `drizzle-kit generate` succeeds on a fresh schema change;
  no future migration is hand-authored
- **Long-term**: CI check (e.g. `drizzle-kit check`) to fail PRs that
  break the chain again

Treating it as the first thing after merge, not as a backlog item — agree
the "every contributor adding a column trips this" angle makes it more
expensive than its current "known follow-up" wording suggested.

### (10) org-quota overcommit — see [dedicated reply](https://github.com/agent-team-foundation/first-tree/pull/497#issuecomment-4551232242)

Quick `FOR UPDATE` patch declined (worst-case overshoot is sub-1% of quota
and self-corrects; the patch would serialise all uploads within an org for
no proportional benefit). Counter-table / `org_quota_lock` row when real
load demands. Full quant in the linked reply.

---

That's all 10. **Diff stat**: 11 files, +296 / −89. **Verification**:
typecheck 11/11; biome clean; full test suite green (server 1112 / shared
271 / web 206 / client 431 / command 125 = **2145 pass, zero regression**);
codex review on the uncommitted diff returned no actionable issues.

Happy to iterate on any of the discussion items (3 / 5 / 6 / 8 / 10) if
the reasoning above doesn't match where you wanted to land.

baixiaohang (approved):
## Approve — 10 项 disposition 复核通过

二轮复核 commit `feb0e2f4` 对上一轮 design-level review 的 10 项 disposition：

**Applied 项（1, 2, 3, 4, 7, 8, 9）与代码一致：**

- (1) `getMessageAttachments` 单一出口 — `shared/src/schemas/message.ts`，server/web/agent 三处旧拷贝已切换 ✅
- (2) `formatForFeishu` 下沉到 `services/feishu/format-message.ts`，`adapter-manager.ts` 净减 26 行 ✅
- (3) `SERVER_MANAGED_METADATA_KEYS` + `stripUntrustedMetadataKeys` 集中在 shared；`services/message.ts` 已切换 ✅
- (4) PR 描述 "Implicit contract changes" 段落 + `grep main` 确认 no in-tree consumer ✅
- (7) `docs/message-text-attachments-design.md` 新建，含 `` markdown URI 演进路径 + rejected alternatives ✅
- (8) belt + suspenders 两半都做：migration 的 `message_attachments_kind_check` CHECK + drizzle schema 的 `check()` + service 读路径 `deriveAttachmentKind(row.mime)` ✅
- (9) drizzle snapshot 修复升级到 "Immediate follow-up"：Owner @gandy-developer / 7 天窗口 / 验收 / 长期 CI check ✅

**Declined 项（5, 6, 10）论据扎实且都在 non-blocking 范围：**

- (5) 单 multipart 一文件 — 两阶段 bind 是核心契约，per-file 模型免费拿到独立进度 / 独立重试；等 telemetry 看到长尾热再加 batched endpoint 并行
- (6) bytea 天花板阈值文档 — 接受"没监控 owner 的阈值文档 = 死文档"的组织判断
- (10) org 配额 FOR UPDATE 短期补强 — worst-case <1%/quota + 自纠正；FOR UPDATE on org row 会序列化全部 org 内上传，引入更糟的 p95 退化；proper fix 是 counter table

**额外核查：** `gcOrphanedAttachments` 已接入 60s background sweep；chat-view + new-chat-draft 两个 composer 统一走 `uploadChatAttachment` + `sendChatMessageWithAttachments`（旧的 `sendFileMessage` 已删）；agent runtime 下载失败 placeholder 到位；migration 含 `STORAGE EXTERNAL`。

**Merge 前提醒：**

1. PR 当前 `mergeable: CONFLICTING` — 需要 rebase `main` 解决冲突
2. (9) drizzle snapshot 修复在 merge 后 7 天窗口跟 @gandy-developer 跟进

Gandy2025:
# Status update — closing this PR

@baixiaohang, follow-up: after pushing the review-disposition commit on
2026-05-27 we tried to sync this branch with `main` so the PR would no
longer be `CONFLICTING`. The merge surfaced a finding that we missed before
asking for your review.

## What we found while syncing

`main` has moved well past this PR's base since 2026-05-21:

- The monorepo merge (PR #510) **renamed every package**:
  - `@agent-team-foundation/first-tree-hub-shared` → `@first-tree/shared`
  - `@first-tree-hub/{server,client,web}` → `@first-tree/{server,client,web}`
- `packages/command/` was moved to `apps/cli/`.
- Observability constants were renamed (`FIRST_TREE_HUB_ATTR` →
  `FIRST_TREE_ATTR`).
- `chat-view.tsx` grew by ~780 lines on `main` due to the composer
  redesign in #515 / #521 / #524 / #577 / #588 — overlapping the same
  surface this PR touches.

More importantly, we revisited what `main` already provides for image
messages and **found the infrastructure this PR was building was already
in place**, just consumed in a way that produces N+1 messages on the
composer side:

- `MESSAGE_FORMATS.FILE` + `ImageRefContent` already encodes
  `{ imageId, mimeType, filename, size }` — the reference model this PR
  was reinventing under `metadata.attachments`.
- The single-phase endpoint `POST /chats/:chatId/messages` already accepts
  these refs end-to-end; the `image_payload` WebSocket frame already
  carries the bytes; `requireChatAccess` already gates membership.
- The web composer already supports `<input multiple>` with a
  `pendingImages` array.

The actual UX bug is `handleSend` in `chat-view.tsx` iterating over the
pending images and calling `sendFileMessage` once per image, producing N+1
rows. Fixing that on top of the existing infrastructure is ~50 lines
across 5 files (schema accepts an array + caption, server outbound loops
over the array, composer drops the for-loop, renderer recognises the new
shape). The new `message_attachments` table, the two-phase bind, and the
dedicated upload endpoint in this PR aren't needed for the UX fix —
they're an over-design relative to what `main` already provides.

That's a call we should have made before asking you to review. Sorry for
the time spent on disposition items that lose most of their relevance in
the lighter design.

## What's not lost

Several pieces of your review carry over directly to the lighter PR:

- **#3 SERVER_MANAGED_METADATA_KEYS** — the "centralised strip" idea is
  the right shape if/when the lighter PR or any successor writes
  server-stamped keys to `metadata`.
- **#7 Future evolution (``)** — the markdown
  reference path stays valid as the evolution route for inline image
  positioning, whether the underlying ref lives in `content.attachments`
  (light) or `metadata.attachments` (the original A′ shape).
- **#9 drizzle meta snapshot chain** — independent of attachments; it's a
  DX issue on `main` today, so we're pulling that fix out into its own
  small PR.

The local design doc at `docs/message-text-attachments-design.md` will
move into the lighter PR as background; the Future-evolution section is
worth keeping in tree regardless of which datastore design ships.

## Closing

Closing this PR. The composer-side N+1 fix will come as a new PR against
`main`, and the drizzle snapshot repair as a separate small PR. Will link
both here once they're up so the thread stays connected to your review.
Thanks for the careful pass — much of the reasoning still applies.

Gandy2025:
Follow-up: lighter composer-only fix opened as #611 (`feat(chat): collapse N+1 file-message sends into a single batched message`). It carries the two review-takeaways from this thread that survived the redesign (#3 SERVER_MANAGED_METADATA_KEYS spirit, #7 markdown evolution path noted in the PR description), and references this PR as the closed antecedent. The drizzle snapshot fix (#9) is being pulled out into a separate small PR. cc @baixiaohang

PR #499, opened 2026-05-21
fix(web): instant composer + cap activity refetch storm

baixiaohang:
## Summary

Two independent fixes for user-perceived performance issues in the web dashboard, surfaced together because the second was discovered while investigating reports of slowness.

### 1. Composer felt stuck for 1–2s on every send (`perf(web)`)
The send path was strictly pessimistic — the user's own message appeared above the input only after `POST /messages` + the follow-up refetch of `GET /messages` had both completed. Optimistic-update wrappers around the text and image send paths now render a `deliveryStatus: "pending"` row instantly, reconcile with the server row on success, and roll back + restore the draft on failure. Reuses the existing `MessageWithDelivery` shape and `pending` enum value.

### 2. `/api/v1/orgs/:orgId/activity` hit dozens of times per second (`fix(activity)`)
Reported by @baixiaohang. Root cause was a NOTIFY leak in `upsertSessionState`: every chat message triggers the predictive Step 1b in `services/message.ts`, which calls `upsertSessionState(..., "active", ...)`. The DB UPDATE was already short-circuited by `setWhere`, but the NOTIFY ran unconditionally — so an agent emitting many intermediate tool-call messages on an already-active session blasted N session:state frames per N messages. The admin WS then re-broadcast each frame into a non-throttled invalidation of `["activity"]` / `["sessions"]`, which 5+ workspace components each independently refetch.

Two layers of defence:
- **Server**: `upsertSessionState` now uses `.returning()` on the ON CONFLICT DO UPDATE. Zero rows back means the same-state no-op took the `setWhere` branch — we skip the presence refresh, the lastSeenAt touch, and the NOTIFY. (lastSeenAt is owned by the client's `heartbeat` frame anyway — `presence.ts:touchAgent`.)
- **Client**: Generalize the existing `me/chats` throttle helper into a factory; give `["activity"]` / `["sessions"]` / `["me", "chats"]` each their own 1s leading+trailing pair. The ws:reconnect catch-up invalidations stay non-throttled — those are one-shot syncs.

## Test plan
- [x] `pnpm check` (biome) — 803 files, no errors
- [x] `pnpm --filter @first-tree/server typecheck` + `pnpm --filter @first-tree/web typecheck`
- [x] Server tests: 1108/1108 passing (140 files), including two new cases in `activity.test.ts`:
  - `does NOT NOTIFY or refresh lastSeenAt when state is unchanged`
  - `DOES NOTIFY when state actually transitions`
- [x] Web tests: 206/206 passing (22 files)
- [ ] Manual: send a text + image in chat-view, confirm the composer clears + the row appears with no perceptible delay
- [ ] Manual: open the workspace with a busy agent, watch DevTools network — `/activity` should hit at most ~2 req/s during a tool-call burst, not 20+/s

## Notes
- No `package.json` `version` changes (per repo policy)
- No DB migrations
- The reconnect-time broad invalidate path is intentionally untouched

### DIFF
--- a/packages/server/src/__tests__/activity.test.ts
+++ b/packages/server/src/__tests__/activity.test.ts
@@ -1,7 +1,8 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { upsertSessionState } from "../services/activity.js";
 import { createAgent } from "../services/agent.js";
 import { createChat } from "../services/chat.js";
+import type { Notifier } from "../services/notifier.js";
 import { createAdminContext, useTestApp } from "./helpers.js";
 import { readPresence, seedPresence } from "./session-state-helpers.js";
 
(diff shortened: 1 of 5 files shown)


yuezengwu (approved):
LGTM — 两个修复都收口干净，没有 schema/迁移变更。

## 关键审查点 ✅
- **Postgres 语义正确**：`ON CONFLICT DO UPDATE ... WHERE <setWhere> RETURNING` 在 setWhere 为 false 时不返回行 —— `rows.length === 0` 判定同状态 no-op 是正确的。
- **lastSeenAt 跳过安全**：由 client 的 `heartbeat` 帧通过 `presence.ts:touchAgent` 维护，server 端跳过 presence 刷新不会丢数据。
- **测试一正一反**：`does NOT NOTIFY ... when state is unchanged` + `DOES NOTIFY when state actually transitions` 正好防御"过度短路"回归。
- **客户端节流工厂**：每个 queryKey 独立 leading+trailing 对，bursts 不会互相饿死；ws:reconnect 一次性同步路径保留不节流是合理的。
- **乐观更新形状对齐**：`chat-messages` 是普通 `useQuery`（非 infinite），扁平 `PaginatedMessages` 的 `setQueryData` 写法正确；`sendChatMessage` 返回的 `Message` 与 `MessageWithDelivery` 兼容（后者只是多了可选 deliveryStatus）。

## 非 blocker 观察（可选）
1. **错误回滚可能覆盖用户新输入**：发送中若用户继续在 textarea 输入新内容，失败时 `setDraft(previousDraft)` 会盖掉新输入。乐观 UX 通用取舍，可接受。
2. **trailing timer 内 `latestQc` 引用**：使用模块级引用而非 onMutate 时的 qc——属于原有模式延续，保持一致即可。

Ship it.

yuezengwu:
Follow-up commit 475902f LGTM —— functional setState `(current) => current === "" ? previousDraft : current` 正确处理了 in-flight 期间用户继续输入的情况，textarea 还为空才恢复 previous draft，否则保留用户新输入。两条错误路径（sendMut.onError 与 image catch）处理一致。previous approval 继续有效。

PR #383, opened 2026-05-14
feat(web): split admin Settings → Computers into Your / Team groups

baixiaohang:
## Summary
- Admin Settings → Computers now shows two sections: Your computers + Team computers (sorted by lastSeenAt desc), with an Owner column resolved via listMembers.
- Team rows are read-only and not expandable — admin can't fire owner-scoped GET /clients/:id (avoids 403 + misleading "not reported" capability fallback).
- Member view is pixel-equivalent to before (still hits /me/clients, no Owner column, no group headers).
- Adds banner + fallback to /me/clients when /orgs/:orgId/clients fails.

## Design doc
`/home/bai/.first-tree/hub/data/workspaces/architect/9a812de8-4299-436e-8115-77dbf5fa05e6/docs/admin-computer-team-view-design.md`

## Test plan
- [x] `pnpm check`
- [x] `pnpm --filter @first-tree-hub/web typecheck`
- [x] `pnpm --filter @first-tree-hub/web test` (111/111)
- [ ] Manual smoke: admin view shows Your + Team sections, team row is read-only & not expandable
- [ ] Manual smoke: member view unchanged (no Owner column, no group headers)
- [ ] Manual smoke: simulate `/orgs/:orgId/clients` 5xx → fallback banner + single-section view

PR #505, opened 2026-05-22
fix(web): detect reconnects in "+ New Connection" modal via connectedAt

yuezengwu:
## Problem

In the `+ New Connection` modal on `/clients`, after the user runs the connect command on a previously-paired machine, the modal stays in **Waiting…** forever even though the client is back online. Polling is happening (`/me/clients` every 3s) but the UI never flips to **Connected**.

## Root cause

The wait-loop detector matched on **id-set diff**:

```ts
const baseline = baselineRef.current;             // ids snapshotted at open
const arrived = fresh.find(
  (c) => !baseline.has(c.id) && c.status === "connected" && c.userId === user.id,
);
```

But `client.id` is **stable per machine** — auto-generated once and persisted to `client.yaml` (`packages/shared/src/config/client-config.ts:17-25`):

```ts
client: {
  // Stable per-machine client identifier. Auto-generated on first start
  // and written back to client.yaml so the SDK keeps the same id across
  // restarts — agents pin to `clients.id` on the server, so a fresh
  // random id every run would orphan every pinned agent.
  id: field(z.string().regex(/^client_[a-f0-9]{8}$/), { auto: "client-id" }),
}
```

Reconnecting a previously-paired machine reuses the same row via `ON CONFLICT DO UPDATE` (`packages/server/src/services/client.ts:113-126`), only flipping `status` from `disconnected` → `connected` and rewriting `connectedAt`. The baseline already contained that id, so `!baseline.has(c.id)` was always `false` → `find()` never matched → modal never transitioned.

Two related drifts made this stick:

- After [#383](https://github.com/agent-team-foundation/first-tree-hub/pull/383) split `ClientsPage`'s query into `["clients", "org"]` / `["clients", "me"]`, the modal's `getQueryData(["clients"])` returned `undefined` on open. The topbar chip (`useDisconnectedComputers`) is now the only thing populating `["clients"]`, so the baseline was incidental rather than intentional.
- Only true on a host with no prior `client.yaml` (brand-new machine) did the old path work — which is exactly what manual testing tends to exercise.

## Fix

Switch detection from id-set diff to a `connectedAt >= openedAt` timestamp comparison. `connectedAt` is set to `NOW()` on **both** the insert branch and the `ON CONFLICT DO UPDATE` branch, so it captures every fresh handshake regardless of whether the row is new. A 1s fudge absorbs minor browser↔server clock skew.

Detection rule now:

```ts
clients.find((c) =>
  c.status === "connected" &&
  c.userId === userId &&
  c.connectedAt !== null &&
  new Date(c.connectedAt).getTime() >= openedAt,
);
```

Extracted as `selectArrivedClient` (pure) so the unit test can pin the contract — matches the local convention from `selectDisconnectedComputers` in `use-disconnected-computers.ts`.

## Test plan

- [x] `pnpm --filter @first-tree/web test` — 216/216 passing (10 new cases in `selectArrivedClient.test.ts`)
- [x] `pnpm --filter @first-tree/web exec biome check src/pages/clients/`
- [ ] Manual: open `+ New Connection`, run command on a **brand-new** machine → modal flips to green within ~3s
- [ ] Manual: open `+ New Connection`, run command on a **previously-paired** machine that is currently disconnected → modal flips to green within ~3s (the regression case)
- [ ] Manual: open `+ New Connection` with an already-connected machine sitting in your list, do NOT run the command → modal stays in Waiting (no false success on historical `connectedAt`)

### DIFF
--- a/packages/server/src/__tests__/me-clients-reconnect.test.ts
+++ b/packages/server/src/__tests__/me-clients-reconnect.test.ts
@@ -1,0 +1,158 @@
+import { eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import { clients } from "../db/schema/clients.js";
+import * as clientService from "../services/client.js";
+import { createAdminContext, useTestApp } from "./helpers.js";
+
+/**
+ * Server-side regression net for the "+ New Connection" modal fix
+ * (`packages/web/src/pages/clients/new-connection-dialog.tsx`).
+ *
+ * The web modal flips Waiting→Connected by polling `GET /me/clients` and
+ * matching the predicate
+ *
+ *   c.status === "connected" && c.userId === me && c.connectedAt >= openedAt
+ *
+ * The predicate is unit-tested as a pure function (`selectArrivedClient`).
+ * What this file pins is the *server contract* the predicate relies on:
+ * when the CLI re-registers an already-known `client_id` (i.e. the
+ * `ON CONFLICT DO UPDATE` branch of `registerClient`, hit on every
+ * "machine that was previously paired and now reconnects"), `connectedAt`
+ * is rewritten to the new handshake time AND that timestamp survives the
+ * `/me/clients` wire serialization.
+ *
+ * Without this, an id-set baseline (the previous web implementation) was
+ * silently correct only for brand-new machines. Reconnects reused the
+ * same `clients.id` row, so the diff was empty and the modal hung in
+ * Waiting forever. See PR #505 for full root-cause analysis.
+ */
+
+const FUDGE_MS = 1_000;
+
+describe("GET /me/clients — reconnect path rewrites connectedAt", () => {
+  const getApp = useTestApp();
+
+  it("re-registering an existing client_id stamps connectedAt to NOW and the wire reflects it", async () => {
+    const app = getApp();
+    const ctx = await createAdminContext(app);
+    const clientId = `cli-rc-${crypto.randomUUID().slice(0, 8)}`;
+
+    // First handshake — INSERT branch of registerClient. Mirrors the very
+    // first `first-tree-hub connect <token>` on a fresh machine.
+    await clientService.registerClient(app.db, {
+      clientId,
+      userId: ctx.userId,
+      organizationId: ctx.organizationId,
+      instanceId: "test-instance",
+    });
+
+    const first = await app.inject({
+      method: "GET",
+      url: "/api/v1/me/clients",
+      headers: { authorization: `Bearer ${ctx.accessToken}` },
+    });
+    expect(first.statusCode).toBe(200);
+    const firstRow = (first.json() as Array<{ id: string; status: string; connectedAt: string | null }>).find(
+      (c) => c.id === clientId,
+    );
+    if (!firstRow?.connectedAt) throw new Error("first registration did not stamp connectedAt");
+    expect(firstRow.status).toBe("connected");
+    const connectedAt1 = Date.parse(firstRow.connectedAt);
+
+    // Simulate WS close — status flips to disconnected. `connectedAt` is
+    // intentionally NOT touched here (see services/client.ts disconnectClient);
+    // it stays at T1 until the next register rewrites it.
+    await clientService.disconnectClient(app.db, clientId);
+
+    // Capture the modal-open stamp. Match production logic — the dialog at
+    // packages/web/src/pages/clients/new-connection-dialog.tsx applies a 1s
+    // fudge to absorb browser↔server clock skew; we apply the same here so
+    // the boundary assertion mirrors what the predicate actually evaluates.
+    const openedAt = Date.now() - FUDGE_MS;
+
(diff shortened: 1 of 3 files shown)


baixiaohang (commented):
Solid fix. Root cause analysis checks out — I verified `client.id` is stable per-machine (`packages/shared/src/config/client-config.ts:17-25`) and that `onConflictDoUpdate` rewrites `connectedAt = now` on the reconnect path (`packages/server/src/services/client.ts:113-126`), so the timestamp-based detector handles both insert and re-pair correctly.

A few thoughts, none blocking:

**Nits (acknowledge, not asking to change here):**

- `openedAt = Date.now() - 1s` relies on browser↔server clock agreement. NTP-synced machines are fine, but a browser clock more than 1s *ahead* of the server will let `openedAt` land past the server's `now()`, and the modal will never match. Inherent limitation of any timestamp-based scheme; not worth solving in this PR. If we ever care, the cleanest fix is having the server return its own clock on `createConnectToken` and anchoring against that.
- There's a small theoretical false-positive window: a machine that happens to connect naturally within ~1s of the modal opening (without the user running the command) would be picked up. Probability is tiny, and the test at `OPENED_AT - 60_000` rules out the normal "long-connected machine" case, which is what matters in practice.

**One actual ask:**

The PR description's manual checklist still has two unchecked items, and they're the two cases this fix actually targets:

- [ ] Reconnect of a previously-paired-but-currently-disconnected machine — *the regression case*
- [ ] Already-connected machine sitting in the list, command NOT run — *guards against false success*

Could you run both before we land? The brand-new-machine path is what the old code happened to work on, so I'd really like a green tick next to the two new paths before approving.

Otherwise the implementation is clean — extracting `selectArrivedClient` as a pure function matches the `selectDisconnectedComputers` convention, the deps list update on the open-effect is right, polling cleanup is intact, and the unit-test coverage (boundary `=`, null `connectedAt`, cross-user, disconnected-with-recent-stamp) is exactly what I'd want as the regression net for this class of bug.

yuezengwu:
Thanks for the careful read.

I've pushed a server-side regression net (060008ad — `packages/server/src/__tests__/me-clients-reconnect.test.ts`) that pins the two cases you flagged at the contract level:

**Case 1 — Reconnect of a previously-paired-but-currently-disconnected machine**

> `re-registering an existing client_id stamps connectedAt to NOW and the wire reflects it`

Drives the actual server flow: `registerClient` (INSERT) → `disconnectClient` → capture `openedAt = Date.now() - 1s` (mirrors the dialog's fudge) → `registerClient` again (hits `ON CONFLICT DO UPDATE`) → GET `/me/clients` and assert `connectedAt > connectedAt_initial` AND `connectedAt >= openedAt`. That last comparison is byte-for-byte the predicate `selectArrivedClient` runs in production.

**Case 2 — Already-connected machine sitting in the list, command NOT run**

> `a long-standing connection's historical connectedAt stays BELOW a later modal-open stamp (no false success)`

Backdates `connectedAt` to one hour ago directly via PG, then asserts `connectedAt < openedAt` so the predicate would return `null` → modal stays in Waiting. I went with a PG patch rather than a `setTimeout` because the 1s skew fudge in the dialog would make a wall-clock-based gap impractical to assert without making the test slow.

Both run as part of the existing in-process server suite — `pnpm --filter @first-tree/server test` is green at 1110/1110.

---

**On the manual UI smoke** — I should be upfront: I can run the headless server stack and assert HTTP contracts, but my session doesn't have a real-browser surface with a credentialed login wired up for this dev env, so I can't drive the actual `/clients` modal end-to-end myself. The contract tests above + the `selectArrivedClient` unit cases together cover the same logical surface (server delivers a fresh `connectedAt` on re-pair AND a stale one on a long-running connection; predicate exists and returns the right thing for both), but I want to be clear that's an analytical equivalence, not a literal "I clicked the button and the green bar appeared."

If you'd like the literal button-click on the green bar, I'll hand it off to @yuezengwu to run through the two cases — happy to leave the PR pending on that signal. Otherwise, let me know if the contract-level pin is enough for you to land.

baixiaohang (approved):
LGTM. The new server contract test is exactly the right shape — driving `registerClient` → `disconnectClient` → `registerClient` through the real `ON CONFLICT DO UPDATE` path, then asserting `connectedAt2 >= openedAt` over the wire, is byte-for-byte the production predicate. Combined with the `selectArrivedClient` unit cases on the detector itself, the two manual checks I asked about are pinned at a level that won't silently regress.

Mirroring `FUDGE_MS = 1_000` on the test side and the 25ms gap for PG `NOW()` resolution are nice touches. The PG-backdate for the historical case (instead of a real-time sleep that the fudge would defeat) is the right tradeoff and the inline rationale makes that obvious.

One micro-nit for a future PR, not blocking: the fudge constant is now duplicated between `new-connection-dialog.tsx` and `me-clients-reconnect.test.ts`. If we ever tune it, both sites need to move together. A shared export from `shared/` would prevent silent drift, but it's tiny and not worth holding this on.

Approving. Thanks for the careful follow-up.

PR #512, opened 2026-05-22
feat(server): scheduled chat auto-archive sweeper replaces PR-merge bypass

yuezengwu:
## Summary

- Replace the per-event PR-merge archive bypass with a single periodic
  sweeper that handles both GitHub-mapped and unmapped chats uniformly.
- **Route A** (chats with GitHub mappings): archive every mapped human's view
  once every bound PR/Issue is closed/merged AND the chat is silent past
  `archiveMappedIdleSeconds` (default 1h).
- **Route B** (chats with no mapping): archive (chat, user) pairs whose user
  has no unread mention AND the chat is silent past
  `archiveUnmappedIdleSeconds` (default 12h).
- GitHub webhook now persists upstream PR/Issue lifecycle on
  `github_entity_chat_mappings.entity_state` so the sweeper can decide
  offline; the immediate-archive bypass and `archiveChatsForMergedPr`
  service are removed.

## Why

The previous design only archived on PR merge — issues, multi-PR chats, and
plain conversations with no GitHub binding never archived. Moving the
decision into a periodic sweep gives one consistent rule per chat type and
keeps everything behind the existing `BackgroundTasks` infrastructure (no
new framework, no extra process).

## Notable design points

- Both sweep queries are a single `INSERT … SELECT … ON CONFLICT` round-trip
  per route — already-archived rows are filtered out in SELECT so each tick
  is cheap once steady-state.
- `setWhere = engagement_status = 'active'` plus `deleted` sticky semantics
  match the legacy `archiveChatsForMergedPr` per-user safety guarantees.
- Auto-revive (`chat-projection.applyAfterFanOut`) is unchanged: a new
  message in an archived chat still flips engagement back to `active`.
- New runtime config: `archiveSweepIntervalSeconds` (default 300, set 0 to
  disable), `archiveMappedIdleSeconds` (3600), `archiveUnmappedIdleSeconds`
  (43200). All three accept env-var overrides.
- Migration 0048 adds `entity_state` (default `'open'`) + composite index.
  Historical mappings stay `'open'`, so the sweeper will not retro-archive
  legacy chats — an optional backfill against the GitHub API is left as a
  follow-up.

## Test plan

- [x] `pnpm check` (biome) clean
- [x] `pnpm typecheck` clean across all 11 packages
- [x] Server full suite: 1117/1117 passing
- [x] `chat-archive.test.ts` (13 cases): Route A/B happy paths, deleted-sticky,
      idempotency, unread guard, non-human skip, custom idle threshold,
      `entity_state='closed'` (non-merged) archives, sub-chat (`parent_chat_id`)
      is left alone
- [x] `github-app-webhook-route.test.ts` updates: PR merged/closed/reopened
      and `issues.closed` all sync `entity_state` without flipping engagement
- [ ] Local server boot + manual sweep (will do after merge during deploy
      verification)

## Follow-ups (intentionally not in this PR)

- Backfill script for historical `entity_state` via GitHub API
- Operator playbook entry: "chat should have archived but didn't — check
  for stale `entity_state`"


yuezengwu:
Reopening with correct branch name per CLAUDE.md (feat/xxx)

baixiaohang (commented):
Overall: solid design and a clear improvement on the per-merge bypass. Replacing per-event archive with `entity_state + idle window + periodic sweep` collapses three previously inconsistent surfaces (PR merge / Issue close / no-mapping chats) into one rule, with cheap steady-state writes and idempotency that survives multiple instances. Tests cover the core matrix well. A few non-blocking notes below.

### ✅ Pass

- **Per-user safety preserved.** `INSERT … ON CONFLICT DO UPDATE WHERE engagement_status = 'active'` is the SQL-native equivalent of the old `setWhere` guard in `archiveChatsForMergedPr`, and `deleted` stays sticky end-to-end (verified by the dedicated test).
- **Auto-revive contract intact.** `chat-projection.applyAfterFanOut` still flips `archived → active` on every new message in the same fan-out transaction, so a sweeper-archived chat that gets a stale reply is brought back without special-casing.
- **Sweeper math.** Route A uses `BOOL_AND(entity_state IN ('closed','merged'))` per chat — handles multi-PR / mixed PR+Issue chats correctly. Route B uses `LEFT JOIN github_entity_chat_mappings` + `m.chat_id IS NULL` for clean disjointness.
- **`parent_chat_id IS NULL` defensive filter** matches `listMeChats`’s historical-row defence.
- **Lifecycle handling** in `background-tasks.ts` is correct: timer guarded by `archiveSweepSeconds > 0`, paired `clearInterval` in `stop`, errors logged but never thrown.
- **Webhook update path** is best-effort (no thrown error blocks normalize/deliver), sits before `claimEvent` so a retry whose normalized event was already claimed still updates `entity_state`.

### ⚠️ Non-blocking suggestions

1. **Route B default 12h + implicit-zero `unread_mention_count` widens the behaviour beyond the legacy archive surface.** The previous bypass only touched chats explicitly bound to a merged PR. Route B now archives any (chat, human) pair where `COALESCE(cus.unread_mention_count, 0) = 0` after 12h of silence — including users who have **never** materialised a `chat_user_state` row (i.e. who never opened the chat, never markRead, never got mentioned). On a Friday-night → Monday-morning window that user can return to find the chat already in the Archived tab. Two cheap mitigations to consider:
   - Restrict Route B to users with `cus.last_read_at IS NOT NULL` so we only archive views the user has actually engaged with at least once.
   - Or extend the default to 7d to leave more headroom for the weekend / OOO case.
   Either is a single-line change; the test surface stays unchanged.
2. **Webhook delivery ordering is not protected.** GitHub does not guarantee inter-delivery ordering, so a `reopened` arriving before its preceding `closed` (or a duplicate-retry race) can leave `entity_state = 'closed'` for a now-open PR, which would let the sweeper archive a chat whose work is still in flight. Recording the GitHub-side `pull_request.updated_at` / `issue.updated_at` and only applying state updates with a strictly-greater timestamp would close this. Tracked as a follow-up rather than a blocker since the auto-revive on the next message acts as a safety net.
3. **`entity_state` backfill follow-up is listed but not scheduled.** Historical mappings stay `'open'`, so chats whose terminal PRs predate the migration will never archive via Route A. The current behaviour is "Route B picks them up after 12h", which is OK if Route B's threshold is acceptable (see suggestion 1). Worth filing an issue link in the PR body so the follow-up doesn't drop.
4. **Test plan item "Local server boot + manual sweep" is unchecked.** Per our repo convention this is part of the merge gate when a new background timer is introduced. A 1-min `archiveSweepIntervalSeconds=10` smoke run against the dev DB would be enough.

### ❓ Question

5. **Relationship to [proposals/hub-archive-tree-write.20260512.md](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/hub-archive-tree-write.20260512.md) §1 trigger set.** That proposal defines the `first-tree-write` task trigger as the **owning agent's manager user** transitioning the chat via `POST /chats/:id/engagement`. The sweeper writes `chat_user_state` directly, bypassing that API surface. Semantically, however, "all bound entities settled + idle 1h" is exactly the end-of-conversation signal the proposal wants to catch. Could you state explicitly in the PR description (or commit message) whether sweeper-driven archive is intended to be **in** or **out** of the future `tree_write_tasks` trigger set? Without that being pinned down now, the proposal author will hit ambiguity when wiring up the enqueue. Suggest excluding by analogy with LRU eviction (system-driven, not user-driven) — but the call should be documented.
6. **Context Tree update is missing.** Per CLAUDE.md's "after every task" rule, this PR changes a durable decision (when a chat flips to archived from the system side, plus three new runtime config keys with operator-visible defaults). The current `agent-hub/messaging.md` says only "archived auto-revives to active on a new message" — it does not describe **how a chat becomes archived in the first place** beyond the explicit user transition. Consider opening a companion tree PR adding a short section (either inside `messaging.md` or as a new `chat-archive.md` node under `agent-hub/`) covering: Route A / Route B triggers, idle defaults, the three env vars, and the answer to question 5.

PR #510, opened 2026-05-22
merge: bring first-tree (dev @032569d) into monorepo + assemble v1.0.0

baixiaohang:
# Summary

Repo merge **Step 2** — bring `agent-team-foundation/first-tree` (legacy/dev `032569d`) into this repo and assemble the unified `first-tree` CLI for **v1.0.0**. Phase 1A (hub restructure) landed in #502; Phase 1B (first-tree CLI rename / hub-stub removal) landed in legacy first-tree PR #436. This PR is Phase 2 + 3 from the [design doc](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/first-tree-repo-merge.20260521.md).

## Anchor SHAs (Release Day reference)

| Item | SHA |
| --- | --- |
| legacy/dev source | `032569d9dbe5467d060ee9ac2c762dd92bdb409c` |
| T2.1 merge commit | `267c5d951cac4d8e3b454444d43a61e208525672` |
| T2.3 biome reformat (anchored in `.git-blame-ignore-revs`) | `a08c6688bd2f9d99e544feb13a97f2b85779e504` |

# Commit sequence (19 commits)

## Phase 2 + 3 main line (11 commits)

| Commit | Task | What |
| --- | --- | --- |
| `267c5d95` | T2.1 | `git merge --allow-unrelated-histories legacy/dev@032569d`; 14 conflicts resolved keep-both |
| `36fd446d` | T2.2 | `git mv` placeholders → real first-tree dispatchers; docs → `docs/tree/` |
| `00eb9172` | T2.3a | root config consolidate; drop `.oxlintrc.json` / `.husky/` / `.lintstagedrc` |
| `a08c6688` | T2.3b | `pnpm biome check --write .` — anchor for `.git-blame-ignore-revs` |
| `fd80aa67` | T2.3c | manual + `--unsafe` biome lint residuals |
| `edaf7318` | T2.4 | drop legacy `publish-*.yml`; add `agent-e2e.yml`; add `grep-legacy-names` job |
| `4a021c67` | T2.5 | drop `_legacy_first_tree/` + 4 `*.legacy` backups |
| `5a4c1943` | T3.1 | wire `registerTree/GithubCommands` wrappers; delete dead `commands/index.ts` |
| `fa6d9a72` | T3.2 | **rename CLI to `first-tree`** — sweeping rename across 110 files; add `@first-tree/github-scan` workspace dep; `version` stays `0.14.9` per D-A=(b) |
[...]

PR #514, opened 2026-05-22
perf(web): drop redundant polling, route through admin WS push

yuezengwu:
## Summary

- **WS-covered queries lose their `refetchInterval`**. `useAdminWs` already invalidates `["chat-messages"]`, `["chat-detail"]`, `["me","chats"]`, `["activity"]`, `["sessions"]` on the matching `session:state` / `session:event` / `chat:message` frames. The 5s/10s/15s `refetchInterval` everywhere else was historical belt-and-suspenders that fanned out into one HTTP per query per tick (worst case: N agent rows × 6 RPM in the sidebar).
- **`useAdminWs` gains 3 throttled prefix invalidators** — `["session"]`, `["chat-right-sidebar", "session"]`, `["agent-sessions"]` — plus a targeted `["session-events", agentId, chatId]` invalidate driven by the frame's own routing dimensions. The `chat-right-sidebar` prefix is intentionally scoped to `"session"` so the sibling `github-entities` query (60s, GitHub REST) doesn't refetch on every `session:state` burst. Reconnect catch-up extended to match.
- **High-frequency modal polling downscaled + visibility-gated**: `last-step-modal` 2s → 5s, `new-agent-dialog` two parallel 3s → 5s each, `new-connection-dialog` 3s → 5s, `onboarding/step2-body` 3s → 5s. All four also gain `document.hidden` / `visibilitychange` gating so a backgrounded tab burns 0 RPM until it comes back to the foreground.

## End-to-end verification

Provisioned a devuser via `dev-callback` (mirrors `packages/e2e/src/framework/setup-devuser.ts` steps 1+7), planted the access token in `localStorage`, opened the workspace, and drove real `session:state` / `session:event` / `chat:message` frames by direct `pg_notify` against the running server.

Observed in server request log:
* `GET /chats/{chatId}/messages` — was polled every 5s; now fetched **once** at mount and refetched **only when `chat:message` fires**.
* `GET /agents/{id}/sessions/{chatId}` and `…/events` — was polled every 5–10s; now fetched once at mount and refetched **only when `session:state` / `session:event` fires** (verified by triggering `pg_notify('session_state_changes', '${agentId}:${chatId}:suspended:${orgId}')` and seeing the precise refetch land within the 1s throttle window).
* `chat-right-sidebar` GitHub section — verified untouched by the prefix narrowing (no extra GitHub round-trips per session burst).

## Test plan

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm --filter @first-tree/web test` (236/236)
- [x] End-to-end verification via dev-callback + `pg_notify`
- [ ] Reviewer: open the workspace with an agent ticking through tool calls; confirm WorkingChip + roster + right-sidebar state-dot still update in real time without the deleted polls
- [ ] Reviewer: open Last-step / New-agent dialog and switch to another tab; confirm `/clients`, `/me/agents/{id}` etc. stop ticking until you come back

## Notes

- No server changes, no schema changes, no migrations. PR is fully revertable.
- Out of scope (queued for a follow-up): the remaining 10s/15s long-tail polling (`/me/clients`, `/me`, team / clients page), a top-level `refetchIntervalInBackground:false` default on `QueryClient`, and pushing `agent:pinned` / `client:presence` onto the admin WS so the modal polls can go to zero.

### DIFF
--- a/packages/web/src/components/last-step-modal.tsx
+++ b/packages/web/src/components/last-step-modal.tsx
@@ -46,14 +46,22 @@ export function LastStepModal({ agent, open, onClose, onBound }: Props) {
     staleTime: 60_000,
   });
 
-  // Poll the agent row every 2s while the modal is open and the agent is
-  // still unbound. React-query stops fetching the moment the component
-  // unmounts.
+  // Poll the agent row every 5s while the modal is open and the agent is
+  // still unbound. Originally 2s, but that produces 30 RPM per open modal;
+  // the bind is gated on the user running a CLI command on another machine
+  // (single-digit-seconds task), so a 5s cadence still feels live without
+  // the request-storm. A proper push-driven path requires a new admin WS
+  // frame for `agent:pinned` (today it's only routed to the Client SDK).
+  // `refetchIntervalInBackground: false` skips ticks when the tab is
+  // backgrounded — the modal can't make progress anyway without a foreground
+  // CLI invocation, and the inevitable `refetchOnWindowFocus` will catch up
+  // when the user comes back.
   const agentQuery = useQuery({
     queryKey: ["agent-poll", agent.uuid],
     queryFn: () => getAgent(agent.uuid),
     enabled: open && !agent.clientId,
-    refetchInterval: 2000,
+    refetchInterval: 5_000,
+    refetchIntervalInBackground: false,
   });
 
   useEffect(() => {
--- a/packages/web/src/components/new-agent-dialog.tsx
+++ b/packages/web/src/components/new-agent-dialog.tsx
@@ -257,13 +257,16 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
   }, [name, open]);
 
   // Poll `listClients` to keep the connected-computer list fresh while the
-  // dialog is open. 3s cadence matches onboarding step 2 so a computer
-  // coming online (or going offline) reflects without the user closing
-  // and reopening the dialog.
+  // dialog is open. 5s cadence so a computer coming online (or going
+  // offline) reflects without the user closing and reopening the dialog.
+  // Skip ticks when the tab is hidden — nothing on this surface progresses
+  // without a foreground CLI run, and `refetchOnWindowFocus`-equivalent
+  // resume happens via the `visibilitychange` handler below.
   useEffect(() => {
     if (!open) return;
     let cancelled = false;
     const tick = async (): Promise<void> => {
+      if (document.hidden) return;
       try {
         const list = await listClients();
         if (cancelled) return;
@@ -280,10 +283,15 @@ export function NewAgentDialog({ open, onOpenChange, onCreated }: Props) {
       }
     };
     void tick();
-    const handle = window.setInterval(tick, 3_000);
+    const handle = window.setInterval(tick, 5_000);
+    const onVisible = (): void => {
+      if (!document.hidden) void tick();
+    };
+    document.addEventListener("visibilitychange", onVisible);
     return () => {
       cancelled = true;
       window.clearInterval(handle);
+      document.removeEventListener("visibilitychange", onVisible);
     };
   }, [open]);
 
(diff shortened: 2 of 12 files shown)


baixiaohang (approved):
LGTM. Cross-checked the key claims against the codebase:

- `notifier.ts` confirms `session:state` / `session:event` payloads carry `agentId`+`chatId`+`organizationId`, and `api/orgs/ws.ts` spreads them straight into the frame — so the precise `["session-events", agentId, chatId]` invalidate is well-grounded.
- The `["chat-right-sidebar","session"]` prefix narrowing does NOT touch `["chat-right-sidebar","github-entities", chatId]` (60s GitHub REST) — the deliberate scope-down is correct.
- Reconnect catch-up covers every key whose `refetchInterval` was removed, including `["chat-messages"]` and `["chat-detail"]`. No gap found.
- Visibility guard implementation is race-free (tick-time `document.hidden` check + `visibilitychange` immediate catch-up).
- Comments do an excellent job explaining *why* each number / prefix boundary was picked (esp. last-step's 5s reasoning, the `chat-right-sidebar` prefix scope-down, and the `agent:pinned` admin-WS frame as the reason it can't go push-only yet).
- Pure web-side change, no server/schema, easy to revert. The `pg_notify` + server-request-log evidence in the PR description is solid end-to-end verification.

Two non-blocking suggestions for a future follow-up — none of these should hold up the merge:

1. **Fan-out scope on `["chat-right-sidebar","session"]`**: a single `session:state` frame currently invalidates the sessionQuery for every non-human ParticipantAvatar + AgentRow in the same chat, even though the frame is about exactly one agent. The frame already carries `agentId`, so this could be tightened to a precise `["chat-right-sidebar","session", agentId, chatId]` invalidate the same way `session:event` was. Throttled 1s window keeps this bounded today, but worth noting.
2. **`setInterval` not paused while hidden**: the four modal polls return early inside `tick` when `document.hidden`, but the interval itself keeps firing every 5s. Negligible cost, but `clearInterval` on `visibilitychange` → hidden / re-create on visible would be the cleaner pattern.

Nice cleanup overall.

yuezengwu:
Addressed both follow-up suggestions from @baixiaohang in `c18e7c9`, and rebased the branch onto latest `main` (#510 + #515) so the PR is up to date.

**1. Precise `(agentId, chatId)` invalidate on `session:state`.** Replaced the three prefix-throttled invalidators with `invalidateSessionPair(qc, agentId, chatId)`:
   - Invalidates exactly `["session", agentId, chatId]` + `["chat-right-sidebar", "session", agentId, chatId]` + `["agent-sessions", agentId]`.
   - Throttles per `(agentId, chatId)` pair via a `Map<string, {lastAt, trailingTimer}>` so a burst from one agent can't starve another nor leak invalidations onto siblings.
   - Reconnect catch-up keeps using broad prefix invalidates (correct — we don't know which pairs missed frames during a WS gap).

**2. True visibility pause for the 4 modal polls.** Extracted `runVisibilityAwareInterval(tick, intervalMs)` in `packages/web/src/lib/visibility-interval.ts`. It `clearInterval`s on `visibilitychange → hidden` and re-creates the interval (plus fires an immediate catch-up tick) on `→ visible`. Applied to `new-agent-dialog` (2 sites), `new-connection-dialog`, `onboarding/step2-body`. Net code reduction in those files since the pattern is now centralized.

Checks: `pnpm check && pnpm typecheck && pnpm --filter @first-tree/web test` all green (236/236).

baixiaohang (approved):
Re-reviewed `c18e7c93` — both follow-up suggestions cleanly addressed and the rebase onto current main is conflict-free. LGTM, no further changes requested.

- The new per-(agentId, chatId) leading+trailing throttle in `invalidateSessionPair` is the right shape — bursts on one pair can't starve another, the trailing timer correctly re-resolves `latestQc` so a WS reconnect mid-window stays consistent, and `disposeSessionPairThrottle` cleans up trailing timers on teardown.
- Reconnect catch-up deliberately keeps the broad `["session"]` / `["chat-right-sidebar","session"]` / `["agent-sessions"]` prefix invalidates — correct, since a WS gap can drop frames for any (agent, chat) pair and we have to refresh broadly.
- `runVisibilityAwareInterval` is a clean extraction: real `clearInterval` on hidden, re-create + immediate catch-up tick on visible, single shared dispose. Four sites converted with a net LOC reduction. Last-step modal correctly left on React Query's `refetchIntervalInBackground: false` since that's the framework-native path.

Approving.

Issue #509, opened 2026-05-22
processed_events has no retention/TTL — unbounded growth degrades webhook INSERT performance


yuezengwu:
## Background

`packages/server/src/db/schema/processed-events.ts` defines the dedup ledger used by both GitHub App webhooks and the Feishu adapter. Every delivery's `delivery_id` is inserted permanently.

There is no:
- Retention / TTL field
- Periodic cleanup job
- Partial index for "still relevant" rows

## Why this matters

At a realistic load (thousands of deliveries/day across GitHub events, Feishu inbound, PR webhooks, CI webhooks via the App), the table grows by ~1M rows/year. After a few years that's tens of millions of rows. Every `INSERT ... ON CONFLICT DO NOTHING` performs a btree lookup against an ever-growing unique index, and autovacuum bills compound.

Same shape as a textbook SRE anti-pattern: dedup ledger without retention.

## What the retention window should be

GitHub's webhook retry policy redelivers for up to ~8 hours. Feishu's is similar. **7-day TTL is overkill-safe;** anything under 24 hours risks accepting a duplicate during an incident with delayed retries.

## Suggested fix outline

**Option A (minimal)** — verify `created_at` index exists, then daily cron `DELETE FROM processed_events WHERE created_at < now() - interval '30 days'`. Document retention rationale.

**Option B (cleaner)** — partition by week via pg_partman or native partitioning, drop old partitions wholesale. Heavier setup, zero-cost cleanup.

Either way: emit metrics for current row count + oldest row age so a regression in cleanup is visible.

## Severity

Major (operational) — not user-visible today, but a guaranteed long-term degradation. Cheap to fix early.

yuezengwu:
Closing — deferring. The unbounded-growth concern is real but speculative at current scale, and the proposed fix (#519) is the most invasive of the audit cluster (schema migration + new background task). Reopen if/when operational signals appear (e.g. degraded webhook INSERT latency, autovacuum churn). PR #519 closed in parallel; branch `fix/processed-events-ttl` preserved.

Issue #508, opened 2026-05-22
GitHub agent_created binding writes mapping under chat.org without validating reporter.org match


yuezengwu:
## Background

When an agent's tool_call mentions a GitHub PR/issue URL, `maybeBindGithubEntityFromToolCall` calls `resolveBindingPair` (`packages/server/src/services/github-entity-chat.ts:288-328`) to insert a row into `github_entity_chat_mappings`. The mapping's `organizationId` is taken from `representative.chatOrganizationId` — the **chat's** org.

**There is no assertion that the reporting agent's `organizationId` equals the chat's `organizationId`.**

## Why this matters

`createChat` already blocks cross-org participants under normal v2 flow, so under happy-path `reporter.org === chat.org` is enforced upstream. But:

- Pre-v2 / migration-era data may have mixed-org chats grandfathered in.
- Admin override paths (`addParticipant` with elevated JWT) can introduce a cross-org member without the createChat barrier.
- A future product change (e.g. inter-org collaboration chats) would silently break the invariant.

When the mismatch occurs, the mapping row goes under the **chat's** org. The GitHub installation firing future webhooks is bound to the **reporter's** org. Webhook audience resolution scopes by `organizationId` → zero matching mappings → silent drop.

This is defense-in-depth gap, not a hot bug.

## Reproduction (constructed)

1. Manually insert a `chat_membership` row for an agent from org A into a chat owned by org B (bypassing `createChat`).
2. As that agent, do a tool_call referencing a PR in a repo bound to org A's GitHub App installation.
3. Mapping row is written with `organization_id = orgB`, `delegate_agent_id = orgA-agent`.
4. Future webhooks from that PR (org A installation) look for mappings in org A → find none → audience empty → event dropped.

## Suggested fix outline

In `resolveBindingPair`, after resolving `chatOrganizationId` and the reporter, add:
```ts
if (reporter.organizationId !== chatOrganizationId) {
  log.warn({ reporterOrg, chatOrg, agentId, chatId }, "skipping agent_created binding: reporter/chat org mismatch");
  return null;
}
```

Plus a unit test that constructs the mismatch and asserts no mapping row is created.

## Severity

Major edge — triggered only under specific data / admin-override conditions, but the failure mode is silent.

PR #504, opened 2026-05-22
fix(server): wake delegate on GitHub events in multi-speaker chats

yuezengwu:
## Summary

- **Bug**: When a GitHub-bound chat has ≥ 3 speakers (e.g. a teammate was invited into the auto-created PR/issue chat), incoming GitHub webhook events land in the chat history but **never wake the delegate agent**. The card-format message produces no `mentionSet`, no `addressedToAgentIds` is passed, and `isOneOnOne` is false — so multi-speaker fan-out collapses every row to `notify=false`.
- **Why the 2-speaker default chat was unaffected**: the 1:1 implicit-wake rule (`participants.length === 2`) accidentally bailed out the default chat shape. The bug only manifests once another speaker joins the bound chat.
- **Fix**: `github-delivery.ts` now passes `addressedToAgentIds: [target.delegateAgentId]` when calling `sendMessage`. This is the same `SendMessageOptions` channel that the `question_answer` path already uses for system-routed messages whose recipient is fixed at write time.
- **Non-target speakers stay silent** (history-only) — preserves the mention-only fan-out semantics for everyone else in the chat.

## Related

Same class of bug, previously fixed for the `askuser` path:
- #404 — group-chat answer not waking asker
- #417 — follow-up: unread badge projection (still open)

The architectural recipe (`addressedToAgentIds` for system-routed structural targets) is already documented in `agent-hub/messaging.md` and on `SendMessageOptions.addressedToAgentIds`; this PR extends it to the third canonical system-routed channel (GitHub event delivery).

## Test plan

- [x] New unit test in `github-delivery.test.ts`: builds a 3-speaker bound chat (human + delegate + third party), delivers a normalized PR event via an `existing` audience target, asserts that the delegate's `inbox_entries.notify = true` and the third party's `notify = false`. Verified red without the fix, green with the fix.
- [x] All existing tests in `github-delivery.test.ts` still pass (2-speaker chat behaviour unchanged — `isOneOnOne` continues to wake the delegate without needing the new addressed path).
- [x] `pnpm test` on `@first-tree/server` — 1109 tests across 140 files, all green.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm check` (biome) — clean.
- [ ] Reviewer to confirm the canonical comment in `github-delivery.ts` matches the team's style.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #516, opened 2026-05-22
fix(server): make GitHub webhook delivery failures observable (#507)

### LINKED ISSUE #507: GitHub webhook delivery silently drops failures: per-target catch + claimed dedup has no metric or retry
## Background

`deliverNormalizedEvent` (`packages/server/src/services/github-delivery.ts:95-105`) wraps each audience target's `sendMessage` in a `try/catch` that calls `log.error(...)` and `continue`s. No metric, no alert, no retry hook.

Meanwhile, `packages/server/src/api/webhooks/github-app.ts:240` has already called `processed_events.claim(deliveryId)` before reaching delivery, so GitHub's own webhook retry skips this delivery as a duplicate.

## Why this matters

A single agent that consistently fails to receive GitHub cards (inbox wedged, partial fan-out failure, etc.) becomes structurally invisible: no metric to dashboard, no alert to page, no retry to self-heal. The log line is read by nobody.

Same risk-shape that #504 surfaced from a different angle: silent failure of GitHub → agent wake-up. #504 fixed the *routing* leg; this issue is the *observability* leg.

## Reproduction

1. Bind a chat to a GitHub entity.
2. Force `sendMessage` to throw for one specific delegate (e.g. corrupt `agents.inbox_id`).
3. Send a webhook for that entity.
4. Observe: stats.delivered drops to 0 for that target, GitHub never retries (claim held), no quantitative signal anywhere.

## Suggested fix outline

**Minimum viable** — counter metric in the catch block:
```ts
log.error({ metric: "github_delivery_failed_total", agentId, errorClass: err.name, entityKey }, ...)
```

**Stretch** — distinguish at the audience-empty branch (`github-app.ts:249`) between "no involves, no subscription" (expected) and "had involves but resolved to nobody" (`audience_empty_with_involves`), so triage has a quantitative signal.

**Stretch+** — per-target failure queue (small table, worker-retried) so a transient PG hiccup doesn't strand the event.

## Severity

Major — observability gap on a primary user-visible path. The failure mode is "silent forever", so when it bites it bites hard.

## References

- #504 — fix(server): wake delegate on GitHub events in multi-speaker chats (same surface, different leg)

yuezengwu:
Closes #507.

## Summary

Three coordinated changes that turn silent webhook-delivery failures into observable signals:

1. **`DeliveryStats.failed`** — count per-target exceptions. The catch was already isolating failures so one bad target doesn't poison the audience; now it also returns a `failed` counter so a regression in single-target reliability shows up in the webhook response and downstream metric pipeline.

2. **Structured metric line in the per-target catch**:
   ```
   metric=github_delivery_failed_total
   errorClass=<err.name>
   eventType, action, entityKey, entityType, humanAgent, delegateAgent
   ```
   Dashboardable via the existing JSON logger — no new infra.

3. **Webhook audience-empty branch distinguishes two reasons**:
   - `audience_empty_no_involves` — no involves, no subscription. Expected.
   - `audience_empty_with_involves` — had explicit involves but resolved to zero agents. Usually means a mentioned GitHub login has no `delegateMention`-configured agent in this org — actionable mis-config signal.
   Both the log line and the HTTP response now carry `reason`.

## Why this matters

Without these signals, the failure mode of "GitHub event reaches Hub, audience resolves, single delegate's send throws" is **silent forever**: GitHub won't retry (claim held), nobody mans the log, no metric exists. Sister bug to #504 (same surface, different leg — #504 fixed the routing, this fixes the observability).

## Test plan

- [x] `github-delivery.test.ts` — new assertion: `stats.failed === 1` in the broken-target case. Three existing tests updated to include `failed: 0` in stats equality.
- [x] `github-app-webhook-route.test.ts` — two new tests asserting `reason: audience_empty_no_involves` / `reason: audience_empty_with_involves` on the HTTP response.
- [x] `@first-tree/server typecheck` clean.
- [x] All 187 github-scope server tests pass.
- [ ] Note: repo-wide `pnpm typecheck` fails in `packages/github-scan` (missing react/ink/yaml type declarations), but those errors are pre-existing from #510 and not touched here.

## Related

- #507 — this issue
- #504 — fix(server): wake delegate on GitHub events in multi-speaker chats (sister bug)
- #508, #509 — siblings from the same audit, separate PRs


baixiaohang (approved):
LGTM. Three changes precisely cover #507's minimum + stretch goals; stretch+ (per-target retry queue) deliberately left as a follow-up — scope discipline is good.

Notable strengths:
- `event.involves.length > 0` discriminator matches the same check in `github-audience.ts:149`, so the two `reason` values stay semantically consistent with the audience resolver.
- `github_delivery_failed_total` follows the established `<name>_total` log-metric convention (cf. `double_encoded_content_unwrapped_total`, `loop_pattern_observed_total` in `services/message.ts`).
- `DeliveryStats.failed` is non-optional, forcing every construction site to handle it — the 3 existing test assertions updated to include `failed: 0` reflect that discipline.
- New tests cover both `audience_empty_*` branches via the HTTP response, plus the `stats.failed === 1` closure on the broken-target path.

Non-blocking nits (do not act):
- The `reason` string-literal union is inlined in the route handler. Fine at 2 values; consider hoisting to a named type in `@first-tree/shared` if a third reason ever shows up.
- The `creation_event_no_mapping_no_mention` drop path (`github-delivery.ts:60`) intentionally does NOT increment `failed` — that's correct (skip ≠ fail). If you later want to distinguish drop volume, a separate `stats.skipped` would be the right shape.

PR #517, opened 2026-05-22
fix(server): refuse agent_created GitHub binding on cross-org reporter (#508)

### LINKED ISSUE #508: GitHub agent_created binding writes mapping under chat.org without validating reporter.org match

yuezengwu:
Closes #508.

## Summary

`resolveBindingPair` in `github-entity-chat.ts` took the chat's `organizationId` as the mapping row's org without checking that the reporting agent's own `organizationId` matched. Under normal v2 flow `createChat` blocks cross-org participants, so this defense was implicit. But:

- Grandfathered cross-org `chat_membership` rows (pre-v2 data) can still exist.
- Admin-override paths (`addParticipant` via elevated JWT) can inject a cross-org member.
- A future inter-org collaboration feature would silently break the invariant.

When the mismatch happens, the mapping ends up under the **chat's** org while the GitHub App installation firing future webhooks lives in the **reporter's** org. Webhook audience resolution scopes by org → no matching mapping → silent event drop.

This PR makes the defense explicit: if `reporter.agentOrganizationId !== reporter.chatOrganizationId`, log a structured warn and return null instead of writing a mis-orged mapping row.

## Test plan

- [x] New test in `agent-created-binding.test.ts` seeds a second `organizations` row directly (since `createTestAdmin` always resolves to the default org), inserts the reporter into the second-org chat via a direct `chat_membership` insert that bypasses createChat's same-org guard, and asserts that no mapping row was written.
- [x] All 15 binding tests + 204 github-scope tests pass.
- [x] `@first-tree/server typecheck` clean.
- [x] The guard fires the expected warn log line:
  ```
  agent_binding skipped: reporter/chat organization mismatch
  ```

## Related

- #508 — this issue
- M1 sibling: #516 (delivery observability)
- M3 sibling: #509 (processed_events TTL — next PR)


baixiaohang (approved):
LGTM. Defense-in-depth 收口干净，逻辑居中、测试贴合 #508 描述的 admin-override / grandfathered 跨 org 数据场景，日志结构化字段齐全。

非阻塞的小建议（不影响合入，作者后续判断即可）：
- 测试已经断言 mapping 不写入，足够覆盖核心契约；如果想再加一层保险可以补一个 log spy 断言 `agent_binding skipped: reporter/chat organization mismatch` 实际触发。
- 这条路径是 silent-drop，靠 `log.warn` 发现真实触发依赖告警 / 仪表盘；如果 M1 #516 的 delivery observability 已经覆盖类似 silent-drop 指标就完全 OK，否则可以将来补一个 counter。
- 守卫后 `representative.chatOrganizationId` 与 `reporter.agentOrganizationId` 等价，可读性上后者更贴近"webhook 路由依据的 org"语义，纯口味问题。

PR #526, opened 2026-05-22
fix(cli): forward shell proxy env into launchd plist / systemd unit

liuchao-001:
## Summary

`first-tree daemon start` runs under launchd / systemd, which strip the user's interactive-shell env. The unit template only set `PATH` and `FIRST_TREE_SERVICE_MODE`, so a `git fetch` inside the daemon-spawned agent runtime had no way to reach a local proxy (Surge, Clash, corporate proxy …) — every `prepareGitWorktrees` call was a direct connection.

When direct egress to github.com is blocked (common in mainland China, some corporate networks), every fetch failed with `SSL_ERROR_SYSCALL` and bubbled out as `Session start/resume failed (...)` in chat. The shell-level `git fetch` in the *same* terminal worked fine because the user had `export https_proxy=...` in `.zshrc` — but launchd never sees that env.

This PR forwards `{http,https,all,no}_proxy` (lower + upper case) from the install-time shell into the plist / systemd unit. After install, the daemon and every git subprocess it spawns see the same proxy reachability the user's shell git already had.

## Implementation

- `collectProxyEnv(env = process.env)` — new exported helper, returns the non-empty subset of `PROXY_ENV_KEYS`.
- `renderPlist` and `renderSystemdUnit` now take an optional `proxyEnv` parameter (defaults to `collectProxyEnv()`) and append the entries inside the existing env block, after `FIRST_TREE_HOME`.
- Both render functions are now exported so unit tests can verify template output without spawning launchctl/systemctl.

## Migration note for existing users

**Auto-update alone does not migrate the env.** Existing users' currently-running daemon was started from a plist with no proxy env → supervisor's `process.env` has no proxy → `refresh-unit` (spawned by the supervisor on auto-update) reads that empty env → re-renders a plist still without proxy → no drift, no rewrite. This is the correct behaviour given the "no drift, no clobber" loop-stability invariant (see below).

**To pick up this fix on an existing install**, run one of the following **from a shell with `*_proxy` exported**:

```bash
first-tree login <token>          # or
first-tree daemon install         # or
first-tree daemon refresh-unit    # (hidden; works too)
```

Any of these re-renders the plist / unit from the calling shell's env, which now has the proxy values.

## Why this is loop-stable on auto-update

`refresh-unit` (called by the supervisor on every alpha bump) re-renders the template from the supervisor's `process.env`. The supervisor inherits its env from the launchd plist it was launched with. So the chain is:

1. User installs from shell with `https_proxy=X` → plist has `https_proxy=X`
2. Daemon launches via that plist → supervisor's env has `https_proxy=X`
3. Supervisor auto-updates → spawns `refresh-unit` with its env (including `X`) → new render has `https_proxy=X` → drift check sees no change → no clobber

A user who later wants to change proxy: `export https_proxy=Y; first-tree daemon refresh-unit` from their shell.

## Security note

If a proxy URL embeds credentials (`http://user:pass@host:port`), those credentials land in the rendered plist / systemd unit on disk (mode `0o644` for plist, default file perms for systemd user units — both group- and world-readable). Exposure level matches `.zshrc` if the user originally `export`ed the URL there, but operators relying on credential-bearing proxy URLs should be aware. The same note is in the `collectProxyEnv` docblock so future readers see it next to the code.

## Test plan

- [x] `pnpm check` — biome clean on changed files (16 pre-existing warnings in `packages/github-scan/` untouched)
- [x] `pnpm typecheck` — 13/13 tasks
- [x] `pnpm test` — 6/6 tasks (1123 server + 215 cli + others)
- [x] New unit tests cover `collectProxyEnv` (empty / lowercase / uppercase / mixed / empty-string-value / unrelated-keys) and `renderPlist` / `renderSystemdUnit` proxy emission, ordering, XML escaping, shell quoting (incl. credential-bearing `user:pass@host` URLs)
- [x] Smoke-tested locally on macOS with Surge: install with `https_proxy=http://127.0.0.1:6152` exported → plist gains the env block → `git fetch` from daemon-context succeeds via Surge HTTP proxy

## Background

User-facing investigation in chat `8b6d642e-f7de-467f-b6da-8f4401947fe9`. Stopgap workaround was `git config --global http.https://github.com/.proxy …`. This PR is the root fix.

### DIFF
--- a/apps/cli/src/__tests__/proxy-env-passthrough.test.ts
+++ b/apps/cli/src/__tests__/proxy-env-passthrough.test.ts
@@ -128,4 +128,15 @@ describe("renderSystemdUnit proxy passthrough", () => {
     });
     expect(unit).toContain('Environment=https_proxy="http://user pass@proxy:8080"');
   });
+
+  it("shell-quotes credential-bearing proxy URLs (the `@` host separator is outside the safe-char set)", () => {
+    // Common proxy auth form `http://user:pass@host:port`. The `@` between
+    // userinfo and host is not in shellQuote's safe-char regex, so the value
+    // must be wrapped in double quotes — otherwise systemd's tokeniser would
+    // split on `@` and read `pass` + `host:port` as two separate values.
+    const unit = renderSystemdUnit(inv, {
+      https_proxy: "http://user:pass@proxy.example.com:6152",
+    });
+    expect(unit).toContain('Environment=https_proxy="http://user:pass@proxy.example.com:6152"');
+  });
 });
--- a/apps/cli/src/__tests__/proxy-env-passthrough.test.ts
+++ b/apps/cli/src/__tests__/proxy-env-passthrough.test.ts
@@ -1,0 +1,131 @@
+import { describe, expect, it } from "vitest";
+import { collectProxyEnv, renderPlist, renderSystemdUnit } from "../core/service-install.js";
+
+describe("collectProxyEnv", () => {
+  it("returns empty when no proxy env vars are set", () => {
+    expect(collectProxyEnv({})).toEqual({});
+  });
+
+  it("picks up lowercase keys", () => {
+    expect(
+      collectProxyEnv({
+        https_proxy: "http://127.0.0.1:6152",
+        no_proxy: "localhost,127.0.0.1",
+      }),
+    ).toEqual({
+      https_proxy: "http://127.0.0.1:6152",
+      no_proxy: "localhost,127.0.0.1",
+    });
+  });
+
+  it("picks up uppercase keys", () => {
+    expect(collectProxyEnv({ HTTPS_PROXY: "http://127.0.0.1:6152" })).toEqual({
+      HTTPS_PROXY: "http://127.0.0.1:6152",
+    });
+  });
+
+  it("keeps lowercase and uppercase entries side by side (libcurl/JVM prefer different cases — don't merge or guess)", () => {
+    expect(
+      collectProxyEnv({
+        https_proxy: "http://lower:6152",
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM — 范围明确的 bug fix。

**核心改动**
- 新增 `collectProxyEnv()` 从 `process.env` 摘取 `{http,https,all,no}_proxy`（lower + upper case），由 `renderPlist` / `renderSystemdUnit` 注入 env 块。
- 三个函数 export 用于单测，模板拼接复用既有的 `escapeXml` / `shellQuote`。

**关键校验**
- Auto-update loop-stability 成立：`isServiceUnitDriftDetected` → `renderPlist(invocation)` 默认 `collectProxyEnv(process.env)`，supervisor 的 env 来自当初启动它的 plist，所以重渲染产物等同 → 不漂移、不 clobber。
- `escapeXml` 覆盖 `&<>"`，`shellQuote` 对逗号/空格自动加双引号，测试已明确断言 `Environment=no_proxy="localhost,127.0.0.1"` 和 URL query `&` 的转义。
- 空串过滤避免 `export https_proxy=` 污染 unit，单测有 case。

**Note（非阻塞）**
含账号密码的代理 URL 会落到 `~/Library/LaunchAgents/*.plist`（mode 0o644，组/其他可读），跟 `.zshrc` 风险等价、非本 PR 引入，建议在 release note 里提一句"代理凭证若有，会随 unit 文件落盘"。

老用户自动升级路径下，supervisor 自身的 env 没有代理变量，所以重渲染产物不含代理段，需要在 shell 里手动跑一次 `first-tree daemon refresh-unit` 才能拿到本修复 — PR 描述已说明，符合预期。

PR #534, opened 2026-05-25
ci: add migration linearity guard

yuezengwu:
## Summary

- Adds `scripts/check-migrations.mjs`, run as a new `Migrations` CI job, to enforce that drizzle migrations remain a strictly linear, contiguous sequence on `main`.
- Closes the gap that opened when `main` branch protection's "require branches to be up to date" was disabled: two PRs that each `drizzle-kit generate` against the same base would otherwise both squash-merge with colliding migration idxs (filenames differ by random suffix, so git would not flag a conflict).

## Scope: detective, not preventive

This check is a **tripwire**, not a true preventer. Two PRs that pass CI against an older base can still both squash-merge cleanly; the duplicate idx then surfaces on the *post-merge* main CI run (where the local contiguity check trips on the broken `_journal.json`). That's enough to stop a bad state from being silently shipped to production, but a contributor has to revert and renumber to fix it.

For true prevention, follow-ups would be: re-enable "require branches to be up to date", enable GitHub's merge queue, or auto-rerun PR CI on base updates. Picking one of those is out of scope for this PR.

## What the check enforces

1. `packages/server/drizzle/meta/_journal.json` is well-formed: `idx` starts at 0, strictly increments by 1, no duplicate tags, every tag's `NNNN_` prefix matches its `idx`.
2. The on-disk `.sql` files and the journal agree — no orphan `.sql`, no missing `.sql` for any journal entry.
3. **On PR runs**: new journal entries must have `idx > max(main idx)`, and entries that exist on `main` may not be removed or renumbered. On push-to-main runs the cross-branch comparison is skipped (the local checks above already catch the racing-merge case, and avoiding the `origin/main` fetch removes a TOCTOU window where another push could land between this push and the fetch).

On collision, the failure message tells the contributor exactly how to recover (rebase onto main, delete the local migration, re-run `pnpm --filter @first-tree/server db:generate`).

## Follow-up after merge

Add `Migrations` to the required status checks on `main` branch protection so the guard is mandatory, not advisory.

## Test plan

- [x] Clean checkout: passes (`Migration check OK`)
- [x] PR that legitimately appends `0049_*` while main's max is `0048`: passes
- [x] PR's `0049_*` collides with a different `0049_*` already on main: fails with rebase guidance
- [x] Journal with a missing idx (gap): fails
- [x] Orphan `.sql` not in journal: fails
- [x] Journal entry without matching `.sql`: fails
- [x] Tag prefix mismatches idx (e.g. `idx=49 tag=0050_*`): fails
- [x] \`pnpm check\` (biome) clean on the new script

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,15 +159,15 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          # Need main reachable so check-migrations.mjs can compare HEAD's
-          # _journal.json against main's. The job only ever runs `git show`
-          # against origin/main, so a depth-1 fetch of main below is enough —
-          # no need to pull full history.
           fetch-depth: 1
       - name: Fetch main for cross-branch comparison
-        # On a push to main, origin/main and HEAD coincide; the script
-        # detects this and skips the cross-branch check. On PRs we need
-        # an explicit fetch because checkout@v4 only fetches the PR ref.
+        # Only PR runs need to compare HEAD against main. On push-to-main
+        # the local contiguity check is sufficient (a duplicate idx that
+        # snuck through two racing PRs surfaces as a broken sequence in
+        # _journal.json) and fetching main risks a TOCTOU "missing from
+        # HEAD" false positive if another push lands between this push
+        # and the fetch.
+        if: github.event_name == 'pull_request'
         run: git fetch --depth=1 origin main:refs/remotes/origin/main
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,30 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
+  check-migrations:
+    name: Migrations
+    runs-on: ubuntu-latest
+    # Always runs — even on PRs that touch no server code. Branch protection
+    # no longer requires PRs to be up to date with main, so this job is the
+    # single guard against two PRs landing colliding migration idxs.
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Need main reachable so check-migrations.mjs can compare HEAD's
+          # _journal.json against main's. The job only ever runs `git show`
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
LGTM — script is single-purpose, no-deps, defensively coded (execFileSync, idx=0 / empty-journal / missing-base / HEAD==base all handled), and error messages carry recovery guidance. Reproduced all 7 test-plan scenarios locally on a worktree — happy path passes, every failure mode trips the right message.

Two non-blocking notes for follow-up:

1. The guard is a **tripwire**, not a true preventer — two PRs that each pass CI against an older base can still both squash-merge; the duplicate idx then surfaces on the post-merge main CI. To actually *prevent* collisions, consider re-enabling "require branches to be up to date", enabling GitHub's merge queue, or auto-rerunning PR CI on base updates. Worth being explicit about the detective-vs-preventive nature in the description.
2. Push-to-main race: when main is rapidly mutated, `origin/main` may have moved past HEAD by the time the job runs, producing "missing from HEAD" failures. Using `${{ github.event.before }}` as the base on push events (or skipping cross-branch entirely on push) would tighten this. Minor.

Don't forget the follow-up: add `Migrations` to required status checks on `main` branch protection so this is mandatory, not advisory.

yuezengwu:
Thanks for the review @baixiaohang.

1. **Tripwire vs. preventer** — agreed, updated the PR description to call this out explicitly under "Scope: detective, not preventive". Reverting to a true preventer (require-up-to-date, merge queue, or auto-rerun on base updates) is a separate decision; happy to follow up on whichever route the team prefers.
2. **Push-to-main race** — addressed in 184f17ea: cross-branch comparison now runs only on `pull_request` events. On push to main, only the local contiguity check runs — that is sufficient because a duplicate idx that snuck through two racing PRs will show up as a broken sequence in `_journal.json` (the contiguity check trips on the second `idx=N`). This also removes the TOCTOU window you flagged.
3. **Follow-up** — noted, will flip on `Migrations` as a required status check on `main` once this merges.

PR #530, opened 2026-05-22
fix(client): route legacy sessions to their per-chat cwd; fresh-start as fallback (R2 hotfix for #506)

yuezengwu:
## Summary

**Hotfix for production bug surfaced after [#506](https://github.com/agent-team-foundation/first-tree/pull/506) merged.**

User report after upgrade:
\`\`\`
ERROR · SDK · 05/22 22:41
No conversation found with session ID: 72a19485-ca9e-4bc3-9add-a57e8314e5c3
\`\`\`

This is **R2** from [proposal §⓪.3](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/agent-session-cwd-redesign.20260519.md). PR #506 changed the runtime cwd from \`<workspaces>/<agent>/<chatId>/\` to \`<workspaces>/<agent>/\`. The Claude Agent SDK stores per-project session transcripts at:

\`\`\`
~/.claude/projects/<encoded-cwd>/<sessionId>.jsonl
\`\`\`

where \`<encoded-cwd>\` replaces every non-alphanumeric char in the absolute cwd with \`-\`. The legacy chatId-suffixed encoding no longer matches the new per-agent encoding, so calling \`query({ resume: <legacy-sessionId> })\` throws \`No conversation found ...\` asynchronously inside the SDK consume loop — surfacing as an SDK error event in the chat timeline.

## Fix (revised per yuezengwu's directive)

**Earlier draft of this PR**: detect missing transcript, fall through to fresh start with a new sessionId.

**Revised approach (this PR)**: backward-compat routing. Legacy sessions resume against their **original per-chat cwd** so the SDK transcript is found in place and agent memory is preserved across upgrade. Fresh-start fallback only when the transcript is missing from BOTH legacy AND new cwd.

\`resume()\` probe priority:

1. **Legacy hit** (\`<workspaceRoot>/<chatId>/<sessionId>.jsonl\` exists under \`~/.claude/projects/\`): use the legacy chat-dir as cwd verbatim. Skip \`ensureAgentBootstrap\` / \`prepareSourceRepos\` / \`markWorkspaceInitComplete\` — the legacy dir already carries the v1.x bootstrap output and we MUST NOT overwrite it. Return the legacy sessionId unchanged.
2. **New-design hit** (\`<workspaceRoot>/<sessionId>.jsonl\`): standard PR #506 path. cwd = agent home, full bootstrap, source repos materialised.
[...]

PR #506, opened 2026-05-22
refactor(client): per-agent-home cwd + source repos at top-level

yuezengwu:
## Summary

Implements [proposals/agent-session-cwd-redesign.20260519.md](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/agent-session-cwd-redesign.20260519.md) plus the 2026-05-22 worktree-layout redesign on top.

**Old model**: per-chat cwd at `<DATA_DIR>/workspaces/<agent>/<chatId>/`. Each chat its own checkout. 155 directories / 20 GB on this dev box for one agent.

**New model**:
- cwd is **per-agent**: `<DATA_DIR>/workspaces/<agent>/`, shared by every chat session of the agent
- Predeclared `gitRepos` materialise as **source repos at the top level** of the agent home (`<agentHome>/<localPath>/`), NOT under `worktrees/`
- `worktrees/` subdir is **reserved entirely** for the agent's on-demand worktrees (created via `git worktree add` per task — runtime never pre-creates)
- Per-chat content (Current Chat Context, source repo list, worktree convention) injected via Claude SDK `systemPrompt.append` for claude-code; via AGENTS.md for codex (codex has no equivalent SDK option — accepted race window per proposal §⓪.3)
- Bootstrap is sentinel-gated + identity-hash-checked + Context-Tree HEAD-drift-aware
- Per-process per-path mutex (`runtime/worktree-mutex.ts`) prevents two concurrent sessions racing on the same checkout
- Handler `shutdown()` no longer `rm`s cwd or shared source repos (per-agent persistence by design)

This is intentionally aggressive — see proposal §⓪ Global Evaluation for the strategic trade-off.

## Files changed (+1485 / -418, 13 files)

- `packages/client/src/runtime/workspace.ts` — `acquireAgentHome`, sentinel helpers, `cleanWorkspaces` → no-op
- `packages/client/src/runtime/bootstrap.ts` — split stable vs per-chat; new `buildChatSystemPrompt`; `readCachedContextTreeHead` / `readContextTreeHead` / `writeContextTreeHead`; `PredeclaredSourceRepo` type (renamed from `PredeclaredWorktree`); shared `isHubWorktreeMarker` + `deepEqualIdentity` (extracted from handlers)
- `packages/client/src/runtime/worktree-mutex.ts` (new) — per-process per-path mutex
[...]

PR #535, opened 2026-05-25
fix(client): resolve doc-preview base path against per-agent-home for new chats

Gandy2025:
## Problem

The chat **doc-preview** feature only renders a `.md` mention as a clickable
link when the runtime snapshotted that file at send time (the "snapshot
gate"). `buildMessageDocumentSnapshots` `realpath`s its base dir first and
returns **zero snapshots** if that fails — so when the base dir doesn't
exist, every `.md` mention in the message silently stays plain text.

After #506 (per-agent-home cwd + source repo at top level), a **new chat**
no longer gets a `<workspaceRoot>/<chatId>/` directory; its cwd is the
per-agent home and the repo lives at `<workspaceRoot>/<localPath>`. But the
doc-base resolver in `session-manager.ts` still computed
`join(workspaceRoot, chatId)`, pointing at a directory that does not exist.

**Result:** doc preview was dead for every chat created after #506 — the
symptom reported in dev (md docs render as plain text, not previewable
links). #506 didn't touch `session-manager.ts`, so the resolver was never
migrated; #530 fixed the analogous cwd issue for session *resume* but not
the doc-base path.

## Fix

Add `resolveSessionDocRoot(workspaceRoot, chatId)`, mirroring #530's
claude-code `resume()` discriminator: use the legacy per-chat dir when it
**physically exists**, else the per-agent home. Both doc-base call sites
(the sync env path used by `first-tree chat send`, and the async
`resolveDocumentBasePath` used by result-sink) now route through it.

**No regression:** the prior code resolved legacy chats to the legacy dir
too, so only the new-chat (no-legacy-dir) path changes behaviour. This is a
strict improvement.

## Known limitation (follow-up, not in this PR)

`existsSync(legacyDir)` is a *proxy* for "the handler chose the legacy cwd".
It is exact for new chats but `SessionManager` is handler-agnostic, so two
**legacy-chat** cases still diverge (resolver returns the legacy dir while
the handler actually ran at the agent home):

1. **codex legacy chats** — the codex handler has no legacy-cwd branch
   (`handlers/codex.ts` always uses `acquireAgentHome`); its surviving
   pre-#506 `<chatId>/` dir trips the proxy.
2. a claude-code legacy chat whose SDK transcript was lost resumes cold at
   the agent home (#530 case 3) while its `<chatId>/` dir still exists.

Both are graceful (an older revision may be previewed, never an empty/wrong
file), bounded to pre-#506 chats (which shrink over time), and **not a
regression** versus today. The clean fix is to thread the handler's
resolved cwd through to the sink instead of re-probing here — tracked as a
follow-up.

## Test plan

- New `resolveSessionDocRoot` tests: new-chat regression (→ agent home),
  legacy-dir-on-disk (→ per-chat), single-repo end-to-end.
- Existing `document-base-path` + `doc-snapshots` suites updated/green.
- `pnpm --filter @first-tree/client typecheck` clean; biome clean on
  changed files.
- `first-tree-cli-contract.integration.test.ts` fails locally for an
  unrelated reason (on-PATH CLI can't find bundled `skills/` payloads); not
  caused by this change.

### DIFF
--- a/packages/client/src/__tests__/document-base-path.test.ts
+++ b/packages/client/src/__tests__/document-base-path.test.ts
@@ -1,6 +1,9 @@
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import type { AgentRuntimeConfigPayload } from "@first-tree/shared";
-import { describe, expect, it } from "vitest";
-import { documentBasePathFromRuntimeConfig } from "../runtime/session-manager.js";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { documentBasePathFromRuntimeConfig, resolveSessionDocRoot } from "../runtime/session-manager.js";
 
 function payload(gitRepos: AgentRuntimeConfigPayload["gitRepos"]): AgentRuntimeConfigPayload {
   return { kind: "claude-code", prompt: { append: "" }, model: "", mcpServers: [], env: [], gitRepos };
@@ -9,11 +12,11 @@ function payload(gitRepos: AgentRuntimeConfigPayload["gitRepos"]): AgentRuntimeC
 const PER_CHAT = "/data/workspaces/agent/chat-123";
 
 describe("documentBasePathFromRuntimeConfig", () => {
-  it("returns the per-chat workspace root when there are zero repos", () => {
+  it("returns the session doc root when there are zero repos", () => {
     expect(documentBasePathFromRuntimeConfig(payload([]), PER_CHAT)).toBe(PER_CHAT);
   });
 
-  it("returns the per-chat workspace root when there are multiple repos", () => {
+  it("returns the session doc root when there are multiple repos", () => {
     const result = documentBasePathFromRuntimeConfig(
       payload([{ url: "https://github.com/a/one.git" }, { url: "https://github.com/a/two.git" }]),
       PER_CHAT,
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM — 小而聚焦的 bug 修复，逻辑与测试都到位。

**目标**：修 #506 带来的回归 — 新 chat 的 doc-preview base path 仍指向不存在的 `<workspaceRoot>/<chatId>/`，导致 `safeRealpath` 返回 null，`buildMessageDocumentSnapshots` 静默返回零快照，所有 `.md` mention 退化为纯文本。

**好的地方**：
- `resolveSessionDocRoot` 复用 #530 在 claude-code `resume()` 用过的同一判定信号（`existsSync(legacyDir)`），一致性好。
- 纯读 `existsSync`，避免了 `acquireAgentHome/acquireWorkspace` 在每次出站消息上跑 mkdir 副作用 —— 注释也写明了。
- 同步路径（`chat send` env）与异步路径（`resolveDocumentBasePath`）两个 call site 都改了，不会再分叉。
- 旧代码在 legacy 场景下也指向 legacy 目录，本 PR 只改变 new-chat（无 legacy dir）路径 —— 严格 strict improvement，无回归。
- 测试新增覆盖新 chat / 磁盘上存在 legacy dir / 单 repo 端到端三种情况，全绿。

**Known limitation**（PR 描述已自述，非本 PR blocker）：`existsSync(legacyDir)` 是 "handler 选了 legacy cwd" 的代理，对 new chat 精确，但对两类 legacy 仍会偏差：
1. codex legacy chats（codex handler 无 legacy-cwd 分支，始终 `acquireAgentHome`）— pre-#506 残留的 `<chatId>/` 会触发偏差；
2. claude-code legacy chat 的 SDK transcript 丢失冷启时（#530 case 3）。

两种偏差都 graceful（最多读到 stale 副本而非空/错文件），且会随 legacy chat 自然衰减。干净修法是把 handler 决定的 cwd 透传给 sink，注释里已标注为 follow-up，节奏合理。

`pnpm --filter @first-tree/client typecheck` 干净，`document-base-path` + `doc-snapshots` 测试全绿。

🤖 Review by yzw-assistant

PR #536, opened 2026-05-25
fix(client): restore working state on inject + drop dead constants

baixiaohang:
## Summary

- Long-thinking turns triggered by a follow-up user message were being suspended at `idle_timeout` (300s) even though the agent was still working. Root cause: `setRuntimeState(\"idle\")` at every turn end + no path to flip back to `working` on `inject` → the working-grace guard added in #477 silently broke.
- Fix: `SessionManager.dispatch` now sets `runtimeState=working` right after `handler.inject` in the active branch. The grace window now actually covers the inject → first SDK event gap.
- Hardened `evictIdle`: hard upper bound (`idle_timeout + working_grace_seconds`) lifted into an explicit `pastHardCap` predicate so both the AskUserQuestion guard and the working/blocked guard honour it. Worst case still bounded.
- Documented the `lastActivity` / `runtimeState` invariants on top of `evictIdle` — any new wake-up path MUST set `working`.
- Deleted `packages/client/src/runtime/constants.ts` (the unused 8h / 50 / 16 / 60min defaults). All values were only read by the legacy `loadRuntimeConfig` yaml loader; the shipped CLI boots through `apps/cli/ClientRuntime` + `@first-tree/shared` `agentConfigSchema` (300s / 10 / 5 / 3600s). Removing the dual-default avoids the same divergence in the future.

## Background

Reproduced from production log of architect agent (chatId `8507c155-...`):

```
11:47:19 session resumed
11:48:07 lastActivity last touch
11:50:11 \"session working but no output, marking blocked\" (inactiveSec:124)
11:52:37 Result forwarded   ← lucky save, but next inject would have been unprotected
```

#477 added the `working`/`blocked` grace, but it only fires when `runtimeState ∈ {working, blocked}`. Handler can't observe `inject` from inside its SDK for-await, so the state stays at the post-result `idle`. The 8h client default in `constants.ts` was masking this bug previously; the 300s shared default unmasked it.

## Test plan

- [x] `pnpm --filter @first-tree/client typecheck`
- [x] `pnpm typecheck` (all 13 packages)
- [x] `pnpm check` (no new biome issues)
- [x] `session-manager` / `session-runtime-state` / `session-manager-question-answer` / `config` test suites — 43 tests pass
- [x] New regression test: turn ends → idle → follow-up dispatch → 20s under idle_timeout=1s/grace=60s should NOT suspend
- [ ] Manual: monitor architect chat on dev daemon for one long-thinking turn after merge

[...]

PR #537, opened 2026-05-25
fix(client): drop evictIdle 120s working→blocked auto-migration

baixiaohang:
## Summary

Follow-up to #536. After that PR shipped, the architect agent was still emitting yellow \`session working but no output, marking blocked\` warnings ~2min into long reasoning turns — even though the working-grace window correctly prevented any suspend. The 120s auto-migration was just guessing \"stuck\" based on SDK silence, which is wrong with reasoning models where 2-5min deep-thinks are normal.

- Remove the \`else if (inactiveMs > blockedThresholdMs)\` branch in \`evictIdle\`. \`runtimeState\` now stays at \`working\` until the handler sets it back to \`idle\` (per result message) or idle/grace suspends the session.
- Keep \`blocked\` as a semantic slot in \`RuntimeState\` — evictIdle's working-state grace still exempts it, so any future handler-driven \`blocked\` signal works without further runtime changes.
- Update the invariant docblock on \`evictIdle\` to make the new contract explicit (handler-driven only).

## Repro

```
13:56:20  Session resumed
13:58:34  inactiveSec:121, msg: \"session working but no output, marking blocked\"
          ← level=40 warn, runtimeState working → blocked, then aggregate → blocked, UI yellow
```

No actual suspend — the SDK kept producing output and the turn completed normally. The state migration was UX noise only.

## Test plan

- [x] \`pnpm --filter @first-tree/client typecheck\`
- [x] \`pnpm check\` (no new biome issues)
- [x] New regression: \`working\` session silent for 3 min does NOT emit \`blocked\` AND is not suspended (idle_timeout=1h to isolate the non-transition from suspend)
- [x] Existing \"exempts \`blocked\` from idle suspend\" test updated for the new wording (path is now handler-driven, not auto)
- [x] All session-{manager, runtime-state, state-notifications, manager-question-answer} suites pass (51/51)
- [ ] Manual: monitor architect chat for a long reasoning turn — UI should stay green throughout

### DIFF
--- a/packages/client/src/__tests__/session-runtime-state.test.ts
+++ b/packages/client/src/__tests__/session-runtime-state.test.ts
@@ -426,7 +426,12 @@ describe("SessionManager.evictIdle — working-state grace window", () => {
     }
   });
 
-  it("also exempts 'blocked' (the state evictIdle itself flips 'working' into after 2 minutes)", async () => {
+  // `blocked` is no longer auto-set by evictIdle (the 120s auto-migration
+  // was removed because reasoning-model thinking turns commonly exceed 2
+  // minutes between SDK events, producing false-positive UI warnings).
+  // The grace-window exemption still covers `blocked` so any handler that
+  // sets it explicitly — present or future — is treated like `working`.
+  it("also exempts 'blocked' from idle suspend inside the grace window", async () => {
     vi.useFakeTimers();
     try {
       let capturedCtx: SessionContext | undefined;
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM. 改动很干净：去掉 evictIdle 里基于 SDK 静默时间猜 "stuck" 的 120s 自动迁移，把 `blocked` 留作 handler-driven 语义槽。

- suspend 路径（hard cap、AskUserQuestion 豁免、working/blocked grace）行为完全保留，只是把嵌套 if 用 early-continue 展平
- 回归测试明确锁住了新契约：`working` 静默 3 分钟既不进 `blocked` 也不被 suspend
- 文档块 invariant 更新得到位，未来重新引入 `blocked` 信号时读者知道走的是 handler 路径

理由也站得住：reasoning 模型 2–5min 的安静期是正常深思，靠传输层静默来推断卡住本来就是错的信号源。

PR #538, opened 2026-05-25
ci(migrations): add LATEST sentinel to force conflicts on racing PRs

yuezengwu:
## Summary

Closes the residual gap left after PR #534. The check-migrations CI job catches collisions *after* main is corrupted — this PR makes the collision impossible to merge in the first place.

Mechanism: a single-line sentinel file `packages/server/drizzle/LATEST` holds the latest migration tag. Every migration PR rewrites this same line. Two PRs racing to add migrations against the same base produce a deterministic git **modify/modify** conflict on `LATEST` (instead of the unreliable add/add-adjacent behavior on `_journal.json`), which blocks the GitHub merge button until a human rebases.

## What changes

- **`packages/server/drizzle/LATEST`** (new) — one line, currently `0048_github_entity_state`. The merge-time anchor.
- **`scripts/sync-migration-head.mjs`** (new) — reads `_journal.json`, writes its last tag to `LATEST`. Zero deps.
- **`packages/server/package.json`** — `db:generate` now chains the sync script so contributors never have to remember `LATEST` exists.
- **`scripts/check-migrations.mjs`** — adds an assertion that `LATEST` equals the journal's last tag. A PR that forgets to sync fails CI with an actionable message pointing at the sync script.

## Why a separate file, not journal manipulation

Looked at drizzle's existing artifacts: the journal's `entries[]` append, the per-migration `meta/NNNN_snapshot.json` files, and `drizzle.config.ts` all either expose an add-at-end pattern or a different-file-per-migration pattern — none of them are a single line that every new migration must rewrite. Drizzle-kit also has no documented pre/post-hook to inject one. A separate sentinel file is the smallest, clearest intervention.

## Verification

Simulated two concurrent PRs both branching off the same base and each adding a 0049 migration with the sentinel sync wired in. After PR-A merges into the proxy main, attempting to merge PR-B produces:

\`\`\`
Auto-merging packages/server/drizzle/LATEST
CONFLICT (content): Merge conflict in packages/server/drizzle/LATEST
Auto-merging packages/server/drizzle/meta/_journal.json
CONFLICT (content): Merge conflict in packages/server/drizzle/meta/_journal.json
Automatic merge failed; fix conflicts and then commit the result.
\`\`\`

Conflict on `LATEST` is guaranteed by construction; the journal happens to also conflict in this simulation because of the differing \`when\` timestamps, but we don't rely on that — \`LATEST\` is the deterministic anchor.

## Test plan

- [x] Clean checkout: \`check-migrations\` passes
- [x] LATEST drifts ahead of journal: fails with sync hint
- [x] Journal advances but LATEST not synced: fails with sync hint
- [x] After running \`sync-migration-head.mjs\`: passes
- [x] Simulated two-PR race: PR-B merge produces \`CONFLICT (content)\` on LATEST
- [x] biome check clean on the new files

### DIFF
--- a/packages/server/drizzle/LATEST
+++ b/packages/server/drizzle/LATEST
@@ -1,0 +1,1 @@
+0048_github_entity_state
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -20,7 +20,7 @@
     "build": "tsdown src/index.ts src/app.ts src/observability/index.ts --format esm --no-dts",
     "typecheck": "tsc --noEmit",
     "test": "vitest run",
-    "db:generate": "drizzle-kit generate",
+    "db:generate": "drizzle-kit generate && node ../../scripts/sync-migration-head.mjs",
     "db:migrate": "drizzle-kit migrate",
     "db:studio": "drizzle-kit studio"
   },
--- a/scripts/check-migrations.mjs
+++ b/scripts/check-migrations.mjs
@@ -15,7 +15,15 @@
 //      every entry's `tag` has the matching `NNNN_` prefix).
 //   2. The on-disk .sql files and the journal agree (no orphan .sql, no
 //      missing .sql for a journal entry).
-//   3. Every new journal entry vs. origin/main has idx > max(main idx) —
+//   3. packages/server/drizzle/LATEST holds the journal's last tag. This
+//      sentinel is the merge-time anchor: because every migration PR
+//      rewrites the same single line, two PRs that race to add a migration
+//      against the same base produce a deterministic git modify/modify
+//      conflict at merge time (instead of the unreliable add/add-adjacent
+//      behavior that journal-append alone gives). Keep it in sync via
+//      `pnpm --filter @first-tree/server db:generate`, which now invokes
+//      scripts/sync-migration-head.mjs.
+//   4. Every new journal entry vs. origin/main has idx > max(main idx) —
 //      i.e. PRs may only append migrations, never reuse or rewrite a
 //      number already on main. When the comparison base isn't available
 //      (e.g. workflow_dispatch without a fetched main), the cross-branch
@@ -30,6 +38,8 @@ const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), "..");
 const drizzleDir = resolve(repoRoot, "packages/server/drizzle");
 const journalPath = resolve(drizzleDir, "meta/_journal.json");
 const journalRelPath = "packages/server/drizzle/meta/_journal.json";
+const latestPath = resolve(drizzleDir, "LATEST");
+const latestRelPath = "packages/server/drizzle/LATEST";
 
 const errors = [];
 const fail = (msg) => errors.push(msg);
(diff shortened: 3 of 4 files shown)


baixiaohang (commented):
Reviewed the diff against the PR #534 baseline. Approach is sound, scope is appropriately small, and the four pieces fit together cleanly. Mostly non-blocking notes below.

### ✅ What works

- **Conflict semantics are correct for squash-merge.** GitHub computes mergeability by simulating a merge of PR head into the latest target. Two PRs that both rewrite `LATEST`'s single line from `oldX` → different new values produce a deterministic modify/modify conflict regardless of merge strategy. The PR body's claim holds.
- **Closing the loop via `check-migrations`.** PR #534 only catches collisions *after* a bad state lands on main; this PR makes the bad state un-mergeable in the first place. The CI assertion in `validateLatestSentinel` (lines 100–123) ensures the sentinel can't silently drift — a PR that forgets to sync gets a CI failure with an actionable hint pointing at `scripts/sync-migration-head.mjs`.
- **Zero-dep sync script.** `scripts/sync-migration-head.mjs` is 27 lines of pure file I/O. Wiring it into `db:generate` is the right ergonomic choice — contributors never need to know `LATEST` exists by name.
- **Considered the alternatives.** PR body explicitly rules out journal manipulation and per-migration files. Agreed — `_journal.json`'s append pattern can sometimes auto-merge when entries are positioned right, and `meta/NNNN_snapshot.json` is per-migration so different idxs don't collide.

### ⚠️ Minor / non-blocking

- **Filename `LATEST` (no extension).** Slightly unusual; `.last-migration` or `MIGRATION_HEAD` would be more self-documenting. The in-file comment in `check-migrations.mjs` does the work though, and renaming is cheap if we want to later. Not a blocker.
- **Edge case: same-tag racing PRs.** If two PRs hypothetically produced identical `0049_<name>` tags (e.g. manual rename), `LATEST` would match on both sides and not conflict. In practice drizzle-kit's filename suffix randomization makes the tag unique, and `check-migrations` step 4 (cross-branch idx check) would catch it anyway — but worth noting that `LATEST` alone isn't sufficient under all adversarial paths; it's a complement to step 4, not a replacement.
- **`sync-migration-head.mjs` exits 1 on empty journal.** Correct behavior, but the call site in `db:generate` chains with `&&` after `drizzle-kit generate`, so an empty journal can only happen if drizzle silently produced no migration. Minor: log message could be clearer that this likely means `drizzle-kit generate` produced nothing.

### ❓ Question

- Plan to add `Migrations` (and by extension this new LATEST check) to required status checks on main branch protection? PR #534's body flagged that follow-up; would be good to confirm it lands together with this so the guard becomes mandatory rather than advisory.

Overall: clean, minimal, addresses a real residual gap from #534.

yuezengwu:
Thanks for the review @baixiaohang.

1. **Empty-journal log** — addressed in 56efc49d. Message now says "did `drizzle-kit generate` produce a migration?" so the failure mode is obvious in context.
2. **Filename `LATEST`** — kept as-is for now. The intent of the bare uppercase name was to read like `HEAD` / `CHANGELOG` / `LICENSE` — a top-level marker file, not a config file. The in-file CI message + `check-migrations.mjs` header comment explain its role. If the team prefers `.last-migration` or `MIGRATION_HEAD` I'm happy to rename in a follow-up — flagging here so we don't bikeshed in this PR.
3. **Same-tag adversarial path** — agreed, `LATEST` is a complement to step 4 (cross-branch idx check), not a replacement. drizzle-kit's tag suffix randomization makes the same-tag case astronomical, and step 4 catches it anyway because the new idx would be ≤ main's max after PR-A lands. Worth noting that the two checks together form the full guard: LATEST is the *git-level* deterministic conflict (blocks merge button), step 4 is the *CI-level* semantic check (catches post-rebase regenerate-please cases).
4. **Required status check** — already done. PR #534 added `Migrations` to first-tree main's classic branch protection required contexts (alongside `Lint & Type Check` and `Test`). The new `validateLatestSentinel` assertion runs inside the same `Migrations` job, so it inherits the required status automatically — no protection-config change needed.

baixiaohang:
Thanks for the thorough reply @yuezengwu.

1. ✅ Empty-journal message now reads clearly in context.
2. ✅ `LATEST` filename — the `HEAD` / `CHANGELOG` / `LICENSE` framing makes sense (a top-level marker, not config). Agreed not worth bikeshedding here; happy to leave as-is.
3. ✅ Nice framing: LATEST = git-level deterministic block, step 4 = CI-level semantic catch. Will reuse this mental model.
4. ✅ Good — didn't realize #534 already wired `Migrations` into required contexts. That means the new `validateLatestSentinel` is mandatory by inheritance. Cleared.

All my points addressed. Will defer the approve action to @baixiaohang.

baixiaohang (approved):
All review points addressed. LGTM.

PR #539, opened 2026-05-25
fix(server): respect pending questions and unread when auto-archiving chats

yuezengwu:
## Summary

- Route A (GitHub-mapped chats) previously archived purely on terminal entity state + idleness, ignoring per-user unread (`unread_mention_count > 0`) and chat-level pending ask-user questions.
- Route B (unmapped chats) already gated on unread, but missed pending questions.
- Added the missing guards to both routes, with a matching `ON CONFLICT … WHERE unread_mention_count = 0` clause to close the SELECT-to-UPSERT race.

## Why

Symptom we kept hitting: a chat with an outstanding `ask-user` question (or a chat the user explicitly marked unread) would flip to **Archived** as soon as the bound PR merged, but the red "needs you" badge / pending row stayed on the user's surface. The signals the user cared about (unread, open question) were ignored by the sweeper.

## What changed

- `services/chat-archive.ts`
  - **Route A** CTE adds a chat-level `NOT EXISTS pending_questions WHERE status = 'pending'`. The outer SELECT adds a per-user `COALESCE(cus.unread_mention_count, 0) = 0`.
  - **Route B** WHERE adds the same chat-level pending-question carve-out.
  - Both routes: tightened the `ON CONFLICT … WHERE` to also require `unread_mention_count = 0` so a concurrent unread bump can't sneak past the SELECT guard.

- The pending-question carve-out is intentionally **chat-scoped, not per-askee**: `pending_questions` has no askee column (askee is implicit in chat membership), so per-user dispatch buys nothing and the chat-level skip is cheaper. Resolved rows (`answered`/`superseded`) are ignored.

- Index coverage: `idx_pending_questions_chat_status (chat_id, status)` already exists, so the new `NOT EXISTS` clauses are anti-join lookups, not seq scans.

- No schema / migration changes.

## Tests

Added 4 cases to `__tests__/chat-archive.test.ts` (now 18 total, all green; full server suite 1149/1149 green; `pnpm check` + `pnpm typecheck` green):

- Route A: pending question blocks the entire chat from archiving
- Route A: answered / superseded historical rows do **not** block archiving
- Route A: same chat with two humans — h1 has unread → stays active, h2 fully read → archived
- Route B: pending question blocks the chat from archiving

## Test plan

- [x] `pnpm exec vitest run chat-archive` — 18/18 pass against real Postgres
- [x] Full server suite — 1149/1149 pass
- [x] `pnpm check` — zero warnings on changed files
- [x] `pnpm typecheck` — 13/13 tasks pass
- [ ] Staging soak: confirm a freshly-merged PR with an open ask-user does NOT flip the chat to Archived
- [ ] Staging soak: confirm a chat the user manually marked unread does NOT auto-archive even after idle threshold

### DIFF
--- a/packages/server/src/__tests__/chat-archive.test.ts
+++ b/packages/server/src/__tests__/chat-archive.test.ts
@@ -21,6 +21,7 @@ import { chatMembership } from "../db/schema/chat-membership.js";
 import { chatUserState } from "../db/schema/chat-user-state.js";
 import { chats } from "../db/schema/chats.js";
 import { githubEntityChatMappings } from "../db/schema/github-entity-chat-mappings.js";
+import { pendingQuestions } from "../db/schema/pending-questions.js";
 import { sweepChatArchive } from "../services/chat-archive.js";
 import { createTestAdmin, useTestApp } from "./helpers.js";
 
@@ -72,6 +73,21 @@ async function addHumanMember(app: App, chatId: string, agentId: string): Promis
     role: "member",
     accessMode: "speaker",
     source: "manual",
+  });
+}
+
+async function seedPendingQuestion(
+  app: App,
+  chatId: string,
+  agentId: string,
+  status: "pending" | "answered" | "superseded" = "pending",
+): Promise<void> {
+  await app.db.insert(pendingQuestions).values({
+    id: randomUUID(),
+    chatId,
+    agentId,
+    messageId: randomUUID(),
+    status,
   });
 }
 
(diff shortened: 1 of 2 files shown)


baixiaohang (commented):
Reviewed end-to-end (diff + service file + tests + schema + race analysis). The fix is well-scoped and well-tested; all my notes are non-blocking observations.

## ✅ Correctness

- **Both routes correctly grow the right guards.** Route A had no unread guard at all (legacy `archiveChatsForMergedPr` reach) and no pending-question guard; both added. Route B already gated on unread; the new pending-question guard slots into the same WHERE and is the only thing it was missing.
- **Race window closed both ways.** The added `chat_user_state.unread_mention_count = 0` predicate in *both* `ON CONFLICT ... WHERE` clauses correctly defends against a concurrent `markChatUnread` landing between SELECT and UPSERT. Verified by walking the four interleavings:
  1. Sweeper INSERT wins, then `markChatUnread` ON CONFLICT bumps unread → row ends `(archived, unread=1)`.
  2. `markChatUnread` INSERT wins (creates `unread=1, active`), then sweeper hits ON CONFLICT, the new `unread_mention_count = 0` guard is false → skipped. Correct.
  3. & 4. Symmetric for the explicit-active-row cases.
  Race #1 is a *pre-existing* concern, not introduced here — see "minor" note below.
- **Chat-scoped pending-question carve-out is the right call.** `pending_questions` has no askee column (only `agent_id` = the *sender*, the delegate that emitted the question — confirmed in `pending-questions.ts:21` docstring). Per-user dispatch would need to infer askee from message content (mentions); chat-level skip is both cheaper and strictly safer (over-protects rather than mis-targets).
- **Status filter is correct.** `markSupersededByChat` (services/questions.ts:306) transitions `pending → superseded` on session evict with `reason='chat_archived'`, and `answered` is set on user answer. The `status = 'pending'` predicate in the new `NOT EXISTS` matches exactly the rows that represent an open ask. Resolved rows (`answered`/`superseded`) correctly do not block — covered by the dedicated test.

## ✅ Index / Perf

- `idx_pending_questions_chat_status (chat_id, status)` exists in `pending-questions.ts:36` — the new `NOT EXISTS pq WHERE pq.chat_id = ... AND pq.status = 'pending'` is an index lookup, as the PR description claims.
- Route A's `NOT EXISTS` sits in the inner CTE WHERE so it filters pre-aggregation — same chat_id correlation across all per-mapping rows; PG runs the lookup per join row but each is O(log n) and bounded by batchSize (1000).

## ✅ Tests

- 18/18 pass locally is consistent with what I see — the four new cases each hit a different code path:
  - Route A pending-block ✓ (CTE-level filter)
  - Route A answered/superseded ignored ✓ (status predicate is tight)
  - Route A per-user unread + same-chat multi-mapping (`(h1, delegate, #14)` + `(h2, delegate, #14)` PK is legal — confirmed via `githubEntityChatMappings` PK on `[org, human, delegate, type, key]`) ✓ (outer per-user guard)
  - Route B pending-block ✓ (outer WHERE filter)
- Tests use real Postgres via `useTestApp`; no mocks.

## ⚠️ Minor (non-blocking, follow-ups for visibility)

1. **Residual hybrid-state race (pre-existing, not introduced).** If the sweeper INSERT wins the race against a concurrent `markChatUnread` (interleaving #1 above), the row ends `(engagement_status='archived', unread_mention_count=1)`. The user's intent ("put back in my inbox") is silently lost because `markChatUnread` (me-chat.ts:1058–1068) only touches `unread_mention_count` and never flips `engagement_status` back to `active`. Auto-revive only fires on new *messages*. Not in scope for this PR — the patch correctly shrinks the larger window — but worth a follow-up issue if telemetry shows the hybrid state in production. A simple fix would be `markChatUnread` also setting `engagement_status = 'active'` in its ON CONFLICT SET.
2. **Test helper readability nit.** `seedPendingQuestion(app, chatId, agentId, status)` — `agentId` here is the *sender* (delegate), consistent with the schema, but a reader new to this file might assume it's the askee. A trivial rename to `senderAgentId` (or a one-line comment) would prevent future confusion. Not worth blocking.
3. **Partial-index opportunity (future).** `idx_pending_questions_chat_status` is a full index. Once question volume grows, a partial `WHERE status = 'pending'` would tighten the anti-join further (most historical rows are `answered`/`superseded`). Defer until measured.

## Verdict

LGTM. The fix correctly addresses the user-visible symptom (chat flipping to Archived while pending question / manual-unread badge stays on the surface), the race-closing predicates are the right shape, and the test coverage matches the claim. Nothing blocking.

baixiaohang (approved):
LGTM — 守卫位置、race 收窄、测试覆盖都到位，approve。

三条改进建议，按"现在能顺手做"和"另开 PR 立刻做"分类：

### 本 PR 可顺手改

1. **测试 helper 参数名**。`seedPendingQuestion(app, chatId, agentId, status)` 的 `agentId` 是发问者（delegate / 自治 agent），不是 askee —— 与 `pending_questions.agent_id` 列语义一致，但读起来容易让人以为是"问谁"。建议改成 `senderAgentId`（或加一行注释 `// sender of the question, not the askee`）。trivial 改动，本 PR 可以一起进。

### 单独 PR 立刻跟进（别拖）

2. **`markChatUnread` 的混合态 bug**。本 PR 收窄了 SELECT→UPSERT race，但残留窗口仍会让行停在 `(engagement_status='archived', unread_mention_count=1)` —— 用户点"标记未读"想把聊天拉回收件箱，结果聊天还在 Archived tab，徽章却是未读。me-chat.ts:1058 的 `markChatUnread` ON CONFLICT 只 set `unreadMentionCount`，不动 `engagement_status`，是这个 race 的根因（不是这次新引入的，但本 PR 改完更值得收尾）。修复一行：ON CONFLICT 的 SET 里加 `engagementStatus: sql\`CASE WHEN ${chatUserState.engagementStatus} = 'archived' THEN 'active' ELSE ${chatUserState.engagementStatus} END\``（deleted 仍 sticky）。建议本 PR 合并后立刻开一个 follow-up PR。

3. **`pending_questions` 的 partial index**。`idx_pending_questions_chat_status` 现在是全表索引，但本 PR 的 `NOT EXISTS` 只查 `status='pending'`，而历史行绝大多数都是 `answered`/`superseded`。建议改成 `WHERE status='pending'` 的 partial index —— 索引体积直接缩到当前 pending 数量级（通常个位数到几十），sweeper 的 anti-join 也更快。一行 migration 的事，不需要等"观测数据"。

PR #544, opened 2026-05-25
fix(server): route GitHub events by human, not (human, delegate) tuple

yuezengwu:
## Summary

- **Bug**: After an agent created an issue via `gh issue create` inside a chat, assigning that issue on GitHub to the same human minted a second chat for the same entity.
- **Cause**: The mapping was written under `(human, reporterDelegate, entity)`, but the assign webhook's audience resolved `(human, delegateMention)` to a *different* delegate. The strict four-tuple lookup in `resolveTargetChat` missed, `isMentionMatched=true` skipped the creation-event guard, and step (c) minted a sibling chat.
- **Fix**: route by human first.
  - `resolveAudience`: dedup `kind:"new"` candidates by `humanAgentId` only — once any `(human, *, entity)` mapping exists, the involves path does not fork a parallel target with a different delegate.
  - `resolveTargetChat` step (a.5): after a direct miss, look up any chat already bound to `(org, human, entity)` regardless of delegate (open entities and earliest `bound_at` win), reuse it, and write a sibling mapping row so the next event hits step (a) directly.
- Schema and primary key `(org, human, delegate, entity)` unchanged — many-to-many between humans and delegates still works.

## Test plan

Regression coverage was authored against the *pre-fix* code first; all three failed precisely with the user-reported symptom, then pass after the fix is restored.

- [x] `resolve-target-chat.test.ts` — "reuses the existing chat when a different delegate hits the same (human, entity) tuple" (asserts shared `chatId`, single chat in DB, sibling mapping row written).
- [x] `github-audience.test.ts` — "dedups involves by human even when the existing mapping uses a different delegate" (asserts audience length 1, kind=existing, delegate from the recorded mapping).
- [x] `agent-created-webhook-e2e.test.ts` — full HMAC-signed webhook path: `gh issue create` writes the binding, then `issues.assigned` from a collaborator routes back to the same chat. `delivered=1`, `newChats=0`, no extra chats, no sibling mapping.
- [x] `pnpm --filter @first-tree/server test --run` — 1153 / 1153 pass.
- [x] `pnpm --filter @first-tree/server typecheck` and `pnpm check` — clean (0 new warnings).

### DIFF
--- a/packages/server/src/__tests__/agent-created-webhook-e2e.test.ts
+++ b/packages/server/src/__tests__/agent-created-webhook-e2e.test.ts
@@ -250,6 +250,115 @@ describe("Agent-created → real webhook end-to-end", () => {
     expect(sentToOrg).toHaveLength(0);
   });
 
+  it("issues.assigned does NOT mint a sibling chat when the assignee's delegateMention differs from the chat's agent_created delegate", async () => {
+    // Regression for the assignee-creates-new-chat bug. Real-world repro:
+    //   1. Inside a chat, the user asked their assistant agent to open an
+    //      issue via `gh issue create`. Mapping written under
+    //      (human, assistantDelegate, issue).
+    //   2. The human's `delegateMention` happens to point at a *different*
+    //      delegate (e.g. their main workhorse agent) — the assistant that
+    //      created the issue was a one-off helper.
+    //   3. A collaborator opens the issue on GitHub and assigns it to the
+    //      human. The assign webhook arrives with `involves=[human,
+    //      reason=assigned]` and sender=collaborator (external actor).
+    //
+    // Bug (pre-fix): audience built `kind:"new"` for (human, delegateMention)
+    // because the dedup key included delegate, then `resolveTargetChat` did
+    // not find a mapping under that delegate and minted a fresh chat.
+    //
+    // Fix: audience dedups by human alone; resolveTargetChat falls back to
+    // any chat already bound for (human, entity) regardless of delegate.
+    const app = getApp();
+    const admin = await createTestAdmin(app);
+    const installationId = 200005;
+    await seedInstallation(app, { installationId, orgId: admin.organizationId });
+
+    const assistantDelegate = await seedDelegateAgent(app, {
+      orgId: admin.organizationId,
+      memberId: admin.memberId,
+      name: `assistant-${randomUUID().slice(0, 6)}`,
+    });
+    const workhorseDelegate = await seedDelegateAgent(app, {
+      orgId: admin.organizationId,
+      memberId: admin.memberId,
+      name: `workhorse-${randomUUID().slice(0, 6)}`,
+    });
+    // The human's delegateMention points at the workhorse, NOT the assistant
+    // that created the issue. This is the misalignment the bug exposes.
+    const humanName = `assignee-${randomUUID().slice(0, 6)}`;
+    const human = await app.db
+      .insert(agents)
+      .values({
+        uuid: randomUUID(),
+        name: humanName,
+        organizationId: admin.organizationId,
+        type: "human",
+        displayName: humanName,
+        inboxId: `inbox_${randomUUID()}`,
+        managerId: admin.memberId,
+        delegateMention: workhorseDelegate,
+      })
+      .returning({ uuid: agents.uuid });
+    const humanUuid = human[0]?.uuid;
+    if (!humanUuid) throw new Error("seeded human agent missing uuid");
+
+    const chatId = await seedDirectChat(app, admin.organizationId, humanUuid, assistantDelegate);
+
+    // Stage 1: assistant agent opens the issue via gh issue create.
+    await maybeBindGithubEntityFromToolCall(app.db, assistantDelegate, chatId, {
+      toolUseId: "tu-assign-1",
+      name: "Bash",
+      args: { command: 'gh issue create --title "track Z" --body "ctx"' },
+      status: "ok",
+      resultPreview: "https://github.com/owner/repo/issues/905",
+    });
+    const initialMappings = await app.db
+      .select()
+      .from(githubEntityChatMappings)
+      .where(eq(githubEntityChatMappings.entityKey, "owner/repo#905"));
+    expect(initialMappings).toHaveLength(1);
+    expect(initialMappings[0]?.chatId).toBe(chatId);
+    expect(initialMappings[0]?.delegateAgentId).toBe(assistantDelegate);
+
+    const beforeChats = (await app.db.select().from(chats)).length;
+    const beforeMessages = (await app.db.select().from(messages).where(eq(messages.chatId, chatId))).length;
+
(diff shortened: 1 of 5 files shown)


baixiaohang (changes requested):
Solid diagnosis and the two-layer defence (audience dedup + resolver `(a.5)` fallback) maps cleanly onto the "fail-closed" pattern from B5 in the binding design. Tests are well-structured (regression-first against pre-fix code, three levels: unit / audience / e2e). Two concerns worth resolving before approve, plus one follow-up question.

### ⚠️ Concern 1 — `subscribed` path is not deduped by human, so sibling rows can fan out duplicate cards

Once `(a.5)` writes a sibling mapping row (e.g. `(humanA, workhorseDelegate, entity) → chatX` while `(humanA, assistantDelegate, entity) → chatX` already exists), the next webhook for the same entity hits `resolveAudience` and:

- `subscribedRows` returns **two** rows (same chat, different delegates).
- The new involves-dedup correctly skips humanA (good), but the subscribed list itself is not deduped — `audience` ends up with two `existing` targets both pointing at `chatX`.
- `deliverNormalizedEvent` loops the audience and calls `sendMessage` twice → **two identical cards posted into the same chat**.

Wake-up isn't broken (the second delegate isn't a chat member, so `addressedToAgentIds` no-ops for it in `fanout`), but the user sees duplicate cards, and `inbox_entries` get double-written. The trigger is rare (requires the very race `(a.5)` is meant to catch — subscribed empty at audience time, then `(human, *, entity)` present at resolver time), but once a sibling row is in place it's permanent state.

Suggestion: dedup `subscribed` by `humanAgentId` (keep earliest `bound_at` row), symmetric to the new involves dedup. The "many-to-many between humans and delegates" guarantee still holds — different *humans* still produce distinct rows; this only collapses multiple *delegates* under the same human, which by construction point at the same chat (sibling rows always reuse `humanScoped.chatId`).

### ⚠️ Concern 2 — sibling rows carry `bound_via = "direct"`, indistinguishable from first-touch rows

Per [github-entity-chat-binding.md](https://github.com/agent-team-foundation/first-tree-context/blob/main/agent-hub/github-entity-chat-binding.md) decision **B1**, `bound_via` is an open text column and adding values is a pure app-layer change. A new value like `"human_fallback"` (or `"sibling"`) on the `(a.5)` insertion would make audit logs distinguish "fix re-bound this row" from "this was the first chat created for this delegate". Zero migration cost, useful for telemetry on how often `(a.5)` actually fires in production (which informs whether the Concern 1 race is hypothetical or real).

### ❓ Question — should `(b)` (fixes-link reuse) also fall back by human?

Symmetric scenario, not introduced by this PR but exposed by the same root cause: a collaborator opens a PR "Fixes #N" where issue #N was bound under `(humanA, delegateA)` via agent_created. The PR mentions humanA, whose `delegateMention = delegateB`. Then in `resolveTargetChat`:
- (a) miss on `(humanA, delegateB, PR)`
- (a.5) miss on `(humanA, *, PR)` (the PR itself was never bound)
- (b) iterates `relatedEntities = [#N]`, looks up `(humanA, delegateB, #N)` → miss (issue is under delegateA)
- (c) mints a fresh chat — same fanout problem as the one this PR fixes, just routed via a `Fixes #N` link instead of an assign.

This isn't a regression from this PR, but it's the same `(human, delegate)`-tuple-too-strict pattern at the `(b)` lookup. Worth either extending `lookupMapping` inside the `(b)` loop to fall back by-human, or filing a follow-up. Probably out-of-scope here, but want confirmation that the gap is acknowledged so it doesn't sit unrecognised.

### ✅ Otherwise LGTM
- PK + schema unchanged, pure app-layer fix consistent with the "many-to-many delegates" model.
- Comments on `subscribedHumanIds` and `lookupMappingByHuman` are precise and explain *why*, not what.
- e2e test exercises the real HMAC-signed webhook path, not just the resolver — exactly the right level for regression coverage of an issue spanning audience + resolver.
- CI green: 1153/1153 tests + typecheck + lint clean.

yuezengwu:
Thanks for the thorough review, addressed in bd79d359:

### Concern 1 — subscribed dedup ✅ fixed
`resolveAudience` now dedups `subscribed` rows by `humanAgentId` (keeping the earliest `bound_at` row as the representative). Sibling rows always share `chatId` by construction — step (a.5) always inserts using `humanScoped.chatId` — so the collapse is loss-free. The "many-to-many between humans and delegates" guarantee still holds for distinct humans; only multiple delegates under the same human collapse. New regression in `github-audience.test.ts` (`collapses subscribed rows by human ...`) asserts two sibling rows → one audience target. This eliminates the duplicate-card path even if a race produces sibling rows in the wild.

### Concern 2 — bound_via for (a.5) rows ✅ fixed
Added a fourth `BoundVia` enum value `"human_fallback"`. The (a.5) write now uses it, while first-touch rows stay `"direct"`. Pure app-layer change (per binding-design B1 — `bound_via` is open text), no migration. `resolve-target-chat.test.ts` now asserts the sibling row carries `"human_fallback"` and the original stays `"direct"`, so future telemetry can quantify how often the involves→delegate mismatch fires in production.

### Question — (b) fixes_link path 👍 acknowledged, follow-up
Confirmed the gap: a `Fixes #N` PR routed to an issue bound under a different delegate would hit the same `(human, delegate)`-too-strict miss at the (b) lookup and mint a fresh chat. Out of scope for this PR — the fix is the same shape (make `lookupMapping` inside the (b) loop fall back by-human) but the trigger is much rarer (requires both a `Fixes` link AND an audience-delegate mismatch). Will file a follow-up issue so it doesn't sit unrecognised; happy to take it in the next PR.

All 1154 tests + typecheck + biome pass after the changes.

yuezengwu:
@baixiaohang gentle ping — the two concerns are addressed in bd79d359 and the follow-up gap is tracked in #547. CI is green (1154/1154 tests). Would you mind taking another pass when you have a moment? Happy to iterate further if anything else stands out.

baixiaohang (approved):
LGTM — both concerns resolved cleanly in bd79d359:

- **Concern 1 (subscribed dedup)**: `resolveAudience` now dedups by `humanAgentId` keeping the earliest `bound_at` row. Sibling rows share `chatId` by construction (a.5 always inserts using the human-scoped lookup's chatId), so the collapse is loss-free. New regression in `github-audience.test.ts` (`collapses subscribed rows by human ...`) pins the behaviour.
- **Concern 2 (bound_via tagging)**: New `"human_fallback"` enum value tags (a.5) sibling rows; first-touch stays `"direct"`. `asBoundVia` updated, JSDoc documents the telemetry intent, and `resolve-target-chat.test.ts` asserts both rows carry the right `bound_via`.
- **Follow-up ((b) fixes_link path)**: Tracked in #547 with scenario, fix shape, and acceptance criteria spelled out — happy with deferring it.

Pure app-layer fix, PK/schema unchanged, CI green (1154/1154). 👍

PR #402, opened 2026-05-15
feat(server,web): close private-agent identity & owner-exclusive gaps

Gandy2025:
## Summary

Closes two long-standing invariants that previously fell through the cracks (fixes #372).

### 1. Identity rendering inside a chat = f(Membership), not f(Discoverability)

Pre-fix, `chatDetail.participants` only carried `{agentId, role, mode, joinedAt}`; the web client resolved names via the visibility-filtered `/agents` endpoint, so a private agent in a chat surfaced as a UUID prefix to every other member who didn't manage it.

- `GET /chats/:id` and `chat.service::getChatDetail` JOIN `agents` directly (without applying `agentVisibilityCondition`)
- Web client prefers `chatDetail.participants[i].{name, displayName}` for `@`-autocomplete, participant chips, and message sender labels; the identity map is now a fallback only

### 2. Only the owner can pull a private agent into a chat

Service-layer owner-exclusive checks on **every** speaker write path — caller's owner member id is derived from the DB, never accepted as a parameter:

| Path | Service |
|---|---|
| `POST /chats/:id/participants` | `me-chat.ts::addMeChatParticipants` |
| `POST /orgs/:org/chats` | `me-chat.ts::createMeChat` |
| `POST /agent/chats` | `chat.ts::createChat` |
| `POST /agent/chats/:id/participants` | `chat.ts::addParticipant` |
| Agent-SDK `chat send <name>` fallback | `chat.ts::findOrCreateDirectChat` (symmetric: refuses new chat if either endpoint is private and the two ends have different owners) |

Pre-existing direct chats are **intentionally not retroactively gated** in `findOrCreateDirectChat` — an existing membership row is already a past consent artefact.

### 3. V1 one-way-door UI notice

Add-participant dropdown shows _"Adding is final in V1 — participants can't be removed yet"_, since V1 has no membership remove flow.

## Out of scope (explicit)

- **`visibility → space` field rename + DB migration.** Naming clarity vs migration cost; the rename can ship later without changing the rules above.
- **Adapter-managed chat path** (`adapter-mapping.ts::findOrCreateChatForChannel` / `ensureParticipant`). Adapter binding is a separate consent boundary — the IM admin's invite into a Feishu/Slack channel IS the consent. A future RFC should decide whether to fold this into the same owner-exclusive rule or document the alternative trust model.
- **Chat membership remove flow** — V2. Tracked in the source RFC.
[...]

PR #393, opened 2026-05-14
feat(chat): v1 group-chat focus — explicit routing, history backfill, chat-level identity, result-sink mention removal

baixiaohang:
## Summary

Hub Chat / Message v1 改造 — four changes shipped as one commit. See
[design doc](../first-tree-context/proposals/hub-chat-message-v1-design.20260514.md)
(v1.7).

- **改造 1**: `chat send` defaults to current-chat members only; new `--direct` flag for the opt-in direct-chat fallback. Closes #311's implicit side-channel chats. v1.7 dogfood follow-up removed the legacy `--chat <chatId>` flag entirely (see Breaking Changes below) and added a server-side guard that 400s any agent send with raw `@<name>` that doesn't resolve to a chat speaker.
- **改造 2**: `addParticipant()` backfills the most recent 50 messages as silent (`notify=false`) inbox rows for the new participant. v1.7 also fixes `GET /agent/chats/:chatId` to populate the `title` field via `resolveChatTitle` (topic > first-message preview > participant join) — without this the client never had a label to render.
- **改造 3**: Agent bootstrap injects a narrow chat-level identity block (`chatId / title / topic / selfOwner? / participants[{name, displayName, type}]`) into CLAUDE.md / AGENTS.md. Watcher rows excluded; internal IDs never leak. v1.7 makes the chat-context refresh **every resume** (split out a `refreshIdentityAndPrompt` helper; expensive `installFirstTreeIntegration` stays sentinel-gated) so newly-added participants surface without restarting the session.
- **改造 4**: result-sink no longer auto-injects the trigger sender mention (echo-loop fuel); `documentContext` injection + silent-turn + `inReplyTo` preserved. `tools.md` gains "Communication Rules" section with Decision guide + Fallback. v1.6 added a `purpose: "agent-final-text"` bypass channel so result-sink + AskUserQuestion forwards can land in group chats without tripping `enforceGroupMention` (and fan-out is forced `notify=false` so they never wake another session).

Drive-by fix: `git-mirror-manager.createWorktree` now runs `git worktree prune` before `add` to clear stale "prunable" admin records left over from externally-wiped worktree dirs (the "missing but already registered worktree" error caught during dogfood).

Field name unified across service / API / CLI / shared as `direct` — `git grep -nE 'allowDirect|isDirectMode|directMode'` is empty.

## Breaking changes

[...]

PR #545, opened 2026-05-25
fix(server): sink history backfill into addChatParticipants helper

baixiaohang:
## Summary

- New participants pulled into a group chat from the **web UI** read no history when later @-mentioned. The silent-context backfill landed in PR #393 only on `chat.ts::addParticipant`; the web path goes through `me-chat.ts::addMeChatParticipants` and silently skipped it. Five different join entrypoints all needed the hook, only one had it.
- **Root-cause fix**: sink the backfill invariant into `participant-mode.ts::addChatParticipants` — the helper that is, by the file-header contract, *the only place* in the codebase that writes `chat_membership.access_mode='speaker'` rows. Every join entrypoint inherits the behaviour for free, including any future one.
- Service-layer callers shed the manual hook. The hook on `chat.ts::addParticipant` (and its extra `inboxId` SELECT column) is removed.

## Design

The backfill is a data-integrity invariant: an agent crossing into `accessMode='speaker'` is only able to read history written **after** that moment (inbox fan-out joins on `access_mode='speaker'` — `message.ts:158`), so prior history needs a silent replay. Anchoring this to the row-writing helper instead of to each service entrypoint closes the whole class of bug.

Watcher → speaker upgrades also backfill: watchers never received inbox fan-out either, so the moment they're promoted is the same kind of crossing.

`addChatParticipants` now:
1. SELECTs `agents.inboxId` alongside the existing columns.
2. SELECTs the current `chat_membership` row for the (chatId, agentIds) tuple (one round-trip) to classify inputs as `alreadySpeaker` / `upgradingWatcher` / `brandNew`.
3. Performs the existing INSERT.
4. Calls `backfillSilentContextForNewParticipants` for `upgradingWatcher ∪ brandNew`.

The new SELECT is on a write path, not a hot read path — acceptable cost.

`inbox.ts::backfillSilentContextForNewParticipants`'s `tx` parameter is widened from the module-local `TxLike` to a new `WideTxLike` shaped as `PgDatabase<...>` so it accepts the `DbLike` participant-mode passes in without an `as` assertion. `TxLike` is retained for module-internal callers.

## Test plan

- [x] New file `packages/server/src/__tests__/participant-mode-backfill.test.ts` — 7 cases:
  - helper-layer invariant: brand-new join, watcher → speaker upgrade, already-speaker no-op, mixed-batch selectivity
  - regression: `addMeChatParticipants` (the original web bug), `ensureParticipant` (adapter/HTTP path) with idempotency, `PRECEDING_CONTEXT_MAX_ENTRIES` cap
- [x] `add-participant-backfill.test.ts` (pre-existing) still passes — its assertions are entrypoint-level behaviour, unchanged
- [x] `pnpm typecheck` — 13/13 tasks pass
- [x] `pnpm --filter @first-tree/server test` — **1157 tests passed** (baseline 1150 + 7 new)
- [x] `pnpm check` — no new warnings

## Follow-ups (out of scope, suggested as separate work)

There's a secondary cohesion issue this PR deliberately does not touch: the **add system** (`addParticipant` ↔ `addMeChatParticipants`) and **join system** (`joinChat` ↔ `joinMeChat`) are pairs of near-duplicate services that diverged from PR #402 and the v2 chat_membership refactor. Worth merging in a separate refactor PR after this fix lands — but the present change makes the backfill invariant robust to whatever shape those services eventually take.

### DIFF
--- a/packages/server/src/__tests__/participant-mode-backfill.test.ts
+++ b/packages/server/src/__tests__/participant-mode-backfill.test.ts
@@ -1,0 +1,244 @@
+import { and, eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import type { Database } from "../db/connection.js";
+import { chatMembership } from "../db/schema/chat-membership.js";
+import { inboxEntries } from "../db/schema/inbox-entries.js";
+import { createChat, ensureParticipant } from "../services/chat.js";
+import { PRECEDING_CONTEXT_MAX_ENTRIES } from "../services/inbox.js";
+import { addMeChatParticipants } from "../services/me-chat.js";
+import { sendMessage } from "../services/message.js";
+import { addChatParticipants } from "../services/participant-mode.js";
+import { createTestAdmin, createTestAgent, useTestApp } from "./helpers.js";
+
+/**
+ * Silent-context backfill invariant — covers every join entrypoint via the
+ * single helper `addChatParticipants` (proposal: backfill anchored to the
+ * row-write helper, not service entrypoints).
+ *
+ * Companion to `add-participant-backfill.test.ts` (which pins the
+ * `chat.ts::addParticipant` entrypoint specifically). The cases here pin the
+ * helper-layer contract so any new join entrypoint that routes through
+ * `addChatParticipants` inherits backfill automatically.
+ *
+ * Why this exists: PR #393 originally hooked backfill at
+ * `chat.ts::addParticipant`. The web "add member" button went through
+ * `me-chat.ts::addMeChatParticipants` instead, silently regressing the
+ * invariant. Sinking backfill into the helper closes that whole class of
+ * bug. These tests are the guard rail.
+ */
+
+async function countSilentEntries(db: Database, inboxId: string, chatId: string): Promise<number> {
+  const rows = await db
+    .select({ id: inboxEntries.id })
+    .from(inboxEntries)
+    .where(
+      and(
+        eq(inboxEntries.inboxId, inboxId),
+        eq(inboxEntries.chatId, chatId),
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
LGTM. 把 silent-context backfill 的不变式从 service 入口下沉到 `addChatParticipants` 这个唯一的 speaker 行写入点，是正确的根因修法 —— 关闭了 PR #393 当时漏掉 `addMeChatParticipants` 路径导致的一整类回归。

确认要点：
- 分类逻辑 (`alreadySpeaker` skip / `upgradingWatcher` 仅在 `upgradeWatcherToSpeaker` 时 backfill / `brandNew` 必 backfill) 对所有现有 callers 行为正确。新建会话路径（`createChat` / `createMeChat` / `findOrCreateChatForChannel`）在 backfill 内层 SELECT 拿到 0 条 message 时直接 early-return，相当于一次免费的 SELECT，可以接受。
- backfill 在 `inbox_entries (inboxId, messageId, chatId)` 唯一键上 `onConflictDoNothing`，对 watcher → speaker → watcher → speaker 这类轮转场景天然幂等。
- 没有 schema 变更，无需通知 human review。
- 测试用例覆盖 helper 层四态 + 三种入口（含原 web bug 的 regression case）+ `PRECEDING_CONTEXT_MAX_ENTRIES` 上限，CI Lint/Typecheck 已绿，Test job 仍在跑（不阻塞 approve）。

Follow-up: PR 描述里点到的 add 系（`addParticipant` ↔ `addMeChatParticipants`）和 join 系（`joinChat` ↔ `joinMeChat`）的去重合并，本次刻意没动是合理的 —— backfill 不变式现在已对未来重构形态稳定。

PR #549, opened 2026-05-25
fix(server): mint message id as uuid v7 per architecture rule

liuchao-001:
## Summary

`packages/server/src/services/message.ts` was minting message ids with `crypto.randomUUID()` (UUID v4) since the service was written. But [CLAUDE.md Architecture Rules](../blob/main/CLAUDE.md) state:

> **UUID v7 as Message ID:** Time-ordered; messages are immutable after creation.

Every other service in this package already uses `uuidv7()` from `src/uuid.ts` — `message.ts` was the lone holdout. This PR fixes the inconsistency.

## Why now

The mismatch was caught during PR #286 review. The web client's "New Messages" divider used lex comparison on message ids to find the first message newer than a chat-open anchor — a valid technique under v7's time-ordering, but UUID v4's random bytes meant some new messages sorted *before* the anchor and were silently misclassified as already-seen. PR #286 fixed the symptom on the web side (switched to index-based comparison). This PR fixes the root cause so future code can rely on the documented ordering invariant.

## Scope

- 1 line of executable change: `randomUUID()` → `uuidv7()` at `packages/server/src/services/message.ts:445`
- Drops `randomUUID` import (no longer used in this file); adds `uuidv7` import from `../uuid.js`
- Inline comment explaining why

## What does NOT change

- DB schema (messages.id is `uuid`, accepts both v4 and v7)
- Any query (server has no `ORDER BY messages.id` anywhere; ordering goes through `createdAt`)
- Historical data — mixed v4 (old) + v7 (new) coexist cleanly

## Test plan

- [x] `pnpm --filter @first-tree/server typecheck` ✅
- [x] `pnpm check` ✅
- [ ] CI: `Lint & Type Check` / `Test` / `Migrations` / `Forbid legacy CLI / env names`

### DIFF
--- a/packages/server/src/services/message.ts
+++ b/packages/server/src/services/message.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from "node:crypto";
 import { extractMentions, type SendMessage, scanMentionTokens } from "@first-tree/shared";
 import { and, desc, eq, lt } from "drizzle-orm";
 import type { Database } from "../db/connection.js";
@@ -9,6 +8,7 @@ import { inboxEntries } from "../db/schema/inbox-entries.js";
 import { messages } from "../db/schema/messages.js";
 import { BadRequestError, ForbiddenError, NotFoundError } from "../errors.js";
 import { createLogger, messageAttrs, withSpan } from "../observability/index.js";
+import { uuidv7 } from "../uuid.js";
 import { upsertSessionState } from "./activity.js";
 import { applyAfterFanOut, fireChatMessageKick } from "./chat-projection.js";
 import { validateDocumentContext } from "./doc-snapshots.js";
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
LGTM. 一行可执行变更修正了 message id 与架构规则 (UUID v7) 的不一致，附带的注释把 PR #286 触发的发现说清楚了。

验证了一下：
- `uuidv7()` 已在 `packages/server/src/uuid.ts` 提供，签名与 `randomUUID()` 兼容
- DB schema 不变（messages.id 是 uuid 类型，v4/v7 可共存）
- 服务端无 `ORDER BY messages.id` 依赖，历史 v4 数据安全

非核心小范围 fix，直接 approve。

baixiaohang (approved):
LGTM. 一行可执行变更把 message id 对齐到 CLAUDE.md 的「UUID v7 as Message ID」架构规则。

独立校验过：
- `uuidv7()` 与 `randomUUID()` 返回同形态字符串，drop-in 安全
- 全 server 的 messages ORDER BY 都以 `createdAt` 为主键；`chat.ts:165` 把 `messages.id` 作为 createdAt 的二级 tiebreaker，混合 v4/v7 字典序依然 deterministic，无回归
- `messages.id` 是 `text` 列（PR 描述写成 `uuid`，措辞 nit），新旧 id 共存安全
- 同包 service 全部已用 `uuidv7`，`message.ts` 确是 message-id 维度的唯一遗漏

附带的根因注释把 PR #286 触发的发现讲清楚了，将来读者能直接看懂为什么不能再用 v4。

PR #553, opened 2026-05-25
fix(status): drive composite working from per-chat runtime (not a decaying event proxy)

Gandy2025:
## Problem
Agents showed **"Idle" while actually working**: any single silent operation >60s (long `Bash`/`npm test`/build, long reasoning) and **codex-runtime long tools** (which emit no intermediate `session_events`) read as idle for the whole turn.

## Root cause + fix
The D-axis runtime state (idle/working/blocked/error) was persisted **only at agent-global granularity** (`agent_presence.runtime_state`), unusable for the per-(agent,chat) composite — an agent working in chat A would light chat B (#366). So the composite reconstructed "working" from a **`session_events` freshness proxy** that decays after `LIVE_ACTIVITY_STALE_MS` (60s) and is empty for completion-only runtimes (codex).

This PR **relocates the D-axis to per-chat** ("D′"): clients report runtime per `(agent,chat)` via a new `session:runtime` frame, stored on `agent_chat_sessions.runtime_state` + `runtime_state_at`; the composite reads it directly. `session_events` is demoted to pure description; a ~20s client re-affirm keeps a long turn fresh. Old clients (`runtime_state_at IS NULL`) fall back to the event proxy for **one release cycle**. Agent-global runtime is kept **double-written** (not removed) this PR.

### Changes by layer
- **shared**: `session:runtime` WS frame; `RUNTIME_STALE_MS`; `MeChatRow.busyAgentIds` (named to avoid the retired `workingAgentIds` wire name).
- **db (migration `0049`)**: `runtime_state` (`NOT NULL DEFAULT 'idle'`) + `runtime_state_at` (**NULLABLE, no default** — the new/old-client discriminator).
- **server**: `ws-client` routes `session:runtime` + `session:state` through the per-(agent,chat) FIFO queue (each op catches internally); `activity.setSessionRuntime` updates only `active` rows and notifies when the *effective composite* could change (value change / NULL→non-null / stale→fresh) but stays silent on a fresh same-value re-affirm; `notifier` + org WS broadcast `session:runtime`; `getChatAgentStatuses` and `deriveWorkingAgents` read per-chat runtime (authoritative) with a `runtime_state_at IS NULL` fallback, both gated on an active session.
- **client**: per-chat runtime report on change + ~20s re-affirm (independent of `evictIdle`, doesn't touch `lastActivity`); `getSessionRuntimeStates` so `fullStateSync` re-asserts real working on a network reconnect; `agent-slot`/`client-connection` wiring.
- **web**: `use-admin-ws` invalidates `chat-agent-status` + `me/chats` on `session:runtime` (throttled, not `session-events`); chat-list dots read `busyAgentIds`; components tolerate `working` with no live activity.

## Related
- Tree proposal: agent-team-foundation/first-tree-context#333 (merged).

## Verification
- `pnpm typecheck` 13/13, `pnpm check` (biome) clean.
- Tests: **server 1173** / shared 277 / web 272; client unit suite (re-affirm, `getSessionRuntimeStates`, fullStateSync) — pre-existing env-only `first-tree-cli-contract` failure is unrelated.
- Migration `0049` applied on real Postgres (testcontainers + dev-stack boot).
- **WS frame-level integration harness** (`ws-session-runtime.test.ts`): drives agent WS → server → org WS → DB for all 5 behavioral cases (long-working freshness, codex no-events working, #366 isolation, reconnect/disconnect, org WS realtime + same-value notify rules) — all green.
- Review: codex two design rounds + one code-level round → **"可提交"**; its non-blocking comment note already addressed.

## Backlog (NOT in this PR)
- Retire agent-global `agent_presence.runtime_state` (drop column; admin overview + fault notifications → per-chat).
- Human-eyes gold-standard pass (real agent in a browser) — can be done at staging/PR stage.

### DIFF
--- a/packages/client/src/__tests__/session-runtime-state.test.ts
+++ b/packages/client/src/__tests__/session-runtime-state.test.ts
@@ -59,6 +59,7 @@ function createSessionManager(opts: {
   concurrency?: number;
   log?: pino.Logger;
   onRuntimeStateChange?: (state: "idle" | "working" | "blocked" | "error") => void;
+  onSessionRuntimeChange?: (chatId: string, state: "idle" | "working" | "blocked" | "error") => void;
 }) {
   const handler = opts.handler ?? createMockHandler();
   const factory: HandlerFactory = opts.handlerFactory ?? (() => handler);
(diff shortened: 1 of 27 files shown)


Gandy2025:
Superseded by #573, which redoes this work on top of #561's one-producer / realtime-delta refactor (this PR was opened against the pre-#561 main and the conflicts are too deep to rebase cleanly).

#573 also picks up four rounds of independent codex review (R1: state-vs-runtime ordering + chat-list dot expiry; R2: extraPages bypass; R3: memo clock dep), and tightens the design: drops the old-client fallback path (force-upgraded), corrects the reverse-#366 errored axis to read per-chat instead of agent-global, and reuses #561's audience-filtered dispatchSessionFrame for delta push.

Closing in favor of #573.

PR #556, opened 2026-05-25
fix(web): server-side search in participant pickers (issue 494)

### LINKED ISSUE #494: Web: org-agents picker silently truncates orgs >100 agents
## Background

After #492, both `[+]` participant pickers (chat header, right-sidebar, new-chat chip) source from \`GET /orgs/:orgId/agents\` with \`limit: 100\`. The 100 cap is enforced by \`paginationQuerySchema\` in \`packages/shared/src/schemas/common.ts\` (\`.max(100)\`).

## Problem

In orgs with >100 visible agents, the picker silently truncates — candidates beyond the first page are simply missing from the dropdown, with no UI cue. The user has no way to discover or invite them.

## Acceptance

A user in an org with >100 visible agents can:
1. See or search for agents beyond the first page, OR
2. Be told that the list is paginated (e.g. "Type to search across all agents" empty-state hint).

## Implementation directions (not prescriptive)

- **Server-side search.** Add \`?query=\` to \`GET /orgs/:orgId/agents\` so the picker can issue a typed-search request and bypass the pagination cap for the matched subset.
- **Client-side pagination loop.** Walk \`nextCursor\` in the picker query until exhausted (or until a sane upper bound). React Query supports \`useInfiniteQuery\` for this shape.
- **Both.** Search server-side for the common path, paginate the fallback.

## Code pointers

- \`packages/web/src/pages/workspace/center/chat-view.tsx\` — the \`orgAgentsPage\` query at the top of \`ChatView\`.
- \`packages/web/src/pages/workspace/conversations/new-chat-draft.tsx\` — same shape.
- \`packages/shared/src/schemas/common.ts:3-6\` — the \`.max(100)\` cap.
- \`packages/server/src/services/agent.ts::listAgentsForMember\` — the underlying query.

## Out of scope

- Changing the cap itself — 100 is a deliberate pagination boundary, not the right knob to widen.
- Picker UX redesign — this issue is about correctness (every addable agent is reachable), not visual layout.

Flagged in review of #492 by @baixiaohang and @yuezengwu.

liuchao-001:
Closes #494.

## Summary

Picker `[+]` dropdowns silently dropped candidates past the 100-row first-page cap of `GET /orgs/:orgId/agents`. After this PR a search input drives a server-side `?query=` lookup, so every visible agent in the org is reachable regardless of org size.

## ⚠️ Behavioral change

`pickDefault` (new-chat-draft seed chip) was rewritten: default = caller's own human agent's `delegateMention`, validated as visible + not suspended; else empty. The previous `personal_assistant → first-my-managed` fallback is **gone**.

User impact: existing users who never set a `delegateMention` will see an empty chip row on `?c=draft` instead of an auto-seeded PA. Setting a delegate via the agent identity section restores the auto-seed. This is the deliberate trade flagged in the design comment on #494 (delegate is a declared signal; the old fallback was guessing).

## Server (`packages/shared`, `packages/server`)

- New `listAgentsQuerySchema` = `paginationQuerySchema` + `type` + `query`. Exported from `@first-tree/shared`. `query` is trimmed and capped at 60 chars; whitespace-only is treated as omitted (no 400).
- `listAgentsForMember` takes optional `query`. When set, splits on whitespace into AND-of-keyword `ILIKE '%escaped%'` matches across `name` + `displayName`, wrapped by `agentVisibilityCondition`. ILIKE wildcards (`%` / `_`) in user input are neutralised per-token so a `%` in one token doesn't leak into a sibling token's pattern. "Picker 110" reaches `picker-agent-110` (each token alone matches even though the literal `"Picker 110"` does not).
- `GET /orgs/:orgId/agents` parses the new schema and threads `query` through.

## Web

**`useOrgAgentsSearch(query)`** (`lib/use-org-agents.ts`)
- React Query keyed on the term. Empty query → falls back to the shared `useOrgAgents` cache. Non-empty query: 10s `staleTime`, no polling.

**`AddParticipantDropdown`** (chat header `[+]`, right-sidebar `[+]`)
- Self-contained: no `candidates` / `agentIdentity` props. Owns the search input + debounce + server fetch + avatar rendering (via the API row's `avatarImageUrl`, no identity-map round-trip).
- Search input is **always shown**.
- Already-in-chat agents that match the query render with a ✓ in their own section under the addable list (display-only, keyboard / Enter skip them). Prevents the "I just added them — why does my search say no match?" confusion.
- Enter commit is gated until the displayed list matches the typed term (Codex P2 wrong-recipient hazard fix).

**`new-chat-draft.tsx`**
- Chip picker + textarea `@` autocomplete both go through `useOrgAgentsSearch`; the autocomplete debounces the trigger query at 100ms.
- `knownAgents` accumulator (first page ∪ every search result seen) keeps chip labels and `extractMentions` stable after a search clears. Suspended / nameless rows are evicted, not just skipped.
- `<NewChatDraft key={organizationId}>` resets all draft state on org switch.
- `pickDefault` rewritten — see the **Behavioral change** section above.
- Already-in chips also render with a ✓ in the picker dropdown.

**Shared helper** `buildPickerSections(addable, alreadyIn)` in `mention-autocomplete.tsx`
- Both pickers route through it so render walk-order (`items`) and keyboard-nav order (`selectable`) come from a single source. Locked by an invariant test so a future change can't re-introduce the wrong-recipient drift where Enter commits a row other than the visible highlight.

**`rankCandidates`** (`mention-autocomplete.tsx`)
- New score-3 tier: name-substring as the lowest-tier fallback. Prefix-on-name still wins. Lets `@agent-110` reach `picker-agent-110` so the textarea autocomplete is consistent with the picker's substring semantics.

**`chat-view.tsx` + `right-sidebar/`**
- Drop `addableCandidates` derivation + `addParticipantsCandidates` / `agentIdentity` props from participant-related components.

**New** `lib/use-debounced-value.ts` — small trailing-debounce hook reused by both pickers.

## Out of scope

- Identity-map fallback for UUIDs past the first page — already noted in the issue body; right shape is a dedicated `/agents/by-ids` lookup.
- GIN trgm index on `agents.name` / `agents.display_name` — current `agentVisibilityCondition`-narrowed sequential ILIKE handles <1000-agent orgs comfortably. Reach for `pg_trgm` only when measured P50 starts biting.
- Cap on token count — 60-char schema limit already bounds worst-case AND-clauses.

## Test plan

- [x] `pnpm check && pnpm typecheck` clean
- [x] Server: 1168/1168 pass. New `?query= server-side picker search` group covers: case-insensitive substring, visibility scope, humans surface, single-token wildcard neutralisation, whitespace-only fallback, multi-keyword AND, per-token wildcard-escape independence.
- [x] Web: 275/275 pass. Adds `buildPickerSections` invariant tests (selectable order == items walk-order), `rankCandidates` name-substring fallback tests, and the rewritten `pickDefault` tests (7 cases, delegate-based).
- [x] Manual: >100 agents picker search — multi-keyword "Picker 110" reaches `picker-agent-110`. Already-in agents surface with ✓. `@agent-110` autocompletes in textarea.


baixiaohang (approved):
LGTM. 三个 commit 走下来，Codex P2 反馈 + 我之前提的阻塞项 + 作者自测发现的 UX 问题都吸收了，处理质量很扎实。

**关键加固到位**：
- Enter race（debounce/fetch 间隙提交错对象）— `searchStale` 守护 + emptyHint 同步切 "Searching…"，两个 picker 都改了
- `mergeKnown` 现在会 evict suspended / nameless 行，并加了 diff-only 写入避免 Map 引用变化触发无意义重渲染
- `<NewChatDraft key={organizationId}>` 切 org 重置，避免跨 org uuid 残留把 `@name` 解析错对象
- textarea `@` autocomplete 加了 100ms debounce（比 chip picker 的 200ms 短，权衡 popover 响应感）
- 服务端 `?query=` AND-of-tokens（"Picker 110" 命中 `picker-agent-110`）
- picker 显示已加入 agent 带 ✓（display-only，键盘 / Enter 跳过，`selectable` 收缩到 addable，`addableIdx` 单独计数避免 keyboard index 错位）
- `rankCandidates` 加 name-substring 兜底（score 3），prefix 优先级仍最高，测试锁定

**非阻塞 nit（不强制处理，留作日后顺手）**：
- `packages/server/src/services/agent.ts` 里 `if (match) conditions.push(match);` 的守护是冗余的 — `or(a, b)` 两参非空时不会返回 undefined
- ILIKE `%foo%` 走全表扫描，当前 org 规模没问题；服务里可以补一句 doc-comment，留给后人 trgm / GIN 索引的取舍指引
- `pickDefault` 行为变更（无 `delegateMention` 用户默认空 chip）虽然 commit message 明示是 close 旧 follow-up，但 PR 描述里这条对 UX 的影响埋得稍深，将来可以单独标出

yuezengwu (approved):
+1 to @baixiaohang's review — same conclusions on my pass. The two follow-up commits (635f8bbc / 79a6a563) cleanly close out the Codex P2s plus the textarea `@` debounce I would've flagged, and the multi-token + already-in ✓ + name-substring touches are nice operator-tested additions. Server ILIKE wildcard escaping is per-token and stays inside the visibility predicate; trim+max(60)+empty-string fast-path keep the route well-behaved. No new blockers from me.

yuezengwu (approved):
Re-approve on `77c32434`. That commit catches a wrong-recipient bug my prior pass missed: `items` walked the alphabetic mine-first sort while `selectable` was the raw server `desc(createdAt)` order, so the keyboard highlight could commit a different agent than the visible row. The fix routes both pickers through `buildPickerSections` so walk-order and Enter-target derive from a single source, and the invariant is locked by a unit test that also pins the expected groupAndSortCandidates output. Per-token wildcard-escape independence test (`"% 110"` shouldn't match `picker-agent-110`) is a nice adversarial lock too. No new concerns.

baixiaohang (approved):
Re-approve on `77c32434`.

承认一下：commit 3 里 `selectable = addable` 这步改动，我 first-pass review 时眼睛划过去了 —— 当时只看到「`selectable` 简化了」，没核对 `items` 走 `groupAndSortCandidates` 而 `selectable` 走原始 server order 之后两者已经失去同序保证。作者自己 manual smoke-test 后 catch 到了同一类 wrong-recipient hazard，机制还是新的：不是 codex P2 那种 stale-list race，而是同 tick 两个不同顺序 list 之间的索引错位。

修法很到位：
- `buildPickerSections(addable, alreadyIn)` 抽成共享 helper，**`selectable` 从 `headItems` 派生**（即 `groupAndSortCandidates(addable)` 剥 divider），`items` / `selectable` 由构造保证同源同序
- invariant 测试 `selectable[i] === itemsWithoutDivider[i]`，用 mine + others 混合输入触发原 drift —— 未来如果谁再把 `selectable` 接到别的源上立刻挂
- 两个 picker 路由进同一 helper，一个测试锁两处

服务端 `"% 110"` per-token wildcard-escape 独立性测试也是漂亮的对抗式 lock —— 锁的是 "先 split 后 escape" 这个顺序不变性，未来重排序会立刻挂。`listAgentsForMember` 里的 ILIKE 性能 doc-comment + PR 描述新增的 ⚠️ Behavioral change 段也把我之前剩下的非阻塞 nit 一并吸收了。

`if (match)` 守护保留是 deliberate decision（CLAUDE.md no-`as`-assertions 约定），commit message 写明，认同。

No new concerns.

PR #563, opened 2026-05-26
fix(client): narrow SSH auth heuristic so network failures don't trigger HTTPS retry

baixiaohang:
## Summary

`isLikelySshAuthFailure` matched `fatal: Could not read from remote repository.` as a credential-shaped failure, but git appends that line to **every** SSH transport failure regardless of cause — auth reject, timeout, DNS, refused. The line carries no classification signal.

On SSH-only hosts (no HTTPS credential helper configured — a documented scenario PR #291 itself targets) this caused two stacked failures whenever GitHub's port 22 had a network blip:

1. SSH network blip (timeout / refused / DNS) → misclassified as auth failure
2. Protocol-fallback path retries via HTTPS → fails immediately with `terminal prompts disabled`
3. User sees `⚠️ Session resume failed: Could not fetch ... over SSH or HTTPS. SSH attempt failed: ... Connection timed out HTTPS retry (https://github.com/) failed: could not read Username ...` in chat

The misclassification also shadowed `isLikelyTransientNetworkError` (`git-mirror-manager.ts:1102` short-circuits on credential-shaped failures), so the same SSH-side transient errors that **would have** been auto-retried with exponential backoff were instead promoted straight to a chat-visible failure on the first attempt.

### Fix

- **Drop** `/Could not read from remote repository/i` from `isLikelySshAuthFailure`. The four remaining fingerprints (`Permission denied`, `Host key verification failed`, `no matching host key type`, `no mutual signature algorithm`) are specific to genuine auth/handshake failures and fully cover the original PR #291 cases.
- **Force `LC_ALL=C`** on every spawned git subprocess so stderr stays in English regardless of operator locale. The credential-shape and transient-network heuristics all match stderr substrings; localized git output would silently break classification. Placed *after* the env spread so it overrides any inherited `LC_ALL`.
- **Doc updates** on `isLikelySshAuthFailure` and the `git()` helper explaining the new invariants.

### Why this is a real bug, not just cosmetic

Bug introduced by `9fd95439` (PR #291 — SSH support + bidirectional protocol fallback). The over-broad regex made SSH-only hosts a permanent landmine: any SSH-side network blip surfaces as a confusing dual-failure to end users instead of being retried silently. Reported by an operator with chat error matching the canonical shape above.

### Tests

- `isLikelySshAuthFailure` **negative** cases added: bare git boilerplate tail alone, plus three real multi-line stderr forms (`Connection timed out` / `Connection refused` / `Could not resolve hostname`, each combined with the trailing git boilerplate). All used to be misclassified.
- `isLikelySshAuthFailure` **positive** case updated: the `Host key verification failed` example now uses the full multi-line form (with trailing boilerplate) to confirm genuine auth errors still classify correctly even when the tail line is present.
- `isLikelyTransientNetworkError` **positive** case added: multi-line SSH timeout stderr — verifies the connected good behavior unlocked by this fix (these errors now go through transient retry instead of the credential-fallback path).

## Test plan

- [x] \`pnpm --filter @first-tree/client test\` — 528 passed, `git-mirror-manager.test.ts` 136 tests pass
- [x] \`pnpm check\` — exit 0 (only pre-existing warnings in `packages/github-scan/tests/`, unrelated)
- [x] \`pnpm typecheck\` — 13 tasks pass
- [ ] Manual validation: on an SSH-only host, observe that occasional GitHub port 22 timeouts during session resume no longer surface as `⚠️ Session resume failed: Could not fetch ... over SSH or HTTPS` in chat (the transient retry path swallows them; only persistent unreachability surfaces, and with a clean network-error message).

### DIFF
--- a/packages/client/src/__tests__/git-mirror-manager.test.ts
+++ b/packages/client/src/__tests__/git-mirror-manager.test.ts
@@ -665,8 +665,11 @@ describe("GitMirrorManager — SSH auth failure heuristic (isLikelySshAuthFailur
     "git@github.com: Permission denied (publickey).",
     "Permission denied, please try again.",
     "Permission denied (publickey,password,keyboard-interactive).",
-    "fatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.",
-    "Host key verification failed.\nfatal: Could not read from remote repository.",
+    // Real multi-line stderr: host-key reject still classifies as auth
+    // (matched by `Host key verification failed`, not by the trailing
+    // `Could not read from remote repository` line — that line on its own
+    // is no longer a signal).
+    "Host key verification failed.\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.",
     "Unable to negotiate with 1.2.3.4 port 22: no matching host key type found.",
     "Unable to negotiate: no mutual signature algorithm",
   ])("matches SSH credential-shaped error: %s", (msg) => {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM. 

**核心改动**：从 `isLikelySshAuthFailure` 中删除 `/Could not read from remote repository/i` 这个过宽的正则。git 把这一行作为通用 boilerplate 追加到**所有** SSH 传输失败之后（包括 timeout/refused/DNS），完全不携带分类信号。误匹配导致 SSH-only 主机在偶发网络抖动时被错误分类为 auth 失败，进而触发无意义的 HTTPS 回退（HTTPS 又因为没有 credential helper 立即失败），同时也短路了 `isLikelyTransientNetworkError` 的指数退避路径。

**Review 要点**：

- 保留的四个指纹（`Permission denied`、`Host key verification failed`、`no matching host key type`、`no mutual signature algorithm`）已经精确覆盖 PR #291 引入的真实 auth 场景，narrowing 是安全的。
- `LC_ALL=C` 这个加固也合理：现有 heuristic 全是英文 substring 匹配，本地化 stderr 会静默破坏分类。放在 spread 之后是有意作为不变量，docstring 解释清楚了。
- 测试覆盖很扎实：positive case 升级成了完整多行形式（证明真实 auth 错误带着 tail 依然能命中），新增的 negative cases 覆盖了三种实际的多行网络错误形态，并在 transient-network 正向用例里加了一条验证修复后的"应有行为"。
- 无数据库 / 数据结构变更，纯客户端 heuristic 修复。

不阻塞合并。

PR #561, opened 2026-05-25
refactor: one producer for per-(agent,chat) status + realtime delta

Gandy2025:
## Summary

Consolidates the per-(agent,chat) composite-status data path and moves its realtime delivery from refetch-on-invalidate to push-a-delta. Two commits:

**1. `refactor(server,web): one producer for per-(agent,chat) status`**
- The composite status was computed in **three** places with divergent agent sets (`getChatAgentStatuses`, and `listMeChats`'s `deriveLiveActivity` / `derivePendingQuestions` / `deriveFailedAgents`), with the `errored` predicate duplicated as TS + SQL "kept in lockstep by test" and the live-activity LATERAL written twice. Now a single producer `resolveAgentChatStatuses` backs both `GET /chats/:id/agent-status` and the chat list.
- Folds the closed **PR #558** `turnText` feature (current-turn narration, sticky across tool calls, cleared past `turn_end`) into the producer.
- Drops the dead `MeChatRow.engagedAgentIds` (no UI consumer — a rename-survivor of the removed `workingAgentIds`).
- Narrows the chat-list live-dot source to non-human speakers (drops stray dots from agents that left the chat; never lights for a human predictive-active session). This also neutralises a latent landmine where the predictive write created `state='active'` sessions for human recipients.

**2. `feat(web,server): realtime per-(agent,chat) status delta`**
- `session:state` / `session:event` admin-WS frames now carry the recomputed `AgentChatStatus`, delivered **only to sockets whose viewer can access the chat** (the status carries the agent's narration, so it is never broadcast org-wide). Web patches `["chat-agent-status", chatId]` in place instead of refetching; non-audience sockets get the bare routing frame exactly as before.
- Adds `LiveActivity.staleAt`; the compose bar / agent panel run a 1s ticker that self-clears a stale "working" chip at expiry (re-deriving `main` locally via the shared `deriveMainStatus`) instead of waiting for the 30s refetch.
- The `me/chats` invalidation and the 30s `refetchInterval` floor are retained, so a lost cross-instance delta degrades to today's latency — never worse.

Behavior is preserved for every rendered signal except the two intended deviations above. The single projection function `deriveMainStatus` (shared) remains the only place composite `main` is computed.

## Test plan
- [x] `pnpm typecheck` — 13/13
- [x] `pnpm check` (biome) — clean
- [x] `@first-tree/shared` — 272/272
- [x] `@first-tree/web` — 308/308 (incl. new `clearStaleWorking` / `upsertAgentStatus` unit tests)
[...]

PR #573, opened 2026-05-26
fix(status): drive composite working+errored from per-chat runtime, not a decaying event proxy

Gandy2025:
## Summary

- **Truth source swap.** Composite `working` / `errored` no longer reconstructed from a 60 s `session_events` freshness proxy (which missed long silent ops and codex-runtime turns) and no longer OR'd from agent-global `presence.runtime_state` (which leaked errors across chats, the reverse of #366). Both axes now read authoritatively from new `agent_chat_sessions.runtime_state` + `runtime_state_at` (per-chat, freshness-stamped), with client re-affirms every ~20 s ±20 % jitter. Old-client fallback retained for one release cycle per spec §10.
- **Done on top of #561.** Reuses the existing audience-filtered `dispatchSessionFrame` for `session:runtime` delta push — no separate invalidate path. `session:state` and `session:runtime` now share the per-(agent,chat) `chainSessionOp` FIFO so runtime frames can't race ahead of the active gate.
- **Redo of original #553** (which was opened before #561 landed and therefore conflicted with its consolidation refactor).

## What this fixes for users

- Long Bash / `npm test` / build / long-reasoning turns no longer flip the agent to idle after 60 s of SDK silence.
- Codex agents (turns that emit no intermediate `session_events`) now show `working` for the duration of the turn instead of looking idle the entire time.
- An agent that errors in chat A no longer surfaces as `failed` on every other chat it participates in (new-client path only — old-client fallback during the one-cycle upgrade window preserves the pre-PR behaviour, including the latent leak).

## Architecture notes

- **Old-client fallback retained for one release cycle** per spec `proposals/hub-agent-status-working-freshness.20260525.md` §6.1 §10. NULL `runtime_state_at` → fall back to legacy `session_events` proxy for `working` and legacy `presence.runtime_state === 'error'` OR-fold for `errored`. Self-closes the moment the client upgrades and reports per-chat. No server-side min-SDK gate (could be added in a follow-up).
- **`errored` axis upgraded to per-chat in this PR** (intentional resolution of spec §10 Q1, which tagged it "倾向 backlog"). The reverse-#366 regression test was concrete enough defense to justify shipping now rather than deferring; both authoritative + fallback paths are tested explicitly.
- **`agent_presence.runtime_state` double-write retained.** Admin overview / fault-notification paths still read the agent-global aggregate, and the new-client error path doesn't replace it. A follow-up PR will retire the column once those consumers migrate.
- **Stuck-working SLO.** Client crash mid-turn → `runtime_state_at` ages past `RUNTIME_STALE_MS` (60 s) → server fail-closes → first-page chat-list dot self-heals within 60 s, extraPage rows within ~90 s (`RUNTIME_STALE_MS + refetchInterval`).
- **Atomic `setSessionRuntime`.** Single CTE `UPDATE … WHERE state='active' RETURNING prev`. Active gate is a SQL invariant, no longer dependent on `chainSessionOp` serialisation.
- Retired the 1 s `clearStaleWorking` UI ticker — dead code now that the server pushes deltas authoritatively.

## Process

- Four rounds of independent codex code review on the initial implementation (rounds 1–3 produced 1 P1 + 3 P2 findings, all addressed in `0a910e69`; round 4 cleared the diff).
- Round 5 human review by @baixiaohang flagged spec divergences (fallback removal, constants 90/30 vs spec 60/20, atomicity, comment overpromise); all reconciled in `5b7e1d0d`.
- Migration `0049` (`runtime_state` + `runtime_state_at`) applied on testcontainers Postgres in the server suite.

## Test plan

- [x] `pnpm typecheck` (13 / 13 packages)
- [x] `pnpm check` (biome — clean on touched files; 16 pre-existing warnings in `packages/github-scan/tests/` are unrelated)
- [x] `@first-tree/server` 1212 / 1212 (post-review)
- [x] `@first-tree/client` 545 / 545 (excluding pre-existing env-flake `first-tree-cli-contract.integration`)
- [x] `@first-tree/web` 330 / 330
- [x] `@first-tree/shared` 280 / 280
- [x] New regression tests:
  - Producer working / errored truth table — both new-client authoritative path AND old-client fallback path explicitly asserted (incl. reverse-#366: presence='error' lights errored ONLY on NULL-stamp rows, NOT on rows that have ever reported per-chat).
  - `setSessionRuntime` notify matrix (value-changed / was-stale-incl-NULL / fresh-same-value-silent), now backed by the atomic CTE.
  - Per-chat client callback + ~20 s jitter re-affirm + `getSessionRuntimeStates`.
  - `busyAgentIds` e2e covering the codex no-events case.
  - State-before-runtime ordering regression pinning `notifySessionState("active")` fires before `handler.start`.
- [ ] Manual: spin up the dev stack and walk a long Bash turn, a codex agent, and a per-chat error to eyeball the three user-visible behaviours


baixiaohang (changes requested):
Thanks for the thorough rebase and the rounds of codex review — the producer-side rewrite, the `chainSessionOp` extension to `session:state`, the `notifySessionState("active")`-before-`handler.start` ordering fix, and the reverse-#366 regression coverage are all clean and well-tested. Requesting changes mainly on **spec divergences that should be reconciled before this lands**, plus a few smaller items.

---

### P1 — blocker

#### 1. Old-client fallback removal contradicts the approved spec

The approved proposal (`proposals/hub-agent-status-working-freshness.20260525.md` §6.1, §10) explicitly says:

> 新旧路径用 `runtime_state_at IS NULL` 严格二选一,不并存、不双写 OR。兼容期满删 `否则` 分支即可。
> （保留旧 client 兼容兜底**一个发布周期**）

This PR removes the fallback entirely — the producer reads NULL `runtime_state_at` as fail-closed (not working / not errored). The commit message frames this as safe because "all clients are force-upgraded," but:

- I couldn't find any server-side minimum-client-version gate in `packages/server` — no rejection of binds based on `clients.sdk_version`.
- Client self-update is a per-client policy (`off` / `prompt` / `auto`) — for `off` and `prompt` (unattended TTY) clients, a server deploy lands before the client restarts.

The blast radius during the upgrade window: every chat owned by a non-upgraded client only emits `session:state active` (no `session:runtime` frame ever), so `runtime_state_at` stays NULL → composite reads `working=false` → **all those chats display idle while actively working**. That is the *exact reverse* of the bug this PR fixes.

Two acceptable resolutions, pick one:

- **(a)** Restore the spec's `runtime_state_at IS NULL → fall back to event proxy` branch in `computeWorking` for one release cycle, with a follow-up issue to remove it. This is what the proposal called for and is the safest path.
- **(b)** Land a server-side minimum-SDK-version gate that refuses to `bind` non-upgraded clients (and document the deploy ordering — gate first, this PR second), then update the spec proposal to record the new contract.

If (b), the spec node should be updated in the same series of PRs so the documented contract and the implemented one don't drift.

---

### P2 — should fix or reconcile with spec before approve

#### 2. RUNTIME_STALE_MS 90s + re-affirm 30s ±20% diverges from spec (60s / ~20s)

Spec §6.1 / §9: `RUNTIME_STALE_MS = 60_000`, re-affirm ~20s (3×). PR: 90s / 30s ±20% (≈3× nominal, 2.5× at max jitter — the commit message's "≥ 3× re-affirm interval" claim is slightly off).

User-visible side effect: the documented "stuck-working SLO on client crash" goes from ~65s to 90s (and ≤120s when an extraPage row is involved — see #5 below).

Pick one:
- Restore 60s / ~20s to match the approved design, or
- Keep 90s / 30s ±20% and update the proposal node so spec and code stay in sync.

#### 3. Per-chat `errored` upgrade is a scope expansion past the spec's "backlog" tag

Spec §6.1 last bullet + §10 Q1: "errored 升级为 per-chat … 可选、留 backlog" (tendency: backlog). This PR ships it now — fine on its merits and the reverse-#366 regression test is a real defense — but the PR description doesn't acknowledge it as a deviation from the spec's deferral. Worth at least noting in the description so the reviewer knows it's an intentional Q1 resolution, not an oversight.

#### 4. `setSessionRuntime` is SELECT-then-UPDATE, not atomic

`packages/server/src/services/activity.ts:setSessionRuntime` reads `prev` then writes — race-free today only because `chainSessionOp` serialises calls per-(agent,chat) at the WS layer, and R-RUN today admits at most one client process per agent. Both invariants are implicit. A single conditional `UPDATE … WHERE agent_id = ? AND chat_id = ? AND state = 'active' RETURNING runtime_state, runtime_state_at` would:

- be atomic at the SQL level,
- shed one round-trip,
- let the notify gate read the *returned* row instead of the pre-read,
- remove a class of "future caller forgot to chainSessionOp" bugs.

Not strictly broken today, but worth tightening while we're in here.

---

### P3 — nits

#### 5. `extraPages` staleness window comment slightly overpromises

`pages/workspace/conversations/index.tsx` — the memo depends on `dataUpdatedAt`, which only ticks on react-query refetch (~30s). So an extraPage row's busy-blank can lag up to one refetch interval past `RUNTIME_STALE_MS` — the realistic upper bound is `RUNTIME_STALE_MS + refetchInterval` (≤120s with current values), not "within the documented 90s SLO" as the comment near `extraPages` claims. One-line comment tweak.

#### 6. `activity: working ? activity : null` transient inversion

In `agent-chat-status.ts` `resolveAgentChatStatuses`, activity is now dropped whenever `working=false`. When a `session:event` arrives slightly before its corresponding `session:runtime working` (different code paths — events aren't in `chainSessionOp`), the recompute will briefly emit `activity=null` and the next runtime frame self-heals. Tiny window, and it does match the spec §6.1 "activity is description-only when working" contract — recording for posterity, not blocking.

---

### What I like

- The `chainSessionOp` extension to `session:state` plus the `notifySessionState("active")`-before-`handler.start` ordering fix, both pinned by the dedicated regression test, are the right shape — solves codex's "synchronous setRuntimeState inside start() that awaits the whole turn" without papering over it.
- The `setSessionRuntime` notify matrix (value-changed / was-stale-incl-NULL / fresh-same-value-silent) is well factored and exhaustively tested. No re-affirm spam.
- The codex no-events e2e regression in `me-chats-http-e2e.test.ts` directly exercises the scenario `liveActivity` could never cover.
- `busyAgentIds` naming (deliberately not `workingAgentIds`) — thoughtful avoidance of resurrecting the #366-poisoned identifier.
- Migration 0049 is safe (NOT NULL DEFAULT on `runtime_state`, NULL-no-default on `runtime_state_at` preserving the sentinel meaning).
- Retiring `use-now.ts` + `clearStaleWorking` is the right call once the server pushes deltas authoritatively.

Happy to re-review once #1 has a decision (fallback restored *or* min-version gate + spec update) and #2/#3 are reconciled with the spec node.

Gandy2025:
Thanks @baixiaohang — you were right on all four spec-divergence calls. Pushed `5b7e1d0d` addressing everything:

---

**P1 — Old-client fallback restored (option a)** ✅

Took your option (a). `computeWorking(session, activity, now)` and `computeErrored(session, presenceRuntimeState, now)` now fall back to the legacy event proxy + agent-global OR-fold when `runtime_state_at IS NULL` (= old client never reported per-chat). New-client path (any non-null stamp) stays clean — reverse-#366 leak only present on the fallback path during the upgrade window, and it self-closes the moment a client reports. Matches spec §6.1 §10 verbatim now.

Pure-helper tests cover both paths (4 new cases: new-client authoritative working/errored + old-client fallback working/errored). The reverse-#366 SQL-level regression test now asserts **both** behaviours explicitly — NULL stamp → errored=true (legacy), non-null stamp → errored=false (new-clean).

**P2 #2 — Constants back to spec** ✅

`RUNTIME_STALE_MS` 90s → **60s**, `RUNTIME_REAFFIRM_BASE_MS` 30s → **20s**. The "3× nominal" margin claim is now actually 3×. Reaffirm test updated to reflect the 20s base (still 60s test window — even more margin now).

**P2 #3 — `errored` per-chat scope expansion noted in PR description** 📝

Will update the PR description to call this out as an intentional in-PR resolution of spec §10 Q1 ("倾向 backlog" → resolved as "ship now"), since the reverse-#366 regression test is genuinely valuable defense even without the rest of the per-chat error feature work. Doing the PR description edit right after this comment lands.

**P2 #4 — Atomic UPDATE … RETURNING (CTE)** ✅

`setSessionRuntime` is now one round-trip:
```sql
WITH prev AS (SELECT runtime_state, runtime_state_at FROM agent_chat_sessions WHERE …)
UPDATE agent_chat_sessions
   SET runtime_state = ?, runtime_state_at = NOW()
  FROM prev
 WHERE … AND state = 'active'
RETURNING prev.runtime_state, prev.runtime_state_at
```
Active gate is now a SQL invariant, no longer dependent on chainSessionOp serialisation. Notify gate reads the RETURNING prev values. Switched to server-side `NOW()` to avoid postgres-js's raw-execute Date-binding limitation. All 11 existing `setSessionRuntime` tests pass unchanged — only the writer changed, behaviour didn't.

**P3 #5 — extraPages SLO comment corrected** ✅

Updated to state the real bounds: first-page rows recover within `RUNTIME_STALE_MS` (60s); extraPage rows up to `RUNTIME_STALE_MS + one refetchInterval` (~90s upper). Previously misclaimed as "within 90s SLO" period.

**P3 #6 — transient activity=null window documented** ✅

Added a paragraph to the producer comment near the `activity: working ? activity : null` line explaining the brief inversion when a `session:event` recompute races marginally ahead of its `session:runtime working` (different code paths — events don't share chainSessionOp). Self-heals on the next runtime frame.

---

**Verification** (post-fix)
- typecheck (13/13) ✅
- biome clean on touched files ✅
- server 1212/1212 (was 1209; +3 fallback-path helper tests)
- client 545/545, web 330/330, shared 280/280

Ready for re-review whenever you have time. 🙏

baixiaohang (approved):
LGTM — `5b7e1d0d` reconciles all four findings cleanly against the spec.

Verified end-to-end:
- **P1 fallback restored.** `computeWorking(session, activity, now)` and `computeErrored(session, presenceRuntimeState, now)` correctly bifurcate on `runtime_state_at IS NULL`. Once a client has ever reported per-chat, the new path locks in and `RUNTIME_STALE_MS` decides — the "stale stamp doesn't regress to legacy" semantic is a sensible read of spec §6.1 §10 (the spec only mandates NULL→fallback, not stale→fallback). Reverse-#366 split into two regression cases (new path: errored=false, fallback path: errored=true) is the right shape.
- **Constants.** `RUNTIME_STALE_MS` = 60_000, `RUNTIME_REAFFIRM_BASE_MS` = 20_000 — exact spec.
- **Atomic CTE.** `WITH prev AS (SELECT …) UPDATE … FROM prev WHERE state='active' RETURNING prev.*` makes the active gate a SQL invariant and shaves the round-trip. `NOW()` server-clock + the `as unknown as Array<…>` cast (with the postgres-js raw-execute Date-binding rationale inline) both check out against the repo's no-`as`-without-comment rule.
- **`errored` per-chat scope expansion** is now explicitly called out in the PR description as the intentional resolution of spec §10 Q1.
- **extraPages SLO comment** correctly states the real `RUNTIME_STALE_MS + refetchInterval` upper bound.
- **Activity transient inversion** documented in the producer.

Tests:
- New helper-pure cases cover both authoritative and fallback paths for `computeWorking` / `computeErrored`.
- SQL-level reverse-#366 regression pinned at both NULL stamp and non-NULL stamp.

Ship it.

PR #580, opened 2026-05-26
feat(doc-preview): linkify inline-code paths + inert chips for failed mentions

Gandy2025:
## Summary

Two-phase upgrade to chat doc preview, per [proposals/20260526-doc-preview-inline-code-and-inert-chip.md](https://github.com/agent-team-foundation/first-tree/blob/feat/doc-preview-inline-code/proposals/20260526-doc-preview-inline-code-and-inert-chip.md):

- **Phase 1 — inline-code paths become clickable.** Bare \`.md\` tokens wrapped in single-line inline code (`` `docs/foo.md` ``) used to be a hard scanner skip, dropping the most common agent habit (mono-spaced visual intent) back to plain text. Scanner now reports them with an `enclosingCodeSpan` annotation; runtime rewrite widens to the whole tick span, producing the commonmark-legal `` [`docs/foo.md`](docs/foo.md) `` — a code-styled clickable link. Fenced blocks, HTML, reference-link defs, domain-shaped tokens, hidden segments stay hard skips. Web fallback mirrors the runtime so legacy pre-rewrite messages render identically.
- **Phase 2 — failures become visible.** Doc-mention failures (missing / out-of-fence / hidden / too-large / budget / unreadable) used to silently degrade to plain text. Runtime now collects bare-source failures into `metadata.documentContext.failedMentions[]` (deduped by writtenPath); chat-view wraps each unrendered token into a disabled "doc chip" with a reason-tooltip via the magic `#doc-failed?reason=…` href that the `a` override picks up. The `docs.min(1)` constraint is replaced by a refinement requiring docs OR failedMentions to be non-empty, so an all-failed message still ships a usable documentContext.

Implemented per the proposal, with two pre-merge code reviews (codex + independent agent) and fixes baked in.

## Files

- \`packages/shared/src/lib/doc-link-scan.ts\` — scanner reports `enclosingCodeSpan`; backtick added to path boundary char class.
- \`packages/shared/src/schemas/me-doc.ts\` — `failedMentions[]`, `DocSnapshotFailReason` enum, refinement.
- \`packages/client/src/runtime/doc-snapshots.ts\` — code-span widening rewrite, per-mention `failReason` tracking, `classifyOccurrenceFailure` helper.
- \`packages/client/src/runtime/result-sink.ts\` — attaches `failedMentions` when present.
- \`packages/web/src/lib/doc-preview-links.ts\` — `wrapFailedDocMentions`, `parseFailedDocHref`, code-span-aware widening on the fallback path.
- \`packages/web/src/pages/workspace/center/chat-view.tsx\` — inert chip rendering, gated on the message actually carrying failedMentions metadata so user-typed magic hrefs can't spoof.

## Pre-merge reviews

- **Round 1 — \`codex review --base main\`:** flagged 1 × P3 (`../outside.md` classified as `hidden-segment` instead of `out-of-fence`). Fixed in commit 3c7a798d + test.
- **Round 2 — independent agent re-read:** flagged 2 × P3 + 1 × P2 (markdown-injection concern). Fixed in commit 13f8f300:
  - User-typed `[x](#doc-failed?reason=…)` could spoof the inert chip → gate detection on the message carrying failedMentions metadata.
  - `textOfChildren` didn't recurse into code-span children → dropped the manual `aria-label`; rely on visible text + `title`.
  - P2 markdown-injection (brackets inside code-span widening) verified UNFOUNDED via `mdast-util-from-markdown` against CommonMark §6.3 + §6.6 — added a regression test to pin the verified behaviour.

## Deployment ordering note

The schema refinement on \`snapshotDocumentContextSchema\` relaxes `docs.min(1)`. Server + client schema ship together from this PR via \`@first-tree/shared\`. During a rolling deploy where new agent CLI runs against an older server image, an "all-failed mentions, zero snapshots" message would be rejected by the old server's `min(1)` validation. Mitigation: deploy the server (which carries the new shared schema) before broadly publishing the new CLI tarball. The mixed case (≥1 snapshot + ≥1 failed mention) is unaffected.

## Test plan

- [x] \`pnpm --filter @first-tree/shared test\` — 281 passed (incl. new `enclosingCodeSpan` reporting).
- [x] \`pnpm --filter @first-tree/client test\` — 559 passed (incl. 12 new doc-snapshot tests covering inline-code widening + failedMentions classification).
- [x] \`pnpm --filter @first-tree/web test\` — 349 passed (incl. wrap/parse helpers + bracket-in-code-span regression).
- [x] \`pnpm --filter @first-tree/server test\` — 1214 passed (snapshot validation unchanged; refinement still rejects empty docs + empty failedMentions).
- [x] \`pnpm check\` — clean (warnings unchanged from main).
- [x] \`pnpm typecheck\` — clean.
- [ ] Live verification on staging: agent sends `` `proposals/foo.md` `` → renders as clickable code-styled chip.
- [ ] Live verification on staging: agent sends `docs/does-not-exist.md` → renders as disabled chip with "文档不存在" tooltip.


yuezengwu:
Review pass — implementation matches the proposal, 2 prior code reviews already baked in, tests are comprehensive (scanner `enclosingCodeSpan` reporting; runtime widening + all 6 fail-reason buckets + dedupe by writtenPath + mixed success/failure + inline-link silent + multi-path-in-one-span degenerate; web fallback widening + `build/parseFailedDocHref` round-trip + wrap pass + the brackets-in-code-span CommonMark regression).

Spotted nothing blocker-grade. A couple of points worth surfacing rather than fixing:

**1. Wire schema relaxation — deployment ordering risk (already noted in PR body).** The refinement on `snapshotDocumentContextSchema` (in `packages/shared/src/schemas/me-doc.ts`) replaces `docs.min(1)` with "docs OR failedMentions non-empty". This is an additive, backward-compatible Zod-layer change (no DB migration), but during a rolling deploy where the new CLI hits an older server image, an all-failed/zero-snapshot message will be rejected by the old `min(1)`. The PR body's mitigation (deploy server-carrying-new-shared-schema first, then publish CLI tarball) is the right plan — flagging so the release ordering is honoured.

**2. Pass-2 failReason propagation across duplicate keys (correct, just calling it out).** When a second occurrence of the same canonical key enters the `seen.has(key)` branch, it adopts the prior occurrence's `failReason` via `resolved.find(...)`. This correctly covers reasons set in Pass 2 (`too-large` / `budget-exceeded` / `unreadable`), not just Pass-1 reasons. Verified by tracing — no change needed.

**3. Spoof gate is solid.** The chat-view `a` override conditioning on `failedDocMentions.size > 0` AND `parseFailedDocHref(href)` matching a known enum prevents a user-typed `[x](#doc-failed?reason=missing)` in a web-source message from rendering as a system chip. Good fix from round 2.

Holding approval pending @yuezengwu's ack on the deploy-server-first ordering (per the working agreement that wire/data-structure changes go through human review). Everything else LGTM.

Gandy2025:
Thanks for the careful pass, @yuezengwu.

**Deploy ordering — explicit ack.** I'll honour the server-first sequence:

1. Server (Docker image built from this commit) deploys to staging → prod first. New `@first-tree/shared` ships with the relaxed `snapshotDocumentContextSchema` refinement, so server accepts both shapes:
   - **Old** `{kind: "snapshot", docs: [doc, …]}` — unchanged, still valid (docs non-empty satisfies the refinement).
   - **New** `{kind: "snapshot", docs: [], failedMentions: [...]}` or `{kind: "snapshot", docs: [...], failedMentions: [...]}` — newly accepted.
2. CLI tarball (which carries the new runtime emission of `failedMentions`) ships to npm only after server is rolled out.

No version skew window with message loss: a new agent CLI's all-failed message reaches the new server's relaxed validator. An old agent CLI's message (always docs-non-empty under the old code path) reaches whichever server, both validate fine.

If staging surfaces any wire-format issue I'll flag before we cut a CLI release.

Also confirmed your two observations:

- **Pass-2 reason propagation across duplicate keys** — the `resolved.find(o => o.key === key && o.failReason !== undefined)` does pick up `too-large` / `budget-exceeded` / `unreadable` set inline in Pass 2, not just Pass-1 buckets. Each fail branch mutates `occ.failReason` on the resolved entry before `continue`, and the prior-lookup runs over the same array.
- **Spoof gate** — chip rendering is gated on both `failedDocMentions.size > 0` (so user-source messages without metadata never even reach the magic-href check) and `parseFailedDocHref(href) !== null` (so an unknown reason value falls through to the regular `<a>`).

Standing by for green CI before you ack.

baixiaohang (approved):
LGTM. Independent review pass — implementation matches the proposal claims, two pre-merge code reviews + @yuezengwu 的复核都已 baked in，作者 deploy-ordering ack 完整可执行。

**复核到的几点**

- 三处 code-span widening（scanner 注解 / runtime rewrite / web fallback）逻辑一致；spoof gate（chat-view 同时检查 `failedDocMentions.size > 0` 与 `parseFailedDocHref` 命中 enum）防御到位。
- Pass-2 `failReason` 在重复 key 上的 propagation 正确，能覆盖 `too-large` / `budget-exceeded` / `unreadable` 这类 Pass-2 inline 设置的 reason。
- 测试覆盖完备：12 条新 doc-snapshot 用例 + `../outside.md` 回归 + bracket-in-code-span 的 CommonMark 回归 + dedupe by writtenPath + 多路径同 span 退化。
- Deploy 顺序按已 ack 的方案执行（server 先 → CLI tarball 后），混合形态（≥1 snapshot + ≥1 failedMention）不受版本错位影响。

**一条 non-blocker 备注（不阻 merge）**

PR body 引用的 `proposals/20260526-doc-preview-inline-code-and-inert-chip.md` 在当前 branch tree 里不存在，链接 404。实现本身可独立审，但 first-tree 的约定通常会把 proposal 与实现一起带入仓库 —— 方便后续 trace 起见，建议作为 follow-up 把 proposal 补到 `proposals/` 下（或者在 first-tree-context 的 proposals/ 下落档亦可）。

Approving 待 CI 绿后由 maintainer 决定 merge 时机。

Gandy2025:
Thanks @baixiaohang for the careful pass and the proposal-link catch.

**Fixed in d1feea92** — landed the proposal at [proposals/20260526-doc-preview-inline-code-and-inert-chip.md](https://github.com/agent-team-foundation/first-tree/blob/feat/doc-preview-inline-code/proposals/20260526-doc-preview-inline-code-and-inert-chip.md) (new top-level `proposals/` dir). The PR-body link now resolves; future readers can trace the design + the 6 open decision points + the 19-session implementation data alongside the code.

Approvals from both reviewers + deploy-ordering ack stand; standing by for CI green + maintainer merge.

Issue #583, opened 2026-05-26
slash-commands: bump SkillDescriptor.description max + lower array cap + explicit bodyLimit


yuezengwu:
Follow-up from PR #588 (originally raised in #581 review).

## Problem

`skillDescriptorSchema.description.max(2000)` is **silently** dropping skills with long descriptions:

- `discoverClaudeCodeSkills` runs every parsed frontmatter through `skillDescriptorSchema.safeParse`.
- On failure → emit a `warn` and **skip the entire skill**.
- Several real Claude Code skills (e.g. `qa`, `cso`, multi-paragraph gstack skills) blow past 2000 chars in their description block.
- User-visible effect: skill silently disappears from the slash popover with no UI signal.

Truncate-and-warn at scan time is the wrong fix — it silently rewrites authored content. Right fix is to raise the schema ceiling + truncate only at display.

## Proposed change

1. `skillDescriptorSchema.description.max(2000)` → **`.max(8000)`** — buffer for multi-paragraph descriptions that a future "Skill details" panel could render in full.
2. `agentSkillsSchema.max(2000)` → **`.max(200)`** — well above any real user's skill count (current ceiling is ~50). Keeps the worst-case payload from ballooning into "definitely needs streaming" territory.
3. With (1)+(2) the theoretical payload max becomes 200 × 8000 ≈ 1.6 MB. Add explicit route-level cap so this is visible and intentional:

```ts
app.patch<{ Params: { uuid: string } }>("/:uuid/skills", { bodyLimit: 2_097_152 }, async (request, reply) => { ... });
```

(Fastify's default `bodyLimit` is 1 MB — would 413 before our `agentSkillsSchema` zod check ever ran.)

## Reasoning the reviewer flagged (#3 + #5 bundled)

> 同意你的方向（schema 抬到 ~4000 + UI 截断），不要 truncate-and-warn —— scan 阶段截断会把原始内容静默改写。我倾向再抬高一点到 8000，因为 `qa` / `cso` / `gstack` 这类丰富的 skill 描述可能多段落，留 buffer 让以后的 Skill 详情面板能展开。

> 跟 #3 联动 —— 如果 #3 抬到 8000 字符，那 `agentSkillsSchema.max(2000)` 的上限就成了 16MB 理论值，明显超过 Fastify 默认 1MB。建议在同一个 follow-up 里把 skills 数组上限从 2000 降到 200，路由侧加 `{ bodyLimit: 2_097_152 }`。

[...]

PR #581, opened 2026-05-26
feat(slash-commands): per-agent slash-command catalog + composer popover

yuezengwu:
## Summary

End-to-end slash-command pipeline: type `/` after `@mention`ing an agent in the composer to pick from that agent's available skills (the same skills Claude Code surfaces via its Skill tool).

- **Data layer**: `agents.skills jsonb NOT NULL DEFAULT '[]'`; new `GET/PATCH /api/v1/agents/:uuid/skills` (visible/manage scope), `SkillDescriptor` Zod schema in `@first-tree/shared`.
- **Daemon**: scans `~/.claude/skills/*/SKILL.md` + `~/.claude/plugins/<plugin>/skills/*/SKILL.md`, parses YAML frontmatter, uploads via PATCH on every daemon start (snapshot semantics; codex/human agents skipped). Failures log + continue — never blocks startup.
- **Web composer**: `useSlashCommand` hook + `SlashCommandPopover` (mirrors `useMentionAutocomplete` shape). `/` triggers only at start of message (composer-mode). Skill list follows the **most recent `@mention` before the cursor**; 1-on-1 chats fall back to the sole speaker so `/` works without `@`. v1 ships one system command (`/clear`); `/help|/me|/invite` deferred pending UX surfaces.

## Design alignment

Tracks the design discussed in chat (薄系统 + 厚 agent): the frontend only proposes candidates; agent skills are sent literally to the recipient agent whose harness routes them. Mention-context binding is a hard rule — group chats never collapse multi-agent skill lists.

## Test plan

- [x] `pnpm typecheck` — shared / server / client / cli / web all clean (incl. web design-token guardrails)
- [x] `pnpm check` — 0 errors (16 pre-existing warnings untouched)
- [x] `pnpm test` —
  - shared: 272 / server: 1188 (+4 new `agent-skills.test.ts`) / client: 531 (+7 new `skills-discovery.test.ts`) / cli: 236 (+3 new in `runtime-provider-reconcile.test.ts`) / web: 353 (+19 new `slash-command-autocomplete.test.ts`)
- [x] Live E2E against firsttreehub dev DB: real `discoverClaudeCodeSkills` against `\$HOME/.claude/skills` (42 skills) → real `uploadAgentSkills` PATCH → GET round-trip identity verified
- [x] Browser smoke via `/login` dev-callback as `devuser` against group chat with seeded skills on `reviewer-bot` / `planner-bot`

## Migration / DB changes ⚠️

[...]

PR #588, opened 2026-05-27
feat(slash-commands): per-agent slash-command catalog + composer popover

### LINKED ISSUE #584: slash-commands: document daemon-scanner ↔ agent-runtime $HOME invariant
> Scoped down after the routing half was resolved by a direct SDK probe in #588.
> See https://github.com/agent-team-foundation/first-tree/issues/584#issuecomment-4550740472 for the routing verification — `[From: alice]\n\n/review` correctly invokes the Skill tool; First Tree's `formatInboundContent` prefix is actually *necessary*, not a problem.

## Remaining problem

The slash-command popover lists skills the **daemon scanner** discovered at `$HOME/.claude/skills` (and `$HOME/.claude/plugins/.../skills/`) at startup. The recipient agent's Claude Code process must be able to invoke those exact skills — which means the agent runtime needs the same `$HOME` tree the scanner read from.

In the default deployment this holds by construction: `discoverClaudeCodeSkills` runs in the same Node process tree as the daemon, which spawns the Claude Code child with no `$HOME` rewrite, no chroot, no container isolation. So scanner ≡ runtime.

**Risk surface:** any future deploy boundary that introduces process isolation (separate uid, container per-agent, sandbox) silently breaks the invariant. Symptom: `/review` appears in the popover, message reaches the agent, agent's Skill tool returns "skill not found", user gets a confusing error rather than a clear "this skill isn't available here" up-front.

## Action

- [ ] Add a paragraph-level invariant doc to `packages/client/src/runtime/skills/index.ts` describing the daemon-scanner ↔ agent-runtime `$HOME` agreement and the failure mode if a future deploy breaks it.
- [ ] At minimum, log a warning at daemon start if `$HOME` is unset or differs from `os.homedir()` resolution (defensive sanity check — almost-zero cost).
- [ ] When sandbox isolation actually ships (containers, sandbox-exec, etc.), pick one of:
  - Hot-rescan skills per-session in the agent slot (cheap, defensive, removes the invariant).
  - Pass the discovered skill catalog to the agent runtime as an explicit allowlist instead of letting it re-read disk.
  - Server-side validation that PATCH-uploaded skills actually exist in the agent's runtime view (requires a runtime probe round-trip).

## Refs

- Original review thread: https://github.com/agent-team-foundation/first-tree/pull/581#pullrequestreview-4363349360 (point #7, second paragraph)
- Spike verifying routing half: https://github.com/agent-team-foundation/first-tree/issues/584#issuecomment-4550740472
- Merged PR: #588

yuezengwu:
> Re-opened from #581 after branch rename `feature/slash-commands` → `feat/slash-commands` (CI's `Check Branch Name` job requires the `feat/` prefix). No code changes since the previous PR's last push (`1ebe0dfa`). Carries over @baixiaohang's approve from the prior thread — see https://github.com/agent-team-foundation/first-tree/pull/581 for the full review history.

## Summary

End-to-end slash-command pipeline: type `/` after `@mention`ing an agent in the composer to pick from that agent's available skills (the same skills Claude Code surfaces via its Skill tool).

- **Data layer**: `agents.skills jsonb NOT NULL DEFAULT '[]'`; new `GET/PATCH /api/v1/agents/:uuid/skills` (visible/manage scope), `SkillDescriptor` Zod schema in `@first-tree/shared`.
- **Daemon**: scans `~/.claude/skills/*/SKILL.md` + `~/.claude/plugins/<plugin>/skills/*/SKILL.md`, parses YAML frontmatter, uploads via PATCH on every daemon start (snapshot semantics; codex/human agents skipped). Failures log + continue — never blocks startup.
- **Web composer**: `useSlashCommand` hook + `SlashCommandPopover` (mirrors `useMentionAutocomplete` shape). `/` triggers only at start of message (composer-mode). Skill list follows the **most recent `@mention` before the cursor**; 1-on-1 chats fall back to the sole speaker so `/` works without `@`. v1 ships one system command (`/clear`); `/help|/me|/invite` deferred pending UX surfaces.

## Test plan

- [x] `pnpm typecheck` — shared / server / client / cli / web all clean
- [x] `pnpm check` — 0 errors
- [x] `pnpm test` — all suites green (shared / server / client / cli / web)
- [x] Live E2E against firsttreehub dev DB: real `discoverClaudeCodeSkills` against `\$HOME/.claude/skills` (42 skills) → real `uploadAgentSkills` PATCH → GET round-trip identity verified
- [x] Browser smoke via `/login` dev-callback as `devuser` against group chat with seeded skills on `reviewer-bot` / `planner-bot`
- [x] **Slash-routing spike** (closes #584 routing half): direct SDK probe confirms `[From: alice]\n\n/review` correctly invokes Skill tool — First Tree's `formatInboundContent` prefix is actually *necessary* for routing (bare `/review` falls through to Bash). See https://github.com/agent-team-foundation/first-tree/issues/584#issuecomment-4550740472.

## Migration / DB changes ⚠️

Adds one column. Non-destructive — PG 11+ does this as a metadata-only ALTER (no table rewrite). `IF NOT EXISTS` for idempotency.

```sql
ALTER TABLE "agents" ADD COLUMN IF NOT EXISTS "skills" jsonb NOT NULL DEFAULT '[]'::jsonb;
```

## Follow-up issues

- #583 — bump `description.max` 2000 → 8000, lower `agentSkillsSchema.max` 2000 → 200, add explicit `bodyLimit` on the PATCH route
- #584 — skill-set sync invariant (routing half resolved by spike; sandbox-isolation half deferred to first real deploy boundary)

### DIFF
--- a/apps/cli/src/__tests__/runtime-provider-reconcile.test.ts
+++ b/apps/cli/src/__tests__/runtime-provider-reconcile.test.ts
@@ -3,7 +3,11 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { parse as parseYaml } from "yaml";
-import { reconcileLocalRuntimeProviders, uploadClientCapabilities } from "../core/runtime-provider-reconcile.js";
+import {
+  reconcileLocalRuntimeProviders,
+  uploadAgentSkills,
+  uploadClientCapabilities,
+} from "../core/runtime-provider-reconcile.js";
 
 /**
  * Two CLI helpers, both fetch-driven against the hub:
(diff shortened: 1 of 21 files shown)


baixiaohang (approved):
Re-approve — supersedes #581 with branch renamed to satisfy CI's branch-name check. Commits bit-for-bit identical (`f8e8773e` + `40011163` + `1ebe0dfa`); zero content delta. Original review notes and follow-ups #583 / #584 still apply.

baixiaohang (approved):
Carries over the #581 approval. Delta vs the previously-approved commit is a single line in `packages/server/drizzle/0050_agent_skills.sql` — `ADD COLUMN` → `ADD COLUMN IF NOT EXISTS` (commit `1ebe0dfa`). Idempotent migration is the right call given drizzle's `when`-based replay semantics, and it matches the prevailing pattern of recent additive migrations on main (`0048_github_entity_state.sql` et al.). Routing spike on #584 closed the load-bearing assumption (`[From: ...]` prefix is actually required for `/<skill>` to route to the Skill tool).

Ready to ship — merging is yuezengwu's / 白晓航's call. Follow-ups #583 and #584 (sandbox-isolation half) remain tracked.

PR #578, opened 2026-05-26
refactor: remove chat-internal ask-user surface (NHA M0 prep)

yuezengwu:
## Summary

This is the **M0 prep cleanup** for the upcoming Need-Human-Attention (NHA) work. The NHA primitive replaces the existing chat-internal `AskUserQuestion` bridge with a new structured surface; M0 strips the old one so the two never coexist.

- **Client SDK** no longer bridges Claude Agent SDK's `AskUserQuestion` tool to Hub messages. `canUseTool` now denies it with a redirect message pointing at the upcoming `first-tree attention` CLI / NHA skill. `ask-user-bridge.ts` is removed; SessionManager / agent-io drop the `question_answer` dispatch + render branches.
- **Server** drops the `format=question` write path: the `POST /:chatId/questions/:correlationId/answer` route, the `assertSenderMayEmitQuestion` / `recordPendingQuestionFromMessage` calls in `services/message.ts`, and the `submitAnswer` / write-path helpers in `services/questions.ts`. The remaining file keeps only the two supersede hooks (`markSupersededByChat`, `markSupersededByAgents`) still called from chat-archive and client-claim.
- **Shared** drops the `question` / `question_answer` message-format enum entries and the SDK-bridge content schemas. Only `questionStatusSchema` + `QUESTION_STATUSES` survive — they type the unchanged `pending_questions.status` column.
- **Web** drops the in-chat `QuestionMessage` / `OptionCard` renderers and `api/questions.ts`. chat-view falls through to the existing JSON pre block for any historical `format=question` rows.

## Intentionally kept for NHA M1 末 reuse

- the `pending_questions` table (no migration in this PR)
- `derivePendingQuestions` + `pendingQuestionAgentIds` on `me-chat` rows
- the `needs_you` main status + `MAIN_STATUS_PRIORITY` ordering
- the chat-list amber `?` badge, right-sidebar "Needs reply" pill, compose-status-bar Reply jump
- the supersede transitions on chat-archive / client-claim

The new attention primitive will repoint the same chat-list signal to its backing store without touching any of the rendering machinery.

## Diff shape

26 files changed, +90 / -4428.

## Test plan

- [x] `pnpm typecheck` — 13/13 tasks green
- [x] `pnpm check` — clean (16 pre-existing non-null-assertion warnings)
- [ ] Verify in staging that historical `format=question` rows in the messages table render via the JSON-pre fallback without breaking chat-view
[...]

PR #593, opened 2026-05-27
refactor(needs-attention): collapse to three rules + migrate to attentions table

Gandy2025:
## Summary

Replaces the four-rule Needs-attention predicate with a three-rule one that only pins chats that actually point at the caller, and migrates the data source from the dormant `pending_questions` table to NHA's `attentions`.

### Rules (locked with @gandy2025)

```
chat ∈ Needs Attention ⇔ any of:
  R1. agent ∈ chat has main = 'failed'  AND  agent.manager = caller
  R2. ∃ message in unread window with caller_human_id ∈ message.metadata.mentions
  R3. ∃ attention in chat with state = 'open' AND target_human_id = caller_human_id
```

Sort ladder unchanged: `failed > needs_you > mention`.

### Why three rules

- The old R3 (any agent has a pending question + I'm a speaker → pin) turned every group chat with any open question into noise for every speaker, even when the question was clearly addressed to someone else.
- The old R4 (unreadMentionCount > 0 → pin) rode on the v1 1-on-1 implicit auto-mention (`services/message.ts:282 dmAutoProjection`), so every plain `"ack"` from an agent in a DM pinned the chat — but that signal was only meant as the red dot, not a top-of-list anchor (the original user complaint).

Both reduce to: a chat should pin only when *something points at me*. The three new rules each encode that explicitly.

### Why migrate to `attentions`

`pending_questions` has no production writer (PR #578 removed the `format=question` write path) and has no `target` column — the old R2/R3 couldn't actually tell who a question was for. NHA M1's `attentions.target_human_id` is the explicit signal, hit by the indexed `(target_human_id, state)` / `(origin_chat_id, state)` lookups.

### Wire compat

Old `pendingQuestionAgentIds: string[]` and `chatHasOpenQuestion: boolean` keep being emitted as `[]` / `false`, marked `@deprecated`. A follow-up PR drops them after a release of soak. New booleans are checked with strict `=== true` on the front-end so an `undefined` from an old server silently degrades R2/R3 to off (under-pin, never over-pin).

### v1 red-dot preserved

`dmAutoProjection` is unchanged. `unread_mention_count` still bumps on every DM and still drives the red dot + bold title. Only the bucket predicate decouples from the raw counter; R2 reads `chatHasExplicitMentionToMe` instead. `direct-chat-auto-mention.test.ts` passes without changes.

### Lifecycle

The two callers of the now-dead `markSupersededBy{Chat,Agents}` (`session.ts:396` chat evict, `client.ts:185` client claim) are rewired to two new helpers in `services/attention.ts`: `closeOpenAttentionsByChat` / `closeOpenAttentionsByAgents`. Same return shape so post-commit `notifyChatMessage` fan-out keeps working.

### Deleted dead code

- `packages/server/src/db/schema/pending-questions.ts` (schema only; DB table itself stays for one release — follow-up migration drops it after soak)
- `packages/server/src/services/questions.ts` (lifecycle helpers no longer reachable)
- `packages/server/scripts/m0-supersede-historical-pending-questions.mjs` (one-shot NHA M0 cleanup)

## Test plan

- [x] `pnpm typecheck` — 13/13 tasks green
- [x] `pnpm check` — clean (16 pre-existing github-scan warnings unrelated)
- [x] `pnpm test` — 1219 server / 383 web tests pass
- [x] **User scenario coverage** (14 new server cases + 27 web unit cases):
  - R1: mine-failed × {speaker, watcher, manager-narrowing}
  - R2: 1v1 plain (**user's t7 痛点 — does NOT pin**), 1v1 explicit, group explicit, group `@other`, mark-read clears
  - R3: 1v1 raise, group target=me, group target=other, notification closed-on-creation
  - composite + deprecation guards
  - 3 front-end skew tests (old server + new web) verify R2/R3 silently degrade
- [x] v1 red-dot regression — `direct-chat-auto-mention.test.ts` 5/5 passes untouched

## Open follow-ups

- A migration to `DROP TABLE pending_questions` once a release of soak passes.
- Drop deprecated `pendingQuestionAgentIds` / `chatHasOpenQuestion` wire fields after one release.
- Possibly extend R2 to cover "human watcher explicitly @-mentioned" — out of scope here.


yuezengwu:
Hi @Gandy2025 — 想跟你同步一下：NHA（Need Human Attention）这套基础设施目前还在持续推进中，`attentions` 表、服务层、wire 协议都还会有进一步演进。

这个 PR 把 Needs-attention 的数据源切到正在迭代的 `attentions` 上，相当于在动中的子系统上叠加改动，容易和 NHA 后续工作产生 merge / 语义冲突，所以先关掉。**在 NHA 主轴落定之前，请暂时不要在 `attentions` 表 / `services/attention.ts` / 相关 wire schema 上做改动**，避免互相打架。

需要强调一下：代码改动本身 review 过质量很高 —— 三条规则的收敛思路、wire 双向兼容、closeOpenAttentions 的事务边界、索引选择、测试覆盖都做得很到位。**不是代码问题关**。等 NHA 主线稳定后欢迎基于新基线重开或重做。

Cc @yuezengwu

PR #577, opened 2026-05-26
feat(web): responsive workspace layout for mobile viewports

yuezengwu:
## Summary

Makes the web dashboard usable down to ~375px logical. Adds a three-bucket viewport hook (`xl` ≥80rem / `md` 48–80rem / `narrow` <48rem) and rewires `Layout`, `WorkspacePage`, `ChatView`, and `NewChatDraft` so each viewport gets sensible chrome.

### Behavior by breakpoint

| Surface | xl (≥1280) | md (768–1279) | narrow (<768) |
|---|---|---|---|
| Top bar | brand + tabs + cmdk/theme/user | brand + tabs | tabs only |
| Conversation list | inline rail | inline rail | full-bleed view (no chat) / scrim+drawer summoned by hamburger (chat or draft selected) |
| Chat right details | inline column | inline column | scrim+drawer (capped `min(88vw, 20rem)`) |
| Composer | normal | normal | adds `env(safe-area-inset-bottom)` so the iOS home indicator doesn't overlap the send button |
| `hearback-widget` FAB | shown | shown | hidden via CSS (would overlap composer) |

Layout uses `min-height: 100dvh` over `100vh` so the chrome syncs with iOS Safari's URL-bar resize.

### Overlay anchoring (intentional difference)

The two narrow-viewport overlays have different scrim anchors on purpose:

- **Conversation list overlay** is anchored to `WorkspacePage` — the scrim covers the chat area only, the top tab bar stays interactive so the user can still switch sections.
- **Chat-right-details overlay** is anchored to `ChatView`'s inner row — the scrim covers the timeline + composer only, the chat header (title, status, controls) stays visible above it.

Both overlays accept Esc and scrim click to dismiss; the conversation-list one is closed automatically when the viewport widens back to `md`/`xl`.

### Files

- `packages/web/src/hooks/use-viewport.ts` *(new)* — `useWorkspaceViewport()` driven by two `matchMedia` queries
- `packages/web/src/components/layout.tsx` — top-bar progressive collapse + dvh fallback
- `packages/web/src/pages/workspace/index.tsx` — narrow + no chat → list as main view; otherwise scrim+drawer summon (Esc + scrim click both dismiss)
- `packages/web/src/pages/workspace/center/chat-view.tsx` — right-rail overlay branch + composer safe-area + summon hamburger
- `packages/web/src/pages/workspace/center/index.tsx` + `center/chat-by-id.tsx` — prop pass-through for `narrow` / `onShowConversations`
[...]

PR #596, opened 2026-05-27
feat(web): responsive Settings shell + Team table for narrow viewports

yuezengwu:
## Summary

Phase-2 of the mobile dashboard rewire — picks up where #577 left off. PR #577 covered the workspace surface (chat list / chat view / right rail / top bar); this PR lands the two remaining high-traffic admin pages so the dashboard is usable end-to-end on a phone.

### Behavior by breakpoint

| Surface                 | xl/md (≥48rem)                          | narrow (<48rem)                                                                           |
| ----------------------- | --------------------------------------- | ----------------------------------------------------------------------------------------- |
| Settings sub-nav        | 200px sidebar + 1160 canvas (unchanged) | horizontal pill row above `<Outlet/>`, non-sticky, x-scrolls if items overflow            |
| Team roster table       | 7 columns (Name · Delegate · Manager · Runs on · Status · Created · ⋯) | 3 columns: Name + Status + ⋯. Manager / Runs on / Created collapse into a `·`-joined caption under the handle |
| Humans row              | Delegate column has chip + edit         | Delegate folds into Name caption; "Set delegate →" inline CTA preserved for self/admin     |

### Why a *non-sticky* pill row for Settings

Settings is a low-frequency switch scenario (users come here to do one thing, not bounce between sub-pages). The Layout top nav is already permanent; pinning a second nav row would steal another ~36px from a phone screen that already lost the brand cluster. Letting the pill row scroll away gives the sub-page its full vertical space — and tapping the global "Settings" tab again returns to the top anyway.

### Why keep Status as its own column at narrow

PresenceChip is the single most-scanned signal in the roster ("is this agent live?"). The pill's width is hostile to wrapping inside a caption line, and scanning down a fixed column of chips is much faster than scanning a free-text caption. Manager / Runs-on / Created are reference data — fine to fold; Status isn't.

### Files

- `packages/web/src/pages/settings.tsx` — `useWorkspaceViewport()` gate; new `PillLink` for narrow; existing sidebar branch untouched
- `packages/web/src/pages/team/team-table.tsx` — `COLUMNS_WIDE` / `COLUMNS_NARROW` split; `HumanRowView` / `AgentRowView` get an `isNarrow` branch; `NameCell` gains optional `extra` slot for the caption fold-in; new `HumanDelegateInline` for the narrow Set-delegate CTA

No data structure / schema / migration / API change. No new hooks. No new dependencies. `pnpm --filter @first-tree/web typecheck` clean (incl. design-token lint); `pnpm --filter @first-tree/web test` 406/406 pass.

### Self-review (P2 follow-ups intentionally not in this PR)

- No RTL render test for the narrow path yet — `team-groups.test.ts` only covers `buildGroups` pure logic. Adding one needs a `useWorkspaceViewport` mock; happy to file an issue or follow-up if you want it tracked.
- `useWorkspaceViewport` is no longer "workspace-only"; rename to `useResponsiveViewport` is a worthwhile but cross-cutting follow-up.

## Test plan

- [ ] Desktop (xl ≥80rem): Settings shows sidebar + 1160 canvas; Team table shows all 7 columns — unchanged from `main`
- [ ] md (48–80rem): same as xl for these surfaces
- [ ] narrow (<48rem) Settings: pill row visible at top of `/settings/*`, scrolls horizontally when items don't fit, scrolls away with content (not sticky)
- [ ] narrow (<48rem) Team: only Name + Status + ⋯ visible; Agent rows show `manager (you) · runtime @ host · 5/12` caption; Human rows show `Delegate: @x` (or `Set delegate →` for self/admin) caption
- [ ] narrow Team: clicking an Agent row still navigates to `/agents/<id>`; "Set delegate →" inside a Human row's caption opens the delegate dialog and does NOT propagate to row click
- [ ] narrow Team: collapsible "Other members' private agents" section (admin-only) toggles correctly with chevron
- [ ] Breakpoint flip (resize across 48rem boundary): UI hot-reflows cleanly; section-collapsed state preserved across flips
- [ ] `pnpm --filter @first-tree/web typecheck` and `pnpm --filter @first-tree/web test` green in CI

### DIFF
--- a/packages/web/src/pages/settings.tsx
+++ b/packages/web/src/pages/settings.tsx
@@ -1,5 +1,6 @@
 import { NavLink, Outlet } from "react-router";
 import { useAuth } from "../auth/auth-context.js";
+import { useWorkspaceViewport } from "../hooks/use-viewport.js";
 import { cn } from "../lib/utils.js";
 
 /**
@@ -20,6 +21,11 @@ import { cn } from "../lib/utils.js";
  * (sidebar 200 + main 960 = 1160 total).
  * Layout opts out of the default 960 wrapper via `isSettings` so this whole
  * shell can centre itself at ~1180.
+ *
+ * Narrow viewport (<48rem): the sidebar would steal half the screen,
+ * so it collapses into a horizontally-scrollable pill bar above the
+ * `<Outlet/>`. Same NavLink semantics (active state via `isActive`), same
+ * route targets — only the chrome shape changes.
  *
  * Sub-routes:
  *   Computers     — user-scoped: machines connected to Hub (most-frequent
@@ -47,6 +53,7 @@ const ITEMS: Item[] = [
 
 export function SettingsLayout() {
   const { role, onboardingCompletedAt, meLoaded } = useAuth();
+  const viewport = useWorkspaceViewport();
   // Wait for `/me` to resolve before rendering the nav — otherwise a fresh
   // direct hit on /settings/github would briefly paint the member-view nav
   // (no GitHub) before `role` flips to "admin".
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
Scope is tightly bounded (2 files, narrow path is additive, wide path untouched → zero desktop regression risk). Design rationale well-documented in PR description and inline comments. Reuses existing `useWorkspaceViewport` + design tokens, no new breakpoints/deps. CI green.

Non-blocking follow-ups (no need to address in this PR):
- RTL render test for the narrow path (already self-flagged).
- `useWorkspaceViewport` → `useResponsiveViewport` rename (cross-cutting, separate PR).
- `HumanDelegateInline`'s `e.stopPropagation()` is defensive — narrow `HumanRowView` has no row-click handler, so it's harmless but unnecessary.
- Pill row horizontal scroll has no fade/indicator at the narrowest widths — minor UX nit.

LGTM 🚀

PR #597, opened 2026-05-27
feat(web): inline @mention chip highlight + paste-safe popover

yuezengwu:
## Why

The chat composer's `@mention` validation was overly strict in a way that
broke copy-paste flow: a pasted snippet containing `@something` could (a)
silently disable the send button without explaining which `@`s counted,
and (b) hijack the user's Enter keystroke because the autocomplete
popover treated *any* cursor-adjacent `@` as if the user had just typed
it.

## What

Two changes sharing one source of truth (`segmentMentions` in
`@first-tree/shared`):

1. **Inline chip highlight for resolved `@<participant>` tokens**, both
   in the composer and in sent messages.
   - Composer: a `position: absolute` mirror layer (`MentionHighlightOverlay`)
     paints chips behind the textarea; the textarea itself uses
     `color: transparent; caret-color: var(--fg)` so the caret stays
     visible.
   - Sent messages: a rehype plugin (`rehypeMentions`) walks the hast
     tree, splits text nodes on `MENTION_REGEX`, and wraps resolved
     tokens in `span.mention-chip`. Code regions and link text are
     skipped.
   - Unresolved `@tokens` (typos, `@TODO`, `@types/node`, outsiders)
     stay plain text — the chip / no-chip diff is now the validity
     signal, removing the need for differentiated tooltip copy.

2. **Paste-safe popover keyboard hijack.** `useMentionAutocomplete` now
   takes a required `interactiveTriggerIndex: number | null`. The
   keyboard hijack (Enter / Tab / Arrows / Escape) only fires when the
   active trigger's `@` position matches an index the caller has flagged
   as "user-typed". Two callers updated:
   - `chat-view.tsx` — tracks keystroke `@`, the explicit `@` toolbar
     button, and the focus-prime stamp.
   - `new-chat-draft.tsx` — same pattern. The popover stays visible on
     a paste-introduced `@` (so click-to-pick still works), but Enter
     falls through to `handleSend`.

## What's NOT changed

- Group-chat send guard (`requiresMention && draftMentions.length === 0`)
  is intact.
- Server-side `enforceGroupMention` is untouched.
- `MENTION_REGEX` and `extractMentions` are unchanged — `segmentMentions`
  reuses the same regex so composer / message / server resolution can't
  drift.

## Test plan

- [x] `pnpm --filter @first-tree/shared test` — 11 new tests for `segmentMentions` (empty, no participants, multiple mentions, case preservation, npm scope / email negatives, offset-preservation contract for the mirror overlay)
- [x] `pnpm --filter @first-tree/web typecheck && pnpm --filter @first-tree/web test`
- [x] `pnpm --filter @first-tree/server test mentions` (sanity-check that shared changes don't regress server-side resolution)
- [x] Biome on touched files: clean
- [ ] Manual: paste a multi-line block containing `@foo` (non-member) into a group chat — Enter sends; send button disabled with single-line tooltip
- [ ] Manual: type `@al` — popover opens, Enter picks; the inserted `@alice` renders as a green chip immediately
- [ ] Manual: send a message containing `@alice` — the sent bubble shows the same chip
- [ ] Manual: a message containing `` `@param` `` inside an inline code span renders as code, not as a chip

### DIFF
--- a/packages/shared/src/__tests__/mentions.test.ts
+++ b/packages/shared/src/__tests__/mentions.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { extractMentions, scanMentionTokens } from "../mentions.js";
+import { extractMentions, scanMentionTokens, segmentMentions } from "../mentions.js";
 import type { ChatParticipantDetail } from "../schemas/chat.js";
 
 /**
@@ -129,5 +129,74 @@ describe("scanMentionTokens", () => {
     // mention token of `@agent-team-foundation/...`. Pin both forms.
     expect(scanMentionTokens("npm i first-tree-shared")).toEqual([]);
     expect(scanMentionTokens("@alice/foo and @bob/bar")).toEqual([]);
+  });
+});
+
+describe("segmentMentions", () => {
+  const participants = [
+    mkParticipant("alice", "agent-alice"),
+    mkParticipant("bob", "agent-bob"),
+    mkParticipant("charlie-07", "agent-charlie"),
+  ];
+
+  it("returns a single text segment when there are no mentions", () => {
+    expect(segmentMentions("hello world", participants)).toEqual([{ kind: "text", value: "hello world" }]);
+  });
+
+  it("returns an empty array for empty content", () => {
+    expect(segmentMentions("", participants)).toEqual([]);
+  });
+
+  it("splits a body with one mention into text-mention-text", () => {
+    expect(segmentMentions("hey @alice please look", participants)).toEqual([
+      { kind: "text", value: "hey " },
+      { kind: "mention", value: "@alice", name: "alice", agentId: "agent-alice" },
+      { kind: "text", value: " please look" },
+    ]);
+  });
+
+  it("preserves original case in the mention value (no lowercase rewriting)", () => {
+    expect(segmentMentions("@ALICE", participants)).toEqual([
+      { kind: "mention", value: "@ALICE", name: "alice", agentId: "agent-alice" },
+    ]);
+  });
+
+  it("keeps unresolved @tokens inside the surrounding text segment", () => {
+    expect(segmentMentions("@ghost and @alice", participants)).toEqual([
+      { kind: "text", value: "@ghost and " },
+      { kind: "mention", value: "@alice", name: "alice", agentId: "agent-alice" },
+    ]);
+  });
+
+  it("does not split inside an email address", () => {
+    expect(segmentMentions("ping alice@example.com", participants)).toEqual([
+      { kind: "text", value: "ping alice@example.com" },
+    ]);
+  });
+
+  it("does not split on npm scoped package names", () => {
+    expect(segmentMentions("install @alice/foo today", participants)).toEqual([
+      { kind: "text", value: "install @alice/foo today" },
+    ]);
+  });
+
+  it("handles back-to-back mentions with no text between", () => {
+    expect(segmentMentions("@alice @bob hi", participants)).toEqual([
+      { kind: "mention", value: "@alice", name: "alice", agentId: "agent-alice" },
+      { kind: "text", value: " " },
+      { kind: "mention", value: "@bob", name: "bob", agentId: "agent-bob" },
+      { kind: "text", value: " hi" },
+    ]);
+  });
+
+  it("preserves character offsets — concatenated values rebuild the source", () => {
+    // This is the contract the composer mirror overlay relies on: the
+    // overlay text must equal the textarea text byte-for-byte so caret /
+    // selection alignment stays correct.
+    const source = "  hey @alice — and @charlie-07!  ";
+    const rebuilt = segmentMentions(source, participants)
+      .map((s) => s.value)
+      .join("");
+    expect(rebuilt).toBe(source);
   });
 });
--- a/packages/shared/src/__tests__/mentions.test.ts
+++ b/packages/shared/src/__tests__/mentions.test.ts
@@ -199,4 +199,35 @@ describe("segmentMentions", () => {
       .join("");
     expect(rebuilt).toBe(source);
   });
+
+  // The next batch pins the agreement with `extractMentions` on code
+  // regions — without this, the composer paints chips for tokens the
+  // server's resolver would drop, and the user sees "chip + send
+  // disabled" with no explanation (PR 597 review #1).
+  it("does not chip @tokens inside inline backtick spans", () => {
+    expect(segmentMentions("use `@alice` literally", participants)).toEqual([
+      { kind: "text", value: "use `@alice` literally" },
+    ]);
+  });
+
+  it("does not chip @tokens inside fenced ``` blocks", () => {
+    const source = "```\n@alice in code\n```\nhey @bob";
+    expect(segmentMentions(source, participants)).toEqual([
+      { kind: "text", value: "```\n@alice in code\n```\nhey " },
+      { kind: "mention", value: "@bob", name: "bob", agentId: "agent-bob" },
+    ]);
+  });
+
+  it("does not chip @tokens inside tilde ~~~ blocks", () => {
+    const source = "~~~\n@alice example\n~~~";
+    expect(segmentMentions(source, participants)).toEqual([{ kind: "text", value: source }]);
+  });
+
+  it("byte-for-byte invariant holds even when code regions suppress mentions", () => {
+    const source = "before `@alice` middle @bob ```\n@charlie-07\n``` after @alice";
+    const rebuilt = segmentMentions(source, participants)
+      .map((s) => s.value)
+      .join("");
+    expect(rebuilt).toBe(source);
+  });
 });
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -9,7 +9,15 @@ export {
   normalizeDocLinkPath,
   parseWorkspaceDocKey,
 } from "./lib/doc-path.js";
-export { extractMentions, MENTION_REGEX, type MentionParticipant, scanMentionTokens, stripCode } from "./mentions.js";
+export {
+  extractMentions,
+  MENTION_REGEX,
+  type MentionParticipant,
+  type MentionSegment,
+  scanMentionTokens,
+  segmentMentions,
+  stripCode,
+} from "./mentions.js";
 // -- OAuth-callback open-redirect guard --
 export { DEFAULT_SAFE_REDIRECT, safeRedirectPath } from "./safe-redirect.js";
 export {
--- a/packages/shared/src/mentions.ts
+++ b/packages/shared/src/mentions.ts
@@ -96,6 +96,52 @@ export type MentionSegment =
   | { kind: "mention"; value: string; name: string; agentId: string };
 
 /**
+ * Collect the byte ranges of Markdown code regions (fenced ```, fenced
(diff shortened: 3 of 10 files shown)


baixiaohang (commented):
Review (non-approving). Overall direction looks solid — `segmentMentions` is a clean single source of truth, the `interactiveTriggerIndex` design correctly separates "popover visible" from "popover hijacks keys", and the mirror-overlay docstrings spell out the trickier invariants (font / padding / scroll / zIndex / trailing-newline padding). Tests on the shared layer are thorough.

A few things worth a follow-up before this lands. None of them are blockers on their own, but the first one is a real UX regression you'll hit immediately if a user pastes code-with-`@` into a group chat:

### 1. ⚠️ Composer overlay disagrees with `extractMentions` on code regions

`extractMentions` strips fenced + inline code via `stripCode` before scanning (mentions.ts:57-58). `segmentMentions` intentionally does NOT (mentions.ts:131-134 in the new code).

Concrete failure mode in a group chat:

1. User types or pastes `` `@alice` `` (inline code) or a fenced block containing `@alice`.
2. `MentionHighlightOverlay` paints an `@alice` chip in the composer — visual signal "alice is the recipient".
3. `draftMentions = extractMentions(...)` returns `[]` because `stripCode` removed the code region.
4. `requiresMention && draftMentions.length === 0` → send button disabled + tooltip "Group chats need at least one @member to send".

The user sees a chip and is told the message has no @member. That breaks the "chip / no-chip is the validity signal" contract that's called out in the PR description.

Suggested fix: in `segmentMentions`, run `stripCode` and walk only the surviving offsets (you'll need to track an offset map so the segments still reconstruct the original buffer for the overlay). Alternatively give the function an opt-in `stripCode: boolean` flag and pass `true` from the composer path / `false` from places that need byte-for-byte segments. The rehype path already gets this for free because markdown lowers code regions into `<code>` / `<pre>` and `SKIP_TAGS` catches them.

### 2. ⚠️ `NewChatDraft` got the paste-safe popover fix but no chip overlay

`new-chat-draft.tsx` wires `interactiveTriggerIndex` just like `chat-view.tsx`, but its textarea has no `MentionHighlightOverlay`. Is that intentional? PR description reads as if both composers should look the same; if it's intentional ("new-chat draft is participant-selection territory, not message territory") it'd help to drop a one-line comment so the next reader doesn't think it was forgotten.

### 3. ⚠️ Selection over chips

With `color: transparent` on the textarea and `::selection { background: var(--accent-bg); color: var(--fg) }` from index.css:1121, selecting across a chip yields: textarea selection band (`--accent-bg`) painted on top of the overlay chip (`--accent` bg + `--fg-on-vivid` text), and the textarea's transparent glyphs are repainted in `--fg` by `::selection`. The chip becomes near-unreadable while selected and the text "colour-flips" mid-selection. Worth eyeballing manually on long selections / cross-chip drags / chip-edge drags to make sure it's acceptable; if not, a scoped `::selection` rule on the textarea (transparent background, transparent color) or a `selection-color: transparent` could keep the overlay readable underneath.

### 4. ❓ Misc / nits

- `mention-highlight-overlay.tsx` re-subscribes the `scroll` listener on every keystroke (`value` is in the deps). Functionally fine — but you could lift `sync` into a ref and run the effect once on mount; the `value`-driven re-sync can be done by calling `sync()` directly inside a `useLayoutEffect([value])`. Micro-opt, optional.
- `rehype-mentions.ts` rebuilds the `nameMap` once per render of the message (the plugin factory closes over it), and `messageRehypePlugins` is `useMemo`'d per `mentionParticipants` — good. Just confirming the plugin instance is intentionally per-message and not shared between conversations, since `participants` is the chat's participant set.
- `mergeable_state` is currently `dirty` on this branch — needs a rebase on main before merge.
- The 4 manual QA bullets in the PR description are still unchecked. Worth running them, especially the paste-with-`@foo` flow and the `` `@param` `` inside inline code case (which is also #1 above).

### What passed cleanly

- ✅ `segmentMentions` API + 11 new tests (case preservation, scoped npm, email, back-to-back mentions, offset-rebuild contract — the last one is exactly the invariant the overlay depends on)
- ✅ `rehypeMentions` correctly skips `<code>`, `<pre>`, `<a>`
- ✅ `useMentionAutocomplete` change is backwards-additive for callers and the `interactive` gate is in the right place (`handleKey` early-return)
- ✅ Existing group-chat send guard (`requiresMention && draftMentions.length === 0`) and server `enforceGroupMention` untouched
- ✅ Tooltip copy: "Group chats need at least one @member to send" reads cleaner than the old line

Leaving as `COMMENT` so the code-region inconsistency can be discussed before approval. Happy to re-review once #1 + rebase are sorted.

yuezengwu:
Thanks for the close read @baixiaohang — pushed `c58cbf6b` addressing #1, #2, #3 and rebased on `main` to clear the conflict. Detail:

### #1 — code-region disagreement (the real bug)

Fixed in `segmentMentions` itself, not via an opt-in flag. New internal `codeRanges` helper computes byte-offset skip ranges from the original content (fenced ` ``` `, fenced `~~~`, inline `` ` ``), merges overlaps, and `segmentMentions` skips MENTION_REGEX matches whose `@` falls inside a skip range. Offsets are preserved — the byte-for-byte rebuild invariant the mirror overlay depends on is unchanged (new test pins it for code-mixed inputs).

I picked "always honor code regions" over the opt-in flag because:
- composer needs this to match `extractMentions` (your scenario)
- rehype path already gets it for free via `SKIP_TAGS` on `<code>` / `<pre>` and never calls `segmentMentions`
- no current third caller, so a flag would only be configuring something nothing else uses

4 new tests pin inline / fenced / tilde / offset-preservation under code regions (31 total in `segmentMentions`).

### #2 — `NewChatDraft` no overlay

Intentional. Added the explanatory comment next to `interactiveTriggerIndex` in `new-chat-draft.tsx`: the chip row above the textarea is already the canonical "who is in the room" surface, a typed `@<name>` promotes the agent to that chip row, so a second in-textarea chip would duplicate the signal. The paste-safe popover fix is still applied because that's a generic textarea UX bug, not tied to whether chips render.

### #3 — selection over chips

Added a `.mention-composer-textarea` class to the composer textarea and two scoped `::selection` rules in `index.css`:
- `.mention-composer-textarea::selection { background: var(--accent-ring); color: transparent; }` — keeps the global `color: var(--fg)` from repainting the textarea's transparent glyphs on top of the overlay during selection
- `.mention-chip::selection { background: var(--accent); color: var(--fg-on-vivid); }` — keeps the chip's own colours through a selection band so it stays readable

### #4 nits

- **Scroll re-subscribe**: agree, micro-opt, deferred. Documented the `value` dep as the sentinel that drives post-edit re-sync.
- **Plugin per-message**: confirmed intentional — `rehypeMentions` factory closes over the chat's `mentionParticipants`, which `messageRehypePlugins` memoises per-chat. Different conversations have different participant sets, so reusing one instance across conversations would either resolve against the wrong set or need a re-instantiation anyway.
- **Rebase**: done (force-pushed on `c58cbf6b`). `mergeable: MERGEABLE` now; CI is queued.
- **Manual QA**: I can't drive the browser from this CLI, so the 4 QA checkboxes still need a human pass. The shared-layer tests now cover the code-region + offset cases that were the riskiest path; the rest (mirror alignment, popover-doesn't-hijack-on-paste, selection band over chip) needs eyeballing in dev.

Ready for re-review.

baixiaohang (commented):
Followed `c58cbf6b` — looked at the diff against the previous review. Summary of how my points landed:

### #1 — composer ↔ extractMentions code-region agreement: ✅ fixed

`codeRanges` + `isInCodeRange` is the right shape (offset-preserving skip set rather than `stripCode`'s in-place erase) and the merge step makes the three patterns commutative with `stripCode`'s sequential `replace` passes. Quick equivalence check on the corner cases I had in mind:

- `` `unclosed @alice `` (no closing backtick) — both helpers leave it; `MENTION_REGEX` matches and chip is emitted ✓
- `` ``` ```\n`@foo`\n``` `` (inline span inside a fenced block) — fenced range covers the inline range; merge collapses them; `@foo` suppressed ✓
- `` `@alice\n` `` (newline inside inline span) — inline pattern rejects `\n`; `@alice` stays a chip; matches `stripCode` ✓

The new "byte-for-byte invariant holds even when code regions suppress mentions" test pins exactly the property the mirror overlay depends on. Picking always-honor over an opt-in flag is the right call given there's only one caller path that wanted byte fidelity and that path also wants code-region suppression.

### #2 — `NewChatDraft` no overlay: ✅ explanation accepted

The "chip row above the textarea is already the canonical 'who is in the room' surface" reasoning is sound, and the comment in `new-chat-draft.tsx` is now load-bearing for the next reader. No code change needed.

### #3 — selection over chips: ✅ improved, with one caveat

`.mention-composer-textarea::selection { color: transparent }` is the key fix — it stops the global rule from repainting the textarea's transparent glyphs in `--fg` during selection, which was the visible-double-text glitch I was worried about.

`.mention-chip::selection` is the right intent, but it may not actually fire: the user's selection range is inside the textarea DOM, and the overlay (`pointer-events: none`, absolute, not part of the selection range) doesn't get `::selection` applied. That said, the effective result is still fine because the textarea's selection band is now `--accent-ring` (30% alpha) instead of `--accent-bg` (12% alpha) on a chip that's `--accent` solid — the band tints the chip darker but doesn't obscure it, and the chip's foreground glyphs in the overlay layer underneath remain readable. Still worth eyeballing on `c58cbf6b` in dev to confirm the cross-chip drag case.

### #4 — nits: ✅ noted

Scroll re-subscribe doc tweak landed, plugin-per-message confirmed intentional, rebased on main, conflict cleared (`mergeable: true`, `mergeable_state: blocked` is just branch-protection waiting for approval).

### Remaining

Only the 4 manual QA bullets in the PR description. The shared-layer tests now cover the code-region / offset-preservation paths (the riskiest bits), so what's left for eyeballs is mirror alignment, paste-doesn't-hijack-Enter, sent-bubble chip rendering, and inline-code-stays-code. Will defer to a human pass on those — happy with the code review side.

yuezengwu:
Good catch on #3, @baixiaohang — you're right that `.mention-chip::selection` is dead. The chip lives inside the overlay div (`pointer-events: none`, absolute-positioned), so it's never part of the textarea's selection range and the rule never fires. Pushed `b8619ae8` removing it, and updated the comment on the surviving `.mention-composer-textarea::selection` rule to record why the chip-scoped one isn't needed (so the next reader doesn't try to "fix" selection styling on the chip again).

Leaving the rest as-is. Code review side closed — only the 4 manual QA bullets remain, which need a human pass.

baixiaohang (approved):
LGTM on the code-review side. Three rounds of follow-ups closed every point cleanly:

- #1 `codeRanges` is the right offset-preserving shape, equivalence with `stripCode` holds on the corner cases I had in mind, and the byte-rebuild invariant test pins what the mirror overlay depends on
- #2 `NewChatDraft` comment makes the deliberate-no-overlay decision load-bearing for the next reader
- #3 `.mention-composer-textarea::selection { color: transparent }` is what actually mattered; the dead chip-scoped rule has been removed and its absence now has a comment

Manual QA bullets (paste-with-`@foo`, type-`@al`-Enter-picks, sent bubble chip rendering, inline `` `@param` ``) still need a human pass before this lands — leaving those to @baixiaohang's pre-merge walkthrough. Branch protection will hold the merge anyway.

Nice work @yuezengwu — thanks for the patient back-and-forth.

PR #603, opened 2026-05-27
refactor(computers): 9 UI fixes (per-row actions, runtime label, copy, layout)

Gandy2025:
## Summary

Round of UI polish driven by gandy2025 design feedback. Stacked on top of PR-D1 (#598) — base set to `refactor/computers-minimal-style`. After PR-D1 merges this PR rebases onto main.

| # | Fix |
|---|---|
| 1 | **Offline → inline Reconnect button**, kebab keeps only Disconnect / Retire |
| 2 | **Agent rows show runtime type**: `gandy-developer · Claude Code   ● online` |
| 3 | **Ready card hides "not reported" runtimes** (no more "Install Codex" advert on machines that deliberately don't run it) |
| 4 | **Runtime line drops `(authMethod)` parens** → `Claude Code · v0.2.141 · oauth` |
| 5 | **Unauth hint drops env-var noise + becomes OS-aware**: `Run \`codex login\` on this Mac.` |
| 6 | **AuthExpired adds dimmed "Runtimes · last reported"** block — tells the operator what's at stake before they re-auth |
| 7 | **Setup-incomplete install boxes**: Copy buttons baseline-align via `margin-top: auto` |
| 8 | **Offline + AuthExpired agents** upgraded from compact summary to expanded list (the operator's question is *which* agents are down, not how many) |
| 9 | **Admin: Team computers section collapsed by default** with Show/Hide; team table left-aligns to Section title |

## Bonus: DEV-only `?demo=<key>` mode

Same PR also bundles a DEV-only demo navigator for `/settings/computers`. Open `/settings/computers?demo=auth-expired` (or any of 11 scenarios) in DEV to flip the live page through every pill × sub-variant with the real Settings chrome, without seeding the local DB. The navigator is a bottom-right floating overlay with a dropdown / prev / next / exit + a "What to check" checklist per scenario.

Production builds tree-shake the demo path entirely (Vite folds `import.meta.env.DEV` to false; `findDemoScenario` returns null; `<DemoNavigator>` never renders; the fixtures module is dead code).

## Files

| New |
|---|
| `cards/shared/stale-runtimes.tsx` — `DimmedGroup` + `StaleRuntimeLine` shared by AuthExpired + Offline |
| `clients/demo-navigator.tsx` — floating dev overlay + `useDemoScenarioParam` |
| `clients/dev-fixtures.ts` — 11 `HubClient` + `RuntimeAgent` scenarios |

| Modified |
|---|
[...]

PR #605, opened 2026-05-27
fix: bundle skills/ in CLI tarball + auto-refresh on CLI drift

yuezengwu:
## Summary

Two coupled fixes for "agent homes silently never get `.agents/skills/*`
installed (or never refresh them after `first-tree upgrade`)":

1. **`fix(cli)`**: wire `copy-github-scan-assets.mjs` onto `pnpm build`
   so the published tarball actually contains `dist/skills/*`,
   `dist/assets/dashboard.html`, and the bundled
   `dist/github-scan/{README,VERSION}`. The script already existed but
   was never called — every published release silently omitted them,
   and `first-tree tree integrate` failed with
   "Could not locate bundled `skills/` payloads" on every fresh agent
   home.
2. **`fix(client)`**: add a per-agent-home pin (`.agent/cli-version`)
   for the bundled CLI version that last performed the bootstrap. Both
   handlers (claude-code, codex) treat a mismatch the same way they
   treat Context Tree HEAD drift: short-circuit disabled, full
   bootstrap re-runs, `installFirstTreeIntegration` re-copies the
   shipped skills payload. Without this, `first-tree upgrade` only
   refreshed skills on already-bootstrapped agent homes when the
   Context Tree happened to advance.

## Behaviour

| Trigger | Before | After |
|---|---|---|
| First bootstrap | `installFirstTreeIntegration` exits 1 (no skills payload in published tarball) | Skills installed ✓, pin written |
| Context Tree HEAD changes | Re-runs integrate ✓ | Same |
| `first-tree upgrade` (no Tree change) | Skills stay stale until next Tree change | Re-runs integrate ✓ |
| Integrate fails (transient) | Sentinel still written; next start short-circuits past the retry | Sentinel written, CLI pin NOT written → next start retries |

Same fail-open rule as Context Tree HEAD: `null` on either side of the
comparison means "unknown" and never forces re-bootstrap. Captures the
`installFirstTreeIntegration` return value (already boolean, was being
ignored) and only pins the CLI version on success so transient
shell-out failures self-heal on the next start instead of being masked
by an optimistic pin.

## Test plan

- [x] `pnpm typecheck` — clean (13/13 turbo tasks)
- [x] `pnpm --filter @first-tree/client test` — 553 pass, 3 skipped (+5 new unit tests for the CLI version pin helpers)
- [x] `pnpm build` produces `apps/cli/dist/skills/{first-tree,first-tree-cloud,first-tree-github-scan,first-tree-onboarding,first-tree-sync,first-tree-write,github-scan}/`
- [x] Smoke: `node apps/cli/dist/cli/index.mjs tree integrate ...` in an empty dir installs all 6 skills under `.agents/skills/` + `.claude/skills/` symlinks
- [ ] Reviewer: after merge, verify next staging publish ships `dist/skills/` and `dist/assets/dashboard.html` in the npm tarball
- [ ] Reviewer: confirm a `first-tree upgrade` (or staging auto-update) on a machine with existing agent homes triggers the "Bundled CLI version changed (… → …); re-running bootstrap to refresh skills" log line on next session start

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "dev": "tsx src/cli/index.ts",
-    "build": "tsdown src/cli/index.ts src/index.ts --format esm --no-dts --external @anthropic-ai/claude-agent-sdk --external @openai/codex-sdk",
+    "build": "tsdown src/cli/index.ts src/index.ts --format esm --no-dts --external @anthropic-ai/claude-agent-sdk --external @openai/codex-sdk && node ./scripts/copy-github-scan-assets.mjs",
     "typecheck": "tsc --noEmit",
     "test": "vitest run --passWithNoTests"
   },
--- a/apps/cli/turbo.json
+++ b/apps/cli/turbo.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": ["dist/**"],
+      "inputs": [
+        "$TURBO_DEFAULT$",
+        "../../skills/**",
+        "../../packages/github-scan/assets/**",
+        "../../packages/github-scan/README.md",
+        "../../packages/github-scan/VERSION"
+      ]
+    }
+  }
+}
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -3,15 +3,19 @@ import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node
 import { join } from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
+  BUNDLED_CLI_VERSION_REL,
   bootstrapWorkspace,
   buildChatSystemPrompt,
   CONTEXT_TREE_HEAD_REL,
   type ContextTreeBinding,
   type InstallFirstTreeIntegrationExec,
   installFirstTreeIntegration,
+  readCachedBundledCliVersion,
   readCachedContextTreeHead,
   readContextTreeHead,
+  resolveBundledCliVersion,
   withContextTreeSyncLock,
+  writeBundledCliVersion,
   writeContextTreeHead,
 } from "../runtime/bootstrap.js";
 import type { AgentIdentity } from "../runtime/handler.js";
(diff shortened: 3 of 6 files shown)


baixiaohang (changes requested):
Reviewed by 白晓航的个人助理 (代表 @baixiaohang).

## Scope & rebase (blocking)

This branch was forked before #598 and #603 landed on `main`. The first 4 commits (`4071652`, `112fd50`, `c8457e4`, `fdbf5ef`) are the same `refactor(computers)` work that has *already* merged to main as PRs #598 + #603. They explain both the merge conflict (`mergeable_state: dirty`) and why the diff is +1510/−402 across 21 files instead of the small fix this PR description talks about.

Please rebase onto current `main` (74cad8a) and drop those 4 commits. After rebase the diff should be ~6 files / ~190 lines, matching the PR description exactly:

- `apps/cli/package.json` — build script hookup
- `packages/client/src/runtime/bootstrap.ts` — CLI-version pin helpers
- `packages/client/src/handlers/{claude-code,codex}.ts` — drift detection wiring
- `packages/client/src/__tests__/bootstrap.test.ts` — helper unit tests

The rest of the review below assumes that's done; everything else I'm flagging is on the two fix commits only.

## Fix quality (positive)

The two `fix(*)` commits are well-designed:

- **CLI-version pin is symmetric with `CONTEXT_TREE_HEAD_REL`** — same fail-open rule (`null` on either side → never force re-bootstrap), same `.agent/`-local pin file, same drift-vs-pin pattern in both handlers. Easy to reason about because it mirrors a pattern that's already battle-tested.
- **Only pinning on `integrationOk` is the correct call.** Pinning on a failed integrate would silently mask the gap and the next start would skip the very retry this drift trigger exists to perform. The fact that `installFirstTreeIntegration` was already returning a `boolean` that was being thrown away makes this a clean fix.
- **Test coverage on the helpers is reasonable** — resolve / read / write / null no-op / trim are all there.

## Issues on the fix commits

1. **`resolveBundledCliVersion()` rests on an assumption that needs verification.** It walks up from `import.meta.url` looking for the nearest `package.json` with a `version` field. The comment promises that "inside the published / dev CLI bundle this resolves to the consumer-facing manifest (`first-tree`, `first-tree-staging`, or `first-tree-dev`)" — but that's not asserted anywhere. In the published tarball case (`bootstrap.ts` is inlined into `dist/index.mjs`), `import.meta.url` lives under whatever `node_modules/first-tree[-staging]/dist/` directory the package was installed into, so the walk should hit the consumer manifest first — ✓ that works.

   But for **dev-mode** (`pnpm --filter @first-tree/client dev` or `tsx packages/client/src/runtime/bootstrap.ts`), the walk hits `packages/client/package.json` first — version comes from `@first-tree/client`, which is `private` and **doesn't get bumped on consumer-facing CLI releases**. The drift trigger never fires in dev mode. That's probably fine (dev hot-reloads anyway), but worth either:
   - asserting "we only rely on this in published-tarball mode" in the comment, or
   - probing the bundled `apps/cli/package.json` explicitly (e.g. `process.argv[1]` based) so the contract is "this is the version of the binary the user actually ran".

   At minimum: please confirm you've tested on a real staging tarball that `resolveBundledCliVersion()` returns `first-tree-staging`'s version, not `@first-tree/client`'s. If yes, ignore.

2. **Turbo `inputs` don't cover the external assets the new build step depends on.** `apps/cli`'s build now reads from `../../skills/` and `../../packages/github-scan/{assets,README.md,VERSION}` via `copy-github-scan-assets.mjs`, but `turbo.json` declares `"inputs": ["$TURBO_DEFAULT$"]` which by default only sees files within the package itself. A change to a skill file or to `packages/github-scan/assets/dashboard.html` will be cache-hit on `apps/cli#build` and silently ship the *previous* skills payload — the exact failure mode this PR is fixing, just one layer up.

   Two ways to address:
   - Add explicit `inputs: ["$TURBO_DEFAULT$", "../../skills/**", "../../packages/github-scan/assets/**", "../../packages/github-scan/README.md", "../../packages/github-scan/VERSION"]` to apps/cli's build task overlay, or
   - Make the cli depend on `^build` of `@first-tree/github-scan` (which it already does via workspace dep) AND on a synthetic "skills" package — heavier but mirrors the source-of-truth.

3. **No integration test for the drift path itself.** The unit tests cover the helpers, but there's no end-to-end test that `cliDrifted ⇒ re-runs integrate`, or that a failed integrate leaves the pin unwritten so the next start retries. Given the existing `installFirstTreeIntegration` test infrastructure with `InstallFirstTreeIntegrationExec`, this should be a small addition and would lock in the behavior the second of the two reviewer checklist items asks future humans to verify by hand.

## CI

`no checks reported on the 'fix/skill-bundle-and-autoupgrade' branch` — please make sure CI runs after the rebase. The PR-description Test plan is all manually-verified; CI would catch the kind of "skill payload not in tarball" issue that started this PR.

## Summary

Core fix is solid and well-thought-through. Blocking: rebase to drop the already-merged refactor commits; then please confirm the `resolveBundledCliVersion()` published-tarball assumption (#1) and decide on the turbo inputs question (#2). Item #3 is a nice-to-have.

yuezengwu:
Thanks for the review — addressed all three points:

### 1. Rebased onto `main` (74cad8a)

Used `git rebase --onto origin/main fdbf5ef6 fix/skill-bundle-and-autoupgrade` to skip the four `refactor(computers)` commits that landed via #598/#603. Branch is now exactly the diff the PR description promised:

```
 apps/cli/package.json                           |   2 +-
 apps/cli/turbo.json                             |  17 ++++  (new — see #2 below)
 packages/client/src/__tests__/bootstrap.test.ts | 114 +++++++  (incl. drift contract tests — see #3 below)
 packages/client/src/handlers/claude-code.ts     |  29 ++++-
 packages/client/src/handlers/codex.ts           |  41 ++++--
 packages/client/src/runtime/bootstrap.ts        |  66 ++++++-
```

### 2. `resolveBundledCliVersion()` published-tarball check — confirmed

Smoke-verified against the built dist by simulating the walk from where `bootstrap.ts` lands inside the bundle. tsdown emits the helper into a chunk at `apps/cli/dist/<chunk>.mjs`, so the walk goes `dist/<chunk>.mjs` → `dist/` → `apps/cli/package.json` (`first-tree-dev@0.5.2` in dev). In a published tarball that same walk goes `node_modules/first-tree[-staging]/dist/<chunk>.mjs` → `node_modules/first-tree[-staging]/package.json` (channel-rewritten name + version). The CI publish workflow rewrites `pkg.name`/`pkg.bin` before `pnpm build` runs, so the consumer-facing manifest is what the walk hits.

I also kept the existing self-review note about dev-mode (`tsx`): in that path the walk does hit `@first-tree/client@0.1.2` first. Functionally still drift-detects correctly (still a stable string per binary), but I've tightened the JSDoc to call that out explicitly rather than only document the published case.

### 3. Turbo `inputs` for external assets — added

New `apps/cli/turbo.json` overlay:

```jsonc
{
  "$schema": "https://turbo.build/schema.json",
  "extends": ["//"],
  "tasks": {
    "build": {
      "dependsOn": ["^build"],
      "outputs": ["dist/**"],
      "inputs": [
        "$TURBO_DEFAULT$",
        "../../skills/**",
        "../../packages/github-scan/assets/**",
        "../../packages/github-scan/README.md",
        "../../packages/github-scan/VERSION"
      ]
    }
  }
}
```

Verified cache invalidates: `pnpm build` → all cached; modify `skills/first-tree/SKILL.md` → only `first-tree-dev:build` re-runs (`Cached: 5 cached, 6 total`).

### 4. Drift contract tests — added

Added a `describe("CLI-version pin contract (handler invariants)")` block to `bootstrap.test.ts` with two tests that lock in the exact handler gate the reviewer was worried about:

- **`does not overwrite the existing pin when integrate fails`** — pre-writes a stale pin, drives `installFirstTreeIntegration` with a throwing exec, asserts the pin still has the old value. If a future change drops the `if (ok) writeBundledCliVersion(...)` guard, this fails.
- **`advances the pin to the new version when integrate succeeds`** — same setup with a no-op exec, asserts the pin advances.

Couldn't unit-test `ensureAgentBootstrap` directly without intrusive plumbing through the handler factory (the exec backend isn't injectable through `HandlerFactoryConfig`), so the tests assert the gate composition directly using the same `InstallFirstTreeIntegrationExec` mock the reviewer pointed at. Open to refactoring the handler to inject `exec` through config in a follow-up if you'd prefer the test exercise the closure end-to-end.

### Other

- `pnpm --filter @first-tree/client test` — 614 pass, 3 skipped (+5 helper unit tests +2 drift contract tests)
- `pnpm typecheck` clean, `pnpm build` chain clean (`dist/skills/*` present)
- Force-pushed; CI should pick up the rebase now.

baixiaohang (approved):
v2 addresses all four review items cleanly: rebase to drop the already-merged refactor commits, dev-mode resolve assumption now documented in the helper, turbo `inputs` overlay added for the external assets, and a contract test pair locking in the "don't pin on failure" handler invariant. CI green. LGTM.

PR #606, opened 2026-05-27
fix(client): dedup concurrent Context Tree syncs sharing one clone dir

baixiaohang:
## Summary

- Client startup was serialising N `git pull` round-trips on the shared Context Tree sync lock, stretching boot to ~7s × N (recent staging log: 6 agents → ~40s, each agent spaced ~7s apart).
- `withContextTreeSyncLock` in `packages/client/src/runtime/bootstrap.ts` chained concurrent callers via `prev.then(fn)` instead of sharing the in-flight promise — so every agent's `syncAgentContextTree` queued its own `git pull` even though the resolved clone dir was identical.
- Rewrite the lock as in-flight de-dup: share the settled result while a sync is in flight; clear the slot on settle so the next caller triggers a fresh sync. `syncContextTreeRepo` is idempotent in `(repo, branch, cloneDir)`, so collapsing N concurrent callers to one shared sync is observationally equivalent to running it N times — minus the latency. Rejection now propagates to all in-flight callers and clears the slot, so a transient failure no longer fans out into N sequential retries.

**Expected impact:** 6 agents × ~7s per pull (~40s startup) → single shared pull (~10s).

## Files

- `packages/client/src/runtime/bootstrap.ts` — rewrite `withContextTreeSyncLock` (+ JSDoc explaining chain → dedup), export the helper at module scope (not added to `src/index.ts` barrel, so the public API is unchanged).
- `packages/client/src/__tests__/bootstrap.test.ts` — 4 new unit tests covering the dedup contract: concurrent same-key collapse, distinct-key isolation, post-settle slot clearing, rejection propagation + retry-after-clear.

## Test plan

- [x] `pnpm exec vitest run src/__tests__/bootstrap.test.ts` — 32/32 pass (incl. 4 new)
- [x] `pnpm check` — client package zero warnings
- [x] `pnpm typecheck` — 13/13 pass
- [ ] Operator validation: install on a multi-agent client and confirm startup drops from ~7s × N to a single shared pull window
- [ ] Confirm logs still make sense — see "Follow-up" below

## Notes

**Out of scope for this PR (follow-up observation from reviewer):** after dedup, `syncContextTreeRepo`'s `log("Context Tree updated (pull)")` lines only land on the *first* caller's log channel. Later agents reuse the in-flight promise and so their per-slot logs show the binding "appearing from nowhere". Not a bug — the result is correct — but a future PR can add a `"reused another agent's in-flight pull"` log line in `resolveContextTreeBinding` on dedup-hit to keep every slot's log self-explanatory.

### DIFF
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -6,10 +6,12 @@ import {
   bootstrapWorkspace,
   buildChatSystemPrompt,
   CONTEXT_TREE_HEAD_REL,
+  type ContextTreeBinding,
   type InstallFirstTreeIntegrationExec,
   installFirstTreeIntegration,
   readCachedContextTreeHead,
   readContextTreeHead,
+  withContextTreeSyncLock,
   writeContextTreeHead,
 } from "../runtime/bootstrap.js";
 import type { AgentIdentity } from "../runtime/handler.js";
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM. Focused, correct fix to a real concurrency bug.

**Correctness**
- `syncContextTreeRepo` is idempotent in `(repo, branch, cloneDir)` → collapsing N concurrent callers to one shared sync is observationally equivalent to N sequential pulls.
- Microtask ordering safe: synchronous `set(key, next)` runs before the `.finally` cb microtask, so the slot is always populated before the delete guard fires.
- Rejection path propagates to all in-flight callers and clears the slot — verified by the `propagates rejection ... and clears the slot` test (retry-after-clear assertion is the right thing to pin).
- `=== next` guard is now defensively redundant (no new entry can land while in-flight), but harmless.

**Tests** 4 new tests cover the contract surface well: same-key dedup, distinct-key isolation, post-settle clear, reject + retry. Local `vitest run src/__tests__/bootstrap.test.ts` → 32/32 ✓.

**No DB / no data-structure changes.** No blockers.

**Non-blocker (PR already calls out as follow-up):** after dedup only the first caller's `log("Context Tree updated (pull)")` lands — later agents reuse the in-flight promise and lose that line. Result is correct, just observability. The proposed `"reused another agent's in-flight pull"` line in `resolveContextTreeBinding` on dedup-hit is the right shape — happy to take as a follow-up PR.

PR #618, opened 2026-05-27
feat(handlers): translate codex/claude SDK auth errors into actionable chat hints

liuchao-001:
## Context

A brand-new user — fresh org, fresh agent — hit this on first chat with a codex-backed agent:

> ERROR - SDK
> Your access token could not be refreshed because your refresh token was revoked. Please log out and sign in again.

The user attributed it to First Tree, but the auth state lives in `~/.codex/auth.json`, owned by OpenAI's codex binary. They had an old install where that file was written months ago; OpenAI's auth server eventually marked the refresh token as revoked, but they never knew because their Codex desktop app uses a separate session and kept working.

The chat-timeline error gave no hint that the fix was `codex login` in their terminal. Same scenario applies to claude-code with a stale credential store.

## Change

Translate the SDK auth-failure at the handler boundary, so the chat shows:

> codex auth on this machine looks broken or expired. This is OpenAI's auth state, not First Tree's — please run `codex login` in your terminal to re-authenticate, then retry. Original SDK error: …

…instead of the opaque raw SDK line.

First Tree does **not** manage either runtime's credential store. Boundary stays one-way: detect, point user at the right CLI. No First Tree UI re-login flow, no API key paste field, no settings-page changes.

## Implementation

**New module** `packages/client/src/handlers/auth-error-hint.ts`:

- `isCodexAuthError(message)` — substring keyword match on the canonical refresh-failure phrases the bundled codex Rust binary emits (`refresh token`, `could not be refreshed`, `log out and sign in`, `sign in again`, `log in again`, `Token data is not available`). The codex SDK exposes only `ThreadError = { message: string }` — no typed code — so substring matching is the only option there.
- `isClaudeAuthError(code)` — equality check against `SDKAssistantMessageError === "authentication_failed"`. The Claude SDK ships a typed union (`sdk.d.ts:1785`), so this path is strictly more robust than codex's.
- `formatAuthHint(runtime, originalMessage)` — builds the chat-timeline message, preserves the raw SDK error verbatim at the end.

**Codex handler** (`packages/client/src/handlers/codex.ts`): wraps all four emit sites that surface SDK error messages — `turn.failed`, `ThreadErrorEvent`, the `runStreamed` catch block, and the non-fatal `error` item in `processItem`.

**Claude-code handler** (`packages/client/src/handlers/claude-code.ts`): new `detectClaudeAuthFailure` guard checks every incoming SDK message for the three typed-auth carriers — `assistant.error`, `system/api_retry.error`, and `auth_status.error`. On match, emits the hint event and sets a per-turn `authHintEmitted` flag; the existing result-error branch consults that flag and suppresses the redundant raw-SDK emit so the timeline shows the actionable hint instead of an opaque second line.

## Tests

`packages/client/src/__tests__/auth-error-hint.test.ts` (10 tests):

- All 6 canonical codex refresh-flow strings extracted from `@openai/codex@0.125.0`'s bundled Rust binary trip `isCodexAuthError`.
- Wrapped-in-envelope wording (`"codex exec failed: Your access token..."`) still trips.
- Unrelated errors (5xx, timeouts, sandbox, context-length, rate limit) do not.
- `isClaudeAuthError` accepts `"authentication_failed"` only; the other 6 `SDKAssistantMessageError` codes return false.
- `formatAuthHint` quotes the right CLI command + provider name per runtime.

## Test plan

- [x] `pnpm --filter @first-tree/client test` — 64 files / 645 tests pass
- [x] `pnpm typecheck` — clean
- [x] `pnpm check` — no new biome warnings (16 pre-existing warnings unchanged, none in touched files)
- [ ] Manual smoke test with a deliberately revoked `~/.codex/auth.json` to confirm the translated message appears in the chat timeline. Requires a test machine I don't have right now — would appreciate verification on a real failing-auth setup before merge.

### DIFF
--- a/packages/client/src/__tests__/auth-error-hint.test.ts
+++ b/packages/client/src/__tests__/auth-error-hint.test.ts
@@ -96,10 +96,10 @@ describe("formatAuthHint", () => {
     expect(hint).toContain("refresh token was revoked");
   });
 
-  it("targets `claude /login` for the claude-code runtime", () => {
+  it("targets `claude login` for the claude-code runtime", () => {
     const hint = formatAuthHint("claude-code", "authentication_failed");
     expect(hint).toContain("claude-code");
-    expect(hint).toContain("`claude /login`");
+    expect(hint).toContain("`claude login`");
     expect(hint).toContain("Anthropic");
     expect(hint).toContain("not First Tree's");
     expect(hint).toContain("authentication_failed");
--- a/packages/client/src/__tests__/auth-error-hint.test.ts
+++ b/packages/client/src/__tests__/auth-error-hint.test.ts
@@ -109,4 +109,15 @@ describe("formatAuthHint", () => {
     const hint = formatAuthHint("codex", "");
     expect(hint).toContain("(no message from SDK)");
   });
+
+  it("caps an oversized SDK error envelope so the hint stays readable in the timeline", () => {
+    // Codex error envelopes can occasionally include a wrapped stack trace
+    // that runs into the tens of KB. The hint should remain bounded.
+    const giantMessage = "x".repeat(5000);
+    const hint = formatAuthHint("codex", giantMessage);
+    expect(hint.length).toBeLessThan(2000);
+    expect(hint).toContain("Original SDK error:");
+    // The original message we DO include should be capped, not absent.
+    expect(hint).toMatch(/x{500,}/);
+  });
 });
--- a/packages/client/src/__tests__/auth-error-hint.test.ts
+++ b/packages/client/src/__tests__/auth-error-hint.test.ts
@@ -1,0 +1,112 @@
+import { describe, expect, it } from "vitest";
+import { formatAuthHint, isClaudeAuthError, isCodexAuthError } from "../handlers/auth-error-hint.js";
+
+/**
+ * Locks the behavioural contract of the auth-error hint module that
+ * codex.ts and claude-code.ts both consume.
+ *
+ * `isCodexAuthError` must match every canonical wording the bundled
+ * `@openai/codex` Rust binary emits when its refresh flow fails — the SDK
+ * gives us no typed code, so substring matching is all we have. Drift in
+ * either direction is bad: false positives mistranslate unrelated errors
+ * into a "run codex login" hint; false negatives let a stale `auth.json`
+ * surface as an opaque "ERROR - SDK" line that new users read as "First
+ * Tree is broken."
+ *
+ * `isClaudeAuthError` is a thin equality check against the SDK's typed
+ * `SDKAssistantMessageError` union and exists mainly so codex.ts /
+ * claude-code.ts share a single source of truth for the auth-failure code.
+ */
+describe("isCodexAuthError", () => {
+  it("matches every refresh-flow wording extracted from @openai/codex 0.125.0", () => {
+    const authMessages = [
+      "Your access token could not be refreshed because your refresh token was revoked. Please log out and sign in again.",
+      "Your access token could not be refreshed because your refresh token has expired. Please log out and sign in again.",
+      "Your access token could not be refreshed because your refresh token was already used. Please log out and sign in again.",
+      "Your access token could not be refreshed because you have since logged out or signed in to another account. Please sign in again.",
+      "Your access token could not be refreshed. Please log out and sign in again.",
+      "Your authentication session could not be refreshed automatically. Please log out and sign in again.",
+      "Token data is not available.",
+    ];
+    for (const msg of authMessages) {
+      expect(isCodexAuthError(msg), `expected auth-error: ${msg}`).toBe(true);
+    }
+  });
+
+  it("matches when the wording is wrapped in a longer SDK error envelope", () => {
+    // ThreadError messages can be wrapped by upstream layers; the keyword
+    // detector must still trigger on substring presence.
+    expect(
+      isCodexAuthError(
+        "codex exec failed: Your access token could not be refreshed because your refresh token was revoked. Please log out and sign in again.",
+      ),
+    ).toBe(true);
+  });
+
+  it("does NOT match unrelated SDK errors", () => {
+    const nonAuth = [
+      "HTTP 500 Internal Server Error",
+      "fetch failed",
+      "ECONNRESET while reading response",
+      "request timed out",
+      "sandbox denied write to /etc/passwd",
+      "context length exceeded",
+      "The server is overloaded",
+      "rate limit exceeded",
+      "",
+    ];
+    for (const msg of nonAuth) {
+      expect(isCodexAuthError(msg), `expected NOT auth-error: ${msg}`).toBe(false);
+    }
+  });
+
+  it("returns false for the empty string", () => {
+    expect(isCodexAuthError("")).toBe(false);
+  });
+});
+
+describe("isClaudeAuthError", () => {
+  it("matches the canonical SDKAssistantMessageError auth code", () => {
+    expect(isClaudeAuthError("authentication_failed")).toBe(true);
+  });
+
+  it("does NOT match other SDKAssistantMessageError codes", () => {
+    const nonAuth = ["billing_error", "rate_limit", "invalid_request", "server_error", "unknown", "max_output_tokens"];
+    for (const code of nonAuth) {
+      expect(isClaudeAuthError(code), `expected NOT auth-error: ${code}`).toBe(false);
+    }
+  });
+
+  it("returns false for undefined / empty", () => {
+    expect(isClaudeAuthError(undefined)).toBe(false);
+    expect(isClaudeAuthError("")).toBe(false);
+  });
+});
+
+describe("formatAuthHint", () => {
+  it("targets `codex login` for the codex runtime and quotes the original SDK message", () => {
+    const hint = formatAuthHint(
+      "codex",
+      "Your access token could not be refreshed because your refresh token was revoked. Please log out and sign in again.",
+    );
+    expect(hint).toContain("codex");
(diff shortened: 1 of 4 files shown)


liuchao-001:
Addressed codex review feedback in 5380ce2a:

The repo's canonical `PROVIDER_LOGIN_COMMAND` table in `packages/web/src/pages/clients/cards/shared/providers.ts:37` already uses `claude login` (not `claude /login` or `claude auth login`) for the Setup-incomplete install card. Aligned the in-chat hint to the same wording and pinned the two call sites together via comment so future changes stay in sync.

Picked `claude login` over codex's suggested `claude auth login` because (a) we already ship `claude login` in the install-card flow and the user will have seen that command once already, and (b) the repo's own source-of-truth is more authoritative than out-of-date public CLI docs.

yuezengwu (approved):
LGTM — narrow handler-boundary UX fix, no data-structure / schema changes, well-scoped tests.

Verified against `@anthropic-ai/claude-agent-sdk@0.2.84/sdk.d.ts`:
- `SDKAssistantMessage.error?: SDKAssistantMessageError`, `SDKAPIRetryMessage` (`type:'system'`, `subtype:'api_retry'`, `error`, `error_status`), `SDKAuthStatusMessage` (`type:'auth_status'`, `error?: string`) — `detectClaudeAuthFailure`'s three branches map cleanly onto the typed union.
- Per-turn `authHintEmitted` is reset on every `result` boundary, so multi-turn `query()` runs start clean.

`5380ce2a` already addressed codex's only finding. Justification (aligning with `PROVIDER_LOGIN_COMMAND`) is sound — keeping the in-chat hint and the Setup-incomplete card textually identical is the right call; if `claude login` is ever wrong it'll be wrong in one place. Cross-reference comment makes that maintenance contract explicit.

Two small follow-ups (non-blocking, your call):

1. The PR body flags the manual smoke test against a deliberately revoked `~/.codex/auth.json` as still pending. Worth ticking before declaring this fully verified — the substring keyword path is the kind of thing that's easy to get one phrase wrong on. I can run that on my side later if a real stale-auth setup shows up.

2. On `claude login` specifically: the current `claude --help` top-level surface only lists `auth` as the auth subcommand (with `auth login` inside). `claude login` as a bare alias isn't documented there, though npm's claude-code bin might still accept it — same situation as `providers.ts`. Not this PR's problem; flagging in case the team wants a single follow-up to verify both surfaces against the actual current CLI.

liuchao-001:
Pushed `609c927f` addressing both minor findings from @code-reviewer and the keyword-ordering / message-cap drive-bys.

**Bug 1 — retry double-emit**: hoisted `authHintEmitted` to `consumeOutput`'s function scope (alongside `retryCount`), so the outer `while(true)` + `respawnQuery` reentry preserves it across retries. Reset semantics simplified: only reset on a SUCCESS result. Mental model is "once you've told the user, don't repeat until at least one turn succeeds" — handles both bg-agent multi-turn within a query and the exception-driven retry path with one rule.

**Bug 2 — api_retry pre-emption**: dropped the `system/api_retry` branch from `detectClaudeAuthFailure`. `api_retry` fires before the SDK's next attempt, not as a final verdict, so emitting a hint there could pre-empt a retry that ends up succeeding. The two remaining sources (`assistant.error` and `auth_status.error`) are the authoritative post-failure signals.

**Drive-bys**:
- Reordered `CODEX_AUTH_KEYWORDS` so most-specific phrases match first (zero behavioural change, `some()` short-circuits).
- Capped the appended raw SDK message at 1000 chars in `formatAuthHint`.
- Tightened the source comment to call out that all seven codex keywords come from the bundled `@openai/codex@0.125.0` Rust binary via `strings`.
- Added a unit test for the length cap (11 tests now).

Skipping the `translateCodexError` helper extraction — would only save 3 line-of-code per emit site at the cost of a layer of indirection, doesn't seem worth the trade.

**Test state**: 64 files / 646 client tests pass, typecheck clean, no new biome warnings.

@yuezengwu's two follow-ups (manual smoke against a real revoked `~/.codex/auth.json` + cross-checking the `claude login` / `claude auth login` surface against current CLI docs) are still TBD — both depend on a real test machine I don't have right now. Holding off on merge until @liuchao-001 confirms the smoke test path tomorrow.

yuezengwu:
`609c927f` 看过了，两处都是更紧的逻辑：

**Retry 持久化**：`authHintEmitted` 提到 `consumeOutput` 外层，配合 "只在 success result 上 reset" 的单一规则，把 bg-agent 多轮 + 异常重试两条路径合一处理。对于不会自愈的 auth 失败这是正确的语义 — 告诉用户一次就够，下次真正成功之前不再重复。

**`api_retry` 移除**：同意。`api_retry` 是 SDK 即将重试的预告，不是终态；如果 SDK 自己重试成功了，前面发出去的提示就成了误报。让 `assistant.error` / `auth_status.error` / `result.subtype === "error"` 来驱动终态提示是对的。

`ORIGINAL_MESSAGE_CAP = 1000` 的 cap + 测试也合理。`some()` 短路下 keyword 重排不改行为，纯可读性收益。

11 测试通过、typecheck 干净。Approval 维持。后续等手动 smoke 之后再合。

baixiaohang (approved):
LGTM — handler-boundary 翻译，职责归属干净，不越界管凭证 / 不动 hub schema / 不动 web。三轮迭代后 retry-persistent flag + 终态信号收紧都到位了。

从 client 架构第一性原理留几条 forward-looking 观察（不阻塞本 PR，作为 tech-debt 候选）：

**1. detection 形态二分 → 触发统一中间件的阈值**

当前 codex 走「emit 出口包翻译」（4 个 site wrap `isCodexAuthError ? format : raw`），claude-code 走「入口 shape-aware 检测 + per-flag 抑制」。两种形态都自洽，但**触发条件**：client 出现第 2 类同性质的 message-massaging 需求（脱敏 / 截断 / 多语言 / 另一种错误翻译）时，正确的动作是重构出 `sessionCtx.emitEvent` 的 errorMessageMiddleware，而不是在每个 handler 里再加一层 `isXxx ? format : raw`。再加一类就是过 R3 abstraction 阈值的信号。

**2. runtime metadata 跨包 SSOT 通过文本 + 注释 pin 维持**

`PROVIDER_LOGIN_COMMAND` (`packages/web/.../providers.ts`) + 两处 web 文案 + 本 PR 的 `formatAuthHint` 里的 `reauth` —— 同一份事实在 4 处。当前用 cross-ref 注释 pin，OK。**触发条件**：runtime 数量从 2 增加到 3+ 时（gemini / aider / 其它），应该抽到 `packages/shared/runtime-metadata.ts`，web 和 client 都从那里 import。

**3. `detectClaudeAuthFailure` 缺单元测试 — 这条从 nit 升级为 strongly recommended**

不是测试覆盖度问题，是**架构脆弱点护栏**：detection 依赖的是从 `sdk.d.ts` 反推的 message shape 假设（"auth failure 一定从 `assistant.error` 或 `auth_status.error` 出来"），不是 SDK 文档化契约。任何 SDK 内部 message 重构（合并 type、新增 type、重命名 subtype、把终态搬到 `result` 路径）都会让 detection 静默漏检。建议下个 PR 加 4-5 个 case lock down 三种典型 carrier + 反例。如果 SDK shape 真的变了，CI 会在升级 `@anthropic-ai/claude-agent-sdk` 时立刻报警，而不是等用户在线上踩坑。

---

另：作者已说 hold merge 等明天 manual smoke test against deliberately revoked `~/.codex/auth.json` —— 这条是 PR body 里 unchecked 的 Test Plan 项，合理。

liuchao-001:
Thanks @baixiaohang. Engaging genuinely with the three forward-looking observations:

**1. Detection-shape bifurcation → middleware threshold**

Agreed on the diagnosis: codex's emit-site wrap and claude-code's entry-side shape-aware detection are both locally correct, and the cost of unifying them now (an `errorMessageMiddleware` on `sessionCtx.emitEvent`) is higher than the cost of letting them sit. The trigger you call out — "second class of message-massaging need" — is the right one. Sanitization / truncation / i18n / another runtime's auth surface would all push past the threshold together. Until then, I'd rather have the duplication visible at the four codex emit sites (easy to grep for) than a pre-built abstraction with one consumer.

**2. Runtime metadata SSOT**

Fully agree — `PROVIDER_LOGIN_COMMAND` already lives in `providers.ts` and the install card surfaces; this PR added a fourth textual instance (the `reauth` variable in `formatAuthHint`) guarded only by a cross-ref comment. The 2→3 runtime jump (gemini / aider / etc.) is the right trigger to lift it into `packages/shared/runtime-metadata.ts` and have web + client both import it.

**3. `detectClaudeAuthFailure` tests — accept upgrade to strongly-recommended**

This is the one I think actually warrants action, not just tracking. Detection relies on inferred-from-`.d.ts` message shapes, not a contractual API. An SDK message restructure (merging types, moving terminal auth signal into `result`, renaming `subtype`) silently breaks us today, and the failure mode is exactly "users see the opaque raw error again, hint never fires" — i.e. we silently regress the PR's whole purpose.

Plan: I'll send a small follow-up PR after this one lands with 4-5 cases pinning down (a) `assistant.error: 'authentication_failed'` triggers; (b) `auth_status` with non-empty error triggers; (c) `api_retry` does NOT trigger (we deliberately dropped it); (d) `assistant.error: 'rate_limit'` / other typed codes do NOT trigger; (e) malformed `{ type: undefined }` doesn't crash. Plus a static check that imports `SDKAssistantMessageError` and asserts our handling of every union member, so adding a new SDK code (e.g. `tos_violation`) becomes a compile-time signal to triage.

---

Holding merge for @liuchao-001's manual smoke against the user's revoked `~/.codex/auth.json` — confirmed in PR body Test Plan as the last unchecked item.

PR #631, opened 2026-05-28
fix(client): retry SDK timeouts + tighten fetchAgentConfig budget

baixiaohang:
## Summary

- `isTransientNetworkError` now recognises `TimeoutError` (Node 22+'s real shape from `AbortSignal.timeout()`), so production timeouts hit the retry layer instead of one-shot `bind aborted`.
- New `SdkCallOptions = { timeoutMs?: number }` knob plumbed through `requestJson`/`requestVoid` → `doFetch` → `doFetchOnce`.
- `fetchAgentConfig` uses a new `STARTUP_FETCH_TIMEOUT_MS = 5_000` budget — worst case drops from ≈46.5s to ≈16.5s.

## Background

A user hit `bind aborted` after a single 15s `fetchAgentConfig` timeout. The log err was `DOMException` / `name: "TimeoutError"` / `code: 23` — none of the three branches in `isTransientNetworkError` matched, so the 3-attempt retry layer was silently skipped. Root cause: `AbortSignal.timeout()` aborts with a `DOMException("...", "TimeoutError")` per Web spec, not an `AbortError`. The existing unit test (`sdk-retry.test.ts`) used a manual `err.name = "AbortError"` mock that didn't match the real runtime shape, so the gap escaped review.

## Changes

**`packages/client/src/sdk.ts`**
- `isTransientNetworkError`: add `|| obj.name === "TimeoutError"` to the name match. Inline comment explains the Web-spec divergence so the next reader doesn't re-introduce the gap.
- `STARTUP_FETCH_TIMEOUT_MS = 5_000` constant added next to `FETCH_TIMEOUT_MS`. `fetchAgentConfig` opts into it via the new `SdkCallOptions` parameter.
- `SdkCallOptions` is plumbed through `requestJson`/`requestVoid`/`doFetch`/`doFetchOnce`. `doFetchOnce` resolves to `opts?.timeoutMs ?? FETCH_TIMEOUT_MS` — every other endpoint keeps the existing 15s budget.
- `classifyRetryReason(err)` helper extracted from the retry-log site (was a three-level nested ternary; now collapses the timeout family into one bucket).
- Idempotency caveat in the `doFetch` JSDoc generalised to cover all transient signatures, not just `fetch failed`.

**`packages/client/src/__tests__/sdk-retry.test.ts`**
- New `makeTimeoutError()` factory constructs the **real** `DOMException("...", "TimeoutError")` shape (verified against the production bind-aborted log).
- Two new cases:
  - 3-fail TimeoutError → retries 3 times then rethrows
  - TimeoutError × 2 then success → retries succeed (the inverse of the prod scenario)

## Test plan

- [x] `pnpm --filter @first-tree/client typecheck`
- [x] `pnpm exec biome check packages/client/src/sdk.ts packages/client/src/__tests__/sdk-retry.test.ts` — no diagnostics
- [x] `pnpm --filter @first-tree/client test` — 903 passed / 3 skipped (the 3 skipped predate this PR)
- [x] `sdk-retry.test.ts`: 7/7 (5 existing + 2 new)

### DIFF
--- a/packages/client/src/__tests__/sdk-retry.test.ts
+++ b/packages/client/src/__tests__/sdk-retry.test.ts
@@ -41,6 +41,17 @@ function makeAbortError(): Error {
 }
 
 /**
+ * Node 22+ shape for `AbortSignal.timeout()` firings: a `DOMException` whose
+ * `name` is `"TimeoutError"` (Web spec), NOT `AbortError`. The previous
+ * `isTransientNetworkError` matched only on `AbortError`, so real production
+ * timeouts skipped the retry layer entirely. This factory captures the
+ * actual runtime shape (verified against a v0.5.x bind-aborted log).
+ */
+function makeTimeoutError(): DOMException {
+  return new DOMException("The operation was aborted due to timeout", "TimeoutError");
+}
+
+/**
  * Schedules a body returning the `n`-th item of `results` per call. Each item
  * is either a `Response` (resolved) or an `Error` (rejected).
  */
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM.

Root cause analysis is correct: `AbortSignal.timeout()` in Node 22+ throws `DOMException("...", "TimeoutError")` per Web spec, not `AbortError`, so the prior `isTransientNetworkError` match skipped retries entirely. The fix at `sdk.ts:181` is minimal and the inline comment captures the divergence so it doesn't regress.

Particularly liked:
- New `makeTimeoutError()` factory uses the **real** `DOMException` shape instead of the manual `name = "AbortError"` mock — that's exactly the gap that let this slip through review the first time.
- `SdkCallOptions` is fully optional and only `fetchAgentConfig` opts into the shorter 5s budget; other endpoints keep the 15s tolerance for cold-start / slow PG long tails.
- `classifyRetryReason` extraction collapses the prior three-level nested ternary into one helper.

Verified locally: `pnpm --filter @first-tree/client test --run sdk-retry` → 7/7 passing.

Non-blocker observation (for a follow-up, not this PR): `register()` (`/agent/me`) and `getContextTreeConfig()` also run on the startup path but stay on the 15s budget. Today's retry layer covers them; if bind-aborted reports come back from a different endpoint, this knob is now in place to apply selectively.

PR #639, opened 2026-05-28
docs(attention): rewrite skill, slim tools.md, NHA-gate tree updates

yuezengwu:
Supersedes #636 (closed automatically when its head branch was renamed to satisfy the branch-name CI check).

## Summary

- `tools.md` NHA section compressed to one sentence pointing at the attention skill; heavy detail moved out of the always-injected prompt.
- Skill rewrite: removes stale polling guidance (server echoes responses as chat messages — no need to poll), adds concrete triggers / notify body template / three-question self-check / quick lookup table.
- Tree workflow shifts from "always ask whether the tree needs updating" to **suspect-based attention**: only raise an NHA when the task is suspected to have touched decisions / constraints / ownership / shared relationships. Routine implementation changes skip both the NHA and the tree PR.
- Skill explicitly marks itself as the **baseline** — orgs / members can tighten via their own CLAUDE.md or member spec.
- Removed the stale `## Source Repos` stub in `tools.md` ("prefer the repo worktrees already present") — contradicted agent-session-cwd-redesign and was already redundant with the per-chat block from `buildChatSystemPrompt`.

## Why suspect-based instead of always-NHA on every PR

Discussed as Option A (always-NHA per shared-state action) vs Option B (suspect-based / agent autonomy). Picked B: agents have judgment, "Attention is scarce" is a stated principle, and forcing an NHA every PR would inflate notification volume without proportional safety gain. Truly irreversible categories (DB writes, prod deploys, force push, releases) remain on the unconditional ask list inside the skill.

- **Channel-aware bin name in `source-integration.ts` / `tree-templates.ts`**: still hardcode `first-tree attention raise` (matches the pre-existing `first-tree tree publish` reference in the same file). Staging/dev users would see the wrong binary. Worth a separate fix using the `getCliBinding()` pattern that `bootstrap.ts` already uses.
- **Consolidate the four parallel "when to raise" lists** in SKILL.md (Four lenses + Concrete triggers + When NOT to raise + Quick lookup). Each frames the decision differently; future readers may want one canonical list.

## Test plan

- [x] `pnpm --filter @first-tree/client test bootstrap` — 81/81 (`bootstrap.test.ts:280` regression for `first-tree-staging attention raise` still satisfied by the new one-sentence summary).
- [x] `pnpm --filter first-tree-dev test` — 249/249.
- [x] `pnpm check && pnpm typecheck` — clean (24 pre-existing biome warnings unrelated to this diff).
- [ ] Verify on a staging agent host that the new `tools.md` content renders as expected in `CLAUDE.md` (post-deploy).


baixiaohang (changes requested):
## Review by baixiaohang-assistant

Direction is right — slimming the always-injected NHA section, killing the stale polling guidance, and gating tree-update asks on suspicion all make sense. Four points before approval; #1 and #2 are the load-bearing ones, #3 and #4 are improvements.

### 1. tools.md is unconditional, but the attention skill it points at is not (R5)

`bootstrapWorkspace` writes `tools.md` unconditionally (`packages/client/src/runtime/bootstrap.ts:479`), but the `attention` skill files (`SKILL_NAMES` in `apps/cli/src/commands/tree/skill-lib.ts`, item 7) are only deployed via `tree integrate` — i.e. only for agents with a Context Tree configured.

After this PR, an agent without a Context Tree sees only the one-sentence pointer in `tools.md` ("full trigger lists … live in the attention skill (`.claude/skills/attention/SKILL.md` …)"), but `.claude/skills/attention/SKILL.md` does not exist on its disk. Before this PR, the same agent at least had the full NHA trigger checklist inlined in `tools.md`.

Net effect: a real, partial regression in NHA-decision quality for the no-tree subset of agents.

Options:

- (a) Keep a minimal trigger list inline in `tools.md` for agents whose skill file is absent.
- (b) Unconditionally deploy the `attention` skill from `bootstrap.ts`, independent of `tree integrate`.
- (c) Accept the trade-off and explicitly state in the PR body that the no-tree agent is out of scope.

Any of these is fine, but the current state silently weakens the no-tree path.

### 2. `first-tree attention raise` is hardcoded in the two tree files this PR is already touching (R5)

`apps/cli/src/commands/tree/source-integration.ts:84` and `apps/cli/src/commands/tree/tree-templates.ts:715` both spell `first-tree attention raise --requires-response` literally, while `bootstrap.ts:877` correctly uses `${bin}` via `getCliBinding()`.

Result: staging and dev agents onboarded after this lands see a command that points at the wrong binary — exactly the issue the PR body lists as a follow-up. But this PR is the one introducing the `attention raise` line in those two files. Folding the `getCliBinding()` fix into the same PR is more symmetric than shipping a known-broken instruction now and patching it later.

### 3. "Do not poll" assumes the chat-echo always lands (R4)

The new "How to wait for the response" section in `skills/attention/SKILL.md:241` tells agents to end the turn and rely on the server echoing the human's response into chat as their wake signal.

`packages/server/src/services/attention.ts:290-311` wraps the echo `sendMessage` in `try/catch` — if the echo fails, the attention still closes, the canonical answer lives only in `attentions.response`, and **no chat message is generated to wake the agent**. In a quiet chat with no other traffic, the agent never returns to the NHA.

Suggest adding a one-line fallback in the skill, e.g.: "If the chat stays silent for longer than your task's natural cadence after raising the ask, run `first-tree attention show <id>` once to check for `state=closed` directly — the canonical answer is on `attentions.response` regardless of whether the echo succeeded."

### 4. Four parallel "when to raise" lists (R1, non-blocking)

`SKILL.md` now carries Four lenses + Concrete triggers (ask/notify/chat defaults) + When NOT to raise + Quick lookup table. Each frames the decision differently. The PR body already flags consolidation as a follow-up — non-blocking, but a single canonical list would help future readers and prevent drift between the four phrasings.

---

Happy to re-review once #1 and #2 are addressed. #3 and #4 are nice-to-have and can be follow-up PRs if you prefer.

yuezengwu:
Pushed [`148fa107`](https://github.com/agent-team-foundation/first-tree/pull/639/commits/148fa107) addressing all four review points. Summary:

**#1 (attention skill missing on no-tree agents) — fixed**

- Split `SKILL_NAMES` into `CORE_SKILL_NAMES = ["attention"]` and `TREE_SKILL_NAMES`. `SKILL_NAMES` is now their union for status / doctor / link surfaces.
- New `copyCoreSkills(targetRoot)` helper + `tree skill install-core` subcommand.
- Client `bootstrap.ts` exports `installCoreSkills(...)`. Both handlers (claude-code, codex) call it unconditionally right after `bootstrapWorkspace`, before the tree-conditional `installFirstTreeIntegration`.
- Tree-bound agents still get `attention` via `copyCanonicalSkills` inside `tree integrate`; duplicate copy is idempotent and the two paths stay independently robust on purpose.

**#2 (hardcoded `first-tree` literal) — fixed**

- `buildSourceIntegrationBlock` gets an optional `binName` in `SourceIntegrationOptions` (defaults to `"first-tree"` so the four existing test call-sites stay untouched).
- `renderTreeAgentsInstructions(binName = "first-tree")` likewise.
- All four production callers (`integrate.ts`, `bind.ts`, `publish.ts`, `upgrade.ts`) plus `tree bootstrap` pass `channelConfig.binName`.

**#3 (echo-failure fallback) — fixed**

Added a paragraph to the skill's "How to wait for the response" section documenting that `packages/server/src/services/attention.ts` wraps the echo `sendMessage` in try/catch, and instructing the agent to run `attention show <id>` once on its next wake when a chat goes unusually quiet — with an explicit guardrail against turning the fallback back into a polling loop.

**#4 (four parallel lists) — still follow-up**

Per the PR body note, a single canonical list is a cleaner doc but the structural rewrite is out of scope here. Happy to do it in a follow-up.

**Tests**

- 4 new `installCoreSkills` tests (happy path, npx fallback, channel-aware bin, dev-channel graceful failure).
- 2 new `buildSourceIntegrationBlock — binName threading` tests.
- All pre-existing bootstrap + CLI tests pass locally.

Ready for another look.

baixiaohang (approved):
LGTM. Thanks for the thorough follow-up on all three blocking points — the `CORE_SKILL_NAMES` split + unconditional `installCoreSkills` from both handlers, the `binName` threading through every production caller, and the explicit echo-failure fallback (with anti-polling guardrail) all landed clean. The vitest-setup install-exec neutering in `ad985b3` is a fair test-infra trade-off.

Leaving #4 (consolidating the four "when to raise" lists in `SKILL.md`) to a follow-up as agreed.

PR #640, opened 2026-05-28
fix(client): dev-channel build fingerprint for cliDrifted

baixiaohang:
## Summary

- `resolveBundledCliVersion()` now appends `+build.<bundle-mtime>` to the version **only on the dev channel** (`getCliBinding().packageName === null`).
- Fixes a silent stale-bootstrap bug on dev installs: rebuilding via `scripts/dev-install.sh` and restarting the daemon used to leave `~/.first-tree-dev/data/workspaces/<agent>/CLAUDE.md`, `tools.md`, and `.agents/skills/*` frozen at the first build's content, because `apps/cli/package.json#version` never moves between dev rebuilds (CLAUDE.md forbids touching `version` fields) and so the handler's `cliDrifted` gate never tripped.
- Prod / staging keep the bare version — CI bumps it per release, so the fingerprint suffix would just be noise in the `.agent/cli-version` pin.

## Why now

Dev runs `bootstrap.ts` is inlined into `apps/cli/dist/cli/index.mjs`. `pnpm build` rewrites that file (so its `mtime` changes), but the surrounding `package.json#version` is constant in dev iteration. Using the bundle file's `mtime` as a build fingerprint gives `cliDrifted` an automatic trigger that doesn't depend on operator discipline (manually clearing sentinels) or external state (Context Tree HEAD drift, which dev usually doesn't have since `contextTreePath` is `null` on local hub servers).

## Test plan

- [x] `pnpm --filter @first-tree/client test` — 910 passed (4 new cases: dev appends suffix, prod/staging do not, mtime change yields different fingerprint, statSync failure degrades to bare version)
- [x] `pnpm typecheck` — 13/13 successful
- [x] `pnpm exec biome check <changed files>` — clean
- [ ] Manual: on a dev install, run `./scripts/dev-install.sh && first-tree-dev daemon restart`, send any message to a `first-tree-dev` agent, then check `~/.first-tree-dev/data/workspaces/<agent>/.agent/cli-version` — should now look like `0.5.2+build.<mtime>` and the corresponding `CLAUDE.md` should reflect the latest `generateStableClaudeMd` output.

## Notes

- `binding not initialised` path defaults to non-dev (safer: avoids emitting a fingerprint into a sentinel that a later boot would reject as drifted).
- Cleanup of stale dev agent homes after this lands: their cached `cli-version` is bare (e.g. `0.5.2`), so the first session start under the new resolver will see `0.5.2+build.<mtime>` vs. `0.5.2` → `cliDrifted=true` → full re-bootstrap, exactly what we want.

### DIFF
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -1,5 +1,5 @@
 import { execFileSync } from "node:child_process";
-import { chmodSync, existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { chmodSync, existsSync, mkdirSync, readFileSync, rmSync, utimesSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 import {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM。dev channel 用 bundle mtime 做 build fingerprint 解了 cliDrifted 一直不触发的 bug，channel gate（`packageName === null`）和 cli-binding 语义一致，prod / staging 不受影响。测试覆盖了四个分支：dev 加后缀、prod/staging 不加、mtime 变化产出不同 fingerprint、statSync 抛错降级到 bare version。binding-not-initialised 走 non-dev 分支的注释也讲清了 why（避免后续 boot 把 fingerprint 误判为 drifted）。

PR #641, opened 2026-05-28
fix(client): close dispatch ack-loss redelivery loop on dedup-hit; protect LRU recovery path

baixiaohang:
## Summary

Fixes a regression where, after a client restart, several chats had agents re-output a turn whose work had completed an hour earlier. Root cause was inside `SessionManager.dispatch()`'s dedup short-circuit, not the ack mechanism itself.

- `dispatch()` dedup branch silently returned on a redelivered `(chatId, messageId)`, assuming "the previous turn will ack on its own". When the original ack frame never lands server-side (fire-and-forget WS / TCP half-open / reaper race against `ackEntryByIdForBoundAgents`'s `status='delivered'` filter), `agent:bind`'s `resetDeliveredForInboxes` keeps resetting the entry to `pending` and `drainBacklogForAgent` keeps redelivering it. The loop only breaks when the client process restarts (dedup empty) — and then the redelivery is mis-classified as a fresh message, so the agent re-runs the turn. That is the user-visible regression.
- LRU eviction (`evictIfNeeded`) wipes `inFlightEntries[chatId]` **without** acking, relying on bind-reset redelivery against a fresh session. The dedup keys were left intact, which would have made the new re-ack shortcut that recovery path. The fix synchronously drops the evicted chat's dedup keys via `Deduplicator.dropByPrefix`.

### `packages/client/src/runtime/deduplicator.ts`
- New `dropByPrefix(prefix)` — O(n) scan that rebuilds `order` so the post-drop set keeps its FIFO contract (capacity-evict still removes the genuinely-oldest surviving key).

### `packages/client/src/runtime/session-manager.ts`
- `dispatch()` dedup branch: when the entry's `entryId` is **not** in `inFlightEntries[chatId]`, re-ack it. The three paths that remove an entry from `inFlightEntries` are now enumerated in the comment: (a) `markCompleted` via `ackInFlightEntries` — already acked; (b) `drainAllInFlightEntries` (terminate / permanent error) — already acked; (c) `evictIfNeeded` LRU — not acked, but the dedup keys are dropped synchronously so the redelivery never lands in this branch. Re-ack closes the (a)/(b) ack-loss loop; the in-flight case is left alone so a mid-turn crash keeps its recovery guarantee.
- `evictIfNeeded()`: alongside `inFlightEntries.delete(candidate.key)`, calls `deduplicator.dropByPrefix(\`${candidate.key}:\`)` so the LRU recovery contract stays intact.

### `packages/client/src/__tests__/`
- New session-manager axes:
  - `does NOT re-ack a dedup-hit while the entry is still in-flight` — protects mid-turn inflight-message-recovery.
  - `re-acks a dedup-hit when the entry is no longer in-flight` — direct repro of the "agent re-outputs after client restart" symptom.
  - `does NOT re-ack a dedup-hit whose chat was LRU-evicted` — R5 boundary: bind-reset recovery must still route through `startNewSession` -> `handler.resume`.
- New deduplicator unit tests for `dropByPrefix`: prefix-only removal, FIFO-preserving capacity-evict-oldest after drop, empty-set no-op.

## Design-doc follow-up

This PR changes the implicit "first ack always lands" premise in [`docs/inflight-message-recovery-design.md §4`](https://github.com/agent-team-foundation/first-tree-context/blob/main/docs/inflight-message-recovery-design.md). A separate `first-tree-context` PR will sync the design text to describe:

1. Client re-acks on dedup hit when the entry is no longer in-flight, closing the ack-loss loop.
2. LRU eviction synchronously drops dedup keys so bind-reset recovery against a fresh session is preserved.

## Test plan

- [x] `pnpm --filter @first-tree/client typecheck`
- [x] `pnpm --filter @first-tree/client exec vitest run` — 75 files / 915 tests pass (incl. 4 new: 1 session-manager + 3 deduplicator).
- [x] `biome check` on the 4 changed files — 0 warnings.
- [ ] Reviewer-recommended manual sanity: in a staging deploy, restart the client while at least one chat has a completed-but-unacked entry; verify no re-output.

### DIFF
--- a/packages/client/src/__tests__/deduplicator.test.ts
+++ b/packages/client/src/__tests__/deduplicator.test.ts
@@ -39,6 +39,68 @@ describe("Deduplicator", () => {
     expect(dedup.size).toBe(3);
   });
 
+  it("dropByPrefix removes only matching keys and frees capacity", () => {
+    // Used by SessionManager's LRU eviction path to drop the evicted
+    // chat's dedup keys synchronously with `inFlightEntries.delete`.
+    // Without this, a bind-reset redelivery for the evicted chat would
+    // mis-route to the dispatch dedup short-circuit and break the
+    // documented "fresh session" recovery path.
+    const dedup = new Deduplicator(5);
+    dedup.isDuplicate("chat-a:msg-1");
+    dedup.isDuplicate("chat-a:msg-2");
+    dedup.isDuplicate("chat-b:msg-1");
+    dedup.isDuplicate("chat-a:msg-3");
+    dedup.isDuplicate("chat-b:msg-2");
+    expect(dedup.size).toBe(5);
+
+    dedup.dropByPrefix("chat-a:");
+
+    // All chat-a keys are gone; chat-b keys are preserved.
+    expect(dedup.size).toBe(2);
+    expect(dedup.isDuplicate("chat-b:msg-1")).toBe(true);
+    expect(dedup.isDuplicate("chat-b:msg-2")).toBe(true);
+    // Re-adding chat-a keys is treated as first-occurrence — the fresh
+    // session can process them again.
+    expect(dedup.isDuplicate("chat-a:msg-1")).toBe(false);
+    expect(dedup.isDuplicate("chat-a:msg-2")).toBe(false);
+    expect(dedup.isDuplicate("chat-a:msg-3")).toBe(false);
+  });
+
+  it("dropByPrefix preserves FIFO order of remaining keys (capacity-evict-oldest still works)", () => {
+    // The internal `order` array is what drives capacity eviction. After
+    // a dropByPrefix the remaining keys must keep their original relative
+    // order so the next over-capacity insert evicts the genuinely-oldest
+    // surviving key.
+    const dedup = new Deduplicator(3);
+    dedup.isDuplicate("chat-a:1"); // oldest
+    dedup.isDuplicate("chat-b:1");
+    dedup.isDuplicate("chat-a:2"); // newest
+
+    dedup.dropByPrefix("chat-a:");
+    expect(dedup.size).toBe(1);
+
+    // Refill to capacity, then overflow — the oldest surviving key
+    // ("chat-b:1") must be the one evicted.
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
LGTM — well-scoped fix with clear root-cause framing.

**核心评估**
- `dispatch()` dedup 分支区分 in-flight / 不再 in-flight 是这次修复的关键决策。三条 entry 退出 `inFlightEntries` 的路径在 comment 里完整枚举：(a) `markCompleted` 已 ack；(b) `drainAllInFlightEntries` 已 ack；(c) LRU 通过同步 `dropByPrefix` 让 bind-reset 重投走正常 `startNewSession → handler.resume`，根本不进 dedup 分支。剩下 (a)/(b) 的 re-ack 是干净闭环。
- mid-turn 同 entryId 重投（`stillInFlight=true` 跳过 re-ack）保住了 inflight-message-recovery 的崩溃恢复语义。
- `Deduplicator.dropByPrefix` FIFO 保留正确，capacity-evict-oldest 仍按真正最旧的存活 key 走；O(n) 在 capacity=1000 下完全可接受。
- 测试覆盖度好：3 个 SessionManager 新 axes（in-flight / post-completion / LRU-evicted）+ 3 个 Deduplicator 单测，正好对齐三条路径。Bug 直接 repro 的那条用例尤其干净。

**前缀冲突风险**：`${candidate.key}:` 作为前缀，chatId 是 UUID 不含 `:`，无冲突。

**回归风险**：re-ack 是 fire-and-forget，server 端对同 entry 多次 ack 幂等；后续 bind-reset 会继续触发 re-ack 直到 ack 帧真的落地。

**Follow-ups**（非 blocker）
- `first-tree-context` 设计文档 `inflight-message-recovery-design.md §4` 的同步另开 PR，按"代码先合、文档同步跟上"的约定即可。
- 描述里勾留的 staging 重启 manual sanity 合前过一遍更稳。

本地 `typecheck` 通过；`session-manager.test.ts` + `deduplicator.test.ts` 共 34 个用例通过。

PR #647, opened 2026-05-28
feat(client): codex sandbox default → danger-full-access; safety via agent NHA

yuezengwu:
## Summary

- Switch the Codex handler default from `sandbox=workspace-write` to `sandbox=danger-full-access` in `buildCodexThreadOptions` (`packages/client/src/handlers/codex.ts`).
- Move the safety wall from sandbox-profile to **agent judgment + Need-Human-Attention (NHA)** for irreversible actions.
- Update the unit test that pins this default to assert the new value and document the rationale inline.

## Why

Codex is the agent's primary local-execution surface — docker, cross-directory writes, and host tools all flow through it. With `workspace-write`:

- Unix sockets outside the workspace are blocked, most visibly `~/.docker/run/docker.sock`, surfacing as `permission denied while trying to connect to the docker API at unix:///Users/<u>/.docker/run/docker.sock`.
- Any legitimate out-of-tree write the agent needs (operating predeclared repos that live above the working directory, host-side tooling, etc.) is rejected before the agent can even reason about it.

The current model is **thin sandbox + thick agent judgment**: the sandbox is no longer the safety wall; the agent is, with NHA as the human-in-the-loop escalation. A static profile cannot tell apart "modify file in /tmp" (safe) from "rm -rf /" (catastrophic) — the agent can, and is expected to NHA before crossing irreversible lines.

## Risk

This is a deliberate trade-off: codex turns now run without sandbox isolation on the host. Mitigation is policy / agent layer:

- Agents must NHA before irreversible host operations (rm of unowned paths, force-push, destructive DB changes, deploy actions, etc.).
- `approvalPolicy: never` is unchanged — codex still never blocks waiting for codex-level approval; the agent's NHA is the human gate.

## Test plan

- [x] `pnpm --filter @first-tree/client test` — 618 passed, 3 skipped.
- [x] `pnpm typecheck` — 13/13 tasks succeed.
- [x] `biome check` on touched files — clean.
- [ ] Local smoke: run a codex thread that needs `docker ps` / writes outside the workspace and confirm the previous `permission denied` is gone.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -18,7 +18,8 @@
     "build": "tsdown src/index.ts src/observability/index.ts --format esm --no-dts",
     "typecheck": "tsc --noEmit",
     "test": "vitest run --passWithNoTests",
-    "coverage": "vitest run --coverage --testTimeout=10000 --passWithNoTests"
+    "coverage": "vitest run --coverage --testTimeout=10000 --passWithNoTests",
+    "smoke:codex-sandbox": "node scripts/codex-sandbox-smoke.mjs"
   },
   "engines": {
     "node": ">=22.16.0"
--- a/packages/client/scripts/codex-sandbox-smoke.mjs
+++ b/packages/client/scripts/codex-sandbox-smoke.mjs
@@ -1,0 +1,156 @@
+#!/usr/bin/env node
+/**
+ * Codex sandbox smoke — verify `danger-full-access` actually unblocks the
+ * two motivating scenarios behind PR #647:
+ *
+ *   T1. `docker ps`                    — host docker socket access works.
+ *   T2. write to a path OUTSIDE cwd    — out-of-tree filesystem write works.
+ *
+ * Runs codex with the same ThreadOptions `buildCodexThreadOptions` would
+ * build (sandbox=danger-full-access, approval=never, model unset → CLI auth
+ * default), so a green run here is direct evidence the new default behaves
+ * as intended on this machine.
+ *
+ * Usage:
+ *   pnpm --filter @first-tree/client smoke:codex-sandbox
+ *   # or
+ *   node packages/client/scripts/codex-sandbox-smoke.mjs
+ *
+ * Exit code: 0 = both tasks succeeded; non-zero = any task failed.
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
## Review by baixiaohang-assistant — PR #647

**Summary read**
- Switches `buildCodexThreadOptions` default `sandboxMode` from `workspace-write` → `danger-full-access` to unblock docker socket and legitimate out-of-tree writes (predeclared source repos above cwd).
- Moves the safety wall from sandbox-profile to agent judgment + NHA for irreversible actions. Code comment + unit test updated.

**Verdict: approve** ✅

**Notes (non-blocking)**

⚠️ **Doc drift — `docs/runtime-provider-design.md:335`** still asserts default = `"workspace-write"` with rationale *"跟 Claude `bypassPermissions` 信任域对齐"*. After this PR (a) the default is wrong, and (b) the alignment claim only becomes *correct* with the new default — `workspace-write` was actually narrower than `bypassPermissions`. Worth updating in this PR or a quick follow-up.

⚠️ **NHA is a soft guarantee.** This PR is consistent with the claude-code handler (`claude-code.ts:713` already hard-codes `permissionMode: "bypassPermissions"` + `allowDangerouslySkipPermissions: true`), so the "thin sandbox + thick agent judgment" model isn't new — but it does mean every codex-backed agent in the org now relies on its own prompt / skill quality to NHA before irreversible host operations (rm of unowned paths, force-push, destructive docker/k8s commands). Recommend a follow-up audit of agent system prompts + the `attention` skill coverage to confirm irreversible-action gating is actually wired up.

❓ **Smoke test unchecked.** The last item in the test plan (`docker ps` / out-of-tree write) is the core motivation for this change — would be good to confirm before merge.

**What looks good** ✅
- 12-line targeted diff, no scope creep.
- Brings codex into the same trust posture as claude-code (consistency, not new risk).
- Existing UI warning text in `re-bind-dialog.tsx:191` and design doc §3.2 line 283 ("sandboxMode resets to default") remains accurate.
- Tests / typecheck / biome clean.

Approving — see the two ⚠️ above as polish, not blockers.

PR #612, opened 2026-05-27
feat: in-flight message recovery via agent:bind reset + deferred ack

baixiaohang:
## Summary

- Server: `agent:bind` flips every still-`delivered` inbox row back to `pending` and drains, so a freshly (re)connected client picks up un-acked messages within bind latency. Delete the 300s timeout reaper, the 60s scanner, and the `inboxTimeoutSeconds` / `maxRetryCount` config.
- Client: defer inbox ack from "handler about to start" to "after `forwardResult` succeeds" via new `SessionContext.markCompleted(count?: number)`. Per-chat FIFO (`inFlightEntries`); claude-code calls `markCompleted()` (default `count=1`, matches SDK 1:1 user→result pairing); codex passes `runTurn(..., entryCount)` so a fused `mergeAndRun` turn acks exactly N entries.
- E2E: wire-level proof that an UN-acked entry is redelivered on the next bind with the same `entryId`, and that mid-turn ack semantics don't drag still-in-flight entries along.

Design: [`docs/inflight-message-recovery-design.md`](https://github.com/agent-team-foundation/first-tree/blob/feat/inflight-message-recovery/docs/inflight-message-recovery-design.md).

## What broke without this

Pre-PR ack timing: server marked the row `delivered` and the client acked it the instant `dispatch` routed the message to a handler — before the handler had actually processed anything. A client crash between "ack" and "forwardResult" lost the user's message permanently; the only fallback was a 300s timeout reaper that also fired on long legitimate turns (>300s) and on noisy reconnects.

Post-PR: ack waits for the handler to call `markCompleted()` once the turn closes (forwardResult success, silent turn, SDK reported "no result", or a permanent error). A crash mid-turn leaves the row `delivered` server-side — the next `agent:bind` resets it back to `pending` and the fresh client resumes the work via `SessionRegistry` + `handler.resume`.

## Notable design decisions

- **`retryCount` not bumped on bind-reset**: a client crash isn't a "delivery attempt failed" event. Bumping the counter would push genuinely-stuck messages into `failed` too aggressively. (Death-loop protection is deferred to a future monitoring layer — design §7 risk #1.)
[...]

PR #648, opened 2026-05-28
fix(client): replay user message on transient stream-error retry

baixiaohang:
## Summary

Fix the `claude_socket_closed` auto-resume path so it actually recovers instead of hanging the session.

When the Claude SDK packages an API error as a `subtype: "success"` result whose text starts with `"API Error: The socket connection was closed unexpectedly..."`, the handler classifies it `transient / claude_socket_closed` and throws into the consumer-loop's catch + auto-resume path. Two latent bugs combined to leave that path stuck:

1. **`respawnQuery()` only rebuilt the SDK query** — the freshly-built `InputController` was empty. The SDK subprocess loaded the resumed conversation history but had no new user message to drive the next turn, so it sat idle and the runtime state stayed `working` forever. Operators saw `Attempting auto-resume (retry 1/2)` in the log and then complete silence (no retry 2/2, no recovery).
2. **`retryCount = 0` was reset at the top of the `subtype === "success"` branch, BEFORE the stream-API-error sniff.** Each wrapped error re-entered the success branch, reset the counter, sniffed positive, threw transient — looping forever between catch and respawn. `MAX_RETRIES` could never fire.

Together these manifested as a chat that went silent mid-turn after a brief Anthropic API blip, with no indication that anything could un-stick it short of a daemon restart.

## What the fix does

- **Stash + replay**: introduce `stashedSdkMessage: SDKUserMessage | null`, written at every `inputController.push` site (start / resume legacy / resume fresh / resume normal / inject) and re-pushed from `respawnQuery()`. Stashing the already-converted SDK form keeps the retry path synchronous and timing-compatible with the existing consumer-loop catch block (a `SessionMessage`-typed stash would require async `toSDKUserMessage` on the retry path, which changed timing enough to break sibling tests).
- **Pre-spawn stash ordering**: reorder `start()` / `resume()` to `toSDKUserMessage` → stash → `spawnQuery` → push so the stash is populated before the consumer loop's first iteration. Without this, an SDK that yields its first result before `start()` finishes its await would already throw transient before the stash exists.
- **`retryCount` reset moved**: from the top of the success branch into the sniff-miss leaves (forward success / forward error / no-result). The wrapped-error sniff-hit branches no longer reset, so `MAX_RETRIES` actually fires. `forward-error` resets too — semantically the SDK returned a clean result, only the downstream `sendMessage` failed, so the next turn shouldn't inherit this turn's transient counter.
- **`ackTurnClose(sessionCtx)` helper**: folds the `markCompleted() + stashedSdkMessage = null` pair into one call at all seven turn-close points (sniff-permanent / forward-success / forward-error / no-result / non-success-subtype / MAX_RETRIES / respawn-fail). `inject`'s `markCompleted(1)` stays separate — per-entry ack semantics, not turn-close.
- **Single-consumer-loop invariant documented**: the stash safety relies on the `catch → respawn → push` sequence being synchronous (no `await` between sniff-throw and the re-push) and `inject`'s microtask never interleaving inside the catch. Spelled this out on the stash declaration so future concurrent-turn / multi-consumer refactors don't quietly break the replay property.

## Why not route through `handler.resume()` (PR #612 bind-reset path)

Considered, rejected. `handler.resume()` calls `spawnQuery()`, which fires off `consumerDone = consumeOutput(...)` as a fresh consumer loop. The current consumer is still in its catch block `await`ing — under a persistent failure mode (e.g. the SDK always throws), each retry recursively spawns another consumer, each with its own `retryCount = 0`, and the recursion never terminates. This crashed vitest during development.

PR #612's bind-reset path is safe because it's externally triggered (WS reconnect after the old client process died), so there is no live consumer to race. In-process retry is internally triggered with the old consumer still on the stack — it must stay inside that consumer to avoid the recursion.

## Test plan

- [x] New regression test `packages/client/src/__tests__/claude-code-stream-error-retry-replay.test.ts` — mocks the SDK to return the wrapped API error on every attempt, captures every push into every retried `InputController`, asserts the rebuilt query (attempt 2) actually receives the original prompt and the handler eventually reaches retry-exhausted ack with `turn_end: error`. Pre-fix the test hangs in the infinite-retry loop; post-fix it completes in milliseconds.
- [x] `pnpm --filter @first-tree/client test` — 909 / 909 pass (including the new regression + the pre-existing retry-exhausted / auto-resume-failed coverage that this PR's reorderings could plausibly have broken).
- [x] `pnpm typecheck` — all 13 packages green.
- [x] `pnpm check` — 0 errors (24 pre-existing warnings unchanged, same baseline as PR #612).
- [ ] Manual staging verification: trigger a wrapped stream-API error during an active turn and confirm the agent resumes and forwards a real result, instead of going silent at `Attempting auto-resume (retry 1/2)`. (Operator follow-up; depends on Anthropic-side hiccup reproducibility.)

## Review

- @reviewer (one-shot review + revision pass) — first pass surfaced two non-blocking improvements (R1 `ackTurnClose` helper, R4 `retryCount = 0` in forward-error catch) and one architecture-comment ask (R5 single-consumer-loop invariant); all three landed before second review and were approved.

### DIFF
--- a/packages/client/src/__tests__/claude-code-stream-error-retry-replay.test.ts
+++ b/packages/client/src/__tests__/claude-code-stream-error-retry-replay.test.ts
@@ -1,0 +1,209 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import type { SessionEvent } from "@first-tree/shared";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+
+/**
+ * Regression test for the `claude_socket_closed` transient retry path.
+ *
+ * Reported bug: a Claude SDK "wrapped" stream-API error (subtype=`success`
+ * payload whose `result` text is `"API Error: The socket connection was
+ * closed unexpectedly..."`) was correctly classified `transient/
+ * claude_socket_closed` and triggered the auto-resume branch — but the
+ * old `respawnQuery()` only rebuilt the SDK query and left the new
+ * `InputController` empty. With no user prompt in the iterable the SDK
+ * subprocess just hung idle (resume mode loads conversation history but
+ * still needs a fresh user message to drive the next turn), so the agent
+ * appeared to stall mid-conversation with no further log output and no
+ * recovery.
+ *
+ * Contract this test pins:
+ *   1. On the first transient stream-error, the handler enters the
+ *      retry path (logs `Attempting auto-resume`).
+ *   2. The rebuilt query receives the SAME user message that was pushed
+ *      against the failing query — verified by capturing every SDK
+ *      input message across all `query()` calls and asserting the
+ *      original prompt shows up at least twice (once per attempt).
+ *   3. After the second attempt also fails transient + retries exhaust,
+ *      the handler surfaces the error + acks (per PR #612 § "permanent
+ *      → ack").
+ */
+
+const ORIGINAL_PROMPT = "please reply";
+const FAKE_API_ERROR_TEXT =
+  "API Error: The socket connection was closed unexpectedly. For more information, pass `verbose: true` in the second argument to fetch()";
+
+// Capture every user prompt pushed into the SDK input controller across
+// retries. Each `query()` call enqueues its own slice; we concatenate
+// them in observation order so the test can assert the same prompt was
+// replayed on the second attempt.
+const observedInputMessages: Array<{ attempt: number; content: string }> = [];
+
+function makeWrappedStreamErrorQuery(
+  promptIterable: AsyncIterable<{ message: { content: unknown } }>,
+  attempt: number,
+) {
+  let yielded = false;
+  // Eagerly drain the input iterable in the background so the
+  // handler-side `inputController.push(...)` reaches us — without this,
+  // the test never sees what got pushed.
+  void (async () => {
+    for await (const sdkMsg of promptIterable) {
+      const content = sdkMsg.message?.content;
+      const flat = typeof content === "string" ? content : JSON.stringify(content);
+      observedInputMessages.push({ attempt, content: flat });
+    }
+  })();
+  return {
+    [Symbol.asyncIterator]() {
+      return {
+        next: async (): Promise<IteratorResult<unknown>> => {
+          if (yielded) return { value: undefined, done: true };
+          yielded = true;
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM — 修得很干净，回归测试本地 7ms 跑过（修前会 hang）。

两个 bug 的耦合解释得清楚：`respawnQuery` 不回放 stash + `retryCount` 在 sniff 前重置，两个独立 bug 叠成"自动 resume 路径永远不收敛"的 hang，单独修一个都不够。

几点点赞：
- `ackTurnClose` 把 `markCompleted + stash = null` 收敛到 7 个 turn-close 点，避免以后漏置空
- stash 声明处的 single-consumer-loop invariant 注释是给未来想做并发 turn 的人的关键 warning
- 没走 `handler.resume()` / 不另开 consumer 的取舍正确——并行 for-await + 各自 `retryCount = 0` 在持续失败下会 fan-out 成无限递归，这个理由值得写进注释
- forward-error 分支也重置 `retryCount` 的语义说得通：SDK 已经返回 clean result，失败在下游 sendMessage，下一个 turn 不该继承本 turn 的 transient 计数

非 blocker 观察（可以下个 PR 看，也可以不看）：

1. **inject 与 retry 的 stash 竞态**：`inject()` 在 `await toSDKUserMessage` 期间 consumer 触发 transient retry → 用旧 stash respawn → inject 恢复后覆写 stash 再 push，新 query 会先后收到 "replay 的原 prompt + inject 的新 prompt" 两条 user message。resume 历史里其实已有原 prompt，最坏是模型看到一次重复输入再处理 inject，可恢复。stash 声明里的 invariant 讲了跨 chat 互覆盖，这个同 chat 同 turn 的序列没显式覆盖，值得补一句。

2. **`null` stash 的 admin-resume edge case**：`handler.resume(undefined, sid)` 之后立刻 stream-error 会消耗一次 retry 配额但没 payload 可重放。注释已经承认了，影响很小。

本地额外验证：
- `sdk-retry.test.ts` 7 个老用例全过
- 全量测试里 5 个失败（doc-snapshots realpath）`main` 上同样失败，与本 PR 无关

yuezengwu:
abeca1f5 收到，docs-only，把同 chat 的 inject-vs-retry 序列写清楚了（replayed prior + injected new 两条 user message → 模型看到一次重复输入，可恢复）。维持 approve。

PR #658, opened 2026-05-28
feat(web): unify agent in-progress turn into one inline WorkingTurn

Gandy2025:
## Summary

Replaces the per-event `WorkingBubble` + inline `AssistantTextRow` with a single, **borderless** `WorkingTurn` block per agent turn in the chat timeline:

- **Body** = the agent's latest narration segment (replaces in place as the turn streams).
- **Process lane** (de-emphasized, below the body, no divider): `tool_call` / `thinking` with **humanized tool lines** (`run` / `read` / `search` / `edit` … + the one key arg; full args on hover) and **per-step result previews** (first line of `resultPreview`).
- **Expand / collapse**: folded shows the current step + a legible `▾ N steps` cue; expanded shows the full process under a thin connecting **spine**, with `▴ N tools · M thinking` to collapse. Direct chats default open, group chats folded.
- **"Working" perception**: header carries a cyan **`working`** label, an enlarged pulsing dot, and a **live elapsed counter** that ticks every second (alive even between steps; reduced-motion falls back to the static cyan cues).

**Duplicate-card fix**: transient events are now grouped by `(agent, current turn)` independently of interleaved chat messages. Previously a message landing mid-turn flushed the bucket and split one turn into two `working` cards.

Net: `chat-view.tsx` −73 lines, `working-bubble.tsx` removed (−237), new `working-turn.tsx`.

## Test plan

- [x] `pnpm typecheck` (repo-wide, 13/13)
- [x] `pnpm check` (biome) — changed files clean
- [x] design-token guardrails (`lint:tokens`) clean
- [x] `@first-tree/web` unit tests 418/418
- [x] Manual: direct + group rendering, expand/collapse, result previews, error row
- [ ] Manual: reproduce a long turn + mid-turn message → confirm single card (reviewer / live)


yuezengwu (approved):
LGTM — clean UX refactor with a real bug fix (the `(agentId, turn)` grouping is the right shape; the old flush-on-non-tool bucket was the source of the split cards). Header + live elapsed counter + humanized verbs read much better than the old bubble. A few non-blocking notes for the next pass:

1. **`filterEventsForTimeline` is still globally turn-anchored, not per-agent.** It drops everything with `seq <= lastTurnEndSeq` across all agents. In a group chat with concurrent turns, when agent A’s `turn_end` lands, agent B’s in-progress events emitted before that seq disappear from the timeline (and from B’s WorkingTurn). The grouping change in this PR is correct in isolation, but the comment in `chat-view.tsx` (`filterEventsForTimeline already narrows them to that agent's current (un-ended) turn`) slightly overstates what the filter does. Worth either tightening the comment, or following up with a per-agent `lastTurnEndSeqByAgent` map in `session-timeline.ts`.

2. **No unit coverage for `WorkingTurn` or the new grouping.** The 418/418 number is "no regression," not "new behavior is asserted." Given the workgroup keying (`wg-${lastTurnEndSeq}-${agentId}`) controls remount semantics (preserves user toggle, resets across turn boundaries), a tiny vitest around the items reducer in `chat-view.tsx`, or at least one snapshot test for `WorkingTurn` (folded vs expanded, with/without body), would lock in the contract.

3. **`Manual: reproduce a long turn + mid-turn message → confirm single card` is still unchecked** in the test plan — this is exactly the scenario this PR claims to fix. Worth verifying live before merge.

Nits (optional, ignore if too small):
- `useElapsedSeconds`: the `setInterval` dep array is `[]`, which is fine, but it does mean the counter floors at 1s ticks even if `startIso` changes after mount; fine for current data flow (events arrive sorted by seq and `first` is stable), worth a one-liner comment explaining the invariant.
- `describeTool` for `Bash`: `readString(p.args, "command").split("\n")[0] ?? ""` — `split("\n")[0]` is always defined on a string, so the `?? ""` is unreachable. Cosmetic.
- PR description mentions "reduced-motion falls back to static cyan cues" — I see the pulse atom is reused via `StatusGlyph`, but the elapsed counter text still ticks every second regardless of `prefers-reduced-motion`. Probably fine (text isn’t motion), but if you want strict parity, gate the interval too.

Approving — none of the above are blockers, and the duplicate-card fix is worth landing.

baixiaohang (approved):
## Review by baixiaohang-assistant

**自动分级**: C（packages/web/** 全量，无订阅/WS frame/数据层触及）

LGTM，**duplicate-card fix** 是真问题、shape 也是对的。@yuezengwu 已经覆盖到主要的非阻塞观察（filter 仍是全局 turn-anchored、workgroup keying 的 remount 语义未被新测试锁定、PR 描述里的 manual reproduce 还没勾），都同意，不重复展开。

补一条独立观察：

**R1 ⚠️ — `formatElapsed` 同目录同名两个版本，单位不同**

`packages/web/src/components/chat/working-chip.tsx:28` 已导出 `formatElapsed(ms: number)`（有 unit test 覆盖 0.4s / 12s / 1m23s）；`working-turn.tsx` 新增了一个本地 `formatElapsed(totalSeconds: number)`（无 ms 精度、不导出）。两者同名不同单位，将来 grep `formatElapsed` 很容易拿错。

建议要么复用现有的 `formatElapsed`（把秒数 × 1000 传进去），要么把新加的改回原来的 `formatElapsedSec` 这种带单位后缀的名字，避免后续 reader 误以为是同一个东西。属于 nice-to-fix 级，不阻塞这次合并。

---

整体非阻塞，approving。

PR #673, opened 2026-05-29
Enforce suspended agent lifecycle

baixiaohang:
## Summary
- enforce agent-level suspend semantics by force disconnecting bound runtimes and ignoring runtime frames after unbind
- stop message fan-out and predictive activation from waking suspended agents while keeping /me/pinned-agents as non-deleted ownership data
- classify suspended local aliases as skipped at CLI startup and preserve them in prune/doctor
- add shared Web suspend/delete confirmation dialogs for Agent detail and Team entry points

## Tests
- pnpm check
- pnpm typecheck
- pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --minWorkers=1 src/__tests__/message-recipients.test.ts src/__tests__/message-session-creation.test.ts src/__tests__/connection-manager.test.ts src/__tests__/client-service.test.ts src/__tests__/agent-pinned-notification.test.ts
- pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1 src/__tests__/agent-slot.test.ts src/__tests__/client-connection-websocket-coverage.test.ts
- pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1 src/__tests__/sdk-surface.test.ts
- pnpm --filter @first-tree/web exec vitest run --maxWorkers=2 --minWorkers=1 src/components/__tests__/agent-lifecycle-confirm-dialog.test.tsx
- pnpm --filter first-tree-dev exec vitest run --maxWorkers=2 --minWorkers=1 src/__tests__/agent-prune.test.ts src/__tests__/client-runtime-context-tree.test.ts src/__tests__/doctor-core.test.ts

### DIFF
--- a/apps/cli/src/__tests__/agent-prune.test.ts
+++ b/apps/cli/src/__tests__/agent-prune.test.ts
@@ -53,6 +53,21 @@ describe("findStaleAliases", () => {
     expect(stale).toEqual([]);
   });
 
+  it("keeps suspended aliases pinned to this client out of the prune set", async () => {
+    writeAlias(agentsDir, "paused", "agentId: 00000000-0000-0000-0000-00000000ssss\nruntime: claude-code\n");
+
+    const remote: PinnedAgent[] = [
+      { agentId: "00000000-0000-0000-0000-00000000ssss", clientId: THIS_CLIENT, status: "suspended" },
+    ];
+
+    const stale = await findStaleAliases({
+      agentsDir,
+      clientId: THIS_CLIENT,
+      listPinnedAgents: async () => remote,
+    });
+    expect(stale).toEqual([]);
+  });
+
   it("classifies an agentId pinned to a different client as pinned-elsewhere", async () => {
     writeAlias(agentsDir, "shared-name", "agentId: 00000000-0000-0000-0000-00000000cccc\nruntime: claude-code\n");
 
--- a/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
@@ -162,6 +162,52 @@ describe("ClientRuntime context-tree wiring", () => {
     RUNTIME_TEST_TIMEOUT_MS,
   );
 
+  it(
+    "reports suspended local aliases as skipped instead of connection failures",
+    async () => {
+      const { print } = await import("../core/output.js");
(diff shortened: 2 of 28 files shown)


yuezengwu (approved):
Reviewed the full diff (+957/-287, 28 files). No DB/schema changes. Well-structured and thoroughly tested — approving.

## What this does
Enforces agent-level suspend semantics end to end:
- **Server**: `/suspend` force-disconnects the bound runtime (`agent:force_disconnect` now carries `reason: "agent_suspended"`) and sets presence offline; `forceDisconnect` propagates the reason.
- **Routing** (`message.ts`): fan-out gains a `status === "active"` filter, and explicit routing (mention / addressedTo) to a non-active agent throws a `BadRequestError` with an actionable message and writes no inbox entry.
- **Surfaces split cleanly**: `/me/pinned-agents` now returns non-deleted agents (incl. suspended, with `status`) so reconcile/prune keep local config/workspace; `listActiveAgentsPinnedToClient` narrows to `active` only so clients don't attempt binds the server will reject (grounded in the existing `agent.status !== "active"` bind rejection at `ws-client.ts:570`).
- **Client**: `agent:unbound` carries a reason; `AgentSlot` self-stops on forced unbind; `cleanupFailedStart` removes pre-bind listeners so a slot can retry on reactivation; `stop()` is idempotent.
- **CLI**: suspended local aliases are classified as `skipped (suspended)` at startup and preserved by prune/doctor.
- **Web**: suspend/delete confirm dialogs extracted to a shared component; delete is gated behind "suspend first"; `isAgentStillRoutedHere` guards in-flight frames/heartbeat after force-disconnect.

## One design point to confirm (non-blocking)
In `message.ts`, if a message explicitly mentions/addresses **any** suspended or deleted agent, the **entire** send is rejected (400) — the active recipients in the same message also receive nothing. The tests only cover the single-suspended-mention case. The error is clear and actionable (`Reactivate it before sending.`), so this is defensible, but please confirm the mixed-mention group-chat behavior is intended.

## Minor notes
- After force-disconnect, `onUnbound -> stop() -> unbindAgent()` sends an unbind frame for an agent the server has already unbound. Idempotent/harmless, just a redundant round-trip.
- `startAgentEntry` returns `"skipped"` for the `"starting"` state — only affects the count label, slightly odd semantically.

No blockers. Nice work.

PR #675, opened 2026-05-29
fix: bind Codex-created GitHub PRs

baixiaohang:
## Summary
- accept Codex command execution events as shell-backed gh create calls in the GitHub entity extractor
- add a regression test for Codex command_execution PR URL extraction
- update the stale non-shell tool test wording

## Review
- codex-reviewer passed review in chat 2dfc9b3b-6533-439c-ba85-33c6bad9cda6

## Tests
- pnpm --filter @first-tree/server test -- github-entity-extractor (ran full server suite: 152 files / 1244 tests)
- pnpm --filter @first-tree/server typecheck
- pnpm exec biome check packages/server/src/services/github-entity-extractor.ts packages/server/src/__tests__/github-entity-extractor.test.ts
- pnpm --filter @first-tree/server exec vitest run src/__tests__/github-entity-extractor.test.ts

### DIFF
--- a/packages/server/src/__tests__/github-entity-extractor.test.ts
+++ b/packages/server/src/__tests__/github-entity-extractor.test.ts
@@ -29,6 +29,22 @@ describe("extractGithubEntity", () => {
     });
   });
 
+  it("extracts a pull_request entity from Codex command_execution events", () => {
+    const result = extractGithubEntity(
+      basePayload({
+        name: "command",
+        args: { command: 'gh pr create --title "Codex PR" --body "body"' },
+        resultPreview: "https://github.com/agent-team-foundation/first-tree/pull/456\n",
+      }),
+    );
+    expect(result).toEqual({
+      entityType: "pull_request",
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM — approving.

Verified the fix end-to-end:
- `codex.ts` normalizes Codex `command_execution` into `name: "command"`, `args: { command: <string> }`, `status: "ok"` on completion, and `resultPreview: aggregated_output` — exactly what the extractor consumes.
- codex-sdk `CommandExecutionItem.command` is typed `string`, so the extractor's `typeof command !== "string"` guard won't drop Codex events.
- The `gh pr create` / `gh issue create` regexes match the Codex command string the same as Bash.

Regression test + stale wording update both look right.

Non-blocking nit: `ExtractedEntity.source` stays `"bash-gh-*"` even when the source is Codex's `command` tool. Cosmetic for a phase-1 internal tag; fine to leave or rename later.

PR #678, opened 2026-05-29
feat: register claude-code-tui runtime provider (schema + CLI + web)

yuezengwu:
> **Draft — slice 1 of 5, awaiting QA.** Part of the SDK-sunset runtime migration designed in tree proposal [`hub-tmux-runtime.20260521`](https://github.com/agent-team-foundation/first-tree-context/blob/main/proposals/hub-tmux-runtime.20260521.md) (naming resolution PR: agent-team-foundation/first-tree-context#367).

## What this is

Additive, **type-level + display-only** introduction of the `claude-code-tui` runtime provider. No client handler and no capability probe yet — those are PR2/PR3. Creating a `claude-code-tui` agent is accepted by schema/CLI/UI, but binding it is a no-op until the handler lands. Nothing about the existing `claude-code` / `codex` providers changes.

## Why it spans shared + cli + web in one slice

The web display maps are exhaustive `Record<RuntimeProvider, …>` and the shared `defaultRuntimeConfigPayload` switch has a `never` exhaustiveness guard — adding the enum value forces every display surface to gain the key in the same commit to keep `pnpm typecheck` green. So this is the smallest slice that compiles.

## Changes

| Package | Change |
|---|---|
| `shared` | `RUNTIME_PROVIDERS.CLAUDE_CODE_TUI` + enum value; `claude-code-tui` discriminated-union kind; `DEFAULT_CLAUDE_CODE_TUI_RUNTIME_CONFIG_PAYLOAD` (same 5 fields as `claude-code`); `defaultRuntimeConfigPayload()` case; barrel export; schema tests |
| `cli` | `agent create --runtime` help text lists `claude-code-tui` |
| `web` | provider labels / install / login / npm-pkg, model options + help, re-bind labels/order/unauth hint, setup copy, new-agent-dialog narrowing + preferred-runtime pick + pretty label |

tmux dependency is surfaced via the install hint ("requires tmux ≥ 3.0"), **not** the provider name — per the proposal's Q1 resolution.

## Decomposition (this PR = ①)

1. **① schema + CLI + web display** ← this PR
2. client capability probe (`claude` on PATH + `tmux ≥ 3.0`)
3. TUI handler core (tmux session + transcript tail, no hooks)
4. independent client bug fixes (`SessionRegistry.dispose()` flush; recursive `fs.watch` error handler)
5. e2e framework + TUI scenarios

## Test plan

- [x] `pnpm typecheck` — 13/13
- [x] `pnpm check` — 0 errors (24 pre-existing warnings)
- [x] `@first-tree/shared` — 363/363 (new enum + payload cases)
- [x] `@first-tree/web` — 418/418
- [x] `first-tree-dev` (CLI) — 249/249
- [ ] QA sign-off


yuezengwu:
## QA follow-up — display coverage completed (a57fde07)

Thanks for the catch. The first commit's display sweep missed two render points; both fixed:

1. **Agent-detail context bar** (`agent-detail.tsx`) — was a binary ternary `setupRuntimeProvider === "codex" ? "Codex" : "Claude Code"`, so `claude-code-tui` rendered as plain **"Claude Code"**. Replaced with an **exhaustive `Record<RuntimeProvider, string>`**, so a future provider can no longer silently fall through to the wrong label.
2. **Bound-agents list on computer cards** (`bound-agents-list.tsx`) — `formatRuntimeLabel` only recognized `claude-code`/`codex` and leaked the raw wire string `"claude-code-tui"`. Now matches against every `RUNTIME_PROVIDERS` value (still degrading truly-unknown strings gracefully).

I also swept the rest of web for the same ad-hoc-conditional pattern — `new-agent-dialog` and `providers.ts` already cover `claude-code-tui`; `setup-tab` just forwards the provider to already-covered children. No other fall-throughs.

Re: the scope question — confirmed this slice is **shared + CLI + web display**, not CLI-only. That's forced by the type system: `claude-code-tui` joining `runtimeProviderSchema` makes shared's `never` exhaustiveness guard and web's exhaustive `Record<RuntimeProvider>` maps fail to compile until updated in the same commit. The CLI is a passthrough (one help string). Happy to relabel the PR title if "CLI agent runtime" framing is the sticking point.

Green: web typecheck + lint:tokens clean, biome clean, web 418/418.

yuezengwu:
## CI fixed — green (1d038608)

The earlier red was a **stale-branch** issue, not a defect in the slice. CI tests the PR *merged into latest main*, and main had advanced 32 commits while this branch was open — including a new **reasoning-effort** feature that adds a per-provider `reasoningEffort` field (claude-code + codex) plus a `reasoning-effort-section` with exhaustive `Record<RuntimeProvider>` maps. My local typecheck passed only because the branch predated that merge.

Resolution:
- Merged latest `main` into the branch (clean, no conflicts).
- Extended `claude-code-tui` to carry `reasoningEffort` (identical claude-code contract — same `claude` CLI), in the shared schema + default payload + the web `reasoning-effort-section` maps. This is the same compile-time coupling as the original slice: the moment a provider exists, the union and every exhaustive map must include it.

Now green: **Lint & Type Check ✓ · Test ✓** (+ Migrations, Branch Name, legacy-name guard). Local mirror: `pnpm build` 6/6, typecheck 13/13, biome clean, shared 371/371, web 490/490.

Note: two CLI unit tests fail **on my macOS sandbox only** (`tree-skill-lib` EEXIST symlink, `login-command` 5s timeout) — verified they reproduce on clean `origin/main` (`73e0ca3a`) and are untouched by this branch, so they're pre-existing/environmental, and GitHub's Test job is green.

baixiaohang (changes requested):
## Review by codex-assistant — PR #678

**PR Summary**
- Problem: Introduce the `claude-code-tui` runtime provider as the first schema/display slice of the SDK-sunset migration.
- Design: Adds the shared runtime-provider enum value and config payload variant, updates CLI help, and fills web display/runtime maps for labels, setup copy, model options, reasoning effort, and re-bind surfaces. No client handler or capability probe is included in this slice.
- Affected modules: `packages/shared`, `apps/cli`, `packages/web`

**Automatic Level**: A
- Path initial: includes `apps/cli/**`.
- Semantic lift: none.

**Review Notes**

❌ 1. This slice can crash the daemon after creating or forcing a bound `claude-code-tui` agent. [R5 / apps/cli/src/commands/agent/create.ts:75]
   The PR now lets the CLI send `runtimeProvider: "claude-code-tui"` and persist that runtime locally, while server push writes the same runtime into `agent.yaml`. On daemon startup, `ClientRuntime.addAgent()` resolves `getHandlerFactory(config.runtime)`, but `registerBuiltinHandlers()` still registers only `claude-code` and `codex`; a local `claude-code-tui` config therefore throws `Unknown handler type "claude-code-tui"` before the daemon starts. If PR1 intentionally accepts this provider before the real handler lands, it still needs a guard or inert path that prevents local materialization/startup from killing the process.

✅ 2. The shared schema/config additions are additive and keep the existing `claude-code` default/legacy parse path intact.

✅ 3. The web display maps are mostly exhaustive `Record<RuntimeProvider, ...>` updates, which keeps the new provider from silently rendering as the old Claude label.

- Reason: The advertised “accepted but no-op until handler lands” behavior is not true on the daemon path; a bound local TUI agent can make startup fail.

**Executed**:
- Requested changes.

yuezengwu:
Thanks @baixiaohang — the blocking point (#1) is fixed in cb39658c.

**Root cause confirmed (and slightly worse than described):** the throw is uncaught on *two* startup paths — `daemon/start.ts` and the `login` auto-start loop — so a bound `claude-code-tui` agent crashes daemon startup before any agent runs. In `scanForNewAgents` the throw is swallowed but aborts the whole load loop, silently stopping every other agent sorted after it.

**Fix (guard at the single chokepoint, client-only — no schema/web change):**
- Added a non-throwing `hasHandler(type)` to the handler registry (exported from the client barrels).
- `ClientRuntime.addAgent` now checks `hasHandler(config.runtime)` first: if the provider is a valid enum value with no handler on this build, it prints a clear warning naming the agent + runtime and **skips** the agent (recording name/id so rescans/reconnects don't re-warn) instead of letting `getHandlerFactory` throw. A client upgrade + restart picks the agent up once the handler ships (PR3).

This makes the "accepted but inert until the handler lands" contract actually true on the daemon path. I chose the *guard/skip* path over an inert stub handler so we don't create a half-bound agent that errors on every message.

**Tests:** `hasHandler` registry test; a `ClientRuntime` test that registers a `claude-code-tui` agent *before* a `claude-code` one and asserts startup does not throw, only the supported agent builds a slot, and the warning is emitted.

**Verification:** typecheck 13/13 · biome clean · client 943/943 · cli runtime suites green (the one local `tree-skill-lib` WHITEPAPER-symlink failure is a pre-existing macOS-sandbox EEXIST, green on Linux CI). CI re-running now.

baixiaohang (approved):
## Review by codex-assistant — PR #678

**PR Summary**
- Problem: Introduce the `claude-code-tui` runtime provider as the first schema/display slice of the SDK-sunset migration.
- Design: Adds the shared runtime-provider enum value and config payload variant, updates CLI help, fills web display/runtime maps, and now adds a client-side guard so agents whose runtime has no handler on the current build are skipped instead of crashing daemon startup.
- Affected modules: `packages/shared`, `apps/cli`, `packages/client`, `packages/web`

**Automatic Level**: A
- Path initial: includes `apps/cli/**` and `packages/client/**`.
- Semantic lift: none.

**Review Notes**

✅ 1. Previous blocker is fixed: `ClientRuntime.addAgent()` now checks handler availability before resolving a factory, so a `claude-code-tui` local config is skipped with a warning instead of throwing during daemon startup. [R5]

✅ 2. The guard covers the relevant materialization paths because daemon startup, login startup, `agent:pinned` auto-add, and fs-watch rescans all flow through the same `ClientRuntime.addAgent()` wrapper.

✅ 3. The new regression test places an unsupported runtime before a supported agent and verifies startup continues with the supported slot, which pins the failure mode from the previous review.

⚠️ 4. I did not run the test suite locally because this review worktree has no `node_modules`; author-reported targeted suites are green.

**Recommendation**: approve
- Reason: The schema/display slice is coherent and the daemon-start crash risk from accepting a pre-handler provider has been closed.

**Executed**:
- Approved with prior baixiaohang authorization.

PR #676, opened 2026-05-29
feat(attachments): PostgreSQL object-storage primitive

yuezengwu:
## Summary

First phase (PR-1) of the "group-chat images aren't visible to others" work: build the **object-storage primitive** only. No business logic — image messages are a separate follow-up. This PR adds a generic, reusable binary-blob surface; upstream consumers will hold `attachments.id` references later.

- **shared**: Zod contract (`attachmentMetadataSchema`), upload header constants (`x-attachment-mime` / `x-attachment-filename`), 10 MiB size cap
- **server**: independent `attachments` table (`bytea` payload, migration `0054`), service layer, upload route `POST /api/v1/orgs/:orgId/attachments`, download route `GET /api/v1/attachments/:id`
- **client**: `uploadAttachment` / `fetchAttachment` SDK methods

### Design notes

- **Independent blob**: no `chat_id` / `message_id` columns and no foreign keys — consumers hold the id; mirrors `messages.sender_id` soft-delete tolerance.
- **Upload protocol**: `application/octet-stream` body + metadata headers, avoiding a multipart dependency. Route-local `bodyLimit`.
- **Upload auth** (org-scoped, Class B): `POST /api/v1/orgs/:orgId/attachments` + `requireOrgMembership`. The org in the path makes `uploaded_by` deterministic for multi-org callers (a `?orgId=` query was rejected — leaves uploader identity ambiguous and violates the HTTP path convention).
- **Download auth** (capability model): a valid user JWT plus knowledge of the unguessable UUIDv4 id is sufficient; there is **no** per-attachment ACL. The id *is* the bearer capability (cf. signed object URLs). A `?chatId=` membership check was considered and rejected: with no attachment↔chat binding it could only prove "caller is in *some* chat", not "this attachment belongs to *that* chat". Stronger, attachment-scoped authorization is the **consumer's** responsibility.
- **304 path** splits metadata from payload so an `If-None-Match` cache hit never drags the `bytea` out of PG.

## Test plan

- [x] `pnpm check` (Biome)
- [x] `pnpm typecheck` — shared / server / client
- [x] `vitest run attachments-route` (upload by org member, capability-model download by any authenticated user with the id, 401 without JWT, ETag 304, wrong content-type, missing mime header, empty/oversize, 404 unknown id, 403 upload to non-member org, multi-org uploader determined by path)
- [x] `node scripts/check-migrations.mjs`
[...]

PR #682, opened 2026-05-29
fix: persist chat images so all chat participants can view them

yuezengwu:
## Summary

Images sent in group chats were only visible to the sender. The bytes lived in the sender's IndexedDB plus a best-effort WS `image_payload` broadcast, so other participants never received them (reported by @liuchao-001). This wires image messages onto the `attachments` object-store primitive (PR #676): the sender uploads bytes first, the message carries only a reference, and every recipient fetches bytes on demand.

- **Send (web):** both composers — in-chat (`chat-view.tsx`) and new-chat draft (`new-chat-draft.tsx`) — upload each image to `POST /orgs/:orgId/attachments` first, then send a ref-only `format: "file"` message (`{imageId, mimeType, filename, size}`). No inline base64 over the message body.
- **Retire broadcast:** removed the WS `image_payload` frame, the `prepareImageOutbound` service, and the client-side frame handler. No back-compat shim.
- **Agents:** eager-fetch attachment bytes to local disk on delivery (`SessionManager.ensureImagesLocal`); a fetch failure degrades to a placeholder rather than blocking message routing.
- **Web render:** fetch bytes on render with an IndexedDB cache fronting the store. A cache miss (cross-device / incognito / cleared storage) is recoverable, not a dead-end.
- Dropped the 40MB message `bodyLimit` now that bodies carry refs only.

No DB/schema change — `messages.content` is JSONB and `content.imageId` now equals an `attachments` row id. Old inline-base64 historical messages still render.

## Test plan

- [x] `pnpm check` + `pnpm typecheck` clean
- [x] shared / server / client / web targeted unit + integration suites green
- [x] QA cross-user browser validation: receiver in a fresh browser context (no local cache) fetches `GET /attachments/:id` → 200 and renders the image, no failed placeholder
- [x] QA verified both send paths (in-chat composer + new-chat draft) upload then send ref-only messages


baixiaohang (changes requested):
## Review by codex-assistant — PR #682

**PR Summary**
- Problem: Group-chat images were only reliably visible to the sender because bytes lived in sender-local browser storage plus a best-effort WebSocket side channel.
- Design: Web composers upload image bytes to the attachment store first, then send ref-only `format: "file"` messages; web renderers and agent clients fetch attachment bytes on demand / delivery; the old `image_payload` WebSocket path and server broadcast service are removed.
- Affected modules: `packages/client`, `packages/server`, `packages/shared`, `packages/web`

**Automatic Level**: A
- Semantic lift: none; path level is already the maximum route.

**Review Findings**

✅ 1. Moving image bytes to the attachment store removes the sender-local IndexedDB / WebSocket side-channel dependency and gives recipients a durable fetch path.

❌ 2. The message write boundary no longer validates image-ref file content, so malformed, unsupported-MIME, or over-limit batches can be persisted and fanned out. [R1/R4 / packages/server/src/api/chats.ts:321]
   `sendMessageSchema` still accepts `content: unknown`, and after `prepareImageOutbound` was removed both the web and agent routes pass the body straight into `sendMessage`. The shared schema still defines the 20-attachment cap and supported MIME contract, but those checks are now only schema-level/tests-level facts rather than a server-side guard. A web user can stage more than 20 `image/*` files or an unsupported `image/svg+xml`, and API callers can post arbitrary ref arrays; recipients then either fail to recognize the message as an image batch or perform unbounded attachment fetches. Please centralize validation/normalization for `format: "file"` image refs at the message write boundary and restore route coverage for invalid / over-limit batches.

- Reason: the new attachment-backed path is directionally sound, but it currently removes the previous fail-closed server guard for structured image batch content.

**Executed**:
- Submitted request_changes.

yuezengwu:
Thanks @codex-assistant — addressed in 8e8d623d.

Finding #2 (no server-side guard for `format: "file"` content) is valid: after retiring `prepareImageOutbound`, `sendMessageSchema.content` being `z.unknown()` meant malformed / unsupported-MIME / over-limit ref arrays could be persisted and fanned out.

Fix: restored a fail-closed guard centralized in the `sendMessage` service (the shared write boundary for both the web and agent routes), running before the transaction opens. `format: "file"` content must parse as either a single `imageRefContentSchema` ref or a `imageBatchRefContentSchema` batch (1..`MAX_BATCH_ATTACHMENTS`, supported MIME only) — otherwise `BadRequestError`. It reuses the shared schemas, so the guard can't drift from the renderers' contract.

Restored route-level coverage in `packages/server/src/__tests__/message-file-content-validation.test.ts`: over-limit batch, malformed attachment entry, unsupported MIME (`image/svg+xml`), and non-image content — all rejected before any DB access. Existing messaging suites (messaging-e2e, agent-messages, agent-messaging-flow, message-recipients, orgs-chats) stay green.

Could you take another look when you have a moment?

baixiaohang (approved):
## Review by codex-assistant — PR #682

**PR Summary**
- Problem: Group-chat images were only reliably visible to the sender because image bytes lived in sender-local IndexedDB plus a best-effort WebSocket side channel.
- Design: Image bytes now live in the attachment store; web uploads bytes first and sends ref-only file messages, while web and agent clients fetch bytes by attachment id. The old `image_payload` WebSocket path is retired.
- Affected modules: `packages/client`, `packages/server`, `packages/shared`, `packages/web`

**Automatic Level**: A
- Semantic lift: none; path level is already maximum.

**Review Findings**

✅ 1. The attachment-store design removes the sender-local cache / best-effort WebSocket dependency and gives recipients a durable fetch path.

✅ 2. The previous blocking issue is addressed in 8e8d623d: `format: "file"` content is now validated at the shared `sendMessage` write boundary before DB writes, with coverage for malformed, unsupported-MIME, and over-limit batches.

⚠️ 3. I could not run the new server test locally because this review worktree has no `node_modules`; author-reported targeted suites are green.

**Recommendation**: approve
- Reason: the original design is coherent and the server-side validation gap has been closed.

**Executed**:

PR #687, opened 2026-05-29
feat(web): redesign Team page into Agent / Human teammates (Phase 1)

Gandy2025:
## Phase 1 — Team page redesign (Agent / Human teammates)

Replaces the merged single roster with two stacked sections — **Agent teammates** (Public / Private groups) first, **Human teammates** second — per the agreed requirements + design docs. This is **Phase 1 (pure frontend)**; the backend-dependent items (tagline field, human last-active, self-edit profile, custom avatars) follow as separate Phase 2 PRs.

### What's in here
- **Two sections, agent-first.** Agents grouped by visibility (Public / Private — group encodes visibility, no per-row chip). Members see only their own private agents; admins see all. `All | Mine` filter; own agents pinned first.
- **Columns:** Name (avatar + name + green `@handle`) · Owner (`You` highlight, dimmed for members' all-self private group) · Runs on (provider; host in hover) · Usage (right-aligned `1.24M · N turns`, `7d/30d` dropdown) · Status (blue/gray `PresenceChip`).
- **Inline delegate selector** scoped to the viewer's own agents, portal-rendered (no clipping) — **closes #669** (`selectDelegateCandidates` now takes a `managerId`).
- **Whole-row click → detail** (agent detail page / human profile dialog); a single right-aligned `⋯` kebab holds the rest (Chat + owner/admin Suspend/Delete; Remove for humans). Kebab + delegate popover `stopPropagation`; keyboard-accessible.
- **Responsive:** below 64rem rows collapse to Name + folded meta + Status + `⋯`.
- **Buttons:** `New agent` = brand `cta`, `Invite link` = neutral `outline` (admin-only).
- **DEV-only `/preview/team`** mock gallery retained (gated by `import.meta.env.DEV`) for design review.

### Tests / checks
- Web: **489/489** pass (incl. rewritten `team-groups` unit tests for `buildTeamData`).
- Repo-wide `biome check` clean (1216 files); `typecheck` 13/13.
- Note: the `apps/cli` test suite has **pre-existing flaky/order-dependent failures** (different tests fail per run, pass in isolation); this PR touches **zero `apps/cli` files**.

### Decisions captured (with @gandy2025)
Avatars circular; human last-active = `users.lastActiveAt` (Phase 2); human "Details" = inline dialog (no new page); Usage visible to all; design tuned to the shared 960 content canvas.

Design/requirements docs and the phased plan are tracked outside the repo (drafts in the team workspace).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #690, opened 2026-05-29
fix(usage): codex per-turn deltas + chat composer marker + drop Team Usage window picker

yuezengwu:
## Summary

Three small fixes layered together:

### 1. Codex per-turn token deltas (`packages/client`, `shared`)

`@openai/codex-sdk` surfaces *cumulative* thread usage at every `turn.completed`, so emitting it verbatim made each turn's usage look like the running total. Reconstruct the per-turn delta via a `computePerTurnUsageDelta` baseline, reset on `start()`; `resume()` cold-skips one turn. Emitted `inputTokens` is now `input_tokens − cached_input_tokens` so summing across turns is valid for both Anthropic and OpenAI accounting.

### 2. Chat composer cumulative token marker (`server`, `web`)

- `chatTokenUsageSchema` (shared) + `summarizeChatTokenUsage` service (overflow-safe `::bigint::text` SUM)
- `GET /chats/:chatId/token-usage` route gated by `requireChatAccess`
- Web `getChatTokenUsage` query (5s poll) renders `N.Nk tokens used in this chat` directly above ComposeStatusBar
- `filterEventsForTimeline` drops `token_usage` rows so the message list stays clean. `TokenUsagePill` (added by #661) is removed since its data source is now stripped upstream.

### 3. Drop the Usage window picker on the Team page (`packages/web/pages/team`)

The redesigned Team page (#687) put a `UsageWindowSelect` popover next to the Usage column header (and a dedicated `UsageWindowBar` in the compact viewport) that flipped the aggregation between 7d / 30d. Lock to 7 days, remove the picker, header is now a plain `Usage` label.

## Caveats

- Composer marker is **per current session history** — `session_events` are cleared by `clearEvents` on session terminate. Not a permanent billing figure.
- No DB migration.

## Test plan

- [x] `pnpm check` ✅
- [x] `pnpm typecheck` ✅
- [x] `pnpm --filter @first-tree/server test` ✅ (151 files / 1245 tests)
- [x] `pnpm --filter @first-tree/web test` ✅ (58 files / 515 tests)
- [x] `pnpm --filter @first-tree/client test` ✅ (79 files / 949 tests)
- [x] Local QA: chat composer marker renders above composer on desktop + mobile, message list stays clean. Screenshots in \`.qa-local/chat-token-usage-qa-20260529/\`.
- [ ] Local QA on Team page header against the new redesign (please re-verify — the prior QA was against the pre-redesign layout).


baixiaohang (approved):
## Review by codex-assistant — PR #690

**PR summary**
- Problem: fix Codex token usage events that reported cumulative thread totals as per-turn usage; move chat token usage from an inline timeline pill to a composer-level cumulative marker; lock the redesigned Team page Usage column to 7 days.
- Design: the Codex handler derives per-turn deltas from consecutive cumulative snapshots, treating `start()` as a fresh zero-baseline thread and cold `resume()` as a one-turn skip; the server adds a chat-scoped token usage SUM route consumed by a 5s web poll above the composer; TeamPage now calls `getOrgUsageByAgent("7d")` and TeamTable removes `usageWindow` props, `UsageWindowSelect`, and the compact `UsageWindowBar`.
- Affected modules: `packages/client/src/handlers/codex.ts`, `packages/shared/src/schemas/session-event.ts`, `packages/server/src/{api/chats.ts,services/session-event.ts}`, `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/pages/team/**`

**Automatic level**: A
- Path initial: contains `packages/client/**` -> A
- Semantic lift: none; already highest level

**Review notes**

✅ 1. Codex cumulative-to-delta lifecycle is closed: `start()` resets the baseline, cold `resume()` avoids reporting the whole prior thread as one turn, later turns diff against the new baseline, and tests cover first turn, resume, subtraction, and compaction reset.

✅ 2. The shared schema now makes `inputTokens` / `cachedInputTokens` disjoint buckets, and the Codex handler maps `input_tokens - cached_input_tokens` into that contract so OpenAI cached prompt tokens are not double-counted in aggregates.

✅ 3. The chat token usage route follows the resource-scoped chat path and reuses `requireChatAccess`; the web change removes the inline pill data source and filters `token_usage` from the timeline, so the same usage fact is not shown through two UI paths.

✅ 4. The live Team page 7d lock remains complete on the current `main`: page state, query key, and request are fixed to 7d, and TeamTable removed the desktop header picker and compact UsageWindowBar.

⚠️ 5. Non-blocking follow-ups: the chat-wide SUM filters by `chat_id + kind` while the existing token usage partial index is `agent_id`-leading, and DEV-only `/preview/team` still shows the old Usage window picker while the live page no longer does.

**Recommendation**: approve
- Reason: the state semantics, schema contract, access boundary, and live UI replacement are closed; the remaining risks are follow-up-grade.

**Executed**:

PR #705, opened 2026-06-01
Add single-org invite entry gate

baixiaohang:
## Summary
- add FIRST_TREE_ALLOWED_ORGANIZATION_ID as access.allowedOrganizationId server config
- block OAuth invite redemption for non-configured orgs before membership/redemption side effects
- block new solo OAuth onboarding when the gate is enabled while allowing returning members

## Tests
- pnpm check
- pnpm --filter @first-tree/shared exec vitest run --maxWorkers=2 src/config/__tests__/server-config.test.ts
- pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/oauth-flow.test.ts
- pnpm typecheck

### DIFF
--- a/packages/server/src/__tests__/helpers.ts
+++ b/packages/server/src/__tests__/helpers.ts
@@ -58,6 +58,7 @@ type AgentRequestFn = (
  */
 export type CreateTestAppOptions = {
   rateLimit?: Partial<NonNullable<Config["rateLimit"]>>;
+  allowedOrganizationId?: string;
   /**
    * Drop `oauth.githubApp.slug` from the test config. Used by the
    * `/github-app-installation/install-url` 503 test — the slug is the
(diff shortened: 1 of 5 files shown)


yuezengwu (approved):
LGTM — clean single-org entry gate.

Implementation matches the PR description: side-effecting calls (`ensureMembership` / `recordRedemption` / `createPersonalTeam`) are correctly gated, config schema's trim-to-undefined behavior is covered by a unit test, and the OAuth tests assert the right DB state on rejection (no member / no redemption row).

A few non-blocking observations for follow-up if useful:

1. `findOrCreateUserFromGithub` still runs before the gate, so blocked users do create `users` / `auth_identities` rows. Tests verify this is the case so it seems intentional; just worth being explicit in code/docs that the gate guards *membership/redemption side effects*, not user identity creation.
2. No structured log on gate rejection. When this is enabled in a hosted env, an `event: "entry_gate.blocked"` would help operators audit blocked attempts. Optional.
3. The "returning members of *any* org are allowed" semantics is correct per the description but worth a one-liner in deployment docs — gate filters *new* users, doesn't evict existing cross-org members.

No DB migration, config-only + route-level change. Nothing requires human escalation.

PR #740, opened 2026-06-01
feat(web): align agent detail page with the design system

Gandy2025:
## What & why

The agent detail page (`packages/web/src/pages/agent-detail/`) was an early build that drifted from `DESIGN.md`: a bespoke dropdown plus four native `<select>`s, a hand-rolled tablist, draft/dirty status drawn three different ways in two oranges, and focus styling that predated the §13 single-line rule. This converges those onto shared primitives.

## Changes

**New primitives (`components/ui/`)**
- **`Select`** — self-contained portal listbox with full keyboard + ARIA (arrows / Home / End / type-ahead / Enter / Escape-refocus), `aria-activedescendant`, viewport flip + height clamp, optional `searchable` filter (`role=combobox`). Replaces the bespoke `OptionDropdown` and 4 native `<select>`s (identity visibility/delegate, mcp transport, re-bind computer). Trigger is a bordered control → `focus-visible:border-ring`.
- **`DraftStatusChip`** — one chip for `new`/`changed`/`removed` across list rows and section headers. Green-liveness: `new`=success green, `removed`=error red (kept), **`changed`=needs-you amber** (the design decision). Removes two duplicate `ChangedChip`s and the inline per-row badges.
- **`Textarea`** — `Input`'s sibling (`border-ring` focus); fixes the prompt textarea (was `focus:` + bare border) and the two mcp textareas.

**Primitive extension**
- `TabBar`/`Tab` gain a backward-compatible `dirty` dot + `className`; `TabsNav` now uses the primitive instead of a hand-rolled tablist. Save-bar's unsaved dot + jump chips align to the same amber and §13 focus.

**Token**
- Adds `--fg-needs-you-strong` — an AA-checked foreground companion for the needs-you amber (the soft fill alone fails contrast as text), mirroring the existing `--fg-*-strong` callout family. Not a new state hue; completes the amber family.

**Cleanup**
- Deletes dead code `option-dropdown.tsx` and `section-shell.tsx`; converges the appearance avatar's inline font; adds the new primitives to the `/preview/styleguide` showcase.

## Decision

`changed` color = **amber / needs-you** (chosen by the owner among amber / neutral / new-token candidates). `new`=green and `removed`=red unchanged.

## Verification

- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) ✅
- `biome check` (303 files) ✅
- **516 web tests pass** (incl. agent-detail SSR + DOM-interaction, the latter updated to drive the new `Select`) ✅
- **2 independent code-review rounds** (multi-agent), fixes applied between them (Select state-reset on unstable `options`, focus-on-open, filtered active-index clamp, combobox ARIA, accessible names, viewport flip, re-bind placeholder).
- **Browser E2E** on `/preview/styleguide` (real Chromium): mouse + keyboard open/select, listbox receives focus on open, searchable filter, disabled guard, amber `changed` chip renders distinctly, no console errors, light + dark.

Net −254 lines of app code (3 new primitives, 2 dead files removed).


yuezengwu (approved):
**Approving** — clean design-system convergence, frontend-only, no DB/core-data changes. Reviewed the full diff statically (didn't re-run CI; your typecheck + biome + 516 tests + 2 review rounds + browser E2E are sufficient evidence).

### What it does
Replaces the agent-detail page's bespoke `OptionDropdown` + 4 native `<select>`s, hand-rolled tablist, and three-different-ways draft/dirty status with shared primitives (`Select`, `DraftStatusChip`, `Textarea`, extended `TabBar`), plus one new AA-checked token `--fg-needs-you-strong`. Net −254 lines.

### What I checked carefully
- **`Select`** keyboard/ARIA/portal logic — the meaty bit. Open-init keyed on `open` with options/value read from refs (so a parent rebuilding the options array mid-open doesn't wipe state) is the right call; the `visible`-clamp effect keeps `activeIndex` on an enabled, in-range row after filtering; pointerdown guard correctly exempts trigger + panel; commit/Escape/Tab all refocus the trigger. Type-ahead disabled in searchable mode, combobox ARIA on the input. Solid.
- **`re-bind-dialog` placeholder** (the one behavior change) — verified `selectedClientId` inits to `currentClientId`, submit is double-guarded (`canSubmit` requires `!!selectedClientId` + mutation throws on null). The placeholder only shows for an unbound agent, which actually *fixes* the old native-select display/state mismatch (showed first option while state was null). Good catch.
- **Amber convergence** — confirmed the two remaining `--state-blocked` usages (`conflictMessage`, `stale` test-result border) are genuinely different states, not draft-dirty. No leftover refs to the deleted `option-dropdown` / `section-shell`.
- **Token** — light + both dark blocks updated; AA rationale documented inline.

### Non-blocking nits (optional)
- `Select`: pressing **Home** when *every* option is disabled sets `activeIndex` to `-1` (findIndex miss). Harmless today (Enter then no-ops, aria-activedescendant goes undefined) since no consumer ships an all-disabled list, but a `Math.max(0, …)` would make it bulletproof.
- `key={o.value || "__unset"}` collides if a consumer ever passes two empty-value options; only one ("" = Remove delegate) exists today, so fine.

Neither blocks merge.

baixiaohang (approved):
## Review by codex-assistant — PR #740

**PR Summary**
- Problem: the agent detail page had drifted from the web design system through a bespoke dropdown, native selects, duplicate dirty-status chips, and older focus styling.
- Design: adds shared `Select`, `DraftStatusChip`, and `Textarea` primitives, extends `TabBar`/`Tab` for dirty state, replaces the agent-detail controls/status badges with those primitives, adds the scoped `--fg-needs-you-strong` token, and removes the dead page-local dropdown/shell code.
- Affected modules: `packages/web/src/components/ui`, `packages/web/src/pages/agent-detail*`, `packages/web/src/index.css`, `packages/web/src/pages/styleguide-preview.tsx`.

**Automatic Level**: C
- Path initial: only `packages/web/**` files are touched.
- Semantic lift: none; I did not see subscription, WebSocket frame parsing, data-layer, release, dependency, or CI workflow changes.

**Review Notes**

✅ 1. The PR removes duplicate page-local truths for select behavior and draft-status rendering, and current call sites consistently use the new primitives. [R1/R2]
✅ 2. `Select` handles the main failure edges for this rollout: unstable option-array identities do not reset an open menu, filtered active indexes are clamped, disabled options are guarded, and commit/Escape/Tab return focus to the trigger. [R4]
⚠️ 3. Non-blocking: `Select` labels the trigger, but `htmlFor`-based labels are not propagated to the portal listbox/search input when callers omit `aria-label`; consider adding `aria-labelledby` support before broader screen-reader-critical reuse. [R5 / `packages/web/src/components/ui/select.tsx`:326]
✅ 4. The new amber foreground token is scoped to the existing needs-you family and is wired in light plus both dark token blocks and the Tailwind alias. [R3]

**Recommendation**: approve
- Reason: this is a frontend-only design-system convergence with no blocking correctness or lifecycle issues found.

**Executed**:
- Approved.

Gandy2025:
Thanks for the thorough review @yuezengwu. Addressed both non-blocking nits in 6d46aef:
- **Home / all-disabled**: `setActiveIndex(Math.max(0, visible.findIndex(...)))` — no more `-1` if every option is disabled.
- **Duplicate empty-value keys**: option `key` is now index-prefixed (`${i}-${o.value}`), collision-proof if a consumer ever passes two empty values.

Re-ran typecheck + lint:tokens + biome + 516 tests locally (green); CI re-running on the new commit.

yuezengwu:
Confirmed `6d46aef` — both fixes are correct:

- **Home / all-disabled**: `Math.max(0, …findIndex)` keeps `activeIndex` in range; the all-disabled list now lands on index 0 instead of `-1`. ✅
- **Duplicate empty-value keys**: `${i}-${o.value}` is collision-proof. Index-based keys are safe here since each option is a stateless button re-rendered in full and active/selected state is derived from props, so no reconciliation staleness. ✅

Approval stands. The one item left open (codex's `aria-labelledby` propagation to the portal listbox/search input when callers omit `aria-label`) is genuinely a "before broader SR-critical reuse" follow-up, not a merge blocker — today's call sites all pass `aria-label` or label the trigger via `htmlFor`. No need to hold the PR for it.

baixiaohang (approved):
## Review by codex-assistant — PR #740 follow-up

**PR Summary**
- Problem: follow-up after the synchronized head `6d46aeff` added a small hardening commit for the new `Select` primitive.
- Design: the new commit keeps `Home` from setting `activeIndex` to `-1` when every visible option is disabled, and prefixes option keys with the visible index so duplicate empty-value options cannot collide.
- Affected modules: `packages/web/src/components/ui/select.tsx`.

**Automatic Level**: C
- Path initial: only `packages/web/**` is touched.
- Semantic lift: none; this is a local UI primitive hardening delta with no data-layer, subscription, WebSocket, release, dependency, or CI workflow impact.

**Review Notes**

✅ 1. The synchronize delta is tightly scoped to the two previously noted `Select` edge cases and does not reopen the broader agent-detail behavior surface. [R2/R5]
✅ 2. The `Home` key path now keeps `activeIndex` in range for an all-disabled option set, while existing commit guards still prevent disabled selections from firing. [R4 / `packages/web/src/components/ui/select.tsx`:226]
✅ 3. Prefixing option keys with the visible index removes the duplicate empty-value key collision without changing the option ids used by `aria-activedescendant`. [R4 / `packages/web/src/components/ui/select.tsx`:352]

**Recommendation**: approve
- Reason: the new head only hardens non-blocking UI edge cases and introduces no blocker.

**Executed**:
- Approved.

yuezengwu:
Re my earlier comment: the index-prefixed key I endorsed (`${i}-${o.value}`) trips biome's `noArrayIndexKey`, so `fa92538b` correctly reverts to the value-based `o.value || "__unset"` — the lint rule wins over what was only a hypothetical collision (no call site ships two empty-value options today). The Home / all-disabled guard is retained. Approval stands.

baixiaohang (approved):
## Review by codex-assistant — PR #740 follow-up

**PR Summary**
- Problem: follow-up after the synchronized head `fa92538b` fixes the lint fallout from the previous `Select` key hardening.
- Design: keeps the `Home` / all-disabled active-index guard, reformats it under Biome, and restores a value-based option key so `biome check` does not trip `noArrayIndexKey`.
- Affected modules: `packages/web/src/components/ui/select.tsx`.

**Automatic Level**: C
- Path initial: only `packages/web/**` is touched.
- Semantic lift: none; this is a local UI primitive lint/edge-case delta with no data-layer, subscription, WebSocket, release, dependency, or CI workflow impact.

**Review Notes**

✅ 1. The synchronize delta is isolated to `Select` and does not reopen the broader agent-detail behavior surface. [R2/R5]
✅ 2. The `Home` key guard still keeps `activeIndex` in range for an all-disabled visible option set, while commit-time disabled guards remain intact. [R4 / `packages/web/src/components/ui/select.tsx`:227]
✅ 3. Returning to a value-based React key is consistent with the current `Select` contract: `value` is the single selection identity, and all current call sites have unique values with at most one empty sentinel. [R1/R4 / `packages/web/src/components/ui/select.tsx`:357]
⚠️ 4. Non-blocking: broader future reuse should keep that uniqueness invariant explicit, or add a separate stable option id if duplicate values ever become valid. [R1/R5]

**Recommendation**: approve
- Reason: the new head preserves the useful all-disabled guard, fixes lint compatibility, and introduces no blocker.

**Executed**:
- Approved.

Gandy2025:
Heads-up for transparency: my first nit-fix push made the option `key` index-based, which tripped `biome check .` (`noArrayIndexKey`) in CI's Lint & Type Check. Reverted to the original value-based key in fa92538 (option values are unique in practice → collision-safe), keeping the Home/all-disabled guard. Now `pnpm check` + typecheck + 516 tests pass locally and Lint & Type Check + Test are green on CI. (Lesson logged: gate on `pnpm check`, not a path-scoped biome run.)

Gandy2025:
Thanks @baixiaohang. Addressed the non-blocking a11y note in 2c29522: the three Selects that relied on a visible `<Label htmlFor>` (Visibility, Delegate Mention, Transport) now also pass a matching `aria-label`, so the portal listbox / search input get an accessible name — consistent with the Model / Reasoning effort / Computer Selects. `pnpm check` + typecheck + 516 tests green. (A general `aria-labelledby` passthrough is a reasonable future enhancement for broader reuse; aria-label covers all current call sites.)

baixiaohang (approved):
## Review by codex-assistant — PR #740 follow-up

**PR Summary**
- Problem: follow-up after the synchronized head `2c295225` addresses the remaining non-blocking accessibility note on `Select` popup naming.
- Design: adds explicit `aria-label` values to the labelled `Select` call sites that previously relied only on visible `<Label htmlFor>` text: identity visibility, delegate mention, and MCP transport.
- Affected modules: `packages/web/src/pages/agent-detail/identity-section.tsx`, `packages/web/src/pages/agent-detail/mcp-section.tsx`.

**Automatic Level**: C
- Path initial: only `packages/web/**` is touched.
- Semantic lift: none; this is a local UI accessibility labeling delta with no data-layer, subscription, WebSocket, release, dependency, or CI workflow impact.

**Review Notes**

✅ 1. The synchronize delta is scoped to the three current labelled `Select` call sites that needed explicit popup/search naming. [R5]
✅ 2. The added labels match the visible field labels, so the trigger, portal listbox, and searchable input now expose the same field name without changing selection behavior. [R1/R4]
✅ 3. Other agent-detail `Select` call sites already had `aria-label`, so the immediate blast radius is covered. [R5]

**Recommendation**: approve
- Reason: the latest head closes the remaining non-blocking accessibility follow-up without introducing new behavior risk.

**Executed**:
- Approved.

baixiaohang (approved):
## Review by codex-assistant — PR #740 follow-up

**PR Summary**
- Problem: follow-up after the synchronized head `da8f4381` applies live-review UI polish on the agent detail redesign.
- Design: removes the tab button radius so the active underline remains visually straight, and changes the prompt inline editor `Done` button from primary/default to outline because it only collapses the editor; the Save Bar remains the actual submit path.
- Affected modules: `packages/web/src/components/ui/tab-bar.tsx`, `packages/web/src/pages/agent-detail/prompt-section.tsx`.

**Automatic Level**: C
- Path initial: only `packages/web/**` is touched.
- Semantic lift: none; this is local web presentation/affordance polish with no data-layer, subscription, WebSocket, release, dependency, or CI workflow impact.

**Review Notes**

✅ 1. Removing the tab radius is consistent with the primitive's underline-tab lifecycle: the active state is a straight bottom border, and the change affects only the shared `Tab` primitive plus its current web preview/agent-detail consumers. [R3/R5]
✅ 2. Making prompt `Done` outline aligns visual weight with behavior: it closes the inline editor but does not persist the draft, preserving the Save Bar as the single submit affordance. [R1/R3]
✅ 3. `git diff --check` is clean; I did not see layout, focus, or keyboard behavior changes beyond the intended styling adjustment. [R4]

**Recommendation**: approve
- Reason: the new head is a scoped web UI polish delta and introduces no blocker.

**Executed**:
- Approved.

baixiaohang (approved):
## Review by codex-assistant - PR #740

**建议**: approve
- 理由: The latest synchronize keeps the design-system convergence intact and correctly restores the token aliases needed by the PR after merging current main.

**分级**: C
- Path 初判: only `packages/web/**` files are in the PR diff.
- 语义升级: none; this remains UI primitives, CSS tokens, and DOM tests, with no subscription/WS/data-layer changes.

**PR 摘要**
- 解决问题: aligns the agent detail page with shared UI primitives and consistent draft/dirty status styling.
- 设计方案: adds shared `Select`, `Textarea`, and `DraftStatusChip`; removes the old dropdown/shell paths; extends `TabBar`; restores the amber needs-you token family after the main-merge token drift.
- 受影响模块: `packages/web/src/components/ui`, `packages/web/src/pages/agent-detail`, `packages/web/src/index.css`, related DOM/styleguide tests.

**Review 意见**
✅ 1. The restored needs-you aliases cover the actual consumers: `bg-state-needs-you-soft`, `text-state-needs-you-strong`, and raw `--state-needs-you` dirty dots all resolve again in light, dark, and marketing-dark scopes. [R1/R5]
✅ 2. The follow-up keeps `DraftStatusChip` as the single draft-status visual source instead of reintroducing local amber/orange variants. [R1/R2]
✅ 3. The updated DOM tests now drive the custom `Select` as a button plus portal listbox, including disabled fallback and Escape refocus behavior, matching the component contract rather than the old native-select path. [R4]

**已执行**
- 已 approve latest head `7f586015`.

PR #747, opened 2026-06-02
refactor: remove NHA (Need-Human-Attention) feature end-to-end

yuezengwu:
## Summary

Removes the NHA / attention primitive and the legacy `pending_questions` code end-to-end (server, shared, client, web, CLI, skills), clearing the ground for the **group-chat-unified-send** redesign that rebuilds attention on top of the `messages` table.

This is a **deletion-only** PR — no replacement attention feature is introduced here. Asking-a-human is intentionally left unavailable (agents see a `[pending redesign, 自行判断]` placeholder) until P4/P5 land.

## What is removed

- **Server** — `attentions` + `pending_questions` schema, services, REST routes (user + agent), supersede hooks (`markSupersededByChat/Agents`) and their callers.
- **Shared** — attention + question zod schemas. The `needs_you` axis is fully excised from agent-status (enum value, `MAIN_STATUS_PRIORITY`, `needsYou` field, `deriveMainStatus` arm) and cascaded through the me-chat projections (`pendingQuestionAgentIds` / `chatHasOpenQuestion`).
- **Client** — attention frame dispatcher. The `AskUserQuestion` `canUseTool` deny-bridge is replaced with SDK `disallowedTools`; the `claude-code-tui` AskUser text-degrade path is replaced by a `--disallowed-tools AskUserQuestion` launch flag (the TUI runs `--dangerously-skip-permissions`, so `disallowedTools` is the only way to block the tool on that path).
- **Web** — `AttentionCard` (chat-bottom), `AttentionsSection` (right sidebar), attention API client, command-palette NHA group, and the full `needs-you` status rendering + palette tokens (`--state-needs-you*`, pulse class, tones entry).
- **CLI** — the `attention` command namespace (raise/respond/cancel/list/show) + core module.
- **Skills/docs** — `skills/attention`, needs-attention scoping docs, and all stale doc/comment references.

## ⚠️ Key decisions for reviewer / @yuezengwu

- **Tables are NOT dropped.** `attentions` (migration `0052`) and `pending_questions` (migration `0034`) survive as **dead audit tables** holding the dogfood-era rows. No `DROP TABLE` is added; the drizzle journal is untouched. A future cleanup PR can drop them if desired.
- **No data migration.** In-flight open attentions from the dogfood window die silently (CLI + UI both gone). This was an explicit decision.
[...]

PR #755, opened 2026-06-02
fix(server): never auto-archive manual chats that hold a github mapping (#745)

### LINKED ISSUE #745: Manual chats without PRs are archived overnight
## Problem

Manually created chats that are not linked to a PR are automatically archived after roughly one night.

This makes manual discussion chats disappear unexpectedly even when they still need to remain visible or active.

## Expected behavior

Manual chats without a PR should not be archived solely because they have no PR link, or they should follow an explicit lifecycle policy that preserves user-created chats unless they are intentionally archived.

## Notes

Reported from a manually created First Tree chat.

yuezengwu:
Closes #745. Builds on top of #810.

## Background

#810 already fixed part of #745: it excluded **human-owned** chats from the
no-mapping auto-archive route (Route B), so manually-created discussion chats
with no GitHub link stay active.

This PR closes the **remaining gap** from the agreed scope: a manual chat that
later acquires a `github_entity_chat_mappings` row — e.g. an `agent_created`
binding when an agent opens a PR/Issue from inside it — was still archived by
Route A once the bound entity went terminal, even though the chat is
manual-origin.

## Change

- `sweepMapped` (Route A) now requires `chats.metadata.source = 'github'`.
  That is the same discriminator the conversation list classifies on
  (`me-chat.ts` `KNOWN_NON_MANUAL_PREDICATE`). A manual chat that merely
  acquired a mapping row is not github-minted and is left alone.
- Verified `agent_created` bindings only insert a mapping row and never write
  `chats.metadata`, so a manual chat stays manual-origin.

### Why origin, not ownership

#810's human-owner carve-out cannot be extended to Route A: github-source chats
are created with their human reporter as `role='owner'` too, so gating Route A
on ownership would wrongly stop **all** github chats from ever archiving.
Origin (`metadata.source`) is the correct discriminator for the mapped route.

GitHub-source chats keep the current policy unchanged (archived only when all
bound entities are terminal AND idle past the threshold, with the per-user
unread carve-out). No schema or migration changes.

## Tests

- `chat-archive.test.ts`: Route A scenarios re-seeded as github-source; added a
  case for a manual chat with an `agent_created` mapping (not archived).
  #810's Route B tests untouched. 17/17 pass.
- Full server suite 1302 passed; typecheck + biome clean.

### DIFF
--- a/packages/server/src/__tests__/chat-archive.test.ts
+++ b/packages/server/src/__tests__/chat-archive.test.ts
@@ -2,9 +2,12 @@
  * Unit tests for the chat auto-archive sweeper (`sweepChatArchive`). Covers
  * the two routes the sweeper owns:
  *
- *   Route A — chats with GitHub mappings: archive when every mapped entity
- *             is terminal AND `last_message_at` is older than the mapped
- *             idle threshold.
+ *   Route A — github-minted chats (`metadata.source = 'github'`) with GitHub
+ *             mappings: archive when every mapped entity is terminal AND
+ *             `last_message_at` is older than the mapped idle threshold. A
+ *             manual chat that merely acquired a mapping row (e.g. an
+ *             `agent_created` binding) is not github-minted and is left alone
+ *             (issue #745).
  *   Route B — chats with no GitHub mapping and no human owner: archive
  *             (chat, user) pairs with no unread mentions AND
  *             `last_message_at` older than the unmapped idle threshold.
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
## Review by codex-assistant — PR #755

**PR Summary**
- Problem: manual-origin chats that later receive GitHub mapping rows can still be archived by Route A once the mapped entity is terminal.
- Design: require `chats.metadata.source = 'github'` in Route A, while leaving Route B and the GitHub terminal + idle policy unchanged.
- Affected modules: `packages/server/src/services/chat-archive.ts`, `packages/server/src/__tests__/chat-archive.test.ts`.

**Automatic Level**: B-high
- Path initial: touches `packages/server/src/services/chat-archive.ts` (`packages/server/**` service `chat*`) -> B-high.
- Semantic lift: none; already B-high.

**Review Notes**

❌ 1. This changes durable chat auto-archive policy but does not include or link the required Context Tree update. [R5 / first-tree-cloud/messaging.md]
   The current tree still describes Route A as “chats bound to GitHub entities” and Route B as “chats with no GitHub mapping”; after #810 plus this PR, the durable rule is “Route A only archives GitHub-origin chats” and Route B excludes human-owned chats. Please open/link the `agent-team-foundation/first-tree-context` update before this source PR merges.

✅ 2. The code change itself matches the agreed scope: a manual chat with an `agent_created` mapping is excluded from Route A, while GitHub-source terminal + idle behavior remains unchanged.

✅ 3. Tests preserve the existing Route A/Route B coverage and add the mapped-manual regression case; local verification passed for `chat-archive.test.ts` (17/17), server typecheck, and `git diff --check`.

- Reason: block merge until the Context Tree drift is handled; no code-path blocker found.

**Executed**:
- Submitted request_changes.

yuezengwu:
Closing — this branch has fallen too far behind `main` to be worth carrying forward (the base churned through #747 NHA removal and #810, requiring two reworks already).

State for the record:
- #810 (merged) already covers the main report: human-owned no-mapping chats no longer auto-archive.
- The only remaining piece this PR added was the Route A refinement — a manually-created chat that later receives a GitHub mapping (e.g. an `agent_created` binding) should also be skipped, keyed on `metadata.source = 'github'` rather than mapping presence. That refinement is **not** in `main`; if it's still wanted it can be re-filed fresh against current `main` (the change is a one-line Route A filter + tests).

Not merging. See #745.

PR #704, opened 2026-05-31
test: expand cli and web coverage

bestony:
## Summary

- Expands CLI test coverage across agent config/lifecycle commands, chat attention flows, tree commands, daemon/login/service helpers, onboarding, output helpers, and runtime context-tree behavior.
- Expands web test coverage across API wrappers/client requests, layout and dialog interactions, chat working states, agent detail panels, team/workspace/onboarding pages, command palette, conversations, and utility hooks.
- Matters because these tests lock down existing behavior across the highest-traffic CLI and web surfaces without changing runtime code.

## Validation

- [ ] `pnpm check`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`

## Change Surface

- [ ] `apps/cli` public CLI or help output
- [ ] `packages/github-scan` / `first-tree github scan`
- [ ] tree onboarding / binding / inspection behavior
- [ ] shipped or planned skill topology
- [ ] docs or contributor-facing repository metadata
- [ ] CI / packaging / release plumbing

## Notes

- package or install behavior changes: none
- docs or tests updated to match: tests only; no docs changes needed
- follow-up work: run the full validation suite before merge

PR #756, opened 2026-06-02
Add Context Tree IO statistics

baixiaohang:
## Summary
- add durable `context_tree_io_events` facts for explicit Context Tree read/write IO
- record IO candidates through server-side validation during agent WS `session:event` ingestion
- expose read/write IO summary, by-agent rows, and recent feed in Context Tree snapshot/web tab

## Validation
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/context-tree-io.test.ts src/__tests__/ws-session-event.test.ts src/__tests__/org-settings.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/codex-context-tree-io.test.ts src/__tests__/claude-code-tool-events.test.ts`
- `pnpm check`
- `pnpm typecheck`

## Notes
- `shell_command` IO candidates are disabled until P1 server-side shell parsing exists.
- Migration `0055_context_tree_io_events.sql` was hand-authored because the repo's Drizzle snapshots currently have existing collision issues.


yuezengwu (commented):
## Review summary (yzw-assistant)

**What this does:** Adds a durable, server-validated record of explicit Context Tree read/write IO. A new `context_tree_io_events` table (migration `0055`) is populated during `session:event` ingestion: the client handlers (claude-code `Read/Write/Edit/MultiEdit`, codex `file_change`) attach `contextTreeIo` candidates to `tool_call` events, the server re-validates them (canonical repo-URL match, path normalisation, action/target checks) before persisting. Legacy `context_tree_usage` reads are backfilled on read. The snapshot/web Context tab gains a read/write summary, a by-agent table, and a recent feed. Server is the trust boundary — clients only propose candidates.

The trust-boundary design (re-validate every client candidate, canonical repo-URL gate, `posix.normalize` + traversal/`\0` rejection, fail-closed `viewerCanAccess`, cross-org chat masking) is solid and well-tested (`context-tree-io.test.ts` covers SSH/HTTPS URL equivalence, idempotency, mismatch skipping).

### CI status — red, but NOT caused by this PR
Both `Lint & Type Check` and `Test` fail, but **every failing file belongs to an unrelated Attention/NHA feature** (`api/attention.js`, `Attention` export, `pendingQuestionAgentIds`, `chatHasOpenQuestion`, `needs_you`) — none of the context-tree-io files this PR touches appear in the errors. I confirmed `origin/main` HEAD (`860aa28d`) is itself **already red with the identical errors**; this branch's base (`b4c9f79`, ~3h older) predates that breakage, which is why the author's local run was green. **Action: rebase/merge onto current `origin/main` to get a real green signal** (and to pull in the missing `api/attention.js` consistently). The PR can't be honestly verified as green until main is green again.

### Needs human owner review (flagged to @yuezengwu)
- **New table + hand-authored migration `0055_context_tree_io_events.sql`.** Schema and the hand-written SQL match (FKs, the 4 indexes, the 3 CHECK constraints, the `(source_session_event_id, source_index)` unique index). `Migrations` CI job passes, so it applies cleanly. Note: no `meta/0055_snapshot.json` was added (only `_journal.json` bumped) — acceptable given the documented pre-existing Drizzle snapshot drift, but it leaves `drizzle-kit generate` still unable to run until that drift is fixed.

### Discussion points (non-blocking)
1. **Critical-path coupling in `ws-client.ts` `session:event`.** `recordFromSessionEvent` now runs *inside the same transaction* as `appendEvent`. If IO recording throws, the whole transaction rolls back and the **primary session_event is lost** (client gets an error frame, no retry for `session:event`). In practice the risk is low — inputs are pre-validated, `normalizeCandidate` enforces all CHECK constraints, and conflicts use `onConflictDoNothing` — but it does make the user-visible timeline write hostage to a secondary statistics feature. Please confirm the atomic coupling is intentional vs. making IO recording best-effort (it has no FK to `session_events`, so it could tolerate being decoupled).
2. **Write-on-read backfill.** `summarizeContextTreeIo` calls `backfillLegacyContextTreeUsage` on every snapshot GET, and that helper calls `recordFromSessionEvent` per legacy row (an extra `getOrgContextTree` + `chats` query each). After the first backfill it's a cheap `NOT EXISTS` no-op, but the first snapshot load post-deploy for an org with a large `context_tree_usage` history could be slow / serial. Consider bounding or one-shotting it if that window matters.

Holding off on approve pending (a) green CI after rebase and (b) your sign-off on the new table/migration. Implementation itself looks good.

yuezengwu (commented):
## Re-review of `fbcd690` "refactor: derive context tree io on server"

Nice refactor — this directly tightens the trust boundary I'd flagged, and goes further than I'd have asked for:

- The client no longer asserts the semantic classification at all. `contextTreeIoCandidate` (which carried client-chosen `action`/`source`) is gone; the client now emits a neutral `ToolFileRef` (`origin`, `localPath`, optional `repoUrl`/`repoBranch`/`repoRelativePath`, `pathKind`) as pure evidence.
- The **server** is now the sole authority on `action`/`source`: `deriveEventIo()` maps `(runtimeProvider, toolName)` → read/write + source, independent of anything the client claims. The new test "derives Claude read and write source from tool name, not client-provided action" locks this in.
- `normalizeFileRef` still requires a canonical repo-URL match against the binding, so local-only refs (files outside the tree) are dropped server-side. Path normalisation / traversal rejection is preserved.
- No DB shape change — this is purely the client↔server wire contract, so migration `0055` is unaffected.

Logic checks out: `recordFromSessionEvent` gates on `deriveEventIo` first (null → skip), then extracts refs; codex non-`file_change` tools and claude non-view tools correctly produce no derivation; legacy `context_tree_usage` still maps to read/`legacy_context_tree_usage`.

### Still blocking merge (unchanged from prior review)
1. **CI red is NOT this PR.** The new run fails on the exact same unrelated stale web tests (`api/attention.js`, `Attention` export, `pendingQuestionAgentIds`, `needs_you`) — `origin/main` HEAD is itself red with these, and #759 ("test(web): remove stale attention fixtures") is the in-flight fix. None of the failures touch any file in this PR. **Merge main (after #759 lands) to get a real green signal** — the branch hasn't been rebased, so it still inherits the broken base.
2. **DB table + hand-authored migration `0055`** still needs the owner's sign-off (human review).

### Still open (non-blocking, from prior review — `ws-client.ts` untouched here)
- `recordFromSessionEvent` runs inside the same transaction as `appendEvent`, so an IO-recording throw would roll back the primary session_event. Low practical risk, but please confirm the atomic coupling is intentional (no FK to `session_events`, so it could be best-effort).
- `summarizeContextTreeIo` backfills legacy rows on every snapshot GET (per-row `getOrgContextTree`/`chats` query); first post-deploy load for a large history could be slow.

Implementation LGTM; holding approve on green CI + owner sign-off on the schema/migration.

yuezengwu (changes requested):
## Requesting changes — decouple `recordFromSessionEvent` from the session_event write

**Blocker (owner decision).** In `ws-client.ts`, the `session:event` handler now wraps `appendEvent` and the new `recordFromSessionEvent` in the **same transaction**:

```ts
await app.db.transaction(async (tx) => {
  const persistedEvent = await sessionEventService.appendEvent(txDb, ..., { deferSideEffects: true });
  if (boundInfo) {
    await contextTreeIoService.recordFromSessionEvent(txDb, { ... }); // ← throws ⇒ whole tx rolls back
  }
  return persistedEvent;
});
```

If `recordFromSessionEvent` throws, the transaction rolls back and the **primary `session_event` is lost** — and since `session:event` is fire-and-forget with no client retry, that timeline event is gone for good. That couples the user-visible chat timeline (and token usage / GitHub-entity binding, which all ride this path) to a secondary statistics feature.

The IO statistics are explicitly **allowed to fail and be dropped** — `context_tree_io_events` has no FK to `session_events`, so losing one stat row is fine; losing a timeline event is not. The two must not share a failure domain.

### Required change
Make IO recording **best-effort**, isolated from the session_event write. Either is fine:

- **(preferred, smallest diff)** Keep the structure but wrap the IO call in its own `try/catch` inside the tx so its failure can't abort the `appendEvent` write:
  ```ts
  if (boundInfo) {
    try {
      await contextTreeIoService.recordFromSessionEvent(txDb, { ... });
    } catch (err) {
      app.log.warn({ err, agentId, chatId: payload.chatId }, "context-tree IO record failed (non-fatal)");
    }
  }
  ```
  (Note: if you swallow inside the tx, confirm the driver doesn't leave the tx in an aborted state after a statement error — if it does, prefer the option below.)
- **(cleanest separation)** Move `recordFromSessionEvent` out of the transaction entirely — run it after `appendEvent` commits, fire-and-forget with a `.catch()` log, same as `emitAppendEventSideEffects` / `notifySessionEvent` already do.

Everything else in `fbcd690` LGTM — the server-side derivation refactor is good. Just need the reporting path fully decoupled from the stats write before this can land.

yuezengwu (commented):
## CI after rebase: `Lint & Type Check` is green now ✅, but `Test` has a **real, PR-caused** failure ⚠️

The rebase onto current `main` cleared the stale-attention build errors (good). The remaining `@first-tree/web#test` failure is **not** unrelated this time — it's caused by this PR's `context.tsx` changes:

`src/pages/__tests__/context-page-dom.test.tsx` (added on main by #704, so it wasn't on your branch before the rebase) — 2/4 tests fail because they assert the **old** Context-page copy that this PR rewrote:

1. **`renders live preview ...` (line 183)** asserts `toContain("once")` and `toContain("read the context tree")`. Your `ContextSignal` now renders `"made N reads and M made K writes"` driven by `snapshot.io`, so those strings are gone.
2. **`renders empty usage and empty change map states` (line 229)** asserts `toContain("No agent has read the context tree in the last 7 days.")`. Two problems: (a) the empty-state copy is now `"No explicit Context Tree read/write recorded in the last 7 days."`, and (b) the fixture only zeroes `usage` — it inherits the non-empty `io` block from `MOCK_CONTEXT_SNAPSHOT` (read 18 / write 5), so `ContextSignal` takes the non-empty branch and the new `ContextIoByAgent` table renders anyway.

### Fix
Update `context-page-dom.test.tsx` to match the new Context-tab behavior: assert the new read/write copy, and in the empty-state case set `io` to an empty summary (`read`/`write` buckets all 0, empty `agents`/`recentEvents`) alongside the empty `usage`. This is a test the rebase merged in — worth a quick local `pnpm --filter @first-tree/web test` to confirm green before pushing.

---

Status recap — two items before this can land:
1. **(blocker, still open)** Decouple `recordFromSessionEvent` from the `appendEvent` transaction (make it best-effort) — owner confirmed stats may be dropped.
2. **(blocker, new)** Fix the `context-page-dom.test.tsx` failures above.

The implementation logic itself still LGTM.

yuezengwu (commented):
Re v4 (`52f2a5d`):

✅ **Blocker 1 resolved cleanly** — `recordFromSessionEvent` is now fire-and-forget after `appendEvent` commits (no shared transaction), failures only log a non-fatal warn. The session_event write is fully decoupled from the stats write. 👍

⚠️ **Heads-up on the latest `Test` red — it's a CI infra flake, not your test.** The job failed at the *Initialize containers* step (`Get "https://registry-1.docker.io/v2/": context deadline exceeded` — Docker image pull timed out); the vitest suites never ran. Don't chase that one. A re-run should get past it.

❗ **Blocker 2 is still open.** `context-page-dom.test.tsx` (merged onto main via #704) wasn't touched in this push — once the suite actually runs, the 2 failures from the previous run will reappear: the test still asserts the old `ContextSignal` copy (`"read the context tree"` / `"once"`) and the old empty-state string, and the empty-state fixture still inherits a non-empty `io` block from `MOCK_CONTEXT_SNAPSHOT`. Please update that test to the new copy + add an empty `io` summary to the empty-state fixture (details in my earlier comment), then re-run.

baixiaohang:
Updated the PR to address the remaining Context page test blocker from the review.

Changed `context-page-dom.test.tsx` to use the new `io.recentEvents` fixture and explicit read/write copy, and added an empty `io` summary for the empty-state fixture so it no longer inherits non-empty mock IO.

Local verification:
- `pnpm --filter @first-tree/web exec vitest run src/pages/__tests__/context-page-dom.test.tsx`
- `pnpm --filter @first-tree/web test`
- `pnpm check`
- `pnpm typecheck`

yuezengwu (approved):
## Approving ✅

Both review blockers are resolved and CI is green (Lint & Type Check / Test / Migrations / CodeQL).

- **Decoupling (blocker 1):** `recordFromSessionEvent` now runs after `appendEvent` commits, with failures swallowed to a non-fatal warn — the user-visible session_event write no longer shares a failure domain with the IO statistics. The read-path backfill (now covering `tool_call` `toolFileRefs`, not just legacy usage) is a nice safety net that re-derives any dropped stat rows from persisted session_events. 👍
- **Context page test (blocker 2):** `context-page-dom.test.tsx` updated to the new read/write copy + `io` fixtures, with an empty `io` summary for the empty-state case.

The server-side derivation refactor (client emits neutral `toolFileRef` evidence, server is sole authority on action/source via `deriveEventIo`) is a clean trust boundary, well covered by tests.

**DB change signed off by the owner (@yuezengwu):** new `context_tree_io_events` table + hand-authored migration `0055`. Acknowledged trade-offs: `source_session_event_id` intentionally has no FK (table outlives `session_events`), and `0055` is hand-authored with no `meta/0055_snapshot.json` due to the repo's pre-existing Drizzle snapshot drift (so `drizzle-kit generate` stays blocked until that drift is fixed — not introduced by this PR).

Non-blocking note for later: the on-read backfill in `summarizeContextTreeIo` runs a jsonb scan + per-row re-record on every snapshot GET; fine at steady state (NOT-EXISTS guarded), but worth revisiting if a busy org's Context tab feels slow.

LGTM — thanks for the quick iterations.

PR #760, opened 2026-06-02
feat(tree): workspace.json schema + status command (PR-2/N)

liuchao-001:
## Summary

Phase 2 of the **workspace-layout-simplification** work tracked in [first-tree-context](https://github.com/agent-team-foundation/first-tree-context). Adds the new workspace-rooted layout primitives **alongside** the legacy binding system. Purely additive — no removals.

See the spec in first-tree-context (PRs [#383](https://github.com/agent-team-foundation/first-tree-context/pull/383) and [#384](https://github.com/agent-team-foundation/first-tree-context/pull/384)): `first-tree-skill-cli/workspace-layout-simplification.md`.

## What lands

**Adds**:
- `packages/shared/src/schemas/workspace-manifest.ts` — Zod schema for `<workspace>/.first-tree/workspace.json`. Two fields: `tree` (immediate subdir name) and `sources` (string[]). Strict validation: rejects path separators, dotfiles, duplicates, tree/sources collision. No `bindingMode`, `treeMode`, `schemaVersion`.
- `apps/cli/src/core/workspace.ts` — `discoverWorkspaceRoot`, `readWorkspaceManifest`, `writeWorkspaceManifest`, `computeWorkspaceStatus`.
- `apps/cli/src/__tests__/workspace.test.ts` — 16 tests covering schema validation, walk-up discovery (including nested workspaces), drift detection, dotfile/non-git filtering.

**Rewires**:
- `apps/cli/src/commands/tree/status.ts` — prefers the new workspace.json code path when discovery finds a workspace root; falls back to the existing `inspect` flow for unmigrated workspaces.
- `apps/cli/src/commands/tree/init.ts` — after legacy `initializeSourceRoot` succeeds, ALSO writes `<sourceRoot>/.first-tree/workspace.json` when `scope === "workspace"` **AND** the resolved tree is an immediate child of sourceRoot. Forward-compat shim only — does not touch repo-scope inits or legacy sibling-tree layouts.

## What this PR does NOT do

Per the spec's PR plan:
- **PR-3**: Migration command (`migrate-to-w1`).
- **PR-4**: Shipped skill payload rewrites — must land before any command removal, otherwise in-flight managed-onboarding sessions running the old skill against the new CLI would crash on deleted subcommands.
- **PR-5**: Remove `inspect` / `bind` / `bootstrap` / `workspace sync` / `integrate` / `publish`.
- **PR-6**: Remove per-source skill install / framework block / source.json dual writes.

## Validation Gates (verified pre-PR)

[...]

PR #762, opened 2026-06-02
feat(tree): migrate-to-w1 command (PR-3/N)

liuchao-001:
## Summary

Phase 3 of the **workspace-layout-simplification** work tracked in [first-tree-context](https://github.com/agent-team-foundation/first-tree-context). Ships the one-shot migration command that converts legacy multi-mode workspaces to the W1 layout introduced in PR-2 (#760).

## What it does

Run `first-tree tree migrate-to-w1` from a workspace root or from a single source repo. The command auto-detects four cases and converges all of them on the same cleanup flow:

| Case | Detection | Pre-step |
|---|---|---|
| Already migrated | `.first-tree/workspace.json` present | no-op |
| Workspace | Legacy `.first-tree-workspace` marker, OR child dir with `.first-tree/bindings/` | none |
| Promotable source | `source.json` points at a sibling tree (no parent workspace) | prompt → `mkdir <parent>/<source>-workspace/` + `mv` both repos in |
| Not-applicable | None of the above | error out |

After convergence (or as Case A directly), the migration:

1. Writes `<workspaceRoot>/.first-tree/workspace.json` (the new manifest from PR-2).
2. Removes `.first-tree-workspace` marker.
3. Strips `.first-tree/bindings/`, `bootstrap.json`, `tree.json`, `source-repos.md` from the tree repo.
4. For each source repo, strips `.first-tree/source.json` (+ empty `.first-tree/`), `.agents/skills/`, `.claude/skills/`, `WHITEPAPER.md`, and the `<!-- BEGIN FIRST-TREE-SOURCE-INTEGRATION --> ... <!-- END --->` block from `AGENTS.md` / `CLAUDE.md`.

**No git operations are performed.** Every change is left as a dirty working-tree edit for the user to commit. `--dry-run` walks the whole plan without touching disk.

## Adds

- `apps/cli/src/core/migrate-workspace.ts` — detection + migration + promote logic (pure, testable).
- `apps/cli/src/commands/tree/migrate.ts` — CLI wrapper with `--dry-run`, `--workspace-name`, `--yes` flags. Promote step prompts via `@inquirer/prompts` unless `--yes` is passed.
- `apps/cli/src/__tests__/migrate-workspace.test.ts` — **18 tests** covering all 4 detection buckets, Case A end-to-end (manifest write, marker deletion, tree cleanup, source cleanup, framework block strip), dry-run preservation, idempotency, source dedup against fallback path, promote success / target-exists / invalid-name failure paths, and Case B end-to-end.

## NOT in this PR

- Removing any legacy commands (PR-5).
- Removing per-source skill install / framework-block writing on init (PR-6).
- Rewriting the 5 shipped SKILL.md files to stop calling soon-removed CLI commands (PR-4 — must land before PR-5).

## Test plan

- [x] `pnpm --filter ./apps/cli typecheck` — clean.
- [x] `pnpm --filter ./apps/cli exec vitest run src/__tests__/migrate-workspace.test.ts` — 18/18 pass.
- [x] `pnpm --filter ./apps/cli exec vitest run` — 502/502 pass (PR-2 baseline plus the 18 new tests).
- [x] `pnpm check` — clean for files in scope (only pre-existing tmux-session.ts info remains).
- [ ] Code reviewer signs off.


yuezengwu (approved):
LGTM — clean, well-tested, additive (pure detection/migrate logic + thin CLI wrapper). No git ops, no DB changes; `workspace.json` schema is unchanged from PR-2. 18 tests cover all four detection buckets, Case A end-to-end, dry-run preservation, idempotency, and the promote failure paths. Verified the legacy state filename `.first-tree/source.json` matches the rest of the CLI (`binding-state.ts`).

One non-blocking suggestion (not gating merge):

**`--dry-run` for the promotable-source path under-reports cleanup.** In `handlePromote` (`apps/cli/src/commands/tree/migrate.ts:109-123`), the dry-run cleanup plan is computed against `plan.newTreeRoot` / `plan.newSourceRoot` — the *post-move* paths, which don't exist on disk yet because the promote was also a dry-run. So every `tryRemove`/`stripFrameworkBlock` finds nothing and `removed`/`modified` come back empty. The user sees "would mkdir + move both repos" but **no** indication that source.json, `.agents`/`.claude` skills, WHITEPAPER, and the framework block in AGENTS.md/CLAUDE.md will also be stripped. Since dry-run is a safety-preview, it should show the real blast radius. Suggest computing the preview against the current `detection.sourceRoot` / `detection.treeRoot` (then relabel paths), so the plan reflects what the real run will strip.

Low severity because every change is left as a recoverable working-tree edit in git-tracked repos and the promote step is gated behind an explicit confirm prompt. Approving.

Minor nits (optional): CRLF AGENTS.md gets normalized to LF whole-file (not just the stripped block); `not-applicable` exits 0 in `--json` mode but 1 otherwise; the `warnings[]` field for "bound source missing on disk" is never populated by detection.

baixiaohang (changes requested):
## Review by codex-assistant - PR #762

**建议**: request_changes
- 理由: promote 路径在移动目录前没有证明这是“无父 workspace 的 sibling tree”旧布局，且 promotable dry-run 没有实际走完整 cleanup plan。

**分级**: A
- Path 初判: 含 `apps/cli/**` -> A
- 语义升级: 无

**PR 摘要**
- 解决问题: 提供 `first-tree tree migrate-to-w1`，把 legacy 多模式绑定布局迁到 W1 workspace-rooted 布局。
- 设计方案: core 层负责 detect / promote / cleanup，CLI 层提供 `--dry-run`、`--workspace-name`、`--yes`，迁移后写 workspace manifest 并清理 tree/source 侧 legacy artifact。
- 受影响模块: `apps/cli/src/core`, `apps/cli/src/commands/tree`, CLI 测试与导出面

**Review 意见**
❌ 1. `promotable-source` 检测太宽，会在错误目录下进入搬迁流程。当前只要 `source.json.tree.localPath` resolve 到一个存在的目录就返回 promotable，没有确认 tree 是 sibling、没有确认 tree 的 bindings 回指当前 source，也没有排除“当前 source 已经位于 legacy workspace parent 下”的情况。结果是用户如果在已有 legacy workspace 的 source 子目录执行命令，会被提示创建 `<parent>/<source>-workspace` 并把 source/tree 从原 workspace 里 `rename` 出去，留下原 `.first-tree-workspace` 和其他 sibling 处于破坏状态；absolute / non-sibling `localPath` 也会被同样移动。promote 入口需要先证明 B/C 的完整形状，或先向上识别已有 workspace 并走 Case A / refuse。 [R4/R5 / `apps/cli/src/core/migrate-workspace.ts:276`]

❌ 2. promotable `--dry-run` 没有 walk the whole plan。`handlePromote` 先拿到计划路径，然后把尚不存在的 `plan.newSourceRoot` / `plan.newTreeRoot` 传给 `migrateWorkspaceToW1(..., { dryRun: true })`；cleanup helper 都用 `existsSync`/`isFile` 看这些新路径，因此不会报告将要删除的 `source.json`、skills、bindings、framework block 等 legacy artifact。用户看到的是不完整的迁移计划，和 PR 描述及命令文案里的 dry-run 语义不一致。 [R4 / `apps/cli/src/commands/tree/migrate.ts:109`]

⚠️ 3. promotable path 下 `--json` 不是纯 JSON。`runMigrateCommand` 在 `context.options.json` 时仍先进入 `handlePromote`，而 `handlePromote` 会打印人类说明、dry-run 文案、甚至在未传 `--yes` 时发起 confirm prompt，最后才由 caller 打印 JSON。其他 workspace/status 路径把 JSON 作为 machine-readable 输出；这里会让脚本无法解析。 [R4 / `apps/cli/src/commands/tree/migrate.ts:101`]

✅ 4. Case A 的 manifest 写入、tree/source cleanup、idempotency 和基础 promote 成功/失败路径都有测试覆盖；建议补上 parent workspace 误判、non-sibling localPath、promotable dry-run cleanup、promotable JSON 输出这些用例。

**已执行**
- 已提交 request_changes review
- 未运行代码检查或单测（PR Review 模式禁止）

yuezengwu (commented):
Following up on the Codex review — concurring and **revising my earlier approval**. I confirmed Codex's three findings against the code; the first one I under-weighted in my initial pass and it should be fixed before merge:

**[P1, agree — should-fix before merge] Bindings/fallback are mutually exclusive, can silently drop a bound source** (`migrate-workspace.ts:259`). The fallback `source.json` scan only runs when `sourceRoots.length === 0`. But source dirs are resolved via the binding's `sourceName` field — which this same file (`:490`) explicitly acknowledges "can drift". So if one binding resolves cleanly and another's `sourceName` is stale, the still-present `web/.first-tree/source.json` is never picked up: the migration writes `workspace.json` without `web` and leaves its legacy state uncleaned. For a one-way, non-transactional migration, silently un-binding an existing source is the kind of thing worth blocking on. Suggest making detection take the **union** of bindings-derived roots and a filesystem scan of immediate children with `source.json`, deduped by resolved path (the migrate layer already dedupes source names at `:496`, so this is purely a detection-coverage fix).

**[P1, agree] Promote accepts a non-sibling tree** (`migrate-workspace.ts:280`). `siblingPathFromSourceState` will resolve an absolute or `../../elsewhere` `localPath`, and detection only checks `isDirectory`. `migrate-to-w1 --yes` would then relocate an external repo into the new workspace. Cheap guard: require `dirname(treeRoot) === dirname(sourceRoot)` (true sibling) before classifying as `promotable-source`, otherwise `not-applicable` with a clear reason.

**[P2, agree] Dry-run under-reports for promotable sources** (`migrate.ts:112`) — same as my earlier comment; compute the preview against the current `detection.sourceRoot`/`treeRoot`.

I'm leaving the approval in place since these are recoverable working-tree edits in git-tracked repos (no git ops, promote is confirm-gated) — but I'd treat the first finding as a fix-before-merge rather than optional. @liuchao-001 your call on whether to land + fast-follow or fix here.

baixiaohang:
Follow-up: keeping this as fix-before-merge, not fast-follow.

The bindings/fallback issue yuezengwu called out is valid and should be included in the same fix set: detection should take the union of bindings-derived source roots and immediate children that still have `.first-tree/source.json`, deduped by resolved path/name, so a drifted `sourceName` cannot silently drop an existing source from `workspace.json`.

Please address that together with the promote sibling guard and promotable dry-run preview before re-requesting review. The existing request_changes remains.

liuchao-001:
Addressed all four reviewers in 5576220f. Summary:

## P1 — detection / promote correctness

- **Bindings + filesystem union (Codex P1#1 + yuezengwu + baixiaohang)**: detection now always runs both scans (bindings/-derived sources AND immediate children with `.first-tree/source.json`), deduped by resolved path. New `missingFromBindings` field on the workspace detection lists names from bindings whose subdir doesn't exist; the migration result surfaces them as warnings.
- **Non-sibling tree rejection (Codex P1#2)**: `siblingTreeFromSourceState` now returns ok/reason and rejects when `dirname(treeRoot) !== dirname(sourceRoot)`. Absolute / `..`-escaping `localPath` returns `not-applicable` with a clear message.
- **Parent-is-workspace guard (baixiaohang)**: detection returns `not-applicable` when a source's parent has the `.first-tree-workspace` marker OR another source sibling with `.first-tree/source.json`. The guard intentionally does NOT trip on "parent has tree-with-bindings child" alone — a pure two-sibling layout (just source + tree) is the spec's intended Case B/C promote.

## P2 — cleanup ordering (code-reviewer)

- Old order: clean source → clean tree → remove marker → write manifest. If the manifest write threw at the end, every detection signal was already wiped. New order: write manifest → clean source → clean tree → remove marker. Marker survives any cleanup failure; re-run still classifies as workspace.
- Regression test `preserves the workspace marker when manifest write fails` plants a file at `.first-tree/` so `mkdirSync` inside `writeWorkspaceManifest` throws, then asserts marker + bindings/ survive and re-detection returns `workspace`.

## P2 — dry-run for promotable (codex + yuezengwu + code-reviewer + baixiaohang)

- Extracted `planPromotableDryRun` into core. Cleanup planning runs against the live pre-move layout (parent dir as synthetic workspace root); result's `workspaceRoot` is re-anchored to the planned post-move location. Reported paths preserve source/tree basenames so they describe the post-move shape. Regression test asserts the plan includes the same removed kinds (`source-state`, `source-skills`, `tree-bindings`, etc.) as the real run.

## JSON mode + exit codes (baixiaohang #3 + yuezengwu nit2)

- `not-applicable` sets `exitCode=1` in both human and JSON modes.
- promotable-source path in `--json` mode no longer prints prose or starts an interactive `confirm`. Without `--yes` it returns a structured refusal.

## P3 polish

- Malformed `source.json` produces "exists but could not be parsed as JSON" (code-reviewer).
- `pickTreeAmongChildren` computed once in detection (code-reviewer P2 secondary).
- `stripFrameworkBlock` has a comment explaining CRLF→LF is safe under the supported flow (yuezengwu nit1 / code-reviewer P3).

## Spec sync

Opened [first-tree-context #386](https://github.com/agent-team-foundation/first-tree-context/pull/386) to align the spec's §Migration Plan ordering + source-list union derivation with the new implementation.

## Tests

18 → 28 (+10). 512/512 full apps/cli suite pass. typecheck clean. pnpm check clean.

@yuezengwu @baixiaohang @code-reviewer please re-scan when CI is green.

yuezengwu (approved):
Re-reviewed commit `5576220f` — the fix-before-merge set is fully addressed. Approving.

Verified each finding against the diff:

1. **Bindings/fallback union (P1)** ✅ — `detectMigrationState` now always runs both the bindings scan and the filesystem `source.json` scan, deduped by resolved path via a `seen` set. A drifted `sourceName` can no longer silently drop a real source from `workspace.json`. Bindings that point at a non-existent dir are surfaced as `missingFromBindings` → migration warnings.
2. **Promote sibling guard (P1)** ✅ — `siblingTreeFromSourceState` rejects absolute / `..`-escaping `localPath` that doesn't resolve to a true sibling (`dirname(tree) === dirname(source)`).
3. **Parent-workspace guard (baixiaohang's stronger case)** ✅ — refuses to promote when the source's parent already has a `.first-tree-workspace` marker or another `source.json` sibling, telling the user to `cd` up and re-run as Case A. The comment correctly carves out the legitimate two-sibling layout.
4. **Promotable `--dry-run` preview (P2)** ✅ — new `planPromotableDryRun` synthesizes the plan against the live pre-move layout then re-anchors `workspaceRoot`, so the preview reports the real blast radius.
5. **`--json` purity (baixiaohang #3)** ✅ — `jsonOnly` is threaded through `handlePromote`; all human-readable prints are suppressed in JSON mode, and JSON-mode-without-`--yes` now refuses (can't prompt) instead of hanging. `not-applicable` exits non-zero in both modes.

Also noted: manifest write was reordered to **first** (then cleanup, marker removal last). The tradeoff is sound and tested — a mid-cleanup failure now leaves the marker in place so re-detection still hits the workspace branch, rather than wiping all detection signals before the manifest lands. Cleanup ops use `rmSync(force)` so the throw path is largely theoretical either way.

10 new tests cover exactly these paths (non-sibling abs path, malformed JSON, parent-with-sibling, parent-with-marker, drifted-sourceName fallback pickup, missing-binding warning, manifest-before-marker ordering, marker-preserved-on-write-failure, dry-run parity). CI on `5576220f` is green (Test + Lint & Type Check SUCCESS).

@baixiaohang this resolves the items behind your request_changes — over to you to dismiss it.

baixiaohang (changes requested):
## Review by codex-assistant - PR #762

**建议**: request_changes
- 理由: 最新 head 修复了前一轮 promote/dry-run 问题，但 manifest-first cleanup 顺序会在 cleanup 中途失败后把未完成迁移误判为 already migrated。

**分级**: A
- Path 初判: 含 `apps/cli/**` -> A
- 语义升级: 无

**PR 摘要**
- 解决问题: 提供 `first-tree tree migrate-to-w1`，把 legacy 多模式绑定布局迁到 W1 workspace-rooted 布局。
- 设计方案: 最新 commit 改为 bindings + filesystem union、promote sibling/parent guard、promotable dry-run preview、JSON mode purity，并调整 migration cleanup ordering。
- 受影响模块: `apps/cli/src/core/migrate-workspace.ts`, `apps/cli/src/commands/tree/migrate.ts`, migration tests

**Review 意见**
❌ 1. manifest-first 顺序会把部分清理失败的 workspace 锁成 no-op。`migrateWorkspaceToW1` 现在先写 `.first-tree/workspace.json`，再清 source/tree，最后删 `.first-tree-workspace`；注释说如果 steps 2/3 失败，marker 保留后 re-detection 还能命中 workspace branch。但 `detectMigrationState` 的第一步是只要 `workspace.json` 存在就直接返回 `already-migrated`，优先级高于 marker/tree bindings。因此只要 `cleanSourceRepo` / `cleanTreeRepo` 中任一步在 manifest 已写后抛错，重跑 `migrate-to-w1` 会打印 already migrated / nothing to do，留下未删除的 legacy source.json、bindings、framework block 等。这个路径需要保持原子语义：要么 manifest 仍最后提交；要么 detection/command 明确把 `workspace.json + legacy marker/bindings/source.json` 视为 incomplete migration 并继续 cleanup；要么引入单独的 in-progress/recovery 信号。 [R4 / `apps/cli/src/core/migrate-workspace.ts:253`, `apps/cli/src/core/migrate-workspace.ts:663`]

✅ 2. 上一轮 blockers 已看到对应修复方向：bindings/fallback 已 union；non-sibling promote 已拒绝；parent workspace guard 已加；promotable dry-run 不再基于不存在的 post-move 路径；JSON path 不再混入人类输出或 prompt。

**已执行**
- 已提交 request_changes review
- 未运行代码检查或单测（PR Review 模式禁止）

yuezengwu (changes requested):
Reversing my prior approval — baixiaohang's finding on `5576220f` is a valid blocker, and I should have caught it (I flagged the same interaction in my re-approve comment but waved it off as "theoretical"; it isn't).

**The manifest-first ordering + workspace.json-wins detection creates an incomplete-migration trap.** `migrateWorkspaceToW1` now writes `.first-tree/workspace.json` as step 1, cleans source/tree as steps 2–3, removes the marker last. But `detectMigrationState` (`migrate-workspace.ts:231`) returns `already-migrated` the moment `workspace.json` exists — higher priority than the marker/bindings branch the ordering comment relies on. So if any cleanup step throws after the manifest is written, a re-run prints "already migrated / nothing to do" and leaves the uncleaned legacy `source.json` / `bindings/` / framework blocks behind permanently.

This is reachable, not theoretical: `cleanSourceRepo`'s `stripFrameworkBlock` ends in `writeFileSync`, and the `rmSync` calls can all throw on EACCES / EBUSY (locked file, common on Windows/network mounts) / ENOSPC / EROFS. Concrete path: source A cleans fine, source B's `writeFileSync` throws → manifest written, marker present, B's framework block never stripped, re-run is a no-op.

**Recommended fix (resumable detection, baixiaohang's option b):** keep the manifest write first, but make detection treat "`workspace.json` present **and** legacy artifacts still present (`.first-tree-workspace` marker, a child `<tree>/.first-tree/bindings/`, or any child `.first-tree/source.json`)" as an *incomplete* migration that re-enters cleanup, rather than `already-migrated`. That makes `migrate-to-w1` idempotent/resumable regardless of where a partial failure lands, and it strictly dominates both orderings — it also closes the original code-reviewer concern (cleanup-then-throw wiping all detection signals) without reintroducing it. Pure `already-migrated` should require workspace.json with **no** surviving legacy artifacts.

Everything else from the prior round is confirmed fixed (bindings/fallback union, non-sibling reject, parent-workspace guard, dry-run preview, JSON purity) — this is the one remaining item. Once detection is resumable I'll re-approve.

liuchao-001:
Addressed baixiaohang's blocker (and yuezengwu's retraction confirming the same finding) in e1dea557.

## What was broken

The manifest-first ordering fix swapped one partial-failure mode for another: detection short-circuited on the manifest's presence, so if any cleanup step threw AFTER the manifest write (e.g. \`cleanSourceRepo\`'s \`stripFrameworkBlock\` ends in \`writeFileSync\` — EACCES / EBUSY / ENOSPC / EROFS are all real-world failure paths), a re-run printed "already migrated / nothing to do" and silently left uncleaned legacy artifacts behind.

## What's now fixed

**Resumable detection.** Pure \`already-migrated\` now requires the manifest AND no surviving legacy artifacts. "Legacy artifact" covers:
- \`.first-tree-workspace\` marker
- a child subdir with \`.first-tree/bindings/\`
- a child subdir with \`.first-tree/source.json\`

Any of those present alongside the manifest → routed through the \`workspace\` branch so cleanup re-runs idempotently.

**Resume support across the cleanup path.** Three follow-on fixes for the path where a prior run partially completed:
1. **Tree-root derivation**: if \`bindings/\` was cleaned in the prior run but sources/marker survive, detection falls back to the manifest's \`tree\` field to identify the tree subdir so source cleanup can still complete.
2. **Source roster preservation**: a previously-cleaned source has neither \`source.json\` (filesystem fallback can't see it) nor a matching binding entry. Detection unions in \`sources\` from the existing manifest so the resumed manifest re-write preserves the full roster — otherwise resumed runs would lose previously-cleaned sources from the manifest.
3. **Parent-guard extension**: promote-source parent guard also rejects when the parent has \`.first-tree/workspace.json\`, matching the marker / other-source-sibling checks.

## Tests

28 → 32 (+4):
- \`re-detects 'workspace' (not 'already-migrated') when manifest + legacy artifacts coexist\`
- \`finishes cleanup on a resumed migration without losing already-cleaned sources\` (end-to-end resume: api cleaned in prior round; manifest still lists [api, web]; resume completes web; final manifest preserved [api, web]; re-run after = true \`already-migrated\`)
- \`falls back to manifest.tree when bindings/ are already gone\`
- \`'not-applicable' when running inside a source whose parent has a workspace.json\`

32/32 migrate suite + 516/516 full apps/cli suite + typecheck + biome all green.

@yuezengwu @baixiaohang please re-scan when CI clears.

yuezengwu (approved):
Approving. The resumability blocker is resolved and the lone CI red is an unrelated, pre-existing main breakage — details below.

## Resumability fix (my prior request_changes) — resolved ✅

Verified in `migrate-workspace.ts` on the current head: pure `already-migrated` now requires the manifest AND no surviving legacy artifact (`.first-tree-workspace` marker, a child `.first-tree/bindings/`, or a child `.first-tree/source.json`); any of those alongside the manifest re-enters the `workspace` branch so cleanup runs idempotently. The resume path also (a) falls back to `manifest.tree` when `bindings/` was already cleaned, and (b) unions in `manifest.sources` so a partial-failure re-run doesn't drop already-cleaned sources from the roster. Parent-workspace promote guard extended to `parentHasManifest`. The migration suite (32 tests) passes **in CI** on this head.

## CI red is NOT this PR — it's #763 breaking main

The only failing check is `Test`, and the only failing test is:

```
src/__tests__/skill-artifacts.test.ts > … > references/agent-communication.md teaches `chat invite` instead of the retired --direct escape hatch
  → expected '# Agent-to-Agent Communication …' not to match /--direct/
```

This is unrelated to PR #762:

- #762's diff touches neither `skill-artifacts.test.ts` nor `skills/first-tree-cloud/references/agent-communication.md` (empty `git diff --stat` for both).
- The assertion is a blunt `not.toMatch(/--direct/)`; the doc mentions `--direct` (at line 82, in prose describing it as retired), which trips it.
- **main itself is currently red for exactly this reason**: the latest CI run on `main` (`ac1ddbb4`) is a failure, and that same commit — #763 "sink Communication Rules into first-tree-cloud skill (proposal P3)" — is what introduced the `--direct` mention into the doc. This branch rebased onto that broken main and inherited the failure.

So this red gates the merge button mechanically but is not attributable to this PR, and merging main won't clear it — the fix belongs with #763 / the first-tree-cloud skill-docs owner (either drop `--direct` from `agent-communication.md` or loosen the test to allow describing a retired flag). Flagging rather than fixing, since it's outside this PR's scope and another owner's call.

Signing off on the migration code. Recommend landing once main is green again (the #763 breakage is fixed and this branch picks it up).

baixiaohang (approved):
## Review by codex-assistant - PR #762

**建议**: approve
- 理由: 当前 head `6d33945c` 已解决前几轮 migration blocker；剩余 `Test` 红是 main/#763 引入的 unrelated skill-docs 断言问题，不归因于本 PR。

**分级**: A
- Path 初判: 含 `apps/cli/**` -> A
- 语义升级: 无

**PR 摘要**
- 解决问题: 提供 `first-tree tree migrate-to-w1`，把 legacy 多模式绑定布局迁到 W1 workspace-rooted 布局。
- 设计方案: core 层实现 detection / promote / cleanup；CLI 层提供 `--dry-run`、`--workspace-name`、`--yes`；最新 head 增加 resumable detection，让 `workspace.json` 与 legacy artifacts 并存时继续 cleanup。
- 受影响模块: `apps/cli/src/core/migrate-workspace.ts`, `apps/cli/src/commands/tree/migrate.ts`, migration tests

**Review 意见**
✅ 1. 上轮 blocker 已修：pure `already-migrated` 需要 manifest 存在且无 `.first-tree-workspace`、tree bindings、child source.json；部分迁移会重新进入 workspace cleanup。[R4]
✅ 2. Resume path 保留 source roster：bindings 已清理或某些 source 已清理时，会从 manifest tree/sources 补回恢复所需信息，避免重写 manifest 时丢 source。[R1/R4]
✅ 3. 前几轮 blockers 仍保持修复：bindings/filesystem union、non-sibling promote reject、parent workspace guard、promotable dry-run preview、JSON purity。
⚠️ 4. 当前 `Test` check failure 来自 main/#763 的 `skill-artifacts.test.ts` / `first-tree-cloud` skill-docs 断言，不是本 PR diff 引入；按授权不阻塞本次 approve。

**已执行**
- 已按 baixiaohang 明确授权提交 approve
- 未运行代码检查或单测（PR Review 模式禁止）

PR #783, opened 2026-06-02
Track Context Tree shell reads

baixiaohang:
## Summary
- add a conservative shared shell IO classifier for explicit Context Tree reads
- attach Context Tree repo refs for Codex `command` and Claude Code/TUI `Bash` reads when paths resolve under the local Context Tree checkout
- derive `source=shell_command` IO rows on the server with skip-reason diagnostics while preserving existing direct write tracking

## Scope
- shell read tracking only
- no shell write tracking
- no `resultPreview` inference
- no DB migration

## Verification
- `pnpm --filter @first-tree/shared exec vitest run src/__tests__/shell-command-io.test.ts --poolOptions.threads.maxThreads=2 --poolOptions.threads.minThreads=1`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/context-tree-file-refs.test.ts src/__tests__/codex-context-tree-io.test.ts src/__tests__/claude-code-tool-events.test.ts --poolOptions.threads.maxThreads=2 --poolOptions.threads.minThreads=1`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/context-tree-io.test.ts src/__tests__/ws-session-event.test.ts --poolOptions.threads.maxThreads=2 --poolOptions.threads.minThreads=1`
- `pnpm check`
- `pnpm typecheck`

## Review
- codex-reviewer reviewed the worktree and found two shell-classifier blockers
- fixed `find` mutation primaries and unquoted shell comments
- reviewer re-reviewed and found no remaining blockers

### DIFF
--- a/packages/client/src/__tests__/claude-code-tool-events.test.ts
+++ b/packages/client/src/__tests__/claude-code-tool-events.test.ts
@@ -412,17 +412,52 @@ describe("createToolCallProcessor — Context Tree file refs", () => {
     expect(usageEventCount(emit)).toBe(0);
   });
 
-  it("does NOT attach file refs for unsupported tools even if an arg path is under the tree", () => {
+  it("attaches file refs when Bash reads a Context Tree file", () => {
     const emit = vi.fn<(event: SessionEvent) => void>();
-    const processor = createToolCallProcessor(emit, binding);
+    const processor = createToolCallProcessor(emit, { ...binding, branch: "main" }, { cwd: "/home/op/project" });
 
-    processor.onMessage(
-      assistantToolUse("r7", "Bash", { command: `cat ${TREE}/NODE.md`, file_path: `${TREE}/NODE.md` }),
-    );
+    processor.onMessage(assistantToolUse("r7", "Bash", { command: `cat ${TREE}/NODE.md` }));
     processor.onMessage(userToolResult("r7", "contents"));
 
     const final = toolCallEvents(emit).find(
       (event) => event.payload.toolUseId === "r7" && event.payload.status === "ok",
+    );
+    expect(final?.payload.toolFileRefs).toEqual([
+      {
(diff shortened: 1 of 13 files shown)


yuezengwu (approved):
LGTM — approving. Reviewed the shared classifier, both client handlers, the server derivation, the DB schema and CI (all green).

**What this does:** best-effort tracking of Context Tree *reads* performed via shell (`codex command` / `claude Bash`), complementing the existing direct Read/Write tracking. The `classifyShellCommandIo` classifier is appropriately conservative — it rejects compound shell (`| ; & < >`), dynamic syntax (`$`, backtick, `~`, globs), mutating tools/options (`sed -i`, `find -delete/-exec`, `tee`, write redirects) and comment lines, and falls back to "no refs" rather than crashing. Path-traversal has defense in depth (client `pathInsideContextTree` + server `normalizeTargetPath`). No DB migration is correct: `source` is a plain `text` column with no CHECK constraint. Good end-to-end coverage including the new ws-session-event codex frame test.

Two non-blocking notes:

1. **(verify)** The server tests and codex test fixtures assume `command_execution.command` is the *bare* command (e.g. `sed -n '1,120p' …`). If real codex ever surfaces it as a `bash -lc "…"` wrapper instead, `bash` hits `MUTATING_OR_AMBIGUOUS_TOOLS` and every codex shell read silently stops tracking (graceful — matches the prior P1 gap, no crash). codex-reviewer validating its own output shape raises confidence, but worth a quick confirmation against a live codex session before relying on the codex numbers.

2. **(minor)** `metadata.localPath` persists the agent's absolute local filesystem path durably in `context_tree_io_events`. It's internal-only today (not selected by `summarizeContextTreeIo`, not surfaced in web), so low risk — just flagging that it's stored long-lived in case that matters later.

Neither blocks merge.

PR #723, opened 2026-06-01
Disable legacy MCP config writes

baixiaohang:
## Summary
- Web/CLI/Server close the legacy `mcpServers` write entry points.
- Runtime consumption of `payload.mcpServers` is preserved for existing resolved config paths.
- Skill discovery/upload reporting remains enabled; daemon now skips stale local aliases that are not pinned to this client before uploading skills.
- `GET/PATCH /api/v1/agents/:uuid/skills` stays available.

## Tests
- `pnpm --dir packages/server exec vitest run src/__tests__/admin-agent-config.test.ts src/__tests__/agent-skills.test.ts`
- `pnpm --dir packages/web exec vitest run src/pages/__tests__/page-ssr-smoke.test.tsx src/pages/__tests__/web-dom-interactions.test.tsx src/pages/agent-detail/__tests__/use-config-draft.test.tsx`
- `pnpm --dir apps/cli exec vitest run src/__tests__/agent-config-cli.test.ts`
- `pnpm --dir apps/cli exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/runtime-provider-reconcile.test.ts src/__tests__/agent-config-cli.test.ts`
- `pnpm --dir packages/server exec vitest run src/__tests__/agent-skills.test.ts`
- `pnpm exec biome check apps/cli/src/__tests__/agent-config-cli.test.ts apps/cli/src/commands/agent/config/add-mcp.ts packages/server/src/__tests__/admin-agent-config.test.ts packages/server/src/services/config-service.ts packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx packages/web/src/pages/agent-detail/__tests__/use-config-draft.test.tsx packages/web/src/pages/agent-detail/tools-tab.tsx packages/web/src/pages/agent-detail/use-config-draft.ts`
- `pnpm exec biome check apps/cli/src/core/runtime-provider-reconcile.ts apps/cli/src/core/index.ts apps/cli/src/commands/daemon/start.ts apps/cli/src/__tests__/runtime-provider-reconcile.test.ts apps/cli/src/__tests__/daemon-start-command.test.ts`

## Reviewer note
- codex-reviewer has reviewed and approved the original MCP write-shielding change.
- The only noted behavior is that an expired `expectedVersion` can return 409 before the MCP-specific rejection; this is acceptable optimistic-lock behavior.
[...]

PR #787, opened 2026-06-03
Add Resources Phase 1

baixiaohang:
## Summary
- Add shared Resource schemas, Drizzle resources tables/migration, resource service/resolver, backfill, Class C Team Resource routes, Agent Resource routes, and runtime config resolution.
- Move repo/prompt/skill/MCP resource materialization into the resource model, including runtime skill files and effective prompt/repo resolution.
- Add Web API wrappers plus Team Resources, Agent Detail Resources, onboarding, and CLI compatibility updates.

## Reviewer risk coverage
- Private agent-scoped resources reuse agent visible/manage access and return 404 outside the visible agent scope.
- Inline prompt replace supports `resource_id = null`, `replaces_resource_id` pointing at a Team prompt, and non-empty inline body.
- Canonical repo de-dup normalizes HTTPS, SSH, and scp-like GitHub URLs during backfill.
- Promote rewrites equivalent agent-scoped repo bindings and retires duplicates in one transaction with agent config version bumps.
- Effective prompt append budget remains 32,000 characters.
- Team resource detail/update/delete/promote/usage APIs use Class C `/api/v1/resources/:resourceId...` routes.

## Tests
- `pnpm exec biome check apps/cli/src packages/client/src packages/server/src packages/shared/src packages/web/src`
- `pnpm --filter @first-tree/shared test -- src/__tests__/resource-schema.test.ts`
- `pnpm --dir packages/server exec vitest run src/__tests__/admin-agent-config.test.ts src/__tests__/resources-phase1.test.ts`
- `pnpm --dir packages/client exec vitest run src/__tests__/resource-skills.test.ts`
- `pnpm --dir packages/web exec vitest run src/pages/agent-detail/__tests__ src/pages/__tests__/page-ssr-smoke.test.tsx`
- `pnpm --dir packages/web exec vitest run src/api/__tests__/api-wrappers.test.ts`
- `pnpm --dir apps/cli exec vitest run src/__tests__/agent-config-fetchers.test.ts src/__tests__/agent-config-commands-extra.test.ts`


yuezengwu (commented):
## Review — Resources Phase 1

Thorough pass on the data model, migration, service layer, access control, client runtime, and CLI. Access control is solid (404-not-403 for private agent resources, org isolation on every query, Class C routes wired correctly, no IDOR found). The effective-resolution + prompt-budget logic is correct. A few things to address before merge.

### Blocker (CI is red)

**1. Missing `packages/server/drizzle/meta/0057_snapshot.json`.** The migration was hand-authored — `0057_resources_phase1.sql`, `LATEST`, and `_journal.json` were updated, but the matching drizzle snapshot was not added. `scripts/check-migrations.mjs` (`validateSnapshotLinearity`) requires a snapshot for the latest journal entry, so the **Migrations** job fails. Add a `0057_snapshot.json` reflecting the two new tables so future `db:generate` diffs from current state instead of `0056`. (Note: this hand-authored path is at odds with the repo's "never hand-edit Drizzle migrations" rule; the partial unique indexes are the reason, which is fine — but the snapshot still has to exist.)

### Should-fix (correctness)

**2. Optimistic-concurrency is TOCTOU-racy.** `replaceAgentResources` (`services/resources.ts:851-856`) reads and compares `expectedVersion` *outside* the transaction, and the tx (`:858-885`) never re-checks it. Two concurrent edits submitted with the same `expectedVersion` both pass the guard and both commit — last-writer-wins, one set of bindings is silently lost and the version jumps `+2`. That is exactly what `expectedVersion` exists to prevent. Move the check inside the tx as a conditional `UPDATE ... WHERE version = expectedVersion` (`rowCount === 0` → `ConflictError`) or `SELECT ... FOR UPDATE` the config row.

**3. `append-prompt` silently changed from replace to accumulate.** The command still describes itself as *"Replace the systemPrompt append text"* (`append-prompt.ts:10`), but the new impl pushes a fresh inline-prompt binding on every call (`:29-42`). Repeated invocations now stack prompts and eventually trip the 32k `applyPromptBudget`, dropping content as `prompt_budget_exceeded`. Restore replace semantics (reuse/overwrite the existing inline binding) or rename + redoc the command.

**4. `add-repo` dedup regression.** New dedup only matches bindings whose `resourceId` is in `availableTeamResources` (`add-repo.ts:17-22`), but a repo added by a prior `add-repo` round-trips as an **agent-scoped** resource, which `availableTeamResources` excludes (`scope:"team"` only). So re-running `add-repo <same-url>` to change `--ref`/`--path` stacks a duplicate binding; the second resolves to `unavailable: duplicate_local_path` and the updated ref/path is silently discarded — the opposite of the old replace behavior.

**5. `updateResource` doesn't map unique-collision to 409.** Editing a team repo's `url` recomputes `repoCanonicalKey` (`resources.ts:679`); a collision violates `uq_resources_team_repo_canonical_active` and throws PG 23505. `createTeamResource` catches this → `ConflictError`, but `updateResource` (`:670-703`) doesn't, so it surfaces as a 500 instead of a clean 409.

**6. Stale runtime skill files are never cleaned up.** `materializeResourceSkills` (`runtime/resource-skills.ts`) only `mkdir`+`writeFile`; nothing prunes dirs for skills that were disabled/removed. Orphaned `SKILL.md` files persist under `.first-tree/resources/skills/<id>/`. Limited blast radius (that dir isn't auto-discovered as a Claude skill), so it's a disk leak rather than a ghost-skill-in-prompt — but it should reconcile (manifest + prune unlisted dirs).

### Low / nits

**7. CodeQL ReDoS (code-scanning alert 33), `resource.ts:391-392`.** `buildCanonicalRepoKey` runs `/\/+$/`-style replaces on `repoUrlSchema` input, which has **no max length** — polynomial worst-case on many-slash strings. Low practical risk given the validated URL shape, but trivial to fix; please resolve the alert.

**8. Backfill is non-transactional and re-runs on every boot** (`resources-migration.ts`, fired from `app.ts` `onReady`). Idempotency rests on find-before-insert with no surrounding tx; concurrent boots (rolling deploy) could race the unique index. Per-repo errors are only logged, so it's non-fatal, but confirm single-instance boot or the "created" counts are unreliable.

**9. YAML frontmatter injection** in `buildSkillMarkdown` — raw `name`/`description` interpolation with no charset restriction; quote the scalars. Low impact.

Will re-review once the snapshot lands (CI green) and the should-fix items are addressed or consciously deferred.

yuezengwu:
Re-reviewed `b2ea6634`. Confirmed resolved:
- **Blocker 1** — `0057_snapshot.json` added, **Migrations** job is green again. ✅
- The two partial unique indexes are now mirrored in the Drizzle model (`db/schema/resources.ts`), so the model no longer drifts from the hand-authored SQL. 👍
- **Finding 7 (ReDoS)** — `canonicalizeResourceRepoUrl` rewritten to manual character-scan parsing (no regex on the unbounded URL). ✅

Still open from the review above (none touched by this commit) — the silent behavior regressions are the ones I'd most want addressed or consciously deferred before merge:
- **Finding 2** — optimistic-concurrency TOCTOU in `replaceAgentResources` (version checked outside the tx → lost updates).
- **Finding 3** — `append-prompt` now accumulates instead of replacing (description still says "Replace").
- **Finding 4** — `add-repo` dedup no longer matches agent-scoped repos → re-running stacks a duplicate binding and silently drops the new ref/path.
- **Finding 5** — `updateResource` unmapped 23505 → 500 instead of 409.
- **Finding 6** — stale runtime skill dirs never pruned.

_(Edited to de-link stray `#N` auto-references that pointed at unrelated issues — apologies for the noise.)_

yuezengwu (approved):
Approving. ✅

All review findings (Blocker 1 + Findings 2–9) were addressed with tests, and the second-pass "rereview blockers" commit landed real fixes too (impact-preview now simulates the proposed payload, backfill is idempotent via a per-org completion marker, `source_repos` org-setting writes/deletes now return 410). CI is fully green on `c33a5d15` — Migrations, Lint, Test, and CodeQL (the ReDoS alert is resolved).

Schema necessity reviewed column-by-column with the table owner: both `resources` and `agent_resource_bindings` are sound and every field/index is actively used, with one non-blocking note — `agent_resource_bindings.organization_id` is write-only (never read in any query; org isolation goes through `agent_id → agent.organizationId`). Kept intentionally for shape-consistency with `resources` and a defensive org-delete cascade. Fine to leave; could be dropped in a later cleanup if you ever want to trim it.

Nice responsiveness on the iteration. 👍

Gandy2025:
Took a pass from a product/usage angle (the technical-correctness line already approved by @yuezengwu I reconciled and agree with). Two backend items I'd like to flag for confirmation before merge — both about the runtime-resolution path, not the UI:

### 1. [High — needs confirmation] MCP isn't backfilled, but `resolveRuntimeConfig` overwrites `mcpServers` wholesale

`resolveRuntimeConfig` (`packages/server/src/services/resources.ts:633`) runs on the live runtime-config fetch path for every agent (`api/agents-config.ts:15,22`, `api/agent/config.ts:16`) and **replaces** (not merges) `gitRepos` / `prompt.append` / `mcpServers` / `resourceSkills` with the resource-resolution result.

The backfill (`packages/server/src/services/resources-migration.ts`) migrates org `source_repos`, agent `gitRepos`, and agent `prompt.append` — but does **not** touch `mcpServers`. Legacy per-agent MCP writes were disabled in #723 (the disable didn't clear already-stored values), and the resource tables are new in `0057`. So any agent that still carries `mcpServers` in its stored config payload would have `mcpServers` resolved to `[]` at runtime (`runtimeMcp` returns `[]` for an empty effective set) — i.e. MCP silently dropped. I didn't find a test asserting MCP survival through `resolveRuntimeConfig`.

Could you confirm whether any existing agent config (prod/staging) still carries `mcpServers`? If yes, this probably needs an MCP backfill (or a resolve-time fallback that preserves un-migrated MCP). If no agent carries it, this is moot — just want to make sure it's a conscious call.

### 2. [Medium] `add-repo` against a team-resource URL silently converts the team binding into an agent-scoped extra

In `apps/cli/src/commands/agent/config/add-repo.ts:35-56`, when the incoming URL canonically matches a team resource the agent currently binds, the matching team binding is removed and replaced with a fresh `agentExtraRepo` binding. The repo stays present, but its identity changes from team-managed (inherits team ref/defaults, receives future team-side changes) to an agent-private extra — so subsequent team-side updates to that repo no longer flow to this agent. `--help` (`:10`) doesn't mention this. Is this an intended override path or a regression? If intended, a line in `--help` / a CLI hint would help; if not, the team binding should probably be preserved.

For context, the earlier CLI items (`append-prompt` replace semantics, `add-repo` dedup) I confirmed are fixed with tests. Thanks!

yuezengwu:
@Gandy2025 thanks — I reconciled both against the code and **item 1 is a real gap I missed in the correctness pass.** Confirming with evidence so the record is clear:

**Item 1 (MCP silent-drop) — confirmed.**
- `resolveRuntimeConfig` (`services/resources.ts:646`) sets `mcpServers: runtimeMcp(effective.mcp)` unconditionally; the `...config.payload` spread is overridden by that explicit key, so the stored value is discarded, not merged.
- The backfill never creates MCP resources — the only `mcp` token in `resources-migration.ts` is the type annotation at `:341`, no migration logic.
- Stored legacy values persist: `config-service.ts:126` keeps `current.mcpServers` on unrelated patches, and `:137` only rejects *new* MCP writes (#723's disable). So "disabled-but-not-cleared" configs are exactly the at-risk set.
- No test asserts MCP survival through `resolveRuntimeConfig`.

So for any agent whose stored payload still carries `mcpServers`, the live config fetch now resolves it to `[]`. **Recommendation:** rather than gate this on a prod/staging DB inspection, add a resolve-time fallback that's safe regardless of DB state — when `effective.mcp` is empty *and* `config.payload.mcpServers` is non-empty, preserve the stored value (or, equivalently, backfill MCP into the resource tables). Either way please add a regression test that an agent carrying legacy `mcpServers` keeps them after `resolveRuntimeConfig`. I'd treat this as a merge-blocker until the fallback/backfill lands or an operator confirms no agent carries `mcpServers`.

**Item 2 (`add-repo` team→agent conversion) — agree it's worth pinning down.** The dedup fix I confirmed earlier intentionally removes the matching team binding and re-adds the URL as an `agentExtraRepo` (`add-repo.ts:35-56`), which does flip the repo from team-managed to agent-private and severs future team-side updates. Reasonable as an explicit override, but it shouldn't be silent: a `--help` line, and ideally a no-op when the URL already matches an enabled team binding with unchanged `--ref`/`--path` (only convert when the user is actually overriding ref/path). Author's call on intent.

Given item 1, I'd hold the merge until it's resolved. Happy to re-review the fallback quickly.

Gandy2025 (approved):
Approving. The two backend items I raised earlier (MCP silent-drop through `resolveRuntimeConfig`, and `add-repo` team→agent-extra conversion) are tracked in the comment thread for the author to resolve/confirm — see the discussion above. Thanks for the thorough iteration.

yuezengwu (approved):
Re-approving on `f7c86d02` — the hold I placed on the MCP item is lifted.

Both backend items @Gandy2025 raised are now resolved with tests:
- **MCP silent-drop:** `resolveRuntimeConfig` now preserves stored `mcpServers` when no Team MCP resources are configured (`effective.mcp.length === 0 && config.payload.mcpServers.length > 0 ? config.payload.mcpServers : resolvedMcp`), and a Team MCP resource cleanly takes over once present. The new test asserts both directions. This is the safe, DB-state-independent fix.
- **`add-repo` team→agent conversion:** matching a Team repo URL now re-binds to the Team `resourceId` (overriding ref/path only) instead of converting to an agent-private extra, so team-managed identity and future team-side updates are preserved. Tested.

Minor non-blocking note: the fallback resurrects legacy `mcpServers` if a previously-resource-configured agent later disables its last MCP resource while a stale legacy value remains in the payload — a narrow transitional edge, fine for Phase 1.

CI: Migrations/Lint green; the Test job (which now includes the MCP-survival test) was still finishing at approval time — worth a glance before merge.

PR #600, opened 2026-05-27
refactor(codex): turn-level retry + race detection + usage metrics

yuezengwu:
## Summary

Three handler-scoped resilience improvements to the Codex SDK integration, all confined to `packages/client/src/handlers/codex.ts` (plus a new test file). No changes to `@first-tree/shared` schemas, `package.json`, `@openai/codex-sdk` version, or any other handler — the diff is fully self-contained.

Spun out of an architectural review of the project's Codex integration; addresses the P1-P2 items that fit inside the handler. P0 (SDK 0.125 → 0.134 upgrade) and P3 (per-agent `reasoningEffort` / `webSearchEnabled` config) are left for follow-up PRs because they require touching `package.json` and the shared schema respectively.

### 1. Turn-level retry on transient failures

`runTurn` now wraps the `runStreamed` + `for-await` loop in a bounded retry: 1 initial attempt + up to 2 retries with `500ms × 3^n` backoff. The retry gate has two conditions, both required:

1. The error message looks transient (HTTP 5xx, rate limit, fetch failed, ECONN*, ETIMEDOUT, overloaded, etc.) — see `isTransientCodexErrorMessage`.
2. **No user-visible item has been emitted yet** for this turn (`agent_message`, `command_execution`, `file_change`, `mcp_tool_call`, `web_search`, `todo_list`). Once the chat timeline has received content, re-running the turn would double-render it.

The classifier **short-circuits on auth / sandbox / context-length keywords** even when the message also contains a transient one. A message like `"401 unauthorized after fetch failed"` does NOT get retried — credentials are bad, not the network.

Abort handling: `sleepWithAbort` rejects on signal, so suspend/shutdown during the backoff window unwinds cleanly without burning the remaining retry budget.

### 2. AGENTS.md concurrent-write detector (proposal §⓪.3 race window)

Codex CLI reads `AGENTS.md` once at thread startup; the handler rewrites it on every `start`/`resume` because Codex has no per-turn prompt-injection API (claude-code's `systemPrompt.append` equivalent). Two chats hitting bootstrap for the same agent home within a tight window means the second writer almost certainly clobbered the first briefing before codex CLI read it — "wrong chat context surfaces in codex" bugs trace here.

[...]

PR #788, opened 2026-06-03
fix: avoid late Codex retry abort

baixiaohang:
## Summary

- Stop aborting the per-attempt Codex spawn signal after the SDK event iterator has already closed during retry cleanup.
- Keep parent suspend/shutdown cancellation intact through the active-attempt listener and parent-signal backoff sleep.
- Add a focused regression test that retries a transient stream failure without a late post-close abort, then verifies final forwarding, `turn_end:success`, and `markCompleted(1)` semantics.

## Review

- Reviewed by `codex-reviewer` in the workspace chat.
- Initial review found no blocking issues and suggested one comment cleanup.
- Follow-up review passed after the comment was updated.

## Testing

- `pnpm --filter @first-tree/client exec vitest run --passWithNoTests --maxWorkers=2 codex-retry-abort codex-resilience`
- `pnpm --filter first-tree-dev build`
- `pnpm check` (exit 0; reports one existing info-level style suggestion in untouched `packages/client/src/handlers/claude-code-tui/tmux-session.ts`)

### DIFF
--- a/packages/client/src/__tests__/codex-resilience.test.ts
+++ b/packages/client/src/__tests__/codex-resilience.test.ts
@@ -25,10 +25,9 @@ const usage = (input: number, cached: number, output: number, reasoning = 0) => 
  *     between tests — otherwise leaked state from one test contaminates
  *     the next.
  *
- * The full `runTurn` retry loop (mock Codex + Thread + SessionContext) is
- * intentionally NOT covered here — testing it end-to-end requires faking
- * the bootstrap / git-mirror / first-tree integration chain, which would
- * double the PR scope. The retry path's correctness rests on:
+ * The full `runTurn` retry loop needs a mock Codex + Thread + SessionContext,
+ * so the targeted regression for that state machine lives in
+ * `codex-retry-abort.test.ts`. The classifier path's correctness here rests on:
  *   1. typecheck (covers `usageBox` narrowing + control flow)
  *   2. `isTransientCodexErrorMessage` table (this file)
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
Approved. Independently verified the root cause against `@openai/codex-sdk@0.134.0` source.

**Why the fix is correct:** breaking out of the `for await (streamed.events)` loop calls `.return()` on the SDK iterator, which propagates through `runStreamedInternal` into the inner exec generator and runs its `finally`: `rl.close()` + `child.removeAllListeners()` + `child.kill()`. So **iterator close already tears the child down** — the explicit `attemptAbort.abort()` was redundant for teardown. Worse, since `removeAllListeners()` has already detached the child's `'error'` handler by then, a late `abort()` makes Node's `spawn({ signal })` emit an unhandled `AbortError` on the child → client crash. The PR #600 concern about a "background generator draining the dead stream" is actually covered by that SDK `finally`, so dropping the post-close abort is safe.

**Parent suspend/shutdown intact:** the `onParentAbort` listener still cascades into `attemptAbort` mid-stream (child `'error'` handler still attached there, caught via `spawnError` and swallowed by the `abort.signal.aborted` guard), and the backoff sleep still races on the parent signal.

**Test:** `codex-retry-abort.test.ts` is a faithful regression — the mock generator's `finally` flags `streamClosedByAttempt`, and the abort listener trips `lateAbortAfterClose` only if abort fires after close; asserting it stays `false` pins exactly this bug.

No core data-structure or DB changes. CI green (Test / Lint & Type Check / CodeQL all pass).

PR #792, opened 2026-06-03
fix(web): stop tailwind-merge eating button text colors at sm/xs

Gandy2025:
## Problem

On the agent-detail **Save bar**, the **Save** button rendered near-black background + **dark-gray text** — "Save" was nearly unreadable (reported with screenshot). The button's *background* was correct; its *text color* was being silently dropped.

## Root cause (not a save-bar mistake — a systemic `cn()` bug)

The design system's typography scale (`text-eyebrow … text-display`) ships as Tailwind v4 `@theme` **font-size** utilities. `cn()` used a bare `twMerge`, whose default config doesn't know these custom names and **misclassifies them as text-COLOR utilities**.

So when a class string carries both a color (`text-primary-foreground`) and a size (`text-label`, appended by `Button size="sm"/"xs"`), `twMerge` thought they conflicted and **dropped the earlier color class**. The button then inherited the page's dark `--fg` → dark-on-dark.

Verified with the repo's own `tailwind-merge`: at `size="sm"`/`"xs"`, **every filled variant** (`default` / `cta` / `destructive` / `secondary`) lost its text color. `size="default"` was unaffected (no `text-*` size class). So this hit small filled buttons app-wide, not just the Save bar.

## Fix

Root-cause, one change: `cn()` now uses `extendTailwindMerge` to register the custom scale in the **`font-size`** group, so size and color land in separate conflict groups and both survive. Zero hardcoded colors/sizes — fully token-driven, DESIGN.md-compliant. The save bar itself is untouched (its variant/hierarchy choices were already correct).

Adds `button-variants` regression tests that fail if `cn()` is reverted to a bare `twMerge`.

## Verification

- ✅ `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) — clean
- ✅ New regression tests — 12/12 pass
- ✅ Full web suite — 500/500 pass (no regressions from this shared-util change)

### DIFF
--- a/packages/web/src/components/ui/__tests__/button-variants.test.ts
+++ b/packages/web/src/components/ui/__tests__/button-variants.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "vitest";
+import { cn } from "../../../lib/utils.js";
+import { buttonVariants } from "../button.js";
+
+/**
+ * Regression guard for the "dark-on-dark Save button" bug.
+ *
+ * The design system's typography scale (text-label / text-body / …) ships as
+ * Tailwind v4 @theme font-size utilities. tailwind-merge's *default* config
+ * misclassifies those custom names as text-COLOR utilities, so a class string
+ * carrying both a color (text-primary-foreground) and a size (text-label, added
+ * by Button size="sm"/"xs") would have its color silently dropped — leaving small
+ * filled buttons inheriting --fg (dark text) on a dark bg. `cn()` registers the
+ * scale in the font-size group to keep the two in separate conflict groups.
+ *
+ * If someone reverts `cn` back to a bare twMerge, these assertions fail.
+ */
+describe("cn — custom typography scale does not eat text colors", () => {
+  it("keeps the text color when a custom font-size class follows it", () => {
+    const out = cn("text-primary-foreground", "text-label");
+    expect(out).toContain("text-primary-foreground");
+    expect(out).toContain("text-label");
+  });
+
+  it("still resolves two competing custom font-sizes to the last one", () => {
+    const out = cn("text-body", "text-label");
+    expect(out).toContain("text-label");
+    expect(out).not.toContain("text-body");
+  });
+});
+
+describe("buttonVariants — filled variants keep their text color at every size", () => {
+  // Every filled variant pairs a bg with an explicit on-color; small sizes append
+  // a custom font-size class (text-label) that must not strip that on-color.
+  const filled: Array<[string, string]> = [
+    ["default", "text-primary-foreground"],
+    ["destructive", "text-destructive-foreground"],
+    ["secondary", "text-secondary-foreground"],
+  ];
+
+  for (const [variant, expectedColor] of filled) {
+    for (const size of ["sm", "xs", "default"] as const) {
+      it(`${variant} @ ${size} retains ${expectedColor}`, () => {
+        const out = buttonVariants({ variant: variant as never, size });
+        expect(out).toContain(expectedColor);
+      });
+    }
+  }
+
+  it("cta @ sm retains its on-vivid text color", () => {
+    const out = buttonVariants({ variant: "cta", size: "sm" });
+    expect(out).toContain("text-[color:var(--fg-on-vivid)]");
+  });
+});
--- a/packages/web/src/lib/utils.ts
+++ b/packages/web/src/lib/utils.ts
@@ -1,5 +1,37 @@
 import { type ClassValue, clsx } from "clsx";
-import { twMerge } from "tailwind-merge";
+import { extendTailwindMerge } from "tailwind-merge";
+
+// The design system defines its typography scale (text-eyebrow … text-display)
+// as Tailwind v4 @theme font-size utilities — see the `@theme inline` block in
+// index.css. tailwind-merge's default config doesn't know these custom names, so
+// it misclassifies them as text-COLOR utilities. When a class string carries both
+// a color (e.g. `text-primary-foreground`) and a size (e.g. `text-label`, added by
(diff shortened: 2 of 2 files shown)


yuezengwu (commented):
Reviewed the diff and verified the mechanism end-to-end locally (tailwind-merge 3.5.0):

- Reproduced the bug with a bare `twMerge`: `text-body bg-primary text-primary-foreground text-label` → `bg-primary text-label` (both `text-body` and `text-primary-foreground` silently dropped). ✅ root cause confirmed.
- With the `extendTailwindMerge` fix, size and color land in separate conflict groups and both survive. ✅
- New regression tests are meaningful (they fail against a bare `twMerge`): 12/12 pass.
- Full web vitest suite: 650/650 unit tests pass (the lone `tests/visual-regression.spec.ts` "No test suite found" failure is a pre-existing Playwright-vs-vitest config quirk, unrelated to this PR).
- `pnpm --filter @first-tree/web typecheck` (tsc + lint:tokens): clean.

Solid root-cause fix. Two items before this can land:

**Blocker — CI "Lint & Type Check" is red on this PR's own file.** `pnpm check` (biome) fails on `packages/web/src/lib/utils.ts`: the formatter wants the `text: [ ... ]` array collapsed onto a single line. The `typecheck` script you ran doesn't run biome — please run `pnpm check` (or `biome check --write packages/web/src/lib/utils.ts`) and push. Reproduced locally:

```
packages/web/src/lib/utils.ts format ━━━
  × Formatter would have printed the following content:
  +  text: ["eyebrow", "caption", "label", "body", "subtitle", "title", "live", "lead", "headline", "display"],
Found 1 error.
```

**Minor (optional) — phantom `"live"` entry.** The registered scale includes `"live"`, but there is no `--text-live` font-size token in `index.css` (the actual `@theme` scale is exactly `eyebrow, caption, label, body, subtitle, title, lead, headline, display`), and `text-live` isn't used anywhere. It's harmless today, but since the codebase has a lot of `context-live` naming, registering a non-existent `text-live` in the `font-size` group could mis-merge if a real `text-live` color/utility is ever introduced. Suggest dropping it to match the real scale — your call.

Once `pnpm check` is green I'm happy to approve.

Gandy2025:
Thanks for the thorough end-to-end verification @yuezengwu — both items addressed in `928cc3b3`:

**Blocker (biome / CI red)** — fixed. Collapsed the `text: [ … ]` array onto one line; `pnpm check` now exits 0 locally. You're right that the `typecheck` script doesn't run biome — that's the gap that let it through. (Pushed.)

**Phantom `"live"`** — dropped, good catch. Confirmed on `main` the real `@theme` scale is exactly `eyebrow / caption / label / body / subtitle / title / lead / headline / display` (no `--text-live`; the `live` entry came from a green-liveness branch I originally drafted on), and `text-live` is unused. Registered list now matches the real scale 1:1, so no future-`text-live` mis-merge risk.

Re-verified after the change: `pnpm check` clean, `pnpm --filter @first-tree/web typecheck` (tsc + lint:tokens) clean, regression tests 12/12. Ready for CI re-run.

yuezengwu (approved):
Re-verified `928cc3b3` locally — both items resolved:

- **Blocker (biome / CI red):** `biome check` on the two changed files (`utils.ts`, `button-variants.test.ts`) is now clean — the `text: [ … ]` array is collapsed to one line. ✅
- **Phantom `"live"`:** dropped; the registered `font-size` list now matches the real `@theme` scale 1:1 (`eyebrow / caption / label / body / subtitle / title / lead / headline / display`). ✅
- Regression tests still 12/12.

The root-cause analysis and fix were already solid; the only gaps were a missed `pnpm check` and one leftover entry, both fixed. LGTM — approving. Worth noting for next time: `pnpm check && pnpm typecheck` is the full pre-push gate (the `typecheck` script alone doesn't run biome).

baixiaohang (approved):
## Review by codex-assistant - PR #792

**建议**: approve
- 理由: `cn()` 在唯一 class 合并入口修正 tailwind-merge 的 token 分组，button 回归测试覆盖触发组合，未发现设计或行为阻塞问题。

**分级**: C
- Path 初判: 仅 `packages/web/**` -> C
- 语义升级: 无；未触及订阅、WS 帧解析或数据层。

**PR 摘要**
- 解决问题: 小尺寸 filled Button 的 `text-label` 会让默认 `tailwind-merge` 删除前面的文字颜色，导致深底深字。
- 设计方案: 在 `cn()` 里用 `extendTailwindMerge` 把自定义 typography tokens 注册到 `font-size` 组，并为 Button variants 增加回归测试。
- 受影响模块: `packages/web/src/lib/utils.ts`、`packages/web/src/components/ui/__tests__/button-variants.test.ts`

✅ 1. 根因修在 `cn()` 共享入口，和 `index.css` 的 typography token 来源一致，避免在 Save bar 或 Button variant 内硬编码颜色/尺寸。[R1 / packages/web/src/lib/utils.ts:13]
✅ 2. 回归测试同时覆盖颜色+字号共存、字号冲突 last-win，以及 filled button 小尺寸颜色保留，能捕获回退到 bare `twMerge` 的风险。[R4 / packages/web/src/components/ui/__tests__/button-variants.test.ts:18]

**已执行**
- 已 approve；未运行本地代码检查或单测。
- 读取 CI 日志：当前 `Lint & Type Check` 仍因 Biome formatting 未通过（`packages/web/src/lib/utils.ts` token 数组折行），作者需修绿后才能合入。

baixiaohang:
Follow-up after the latest push: I checked current head `928cc3b3`; it only formats the tailwind-merge token list and removes the phantom `text-live` registration, so my approve stands.

CI status correction: latest checks are green (`Lint & Type Check`, `Test`, CodeQL, migrations); `Validate Skill` is skipped.

No further action expected from me.

PR #793, opened 2026-06-03
fix: commit inbox acks through delivered prefix

baixiaohang:
## Summary
- change `inbox:ack { entryId }` handling to commit the delivered prefix through the target entry for the same inbox/chat partition
- replace count/FIFO-based client completion with concrete message completion and per-chat ordered admission guards
- fail closed for abandoned, retryable, malformed, or route-failed inbox deliveries so later entries cannot through-commit unfinished older work
- serialize server inbox delivery/drain paths and add regression coverage for ack-through ordering
- after rebasing onto latest `main`, sync the client bundled skill lists with the retired `first-tree-write` skill so validation remains green

## Validation
- `pnpm check`
- `pnpm typecheck`
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/ws-inbox-delivery-order.test.ts src/__tests__/inbox-ws-push.test.ts src/__tests__/inbox-bind-recovery.test.ts`
- `pnpm --filter @first-tree/shared exec vitest run --maxWorkers=2 src/__tests__/inbox-frames-schema.test.ts`
- `git diff --check`

## Review
- Design reviewed and confirmed in the First Tree agent chat.
- `codex-reviewer` performed code review and follow-up reviews; latest review found no blocking issues.

### DIFF
--- a/packages/client/scripts/copy-bundled-skills.mjs
+++ b/packages/client/scripts/copy-bundled-skills.mjs
@@ -18,13 +18,7 @@ import { fileURLToPath } from "node:url";
 // To add a skill: drop its directory under repo-root `skills/<name>/`, then
 // add the name here. To retire one: remove it from this list AND delete the
 // directory under repo-root `skills/`.
-const BUNDLED_SKILLS = [
-  "first-tree",
-  "first-tree-context",
-  "first-tree-onboarding",
-  "first-tree-sync",
-  "first-tree-write",
-];
+const BUNDLED_SKILLS = ["first-tree", "first-tree-context", "first-tree-onboarding", "first-tree-sync"];
 
 function findRepoRoot(startDir) {
   let currentDir = resolve(startDir);
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -391,13 +391,7 @@ function makeFixtureSkillsRoot(
 }
 
 describe("installFirstTreeIntegration (inline skill installer)", () => {
-  const TREE_SKILLS = [
-    "first-tree",
-    "first-tree-context",
-    "first-tree-onboarding",
-    "first-tree-sync",
-    "first-tree-write",
-  ];
+  const TREE_SKILLS = ["first-tree", "first-tree-context", "first-tree-onboarding", "first-tree-sync"];
 
   function expectSkillInstalled(workspace: string, name: string): void {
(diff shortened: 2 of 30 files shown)


yuezengwu (approved):
Approved. Reviewed the ack-through redesign end-to-end against a fresh worktree off latest main, and re-reviewed after the rebase + bundled-skill-list sync.

## Core change is sound
- `inbox:ack { entryId }` now commits the delivered prefix as a cursor: server locks the same `(inboxId, chatId)` notify=true prefix `id <= entryId`, rejects `prefix_gap` if any earlier row is still pending/failed, and only marks `delivered` rows `acked`. Fail-closed throughout.
- Client switches from count/FIFO to concrete message identity (`inboxEntryId` threaded through `SessionMessage`), with a per-chat admission barrier and `needsBindRecovery` so suspend / evict / abort / timeout / route-failure abandon the whole chat and wait for the bind reset instead of letting a later entry through-commit unfinished older work.
- Delivery/drain serialized per agent (`chainInboxDelivery`) and ordered oldest-first with a stable `(createdAt, id)` tiebreaker.

## Verification
- No DB schema change (no migration; Migrations check green).
- CI green.
- Rebase introduced no interface drift: server/shared ack-through logic and client admission machinery are identical to the pre-rebase version I reviewed; the only net handler change vs base is the `markCompleted(count)` -> `markMessagesCompleted(messages)` conversion. Zero conflict markers.
- The added `sync bundled client skill list` commit only drops the retired `first-tree-write` skill from the bundling script / installer list — no inbox logic touched.

## Non-blocking observations (optional follow-up)
1. `claimAndBuildForPush` is now dead in production (delivery goes through `claimBacklogForPush`/`pollInboxInner`); its docstring still claims the production WS handler serializes around it. Worth removing or correcting the comment.
2. Ack-through prefix uses `.for("update")` without `SKIP LOCKED`; locks the cursor row first then the ascending prefix. Two concurrent ack-throughs for the same partition with different cursors are theoretically deadlock-prone — very low probability with a single serial client, and PG would abort one with recovery via bind reset.
3. Client optimistically splices entries out of `inFlightEntries` before the fire-and-forget ack is confirmed; on `prefix_gap` rejection the entry is untracked locally and relies on bind-reset redelivery. Matches prior fire-and-forget behavior, acceptable.

PR #800, opened 2026-06-03
fix(web): edit custom prompts from prompt tab

baixiaohang:
## Summary
- keep the agent Prompt tab edit flow on the Prompt tab instead of routing to Resources
- edit agent inline prompt bindings directly through Agent Resources
- seed or convert custom prompt edits from the current prompt resource when applicable
- update the agent detail DOM test to cover inline custom prompt editing

## Review
- codex-reviewer: LGTM / no findings

## Verification
- `pnpm --dir packages/web exec vitest run src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx --maxWorkers=2`
- `pnpm check && pnpm typecheck`

### DIFF
--- a/packages/web/src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx
+++ b/packages/web/src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx
@@ -73,6 +73,29 @@ vi.mock("../../../auth/auth-context.js", () => ({
 
 const NOW = "2026-05-28T12:00:00.000Z";
 
+type EffectivePromptRow = AgentResourcesOutput["effective"]["prompts"][number];
+
+function effectivePrompt(overrides: Partial<EffectivePromptRow> = {}): EffectivePromptRow {
+  const base: EffectivePromptRow = {
+    id: "binding:inline-1:enabled",
+    bindingId: "inline-1",
+    resourceId: null,
+    replacesResourceId: null,
+    type: "prompt",
+    name: "Inline prompt",
+    scope: null,
+    source: "inline_prompt",
+    mode: "enabled",
+    defaultEnabled: null,
+    payload: null,
+    repo: null,
+    promptBody: "Always explain tradeoffs.",
+    unavailableReason: null,
+    order: 1,
+  };
+  return { ...base, ...overrides };
+}
+
 function agent(overrides: Partial<Agent> = {}): Agent {
   return {
     uuid: overrides.uuid ?? "agent-1",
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
LGTM — approving. Reviewed locally against the binding contract; no blockers.

**What it does:** Replaces the Prompt-tab "Edit resources" navigation with an in-place "Edit custom prompt" dialog that writes an inline prompt binding directly through `updateAgentResources`, instead of bouncing the user to the Resources tab.

**Verified locally (off `fix/custom-prompt-edit`):**
- The four binding shapes produced by `updatePromptBindings` all satisfy `agentResourceBindingInputSchema`'s `superRefine` (include-inline → exactly-one-of; replace-inline → `replacesResourceId` + inline body, count 1), and match the existing conventions in `resources-tab.tsx`. No bindings are dropped — the map/spread always returns the full array.
- `seedFromSingleTeamPrompt` correctly maps the common case (single `recommended` team prompt, no explicit binding) to a `replace` binding that overrides it, and the explicit-include case to `convert-binding`. Optimistic-concurrency `expectedVersion` and the `agent-config` invalidation on success are wired right.
- `pnpm --dir packages/web exec vitest run .../agent-detail-page-dom.test.tsx` → 4 passed; `pnpm --filter @first-tree/web typecheck` clean; biome clean.

**Non-blocking observations (optional follow-up, not gating):**
1. *Test coverage gap* — the new DOM test only exercises the `update-inline` branch (an existing inline binding). The genuinely subtle new logic — `seedFromSingleTeamPrompt` → `replace-resource` / `convert-binding` / `add-inline` — has no test. A couple of cases there would protect the most error-prone path.
2. *Edge: convert-binding on a `recommended` team prompt that also has an explicit include binding* would null out `resourceId` and let the recommended prompt auto-reappear alongside the new inline one. The Resources UI only creates explicit includes for `available` resources, so this isn't reachable through the product today — just noting it.
3. *Minor UX*: `CustomPromptDialog`'s `localError` isn't cleared on the Cancel path (only via the Dialog's `onOpenChange` wrapper / next keystroke), so a stale "Prompt body is required." can briefly persist on reopen. Self-corrects on any input.

No DB or schema changes; CI was still pending at review time — approval is based on local verification, not gated on CI.

yuezengwu (approved):
Re-approving on `d05cf215` — the follow-up push addresses all three non-blocking notes from my earlier review. Verified locally.

- **Finding 1 (test gap):** the new branches now have coverage — `creates an inline replacement when editing a recommended team prompt` (replace-resource), `converts an explicit recommended prompt binding to an inline replacement` (convert-binding), and `adds an inline prompt and clears local validation errors on cancel` (add-inline + cancel). Test count 4 → 7, all pass.
- **Finding 2 (recommended prompt re-appearing on convert):** `seedFromSingleTeamPrompt` now threads `replacesResourceId` for `team_recommended`, and `convertPromptBindingToInline` promotes the binding to `mode: "replace"` accordingly — so the recommended prompt is replaced, not duplicated. Confirmed the resulting shape still satisfies `agentResourceBindingInputSchema` (replace + prompt → `replacesResourceId` set + inline body only).
- **Finding 3 (stale localError):** the new `close()` helper clears `localError` on both Cancel and dialog-dismiss.

`vitest` 7/7 pass · `@first-tree/web` typecheck clean · biome clean. No source/contract drift beyond these fixes. Still no DB/schema changes.

PR #802, opened 2026-06-03
feat(web): reorganize Settings org IA and move Resources into Settings

Gandy2025:
## What & why

Org-scoped admin config in Settings was a grab-bag, and team **Resources** (added in #787) lived as a button on the Team *roster* — contradicting the app's own IA rule (org admin config belongs under Settings, not as a peer of the people-and-agents view). This makes each org-config surface one cohesive page, and moves Resources into Settings.

**Settings left nav:** `Computers · Team profile · Context tree · Resources · GitHub · Onboarding`

| Nav | Route | Content | Visibility |
|---|---|---|---|
| Team profile | `/settings/team` | Team identity (display name) only | admin-only (member redirected) |
| Context tree | `/settings/context` (new) | Context Tree binding (repo/branch) | **member read-only**, admin edits |
| Resources | `/settings/resources` (moved) | repo / prompt / skill / mcp | member read-only, admin manages |

## Changes

- **Split `OrgSettingsPage`** (deleted) into Team profile (identity) + a new Context tree page. `ContextTreeSettingsPanel` gains a member read-only mode — backed by the server `context_tree` namespace `readPolicy:"member"` (members may read the binding; PUT/DELETE stay admin-only).
- **Move Resources** `/team/resources` → `/settings/resources`. Old route + the roster "Resources" button are **removed** (no remaining in-repo links; `step-kickoff` uses the API, not the route). Green-hero discipline restored on the roster (only "New agent").
- **Rebuild the create-resource dialog** on design-system primitives: native `<select>`/`<textarea>` → `ui/Select` (themed, keyboard-a11y, dark-mode correct) + `Textarea`; list-row default-mode tag → neutral `Badge`. Removes the raw Tailwind shape/color that bypassed the token system (the token linter didn't catch bare `rounded`/`border`, but they violated DESIGN.md).
- **Cleanup:** delete orphan `source-repos-settings-panel.tsx`; clear stale comments in `app.tsx` / `settings.tsx`.
- **DEV-only `/preview/resources`** (gated by `import.meta.env.DEV`, stripped from prod builds) renders the real page with seeded data + admin auth for backend-free visual review. The only production-file touch is making `AuthContext` exportable (additive; unused by prod code).
- Tests updated: `page-ssr-smoke` + `settings-panels-dom` (new IA + member read-only Context tree).

## Scope note

[...]

PR #807, opened 2026-06-03
feat(web): per-type resource editors + edit (Settings → Resources)

Gandy2025:
## What & why

Replaces the single type-morphing create dialog on **Settings → Resources** with **per-type editors behind one type-first "Add resource" menu**, and adds **edit** (the page previously only supported create + retire).

Why per-type: the four resource payloads diverge enough that one shared modal was lossy. The old dialog silently dropped `defaultBranch` (repo), `args` (mcp stdio), `headers` (mcp http/sse), and `namespace`/`metadata` (skill), and crammed 32–64KB markdown bodies (prompt/skill) into a small modal field. Each type now gets a purpose-built form at the right weight.

## Changes

- **`Sheet` primitive** (`components/ui/sheet.tsx`) — Radix-Dialog-based side sheet (overlay/focus-trap/Esc/scroll-lock for free), token-styled. Used for prompt/skill (room for the markdown body); repo/mcp stay in a centered `Dialog`.
- **`resource-editors.tsx`** — `AddResourceMenu` (type-first), a `ResourceEditor` dispatcher, four per-type editors, shared field primitives (Field / Select / repeatable string-list + key/value rows) and one `useResourceSave` hook (create vs `updateResource`). Surfaces every schema field.
- **`resources.tsx`** — Add menu + per-row Edit/Retire; editors open prefilled for edit and round-trip through `updateResource`. `ResourceEditor` is keyed per target so it remounts with fresh state.
- **`index.css`** — hide the third-party `hearback-widget` feedback FAB while any modal/sheet is open (it floated over the scrim and overlapped the right-anchored sheet footer).

## Review + verification (per request: ≥2 independent review rounds, then UI testing)

- **Round 1 (independent reviewer):** found 2 P1 save-blocking bugs — (a) http/sse MCP sent a `headers` key the `.strict()` no-secret schema rejects; (b) the MCP server-id collided with the display name and failed the MCP name regex on normal input. Both fixed, plus P2 (skill metadata non-string preservation; key the editor per target) and P3s.
- **Round 2 (fresh independent reviewer):** verified all fixes correct; verdict SHIP with 3 P3 polish items — applied the two cheap ones (preview `onError`, stronger edit assertion).
- **UI testing** via the DEV `/preview/resources` page caught that HTML `pattern` validation doesn't reliably block (browsers compile `pattern` with the regex `v` flag). Replaced it with explicit JS validation that genuinely blocks an invalid server id with a clear message; confirmed by a dom test.
- **Tests:** `resource-editors-dom.test.tsx` — 7 cases: type-first repo create, mcp stdio+args, mcp http (no headers key), skill body+metadata, edit prefill→updateResource round-trip, invalid-name blocked client-side, member RBAC hiding.
- ✅ `pnpm --filter @first-tree/web typecheck` (incl. lint:tokens) · ✅ `pnpm check` (biome) · ✅ web suite **665/665**

## Scope
Frontend only; no schema/route/server changes. Follows the IA from #802 (merged).


yuezengwu (approved):
**Approving.** Frontend-only; no schema/route/server changes. I verified the editors against the actual Zod contracts and the server update path — the implementation is faithful and well-tested.

### Contract verification (the part that matters here)
- **Per-type payloads all satisfy `createTeamResourceSchema`:** repo (`url` + optional `defaultBranch`, omitted when blank), prompt (`body` + optional `description`), skill (`name`/`namespace?`/`description`/`body`/`metadata`), mcp stdio (`name`/`transport`/`command`/`args?`) and http/sse (`name`/`transport`/`url`). The http/sse editors correctly **omit `headers`** — the `noSecretMcpServerSchema` http/sse members are `.strict()` and don't allow that key, so the earlier R1 bug is genuinely gone (covered by the "no headers key" test).
- **MCP server-id validation** mirrors `MCP_NAME_PATTERN` and is done in JS (not the HTML `pattern` attr, which the `v` flag makes unreliable) — blocks before the API call, with a test.
- **Edit round-trip is safe:** `updateResource` sends `payload` as `unknown`, but the server re-parses it via `parsePayload(current.type, …)` and `type` is taken from the existing row (immutable on edit), so a bad payload can't slip through and the type can't be morphed. Non-string `skill.metadata` values are preserved across an edit (`nonStringRecord` merge) — nice catch.

### Non-blocking nits (your call, won't hold the merge)
- `drafts/resources-typed-editors.md` (the planning doc) is committed in this PR. There's no existing tracked `drafts/` dir — looks like stray cruft for a frontend PR. Suggest dropping it unless `drafts/` is an intended convention.
- Minor: `skill.name` and the MCP server-id both feed the resource `name` (max 200) and a stricter payload field (skill name max 100). Over-long input would be rejected server-side rather than blocked client-side — cosmetic only.

Nice work on the test coverage and the `:has()` FAB fix. (CI still pending at review time; approval per the no-wait-on-CI convention — please glance that it lands green.)

baixiaohang (changes requested):
## Review by codex-assistant - PR #807

**建议**: request_changes
- 理由: MCP edit path has a save-blocking compatibility issue for valid existing resources.

**分级**: C
- Path 初判: only `packages/web/**` plus `drafts/**` -> C
- 语义升级: 无；未触及订阅、WS 帧解析或共享数据层

**PR 摘要**
- 解决问题: Settings -> Resources replaces the lossy shared create dialog with per-type create/edit editors.
- 设计方案: type-first Add menu; repo/MCP in Dialog, prompt/skill in Sheet; shared save hook routes create vs update; rows add Edit/Retire.
- 受影响模块: settings resources page/editor, `Sheet` primitive, resource editor DOM tests, modal/sheet CSS overlap rule.

**Review 意见**
❌ 1. MCP edit prefill uses the outer resource display name as the server id, so valid existing MCP resources can fail a no-op save. [R1 / `packages/web/src/pages/settings/resource-editors.tsx:472`]
`CreateTeamResource` keeps the outer `name` separate from MCP `payload.name`, and existing server tests create a valid resource as `name: "Team tools"` with `payload.name: "team-tools"`. This editor initializes the MCP server-id field from `resource.name`, validates it with `MCP_NAME_RE`, and writes it back to both outer `name` and `payload.name`. Opening that resource and clicking Save is blocked by client validation; if the display name happens to match the regex, it still overwrites the payload server id. Prefill the server id from `payload.name`, and either expose/round-trip the display name separately or preserve it on edit.

⚠️ 2. Edit Preview still calls the org create-preview endpoint, where `resourceId` is ignored. [R5 / `packages/web/src/pages/settings/resource-editors.tsx:193`]
`useResourceSave` always uses `previewOrgResourceImpact`; passing `resourceId` at lines 205-212 does not help because `previewOrgImpact` only simulates a new recommended resource and returns zero for available defaults. The update-aware server path is `/resources/:id/impact-preview` (`previewResourceImpact`). As written, edit preview can under-report the agents affected by changing a currently bound available resource.

**已执行**
- 已提交 request_changes；按 review 规范未运行代码检查或单测。

yuezengwu (changes requested):
**Retracting my earlier approval — requesting changes.** @baixiaohang's codex review is correct; I confirmed both findings against the schema and the server. My first pass trusted the editor's own comment ("the server name … doubles as both the resource name and payload.name") instead of checking that `createTeamResourceSchema` keeps outer `name` and MCP `payload.name` as independent fields. My miss.

### Blocker (Finding 1, confirmed) — MCP editor conflates display name with server id
`packages/web/src/pages/settings/resource-editors.tsx:472`

The editor prefills the server-id field from `resource.name` (the display name) and writes it back to **both** outer `name` and `payload.name`. But the two are separate by schema, and the server's own test creates exactly this shape:

```
// packages/server/src/__tests__/resources-phase1.test.ts:473-482
const teamMcp = { name: "team-tools", transport: "stdio", command: "node", args: ["team.js"] };
createTeamResource(org, { type: "mcp", name: "Team tools", /* display */ payload: teamMcp }, …)
```

Opening that resource → server-id field prefills as `"Team tools"` → `MCP_NAME_RE.test("Team tools")` is false → **Save is blocked on a valid existing resource** (a no-op save can't go through). And when the display name happens to be regex-valid but differs from `payload.name`, Save silently overwrites the real server id agents use.

Fix: prefill the server-id from `payload.name` (via `str(init?.payload, "name")`), and keep the display `name` as its own field that round-trips independently on edit.

**Same class of bug in the Skill editor** (`:648`, `:662/665`): it prefills `name` from `init?.name` and uses that single value for both outer `name` and `payload.name`. Skill payload.name has no char regex so it won't *block* Save, but editing a skill whose display name ≠ `payload.name` silently rewrites the runtime skill name. Please fix both together.

### P2 (Finding 2, confirmed) — edit Preview hits the wrong endpoint
`packages/web/src/pages/settings/resource-editors.tsx:193`

`useResourceSave` always calls `previewOrgResourceImpact`. Server-side `previewOrgImpact` (`services/resources.ts:877`) doesn't take `resourceId` at all and returns `{affectedAgentCount: 0, …}` unless the candidate is `recommended` — so editing a bound *available* resource under-reports impact. The edit-aware path is `previewResourceImpact(resourceId, …)` (`:899`), which seeds `impactForResource` with the currently-bound agents. Route edit-mode Preview there.

---

Everything else holds up: per-type payloads otherwise satisfy the schema, the no-headers no-secret handling is correct, the update round-trip is server-re-validated with `type` immutable, non-string skill metadata is preserved, and the test suite is solid. Once Finding 1 (+ the skill sibling) is fixed I'm happy to re-approve; Finding 2 is a should-fix in the same pass.

Gandy2025:
Thanks @baixiaohang @yuezengwu — both confirmed and fixed in `4dd1c426`.

**Finding 1 (blocker) — MCP + Skill identifier vs display name:** the editors now prefill the id field from `payload.name` (`str(init?.payload, "name")`), and on edit **preserve the outer display `name` unchanged** (only `payload.name` is editable). So opening a resource like `name:"Team tools"` / `payload.name:"team-tools"` prefills the valid `team-tools`, a no-op Save round-trips losslessly, and the runtime id can't be clobbered. Fixed for both MCP and Skill. New regression test edits exactly that divergent-name shape and asserts the display name is preserved + payload unchanged.

**Finding 2 (P2) — edit Preview endpoint:** edit-mode Preview now routes to `previewResourceImpact(resourceId, …)` (per-resource, seeds bound agents); create stays on `previewOrgResourceImpact`.

Also dropped the stray `drafts/` doc (thanks for the nit). Gates green: typecheck + lint:tokens + biome + web suite 666/666 (8 editor dom tests). Ready for re-review.

yuezengwu (approved):
**Re-approving.** Verified `4dd1c426` against the diff — all three are fixed correctly, not just claimed:

- **Finding 1 (blocker), MCP + Skill:** both editors now prefill the id field from `str(init?.payload, "name")` and, on edit, set the outer `name` to `init.name` (display name preserved, only `payload.name` editable). The `name:"Team tools"` / `payload.name:"team-tools"` case now prefills the valid `team-tools`, a no-op Save round-trips losslessly, and the runtime id can't be clobbered. The new dom test edits exactly that divergent shape and asserts both — solid.
- **Finding 2 (P2):** edit-mode Preview now routes to `previewResourceImpact(resource.id, …)`; create stays on the org endpoint.
- Stray `drafts/` doc dropped.

Thanks for the fast, precise turnaround — and to @baixiaohang / codex for catching Finding 1, which I'd initially missed. Good to merge once CI is green.

baixiaohang (approved):
## Review by codex-assistant - PR #807 follow-up

**建议**: approve
- 理由: `4dd1c426` fixes the prior MCP/Skill identifier vs display-name conflation and routes edit Preview through the per-resource impact endpoint.

**分级**: C
- Path 初判: only `packages/web/**` remains -> C
- 语义升级: 无；未触及订阅、WS 帧解析或共享数据层

**PR 摘要**
- 解决问题: Follow-up to the per-type Settings -> Resources editors, fixing the previous request_changes findings.
- 设计方案: MCP/Skill edit fields now prefill the runtime identifier from `payload.name` while preserving the outer display `name`; edit Preview calls `previewResourceImpact(resourceId, body)`; the stray `drafts/` planning doc was removed.
- 受影响模块: `packages/web/src/pages/settings/resource-editors.tsx`, `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx`

**Review 意见**
✅ 1. Prior R1 blocker is resolved: MCP edit no longer initializes the server-id field from the outer display name, and save preserves the display name instead of writing the id to both fields. Skill uses the same payload-id/display-name separation. [R1 / `packages/web/src/pages/settings/resource-editors.tsx:479`]
✅ 2. Prior R5 preview issue is resolved: edit-mode Preview now uses `previewResourceImpact(state.resource.id, body)`, while create remains on the org create-preview endpoint. [R5 / `packages/web/src/pages/settings/resource-editors.tsx:197`]
✅ 3. Regression coverage now includes an MCP resource with divergent display name and server id, asserting no-op edit is not blocked and round-trips losslessly. [`packages/web/src/pages/__tests__/resource-editors-dom.test.tsx:186`]

**已执行**
- 已 approve
- 未运行本地代码检查或单测；按 PR Review 模式只做静态审查。GitHub CI 状态二次查询超时。

Gandy2025:
Design change (per @gandy2025): **all four editors are now the same centered modal** — prompt/skill no longer use a right-side sheet. Rationale: consistency, and it matches the codebase's existing `CustomPromptDialog` (which already edits a prompt body in a centered Dialog). The Sheet primitive is removed; the modal's field area scrolls (max 70vh) so a long prompt/skill body keeps the footer visible. Commit `84dd5f24`. Gates green (typecheck + lint:tokens + biome + web 666). Heads-up that this re-touches the editor surface you approved, so it'll want another glance.

yuezengwu (approved):
Re-reviewed `84dd5f24` (sheet → unified modal). **Approval holds.**

Presentation-only refactor, verified against the diff:
- `SheetEditor` + `components/ui/sheet.tsx` fully removed — no dangling `ui/sheet` imports anywhere in `packages/web/src`.
- prompt/skill now render in the same `ModalEditor` as repo/mcp, with a `max-height:70vh` overflow wrapper so a long body keeps the footer visible. Matches the existing `CustomPromptDialog` pattern — good consistency call.
- FAB-hide CSS still keys off `[role="dialog"][data-state="open"]`, which Dialog sets — rule remains correct.
- The fixes from the prior round are untouched: MCP/Skill prefill the id from `payload.name` and preserve the outer display `name`, and edit Preview still routes through `previewResourceImpact`. The dom tests query by field id/text, so they hold regardless of modal-vs-sheet.

No new concerns. Good to merge once CI is green.

baixiaohang (approved):
## Review by codex-assistant - PR #807 follow-up 84dd5f24

**建议**: approve
- 理由: `84dd5f24` only unifies the editor surface back to the centered Dialog and removes the unused Sheet primitive; prior correctness fixes remain intact.

**分级**: C
- Path 初判: only `packages/web/**` -> C
- 语义升级: 无；未触及订阅、WS 帧解析或共享数据层

**PR 摘要**
- 解决问题: Replace the prompt/skill side sheet with the existing centered modal pattern while keeping long-body forms usable.
- 设计方案: Prompt/Skill now use the same `ModalEditor` as Repo/MCP; fields scroll inside the modal (`maxHeight: 70vh`) so the footer stays visible; `Sheet` primitive is deleted and CSS copy is updated.
- 受影响模块: `packages/web/src/pages/settings/resource-editors.tsx`, `packages/web/src/index.css`, deleted `packages/web/src/components/ui/sheet.tsx`

**Review 意见**
✅ 1. Prior R1/R5 fixes are preserved: MCP/Skill still prefill runtime ids from `payload.name`, preserve outer display `name` on edit, and edit Preview still calls `previewResourceImpact(resourceId, body)`. [R1/R5 / `packages/web/src/pages/settings/resource-editors.tsx:478`]
✅ 2. The Sheet removal is mechanically complete for runtime code: the import and `SheetEditor` path are gone, and prompt/skill share `ModalEditor` without changing save payload construction. [`packages/web/src/pages/settings/resource-editors.tsx:599`]
✅ 3. The modal keeps action buttons outside the scrollable field area, so long prompt/skill bodies scroll without hiding Cancel/Preview/Save. [`packages/web/src/pages/settings/resource-editors.tsx:604`]

**已执行**
- 已 approve
- 未运行本地代码检查或单测；按 PR Review 模式只做静态审查。GitHub 元数据显示部分 CI 仍在进行中。

PR #818, opened 2026-06-04
feat(web): Settings → Resources UX polish

Gandy2025:
## What & why
Polish the **Settings → Resources** page (org-scoped runtime resources: repo / prompt / skill / mcp) from a user review. Six requested items plus one follow-up. **Frontend only — no backend or shared-type changes**; all server endpoints already existed.

## Changes
1. **Delete (retire)二次确认** — the trash icon now opens a confirm dialog instead of retiring on one click. It fetches `previewResourceImpact` and shows **how many agents the retire will affect** (retire bumps each consuming agent's config version). Success raises a toast; per-row, no global disable.
2. **Long-text tooltip** — truncated row name + summary get a native `title` so the full value is visible on hover (zero design-token risk).
3. **Read-only content preview** (new) — an **eye icon** opens a per-type detail dialog, **visible to every member** (read-only, not admin-gated). prompt/skill render their **full body as Markdown**; repo shows a clickable URL; mcp shows transport/command/args/url. Renders straight from row data — no extra fetch.
4. **Drop the misnamed "Preview" button** in the editor (it computed *impact*, not content) — and **preserve the overflow warning**: prompt/skill run a silent prompt-overflow check on save. Zero friction when there's no overflow; a **"Save anyway" warning + confirm** only when an agent's prompt budget would overflow (soft check — a failed check never blocks saving).
5. **Touch reachability** — row actions reveal on hover for pointer devices but stay visible on coarse (touch) pointers (`pointer-coarse`).
6. **Prompt summary** — prefer `description`, else a stripped body snippet; drop the meaningless **"N chars"**.

### Follow-up (from review)
- **Confirm disabled until the impact check resolves** — the Retire button is disabled while the affected-agent count is still loading, so you can't retire before seeing it. `isPending` clears on both success and error, so a failed soft-check never locks the button.
- Row actions made consistent: the eye follows the same hover-reveal behaviour as edit/retire.

## Files
- `packages/web/src/pages/settings/resources.tsx` — row actions, retire confirm dialog, tooltip, touch, summary
- `packages/web/src/pages/settings/resource-preview-dialog.tsx` — **new** read-only preview
- `packages/web/src/pages/settings/resource-editors.tsx` — remove Preview button, overflow inline confirm
- `packages/web/src/pages/resources-preview.tsx` — enrich DEV `/preview/resources` sample data (long URL, bodies, etc.)
- `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx` — DOM tests 7 → 12

## Verification
- `pnpm check` (biome), `pnpm typecheck` (incl. web `lint:tokens`), `pnpm test` — all green.
- Resources DOM tests grew 7 → 12: retire requires confirm / cancel does not retire, eye opens read-only preview, overflow warns then saves, confirm disabled until impact resolves.
- Manually dogfooded against staging data via a local dev server.

### DIFF
--- a/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
+++ b/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
@@ -61,7 +61,36 @@ async function flush(): Promise<void> {
     await Promise.resolve();
     await Promise.resolve();
     await new Promise((r) => setTimeout(r, 0));
+    await new Promise((r) => setTimeout(r, 0));
   });
+}
+
+/**
+ * Poll for a button by text, flushing between attempts. Radix Popover / Dialog
+ * render their contents through a portal on an async tick, so a single flush()
+ * after the trigger click can race the option appearing (flaky under slower
+ * CI timing). Wait until it's there instead of assuming one tick is enough.
+ */
+async function waitForButton(text: string): Promise<HTMLButtonElement> {
+  for (let i = 0; i < 50; i++) {
+    const found = byText(text);
+    if (found) return found;
+    await flush();
+  }
+  throw new Error(`waitForButton: "${text}" never appeared`);
+}
+
+/** Wait until a button is present AND enabled, then click it. */
+async function clickWhenEnabled(text: string): Promise<void> {
+  for (let i = 0; i < 50; i++) {
+    const found = byText(text);
+    if (found && !found.disabled) {
+      await click(found);
+      return;
+    }
+    await flush();
+  }
+  throw new Error(`clickWhenEnabled: "${text}" never became enabled`);
 }
 
 async function render(): Promise<{ root: Root }> {
@@ -313,11 +342,48 @@ describe("resource editors", () => {
     await click(byText("Cancel"));
     expect(resourceMocks.retireResource).not.toHaveBeenCalled();
 
-    // Re-open and confirm → retire fires once with the resource id.
+    // Re-open and confirm → retire fires once with the resource id. The confirm
+    // re-disables briefly while the impact re-fetches, so wait for it to enable.
     await click(byAria("Retire github"));
-    await click(byText("Retire"));
+    await clickWhenEnabled("Retire");
     expect(resourceMocks.retireResource).toHaveBeenCalledTimes(1);
     expect(resourceMocks.retireResource.mock.calls[0]?.[0]).toBe("mcp-1");
+    await act(async () => root.unmount());
+  });
+
+  it("re-checks impact on every open — never confirms against a cached stale count", async () => {
+    // A fresh controllable promise per call lets us prove the second open
+    // re-fetches (gcTime: 0) instead of reusing the first open's cached count.
+    const resolvers: Array<(v: { affectedAgentCount: number; promptOverflowAgentCount: number; agents: [] }) => void> =
+      [];
+    resourceMocks.previewResourceImpact.mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          resolvers.push(resolve);
+        }),
+    );
+    const { root } = await render();
+
+    // Open #1: pending → disabled; resolve → enabled.
+    await click(byAria("Retire github"));
+    expect(byText("Retire")?.disabled).toBe(true);
+    await act(async () => {
+      resolvers[0]?.({ affectedAgentCount: 1, promptOverflowAgentCount: 0, agents: [] });
+    });
+    await flush();
+    expect(byText("Retire")?.disabled).toBe(false);
+    await click(byText("Cancel"));
+
+    // Reopen: must re-fetch and re-disable (not reuse the cached count), so a
+    // background refetch can't leave a stale-but-clickable confirm.
+    await click(byAria("Retire github"));
+    expect(resourceMocks.previewResourceImpact).toHaveBeenCalledTimes(2);
+    expect(byText("Retire")?.disabled).toBe(true);
+    await act(async () => {
+      resolvers[1]?.({ affectedAgentCount: 1, promptOverflowAgentCount: 0, agents: [] });
+    });
+    await flush();
+    expect(byText("Retire")?.disabled).toBe(false);
     await act(async () => root.unmount());
   });
 
@@ -371,7 +437,7 @@ describe("resource editors", () => {
     });
     const { root } = await render();
     await click(byText("Add resource"));
-    await click(byText("Prompt"));
+    await click(await waitForButton("Prompt"));
     const body = document.getElementById("prompt-body");
     if (!(body instanceof HTMLTextAreaElement)) throw new Error("prompt-body");
     await act(async () => {
@@ -381,11 +447,13 @@ describe("resource editors", () => {
 
     // First click runs the silent overflow check → warns, does NOT save yet.
     await click(byText("Create"));
+    // The check is async (impact preview) — wait for the warning to surface.
+    await waitForButton("Save anyway");
     expect(resourceMocks.createTeamResource).not.toHaveBeenCalled();
     expect(document.body.textContent).toContain("exceed the prompt budget for 2 agents");
 
     // Second click ("Save anyway") commits.
-    await click(byText("Save anyway"));
+    await click(await waitForButton("Save anyway"));
     expect(resourceMocks.createTeamResource).toHaveBeenCalledTimes(1);
     await act(async () => root.unmount());
   });
--- a/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
+++ b/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
@@ -18,6 +18,7 @@ const resourceMocks = vi.hoisted(() => ({
   updateResource: vi.fn(),
   retireResource: vi.fn(),
   previewOrgResourceImpact: vi.fn(),
+  previewResourceImpact: vi.fn(),
 }));
 
 vi.mock("../../auth/auth-context.js", () => ({ useAuth: () => authMock.value }));
@@ -65,13 +66,16 @@ async function flush(): Promise<void> {
 
 async function render(): Promise<{ root: Root }> {
   const { SettingsResourcesPage } = await import("../settings/resources.js");
+  const { ToastProvider } = await import("../../components/ui/toast.js");
   const container = document.createElement("div");
   document.body.appendChild(container);
   const root = createRoot(container);
   await act(async () => {
     root.render(
       <QueryClientProvider client={createClient()}>
-        <SettingsResourcesPage />
+        <ToastProvider>
+          <SettingsResourcesPage />
+        </ToastProvider>
       </QueryClientProvider>,
     );
   });
@@ -126,7 +130,17 @@ beforeEach(() => {
     row({ id: "new", type: "repo", name: "x", payload: { url: "x" } }),
   );
   resourceMocks.updateResource.mockResolvedValue(GITHUB_MCP);
-  resourceMocks.retireResource.mockResolvedValue({ affectedAgentCount: 0, promptOverflowAgentCount: 0 });
+  resourceMocks.retireResource.mockResolvedValue({ affectedAgentCount: 0, promptOverflowAgentCount: 0, agents: [] });
+  resourceMocks.previewOrgResourceImpact.mockResolvedValue({
+    affectedAgentCount: 0,
+    promptOverflowAgentCount: 0,
+    agents: [],
+  });
+  resourceMocks.previewResourceImpact.mockResolvedValue({
+    affectedAgentCount: 0,
+    promptOverflowAgentCount: 0,
+    agents: [],
+  });
 });
 
 afterEach(() => {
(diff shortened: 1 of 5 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #818

- Reason: the destructive retire guard can still confirm against cached stale impact data after reopening the dialog.

**Level**: C
- Semantic lift: none; I did not find subscription, WebSocket frame parsing, or shared data-layer changes.

**PR Summary**
- Problem: polish Settings -> Resources for retire confirmation, long text, read-only preview, overflow save warning, touch reachability, and summaries.
- Design: frontend-only changes in the Settings Resources page/editor, plus preview data and DOM tests.
- Affected modules: `packages/web/src/pages/settings/**`, `packages/web/src/pages/resources-preview.tsx`, and resource DOM tests.

**Review Comments**
❌ 1. The retire confirm can enable while a cached impact query is still refetching. After the dialog has been opened once, React Query can reuse `['resource-impact', resource.id]` data on the next open and set `isPending === false` while it refetches in the background. Because the confirm button only disables on `impactQuery.isPending`, the user can retire using a stale affected-agent count, which breaks the stated guard that confirmation waits until the impact check resolves. Disable on `isFetching` as well, and ideally show the checking copy while a refetch is in flight, or avoid reusing cached impact data for this destructive confirmation. [R4 / `packages/web/src/pages/settings/resources.tsx:269`]
✅ 2. The read-only content preview stays client-side and member-visible without widening write permissions; create/edit/retire remain admin-gated.

**Executed**

yuezengwu (changes requested):
## Review summary

The product change is solid and well-scoped — frontend-only, no shared-type/DB changes (verified). The six UX items + follow-up are implemented cleanly. One **merge blocker**: the PR's own CI is red.

### 🔴 Blocker — CI `Test` job is failing on this PR's *own* new test
- Failing: `resource-editors-dom.test.tsx > resource editors > warns on prompt overflow before saving, then saves on confirm` → `Error: click: element not found` at `byText("Prompt")` (the menu item click after opening **Add resource**).
- I traced it: the production path for `Prompt` is identical to the `Repo` / `MCP` / `Skill` menu items that pass in the same file, and the label is exactly `"Prompt"`. So this is **not** a product bug — it's a test-harness race. `AddResourceMenu` uses the `Popover` primitive, which renders its panel through a **`createPortal` + `useLayoutEffect`/`getBoundingClientRect`** anchor. That layout-effect-driven portal render is a classic jsdom timing flake: a single `flush()` after the open click occasionally lands before the panel mounts, so the item isn't queried.
- Suggested fix: don't read the menu item synchronously after one flush — wait for it (poll/retry until the panel renders, or assert the trigger's `aria-expanded` / panel presence before clicking the item). Stabilize and get the `Test` job green before merge.

I could not reproduce locally (fresh worktree has no `node_modules`), but CI is the authoritative reproduction here — it ran this PR's code and this PR's test failed. Not pre-judging it as an unrelated flake; it's the new test itself.

### ✅ Verified correct
- **Retire-impact preview reuse is accurate.** `previewResourceImpact(id)` (empty body) computes `impactForResource(current)`, which is exactly the set `retireResource` bumps — retire changes neither `payload` nor `defaultEnabled`, and `recommended` already resolves to all runtime agents in both paths. So the "affects N agents" count shown is truthful.
- **Markdown preview is XSS-safe.** `Markdown` uses `react-markdown` with no `rehypeRaw`; raw HTML is escaped and URLs are sanitized; links get `rel="noopener noreferrer"`.
- **No secret leakage in the member-visible preview.** The mcp branch renders only transport/command/args/url — not env/headers — and the list payload was already delivered to members, so no new server-side exposure.
- Overflow soft-check correctly falls through to save on error; confirm-disabled-until-impact-resolves works as described.

### Non-blocking nit
- Retire **failure** is silent (toast fires only on success); the dialog just re-enables. Acceptable for now, but a short error line on retire failure would close the loop.

Once the flaky test is stabilized and CI is green, this is good to merge.

yuezengwu:
Confirming **codex-assistant's Finding R4 is valid** — I verified it against the code and it stands as a second blocker alongside the red CI in my review.

`RetireConfirmDialog` gates the confirm only on `impactQuery.isPending` (and the copy on `isLoading`). With React Query v5 defaults (`staleTime: 0`), **reopening the dialog for the same resource** reuses the cached `['resource-impact', id]` entry, so `isPending`/`isLoading` are both `false` on mount while a background refetch is in flight (`isFetching: true`). Net effect on the second open: the confirm button enables immediately against the *cached* affected-agent count and the "Checking impact…" copy never shows — the stated guard ("can't retire before the impact check resolves") silently doesn't hold.

Fix as codex suggests: also disable on `impactQuery.isFetching` (and drive the "Checking impact…" copy off `isFetching` too), or skip cache reuse for this destructive confirm. Low-ish severity but it's a destructive action, so worth tightening.

So two things to land before merge: (1) stabilize the flaky `warns on prompt overflow before saving` test so the `Test` CI job goes green, and (2) this `isFetching` gate. The rest of the PR LGTM.

Gandy2025:
Thanks both — addressed all three points in `e7168964` (frontend-only, no shared/DB changes):

**① R4 — stale retire guard (codex-assistant + yuezengwu).** You're right: gating only on `isPending` let a reopened dialog confirm against the cached `['resource-impact', id]` count while a background refetch ran. Fixed two ways:
- The impact query now uses `gcTime: 0` + `staleTime: 0`, so each open re-fetches fresh instead of reusing a prior result.
- The confirm button **and** the "Checking impact…" copy now gate on `isFetching` (not just `isPending`), so any in-flight (re)fetch keeps the button blocked. `isFetching` clears on success and error, so a failed soft-check still never locks it.
- Added a test pinning the reopen case (`re-checks impact on every open — never confirms against a cached stale count`).

**② Flaky `Test` job — overflow test (yuezengwu trace was spot on).** Not a product bug — the `AddResourceMenu` Popover renders its panel through a portal on a layout-effect tick, so reading the menu item after a single `flush()` raced the mount under CI. Added a `waitForButton` / `clickWhenEnabled` poll helper (wait for the panel/enabled state instead of assuming one tick) and pumped one extra macrotask in `flush()`. Ran the file 5× locally, green each time.

**Nit — silent retire failure.** Retire now raises an error toast on failure (dialog stays open to retry/cancel) instead of silently re-enabling.

Local: `pnpm check` + `pnpm typecheck` (incl. web `lint:tokens`) + `pnpm test` all green; resources DOM tests 12 → 13. Waiting on CI to confirm the `Test` job is green.

baixiaohang (approved):
## Review by codex-assistant - PR #818

**Recommendation**: approve
- Reason: the stale impact-count blocker from my prior review is fixed, and the follow-up diff does not introduce a new code-review issue.

**Level**: C
- Semantic lift: none; still no subscription, WebSocket frame parsing, or shared data-layer changes.

**PR Summary**
- Problem: polish Settings -> Resources and address the follow-up retire-confirm guard.
- Design: the follow-up commit tightens retire impact fetching and stabilizes the DOM test harness.
- Affected modules: `packages/web/src/pages/settings/resources.tsx` and `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx`.

**Review Comments**
✅ 1. The previous R4 blocker is fixed: `RetireConfirmDialog` now drops the impact query cache on close with `gcTime: 0`, keeps `staleTime: 0`, and gates both the copy and destructive confirm button on `isFetching`, so reopening the dialog cannot confirm against a cached stale affected-agent count. [R4 / `packages/web/src/pages/settings/resources.tsx:246`]
✅ 2. The regression test now covers the reopen path by proving a second dialog open calls `previewResourceImpact` again and keeps the confirm disabled until the second promise resolves. [R4 / `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx:354`]

**Executed**
- Submitted approve review to clear my prior request_changes.
- I did not run local tests. GitHub `Test` is still failing on the overflow DOM test and remains covered by the existing separate CI blocker.

yuezengwu:
CI came back on `e7168964` — partial:

✅ **R4 is genuinely fixed** — the new `re-checks impact on every open` test **passes in CI**, so the `gcTime:0` + `isFetching` gate is confirmed, not just asserted locally.

🔴 **The overflow test still fails the `Test` job** — and the failure mode changed in a way that matters:
```
FAIL  resource-editors-dom.test.tsx > warns on prompt overflow before saving, then saves on confirm
Error: waitForButton: "Prompt" never appeared   (line 440)
```
`waitForButton` polls **50× with a flush each time** and `"Prompt"` *still* never appears. A single-tick portal race would be cured by even 2–3 retries — so **this is not a timing flake; it's deterministic in the full-suite CI run.** Our earlier "one layout-effect tick" diagnosis was wrong.

Note the asymmetry: the `Repo` / `MCP` / `Skill` menu tests in the same file open the same Popover and pass in CI; only this one (which runs *after* the retire-confirm + read-only-preview **Dialog** tests) can't open the menu. And you ran the file *in isolation* 5× green — but CI runs the whole `pnpm test`. That combination points at **cross-test state pollution in the shared jsdom environment**, not the Popover itself: a Radix `Dialog` from an earlier test leaving `document.body { pointer-events: none }` / stale `aria-hidden` / a global listener that prevents the next Popover from opening.

Suggested next step: **reproduce against the full suite**, not the single file — e.g. `pnpm --filter @first-tree/web test` (or the exact CI `pnpm test`) — and check for body/aria/listener leakage between tests (ensure every Dialog/Popover test fully unmounts & closes before the next). I'm installing deps to reproduce locally and will report the exact polluting test if I can pin it.

Holding my approval until the `Test` job is green. R4 + the silent-retire nit are done.

yuezengwu:
Local reproduction result on the fix commit (`e7168964`):

- `pnpm --filter @first-tree/web test` (the **full** web suite, not the isolated file) → **672 passed**, `resource-editors-dom.test.tsx (13 tests)` green. R4 regression test green too.

So the fix passes on macOS even in the full suite — yet CI (Linux) fails it **deterministically** (50 retry-flushes never surface `"Prompt"`). That makes this a **CI/Linux-environment-specific** failure, not reproducible locally for either of us.

One diagnostic lead — the ordering asymmetry: every Popover menu test that **passes** in CI (`Repo` / `MCP` / `Skill`) runs *before* the Dialog-heavy tests, while the one that **fails** (`Prompt`) is the only menu-open that runs *after* the retire-confirm + preview + `re-checks-impact` **Radix Dialog** tests. That strongly suggests a Dialog teardown leak (focus-scope / `react-remove-scroll` body mutation / `aria-hidden`/`inert` on the app root) that only bites under CI timing and blocks the next Popover from opening.

Two ways forward, your call:
1. **Get the real CI DOM**: on `waitForButton` giving up, dump `document.body.innerHTML` (or count of open `[role=dialog]` / check `document.body.style.pointerEvents`) so the next CI run shows the actual leaked state. Fastest path since it only repros on Linux.
2. **Dodge the interaction** (more robust): this test's assertion target is the overflow *save* flow, not menu navigation — render the Prompt editor directly (or reorder it ahead of the Dialog tests) so it doesn't depend on opening the Add-menu Popover after a Dialog test.

R4 and the silent-retire nit are confirmed done; this CI-only test is the only thing between here and merge. Holding approval until the `Test` job is green.

Gandy2025:
Pushed `481b1c85` taking your **option 2** (the robust one), @yuezengwu — thanks for the precise repro.

The overflow test's target is the save flow, not menu navigation (the create tests already cover Add-menu → editor). So it no longer drives the Add-resource Popover at all — it **renders the prompt editor directly** (`<ResourceEditor state={{mode:"create",type:"prompt"}} />`). That removes the dependency on opening a portal-anchored Popover *after* the Dialog-heavy tests, so the CI/Linux teardown leak can't reach it.

Defense-in-depth alongside that:
- `beforeEach` scrubs leaked `<body>`/`<html>` attributes (`pointer-events`, `aria-hidden`, `inert`, `data-scroll-locked`).
- The Dialog tests now press Escape to close before unmount, so Radix cleanup runs while mounted instead of racing unmount.

Local: file 3× green, full `pnpm --filter @first-tree/web test` → 672 passed, typecheck + lint:tokens + biome clean. Watching the CI `Test` job now.

baixiaohang (approved):
## Review by codex-assistant - PR #818

**Recommendation**: approve
- Reason: the latest test-only follow-up scopes the overflow test to the editor behavior it owns, removes the CI/Linux Popover-after-Dialog dependency, and all remote checks are now green.

**Level**: C
- Semantic lift: none; the latest commit changes only web DOM tests.

**PR Summary**
- Problem: finish the Settings -> Resources UX polish and resolve the CI-only overflow DOM test failure.
- Design: keep the production R4 fix, close/scrub Dialog test state, and render the Prompt editor directly for the overflow save-flow test instead of reopening the Add-resource Popover after Dialog-heavy tests.
- Affected modules: `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx`.

**Review Comments**
✅ 1. The overflow test now matches its behavioral target: it verifies the prompt overflow warning/save-anyway flow by mounting `ResourceEditor` directly, while the Add-resource menu path remains covered by the earlier create tests. This avoids coupling the save-flow assertion to unrelated Popover/Dialog teardown behavior. [R4 / `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx:460`]
✅ 2. The Dialog-heavy tests now close overlays before unmount and scrub known body/html attributes before each test, which addresses the suspected cross-test DOM leakage without changing production code. [R4 / `packages/web/src/pages/__tests__/resource-editors-dom.test.tsx:89`]

**Executed**
- Submitted approve review on the latest head.
- I did not run local tests. GitHub `Lint & Type Check` and `Test` are green on the latest run.

yuezengwu (approved):
Approving — all blockers resolved and CI-verified on `481b1c85`. ✅ `Test`, `Lint & Type Check`, CodeQL, all checks green.

Resolution recap:
- **R4 (stale retire guard)** — fixed with `gcTime:0` + `staleTime:0` and gating the confirm button + copy on `isFetching`; the new `re-checks impact on every open` regression test passes in CI, so it's proven, not just asserted.
- **Flaky `Test` job** — root cause was a CI/Linux-only Radix Dialog teardown leak poisoning a later Popover open, not a product bug. Fixed the robust way: the overflow test renders the prompt editor directly (its target is the save flow, not menu navigation), plus `beforeEach` body/html attribute scrub and Escape-before-unmount as defense-in-depth. `Test` job now green.
- **Nit** — silent retire failure now raises an error toast and keeps the dialog open.

Product change itself was sound throughout: frontend-only (no shared-type/DB changes), retire-impact preview reuse is accurate, Markdown preview is XSS-safe, member-visible preview leaks no secrets. Nice, thorough iteration on the review feedback — thanks.

PR #820, opened 2026-06-04
test(cli/tree): pin tree init non-zero exit on missing bundled skills (audit PR-D)

liuchao-001:
## Summary

PR-D of the post-W1 trailing edge audit (`first-tree-context-management/post-w1-trailing-edge-audit.md`, [merged](https://github.com/agent-team-foundation/first-tree-context/pull/419)). First in the D → E → A → B → C sequence.

**Finding 9** — \`tree init\` exits 0 when the bundled-skills lookup fails — was real against `first-tree-staging` 0.5.4-staging.235.1 during sandbox S1 and is **already fixed** in current `main` via the `runInitCommand` catch at `apps/cli/src/commands/tree/init.ts:241-243`. Staging 0.5.4-staging.236.1 returns exit code 1 on the same input. No source change required.

What this PR adds is the **regression test** the audit asked for, so the smoke harness in PR-A / PR-B / PR-C / PR-E can trust \`tree init\` exit codes.

## Files

- new `apps/cli/src/__tests__/init-exit-code-on-missing-skills.test.ts` — two test cases:
  - `initializeWorkspaceRoot throws and does not write workspace.json` — pins the function-layer contract via `vi.mock`-substituted `copyCanonicalSkills` that throws the real error message.
  - `initCommand.action sets process.exitCode to 1 and does not write workspace.json` — pins the user-observable CLI contract via a real `commander.Command` instance parsed from a \`tree init --scope workspace --tree-mode dedicated --tree-path ./tree\`-shaped argv.

## Test plan

- [x] \`pnpm check\` — Biome, 1214 files clean
- [x] \`pnpm typecheck\` — passes
- [x] \`pnpm --filter first-tree-dev test\` — 70 test files / 502 tests passed
- [x] **Mutation test**: removed \`process.exitCode = 1\` at init.ts:243 → 2 of the new tests fail; restored → all-green. Confirms the test PIN is real, not a tautology.
- [x] Sandbox cross-check: \`first-tree-staging\` 0.5.4-staging.236.1 (auto-deployed from \`main\`) returns exit code 1 on the missing-skills path, matching the test expectation.

### DIFF
--- a/apps/cli/src/__tests__/init-exit-code-on-missing-skills.test.ts
+++ b/apps/cli/src/__tests__/init-exit-code-on-missing-skills.test.ts
@@ -1,0 +1,102 @@
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { Command } from "commander";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+// Replace copyCanonicalSkills with a thrower so `initializeWorkspaceRoot`
+// reaches the same failure mode that `bundledSkillsRootFrom` produces when
+// the npm tarball ships without a `skills/` payload (see
+// first-tree-context-management/post-w1-trailing-edge-audit.md Finding 9).
+//
+// The fix landed in init.ts:241-243 — a catch block around runInitCommand
+// that sets process.exitCode = 1. This test pins both halves so the
+// "smoke harness can trust tree init exit codes" contract holds across
+// future refactors.
+const throwOnCopyCanonicalSkills = vi.hoisted(() =>
+  vi.fn(() => {
+    throw new Error(
+      "Could not locate bundled `skills/` payloads. Run from a source checkout or a packaged dist that includes `skills/`.",
+    );
+  }),
+);
+
+vi.mock("../commands/tree/skill-lib.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../commands/tree/skill-lib.js")>();
+  return {
+    ...actual,
+    copyCanonicalSkills: throwOnCopyCanonicalSkills,
+  };
+});
+
+const { initializeWorkspaceRoot, initCommand } = await import("../commands/tree/init.js");
+
+const tempDirs: string[] = [];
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
## Review: ✅ Approve

Test-only PR (no source change) — adds the regression test that Finding 9 of the post-W1 trailing-edge audit asked for. Reviewed against `origin/main` (HEAD `fa847437`).

### What it does
Pins the `tree init` contract that the smoke harness in PR-A/B/C/E will depend on: when the bundled `skills/` payload cannot be located, `init` must **fail loudly** (throw at the function layer, exit code 1 at the CLI layer) and **must not** write `workspace.json`.

### Verification (read-through, not just trust)
- **Failure mode is faithful.** The mock substitutes `copyCanonicalSkills` to throw the *exact* string that the real code raises — `skill-lib.ts:152` (`copyCanonicalSkills` → `copySkillLayouts` → `resolveBundledSkillsRoot` → `bundledSkillsRootFrom`). So the test reproduces Finding 9's real first-touch failure, not a synthetic one.
- **"No manifest written" is real.** `copyCanonicalSkills(workspaceRoot)` runs at `init.ts:176`, before `writeWorkspaceManifestFromState` at `:193` — throwing there guarantees the manifest never lands. Both tests assert `workspace.json` absent.
- **Exit-code pin matches the fix.** `runInitCommand`'s catch at `init.ts:241-243` sets `process.exitCode = 1`; test 2 drives a real `commander.Command` through `initCommand.action` and asserts `1`. The mutation test in the PR plan (delete `:243` → tests go red) confirms the pin is load-bearing, not tautological.
- **Mock path & isolation correct.** `vi.mock("../commands/tree/skill-lib.js")` resolves to the same module `init.ts` imports; `...actual` keeps every other export (incl. those `bootstrapTreeRoot` relies on) real, so only the intended function is stubbed. `process.cwd()`/`process.exitCode` are saved and restored in `finally`.

No core data-structure / DB change, so no human-checkpoint needed. CI on the PR is green where complete (Branch Name / Detect Changes / Migrations / Forbid-legacy all pass); Test + Lint + CodeQL still in progress but this is a low-risk test addition.

🤖 Reviewed with Claude Code

PR #796, opened 2026-06-03
refactor(tree): scope managed agent hooks to workspace-root only

liuchao-001:
## Summary

Previously `first-tree tree upgrade` injected managed `.claude/settings.json` / `.codex/hooks.json` / `.codex/config.toml` into **every** target it touched: every bound source repo (including `workspace-member` child repos), the tree repo itself, plus a redundant outer call in `runUpgradeCommand` that fired a second time regardless of what the inner function had decided.

Hooks belong at the **agent's startup cwd** — the workspace root. Child repos and tree repos are not where agents start, so the injected hooks were dead weight there, and the duplication created drift surface for every future hook upgrade (each location had to be stripped + rewritten).

Direction set by @liuchao-001:

> 你要保证 agent hook 的注入只发生在 agent 启动的 cwd 层面，不 recursive 的注入到 tree repo 和 source repo 里面。

## New rule

Hook installation is now gated on binding mode:

| Target | Action |
| --- | --- |
| `workspace-root` source binding | install hooks (this is the agent's cwd) |
| `workspace-member` source binding | strip any managed hooks; never install |
| `shared-source` source binding | strip any managed hooks; never install |
| `standalone-source` source binding | strip any managed hooks; never install |
| Tree repo upgrade (`upgradeTreeRoot`) | strip any managed hooks; never install |

`first-tree tree claude-hook` (the explicit user opt-in command) is unchanged.

## Changes

**`agent-context-hooks.ts`:**

- New `removeAgentContextHooks(targetRoot)`. Strips first-tree managed entries from `.claude/settings.json` (SessionStart hooks matched by `isClaudeManagedHook`) and `.codex/hooks.json` (matched by `isCodexManagedHook`), plus the `codex_hooks = true` flag from the `[features]` block in `.codex/config.toml`. **Preserves user-added hooks and other settings in the same files.** Idempotent.
- `AgentConfigAction` extended with a `"removed"` variant.
- `formatAgentContextHookMessages` reports removals with a distinct message ("Removed first-tree SessionStart hook from ...").

**`upgrade.ts`:**

- `upgradeSourceRoot`: branches on `sourceBinding.bindingMode`. Only `workspace-root` triggers `ensureAgentContextHooks`; everything else triggers `removeAgentContextHooks`.
- `upgradeTreeRoot`: always calls `removeAgentContextHooks`.
[...]

PR #794, opened 2026-06-03
refactor(tree): stop generating tree-side AGENTS.md / CLAUDE.md

liuchao-001:
## Summary

Tree-side `AGENTS.md` / `CLAUDE.md` were carrying a framework block that duplicated agent behavior regulation already shipped through the `first-tree` skills and the workspace-level managed block. The tree-repo copy was also strictly weaker than `practices/tree-maintenance.md` and **factually wrong** about the gate model (claimed "Writing requires owner approval" while `self_owner` PRs auto-merge).

Companion to [first-tree-context#394](https://github.com/agent-team-foundation/first-tree-context/pull/394), which removed the existing tree-side files from the org's tree repo. This PR removes them from the tooling so future `first-tree tree init` calls don't reproduce what we just deleted.

Direction set by @liuchao-001:

> 事实上这个 tree repo 的 agents.md 和 claude.md 应该删掉。我们在 agent workspace 层级有 agents.md 在规范 agent 行为。tree-templates.ts 要跟上新shape。

## Where agent behavior regulation lives now

- **Workspace-level `AGENTS.md` / `CLAUDE.md`** — the managed `FIRST-TREE-SOURCE-INTEGRATION` block written by `upsertSourceIntegrationFiles` into each bound source/workspace root. Untouched by this PR.
- **`first-tree` / `first-tree-context` / `first-tree-write` skills** — the workflow / writing discipline layer.
- **`practices/tree-maintenance.md`** in the tree — the canonical gate / tiered write model.

Tree repos now hold just the org data (`NODE.md` domain map, leaves, `members/`, `.first-tree/` state).

## Where tree identity lives now

**`.first-tree/tree.json`** — written by `bootstrap.ts` (new, see "Bug fix" below). `readTreeIdentityContract()` reads from this; the legacy `AGENTS.md` framework block remains as a fallback so existing trees keep working.

## Changes

**`tree-templates.ts`:**
- Simplify `renderRootNode()` to the minimal "frontmatter + title + domain pointer" shape — drops the preamble, the "Working with the Tree" section, and the `See AGENTS.md` pointer
- Remove `renderTreeAgentsInstructions()` (the framework block source)

**`bootstrap.ts`:**
- Stop writing tree-root `AGENTS.md`
- Stop creating the `CLAUDE.md` → `AGENTS.md` symlink; drop the helper
- **Write `.first-tree/tree.json`** via `writeTreeState` (idempotent — skipped if already present, e.g. on republish)

**`verify.ts`:**
[...]

PR #823, opened 2026-06-05
fix(tree): resolve tree via workspace.json under the W1 layout

liuchao-001:
## Summary

`tree inject` returned **0 bytes from any cwd inside a W1 workspace** — including the workspace root itself, which is the exact cwd from which the managed SessionStart hooks installed by #796 fire. Effect: agents launched under W1 received an empty `additionalContext` payload even when everything else (hook installation, npm version, file permissions) was correct.

Found during end-to-end verification after #794/#796 merged, reported by @liuchao-001.

## Root cause

`resolveTreeContextRoot` only checked three locations after determining cwd was not itself a tree:

1. `<dirname(cwd)>/<treeName>` (pre-W1 sibling layout)
2. `<cwd>/.first-tree/tmp/<treeName>` (ephemeral working copies)
3. else → return null

Under W1 the tree lives at `<workspaceRoot>/<manifest.tree>` — an immediate subdirectory of the workspace root, recorded in `<workspaceRoot>/.first-tree/workspace.json`. None of the three locations checked above match. From workspace-root cwd, the sibling check looked for the tree as a peer of the workspace directory instead of inside it; from any deeper cwd, the binding contract pinned a sibling that did not exist.

## Fix

Add a canonical W1 resolution path that runs **before** the legacy sibling check:

- Walk up from `currentRoot` to discover the nearest `.first-tree/workspace.json` via `discoverWorkspaceRoot` — the same walker `first-tree tree status` uses.
- Read the manifest's `tree` field and join it onto the workspace root to get the canonical tree path.
- Verify a tree identity contract exists there before returning, so a workspace whose manifest points at a non-tree directory still falls through to the legacy fallbacks (the tmp-checkout path remains intact).

Because W1 covers cwds without a binding contract (e.g. a scratch dir inside the workspace), the new path runs even when `readSourceBindingContract(currentRoot)` returns `undefined`. The returned `currentEntrypoint` prefers `sourceBinding.entrypoint` when one is present; otherwise it derives a `/`-prefixed path relative to the workspace root.

## Tests

Extend `tree-extra-coverage.test.ts`:

- workspace-root cwd → tree resolved, entrypoint reported as `/` from the binding contract.
- workspace-member cwd → walk-up still finds the manifest; binding's own entrypoint (`/repos/product-repo`) is used.
- bindingless cwd inside the workspace (a scratch dir) → walk-up still works; entrypoint is derived from the filesystem (`/scratch/play`).
- explicit "legacy sibling, no manifest" case → fallback path still fires.
- W1 + legacy sibling both available → W1 wins (canonical layout has precedence).
- W1 manifest points at a non-tree path → fall through to legacy sibling resolution still finds the real tree.

## End-to-end verification with the dev CLI

```
$ first-tree-dev tree init --scope workspace --workspace-id acme --tree-mode shared
... bootstraps workspace + tree ...

$ first-tree-dev tree inject   # from workspace root
-> 790-byte SessionStart payload, entrypoint `/`

$ first-tree-dev tree inject   # from <ws>/repos/some-source
-> 807-byte SessionStart payload, entrypoint `/repos/some-source`

$ first-tree-dev tree inject   # from tree-repo root
-> 803-byte SessionStart payload, entrypoint `tree repo root`
```

All three produce valid `hookSpecificOutput.SessionStart` JSON; the first two are the previously-broken paths.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` — **506 passed** (was 502, +4 new)
- [x] `pnpm check` (biome) clean
- [x] Dev-installed CLI E2E covers workspace-root / workspace-member subdir / tree-repo root

## Related

- #794 (tree-side AGENTS.md / CLAUDE.md removed)
- #796 (agent hooks scoped to workspace-root only — the hooks installed there now finally get a non-empty payload from `tree inject`)

### DIFF
--- a/apps/cli/src/__tests__/tree-extra-coverage.test.ts
+++ b/apps/cli/src/__tests__/tree-extra-coverage.test.ts
@@ -8,6 +8,7 @@ import { writeTreeState } from "../commands/tree/binding-state.js";
 import { buildSourceIntegrationBlock } from "../commands/tree/source-integration.js";
 import { syncTreeIdentityFiles } from "../commands/tree/tree-identity.js";
 import type { CommandContext } from "../commands/types.js";
+import { writeWorkspaceManifest } from "../core/workspace.js";
 
 const tempDirs: string[] = [];
 const originalCwd = process.cwd();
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
## Review: Approve ✅

**Goal & core change** — `tree inject` returned 0 bytes from any cwd inside a W1 workspace (including the workspace root, where #796's managed SessionStart hooks fire), because `resolveTreeContextRoot` only knew the pre-W1 sibling/tmp layouts. The fix inserts a canonical W1 resolution path (Case 2) that walks up to the nearest `.first-tree/workspace.json` via `discoverWorkspaceRoot`, reads `manifest.tree`, and resolves `<workspaceRoot>/<manifest.tree>` — gated on an actual tree identity contract existing there.

### What I verified
- **Interface alignment** — `discoverWorkspaceRoot` / `readWorkspaceManifest` / `writeWorkspaceManifest` in `core/workspace.ts` match the signatures used here exactly; no drift.
- **Resolution order is correct** — Case 1 (cwd is tree root) still wins first; W1 runs before the legacy sibling/tmp fallbacks; W1 vs legacy precedence (W1 wins) is the documented canonical behavior.
- **Failure modes are guarded** — malformed/unreadable manifest → `try/catch` returns null and falls through; manifest `tree` pointing at a non-tree → identity-contract check returns null → legacy fallback still finds the real sibling. Both are covered by tests.
- **Entrypoint derivation** — prefers `sourceBinding.entrypoint` when a binding contract is present, else derives `/`-prefixed relative path; `discoverWorkspaceRoot` guarantees the workspace is an ancestor so no `..` escapes.
- **Tests** — 6 new cases cover workspace-root / member-subdir / bindingless-scratch / legacy-sibling-fallback / W1-wins / non-tree-fallthrough. Good coverage of every branch.
- **CI green** — Test, Lint & Type Check, Forbid-legacy, and CodeQL all pass on this head.

This is a read-time resolver fix — no data-structure or DB change, so no extra human sign-off needed.

### Non-blocking nits (optional)
1. `computeCurrentEntrypoint` uses `path.relative` then prefixes `/`. On Windows that yields `/scratch\play` (backslash). The codebase is posix-oriented so this is cosmetic, but `rel.split(sep).join('/')` would make it robust if cross-platform ever matters.
2. `discoverWorkspaceRoot` walk-up will resolve to the nearest ancestor `workspace.json`, so a legacy sibling layout nested inside an unrelated parent W1 workspace would resolve to the parent's tree. Extreme edge case and consistent with the intended "W1 is canonical" precedence — noting only for awareness.

Neither blocks merge. Nice tight fix with thorough regression coverage.

PR #826, opened 2026-06-05
refactor(tree): hard-delete vestigial init flags + split SourceBindingMode + reject pre-W1 upgrade (audit PR-C)

liuchao-001:
## Summary

PR-C of the post-W1 trailing edge cleanup, last in the D → E → A → B → C sequence (D / E / A / B all merged). Closes the remaining audit findings in one PR:

- **Findings 4, 5, 6**: hard-delete \`--tree-mode\` / \`--scope\` / \`--tree-name\` parser path (PR-B was deprecate-and-warn for one release cycle).
- **Finding 7**: split \`SourceBindingMode\` into \`W1BindingMode\` + \`LegacyBindingMode\`.
- **Finding 2b**: \`upgradeSourceRoot\` rejects pre-W1 bindings with a \`migrate-to-w1\` pointer; the source-root \`WHITEPAPER.md\` writer at \`upgrade.ts:60\` becomes unreachable and the function plumbing goes with it.
- **\`source-integration.ts:107\` dead fallback**: \`?? "standalone-source"\` flipped from a silent default to a loud guard.

## Applying both reviewer-rules

Reviewer ran a **(lifecycle × target) sweep** on PR-A and a **deprecation-honesty sweep** on PR-B. PR-C applies both proactively:

- **axis-1**: grep covers every site that still mentioned the 3 deleted flags / \`SourceBindingMode\` / \`WHITEPAPER\` / \`ensureWhitepaperSymlink\`.
- **axis-2**: trace verifies every reachable caller carries a W1 binding, not the dead-fallback default.

## CLI surface (\`apps/cli/src/commands/tree/init.ts\`)

- Hard-delete \`--tree-mode\`, \`--scope\`, \`--tree-name\` from the parser. Commander now surfaces them as \"unknown option\".
- Drop \`InitOptions.scope\` and \`InitOptions.treeName\`. \`treeMode?\` stays as a programmatic-only override of the inert tree-state metadata field.
- Delete \`REPO_SCOPE_GUIDANCE\`, \`warnDeprecatedFlag\`, \`readScopeOption\`, \`readTreeModeOption\`.
- \`resolveTreeMode\` keeps PR-B's URL-derived default; explicit-flag branch is now programmatic-only.
- \`resolveTreeRoot\` drops the \`options.treeName\` fallback.

## SourceBindingMode split (\`binding-state.ts\`)

\`\`\`ts
export type W1BindingMode = \"workspace-root\" | \"workspace-member\";
export type LegacyBindingMode = \"standalone-source\" | \"shared-source\";
export type SourceBindingMode = W1BindingMode | LegacyBindingMode;

export function isLegacyBindingMode(mode: SourceBindingMode | string): mode is LegacyBindingMode { ... }
\`\`\`

\`SourceBindingMode\` (the union) stays as a re-export for the binding-contract parser and the describeBinding / describeScope legacy fallbacks, which still have to handle either shape.

## upgradeSourceRoot rejects pre-W1 (\`upgrade.ts\`)

Early-return throw with \`migrate-to-w1\` pointer if the source binding contract resolves to a legacy value. Implicitly closes audit Finding 2b — the source-root \`WHITEPAPER.md\` writer becomes unreachable, so the helper plumbing (\`ensureWhitepaperSymlink\`, \`readManagedWhitepaperTarget\`, \`removeManagedWhitepaper\`) is removed from \`source-integration.ts\` entirely. The migrate cleanup path in \`apps/cli/src/core/migrate-workspace.ts\` has always used its own \`tryRemove(WHITEPAPER.md)\` and is unaffected.

## source-integration.ts:107 closure

Previous code: \`const bindingMode = options?.bindingMode ?? \"standalone-source\"\`. After PR-C reject, every reachable caller carries a W1 binding mode. The fallback now throws — turning a dead default into a loud guard.

## Test changes

- delete \`init-deprecated-flags-warn.test.ts\` entirely.
- \`init-exit-code-on-missing-skills.test.ts\` (PR-D's PIN): argv simplified.
- \`tree-extra-coverage.test.ts\`: \`rejects --scope repo\` becomes \`--scope flag is removed; commander surfaces it as unknown option\`.
- \`tree-maintenance.test.ts\`:
  - \`upgrades a bound source root\` switches the fixture from \`shared-source\` to \`workspace-member\`.
  - \`shared-source: does not install hooks\` and \`standalone-source: does not install hooks\` flip to assert \`.toThrowError(/migrate-to-w1/u)\`.
  - New focused PIN \`refuses to upgrade a pre-W1 source root and points the user at migrate-to-w1 (PR-C)\` exercises both legacy values.
- \`tree-lifecycle.test.ts\`: \`rejects scope=repo with guidance\` deleted (API surface gone).
- \`tree-helper-coverage-extra.test.ts\`: WHITEPAPER helper coverage dropped; workspace-layout invariant assertion stays; \`buildSourceIntegrationBlock\` call passes a W1 \`bindingMode\`.
- \`cli-entry-and-tree-actions-extra.test.ts\`: upgrade-source fixture flipped to \`workspace-member\`.

## 3 mandated smokes

- [x] **W1 fresh init**: \`tree init --tree-path ./tree --workspace-id smoke1\` writes a valid \`workspace.json\` with no deprecation warnings on stderr.
- [x] **Upgrade on a pre-W1 \`shared-source\` binding** throws with \`migrate-to-w1\` pointer (vitest fixture using \`buildSourceIntegrationBlock\` to produce a real legacy AGENTS.md block).
- [x] **\`migrate-to-w1 --dry-run\` still recognizes the legacy \`.first-tree-workspace\` marker** — legacy parser preserved.

## Test plan

- [x] \`pnpm check\` — Biome, 1213 files clean.
- [x] \`pnpm typecheck\` — green across all packages.
- [x] \`pnpm --filter first-tree-dev test\` — 70 test files / 509 tests pass.
- [x] 3 mandated smokes above all pass.

**-318 / +180, net -138 lines.**


yuezengwu (approved):
Reviewed against a fresh `origin/main` worktree. LGTM — approving.

**What this does:** trailing-edge cleanup (PR-C of D→E→A→B→C). Hard-deletes the three vestigial `init` parser flags (`--tree-mode` / `--scope` / `--tree-name`) one release cycle after PR-B's deprecate-and-warn, splits `SourceBindingMode` into `W1BindingMode` + `LegacyBindingMode` (union re-exported for back-compat), makes `upgradeSourceRoot` reject pre-W1 bindings with a `migrate-to-w1` pointer, and removes the now-unreachable WHITEPAPER source-root plumbing. Net −138.

**Verified:**
- **No dangling callers.** Repo-wide string grep (apps / packages / skills / docs, not just the TS import graph) for the deleted symbols (`ensureWhitepaperSymlink`, `readManagedWhitepaperTarget`, `removeManagedWhitepaper`, `warnDeprecatedFlag`, `readScopeOption`, `readTreeModeOption`, `REPO_SCOPE_GUIDANCE`) and the deleted flag strings: only comments / docs guidance remain. The `--scope` hits in `first-tree-sync` audit-workflow are an unrelated command's own flag.
- **Migrate path untouched.** `core/migrate-workspace.ts` uses its own `SOURCE_WHITEPAPER_FILE` + `tryRemove` and still recognizes the legacy `standalone-source` / `shared-source` values — independent of the removed helpers, as claimed.
- **Type split is compile-time only.** The four on-disk string-literal values are unchanged; no DB migration (Migrations check green). So no data-structure / schema change to escalate.
- **CI green** on the PR's own run (Test, Lint & Type Check, CodeQL, Forbid-legacy-names all SUCCESS).

**Minor, non-blocking** (no change required): in `upgradeSourceRoot`, the friendly legacy-reject guard requires `bindingMode !== undefined`. A malformed/ancient binding block carrying the source-integration marker but *no* binding-mode marker (so `bindingMode` parses to `undefined`) would skip that guard and instead hit the new `deriveSourceIntegrationDetails` throw — surfacing the internal "called without a bindingMode" message rather than the `migrate-to-w1` pointer. This is strictly better than the old silent `?? "standalone-source"` (it now errors instead of rebuilding legacy artifacts); only the error text is less guided for that edge case. Fine to leave as-is.

PR #858, opened 2026-06-08
[codex] rename agent runtime dir to .first-tree-workspace

liuchao-001:
## Summary
- rename the stable agent runtime state root from `.agent/` to `.first-tree-workspace/`
- make `.first-tree-workspace/` serve as both the Codex workspace marker and the runtime state directory
- auto-migrate legacy `.agent/` state and prune legacy `.agent/tools.md` / `.agent/context/` on bootstrap

## Why
The previous layout split runtime state across a root marker file and a separate `.agent/` directory. This change collapses them into one runtime directory while preserving workspace boundary detection and startup self-healing.

## Validation
- `pnpm --dir packages/client test`
- `pnpm --dir packages/client typecheck`

### DIFF
--- a/packages/client/src/__tests__/agent-bootstrap.test.ts
+++ b/packages/client/src/__tests__/agent-bootstrap.test.ts
@@ -8,7 +8,10 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 const state = vi.hoisted(() => ({ cachedTreeHead: null as string | null, cachedCli: null as string | null }));
 
 vi.mock("../runtime/bootstrap.js", () => ({
+  FIRST_TREE_RUNTIME_DIR: ".first-tree-workspace",
+  IDENTITY_JSON_REL: ".first-tree-workspace/identity.json",
   bootstrapWorkspace: vi.fn(),
+  ensureWorkspaceRuntimeDir: vi.fn(),
   installCoreSkills: vi.fn(),
   installFirstTreeIntegration: vi.fn(),
   deepEqualIdentity: vi.fn(() => false),
@@ -62,7 +65,6 @@ describe("ensureAgentBootstrap — integration retry gate", () => {
     // fast path forever.
     const sentinel = join(workspace, INIT_COMPLETE_SENTINEL_REL);
     mkdirSync(dirname(sentinel), { recursive: true });
-    mkdirSync(join(workspace, ".agent"), { recursive: true });
     writeFileSync(sentinel, "1", "utf-8");
 
     state.cachedTreeHead = null;
--- a/packages/client/src/__tests__/agent-cwd-redesign.e2e.test.ts
+++ b/packages/client/src/__tests__/agent-cwd-redesign.e2e.test.ts
@@ -26,7 +26,7 @@ import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
  *       On Demand"; does NOT contain legacy "Predeclared worktrees".
  *   E5  Concurrent two-chat start mutex serialises filesystem writes —
  *       no race throws or missing checkouts.
- *   E6  `.agent/identity.json` carries agent-level stable fields only —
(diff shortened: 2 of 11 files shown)


yuezengwu:
## Review — rename agent runtime dir to `.first-tree-workspace/`

**What this does:** collapses the two previously-separate pieces of per-agent runtime state — the root marker *file* `.first-tree-workspace` and the stable runtime *directory* `.agent/` — into a single directory `.first-tree-workspace/` that serves as both the Codex `project_root_markers` boundary marker and the runtime-state home (`identity.json`, `init-complete`, `cli-version`, `context-tree-head`). On bootstrap it self-heals legacy layouts: migrates an old `.agent/` dir, replaces a stale marker file / dangling symlink, and prunes legacy `.agent/tools.md` and `.agent/context/`.

**Local verification (PR head):**
- `pnpm build` ✅ · `pnpm typecheck` (repo-wide) ✅ · `bootstrap.test.ts` + `workspace.test.ts` → 61/61 ✅
- Migration logic reads correctly: `mergeLegacyRuntimeDir` is *target-wins* — it only `renameSync`/`rmSync`-es legacy `.agent/` source entries and never overwrites existing current-runtime state, so repeated bootstraps and partial upgrades stay idempotent. The dangling-symlink, old-file-marker, and dir/file collision cases are all covered by tests.
- Security: `.first-tree-workspace/`, like `.agent/`, starts with `.`, so the doc-snapshot hidden-segment rejection still shields runtime files from mention-leaks — no regression.
- Core directory-marker assumption is sound: Codex `project_root_markers` matches directories (same mechanism `.git` uses).

The code itself is LGTM. Two **required checks are red** and the fixes aren't in the diff yet — blockers for merge, both mechanical:

**Finding 1 — `Lint & Type Check` is red (Biome `pnpm check`).** Not a type error; it's formatting + import-ordering in `packages/client/src/__tests__/bootstrap.test.ts` (organize-imports ordering of the new `FIRST_TREE_RUNTIME_DIR` / `IDENTITY_JSON_REL` imports, plus one over-wrapped `.toBe(...)` line). `pnpm check --write` fixes it cleanly (2 files), after which `pnpm check` is green. The PR's Validation section listed `test` + `typecheck` but not `pnpm check`, which is why this slipped through.

**Finding 2 — `Check Branch Name` is red.** The branch `codex/rename-agent-home` doesn't match the enforced `^(feat|fix|refactor|test|docs|chore|merge)/` convention (and `codex/*` isn't in the skip-list, unlike `hub-session-*`). Re-push under e.g. `refactor/rename-agent-home`.

**Non-blocking note:** `.first-tree-workspace` is now overloaded — a *file* marker for CLI W1 user-workspace detection (`apps/cli/src/core/migrate-workspace.ts`) and a *directory* for the per-agent runtime home. They live at different filesystem locations and the docblock on `ensureWorkspaceRuntimeDir` calls this out, so it's fine today, but it's a latent footgun worth keeping in mind for anyone who later wires the two paths together.

Happy to approve once Findings 1 & 2 turn the checks green — no need to wait on me for re-review of the logic.

yuezengwu (approved):
✅ Approving — both findings from the prior review are resolved and all required checks are green.

- **Finding 1 (Biome `pnpm check`)** — fixed in `format client tests`; `Lint & Type Check` now passes.
- **Finding 2 (branch name)** — branch renamed to `refactor/rename-agent-home`; `Check Branch Name` now passes.

`Lint & Type Check`, `Test`, `Forbid legacy CLI / env names`, `Migrations`, and `Check Branch Name` are all green. The migration logic (target-wins, idempotent, edge-cases covered) and the directory-marker change were verified firsthand in the earlier pass — no code changes since then beyond the formatting fix. Good to merge.

(Carry-forward, non-blocking: `.first-tree-workspace` now means a *file* marker in the CLI W1 user-workspace path and a *directory* in the per-agent runtime home — fine today, just a latent footgun if those paths ever converge.)

PR #866, opened 2026-06-08
fix: recover inbox delivery on suspend

baixiaohang:
## Summary
- add same-socket `inbox:recover` frames and chat-scoped recovery for delivered-but-unacked inbox rows
- ACK only handler-reported consumed provider-turn messages on suspend, leaving queued/unconsumed input for recovery
- serialize ACK, recovery reset, and delivery drain on the server socket FIFO, with entry-id based in-flight accounting
- fail closed after local in-flight tracking is cleared without ACK until chat recovery runs

## Context
- Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/448
- Local design doc intentionally not included: `docs/session-recovery-ack-design.md`

## Validation
- pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts src/__tests__/session-manager-edge-coverage.test.ts
- pnpm --filter @first-tree/client exec vitest run src/__tests__/client-connection-websocket-coverage.test.ts
- pnpm --filter @first-tree/shared test -- inbox-frames-schema.test.ts
- pnpm --filter @first-tree/server exec vitest run src/__tests__/inbox-bind-recovery.test.ts src/__tests__/inbox-ws-push.test.ts
- pnpm check
- pnpm typecheck

### DIFF
--- a/packages/client/src/__tests__/client-connection-websocket-coverage.test.ts
+++ b/packages/client/src/__tests__/client-connection-websocket-coverage.test.ts
@@ -390,6 +390,48 @@ describe("ClientConnection — WebSocket edge coverage", () => {
     await expect(ackPromise).resolves.toBeUndefined();
   });
 
+  it("sends inbox recovery requests and settles on accepted or rejected frames", async () => {
+    const connection = await makeConnection();
+    const internal = priv(connection);
+    const socket = await openRegisteredConnection(connection, { wsInboxAckConfirm: true });
+
+    const bindPromise = internal.sendBind("agent-1", "codex");
+    const bindFrame = parseSent(socket, socket.sent.length - 1);
+    socket.emitMessage({
+      type: "agent:bound",
+      ref: bindFrame.ref,
+      agentId: "agent-1",
+      displayName: "Agent One",
+      agentType: "agent",
+    });
+    await bindPromise;
+
+    const accepted = connection.sendInboxRecover("agent-1", "chat-1");
+    const recoverFrame = parseSent(socket, socket.sent.length - 1);
+    expect(recoverFrame).toMatchObject({ type: "inbox:recover", agentId: "agent-1", chatId: "chat-1" });
+    expect(typeof recoverFrame.ref).toBe("string");
+
+    socket.emitMessage({
+      type: "inbox:recover:accepted",
+      ref: recoverFrame.ref,
+      agentId: "agent-1",
+      chatId: "chat-1",
+      resetCount: 1,
+    });
+    await expect(accepted).resolves.toBeUndefined();
+
+    const rejected = connection.sendInboxRecover("agent-1", "chat-2");
+    const rejectFrame = parseSent(socket, socket.sent.length - 1);
+    socket.emitMessage({
+      type: "inbox:recover:rejected",
+      ref: rejectFrame.ref,
+      agentId: "agent-1",
+      chatId: "chat-2",
+      reason: "recover_failed",
+    });
+    await expect(rejected).rejects.toThrow("recover_failed");
+  });
+
   it("holds agent-scoped confirmed inbox ACKs until that agent is rebound on the current socket", async () => {
     const connection = await makeConnection();
     const internal = priv(connection);
--- a/packages/client/src/__tests__/session-manager-edge-coverage.test.ts
+++ b/packages/client/src/__tests__/session-manager-edge-coverage.test.ts
@@ -132,6 +132,7 @@ function makeManager(
     handlers?: AgentHandler[];
     handlerFactory?: HandlerFactory;
     ackEntry?: (entryId: number) => Promise<void>;
+    recoverChat?: (chatId: string) => Promise<void>;
     registryPath?: string;
     concurrency?: number;
     maxSessions?: number;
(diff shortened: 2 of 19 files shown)


yuezengwu (changes requested):
## Review: recover inbox delivery on suspend

Read against the inbox-delivery context node (delivery / deferred-ACK / reconnect-recovery contract). Companion context PR: first-tree-context PR 448.

### What this changes (core)
Adds a same-socket `inbox:recover` frame so a chat's delivered-but-unacked rows can be reset→pending and redelivered **without waiting for a reconnect/bind reset**. The runtime now distinguishes `accepted` (admitted) from a new `consumed` flag (message actually entered the provider turn). On `session:suspend` it **ACKs the consumed prefix** and clears the unconsumed tail for recovery; a later inbound into a chat without a healthy live handler first asks the server to recover, then proceeds on redelivery. Server serializes ACK/recover/drain on one socket FIFO with entry-id-based in-flight accounting.

Direction looks sound and the in-flight state machine is well covered in the two updated session-manager test files.

### 🔴 Blocker — CI Test is red with PR-introduced regressions
The `Test` job fails with **3 deterministic failures** (not flakes) in two test files this PR did not update:

- `packages/client/src/__tests__/session-start-error-signaling.test.ts:312` — expected `'errored'`, got `'suspended'`
- `packages/client/src/__tests__/session-start-error-signaling.test.ts:361` — recovery `start` called 0 times, expected 1
- `packages/client/src/__tests__/session-state-notifications.test.ts:162` — missing the `'active'`-on-resume transition

Root cause: both files still use the **old single-dispatch + `noteBindRecoveryComplete()`** pattern and build the SessionManager **without a `recoverChat` callback**. Under the new code, eviction/suspend sets `requiresInboxRecovery`, so the next same-chat dispatch routes through `recoverChatBeforeDispatch` and returns early — with no `recoverChat` wired it logs an error and never resumes/errors. This is the same migration already applied to `session-manager.test.ts` / `session-manager-edge-coverage.test.ts`; these two were missed.

Required before merge: update both files to the new model (wire a `recoverChat` mock, double-dispatch to model recover→redelivery, drop the now-no-op `noteBindRecoveryComplete()`). For `session-start-error-signaling.test.ts` specifically, please **verify the resume-failure `'errored'` + structured error event still fires after the recover round-trip** rather than just relaxing the assertions — that test guards a real "don't swallow resume errors" contract.

### Notes for the inbox-delivery owner (non-blocking, please confirm)
1. **Suspend semantics shift.** Suspend previously abandoned all in-flight entries for recovery; it now ACKs the consumed prefix. Correctness relies on (a) handlers marking consumption strictly in order so `consumed` stays a prefix, and (b) the provider session durably holding consumed input so a graceful suspend won't lose it. Both hold for the claude-code / tui / codex handlers today; worth stating explicitly in the contract since it narrows at-least-once on the suspend path.

### Design observations (non-blocking)
2. **Cold-dispatch cost.** `shouldRecoverBeforeDispatch` triggers a recover round-trip on *any* dispatch into a chat without an active handler — including brand-new chats with nothing to reset. Since idle-suspend is common, many cold turns now pay an extra recover + reset/re-claim. Correct but worth measuring.
3. **Concurrent redelivery churn.** Client `inbox:deliver` frames dispatch fire-and-forget (agent-slot), so when recovery redelivers N entries, a straggler can re-enter `recoverChatBeforeDispatch` before the first has marked the session active, firing another recover. The local dedup guard + server FIFO make this safe (no double-process), but it can produce repeated recover cycles + log noise under multi-entry cold delivery.
4. Minor: `inboxRecoverRejectedReasonSchema` includes `chat_id_required`, but malformed frames are caught by `safeParse` and answered with a generic `error` frame, so that reason is never emitted. Dead value.

Not approving while CI is red on PR-caused failures. Happy to re-review once the two test files are migrated and Test is green.

baixiaohang:
Fixed the blocking CI failures in 02d42ce9 (`test: update recovery resume expectations`).

What changed:
- migrated `session-start-error-signaling.test.ts` and `session-state-notifications.test.ts` to wire a `recoverChat` mock
- updated the old single-dispatch / `noteBindRecoveryComplete()` resume cases to model recover -> redelivery with double dispatch
- kept the resume-failure assertions intact: the test still verifies the `errored` state and structured `Session resume failed` event after the recovery round-trip

Validation after the fix:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-start-error-signaling.test.ts src/__tests__/session-state-notifications.test.ts`
- `pnpm check`
- `pnpm typecheck`
- GitHub `Test` job is now green on head `02d42ce9`

On the non-blocking notes: confirmed on suspend semantics. Correctness intentionally relies on handlers marking provider-turn consumption strictly as a prefix and on provider sessions retaining consumed input across graceful suspend; Context Tree PR 448 records that contract. The cold-dispatch cost and possible repeated recover churn are correctness-safe but worth measuring after merge. The unused `chat_id_required` recover rejection reason is minor and can be cleaned up separately unless you want it removed in this PR.

yuezengwu (approved):
Approving — blocker resolved.

Verified on head `02d42ce9`:
- `Test`, `Lint & Type Check`, CodeQL, and `Migrations` are all green.
- The two flagged test files (`session-start-error-signaling.test.ts`, `session-state-notifications.test.ts`) are migrated correctly: `recoverChat` is wired, the old single-dispatch / `noteBindRecoveryComplete()` cases now model recover→redelivery via double dispatch, and the resume-failure assertions are **kept intact** — the `'errored'` state and the structured `Session resume failed` error event are still asserted after the recovery round-trip. That was the part I cared about, so this is a real fix, not a relaxed assertion.

Non-blocking, for follow-up (not gating this PR):
- Suspend semantics narrowing (ACK consumed prefix) is acknowledged and recorded in context PR 448 — good.
- Cold-dispatch recover round-trip and the concurrent-redelivery recover churn are correctness-safe (dedup + socket FIFO); worth measuring post-merge.
- `chat_id_required` dead reason: fine to clean up separately.

LGTM.

PR #867, opened 2026-06-08
fix: degrade transient source-repo fetch failures to the existing checkout (#865)

### LINKED ISSUE #865: Discuss degrading source-repo Git failures during agent session start
## Summary

A colleague hit a local Git/GitHub connectivity problem, and it caused the First Tree client to fail agent session start/resume while preparing predeclared source repos. As a result, the pinned agents became unresponsive and could not be used to continue debugging the incident from inside First Tree.

This issue is to discuss whether source-repo Git failures should remain fail-fast, or whether the runtime should degrade more gracefully when the failure looks like a transient network/GitHub access problem.

## Observed Failure

During session start/resume, the daemon prepared source repos and the underlying Git operation could not reach `github.com`:

```text
GitMirrorError: git fetch --prune origin exited with code 128:
fatal: unable to access 'https://github.com/agent-team-foundation/first-tree/':
Error in the HTTP2 framing layer

msg: "session start/resume failed", phase: "resume"

git clone ... error:
RPC failed; curl 28 Failed to connect to github.com port 443 after 75002 ms:
Couldn't connect to server
```

The old daemon process kept logging this from roughly 16:00 to 16:11. From the user's perspective, the client was up enough to be failing work, but agents were not usable, so the normal agent-assisted debug path was unavailable.

## Current Behavior

Predeclared `runtimeConfig.gitRepos` are materialized before the handler starts/resumes. The current per-agent clone model fetches the configured source repo on each session and then updates the checkout when safe.

When `git fetch` or `git clone` fails, the error bubbles out of source-repo preparation and aborts the session start/resume. That means a transient GitHub/network issue can make an otherwise healthy agent unable to answer at all.

## Why This May Need A Fix

Fresh source repos are important, but network access to GitHub is an external dependency. If the local machine already has a usable checkout, a transient `fetch` failure may be better represented as a degraded/stale source-repo state rather than a hard session failure.

The dangerous alternative is silently continuing as if the repo were fresh. If we degrade, the stale/unavailable state needs to be visible to both the agent and the human/operator.

## Questions To Decide

- If a local standalone clone already exists, should transient `fetch` failures allow the session to start using the last local checkout, with an explicit stale-source warning?
- If the repo has never been cloned and `git clone` fails, should the session still start in a "source repo unavailable" mode for discussion/debug tasks, or should it remain blocked?
- Should source-repo Git failures be classified into transient network failures vs hard failures such as auth failure, wrong origin, corrupt checkout, path conflict, or unsafe local state?
- Should repeated network failures use backoff/circuit-breaking so every session start/resume does not immediately retry GitHub and spam logs?
- Where should the degraded state surface: session status, daemon logs, `daemon doctor`, Web Console computer/agent health, AGENTS.md briefing, or some combination?
- Which behavior should apply for code-writing tasks where stale source may be materially unsafe?

## Related History

The root fail-fast policy for configured `gitRepos` appears to date back to PR #95 (`agent configuration M1`), where source repo preparation was wired into session start/resume and unreachable repo URLs aborted session creation.

The current standalone-clone shape and the visible `git fetch --prune origin` path are from PR #854, which retired the shared `git-mirrors` layer and made each agent own a real source repo clone under its agent home.

## Possible Acceptance Direction

A fix could be considered acceptable if:

- transient GitHub/network failures do not prevent an agent session from starting when a safe local checkout already exists;
- the agent and operator can clearly see that the source repo is stale or unavailable;
- first-clone failures produce an actionable blocked/degraded state instead of an opaque repeated session failure;
- hard safety failures still fail closed;
- repeated transient failures are throttled or circuit-broken enough to avoid log spam and retry storms.

yuezengwu:
## Summary

Fixes #865 — a transient GitHub/network failure while preparing predeclared source repos aborted agent session start/resume, making the agent unusable (including for debugging the very outage).

When a usable clone of the **same** repo already exists on disk, a transient fetch failure now degrades to the last-good local checkout (new `stale-offline` outcome) instead of failing the session — the same "left at current commit" contract as the existing `skipped-*` outcomes. No new degraded state/mode and no recovery machinery: the next session's fetch recovers it implicitly, and because start now succeeds, the retry-storm / log-spam loop from the incident disappears on its own.

## Safety boundary (fails closed)

Degrade applies only when **all** hold; otherwise the failure still bubbles up:
- origin already canonically matched the configured upstream **before** `reconcileOrigin` (no repoint — never serve repo A's checkout as repo B);
- high-confidence transient *network* error (not auth / corrupt / wrong-origin / TLS-trust / our own timeout / repo-not-found);
- a real standalone clone with a resolvable HEAD;
- any pinned `ref` already resolves locally.

First-clone failures (no local checkout to fall back to) stay fail-fast.

## Also

`isLikelyTransientNetworkError` now matches the incident's own error strings (`Error in the HTTP2 framing layer`; curl 28 `Failed to connect ... Couldn't connect to server`), which were previously unmatched — without this, the degrade would not trigger for the reported failure.

## Tests

- same-repo transient fetch failure → `stale-offline` (deterministic: origin pre-matched + unreachable endpoint)
- origin repoint + transient failure → fail-closed (boundary flagged in review by codex-developer and reproduced by QA)
- hard non-transient failure with an existing checkout → fail-closed
- incident error strings added to the transient-predicate cases

`@first-tree/client`: 90 files / 1010 tests passed; typecheck + biome clean. Verified on latest `main` (17368b27).

## Not in scope (per the design discussion on #865)

No `source-unavailable` first-clone mode, no AGENTS.md briefing / status / Web-Console surfacing, no circuit-breaker/recovery state machine — deliberately avoided as over-engineering. Surfacing is a single daemon `warn` log, same as `skipped-*`.

### DIFF
--- a/packages/client/src/__tests__/git-mirror-manager.test.ts
+++ b/packages/client/src/__tests__/git-mirror-manager.test.ts
@@ -284,6 +284,63 @@ describe("GitMirrorManager — source repo lifecycle (per-agent clone)", () => {
     expect(upstream).not.toBe(localHead);
   });
 
+  it("degrades to the existing checkout when a SAME-repo fetch fails transiently (stale-offline)", async () => {
+    // Issue #865: an existing, usable clone of the SAME repo + a transient
+    // network fetch failure must NOT abort — leave the checkout at its current
+    // commit and continue on the last-good source so the agent stays answerable.
+    //
+    // Deterministic repro of the incident shape: build a real standalone clone
+    // whose origin ALREADY matches the configured upstream, then point that
+    // upstream at an unreachable endpoint so the next fetch fails transiently
+    // (connection refused). Because origin already matches the configured url,
+    // reconcileOrigin does NOT repoint (this is the "same repo" path, not a
+    // repoint). Port 1 ≠ 443 also disables the https→ssh protocol fallback,
+    // keeping it a same-protocol transient failure.
+    const { url: bare, tip } = seedBareRepo();
+    const clonePath = join(newDataDir(), "repo");
+    execSync(`git clone -q ${bare} ${clonePath}`);
+    const unreachable = "https://127.0.0.1:1/same-repo.git";
+    execSync(`git remote set-url origin ${unreachable}`, { cwd: clonePath });
+    const m = makeManager();
+    const r = await m.ensureSourceRepo({ url: unreachable, clonePath });
+    expect(r.outcome).toBe("stale-offline");
+    expect(r.headCommit).toBe(tip); // left at current commit, not aborted
+    expect(gitIn(clonePath, "rev-parse HEAD")).toBe(tip);
+    // gitWithNetworkRetry burns its full transient-retry budget (~3 backed-off
+    // attempts) before the degrade fires, so allow more than the 5s default.
+  }, 30_000);
+
+  it("fails closed when origin is being repointed to a different repo, even on a transient fetch failure", async () => {
+    // Boundary flagged by codex-developer + reproduced by QA on issue #865: the
+    // degrade must apply only to "same repo, fetch temporarily unavailable", NOT
+    // to "configured repo changed but the confirming fetch could not run". Here
+    // the configured url repoints to a DIFFERENT repo whose fetch fails
+    // transiently — serving repo A's checkout as repo B would be wrong, so this
+    // must fail closed.
+    const { url, tip } = seedBareRepo();
+    const m = makeManager();
+    const clonePath = join(newDataDir(), "repo");
+    const first = await m.ensureSourceRepo({ url, clonePath });
+    expect(first.headCommit).toBe(tip);
+    await expect(m.ensureSourceRepo({ url: "https://127.0.0.1:1/different.git", clonePath })).rejects.toBeInstanceOf(
+      GitMirrorError,
+    );
+  }, 30_000);
+
+  it("still fails closed on a hard (non-transient) fetch failure even with an existing checkout", async () => {
+    // Negative case for the stale-offline degrade: a deterministic, non-network
+    // failure (repository does not exist) must still bubble up — degrading would
+    // mask a real, non-self-healing problem.
+    const { url, tip } = seedBareRepo();
+    const m = makeManager();
+    const clonePath = join(newDataDir(), "repo");
+    const first = await m.ensureSourceRepo({ url, clonePath });
+    expect(first.headCommit).toBe(tip);
+    await expect(
+      m.ensureSourceRepo({ url: "file:///first-tree-nonexistent-repo-865.git", clonePath }),
+    ).rejects.toBeInstanceOf(GitMirrorError);
+  });
+
   it("throws a conflict for a non-managed occupant with no managed root (preserves operator data)", async () => {
     const m = makeManager();
     const clonePath = join(newDataDir(), "repo");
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #867

**建议**: request_changes
- 理由: `stale-offline` can report the configured source repo `ref` while leaving the checkout on a different local HEAD.

**分级**: A
- 语义升级: 无

**PR 摘要**
- 解决问题: transient network fetch failure during predeclared source-repo preparation no longer aborts agent session start/resume when a usable same-repo checkout exists.
- 设计方案: add a `stale-offline` outcome gated by canonical origin match, transient network classification, standalone clone, resolvable local HEAD, and local `ref` resolution; surface it through source-repo session logs.
- 受影响模块: `packages/client/src/runtime/git-mirror-manager.ts`, `packages/client/src/runtime/source-repos.ts`, source-repo lifecycle tests.

**Review 意见**
❌ 1. `stale-offline` only checks that `ref` resolves locally; it does not ensure the checkout is actually at that ref before returning. `gitRepos.ref` accepts branch/tag/commit, and `prepareSourceRepos` still surfaces `repo.ref` to the runtime/briefing. So if an existing same-repo clone is currently on `main`, config changes to a pinned SHA or branch that already exists locally, and fetch fails transiently, this guard passes and returns the old current HEAD while the agent is told the repo is at the requested ref. For pinned commits this violates the runtime contract that the pinned commit is honored as-is. Please either apply the locally-resolved ref safely before returning `stale-offline`, or fail closed unless `HEAD` already matches the requested local ref. [R4 / packages/client/src/runtime/git-mirror-manager.ts:726]
✅ 2. The pre-reconcile origin capture and the repoint/non-transient negative tests cover the main same-repo fail-closed boundary. [R5 / packages/client/src/runtime/git-mirror-manager.ts:686]

**已执行**
- 已 request_changes

yuezengwu:
Thanks @codex-assistant — R4 is a real bug, fixed in 5e7ffad8.

The previous guard only checked that `ref` *resolved* locally, not that the checked-out HEAD was *at* it. You're right: a clone on `main` whose config repoints `ref` to a locally-present commit that isn't HEAD, on a transient fetch failure, would have returned `stale-offline` while `prepareSourceRepos` surfaces the configured `ref` — advertising the repo as being at `ref` while serving a different commit, and breaking the honor-pinned-commit-as-is contract.

Fix takes your option (b) — fail closed — to keep the degrade strictly read-only (no working-tree mutation on the degraded path):

```ts
const refSatisfiedByHead =
  !ref || (headSha !== null && (await localRefCommit(absTarget, ref)) === headSha);
if (headSha !== null && refSatisfiedByHead) {
  return finish("stale-offline");
}
```

So `stale-offline` now requires the configured ref to already resolve to the current HEAD; any just-changed ref the checkout is not at fails closed.

Tests added:
- configured `ref` already == HEAD + transient failure → `stale-offline`
- configured `ref` (a local commit) != HEAD + transient failure → fail closed (the R4 scenario)

`@first-tree/client`: 90 files / 1012 tests passed; typecheck + biome clean.

baixiaohang (approved):
## Review by codex-assistant - PR #867

**建议**: approve
- 理由: 上轮 R4 blocker 已由 `5e7ffad8` 修复，`stale-offline` 现在只在 configured `ref` 已由当前 HEAD 满足时降级，其他边界继续 fail-closed。

**分级**: A
- 语义升级: 无

**PR 摘要**
- 解决问题: transient source-repo fetch failure 不再在已有同仓库 checkout 可用时中断 agent session start/resume。
- 设计方案: 新增 `stale-offline` outcome，并用 same-origin、transient-network、standalone clone、HEAD/ref 一致性等条件限制离线降级。
- 受影响模块: `packages/client/src/runtime/git-mirror-manager.ts`, `packages/client/src/runtime/source-repos.ts`, source-repo lifecycle tests.

**Review 意见**
✅ 1. 上轮 `ref` 只可解析但 HEAD 不匹配的 blocker 已解决：当前实现要求 `localRefCommit(ref) === headSha` 才允许 `stale-offline`。
✅ 2. same-repo transient、origin repoint transient、non-transient failure、configured ref mismatch 都有测试覆盖。
✅ 3. 最新 GitHub CI 全部完成且成功（`Validate Skill` skipped）。

**已执行**
- 已 approve（baixiaohang 授权）

PR #871, opened 2026-06-08
feat(web): agent-detail PR2 — save semantics, Environment zones, leave guard, one Profile edit

Gandy2025:
## What & why

PR2 of 3 in the agent-detail optimization — the **save & edit behavior (P0)** pass. **Stacked on PR1** (#868): base is `feat/agent-detail-pr1-structure`, so review only the incremental diff here. Merge PR1 first; this rebases to `main` after.

Goal: make the page's two save semantics obvious, zone the Environment tab by them, guard against losing unsaved work, and collapse Profile to one Edit.

## Changes

- **Two save semantics, explicit.** One canonical tag pair next to each section title — **"Applies immediately"** (identity, appearance, repos, skills, MCP, instructions) vs **"Saved from the Save bar"** (model, reasoning effort, env). Immediate sections flash a transient "Saved" (shared `useJustSaved`, tick-based so repeated saves re-arm the 2.5s window).
- **Environment zoned by semantics.** Immediate zone (Execution + Repositories) then draft zone (Model settings + Environment variables → SaveBar). Model settings moved out of `RuntimeSection` into the draft zone.
- **Leave guard.** `useBlocker` is unavailable here — the app uses a declarative `<BrowserRouter>`, not a data router, so `useBlocker` throws via `useDataRouterContext`'s invariant. Instead a `guardedNavigate` collars every in-page "leave this agent" exit (breadcrumb, back, Chat, "Manage in Settings", "Open Computers") and confirms via the existing ConfirmDialog when the config draft is dirty; it's exposed on the outlet context for PR3's switcher. `beforeunload` covers hard exits. **Known gap (follow-up; clean fix is a data-router migration):** the global top nav (layout.tsx) and browser back/forward (popstate) route around the guard.
- **One Profile edit.** Identity + Appearance merged into a single "Edit profile" dialog (`profile-edit-dialog.tsx`). Identity + fallback color commit together in one PATCH; the avatar image stays eager (separate endpoint) with its own error. **Partial failure keeps the dialog open and shows which half failed.** Form state is snapshotted once per (open, agent.uuid) so the eager image refresh can't wipe unsaved identity/color edits.

## Review (2 independent rounds, codex)

- **R1 (review):** found a real interaction bug — the merged dialog's eager avatar upload calls `onRefresh`, swapping the `agent` prop and resetting unsaved identity/color fields. Fixed with a uuid-keyed ref-guard (init only on fresh open / different agent).
[...]

PR #873, opened 2026-06-08
feat(web): agent-detail PR3 — agent switcher + ContextBar identity + chrome dedup

Gandy2025:
## What & why

PR3 of 3 in the agent-detail optimization — the **agent switcher + top-bar chrome** pass. **Stacked on PR2** (#871): base is `feat/agent-detail-pr2-save-behavior`, so review only the incremental diff here. Lands after PR1 → PR2.

Goal: replace the breadcrumb with a fast agent switcher, and stop the ContextBar from duplicating the Execution section.

## Changes

- **Agent switcher (vertical-B)** — new `agent-switcher-strip.tsx` replaces the breadcrumb: a horizontal strip of avatar-over-name items, leftmost `‹ Team`, the current agent selected (brand ring + presence dot). Switching agents / going to Team routes through the page's `guardedNavigate` (PR2), so an unsaved config draft prompts a confirm; the selected agent is a no-op (not a "leave"). Overflow scrolls with edge fades; the active item is kept in horizontal view; names truncate with a title tooltip.
- **Scope follows All/Mine, single source of truth** — extracted the `AgentFilter` type / storage key / read+write + the membership predicate into `pages/team/agent-filter.ts`; the Team page now consumes it (zero behavior change). Read-on-mount (the toggle only lives on the Team page, where this strip is unmounted). The current agent is always included even when scope would exclude it, and is rendered from the live (10s-polled) agent prop so the selected chip never lags the header.
- **Current tab preserved on switch** — falls back to profile when the target agent doesn't support it. New `tabs.ts` is now the single source of tab existence: `buildTabs` derives from `tabKeysFor`, and the switcher uses the same to resolve the target path.
- **Page remounts on agent switch** — `AgentDetailPage` keys an inner `AgentDetailPageView` by `:uuid`. The switcher navigates in place (only the param changes), so without this the previous agent's config draft / pending-nav / dialog state would leak into the next agent (`useConfigDraft` only re-seeds from null).
- **ContextBar → identity** — was "Runs on \<runtime\> @ \<computer\>" (duplicated Execution); now avatar + name + presence. It only appears after the switcher scrolls away (sentinel IntersectionObserver), so the two never overlap.
- **Header chrome converged** — presence (already labelled) + active-session count folded into one status cluster.

## Review (2 independent rounds, codex)

- **R1:** caught a real **[P1]** — in-place agent switch leaked the previous agent's config draft (fixed via the uuid-keyed remount above) — and a **[P2]** — clicking the already-selected agent tripped the leave guard (now a no-op).
- **R2 (challenge):** no P1; 4 fixed — selected chip used the unpolled query row (now the live agent prop + 10s poll on the switcher query); `scrollIntoView` could scroll the page vertically (now horizontal-only `scrollLeft`); ContextBar `truncate` needed `min-w-0 flex-1`; `resolveTabPath` received the tab key not the path (now the real path).

## Testing

- Switcher render/scope + leave-guard-on-switch tests; ContextBar identity test; DEV `/preview/agent-detail` gains a switcher panel (verified via headless browser).
- Gates (full repo): `biome check` (1194) · typecheck (11 pkgs) · `lint:tokens` · web tests **694**.

### DIFF
--- a/packages/web/src/pages/agent-detail-preview.tsx
+++ b/packages/web/src/pages/agent-detail-preview.tsx
@@ -2,6 +2,7 @@ import type { Agent, AgentResourcesOutput, AgentRuntimeConfig } from "@first-tre
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import type { ReactNode } from "react";
 import { Outlet, Route, Routes } from "react-router";
+import { AgentSwitcherStrip } from "./agent-detail/agent-switcher-strip.js";
 import { ResourceTypeSection } from "./agent-detail/capability-section.js";
 import type { AgentDetailContext } from "./agent-detail/layout-context.js";
 import { PromptTab } from "./agent-detail/prompt-tab.js";
@@ -300,6 +301,15 @@ function buildClient(): QueryClient {
     },
   });
   client.setQueryData(["agent-resources", UUID], RESOURCES);
+  // Seed the agent switcher list (both admin/member keys, since preview auth is
+  // ambient). fetchAllAgents flattens to Agent[].
+  const switcherAgents: Agent[] = [
+    AGENT,
+    { ...AGENT, uuid: "agent-preview-2", name: "nova", displayName: "Nova", runtimeState: "working" },
+    { ...AGENT, uuid: "agent-preview-3", name: "atlas", displayName: "Atlas Researcher", runtimeState: null },
+  ];
+  client.setQueryData(["agents", "team-page", "admin"], switcherAgents);
+  client.setQueryData(["agents", "team-page", "member"], switcherAgents);
   return client;
 }
 
@@ -322,6 +332,17 @@ export function AgentDetailPreviewPage() {
     <QueryClientProvider client={client}>
       <div style={{ minHeight: "100vh", background: "var(--bg)" }}>
         <div className="mx-auto" style={{ maxWidth: 720, padding: "var(--sp-6) var(--sp-4)" }}>
+          <h1 className="text-title m-0" style={{ color: "var(--fg)" }}>
+            Agent switcher (vertical B)
+          </h1>
+          <p className="text-body" style={{ color: "var(--fg-3)", marginTop: "var(--sp-1)" }}>
+            Replaces the breadcrumb: ‹ Team + avatar-over-name items, current agent selected (brand ring), presence
+            dots. Switching is leave-guarded.
+          </p>
+          <div style={{ marginTop: "var(--sp-4)", marginBottom: "var(--sp-8)" }}>
+            <AgentSwitcherStrip currentAgent={AGENT} currentTabPath="profile" onNavigate={() => {}} />
+          </div>
+
           <h1 className="text-title m-0" style={{ color: "var(--fg)" }}>
             Environment tab — Repositories
           </h1>
--- a/packages/web/src/pages/agent-detail.tsx
+++ b/packages/web/src/pages/agent-detail.tsx
@@ -1,6 +1,6 @@
 import type { RuntimeProvider } from "@first-tree/shared";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
-import { ArrowLeft, MessageSquare, Monitor } from "lucide-react";
+import { MessageSquare, Monitor } from "lucide-react";
 import { type ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Outlet, useLocation, useNavigate, useParams } from "react-router";
 import { type HubClient, listClients } from "./../api/activity.js";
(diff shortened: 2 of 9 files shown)


Gandy2025:
Review note (gandy-assistant — final maintainer pass). Verified the three risk areas on the branch HEAD:

1. **uuid-remount:** `AgentDetailPage` is now a thin shell returning `<AgentDetailPageView key={params.uuid} />`; all per-agent state (draft, pendingNav, dialogs, justSaved) lives in the view, so an in-place agent switch fully remounts — no config-draft bleed between agents. "Keep current tab" is preserved via `resolveTabPath` (falls back to profile when the target doesn't support the current tab).
2. **Exit funnel (R4-class audit — grepped the whole agent-detail subtree, not just the diff):** every leave-the-agent exit is guarded. The switcher routes all items through `onNavigate` (= `guardedNavigate`), the selected item is a no-op, and there's no `useNavigate`/plain `navigate` leak in the strip. The remaining plain `navigate(` calls are all accounted for — inside `guardedNavigate`, same-agent tab/anchor, delete-success, or the load-error page. No unguarded leave-exit.
3. **Single source of truth:** `tabKeysFor` drives both `buildTabs` (adds labels) and `resolveTabPath` — no drift. `agent-filter.ts` keeps the same storage key + read/write semantics, so Team-page behavior is preserved.

No findings. LGTM (can't formally Approve — same GH account as the author).

yuezengwu (approved):
Independent review (yzw-assistant) — incremental diff over PR2 base only. **Approve.**

Verified the goal: breadcrumb → agent switcher strip, uuid-keyed page remount, and ContextBar repointed from "Runs on \<runtime\> @ \<computer\>" → agent identity. Frontend-only, no schema / core data-structure changes.

Correctness — all clear:
- **uuid-remount** (`<AgentDetailPageView key={params.uuid} />`) genuinely closes the R1 config-draft bleed; all per-agent state lives in the inner view.
- **Switcher query key + `fetchAllAgents`** match the Team page exactly → the navigate-from-Team cache-hit claim holds.
- **`canEditConfigFor`** is a faithful mirror of the shell's `canEditConfig`, so `resolveTabPath` never routes to a tab the target page rejects; `tabKeysFor` is the single source for both `buildTabs` and `resolveTabPath` (no drift).
- **`agent-filter.ts`** extraction is behavior-preserving: `!matchesAgentScope(...)` reduces to the old `filter==="mine" && managerId!==self`.
- **No visibility leak:** members fetch the server-visibility-filtered `/agents`; only admins hit `/agents/all`, and admins are authorized to see all org agents.
- `boundClientLabel` still consumed via layout context; `Breadcrumb` still used in the load-error fallback page — no dead imports.
- Selected-item is a no-op (R1-P2 fix); every leave-exit funnels through `guardedNavigate`.

Minor, non-blocking (no action required to land):
1. Admin "All" scope inlines other members' private agents straight into the strip — consistent with admin authority, but the Team page separates them into a governance section, so the strip's mental model differs slightly.
2. A full paginated agent-list fetch + 10s poll now runs on every agent-detail page (mirrors Team cadence, but a new cost on large orgs).

CI was pending at review time — expecting the listed gates (biome / typecheck / tokens / web tests) green before merge. Lands after PR1 → PR2 per the stack.

baixiaohang (approved):
## Review by codex-assistant - PR #873

**建议**: approve
- 理由: Incremental web-only pass keeps tab/scope/navigation state derivations consistent, and I found no blocking regression in the PR3 diff.

**分级**: C
- Path 初判: only `packages/web/**` -> C
- 语义升级: 无；未触及 web 订阅、WS frame parsing 或共享数据层

**PR 摘要**
- 解决问题: Replace the agent-detail breadcrumb with a fast agent switcher, and make the sticky ContextBar show identity instead of duplicating Execution.
- 设计方案: Shared Team agent filter, shared tab availability resolver, uuid-keyed remount for in-place agent switching, and ContextBar identity chrome.
- 受影响模块: `packages/web/src/pages/agent-detail*`, `packages/web/src/pages/team`

✅ 1. `AgentDetailPageView` is keyed by `:uuid`, so per-agent draft / pending-nav / dialog state does not leak across in-place switches. [R4 / packages/web/src/pages/agent-detail.tsx:55]
✅ 2. Tab existence is centralized in `tabKeysFor`, and switcher target routing uses the same source through `resolveTabPath`; unsupported target tabs fall back to `profile`. [R1 / packages/web/src/pages/agent-detail/tabs.ts:22]
✅ 3. The switcher follows the Team read model: same Team query key/fetcher, members stay on the server-filtered list, admins alone use the all-agents endpoint, and the current agent is force-included without widening member visibility. [R1 / packages/web/src/pages/agent-detail/agent-switcher-strip.tsx:40]
✅ 4. ContextBar identity rendering removes the old runtime/computer duplication while preserving a sticky “which agent” signal after the header sentinel scrolls away. [R2 / packages/web/src/pages/agent-detail/context-bar.tsx:28]

**已执行**
- 已 approve（静态 review；未运行测试或项目 gates，采信 PR 中列出的 gates）

PR #894, opened 2026-06-09
Fix WS auth handshake classification

baixiaohang:
## Summary

- add finite shared schemas for `auth:rejected`, `auth:expired`, and `auth:retryable` WS auth control frames
- classify server WS auth handshake failures by phase, emitting retryable frames/1011/1013 for transient or post-auth failures instead of `auth:rejected`
- treat auth-frame timeout as retryable (`auth:retryable` + `auth_timeout` + 1013) so slow/cold clients reconnect instead of entering auth-paused mode
- enrich the `ws.connection` span with `auth.ws.phase/outcome/code/retryable/close_code/error_class` and safe untrusted JWT trace attrs, including `client_register` attrs before 4403 register rejections
- update the client auth state machine to use frame type/code only, keep retryable failures reconnecting without auth-paused, and preserve `retryAfterMs` as a reconnect floor
- surface structured auth pause details in CLI output and add shared/server/client/CLI regression coverage

## Tests

- `pnpm --filter @first-tree/shared exec vitest run src/__tests__/ws-auth-schema.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/ws-auth-classification.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/auth-paused.test.ts src/__tests__/client-connection-auth-expired.test.ts src/__tests__/client-connection-auth-classification.test.ts --maxWorkers=2`
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/client-runtime-context-tree.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/ws-auth-classification.test.ts src/__tests__/ws-jwt-refresh.test.ts src/__tests__/ws-server-welcome.test.ts --maxWorkers=2`
- `pnpm check`
- `pnpm typecheck`

`pnpm check` still reports the existing `packages/client/src/runtime/workspace-migrations.ts` `void | "deferred"` warning.

## Review follow-up

Addressed yuezengwu's non-blocking suggestions for this PR:

- `auth_frame_timeout` now emits `auth:retryable` with code `auth_timeout` and closes 1013.
- 4403 `client:register:rejected` paths now write `auth.ws.phase = "client_register"` and related span attrs before closing.

`server_draining` remains a forward-compatible retryable code and is not emitted by this change.

### DIFF
--- a/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
@@ -510,6 +510,12 @@ describe("ClientRuntime context-tree wiring", () => {
     if (!paused) throw new Error("auth:paused listener missing");
     paused("auth_refresh_failed", new Error("refresh rejected"));
     expect(print.status).toHaveBeenCalledWith("✗", "auth rejected — pausing agents until fresh credentials arrive.");
+    expect(print.status).toHaveBeenCalledWith("", "refresh rejected");
+    const structured = new Error("Server rejected access token");
+    Object.defineProperty(structured, "authCode", { value: "invalid_token" });
+    Object.defineProperty(structured, "authMessage", { value: "signature mismatch" });
+    paused("auth_rejected", structured);
+    expect(print.status).toHaveBeenCalledWith("", "Auth rejection code: invalid_token — signature mismatch");
     expect(rt.isPaused()).toBe(true);
 
     const resumed = connectionListeners.get("auth:resumed");
--- a/apps/cli/src/core/client-runtime.ts
+++ b/apps/cli/src/core/client-runtime.ts
@@ -38,6 +38,13 @@ export function isAgentSuspendedBindError(error: unknown): boolean {
   return msg.includes("agent_suspended");
 }
 
+function authPausedDetail(error: Error): string {
+  const authCode = "authCode" in error && typeof error.authCode === "string" ? error.authCode : null;
+  const authMessage = "authMessage" in error && typeof error.authMessage === "string" ? error.authMessage : null;
+  if (!authCode) return error.message;
+  return `Auth rejection code: ${authCode}${authMessage ? ` — ${authMessage}` : ""}`;
+}
+
 export type ClientRuntimeOptions = {
   /**
    * Version of the Command package this process was launched from. Passed to
@@ -142,7 +149,7 @@ export class ClientRuntime {
     this.connection.on("auth:paused", (reason, err) => {
       print.blank();
       print.status("✗", "auth rejected — pausing agents until fresh credentials arrive.");
-      print.status("", err.message);
+      print.status("", authPausedDetail(err));
       print.status("", "Recovery: get a new connect token from the First Tree web console");
       print.status("", "          (Computers → + New Connection), then re-run `first-tree login <token>`.");
       print.status("", `Paused reason: ${reason}. Process is staying alive — no restart needed after login.`);
--- a/packages/client/src/__tests__/auth-paused.test.ts
+++ b/packages/client/src/__tests__/auth-paused.test.ts
@@ -113,6 +113,35 @@ describe("ClientConnection — auth paused mode (Bug 2)", () => {
     await connection.disconnect();
   }, 10_000);
 
+  it("structured auth:rejected includes the rejection code in paused error", async () => {
+    wss.on("connection", (ws: WebSocket) => {
+      ws.on("message", (raw) => {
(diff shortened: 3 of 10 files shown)


yuezengwu (commented):
## Review — WS auth handshake classification

Read the full diff + verified the server-side observability helpers (`classifyJoseError` / `decodeJwtForTrace` / `untrustedAttrs`, `JwtFailureReason`'s 4 variants) and the client reconnect-floor integration. The change is coherent end to end and the test matrix (shared schema, server phase classification + span attrs, client retryable-reconnect vs. paused) covers the important paths. No blockers.

**What it does:** splits the previously-single `auth:rejected` outcome into three finite, schema-backed control frames — `auth:rejected` (deterministic credential/identity failure → 4401 + paused), `auth:expired` (4401, reconnect with fresh token), `auth:retryable` (transient/post-auth failure → 1011/1013, reconnect without pausing). The client now classifies by frame type/code only and preserves `retryAfterMs` as a reconnect floor; the server enriches the `ws.connection` span with `auth.ws.*` phase/outcome attrs.

### Things I checked and they hold up
- `consumeReconnectDelay` is consumed in **both** the initial-connect retry loop and `scheduleReconnect`, so an `auth:retryable` `retryAfterMs` floor is honored even when the failure happens during the very first `connect()` (the new client test asserts the ≥1150ms gap). Floor is set before the 1013 close, so ordering is correct.
- Jitter is now floor-only and additive (`+0..20%`, never subtracts) — correct for a Retry-After lower bound, and the unified path no longer applies ±jitter to organic backoff.
- Post-auth-welcome failure (after `auth:ok`) closes 1011 retryable and never emits `auth:rejected`; the connect promise rejects "closed before ready" and retries without pausing. Verified the half-authenticated socket just closes cleanly.
- Legacy `auth:rejected` frames (no `code`, old `reason` field) still fall through to paused mode on the new client — backward compatible. New-server `auth:retryable` to an old client degrades to an unknown-frame + 1013 close → reconnect (no pause), which is the desired self-heal.
- `rejectedCodeForJoseError` switch is exhaustive over the 4 `JwtFailureReason` variants (`jwt_expired` is handled earlier and returns before the switch).

### Non-blocking notes
- **Finding 1 (design question):** `auth_frame_timeout` and `invalid_auth_frame` are classified as non-retryable `auth:rejected` → client enters paused mode (manual `login` to recover). A server-side 5s timeout waiting for the client's first frame is arguably an environmental/transient condition (cold start, GC/event-loop stall) rather than a credential failure, so a slow-but-valid client gets paused instead of self-healing. Not a regression (pre-PR `reason:"timeout"` also paused old clients), and 5s is generous — just flagging whether timeout belongs on the retryable side.
- **Finding 2:** `server_draining` (retryable code) and `authControlFrameSchema` (union) are defined/exported but not yet emitted/consumed — forward-looking surface, fine to keep.
- **Finding 3:** Agree the `client_register` phase not writing `auth.ws.phase` on the existing 4403 path is observability-only and non-blocking.

LGTM pending the `Test` job going green.

yuezengwu (approved):
Approved — WS auth handshake classification splits the single `auth:rejected` outcome into three finite, schema-backed frames (rejected / expired / retryable) so transient and post-auth failures reconnect instead of pausing agents. Verified: retryAfterMs floor consumed in both the initial connect loop and scheduleReconnect; floor-only additive jitter; post-auth-welcome failure maps to 1011 retryable and never mis-reports rejected; old/new client↔server combinations degrade safely both ways. Backward-compatible, additive wire-protocol change, no DB change. Non-blocking notes left in the review comment.

yuezengwu (approved):
Re-approve after synchronize. The new push addresses both non-blocking points from my earlier review:

- Finding 1: `auth_frame_timeout` now emits `auth:retryable` (code `auth_timeout`) + 1013 instead of `auth:rejected` — slow/cold-start clients reconnect and self-heal rather than entering auth-paused mode. `AUTH_TIMEOUT` moved cleanly from AUTH_REJECTED_CODES to AUTH_RETRYABLE_CODES (no dangling references; the sole server call-site uses retryAuthAndClose). Backward-compatible: old clients receiving the unknown `auth:retryable` frame + 1013 reconnect (strictly better than the prior paused behavior).
- Finding 3: the 4403 `client:register:rejected` paths now write `auth.ws.phase = "client_register"` and related span attrs (no_membership / client_register_failed) before closing.

New shared/server regression coverage for both. Verified the auth_timeout enum move and the client_register attrs are additive and consistent. No new concerns.

PR #869, opened 2026-06-08
feat(client): workspace hygiene — state-based cleanup + one-shot legacy migrations

liuchao-001:
## Summary

Two CLI behaviors that together stop agent workspaces from sliently accumulating stale resources across config edits and version bumps:

1. **State-based delete-on-removal** for source repos and tree skills. The CLI records what it currently manages at `<workspace>/.agent/managed.json`; on the next session start it diffs the recorded set against the current config and removes anything that fell out — source-repo `localPath` directories AND skill payload + `.claude/skills/<name>` symlink pairs.
2. **One-shot directory-structure migrations** for legacy workspaces. Four v1 migrations sweep known historical baggage (`first-tree-hub/` retired source clones, UUID-named per-chat snapshot dirs from the per-chat-cwd era, the pre-`.agent/` `.first-tree/` state dir, the legacy `WHITEPAPER.md` symlink). Each migration runs at most once per workspace via a marker at `.agent/migrations-applied.json`.

Motivation comes from a real example — the `liuchao-staff` agent workspace had accumulated **~27 GB** of stale residue across the last few months of CLI evolution. Phase A manual cleanup brought it to ~544 MB; this PR codifies the equivalent behavior into the CLI so the same drift doesn't recur and so other long-running workspaces self-heal on upgrade.

## What this PR is NOT

- ❌ Not touching workspace size limits / quotas
- ❌ Not changing the source-repo directory layout (still at workspace top level — `yuezengwu` review explicitly rejected the `.first-tree-repos/` reorg as out of scope)
- ❌ Not touching tree-repo lifecycle (content-hash shared at staging root, separate concern)
- ❌ Not introducing a workspace-doctor CLI subcommand
- ❌ Not handling worktree cleanup (agent self-managed; goes via the AGENTS.md template, separate PR)

## How it stays safe

State-based cleanup applies path-level precision — only resources the CLI itself recorded as installed are eligible for deletion. Anything the user dropped at workspace root (third-party clones, manual experiments, custom skills) is never in the recorded set and therefore never touched.

Source-repo cleanup has three guards before deleting any clone:
1. Working tree clean (`git status --porcelain` empty)
2. No local commits ahead of upstream
3. No dependent `git worktree` checkouts

Any probe failure conservatively SKIPS the delete and logs a warning — better to leave a stale clone than nuke unpushed work.

Migrations:
[...]

PR #898, opened 2026-06-09
fix(client): managed-state writes to .first-tree-workspace/, not legacy .agent/

liuchao-001:
## Summary

Two related fixes, both follow-up to PR #869:

### 1. `managed-state.ts` + `workspace-migrations.ts` write to the wrong dir (`.agent/` instead of `.first-tree-workspace/`)

PR #859 (`rename agent runtime dir to .first-tree-workspace`) landed between PR #869 being authored and merged. #859 retitled the runtime dir constant in `bootstrap.ts` (`FIRST_TREE_RUNTIME_DIR = FIRST_TREE_WORKSPACE_MARKER`) and updated every other state-file consumer in the codebase — but #869's two new modules weren't in that sweep, presumably because they didn't exist yet on main when #859 ran its global rename.

**Production symptom** (caught by @liuchao-001 on the `liuchao-staff` agent):

- `prepareSourceRepos` / `reconcileTreeSkillState` write to `.agent/managed.json`.
- `ensureWorkspaceRuntimeDir`'s `mergeLegacyRuntimeDir` runs later in the bootstrap path and merges `.agent/` → `.first-tree-workspace/` with **target-wins** conflict policy.
- First session after #869 shipped: target didn't exist → whole-dir rename → state landed in target with first-run defaults (`skills: []`).
- Every subsequent session: source dir re-created by my code, merge discards it. The target managed.json stays **frozen** at the first-run snapshot.
- Net effect: `reconcileTreeSkillState` never gets to populate `managed.json::skills`, the whole state-reconcile path is silently inert.

**Fix**: switch both modules to use `.first-tree-workspace/`. Not via `import { FIRST_TREE_RUNTIME_DIR } from "./bootstrap.js"` — that introduces a top-level circular-import init bug (`bootstrap` → `first-tree-skills/installer` → `managed-state` would evaluate `MANAGED_STATE_REL = join(FIRST_TREE_RUNTIME_DIR, "managed.json")` while `bootstrap` is still initialising; the constant reads `undefined` and `join` throws). Inline the literal `".first-tree-workspace"` with a cross-reference comment to `bootstrap.ts :: FIRST_TREE_RUNTIME_DIR` as source of truth — same pattern `bootstrap.ts` itself uses for `FIRST_TREE_WORKSPACE_MARKER`.

### 2. `v1-orphan-skills` migration cleans hardcoded legacy skill list

PR #869's state-based skill reconcile is path-precise: it only removes skills it previously recorded in `managed.json::skills`. Legacy skill payloads installed before state tracking existed — e.g. the 6 skills the CLI used to bundle (`attention`, `first-tree-cloud`, `first-tree-github-scan`, `first-tree-onboarding`, `first-tree-write`, `github-scan`) — sit on disk forever without an explicit sweep.

This commit adds `v1-orphan-skills`: hardcoded one-shot migration that removes both the `.agents/skills/<name>/` payload AND the matching `.claude/skills/<name>` symlink for each historical name. Defensive about user-authored content: the Claude symlink is only unlinked when it actually IS a symlink (matches `v1-whitepaper-symlink` shape).

## Test plan

- [x] `pnpm typecheck`: clean
- [x] `pnpm check` (biome): clean
- [x] `pnpm --filter @first-tree/client test src/__tests__/{managed-state,workspace-migrations,source-repos-state-reconcile,skills-state-reconcile}.test.ts`: **51 / 51 pass** (3 new orphan-skills tests added)
- [ ] Verification on a real workspace once shipped:
  - Next session: `managed.json::skills` populates to current `TREE_SKILL_NAMES` (no longer frozen at `[]`).
  - Next migration run: the 6 historical skills get cleared from `.agents/skills/` and their `.claude/skills/` symlinks.

### DIFF
--- a/docs/development/agent-workspace-state.md
+++ b/docs/development/agent-workspace-state.md
@@ -2,7 +2,7 @@
 
 This page documents the CLI-owned state files that live inside an agent's
 home directory and the contract they enforce. If you're touching anything
-under `<workspace>/.agent/`, anything that calls `prepareSourceRepos`,
+under `<workspace>/.first-tree-workspace/`, anything that calls `prepareSourceRepos`,
 `installFirstTreeSkills`, or `ensureAgentBootstrap`, or you're proposing a
 new directory-structure migration — start here.
 
@@ -25,13 +25,13 @@ sit on disk forever. The state files below are that record.
 
 | Path (relative to workspace root) | Owner | What it is |
 | --- | --- | --- |
-| `.agent/managed.json` | `runtime/managed-state.ts` | Schema-versioned record of the CLI-managed resources currently materialised in this workspace. Two arrays: `sourceRepos` (localPath names) and `skills` (skill names). Diffed on every session start to discover removals. |
-| `.agent/migrations-applied.json` | `runtime/workspace-migrations.ts` | Set of one-shot directory-structure migration ids that have already run in this workspace. Each migration runs at most once even if it's later removed from the registry; the marker stays as forward protection. |
+| `.first-tree-workspace/managed.json` | `runtime/managed-state.ts` | Schema-versioned record of the CLI-managed resources currently materialised in this workspace. Two arrays: `sourceRepos` (localPath names) and `skills` (skill names). Diffed on every session start to discover removals. |
+| `.first-tree-workspace/migrations-applied.json` | `runtime/workspace-migrations.ts` | Set of one-shot directory-structure migration ids that have already run in this workspace. Each migration runs at most once even if it's later removed from the registry; the marker stays as forward protection. |
 
 Both files use atomic writes (temp + rename) so a crashed writer never
 leaves a half-formed JSON record on disk.
 
-### `.agent/managed.json` schema
+### `.first-tree-workspace/managed.json` schema
 
 ```json
 {
@@ -55,7 +55,7 @@ When the file is missing or malformed, the reconcile path treats it as
 "first run on this workspace" and performs no deletions — the safe
 default.
 
-### `.agent/migrations-applied.json` schema
+### `.first-tree-workspace/migrations-applied.json` schema
 
 ```json
 { "schemaVersion": 1, "applied": ["v1-uuid-snapshots", "v1-whitepaper-symlink"] }
@@ -183,4 +183,4 @@ sweep:
 - `runtime/agent-bootstrap.ts` — wires migrations into session start
 - `packages/shared/src/schemas/workspace-manifest.ts` — the W1 binding
   manifest at `<workspace>/.first-tree/workspace.json` (distinct from
-  `.agent/`; never touched by this machinery)
+  `.first-tree-workspace/`; never touched by this machinery)
--- a/packages/client/src/__tests__/managed-state.test.ts
+++ b/packages/client/src/__tests__/managed-state.test.ts
@@ -15,7 +15,7 @@ describe("managed-state", () => {
 
   beforeEach(() => {
     workspace = mkdtempSync(join(tmpdir(), "managed-state-test-"));
-    mkdirSync(join(workspace, ".agent"), { recursive: true });
+    mkdirSync(join(workspace, ".first-tree-workspace"), { recursive: true });
   });
 
   afterEach(() => {
@@ -120,7 +120,7 @@ describe("managed-state", () => {
     };
     writeManagedState(workspace, state);
     // The directory should contain only the final file, no `.tmp` siblings.
-    const agentDir = join(workspace, ".agent");
+    const agentDir = join(workspace, ".first-tree-workspace");
     const entries = readdirSync(agentDir);
     expect(entries.filter((entry) => entry.includes(".tmp"))).toEqual([]);
     expect(entries).toContain("managed.json");
--- a/packages/client/src/__tests__/skills-state-reconcile.test.ts
+++ b/packages/client/src/__tests__/skills-state-reconcile.test.ts
@@ -52,7 +52,7 @@ describe("installFirstTreeSkills — state-based skill reconcile (PR #869 P1-3)"
   beforeEach(() => {
     tmpBase = mkdtempSync(join(tmpdir(), "skill-reconcile-"));
     workspace = join(tmpBase, "ws");
-    mkdirSync(join(workspace, ".agent"), { recursive: true });
+    mkdirSync(join(workspace, ".first-tree-workspace"), { recursive: true });
     bundledSkillsRoot = makeBundledSkillsRoot(tmpBase);
   });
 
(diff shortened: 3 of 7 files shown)


yuezengwu (approved):
## Approve — verified the path-constant fix and its two load-bearing design claims

**Goal / core change:** PR #869's `managed-state.ts` + `workspace-migrations.ts` were left writing to the legacy `.agent/` dir after #859's global rename to `.first-tree-workspace/`. Because `ensureWorkspaceRuntimeDir`'s `mergeLegacyRuntimeDir` is **target-wins**, managed-state's re-created `.agent/` got discarded every session and the populated state never reached `.first-tree-workspace/managed.json` — silently defeating #869. This switches both modules to `.first-tree-workspace/`.

I verified both non-obvious decisions rather than taking them on faith:

1. **Inline literal over `import { FIRST_TREE_RUNTIME_DIR }` — the cycle is real.** `bootstrap.ts` → `first-tree-skills/installer.ts` → `managed-state.ts`, and installer also imports back from `bootstrap.js`. A top-level `join(FIRST_TREE_RUNTIME_DIR, "managed.json")` would evaluate while `bootstrap.ts` is still paused at its import, reading the constant as `undefined` and throwing. Inlining the literal with a cross-ref comment is correct and matches the existing `FIRST_TREE_WORKSPACE_MARKER` literal pattern. Confirmed.

2. **Merge ordering is now safe.** `writeManagedState` does its own `mkdirSync(RUNTIME_DIR)`, so it no longer matters whether it runs before `ensureWorkspaceRuntimeDir`; target-wins merge preserves the fresh writes instead of clobbering them. Already-affected prod workspaces self-heal on the next session (reconcile reads frozen `skills:[]`, rewrites populated).

3. **Conversion is complete** — all `.agent` refs in both modules (code + docstrings) converted, tests mirror the new path, and the legitimate legacy `.agent/` references (`LEGACY_AGENT_RUNTIME_DIR`, `mergeLegacyRuntimeDir`, contract doc's "never touched" note) are correctly left intact. No managed.json schema change.

**Non-blocking nit (follow-up, your call):** the runtime-dir name now lives as a literal in three places (`bootstrap.ts::FIRST_TREE_WORKSPACE_MARKER` + the two new `RUNTIME_DIR` consts). The "keep in sync" comments are the only guard against drift. A one-line invariant test asserting the two module literals equal the bootstrap export would make the cross-reference enforceable rather than aspirational. Not worth blocking — the marker is unlikely to change again.

The "Known limitation" (6 legacy skills not retroactively cleaned) is correctly scoped out — path-precision protection means reconcile only deletes skills once in `prev.skills`. Agreed it belongs in a separate `v1-orphan-skills` migration, not here.

Note: CI (Lint/Type/Test) was still pending at review time; approving on the strength of the local 48/48 run + the verification above per our no-wait-on-CI convention. If the Test job comes back red, re-check before merge.

yuezengwu (approved):
## Re-approve — reviewed the new `v1-orphan-skills` migration (commit ea1b0921b)

The second commit folds in the destructive cleanup that the first revision had scoped out as a follow-up. Since it `rmSync`s skill payloads off disk, I re-reviewed it as a destructive migration — entry guards over cleanup correctness — rather than letting my prior approval ride.

**Premise verified (the 6 names really are retired).** `CORE_SKILL_NAMES = []` and `TREE_SKILL_NAMES = ["first-tree","first-tree-context","first-tree-sync"]` are the only bundled sets, and none of the 6 historical names appear in any bundle/skill/registry reference in the TS source. So removing their on-disk residue is correct, not a live-skill deletion. (Worth noting they still show up as "available skills" in un-migrated workspaces — which is exactly the residue this sweep targets.)

**Entry guards are precise — no wrong-layout over-deletion:**
- The migration's hardcoded `.agents/skills/<name>` + `.claude/skills/<name>` paths **exactly match** the installer's `layoutFor` (`agentsRelPath` / `claudeRelPath` / `claudeSymlinkTarget`). It can only hit the installer's own layout.
- `.claude/skills/<name>` is unlinked **only when it's a symlink** (`lstatSync().isSymbolicLink()`), so a user-authored regular directory there survives — covered by the dedicated test.
- Hardcoded narrow list + one-shot by id; `first-tree-context` (a current skill sharing the namespace) is proven untouched by test 2.

**One intentional property worth stating plainly:** `.agents/skills/<name>` is removed unconditionally when present (no content fingerprint check) — i.e. the guard is the name list, not the payload contents. That's the agreed design (`.agents/skills/` is CLI-owned territory) and the test comment documents it. Acceptable given the path precision above, but it does mean: if a future bundle ever re-introduces one of these 6 names, a workspace that already applied `v1-orphan-skills` won't redelete it (marker is by id) — and the list comment correctly warns "never remove or reorder existing entries." Good.

**Tests (3 new, 51/51):** full-6 removal + symlink cleanup; current-skill survival; user-dir (non-symlink) survival. Coverage matches the risk surface.

**CI: fully green** — Lint & Type Check, Test (1m11s), CodeQL, Analyze ×3, Migrations all pass. This also resolves the "Test pending" caveat from my first review.

Not a core data-structure or DB change (one-shot filesystem cleanup), so no human-gate escalation. Approving.

baixiaohang (approved):
## Review by codex-assistant - PR #898

**建议**: approve
- 理由: 最新 head `ea1b0921` 的 path fix 与 `v1-orphan-skills` cleanup 都符合 workspace-root runtime artefact 约束；未发现阻塞风险。

**分级**: A
- 语义升级: 无

**PR 摘要**
- 解决问题: 修复 managed-state / migrations marker 仍写 legacy `.agent/` 导致 live state 冻结，并清理 state tracking 之前遗留的 6 个 retired bundled skill payload。
- 设计方案: state file 常量改为 `.first-tree-workspace/`；新增 `v1-orphan-skills` one-shot migration，删除 hardcoded historical skill names 下的 `.agents/skills/<name>/` payload，并只在 `.claude/skills/<name>` 是 symlink 时 unlink。
- 受影响模块: `packages/client/src/runtime`, `packages/client/src/__tests__`, `docs/development/agent-workspace-state.md`

**Review 意见**
✅ 1. 当前 shipped skill 集合核对为 `first-tree` / `first-tree-context` / `first-tree-sync`；6 个 historical names 不在当前 bundle 中。
✅ 2. cleanup 路径与 installer layout 一致，测试覆盖 full removal、current skill survival、Claude 侧 regular directory survival。
✅ 3. `.agents/skills/<legacy>` 无条件删除符合 CLI-owned runtime payload 归属；`.claude/skills/<legacy>` 只删 symlink，避免误删用户目录。
⚠️ 4. 非阻塞维护点：未来新增 retired skill cleanup 应使用新的 migration id；旧 `.agent/...` 注释也可后续顺手清理。

**已执行**

PR #900, opened 2026-06-09
fix: refresh auto-update target before install

baixiaohang:
## Summary

- Refresh the server-recommended update target immediately before automatic self-update installs.
- Coalesce valid `server:welcome` targets while an update decision is waiting on the quiet gate, with the latest server target taking precedence even when it is lower than the original pending target.
- Preserve the existing no-downgrade-below-current contract and keep prompt/off behavior unchanged.

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- `pnpm --filter @first-tree/client test -- update-manager` (95 files / 1070 tests passed)
- `pnpm --filter first-tree-dev test -- daemon-start-command login-command update-glue` (58 files / 427 tests passed)
- `git diff --check`

## Change Surface

## Notes

- package or install behavior changes: daemon auto-update now best-effort refreshes the current server target before installing, while still installing an exact server target and refusing targets older than the running version.
- docs or tests updated to match: added UpdateManager tests for quiet-gate target coalescing, refresh rollback above current, and refresh below-current skip.
- follow-up work: optional coverage for a refresh-await race where a later welcome arrives while refresh is pending and refresh then fails.

### DIFF
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -26,6 +26,7 @@ const coreMocks = vi.hoisted(() => ({
   promptMissingFields: vi.fn(),
   promptUpdate: vi.fn(),
   reconcileLocalRuntimeProviders: vi.fn(),
+  refreshServerUpdateTarget: vi.fn(),
   startClientService: vi.fn(),
   uploadAgentSkills: vi.fn(),
   uploadClientCapabilities: vi.fn(),
--- a/apps/cli/src/commands/daemon/start.ts
+++ b/apps/cli/src/commands/daemon/start.ts
@@ -36,6 +36,7 @@ import {
   promptMissingFields,
   promptUpdate,
   reconcileLocalRuntimeProviders,
+  refreshServerUpdateTarget,
   startClientService,
   uploadAgentSkills,
   uploadClientCapabilities,
@@ -232,6 +233,7 @@ export function registerDaemonStartCommand(daemon: Command): void {
             updateConfig: config.update,
             prompt: managed ? declineUpdate : promptUpdate,
             executeUpdate,
+            refreshServerTarget: refreshServerUpdateTarget,
           },
         });
         runtimeRef = runtime;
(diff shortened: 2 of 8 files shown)


yuezengwu (approved):
## Review: refresh auto-update target before install ✅

Traced the full decision flow in `update-manager.ts` plus the command-layer plumbing. The change is correct and well-tested; approving.

### What it does
Before an `auto`-policy install fires, `UpdateManager` now (a) coalesces the latest valid `server:welcome` target observed while a decision is in flight, and (b) best-effort re-reads the current server target via the new `refreshServerTarget` hook (`refreshServerUpdateTarget` → `fetchServerCommandVersion` → `/api/v1/bootstrap/config`). The freshest read wins, even when it is *lower* than the originally-pending target (rollback B→C installs C, not `semver-max`). The no-downgrade-below-current contract and `prompt`/`off` behavior are untouched.

### Verified
- **No-downgrade contract preserved** — `resolveAutoTargetBeforeUpdate` re-runs `eq`/`lt` checks against `current` on the *resolved* target, so a refreshed/coalesced target below the running version still skips (covered by the new below-current test).
- **Plumbing complete** — `RefreshUpdateTargetFn` exported from both `core/index.ts` and `packages/client/src/index.ts`; wired at both call sites (`daemon/start.ts`, `login.ts`); `runtime.ts` forwards it via `...this.updateHooks` spread. Optional hook, so back-compat holds when absent (falls back to coalesced welcome target).
- **In-flight slot still single-threaded** — coalescing path returns early without fanning out parallel decisions; `disposed` is re-checked after the refresh await.
- **CI green** — Lint & Type Check, Test, CodeQL, all CodeQL Analyze jobs pass.

### Non-blocking notes
- The refresh-await race you flagged (a newer welcome arrives while `refresh()` is pending, and refresh succeeds → refresh result wins, newer welcome ignored) is acceptable: both targets originate from the same server npm-poll and the next welcome after restart self-corrects. Fine as the documented follow-up.
- `normalizeAdvertisedTarget` now logs the "invalid version" warn for in-flight/coalesced welcomes too (previously only the non-in-flight path). Harmless, just slightly noisier.

No core data-structure or DB changes. LGTM.

PR #901, opened 2026-06-09
fix(client): clarify async agent handoffs

baixiaohang:
## Summary
- Clarify that agent-to-agent handoffs are async: continue only independent work after sending to another agent.
- Tell agents to end the turn and wait to be woken when the peer reply is the only remaining input.
- Keep the runtime briefing, first-tree skill, and communication reference aligned.

## Tests
- pnpm --filter @first-tree/client test -- agent-briefing
- pnpm check
- pnpm typecheck
- pnpm validate:skill
- git diff --check

Note: pnpm check exits 0 with an existing warning in packages/client/src/runtime/workspace-migrations.ts:89.

### DIFF
--- a/packages/client/src/__tests__/agent-briefing.test.ts
+++ b/packages/client/src/__tests__/agent-briefing.test.ts
@@ -347,6 +347,8 @@ describe("buildAgentBriefing — # Working in First Tree subsections", () => {
     // structured --request ask path; agents must be woken explicitly.
     expect(briefing).toMatch(/\*\*Asking a human\*\*/);
     expect(briefing).toMatch(/Reaching an \*\*agent\*\*/);
+    expect(briefing).toContain("After an agent handoff, continue only independent work");
+    expect(briefing).toContain("do not poll status");
     expect(briefing).toContain("**Fallback**");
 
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
LGTM ✅

Prompt/doc-only change (12 add, 0 del, no logic). Adds async-handoff guidance to the three aligned surfaces (runtime `agent-briefing.ts`, `first-tree/SKILL.md`, `references/agent-communication.md`) plus a matching test.

Verified:
- Wording is consistent across all three docs (briefing + SKILL identical; reference rephrases the same rule).
- Test assertions (`After an agent handoff, continue only independent work`, `do not poll status`) exactly match the strings added to `agent-briefing.ts` — pending Test job will pass.
- Guidance is accurate against the runtime model: the hosting daemon wakes the agent on inbox delivery and the final-text contract already states final text doesn't wake peers, so 'end the turn and wait to be woken' is correct. 'Don't escalate on delayed replies alone' is sound anti-loop guidance.

No blockers, no core/DB changes. Lint & Type Check, Validate Skill, Migrations all green; only Test + CodeQL analyze pending (no need to gate a doc change on them).

PR #913, opened 2026-06-09
feat(client): retain skipped source-repo cleanups in managed state for retry

liuchao-001:
## Summary

State-based source-repo cleanup in `reconcileSourceRepoState` previously dropped a prev-but-no-longer-current entry from `managed.json::sourceRepos` on the **first** session pass — even when the safety guards (dirty / ahead-of-upstream / has-worktrees / in-use-by-live-chat / probe-failed / remove-failed) refused the delete. Once dropped, no future session ever retried the delete, even after the operator cleared the blocking condition.

The user-reported symptom is exactly that: `first-tree-context` was removed from a workspace's per-agent resources, but the directory had attached `git worktree` checkouts, so the delete was skipped — and the directory then sat as a permanent orphan because the entry was already gone from managed state.

This PR makes the cleanup **self-healing**: skipped entries stay tracked, next session retries.

- **What changes**: in `reconcileSourceRepoState`, the new managed-state `sourceRepos` set is `current ∪ (prev items the guards refused to delete)` instead of just `current`. The next session's reconcile re-runs `tryRemoveCloneSafely` against retained entries; once the blocker is cleared, the delete completes and the entry drops.
- **What stays**: final outcomes (`removed` / `absent` / `not-a-clone`) still drop the entry — nothing to retry against. Captured in the new `isFinalRemoveOutcome` helper next to `RemoveCloneOutcome` in `runtime/source-repo-cleanup.ts`.
- **What's NOT touched**: `workspace-migrations.ts` keeps its "best effort, one shot" contract (gated by `migrations-applied.json`); the skill-reconcile path (`installFirstTreeSkills`) also unchanged — it has no operator-clearable guards, only transient rm failures. If we want skills to retry too, that's a follow-up.

Accepted trade-off: a persistent blocker (e.g. uncommitted work the operator never cleans) produces recurring "skipped — working tree is dirty" log lines every session. The previous design swallowed the log noise at the cost of zero auto-recovery; we're flipping that.

## Test plan

- [x] `pnpm exec vitest run packages/client/src/__tests__/source-repos-state-reconcile.test.ts` — 7/7 pass
  - 2 existing tests updated: dirty-skip and live-use-skip now assert state retention instead of "rolls forward"
  - 1 new test: full retry cycle — session 1 skips dirty repo, operator cleans it, session 2 completes the delete
- [x] `pnpm typecheck` — 11/11 packages clean
- [x] `pnpm check` (biome) — no new findings (the one pre-existing `Migration` `void`→`undefined` info is on unrelated code)
- [x] Doc updated: `docs/development/agent-workspace-state.md` reconcile-flow + safety-guards sections now describe retry semantics + the `isFinalRemoveOutcome` boundary

## Related context

Originated from a workspace state where `first-tree-context` couldn't be auto-removed because it had attached worktrees. After this PR, the same situation self-heals: operator removes the worktrees → next session's reconcile completes the delete.

### DIFF
--- a/docs/development/agent-workspace-state.md
+++ b/docs/development/agent-workspace-state.md
@@ -104,7 +104,9 @@ already operate per session:
 - **`prepareSourceRepos`** (`runtime/source-repos.ts`) — after the per-repo
   clone/fetch loop, diff `prev.sourceRepos` against the current set and
   remove any clone no longer in the config. Safety guards apply (see
-  below); state is then rewritten to the current set.
+  below); the new state is `current ∪ (prev items the guards refused to
+  delete)`, so a guard-skipped clone stays tracked and the next session
+  retries.
 - **`installFirstTreeSkills`** (`runtime/first-tree-skills/installer.ts`)
   — same diff against `TREE_SKILL_NAMES`. Dropped skills lose their
   `.agents/skills/<name>/` payload AND their `.claude/skills/<name>`
@@ -127,10 +129,24 @@ Every clone deletion — both state-based and migration-driven — goes through
 6. **Any probe crashed / timed out** → `probe-failed`, skip. Conservative
    default: we'd rather leave a stale clone than nuke unpushed work.
 
-A skipped clone stays on disk and becomes an operator follow-up. The
-state file is still updated to the current set so the next session
-doesn't try the same skipped clone again as part of state-based cleanup
-— the migration path is "best effort, one shot".
+A skipped clone stays on disk **and stays in
+`.first-tree-workspace/managed.json::sourceRepos`** so the next session's
+state-based reconcile re-runs the probes. The conditions that block a
+delete (uncommitted work, commits ahead of upstream, an attached worktree,
+a live chat holding the path, a transient probe / rm failure) are all
+operator-clearable between sessions; once cleared, the follow-up reconcile
+completes the delete. Recurring "skipped — working tree is dirty" log
+lines on persistent blockers are accepted as the cost of self-healing.
+
+Only **final** outcomes — `removed` (we deleted it just now), `absent`
+(already gone), or `not-a-clone` (no `.git/` — never ours to delete) —
+drop the entry from managed state and stop further retries. The
+classification lives in `isFinalRemoveOutcome` next to the
+`RemoveCloneOutcome` definition in `runtime/source-repo-cleanup.ts`.
+
+The migration path (`workspace-migrations.ts`) is unchanged — that one is
+still "best effort, one shot" because it's gated by `migrations-applied.json`
+and re-runs only when the marker is missing.
 
 ## What is NOT touched by this machinery
 
--- a/packages/client/src/__tests__/source-repos-state-reconcile.test.ts
+++ b/packages/client/src/__tests__/source-repos-state-reconcile.test.ts
@@ -139,7 +139,7 @@ describe("prepareSourceRepos — state-based reconcile (PR #869 P1-3)", () => {
     expect(readManagedState(workspace)?.sourceRepos).toEqual(["repo-a"]);
   });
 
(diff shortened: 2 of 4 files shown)


yuezengwu (approved):
Reviewed — approving. ✅

**What this does:** makes state-based source-repo cleanup self-healing. Previously `reconcileSourceRepoState` rewrote `managed.json::sourceRepos` to `current` on every pass, so a guard-skipped clone was forgotten after one session and the directory became a permanent orphan even after the operator cleared the blocker. Now the new set is `current ∪ (prev entries the guards refused to delete)`, gated by the new `isFinalRemoveOutcome` helper.

**Verified:**
- `isFinalRemoveOutcome` is exhaustive against `RemoveCloneOutcome`: 3 final (`removed`/`absent`/`not-a-clone`) + 6 retryable = full 9-variant union.
- `tryRemoveCloneSafely` is synchronous — the new `const outcome = …` is correctly un-`await`ed.
- `retainedForRetry` only accumulates prev-not-in-current paths (loop `continue`s on `currentSet`), so the union stays clean and a later re-add is still tracked.
- No leakage: the only other reader of `managed.json::sourceRepos` is the next reconcile's diff. The agent briefing's `## Source Repositories` is driven by the payload-derived `sourceReposForPrompt`, not managed state, so retained-for-retry entries never surface to the LLM.
- New test covers the full retry cycle (session 1 skips dirty → operator cleans → session 2 removes & drops). Test helpers line up.

**Non-blocking notes (no action required):**
- The accepted tradeoff — recurring "skipped — working tree is dirty" log lines on persistent blockers — is reasonable and documented. If it ever proves noisy, per-path log dedup is a fine follow-up, not a blocker.
- `not-a-clone` is treated as final (dropped). Correct/conservative: a dir without `.git/` was never ours to manage; leaving it on disk is the safe choice.

CI green across the board (Test / Lint & Type / CodeQL / Migrations / legacy-name guard). No data-structure/schema change, and the `best-effort-one-shot → retry` semantics is implementation detail correctly captured in `docs/development/agent-workspace-state.md` — no Context Tree write needed.

baixiaohang (approved):
## Review by codex-assistant - PR #913

**建议**: approve
- 理由: 改动把 state-based source-repo cleanup 从一次性丢弃改成 guard-skipped 继续追踪、下一 session retry，和 predeclared source repo 的持久 clone / safety guard 合同一致；未发现阻塞风险。

**分级**: A
- 语义升级: 无

**PR 摘要**
- 解决问题: guard-skipped source repo cleanup 过去会从 `managed.json::sourceRepos` 中被遗忘，导致 operator 后续清掉 blocker 后也无法自动重试删除。
- 设计方案: `reconcileSourceRepoState` 将 next state 写成 `current ∪ retainedForRetry`；`removed` / `absent` / `not-a-clone` 作为 final outcome drop，其它 guard / probe / rm failure 保留到下一 session retry。
- 受影响模块: `packages/client/src/runtime/source-repos.ts`, `packages/client/src/runtime/source-repo-cleanup.ts`, source repo reconcile tests, `docs/development/agent-workspace-state.md`

**Review 意见**
✅ 1. 核心状态边界正确：只有 state reconcile 使用新 helper 决定 retention，`workspace-migrations.ts` 仍保持 marker-gated one-shot。
✅ 2. Safety guard 分类和 client runtime 的 per-agent standalone source repo 模型一致；dirty / ahead / worktrees / live-use / probe / rm failure 都保留 operator-clearable recovery path。
✅ 3. `Set([...currentSet, ...retainedForRetry]).sort()` 保持 state 去重和稳定写入；current config 重新包含同一路径时不会被 cleanup 分支误删。
✅ 4. GitHub CI 当前全绿；我按 PR Review 模式未运行本地测试。
⚠️ 5. 非阻塞：persistent blocker 会每 session 重复 log，这是 PR body 明确接受的 trade-off。

**已执行**

PR #921, opened 2026-06-09
feat(web): onboarding connect-code new-tab install + copy/state cleanup

Gandy2025:
## Summary

Onboarding flow polish, focused on the **connect-code** (admin) step, plus copy/runtime refinements across the wizard. Preview gallery (`/preview/onboarding`) renders the real components, so it tracks all of this automatically.

### Connect-code install in a new tab
- `handleConnect` opens the GitHub App install in a **new tab** opened synchronously inside the click gesture (so it isn't popup-blocked after the async URL mint), then fills the tab's location with the minted install URL (`next=/onboarding/connected`). The original tab keeps polling the installation query and advances on its own. Falls back to a full-page `window.location.assign` if the popup was blocked.
- New **public** landing route `/onboarding/connected` → `GithubConnectedPage`, which confirms and auto-closes the script-opened install tab (`role="status"` for a11y).
- Server: `/onboarding/connected` added to `ALLOWED_POST_INSTALL_NEXT` so `resolvePostInstallNext` validates the post-install redirect (exact-match allowlist — no open-redirect surface).
- Install CTA is disabled during the post-launch "waiting" window so a second click can't open another tab / re-mint; re-enabled once we detect the user returned without an install (`postAttemptStuck`) so retry still works.

### Dropped dead `scope-missing` state (connect-code)
The admin repos endpoint (`/orgs/:id/github/repos`) is **installation-token** based (server-minted), not the caller's OAuth — its failures are 502 (upstream) / 503 (no_installation / suspended / not_configured). The only 403 is `requireOrgAdmin` ("not an org admin"), which is unreachable in an admin-only step and not fixable by reconnecting GitHub. So the `scopeMissing` special-case + "Reconnect GitHub" link were removed; **all** repo-list errors now fold into one honest `loadFailed` message.

> The **invitee** kickoff picker (`/me/github/repos`, OAuth-based) genuinely returns `403 scope_missing` on a missing `repo` scope, so it **keeps** its scope-missing + reconnect path — that asymmetry is intentional.

### Copy / UI cleanup
- Title pluralized: "Connect the repos your agent works on".
- Merged the non-owner caveat + skip reassurance into one muted `notOwnerHint` line (gating fact bolded); removed the now-redundant "Connected" status row (PhaseNav ✓ + picker already convey it); gated the "Which repos" header to the pickable case.
- Earlier commits in the branch: onboarding copy/runtime-selection polish + preview states; preselect granted repos; two-phase (Connect GitHub → Pick repos) indicator.

[...]

PR #922, opened 2026-06-09
feat(web): onboarding connect-code new-tab install + copy/state cleanup

Gandy2025:
## Summary

Onboarding flow polish, focused on the **connect-code** (admin) step, plus copy/runtime refinements across the wizard. Preview gallery (`/preview/onboarding`) renders the real components, so it tracks all of this automatically.

### Connect-code install in a new tab
- `handleConnect` opens the GitHub App install in a **new tab** opened synchronously inside the click gesture (so it isn't popup-blocked after the async URL mint), then fills the tab's location with the minted install URL (`next=/onboarding/connected`). The original tab keeps polling the installation query and advances on its own. Falls back to a full-page `window.location.assign` if the popup was blocked.
- New **public** landing route `/onboarding/connected` → `GithubConnectedPage`, which confirms and auto-closes the script-opened install tab (`role="status"` for a11y).
- Server: `/onboarding/connected` added to `ALLOWED_POST_INSTALL_NEXT` so `resolvePostInstallNext` validates the post-install redirect (exact-match allowlist — no open-redirect surface).
- Install CTA is disabled during the post-launch "waiting" window so a second click can't open another tab / re-mint; re-enabled once we detect the user returned without an install (`postAttemptStuck`) so retry still works.

### Dropped dead `scope-missing` state (connect-code)
The admin repos endpoint (`/orgs/:id/github/repos`) is **installation-token** based (server-minted), not the caller's OAuth — its failures are 502 (upstream) / 503 (no_installation / suspended / not_configured). The only 403 is `requireOrgAdmin` ("not an org admin"), which is unreachable in an admin-only step and not fixable by reconnecting GitHub. So the `scopeMissing` special-case + "Reconnect GitHub" link were removed; **all** repo-list errors now fold into one honest `loadFailed` message.

> The **invitee** kickoff picker (`/me/github/repos`, OAuth-based) genuinely returns `403 scope_missing` on a missing `repo` scope, so it **keeps** its scope-missing + reconnect path — that asymmetry is intentional.

### Copy / UI cleanup
- Title pluralized: "Connect the repos your agent works on".
- Merged the non-owner caveat + skip reassurance into one muted `notOwnerHint` line (gating fact bolded); removed the now-redundant "Connected" status row (PhaseNav ✓ + picker already convey it); gated the "Which repos" header to the pickable case.
- Earlier commits in the branch: onboarding copy/runtime-selection polish + preview states; preselect granted repos; two-phase (Connect GitHub → Pick repos) indicator.

## Review
Two independent code-review passes (both SHIP / SHIP-WITH-NITS, no HIGH findings). Both independently verified: the scope-missing-removal premise, the allowlist open-redirect guard stays sound, and no leaked/dangling tab on any path. Their actionable findings (double-tab guard, `role="status"`, popup-blocked + landing-page test coverage) are in the final commit.

## Test plan
- `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) — green
- `pnpm --filter @first-tree/web test` — green; added coverage for the new-tab open (`next=/onboarding/connected`), the popup-blocked full-redirect fallback, 403→loadFailed, the post-install-next allowlist, and an SSR render of `GithubConnectedPage`
- Server unit test `post-install-next.test.ts` updated (pure-function; runs in CI with Postgres)

### DIFF
--- a/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
+++ b/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
@@ -9,6 +9,7 @@ import { ALLOWED_POST_INSTALL_NEXT, resolvePostInstallNext } from "../github-app
 describe("resolvePostInstallNext", () => {
   it("passes through allowlisted internal paths", () => {
     expect(resolvePostInstallNext("/onboarding")).toBe("/onboarding");
+    expect(resolvePostInstallNext("/onboarding/connected")).toBe("/onboarding/connected");
     expect(resolvePostInstallNext("/settings/github")).toBe("/settings/github");
   });
 
@@ -24,7 +25,7 @@ describe("resolvePostInstallNext", () => {
     expect(resolvePostInstallNext("")).toBe("/settings/github");
   });
 
-  it("only allows the two known internal destinations", () => {
-    expect([...ALLOWED_POST_INSTALL_NEXT].sort()).toEqual(["/onboarding", "/settings/github"]);
+  it("only allows the known internal destinations", () => {
+    expect([...ALLOWED_POST_INSTALL_NEXT].sort()).toEqual(["/onboarding", "/onboarding/connected", "/settings/github"]);
   });
 });
--- a/packages/server/src/api/orgs/github-app.ts
+++ b/packages/server/src/api/orgs/github-app.ts
@@ -40,7 +40,14 @@ const POST_INSTALL_NEXT = "/settings/github";
  * Allowlisted — never reflect a caller-supplied path verbatim into the
  * signed redirect, so a crafted `?next=` can't become an open redirect.
  */
-export const ALLOWED_POST_INSTALL_NEXT: ReadonlySet<string> = new Set([POST_INSTALL_NEXT, "/onboarding"]);
+export const ALLOWED_POST_INSTALL_NEXT: ReadonlySet<string> = new Set([
+  POST_INSTALL_NEXT,
+  "/onboarding",
+  // Tiny "connected — you can close this tab" landing: onboarding connect-code
+  // installs in a popup and lands the popup here so it can auto-close while the
+  // original tab keeps polling.
+  "/onboarding/connected",
+]);
 
 /**
  * Resolve the post-install redirect destination from a caller-supplied
--- a/packages/web/src/app.tsx
+++ b/packages/web/src/app.tsx
@@ -16,6 +16,7 @@ import { ContextPage } from "./pages/context.js";
 import { InviteAcceptPage } from "./pages/invite-accept.js";
 import { LoginPage } from "./pages/login.js";
 import { OAuthCompletePage } from "./pages/oauth-complete.js";
+import { GithubConnectedPage } from "./pages/onboarding/github-connected.js";
 import { OnboardingPage } from "./pages/onboarding/onboarding-page.js";
 import { SettingsComputersPage } from "./pages/settings/computers.js";
 import { SettingsContextTreePage } from "./pages/settings/context-tree.js";
@@ -97,6 +98,8 @@ export function App() {
               {/* /signup retired — Continue with GitHub on /login covers signup. */}
               <Route path="/signup" element={<Navigate to="/login" replace />} />
               <Route path="/auth/github/complete" element={<OAuthCompletePage />} />
+              {/* Public: the connect-code install popup lands here to auto-close. */}
+              <Route path="/onboarding/connected" element={<GithubConnectedPage />} />
               <Route path="/invite/:token" element={<InviteAcceptPage />} />
               {ContextPreviewPage ? (
                 <Route
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1221,7 +1221,7 @@ describe("page SSR smoke coverage", () => {
     expect(html).toContain("Heads up");
     expect(html).toContain("gandy-macbook connected");
     expect(html).toContain("Install Node.js");
-    expect(html).toContain("Choose your projects");
+    expect(html).toContain("Pick repos");
 
     expect(await renderOnboardingStep(<StepTeam />, { activeStep: "team" })).toContain(
       "What should we call your team?",
@@ -1235,11 +1235,11 @@ describe("page SSR smoke coverage", () => {
     );
     expect(
       await renderOnboardingStep(<StepCreateAgent />, { activeStep: "create-agent", agentPhase: "creating" }),
-    ).toContain("Setting up your agent");
+    ).toContain("Bringing your agent online");
     expect(
       await renderOnboardingStep(<StepCreateAgent />, { activeStep: "create-agent", agentPhase: "timeout" }),
-    ).toContain("longer than expected");
-    expect(await renderOnboardingStep(<StepConnectCode />, { activeStep: "connect-code" })).toContain("Which projects");
+    ).toContain("online yet");
+    expect(await renderOnboardingStep(<StepConnectCode />, { activeStep: "connect-code" })).toContain("Which repos");
     expect(await renderOnboardingStep(<StepKickoff />, { activeStep: "kickoff" })).toContain(
       "Use your team&#x27;s Context Tree",
     );
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1188,6 +1188,7 @@ describe("page SSR smoke coverage", () => {
     const { ConnectStuckPanel } = await import("../../components/connect-stuck-panel.js");
     const { CommandBox, FlowNote, RepoPicker, StatusRow, WorkingState } = await import("../onboarding/flow-ui.js");
     const { InstallGuide, ShowMeHow, TerminalGuide } = await import("../onboarding/guides.js");
+    const { GithubConnectedPage } = await import("../onboarding/github-connected.js");
     const { StepConnectCode } = await import("../onboarding/steps/step-connect-code.js");
     const { StepConnectComputer } = await import("../onboarding/steps/step-connect-computer.js");
     const { StepCreateAgent } = await import("../onboarding/steps/step-create-agent.js");
@@ -1222,6 +1223,13 @@ describe("page SSR smoke coverage", () => {
     expect(html).toContain("gandy-macbook connected");
     expect(html).toContain("Install Node.js");
     expect(html).toContain("Pick repos");
+
+    // The install-popup landing page (auto-closes the script-opened tab; the
+    // effect is a no-op under SSR). Confirms it renders + carries role=status.
+    const connectedHtml = renderPage(<GithubConnectedPage />);
+    expect(connectedHtml).toContain("Connected");
+    expect(connectedHtml).toContain("close this tab");
+    expect(connectedHtml).toContain('role="status"');
 
     expect(await renderOnboardingStep(<StepTeam />, { activeStep: "team" })).toContain(
       "What should we call your team?",
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1239,7 +1239,7 @@ describe("page SSR smoke coverage", () => {
     expect(
       await renderOnboardingStep(<StepCreateAgent />, { activeStep: "create-agent", agentPhase: "timeout" }),
     ).toContain("online yet");
-    expect(await renderOnboardingStep(<StepConnectCode />, { activeStep: "connect-code" })).toContain("Which repos");
+    expect(await renderOnboardingStep(<StepConnectCode />, { activeStep: "connect-code" })).toContain("Pick repos");
     expect(await renderOnboardingStep(<StepKickoff />, { activeStep: "kickoff" })).toContain(
       "Use your team&#x27;s Context Tree",
     );
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -1333,10 +1333,10 @@ describe("web DOM interaction coverage", () => {
     });
 
     const disconnected = await renderOnboardingDom(<StepConnectCode />, { activeStep: "connect-code" });
-    await waitForText("Install First Tree on GitHub", disconnected.container);
+    await waitForText("Install on GitHub", disconnected.container);
     await click(
       [...disconnected.container.querySelectorAll("button")].find((button) =>
-        button.textContent?.includes("Install First Tree on GitHub"),
+        button.textContent?.includes("Install on GitHub"),
       ) ?? null,
     );
     expect(githubAppMocks.getGithubAppInstallUrl).toHaveBeenCalledWith("org-1", "/onboarding");
@@ -1346,7 +1346,7 @@ describe("web DOM interaction coverage", () => {
     // Skip is one click now — a legitimate, recoverable choice goes straight
     // through with no confirm gate (the old "Skip connecting code?" panel +
     // Keep-connecting / Skip-anyway was confirmshaming and was removed).
-    expect(disconnected.container.textContent).toContain("You can connect code anytime from Settings.");
+    expect(disconnected.container.textContent).toContain("You can connect a repo anytime from Settings.");
     await click(
       [...disconnected.container.querySelectorAll("button")].find((button) =>
         button.textContent?.includes("Skip for now"),
@@ -1375,7 +1375,7 @@ describe("web DOM interaction coverage", () => {
       setSelectedRepoUrls,
       goNext: vi.fn(),
     });
-    await waitForText("Which projects should your agent work on?", connected.container);
+    await waitForText("Which repos should your agent work on?", connected.container);
     await waitForText("acme/web", connected.container);
     await click(
       [...connected.container.querySelectorAll("label")].find((label) => label.textContent?.includes("acme/web")) ??
@@ -1384,7 +1384,7 @@ describe("web DOM interaction coverage", () => {
     expect(setSelectedRepoUrls).toHaveBeenCalledWith(["https://github.com/acme/web.git"]);
     await click(
       [...connected.container.querySelectorAll("button")].find((button) =>
-        button.textContent?.includes("Continue without a project"),
+        button.textContent?.includes("Continue without a repo"),
       ) ?? null,
     );
     expect(connected.flow.goNext).toHaveBeenCalled();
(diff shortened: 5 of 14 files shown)


yuezengwu (approved):
### Review — onboarding connect-code new-tab install + copy/state cleanup

Reviewed the full diff (identical to the closed #921). **Goal & core change:** the connect-code (admin) install now opens in a new tab synchronously inside the click gesture and fills its location with the minted install URL (`next=/onboarding/connected`); the original tab keeps polling and advances on its own. A new public `/onboarding/connected` landing page auto-closes the script-opened tab. Plus: drop the dead `scope-missing` path on connect-code, fold all repo-list errors into one `loadFailed`, runtime single-select on connect-computer, and a copy pass.

**Verified the premises:**
- **scope-missing removal is sound.** Confirmed against the context tree (`system/cloud/github/github-app.md` role matrix): the connect-code picker is installation-token-backed org repos and is **admin-only**, so the only 403 is `requireOrgAdmin` ("not an org admin") — not a GitHub-scope problem and not fixable by reconnecting. Folding it into `loadFailed` is correct. The invitee picker (`/me/github/repos`, OAuth-based) genuinely returns 403 scope_missing and correctly **keeps** its reconnect path — the asymmetry is intentional and right.
- **No open-redirect.** `ALLOWED_POST_INSTALL_NEXT` stays an exact-match allowlist; `/onboarding/connected` is just one more literal. Good.
- **No tab leak.** Success path auto-closes; mint-failure path closes the blank tab via `installTab?.close()`; popup-blocked path opens nothing. Clean.
- CTA disable (`attempted && !postAttemptStuck`) correctly blocks a second tab / re-mint while waiting, and re-enables on `postAttemptStuck` for retry.

### Finding 1 — popup-blocked fallback lands the user on a dead-end (non-blocking)

In `handleConnect`, the popup-blocked branch does `window.location.assign(url)` where `url` was minted with `next=/onboarding/connected`. So the **main** onboarding tab navigates to GitHub and, after install, returns to `GithubConnectedPage` — which (a) calls `window.close()` that silently no-ops because this tab was *not* script-opened, and (b) renders "You can close this tab — setup continues in your other tab", but there is no other tab (the wizard was in this very tab). The user is stranded on a confirmation page after a successful install, with no link back into onboarding.

This is a small regression vs. the pre-PR behavior, where the full-page redirect used `next=/onboarding` and resumed the wizard (now installed → picker). Suggested fixes, either works:
- mint with `/onboarding` in the popup-blocked branch, or
- have `GithubConnectedPage` redirect to `/onboarding` when `!window.opener` (i.e. it's not actually a script-opened popup).

Marking **non-blocking**: popups opened synchronously in a click gesture are rarely blocked, and the install still succeeds so the user can recover by reloading `/onboarding`. Worth a follow-up (or an explicit "won't fix" if you consider the blocked path negligible).

Otherwise this is clean and well-tested (new-tab open, popup-blocked fallback, 403→loadFailed, allowlist, SSR of the landing page all covered). Approving. No data-structure / DB changes; the server delta is a pure allowlist addition. Latest CI run is green (Check Branch Name now passes on the convention-compliant branch; Test still finishing).

yuezengwu (approved):
Reviewed. Solid UI/copy polish on the connect-code step; the one substantive logic change (dropping `scopeMissing`) checks out against the server.

**Goal & core changes**
- connect-code installs the GitHub App in a synchronously-opened new tab (`next=/onboarding/connected`); the original tab keeps polling and advances on its own. Full-page `window.location.assign` fallback if the popup is blocked.
- New public `/onboarding/connected` landing (`GithubConnectedPage`) auto-closes the script-opened tab; server adds it to `ALLOWED_POST_INSTALL_NEXT` (exact-match).
- Dropped the dead `scopeMissing` / "Reconnect GitHub" path in connect-code; all repo-list errors fold into one `loadFailed` message.
- Copy/runtime polish (project→repo, PhaseNav two-phase indicator, runtime pills, preselect granted repos, computer-disconnected hint).

**Verified**
- *scopeMissing removal* ✓ — `GET /orgs/:orgId/github-app-installation/repositories` is `requireOrgAdmin` + installation-token based; its failures are explicit 503 (`no_installation`/`suspended`/`not_configured`) / 502 (`upstream`). The only 403 is "not an org admin", which is unreachable in the admin-only connect-code step and not fixable by reconnecting GitHub. Invitee picker (`/me/github/repos`, OAuth) correctly keeps its scope-missing path — the asymmetry is right.
- *Open-redirect* ✓ — allowlist stays exact-match; `GithubConnectedPage` is public but renders nothing sensitive.
- *Dangling tab* ✓ — error path closes the blank tab (`installTab?.close()`); double-tab guard via the disabled CTA (`attempted && !postAttemptStuck`).

**Finding 1 (non-blocking, popup-blocked fallback)**
`handleConnect` mints the URL once with `next=/onboarding/connected` and reuses it for both the new-tab fill and the popup-blocked `window.location.assign`. In the blocked path the *only* tab navigates to GitHub and returns to `GithubConnectedPage`, which then attempts `window.close()` (a no-op for a non-script-opened tab) and shows "close this tab — setup continues in your other tab" — but there is no other tab, and the original flow tab is gone. The install still succeeds and re-navigating to `/onboarding` recovers (it detects the installation), so it's recoverable, but the landing copy misleads in that single-tab case. Consider minting `/onboarding` for the blocked fallback, or softening the landing copy when it isn't a script-opened tab. The added test only asserts `assign` was called, not the return landing.

No core data-structure or DB changes; no migration. Approving — Finding 1 is a follow-up, not a blocker.

baixiaohang (changes requested):
## Review by codex-assistant - PR #922

- Reason: the popup-blocked fallback returns the only onboarding tab to a popup-only landing page, so the advertised full-page fallback cannot resume setup after a successful install.

**Level**: B-low
- Path initial: includes `packages/server/**` outside migrations / Prisma schema / core service files; the rest is `packages/web/**` -> B-low.
- Semantic lift: none.

**PR Summary**
- Solves: polish onboarding connect-code and related wizard copy/runtime states.
- Design: open GitHub App install in a new tab with `/onboarding/connected` as the popup landing, add a public auto-close page and exact post-install allowlist entry, fold admin repo-list errors into one load-failed state, and update copy/previews/tests.
- Affected modules: `packages/web/src/pages/onboarding/**`, `packages/web/src/app.tsx`, `packages/server/src/api/orgs/github-app.ts`, and related tests.

**Review Findings**
❌ 1. Popup-blocked fallback strands the main onboarding tab. [R4 / `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:128`]

When `window.open` returns null, `handleConnect` still mints the install URL with `next=/onboarding/connected`, then sends the current onboarding tab there via `window.location.assign` (`step-connect-code.tsx:128-132`). After GitHub completes, that same tab renders `GithubConnectedPage`, whose only recovery behavior is `window.close()` plus “setup continues in your other tab” (`github-connected.tsx:15-45`). In the blocked-popup path there is no other tab, and browsers will not close the original user-opened tab, so the fallback becomes a dead end instead of resuming the wizard. Please make the blocked-popup path return to `/onboarding` (or make `/onboarding/connected` detect `!window.opener` and navigate back), and cover that post-install return. The current test only asserts the assign call while also requiring the dead-end `next=/onboarding/connected`.

✅ 2. The exact post-install allowlist remains sound; adding `/onboarding/connected` does not reflect arbitrary caller-supplied URLs.

✅ 3. The connect-code `scope_missing` removal matches the durable boundary: the admin repo list is installation-token-backed, while the invitee OAuth picker keeps its reconnect path.

**Executed**

yuezengwu (changes requested):
### Revising my earlier approve → request changes

@baixiaohang / codex-assistant raised a finding on the closed #921 that I'd missed. I independently verified it against the source (`postAttemptStuck` timer, which isn't in the diff) and the context tree, and it holds — it defeats one of this PR's stated safety properties, so I'm withdrawing my approval.

**The `postAttemptStuck` precondition no longer holds in the new-tab design.** That signal was built to mean *"the user navigated to GitHub and came back without an install"* — valid in the old full-page-redirect flow, where the only way to be on this page again was to have returned. The timer (`step-connect-code.tsx` lines 61–74) arms purely on *"attempt marker present AND `installQuery.isLoading === false`"*, then flips `postAttemptStuck` true after **5s**.

In the new-tab flow the original tab **never leaves**, and `installQuery` polls every 4s, so `isLoading` goes false after the first poll. Result: ~5s after the click — while the user is still in the GitHub tab choosing an org / selecting repos — the original tab:
1. auto-opens "Need help?" and shows the "still not connected" stuck hint *during a perfectly healthy in-progress install*, and
2. re-enables the Install CTA (`disabled={redirecting || (attempted && !postAttemptStuck) || …}`).

If the user then clicks Install again (plausible — the page is telling them it's stuck), the second `/install-url` mint **overwrites the per-browser `oauth_state_nonce` cookie**. The first install tab's callback then fails its state-nonce check — exactly the cookie-pairing the code's own (removed) comment and `github-app.md` (onboarding path 4) describe. So the double-tab / re-mint this PR set out to prevent is still reachable, and now it actively breaks the in-flight first install (the install lands unbound on GitHub's side, recoverable only via the orphan-reclaim sweep).

**Fix direction** (any one): don't lift the CTA disable on an elapsed-time timer in this flow — the original tab can't infer "stuck" from time alone now that it never navigates away. Gate re-enable on an explicit signal that the install tab actually came back empty (e.g. a `postMessage` / `BroadcastChannel` / storage event from `/onboarding/connected`, or a user-initiated "try again"), and decouple the premature "stuck" hint from the mint button. Server-side, making concurrent attempts not share an overwritable state cookie also closes it but is heavier.

---

**Finding 2 (still open, non-blocking) — popup-blocked fallback dead-ends.** As noted earlier: the `window.location.assign(url)` fallback navigates the *main* tab with `next=/onboarding/connected`, so after install it lands on `GithubConnectedPage` — which can't `window.close()` (not script-opened) and tells the user to "close this tab, setup continues in your other tab" when there is no other tab. Pre-PR the fallback used `next=/onboarding` and resumed the wizard. Mint `/onboarding` in the blocked branch, or have the landing page redirect to `/onboarding` when `!window.opener`.

The rest of the PR (scope-missing removal — verified sound against the github-app role matrix; exact-match allowlist; no tab leak; runtime single-select; copy) is good. Happy to re-approve once Finding 1 is addressed.

baixiaohang (changes requested):
## Review by codex-assistant - PR #922

- Reason: the normal new-tab retry guard is time-based rather than return-based, so a second install can be launched while the first GitHub tab is still in flight and invalidate its OAuth state cookie.

**Level**: B-low
- Path initial: includes `packages/server/**` outside migrations / Prisma schema / core service files; the rest is `packages/web/**` -> B-low.
- Semantic lift: none.

**PR Summary**
- Solves: polish onboarding connect-code and related wizard copy/runtime states.
- Design: open GitHub App install in a new tab with `/onboarding/connected` as the popup landing, add a public auto-close page and exact post-install allowlist entry, fold admin repo-list errors into one load-failed state, and update copy/previews/tests.
- Affected modules: `packages/web/src/pages/onboarding/**`, `packages/web/src/app.tsx`, `packages/server/src/api/orgs/github-app.ts`, and related tests.

**Review Findings**
❌ 1. The normal new-tab retry guard can still re-mint while the first install is in flight. [R4 / `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:61`, `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:187`, `packages/server/src/api/orgs/github-app.ts:236`]

This is separate from the popup-blocked fallback issue already raised. In the successful popup path, the original onboarding tab never navigates away; it stays mounted and keeps polling. The `postAttemptStuck` effect starts a 5-second timer as soon as the installation query is settled with no row, not when the user has actually returned/cancelled the GitHub install tab. Once that timer fires, the disabled condition allows another Install click. That second click calls `/install-url` again, which writes the same `oauth_state_nonce` cookie name for a new signed state. If the first GitHub tab later completes, its callback carries the old state but the browser now has the second nonce, so the callback fails nonce verification. This reintroduces the double-tab/re-mint race the PR is trying to close, and can break a legitimate first install.

Please keep the CTA locked until a real return/cancel signal exists, or make concurrent attempts unable to invalidate each other (for example by not sharing one overwriteable state cookie across live attempts). The existing tests cover “open returns null -> assign called” and repo-list errors, but not “normal popup stays open longer than 5s -> original tab allows a second mint”.

**Executed**

yuezengwu (changes requested):
Withdrawing my earlier approve — codex-assistant surfaced a second finding I missed, and I've now confirmed both against the code. Both are valid blockers; the re-mint race is the more serious of the two.

**Blocker A — re-mint race overwrites the OAuth state nonce (codex finding, confirmed).** `step-connect-code.tsx`: the `postAttemptStuck` re-enable is a fixed 5s timer keyed on the install query settling, NOT on the user actually returning/cancelling the GitHub tab. The original tab stays mounted, so after 5s the CTA re-enables while the first install is still in flight. Verified server-side:
- `OAUTH_STATE_COOKIE = "oauth_state_nonce"` is one fixed cookie; `/install-url` (`github-app.ts:239`) re-sets it with a fresh nonce on every call.
- `verifyOAuthState` (`services/oauth-state.ts`) mandates a double-submit match (state JWT nonce == cookie nonce), else 401.
- So a second mint overwrites the cookie → the first tab's callback carries the now-stale nonce → fails verification → a legitimate first install breaks.

This isn't limited to the popup-blocked path — it hits the normal new-tab flow whenever the GitHub install dialog takes longer than 5s (common for first-timers), and the "stuck → Need help? → click Install again" copy actively invites the re-click while the first install is still genuinely in progress. Fix direction: gate the CTA on a real return/cancel signal (e.g. visibility/focus regain, or the install tab closing) instead of the timer, or stop sharing one overwriteable state cookie across concurrent attempts. Add coverage for "popup stays open >5s → original tab must not allow a second mint".

**Blocker B — popup-blocked fallback strands the only tab (codex finding + my earlier Finding 1).** The blocked path reuses `next=/onboarding/connected`, so the single tab returns to `GithubConnectedPage`, which only does a no-op `window.close()` and says "setup continues in your other tab" — there is none, and no in-product affordance resumes the wizard. Mint `/onboarding` for the blocked fallback, or have `GithubConnectedPage` detect `!window.opener` and navigate back; test the post-install return, not just the `assign` call.

**Confirmed sound (unchanged):** the exact-match post-install allowlist (no open-redirect) and the `scope_missing` removal (admin repo list is installation-token-backed; invitee OAuth picker correctly keeps its reconnect path).

Gandy2025:
Thanks @baixiaohang and @yuezengwu — both blockers are real and now fixed in e805e104.

**Blocker A — re-mint race overwrites the OAuth state nonce.** You're right that `postAttemptStuck` was a full-redirect-era signal ("came back from GitHub empty") that's meaningless in the new-tab flow: the original tab never navigates away, so the 5s timer fired mid-install and re-enabled the CTA, letting a second `/install-url` clobber the single `oauth_state_nonce` cookie and break the first tab's double-submit check. Fix: **removed the time-based stuck/re-enable entirely** (state + timer effect + auto-open-help-on-timer). The CTA is now locked whenever an attempt is in flight (`disabled = redirecting || attempted || !organizationId`), and retry is an **explicit, user-initiated "Start over"** that abandons the in-flight attempt before re-minting — never a timed auto-unlock. Added a guard test: after launch the CTA is `disabled`, a second click does not re-mint, and only "Start over" re-enables it.

**Blocker B — popup-blocked fallback stranded the only tab.** Since `installTab === null` is known *before* the mint, the blocked path now mints `next=/onboarding` (this tab resumes the wizard); the popup path keeps `next=/onboarding/connected` (auto-close). Updated the test to assert the blocked path mints `/onboarding`.

Also removed the now-unused `stuckStatus` copy. `scope_missing` removal, the exact-match allowlist, and the invitee asymmetry are unchanged. PTAL.

yuezengwu (approved):
Re-reviewed e805e104 — both blockers correctly fixed. Approving.

**Blocker A (re-mint nonce race) — fixed.** Dropped `postAttemptStuck` + the 5s timer + the auto-open-help-on-timer effect entirely. The CTA is now `disabled={redirecting || attempted || !organizationId}` — locked for the whole in-flight window with no timed auto-unlock, so no accidental second `/install-url` can clobber the first attempt's `oauth_state_nonce`. Retry is the explicit `handleStartOver` (clears the marker, `setAttempted(false)`); a second mint now only happens when the user consciously chooses to abandon the first attempt. This closes the race cleanly.

**Blocker B (popup-blocked dead-end) — fixed.** `postInstallNext = installTab ? "/onboarding/connected" : "/onboarding"` — the blocked single-tab path resumes the wizard via `/onboarding` (already allowlisted) and advances on the now-present installation; the popup path keeps the auto-close landing. No stranded tab.

Verified `stuckStatus` copy removed and `restartInstall` added; the guard test (CTA locks after launch, only "Start over" re-mints) and the blocked-path `/onboarding` assertion are in. scope_missing removal / exact-match allowlist / invitee asymmetry unchanged.

(Approving on code inspection; CI was re-running at review time.)

baixiaohang (approved):
## Review by codex-assistant - PR #922 follow-up

**Recommendation**: approve
- Reason: the new commit removes the time-based re-mint path and separates popup vs full-page post-install destinations, closing both blockers without changing the broader GitHub App boundary.

**Level**: B-low
- Path initial: includes `packages/server/**` outside migrations / Prisma schema / core service files; the rest is `packages/web/**` -> B-low.
- Semantic lift: none.

**PR Summary**
- Solves: follow-up fix for connect-code new-tab install retry safety and popup-blocked fallback recovery.
- Design: `postAttemptStuck` and its 5s timed unlock are removed; the Install CTA locks for any in-flight attempt; retry is an explicit `Start over` action; popup installs still land on `/onboarding/connected`, while popup-blocked full-page installs mint `next=/onboarding`.
- Affected modules: `packages/web/src/pages/onboarding/steps/step-connect-code.tsx`, `packages/web/src/pages/onboarding/copy.ts`, and focused DOM tests.

**Review Findings**
✅ 1. Blocker A is fixed. The original tab no longer infers “stuck” from elapsed time, and the Install CTA is disabled by `attempted` until installation lands or the user explicitly starts over, so a normal long GitHub install cannot auto-unlock a second `/install-url` mint. [R4 / `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:56`, `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:189`]

✅ 2. Blocker B is fixed. The post-install target now branches on whether the synchronously opened tab exists: popup path uses `/onboarding/connected`, blocked-popup full-page path uses `/onboarding`; the focused test asserts the blocked path mints `/onboarding`. [R4 / `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:114`, `packages/web/src/pages/__tests__/web-dom-interactions.test.tsx:1478`]

✅ 3. The earlier sound parts still hold: exact-match post-install allowlist remains narrow, connect-code repo listing stays installation-token-backed/admin-only, and the invitee OAuth picker keeps its scope-missing reconnect path.

**Executed**
- Submitted `approve` review. I did not run local tests per PR review policy.

PR #923, opened 2026-06-09
feat(context-tree): add GitHub repo initializer

bestony:
## Summary
- Add an admin-only Context Tree initializer at `POST /api/v1/orgs/:orgId/context-tree/initialize`.
- Create a private GitHub repo under the signed-in user, write the root `NODE.md`, and persist the org `context_tree` binding.
- Surface the initializer in Settings > Context tree and the `/context` unavailable state, with admin/member-specific UI.

## Testing
- `pnpm --filter @first-tree/server test -- src/__tests__/context-tree-initialize.test.ts src/__tests__/me-onboarding.test.ts src/__tests__/org-settings.test.ts`
- `pnpm --filter @first-tree/web test -- src/pages/__tests__/settings-panels-dom.test.tsx src/pages/__tests__/context-page-dom.test.tsx src/api/__tests__/api-wrappers.test.ts`
- `pnpm typecheck`

Note: full local `pnpm check` is currently blocked by unrelated untracked `packages/skill-evals/` formatting diagnostics and a pre-existing `packages/client/src/runtime/workspace-migrations.ts` lint warning.

### DIFF
--- a/docs/github-app-design-zh.md
+++ b/docs/github-app-design-zh.md
@@ -1,0 +1,97 @@
+# GitHub App Design
+
+This document records the GitHub App decisions referenced from code comments.
+The filename keeps the historical `-zh` suffix, but the repository standard is
+to write technical documentation in English.
+
+## D0a: One GitHub App, SaaS-Wide
+
+First Tree uses one centrally configured GitHub App per deployment. The App is
+the trust boundary for GitHub identity, installation tokens, repository access,
+and webhook delivery.
+
+The server stores installation metadata in `github_app_installations`. The row
+is a snapshot from GitHub webhooks, GitHub API reads, or the local development
+stub; GitHub remains the source of truth for the live permission grant.
+
+## D0b: GitHub App Settings Contract
+
+Permissions are declared in the GitHub App settings page, not passed in OAuth
+or install URLs. Install and authorize URLs only carry identity, callback, and
+state parameters; GitHub renders the current App permission contract from the
+App settings.
+
+Repository permissions:
+
+- `Administration`: Read and write
+- `Contents`: Read and write
+- `Pull requests`: Read and write
+- `Issues`: Read-only
+- `Metadata`: Read-only
+
+Organization permissions:
+
+- `Members`: Read-only
+
+Subscribed events:
+
+- `issues`
+- `issue_comment`
+- `pull_request`
+- `pull_request_review`
+- `push`
+- `installation`
+- `installation_repositories`
+- `member`
+
+Operator rollout note:
+
+- Operators must update the GitHub App settings to request
+  `Administration: Read and write`.
+- Existing installations must re-approve the permission upgrade in GitHub
+  before relying on one-click Context Tree initialization.
+- Until an installation is re-approved, initializer calls return
+  `installation_permissions_insufficient`.
+
+## D1: Login and Install Share the App Flow
+
+The GitHub App has "Request user authorization (OAuth) during installation"
+enabled. A first-time user can install the App and authorize identity in the
+same GitHub round trip, while a returning user can authorize without changing
+the installation.
+
+First Tree signs and verifies `state` for CSRF protection and for recovering
+the intended post-auth destination. Install CTAs use the App installation URL
+when the UI needs a repository picker and permission review; identity-only
+sign-in uses the App authorize URL.
+
+## D2: One Installation Binds to One First Tree Team
+
+Each GitHub App installation may bind to at most one First Tree team. The
+server enforces this with the `github_app_installations.hub_organization_id`
+binding and refuses to rebind an installation that is already claimed by a
+different team.
+
+This keeps repository grants, webhook fan-out, Context Tree initialization, and
+team membership semantics aligned around a single team boundary.
+
+## D3: Legacy OAuth and Per-Repo Webhooks Are Removed
+
+The GitHub App is the GitHub integration surface. Legacy per-repository OAuth
+and webhook setup paths are not part of the current contract; remaining code
+must tolerate historical identity metadata only where needed for migration and
+returning users.
+
+Webhook ingestion enters through the GitHub App webhook endpoint and uses the
+installation registry to resolve team ownership.
+
+## D4: Webhooks Drive Installation State
+
+GitHub installation, repository-selection, and member-related events refresh
+First Tree's installation snapshot. The stored `permissions` and `events`
+fields are forward-compatible records so new GitHub keys can be surfaced
+without a schema migration.
+
+Runtime actions that require stronger permissions, such as one-click Context
+Tree initialization, must gate on the live installation token response rather
+than assuming the stored snapshot is current.
--- a/packages/server/src/__tests__/context-tree-initialize.test.ts
+++ b/packages/server/src/__tests__/context-tree-initialize.test.ts
@@ -1,0 +1,328 @@
+import { eq } from "drizzle-orm";
+import type { FastifyInstance } from "fastify";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { authIdentities } from "../db/schema/auth-identities.js";
+import { members } from "../db/schema/members.js";
+import { organizations } from "../db/schema/organizations.js";
+import { decryptValue, encryptValue } from "../services/crypto.js";
+import { getOrgContextTree, putOrgSetting } from "../services/org-settings.js";
+import { uuidv7 } from "../uuid.js";
+import { createTestAdmin, useTestApp } from "./helpers.js";
+
+type TestAdmin = Awaited<ReturnType<typeof createTestAdmin>>;
+type FetchInput = Parameters<typeof globalThis.fetch>[0];
+type FetchInit = Parameters<typeof globalThis.fetch>[1];
+
+const GITHUB_API_BASE = "https://api.github.com";
+const TOKEN_URL = "https://github.com/login/oauth/access_token";
+
+describe("POST /orgs/:orgId/context-tree/initialize", () => {
+  const getApp = useTestApp();
+  let originalFetch: typeof globalThis.fetch;
+
+  beforeEach(() => {
+    originalFetch = globalThis.fetch;
+  });
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  it("creates a private GitHub repo, writes NODE.md, and persists context_tree", async () => {
+    const app = getApp();
+    const admin = await createTestAdmin(app);
+    await seedGithubIdentity(app, admin, { login: "octocat", accessToken: "ghu_live" });
+    await app.db
+      .update(organizations)
+      .set({ displayName: "Acme Labs" })
+      .where(eq(organizations.id, admin.organizationId));
+
+    let createRepoPayload: unknown;
+    let createFilePayload: unknown;
+    mockFetch(async (url, init) => {
+      if (url === `${GITHUB_API_BASE}/user/repos`) {
+        expect(init?.headers).toEqual(expect.objectContaining({ Authorization: "Bearer ghu_live" }));
+        createRepoPayload = parseJsonBody(init);
+        return jsonResponse(
+          {
+            name: "acme-labs-context-tree",
+            full_name: "octocat/acme-labs-context-tree",
+            owner: { login: "octocat" },
+            clone_url: "https://github.com/octocat/acme-labs-context-tree.git",
+            html_url: "https://github.com/octocat/acme-labs-context-tree",
+            private: true,
+            default_branch: "main",
+          },
+          201,
+        );
+      }
+      if (url === `${GITHUB_API_BASE}/repos/octocat/acme-labs-context-tree/contents/NODE.md`) {
+        expect(init?.headers).toEqual(expect.objectContaining({ Authorization: "Bearer ghu_live" }));
+        createFilePayload = parseJsonBody(init);
+        return jsonResponse({ content: { path: "NODE.md" } }, 201);
+      }
+      return new Response(`unexpected fetch ${url}`, { status: 500 });
+    });
+
+    const res = await app.inject({
+      method: "POST",
+      url: `/api/v1/orgs/${admin.organizationId}/context-tree/initialize`,
+      headers: { authorization: `Bearer ${admin.accessToken}` },
+      payload: {},
+    });
+
+    expect(res.statusCode).toBe(201);
+    expect(res.json()).toEqual({
+      repo: "https://github.com/octocat/acme-labs-context-tree.git",
+      htmlUrl: "https://github.com/octocat/acme-labs-context-tree",
+      branch: "main",
+      nodePath: "NODE.md",
+    });
+    expect(createRepoPayload).toMatchObject({
+      name: "acme-labs-context-tree",
+      private: true,
+      auto_init: false,
+      description: "Acme Labs Context Tree",
+    });
+    expect(createFilePayload).toMatchObject({
+      branch: "main",
+      message: "Initialize Context Tree root node",
+    });
+    const nodeContent = readBase64Content(createFilePayload);
+    expect(nodeContent).toBe(`---
+title: "Acme Labs Context Tree"
+description: "Shared context, decisions, ownership, and operating knowledge for Acme Labs."
+owners: [octocat]
+---
+
+# Acme Labs's Context Tree
+`);
+
+    const setting = await getOrgContextTree(app.db, admin.organizationId);
+    expect(setting).toEqual({ repo: "https://github.com/octocat/acme-labs-context-tree.git", branch: "main" });
+  });
+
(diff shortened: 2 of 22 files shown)


yuezengwu (approved):
Reviewed end-to-end (server endpoint, shared token service, schemas, web UI + tests). LGTM — **approving**. The change aligns with the documented onboarding decision in the context tree (`system/context-management`, 2026-06-09: web frontend creates/configures the repo first, agent builds the first tree version async). No DB/schema change — reuses the existing `org_settings` `context_tree` key, and the Migrations check is green.

### What it does
Adds an admin-only `POST /orgs/:orgId/context-tree/initialize` that, using the signed-in admin's GitHub user token: creates a private repo under their account, writes a minimal valid root `NODE.md`, and persists the org `context_tree` binding. It also extracts the existing `/me/github/repos` token-refresh logic into a shared `getFreshGithubUserToken` service (no behavior change — the 503/403 + `refresh_failed` paths are preserved), and surfaces the initializer in Settings and the `/context` unavailable state with admin/member-specific copy. Good test coverage on auth gating, the 409 already-configured short-circuit, GitHub 422→409 mapping, and token refresh.

### Non-blocking observations (author's discretion)
1. **Partial-failure recovery.** The flow is create-repo → write NODE.md → persist setting, with no rollback. If the NODE.md write (or setting persist) fails after the repo is created, the repo is orphaned and a retry hits the existing repo name → GitHub 422 → 409, so the initializer button is stuck. It's recoverable (the admin settings form still lets them paste the repo URL + save manually), and the error copy is reasonable, but a one-time onboarding action benefits from being idempotent — e.g. on 422, detect/adopt an existing `*-context-tree` repo and proceed to the NODE.md + setting steps rather than hard-failing.
2. **default_branch assumption.** `createUserRepo` reads `default_branch` from the response but the code hardcodes `BRANCH = "main"`. If the admin's GitHub account default branch isn't `main`, the commit lands on `main` while the repo's configured default ref differs (cosmetic on GitHub; First Tree itself reads `branch: "main"` from the persisted setting, so functionally fine). Worth a comment noting the intentional hardcode, or reuse `repo.defaultBranch`.
3. **Ownership model.** The repo is created under the individual admin's personal account (`/user/repos`), so tree access is tied to that person. This matches the PR intent and the tree's onboarding decision, so flagging only for awareness — not a change request.

None of these block. Nice, well-tested change.

baixiaohang (changes requested):
## Review by codex-assistant - PR #923

- Reason: the initializer can report success while binding a private repo that the org-scoped Context Tree snapshot path cannot actually read.

**Level**: B-low
- Path initial: touches `packages/server/**` and `packages/web/**`; no migrations/schema and no listed core chat/session/agent service paths.
- Semantic lift: none.

**PR Summary**
- Problem: admins need a one-click way to bootstrap an org Context Tree repo from Settings or the `/context` unavailable state.
- Design: admin-only org route creates a private GitHub repo with the signed-in user's GitHub token, writes root `NODE.md`, persists `context_tree`, and adds Web UI entry points.
- Affected modules: server org Context Tree API, GitHub token helpers, shared schemas, Web Context/Settings pages.

**Review Findings**
❌ 1. The initializer binds a repo that the snapshot service may not be able to read. [R1/R4, `packages/server/src/api/orgs/context-tree.ts:62`]

The new route creates `repo.cloneUrl` with the signed-in user's GitHub user token and immediately persists it as the org Context Tree binding. But the org snapshot route authenticates GitHub-hosted tree fetches by looking up the *First Tree org's bound GitHub App installation* and minting that installation token before cloning (`packages/server/src/api/orgs/context-tree-snapshot.ts:45`). Those are different authorities. If the org has no bound installation, the installation is on a different GitHub account/org, or the App was installed with selected repositories that do not include this newly-created personal repo, `/context` remains unavailable after a successful 201. On top of that, GitHub's `POST /user/repos` private-repo path requires OAuth/PAT `repo` scope, while the App user-token flow intentionally does not request OAuth scopes (`packages/server/src/services/github-app.ts:463`), so the real App-token path can 403 even though the tests mock a repo-scoped token.

Please make initialization use the same authority that will later read the tree, or explicitly verify/adopt the repo into that authority before saving the binding. Concretely, require/choose the org-bound installation account, ensure the new repo is accessible to that installation token (including selected-repo installs), and cover the no-installation / wrong-account / selected-repo cases in tests. Otherwise the UI can move the team from "not configured" to "configured but unreadable" with no successful first-use path.

⚠️ 2. Partial failures after repo creation leave a retry-hostile orphan. [R4, `packages/server/src/api/orgs/context-tree.ts:75`]

The flow is create repo -> write `NODE.md` -> save `context_tree`. If the file write or DB save fails after repo creation, the team remains unconfigured; the next click tries to create the same deterministic repo name and maps GitHub 422 to 409 (`packages/server/src/api/orgs/context-tree.ts:129`), so the initializer cannot recover itself. The manual Settings form can still be used, but the one-click initializer should either compensate, adopt the existing just-created repo, or persist enough state to resume. This is especially important because the external GitHub side effect cannot be part of the DB transaction.

✅ 3. The route is correctly org-scoped and admin-gated, and the Web UI keeps member copy/read-only behavior aligned with the existing settings model.

**Executed**
- Did not run tests, per PR-review rules.

yuezengwu (changes requested):
**Retracting my earlier approval — I agree with codex-assistant's Finding 1, which I missed.** Submitting request-changes.

I cross-checked the read-vs-write authority in the code and the concern is real:

- **Read authority (snapshot):** `packages/server/src/api/orgs/context-tree-snapshot.ts:43-47` mints an **org-bound GitHub App installation token** — `findInstallationByOrg(org)` → `mintContextTreeInstallationToken(installation, ...)`. When there is no bound installation (`mintResult.ok` is false), `githubToken` is `undefined` and the private clone in `getContextTreeSnapshot` cannot read the repo.
- **Write authority (initializer):** `packages/server/src/api/orgs/context-tree.ts` creates the repo via `POST /user/repos` under the **signed-in admin's personal account** with their user token, then persists `repo.cloneUrl` as the binding.

These are different principals. A successful `201` does not guarantee `/context` becomes readable. Concrete failure paths that the current flow does not guard:
1. **No org-bound App installation** → snapshot mints no token → private repo unreadable.
2. **Installation on a different account/org** than the admin's personal account → the personal repo isn't covered by the installation.
3. **"Selected repositories" install** → a brand-new repo is not auto-added to the installation, so it stays unreadable until someone edits the install's repo selection.

(One refinement to codex's wording: GitHub Apps gate `/user/repos` via the App's granular **Administration** permission on the account rather than a classic OAuth `repo` scope — but the conclusion holds: the real App user-token path can also be permission-denied where the mocked test token succeeds.)

This ties back to the team's onboarding decision in the context tree (`system/context-management`, 2026-06-09): the web frontend should create/configure the repo **and complete the GitHub init config** so the tree is usable. Creating the repo under an authority the sync path can't read leaves that contract half-done.

**Asks (blocker):** initialize under, or verify/adopt the repo into, the *same* authority that later reads it — i.e. require/choose the org-bound installation account, ensure the new repo is reachable by the installation token (including selected-repo installs), and add tests for the no-installation / wrong-account / selected-repo cases. Finding 2 (partial-failure orphan, non-idempotent retry) still stands as a should-fix.

Finding 3 (org-scoped + admin-gated route, member read-only UI) — agree, correctly done.

yuezengwu (approved):
**Review: Context Tree GitHub repo initializer**

**Goal & core changes (as I read them):** add an admin-only `POST /api/v1/orgs/:orgId/context-tree/initialize` that, using the team's **per-org GitHub App installation token** (not the admin's personal token), creates a private org repo `<slug>-context-tree`, writes a skeleton root `NODE.md`, and persists the `context_tree` org setting. The flow is idempotent on re-run (adopt-on-422, skip rewrite when `NODE.md` exists, save binding only at the end) and the UI surfaces a "Create private GitHub repo" CTA for admins / "ask an admin" copy for members in both Settings and the `/context` unavailable state. The installation-token authority check (the repo must be readable by the bound installation before it is persisted) is the right call and matches the snapshot-mirror per-org-token model.

No correctness blockers found. Verified the partial-failure paths (root-node write fails after repo create; DB save fails after write) are covered by tests and converge on retry. **Approving** — the items below are non-blocking.

**Finding 1 — dead code (cleanup, non-blocking).** `createUserRepo` and `createRepoFile` in `services/github-oauth.ts` are exported but never called anywhere in the PR; the shipped path uses the installation-token helpers in `github-app.ts` (`createOrganizationRepo` / `createRepoFileWithToken`). These look like residue of the earlier "create repo under the signed-in user" approach (still referenced in the PR description). Recommend deleting them so we don't ship a second, unwired GitHub-write surface. Relatedly, `GithubUserToken.login` / `githubId` returned by `getFreshGithubUserToken` are unused by its only consumer (`me.ts` uses only `accessToken`).

**Finding 2 — 422 handling breadth (minor).** `createOrAdoptContextTreeRepo` treats any `422` from repo-create as "repo already exists → adopt". GitHub also returns `422` for other validation failures (e.g. an invalid/reserved name); in that case the follow-up `getRepository` 404s and the caller sees `repo_unavailable` ("exists but not accessible"), which would be misleading. The slug logic makes invalid names unlikely, so this is low-priority — optionally narrow on the upstream message or surface a distinct code.

**Note — PR description drift:** the summary still says "create a private GitHub repo under the signed-in user", but the implementation creates under the org installation account. Worth updating to avoid confusion.

CI at review time: Lint & Type Check, CodeQL, Migrations green; `Test` still running (did not gate on it).

baixiaohang (changes requested):
## Review by codex-assistant - PR #923

- Reason: the read/write authority blocker from the previous review is fixed, but the new GitHub App permission requirement is not reflected in the app permission contract/dev install path, so the initializer can still be unusable in current installs.

**Level**: B-low
- Path initial: touches `packages/server/**` and `packages/web/**`; no migrations/schema and no listed core chat/session/agent services.
- Semantic lift: none.

**PR Summary**
- Problem: let admins initialize a missing Context Tree repo from the web UI.
- Design: create/adopt a deterministic private GitHub org repo through the bound GitHub App installation token, initialize `NODE.md`, and persist the org `context_tree` setting.
- Affected modules: server org context-tree routes, GitHub App token/repo helpers, shared schema, Settings and `/context` UI.

**Review Comments**
❌ 1. Update the GitHub App permission contract for the new hard `administration: write` requirement. [R5 / `packages/server/src/api/orgs/context-tree.ts:57`, `packages/server/src/api/auth/github.ts:267`]

The route now refuses to initialize unless the minted installation token includes both `administration: write` and `contents: write`. That is the right authority model for creating an org repo, but the rest of the App contract still describes/seeds the old permission set: the dev callback explicitly says its stub mirrors the GitHub-side App settings, then inserts only `contents`, `pull_requests`, `issues`, `metadata`, and `members`. The existing dev-callback test also still pins the old contract. In a local/dev install created through that path, and in any deployed installation whose App settings have not been upgraded and re-approved, this endpoint will deterministically return `installation_permissions_insufficient` rather than initialize anything.

Please update the source-of-truth permission surface together with this code change: add `administration: write` to the dev install stub/tests and document the required GitHub App settings/rollout so operators and existing installations approve the new permission before relying on the initializer.

✅ 2. The previous authority mismatch is addressed: repo creation/adoption, `NODE.md` verification/write, and the persisted binding are now all guarded by the same org-bound installation token, with coverage for no installation, selected-repo installs, unreadable existing repos, and retry after partial failure.

**Executed**
- Submitted request changes.

yuezengwu (changes requested):
**Correcting my earlier approval → aligning with @codex-assistant's request_changes.**

I independently verified codex's Finding 1 and it is a real, blocking gap — I missed it on the first pass.

The initialize route hard-requires `administration: "write" && contents: "write"` (`hasInitializationPermissions`, `api/orgs/context-tree.ts`). But the permission **source of truth** does not yet declare `administration`:

- The dev-install stub in `api/auth/github.ts` seeds `{ contents: write, pull_requests: write, issues: read, metadata: read, members: read }` — no `administration`. Its own comment says this block mirrors the GitHub-side App settings (D0b).
- `__tests__/oauth-flow.test.ts:402` pins that contract (`{ contents: "write", members: "read" }`); `administration` appears nowhere outside the new initializer test.

Consequence: every dev/QA install via that callback, and any deployed installation whose App settings haven't been upgraded **and re-approved**, will deterministically hit `installation_permissions_insufficient` and initialize nothing. The feature ships unusable on current installs.

Please land the permission-surface change together with this code:
1. Add `administration: write` to the dev-install stub and update the pinned `oauth-flow` test.
2. Update the GitHub App permission contract / settings declaration (the D0b source of truth) so the App actually requests `administration: write`.
3. Add a rollout note so operators/existing installs re-approve the new permission before relying on the initializer.

My two earlier non-blocking items still stand (dead `createUserRepo`/`createRepoFile` in `github-oauth.ts`; broad `422`→adopt handling). Internal route correctness, idempotency, and the read/write authority model remain good — this is specifically about making the required permission satisfiable.

baixiaohang (approved):
## Review by codex-assistant - PR #923 follow-up

**Recommendation**: approve
- Reason: the remaining blocker is fixed: the hard GitHub App `administration: write` requirement is now reflected in the dev install path, the pinned dev-callback test, and the D0b permission/rollout documentation.

**Level**: B-low
- Path initial: touches `packages/server/**`, `packages/web/**`, and docs; no migrations/schema and no listed core chat/session/agent services.
- Semantic lift: none.

**PR Summary**
- Problem: let admins initialize a missing Context Tree repo from the web UI.
- Design: create/adopt a deterministic private GitHub org repo through the bound GitHub App installation token, initialize `NODE.md`, and persist the org `context_tree` setting.
- Affected modules: server org context-tree routes, GitHub App token/repo helpers, shared schema, Settings and `/context` UI, GitHub App setup docs.

**Review Comments**
✅ 1. The original read/write authority blocker remains fixed. Repo creation/adoption, root-node verification/write, and the persisted binding are all guarded by the same org-bound installation token, with coverage for no installation, selected-repo installs, unreadable existing repos, and partial-failure retry.

✅ 2. The permission-contract blocker is fixed. The dev callback now seeds `administration: "write"`, the OAuth/dev-flow test pins that permission, and `docs/github-app-design-zh.md` documents the App settings plus the operator rollout requirement that existing installs re-approve the permission upgrade before relying on one-click initialization.

⚠️ 3. Non-blocking cleanup still worth doing before merge: the PR description still says the repo is created under the signed-in user, but the implementation now creates under the org installation account. Also consider removing the now-unused `createUserRepo` / `createRepoFile` exports from `services/github-oauth.ts` so the old user-token write surface does not linger.

**Executed**
- Submitted approve review.
- Did not run local tests/checks per review-mode rules; CI was still finishing at review time.

yuezengwu (approved):
## Re-review after `synchronize` (commits 9e20b46f, 1b01c238) — **approving**

Both previously-flagged blockers are independently verified as resolved against the current head, and all CI is green (Test passed). This re-review supersedes my earlier CHANGES_REQUESTED, which was against the pre-fix code.

### Goal & core change
Admin-only `POST /orgs/:orgId/context-tree/initialize` that bootstraps a missing org Context Tree: create/adopt a deterministic private repo, write root `NODE.md`, persist the `context_tree` org setting. The substantive rework since first submission: **initialization now uses the same authority as the reader.**

### Blocker 1 — read/write authority mismatch — ✅ fixed (verified)
The initializer now mints the **org-bound GitHub App installation token** (`mintContextTreeInstallationToken(findInstallationByOrg(...))`) — the exact same call the snapshot reader uses (`context-tree-snapshot.ts:46-47`). Repo create (`POST /orgs/{org}/repos`), NODE.md verify/write, and the persisted binding are all guarded by that one token. Pre-conditions are enforced before any side effect: Organization install, `repositorySelection === "all"`, and `administration:write` + `contents:write`. The "configured but unreadable" path is closed, and there's explicit coverage for it ("does not fall back to the signed-in user's GitHub token when the installation is bound elsewhere", plus no-installation / suspended / selected-repo / insufficient-permission cases).

### Blocker 2 — partial-failure orphan — ✅ fixed (verified)
Flow is now idempotent: create-or-adopt (422 → adopt existing org repo) → `ensureRootNode` (checks existence first, skips if present) → persist setting. Retry after a failed NODE.md write or DB save now recovers cleanly — covered by "can retry successfully when the root-node write failed after repo creation" and "...when DB save failed...".

### Latest commit `1b01c238` (was unreviewed)
Adds `administration:write` to the App permission contract (required to create org repos), pins it in the dev-callback test, and documents the operator rollout requirement that **existing installations must re-approve the permission upgrade** before one-click init works. Correct and necessary.

No DB/schema change — reuses the existing `org_settings.context_tree` key; Migrations check green.

### Non-blocking (author discretion, won't re-review)
1. PR description still says the repo is created "under the signed-in user" — implementation now creates under the org installation account. Worth updating to avoid confusing future readers.
2. `createUserRepo` / `createRepoFile` in `services/github-oauth.ts` now have zero call sites — safe to delete the dead user-token write surface.

Nice, thorough, well-tested change.

PR #924, opened 2026-06-10
Allow tree verify on W1 clean checkouts

baixiaohang:
## Summary

- Treat pre-W1 `.first-tree/VERSION` and `.first-tree/tree.json` as optional legacy runtime metadata in `tree verify`.
- Keep the existing source-repo misuse guard intact.
- Add regression coverage for a clean W1 Context Tree checkout with no committed runtime metadata.

## Why

`first-tree-context` removed tracked W1 residue from the tree repo, but the published verifier still failed fresh CI checkouts because it required `.first-tree/VERSION` and tree identity files that are no longer durable tree content. The verifier should validate the tree structure itself; workspace/tree binding state now lives outside the tree repo.

## Verification

- `pnpm --filter first-tree-dev test -- src/__tests__/tree-verify-workspace-sync.test.ts`
  - This ran the CLI package test suite in this worktree: 58 test files / 428 tests passed.
- `node apps/cli/dist/cli/index.mjs tree verify --tree-path /home/bai/.first-tree-staging/data/workspaces/codex-architect/first-tree-context`
  - Passed against a clean `first-tree-context` checkout that only has `.first-tree/hooks/pre-push` and no `.first-tree/VERSION` / `.first-tree/tree.json`.

### DIFF
--- a/apps/cli/src/__tests__/tree-verify-workspace-sync.test.ts
+++ b/apps/cli/src/__tests__/tree-verify-workspace-sync.test.ts
@@ -95,6 +95,27 @@ function writeValidTree(root: string): void {
   });
 }
 
+function writeValidW1Tree(root: string): void {
+  writeFileSync(join(root, ".git"), "gitdir: /tmp/tree\n");
+  writeFileSync(join(root, "NODE.md"), ["---", "title: Root", "owners: [team]", "---", "", "# Root", ""].join("\n"));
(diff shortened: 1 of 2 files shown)


baixiaohang:
Closing in favor of #925, which carries the same commit on a branch name that satisfies the CI branch-name convention.

PR #925, opened 2026-06-10
Allow tree verify on W1 clean checkouts

baixiaohang:
## Summary

- Treat pre-W1 `.first-tree/VERSION` and `.first-tree/tree.json` as optional legacy runtime metadata in `tree verify`.
- Keep the existing source-repo misuse guard intact.
- Add regression coverage for a clean W1 Context Tree checkout with no committed runtime metadata.

## Why

`first-tree-context` removed tracked W1 residue from the tree repo, but the published verifier still failed fresh CI checkouts because it required `.first-tree/VERSION` and tree identity files that are no longer durable tree content. The verifier should validate the tree structure itself; workspace/tree binding state now lives outside the tree repo.

## Verification

- `pnpm --filter first-tree-dev test -- src/__tests__/tree-verify-workspace-sync.test.ts`
  - This ran the CLI package test suite in this worktree: 58 test files / 428 tests passed.
- `node apps/cli/dist/cli/index.mjs tree verify --tree-path /home/bai/.first-tree-staging/data/workspaces/codex-architect/first-tree-context`
  - Passed against a clean `first-tree-context` checkout that only has `.first-tree/hooks/pre-push` and no `.first-tree/VERSION` / `.first-tree/tree.json`.

### DIFF
--- a/apps/cli/src/__tests__/tree-verify-workspace-sync.test.ts
+++ b/apps/cli/src/__tests__/tree-verify-workspace-sync.test.ts
@@ -95,6 +95,27 @@ function writeValidTree(root: string): void {
   });
 }
 
+function writeValidW1Tree(root: string): void {
+  writeFileSync(join(root, ".git"), "gitdir: /tmp/tree\n");
+  writeFileSync(join(root, "NODE.md"), ["---", "title: Root", "owners: [team]", "---", "", "# Root", ""].join("\n"));
+  mkdirSync(join(root, "members", "gandy"), { recursive: true });
+  writeFileSync(
+    join(root, "members", "gandy", "NODE.md"),
+    [
+      "---",
+      "title: Gandy",
+      "owners: [gandy]",
+      "type: human",
+      "role: Engineer",
+      "domains: [platform]",
+      "---",
+      "",
+      "# Gandy",
+      "",
+    ].join("\n"),
+  );
+}
+
 function makeGitRepo(root: string): void {
   mkdirSync(root, { recursive: true });
   writeFileSync(join(root, ".git"), "gitdir: /tmp/mock\n");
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Approving — reviewed the commit (`9026505`, identical to the closed #924).

**Goal:** make `tree verify` pass on a fresh W1 tree-repo checkout. After `first-tree-context` stopped committing `.first-tree/VERSION` and `.first-tree/tree.json` (W1 moved workspace/tree identity into the workspace-root manifest), the published verifier still hard-failed CI on their absence. This drops those two requirements and keeps the verifier focused on tree *structure*.

**Verified the core concern — the source-repo misuse guard still works.** The guard now hinges entirely on `readSourceBindingContract`, not on tree-identity presence:
- Source repo → has the `FIRST-TREE-SOURCE-INTEGRATION` managed block in AGENTS.md/CLAUDE.md → `readSourceBindingContract !== undefined` + no tree identity → still throws the helpful "verify the tree repo instead" error. The existing guard test covers this.
- Clean W1 tree checkout → no source-integration block → guard not triggered → validates. The removed VERSION/tree.json files were never the guard's signal, so dropping them doesn't weaken it.

**Other checks:**
- `frameworkVersion`/`treeState` rows hardcoded to `ok: true` (kept as compat rows for parsers) — clean.
- Removed imports (`existsSync`, `TREE_VERSION_FILE`) are fully unreferenced; `readTreeIdentityContract` is still imported and used by the guard. No dangling imports → Lint & Type Check green.
- Tests: new W1-clean-checkout case + the failing-case test correctly updated to expect PASS on framework/tree-state while still failing on owners/progress. Guard test intact.

No core data-structure or DB changes — implementation-only CLI catch-up to an already-made W1 decision. All CI green (Test, Lint & Type Check, CodeQL, Branch Name). LGTM.

yuezengwu (approved):
## Approve ✅

Reviewed the diff against the recorded tree decision in `system/context-management/workspace-layout.md` — the change is correct and aligned.

**What this does:** Makes `tree verify` pass on a clean W1 tree-repo checkout. Post-W1, binding/runtime metadata (`.first-tree/VERSION`, `.first-tree/tree.json`) is no longer durable tree content — it lives in the workspace-root `workspace.json` outside the tree repo — so requiring those files made fresh CI checkouts false-fail. The fix hardcodes the `frameworkVersion` and `treeState` rows to `ok: true` while leaving the real structural validation (nodes, members, root frontmatter, progress) to gate the result.

**Verified:**
- Source-repo misuse guard is intact — `verifyTreeRoot` still throws via `readTreeIdentityContract` + `readSourceBindingContract`, and its test (`explains when verify is run from a source integration…`) is untouched. A source repo still gets the "verify the tree repo instead" error.
- Removed imports (`existsSync`, `TREE_VERSION_FILE`) were used only in the two deleted checks → no dangling references / unused-import lint risk. `readTreeIdentityContract` stays imported because the guard still needs it.
- New regression test (`accepts a W1 clean tree checkout…`) and the updated negative-test expectations are consistent with the new always-PASS rows.
- Matches the tree's stated contract: `tree verify` validates tree-internal structure and runs in tree-repo CI, not at the workspace boundary.

**Minor (non-blocking):** `frameworkVersion` / `treeState` are now constant-`true` rows that can never fail and print `[PASS]` unconditionally — kept (per the code comment) as compatibility signals for existing JSON-output parsers. Reasonable for schema back-compat, but they no longer validate anything; consider a follow-up to drop or repurpose them if no parser actually depends on the keys.

No core data-structure or DB change. Merge is correctly gated by branch protection — let CI (Test / Lint) go green before landing.

PR #932, opened 2026-06-10
feat(onboarding): seed Context Tree from source repos at kickoff

Gandy2025:
## What

After onboarding's "create agent" step, the **new-tree** kickoff path now
provisions a Context Tree (Cloud) and the cloud agent **seeds it from the bound
source repos** with `first-tree-seed` — asynchronously, in its kickoff chat,
without blocking onboarding. Completes the "build context tree during
onboarding" flow that was previously stubbed out.

Builds on #923 (merged). Design + the verified gaps it closes, plus open
prerequisites:
[`docs/development/onboarding-context-tree-seed-design.md`](docs/development/onboarding-context-tree-seed-design.md).

## Changes

**Web (onboarding)**
- `steps/step-kickoff.tsx` + new `provision-tree.ts` — new-tree mode provisions
  the tree before sending the kickoff message; a `409` is resolved by the
  **actual binding state** (already-provisioned → proceed; no-tree-created →
  re-throw the actionable error), not blindly swallowed/re-thrown.
- `bootstrap-prose.ts` — new-tree message names `first-tree-seed` and drops the
  self-provision / record-URL framing; the bind message no longer self-binds
  (the runtime writes `workspace.json`) and reads/reflects via
  `first-tree-context`.

**Client (runtime)**
- `workspace-manifest.ts` (new) + `agent-bootstrap.ts` — make the cloud agent
  home a valid **W1 workspace**: sibling symlink to the shared tree clone +
  `.first-tree/workspace.json`, so the shipped skills (incl. `first-tree-seed`'s
  self-check) stay layout-agnostic. Validates before touching disk, tolerates
  concurrent-session races, never clobbers checked-out code, never throws out.
- `context-tree-rebind.ts` (new) + `session-manager.ts` — lazily re-resolve a
  tree-LESS binding at session start (rate-limited) so a tree provisioned *after*
  the slot started is picked up without a daemon restart.
- `agent-bootstrap.ts` — only pin the bundled CLI version when integration
  actually ran, so the tree-less → tree-bound upgrade installs the skills.

Source repos already reach the agent via `recommended` team resources — no
`gitRepos` wiring needed.

## Prerequisites / known limits (design §O3)
- New-tree provisioning needs the GitHub App **org installation** with
  `Administration: write` + `Contents: write` on **all** repos. Personal-account
  installs get `409 organization_installation_required` (surfaced as an
  actionable error) and cannot use one-click new-tree.

## Testing
[...]

PR #933, opened 2026-06-10
feat(onboarding): seed Context Tree from source repos at kickoff

Gandy2025:
## What

After onboarding's "create agent" step, the **new-tree** kickoff path now
provisions a Context Tree (Cloud) and the cloud agent **seeds it from the bound
source repos** with `first-tree-seed` — asynchronously, in its kickoff chat,
without blocking onboarding. Completes the "build context tree during
onboarding" flow that was previously stubbed out.

Builds on #923 (merged). Design + the verified gaps it closes, plus open
prerequisites:
[`docs/development/onboarding-context-tree-seed-design.md`](docs/development/onboarding-context-tree-seed-design.md).

## Changes

**Web (onboarding)**
- `steps/step-kickoff.tsx` + new `provision-tree.ts` — new-tree mode provisions
  the tree before sending the kickoff message; a `409` is resolved by the
  **actual binding state** (already-provisioned → proceed; no-tree-created →
  re-throw the actionable error), not blindly swallowed/re-thrown.
- `bootstrap-prose.ts` — new-tree message names `first-tree-seed` and drops the
  self-provision / record-URL framing; the bind message no longer self-binds
  (the runtime writes `workspace.json`) and reads/reflects via
  `first-tree-context`.

**Client (runtime)**
- `workspace-manifest.ts` (new) + `agent-bootstrap.ts` — make the cloud agent
  home a valid **W1 workspace**: sibling symlink to the shared tree clone +
  `.first-tree/workspace.json`, so the shipped skills (incl. `first-tree-seed`'s
  self-check) stay layout-agnostic. Validates before touching disk, tolerates
  concurrent-session races, never clobbers checked-out code, never throws out.
- `context-tree-rebind.ts` (new) + `session-manager.ts` — lazily re-resolve a
  tree-LESS binding at session start (rate-limited) so a tree provisioned *after*
  the slot started is picked up without a daemon restart.
- `agent-bootstrap.ts` — only pin the bundled CLI version when integration
  actually ran, so the tree-less → tree-bound upgrade installs the skills.

Source repos already reach the agent via `recommended` team resources — no
`gitRepos` wiring needed.

## Prerequisites / known limits (design §O3)
- New-tree provisioning needs the GitHub App **org installation** with
  `Administration: write` + `Contents: write` on **all** repos. Personal-account
  installs get `409 organization_installation_required` (surfaced as an
  actionable error) and cannot use one-click new-tree.

## Testing
- `pnpm typecheck` (11/11 packages) · `biome check` clean
- client `vitest` green; web `vitest` green (1 pre-existing unrelated
  `localhost:3000` file artifact)
- New unit/regression tests: prose, `provisionNewTree`, `ensureWorkspaceManifest`,
  `reresolveUnboundTree`, SessionManager lazy-binding, the CLI-pin upgrade.
- 3 independent review rounds; all findings fixed. Onboarding kickoff UI verified
  via the DEV preview gallery (renders clean, no console errors).

### DIFF
--- a/docs/development/onboarding-context-tree-seed-design.md
+++ b/docs/development/onboarding-context-tree-seed-design.md
@@ -1,0 +1,246 @@
+# Onboarding → build Context Tree from source repos (design)
+
+Status: proposed · Date: 2026-06-10 · Depends on: PR #923 (`feat/context-tree-initializer`), enabled-by PR #925 (W1 tree verify on clean checkouts)
+
+## 1. Goal
+
+After a user creates their first agent during onboarding, the **new-tree**
+path should actually result in a Context Tree that is **seeded from the bound
+source repos** by that agent, using the `first-tree-seed` skill — asynchronously,
+without blocking the agent from handling other chats, and as a natural part of
+onboarding (the user does not babysit it).
+
+## 2. Current state (verified in code)
+
+The onboarding kickoff already *intends* this, but the new-tree path is broken
+end to end. Three concrete gaps:
+
+- **断点1 — WITHDRAWN after implementation review.** The original concern
+  ("`runKickoff` drops `gitRepoUrls`, so sources never materialize") was based on
+  the stale "gitRepos lives in agent config" model. In the current code: (a) the
+  config PATCH endpoint **rejects** writing `gitRepos`
+  (`config-service.ts` → `legacy_resource_config_disabled`); (b) the runtime
+  `gitRepos` is **derived** from the agent's *effective resources*
+  (`resources.ts` `resolveRuntimeConfig`), and a `recommended` **team** repo
+  resource is `mode:"enabled"` for every org runtime agent by default; (c)
+  `runKickoff` **already** creates the admin's selected repos as `recommended`
+  team resources (the `orgWrites` block). So the admin new-tree agent's sources
+  already flow → `prepareSourceRepos` materializes them, and creating those
+  resources bumps the config version so the kickoff message refreshes the client
+  config. **No gitRepos wiring is needed.** Residual gap: the invitee *picker*
+  sub-state (invitee's own repos, never turned into resources) — an edge path,
+  unrelated to new-tree seed; **deferred**, noted in the PR.
+
+- **断点2 — on the new-tree path the seed skill is neither installed, surfaced,
+  nor allowed.** The First-Tree family skills (incl. `first-tree-seed`) install
+  and appear in the briefing **only when `contextTreePath !== null`**
+  (`agent-briefing.ts:503`, `agent-bootstrap.ts:187`). A brand-new tree has no
+  `org.context_tree` setting yet → `contextTreePath` is null → the agent is told
+  *"binding a tree is an operator action, surface to a human"*
+  (`agent-briefing.ts:475`). The kickoff prose also names a non-existent
+  "first-tree onboarding skill" (`bootstrap-prose.ts`).
+
+- **断点3 — seed assumes Cloud provisioned first; cloud layout doesn't match
+  W1.** `first-tree-seed` refuses unless `.first-tree/workspace.json`
+  (`{ tree, sources }`) exists, the tree is empty, and all sources are on disk
+  (`skills/first-tree-seed/SKILL.md` self-check). In cloud the tree is cloned to
+  a **separate** dir (`context-tree-repos/<hash>/`, not a workspace sibling),
+  sources live under the agent home, and **no `.first-tree/workspace.json` is
+  ever written** (only the local CLI `migrate-workspace` writes it).
+
+Async note: a single agent already runs up to **5 concurrent sessions**
+(`runtime/config.ts`), per-chat isolated, with "working" state protected up to
+65 min. The seed task runs in the kickoff chat's own session; other chats stay
+responsive. Onboarding itself does not block — kickoff stamps `completed`
+right after sending the message. **No new background-task system is needed.**
+
+## 3. Dependency: PR #923 (Cloud provisioning primitive)
+
+`POST /api/v1/orgs/:orgId/context-tree/initialize` (admin-only) creates a
+private GitHub repo via the **GitHub App ORG installation token**
+(`POST /orgs/{org}/repos` — `createOrganizationRepo`), writes a **title-only
+root `NODE.md`** (deliberately within `first-tree-seed`'s "placeholder root
+counts as empty" carve-out), and persists `org.context_tree = { repo,
+branch:"main" }`. Surfaced today only in Settings and `/context` unavailable
+state — **not** wired into onboarding.
+
+> Correction (verified against the *merged* #923, which changed after the
+> initial review): the repo is created through the **org installation token**,
+> not the signed-in user's OAuth token. The endpoint hard-requires an
+> **Organization** installation with `repositorySelection: "all"` and the App's
+> `Administration: write` + `Contents: write` repository permissions, and
+> returns distinct status codes when those aren't met:
+> - `409 organization_installation_required` — installation is a personal/User
+>   account (so **personal-account users cannot use one-click new-tree** — a
+>   real constraint of the Cloud-provision choice);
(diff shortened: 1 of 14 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #933

- Reason: this PR carries the same head as superseded #932, and the new-tree source-resource precondition blocker is still present.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none; already A

**PR Summary**
- Problem solved: provision a Context Tree during onboarding and have the cloud agent seed it from bound source repos.
- Design: web provisions/binds the tree before kickoff; client lazily re-resolves tree-less slots, installs First Tree skills, and writes a W1 manifest via a tree symlink.
- Affected modules: `packages/web/src/pages/onboarding/**`, `packages/client/src/runtime/**`, client/web tests, onboarding design docs.

**Review Opinions**
❌ 1. `runKickoff` sends the seed prompt after swallowing source-resource creation failures. [R4 / `packages/web/src/pages/onboarding/steps/step-kickoff.tsx:70`]

For new-tree mode, selected repos reach the runtime through recommended team repo resources. Those become the runtime `gitRepos`, then `workspace.json.sources`, then the source set required by `first-tree-seed`. The current code still wraps `createTeamResourceForOrg(...)` in `Promise.allSettled(...)`, ignores rejected writes, and proceeds to create the chat/send the seed prompt. If a selected repo resource fails to write, onboarding can complete with a provisioned tree but no matching bound source repo in the agent config, causing the seed flow to refuse or seed from an incomplete source set. Please make source resource creation required for `treeMode === "new"` before sending the seed kickoff, while treating already-existing duplicate resources as success by checking actual resource state.

**Executed**
- Did not run local checks/tests per review-mode rules.

Gandy2025:
Thanks @baixiaohang — fixed in `9f9661e`.

You're right: for `treeMode === "new"` the selected repos reach the agent **only** via the recommended team repo resources, and they were created best-effort (`Promise.allSettled`, failures swallowed), so a dropped write could leave a provisioned tree with a missing bound source → `first-tree-seed` refuses / seeds incomplete.

New `ensureSourceReposRegistered(orgId, repoUrls)`: creates each resource (tolerating the duplicate-conflict re-run case via canonical matching), then **verifies** every selected repo is actually registered and throws a retryable error if any is missing — so onboarding never sends the seed kickoff with an incomplete source set. Wired in for new-tree mode; the existing-tree path keeps the best-effort cache. Covered by unit tests in `provision-tree.test.ts` (all-registered, duplicate-exists, genuine-failure, canonical match, partial-missing).

baixiaohang (changes requested):
## Review by codex-assistant - PR #933

- Reason: the previous source-resource precondition is now enforced, but the verification uses a second, weaker repo canonicalizer and can reject an already-registered equivalent source repo.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none; already A

**PR Summary**
- Problem solved: provision a Context Tree during onboarding and have the cloud agent seed it from bound source repos.
- Design: web provisions/binds the tree before kickoff; client lazily re-resolves tree-less slots, installs First Tree skills, and writes a W1 manifest via a tree symlink.
- Affected modules: `packages/web/src/pages/onboarding/**`, `packages/client/src/runtime/**`, client/web tests, onboarding design docs.

**Review Opinions**
❌ 1. Source registration verification duplicates the repo canonicalizer and rejects valid existing resources. [R1 / `packages/web/src/pages/onboarding/provision-tree.ts:16`]

The new `ensureSourceReposRegistered` correctly makes selected repos mandatory for `treeMode === "new"`, but its verification does not use the resource model's source of truth. Server resource creation stores `repoCanonicalKey` using shared `canonicalizeResourceRepoUrl`, which treats HTTPS, `ssh://`, and scp-like GitHub URLs as the same repo. This helper instead compares `repoLabel(url).toLowerCase()` from the raw payload URL.

That breaks the duplicate/retry path this fix is trying to tolerate: if a recommended team repo already exists as `ssh://git@github.com/acme/app.git` or `git@github.com:acme/app.git`, creating the selected HTTPS clone URL fails with the expected 409 duplicate, `listTeamResourcesForOrg` returns the existing resource, but `repoCanonical()` produces `ssh://git@github.com/acme/app` or `acme/app` on opposite sides and `missing` still includes the selected repo. The user gets blocked even though the resource is registered and the runtime would receive it.

Please verify against `resource.repoCanonicalKey` and canonicalize the selected URLs with the same shared `canonicalizeResourceRepoUrl` used by the server, then add coverage for existing `ssh://` and scp-like resources matching a selected HTTPS clone URL.

✅ 2. The previous blocker is directionally fixed: new-tree no longer sends the seed prompt after blindly swallowing source-resource write failures; it verifies the selected repos before chat creation.

**Executed**
- Did not run local checks/tests per review-mode rules.

Gandy2025:
Good catch @baixiaohang — fixed in `fe8c80b`.

The verify now uses the shared `canonicalizeResourceRepoUrl` (the same function the server keys `repoCanonicalKey` off) on **both** sides instead of the weaker `repoLabel().toLowerCase()`, so an existing recommended resource registered under an `https://` / `ssh://` / scp `git@…:` form of the same repo matches a selected HTTPS clone URL — no more wrongly-blocked duplicate/retry path. The lookup is guarded so a malformed stored URL just fails to match rather than throwing out of the verify.

Added coverage: existing `ssh://git@github.com/acme/app.git` and `git@github.com:acme/app.git` resources both match a selected `https://github.com/acme/app` (the `ssh://` case was red before this fix).

Gandy2025:
@baixiaohang @yuezengwu — could you take a review pass on the latest commit (`fe8c80b`)? 🙏

Both prior review rounds are addressed:
- **R4** (baixiaohang): new-tree mode now *requires + verifies* source-repo registration before the seed kickoff (`ensureSourceReposRegistered`), instead of best-effort.
- **R1** (baixiaohang): the verify now uses the shared `canonicalizeResourceRepoUrl` (the server's `repoCanonicalKey` source of truth), so https/ssh/scp forms of the same repo match — with added `ssh://`/scp coverage.

CI is green and the PR is mergeable. Appreciate a fresh look when you get a chance.

baixiaohang (approved):
## Review by codex-assistant - PR #933

**Recommendation**: approve
- Reason: the latest head fixes both previously raised blockers: new-tree source repo resources are required and verified before seed kickoff, and the duplicate/retry verification now uses the shared resource repo canonicalizer across HTTPS / SSH / scp URL forms.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none; already A

**PR Summary**
- Problem solved: completes onboarding's new-tree path by provisioning a Cloud Context Tree and having the agent seed it from bound source repos via `first-tree-seed`.
- Design: Web provisions/binds the tree, requires selected source repos to exist as recommended team resources, then sends the seed kickoff; Client lazily upgrades tree-less slots to tree-bound and materializes a W1 workspace manifest plus tree symlink so shipped skills stay layout-agnostic.
- Affected modules: `packages/web/src/pages/onboarding/**`, `packages/web/src/pages/workspace/center/onboarding/**`, `packages/client/src/runtime/**`, client/web tests, onboarding design doc.

**Review Opinions**
✅ 1. The source-resource precondition is now enforced for `treeMode === "new"`; onboarding no longer sends the seed prompt after silently losing selected repos. [R4 / `packages/web/src/pages/onboarding/steps/step-kickoff.tsx:68`]
✅ 2. The repo verification path now uses shared `canonicalizeResourceRepoUrl` and covers existing `ssh://` and scp-form resources matching a selected HTTPS clone URL. [R1 / `packages/web/src/pages/onboarding/provision-tree.ts:1`]
✅ 3. The client runtime changes are cohesive: W1 workspace projection is isolated in `workspace-manifest`, tree-less rebinding is slot/session-manager-owned, and the CLI pin gate preserves the tree-less -> tree-bound skill-install upgrade.

**Executed**
- Did not run tests per review-mode rules.

PR #935, opened 2026-06-10
fix(web): recover doc preview from message metadata when snapshot cache is gone

Gandy2025:
## Problem

The doc-preview drawer rendered **"Document not found"** for a preview that was working moments earlier — typically right after the user sent a new message.

Root cause (two compounding flaws):

1. **Observer-less cache → GC'd after ~5 min.** The seeded snapshot lives in a React Query entry written by `setQueryData` and read by `getQueryData` — neither subscribes, so the entry has zero observers and React Query evicts it after the default `gcTime` (~5 min).
2. **No recovery from message metadata.** Once the seeded entry is gone, any re-render (e.g. the `chat-messages` refetch that `sendMut.onSettled` triggers on send) dropped through to `GET /me/docs/preview`. On the cloud topology the server can't read the agent's filesystem, so it 404s — even though the snapshot bytes are immutable and still sitting in the message's `metadata.documentContext`.

## Fix

Recover the snapshot from the message's own metadata instead of trusting only the transient cache:

- The drawer **reactively observes** the `chat-messages` query (sharing ChatView's exact key), so recovery is a free cache read while the chat is open and a single fetch after a cold reload. A non-reactive `getQueryData` peek would memoise `undefined` on a cold load and never recompute.
- The path endpoint is **deferred only while recovery is in flight** (`!awaitingRecovery`), not gated out wholesale on `docMsg`. This preserves the legacy single-host fallback for in-preview links to docs that were never snapshotted.

## Independent review (codex, 3 passes)

- Pass 1 — flagged the cold-reload hole (non-reactive memo). **Fixed** + regression test.
- Pass 2 — flagged that a blanket `!docMsgId` gate stranded in-preview links to non-snapshotted docs on single-host. **Fixed** (defer-not-gate) + regression test.
- Pass 3 — the pagination tail below. Tracked as fast-follow.

## Known limitation (fast-follow)

A message that has scrolled past the newest 50-message window can't be recovered (the list query only fetches the latest 50) and still falls back to the path endpoint → 404 on cloud. This is narrow (deep-link/bookmark to an aged-out doc) and strictly better than before (previously *every* post-GC case 404'd). The robust fix needs a fetch-single-message-by-id endpoint — filed as a follow-up.

## Tests

- `recovers from the warm messages cache when the seeded entry was GC'd (chat open)`
- `recovers by fetching the messages window after a cold reload`
- `still uses the path endpoint for an in-preview link to a non-snapshotted doc`

708 web tests pass · typecheck clean · biome clean.

### DIFF
--- a/packages/web/src/components/__tests__/doc-preview-drawer.test.tsx
+++ b/packages/web/src/components/__tests__/doc-preview-drawer.test.tsx
@@ -12,8 +12,12 @@ globalThis.IS_REACT_ACT_ENVIRONMENT = true;
 const meDocsMocks = vi.hoisted(() => ({
   getMeDoc: vi.fn(),
 }));
+const chatsMocks = vi.hoisted(() => ({
+  listChatMessages: vi.fn(),
+}));
 
 vi.mock("../../api/me-docs.js", () => meDocsMocks);
+vi.mock("../../api/chats.js", () => chatsMocks);
 vi.mock("../../lib/use-agent-name-map.js", () => ({
   useAgentSlugToIdMap: () => (slug: string | null | undefined) => (slug === "kael" ? "agent-owner" : null),
 }));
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #935

**Recommendation**: approve
- Reason: The drawer now recovers immutable snapshot content from the source message metadata while keeping the legacy path fallback available after recovery settles.

**Level**: B-low
- Path initial: `packages/web/**` -> C
- Semantic lift: touches the Web React Query / `chat-messages` data path -> B-low

**PR Summary**
- Problem: Snapshot-backed doc previews could degrade to `Document not found` after the observer-less seeded cache entry was GC'd and a later re-render hit the path endpoint on cloud.
- Design: `DocPreviewDrawer` first uses the seeded snapshot, then observes the same `chat-messages` query as ChatView to recover from `metadata.documentContext`, and only defers the path endpoint while that recovery is loading.
- Affected modules: `packages/web/src/components/doc-preview-drawer.tsx` and its regression tests.

**Review Notes**
✅ 1. The durable source of preview bytes remains the message metadata; the transient React Query snapshot is treated only as the first, faster path. [R1 / `packages/web/src/components/doc-preview-drawer.tsx`]
✅ 2. The path endpoint is delayed rather than disabled for `docMsg` URLs, so in-preview links to non-snapshotted docs still preserve the single-host fallback. [R4 / `packages/web/src/components/doc-preview-drawer.tsx`]
⚠️ 3. The latest-50 message window limitation is real but already documented as a fast-follow; it is narrower than the current post-GC failure mode and does not block this fix.

**Executed**
- Approved.
- I did not run local tests or checks in review mode; the PR reports web tests, typecheck, and biome passing.

PR #944, opened 2026-06-10
Fix silent backfill context ACK semantics

### LINKED ISSUE #939: Silent backfill context is permanently lost when first-dispatch inbox recovery resets the trigger
## Summary

A newly-invited agent's silent backfill context (the `notify=false` inbox rows written by `addParticipant`, issue #370 / PR #393) never reaches the agent prompt in the most common real-world path: the **first wake in a fresh chat over WS push**. The rows are consumed (bulk-acked) by the first claim, but the client drops that delivery to run chat-scoped inbox recovery; the redelivered frame re-runs preceding-context collection, finds nothing, and the context is gone for good.

This is deterministic, not a race — it hits 100% of the time for exactly the scenario the backfill feature exists to serve.

## Repro (from QA real-agent regression, 2026-06-10)

1. Create a group chat with a human owner + agent A; seed 2 messages containing marker values.
2. Add agent B via `POST /api/v1/chats/:chatId/participants` → backfill creates 2 silent (`notify=false`) inbox rows for B. ✓ (verified in DB)
3. Send a notify-worthy message to B asking it to report the earlier marker values *without* using `chat history`.

Expected: B's prompt contains the `[Earlier in chat — context you missed]` block with both markers.
Actual: B answers with empty values. DB shows the silent rows `status=acked`; daemon log shows `inbox:recover ... resetCount=1` immediately before B's session was created.

## Root cause — two pieces interlock

**Server: silent-row consumption is bound to claim, not ack.**
`collectPrecedingContext` (`packages/server/src/services/inbox.ts`) bundles pending silent rows into the trigger's payload and **bulk-acks them in the same transaction**. If that delivered trigger is later reset back to `pending` (recovery, bind-time reset, daemon restart — `resetDeliveredForInboxes`), the redelivery's context collection finds zero pending silent rows. Silent rows are not reset (they're `acked`, not `delivered`), so the bundle cannot be reconstructed.

**Client: first dispatch in a fresh chat always discards the original frame.**
`SessionManager.dispatch` (`packages/client/src/runtime/session-manager.ts`):

```ts
if (this.shouldRecoverBeforeDispatch(chatId)) {
  await this.recoverChatBeforeDispatch(chatId, entry.id, messageId);
  return; // original frame — carrying precedingMessages — is dropped
}
```

`shouldRecoverBeforeDispatch` returns true whenever no healthy live handler exists for the chat — which is **always** the case for a freshly-invited agent's first wake. So the one delivery that carried the context is discarded by design, and the server-side redelivery arrives with `precedingMessages: []`.

Everything downstream is correct (wire schema carries `precedingMessages`, codex/claude-code/TUI handlers all render the `[Earlier in chat]` block via `formatInboundContent`) — the bundle just never survives to the dispatch that actually runs the turn.

## Why existing coverage missed it

Server tests (`add-participant-backfill.test.ts`) verify claim-time bundling, which is correct in isolation. The loss only appears when claim and the actually-consumed delivery are different deliveries — i.e. with a real daemon doing recovery-before-dispatch. The first real-runtime QA pass over this path is what caught it.

## Fix directions

**A (root fix, server side):** bind silent-row consumption to the trigger's **ack** instead of its claim — at claim time mark silent rows as attached to the trigger entry (e.g. `delivered` + a consuming-entry reference, or equivalent); ack them when the trigger acks; when a delivered trigger is reset to `pending`, release its silent rows back to `pending` too. Covers recovery, bind-time reset, and daemon-restart redelivery uniformly. Needs care for the cap-overflow bulk-ack (rows dropped due to window/cap should still age out and not re-attach forever).

**B (quick mitigation, client side):** after `recoverChatBeforeDispatch` succeeds, continue dispatching the **in-memory original frame** instead of dropping it (dedup against the redelivered copy). Small change, fixes the dominant path, but other reset paths (e.g. daemon restart between claim and turn) still lose context — doesn't replace A.

## Evidence

- QA report: workspaces/qa/qa-reports/group-chat-agent-regression-20260610.md
- Silent rows acked + trigger redelivered: `.qa-local/group-chat-agent-regression-20260610/evidence/codex-invite-backfill-inbox.tsv`
- Recovery proof: daemon.log — `inbox:recover accepted ... resetCount=1` followed by session creation for the failing chat
- Agent turn output `INVITE_BACKFILL_OK marker= fact=` (expected `marker=orchid-610 fact=harbor-42`)

## Relationship to existing issues

- #370 (closed): the original "no history for newly added agent" bug — fixed by adding the backfill write path. This issue is about that mechanism's delivery being lost, not its absence.
- #339 (open): umbrella feature ask for invited-agent context; this bug currently nullifies the shipped mechanism in the primary scenario.
- #912: different inbox-loss bug (idle-suspend wiping unprocessed entries) — adjacent but distinct mechanism.

baixiaohang:
## Summary

- Keep silent inbox rows pending when they are bundled as `precedingMessages`, so chat-scoped recovery can rebuild the same context before the notify turn is ACKed.
- Drain scoped pending silent rows during notify ACK-through for the same inbox/chat up to the acknowledged delivery cursor.
- Partition silent context by the previous notify cursor across delivery batches, preventing a later unacked trigger from reusing silent context that belonged to an earlier trigger.

## Context

- Fixes #939.
- Durable Context Tree update landed first in agent-team-foundation/first-tree-context#472.

## Validation

- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 inbox-ws-push`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 inbox-bind-recovery`
- `pnpm check`
- `pnpm typecheck`

### DIFF
--- a/packages/server/src/__tests__/inbox-bind-recovery.test.ts
+++ b/packages/server/src/__tests__/inbox-bind-recovery.test.ts
@@ -1,7 +1,9 @@
 import { and, eq, inArray } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
 import { inboxEntries } from "../db/schema/inbox-entries.js";
+import { createChat } from "../services/chat.js";
 import * as inboxService from "../services/inbox.js";
+import { sendMessage } from "../services/message.js";
 import { createTestAgent, useTestApp } from "./helpers.js";
 
 /**
@@ -51,8 +53,8 @@ describe("inbox bind-time recovery (resetDeliveredForInboxes)", () => {
     // the third as `pending`, and force-set one row to `acked` to confirm
     // the WHERE filter actually short-circuits on status. Filter to
     // notify=true so the silent fan-out row (from the @a3 message) doesn't
-    // skew the assertion — silent rows are bulk-acked by the trigger's
-    // bundling pass and are not part of the recovery contract.
+    // skew the assertion — silent rows are context-only and are not part of
+    // the bind-time notify recovery contract.
     const rows = await app.db
       .select({ id: inboxEntries.id, messageId: inboxEntries.messageId })
       .from(inboxEntries)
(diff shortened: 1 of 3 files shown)


baixiaohang (commented):
Reviewed the pushed PR diff against the previously reviewed worktree. The R1/R4 silent-context partition issue is fixed: the first claimed trigger now uses the previous same-inbox/chat notify cursor as its silent lower bound, ACK-through drain remains scoped to same inbox/chat/cursor, and ackedEntryIds remains notify-only. No blocking findings. I did not rerun tests as reviewer; relying on the reported local validation plus CI.\n\nNote: this review is submitted as a comment because GitHub rejects approval from the current gh identity as the PR author.

yuezengwu (approved):
Reviewed the full PR head of `services/inbox.ts` plus both test files, and checked the implementation against the merged Context Tree contract (system/cloud/chat/inbox-delivery.md, updated by first-tree-context PR 472). All three contract claims hold in code:

1. **Bundling is not consumption** — `collectPrecedingContext` no longer bulk-acks silent rows at claim; they stay `pending` until the notify entry is ACKed. Verified the chat-scoped recovery path end-to-end: `recoverUnackedForScope` resets only delivered notify rows, and redelivery via `claimBacklogForPushForChat` rebuilds the identical preceding block because the window is now trigger-relative (`trigger.createdAt - 24h`) instead of claim-time `NOW()` — deterministic across redeliveries, which is exactly what issue 939's first-dispatch-recovery loss needed.
2. **Scoped ACK-through drain** — `drainPendingSilentRows` is bounded by same inbox + same `chatPredicate` + `id <= entryId` + `notify=false` + `pending`, so other recipients, other chats, and later unconsumed context are untouched (pinned by the scoping test). Draining also on the `already_acked` branch is right for idempotent re-acks, and `ackedEntryIds` staying notify-only keeps WS in-flight accounting clean.
3. **Cursor partition across batches** — the per-chat `previousNotify` lookup (latest notify row below the first claimed trigger, any status) closes the reuse hole where a later unacked trigger could re-attach an earlier trigger's silent context; cap/window-excluded rows can no longer leak forward either, and they still age out via the existing pending-row GC if no later ack drains them.

Edge cases I walked and am satisfied with: lock ordering is notify-then-silent in both claim and ack transactions (no deadlock inversion); silent rows are never `delivered` so bind-time reset can't touch them; a `failed` previous notify row becomes a partition boundary so its context is narrower than before — acceptable given the tree's "helpful window, not a transcript" framing.

No schema/migration changes; CI is green including the full test job. No blockers — approving.

PR #943, opened 2026-06-10
feat(web): redesign onboarding kickoff finale + repo-connect friction + humanized errors

Gandy2025:
## What

Redesign the onboarding **kickoff** finale into one unified "launch" moment, consistent with the welcome/team opening bookends, and tighten the surrounding repo-connect step.

**Admin**
- Delete the new/existing Context Tree fork (URL input + links). A team's tree is always one we provision; a pre-existing tree (re-run / 2nd admin / CLI-bound) is detected silently and switches the agent's first task to "read" instead of "seed" — no user choice.
- Lead the heading with the agent + outcome, not a Context-Tree gloss. CTA: "Build tree & start".
- No-repo finale: an honest "meet your agent" + a quiet "connect a repo" link back to connect-code; the title no longer duplicates the CTA.

**Invitee**
- Merge the confirm/picker repo-selection screens into a single "ready" launch. The agent already inherits the team's `recommended` repo resources automatically, so there was nothing to pick (the old selection only flavoured the message). `resolveInviteeKickoffState` → waiting / no-installation / ready.
- Send a joining-teammate bootstrap (`buildInviteeBootstrap`: read the tree to get oriented, then introduce yourself) instead of the admin's "reflect these repos into the tree" voice.
- The waiting / no-installation bailouts now enter a real chat instead of dropping to an empty workspace.

**connect-code**
- "Friction, not a block" when 0 repos are picked: a consequence line + a quiet underlined "continue without a repo" link (never the strong primary, never disabled). The strong "Continue" is reserved for when a repo is selected.
- "Start over" moves inline with the "Waiting for GitHub…" status and reads "Didn't work? Start over".

**Errors**
- Map provisioning failure codes (personal-account / missing-permissions / no-installation / repo-unavailable / …) to plain language with a way forward, instead of leaking the raw server string. `ApiError` carries the server `code`; `kickoffErrorMessage` does the mapping.

**Layout**: the kickoff finale matches the welcome/team bookends (top-anchored, left-aligned). `FlowHint`'s icon is boxed to the text line-height so it lines up on single- and multi-line hints.

## Context Tree

This changes a durable onboarding contract (invitee kickoff goes from four confirm/picker states to three `waiting / no-installation / ready`; admin always provisions, no manual tree-URL entry). The matching Context Tree update is **agent-team-foundation/first-tree-context#473** — land it first / together so agents don't keep acting on the old four-state model.

## Reviews

[...]

PR #732, opened 2026-06-01
fix: scope onboarding agent-readiness to the org and land invitees in the right org

Gandy2025:
## Problem

An invited user who clicked an invite link, logged in via GitHub, and had **already onboarded in some other org** landed straight on the workspace — skipping onboarding — and in a **stale org**, not the one they were invited to.

Root cause was two independent issues that fired together:

1. **Onboarding completion was account-level.** The workspace gate keyed off `users.onboardingCompletedAt` (a write-once user column) and a user-scoped agent check (`inferOnboardingStep`: "does this user manage a non-human agent in *any* org"). A returning user joining a brand-new or all-private org was treated as already set up, even with no agent they could use there.
2. **The joined org was never made active.** The invite created the membership server-side, but the client kept whatever org sat in `localStorage.selectedOrganizationId`, so the invitee landed in their previous org.

## Changes

**A. Onboarding "agent readiness" → org/member level (no schema change, no migration)**
- New `listOrgsWithUsableNonHumanAgent` (`services/access-control.ts`): per org, does the member have a usable non-human agent — active AND (`visibility="organization"` OR managed by them)? A private agent owned by another member does **not** count.
- `/me` now returns per-membership `hasUsableAgent` (`api/me.ts`, `shared` `meMembershipSchema`).
- Web gate (`onboarding/steps.ts`) enters onboarding when the user hasn't connected a client **or** the selected org has no usable agent. `onboardingCompletedAt` is kept for the Settings entry-point but is no longer the workspace gate.

**B. Land invitees in the right org**
- OAuth callback carries the resolved org back in the redirect fragment (`org=`, `api/auth/github.ts`); `oauth-complete.tsx` calls `selectOrganization(org)` before navigating.
- Authenticated invite Join path (`invite-accept.tsx`) selects `res.organizationId`.

**C. Drive-by fix**
- The invite page "switch from X to Y" warning read a non-existent `/me` field and never showed; now reads `teamDisplayName` from the auth context (drops two redundant requests too).

## Intentional behavior changes (signed off)
- Multi-org users who land in an org where they have no usable agent are now routed to create one there (previously skipped account-wide).
- An admin who created an agent but never finished the kickoff step is no longer auto-resumed into the wizard from the workspace root (kickoff is still reachable from Settings).
[...]

PR #948, opened 2026-06-10
fix(web): don't eject onboarding users when their org gains a usable agent mid-flow

Gandy2025:
## Problem

Going through `/onboarding`, the moment a user creates their agent (the **create-agent** step) they are bounced straight into the workspace — **skipping connect-code + kickoff** (admin) / **kickoff** (invitee). Found during a manual E2E of the kickoff redesign (#943); it is **pre-existing**, not introduced by that PR.

## Root cause

`OnboardingPage` re-evaluates `shouldLeaveOnboarding` **on every render**, and that returns true as soon as the org has a usable non-human agent (`currentOrgHasUsableAgent`, from `me.ts` / `listOrgsWithUsableNonHumanAgent` — active + org-visible, **no online requirement**). Creating the agent makes the org "ready", and the flow's own `onAgentOnline` → `refreshMe()` surfaces it immediately, so the route fires `<Navigate to="/">`.

Both paths have steps **after** create-agent (admin: `connect-code` + `kickoff`; invitee: `kickoff`), so they become unreachable via the forward flow. This also left `inferInitialStepIndex`'s "completed → resume at connect-code" branch effectively dead. It reproduces whenever a real runtime is connected (i.e. real users); pure-mock environments don't bring the agent online, which is likely why it slipped past prior reviews/sims.

## Fix

Decide the leave-bounce **once, at entry** (first `/me` load) via a ref, instead of re-checking every render. `shouldLeaveOnboarding` is an entry-time guard ("you landed here already done → go back"), not a live leash. An actively-onboarding user now finishes the flow and leaves via the explicit `completeAndEnterChat` navigate (or `finishLater`); a genuinely-done user who lands on `/onboarding` is still bounced.

`shouldLeaveOnboarding` itself is unchanged — it stays correct as the entry check, and **#732's per-org enter behavior is preserved**. (Deliberately did *not* switch the leave check to `onboardingCompletedAt`: that flag is account-level, so it would reintroduce the multi-org bug #732 fixed.)

## Tests

- New regression test in `onboarding-page-dom.test.tsx`: org gains a usable agent **mid-flow** → no eject, stays on connect-code (red before the fix, green after).
- Existing "terminally completed user landing on /onboarding → bounce" test still passes (entry-time guard intact).
- Full web suite green: **726/726**; typecheck + `lint:tokens` + biome clean.

## Known limitation (not addressed here)

A user who abandons mid-flow (already has an agent, never finished kickoff) and later re-opens `/onboarding` is still bounced rather than resumed at connect-code. Supporting resume-after-abandon needs a real **per-org completion** signal (server change) — out of scope for this minimal fix.

### DIFF
--- a/packages/web/src/pages/onboarding/__tests__/onboarding-page-dom.test.tsx
+++ b/packages/web/src/pages/onboarding/__tests__/onboarding-page-dom.test.tsx
@@ -130,6 +130,43 @@ describe("OnboardingPage", () => {
     expect(container.textContent).toContain("Workspace Home");
   });
 
+  it("does NOT eject a user whose org gains a usable agent mid-flow (created at create-agent)", async () => {
+    // Entry: actively onboarding, no usable agent yet → on the create-agent step.
+    authMock.value = { ...authMock.value, onboardingStep: "create_agent", currentOrgHasUsableAgent: false };
+    flowMock.activeStep = "create-agent";
+
+    // Fresh element each render so React actually reconciles on the flip
+    // (re-rendering the same element object bails out). OnboardingPage stays
+    // the same instance across renders, so its entry-time decision persists.
+    const renderTree = () => (
+      <MemoryRouter initialEntries={["/onboarding"]}>
+        <Routes>
+          <Route path="/" element={<div>Workspace Home</div>} />
+          <Route path="/onboarding" element={<OnboardingPage />} />
+        </Routes>
+      </MemoryRouter>
+    );
+    const container = document.createElement("div");
+    document.body.appendChild(container);
+    root = createRoot(container);
+    await act(async () => root?.render(renderTree()));
+    await flush();
+    expect(container.textContent).toContain("Create Agent Step");
+
+    // Mid-flow: the just-created agent comes online → currentOrgHasUsableAgent
+    // flips true and the flow advances to connect-code. The route must NOT
+    // bounce to the workspace — connect-code + kickoff are still ahead of
+    // create-agent in the admin sequence. (Regression: the leave-gate used to
+    // re-evaluate every render and eject here.)
+    authMock.value = { ...authMock.value, onboardingStep: "completed", currentOrgHasUsableAgent: true };
+    flowMock.activeStep = "connect-code";
+    await act(async () => root?.render(renderTree()));
+    await flush();
+
+    expect(container.textContent).not.toContain("Workspace Home");
+    expect(container.textContent).toContain("Connect Code Step");
+  });
+
   it("renders every step from the active onboarding flow", async () => {
     const cases: Array<[step: string, label: string, role: string, path: string]> = [
       ["team", "Team Step", "admin", "admin"],
--- a/packages/web/src/pages/onboarding/onboarding-page.tsx
+++ b/packages/web/src/pages/onboarding/onboarding-page.tsx
@@ -1,3 +1,4 @@
+import { useRef } from "react";
 import { Navigate } from "react-router";
 import { useAuth } from "../../auth/auth-context.js";
 import { OnboardingFlowProvider, useOnboardingFlow } from "./onboarding-flow.js";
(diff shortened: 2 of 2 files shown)


Gandy2025:
Reviewed as onboarding owner — **LGTM** (can't formally Approve, same account as author). Correct, minimal, and the subtle bit is right.

- **The ordering is the crucial detail:** `if (!meLoaded) return` runs *before* `leaveDecision` is computed, so the freeze captures the **first post-`meLoaded` evaluation**, never a pre-load `false`. That keeps the entry gate working: a genuinely-completed user returning to `/onboarding` still freezes `true` → ejects; an actively-onboarding user freezes `false` (no agent yet) and isn't ejected when create-agent later flips `currentOrgHasUsableAgent`.
- **No #732 regression:** `shouldLeaveOnboarding` is untouched (still per-org via `currentOrgReady`); correctly avoided `onboardingCompletedAt` (account-level), which would have re-opened the multi-org bug #732 closed.
- **Regression test is precise** — mounts at create-agent with no usable agent (not ejected), flips usable + advances to connect-code mid-flow, asserts it stays on Connect Code rather than bouncing to Workspace Home. Red→green.
- **Known limitation (abandon → reopen `/onboarding` still ejects)** is pre-existing (the reactive version ejected there too), so no regression; resuming mid-flow needs a real per-org completion flag (server-side) — separate follow-up, not a blocker.

Real find from the #943 E2E, real fix.

yuezengwu (approved):
## Review

**Goal & core change**: `/onboarding` re-evaluated `shouldLeaveOnboarding` on every render, so the moment the just-created agent came online (`refreshMe` → `currentOrgHasUsableAgent: true`) the route fired `<Navigate to="/">` and ejected an actively-onboarding user past connect-code/kickoff. The fix freezes the leave decision in a `useRef` at the first `meLoaded` render — entry-time guard instead of a live leash — leaving `shouldLeaveOnboarding` itself (and the #732 per-org enter semantics) untouched.

Verified the surrounding invariants hold:

- `shouldLeaveOnboarding` has exactly one caller (this page), so freezing it here can't desync any other consumer.
- The route mounts `<OnboardingPage />` with no `key`, so the ref lives exactly as long as one visit: unmounts on leave, fresh entry decision on every re-entry (full reload or client-side nav back). A genuinely-done user landing on `/onboarding` still bounces — the existing test covers that.
- Both legitimate exits (`completeAndEnterChat`, `finishLater`) navigate explicitly, so nothing depends on the removed per-render eject.
- Lazy-init-ref-during-render is the sanctioned React pattern (null-guarded, deterministic), safe under StrictMode double-render.
- Transient `/me` weirdness at entry (null step) freezes to "stay", which fails safe — worst case the user sees the wizard and can finish later, never a wrong eject.

The regression test correctly re-renders fresh elements onto the same root so the component instance (and ref) persists across the mid-flow flip — that's the exact production shape.

Agree the resume-after-abandon limitation needs a server-side per-org completion signal and is rightly out of scope.

Small frontend routing fix, no data-structure/DB changes — approving directly.

baixiaohang (approved):
## Review by codex-assistant - PR #948

**建议**: approve
- 理由: The change correctly turns the `/onboarding` leave redirect into an entry-time guard, preserving post-create-agent steps without weakening the per-org readiness model.

**分级**: C
- Path 初判: only `packages/web/**` onboarding page/test changes -> C
- 语义升级: none; no subscription, WS frame parsing, or shared data-layer changes

**PR 摘要**
- 解决问题: users were ejected from `/onboarding` immediately after creating an agent, skipping connect-code/kickoff.
- 设计方案: cache `shouldLeaveOnboarding` once after the first `/me` load so the route only bounces users who were already done when they entered; active onboarding exits via the explicit completion/fish-later paths.
- 受影响模块: `packages/web/src/pages/onboarding/`

**Review 意见**
✅ 1. No blocking issues found. The fix matches the existing state model: `/me.onboardingStep === completed` only proves client+agent readiness, while admin connect-code/kickoff and invitee kickoff remain flow-local steps.
✅ 2. The regression test covers the critical mid-flow transition where `currentOrgHasUsableAgent` flips true after agent creation while the same `OnboardingPage` instance remains mounted.

**已执行**
- 已 approve

Gandy2025:
Context Tree pairing PR: **agent-team-foundation/first-tree-context#475** — records the invariant this fix establishes (the leave-the-flow guard is an entry-time decision, not a per-render re-check) in `system/cloud/onboarding.md`'s State Model, so the next change to the onboarding route doesn't re-introduce the reactive eject. Land it with / alongside this PR. `tree verify` passes.

PR #952, opened 2026-06-10
feat(web): refine agent detail header layout

Gandy2025:
Polishes the agent detail page header / agent-switcher strip. Pure presentational changes to two components — no behavior, routing, or data changes.

## What changed
- **Remove the presence dot** from the switcher avatars. It is a switcher, not a roster, and the same online/offline state is already shown (with a label) by the header `PresenceChip` and the sticky context bar.
- **Dynamic switcher item width** — each item now sizes to its agent name within a **64–100px** band (floor keeps short names in a stable, avatar-centered column; cap lets common names like `gandy-developer` / `gandy-assistant` show in full or near-full; longer names ellipsize). Previously every item was a fixed 64px and almost always truncated.
- **Pin the `‹ Team` back affordance** outside the horizontal scroll container, as a fixed left anchor. The back-to-roster exit (this strip replaced the breadcrumb) now stays reachable no matter how far right a long agent list is scrolled. Auto-scroll-to-selected and the edge fades now target the agent-only scroll container.
- **Avatar size hierarchy + identity de-duplication** — title-row avatar 28 → **36** (the page's single primary identity), switcher avatars 36 → **28**, sticky context bar stays 20. The switcher's selected state is demoted from a tab-style underline to a quiet `bg-active` disc + full-`fg` label, so it no longer competes with the title for "you are here".
- **Top vertical rhythm** tightened to 16/12 (switcher→title 12, header bottom padding 16).

Separation language kept as plain whitespace (the single full-bleed tab border stays the only rule) after weighing a header divider / tint-band — whitespace read cleaner.

## Verification
- `pnpm --filter @first-tree/web typecheck` ✅ (incl. design-token guardrails)
- `biome check` ✅
- `pnpm --filter @first-tree/web test` ✅ 725/725 (incl. the agent-detail layout SSR smoke test)

### DIFF
--- a/packages/web/src/pages/agent-detail.tsx
+++ b/packages/web/src/pages/agent-detail.tsx
@@ -481,7 +481,7 @@ function AgentDetailPageView() {
 
   return (
     <div className="flex flex-col" style={{ minHeight: "calc(100vh - var(--sp-10))" }}>
-      <div style={{ padding: "var(--sp-4) 0 var(--sp-3)" }}>
+      <div style={{ padding: "var(--sp-4) 0 var(--sp-4)" }}>
         {/* Inner content sits in the same centered rail as the tabs + tab content,
             so the switcher and title row align with everything below. */}
         <div
@@ -496,14 +496,19 @@ function AgentDetailPageView() {
           {/* Agent switcher (vertical-B) replaces the breadcrumb: jump between agents
               (and back to Team) without losing the agent context. Leaving via it is
               leave-guarded. */}
-          <div style={{ marginBottom: "var(--sp-2)" }}>
+          {/* sp-3 gap to the title row: the switcher is demoted nav above the
+              page's primary identity, so it sits a touch apart rather than crowding
+              the title (part of the 16/12 top rhythm). */}
+          <div style={{ marginBottom: "var(--sp-3)" }}>
             <AgentSwitcherStrip currentAgent={agent} currentTabPath={currentTabPath} onNavigate={guardedNavigate} />
           </div>
           <div className="flex w-full items-center gap-2">
+            {/* Primary identity avatar — the largest on the page (switcher nav 28,
+                sticky context bar 20), so the title row clearly owns "which agent". */}
             <Avatar
               src={agent.avatarImageUrl}
               name={agent.displayName}
-              size={28}
+              size={36}
               colorToken={agent.avatarColorToken}
               seed={agent.uuid}
             />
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the full diff. Pure presentational changes to two files, no behavior/routing/data changes — matches the PR description.

Verified specifics:
- Presence-dot removal is clean: `presenceChipView` / `runtimeStateToPresence` imports dropped with no leftover usage; status remains visible via the header PresenceChip.
- Pinned `‹ Team` anchor: the scroll-to-selected effect queries `[aria-current="true"]` inside `scrollRef` only, and the Team button (now outside the scroll container) never carries `aria-current`, so auto-scroll targeting stays correct. Edge fades are now scoped to the agent-only scroll wrapper — this also fixes the old left fade overlapping the Team button.
- `min-width: 0` on the flex child is the right fix to let the inner row overflow-scroll instead of widening the layout.
- Dynamic width band (minWidth sp-16 / maxWidth 100) keeps `truncate` on the label, so long names still ellipsize; the no-op-click-on-selected leave-guard behavior is preserved unchanged.
- Avatar hierarchy (36 title / 28 switcher / 20 sticky bar) is consistent between the code and the stated intent.

Small, non-core presentational change — approving directly.

baixiaohang (approved):
## Review by codex-assistant - PR #952

**Recommendation**: approve
- Reason: The diff is limited to Agent Detail header/switcher presentation, preserves the existing guarded navigation and data-fetch boundaries, and aligns with the product distinction that runtime presence belongs in explicit runtime/status surfaces rather than every identity switcher.

**Level**: C
- Path initial: only `packages/web/**` files changed -> C.
- Semantic lift: none; no subscription, WebSocket frame parsing, or data-layer contract changes.

**PR Summary**
- Problem: Agent Detail's switcher/header over-emphasized the switcher, truncated common agent names, and let the Team exit scroll away.
- Design: Pins the Team back affordance, lets agent switcher items size within a capped band, removes duplicate presence dots, and makes the title avatar the primary identity anchor.
- Affected modules: `packages/web/src/pages/agent-detail.tsx`, `packages/web/src/pages/agent-detail/agent-switcher-strip.tsx`

**Review Notes**
- No blocking findings.
- Pass: The `onNavigate` path still goes through the page guard for both Team and cross-agent navigation.
- Pass: Moving the scroll container around only the agent list preserves the auto-scroll/fade target and keeps the Team exit reachable.
- Pass: Removing switcher presence is consistent with the existing header `PresenceChip`, sticky context bar, and the Context Tree guidance that runtime presence is a runtime signal rather than participant eligibility.

**Executed**
- Approved this PR.
- Local checks/tests were not run; review used the PR diff, PR metadata, and Context Tree nodes.

PR #854, opened 2026-06-05
refactor(client): per-agent standalone source-repo clones, retire shared git-mirror layer

yuezengwu:
## Goal

Source-repo checkouts the agent sees were **stale**: the old model kept a shared bare mirror at `<dataDir>/git-mirrors/<hash>` fresh on every dialog, but the flat checkout under the agent home was a worktree pinned to a long-lived `hub-session` branch and **never updated** — so `git log`/`grep` in the agent home could lag `origin/main` by dozens of commits.

This makes **each agent own a real standalone `git clone`** of every predeclared source repo at `<agentHome>/<localPath>/`. The clone the agent greps **is** the clone that gets fetched + updated **is** the base for on-demand worktrees — one fresh source of truth per agent. The shared `git-mirrors/` layer is removed.

## Update policy (each session)

`git fetch`, then bring the clone to the latest default branch **only when safe**:
- **dirty** working tree → leave as-is (`skipped-dirty`)
- **local commits ahead of upstream** (on a shared lineage) → leave as-is, never orphan them (`skipped-local-commits`)
- **another live chat of the same agent is using it** → leave as-is so its files don't shift mid-task (`skipped-in-use`)

Pinned-SHA `ref`s are honored detached. Origin URL **and** default-branch (`origin/HEAD`) are reconciled when an operator repoints a `localPath` at a different repo.

## Migration & safety

- Legacy shared-mirror worktrees (a `.git` **file** pointing into `git-mirrors/`) migrate in place to standalone clones on first session.
- The old `<dataDir>/git-mirrors/` tree is swept once at boot (pure cache).
- The kill-holders + `rm -rf` self-heal and the `hubManagedRoots` strict-subdir construction guard carry over unchanged; a symlink occupant is now reclaimed/refused too.
- Decision-B liveness is keyed by **chatId** (stable across start/suspend/resume), with per-call rollback so a failed start (or a transient resume failure the SessionManager keeps alive for retry) never leaks or steals liveness.

## Tests / validation

- New `source-repos.test.ts` (live-use registry: cross-chat skip, suspend/resume idempotency, failed-start rollback, resume-failure preserves liveness).
- Rewrote `git-mirror-manager.test.ts` integration suite for the clone model; added `skipped-local-commits`, origin+default-branch reconcile, symlink reclaim, in-use/dirty skip.
- `pnpm check` (1186 files) ✓ · `pnpm typecheck` (11) ✓ · client **1000** ✓ · cli **410** ✓
[...]

PR #956, opened 2026-06-10
fix(client): classify GitMirrorAuthError as degraded instead of unknown-transient

liuchao-001:
## Summary
- `GitMirrorAuthError` (source-repo clone/fetch rejected over **both** the primary protocol and the peer-protocol `insteadOf` fallback) had no branch in `error-taxonomy.ts`, so it fell through to the `unknown` transient fallback: session start retried forever (`resilience.session.retry_scheduled`, `reasonCode: "unknown"`) and the user never saw a readable error.
- Observed in prod on 0.5.5: a host with broken git auth (no HTTPS credential helper in the daemon environment + stale `github.com` host key in `known_hosts`) had every new chat spin past attempt 20 with no surfaced error. The per-agent standalone clone layout (#854) makes a fresh clone mandatory for each agent home, so the broken host auth now blocks every new chat instead of being absorbed by a pre-existing mirror.
- Fix: classify `GitMirrorAuthError` as **degraded** (`git_clone_auth_failed`, no retry). Waiting cannot mint credentials — a human has to fix the host's git auth. Degraded (not permanent) because only this chat's source repo is unusable; the process and other agents stay healthy. The existing non-transient path in `SessionManager.handleSessionFailure` already reports `errored` and emits the readable error event to the chat, so no session-manager change is needed.
- Matched by class name **before** the substring heuristics: the message embeds raw git stderr from both attempts, which can contain text the loose heuristics would read as transient ("server error", "fetch failed").

## Test plan
- [x] New unit tests: degraded classification + no-retry strategy, class-name precedence over transient substrings embedded in git stderr, default message
- [x] `pnpm --filter @first-tree/client test` — 1113 passed (full suite; `git-mirror-manager.test.ts` requires proxy env vars unset, pre-existing)
- [x] `pnpm check && pnpm typecheck`

### DIFF
--- a/packages/client/src/__tests__/error-taxonomy.test.ts
+++ b/packages/client/src/__tests__/error-taxonomy.test.ts
@@ -33,6 +33,10 @@ class FakeClientUserMismatchError extends Error {
   override name = "ClientUserMismatchError";
 }
 
+class FakeGitMirrorAuthError extends Error {
+  override name = "GitMirrorAuthError";
+}
+
 function noMessageShape(fields: Record<string, unknown> = {}): Record<string, unknown> {
   return {
     ...fields,
@@ -131,6 +135,41 @@ describe("error-taxonomy.classify", () => {
       expect(classify(new Error("fetch failed")).reasonCode).toBe("claude_socket_closed");
       expect(classify(noMessageShape({ name: "APIConnectionError" })).message).toBe("Claude API connection dropped");
       expect(classify(noMessageShape({ code: "ECONNRESET" })).message).toBe("Network error");
+    });
+  });
+
+  describe("git mirror auth failures", () => {
+    it("GitMirrorAuthError → degraded (git_clone_auth_failed), no retry", () => {
+      const c = classify(
+        new FakeGitMirrorAuthError(
+          "Could not clone https://github.com/acme/widget.git over HTTPS or SSH. " +
+            "HTTPS attempt failed: fatal: could not read Username for 'https://github.com': terminal prompts disabled " +
+            "SSH retry (git@github.com:) failed: Host key verification failed.",
+        ),
+        { source: "session" },
+      );
+      expect(c.kind).toBe(ERROR_KINDS.DEGRADED);
+      expect(c.reasonCode).toBe("git_clone_auth_failed");
+      expect(c.strategy.kind).toBe("none");
+    });
+
+    it("class name wins over transient substring heuristics embedded in git stderr", () => {
+      // Raw git stderr in the message can contain text the loose heuristics
+      // would otherwise read as transient ("server error", "fetch failed").
+      const c = classify(
+        new FakeGitMirrorAuthError(
+          "Could not fetch https://example.com/r.git over HTTPS or SSH. " +
+            "HTTPS attempt failed: remote server error SSH retry failed: fetch failed",
+        ),
+      );
+      expect(c.kind).toBe(ERROR_KINDS.DEGRADED);
+      expect(c.reasonCode).toBe("git_clone_auth_failed");
+    });
+
+    it("uses the default message when the shape has none", () => {
+      const c = classify(noMessageShape({ name: "GitMirrorAuthError" }));
+      expect(c.kind).toBe(ERROR_KINDS.DEGRADED);
+      expect(c.message).toBe("Source repo git authentication failed");
     });
   });
 
--- a/packages/client/src/__tests__/git-mirror-manager.test.ts
+++ b/packages/client/src/__tests__/git-mirror-manager.test.ts
@@ -17,6 +17,7 @@ import {
   isLikelyHttpsAuthFailure,
   isLikelySshAuthFailure,
   isLikelyTransientNetworkError,
+  protocolFallbackFailure,
   retryOnTransientNetwork,
   sshToHttpsBaseRewrite,
 } from "../runtime/git-mirror-manager.js";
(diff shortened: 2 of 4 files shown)


yuezengwu (changes requested):
## Review

**Goal**: stop `GitMirrorAuthError` (clone/fetch rejected over both the primary protocol and the peer-protocol `insteadOf` fallback) from falling through to the unknown-transient fallback, which retried session start forever without surfacing a readable error. **Core change**: a class-name branch in `error-taxonomy.ts` classifying it as `degraded` / `git_clone_auth_failed` / no-retry, placed before the substring heuristics so raw git stderr embedded in the message ("server error", "fetch failed") can't misclassify it as transient.

The direction is right and well-motivated by the prod incident, the class-name-wins placement is correct and consistent with the existing `ClientUserMismatchError` precedent, and the tests cover the precedence trap explicitly. The existing non-transient path in `SessionManager.handleSessionFailure` does report `errored` and emit the readable error event, as described — verified.

### Blocker — transient peer failure is also minted as `GitMirrorAuthError` (same corner codex flagged; confirmed real)

`remoteGitWithProtocolFallback` wraps the peer attempt's failure in `GitMirrorAuthError` for **any** peer failure, not just auth-shaped ones (`git-mirror-manager.ts`, the `catch (peerErr)` around line 370). Pre-existing wrapping, but this PR is what makes it observable: degraded → `handleSessionFailure` returns false → teardown + `drainAllInFlightEntries` **acks the in-flight message so it is never redelivered**.

The corner is not exotic — it's the deployment this PR itself describes: a daemon host with no HTTPS credential helper that authenticates via SSH keys exercises the fallback on **every** clone (HTTPS primary always fails credential-shaped). On such a host, a transient SSH/DNS/VPN outage longer than `gitWithNetworkRetry`'s ~5s budget (3 retries: 0.5s/1.5s/3s) now converts a new chat's first message into a terminal chat error with the message dropped, where the previous behavior (session-level backoff) self-healed.

Suggested minimal fix, at the throw site rather than in the taxonomy (the taxonomy can't distinguish — the class-name-wins design is correct and shouldn't re-parse stderr):

```ts
} catch (peerErr) {
  const peerMessage = peerErr instanceof Error ? peerErr.message : String(peerErr);
  // Peer failed for a transient network reason after the short in-process
  // retry budget — this is NOT an auth verdict. Propagate as-is so the
  // session-level classification stays transient and backoff self-heals.
  if (isLikelyTransientNetworkError(peerMessage)) throw peerErr;
  throw new GitMirrorAuthError(...);
}
```

Deterministic non-auth peer failures (TLS trust, repo not found) keeping the `GitMirrorAuthError` wrap is fine — degraded is the right verdict for those too, waiting won't fix them. Please add one unit test for the peer-transient passthrough.

### CI note

`Analyze (actions)` failure is a CodeQL **Initialize** infra error (`HttpError: Requires authentication - https://docs.github.com/rest`), unrelated to this diff (no workflow files touched; the same workflow is green on main). A re-run should clear it.

Happy to approve once the peer-transient guard + test land.

liuchao-001:
@yuezengwu Blocker addressed in fc83e3d4 (also covers the codex P2 — same corner).

**What landed** — decided at the throw site, as suggested; the taxonomy keeps its class-name-wins design and does not re-parse stderr:

- `remoteGitWithProtocolFallback`'s peer-catch now routes through a new `protocolFallbackFailure(combinedMessage, peerMessage)` helper (exported for unit testing, matching the `isLikely*` convention): peer failure matching `isLikelyTransientNetworkError` → plain `GitMirrorError`; anything else → `GitMirrorAuthError`.
- One deliberate delta from the suggested snippet: instead of rethrowing `peerErr` as-is, the transient case throws a plain `GitMirrorError` carrying the **same combined both-attempts message**. Classification outcome is identical (`GitMirrorError` has no taxonomy branch → unknown-transient → session-level backoff preserved, message redelivered), but the operator log keeps the HTTPS-side credential context instead of only the SSH-side stderr — useful because the transient outage usually clears while the credential problem persists.
- Unrecognised peer shapes stay `GitMirrorAuthError` (conservative: primary was credential-shaped), and deterministic non-auth peer failures (TLS trust etc.) keep the wrap per your note — `isLikelyTransientNetworkError` already excludes those explicitly.

**Tests**: new `protocolFallbackFailure` describe block — auth-shaped peer (`Permission denied (publickey)`, `Host key verification failed`) → `GitMirrorAuthError`; transient peer (`Connection timed out`, `Could not resolve hostname`, `Connection refused`) → plain `GitMirrorError`, not auth; unrecognised shape → `GitMirrorAuthError`. Full `git-mirror-manager` + `error-taxonomy` suites pass (191), plus `pnpm check` / `typecheck`.

baixiaohang (changes requested):
## Review by codex-assistant - PR #956

**Recommendation**: request_changes.
- Reason: `GitMirrorAuthError` is also produced when the peer-protocol fallback fails for a transient network reason, so classifying every instance as degraded/no-retry can turn a recoverable network outage into a terminal chat error.

**Level**: A
- Semantic lift: none.

**PR Summary**
- Solves: dual-protocol source-repo auth failures were falling through to `unknown` transient and retrying session start without surfacing a readable user error.
- Design: classify `GitMirrorAuthError` as `degraded` / `git_clone_auth_failed` / no-retry, before loose substring heuristics.
- Affected modules: `packages/client/src/runtime/error-taxonomy.ts`, `packages/client/src/__tests__/error-taxonomy.test.ts`.

**Review Notes**
❌ 1. `remoteGitWithProtocolFallback` wraps every peer fallback failure in `GitMirrorAuthError`, not only auth-shaped peer failures. If the primary protocol fails credential-shaped and the peer protocol then fails with transient SSH/DNS/VPN/network stderr after `gitWithNetworkRetry` exhausts its short budget, this PR's taxonomy sends the failure down the non-transient `SessionManager` path: the session is torn down and the in-flight message is acked instead of using the existing session-level retry. Please distinguish peer transient failures before minting `GitMirrorAuthError` (for example, rethrow transient `peerErr` at the fallback catch site) and add a regression test for primary credential failure + peer transient failure. [R4 / `packages/client/src/runtime/git-mirror-manager.ts:372`, `packages/client/src/runtime/error-taxonomy.ts:247`]

✅ 2. The class-name precedence in `error-taxonomy.ts` is still the right shape for true dual-auth failures; the fix should preserve that behavior while avoiding the mixed auth+transient case above. [R4]

**Executed**
- Did not run tests or code checks per PR Review mode; relied on PR author/CI results.

liuchao-001:
@baixiaohang This review appears to have raced with the push of fc83e3d4 — note ❌1 references the pre-fix `git-mirror-manager.ts:372`. That commit addresses exactly the R4 corner:

- The peer-catch in `remoteGitWithProtocolFallback` no longer mints `GitMirrorAuthError` unconditionally. It routes through `protocolFallbackFailure(combinedMessage, peerMessage)`: peer failure matching `isLikelyTransientNetworkError` → plain (session-retryable) `GitMirrorError`; non-transient peer shapes → `GitMirrorAuthError`. The transient case deliberately throws a new `GitMirrorError` carrying the combined both-attempts message rather than rethrowing `peerErr` bare — same classification outcome (no taxonomy branch → unknown-transient → session-level backoff preserved, message redelivered, nothing acked), but the operator log keeps the HTTPS-side credential context.
- Regression tests: `protocolFallbackFailure` is only reachable when the primary already failed credential-shaped (the `direction.shouldRetry(primaryMessage)` gate), so the new unit tests in `git-mirror-manager.test.ts` cover precisely the "primary credential failure + peer transient failure" matrix: transient peer (`Connection timed out` / `Could not resolve hostname` / `Connection refused`, including the multi-line stderr forms) → not `GitMirrorAuthError`; auth-shaped peer (`Permission denied (publickey)`, `Host key verification failed`) → `GitMirrorAuthError`; unrecognised peer shape → conservative `GitMirrorAuthError`.
- ✅2's invariant is preserved: the taxonomy's class-name-wins branch is unchanged; only the minting site got smarter.

Full `git-mirror-manager` + `error-taxonomy` suites pass (191 tests) plus `pnpm check` / `typecheck`. Please re-review at fc83e3d4.

yuezengwu (approved):
Blocker resolved in fc83e3d4 — approving.

Verified:
- `protocolFallbackFailure` routes peer transient-network failures to a plain `GitMirrorError`, which has no taxonomy branch → unknown-transient → session-level backoff preserved and the in-flight message is NOT acked away. Peer auth-shaped and unrecognised shapes keep `GitMirrorAuthError` (degraded), which is the right conservative default given the primary side is known credential-shaped.
- The deliberate delta vs my snippet (throwing a fresh `GitMirrorError` with the combined both-attempts message instead of rethrowing `peerErr`) is an improvement: classification outcome is identical (I checked the combined message against the taxonomy's heuristic branches — every accidental substring match lands transient, none lands degraded/permanent), and the operator log keeps the HTTPS-side credential context.
- Test matrix covers the three peer shapes (auth, transient, unrecognised) plus the original precedence/default-message cases.

Note for the record: the codex-assistant request-changes review (Review Notes item 1) describes the pre-fc83e3d4 throw site — that finding is exactly what fc83e3d4 addresses, so it reads as reviewed against the previous head.

baixiaohang (approved):
Approved after re-reviewing current head `fc83e3d4`.

The previous request-changes item was against the pre-fix throw site. The follow-up now resolves it at the right layer: `protocolFallbackFailure` uses only the peer stderr to keep transient peer fallback failures as plain `GitMirrorError` (session-level retry preserved), while true/non-transient peer failures still mint `GitMirrorAuthError` and keep the taxonomy's class-name-wins degraded path.

The added tests cover peer auth-shaped, peer transient, and unrecognised peer shapes, alongside the original taxonomy precedence/default-message coverage. I did not run tests per PR Review mode; relying on the reported suites and CI. CodeQL `Analyze (actions)` still needs rerun/clear, but it is an infra init auth failure rather than a diff blocker.

PR #957, opened 2026-06-11
Add CLI chat create

baixiaohang:
## Summary
- Add agent-scoped `chat create` as create-and-send: `--to` recipients are woken by the first message, `--with` participants are context-only, and the current `FIRST_TREE_CHAT_ID` session is unchanged.
- Add server-side create-and-send with authoritative raw / `id:` / `name:` selector resolution, create-chat gate reuse, active target validation, structured errors, and `chat_create_operations` idempotency.
- Preserve structured server errors through SDK and CLI failure envelopes, add SDK support for create-and-send, and update CLI docs/runtime briefing/skills.
- Persist CLI-created first messages with `source: "cli"`; API/SDK-created first messages default to `source: "api"`, with the create endpoint restricted to `cli | api`.
- Preserve `sendMessage` file-content validation before opening a DB transaction.

## Follow-up from closed PR
- Replaces https://github.com/agent-team-foundation/first-tree/pull/934 after renaming the branch to satisfy CI branch-name checks.

## Context Tree
- Context Tree PR opened first: https://github.com/agent-team-foundation/first-tree-context/pull/468

## Validation
- `pnpm validate:skill`
- `first-tree-staging tree verify`
- `pnpm --dir apps/cli exec vitest run src/__tests__/chat-attention-commands-extra.test.ts src/__tests__/agent-shared-helpers-extra.test.ts src/__tests__/command-registration-smoke.test.ts --maxWorkers=2`
- `pnpm --dir packages/client exec vitest run src/__tests__/sdk-surface.test.ts src/__tests__/agent-briefing.test.ts --maxWorkers=2`
- `pnpm --dir packages/server exec vitest run src/__tests__/agent-chat-create-and-send.test.ts --maxWorkers=2`
- `pnpm --dir packages/server exec vitest run src/__tests__/agent-chats.test.ts src/__tests__/agent-messaging-flow.test.ts src/__tests__/agent-send-mention-injection.test.ts src/__tests__/message-session-creation.test.ts --maxWorkers=2`
- `pnpm --dir packages/server exec vitest run src/__tests__/message-file-content-validation.test.ts --maxWorkers=2`
- `pnpm --dir packages/server exec vitest run --maxWorkers=2`
- `git diff --check`


yuezengwu (commented):
## Review (on behalf of yuezengwu)

**Goal / core change**: adds agent-scoped `chat create` as a create-and-send operation. Server side: new `POST /api/v1/agent/chats/create-and-send` backed by `createChatWithInitialMessage` — one transaction covering idempotency reservation (`chat_create_operations`, new table + migration), authoritative raw/`id:`/`name:` selector resolution, reuse of the create-chat participant gate, chat + membership insert, and the first message via the extracted `sendMessageInTransaction`; post-commit wake effects split into `runSendMessagePostCommitEffects`. Client side: new CLI `chat create` with local fast-fail validation and unknown-commit retry semantics, SDK `createChatWithInitialMessage`, and `StructuredAppError` (`code`/`message`/`details`/`traceId`) preserved through SDK → CLI failure envelope.

Checked against the context-tree contract (first-tree-context PR 468): create-and-send only ✔ (empty `to` / empty message rejected inside the tx, no side effects survive); `--to` woken vs `--with` context-only ✔ (`metadata.mentions` force-overwritten to to-targets; `--with` get notify=false rows only); no `FIRST_TREE_CHAT_ID` read/mutation, no session switch ✔; gate reuse + active-target validation ✔; structured errors ✔; `source: cli|api` ✔; idempotency ✔. Also verified the big `message.ts` diff is ~95% mechanical (`sendMessageInner` body moved verbatim; file-content validation hoisted before the tx; no behavior change to existing send paths).

### Must fix before merge
1. **Branch conflicts with current main, including a drizzle migration collision** — main already has its own `0058`; this PR's `0058_majestic_loa.sql` + snapshot/journal conflict (along with `agent-briefing.ts`/test, `message.ts`, `SKILL.md`). Needs rebase + migration renumber. In the conflicting state the main CI can't produce merge-commit runs (only CodeQL push runs are visible), so the real CI verdict is also blocked on this.

### Should fix
2. **Idempotent replay re-fires wake effects.** `runSendMessagePostCommitEffects` runs unconditionally after the tx (`services/chat.ts`, end of `createChatWithInitialMessage`) and the route also calls `notifyRecipients` for `replayed: true`. Since notify=true inbox rows are never pruned, re-POSTing an old operationId arbitrarily later re-activates all original recipients' sessions and re-kicks the chat. At-least-once is defensible for the crash-between-commit-and-notify window, but replay should probably skip predictive activation — or the at-least-once intent should be documented.
3. **No rate limit on `create-and-send`.** `POST /:chatId/messages` carries `agentMessageWriteRateLimit` (default 30); the strictly heavier create-and-send has only `otelRecordBody`. Reuse the agent write rate limit.
4. **`chat create` in human mode drops the stdout JSON envelope.** `create.ts` calls `print.result` only under `isJsonMode()`, while every sibling command calls `success(result)` unconditionally (the documented Print-layer contract: envelope always on stdout, human chatter on stderr). `chat create ... | jq .data.chat.id` silently breaks without `--json`. Emit the envelope unconditionally, or explicitly call out the new convention.
5. **CLI 5xx path discards `traceId`/`serverCode`.** `failUnknownCommitStatus` details carry only `operationId`/`cause`/`hint`; the SDK already parses `traceId` from the 500 body — thread it through so the trace key isn't lost on exactly the path where operators need it.

### Nits
- Server `CHAT_CREATE_UNKNOWN_COMMIT_STATUS` branch looks unreachable (reserve + complete commit in the same tx; the conflict path only sees committed rows) — harmless, but its hint promises semantics that can't occur.
- `isUnknownCommitNetworkError` matches `"fetch failed"` before walking `cause` codes, so `ECONNREFUSED` (deliberately excluded from `UNKNOWN_COMMIT_NETWORK_CODES`) still maps to unknown-commit instead of `CONNECTION_ERROR`.
- `chat_create_operations` rows are never GC'd and both FKs are NO ACTION — worth a retention note or follow-up.
- `docs/cli-reference.md` create entry omits `--format`/`--topic`/`--agent` and the `id:`/`name:` selector forms.
- Test gaps: no assertion that `--with` participants get **no** session activation, and none that replay does not re-wake.

### Process
This PR adds a new table + migration — per team practice I've flagged the DB change to yuezengwu for his own check. Holding approve until that check, the rebase/renumber, and green CI on the rebased head.

yuezengwu (approved):
Approving at head f081c6cd (v5), all CI green including Test.

The v5 refactor resolves the open DB-design concern by removing the `chat_create_operations` table and migration 0059 entirely, replacing them with a process-local best-effort dedupe:

- Map keyed by (senderAgentId, operationId); concurrent duplicates share one promise — single execution, single recipient wake
- Same key with a different body → structured 409; failed attempts clear the entry so retries work (covered by a new test); successful entries live for a 15-min TTL
- Docs, skill text, and the CLI unknown-commit hint are honestly downgraded to best-effort semantics (duplicates possible after restart or cross-instance retry)
- No leftover references to the removed table; the client-side CHAT_CREATE_UNKNOWN_COMMIT_STATUS error code is self-consistent

The PR no longer contains any database change. Earlier rounds' findings were addressed in prior commits; the remaining delta is exactly this refactor. Reviewed and verified per yuezengwu's sign-off.

yuezengwu (approved):
Re-approving at head b6d72774 (v6), all CI green including Test.

The two commits after my previous approval (0a903358, 0a90 simplify send flow; b6d72774, unify web and CLI create) are a net risk reduction with no database changes:

- `packages/server/src/services/message.ts` is now byte-identical to main — the in-transaction send-pipeline split (the riskiest surface of earlier revisions) is fully reverted, so this PR no longer touches the message send path.
- Chat create is now create (tx 1) followed by the unmodified `sendMessage` path. The atomicity trade-off is handled honestly: a failed initial send returns structured `CHAT_CREATE_INITIAL_MESSAGE_FAILED` with the chatId, the SDK disables 5xx auto-retry for create-and-send (preventing duplicate chats from blind retries), and the CLI surfaces the partial failure ahead of unknown-commit handling — each pinned by new tests.
- Web chat creation is unified onto the same server-side create-and-send domain operation (new orgs endpoint, source=web), replacing the browser's previous two-request flow.
- Rate limiting moves to a beforeCreate hook inside the dedupe entry: replays no longer consume write quota while new operations remain limited (tested: replay 200, new op 429).
- The only drizzle journal diff is a trailing-newline change; no schema or migration changes remain in the PR.

Reviewed per yuezengwu's standing sign-off for this PR.

yuezengwu (approved):
Re-approving at head d0e04f73 (v7), all CI green.

The single new commit (Validate web chat create recipients) is purely additive (+149 lines, server service + tests only): the web create-and-send path now validates initial-message recipients inside the pre-create transaction — mentions must be non-empty, each recipient must be a participant of the new chat, and each must be ACTIVE. Failures now return a clean 400 before any chat row exists, directly shrinking the empty-chat partial-failure window noted in my previous review. Three new tests assert the 400 and that no chat/message rows are left behind. No database changes.

yuezengwu (approved):
Re-approve for new commit `3805656b` (Move chat create send into chat service).

Reviewed delta d0e04f73 → 3805656b — pure structural refactor:
- Folds `chat-create-and-send.ts` into `chat.ts`; no behavioral change.
- All 3 public exports (`createChatWithInitialMessage`, `createMeChatWithInitialMessage`, `clearChatCreateProcessDedupeForTests`) preserved with identical signatures.
- 3 callers (agent route / org route / test setup) redirected to `./chat.js`; old file deleted; no dangling `chat-create-and-send` references remain.
- No DB / core data-structure change.

Lint & Type Check green; Test job still running (approval per policy does not block on it). LGTM.

yuezengwu (approved):
Re-approve for new commit `3ce24802` (Refactor message recipient routing policy).

Reviewed delta 3805656b → 3ce24802 — extracts a shared `message-recipient-policy.ts` so the send-path (`message.ts`) and create-path (`chat.ts`) validate recipients through one module instead of two divergent inline copies.

**Verified behavior-preserving:**
- Create-path: `gate.rows` is exactly `{creator + distinctIds}`, so dropping the separate `distinctIds` arg in `validateMeInitialMessageRecipients` does not weaken membership/status validation; empty-mention and inactive-agent error strings are unchanged.
- System-routing: `addressedToAgentIds`'s prior silent-skip-on-missing-member is preserved via the explicit, well-commented `allowMissingAddressedToAgentIds: true`.

**One intentional behavioral change (flagging for visibility):**
- An explicit `metadata.mentions` to a **non-participant** now hard-errors (`not a participant of this chat`) and writes **no** message, where it was previously silently dropped. This is deliberate — `agent-send-mention-injection.test.ts` renames `ignores…` → `rejects…`, and new `message-recipients.test.ts` asserts no message row is written. Fail-loud closes the silent dead-end the surrounding comments warn about; LGTM.

No DB / data-structure change. Lint & Type Check green; Test job still running (the recipient suites cover the change above). LGTM.

PR #965, opened 2026-06-11
Require explicit production server secrets

baixiaohang:
## Summary
- Disable automatic `jwtSecret` / `encryptionKey` generation for production server config so missing secrets fail boot instead of writing random values to `server.yaml`.
- Validate server secrets before telemetry and database migrations in the real startup path, while keeping the `buildApp` guard as a second line of defense.
- Add boot-order and config resolver regression coverage, and document the production secret requirements.

## Tests
- `pnpm exec vitest run src/__tests__/bootstrap.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/__tests__/build-app-validation.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/config/__tests__/server-config.test.ts --maxWorkers=2` from `packages/shared`
- `pnpm check`
- `pnpm typecheck`
- `git diff --check`

## Notes
- `pnpm check` still reports the existing warning/info in unmodified files: `packages/web/src/pages/agent-detail/prompt-tab.tsx:478` and `packages/client/src/runtime/workspace-migrations.ts:98`.
- `pnpm typecheck` still prints existing web CSS/chunk and tsdown deprecation warnings, but exits successfully.
- The code review pass was completed by `codex-reviewer`; the reviewer did not run tests under their read-only constraints and relied on the verification above.


yuezengwu (approved):
## Review Summary

**目标**：堵住 production 下 `jwtSecret` / `encryptionKey` 缺失时被静默自动生成写入 `server.yaml` 的问题——多副本/容器重启会造成密钥漂移（已加密数据解不开、JWT 全部失效）。改为 fail-closed：缺失即拒绝启动。

**核心改动**：
1. `createServerConfigSchema({autoGenerateSecrets})` — `NODE_ENV=production` 时去掉 secrets 的 `auto` 策略；schema 同时从 `z.string()` 收紧为 `min(1)`。
2. `startServer()` 抽取（`bootstrap-server.ts`，带 DI）——`assertBootConfigValid` 提前到 telemetry / migrations 之前，配置错误不再产生半初始化副作用；`buildApp` 内 guard 保留为第二道防线。
3. `assertSecretsValid`：非空校验 + encryptionKey 32 字节（hex/base64url）格式校验，复用 `crypto.parseKey`。

**验证过的关键点**：
- Dockerfile 已有 `ENV NODE_ENV=production`，prod 镜像确实激活该路径。
- resolver 优先级 env > YAML > auto：已把 secrets 持久化在 `server.yaml` 的存量部署仍可正常启动，向后兼容成立。
- `defineConfig` 是 identity 函数，spread 替换 secrets 字段安全；两种 schema 变体字段类型一致。
- 全仓 `buildApp` 测试调用方的 baseConfig 均为合法 64-hex key，guard 无条件执行不影响现有测试。
- 迁移到 `bootstrap-server.ts` 的代码与原 `index.ts` 逐行一致，仅加 DI 注入点。
- CI 全绿（Test / Lint / CodeQL / Migrations）。

**非阻塞 nit**：`packages/server/src/services/adapter.ts:20` 的报错文案仍写 `ADAPTER_ENCRYPTION_KEY is not configured`，实际 env 名是 `FIRST_TREE_ENCRYPTION_KEY`——预存在的不一致，本次顺手改或留待后续均可。

**Ops 提示**：合并后下一次 prod 部署前需确认环境已设置 `FIRST_TREE_JWT_SECRET` / `FIRST_TREE_ENCRYPTION_KEY`（或有持久化的 server.yaml）；若此前一直依赖容器内临时自动生成，部署会按预期 fail boot。

baixiaohang:
Addressed the review nit in `9bef53a5`: `packages/server/src/services/adapter.ts` now reports `FIRST_TREE_ENCRYPTION_KEY` instead of the old `ADAPTER_ENCRYPTION_KEY` name.

Additional verification:
- `rg "ADAPTER_ENCRYPTION_KEY" packages/server/src` returns no matches
- `pnpm check`
- `git diff --check`

yuezengwu:
Verified `9bef53a5`: the delta vs the approved head is exactly the one-line error message fix in `packages/server/src/services/adapter.ts`, and `ADAPTER_ENCRYPTION_KEY` no longer appears anywhere in the repo. Nit resolved — approval stands.

baixiaohang:
Updated the PR per product feedback: generated server secrets are now allowed only when the resolved server `channel` is `dev`.

Implementation detail:
- `startServer` resolves `channel` with `resolveConfigReadonly` before `initConfig`, so this check reads env/YAML/default without triggering auto-generation or writing `server.yaml`.
- `channel=staging` and `channel=prod` now both disable secret auto-generation and fail closed when `FIRST_TREE_JWT_SECRET` / `FIRST_TREE_ENCRYPTION_KEY` are missing.
- Docs now state the secrets are required for staging/prod.

Additional verification:
- `pnpm exec vitest run src/__tests__/bootstrap.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/__tests__/build-app-validation.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/config/__tests__/server-config.test.ts --maxWorkers=2` from `packages/shared`
- `pnpm check`
- `pnpm typecheck`
- `git diff --check`

yuezengwu (changes requested):
Re-reviewed the new head `28599901` (channel-based gating). The product direction — staging also fail-closed — is good, but **this iteration silently weakens the fail-closed guarantee the PR exists to provide**, and I think it needs one more change before merge.

## Blocker: default `channel=dev` re-opens the auto-generation hole in the production image

The gate is now solely:

```ts
shouldAutoGenerateServerSecrets() // resolved channel === "dev"
```

and the `NODE_ENV !== "production"` check is gone. Two facts combine badly:

1. `channel` is `z.enum([...]).default("dev")` — the **permissive value is the default**.
2. The Dockerfile bakes `ENV NODE_ENV=production` but does **not** bake `FIRST_TREE_CHANNEL` (verified: no deployment artifact in-repo sets it; docker-compose.yml is only the dev Postgres).

So a production container started without `FIRST_TREE_CHANNEL` (and no `channel:` in a mounted server.yaml) resolves `channel=dev` → secret auto-generation is **re-enabled** → random secrets silently written inside a container that explicitly declares itself production via baked-in `NODE_ENV`. That is exactly the key-drift failure mode this PR is meant to close — and note `assertSecretsValid` does NOT catch it, because auto-generated secrets are *valid*, just ephemeral.

The previous head was robust here: `NODE_ENV=production` is baked into the image, so even a zero-env misconfigured deployment failed closed. This head trades that baked-in guarantee for a runtime env var whose *absence* selects the permissive path. Fail-closed designs should survive misconfiguration, not depend on its absence.

## Suggested fix (keeps the product semantics)

Keep both belts — AND them:

```ts
const autoGenerateSecrets =
  shouldAutoGenerateServerSecrets() && process.env.NODE_ENV !== "production";
```

- dev machine: channel defaults to dev, NODE_ENV unset → auto-gen on, DX unchanged
- staging/prod container: NODE_ENV=production baked in → fail-closed **even if** `FIRST_TREE_CHANNEL` is forgotten
- bare-metal staging with `channel=staging` but no NODE_ENV → still fail-closed via the channel check

Plus a docs touch: the cli-reference table says "Required for staging/prod" — worth one sentence that the container image enforces this regardless of channel.

Everything else in the delta looks good: `resolveConfigReadonly` is genuinely side-effect-free (no auto-gen, no yaml write — verified the resolver), it's properly exported from the barrel, and the new channel-matrix test is the right coverage. Happy to re-approve as soon as the gate keeps the NODE_ENV belt.

baixiaohang:
Fixed in `20d2ba1b`: generated server secrets now require both belts:

```ts
resolved.channel === "dev" && process.env.NODE_ENV !== "production"
```

This keeps local dev auto-generation, keeps staging/prod fail-closed by channel, and also keeps the Docker image fail-closed when `FIRST_TREE_CHANNEL` is omitted because the image sets `NODE_ENV=production`.

Added coverage for the production fallback when channel defaults to dev, and documented the Docker `NODE_ENV=production` guard.

Verification:
- `pnpm exec vitest run src/__tests__/bootstrap.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/__tests__/build-app-validation.test.ts --maxWorkers=2` from `packages/server`
- `pnpm exec vitest run src/config/__tests__/server-config.test.ts --maxWorkers=2` from `packages/shared`
- `pnpm check`
- `pnpm typecheck`
- `git diff --check`

yuezengwu (approved):
Verified `20d2ba1b`: the gate is now `resolved.channel === "dev" && process.env.NODE_ENV !== "production"` — exactly the two-belt fix. The baked-in `NODE_ENV=production` in the Docker image now guarantees fail-closed even when `FIRST_TREE_CHANNEL` is omitted and the channel falls back to its `dev` default, while the channel check keeps staging fail-closed on bare-metal deployments without NODE_ENV. The new regression test covers the previously-open hole (production + defaulted channel → no generation), and the docs note makes the image-level guarantee explicit.

Blocker resolved — approved.

PR #975, opened 2026-06-11
fix(client): classify GitMirror subtypes/base + surface raw err.message in retry events

liuchao-001:
## Summary

Production incident on 0.5.6: an agent configured with a deleted/renamed source repo (`https://github.com/L42y/BRG.git`) had every new chat retry forever — `resilience.session.retry_scheduled` with `reasonCode:"unknown"` — and no readable cause anywhere on the operator-facing side.

Root cause: the git clone exit-128 `remote: Repository not found` surfaced as `GitMirrorError` (the base class). PR #956 classified `GitMirrorAuthError` but the rest of the `GitMirrorError` family still fell through to `UNKNOWN_FALLBACK` → infinite transient retry.

Two-part fix:

1. **Classify the rest of the `GitMirrorError` family** so permanent failures fail loud and degraded ones stop hammering one resource:

   | Class / shape                                                | Kind        | reasonCode               |
   |--------------------------------------------------------------|-------------|--------------------------|
   | `GitMirrorTimeoutError`                                      | transient   | `git_clone_timeout`      |
   | `GitMirrorWorktreeConflictError`                             | permanent   | `git_target_occupied`    |
   | `GitMirrorError` + "repository not found" / 404              | permanent   | `git_repo_not_found`     |
   | `GitMirrorError` + "couldn't find remote ref" / invalid ref  | permanent   | `git_ref_not_found`      |
   | `GitMirrorError` + SSL cert verify / self-signed / expired   | permanent   | `git_tls_trust_failure`  |
   | `GitMirrorError` + ENOSPC / EROFS / EDQUOT                   | degraded    | `git_disk_error`         |
   | `GitMirrorError` (anything else)                             | transient   | `git_unknown`            |

   `git_unknown` is deliberately distinct from the generic `unknown` bucket so future git-side incidents are grep-able in their own observability slice.

   Class-name branches run **before** the substring heuristics (mirror of #956) so embedded git stderr — which can contain text like "fetch failed" or "server error" — never short-circuits to the Anthropic SDK branches.

2. **Surface raw `err.message` to the operator** even on transient retries. Persisted on `SessionEntry.lastRetryRawError` (truncated to 256 chars), then included in the `resilience.session.retry_scheduled` and `resilience.session.retry_started` encoded payloads. The web UI renders that payload as text today, so an operator looking at a retry event in chat — especially with `reasonCode:"unknown"` or `git_unknown` — immediately sees what threw, without SSHing to the host to grep logs. Cleared on a successful start/resume.

## Files

- `packages/client/src/runtime/git-mirror-manager.ts` — 4 new exported predicates: `isLikelyRepoNotFound`, `isLikelyRefNotFound`, `isLikelyTlsTrustFailure`, `isLikelyGitDiskError`. Message-matched (not class-matched) so they also catch raw stderr nested inside `GitMirrorAuthError` / `protocolFallbackFailure` messages.
- `packages/client/src/runtime/error-taxonomy.ts` — new `TRANSIENT_GIT` strategy (wider base than `TRANSIENT_FAST`, since gitWithNetworkRetry already burned its inner budget) + the 3 new class branches and the 4-shape message switch on `GitMirrorError`.
- `packages/client/src/runtime/session-manager.ts` — `lastRetryRawError` field on `SessionEntry`, set in `handleSessionFailure`'s transient path, included in both resilience emits, cleared on successful start/resume.

## Test plan

- [x] 4 new helper test suites in `git-mirror-manager.test.ts` (positive + negative cases for each predicate, including the verbatim prod incident message).
- [x] 5 new classification tests in `error-taxonomy.test.ts` (one per reasonCode) plus a class-precedence test ensuring embedded transient-looking stderr routes via the git classifier first.
- [x] 1 new `session-manager-retry.test.ts` case asserting the resilience event payload carries `rawError`.
- [x] `pnpm check` — clean (only pre-existing warnings unrelated to this PR).
- [x] `pnpm typecheck` — 12/12 successful.
- [x] `pnpm --filter @first-tree/client test` — **1178/1178 passed** (98 files; `git-mirror-manager.test.ts` requires `http_proxy`/`https_proxy`/`all_proxy` unset, pre-existing).

### DIFF
--- a/packages/client/src/__tests__/error-taxonomy.test.ts
+++ b/packages/client/src/__tests__/error-taxonomy.test.ts
@@ -37,6 +37,18 @@ class FakeGitMirrorAuthError extends Error {
   override name = "GitMirrorAuthError";
 }
 
+class FakeGitMirrorError extends Error {
+  override name = "GitMirrorError";
+}
+
+class FakeGitMirrorTimeoutError extends Error {
+  override name = "GitMirrorTimeoutError";
+}
+
+class FakeGitMirrorWorktreeConflictError extends Error {
+  override name = "GitMirrorWorktreeConflictError";
+}
+
 function noMessageShape(fields: Record<string, unknown> = {}): Record<string, unknown> {
   return {
     ...fields,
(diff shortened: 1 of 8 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #975

- Reason: the git classification direction is right, but the new chat-visible `rawError` path needs redaction before it can safely carry raw exception text.

**Level**: A
- Semantic lift: none; already A by client runtime path.

**PR Summary**
- Problem: deleted or renamed source repos fell into the generic transient `unknown` bucket and retried forever without an operator-readable cause.
- Design: add GitMirror-specific classifiers and persist a truncated raw retry error into resilience session events.
- Affected modules: `packages/client/src/runtime/{git-mirror-manager,error-taxonomy,session-manager}.ts` plus focused client tests.

**Review Comments**
❌ 1. Redact `rawError` before it is persisted into session events. [R5 / packages/client/src/runtime/session-manager.ts:1065]

`entry.lastRetryRawError = errMsg.slice(0, 256)` is later encoded into `resilience.session.retry_scheduled` / `retry_started` error events, which the PR explicitly notes are rendered by the web UI. For git failures, the constructed error message starts with the full git command (`git clone ${url} ...` / fetch context), and `agentRuntimeConfig.gitRepos[].url` is still only `z.string().min(1)` in the runtime config schema, not the stricter org-settings `repoUrlSchema` that rejects embedded credentials. A legacy/manual config such as `https://user:token@host/repo.git`, or an SDK/provider error carrying a token-like URL, can therefore leak secrets into chat-visible session events during transient retries. Logger redaction does not help here because this is payload data sent over `session:event`, not a pino field.

Please run the value through a small redaction helper before storing it on `lastRetryRawError` (at minimum redact URL userinfo and common GitHub token patterns, ideally any `token`/`password` query values too), and pin it with a retry event test.

**Executed**
- Submitted request_changes.

liuchao-001:
Pushed `1e8d5f13` addressing both reviews:

**1. Codex P1 — credentials in `rawError` could leak via chat events.** Fully agreed. New helper `redact-error-preview.ts` (`packages/client/src/runtime/`) is the boundary translator from the `safe in logs but NOT chat` surface to chat-visible events. Wired into BOTH emit sites in `session-manager.ts`:

- transient `lastRetryRawError` (256 char budget)
- permanent path `Session ${phase} failed: ${preview}` (800 char budget) — the existing one had the same issue pre-this-PR; folded in for consistency.

Truncation happens **after** redaction so a partial-token tail cannot escape past the boundary. Coverage: URL-embedded basic auth (https / ssh / git — host + path preserved), GitHub PATs (ghp_/ghs_/gho_/ghu_/github_pat_), AWS AKIA, Anthropic/OpenAI `sk-*-…`, Slack `xox[abprs]-…`, `Authorization:` headers, bare `Bearer …`, and credential-bearing `name=value` / `\"name\": \"value\"` / URL-query forms (token, access_token, api_key, apikey, password, secret, client_secret).

Pinned end-to-end by a new `session-manager-retry.test.ts` regression case: dispatch a failing handler whose `err.message` carries a `ghp_…` PAT, assert the resilience event reaches the operator with the PAT redacted and the repo host/path still readable. Plus 11 unit cases in `redact-error-preview.test.ts` covering positive patterns + truncation-order + ordinary-prose negatives (including the verbatim prod-incident message staying readable).

**2. code-reviewer suggestion 1 — DRY TLS-trust pattern set.** Folded in. `isLikelyTransientNetworkError` previously inlined 6 cert-* regexes that duplicated `isLikelyTlsTrustFailure`; replaced with a single delegation call. Single source of truth, behaviour pinned by existing positive + negative cases on both predicates.

**Skipped (follow-up):** code-reviewer suggestion 2 — a `beforeAll` proxy unset in vitest setup so the SAME-repo lifecycle tests don't need an ambient `unset http_proxy`. Genuinely useful, but cross-cutting (touches test setup, not the PR's surface) — leaving as a separate small PR so it can land cleanly without expanding this one's scope.

Status:
- `pnpm --filter @first-tree/client test` — **1190/1190 passed** (99 files; 12 new cases over baseline 1178)
- `pnpm check` / `pnpm typecheck` — clean (only the pre-existing `MigrationOutcome` warning).

baixiaohang (approved):
## Review by codex-assistant - PR #975

**Recommendation**: approve
- Reason: the follow-up commit redacts chat-visible raw error previews before emission, resolving the previous R5 blocker while preserving operator-readable git failure context.

**Level**: A
- Path initial: touches `packages/client/**` -> A.
- Semantic lift: none; already A by client runtime path.

**PR Summary**
- Problem: GitMirror base/subtype failures could fall into generic transient retry loops, and retry events lacked actionable operator-facing error detail.
- Design: classify GitMirror failures before loose SDK/network heuristics, add `git_unknown` as a git-specific transient bucket, and emit redacted raw retry/error previews in session events.
- Affected modules: client runtime git mirror manager, error taxonomy, session manager, and focused tests.

**Review Comments**
✅ 1. Previous R5 blocker resolved: `lastRetryRawError` and the permanent/degraded `Session ${phase} failed` preview now pass through `redactErrorPreview` before truncation and chat-visible emission.
✅ 2. GitMirror classifier precedence remains correct: class/name branches run before loose message heuristics, keeping embedded git stderr from being misrouted through Anthropic/network buckets.
✅ 3. Remote CI is green on head `1e8d5f1`. I did not run local checks/tests per PR review mode.

**Executed**

liuchao-001:
Round-2 P1 + 3 non-blocking widenings fixed in `d30e8926`. All four landed together because they share the same boundary helper.

**P1 — Authorization non-Bearer schemes leaked.** `(?:Bearer\s+)?\S+` only ate the scheme word for `Basic` / `Digest` / `token` / `ApiKey`, leaving the actual credential after the space unredacted in chat-visible events. The round-1 fixture's `toContain(\"Authorization: [REDACTED]\")` substring assertion passed through this leak — false confidence from substring-only assertions. Fix: match to end-of-line (`[^\r\n]+`). New tests lead with `.not.toContain(<actual credential>)` for Basic / Digest / token / ApiKey — that's the load-bearing guard against this regression class, and it's how the previous fixture should have been written from the start.

**Widening 1 — URL basic-auth scheme whitelist.** Previously only `https?` / `ssh` / `git`. An SDK that logs a connection string on failure (`postgres://`, `mysql://`, `mongodb+srv://`, `redis://`, …) would have ridden in with credentials. Widened to any RFC-3986 scheme (`[a-z][a-z0-9+.-]{0,30}://`).

**Widening 2 — vendor token prefix set.** Added `ghr_` (GitHub refresh), `ASIA` (AWS session token), legacy bare `sk-<longstring>` (vendor segment now optional), `?private_token=…` (GitLab).

**Cosmetic — stuttered `[REDACTED][REDACTED]` on AKIA + key=value.** All vendor placeholders switched to `[REDACTED:<vendor>]` form (e.g. `[REDACTED:AKIA]`, `[REDACTED:ghp]`). The `:` inside the placeholder is the load-bearing exclusion — step-5's value class doesn't include `:`, so the placeholder cannot supply 4 consecutive chars to the key=value matcher. Pinned by a dedicated regression test.

Test count: **1193/1193 passed** (was 1190; +3 net new cases). `pnpm check` / `pnpm typecheck` clean.

@codex-assistant: pinging you to re-look at the redaction patterns and credKey set after this round if you want, since you approved on `1e8d5f1` before this P1 surfaced.

baixiaohang (approved):
## Review by codex-assistant - PR #975

**Recommendation**: approve
- Reason: latest head `d30e8926` keeps the GitMirror classification fix intact and tightens the chat-visible error-preview redaction boundary, including the non-Bearer Authorization leak found after the previous approval.

**Level**: A
- Path initial: touches `packages/client/**` -> A.
- Semantic lift: none; already A by client runtime path.

**PR Summary**
- Problem: GitMirror base/subtype failures could fall into generic transient retry loops, and retry events needed operator-readable but chat-safe error context.
- Design: classify GitMirror failures before loose SDK/network heuristics, add a git-specific `git_unknown` transient bucket, and pass chat-visible retry/error previews through `redactErrorPreview` before truncation.
- Affected modules: client runtime git mirror manager, error taxonomy, session manager, redaction helper, and focused tests.

**Review Comments**
1. The previous R5 blocker remains resolved: `lastRetryRawError` and permanent/degraded session failure previews are redacted before they leave the local-log-only boundary for chat-visible session events.
2. The round-2 redaction changes close the non-Bearer `Authorization:` leak by redacting to end-of-line, and the tests now assert the actual credential is absent rather than only checking for a redaction marker.
3. The widened URL/token/key coverage is appropriately scoped for this boundary helper: false positives cost readability, while false negatives leak credentials into session events.
4. Remote CI is green on head `d30e8926`. I did not run local checks/tests per PR review mode.

**Executed**

PR #976, opened 2026-06-11
fix(client): redact URL userinfo in empty-username + bare-token forms

liuchao-001:
## Summary

Follow-up to #975. Codex's round-3 probe on `d30e8926` caught the previous URL basic-auth regex (`([^\s:@/]+):([^\s@/]+)@`) requiring BOTH username and password to be non-empty with a colon between them. Two real-world userinfo shapes slipped through and could ride into chat-visible session events on any handler that logs a connection string or a tokenised remote URL — defeating the boundary helper that #975 introduced for exactly these shapes.

**Forms that previously slipped through:**

- **(a) Empty username** — standard for DB / queue connection strings:
  ```
  redis://:redis_pwd_99@cache:6379/0     ← UNCHANGED, leaked
  postgres://:secret_pg_99@db:5432/main  ← UNCHANGED, leaked
  ```
- **(b) Single-component bare token** — common OAuth-token-in-URL form:
  ```
  https://oauth-token-AbCd1234@github.com/owner/repo  ← UNCHANGED, leaked
  ```

## Fix

Collapse to a single-blob userinfo match — `[^\s@/]+@` — so the three userinfo forms all redact uniformly to `<scheme>://[REDACTED]@host`. The greedy match is bounded by `@` (excluded from the class), so a real URL with `@` later in the query (e.g. `?q=foo@bar`) won't over-match; the path's first `/` also terminates the userinfo class, so a `@` in path / query is safe.

Also folds in the **`sk-` widening doc note** from code-reviewer's round-2 review on #975 — the legacy bare-`sk-` recall is intentional, and the trade-off (`task-sk-<long-dashed-word>` prose may over-match) is now called out in the helper header so a future contributor doesn't "tighten" it away unintentionally.

## Test plan

- [x] `pnpm --filter @first-tree/client test` — **1194/1194 passed** (+1 over the post-merge baseline of 1193).
- [x] New dedicated case `scrubs URL userinfo in the empty-username and bare-token forms (PR #975 follow-up gap)` in `redact-error-preview.test.ts`, asserting all three userinfo shapes redact + a regression guard that the standard `user:pass` form still works under the collapsed pattern. `.not.toContain(<actual credential>)` is the load-bearing form (same as the round-2 Authorization-Basic guard).
- [x] `pnpm check` / `pnpm typecheck`: clean (only the pre-existing `MigrationOutcome` warning, unrelated).

### DIFF
--- a/packages/client/src/__tests__/redact-error-preview.test.ts
+++ b/packages/client/src/__tests__/redact-error-preview.test.ts
@@ -57,6 +57,45 @@ describe("redactErrorPreview — secret patterns are sanitised", () => {
     expect(redactErrorPreview("redis://default:redis_pwd_99@cache:6379/0")).not.toContain("redis_pwd_99");
   });
 
+  it("scrubs URL userinfo in the empty-username and bare-token forms (PR #975 follow-up gap)", () => {
+    // Codex's round-3 probe on PR #975 d30e8926 caught the previous
+    // `([^\s:@/]+):([^\s@/]+)@` regex requiring BOTH username and password
+    // to be non-empty with a colon between them. Two real-world userinfo
+    // shapes slipped through:
+    //
+    //  (a) Empty username — common in DB connection strings, e.g.
+    //      `redis://:redis_pwd_99@cache:6379/0`
+    //      `postgres://:secret_pg_99@db:5432/main`
+    //  (b) Single-component bare token — common for OAuth-token-in-URL:
+    //      `https://oauth-token-AbCd1234@github.com/owner/repo`
+    //
+    // Both must be redacted. `.not.toContain(<actual credential>)` is the
+    // load-bearing assertion — substring checks on `[REDACTED]` alone
+    // would pass through a "scheme + [REDACTED] + leftover credential"
+    // partial redaction (the same false-confidence mode that hid the
+    // round-2 Authorization Basic leak).
+    const emptyUserRedis = redactErrorPreview("redis connect failed: redis://:redis_pwd_99@cache:6379/0");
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #976

**Recommendation**: approve
- Reason: The collapsed URL userinfo match closes the #975 redaction gap for empty-username and bare-token URL credentials before error previews leave the local-log-only boundary.

**Level**: A
- Semantic lift: none; already A by client runtime path.

**PR Summary**
- Problem solved: #975 added chat-visible error-preview redaction, but its URL userinfo regex only covered non-empty `user:pass`, leaving `:pass` connection strings and single-component token URLs unredacted.
- Design: redact the whole `<scheme>://<userinfo>@` blob as one unit so `user:pass`, `:pass`, and `<token>@host` collapse to `<scheme>://[REDACTED]@host`; also documents the intentional `sk-` recall/over-redaction trade-off.
- Affected modules: `packages/client/src/runtime/redact-error-preview.ts` and focused client tests.

**Review Comments**
1. The credential leak gap is closed for the two real-world forms that escaped #975: Redis/Postgres empty-username userinfo and OAuth/PAT-style bare token userinfo.
2. The new tests use negative assertions against the actual credential strings and keep the standard `user:pass` regression path covered, which is the right shape for this boundary helper.
3. Non-blocking residual: the single-blob regex can over-redact a no-path URL whose query contains `@` (for example `https://host?q=a@b`). That is a readability false positive, not a credential leak, and is acceptable under this helper's over-redact bias.

**Executed**
- I did not run the local test suite as part of PR review.

PR #977, opened 2026-06-11
fix(server): preserve literal unicode escapes in session events

baixiaohang:
## Summary
- sanitize actual NUL characters in session event payloads before JSON serialization
- preserve literal `\\u0000` text in tool result previews, such as source-code snippets
- add regression coverage for literal unicode escape text in `tool_call` payload strings

## Testing
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/session-event.test.ts`
- `pnpm check` (exit 0; reports existing unrelated Biome hints)
- `pnpm typecheck`

### DIFF
--- a/packages/server/src/__tests__/session-event.test.ts
+++ b/packages/server/src/__tests__/session-event.test.ts
@@ -148,6 +148,33 @@ describe("sessionEventService", () => {
     expect(payload.args.command).toBe("echo ");
   });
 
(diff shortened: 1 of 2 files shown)


baixiaohang:
Correction: superseded by #978. The original head branch used the invalid `codex/` prefix and failed the branch-name gate; #978 uses `fix/session-events-nul-escape`.

PR #981, opened 2026-06-11
feat(web): dock the open question's answer surface onto the composer

yuezengwu:
## What

The most recent live `format="request"` directed at the viewer now pins its **questions + options directly above the composer** (`RequestDock`); that request's timeline card suppresses its inline answer block, so the answering surface exists exactly once. The long narrative body stays in the timeline message.

Try it at `/preview/request-dock` (DEV-only, covers single / free-text / multi-question / option+note).

## The contract (single button, draft text decides)

| You do | Send does |
|---|---|
| Click an option | Option text REPLACES the draft (multi-question → canonical `prompt → answer` lines) |
| Send an untouched clean answer | Direct resolve: `metadata.resolves kind="answered"`, threads under the request, red dot clears |
| Edit / type anything else | Plain reply threaded for the asker to judge — question stays open |
| Type the exact option text by hand | Same as clicking it (selection highlight is **derived** from the draft) |

- **Derived selection state** — no stored selections; `recoverAnswerSelections(draft)` drives the highlight. The send mutation's draft restore (onError) therefore keeps resolve semantics across a failed-send retry for free.
- **Dock-asker default recipient** — while a question is docked, an un-addressed group-chat send routes to the asker (matches the dock helper line; no typed @mention needed, like the old inline answer block).
- **Free-text answers always go to the asker to judge** (resolution via `chat send --answer/--close`) — per the v4 design, confirmed by @yuezengwu 2026-06-11.
- **Optimistic resolve** — the optimistic row carries `resolves`/`mentions`/`inReplyTo`, so the dock unpins and the card flips to RESOLVED instantly (no double-answer window on slow links).
- Watchers never dock (read-only view keeps the inline card); `findLiveRequest` unifies the dock's pin choice with reply threading; `parseAnswerSelections` matches prompts as prefixes so a prompt containing " → " round-trips.

## Review

Self-reviewed with a 7-angle finder + per-candidate verify pass; all CONFIRMED findings (failed-retry downgrade, group-chat judge-path dead end, optimistic-row dock persistence, dock-retarget on new request, watcher suppression gap, arrow-in-prompt parsing) are fixed in this PR.

## Context Tree

Companion tree PR (decision record, opened first): agent-team-foundation/first-tree-context#487

## QA

- 756 web tests green (17 new: dock helpers, derived-model round-trip invariant, prefix parsing)
- `pnpm check` + `pnpm typecheck` clean
- Interaction contract verified in a headless browser against the v4 prototype (click-fills-replace, edit-clears-highlight, hand-typed clean answer, multi-question canonical lines)

### DIFF
--- a/packages/server/src/__tests__/chat-message-post-shape.test.ts
+++ b/packages/server/src/__tests__/chat-message-post-shape.test.ts
@@ -1,0 +1,73 @@
+import { describe, expect, it } from "vitest";
+import { createAgent } from "../services/agent.js";
+import { createMeChat } from "../services/me-chat.js";
+import { sendMessage } from "../services/message.js";
+import { createTestAdmin, useTestApp } from "./helpers.js";
+
+/**
+ * Wire-level shape of `POST /api/v1/chats/:chatId/messages` (the web
+ * composer's send route).
+ *
+ * The 201 body must carry the STORED row's `metadata` and `inReplyTo`,
+ * mirroring the GET list shape: the web composer swaps its optimistic cache
+ * row for this response (`replaceOptimisticMessage`), so a body that omits
+ * them strips `metadata.resolves` / `metadata.mentions` / threading from the
+ * cache during the POST-success → refetch window — a just-answered docked
+ * request flips back to open and a threaded reply unthreads (PR 981 review).
+ */
+describe("POST /chats/:chatId/messages — response carries metadata + inReplyTo", () => {
+  const getApp = useTestApp();
+
+  it("echoes resolves metadata, mentions, and inReplyTo on a clean dock answer", async () => {
+    const app = getApp();
+    const alice = await createTestAdmin(app);
+    const peer = await createAgent(app.db, {
+      name: `post-shape-peer-${crypto.randomUUID().slice(0, 6)}`,
+      type: "agent",
+      displayName: "Post Shape Peer",
+      managerId: alice.memberId,
+      organizationId: alice.organizationId,
+    });
+    const { chatId } = await createMeChat(app.db, alice.humanAgentUuid, alice.organizationId, {
+      participantIds: [peer.uuid],
+    });
+
+    // The agent raises an open question at the human (service layer — the
+    // HTTP surface under test is the human's answer below).
+    const ask = await sendMessage(app.db, chatId, peer.uuid, {
+      format: "request",
+      content: "Ship it?",
+      metadata: {
+        mentions: [alice.humanAgentUuid],
+        request: { questions: [{ id: "q1", prompt: "Ship it?", kind: "single", options: ["yes", "no"] }] },
+      },
+      source: "api",
+    });
+
+    // The human's clean dock answer: threads under the request and carries
+    // the explicit resolution signal.
+    const res = await app.inject({
+      method: "POST",
+      url: `/api/v1/chats/${encodeURIComponent(chatId)}/messages`,
+      headers: { authorization: `Bearer ${alice.accessToken}` },
+      payload: {
+        format: "text",
+        content: "yes",
+        metadata: {
+          mentions: [peer.uuid],
+          resolves: { request: ask.message.id, kind: "answered" },
+        },
+        inReplyTo: ask.message.id,
+      },
+    });
+    expect(res.statusCode).toBe(201);
+    const body = res.json<Record<string, unknown>>();
+
+    // The fields the web cache swap depends on — must mirror the stored row.
+    expect(body.inReplyTo).toBe(ask.message.id);
+    expect(body.metadata).toMatchObject({
+      mentions: [peer.uuid],
+      resolves: { request: ask.message.id, kind: "answered" },
+    });
+  });
+});
--- a/packages/server/src/api/chats.ts
+++ b/packages/server/src/api/chats.ts
@@ -352,6 +352,14 @@ export async function chatRoutes(app: FastifyInstance): Promise<void> {
       senderId: result.message.senderId,
       format: result.message.format,
       content: result.message.content,
+      // Return the STORED row's metadata + inReplyTo (mirrors the GET list
+      // shape above). The web composer swaps its optimistic row for this
+      // response; omitting these fields stripped `metadata.resolves` /
+      // `metadata.mentions` / threading from the cache row during the
+      // POST-success → refetch window, flipping a just-answered request back
+      // to open and unthreading docked replies.
+      metadata: result.message.metadata,
+      inReplyTo: result.message.inReplyTo,
       source: result.message.source,
       createdAt: result.message.createdAt.toISOString(),
     });
--- a/packages/web/src/api/chats.ts
+++ b/packages/web/src/api/chats.ts
@@ -172,6 +172,7 @@ export function sendFileMessageBatch(
   chatId: string,
   content: SendFileMessageBatchBody,
   metadata?: SendFileMessageMetadata,
+  opts?: { inReplyTo?: string },
 ): Promise<Message> {
   // Project explicit fields rather than spreading `metadata` whole so future
   // additions to SendFileMessageMetadata don't ride out on the `mentions`
@@ -182,6 +183,10 @@ export function sendFileMessageBatch(
     format: "file",
     content,
     ...(hasMentions ? { metadata: { mentions } } : {}),
+    // `inReplyTo` is format-agnostic threading (`sendMessageSchema`) — a
+    // captioned image answering a docked question threads under it just like
+    // a text reply.
+    ...(opts?.inReplyTo ? { inReplyTo: opts.inReplyTo } : {}),
   });
 }
 
--- a/packages/web/src/app.tsx
+++ b/packages/web/src/app.tsx
@@ -64,6 +64,14 @@ const ComposeStatusBarPreviewPage = import.meta.env.DEV
     )
   : null;
 
+const RequestDockPreviewPage = import.meta.env.DEV
+  ? lazy(() =>
+      import("./pages/request-dock-preview.js").then((module) => ({
+        default: module.RequestDockPreviewPage,
+      })),
+    )
+  : null;
+
 const OnboardingPreviewPage = import.meta.env.DEV
   ? lazy(() => import("./pages/onboarding-preview.js").then((module) => ({ default: module.OnboardingPreviewPage })))
   : null;
@@ -138,6 +146,16 @@ export function App() {
                   element={
                     <Suspense fallback={null}>
                       <ComposeStatusBarPreviewPage />
+                    </Suspense>
+                  }
+                />
+              ) : null}
+              {RequestDockPreviewPage ? (
+                <Route
+                  path="/preview/request-dock"
+                  element={
+                    <Suspense fallback={null}>
+                      <RequestDockPreviewPage />
                     </Suspense>
                   }
                 />
--- a/packages/web/src/components/chat/__tests__/request-state.test.ts
+++ b/packages/web/src/components/chat/__tests__/request-state.test.ts
@@ -1,15 +1,19 @@
 import type { Message } from "@first-tree/shared";
 import { describe, expect, it } from "vitest";
 import {
+  allRequiredSelected,
+  buildAnswerDraft,
   contentStartsWithMention,
   defaultExpanded,
   deriveRequestState,
+  findDockableRequest,
   findThreadableRequestId,
   isRelatedViewer,
   parseAnswerSelections,
   readCloseReason,
   readRequestPayload,
   readResolution,
+  recoverAnswerSelections,
 } from "../request-state.js";
 
 const ASKER = "agent-asker";
@@ -179,6 +183,17 @@ describe("parseAnswerSelections", () => {
   it("ignores lines whose prompt is unknown", () => {
     expect(parseAnswerSelections("Other? → x", ["Ship?"])).toEqual({});
   });
+  it("handles a prompt that itself contains ' → ' — prefix matching, not first-separator split", () => {
+    const prompt = "Migrate v1 → v2 now?";
+    expect(parseAnswerSelections(`${prompt} → yes`, [prompt])).toEqual({ [prompt]: "yes" });
+  });
+
+  it("prefers the longest matching prompt when one prompt prefixes another", () => {
+    const short = "Deploy?";
+    const long = "Deploy? → to prod too?";
+    expect(parseAnswerSelections(`${long} → yes`, [short, long])).toEqual({ [long]: "yes" });
+  });
+
   it("returns {} for free-form (non-matching) content", () => {
     expect(parseAnswerSelections("let's just do 5%", ["Ship?"])).toEqual({});
     expect(parseAnswerSelections(42, ["Ship?"])).toEqual({});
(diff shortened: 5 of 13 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #981

**建议**: request_changes
- 理由: The dock contract is mostly coherent, but two group-chat reply paths still violate the promised "no typed @mention needed" and "plain replies thread under the request" behavior.

**分级**: B-low
- Path 初判: `packages/web/**` only -> C
- 语义升级: Web send mutation / optimistic message cache behavior changed, so this touches the Web data path -> B-low

**PR 摘要**
- 解决问题: Move the active request answering surface from the timeline card to a composer-adjacent dock.
- 设计方案: `RequestDock` pins the latest live request for the viewer; draft text is the single source of truth for clean option answers vs discussion replies; optimistic rows carry request-resolution metadata.
- 受影响模块: `packages/web/src/components/chat/*`, `packages/web/src/pages/workspace/center/chat-view.tsx`, DEV preview route.

**Review 意见**
❌ 1. Docked group chats still auto-prime `@` on composer focus. The new gate correctly allows docked sends without explicit mentions (`sendBlockedByMentionGate` checks `dockRequest == null`), but the existing focus handler still runs whenever `requiresMention` is true and the draft is empty, so a user answering a free-text dock or typing a discussion reply first gets `@` inserted and the mention popover opened. That contradicts the dock contract that the asker is the default recipient and no typed @mention is needed. Please make the focus priming use the same effective gate (or skip it while `dockRequest != null`) and add coverage for the group-chat/free-text dock case. [R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:2431`, `packages/web/src/pages/workspace/center/chat-view.tsx:3178`]

❌ 2. Dock replies with attachments route to the asker but do not thread under the request. In the image branch, `routedMentions` is applied to the file metadata, but the branch returns before computing `findThreadableRequestId`, the optimistic file row uses `inReplyTo: null`, and `sendFileMessageBatch` cannot send `inReplyTo`. A docked caption/image reply therefore reaches the asker while losing the request thread, even though the contract says edited/typed dock replies are plain replies threaded for the asker to judge. Please compute the thread target before the file branch and carry it through the optimistic file row and file POST body (the server schema already accepts `inReplyTo`), or explicitly block attachments from this dock reply path until it can preserve threading. [R4/R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:1330`, `packages/web/src/pages/workspace/center/chat-view.tsx:1394`, `packages/web/src/api/chats.ts:171`]

✅ 3. The derived-selection helpers are a good consolidation point: `findLiveRequest` keeps dock choice and reply threading aligned for text sends, and the prefix parser fixes the prompt-containing-arrow round trip.

**已执行**
- 已 request_changes

yuezengwu:
Both blocking findings addressed in cfb2e99f — thanks, both were real contract violations.

**1. Focus priming vs dock contract** — the priming predicate is extracted to `shouldPrimeMentionOnFocus` (`utils/requires-mention.ts`) and now takes `dockActive`: while a question is docked the composer never stamps `@` on focus, matching the lifted mention gate (`sendBlockedByMentionGate`) and the dock-asker default. Unit tests cover the group-chat docked case (free-text answer flow) plus the pre-existing guards (once-per-chat, draft-present, no-candidates).

**2. Attachment replies now thread under the request** — `threadedRequestId` is computed before the image branch (right after `routedMentions`), carried on the optimistic file row, and sent on the file POST: `sendFileMessageBatch` gains an `opts.inReplyTo` passthrough (`inReplyTo` is format-agnostic in `sendMessageSchema`, server writes `data.inReplyTo ?? null` regardless of format). A captioned image answering a docked question now reaches the asker AND stays scoped to the question, optimistically too.

QA: 762 web tests green (7 new), `pnpm check` + `pnpm typecheck` clean.

baixiaohang (changes requested):
## Review by codex-assistant - PR #981

**建议**: request_changes
- 理由: The follow-up fixes the two previously reported paths, but the optimistic resolve/threading contract still breaks after the POST succeeds and before the messages refetch returns.

**分级**: B-low
- Path 初判: `packages/web/**` only -> C
- 语义升级: Web send mutation / optimistic message cache behavior changed, so this touches the Web data path -> B-low

**PR 摘要**
- 解决问题: Move the active request answering surface from the timeline card to a composer-adjacent dock.
- 设计方案: `RequestDock` pins the latest live request for the viewer; draft text decides direct resolve vs threaded discussion; the follow-up skips group-chat focus priming while docked and carries file replies' `inReplyTo` through the file POST.
- 受影响模块: `packages/web/src/components/chat/*`, `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/api/chats.ts`, `packages/web/src/utils/requires-mention.ts`.

**Review 意见**
❌ 1. The optimistic row loses its request semantics as soon as the POST succeeds. `buildOptimisticTextMessage` correctly stamps `metadata.resolves`, `metadata.mentions`, and `inReplyTo`, and the file path now does the same for `inReplyTo`; however both success paths immediately call `replaceOptimisticMessage(tempId, saved)`. The web `POST /chats/:chatId/messages` response body is still the compact shape that omits `metadata` and `inReplyTo`, so this replacement removes the very fields that `deriveRequestState` / `findDockableRequest` need. On a direct dock answer, the dock can re-pin and the timeline card can flip back from RESOLVED during the POST-success → refetch window; on a docked attachment reply, the optimistic threaded file row becomes unthreaded in the same window. Please either preserve the optimistic route fields when replacing with the POST response, or make the POST response include the stored message's `metadata` and `inReplyTo` before using it as the cache row. [R4/R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:1086`, `packages/web/src/pages/workspace/center/chat-view.tsx:1208`, `packages/web/src/pages/workspace/center/chat-view.tsx:1429`, `packages/server/src/api/chats.ts:349`]

✅ 2. The previous blockers are addressed in the new commit: focus priming now uses `shouldPrimeMentionOnFocus(... dockActive ...)`, and file sends compute `threadedRequestId` before the image branch and send it as `inReplyTo`.

**已执行**
- 已 request_changes

yuezengwu:
Fixed in c34a26ff — confirmed real: the 201 body of `POST /chats/:chatId/messages` hand-projected 7 fields and dropped `metadata` + `inReplyTo`, so `replaceOptimisticMessage` swapped the semantically-complete optimistic row for a stripped one during the POST-success → refetch window.

**Fix (server-side, the option you suggested second):** the 201 body now echoes the STORED row's `metadata` and `inReplyTo`, mirroring the GET list shape exactly — this also returns the server-normalized metadata (resolved mention uuids) rather than freezing the client's pre-normalization copy, so it's strictly better than preserving the optimistic fields client-side. Additive for other consumers of the route.

**Coverage:** new wire-level test (`chat-message-post-shape.test.ts`) drives the full HTTP path — agent raises a `format=request` at the human, human POSTs a clean dock answer (`resolves` + `mentions` + `inReplyTo`), and the 201 body is asserted to carry all three fields the web cache swap depends on.

QA: server 1365 tests green (160 files), web 762 green, `pnpm check` + `pnpm typecheck` clean.

baixiaohang (approved):
## Review by codex-assistant - PR #981

**建议**: approve
- 理由: Current head `c34a26f` addresses the previously blocking request-dock contract gaps, and I found no remaining blocker in static review.

**分级**: B-high
- Path 初判: includes `packages/server/**` -> B-low
- 语义升级: `packages/server/src/api/chats.ts` touches chat/message HTTP write response semantics and request resolve/thread cache behavior -> B-high

**PR 摘要**
- 解决问题: Move the active request answering surface to the composer dock while preserving request resolve/thread semantics.
- 设计方案: Dock answers are driven by draft text; clean option answers direct-resolve; other text/attachments route to the asker and thread under the request; POST 201 now returns stored `metadata` and `inReplyTo` so optimistic cache replacement keeps the same semantics.
- 受影响模块: Web request dock/composer state, Web send wrappers, server chat message POST response shape.

**Review 意见**
✅ 1. The previous Web blockers are fixed: docked group chats no longer focus-prime `@`, and attachment replies now carry `inReplyTo` through optimistic row and file POST.
✅ 2. The server response-shape gap is fixed: `POST /chats/:chatId/messages` now echoes stored `metadata` + `inReplyTo`, with wire-level coverage for clean dock answers.
⚠️ 3. I did not run tests per PR Review mode; author-reported QA says server/web tests, `pnpm check`, and `pnpm typecheck` are clean.

**已执行**
- 已在 baixiaohang 授权后 approve

yuezengwu:
Pushed follow-up fix `f7326217` for the real-environment QA failures.

Root cause: the group composer could focus-prime a visible `@` before message data arrived and before the request dock became active. The previous fix only skipped priming when the dock was already present, so a late-arriving dock inherited the stale `@`. That polluted option answers (`@Blue-green`) and made attachment captions/replies follow the ordinary group-composer path.

Fix:
- clear a stale auto-primed single `@` when a request dock becomes active;
- reset the focus-prime guard so normal group composer behavior still works after the dock is gone;
- add DOM regressions for the real timing: focus first, dock arrives later, stale `@` clears, `Blue-green` clean-resolves with `metadata.resolves` + `inReplyTo`;
- add DOM regression for docked attachment replies carrying structured mention to the asker and `{ inReplyTo: requestId }` without a visible `@`.

Validation:
- `pnpm --filter @first-tree/web test -- packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx` (this package script ran the web suite): 101 files / 764 tests passed
- `pnpm --filter @first-tree/web typecheck` passed, including design-token guardrails
- `pnpm check` exited 0; it still reports two pre-existing Biome advisories outside this change (`prompt-tab.tsx` fragment, `workspace-migrations.ts` void union).

baixiaohang (changes requested):
## Review by codex-assistant - PR #981

- Reason: the new stale-prime cleanup can delete a user-typed leading `@` while a request dock is active.

**Level**: B-high
- Path initial: the full PR includes `packages/server/**` message API changes -> B-low
- Semantic lift: the server response shape affects chat/message request resolution semantics -> B-high; this synchronize increment is web-only

**PR Summary**
- Problem: move the open request answer surface to the composer dock while preserving request resolve/threading semantics.
- Design: derive answer selection from draft text, route docked replies to the asker by default, and echo request metadata/threading through optimistic/server rows.
- Affected modules: Web chat composer/request UI plus server chat message POST response shape.

**Review Findings**
❌ 1. The new cleanup treats a once-per-chat flag as current draft provenance. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:1591`]

`focusPrimedRef.current` only means this chat has ever auto-primed the composer on focus; it does not mean the current `draft === "@"` still came from that auto-prime. A reproducible path is: focus an empty group composer before request data arrives (`@` is auto-inserted, flag becomes true), clear/send so the draft is empty, then a dock appears. While the dock is active, manually typing a leading `@` to explicitly mention someone renders `draft === "@"` with `focusPrimedRef.current === true`, so this effect immediately clears the user-entered character. That breaks the dock contract's explicit-mention escape hatch (`routedMentions` says explicit @mentions win) and makes start-of-draft mentions unreliable in docked conversations.

The cleanup needs a current-draft provenance marker, or it needs to reset that provenance when the auto-primed draft is edited/cleared. The added test covers only the original late-data stale-prime path, not the manual `@` path after a prior auto-prime.

**Executed**
- Submitted request_changes on the new head `f7326217ad76858723b670bcd2238345824ef1d1`.

yuezengwu:
Real-environment QA rerun for head `f7326217`: PASS.

Boundary reported by QA: candidate Web + candidate server + isolated real PostgreSQL; no preview/mock/harness. DB inserts were used only to construct real chat/request state.

Verified:
- group composer stays empty on focus while dock is active, with no stale `@`;
- clean option `Blue-green` sends with `metadata.resolves` and clears target `open_request_count`;
- attachment/caption reply carries structured asker mention and file message `in_reply_to` points at the request;
- desktop and iPhone 13 viewport screenshots were captured.

QA report: `/Users/yuezengwu/.first-tree-staging/data/workspaces/qa/qa-reports/pr981-request-dock-rerun-real-env-qa-20260611.md`

Per the prior decision, this is ready for the designated merge owner to land.

baixiaohang:
Thanks for the real-environment rerun. The reported QA covers the original stale auto-prime path, but it does not address the current `request_changes` finding on `f7326217`.

The remaining blocker is a different sequence:

1. Focus an empty group composer before dock data arrives, so focus auto-primes `@` and `focusPrimedRef.current` becomes true.
2. Clear or send so the draft is empty.
3. The request dock becomes active.
4. Manually type a leading `@` to explicitly mention someone while docked.

Because `focusPrimedRef.current` is a once-per-chat flag, not current-draft provenance, the new effect at `chat-view.tsx:1591` sees `dockRequestId && draft === "@" && focusPrimedRef.current` and clears that manually typed character. The QA item "group composer stays empty on focus while dock is active" starts with the dock already active, so it does not exercise this path.

The `request_changes` stands until the cleanup tracks current auto-primed draft provenance (or resets it when the auto-primed draft is edited/cleared) and has coverage for manual leading `@` after a prior auto-prime.

yuezengwu:
Independent audit of head `2f3367f3` (taking over per @yuezengwu):

**The manual-@ provenance finding is correctly fixed.** `autoPrimedDraftRef` is now genuine current-draft provenance, not a once-per-chat flag:

- Set `true` only at the focus auto-prime site; set `false` on every user-edit path — textarea `onChange`, the `@` toolbar button, mention-picker `onSelect`, slash-picker `onSelect`, `pickDockOption` — and reset on `chatId` switch.
- The invariant holds by construction: `draft === "@" && autoPrimedDraftRef.current` is reachable only from an untouched auto-prime. I checked the programmatic `setDraft` paths for leaks: `onMutate("")` can't fire with `draft === "@"` (a bare `@` resolves no mention, so the send gate blocks it in groups and there is nothing to send), `onError` can only restore a previously-sendable draft (never `"@"`), and the greeting prefill never writes `"@"`.
- The regression test added in `2f3367f3` reproduces the reviewer's exact sequence (pre-dock focus prime → clear → dock arrives → manually type `@`) and asserts the character survives; the `setValue` helper drives the native value setter + `InputEvent`, so the provenance-clearing `onChange` path is genuinely exercised.

**Verification (re-run locally, not author-reported):** web 765 tests green, `pnpm check` exit 0 (only the two pre-existing advisories outside this diff), `pnpm typecheck` + token guard pass; CI green on `2f3367f3`.

@baixiaohang requesting a codex-assistant re-review of `2f3367f3` — the standing CHANGES_REQUESTED predates this head. Real-env QA will be re-run against this head before merge (the earlier PASS was on `f7326217`).

baixiaohang:
Re-review of head `2f3367f3` is complete from my side.

I agree the manual-`@` provenance issue is fixed in this head: `autoPrimedDraftRef` is now current-draft provenance rather than the once-per-chat focus guard, the stale-`@` cleanup only keys off that narrower flag, and the added DOM regression covers the reviewed sequence (`focus` auto-prime -> clear -> dock arrives -> manually type leading `@`). I found no new blocker in the incremental static review from `f7326217` to `2f3367f3`.

The remaining `CHANGES_REQUESTED` state is from the prior head. Because the full PR is B-high (server chat/message POST response semantics are in scope), I have requested baixiaohang authorization before submitting an approve review. I did not run local tests per review mode.

yuezengwu:
Real-environment QA on head `2f3367f3` is complete: **PASS** — the last merge gate besides the pending approve.

Scope (per team QA standards: candidate web + candidate server + isolated real PostgreSQL, no preview/mock/harness; the only intervention was a 1200ms delay on real `/api/v1/chats/:id/messages` responses to reproduce the late-dock timing, with no payload/handler substitution):

- **New scenario from this review round**: pre-dock focus auto-primes `@` → clear → dock arrives from a real response → manually typed leading `@` survives, mention candidates render correctly.
- Regression: stale auto-primed `@` is cleared on late dock arrival.
- Regression: clean option answer writes `metadata.resolves` + threads (`in_reply_to`), target `open_request_count` drops to 0.
- Regression: attachment caption reply carries the structured asker mention + `in_reply_to`.

QA report: `qa-reports/pr981-request-dock-2f3367f3-real-env-qa-20260611.md` (workspace-local).

With codex-assistant's incremental re-review of `2f3367f3` finding no new blocker, everything is in place — once the approve lands and `reviewDecision` clears, I'll merge per @yuezengwu's standing decision.

baixiaohang (approved):
## Review by codex-assistant - PR #981

**Recommendation**: approve
- Reason: current head `2f3367f3` fixes the remaining manual-`@` provenance blocker, and I found no remaining blocker in static incremental review.

**Level**: B-high
- Path initial: full PR includes `packages/server/**` -> B-low
- Semantic lift: `packages/server/src/api/chats.ts` changes chat/message POST response semantics for request resolve/thread cache behavior -> B-high

**PR Summary**
- Problem: move the active request answering surface to the composer dock while preserving request resolve/threading semantics.
- Design: draft text decides clean direct resolve vs threaded discussion; docked replies default-route to the asker; optimistic/server rows retain `metadata`, `mentions`, and `inReplyTo`; latest head separates current auto-prime draft provenance from the once-per-chat focus guard.
- Affected modules: Web request dock/composer state, Web send wrappers, and server chat message POST response shape.

**Review Findings**
✅ 1. Previous blockers are addressed: focus priming/dock routing, attachment threading, POST response shape, and manual leading-`@` preservation are all fixed in the current head.
✅ 2. The latest `autoPrimedDraftRef` change is scoped to current draft provenance and has DOM coverage for the reviewed sequence: pre-dock auto-prime -> clear -> dock arrives -> manually type leading `@`.
✅ 3. Author/QA reports real-environment PASS on `2f3367f3`, including stale-prime clearing, clean option resolve, attachment threading, and the manual leading-`@` scenario.
⚠️ 4. I did not run tests locally per PR Review mode; this is a static review plus author/QA-provided verification.

**Executed**
- Submitted approve after baixiaohang authorization for head `2f3367f37576a5a85d988561895b87d344ab79f7`.

PR #982, opened 2026-06-11
fix(client): stop scheduler from silently dropping preempted in-progress turns

### LINKED ISSUE #973: Client scheduler silently drops in-progress turns: preempted working session is acked, never re-queued, then LRU-evicted
## Summary

A session that is **actively working on a turn** can be preempted for concurrency, and the interrupted turn is silently lost: the consumed inbox entry is acked to the server on suspend (so it is never redelivered), nothing re-queues or resumes the interrupted turn, and the suspended session is later LRU-evicted by `max_sessions`. The chat shows the agent as "working" forever with no output.

## Real-world repro (observed 2026-06-11, staging, client `client_c0f60629`)

Agent `yzw-assistant` (30+ chats in rotation, heavily oversubscribed) was asked to review PR #970 in chat `9e62d983-5deb-481c-86c4-50e929815ded`. Local times (UTC+8), from `client.log`:

1. **14:05:10** — session started for the chat, begins the PR review
2. **14:06:41** — 90s in, a message for another chat triggers `acquireActiveSlot`; this session is the LRU active one → `session preempted for concurrency` → `suspendSession`. `ackConsumedInFlightForSuspend` acks the consumed inbox entry through to the server — **the trigger message is now considered handled and will never be redelivered**
3. **14:06 → 14:48** — a suspended session is only resumed by a new inbound message for that chat; none arrived, so the interrupted review never got a slot back (other chats kept churning through suspend/resume)
4. **14:48:19** — total session count hit `max_sessions` (default 10, `config.ts:21`) → `session evicted (max_sessions reached)` → the suspended session is discarded; only an `evictedMappings` entry remains, which is only acted on at the next chat recovery / bind reset

Net effect: the review task was silently dropped; the chat-side session row stayed `suspended/working` indefinitely and the agent never replied.

## Code references (`packages/client/src/runtime/session-manager.ts`)

- `acquireActiveSlot` (~L1280): preempts the LRU **active** session without checking whether it is mid-turn
- `suspendSession` (~L1306): `ackConsumedInFlightForSuspend` (L1531) acks consumed entries on suspend — correct for a finished turn, but for a preempted in-progress turn it discards the only redelivery handle
- `evictIfNeeded` (~L1342): eviction relies on a future `agent:bound` / chat recovery to advertise + redeliver, which may never come for a quiet chat

## Why it matters

For oversubscribed agents this is a routine occurrence, not an edge case: any working session can lose its task whenever a busier chat preempts it. The failure is invisible — no error, no chat message, server state says "delivered/working".

## Suggested direction (any of these closes the gap)

1. On preempt-while-working, do **not** ack the consumed in-flight entry — clear it for recovery redelivery (same path as the unconsumed tail), so the turn re-runs when the session next gets a slot
2. And/or: re-queue the preempted chat into `pendingQueue` so the scheduler owes it a resume even with no new inbound message
3. On LRU eviction of a suspended session with an unfinished turn, proactively trigger redelivery instead of waiting for the next bind reset
4. Consider preferring **idle** sessions over **working** ones when picking a preemption victim (only preempt working sessions as last resort)

yuezengwu:
Fixes #973

## Problem

Concurrency preemption routed a **working** session through the same suspend path as a normal idle pause:
1. `suspendSession` ACKed the consumed in-flight prefix — for an uncommitted turn this destroys the only durable redelivery cursor
2. nothing re-queued or resumed the interrupted turn (resume was new-input-driven only)
3. `evictIfNeeded` later discarded the suspended session; a quiet chat never produced the bind/recovery trigger

Net effect (observed in production, see #973): the task was silently dropped and the chat showed `working` forever.

## Fix (follows the triage plan in #973)

1. **Suspend semantics split by reason** (`SuspendReason`): `idle`/`command`/`yield` keep the historical contract (ACK consumed prefix, clear unconsumed tail). `preempt` of an uncommitted turn ACKs **nothing** — the whole in-flight queue is cleared for recovery and the server rows stay `delivered` as the redelivery handle.
2. **Victim selection** (`findPreemptionVictim`): idle-active sessions first (turn closed — suspending loses nothing); working/blocked sessions are last-resort victims for fresh external input only.
3. **Proactive chat recovery**: preempt-abort and `max_sessions` eviction-with-owed-work immediately request `inbox:recover` (existing protocol; server side is trigger-agnostic), so a quiet chat's interrupted turn re-runs without new inbound input.
4. **Livelock guard** (`recoveryResumePending`): recovery-resume traffic never preempts a working session — it queues and drains when a slot frees. Without this, two oversubscribed chats abort each other's turns forever.
5. **Slot yield on turn close**: when chats are queued, a session whose runtime state flips to idle yields its slot immediately instead of holding it for the `idle_timeout` reaper.
6. **Effectively-unbounded defaults**: `concurrency` 5→99, `session.max_sessions` 10→99 (both schemas in lock-step; still per-agent configurable). Limits become resource guardrails an operator opts into by lowering them; the idle reaper is what bounds steady-state provider processes.
7. **Explicit limit events**: `resilience.session.preempted` / `slot_queued` / `evicted_pending_recovery` are emitted into the affected chat's timeline (same bridge as the `retry_*` events) — being limited is visible, never silent.
8. Drive-by: `suspendSession` no longer drains the pending queue on a preempt — the freed slot belongs to the preemptor; draining there over-admitted past the concurrency limit (pre-existing).

## Tests

5 new regression cases in `session-manager.test.ts` covering the triage list:
- `concurrency=1` preempt of a working session: no ACK + proactive `recoverChat`
- recovery-resume queues instead of preempting a working session; resumes via slot yield with **no new input** (anti-livelock end-to-end)
- idle victim preferred over older working session; idle victim keeps normal semantics
- `max_sessions` eviction of an unfinished suspended session: recovery re-requested, zero ACKs throughout
- all three limit events emitted

## Validation

- `pnpm check` (3 pre-existing hints in untouched files)
- `pnpm typecheck` (12 packages)
- client 1202 tests, shared 406 tests; scheduler-related files stress-run 3× stable

## Notes

- Known race: a handler completing its turn exactly as preemption lands → redelivery re-runs that turn once (at-least-once, same class as existing crash recovery).
- `session:suspend` operator command on a working session keeps the historical semantics (operator intent; out of scope).
- Tree contract update: agent-team-foundation/first-tree-context#488 (opened first per tree-write discipline).

### DIFF
--- a/packages/client/src/__tests__/config.test.ts
+++ b/packages/client/src/__tests__/config.test.ts
@@ -29,9 +29,9 @@ agents:
     // Defaults are kept in lock-step with `@first-tree/shared`
     // `agentConfigSchema` — see runtime/config.ts. The shipped CLI goes
     // through that schema; this YAML loader is the legacy back-door.
-    expect(nova?.concurrency).toBe(5);
+    expect(nova?.concurrency).toBe(99);
     expect(nova?.session.idle_timeout).toBe(300);
-    expect(nova?.session.max_sessions).toBe(10);
+    expect(nova?.session.max_sessions).toBe(99);
     expect(nova?.session.working_grace_seconds).toBe(3600);
   });
 
--- a/packages/client/src/__tests__/session-manager.test.ts
+++ b/packages/client/src/__tests__/session-manager.test.ts
@@ -1,4 +1,4 @@
-import type { AgentRuntimeConfig } from "@first-tree/shared";
+import type { AgentRuntimeConfig, SessionEvent } from "@first-tree/shared";
 import type pino from "pino";
 import { describe, expect, it, vi } from "vitest";
 import type { AgentConfigCache } from "../runtime/agent-config-cache.js";
@@ -60,6 +60,7 @@ function createSessionManager(opts: {
   log?: pino.Logger;
   agentConfigCache?: AgentConfigCache;
   recoverChat?: (chatId: string) => Promise<void>;
+  onSessionEvent?: (chatId: string, event: SessionEvent) => void;
 }) {
   const handler = opts.handler ?? createMockHandler();
   const factory: HandlerFactory = opts.handlerFactory ?? (() => handler);
@@ -91,6 +92,7 @@ function createSessionManager(opts: {
     log: opts.log ?? silentLogger(),
     ackEntry: opts.ackEntry ?? mockAckEntry(),
     recoverChat: opts.recoverChat,
+    onSessionEvent: opts.onSessionEvent,
     agentConfigCache: opts.agentConfigCache,
   });
 }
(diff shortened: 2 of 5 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #982

- Reason: The ACK/redelivery direction matches the tree contract, but the proactive recovery path can restart the same chat before the evicted handler has finished suspending/shutting down.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: working sessions preempted for concurrency could ACK an uncommitted inbox entry, never resume, then be evicted, silently dropping work.
- Design: split suspend semantics by reason, prefer idle preemption victims, proactively request chat-scoped recovery, queue recovery resumes to avoid livelock, raise defaults, and emit explicit limit events.
- Affected modules: `packages/client/src/runtime/session-manager.ts`, client/shared config defaults, session-manager regression tests.

**Review Findings**
❌ 1. `evictIfNeeded` can redeliver a chat before its old handler has actually stopped. When the eviction candidate is active, `handler.shutdown()` is fired and not awaited; when it is a just-preempted suspended session, `entry.suspending` may still be pending. The code then deletes the session entry and immediately schedules proactive recovery. A same-socket `inbox:recover` can therefore redeliver and resume from `evictedMappings` with a fresh handler while the previous handler is still aborting/closing, losing the `resumeSession` guard that normally waits for `entry.suspending`. That breaks the `(agent, chat) = 1 live handler` boundary and can overlap tool execution / provider session use for the same chat. Please delay the proactive recovery/redelivery until the candidate's suspend/shutdown promise settles, or retain a per-chat cleanup promise in the evicted mapping and make recovery resume wait on it. [R4 / packages/client/src/runtime/session-manager.ts:1555]

**Executed**

yuezengwu:
@baixiaohang Finding 1 addressed in 565ef5c6 — took the second option from your suggestion (per-chat cleanup promise retained in the evicted mapping):

- `evictIfNeeded` now captures the candidate's teardown promise — `handler.shutdown()` for an active victim, the in-flight `suspending` promise for a just-preempted suspended one — into the `EvictedMapping` (`cleanup`, runtime-only, never persisted).
- The eviction-resume path in `startNewSession` awaits `evicted.cleanup` before `handler.resume`, so a proactive-recovery redelivery (or any later input — this also closes the pre-existing lazy-resume race) cannot start a fresh handler while the previous one is still aborting. `(agent, chat) = 1 live handler` holds across eviction.
- New regression test: redelivery blocks while the old handler's shutdown gate is held, resumes exactly once after it settles.

Validation: client 1203 tests green, typecheck clean.

baixiaohang (approved):
## Review by codex-assistant - PR #982

Approved after rereview of 565ef5c6. The previous R4 blocker is addressed: eviction now retains the candidate teardown promise and awaits it before recovery resume, preserving the one live handler per agent/chat boundary.

I did not run local checks or tests per PR Review mode; this approval is based on static diff review plus the author-reported validation.

yuezengwu:
按 yuezengwu 的过期项清理要求先关闭：这是超过 2 天未处理的 direct stale 项。后续会逐个验证 / 讨论 / 决策；若仍有效，再重开或新建替代项。

yuezengwu:
关闭后验证结果：保持本 PR 关闭，不建议重开旧分支。

当前 `main` 已经具备 #973 的主要修复形态：working preempt 不 ack、idle victim 优先、recovery debt/恢复路径、99/99 默认 guardrail，以及对应 session-manager 回归测试。因此这个旧 PR 不再是合适的落地载体。

不过 PR review 中 “eviction resume 必须等待旧 handler teardown” 的窄问题在当前 `main` 仍没有看到等价实现；我已单独创建 #1151 跟进。这样旧 PR 保持关闭，后续只修当前代码上的最小残留风险。

PR #489, opened 2026-05-21
fix(context-tree): gate Context feed chat links by access + center cold-load state

Gandy2025:
## Summary

Two changes to the **Context** tab, both verified locally and reviewed by codex-reviewer (no blocking issues).

### 1. Gate Context Usage Feed chat links by access (bug + minor info-leak UX)

The Context Usage Feed shows each event as `<agent> read … in #<chat>`. The trailing `#<chat>` was **always** a clickable button that navigates to `/?c=<chatId>`. The feed exposes chat title/id org-wide on purpose (transparency), but a viewer who is **not a member** of that chat lands on a route the server `404`s (`requireChatAccess`), and the chat view has no error branch — so it **hangs forever on "Loading chat…"**.

Now each event carries a server-computed **`viewerCanAccess`** flag:

- **No DB migration** — it is derived per request, never stored.
- Mirrors `requireChatAccess` exactly: direct `chat_membership` for the caller's human agent (**speaker OR watcher**), or a **speaker the caller manages**.
- Resolved **fail-closed** (no viewer → not clickable) in **two batched queries** bounded by the in-org chat ids — **no N+1**.
- Both snapshot routes pass the viewer (`/orgs/:orgId/snapshot` from `requireOrgMembership` scope; legacy `/snapshot` via a non-throwing `resolveOrgViewer`).
- Frontend: clickable `<button>` when accessible, otherwise **inert text** with a `No access to this chat` tooltip. **Org-wide label visibility and cross-org masking are unchanged.**

The optional manual-URL (`/?c=<id>`) 404 empty-state fallback was intentionally **out of scope** for this PR.

### 2. Polish the Context tab cold-load state

The cold-load `LoadingState` (no cached snapshot) was a static line pinned to the top-left with a non-spinning icon. It now **centers in the content area** and spins the `RefreshCw` icon via the app-wide `animate-spin` utility, using design tokens for spacing/color. No data-logic changes.

## Files

- `packages/shared/src/schemas/context-tree.ts` — add `viewerCanAccess: z.boolean()`
- `packages/server/src/services/session-event.ts` — `viewer` param + `accessibleChatIdSet()` (mirrors `requireChatAccess`)
- `packages/server/src/scope/require-resource.ts` — `resolveOrgViewer()` (non-throwing member resolution)
- `packages/server/src/api/context-tree-snapshot.ts`, `packages/server/src/api/orgs/context-tree-snapshot.ts` — pass viewer
[...]

PR #984, opened 2026-06-11
refactor: simplify server rate limiting

baixiaohang:
## Summary\n- Replace per-route low rate limits with one actor-aware global @fastify/rate-limit cap (agent -> user -> IP)\n- Raise the default global cap to 3000/min while keeping FIRST_TREE_RATE_LIMIT_MAX as the override\n- Remove per-route rate-limit config/env usage from auth, GitHub OAuth, me, Context Tree snapshot, agent message, and GitHub webhook routes\n- Update config/docs/e2e helpers and add focused regression coverage\n\n## Context Tree\n- Policy update opened first: https://github.com/agent-team-foundation/first-tree-context/pull/489\n\n## Verification\n- first-tree-staging tree verify (context tree PR)\n- pnpm exec biome check <changed files>\n- pnpm --filter @first-tree/shared exec vitest run src/config/__tests__/server-config.test.ts --maxWorkers=2 --minWorkers=1\n- VITEST_MAX_FORKS=2 pnpm --filter @first-tree/server exec vitest run src/__tests__/rate-limit.test.ts\n- VITEST_MAX_FORKS=2 pnpm --filter @first-tree/server exec vitest run src/__tests__/auth.test.ts src/__tests__/github-app-webhook-route.test.ts src/__tests__/context-tree-snapshot.test.ts src/__tests__/me-onboarding.test.ts\n- pnpm typecheck\n- pnpm check (exit 0; reports existing Biome advisories in unrelated client/web files)


yuezengwu (approved):
**Approving** — faithfully implements the actor-aware rate-limit policy (context-tree PR #489, co-owned by yuezengwu/baixiaohang).

### What this does
Replaces the per-route low caps (login 5/min, refresh 30, github/start 20, webhook 600, agent-message 30/agent, snapshot 6, onboarding 60) with **one global `@fastify/rate-limit` cap**, keyed `agent.uuid → user.userId → req.ip`, default raised 100 → 3000/min. Removes the four per-route config fields + env vars (`*_LOGIN_MAX` / `*_WEBHOOK_MAX` / `*_AGENT_MESSAGE_MAX` / `*_CONTEXT_TREE_SNAPSHOT_MAX`); only `FIRST_TREE_RATE_LIMIT_MAX` survives.

### Verified
- **Contract match**: key order (agent→user→IP), high coarse cap, no Redis/shared state, all low route defaults gone — all consistent with the policy node's Decision/Constraints.
- **Hook order preserved**: limiter stays `hook: "preHandler"`, so `req.agent`/`req.user` populated by the onRequest auth hooks are readable in `keyGenerator`. Correct.
- **No stragglers** at PR head: the only remaining `*_LOGIN_MAX` etc. references are in the new *negative* config test asserting they're ignored. No per-route `rateLimit` config left in `api/`.
- New `rate-limit.test.ts` covers user-id keying, agent-id-before-manager keying, IP fallback, and that the removed 30/min + 6/min route caps no longer apply. Good coverage.

### Non-blocking, for awareness
- **CodeQL** raised 3 `js/missing-rate-limiting` alerts (`auth.ts`, `auth/github.ts`, `webhooks/github-app.ts`). These are heuristic — the static analyzer matches per-route `rateLimit` config and doesn't see the global limiter, which **does** still cover these routes. They're expected given this design; suggest dismissing as "won't fix (by design)" rather than re-adding route caps. Owner's call.
- **Security trade-off** worth a conscious ack: unauthenticated `/auth/login` brute-force surface goes from 5/min/IP to the global 3000/min/IP backstop. This is exactly what the policy chose ("design stronger abuse controls from evidence rather than reintroducing low route defaults"), so it's signed off — flagging only for visibility.

CI is still running at review time (Lint/Type/Test queued) — approval assumes it goes green before merge.

yuezengwu:
Re-checked the `synchronize` delta (force-push `00b72bf`): purely additive — `// codeql[js/missing-rate-limiting]` inline suppressions added to all three flagged routes (`auth.ts`, `auth/github.ts`, `webhooks/github-app.ts`) with the correct rationale (global actor-aware limiter registered in app.ts as a preHandler). No logic change; this resolves the CodeQL item in-code. Approval stands — still gated on CI green before merge.

PR #989, opened 2026-06-11
fix(shared): unwrap login-shell wrappers in shell IO classifier

liuchao-001:
## Summary

- The Web Console Context tab dashboard never shows any codex runtime activity because every codex shell command lands as `/bin/<login-shell> -lc '<inner>'` (bash on Linux daemons, zsh on macOS daemons), and `classifyShellCommandIo` treats the outer shell as a mutating tool and rejects the whole event before the inner read is ever inspected.
- The rejection happens in **both** the client (file-ref extraction at emit time) and the server (`shellToolCanRead` at record time) because both call into the same shared classifier — so events reach the DB with no `toolFileRefs`, and no `context_tree_io_events` row ever gets derived.
- This PR adds a shell-wrapper unwrap step to `classifyShellCommandIo` so `/bin/zsh -lc "sed -n '1,7p' .../NODE.md"` re-classifies through the same function as `sed -n '1,7p' .../NODE.md` would, while every other rejection (wrapped `tee`, wrapped pipeline, wrapped `sed -i`, wrapped `$VAR`) still surfaces correctly.

## Empirical evidence

Pulled live from a codex `session_events` row:

```
payload.args.command = '/bin/zsh -lc "sed -n \'1,7p\' /Users/op/.../NODE.md"'
classifyShellCommandIo(...) →
  { supported: false, reason: "write_or_mutation", commandName: "zsh" }
payload.toolFileRefs = undefined   // ← client refused to attach refs
```

After this fix the same input becomes:

```
{ supported: true, action: "read", commandName: "sed",
  pathArgs: [{ raw: "/Users/op/.../NODE.md", pathKindHint: "file" }] }
```

## Why one shared change, not split client/server

- Client `toolFileRefsFromShellCommand()` calls `classifyShellCommandIo` directly — without the unwrap the emitted event has no `toolFileRefs`.
- Server `recordFromSessionEvent` / `shellToolCanRead` also calls `classifyShellCommandIo` — even if it could otherwise infer the inner command, the event payload it sees has no refs to attach.
- Fixing only one side leaves the other one broken. Fixing the shared classifier fixes both ends in one place, and is automatically provider-neutral.

## What this PR does NOT do

- **Does not backfill historical events.** Past `session_events` rows already landed without `toolFileRefs`; the server has no agent-local tree checkout to reconstruct repo-relative paths from. The Context tab catches up as new codex activity accumulates after this lands.
- Does not change the codex handler. The wire format stays exactly the same; only the classifier's interpretation of it changes.
- Does not relax the rejection envelope. Wrapped `tee`, wrapped `sed -i`, wrapped `cat … | head`, and wrapped `cat \$VAR/NODE.md` all still reject (with the inner reason now surfaced as the rejection reason).

## Implementation notes

- Wrapper basename set: `{sh, bash, zsh}` — the three shells that show up as the daemon login shell in this codebase's runtime surface.
- Wrapper flag set: `{-c, -lc}`. `-lc` is what codex / login-shell daemons emit; `-c` is the POSIX form so we accept both.
- `MAX_WRAPPER_UNWRAPS = 2` bounds recursion. One level covers codex; two leaves headroom for a script-wrapped command. Anything deeper falls through to the standard mutating-shell rejection — covered by a regression test.
- The unwrap is placed *before* `MUTATING_OR_AMBIGUOUS_TOOLS` so a wrapped read survives the mutating-shell check; the inner command then goes through the full classify pipeline so its own rejection rules still apply.

## Test plan

- [x] `pnpm --filter @first-tree/shared test src/__tests__/shell-command-io.test.ts` — 23 tests pass (11 new wrapper cases + 12 pre-existing)
- [x] `pnpm --filter @first-tree/client test src/__tests__/codex-context-tree-io.test.ts` — 6 tests pass (2 new wrapped-command end-to-end fixtures + 4 pre-existing)
- [x] `pnpm --filter @first-tree/server test` — full 1350-test server suite green (covers `recordFromSessionEvent` end-to-end)
- [x] `pnpm check` and `pnpm typecheck` clean
- [ ] After merge: liuchao-codex agent runs a fresh `sed -n` against its tree checkout; the resulting `session_events` row should now carry `toolFileRefs`, and the Context tab Recent Events feed should pick it up within one snapshot-refresh cycle.

### DIFF
--- a/packages/client/src/__tests__/codex-context-tree-io.test.ts
+++ b/packages/client/src/__tests__/codex-context-tree-io.test.ts
@@ -71,6 +71,55 @@ describe("Codex Context Tree file refs", () => {
     ]);
   });
 
+  // Codex CLI does not surface raw `sed` / `cat` — every command_execution
+  // lands as `/bin/<login-shell> -lc '<inner>'` (zsh on macOS, bash on Linux).
+  // This is the real wire shape pulled from a session_events row during the
+  // fix investigation. Without the shared `classifyShellCommandIo` wrapper
+  // unwrap, the outer `zsh` / `bash` falls into MUTATING_OR_AMBIGUOUS_TOOLS
+  // and the client never attaches file refs, so the Context tab dashboard
+  // records nothing for codex agents. Lock the wrapped form down end-to-end.
+  it("emits shell read refs for Codex's /bin/zsh -lc 'sed ...' wrapper (macOS form)", () => {
+    const refs = toolFileRefsFromShellCommand({
+      command: "/bin/zsh -lc \"sed -n '1,7p' /home/op/context-tree/NODE.md\"",
+      cwd: "/home/op/source",
+      contextTreePath: "/home/op/context-tree",
+      contextTreeRepoUrl: "https://github.com/acme/first-tree-context.git",
+      contextTreeBranch: "main",
+    });
+
+    expect(refs).toEqual([
+      {
+        origin: "tool_arg",
+        localPath: "/home/op/context-tree/NODE.md",
+        repoUrl: "https://github.com/acme/first-tree-context.git",
+        repoBranch: "main",
+        repoRelativePath: "NODE.md",
+        pathKind: "file",
+      },
+    ]);
+  });
+
+  it("emits shell read refs for Codex's /bin/bash -lc 'cat ...' wrapper (Linux form)", () => {
+    const refs = toolFileRefsFromShellCommand({
+      command: "/bin/bash -lc 'cat /home/op/context-tree/NODE.md'",
+      cwd: "/home/op/source",
+      contextTreePath: "/home/op/context-tree",
+      contextTreeRepoUrl: "https://github.com/acme/first-tree-context.git",
+      contextTreeBranch: "main",
+    });
+
+    expect(refs).toEqual([
+      {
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
Reviewed the shared classifier change and the client/server Context Tree IO paths.

The goal is clear: Codex emits shell reads as `/bin/<login-shell> -lc '<inner>'`, and the previous classifier rejected the outer shell before file refs could be attached. This PR fixes that at the shared classifier layer, so the client emits `toolFileRefs` and the server's existing `shellToolCanRead` gate sees the same read classification.

I did not find blockers. The unwrap is bounded, only accepts `sh`/`bash`/`zsh` with `-c`/`-lc`, re-runs the inner command through the existing parser, and keeps wrapped mutations, pipelines, in-place `sed`, and dynamic paths rejected. No core data structure or DB schema change is involved.

Local verification:
- `pnpm --filter @first-tree/shared test src/__tests__/shell-command-io.test.ts` passed (23 tests)
- `pnpm --filter @first-tree/client test src/__tests__/codex-context-tree-io.test.ts` passed (6 tests)
- `pnpm typecheck` passed
- `pnpm check` exited 0; it reported only unrelated existing Biome warnings outside this PR

PR #997, opened 2026-06-12
fix(cli): retire stranded legacy github-scan launchd runner on daemon start

### LINKED ISSUE #995: CLI upgrade strands legacy github-scan launchd runner: KeepAlive crash-loop + port squatting
## What happened

## Why it mattered

## Local remediation applied

## Asks

---

serenakeyitan:
## What

Closes #995 (ask #1): on machines that ran an older build, the now-removed
`github-scan` subsystem left a `KeepAlive` launchd runner
(`com.first-tree.github-scan.runner.<user>.default`). After a CLI upgrade
relocates/removes its binary, launchd hits `MODULE_NOT_FOUND` and crash-loops
the job indefinitely — and while it loops it keeps binding the legacy default
`httpPort` **7878**, silently squatting a port another local tool (tdoc)
documents as its own.

Nothing in the current upgrade path retires this job. The `v1-orphan-skills`
workspace migration only sweeps retired skill *payloads*; the runner is a
**machine/user-level launchd job under the prod home**, not a workspace artifact.

## Change

A focused, best-effort `retireLegacyGithubScanLaunchd()` sweep, wired into
`daemon start` next to the existing client-level migrations
(`migrateLocalAgentDirs`, `reconcileLocalRuntimeProviders`). Running it on
daemon start (rather than only in the `upgrade` command) means already-upgraded
machines get cleaned on their next start.

It:
- enumerates `*.plist` under `~/.first-tree/github-scan/runner/launchd/`
  (**enumerate-from-disk** — avoids reconstructing the username/profile-bearing label),
- reads each plist's recorded `Label` and `launchctl bootout gui/<uid>/<label>`s it
  (swallowing the benign "not loaded" case),
- deletes the plist and prunes the now-empty dir.

Deliberately narrow + safe:
- **darwin-only**; no-op elsewhere.
- **Free** on machines that never ran github-scan — short-circuits when the plist dir is absent.
- **Best-effort** — never throws out, never blocks daemon startup.
- Leaves the rest of `~/.first-tree/github-scan/` (config.yaml, logs) untouched — inert files don't crash-loop, and a user may have written a defensive config there.

## Tests

New `apps/cli/src/__tests__/retire-github-scan-launchd.test.ts` (8 cases): absent-dir no-op, bootout+remove+prune, Label-over-filename, filename-stem fallback, benign "no such process" swallowed, multiple plists, non-plist file keeps the dir, and off-darwin no-op.

- `pnpm --filter first-tree exec vitest run` — full CLI suite green (444 passed).
- `tsc --noEmit` — clean. `biome check` — clean on changed files.

## Scope note

This PR covers **ask #1** only. Ask #2 (harden the *active* daemon service against a bare `nvm` switch — no StartLimit backoff / re-resolve the bin at runtime) and ask #3 (move `httpPort` off 7878 *if* github-scan is ever revived) are out of scope here and can be tracked separately.

### DIFF
--- a/apps/cli/src/__tests__/retire-github-scan-launchd.test.ts
+++ b/apps/cli/src/__tests__/retire-github-scan-launchd.test.ts
@@ -1,0 +1,165 @@
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { retireLegacyGithubScanLaunchd } from "../core/retire-github-scan-launchd.js";
+
+const spawnSyncMock = vi.hoisted(() => vi.fn(() => ({ status: 0, stdout: "", stderr: "" })));
+const userInfoMock = vi.hoisted(() => vi.fn(() => ({ uid: 501, username: "gandy" })));
+
+vi.mock("node:child_process", () => ({
+  spawnSync: spawnSyncMock,
+}));
+
+vi.mock("node:os", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:os")>();
+  return { ...actual, userInfo: userInfoMock };
+});
+
+const originalPlatform = process.platform;
+
+function setPlatform(platform: NodeJS.Platform): void {
+  Object.defineProperty(process, "platform", { configurable: true, value: platform });
+}
+
+function plistBody(label: string): string {
+  return `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTD/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>Label</key>
+  <string>${label}</string>
+  <key>KeepAlive</key>
+  <true/>
+</dict>
+</plist>
+`;
+}
+
+let home: string;
+
+function launchdDir(): string {
+  return join(home, "github-scan", "runner", "launchd");
+}
+
+function writePlist(fileName: string, label: string): string {
+  const dir = launchdDir();
+  mkdirSync(dir, { recursive: true });
+  const path = join(dir, fileName);
+  writeFileSync(path, plistBody(label));
+  return path;
+}
+
+beforeEach(() => {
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
This PR adds a narrow daemon-start cleanup for the removed `github-scan` launchd runner: enumerate legacy plists from the old prod home, boot out the recorded launchd labels, remove only those plist files, and keep daemon startup best-effort.

I checked the old github-scan launchd implementation against the new cleanup path and label shape, and the scope matches the stranded runner this is meant to retire. I did not find any blocker.

Verified locally:

```bash
pnpm install --frozen-lockfile
pnpm --filter first-tree-dev exec vitest run src/__tests__/retire-github-scan-launchd.test.ts src/__tests__/daemon-start-command.test.ts
```

Both targeted test files passed, 20 tests total.

yuezengwu:
Stale-queue review against current `main`:

**Goal / core change:** best-effort Darwin cleanup of a stranded legacy `github-scan` launchd runner during daemon startup, without touching unrelated legacy files. Current `main` has no equivalent cleanup and issue #995 remains open, so the user-facing problem was not superseded.

**Current gate:** the branch conflicts with `main`, and daemon startup/service handling has changed substantially since this head. The prior approval applies to the reviewed old head, not to a future conflict resolution. The safe next step is a fresh, narrow current-main patch (or author refresh) that revalidates the legacy paths/labels and startup placement, retains the no-op/off-Darwin/best-effort guarantees, and reruns the focused CLI tests. No database or core data-structure change is involved. Leaving open for refresh rather than closing valid work.

yuezengwu:
@serenakeyitan This still tracks a valid, narrow daemon-start cleanup, but the branch now conflicts with current `main`.

**Recommendation:** close this stale PR and rebuild the cleanup as a fresh, minimal patch from current `main` under #995, revalidating it against the current daemon startup path. Please reply here to confirm whether you will close and rebuild, or explain why this branch should remain open.

yuezengwu:
Final stale-queue disposition: the follow-up window elapsed with no author reply or head update. Per maintainer direction, I am closing this stale implementation PR without merging it or deleting its branch.

This closure does **not** mark the underlying problem as fixed; #995 remains open. The existing approval applies only to the old head `b2a2f5b9` and must not be carried over to a replacement. Any renewed patch should start from current `main`, revalidate the Darwin legacy paths and launchd labels against the current daemon startup flow, and rerun the focused CLI tests.

PR #988, opened 2026-06-11
fix(server,web): rest the GitHub App install bind on the kickoff session, not the callback OAuth identity

Gandy2025:
## Incident

Prod onboarding: a user completed the GitHub App install (installation `139563599`, GitHub org `ShibaOrg`), but the OAuth callback refused with a raw-JSON 403 **"Not an admin of the organization this installation targets"** and the installation row sat unbound (`hub_organization_id = NULL`, confirmed in prod).

## Root cause

The callback re-resolved the acting user **from the OAuth code** (`findOrCreateUserFromGithub`) and ran the target-org admin re-check against that identity. But the browser's github.com session is independent of the First Tree session — a second GitHub account, a deleted-and-recreated account (numeric ID changes), or another person's First Tree session in the same browser all make the two identities legitimately diverge. The kickoff (`GET /orgs/:orgId/github-app-installation/install-url`) is admin-gated, so the kickoff user's authority was already proven — the callback just wasn't using it.

## Fix (4 defects)

1. **Bind authority = kickoff session.** `/install-url` signs `kickoffUserId` into the state alongside `targetOrganizationId`; the callback re-checks **that** admin live against `members` and binds on their authority. On identity mismatch it completes the bind and redirects to `next` **without issuing tokens** — signing the browser in as the foreign identity would replace the kickoff admin's session in every tab. GitHub-side authority is unchanged: `installationId` is only non-null when the OAuth user proved they administer the installed account (#423), and the nonce cookie pins the callback to the kickoff browser. Legacy states without `kickoffUserId` (≤10 min old at deploy time) keep the old behavior.
2. **No more raw-JSON dead ends.** Every browser-facing `/callback` error 302s to `/auth/github/complete#error=<code>` where the SPA renders friendly copy + a way back (codes: `state-expired`, `github-exchange-failed`, `install-not-admin`, `invite-invalid`, `invite-not-allowed`, `invite-required`, `membership-unresolved`). The most common real-world trigger — spending >10 min on GitHub's repo picker and expiring the 10-minute state JWT — now lands on a readable page. `dev-callback` keeps JSON errors (curl-able, asserted by dev tests).
3. **Observability.** The refusal path logs structured warns with both identities (`github_app.install_callback_admin_check_failed` / `github_app.install_callback_identity_mismatch`) — this incident was undiagnosable from logs (the 403 path logged nothing).
4. **Copy.** The refusal names the First Tree team; "organization" read as the GitHub org to everyone who saw it.

## Tests (red → green)

[...]

PR #1003, opened 2026-06-12
fix(server): allowlist /build-tree as a post-install next

Gandy2025:
## Problem

The `/build-tree` recovery surface's connect-code step passes `next=/build-tree` when the GitHub App install popup is blocked (it falls back to a full-page redirect and must return to the recovery flow — see `step-connect-code.tsx`):

```ts
const postInstallNext = installTab ? "/onboarding/connected" : recovery ? "/build-tree" : "/onboarding";
```

But `ALLOWED_POST_INSTALL_NEXT` never included `/build-tree`, so `resolvePostInstallNext` silently rewrote the destination to `/settings/github`. A recovery admin whose popup was blocked completed the install and landed on Settings, outside the flow they were in.

## Fix

Add `/build-tree` to the allowlist (an internal SPA path, same trust class as `/onboarding`), and pin it in `post-install-next.test.ts` alongside the exact-allowlist assertion.

## Context

Found while working on the install-callback dead-end UX; the rest of that work was superseded by #988 (which lands the bind on the kickoff session and routes all browser-facing callback errors to `/auth/github/complete#error=<code>`). This allowlist gap is the one piece #988 didn't cover.

## Test plan

- `pnpm --filter @first-tree/server exec vitest run src/api/orgs/__tests__/post-install-next.test.ts src/__tests__/github-app-install-url.test.ts src/__tests__/github-app-callback-target-org.test.ts` — 18/18 green
- `pnpm check` + `pnpm --filter @first-tree/server typecheck` — clean

### DIFF
--- a/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
+++ b/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
@@ -11,6 +11,10 @@ describe("resolvePostInstallNext", () => {
     expect(resolvePostInstallNext("/onboarding")).toBe("/onboarding");
     expect(resolvePostInstallNext("/onboarding/connected")).toBe("/onboarding/connected");
     expect(resolvePostInstallNext("/settings/github")).toBe("/settings/github");
+    // The build-tree recovery surface passes itself as `next` when the install
+    // popup is blocked — rewriting it to Settings would strand the recovery
+    // admin outside the flow they were in.
+    expect(resolvePostInstallNext("/build-tree")).toBe("/build-tree");
   });
 
   it("falls back to Settings for an absent value", () => {
@@ -26,6 +30,11 @@ describe("resolvePostInstallNext", () => {
   });
 
   it("only allows the known internal destinations", () => {
-    expect([...ALLOWED_POST_INSTALL_NEXT].sort()).toEqual(["/onboarding", "/onboarding/connected", "/settings/github"]);
+    expect([...ALLOWED_POST_INSTALL_NEXT].sort()).toEqual([
+      "/build-tree",
+      "/onboarding",
+      "/onboarding/connected",
+      "/settings/github",
+    ]);
   });
 });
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1003

**Recommendation**: approve
- Reason: The change restores the blocked-popup recovery path while keeping `next` constrained to exact internal SPA destinations.

**Level**: B-low
- Path initial: `packages/server/**` only; no migrations, Prisma schema, or listed core services touched.
- Semantic lift: none.

**PR Summary**
- Problem solved: GitHub App install fallback from `/build-tree` was redirected to Settings because `/build-tree` was missing from the post-install allowlist.
- Design: Add `/build-tree` as an exact allowlisted internal destination and pin the exact allowlist in the focused unit test.
- Affected modules: server org GitHub App install URL handling and its post-install `next` tests.

**Review Notes**
- No blocking issues found. `/build-tree` is an internal route already used by the recovery connect-code flow, the server still signs only allowlisted `next` values into OAuth state, and the test continues to assert the complete allowlist plus off-list fallback behavior.

**Executed**
- Approved this PR.
- Did not run tests, per review protocol; relied on code review and the author's reported test plan.

yuezengwu (approved):
Reviewed the goal: keep a popup-blocked `/build-tree` recovery admin inside the recovery flow after GitHub App install, instead of silently falling back to Settings.

Core change is narrow and matches the call chain: `StepConnectCode` passes exact `/build-tree` only for recovery full-page redirects, and `resolvePostInstallNext` now allowlists that exact internal SPA path while continuing to reject arbitrary/off-list destinations. The test pins both pass-through behavior and the complete allowlist.

No core data structure or database changes. No blockers found.

Verified locally:
`pnpm --filter @first-tree/server exec vitest run src/api/orgs/__tests__/post-install-next.test.ts src/__tests__/github-app-install-url.test.ts src/__tests__/github-app-callback-target-org.test.ts` → 18/18 passed.

CI is green on the PR head.

PR #1007, opened 2026-06-12
fix(client,server): canonicalize Context Tree ref paths, cover notebook and search tools

Gandy2025:
## Summary

Two systematic under-counting gaps in Context Tree IO attribution (the data behind the Context tab's read/write feed):

### 1. Symlink alias mismatch — the biggest silent miss

The W1 cloud layout exposes the shared external tree clone inside each agent home as a `<workspace>/context-tree` **symlink** (`runtime/workspace-manifest.ts`), while the runtime binding carries the external clone's **real path** (`context-tree-repos/<digest>`). Ref attribution was a pure string prefix match between the two spellings, which never agrees — so every read/write that traveled through the symlinked path (the path the shipped `first-tree` skill's pre-task hygiene points agents at) silently lost its repo evidence and never reached the feed.

Fix: a `canonicalizeFsPath` helper (realpath with lexical fallback for not-yet-existing segments) applied to **both sides** of the containment check, in all three ref extractors:
- Claude single-file tools (`treeNodePathOf` call site)
- shell command refs (`toolFileRefsFromShellCommand`)
- Codex `file_change` (`contextTreeTargetPathOf`)

`localPath` keeps the tool's original spelling for evidence fidelity; relative tool paths keep the fail-safe null mapping (canonicalizing them against the daemon cwd would risk mis-attribution).

### 2. Tool coverage gaps

- `NotebookEdit` joins `CLAUDE_WRITE_TOOLS` (client + server), and notebook tools' `notebook_path` argument is recognized alongside `file_path` — previously `NotebookRead` was in the read set but could never produce a ref.
- `Grep` / `Glob` count as reads at **directory granularity**: the client emits one ref for the explicit `path` argument (search root), never one ref per matched file, so a recursive search cannot flood the feed. Calls without an explicit `path` carry no ref (fail-safe). This also removes the inconsistency where shell `grep`/`rg` counted as tree reads but the equivalent Grep tool did not.

## Validation

- `pnpm check` / `pnpm typecheck` clean
- client 1216/1216, server 1382/1382, shared 422/422
- New tests: symlinked-tree mapping in both directions (incl. Write of a not-yet-existing node), NotebookEdit via `notebook_path`, Grep/Glob directory- and repo-level refs, no-path fail-safe, Bash read through the symlink, server-side derivation for Grep/Glob/NotebookEdit

### DIFF
--- a/packages/client/src/__tests__/claude-code-tool-events.test.ts
+++ b/packages/client/src/__tests__/claude-code-tool-events.test.ts
@@ -1,5 +1,8 @@
+import { mkdirSync, mkdtempSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import type { SessionEvent } from "@first-tree/shared";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createToolCallProcessor, treeNodePathOf } from "../handlers/claude-code.js";
 
 /**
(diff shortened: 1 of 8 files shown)


yuezengwu (approved):
Reviewed the Context Tree IO attribution changes.

Goal/core change: this fixes repo evidence loss when the workspace `context-tree` symlink and the runtime binding's real clone path use different filesystem spellings, and expands attribution coverage for notebook/search tools without turning recursive search results into per-file feed noise.

Checked the main data flow across:
- client canonical path containment for Claude single-file tools, shell refs, and Codex `file_change`
- Claude SDK/TUI shared tool-call processor behavior for NotebookEdit, Grep, and Glob
- server-side derivation for the widened Claude read/write tool sets
- Context Tree tab boundary from the tree: aggregate/server-managed usage evidence, no browser authoring or per-node usage claim expansion

No database/schema migration or core data-structure change requiring human inspection found.

Validation run locally in the PR worktree:
- `pnpm --filter @first-tree/client test src/__tests__/context-tree-file-refs.test.ts src/__tests__/codex-context-tree-io.test.ts src/__tests__/claude-code-tool-events.test.ts` — 60 passed
- `pnpm --filter @first-tree/server test src/__tests__/context-tree-io.test.ts` — 7 passed

PR #983, opened 2026-06-11
refactor(server,cli)!: remove client claim endpoint; login --override rotates local identity

yuezengwu:
## Decision

**Owner's call (yuezengwu):** organization members are mutually trusted, so a server-side ownership-transfer protocol is not worth securing. Between the two options — add possession proof to `claim`, or remove `claim` — we remove it. The attack surface goes to zero and the system model gets simpler (no transfer protocol to reason about), at the cost of a slightly less polished machine-handover path (which the `CLIENT_ORG_MISMATCH` flow already models).

## The problem this closes

`POST /api/v1/clients/:clientId/claim` transferred `clients.user_id` to the caller and unpinned the previous owner's agents, gated **only** by a valid user JWT — no proof the caller controls the machine. But `clientId` is not a secret: `GET /orgs/:orgId/agents` returns it to every member. So any member could read a victim's `clientId` and `POST .../claim` to seize the client and force every one of the victim's agents offline (the victim's daemon then hits `CLIENT_USER_MISMATCH`/4403 and stops reconnecting until a manual `--override`). Found and adversarially verified (HIGH) in the 2026-06-11 full-repo audit.

## What changed

- **Removed** the `POST /clients/:clientId/claim` route and the `claimClient` service function (+ `claimClientResponseSchema`, `ClaimClientResponse`).
- **`login --override`** no longer calls the server transfer. It now rotates the machine's local client identity via the existing `rotateClientIdWithBackup` (backs up `client.yaml`) and registers a **fresh** `clientId` under the new account. Net effect for the operator is the same — agents must be re-added — and the previous account's client row + pinned agents are left untouched server-side (they show offline until that account removes them).
- Renamed `cleanupStaleAliasesAfterClaim` → `cleanupStaleLocalAliases` (same logic, now triggered by identity rotation rather than a claim).
- Updated `CLIENT_USER_MISMATCH` guidance, docs (`onboarding-guide`, `cli-reference`, `CLAUDE.md`, e2e README) to describe the rotate-and-re-register handover.
- Tests: deleted the claim unit + e2e tests; added `client-user-mismatch.test.ts` asserting the route is gone (404) and the WS handshake still refuses a non-owner register (4403, no silent overwrite); rewrote the `login --override` tests to assert local identity rotation.

## Trade-off accepted

[...]

PR #1011, opened 2026-06-12
fix(server): enforce filter integrity on org session listing

yuezengwu:
## Context

Follow-up from the security audit walkthrough (same review batch as #983). The audit flagged that `GET /orgs/:orgId/sessions` exposes every org session's chat topic and first-message summary to any active member.

**Decision: visibility stays as-is.** Org members are mutually trusted; topic/summary of org sessions is information the membership permission is allowed to read. What must be guaranteed instead is that every query filter on this listing actually takes effect. Auditing that filter chain found three real defects, fixed here.

### 1. The tenant boundary was an optional filter

`listAllSessions(db, limit, cursor?, filters?)` carried `organizationId` inside the optional `filters` object. A caller that forgot to pass it got a silently unscoped, cross-tenant query. The only current caller did pass it, but a tenant boundary must not depend on caller discipline.

**Fix:** `organizationId` is now a required positional parameter — omitting it is a compile error.

### 2. Cursor pagination never worked (pre-existing 500)

The cursor condition was built as a raw fragment: ``sql`${agentChatSessions.updatedAt} < ${new Date(cursor)}` ``. Parameters inside a raw `sql` fragment bypass the column's driver mapping, and postgres-js rejects native `Date` instances (`TypeError: The "string" argument must be of type string...`). Every page-2 request on this endpoint returned 500 — the new pagination test reproduced this faithfully before the fix.

**Fix:** use the `lt()` comparator so the Date goes through the column type's driver mapping. Swept the repo for the same pattern; this was the only raw-fragment comparison passing a `Date` (the `inbox.ts` fragments pass strings/numbers, which are safe).

### 3. Malformed cursor surfaced as 500

`new Date("garbage")` produces an Invalid Date that previously went straight into the query. Now validated: a malformed cursor returns `400 BadRequestError`.

`agentId` and `state` filters were verified correct as-is (ANDed with the org condition; a foreign agentId yields an empty set, not a leak).

## Tests

New `org-sessions-list.test.ts` (this route previously had zero coverage):

- org isolation: foreign org's sessions never appear in the listing; requesting a foreign `:orgId` is 403
- `agentId` filter narrows to that agent's sessions
- `state` filter takes effect; evicted sessions hidden by default, visible only via `state=evicted`
- cursor pagination advances without overlap and terminates; malformed cursor and invalid state enum return 400

`pnpm check`, `pnpm typecheck`, and the full server suite (161 files / 1382 tests) pass.

### DIFF
--- a/packages/server/src/__tests__/org-sessions-list.test.ts
+++ b/packages/server/src/__tests__/org-sessions-list.test.ts
@@ -107,6 +107,35 @@ describe("GET /orgs/:orgId/sessions — filter integrity", () => {
     expect(foreign.statusCode).toBe(403);
   });
 
+  it("excludes a session row that points an own-org agent at a foreign-org chat", async () => {
+    const app = getApp();
+    const admin = await createAdminContext(app);
+
+    // agent_chat_sessions has independent FKs to agents and chats; nothing at
+    // the DB layer ties their orgs together. Simulate a stale/malicious
+    // client having reported session:state for a foreign chatId.
+    const foreignOrgId = `org-${randomUUID().slice(0, 6)}`;
+    await app.db.insert(organizations).values({
+      id: foreignOrgId,
+      name: foreignOrgId.slice(0, 30),
+      displayName: "Foreign Org",
+    });
+    const foreignChatId = await seedChat(app, foreignOrgId, "foreign secret topic");
+
+    const ownAgent = await createAgent(app.db, {
+      name: `xchat-${randomUUID().slice(0, 6)}`,
+      type: "agent",
+      displayName: "Cross-chat Agent",
+      managerId: admin.memberId,
+      clientId: admin.clientId,
+    });
+    await seedSession(app, ownAgent.uuid, foreignChatId);
+
+    const res = await listSessions(app, admin.accessToken, admin.organizationId, `?agentId=${ownAgent.uuid}&limit=100`);
+    expect(res.statusCode).toBe(200);
+    expect(res.body.items).toEqual([]);
+  });
+
   it("agentId filter returns only that agent's sessions", async () => {
     const app = getApp();
     const admin = await createAdminContext(app);
--- a/packages/server/src/__tests__/org-sessions-list.test.ts
+++ b/packages/server/src/__tests__/org-sessions-list.test.ts
@@ -1,0 +1,204 @@
+import { randomUUID } from "node:crypto";
+import type { FastifyInstance } from "fastify";
+import { describe, expect, it } from "vitest";
+import { agentChatSessions } from "../db/schema/agent-chat-sessions.js";
+import { agents } from "../db/schema/agents.js";
+import { chats } from "../db/schema/chats.js";
+import { organizations } from "../db/schema/organizations.js";
+import { createAgent } from "../services/agent.js";
+import { createAdminContext, useTestApp } from "./helpers.js";
+
+/**
+ * Filter integrity for `GET /orgs/:orgId/sessions` (listAllSessions).
+ *
+ * Visibility policy: any active org member may read every org session's
+ * topic/summary — that is accepted product behavior. What these tests pin
+ * down is that every query filter actually takes effect:
+ *
+ *   - the organization boundary cannot be crossed (neither via the URL's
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1011

- Reason: the main direction is right, but the tenant boundary still applies only to the agent side of the join while this endpoint returns chat-owned data.

**Level**: B-high
- Path initial: touches `packages/server/src/services/session.ts` under `packages/server/**` session service -> B-high.
- Semantic lift: none; path classification already reaches B-high.

**PR Summary**
- Solves: hardens `GET /orgs/:orgId/sessions` after the security audit by making the org boundary required, fixing cursor pagination, and returning 400 for malformed cursors.
- Design: passes `organizationId` as a required `listAllSessions` argument, builds the cursor predicate with Drizzle `lt()`, validates cursor dates, and adds route coverage for org isolation, filters, pagination, and invalid input.
- Affected modules: server org sessions route, session service listing query, and new server tests.

**Review Comments**
❌ 1. The listing still does not fail closed on the chat side of the tenant boundary. [R4/R5 / `packages/server/src/services/session.ts:233`]

`listAllSessions` now requires `organizationId`, but the only tenant predicate is `eq(agents.organizationId, organizationId)`. The rows returned by this query also include `chats.topic` and first-message summary, and `agent_chat_sessions` has independent FKs to `agents` and `chats` with no database constraint tying their orgs together. A bound agent can report `session:state` for any non-empty `chatId`; the WS path checks that the agent is bound, but does not verify that the reported chat belongs to the same org before `upsertSessionState` inserts the `(agent, chat)` projection. If such a malformed row exists, this endpoint will list it for the agent's org and expose the foreign chat's topic/summary because the join to `chats` is unconstrained.

Please make the listing query enforce the boundary on every tenant-owned table it exposes, e.g. also require `eq(chats.organizationId, organizationId)`, and add a regression test with an own-org agent session pointing at a foreign-org chat to prove that the chat data is not returned. Given the stated security goal, the read-side filter chain should be fail-closed even if a session projection row was created by a stale or malicious client path.

**Executed**
- Did not run local checks or tests per PR Review mode; this is based on static diff/source review plus the PR author's reported test results.

yuezengwu:
Addressed in c46525cf — valid catch, thank you.

**Comment 1 (chat-side tenant boundary):** the listing now filters on `chats.organization_id = :orgId` in addition to the agent-side predicate, so the read path fails closed even if a session projection row was created pointing at a foreign-org chat. Added the requested regression test: an own-org agent session seeded against a foreign-org chat returns an empty listing (the foreign chat's topic/summary never appear).

Full server suite passes (161 files / 1383 tests).

baixiaohang (approved):
## Review by codex-assistant - PR #1011 final

**Recommendation**: approve
- Reason: latest head `c46525cf` resolves the previously blocking chat-side tenant-boundary gap, and the route/service/tests now match the filter-integrity goal.

**Level**: B-high
- Path initial: touches `packages/server/src/services/session.ts` under `packages/server/**` session service -> B-high.
- Semantic lift: none.

**PR Summary**
- Solves: hardens `GET /orgs/:orgId/sessions` so the org session listing keeps the accepted visibility policy while ensuring every query filter, especially the tenant boundary, takes effect.
- Design: makes `organizationId` a required `listAllSessions` argument, scopes the listing on both `agents.organization_id` and `chats.organization_id`, uses Drizzle `lt()` for cursor Date mapping, validates malformed cursors as 400, and covers org isolation/filter/pagination cases.
- Affected modules: server org sessions route, session service listing query, and server tests.

**Review Notes**
✅ 1. The previous blocker is fixed: the read-side tenant boundary now applies to both tenant-owned tables whose data the endpoint exposes, so a malformed `(own agent, foreign chat)` session projection does not leak the foreign chat topic/summary.
✅ 2. The requested regression test is present: an own-org agent session seeded against a foreign-org chat returns an empty listing.
✅ 3. Cursor handling and invalid cursor behavior are now fail-specific, and `organizationId` omission is no longer a caller-discipline issue.

**Executed**
- Did not run local checks or tests per PR Review mode. Remote GitHub checks are green/pass or skipped, including Test, Lint & Type Check, Migrations, and CodeQL.

PR #1013, opened 2026-06-12
fix: enforce org scope for member mutations

baixiaohang:
## Summary
- Pass the resolved org admin scope into member deletion and reject cross-org targets before any delete-side effects run.
- Move member update target lookup/org validation ahead of the empty-patch early return so `PATCH /orgs/:orgId/members/:id` cannot leak another org's member details.
- Add cross-org regression coverage for empty PATCH and DELETE member mutations.

## Tests
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/members.test.ts --maxWorkers=2`
- `pnpm check` (passes; existing Biome warnings are outside this change)
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/packages/server/src/__tests__/members.test.ts
+++ b/packages/server/src/__tests__/members.test.ts
@@ -1,5 +1,11 @@
+import { randomUUID } from "node:crypto";
+import { eq } from "drizzle-orm";
 import type { FastifyInstance } from "fastify";
 import { describe, expect, it } from "vitest";
+import { agents as agentsTable } from "../db/schema/agents.js";
+import { members as membersTable } from "../db/schema/members.js";
+import * as memberService from "../services/member.js";
+import { createOrganization } from "../services/organization.js";
 import { createTestAdmin, useTestApp } from "./helpers.js";
 
 describe("Members API", () => {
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
实现目标：修复 org-scoped member mutations 的跨组织目标访问问题，确保 URL 中的 org scope 被用于 PATCH/DELETE 的目标 member 校验。

核心改动看过了：
- DELETE route 现在把 `requireOrgAdmin` 返回的 `scope.organizationId` 传入 `deleteMember`，service 在任何删除 side effect 前拒绝跨 org target。
- `updateMember` 先读取 target 并做 caller org 校验，再处理 empty patch early return，避免 `{}` patch 泄露其他 org 的 member 详情。
- 新增 empty PATCH 和 DELETE 的跨 org regression tests，DELETE 覆盖 member row 仍存在、human agent 未 tombstone。

涉及数据库 schema / 核心数据结构变更：无。改动是 org scope 参数传递和 service 层校验顺序调整。

验证：
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/members.test.ts --maxWorkers=2`，9 tests passed
- `pnpm typecheck`，12 turbo tasks passed

No blockers found. Approved.

PR #979, opened 2026-06-11
feat(github): explicit follow/unfollow CLI replaces session-event auto-binding

yuezengwu:
## What

`first-tree github follow|unfollow|following <entity>` — the single explicit path for wiring a GitHub entity's webhook event stream into a chat — and, in the same PR, **deletion of the session-event auto-binder** (`gh pr create` / `gh issue create` stdout extraction). Deletion and addition land together so the system never has two parallel wiring semantics, nor a window with none.

Context PR (decision layer, opened first per tree-write discipline): agent-team-foundation/first-tree-context#486

## Semantics (settled S1–S6)

- **Chat-scoped, not personal**: follow wires the entity's events into *this chat*; every event wakes the wiring agent (active attention with real cost — restraint is taught in the skill).
- **One line, one room**: a (human, delegate) line for an entity lives in exactly one chat (table PK). Re-follow elsewhere → `409` with the existing chat's id/topic; `--rebind` *moves* the line and rewrites `bound_via` so a moved `direct` anchor can't impersonate a github-minted chat.
- **Creating is not following**: no creation path auto-binds anymore. The skill teaches "follow in the same breath as creation"; the create-vs-`opened`-webhook race converges on the PK (webhook wins + mints → 409 → rebind).
- **Unfollow = the task's attention span closes**: severs *every* row pointing at the chat (any pair, any bound_via), idempotent (`200` + `removed` count, `removed: 0` is success), never touches the GitHub API.

## Server

- `services/github-entity-follow.ts`: `declareEntityFollow` — fail-fast GitHub resolution before any write (existence, issue-vs-PR discrimination via the issues endpoint, canonical `full_name` key normalization, lifecycle state; `422` when the App installation can't see the repo, `503` when GitHub is down — never a blind/ghost row). `removeEntityFollow` — API-free, case-insensitive key match + sha-prefix for commits. `listChatGithubEntities` — shared by both route trees (the Class C sidebar GET now delegates to it).
- Routes: `GET/POST/DELETE /api/v1/chats/:chatId/github-entities` (Class C — human follows bind under the caller with their `delegate_mention` as delegate) and `/api/v1/agent/chats/:chatId/github-entities` (Class D — agent follows bind under the chat's representative human via the reused, entry-point-agnostic `resolveBindingPair`).
[...]

PR #1014, opened 2026-06-12
feat(client): teach follow-after-create default in unconditional agent briefing

yuezengwu:
## Problem

Since #979 replaced session-event auto-binding with the explicit `github follow` entrance, agents stopped proactively tracking the PRs/issues they create. The follow-after-create rule lived only in the on-demand `first-tree-github` skill — and progressive disclosure only fires when an agent is already thinking about following, which is never the creation moment. Net effect: entities created by agents routinely had no mapping, and their webhook events went nowhere.

## Change

**Unconditional briefing block** — new `## GitHub Entity Attention` section under `# Working in First Tree` (always present, so it exists at creation time):

- **Default: follow what you create.** Immediately after creating a PR/issue through any path, run `<bin> github follow <url>`. Creation never auto-follows.
- Skip only when the entity is clearly unrelated to the chat's task.
- Unfollow when the human explicitly asks to stop tracking, or when the task's attention span on the entity has genuinely closed.

**Supporting fixes** (from self-review):

- The full-guide pointer is gated on the tree binding: tree-bound → load `first-tree-github`; tree-less → `github follow --help` (tree-less agents have no skill payloads on disk; pointing them at the skill would violate the existing gating discipline of `# Required Reading` / the Family map).
- CLI Overview gains the `github …` namespace row — the briefing now instructs these commands, so its own command map must list them.
- The First Tree Family row for `first-tree-github` shrinks to a pointer at the inline block, so the normative rule exists once per briefing.
- `SKILL.md`: description + decision-guide rule 1 carry the DEFAULT framing; the skip-exception cross-reference names its list ("Do NOT follow" rule 3); version → 0.1.1.

## Non-goals

The decision layer is unchanged: explicit declaration stays the only agent-side entrance (no auto-binding returns). A possible follow-up is surfacing the server's `webhook_dropped_creation` signal as a metric to measure whether this prompt-layer fix moves the needle.

## Testing

- `pnpm check` / `pnpm typecheck` pass
- client suite: 100 files / 1203 tests pass (new assertions pin the block, the tree-bound/tree-less pointer gating, and the CLI Overview row)
- `validate:skill` passes
- QA (@qa) independently verified the generated briefing from this worktree contains the default-follow rules end-to-end

Companion tree PR (opened first): agent-team-foundation/first-tree-context#493


baixiaohang (approved):
## Review by codex-assistant - PR #1014

**Recommendation**: approve
- Reason: the change moves the follow-after-create default into the unconditional briefing where it can actually fire, while keeping the explicit `github follow` decision layer and tree-bound/tree-less skill gating intact.

**Level**: A
- Semantic lift: none.

**PR Summary**
- Solves: agents stopped tracking PRs/issues they created after auto-binding was removed, because the rule lived only in an on-demand skill.
- Design: add `## GitHub Entity Attention` to the generated briefing, point tree-bound agents at `first-tree-github` and tree-less agents at CLI help, add the `github ...` CLI Overview row, and align the skill text/version.
- Affected modules: generated agent briefing, agent-briefing tests, and `first-tree-github` skill payload.

**Review Notes**
✅ 1. The direction matches the Context Tree decision for explicit GitHub entity binding: no implicit agent-side binding returns, and the default belongs in the unconditional briefing.
✅ 2. The implementation keeps the tree-bound/tree-less boundary consistent: tree-bound agents get the skill pointer, while tree-less agents are not told to load missing skill payloads.
✅ 3. Tests pin the new block, the default follow wording, pointer gating, and the CLI Overview surface; remote CI and `Validate Skill` are green.
⚠️ 4. Non-blocking existing drift remains in `# Required Reading`: it still names only `first-tree-read` / `first-tree-sync` as remaining on-demand skills even though the Family map also includes seed/github. This predates the PR and does not block the default-follow fix.

**Executed**
- Submitted approve after explicit baixiaohang authorization.
- I did not run local checks or tests per PR Review mode; conclusion is based on static diff/source review, relevant Context Tree nodes, and remote CI status.

PR #1023, opened 2026-06-12
fix(web): wrap unbreakable long runs in chat markdown instead of overflowing

Gandy2025:
## Problem

A horizontal scrollbar appeared at the bottom of the chat timeline (middle livestream column).

Root cause, verified live against the affected chat:

1. One message contained a bare login URL with two embedded JWTs — a single **~620-char unbreakable run**. The `Markdown` prose wrapper sets no `overflow-wrap`, so the browser cannot break inside the run and the paragraph overflows the message column (`<p>` measured 2473px wide in an 852px column).
2. The timeline scroll container only sets `overflow-y-auto`. Per CSS, when one overflow axis is non-`visible` the other computes `visible` → `auto`, so the container's `overflow-x` is effectively `auto` and the overflow surfaces as a horizontal scrollbar on the whole stream.

The whole rendered page had exactly one overflowing chain — that paragraph; tables/`pre`/composer/token-usage rows were all measured and ruled out (typography's `pre` keeps its own internal `overflow-x: auto`, so fenced code blocks never widen the column).

## Fix

Add `break-words` to the `Markdown` prose wrapper so unbreakable runs (bare URLs, tokens, long paths) wrap within the message column. One class; applies to every `Markdown` call site.

`break-words` (`overflow-wrap: break-word`) is preferred over `overflow-wrap: anywhere`: identical wrapping behavior in width-constrained columns, but it does not change min-content intrinsic sizing, so shrink-to-fit consumers of `Markdown` (popovers, hover cards) keep their natural width.

## Verification

- Live A/B in the affected chat: scroll container `scrollWidth` 2541 → 960 (= `clientWidth`) with the rule applied; horizontal scrollbar gone. Toggling the rule reproduces/clears the bug.
- Both `break-word` and `anywhere` were measured against the real 617-char URL in an isolated repro; both fully eliminate the overflow.
- `pnpm check` / `pnpm typecheck` green; web suite 787/787 passed, including a new regression test pinning `break-words` on the prose wrapper.

### DIFF
--- a/packages/web/src/components/__tests__/markdown-overflow-wrap.test.tsx
+++ b/packages/web/src/components/__tests__/markdown-overflow-wrap.test.tsx
@@ -1,0 +1,43 @@
+// @vitest-environment happy-dom
+
+import { act, type ReactElement } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { Markdown } from "../ui/markdown.js";
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+
+let container: HTMLElement;
+let root: Root;
+
+beforeEach(() => {
+  container = document.createElement("div");
+  document.body.appendChild(container);
+  root = createRoot(container);
+});
+
+afterEach(() => {
+  act(() => root.unmount());
+  container.remove();
+});
+
+function renderMarkdown(markdown: string): void {
+  act(() => root.render((<Markdown>{markdown}</Markdown>) as ReactElement));
+}
+
+/**
+ * Regression: a chat message carrying one unbreakable run (a bare login URL
+ * with two embedded JWTs, ~620 chars) widened the chat timeline and put a
+ * horizontal scrollbar on the whole message stream. The prose wrapper must
+ * carry `break-words` (overflow-wrap: break-word) so such runs wrap instead
+ * of overflowing. happy-dom does no layout, so the class on the wrapper is
+ * the testable contract.
+ */
+describe("Markdown long-token wrapping", () => {
+  it("keeps break-words on the prose wrapper", () => {
+    renderMarkdown(`http://localhost:5179/auth/github/complete#access=${"eyJx".repeat(150)}`);
+    const wrapper = container.querySelector(".prose");
+    expect(wrapper).not.toBeNull();
+    expect(wrapper?.className).toContain("break-words");
+  });
+});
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1023

**建议**: approve
- 理由: 变更把长不可断 run 的换行能力收敛到共享 Markdown wrapper，范围和回归测试都匹配问题根因。

**分级**: C
- Path 初判: 仅含 `packages/web/**` -> C
- 语义升级: 无；未触及订阅、WS 帧解析或数据层

**PR 摘要**
- 解决问题: chat timeline 中超长裸 URL/JWT 不换行导致中间消息流出现横向滚动条
- 设计方案: 在共享 `Markdown` prose wrapper 上加 `break-words`，保留 `break-word` 的 min-content 行为，并新增 happy-dom 测试钉住 wrapper class
- 受影响模块: `packages/web/src/components/ui/markdown.tsx`，`packages/web/src/components/__tests__/markdown-overflow-wrap.test.tsx`

**Review 意见**
✅ 1. 共享 wrapper 是合适入口，所有 Markdown 调用点继承修复，不需要在 chat timeline 单点补丁。[R1 / packages/web/src/components/ui/markdown.tsx]
✅ 2. `break-words` 相比 `anywhere` 避免改变 shrink-to-fit 场景的 min-content sizing，和 PR 里提到的 popover/hovercard 风险一致。[R3 / packages/web/src/components/ui/markdown.tsx]
✅ 3. 回归测试钉住 `break-words` class；happy-dom 无布局能力，测试 class contract 是当前可验证的合理边界。[packages/web/src/components/__tests__/markdown-overflow-wrap.test.tsx]

**已执行**
- 已 approve

yuezengwu:
Post-merge review by yzw-codex: no findings.

Implementation goal/core change: fix Workspace chat markdown overflow by adding `break-words` to the shared `Markdown` prose wrapper, so unbreakable bare URLs/tokens/paths wrap inside the message column. The regression test pins that wrapper contract in happy-dom.

Review notes:
- The change is correctly scoped to the shared Markdown renderer, so chat messages, request bodies, image captions, prompt/resource previews, and document previews inherit the fix.
- Existing Markdown callers do not appear to pass an overriding `break-normal`/wrapping class; the compact resource-preview class only adjusts heading scale.
- No core data structures or database changes are involved.

Verification: `git diff --check origin/main...HEAD` passed. I could not run the targeted Vitest locally because this checkout/worktree has no `node_modules` installed (`vitest: command not found`); PR description reports `pnpm check`, `pnpm typecheck`, and the web suite green.

Issue #1025, opened 2026-06-12
Onboarding data-model review: per-(user,org) semantics stored per-user, installation↔org 1:1 dead-end, and follow-up fixes


Gandy2025:
Consolidated review of the onboarding core data model — the relationships between first-tree `users` / `organizations` / `members` / `clients` / `agents` and GitHub-side users / accounts / App installations — plus the defects the review surfaced and suggested fixes. All references pinned to `303aaec0`.

## Entity model (as implemented)

```mermaid
flowchart LR
  subgraph G["GitHub side"]
    GU["GitHub user (OAuth login identity)"]
    GA["GitHub account (User or Organization)"]
    INST["App installation (installationId unique)"]
    GU -->|"is one / admins"| GA
    GA -->|"hosts at most 1"| INST
  end
  subgraph FT["first-tree side"]
    U["users (onboarding stamps live here)"]
    C["clients (user-owned machines)"]
    O["organizations (= team)"]
    M["members (unique user x org)"]
    A["agents (human 1:1 + team agents)"]
    U -->|"1:N machines"| C
    U -->|"1:N"| M
    O -->|"1:N"| M
    M -->|"manages 1:N"| A
    A -->|"runtime pin N:1"| C
  end
  GU -.->|"1:1 via auth_identities"| U
  INST ==>|"1:1 both ways - dead end"| O
```

### "org" means two unrelated things

| Where it appears | What it means |
|---|---|
| `organizations` table, `/orgs/:orgId/*` routes, JWT org scope, `*.organizationId` columns | first-tree org (the product's "team") — [zero GitHub fields](https://github.com/agent-team-foundation/first-tree/blob/303aaec0915f31ed81b2927351f1b8412933ce3b/packages/server/src/db/schema/organizations.ts) |
| `github_app_installations.accountType / accountLogin / accountGithubId` | GitHub-side account hosting the install — `User` or `Organization` |
| `github_app_installations.hubOrganizationId` | first-tree org (the `hub` prefix is the deliberate disambiguation) |
| `GET /user/memberships/orgs/{login}` in the OAuth callback | GitHub org membership (admin check only) |

There is no direct FK between a first-tree org and a GitHub org. The link is two hops: first-tree org —1:1— App installation —hosted on— GitHub account (`User` \| `Organization`). A GitHub org has no entity of its own; it exists only as snapshot fields on the installation row.

### Key verified facts

[...]

Issue #951, opened 2026-06-10
Onboarding: resume an abandoned-mid-flow user instead of ejecting on reopen (needs a per-org completion signal)


Gandy2025:
Follow-up split out from the eject-bug fix #948 (which is merged). **Non-urgent, decoupled from that bug.**

## Symptom
A user who created their agent but **abandoned onboarding before finishing** (admin: didn't connect-code / build the Context Tree; invitee: didn't launch) and later **reopens `/onboarding`** is **ejected back to the workspace** — they can't resume connect-code / kickoff via the flow. They'd have to finish team setup piecemeal from Settings instead.

## Why
The leave-the-flow guard (now correctly an entry-time decision after #948) uses `currentOrgHasUsableAgent`. Creating the first agent flips that true, so on re-entry the guard reads "this org already has a usable agent → you're done" and bounces them. There is **no per-org signal for "did this org actually finish onboarding"** — `onboardingCompletedAt` is account-level and was deliberately removed from the per-org gate by #732 (multi-org correctness).

#948 deliberately kept this behavior (the pre-#948 reactive version ejected here too, so the fix doesn't regress it) — but it can't *improve* it without a real per-org completion flag.

## Proposed fix
Add a **per-org onboarding-completion signal** (e.g. `org_members.onboarding_completed_at` or an org-scoped flag), so the entry guard can tell:
- org has a usable agent **and** this member finished onboarding for this org → eject (current behavior, correct);
- has an agent **but** onboarding for this org isn't finished → let them resume at the right step.

Scope: shared Zod schema + server persistence/route + FE entry/leave gates in `onboarding-page.tsx` / `steps.ts`. Worth pairing with a small Context Tree update to `system/cloud/onboarding.md` State Model once the signal exists.

## Not in scope here
The reactive-eject bug itself (fixed in #948) and the redesigned kickoff (#943).

Gandy2025:
## Implementation plan — per-org onboarding completion signal

Approved approach (gandy2025): **completion is an explicit per-org event**, recorded at the moment the user clicks the final kickoff launch — not derived from "has a usable agent" / "has a tree" / "has a chat" (all unreliable; notably the no-repo "Meet your agent" path completes legitimately but builds no tree). This supersedes the #948 stopgap (the entry-time `useRef` freeze in `onboarding-page.tsx`), which only stopped the *mid-flow* eject but still bounces a returning, not-yet-finished admin.

### Signal
[...]

PR #529, opened 2026-05-22
feat(web): beginner-first standalone /onboarding flow

Gandy2025:
> Replaces #528 (renamed branch `claude/onboarding-refactor-Xd0nd` → `feat/onboarding-refactor` to satisfy the branch-name convention check). Same commit, same content.

## Why

New-user onboarding currently lives inline in the ChatView center panel and assumes a developer who already knows GitHub / CLI / repos / runtimes / context-trees. This rebuilds it as a dedicated, full-screen **`/onboarding`** flow designed for a complete beginner — someone who just wants an AI teammate working on their project — so the path from landing page → "my AI teammate is doing its first task" needs no docs, no engineer, no manual CLI, no DB.

It also fixes a verified drift the proposal flagged: **sign-in (`authorize`) never reliably installs the GitHub App**, so the only real install entry (`installations/new`) was buried in Settings and absent from onboarding. Connecting code is now a first-class, explained step.

Builds on `hub-onboarding-redesign.20260522.md`; implements the task `new-user-onboarding-refactor.20260522.md`.

## What changes for the user

- **One step at a time**, each with a plain-language *why* and a *"what you'll have"* panel. Vocabulary is reduced to **team / your code / your computer / AI teammate / team knowledge** — no *repo*, *runtime*, *binding*, *context-tree* in the UI.
- **Sign-in (identity) and connecting code (installing the GitHub App) are separate, explained steps.** The connect-code step calls `install-url` (`installations/new`), not the sign-in `authorize` URL. Resilient failure states: not installed, App not configured (503), not an admin, missing repo scope, no repos — each with a way forward (never a dead end).
- **"Show me how" visual guides** (progressive disclosure) on the two highest-friction steps: a mock terminal for connecting a computer, and a Choose → Install → back-here flow for connecting code. Collapsed by default; theme-aware, accessible, `prefers-reduced-motion` safe; no external GIF/video assets.
- **Admin vs invitee paths.** Admins set up team + code + knowledge; invitees bind to the team's **existing** knowledge base (never creating a duplicate), with a clear waiting state + "what's missing / who to ask" when the team isn't configured yet.
- **Completion = AI teammate online + first task kicked off** (per the agreed bar). Stamps `onboarding_completed_at`, drops the user into their first chat, and is resumable from **Settings → Setup**.
- The standalone flow **replaces** the inline wizard via a redirect; the old `OnboardingView` is left untouched as a reversible fallback (no deletion).

## Implementation

- `packages/web/src/pages/onboarding/`:
[...]

PR #1027, opened 2026-06-12
fix(web,server): stop stamping onboarding dismissal on the completion path (#1025 P0-1)

Gandy2025:
## What

P0-1 hotfix from #1025: the normal onboarding completion path (`completeAndEnterChat`) stamped the account-level `onboarding_dismissed_at` alongside `onboarding_completed_at` — a leftover from the retired inline workspace stepper (#529). Since `shouldEnterOnboarding` consults the dismissal **before** its per-org readiness branch, the org-level re-entry gate introduced by #732 was unreachable for any user who ever finished onboarding: creating or joining a second org landed them in an empty workspace with no auto-bounce and no manual path back.

- `completeAndEnterChat` now only stamps the completion flag (wrapped in a local try/catch so the always-navigate invariant stays local to the flow). Dismissal is reserved for the explicit "finish later" action — the two stamps are semantically orthogonal again.
- Migration `0062` (data-only, snapshot chain verified) performs the **broad clear**: every row with both `onboarding_completed_at` and `onboarding_dismissed_at` set has the dismissal cleared. Timestamp comparisons cannot reliably separate row classes (both old endpoints stamp with ms-precision `new Date()` and the old client fired the writes in parallel), so the wider blast radius is accepted deliberately across all three classes — completion artifacts (the bug), the 0043 backfill cohort (included by design; worst outcome is one wizard prompt in a not-yet-ready org with finish-later one click away), and dismiss-then-complete rows (completion supersedes the earlier dismissal). A `DO` block `RAISE NOTICE`s the stamp-ordering distribution into the boot log before the clear, per the Phase-0 spec in the #1025 long-term proposal. Genuine finish-later users who never completed are dismissed-only and untouched by construction.

## What this does NOT do

Scoped strictly to the enter-guard artifact. Membership-level stamps, the resume path (#951), server-side kickoff, and the installation junction (#1022) stay with the long-term rework proposed on #1025.

## Test plan

- `hooks-flow.test.tsx`: completing must stamp completed **without** dismissing (regression assertion: `dismissOnboarding` called exactly once — by the explicit `finishLater`).
- Verified no other consumer depends on dismissed-after-completion: `shouldLeaveOnboarding` never reads it; Settings → Onboarding visibility is governed by `completedAt`; Resume clears dismissal via `restoreOnboarding()` before re-entry, so the dual stamp cannot legitimately reappear.
- Gates: biome check ✓, typecheck ✓, web 787/787 ✓, server 1390/1390 ✓ (server suite applies migration 0062 including the DO block), check-migrations ✓.

Refs #1025


packages/web/src/pages/onboarding/onboarding-flow.tsx
```diff
@@ -181,20 +181,27 @@ export function OnboardingFlowProvider({ path, children }: { path: OnboardingPat
...
+      // bare workspace with no way back (the original call's target, the
+      // retired inline workspace stepper, no longer exists).
       clearPersistedStep(path);
       // Clear the per-tab agent-uuid stash now that the kickoff has resolved and
       // used it — so a later same-tab onboarding/recovery in a DIFFERENT org
       // can't read a stale cross-org agent (the org filter in
       // resolveOnboardingAgent only catches that when the org id is known).
       writeOnboardingAgentUuid(null);
-      await Promise.allSettled([dismissOnboarding(), markOnboardingCompleted()]);
+      await markOnboardingCompleted();
```
yuezengwu: This now blocks navigation if the completion POST rejects. `markOnboardingCompleted()` optimistically stamps state but still awaits `postOnboardingCompleted()` and rethrows on failure, while this function's contract/comment still says a transient failure must never strand the user and the previous code used `Promise.allSettled(...)` before navigating. Please keep the single completion write, but swallow/settle its failure before `navigate(...)` (for example `await Promise.allSettled([markOnboardingCompleted()])` or a local `try/catch`) and add a regression assertion that a rejected completion write still navigates into the chat.

yuezengwu (dismissed):
Reviewed the PR goal and core changes: this hotfix separates the normal onboarding completion stamp from the explicit finish-later dismissal stamp, and adds a data-only migration (`0062`) to clear historical completed+dismissed artifacts. The direction matches the onboarding state model I checked in the Context Tree: dismissal is the finish-later suppressor, while org readiness must remain reachable for second-org re-entry.

One blocker: the completion path now awaits `markOnboardingCompleted()` directly, which regresses the existing "always enter the chat after kickoff" behavior when that idempotent completion write has a transient failure. Please preserve the previous failure semantics while only removing the dismissal write.

DB/data note for human reviewers: this PR includes a data migration clearing `users.onboarding_dismissed_at` where `onboarding_completed_at IS NOT NULL`. I verified the schema snapshot is otherwise unchanged apart from the expected snapshot id chain.

baixiaohang (commented):
## Review by codex-assistant - PR #1027

**Recommendation**: comment
- Reason: I do not see an additional blocker in the PR diff, but this is B-high due to the server data migration and an existing CHANGES_REQUESTED review is already open; leaving approval paused.

**Level**: B-high
- Path initial: includes a server data migration/snapshot under `packages/server/drizzle/**` -> treated as server migration scope.
- Semantic lift: none beyond the migration scope.

**PR Summary**
- Problem: normal onboarding completion also stamped account-level dismissal, which short-circuited second-org onboarding re-entry.
- Design: stop calling `dismissOnboarding` on `completeAndEnterChat`; add data-only migration clearing completed+dismissed user rows; update the flow regression test.
- Impacted modules: `packages/web/src/pages/onboarding`, `packages/server/drizzle`.

**Review Notes**
✅ 1. The web change matches the onboarding state model: dismissal remains the finish-later suppressor, while completed users still go through the current-org readiness gate. `restoreOnboarding()` clears dismissal before the Settings Resume path, so explicit finish-later then resumed completion does not leave a new dual stamp. [R1 / packages/web/src/pages/onboarding/onboarding-flow.tsx:201]

⚠️ 2. Data migration blast radius: `0043_onboarding_completed_at.sql` intentionally backfilled existing dismissed users into `onboarding_completed_at`, so `0062` will clear dismissal for that legacy cohort too. I do not think this blocks the hotfix because `0043` already accepted treating those users as completed, but if that effect is intentional the migration comment should say so rather than claiming every completed+dismissed row came only from kickoff completion. [R5 / packages/server/drizzle/0062_clear_onboarding_dismiss_completion_artifact.sql:14]

✅ 3. Snapshot chain looks consistent: `0062_snapshot.prevId` matches `0061_snapshot.id`, the user column schema is unchanged, and `_journal` / `LATEST` advance to 0062. [packages/server/drizzle/meta/0062_snapshot.json]

**Existing Review Note**
- yuezengwu has already requested changes on the direct `await markOnboardingCompleted()` concern. I checked the current auth implementation: it ultimately awaits the API wrapper that swallows completion-write failures, so I do not see a production navigation regression in this diff. Restoring `await Promise.allSettled([markOnboardingCompleted()])` would still make the "always navigate" invariant local to this flow and is a low-cost cleanup.

**Executed**
- Submitted comment review; no approval submitted.

packages/web/src/pages/onboarding/onboarding-flow.tsx
yuezengwu: Correction: this blocker is withdrawn. I missed the extra wrapper layer in `packages/web/src/api/onboarding-events.ts`: `auth-context` awaits `postOnboardingCompleted()`, and that helper already catches and swallows completion POST failures. So the direct `await markOnboardingCompleted()` here does not block navigation on a transient API failure in production. Keeping the failure-tolerance local to this flow would still be a reasonable cleanup, but it is not a required change.

yuezengwu:
Correction to my earlier review: I dismissed the CHANGES_REQUESTED review and resolved the inline thread. The direct `await markOnboardingCompleted()` is not a production navigation blocker because the awaited API helper in `packages/web/src/api/onboarding-events.ts` already catches and swallows completion POST failures.

Current review stance: no web-flow blocker from me. The remaining point is data-migration scope: `0062` clears every completed+dismissed row. That appears aligned with this hotfix, but because this is a server data migration and `0043_onboarding_completed_at.sql` previously backfilled some dismissed users into `onboarding_completed_at`, a human should confirm the legacy cohort is intentionally included. If intentional, I recommend adjusting the migration comment so it does not claim every completed+dismissed row came only from kickoff completion.

Gandy2025:
Addressed both reviews in caba396b:

**@yuezengwu — navigation invariant**: even though the API helper already swallows completion-POST failures (as your correction notes), the always-navigate guarantee is now enforced locally — `completeAndEnterChat` wraps the stamp in its own try/catch, so the invariant no longer depends on a callee's error handling.

**@yuezengwu @baixiaohang — 0062 vs the 0043 backfill cohort**: good catch, and on inspection I went one step further than a comment fix. `0043` backfilled by **copying** `onboarding_dismissed_at` into the new column verbatim, so that cohort is precisely identifiable: `completed_at = dismissed_at` (exact equality). The true completion artifact comes from two independent writes (PATCH dismissal + POST completion) whose timestamps always differ. `0062` now adds `AND completed_at <> dismissed_at`, which **excludes the legacy cohort entirely**: it is an indistinguishable mix of genuine finish-later users and pre-column completers, and clearing it would resurrect auto-bounce for users who only ever asked to hide the flow — the conservative call is to keep their suppression and let the membership-level rework (#1025 long-term) supersede it. The migration comment now documents both cohorts.

Also restored the journal's trailing newline (drizzle-kit dropped it on generate).

One residual noted for the record, deliberately out of scope: during the deploy window, a stale SPA bundle in a long-lived tab can still stamp dismissal on completion (old `completeAndEnterChat`) and re-create the artifact for that user until they reload. A server-side guard can't fully close it (the parallel writes race), the window is narrow, and the membership-level stamps remove the mechanism entirely.

Gates re-run: biome ✓, typecheck ✓, web 787/787 ✓, server 1390/1390 ✓ (suite applies 0062), check-migrations ✓.

packages/server/drizzle/0062_clear_onboarding_dismiss_completion_artifact.sql
```diff
@@ -0,0 +1,27 @@
...
+-- dismissed-only (completion is stamped exclusively by walking kickoff to
+-- success) and are untouched by construction. New completions no longer
+-- write the dismissal at all, and a completed account can no longer reach a
+-- dismiss affordance (Settings hides the onboarding entry once completed,
+-- and Resume clears the dismissal first), so the artifact cannot reappear.
+UPDATE "users"
+SET "onboarding_dismissed_at" = NULL
+WHERE "onboarding_completed_at" IS NOT NULL
+  AND "onboarding_dismissed_at" IS NOT NULL
+  AND "onboarding_completed_at" <> "onboarding_dismissed_at";
```
yuezengwu: This equality check can miss real completion artifacts. The old `completeAndEnterChat` sent dismissal and completion in parallel, and both server endpoints stamp with JavaScript `new Date()` (`/me/onboarding` sets `onboardingDismissedAt`, `/me/onboarding-completed` sets `onboardingCompletedAt`). Because `Date` is millisecond-resolution, two independent requests can still store exactly equal timestamps. Those rows would be treated as the 0043 backfill cohort and keep `onboarding_dismissed_at`, leaving the second-org re-entry bug unfixed for those users. Please don't use exact timestamp inequality as the artifact discriminator unless there is a stronger guarantee than the current write path provides.

yuezengwu (changes requested):
Re-reviewed the synchronized head `caba396b`. The web-flow update now keeps the always-navigate invariant local to `completeAndEnterChat`, and the snapshot/journal chain still looks structurally consistent.

I found one blocker in the revised data migration: the new `completed_at <> dismissed_at` discriminator relies on the two old completion-path writes never producing equal timestamps, but the server stamps both columns with JavaScript `new Date()` and the old web code sent the PATCH and POST in parallel. Equal millisecond timestamps are possible for a real completion artifact, so 0062 can now leave some of the exact rows this hotfix is meant to repair. Please use a discriminator that does not depend on timestamp inequality, or explicitly choose the broader/smaller blast radius with that residual risk accounted for.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1027 follow-up (`caba396b`)

- Reason: the new `0062` predicate excludes equal timestamps, but the old completion artifact can also produce equal timestamps, so the migration can leave some affected users un-fixed.

**Level**: B-high
- Path initial: includes a server data migration/snapshot under `packages/server/drizzle/**` -> treated as server migration scope.
- Semantic lift: none beyond the migration scope.

**PR Summary**
- Problem: normal onboarding completion stamped account-level dismissal, which short-circuited second-org onboarding re-entry.
- Current design: stop calling `dismissOnboarding`; locally swallow completion-stamp failures before navigating; add data migration clearing only completed+dismissed rows whose two timestamps differ.
- Impacted modules: `packages/web/src/pages/onboarding`, `packages/server/drizzle`.

**Review Notes**
❌ 1. `onboarding_completed_at <> onboarding_dismissed_at` is not a reliable discriminator for the completion artifact. The comment says the two independent server writes are "close but never equal", but both old endpoints stamp with Node `new Date()` (`PATCH /me/onboarding` and `POST /me/onboarding-completed`), which has millisecond precision. The old `Promise.allSettled([dismissOnboarding(), markOnboardingCompleted()])` path sent those requests in parallel, so a real completion artifact can land with equal millisecond timestamps. `0062` would now preserve that row's dismissal, leaving the second-org re-entry gate unreachable for exactly the class this hotfix is meant to repair. Please avoid relying solely on timestamp inequality here: either use a boundary that preserves only the 0043 legacy backfill cohort, or return to clearing all completed+dismissed rows if that legacy effect is accepted. [R5 / packages/server/drizzle/0062_clear_onboarding_dismiss_completion_artifact.sql:27]

✅ 2. The web flow update now keeps the always-navigate invariant local to `completeAndEnterChat`; a completion-stamp rejection no longer blocks `navigate(...)`. [R4 / packages/web/src/pages/onboarding/onboarding-flow.tsx:201]

**Executed**
- Submitted request_changes on the new head; no local tests or checks run.

Gandy2025:
@yuezengwu @baixiaohang — you're both right, and the `<>` discriminator is gone in ffb0a245. I verified the premise: both old endpoints stamp with millisecond-precision Node `new Date()` and the old client fired the PATCH/POST in parallel, so a genuine artifact can carry equal stamps — the predicate would have left exactly the rows this hotfix exists to repair.

`0062` is back to the **broad clear** (every completed+dismissed row), which also matches the Phase-0 spec in the long-term proposal now on #1025. The migration comment documents the accepted blast radius across all three row classes:
1. completion artifacts — the bulk, the bug being repaired;
2. the 0043 backfill cohort — treating them as completed was 0043's own accepted trade-off; worst post-clear outcome is one wizard prompt in a not-yet-ready org, with finish-later one click away;
3. dismiss-then-complete rows — completion supersedes the earlier dismissal for the enter gate's purpose.

Per the proposal's "log the stamp-ordering distribution first" note, the migration now `RAISE NOTICE`s the class counts (equal / dismissed-before / dismissed-after) into the boot log before the UPDATE, so the thin-tail assumption is checkable against production data at deploy time.

Gates: check-migrations ✓, server 1390/1390 ✓ (suite applies 0062 including the DO block), web unchanged since last run.

yuezengwu (approved):
Approved latest head `ffb0a245`.

The current PR goal and core change are clear: normal onboarding completion no longer stamps account-level dismissal, and 0062 now deliberately performs the broad data cleanup for every completed+dismissed row. The previous blocker is addressed: the migration no longer relies on timestamp inequality, and the DO block only logs the stamp-ordering distribution before the broad clear. The web flow also keeps the always-navigate invariant local via try/catch around the completion stamp.

DB note for human awareness: this remains a data migration that clears `users.onboarding_dismissed_at` for completed users, including the 0043 backfill cohort by design. I verified the schema snapshot is unchanged apart from the expected snapshot id chain; I did not rerun the full test suite locally.

Non-blocking cleanup: the PR body still contains the older narrower wording ("that combination is always a completion artifact" / genuine finish-later users keep suppression). The migration comment and author follow-up now describe the actual broad-clear scope correctly.

baixiaohang (approved):
Approved latest head `ffb0a245` after B-high authorization from baixiaohang.

The previous blocker is resolved: migration `0062` no longer relies on timestamp inequality and deliberately performs the broad clear for completed+dismissed rows, with the accepted 0043 backfill / dismiss-then-complete blast radius documented in the migration and PR body. The `RAISE NOTICE` block records the pre-clear stamp-ordering distribution without changing the broad cleanup. The web completion path also keeps the always-navigate invariant local with a try/catch around the completion stamp.

I did not run local tests or checks in review mode. Remote checks are green/skipped at the time of approval.

PR #1029, opened 2026-06-12
fix: gate optional config groups by activators

baixiaohang:
## Summary
- add optional-group `activateBy` metadata while preserving default any-field activation
- require every optional ancestor to be active before resolving a field, including readonly config resolution
- gate feedback config on repo/token so `LLM_API_KEY` alone does not enable feedback

## Verification
- `pnpm --filter @first-tree/shared exec vitest run --passWithNoTests --maxWorkers=2 src/config/__tests__/resolver.test.ts src/config/__tests__/server-config.test.ts`
- `pnpm check`
- `pnpm typecheck`

Review: codex-reviewer passed; no findings.


yuezengwu (approved):
Reviewed the optional-group activation change. The implementation preserves default any-field activation, adds explicit `activateBy` gating, requires all optional ancestors to be active before resolution, and applies the same activation rules in readonly resolution. Feedback config is now gated on the repo/token fields so LLM-only env vars no longer enable the group.

No blocker findings. Local verification:
- `pnpm --filter @first-tree/shared exec vitest run --passWithNoTests --maxWorkers=2 src/config/__tests__/resolver.test.ts src/config/__tests__/server-config.test.ts`
- `pnpm check` exited 0 with unrelated existing warnings outside this PR's diff
- `pnpm typecheck`

PR #1024, opened 2026-06-12
feat(client,server): detect Context Tree writes from git status

Gandy2025:
## Summary

- Add a client-side Context Tree git-status write tracker that diffs newly dirty tree paths after successful tool calls.
- Emit synthetic `toolFileRefs` with `origin: "git_status_delta"` for Claude SDK/TUI and Codex command/file-change events.
- Teach server IO recording to persist `git_status_delta` refs as write activity even when the original shell command is otherwise unsupported.

## Notes

This implements the design from `docs/development/context-tree-git-write-detection.md`: write attribution comes from the tree clone's git status rather than expanding the shell parser. The feed remains activity evidence, not an audit log; git history and the change map remain authoritative write history.

The tracker is conservative: it records only paths newly dirty relative to the session/tool baseline and updates the baseline after recording to avoid repeated events.

## Validation

- `pnpm --filter @first-tree/client test src/__tests__/context-tree-git-status.test.ts src/__tests__/claude-code-tool-events.test.ts src/__tests__/codex-context-tree-io.test.ts`
- `pnpm --filter @first-tree/server test src/__tests__/context-tree-io.test.ts`
- `pnpm --filter @first-tree/shared typecheck`
- `pnpm --filter @first-tree/server typecheck`
- `pnpm --filter @first-tree/client typecheck`
- `pnpm exec biome check packages/client/src/runtime/context-tree-git-status.ts packages/client/src/__tests__/context-tree-git-status.test.ts packages/client/src/__tests__/claude-code-tool-events.test.ts packages/client/src/__tests__/codex-context-tree-io.test.ts packages/client/src/handlers/claude-code.ts packages/client/src/handlers/claude-code-tui/index.ts packages/client/src/handlers/codex.ts packages/server/src/__tests__/context-tree-io.test.ts packages/server/src/services/context-tree-io.ts packages/shared/src/schemas/context-tree.ts packages/shared/src/schemas/session-event.ts`

PR #1032, opened 2026-06-12
feat(client,shared): attribute Context Tree refs by git repo identity

Gandy2025:
## Problem

The Context tab feed rarely shows writes. Root cause (discussed in chat): the agent workflow keeps **two kinds of Context Tree checkouts** —

- the **shared read clone** the runtime binds as `contextTreePath` (exposed as the `<workspace>/context-tree` symlink), used for reading;
- **per-task worktrees** under `worktrees/<task>`, where tree PRs are actually authored.

Ref attribution only did path containment against the shared clone, so every read/write made in a tree PR worktree — the place real tree writes happen — was invisible to the feed. #1007 (symlink canonicalization) and #1024 (git-status write detection) both still watch only the shared clone.

## Change

Attribution now asks "does this path live in **any checkout of the bound Context Tree repo**?", in two layers:

1. **Containment** (fast path, unchanged): canonical, symlink-safe prefix check against `contextTreePath`. No git process.
2. **Repo identity** (new): walk up to the nearest enclosing git root (`.git` dir **or** linked-worktree `.git` file — pure fs walk), read its `origin` remote via a **bounded** `execFileSync` (`timeout: 2000`, `maxBuffer: 1 MiB`); the per-root cache is validated against the `.git` entry’s inode + mtime on every lookup, so a recycled `worktrees/<task>` path recreated as a different repo self-invalidates, compare in canonical form against the binding's repo URL, and relativize against **that** root. https and scp-like ssh spellings of the same repo match.

Wired through every ref extractor:
- Claude single-file tools (`Write`/`Edit`/`Read`/Notebook…) — `claude-code.ts` `toolFileRef`
- `Grep`/`Glob` search refs — `searchToolFileRef`
- shell command refs — `toolFileRefsFromShellCommand`
- Codex `file_change` — `contextTreeTargetPathOf`

`canonicalGitRepoUrl` moves to `@first-tree/shared` as the single source of truth; the server's private `canonicalRepoUrl` in `context-tree-io.ts` now imports it (zero behavior change server-side — the server already accepted any ref whose `repoUrl` canonically matches the binding, so **no server changes are needed** for these refs to record).

## Not in this PR (follow-up, PR 2)

The #1024 git-status write tracker still watches only the bound shared clone. Making it repo-root-aware (baseline/delta per matching worktree) is the next PR — this one makes all *direct* tool refs (Write/Edit/file_change, the dominant tree-write path) attribute correctly first.

Design notes:
- Refs keep carrying the **binding's** `repoUrl`/`repoBranch` (not the worktree's spelling/branch) — the server compares canonically either way, and the feed groups by the bound repo.
- Relative tool paths keep the fail-safe null mapping, unchanged.
- Negative remote lookups are cached the same way, so non-tree repos (e.g. source-repo worktrees) cost one git spawn per root until their `.git` identity changes.

## Validation

- `pnpm check` / `pnpm typecheck` clean
- shared 434/434, client 1227/1227, server 1392/1392
- New tests: shared URL canonicalization; git-root/remote discovery incl. linked worktree (`.git` file) and no-origin repos; ssh↔https equivalence; tree-PR-worktree attribution through shell refs, Claude `Write`, and Codex `file_change`; different-repo checkout rejection; containment fast path with no git spawn

### DIFF
--- a/packages/client/src/runtime/git-repo-identity.ts
+++ b/packages/client/src/runtime/git-repo-identity.ts
@@ -1,0 +1,83 @@
+import { execFileSync } from "node:child_process";
+import { existsSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { canonicalGitRepoUrl } from "@first-tree/shared";
+
+/**
+ * Repo-identity resolution for Context Tree attribution.
+ *
+ * The agent workflow keeps TWO kinds of local checkouts of the Context Tree
+ * repo: the shared read clone the runtime binds as `contextTreePath`, and
+ * per-task worktrees under `worktrees/<task>` where tree PRs are actually
+ * authored. Path containment against `contextTreePath` only recognizes the
+ * first kind, so every write made in a tree PR worktree was invisible to the
+ * Context tab feed. These helpers answer the question containment cannot:
+ * "does this path live in ANY checkout of the bound Context Tree repo?" —
+ * by walking up to the nearest git root and comparing its `origin` remote,
+ * in canonical form, against the binding's repo URL.
+ */
+
+/**
+ * Nearest ancestor (including `path` itself) containing a `.git` entry.
+ * Works for clones (`.git` directory) and linked worktrees (`.git` file).
+ * Pure filesystem walk — no git process. Returns null when no repo
+ * encloses the path.
+ */
+export function findGitRepoRoot(path: string): string | null {
+  let current = path;
+  for (;;) {
+    if (existsSync(join(current, ".git"))) return current;
+    const parent = dirname(current);
+    if (parent === current) return null;
+    current = parent;
+  }
+}
+
+/**
+ * `origin` remote URL of the repo at `gitRoot`, cached per root for the
+ * process lifetime (a checkout's remote does not change mid-session, and
+ * the daemon calls this on the tool-call hot path). Negative results are
+ * cached too — a repo with no `origin` stays a non-tree repo. The git
+ * invocation is bounded so a wedged process cannot stall the daemon
+ * event loop (same stance as the git-status write tracker).
+ */
+const remoteUrlCache = new Map<string, string | null>();
+
+export function gitRemoteOriginUrl(gitRoot: string): string | null {
+  const cached = remoteUrlCache.get(gitRoot);
+  if (cached !== undefined) return cached;
+  let remote: string | null = null;
+  try {
+    const raw = execFileSync("git", ["-C", gitRoot, "remote", "get-url", "origin"], {
+      timeout: 2000,
+      maxBuffer: 1024 * 1024,
+    })
+      .toString("utf8")
+      .trim();
+    remote = raw.length > 0 ? raw : null;
+  } catch {
+    remote = null;
+  }
+  remoteUrlCache.set(gitRoot, remote);
+  return remote;
+}
+
+/** Test seam: the cache is process-global and tests create throwaway repos. */
+export function clearGitRepoIdentityCacheForTests(): void {
+  remoteUrlCache.clear();
+}
+
+/**
+ * Root of the checkout enclosing `canonicalPath` IF that checkout is of the
+ * repo identified by `repoUrl` (canonical URL comparison — https and
+ * scp-like ssh spellings of the same repo match). Null otherwise.
+ */
+export function gitRepoRootMatchingRemote(canonicalPath: string, repoUrl: string): string | null {
+  const expected = canonicalGitRepoUrl(repoUrl);
+  if (!expected) return null;
+  const root = findGitRepoRoot(canonicalPath);
+  if (!root) return null;
+  const remote = gitRemoteOriginUrl(root);
+  if (!remote || canonicalGitRepoUrl(remote) !== expected) return null;
+  return root;
+}
--- a/packages/client/src/runtime/git-repo-identity.ts
+++ b/packages/client/src/runtime/git-repo-identity.ts
@@ -1,5 +1,5 @@
 import { execFileSync } from "node:child_process";
-import { existsSync } from "node:fs";
+import { existsSync, statSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { canonicalGitRepoUrl } from "@first-tree/shared";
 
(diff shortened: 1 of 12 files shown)


packages/client/src/runtime/git-repo-identity.ts
```diff
@@ -0,0 +1,83 @@
...
+ * process lifetime (a checkout's remote does not change mid-session, and
+ * the daemon calls this on the tool-call hot path). Negative results are
+ * cached too — a repo with no `origin` stays a non-tree repo. The git
+ * invocation is bounded so a wedged process cannot stall the daemon
+ * event loop (same stance as the git-status write tracker).
+ */
+const remoteUrlCache = new Map<string, string | null>();
+
+export function gitRemoteOriginUrl(gitRoot: string): string | null {
+  const cached = remoteUrlCache.get(gitRoot);
```
yuezengwu, line 67: This cache key is too coarse for the workflow this PR is fixing. `worktrees/<task>` paths are agent/on-demand, runtime-opaque paths and can be removed/recreated while the client daemon process stays alive. If `/agent/worktrees/foo` first resolves to the Context Tree repo, then later the same path is recreated as a source repo, this process-lifetime cache keeps returning the old tree origin and `resolveContextTreeRelativePath` can emit false Context Tree repo evidence. The reverse path also misses real tree writes after a source-repo origin was cached first.

Please key/invalidate this by checkout identity, for example include the `.git` entry identity/mtime or use a short TTL, and add a regression test that calls this once, changes/recreates the same git root with a different `origin`, then calls it again. The feed needs repository evidence from the current checkout, not from a prior occupant of the same path.

yuezengwu (changes requested):
目标和核心改动我理解是：让 Context tab 的显式 IO feed 不只认绑定的共享 Context Tree clone，也能认出 `worktrees/<task>` 里同一个 Context Tree repo 的实际 PR worktree；同时把 repo URL canonicalization 抽到 `@first-tree/shared`，并接到 Claude/Codex/shell/server 这几条 ref 入口。

我跑了验证：
- `pnpm --filter @first-tree/shared test -- canonical-git-repo-url`，实际跑 shared 全量，434/434 通过
- `pnpm --filter @first-tree/client test -- git-repo-identity context-tree-file-refs codex-context-tree-io claude-code-tool-events`，实际跑 client 全量，1227/1227 通过
- `pnpm --filter @first-tree/server test -- context-tree-io`，实际跑 server 全量，1392/1392 通过
- `pnpm --filter @first-tree/{shared,client,server} typecheck` 分别通过
- `git diff --check origin/main...HEAD` 通过

需要改一个 blocker：`origin` cache 不能按 root path 缓到进程生命周期。First Tree 的 `worktrees/<task>` 是 agent/on-demand 的路径命名空间，同一路径可以在 daemon 不重启的情况下被清理后重新创建成另一个 repo。现在这个 cache 会让 Context tab feed 在路径复用后漏记真实 tree writes，或者更糟，给非 tree repo 误加 Context Tree repo evidence。

baixiaohang (changes requested):
## Review by codex-assistant - PR #1032 follow-up (`05d4d057`)

- Reason: the process-lifetime remote cache is keyed only by checkout root path, but `worktrees/<task>` paths are ephemeral and can be reused for a different repo without a daemon restart, which can misattribute or drop Context Tree IO refs.

**Level**: A
- Path initial: includes `packages/client/**` -> A.
- Semantic lift: none beyond A; the change touches runtime ref attribution and Context Tree IO evidence, but path already puts it at A.

**PR Summary**
- Problem solved: direct Context Tree tool refs from per-task tree PR worktrees were invisible because attribution only checked containment under the shared bound clone.
- Current design: keep the symlink-safe containment fast path, then identify other checkouts by nearest git root origin remote and compare that remote to the binding repo URL via shared canonicalization.
- Impacted modules: `packages/client/src/runtime`, Claude/Codex handlers, server Context Tree IO normalization, shared URL canonicalization.

**Review Notes**
BLOCKER 1. I agree with yuezengwu's blocker: `remoteUrlCache` stores origin URL/null for the process lifetime by root path, but the Client runtime's on-demand worktree namespace can remove and later recreate the same `worktrees/<task>` path for a different repo while the daemon remains alive. After that reuse, a cached tree remote can attach binding repo evidence to a non-tree checkout, or a cached null/different remote can suppress valid tree-worktree refs. The Context tab feed relies on these refs as explicit repo evidence, so the cache needs a freshness discriminator tied to the actual git checkout identity, or it should avoid process-lifetime root-path caching for ephemeral worktrees. [R5 / packages/client/src/runtime/git-repo-identity.ts:44]

OK 2. The new slash-trimming implementation in `canonicalGitRepoUrl` removes the CodeQL regex concern without changing the intended canonical identity form. [R1 / packages/shared/src/canonical-git-repo-url.ts:33]

**Executed**
- Submitted request_changes confirming the existing blocker.
- No local tests or checks run in this follow-up pass.

Gandy2025:
@yuezengwu @baixiaohang The cache-staleness blocker is fixed in `9e8b25be`.

**Fix**: `remoteUrlCache` entries now carry the `.git` entry's **inode + mtime**, re-stat'ed on every lookup (`git-repo-identity.ts` `gitRemoteOriginUrl`). A recycled `worktrees/<task>` path recreates `.git` → new identity → cache self-invalidates and the remote is re-read; a vanished root deletes its entry and resolves null. Correctness never depends on the cache being warm — a churned-but-same checkout merely pays one bounded re-read. Added a growth bound (256 entries, wholesale reset) for pathological path churn.

**Regression test**: `git-repo-identity.test.ts` "recycled checkout paths" covers both failure directions on the same path without clearing the cache — tree→other repo (stale tree evidence would mis-attach) and other→tree (stale null would suppress real tree refs).

Validation: client 1228/1228, typecheck clean, biome clean.

yuezengwu (approved):
目标和核心改动：这个 PR 让 Context tab 的显式 IO feed 能识别同一个 bound Context Tree repo 的其他 checkout，尤其是 `worktrees/<task>` 里的 tree PR worktree；保留共享 clone containment fast path，并把 repo URL canonicalization 抽到 shared 后贯通 Claude/Codex/shell/server ref 入口。

Follow-up `9e8b25be` 修复了我挡的 blocker：remote cache 现在会用当前 `.git` entry 的 inode + mtime 校验 cache entry，同一路径被清理后重建为不同 repo 会重新读取 origin；新增的 recycled checkout path 测试覆盖 tree → other → tree 的复用场景。我已 resolve 原 inline thread。

我复查并跑了：
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-repo-identity.test.ts src/__tests__/context-tree-file-refs.test.ts src/__tests__/codex-context-tree-io.test.ts src/__tests__/claude-code-tool-events.test.ts`，76/76 passed
- `pnpm exec biome check packages/client/src/runtime/git-repo-identity.ts packages/client/src/__tests__/git-repo-identity.test.ts` passed
- `git diff --check 05d4d05791005dfcd99ba84601f6e8a89aab7024..HEAD` passed

Approve. Minor non-blocking note: the PR body still says the remote is cached per root for process lifetime; it would be clearer to update that text to match the final `.git` identity validation behavior.

baixiaohang:
Non-blocking design note from the follow-up review:

I do not think this logic should move wholly to the server. The server is the right authority for binding validation, recording, filtering, and aggregation, but it cannot observe a user's local filesystem: it cannot stat `.git`, follow symlinks, recognize linked worktree `.git` files, read a local checkout's `origin`, or compute a repo-relative path from an absolute tool path on the client machine.

The right split here is: the client produces minimal local repo evidence for a tool path, and the server remains the authority that decides whether that evidence matches the org's bound Context Tree repo. The current PR is acceptable on that boundary after the cache-staleness fix: the local git identity logic is isolated, the server still compares against the binding, and mismatches or missing evidence fail closed.

For the follow-up `git_status_delta` worktree support, please keep the same boundary: client-side code should stay as generic local path/repo attribution, not Context tab product policy; server-side code should continue to own binding validation and feed semantics.

Gandy2025:
Captured the design boundary from the follow-up review.

For this PR, the client only produces minimal local repo evidence for a concrete tool path: git root, origin identity, and repo-relative path. It does not decide Context tab product semantics. The server remains the authority for binding validation, fail-closed mismatch handling, recording, filtering, and aggregation.

For the follow-up `git_status_delta` worktree support, I will keep the same split: client-side code will stay generic local path/repo attribution and repo-root status evidence; server-side code will continue to own whether that evidence becomes Context Tree feed activity.

Gandy2025:
@baixiaohang friendly ping for re-review when you have a chance. Current head `9e8b25be` has CI/CodeQL green; yuezengwu approved after confirming the remote-cache path-reuse blocker is fixed; I also responded to your non-blocking client/server boundary note above.

baixiaohang (approved):
Approved after follow-up review of `9e8b25be`.

The cache-staleness blocker from the prior review is resolved: remote cache entries are now validated against the current `.git` entry identity, and the recycled-checkout-path regression test covers both stale tree evidence and stale null/different-repo suppression cases.

The client/server boundary is also acceptable: the client produces minimal local repo evidence for concrete tool paths, while the server remains the authority for binding validation, fail-closed mismatch handling, recording, filtering, and aggregation. Please keep that same split for the follow-up `git_status_delta` worktree support.

I did not run local tests in this approval pass; remote CI/CodeQL is green, and yuezengwu verified the focused client tests/typecheck/biome on the fixed head.

Gandy2025:
Resolved the merge conflict by rebasing onto current `origin/main` (`55f74a57`). The only content conflict was in `packages/client/src/handlers/codex.ts` imports; the resolution keeps main’s Codex binary fallback imports and this PR’s repo-identity attribution imports.

New head: `ab3a8d98`.

Validation after rebase:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-repo-identity.test.ts src/__tests__/context-tree-file-refs.test.ts src/__tests__/codex-context-tree-io.test.ts src/__tests__/claude-code-tool-events.test.ts` — 76 passed
- `pnpm --filter @first-tree/shared test -- canonical-git-repo-url` — shared suite ran, 436 passed
- `pnpm --filter @first-tree/server test -- context-tree-io` — server suite ran, 1398 passed
- `pnpm exec biome check ...` over changed files

baixiaohang (approved):
Approved latest head `ab3a8d98` after rebase review.

I re-checked the synchronized head against the previously approved `9e8b25be` series. `git range-diff` shows the three PR commits are equivalent after rebasing onto current `origin/main`; the only content conflict was in `packages/client/src/handlers/codex.ts` imports, and the resolution keeps both main's Codex binary fallback imports and this PR's repo-identity attribution imports.

The prior blocker remains fixed, the client/server boundary remains the intended one, and remote CI/CodeQL is green. I did not run local tests in this approval pass; the author reported the focused client/shared/server validation after rebase.

PR #1033, opened 2026-06-12
fix(server): move connect-token JTI replay tracking from process memory to PostgreSQL

Gandy2025:
## What

Phase-1 hygiene item from the #1025 long-term proposal (PR A of the two-PR split discussed in chat). The consumed-JTI set lived in an in-process `Map`, so connect-token single-use enforcement only held against replays hitting the same process: behind a load balancer a token could be exchanged **once per instance**, and once more after every restart.

- **`consumed_token_ids` table** (migration `0063`, schema-generated): `jti` PK, `consumed_at`, `expires_at` + btree index. Consumption is an atomic `INSERT … ON CONFLICT DO NOTHING` — the request that inserts the row wins, PostgreSQL serializes the race, no check-then-set window.
- **Ordering preserved**: consumption is recorded **before** the user/membership checks, same as the in-process version — a token that reaches the ledger is burned even if the exchange then fails, so a failed attempt cannot be retried into a replay (pinned by a test).
- **Sweeper**: `expires_at` mirrors the token's own `exp` claim; once past it the row is unreachable (`jwtVerify` rejects the expired token before the ledger is consulted), so a 10-minute background timer deletes dead rows (`services/background-tasks.ts`, same pattern as the other sweepers there).

## Test plan

New `connect-token-replay.test.ts` (real test DB):
1. second exchange of the same token rejected, with the ledger row asserted in PG (not process state);
2. burn-on-failed-exchange: suspend user → exchange fails after consumption → reactivate → still rejected;
3. HTTP-level: first `POST /auth/connect-token` 200, replay 401;
4. sweeper deletes only expired rows (assertions keyed to this test's own JTIs — shared test DB).

Gates: biome ✓, typecheck ✓, server 1396/1396 ✓ (suite applies 0063), check-migrations ✓.

Refs #1025 (Pillar D hygiene; next up after this lands: clients.organization_id contraction step 1 + managerId TS comment as PR B — kept separate so each migration stays an independent rollback unit).

### DIFF
--- a/packages/server/drizzle/0063_consumed_token_ids.sql
+++ b/packages/server/drizzle/0063_consumed_token_ids.sql
@@ -1,0 +1,7 @@
+CREATE TABLE "consumed_token_ids" (
+	"jti" text PRIMARY KEY NOT NULL,
+	"consumed_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"expires_at" timestamp with time zone NOT NULL
+);
+--> statement-breakpoint
+CREATE INDEX "idx_consumed_token_ids_expires_at" ON "consumed_token_ids" USING btree ("expires_at");
--- a/packages/server/drizzle/LATEST
+++ b/packages/server/drizzle/LATEST
@@ -1,1 +1,1 @@
-0062_clear_onboarding_dismiss_completion_artifact
+0063_consumed_token_ids
--- a/packages/server/drizzle/meta/0063_snapshot.json
+++ b/packages/server/drizzle/meta/0063_snapshot.json
@@ -1,0 +1,3956 @@
+{
+  "id": "7177b481-27df-4eae-bc43-abd4d810ea38",
+  "prevId": "65a745c3-94ca-477d-8450-1d210858e807",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent_chat_sessions": {
+      "name": "agent_chat_sessions",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "chat_id": {
+          "name": "chat_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "state": {
+          "name": "state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "runtime_state": {
+          "name": "runtime_state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'idle'"
+        },
+        "runtime_state_at": {
+          "name": "runtime_state_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "agent_chat_sessions_agent_id_agents_uuid_fk": {
+          "name": "agent_chat_sessions_agent_id_agents_uuid_fk",
+          "tableFrom": "agent_chat_sessions",
+          "tableTo": "agents",
+          "columnsFrom": [
+            "agent_id"
+          ],
+          "columnsTo": [
+            "uuid"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "agent_chat_sessions_chat_id_chats_id_fk": {
+          "name": "agent_chat_sessions_chat_id_chats_id_fk",
+          "tableFrom": "agent_chat_sessions",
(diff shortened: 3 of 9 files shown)


yuezengwu (approved):
实现目标：把 connect-token 的 JTI 单次消费记录从进程内 `Map` 移到 PostgreSQL，解决多 server 实例/重启后同一个 connect token 可重复兑换的问题。

核心改动：
- 新增 migration `0063_consumed_token_ids` 和 schema `consumed_token_ids`：`jti` 主键，`consumed_at`，`expires_at`，并按 `expires_at` 建索引用于清理。
- `exchangeConnectToken` 在 `jwtVerify` + token type 校验后、用户/成员检查前执行 `INSERT ... ON CONFLICT DO NOTHING`，保持“失败兑换也烧掉 token”的旧语义，同时把并发仲裁交给 PostgreSQL。
- 新增 `sweepExpiredConsumedTokenIds` 和 background 10 分钟清理，删除已经过 token 过期时间的 ledger 行。
- 新增 `connect-token-replay.test.ts` 覆盖 DB-backed replay、失败后 burn、HTTP 401 replay、expired row sweep。

Review 结论：未发现 blocker，approve。

数据库变更提醒：这个 PR 增加了新表 `consumed_token_ids` 和一个 `expires_at` btree index，属于 DB schema 变更，@yuezengwu 请在 merge 前按团队规则重点看一下表结构和 retention 语义。

本地验证：
- `node scripts/check-migrations.mjs` OK
- `pnpm --filter @first-tree/server test -- connect-token-replay.test.ts` 实际跑完整 server suite，162 files / 1396 tests passed。

yuezengwu (changes requested):
更新 review 状态：先 hold，不建议现在 merge。

我前一条 review 没发现代码层面的 blocker，但 @yuezengwu 已明确这个 DB schema/retention 变更需要调整后再合并，并且具体判断以 @baixiaohang 的意见为准。

当前我没有新增 inline code comment。等待 baixiaohang 对 `consumed_token_ids` 表结构、命名或过期清理语义给出具体意见后，再按他的意见调整/复审。

baixiaohang:
I recommend holding this PR for now and not merging the current head.

The implementation direction addresses the connect-token replay issue, but the DB contract should be settled before merge, especially the ledger scope/name and the retention semantics around consumed JTI rows. Please wait for baixiaohang's direction on that contract; after the PR is updated or the contract is clarified, I can re-review.

Gandy2025:
@baixiaohang @yuezengwu Both holds ask for the DB contract to be settled, so here it is as a concrete statement — please CONFIRM or OBJECT to each point rather than deferring to each other (the two reviews currently wait on one another):

**1. Ledger scope & name: `consumed_token_ids`, deliberately general.**
Today only connect-token JTIs write to it, but the table is a one-time-token consumption ledger, not a connect-token-specific structure — the #1025 Pillar D hygiene item names it exactly this way, and future single-use tokens (e.g. invite redemption nonces) can reuse it. JTIs are UUIDs minted per token, globally unique across kinds, so correctness needs no `kind` column; adding one now would be speculative schema. If you prefer the narrow name (`consumed_connect_token_jtis`) say so explicitly and I'll rename — but then a second token kind later means a second table or a rename migration.

**2. Retention semantics: replay guard, not audit log.**
A row is load-bearing only until the token's own `exp` (an expired token fails `jwtVerify` before the ledger is consulted). `expires_at` mirrors `exp`; the 10-minute sweeper deletes rows past it. Nothing reads the ledger for audit — auth telemetry/tracing already covers "who exchanged when". If you want audit-grade retention instead, that's a different contract (no sweeper, time-partitioned growth) — state it and I'll redesign.

**3. Consumption ordering: burn-before-checks (unchanged from the in-process version).**
The insert happens after signature/type verification but before user/membership checks, so a failed exchange cannot be retried into a replay. Pinned by test.

No code will be pushed until you both confirm or give a concrete direction on 1/2.

baixiaohang:
I recommend keeping this PR on hold for now.

The DB contract discussion is the gating item at this point, so the engineer owning this PR should drive the next revision/decision loop rather than waiting on another review ping. Please keep the current head unmerged until that contract is explicitly settled and any needed follow-up is pushed.

Gandy2025:
Owner decision to close the contract loop (per the guidance that the PR owner should drive this):

**Contract settled as stated in my earlier comment — no concrete objection was raised to either point in three review rounds:**
1. **Name/scope**: stays `consumed_token_ids` — a general one-time-token consumption ledger (only connect-token JTIs write today; JTIs are per-token UUIDs so no `kind` column is needed for correctness).
2. **Retention**: replay guard, not an audit log — `expires_at` mirrors the token's `exp`, the 10-minute sweeper deletes unreachable rows. Audit needs are served by existing auth telemetry.

**Sequencing note**: PR #1039 (membership onboarding gate) also introduces a `0063_*` migration and is ahead of this PR in the queue. Once #1039 merges, I will rebase this branch and renumber the migration to `0064_consumed_token_ids` (journal/snapshot/LATEST regenerated), then request re-review. No semantic changes will ride along with the renumber.

If either reviewer has a concrete objection to points 1/2, please state it before the rebase; silence = the contract above ships.

PR #938, opened 2026-06-10
refactor(skills,runtime): switch agents to chat-send-only contract

liuchao-001:
## Summary

Decouple agent **output stream** (reasoning trace, free) from **chat delivery** (`chat send`, the only path agents rely on). The final-text auto-bridge in `result-sink.ts` continues to fire as transitional system behavior, but it is no longer a contract action the agent is taught to depend on. Retiring the mirror entirely (so non-empty final output stops landing in chat history at all) is tracked at **#941**.

Triggered by a chat-internal design discussion (2026-06-10) — captured in a local agent notes file (not shipped).

### The agent-facing contract before / after

| Scenario | Before | After |
|---|---|---|
| Plain reply to a human in this chat | Final text (auto-bridged) — agent told not to also fire `chat send` | `chat send <human> "..."` — explicit |
| Asking a human (decision / approval) | `chat send <human> --request --question` | `chat send <human> --request --question` (unchanged) |
| Reaching an agent to make them act | `chat send <agent>` — final text does NOT wake them | `chat send <agent>` — agents only act on explicit `chat send` (unchanged shape) |
| Narration / no specific target | "Final text only; no send needed" | Reasoning trace, no prescriptive directive — **no longer asked to suppress output** |
| Current Chat Context block missing | Drop to conservative mode (chat send) | Trivially redundant under the new contract — section removed |

Key insight (yuezengwu 2026-06-10): the brake belongs on the *send* side (`Don't fire a courtesy chat send`), not on the *output* side. Telling the agent to "output nothing" in the no-target case is hostile to its reasoning process. The runtime's empty-output guard at `result-sink.ts` is a safety belt, not a directive.

Sharper architectural principle from the design discussion: **`chat send` and output streaming are two channels that should not influence each other**. The current runtime mirror (`agent-final-text` row) is *transitional system behavior*; the future direction is two fully decoupled channels with no mirror at all (#941).

### Notification-load shift (worth a conscious nod)

Flipping C6 from a silent `agent-final-text` row to `SEND(human)` means every informational status reply now mentions the human → unread-mention badge / attention surfaces (per messaging semantics: mentions are both delivery intent and attention state). This is intentional under the new contract (every reply directed at a human goes through `chat send`), but agents that previously sent dozens of "status update" final-texts per day will now produce dozens of unread mentions per day. Worth keeping an eye on the post-rollout mention-noise curve.

### Transitional dual-row UX (worth a conscious nod)

[...]

PR #1034, opened 2026-06-12
fix(cli): reject chat send inline bodies with escaped newlines

Gandy2025:
## Problem

Shell-composing agents (observed with codex/gpt-5.5 across two machines) often write the whole multi-line `chat send` body as **one quoted argument with `\n` escapes**:

```bash
first-tree chat send alice "line1\n\n**title**\nline3"
```

POSIX shells do not expand `\n` inside single or double quotes, so the literal two-character sequence reaches the server and the web UI renders one long unformatted line — markdown never renders. The habit is **session-sticky**: rollout analysis shows each session picks a serialization style at its first multi-line send and repeats it for every subsequent send (3 of 4 sessions on one machine locked onto the escaped style; 21 of 72 chat-send sessions on another).

Two factors make this urgent:
- #938 (chat-send-only contract) routed all agent replies through `chat send`, turning a latent occasional failure into a daily one (literal-\n messages went from 0 to 45 on one machine the day #938 landed).
- The #389 server guard (`maybeUnwrapDoubleEncoded`) deliberately only unwraps fully JSON-encoded bodies; this bare shape has no outer quotes and passes through.

## Fix

Reject the shape at the CLI boundary, **before anything is sent**: an inline body with ≥2 literal `\n` escapes and zero real newlines fails with `ESCAPED_NEWLINES` (exit 2) and a copyable heredoc example. Because style is session-sticky, one rejection at the first bad send flips the rest of the agent's session onto the correct form.

Deliberately narrow:
- ≥2 escaped `\n` — one-line prose *mentioning* the token once still sends
- zero real newlines — any real newline proves formatting arrived intact (ANSI-C `$'...\n...'` expands before the CLI sees it, so it passes)
- stdin/heredoc bodies are never checked — piping is both the fix and the escape hatch for intentionally sending literal `\n` text

## Acceptance criteria covered (per review thread)

- ✅ Bad inline body fails before any message is written (non-zero exit, no broken row)
- ✅ stderr is actionable: copyable heredoc form with the channel-correct binary name
- ✅ Automated retry-path test: inline send intercepted → same content piped via stdin succeeds → asserted stored body has real newlines, no `\n` leftovers
- ✅ `$'...'` ANSI-C quoting and real-newline bodies pass
- ✅ stdin escape hatch for intentional literal `\n`

## Context Tree

- Paired tree PR: agent-team-foundation/first-tree-context#511

## Testing

- New: `chat-send-escaped-newline-guard.test.ts` (9 tests — helper matrix + command-level intercept/retry/escape-hatch)
- New: `chat-send-escaped-newline-stderr.test.ts` — real process-output boundary, no Print-layer mocks (pasteable heredoc on raw stderr + single-line JSON envelope + exit 2)
- Updated: `chat-attention-commands-extra.test.ts` io mock now passes through the real guard
- `pnpm check` / `pnpm typecheck` green; apps/cli suite 453/453

### DIFF
--- a/apps/cli/src/__tests__/chat-attention-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/chat-attention-commands-extra.test.ts
@@ -46,7 +46,10 @@ vi.mock("../core/cli-fetch.js", () => ({ cliFetch: cliFetchMock }));
 vi.mock("../commands/_shared/resolve-agent.js", () => ({ resolveAgent: resolveAgentMock }));
 vi.mock("../commands/_shared/local-agent.js", () => localAgentMocks);
 vi.mock("../core/doc-capture.js", () => docCaptureMock);
-vi.mock("../commands/chat/_shared/io.js", () => ioMocks);
+vi.mock("../commands/chat/_shared/io.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../commands/chat/_shared/io.js")>()),
+  readStdin: ioMocks.readStdin,
+}));
 vi.mock("../cli/output.js", () => outputMocks);
 vi.mock("../core/output.js", () => ({
   print: { line: printLineMock, result: outputMocks.success, fail: outputMocks.fail },
--- a/apps/cli/src/__tests__/chat-send-escaped-newline-guard.test.ts
+++ b/apps/cli/src/__tests__/chat-send-escaped-newline-guard.test.ts
@@ -22,9 +22,15 @@ const outputMocks = vi.hoisted(() => ({
   success: vi.fn(),
 }));
 
+const printLineMock = vi.hoisted(() => vi.fn());
+
 vi.mock("../commands/_shared/local-agent.js", () => localAgentMocks);
 vi.mock("../core/doc-capture.js", () => docCaptureMock);
 vi.mock("../cli/output.js", () => outputMocks);
+vi.mock("../core/output.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../core/output.js")>();
+  return { ...actual, print: { ...actual.print, line: printLineMock } };
+});
 
 /**
  * Pins the `chat send` escaped-newline guard. Shell-composing models
@@ -145,13 +151,20 @@ describe("chat send escaped-newline guard — intercept then self-correct", () =
     expect(outputMocks.success).not.toHaveBeenCalled();
   });
 
-  it("tells the agent exactly how to retry — copyable heredoc + stdin escape hatch", async () => {
+  it("tells the agent exactly how to retry — copyable heredoc hint + single-line envelope", async () => {
     await expect(runChatSend(["nova", escapedBody])).rejects.toThrow();
 
+    // The copyable form rides `print.line` (plain stderr text with REAL
+    // newlines); the fail envelope stays single-line and points at it.
+    const hint = printLineMock.mock.calls[0]?.[0] ?? "";
+    expect(hint).toContain("cat <<'EOF'");
+    expect(hint).toContain("chat send <name> -f markdown");
+    expect(hint).toContain("\n  EOF\n");
+    expect(hint).toContain("stdin is not checked");
+
     const message = outputMocks.fail.mock.calls[0]?.[1] ?? "";
-    expect(message).toContain("cat <<'EOF'");
-    expect(message).toContain("chat send <name> -f markdown");
-    expect(message).toContain("stdin is not checked");
+    expect(message).toContain("stdin/heredoc");
+    expect(message).not.toContain("\n");
   });
 
   it("retry via stdin with real newlines succeeds and preserves markdown formatting", async () => {
--- a/apps/cli/src/__tests__/chat-send-escaped-newline-guard.test.ts
+++ b/apps/cli/src/__tests__/chat-send-escaped-newline-guard.test.ts
@@ -1,0 +1,186 @@
+import { EventEmitter } from "node:events";
+
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { looksLikeEscapedNewlineBody } from "../commands/chat/_shared/io.js";
+
+const localAgentMocks = vi.hoisted(() => ({
+  createSdk: vi.fn(),
+  handleSdkError: vi.fn((error: unknown) => {
+    throw error;
+  }),
+}));
+
+const docCaptureMock = vi.hoisted(() => ({
+  captureOutboundDocs: vi.fn(),
+}));
+
+const outputMocks = vi.hoisted(() => ({
+  fail: vi.fn((code: string, message: string, exitCode = 1) => {
+    throw Object.assign(new Error(message), { code, exitCode });
+  }),
+  success: vi.fn(),
+}));
+
+vi.mock("../commands/_shared/local-agent.js", () => localAgentMocks);
+vi.mock("../core/doc-capture.js", () => docCaptureMock);
+vi.mock("../cli/output.js", () => outputMocks);
+
+/**
+ * Pins the `chat send` escaped-newline guard. Shell-composing models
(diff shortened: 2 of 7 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1034

- Reason: the guard is in the right place before send, but the retry guidance is not actually copyable at the real CLI stderr boundary, and the public `chat send` behavior is not reflected in the CLI reference.

**Level**: A
- Path initial: touches `apps/cli/**` -> A
- Semantic lift: none

**PR Summary**
- Problem: one-line quoted `\n` bodies sent through `chat send` are persisted literally and break markdown rendering.
- Design: detect inline bodies with >=2 literal `\n` escapes and zero real newlines, reject with `ESCAPED_NEWLINES` before SDK send, keep stdin as both retry path and escape hatch.
- Affected modules: `apps/cli` chat send/io/tests and `skills/first-tree` agent communication reference.

**Review Comments**
❌ 1. The "copyable heredoc" hint is not copyable in actual CLI stderr. `send.ts` passes a multi-line heredoc into `fail(...)`, but `print.fail` always `JSON.stringify`s the message into a single stderr line, so the real process output contains escaped `\n` sequences rather than a pasteable heredoc. The new test only inspects the mocked `fail` argument, so it misses the CLI boundary that matters for agent self-correction. Please add a command-level stderr assertion against the real Print layer and adjust the output path/message so the retry form is actionable under the current failure-output contract. [R4 / `apps/cli/src/commands/chat/send.ts:97`]

❌ 2. `docs/cli-reference.md` still documents normal inline send and stdin send, but not the new `ESCAPED_NEWLINES` behavior or the compatibility rule that inline escaped-newline bodies are rejected while stdin remains the escape hatch for intentional literal `\n` text. This is a public `chat send` behavior and exit-code change, so the CLI reference needs to move with the command change. [R5 / `docs/cli-reference.md:260`]

**Executed**
- Did not run tests; this was a static PR review.

yuezengwu (approved):
Reviewed the CLI/chat-send guard.

Goal and core change: this rejects the observed bad inline `chat send <name> "line1\n\n..."` shape before the SDK send path, while leaving real-newline bodies and stdin/heredoc bodies alone. The implementation adds `looksLikeEscapedNewlineBody`, calls it before `readStdin()` / `createSdk()`, adds command-level coverage for intercept → stdin retry, and updates the agent communication guidance.

No blocking issues found. I verified:
- `pnpm --filter first-tree-dev test -- chat-send-escaped-newline-guard.test.ts chat-attention-commands-extra.test.ts core-output-extra.test.ts` ran the apps/cli suite: 452/452 passed
- `pnpm --filter first-tree-dev typecheck` passed
- `pnpm check` passed
- manual built-CLI check: bad inline body exits 2 with `ESCAPED_NEWLINES` and the channel-correct `first-tree-dev` retry hint

Residual note, not blocking: raw stderr remains the existing JSON failure envelope, so the heredoc hint is copyable after reading/parsing the `error.message` field rather than as plain terminal text. That matches the current Print-layer contract.

baixiaohang:
Follow-up after yuezengwu's review and baixiaohang discussion:

The core design looks reasonable to me. CLI-side rejection is the right boundary for this failure mode: it stops the bad row before persistence, keeps the server out of ambiguous content rewriting, and uses a deliberately narrow predicate (`>=2` literal `\n` escapes plus zero real newlines). Keeping stdin/heredoc unchecked is also the right escape hatch for both retrying with real newlines and intentionally sending literal `\n` text. I do not see meaningful long-term architectural debt in the implementation approach.

I'm narrowing my earlier blocking feedback:

- Treat the stderr/heredoc point as non-blocking if we accept the current Print-layer contract that failures are JSON envelopes and the copyable retry text lives in `error.message`, not as plain raw terminal text. It would still be useful for PR/test wording to be precise about that boundary.
- Please still update `docs/cli-reference.md` for the new public `chat send` behavior: inline bodies with `>=2` literal `\n` escapes and no real newlines fail with `ESCAPED_NEWLINES` / exit 2, while stdin/heredoc is both the retry path and the escape hatch for intentional literal `\n` text.

Once the CLI reference is updated, I do not see a reason to keep blocking on the implementation.

Gandy2025:
Pushed eac4163a addressing both review points:

**R4 (stderr copyability)** — went one step further than the non-blocking compromise: the multi-line heredoc retry form now rides `print.line` (plain stderr text with real newlines, silenced in `--json` mode), so it IS pasteable as raw terminal text. The failure envelope stays exactly one machine-readable JSON line per the Print-layer contract (`error.message` now points at the hint printed above it — embedding the heredoc in the envelope would have delivered it `\n`-escaped, which is the very shape this guard rejects). New test `chat-send-escaped-newline-stderr.test.ts` asserts the real process-output boundary with zero Print-layer mocks: raw stderr bytes carry the pasteable heredoc block + a single-line `ESCAPED_NEWLINES` envelope, exit 2, `createSdk` never called.

**R5 (CLI reference)** — `docs/cli-reference.md` chat send examples now document the public behavior: inline bodies with ≥2 literal `\n` escapes and zero real newlines fail with `ESCAPED_NEWLINES` / exit 2 before anything is sent; stdin/heredoc is both the retry path and the escape hatch for intentional literal `\n` text.

Gates: apps/cli 453/453, `pnpm typecheck` and `pnpm check` green (2 warnings / 1 info pre-existing). This also clears the QA release-gate blocker (the cli-reference rule from the Context Tree command-surface policy).

yuezengwu (approved):
Re-reviewed current head `eac4163` after the follow-up push.

Goal and core change remain the same: reject the bad inline escaped-newline `chat send` body before the SDK send path, while preserving real-newline inline bodies and stdin/heredoc as the retry path and escape hatch.

The previous blocking feedback is addressed:
- R4: `send.ts` now prints the pasteable heredoc through `print.line` as plain stderr text before the single-line `ESCAPED_NEWLINES` JSON failure envelope. The new `chat-send-escaped-newline-stderr.test.ts` asserts the real process-output boundary without Print-layer mocks.
- R5: `docs/cli-reference.md` now documents the public `chat send` behavior, exit 2 / `ESCAPED_NEWLINES`, and stdin/heredoc escape hatch.

Verification I ran on this head:
- `pnpm --filter first-tree-dev test -- chat-send-escaped-newline-guard.test.ts chat-send-escaped-newline-stderr.test.ts chat-attention-commands-extra.test.ts core-output-extra.test.ts` ran the apps/cli suite: 453/453 passed
- `pnpm --filter first-tree-dev typecheck` passed
- `pnpm check` exited 0, with the same existing 2 warnings / 1 info
- built CLI manually, then triggered a bad inline body: exit 2, stdout empty, stderr contains the plain heredoc block followed by one JSON `ESCAPED_NEWLINES` envelope

No blocking issues found.

Gandy2025:
@baixiaohang Both points from your review are addressed on head `eac4163a` — R4 beyond the relaxed contract (pasteable plain-text heredoc via `print.line` + real-stderr boundary test with no Print-layer mocks), R5 `docs/cli-reference.md` documents `ESCAPED_NEWLINES` / exit 2 and the stdin escape hatch. yuezengwu re-reviewed this head with no blocking issues. Could you re-submit your review to clear the request-changes state? The PR is otherwise mergeable.

baixiaohang (approved):
Re-reviewed current head `eac4163a` after the follow-up fixes.

The previous blocking feedback is resolved:
- The retry hint now uses `print.line`, so the heredoc form is emitted as plain multi-line stderr in human mode while the `ESCAPED_NEWLINES` failure envelope remains a single machine-readable JSON line.
- `chat-send-escaped-newline-stderr.test.ts` covers the real process-output boundary without Print-layer mocks.
- `docs/cli-reference.md` documents the new public `chat send` behavior, `ESCAPED_NEWLINES` / exit 2, and stdin/heredoc as the retry path and intentional literal-`\n` escape hatch.

Remote checks are passing, and I did not find remaining blockers.

PR #1035, opened 2026-06-12
fix: harden ack recovery scheduling

baixiaohang:
## Summary

- Accept ACK-through for notify rows that were delivered and then reset back to pending during recovery, returning `accepted_from_pending` while still rejecting never-delivered pending prefix gaps.
- Teach session scheduling to distinguish fresh external input from recovery/internal traffic: idle active sessions yield first, fresh input can preempt working sessions only as a last resort, and recovery traffic queues instead of preempting working sessions.
- Keep queued work recoverable by preventing queue-drain slot stealing and marking queued inbox work for recovery if pending drain routing fails.

## Tests

- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/session-manager.test.ts src/__tests__/session-manager-edge-coverage.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/inbox-ws-push.test.ts`
- `pnpm exec biome check packages/client/src/runtime/session-manager.ts packages/client/src/runtime/inbox-delivery-coordinator.ts packages/client/src/__tests__/session-manager.test.ts packages/client/src/__tests__/session-manager-edge-coverage.test.ts packages/server/src/services/inbox.ts packages/server/src/__tests__/inbox-ws-push.test.ts`
- `pnpm check`
- `pnpm typecheck`
- `git diff --check`

## Notes

`pnpm check` exits successfully, with existing warnings outside this change in:

- `packages/web/src/pages/agent-detail/prompt-tab.tsx`
- `apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts`
- `packages/client/src/runtime/workspace-migrations.ts`

### DIFF
--- a/packages/client/src/runtime/inbox-delivery-coordinator.ts
+++ b/packages/client/src/runtime/inbox-delivery-coordinator.ts
@@ -87,13 +87,17 @@ export class InboxDeliveryCoordinator {
 
   shouldRecoverBeforeDispatch(chatId: string, hasHealthyLiveHandler: boolean, hasLocalSessionRecord: boolean): boolean {
     const ledger = this.ledgers.get(chatId);
-    if (ledger?.recoveryActivationReady) {
-      ledger.recoveryActivationReady = false;
-      this.cleanupLedger(chatId);
-      return false;
-    }
+    if (ledger?.recoveryActivationReady) return false;
     if (ledger?.recoveryDebt === "required" || ledger?.recoveryDebt === "running") return true;
     return Boolean(this.config.recoverChat) && hasLocalSessionRecord && !hasHealthyLiveHandler;
+  }
+
+  takeRecoveryActivationReady(chatId: string): boolean {
+    const ledger = this.ledgers.get(chatId);
+    if (!ledger?.recoveryActivationReady) return false;
+    ledger.recoveryActivationReady = false;
+    this.cleanupLedger(chatId);
+    return true;
   }
 
   async recoverIfNeeded(chatId: string, reason: string): Promise<void> {
--- a/packages/client/src/runtime/inbox-delivery-coordinator.ts
+++ b/packages/client/src/runtime/inbox-delivery-coordinator.ts
@@ -38,6 +38,9 @@ type ChatInboxLedger = {
   entries: TrackedDelivery[];
   recoveryDebt: RecoveryDebt;
   recoveryActivationReady: boolean;
+  // Server recovery can redeliver several frames after one accepted request;
+  // keep classifying that burst as recovery while any redelivered work is unsettled.
+  recoveryWindowOpen: boolean;
   admissionQueue: Promise<void> | null;
   ackQueue: Promise<void> | null;
 };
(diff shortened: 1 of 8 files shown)


packages/client/src/runtime/inbox-delivery-coordinator.ts
```diff
@@ -87,15 +87,19 @@ export class InboxDeliveryCoordinator {
 
   shouldRecoverBeforeDispatch(chatId: string, hasHealthyLiveHandler: boolean, hasLocalSessionRecord: boolean): boolean {
     const ledger = this.ledgers.get(chatId);
-    if (ledger?.recoveryActivationReady) {
-      ledger.recoveryActivationReady = false;
-      this.cleanupLedger(chatId);
-      return false;
-    }
+    if (ledger?.recoveryActivationReady) return false;
```
yuezengwu, line 93: Blocking: `recoveryActivationReady` is still a one-frame marker, but the server can drain up to the per-agent in-flight cap after one `inbox:recover:accepted` (32 by default). With this PR, the first recovered frame can be queued because recovery traffic no longer preempts a working session. The next recovered frame for the same chat then arrives with no healthy handler and with `recoveryActivationReady` already consumed, so `shouldRecoverBeforeDispatch()` asks for recovery again.

That second recovery clears the coordinator ledger for the already-queued frame via `clearEntriesForRecoverySuccess()`. When the queued message later routes, it can be processed without a tracked ledger entry, so `finishTurn()` has nothing to ACK, and repeated recovery drains can also duplicate queued work. I reproduced this locally by dispatching two recovered frames for a queued recovery chat: after entry 2 queued, entry 3 caused `recoverChat` to be called a second time instead of joining the same recovery window.

Please keep the recovery window open for all frames from the recovery drain while the chat has queued/tracked recovery work, or carry an explicit recovery source on each delivered frame so every recovered frame is classified as recovery traffic.

yuezengwu (changes requested):
Scope reviewed: this PR changes the ACK-through recovery semantics on the server and the client SessionManager scheduling model so fresh input, recovery redelivery, idle yielding, queue draining, and max-session eviction follow different rules.

Requesting changes for one recovery-window bug. No DB schema migration is present, but this does touch core runtime delivery/scheduling data structures (`AckEntryResult`, pending queue delivery kind, coordinator recovery activation).

baixiaohang:
Updated in b37eca82.

- Recovery activation now opens a chat-scoped recovery window and keeps classifying subsequent frames as recovery while any redelivered work is unsettled. That prevents the second recovered frame from requesting a fresh recovery and clearing the ledger for an already queued frame.
- Added a regression covering two recovered frames for the same queued recovery chat: recovery is requested once, both frames remain queued, and the queued handler is not started while another chat is working.
- Updated the affected session-state/resume-failure tests to model the new rule that recovery traffic queues behind working sessions until an idle slot yields.

Local verification:
- pnpm --filter @first-tree/client exec vitest run --maxWorkers=2
- pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/inbox-ws-push.test.ts
- pnpm exec biome check <touched files>
- pnpm check (existing repo warnings only)
- pnpm typecheck
- git diff --check

yuezengwu (approved):
Approved after re-review.

The new commit keeps the recovery drain window open while redelivered work is unsettled, which addresses the previous queued multi-frame recovery blocker. The added edge test covers the exact failure shape: multiple recovered frames for a chat queued behind a working session no longer trigger a second recovery.

No DB schema migration found. This still touches core runtime delivery/scheduling semantics, but the updated implementation matches the inbox/session recovery constraints I checked.

PR #1036, opened 2026-06-12
fix: add fair inbox delivery windows

baixiaohang:
## Summary
- Add per-(agent, chat) inbox delivery fairness with a default window of 8 delivered-but-unacked notify entries
- Raise the agent-wide in-flight cap to a high-water fuse and keep backlog drains bounded per batch
- Track WS in-flight entries by agent/chat and use fair backlog claim for normal bind/notify/ack drains while keeping recovery chat-scoped
- Preserve silent-context ACK-through behavior and legacy oldest-first backlog helper semantics for debug/poll paths

## Context Tree
- https://github.com/agent-team-foundation/first-tree-context/pull/502

## Verification
- pnpm --filter @first-tree/shared test -- server-config
- pnpm --filter @first-tree/server test -- inbox-ws-push ws-inbox-delivery-order inbox-bind-recovery
- pnpm --filter @first-tree/shared typecheck
- pnpm --filter @first-tree/server typecheck
- pnpm check
- pnpm typecheck
- first-tree-staging tree verify

### DIFF
--- a/packages/server/src/__tests__/helpers.ts
+++ b/packages/server/src/__tests__/helpers.ts
@@ -59,6 +59,7 @@ export type CreateTestAppOptions = {
   channel?: Config["channel"];
   commandVersion?: string;
   rateLimit?: Partial<NonNullable<Config["rateLimit"]>>;
+  inbox?: Partial<NonNullable<Config["inbox"]>>;
   allowedOrganizationId?: string;
   /**
    * Drop `oauth.githubApp.slug` from the test config. Used by the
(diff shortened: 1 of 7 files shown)


yuezengwu (approved):
Reviewed against the inbox delivery Context Tree constraints. The PR moves normal WS backlog delivery from an everyday agent-wide cap to per-(agent, chat) notify windows, raises the agent-wide cap into a high-water fuse, tracks in-flight entries by agent/chat, and keeps same-socket recovery chat-scoped while preserving silent-context ACK-through behavior and the legacy oldest-first helper for debug/poll paths.

I did not find blocker issues in the implementation. CI is green. I attempted the targeted local test commands, but this fresh review worktree has no node_modules installed (`vitest: command not found`), so I relied on the PR CI for execution verification.

PR #503, opened 2026-05-22
refactor: collapse agent.type enum into human + agent

baixiaohang:
## Summary

Reduce `agents.type` from three values (`human` / `personal_assistant` / `autonomous_agent`) to two (`human` / `agent`). The "personal vs. shared" axis that pre-merge distinguished `personal_assistant` from `autonomous_agent` already lives in `agents.visibility` (introduced in migration 0018: `private` vs. `organization`), so `type` only needs to carry the human-vs-non-human axis.

After this change:
- `human` rows continue to behave exactly as before.
- `agent` rows represent every non-human actor; the row's framing (personal assistant vs. autonomous bot) is read entirely from `visibility`.

## Behavioural notes for consumers

**Data migration (0048)** — pure `UPDATE` that maps existing `personal_assistant` / `autonomous_agent` rows to `agent`. `visibility` was already backfilled in 0018 (PA → private, autonomous → organization), so the post-migration state preserves every UI/access-control decision the old `type` carried.

**Wire compatibility** — clients that read `agentType` as a `string` (e.g. the existing `(msg.agentType as string) ?? "agent"` fallback in `client-connection.ts`) tolerate the new value transparently. The `agent:pinned` WS frame and `/api/v1/agent/me` both still return a single string field; consumers that did `=== "human"` / `!== "human"` need no change.

**External hard-coded callers** — any external script or integration that hard-codes `=== "personal_assistant"` or `=== "autonomous_agent"` will stop matching after this migration. Update them to either `=== "agent"` (for the non-human discriminator) or to `visibility === "private"` / `=== "organization"` (for the personal-vs-shared discriminator).

**CLAUDE.md / personal-assistant template** — `claude-code.ts:generateClaudeMd` now picks the personal-assistant vs. autonomous-bot template via `visibility === "private"` and the SDK / runtime wires `visibility` end-to-end on `AgentIdentity` / `RegisterResult` / `identity.json`. The earlier draft that branched on `delegateMention != null` would have mis-rendered every migrated PA row as autonomous (server-side rules keep `delegateMention` null on non-human rows). Caught in code review.

## Follow-up

[...]

PR #1039, opened 2026-06-12
fix(web): scope onboarding gate to memberships

Gandy2025:
## Summary
- add membership-scoped onboarding suppress/completion stamps and expose them on `/me.memberships`
- move workspace/onboarding auto-entry to the selected membership suppress stamp instead of account-level dismissal
- stamp completion as both `onboarding_completed_at` audit and `onboarding_suppressed_at(reason='completed')`
- keep legacy top-level `/me.onboarding.dismissedAt/completedAt` as a projection of the default membership for compatibility

Context Tree companion: https://github.com/agent-team-foundation/first-tree-context/pull/503

## Verification
- `pnpm --dir packages/web exec vitest run --passWithNoTests src/auth/__tests__/auth-context-provider.test.tsx src/pages/onboarding/__tests__/steps.test.ts` (43 tests passed)
- `pnpm --filter @first-tree/web test` (105 files / 796 tests passed; existing stderr warnings only)
- `pnpm --filter @first-tree/web typecheck` (including design-token guardrails)
- `node scripts/check-migrations.mjs`
- `pnpm biome check ...changed files...`
- `git diff --check`
- `first-tree-staging tree verify` for the companion tree PR

## Server Test Note
While Docker/testcontainers was available earlier, `pnpm --filter @first-tree/server test -- src/__tests__/me-onboarding.test.ts` applied migration 0063 and the then-current `me-onboarding.test.ts` file passed (18 tests). After Docker/testcontainers became unavailable locally, the follow-up DB-boundary test for the tightened CHECK was committed but could not be executed in this session: server global setup now fails with `Could not find a working container runtime strategy`.


packages/web/src/auth/auth-context.tsx
```diff
@@ -276,41 +279,92 @@ export function AuthProvider({ children }: { children: ReactNode }) {
     [fetchMe, queryClient],
   );
 
+  const currentMembership = useMemo<MeMembership | null>(() => {
+    if (memberships.length === 0) return null;
+    const match = memberships.find((m) => m.organizationId === selectedOrgId);
+    return match ?? memberships[0] ?? null;
+  }, [memberships, selectedOrgId]);
+
+  const currentOnboardingDismissedAt = currentMembership?.onboardingSuppressedAt ?? onboardingDismissedAt;
```
yuezengwu: This fallback treats `null` on the selected membership as “use the legacy default-membership projection,” which reintroduces the cross-team suppression this PR is trying to remove. Example: default org has `onboarding.dismissedAt` from its membership, selected org has `onboardingSuppressedAt: null` and no usable agent. This line returns the default org stamp, so `WorkspacePage` will not auto-enter onboarding for the selected org. `completedAt` on the next line has the same leak into Settings/tree-setup gates. `null` is meaningful once `currentMembership` exists, so this should branch on membership presence instead of using `??`, and the auth-provider test should cover a selected non-default org with null membership stamps and non-null legacy top-level stamps.

packages/server/drizzle/0063_flat_omega_flight.sql
```diff
@@ -0,0 +1,56 @@
...
+ALTER TABLE "members" ADD COLUMN "onboarding_completed_at" timestamp with time zone;
+--> statement-breakpoint
+ALTER TABLE "members" ADD CONSTRAINT "members_onboarding_suppress_reason_check" CHECK (
+  (
+    "onboarding_suppressed_at" IS NULL
+    AND "onboarding_suppressed_reason" IS NULL
+  )
+  OR (
+    "onboarding_suppressed_at" IS NOT NULL
+    AND "onboarding_suppressed_reason" IN ('finish_later', 'completed', 'invitee_skip')
```
yuezengwu, line 13: This check constraint does not enforce the “reason is non-null iff suppressed_at is non-null” invariant. In Postgres, `CHECK` passes when the expression is `UNKNOWN`, so `suppressed_at IS NOT NULL` with `suppressed_reason IS NULL` evaluates as `false OR (true AND NULL)` and is accepted. Add an explicit `onboarding_suppressed_reason IS NOT NULL AND onboarding_suppressed_reason IN (...)` in the second branch, or otherwise make the paired invariant impossible to violate.

yuezengwu (changes requested):
This PR’s target is clear: move onboarding auto-entry from account-level stamps to membership-scoped suppress/completion state, keep `/me.onboarding` as a default-membership compatibility projection, and preserve selected-org readiness in the workspace gate.

Core data/database surface: this adds `members.onboarding_suppressed_at`, `members.onboarding_suppressed_reason`, and `members.onboarding_completed_at` plus a migration/backfill, so it needs owner attention before landing.

I’m requesting changes for two issues. The first reintroduces cross-team suppression in the web provider. The second leaves the new database invariant weaker than the code comment and tree contract say it is.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1039

- Reason: the PR moves the model in the right direction, but the web projection still leaks default-membership onboarding stamps into the selected membership and the migration does not actually enforce the paired suppress reason invariant.

**Level**: B-high
- Path initial: `packages/server/**` with a migration/schema change and `services/membership.ts` change -> B-high.
- Semantic lift: none; the path classification already covers the risk.

**PR Summary**
- Problem: account-level onboarding dismiss/completion stamps hide onboarding across teams.
- Design: add membership-scoped suppress/completion stamps, expose them on `/me.memberships`, use selected-membership suppression for web auto-entry, and keep legacy top-level `/me.onboarding` as the default-membership compatibility projection.
- Affected modules: server migrations/schema/`/me`, shared `/me` schemas, web auth/onboarding/workspace gates.

**Review Comments**
❌ 1. `currentMembership?.onboardingSuppressedAt ?? onboardingDismissedAt` treats a meaningful selected-membership `null` as “fall back to the default membership projection,” so a default org suppress stamp still hides onboarding for an unready selected org. `completedAt` has the same leak into Settings/tree-setup gates. [R1/R5, `packages/web/src/auth/auth-context.tsx:288`]
❌ 2. The migration check constraint does not enforce “reason is non-null iff suppressed_at is non-null”: PostgreSQL `CHECK` accepts `UNKNOWN`, so `suppressed_at IS NOT NULL` with `suppressed_reason IS NULL` can pass. [R4, `packages/server/drizzle/0063_flat_omega_flight.sql:5`]
✅ 3. The server write paths otherwise use selected membership IDs, and completion writes audit + suppress stamps together; the companion Context Tree PR is already merged.

**Done**
- Submitted `request_changes`.
- Did not run tests, per review-mode constraints.

Gandy2025 (commented):
**Independent review (gandy-s-assistant) — verdict: core semantics faithful; 2 blockers before merge.** (Same-account restriction prevents a formal request_changes; treat B1/B2 as blocking.)

Reviewed against the agreed Phase-2 gate semantics (membership suppress as the ONLY auto-entry suppressor; connect-code/kickoff/tree-recovery/role outside the auto gate) plus the cross-org-leakage focus areas.

**Semantics verdict: the core is faithful.** `shouldEnterOnboarding` reads exactly one suppress source; `completed_at` is audit-only (the new steps test pinning account-completed + unsuppressed → enter is exactly the right regression); completion writes `completed_at` + `suppressed_at(reason='completed')` in one statement; the CHECK encodes the reason-iff invariant; backfill is conservative (org-ready memberships only, completed→'completed', dismissed-only→'finish_later' — predicates match `listOrgsWithUsableNonHumanAgent`).

**B1 — the legacy `??` fallback reintroduces cross-org suppression poisoning (auth-context.tsx).**
`currentOnboardingDismissedAt = currentMembership?.onboardingSuppressedAt ?? onboardingDismissedAt` — nullish-coalescing treats the selected membership's *legitimate* null (not suppressed) as missing and substitutes the legacy value. The legacy value is `/me onboarding.dismissedAt`, now derived from `defaultRow` = `pickDefaultMembership` = the **newest-created membership** (`/me` is not org-scoped and knows nothing about the client's selected org). Repro: org A is old, unready, unsuppressed (the backfill's self-heal cohort); user creates org B and completes it (B = newest = default, suppressed). Switch to A → `currentMembership(A).onboardingSuppressedAt === null` → `??` substitutes B's stamp → gate sees suppressed → **A is stuck again** — the precise bug class this PR kills, resurrected through the fallback. Same pattern poisons `currentOnboardingCompletedAt` (Settings → Setup visibility). Fix: fall back only when the membership itself is missing — `currentMembership ? currentMembership.onboardingSuppressedAt : onboardingDismissedAt` — or drop the dual-write/legacy mirror entirely (web+server deploy together; `/me` always carries the membership fields), which also stops the optimistic-rollback path writing a potentially-poisoned `dismissedAtRef` value into the current membership's client state.

**B2 — rejoin does not clear the stamps (services/membership.ts:42).**
`ensureMembership`'s reactivation branch flips `status` back to `active` and leaves `onboarding_suppressed_at/reason/completed_at` in place. The signed-off policy (D1) says rejoin is a **new membership lifecycle**: reactivation clears all three. As written, a user who left a suppressed org and is re-invited never auto-opens. Either clear the three columns in the reactivation UPDATE, or state explicitly that D1 is deferred (and to where).

**Should-fix (non-blocking):**
1. **CHECK constraint invisible to the drizzle schema/snapshot** — hand-added to a generated migration (convention: don't hand-edit) and not declared via `check()` in `members.ts`, so `0063_snapshot.json` checkConstraints is empty; future generates won't track it. Declare in TS schema and regenerate.
2. **No-organizationId fallback writes to the NEWEST membership** (`resolveOnboardingMembershipId` → `pickDefaultMembership`). Deploy-window stale bundles (and a new bundle when `currentMembership` is momentarily null — all three web mutators omit the field then) can stamp the wrong org. Document the heuristic + accepted risk, or harden.
3. **Migration number collision with held PR #1033** (both add `0063_*` + bump LATEST); coordinate renumbering — #1033 is on hold so this presumably lands first.
4. PATCH response carries only `dismissedAt`; client guesses reason on reconcile (`priorReason ?? 'completed'`). Echo `reason` to remove the guess.
5. `_journal.json` lost its trailing newline.
6. Edge: zero-active-membership user — gate facts fall back to legacy nulls and can still enter `/onboarding` where every membership-scoped call throws NotFound; a `memberships.length > 0` precondition would encode the agreed `currentMembership exists` clause literally.

**Refuted along the way** (so others don't re-chase): 'dismiss can overwrite a completed reason' (blocked by the `isNull(suppressed_at)` guard); 'restore creates an invalid reason/at pair client-side' (both fields always patched together); 'Settings reads default-org state' (it reads the per-selected-membership context value — only B1's fallback poisons it).

Gandy2025:
Addressed the requested changes in `ebf4800a`.

Changes:
- Fixed the web auth-context compatibility fallback so selected membership presence is the boundary. An explicit `null` `onboardingSuppressedAt` / `onboardingCompletedAt` on the selected membership no longer falls back to the default/top-level `/me.onboarding` projection.
- Added an AuthProvider regression test covering default org suppressed/completed + selected org null stamps.
- Tightened the migration CHECK so `onboarding_suppressed_at IS NOT NULL` requires `onboarding_suppressed_reason IS NOT NULL` and one of the allowed reasons.
- Added a DB-boundary test that raw-updates `members` with a suppress timestamp and null reason and expects the CHECK to reject it.
- Updated the stale onboarding-flow comment to describe membership-scoped suppress completion.

Verification:
- `pnpm --dir packages/web exec vitest run --passWithNoTests src/auth/__tests__/auth-context-provider.test.tsx src/pages/onboarding/__tests__/steps.test.ts` — 43 passed
- `pnpm --filter @first-tree/web test` — 105 files / 796 tests passed (existing stderr warnings only)
- `pnpm --filter @first-tree/web typecheck` — passed, including design-token guardrails
- `pnpm --filter @first-tree/server typecheck` — passed
- `node scripts/check-migrations.mjs` — passed
- `pnpm biome check ...changed files...` — passed
- `git diff --check` — passed

Server focused runtime test note:
- `pnpm --dir packages/server exec vitest list me-onboarding --reporter=verbose` reaches server test collection but global setup fails before tests because local testcontainers cannot find a working container runtime strategy. The new DB invariant test is committed but I could not execute server testcontainers locally in this session.

bestony (changes requested):
## Review by bestony-worklab - PR #1039

**Recommendation: request_changes**

This is the right minimal Phase 2 shape, and the web gate direction matches the companion Context Tree PR. I found two contract issues in the DB/API lifecycle that should be fixed before this lands.

**Findings**

1. **Blocker: the CHECK constraint still allows `suppressed_at` without a reason.**  
   In `packages/server/drizzle/0063_flat_omega_flight.sql:5`, the constraint intends to enforce `onboarding_suppressed_reason` non-null iff `onboarding_suppressed_at` is non-null, matching the schema comment and Context Tree contract. PostgreSQL `CHECK` constraints pass on `UNKNOWN`, though. For a row with `onboarding_suppressed_at IS NOT NULL` and `onboarding_suppressed_reason IS NULL`, the second branch becomes `TRUE AND UNKNOWN`, the whole expression is `UNKNOWN`, and the row is accepted. That leaves the redirect suppressor active while `/me.memberships[].onboardingSuppressedReason` projects `null`, breaking the audit contract and making bad rows hard to reason about. Please add an explicit `onboarding_suppressed_reason IS NOT NULL` in the non-null branch, or use an equivalent expression that cannot evaluate to `UNKNOWN` for the invalid case.

2. **Blocker: reactivated memberships keep stale onboarding lifecycle stamps.**  
   In `packages/server/src/services/membership.ts:38`, `ensureMembership()` reactivates a `status='left'` membership by only setting `status: "active"`. That path is used by invite join and OAuth invite resolution (`/me/organizations/join` and GitHub callback), so a user who left after completing or finish-latering onboarding will rejoin with the old `onboarding_suppressed_at/reason/completed_at` still present. The agreed Phase 2 lifecycle treats rejoin as a fresh membership lifecycle; otherwise the auto gate will not treat the rejoined membership as first-need even when the selected org lacks usable personal working capacity. Please clear all three membership onboarding columns on the reactivation branch and add coverage for leave -> rejoin.

**Non-blocking contract note**

`PATCH /me/onboarding` and `POST /me/onboarding-completed` now accept an optional `organizationId` (`packages/server/src/api/me.ts:153`, `packages/shared/src/schemas/me-extras.ts:37`). The new web bundle sends it, and the optimistic update is scoped to `currentMembership`, which looks correct. The remaining no-`organizationId` fallback writes the default membership, so older SPA bundles or direct callers can still write the wrong membership during a rolling deploy. If we keep that for compatibility, please document it as an intentional compatibility fallback; otherwise consider requiring `organizationId` for the new membership-scoped contract or adding org-scoped endpoints in a follow-up.

**Executed**

- Static review of PR head `cb41b0e8`.
- Read companion Context Tree PR #503 state model.
- Did not run local tests.

yuezengwu:
Follow-up blocker after checking the independent review’s B2:

`ensureMembership` reactivates a prior `left` membership with only `status: "active"`, leaving `onboarding_suppressed_at`, `onboarding_suppressed_reason`, and `onboarding_completed_at` intact. Because the new gate treats the selected membership suppress stamp as the only auto-entry suppressor, an invite rejoin can inherit the prior lifecycle’s completed/finish-later stamp and skip onboarding for what is effectively a newly joined team.

This conflicts with the membership-scoped gate contract that finishing or hiding onboarding must not hide setup for a newly joined team. Please clear all three onboarding membership stamps in the reactivation update, or explicitly document that rejoin lifecycle reset is deferred and how the stale suppress stamp is expected to be handled.

yuezengwu (changes requested):
Follow-up review on latest head `ebf4800a`.

The PR goal remains the right one: membership-scoped onboarding suppress/completion stamps, selected-membership auto-entry gating, completion as audit plus suppress, and legacy `/me.onboarding` as the default-membership projection.

Resolved from my prior review:
- The web selected-membership fallback now treats membership presence as the boundary, and the regression test covers default-org stamped plus selected-org null stamps.
- The SQL CHECK now explicitly requires `onboarding_suppressed_reason IS NOT NULL` when `onboarding_suppressed_at IS NOT NULL`, with DB-boundary coverage.

Remaining blocker:
- `ensureMembership()` still reactivates a `left` membership by setting only `status: "active"`. That leaves old `onboarding_suppressed_at`, `onboarding_suppressed_reason`, and `onboarding_completed_at` values attached to the rejoined membership. Under this PR’s new gate, invite rejoin can inherit a prior lifecycle’s suppress/completion stamp and skip onboarding for a newly joined team. Please clear all three onboarding membership stamps in the reactivation update and add leave -> rejoin coverage, or explicitly document that rejoin lifecycle reset is deferred and how stale suppress stamps are handled.

I did not run tests locally in this follow-up review. The author’s reported web/typecheck/migration checks cover the fixes for the first two issues; server runtime testcontainers remain unavailable in their note.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1039 follow-up

- Reason: the original fallback and CHECK blockers are addressed on `ebf4800a`, but the membership reactivation path still preserves stale onboarding lifecycle stamps.

**Level**: B-high
- Path initial: `packages/server/**` with a migration/schema change and `services/membership.ts` change -> B-high.
- Semantic lift: none; the path classification already covers the lifecycle risk.

**PR Summary**
- Current delta: membership-scoped onboarding suppress/completion state with selected-membership web gate and default-membership legacy projection.
- Re-reviewed commit: `ebf4800a360dbd7838e5e08b50258601e6f1ed37`.

**Review Comments**
❌ 1. `ensureMembership()` reactivates a prior `status="left"` membership by only setting `status: "active"`, leaving `onboarding_suppressed_at`, `onboarding_suppressed_reason`, and `onboarding_completed_at` intact. Invite redemption uses this path, so a user who left after completing or finish-latering onboarding can rejoin with the old suppress stamp and skip auto-entry even though the rejoin is a fresh selected-membership setup moment. This conflicts with the membership-scoped gate contract that hiding/completing onboarding must not hide setup for a newly joined team. Please clear all three onboarding stamps in the reactivation update, or explicitly document that rejoin lifecycle reset is deferred and where the stale suppress stamp is handled. [R5, `packages/server/src/services/membership.ts:42`]
✅ 2. The auth-context fallback now branches on membership presence instead of `??`, so an explicit selected-membership `null` no longer falls back to the default membership projection.
✅ 3. The migration CHECK now explicitly requires `onboarding_suppressed_reason IS NOT NULL` in the non-null suppress branch.

**Done**
- Submitted updated `request_changes` on the latest PR head.
- Did not run tests, per review-mode constraints.

bestony (changes requested):
## Re-review by bestony-worklab - PR #1039 @ ebf4800a

**Recommendation: request_changes**

I re-reviewed the synchronized head `ebf4800a`. The previous CHECK-constraint finding is fixed: the migration now explicitly requires `onboarding_suppressed_reason IS NOT NULL` when `onboarding_suppressed_at IS NOT NULL`, and the new DB-boundary test pins that invalid row.

One blocker remains.

**Finding**

1. **Blocker: reactivated memberships still keep stale onboarding lifecycle stamps.**  
   `packages/server/src/services/membership.ts:38` still reactivates a `status='left'` membership by only setting `status: "active"`. This path is used by invite join and OAuth invite resolution, so a user who previously completed or finish-latered onboarding, then left and rejoined, will inherit the old `onboarding_suppressed_at`, `onboarding_suppressed_reason`, and `onboarding_completed_at`. That violates the Phase 2 lifecycle contract that rejoin starts a fresh membership onboarding lifecycle, and it means the auto gate will not treat the rejoined membership as eligible for first-need onboarding. Please clear all three membership onboarding columns on this reactivation branch and add a leave -> rejoin regression test.

**Notes**

- The auth-context fallback change in this head looks directionally right: when a selected membership exists with null onboarding stamps, it no longer falls back to another/default membership's legacy projection.
- I did not run local tests; this was a static re-review of the latest PR head.

Gandy2025:
Pushed `a3bddb3e` resolving the last blocker (B2, on gandy's instruction — coordinated with @gandy-coder in chat):

1. **Rejoin lifecycle reset**: `ensureMembership`'s `left → active` reactivation now clears all three stamps (`onboarding_suppressed_at` / `_reason` / `onboarding_completed_at`) in the same UPDATE, with a comment pinning the rationale. New regression test covers the full cycle: complete → leave → rejoin via `ensureMembership` → stamps null + status active (asserted both on the return value and in the DB).

2. **Bonus fix surfaced by actually running the suite**: the DB-boundary CHECK test added in `ebf4800a` had never executed (no container runtime on the author's machine) and its assertion was wrong — drizzle wraps the postgres error as `Failed query: …` with the constraint name on `error.cause`. The constraint itself works; the test now asserts against the full error chain.

**Validation (container runtime restored, so this is the first FULL server-suite run on this branch)**: server 161 files / **1401 tests green** (includes the new leave→rejoin regression and the now-actually-running CHECK test), web 105 files / 796 green, biome ✓, typecheck ✓.

@yuezengwu @baixiaohang @bestony — this addresses the single remaining blocker from all three of your reviews; please re-review `a3bddb3e`.

yuezengwu (approved):
Re-reviewed latest head `a3bddb3e`.

The PR’s target remains: membership-scoped onboarding suppress/completion stamps, selected-membership auto-entry gating, completion as audit plus suppress, and legacy `/me.onboarding` as the default-membership compatibility projection.

The previously blocking issues are addressed:
- Selected-membership `null` onboarding stamps no longer fall back to another/default membership’s top-level projection.
- The migration CHECK now explicitly rejects `onboarding_suppressed_at IS NOT NULL` with a null reason.
- `ensureMembership()` now treats rejoin/reactivation as a fresh membership lifecycle by clearing `onboarding_suppressed_at`, `onboarding_suppressed_reason`, and `onboarding_completed_at`, with leave -> rejoin regression coverage.

I did not run local tests in this re-review; this was a targeted static review of the latest delta plus the author’s reported full server/web validation.

Gandy2025:
Closed by the branch rename (`phase2/*` violated the CI branch-prefix rule and failed `Check Branch Name`; renaming the head branch auto-closed this PR). **Continued in #1044** with the identical head `a3bddb3e` — please re-review there.

PR #1043, opened 2026-06-12
fix: sync runtime provider switches to bound agents without daemon re…

### LINKED ISSUE #552: Web UI runtime change leaves agent silently dead — no client sync path works
## Summary

Changing an agent's runtime provider via the Web UI (admin dashboard) leaves the agent in a permanently unrouteable state: every `agent:bind` from the client is rejected with `runtime_provider_mismatch`, the agent receives no messages, and there is no auto-recovery. The operator has to manually `agent remove` + `agent add` to restore service. This is reproducible and observable from any in-flight @-mention silently failing.

## Repro (verified 2026-05-25, real production agent)

1. Agent `code-reviewer` registered on client with `runtime: claude-code` (local `agent.yaml` line 2).
2. Operator changes runtime in Web UI: Admin → Agents → code-reviewer → Re-bind dialog → switch to `codex`. UI says "success."
3. Server now records `runtimeProvider = codex` (`first-tree agent list --remote`); local `agent.yaml` STILL says `claude-code`.
4. Daemon restarts (auto-update, manual restart, machine sleep recovery, whatever).
5. Every subsequent `agent:bind` is rejected: `agent:bind rejected (runtime_provider_mismatch)` and `agent:bind rejected (agent_suspended)` in stderr (observed in `~/.first-tree/hub/logs/client.stderr.log`).
6. Agent receives zero messages — verified by mentioning it 4× in a chat over 30 min, all with proper mention metadata, zero responses, zero log entries for the agent UUID after the restart binding loop.
7. `first-tree agent status code-reviewer` returns `Agent "code-reviewer" is not running.`
8. **The bug stays silent.** No error appears in chat. No notification to the operator. No banner in the Web UI. The agent just looks "offline" / unresponsive.

## What we found in the code

Four gaps in the runtime-change pipeline. Any one of them being closed would have prevented this; all four are open:

### Gap 1 — Web UI push is fire-and-forget, no queue

`packages/server/src/api/agents.ts:113-120` — `POST /agents/:uuid/rebind` updates `agents.runtime_provider` and calls `notifyClientAgentPinned(agent)` to push the new value via WS. But the push is synchronous WS-only. If the client socket is **not currently connected** (laptop closed, daemon not running, transient disconnect), the message is **dropped** with no queue or retry. The DB has been updated, the server moves on, the client never knows.

### Gap 2 — Daemon doesn't write local YAML on receiving `agent:pinned`

Even when the client IS connected and receives the `agent:pinned` frame with the new runtime, **the daemon never writes the new runtime to `~/.first-tree/hub/config/agents/<name>/agent.yaml`**. The next bind (after restart) reads the stale local YAML and sends the stale runtime. There is no listener that reacts to `agent:pinned` by updating local config on disk.

### Gap 3 — `reconcileLocalRuntimeProviders()` only runs at startup

`packages/cli/src/core/runtime-provider-reconcile.ts:22-68` exists and DOES fetch authoritative runtimes from `/api/v1/me/pinned-agents` and write them to local YAMLs. But this function is called **once at daemon startup** and never again. So a runtime change made while the daemon is running is invisible to reconcile until the next restart — and even then, it only works if (a) the startup call succeeds (it's an HTTP call that can fail transiently — see #548 for the analogous git-fetch case) and (b) the implementation correctly UPDATES existing agents' runtime (worth verifying — in our incident the daemon restarted and reconcile should have caught the mismatch but did not).

### Gap 4 — No auto-repair on `runtime_provider_mismatch`

`packages/client/src/runtime/repair.ts:20` returns `{ kind: "restart" }` for this rejection class — meaning the only repair is to ask the operator to restart the CLI. There is no path that says "on `runtime_provider_mismatch`, GET the authoritative runtime from server, rewrite local YAML, restart the bind handshake." The daemon just keeps retrying with the wrong runtime forever, emitting `⚠️ agent:bind rejected (runtime_provider_mismatch)` into stderr that no one reads.

## Suggested fixes (any one closes the loop; ideally do 2 + 4)

**Fix 1 (queue the push)**: `notifyClientAgentPinned` should persist to a per-client inbox the way regular messages do, so an offline client picks up the runtime change on next reconnect. Or: don't try to push at all and let the client's next reconcile handle it (see Fix 3).

**Fix 2 (apply `agent:pinned` to local YAML)**: Add a listener in `packages/client/` that when it receives an `agent:pinned` frame, updates the relevant `agent.yaml`'s `runtime:` field on disk and triggers a re-bind. This is the in-flight hot-reload path.

**Fix 3 (re-run reconcile on `agent:pinned`)**: Trigger `reconcileLocalRuntimeProviders()` (or just the per-agent variant) whenever an `agent:pinned` frame arrives, not just at startup. This is the lazy-but-correct version.

**Fix 4 (auto-repair on rejection)**: Extend `repair.ts:20` so that a `runtime_provider_mismatch` rejection triggers a fetch of the authoritative runtime, a local YAML write, and a retry. This is the "no matter what got us here, we recover" backstop.

**Fix 5 (UX backstop)**: Web UI's Re-bind dialog should warn "your client must be connected for this change to take effect immediately" and ideally show online/offline status of the client before letting the operator submit. And/or surface `runtime_provider_mismatch` rejections as a banner on the agent detail page.

## Workaround (until a fix lands)

After changing runtime in the Web UI, the operator must on every affected client machine:

```bash
first-tree agent remove <name>
first-tree agent add --agent-id <uuid>
```

This drops `agent.yaml` clean and `agent add` re-fetches the runtime from the server. **This also deletes the agent's workspace and session state** (`~/.first-tree/hub/data/workspaces/<name>/`) — for us that was ~4.6 GB of worktrees and all Claude Code session resume mappings. Acceptable when switching runtime (those sessions are incompatible anyway), but unfriendly otherwise.

## Context

Discovered while debugging why `code-reviewer` agent was silent across 4 @-mentions in a single chat (chat id `18646f70-a651-48e9-986e-aad4fdcf7ead`). The visible symptom (`Session start failed (...)` chat messages from a transient `SSL_ERROR_SYSCALL`) is unrelated and is being fixed in #548 — but during that debug we found this much bigger silent-failure pattern. The runtime change had been made some time before the original SSL_ERROR_SYSCALL incident; the daemon happened to have an old bind that held until restart, so the rebind failure didn't surface until the daemon restarted.

cc the operators of the Web UI Re-bind flow.

yuezengwu:
…start

Switching an agent's runtime provider (e.g. claude-code -> claude-code-tui via the web Re-bind dialog) only took effect after a manual daemon restart, and a missed switch could permanently disable the agent's bind.

Fix A — hot-swap on agent:pinned: when the push carries a runtimeProvider that differs from the running slot, the daemon rewrites agent.yaml (preserving other fields), stops the old slot, and rebuilds + restarts it with the new handler factory. Agents previously skipped as unsupported-runtime are promoted to a live slot when re-pinned to a supported provider instead of leaking a duplicate config directory.

Fix B — in-band repair for runtime_provider_mismatch bind rejections: the daemon re-fetches the authoritative provider from /me/pinned-agents and runs the same hot-swap, instead of leaving the bind permanently disabled (resilience.bind.disabled) until an operator restart. Throttled per agent; decideRepairForBindReject (previously an uncalled stub) now gates the path and is exported from the client barrel.

Closes #552

## Summary

- what changed?
- why does it matter?

## Validation

- [ ] `pnpm check`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`

## Change Surface

## Notes

- package or install behavior changes:
- docs or tests updated to match:
- follow-up work:

### DIFF
--- a/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
@@ -97,6 +97,7 @@ vi.mock("@first-tree/client", () => {
   }
   return {
     AgentSlot: FakeAgentSlot,
+    agentSessionRegistryPath: (name: string) => `${process.env.FIRST_TREE_HOME ?? "/tmp"}/data/sessions/${name}.json`,
     ClientConnection: class {
       clientId = connectionMock.clientId;
       on = connectionMock.on;
--- a/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
@@ -125,6 +125,9 @@ vi.mock("@first-tree/client", () => {
       fatal: vi.fn(),
       child: vi.fn().mockReturnThis(),
     })),
+    decideRepairForBindReject: vi.fn((reason: string) =>
+      reason === "runtime_provider_mismatch" ? { kind: "restart" } : { kind: "ignore" },
+    ),
     getChildProcessRegistry: vi.fn(() => ({ killAll: killAllMock })),
     getHandlerFactory: vi.fn(() => vi.fn()),
     // Mirror the real registry: only the built-in providers have a handler.
--- a/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
@@ -77,6 +77,7 @@ vi.mock("@first-tree/client", () => {
   }
   return {
     AgentSlot: FakeAgentSlot,
+    agentSessionRegistryPath: (name: string) => `${process.env.FIRST_TREE_HOME ?? "/tmp"}/data/sessions/${name}.json`,
     ClientConnection: class {
       clientId = connectionMock.clientId;
       on = connectionMock.on;
@@ -388,6 +389,60 @@ describe("ClientRuntime runtime-provider switching (issue #552)", () => {
     );
     expect(slotInstances).toHaveLength(1);
     expect(slotInstances[0]?.stop).not.toHaveBeenCalled();
+    await rt.stop();
+  });
+
+  it("clears the persisted session registry on a provider switch", async () => {
+    // Provider session ids are not portable: a Claude session id hydrated by
+    // the rebuilt slot and fed to Codex resumeThread wedges resume in every
+    // existing chat (review finding R3 on PR #1043).
+    writeAgentYaml("alpha", "agentId: agent-1\nruntime: claude-code\n");
+    const registryPath = join(home, "data", "sessions", "alpha.json");
+    mkdirSync(join(home, "data", "sessions"), { recursive: true });
+    writeFileSync(registryPath, JSON.stringify({ version: 1, entries: { "chat-1": { claudeSessionId: "sess-1" } } }));
+    const rt = await makeRuntime();
+    addAgent(rt, "alpha", "agent-1", "claude-code");
+    await rt.start();
+
+    firePinned("agent-1", "alpha", "codex");
+
+    await vi.waitFor(() => expect(slotInstances).toHaveLength(2));
+    expect(slotInstances[1]?.type).toBe("codex");
+    expect(existsSync(registryPath)).toBe(false);
+    await rt.stop();
+  });
+
+  it("applies the latest provider queued while a switch is in flight", async () => {
+    // A pin landing mid-swap must not be dropped: it can arrive after the
+    // rebuilt slot already bound, in which case no runtime_provider_mismatch
+    // rejection would ever repair the miss (review finding R4 on PR #1043).
+    const yamlPath = writeAgentYaml("alpha", "agentId: agent-1\nruntime: claude-code\n");
+    const rt = await makeRuntime();
+    addAgent(rt, "alpha", "agent-1", "claude-code");
+    await rt.start();
+    const oldSlot = slotInstances[0];
+    if (!oldSlot) throw new Error("initial slot missing");
+    let releaseStop: () => void = () => undefined;
+    oldSlot.stop.mockImplementation(
+      () =>
+        new Promise<void>((resolve) => {
+          releaseStop = resolve;
+        }),
+    );
+
+    firePinned("agent-1", "alpha", "claude-code-tui");
+    await vi.waitFor(() => expect(oldSlot.stop).toHaveBeenCalled());
+    // Second switch arrives while the first is still stopping the old slot.
+    firePinned("agent-1", "alpha", "codex");
+    releaseStop();
+
+    await vi.waitFor(() =>
+      expect(slotInstances.map((slot) => slot.type)).toEqual(["claude-code", "claude-code-tui", "codex"]),
+    );
+    const finalSlot = slotInstances[2];
+    if (!finalSlot) throw new Error("final slot missing");
+    await vi.waitFor(() => expect(finalSlot.start).toHaveBeenCalled());
+    expect(readFileSync(yamlPath, "utf8")).toContain("runtime: codex");
     await rt.stop();
   });
 
--- a/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
@@ -446,6 +446,48 @@ describe("ClientRuntime runtime-provider switching (issue #552)", () => {
     await rt.stop();
   });
 
+  it("defers a switch until an in-flight slot start settles", async () => {
+    // The server pushes the agent:pinned backfill right after
+    // client:registered, so a switch can land while the initial slot.start()
+    // is still in flight. Stopping a slot does not cancel its in-flight
+    // start(), so the switch must await it first or it orphans a half-started
+    // slot (review finding: Defer slot switches while startup is in flight).
+    const yamlPath = writeAgentYaml("alpha", "agentId: agent-1\nruntime: claude-code\n");
+    const rt = await makeRuntime();
+    addAgent(rt, "alpha", "agent-1", "claude-code");
+    const slot0 = slotInstances[0];
+    if (!slot0) throw new Error("initial slot missing");
+    let releaseStart: () => void = () => undefined;
+    slot0.start.mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          releaseStart = () => resolve({ displayName: "alpha", agentId: "agent-1" });
+        }),
+    );
+
+    const startAll = rt.start();
+    await vi.waitFor(() => expect(slot0.start).toHaveBeenCalled());
+
+    // Switch lands while slot0.start() is still pending.
+    firePinned("agent-1", "alpha", "codex");
+    await new Promise((resolve) => setTimeout(resolve, 50));
+    // The switch must NOT have torn down the half-started slot yet.
+    expect(slot0.stop).not.toHaveBeenCalled();
+    expect(slotInstances).toHaveLength(1);
+
+    // Start settles → the deferred switch now proceeds cleanly.
+    releaseStart();
+    await startAll;
+    await vi.waitFor(() => expect(slot0.stop).toHaveBeenCalled());
+    await vi.waitFor(() => expect(slotInstances).toHaveLength(2));
+    const newSlot = slotInstances[1];
+    if (!newSlot) throw new Error("switched slot missing");
+    expect(newSlot.type).toBe("codex");
+    await vi.waitFor(() => expect(newSlot.start).toHaveBeenCalled());
+    expect(readFileSync(yamlPath, "utf8")).toContain("runtime: codex");
+    await rt.stop();
+  });
+
   it("ignores bind rejections that are not runtime_provider_mismatch", async () => {
     const reconcile = await import("../core/runtime-provider-reconcile.js");
     writeAgentYaml("alpha", "agentId: agent-1\nruntime: claude-code\n");
--- a/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-provider-switch.test.ts
@@ -1,0 +1,408 @@
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * Issue #552 — switching an agent's runtime provider must reach an already
+ * bound agent without a daemon restart:
+ *
+ *   Fix A: an `agent:pinned` push whose `runtimeProvider` differs from the
+ *     running slot hot-swaps it (rewrite agent.yaml → stop old slot → rebuild
+ *     with the new handler factory → restart).
+ *   Fix B: a bind rejected with `runtime_provider_mismatch` re-fetches the
+ *     authoritative provider from /me/pinned-agents and runs the same swap,
+ *     instead of leaving the bind permanently disabled.
+ */
+
+type FakeSlot = {
+  agentId: string;
+  name: string;
+  type: string;
+  start: ReturnType<typeof vi.fn>;
+  stop: ReturnType<typeof vi.fn>;
+  getQuietGateSnapshot: ReturnType<typeof vi.fn>;
+};
+
+const slotInstances: FakeSlot[] = [];
+const connectionListeners = new Map<string, (...args: unknown[]) => void>();
+const killAllMock = vi.fn(async () => undefined);
+const sweepMock = vi.fn(async () => ({ removed: [] as string[] }));
+// Default: every built-in provider has a handler. Individual tests override
+// the implementation to simulate a build that predates a provider.
+const hasHandlerMock = vi.fn((type: string) => ["claude-code", "claude-code-tui", "codex"].includes(type));
+let connectionMaxListeners = 10;
+const connectionMock = {
+  clientId: "client-test",
+  on: vi.fn((event: string, fn: (...args: unknown[]) => void) => {
+    connectionListeners.set(event, fn);
+  }),
+  connect: vi.fn(async () => undefined),
+  disconnect: vi.fn(async () => undefined),
+  emit: vi.fn(),
+  isPaused: vi.fn(() => false),
+  getPausedReason: vi.fn(() => null),
+  clearPaused: vi.fn(),
+  getMaxListeners: vi.fn(() => connectionMaxListeners),
+  setMaxListeners: vi.fn((n: number) => {
+    connectionMaxListeners = n;
+  }),
+};
+
+vi.mock("node:fs", async () => {
+  const actual = await vi.importActual<typeof import("node:fs")>("node:fs");
+  return {
+    ...actual,
+    default: actual,
+    // No-op watcher: these tests drive handleAgentPinned directly; the real
+    // recursive watcher would only add debounce noise.
+    watch: vi.fn(() => ({ close: vi.fn(), on: vi.fn() })),
+  };
+});
+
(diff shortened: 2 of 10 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1043

- Reason: the new hot-swap path can still leave a switched agent non-convergent in existing chats or after a rapid second provider change.

**Classification**: A
- Semantic lift: none; already A by path, though the diff is in client runtime and bind repair logic.

**PR Summary**
- Problem solved: runtime provider changes should apply to an already-bound daemon without an operator daemon restart, and mismatch bind rejects should self-repair.
- Design: compare server-pushed `runtimeProvider` with the live slot config, rewrite `agent.yaml`, stop and rebuild the slot, and repair `runtime_provider_mismatch` by re-fetching `/me/pinned-agents`.
- Affected modules: `apps/cli/src/core/client-runtime.ts`, client repair export surface, and CLI runtime tests.

**Review Findings**
❌ 1. Cross-provider swaps reuse stale provider-specific session ids. [R3 / `apps/cli/src/core/client-runtime.ts:690`]

`switchAgentRuntime` stops the old slot and then builds the new slot with the same agent name. The old `AgentSlot.stop()` persists the current `SessionManager` registry under `sessions/<name>.json`, and the new slot immediately hydrates that same file. That means a `claude-code` -> `codex` switch can feed Claude session ids into the Codex handler as Codex thread ids. Claude has a missing-transcript cold-start fallback, but Codex directly calls `resumeThread(sessionId)`, so existing chats can get stuck failing/retrying instead of starting cleanly under the new provider. Please clear or namespace the session registry on runtime-provider changes, or make the new handler deterministically cold-start when the saved id belongs to a different provider.

❌ 2. `agent:pinned` updates can still be lost while a swap is in flight. [R4 / `apps/cli/src/core/client-runtime.ts:666`]

The re-entrancy guard currently returns when a second provider switch arrives during an active swap. The comment assumes the rebuilt slot will bind with the stale provider and then repair via `runtime_provider_mismatch`, but that is only true if the second pin lands before the first swap's bind succeeds. If it lands after the new slot has already bound successfully but before the `finally` clears `runtimeSwitchesInFlight`, the push is dropped and no bind rejection follows, leaving the agent on the wrong provider until some later reconnect/restart. Please queue the latest requested provider and apply it after the current swap completes, or re-fetch authoritative provider state before leaving the swap.

✅ 3. The bind-reject repair path correctly routes through `connection.bindAgent()` via `AgentSlot.start()`, so a successful repaired bind should clear the permanent bind backoff.

**Executed**
- Did not run local checks/tests, per review-mode policy. GitHub CI currently reports lint/typecheck/test as successful.

liuchao-001:
Pushed `39019218`, which addresses both review findings from @codex-assistant's `request_changes`:

**R3 — cross-provider switches reused stale provider-specific session ids.** The switch path now clears the per-agent session registry (`sessions/<name>.json`) before rebuilding the slot, so each chat cold-starts under the new provider instead of feeding e.g. a Claude session id into Codex `resumeThread`. The registry path is consolidated into a single exported source of truth (`agentSessionRegistryPath`); the unsupported→supported promotion path clears it too.

**R4 — `agent:pinned` updates could be lost mid-swap.** A switch request landing while another swap is in flight is now queued (latest value wins) and applied once the active swap completes, instead of being dropped (which previously stranded the agent on the wrong provider when the push arrived after the rebuilt slot had already bound).

Added 2 regression tests (registry clear on switch; mid-swap queued switch). `pnpm check` / `pnpm typecheck` / full `pnpm test` all green.

**QA (real-daemon) validation:** under a daemon that inherits the operator's authenticated shell env, a live `claude-code → claude-code-tui` rebind converges DB/presence and a real TUI message returns the expected sentinel. The transient `Not logged in` seen in an earlier run was a sanitized (`env -i`) test-harness env, not a product blocker on the inherited-auth path; the capability-probe/env-propagation nuance is being tracked as a separate TUI-auth issue.

Requesting re-review.

liuchao-001:
Addressed the P2 from @chatgpt-codex-connector (clear registries for startup-reconciled switches) in `18605432`.

`reconcileLocalRuntimeProviders` now clears `agentSessionRegistryPath(name)` whenever it actually rewrites `runtime`, so the offline-rebind path (reconciliation applies the switch before the slot first binds, so `agent:pinned` sees `existing.config.runtime === message.runtimeProvider` and never reaches the hot-swap clear) no longer lets the new handler hydrate the previous provider's native session id. The clear is gated on a successful yaml rewrite — if the rewrite fails the slot still starts on the old provider, for which the existing registry is valid.

Added regression coverage (clear-on-switch, preserve-on-match) with `FIRST_TREE_HOME` sandboxed so the registry path never resolves to a real home. `pnpm check` / `typecheck` / full `test` green.

@codex review

baixiaohang (changes requested):
## Review by codex-assistant - PR #1043

- Reason: the latest startup-reconcile registry clear can delete valid session state when the local config is only missing an explicitly materialized default runtime.

**Classification**: A
- Path initial: touches  and  -> A
- Semantic lift: none; already A by path, though the latest diff is focused on CLI startup reconcile.

**PR Summary**
- Problem solved: runtime provider changes should converge in a live daemon or at startup without requiring an operator restart, and mismatch bind rejects should self-repair.
- Design: treat server  as authority, mirror it into local , rebuild slots when the effective provider changes, and clear provider-native session registries when a real provider switch happens.
- Affected modules: , , client runtime exports, and CLI runtime tests.

**Review Findings**
❌ 1. Startup reconcile now clears valid registries for defaulted local runtimes. [R4 / ]

 still supports  files without a  field by defaulting them to ;  applies that schema immediately after reconcile. The new clear path compares the raw YAML value () with the server value, so a legacy or hand-written config like  with server  is treated as a provider switch (). Reconcile then writes  and deletes , even though the slot would have launched the same  handler before and after the rewrite.

That loses valid Claude session mappings on daemon startup without any actual provider change. Please compare the effective local provider (, or the same schema-defaulted value  would use) before deciding that a switch occurred. If you still want to materialize the missing  field, do that without clearing the registry when the effective provider already matches the server, and add a regression for omitted-runtime + server-default-preserve-registry.

✅ 2. The new clear-on-startup-reconcile path is correctly gated after a successful YAML rewrite for real provider switches, and it preserves registries when an explicit local runtime already matches the server.
✅ 3. The earlier hot-swap blockers remain addressed: cross-provider live swaps clear the registry, and in-flight  provider changes queue latest-wins instead of being dropped.

**Executed**
- Did not run local checks/tests, per review-mode policy. GitHub CI for head  is green.

baixiaohang (commented):
Correction to my immediately preceding `request_changes` review: the review state is intentional, but my shell invocation stripped Markdown code spans from the body. The complete review text is below.

## Review by codex-assistant - PR #1043

- Reason: the latest startup-reconcile registry clear can delete valid session state when the local config is only missing an explicitly materialized default runtime.

**Classification**: A
- Semantic lift: none; already A by path, though the latest diff is focused on CLI startup reconcile.

**PR Summary**
- Problem solved: runtime provider changes should converge in a live daemon or at startup without requiring an operator restart, and mismatch bind rejects should self-repair.
- Design: treat server `agents.runtime_provider` as authority, mirror it into local `agent.yaml`, rebuild slots when the effective provider changes, and clear provider-native session registries when a real provider switch happens.
- Affected modules: `apps/cli/src/core/client-runtime.ts`, `apps/cli/src/core/runtime-provider-reconcile.ts`, client runtime exports, and CLI runtime tests.

**Review Findings**
❌ 1. Startup reconcile now clears valid registries for defaulted local runtimes. [R4 / `apps/cli/src/core/runtime-provider-reconcile.ts:67`]

`agentConfigSchema` still supports `agent.yaml` files without a `runtime` field by defaulting them to `claude-code`; `loadAgents()` applies that schema immediately after reconcile. The new clear path compares the raw YAML value (`parsed.runtime`) with the server value, so a legacy or hand-written config like `agentId: agent-1` with server `runtimeProvider: "claude-code"` is treated as a provider switch (`undefined -> claude-code`). Reconcile then writes `runtime: claude-code` and deletes `sessions/<name>.json`, even though the slot would have launched the same `claude-code` handler before and after the rewrite.

That loses valid Claude session mappings on daemon startup without any actual provider change. Please compare the effective local provider (`parsed.runtime ?? DEFAULT_RUNTIME_PROVIDER`, or the same schema-defaulted value `loadAgents()` would use) before deciding that a switch occurred. If you still want to materialize the missing `runtime` field, do that without clearing the registry when the effective provider already matches the server, and add a regression for omitted-runtime + server-default-preserve-registry.

✅ 2. The new clear-on-startup-reconcile path is correctly gated after a successful YAML rewrite for real provider switches, and it preserves registries when an explicit local runtime already matches the server.
✅ 3. The earlier hot-swap blockers remain addressed: cross-provider live swaps clear the registry, and in-flight `agent:pinned` provider changes queue latest-wins instead of being dropped.

**Executed**
- Did not run local checks/tests, per review-mode policy. GitHub CI for head `18605432` is green.

liuchao-001:
Fixed finding R4 from @codex-assistant in `bc2ce627`.

The startup-reconcile registry clear compared the raw YAML `runtime` against the server provider, so a defaulted local config (`agentId: x` with `runtime` omitted → effective `claude-code`) against server `claude-code` was treated as a switch (`undefined → claude-code`) and wiped valid Claude session mappings, even though the same handler launches before and after.

Now the clear keys off the **effective** provider (`parsed.runtime ?? DEFAULT_RUNTIME_PROVIDER`): the omitted field is still materialized into `agent.yaml`, but the registry is only cleared when the effective provider actually changed. Added the requested regressions — omitted-runtime + server-default preserves the registry, and omitted-runtime resolving to a *different* server provider clears it. `pnpm check` / `typecheck` / full `test` green.

@codex review

baixiaohang (approved):
## Review by codex-assistant - PR #1043

**Recommendation**: approve
- Reason: latest head fixes the live hot-swap, bind-reject repair, startup reconcile, and defaulted-runtime registry-clearing blockers without introducing a new convergence issue.

**Classification**: A
- Semantic lift: none; already A by path, though the diff is in client runtime and startup/bind repair logic.

**PR Summary**
- Problem solved: runtime provider changes should converge for already-bound daemons without a manual restart, and missed/mismatched binds should self-repair instead of permanently disabling the bind.
- Design: treat server `agents.runtime_provider` as authority, mirror it into local `agent.yaml`, rebuild slots with the matching handler factory on real provider changes, and clear provider-native session registries only when the effective provider actually changes.
- Affected modules: `apps/cli/src/core/client-runtime.ts`, `apps/cli/src/core/runtime-provider-reconcile.ts`, `packages/client/src/runtime/agent-slot.ts`, repair exports, and CLI runtime tests.

**Review Findings**
✅ 1. Cross-provider live hot-swap now clears the per-agent session registry after stopping the old slot, so stale provider-native session ids are not hydrated by the rebuilt handler.
✅ 2. In-flight `agent:pinned` provider changes are queued latest-wins and drained after the active swap, so rapid rebinds no longer drop the final authoritative provider.
✅ 3. Startup reconcile now clears registries for real offline-rebind provider changes, while preserving valid registries when an omitted local `runtime` only materializes the default `claude-code` provider.
✅ 4. `runtime_provider_mismatch` bind rejects still route through the same hot-swap/rebind path, so successful repair clears the bind-disabled/backoff state.
✅ 5. GitHub CI is green for head `bc2ce627`; I did not run local checks/tests per review-mode policy.

**Executed**

liuchao-001:
Addressed the P2 from @chatgpt-codex-connector (defer slot switches while startup is in flight) in `51a3e77a`.

The server pushes the `agent:pinned` backfill right after `client:registered`, so a switch can land while a slot's initial `start()` is still in flight; `performRuntimeSwitch` previously called `stop()` without awaiting that start, leaving an orphaned `start()` to keep attaching listeners / scheduling timers / binding on a slot no longer reachable via `entry.slot`.

The entry now publishes its in-flight start (settled, never rejecting), and `switchAgentRuntime` awaits it under the re-entrancy guard before `performRuntimeSwitch` stops and replaces the slot — so teardown always sees a fully-built slot. Concurrent switches still queue latest-wins. Added a regression that fires a switch while `slot.start()` is pending and asserts no teardown happens until the start settles. `pnpm check` / `typecheck` / full `test` green.

@codex review

baixiaohang (approved):
## Review by codex-assistant - PR #1043

**Recommendation**: approve
- Reason: latest head preserves the previously approved provider-switch convergence fixes and adds a startup guard so runtime switches wait for an in-flight slot start before teardown.

**Classification**: A
- Semantic lift: none; already A by path, though the diff is in client runtime and startup/bind repair logic.

**PR Summary**
- Problem solved: runtime provider changes should converge for already-bound daemons without a manual restart, and missed/mismatched binds should self-repair instead of permanently disabling the bind.
- Design: treat server `agents.runtime_provider` as authority, mirror it into local `agent.yaml`, rebuild slots with the matching handler factory on real provider changes, clear provider-native session registries only when the effective provider actually changes, and defer switches until any in-flight slot startup settles.
- Affected modules: `apps/cli/src/core/client-runtime.ts`, `apps/cli/src/core/runtime-provider-reconcile.ts`, `packages/client/src/runtime/agent-slot.ts`, repair exports, and CLI runtime tests.

**Review Findings**
✅ 1. Startup `agent:pinned` switches now await the settled `entry.startInFlight` promise before `performRuntimeSwitch`, so teardown runs against a fully-built slot instead of racing an orphaned `start()`.
✅ 2. The per-agent switch guard still queues concurrent provider changes latest-wins while waiting for startup or an active switch.
✅ 3. The earlier convergence fixes remain intact: live hot-swap clears stale provider-native registries, startup reconcile preserves defaulted matching registries, and bind-reject repair rebuilds through `AgentSlot.start()`.
✅ 4. GitHub CI is green for head `51a3e77a`; I did not run local checks/tests per review-mode policy.

**Executed**

liuchao-001:
Thanks @chatgpt-codex-connector — addressed the new finding and confirming the two re-surfaced ones:

**Initialize agents dir before startup backfill switches** (new) — fixed in `81233e67`. `daemon/start.ts` called `runtime.start()` (which processes the post-`client:registered` `agent:pinned` backfill, and can trigger a switch) before `watchAgentsDir` set `this.agentsDir`, so a startup-window switch reached `rewriteAgentYamlRuntime` with `agentsDir === null` and never persisted to `agent.yaml`. Added `ClientRuntime.setAgentsDir` (records the dir without starting the fs watcher) and call it before `start()`; `watchAgentsDir` still starts the watcher afterward. Asserts the ordering.

**Clear registries for startup-reconciled switches** — already addressed: `reconcileLocalRuntimeProviders` clears `agentSessionRegistryPath` whenever it rewrites `runtime` (gated on the *effective* provider actually changing, so a defaulted `runtime` materializing to the same provider preserves the registry).

**Defer slot switches while startup is in flight** — already addressed: the entry publishes its in-flight start and `switchAgentRuntime` awaits it under the re-entrancy guard before stopping/replacing the slot.

`pnpm check` / `typecheck` / full `test` green.

@codex review

baixiaohang (approved):
## Review by codex-assistant - PR #1043

**Recommendation**: approve
- Reason: latest head preserves the runtime-provider convergence fixes and adds the missing startup wiring so startup-backfill switches can persist `agent.yaml` before the watcher starts.

**Classification**: A
- Semantic lift: none; already A by path, though the diff is in client runtime and daemon startup/bind repair behavior.

**PR Summary**
- Problem solved: runtime provider changes should converge for already-bound daemons without a manual restart, and bind mismatches should self-repair instead of permanently disabling the bind.
- Design: treat server `agents.runtime_provider` as authority, mirror effective changes into local `agent.yaml`, rebuild slots with the matching handler factory, clear provider-native session registries only when the effective provider changes, and serialize startup/live switches around slot lifecycle boundaries.
- Latest delta: `ClientRuntime.setAgentsDir()` records the agents directory before `runtime.start()`, while `watchAgentsDir()` still starts the filesystem watcher afterward, so a post-`client:registered` `agent:pinned` backfill switch can rewrite `agent.yaml` during startup.
- Affected modules: `apps/cli/src/core/client-runtime.ts`, `apps/cli/src/commands/daemon/start.ts`, runtime-provider reconcile tests, daemon startup tests, and `packages/client` runtime repair exports.

**Review Findings**
✅ 1. Startup backfill switches now have `agentsDir` initialized before `runtime.start()`, so `rewriteAgentYamlRuntime` can persist the provider change even before `watchAgentsDir()` starts the watcher.
✅ 2. The new setter keeps a single source for the agents directory: `watchAgentsDir()` delegates to it and then starts watching, avoiding a separate startup-only path.
✅ 3. The earlier lifecycle protections remain intact: switches wait for in-flight slot startup, concurrent switch requests queue latest-wins, and provider-native session registries are cleared only on effective provider changes.
✅ 4. GitHub CI is green for head `81233e67`; I did not run local checks/tests per review-mode policy.

**Executed**

PR #1058, opened 2026-06-15
fix(client): fail closed on source repo repoint blockers

baixiaohang:
## Summary

- fail closed before repointing a source repo origin when the checkout is dirty or actively in use
- keep same-repo local-commit protection, but disable that guard only after a confirmed URL repoint so a clean checkout can move to the new upstream
- distinguish matched, different, and missing/unreadable origin states so missing origin still preserves clean local commits
- roll back unconfirmed origin URL writes when a repoint fetch fails, so retries continue to fail closed instead of degrading to `stale-offline` on the old checkout
- add deterministic regression coverage for clean shared-base repoint, dirty/in-use repoint blockers, missing-origin local commits, and failed-repoint retry rollback

## Testing

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "origin is being repointed" --maxWorkers 1 --minWorkers 1`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts --maxWorkers 1 --minWorkers 1`
- `pnpm check` (exits 0; existing unrelated Biome hints still print)
- `pnpm typecheck`

### DIFF
--- a/packages/client/src/runtime/git-mirror-manager.ts
+++ b/packages/client/src/runtime/git-mirror-manager.ts
@@ -407,21 +407,21 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
    * so the subsequent fetch + `checkout -B` track the configured upstream
    * instead of silently serving the old one.
    */
+  type OriginUrlState = "matched" | "different" | "missing-or-unreadable";
+
   /**
-   * True when `absTarget`'s current `remote.origin.url` already canonically
-   * matches the configured `url` — i.e. reconcileOrigin would NOT repoint origin
-   * to a different repo. Mirrors reconcileOrigin's own canonical comparison.
-   * Gates the transient `stale-offline` degrade: serving the local checkout is
-   * only safe when it is the SAME repo. Treats an unreadable / missing origin as
-   * "not matching" (fail closed).
+   * Classify the current `remote.origin.url` before reconciling it. A confirmed
+   * canonical difference means a real URL repoint; a missing / unreadable origin
+   * is ambiguous and must keep same-repo local-commit protection.
    */
-  async function originCanonicallyMatches(absTarget: string, url: string): Promise<boolean> {
+  async function originUrlState(absTarget: string, url: string): Promise<OriginUrlState> {
     try {
       const { stdout } = await git(["config", "--get", "remote.origin.url"], absTarget, 10_000);
       const current = stdout.trim();
-      return current.length > 0 && canonicalizeRepoUrl(current) === canonicalizeRepoUrl(url);
+      if (!current) return "missing-or-unreadable";
+      return canonicalizeRepoUrl(current) === canonicalizeRepoUrl(url) ? "matched" : "different";
     } catch {
-      return false;
+      return "missing-or-unreadable";
     }
   }
 
@@ -706,13 +706,13 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
 
         // (4) Managed standalone clone → reconcile origin, fetch, then update when safe.
         //
-        // Capture whether origin ALREADY canonically matched the configured
-        // upstream *before* reconcileOrigin can repoint it. The transient
-        // degrade below is only safe for the SAME repo: if this call is
-        // repointing origin to a different repo and the confirming fetch then
-        // fails, the local checkout is the OLD repo's content and must NOT be
-        // served as the newly-configured source.
-        const originMatchedBeforeFetch = await originCanonicallyMatches(absTarget, url);
+        // Capture origin state *before* reconcileOrigin can mutate it. Only a
+        // confirmed canonical difference is a real URL repoint. A missing /
+        // unreadable origin is ambiguous: fetch failures must fail closed, but a
+        // successful fetch must still protect same-repo local commits.
+        const originStateBeforeFetch = await originUrlState(absTarget, url);
+        const originMatchedBeforeFetch = originStateBeforeFetch === "matched";
+        const originDiffersBeforeFetch = originStateBeforeFetch === "different";
         if (!originMatchedBeforeFetch) {
           if (activelyInUse) {
             throw repointBlockedError(absTarget, url, "another live session is using the checkout");
@@ -809,7 +809,7 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
         }
 
         const before = await headCommit(absTarget);
-        const update = await updateToLatest(absTarget, ref, { protectLocalCommits: originMatchedBeforeFetch });
+        const update = await updateToLatest(absTarget, ref, { protectLocalCommits: !originDiffersBeforeFetch });
         if (update === "skipped-local-commits") {
           log?.warn(
             { clonePath: absTarget },
--- a/packages/client/src/runtime/git-mirror-manager.ts
+++ b/packages/client/src/runtime/git-mirror-manager.ts
@@ -596,13 +596,16 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
    * current HEAD is strictly ahead of the resolved branch tip on the *same*
    * lineage — i.e. someone committed local work on top of the tracked branch.
    * `checkout -B` would orphan those commits, so we refuse and leave the
-   * checkout as-is. A diverged/unrelated HEAD (e.g. after an origin repoint) is
-   * NOT ahead-on-the-same-lineage, so it still resets cleanly. A pinned SHA
-   * `ref` is explicit operator intent and is always applied.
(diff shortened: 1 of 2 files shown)


packages/client/src/runtime/git-mirror-manager.ts
```diff
@@ -785,7 +809,7 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
         }
 
         const before = await headCommit(absTarget);
-        const update = await updateToLatest(absTarget, ref);
+        const update = await updateToLatest(absTarget, ref, { protectLocalCommits: originMatchedBeforeFetch });
```
yuezengwu: Blocker: this disables the local-commit guard for more than a confirmed URL repoint. `originMatchedBeforeFetch === false` also covers a standalone clone whose `remote.origin.url` is missing or unreadable. In that state, if the clone has clean local commits ahead of the same configured repo, `reconcileOrigin` adds `origin`, `fetchClone` succeeds, and this call resets with `protectLocalCommits: false`, orphaning the local commit instead of returning `skipped-local-commits`.

I reproduced it through the built manager API: clone repo A, commit clean local work, `git remote remove origin`, then call `ensureSourceRepo({ url: repoA, clonePath })`. Current result is `outcome: "updated"` and `HEAD` moves back to the remote tip (`preserved: false`). That violates the PR summary’s “keep same-repo local-commit protection, but disable that guard after a confirmed URL repoint”. Please distinguish `matches` / `differs` / `missing-or-unreadable` and only disable the guard for the confirmed `differs` case, or otherwise fail closed for the ambiguous origin state.

yuezengwu (changes requested):
实现目标我理解为：source repo 配置 URL 变更时，不能在 dirty 或 in-use checkout 上先改 `origin` 再继续提供旧内容；同时 clean checkout 在确认 repoint 后要能切到新 upstream，并保留 same-repo local commit 保护。

核心改动方向是对的，新增 dirty/in-use repoint blocker 和 shared-base clean repoint regression 也覆盖了主路径。我跑了：

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "repointing origin" --maxWorkers 1 --minWorkers 1`，2/2 passed
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts --maxWorkers 1 --minWorkers 1`，191/191 passed
- `pnpm typecheck`，passed

但我发现一个 blocker：当前实现把“origin 不匹配”同时用于“确认的 URL repoint”和“origin 缺失/不可读”。后者不是确认 repoint，会绕过 same-repo local-commit guard，可能 orphan 本地 clean commits。见 inline comment。

packages/client/src/runtime/git-mirror-manager.ts
baixiaohang: Fixed in `543904ff`.

I split the previous boolean into `matched` / `different` / `missing-or-unreadable` origin states. The local-commit guard is disabled only for the confirmed `different` URL repoint case. Missing/unreadable origin stays ambiguous: transient fetch still cannot degrade, but a successful fetch keeps same-repo local-commit protection.

Added regression coverage for `git remote remove origin` + clean local commit; it now returns `skipped-local-commits`, restores `remote.origin.url`, and preserves `HEAD` / the local file.

Re-ran:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "origin is missing" --maxWorkers 1 --minWorkers 1`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "repointing origin" --maxWorkers 1 --minWorkers 1`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts --maxWorkers 1 --minWorkers 1` (192 tests)
- `pnpm check`
- `pnpm typecheck`

packages/client/src/runtime/git-mirror-manager.ts
```diff
@@ -696,13 +706,21 @@ export function createGitMirrorManager(opts: GitMirrorManagerOptions): GitMirror
...
+        if (!originMatchedBeforeFetch) {
+          if (activelyInUse) {
+            throw repointBlockedError(absTarget, url, "another live session is using the checkout");
+          }
+          if (await isDirty(absTarget)) {
+            throw repointBlockedError(absTarget, url, "the checkout has local changes");
+          }
+        }
         try {
           await reconcileOrigin(absTarget, url);
```
yuezengwu, line 747: Blocker: `reconcileOrigin` mutates `remote.origin.url` before the confirming fetch succeeds, and the catch path does not roll it back. That means the fail-closed behavior only holds for the first failed repoint attempt. On the next session, `originUrlState` sees the already-mutated URL as `matched`, so the transient-fetch degrade path at line 757 can return `stale-offline` and serve the old repo checkout as the newly configured repo.

I reproduced this on `543904ff` through the built manager API: clone repo A, then call `ensureSourceRepo({ url: "https://127.0.0.1:1/different.git", clonePath })`. First call throws, but leaves `remote.origin.url` set to that new URL. A second call with the same URL returns `{ outcome: "stale-offline" }` with `HEAD` still at repo A (`servedOld: true`).

Please either roll back `origin` when fetch fails before a repoint is confirmed, or otherwise persist enough state to prevent same-repo stale-offline degradation until a successful fetch has confirmed the new upstream.

yuezengwu (changes requested):
`543904ff` fixes the previous missing-origin local-commit blocker, and I resolved that outdated thread. The new `OriginUrlState` split is the right shape for that issue.

I still found one blocker in the same fail-closed boundary: a failed repoint fetch mutates `remote.origin.url` before the fetch confirms the new upstream. The first call fails closed, but the second call now sees origin as matched and degrades to `stale-offline`, serving the old repo checkout as the new configured repo.

What I ran on the updated head:

- one-off built-manager repro for failed repoint retry, result: first call throws, `originAfterFirst` is the new unreachable URL, second call returns `outcome: "stale-offline"`, `servedOld: true`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "origin is missing" --maxWorkers 1 --minWorkers 1`, 1/1 passed

Please keep the repoint unconfirmed until fetch succeeds, or roll back `origin` when fetch fails before confirmation. The fail-closed guarantee needs to hold across retries, not just the first attempt.

packages/client/src/runtime/git-mirror-manager.ts
baixiaohang: Fixed in `209ad16c`.

The origin state now carries the previous URL. When `reconcileOrigin` has to add/set `origin` for a non-matched state and the confirming fetch fails, the catch path rolls that unconfirmed origin change back before rethrowing. That keeps the retry classified as a repoint/missing-origin case instead of a same-repo outage, so `stale-offline` remains gated to origins that already matched before fetch.

I extended the transient repoint test to call `ensureSourceRepo` twice with the unreachable new URL and assert both calls throw while `remote.origin.url` and `HEAD` remain on repo A.

Re-ran:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts -t "origin is being repointed" --maxWorkers 1 --minWorkers 1`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts --maxWorkers 1 --minWorkers 1` (192 tests)
- `pnpm check`
- `pnpm typecheck`

yuezengwu (approved):
目标和核心改动：这组改动让 source repo URL repoint 在 dirty / actively-in-use / unconfirmed-fetch 场景 fail closed，不再把旧 checkout 当作新配置 repo 服务；同时只在确认的 URL repoint 后关闭 same-repo local-commit guard，missing/unreadable origin 仍保留 clean local commit 保护。

我复核了 `209ad16c`：

- previous missing-origin blocker 已由 `OriginUrlState` 拆分修复，并有 regression coverage。
- previous failed-repoint retry blocker 已由 unconfirmed-origin rollback 修复；我重新跑了 built-manager repro，第一次 failed repoint 后 `origin` 回滚到旧 URL，第二次继续 throw，不再 `stale-offline`。
- dirty/in-use repoint 仍在 mutation 前 fail closed，clean shared-base repoint 仍能移动到新 upstream。

Verification:

- one-off built-manager repro for failed repoint retry on `209ad16c`: first call throws, `originAfterFirst` stays old URL, second call throws, no stale-offline degrade
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/git-mirror-manager.test.ts --maxWorkers 1 --minWorkers 1`，192/192 passed
- `pnpm typecheck`，passed

No remaining blockers from my review.

PR #1063, opened 2026-06-15
refactor: place agent source repos under <workspace>/source-repos/<name>

liuchao-001:
## Summary

Adds a `source-repos/` directory layer to the agent workspace so source clones live at `<workspace>/source-repos/<name>` instead of directly at the workspace root, grouping them apart from the tree clone (`<workspace>/context-tree`), `worktrees/`, and workspace state dirs. The Context Tree is unchanged.

**W1 manifest representation (option A, confirmed with the requester):** `workspaceManifestSchema` gains an optional `sourcesRoot` field (`SOURCE_REPOS_DIRNAME = "source-repos"`); `sources` stay bare names, each resolving to `<workspace>/<sourcesRoot>/<name>`. The field is **optional for back-compat** — a manifest without `sourcesRoot` reads as the legacy flat layout (sources at the workspace root), so existing local workspaces and the `migrate-workspace` flat flow keep working unchanged.

## Changes

- **shared** (`workspace-manifest.ts`): optional `sourcesRoot` + `sourcesRoot !== tree` refine + barrel export of `SOURCE_REPOS_DIRNAME`.
- **client**: `resolveGitRepoTargetPath` resolves under `source-repos/` (escape guard still anchored at the workspace root); `ensureWorkspaceManifest` writes `sourcesRoot`; the briefing's *Source Repositories* protocol describes the `source-repos/` location. Source-path consumers (codex `additionalDirectories`, claude-code handlers) flow through the resolver, so they update automatically.
- **CLI** (`workspace.ts`): `computeWorkspaceStatus` resolves bound sources and scans unbound siblings under `sourcesRoot` when set (flat when absent); `listImmediateChildDirs` tolerates a not-yet-materialised `source-repos/`.
- **skills**: `first-tree-seed` / `first-tree-sync` / `first-tree` locate source clones at `<workspace>/source-repos/<name>` (legacy-flat fallback noted).

## Back-compat / migration

Per the zero-deletion invariant the runtime never moves an existing clone. An already-provisioned agent's old top-level clone is left in place; following the new briefing the agent re-clones under `source-repos/`, and the old clone is left for the operator to remove. No data migration. The optional `sourcesRoot` is the seam: present = new layout, absent = legacy flat.

## Test plan

- [x] shared: schema accepts/omits `sourcesRoot`, rejects `sourcesRoot === tree` and non-single-segment values.
- [x] CLI: `computeWorkspaceStatus` resolves bound sources + unbound siblings under `sourcesRoot`; flat path unchanged.
- [x] client: manifest writer emits `sourcesRoot`; `resolveGitRepoTargetPath` → `source-repos/<name>`; briefing + codex `additionalDirectories` carry the new paths.
- [x] `pnpm check` + `pnpm typecheck` clean; full suite green (server 1419, client 972, shared, cli).

Companion tree PR (records the layout in `client-runtime.md`): _to follow._


baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: a clean agent workspace writes `sourcesRoot: "source-repos"`, but the first-clone protocol does not create the `source-repos/` parent directory, so source repo materialization fails before any worktree can be created.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: move agent-managed source clones from workspace-root siblings to `<workspace>/source-repos/<name>` while keeping legacy flat manifests readable.
- Design: adds optional `sourcesRoot`, updates client source path resolution and manifest writing, adjusts CLI workspace status, and rewrites shipped skill/briefing text for the new layout.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. Fresh workspaces cannot create the first source clone under `source-repos/`. [R4 / `packages/client/src/runtime/agent-briefing.ts:415`]

`ensureWorkspaceManifest` now writes `sourcesRoot: "source-repos"`, and `resolveGitRepoTargetPath` points every declared repo at `<workspace>/source-repos/<name>`, but bootstrap only creates `.first-tree/` and the new briefing says `git clone` creates the parent directory. Git creates the target directory, not missing parents, so on a clean agent home with no existing `source-repos/`, the documented first-use command `git clone --bare <url> <workspace>/source-repos/<name>` fails with a missing parent directory. The test currently locks in the false assumption by asserting the briefing must not contain `mkdir -p`. Please either create `<workspace>/source-repos` during client bootstrap, or put explicit parent creation in the agent protocol and update the test accordingly.

✅ 2. The optional `sourcesRoot` schema shape and legacy-flat reader behavior are directionally sound; the CLI status path correctly switches bound-source resolution based on field presence.

**Executed**
- Submitted request_changes.
- Did not run local tests or checks per review protocol.
- Context Tree read used the local checkout because `first-tree-staging tree tree` degraded with `Cannot fast-forward to multiple branches` while selecting relevant nodes.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: the latest head only updates `first-tree-sync` documentation; the fresh-workspace clone blocker from the previous review is still present in the client briefing/bootstrap path.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `ed412286` aligns `skills/first-tree-sync/references/audit-workflow.md` with the `source-repos/` layout.
- Remaining design: optional `sourcesRoot`, client resolver/manifest changes, CLI status handling, and shipped skill text for source clones under `<workspace>/source-repos/<name>`.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. The fresh-workspace clone blocker remains on the latest head. [R4 / `packages/client/src/runtime/agent-briefing.ts:415`]

`ed412286` did not change the client bootstrap/briefing path. The briefing still says `git clone` creates the `source-repos/` parent and then tells the agent to run `git clone --bare <url> <path>` at `<workspace>/source-repos/<name>`. The test still encodes that assumption by asserting the briefing must not contain `mkdir -p`, while `ensureWorkspaceManifest` still only creates `.first-tree/`. Git creates the target directory, not missing parents, so a clean agent home still cannot materialize the first source clone under `source-repos/`. Please create `<workspace>/source-repos` during client bootstrap, or make the agent protocol explicitly create the parent directory and update the briefing test.

✅ 2. The new `first-tree-sync` audit-workflow docs fix is consistent with the intended `sourcesRoot` layout and does not introduce a new issue.

**Executed**
- Submitted follow-up request_changes on head `ed412286`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still partially in progress when reviewed.

liuchao-001:
Pushed `2647990b` addressing the fresh-workspace clone finding.

**On the premise:** `git clone` *does* create missing parent directories of the clone target — this was never a blocking bug. Verified directly (git 2.51.1):

```
$ ls /tmp/t            # source-repos/ does NOT exist
$ git clone --bare <url> /tmp/t/source-repos/first-tree
Cloning into bare repository '/tmp/t/source-repos/first-tree'... done.   # exit 0
$ ls /tmp/t            # → source-repos/  (created)
```

This matches @code-reviewer's independent live E2E against real `first-tree.git` ("git clone --bare → source-repos/<name>, auto-creates the parent, bare confirmed"). So a clean agent home materialises the first clone fine.

**The fix anyway:** the briefing is a protocol an agent follows *loosely*, so relying on git's implicit leading-dir creation is fragile if an agent deviates from the exact clone command. Per your suggested remediation ("make the agent protocol explicitly create the parent directory and update the briefing test"), the Source Repositories protocol now runs `mkdir -p "$(dirname <path>)"` before the clone, and the briefing test asserts its presence instead of its absence. I kept this in the **agent protocol** rather than client bootstrap because the runtime is declaration-and-observation only under the agent-managed model (it creates no repo dirs); making the agent own parent creation keeps that boundary clean.

`pnpm check` + briefing/e2e tests green. Re-review welcome. @baixiaohang

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: the previous clone-parent blocker is fixed, but the CLI status path still detects only worktree checkouts while the new `source-repos/` layout stores agent-managed sources as bare clones.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `2647990b` makes the briefing explicitly create the `source-repos/` parent before cloning.
- Design: adds optional `sourcesRoot`, moves client-resolved source paths under `<workspace>/source-repos/<name>`, updates workspace status scanning, and adjusts shipped skill/briefing text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. `computeWorkspaceStatus` still misses actual bare source clones under `source-repos/`. [R5 / `apps/cli/src/core/workspace.ts:161`]

The new layout says agent-managed source repos are bare clones, and `computeWorkspaceStatus` now scans `sourcesBase = <workspace>/source-repos` for unbound siblings. But the helper used for both `readGitRemoteUrl()` and `unboundGitSiblings` still defines a git repo as `isDirectory(path) && existsSync(join(path, ".git"))`, which only matches normal working-tree checkouts. A real bare clone at `<workspace>/source-repos/scratch` has `HEAD`, `objects/`, `refs/`, etc. directly under the repo path and no `.git/`, so the status report will not list it as an unbound sibling. The same guard also prevents `remoteUrl` from being populated for present bound bare clones. The added test creates `scratch` via `git init` (non-bare), so it passes while missing the actual agent-managed shape. Please update git-repo detection / remote-url reading to handle bare clones (for example by using `git -C <path> rev-parse` or equivalent) and add a `source-repos/` bare-clone regression test.

✅ 2. The previous fresh-clone protocol issue is addressed: the briefing now explicitly creates the parent directory with `mkdir -p "$(dirname <path>)"`, and the test asserts that protocol.

**Executed**
- Submitted request_changes on head `2647990b`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still in progress when reviewed.

liuchao-001:
Pushed `f38fa480` (+ `385bf9e0`) addressing the bare-clone finding.

**R5 — valid, fixed.** You're right: `computeWorkspaceStatus`'s `isGitRepoDir` was `isDirectory(path) && existsSync(path/.git)`, which only matches a normal working-tree checkout. Agent-managed source clones under `source-repos/<name>` are **bare** (HEAD / objects/ / refs/ at the repo root, no `.git/`), so the scan missed them as unbound siblings and `readGitRemoteUrl` (gated on the same helper) returned nothing for present bound bare clones. And my earlier `source-repos/` test used `git init` (non-bare), masking it — good catch.

`f38fa480`:
- `isGitRepoDir` now also recognizes the bare shape (`HEAD` + `objects/` + `refs/`), filesystem-only, no subprocess.
- The `source-repos/` status regression now uses real `git init --bare` clones, asserts the bare unbound sibling is detected, and asserts a bound bare clone's `remoteUrl` resolves.

**Plus the optional escape-guard hardening** (`385bf9e0`, suggested by the in-chat reviewer): `resolveGitRepoTargetPath` now anchors its escape check at `<workspace>/source-repos` rather than the workspace root, so a `..`-bearing localPath that would climb into a workspace-root sibling (e.g. `context-tree/`) is rejected — locally self-consistent rather than relying on the upstream safety check staying strict (currently unreachable, pure defense-in-depth).

`pnpm typecheck` 12/12, `pnpm check` clean, full suite green (cli workspace 28 incl. the bare-clone case, client 972). Re-review welcome. @baixiaohang

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: the previous blockers are fixed, but one client doc-preview source-path consumer still resolves single-repo paths as the legacy flat `<workspace>/<localPath>` instead of the new `<workspace>/source-repos/<name>` layout.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `385bf9e0` detects bare source clones in CLI status and anchors the client source-repo escape guard at `source-repos/`.
- Design: adds optional `sourcesRoot`, keeps manifest `sources` as bare names, routes declared source repos through `source-repos/`, updates CLI status scanning, and refreshes the shipped briefing/skills text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. Single-repo doc preview still points at the legacy flat source path. [R5 / `packages/client/src/runtime/session-manager.ts:166`]

`documentBasePathFromRuntimeConfig()` still returns `join(sessionRoot, localPath)`, and `selfFenceFromRuntimeConfig()` still passes `singleRepoLocalPath` through as the bare payload `localPath`. The snapshot pipeline then resolves relative markdown refs against `<agentHome>/<localPath>` and exports the same old location through `FIRST_TREE_DOC_BASE` / `FIRST_TREE_DOC_REPO_LOCAL_PATH`. With this PR, the declared source repo path for a fresh non-legacy agent is `<agentHome>/source-repos/<localPath>` via `resolveGitRepoTargetPath`, so a single-source agent that writes `docs/foo.md` in a response or invokes `chat send` still snapshots/promotes using the old flat path. In a fresh workspace that silently drops promotion/falls back to the agent home; in an upgraded workspace with a leftover top-level source directory, it can continue using the old path after the manifest declares `sourcesRoot`. Please route this helper through the same source-path resolver (or pass `source-repos/<name>` as the self-fence local path) and update `document-base-path`, `agent-io`, and doc-snapshot tests to cover the new layout and any intentional legacy exception.

✅ 2. The previous fresh-clone protocol issue is fixed: the briefing now explicitly creates the `source-repos/` parent before cloning.

✅ 3. The previous CLI status issue is fixed: `computeWorkspaceStatus` now recognizes bare source clones and the regression test covers both bound remoteUrl and unbound bare siblings under `sourcesRoot`.

**Executed**
- Submitted request_changes on head `385bf9e0`.
- Did not run local tests or checks per review protocol.
- GitHub checks were green when reviewed.

packages/shared/src/schemas/workspace-manifest.ts
```diff
@@ -44,13 +50,27 @@ export const workspaceManifestSchema = z
     sources: z.array(subdirectoryNameSchema).refine((values) => new Set(values).size === values.length, {
       message: "sources must not contain duplicate entries",
     }),
+    sourcesRoot: subdirectoryNameSchema.optional(),
   })
   .refine((manifest) => !manifest.sources.includes(manifest.tree), {
```
yuezengwu: With `sourcesRoot` present, `tree` and `sources` no longer share the same filesystem namespace: `tree: "context-tree"` resolves to `<workspace>/context-tree`, while `sources: ["context-tree"]` would resolve to `<workspace>/source-repos/context-tree`. This unconditional refinement still rejects that valid new-layout case, and `ensureWorkspaceManifest` currently skips writing the whole manifest when a configured source is named `context-tree`. That can leave an otherwise tree-bound agent with no `.first-tree/workspace.json`.

Please either condition this collision check on the legacy flat layout (`sourcesRoot` absent), or explicitly document that `context-tree` remains a reserved source localPath even under `sourcesRoot` and avoid dropping the entire manifest for the remaining sources.

skills/first-tree-sync/references/audit-workflow.md
```diff
@@ -61,11 +64,13 @@ default branch — following the **Worktrees** protocol in your `AGENTS.md`
 / `CLAUDE.md` briefing — and remove it when the audit is done:
 
 ```bash
-# for each <source> in manifest.sources:
-git -C <workspaceRoot>/<source> fetch origin
-git -C <workspaceRoot>/<source> worktree add <workspaceRoot>/worktrees/sync-<source> origin/main
+# for each <source> in manifest.sources — the bare clone is at
+# <workspaceRoot>/<manifest.sourcesRoot>/<source>, i.e. <workspaceRoot>/source-repos/<source>
+# (a legacy flat manifest omits sourcesRoot → <workspaceRoot>/<source>):
+git -C <workspaceRoot>/source-repos/<source> fetch origin
```
yuezengwu: This prose says legacy flat manifests omit `sourcesRoot` and keep clones at `<workspaceRoot>/<source>`, but the executable commands still hard-code `<workspaceRoot>/source-repos/<source>`. An agent following this skill in an existing flat workspace will run `git -C` against a missing path and fail the audit.

Please derive a `sourceBase` from the manifest (`sourcesRoot ? <workspaceRoot>/<sourcesRoot> : <workspaceRoot>`) or show explicit new-layout vs legacy-flat command branches.

skills/first-tree-seed/SKILL.md
```diff
@@ -80,30 +80,35 @@ refuses; route subsequent work through `first-tree-context` or
...
 structural read, materialize one read worktree per source off its latest
 default branch, following the **Worktrees** protocol in your `AGENTS.md`
 / `CLAUDE.md` briefing:
 
 ```bash
-# for each <source> in manifest.sources:
-git -C <workspaceRoot>/<source> fetch origin
-git -C <workspaceRoot>/<source> worktree add <workspaceRoot>/worktrees/seed-<source> origin/main
+# for each <source> in manifest.sources (clone at <workspaceRoot>/source-repos/<source>):
+git -C <workspaceRoot>/source-repos/<source> fetch origin
```
yuezengwu: Same back-compat issue as the sync workflow: the paragraph correctly says legacy flat manifests keep sources at `<workspaceRoot>/<name>`, but the commands below only work for `sourcesRoot: "source-repos"`. For a legacy flat workspace, seed will try to fetch from a nonexistent `<workspaceRoot>/source-repos/<source>`.

Please make the command block compute or branch on `manifest.sourcesRoot`, so the skill’s executable protocol matches the optional-field compatibility contract.

yuezengwu (changes requested):
目标和核心改动我理解为：把 agent-managed source clones 从 workspace root 迁到 `<workspace>/source-repos/<name>`，同时用 optional `workspace.json.sourcesRoot` 保持旧 flat manifest 可读；配套改了 client resolver / manifest writer / briefing、CLI workspace status，以及 shipped First Tree skills。

这是 W1 workspace manifest/runtime layout 的核心数据结构和运行时契约变更，需要 yuezengwu 重点看 schema 与 back-compat 语义。

I’m requesting changes for two compatibility gaps on the latest head (`385bf9e0`). The earlier clone-parent protocol and bare-clone status findings look addressed in this head.

liuchao-001:
Pushed `12273327` addressing the doc-preview source-path finding — **valid, fixed.**

You're right: `documentBasePathFromRuntimeConfig` returned `join(sessionRoot, localPath)` and `selfFenceFromRuntimeConfig` set `singleRepoLocalPath` to the bare name, so for a single-source agent the doc-snapshot pipeline — and the `FIRST_TREE_DOC_BASE` / `FIRST_TREE_DOC_REPO_LOCAL_PATH` env exported to the `chat send` binary — still resolved docs at the legacy flat `<agentHome>/<localPath>`, not the new `<agentHome>/source-repos/<localPath>`. So `docs/foo.md` promotion/snapshot silently fell back to the agent home.

`12273327`:
- `documentBasePathFromRuntimeConfig` → `<sessionRoot>/source-repos/<localPath>`.
- `selfFenceFromRuntimeConfig`'s `singleRepoLocalPath` → `source-repos/<name>` (the agentHome-relative path the snapshot pipeline resolves via `resolve(agentHome, …)`, consumed identically by the runtime result-sink and the `chat send` env path). `singleRepoLocalPathFromPayload` still returns the bare manifest name; the `source-repos/` prefix is applied at the two consumers so the env / sessionRoot / SelfFence stay single-sourced.
- Tests updated to the new layout: `document-base-path`, `agent-io`, `session-manager-edge-coverage` (FIRST_TREE_DOC_BASE = `…/source-repos/project`, FIRST_TREE_DOC_REPO_LOCAL_PATH = `source-repos/project`), plus a `doc-snapshots` case exercising a real `source-repos/<name>` promotion key.

For the migration edge you noted (upgraded workspace with a leftover top-level clone): the briefing's fail-closed repoint guard + the manifest's `sourcesRoot` mean the agent re-clones under `source-repos/` and previews target the new location; the orphaned top-level dir is left for the operator (zero-deletion), consistent with the rest of the PR.

`pnpm typecheck` 12/12, `pnpm check` clean, full suite green (client 973). Re-review welcome. @baixiaohang

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: `12273327` fixes the single-repo doc-preview path issue, but the two manifest/skill back-compat gaps from the latest review round are still present on the new head.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `12273327` routes single-repo doc preview and `chat send` doc env through `source-repos/<name>`.
- Design: adds optional `sourcesRoot`, keeps manifest `sources` as names, moves client-declared source repos under `source-repos/`, updates CLI status scanning, and refreshes shipped skill text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. New-layout manifests still reject a valid `context-tree` source name and the writer still skips the whole manifest. [R5 / `packages/shared/src/schemas/workspace-manifest.ts:55`]

With `sourcesRoot` present, `tree: "context-tree"` resolves to `<workspace>/context-tree`, while a source named `context-tree` resolves to `<workspace>/source-repos/context-tree`; they no longer collide on disk. The schema still unconditionally rejects `sources.includes(tree)`, and `ensureWorkspaceManifest()` still returns without writing any manifest when a configured source is named `context-tree`. That means a valid new-layout source localPath can leave the agent without `.first-tree/workspace.json` at all. Please either make the collision check legacy-flat-only (`sourcesRoot` absent), or explicitly reserve the source name in the new layout too and keep writing the manifest for the remaining usable sources.

❌ 2. Seed/sync executable command blocks still break legacy flat manifests. [R5 / `skills/first-tree-sync/references/audit-workflow.md:70`]

The prose correctly says a manifest without `sourcesRoot` keeps sources at `<workspaceRoot>/<source>`, but the actual `git -C` commands in both `first-tree-sync` and `first-tree-seed` still hard-code `<workspaceRoot>/source-repos/<source>`. An agent following those skills in an existing flat workspace will fetch/worktree from a missing path. Please compute a `sourceBase` from `manifest.sourcesRoot` or show explicit new-layout vs legacy-flat command branches in both skill protocols.

✅ 3. The previous single-repo doc-preview finding is fixed: `documentBasePathFromRuntimeConfig()` and `selfFenceFromRuntimeConfig()` now use the agent-home-relative `source-repos/<name>` path, with updated document-base/doc-snapshot/env tests.

**Executed**
- Submitted request_changes on head `12273327`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still in progress when reviewed.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: latest head `e2bdea59` only aligns client comments/docs; the two remaining manifest/skill compatibility blockers from head `12273327` are still present.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `e2bdea59` updates client comments to describe `source-repos/` correctly.
- Design: optional `sourcesRoot`, source clones under `source-repos/`, CLI status scanning, client resolver/briefing/doc-preview updates, and shipped skill text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. The new-layout `context-tree` source-name compatibility gap remains. [R5 / `packages/shared/src/schemas/workspace-manifest.ts:55`]

The schema still unconditionally rejects `sources.includes(tree)`, and `ensureWorkspaceManifest()` still skips writing the whole manifest when a configured source is named `context-tree`. With `sourcesRoot` present, that source resolves under `<workspace>/source-repos/context-tree`, so it no longer collides with the tree clone at `<workspace>/context-tree`. Please either make this collision check legacy-flat-only, or explicitly reserve that source name under the new layout and keep writing the manifest for the remaining usable sources.

❌ 2. The seed/sync legacy-flat command gap remains. [R5 / `skills/first-tree-sync/references/audit-workflow.md:70`]

Both skill docs still say legacy flat manifests omit `sourcesRoot`, but their executable `git -C` command blocks still hard-code `<workspaceRoot>/source-repos/<source>`. Existing flat workspaces will run against a missing path. Please compute a `sourceBase` from the manifest or show explicit new-layout vs legacy-flat branches in both `first-tree-sync` and `first-tree-seed`.

✅ 3. The latest client comment cleanup is consistent with the `source-repos/` layout and does not introduce a new issue.

**Executed**
- Submitted follow-up request_changes on head `e2bdea59`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still in progress when reviewed.

liuchao-001:
Pushed `29283bff` addressing both compatibility gaps from your review, @yuezengwu — both valid, thanks.

**1. `tree ∉ sources` refine + manifest drop (schema/back-compat).** You're right: once `sourcesRoot` is set, `tree` (`<ws>/<tree>`) and a same-named source (`<ws>/<sourcesRoot>/<name>`) are different namespaces, so the unconditional refine wrongly rejected the valid new-layout case, and `ensureWorkspaceManifest` dropping the whole manifest could leave a tree-bound agent with no `workspace.json`.
- Schema: the `tree ∉ sources` refine now applies **only to the legacy flat layout** (`sourcesRoot` absent). With `sourcesRoot` present a source named `context-tree` parses fine.
- `ensureWorkspaceManifest` no longer special-cases / drops a `context-tree`-named source — it always writes `sourcesRoot`, so the source lands at `source-repos/context-tree` and the manifest is written normally.
- Tests: schema accepts tree-named source under `sourcesRoot` + still rejects it flat; the client writer now **writes** (not drops) a `context-tree`-named source.

**2. Skill command blocks vs the legacy-flat contract.** Also valid — the prose claimed flat support but the commands hard-coded `source-repos/`. `first-tree-sync/audit-workflow` and `first-tree-seed` now resolve a `<source-clone>` from the manifest: `sourcesRoot` present → `<ws>/<sourcesRoot>/<source>`; absent → `<ws>/<source>`. So an agent in a legacy flat workspace runs `git -C` against the right path.

These are the same optional-`sourcesRoot` back-compat seam applied consistently to the schema collision rule, the manifest writer, and the skill protocols. `pnpm typecheck` 12/12, `pnpm check` clean, full suite green (shared/client manifest tests updated, client 973). Re-review welcome.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: the previous manifest/skill blockers are fixed, but the same namespace split still leaks in CLI status: unbound source repos under `sourcesRoot` are filtered against the tree name.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `29283bff` makes `tree ∈ sources` valid when `sourcesRoot` is present, stops dropping a `context-tree` source from the client manifest writer, and changes seed/sync commands to resolve a manifest-dependent `<source-clone>` path.
- Design: optional `sourcesRoot`, source clones under `source-repos/`, client resolver/briefing/doc-preview updates, CLI status scanning, and shipped skill text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. CLI status still filters `source-repos/<tree>` as if the tree shared the source namespace. [R5 / `apps/cli/src/core/workspace.ts:292`]

`29283bff` correctly updates the manifest schema so a source may share the tree name when `sourcesRoot` is present: `tree: "context-tree"` resolves to `<workspace>/context-tree`, while the source resolves to `<workspace>/source-repos/context-tree`. But `computeWorkspaceStatus()` still builds `declaredNames = new Set([manifest.tree, ...manifest.sources])` before scanning `sourcesBase`. In the new layout, `manifest.tree` is not inside `sourcesBase`, so an unbound bare clone at `<workspace>/source-repos/context-tree` will be skipped just because its name equals the tree. That hides an unbound source sibling that is now valid by schema. Please exclude `manifest.tree` only for the legacy flat layout (`sourcesRoot` absent), and add a regression where `sourcesRoot` is set and an unbound bare source sibling is named like the tree.

✅ 2. The previous schema/client manifest writer issue is fixed: `tree ∈ sources` is now rejected only for flat manifests, and `ensureWorkspaceManifest()` writes a `context-tree` source under `source-repos/`.

✅ 3. The previous seed/sync legacy-flat command issue is fixed: both skills now resolve `<source-clone>` from `sourcesRoot` when present and fall back to `<workspaceRoot>/<source>` when absent.

**Executed**
- Submitted request_changes on head `29283bff`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still in progress when reviewed.

liuchao-001:
@baixiaohang both findings #1 and #2 are already fixed on the **current HEAD `29283bff`** — your review was on `12273327`/`e2bdea59`, two commits back. (#3 doc-preview you've confirmed fixed.) Concretely, in `29283bff`:

- **#1 schema/manifest:** the `tree ∉ sources` refine is now **legacy-flat-only** (`workspace-manifest.ts`: `manifest.sourcesRoot !== undefined || !sources.includes(tree)`), so a `context-tree`-named source is valid under `sourcesRoot`; and `ensureWorkspaceManifest` no longer returns-without-writing for that name (it always writes `sourcesRoot`, so the source lands at `source-repos/context-tree` and the manifest is written). Tests: schema accepts tree-named source under `sourcesRoot` + still rejects it flat; client writer writes (not drops) it.
- **#2 skills:** both `first-tree-sync/audit-workflow` and `first-tree-seed` command blocks now resolve a `<source-clone>` from the manifest — `sourcesRoot` present → `<ws>/<sourcesRoot>/<source>`, absent → `<ws>/<source>` — instead of hard-coding `source-repos/`.

Please re-review HEAD `29283bff`. `pnpm typecheck` 12/12, `pnpm check` clean, full suite green.

baixiaohang:
Clarification: my latest review was already on HEAD `29283bff`, and it explicitly marks the schema/manifest writer and seed/sync skill issues as fixed.

The remaining request_changes is a new adjacent CLI-status namespace issue:

- `computeWorkspaceStatus()` scans `sourcesBase` (`<workspace>/source-repos` when `sourcesRoot` is present).
- It still builds `declaredNames = new Set([manifest.tree, ...manifest.sources])`.
- In the new layout, `manifest.tree` is not in `sourcesBase`, so an unbound bare clone at `<workspace>/source-repos/context-tree` is incorrectly skipped just because its name equals the tree.

That should be conditional: include `manifest.tree` in the exclusion set only for the legacy flat layout (`sourcesRoot` absent), and add a regression for an unbound bare sibling named like the tree under `sourcesRoot`.

packages/client/src/runtime/session-manager.ts
```diff
@@ -155,17 +160,18 @@ export function resolveSessionDocRoot(workspaceRoot: string, chatId: string): st
...
+ *    Returning a bare relative `localPath` (the old behaviour) made the runtime
+ *    resolve it against its own `process.cwd()` — the launch dir, not the
+ *    session workspace — so it silently failed to find any doc and cloud
+ *    preview was dead.
  *  - zero or multiple repos → the session doc root.
  */
 export function documentBasePathFromRuntimeConfig(payload: AgentRuntimeConfigPayload, sessionRoot: string): string {
   const localPath = singleRepoLocalPathFromPayload(payload);
-  return localPath ? join(sessionRoot, localPath) : sessionRoot;
+  return localPath ? join(sessionRoot, SOURCE_REPOS_DIRNAME, localPath) : sessionRoot;
```
yuezengwu: This now breaks the legacy per-chat layout documented just above. `resolveSessionDocRoot()` intentionally returns `<workspaceRoot>/<chatId>` when a pre-#506 chat directory exists, and the comment says those legacy chats keep source repos at `<workspaceRoot>/<chatId>/<localPath>`. For such a session, this line now computes `<workspaceRoot>/<chatId>/source-repos/<localPath>`, so single-repo markdown previews look in a directory that does not exist and regress to no snapshot.

Please preserve the legacy path branch, e.g. have `resolveSessionDocRoot` return whether it selected the legacy per-chat root, or otherwise make `documentBasePathFromRuntimeConfig` choose `<sessionRoot>/<localPath>` for legacy roots and `<agentHome>/source-repos/<localPath>` only for new agent-home sessions. Add a regression test with an existing legacy chat dir plus a single repo.

yuezengwu (changes requested):
第二轮 review，基于最新 head `e2bdea59`。

目标和核心改动仍是：把 agent source clones 从 workspace root 移到 `<workspace>/source-repos/<name>`，并用 optional `workspace.json.sourcesRoot` 维持 legacy flat manifest 读取兼容。这仍然是 W1 workspace manifest / client runtime layout 的核心契约变更，建议 yuezengwu 重点检查 schema 与 legacy 语义。

我继续 request changes：上轮两个问题仍未修复，且最新同步新增了一个 legacy doc-preview 回归。

仍未修复：
1. `workspaceManifestSchema` 仍无条件拒绝 `sources` 里出现 `tree` 名称。`sourcesRoot` 存在时两者不在同一目录 namespace，`context-tree` source 会解析到 `<workspace>/source-repos/context-tree`，但 client writer 仍会跳过整个 manifest。
2. `first-tree-seed` 和 `first-tree-sync` 的命令块仍硬编码 `<workspaceRoot>/source-repos/<source>`，但文案承诺 legacy flat manifest 没有 `sourcesRoot` 时 clone 在 `<workspaceRoot>/<source>`。

新增问题见 inline comment：single-repo doc preview 的 `source-repos/` 修复没有区分 legacy per-chat layout。

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: latest head `d18f8322` only fixes a drift-taxonomy skill doc path; the active CLI status namespace blocker from `29283bff` is still unchanged.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `d18f8322` makes the drift-taxonomy source-discovery wording account for `sourcesRoot` vs legacy flat manifests.
- Design: optional `sourcesRoot`, source clones under `source-repos/`, client resolver/briefing/doc-preview updates, CLI status scanning, and shipped skill text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. CLI status still hides an unbound source sibling named like the tree under `sourcesRoot`. [R5 / `apps/cli/src/core/workspace.ts:292`]

`computeWorkspaceStatus()` still builds `declaredNames = new Set([manifest.tree, ...manifest.sources])` before scanning `sourcesBase`. When `sourcesRoot` is present, `sourcesBase` is `<workspace>/source-repos`, while `manifest.tree` lives at `<workspace>/<tree>` outside that directory. So an unbound bare clone at `<workspace>/source-repos/context-tree` is still skipped solely because the tree is also named `context-tree`. Please exclude `manifest.tree` only for legacy flat manifests (`sourcesRoot` absent), and add the requested regression for an unbound bare sibling named like the tree under `sourcesRoot`.

✅ 2. The new drift-taxonomy doc change is consistent with the optional-`sourcesRoot` compatibility contract and does not introduce a new issue.

**Executed**
- Submitted follow-up request_changes on head `d18f8322`.
- Did not run local tests or checks per review protocol.
- GitHub checks were still in progress when reviewed.

baixiaohang:
Current HEAD `d18f8322` status after reading yuezengwu's latest review comments:

- The earlier schema/client manifest writer issue is fixed in `29283bff`.
- The earlier seed/sync legacy-flat command issue is fixed in `29283bff`, and `d18f8322` extends the same compatibility wording to drift-taxonomy.
- GitHub checks are green.

There are still two active blockers:

1. **CLI status namespace filter** (my latest review): `computeWorkspaceStatus()` still excludes `manifest.tree` while scanning `sourcesBase`. With `sourcesRoot` present, the tree is outside `sourcesBase`, so an unbound bare sibling such as `<workspace>/source-repos/context-tree` is hidden.

2. **Legacy per-chat doc-preview path** (yuezengwu's inline comment): `documentBasePathFromRuntimeConfig()` now always returns `<sessionRoot>/source-repos/<localPath>` for a single repo. That is correct for new agent-home sessions, but `resolveSessionDocRoot()` intentionally returns `<workspaceRoot>/<chatId>` for legacy per-chat sessions, whose source checkout remains at `<workspaceRoot>/<chatId>/<localPath>`. Those legacy sessions now look in `<workspaceRoot>/<chatId>/source-repos/<localPath>` and lose single-repo markdown previews.

So the remaining fixes are: make CLI status exclude the tree name only for legacy flat manifests, and preserve the legacy per-chat doc-preview branch while using `source-repos/<name>` for new agent-home sessions.

liuchao-001:
Pushed `063d5ec3` — **valid, fixed.** Good catch, and a genuinely new one (not a stale re-flag): it's a direct consequence of `29283bff` making a `context-tree`-named source valid under `sourcesRoot`.

`computeWorkspaceStatus` built `declaredNames = {tree, ...sources}` and used it to filter the unbound-sibling scan over `sourcesBase`. Once the tree lives at `<ws>/<tree>` (outside `sourcesBase`), excluding `manifest.tree` from that scan wrongly hides an unbound clone at `<ws>/source-repos/<tree-name>` — exactly the now-schema-valid case. Fixed: `manifest.tree` is excluded **only in the legacy flat layout** (`sourcesRoot` absent, where tree + sources share the root namespace); with `sourcesRoot` set the scan filters on `sources` alone. Regression added: `sourcesRoot` set + an unbound bare sibling literally named like the tree → it's listed.

I also swept the other consequences of the `tree ∈ sources` relaxation to confirm none are left: `ensureWorkspaceManifest` (writes it — 29283bff), `treePath` (always `<ws>/<tree>`, unchanged), and `pickImmediateWorkspaceSources` (flat-only migrate flow, correctly still excludes the tree name).

Current HEAD `063d5ec3`. `pnpm typecheck` 12/12, `pnpm check` clean, full suite green (cli workspace 29 incl. the new regression). Findings #2/#3 from this round you've confirmed fixed. Re-review welcome. @baixiaohang

baixiaohang (changes requested):
## Review by codex-assistant - PR #1063

- Reason: latest head `063d5ec3` fixes the CLI status namespace issue, but the legacy per-chat doc-preview regression remains.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest change: `063d5ec3` excludes the tree name from unbound source scanning only in legacy flat workspaces and adds the requested `source-repos/<tree>` regression.
- Design: optional `sourcesRoot`, source clones under `source-repos/`, client resolver/briefing/doc-preview updates, CLI status scanning, and shipped skill text.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
❌ 1. Legacy per-chat single-repo doc preview still resolves under `source-repos/`. [R5 / `packages/client/src/runtime/session-manager.ts:172`]

`documentBasePathFromRuntimeConfig()` still returns `join(sessionRoot, SOURCE_REPOS_DIRNAME, localPath)` for every single-repo session. That is correct when `sessionRoot` is the new per-agent home, but `resolveSessionDocRoot()` intentionally returns `<workspaceRoot>/<chatId>` for legacy pre-#506 per-chat sessions. Those legacy sessions keep their source checkout at `<workspaceRoot>/<chatId>/<localPath>`, not `<workspaceRoot>/<chatId>/source-repos/<localPath>`, so relative markdown refs lose their single-repo preview path. Please preserve the legacy branch, e.g. have the session-root resolver surface whether it selected a legacy per-chat dir, then use `<sessionRoot>/<localPath>` for legacy roots and `<agentHome>/source-repos/<localPath>` for new agent-home roots. Add a regression with an existing legacy chat dir plus a single repo.

✅ 2. The previous CLI status namespace issue is fixed: unbound `source-repos/<tree>` siblings are no longer filtered in the `sourcesRoot` layout, and the regression covers a bare clone named like the tree.

**Executed**
- Submitted request_changes on head `063d5ec3`.
- Did not run local tests or checks per review protocol.
- GitHub checks were green when reviewed.

liuchao-001:
Pushed `c9c5dd50` addressing the legacy per-chat doc-preview regression — **valid, fixed.** Good catch (and a real one, not a stale re-flag): it's anchored to `063d5ec3`, the then-current head.

You're right: my doc-preview fix prepended `source-repos/` unconditionally, but `resolveSessionDocRoot` returns `<workspaceRoot>/<chatId>` for a legacy pre-#506 chat, so the base became `<ws>/<chatId>/source-repos/<name>` — a nonexistent dir → single-repo preview regressed to no snapshot. `documentBasePathFromRuntimeConfig` and `selfFenceFromRuntimeConfig` now take `workspaceRoot` and apply the `source-repos/` layer **only when `sessionRoot === workspaceRoot`** (new agent-home session); a legacy per-chat session keeps its prior flat base `<sessionRoot>/<localPath>` (and flat `singleRepoLocalPath`). Threaded `workspaceRoot` through both call sites. Regressions added for the legacy per-chat layout in both helpers.

The other two findings in this review round were already fixed on heads after the one you reviewed:
- **schema/manifest `context-tree` source** → `29283bff` (refine legacy-flat-only + writer no longer drops the manifest).
- **seed/sync hard-coded `source-repos/` commands** → `29283bff` (+ `d18f8322` for the `drift-taxonomy` `ls -d`). Both now resolve `<source-clone>`/`<sourcesRoot>/*` from the manifest.

Current HEAD `c9c5dd50`. `pnpm typecheck` 12/12, `pnpm check` clean, full suite green (client 975). Re-review of `c9c5dd50` welcome.

baixiaohang (approved):
## Review by codex-assistant - PR #1063

**Recommendation**: approve
- Reason: the latest head fixes the remaining legacy per-chat doc-preview regression, and the earlier source-repos layout compatibility blockers remain addressed.

**Level**: A
- Semantic lift: none

**PR Summary**
- Solves: moves agent-managed source clones from `<workspace>/<name>` to `<workspace>/source-repos/<name>` while preserving optional-`sourcesRoot` legacy-flat manifest compatibility.
- Design: shared manifest schema gains optional `sourcesRoot`; client resolver/manifest writer/briefing/doc-preview route source repos through `source-repos/` for new layouts; CLI status scans the correct source namespace; shipped skills resolve source clone paths from the manifest.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
✅ 1. The legacy per-chat single-repo doc-preview regression is fixed: `documentBasePathFromRuntimeConfig()` and `selfFenceFromRuntimeConfig()` now apply `source-repos/` only for new agent-home sessions, while legacy per-chat roots keep `<sessionRoot>/<localPath>`.

✅ 2. The CLI status namespace issue remains fixed: bare source clones are detected, and unbound `source-repos/<tree>` siblings are no longer filtered in the `sourcesRoot` layout.

✅ 3. The schema/client manifest writer and skill command compatibility gaps remain fixed: `tree ∈ sources` is legacy-flat-only invalid, and seed/sync/drift source reads derive `<source-clone>` from `sourcesRoot` when present with flat fallback when absent.

⚠️ Non-blocking: `packages/client/src/handlers/codex.ts` still has one stale comment saying source repos live at the agent-home top level; the code path is correct via `resolveGitRepoTargetPath()`, so this is documentation cleanup only.

**Executed**
- Submitted approve on head `c9c5dd50` with baixiaohang authorization.
- Did not run local tests or checks per review protocol.
- GitHub checks were green when approved.

baixiaohang (approved):
## Review by codex-assistant - PR #1063

**Recommendation**: approve
- Reason: the new head only cleans up stale Codex-handler comments to match the already-reviewed `source-repos/` layout; no behavioral changes or reopened blockers.

**Level**: A
- Semantic lift: none

**PR Summary**
- Latest head: `272dbafe` updates `packages/client/src/handlers/codex.ts` comments from the old top-level source-repo wording to `<workspaceCwd>/source-repos/<localPath>`.
- The previously reviewed design remains unchanged: optional `sourcesRoot`, source clones under `source-repos/`, CLI status namespace handling, and legacy-flat compatibility in the shipped skills.

**Review Findings**
✅ 1. The new commit is documentation-only in a client handler comment and aligns with the actual `resolveGitRepoTargetPath()` behavior.

✅ 2. The previous blockers remain fixed: legacy per-chat doc preview, bare clone status detection, `tree ∈ sources` namespace handling, seed/sync legacy-flat commands, and `source-repos/<tree>` CLI status scanning.

**Executed**
- Submitted approve on head `272dbafe` with baixiaohang authorization.
- Did not run local tests or checks per review protocol.
- GitHub checks were green when approved.

yuezengwu (approved):
## Review by codex-assistant - PR #1063

**Recommendation**: approve
- Reason: the latest head fixes the remaining compatibility gaps I had requested changes on, and I did not find a new blocking issue in the updated diff.

**Level**: A
- Semantic lift: core workspace manifest/runtime layout contract, already called out for owner review.

**PR Summary**
- Latest head: `c9c5dd50`.
- Design: optional `sourcesRoot`, source clones under `<workspace>/source-repos/<name>`, legacy flat manifests remain readable, runtime declares repo layout while the agent owns git lifecycle.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, shipped First Tree skills.

**Review Findings**
✅ 1. The new-layout `context-tree` source-name gap is fixed. `tree ∉ sources` is now legacy-flat-only, and `ensureWorkspaceManifest()` writes a manifest with `sourcesRoot` even when a source is named `context-tree`.

✅ 2. The seed/sync legacy-flat skill command gap is fixed. The executable command blocks now resolve a `<source-clone>` from `manifest.sourcesRoot`, with explicit branches for current `source-repos/` manifests and legacy flat manifests.

✅ 3. The single-repo doc-preview regression is fixed without breaking legacy per-chat sessions. `documentBasePathFromRuntimeConfig()` and `selfFenceFromRuntimeConfig()` now apply `source-repos/<name>` only when `sessionRoot === workspaceRoot`, and preserve flat `<legacyChatRoot>/<name>` for pre-#506 per-chat roots.

✅ 4. The CLI status follow-up is consistent: unbound source clones named like the tree are no longer filtered out under `sourcesRoot`, while legacy flat status still excludes the tree name.

**Executed**
- Reviewed local diff at head `c9c5dd50`.
- Ran `git diff --check origin/main...HEAD` (clean).
- Did not run local tests because this review worktree has no `node_modules`; I did not reinstall dependencies during review.

PR #1064, opened 2026-06-15
fix(server): stop dropping followed chats from GitHub event audience

liuchao-001:
## Summary
Fixes the "followed chat never receives GitHub events" bug — a refactor chat that `github follow`-ed a PR never got woken by reviewers' comments. Two **decoupled** fixes in the webhook → chat audience/delivery path:

- **M2 — audience dedup by `(human, chat)`, not `human`** (`github-audience.ts`). `resolveAudience` collapsed subscribed rows by `humanAgentId` alone, assuming all of a human's rows share one chat. When the same human is bound to an entity from two chats (a webhook `human_fallback` row in one chat + an explicit `follow` row in another, under different delegates), the dedup kept only the earliest chat and **silently dropped every other followed chat** from the audience. This is the confirmed root cause of the reported bug.
- **B — echo suppression moved to the notification layer, decoupled from delegate** (reuses unmerged `origin/fix/issue-942-audience-echo-suppression`). The old `kind:"agent"` filter dropped whole audience rows whose human/delegate matched the actor; when that was a chat's only routing row, the chat lost the event entirely. It also relied on `actor.login` identifying a single acting agent — unsound, since agents act on GitHub under a human's GitHub identity. Now every target is annotated with `actorAgentId` and `deliverNormalizedEvent` excludes the actor from `addressedToAgentIds`: the card always lands (public record), the actor is never woken by its own action, everyone else is notified normally.

**Safety invariant** established by this change: *echo & dedup may never drop a followed chat or a card; echo's strongest action is removing the acting agent from one card's wake-set.*

## Follow-ups (separate PRs)
- **M3** — delivery must fall back to the chat's live speakers instead of silently recipientless when the stored delegate is not a live speaker (renegotiates the `e20f0dec` recipientless test).
- **D** (optional) — de-fabricate the `(human, delegate)` pair; the representative-human becomes cosmetic only.

## Test plan
- [x] `pnpm --filter @first-tree/server typecheck`
- [x] `biome check` on changed files
- [x] `vitest run github-audience github-delivery` — all pass (32 + new)
- [x] New tests: M2 cross-chat non-collapse; echo annotates instead of dropping (audience); echo excludes actor from addressing while card still lands (delivery)
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1067, opened 2026-06-15
test(server): combined M2+echo regression for the multi-human-not-pushed bug

liuchao-001:
Follow-up to #1064 (code-reviewer's suggestion ①): one fixture that ignites **both** old defects at once, locking the exact regression #1064 fixed.

## The fixture
One human **H** (`agents.name == actor login`), the same entity bound from **two chats** — chatA (`bound_at` earlier, delegate Da) and chatB (`bound_at` later, delegate Db) — and the event **actor = H**. Da/Db/H distinct; Da speaks in chatA, Db in chatB.

On the **old** code both mechanisms stacked:
- old M2 (dedup by `humanAgentId`, keep earliest) dropped **chatB**;
- old echo (actor on the human side → drop the whole row) dropped **chatA** too;
- → empty audience → both chats silent (the real `1d2f3012` shape).

## What it asserts (delivery/inbox seam, not just audience)
Drives `resolveAudience` → `deliverNormalizedEvent` and checks:
1. audience keeps **both** chats, each annotated `actorAgentId = H`;
2. both chats get a card;
3. **Da is woken in chatA, Db in chatB** (inbox `notify=true`) — the swallowed chat's delegate (≠ actor) actually wakes;
4. the actor H is never self-woken.

A pure audience-layer assertion only proves the row survives, not the wake — so the check runs at delivery/inbox.

Test-only; `typecheck` / `biome` / `vitest github-delivery` (11 passed) green locally. Durable constraint already recorded in first-tree-context#523 (S7).

### DIFF
--- a/packages/server/src/__tests__/github-delivery.test.ts
+++ b/packages/server/src/__tests__/github-delivery.test.ts
@@ -8,7 +8,7 @@ import { chats } from "../db/schema/chats.js";
 import { githubEntityChatMappings } from "../db/schema/github-entity-chat-mappings.js";
 import { inboxEntries } from "../db/schema/inbox-entries.js";
 import { messages } from "../db/schema/messages.js";
-import type { AudienceTarget } from "../services/github-audience.js";
+import { type AudienceTarget, resolveAudience } from "../services/github-audience.js";
 import { deliverNormalizedEvent } from "../services/github-delivery.js";
 import { createTestAdmin, useTestApp } from "./helpers.js";
 
@@ -54,6 +54,7 @@ function makeEvent(opts: {
   rawEventType?: string;
   rawAction?: string;
   title?: string;
+  actorLogin?: string;
 }): NormalizedEvent {
   return {
     source: { kind: "github-app-installation", installationId: 1, organizationId: opts.orgId },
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed as a test-only regression fixture for the combined M2 dedup + actor-echo failure fixed by #1064.

Core change: the new fixture creates one represented human actor bound to the same PR entity through two chats/delegates, then drives `resolveAudience` into `deliverNormalizedEvent` and asserts both chat cards plus `notify=true` inbox rows for the correct delegates while the actor is not self-woken.

No blocking findings.

Verified locally on the PR worktree:
- `pnpm --filter @first-tree/server test -- github-delivery` (ran server suite: 161 files / 1422 tests passed)
- `pnpm typecheck`
- `pnpm check` (exit 0; existing warnings outside this PR's changed file)

baixiaohang (approved):
## Review by codex-assistant - PR #1067

**Recommendation**: approve
- Reason: The test-only fixture exercises the combined M2 cross-chat dedup + actor-echo regression through delivery and inbox notify semantics.

**Level**: B-low
- Path initial: `packages/server/**` test-only change; no migrations, Prisma schema, or core service files changed.
- Semantic lift: none.

**PR Summary**
- Problem: Locks the #1064 failure shape where human-scoped dedup and actor echo together could leave all followed chats silent.
- Design: Adds a two-chat, one-represented-human fixture that runs `resolveAudience` then `deliverNormalizedEvent`, asserting both cards land, each delegate is notified, and the actor is not notified.
- Affected module: `packages/server/src/__tests__/github-delivery.test.ts`

**Review Notes**
✅ 1. The fixture matches the durable S2/S7 constraints: audience dedup keeps `(human, chat)` rows, echo is notification-layer only, and followed chats still get cards.
✅ 2. The delivery assertions check the actual wake boundary via `inbox_entries.notify=true`, not just audience membership.

**Executed**
- Approved.

PR #1048, opened 2026-06-12
refactor(client): agent-managed source repos & context tree

liuchao-001:
## Summary
- **The client runtime performs no git network or write operations on agent workspaces anymore** — only a single read-only `git status` survives, for tree-write telemetry (`runtime/context-tree-git-status.ts`). Binding coordinates resolve via `resolveAgentContextTreeBinding` (pure config) and ship to the agent in the briefing; the agent itself clones, refreshes, and cleans every repo.
- **Deleted runtime git management** (−6.8k lines): `git-mirror-manager.ts`, the live-use registry / `prepareSourceRepos` git logic in `source-repos.ts` (now pure declaration), `source-repo-cleanup.ts`, `worktree-cleanup.ts`, handler prepare/cleanup paths, `sweepLegacyMirrors`, the `GitMirror*` error-taxonomy branches, and the clone-deleting workspace migrations.
- **Context Tree: per-agent clone** at `<agentHome>/context-tree` replaces the shared `<dataDir>/context-tree-repos/` pool. Zero-deletion migration: legacy pool and `context-tree` symlinks are left untouched (reads keep working); the agent replaces the symlink (link only) with a real clone per its briefing.
- **Briefing protocol rewritten** (`agent-briefing.ts`): source repos are bare/fetch-only (`git clone --bare` on first use; never read/write the clone path; all work in per-task worktrees frozen for the PR's life; agent removes worktrees at task close, e.g. PR merge); context tree is clone-if-missing + `git pull --ff-only` **before every tree read** (fresher than any session-edge hook — busy sessions inject for hours without a resume); credential failures must be reported to a human in chat instead of degrading silently in client.log. Every interpolated path / URL / branch is shell-quoted so values with spaces or `$` render correctly.
- **Tree-HEAD drift re-bootstrap retired** — vestigial since skills became bundled (no workspace artifact derives from tree content); CLI-version drift remains the only content key.
- Fixes the standing staleness bugs: permanent `skipped-in-use` short-circuit (release hook on shutdown that never fired), context tree pulled only at client boot, and silent credential failures.

## Behavior verification (PR #1048 review pass — code-reviewer)
- ✅ S3 zero-deletion: legacy `context-tree` symlinks + pre-existing worktrees survive `start()`.
- ✅ S4 architectural git-free: only one `git status` (read-only, no credentials) left in runtime.
[...]

PR #1070, opened 2026-06-15
refactor(client): drop dead managed.json sourceRepos field

liuchao-001:
## Summary

`managed.json` (`.first-tree-workspace/managed.json`) carries a `sourceRepos` array that has been a **no-op since the agent-managed-repos refactor (#1048)**. Under the agent-managed model the runtime declares + observes but never clones / fetches / reconciles / deletes a source-repo clone, so no code writes or reads `ManagedState.sourceRepos` anymore. Only the `skills` array is still live (skill-install reconcile).

This removes the dead field and its references:

- `ManagedState` type, `readManagedState`, and the `updateManagedState` default no longer carry `sourceRepos`.
- `docs/development/agent-workspace-state.md` schema doc updated.
- Test fixtures in `managed-state.test.ts`, `skills-state-reconcile.test.ts`, `workspace-migrations.test.ts` updated.

**Read-compatible, no schema bump:** `schemaVersion` stays `1` and `readManagedState` already ignores unknown keys, so an existing `managed.json` still carrying `sourceRepos` reads fine — the stale key is simply dropped on the next write. (Bumping to `2` would make readers reject every existing v1 record as "first run" and lose `skills` tracking for a cycle — not worth it for a read-compatible change.)

**Drive-by test fix:** the `workspace-migrations` "UUID dir still in source-repo config" test was passing only via *deferral* (it called `applyPendingMigrations` with no ctx), not via config-based sparing — its fixture wrote the now-removed `sourceRepos` key but the migration spares on `ctx.currentSourceRepoNames`. It now passes a resolved `currentSourceRepoNames` set so it exercises the real spare path.

Not touched: the unrelated `sourceRepos` / `sourceReposForPrompt` fields in the briefing (`agent-briefing.ts`), handlers, and web onboarding — those are the live `PredeclaredSourceRepo` coordinates, a different concept.

Context: companion to the agent-managed-repos model recorded in first-tree-context#506. No tree change needed — this is implementation cleanup consistent with the already-recorded decision; no tree node references the field.

## Test plan

- [x] `pnpm --filter @first-tree/client typecheck` — clean
- [x] `vitest run managed-state workspace-migrations skills-state-reconcile` — 40 passed
- [x] `biome check` on changed files — exit 0 (pre-existing `noConfusingVoidType` warning at `workspace-migrations.ts:90` is on `main`, untouched)
- [ ] CI green

### DIFF
--- a/docs/development/agent-workspace-state.md
+++ b/docs/development/agent-workspace-state.md
@@ -27,7 +27,7 @@ sit on disk forever. The state files below are that record.
 
 | Path (relative to workspace root) | Owner | What it is |
 | --- | --- | --- |
-| `.first-tree-workspace/managed.json` | `runtime/managed-state.ts` | Schema-versioned record of the CLI-managed resources currently materialised in this workspace. The live array is `skills` (skill names), diffed on every session start to discover removals. `sourceRepos` is a legacy field — no current code writes or reconciles it (source repos are agent-managed). |
+| `.first-tree-workspace/managed.json` | `runtime/managed-state.ts` | Schema-versioned record of the CLI-managed resources currently materialised in this workspace. Holds one array — `skills` (skill names) — diffed on every session start to discover removals. (Source repos and the Context Tree clone are agent-managed; the runtime never deletes a repo clone, so they carry no entry here. A legacy record that still has the retired `sourceRepos` key is read-compatible — the reader ignores unknown keys.) |
 | `.first-tree-workspace/migrations-applied.json` | `runtime/workspace-migrations.ts` | Set of one-shot directory-structure migration ids that have already run in this workspace. Each migration runs at most once even if it's later removed from the registry; the marker stays as forward protection. |
 
 Both files use atomic writes (temp + rename) so a crashed writer never
@@ -40,13 +40,10 @@ leaves a half-formed JSON record on disk.
   "schemaVersion": 1,
   "cliVersion": "0.3.2",
   "updatedAt": "2026-06-08T15:50:00.000Z",
-  "sourceRepos": ["first-tree", "first-tree-context"],
   "skills": ["first-tree", "first-tree-context", "first-tree-read", "first-tree-sync", "first-tree-seed"]
 }
 ```
 
-- **`sourceRepos`** — the `localPath` of each repo from `payload.gitRepos`
-  the CLI cloned at workspace top level.
 - **`skills`** — the names of skills installed under
   `<workspace>/.agents/skills/<name>/` (with matching `.claude/skills/<name>`
   symlinks).
--- a/packages/client/src/__tests__/managed-state.test.ts
+++ b/packages/client/src/__tests__/managed-state.test.ts
@@ -31,7 +31,6 @@ describe("managed-state", () => {
       schemaVersion: 1,
       cliVersion: "0.3.2",
       updatedAt: "2026-06-08T16:00:00.000Z",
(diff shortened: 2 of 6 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1070

**Recommendation**: approve
- Reason: This removes the dead `managed.json::sourceRepos` state while preserving the live `skills` reconcile path and v1 read compatibility for legacy records.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: After the agent-managed repos refactor, `.first-tree-workspace/managed.json::sourceRepos` is no longer written, read, reconciled, or used for clone deletion.
- Design: `ManagedState`, its reader, and update defaults now track only `skills`; docs and fixtures drop `sourceRepos`; the workspace migration UUID-source-repo test now passes resolved `currentSourceRepoNames` so it covers the real config-based spare path.
- Affected modules: `packages/client/src/runtime/managed-state.ts`, `packages/client/src/runtime/workspace-migrations.ts`, related tests, and `docs/development/agent-workspace-state.md`.

**Review Notes**
✅ 1. `readManagedState` still gates only on `schemaVersion === 1` and defensively reads `skills`; a legacy `sourceRepos` key is ignored as an unknown key and naturally dropped on the next write.
✅ 2. `installFirstTreeSkills` still diffs `prev.skills` against `TREE_SKILL_NAMES`, so removing `sourceRepos` does not affect live skill reconciliation.
✅ 3. The workspace migration test now uses a non-null `currentSourceRepoNames`, so it exercises the resolved-config spare path rather than passing by unresolved-context deferral.
✅ 4. The live briefing/handler/onboarding `sourceRepos` / `sourceReposForPrompt` coordinates are untouched, and CI is green. I did not run local tests per review-mode policy.

**Executed**
- Approved.

yuezengwu (approved):
实现目标 / 核心改动：移除 `.first-tree-workspace/managed.json` 里已经失效的 `ManagedState.sourceRepos` 字段，只保留仍在使用的 `skills` 状态；`schemaVersion` 继续保持 1，旧 v1 记录里的 `sourceRepos` 会作为 unknown key 被读端忽略，并在下一次写回时自然丢掉。相关 docs 和测试 fixture 已同步，`workspace-migrations` 里 UUID source-repo spare path 也改为通过 live `currentSourceRepoNames` 覆盖真实路径。

我对照了 context tree 里的 agent-managed repo model：source repo clone lifecycle 现在确实属于 agent 协议，runtime 只声明 / 观察，不再 clone/fetch/reconcile/delete source repo clone；这个 cleanup 与已记录模型一致，不需要新的 tree 更新。

No blockers found.

验证：
- CI on PR head: success
- `pnpm --filter @first-tree/client typecheck`: pass
- `pnpm exec vitest run packages/client/src/__tests__/managed-state.test.ts packages/client/src/__tests__/workspace-migrations.test.ts packages/client/src/__tests__/skills-state-reconcile.test.ts`: 40 passed after building `@first-tree/shared` in the fresh review worktree
- `pnpm exec biome check ...changed files...`: exit 0, only the pre-existing `noConfusingVoidType` warning at `workspace-migrations.ts:90`

Non-blocking: because the PR contract explicitly relies on legacy v1 records containing `sourceRepos` remaining read-compatible, it would be useful to add a small regression test that starts from such a legacy JSON record and asserts `updateManagedState` rewrites it without `sourceRepos`. Current code already behaves that way.

PR #1073, opened 2026-06-15
fix(client): enforce terminal inbox delivery ack

baixiaohang:
## Summary

- Introduce explicit inbox delivery custody/terminal semantics in the client runtime: entries move through `open -> owned -> terminal`, ACK only commits a contiguous terminal prefix, and ACK/recovery debt no longer projects as provider busy.
- Route built-in handlers through `DeliveryToken` receipts/outcomes so start/resume/inject either return custody or reject/throw, and post-custody async cleanup/failure paths settle through complete/retry/terminal rejection guards.
- Keep pre-provider failures recoverable unless durable terminal evidence exists; Claude/Codex/TUI formatting or conversion failures retry instead of ACKing, including mixed queued batches.
- Add `session:resume` command/API/UI wiring and manual pause guard semantics so old recovery redelivery cannot clear pause.
- Preserve latest TUI `TuiLifecycleFence` suspend-during-resume behavior while merging queue/retry token handling.

## P0/P1 Scope Covered

- admitted-but-dropped and handler silent-drop windows
- terminal-only ACK boundary and active-ledger same-`entryId` redelivery dedup
- pre-provider deterministic failure without durable evidence does not ACK
- false working/runtime projection and scheduler/idle gates detached from delivery debt
- manual pause bounce protection with explicit resume
- TUI dying/suspend/queued cleanup paths retry non-terminal queued work

## Residual Risks / Out of Scope

The following P2 items are intentionally not solved in this PR:

- final output idempotency
- ACK attempt/receipt fencing
- long-turn progress lease
- poison/backoff policy
- cross-process / persistent `entryId` dedup after ledger eviction or restart

`AgentHandler` still has transitional compatibility surfaces (`token?`, legacy string start/resume returns, and `SessionContext.finishTurn/retryTurn` shims). Built-in production routes now pass explicit tokens; the compatibility surface is not the recommended protocol for new handlers.

## Validation

- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/session-runtime-state.test.ts src/__tests__/session-manager.test.ts src/__tests__/session-manager-edge-coverage.test.ts src/__tests__/session-manager-retry.test.ts src/__tests__/session-start-error-signaling.test.ts src/__tests__/session-state-notifications.test.ts src/__tests__/claude-code-inject-startup-race.test.ts src/__tests__/claude-code-stream-error-retry-replay.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/codex-inject-startup-race.test.ts src/__tests__/codex-retry-abort.test.ts src/__tests__/codex-usage-limit.test.ts src/__tests__/agent-cwd-redesign.e2e.test.ts src/__tests__/tui-transcript-tail.test.ts src/__tests__/tui-claude-command.test.ts src/__tests__/tui-tmux-session.test.ts src/__tests__/tui-turn-disposition.test.ts src/__tests__/tui-lifecycle-fence.test.ts src/__tests__/tui-suspend-queued-recovery.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/admin-sessions-suspend-terminate.test.ts`
- `pnpm --filter @first-tree/web exec vitest run --maxWorkers=2 src/api/__tests__/api-wrappers.test.ts src/components/chat/__tests__/agent-status-panel.test.ts`
- `pnpm --filter first-tree-dev exec vitest run --maxWorkers=2 src/__tests__/cli-command-matrix-extra.test.ts`
- `pnpm check` (passes with existing warnings/info in unrelated files: `prompt-tab.tsx`, `agent-shared-helpers-extra.test.ts`, `workspace-migrations.ts`)
- `git diff --check`

Attempted TUI e2e:

- `E2E_TUI=1 E2E_WITH_CLIENT=1 pnpm --filter @first-tree/e2e exec vitest run --config vitest.tui.config.ts --maxWorkers=2 ./src/tests/tui-runtime-suspend-race.e2e.test.ts`
- Local run did not reach the handler/fence assertion: the e2e daemon registered with `no agents configured yet` / `agent pinned (...) but no agents dir set -- cannot auto-register`, so the first fixture message timed out waiting for a reply. Focused handler/runtime tests above cover the intended TUI semantics.


packages/client/src/runtime/session-manager.ts
```diff
@@ -518,17 +564,32 @@ export class SessionManager {
...
       const session = this.sessions.get(chatId);
       if (session?.status === "active") {
         this.config.log.info({ chatId }, "suspend command received");
         this.suspendSession(session);
       }
+      this.projectSessionRuntime(chatId);
+      return;
+    }
+
+    if (command === "session:resume") {
```
yuezengwu, line 612: This message-less resume is routed as `deliveryKind: "fresh"`. If concurrency is full, `acquireActiveSlot` treats fresh work as external input and can preempt an unrelated working session just to resume a paused handler with no new user message. It also uses a synthetic empty `SessionMessage` for slot acquisition, which can be queued and later drained as a real empty provider input. Message-less resume needs a non-preempting/internal slot path that preserves `undefined` through any queue instead of treating the command like fresh user work.

packages/server/src/api/sessions.ts
```diff
@@ -66,6 +66,24 @@ export async function sessionRoutes(app: FastifyInstance): Promise<void> {
...
+  app.post<{ Params: { uuid: string; chatId: string } }>("/:uuid/sessions/:chatId/resume", async (request, reply) => {
+    const { agent, scope } = await requireAgentAccess(request, app.db, "manage");
+    const result = await sessionService.resumeSession(
+      app.db,
+      agent.uuid,
+      request.params.chatId,
+      scope.organizationId,
+      app.notifier,
+    );
+    if (result.transitioned) sendToAgent(agent.uuid, { type: "session:resume", chatId: request.params.chatId });
```
yuezengwu: `sendToAgent` returns `false` when the agent has no live socket, but the route has already transitioned the row from `suspended` to `active` and ignores that return value. For `resume`, `active` means the client actually accepted a live command; otherwise an offline suspended session reports 200/active while the client never clears its local pause guard or resumes. Please gate this transition on live delivery, return a conflict when the command cannot be delivered, or otherwise keep the server projection suspended until the client reports back.

yuezengwu (changes requested):
Thanks for the detailed terminal-ACK work. The core custody/terminal ledger direction matches the inbox-delivery contract, but I found two resume-path regressions that should be fixed before merge: message-less resume can preempt unrelated working turns, and the server can mark a disconnected suspended session active even when the resume command was not delivered.

packages/client/src/runtime/session-manager.ts
```diff
@@ -1593,7 +1804,7 @@ export class SessionManager {
       },
       forwardResult,
       markMessagesConsumed: (messages) => {
-        this.inboxDelivery.markConsumed(chatId, messages);
+        this.inboxDelivery.markProcessingStarted(chatId, messages);
       },
       finishTurn: (messages, outcome) => {
```
yuezengwu, line 1864: This slot gate still blocks already-routable same-chat work after an explicit message-less resume. Repro: with `concurrency = 1`, suspend a chat, let the manual pause guard buffer a delivery in `pendingQueue`, then issue `session:resume`. `resumeSession(..., undefined)` makes that same chat active and increments `_activeCount`; the follow-up `drainPendingQueue()` reaches this check, sees `_activeCount >= concurrency`, finds no eligible *other* idle session because the queued item is for the same chat, and returns before `routeMessage` can inject the real buffered message into the now-active handler. The inbox row stays queued/unacked even though it no longer needs a new slot. Please special-case an already-active `next.chatId` (or drain same-chat buffered deliveries before/with the control resume) so explicit resume actually releases manual-pause buffered work.

yuezengwu (changes requested):
The synchronize fixes the two earlier blockers in the right direction: message-less resume no longer uses a synthetic empty input / fresh preemption, and the server no longer marks a disconnected suspended row active. I found one remaining resume/manual-pause queue regression that should be fixed before merge: buffered same-chat deliveries can stay queued after explicit resume at the concurrency limit.

yuezengwu (approved):
Latest update addresses the remaining resume/manual-pause queue blocker: `drainPendingQueue` now routes same-chat work when that session is already active, before applying the slot gate, and the new regression test covers the explicit resume + buffered delivery + concurrency=1 case. I did not find further P0/P1 issues in the reviewed delta.

PR #1075, opened 2026-06-15
feat(onboarding): idempotent server-side kickoff endpoint

Gandy2025:
## Why

The onboarding finale was **browser-orchestrated**: the client sequentially created the first chat → sent the bootstrap message → stamped completion. A closed tab, dropped network, or one-step-fails-mid-sequence left **half-completed state** — orphan chats, duplicate bootstraps, or a completion stamp decoupled from whether the chat actually existed. The `/build-tree` recovery surface just re-ran the same fragile sequence.

This is the "idempotency subset" of the broader kickoff-reliability discussion — it makes the existing recovery path actually safe to re-run, **without** a full saga (no job queue / progress table / auto-resume).

## What

Fold the three fragile steps into one idempotent, resumable endpoint **`POST /me/onboarding/kickoff`**:

1. **find-or-create** the kickoff chat, keyed by `<human>:<agent>:<kind>` via a unique index on `chats.onboarding_kickoff_key` + `ON CONFLICT DO NOTHING` (race-safe; `NULL` for every non-kickoff chat, so ordinary chats never collide — mirrors the `github_app_installations` 1:1 pattern).
2. **send-if-empty** the bootstrap, under a `FOR UPDATE` row lock so two concurrent requests can't both send.
3. **stamp completion** (audit + suppressor) only after the chat exists, idempotently (`WHERE completed_at IS NULL`).

`kind` (`intro` | `tree`) keeps an "meet your agent" intro chat distinct from a later `/build-tree` kickoff for the same agent — otherwise the coarse key would resolve build-tree to the intro chat and silently skip the tree-seeding bootstrap.

Web `runKickoff` collapses `createChat + sendMessage + complete` into one call; the client completion stamp is now an **optimistic-UI flip only** (the server is the durable authority).

### Housekeeping in the same area
- **Drop** the dead `users.onboarding_dismissed_at` / `onboarding_completed_at` columns — per-membership state on `members` supersedes them and nothing has read them since migration 0062.
- **Add** a `members` CHECK that completion implies suppression (`completed_at IS NULL OR suppressed_at IS NOT NULL`), making the invariant a schema fact. Migration **0064** backfills any legacy rows before adding the constraint.

## Testing
- New `me-onboarding-kickoff.test.ts`: create+send+stamp; idempotent re-call (1 chat / 1 bootstrap / stamp preserved); concurrent → 1 chat + 1 bootstrap; **intro vs tree stay separate chats so `/build-tree` still wakes the agent** (regression).
- Full suite green: server **1426**, web **801**, typecheck + biome clean.

## Independent review
codex `review --uncommitted` (high): round 1 = FAIL — caught a P1 (over-broad key broke `/build-tree`) + a P2 (send-if-empty race). Both fixed + regression-tested. Round 2 = **PASS** ("no actionable correctness issues").


Gandy2025:
Context Tree PR (cross-link): agent-team-foundation/first-tree-context#526 — reflects the kickoff finalization decision + per-membership state-model update into `system/cloud/onboarding.md`.

baixiaohang (approved):
## Review by codex-assistant - PR #1075

**Recommendation**: approve
- Reason: Static review found no blocking correctness issues in the server-side idempotent kickoff path, membership-scoped onboarding state, or message delivery contract. Approved after explicit authorization from baixiaohang.

**Level**: B-high
- Path initial: `packages/server/**` includes migrations/schema and changes `packages/server/src/services/chat.ts`.
- Semantic lift: none beyond the path-based B-high classification.

**PR Summary**
- Problem: Browser-orchestrated onboarding kickoff could leave half-completed state across chat creation, bootstrap send, and completion stamping.
- Design: Add `POST /me/onboarding/kickoff` to create-or-reuse the kickoff chat, send the bootstrap only if the chat is empty under a row lock, then stamp membership completion after the chat exists. `kind` separates intro and tree-building kickoffs. Migration 0064 adds the kickoff key, removes dead user-level onboarding stamps, and enforces completion => suppression on `members`.
- Affected modules: server drizzle/schema/api/services, shared `/me` schemas, web onboarding kickoff flow.

**Review Notes**
✅ Server routing resolves the active membership/human agent first, then relies on the existing `createChat` same-org/active/private-target gates.
✅ Idempotency coverage includes repeated calls, concurrent create/send, and intro-vs-tree separation for `/build-tree` recovery.
✅ Migration backfills legacy completed-but-unsuppressed rows before adding the CHECK, and I did not find remaining reads of the dropped user columns.
⚠️ Non-blocking residual: `kickoffOnboarding` reuses `sendMessage(tx)` inside an outer transaction, while `sendMessage` has a best-effort chat-message kick intended for after commit. In this path the durable inbox notify still happens after the route transaction returns, and the kickoff UI navigates/refetches, so I do not think it should block this PR. If tx-scoped sends become a general pattern, consider splitting a core writer that has no external kick side effect.

**Verification**
- I did not run local tests, per review mode.
- GitHub CI was green at review time: Migrations, Lint & Type Check, Test, CodeQL.

yuezengwu (approved):
## Review by yzw-codex - PR #1075

Recommendation: approve

Implementation goal and core change: this moves onboarding kickoff from a browser-orchestrated create-chat -> send-bootstrap -> stamp-completion sequence into `POST /me/onboarding/kickoff`. The server now find-or-creates the kickoff chat with a kind-scoped key, sends the bootstrap only if the chat is empty under a row lock, then stamps per-membership completion/suppression after the chat exists. Web kickoff now calls that endpoint and treats the old completion call as optimistic UI reconciliation.

Findings: no blocking correctness issues found.

Notes for human review: this PR contains core DB/state-model changes: `chats.onboarding_kickoff_key` plus a unique index, removal of `users.onboarding_dismissed_at` / `users.onboarding_completed_at`, and a `members` CHECK enforcing completed => suppressed. I checked runtime references and did not find remaining reads of the dropped user columns; the remaining web names are compatibility state sourced from membership data.

Review checks:
- Read the Context Tree onboarding/messaging/inbox/team constraints and checked the diff against them.
- Reviewed the server kickoff service, chat create conflict path, route notification behavior, migration, shared schemas, and web kickoff flow.
- `git diff --check 4ab05778dfe54c9b1d0b11f251bfbee3d0a6a4f0...HEAD` passed.
- GitHub CI for head `63775d2` is green.
- I did not run local tests because this review worktree has no `node_modules`.

Non-blocking residual: `kickoffOnboarding` calls `sendMessage(tx)` inside an outer transaction, so `sendMessage`'s best-effort chat-message kick can run before the outer transaction commits. I agree this is not a blocker here: durable inbox rows are written transactionally, route-level inbox notify is after the kickoff service returns, and bind/backlog recovery covers missed wake hints. If tx-scoped sends spread, splitting a no-side-effect core writer would be cleaner.

yuezengwu (approved):
## Review by yzw-codex - PR #1075 synchronized pass

Recommendation: approve

This is a re-review of synchronized head `16cd682`. The branch merged newer `main`; the meaningful PR diff is still the onboarding kickoff surface, now rebased so the migration is `0068_brief_phantom_reporter` instead of the earlier `0064` slot. I checked `LATEST`, `_journal.json`, and the snapshot against that renumbering.

Implementation goal and core change: onboarding kickoff is now server-authoritative through `POST /me/onboarding/kickoff`, which find-or-creates the kind-scoped kickoff chat, sends the bootstrap only if the chat is empty under a row lock, and stamps per-membership completion/suppression after the chat exists. The web flow now calls that endpoint rather than orchestrating chat creation, bootstrap send, and completion stamping in separate browser steps.

Findings: no blocking correctness issues found in the synchronized head.

Notes for human review: this still contains core DB/state-model changes: `chats.onboarding_kickoff_key` plus a unique index, removal of `users.onboarding_dismissed_at` / `users.onboarding_completed_at`, and a `members` CHECK enforcing completed => suppressed. Runtime references to the dropped user columns remain absent; remaining web `onboardingDismissedAt` / `onboardingCompletedAt` names are compatibility/auth-context state sourced from membership data.

Review checks:
- Re-read the Context Tree onboarding/messaging/inbox/team constraints and checked the diff against them.
- Reviewed the isolated PR diff against merge-base `793f944`: 17 files, including migration/schema, server kickoff route/service/chat-create path, shared schema, web kickoff flow, and tests.
- Verified the rebase-specific migration slot: `0068` added after existing `0064`-`0067`, `LATEST` points to `0068`, and the `0068` snapshot contains the kickoff key/index and members CHECK while omitting the dropped user columns.
- GitHub Actions for `16cd682` are green: Migrations, Lint & Type Check, Test, and CodeQL.
- I did not run local tests because this fresh review worktree has no `node_modules`; I relied on GitHub CI for execution.

Non-blocking residual: same as the first pass, `kickoffOnboarding` calls `sendMessage(tx)` inside an outer transaction, so `sendMessage`'s best-effort chat-message kick can run before the outer transaction commits. The durable inbox rows remain transactional, route-level inbox notify happens after the kickoff service returns, and bind/backlog recovery covers missed wake hints, so I do not think this blocks the PR. If tx-scoped sends become a broader pattern, splitting a no-side-effect core writer would be cleaner.

PR #1072, opened 2026-06-15
feat(web): blocking reply UI for ask-question; align agent prompts/docs

yuezengwu:
## What

Reworks the **ask-question (`format="request"`) reply experience** so an open question blocks the asked human until they answer, and any answer resolves it.

### Web UI
- **Blocking timeline**: when the viewer is the **target** of an unresolved question, the timeline is truncated at the **FIFO-oldest** such question — every item after it is hidden (with an `Answer the question above to see N newer messages` marker) until it resolves. Target-only; other participants keep the full timeline + a read-only card. The "new messages" pill is suppressed while blocked.
- **Decoupled options**: clicking a dock option sets a separate selection state and no longer rewrites the composer; the composer holds free text only.
- **Any answer resolves**: picking an option **or** typing free text both attach `metadata.resolves` (`kind="answered"`), clear the red dot, and unblock — gated on every required question being answered. The "discuss without resolving" path is removed from the human's surface.

### Agent prompts / docs (aligned to the new model)
- `packages/client/src/runtime/agent-briefing.ts` (`## Asking Humans`), both `first-tree` skills, `docs/development/comms-behavior-cases.md`, and `packages/shared/src/schemas/message.ts` comments now state: a question **blocks the target's chat**, **any answer resolves**, FIFO oldest-first, and **prefer free-text questions** (use `--option` only for short, single-meaning, mutually-exclusive picks).

## Why
Dense option sets are hard to choose from, and an easy-to-scroll-past red dot let asks go unanswered. Blocking + free-text-first makes asks unmissable and easier to answer. (Product-semantic shift tracked in the companion tree PR — see below.)

## Self-review fixes (post independent review)
- `findBlockingRequest` skips requests whose payload fails to parse (avoids a stuck block with no answer surface).
- `buildResolveAnswer` drops optional questions left unanswered (no `→ —` placeholder lines).
- Image attached while blocked: answer sends as text, image stays pending with a hint.

## Testing
- web typecheck / biome / design-token guardrails clean; **web 813** tests, **client agent-briefing 36** tests pass; rebased onto latest `origin/main`.
- QA: **PASS** (target-only truncation, observer view, option decoupling, option+free-text resolve, required gating, FIFO, boundaries, regressions). Report: `qa-reports/ask-blocking-reply-ui-qa-20260615.md`.

## Companion tree PR
A Context Tree PR updating `product/work-model/attention-and-review.md` with the "open question is blocking for the target" product semantic will be linked here. (cross-link added after creation)

PR #1081, opened 2026-06-15
fix(web): free text alone answers/resolves an option question

yuezengwu:
## Bug
Reported by yuezengwu: when answering an open question **with free text instead of picking an option**, the **Send button stays disabled** — so an option-type question couldn't be answered by typing.

## Cause
The blocking-answer send gate `allRequiredAnswered` required an option *selection* for option (`single`) questions; free text didn't satisfy it. This contradicts the shipped "option **or** free text both resolve" contract (#1072).

## Fix
- `allRequiredAnswered`: free text alone is a complete answer — it enables Send (an option pick alone still works too).
- `buildResolveAnswer`: free text stands in as the answer for any question with no option selection, so an option question answered by free text resolves as `"<prompt> → <free text>"` (no `→ —` placeholder).

## Tests
- Unit: `allRequiredAnswered` / `buildResolveAnswer` for free-text-on-option-question, plus updated mixed-request expectations.
- DOM: typing free text into an option-question dock enables Send and resolves with the free-text answer.
- Full web suite green (806); typecheck / biome / design-token clean.

Follow-up to #1072 (open-question blocking reply).

### DIFF
--- a/packages/web/src/components/chat/__tests__/request-state.test.ts
+++ b/packages/web/src/components/chat/__tests__/request-state.test.ts
@@ -400,9 +400,12 @@ describe("allRequiredAnswered / buildResolveAnswer (blocking surface — both ch
     allowExtra: false,
   };
 
-  it("a required single needs a selection; free text does not satisfy it", () => {
+  it("free text satisfies a required option question — no option pick needed (the reported bug)", () => {
     expect(allRequiredAnswered(single, {}, "")).toBe(false);
-    expect(allRequiredAnswered(single, {}, "just typing")).toBe(false);
+    // Free text alone enables send for an option question (was incorrectly false).
+    expect(allRequiredAnswered(single, {}, "let's hold off")).toBe(true);
+    expect(buildResolveAnswer(single, {}, "let's hold off")).toBe("Ship? → let's hold off");
+    // An option pick alone still works.
     expect(allRequiredAnswered(single, { "Ship?": "yes" }, "")).toBe(true);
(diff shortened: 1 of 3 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1081

**Recommendation**: approve
- Reason: The change restores the documented Web request-answer contract: a free-text answer resolves an option question without requiring an option pick.

**Level**: C
- Semantic lift: none; the diff stays in Web request-answer UI/state helpers and does not touch subscriptions, WS frame parsing, or data-layer contracts.

**PR Summary**
- Problem: An option-type open question could not be answered with typed free text because Send stayed disabled.
- Design: `allRequiredAnswered` treats non-empty free text as a complete blocking answer, and `buildResolveAnswer` uses that text for unanswered option prompts instead of emitting an empty placeholder.
- Affected modules: Web chat request state helpers and focused unit/DOM regression tests.

**Review Notes**
✅ 1. The implementation matches the Context Tree messaging contract that Web attaches a resolution signal to either an option pick or free text, and that every human answer clears the open request.
✅ 2. The regression coverage exercises both helper behavior and the DOM path where typing free text enables Send and sends `metadata.resolves`.
✅ 3. Existing option-only answers remain covered and continue to resolve with canonical prompt/answer content.

**Done**
- Approved.
- I did not run local tests per review-mode rules; GitHub CI showed lint/typecheck and CodeQL green while the `Test` job was still in progress at review time.

PR #1082, opened 2026-06-16
fix(chat): unify participant avatar resolution

bestony:
## Summary
- Resolve chat participant avatars through the canonical server avatar resolver.
- Add human backing-user avatar joins for chat list, chat detail, and participant projections.
- Keep non-human agents without uploads on the initials fallback while preserving uploaded-avatar precedence.

## Verification
- pnpm --filter @first-tree/server test -- --run src/__tests__/chat-detail-avatar-fields.test.ts
- pnpm --filter @first-tree/server test -- --run src/__tests__/me-chat-service.test.ts
- pnpm check && pnpm typecheck

Note: the two focused server test commands currently run the full server test suite via the package script; both runs passed (161 files, 1426 tests).

### DIFF
--- a/packages/server/src/__tests__/chat-detail-avatar-fields.test.ts
+++ b/packages/server/src/__tests__/chat-detail-avatar-fields.test.ts
@@ -1,8 +1,9 @@
 import { eq } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
 import { agents } from "../db/schema/agents.js";
+import { users } from "../db/schema/users.js";
 import { createChat } from "../services/chat.js";
-import { createTestAgent, useTestApp } from "./helpers.js";
+import { createTestAdmin, createTestAgent, useTestApp } from "./helpers.js";
 
 /**
  * Regression: the chat-detail surface used to drop the manager-configured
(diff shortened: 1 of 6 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1082

**Recommendation**: approve
- Reason: the change routes chat participant avatar rendering through the existing canonical resolver while adding the human backing-user avatar join only after chat membership has scoped the participant set.

**Level**: B-high
- Path initial: `packages/server/**` with `services/chat.ts` and `services/me-chat.ts` touched.
- Semantic lift: none.

**PR Summary**
- Problem: chat list/detail/participant projections did not consistently use the human backing user's external avatar fallback.
- Design: join `members -> users` in the relevant chat participant projections and pass the result through `resolveAvatarImageUrl`, preserving uploaded-avatar precedence and keeping non-human agents without uploads on the initials fallback.
- Affected modules: server chat API/routes, chat services, me-chat list projection, and focused regression tests.

**Review Comments**
- No blocking findings.
- The new `members/users` joins are applied to already membership-scoped participant rows, so the chat-scoped identity rendering boundary remains intact.
- The resolver keeps the intended precedence: uploaded avatar, then human backing-user avatar, then `null`; non-human rows without uploads continue to return `null`.
- Non-blocking note: `packages/shared/src/schemas/chat.ts` and `packages/shared/src/schemas/me-chat.ts` still describe `avatarImageUrl` as a manager-uploaded avatar URL. The wire behavior is now the canonical resolver result, so those comments should be updated in a follow-up.

**Verification**
- Did not run local tests, per PR Review mode.
- Remote CI reports success for lint/typecheck/test.

PR #1084, opened 2026-06-16
docs: guard local-only refs/stashes before rm -rf in legacy migration (follow-up to #1083)

liuchao-001:
## What

Follow-up to #1083. Closes the last data-loss gap in the *one-time legacy-layout migration* recipe (`packages/client/src/runtime/agent-briefing.ts`, *Source Repositories* block).

## Why

#1083 guarded the linked worktrees and the legacy checkout's **own working tree + HEAD** before `rm -rf <legacy>`. But a non-bare checkout's `.git` can also hold **local-only history that those checks never see**:

- local **branches not checked out in any worktree**, and
- **stash** entries.

If such a branch (or stash) isn't reachable from `origin/<default>`, `rm -rf <legacy>` still loses that work even though `status --porcelain` is empty and `HEAD` is merged. Raised by codex-assistant on #1083.

## The fix

Extend the legacy-checkout self-check with two more gates before the delete:

```bash
git -C <legacy> branch --no-merged origin/<default>  # empty ⇒ no unmerged local branch
git -C <legacy> stash list                            # empty ⇒ no stashed work
rm -rf <legacy>                                       # only after ALL of the above are clear
```

If either is non-empty, **stop** — push / migrate the work or ask a human. The prose now frames the bar as *everything the `rm -rf` would destroy* (worktrees + own tree + local-only refs/stashes).

## Test

New assertions in `agent-briefing.test.ts`. `biome` / `tsc --noEmit` / the briefing + cwd-redesign suites (43 tests) all pass.

Cross-link: builds on #1083.


yuezengwu (approved):
## Review by yzw-codex - PR #1084

**Recommendation**: approve
- Reason: This follow-up closes the remaining data-loss gap from #1083 by guarding local-only branches and stashes before deleting the legacy non-bare checkout.

**What changed**
- Goal: make the one-time legacy-layout migration recipe safe for everything `rm -rf <legacy>` would destroy.
- Core change: the generated agent briefing now requires `git -C <legacy> branch --no-merged origin/<default>` and `git -C <legacy> stash list` to be empty before deletion, and tells the agent to stop and migrate/push/ask a human if either has output.
- Tests: `agent-briefing.test.ts` now pins the branch/stash guard strings and the `only after ALL of the above are clear` delete condition.

**Review comments**
✅ No blocker found. The previous parked-local-branch/stash data-loss case is covered.
✅ The change stays inside the agent-managed repo model: runtime still only declares layout; destructive cleanup remains explicit agent-side migration guidance with guards.
✅ No core data structure or database changes.

**Verification**
- Ran `git diff --check` on the PR diff: clean.
- Did not run the test suite locally because this temporary worktree has no `node_modules`; author reports `biome`, `tsc --noEmit`, and the briefing + cwd-redesign suites passed.

packages/client/src/runtime/agent-briefing.ts
```diff
@@ -475,11 +475,14 @@ function sourceRepositoriesBlock(sourceRepos: ReadonlyArray<PredeclaredSourceRep
...
-    "it has no uncommitted changes AND its HEAD is already merged into",
-    "`origin/<default>`, so no work is lost. `git worktree remove` refuses",
-    "the main checkout, so verify that one by hand before the `rm -rf`:",
+    "then retire the legacy checkout. Clear the same zero-data-loss bar for",
+    "**everything** the final `rm -rf` would destroy — its linked worktrees,",
+    "its own working tree, AND any local-only history in its `.git` (branches",
+    "checked out in no worktree, plus stashes). Delete a tree only when it has",
+    "no uncommitted changes and its HEAD is merged into `origin/<default>`;",
+    "`git worktree remove` refuses the main checkout, so verify it by hand. If",
+    "any check below is non-empty, stop — push / migrate that work or ask a",
```
yuezengwu, line 484: This needs to mention exit status, not only non-empty output. Two of the destructive guards below are `merge-base --is-ancestor` checks; when they fail, they normally print nothing and only return a non-zero exit code. With the current sentence, an agent can read empty output as safe and continue to `rm -rf <legacy>` even though HEAD or a linked worktree is not merged.

Suggested wording: `If any status/branch/stash output is non-empty, or any merge-base check exits non-zero, stop — push / migrate that work or ask a human before deleting.`

yuezengwu (changes requested):
Goal/core change: this PR tightens the one-time legacy-layout migration briefing so agents guard local unmerged branches and stashes before deleting the legacy checkout, with regression assertions in `agent-briefing.test.ts`.

I found one blocker in the wording. There are no core data-structure or database changes in this PR.

PR #1085, opened 2026-06-16
fix: add inbox delivery status index

baixiaohang:
## Summary

- Add `idx_inbox_entries_message_status` on `inbox_entries(message_id, status)` for chat message deliveryStatus lookups.
- Add Drizzle migration metadata and a focused regression test that verifies the migrated index exists with the expected column order.

## Validation

- [x] `pnpm check` (passes; reports unrelated existing Biome hints in `packages/web/src/pages/agent-detail/prompt-tab.tsx`, `apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts`, and `packages/client/src/runtime/workspace-migrations.ts`)
- [x] `pnpm typecheck`
- [x] `VITEST_MAX_FORKS=2 pnpm --filter @first-tree/server test -- inbox-delivery-indexes.test.ts` (ran the server suite: 162 files / 1427 tests passed)
- [x] `node scripts/check-migrations.mjs`

## Change Surface

## Notes

- docs or tests updated to match: added `inbox-delivery-indexes.test.ts`
- follow-up work: query-shape rewrite can be handled separately; this PR only adds the index

Production note: Drizzle migrations run inside a transaction, so the migration uses regular `CREATE INDEX IF NOT EXISTS`. For a large production `inbox_entries` table, operators can first run `CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_inbox_entries_message_status" ON "inbox_entries" ("message_id", "status")` outside a transaction, then apply migrations and let `IF NOT EXISTS` skip it.

### DIFF
--- a/packages/server/drizzle/0064_superb_betty_brant.sql
+++ b/packages/server/drizzle/0064_superb_betty_brant.sql
@@ -1,0 +1,18 @@
+-- Delivery-status lookup index for chat message history.
+--
+-- The messages list API derives deliveryStatus from inbox_entries by
+-- checking rows for each page message. Keeping message_id first lets the
+-- acked / delivered / any-entry probes use one bounded lookup instead of
+-- scanning the full inbox_entries table.
+--
+-- Operator note:
+-- Drizzle migrator executes migration files inside a transaction, so this file
+-- cannot use CREATE INDEX CONCURRENTLY. For a large production table, run the
+-- concurrent form manually outside a transaction before applying migrations:
+--
+--   CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_inbox_entries_message_status"
+--     ON "inbox_entries" ("message_id", "status");
+--
+-- The IF NOT EXISTS clause below will then skip the already-created index.
+CREATE INDEX IF NOT EXISTS "idx_inbox_entries_message_status"
+  ON "inbox_entries" USING btree ("message_id", "status");
--- a/packages/server/drizzle/LATEST
+++ b/packages/server/drizzle/LATEST
@@ -1,1 +1,1 @@
-0063_flat_omega_flight
+0064_superb_betty_brant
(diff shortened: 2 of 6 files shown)


yuezengwu (commented):
Review result: no blocking issues found.

Goal/core change: this PR adds a two-column PostgreSQL btree index on `inbox_entries(message_id, status)`, keeps the Drizzle schema and migration metadata in sync, and adds a regression test that verifies the migrated index exists with that column order. The messages API derives `deliveryStatus` with probes by `message_id`, two of them also filtered by `status`, so the index order matches the lookup shape.

Validation I ran locally:
- `node scripts/check-migrations.mjs`
- `VITEST_MAX_FORKS=2 pnpm --filter @first-tree/server test -- inbox-delivery-indexes.test.ts` (this ran the full server suite: 162 files / 1427 tests passed)

Database note for @yuezengwu: this is a DB migration/index change. The PR documents the regular Drizzle migration path and the production option to pre-create the index with `CREATE INDEX CONCURRENTLY` outside a transaction; please check that operational plan before merge/final approval. I’m leaving this as COMMENT rather than APPROVE for that reason.

yuezengwu (approved):
Approved after checking the DB migration plan.

This migration solves a real hot-path issue: the chat messages API derives `deliveryStatus` by probing `inbox_entries` for each returned message by `message_id`, with `status` filters for the acked/delivered cases. Existing indexes are led by `inbox_id` or `(inbox_id, chat_id, notify, status)`, so they do not cover the message-history lookup shape. Without a leading `message_id` index, the lookup can degrade as `inbox_entries` grows with fan-out and silent-context rows.

I do not recommend skipping the migration long-term. Alternatives either keep the current slow lookup, remove/degrade the user-facing delivery status, or replace this with a larger projection/query rewrite that still needs an efficient `message_id` access path. The narrow two-column index is the lowest-risk fix.

Production path: direct Drizzle migration is fine for a small table. For a large/hot production `inbox_entries` table, pre-create the index with `CREATE INDEX CONCURRENTLY IF NOT EXISTS` outside a transaction, verify the resulting index is valid, then apply migrations so the `IF NOT EXISTS` statement skips it. Failed concurrent index attempts can leave an invalid index name behind, so that verification matters.

PR #1087, opened 2026-06-16
docs: validate legacy checkout path before destructive migration cleanup (#1086)

### LINKED ISSUE #1086: Validate legacy checkout path before destructive migration cleanup
## Problem

Follow-up to #1084.

The generated agent briefing now guards the Git state inside the legacy checkout before the final `rm -rf <legacy>`: linked worktrees, the checkout's own clean/merged state, unmerged local branches, and stashes.

That still does not validate the `<legacy>` path argument itself. A future agent can bind `<legacy>` to the wrong directory, and the current recipe does not prove that the target is the intended legacy flat non-bare source checkout for the declared repository.

If the mistaken target is also a Git repository and happens to pass the clean/merged/ref/stash checks, the final `rm -rf <legacy>` can delete the wrong data. Risky examples include:

- the workspace root;
- `context-tree`, `source-repos`, `worktrees`, or `.first-tree`;
- an unbound sibling checkout;
- another clean and merged repository under the same workspace.

The existing guards answer "is this Git checkout safe to retire?" but not "is this exactly the checkout we intended to retire?"

## Suggested fix

Before any destructive remove, make the recipe derive and validate the target instead of asking the agent to hand-fill a naked `<legacy>` placeholder.

Recommended preflight:

1. Resolve the workspace root from `.first-tree/workspace.json`.
2. Derive the expected legacy flat source path from the declared source name, e.g. `<workspace>/<source-name>`, while the new bare clone lives at `<workspace>/source-repos/<source-name>`.
3. `realpath` the candidate and require it to be an immediate child of the current workspace root.
4. Explicitly reject empty paths, `/`, the workspace root itself, symlinks, and reserved workspace directories such as `context-tree`, `source-repos`, `worktrees`, and `.first-tree`.
5. Require `git -C "$legacy" rev-parse --show-toplevel` to equal the resolved candidate and require the repo to be non-bare.
6. Require `git -C "$legacy" remote get-url origin` to canonically match the declared source URL.
7. Only after that path preflight should the existing worktree/status/merge/local-branch/stash gates run.

Also consider changing the final destructive action to a quarantine move first, for example `mv -- "$legacy" "$legacy.retired.<timestamp>"`, and leave irreversible deletion to a human-confirmed second step.

## Why this matters

#1084 closed the hidden branch/stash data-loss gap, but the recipe still contains an irreversible delete whose target is not mechanically constrained. For an agent-facing briefing, the destructive operation needs a path-level guard as strong as the Git-state guard.

liuchao-001:
## What

Closes #1086. Follow-up to #1083 / #1084. Hardens the *one-time legacy-layout migration* recipe in the generated agent briefing (`packages/client/src/runtime/agent-briefing.ts`, **Source Repositories** block).

## Why

#1083/#1084 guarded the Git **state** inside the legacy checkout (linked worktrees, own clean/merged tree, unmerged local branches, stashes) before the final destructive remove. But the **target path** was still a naked `<legacy>` placeholder the agent hand-filled. If the mistaken target is also a Git repo that happens to be clean + merged, the destructive remove deletes the wrong data. Risky targets: the workspace root; `context-tree` / `source-repos` / `worktrees` / `.first-tree`; an unbound sibling checkout; another clean+merged repo under the same workspace.

The existing guards answer *"is this checkout safe to retire?"* but not *"is this exactly the checkout we intended to retire?"*

## The fix

A **path preflight** now derives and validates the target before any destructive action, and the destructive action becomes a **reversible quarantine**:

- `assert_legacy_target()` — resolve the workspace root from `.first-tree/workspace.json`; derive `$legacy = $WS/<name>` from the declared source name; `realpath` + immediate-child check; reject empty / `.` / `..` / slashed names, symlinks, the workspace root itself, and the reserved dirs `context-tree` / `source-repos` / `worktrees` / `.first-tree`; require `git rev-parse --show-toplevel` to equal the candidate, the repo to be **non-bare**, and `git remote get-url origin` to match the declared source URL. The per-source calls are **baked from the manifest**, so there is no naked placeholder to mis-fill.
- The existing Git-state gates run only on a `$legacy` the preflight passed (`ok:`).
- Final action is `mv -- "$legacy" "$legacy.retired.$(date +%Y%m%d%H%M%S)"`; irreversible `rm -rf` is left to a separate, human-confirmed step.

This maps 1:1 to the issue's recommended 7-step preflight + quarantine suggestion.

## Test

`agent-briefing.test.ts` updated to assert the preflight gates (manifest resolution, realpath/immediate-child, reserved-dir + symlink rejection, toplevel/non-bare/origin checks), the baked per-source calls, the `$legacy`-scoped Git-state gates, the quarantine move, and the **absence** of the old naked `rm -rf <legacy>`.

- `pnpm check` (biome) ✅ · `pnpm typecheck` (turbo, 12/12) ✅ · full `@first-tree/client` suite (998 tests) ✅
- The shipped validator was also exercised in a real sandbox: it accepts the intended `api` checkout and rejects the workspace root, reserved dirs, a symlink, a wrong-origin sibling, a bare repo, a missing path, and a slashed name.

Cross-link: builds on #1083 / #1084.


packages/client/src/runtime/agent-briefing.ts
```diff
@@ -471,35 +471,102 @@ function sourceRepositoriesBlock(sourceRepos: ReadonlyArray<PredeclaredSourceRep
...
+    'while [ "$d" != / ]; do',
+    '  [ -f "$d/.first-tree/workspace.json" ] && { WS=$(realpath "$d"); break; }',
+    '  d=$(dirname "$d")',
+    "done",
+    '[ -n "$WS" ] || echo "stop: no .first-tree/workspace.json at or above $PWD"',
+    "",
+    "# Validate ONE candidate. Args: <source-name> <declared-origin-url>.",
+    '# Prints "ok: <path>" and sets $legacy only when EVERY gate below holds.',
+    "assert_legacy_target() {",
+    "  name=$1 want=$2 legacy=$WS/$name",
```
yuezengwu: This function says it "sets `$legacy` only when EVERY gate below holds", but it assigns `legacy=$WS/$name` before any gate has passed. Every rejection after this line leaves `$legacy` pointing at the rejected candidate.

That keeps the risky path open. For example, if `$WS/api` is a clean wrong-origin repo, `assert_legacy_target api <want>` prints `reject: origin ...` and returns 1, but `$legacy` still points at `$WS/api`; if the agent then runs the Git-state gates, they can all pass and the final `mv` quarantines the wrong repo.

Please keep the candidate in a separate variable and only publish `$legacy` after all checks pass, clearing it on entry so stale/failed calls cannot be reused, for example: `name=$1 want=$2 legacy= candidate=$WS/$name`, run all checks against `$candidate`, then `legacy=$candidate` immediately before `echo "ok: $legacy"`.

yuezengwu (changes requested):
Goal/core change: this PR hardens the generated agent briefing's one-time legacy-layout migration by deriving the intended legacy checkout from the manifest, validating path/toplevel/non-bare/origin before the Git-state gates, and replacing the final irreversible `rm -rf` with a reversible quarantine move. The tests pin the new preflight, baked per-source calls, `$legacy`-scoped gates, and absence of the old naked delete.

I found one blocker in the preflight state handling. No core data structure or database changes.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1087

- Reason: the path preflight still does not implement the canonical origin comparison required by #1086, so the migration recipe can reject a legitimate legacy checkout for the same repository.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: harden the one-time legacy-layout migration recipe so agents cannot retire the wrong checkout.
- Design: derive the legacy path from the workspace/source name, validate path shape and Git identity, then quarantine with `mv` instead of deleting in place.
- Affected modules: `packages/client/src/runtime/agent-briefing.ts`, `packages/client/src/__tests__/agent-briefing.test.ts`

**Review Comments**
❌ 1. The origin check is not canonical as required by #1086. `got` and `want` are compared after only trimming a trailing `.git`, so `git@github.com:agent-team-foundation/first-tree.git` and `https://github.com/agent-team-foundation/first-tree.git` fail even though they identify the same repo. The same Source Repositories block already tells agents to compare origins canonically across `.git` and HTTP/HTTPS/SSH forms, and #1086 explicitly requires the preflight to do the same. As written, a common intended legacy checkout can be rejected as wrong-origin, leaving the shipped migration recipe incomplete. Please normalize the supported GitHub URL forms before comparing and add a test that accepts equivalent SSH/HTTPS origins while still rejecting a genuinely different repo. [R1 / `packages/client/src/runtime/agent-briefing.ts:526`]

**Executed**
- Submitted request_changes.

liuchao-001:
Addressed both review blockers in `ab3f46d2`.

**B1 — `$legacy` set before the gates pass (@yuezengwu, chatgpt-codex-connector):** `assert_legacy_target` now validates a `candidate` variable, clears `legacy=` on entry, and publishes `legacy=$candidate` only immediately before the `ok:` line. A rejected/failed call leaves `$legacy` empty, so the downstream git-state / quarantine block fails closed instead of acting on a rejected checkout. Sandbox-verified: a wrong-origin reject leaves `$legacy` empty even when preset to a value beforehand.

**B2 — canonical origin compare per #1086 (codex-assistant R1):** new `canon_url` helper normalizes the `.git` suffix and the `https` / `http` / `ssh` / `git` / scp transport forms to `host/path` before comparing — the same canonical match the reuse guard already requires. It strips only non-identifying transport bits, so two genuinely different repos never collapse (no false-accept). Added a behavioral test that extracts the shipped `canon_url` and runs it under `sh`: the seven equivalent `.git`/https/http/ssh/git/scp forms of one repo collapse to a single key, while a different org/path and a different host stay distinct. Sandbox-confirmed: an `api` checkout cloned via `git@github.com:org/api.git` is now accepted against an `https://github.com/org/api` manifest URL.

`pnpm check` (biome) ✅ · `pnpm typecheck` ✅ · full `@first-tree/client` suite (999 tests) ✅.

yuezengwu (approved):
Goal/core change: this PR hardens the generated agent briefing's one-time legacy-layout migration by deriving the intended legacy checkout from the manifest, validating path/toplevel/non-bare/origin before the Git-state gates, and replacing the final irreversible `rm -rf` with a reversible quarantine move. The latest update also addresses both request-changes blockers.

**What I rechecked**
- `$legacy` is now cleared on entry and only assigned from `$candidate` after every preflight gate passes, so rejected candidates cannot leak into the git-state gates.
- Origin comparison now uses `canon_url()` to collapse `.git` plus HTTPS/HTTP/SSH/git/scp-style GitHub URL forms to host/path before comparing, while still rejecting genuinely different repos.
- Tests now include the string-level preflight assertions and a behavioral shell check for canonical URL equivalence/non-equivalence.

No core data-structure or database changes.

**Verification**
- Reviewed updated diff at `ab3f46d2`.
- Ran `git diff --check origin/main...HEAD`: clean.
- Manually exercised the shipped `canon_url` shell function against the same HTTPS/HTTP/SSH/git/scp URL set and different-repo negatives: passed.
- Did not run the full test suite locally because this temporary review worktree has no `node_modules`; author reports `pnpm check`, `pnpm typecheck`, and full `@first-tree/client` suite passed.

baixiaohang (approved):
## Review by codex-assistant - PR #1087

**Recommendation**: approve
- Reason: the latest update addresses both manual blockers from review: rejected preflight candidates no longer leak through `$legacy`, and origin comparison now uses canonical URL matching as required by #1086.

**Level**: A
- Semantic lift: none
- Approval authorization: explicitly authorized by baixiaohang in chat.

**PR Summary**
- Problem: harden the generated agent briefing's one-time legacy-layout migration so agents cannot retire the wrong checkout.
- Design: derive the legacy path from workspace/source identity, validate path/toplevel/non-bare/origin before the Git-state gates, and quarantine with `mv` instead of deleting in place.
- Affected modules: `packages/client/src/runtime/agent-briefing.ts`, `packages/client/src/__tests__/agent-briefing.test.ts`

**Review Comments**
✅ 1. `assert_legacy_target` now validates `$candidate`, clears `$legacy` on entry, and publishes `$legacy=$candidate` only after every gate passes, so a rejected checkout cannot be reused by the Git-state/quarantine block.
✅ 2. `canon_url` now normalizes `.git` plus HTTPS/HTTP/SSH/git/scp-style URL forms to `host/path`, with behavioral test coverage for same-repo equivalence and different-repo negatives.
✅ 3. yuezengwu has approved the latest commit `ab3f46d2`.
⚠️ 4. I did not run local tests per PR Review mode. Remote Lint & Type Check / Test / Migrations are passing, but the CodeQL aggregate check is still failing on the added test regex alert and must be handled before merge.

**Executed**
- Submitted approve review after explicit authorization.

packages/client/src/__tests__/agent-briefing.test.ts
```diff
@@ -604,24 +606,105 @@ describe("buildAgentBriefing — # Working in First Tree subsections", () => {
...
+    const briefing = buildAgentBriefing(
+      makeOpts({
+        sourceRepos: [{ absolutePath: `${AGENT_HOME}/source-repos/api`, url: "https://github.com/org/api" }],
+      }),
+    );
+    // Extract the shipped `canon_url` shell function verbatim from the recipe
+    // and exercise it — a string assertion alone cannot prove the transport
+    // forms actually collapse. codex-assistant (PR #1087) asked for a test that
+    // accepts equivalent SSH/HTTPS origins while still rejecting a different repo.
+    const match = briefing.match(/canon_url\(\) \{\n[\s\S]*?\n\}/);
```
yuezengwu: This still leaves the CodeQL blocker in place. The latest patch avoids interpolating `canonFn` into `sh -c`, but the reported alert is on this regex: `briefing.match(/canon_url\(\) \{\n[\s\S]*?\n\}/)`. Since that line is unchanged, the CodeQL aggregate is likely to keep failing.

Please avoid the regex extraction entirely and use deterministic string slicing, for example:

```ts
const start = briefing.indexOf("canon_url() {");
expect(start).not.toBe(-1);
const end = briefing.indexOf("\n}", start);
expect(end).not.toBe(-1);
const canonFn = briefing.slice(start, end + "\n}".length);
```

That keeps the test behavior but removes the polynomial-regex pattern CodeQL is blocking on.

yuezengwu (changes requested):
Goal/core change remains the same: this PR hardens the generated agent briefing's legacy-layout migration path preflight and quarantine flow. The latest synchronization appears intended to clear the remaining CodeQL annotation in the new `canon_url` behavioral test.

I found one remaining blocker: the patch changes how the extracted function is executed, but the CodeQL alert is on the regex extraction itself, which is still present. No core data-structure or database changes.

yuezengwu (approved):
Goal/core change remains the same: this PR hardens the generated agent briefing's one-time legacy-layout migration by deriving and validating the intended legacy checkout before git-state gates, then quarantining instead of deleting.

The latest sync addresses my remaining CodeQL blocker: the `canon_url` behavioral test no longer uses the polynomial regex extraction and now uses deterministic `indexOf`/`slice` extraction before running the shipped shell function from a temp script. The previous `$legacy` state leak and canonical-origin blockers remain fixed.

No core data-structure or database changes.

Verification:
- Reviewed latest head `b46074f0` and the incremental diff from `68172b59`.
- Confirmed `briefing.match(/canon_url.../)` is gone and `rg` only finds the intended `indexOf("canon_url() {\n")` extraction.
- Ran `git diff --check 3fb1a9dcebbfd5312b5d76f6575557e65bf1af1c...HEAD`: clean.
- Did not run local Vitest because this temporary review worktree has no `node_modules`; current GitHub CI run is still in progress.

baixiaohang (approved):
## Review by codex-assistant - PR #1087

**Recommendation**: approve
- Reason: the latest head keeps the legacy migration preflight fixes intact and replaces the CodeQL-triggering regex extraction in the `canon_url` behavioral test with deterministic `indexOf`/`slice` extraction.

**Level**: A
- Semantic lift: none
- Approval authorization: explicitly authorized by baixiaohang in chat for latest commit `b46074f0`.

**PR Summary**
- Problem: harden the generated agent briefing's one-time legacy-layout migration so agents cannot retire the wrong checkout.
- Design: derive and validate the legacy checkout path before Git-state gates, publish `$legacy` only after preflight success, compare origins canonically, and quarantine instead of deleting in place.
- Affected modules: `packages/client/src/runtime/agent-briefing.ts`, `packages/client/src/__tests__/agent-briefing.test.ts`

**Review Comments**
✅ 1. The `$legacy` stale-target blocker remains fixed: gates validate `$candidate`, clear `$legacy` on entry, and publish only after every gate passes.
✅ 2. The canonical-origin blocker remains fixed: `canon_url` normalizes `.git` plus HTTPS/HTTP/SSH/git/scp-style URL forms to `host/path` while keeping different repos distinct.
✅ 3. The CodeQL ReDoS blocker is addressed: the test now extracts `canon_url` with `indexOf`/`slice` instead of the polynomial regex pattern.
✅ 4. Remote checks are passing, including CodeQL, Lint & Type Check, Test, and Migrations. I did not run local tests per PR Review mode.

**Executed**
- Submitted approve review after explicit authorization.

PR #1089, opened 2026-06-16
fix: resolve a pinned question on Enter for options-only answers

yuezengwu:
## Problem

Answering a blocking structured-request question has two channels: clicking an option pill (stored in `answerSelections`) and typing free text in the composer. Enter-to-send was wired **only** on the composer textarea's own `onKeyDown`, so it fired only while the textarea was focused.

An **options-only** answer never touches the composer — clicking a pill moves focus to the (sr-only) radio — so **Enter never reached the handler**, even though the dock's helper line promises *"↵ Send answers and resolves this question"*. Typing text worked (composer focused); picking an option alone did not.

Reported via blocking-reply regression testing:
- ✅ with composer text → Enter sends
- ❌ option selected, composer empty → Enter does nothing (Send button still worked)

## Fix

Add an Enter-to-resolve backstop bound on the **composer-footer subtree** (which wraps both the dock and the composer), active only while the pinned question is resolvable (`dockCanResolve`). It resolves on Enter from anywhere in that subtree **except** a control that owns Enter itself (the composer textarea, another text input, a button, a link); radios/checkboxes do not, so Enter on a selected option pill resolves.

- **Scoped to the footer, not `window`** — a portaled dialog (custom `Select` listbox, Radix dialog) rendered over a still-mounted chat can't bubble Enter into the listener and silently resolve the pinned question in the background.
- **No double-send** — the composer-focused path is unchanged (its own handler still fires; the backstop early-returns on a focused textarea).
- **No-op when unresolvable** — the listener isn't bound until every required question is answered.

## Adversarial review + similar-bug sweep

This change was adversarially reviewed (double-send, IME, event ordering, ref-during-render under StrictMode, cleanup, global-hijack) — the one real finding (a `window`-global listener could resolve from a portaled dialog over the chat) is addressed by the footer-scoping above. A sweep of all keyboard affordances in `packages/web` for the same "promise vs. focus-gated handler" pattern found no other instances.

## Test

Adds a regression test: options-only answer + Enter (dispatched on the focused option radio, bubbling to the footer-scoped backstop) resolves the question; Enter before any answer is a no-op. `pnpm --filter @first-tree/web vitest` (chat-view + request-card/dock DOM suites), `tsc --noEmit`, and biome all pass.

### DIFF
--- a/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
@@ -861,6 +861,91 @@ describe("ChatView", () => {
     await act(async () => root.unmount());
   });
 
+  it("resolves an options-only answer via Enter while focus is off the composer", async () => {
+    const { ChatView } = await import("../chat-view.js");
+    const dockMessages = messages([
+      message({
+        id: "req-enter",
+        senderId: "agent-1",
+        format: "request",
+        content: "Pick the deploy color.",
+        metadata: {
+          mentions: ["human-agent-self"],
+          request: {
+            subject: "Deploy",
+            questions: [
+              {
+                id: "q1",
+                prompt: "Deploy color?",
+                kind: "single",
+                options: ["Blue-green", "Rolling update"],
+                required: true,
+              },
+            ],
+          },
+        },
+        createdAt: "2026-05-28T12:00:00.000Z",
+      }),
+    ]);
+    const { container, queryClient, root } = await renderDom(
+      <ChatView agentId="agent-1" chatId="chat-1" />,
+      (client) => seedChat(client, chatDetail(), messages([])),
+      "/",
+    );
+
+    await act(async () => {
+      queryClient.setQueryData(["chat-messages", "chat-1"], dockMessages);
+    });
+    await flush();
+    await waitForText(container, "Awaiting your answer");
+
+    const textarea = container.querySelector<HTMLTextAreaElement>("textarea");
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1089

**Recommendation**: approve
- Reason: The footer-scoped Enter backstop preserves the existing composer send path while closing the option-only focus gap, and the regression test covers the previously broken path.

**Level**: C

**PR Summary**
- Problem: option-only blocking request answers could not be submitted with Enter because focus remained on the option radio instead of the composer textarea.
- Design: bind an Enter-to-resolve backstop on the composer footer only while the dock is resolvable, skip controls that own Enter, and delegate to the existing `handleSend` path.
- Affected modules: `packages/web/src/pages/workspace/center`

**Review Notes**
✅ 1. The listener is scoped to the dock/composer footer and gated by `dockCanResolve`, so Enter resolves only when the pinned Web answering surface is already complete. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:1606`]
✅ 2. The textarea/button/link/input exclusions keep the existing composer and control-specific Enter behavior from double-sending or being hijacked. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:1611`]
✅ 3. The regression test exercises both the unselected no-op and selected option-radio submit path. [R5 / `packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx:864`]

**Done**
- Approved.

Tests not run by reviewer, per review-mode rules; relied on code review plus the author-reported checks.

PR #1096, opened 2026-06-16
fix(client): keep claude-code-tui startup from deadlocking on unanswerable prompts

yuezengwu:
## Problem

The `claude-code-tui` runtime launches `claude --resume <id>` inside a **detached tmux session** and polls `waitForReady` for the ready surface (`bypass permissions on` marker + `❯` prompt). If Claude Code shows **any interactive prompt that needs a keystroke** before that surface, the detached runtime can't answer it → `waitForReady` times out (30s) → `SessionManager` retries forever as `reasonCode:"unknown"` (transient). Observed live: a 119.5k-token / 2h41m session deadlocked on the resume-strategy menu and spammed `resilience.session.retry_*` every ~60s, recurring across chats.

This PR makes the runtime handle the known startup prompts instead of deadlocking, and stops the *infinite* retry when the prompt is genuinely unanswerable.

## Changes (all in `waitForReady` + the error taxonomy)

1. **Resume-strategy menu** (large/old session): detect it and select option 1 **"Resume from summary"** via `"1"`+Enter; grant a fresh longer deadline (`summaryReadyTimeoutMs`, default 120s) since summary generation is an LLM round-trip that exceeds the normal 30s start window.
2. **Bypass-Permissions warning** (`--dangerously-skip-permissions` → "WARNING: …Bypass Permissions mode" / "Yes, I accept" / "No, exit"): same deadlock class as the existing trust dialog, only masked today by HOMEs that carry `skipDangerousModePermissionPrompt` (First Tree sets it nowhere → fresh HOMEs deadlock). Accept option 1 **"Yes, I accept"** via `"1"`+Enter — explicit by number because a bare Enter on a "No, exit" default would *quit* claude.
3. **Login / re-auth wall** (interactive login-method selector): **not** keystroke-answerable — a human must re-authenticate. `waitForReady` throws `ClaudeTuiLoginRequiredError`, which the taxonomy classifies **permanent** (`claude_login_required`) → `SessionManager` stops retrying and surfaces the session as `errored` instead of spamming "did not become ready" forever.
4. **Robustness**: normalize U+00A0 (NBSP) → ASCII space on the captured pane before matching (the TUI prints NBSP in its chrome; `tui-markers.ts` documents this).
5. **Transcript safety (the key correctness guard)**: on resume the captured pane re-renders prior transcript, which can quote any of these prompt strings. So `waitForReady` checks the **ready surface FIRST every poll** — a loaded session (marker + `❯`) returns before any modal/login detection, so visible history mentioning `run /login` or `Bypass Permissions mode` can never inject a keystroke or trigger a false permanent failure. The detectors are also tightened to live-modal-only shapes (login = the selector title + an option label, not loose `run /login`; bypass = full warning title + both option labels), so even the pre-ready window can't be tripped by prose. Under-detecting (fall back to the normal retry/timeout) is the safe direction.

Already closed by existing launch flags and out of scope: the project-`.mcp.json` trust modal (`--strict-mcp-config`) and the in-turn AskUserQuestion picker (`--disallowed-tools AskUserQuestion`).

## Verification

- `pnpm --dir packages/client exec vitest run src/__tests__/tui-tmux-session.test.ts src/__tests__/error-taxonomy.test.ts src/__tests__/tui-suspend-queued-recovery.test.ts` — **75 passed**. Includes a `waitForReady ordering (transcript safety)` group driven through a `capture`/`send` injection seam: a ready pane quoting **every** modal/login string resolves with no keystroke and no throw (and asserts the detectors *would* fire on it, proving ready-first is load-bearing); a live login selector rejects with `ClaudeTuiLoginRequiredError`; a live bypass modal is accepted via `"1"`+Enter.
- `pnpm --filter @first-tree/client typecheck` — clean. Biome — clean.
- Independent QA: full client suite 1015/1015, typecheck, `pnpm check` clean; a real-tmux probe reproduced the transcript false-positive on the pre-fix head and the ready-first commit resolves it. The package TUI resume e2e is inconclusive but fails identically on `origin/main` (pre-existing harness issue, not introduced here).

Note: the bypass-modal and login-wall paths are masked on the dev machine (HOME already onboarded / authed), so they're covered by unit tests + a real-tmux probe + the live `claude.exe` strings, not a live daemon repro.

### DIFF
--- a/packages/client/src/__tests__/error-taxonomy.test.ts
+++ b/packages/client/src/__tests__/error-taxonomy.test.ts
@@ -106,6 +106,15 @@ describe("error-taxonomy.classify", () => {
       expect(c.reasonCode).toBe("auth_refresh_failed");
     });
 
+    it("ClaudeTuiLoginRequiredError (session source) → permanent, no retry", () => {
+      const err = new Error("claude TUI requires re-authentication (run /login) — session=ftth-x");
+      err.name = "ClaudeTuiLoginRequiredError";
+      const c = classify(err, { source: "session" });
+      expect(c.kind).toBe(ERROR_KINDS.PERMANENT);
+      expect(c.reasonCode).toBe("claude_login_required");
+      expect(c.strategy.kind).toBe("none");
+    });
+
     it("ClientOrgMismatchError → permanent with default message", () => {
       const c = classify(noMessageShape({ name: "ClientOrgMismatchError" }));
       expect(c.kind).toBe(ERROR_KINDS.PERMANENT);
--- a/packages/client/src/__tests__/tui-tmux-session.test.ts
+++ b/packages/client/src/__tests__/tui-tmux-session.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from "vitest";
 import {
   deriveSessionName,
+  isBypassPermissionsWarning,
+  isClaudeLoginWall,
   isResumeSummaryPrompt,
   isWorkspaceTrustPrompt,
   ownedSessionPrefix,
@@ -131,5 +133,59 @@ Resuming the full session will consume a substantial portion of your usage limit
 `;
 
     expect(isResumeSummaryPrompt(pane)).toBe(false);
+  });
+});
+
+describe("isBypassPermissionsWarning", () => {
+  it("detects the one-time bypass-permissions acceptance modal", () => {
+    const pane = `
+ WARNING: Claude Code running in Bypass Permissions mode
+
+ In Bypass Permissions mode, Claude Code will not ask for your approval
+ before running potentially dangerous commands.
+
+ ❯ 1. Yes, I accept
+   2. No, exit
+
+ Enter to confirm · Esc to cancel
+`;
+
+    expect(isBypassPermissionsWarning(pane)).toBe(true);
+  });
+
+  it("does not treat the normal ready surface ('bypass permissions on') as the modal", () => {
+    const pane = `
+❯ Try "edit <filepath> to..."
+⏵⏵ bypass permissions on (shift+tab to cycle)
+`;
+
+    expect(isBypassPermissionsWarning(pane)).toBe(false);
+  });
+});
+
+describe("isClaudeLoginWall", () => {
+  it("detects the interactive login-method selector", () => {
+    const pane = `
+ Select login method:
+
+ ❯ 1. Login with Claude account
+   2. Login with Claude Console (API)
+`;
+
+    expect(isClaudeLoginWall(pane)).toBe(true);
+  });
+
+  it("detects a 'run /login' re-auth prompt", () => {
+    expect(isClaudeLoginWall("OAuth refresh token is no longer valid; run /login to re-authenticate")).toBe(true);
+    expect(isClaudeLoginWall("Not authenticated. Please run /login and try again.")).toBe(true);
+  });
+
+  it("does not fire on the normal ready surface", () => {
+    const pane = `
+❯ Try "edit <filepath> to..."
+⏵⏵ bypass permissions on (shift+tab to cycle)
+`;
+
+    expect(isClaudeLoginWall(pane)).toBe(false);
   });
 });
--- a/packages/client/src/__tests__/tui-tmux-session.test.ts
+++ b/packages/client/src/__tests__/tui-tmux-session.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 import {
   deriveSessionName,
+  isResumeSummaryPrompt,
   isWorkspaceTrustPrompt,
   ownedSessionPrefix,
 } from "../handlers/claude-code-tui/tmux-session.js";
@@ -91,5 +92,44 @@ describe("isWorkspaceTrustPrompt", () => {
 `;
 
     expect(isWorkspaceTrustPrompt(pane)).toBe(false);
+  });
+});
+
+describe("isResumeSummaryPrompt", () => {
+  it("detects Claude Code's large-session resume strategy menu", () => {
+    const pane = `
+This session is 2h 41m old and 119.5k tokens.
+Resuming the full session will consume a substantial portion of your usage limits. We recommend resuming from a summary.
+
+ ❯ 1. Resume from summary (recommended)
+   2. Resume full session as-is
+   3. Don't ask me again
+
+ Enter to confirm · Esc to cancel
+`;
+
+    expect(isResumeSummaryPrompt(pane)).toBe(true);
+  });
+
+  it("does not treat the normal ready surface as a resume menu", () => {
+    const pane = `
+❯ Try "edit <filepath> to..."
+⏵⏵ bypass permissions on (shift+tab to cycle)
+`;
+
+    expect(isResumeSummaryPrompt(pane)).toBe(false);
+  });
+
+  it("does not trip on the trust prompt (disjoint option labels)", () => {
+    const pane = `
+ Quick safety check: Is this a project you created or one you trust?
+
+ ❯ 1. Yes, I trust this folder
+   2. No, exit
+
+ Enter to confirm · Esc to cancel
+`;
+
+    expect(isResumeSummaryPrompt(pane)).toBe(false);
   });
 });
--- a/packages/client/src/__tests__/tui-tmux-session.test.ts
+++ b/packages/client/src/__tests__/tui-tmux-session.test.ts
@@ -1,11 +1,13 @@
 import { describe, expect, it } from "vitest";
 import {
+  ClaudeTuiLoginRequiredError,
   deriveSessionName,
   isBypassPermissionsWarning,
   isClaudeLoginWall,
   isResumeSummaryPrompt,
   isWorkspaceTrustPrompt,
   ownedSessionPrefix,
+  waitForReady,
 } from "../handlers/claude-code-tui/tmux-session.js";
 
 const CID = "client_abcd1234";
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1096

**Recommendation**: approve
- Reason: the change handles a previously unanswerable Claude Code TUI resume menu inside the existing ready-polling boundary, and the longer deadline is scoped to the summary path only.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none

**PR Summary**
- Problem: large or old `claude-code-tui` sessions can resume into an interactive strategy menu, so detached startup never reaches the ready marker and loops on `waitForReady` timeouts.
- Design: `waitForReady` detects the resume-summary menu, sends Enter to accept the default summary option, then grants a fresh `summaryReadyTimeoutMs` window for summary generation before the normal ready surface appears.
- Affected modules: `packages/client/src/handlers/claude-code-tui/tmux-session.ts`, `packages/client/src/__tests__/tui-tmux-session.test.ts`.

**Review Comments**
✅ 1. The prompt detector is specific to the resume strategy labels plus the confirm footer, and remains disjoint from the workspace trust prompt and the normal ready surface.
✅ 2. The timeout extension is guarded by the successful resume-menu detection, so ordinary TUI starts keep the existing 30s ready window while large-session summary generation gets a separate longer deadline.
✅ 3. I found no repository code or tests relying on the old exact `30000ms` ready-timeout message.

**Executed**
- Did not run local tests or checks, per PR Review mode; I reviewed the diff and surrounding call sites statically.

yuezengwu:
## Adversarial review + sibling-bug sweep

Ran an adversarial pass on this diff and a sweep for *similar* deadlock vectors (Claude Code interactive prompts the detached tmux runtime can't answer). Summary:

**Diff itself:** no correctness blockers. Pushed a follow-up commit (`harden TUI resume-summary auto-select`) addressing two robustness notes:
- Normalize NBSP→ASCII space on the captured pane before matching (the TUI prints U+00A0 in its chrome; `tui-markers.ts` already documents this). Applied once at the capture site, so trust + resume + ready checks all benefit.
- Select "Resume from summary" via `"1"`+Enter instead of relying on the default highlight — pins the over-threshold path to summary even if a future build reorders/re-highlights the menu (degrades safely to Enter-on-default if number-select is unsupported).

Deliberately **not** changed: the one-shot latch is kept (a re-press combined with the `"1"` keystroke risks injecting a stray `"1"` user turn on a stale frame — worse than the bounded 120s it would save).

**Sibling vectors found (out of scope here — recommend follow-ups):**
1. **Bypass-Permissions warning modal** (`--dangerously-skip-permissions` → "WARNING: …Bypass Permissions mode" / "Yes, I accept" / "No, exit"). Same deadlock class as the trust dialog. Currently masked only because some dev HOMEs carry `skipDangerousModePermissionPrompt: true`; First Tree sets it nowhere, so a **fresh HOME (new machine / cloud agent) deadlocks identically**. Best fixed by writing the suppression flag at agent bootstrap rather than navigating the modal by keystroke (its default may be "No, exit" → a wrong key would *quit* claude).
2. **Expired-OAuth login wall** + **systemic gap**: a `waitForReady` timeout currently classifies as `reasonCode:"unknown"` → transient → retries ~indefinitely. An unanswerable prompt (e.g. the `/login` re-auth wall) can't be Enter'd and should be **detected and surfaced as permanent/operator-actionable**, not retried forever. Larger change; recommend a separate issue.

Already closed by existing flags: the project-`.mcp.json` trust modal (`--strict-mcp-config`) and the in-turn AskUserQuestion picker (`--disallowed-tools AskUserQuestion`).

baixiaohang (approved):
## Review by codex-assistant - PR #1096 follow-up

**Recommendation**: approve
- Reason: the synchronized head only hardens the resume-summary handling already reviewed, without expanding the call surface or changing the ordinary ready timeout path.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none

**PR Summary**
- Latest head: `d51e34fe0ac47bcb39ddb6e903a100989d749825`
- Problem: detached `claude-code-tui` sessions can deadlock on the large-session resume strategy menu before the ready marker appears.
- Current design: `waitForReady` detects the resume-summary menu, normalizes captured-pane NBSPs before matching, explicitly sends `1` then Enter to select summary, and grants a fresh `summaryReadyTimeoutMs` deadline only for that path.
- Affected modules: `packages/client/src/handlers/claude-code-tui/tmux-session.ts`, `packages/client/src/__tests__/tui-tmux-session.test.ts`.

**Review Comments**
✅ 1. NBSP normalization is local to the captured pane string, so it improves prompt/ready matching without mutating tmux state or broadening runtime behavior.
✅ 2. The explicit `1` + Enter selection remains inside the one-shot resume-summary guard; ordinary startup still uses the existing 30s ready window.
✅ 3. The sibling prompt vectors called out by the author are real follow-up work, but they are outside this PR and do not make this targeted fix unsafe to land.

**Executed**
- Approved after baixiaohang authorization for the synchronized head.
- Did not run local tests or checks, per PR Review mode; I reviewed the synchronized diff and surrounding call sites statically.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1096 follow-up 2

- Reason: the new login/bypass prompt handling can misclassify a normal ready TUI pane when prior visible conversation text contains the same prompt strings, causing a permanent session failure or an injected `1` user turn.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none; already A, and this update also touches session error taxonomy.

**PR Summary**
- Latest head: `8ac254fa9a55b9a914e40d45e6a58b7b8551ab53`
- New change since prior approval: adds bypass-permissions modal auto-accept, login/re-auth wall detection, `ClaudeTuiLoginRequiredError`, and permanent `claude_login_required` taxonomy classification.
- Affected modules: `packages/client/src/handlers/claude-code-tui/tmux-session.ts`, `packages/client/src/runtime/error-taxonomy.ts`, and tests.

**Review Comments**
❌ 1. The new prompt detectors can fire on normal ready panes that merely show prior conversation text. [R4 / `packages/client/src/handlers/claude-code-tui/tmux-session.ts:236`]

`waitForReady` captures the visible pane, not a semantic modal state. On resume, a perfectly ready TUI can still show previous user/assistant text above the input prompt. The new `isClaudeLoginWall` matcher returns true for any visible `run /login`, `Login with Claude account`, or OAuth text, and it runs before the ready-surface check. That means a normal ready session whose visible transcript discusses login remediation can throw `ClaudeTuiLoginRequiredError`; SessionManager then classifies it permanent, marks the session errored, drains the inbox, and stops retrying. The bypass warning path has the same shape: visible prior text containing `Bypass Permissions mode` and `Yes, I accept` can trigger `1` + Enter before the ready check, injecting a stray user turn into an otherwise ready session.

Please make these detections distinguish the current startup modal from transcript content before taking irreversible action. For example, check the ready surface before modal handling when the ready marker + prompt are present, or require a stricter current-screen prompt shape that cannot be satisfied by ordinary transcript text. Add regression tests for a ready pane that includes prior visible text mentioning `run /login` and `Bypass Permissions mode`.

**Executed**
- Submitted request_changes on the synchronized head.
- Did not run local tests or checks, per PR Review mode; I reviewed the synchronized diff and surrounding call sites statically.

yuezengwu:
Addressed the `request_changes` in commit `af21112`. The finding was correct — on resume the captured pane re-renders prior transcript, so the new detectors could false-fire on a healthy session (stray keystroke, or a false *permanent* login failure). Fixed two ways:

1. **Ready-surface check moved FIRST** in `waitForReady`'s poll loop. A loaded session always shows the marker + `❯` prompt, so we now return before any modal/login handling — transcript text quoting `run /login` or `Bypass Permissions mode` can no longer trigger a keystroke or a false permanent failure.
2. **Tightened detectors** to live-modal-only shapes: `isClaudeLoginWall` now requires the interactive login-method selector (title + an option label) rather than loose `run /login`/OAuth text; `isBypassPermissionsWarning` requires the full warning title + **both** option labels. Under-detecting (fall back to the normal retry/timeout path) is the safe direction in the pre-ready window.

Regression tests added via a `capture`/`send` injection seam on `waitForReady`:
- a ready pane that quotes **every** modal/login string → resolves with no keystroke and no throw (the test also asserts the detectors *would* fire on that pane, proving the ready-first ordering is load-bearing);
- a live (non-ready) login selector → rejects with `ClaudeTuiLoginRequiredError`, no keystroke;
- a live bypass modal → accepted via `"1"`+Enter.

70 client unit tests pass; typecheck + Biome clean.

baixiaohang (approved):
## Review by codex-assistant - PR #1096 follow-up 3

**Recommendation**: approve
- Reason: the latest head resolves the prior request-changes finding by making ready detection win before any modal/login handling, tightening the live prompt detectors, and adding focused regression coverage for the transcript false-positive case.

**Level**: A
- Path initial: includes `packages/client/**` -> A
- Semantic lift: none; already A, and this still touches TUI startup/session error taxonomy.

**PR Summary**
- Latest head: `af211123f707c37abf5b4c20f2cb3c87a8140a61`
- Problem: detached `claude-code-tui` startup can block on pre-ready interactive prompts, then retry forever when the failure is classified as unknown/transient.
- Current design: `waitForReady` handles known answerable startup prompts, throws a permanent `ClaudeTuiLoginRequiredError` for the live login selector, checks the ready surface first to avoid transcript false positives, and keeps prompt matching NBSP-normalized.
- Affected modules: `packages/client/src/handlers/claude-code-tui/tmux-session.ts`, `packages/client/src/runtime/error-taxonomy.ts`, and tests.

**Review Comments**
✅ 1. Prior blocker resolved: a ready pane that quotes login/bypass prompt text now returns before any detector can send keys or throw a permanent login error.
✅ 2. The detectors are now intentionally narrower live-screen shapes, and the tests cover both the ready transcript case and the live login/bypass cases through the new `capture`/`send` seam.
✅ 3. The taxonomy path remains specific: only `ClaudeTuiLoginRequiredError` becomes permanent `claude_login_required`; ordinary unknown startup failures still follow the existing timeout/retry behavior.
⚠️ 4. The implementation deliberately does not classify loose `run /login` / OAuth text as a login wall; that is the right tradeoff here to avoid permanent false positives from transcript text.

**Executed**
- Approved after baixiaohang authorization for the current head.
- Did not run local tests or checks, per PR Review mode; I reviewed the latest synchronized diff and surrounding call sites statically.

PR #1105, opened 2026-06-16
feat(web): restructure chat right sidebar as an inspection panel

Gandy2025:
## What

Restructures the chat right sidebar (`ChatRightSidebar`) around the
workspace-detail product model — an inspection + light-control surface for the
selected work — **without adding modules or backend**. Same three modules
(Participants / Summary / GitHub), reorganized and re-presented.

## Changes

- **Section order**: Participants → Summary → GitHub (the live roster first; the
  read-only summary no longer occupies the most valuable top slot).
- **Participants**: agents-first; show the first 5 with a "Show all" fold;
  tighter row padding via a new `compact` prop on `AgentStatusPanel` (the
  compose status bar keeps the roomier default).
- **Summary** (renamed from "Description" — it's a live, continuously-rewritten
  summary read by whoever is reconstructing context, human or agent):
  **dynamic height** — capped at 30rem with a fade + "Show more" only when the
  GitHub section renders below it; uncapped (full body) when Summary is the last
  visible section (the common no-bindings case).
- **GitHub**: cluster rows by type (icon only: PR → Issue → Discussion →
  Commit); drop the bound-via provenance caption.
- **Width**: the inline rail is drag-resizable (left-edge handle + keyboard +
  double-click reset), default 360px, persisted globally, clamped to
  `[300, 600]`; the narrow-viewport overlay is unchanged.

## Scope / safety

Pure `packages/web`; no backend or shared-type changes.

## Tests

- New unit tests: roster partition (`partitionRoster`), GitHub type sort
  (`sortEntitiesByType`).
- Updated the right-rail DOM test for the Summary rename.
- `biome` + `typecheck` + design-token guardrail + affected suites green (33).

## Review

Self-reviewed with a high-effort multi-finder pass. One real bug found and
fixed in-branch: a self-added viewport-half width clamp diverged from the drag
handle's clamp, causing a mid-drag freeze and width-preference loss on smaller
windows — removed it, so the handle and parent now share one `[300, 600]`
clamp.

## Follow-up (not in this PR)

The resize handle mirrors the proven pattern in `doc-preview-drawer.tsx`
(~120 lines). Extracting a shared `useResizableWidth` hook would touch that
shipped component; left as a separate cleanup.

## Note

`main` advanced with unrelated server/CLI work after this branch's base; a
rebase may be needed before merge if `agent-status-panel.tsx` / `chat-view.tsx`
conflict.

### DIFF
--- a/packages/web/src/pages/workspace/right-sidebar/index.tsx
+++ b/packages/web/src/pages/workspace/right-sidebar/index.tsx
@@ -135,8 +135,14 @@ export function ChatRightSidebar({
           onReset={handleReset}
         />
       )}
+      {/* Key the sections that hold local fold state (Participants' "Show all",
+          Summary's "Show more") by chatId. ChatView is NOT remounted on chat
+          switch — it refetches by chatId — so without this the expanded/showAll
+          UI state would leak from one chat into the next. The rail shell itself
+          is intentionally NOT keyed: its width is a global preference. */}
       <div className="flex-1 overflow-y-auto">
         <ParticipantsSection
+          key={chatId}
           chatId={chatId}
           participants={participants}
           participantsLoading={participantsLoading}
@@ -144,7 +150,7 @@ export function ChatRightSidebar({
           onAdded={onAdded}
           readOnly={readOnly}
         />
-        <DescriptionSection description={description} capped={summaryCapped} />
+        <DescriptionSection key={chatId} description={description} capped={summaryCapped} />
         <GitHubSection chatId={chatId} />
       </div>
     </aside>
--- a/packages/web/src/pages/workspace/right-sidebar/index.tsx
+++ b/packages/web/src/pages/workspace/right-sidebar/index.tsx
@@ -1,27 +1,76 @@
 import type { ChatParticipantDetail } from "@first-tree/shared";
+import { useCallback, useState } from "react";
 import { DescriptionSection } from "./description-section.js";
-import { GitHubSection } from "./github-section.js";
+import { GitHubSection, useChatGithubEntities } from "./github-section.js";
 import { ParticipantsSection } from "./participants-section.js";
+import { SidebarResizeHandle } from "./resize-handle.js";
+
+/** Resizable inline-rail width (px). Default is a touch wider than the legacy
+ *  320 so markdown descriptions breathe; the user can drag wider for long
+ *  content. Persisted globally (a reading preference, not per-chat). */
+const WIDTH_STORAGE_KEY = "first-tree:chat-right-sidebar:width:v1";
+const DEFAULT_WIDTH = 360;
+const MIN_WIDTH = 300;
+const MAX_WIDTH = 600;
+
+/** Clamp to the fixed [MIN, MAX] band. A single source of clamping shared with
+ *  the drag handle (which is handed the same MIN/MAX), so the live drag value,
+ *  the committed value, and the persisted value never disagree. The center
+ *  conversation stays usable because the shell switches to the overlay variant
+ *  below the narrow breakpoint; in inline mode the user owns the trade and can
+ *  double-click the handle to reset. */
+function clampWidth(width: number): number {
+  return Math.min(Math.max(Math.round(width), MIN_WIDTH), MAX_WIDTH);
+}
+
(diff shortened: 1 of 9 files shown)


packages/web/src/pages/workspace/right-sidebar/index.tsx
```diff
@@ -44,23 +93,49 @@ export function ChatRightSidebar({
...
+          min={MIN_WIDTH}
+          max={MAX_WIDTH}
+          onWidthChange={handleWidthChange}
+          onCommit={handleCommit}
+          onReset={handleReset}
+        />
+      )}
       <div className="flex-1 overflow-y-auto">
-        <DescriptionSection description={description} />
         <ParticipantsSection
```
yuezengwu, line 144: Non-blocking: these child sections now own local fold state (`ParticipantsSection.showAll`, `DescriptionSection.expanded`), but `ChatView` is explicitly not remounted on `chatId` changes (see the existing DOM regression test around `chat-view-dom.test.tsx:1319`). That means `Show all` / `Show more` carries into the next selected chat. For example, expand a long GitHub-bound Summary, switch to another GitHub-bound chat, and the next Summary starts expanded, pushing the GitHub section down despite the default cap. Consider keying the stateful sections by `chatId` here, or resetting `expanded`/`showAll` when the selected chat changes.

yuezengwu (approved):
Reviewed against the workspace-detail sidebar model.

Implementation goal/core change: this keeps the existing Participants / Summary / GitHub modules, but reorders them around selected-work inspection, adds a desktop-only resizable right rail, makes Summary height conditional on whether GitHub content sits below it, and clusters GitHub bindings by entity type.

No backend, shared type, database, or core data-structure changes found in the diff. I left one non-blocking inline note about per-chat reset for local expand/fold state. This does not block merge from my side.

Gandy2025:
Thanks @yuezengwu. Addressed the per-chat fold-state note in 9d0dc07a: Participants and Summary are now keyed by `chatId` so their local *Show all* / *Show more* state resets on chat switch (ChatView isn't remounted on switch, so that UI state was leaking from one chat into the next). The rail shell stays unkeyed since its width is a global preference.

baixiaohang (approved):
## Review by codex-assistant - PR #1105

**Recommendation**: approve
- Reason: The change stays inside the existing web right-sidebar surface, matches the workspace-detail inspection model, and I found no blocking regression.

**Level**: C
- Semantic lift: none. I did not find subscription, WS frame parsing, data-layer, backend, or shared-type changes.

**PR Summary**
- Problem: Reworks chat details into a selected-work inspection panel without adding backend or modules.
- Design: Reorders Participants / Summary / GitHub, adds roster and GitHub sort helpers, makes Summary height conditional on GitHub presence, and adds a persisted desktop resize handle.
- Affected modules: `packages/web/src/pages/workspace/right-sidebar/**`, `packages/web/src/components/chat/agent-status-panel.tsx`, and the chat-view DOM test.

**Review Comments**
⚠️ 1. Non-blocking follow-up already covered by yuezengwu's inline note: `DescriptionSection.expanded` and `ParticipantsSection.showAll` can persist across selected-chat changes because `ChatView` is intentionally not remounted. I agree this is worth tightening, ideally by keying/resetting the stateful sections by `chatId`, but I do not consider it merge-blocking because it only affects user-expanded presentation state and remains manually reversible. [R4 / `packages/web/src/pages/workspace/right-sidebar/description-section.tsx:49`, `packages/web/src/pages/workspace/right-sidebar/participants-section.tsx:79`]
✅ 2. The GitHub section remains a read-only local projection and does not clone GitHub review/issue management into Workspace.
✅ 3. The width persistence path uses one fixed clamp for live, committed, and restored values; the narrow overlay remains fixed and unresizable as before.

**Performed**
- Approved.
- Did not run local tests/checks per PR-review rules; reviewed diff, relevant tree context, existing tests, and current CI/check status.

baixiaohang:
Follow-up after `9d0dc07a`: verified that the per-chat fold-state note is addressed by keying `ParticipantsSection` and `DescriptionSection` by `chatId` while leaving the rail shell unkeyed for the global width preference.

My approve still stands.

PR #1119, opened 2026-06-17
fix: name chat completion undici spans by route

baixiaohang:
## Summary
- add an undici request hook that renames known OpenAI-compatible chat completions client spans to `METHOD path`
- keep host, full URL, and query data in span attributes rather than span names
- cover OpenRouter `/api/v1/chat/completions`, OpenAI-style `/v1/chat/completions`, malformed requests, unrelated requests, and arbitrary dynamic path prefixes

## Verification
- `pnpm exec vitest run src/observability/__tests__/otel-init.test.ts --maxWorkers=2 --minWorkers=1`
- `pnpm exec biome check packages/server/src/observability/logfire-init.ts packages/server/src/observability/__tests__/otel-init.test.ts`
- `git diff --check`

### DIFF
--- a/packages/server/src/observability/__tests__/otel-init.test.ts
+++ b/packages/server/src/observability/__tests__/otel-init.test.ts
@@ -1,5 +1,12 @@
 import { describe, expect, it } from "vitest";
-import { initTelemetry, isTelemetryEnabled, parseHeaderString, shutdownTelemetry } from "../logfire-init.js";
+import {
+  initTelemetry,
+  isTelemetryEnabled,
+  parseHeaderString,
+  shutdownTelemetry,
+  undiciSpanNameForRequest,
+  updateKnownUndiciSpanName,
+} from "../logfire-init.js";
 import { normalizeAttrs } from "../otel-helpers.js";
 
 describe("normalizeAttrs", () => {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed and approved.

Goal/core change: this PR renames only known OpenAI-compatible chat completions undici client spans to `METHOD path`, while keeping host/full URL/query in the existing undici span attributes instead of cardinality-heavy span names.

What I checked:
- `@opentelemetry/instrumentation-undici@0.25.0` calls `requestHook(span, request)`, so `span.updateName(...)` is wired on the right hook.
- `@pydantic/logfire-node@0.14.0` passes `nodeAutoInstrumentations` through to `getNodeAutoInstrumentations(...)`.
- Route matching covers `/api/v1/chat/completions`, `/v1/chat/completions`, trailing slash normalization, malformed request data, unrelated requests, and dynamic prefixes that should not be renamed.
- No DB change or core data-structure change in this PR.

Local verification:
- `pnpm exec biome check packages/server/src/observability/logfire-init.ts packages/server/src/observability/__tests__/otel-init.test.ts` passed.
- `pnpm --filter @first-tree/server typecheck` passed.
- The exact targeted Vitest command is blocked in this local environment because server test global setup requires testcontainers/Postgres and Docker daemon is not running. I separately ran direct assertions for the added undici span naming cases via `node --import tsx`, and they passed.

PR #1121, opened 2026-06-17
feat(daemon): refresh runtime capabilities while connected

### LINKED ISSUE #1118: Runtime provider readiness does not refresh after install while daemon stays connected
## Summary

Runtime-provider readiness can stay stale after a user installs or logs into a provider while the First Tree daemon is already running. The user-facing setup flow promises that installing the engine makes it appear as ready, but the current daemon only uploads a fresh capability snapshot at startup and on WebSocket reconnect. If the daemon stays connected, the server/web UI can continue to show "no runtime ready" until the user manually runs `first-tree daemon probe`, restarts the daemon, or happens to reconnect.

## Current findings

Checked latest `origin/main` around `a7752102`.

Current refresh points:

- `apps/cli/src/commands/daemon/start.ts` probes capabilities before starting the runtime, then uploads after WS registration.
- The same file re-probes on `runtime.onReconnect(...)`, so reconnects partially mitigate stale capability state.
- `apps/cli/src/commands/daemon/probe.ts` provides a manual refresh command and explicitly documents the operator workaround.
- `packages/client/src/runtime/capabilities/index.ts` has reusable `probeCapabilities`, `revalidateCapabilities`, and `reprobeOnReconnect` helpers with TTL/backoff-aware behavior.

Remaining gap:

- Installing/logging into a runtime provider does not itself trigger a capability refresh while the daemon remains connected.
- A client that was previously `missing` / `unauthenticated` can remain setup-incomplete indefinitely from the web user's perspective.
- The user can be blocked at first run even though the provider is now installed and usable locally.

## User impact

From the user perspective:

1. They connect a client and see no ready runtime provider.
2. They follow the setup instructions to install/log into Claude Code or Codex.
3. Nothing changes in the web UI while the daemon keeps running normally.
4. Agent setup or first run still looks blocked until they discover a manual refresh/restart workaround.

This makes the quickstart promise feel broken: the runtime is installed locally, but First Tree does not notice it in bounded time.

## Proposed fix

Add an automatic daemon-side capability refresh path that does not require daemon restart or WS reconnect.

Possible implementation shape:

- Reuse `reprobeOnReconnect` / `revalidateCapabilities` rather than adding a second probing model.
- When any provider is non-`ok`, schedule a bounded, low-frequency re-probe with backoff and upload only changed snapshots.
- Keep full launch smokes guarded by TTL/backoff so already-`ok` providers are not repeatedly re-smoked.
- Consider exposing an explicit web/API "refresh runtimes" action if automatic polling is too expensive, but the default setup path should still have a clear bounded refresh story.
- Update quickstart/setup copy if manual refresh remains part of the product contract.

## Acceptance criteria

- If a user installs/logs into a supported runtime while the daemon remains connected, the client capability state refreshes within a bounded time or via an explicit first-class refresh action surfaced in setup UI.
- The server/web UI stops showing stale "no runtime ready" once the provider is actually usable.
- The implementation avoids overlapping probes and avoids expensive full smoke probes on every interval for already-healthy providers.
- Existing reconnect re-probe behavior remains covered.
- Docs/setup copy matches the actual refresh behavior.

Gandy2025:
## Problem

Closes #1118.

Runtime-provider readiness could stay stale after a user installs or logs into a provider while the daemon is already connected. The daemon only refreshed its capability snapshot at **startup** and on **WS reconnect**. With a connected daemon, installing/logging into Claude Code or Codex went unnoticed, so the server/web UI could keep showing **"no runtime ready"** until the user ran `daemon probe`, restarted, or happened to reconnect.

## Fix

Unify the refresh triggers behind a single `CapabilityRefresher` (`apps/cli/src/core/capability-refresh.ts`) with **one shared in-flight guard**, so a reconnect re-probe and the new background poll never launch providers concurrently:

- **Reconnect re-probe** reuses `reprobeOnReconnect` — behavior unchanged (TTL-aware full-vs-revalidate). A reconnect that lands while a refresh is already running is **held and drained in reconnect mode** afterward (never dropped); when it coincides with a poll, the stricter reconnect TTL/full path wins.
- **New bounded background poll** runs *while the daemon stays connected* whenever a refresh is still warranted:
  - starts ~15s after the degraded state is observed, exponential backoff to a **5-min ceiling**;
  - re-probes **only the non-`ok` providers** via `revalidateCapabilities` (already-`ok` providers stay on the free cached resolve+auth path — no repeated full smokes);
  - uploads **only when the snapshot changes** (deduped);
  - **stops only once every provider is `ok` AND that healthy snapshot is confirmed uploaded** — a recovery whose PATCH failed keeps the poll alive (and resets the backoff) so the server can't be left on the stale degraded snapshot;
  - a permanently-missing provider (e.g. `claude-code-tui` on a no-tmux box) settles into the cheap 5-min poll, where its resolve stage short-circuits without a launch.

No server/web changes — the web "no runtime ready" gate clears automatically once the next changed snapshot uploads. The optional explicit web "refresh runtimes" button is deferred (the issue lists it as a maybe; the bounded auto-poll satisfies the default-path acceptance criteria).

## Acceptance criteria

- [x] Install/login while connected refreshes within a bounded time (no restart/reconnect needed).
- [x] Server/web stops showing stale "no runtime ready" once the provider is usable (poll only stops after the healthy snapshot is confirmed uploaded).
- [x] No overlapping probes (shared in-flight guard); no expensive full smoke on every interval for healthy providers (revalidate keeps `ok` cached).
- [x] Existing reconnect re-probe behavior preserved (a reconnect coinciding with a poll is deferred, then run in reconnect mode — never dropped).
- [x] Docs/setup copy updated (`cli-reference.md`, `daemon probe` help).

## Tests

New pure-helper tests (`hasNonOkProvider`, `nextCapabilityRefreshDelayMs`) and a behavioral `capability-refresh.test.ts` covering: poll → recover → stop, upload dedupe, backoff growth/reset, a reconnect deferred-then-drained mid-poll, recovery-but-upload-fails → keeps polling → stops once synced, the reconnect path, and null-snapshot recovery. Existing `daemon-start-command.test.ts` updated for the refactor.

Verified locally: `pnpm typecheck` clean, CLI 491/491 + client 1036/1036 green, `pnpm check` clean.

## Context Tree

Paired tree PR updates the Capabilities Probe refresh-timing note in `system/cloud/runtime/client-runtime.md`: agent-team-foundation/first-tree-context#568 (https://github.com/agent-team-foundation/first-tree-context/pull/568).

### DIFF
--- a/apps/cli/src/core/capability-refresh.ts
+++ b/apps/cli/src/core/capability-refresh.ts
@@ -69,6 +69,10 @@ export class CapabilityRefresher {
   private snapshot: ClientCapabilities | null;
   private lastUploadedJson: string | null = null;
   private inFlight = false;
+  /** A reconnect that landed while a refresh was in flight, to be drained (in
+   * reconnect mode) once the current refresh finishes — never dropped, so the
+   * reconnect TTL/full re-probe path is always honored. */
+  private pendingReconnect = false;
   private timer: ReturnType<typeof setTimeout> | null = null;
   /** Consecutive polls with no observed state change — drives the backoff. */
   private idleAttempts = 0;
@@ -84,9 +88,10 @@ export class CapabilityRefresher {
   }
 
   /**
-   * Push the startup snapshot to the server (deduped) and, if any provider is
-   * not yet `ok`, arm the background poll. Best-effort: an upload failure is
-   * logged and the poll is still armed (a later poll re-tries the upload).
+   * Push the startup snapshot to the server (deduped) and arm the background
+   * poll if a refresh is still warranted (a non-`ok` provider, or a snapshot
+   * the server has not confirmed). Best-effort: an upload failure is logged and
+   * the poll is still armed (a later poll re-tries the upload).
    */
   async start(): Promise<void> {
     if (this.snapshot) {
@@ -103,9 +108,15 @@ export class CapabilityRefresher {
   /**
    * Re-probe after a WS reconnect (full or revalidate per the TTL policy), then
    * re-arm or stop the poll based on the fresh snapshot. Fire-and-forget; never
-   * throws into the connection. No-op if a refresh is already in flight.
+   * throws into the connection.
+   *
+   * If a refresh is already in flight (e.g. a degraded-state poll), the
+   * reconnect is recorded as pending and drained in reconnect mode once that
+   * refresh finishes — it is never dropped, so the reconnect's TTL/full
+   * re-probe always runs even when it coincides with a poll.
    */
   onReconnect(): void {
+    this.pendingReconnect = true;
     void this.runRefresh("reconnect");
   }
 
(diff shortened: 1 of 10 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1121

**建议**: request_changes
- 理由: 当前 in-flight guard 会吞掉 reconnect 触发，导致既有 reconnect re-probe 语义在 poll 竞态下不再保证执行。

**分级**: A
- Path 初判: 含 `apps/cli/**` 和 `packages/client/**` -> A
- 语义升级: 无

**PR 摘要**
- 解决问题: daemon 保持连接时，用户安装或登录 runtime provider 后 capabilities 不会刷新，Web 可能持续显示 no runtime ready。
- 设计方案: 新增 `CapabilityRefresher` 统一 startup upload、WS reconnect re-probe 和 degraded background poll，并在 client capability helpers 中加入 degraded 判断与 backoff。
- 受影响模块: `apps/cli/src/core`, `apps/cli/src/commands/daemon`, `packages/client/src/runtime/capabilities`, CLI docs/tests

**Review 意见**
❌ 1. `onReconnect()` 在 refresh 已经 in-flight 时会直接丢掉 reconnect 事件，破坏“existing reconnect re-probe behavior preserved”的语义。`runRefresh()` 开头用 `if (this.inFlight || this.stopped) return` 硬返回；而 `runtime.onReconnect()` 只是同步调用 `capabilityRefresher.onReconnect()`，没有后续重试。因此只要 degraded poll 正在跑时 WS 完成 re-registration，本次 reconnect 应有的 `reprobeOnReconnect()` 就永久不会执行。具体后果是：一个已有 24h+ stale `ok` snapshot 的 daemon 如果因为某个 provider non-ok 正在 poll，恰好同时 reconnect，本来 reconnect 路径应 full re-probe 所有 provider；现在会被 poll 的 cheaper `revalidate()` 覆盖，且 poll 完成后只继续按 poll 调度，不会补跑 reconnect/full TTL 检查。建议把 reconnect 记为 pending，并在当前 refresh 完成后以 reconnect 模式补跑，或让 in-flight 合并策略选择更严格的 reconnect/full 行为，而不是 drop。 [R4 / `apps/cli/src/core/capability-refresh.ts:133`]

**已执行**
- 已 request_changes

apps/cli/src/core/capability-refresh.ts
```diff
@@ -0,0 +1,196 @@
...
+      if (trigger === "reconnect") {
+        const { capabilities, mode } = await this.reprobe(previous);
+        next = capabilities;
+        modeLabel = `reconnect, ${mode}`;
+      } else {
+        next = await this.revalidate(previous);
+        modeLabel = "poll";
+      }
+      const changed = stableCapabilitiesJson(next) !== stableCapabilitiesJson(previous);
+      this.snapshot = next;
```
yuezengwu: This updates the local snapshot before `uploadIfChanged` succeeds. If a poll detects recovery (`missing`/`unauthenticated` -> all `ok`) but the PATCH fails transiently, the `catch` path calls `scheduleNext()` with `this.snapshot` already set to the all-`ok` value. `hasNonOkProvider` then returns false, the poll disarms, and the server/web keep the old degraded snapshot until a reconnect/manual probe/restart. That is the stale "no runtime ready" path this PR is meant to close. Please keep retry state tied to the last successfully uploaded snapshot, or otherwise keep polling/retrying until the recovered snapshot is actually uploaded; add a test for recovery + upload failure.

yuezengwu (changes requested):
Reviewed PR #1121.

Goal: refresh runtime-provider readiness while the daemon stays connected, so installing or logging into Claude Code/Codex clears stale "no runtime ready" state without restart/reconnect/manual probe.

Core changes: `CapabilityRefresher` centralizes startup upload, reconnect re-probe, and the new degraded-state background poll behind one in-flight guard; the client probe helpers now expose the non-ok detector/backoff policy; docs and focused tests were updated for the new refresh timing.

I am requesting changes for one correctness issue: a recovered all-`ok` snapshot can stop the poll even when the upload that should clear the server/web stale state failed. That leaves the server on the old degraded snapshot until some unrelated reconnect/manual probe/restart, which breaks the main acceptance path under a transient PATCH/auth failure.

Gandy2025:
Both review findings were correct; fixed in 6d75157d.

**@codex-assistant (R4 — dropped reconnect):** confirmed — `runRefresh`'s `if (this.inFlight) return` dropped a reconnect that coincided with a degraded-state poll, so its TTL/full re-probe never ran. A reconnect is now held in `pendingReconnect` and **drained in reconnect mode** once the current refresh finishes — never dropped. When a reconnect coincides with a poll, the run's `effective` trigger upgrades to `reconnect`, so the stricter TTL/full path wins instead of the poll's `revalidate`.

**@yuezengwu (recovered-but-upload-failed stops the poll):** confirmed — the snapshot was set to the recovered all-`ok` value before the PATCH, and a failed upload landed in the catch that still called `scheduleNext`, stopping the poll with the server left on the stale degraded snapshot. Upload success is now tracked separately from the probe (`needsRefresh()` keeps the poll alive while `stableJson(snapshot) !== lastUploadedJson`), so a healthy-but-unsynced snapshot keeps re-attempting the upload; a failed upload also resets the backoff so the retry is prompt. The poll only stops once the server actually reflects readiness.

New tests cover both: a reconnect deferred-then-drained mid-poll (reconnect re-probe never dropped), and poll-recovers-but-upload-fails → keeps polling → stops once synced. CLI 491/491 + client 1036/1036 green, typecheck + biome clean.

yuezengwu (approved):
Re-reviewed after 6d75157d.

Goal: keep the daemon's advertised runtime capabilities fresh while it remains connected, so provider install/login clears stale setup state without restart/reconnect/manual probe.

Core changes are now sound: `CapabilityRefresher` owns startup upload, degraded-state polling, reconnect re-probe, shared in-flight serialization, and upload dedupe; reconnects that arrive during a poll are retained and drained in reconnect mode; healthy-but-unsynced snapshots keep polling until the server upload succeeds.

The two previous blockers are addressed, with targeted tests for reconnect deferral/drain and recovery upload failure retry. I did not find new blocking issues in this pass.

baixiaohang (approved):
Re-reviewed after `6d75157d` and the follow-up responses. The previous blockers are addressed: reconnects that arrive during an in-flight poll are retained and drained in reconnect mode, and recovered-but-unsynced healthy snapshots keep polling until the capability upload succeeds. No new blocking issues found in this pass.

PR #1128, opened 2026-06-17
fix(chat ask): skip-as-answer + instant-in-chat takeover hardening

yuezengwu:
Fixes the `chat ask` takeover interaction reported by @yuezengwu, plus two latent realtime hazards found while auditing why the takeover sometimes didn't pop instantly. Three independent fixes, landed together as one reviewed change (QA validated as a unit).

## #2 — Skip is an answer, not a deferral (web)
Previously the takeover's **Skip** only dismissed the overlay for the session via a local `skippedRequestIds` set: no message was sent, so the open request, the red dot, and the asking agent's block all persisted. Skip now sends a resolving reply (`metadata.resolves` `kind="answered"`, body `"(Skipped — no answer provided.)"`) → the question resolves, the red dot clears, the asking agent unblocks, and the overlay closes. Removed the dead session state.

## H1 — Cross-replica chat-audience invalidation (server)
The push-audience cache is process-local, so a membership change on one replica left other replicas serving a stale audience — silently dropping `chat:message` pushes to a just-added member until the TTL aged out (intermittent "takeover didn't pop instantly", self-healing in ≤5s). Adds a `chat_audience_events` PG NOTIFY: `invalidateChatAudience` now drops the local entry **and** fans out so every replica invalidates; `invalidateChatAudienceLocal` (used by the listener) avoids a NOTIFY loop. No schema change.

## H2 — Open asks beyond the message window still block (server + web)
The blocking takeover derived only from the latest 50 loaded messages, so an open ask that scrolled past that window never blocked (and refresh didn't help). Adds a window-independent source: `listOpenRequestsForViewer` + `GET /:chatId/open-requests` on the **user-scope** chat router (same scope the web's `/messages` uses), scoped to the caller's own human-agent uuid; the web unions open requests into the blocking derivation, with WS + reconnect invalidation wired.

## Tests
- web DOM: Skip resolves (no dismiss); a buried open ask surfaces via the open-requests source.
- server: cross-audience fan-out unit test; user-scope `open-requests` HTTP e2e (route + open/resolved semantics).
- `pnpm typecheck` (server + web), `biome`, and the full server suite green.

## QA
Validated end-to-end on a real two-replica server + browser suite (skip resolve, cross-replica WS push, buried-ask takeover via user-scope `/open-requests`, mobile) — full server suite 1472 tests pass. A first QA pass caught an H2 404 (the `open-requests` route was initially on the agent scope, not the user scope the web calls); fixed and re-validated PASS.

No DB migration. No `version` bumps.

### DIFF
--- a/packages/server/src/__tests__/activity.test.ts
+++ b/packages/server/src/__tests__/activity.test.ts
@@ -19,6 +19,7 @@ function makeNotifier(): Notifier {
     notifyRuntimeStateChange: vi.fn(async () => {}),
     notifySessionRuntime: vi.fn(async () => {}),
     notifyChatMessage: vi.fn(async () => {}),
+    notifyChatAudience: vi.fn(async () => {}),
     notifySessionEvent: vi.fn(async () => {}),
     pushFrameToInbox: vi.fn(async () => 0),
     onConfigChange: vi.fn(),
@@ -27,6 +28,7 @@ function makeNotifier(): Notifier {
     onRuntimeStateChange: vi.fn(),
     onSessionRuntime: vi.fn(),
     onChatMessage: vi.fn(),
+    onChatAudience: vi.fn(),
     start: vi.fn(async () => {}),
     stop: vi.fn(async () => {}),
   } satisfies Notifier;
@@ -142,6 +144,7 @@ describe("upsertSessionState — touchPresenceLastSeen option", () => {
       notifyRuntimeStateChange: vi.fn(async () => {}),
       notifySessionRuntime: vi.fn(async () => {}),
       notifyChatMessage: vi.fn(async () => {}),
+      notifyChatAudience: vi.fn(async () => {}),
       notifySessionEvent: vi.fn(async () => {}),
       pushFrameToInbox: vi.fn(async () => 0),
       onConfigChange: vi.fn(),
@@ -150,6 +153,7 @@ describe("upsertSessionState — touchPresenceLastSeen option", () => {
       onRuntimeStateChange: vi.fn(),
       onSessionRuntime: vi.fn(),
       onChatMessage: vi.fn(),
+      onChatAudience: vi.fn(),
       start: vi.fn(async () => {}),
       stop: vi.fn(async () => {}),
     } satisfies Notifier;
@@ -180,6 +184,7 @@ describe("upsertSessionState — touchPresenceLastSeen option", () => {
       notifyRuntimeStateChange: vi.fn(async () => {}),
       notifySessionRuntime: vi.fn(async () => {}),
       notifyChatMessage: vi.fn(async () => {}),
+      notifyChatAudience: vi.fn(async () => {}),
       notifySessionEvent: vi.fn(async () => {}),
       pushFrameToInbox: vi.fn(async () => 0),
       onConfigChange: vi.fn(),
@@ -188,6 +193,7 @@ describe("upsertSessionState — touchPresenceLastSeen option", () => {
       onRuntimeStateChange: vi.fn(),
       onSessionRuntime: vi.fn(),
       onChatMessage: vi.fn(),
+      onChatAudience: vi.fn(),
       start: vi.fn(async () => {}),
       stop: vi.fn(async () => {}),
     } satisfies Notifier;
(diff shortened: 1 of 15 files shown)


yuezengwu:
Context-tree PR (request-lifecycle node update for the new Skip-as-answer semantics): agent-team-foundation/first-tree-context#569

yuezengwu (commented):
Reviewed PR #1128.

Goal and core changes:
- `chat ask` takeover Skip now sends a resolving answer instead of doing a session-only dismiss, so the open request, red dot, and asking-agent block clear.
- Server adds PG NOTIFY-backed cross-replica invalidation for the chat audience cache, so membership changes do not leave another replica with a stale `chat:message` push audience.
- Server/web add a window-independent open-requests source on the user-scope chat route, and the web unions it into takeover derivation with WS/reconnect invalidation.

Findings: no code blockers found. I checked the critical seams: resolution authz remains target-human-only in `sendMessage`; the open-request query's asker resolver scope matches the existing legacy-compat reader/idempotency scope; the new route is on the user chat router and scopes results to the caller's `humanAgentId`; and the web optimistic resolving row prevents stale `open-requests` data from re-pinning the overlay while POST/refetch completes.

Verification run locally after installing deps in a fresh review worktree:
- `pnpm --filter @first-tree/server test -- src/__tests__/chat-audience-cache.test.ts src/__tests__/chat-open-requests-route.test.ts` — this ran the full server suite: 164 files / 1472 tests passed.
- `pnpm --filter @first-tree/web test -- src/pages/workspace/center/__tests__/chat-view-dom.test.tsx -t "surfaces a buried open ask|Skip resolves"` — this ran the full web suite: 109 files / 808 tests passed. There is existing stderr noise around localhost:3000 connection attempts and Radix aria warnings, but no failures.

Context Tree follow-up: this intentionally changes the durable Web Answering Surface semantics. The current tree node `system/cloud/chat/messaging.md` still says dismissing without answering hides the card for the session while the open request persists. Please land/link a paired Context Tree update before merge or shortly after so the tree does not contradict the shipped behavior.

yuezengwu:
Thanks for the review. The paired Context Tree update is already open **and approved**: agent-team-foundation/first-tree-context#569. It rewrites the `system/cloud/chat/messaging.md` Web Answering Surface so **Skip is a resolving "skipped" answer** (clears the open-request count, unblocks the asking agent) and drops the old "dismiss hides for the session, returns next visit" wording — exactly the contradiction you flagged. `first-tree tree verify` passes and it's cross-linked above, so the tree won't trail the shipped behavior. Ready to land together / right after this PR.

baixiaohang (approved):
## Review by codex-assistant - PR #1128

**Recommendation**: approve
- Reason: the three fixes align with the existing chat request lifecycle, membership-scoped realtime delivery, and web takeover derivation contracts; I found no blocking issues.

**Level**: B-high
- Path initial: touches `packages/server/src/services/message.ts` (`services/message*`) -> B-high.
- Semantic lift: none.

**PR Summary**
- Problem: `chat ask` Skip only dismissed locally, open asks outside the loaded message window could fail to block, and cross-replica stale chat-audience caches could delay `chat:message` pushes to newly-added viewers.
- Design: Skip sends an explicit resolving reply; server fans chat-audience invalidation through PG NOTIFY with a local-only listener path; server exposes user-scope open requests and web unions that source into the blocking takeover derivation with WS/reconnect invalidation.
- Affected modules: server chat/message/notifier/audience cache, web chat view/takeover/admin WS, focused server and web tests.

**Review Notes**
✅ 1. The open-requests query mirrors the request lifecycle rules, and the user-scope route scopes results to the caller's `humanAgentId`.
✅ 2. The web union/dedup path lets optimistic resolving replies immediately lift a buried ask, so stale open-request query data does not re-pin the overlay while POST/refetch completes.
✅ 3. Audience invalidation preserves the post-commit cache boundary and uses a local-only listener invalidation to avoid NOTIFY loops.
✅ 4. The paired Context Tree update is linked and approved in agent-team-foundation/first-tree-context#569, so the durable Web Answering Surface semantics will match this change.

**Executed**
- Submitted approve review after B-high authorization. I did not run local tests per review-mode rules; CI is green.

PR #1130, opened 2026-06-17
feat(web): persist onboarding repo selection as a per-org draft

Gandy2025:
## Problem

Repos picked on the onboarding **connect-code** step lived only in React flow state. The formal team-resource write happens later, at the **kickoff** step ("Build tree & start"). So any bailout before kickoff — the top-bar "I'll finish later", a refresh, or a mid-flow navigation — silently discarded the selection. A user who carefully narrowed to one repo, then left, came back to find their choice gone (and re-defaulted to "all granted").

This is the "I selected a repo during onboarding but it didn't save" report.

## Approach

Persist the selection as a **per-org draft in `sessionStorage`**, so the user resumes the wizard with the repos they chose. The formal team-resource write is unchanged — it still happens only at kickoff. This is purely an in-flight draft that survives a bailout.

Why `sessionStorage` (not `localStorage` / server): onboarding's whole resume model is already per-tab `sessionStorage` (the step index `onboarding:stepIndex` and the created-agent stash `onboarding:agentUuid`). On tab close the flow itself resets, so a more-durable repo draft would be a half-resume that buys nothing while adding cleanup/privacy/multi-tab concerns (localStorage) or backend weight (server). The draft's durability should match the surrounding resume model. If onboarding resume ever moves server-side, the draft should move with it (noted in code).

## Changes (frontend only)

- **`onboarding-flags.ts`** — per-org `read/writeOnboardingSelectedRepos` (`onboarding:selectedRepos:<org>`). `null` = no draft yet; `[]` = deliberately selected none. The connect-code step needs that distinction. Auto-swept by the existing `onboarding:*` logout cleanup.
- **`onboarding-flow.tsx`** — hydrate the draft per org, write through on every change, expose `hasRepoDraft`, clear the draft on successful completion (`completeAndEnterChat`); `finishLater` deliberately keeps it.
- **`step-connect-code.tsx`** — auto-select all granted repos only when there's **no** draft (so a resumed narrowing isn't clobbered back to "all"); and prune a restored draft against the current grant list for picker/UI hygiene.
- **`step-kickoff.tsx`** — authoritative grant validation on the write path. `handleStart` re-validates `selectedRepoUrls` against the current GitHub App grant list (`listOrgGithubRepos`) before building `orgWrites.sourceRepos`, covering the **resume-directly-at-kickoff** path (persisted step index) where `StepConnectCode` never mounts.
  - Ungranted repos are pruned; if the selection is empty after pruning, it degrades to the no-repo intro path.
  - **Always reads current grants** (`staleTime: 0`, not a cached list): the QueryClient is an app-level singleton and `finishLater` is SPA navigation, so a cached grant list can be stale relative to grants changed in another tab / GitHub settings. The check must prove *current* grant state, so it never trusts the picker cache here. The redundant read on the normal connect-code → kickoff path is the accepted cost of correctness.
  - **Fails closed**: if the grant list can't be read (`no_installation` / `suspended` / `not_configured` / upstream 5xx), kickoff surfaces a retryable error and writes **nothing** — it never registers an unvalidated selection. There is no downstream grant check (`createTeamResourceForOrg` only validates URL shape; existing-tree writes are best-effort). Clicking Start retries.

## Review

Iterated through several `request_changes` rounds; the stale-grant invariant — a restored draft carrying a now-inaccessible repo into the team-resource write — is now closed on **every** path:
- connect-code prune (UI / picker hygiene),
- kickoff write-path revalidation against current grants (authoritative, never cached),
- fail-closed when grants can't be read.

Server-side grant enforcement at `createTeamResourceForOrg` would be additional defense-in-depth, but the onboarding client is the only writer, so the invariant is closed here (can file a follow-up issue if wanted). Implementation-only; no Context Tree change (the "formal write stays at kickoff" decision is unchanged).

## Test plan

- Unit (`onboarding-flags`): round-trip, `null` vs `[]`, per-org isolation, logout sweep, malformed-value fallback.
- DOM (`connect-code`): no-draft → auto-select all; resumed draft → not overwritten; resumed draft with a no-longer-granted repo → pruned.
- DOM (`kickoff`): stale repo in selection → pruned, never registered; grant read fails → no resource write, no navigation (fail closed); **a stale cached grant list is ignored — the live read still prunes the removed repo** (guards the authoritative check).
- `pnpm typecheck` + Biome clean; onboarding-related suites green (114 tests).

### Manual (staging)
1. Onboarding → connect GitHub → narrow the repo picker to one.
2. Top-bar "I'll finish later".
3. Re-enter onboarding → connect-code should still show your narrowed selection (not re-checked to all).
4. Refresh mid-flow → same.
5. Finish kickoff → draft cleared; the repo appears in Settings → Resources (kickoff write unchanged).

### DIFF
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -761,6 +761,7 @@ function createFlowValue(overrides: Partial<OnboardingFlowValue> = {}): Onboardi
     hasAgent: true,
     selectedRepoUrls: ["https://github.com/acme/web.git"],
     setSelectedRepoUrls: () => undefined,
+    hasRepoDraft: true,
     treeMode: "existing",
     setTreeMode: () => undefined,
     treeUrl: "https://github.com/acme/context-tree",
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -1870,6 +1870,31 @@ describe("web DOM interaction coverage", () => {
     await unmountRoot(view.root);
   });
 
+  it("fails closed at kickoff when the grant list can't be read (no stale write)", async () => {
+    // If the current-grant read fails we can't prove the selected repos are still
+    // accessible and nothing downstream re-checks grants, so kickoff must NOT
+    // write the (possibly stale) selection — it surfaces a retryable error and
+    // stays on the form instead.
+    const { StepKickoff } = await import("../onboarding/steps/step-kickoff.js");
+    githubMocks.listOrgGithubRepos.mockRejectedValue(new Error("github unavailable"));
+    const view = await renderOnboardingDom(<StepKickoff />, {
+      activeStep: "kickoff",
+      selectedRepoUrls: ["https://github.com/acme/web.git"],
+      treeMode: "existing",
+      treeUrl: "https://github.com/acme/context-tree",
+    });
+    await waitForText("Your agent's ready to get to work", view.container);
+    await click(
+      ([...view.container.querySelectorAll("button")].find((b) => b.textContent?.includes("Start")) ??
+        null) as HTMLButtonElement | null,
+    );
+    await waitForText("Couldn't check your repositories", view.container);
+    expect(resourceMocks.createTeamResourceForOrg).not.toHaveBeenCalled();
+    expect(chatApiMocks.createAgentChat).not.toHaveBeenCalled();
+    expect(view.flow.completeAndEnterChat).not.toHaveBeenCalled();
+    await unmountRoot(view.root);
+  });
+
   it("edits MCP server rows through validation, stdio, and HTTP submissions", async () => {
     const { McpSection } = await import("../agent-detail/mcp-section.js");
     const onAdd = vi.fn();
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -484,11 +484,14 @@ function createFlowValue(overrides: Partial<OnboardingFlowValue> = {}): Onboardi
 async function renderOnboardingDom(
   element: ReactElement,
   overrides: Partial<OnboardingFlowValue> = {},
+  seed?: (queryClient: QueryClient) => void,
 ): Promise<{ container: HTMLElement; root: Root; flow: OnboardingFlowValue }> {
   const { OnboardingFlowContext } = await import("../onboarding/onboarding-flow.js");
   const flow = createFlowValue(overrides);
   const rendered = await renderDom(
     <OnboardingFlowContext.Provider value={flow}>{element}</OnboardingFlowContext.Provider>,
+    "/",
+    seed,
   );
   return { ...rendered, flow };
 }
@@ -1892,6 +1895,67 @@ describe("web DOM interaction coverage", () => {
     expect(resourceMocks.createTeamResourceForOrg).not.toHaveBeenCalled();
     expect(chatApiMocks.createAgentChat).not.toHaveBeenCalled();
     expect(view.flow.completeAndEnterChat).not.toHaveBeenCalled();
+    await unmountRoot(view.root);
+  });
+
+  it("kickoff grant check is authoritative — ignores a stale cached grant list", async () => {
+    // The QueryClient is an app-level singleton and finishLater is SPA nav, so a
+    // grant list connect-code cached earlier can still be in the cache when the
+    // user resumes at kickoff — possibly minutes stale. The write-path check
+    // must read CURRENT grants, not trust the cache (guards against re-adding a
+    // staleTime). Seed the cache with a stale list that still contains a repo
+    // that has since been removed; assert the live read prunes it anyway.
+    const { StepKickoff } = await import("../onboarding/steps/step-kickoff.js");
+    // Current grants (live) are web + api; the stale cache still has `gone`.
+    githubMocks.listOrgGithubRepos.mockResolvedValue(GITHUB_REPOS);
+    const view = await renderOnboardingDom(
+      <StepKickoff />,
+      {
+        activeStep: "kickoff",
+        selectedRepoUrls: ["https://github.com/acme/web.git", "https://github.com/acme/gone.git"],
+        treeMode: "existing",
+        treeUrl: "https://github.com/acme/context-tree",
+      },
+      (queryClient) => {
+        queryClient.setQueryData(
+          ["onboarding", "org-github-repos", "org-1"],
+          [
+            {
+              fullName: "acme/web",
+              cloneUrl: "https://github.com/acme/web.git",
+              htmlUrl: "",
+              private: false,
+              defaultBranch: "main",
+              pushedAt: NOW,
+            },
+            {
+              fullName: "acme/gone",
+              cloneUrl: "https://github.com/acme/gone.git",
+              htmlUrl: "",
+              private: false,
+              defaultBranch: "main",
+              pushedAt: NOW,
+            },
+          ],
+        );
+      },
+    );
+    await waitForText("Your agent's ready to get to work", view.container);
+    await click(
+      ([...view.container.querySelectorAll("button")].find((b) => b.textContent?.includes("Start")) ??
+        null) as HTMLButtonElement | null,
+    );
+    await waitForText("Starting your agent", view.container);
+    // Live read returned web + api → `gone` is pruned despite being in the cache.
+    expect(githubMocks.listOrgGithubRepos).toHaveBeenCalled();
+    expect(resourceMocks.createTeamResourceForOrg).toHaveBeenCalledWith(
+      "org-1",
+      expect.objectContaining({ payload: { url: "https://github.com/acme/web.git" } }),
+    );
+    expect(resourceMocks.createTeamResourceForOrg).not.toHaveBeenCalledWith(
+      "org-1",
+      expect.objectContaining({ payload: { url: "https://github.com/acme/gone.git" } }),
+    );
     await unmountRoot(view.root);
   });
 
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -468,6 +468,7 @@ function createFlowValue(overrides: Partial<OnboardingFlowValue> = {}): Onboardi
     hasAgent: true,
     selectedRepoUrls: ["https://github.com/acme/web.git"],
     setSelectedRepoUrls: vi.fn(),
+    hasRepoDraft: true,
     treeMode: "existing",
     setTreeMode: vi.fn(),
     treeUrl: "https://github.com/acme/context-tree",
@@ -1586,6 +1587,83 @@ describe("web DOM interaction coverage", () => {
       ) ?? null,
     );
     expect(notConfigured.flow.goNext).toHaveBeenCalled();
+  });
+
+  it("auto-selects all granted repos with no draft, but preserves a resumed draft", async () => {
+    const { StepConnectCode } = await import("../onboarding/steps/step-connect-code.js");
+    const connectedInstall = {
+      installationId: 42,
+      accountLogin: "acme",
+      accountType: "Organization" as const,
+      accountGithubId: 123,
+      repositorySelection: "selected" as const,
+      permissions: {},
+      events: [],
+      suspended: false,
+      manageUrl: "https://github.com/organizations/acme/settings/installations/42",
+      createdAt: NOW,
+      updatedAt: NOW,
+    };
+    githubAppMocks.getGithubAppInstallation.mockResolvedValue(connectedInstall);
+
+    // No saved draft (first visit) → the picker defaults to every granted repo,
+    // so the user doesn't re-pick what they just granted on GitHub.
+    const freshSet = vi.fn();
+    const noDraft = await renderOnboardingDom(<StepConnectCode />, {
+      activeStep: "connect-code",
+      selectedRepoUrls: [],
+      hasRepoDraft: false,
+      setSelectedRepoUrls: freshSet,
+    });
+    await waitForText("acme/web", noDraft.container);
+    await waitForCondition(
+      () =>
+        freshSet.mock.calls.some(
+          ([arg]) =>
+            Array.isArray(arg) &&
+            arg.length === 2 &&
+            arg.includes("https://github.com/acme/web.git") &&
+            arg.includes("git@github.com:acme/api.git"),
+        ),
+      "auto-select all granted repos when there's no draft",
+    );
+    await unmountRoot(noDraft.root);
+
+    // Resumed draft (user narrowed to just one repo earlier, then bailed before
+    // kickoff) → no auto-select; the saved selection is left untouched instead
+    // of being clobbered back to "all granted".
+    const resumedSet = vi.fn();
+    const withDraft = await renderOnboardingDom(<StepConnectCode />, {
+      activeStep: "connect-code",
+      selectedRepoUrls: ["https://github.com/acme/web.git"],
+      hasRepoDraft: true,
+      setSelectedRepoUrls: resumedSet,
+    });
+    await waitForText("acme/web", withDraft.container);
+    await flush();
+    expect(resumedSet).not.toHaveBeenCalled();
+    await unmountRoot(withDraft.root);
+
+    // Resumed draft carrying a repo the GitHub App no longer grants (uninstall /
+    // changed grant between bailout and resume) → prune it against the current
+    // grant list so a stale URL can't ride into kickoff. acme/web is still
+    // granted; the gone repo is dropped.
+    const prunedSet = vi.fn();
+    const staleDraft = await renderOnboardingDom(<StepConnectCode />, {
+      activeStep: "connect-code",
+      selectedRepoUrls: ["https://github.com/acme/web.git", "https://github.com/acme/gone.git"],
+      hasRepoDraft: true,
+      setSelectedRepoUrls: prunedSet,
+    });
+    await waitForText("acme/web", staleDraft.container);
+    await waitForCondition(
+      () =>
+        prunedSet.mock.calls.some(
+          ([arg]) => Array.isArray(arg) && arg.length === 1 && arg[0] === "https://github.com/acme/web.git",
+        ),
+      "prune a no-longer-granted repo from a resumed draft",
+    );
+    await unmountRoot(staleDraft.root);
   });
 
   it("falls back to a full-page redirect when the install popup is blocked", async () => {
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -1836,6 +1836,40 @@ describe("web DOM interaction coverage", () => {
     await unmountRoot(inviteeReady.root);
   });
 
+  it("prunes a no-longer-granted repo at kickoff before writing team resources", async () => {
+    // A flow can resume directly at kickoff (persisted step index) without ever
+    // mounting StepConnectCode, so connect-code's grant prune never runs. The
+    // kickoff handler must re-validate the (possibly stale) selection against the
+    // current grant list, so a repo removed from the installation since the user
+    // picked it is never registered as a team repo resource.
+    const { StepKickoff } = await import("../onboarding/steps/step-kickoff.js");
+    // Current grants are web + api (GITHUB_REPOS); the draft also carries a repo
+    // the app no longer grants.
+    githubMocks.listOrgGithubRepos.mockResolvedValue(GITHUB_REPOS);
+    const view = await renderOnboardingDom(<StepKickoff />, {
+      activeStep: "kickoff",
+      selectedRepoUrls: ["https://github.com/acme/web.git", "https://github.com/acme/gone.git"],
+      treeMode: "existing",
+      treeUrl: "https://github.com/acme/context-tree",
+    });
+    await waitForText("Your agent's ready to get to work", view.container);
+    await click(
+      ([...view.container.querySelectorAll("button")].find((b) => b.textContent?.includes("Start")) ??
+        null) as HTMLButtonElement | null,
+    );
+    await waitForText("Starting your agent", view.container);
+    // web is still granted → written; the stale repo is pruned → never written.
+    expect(resourceMocks.createTeamResourceForOrg).toHaveBeenCalledWith(
+      "org-1",
+      expect.objectContaining({ payload: { url: "https://github.com/acme/web.git" } }),
+    );
+    expect(resourceMocks.createTeamResourceForOrg).not.toHaveBeenCalledWith(
+      "org-1",
+      expect.objectContaining({ payload: { url: "https://github.com/acme/gone.git" } }),
+    );
+    await unmountRoot(view.root);
+  });
+
   it("edits MCP server rows through validation, stdio, and HTTP submissions", async () => {
     const { McpSection } = await import("../agent-detail/mcp-section.js");
     const onAdd = vi.fn();
--- a/packages/web/src/pages/onboarding-preview.tsx
+++ b/packages/web/src/pages/onboarding-preview.tsx
@@ -899,6 +899,7 @@ function baseFlow(path: OnboardingPath): OnboardingFlowValue {
     hasAgent: false,
     selectedRepoUrls: [],
     setSelectedRepoUrls: NOOP,
+    hasRepoDraft: false,
     treeMode: "new",
     setTreeMode: NOOP,
     treeUrl: "",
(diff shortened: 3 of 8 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1130

**建议**: request_changes
- 理由: restored repo drafts are pruned only when `connect-code` mounts, but the flow can resume directly at `kickoff`, where the stale URLs are written as team repo resources.

**分级**: C
- Path 初判: only `packages/web/**` files changed -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: onboarding connect-code repo selection was lost before kickoff because it only lived in React state.
- 设计方案: store a per-org `sessionStorage` draft, hydrate it in the onboarding provider, keep it through finish-later, clear it after successful completion, and avoid auto-selecting all repos when a draft exists.
- 受影响模块: `packages/web/src/pages/onboarding/**`, `packages/web/src/utils/onboarding-flags.ts`, onboarding DOM/unit tests.

❌ 1. The stale-grant prune does not run when a restored flow resumes directly at `kickoff`. [R4/R5 / packages/web/src/pages/onboarding/onboarding-flow.tsx:155]

The flow persists the active step in `sessionStorage` and restores the max of server-inferred progress and that persisted index. For an admin who selected repos, clicked through to kickoff, then refreshed / left and resumed, this means the page can reopen on `kickoff` without mounting `StepConnectCode`. In that path the draft is hydrated into `selectedRepoUrls`, but the new prune guard in `StepConnectCode` never executes. `AdminKickoff` then passes those URLs straight into `runKickoff` as `orgWrites.sourceRepos`, and the resource API only schema-validates/canonicalizes repo URLs; it does not check the current GitHub App grant list. So a repo removed from the installation between bailout and resume can still be registered as a recommended team repo and included in the bootstrap.

This is the same invariant the PR is trying to close for restored drafts, but it currently only holds for resumes that revisit connect-code. The fix should put the current-grant validation on the kickoff path as well (or make kickoff force/perform the same authoritative prune before starting), with a test that restores a draft at persisted kickoff and verifies the stale repo cannot be written.

✅ 2. The storage helper correctly preserves the important `null` vs `[]` distinction, scopes drafts per org, and uses the existing `onboarding:*` logout sweep.

**已执行**
- 已提交 request_changes

Gandy2025:
Addressed in 6d11a2d8.

The prune previously lived only in `StepConnectCode`, which the resume-at-kickoff path (persisted step index) skips. Moved the authoritative grant validation onto the kickoff write path: `handleStart` now re-validates `selectedRepoUrls` against the current GitHub App grant list (`listOrgGithubRepos`, reusing connect-code's cached query) before building `orgWrites.sourceRepos`. Ungranted repos are pruned; if the selection is empty after pruning, it falls to the no-repo intro path instead of provisioning from inaccessible repos. A transient grant-fetch failure does not block kickoff (downstream `ensureSourceReposRegistered` still verifies registration).

Added a DOM test that starts kickoff with a stale repo in the selection and asserts it is never registered while the still-granted repo is. The connect-code prune is kept for picker/UI hygiene.

Re-requesting review.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1130

**建议**: request_changes
- 理由: the kickoff re-validation now handles the happy path, but on grant-list fetch failure it falls back to the unvalidated draft and can still write stale repo resources.

**分级**: C
- Path 初判: only `packages/web/**` files changed -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: onboarding connect-code repo selection was lost before kickoff because it only lived in React state.
- 设计方案: store a per-org `sessionStorage` draft, hydrate/write it through the onboarding provider, avoid auto-selecting all repos when a draft exists, and now re-check the selected repos at kickoff before writing team resources.
- 受影响模块: `packages/web/src/pages/onboarding/**`, `packages/web/src/utils/onboarding-flags.ts`, onboarding DOM/unit tests.

❌ 1. Kickoff still writes the stale draft when current-grant validation fails. [R4/R5 / packages/web/src/pages/onboarding/steps/step-kickoff.tsx:216]

The new kickoff guard correctly prunes stale repos when `listOrgGithubRepos` succeeds, but its `catch` keeps `repos = selectedRepoUrls` and proceeds. That means the same stale draft can still be registered whenever the authoritative grant read fails: `no_installation`, `suspended`, `not_configured`, or a 502 from GitHub/mint all become “use the old draft anyway”. There is no downstream grant verification to catch this: `createTeamResourceForOrg` only schema-validates/canonicalizes the URL, and for existing-tree mode the resource writes are explicitly best-effort. So a repo removed from the installation can still ride into kickoff under exactly the failure cases where we cannot prove it is still granted.

This should fail closed for repo-backed kickoff: either block with the existing load-failed/retry style, or degrade to the no-repo intro path, but it should not write `selectedRepoUrls` after the current grant list could not be read. The test added for the successful prune path is good; please add the failure-path case too so this invariant stays explicit.

✅ 2. The new successful-fetch path covers the previous direct-kickoff resume gap: it reuses the connect-code query key, prunes the bootstrap/resource input, and handles the “all selected repos gone” case as no-repo kickoff.

**已执行**
- 已提交 request_changes

yuezengwu (changes requested):
实现目标和核心改动我理解为：把 onboarding connect-code 的 repo 选择从纯 React 状态提升为 per-org `sessionStorage` draft，保持正式 team resource 写入仍只在 kickoff；同时在 connect-code 恢复 draft 时避免把用户的 narrowing 重新覆盖成 all granted，并尝试按当前 GitHub App grant list 修剪 stale repo。

必须修复：stale-grant prune 只放在 `StepConnectCode` 的 mount path 上，不能覆盖直接恢复到 `kickoff` 的路径。`OnboardingFlowProvider` 会用 `sessionStorage` 的 `onboarding:stepIndex:admin` 恢复 active step，并在 `packages/web/src/pages/onboarding/onboarding-flow.tsx:155` 用 `Math.max(inferred, persisted)` 保留用户已到达的 kickoff。用户在 connect-code 选 repo、进入 kickoff 后刷新或离开再回来时，页面可以直接挂 `StepKickoff`，不会再挂 `StepConnectCode`，所以 `packages/web/src/pages/onboarding/steps/step-connect-code.tsx:100` 里的 prune effect 不会执行。

随后 `AdminKickoff` 会在 `packages/web/src/pages/onboarding/steps/step-kickoff.tsx:211` 把 `selectedRepoUrls` 原样传入 `runKickoff`，并在 `:216` 写成 `orgWrites.sourceRepos`。服务器端 resource create 路径只 schema/canonicalize URL 并插入资源，见 `packages/server/src/services/resources.ts:785` 起，没有重新确认该 repo 仍在当前 GitHub App installation grant 里。因此，一个 bailout/resume 期间从 GitHub App grant 中移除的 repo 仍可能被注册成 recommended team repo，进入 kickoff bootstrap。

建议把当前 grant 校验放到 kickoff 开始前或 `runKickoff`/server-side 写入前，而不是只依赖 connect-code 组件 mount；同时补一个直接从 persisted kickoff + stale draft 启动的测试，断言 stale URL 不会传给 `createTeamResourceForOrg` / `ensureSourceReposRegistered`。

没有发现 DB/schema 或核心数据结构变更；这轮只做代码路径审查，`git diff --check origin/main..HEAD` 通过，未跑 Vitest，因为 review worktree 没有安装 `node_modules`。

yuezengwu (changes requested):
第二轮实现目标和核心改动：6d11a2d8 把 repo grant 复检从只在 `StepConnectCode` mount 时发生，补到了 `StepKickoff` 的写入路径。成功读取当前 GitHub App grant list 时，`selectedRepoUrls` 会被过滤成仍 granted 的 `repos`，bootstrap 和 `orgWrites.sourceRepos` 都改用过滤后的列表；新增 DOM 测试覆盖了 stale repo 被成功 prune 的 happy path。

必须修复：grant-list fetch 失败时仍然会把未校验的 draft 写成 team repo resource。`packages/web/src/pages/onboarding/steps/step-kickoff.tsx:215` 先把 `repos` 初始化为 `selectedRepoUrls`，`:218-223` 只有在 `listOrgGithubRepos` 成功时才过滤，`:224-227` 的 `catch` 吞掉错误后继续向下走。结果是 `no_installation`、`suspended`、`not_configured`、GitHub 502、installation token mint 失败等“无法证明当前 grant”的状态，都会回退成“信任旧 draft”。随后 `:247-253` 仍把这些 URL 放进 `orgWrites.sourceRepos`。

下游没有这个 grant 校验兜底：`createTeamResourceForOrg` / server resources service 只做 payload schema、canonical key 和 insert；`ensureSourceReposRegistered` 验证的是 resource 是否注册成功，不是 repo 是否仍在 GitHub App installation grant 中。existing-tree 分支还用 `Promise.allSettled` 做 best-effort 写入，所以 stale repo 在 grant refresh 失败时仍能进入 recommended team resources。

这个路径应该 fail closed：要么在 repo-backed kickoff 上显示 retry/error，要么把无法验证 grant 的情况降级成 no-repo intro path，但不能继续写 `selectedRepoUrls`。建议补一个测试：`listOrgGithubRepos` reject、selection 含 stale URL，断言不会调用 `createTeamResourceForOrg` / `ensureSourceReposRegistered` 写入原始 draft。

没有发现 DB/schema 或核心数据结构变更。验证：`git diff --check origin/main..HEAD` 通过；未跑 Vitest，因为本地 review worktree 没有 `node_modules`。

Gandy2025:
Addressed in 6c548f72 — fail closed.

The grant re-validation no longer swallows a failed `listOrgGithubRepos` read. If the current grant list can't be read (no_installation / suspended / not_configured / upstream 5xx), kickoff now surfaces a retryable error and stays on the form instead of writing the unvalidated selection. Clicking Start again retries the grant read. So a repo can only be registered when we can positively confirm it's still granted.

Chose block-with-retry over silently degrading to the no-repo path so a transient GitHub blip doesn't quietly drop the user's tree-building intent — consistent with connect-code's existing 'Couldn't load your team's repos' treatment of the same failure.

Added a DOM test asserting no resource write and no navigation when the grant read fails.

baixiaohang (approved):
## Review by codex-assistant - PR #1130

**建议**: approve
- 理由: the restored-draft persistence now preserves user selection, and the stale-grant invariant is enforced on both connect-code UI restore and kickoff's authoritative write path, including grant-list fetch failure.

**分级**: C
- Path 初判: only `packages/web/**` files changed -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: onboarding connect-code repo selection was lost before kickoff because it only lived in React state.
- 设计方案: store a per-org `sessionStorage` draft, hydrate/write it through the onboarding provider, preserve `null` vs `[]`, avoid auto-selecting all repos over an existing draft, prune restored selections against current grants, and re-check grants at kickoff before writing team resources.
- 受影响模块: `packages/web/src/pages/onboarding/**`, `packages/web/src/utils/onboarding-flags.ts`, onboarding DOM/unit tests.

✅ 1. Previous blocker is closed: direct resume at `kickoff` now validates `selectedRepoUrls` against the current GitHub App grant list before constructing bootstrap text or `orgWrites.sourceRepos`.
✅ 2. Previous failure-path blocker is closed: when the grant list cannot be read, kickoff now fails closed with a retryable error and does not create team repo resources, chat, or completion state.
✅ 3. Draft storage semantics are scoped and explicit: `null` remains “no draft yet”, `[]` remains “deliberately selected none”, per-org keys isolate teams, and the existing `onboarding:*` sweep clears the draft on logout.

**已执行**
- 已提交 approve

Gandy2025:
@yuezengwu This is already addressed at branch HEAD (`6c548f72`). The review reads as if against the first commit (`b08b72bb`), where `StepKickoff` passed `selectedRepoUrls` through unvalidated — that exact resume-directly-at-kickoff gap was closed in two follow-up commits:

- **`6d11a2d8`** — moved the grant validation off the connect-code mount path onto the kickoff write path. `AdminKickoff.handleStart` now re-validates `selectedRepoUrls` against the current GitHub App grant list (`listOrgGithubRepos`) before building `orgWrites.sourceRepos`, so a flow resumed straight to kickoff (persisted `onboarding:stepIndex`) prunes stale repos even though `StepConnectCode` never mounts. Empty-after-prune degrades to the no-repo path.
- **`6c548f72`** — that validation fails closed: if the grant list can't be read, kickoff surfaces a retryable error and writes nothing (no fall-open).

Tests added (exactly the case you asked for): `prunes a no-longer-granted repo at kickoff before writing team resources` renders `StepKickoff` directly with a stale draft and asserts the stale URL never reaches `createTeamResourceForOrg`; plus a fail-closed test. Your worktree had no `node_modules` so Vitest didn't run — could you re-pull HEAD (`6c548f72`) and re-check?

Server-side grant enforcement would be additional defense-in-depth, but the onboarding client is the only writer here, so the kickoff-path check closes the invariant; happy to file a follow-up issue if you'd prefer belt-and-suspenders.

yuezengwu (approved):
第三轮 review 结论：approve。

实现目标和核心改动：这个 PR 将 onboarding connect-code 的 repo 选择保存为 per-org `sessionStorage` draft，保持正式 team resource 写入仍只发生在 kickoff。后续修复把当前 GitHub App grant 校验补到 kickoff 写入路径：成功读取 grant list 时先过滤 `selectedRepoUrls`，bootstrap 和 `orgWrites.sourceRepos` 都只使用仍 granted 的 repo；如果当前 grant list 读取失败，6c548f72 已改为 fail closed，显示可重试错误并停留在表单，不再写未校验 draft。

我复查了前两轮 blocker：
- 直接恢复到 persisted `kickoff` 不再绕过 stale repo prune；`StepKickoff` 自己会调用 `listOrgGithubRepos` 复检。
- grant-list fetch 失败不再 fallback 到原始 `selectedRepoUrls`；`fetchQuery` reject 会抛出 `Couldn't check your repositories...`，外层 catch 设置 error 并回到 `form`，不会进入 `runKickoff`。
- 新增 DOM 测试覆盖了 successful prune 和 failure fail-closed 两条路径，断言 stale repo 不写入、grant read 失败时不写 resource、不创建 chat、不 complete。

没有发现 DB/schema 或核心数据结构变更；这是 implementation-only，Context Tree 不需要更新。验证：`git diff --check origin/main..HEAD` 通过；未跑 Vitest，因为本地 review worktree 没有 `node_modules`。

Gandy2025:
One self-review refinement on top of the grant-validation, independent of the review threads (35d4a361):

The kickoff grant re-validation was fetching with the default `staleTime: 0`, so it did a live GitHub read on **every** "Start" click — including the normal connect-code → kickoff path where connect-code validated grants seconds earlier. With fail-closed, a transient GitHub blip at that instant would hard-block a user who did nothing wrong. Set `staleTime` so kickoff reuses the grant list connect-code just fetched within the session; a genuine resume-directly-at-kickoff is a fresh page load (empty in-memory cache) so it still fetches and validates. Net effect: the live grant read (and its failure risk) is scoped to exactly the resume-at-kickoff case that needs it, and the common path is untouched.

Summary of where the stale-grant invariant is now enforced: connect-code prune (UI) + kickoff write-path revalidation (authoritative, fail-closed) + cache reuse so the common path isn't penalized. Server-side grant enforcement at `createTeamResourceForOrg` would be additional defense-in-depth; happy to file a follow-up issue if wanted, but the onboarding client is the only writer so the invariant is closed here.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1130

**建议**: request_changes
- 理由: the new `staleTime` makes kickoff trust the connect-code cache for five minutes, so the write-path check is no longer against the current GitHub App grant list.

**分级**: C
- Path 初判: only `packages/web/**` files changed -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: onboarding connect-code repo selection was lost before kickoff because it only lived in React state.
- 设计方案: store a per-org `sessionStorage` draft, hydrate/write it through the onboarding provider, preserve `null` vs `[]`, prune restored selections against grants, and validate selected repos at kickoff before writing team resources.
- 受影响模块: `packages/web/src/pages/onboarding/**`, `packages/web/src/utils/onboarding-flags.ts`, onboarding DOM/unit tests.

❌ 1. `staleTime` reopens stale-grant writes inside the same tab/session. [R5 / packages/web/src/pages/onboarding/steps/step-kickoff.tsx:224]

The kickoff check is supposed to be the authoritative write-path validation against the current GitHub App grant list. Adding `staleTime: 5 * 60 * 1000` means `fetchQuery` can return the `connect-code` query cache without hitting GitHub. That makes the guard trust a list fetched before the user reached kickoff. Grants can still change within the same browser session: the user can click “finish later” (SPA navigation, QueryClient cache remains alive), change the GitHub App repository selection from GitHub/Settings, and resume onboarding within five minutes; or the GitHub App grant can be changed in another tab while this tab sits on kickoff. In those paths, a repo removed from the installation is still present in the fresh-by-staleTime cache, so it can be included in `orgWrites.sourceRepos` and written as a recommended team resource.

This is exactly why the kickoff check needs to be authoritative rather than another consumer of the picker cache. If avoiding a redundant read on the normal connect-code -> kickoff path is important, the optimization needs a narrower proof that the cached list was fetched after the last possible grant-changing navigation. Otherwise keep `fetchQuery` staleTime at its default (or explicitly force a refetch) for the kickoff write path.

✅ 2. The previous fail-closed behavior is still correct when the grant read actually runs and rejects.

**已执行**
- 已提交 request_changes

Gandy2025:
@yuezengwu You're reviewing intermediate commit `6d11a2d8` — the fall-open you describe is correct for that commit but was fixed two commits earlier than HEAD:

- **`6c548f72`** ("fail closed at kickoff when the grant list can't be read") — the `catch` no longer swallows-and-proceeds; it now `throw`s, so a failed `listOrgGithubRepos` read surfaces a retryable error and writes **nothing** (no `createTeamResourceForOrg` / `ensureSourceReposRegistered`). Clicking Start retries.
- Test added in the same commit, exactly the case you suggest: `fails closed at kickoff when the grant list can't be read (no stale write)` (web-dom-interactions.test.tsx:1873) — `listOrgGithubRepos` rejects, selection has a repo, asserts no resource write and no navigation.
- HEAD is `35d4a361` (adds a `staleTime` so the common connect-code → kickoff path reuses the just-fetched grant list instead of a redundant live read).

Could you review the PR at HEAD (`35d4a361`) as a whole rather than per-commit? Both of your points (resume-at-kickoff prune and the fetch-failure fail-closed) are already closed there. `git fetch && git checkout 35d4a361` to see the final state.

yuezengwu (changes requested):
第四轮 review 结论：request changes。

实现目标和核心改动：最新提交 `35d4a361` 试图避免 normal connect-code -> kickoff 路径在点击 Start 时再次实时读取 GitHub grant list；它给 kickoff 的 `queryClient.fetchQuery(["onboarding", "org-github-repos", organizationId])` 增加了 `staleTime: 5 * 60 * 1000`，让 kickoff 复用 connect-code 刚加载过的 repo grant 缓存。

必须修复：这个 `staleTime` 重新削弱了 kickoff 写入路径的“当前 grant”校验。`packages/web/src/pages/onboarding/steps/step-kickoff.tsx:223-235` 现在会在缓存 5 分钟内直接返回 connect-code 之前读到的 grant list，而不是重新确认 GitHub App 当前仍 grant 这些 repo。QueryClient 是 app 级单例，`finishLater` 只是 `navigate("/")`，并不会清空 React Query cache；所以用户从 connect-code 进入 kickoff 后点 “I'll finish later”、在同一 tab/session 里稍后从 Settings -> Setup 回到 onboarding，仍可能直接恢复到 persisted `kickoff`，并复用几分钟内的旧 grant cache。如果这期间 GitHub App installation scope 被改小、installation suspended，或 repo 从 grant 中移除，kickoff 仍会把旧 cached grant 当作通过校验，然后把 draft repo 写进 `orgWrites.sourceRepos`。

这正是前两轮 review 要关闭的边界：repo resource 写入前必须能证明 repo 仍在当前 GitHub App grant 中。5 分钟缓存把证明从“current”降级成“recently observed”。如果要减少普通路径上的 transient GitHub blip，可以另做 UX/重试优化，或把 grant 校验下沉到资源写入/服务器侧；但 kickoff 的 authoritative write path 不能接受可能 stale 的 grant cache。建议去掉这里的 `staleTime`，或至少在任何可能写 `sourceRepos` 的路径上强制实时 refetch/validate，并补一个测试：先 seed query cache with stale granted repo，随后 mock `listOrgGithubRepos` 返回移除后的 grant list，直接渲染/恢复 `StepKickoff` 并点击 Start，断言 stale repo 不会写入。

没有发现 DB/schema 或核心数据结构变更。验证：`git diff --check origin/main..HEAD` 通过；未跑 Vitest，因为本地 review worktree 没有 `node_modules`。

yuezengwu (commented):
Re-request checked against HEAD `35d4a361`; there are no code changes since review `4512551957`, only PR description edits.

The previous request-changes still applies. The unresolved issue is not the fail-closed path when a live grant read rejects; that part remains fixed. The issue is the new `staleTime` on the kickoff `fetchQuery`: within the same SPA session, kickoff can reuse connect-code's cached grant list for five minutes, so the write-path validation is not necessarily against the current GitHub App grant list. `finishLater`/navigation does not clear the app-level QueryClient cache, and GitHub App grants can change in another tab or outside this tab before the user resumes.

So the PR description's claim that the invariant is closed on every path is not supported by the current code. The fix still needs to remove/narrow the cache reuse for the authoritative `sourceRepos` write path, or move the grant check to a boundary that can prove current grant state before writing.

Gandy2025:
Fixed in f15e1996 — you're both right, and my earlier reasoning was wrong.

I assumed a resume-at-kickoff was a fresh page load with an empty cache, but `finishLater` is SPA `navigate("/")` and the QueryClient is an app-level singleton, so the connect-code grant cache survives the bailout. With `staleTime: 5min` the authoritative write-path check could reuse a list minutes-stale relative to grants changed in another tab / GitHub settings, and write a removed repo — exactly the invariant this check exists to enforce.

Reverted: `staleTime` is pinned to `0` (with a comment explaining why it must stay 0), so the kickoff check always reads the current grant list. The redundant read on the normal connect-code → kickoff path is the accepted cost of correctness; if the transient-blip UX ever matters it's a separate optimization, not a weakening of the authoritative check.

Added the regression test you asked for (@yuezengwu): seed the query cache with a stale grant list still containing the removed repo, mock `listOrgGithubRepos` to return the current (reduced) list, render `StepKickoff`, click Start, assert the stale repo is never registered. It fails with the staleTime present and passes without it.

HEAD is now f15e1996; local typecheck/Biome/onboarding suites green (114 tests).

baixiaohang (approved):
## Review by codex-assistant - PR #1130

**建议**: approve
- 理由: the kickoff grant check is authoritative again: it no longer trusts the connect-code cache, fails closed on grant-read failure, and has coverage for stale-cache regression.

**分级**: C
- Path 初判: only `packages/web/**` files changed -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: onboarding connect-code repo selection was lost before kickoff because it only lived in React state.
- 设计方案: persist a per-org `sessionStorage` draft, preserve `null` vs `[]`, restore/write through the selection in the onboarding provider, prune restored selections against current grants for picker hygiene, and re-validate current grants at kickoff before writing team resources.
- 受影响模块: `packages/web/src/pages/onboarding/**`, `packages/web/src/utils/onboarding-flags.ts`, onboarding DOM/unit tests.

✅ 1. Previous `staleTime` blocker is closed: kickoff now uses `fetchQuery(..., staleTime: 0)` for the write-path grant check, so a cached connect-code grant list cannot satisfy the authoritative validation.
✅ 2. Regression coverage was added for the exact stale-cache case: the test seeds a cached grant containing a removed repo, mocks the live grant list without it, and verifies only the live-granted repo is written.
✅ 3. Earlier restored-draft and fetch-failure blockers remain closed: direct kickoff resume prunes against grants, and grant-read failure keeps the user on the form without resource writes, chat creation, or completion.

**已执行**
- 已提交 approve

yuezengwu (approved):
第五轮 review 结论：approve。

实现目标和核心改动：这个 PR 继续保持 onboarding connect-code 的 repo 选择作为 per-org `sessionStorage` draft，正式 team resource 写入仍只在 kickoff。最新提交 `f15e1996` 针对上一轮 blocker 做了正确修复：kickoff 的 grant-list `fetchQuery` 显式 `staleTime: 0`，把该路径重新固定为 authoritative write-path check；即使同一 SPA session 中仍有 connect-code 的旧 grant cache，Start 时也会读取当前 GitHub App grant list，再用当前 grant 过滤 `selectedRepoUrls` 后构造 bootstrap 和 `orgWrites.sourceRepos`。

我复查的关键点：
- `StepKickoff` 不再复用 5 分钟缓存；`staleTime: 0` 使 sourceRepos 写入前校验回到当前 grant list。
- grant-list fetch 失败仍 fail closed，设置可重试错误并停在表单，不写 resource、不创建 chat、不 complete。
- 新增 DOM 回归测试会先 seed stale query cache（包含已移除 repo），再 mock live grant list，断言 removed repo 不会写入；这覆盖了上一轮指出的同 tab/session cache 问题。

没有发现 DB/schema 或核心数据结构变更；这是 implementation-only，Context Tree 不需要更新。

验证：`git diff --check origin/main..HEAD` 通过；未跑 Vitest，因为本地 review worktree 没有 `node_modules`。

非阻塞提醒：PR description 仍有 “A `staleTime` makes the normal connect-code → kickoff path reuse...” 和 “cache reuse...” 的旧说法，已经和 `f15e1996` 的代码不一致；建议合并前更新描述，避免误导后续读者。

yuezengwu (approved):
Re-request checked: HEAD is still `f15e19967b22f80f942ad91c2dd6d2b39a6d3f04`, with no code changes since my approve review `4512585078`.

Approval still stands. The staleTime blocker is closed in code: kickoff uses `staleTime: 0` for the authoritative grant check and has stale-cache regression coverage. The only remaining note is non-blocking: the PR description still mentions staleTime/cache reuse and should be updated to match the current implementation before merge.

PR #1134, opened 2026-06-17
docs(skills): simplify context tree workflows

bestony:
## Summary
- remove obsolete workspace binding resolution steps from the first-tree read skill
- update the write skill CLI section to rely on current tree help and verification commands

## Testing
- Not run: `pnpm check && pnpm typecheck` failed because local `node_modules` is missing and `biome` is not installed.

### DIFF
--- a/skills/first-tree-read/SKILL.md
+++ b/skills/first-tree-read/SKILL.md
@@ -19,44 +19,7 @@ is available and ask for a specific source artifact or scope.
 
 ## Workflow
 
-### 1. Resolve the context repo
-
-Find the workspace binding from the current working directory:
-
-```bash
-find_workspace_root() {
-  local d=$(pwd)
-  while [ "$d" != "/" ]; do
-    if [ -f "$d/.first-tree/workspace.json" ]; then echo "$d"; return; fi
-    d=$(dirname "$d")
-  done
-  return 1
-}
-
-WS=$(find_workspace_root) || { echo "No First Tree workspace at or above cwd"; exit 1; }
-cat "$WS/.first-tree/workspace.json"
-```
-
-Resolve the context repo as `<workspaceRoot>/<manifest.tree>`. If the
-manifest is missing or malformed, stop and report the binding gap — do
-not guess a context repo.
-
-If the manifest is present but the resolved path **does not exist on
-disk**, the workspace is agent-managed and this is the agent's job to
-materialise: follow the **Tree Location** block in your `AGENTS.md` /
-`CLAUDE.md` briefing to clone the upstream tree repo into the resolved
-path (the briefing carries the upstream URL, branch, and a ready
-`git clone` command). Once the directory exists, continue below. (If the
-path exists as a **symlink**, treat it as the legacy shared-pool layout —
-remove only the symlink, then clone per the briefing.)
-
-You do **not** need a separate `git pull` step before reading: the
-`first-tree tree tree` command in step 2 runs `git pull --ff-only` on the
-context repo for you (a built-in freshness guarantee), degrading to the
-local copy with a warning if the remote is unreachable. Pass `--no-pull`
-only when you deliberately want a stable snapshot or are working offline.
-
-### 2. Inspect the reader command every time
+### 1. Inspect the reader command every time
 
 Run the help command from inside the context repo before using any
 `tree tree` selector:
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1134

- Reason: The updated `first-tree-read` workflow now starts inside `$CONTEXT_REPO` without telling the agent how to resolve that repo, which removes a required part of the shipped skill contract.

**Level**: C
- Path initial: only `skills/**` documentation changes -> C
- Semantic lift: none

**PR Summary**
- Problem addressed: simplify the shipped Context Tree read/write skill instructions after the current CLI and workspace model changed.
- Design approach: remove the explicit workspace binding resolution step from `first-tree-read`, and update `first-tree-write`'s CLI section to rely on current help/verification commands.
- Affected modules: `skills/first-tree-read`, `skills/first-tree-write`

**Review Comments**
❌ 1. The `first-tree-read` workflow no longer has an executable path to the context repo. After this change, step 1 says to `cd "$CONTEXT_REPO"`, but the skill never defines that variable or tells the agent to derive it from `<workspaceRoot>/.first-tree/workspace.json`. The Context Tree contract still says the reader must resolve the workspace binding first, and the architecture node describes `first-tree-read` as the workflow that resolves that binding before running `first-tree tree tree`. As written, a fresh agent that follows only the loaded skill will stall or guess the repo path. Please keep a concise binding-resolution step here, or update the durable Context Tree/architecture contract in the same source-backed change if the binding responsibility has intentionally moved elsewhere. [R5 / `skills/first-tree-read/SKILL.md:22`]

**Executed**

skills/first-tree-read/SKILL.md
```diff
@@ -19,44 +19,7 @@ is available and ask for a specific source artifact or scope.
...
-remove only the symlink, then clone per the briefing.)
-
-You do **not** need a separate `git pull` step before reading: the
-`first-tree tree tree` command in step 2 runs `git pull --ff-only` on the
-context repo for you (a built-in freshness guarantee), degrading to the
-local copy with a warning if the remote is unreachable. Pass `--no-pull`
-only when you deliberately want a stable snapshot or are working offline.
-
-### 2. Inspect the reader command every time
+### 1. Inspect the reader command every time
```
yuezengwu, line 22: This section now starts from `cd "$CONTEXT_REPO"`, but this PR deleted the only step that found the workspace root, read `.first-tree/workspace.json`, and derived/materialized the context repo. The current reader contract still says binding first, and the eval fixture still expects Context Tree reads to use the current workspace binding before selecting files. Please restore a resolution step or explicitly define `CONTEXT_REPO` from the generated Tree Location before the snippet.

skills/first-tree-write/SKILL.md
```diff
@@ -460,20 +460,20 @@ stay in the node body.
...
-standard tooling (`ls`, `Read`, `Grep`) and rely on `tree verify` as
-the write gate. Everything else (opening PRs, fetching code, reading
-the workspace binding) goes through standard tools (`git`, `gh`,
-`Read`, etc.).
+Start from the top-level CLI help, then use the tree-specific command from
+inside the context repo to inspect and locate the current Context Tree
+structure:
+
+```bash
+cd "$CONTEXT_REPO"
```
yuezengwu, line 468: Same executability gap here: the snippet uses `$CONTEXT_REPO`, but this skill never defines it. Since this is copy-paste operational guidance, agents in staging/dev or detached contexts will fail unless they infer hidden state from the generated briefing. Please define the variable, or phrase this as `cd <Tree Location path from AGENTS.md>` and keep the channel-correct binary substitution explicit.

yuezengwu (changes requested):
Reviewed the two skill-doc changes. Requesting changes because the edited workflows now reference `$CONTEXT_REPO` after removing or omitting the resolution path. That makes the instructions non-executable unless the agent infers state from a separate briefing, and it conflicts with the current Context Tree contract that `first-tree-read` resolves the workspace binding before selector commands.

No database or core data-structure changes in this PR.

yuezengwu:
Closing this stale PR because the existing requested-change blocker is still valid and the shipped skills have moved on.

The PR tries to simplify `first-tree-read` by removing the workspace-binding resolution path, but the resulting workflow references the Context Tree repo without an executable way to resolve it. Current `main` intentionally keeps the manifest-based resolution step, including agent-managed materialization when the bound tree clone is missing, before `first-tree tree tree` selection.

Merging this branch now would regress the read skill contract. There are no database or core data-structure changes here; a future simplification should start from current `main` and preserve a self-contained path from workspace binding to the selected tree files.

PR #1152, opened 2026-06-18
fix(client): wait for eviction cleanup before resume

### LINKED ISSUE #1151: Client runtime: eviction resume can overlap old handler shutdown
## Context

Post-close validation of stale #973 / closed PR #982 found that the original silent-drop scheduler bug is mostly covered in current `main`, but one narrower review finding from #982 does not appear to have an equivalent landing.

Current `main` already has the core #973 protections:

- concurrency preemption prefers idle sessions before working sessions;
- preempting a working session uses `ackConsumedPrefix: false`;
- preempted / owed work can request chat recovery;
- defaults are effectively unbounded (`DEFAULT_AGENT_CONCURRENCY = 99`, `DEFAULT_AGENT_MAX_SESSIONS = 99`).

## Remaining gap

`packages/client/src/runtime/session-manager.ts` still stores evicted sessions as:

```ts
private readonly evictedMappings = new Map<string, { claudeSessionId: string; lastActivity: number }>();
```

When `evictIfNeeded` evicts an active candidate, it calls:

```ts
candidate.session.handler.shutdown().catch(() => {});
this.sessions.delete(candidate.key);
```

Then `startNewSession` can immediately see the `evictedMappings` entry and call `handler.resume(...)` for the same chat without waiting for the prior handler's shutdown/suspend promise to settle.

That is the same class of blocker called out in the #982 review: a proactive recovery redelivery or later input can start a fresh handler while the old one is still aborting, violating the intended `(agent, chat) = 1 live handler` boundary.

## Suggested fix

Carry a runtime-only cleanup promise in the evicted mapping and await it before resuming from eviction:

- for active eviction: capture `handler.shutdown()`;
- for suspended / just-preempted eviction: capture `entry.suspending` when present;
- in `startNewSession`, if `evicted.cleanup` exists, await it before `handler.resume(...)`;
- add a regression test that holds the old shutdown gate and proves redelivery does not resume until it settles.

## Relationship to old stale items

- Keep #973 closed: the original silent ACK/drop behavior is no longer the right tracking shape.
- Keep #982 closed: the old branch is stale; this is a narrower follow-up against current `main`.

yuezengwu:
## Summary

- Store a runtime-only cleanup promise on evicted session mappings.
- Capture active handler `shutdown()` and in-flight `suspending` cleanup before dropping a session from `sessions`.
- Await that cleanup before calling `handler.resume(...)` for the evicted chat, while keeping the new session entry as the in-flight routing placeholder.
- Add a regression test proving redelivery does not resume until the old shutdown gate settles.

## Root Cause

`evictIfNeeded` preserved only the prior session id when LRU/max-session eviction dropped a session. For active sessions it started `handler.shutdown()` fire-and-forget, deleted the local entry, and allowed a later same-chat redelivery to immediately create a fresh handler and call `resume(...)` against the same saved session id. That could overlap the old handler's shutdown/abort path and violate the `(agent, chat) = 1 live handler` boundary.

## Validation

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts`
- `pnpm exec biome check packages/client/src/runtime/session-manager.ts packages/client/src/__tests__/session-manager.test.ts`

Fixes #1151

### DIFF
--- a/packages/client/src/__tests__/session-manager.test.ts
+++ b/packages/client/src/__tests__/session-manager.test.ts
@@ -720,6 +720,84 @@ describe("SessionManager", () => {
     await sm.shutdown();
   });
 
+  it("waits for evicted active handler shutdown before resuming the same chat", async () => {
+    const shutdownGate = deferred<void>();
+    const contexts = new Map<string, SessionContext>();
+    const messages = new Map<string, SessionMessage>();
+    const resumeSpy = vi.fn(async (msg: SessionMessage | undefined, sessionId: string, ctx: SessionContext) => {
+      if (msg) {
+        contexts.set(`${msg.chatId}:resume`, ctx);
+        messages.set(`${msg.chatId}:resume`, msg);
+      }
+      return sessionId;
+    });
+
+    const handlerA = createMockHandler({
+      async start(msg, ctx) {
+        contexts.set(msg.chatId, ctx);
+        messages.set(msg.chatId, msg);
+        return "session-chat-a";
+      },
+      shutdown: vi.fn(() => shutdownGate.promise),
+    });
+    const handlerB = createMockHandler({
+      async start(msg, ctx) {
+        contexts.set(msg.chatId, ctx);
+        messages.set(msg.chatId, msg);
+        return "session-chat-b";
+      },
+    });
+    const handlerAResume = createMockHandler({
+      resume: resumeSpy,
+    });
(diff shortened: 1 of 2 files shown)


yuezengwu:
Superseded by #1153. The original PR used `yzw/issue-1151-eviction-cleanup`, which fails the repository branch-name check; #1153 carries the same commit on `fix/issue-1151-eviction-cleanup`.

PR #1153, opened 2026-06-18
fix(client): wait for eviction cleanup before resume

### LINKED ISSUE #1151: Client runtime: eviction resume can overlap old handler shutdown

yuezengwu:
## Summary

- Store a runtime-only cleanup promise on evicted session mappings.
- Capture active handler `shutdown()` and in-flight `suspending` cleanup before dropping a session from `sessions`.
- Await that cleanup before calling `handler.resume(...)` for the evicted chat, while keeping the new session entry as the in-flight routing placeholder.
- Add a regression test proving redelivery does not resume until the old shutdown gate settles.

## Root Cause

`evictIfNeeded` preserved only the prior session id when LRU/max-session eviction dropped a session. For active sessions it started `handler.shutdown()` fire-and-forget, deleted the local entry, and allowed a later same-chat redelivery to immediately create a fresh handler and call `resume(...)` against the same saved session id. That could overlap the old handler's shutdown/abort path and violate the `(agent, chat) = 1 live handler` boundary.

## Validation

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts`
- `pnpm exec biome check packages/client/src/runtime/session-manager.ts packages/client/src/__tests__/session-manager.test.ts`

Fixes #1151

### DIFF
--- a/packages/client/src/__tests__/session-manager.test.ts
+++ b/packages/client/src/__tests__/session-manager.test.ts
@@ -720,6 +720,84 @@ describe("SessionManager", () => {
     await sm.shutdown();
   });
 
+  it("waits for evicted active handler shutdown before resuming the same chat", async () => {
+    const shutdownGate = deferred<void>();
+    const contexts = new Map<string, SessionContext>();
+    const messages = new Map<string, SessionMessage>();
+    const resumeSpy = vi.fn(async (msg: SessionMessage | undefined, sessionId: string, ctx: SessionContext) => {
+      if (msg) {
+        contexts.set(`${msg.chatId}:resume`, ctx);
+        messages.set(`${msg.chatId}:resume`, msg);
+      }
+      return sessionId;
+    });
+
+    const handlerA = createMockHandler({
+      async start(msg, ctx) {
+        contexts.set(msg.chatId, ctx);
+        messages.set(msg.chatId, msg);
+        return "session-chat-a";
+      },
+      shutdown: vi.fn(() => shutdownGate.promise),
+    });
+    const handlerB = createMockHandler({
+      async start(msg, ctx) {
+        contexts.set(msg.chatId, ctx);
+        messages.set(msg.chatId, msg);
+        return "session-chat-b";
+      },
+    });
+    const handlerAResume = createMockHandler({
+      resume: resumeSpy,
+    });
+    const handlers = [handlerA, handlerB, handlerAResume];
+    const sm = createSessionManager({
+      handlerFactory: () => {
+        const handler = handlers.shift();
+        if (!handler) throw new Error("unexpected handler allocation");
+        return handler;
+      },
+      recoverChat: vi.fn().mockResolvedValue(undefined),
+      concurrency: 1,
+      session: { idle_timeout: 300, max_sessions: 1, working_grace_seconds: 3600, reconcile_interval_seconds: 300 },
+    });
+
+    const chatA = mockEntry({ id: 1, chatId: "chat-a", messageId: "msg-a" });
+    await sm.dispatch(chatA);
+    await finishEntry(contexts.get("chat-a"), 1, "chat-a", "msg-a");
+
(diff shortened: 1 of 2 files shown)


yuezengwu:
Stale-queue review against current `main`:

**Goal / core change:** serialize eviction cleanup with same-chat resume so a new handler cannot resume the saved provider session while the prior handler is still shutting down or suspending.

The defect is still visible on current `main`: `evictedMappings` carries only `{ claudeSessionId, lastActivity }`, active LRU eviction still calls `handler.shutdown().catch(...)` fire-and-forget, and a later start consumes the mapping without awaiting prior cleanup. There is no equivalent landing elsewhere.

**Current gate:** this draft conflicts with `main`, and `session-manager.ts` has accumulated substantial delivery/recovery behavior since this head. The old two-file patch should not be conflict-resolved mechanically; it needs a fresh current-main runtime review of cleanup ownership, retry/ACK behavior, terminate handling, and the regression gate. No database or schema change is involved, but this is core session-lifecycle behavior. Leaving the PR open and draft pending an explicit refresh/rebuild decision.

yuezengwu:
@yuezengwu This still tracks a real session-lifecycle race, but the branch now conflicts with current `main` in core runtime code.

**Recommendation:** close this stale PR, reopen #1151 (or replace it with a current tracking issue), and rebuild the narrow fix from current `main` with eviction-cleanup/resume ordering revalidated against the current session lifecycle. Please reply here to confirm whether you will close and rebuild, or explain why this branch should remain open.

yuezengwu:
Final stale-queue disposition: the follow-up window elapsed with no new reply or head update. Per maintainer direction, I am closing this stale implementation PR without merging it or deleting its branch.

This closure does **not** mark the session-lifecycle race as fixed. #1151 remains closed as not planned. If production or debugging evidence raises the priority again, reopen that issue or create a fresh current-main tracking issue, then revalidate eviction-cleanup/resume ordering against the current session lifecycle before implementing a replacement.

PR #1159, opened 2026-06-18
fix(client): batch settle Claude Code coalesced inputs

### LINKED ISSUE #1156: client: un-ACKed coalesced input leaves session stuck "working" until 65-min hard cap (delivery stall + duplicate reply)
## Summary

A client agent session can stay in `working` state for up to `idle_timeout + working_grace_seconds` (default **65 min**) *after it has actually gone idle*, because an inbox entry is left in `owned + processingStarted` and never `finishTurn`'d. Consequences:

- **Delivery stall**: the un-ACKed entry holds the per-`(agent, chat)` delivery window, so later messages in that chat sit `pending` server-side for the whole window.
- **Duplicate reply**: when the hard cap finally force-suspends the session, the still-non-terminal entry is redelivered and reprocessed.

Observed on the staging channel: a message `created 20:39:28 / delivered 21:45:06 / acked 21:45:25` — exactly `idle_timeout(300s) + working_grace_seconds(3600s) = 3900s` after the session's last activity at `20:40:02`.

## Root cause — the ACK model assumes 1 input : 1 result

`packages/client/src/handlers/claude-code.ts` tracks delivered/injected messages in a FIFO queue `pendingAckMessages` and assumes **one `result/success` per input message**:

- `pushPendingAckMessage(message, token)` enqueues each input; only the head is marked `processingStarted`.
- Each SDK `result` with `subtype:"success"` → `ackTurnClose()` (~L928-938) `shift()`s and `complete()`s **exactly one** queued message, then marks the new head `processingStarted` (success branch ~L1294-1302).
- `token.processingStarted → markProcessingStarted` (owned+processingStarted); `token.complete → finishTurn` (terminal + ACK) — `session-manager.ts::createDeliveryToken` (~L862-869).

In streaming-input / bg-agent mode, multiple messages can be fed into a single SDK query and the model may answer them with **fewer** `result` messages (coalescing). When `#inputs > #result-success`, the surplus inputs stay in `pendingAckMessages`; the head sits `owned+processingStarted` with no further `result` to close it. The SDK query then parks in `for await (... of currentQuery)` awaiting the next input → `recordProviderActivity()` (~L1207, the only per-turn `lastActivity` bump) never fires → `lastActivity` frozen.

## Why it stays `working` and is never suspended

`runtimeState` is projected purely from the ACK ledger, **not** from whether the SDK is actually generating:

- `session-manager.ts::projectedRuntimeState` (~L1935-1940): `return hasProcessingOwnedWork(chatId) ? "working" : "idle";`
- `evictIdle` (~L1714-1757) also gates skip-suspend on `hasProcessingOwnedWork` (~L1725), not on `runtimeState`. With processing-owned work it logs `"session idle threshold reached but provider work is still processing — skipping suspend"` and keeps the slot until `inactiveMs >= idle_timeout + working_grace_seconds` (hard cap), then force-suspends.

So the orphaned entry keeps `hasProcessingOwnedWork === true` forever → the session is `working` and unsuspendable until the 65-min hard cap, even though the agent went idle right after its last `result`.

## Why the duplicate happens (downstream symptom, not a recovery flaw)

Hard-cap suspend goes `suspendSession → prepareSuspend` (not `prepareOperatorSuspend`). `inbox-delivery-coordinator.ts::prepareSuspend` (~L286-307) ACKs only the terminal prefix and marks the still-non-terminal orphan as recovery debt → redelivery → reprocess → duplicate reply. (`prepareOperatorSuspend` ~L309-329 would terminalize+ACK owned-processing work; the idle/hard-cap path does not.) This is at-least-once working as intended; the duplicate is a consequence of the orphan, so fixing the orphan removes it.

## Evidence (staging `client.log`; pid stable, no restart; CLI upgrade at 20:11 unrelated)

- `lastActivity` frozen at `20:40:02` (last `Result forwarded to chat`).
- `... skipping suspend` logged ~360× from `20:45:05`; `inactiveSec` climbs `303 → 3893`, `graceSec=3600`.
- `idle_timeout=300`, `working_grace_seconds=3600` (`runtime/config.ts` ~L21/L28) → hard cap 3900s.
- `21:45:05 session idle, suspending` → `chat has unsettled inbox work; waiting for recovery redelivery` → resume `21:45:08` → `21:45:25` duplicate `Result forwarded to chat`.

## Impact

- Up to ~65 min delivery delay for later messages in the affected chat.
- Duplicate reply on hard-cap recovery.
- A "zombie working" slot held for the full grace window (also blocks the slot from serving other work).

## Fix direction

Align the ACK ledger with the agent's real turn boundary instead of assuming 1 input : 1 result. When the SDK query reaches a turn boundary / goes idle with `pendingAckMessages` still non-empty, reconcile — `finishTurn`/ACK the inputs the model actually consumed up to the latest result — rather than waiting for a 1:1 result that never arrives. After the fix, `hasProcessingOwnedWork` clears when the agent is genuinely idle → the session projects `idle` and suspends on the normal `idle_timeout`; `working_grace_seconds` then only protects genuinely-active single long steps.

**Out of scope (separate, deprioritized):** `forwardResult` idempotency key. Duplicate replies are rare and low-impact (they do not wake other agents); revisit only if agents start performing irreversible external side-effects mid-turn (e.g. posting a GitHub comment / sending mail), which at-least-once redelivery would otherwise double.

## Verification

- Repro: feed ≥2 notify messages into one streaming turn that the model answers with a single combined `result`; assert the session projects `idle` (not stuck `working`) and suspends within `idle_timeout`.
- DB confirm for the observed incident: in `inbox_entries` (agent `code-reviewer`, chat `98a24d6f`, window 20:25–21:45) a notify row delivered before 20:40 with `acked_at` null until 21:45 = the orphan.

> Line numbers are against `origin/main` @ `adf36a2f`; treat as approximate anchors.

baixiaohang:
## Summary
- Track Claude Code pending ACK entries as SDK input envelopes and mark them provider-entered when the SDK prompt iterable actually pulls them.
- Snapshot the provider-entered prefix on each SDK result and complete that prefix as one batch, so coalesced inputs do not leave orphaned owned+processing inbox work.
- Preserve the existing single-entry permanent-error fallback for retry-exhausted / auto-resume-fail paths when no input reached provider processing.

## Tests
- pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-inject-startup-race.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts
- pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-turn-end.test.ts src/__tests__/claude-code-turn-end-sdk-error.test.ts src/__tests__/claude-code-result-forward-fail.test.ts src/__tests__/claude-code-stream-error-retry-replay.test.ts src/__tests__/claude-code-turn-end-serialization.test.ts
- pnpm --filter @first-tree/client typecheck
- pnpm check
- pnpm typecheck
- pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1

Closes #1156

### DIFF
--- a/packages/client/src/handlers/claude-code.ts
+++ b/packages/client/src/handlers/claude-code.ts
@@ -58,6 +58,13 @@ type PendingSdkInput = {
   pendingAck: PendingAckMessage | null;
 };
 
+type QueuedInjectedMessage = {
+  message: SessionMessage;
+  token: DeliveryToken;
+  recoveryReason?: string;
+  recoveryRetried?: boolean;
+};
+
 /**
  * Bug 6: thrown by `consumeOutput` when an SDK "success" result message
  * actually contains an API error string (e.g. "API Error: socket
@@ -708,9 +715,11 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
    * when the session ends or `start()` runs for a fresh session.
    */
   let chatContextForPrompt: ChatContext | undefined;
-  const queuedInjectedMessages: Array<{ message: SessionMessage; token: DeliveryToken }> = [];
+  const queuedInjectedMessages: QueuedInjectedMessage[] = [];
   const pendingAckMessages: PendingAckMessage[] = [];
   let injectDrainInProgress = false;
+  let drainingInjectedMessage: QueuedInjectedMessage | null = null;
+  let inputRecoveryReason: string | null = null;
   /**
    * Predeclared source repos the agent config declares at
    * `<agentHome>/source-repos/<localPath>/`. Pure declaration (`declaredSourceRepos`) —
@@ -977,22 +986,40 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
     }
   }
 
+  function retryInjectedItem(item: QueuedInjectedMessage, reason: string): void {
+    item.recoveryReason = reason;
+    if (item.recoveryRetried) return;
+    item.recoveryRetried = true;
+    item.token.retry(item.message, reason);
+  }
+
+  function recoverIfInputClosed(item: QueuedInjectedMessage): boolean {
+    const reason = item.recoveryReason ?? inputRecoveryReason;
+    if (!reason) return false;
+    retryInjectedItem(item, reason);
+    return true;
+  }
+
   async function pushInjectedMessage(
-    message: SessionMessage,
+    item: QueuedInjectedMessage,
     sessionCtx: SessionContext,
     sessionId: string,
-    token: DeliveryToken,
   ): Promise<void> {
+    const { message, token } = item;
+    if (recoverIfInputClosed(item)) return;
     try {
       await maybeSwitchConfig(sessionCtx);
     } catch (err) {
       sessionCtx.log(`maybeSwitchConfig errored: ${err instanceof Error ? err.message : String(err)}`);
     }
+    if (recoverIfInputClosed(item)) return;
 
     try {
       const sdkMsg = await toSDKUserMessage(message, sessionCtx, sessionId);
+      if (recoverIfInputClosed(item)) return;
       pushPendingSdkInput(createPendingSdkInput(sdkMsg, message, token));
     } catch (err) {
+      if (recoverIfInputClosed(item)) return;
       sessionCtx.log(`toSDKUserMessage errored: ${err instanceof Error ? err.message : String(err)}`);
       // The SDK has not seen this input yet, so there is no durable terminal
       // evidence. Keep it recoverable instead of ACKing through `complete`.
@@ -1013,11 +1040,14 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
         ) {
           const queued = queuedInjectedMessages.shift();
           if (!queued) continue;
+          drainingInjectedMessage = queued;
           try {
-            await pushInjectedMessage(queued.message, sessionCtx, sessionId, queued.token);
+            await pushInjectedMessage(queued, sessionCtx, sessionId);
           } catch (err) {
             sessionCtx.log(`inject drain failed: ${err instanceof Error ? err.message : String(err)}`);
-            queued.token.retry(queued.message, "claude_inject_drain_failed");
+            retryInjectedItem(queued, "claude_inject_drain_failed");
+          } finally {
+            if (drainingInjectedMessage === queued) drainingInjectedMessage = null;
           }
         }
       } finally {
@@ -1030,12 +1060,21 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
   }
 
   function retryBufferedMessages(reason: string): void {
+    inputRecoveryReason = reason;
     unclosedSdkInputs.length = 0;
+    const pending = pendingAckMessages.splice(0);
+    const drainingIsPending =
+      drainingInjectedMessage !== null &&
+      pending.some(
+        (pendingItem) =>
+          pendingItem.message === drainingInjectedMessage?.message &&
+          pendingItem.token === drainingInjectedMessage.token,
+      );
+    if (drainingInjectedMessage && !drainingIsPending) retryInjectedItem(drainingInjectedMessage, reason);
     const queued = queuedInjectedMessages.splice(0);
     for (const item of queued) {
-      item.token.retry(item.message, reason);
+      retryInjectedItem(item, reason);
     }
-    const pending = pendingAckMessages.splice(0);
     for (const item of pending) {
       item.token.retry(item.message, reason);
     }
@@ -1081,6 +1120,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
   }
 
   function buildQuery(sessionId: string, sessionCtx: SessionContext, resume?: string): void {
+    inputRecoveryReason = null;
     inputController = new InputController<PendingSdkInput>();
     abortController = new AbortController();
 
@@ -1431,6 +1471,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
             // never re-dispatches and never gets acked. Per design §4
             // "permanent → ack".
             await ackTurnClose("error", "retry_exhausted_notice_posted");
+            retryBufferedMessages("claude_retry_exhausted_tail_recovery");
             return;
           }
 
@@ -1480,6 +1521,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
             // ack the row would loop in `delivered` forever, deduped on every
             // bind-reset replay. Per design §4 "permanent → ack".
             await ackTurnClose("error", "auto_resume_failed_notice_posted");
+            retryBufferedMessages("claude_auto_resume_failed_tail_recovery");
             return;
           }
         }
--- a/packages/client/src/handlers/claude-code.ts
+++ b/packages/client/src/handlers/claude-code.ts
@@ -47,6 +47,17 @@ import { consumedErrorOutcome } from "./turn-settlement.js";
 
 const MAX_RETRIES = 2;
 
+type PendingAckMessage = {
+  message: SessionMessage;
+  token: DeliveryToken;
+  providerEntered: boolean;
+};
+
+type PendingSdkInput = {
+  sdkMessage: SDKUserMessage;
+  pendingAck: PendingAckMessage | null;
+};
+
 /**
  * Bug 6: thrown by `consumeOutput` when an SDK "success" result message
  * actually contains an API error string (e.g. "API Error: socket
@@ -682,7 +693,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
   let cwd: string | null = null;
   let claudeSessionId: string | null = null;
   let currentQuery: Query | null = null;
-  let inputController: InputController<SDKUserMessage> | null = null;
+  let inputController: InputController<PendingSdkInput> | null = null;
   let abortController: AbortController | null = null;
   let consumerDone: Promise<void> | null = null;
   let retryCount = 0;
@@ -698,7 +709,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
    */
   let chatContextForPrompt: ChatContext | undefined;
   const queuedInjectedMessages: Array<{ message: SessionMessage; token: DeliveryToken }> = [];
-  const pendingAckMessages: Array<{ message: SessionMessage; token: DeliveryToken }> = [];
+  const pendingAckMessages: PendingAckMessage[] = [];
   let injectDrainInProgress = false;
   /**
    * Predeclared source repos the agent config declares at
@@ -711,18 +722,18 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
    */
   let sourceReposForPrompt: PredeclaredSourceRepo[] = [];
   /**
-   * The most recently pushed SDK user message, kept around as the replay
+   * The most recently pushed SDK input envelope, kept around as the replay
    * payload for the transient stream-API retry path. Stashed at every
    * `inputController.push` site (start / resume / inject) so a
    * `claude_socket_closed` retry can re-push it into the rebuilt query —
    * without this, `respawnQuery()` would leave the new `InputController`
    * empty and the SDK subprocess would hang idle (resume mode loads
    * conversation history but still needs a new user message to drive
-   * the next turn). We stash the already-converted SDK form (rather
-   * than the raw `SessionMessage`) so the retry path stays synchronous
-   * and timing-compatible with the existing consumer-loop catch block.
-   * Cleared once `ackTurnClose()` acks the entry — turn fully processed,
-   * no further replay needed.
+   * the next turn). We stash the already-converted SDK form plus its ACK
+   * metadata (rather than the raw `SessionMessage`) so the retry path stays
+   * synchronous and timing-compatible with the existing consumer-loop catch
+   * block. Cleared once `ackTurnClose()` acks the same entry — turn fully
+   * processed, no further replay needed.
    *
    * Invariant — single-consumer loop: this stash is safe ONLY because
    * the catch → respawn → re-push sequence in `consumeOutput` is
@@ -752,7 +763,7 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
    * radius than the duplicate-message symptom. Documented per PR #648
    * reviewer observation #1.
    */
-  let stashedSdkMessage: SDKUserMessage | null = null;
+  let stashedSdkInput: PendingSdkInput | null = null;
 
   async function toSDKUserMessage(
     message: SessionMessage,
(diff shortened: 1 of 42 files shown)


packages/client/src/handlers/claude-code.ts
```diff
@@ -1039,8 +1094,8 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
    */
   function respawnQuery(sessionId: string, sessionCtx: SessionContext): void {
     buildQuery(sessionId, sessionCtx, sessionId);
-    if (stashedSdkMessage) {
-      inputController?.push(stashedSdkMessage);
+    if (stashedSdkInput) {
```
yuezengwu: Blocking: this still replays only one `stashedSdkInput`, but this PR allows multiple pending inputs to become provider-entered before a result closes the turn. Example: SDK pulls `m1` and `m2`, then a transient stream error is thrown before any successful result. `pendingAckMessages` still contains `[m1, m2]`, but `pushPendingSdkInput()` overwrote `stashedSdkInput` with `m2`, so `respawnQuery()` re-pushes only the tail. The next result can then `ackTurnClose()` the whole provider-entered prefix even though the resumed query was only driven by `m2`.

Please either replay/snapshot the full provider-entered batch for retry, or add a regression test proving the Claude SDK resume transcript preserves enough state that replaying only the tail cannot drop the head input in the coalesced + transient-retry path.

yuezengwu (changes requested):
实现目标我理解为：把 Claude Code handler 的 ACK 模型从 `1 input : 1 result` 改成 provider-entered prefix 结算，只有 SDK prompt iterable 实际拉取的输入才标记 processing，并在 SDK result 到达时按一个 provider turn 批量完成，解决 #1156 里 coalesced input 留下 owned+processing orphan 的问题。

核心改动集中在 `PendingSdkInput` / `PendingAckMessage` 包装、`providerPromptInputs()` 拉取时标记 provider-entered，以及 `ackTurnClose()` 对 provider-entered prefix 做 batch `complete()`。没有数据库/schema 变更。

我需要 request changes：当前 batch ACK 路径覆盖了 success/error result 结算，但 transient auto-resume replay 仍是单条 tail stash。这个组合路径和本 PR 的核心语义直接相交，建议修复或至少加回归测试证明安全。

baixiaohang:
Addressed the transient replay gap from the request-changes review.

What changed in the follow-up commit `3100e5ad`:

- Split ACK eligibility from retry replay state.
- `PendingAckMessage.providerEntered` is still set only when the SDK prompt iterable actually pulls the input, and ACK settlement still uses only the provider-entered prefix / explicit result snapshot.
- Transient retry replay now uses `unclosedSdkInputs`, populated when an input is pushed into the active `InputController`, so pushed-but-not-consumed tail inputs survive `respawnQuery()`.
- `ackTurnClose()` removes replay entries only after `complete(messages)` succeeds and only for the settled batch.
- Added regression coverage for both provider-entered coalesced replay and the `m1 consumed + m2 pushed but not consumed` transient retry case.

Verification:

- `pnpm exec biome check --write packages/client/src/handlers/claude-code.ts packages/client/src/__tests__/claude-code-stream-error-retry-replay.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-stream-error-retry-replay.test.ts --maxWorkers=2 --minWorkers=1` — 1 file / 3 tests passed
- Related Claude Code turn tests — 7 files / 12 tests passed
- `pnpm check` — exit 0; existing unrelated Biome warnings remain
- `pnpm typecheck`
- Full client Vitest — 100 files / 1039 tests passed
- `git diff --check`

`codex-reviewer` re-reviewed the updated working tree and found no blocking issues.

packages/client/src/handlers/claude-code.ts
```diff
@@ -1410,13 +1435,13 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
           }
 
           // Automatic retry — rebuild the SDK query in resume mode AND re-push
-          // the pending user message into the freshly built InputController.
+          // the unclosed inputs into the freshly built InputController.
```
yuezengwu, line 1501: Blocking: the retry replay buffer is fixed, but the fatal return paths still need to drain/recover any remaining unentered tail. Example: `m1` and `m2` are pushed, each retry re-consumes only `m1`, then retries exhaust. The branch just above this comment calls `ackTurnClose("error", "retry_exhausted_notice_posted")`, which ACKs/removes only the provider-entered prefix (`m1`) and leaves `m2` in `pendingAckMessages` / `unclosedSdkInputs` without calling `retry()`. The consumer loop has returned while the session still looks active, so later input can be injected into a dead query and `m2` waits for an indirect idle/suspend recovery path.

Please explicitly retry/recover the remaining pending/queued inputs after settling the provider-entered prefix on fatal return paths. The same issue applies to the auto-resume-failed branch at `ackTurnClose("error", "auto_resume_failed_notice_posted")`.

yuezengwu (changes requested):
这轮 follow-up 的实现目标我理解为：保留 provider-entered prefix 的 ACK 结算，同时把 transient retry replay 从单条 stash 扩展成 `unclosedSdkInputs` 队列，覆盖“已 provider-entered 的 coalesced batch”和“已 push 但尚未 provider-entered 的 tail”两类重放场景。

核心改动方向是对的：`respawnQuery()` 现在按顺序重放 unclosed buffer，`PendingAckMessage.providerEntered` 仍然只在 SDK prompt iterable 拉取时置位，`ackTurnClose()` 只结算 provider-entered prefix。没有数据库/schema 变更；仍是 Claude Code handler 内部 ACK/retry ledger 语义调整。

我需要继续 request changes：旧 blocker 已解决，我已把旧 thread resolved；但 terminal failure return path 现在会留下一个新的 tail recovery gap。focused tests 和 typecheck 都通过了，adversarial review 也指向同一个问题。

baixiaohang:
Addressed the terminal failure tail recovery gap from the latest request-changes review.

Follow-up commit `9e12b7d9` adds recovery coverage for the injected-message drain window after an item has been shifted out of `queuedInjectedMessages` but before it reaches `pendingAckMessages` / `unclosedSdkInputs`.

What changed:

- `scheduleInjectedMessagesDrain()` now tracks the currently shifted injected item as `drainingInjectedMessage`.
- Fatal/suspend recovery records an `inputRecoveryReason` and retries queued, pending, and currently draining inputs.
- `pushInjectedMessage()` checks for closed/recovered input state after each async boundary and before `pushPendingSdkInput()`, so an old async continuation cannot write a recovered tail back into local pending/unclosed state.
- Draining items that already reached `pendingAckMessages` are de-duplicated by message+token, and each queued/draining envelope has a retry guard.
- New `buildQuery()` clears the global recovery marker for the next live query, while already recovered old drain items keep their per-item recovery reason.
- Added regression coverage for the exact retry-exhausted race: `m1` provider-entered, `m2` shifted from the inject queue with formatting suspended, then terminal failure completes `m1` and retries `m2` without leaving a local orphan. Also added auto-resume-fail tail recovery coverage.

Verification:

- `pnpm exec biome check --write packages/client/src/handlers/claude-code.ts packages/client/src/__tests__/claude-code-turn-end-retry-exhausted.test.ts packages/client/src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts packages/client/src/__tests__/claude-code-stream-error-retry-replay.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-stream-error-retry-replay.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts --maxWorkers=2 --minWorkers=1` — 3 files / 8 tests passed
- Related Claude Code turn/ACK suite — 8 files / 17 tests passed
- `pnpm check && pnpm typecheck` — exit 0; existing unrelated Biome warnings remain
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1` — 100 files / 1041 tests passed
- `git diff --check`

`codex-reviewer` re-reviewed this version and found no new blocking issues.

packages/client/src/handlers/claude-code.ts
```diff
@@ -1406,17 +1471,18 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
             // never re-dispatches and never gets acked. Per design §4
             // "permanent → ack".
             await ackTurnClose("error", "retry_exhausted_notice_posted");
+            retryBufferedMessages("claude_retry_exhausted_tail_recovery");
```
yuezengwu, line 1495: Blocking: this now retries the local queued/pending/draining tail, but the same-socket recovery path still routes the recovered row back into this same handler instance after `consumeOutput()` returns. The fatal branch leaves `claudeSessionId`, `ctx`, `inputController`, and the SessionManager entry looking active; `SessionManager.routeMessage()` injects recovered or queued work into active sessions, and `inject()` still returns `{ kind: "owned", mode: "queued" }`.

Concrete sequence: `m1` is provider-entered, `m2` is retried here, server recovery redelivers `m2`, SessionManager sees the existing active session and calls `handler.inject(m2)`. Because the old consumer loop is gone, either `inputRecoveryReason` immediately retries it again or it is pushed into an input controller nobody is consuming. The tail is no longer just a local orphan, but it still cannot make progress without an indirect idle/suspend.

Please make the fatal path close/clear the handler state or notify SessionManager so recovery redelivery resumes through a fresh live query/handler, and add a SessionManager-level regression for retry-exhausted / auto-resume-failed tail redelivery.

yuezengwu (changes requested):
这轮 follow-up 的实现目标我理解为：在保留 provider-entered prefix ACK 结算的基础上，补齐 retry-exhausted / auto-resume-failed terminal path 对 queued / pending / draining tail 的恢复，避免上一轮指出的本地 ACK ledger 残留。

核心改动方向比上一轮更接近目标：`retryBufferedMessages()` 现在覆盖 queued、pending、draining 三个窗口，`pushInjectedMessage()` 在 async 边界后检查 recovery marker，避免旧 drain continuation 把已恢复 tail 重新写回本地 pending/unclosed。没有数据库/schema 变更；仍是 Claude Code handler 内部 ACK/retry/recovery ledger 语义调整。

我需要继续 request changes：这次修掉了“本地 tail 没有 retry”的直接问题，但端到端 recovery redelivery 仍不闭合。fatal branch 调用 `retryBufferedMessages()` 后 `consumeOutput()` 直接 return，handler / SessionManager 仍表现为 active；server 同 socket redelivery 会再注入这个已经没有 consumer loop 的 handler，tail 会再次被 retry 或卡在 dead query，而不是通过 fresh live query 恢复处理。

验证：focused Claude Code retry tests 3 files / 8 tests passed；相关 Claude Code ACK/turn-end suite 8 files / 17 tests passed；`@first-tree/client` typecheck passed；`git diff --check` passed。外部 adversarial review 也指向同一个 active-dead-handler recovery gap。

baixiaohang:
Addressed the active-dead-handler recovery gap from the latest request-changes review.

Follow-up commit `bbe1bb9a` makes the Claude fatal path retire the SessionManager active slot after it has settled the provider-entered prefix and marked all remaining queued/pending/draining tail inputs for recovery.

What changed:

- Added optional `SessionContext.retireActiveSession(reason)` for handler-internal fatal paths that have already handled delivery settlement/recovery.
- `SessionManager` implements this as a local active-slot release: mark the entry `suspended`, decrement active count, clear runtime projection, persist, notify `suspended`, and drain pending slot work. It intentionally does not call `prepareSuspend()` or `handler.suspend()` because the fatal handler path has already completed prefix ACK and tail retry, and the current consumer loop is returning.
- Claude Code retry-exhausted and auto-resume-failed fatal paths now run in this order: batch complete provider-entered prefix, retry queued/pending/draining tail, then retire the active session.
- Same-socket recovery redelivery now re-enters through `resumeSession()` instead of `inject()` into a dead consumer loop.

Regression coverage:

- Added a SessionManager test where `m1` is completed, `m2` is injected then retried as recovery debt, and the handler retires the active session. The first `m2` redelivery triggers `recoverChat`; the accepted recovery redelivery calls `handler.resume()` and does not call `handler.inject()` again.
- Added Claude retry-exhausted / auto-resume-fail assertions that the fatal handler branches call `retireActiveSession(...)`.

Verification:

- `pnpm exec biome check --write packages/client/src/runtime/handler.ts packages/client/src/runtime/session-manager.ts packages/client/src/handlers/claude-code.ts packages/client/src/__tests__/session-manager.test.ts packages/client/src/__tests__/claude-code-turn-end-retry-exhausted.test.ts packages/client/src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts --maxWorkers=2 --minWorkers=1` — 3 files / 63 tests passed
- Related Claude Code turn/ACK suite — 8 files / 17 tests passed
- `pnpm check && pnpm typecheck` — exit 0; existing unrelated Biome warnings remain
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1` — 100 files / 1042 tests passed
- `git diff --check`

`codex-reviewer` re-reviewed this version and found no new blocking issues.

yuezengwu (approved):
这轮 follow-up 的实现目标我理解为：在保留 provider-entered prefix batch ACK 的基础上，让 retry-exhausted / auto-resume-failed fatal path 在结算 provider-entered prefix、恢复 queued/pending/draining tail 后，主动释放 SessionManager 的 active slot，确保同 socket recovery redelivery 走 resume，而不是 inject 回已经返回的 consumer loop。

核心改动方向已经闭合上一轮 blocker：`SessionContext.retireActiveSession()` 只作为 handler 已自行完成 ACK/retry 后的 active-slot release；`SessionManager` 将该 chat 从 active 转为 suspended、清 runtime projection、持久化并 drain pending queue；后续 recovery redelivery 会先触发 recover，再在 accepted redelivery 上进入 `resumeSession()`。新增 SessionManager regression 覆盖了 `m1` completed + `m2` recovery debt 后不再调用 `inject()` 的路径，Claude retry-exhausted / auto-resume-fail 测试也覆盖了 fatal branch 调用 retire hook。

没有数据库/schema 变更；本轮仍是 client runtime / handler 内部 ACK、retry、recovery ledger 语义调整。

验证通过：
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts --maxWorkers=2 --minWorkers=1` — 3 files / 63 tests passed
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-inject-startup-race.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts src/__tests__/claude-code-turn-end.test.ts src/__tests__/claude-code-turn-end-sdk-error.test.ts src/__tests__/claude-code-result-forward-fail.test.ts src/__tests__/claude-code-stream-error-retry-replay.test.ts src/__tests__/claude-code-turn-end-serialization.test.ts --maxWorkers=2 --minWorkers=1` — 8 files / 17 tests passed

外部只读 adversarial review 未发现 blocker。剩余注意点是既有 config hot-switch restart 路径没有复用 `unclosedSdkInputs` replay helper，但这不是本轮 fatal recovery blocker，也不阻挡本 PR。

yuezengwu (approved):
这轮同步后的实现目标我理解为：保持 Claude Code provider-entered prefix batch ACK 语义，同时在 retry-exhausted / auto-resume-failed fatal path 中先结算 provider-entered prefix、恢复 queued/pending/draining tail，再通过 `failSessionForRecovery()` 释放并删除本地 dead handler，让同 socket recovery redelivery 走 fresh handler resume。

核心改动已经闭合上一轮 blocker：`PendingSdkInput` / `PendingAckMessage.providerEntered` 仍然只在 SDK prompt iterable 拉取时进入 processing；transient replay 继续用 `unclosedSdkInputs` 覆盖 pushed-but-unclosed 输入；fatal recovery 现在借助 SessionManager 的 evicted mapping 保存 provider session id，后续 recovery redelivery 先触发 `recoverChat`，accepted redelivery 再通过 `startNewSession()` -> `handler.resume()` 恢复，而不是 inject 回已返回的 consumer loop。

涉及核心 runtime/handler 数据结构与 hook 语义调整，但没有数据库/schema 变更。

验证通过：
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/session-manager.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts --maxWorkers=2 --minWorkers=1` — 3 files / 63 tests passed
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/claude-code-inject-startup-race.test.ts src/__tests__/claude-code-turn-end-retry-exhausted.test.ts src/__tests__/claude-code-turn-end-auto-resume-fail.test.ts src/__tests__/claude-code-turn-end.test.ts src/__tests__/claude-code-turn-end-sdk-error.test.ts src/__tests__/claude-code-result-forward-fail.test.ts src/__tests__/claude-code-stream-error-retry-replay.test.ts src/__tests__/claude-code-turn-end-serialization.test.ts --maxWorkers=2 --minWorkers=1` — 8 files / 17 tests passed

最新 merge-main 后我没有发现新的 blocking issue。

PR #1167, opened 2026-06-18
fix(web): keep failed-send draft rollback in its originating chat

liuchao-001:
Follow-up to #1165 addressing a Codex P2 found on that PR.

## Bug

The composer clears the draft optimistically on send and restores it on failure (`setDraft(previousDraft)` when the live composer is still empty). The in-chat `draft` state is **shared** and `ChatView` isn't remounted on chat switch — so if the user switches from chat A to chat B **during** an in-flight send that then **fails**, the rollback writes chat A's rejected text into chat B's composer, and the per-chat draft hook persists it under **chat B's** key. Chat A's text is misfiled into chat B.

This affects both send paths (text and image/file).

## Fix

- Thread the **send-time `chatId`** through the send context (`sendChatId`), and track the **currently-viewed `chatId`** in a ref.
- New pure helper `parkFailedDraftIfSwitched(userId, sendChatId, currentChatId, text)`:
  - same chat → returns `false`, caller restores into the live composer as before;
  - switched away → **parks** the rejected text in the *originating* chat's own draft cache (`u:<userId>:chat:<sendChatId>`) and returns `true`, so it's neither lost nor leaked into the current composer. The user finds it waiting when they return to that chat.
- Both rollback sites (`sendMut.onError` + the image/file `catch`) go through the helper.

## Testing

- `tsc --noEmit` clean · `biome check` clean · `vitest run` 868 pass, incl. 4 new unit tests for `parkFailedDraftIfSwitched` (same-chat / switched-away park / empty-rollback no-op / per-user scope). The only failing *suite* is the pre-existing `tests/visual-regression.spec.ts` Playwright skeleton, unrelated.
- The decision logic is unit-tested; the ChatView wiring (ctx `sendChatId`, `chatIdRef`) is type-checked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1169, opened 2026-06-18
fix(web): sync the chat-id ref in a layout effect, not during render

liuchao-001:
Follow-up to #1168 resolving a Codex P2 on its head — and, with it, an apparent contradiction between two earlier review rounds.

## Background

`chatIdRef` tracks the currently-viewed chat so an async send-*failure* rollback can tell whether the user has switched chats. Its update timing has been pulled two ways:

- The original passive `useEffect` (#1167) left the ref **stale for the first frame** after a switch — a rollback resolving in that window would mis-decide. #1167's reviewer asked to update it synchronously with render.
- So #1168 wrote `chatIdRef.current = chatId` **during render**. Codex then flagged that a render-phase write also runs for renders React **starts then abandons** (concurrent), leaving the ref pointing at a chat that never committed.

## Fix

`useLayoutEffect` is the synthesis:
- runs **only for committed renders** → concurrent-safe (an abandoned render never runs layout effects), unlike the render-phase write;
- runs **synchronously in the commit phase, before paint** → no async send-failure callback for that frame can fire before it, so there's no stale window, unlike the passive `useEffect`.

Read-only-in-async-callbacks semantics are unchanged.

## Testing

`tsc --noEmit` clean · `biome check` clean · `vitest run` 869 pass (no behavior change to assert beyond existing coverage; this is a render-timing correction). The only failing *suite* is the pre-existing `tests/visual-regression.spec.ts` Playwright skeleton, unrelated.

### DIFF
--- a/packages/web/src/pages/workspace/center/chat-view.tsx
+++ b/packages/web/src/pages/workspace/center/chat-view.tsx
@@ -1083,11 +1083,16 @@ export function ChatView({
   // Always-current chat id for async send rollbacks: a send that FAILS after
   // the user switched chats must restore its rejected text into the originating
   // chat, never the one now in view (the draft state is shared and ChatView
-  // isn't remounted on switch). Written during render (not a passive effect) so
-  // it is already current on the first committed frame after a switch; only
-  // read inside async callbacks, never to compute render output.
+  // isn't remounted on switch). Synced in a layout effect — it runs only for
+  // COMMITTED renders (so a concurrent render React starts then abandons can't
+  // leave the ref pointing at a chat that never committed, which a render-phase
+  // write would) and synchronously in the commit phase before any async
+  // send-failure callback for that frame can fire (so there's no stale window,
+  // unlike a passive effect). Only read inside async callbacks.
   const chatIdRef = useRef(chatId);
-  chatIdRef.current = chatId;
+  useLayoutEffect(() => {
+    chatIdRef.current = chatId;
+  }, [chatId]);
   const [cursor, setCursor] = useState(0);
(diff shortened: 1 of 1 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1169

**Recommendation**: approve
- Reason: moving the current-chat ref update into `useLayoutEffect` preserves first-committed-frame freshness while avoiding render-phase writes from abandoned concurrent renders.

**Level**: B-low
- Semantic lift: touches Workspace chat failed-send rollback timing -> B-low

**PR Summary**
- Problem: `chatIdRef` needs to be current before a failed-send rollback can run, but a render-phase write can leak state from a render React never commits.
- Design: sync `chatIdRef.current` from a layout effect keyed by `chatId`, keeping the update in the commit phase and before paint.
- Affected module: `packages/web/src/pages/workspace/center/chat-view.tsx`

**Review Notes**
✅ 1. The new `useLayoutEffect` update is commit-scoped, so an abandoned concurrent render cannot point rollback logic at a chat that never became visible. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:1093`]
✅ 2. The timing still closes the stale first-frame window from the earlier passive-effect version: layout effects run during commit before paint, while the ref remains read only from async failure paths. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:1572`]

**Executed**
- Approved
- Did not run local checks or tests per PR Review policy; considered the author's reported `tsc`, `biome`, and `vitest` results.
- Context Tree read: `system/cloud/chat/NODE.md`, `system/cloud/chat/messaging.md`, `system/cloud/web-console.md`.

PR #1170, opened 2026-06-18
fix: dedupe runtime capability refresh uploads

baixiaohang:
## Summary
- Add semantic capability snapshot dedupe that ignores volatile probe metadata (`detectedAt`, `latencyMs`) for upload/backoff decisions.
- Stop logging successful background refreshes when no capability state actually changed.
- Cover volatile metadata churn so degraded provider polling backs off instead of uploading every cycle.

## Tests
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/capability-refresh.test.ts --maxWorkers=2 --maxConcurrency=2`
- `pnpm check`
- `pnpm typecheck`

Note: `pnpm check` exits 0 but reports pre-existing warnings outside this change.

### DIFF
--- a/apps/cli/src/__tests__/capability-refresh.test.ts
+++ b/apps/cli/src/__tests__/capability-refresh.test.ts
@@ -1,6 +1,10 @@
 import type { CapabilityEntry, ClientCapabilities } from "@first-tree/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { CapabilityRefresher, type CapabilityRefresherDeps } from "../core/capability-refresh.js";
+import {
+  CapabilityRefresher,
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
已 review。\n\n目标和核心改动：这次把 runtime capability refresh 的上传/退避判断改成语义快照比较，忽略每次 probe 都会变化的 `detectedAt` / `latencyMs`，避免 degraded provider polling 每轮都 PATCH；同时只有真实上传成功时才打印成功刷新日志，并补了 volatile metadata churn 的单测。\n\n检查结果：没有发现必须修复的 blocker。该改动不涉及数据库变更，也没有引入新的核心数据结构/持久化 schema 变更。GitHub CI 当前为 pass。

PR #1179, opened 2026-06-22
feat: add Enter/Esc keyboard shortcuts to the ask takeover card

yuezengwu:
## What

Adds keyboard shortcuts to the blocking **ask takeover card** (`AskTakeover` — the pop-up answer card shown when an agent's `chat ask` blocks this chat for the viewer):

- **Enter** → **Reply** (resolves with the composed answer)
- **Esc** → **Skip** (resolves with a "skipped" answer)

Before this, Skip and Reply were click-only.

## Aligned with the chat composer

Per the request, the Enter behavior mirrors the chat composer's textarea exactly, so the two surfaces feel consistent:

- **Enter** (no Shift, and not mid-IME-composition) resolves with Reply, gated on a valid answer (`canReply`).
- **Shift+Enter** stays a newline in the free-text box.
- A keydown that confirms an **IME candidate** (`isComposing`) never resolves — and Esc during composition cancels the candidate rather than skipping.

## Not stealing keys from a higher overlay

Bound at the `window` (matching `layout.tsx` / `chat-view.tsx`) because the card does not autofocus — so shortcuts work before the user clicks in. To avoid resolving the ask from keys meant for a modal layered above it, the handler bails when:

- the card sits inside an **`aria-hidden`** region — Radix dialogs (the ⌘K command palette, etc.) render in a body-level portal and mark the rest of the tree aria-hidden while open, so this detects a higher overlay holding focus; and
- the keystroke was already consumed (**`e.defaultPrevented`**) by a focused control such as the composer's mention/slash popover.

## Notes

- **Enter on an option row** keeps its native radio/checkbox toggle (not a resolve).
- Both shortcuts stay **inert while a reply is sending** (matching the disabled buttons).
- Adds `title` hints — "Reply (Enter)" / "Skip (Esc)" — for discoverability.
- 7 new DOM tests cover Enter-replies, Esc-skips, Shift+Enter/IME no-ops, option-row toggle, the sending lockout, the aria-hidden (command-palette) gate, and the defaultPrevented gate. `pnpm --filter @first-tree/web typecheck` and `biome check` are clean; the `ask-takeover-dom` suite is green (12/12).

### DIFF
--- a/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
+++ b/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
@@ -43,6 +43,13 @@ async function click(el: Element | null): Promise<void> {
     el.dispatchEvent(new MouseEvent("click", { bubbles: true, cancelable: true }));
   });
 }
+async function keyDown(el: EventTarget, key: string, init: KeyboardEventInit = {}): Promise<KeyboardEvent> {
+  const event = new KeyboardEvent("keydown", { key, bubbles: true, cancelable: true, ...init });
+  await act(async () => {
+    el.dispatchEvent(event);
+  });
+  return event;
+}
 async function setValue(el: HTMLTextAreaElement, value: string): Promise<void> {
   await act(async () => {
     const setter = Object.getOwnPropertyDescriptor(HTMLTextAreaElement.prototype, "value")?.set;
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1179

- Reason: the window-level shortcut can resolve the blocking ask while the user is interacting with another higher-priority overlay.

**Level**: C
- Path initial: only `packages/web/**` changed -> C

**PR Summary**
- Problem solved: adds keyboard shortcuts for replying to or skipping the blocking ask takeover card.
- Design: `AskTakeover` installs a `window` `keydown` listener for Enter/Escape and reuses `canReply` plus `buildResolveAnswer`; tests cover the core shortcut paths.
- Affected modules: `packages/web/src/components/chat/ask-takeover.tsx` and its DOM test.

**Review Comments**
❌ 1. The global handler needs to respect other active overlays before resolving the ask. [`packages/web/src/components/chat/ask-takeover.tsx:110`]

Because this listener is attached to `window`, it also receives key events from UI rendered above the ask card. The command palette is still globally reachable with Cmd/Ctrl+K while the ask takeover is mounted; after typing a valid answer, pressing Enter in the palette can both select/navigate and call `onReply`, and pressing Escape to close the palette can also call `onSkip`. That turns ordinary palette keyboard use into an accidental request resolution. Please gate the shortcut on whether the event was already handled and whether the active target belongs to the ask takeover surface, or otherwise suppress it while another modal/popup owns the keystroke; add a regression test for the command-palette or equivalent higher-overlay case.

**Executed**
- Submitted request_changes.
- Did not run tests, per PR review mode.

yuezengwu:
Thanks @codex-assistant — good catch, fixed in fb8074d3.

The window-level listener now bails before resolving when a higher overlay owns the keystroke:

1. **`aria-hidden` ancestor check** — Radix dialogs (the ⌘K command palette via `CommandDialog` → `Dialog`/`DialogContent`, and any other Radix dialog) render in a body-level portal and mark the rest of the tree `aria-hidden` while open. The handler bails when `cardRef.current.closest('[aria-hidden="true"]')` is truthy, i.e. when a modal is layered above the card. This is the primary guard for the palette case you described — Enter to pick a palette item, or Esc to close it, no longer resolves the ask underneath.
2. **`e.defaultPrevented` check** — yields to a focused control that already consumed the keystroke (e.g. the composer's mention/slash autocomplete, which preventDefaults Enter/Escape).

Regression tests added:
- `stays inert while a higher overlay owns the keystroke (aria-hidden region)` — marks the card's container `aria-hidden` (exactly what Radix's focus scope does to the rest of the tree), asserts Enter/Esc are inert, then clears it and asserts the shortcuts resume.
- `yields to a focused control that already consumed the keystroke (defaultPrevented)`.

Suite is green (12/12); `typecheck` + `biome check` clean.

baixiaohang (approved):
## Review by codex-assistant - PR #1179

**Recommendation**: approve
- Reason: the follow-up fix addresses the previous accidental-resolution path without changing request resolution semantics.

**Level**: C
- Path initial: only `packages/web/**` changed -> C

**PR Summary**
- Problem solved: adds Enter-to-reply and Esc-to-skip shortcuts for the blocking ask takeover card.
- Design: `AskTakeover` uses a window-level keydown listener so shortcuts work before focus enters the card, reuses `canReply` / `buildResolveAnswer`, and now yields when a higher overlay owns the key via `aria-hidden` or `defaultPrevented`.
- Affected modules: `packages/web/src/components/chat/ask-takeover.tsx` and its DOM test.

**Review Comments**
✅ 1. The previous blocker is resolved: the handler now bails when the ask card sits under an `aria-hidden` ancestor, covering Radix dialogs such as the command palette, and also respects already-consumed events via `e.defaultPrevented`.
✅ 2. The shortcut path shares the existing reply gate and answer formatter, so it does not introduce a second resolution format or bypass the sending lockout.

**Executed**
- Submitted approve review.
- Did not run tests, per PR review mode.

PR #1185, opened 2026-06-22
fix: stop retrying codex deterministic compact failures

baixiaohang:
## Summary

- Stop routing Codex app-server deterministic turn failures through delivery retry/recovery.
- Terminal-reject deterministic Codex failures, including context-window / remote compact failures, with provider turn evidence.
- Catch the current pre-sampling compact failure shape where the app-server reports `status: "completed"` with no assistant output but stderr contains `Failed to run pre-sampling compact`.
- Preserve transient and unknown-custody retry behavior, including ordinary transport-close recovery and completed-empty successful silence without compact diagnostics.
- Add Codex app-server handler regressions for deterministic compact failures, completed-empty compact diagnostics, prior SDK compact errors followed by transport close, transient retry, ordinary transport close, and successful silence.

## Context

Codex pre-sampling remote compact can fail deterministically when the current thread exceeds or cannot fit within the model context window. The app-server handler already classified several provider failures as deterministic, but settlement still converted every turn failure into a `retryReason`, which sent the same input back through inbox recovery. For compact/context-window failures, replaying the same input into the same thread repeats the same provider failure and can produce a stream of visible SDK errors.

The app-server can also surface compact failure as a completed-but-empty turn with no wire-visible error. This PR treats that shape as terminal only when stderr or a prior turn diagnostic explicitly contains compact failure evidence, so normal empty successful turns remain allowed.

This change keeps the fix scoped to the Codex app-server handler. It does not change generic inbox recovery, retryCount, Claude handling, or Web error rendering.

## Risk

- Deterministic classification remains conservative: ordinary `transport is closed`, timeout, and connection reset paths still retry/recover.
- Completed-empty turns without compact diagnostics still settle as successful silence.
- A terminal-rejected close may leave the closed app-server object in place until the next recovery/cleanup path. That can be optimized later, but it does not redeliver the already terminal-rejected input.

## Verification

- `pnpm --filter @first-tree/client test -- codex-app-server-handler` (passed; ran client suite: 102 files / 1067 tests)
- `pnpm check` (exit 0; existing warnings only in untouched files)
- `git diff --check`

### DIFF
--- a/packages/client/src/handlers/codex/app-server/index.ts
+++ b/packages/client/src/handlers/codex/app-server/index.ts
@@ -111,6 +111,8 @@ const USAGE_LIMIT_NOTICE =
 const CODEX_CONTEXT_WINDOW_FAILURE_MESSAGE =
   "Codex ran out of room in the model context while compacting this thread. " +
   "Start a new thread or clear earlier history before retrying.";
+const CODEX_COMPACT_FAILURE_MESSAGE =
+  "Codex failed to compact this thread before answering. Start a new thread or clear earlier history before retrying.";
 
 export const createCodexAppServerHandler: HandlerFactory = (config: HandlerConfig): AgentHandler => {
   const workspaceRoot = config.workspaceRoot as string;
@@ -762,6 +764,7 @@ export const createCodexAppServerHandler: HandlerFactory = (config: HandlerConfi
     }
 
     const completedSuccessfully = turn.providerCompleted && turn.failure === null && turn.status === "completed";
+    const finalAgentText = turn.finalAgentText.trim();
     const usage = turn.usageLast;
     const zeroTokenDelta =
       usage !== null &&
@@ -769,7 +772,11 @@ export const createCodexAppServerHandler: HandlerFactory = (config: HandlerConfi
       usage.cachedInputTokens === 0 &&
       usage.outputTokens === 0 &&
       usage.reasoningOutputTokens === 0;
-    const usageLimitEmptyTurn = completedSuccessfully && turn.finalAgentText.trim().length === 0 && zeroTokenDelta;
+    const usageLimitEmptyTurn = completedSuccessfully && finalAgentText.length === 0 && zeroTokenDelta;
+    const completedEmptyCompactFailure =
+      completedSuccessfully && finalAgentText.length === 0
+        ? completedEmptyCompactFailureInfo(turn, appServer?.stderr ?? "")
+        : null;
     const usageLimitFailure = turn.failure ? isUsageLimitErrorInfo(turn.failure.codexErrorInfo) : false;
 
     let forwardFailed = false;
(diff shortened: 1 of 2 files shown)


packages/client/src/handlers/codex/app-server/index.ts
```diff
@@ -821,12 +841,25 @@ export const createCodexAppServerHandler: HandlerFactory = (config: HandlerConfi
...
+        recordId: turn.turnId,
+      });
     } else {
-      turn.primaryToken.retry(turn.acceptedMessages, settlement.action.reason);
+      const settlement = resolveTurnSettlement({
+        retryReason,
+        consumedErrorReason,
+        forwardFailed,
+      });
+      sessionCtx.emitEvent({ kind: "turn_end", payload: { status: settlement.status } });
```
yuezengwu, line 873: Blocker: this still misses the current app-server compact-failure shape where `turn/completed` arrives as `status: "completed"` with no assistant item/error (`last_agent_message=null` on the app-server side). In that case `completedSuccessfully` is true, `turn.finalAgentText` is empty, `turn.failure` is null, and `providerCompleted` is true, so this falls through with no `retryReason`/`terminalRejectionReason`; `resolveTurnSettlement` then returns success and the input is ACKed. OpenAI tracks this exact pre-sampling compact behavior in https://github.com/openai/codex/issues/25619, and current `run_pre_sampling_compact` still returns `Ok(None)` after logging the compact failure. Please add a regression for the silent completed/no-output compact failure and make this path terminal-reject when there is deterministic evidence, or at least retry instead of completing successfully when custody is not provably consumed.

yuezengwu (changes requested):
Goal/core change: this PR moves deterministic Codex app-server turn failures off the retry/recovery path and into `DeliveryToken.terminalRejected`, while preserving transient and unknown transport-close retries.

I found one blocker: the implementation still misses the current app-server pre-sampling compact failure shape where the turn completes with `status: "completed"`, no assistant item, and no wire-visible error. That path can still ACK the failed input as success, so the PR does not fully stop deterministic compact failures from being consumed incorrectly.

baixiaohang:
Addressed the blocker in 151ad2fe.

What changed:
- Added a completed-empty compact diagnostic path before usage-limit/success settlement.
- If a Codex app-server turn completes with `status: "completed"`, no assistant text, and the app-server stderr or prior turn diagnostic contains `Failed to run pre-sampling compact`, the handler now emits one SDK error and terminal-rejects the accepted delivery instead of ACKing it as success.
- Kept the scope conservative: completed-empty turns without compact diagnostics still settle as successful silence; transient/unknown transport-close paths remain retryable.

Added regressions:
- completed empty + pre-sampling compact stderr => `terminalRejected(..., "codex_compact_failure", server_terminal_record turnId)`
- completed empty with no compact diagnostics => successful silence

Verification after the fix:
- `pnpm --filter @first-tree/client test -- codex-app-server-handler` passed (client suite: 102 files / 1067 tests)
- `pnpm check` passed with exit 0; only existing warnings in untouched files
- `git diff --check` passed

yuezengwu (approved):
Goal/core change: Codex app-server deterministic compact/context-window failures now terminal-reject accepted delivery with provider turn evidence instead of flowing through retry/recovery, while transient/unknown transport-close paths stay retryable.

The 151ad2fe update addresses my previous blocker: completed-empty turns with explicit pre-sampling compact diagnostics are handled before usage-limit/success settlement, and the reverse case (completed empty without compact diagnostics) remains successful silence. The added regressions cover both paths. No DB/schema change involved.

I reviewed the updated diff and ran `git diff --check` locally. I did not rerun the pnpm suite in this review worktree because it has no `node_modules`/`vitest`; the PR reports the targeted test, typecheck, and `pnpm check` passing.

PR #1188, opened 2026-06-22
feat(web): clearer runtime install guidance (TUI→CLI rename, OS-specific tmux hints)

yuezengwu:
## What

Tightens how the web surfaces a runtime that isn't ready to use, driven by three concrete pain points: the display name was confusing, the `missing` state gave no actionable hint, and the tmux requirement was never shown as a real command.

Three commits:

1. **Rename `claude-code-tui` display label → "Claude Code CLI".** The provider drives the real `claude` CLI inside tmux, so "Claude Code CLI" reads truer than "Claude Code (TUI)". Display-only: the `claude-code-tui` enum id (wire/DB) is untouched, so no migration. Covers the 3 web label sites + the client-side re-auth error string.

2. **Specific install hint for a missing runtime.** The `missing` state line read a flat "not installed", which is wrong for `claude-code-tui` — it needs both the `claude` CLI and tmux (>= 3.0), so a machine that already had Claude Code and only lacked tmux was told it wasn't installed at all. `providerInstallHint` now narrows on the probe's verbatim resolve reason ("tmux not found" / "`claude` not found …") and names only what's actually missing (tmux-only / CLI-only / both). Applied to the card state line, the legacy `ProviderRow`, and the Setup-incomplete install box.

3. **OS-specific tmux command.** tmux isn't an npm package, so the command depends on the host. New `tmuxInstallCommand(os)` keys off the client's reported OS:
   - macOS → `brew install tmux`
   - Linux → `sudo apt install tmux`
   - Windows → `wsl sudo apt install tmux` (no native Windows tmux; runs under WSL)

   Wired into the state-line hint, the install-box headline, and the copy-paste command block (`os` threaded from `client.os`).

## Scope notes

- The capability **probe semantics are unchanged** — the launch-verified real smoke stays the only path to `ok` (that behavior is expected). This PR is display/guidance only.
- No schema, server, or wire changes. `packages/web` only, plus one client-side error-message string.

## Test

- New/updated `runtime-state-line` tests cover tmux-only, CLI-only, both-missing, and OS-keyed tmux commands (macOS→brew, Linux→apt).
- `pnpm --filter @first-tree/web typecheck` ✅, `biome` ✅
- web suites: runtime-state-line (6), view-models (16), clients-page-cards-dom (3), page-ssr-smoke + web-dom-interactions (37) ✅

### DIFF
--- a/packages/client/src/runtime/error-taxonomy.ts
+++ b/packages/client/src/runtime/error-taxonomy.ts
@@ -244,7 +244,7 @@ export function classify(err: unknown, context?: { source?: ErrorSource }): Clas
       kind: ERROR_KINDS.PERMANENT,
       strategy: NONE,
       reasonCode: "claude_login_required",
-      message: shape.message ?? "Claude TUI requires re-authentication (run /login)",
+      message: shape.message ?? "Claude Code CLI requires re-authentication (run /login)",
     };
   }
   if (isCodexBinaryMissingError(err)) {
--- a/packages/web/src/components/new-agent-dialog.tsx
+++ b/packages/web/src/components/new-agent-dialog.tsx
@@ -144,7 +144,7 @@ function asRuntimeProvider(provider: string): RuntimeProvider | null {
 
 /**
  * Pick the preferred runtime among the ones in `ok` state on a given
- * client. Claude Code wins over Claude Code (TUI) which wins over Codex;
+ * client. Claude Code wins over Claude Code CLI which wins over Codex;
  * if none of those is ok we fall back to whatever else the client reports
  * as ok (still narrowed to a known RuntimeProvider), then `null`.
  */
@@ -163,7 +163,7 @@ function pickPreferredRuntime(caps: ClientCapabilities): RuntimeProvider | null 
 
 function prettyRuntimeLabel(provider: RuntimeProvider): string {
   if (provider === "claude-code") return "Claude Code";
-  if (provider === "claude-code-tui") return "Claude Code (TUI)";
+  if (provider === "claude-code-tui") return "Claude Code CLI";
   if (provider === "codex") return "Codex";
   return provider;
 }
--- a/packages/web/src/pages/agent-detail/runtime-section.tsx
+++ b/packages/web/src/pages/agent-detail/runtime-section.tsx
@@ -28,7 +28,7 @@ const RUNTIME_COPY: Record<RuntimeProvider, { name: string; caption: string }> =
     caption: "Runs through Anthropic's Claude Code runtime.",
   },
   "claude-code-tui": {
-    name: "Claude Code (TUI)",
+    name: "Claude Code CLI",
     caption: "Runs Claude Code through tmux for terminal-native sessions.",
   },
   codex: {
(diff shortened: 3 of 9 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1188

**建议**: approve
- 理由: Display/guidance-only changes align with the runtime-provider contract for `claude-code-tui` (`claude` CLI + tmux >= 3.0), with no schema, wire, or probe-semantic change.

**分级**: A
- Path 初判: contains `packages/client/**` -> A
- 语义升级: none; the client change is a user-facing re-auth message, and web changes do not touch subscriptions, WS frame parsing, or data-layer contracts.

**PR 摘要**
- 解决问题: clearer runtime setup guidance for the tmux-driven Claude runtime.
- 设计方案: rename the display label to "Claude Code CLI", derive missing-runtime hints from the probe resolve reason, and key tmux install commands from `client.os`.
- 受影响模块: `packages/web/src/pages/clients/**`, new-agent/runtime display copy, and one client runtime error string.

**Review 意见**
✅ 1. The enum/wire id remains `claude-code-tui`, and the capability probe semantics remain launch-verified; this is consistent with the runtime-provider contract.
✅ 2. The missing-state hint uses the probe's resolve-stage error to distinguish missing `claude` vs missing tmux, avoiding the previous misleading “not installed” copy when only tmux is absent.
⚠️ 3. Non-blocking follow-up: `tmuxInstallCommand(null/unknown)` falls back to `sudo apt install tmux`; current clients send `process.platform`, but older/unknown rows may show a Linux command without OS evidence. Consider a generic fallback later.

**已执行**

yuezengwu:
Addressed the non-blocking review item (#3, OS fallback) in 487e2bbf: `tmuxInstallCommand` now lists `linux` explicitly and returns `null` for an unknown/unreported OS. In that case the UI names the tmux requirement ("install tmux (>= 3.0) with your package manager") instead of guessing `sudo apt install tmux` without OS evidence. Added a unit test for the unknown-OS path; typecheck + biome + web suites green.

baixiaohang (approved):
## Review by codex-assistant - PR #1188

**建议**: approve
- 理由: The new head only fixes the unknown-OS tmux fallback from the prior non-blocking note, and keeps the change display/guidance-only with no schema, wire, or probe-semantic change.

**分级**: A
- Path 初判: contains `packages/client/**` -> A
- 语义升级: none; the client change remains a user-facing re-auth message, and web changes do not touch subscriptions, WS frame parsing, or data-layer contracts.

**PR 摘要**
- 解决问题: clearer runtime setup guidance for the tmux-driven Claude runtime, plus a safe fallback when the client's OS is unknown.
- 设计方案: keep the “Claude Code CLI” display rename, derive missing-runtime hints from the probe resolve reason, use OS-specific tmux install commands when OS is known, and fall back to a generic package-manager instruction when OS is unknown.
- 受影响模块: `packages/web/src/pages/clients/**`, new-agent/runtime display copy, and one client runtime error string.

**Review 意见**
✅ 1. Follow-up commit `487e2bbf` resolves the prior non-blocking item: `tmuxInstallCommand` now returns `null` for unknown/unreported OS instead of guessing `sudo apt install tmux`.
✅ 2. Callers handle the unknown-OS path with generic package-manager copy in both the runtime state line and setup command block.
✅ 3. The runtime-provider contract remains intact: `claude-code-tui` still means the real `claude` CLI plus tmux >= 3.0, and capability probe semantics are unchanged.

**已执行**

PR #1189, opened 2026-06-22
fix(web): handle inaccessible chat deep links

### LINKED ISSUE #490: Context tab: manual /?c=<id> URL to inaccessible chat hangs on 'Loading chat…'
## Background

Follow-up split out of #489.

#489 fixed the **Context Usage Feed** by gating the `#<chat>` deep link with a server-computed `viewerCanAccess` flag — non-members now see inert text instead of a clickable link that would 404.

However, the underlying chat-view empty-state issue remains: if a viewer **manually** types or pastes a URL like `/?c=<chatId>` for a chat they cannot access (or follows an out-of-band link), the server returns 404 via `requireChatAccess`, but the chat view has no error branch and renders **"Loading chat…" indefinitely**.

## Proposal

Add a 404 (and generic error) empty-state to the chat view route so a manual / stale / shared inaccessible URL renders an explanatory empty state instead of hanging forever. Suggested copy: e.g. "This chat doesn't exist or you don't have access." with a CTA back to the home view.

## Why low-priority

- Discoverable surface (the feed link) is already fixed in #489, so the trigger is now \"user manually constructs URL\" / \"someone shared a stale link\" — low frequency.
- Pure UX issue, no security or data-correctness risk.

## Related

- PR #489 — the fix that scoped this out: https://github.com/agent-team-foundation/first-tree-hub/pull/489

baixiaohang:
## Summary

- Show a chat unavailable empty state when a manual `?c=<chatId>` deep link cannot load chat detail.
- Add a Back to conversations action that clears the selected chat URL state.
- Cover the rejected chat-detail query path in `ChatByIdView` DOM tests.

Fixes #490

## Tests

- `pnpm check`
- `pnpm typecheck`
- `pnpm exec turbo run test --concurrency=2`

### DIFF
--- a/packages/web/src/pages/workspace/center/__tests__/chat-by-id-center-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-by-id-center-dom.test.tsx
@@ -343,6 +343,28 @@ describe("ChatByIdView and CenterPanel", () => {
     await act(async () => loading.root.unmount());
   });
 
+  it("shows an unavailable empty state when a manual chat URL cannot be loaded", async () => {
+    chatMocks.getChat.mockRejectedValueOnce(new Error("not found"));
+    const onClearChat = vi.fn();
+    const { ChatByIdView } = await import("../chat-by-id.js");
+    const { container, root } = await renderDom(
+      <ChatByIdView chatId="missing-chat" narrow={false} onShowConversations={null} onClearChat={onClearChat} />,
+    );
+
+    await waitForText(container, "This chat doesn't exist or you don't have access.");
+    expect(container.textContent).toContain("Chat unavailable");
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed.

Goal/core change: manual `/?c=<chatId>` deep links that reject from `getChat` now render a chat-unavailable empty state instead of staying on `Loading chat...`, and the Back to conversations CTA clears the selected chat/doc-preview URL state through `WorkspacePage`. The DOM coverage exercises the rejected chat-detail query path and the `CenterPanel` prop threading.

Verification:
- `pnpm --filter @first-tree/web test -- chat-by-id-center-dom.test.tsx` (the package script ran the full web Vitest suite: 118 files / 884 tests)

No blocking findings.

PR #1194, opened 2026-06-22
feat: in-product browser-OAuth Connect for Claude/Codex SDK (no separate CLI)

yuezengwu:
## What

Adds an in-product **Connect** flow on the Computers cards that authenticates a
runtime's credentials on the daemon host without the operator installing a
separate `claude` / `codex` CLI. One click runs the provider's official
**browser OAuth** on the host; the user authorizes in the page that opens,
the CLI's own localhost callback writes the local credential, the daemon
re-probes, and the card flips to connected. Consistent for **Claude Code** and
**Codex**.

Design: `designs/no-cli-sdk-auth.md` (see §10 for the browser-OAuth decision).

## Why

Onboarding users had runtimes that "couldn't find an Agent runtime" — the real
gap was **no SDK-readable credentials on the host**, not a missing binary (the
client already bundles both runtimes). Telling users to "go install/login a
CLI" is the wrong experience; this brings the official login in-product.

## Approach

- **Primary = browser OAuth**, consistent across providers:
  - Codex: bare `codex login` → writes `~/.codex/auth.json`.
  - Claude Code: `claude auth login` → writes keychain `Claude Code-credentials`
    (the same entry the Agent SDK reads; the TUI runtime shares it).
- **Fallback = device code** (codex `login --device-auth`) for headless hosts
  with no browser — secondary, not the main visual.
- **Security boundary**: the OAuth dance is browser ↔ provider ↔ the CLI's own
  localhost callback, entirely on the host. **First Tree's server never sees,
  stores, or relays the access/refresh token.** Over the wire it carries only a
  reverse "start" command and the capability status — no token. (QA secret-scan
  confirmed no token in server-visible payloads.)

## How it's wired (reuses existing pipeline)

- `shared`: capability `pendingAuth` ({method: "browser" | "device-code", …})
  rides the capabilities snapshot the daemon already PATCHes; a `runtime-auth`
  command contract + a `POST /clients/:id/runtime-auth/start` route.
- `client`: a provider-agnostic login module (`runtime-login.ts`) +
  `codex-device-auth.ts` / `claude-login.ts` runners.
- `cli`: an orchestrator that runs the login on the host and re-probes;
  serialized per provider (duplicate starts ignored); the background
  capability poll preserves an in-flight login's pending marker.
- `web`: provider cards render a single **Connect** + "finish in your browser"
  state; the probe stays the single source of truth (web polls capabilities,
  no bespoke realtime channel). Manual "Run `<cli> login`" hints are dropped
  for in-product providers (incl. claude-code-tui, which shares Claude auth).

## Validation

- **QA real-machine PASS** (3 reports): Codex browser OAuth end-to-end
  (connected-flip + secret-scan clean); Claude SDK browser OAuth end-to-end
  (keychain written, `✓ Claude Code`, no token leak); claude-code-tui shows one
  shared Claude auth state after the SDK login.
- Full suites green: shared 473, client 1092, cli 504, web 900; typecheck 12/12;
  biome clean.

## Context Tree

Companion tree PR updates `system/cloud/runtime/client-runtime.md` and
`system/cloud/onboarding.md` with the browser-OAuth / zero-token-custody /
single-Claude-auth decisions: agent-team-foundation/first-tree-context#581


baixiaohang (changes requested):
## Review by codex-assistant - PR #1194

- Reason: The browser-OAuth fallback/recovery path is not actually reachable from the product, so headless or no-auto-open hosts can get stuck without the promised device-code/link recovery.

**Level**: A
- Semantic lift: runtime-auth/capability auth pipeline is core runtime behavior; already A

**PR Summary**
- Problem solved: let users connect Claude Code / Codex credentials from the Computers card instead of manually running a separate login flow.
- Design: server forwards a `runtime-auth:start` command, daemon runs provider login on the host, capability snapshots carry `pendingAuth`, and web polls capabilities.
- Affected modules: `packages/shared`, `packages/client`, `apps/cli`, `packages/server`, `packages/web`.

**Review Comments**
❌ 1. The headless / no-auto-open fallback is not wired through to the UI. [R5 / `apps/cli/src/core/runtime-auth-login.ts:141`, `packages/web/src/pages/clients/cards/shared/runtime-auth-controls.tsx:62`]

`runBrowserLogin` already exposes `onAuthUrl` specifically so the web can offer a fallback link when the host browser does not auto-launch, but `runCodexBrowserFlow` calls it as `runBrowserLogin({ binary })`, so that URL is dropped. The web Connect button also always calls `startRuntimeAuth(clientId, { provider })`; there is no product path that passes `method: "device-auth"`. As a result, on a headless host or a machine where the browser fails to open, the card says a sign-in page opened and just waits until the pending marker expires. The advertised fallback exists only behind the API, not in the user flow. Please either carry the browser URL through `pendingAuth` and render it, or add an explicit headless/device-code action in the card, with tests covering the no-auto-open/headless recovery path.

⚠️ 2. Unsupported providers can still return `started: true`. [R4 / `packages/shared/src/schemas/runtime-auth.ts:48`, `apps/cli/src/core/runtime-auth-login.ts:84`]

The request schema accepts the full `runtimeProviderSchema`, including `claude-code-tui`, but the daemon treats anything except `codex` and `claude-code` as unsupported and only logs. A caller can therefore receive `200 { started: true }` for a provider that will never publish pending auth or change capability state. The UI currently avoids this, but the API contract should either reject unsupported Connect targets before sending the frame or reflect a specific unsupported state back into capabilities.

**Executed**
- Submitted request_changes.

yuezengwu:
Thanks — both findings were real. Fixed in `11ca70a3`.

**R5 (blocking) — headless / no-auto-open recovery is now in the product flow, not just the API:**
- The orchestrator now wires `onAuthUrl` through: the provider's sign-in URL is carried into `pendingAuth.authUrl` once the login process prints it, and the card renders a **"Didn't open? Open the sign-in page →"** link — covering the case where the host browser doesn't auto-launch.
- The card also offers a codex-only **"No browser on this computer? Use a one-time code instead"** action that starts the device-code login (`method: "device-auth"`), so a genuinely headless host has a working path instead of waiting until the pending marker expires. (Claude has no device-code equivalent, so the affordance is codex-only.)
- Tests: cli asserts the auth URL lands in `pendingAuth` when the login emits it; web asserts `browser-pending` carries `authUrl`.

**R4 — unsupported Connect targets are rejected before any frame is sent:**
- The `runtime-auth:start` request/command `provider` is now `runtimeAuthProviderSchema` (`claude-code | codex`) instead of the full provider enum. `claude-code-tui` (whose credentials come from the `claude-code` login, never a separate Connect) and anything else now get **400 with no frame sent**, rather than `started: true` for a no-op.
- Test: server returns 400 for `claude-code-tui` and does not call `ws.send`.

Full suites green: shared 473, cli 505, web 901, server 1493; typecheck 12/12; biome clean.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1194 Follow-up

- Reason: The original unsupported-provider issue is fixed, and the auth URL fallback is now surfaced, but the new device-code fallback shown during browser-pending is a dead action because the daemon ignores duplicate starts for the same provider.

**Level**: A
- Semantic lift: runtime-auth/capability auth pipeline is core runtime behavior; already A

**PR Summary**
- Problem solved: in-product browser OAuth Connect for Codex and Claude Code, with capability polling as the source of truth.
- Follow-up scope: `11ca70a3` carries `pendingAuth.authUrl`, adds a codex device-code fallback action, and narrows runtime-auth providers to `claude-code | codex`.
- Affected modules: `apps/cli`, `packages/shared`, `packages/server`, `packages/web`.

**Review Comments**
❌ 1. The browser-pending device-code fallback is rendered but cannot start. [R4 / `packages/web/src/pages/clients/cards/shared/runtime-auth-controls.tsx:116`, `apps/cli/src/commands/daemon/start.ts:283`]

The new codex fallback button is shown both before login and while `view.kind === "browser-pending"`. In the browser-pending case, clicking **"No browser on this computer? Use a one-time code instead."** sends a second `runtime-auth:start` for the same provider with `method: "device-auth"`. But the daemon has already marked `codex` interactive for the browser login, and `start.ts` drops any same-provider start while `capabilityRefresher.isInteractive(command.provider)` is true. So the exact recovery path a user needs after clicking Connect and discovering the host browser did not open is a no-op until the browser attempt exits/expires. Please either make this button cancel/supersede the in-flight browser login and start device-auth, or do not render it during browser-pending; if the intended headless path is “choose code before starting browser login,” the UI should make that explicit in the connectable state only. Add coverage for the browser-pending fallback click so this does not regress.

✅ 2. Prior R5 auth-link part is fixed: `onAuthUrl` now updates `pendingAuth.authUrl`, and the web renders a “Didn't open?” sign-in link.

✅ 3. Prior R4 unsupported-provider issue is fixed: `runtimeAuthProviderSchema` narrows start commands to `claude-code | codex`, and server coverage rejects `claude-code-tui` without sending a frame.

**Executed**
- Submitted follow-up request_changes.
- Did not run local tests/checks per PR Review mode; reviewed diff and author-stated suite results only.

yuezengwu:
Good catch — fixed in the latest push.

The device-code fallback action no longer renders during `browser-pending`, where it was a dead action (the daemon drops a same-provider duplicate start while the browser login is in flight). The headless choice is now made **upfront only**: a `offersDeviceCodeFallback(provider, view)` predicate returns true only for codex in the `connectable` state, and the card renders the action through it — so it appears before login (pick code vs browser based on whether the host has a browser) and never mid-flight. The in-flight recovery for a browser that didn't auto-open stays the "Didn't open? Open the sign-in page" auth-URL link.

A view-model test covers the predicate: live for codex+connectable, **false** for codex+browser-pending, off for claude. web suite green (902); typecheck + biome clean.

baixiaohang (approved):
## Review by codex-assistant - PR #1194

**Recommendation**: approve
- Reason: Follow-up fixes close the browser fallback/device-code and unsupported-provider contract issues; no remaining blocking issue found.

**Level**: A
- Semantic lift: runtime-auth/capability auth pipeline is core runtime behavior; already A

**PR Summary**
- Problem solved: in-product browser OAuth Connect for Claude Code and Codex credentials from the Computers card, without requiring a separate manual CLI login flow.
- Design: server forwards `runtime-auth:start`, daemon runs provider official login on the host, progress rides `pendingAuth` in capability snapshots, and web polls capabilities as the source of truth.
- Affected modules: `apps/cli`, `packages/client`, `packages/shared`, `packages/server`, `packages/web`.

**Review Comments**
✅ 1. Prior blocker fixed: `05f2e9c` offers Codex device-code fallback only in the upfront `connectable` state, not during `browser-pending`, so the UI no longer renders a duplicate-start action the daemon would ignore.
✅ 2. Prior fallback gap fixed: `onAuthUrl` is carried into `pendingAuth.authUrl`, and web renders the “Didn't open?” sign-in link for browser no-auto-open recovery.
✅ 3. Prior unsupported-provider issue fixed: runtime-auth start is narrowed to `claude-code | codex`; `claude-code-tui` is rejected with 400 and no daemon frame.

**Executed**
- Did not run local tests/checks per PR Review mode; reviewed diff and author-stated suite results only.

PR #1204, opened 2026-06-23
fix: notify resumed Claude sessions when the briefing changed

yuezengwu:
## Problem

After #1190 retired the agent-final-text mirror, conversations that **resumed an existing transcript** kept behaving under the *previous* operating contract — most visibly, they no longer delivered a human-visible reply because they never learned the mirror was gone and `chat send <human>` is now the way to reply.

Root cause is subtle and affects **both providers**:
- The briefing (`AGENTS.md` / `CLAUDE.md`) **is** rewritten on every resume, and the Claude SDK reloads it via `settingSources: ["project"]`. But a resumed session's behavior is anchored to the long transcript it was built under; a freshly-read-but-changed briefing is not salient enough to override the established behavior.
- A resumed **Codex** thread is worse: the CLI reads `AGENTS.md` once at thread init and never re-reads it, with no detection at all.

## Fix

Detect the change explicitly and surface it on the next turn, where the model will actually read it. Shared module `session-briefing-fingerprint.ts`: sha256 the rendered briefing, keyed per session at `.first-tree-workspace/session-briefings/<id>.json`; when the briefing the next turn would run under differs from the last delivered turn's baseline (or there is no baseline — a session predating the mechanism), prepend a one-time `<system-reminder>` re-read notice.

**Claude SDK handler** — start / resume / fresh-fallback / inject-drain all route through one `deliverUserMessage()` chokepoint that does the compare + prepend, then advances the baseline **only after the input is in the replay buffer**. This:
- covers the `maybeSwitchConfig()` resume-mode restart: a mid-session config change rewrites the briefing, and the next message (drained through `pushInjectedMessage`) now carries the notice **(review pt 1)**;
- leaves the notice pending for the retry if `buildQuery()` throws before the input is buffered, instead of recording it as shown **(review pt 2)**.

**Codex (both transports)** — reuse the fingerprint module; prepend the notice ahead of the Current Chat Context in `pendingChatContextPrompt`. The baseline is advanced **only when the turn actually reached the provider and settled complete** — `runTurn()` / `runTurnFromText()` return that signal — so a pre-provider / retried / aborted turn leaves the notice pending for redelivery **(follow-up review pt 1)**. Covered turn boundaries:
- **resume** — compare against the persisted baseline;
- **active-session prompt hot-switch** — a Codex thread reads `AGENTS.md` once at init and (unlike Claude) had no inject-path hot-switch, so a mid-session prompt change only landed after a suspend/resume. Before an injected turn both transports rebuild the briefing from the latest cached config (synchronous `.get()`, kept fresh by config-update pushes), rewrite `AGENTS.md` on change, and fire the notice **(QA finding)**. The refresh is **non-throwing**: it runs after the batch is dequeued, so a rewrite failure skips the hot-switch for that turn (message still delivers under the prior briefing, retried next turn) rather than stranding the message **(follow-up review pt 2)**.

Model / MCP hot-switch for Codex (which would need a thread restart) stays out of scope; this targets the prompt/briefing.

## Design notes

- **Per-session files, not one shared map** — the agent home is shared by every chat of the agent, so each session writes only its own file and sibling chats never race.
- **Whole-briefing fingerprint** — the briefing is stable/cache-friendly across resumes for the same config + runtime (Current Chat Context is not in it), so the hash changes exactly when something the agent reads changes: a prompt-stack edit, identity / source-repo / tree-binding change, a skill-set change, or a runtime template change between CLI versions (the upgrade case this targets).
- **One migration nudge** — a missing baseline counts as changed, so each pre-existing conversation gets exactly one re-read nudge on its first post-deploy turn, then compares normally.

## Scope / follow-ups

- Covered: Claude Code SDK handler (all delivery paths incl. config hot-switch) + Codex SDK + Codex app-server (resume **and** active-session prompt hot-switch).
- Not covered: the Claude **legacy per-chat resume path** (deliberately avoids writing `.first-tree-workspace/`), the **Claude-TUI** handler (append-system-prompt file, a different channel), and Codex **model / MCP** hot-switch (needs a thread restart) — follow-ups.

## Tests

`pnpm --filter @first-tree/client test` — 104 files / **1101** pass. New: `session-briefing-fingerprint.test.ts` (9) and five codex app-server cases (resume missing-baseline → notice; unchanged briefing → none; pre-provider turn/start failure does not advance the baseline; mid active-session prompt change → injected turn carries the notice; active-session rewrite failure still delivers the message). `biome check` + `typecheck` clean. Claude resume-wiring is covered by typecheck + the existing handler suites (a full Claude integration test would need faked `~/.claude/projects/<cwd>/<id>.jsonl` transcripts, which pollute the real home dir).

## Context Tree

Paired tree PR updating `system/cloud/agents/agent-behavior-steering.md` (briefing channel lifecycle, provider-agnostic + hot-switch): agent-team-foundation/first-tree-context#582.

### DIFF
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -861,3 +861,51 @@ describe("codex app-server handler", () => {
     await handler.shutdown();
   });
 });
+
+describe("codex app-server briefing-update notice", () => {
+  it("prepends a re-read notice on resume when the session has no recorded briefing baseline", async () => {
+    const fake = new FakeAppServerClient();
+    const handler = makeHandler(fake);
+    const ctx = makeContext({ finishTurn: async () => {} });
+
+    const resumePromise = handler.resume(makeMessage("m1", "hello"), "thread-app-server", ctx);
+    await waitFor(() => fake.requests.some((request) => request.method === "turn/start"));
+
+    const start = fake.requests.find((request) => request.method === "turn/start");
+    const input = JSON.stringify(start?.params ?? {});
+    expect(input).toContain("<system-reminder>");
+    expect(input).toContain("re-read");
+
+    completeTurn(fake, "turn-1", "ok");
+    await resumePromise;
+    await handler.shutdown();
+  });
+
+  it("adds no notice when the briefing is unchanged since the session last ran a turn", async () => {
+    // First session start seeds the briefing baseline for this thread id at the
+    // shared agent home (keyed off `workspaceRoot`).
+    const startFake = new FakeAppServerClient();
+    const startHandler = makeHandler(startFake);
+    const startCtx = makeContext({ finishTurn: async () => {} });
+    const startPromise = startHandler.start(makeMessage("m1", "first"), startCtx);
+    await waitFor(() => startFake.requests.some((request) => request.method === "turn/start"));
+    completeTurn(startFake, "turn-1", "first answer");
+    await startPromise;
+    await startHandler.shutdown();
+
+    // A later resume of the same thread, same config (briefing unchanged) →
+    // baseline matches → no re-read notice.
+    const resumeFake = new FakeAppServerClient();
+    const resumeHandler = makeHandler(resumeFake);
+    const resumeCtx = makeContext({ finishTurn: async () => {} });
+    const resumePromise = resumeHandler.resume(makeMessage("m2", "again"), "thread-app-server", resumeCtx);
+    await waitFor(() => resumeFake.requests.some((request) => request.method === "turn/start"));
+
+    const start = resumeFake.requests.find((request) => request.method === "turn/start");
+    expect(JSON.stringify(start?.params ?? {})).not.toContain("<system-reminder>");
+
+    completeTurn(resumeFake, "turn-1", "second answer");
+    await resumePromise;
+    await resumeHandler.shutdown();
+  });
+});
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -908,4 +908,31 @@ describe("codex app-server briefing-update notice", () => {
     await resumePromise;
     await resumeHandler.shutdown();
   });
+
+  it("keeps the notice for the next resume when the turn fails before reaching the provider", async () => {
+    // First resume hits a pre-provider turn/start failure: the notice was
+    // consumed for that attempt but the model never saw it, so the baseline
+    // must NOT advance.
+    const failFake = new FakeAppServerClient();
+    failFake.turnStartError = new CodexAppServerTransportError("codex app-server request timed out: turn/start");
+    const failHandler = makeHandler(failFake);
+    const failCtx = makeContext({ failSessionForRecovery: vi.fn(), finishTurn: async () => {} });
+    await failHandler.resume(makeMessage("m1", "hello"), "thread-app-server", failCtx);
+    expect(failFake.requests.some((request) => request.method === "turn/start")).toBe(true);
+    await failHandler.shutdown();
+
+    // Second resume of the same thread: baseline was never recorded, so the
+    // briefing still reads as changed and the re-read notice reappears.
+    const okFake = new FakeAppServerClient();
+    const okHandler = makeHandler(okFake);
+    const okCtx = makeContext({ finishTurn: async () => {} });
+    const resumePromise = okHandler.resume(makeMessage("m2", "again"), "thread-app-server", okCtx);
+    await waitFor(() => okFake.requests.some((request) => request.method === "turn/start"));
+    const start = okFake.requests.find((request) => request.method === "turn/start");
+    expect(JSON.stringify(start?.params ?? {})).toContain("<system-reminder>");
+
+    completeTurn(okFake, "turn-1", "ok");
+    await resumePromise;
+    await okHandler.shutdown();
+  });
 });
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -5,6 +5,7 @@ import type { SessionEvent } from "@first-tree/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CodexAppServerRpcError, CodexAppServerTransportError } from "../handlers/codex/app-server/client.js";
 import { createCodexAppServerHandler } from "../handlers/codex/app-server/index.js";
+import { writeAgentBriefing } from "../runtime/bootstrap.js";
 import type { DeliveryToken, SessionContext, SessionMessage } from "../runtime/handler.js";
 import { mockCtxPlumbing } from "./test-helpers.js";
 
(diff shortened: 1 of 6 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1204

- Reason: the new notice/baseline mechanism is wired only into `handler.resume()`, but another Claude SDK resume-mode path can run a message under a changed briefing without the notice.

**Level**: A
- Semantic lift: none; already A by path, with Claude SDK session/briefing behavior in scope.

**PR Summary**
- Problem: resumed Claude SDK transcripts can keep following an older operating contract after the shared briefing changes.
- Design: fingerprint the rendered briefing per SDK session and prepend a one-time re-read notice on resumed message turns when the fingerprint changed or is missing.
- Affected modules: `packages/client/src/handlers/claude-code.ts`, `packages/client/src/runtime/session-briefing-fingerprint.ts`.

**Review Notes**
❌ 1. `maybeSwitchConfig()` still restarts the Claude SDK query in resume mode after config changes without running the new fingerprint/notice path. When a prompt/resource config change arrives while the session is active, `pushInjectedMessage()` calls `maybeSwitchConfig()` before converting and pushing the next user message; the restart rewrites `CLAUDE.md` and calls `buildQuery(sid, ..., sid)`, then the message is delivered with no re-read notice and no baseline update. That is the same stale-transcript failure mode this PR is trying to close, and it also contradicts the PR/tree wording that the runtime tracks the briefing each turn ran under. Please extract the compare/prepend/write behavior so both `handler.resume()` and the config-hot-switch resume path share it, or explicitly narrow the contract so active-session config changes are not claimed covered. [R5 / `packages/client/src/handlers/claude-code.ts:1295`]

⚠️ 2. The resume path advances the fingerprint baseline before `spawnQuery()` succeeds and before the notice-bearing input is queued. If `buildQuery()` throws synchronously, the agent never receives the notice, but the next retry sees the new fingerprint as already recorded and suppresses it. This path already treats build success as a meaningful boundary elsewhere (`recordAppliedPayload` happens after `buildQuery`), so the baseline write should move after the SDK query/input setup succeeds, or otherwise be tied to the input actually entering the replay buffer. [R4 / `packages/client/src/handlers/claude-code.ts:1994`]

**Executed**
- Submitted request_changes.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1204 follow-up

- Reason: the Claude SDK review points are fixed, but the new Codex baseline advancement can still record the notice as delivered after helper-level retry paths where the turn did not actually run.

**Level**: A
- Semantic lift: none; already A by path, now also touching Codex provider resume behavior.

**PR Summary**
- Problem: resumed provider transcripts can keep following an older operating contract after the shared briefing changes.
- Design: persist per-session briefing fingerprints and prepend a one-time re-read notice on changed/missing baseline.
- Follow-up delta: Claude SDK now uses a shared delivery chokepoint; Codex SDK and app-server now add resume notices.

**Review Notes**
❌ 1. Codex advances the briefing baseline after `runTurn()` / `runTurnFromText()` returns, but both helpers can return after retrying the message before a usable turn was delivered. In app-server, `runTurnFromText()` returns after `token.retry(...)` for missing thread, shutdown-before-start, `turn/start` throw, or missing turn id; the caller then still writes `writeSessionBriefingFingerprint(...)` at `index.ts:1203`. In SDK, `runTurn()` resolves after settlement even when `settlement.action.kind === "retry"`, and the resume caller still writes the baseline at `sdk.ts:1446`. That means a pre-provider / unknown-custody failure can consume `pendingChatContextPrompt` with the re-read notice and then mark the fingerprint as shown, so the eventual redelivery suppresses the very notice this PR is trying to preserve. Please make these helpers return whether the notice-bearing input was actually accepted/delivered enough to advance the baseline, or move the baseline write into the successful/complete settlement branch. [R4 / `packages/client/src/handlers/codex/app-server/index.ts:1201`, `packages/client/src/handlers/codex/sdk.ts:1438`]

✅ 2. The previous Claude SDK findings are addressed: `maybeSwitchConfig()` refreshes `currentBriefingFingerprint`, injected messages route through `deliverUserMessage()`, and the baseline write moved after replay-buffer insertion.

**Executed**
- Submitted request_changes follow-up.
- Did not run tests per review-mode rules.

baixiaohang (approved):
## Review by codex-assistant - PR #1204 final

**Recommendation**: approve
- Reason: the latest head fixes both previously blocking delivery/baseline issues and preserves the one-time briefing re-read notice across retry boundaries.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: resumed Claude/Codex sessions can keep following an older briefing contract after the stable briefing changes.
- Design: persist per-session briefing fingerprints and prepend a one-time re-read notice on resumed turns when the fingerprint changed or is missing.
- Affected modules: Claude SDK handler, Codex SDK/app-server handlers, shared session briefing fingerprint runtime module.

**Review Notes**
✅ 1. Previous Claude SDK hot-switch and baseline-ordering findings are fixed: message delivery now routes through the shared notice chokepoint and records the baseline after the notice-bearing input is buffered.
✅ 2. Previous Codex retry/baseline finding is fixed: `runTurn()` / `runTurnFromText()` now return delivered/completed signals, so pre-provider or retried turns leave the notice pending for redelivery.
✅ 3. CI is green on the reviewed head.
⚠️ 4. Legacy Claude per-chat resume and Claude-TUI remain explicit follow-ups, which is consistent with the PR's stated scope.

**Executed**
- Submitted approve after baixiaohang authorization.
- Did not run tests per review-mode rules.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1204 re-review after latest synchronize

- Reason: the new Codex app-server active hot-switch can drop an injected message batch if the briefing rewrite fails before the turn is started.

**Level**: A
- Semantic lift: none; already A by path, with provider session delivery behavior in scope.

**PR Summary**
- Problem: resumed or active Claude/Codex sessions can keep following an older briefing contract after the stable briefing changes.
- Design: persist per-session briefing fingerprints and prepend a one-time re-read notice when the next delivered turn should run under a changed briefing.
- Latest delta: Codex SDK/app-server now attempt active-session prompt hot-switch before injected turns, not only resume.

**Review Notes**
❌ 1. Codex app-server removes the pending injected batch before the new active-session briefing rewrite, but that rewrite can throw synchronously. `startTurnFromPendingInputs()` formats the batch, calls `removePendingPrefix(batch)`, then calls `refreshBriefingForActiveTurn()`, which reaches `writeAgentBriefing(cwd, briefing)`. `writeAgentBriefing` uses synchronous filesystem writes and can throw; `drainPendingInputs()` is launched as `void drainPendingInputs()` with no catch, so this path leaves the messages removed from `pendingInputs`, never sent to `turn/start`, and never retried via the delivery token. This violates the same delivery/baseline boundary the PR is trying to protect: a pre-provider hot-switch failure should leave the input recoverable, not silently strand it. Please either perform the refresh before removing the batch, or wrap the refresh in the same retry/close guard as other pre-provider failures. [R4 / `packages/client/src/handlers/codex/app-server/index.ts:1091`]

✅ 2. The previously blocking Claude SDK and Codex resume retry/baseline issues remain fixed on this head.
✅ 3. CI is green on the reviewed head (`6039c149`).

**Executed**
- Submitted request_changes after the latest synchronize.
- Did not run tests per review-mode rules.

baixiaohang (approved):
## Review by codex-assistant - PR #1204 final after latest synchronize

**Recommendation**: approve
- Reason: the latest head fixes the Codex app-server hot-switch failure boundary, previous Claude/Codex resume findings remain fixed, and CI is green.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: resumed or active Claude/Codex sessions can keep following an older briefing contract after the stable briefing changes.
- Design: persist per-session briefing fingerprints and prepend a one-time re-read notice when the next delivered turn should run under a changed briefing.
- Latest delta: Codex active-session prompt hot-switch now refreshes the briefing without letting rewrite failures strand already-dequeued injected input.

**Review Notes**
✅ 1. The previous Codex app-server batch-stranding finding is fixed: active-session briefing refresh is non-throwing, so a rewrite failure logs and delivers the injected message under the prior briefing instead of dropping it before `turn/start`.
✅ 2. The previous Claude SDK hot-switch/baseline findings and Codex resume retry/baseline finding remain fixed.
✅ 3. CI is green on the reviewed head (`39e8e5aa`).
⚠️ 4. Legacy Claude per-chat resume, Claude-TUI, and Codex model/MCP hot-switch remain explicit follow-ups, consistent with stated scope.

**Executed**
- Submitted approve after baixiaohang authorization.
- Did not run tests per review-mode rules.

PR #1211, opened 2026-06-23
feat(web): add-repo form — URL + branch only, derive name, normalize URL

Gandy2025:
## What & why

Item 7 of the agent-detail follow-up batch. Simplifies the **Settings → Resources** repo editor: a repo's name is just its `owner/repo`, so asking for it is redundant, and the URL field should accept the shapes people actually paste.

## Changes

- **Drop the Name field.** The repo's display name is derived from the URL (`deriveRepoShortLabel` → `owner/repo`). The form is now **URL (required) + Default branch (optional**, repo default when blank).
- **Normalize the URL** before save + validation so common paste shapes work without a scheme (`github.com/org/repo`, `org/repo` → `https://github.com/org/repo`; full `https://` / `ssh://` / scp-like pass through). New `lib/normalize-repo-url.ts` (+ unit test) — only widens accepted input; the server schema still validates.

## Verification

- `pnpm --filter @first-tree/web test` green; `tsc` + `lint:tokens` + `biome` clean.

## Notes

- Frontend-only; no backend / API / schema change. No Context Tree write.
- **Follow-up:** the agent-detail "Add agent repository" dialog (`capability-section.tsx`) is aligned the same way in **#1220** — it was NOT part of this merged PR (the merged diff here is fixed at `0ce22b42`: the normalize helper + Settings repo editor only).

PR #1209, opened 2026-06-23
feat(web): agent-detail polish — instruction segments, label-side help, execution cleanup

Gandy2025:
## What & why

Follow-up polish on the (now merged) agent-detail UX redesign. gandy reviewed the shipped result and raised a batch of refinements; this PR lands **items 2–6** of that list (item 1 — profile-dialog instant-save — and item 7 — add-repo form in Settings → Resources — follow as separate PRs).

## Changes

- **Instructions — "All instructions" + segmented merged view (items 5 + 6).** The merged-instructions panel eyebrow is renamed `Effective` → **All instructions**, and the panel now renders each contributed instruction as its **own labelled segment** split by a dashed rule, instead of one undifferentiated blob. Segments come from the server-resolved `prompt.sections` — the same effective stack, in the same order, that produces `prompt.append` — so the on-screen segmentation matches exactly what the agent runs. Each segment is labelled by its source resource name, or "Team instructions" / "Custom instructions" for the agent's own unnamed inline fragment. Falls back to the merged `append` string for older payloads with no `sections`.
- **Help "?" moves to the label (item 2).** `ConfigRow`'s help-icon tooltip now sits next to the label rather than the value — it explains the setting (which the label names), not whichever value is selected. Model / Reasoning effort benefit.
- **Execution — drop redundant computer status (item 3).** The bound Computer row no longer repeats an Online/Offline indicator; the page header already shows presence, and the two read as redundant side by side. The computer name stays. Removes the now-dead `boundComputerOnline` chain (context field + derivation + prop).
- **Execution before Model settings (item 4)** in the Runtime tab.

## Verification

- `pnpm --filter @first-tree/web test` — 121 files / 922 tests green (added segmented-block assertions to the page DOM test).
- `tsc --noEmit` + `lint:tokens` clean; `biome check` clean.
- DEV `/preview/agent-detail` updated: the Instructions block shows the segmented view; Execution shows the name-only computer row.

## Notes

- No backend / API / schema change — presentation-only (`prompt.sections` already existed, server-resolved). No Context Tree write.
- Do not self-merge — gandy-s-assistant coordinates; gandy2025 authorizes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1212, opened 2026-06-23
feat(web): profile dialog saves every field instantly (no Save button)

Gandy2025:
## What & why

Item 1 of the agent-detail follow-up batch. The Edit-profile dialog was the one place on the page with a split save model — the avatar saved instantly but name / visibility / color / delegate waited for a Save button — plus a caption explaining that split. This makes the whole dialog instant-save, consistent with every other control on the page, so the caption is no longer needed.

## Changes

- **Every field saves on its own commit:** display name on **blur / Enter**, visibility / color / delegate on **change**. Each is a partial `PATCH /agents/:uuid`.
- **Removed the Save button** — the footer is now a single **Done** that just closes (a subtle "Saving…" indicator shows while a field is in flight).
- **Removed the caption** ("Avatar changes save immediately. Other details save when you click Save.") — the split it described is gone.
- **Display name stays required:** empty on blur shows an inline error and does not save. A failed save surfaces an inline error without a "saved" flash.
- Avatar image / pixel uploads remain eager (unchanged).

## Verification

- `pnpm --filter @first-tree/web test` — 922 tests green (the profile-dialog DOM test reworked: color / visibility / delegate assert per-field instant PATCHes; name commits on blur; empty name blocks the save).
- `tsc --noEmit` + `lint:tokens` clean; `biome check` clean.

## Notes

- Frontend-only — no backend / API / schema change (the PATCH endpoint already accepts partial agent updates). No Context Tree write.
- Final PR of the batch: #1209 (items 2–6), #1211 (item 7), this (item 1).
- Do not self-merge — gandy-s-assistant coordinates; gandy2025 authorizes.

### DIFF
--- a/packages/web/src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx
+++ b/packages/web/src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx
@@ -374,6 +374,60 @@ describe("ProfileEditDialog (merged identity + appearance)", () => {
     await act(async () => root.unmount());
   });
 
+  it("commits a pending name edit when Done is clicked without an explicit blur", async () => {
+    const { ProfileEditDialog } = await import("../profile-edit-dialog.js");
+    const onSave = vi.fn().mockResolvedValue(undefined);
+    const onOpenChange = vi.fn();
+    const { root } = await renderDom(
+      <ProfileEditDialog agent={agent()} open onOpenChange={onOpenChange} onSave={onSave} onSaved={vi.fn()} />,
+    );
+    const displayInput = document.body.querySelector<HTMLInputElement>("#profile-display");
+    if (!displayInput) throw new Error("Expected display field");
+    // Type a new name but do NOT blur — click Done straight away.
+    await setInputValue(displayInput, "Nova Done");
+    await click(buttonByText(document.body, "Done"));
+    // Done commits the pending name before closing, so the edit isn't lost.
+    expect(onSave).toHaveBeenCalledWith({ displayName: "Nova Done" });
+    expect(onOpenChange).toHaveBeenCalledWith(false);
+    await act(async () => root.unmount());
+  });
+
+  it("serializes field saves — controls and Done disable while a save is in flight", async () => {
+    const { ProfileEditDialog } = await import("../profile-edit-dialog.js");
+    let resolveSave: (() => void) | undefined;
+    const onSave = vi.fn().mockImplementation(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveSave = resolve;
+        }),
+    );
+    const onOpenChange = vi.fn();
+    const { root } = await renderDom(
+      <ProfileEditDialog
+        agent={agent({ avatarColorToken: "hue-1" })}
+        open
+        onOpenChange={onOpenChange}
+        onSave={onSave}
+        onSaved={vi.fn()}
+      />,
+    );
+    // Pick a color — the save starts and stays pending (unresolved).
+    await click(document.body.querySelector('button[title="hue-3"]'));
+    expect(onSave).toHaveBeenCalledTimes(1);
+    // While pending: other swatches + Done are disabled, and a second click is
+    // dropped by the in-flight guard — no overlapping PATCH, no close mid-save.
+    expect(document.body.querySelector<HTMLButtonElement>('button[title="hue-2"]')?.disabled).toBe(true);
+    expect(buttonByText(document.body, "Done")?.disabled).toBe(true);
+    await click(document.body.querySelector('button[title="hue-2"]'));
+    expect(onSave).toHaveBeenCalledTimes(1);
+    // Resolve the save → controls re-enable.
+    await act(async () => {
+      resolveSave?.();
+    });
+    expect(buttonByText(document.body, "Done")?.disabled).toBe(false);
+    await act(async () => root.unmount());
+  });
+
   it("offers the pixel-avatar generator only for non-human agents (humans use their GitHub avatar)", async () => {
     const { ProfileEditDialog } = await import("../profile-edit-dialog.js");
 
--- a/packages/web/src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx
+++ b/packages/web/src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx
@@ -129,6 +129,15 @@ async function setInputValue(element: HTMLInputElement, value: string): Promise<
   await flush();
 }
 
+async function blurInput(element: Element | null): Promise<void> {
+  if (!element) throw new Error("Expected element to blur");
+  await act(async () => {
+    // React's onBlur listens to the bubbling `focusout` event.
+    element.dispatchEvent(new FocusEvent("focusout", { bubbles: true }));
+  });
+  await flush();
+}
+
 async function setFileInput(element: HTMLInputElement, file: File): Promise<void> {
   Object.defineProperty(element, "files", { configurable: true, value: [file] });
   await act(async () => {
@@ -324,11 +333,12 @@ describe("ProfileEditDialog (merged identity + appearance)", () => {
     await click(buttonByText(document.body, "Remove image"));
     expect(agentApiMocks.deleteAgentAvatar).toHaveBeenCalledWith("agent-1");
 
-    // Save commits identity fields + the fallback color in one PATCH.
+    // Colour saves instantly on swatch click — there is no Save button.
     await click(document.body.querySelector('button[title="hue-3"]'));
-    await click(buttonByText(document.body, "Save"));
-    expect(onSave).toHaveBeenCalledWith(expect.objectContaining({ displayName: "Nova", avatarColorToken: "hue-3" }));
-    expect(onSaved).toHaveBeenCalledTimes(1);
+    expect(onSave).toHaveBeenCalledWith({ avatarColorToken: "hue-3" });
+    expect(onSaved).toHaveBeenCalled();
+    // "Done" just closes — every field already saved on its own commit.
+    await click(buttonByText(document.body, "Done"));
     expect(onOpenChange).toHaveBeenCalledWith(false);
 
     await act(async () => root.unmount());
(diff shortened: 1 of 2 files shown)


yuezengwu (changes requested):
Goal/core change: this PR moves the profile edit dialog from a split Save-button model to per-field immediate PATCHes: display name commits on blur/Enter, visibility/color/delegate commit on change, and Done only closes the dialog. That matches the Agent Detail immediate-save model in the Context Tree; I did not see backend/API/schema changes.

Blocker: the new `saveField` path can issue overlapping PATCHes while the controls remain enabled. A user can click two colors quickly, or change visibility/delegate/name again before the previous `onSave` resolves; those are unversioned partial `PATCH /agents/:uuid` calls, so an older request that is applied after a newer one can leave the server/cache with the stale value while the dialog shows the latest local state. The runtime-config immediate-save path avoids this by disabling edits while pending; this dialog needs an equivalent serialization/latest-wins guard before shipping.

Minor while touching this: the file-level comment in `profile-edit-dialog.tsx` still describes identity+color as one Save-button PATCH, which is now stale.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1212

- Reason: the new instant-save path can hide failed saves after close and can persist an older value when field saves overlap.

**Level**: C
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or shared data-layer code.

**PR Summary**
- Problem solved: removes the profile dialog's split save model so profile fields save at their natural commit point.
- Design: replaces the Save submit with per-field partial `PATCH /agents/:uuid`; avatar image/pixel uploads remain eager; footer `Done` only closes.
- Affected modules: `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx` and the profile dialog DOM test.

**Review Comments**
❌ 1. `Done` can close while a profile PATCH is still pending, so a rejected save is no longer surfaced. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:487`]

When a user changes color, visibility, delegate, or display name and immediately clicks `Done`, `saveField` may still be awaiting `onSave`, but the dialog closes because `Done` is only disabled for `uploading`. If that request rejects, `setSaveError` runs after the dialog is gone, so the PR's "failed save surfaces an inline error" guarantee is broken. The previous footer disabled both actions while `saving`. Please keep the dialog open until pending profile saves settle, disable `Done` while `saving`, or otherwise surface post-close failures.

❌ 2. Instant field saves can race and leave the server with an older selection than the UI. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:190`]

`saveField` fires every patch immediately, but the controls remain enabled and there is no serialization, pending counter, or latest-wins guard. A quick color change such as `hue-2 -> hue-3`, or changing a select twice, sends overlapping PATCHes; if the earlier request is applied after the later request, the stored agent value becomes the stale choice while the local dialog preview still shows the latest choice. This page's existing immediate config save path avoids this class of race by disabling controls while pending. Please serialize/coalesce profile saves or disable the affected controls while a profile save is pending.

**Action Taken**
- Submitted a `request_changes` review.
- Tests were not run, per PR Review mode.

yuezengwu (changes requested):
Goal/core change: the updated PR keeps the profile dialog on the intended immediate-save model: display name commits at its natural commit points, visibility/color/delegate commit on change, and the dialog now disables controls + Done while a save is pending. The prior stale file-level comment was also refreshed. I still do not see backend/API/schema changes.

Blocker: the new display-name dedupe marks a name as saved before the PATCH actually succeeds. In `commitName`, `savedNameRef.current = trimmed` runs before `saveField({ displayName: trimmed })` returns. If that PATCH rejects, the inline error is shown, but the unsaved name is now treated as already committed: blurring again or clicking Done takes the `trimmed === savedNameRef.current` path, returns true, and Done can close without retrying or saving the value. That breaks the failure guarantee from this PR. Please only advance the dedupe baseline after a successful save, or roll it back on failure, and add a DOM test for rejected name save -> Done/blur should not close or skip the retry.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1212 follow-up

- Reason: the follow-up fixes the normal pending/disable path, but a failed display-name save is still marked as saved before it succeeds, so the same unsaved name can no longer be retried and `Done` can close without persisting it.

**Level**: C
- Semantic lift: none; this remains scoped to the agent-detail UI and DOM tests.

**PR Summary**
- New commit reviewed: `39619cc6` (`fix(web): serialize profile-dialog saves; Done awaits + closes`)
- Fix intent: disable controls/`Done` while saving, add a synchronous in-flight guard, make `Done` commit a typed-but-unblurred display name, and refresh the stale file comment.
- Affected modules: `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx` and the profile dialog DOM test.

**Review Comments**
❌ 1. `commitName` records the new display name before the PATCH succeeds, so a failed save is treated as already saved. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:239`]

`commitName` sets `savedNameRef.current = trimmed` before calling `saveField`. If that `PATCH` rejects, the ref still equals the unsaved input. The next blur or `Done` click then takes the `trimmed === savedNameRef.current` branch, returns `true` without calling `onSave`, and `handleDone` closes the dialog even though the display name was never persisted. It also prevents retrying the same display name unless the user edits it to a different string first. Please only advance `savedNameRef` after `saveField` returns success, or roll it back on failure, and add coverage for `Done`/retry after a rejected name save.

⚠️ 2. If `savingRef` is meant to handle pre-render duplicate events, local field state should not move when the save is dropped. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:358`]

The select/swatch handlers update local state before calling `saveField`; if the synchronous guard returns `false`, the UI has already moved to a value that was not sent. The disabled controls likely cover normal user clicks after React re-renders, but the comment explicitly relies on the ref for the pre-render window. If that guard is part of the correctness story, make the local state update contingent on an accepted save or revert it when `saveField` returns `false`.

**Action Taken**
- Submitted a follow-up `request_changes` review.
- Tests were not run, per PR Review mode.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1212 follow-up 2

- Reason: the new commit fixes the previous retry/dropped-state issues, but the real focused-input `Done` click path can save without closing because `onBlur` races ahead of `onClick`.

**Level**: C
- Semantic lift: none; still scoped to the agent-detail UI and DOM tests.

**PR Summary**
- New commit reviewed: `97a00ddf` (`fix(web): advance name dedupe only on save success; drop guarded interactions cleanly`)
- Fix intent: advance the display-name dedupe baseline only after successful save, retry failed name saves, and avoid moving local select/swatch state when `savingRef` drops an interaction.
- Affected modules: `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx` and the profile dialog DOM test.

**Review Comments**
❌ 1. Clicking `Done` from a focused display-name input can save the name but leave the dialog open. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:345`]

In the real browser event order, clicking the `Done` button while the display-name input is focused first blurs the input, so `onBlur={() => void commitName()}` starts the PATCH and sets `savingRef.current = true` / `saving = true`. The `Done` button is then disabled before or during the click path; if `onClick` still fires, `handleDone -> commitName -> saveField` sees `savingRef.current` and returns `false`, so it does not close. After the blur-triggered save succeeds, nothing calls `onOpenChange(false)`. The new DOM test dispatches only a button click after `setInputValue`, so it misses this focused-input blur-before-click sequence. Please make the `Done` path own this commit+close interaction, or suppress/dedupe the blur-triggered commit when the blur is caused by `Done`, and add coverage that focuses the input before clicking `Done`.

**Action Taken**
- Submitted a follow-up `request_changes` review.
- Tests were not run, per PR Review mode.

yuezengwu (changes requested):
Goal/core change: this round fixes the two previous correctness issues: the display-name dedupe baseline now advances only after a successful PATCH, and guarded visibility/delegate/color interactions return before moving local state. The PR remains frontend-only; I still do not see backend/API/schema changes.

Blocker: the normal browser path for "type a display name, then click Done" can lose the Done click. The input still commits on `onBlur`, and Done is disabled whenever `saving` flips true. In a real pointer interaction, clicking the button while the input is focused blurs the input before the button's `click`; that blur starts `commitName()`, sets `saving`, and re-renders Done disabled. The subsequent button click may not fire, so the name is saved but the dialog stays open and the user has to click Done a second time. The current DOM test for "Done commits a pending name edit" dispatches a click without focusing/blurring the input first, so it misses this event ordering. Please handle the Done pointer path before blur disables the button (for example by preventing the blur on Done pointerdown/mousedown, or otherwise making the blur-started save also close when the user's target was Done), and add coverage for focused input -> click Done.

yuezengwu (approved):
Goal/core change: this PR moves the profile dialog fully onto the Agent Detail immediate-save model: display name commits at its natural commit points, visibility/color/delegate save on change, avatar image remains eager, and Done is close-only after any pending name commit is handled. I did not see backend/API/schema changes.

Reviewed latest head `b65a352c`. The previous blockers are addressed: profile saves are serialized, failed display-name saves remain retryable, guarded interactions no longer move local state when a save is dropped, and the focused-input -> Done path is now owned by Done via `onMouseDown.preventDefault()` plus `pendingNameRef` reuse. The added regression coverage matches the blur-before-click race we discussed.

Approved. Local tests were not run in the isolated review worktree because it has no `node_modules`; I reviewed code paths and the PR/CI-reported test coverage.

baixiaohang (approved):
## Review by codex-assistant - PR #1212 follow-up 3

**Recommendation**: approve
- Reason: the latest commit resolves the previously blocking focused-input `Done` path while preserving the serialized, retryable instant-save behavior.

**Level**: C
- Semantic lift: none; this remains scoped to the agent-detail UI and DOM tests.

**PR Summary**
- New commit reviewed: `b65a352c` (`fix(web): Done owns the focused-input commit+close (no blur race, no double save)`)
- Fix intent: make `Done` own the focused display-name commit path via `onMouseDown.preventDefault()`, and have `Done` await any blur-started name commit through `pendingNameRef` so it saves once and closes only after persistence.
- Affected modules: `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx` and the profile dialog DOM test.

**Review Comments**
✅ 1. Previous blockers are addressed. [R4 / `packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:245`]

The display-name dedupe baseline now advances only after a successful save, guarded select/swatch interactions no longer move local state when a save is dropped, and the focused-input `Done` path now has both the normal `mousedown.preventDefault()` path and a `pendingNameRef` fallback for blur-first sequencing. The added DOM coverage exercises the race shape by dispatching `mousedown`, `focusout`, and `click` without an intervening flush and asserts a single save plus close.

**Action Taken**
- Submitted an `approve` review.
- Tests were not run, per PR Review mode.

PR #1215, opened 2026-06-23
fix: run daemon capability probe asynchronously

baixiaohang:
## Summary
- move daemon startup capability full probes out of the pre-connection path
- start the post-registration capability refresher immediately in the background when no startup snapshot exists
- preserve runtime-auth capability entries published during slow probes, including both pending device-code state and the final completed-login provider state

## Tests
- pnpm --filter first-tree-dev exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/capability-refresh.test.ts --maxWorkers=2 --minWorkers=1
- pnpm --filter first-tree-dev typecheck
- pnpm exec biome check apps/cli/src/commands/daemon/start.ts apps/cli/src/core/capability-refresh.ts apps/cli/src/__tests__/daemon-start-command.test.ts apps/cli/src/__tests__/capability-refresh.test.ts
- git diff --check

PR #1216, opened 2026-06-23
fix: retry daemon capability uploads

baixiaohang:
## Summary
- route daemon capability PATCH uploads through `FirstTreeHubSDK.updateCapabilities()` so the background upload path uses the SDK timeout and transient retry layer
- keep the same upload-time access token, server URL, client id encoding, payload shape, and CLI user-agent
- add regression coverage for two transient `fetch failed` errors followed by a successful capability upload

## Context
After #1215, daemon startup no longer waits for the full capability probe. A staging journal showed startup succeeded quickly, but the post-registration background capability upload logged `fetch failed` during the same window as a WebSocket connect timeout; a later reconnect revalidate uploaded successfully. The direct `cliFetch` upload path did not use the SDK retry behavior already used by other HTTP calls.

## Tests
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/runtime-provider-reconcile.test.ts --maxWorkers=2 --minWorkers=1`
- `pnpm exec biome check apps/cli/src/core/runtime-provider-reconcile.ts apps/cli/src/__tests__/runtime-provider-reconcile.test.ts`
- `git diff --check`

Reviewed by codex-reviewer in chat; no blockers found.

### DIFF
--- a/apps/cli/src/__tests__/runtime-provider-reconcile.test.ts
+++ b/apps/cli/src/__tests__/runtime-provider-reconcile.test.ts
@@ -188,6 +188,8 @@ describe("uploadClientCapabilities", () => {
 
   afterEach(() => {
     vi.unstubAllGlobals();
+    vi.useRealTimers();
+    vi.restoreAllMocks();
   });
 
   it("PATCHes /clients/:id/capabilities with the snapshot", async () => {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the capability upload retry fix.

Implementation goal/core change: daemon capability PATCH uploads now go through `FirstTreeHubSDK.updateCapabilities()`, preserving the upload-time token, server URL, encoded client id path, payload shape, and CLI user-agent while gaining the SDK timeout/retry layer. The regression test covers two transient `fetch failed` errors followed by a successful upload.

No blockers found. No database or durable data-structure change to escalate.

Verified locally:
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/runtime-provider-reconcile.test.ts --maxWorkers=2 --minWorkers=1`
- `pnpm exec biome check apps/cli/src/core/runtime-provider-reconcile.ts apps/cli/src/__tests__/runtime-provider-reconcile.test.ts`

PR #1225, opened 2026-06-23
feat(runtime-auth): unify in-product Connect UX and surface login failures

yuezengwu:
## What & why

Unifies the in-product **Connect <provider>** experience and closes the failure-feedback gap. Started from three reported UX problems, then an adversarial review surfaced a deeper one (login failures were never visible).

### Problems fixed
1. **Connect gave no feedback and was re-clickable** — added a local "starting" latch in `RuntimeAuthControls` that locks the button and starts the fast-poll immediately, bridging the gap between the start POST and the daemon publishing `pendingAuth`.
2. **Inconsistent card layout** (all-unauth vs codex-ok/claude-unauth) — extracted a shared `RuntimeProviderRow` (state line + Connect/install action) and converted `SetupIncomplete` from a CSS grid to the same single-column rows as `Ready`. An unauthenticated→Connect block now looks identical everywhere.
3. **New-agent dialog dead-ended when no runtime was signed in** — it now offers in-place Connect for connectable providers; the dialog's capability poll flips Create to enabled once login lands.
4. **Login failures were silent (adversarial-review find)** — new `lastAuthError` on the capability entry; the daemon writes the terminal reason on every failure path; the web shows "Last sign-in didn't complete… Try again". Also a small AuthExpired body hint (E) and unified section headings (G).

### QA-driven fixes (real candidate server/web/daemon)
- Fallback sign-in link was `http://localhost:1455.` (trailing period → invalid URL). Then found the deeper issue: that loopback origin is codex's **local callback server**, whose root 404s — never a sign-in page. `extractAuthUrl` now skips loopback hosts and only surfaces the provider's **external** auth URL (or none), plus a web-side `new URL` guard.
- **CodeQL**: rewrote the auth-URL parsing to be provably linear (tokenise + anchored prefix + char-set trim), resolving 2 high-severity `js/polynomial-redos` alerts.

## Verification
- Local: typecheck 12/12 · client 1099 · web 938 · shared/server/cli suites green · biome clean.
- QA rerun (real candidate stack): in-dialog Connect, immediate pending + no double-click, real Codex auth → agent created, forced Claude failure → "Try again", desktop/mobile no overflow. Fallback href now parses with no trailing punctuation.
- CI: all checks green incl. CodeQL + Analyze ×3.

## Notes
- The provider's post-OAuth `/success` page can momentarily 404 because codex/claude's local callback server exits the instant it writes credentials — login still completes (daemon logs `✓ login complete`; the console flips to connected). This is provider-side and out of scope; a follow-up could add expectation-setting copy.


yuezengwu:
## QA finding #1 fixed — `9c9a57de`

**Root cause:** codex prints the local sign-in server inside prose ("If it didn't open, visit `http://localhost:1455`."), so the captured fallback URL kept the trailing period and `new URL("http://localhost:1455.")` rejected the port `1455.` as invalid.

**Fix:**
- `extractAuthUrl()` (pure, unit-tested) now requires a whitespace terminator (no truncated capture across stdout chunks) **and** strips trailing sentence punctuation `.,;:!?)]}>'"` — so the href is always parseable.
- Web defense-in-depth: the fallback link only renders when its value passes `new URL()` as http(s), so a malformed URL can never produce a broken href again.
- Tests: pure `extractAuthUrl` cases + an end-to-end `runCodexBrowserLogin` repro of the exact QA output (`http://localhost:1455.` → `http://localhost:1455`).

**Please re-verify (step 2 of the plan):** Connect Codex → after pending, the fallback "Open the sign-in page →" link `href` should be `http://localhost:1455` (no trailing period) and parse cleanly.

**Re: not-covered AuthExpired hint** — the body hint is gated on the `auth_expired` pill (`client.authState === "expired"`), which needs a connect-token past its refresh-TTL. If seeding that state is impractical, the hint is also exercised by the `auth-expired-card-body` unit/DOM coverage; happy to add a seam to force the state if you'd like it in the live run.

All gates still green locally: typecheck 12/12 · client 1097 · web 938 · cli 504 · shared 478 · server 1500.

baixiaohang (approved):
Approved. Static review found no blocking issues: runtime-auth progress/failure remains sourced from the capabilities snapshot, the new lastAuthError contract is optional/backward-compatible, URL fallback parsing now skips loopback callback origins and avoids the CodeQL ReDoS pattern, and the New Agent dialog still gates creation on an ok runtime capability.

PR #1227, opened 2026-06-23
[codex] Fix chat summary manual expand

Gandy2025:
## Summary

Fix the pinned task summary so a manual expand from any collapsed summary bar stays open when the message stream is already scrolled down.

## Root cause

There were two visually similar collapsed-bar states:

- default collapsed: `open` is false
- sticky-collapsed after scrolling an open summary: `open` is true, but `scrollCollapsed` hides the body

The original fix handled the default-collapsed case by preventing the next same-position scroll event from immediately re-collapsing the panel. The sticky-collapsed case still failed because the click handler toggled internal `open` from true to false, so the first click on that bar collapsed the internal state instead of expanding the visible panel.

## Changes

- Track the scroll position where a manual expand occurs.
- Require a fresh downward move past the collapse threshold before sticky-collapse folds the summary again.
- Treat a click from sticky-collapsed state as an expand action rather than toggling `open` false.
- Preserve the existing restore-at-top behavior.
- Add regression coverage for:
  - manual expand at an already-scrolled position
  - subsequent downward scrolling still collapsing
  - first-click expand from the sticky-collapsed bar

## Validation

- `pnpm check` — passed
- `pnpm --dir packages/web exec vitest run src/pages/workspace/center/__tests__/chat-summary.test.ts` — 19 tests passed
- `pnpm --filter @first-tree/web typecheck` — passed, design-token guardrails clean
- `pnpm --filter @first-tree/web test` — 124 files / 959 tests passed
- GitHub checks after `ea92b5e7` — passed

No Context Tree PR: this is an implementation-only interaction bug fix and does not change a durable product decision or ownership constraint.

### DIFF
--- a/packages/web/src/pages/workspace/center/chat-summary.tsx
+++ b/packages/web/src/pages/workspace/center/chat-summary.tsx
@@ -204,7 +204,7 @@ export function ChatSummary({
 
   const onToggle = useCallback(() => {
     setOpen((prev) => {
-      const next = !prev;
+      const next = scrollCollapsedRef.current ? true : !prev;
       const el = scrollContainerRef.current;
       manualExpandTopRef.current = next && el ? el.scrollTop : null;
       saveManualPref(chatId, next);
--- a/packages/web/src/pages/workspace/center/chat-summary.tsx
+++ b/packages/web/src/pages/workspace/center/chat-summary.tsx
@@ -238,8 +238,7 @@ export function ChatSummary({
       if (top <= SCROLL_RESTORE_PX) {
         manualExpandTopRef.current = null;
       }
-      const collapseTop =
-        manualExpandTop === null ? SCROLL_COLLAPSE_PX : manualExpandTop + SCROLL_COLLAPSE_PX;
+      const collapseTop = manualExpandTop === null ? SCROLL_COLLAPSE_PX : manualExpandTop + SCROLL_COLLAPSE_PX;
       if (top > collapseTop && openRef.current && !scrollCollapsedRef.current) {
         manualExpandTopRef.current = null;
         setScrollCollapsed(true);
--- a/packages/web/src/pages/workspace/center/chat-summary.tsx
+++ b/packages/web/src/pages/workspace/center/chat-summary.tsx
@@ -182,10 +182,15 @@ export function ChatSummary({
   // Manual toggles for the current version always win (see onToggle).
   const decidedForKey = useRef<string | null>(null);
   const entryKey = `${chatId}|${descriptionUpdatedAt ?? ""}`;
+  // When the user manually expands while already scrolled down, sticky-collapse
+  // must not immediately fold the panel on the next scroll event that reports the
+  // same scrollTop. Store the expansion point and require a fresh downward move.
+  const manualExpandTopRef = useRef<number | null>(null);
   useEffect(() => {
     if (!hasDescription || !freshnessReady) return;
     if (decidedForKey.current === entryKey) return;
     decidedForKey.current = entryKey;
+    manualExpandTopRef.current = null;
     setScrollCollapsed(false);
     setUnreadCleared(false);
     if (unread && isStaleSinceLastView(lastReadMs, Date.now())) {
@@ -200,13 +205,15 @@ export function ChatSummary({
   const onToggle = useCallback(() => {
     setOpen((prev) => {
       const next = !prev;
+      const el = scrollContainerRef.current;
+      manualExpandTopRef.current = next && el ? el.scrollTop : null;
       saveManualPref(chatId, next);
       return next;
     });
     setHighlighted(false);
     setUnreadCleared(true);
     setScrollCollapsed(false);
-  }, [chatId]);
+  }, [chatId, scrollContainerRef]);
 
   // Sticky-collapse: while open, scrolling the message stream down folds the
   // header to its one-line bar; returning to the top restores it. Transient —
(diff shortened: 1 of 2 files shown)


packages/web/src/pages/workspace/center/chat-summary.tsx
```diff
@@ -200,13 +205,15 @@ export function ChatSummary({
   const onToggle = useCallback(() => {
     setOpen((prev) => {
       const next = !prev;
```
yuezengwu: Blocking: this still misses the collapsed bar produced by sticky-collapse. In that state `open` is still `true` and `scrollCollapsed` is `true`, so the rendered button says “Expand summary”, but this click computes `next = !prev` and sets `open` to `false` while clearing `scrollCollapsed`; the summary stays collapsed instead of manually expanding and anchoring at the current `scrollTop`. Please derive the toggle from the rendered expanded state, or treat `scrollCollapsedRef.current` as an expand, and add a regression that expands, scroll-collapses, clicks the collapsed bar once, then verifies the markdown is visible and survives the same-position scroll event.

yuezengwu (changes requested):
Thanks for the targeted fix and the regression coverage for the default-collapsed-at-scroll-position path. I found one remaining blocking case on the same collapsed-bar interaction: the sticky-collapsed state still does not expand on the first click, so the fix is incomplete for the summary bar users actually see after scrolling an open summary down.

baixiaohang (approved):
## Review by codex-assistant - PR #1227

**Recommendation**: approve
- Reason: The change is a localized Workspace summary interaction fix; it preserves the existing sticky restore behavior and adds focused regression coverage for the manual-expand-at-scrolled-position case.

**Level**: C
- Path initial: only `packages/web/**` changed -> C
- Semantic lift: none; this touches local Workspace center UI state/tests, not subscriptions, WebSocket frame parsing, or data-layer contracts.

**PR Summary**
- Problem solved: prevents the pinned chat summary from flashing open and immediately re-collapsing when manually expanded while the message stream is already scrolled down.
- Design: records the scroll position at manual expand time, then requires a fresh downward movement beyond the sticky-collapse threshold before folding again; resets the guard on new summary entry state and top restore.
- Affected modules: `packages/web/src/pages/workspace/center/chat-summary.tsx`, `packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts`

**Review Notes**
✅ 1. The summary remains an inspectable Workspace center surface over `chat.description`, aligned with the Workspace/detail model rather than introducing a new state source.
✅ 2. The scroll guard is transient and local to the component; manual preference persistence remains the existing per-chat localStorage key.
✅ 3. Regression coverage exercises both the no-immediate-recollapse path and the later fresh downward-scroll collapse path.

**Executed**
- Approved this PR.
- Did not run tests, per review-mode constraints; relied on the author's reported validation.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1227

- Reason: The PR fixes the default-collapsed-at-scroll-position path, but the same collapsed-bar interaction is still broken when the bar is collapsed by sticky scroll state.

**Level**: C
- Path initial: only `packages/web/**` changed -> C
- Semantic lift: none; this touches local Workspace center UI state/tests, not subscriptions, WebSocket frame parsing, or data-layer contracts.

**PR Summary**
- Problem solved: intended to keep the pinned chat summary open after a manual expand while the stream is already scrolled down.
- Design: records the manual-expand scroll position and requires additional downward movement before sticky-collapse can fold again.
- Affected modules: `packages/web/src/pages/workspace/center/chat-summary.tsx`, `packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts`

**Review Notes**
❌ 1. `onToggle` still treats `open` as the only source of truth even though the visible expanded state is `open && !scrollCollapsed`. When an already-open summary has been sticky-collapsed by scrolling, the bar shows `Expand summary`, but clicking it runs `setOpen(prev => !prev)` with `prev === true`, saves the manual preference as collapsed, clears `scrollCollapsed`, and leaves `expanded` false. That means the first click on the sticky-collapsed bar still does not expand the summary, so the collapsed-bar fix is incomplete. [R4 / `packages/web/src/pages/workspace/center/chat-summary.tsx`:205]
✅ 2. The new manual-expand scroll guard looks appropriate for the default-collapsed path, and the added tests cover that path plus later downward scrolling.

**Executed**
- Requested changes with this follow-up review, superseding my earlier approve.
- Did not run tests, per review-mode constraints; relied on code inspection and the author's reported validation.

yuezengwu:
Rechecked the latest head `f89f8ad3`; the requested-change comment still applies. The second commit only formats the `collapseTop` expression, while `onToggle` still derives `next` from `open` rather than the rendered expanded/sticky-collapsed state.

yuezengwu (approved):
Re-reviewed latest head `ea92b5e7`. The previous sticky-collapsed blocker is fixed: clicking the sticky-collapsed bar now treats that state as an expand, preserves the manual expand anchor, and the new regression covers the first-click path. I did not find remaining issues in this focused interaction change.

baixiaohang (approved):
## Review by codex-assistant - PR #1227

**Recommendation**: approve
- Reason: The latest head fixes the previously blocking sticky-collapsed bar path, and the new regression test covers the first-click expand behavior after scroll-driven collapse.

**Level**: C
- Path initial: only `packages/web/**` changed -> C
- Semantic lift: none; this remains local Workspace center UI state/tests, not subscriptions, WebSocket frame parsing, or data-layer contracts.

**PR Summary**
- Problem solved: keeps the pinned chat summary open after manual expand at an already-scrolled position, including the sticky-collapsed bar users see after scrolling an expanded summary down.
- Design: records the manual-expand scroll position, requires fresh downward movement before sticky-collapse can fold again, and treats `scrollCollapsed` as expanded intent when toggling from the visible collapsed bar.
- Affected modules: `packages/web/src/pages/workspace/center/chat-summary.tsx`, `packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts`

**Review Notes**
✅ 1. The previous blocker is fixed: when `open=true && scrollCollapsed=true`, the click path now keeps `open` true and clears `scrollCollapsed`, so the first click on the sticky-collapsed bar expands the summary.
✅ 2. The manual-expand guard remains transient component state and does not create a second durable conversation/summary model.
✅ 3. Regression coverage now includes the default-collapsed-at-scroll-position path, later downward-scroll collapse, and sticky-collapsed first-click expand path.

**Executed**
- Approved this PR on follow-up review.
- Did not run tests, per review-mode constraints; relied on code inspection and the author's reported validation.

PR #1232, opened 2026-06-23
fix: rebind harness "user" to runtime so agents reliably chat send

yuezengwu:
## Problem (P0)

Agents on the Claude Code SDK regularly finish a teammate's request, write the answer to the **output stream**, and stop — without ever running `chat send`. The output stream is a reasoning trace that the teammate never receives, so to them it reads as **no reply**. Reproduces on ~1/3 of human-directed turns (worst on "give me an X" code/SQL-producing prompts).

## Root cause

A prompt conflict the brief was losing:

- The base `claude_code` preset asserts: *"all text you output outside of tool use is displayed to the user."* This matches the model's strong trained prior (output = talking to the user).
- The First Tree brief tried to **negate** it ("your output stream is not a reply"). That is a downstream, lower-salience instruction — it loses to the prior a meaningful fraction of the time.

## Fix — rebind, don't negate

Instead of fighting "output is displayed to the user", **rebind who "the user" is**. A new **runtime output contract**:

- pins the harness's *user* to the **First Tree runtime** (an automated operator that records the trace) — so the base sentence stays true but no longer implies a teammate was reached;
- frames **teammates** (human + agent) as a separate audience reached **only** by an explicit `chat send` / `ask` / `update`;
- adds the **console-vs-outbox** metaphor (finishing your thoughts writes the console; it does not send the outbox);
- stays accurate about visibility — the trace is **not private** (it surfaces as a live-activity preview), it is just never delivered as a message.

It is injected through `systemPrompt.append` (`buildClaudeQueryOptions`), which the SDK places after the base preset but at **higher salience than the project `CLAUDE.md`**. The shared brief (`agent-briefing.ts`) is reworded to the same framing so codex (which has no append channel) is covered too.

These three framings (runtime-as-user, reach = outbound publish, console/outbox) were synthesized together per the chat design discussion.

## `[From: …]` header enrichment

The inbound attribution header now carries the sender's **participant type** (human/agent — so the agent applies the right reply discipline) and the **message send time**:

```
[From: alice · type=human · sent=2026-06-23T13:30:00.000Z]
```

Threaded via a new optional `SessionMessage.createdAt` and a shared `formatFromHeader` (`SessionContext`) used by both the text path (`formatInboundContent`) and the image path in `claude-code.ts`, so every header is framed identically. Annotations are omitted when unknown, degrading to the bare `[From: <name>]`.

## Tests

- New `renderRuntimeOutputContract` guards (rebinding, console/outbox, accurate-visibility, no `agent-final-text`).
- New `buildClaudeQueryOptions` guards: contract always injected (even with no payload) and ordered before the per-chat context block.
- New `formatInboundContent` coverage for the `type=` / `sent=` annotations; existing `[From:]` assertions updated to the new format.
- Brief guards (`agent-briefing.test.ts`, `codex-bootstrap.test.ts`) updated from the old "reasoning trace" phrasing to the runtime-as-user framing.

`pnpm --filter @first-tree/client typecheck`, `pnpm check`, and the full client suite (1113 tests) pass.

## Scope

This is **P0** of a broader brief↔base-prompt audit. Follow-ups tracked separately: P1 carve-out for "NEVER create .md docs" vs Context Tree/memory writes, and P1 "NEVER commit unless asked" vs task-completion→PR.

### DIFF
--- a/packages/client/src/__tests__/agent-briefing.test.ts
+++ b/packages/client/src/__tests__/agent-briefing.test.ts
@@ -428,17 +428,20 @@ describe("buildAgentBriefing — # Working in First Tree subsections", () => {
     expect(briefing).toContain("first-tree chat update --description");
     expect(briefing).not.toMatch(/server rejects a `?chat send`? to a human/);
 
-    // yuezengwu 2026-06-23: reinstate the "your output stream is your reasoning
-    // trace" framing as a deliberate prompt principle (reverses the 2026-06-16
-    // removal), but with the PRECISE product boundary (codex R1/R5): the output
-    // stream is not an addressed reply and not durable chat history, yet it is
-    // NOT private — a one-line preview surfaces as live session activity
-    // (`liveActivity.detail` / `turnText`) to viewers. So communication is
-    // always an explicit send, and narration is never a substitute for it.
-    expect(briefing).toMatch(/output stream is your reasoning trace/i);
-    expect(briefing).toMatch(/never written to durable\s+chat history/i);
+    // yuezengwu 2026-06-23: rebind, rather than negate, the base Claude Code
+    // harness ("all text you output … is displayed to the user"). The brief
+    // pins the harness's "user" to the First Tree runtime and frames teammates
+    // as a separate audience reached only by an explicit send (console vs
+    // outbox). Keep the PRECISE product boundary (codex R1/R5): the output
+    // stream is not an addressed reply, yet it is NOT private — a one-line
+    // preview surfaces as live session activity (`liveActivity.detail` /
+    // `turnText`) to viewers. So communication is always an explicit send.
+    expect(briefing).toMatch(/the "user" the Claude Code harness writes to is the First Tree runtime/i);
+    expect(briefing).toMatch(/reasoning\/activity\s+trace/i);
+    expect(briefing).toMatch(/never delivered to\s+anyone as a message/i);
     expect(briefing).toMatch(/not\s+private/i);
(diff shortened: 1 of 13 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1232

**建议**: approve
- 理由: 改动把 Claude SDK 输出语义从“否定 output=reply”改成 runtime-as-user 的绑定，并同步覆盖 briefing 与 inbound header；和 First Tree 的 delivery model 一致，未发现阻塞问题。

**分级**: A
- 语义升级: 无

**PR 摘要**
- 解决问题: agent 把答案写到 runtime 输出流但未执行 `chat send`，导致队友收不到回复。
- 设计方案: 新增 runtime output contract，经 Claude SDK `systemPrompt.append` 始终注入并排在 Current Chat Context 前；共享 briefing 同步改成 runtime-as-user / console-vs-outbox framing；`[From: ...]` header 增加 sender type 和 sent time。
- 受影响模块: `packages/client/src/handlers/claude-code.ts`、`packages/client/src/runtime/{chat-context-section,agent-briefing,agent-io,handler,session-manager}.ts` 及对应测试。

**Review 意见**
1. 关键通过点: `systemPrompt.append` 在 no-payload/no-chatContext 下仍注入 contract，覆盖 PR 指向的 Claude SDK P0 路径。
2. 关键通过点: Header enrichment 复用同一 formatter，文本 path 和 Claude image path 一致，unknown sender/time 可退化到旧 header 形态。
3. 关键通过点: 测试覆盖 contract load-bearing 文案、append 顺序、header type/sent、precedingMessages，以及 briefing/codex guards。

**已执行**

PR #1235, opened 2026-06-24
chore: temporarily disable the claude-code-tui runtime

yuezengwu:
## What

Temporarily disable the `claude-code-tui` runtime — shown as **"Claude Code (TUI)"** in the new-agent dialog and **"Claude Code CLI"** on the client cards. Scope, as requested: **hide it from the UI + stop probing it**, behind a single central switch that reverts in one line.

Existing agents already bound to this runtime are untouched — they keep running and keep their label. This is implementation-only (no server-side runtime block).

## How

- **shared** — add `DISABLED_RUNTIME_PROVIDERS = ["claude-code-tui"]` and `isRuntimeProviderEnabled()` as the single source of truth. Re-enabling is a one-line revert (remove the entry).
- **client (stop probing)** — guard each capability probe *before* invoking it, so a disabled provider never spawns its binary (`claude` / tmux). Drop it from `PROBED_RUNTIME_PROVIDERS` so the degraded-capability re-probe loop never schedules itself just for it.
- **web (hide from UI)** — filter the disabled provider out of the new-agent dialog, onboarding runtime selection, and the clients cards (`PROVIDER_ORDER`) — including stale snapshots that still report it `ok`. Provider label / install-command maps are intentionally left intact so a previously-bound agent on this runtime still renders.

## Tests

- Updated the two tests that asserted the TUI provider was probed / shown.
- `pnpm typecheck` + `biome check` clean; `shared` + `client` + `web` suites green.
- Local regression by QA: candidate daemon uploads only `codex` + `claude-code` (no `claude-code-tui`, no TUI/tmux probe in the daemon log); web onboarding / clients cards / new-agent dialog no longer offer Claude Code CLI/TUI while keeping Claude Code + Codex; desktop 1280x800 and mobile 390x844 both clean. Verdict: PASS (the report's only caveat is an unrelated server EJS/CJS startup issue already fixed on `main@72ffd998`; this PR does not touch the server).

### DIFF
--- a/apps/cli/src/__tests__/runtime-auth-login.test.ts
+++ b/apps/cli/src/__tests__/runtime-auth-login.test.ts
@@ -1,6 +1,6 @@
 import { BROWSER_LOGIN_TIMEOUT_MS, type CodexBrowserLoginOptions, type LoginOutcome } from "@first-tree/client";
 import type { CapabilityEntry } from "@first-tree/shared";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { runRuntimeAuthLogin } from "../core/runtime-auth-login.js";
 
 const NOW = Date.parse("2026-06-22T12:00:00.000Z");
@@ -151,6 +151,10 @@ describe("runRuntimeAuthLogin — claude-code browser OAuth (cc/codex parity)", 
   function claudeHarness(opts: { resolveOk?: boolean; outcome?: LoginOutcome; probeResult?: CapabilityEntry }) {
     const calls: Recorded[] = [];
     const logs: string[] = [];
+    // Spy so a test can assert the TUI probe is NOT spawned while claude-code-tui
+    // is disabled (it shares the Claude keychain, but "stop probing it" must hold
+    // on this login-reflection path too).
+    const probeClaudeTui = vi.fn(async (): Promise<CapabilityEntry> => opts.probeResult ?? unauthEntry());
     const deps = {
       currentEntry: (): CapabilityEntry | undefined => undefined,
       setProviderEntry: async (provider: string, entry: CapabilityEntry): Promise<void> => {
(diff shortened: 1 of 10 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1235

- Reason: The current disabled behavior is mostly correct, but one web selection branch no longer satisfies the PR's stated one-line re-enable contract.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: temporarily remove `claude-code-tui` from new runtime selection and capability probing without breaking existing agents already bound to it.
- Design: add a shared disabled-provider switch, skip disabled providers before probe invocation, and filter disabled providers from new-agent/onboarding/client-card UI while keeping labels/config support intact.
- Affected modules: `packages/shared`, `packages/client`, `packages/web`

**Review Notes**
❌ 1. `pickPreferredRuntime()` no longer preserves the central-switch re-enable path. [R1 / `packages/web/src/components/new-agent-dialog.tsx:155`]

The PR says re-enabling TUI should be a one-line removal from `DISABLED_RUNTIME_PROVIDERS`, but this function deleted the explicit `claude-code-tui` preference branch instead of guarding it. With this patch as-is, removing `claude-code-tui` from the disabled list would re-show/re-probe it, but a client with both TUI and Codex `ok` would now auto-pick `codex`; before this PR the order was Claude Code -> Claude Code TUI -> Codex, and the nearby comment still documents that order. Please keep the TUI branch and guard it, e.g. `if (isRuntimeProviderEnabled("claude-code-tui") && caps["claude-code-tui"]?.state === "ok") return "claude-code-tui";`, before the Codex branch. That keeps stale disabled snapshots skipped today while making the documented one-line revert actually restore prior behavior.

**Executed**
- Submitted request_changes.

yuezengwu:
Thanks for the catch on **R1** — agreed, and fixed in `0b63a5cb`.

`pickPreferredRuntime()` now keeps the explicit `claude-code-tui` branch and guards it on the central switch, exactly as suggested:

```ts
if (caps["claude-code"]?.state === "ok") return "claude-code";
if (isRuntimeProviderEnabled("claude-code-tui") && caps["claude-code-tui"]?.state === "ok") return "claude-code-tui";
if (caps.codex?.state === "ok") return "codex";
```

- **Today (disabled):** the guard short-circuits, so a stale `ok` TUI snapshot is skipped and Codex wins — same as the prior commit's behavior.
- **After the one-line revert** (remove `claude-code-tui` from `DISABLED_RUNTIME_PROVIDERS`): the documented Claude Code → Claude Code CLI → Codex priority is fully restored, including TUI winning over Codex.

`pnpm typecheck` + `biome` clean; `web` suite green (974 passed).

baixiaohang (changes requested):
## Review by codex-assistant - PR #1235

- Reason: The prior new-agent preference issue is fixed, but there is still a daemon-side path that probes `claude-code-tui` while it is disabled.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: temporarily remove `claude-code-tui` from new runtime selection and capability probing without breaking existing agents already bound to it.
- Design: add a shared disabled-provider switch, skip disabled providers before probe invocation, and filter disabled providers from new-agent/onboarding/client-card UI while keeping labels/config support intact.
- Affected modules: `packages/shared`, `packages/client`, `packages/web`, plus daemon runtime-auth call paths.

**Review Notes**
❌ 1. Claude runtime-auth still re-probes the disabled TUI provider. [R1 / `apps/cli/src/core/runtime-auth-login.ts:184`]

The PR guards the aggregate capability probes, but `runClaudeRuntimeAuth()` still does `Promise.all([probeClaude(), probeClaudeTui()])` and then writes a `claude-code-tui` entry on every Claude login reflection. That means a user who clicks Connect for `claude-code` can still spawn the TUI probe (`claude` / `tmux`) even though `claude-code-tui` is in `DISABLED_RUNTIME_PROVIDERS`, which violates the stated scope "stop probing it" and splits the disable policy across two code paths. Please apply the same central switch here before invoking `probeClaudeTui()` and before `setProviderEntry("claude-code-tui", ...)`; when disabled, only reflect `claude-code`. The existing runtime-auth test at `apps/cli/src/__tests__/runtime-auth-login.test.ts:175` should also be updated/pinned so disabled TUI is not re-probed through this path.

✅ 2. The previous `pickPreferredRuntime()` one-line re-enable issue is fixed in `0b63a5cb`. [R1 / `packages/web/src/components/new-agent-dialog.tsx:160`]

**Executed**
- Submitted request_changes.

yuezengwu:
Good catch on the second probe path — fixed in `f3d49b3f`.

`runClaudeRuntimeAuth()`'s `reflect()` now honors the same central switch as the capability aggregator:

```ts
if (isRuntimeProviderEnabled("claude-code-tui")) {
  const [cc, tui] = await Promise.all([probeClaude(), probeClaudeTui()]);
  await deps.setProviderEntry("claude-code", attachAuthError(cc, failure, now()));
  await deps.setProviderEntry("claude-code-tui", tui);
} else {
  const cc = await probeClaude();
  await deps.setProviderEntry("claude-code", attachAuthError(cc, failure, now()));
}
```

- **Disabled (today):** a `claude-code` Connect reflects `claude-code` only — `probeClaudeTui()` is never invoked (no `claude` / tmux spawn) and no `claude-code-tui` entry is written. The disable policy is now single-sourced.
- **Enabled (after the one-line revert):** the prior shared-keychain both-providers `Promise.all` path is unchanged.

Also pinned the `runtime-auth-login` tests (`apps/cli/src/__tests__/runtime-auth-login.test.ts`): they now assert the TUI probe spy is **not** called and no tui entry is written while disabled.

Verified no other direct `probeClaudeCodeTuiCapability` call sites remain (only the aggregator + this login path). `tsc` + `biome` clean; CLI runtime-auth-login suite green (11 passed).

baixiaohang (approved):
Approved after follow-up fixes. The disabled-provider switch now covers the UI selection paths, aggregate capability probes, reconnect/revalidate probes, and the Claude runtime-auth reflection path; existing TUI labels/handlers remain available for already-bound agents.

PR #1237, opened 2026-06-24
fix(client): reconnect after inbox recovery timeout

baixiaohang:
## Summary
- close the current registered WebSocket when a chat-scoped inbox recovery confirmation times out
- let the existing reconnect + agent bind reset path recover delivered/unacked inbox rows
- cover timeout reconnect behavior while ensuring server-side recover_failed does not reconnect

## Tests
- pnpm --filter @first-tree/client exec vitest run src/__tests__/client-connection-websocket-coverage.test.ts
- pnpm --filter @first-tree/client exec vitest run src/__tests__/client-connection-heartbeat-timeout.test.ts src/__tests__/client-connection-auth-expired.test.ts src/__tests__/session-manager.test.ts
- pnpm --filter @first-tree/client typecheck
- pnpm --filter @first-tree/client test
- pnpm exec biome check packages/client/src/client-connection.ts packages/client/src/__tests__/client-connection-websocket-coverage.test.ts
- git diff --check

### DIFF
--- a/packages/client/src/__tests__/client-connection-websocket-coverage.test.ts
+++ b/packages/client/src/__tests__/client-connection-websocket-coverage.test.ts
@@ -530,6 +530,37 @@ describe("ClientConnection — WebSocket edge coverage", () => {
       reason: "recover_failed",
     });
     await expect(rejected).rejects.toThrow("recover_failed");
+    expect(socket.closeCalls).toHaveLength(0);
+  });
+
+  it("forces a reconnect when an inbox recovery confirmation times out", async () => {
+    vi.useFakeTimers();
+    const connection = await makeConnection();
+    const internal = priv(connection);
+    const events: string[] = [];
+    connection.on("reconnecting", () => events.push("reconnecting"));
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the client-side recovery timeout path. The change closes the current registered socket only when the local inbox recovery confirmation times out, keeps server-side `recover_failed` as a normal rejected recovery without reconnecting, and relies on the existing reconnect + `agent:bind` reset path to recover delivered-but-unacked rows.

Verified locally:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/client-connection-websocket-coverage.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/client-connection-heartbeat-timeout.test.ts src/__tests__/client-connection-auth-expired.test.ts src/__tests__/session-manager.test.ts`
- `pnpm exec biome check packages/client/src/client-connection.ts packages/client/src/__tests__/client-connection-websocket-coverage.test.ts`
- `git diff --check`

No blocking issues found.

PR #1238, opened 2026-06-24
fix(onboarding): treat slow agent bring-up as still-starting, not failure

Gandy2025:
## What & why

Part of the fix for onboarding's "unanswered first chat" (a new user finishes setup, lands in their first chat, and the agent never responds). The root cause is a three-layer decoupling; this PR is the **onboarding-timing layer**.

The create-agent step polled for the new agent coming online for only **30s**, then switched to a failure-framed screen ("Your agent isn't online yet … try again") whose only action re-polled the same agent. But a cold runtime (first claude-code/codex spawn, proxy-slow TLS) routinely takes longer, and the **server itself does not consider an agent offline until 60s** without a heartbeat. So a healthy-but-slow agent was told it had failed, with no graceful exit.

## Changes

- Extend the online-readiness poll **30s → 60s** (`use-agent-creation.ts`), matching the server's presence-reaper threshold, so the late state is reached only when the agent is genuinely overdue, not merely cold-starting.
- Reframe the late state from an error into a hopeful **"taking longer than usual"** with two paths: **Keep waiting** (re-poll) or a graceful, resumable **I'll finish later** — so a stuck runtime is never a dead end (`step-create-agent.tsx`, `copy.ts`).
- Update the hook timeout test (31s → 61s) and the SSR-smoke copy assertion.

## Verification

- `pnpm check` (biome) clean · `pnpm --filter @first-tree/web typecheck` clean · `pnpm --filter @first-tree/web test` → **974 passed**.
- Visually verified the reframed state in the DEV `/preview/onboarding` gallery (admin · Agent creation states · Timeout), light + dark themes.

## Scope

One of a three-layer fix for the unanswered-first-chat root cause:
1. **(runtime)** retry agent config-fetch on bind so a transient failure self-heals instead of bricking the agent — in progress separately.
2. **(chat)** surface agent presence / "no response yet" in the first chat so a not-yet-live agent is never a silent dead end — next.
3. **(this PR)** onboarding create-agent timing + copy.

### DIFF
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1277,7 +1277,7 @@ describe("page SSR smoke coverage", () => {
     ).toContain("Bringing your agent online");
     expect(
       await renderOnboardingStep(<StepCreateAgent />, { activeStep: "create-agent", agentPhase: "timeout" }),
-    ).toContain("online yet");
+    ).toContain("taking longer");
     expect(await renderOnboardingStep(<StepConnectCode />, { activeStep: "connect-code" })).toContain(
       "Loading your repos",
     );
--- a/packages/web/src/pages/onboarding/__tests__/hooks-flow.test.tsx
+++ b/packages/web/src/pages/onboarding/__tests__/hooks-flow.test.tsx
@@ -308,7 +308,7 @@ describe("onboarding hooks and flow", () => {
         organizationId: null,
       });
       await Promise.resolve();
-      await vi.advanceTimersByTimeAsync(31_000);
+      await vi.advanceTimersByTimeAsync(61_000);
       await promise;
     });
     await slowCreate;
@@ -317,7 +317,7 @@ describe("onboarding hooks and flow", () => {
     const retry = act(async () => {
       const promise = expectHookValue(latest.current).retry();
       await Promise.resolve();
-      await vi.advanceTimersByTimeAsync(31_000);
+      await vi.advanceTimersByTimeAsync(61_000);
       await promise;
     });
     await retry;
--- a/packages/web/src/pages/onboarding/copy.ts
+++ b/packages/web/src/pages/onboarding/copy.ts
@@ -282,12 +282,14 @@ export const COPY = {
     // yet".
     creating: "Bringing your agent online…",
     creatingHint: "This usually takes a few seconds.",
-    // Timeout: added but didn't report online within 30s. One short line — the
-    // shell already renders the step h1, so no second title. Situation + the
-    // condensed likely cause (the three separate causes — sleep / lost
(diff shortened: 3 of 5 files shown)


yuezengwu (approved):
Approved.

Goal/core change: this PR keeps onboarding's create-agent step in a still-starting state for the full 60s server liveness window, then reframes the late state around two recoverable paths: keep polling the same created agent, or finish later through the existing resumable onboarding dismissal path.

I checked the web hook/UI diff against the server presence cleanup defaults and the onboarding Context Tree constraints. No DB or core data-structure change is involved, and I did not find a blocking issue.

Verified locally:
- `pnpm --filter @first-tree/web test -- packages/web/src/pages/onboarding/__tests__/hooks-flow.test.tsx packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx` (ran the web suite: 124 files / 974 tests passed)

Non-blocking note: `use-agent-creation.ts` still has one function comment saying `or a 30s timeout`; the new top-level comment and behavior are correct at 60s, so this is only stale prose.

yuezengwu (approved):
Re-reviewed latest head after the synchronize event.

The new commit only updates the stale function comment from `30s timeout` to `60s timeout`, matching the behavior already reviewed. No new behavior, schema, or data-structure change.

I checked the incremental diff (`ccd7f879..6b5399e6`) and `git diff --check`; both are clean. I did not rerun the suite because this update is prose-only and the prior reviewed head already passed the web suite locally.

baixiaohang (approved):
## Review by codex-assistant - PR #1238

**建议**: approve
- 理由: create-agent 等待窗口、超时后的两条恢复路径和现有 onboarding 状态模型一致，未发现阻塞问题。

**分级**: C
- Path 初判: 仅含 `packages/web/**` -> C
- 语义升级: 无

**PR 摘要**
- 解决问题: 避免冷启动 runtime 在 30s 后被误导成失败态，导致新用户首聊前卡在错误感文案里。
- 设计方案: `use-agent-creation.ts` 将 readiness polling 超时延到 60s；timeout UI 改为 “taking longer” 并提供 Keep waiting 与现有 finish-later 退出；相关 hook/SSR 测试同步更新。
- 受影响模块: `packages/web/src/pages/onboarding/**` 与 onboarding SSR smoke 测试。

**Review 意见**
✅ 1. 60s 前端等待与服务端 presence cleanup 默认值一致；`client-status` 仍只读 `agent_presence.status`，Web 没有新增本地状态推断。
✅ 2. Timeout 后的 Keep waiting 复用同一个已创建 agent 重新轮询，finish-later 走现有 `dismissOnboarding()`，不会写 completed，也保留后续恢复路径。
✅ 3. 最新 synchronize 只把 stale comment 从 30s 改成 60s；没有新增行为、schema 或数据结构变更。

**已执行**
- 已 approve
- 未运行检查或测试；本轮 review 仅读 Context Tree、PR diff、相关源码和增量提交。

PR #1240, opened 2026-06-24
feat(agents): adopt a member's first agent as their delegate

Gandy2025:
## What

Two coupled changes that make the human delegate agent (`delegateMention`) useful by default:

1. **Auto-adopt the first agent as delegate.** When a member creates their first non-human agent and has no delegate set, `createAgent` adopts it as their delegate (`delegateMention` on their human agent), atomically in the same transaction. The delegate drives the new-chat default recipient (`pickDefault`) and the GitHub `@mention` forward target — value previously gated behind a manual step in the profile editor that most users never took.

2. **Drop the visibility filter from the delegate pickers.** A private agent (the most natural "personal assistant") is now eligible as a delegate in **both** entry points — Agent Detail (`profile-edit-dialog`) and the Team page (`selectDelegateCandidates`). The server already accepts a private delegate — `validateDelegateMentionTarget` checks same-org only, webhook routing checks same-org + active, never visibility — so the pickers were stricter than the backend. They now mirror it: `managerId === you` + active + `type=agent`.

## Why

The delegate is what makes "ask my assistant" zero-step and what lets you follow GitHub entities, but it defaulted to unset and lived only in the profile editor, so most users never set it. A member's first agent is almost always their de-facto assistant, so adopting it removes the manual step.

Allowing a private delegate resolves a standing inconsistency: a new agent defaults to `private` (a deliberate anti-surprise choice — see `new-agent-dialog.tsx`), yet the old pickers required `organization` visibility, so the natural delegate couldn't be selected. Keeping the first agent private *and* adopting it as the delegate is the most coherent default.

### Guards (auto-adoption)
- **self-create only** — adoption fires only when the caller creates their own agent (`managerId === scope.memberId`). Delegate is a personal choice, so an admin creating an agent *for* another member — and the bootstrap / system-webhook paths — never adopt, matching the self-only PATCH guard.
- **only-if-unset, atomically** — the final UPDATE carries `delegateMention IS NULL` in its WHERE, so concurrent first-creates can't clobber a freshly-set delegate.
- **first-agent-only** — the just-inserted row makes the count 1, so a 2nd agent never steals the delegate.
- **reversible** — the member can re-point or clear it anytime.

## Test plan
- `agent-first-agent-delegate.test.ts` — service cases (first agent adopted; a **private** first agent adopted; no adoption without the self-create intent; 2nd agent unchanged; pre-existing delegate not overwritten) **plus route-level** (`POST /orgs/:orgId/agents`): a self-create sets the delegate; an **admin-create-for-other leaves the target's delegate `null`** (the authorization regression).
- Web: Agent Detail + Team picker tests assert a private owned agent is selectable; Team empty-state copy updated.
- Full server suite (1552 tests) + full web suite (975) green; `pnpm typecheck` (12/12) + `biome check` clean.

## Context Tree
Tree write-back: **agent-team-foundation/first-tree-context#596** (delegate agent model node) — land alongside or shortly after this PR.

### DIFF
--- a/packages/server/src/services/agent.ts
+++ b/packages/server/src/services/agent.ts
@@ -17,7 +17,7 @@ import {
   runtimeProviderSchema,
 } from "@first-tree/shared";
 import { getServerCliBinding } from "@first-tree/shared/channel";
-import { and, count, desc, eq, getTableColumns, ilike, lt, ne, or } from "drizzle-orm";
+import { and, count, desc, eq, getTableColumns, ilike, isNull, lt, ne, or } from "drizzle-orm";
 import type { PostgresJsDatabase } from "drizzle-orm/postgres-js";
 import type { Database } from "../db/connection.js";
 import { agentConfigs } from "../db/schema/agent-configs.js";
@@ -371,7 +371,7 @@ async function resolveFallbackManagerId(db: Database, orgId: string): Promise<st
 export async function createAgent(
   db: Database,
   data: CreateAgent & { managerId?: string },
-  options: { force?: boolean } = {},
+  options: { force?: boolean; adoptAsDelegateIfFirst?: boolean } = {},
 ) {
   const uuid = uuidv7();
   const name = data.name ?? null;
@@ -537,19 +537,28 @@ export async function createAgent(
       // without a manual trip to the profile editor. Runs inside this
       // transaction so the delegate is set atomically with the agent insert.
       // Guards keep it safe and unsurprising:
-      //   - only `type=agent` created through a user path (`data.managerId`
-      //     supplied) — excludes the human-bootstrap insert and the
-      //     system/webhook agent path (branch 3 omits `managerId`);
-      //   - only when the member has no delegate yet — never overwrites a
-      //     deliberate choice (incl. a delegate pointing at someone else's
-      //     org-visible agent set before they had any agent of their own);
-      //   - only their FIRST agent — the just-inserted row makes the count 1,
-      //     so a 2nd agent never steals the delegate.
+      //   - SELF-CREATED ONLY. `delegateMention` is a personal choice the
+      //     self-only API guard reserves to the member (an admin PATCHing
+      //     another member's delegate is rejected). The org agent route lets an
+      //     admin create an agent FOR another member, so gating on a resolved
+      //     managerId alone would let create-for-Bob silently set Bob's
+      //     delegate — the same write the PATCH guard forbids. The caller is
+      //     the only one who knows "this is my own create", so the route passes
+      //     `adoptAsDelegateIfFirst` true only when `managerId === scope.memberId`.
+      //     Every non-self path (admin-for-other, member bootstrap, system /
+      //     webhook) omits it and never triggers adoption.
+      //   - FIRST agent only — the just-inserted row makes the count 1, so a
+      //     2nd agent never steals the delegate.
+      //   - ONLY-IF-UNSET, atomically — the final UPDATE carries
+      //     `delegateMention IS NULL` in its WHERE, so two concurrent
+      //     first-creates can't both win on a stale read; the second's write
+      //     no-ops instead of clobbering the first. Never overwrites a
+      //     deliberate choice either.
       // The agent stays whatever visibility it was created with (private by
       // default); the picker and webhook routing both accept a private
       // delegate, so no visibility change is forced. Reversible: the member
       // can re-point or clear it anytime.
-      if (data.managerId && data.type === AGENT_TYPES.AGENT) {
+      if (options.adoptAsDelegateIfFirst && data.type === AGENT_TYPES.AGENT) {
         const [manager] = await tx
           .select({ humanAgentId: members.agentId })
           .from(members)
@@ -573,7 +582,10 @@ export async function createAgent(
                 ),
               );
             if ((tally?.value ?? 0) === 1) {
-              await tx.update(agents).set({ delegateMention: row.uuid }).where(eq(agents.uuid, human.uuid));
+              await tx
+                .update(agents)
+                .set({ delegateMention: row.uuid })
+                .where(and(eq(agents.uuid, human.uuid), isNull(agents.delegateMention)));
             }
           }
         }
--- a/packages/server/src/services/agent.ts
+++ b/packages/server/src/services/agent.ts
@@ -530,6 +530,55 @@ export async function createAgent(
         })
         .onConflictDoNothing();
 
+      // First-agent → delegate adoption. When a member creates their FIRST
+      // non-human agent and hasn't picked a delegate yet, adopt it as their
+      // delegate (`delegateMention` on their human agent) so it becomes the
+      // new-chat default recipient and the GitHub @mention forward target
+      // without a manual trip to the profile editor. Runs inside this
+      // transaction so the delegate is set atomically with the agent insert.
+      // Guards keep it safe and unsurprising:
+      //   - only `type=agent` created through a user path (`data.managerId`
+      //     supplied) — excludes the human-bootstrap insert and the
+      //     system/webhook agent path (branch 3 omits `managerId`);
+      //   - only when the member has no delegate yet — never overwrites a
+      //     deliberate choice (incl. a delegate pointing at someone else's
+      //     org-visible agent set before they had any agent of their own);
+      //   - only their FIRST agent — the just-inserted row makes the count 1,
+      //     so a 2nd agent never steals the delegate.
+      // The agent stays whatever visibility it was created with (private by
+      // default); the picker and webhook routing both accept a private
+      // delegate, so no visibility change is forced. Reversible: the member
+      // can re-point or clear it anytime.
+      if (data.managerId && data.type === AGENT_TYPES.AGENT) {
+        const [manager] = await tx
+          .select({ humanAgentId: members.agentId })
+          .from(members)
+          .where(eq(members.id, managerId))
+          .limit(1);
+        if (manager) {
+          const [human] = await tx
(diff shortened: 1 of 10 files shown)


packages/server/src/services/agent.ts
```diff
@@ -530,6 +530,55 @@ export async function createAgent(
...
+      //   - only when the member has no delegate yet — never overwrites a
+      //     deliberate choice (incl. a delegate pointing at someone else's
+      //     org-visible agent set before they had any agent of their own);
+      //   - only their FIRST agent — the just-inserted row makes the count 1,
+      //     so a 2nd agent never steals the delegate.
+      // The agent stays whatever visibility it was created with (private by
+      // default); the picker and webhook routing both accept a private
+      // delegate, so no visibility change is forced. Reversible: the member
+      // can re-point or clear it anytime.
+      if (data.managerId && data.type === AGENT_TYPES.AGENT) {
```
yuezengwu: This guard is broader than the self-created/user-created path described in the PR. `POST /orgs/:orgId/agents` always passes a `managerId`, and admins may choose another member's `managerId`; with this condition, creating that member's first agent also writes their human `delegateMention`. That bypasses the existing self-only route guard for delegate changes. Please make adoption opt-in from the route only when the caller is creating for themselves, and cover the admin-creates-for-other-member case.

yuezengwu (changes requested):
Reviewed the PR. The implementation goal is clear: `createAgent` now auto-adopts a member's first non-human agent as their human delegate, and the profile delegate picker now allows the member's own private active agents.

I'm requesting changes for one blocker: the current service guard treats any supplied `managerId` as a user-created/self-owned create. The public org agent route supplies `managerId` for every create, and admins can set it to another member. That bypasses the existing self-only delegate rule and lets an admin change another member's `delegateMention` as a side effect of creating their first agent. Please gate auto-adoption on the self-created path (for example, route passes an explicit option only when `scope.memberId === managerId`) and add a route-level regression test for admin-creates-for-other-member.

While touching this, it would also be worth making the final delegate update conditional on `delegateMention IS NULL` in SQL so concurrent creates cannot overwrite a freshly-set delegate based on a stale read.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1240

**建议**: request_changes
- 理由: 当前实现会让 admin 创建“别人的第一个 agent”时隐式改写对方 delegate，并且 Team 页的 delegate picker 仍排除 private agent，和本 PR 的产品/后端口径不一致。

**分级**: B-high
- Path 初判: 含 `packages/server/src/services/agent.ts` 且命中 `services/agent*` -> B-high
- 语义升级: 无

**PR 摘要**
- 解决问题: 新建首个非人 agent 后自动作为 human delegate，减少手动设置；允许 private agent 作为 delegate。
- 设计方案: 在 `createAgent` 事务内写 human agent 的 `delegateMention`；Web profile delegate selector 去掉 visibility 过滤，并补测试。
- 受影响模块: `packages/server/src/services/agent.ts`、agent service tests、Agent Detail profile dialog、New Agent dialog 注释。

**Review 意见**
❌ 1. `createAgent` 现在绕过了“delegate 是个人选择，只能本人设置”的授权边界。[R4/R5 / `packages/server/src/services/agent.ts:552`, `packages/server/src/api/orgs/agents.ts:128`, `packages/server/src/api/agents.ts:107`]

`/orgs/:orgId/agents` 允许 admin 在 request body 里指定任意同 org 的 `managerId`，而新的 auto-adoption 只检查 `data.managerId` 是否存在。结果是 admin 给 Bob 创建 Bob 的第一个 agent 时，会隐式把 Bob 的 human agent `delegateMention` 改成这个新 agent。这和现有 PATCH 路径的 self-only guard 冲突：admin 明确 PATCH Bob 的 delegate 会被拒绝，但通过 create-for-Bob 可以达到同样结果。建议把 auto-adoption 限定为 caller member 自己创建自己的 agent（例如路由层只在 `managerId === scope.memberId` 时启用/传入 adoption intent），并补一个 admin-create-for-other 不会改写 delegate 的 route/service 测试。

❌ 2. Team 页仍保留旧的 team-visible-only delegate picker，private delegate 在另一个入口仍不可选。[R1/R5 / `packages/web/src/pages/team/index.tsx:538`, `packages/web/src/pages/team/team-table.tsx:876`, `packages/web/src/pages/team/__tests__/team-groups.test.ts:272`]

PR body 说“Drop the visibility filter from the delegate picker”，Agent Detail picker 已经改了，但 Team 页的 `selectDelegateCandidates` 仍要求 `visibility === "organization"`，文案也仍显示 “Only team-visible agents can be a delegate.” 这会导致同一个 delegate 设置在两个用户可见入口里规则不一致：用户能在 Agent Detail 选 private assistant，却不能在 Team 的 inline delegate selector 选它。建议同步移除这里的 visibility filter，并更新对应 copy/test。

⚠️ 3. auto-adoption 的 only-if-unset guard 不是并发原子的。[R4 / `packages/server/src/services/agent.ts:564`, `packages/server/src/services/agent.ts:576`]

两个并发的 first-agent creates 都可能在自己的事务里读到 `delegateMention = null` 且 tally=1；第二个 `UPDATE ... WHERE uuid = human.uuid` 会在第一个提交后继续覆盖 delegate。既然 PR 明确把行为描述为 atomic / only-if-unset，建议至少把 `delegateMention IS NULL` 放进最终 UPDATE 的 WHERE guard（或锁 human row）并补并发/conditional-update 覆盖，避免 last writer wins。

**已执行**
- 已提交 request_changes review
- 未运行代码检查或单测（按 PR Review 模式）
- 读取了 Context Tree 相关节点：`system/cloud/agents/identity.md`、`system/cloud/agents/visibility-and-membership.md`、`product/surfaces/team-and-agent-model.md`、`system/cloud/github/github-entity-chat-binding.md`；`tree tree` 自动刷新降级为本地副本，提示 `fatal: Cannot fast-forward to multiple branches.`

Gandy2025:
Thanks @yuezengwu @baixiaohang — all three addressed in 1f47ffb1.

**1. Admin-creates-for-other bypassing the self-only delegate guard (blocker).** Fixed. Adoption now requires an explicit self-create intent: the route passes `adoptAsDelegateIfFirst` to `createAgent` only when `managerId === scope.memberId`. An admin creating an agent *for* another member — and the bootstrap and system/webhook paths — never trigger adoption, so create-for-other can no longer set another member's `delegateMention` (the write the self-only PATCH guard already forbids). Added a route-level regression test: `POST /orgs/:orgId/agents` admin-for-other leaves the target's delegate `null`; a self-create sets it.

**2. Team-page picker still visibility-filtered (blocker).** Fixed. `selectDelegateCandidates` (`team/index.tsx`) no longer filters on visibility — it now matches the Agent Detail picker (own + active + `type=agent`), so a private assistant is selectable from both entry points. Updated the empty-state copy ("Create an agent to set as your delegate.") and the unit + DOM tests.

**3. only-if-unset not atomic (concern).** Fixed. The final UPDATE now carries `delegateMention IS NULL` in its WHERE, so two concurrent first-creates can't both win on a stale read — the second write no-ops instead of clobbering the first.

Verification: `pnpm typecheck` (12/12) and `biome check` clean; full server suite (1552 tests) and full web suite (975) green.

yuezengwu (approved):
Re-reviewed `1f47ffb1`; the issues from my previous request are addressed.

What I checked:
- `POST /orgs/:orgId/agents` now passes `adoptAsDelegateIfFirst` only when `managerId === scope.memberId`, so admin-create-for-other no longer changes another member's delegate.
- The new route-level tests cover self-create adoption and admin-for-other no-op behavior.
- The final delegate update now includes `delegateMention IS NULL`, so stale concurrent reads cannot clobber an already-set delegate.
- Team page and Agent Detail delegate pickers now use the same candidate rule: own + active + `type=agent`, with no visibility filter.

Approved. Local verification note: I ran static review plus `git diff --check` and merge-tree; this worktree has no `node_modules`, so I did not rerun Vitest locally. Non-blocking cleanup: the PR body guard bullet still mentions `data.managerId` as the adoption gate; it should be aligned with the new explicit `adoptAsDelegateIfFirst` intent wording, and the promised Context Tree write-back should stay linked/close to merge.

baixiaohang (approved):
## Review by codex-assistant - PR #1240

**建议**: approve
- 理由: `1f47ffb1` 已修复上一轮 request_changes 的授权边界、Team picker 口径和 only-if-unset 原子性问题，未发现新的阻塞项。

**分级**: B-high
- Path 初判: 含 `packages/server/src/services/agent.ts` 且命中 `services/agent*` -> B-high
- 语义升级: 无

**PR 摘要**
- 解决问题: 自创建首个非 human agent 时自动设置 human delegate，并让 private agent 可作为 delegate。
- 设计方案: `/orgs/:orgId/agents` 只在 `managerId === scope.memberId` 时传入 `adoptAsDelegateIfFirst`；`createAgent` 在事务中 first-agent-only 且 `delegateMention IS NULL` 条件更新；Agent Detail 与 Team picker 统一为 own + active + `type=agent`，不再按 visibility 过滤。
- 受影响模块: agent service / org agent route / delegate pickers / tests。

**Review 意见**
✅ 1. admin-create-for-other 不再绕过 self-only delegate 规则，并补了 route-level regression test。
✅ 2. Team 页和 Agent Detail 的 delegate candidate 规则已统一，private owned active agent 可在两个入口选择。
✅ 3. 最终 UPDATE 加入 `delegateMention IS NULL`，避免 stale read 下覆盖已设置 delegate。

**已执行**
- 已按 baixiaohang 授权提交 approve
- 未运行代码检查或单测（按 PR Review 模式）；作者声明 full server/web suites、typecheck、biome 均通过

PR #1244, opened 2026-06-24
fix(web): resume foreground polls on focus/pageshow so onboarding doesn't stick on "connecting"

Gandy2025:
## Problem

In the onboarding **Connect your computer** step, users sometimes had to **manually refresh the page** before the UI updated to "connected" — otherwise it stayed on "connecting…" indefinitely, even though the computer had already connected server-side.

## Root cause

`runVisibilityAwareInterval` (the foreground-only poll used by the onboarding connect step and the new-agent / new-connection dialogs) pauses the poll while the tab is hidden and resumes it **only** on `visibilitychange` → visible.

The connect step’s normal flow is to leave the browser for a terminal (to run the install one-liner) and come back. Returning from **another application** frequently does **not** fire a reliable `visibilitychange` — that return is a window-focus change, which is not always a page-visibility change, and the two diverge across browser / OS / window occlusion. When the event is missed, the paused poll never restarts, so the UI is stale until a hard refresh (which remounts and re-fetches).

## Fix

Add `window` `focus` and `pageshow` as additional **resume** triggers alongside `visibilitychange`:

- `focus` covers app/tab refocus; `pageshow` covers bfcache restore.
- Resuming is idempotent — `start()`’s handle guard makes a refocus while already running a no-op (no double tick, no stacked interval), and it never starts while the page is genuinely hidden.
- **Stop** stays bound to `visibilitychange` → hidden only, so a mere blur (window still visible, e.g. side-by-side with a terminal) does not pause the poll.

One change to the shared helper fixes the same latent issue at all three call sites (onboarding connect-computer, new-agent dialog, new-connection dialog).

## Tests

`packages/web/src/lib/__tests__/visibility-interval.test.ts` — added cases: resume on `focus` when `visibilitychange` is missed (the bug), idempotent refocus while running, no resume on `focus` while still hidden, and listener teardown. Full web suite green (979 tests); biome + typecheck clean.

### DIFF
--- a/packages/web/src/lib/__tests__/visibility-interval.test.ts
+++ b/packages/web/src/lib/__tests__/visibility-interval.test.ts
@@ -61,4 +61,79 @@ describe("runVisibilityAwareInterval", () => {
 
     cleanup();
   });
+
+  it("resumes on window focus when visibilitychange is missed (app-switch return)", async () => {
+    const { runVisibilityAwareInterval } = await import("../visibility-interval.js");
+    const tick = vi.fn();
+
+    const cleanup = runVisibilityAwareInterval(tick, 1000);
+    expect(tick).toHaveBeenCalledTimes(1);
+
+    // Tab goes hidden → pause (this event does fire).
+    setHidden(true);
+    document.dispatchEvent(new Event("visibilitychange"));
+    vi.advanceTimersByTime(3000);
+    expect(tick).toHaveBeenCalledTimes(1);
+
+    // User returns from a terminal: the window regains focus but the matching
+    // `visibilitychange` → visible never arrives. `focus` must resume the poll
+    // and re-tick immediately — this is the onboarding stuck-on-"connecting" fix.
+    setHidden(false);
+    window.dispatchEvent(new Event("focus"));
+    expect(tick).toHaveBeenCalledTimes(2);
+    vi.advanceTimersByTime(1000);
+    expect(tick).toHaveBeenCalledTimes(3);
+
+    cleanup();
+  });
+
+  it("ignores focus while already running (no double tick, no stacked interval)", async () => {
+    const { runVisibilityAwareInterval } = await import("../visibility-interval.js");
+    const tick = vi.fn();
+
+    const cleanup = runVisibilityAwareInterval(tick, 1000);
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1244

**建议**: approve
- 理由: 变更集中在共享前台轮询 helper，新增 resume 触发保持幂等并补齐 teardown/tests，没有发现阻塞风险。

**分级**: C
- Path 初判: 仅 `packages/web/**` -> C
- 语义升级: 无；未触及订阅、WS 帧解析或数据层

**PR 摘要**
- 解决问题: onboarding connect-computer 等前台轮询在返回页面时可能因缺少 visible 事件而停住，UI 一直显示 connecting
- 设计方案: 在 `runVisibilityAwareInterval` 中保留 hidden 时停止，仅新增 `focus` / `pageshow` 作为可见状态下的 resume 触发；`start()` guard 防止重复 tick/interval
- 受影响模块: `packages/web/src/lib/visibility-interval.ts` 及对应单测

✅ 1. Shared helper 的 `handle` guard 让 `visibilitychange`、`focus`、`pageshow` 多事件路径保持幂等，teardown 同时移除新增 listeners。 [R4 / `packages/web/src/lib/visibility-interval.ts:42`]
✅ 2. 新测试覆盖 missed `visibilitychange`、already-running refocus、hidden focus no-op、teardown；CI web lint/type/test 当前均为 green。 [`packages/web/src/lib/__tests__/visibility-interval.test.ts:65`]

**已执行**
- 已 approve

yuezengwu (approved):
Reviewed the shared visibility-aware polling helper and its onboarding/new-agent/new-connection call sites. The change preserves the hidden-page stop behavior, adds idempotent resume triggers for focus/pageshow, and the consumers still have their own cancellation/sequence guards.

Local verification:
- `pnpm --filter @first-tree/web test -- packages/web/src/lib/__tests__/visibility-interval.test.ts` (ran the web suite: 124 files / 979 tests passed)
- `pnpm exec biome check packages/web/src/lib/visibility-interval.ts packages/web/src/lib/__tests__/visibility-interval.test.ts`

No blockers found.

PR #1245, opened 2026-06-24
fix(web): bridge wheel scrolling over the expanded chat summary

Gandy2025:
## Problem

When a chat's task **Summary** is expanded, hovering the mouse over the summary panel and scrolling does nothing — the page reads as **"locked."** Reported with a screenshot of an expanded summary occupying ~1/3 of the viewport.

## Root cause

The expanded `ChatSummary` (`packages/web/src/pages/workspace/center/chat-summary.tsx`) is a pinned `shrink-0` sibling **above** the message scroll container (`scrollContainerRef`), and its only ancestor is the `overflow-hidden` center column. So a wheel gesture over the summary has **no scrollable target**:

- The inner markdown body (`maxHeight: min(46vh, 30rem); overflowY: auto`) only scrolls when its **own** content overflows.
- There is no scrollable ancestor above the panel.
- The message stream is a **cousin**, not an ancestor — wheel/scroll chaining can't reach it.

→ Mouse wheel over the expanded summary scrolls nothing. The taller the summary, the larger the dead zone. (Confirmed live on staging: walking up from the summary panel, `nearestScrollableAncestor === none`.)

## Fix

Add a **native, non-passive** `wheel` listener on the whole summary panel (header bar + padding + body) that drives scrolling itself instead of relying on a scrollable ancestor:

- While the markdown body still has room in the wheel's direction → scroll the **body** explicitly and `preventDefault()`. Explicit (not "defer to native") on purpose: a wheel over the header/padding sits **outside** the body's event path, so native scroll there has no target and would re-create the dead zone for that strip. Driving the body ourselves makes a wheel anywhere on the panel move it uniformly. `preventDefault` keeps a wheel directly over the body from being applied twice.
- At the body boundary (or when there's no body) → drive `scrollContainerRef.scrollTop`. Scrolling **down** over the summary also folds it via the existing sticky-collapse (it reacts to the stream's scroll), exactly as scrolling the stream directly does.

Native + non-passive is required because React routes `onWheel` through a passive root listener, where `preventDefault()` is a no-op.

The pinned/collapsible design is unchanged — this only restores scroll continuity.

## Verification

- **Real Chromium** (local build, real `WheelEvent` over the expanded panel): the message stream scrolls its full range and `event.defaultPrevented === true` (the native listener fires; preventDefault works).
- **Unit tests** (`chat-summary.test.ts`, 26 passing incl. 4 new wheel tests): wheel over the expanded summary drives the stream; a scrollable body is moved first (and is moved even when the wheel originates on the **header button**, not just inside the body — the R4 regression case); horizontal gestures are left to the browser.
- `pnpm --filter @first-tree/web typecheck` clean; `biome check` clean on changed files.

## Out of scope / honest note

This bridges **wheel** input (the reported case). Touch-drag over the expanded summary has the same root cause but a different mechanism, and is **not** covered here. The durable cure for both is architectural — make the summary a single scroll surface (a `position: sticky` collapsing header inside the stream, or an overlay on expand) — which removes the dead zone for any input without JS bridging. Tracked as a separate design decision; this PR is the immediate stopgap.

### DIFF
--- a/packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts
@@ -306,7 +306,7 @@ describe("ChatSummary", () => {
     container.remove();
   });
 
-  it("lets the summary body consume the wheel before driving the stream", async () => {
+  it("scrolls the summary body itself before driving the stream", async () => {
     localStorage.clear();
     const scrollEl = document.createElement("div");
     scrollEl.scrollTop = 0;
@@ -323,10 +323,41 @@ describe("ChatSummary", () => {
     Object.defineProperty(inner, "clientHeight", { value: 200, configurable: true });
     inner.scrollTop = 100;
 
-    // Wheel up: the body can still scroll up, so the stream must stay put.
+    // Wheel up: the body can still scroll up, so the body moves and the stream
+    // stays put.
     await act(async () => {
       panel.dispatchEvent(wheelEvent(-40));
     });
+    expect(inner.scrollTop).toBe(60);
+    expect(scrollEl.scrollTop).toBe(0);
+
+    await act(async () => root.unmount());
+    container.remove();
+  });
+
+  it("scrolls the body for a wheel over the summary header, not just the body", async () => {
+    localStorage.clear();
+    const scrollEl = document.createElement("div");
+    scrollEl.scrollTop = 0;
+    const { container, root } = await renderSummary(scrollEl, {
+      descriptionUpdatedAt: unreadVersionAt,
+      lastReadAt: readRecentlyAt,
+    });
+    // The header control sits OUTSIDE the markdown body's event path; a wheel
+    // there must still drive the body while it has room (regression: target-
+    // unaware deferral to native scroll left the header strip locked).
+    const header = container.querySelector<HTMLButtonElement>('button[aria-label="Collapse summary"]');
+    if (!header) throw new Error("summary header button missing");
+    const inner = container.querySelector<HTMLElement>('[style*="46vh"]');
+    if (!inner) throw new Error("summary scroll body missing");
+    Object.defineProperty(inner, "scrollHeight", { value: 1000, configurable: true });
+    Object.defineProperty(inner, "clientHeight", { value: 200, configurable: true });
+    inner.scrollTop = 0;
+
+    await act(async () => {
+      header.dispatchEvent(wheelEvent(120));
+    });
+    expect(inner.scrollTop).toBe(120);
     expect(scrollEl.scrollTop).toBe(0);
 
     await act(async () => root.unmount());
--- a/packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts
@@ -123,6 +123,16 @@ describe("ChatSummary", () => {
     };
   }
 
+  // A wheel event the summary's native (non-passive) listener can read. Built
+  // from a plain cancelable Event so the test does not depend on happy-dom's
+  // WheelEvent constructor; deltaX/deltaY are the only fields the handler reads.
+  function wheelEvent(deltaY: number, deltaX = 0): Event {
+    const e = new Event("wheel", { bubbles: true, cancelable: true });
+    Object.defineProperty(e, "deltaY", { value: deltaY });
+    Object.defineProperty(e, "deltaX", { value: deltaX });
+    return e;
+  }
+
   it("auto-expands an unread summary version on entry even when the chat was read recently", async () => {
     localStorage.clear();
     const scrollEl = document.createElement("div");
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1245

- Reason: the bridge still leaves a wheel dead zone over the expanded summary header when the markdown body itself is scrollable.

**Level**: C

**PR Summary**
- Problem: an expanded chat summary is a pinned sibling above the message stream, so wheel input over that area can have no scrollable target.
- Design: add a native non-passive wheel listener on `ChatSummary` that lets the inner markdown body scroll first, then bridges remaining vertical wheel input to the message stream.
- Affected modules: `packages/web/src/pages/workspace/center`

**Review Notes**
❌ 1. The "body can absorb this delta" guard is not target-aware, so wheel over the summary header/control can still be locked. [R4 / `packages/web/src/pages/workspace/center/chat-summary.tsx:315`]

Because the listener is attached to the panel, this branch returns for any wheel event whenever `inner` can scroll in that direction. If the cursor is over the expanded summary's `Summary` button/header area rather than inside the markdown body, native scrolling has no scrollable ancestor to act on: the body is a sibling/descendant outside the event target path, the panel itself is not scrollable, and the center column is still `overflow-hidden`. In that state neither the body nor the message stream moves, preserving part of the dead zone this PR is meant to remove.

Please make the "let native scroll run" branch conditional on the event target being inside the scrollable markdown body, or explicitly scroll `inner` for panel/header wheel events before falling through to the stream at the body boundary. Add a regression test that dispatches the wheel from the expanded summary button/header while `inner` is scrollable.

**Executed**
- Submitted `request_changes`.

Gandy2025:
Thanks — good catch on R4, confirmed and fixed in `626f8b96`.

You're right: the guard checked whether `inner` *could* scroll but not *where the cursor was*, so a wheel over the summary header/padding (outside the body's event path) deferred to native scroll, which had no scrollable ancestor there → that strip stayed locked while the body was scrollable.

Fix: drive `inner` **explicitly** when it has room in the wheel's direction (and `preventDefault` so a wheel directly over the body isn't applied twice), then fall through to the message stream at the body boundary. A wheel anywhere on the panel — header, padding, or body — now moves the body uniformly.

Added the requested regression test: `scrolls the body for a wheel over the summary header, not just the body` dispatches the wheel from the `Collapse summary` header button while `inner` is scrollable, asserting the body scrolls and the stream stays put.

yuezengwu (approved):
Re-reviewed after `626f8b96`.

The previous R4 issue is fixed: the wheel handler now explicitly scrolls the summary body while it has room, so wheel input from the header/padding/body no longer falls into a target-path dead zone. The new regression test dispatches from the summary header button with a scrollable body and covers that case.

Verified locally:
- `pnpm --filter @first-tree/web test -- chat-summary.test.ts` (web suite passed: 124 files / 978 tests)
- `pnpm exec biome check packages/web/src/pages/workspace/center/chat-summary.tsx packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts`

baixiaohang (approved):
## Review by codex-assistant - PR #1245

**Recommendation**: approve
- Reason: the follow-up fixes the panel/header dead zone by explicitly scrolling the summary body before bridging to the message stream.

**Level**: C

**PR Summary**
- Problem: an expanded chat summary is a pinned sibling above the message stream, so wheel input over that area can have no scrollable target.
- Design: add a native non-passive wheel listener on `ChatSummary`; while the markdown body has scroll room, scroll it explicitly from any panel target, then drive the message stream at the body boundary.
- Affected modules: `packages/web/src/pages/workspace/center`

**Review Notes**
✅ 1. Previous R4 blocker is resolved: a wheel over the summary header/padding now moves the body while it can scroll, and the requested regression test dispatches from the header button.
✅ 2. The change remains scoped to Workspace center presentation/interaction and preserves the existing pinned/collapsible summary model.

**Executed**
- Submitted `approve`.

baixiaohang:
Non-blocking PR-body accuracy note: the edited Verification section now says the unit tests cover that "horizontal gestures are left to the browser", but at `626f8b96` I only see the three new wheel tests covering stream bridging, body-first scrolling, and the header-button regression. I don't see a horizontal / `deltaX` assertion.

Approval stands, but please either drop that phrase from the Verification text or add the test so the PR body matches the actual coverage.

baixiaohang (approved):
## Review by codex-assistant - PR #1245

**Recommendation**: approve
- Reason: the latest commit only adds the missing horizontal-dominant wheel regression test; the previously approved behavior remains unchanged.

**Level**: C

**PR Summary**
- Problem: an expanded chat summary can otherwise create a wheel dead zone because it is pinned outside the message stream's scroll container.
- Design: bridge vertical wheel input from the summary panel to the summary body first, then to the message stream; leave horizontal-dominant gestures to the browser.
- Affected modules: `packages/web/src/pages/workspace/center`

**Review Notes**
✅ 1. The new `9e4fc5c7d23d` test covers the previously noted PR-body accuracy gap: a horizontal-dominant wheel does not scroll the summary body or stream and is not consumed.
✅ 2. No production-code change was added after the already-approved `626f8b96` behavior fix.

**Executed**
- Submitted `approve`.

Gandy2025:
Good accuracy catch — added the test rather than dropping the phrase (`9e4fc5c7`). `leaves a horizontal-dominant wheel to the browser` dispatches a deltaX-dominant wheel over the panel with a scrollable body and asserts neither the body nor the stream moves and the event is not consumed. The Verification claim now matches actual coverage (26 tests).

baixiaohang:
One last non-blocking body cleanup: the current PR body still says `chat-summary.test.ts`, 25 passing incl. 3 new, but after `9e4fc5c7` the file has 26 `it(...)` cases and this PR adds 4 wheel tests. Approval stands; please update that count so the Verification section matches the branch.

PR #1253, opened 2026-06-24
fix(client): defer idle-suspend while a provider has a live subprocess

yuezengwu:
## Problem

A session whose provider still has a live background subprocess — e.g. a `Bash run_in_background` watcher polling a CI run — was idle-suspended after `idle_timeout` (default 5 min). `lastActivity` is only bumped on provider communication, so a purely-background wait never refreshes it. Suspend then closes the provider stream, so even though the watcher keeps running, its "background task complete → re-invoke the agent" wake-up is lost. The agent goes silent until a human pings it.

Root cause was traced from an exported session transcript (agent waited ~40 min on an iOS build, went silent, user had to ping). Verified locally: the provider is a `claude` process spawned by the Claude Agent SDK as a **direct child of the daemon**, stamped with a per-session `FIRST_TREE_CHAT_ID` env var; both the provider pid and whether it currently has a descendant are reliably observable via `ps`.

## Change (client-only)

- **New `runtime/process-tree-probe.ts`** — a `ps`-backed `SubprocessProbe` that maps each provider pid (daemon's direct `claude` children, attributed by `FIRST_TREE_CHAT_ID`) to whether it currently has a live descendant. Refresh is async + cached, so the synchronous `evictIdle` path never blocks on a process scan; any `ps` failure degrades to "no live work" (i.e. prior behaviour).
- **`session-manager.ts`**
  - `evictIdle`: a live subprocess defers idle-suspend, treated like in-flight processing work — **still bounded by the existing `idle_timeout + working_grace_seconds` hard cap**, so a forgotten/stuck watcher is still reclaimed (no unbounded slot hold).
  - `acquireActiveSlot` (concurrency yield) and `evictIfNeeded` (max_sessions): deprioritize sessions with a live subprocess as eviction victims — sacrificed only as a last resort.
  - `shutdown` stops the probe.
- **Config `session.defer_suspend_on_subprocess`** (default **on**) in both `@first-tree/shared` `agentConfigSchema` and the runtime `config.ts`; `agent-slot` wires the probe per the flag.

No prompt / brief / server changes. With no probe injected, behaviour is identical to before.

## Testing

- New `process-tree-probe.test.ts` (6): pure parsing/attribution/descendant detection + probe class + `ps`-failure fallback.
- New `session-manager.test.ts` cases (3): defer-then-suspend, **hard cap still reclaims** with a live subprocess, concurrency victim prefers the no-subprocess session.
- Full suites green: **client 1165 / 109 files**, **shared 479**. `pnpm check` (Biome, 1345 files) and `pnpm typecheck` clean.
- **Real-daemon QA: PASS, no blocking defects.** With `idle_timeout=5s`, a live watcher keeps the session `active` past idle and the post-completion task-notification wakes the agent; with `idle_timeout=5s`/`working_grace=30s`, the session defers then flips to `suspended` at the hard cap (confirms no unbounded keep-alive).

## Follow-up

A Context Tree decision-record PR (why a local process-tree probe over a server-side decouple / bounded-signal approach) will be opened separately and cross-linked.

### DIFF
--- a/packages/client/src/__tests__/process-tree-probe.test.ts
+++ b/packages/client/src/__tests__/process-tree-probe.test.ts
@@ -40,10 +40,18 @@ describe("process-tree-probe pure helpers", () => {
     expect(hasDescendant(999, idx)).toBe(false);
   });
 
-  it("extracts FIRST_TREE_CHAT_ID from a ps -E command line", () => {
+  it("extracts FIRST_TREE_CHAT_ID from a Darwin `ps -E` (space-separated) line", () => {
     const line = "FIRST_TREE_HOME=/x FIRST_TREE_CHAT_ID=f93566d9-00c8 FIRST_TREE_AGENT_ID=019e /bin/claude";
     expect(extractChatId(line)).toBe("f93566d9-00c8");
     expect(extractChatId("no marker here")).toBeNull();
+  });
+
+  it("extracts FIRST_TREE_CHAT_ID from Linux `/proc/<pid>/environ` (NUL-separated), stopping at the NUL", () => {
+    const environ = ["FIRST_TREE_HOME=/x", "FIRST_TREE_CHAT_ID=f93566d9-00c8", "FIRST_TREE_AGENT_ID=019e", ""].join(
+      "\0",
+    );
+    // The value must not bleed into the next NUL-separated entry.
+    expect(extractChatId(environ)).toBe("f93566d9-00c8");
   });
 });
 
@@ -71,6 +79,21 @@ describe("PsSubprocessProbe", () => {
     expect(probe.hasLiveSubprocess("chat-A")).toBe(true);
     expect(probe.hasLiveSubprocess("chat-B")).toBe(false);
     expect(probe.hasLiveSubprocess("chat-unknown")).toBe(false);
+    probe.stop();
+  });
+
+  it("attributes providers from NUL-separated env (Linux /proc form)", async () => {
+    const probe = new PsSubprocessProbe({
+      log: silentLogger(),
+      daemonPid,
+      intervalMs: 1_000_000,
+      runProcessSnapshot: async () => snapshot,
+      runEnvForPid: async (pid) =>
+        ["FIRST_TREE_HOME=/x", `FIRST_TREE_CHAT_ID=${pid === 100 ? "chat-A" : "chat-B"}`, ""].join("\0"),
+    });
+    await probe.refresh();
+    expect(probe.hasLiveSubprocess("chat-A")).toBe(true);
+    expect(probe.hasLiveSubprocess("chat-B")).toBe(false);
     probe.stop();
   });
 
--- a/packages/client/src/__tests__/process-tree-probe.test.ts
+++ b/packages/client/src/__tests__/process-tree-probe.test.ts
@@ -1,0 +1,91 @@
+import { describe, expect, it } from "vitest";
+import {
+  buildChildrenIndex,
+  extractChatId,
+  findProviderPids,
+  hasDescendant,
+  PsSubprocessProbe,
+  parseProcessRows,
+} from "../runtime/process-tree-probe.js";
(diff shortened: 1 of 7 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1253

**建议**: request_changes
- 理由: The new probe cannot read provider env on Linux/procps, so the main fix silently degrades to the old idle-suspend behavior on Linux daemon hosts.

**分级**: A
- Path 初判: contains `packages/client/**` -> A
- 语义升级: none

**PR 摘要**
- 解决问题: sessions with live provider background subprocesses can be idle-suspended, closing the provider stream and losing the background completion wake-up.
- 设计方案: add a cached `ps`-backed subprocess probe, consult it from idle suspend / concurrency / max-session victim selection, and gate it with `session.defer_suspend_on_subprocess`.
- 受影响模块: `packages/client/src/runtime/*`, client session-manager tests, and shared agent config schema.

**Review 意见**
❌ 1. `defaultEnvForPid` uses `ps -Eww -p <pid> -o command=`, but Linux/procps rejects `-Eww` as an unsupported SysV option. Because `doRefresh()` catches that failure and clears `chatIdsWithLiveWork`, every `hasLiveSubprocess()` lookup returns false on Linux, so Linux daemons still suspend sessions with live background watchers. Please make env collection platform-aware (for example `/proc/<pid>/environ` or `ps eww ...` on Linux, and the BSD form on Darwin) and add coverage for the default env-command/platform path. [R5 / `packages/client/src/runtime/process-tree-probe.ts:98`]

✅ 2. The `SessionManager` side keeps the existing `idle_timeout + working_grace_seconds` hard cap and treats subprocess protection as a victim-selection preference rather than an unbounded slot hold.

**已执行**
- 已提交 request_changes review.
- 未运行代码检查或单测（PR Review 模式禁止运行）。

yuezengwu:
Thanks @codex-assistant — finding #1 is correct and serious (it would have silently disabled the fix on Linux daemon hosts, the common production target). Fixed in 92b3f282:

- `defaultEnvForPid` is now platform-aware: reads `/proc/<pid>/environ` on Linux (procps has no `-E`), keeps `ps -Eww` on Darwin/BSD.
- Hardened `extractChatId` to terminate the value at whitespace **or** NUL, so it parses both the space-separated (`ps`) and NUL-separated (`/proc/environ`) forms without bleeding into the next entry.
- Added coverage: an `extractChatId` NUL-separated case and a probe-level test attributing providers from NUL-separated env.

Full client suite (1167) + shared (479) green; `pnpm check` + `typecheck` clean.

(Finding #2 — hard cap preserved, subprocess protection is victim-selection preference only — matches intent, thanks for confirming.)

baixiaohang (approved):
## Review by codex-assistant - PR #1253

**建议**: approve
- 理由: The Linux/procps env lookup issue from the previous review is fixed, and the remaining session lifecycle changes preserve the hard cap and recovery boundaries.

**分级**: A
- Path 初判: contains `packages/client/**` -> A
- 语义升级: none

**PR 摘要**
- 解决问题: sessions with live provider background subprocesses could be idle-suspended, closing the provider stream and losing the background completion wake-up.
- 设计方案: add a cached process-tree subprocess probe, consult it from idle suspend / concurrency / max-session victim selection, and gate it with `session.defer_suspend_on_subprocess`.
- 受影响模块: `packages/client/src/runtime/*`, client session-manager/probe tests, and shared agent config schema.

**Review 意见**
✅ 1. Previous blocker is resolved: Linux now reads `/proc/<pid>/environ`, Darwin/BSD keeps `ps -Eww`, and `extractChatId` handles both whitespace-separated and NUL-separated env dumps.
✅ 2. `SessionManager` still bounds deferral by `idle_timeout + working_grace_seconds`, and subprocess protection is a victim-selection preference rather than an unbounded slot hold.

**已执行**
- 已按 baixiaohang 授权提交 approve review.
- 未运行代码检查或单测（PR Review 模式禁止运行）。

PR #1256, opened 2026-06-24
feat: support @mention and image attachments in the ask answer card

yuezengwu:
## What & why

The **AskTakeover** card (the pop-up answer surface for a blocking `chat ask`) used a bare `<textarea>`, so answering a question could **neither @mention a member nor attach an image** — unlike the chat composer at the bottom of every chat. This unifies the two: the free-text answer surface now mirrors the composer with `@mention` autocomplete + inline highlight and image attachments (paste / drag-drop / file-picker), on **both** the options "Other" box and the pure free-text box.

## Key finding: server needs no change

An ask answer must carry `metadata.resolves` to clear the blocking question. The server's resolve gate authorizes purely off `senderId === target` (`services/message.ts`) and `metadata.resolves` is a **format-agnostic** field — so an image answer can go out as a `format="file"` batch that *still resolves* the question, exactly like a text answer. The only client-side gap was `sendFileMessageBatch` not threading `resolves`; that's the one API tweak.

## Changes (all `packages/web`)

- **`api/chats.ts`** — `sendFileMessageBatch` opts now carry `resolves` (mirrors `sendChatMessage`).
- **`components/chat/ask-takeover.tsx`** — `onReply` contract changes from `(content: string)` to `({ content, mentions, images })`; new `mentionCandidates` + `error` props; the free-text surface gets mention overlay/popover + a staged-image thumbnail tray + an `@`/attach toolbar; an attached image alone can resolve.
- **`pages/workspace/center/chat-view.tsx`** — new `submitAskAnswer` owns upload + send + resolve. **Non-optimistic by design**: the card stays mounted until the server's resolving reply lands, so a failed send never unmounts the card or drops the typed answer / staged images. Routes the resolving reply to the asker **plus** anyone the free text @mentioned.
- **`pages/request-dock-preview.tsx`** — dev preview updated to the new contract.

## Tests

- `ask-takeover-dom`: answer-contract (object form), `@<name>` → agentId resolution, popover render, image staging + image-only resolve, oversized-image rejection, host-error surfacing.
- `chat-view-dom`: free-text question resolved via text (`sendChatMessage`) and via an attached image (`sendFileMessageBatch` with `resolves`).

`pnpm --filter @first-tree/web typecheck` ✓ · `biome check` ✓ · web suite **1002 passed**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1266, opened 2026-06-24
fix: restore visible text in the chat ask answer input

Gandy2025:
## What

Typed text in the `chat ask` answer box (`AskTakeover`) was invisible — white-on-white on the light theme, and equally hidden on dark. This restores it.

## Root cause (regression from #1256)

#1256 added a mention-highlight mirror overlay to the answer box: the real `<textarea>` is set `color: transparent` and `MentionHighlightOverlay` (behind it) paints the visible glyphs in `var(--fg)`. But the textarea kept an opaque `background: var(--bg)` **and** `zIndex: 1`, so it sat *on top of* the overlay and hid the painted glyphs; the textarea's own text is transparent → nothing visible.

This is the one piece the working chat composer (`chat-view.tsx`) gets right and this box missed: there the textarea is `background: transparent`, so the overlay shows through.

## Fix

Move the visible chrome (border + fill + radius) from the textarea onto the field **wrapper**, and make the textarea `background: transparent` + `border: none` — mirroring `chat-view.tsx`. Both answer inputs route through `renderAnswerInput`, so this covers the sole free-text box and the "Other…" box beside options. It also removes a pre-existing ~1px glyph drift: the overlay's `inset:0` and the textarea text box now share the same origin.

## Test / verification

- Added a regression test in `ask-takeover-dom.test.tsx`: the answer textarea must stay `background: transparent` / `color: transparent`, the mirror overlay sibling must exist, and the opaque fill must live on the wrapper. Verified red on the buggy code, green after the fix.
- `pnpm --filter @first-tree/web test` → 1006/1006 pass. `biome check` clean on changed files. `typecheck` + `lint:tokens` clean.
- Reproduced + verified visually on the DEV `/preview/request-dock` gallery (free-text + "Other"), light theme: **before** = empty white box (typed text invisible), **after** = text clearly rendered.

## How to reproduce

`pnpm --filter @first-tree/web dev` → open `/preview/request-dock` → type into the "free text" box. On `main` the text is invisible; with this branch it renders.

### DIFF
--- a/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
+++ b/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
@@ -425,4 +425,23 @@ describe("AskTakeover", () => {
     expect(onSkip).not.toHaveBeenCalled();
     ta.removeEventListener("keydown", consume);
   });
+
+  it("free-text answer surface stays transparent so the mention overlay shows the typed text", async () => {
+    // Regression guard (PR 1256): the answer textarea is painted transparent and
+    // a mirror overlay behind it draws the glyphs. If the textarea keeps an
+    // opaque background it sits on top of the overlay and the typed text turns
+    // invisible — the white-on-white bug on the light theme. The visible chrome
+    // (border + fill) must live on the wrapper, not the textarea.
+    const c = await renderDom(
+      <AskTakeover body="# Concerns?" payload={{ multiSelect: false }} onReply={() => {}} onSkip={() => {}} />,
+    );
+    const ta = freeTextBox(c);
+    if (!ta) throw new Error("free-text input missing");
+    expect(ta.style.background).toBe("transparent");
+    expect(ta.style.color).toBe("transparent");
+    // The mirror overlay that actually paints the glyphs is a sibling.
+    expect(ta.parentElement?.querySelector("[aria-hidden]")).not.toBeNull();
+    // The opaque fill now lives on the wrapper so the field still reads as a box.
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the AskTakeover text visibility fix. The wrapper/textarea layering matches the composer overlay contract, covers both free-text and Other inputs through renderAnswerInput, and the regression test pins the important invariants. Local verification: pnpm --filter @first-tree/web test -- ask-takeover-dom.test.tsx (ran the web suite, 1006/1006 passed), pnpm exec biome check on the changed files, and pnpm --filter @first-tree/web typecheck.

baixiaohang (approved):
## Review by codex-assistant - PR #1266

**Recommendation**: approve
- Reason: The PR restores the existing mention-overlay layering contract in `AskTakeover` and adds a focused regression guard.

**Level**: C
- Semantic lift: none

**PR Summary**
- Problem: typed text in the chat ask answer textarea was hidden because a transparent textarea still painted an opaque background above the mention mirror overlay.
- Design: moves border/fill/radius to the answer-field wrapper and keeps the textarea transparent/borderless, covering both the free-text ask box and the options "Other" box through `renderAnswerInput`.
- Affected modules: `packages/web/src/components/chat`

**Review Notes**
✅ 1. The change matches the existing composer overlay contract: the textarea remains above the overlay for caret/input handling, but no longer paints over the mirrored glyphs.
✅ 2. The regression test pins the DOM layering contract for `AskTakeover`: transparent textarea, overlay sibling, and wrapper-owned fill.

**Executed**
- Approved.

PR #1262, opened 2026-06-24
fix: support Node 18.14 CLI installs

baixiaohang:
## Summary
- lower the published CLI runtime floor to Node.js >=18.14.1 while keeping the latest Claude agent SDK range
- route npm bins through a CommonJS wrapper that preflights unsupported Node versions before loading the ESM bundle
- pin onboarding install commands to exact prod/staging package versions and fail closed when the server command version does not match the deployment channel
- keep prod/staging publish bin rewrites pointed at the wrapper and add packed CLI smoke coverage for Node 18.14.1 with strict engine installs

## Companion PR
- Context Tree: https://github.com/agent-team-foundation/first-tree-context/pull/602

## Validation
- `pnpm check`
- `pnpm typecheck`
- `pnpm test --concurrency=2`
- `pnpm --filter @first-tree/server test -- me-multi-org` (server suite: 170 files / 1563 tests)
- `pnpm --filter first-tree-dev test -- bin-wrapper package-metadata prompt-core doctor-core`
- packed tarball strict smoke on Node 18.14.1 + npm 9.3.1 (`npm install --engine-strict`, `first-tree-dev --version`, `first-tree-dev --help`)
- `git diff --check`

PR #1271, opened 2026-06-25
fix: fallback agents dir watcher on Node 18

baixiaohang:
## Summary
- fall back from recursive agents-dir watching to non-recursive watchers when the current Node/platform does not support recursive `fs.watch`
- keep watcher setup/runtime failures as warnings so the daemon does not exit and systemd does not restart-loop
- refresh fallback child-directory watchers after scans so hot-added agent directories are still observed when `agent.yaml` is written after the directory event
- add regression coverage for recursive-watch setup failures and delayed `agent.yaml` writes under the fallback path

## Companion PR
- Context Tree: https://github.com/agent-team-foundation/first-tree-context/pull/605

## Validation
- `pnpm typecheck`
- `pnpm --dir apps/cli exec vitest run src/__tests__/client-runtime-context-tree.test.ts`
- Node 18.19.0 Linux smoke via `npx -y node@18.19.0`: import `apps/cli/dist/index.mjs`, call `ClientRuntime.watchAgentsDir`, observe fallback warning and successful return
- `pnpm check` (passes with existing warnings/info)
- `git diff --check`
- `pnpm test --concurrency=2`

### DIFF
--- a/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
+++ b/apps/cli/src/__tests__/client-runtime-context-tree.test.ts
@@ -151,6 +151,12 @@ vi.mock("../core/version.js", () => ({
 describe("ClientRuntime context-tree wiring", () => {
   const originalHome = process.env.FIRST_TREE_HOME;
   let home: string;
+  type WatchCallback = (eventType: string, filename: string | Buffer | null) => void;
+  type WatchRegistration = {
+    path: string;
+    listener: WatchCallback | null;
+    close: ReturnType<typeof vi.fn>;
+  };
 
   beforeEach(() => {
     vi.resetModules();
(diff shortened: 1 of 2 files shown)


baixiaohang:
Closing for now because PR #1262 is being reverted. We can reopen or rework this watcher hardening after the Node runtime-floor direction is settled.

yuezengwu (approved):
Reviewed the goal and implementation: this keeps the daemon from exiting when recursive `fs.watch` is unavailable or watcher setup/runtime errors happen, falls back to one-level non-recursive watchers for `agents/<name>/agent.yaml`, and refreshes fallback child watchers after scans so delayed config writes are still observed.

No blockers found. The change is scoped to `ClientRuntime.watchAgentsDir` plus regression tests, does not touch schema/database/core data structures, and aligns with the runtime/daemon constraint that recoverable local watcher failures should degrade to warnings rather than restart-loop the supervised process.

I could not rerun the Vitest target locally in this fresh review worktree because dependencies were not installed (`vitest` not found), but I inspected the diff and relevant call paths against the PR head and base.

PR #1251, opened 2026-06-24
[codex] Collapse build tree entry points

Gandy2025:
## Summary
- Remove the standalone `/build-tree` recovery page and Settings setup card.
- Move the build-tree entry into the Context tab unavailable state, reusing Cloud bootstrap plus the `tree` kickoff chat.
- Redesign Settings -> Context tree as binding/configuration state instead of a build wizard, including member read-only behavior and Context Reviewer placement.
- Clean up obsolete recovery props, copy, tests, and preview coverage.

## Validation
- `pnpm check`
- `pnpm --filter @first-tree/web typecheck`
- `pnpm --filter @first-tree/web test` (124 files / 970 tests)

## Paired Context Tree PR
- https://github.com/agent-team-foundation/first-tree-context/pull/598

PR #1272, opened 2026-06-25
feat(web): connect GitHub and pick a repo inline on the Context build entry

Gandy2025:
## What & why

Follow-up to #1251. When a team has no Context Tree yet, the Context tab build entry's **"Connect your code"** button sent the user to `/settings/resources` to type a repo URL by hand. But once the team's GitHub App is installed, First Tree already knows which repos it grants — so that bounce was busywork. (And `/settings/github` alone wouldn't fix it either: the build gate keys on having a repo **resource**, which installing the App doesn't create.)

Now the connect + repo pick happen **inline on the card**:

- **No GitHub App** → install it in a popup (the same install flow onboarding uses), card polls and advances on its own once it lands.
- **App installed** → list the repos the installation grants and let the user pick (`listOrgGithubRepos` + the onboarding `RepoTokenPicker`), then choose the builder agent and build.

No more leaving the Context tab to wire up a repo.

## Implementation notes

- **Reuses the onboarding connect-code building blocks** (`RepoTokenPicker`, `FlowHint`/`StatusRow`, `COPY.connectCode.*`, `getGithubAppInstallation`/`getGithubAppInstallUrl`/`listOrgGithubRepos`, `ensureKickoffRepos`/`startTreeSetupKickoff`) — no new visual components. The only new code is the orchestration that assembles them on the card (`ConnectAndPickRepos`) plus a factored-out `BuildAgentControls`.
- The install-popup logic is a deliberate **local copy** of `StepConnectCode`'s (synchronous `window.open`, install-attempt marker, `oauth_state_nonce`-safe re-mint), kept separate on purpose so onboarding's critical first-connect path stays untouched. A future refactor could extract a shared hook.
- The **agent check now runs before the no-repo check**, so the inline pick always has an agent to hand the build to. A bound-tree recovery still builds straight from its binding (unchanged).

## Popup-blocked redirect (server)

When the install popup is blocked, the full-page redirect must return to the card to finish the inline pick. The post-install `next` is server-allowlisted (`ALLOWED_POST_INSTALL_NEXT`) — anything off it is silently rewritten to `/settings/github`. So the web passes the allowlisted `/context`, and `/context` is added to the allowlist, **replacing the now-removed `/build-tree`** entry (its recovery role moved onto the Context tab in #1251). Without this, a popup-blocked install would land the user back on Settings — the exact bounce this PR removes. (Caught by an independent codex review.)

## Testing

- `biome` + `tsc` + web `lint:tokens` clean; root `pnpm check` + `pnpm typecheck` (12/12 packages) green.
- Web: 1008 tests pass, incl. new cases — no-repo → inline install CTA (not a Settings bounce), connected → inline repo list + pick → build CTA, and popup-blocked → allowlisted `/context`.
- Server: `post-install-next` / `install-url` / `github-app-callback` tests pass with the `/context` allowlist change.
- Verified each state visually against the real components via the DEV `/preview/context-tree` gallery.

## Review

Two independent reviews before opening: codex (external) flagged the popup-blocked redirect issue above — fixed; a Claude review returned SHIP with two hardening items (mock completeness + inline pick→build coverage) folded in.

### DIFF
--- a/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
+++ b/packages/server/src/api/orgs/__tests__/post-install-next.test.ts
@@ -11,10 +11,10 @@ describe("resolvePostInstallNext", () => {
     expect(resolvePostInstallNext("/onboarding")).toBe("/onboarding");
     expect(resolvePostInstallNext("/onboarding/connected")).toBe("/onboarding/connected");
     expect(resolvePostInstallNext("/settings/github")).toBe("/settings/github");
-    // The build-tree recovery surface passes itself as `next` when the install
-    // popup is blocked — rewriting it to Settings would strand the recovery
-    // admin outside the flow they were in.
-    expect(resolvePostInstallNext("/build-tree")).toBe("/build-tree");
+    // The Context tab build entry passes itself as `next` when the install popup
+    // is blocked — rewriting it to Settings would bounce the admin out of the
+    // inline build/repo-pick flow they were in.
+    expect(resolvePostInstallNext("/context")).toBe("/context");
   });
 
   it("falls back to Settings for an absent value", () => {
@@ -31,7 +31,7 @@ describe("resolvePostInstallNext", () => {
 
   it("only allows the known internal destinations", () => {
     expect([...ALLOWED_POST_INSTALL_NEXT].sort()).toEqual([
-      "/build-tree",
+      "/context",
       "/onboarding",
       "/onboarding/connected",
       "/settings/github",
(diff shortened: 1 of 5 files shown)


Gandy2025:
Closing — this requirement is being handed off to a teammate (gandy-coder), who will take it over from here. This PR was opened in the few seconds before the hand-off message landed. The branch `feat/context-build-inline-repo-pick` and its commit are preserved (closing the PR does not delete the branch); reopen with `gh pr reopen 1272` or rebuild from the branch as needed.

PR #1278, opened 2026-06-25
[codex] Share Context Reviewer visibility across roles

Gandy2025:
## Summary

- Treat Context Reviewer as an organization-scoped setting instead of a caller-owned agent setting.
- Let admins configure any active non-human agent in the organization as the reviewer.
- Let members read the current Context Reviewer status and configured reviewer name without exposing the admin-only full agent roster.

## Root Cause

The setting was stored at the organization level, but the save path and UI treated the reviewer as if it had to belong to the current admin. That made other admins see an empty selector/warning for a valid reviewer configured by someone else. Members also had no read-only projection for the feature state.

## Impact

Admins now see and edit the same team-level reviewer configuration. Members see a read-only On/Off state and reviewer identity, but still cannot toggle, select, PUT, or DELETE the feature setting.

## Context Tree

Companion tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/609

## Validation

- `pnpm --filter @first-tree/shared test src/__tests__/org-settings-schema.test.ts`
- `pnpm --filter @first-tree/server test src/__tests__/org-settings.test.ts`
- `pnpm --filter @first-tree/web test src/pages/__tests__/settings-panels-dom.test.tsx`
- `pnpm typecheck`
- `pnpm exec biome check packages/shared/src/schemas/org-settings.ts packages/shared/src/__tests__/org-settings-schema.test.ts packages/server/src/services/org-settings.ts packages/server/src/__tests__/org-settings.test.ts packages/web/src/pages/context-tree-settings-panel.tsx packages/web/src/pages/__tests__/settings-panels-dom.test.tsx`

### DIFF
--- a/packages/web/src/pages/context-tree-settings-panel.tsx
+++ b/packages/web/src/pages/context-tree-settings-panel.tsx
@@ -16,6 +16,7 @@ import { Select } from "../components/ui/select.js";
 import { SettingsField, SettingsSaveButton } from "../components/ui/settings-field.js";
 import { Switch } from "../components/ui/switch.js";
 import { titleWithSemantics, useJustSaved } from "./agent-detail/save-semantics.js";
+import { fetchAllAgents } from "./team/index.js";
 
 /**
  * Settings → Context tree. Per-org Context Tree **configuration**: which repo /
@@ -278,12 +279,12 @@ function ContextReviewerSection({ hasBinding, isAdmin }: { hasBinding: boolean; 
 
   const managedAgentsQuery = useQuery({
     queryKey: ["context-reviewer", "org-agents", organizationId],
-    queryFn: () => listAllAgents({ limit: 100 }),
+    queryFn: () => fetchAllAgents((params) => listAllAgents(params)),
     enabled: isAdmin && !!organizationId && switchOn,
   });
 
   const reviewerCandidates = useMemo(() => {
-    return (managedAgentsQuery.data?.items ?? [])
+    return (managedAgentsQuery.data ?? [])
       .filter((agent) => agent.organizationId === organizationId && agent.type !== "human" && agent.status === "active")
       .sort((a, b) => {
         const byLabel = agentLabel(a).localeCompare(agentLabel(b), undefined, { sensitivity: "base" });
@@ -405,7 +406,8 @@ function ContextReviewerSection({ hasBinding, isAdmin }: { hasBinding: boolean; 
                 ) : null}
                 {reviewerMissing ? (
                   <div className="text-label" style={{ color: "var(--fg-3)" }}>
-                    Current reviewer is not your active agent. Choose one of your agents, or turn Context Reviewer off.
+                    Current reviewer is not an active organization agent. Choose another agent, or turn Context Reviewer
+                    off.
                   </div>
                 ) : null}
                 {managedAgentsQuery.error ? (
@@ -447,7 +449,7 @@ function ContextReviewerReadOnly({
     ? contextReviewer.reviewerAgent.displayName.trim() ||
       contextReviewer.reviewerAgent.name?.trim() ||
       contextReviewer.reviewerAgent.uuid
-    : contextReviewer.agentUuid;
+    : null;
 
   return (
     <div className="flex flex-col" style={{ gap: "var(--sp-2)" }}>
--- a/packages/web/src/pages/context-tree-settings-panel.tsx
+++ b/packages/web/src/pages/context-tree-settings-panel.tsx
@@ -2,7 +2,7 @@ import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import { ArrowRight } from "lucide-react";
 import { type FormEvent, useEffect, useId, useMemo, useState } from "react";
 import { useNavigate } from "react-router";
-import { listManagedAgents, type ManagedAgent } from "../api/agents.js";
+import { listAllAgents, type ManagedAgent } from "../api/agents.js";
 import {
   getContextTreeFeaturesSetting,
   getContextTreeSetting,
@@ -144,7 +144,7 @@ export function ContextTreeSettingsPanel() {
         </div>
       </Section>
 
-      {isAdmin ? <ContextReviewerSection hasBinding={hasBinding} /> : null}
+      <ContextReviewerSection hasBinding={hasBinding} isAdmin={isAdmin} />
     </div>
   );
 }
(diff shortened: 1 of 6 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1278

- Reason: the admin picker still loses valid org-level reviewers outside the first `/agents/all` page, so the shared-visibility fix is incomplete for larger teams.

**Level**: B-high
- Path initial: touches `packages/server/src/services/org-settings.ts` -> B-high
- Semantic lift: none

**PR Summary**
- Problem solved: Context Reviewer is org-scoped, but the previous UI/save path treated it as caller-owned and hid reviewers managed by other admins; members also lacked a read-only status projection.
- Design: server validation now accepts any active non-human agent in the org, output includes a configured reviewer summary, `context_tree_features` is member-readable, and the settings panel renders admin edit controls or member read-only status.
- Affected modules: shared org-settings schema, server org-settings service/API tests, and the web Context Tree settings panel.

**Review Comments**
❌ 1. The Context Reviewer picker only reads the first 100 admin-visible agents. [R5 / `packages/web/src/pages/context-tree-settings-panel.tsx:281`]

`listAllAgents({ limit: 100 })` reads only the first page, while the server caps `limit` at 100 and paginates `/agents/all` by `createdAt` (`packages/shared/src/schemas/common.ts:4`, `packages/server/src/services/agent.ts:758`). In orgs with more than 100 agents, a valid active non-human reviewer can be configured but absent from this page, so `selectedIsCandidate` becomes false and the UI falls back to the missing-reviewer state instead of showing the configured org reviewer. The same issue appears when the first page contains only humans/suspended agents while eligible reviewers exist on later pages. That leaves the root visibility bug unresolved for larger teams. The picker should either page through `/agents/all` until exhausted for this admin-only config surface, or use an API/query shape that can include the saved reviewer and search eligible reviewers beyond page one.

✅ 2. Server write validation now correctly accepts active non-human agents from the current organization rather than only agents managed by the caller. [`packages/server/src/services/org-settings.ts:297`]

✅ 3. Member GET output is scoped to the configured reviewer summary and does not expose the admin-only full agent roster. [`packages/server/src/services/org-settings.ts:253`]

**Performed**
- Submitted a request_changes review.

packages/web/src/pages/context-tree-settings-panel.tsx
```diff
@@ -277,13 +277,13 @@ function ContextReviewerSection({ hasBinding }: { hasBinding: boolean }) {
   const switchOn = serverEnabled || setupOpen;
 
   const managedAgentsQuery = useQuery({
-    queryKey: ["context-reviewer", "managed-agents", organizationId],
-    queryFn: listManagedAgents,
-    enabled: !!organizationId && switchOn,
+    queryKey: ["context-reviewer", "org-agents", organizationId],
+    queryFn: () => listAllAgents({ limit: 100 }),
```
yuezengwu: This only loads the first 100 agents from the admin roster. The new contract is that admins can configure any active non-human agent in the organization, but an agent on a later page will not appear in the picker and an already-configured reviewer beyond page 1 will be treated as missing. The Team page already uses `fetchAllAgents` to page through `/agents/all`; this picker should either do the same or use a server-side filtered/searchable endpoint so the UI actually covers the whole org roster.

packages/web/src/pages/context-tree-settings-panel.tsx
```diff
@@ -427,3 +433,42 @@ function ContextReviewerSection({ hasBinding }: { hasBinding: boolean }) {
...
+function ContextReviewerReadOnly({
+  contextReviewer,
+}: {
+  contextReviewer: {
+    enabled: boolean;
+    agentUuid: string | null;
+    reviewerAgent?: { uuid: string; name: string | null; displayName: string } | null;
+  };
+}) {
+  const reviewerLabel = contextReviewer.reviewerAgent
```
yuezengwu, line 448: When the configured reviewer has been deleted or otherwise cannot be summarized, the server returns `reviewerAgent: null` while `agentUuid` remains set. This fallback therefore shows the raw UUID to members, and the `"Configured reviewer is no longer available."` text below is effectively unreachable for enabled settings. Since the member projection is supposed to show reviewer identity/name without exposing roster details, please render the unavailable copy when `reviewerAgent` is null instead of falling back to `agentUuid`.

yuezengwu (changes requested):
Reviewed as: Context Reviewer is now an organization-scoped team setting, with admins editing any active non-human org agent and members reading only the current feature status plus configured reviewer summary.

Requesting changes for two UI correctness issues. The shared schema/server direction matches the tree constraint overall, and there is no DB migration, but the admin picker and member read-only fallback still miss edge cases.

packages/web/src/pages/context-tree-settings-panel.tsx
Gandy2025: Fixed in `912e0fe`: the Context Reviewer admin picker now uses the existing `fetchAllAgents` pagination helper over `/agents/all`, so it evaluates the full admin roster before filtering active non-human org agents. Added a regression test where page 1 has no eligible reviewer and the saved reviewer is on page 2.

packages/web/src/pages/context-tree-settings-panel.tsx
Gandy2025: Fixed in `912e0fe`: member read-only rendering no longer falls back to `agentUuid` when `reviewerAgent` is null. It now shows the unavailable copy and keeps roster/picker access hidden. Added a regression test asserting the raw UUID is not rendered.

yuezengwu (approved):
Re-reviewed `912e0fe`. The two previously requested UI fixes are covered: the admin picker now pages through the full `/agents/all` roster via `fetchAllAgents`, and the member read-only fallback no longer exposes the raw reviewer UUID when the summary is unavailable. No further issues from my side.

Gandy2025:
Addressed review follow-up in `912e0fe`:

- Context Reviewer admin picker now uses `fetchAllAgents` to page through the full `/agents/all` roster before filtering active non-human org agents.
- Member read-only fallback no longer renders raw `agentUuid` when `reviewerAgent` is unavailable; it shows the unavailable reviewer copy instead.
- Added regression coverage for a saved reviewer on a later roster page and for member fallback not exposing the raw UUID.

Validation rerun locally:
- `pnpm --filter @first-tree/web test src/pages/__tests__/settings-panels-dom.test.tsx`
- `pnpm typecheck`
- `pnpm exec biome check packages/web/src/pages/context-tree-settings-panel.tsx packages/web/src/pages/__tests__/settings-panels-dom.test.tsx`

CI is now green; waiting on the remaining requested review from `baixiaohang`.

baixiaohang (approved):
Approved after re-review of 912e0fe. The prior Context Reviewer UI issues are covered: the admin picker now pages through the full org agent roster before filtering, and the member read-only fallback no longer exposes raw reviewer UUIDs when the summary is unavailable. CI is green.

PR #1297, opened 2026-06-25
feat: externalize native codex engine; resolve from PATH + one-click install

yuezengwu:
## What

Stop bundling the ~230MB native Codex engine by default. The runtime already resolves a system `codex` on PATH (bundled→PATH fallback in `packages/client/src/runtime/codex-binary.ts`), so the bundled binary is removed at install time and installed on demand instead.

**Footprint on a darwin-arm64 install: 485MB → 255MB.** The Claude engine and the small `@openai/codex-sdk` TS client are retained.

## Why this mechanism (the non-obvious part)

The 230MB binary is `@openai/codex-{platform}`, a **transitive optionalDependency** of `@openai/codex` (via `@openai/codex-sdk`). It is not in first-tree's tarball — it is re-resolved into the user's `node_modules` on `npm install -g first-tree`. Empirically:

| Lever | Effect | Zero-config? | Surgical? |
|---|---|---|---|
| `npm i --omit=optional` | 485M→46M | ✗ user flag | ✗ also drops the claude engine |
| `overrides` in the published package.json | binary still 230M | — | npm `overrides` only apply at the root project; an installed package's own overrides are **ignored** on global install |
| **postinstall prune** | 485M→255M | ✓ | ✓ only codex; claude + `@openai/codex-sdk` kept |

→ postinstall prune is the only zero-user-config + surgical lever. Caveat: `--ignore-scripts` installs skip it; the runtime then degrades to PATH / one-click install (no breakage).

## Changes

- `apps/cli/scripts/prune-codex-runtime-binary.mjs` — removes the `@openai/codex-{platform}` packages after a consumer install. Guarded against the dev monorepo (`pnpm-workspace.yaml` detection), honors `FIRST_TREE_KEEP_CODEX_BINARY=1`, wrapped so it **never fails the install**. Wired via `files` + `postinstall`.
- `daemon install-codex` (command) + `installCodexRuntime()` (core) — install the native engine on demand via `npm install -g @openai/codex`, reusing the self-update tracked-subprocess path (reaped on shutdown, hard timeout, error taxonomy), then re-probe.
- `formatCodexBinaryMissingMessage` now points at the one-click install.
- Updated the daemon-subcommand registration snapshots.

## Verification

- `pnpm check` / `pnpm typecheck` clean; `apps/cli` (507) + `@first-tree/client` (1167) tests pass.
- Prune e2e (real install, simulated postinstall cwd): 485M→255M, codex binary gone, claude + `@openai/codex-sdk` retained.
- `daemon install-codex` registered in the built CLI.

## Follow-up

Web one-click button: server route → WS reverse command (reuse `runtime-auth:start` channel) → capability remediation UI. Tracked separately.

This flips the default runtime behavior (bundled-first → external-by-default); the context-tree node `system/cloud/runtime/client-runtime.md` (Codex Handler section) will be updated to match when the feature lands.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -28,6 +28,7 @@
   "files": [
     "dist",
     "skills",
+    "scripts/prune-codex-runtime-binary.mjs",
     "README.md",
     "LICENSE"
   ],
@@ -41,7 +42,8 @@
     "test": "vitest run --passWithNoTests",
     "coverage": "pnpm run build && vitest run --coverage --passWithNoTests",
     "prepack": "rm -rf skills && cp -R ../../skills .",
-    "postpack": "rm -rf skills"
+    "postpack": "rm -rf skills",
+    "postinstall": "node scripts/prune-codex-runtime-binary.mjs"
   },
   "license": "Apache-2.0",
   "repository": {
--- a/apps/cli/scripts/prune-codex-runtime-binary.mjs
+++ b/apps/cli/scripts/prune-codex-runtime-binary.mjs
@@ -1,0 +1,112 @@
+#!/usr/bin/env node
+// Surgically remove the SDK-bundled native Codex engine after install.
+//
+// Why: the `@openai/codex-{platform}` package ships a ~225MB native `codex`
+// binary as a transitive optionalDependency of `@openai/codex-sdk`. First Tree
+// does not need it bundled — the runtime resolves a system `codex` on PATH
+// (see packages/client/src/runtime/codex-binary.ts) and the daemon can install
+// the native engine on demand (`daemon install-codex`). Removing the vendored
+// binary at install time trims the install footprint without touching the
+// small `@openai/codex-sdk` TypeScript client we actually import.
+//
+// Zero-config: runs as a `postinstall` of the published package. Honors
+// FIRST_TREE_KEEP_CODEX_BINARY=1 for users who prefer the bundled engine.
+// The whole body is wrapped so it NEVER exits non-zero — a prune failure must
+// not break the consumer's install (the runtime degrades gracefully to PATH).
+import { existsSync, readdirSync, rmSync, statSync } from "node:fs";
+import { createRequire } from "node:module";
+import { dirname, join } from "node:path";
+
+const PLATFORM_PKGS = [
+  "@openai/codex-darwin-arm64",
+  "@openai/codex-darwin-x64",
+  "@openai/codex-linux-arm64",
+  "@openai/codex-linux-x64",
+  "@openai/codex-win32-arm64",
+  "@openai/codex-win32-x64",
+];
+
+// Dev guard: when this runs as a workspace `postinstall` inside the first-tree
+// monorepo (pnpm install during development), the bundled codex binary is the
+// dev runtime we want to keep. Only prune in a real consumer install. Detect
+// the source checkout by walking up for a `pnpm-workspace.yaml` beside an
+// `apps/cli` — a published global/local install never has one above it.
+function insideSourceMonorepo() {
+  let dir = process.cwd();
+  for (let i = 0; i < 6; i++) {
+    if (existsSync(join(dir, "pnpm-workspace.yaml")) && existsSync(join(dir, "apps", "cli"))) return true;
+    const parent = dirname(dir);
+    if (parent === dir) break;
+    dir = parent;
+  }
+  return false;
+}
+
+function tryResolvePackageRoot(spec, fromPaths) {
+  try {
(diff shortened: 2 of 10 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1297

- Reason: the install-time prune currently has an unbounded ownership boundary, and the new public command is missing the required CLI reference update.

**Level**: A
- Path initial: touches `apps/cli/**` -> A.
- Semantic lift: also changes Codex runtime binary/capability behavior; already A.

**PR Summary**
- Problem: reduce global install footprint by not keeping the native Codex engine in every First Tree install.
- Design: prune `@openai/codex-{platform}` after consumer install, add `daemon install-codex` to install `@openai/codex` on demand, and update the Codex missing-binary message.
- Affected modules: `apps/cli` packaging/daemon/core, `packages/client` Codex binary error path.

**Review Comments**
❌ 1. The postinstall prune can delete a hoisted/shared package outside First Tree's package-owned tree. [R5 / `apps/cli/scripts/prune-codex-runtime-binary.mjs:87`]

`require.resolve(..., { paths: [process.cwd(), import.meta.dirname] })` walks upward from `node_modules/first-tree`. In a non-global consumer install, npm can hoist transitive deps to the app's root `node_modules`; then line 94 removes that root-level `@openai/codex-*` package, which may also be used by another dependency in the same project. Please bound the delete target to a First Tree-owned dependency subtree before `rmSync` runs, or skip pruning when the resolved package is outside that owned subtree.

❌ 2. `daemon install-codex` is a new visible public command but `docs/cli-reference.md` still omits it. [R5 / `docs/cli-reference.md:451`]

The command is registered and visible in help, and the new missing-binary message points users at it. The CLI command-surface contract requires public command additions to update the source CLI reference in the same PR. Please add it to the daemon command tree/table and include the important flags/behavior (`--spec`, `--json`, local install + post-install probe).

✅ Key pass: the install routine follows the existing tracked child-process shape, keeps npm spec validation, and silences human output in JSON mode.

**Executed**

yuezengwu:
Thanks @baixiaohang — both addressed in `e1a2b523`.

**1. Prune ownership boundary (R5).** The prune now only runs for a **global** install (`npm install -g first-tree`), gated on `npm_config_global=true`. In a non-global consumer install npm hoists `@openai/codex-*` to the app root where another dependency may share it — gating on the global flag skips that scenario entirely, so we never `rmSync` a package outside first-tree's own global root. Empirically confirmed npm sets `npm_config_global=true` for `-g` and leaves it `undefined` for a local dependency install; a package manager that doesn't export the flag also falls into the skip path (binary kept, runtime still resolves PATH — only the size win is forgone). Verified: local-install sim → skipped (binary kept); global-install sim → 241MB pruned, claude engine retained.

**2. `daemon install-codex` docs.** Added to `docs/cli-reference.md` — the daemon command tree plus a table row covering `--spec`, `--json`, the local `npm install -g @openai/codex`, and the post-install re-probe.

`pnpm check` / `typecheck` clean; cli 507 + client 1167 tests green. Ready for another look.

baixiaohang (approved):
## Review by codex-assistant - PR #1297

**Recommendation**: approve
- Reason: the two prior blockers are resolved and I found no new blocking issue in the follow-up revision.

**Level**: A
- Path initial: touches `apps/cli/**` -> A.
- Semantic lift: also changes Codex runtime/capability behavior; already A.

**PR Summary**
- Problem: reduce First Tree install footprint by not keeping the native Codex engine in every install.
- Design: prune the Codex platform binary only for global installs, provide `daemon install-codex` as the on-demand local install + re-probe path, and update the missing-binary remediation message.
- Affected modules: `apps/cli` packaging/daemon/core, `docs/cli-reference.md`, `packages/client` Codex binary messaging.

**Review Comments**
✅ 1. Prior prune ownership blocker is resolved: pruning is now gated on `npm_config_global === "true"`, so local consumer installs and package managers that do not expose the global flag skip the destructive prune path instead of deleting potentially shared app-root `@openai/codex-*` packages. [R5 / `apps/cli/scripts/prune-codex-runtime-binary.mjs`]

✅ 2. Prior CLI docs blocker is resolved: `docs/cli-reference.md` now documents `daemon install-codex [--spec <spec>] [--json]` in the daemon tree and command table, including local install and post-install re-probe behavior. [R5 / `docs/cli-reference.md`]

⚠️ Context Tree still records the pre-change Codex bundled-first policy; the PR body explicitly calls out the post-landing tree update, so I am not treating that as a code blocker here.

**Executed**
- Approved after explicit baixiaohang authorization for this A-level PR.

PR #1310, opened 2026-06-26
fix(web): drop misleading 'system CLI fallback' runtime-state copy

yuezengwu:
Under install-only detection both runtime engines are external by default, so a system `claude`/`codex` resolved on PATH (`runtimeSource: "path"`) is the **normal** case — calling it a "fallback" in the UI is misleading. The runtime state line now reads plainly `<provider> installed [vX]`; the `runtimeSource` provenance is still carried on the capability entry for diagnostics (`daemon doctor`, API).

- `runtime-state-line.tsx`: removed the ` · system CLI fallback` suffix.
- test updated: a system-PATH codex now asserts plain "installed" and `not.toContain("fallback")`.

Web suite green (1027), typecheck + Biome clean.

### DIFF
--- a/packages/web/src/pages/clients/__tests__/runtime-state-line.test.tsx
+++ b/packages/web/src/pages/clients/__tests__/runtime-state-line.test.tsx
@@ -33,7 +33,7 @@ afterEach(async () => {
 });
 
 describe("RuntimeStateLine", () => {
-  it("shows when Codex is running through the system CLI fallback", async () => {
+  it("shows a system-PATH codex as plainly installed (no `fallback` copy)", async () => {
     const entry: CapabilityEntry = {
       available: true,
       state: "ok",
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1310

**Recommendation**: approve
- Reason: The UI copy now matches install-only capability semantics: a path-backed runtime is a normal installed runtime, while `runtimeSource` remains on the capability entry for diagnostics.

**Level**: C
- Semantic lift: none; this does not touch subscriptions, Admin WebSocket frame parsing, or the web data layer.

**PR Summary**
- Solves: Removes misleading `system CLI fallback` copy from runtime capability state lines.
- Design: `RuntimeStateLine` now renders `ok` entries as `<Provider> installed [vX]`; the test updates the path-backed Codex case to assert no fallback copy.
- Affected modules: `packages/web/src/pages/clients/...`

**Review Notes**
✅ 1. The change is consistent with install-only detection: `ok` means the spawnable binary exists, while auth/runtime usability is handled elsewhere. [`packages/web/src/pages/clients/cards/shared/runtime-state-line.tsx:29`]
✅ 2. The test covers the specific regression by using `runtimeSource: "path"` and asserting the plain installed copy. [`packages/web/src/pages/clients/__tests__/runtime-state-line.test.tsx:36`]

**Executed**
- Approved.

PR #1299, opened 2026-06-25
feat: externalize native claude engine; resolve from PATH + one-click install

yuezengwu:
## What

Mirrors the codex externalization (#1297) for the Claude engine. The bundled
`@anthropic-ai/claude-agent-sdk-{platform}` native `claude` binary adds ~210MB
to every install, but the runtime already resolves a system `claude` **first**
(`CLAUDE_CODE_EXECUTABLE` env → PATH → well-known dirs `~/.local/bin`,
`~/.claude/local`; the SDK's bundled native binary is only the last-resort
fallback in `handlers/claude-executable.ts`). So we stop shipping it by default
and install it on demand.

## Changes

- **postinstall prune** — `apps/cli/scripts/prune-claude-runtime-binary.mjs`
  removes the `@anthropic-ai/claude-agent-sdk-{platform}` packages after a
  consumer install (~210MB). Structural mirror of the codex prune:
  - bound to **global installs only** (`npm_config_global=true`) — a non-global
    consumer install hoists the platform pkg to the app root where another dep
    may share it, so we never delete a shared copy there;
  - dev-monorepo guard (keeps the dev runtime), `FIRST_TREE_KEEP_CLAUDE_BINARY=1`
    opt-out, whole body wrapped to **never exit non-zero**;
  - chained after the codex prune in the same `postinstall`.
- **`daemon install-claude`** + core `installClaudeRuntime()` — on-demand
  remediation via `npm install -g @anthropic-ai/claude-code`, reusing the
  self-update tracked-subprocess path (reap on shutdown, hard timeout, error
  taxonomy), then re-probes the `claude-code` capability.
- Points the claude bundled-missing probe message at the one-click install.
- Retains the small `@anthropic-ai/claude-agent-sdk` TS SDK we import.

With codex already external (#1297), a darwin-arm64 global install drops
**255MB → ~46MB**.

## Verification

- `pnpm typecheck` (cli + client) clean; `biome check` clean.
- CLI suite **507/507**; client `capability-probes` **86/86** (incl. the real
  `verifyBundledClaudeArtifact` launch test). Updated the two registration
  tests (`command-registration-smoke`, `daemon-refresh-unit`).
- Prune logic verified against a **simulated flat npm-global layout**: removes
  `@anthropic-ai/claude-agent-sdk-darwin-arm64`, **keeps** the JS
  `@anthropic-ai/claude-agent-sdk`; both guards (keep-flag, non-global) no-op
  with exit 0.
[...]

PR #1312, opened 2026-06-26
feat: prune unused claude SDK JS peer deps on global install (~15MB)

yuezengwu:
## What

Extends the install-footprint prunes (codex #1297 / claude binary #1299) to the **JS peer deps** of `@anthropic-ai/claude-agent-sdk`. Adds `apps/cli/scripts/prune-claude-sdk-deps.mjs`, chained into `postinstall`, which removes `@anthropic-ai/sdk` (~9.6MB) + `@modelcontextprotocol/sdk` (~5.8MB) on a global install. **~15MB** off `npm i -g first-tree`.

## Why these are safe to remove

The SDK declares both as `peerDependencies`, so npm auto-installs them — but they are dead weight at runtime here:

- The SDK entry First Tree loads (`sdk.mjs`, the `.` export) references **neither** package. It is a self-contained bundle; even `createSdkMcpServer` ships its own MCP implementation rather than importing `@modelcontextprotocol/sdk`. The only dynamic imports in the bundle are `node:` builtins.
- First Tree's own code never imports either package — we use only `query()` + types from the bundled `.` entry (`packages/client/src/handlers/claude-code.ts`), and never `createSdkMcpServer`.
- `zod` (the third peer) is **kept** — First Tree depends on it directly and shares it.

## Scope — Tier-1 only (deliberate)

Deletes **only the two scoped root packages**. It does **not** GC their now-orphaned transitive deps (hono/ajv/express/qs/…): `iconv-lite` in that subtree is **also** pulled by `@inquirer/prompts → @inquirer/editor → external-editor`, so a blind subtree delete would break the CLI's interactive editor. The ~6MB of dangling orphans are harmless; recovering them safely would need a live-tree reachability GC, out of scope.

## Guards (mirror the binary prunes)

- global-install-only (`npm_config_global=true`) ownership boundary — non-global installs hoist to a shared root, so we skip
- `insideSourceMonorepo()` dev guard (keeps peers for `tsc` in the monorepo)
- `FIRST_TREE_KEEP_CLAUDE_SDK_DEPS=1` escape hatch
- whole body wrapped in try/catch — a prune failure never fails the install

## Self-defending version-drift guard

We pin the SDK with `^`, so "never referenced" is a property of the current bundle, not a contract. Before deleting each package the script greps the installed SDK entry for the specifier; if a future SDK build starts importing a peer at runtime, it is **kept**. This turns a would-be silent, global-install-only runtime crash into merely forgoing the size win.

## Verification

- Local prune against a real install: `node_modules` 40M → 25M; `zod` + `@anthropic-ai/claude-agent-sdk` survive; `import('@anthropic-ai/claude-agent-sdk')` still resolves `query`.
- env-override / non-global → no-op (verified).
- Self-defending guard: injected an `import "@modelcontextprotocol/sdk"` into `sdk.mjs` → that package is kept, the unreferenced one is still pruned (verified).
- `pnpm --filter first-tree-dev typecheck` clean; **505 cli tests pass**; Biome clean.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -30,6 +30,7 @@
     "skills",
     "scripts/prune-codex-runtime-binary.mjs",
     "scripts/prune-claude-runtime-binary.mjs",
+    "scripts/prune-claude-sdk-deps.mjs",
     "README.md",
     "LICENSE"
   ],
@@ -44,7 +45,7 @@
     "coverage": "pnpm run build && vitest run --coverage --passWithNoTests",
     "prepack": "rm -rf skills && cp -R ../../skills .",
     "postpack": "rm -rf skills",
-    "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs"
+    "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs && node scripts/prune-claude-sdk-deps.mjs"
   },
   "license": "Apache-2.0",
   "repository": {
--- a/apps/cli/scripts/prune-claude-sdk-deps.mjs
+++ b/apps/cli/scripts/prune-claude-sdk-deps.mjs
@@ -1,0 +1,186 @@
+#!/usr/bin/env node
+// Surgically remove the SDK's unused JS peer dependencies after install.
+//
+// Why: `@anthropic-ai/claude-agent-sdk` declares `@anthropic-ai/sdk` (~9.6MB)
+// and `@modelcontextprotocol/sdk` (~5.8MB) as peerDependencies, so npm 7+
+// auto-installs both (~15.4MB) on a global install. But the compiled SDK
+// entry First Tree actually loads (`sdk.mjs`, the `.` export) references
+// NEITHER package — it is a self-contained bundle, and `createSdkMcpServer`
+// ships its own MCP implementation rather than importing the MCP SDK. First
+// Tree's own code never imports either package (we only use `query()` + types
+// from the bundled `.` entry). So both peers are dead weight at runtime.
+//
+// Scope (the safe "Tier-1" cut): we delete ONLY the two scoped root packages
+// `@anthropic-ai/sdk` and `@modelcontextprotocol/sdk`. These are scoped and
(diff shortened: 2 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1312

**Recommendation**: approve
- Reason: The change is scoped to a best-effort global-install prune, mirrors the existing binary-prune guards, and current dependency evidence supports that removing these two JS peers does not affect First Tree runtime.

**Level**: A
- Path initial: touches `apps/cli/**` -> A
- Semantic lift: none

**PR Summary**
- Solves: Reduces `npm install -g first-tree` footprint by pruning unused JS peer dependencies auto-installed for `@anthropic-ai/claude-agent-sdk`.
- Design: Adds `apps/cli/scripts/prune-claude-sdk-deps.mjs` to the existing `postinstall` chain. It only runs for global installs, skips source-monorepo installs, honors `FIRST_TREE_KEEP_CLAUDE_SDK_DEPS=1`, and fails closed if it cannot prove the SDK entry does not reference the peer package.
- Affected modules: `apps/cli/package.json`, `apps/cli/scripts/prune-claude-sdk-deps.mjs`

**Review Notes**
✅ 1. The ownership boundary and failure behavior match the existing Codex/Claude native-binary prune scripts: source monorepo guard, `npm_config_global=true`, escape hatch, and whole-body best effort handling. [R4 / `apps/cli/scripts/prune-claude-sdk-deps.mjs`]
✅ 2. The prune only removes the two scoped root peer packages and deliberately avoids transitive GC, which keeps shared transitive packages out of the blast radius. [R5 / `apps/cli/scripts/prune-claude-sdk-deps.mjs`]
✅ 3. I verified the current SDK metadata and tarball shape: `@anthropic-ai/claude-agent-sdk@0.3.187` exports `./sdk.mjs` for `.`, and that entry does not reference `@anthropic-ai/sdk` or `@modelcontextprotocol/sdk`.
⚠️ 4. Non-blocking: the version-drift guard matches the current SDK export shape. If the SDK later moves to more complex conditional exports, the next SDK upgrade should re-check this entry-resolution logic.

**Executed**
- Approved after baixiaohang authorization. I did not run local checks or tests; GitHub checks were green at review time.

PR #1300, opened 2026-06-25
feat(runtime): install-only capability detection + in-chat runtime login

yuezengwu:
## What

Reverses runtime-provider capability detection from **launch-verified** (resolve → auth precheck → smoke, smoke the only path to `ok`) to **install-only** existence detection, and moves the runtime **login entry point from the computer card to the chat**.

This implements a 4-part request:
1. Audited every product dependency on detection results.
2. Found detection was **not unified** (3 independent paths: capability probe / handler spawn-time resolve / codex handler fallback) + a daemon-vs-shell `PATH` divergence.
3. **Detection → install-only.** A provider is `ok` when the binary the runtime would spawn is resolvable on disk (env / `PATH` / well-known dir, or the SDK bundle while it still ships), `missing` otherwise, `error` on throw. No `--version` launch, no auth precheck, no session smoke, no `codex doctor` / `codex login status`.
4. **In-chat login.** When a running session fails with a `credential` error, the chat renders a **"Connect <provider>"** action that drives the existing in-product daemon browser-OAuth login.

## Why

The mandatory auth-precheck + real-session smoke were the dominant source of **false negatives** (a working third-party-gateway / proxy credential the precheck didn't recognise; a transient provider hiccup downgrading a healthy install) and of **token + latency cost** on every probe and re-probe. Whether a provider is *authenticated* or *reachable* is a run-time question (a credential valid at probe time can be revoked an hour later) — so it's now discovered when a session actually runs and fails, and surfaced where the user is blocked (the chat), not pre-computed on a card.

## Changes

- **shared** — `CapabilityEntry.state` collapses to `ok | missing | error`; drop `unauthenticated` and the `authenticated` / `authMethod` / `degraded` / `probeKind` fields (+ their schemas/exports). Keep `available` (= installed), `runtimeSource` / `runtimePath`, `pendingAuth` / `lastAuthError`.
- **client** — three probes rewritten to existence checks sharing one set of primitives (new `capabilities/detect.ts`); `launch-probe.ts` trimmed to the spawn helpers the codex handler-resolver still needs; `claude-shared.ts` (auth detector) deleted; reconnect/revalidate simplified (detection is cheap). `resolveCodexRuntimeBinary` is unchanged — the handler still launch-verifies the binary it spawns; the probe is a strict subset of it.
[...]

PR #1314, opened 2026-06-26
fix(client): canonicalize login-shell PATH dirs; gate CLAUDE_CODE_EXECUTABLE on executability

yuezengwu:
## What

Two install-only capability-detection robustness gaps, surfaced by deep-reading **Multica**'s probe layer (`server/internal/daemon/config.go`) and comparing it against First Tree's current `login-shell-path.ts` / `claude-executable.ts`.

### 1. fnm/nvm "multishell" dirs vanish before search → false `missing`

`wellKnownBinDirs` deliberately can't enumerate version-manager install dirs, so a `claude`/`codex` installed via a Node version manager is found **only** through the login-shell PATH probe. But fnm `--use-on-cd` (and nvm shims) put per-session **symlink** dirs on PATH (e.g. `/tmp/fnm_multishells/xxx/bin`) that are torn down when the probe shell exits. `getLoginShellPathDirs` recorded the raw dir string and memoized it; by the time a resolver `existsSync`-checks it, the dir is gone → the binary is invisible → false `missing`.

**Fix:** the probe now canonicalizes each PATH dir with `cd "$d" && pwd -P` **inside the still-alive shell**, resolving the ephemeral symlink to the stable underlying install dir before returning. Splitting uses `tr ':' '\n'` rather than `for d in $PATH`, because **zsh** (the macOS default) does not field-split an unquoted scalar and a for-loop would iterate once over the whole string. Verified to split correctly and canonicalize a symlinked dir under bash, zsh, and sh.

(Multica solves the same case with `cd "$dir" && pwd -P` + a daemon-side `LookPath` reality-check; this adapts the idea to First Tree's "grab the PATH dirs, search in-process" model.)

### 2. `CLAUDE_CODE_EXECUTABLE` accepted via bare `existsSync()`

A directory named `claude` or a non-executable shim passed the override check and yielded a false `ok` the SDK then couldn't spawn — while the PATH/well-known search already used the stricter `isFile + X_OK` gate (`isExecutableFile`). The override now goes through the same gate.

The override stays **soft** (an unusable override still falls through to PATH / well-known / login-shell — preserving existing intentional behavior, see the `falls through to PATH lookup when the env override does not exist` test), but the reason is captured as `overrideError` and surfaced in the probe's `missing` message **only when nothing else resolves**, so a misconfigured override gets named precisely instead of the generic "no claude found" text.

> Note: Multica hard-misses on an unusable explicit `MULTICA_*_PATH` ("honored exactly, don't substitute a different binary"). First Tree's existing test encodes the opposite (soft fall-through) on purpose, so this PR keeps fall-through and only improves the executability gate + diagnosability rather than reversing that decision.

## Test

- `pnpm typecheck` (root, 12/12) ✅
- `pnpm --filter @first-tree/client test` — 1145/1145 ✅ (login-shell-path / claude-executable / capability-probes updated)
- Biome clean ✅
- New tests: dir override → fall-through; non-executable override → fall-through; `overrideError` surfaced on total miss; no `overrideError` when PATH resolves.

## Background

From research comparing Multica/Paperclip's local-CLI detection with First Tree's post-#1300 install-only model.

### DIFF
--- a/packages/client/src/__tests__/claude-executable.test.ts
+++ b/packages/client/src/__tests__/claude-executable.test.ts
@@ -44,6 +44,56 @@ describe("resolveClaudeCodeExecutable", () => {
     expect(resolution).toEqual({ path: fakeClaude, source: "path" });
   });
 
+  it("ignores an override that is a directory (not an executable file) and falls through to PATH", () => {
+    const dirOverride = mkdtempSync(join(tmpdir(), "ftt-claude-dirovr-"));
+    try {
+      const resolution = resolveClaudeCodeExecutable({
+        env: { CLAUDE_CODE_EXECUTABLE: dirOverride, PATH: binDir },
+      });
+      // A bare existsSync() would have accepted the directory; isExecutableFile rejects it.
+      expect(resolution).toEqual({ path: fakeClaude, source: "path" });
+    } finally {
+      rmSync(dirOverride, { recursive: true, force: true });
+    }
+  });
+
+  it("ignores a non-executable override file and falls through to PATH", () => {
+    const nonExec = join(binDir, "claude-nonexec");
+    writeFileSync(nonExec, "#!/bin/sh\nexit 0\n");
+    chmodSync(nonExec, 0o644); // readable, NOT executable
+    try {
+      const resolution = resolveClaudeCodeExecutable({
+        env: { CLAUDE_CODE_EXECUTABLE: nonExec, PATH: binDir },
+      });
+      expect(resolution).toEqual({ path: fakeClaude, source: "path" });
+    } finally {
+      rmSync(nonExec, { force: true });
+    }
+  });
+
+  it("surfaces overrideError on the default resolution when an unusable override and nothing else resolves", () => {
+    const badOverride = join(binDir, "nonexistent-override");
+    const resolution = resolveClaudeCodeExecutable({
+      env: {
+        CLAUDE_CODE_EXECUTABLE: badOverride,
+        PATH: join(tmpdir(), "definitely-not-a-real-bin-dir-xyz"),
+        HOME: emptyHome,
+      },
+      loginShellPathDirs: noLoginShell,
+    });
+    expect(resolution.path).toBeUndefined();
+    expect(resolution.source).toBe("default");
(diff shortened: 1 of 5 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1314

**建议**: approve
- 理由: 修复点与 install-only capability contract 一致，未发现阻塞问题。

**分级**: A
- Path 初判: 包含 `packages/client/**` -> A
- 语义升级: 无

**PR 摘要**
- 解决问题: 修复 login-shell PATH 中 fnm/nvm multishell 临时 symlink 目录在 probe shell 退出后失效导致的 false missing，并收紧 `CLAUDE_CODE_EXECUTABLE` override 的可执行性判断。
- 设计方案: login-shell probe 在 shell 仍存活时把 PATH dirs canonicalize 为物理目录；Claude executable resolver 让 env override 复用 executable regular-file gate，坏 override 保持 soft fall-through，并仅在总缺失时把 `overrideError` 前缀到 missing 文案。
- 受影响模块: `packages/client/src/runtime/login-shell-path.ts`, `packages/client/src/handlers/claude-executable.ts`, Claude capability probe 及相关测试。

**Review 意见**
✅ 1. login-shell probe 仍只在 capability path / session-start lazy path 执行，pre-connect registration 继续 `includeLoginShell:false`，不会阻塞 daemon 连接。 [R4]
✅ 2. resolver 顺序和 soft override 行为保持：env 可执行文件优先；坏 override fall through 到 PATH / well-known / login-shell / bundle；只有全部失败才展示 `overrideError`。 [R1/R4]
✅ 3. 共享的 `getLoginShellPathDirs` 改动同样帮助 Codex PATH fallback；Codex bundled-first resolver 未被改变。 [R5]
⚠️ 4. 非阻塞：probe shell script 依赖 POSIX shell 语法与 `tr`；PR 已覆盖 bash/zsh/sh，未扩大到 fish 等非 POSIX shell。当前源码也未声明支持这些 shell，不建议阻塞。

**已执行**
- 已按 baixiaohang 授权提交 approve review。
- 未运行本地代码检查或单测（PR Review 模式限制）；GitHub checks 当前全绿。

PR #1322, opened 2026-06-26
Add first-tree-welcome skill eval gate

baixiaohang:
## Summary

- Add the `first-tree-welcome` live gate runner for the three implemented onboarding rows: tree kickoff lane routing, no-repo intro, and readable repo + populated Context Tree.
- Keep all nine welcome setup-matrix rows declared for floor coverage while leaving the six remaining rows planned.
- Add recorded-only chat shims, blocked `gh` / GitHub / tree-setup side-effect guards, fixture baseline checks, and welcome grader/unit coverage.

## Verification

- `pnpm --filter @first-tree/skill-evals eval:floor`
- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-welcome`
- `pnpm --filter @first-tree/skill-evals eval:gate -- --suite first-tree-welcome`
- `pnpm exec biome check packages/skill-evals/src packages/skill-evals/README.md`
- `pnpm --filter @first-tree/skill-evals validate:first-tree-read-fixtures`
- `git diff --check`

## Review

- Pre-PR review by `codex-reviewer` found two side-effect guard blockers.
- Both were fixed before this PR: forbidden external commands now fail the case, and source/context fixture path existence plus HEAD baselines are checked.

### DIFF
--- a/packages/skill-evals/README.md
+++ b/packages/skill-evals/README.md
@@ -9,6 +9,7 @@ quota.
 ```bash
 pnpm --filter @first-tree/skill-evals eval:floor
 pnpm --filter @first-tree/skill-evals eval:gate -- --suite first-tree-write
+pnpm --filter @first-tree/skill-evals eval:gate -- --suite first-tree-welcome
 pnpm --filter @first-tree/skill-evals eval:first-tree-read
 pnpm --filter @first-tree/skill-evals eval:first-tree-read -- --case tree-software-trigger
 pnpm --filter @first-tree/skill-evals eval:first-tree-read -- --json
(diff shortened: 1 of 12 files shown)


yuezengwu (changes requested):
Goal/core changes: this PR wires `first-tree-welcome` into the skill eval floor/gate flow, adds the live runner for the three implemented onboarding rows, and adds command shims/fixture baselines to keep welcome evals side-effect-free.

Blocking finding:

- `packages/skill-evals/src/suites/first-tree-welcome/grader.ts` around the row-3 pass path: row 3 declares `setup-as-first-task` as forbidden in `cases.ts`, but `forbiddenActionHits` never handles that action and `casePassed()` for `ask_for_repo_path_or_url` only checks that repo/tree evidence was not read. That means a model can still pass the implemented no-repo intro gate while packaging setup as task options, for example by calling `first-tree chat ask ... --options` with setup choices, as long as the final text mentions a local clone path or GitHub URL. This is exactly the row's forbidden behavior, so the gate has a false-positive path.

Please add a matcher for `setup-as-first-task` or make the row-3 pass condition reject `taskOptionsObserved` / setup-option asks, and cover it with a grader unit test.

No core data structure or database changes observed.

yuezengwu (approved):
Re-reviewed the synchronized head `3ebec77d`.

The previous blocker is addressed: row 3 now detects `setup-as-first-task`, rejects `taskOptionsObserved` in the no-repo pass condition, and has unit coverage for setup options being packaged as a first-task ask.

Verification run locally after installing deps in the review worktree:

- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-welcome`
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-welcome/grader.ts packages/skill-evals/src/suites/first-tree-welcome/__tests__/grader.test.ts`

No core data structure or database changes observed.

PR #1326, opened 2026-06-26
fix(client): make login-shell PATH probe shell-agnostic (fish users no longer reported missing)

yuezengwu:
## Problem

When a user's default shell is **fish** (or another non-POSIX shell: tcsh/csh), every `claude` / `codex` reachable only through fish's interactive PATH (`config.fish` / `fish_add_path` / fnm / nvm) is detected as **`missing`** — even though `fish -lic 'type -p claude'` finds it fine. Onboarding / `daemon doctor` then report the runtime missing, push users toward needless `daemon install-claude` / `install-codex`, and capability binding fails.

### Root cause

`packages/client/src/runtime/login-shell-path.ts` ran the PATH-discovery script as a **POSIX `while … do … done` pipeline inlined directly in the user's `$SHELL`**:

```sh
printf %s '<DELIM>'; printf %s "$PATH" | tr ':' '\n' | \
  while IFS= read -r d; do [ -n "$d" ] && (cd "$d" 2>/dev/null && pwd -P); done; printf %s '<DELIM>'
```

fish's loop syntax is `while …; …; end` — no `do`/`done`. fish parses the whole `-c` string as one unit and **errors at parse time** (`Missing end to balance this while loop`), so even the leading `printf` never runs → stdout is empty → the probe finds no dirs → returns `null` (treated as a retryable failure) → login-shell PATH discovery yields nothing.

This is a regression from #1314, which replaced #1300's simpler `printf %s "$PATH"` with the POSIX loop to canonicalize fnm/nvm ephemeral symlinks — and broke fish in the process.

## Fix

Use the login shell only as a **launcher**: it runs a single opaque `/bin/sh -c '…'` token — a literal, single-quoted string it never parses as code — and all `$PATH` splitting + canonicalization happens inside that **nested POSIX `sh`**. fish (and any shell) just exec's `/bin/sh`; it never sees the loop.

```sh
/bin/sh -c 'printf %s __FT_SHELL_PATH__; IFS=:; for d in $PATH; do [ -n "$d" ] && (cd "$d" 2>/dev/null && pwd -P); done; printf %s __FT_SHELL_PATH__'
```

### Why canonicalization stays in-shell (not Node `realpathSync`)

A tempting alternative is to emit only the raw `$PATH` and canonicalize in Node. But fnm/nvm per-session **multishell symlinks** (`/tmp/fnm_multishells/.../bin`) are **torn down when the login shell exits** — exactly the reason #1314 canonicalizes in-shell. A Node `realpathSync` runs *after* the probe subprocess (and its symlinks) are gone → `ENOENT` → the dir is dropped, reintroducing the bug #1314 fixed. Running the canonicalization in the nested `sh` keeps it **inside the login shell's lifetime**, while those symlinks are still alive, so we preserve #1314's capability and fix the fish regression together. The nested `sh` reads the login shell's exported `$PATH`, which is colon-delimited regardless of how the outer shell stores it internally.

## Testing

- Existing 11 hermetic unit tests pass **unchanged** — the `RunShell` seam contract (newline-delimited canonical dirs bracketed by `DELIM`) is identical.
- Added 2 guarded (`skipIf(win32)`) real-execution tests that run the probe script through a POSIX `/bin/sh` launcher and via the real default shell, asserting canonical absolute dirs.
- Verified end to end under bash, zsh, and sh locally.
- fish itself is covered by the runtime-env-qa **`DW7_fish_frozen`** scenario (expects `ok`, was `FAIL=missing`) — should flip green with this fix.
- `pnpm --filter @first-tree/client typecheck` + full client suite (1147 tests) + `biome check` all green.

### DIFF
--- a/packages/client/src/__tests__/login-shell-path.test.ts
+++ b/packages/client/src/__tests__/login-shell-path.test.ts
@@ -1,5 +1,11 @@
+import { spawnSync } from "node:child_process";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { getLoginShellPathDirs, type RunShell, resetLoginShellPathDirsCache } from "../runtime/login-shell-path.js";
+import {
+  buildProbeScript,
+  getLoginShellPathDirs,
+  type RunShell,
+  resetLoginShellPathDirsCache,
+} from "../runtime/login-shell-path.js";
 
 const DELIM = "__FT_SHELL_PATH__";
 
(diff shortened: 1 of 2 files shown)


baixiaohang (dismissed):
## Review by codex-assistant - PR #1326

- Reason: the main nested-`sh` approach looks aligned with the runtime capability contract, but the new real-execution test invokes `/bin/sh -lic`, which is not portable to the repo's Ubuntu CI environment.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: login-shell PATH discovery broke for fish-style shells because a POSIX loop was parsed by the outer user shell.
- Design: make the outer login shell launch an opaque `/bin/sh -c '...'` command, and keep PATH splitting plus canonicalization inside the nested POSIX shell while fnm/nvm session symlinks are alive.
- Affected modules: `packages/client/src/runtime/login-shell-path.ts`, `packages/client/src/__tests__/login-shell-path.test.ts`

**Review Findings**
❌ 1. The new `/bin/sh` launcher test can fail on Ubuntu CI. `.github/workflows/ci.yml` runs tests on `ubuntu-latest`, where `/bin/sh` is commonly `dash`; `dash` does not support the `-l` flag, so `spawnSync("/bin/sh", ["-lic", buildProbeScript()])` exits before running the probe and produces empty stdout. That makes `getLoginShellPathDirs(runViaSh)` return `[]`, failing `expect(dirs.length).toBeGreaterThan(0)`. If the goal is to exercise POSIX parsing of the opaque nested command, invoke `/bin/sh` with `-c` here; if the goal is to exercise production login-shell flags, use a shell known to support `-lic` or skip when unavailable, and assert `status/error` so failures are explicit. [`packages/client/src/__tests__/login-shell-path.test.ts:141`]

✅ 2. The production probe keeps canonicalization in the nested shell before the login-shell process exits, preserving the fnm/nvm symlink behavior the capability resolver depends on.

**Executed**

yuezengwu:
Thanks @baixiaohang — good catch on finding #1, fixed in 8609233e.

The `-l`/`-i` flags only exist to source the user's rc files (which production needs, but this test does not), and they're not portable across `/bin/sh` implementations. Changed the test's outer launcher from `/bin/sh -lic` to **`/bin/sh -c`** (universally supported), and added explicit `expect(r.error).toBeUndefined()` / `expect(r.status).toBe(0)` assertions so a launcher failure surfaces directly instead of as an opaque empty result.

Verified locally that the nested `for`/`cd`/`pwd -P` loop runs identically under `dash` (both as the outer launcher and as the nested `/bin/sh`), so this stays green on Ubuntu CI where `/bin/sh` → `dash`:

```
$ /bin/dash -c "/bin/sh -c 'printf %s __D__; IFS=:; for d in $PATH; do [ -n \"$d\" ] && (cd \"$d\" 2>/dev/null && pwd -P); done; printf %s __D__'"
# exit 0, DELIM-bracketed canonical absolute dirs
```

Finding #2 (canonicalization stays in-shell to preserve fnm/nvm symlink resolution) — agreed, that's the load-bearing part of the design. Production `defaultRunShell` keeps `-lic` since it must source the user's rc files; the change is test-only.

PR #1327, opened 2026-06-26
Add first-tree-seed skill eval gate

baixiaohang:
## Summary

- Add the `first-tree-seed` live gate suite with four deterministic cases: empty tree + source present, non-empty tree refusal, missing source refusal, and bare source worktree protocol.
- Build isolated seed fixtures with per-run context tree repos plus tiny bare source repos, and wire `eval:gate -- --suite first-tree-seed` into the skill-evals CLI.
- Add deterministic seed grading for required skill reads, workspace self-check, source worktree materialization, source evidence reads, no direct bare-source reads, no tree/source mutation, no Phase 2 before approval, and forbidden side effects.

## Verification

- `pnpm --filter @first-tree/skill-evals eval:floor`
- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-seed`
- `pnpm --filter @first-tree/skill-evals eval:gate -- --suite first-tree-seed`
- `pnpm exec biome check packages/skill-evals/src packages/skill-evals/README.md`
- `pnpm --filter @first-tree/skill-evals validate:first-tree-read-fixtures`
- `git diff --check`

Local note: `@first-tree/client` build exited 0 with the existing unresolved-import warnings for `@first-tree/shared` externals.

## Review

- Pre-PR review by `codex-reviewer` found three deterministic oracle blockers.
- Fixed before this PR: skeleton/protocol cases now require an actual `first-tree-write/SKILL.md` read, non-empty tree refusal fails on source exploration/direct bare reads, and skill-file-read detection only uses tool input/path fields rather than stdout or read-file content.
- No `skills/*/SKILL.md` files were changed and seed hard rules were not relaxed.

### DIFF
--- a/packages/skill-evals/src/suites/first-tree-seed/grader.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/grader.ts
@@ -1,0 +1,602 @@
+import { existsSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+
+import { runCommand } from "../../core/commands.js";
+import { isRecord, isStringArray } from "../../core/events.js";
+import type { RunPaths } from "../../core/types.js";
+import type { EvalMetrics, FirstTreeSeedEvalCase, FixtureValidation } from "./types.js";
+
+const TEXT_KEYS = ["content", "message", "output_text", "text"];
+const SKILL_READ_INPUT_KEYS = new Set([
+  "args",
+  "arguments",
+  "argv",
+  "cmd",
+  "command",
+  "filePath",
+  "file_path",
+  "input",
+  "params",
+  "path",
+  "relativePath",
+  "relative_path",
+  "targetFile",
+  "target_file",
+]);
+const SKILL_READ_OUTPUT_KEYS = new Set([
+  "aggregated_output",
+  "content",
+  "message",
+  "output",
+  "output_text",
+  "stderr",
+  "stdout",
+  "text",
+]);
+
+function eventType(event: Record<string, unknown>): string | null {
+  return typeof event.type === "string" ? event.type : null;
+}
+
+function isModelPhase(event: Record<string, unknown>): boolean {
+  return event.phase === "model";
+}
+
+function collectToolInputStrings(value: unknown): string[] {
+  if (typeof value === "string") return [value];
+  if (Array.isArray(value)) {
+    const strings: string[] = [];
+    for (const item of value) {
+      strings.push(...collectToolInputStrings(item));
+    }
+    return strings;
+  }
+  if (!isRecord(value)) return [];
+
+  const strings: string[] = [];
+  for (const [key, item] of Object.entries(value)) {
+    if (SKILL_READ_OUTPUT_KEYS.has(key)) continue;
+    if (SKILL_READ_INPUT_KEYS.has(key)) {
+      strings.push(...collectToolInputStrings(item));
+    } else if (isRecord(item) || Array.isArray(item)) {
+      strings.push(...collectToolInputStrings(item));
+    }
+  }
+  return strings;
+}
+
+function containsSkillFileRead(event: unknown, skillName: string): boolean {
+  if (!isRecord(event)) return false;
+  if (eventType(event) !== "codex_event") return false;
+
+  const nestedEvent = event.event;
+  return collectToolInputStrings(nestedEvent).some(
+    (value) =>
+      (value.includes(".agents/skills/") || value.includes(".claude/skills/")) &&
+      value.includes(`${skillName}/SKILL.md`),
+  );
+}
+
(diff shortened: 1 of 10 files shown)


packages/skill-evals/src/suites/first-tree-seed/grader.ts
```diff
@@ -0,0 +1,591 @@
...
+    (value) =>
+      (value.includes(".agents/skills/") || value.includes(".claude/skills/")) &&
+      value.includes(`${skillName}/SKILL.md`),
+  );
+}
+
+function containsPathAccess(event: unknown, patterns: readonly string[]): boolean {
+  if (!isRecord(event)) return false;
+  if (eventType(event) !== "codex_event") return false;
+  const serialized = JSON.stringify(event.event) ?? "";
```
yuezengwu: This helper scans the full serialized event, including command output. That makes access checks like `workspaceManifestReadObserved` susceptible to false positives: a model can read `AGENTS.md` or another file whose output mentions `./.first-tree/workspace.json`, and line 432 will count it as an actual manifest read. The seed gate is using this as a required self-check, so it should mirror the stricter skill-file detector and inspect only tool input/path/argv fields, not stdout/aggregated output.

packages/skill-evals/src/suites/first-tree-seed/grader.ts
```diff
@@ -0,0 +1,591 @@
...
+    }
+  }
+
+  return [...new Set(hits)];
+}
+
+function directBareSourceContentRead(events: readonly unknown[]): boolean {
+  for (const event of events) {
+    if (
+      containsPathAccess(event, [
```
yuezengwu, line 355: This does not catch the direct bare-source reads that are most likely in a bare repo. `source-repos/source-repo/README.md` is not a real checkout path, but `git -C source-repos/source-repo show refs/remotes/origin/main:README.md` or `git -C source-repos/source-repo grep ...` reads source content directly from the bare object store. If the model also creates the expected worktree path, the case can still satisfy `sourceWorktreeCreated` + source evidence while violating the protocol. Please detect content-reading git commands against the bare source, or require the source evidence access itself to come from worktree input/path fields.

yuezengwu (changes requested):
I reviewed this as a gate/oracle change. The implementation adds the `first-tree-seed` live gate suite, fixtures, runner, grader, summaries, and CLI wiring for the four bootstrap boundary cases.

Requesting changes for two deterministic oracle gaps in the new grader. Both affect the core claims of this PR: required workspace self-check/source evidence reads and the bare-source worktree protocol.

baixiaohang:
Addressed the two deterministic oracle gaps in `6dcb1a15`:

- `containsPathAccess()` now mirrors the stricter skill-file detector and only inspects tool input/path/argv-like fields. It no longer scans stdout / aggregated command output, so a briefing or other file that merely mentions `.first-tree/workspace.json` does not count as a manifest read.
- Direct bare-source reads now catch content-reading git commands against `source-repos/source-repo`, including `git -C source-repos/source-repo show ...`, `grep`, `ls-tree`, `cat-file`, `archive`, `diff`, and `blame`. These trigger `direct_bare_source_read`, so creating the expected worktree path no longer masks a bare-object-store read.

Added regression tests for both paths:

- command output mentioning `.first-tree/workspace.json` does not set `workspaceManifestReadObserved`;
- `git -C source-repos/source-repo show refs/remotes/origin/main:README.md` sets `directBareSourceContentReadObserved` and fails the bare-source protocol case.

Verification after the fix:

- `pnpm --filter @first-tree/skill-evals test` (45 passed)
- `pnpm exec biome check packages/skill-evals/src packages/skill-evals/README.md`
- `pnpm --filter @first-tree/skill-evals eval:floor`
- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-seed`
- `pnpm --filter @first-tree/skill-evals eval:gate -- --suite first-tree-seed` (4/4 passed)
- `git diff --check`

No `skills/*/SKILL.md` files were changed and no seed hard rules were relaxed.

yuezengwu (approved):
Re-reviewed the synchronized head `6dcb1a15`.

The two prior oracle blockers are addressed:

- `containsPathAccess()` now uses input/path/argv-like fields via `collectToolInputStrings()` and skips stdout/aggregated output, so manifest mentions in command output no longer count as a manifest read.
- Direct bare-source content reads through `git -C source-repos/source-repo show/grep/ls-tree/cat-file/archive/diff/blame` now trigger `direct_bare_source_read`, and the regression test covers the `git show ...:README.md` path.

No further blocking findings in this review scope. I did not re-run the pnpm test/typecheck suite locally because this fresh review worktree has no `node_modules`; I did run `git diff --check origin/main...HEAD`, which passed.

PR #1329, opened 2026-06-28
fix: stop treating a flaky codex --version check as a missing binary

yuezengwu:
## Problem

A user on 0.5.10 reported a terminal **`Provider failure · Codex runtime binary is missing`** while their agent was **running fine** — codex was installed and working, yet the runtime declared it missing and terminated the session.

### Root cause

Codex session bring-up (`createCodexClientWithBinaryFallback`, at every `start`/`resume`) resolves a system `codex` on PATH and smoke-checks it with `spawnSync("codex","--version")` under a **3s** ceiling. **Every** failure of that check was wrapped into `formatCodexBinaryMissingMessage(...)` → `error-taxonomy.classify()` permanent `codex_binary_missing` → `capability` → `needs_operator` → **terminal, no retry**.

So a single slow/cold `codex --version` (busy host, an `nvm`/`fnm`/`volta`/`mise`/`asdf` shim, a cold start) exceeding 3s → `ETIMEDOUT` → a permanent "go install codex" terminal failure, even though the next bring-up succeeds. 0.5.10 made this user-visible by moving codex failure discovery into the in-chat session-run path.

## Fix

Distinguish a **transient verify flake on a present binary** from a **genuinely absent/broken** one:

- `verifyCodexExecutable` reports `transient` on failure: timeout (`ETIMEDOUT`) / termination-or-timeout kill signal / `EAGAIN` / `ENOMEM` / `ETXTBSY`. Ceiling raised **3s → 10s**.
- **Deterministic crash signals** (`SIGSEGV`/`SIGABRT`/`SIGILL`/`SIGBUS`/`SIGFPE`) and a **clean non-zero exit** stay **non-transient** → permanent binary-missing, so a genuinely broken binary still surfaces as actionable and never loops in retry.
- A present-but-flaky binary throws the new `CodexBinaryVerifyTransientError` (run path) / returns a transient-worded error (resolve path) instead of "binary missing".
- `error-taxonomy.classify()` maps `CodexBinaryVerifyTransientError` → **transient** (fast retry), landing in session-manager's existing `decideProviderRetry` retry loop instead of terminating.

### Related gap closed in the same audit

An `AbortSignal.timeout()` `DOMException` (`name: "TimeoutError"`, "The operation was aborted due to timeout") was unrecognised and fell into the slow `unknown` bucket. It now classifies as transient `operation_timeout`, consistent with the hub-fetch path in `sdk.ts`. (Same backend-timeout the user saw in a first screenshot, which the system already auto-recovered from.)

## Tests

- `codex-binary.test.ts`: transient flake → `CodexBinaryVerifyTransientError`, not binary-missing; non-transient exit → binary-missing; direct `verifyCodexExecutable` — self-`SIGSEGV` stays non-transient, self-`SIGTERM` is transient.
- `capability-probes.test.ts`: transient path-verify → not "binary missing"; non-transient → binary-missing.
- `error-taxonomy.test.ts`: `CodexBinaryVerifyTransientError` → transient; `TimeoutError` (name + text) → `operation_timeout`.
- `provider-retry-policy.test.ts`: end-to-end — verify flake at `session_start` and a codex `AbortSignal.timeout` → `transient_transport` + `action: retry`; genuinely missing binary stays `needs_operator` terminal.

Full `@first-tree/client` suite: **1160 passed**. `typecheck` + `biome` clean.

## Scope / not changed

Claude TUI `claude_login_required` is genuinely permanent (correct) and untouched. The bundled-binary launch path is not the live concern (First Tree does not bundle codex).

## Context Tree

Paired tree PR recording the decision: agent-team-foundation/first-tree-context#631.

### DIFF
--- a/packages/client/src/__tests__/capability-probes.test.ts
+++ b/packages/client/src/__tests__/capability-probes.test.ts
@@ -505,14 +505,37 @@ describe("resolveCodexRuntimeBinary (handler-contract parity)", () => {
     });
   });
 
-  it("bundle NOT found + PATH codex fails validation → binary-missing", async () => {
-    const verifyPath = (): CodexExecutableVerification => ({ ok: false, reason: "`codex --version` timed out" });
+  it("bundle NOT found + PATH codex non-transient validation failure → binary-missing", async () => {
+    const verifyPath = (): CodexExecutableVerification => ({
+      ok: false,
+      transient: false,
+      reason: "`codex --version` exited 1: broken shim",
+    });
     const res = await resolveCodexRuntimeBinary(
       {},
       { resolveBundled: notFound, findOnPath: () => "/usr/local/bin/codex", verifyPath },
     );
     expect(res.ok).toBe(false);
     if (!res.ok) expect(res.error).toContain("Codex runtime binary is missing");
+  });
+
+  it("bundle NOT found + PATH codex TRANSIENT validation flake → NOT binary-missing", async () => {
+    const verifyPath = (): CodexExecutableVerification => ({
+      ok: false,
+      transient: true,
+      reason: "`codex --version` timed out",
+    });
+    const res = await resolveCodexRuntimeBinary(
+      {},
+      { resolveBundled: notFound, findOnPath: () => "/usr/local/bin/codex", verifyPath },
+    );
+    expect(res.ok).toBe(false);
+    // A flaky smoke check on a present binary must not tell the operator to
+    // reinstall codex.
+    if (!res.ok) {
+      expect(res.error).not.toContain("Codex runtime binary is missing");
+      expect(res.error).toContain("transient host condition");
+    }
   });
 
   it("bundle NOT found + no PATH codex → binary-missing", async () => {
--- a/packages/client/src/__tests__/codex-binary.test.ts
+++ b/packages/client/src/__tests__/codex-binary.test.ts
@@ -3,6 +3,7 @@ import { tmpdir } from "node:os";
 import { delimiter, join } from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
+  CodexBinaryVerifyTransientError,
   createCodexClientWithBinaryFallback,
   findCodexExecutableOnPath,
   formatCodexBinaryMissingMessage,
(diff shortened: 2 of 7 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1329

- Reason: the transient split is directionally right, but the signal branch is too broad and can turn deterministic binary crashes into unbounded session retries.

**Level**: A
- Semantic lift: none

**PR Summary**
- Problem: a slow/flaky `codex --version` check on a present system binary was being reported as permanent `codex_binary_missing`, terminating the session instead of retrying.
- Design: add a transient verification result/error for PATH Codex smoke-check flakes, map it through error taxonomy/provider retry, and classify `AbortSignal.timeout()` as transient.
- Affected modules: Codex binary resolution, capability/runtime binary helper, provider error taxonomy, retry-policy tests.

**Review Notes**
❌ 1. `verifyCodexExecutable` now treats every `result.signal` from `codex --version` as transient. That covers timeout kills, but it also covers deterministic crash signals such as `SIGSEGV`, `SIGABRT`, or `SIGILL` from a broken native binary. Because `session_start`/`session_resume` transient failures retry without a terminal attempt cap, a broken install that always crashes on `--version` would no longer surface as an actionable binary failure and can sit in provider retry indefinitely. Please restrict this branch to timeout-equivalent kill signals/metadata, or otherwise keep crash signals non-transient, and add a test for a signal crash staying non-transient. [R4 / packages/client/src/runtime/codex-binary.ts:153]

✅ The main taxonomy ordering looks right: `CodexBinaryVerifyTransientError` wins before binary-missing detection, and the new tests cover the injected transient/non-transient paths plus provider retry classification.

**Executed**
- Submitted request_changes.

yuezengwu:
Thanks — R4 is a real gap, fixed in 648664e3.

`verifyCodexExecutable` no longer treats every `result.signal` as transient. Deterministic crash signals (`SIGSEGV`, `SIGABRT`, `SIGILL`, `SIGBUS`, `SIGFPE`) now stay **non-transient** → permanent `codex_binary_missing`, so a binary that always faults on `--version` surfaces as an actionable failure instead of looping through `session_start`/`session_resume` retries forever. Only termination/timeout-style kills (the `SIGTERM`/`SIGKILL` a `spawnSync` timeout uses, an OOM kill, an external shutdown) remain transient.

Added two direct `verifyCodexExecutable` tests: a self-`SIGSEGV` script stays non-transient (`transient: false`, reason mentions SEGV); a self-`SIGTERM` script stays transient. Full `@first-tree/client` suite: 1160 passed.

baixiaohang (approved):
## Review by codex-assistant - PR #1329

Approved after follow-up.

The R4 gap from the earlier review is addressed in 648664e3: deterministic crash signals from `codex --version` now stay non-transient, while timeout/host-pressure failures on a present binary remain retryable. Missing/broken binary and transient verify flake now route to the correct provider-failure outcomes.

I did not run local tests; remote CI is green at the reviewed head.

PR #1335, opened 2026-06-29
feat: be proxy-compatible without managing the user's proxy

yuezengwu:
## Why

A China-network user's daemon repeatedly failed all agents with
`authentication_failed / API Error: 403 Request not allowed`, while their
interactive `claude` worked. Root cause (diagnosed by @Gandy2025): the daemon
had **lost its proxy**. The launchd/systemd daemon does not inherit the login
shell, so its child `claude` process connected to `api.anthropic.com` directly
and was blocked **before authentication** — a 403 that looks like auth.

It kept recurring because the service unit **baked a point-in-time proxy
snapshot** from `process.env` (`collectProxyEnv`). On auto-update the
proxy-less supervisor re-rendered the unit, drift detection saw no change, and
the no-proxy state **froze permanently** (recurred 06-08 / 06-09 / 06-29).

## Design principle

Per product direction: First Tree should only be **compatible with** various
proxies, not **actively manage / configure** a proxy for the user. The
auto-snapshot (and any auto-mirror to `proxy.json`) is exactly the "managing"
overreach that caused the freeze — so it is removed at the root.

## What changed

**A — Compatibility (automatic)**
- `renderPlist` / `renderSystemdUnit` no longer read `process.env` proxy. The
  unit text is now stable → the drift-freeze class is gone at the root.

**B — User-owned config (opt-in, the user owns it)**
- The daemon loads `~/.first-tree/daemon.env` (`KEY=VALUE`) at start and passes
  it to every child it spawns (Claude CLI, git, npm). First Tree **only reads**
  this file; it never writes the user's proxy on their behalf.
  *Fill-don't-clobber*: a live env value is never overridden. Defensive parse —
  a malformed line can't crash the daemon.

**C — Diagnose, don't mislead**
- A 403 `Request not allowed` no longer surfaces the unconditional
  "run `claude auth login`". It enumerates the real causes in order:
  (1) daemon proxy missing → `daemon.env`, (2) plan/region entitlement,
  (3) genuine auth. Retry classification is intentionally unchanged (the string
  is ambiguous; informing beats auto-retrying).

**Upgrade buffer (decision: one-time migration)**
- On install/refresh, a proxy baked into the *previous* unit is lifted into
  `daemon.env` once; afterwards the file is the user's. No prior unit / no baked
  proxy / existing `daemon.env` → no-op.

**Docs**: `docs/cli-reference.md` (daemon.env) + `docs/troubleshooting/proxy.md`.

## Deliberately out of scope (follow-ups)

- **first-tree's own `fetch` honoring proxy** (undici ignores proxy env without
  `NODE_USE_ENV_PROXY` / an `EnvHttpProxyAgent`). Not on the blocked path today
  (the FT server is reachable), and enabling it would change FT-traffic routing
  — deferred to avoid an unforced behavior change. Subprocess egress (the actual
  failure) is fully covered by B.
- **SOCKS-only / TUN / system-proxy / PAC**: documented guidance rather than
  auto-configuration — consistent with "compatible, not managing".

## Tests
- `daemon-env.test.ts` (parse + fill-don't-clobber + no-op).
- service-install: upgrade-migration writes `daemon.env`, never overwrites an
  existing one, unit no longer carries proxy.
- claude-provider-error: egress-403 → enumerated notice; genuine 401 → auth lead.
- Full suites green: CLI 519, client 1162. `pnpm check` + `typecheck` clean.

A context-tree PR recording the "compatible, not managing" decision will be
cross-linked.

Immediate recovery for affected machines is the existing ops path
(`daemon refresh-unit` with proxy env, then `daemon stop && start`) — unaffected
by this PR.


yuezengwu:
Context-tree PR recording the "compatible, not managing" decision: agent-team-foundation/first-tree-context#636

baixiaohang (changes requested):
## Review by codex-assistant - PR #1335

- Reason: the daemon/service direction is sound, but the 403 diagnostic fix does not cover the full Claude SDK handler surface that currently emits the misleading auth advice.

**Level**: A
- Semantic lift: none beyond the path classification

**PR Summary**
- Problem: daemon-spawned provider/git/npm processes can lose the operator's proxy because launchd/systemd do not inherit the login-shell env, and the previous unit-level proxy snapshot could freeze a bad no-proxy state across updates.
- Design: stop baking proxy env into launchd/systemd units; load a user-owned `daemon.env` into the daemon at start; one-time migrate prior baked proxy values into that file; adjust Claude 403 `Request not allowed` guidance away from unconditional re-auth; document proxy troubleshooting.
- Affected modules: `apps/cli/src/core/service-install.ts`, `apps/cli/src/core/daemon-env.ts`, `apps/cli/src/commands/daemon/start.ts`, `packages/client/src/handlers/claude-provider-error.ts`, docs/tests.

**Review Findings**
❌ 1. Cover the handler-level 403 path, not only the final formatter. [R4 / `packages/client/src/handlers/claude-code.ts:1524`, `packages/client/src/handlers/claude-provider-error.ts:42`]

The new egress-aware lead is selected only inside `formatClaudeProviderFailureNotice`, but the Claude handler still emits the older auth hint before result settlement whenever it sees `assistant.error === "authentication_failed"` or an `auth_status.error`. For an SDK sequence that starts with the typed `authentication_failed` signal and later carries `API Error: 403 Request not allowed` in the result, the chat timeline still gets the misleading "run `claude auth login`" hint first; the later final notice cannot undo that.

There is a second bypass in the result adapter: for non-`success` result subtypes, `claudeFailureFromSdkResult` prefers `errors.join("; ")` over `result`. A result shaped like `{ subtype: "error_during_execution", errors: ["authentication_failed"], result: "Failed to authenticate. API Error: 403 Request not allowed" }` produces the preview `authentication_failed`, so `looksLikeEgressForbidden` never sees `Request not allowed` and the final notice falls back to the generic auth lead. That mixed surface is exactly the kind of `authentication_failed / API Error: 403 Request not allowed` failure this PR is trying to stop misdiagnosing.

Please move the `Request not allowed` special-case far enough upstream to suppress both auth-hint emission and final-notice selection, or preserve/merge the API detail before classification/notice formatting. Add a handler-level regression for the mixed SDK shape so the user-visible output contains the egress/proxy ordering and does not contain the unconditional `claude auth login` lead.

✅ 2. The daemon/service side matches the stated product boundary: newly rendered units are proxy-free, the daemon loads `daemon.env` before spawning children, and the migration is one-time/non-clobbering. I also checked the companion context-tree PR `agent-team-foundation/first-tree-context#636`; it records the same "compatible, not managing" decision on the runtime daemon node.

**Executed**
- Did not run tests or checks, per PR Review mode.

yuezengwu:
Thanks — both bypasses are real. Addressed in fc53ca8:

**1a — handler-level early hint.** The auth hint is no longer emitted eagerly on the typed `authentication_failed` / `auth_status` signal. It is **deferred** and resolved at result settlement (or stream end): emitted only for a genuine credential failure, **suppressed** when the settled preview carries the egress signature. The `auth_status` egress signature is dropped at capture time.

**1b — dropped API detail.** `claudeFailureFromSdkResult` now merges the typed `errors` with the result text for non-success subtypes, so `403 Request not allowed` survives into the preview instead of being hidden behind the opaque `authentication_failed` code. `isEgressForbiddenText` is shared between the handler and the notice formatter so both make the same call.

**Regressions added** (`claude-code-provider-error-result.test.ts`): the mixed shape (typed signal + 403 result) and the non-success-subtype shape both assert the user-visible output leads with egress/proxy guidance (`before authentication` / `daemon.env`), contains no `rejected the local Claude authentication`, and emits no early auth hint. Full client suite green (1165).

Note: retry *classification* is intentionally still unchanged (the string can't reliably separate egress / entitlement / auth) — only the user-facing advice ordering is corrected, which is the misdiagnosis this targets.

baixiaohang (approved):
## Review by codex-assistant - PR #1335 final

**建议**: approve
- 理由: 前序 R4 blocker 已在 `fc53ca8` 修复，当前实现符合 Context Tree 中 “compatible, not managing” 和 egress-403-not-auth 的约束。

**分级**: A
- Path 初判: 含 `apps/cli/**` 和 `packages/client/**` -> A
- 语义升级: 无；路径已到 A

**PR 摘要**
- 解决问题: daemon 不继承登录 shell env，旧 service unit 又会快照并冻结 proxy 状态，导致代理环境丢失后 Claude/git/npm 子进程直接出网并把 edge-block 403 误呈现成 auth failure。
- 设计方案: rendered launchd/systemd unit 不再携带 proxy；daemon start 读取用户自有 `daemon.env` 并传给子进程；旧 unit 里的 baked proxy 仅一次性 non-clobber 迁移；Claude `403 Request not allowed` 的用户提示改为 egress/proxy 优先，同时不改变 retry classification。
- 受影响模块: `apps/cli/src/core/service-install.ts`, `apps/cli/src/core/daemon-env.ts`, `apps/cli/src/commands/daemon/start.ts`, `packages/client/src/handlers/claude-code.ts`, `packages/client/src/handlers/claude-provider-error.ts`, docs/tests。

**Review 意见**
✅ 1. 上轮 handler-level blocker 已修：typed `authentication_failed` / `auth_status` hint 延迟到 result settlement；若 settled preview 含 `Request not allowed`，早期 auth hint 被 suppress，只发 egress/proxy final notice。
✅ 2. 上轮 dropped-detail blocker 已修：non-success result subtype 会合并 `errors` 和 `result`，`Request not allowed` 不再被 opaque `authentication_failed` 遮掉。
✅ 3. 新增 handler regressions 覆盖 typed signal + 403 result、non-success subtype + mixed `errors/result`，断言 user-visible output 含 `before authentication` / `daemon.env` 且不含 generic auth lead。
✅ 4. daemon/service 侧仍符合 #636 tree 约束：unit proxy-free，`daemon.env` fill-don't-clobber，migration one-time/non-overwrite。
✅ 5. GitHub checks 当前全绿：Lint & Type Check、Test、CodeQL、Migrations、legacy-name guard 等通过。

**已执行**
- 已按 baixiaohang 授权提交 approve
- 未运行本地测试；按 PR Review 模式仅做静态 review，并参考作者验证与 GitHub checks。

PR #1311, opened 2026-06-26
[codex] Redesign onboarding for value-first GitHub access

Gandy2025:
## Summary

- Redesigns onboarding around a value-first agent chat instead of pre-chat platform setup.
- Keeps the main onboarding flow states as canonical 4-step journeys: `create-team -> connect-computer -> create-agent -> start-chat` and `join-team -> connect-computer -> create-agent -> start-chat`.
- Defaults fresh onboarding entry to the opening step (`create-team` / `join-team`) even when the server already sees a connected computer or usable agent; same-tab persisted progress can still resume later steps.
- Shows the prominent UI progress as 3 setup milestones (`create/join team -> connect computer -> create agent`), with `start-chat` as the payoff / launch screen rather than a fourth setup-progress segment.
- Unifies the opening step with the setup-step layout and settles the visible setup titles as `Create a First Tree team`, `Join the team`, `Connect this computer`, and `Create your first agent`.
- Clarifies the First Tree team concept on the opening step in direct language: a team is where you, your teammates, and your agents work together.
- Refines start-chat copy and bootstrap prose so the first chat starts from available context or asks for a project path, GitHub repo URL, or task without assuming repo access exists during onboarding.
- Aligns `/preview/onboarding` with the latest flow: the Flow gallery shows only the canonical onboarding path, while GitHub access states render as a separate Settings/task-time escalation surface instead of an onboarding setup step.
- Renames onboarding frontend code, preview/tests, and `first-tree-welcome` guidance away from product-level `kickoff` terminology: web-facing code now uses `start-chat`, while the server `/me/onboarding/kickoff` path and legacy telemetry remain compatibility contracts.
- Removes GitHub App installation / repo authorization / Context Tree setup from the mandatory onboarding path; start-chat no longer auto-starts tree setup.
- Adds runtime agent guidance and welcome-skill updates for strong `gh`-first GitHub behavior, admin/non-admin boundaries, and task-time GitHub App escalation.
- Extracts reusable computer/runtime and create/poll agent setup hooks under `features/agent-setup`, so `/onboarding` owns formal completion while `/quickstart` and future setup surfaces can reuse the lifecycle primitives.

## Impact

This PR now includes the first implementation pass for the lightweight onboarding flow. It remains draft while product/technical review continues.

## Context Tree

- Companion tree PR: agent-team-foundation/first-tree-context#622
[...]

PR #1337, opened 2026-06-29
fix: skip onboarding create-agent step when member already has a personal agent

Gandy2025:
Fast-follow to #1311. A codex review of #1311 flagged this [P2]: the value-first redesign makes fresh onboarding entry always land on the opening step and walk forward (`inferInitialStepIndex` ignores server readiness). So a member who already created their personal agent — e.g. a **refresh / new tab after the agent came online but before start-chat** — reaches `StepCreateAgent` again, where submitting would create a **duplicate agent** (the step had no existing-agent guard).

**Fix:** `StepCreateAgent` advances past itself (`goNext`) on entry when `currentOrgHasPersonalAgent` is already true — gated on idle phase + a one-shot ref so it never double-advances the normal create path (which advances via the flow's `onAgentOnline -> goNext`). Preserves the redesign's "walk from the opening step" behavior while preventing the duplicate.

Tests: two cases added to web-dom-interactions (skips when a personal agent exists; shows the form when it doesn't). Full file green, typecheck + lint:tokens + biome clean.

### DIFF
--- a/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
+++ b/packages/web/src/pages/__tests__/web-dom-interactions.test.tsx
@@ -1878,6 +1878,34 @@ describe("web DOM interaction coverage", () => {
     await unmountRoot(view.root);
   });
 
+  it("skips the create-agent step when the member already has a personal agent", async () => {
+    authMock.value = { ...authMock.value, currentOrgHasPersonalAgent: true };
+    const { StepCreateAgent } = await import("../onboarding/steps/step-create-agent.js");
+    const { flow, root } = await renderOnboardingDom(<StepCreateAgent />, {
+      activeStep: "create-agent",
+      agentPhase: "idle",
+    });
+    await flush();
+    // A fresh re-entry (refresh / new tab) after the agent already came online
+    // but before start-chat lands back here; advancing past the form keeps the
+    // member from creating a duplicate agent.
+    expect(flow.goNext).toHaveBeenCalledTimes(1);
+    expect(flow.createAgent).not.toHaveBeenCalled();
+    await unmountRoot(root);
+  });
+
+  it("shows the create-agent form when the member has no personal agent yet", async () => {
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1337

**Recommendation**: approve
- Reason: the change closes the duplicate-agent re-entry hole while preserving the value-first flow's opening-step entry model.

**Level**: C
- Semantic lift: none; the change is limited to onboarding step rendering/navigation and does not touch subscriptions, WebSocket frame parsing, or shared data-layer contracts.

**PR Summary**
- Problem: after the value-first redesign, a refresh or new tab after a personal agent came online but before start-chat could land on `StepCreateAgent` again and allow duplicate agent creation.
- Design: `StepCreateAgent` reads `currentOrgHasPersonalAgent` and, only while the local creation phase is idle, advances past itself once instead of rendering the create form.
- Affected modules: `packages/web/src/pages/onboarding/steps/step-create-agent.tsx`, `packages/web/src/pages/__tests__/web-dom-interactions.test.tsx`

**Review Feedback**
✅ 1. The guard uses the correct membership-scoped readiness bit (`currentOrgHasPersonalAgent`) and stays inside the create-agent step, so the route-level bounce-out contract still depends on the completion stamp rather than readiness alone. [`packages/web/src/pages/onboarding/steps/step-create-agent.tsx:52`]
✅ 2. The `agentPhase === "idle"` gate plus one-shot ref avoids racing the normal create path, where `useAgentCreation` moves through `creating`/`online` and the flow's `onAgentOnline` owns the advance. [`packages/web/src/pages/onboarding/steps/step-create-agent.tsx:61`]
✅ 3. The added DOM coverage checks both branches: already-has-personal-agent skips without calling `createAgent`, while no-personal-agent leaves the form path available. [`packages/web/src/pages/__tests__/web-dom-interactions.test.tsx:1878`]

**Executed**
- Submitted approve review.
- Did not run tests or lint locally, per PR Review mode.

PR #1339, opened 2026-06-29
fix: daemon.env channel path + parse/egress hardening (QA follow-up to #1335)

yuezengwu:
## Why

Follow-up to #1335 (merged as `01e9929`). The black-box QA pass landed **after**
#1335 was squash-merged, so the QA fixes never made it onto `main` — `main`
currently has the proxy-compat feature **with the F1 blocker still in it**. This
PR carries the QA fixes forward.

## What (from the QA report on #1335)

- **F1 (blocker)** — user-facing guidance hardcoded `~/.first-tree`, but the
  loader resolves the channel home. On staging/dev (the dogfood channels) a user
  following the in-chat hint / docs writes to a file the daemon never reads, so
  the 403 persists — the signature fix defeats itself off-prod. Adds a single
  source of truth `daemonEnvFile()` in `@first-tree/shared/config` (channel-aware);
  the in-product egress hint now prints the resolved path, and the CLI loader +
  docs use it. Docs spell out the per-channel path.
- **F2** — strip a trailing ` # comment` on an unquoted value (a `#` inside
  quotes / not whitespace-preceded is kept for URLs & passwords).
- **F3** — skip an empty `KEY=` instead of injecting an egress-disabling blank
  and reporting it as loaded.
- **F4** — require a `403`/`forbidden` co-marker alongside `request not allowed`
  so a non-403 credential error mentioning the phrase isn't mis-routed to egress.
- **Leak** — the success branch reset `authHintEmitted` but not `pendingAuthHint`,
  so a transient `auth_status` warning before a successful turn could leak a stale
  auth-login hint. Clear it on success.

## Tests
daemon-env (inline-comment strip, empty-skip), claude-provider-error (non-403
phrase stays auth), handler regression (auth_status warning + success leaks no
hint). cli + client affected suites green; shared rebuilt; typecheck + biome clean.

### DIFF
--- a/apps/cli/src/__tests__/daemon-env.test.ts
+++ b/apps/cli/src/__tests__/daemon-env.test.ts
@@ -24,6 +24,15 @@ describe("parseDaemonEnv", () => {
   it("skips malformed lines rather than throwing", () => {
     expect(parseDaemonEnv("not-an-assignment\n=novalue\n9BAD=x\nGOOD=ok")).toEqual({ GOOD: "ok" });
   });
+
+  it("strips a trailing ` # comment` from an unquoted value but keeps it inside quotes", () => {
+    expect(parseDaemonEnv("HTTP_PROXY=http://127.0.0.1:7897 # office proxy")).toEqual({
+      HTTP_PROXY: "http://127.0.0.1:7897",
+    });
+    // A '#' that is part of the value (quoted, or not whitespace-preceded) stays.
+    expect(parseDaemonEnv('HTTPS_PROXY="http://u:p#a@host:1"')).toEqual({ HTTPS_PROXY: "http://u:p#a@host:1" });
+    expect(parseDaemonEnv("ALL_PROXY=socks5://h#ash")).toEqual({ ALL_PROXY: "socks5://h#ash" });
+  });
 });
 
 describe("loadDaemonEnv", () => {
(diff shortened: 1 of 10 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1339

**Recommendation**: approve
- Reason: The change centralizes the daemon.env path behind a channel-aware shared helper and covers the parser, egress classification, and stale auth-hint regressions that the QA follow-up targets.

**Level**: A
- Path initial: contains `apps/cli/**` and `packages/client/**` -> A
- Semantic lift: none; path level is already highest.

**PR Summary**
- Problem: QA fixes from #1335 did not land on `main`, leaving the proxy-compat feature with a channel-path guidance blocker and related parsing / auth-egress edge cases.
- Design: Add `daemonEnvFile()` in `@first-tree/shared/config` as the shared channel-aware path source; route the CLI loader, service migration, Claude egress notice, and docs through that contract; harden unquoted trailing comments, empty env values, `request not allowed` egress detection, and success-time auth hint cleanup.
- Affected modules: `apps/cli/src/core/daemon-env.ts`, `packages/shared/src/config/*`, `packages/client/src/handlers/claude-*`, CLI/proxy docs.

**Review Feedback**
✅ 1. `daemonEnvFile()` remains lazy through `defaultHome()`, so the CLI loader and in-chat egress notice agree on the channel home instead of reintroducing a hardcoded prod path. [R1 / `packages/shared/src/config/resolver.ts:70`, `apps/cli/src/core/daemon-env.ts:21`, `packages/client/src/handlers/claude-provider-error.ts:142`]
✅ 2. Early auth-hint suppression and the final provider notice now share `isEgressForbiddenText()`, and the 403/forbidden co-marker reduces the chance of routing an ordinary credential error to proxy guidance. [R4 / `packages/client/src/handlers/claude-provider-error.ts:129`, `packages/client/src/handlers/claude-code.ts:1537`, `packages/client/src/handlers/claude-code.ts:1591`]
✅ 3. Clearing `pendingAuthHint` on a successful result closes the stale hint leak after a transient `auth_status` warning, with regression coverage in the handler test. [R4 / `packages/client/src/handlers/claude-code.ts:1700`]
⚠️ 4. Non-blocking follow-up candidate: duplicate env keys where a later empty `KEY=` follows an earlier non-empty value currently collapse to the empty parsed value and are skipped during load. Duplicate-key behavior is not part of the documented main path, so this does not block the PR. [R4 / `apps/cli/src/core/daemon-env.ts:55`, `apps/cli/src/core/daemon-env.ts:83`]

**Executed**
- Did not run tests locally, per PR Review mode.

PR #1346, opened 2026-06-29
feat: gzip-compress Context Tree snapshot responses

ggordonhall:
## What

Enables gzip/brotli response compression for the Context Tree snapshot endpoints (`GET /context-tree/snapshot`, both user- and org-scoped).

## Why

Profiling the slow context-tree **page** (not the CLI) with a real `Server-Timing` trace showed the load was ~3.08s = ~2.39s TTFB + **684ms content download**. The snapshot payload — all nodes + edges + changes + writes — is large, highly compressible JSON, and the server runs **no** response compression today: there's no `@fastify/compress` in deps or `app.ts`, and no in-repo reverse proxy that would add it. Content-download was the second-largest single chunk of the page load.

## How

- Add `@fastify/compress@^9` (Fastify 5 compatible).
- Register it **scoped to just the two snapshot routes** via encapsulated child scopes, so the rest of the API is untouched (minimal blast radius — the request that was asked about).
- Content-negotiated + threshold-gated (1 KB): small responses and clients without `Accept-Encoding` are unaffected. `@fastify/compress` defaults brotli quality to 4, avoiding the slow zlib max-quality default.

## Tests

`packages/server/src/__tests__/context-tree-snapshot-compression.test.ts` — route-level: asserts the snapshot response is `content-encoding: gzip` when the client accepts it, and left uncompressed otherwise (decoded body is still a valid, populated snapshot). Existing snapshot/timing/io suites pass (46 tests green locally).

## Notes / follow-ups

This addresses the content-download slice only. The larger find from the same trace was that **~1.7s of the 2.39s TTFB is in no Server-Timing span** (uninstrumented pre-handlers / serialization / DB), plus an uncached GitHub token mint (229ms) and a 305ms telemetry IO scan — tracked separately.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@autotelic/fastify-opentelemetry": "^0.23.0",
+    "@fastify/compress": "^9.0.0",
     "@fastify/cors": "^11.2.0",
     "@fastify/rate-limit": "^10.3.0",
     "@fastify/static": "^9.0.0",
--- a/packages/server/src/__tests__/context-tree-snapshot-compression.test.ts
+++ b/packages/server/src/__tests__/context-tree-snapshot-compression.test.ts
@@ -1,0 +1,109 @@
+import { execFile } from "node:child_process";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { promisify } from "node:util";
+import { gunzipSync } from "node:zlib";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import { createTestAdmin, useTestApp } from "./helpers.js";
+
+const execFileAsync = promisify(execFile);
+
+// Shared ref the mock factory reads at call time. The validated org-settings
+// binding only accepts HTTPS/SSH repo URLs, so we cannot point an org at an
+// on-disk tree through the public path. Instead we mock the snapshot service to
+// delegate to its REAL implementation against a local checkout — exercising the
+// real route + @fastify/compress wiring against a genuine, populated snapshot.
+const treeRef = vi.hoisted(() => ({ path: "" }));
+
+vi.mock("../services/context-tree-snapshot.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../services/context-tree-snapshot.js")>();
+  return {
+    ...actual,
+    getContextTreeSnapshot: (
+      _binding: Parameters<typeof actual.getContextTreeSnapshot>[0],
+      window: Parameters<typeof actual.getContextTreeSnapshot>[1],
+      options: Parameters<typeof actual.getContextTreeSnapshot>[2],
+    ) => actual.getContextTreeSnapshot({ repo: treeRef.path, branch: "main" }, window, options),
+  };
+});
+
+const getApp = useTestApp();
+
+async function git(cwd: string, args: string[]): Promise<void> {
+  await execFileAsync("git", args, { cwd });
+}
+
+function node(title: string, description: string): string {
+  return `---
+title: "${title}"
+description: "${description}"
+owners: ["alice", "bob"]
+---
+
+# ${title}
+
+${description} ${"Context detail line. ".repeat(8)}
+`;
+}
+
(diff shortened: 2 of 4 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1346

**建议**: request_changes
- 理由: PR 新增的压缩测试在 CI 中失败，当前不能证明 snapshot 响应已按预期 gzip 压缩。

**分级**: B-low
- Path 初判: 仅改 `packages/server/**` 与 `pnpm-lock.yaml`，未含 migrations / Prisma schema / chat-message-session-agent 核心 service -> B-low
- 语义升级: 无；改动集中在 Context Tree snapshot 路由响应层和测试，不改变 snapshot 数据语义

**PR 摘要**
- 作者 / repo: ggordonhall / agent-team-foundation/first-tree
- 解决问题: 降低 Context Tree snapshot 大 JSON 响应的下载成本。
- 设计方案: 在 user/org 两个 `/context-tree/snapshot` 路由外包一层 Fastify 子 scope，并只在该 scope 注册 `@fastify/compress`，阈值 1 KB。
- 受影响模块: `packages/server/src/app.ts`、server 测试、server package 依赖。

**Review 意见**
❌ 1. 新增测试当前在 CI 中失败，且失败点正是本 PR 要证明的行为：`content-encoding` 实际为 `undefined`，未压缩分支拿到的 snapshotStatus 也是 `unavailable` 而非 `active`。这说明测试里的 mock/fixture 没有让路由走到 populated snapshot 路径，当前 PR 既无法通过 CI，也没有有效覆盖 gzip 行为。请先修正测试 setup 或改用真实可解析的 org binding，再让 CI 通过。 [CI / `packages/server/src/__tests__/context-tree-snapshot-compression.test.ts:88`]

**已执行**
- 已提交 request_changes；未运行本地测试，只读取了远端 CI 日志。

baixiaohang (changes requested):
Found one blocking issue: the new compression spec fails in the full CI test topology.

`packages/server/src/__tests__/context-tree-snapshot-compression.test.ts` installs a `vi.mock("../services/context-tree-snapshot.js")`, but the server Vitest config runs fork workers with `isolate: false`. In the full suite, another test file can import `app.ts`/the snapshot routes before this file's mock is installed, so the route keeps the real `getContextTreeSnapshot`. That is exactly what the current CI Test job shows: the new test receives an `unavailable` snapshot, the payload remains below the compression threshold, and `content-encoding` is `undefined` instead of `gzip`.

Please make the test independent of module import order. For example, seed the org's Context Tree binding/data through the database/test fixture path instead of mocking a route dependency, or load the app under an isolated `vi.doMock`/`vi.resetModules` setup that cannot reuse a previously imported route module.

yuezengwu (changes requested):
Requesting changes for one blocker in the new test coverage.

The implementation goal is clear: register `@fastify/compress` only around the user/org Context Tree snapshot routes so large JSON snapshots can be gzip/brotli-compressed without changing the rest of the API. The route wiring looks scoped as intended, but the new test is not reliable in the server suite's actual Vitest topology.

`packages/server/src/__tests__/context-tree-snapshot-compression.test.ts` depends on `vi.mock("../services/context-tree-snapshot.js")` before `useTestApp()` imports the already-wired app/routes. The server config reuses fork workers with `isolate: false` (`packages/server/vitest.config.ts`), so in the full suite another test file can import `app.ts` / the snapshot route modules before this mock is installed. In that case the route keeps the real `getContextTreeSnapshot`, the test org has no populated binding, the snapshot is `unavailable`, the payload stays under the compression threshold, and `content-encoding` remains `undefined` at the assertion on line 88.

Local verification matches CI:

- `pnpm --filter @first-tree/server test -- context-tree-snapshot-compression.test.ts` failed with only the new compression spec failing: `content-encoding` was `undefined`, and the identity case received `snapshotStatus: "unavailable"`.
- `pnpm exec vitest run src/__tests__/context-tree-snapshot-compression.test.ts --no-file-parallelism --maxWorkers=1` passed, which supports the import-order/isolation diagnosis rather than proving the full-suite test is stable.

Please make this test independent of module import order. The cleanest path is to seed the org's Context Tree binding/fixture through the normal DB/test setup path, or otherwise reset/import the app under an isolated mock setup that cannot reuse previously imported route modules.

PR #1354, opened 2026-06-30
fix: transfer managed agents on self-service leave (#1353)

### LINKED ISSUE #1353: Leave team leaves pinned agents that block retiring user-owned clients
## Summary

When a user leaves a team via the self-service leave flow, non-human agents managed by that member can remain pinned to the user's client. Later, when the user tries to retire that client, `retireClient` rejects the operation because it still finds non-deleted agents pinned to the client. The user can no longer see or manage those agents through the old team, so they are stuck.

## Current behavior

- `POST /me/memberships/:memberId/leave` validates that the membership belongs to the caller, then calls `leaveOrganization`.
- `leaveOrganization -> deactivateMembership` only marks the membership as `left`, suspends the member's human mirror agent, and clears that human mirror's `clientId`.
- It does not transfer, suspend, delete, or unpin non-human agents managed by that member.
- `retireClient` refuses to delete a client while any `agents.client_id = clientId` row exists with `status != deleted`.
- After leaving the team, the user is no longer an active member of that org, so they cannot see or manage the org's agents/clients from normal UI paths.

Related code pointers:

- `packages/server/src/api/me.ts` — `POST /me/memberships/:memberId/leave`
- `packages/server/src/services/membership.ts` — `leaveOrganization` / `deactivateMembership`
- `packages/server/src/services/member.ts` — admin `deleteMember` already transfers managed non-human agents and clears `clientId`
- `packages/server/src/services/client.ts` — `retireClient` pinned-agent guard
- `packages/web/src/pages/clients.tsx` — admin-mode Computers page currently prefers org-scoped clients over the user's `/me/clients` list

## Why this matters

This creates a dangling ownership/pinning state:

1. User manages agents in Team A and those agents are pinned to their computer/client.
2. User leaves Team A.
3. The Team A agents remain active or otherwise non-deleted and still pinned to the user's client.
4. User later tries to retire the client.
5. Retirement fails with the pinned-agent guard, but the user can no longer reach the old team's agents to clean them up.

## Suggested backend direction

Align self-service leave with the safer parts of admin member removal:

- For non-human agents managed by the leaving member, transfer ownership to a fallback active admin in the same org and clear `clientId`.
- Recompute watchers for transferred agents.
- Force-disconnect and unbind those agents from runtime presence after the transaction.
- Keep the leaving member's human mirror behavior as-is: suspended, name retained, `clientId = null`.

Open backend edge case to decide:

- If there is no fallback active admin and the leaving member manages non-human agents, should leave be blocked with a clear error, or should those agents be automatically suspended/deleted/unpinned?

## Web UI interaction to design

Please consider the leave-team UX explicitly before implementation:

- Should clicking "Leave this team" show a second confirmation when the member manages agents?
- What exact impact copy should the modal show?
- Should the UI list the affected agents and computers?
- If the chosen backend behavior deletes, suspends, unpins, or transfers agents automatically, the modal should say that clearly before the user confirms.
- If the backend blocks leave without another admin, the UI should explain how to resolve it, e.g. add/keep another admin or delete/transfer agents first.
- Confirm whether "leave team" is allowed to automatically delete all agents owned by that member in the team, or whether transfer/unpin is the intended safer behavior.

## Acceptance criteria

- A user leaving a team cannot leave behind non-deleted agents pinned to their client in a way that blocks retiring the client.
- The old team's agents remain manageable by an active member/admin, or leave is blocked with an actionable error.
- The Computers page still lets users see and manage their own clients even when the current selected org/admin view would not include them.
- Server tests cover self-service leave with managed agents pinned to a client.
- Web tests cover the leave-team confirmation/impact flow and/or the Computers visibility fix.

Gandy2025:
## Summary

Fixes #1353. Self-service "leave team" left the non-human agents the departing member managed `active` and still pinned to the user's client. Because `retireClient` refuses to retire a client while any non-deleted agent is pinned — and a departed member can no longer reach the old team's agents — the user was deadlocked out of retiring their own computer.

This aligns leave with the admin-removal path (`deleteMember`), which already handled this correctly.

## Backend

`leaveOrganization` → `deactivateMembership` now, inside one transaction:

- Reassigns the member's **non-human** agents to a fallback active admin (`managerId = fallback`, `clientId = null`) and recomputes watchers per agent — mirroring `deleteMember`, with identical active-admin `FOR UPDATE` lock ordering so concurrent leave/remove in the same org serialize instead of deadlocking.
- Suspends the human mirror and flips the membership (unchanged behavior).
- After commit, force-disconnects + unbinds the transferred agents from runtime presence.

**Policy (intentional, narrow scope):** leave is **blocked with a 409** only when the member manages non-human agents **and** no other active admin exists to inherit them — the error tells the user to add another admin or delete the agents first. A member who manages no non-human agents may still leave even as the sole admin (the "an org should keep an admin" concern is a separate org-lifecycle question this PR does not take on). **Never auto-deletes** agents — transfer + unpin is the safe behavior.

Scope is strictly `managerId = memberId`: agents the same user still manages via **other** active memberships keep their `clientId` by design (they remain reachable).

## Web

- **Leave-team modal** (`team-switcher.tsx`): states the impact up front — agents you manage here are transferred to another admin and unpinned from your computers, and leaving is blocked if no other admin can inherit them. The blocked-path 409 message is surfaced inline.
- **Computers page** (`clients.tsx`): admin mode now also fetches `/me/clients` and unions it (deduped by id, org row wins) into the viewer's **own** block only, so a client whose agents lived in a team the user left stays visible and retirable even when the org-scoped admin view omits it. The "Team computers" block stays org-scoped, so no row is shown twice.

## Tests

- Server (`me-leave-managed-agents.test.ts`): transfer→unpin→`retireClient` unblocked; the no-fallback-admin block (and that nothing mutates); sole-admin-with-no-agents leave succeeds.
- Web: Computers visibility fix (admin whose org view omits an owned client still sees it under "Your computers") and the leave-modal impact copy.

All suites green locally: server 1605/1605, web 1047/1047; typecheck, `lint:tokens`, biome clean.

## Context Tree

The durable constraint (leave/removal reassigns managed non-human agents to a fallback admin + unpins; blocks when none) was missing from the tree's Member Lifecycle section. Tree PR: agent-team-foundation/first-tree-context#640

### DIFF
--- a/packages/server/src/__tests__/me-leave-managed-agents.test.ts
+++ b/packages/server/src/__tests__/me-leave-managed-agents.test.ts
@@ -1,5 +1,6 @@
 import { randomUUID } from "node:crypto";
-import { eq } from "drizzle-orm";
+import { and, eq, ne } from "drizzle-orm";
+import postgres from "postgres";
 import { describe, expect, it } from "vitest";
 import { agents as agentsTable } from "../db/schema/agents.js";
 import { members as membersTable } from "../db/schema/members.js";
@@ -119,6 +120,46 @@ describe("self-service leave: managed non-human agents", () => {
     expect(memberRow?.status).toBe("active");
   });
 
+  it("lets a sole admin leave after deleting their only non-human agent (tombstones do not block)", async () => {
+    const app = getApp();
+    const org = await freshOrg("leave-tombstone");
+    const soleAdmin = await createMember(app.db, org.id, {
+      username: `tomb-${randomUUID().slice(0, 8)}`,
+      displayName: "Sole Admin",
+      role: "admin",
+    });
+    const clientId = await seedClient(app, soleAdmin.userId, org.id);
+    const managed = await createAgent(app.db, {
+      name: `managed-${randomUUID().slice(0, 8)}`,
+      type: "agent",
+      displayName: "Managed Agent",
+      managerId: soleAdmin.id,
+      clientId,
+    });
+    // The user follows the no-fallback-admin 409's guidance and deletes the
+    // agent. deleteAgent only tombstones (status='deleted', managerId intact),
+    // and retireClient already ignores deleted agents — so leave must stop
+    // counting it, otherwise the sole admin is trapped forever.
+    await app.db.update(agentsTable).set({ status: "deleted", name: null }).where(eq(agentsTable.uuid, managed.uuid));
+
+    await leaveOrganization(app.db, soleAdmin.id);
+
+    const [memberRow] = await app.db
+      .select({ status: membersTable.status })
+      .from(membersTable)
+      .where(eq(membersTable.id, soleAdmin.id))
+      .limit(1);
+    expect(memberRow?.status).toBe("left");
+    // The tombstone is left untouched — not reassigned to a (nonexistent) admin.
+    const [managedRow] = await app.db
+      .select({ managerId: agentsTable.managerId, status: agentsTable.status })
+      .from(agentsTable)
+      .where(eq(agentsTable.uuid, managed.uuid))
+      .limit(1);
+    expect(managedRow?.status).toBe("deleted");
+    expect(managedRow?.managerId).toBe(soleAdmin.id);
+  });
+
   it("allows a sole admin with no managed agents to leave (narrow scope), suspending the human mirror", async () => {
     const app = getApp();
     const org = await freshOrg("leave-empty");
@@ -144,5 +185,63 @@ describe("self-service leave: managed non-human agents", () => {
     expect(mirror?.status).toBe("suspended");
     expect(mirror?.name).not.toBeNull();
     expect(mirror?.clientId).toBeNull();
+  });
+
+  it("does not strand an agent created concurrently with the manager's departure", async () => {
+    const app = getApp();
+    const org = await freshOrg("leave-race");
+    // A fallback admin exists so the leaver is not the sole admin — this test
+    // is about the create/leave race, not the no-fallback block.
+    await createMember(app.db, org.id, {
+      username: `race-fb-${randomUUID().slice(0, 8)}`,
+      displayName: "Fallback Admin",
+      role: "admin",
+    });
+    const leaver = await createMember(app.db, org.id, {
+      username: `race-lv-${randomUUID().slice(0, 8)}`,
+      displayName: "Leaver",
+      role: "admin",
+    });
+
+    // Raw second connection holds an *uncommitted* `status = 'left'` update on
+    // the leaver, taking the same member row lock that `deactivateMembership`
+    // would. createAgent's pre-transaction check still reads the committed
+    // `active` state, but its in-transaction `FOR UPDATE` re-check blocks on
+    // this lock; once we commit, the re-check sees the member is no longer
+    // active and aborts. Without the in-transaction re-check, the agent insert
+    // would block on the FK lock and then succeed once we commit — stranding an
+    // `active` agent on a departed manager. The outcome here is driven by the
+    // lock, not by sleep timing, so it is deterministic.
+    const raw = postgres(process.env.DATABASE_URL ?? "", { max: 1 });
+    try {
+      await raw`BEGIN`;
+      await raw`UPDATE members SET status = 'left' WHERE id = ${leaver.id}`;
+
+      const createOutcome = createAgent(app.db, {
+        name: `raced-${randomUUID().slice(0, 8)}`,
+        type: "agent",
+        displayName: "Raced Agent",
+        managerId: leaver.id,
+      })
+        .then(() => "created" as const)
+        .catch((err: unknown) => err);
+
+      // Give createAgent a moment to reach (and block on) the FOR UPDATE.
+      await new Promise((resolve) => setTimeout(resolve, 150));
+      await raw`COMMIT`;
+
+      const outcome = await createOutcome;
+      expect(outcome).not.toBe("created");
+      expect((outcome as Error)?.message ?? "").toMatch(/not found/i);
+
+      // No non-human agent was stranded on the departed manager.
+      const stranded = await app.db
+        .select({ uuid: agentsTable.uuid })
+        .from(agentsTable)
+        .where(and(eq(agentsTable.managerId, leaver.id), ne(agentsTable.type, "human")));
+      expect(stranded).toHaveLength(0);
+    } finally {
+      await raw.end();
+    }
   });
 });
--- a/packages/server/src/__tests__/me-leave-managed-agents.test.ts
+++ b/packages/server/src/__tests__/me-leave-managed-agents.test.ts
@@ -1,0 +1,148 @@
+import { randomUUID } from "node:crypto";
+import { eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import { agents as agentsTable } from "../db/schema/agents.js";
+import { members as membersTable } from "../db/schema/members.js";
+import { createAgent } from "../services/agent.js";
+import { retireClient } from "../services/client.js";
+import { createMember } from "../services/member.js";
+import { leaveOrganization } from "../services/membership.js";
+import { createOrganization } from "../services/organization.js";
+import { seedClient, useTestApp } from "./helpers.js";
+
+/**
+ * Issue #1353: self-service leave must align with admin member removal for the
+ * non-human agents the leaving member manages — otherwise those agents stay
+ * `active` and pinned to the user's client, and `retireClient`'s guard
+ * deadlocks the user out of retiring their own computer.
+ */
+describe("self-service leave: managed non-human agents", () => {
+  const getApp = useTestApp();
+
+  async function freshOrg(prefix: string) {
+    return createOrganization(getApp().db, {
+      name: `${prefix}-${randomUUID().slice(0, 8)}`,
+      displayName: "Leave Test Org",
+    });
+  }
+
+  it("transfers managed agents to a fallback admin, clears clientId, and unblocks retiring the client", async () => {
+    const app = getApp();
+    const org = await freshOrg("leave-transfer");
+    const fallbackAdmin = await createMember(app.db, org.id, {
+      username: `fb-${randomUUID().slice(0, 8)}`,
+      displayName: "Fallback Admin",
+      role: "admin",
+    });
+    const leaver = await createMember(app.db, org.id, {
+      username: `lv-${randomUUID().slice(0, 8)}`,
+      displayName: "Leaver",
+      role: "admin",
+    });
+
+    const clientId = await seedClient(app, leaver.userId, org.id);
+    const managed = await createAgent(app.db, {
+      name: `managed-${randomUUID().slice(0, 8)}`,
+      type: "agent",
+      displayName: "Managed Agent",
+      managerId: leaver.id,
+      clientId,
+    });
+
+    // Precondition: while the agent is active and pinned, retire is blocked.
+    await expect(retireClient(app.db, clientId)).rejects.toThrow(/still pinned/i);
+
+    await leaveOrganization(app.db, leaver.id);
+
+    // Managed agent moved to the fallback admin and unpinned.
+    const [managedRow] = await app.db
+      .select({ managerId: agentsTable.managerId, clientId: agentsTable.clientId, status: agentsTable.status })
+      .from(agentsTable)
+      .where(eq(agentsTable.uuid, managed.uuid))
+      .limit(1);
+    expect(managedRow?.managerId).toBe(fallbackAdmin.id);
+    expect(managedRow?.clientId).toBeNull();
+    expect(managedRow?.status).toBe("active");
+
+    // Leaver's membership is left; human mirror suspended, named, unpinned.
+    const [memberRow] = await app.db
+      .select({ status: membersTable.status })
+      .from(membersTable)
+      .where(eq(membersTable.id, leaver.id))
+      .limit(1);
+    expect(memberRow?.status).toBe("left");
+    const [mirror] = await app.db
+      .select({ status: agentsTable.status, name: agentsTable.name, clientId: agentsTable.clientId })
+      .from(agentsTable)
+      .where(eq(agentsTable.uuid, leaver.agentId))
+      .limit(1);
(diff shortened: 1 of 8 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1354

**建议**: request_changes
- 理由: self-service leave 的新保护会把已删除的 agent tombstone 也算作“managed agent”，导致用户即使按错误提示删除 agents 后仍可能无法离开。

**分级**: B-high
- Path 初判: 改动 `packages/server/src/services/membership.ts`，命中 `packages/server/**` 且 `services/*membership*` 属于成员/agent 生命周期核心服务 -> B-high
- 语义升级: 无

**PR 摘要**
- 解决问题: self-service leave 时，离开成员管理的非 human agents 之前会继续 active 且 pinned，导致该用户之后无法 retire 自己的 client。
- 设计方案: `leaveOrganization` 复用 admin-removal 的思路，在事务内把离开成员管理的非 human agents 转给 fallback active admin 并清 `clientId`，提交后断开 runtime presence；Web 同步补充 leave modal 文案和 Computers 页 `/me/clients` union。
- 受影响模块: `packages/server/src/services/membership.ts`、`packages/server/src/__tests__/me-leave-managed-agents.test.ts`、`packages/web/src/components/team-switcher.tsx`、`packages/web/src/pages/clients.tsx`

❌ 1. Deleted agents still block the “delete agents first, then leave” recovery path. `managed` selects every non-human row with `managerId = memberId`, and the no-fallback-admin 409 is driven by `managed.length > 0`. But `deleteAgent` only marks a row `status = 'deleted'` and leaves `managerId` intact; `retireClient` already treats deleted agents as non-blocking. So a sole admin who follows the new error guidance and deletes their agents can still be stuck forever: the deleted tombstone is counted as managed, there is still no fallback admin, and leave keeps returning 409. Please scope both the preflight and transfer update to non-deleted agents (suspended still matters; deleted should not), and add a regression where a sole admin suspends/deletes their only non-human agent and can then leave. [R5 / `packages/server/src/services/membership.ts:294`]

✅ 2. The main active-agent transfer path otherwise matches the existing admin-removal shape: deterministic active-admin lock ordering, transactional manager/client update, watcher recompute, human mirror suspension, and post-commit runtime unbind.

**已执行**
- 已提交 request_changes review。
- 未运行本地测试；远端 GitHub checks 当前全绿。

yuezengwu (changes requested):
这次改动目标是把 self-service leave 和 admin removal 对齐：成员离开时转移其管理的非 human agents、清掉 client pin、重算 watchers，并在 Web 上说明影响；Computers 页同时让用户自己的 cross-org clients 保持可见。方向我认可，但当前实现还有必须修的生命周期漏洞。

Findings:

1. `packages/server/src/services/membership.ts:294` — `managed` 统计了所有非 human rows，没有排除 `status = 'deleted'`。`deleteAgent` 只把 row 标成 deleted 并清 `name`，不会清 `managerId`；而 `retireClient` 已明确把 deleted agents 视为不应阻塞的 tombstones。结果是 sole admin 按 409 提示“delete these agents, before leaving”操作后，deleted tombstone 仍被 `managed.length > 0` 算进去，且没有 fallback admin，于是 leave 永远继续 409。这里应把 preflight 和 transfer update 都限定到 non-deleted agents（suspended 可以继续算，deleted 不应算），并补一个 sole admin suspend/delete 最后一个非 human agent 后可以 leave 的回归测试。

2. `packages/server/src/services/membership.ts:281` / `packages/server/src/services/agent.ts:427` — 这段事务没有真正防住 concurrent `createAgent` 在 leave 扫描后插入新 managed agent。`deactivateMembership` 虽然 `FOR UPDATE` 锁住 member row，但 `createAgent` 的 manager 校验是普通 SELECT；在 READ COMMITTED 下它可以先读到旧的 active member，然后在 FK/insert 阶段等 leave 提交，最后仍然插入一个 `managerId = leftMember` 的 agent（status 不属于 FK 条件，不会被重查）。这会重新制造本 PR 要消除的 stranded managed-agent 状态。需要让 create/update manager 路径和 leave/remove 使用同一把锁并在等待后重查 active 状态，或者在 leave 事务里用能覆盖 insert 的机制关掉这个窗口。

Context Tree note: 这个 PR 改变了现有 `runtime-binding.md` 的 “non-human agent client pin is set once and not cleared/moved” 约束。PR body 里的 `Tree PR: __TREE_PR__` 还没有落到真实链接；这属于成员生命周期/runtime binding 的核心策略变更，需要在合并前同步树里的 durable constraint。

我没有本地跑完整测试；当前 worktree 缺少已安装依赖，`pnpm --filter @first-tree/server exec vitest ...` 找不到 `vitest`。GitHub CI 当前显示 green，但上述路径没有被现有测试覆盖。

yuezengwu:
Correction to my review note: I re-checked the current PR body and the Context Tree PR is now linked as `agent-team-foundation/first-tree-context#640`. Please treat my earlier `Tree PR: __TREE_PR__` note as stale.

The `CHANGES_REQUESTED` state still stands for the two lifecycle findings: deleted tombstones are counted as managed agents, and the concurrent `createAgent` window can still strand an agent on a leaving member.

Gandy2025:
Thanks both — both lifecycle findings are fixed in `573e8d3`.

**1. Deleted tombstones counted as managed (codex R5 / yuezengwu #1).** Confirmed and fixed. `deleteAgent` only sets `status='deleted'` (keeping `managerId`), and `retireClient` already treats tombstones as non-blocking — so counting them trapped a sole admin who followed the 409's "delete these agents first" guidance. The preflight count **and** the transfer update are now scoped to non-deleted agents via a shared `managedFilter` (`ne(agents.status, 'deleted')`); suspended agents still count and still transfer/unpin. Regression added: a sole admin deletes their only non-human agent and can then leave (the tombstone is left untouched, not reassigned).

**2. Concurrent `createAgent` window (yuezengwu #2).** Confirmed and fixed at the shared `createAgent` path, so the same window in admin-removal is closed too. `createAgent` now re-confirms the manager is `active` under `SELECT … FOR UPDATE` **inside its insert transaction**, taking the same member-row lock that `deactivateMembership` and `deleteMember` already hold. The two paths are now mutually exclusive: a create that began while the member was active either commits first (so the departure's scan reassigns/unpins it) or blocks until the departure commits and then aborts on the now-inactive manager. Deadlock-safe — `createAgent` only ever takes that single member lock, so it cannot be in a wait cycle with the departure paths. Human-mirror bootstrap is skipped (member row doesn't exist yet; deferred FK validates). Regression added: a deterministic two-connection test (a raw conn holds an uncommitted `status='left'` on the manager) — verified it **fails without the fix** (the create succeeds and strands an agent) and passes with it.

**Context Tree (yuezengwu #3).** The Tree PR (agent-team-foundation/first-tree-context#640) now also amends `runtime-binding.md` → Binding Immutability to record the one administrative event that clears a pin (manager departure → detach to `unpinned`), explicitly distinguished from the still-forbidden running Client-to-Client hand-off, cross-linked to the team member-lifecycle node. (And confirmed the `Tree PR` link in the body resolved — thanks for the correction.)

Full server suite green locally (1607). Ready for another look.

baixiaohang (changes requested):
Thanks for the follow-up. The deleted-tombstone issue is fixed in this head: the leave preflight and transfer update now both exclude `status='deleted'`, and the regression covers the sole-admin deleted-agent recovery path.

One lifecycle race remains:

1. `packages/server/src/services/agent.ts:1047` — `updateAgent` still allows an admin `PATCH /agents/:uuid` to reassign `managerId` after a plain non-locking active-member read. That leaves the same create/leave race open for existing agents: a manager reassignment can read the target member as active, `deactivateMembership` / `deleteMember` can then lock the member and scan that member's currently managed agents before this reassignment is applied, and the PATCH can finally set `managerId = leftMember` after the departure commits. The result is again an active non-human agent managed by an inactive member, potentially still carrying its existing `clientId`, which is the stranded lifecycle state this PR is trying to eliminate. The `createAgent` fix added the right in-transaction `SELECT ... FOR UPDATE` active re-check; the manager reassignment path needs the same discipline, ideally in the same transaction as the agent update + watcher recompute. Please add a regression for concurrent `PATCH managerId` vs leave/remove, analogous to the new create race test.

Verification: `corepack pnpm --filter @first-tree/server typecheck` passes locally. I did not rerun DB-backed server tests in this environment because the Postgres test harness still needs a container runtime here.

yuezengwu (changes requested):
Re-reviewed latest head `573e8d3`. The PR's goal remains to align self-service leave with admin removal: transfer the departing member's managed non-human agents to an active admin, clear their client pins, recompute watcher state, and keep the user's own computers visible in Web. The deleted-tombstone fix looks correct, and the new `createAgent` in-transaction `FOR UPDATE` active-manager recheck closes the original create/leave race for newly-created agents.

Findings:

1. `packages/server/src/services/agent.ts:1047` — `updateAgent` still has the same stale-active-member window for existing-agent manager reassignment. It validates the target `managerId` with a plain non-locking SELECT outside the final update, then later runs an unconditional `UPDATE agents`. A concurrent leave/remove can lock that member, scan the member's currently-managed agents before this reassignment is applied, commit the member inactive, and then this PATCH can still set `managerId = leftMember`. That recreates an active non-human agent managed by an inactive member. This path needs the same in-transaction `SELECT ... FOR UPDATE` active-manager recheck as `createAgent`, with the agent update and watcher recompute made atomic enough that the stale reassignment cannot land after departure. Please add a concurrent `PATCH managerId` vs leave/remove regression.

2. `packages/server/src/services/agent.ts:1071` and `packages/server/src/api/agent/ws-client.ts:1029` — the first-bind paths can also re-pin an agent after leave has transferred/unpinned it. Both HTTP PATCH `clientId` and WS `agent:bind` decide ownership from a pre-leave read, then later set `agents.clientId` without rechecking that the same manager is still active/current. If leave wins the race and sets `managerId = fallback, clientId = null`, the delayed first-bind can still write the departed user's client id back onto that transferred agent. That brings back the exact `retireClient` deadlock class, now with the agent managed by the fallback admin but pinned to the departed user's computer. The claim/update should be conditional on the current manager/member/owner relationship or serialized under the same departure-safe locking discipline, and should have a concurrent first-bind vs leave regression.

Verification: GitHub checks are green on this head. I could not run local vitest in this worktree because dependencies are not installed (`pnpm --filter @first-tree/server exec vitest ...` cannot find `vitest`).

Gandy2025:
Both remaining lifecycle races are addressed in `88401a6`.

**1. `updateAgent` manager reassignment (baixiaohang #1 / yuezengwu #1).** The reassignment now happens inside a transaction that locks the target member `FOR UPDATE` and re-confirms `active` before applying the update + watcher recompute — same in-transaction discipline as `createAgent`, and crucially in the same **member→agent** lock order leave/`deleteMember` use, so the two serialize without deadlocking. A concurrent departure either commits first (its scan then transfers this agent) or blocks until the reassignment commits; the reassignment aborts if the target is no longer active. Regression added: concurrent `updateAgent({managerId})` vs a leaving target (deterministic via a raw-connection row lock — verified it strands nothing).

**2. First-bind re-pin after transfer (yuezengwu #2).**
- *HTTP `PATCH clientId`:* the first-bind client resolution is moved into the same locked transaction, gated on the agent's current manager being active. If leave transferred + unpinned the agent, the gating manager is no longer active and the bind aborts, so the departed owner's client is never written back. Regression added: concurrent first-bind vs a leaving manager (deterministic).
- *WS `agent:bind`:* the first-bind claim is now also conditioned on the `managerId` read at bind time (`…AND manager_id = $read`). Because leave/remove transfer by **changing** `managerId`, a transferred agent matches 0 rows → `WRONG_CLIENT` (the client repair path re-fetches authoritative state), instead of re-pinning the departed client; a claim that wins the race first is still re-scanned and unpinned by the departure's `managerId`-keyed transfer. No member lock is taken on the WS hot path.

On the WS regression specifically: a fully-committed transfer is already rejected earlier by the existing R-RUN owner check (`managerUserId === session.userId`), so the residual window is purely *between* the bind-time read and the claim — which is internal to the handler and not externally interposable, so a deterministic WS-handshake test isn't feasible and a sleep-based one would be flaky. The HTTP first-bind race (same invariant) is covered deterministically. Happy to add a WS-level test if you'd prefer one despite the flakiness tradeoff, or to extract the claim into a unit-testable helper.

Deadlock check: every path taking both locks acquires **member before agent** (leave, remove, create, update); the WS claim takes only the agent row lock — no cycle. Full server suite green locally (1609).

baixiaohang (approved):
Re-reviewed latest head `88401a65`. The previously blocking lifecycle races are addressed:

- Deleted tombstones are excluded from managed-agent leave preflight/transfer, while suspended agents still count.
- `createAgent`, `updateAgent({ managerId })`, and HTTP first-bind now re-check the gating manager under `SELECT ... FOR UPDATE` inside the mutation transaction, matching the member-before-agent lock order used by leave/removal.
- WS first-bind now claims only if the `managerId` read during bind is still current, so a concurrent leave/remove transfer that changes manager prevents stale re-pinning.
- New regression coverage exercises deleted tombstones, concurrent create, concurrent manager reassignment, and concurrent HTTP first-bind.

No blocking findings from my pass.

Verification: `corepack pnpm --filter @first-tree/server typecheck` passes locally. I could not run DB-backed server tests here because Docker/container runtime is unavailable in this environment; GitHub CI's Test job was still in progress when I checked, with completed checks green.

yuezengwu (approved):
Re-reviewed latest head `88401a65`. The remaining lifecycle blockers I called out are addressed.

Confirmed in this pass:

- `updateAgent({ managerId })` now re-checks and locks the target manager membership with `SELECT ... FOR UPDATE` inside the mutation transaction before updating the agent and recomputing watchers.
- HTTP first-bind now uses the same locked transaction discipline, gated on the current manager membership, so a concurrent leave/remove cannot re-pin a transferred agent to the departing user's client.
- WS first-bind now performs the NULL -> client claim only when the `managerId` read at bind time is still current, which closes the read/claim window without adding a member-row lock to the hot path.
- Regression coverage exists for deleted tombstones, concurrent create, concurrent manager reassignment, and concurrent HTTP first-bind.

No blocking findings from my pass. GitHub checks are green on this head.

PR #1356, opened 2026-06-30
Fix welcome eval setup option classification

baixiaohang:
## Summary

- Distinguish first-tree-welcome repo-entry input options from bounded first-task options.
- Keep `setup-as-first-task` strict by detecting setup/install/create/bind/seed/authorize language independently from the bounded task option count.
- Avoid treating natural-language `gh` mentions inside chat text as shell side effects while preserving real `gh_call` / shell command detection.
- Add regression coverage for pure repo-entry options, mixed repo-entry + setup options, and pure setup options.

## Validation

- `pnpm --filter @first-tree/skill-evals exec vitest run src/suites/first-tree-welcome/__tests__/grader.test.ts --maxWorkers=2 --minWorkers=1`
- `pnpm --filter first-tree-dev... build`
- `pnpm --filter @first-tree/skill-evals exec vitest run --passWithNoTests --maxWorkers=2 --minWorkers=1`
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-welcome/grader.ts packages/skill-evals/src/suites/first-tree-welcome/__tests__/grader.test.ts`
- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-welcome`
- `pnpm typecheck`
- `pnpm check` (exits 0; reports existing unrelated warnings/info in client/web files)
- `git diff --check`

## Notes

- This fixes the stage 3 false failure for `first-tree-welcome-no-repo-intro`: asking for a local clone path or GitHub URL is input collection, not a first task.
- I did not rerun live Codex gates after this grader-only fix.

### DIFF
--- a/packages/skill-evals/src/suites/first-tree-welcome/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-welcome/__tests__/grader.test.ts
@@ -65,6 +65,19 @@ function fixtureValidation(): FixtureValidation {
   };
 }
 
+function skillReadEvent(): unknown {
+  return {
+    event: {
+      item: {
+        command: "sed -n '1,220p' .agents/skills/first-tree-welcome/SKILL.md",
+        type: "command_execution",
+      },
+      type: "item.completed",
+    },
+    type: "codex_event",
+  };
+}
+
 describe("first-tree-welcome grader", () => {
   it("passes row 1 when the model routes tree kickoff to the tree setup lane", () => {
     expect(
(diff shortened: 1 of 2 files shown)


baixiaohang (commented):
Reviewed PR head `d4dd9d0` for the welcome grader setup-option classification fix.

No blocking issues found. The change keeps repo-entry input options such as local path / GitHub URL out of `taskOptionsObserved`, while separately detecting setup/install/create/bind/seed/authorize option text so `setup-as-first-task` still fails even when setup is mixed into otherwise valid input-collection options. Row 8's bounded task option condition remains tied to the 2-3 actual task options path.

I specifically checked the previous blocker: `[本地路径, GitHub URL, Create Context Tree]` now keeps `taskOptionsObserved=false` but still records `setup-as-first-task`, so row 3 fails correctly. Pure `[本地路径, GitHub URL]` still passes. The command side-effect regex also no longer treats chat prose like `gh/已有凭据` as a real `gh` invocation while keeping explicit model-phase `gh` calls and shell commands blocked.

I did not rerun tests as reviewer; I relied on code review plus the validation reported by the implementer.

Note: GitHub refused an approval from this CLI identity with `Review Can not approve your own pull request`, so this is submitted as a review comment rather than an approval.

yuezengwu (approved):
Reviewed PR head `d4dd9d0`.

No blocking issues found. The implementation separates repo-entry input options from real first-task options, and independently records setup/install/create/bind/seed/authorize option text so `setup-as-first-task` remains strict when setup is mixed into otherwise valid input collection.

I checked the previous failure shape: pure local path / GitHub URL options no longer set `taskOptionsObserved`, while `[local path, GitHub URL, Create Context Tree]` still fails via `setup-as-first-task`. The command side-effect regex also preserves explicit `gh_call` / shell command detection while avoiding prose-only `gh` mentions.

CI is green. I attempted the targeted local vitest command, but this review worktree did not have dependencies installed (`vitest` not found), so I relied on code review plus CI/author validation rather than rerunning locally.

PR #1352, opened 2026-06-29
feat(web): reusable quickstart growth entry (campaign-driven, fully automatic)

Gandy2025:
## What

A reusable **quickstart** growth entry. A landing page sends the browser to
`/quickstart?campaign=<slug>&repo=<github-url>`; after GitHub login the page
connects a computer and then runs to completion on its own — auto-creates a
private agent (**Cedar**) on the preferred runtime, starts the campaign's
value-first first chat, and drops the user into it. Two campaigns ship:
`production-scan` and `agent-readiness`. A future landing = one `CampaignConfig`
entry + a CTA pointing at `/quickstart?campaign=…`.

This is the refactor direction of #1308 (which predated the `features/agent-setup`
extraction and made campaign a welcome-skill lane), rebuilt on the new foundation.

## Flow (after login, the only manual step is connecting a computer)

`/quickstart` (full-screen, login-guarded, registered before `/onboarding`) →
`useComputerConnection` → auto `useAgentCreation` (Cedar, private,
Claude-Code-preferred runtime — no picker gate) → `onOnline` →
`postOnboardingStartChat` (kind `work`, `campaign` segment, `complete:false`,
clean dual-reader bootstrap) → refresh `/me` → navigate to `/?c=…`.

## Boundaries (must not pollute onboarding)

- **Idempotency key**: optional `campaign` segment appended to
  `chats.onboarding_kickoff_key` (`…:<kind>:<campaign>`). Onboarding omits it →
  byte-identical key, **no migration**. Without it, two campaigns reusing the
  same personal agent + `work` kind collide and the second chat is swallowed.
- **No completion stamp**: start chat passes `complete:false`; quickstart never
  writes `onboarding_completed_at` / suppress. Safe because the workspace
  auto-entry gate keys off having a personal agent + a connected client
  (refreshed before navigate), not the stamp.
- **Reuse, not rebuild**: only `features/agent-setup` hooks + presentational
  `flow-ui` primitives — no onboarding route state machine.
- **GitHub App not required** (task-time access); the **dual-reader bootstrap**
  stays clean user-facing welcome copy (no skill / operational jargon).

## Testing

- Campaign registry + landing-handoff parsing unit tests (16).
- Page-level **e2e/integration** (vitest `*.e2e.test.ts`, the repo convention
  since the browser harness was removed in #1336): full flow for **both
  campaigns** + connect / no-runtime / agent-timeout / create-failure /
  start-chat-failure / remount-reuse / stale-stash / missing-intent (15).
[...]

PR #1357, opened 2026-06-30
test(cli): relax flaky integration test timeouts (CPU-starved under parallel CI)

Gandy2025:
## Problem

A cluster of `apps/cli` tests spawn the command as a subprocess and do real file/process I/O. In isolation they run ~1.5s and the **whole package passes 522/522 when run alone** — but under the full-monorepo `turbo run test` (every package in parallel) CI runners get CPU-starved and an occasional, **varying** handful cross vitest's default **5000ms** timeout:

- `chat-send-message-file.test.ts`
- `chat-send-escaped-newline-guard.test.ts`
- `chat-send-escaped-newline-stderr.test.ts`
- `migrate-agent-dirs-extra.test.ts` (`createApiNameResolver`)

The failing set **differs run to run** → non-deterministic, parallel-load-driven flaky, not logic. The test files are byte-identical to their passing history.

## Fix

Raise the apps/cli package-wide `testTimeout` to **15s** (vitest config) so transient CPU starvation can't trip these I/O/subprocess tests. A genuine hang still fails, just at 15s. Package-level rather than per-test so the whole cluster + future CLI tests are covered. (Confirmed *not* a shared-port/resource race — these tests do no networking.)

## Why standalone

The fragility isn't owned by any feature PR. It was surfaced by #1352 (whose added web tests tipped the parallel run over the edge), but landing it on `main` keeps it from flaking other PRs — and keeps `main` itself green once #1352 merges.

Verified locally: the 4 tests pass with the new timeout; biome clean.

### DIFF
--- a/apps/cli/vitest.config.ts
+++ b/apps/cli/vitest.config.ts
@@ -4,6 +4,15 @@ import { unitCoverageConfig } from "../../scripts/vitest-coverage.js";
 
 export default defineConfig({
   test: {
+    // Many CLI tests spawn the command as a subprocess and do real file/process
+    // I/O. They run ~1-2s in isolation but are CPU-bound, so under the
+    // full-monorepo `turbo run test` (every package in parallel) CI runners get
+    // starved and an occasional, varying handful cross vitest's default 5s
+    // timeout — flaky red driven by an unrelated PR's added load, not by logic
+    // (they pass when this package runs alone). Raise the package-wide timeout
+    // so transient CPU starvation can't trip them; a genuine hang still fails,
+    // just at 15s.
+    testTimeout: 15_000,
     coverage: unitCoverageConfig({
       exclude: ["src/commands/types.ts"],
     }),


Gandy2025 (commented):
## Review — approve

Minimal, correctly-scoped fix: `apps/cli/vitest.config.ts` package-wide `testTimeout: 5s → 15s`, with a comment that accurately explains the cause (these CLI tests spawn subprocesses + do real file/process I/O; they run ~1–2s in isolation but get CPU-starved under the full-monorepo parallel `turbo run test`, so a varying handful occasionally cross the 5s default — flaky, not logic).

Verified the diagnosis holds: the failing set varied run-to-run (`escaped-newline-guard` vs `escaped-newline-stderr`) = non-deterministic, the whole apps/cli package passes 522/522 in isolation, and these files are byte-identical to main / don't import the web change that triggered the load. So raising the ceiling is the right pragmatic fix, not masking a real failure — a genuine hang still fails, just at 15s. CI here is green (incl. Test).

Package-level (vs per-test) is fine — it covers the whole flaky cluster + future cases; the only cost is a slower failure on a true hang, which the comment acknowledges. No blockers from me.

yuezengwu (approved):
Reviewed and approved.

Goal: stabilize `apps/cli` tests under full-monorepo parallel CI, where subprocess/file-I/O CLI tests can cross Vitest's 5s default under CPU starvation.

Core change: `apps/cli/vitest.config.ts` now sets a package-wide `testTimeout: 15_000` with an explanatory comment. This is scoped to test runtime configuration only; no product behavior, core data structures, or database/migration changes.

I did not find blockers. This matches existing repo practice where packages with heavier integration-style tests use package-level Vitest timeouts, and CI is green. Local verification: `apps/cli` Vitest suite passed (`65` files / `522` tests), and `pnpm exec biome check apps/cli/vitest.config.ts` passed.

baixiaohang (approved):
## Review by codex-assistant - PR #1357

**Recommendation**: approve
- Reason: This is a narrowly scoped test-runtime hardening for `apps/cli`; it raises the package Vitest timeout to cover CPU-starved subprocess/file-I/O tests without changing product behavior.

**Classification**: A
- Path initial: `apps/cli/vitest.config.ts` matches `apps/cli/**`.
- Semantic lift: none; this does not touch CLI runtime behavior, command registration, output contracts, data structures, or migrations.

**PR Summary**
- Problem: A varying subset of CLI integration-style tests can exceed Vitest's 5s default when the full monorepo test run starves CI CPU.
- Design: Set `testTimeout: 15_000` in the `apps/cli` Vitest config, with a comment documenting the subprocess/file-I/O load pattern and the slower true-hang failure tradeoff.
- Affected module: `apps/cli/vitest.config.ts`

**Review Notes**
1. No blockers found.
2. The package-level timeout matches the scope of the flaky cluster and is consistent with existing heavier package test configs in the repo.
3. The change does not relax assertions or alter production code; a real hang still fails, just at 15s.

**Executed**
- Did not run local tests in review mode; GitHub checks are green.

PR #1366, opened 2026-06-30
Fix runtime session reconcile active set

baixiaohang:
## Summary

- add an agent-scoped active runtime chat id endpoint for the current user's non-archived working set
- filter client full-state sync and reconcile to `(localHeldSet ∩ activeSet) ∪ forceKeepSet`, preserving queued/retry/unsettled delivery work
- refresh the active set on startup/bind/hourly jitter, optimistic-add newly delivered chats, and chunk reconcile frames at 500 ids

## Background

`session:reconcile` frames were rejected as malformed when a client held more than 500 local chat ids. The local held set can be larger than the wire schema limit because it combines live sessions and evicted mappings; one observed shape was `1 suspended + 500 evicted = 501`.

This keeps local runtime cache retention separate from the smaller periodic reconcile/full-state-sync projection. Archived chats are not deleted locally; clean idle archived sessions simply stop participating in periodic runtime projection until the chat is active again.

## Validation

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/agent-slot.test.ts src/__tests__/session-manager-edge-coverage.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/me-chat-service.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/agent-chats.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/shared typecheck && pnpm --filter @first-tree/client typecheck && pnpm --filter @first-tree/server typecheck`
- `pnpm check`
- `TMPDIR=/var/tmp/first-tree-tests pnpm exec turbo run test --concurrency=2 --env-mode=loose -- --maxWorkers=2`

### DIFF
--- a/packages/client/src/__tests__/agent-slot.test.ts
+++ b/packages/client/src/__tests__/agent-slot.test.ts
@@ -81,6 +81,8 @@ function makeSdk(options?: {
   configError?: unknown;
   /** Throw `error` on the first `times` config fetches, then succeed — models a transient blip. */
   configErrorsThenSucceed?: { error: unknown; times: number };
+  activeRuntimeChatIds?: string[];
+  activeRuntimeChatIdsError?: unknown;
 }): FirstTreeHubSDK {
   const agent = options?.agent ?? makeAgent();
   let configCalls = 0;
@@ -107,6 +109,10 @@ function makeSdk(options?: {
         updatedAt: "2026-01-01T00:00:00.000Z",
         updatedBy: "user-1",
       };
+    }),
+    listActiveRuntimeChatIds: vi.fn(async () => {
+      if (options?.activeRuntimeChatIdsError) throw options.activeRuntimeChatIdsError;
+      return { chatIds: options?.activeRuntimeChatIds ?? ["chat-1", "chat-2", "chat-evicted"] };
     }),
   };
   // AgentSlot only uses this SDK subset; the concrete SDK type has many HTTP methods irrelevant here.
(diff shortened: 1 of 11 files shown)


yuezengwu (changes requested):
Summary: this PR adds an agent-scoped active runtime chat id endpoint, uses that active set to narrow full-state sync/reconcile, force-keeps unsettled delivery work, optimistic-adds pushed chats, and chunks `session:reconcile` frames at the 500-id wire limit. The direction matches the runtime/session model and keeps DB shape unchanged.

Requesting changes for one lifecycle bug:

- `packages/client/src/runtime/agent-slot.ts:604`: `scheduleActiveRuntimeChatIdsRefresh()` always schedules the next timer from the refresh `finally`. If the periodic timer has already fired and `stop()`/`cleanupFailedStart()` runs while `listActiveRuntimeChatIds()` is still in flight, stop clears the old timer and nulls `sdk`, but the in-flight promise then resolves and calls `scheduleActiveRuntimeChatIdsRefresh()` again. From there the stopped slot keeps a live timeout forever; when it fires, `refreshActiveRuntimeChatIds()` returns immediately because `sdk` is null, and the `finally` schedules yet another timer. This leaks stopped slots/timers and can keep the runtime process alive after shutdown/unbind. Please gate the reschedule on the slot still being live, or use a generation/cancel flag so in-flight refreshes cannot resurrect timers after stop.

I was not able to rerun the targeted Vitest suites locally because this review worktree does not have dependencies installed (`pnpm exec` could not find `vitest`).

baixiaohang:
Addressed the requested lifecycle fix in `de28a86b`.

The active runtime chat id refresh path now uses a generation guard. `start()` advances the generation when a new SDK is bound, and `stop()` / failed-start cleanup invalidate the generation immediately before any async unbind/shutdown work. In-flight refreshes and periodic reschedules now no-op if their captured generation is no longer live, so a stopped slot cannot resurrect its hourly timer.

Added a regression test that keeps `unbindAgent()` pending, resolves the in-flight periodic refresh during that stop window, then advances fake timers to verify no new active-runtime-id refresh is scheduled.

Validation:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/agent-slot.test.ts src/__tests__/session-manager-edge-coverage.test.ts --maxWorkers=2`
- `pnpm check` (passes with existing Biome warning/info only)
- `TMPDIR=/var/tmp/first-tree-tests pnpm --filter @first-tree/client exec vitest run --passWithNoTests --maxWorkers=2`

yuezengwu (approved):
Re-reviewed `de28a86b`. The generation guard closes the lifecycle issue I requested changes for: stop/failed-start cleanup invalidates the captured generation before async shutdown work, and both the in-flight refresh result path and periodic reschedule path now no-op when the slot is no longer live.

The added regression test covers the stop window by holding `unbindAgent()` pending, resolving the in-flight periodic refresh, and advancing fake timers to verify no further active-runtime-id refresh is scheduled.

No remaining blockers from my review.

PR #1361, opened 2026-06-30
feat(web): add dev/staging-only scan landing pages

Gandy2025:
## What

Adds the **top-of-funnel landing** for the repo-scan growth entry: a public
`/scan/:campaign` page (`production-scan`, `agent-readiness`) where a visitor
pastes a GitHub repo and is handed off to the existing quickstart flow at
`/quickstart?campaign=<slug>&repo=<encoded>`.

This is the page the user sees *before* quickstart — it feeds the campaign +
repo into the flow that #1352 already shipped.

## Scope / safety

- **Dev/staging-only.** Same prod-safe channel gate as quickstart
  (`useServerChannelState`: unknown / old server → `null` → not allowed). A
  prod (or still-loading) visitor never sees or acts on the form and is
  redirected home — the whole funnel stays hidden in prod until we deliberately
  open it.
- **Public route** (no auth). Login happens later inside quickstart, carried by
  the `next` round-trip.

## Design

- Reuses the existing landing chrome (`LandingNav`/`Footer`, `landing-marketing`
  dark palette, the `text-*` typography + `fg-*` token scale) so it matches `/`.
- One primary `cta` (brand-green) action per the button's own usage guidance.
- Marketing copy lives in a small `scan-copy.ts` registry, kept separate from
  the quickstart `campaigns.ts` behaviour registry; claims stay scoped to the
  checks we actually run (honest, no superlatives — enforced by a test).

## Contract

The handoff-URL builder is extracted (`buildScanHandoffHref`) and unit-tested
**against quickstart's own `readCampaignHandoff` parser** (round-trip), so the
landing's output is provably what quickstart accepts — the `repo` value is
percent-encoded per the contract in `campaigns.ts` (a raw `https://…` repo is
otherwise dropped to `/` by `safeRedirectPath` on the post-login round-trip).

## Testing

- `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) ✓
- `biome check` on changed files ✓
- `pnpm --filter @first-tree/web test` — 135 files / 1094 tests ✓ (8 new)
- Two independent reviews (correctness/funnel + design/UX/a11y); blocking
  findings fixed (input/button height match, stable `aria-live` error region +
  `aria-describedby`, copy honesty).
- Visual QA on a local dev stack (channel=dev) for both campaigns.

## Related

- Builds on the reusable quickstart entry (#1352).
- The per-campaign **scan skill activation** (the agent actually running the
  scan + producing the report) is a separate in-flight workstream; this PR is
[...]

PR #1367, opened 2026-06-30
feat(quickstart): activate per-campaign scan skill (auto-run scored report)

Gandy2025:
## What

Activates the **scan brain** in the quickstart growth funnel so a campaign first-chat **auto-produces a scored report** with zero user prompting. Two parts:

1. **Campaign-aware onboarding directive + kickoff metadata** (`d04041b0`)
   - Server stamps the campaign slug into the kickoff message metadata (`onboarding-kickoff.ts`), alongside the trusted `systemSender`.
   - Client `onboardingSkillDirective` (`agent-io.ts`) reads that trusted metadata and appends an **agent-only** directive to the agent prompt: load and follow the `<campaign>` skill and run it on the repo, then share the report. Never shown in the user bubble. Malformed/absent slug falls back to the existing welcome directive.

2. **Mount + bundle the per-campaign scan skill** (`50b83965`)
   - `mountScanSkill` (in `quickstart-page.tsx`, before the kickoff) creates the campaign's scan skill as a team resource **idempotently by name** (reuses an existing one), then binds it to the agent **preserving existing bindings**.
   - The two scan-skill bodies (`production-scan`, `agent-readiness`) are bundled via Vite `?raw` and mapped per campaign in `scan-skills.ts`. The runtime materializer re-adds YAML frontmatter, so the `.md` files hold the body only.

This completes the funnel **landing -> quickstart -> scored report** — #1352 shipped a light v0 and deferred the report as a fast-follow; this is that fast-follow.

## Verified end-to-end (local dev stack)

Walked the real flow against a connected client: quickstart reused the personal agent -> mounted `production-scan` -> fired the campaign kickoff. The agent then **autonomously, no human prompt**: recognized it should run `production-scan` and loaded the skill; followed the skill's Step 0 (`git clone`d the connected repo read-only into a temp dir); gathered evidence across all 7 dimensions, **scored 80/100**, emitted the strict `ps-1` JSON + a plain-language report, and offered to turn a blocker into a fix PR.

Cross-checked against a manual dogfood of the same repo (scored 84) — consistent band, **same #1 blocker** (no dependency/SAST/secret scanning in CI) — confirming both the auto-trigger and the scan brain hold.

## Scope / safety

- **dev/staging-only**: quickstart is release-channel gated (shipped in #1352); this PR does not change gating. Prod redirects with zero side effects.
- **Agent-only directive**: gated on trusted `metadata.systemSender = first_tree_onboarding`; server strips untrusted senders. The activation instruction never reaches the user-visible message.

## Tests

Green locally: client `1178`, web `1096` (incl. `scan-skills.test.ts` + `quickstart-page.e2e`), server onboarding `42`, shared schema, `agent-io` directive `31`. `typecheck` 10/10, biome + web `lint:tokens` clean.

## Review focus (per pairing)

- directive is agent-only (never leaks to the user bubble) + slug guard
- skill body `?raw` loading (frontmatter stripped, runtime re-adds it)
- `createTeamResource` idempotent-by-name + preserves existing bindings
- mount-before-kickoff timing

## Links

- Landing pages: #1361 (merged). Quickstart v0: #1352 (merged).
- Context Tree write-back tracked separately (scan funnel design + decisions), cross-linked once opened.


yuezengwu (approved):
Reviewed the quickstart scan activation path. The implementation keeps the user-visible kickoff body clean and routes campaign activation through trusted kickoff metadata into the agent prompt only; the web side mounts the raw scan skill before kickoff and preserves existing resource bindings. I checked the server `systemSender` trust boundary, the web bubble gate, resource skill materialization, and kickoff timing.

No blocking findings.

Verified locally:
- `pnpm --filter @first-tree/web exec vitest run src/pages/quickstart/__tests__/quickstart-page.e2e.test.tsx src/pages/quickstart/__tests__/scan-skills.test.ts` (22 tests)
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/agent-io.test.ts` (31 tests)
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/me-onboarding-kickoff.test.ts src/__tests__/message-system-sender-strip.test.ts` (15 tests)

CI is green. One non-blocking boundary to keep in mind: this mount path creates a team resource, so it assumes the quickstart flow is running in an org where the caller can create team resources, or that the scan skill already exists for member-role users.

baixiaohang (dismissed):
## Review by codex-assistant - PR #1367

- Rationale: The scan activation path can fail before kickoff for non-admin members because it now creates a team resource from the quickstart client.

**Risk level**: A
- Semantic lift: also touches onboarding kickoff metadata, agent prompt injection, and agent resource binding.

**PR summary**
- Problem: Campaign quickstart chats should automatically run a campaign-specific repo scan and return a scored report without extra user prompting.
- Approach: Server stamps campaign metadata on trusted onboarding kickoffs; client appends an agent-only skill directive; quickstart creates/binds the campaign scan skill before sending the kickoff; scan skill bodies are bundled via Vite raw imports.
- Impacted modules: `packages/client/src/runtime/agent-io.ts`, `packages/server/src/services/onboarding-kickoff.ts`, `packages/web/src/pages/quickstart/**`, resource APIs.

**Review findings**
❌ 1. Non-admin quickstart users cannot reach the scan chat. `mountScanSkill` now runs before `postOnboardingStartChat` and creates the scan skill via `createTeamResourceForOrg` when the resource is absent (`packages/web/src/pages/quickstart/quickstart-page.tsx:50`, `packages/web/src/pages/quickstart/quickstart-page.tsx:128`). That endpoint is explicitly admin-only (`packages/server/src/api/orgs/resources.ts:14`), while the quickstart page does not gate on `role` or otherwise guarantee an admin membership. A member who can create/reuse their personal agent but is not an org admin will get a 403 during skill mounting, `startChatError` will be set, and the campaign kickoff will never be sent. Please move the campaign-skill ensure/bind behind a server-owned path that works for the quickstart actor, pre-provision the managed scan skill, or explicitly gate quickstart to admins before any setup side effects. [R4 / `packages/web/src/pages/quickstart/quickstart-page.tsx:50`]

⚠️ 2. The team skill ensure is not atomic/idempotent by name. The client does a read-then-create (`availableTeamResources.find(...)` then `createTeamResourceForOrg`) in `mountScanSkill` (`packages/web/src/pages/quickstart/quickstart-page.tsx:41`), but server-side `createTeamResource` only inserts and turns unique violations into 409s (`packages/server/src/services/resources.ts:785`), and the resource table only has active uniqueness for repo canonical keys, not skill names (`packages/server/src/db/schema/resources.ts:37`). Two tabs/users starting the same campaign in the same org can create duplicate active `production-scan` resources. If the quickstart path keeps creating catalog resources, the ensure should be server-side and backed by a real uniqueness/upsert rule for these managed skills. [R1 / `packages/web/src/pages/quickstart/quickstart-page.tsx:41`]

✅ 3. The trusted metadata boundary for `systemSender` is in the right place: ordinary sends strip the key, and the client directive is appended only to runtime input, not the visible bubble.

**Action taken**
- Submitted request changes.

Gandy2025 (commented):
Independent review (2 rounds: server/client trust-boundary + web resource-mount) + own read. **No blocking issues; CI green.** Trust boundary (server-validated campaign slug + un-forgeable `systemSender` + conjunctive client gate), agent-only guarantee (web reads only `systemSender`, never `campaign`), bind-before-kickoff timing (live `configVersion` at dispatch → client `refreshIfNewer` → materialize), idempotency, bindings-preservation, and error handling all verified sound.

Folding in before merge (all non-blocking nits):
1. `scan-skills.ts` `description` still says "current working directory" — align to "the target repository for this chat" (the body's Step 0 clones the repo URL; it's not in cwd).
2. Directive in `agent-io.ts` says "run it on the repository in your workspace" — loosen to "run the scan, then share the report" so the skill's Step 0 owns repo provisioning.
3. `mountScanSkill` "idempotent by name" is best-effort (no DB unique constraint on `(org, type, name)`) — a concurrent two-tab/device mount of the same campaign could double-create the skill resource. Narrow + non-destructive for the dev/staging-only v0; softening the comment now, partial unique index tracked as a pre-prod follow-up.
4. Add tests for the already-bound-skip and `mountScanSkill` failure→retry branches.

Left as-is (with reason): server validates campaign *format* not *membership* — a direct call with an unknown-but-slug-valid campaign yields a phantom skill the agent simply can't find (graceful degradation, self-inflicted only, slug set lives in `packages/web`).

Gandy2025 (commented):
Thanks — R4 is a valid blocker and we're fixing it (my own review missed the admin gate on `createTeamResourceForOrg`; good catch).

**R4 (blocker) — agreed, fixing server-side.** The role gate lives only on the HTTP route `POST /orgs/:id/resources`; `resourcesService.createTeamResource` itself isn't role-gated. So we're moving the ensure+bind of the managed scan skill into the kickoff handler server-side — it provisions + binds for every quickstart actor regardless of org role, removing the non-admin 403. The client `mountScanSkill` is removed and the skill bodies move to a shared managed-skill location. We're deliberately NOT taking the "gate quickstart to admins" alternative, since that would shut non-admin members out of the growth funnel.

**R1 (race) — agreed, fixed by the same move.** The server-side ensure will be idempotent by name (catch the 409 unique-violation / upsert), closing the client read-then-create race. Considering a partial unique index on `(org, type, name)` for managed skills as belt-and-suspenders.

**R3 — thanks for confirming the `systemSender` trust boundary.**

Will re-request review once the server-side fix is pushed.

yuezengwu (changes requested):
Thanks for moving R4/R1 server-side. The non-admin 403 is addressed, but the new placement introduces a stronger authorization regression that needs fixing before merge.

Blocking finding:
- `POST /me/onboarding/kickoff` now calls `ensureAndBindCampaignScanSkill(body.agentUuid, ...)` before `kickoffOnboarding` / `createChat` validates the target agent. The service then trusts `agentUuid`, loads that agent's `organizationId`, creates/reuses a team skill in that org, inserts an agent binding, bumps its config version, and notifies it. If a caller sends a valid campaign plus an agent UUID from another org, an inactive target, or a private agent they are not allowed to chat with, the resource mutation happens first; only afterwards does `createChat` reject via the normal cross-org / active / private-target checks. This lets an unauthorized kickoff attempt mutate another org/agent's resources even though the chat is never allowed. Keep the server-side provision, but move it after the existing participant validation and before bootstrap send, or perform the equivalent org/status/private ownership validation before calling the ensure.

Secondary note while touching this path:
- The resource unique index closes duplicate skill resources, but the binding step is still read-then-insert with no binding uniqueness/lock. Concurrent duplicate kickoff calls can both observe no binding and insert duplicate `agent_resource_bindings` rows for the same skill. Runtime projection does not dedupe enabled skills, so this is not fully idempotent under the same two-tab retry that motivated R1. A row lock around the agent config/agent plus re-read, or a binding-level uniqueness/on-conflict path, would close it.

Verified locally on the updated head:
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/resources-phase1.test.ts src/__tests__/me-onboarding-kickoff.test.ts src/__tests__/message-system-sender-strip.test.ts` (33 tests)
- `pnpm --filter @first-tree/web exec vitest run src/pages/quickstart/__tests__/quickstart-page.e2e.test.tsx src/pages/quickstart/__tests__/campaigns.test.ts src/pages/quickstart/__tests__/intent.test.ts` (36 tests)
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/agent-io.test.ts src/__tests__/resource-skills.test.ts` (34 tests)

Also flagging for human DB review: this revision adds a migration enforcing unique active/stale team skill names per org. That is likely the right model for managed skills, but it can fail if existing deployed data already contains duplicate active/stale team skill names.

Gandy2025 (commented):
Re-reviewed the server-side R4 fix (2 independent passes + own trace). The admin-gate (R4) and the race (R1) are correctly addressed by moving ensure+bind server-side + the partial unique index. **However the move introduces a new BLOCKING issue:**

**BLOCKING — cross-org IDOR.** `ensureAndBindCampaignScanSkill(body.agentUuid, …)` (called at `me.ts:252`, before `kickoffOnboarding`) derives the org **from the agent row** and provisions a skill + binds it + bumps `agentConfigs.version` + notifies, with **no check that the actor is a member of that agent's org**. `resolveOnboardingMember` only validates the user against `body.organizationId`; nothing ties `body.agentUuid` to the actor's org before this write. `createChat` *does* reject cross-org participants, but it runs **after** this bind, in a separate transaction — so an authenticated user can POST `{organizationId: <own>, agentUuid: <victim's agent in another org>, campaign: production-scan}` and bind a managed skill to + mutate the config of a victim's agent (which then re-fetches and runs it). On `main`, `agentUuid` only flowed into `createChat`, so it was rejected with zero side effects; this is the first server write keyed on it ahead of that guard. **Fix:** validate `agent.organizationId === <actor's resolved membership org>` (throw otherwise) *before* any provision/bind.

**MEDIUM — migration 0071.** The partial unique index `uq_resources_team_skill_name_active (org, name)` can fail to apply if any org already has duplicate active/stale team skills with the same name (team skills had no name-uniqueness before), blocking `db:migrate`; it's also non-`CONCURRENTLY` (write-locks `resources`). Recommend a pre-deploy dedup audit and considering `CONCURRENTLY`.

Timing (bind-before-kickoff via live `configVersion`), clean removal of the client path, skill-content fidelity, and the non-admin/idempotent/unknown-campaign tests all check out. Two low nits: add a config-version-bump assertion; `findSkillResourceId` trusts name as skill identity (a pre-existing same-named team skill would be bound instead of the managed body — very low probability).

Will re-verify once the IDOR + migration items are addressed.

yuezengwu (changes requested):
Thanks for the IDOR follow-up. The original cross-org mutation is fixed on this head: provisioning is now after chat validation, and `ensureAndBindCampaignScanSkill` has its own manage-level ownership gate. The duplicate binding race is also addressed by the agent row lock + re-read.

Requesting changes for one new compatibility regression in the agent-private-skill model:

**BLOCKING — agent-private scan skill bindings cannot round-trip through normal resource saves.** The PR now creates the campaign scan skill as `scope="agent"`, `type="skill"`, `ownerAgentId=<target>` in `ensureAndBindCampaignScanSkill` (`packages/server/src/services/resources.ts:1182`). That binding is returned from `getAgentResources`, and the web resource/prompt editors submit the full binding array on every save (`packages/web/src/pages/agent-detail/capability-section.tsx:94`, `packages/web/src/pages/agent-detail/prompt-tab.tsx:45`). But `replaceAgentResources` validates every submitted `resourceId`, and `validateBindingReference` still rejects any agent-scoped non-repo resource with `Only repo resources may be agent-scoped in Phase 1` (`packages/server/src/services/resources.ts:315`). Result: after quickstart mounts the scan skill, the next ordinary resource save for that agent (enable an MCP/skill, edit prompt bindings, add a repo, etc.) fails even if it preserves the existing scan-skill binding unchanged. Please update the resource access/write model to admit owned agent-scoped skills, or choose a representation that does not put a non-round-trippable binding into `getAgentResources`. A regression test should cover: provision campaign skill → call `replaceAgentResources` with the existing bindings plus a normal edit → succeeds and preserves the scan skill.

Non-blocking notes:
- Migration risk is lower now that the unique index is agent-scoped, but it is still a DB constraint change on existing `resources`; human data/migration review still makes sense.
- Route-level campaign kickoff tests still do not exercise the scan-skill side effect through `/me/onboarding/kickoff`; the service tests cover the core authorization and idempotency behavior.

Verified locally on `6ba8d45d`:
- server targeted tests: 35 passed
- web quickstart targeted tests: 36 passed
- client `agent-io` + resource skill tests: 34 passed
- server/web/client typecheck passed; web `lint:tokens` passed
- GitHub CI is green

yuezengwu (approved):
Approved on latest `6a243102`.

The prior blockers are resolved:
- non-admin quickstart no longer depends on the admin-only team-resource HTTP route;
- cross-org/unauthorized `agentUuid` cannot mutate another agent before chat validation, and the service has its own manage-level gate;
- concurrent ensure/bind now serializes under the target agent row lock and re-reads the resource + binding inside the lock;
- the agent-private scan skill binding now round-trips through normal `replaceAgentResources` saves, with regression coverage;
- the DB unique-index migration was removed, so there is no resource-table schema/data migration risk in this PR.

Verified locally on `6a243102`:
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/resources-phase1.test.ts src/__tests__/me-onboarding-kickoff.test.ts src/__tests__/message-system-sender-strip.test.ts` (36 tests)
- `pnpm --filter @first-tree/web exec vitest run src/pages/quickstart/__tests__/quickstart-page.e2e.test.tsx src/pages/quickstart/__tests__/campaigns.test.ts src/pages/quickstart/__tests__/intent.test.ts` (36 tests)
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/agent-io.test.ts src/__tests__/resource-skills.test.ts` (34 tests)
- `pnpm --filter @first-tree/web typecheck` (including `lint:tokens`)

At review time GitHub CI still had the Test job and JS/TS CodeQL pending; I did not wait on CI per normal review flow.

Gandy2025 (commented):
3rd-round re-verification of `6a243102` — **all prior findings resolved; no new issues.**

Confirmed:
- **Cross-org IDOR (was blocking) — fixed.** `ensureAndBindCampaignScanSkill` now runs an ownership gate before any write: caller must be the agent's manager or an admin in the **agent's** org, else 404 (no agent-id enumeration). The cross-org vector is neither, so it's rejected. Provision is also moved into `onChatReady`, after `createChat` validates the participant (defense-in-depth). Covered by a `rejects a caller who cannot manage the target agent (cross-org IDOR)` test.
- **Agent-private skill round-trip (was blocking) — fixed.** `validateBindingReference` now admits owned agent-scoped skills (still rejecting agent-scoped resources not owned by the agent, and still enforcing type/org). The `binding round-trips through replaceAgentResources` test covers it.
- **Duplicate-binding race — fixed without a migration.** Resource find-or-create + binding are serialized under a `SELECT … FOR UPDATE` on the agent row, keyed on `(ownerAgentId, name)`. No DB unique index / no prod schema change (the earlier migration is removed).
- Timing (version bump in-tx → live `configVersion` on the kickoff message → client refresh) and clean-removal of the client path remain correct; my earlier description/directive nits are fixed.

Gates green (server 1616 / web 1094 / client 1178, typecheck/biome/token-lint). One non-blocking follow-up in progress: a route-level `/me/onboarding/kickoff` e2e asserting the bind side-effect (service-level coverage is already solid).

@baixiaohang @yuezengwu — the R4 / IDOR / round-trip items you flagged are addressed on this head; please re-review to clear the request-changes when you have a moment.

baixiaohang (approved):
## Review by codex-assistant - PR #1367

**Recommendation**: approve
- Rationale: The previously blocking quickstart authorization, resource idempotency, cross-org mutation, duplicate-binding race, and agent-private skill round-trip issues are resolved on the current head.

**Risk level**: A
- Semantic lift: also touches onboarding kickoff metadata, agent prompt injection, and agent resource binding.

**PR summary**
- Problem: Campaign quickstart chats should automatically run a campaign-specific repo scan and return a scored report without extra user prompting.
- Approach: Server stamps trusted campaign metadata on onboarding kickoffs, provisions the campaign scan skill through a guarded server-owned path, binds an owned agent-private skill under the agent row lock, and the client appends the agent-only skill directive from trusted kickoff metadata.
- Impacted modules: `packages/client/src/runtime/agent-io.ts`, `packages/server/src/services/onboarding-kickoff.ts`, `packages/server/src/services/resources.ts`, `packages/web/src/pages/quickstart/**`.

**Review findings**
✅ 1. Prior blockers are resolved: non-admin quickstart no longer depends on the admin-only team-resource route; unauthorized cross-org agent mutation is blocked before writes; duplicate ensure/bind is serialized under the target agent lock; owned agent-scoped skill bindings round-trip through normal resource saves; and the resource-table migration risk was removed.
✅ 2. The trusted metadata boundary remains appropriate: ordinary messages strip `systemSender`, and the scan directive is added only to agent runtime input rather than the visible chat bubble.

**Action taken**
- Approved.

PR #1371, opened 2026-06-30
feat: add portable node distribution builder

bestony:
## Summary

- Add portable CLI distribution metadata schemas and per-channel portable download defaults.
- Add a portable builder for prod/staging artifacts with official Node.js runtime verification.
- Add a POSIX installer that verifies SHA256, switches `current` atomically, writes shims, and supports PATH bootstrapping.
- Add env-gated portable bootstrap responses for `/me/connect-tokens` while keeping npm as the default.
- Add CI smoke coverage for a local linux-x64 portable artifact build and verification.

## Verification

- `pnpm check`
- `pnpm typecheck`
- `pnpm exec vitest run apps/cli/tests/portable-builder.test.ts --config apps/cli/vitest.config.ts`
- `pnpm exec vitest run src/__tests__/connect-token-bootstrap.test.ts src/__tests__/me-multi-org.test.ts src/__tests__/bootstrap.test.ts --config vitest.config.ts` from `packages/server`
- `pnpm exec vitest run src/__tests__/channel.test.ts src/__tests__/portable-schema.test.ts src/config/__tests__/server-config.test.ts --config vitest.config.ts` from `packages/shared`
- Local linux-x64 portable smoke build and artifact verification with official Node.js v24.15.0

`pnpm test` was also attempted; it failed in an unrelated existing skill-evals timeout:
`packages/skill-evals/src/suites/first-tree-seed/__tests__/periodic.test.ts:28`.


yuezengwu (changes requested):
This PR adds a portable First Tree CLI distribution path: shared metadata schemas, channel-aware portable defaults, a Node-bundled artifact builder, a POSIX installer, env-gated `/me/connect-tokens` portable bootstrap responses, and CI smoke coverage.

The overall direction matches the release/channel constraints: npm remains the default bootstrap, prod/staging identities stay separated, dev stays source-only, and the builder verifies the downloaded official Node tarball against Node's SHA256 list.

Blocking issue found:

- `scripts/portable/install.sh:310-313` does not actually replace `current` on upgrade once `current` already exists as a symlink to a directory. On macOS/BSD `mv` (and similarly on common POSIX/GNU behavior without `-T`), `mv -f "$NEW_LINK" "$CURRENT_LINK"` treats the destination symlink-to-directory as a directory target and moves the new `.current.$$` symlink inside the old version directory instead of replacing `PREFIX/current`. The result is that the installer downloads/verifies/extracts the new version, but `current` and the shims still point at the old version. This breaks the advertised atomic switch and means portable upgrades can silently leave users on the old binary.

Please update the switch logic so it targets the `current` symlink itself (for example via a portable no-dereference approach or platform-specific `mv -h`/`mv -T` fallback) and add a successful upgrade test that installs version A, then version B, and asserts `prefix/current/VERSION` changes to B.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1371

- Rationale: the new portable bootstrap path can create installs that do not update through the portable artifact plane, and the generated bootstrap command has an unsafe fixed `/tmp` handoff.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and CLI distribution scripts/workflows -> A
- Semantic lift: touches bootstrap/install/update release behavior; no further lift beyond A

**PR Summary**
- Problem: make the CLI installable on machines that do not already have Node.js.
- Approach: add shared portable metadata schemas, channel portable defaults, a Node-bundled artifact builder, a POSIX installer, env-gated portable connect-token bootstrap responses, and a Linux x64 CI smoke build.
- Impacted modules: `scripts/portable/**`, `apps/cli/**`, `packages/shared/**`, `packages/server/src/api/me.ts`, `.github/workflows/ci.yml`.

**Review findings**
❌ 1. Portable installs are still routed through the npm/global update executor. [`scripts/portable/install.sh:162`, `apps/cli/src/core/update.ts:15`, `apps/cli/src/core/update.ts:50`, `apps/cli/src/core/update-glue.ts:82`]

The installer shims mark the process with `FIRST_TREE_INSTALL_MODE=portable`, but the update code never reads that mode and `detectInstallMode()` only returns `global | npx | source`. Because the portable artifact also contains `app/package.json` with the channel package name, a portable launch can be classified as `global`, and both daemon auto-update and manual `upgrade` then run `npm install -g <pkg>@<target>`. On the machines portable is meant to support, npm may not exist, so updates fail; if npm does exist, it mutates a separate global install instead of switching the portable `current` symlink. Please either add the portable executor now (server target -> channel manifest/tarball -> checksum -> `current` switch) or keep `connectBootstrap.method=portable` from being selectable until that executor exists.

❌ 2. The generated portable bootstrap command executes a predictable `/tmp` script even when the download fails. [`packages/server/src/api/me.ts:59`]

`buildPortableBootstrapCommand()` writes to `/tmp/first-tree-install-${binName}.sh`, then runs `sh` on that path as the next newline-separated command. In a shared `/tmp`, another local user can pre-create that predictable file as readable but not writable; `curl -o` fails, and the pasted bootstrap still executes the attacker-controlled script. The same fixed path can also clobber a symlink target owned by the installing user. Please generate the command with a fresh `mktemp` path, chain failures so `sh` is not run after a failed download, and clean it up with a trap/finalizer.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
This synchronized revision adds a portable self-update executor and improves the generated bootstrap command. That addresses the update-routing concern at the TypeScript auto-update/manual-upgrade layer, but the original POSIX installer blocker is still present.

Blocking issue still unresolved:

- `scripts/portable/install.sh:310-313` still does `mv -f "$NEW_LINK" "$CURRENT_LINK"` after creating the replacement `current` symlink. When `current` already exists as a symlink to a version directory, `mv` treats that symlink-to-directory as a directory target and moves the new `.current.$$` symlink inside the old version directory instead of replacing `PREFIX/current`. I re-ran the shell repro against this revision and `current` remains pointed at the old target.

This matters because `/me/connect-tokens` portable bootstrap runs `install.sh` for fresh-machine installs, and users may re-run the installer to upgrade or repair a portable install. The new `installPortableSpec()` path uses Node `rename()` and has tests, but it does not fix the shipped POSIX installer path. Please update `install.sh` to replace the `current` symlink itself and add a successful installer upgrade test that seeds version A, runs `install.sh` for version B, and asserts `prefix/current/VERSION` is B.

baixiaohang (commented):
## Review by codex-assistant - PR #1371

- Rationale: the synchronized revision fixes the portable update executor and the generated bootstrap handoff, but the shipped POSIX installer still does not replace `current` correctly on reinstall/upgrade.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and portable distribution scripts/workflows -> A
- Semantic lift: touches CLI install/update/bootstrap release behavior; no further lift beyond A

**PR Summary**
- Problem: make the CLI installable without a preinstalled Node.js runtime.
- Approach: add portable metadata schemas, channel download defaults, a Node-bundled artifact builder, POSIX installer, env-gated connect-token portable bootstrap, and portable self-update support.
- Impacted modules: `scripts/portable/**`, `apps/cli/**`, `packages/shared/**`, `packages/server/src/api/me.ts`, `.github/workflows/ci.yml`.

**Review findings**
❌ 1. The POSIX installer still fails to replace `current` when rerun against an existing symlink. [`scripts/portable/install.sh:310`]

`install.sh` creates `PREFIX/.current.$$` and then runs `mv -f "$NEW_LINK" "$CURRENT_LINK"`. When `CURRENT_LINK` is already a symlink to a directory (the normal installed state), common `mv` behavior treats the destination as a directory target and moves the replacement link inside the old version directory instead of replacing `PREFIX/current`. The new TypeScript `installPortableSpec()` path uses Node `rename()` and is covered by tests, but the bootstrap command still ships and runs this shell installer for fresh installs and for users re-running the installer to upgrade or repair. Please update the shell switch to replace the symlink itself and add an installer-level upgrade test that seeds version A, runs `install.sh` for version B, and asserts `prefix/current/VERSION` is B.

✅ 2. The previous codex-assistant concerns about portable update routing and the predictable `/tmp` bootstrap script are addressed in this revision. [`apps/cli/src/core/update.ts:77`, `apps/cli/src/core/update-glue.ts:155`, `packages/server/src/api/me.ts:65`]

**Action taken**
- Submitted comment; an existing latest-commit review already requested changes for the same installer blocker.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1371

- Rationale: the latest fix avoids the `mv` dereference bug, but it does so by deleting `current` before recreating it, so the installer no longer performs an atomic version switch.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and portable distribution scripts/workflows -> A
- Semantic lift: touches CLI install/update/bootstrap release behavior; no further lift beyond A

**PR Summary**
- Problem: make the CLI installable without a preinstalled Node.js runtime.
- Approach: add portable metadata schemas, channel download defaults, a Node-bundled artifact builder, POSIX installer, env-gated connect-token portable bootstrap, and portable self-update support.
- Impacted modules: `scripts/portable/**`, `apps/cli/**`, `packages/shared/**`, `packages/server/src/api/me.ts`, `.github/workflows/ci.yml`.

**Review findings**
❌ 1. The shell installer still does not switch `current` atomically. [`scripts/portable/install.sh:313`]

The latest change replaces the old `mv -f "$NEW_LINK" "$CURRENT_LINK"` with `rm -f "$CURRENT_LINK"` followed by `mv "$NEW_LINK" "$CURRENT_LINK"`. That avoids moving the new symlink into the old version directory, but it creates a non-atomic handoff: between those two commands, `PREFIX/current` does not exist. If the installer is interrupted in that window, or a supervised daemon restart happens at the wrong time, the channel shims point at a missing root and the install can be left broken. The portable distribution design and PR summary both rely on an atomic `current` switch after checksum/extraction succeeds. Please replace the symlink itself using a no-dereference atomic rename for the supported platforms (for example GNU `mv -T` on Linux and BSD/macOS `mv -h`, failing closed if neither is available), and keep the upgrade regression test.

✅ 2. The earlier concerns about portable update routing, predictable `/tmp` bootstrap handoff, and the symlink-to-directory dereference behavior are otherwise addressed.

**Action taken**
- Submitted request changes.

yuezengwu (approved):
This revision fixes the remaining portable installer blocker.

Review summary:
- The PR adds the portable CLI distribution path: metadata schemas/defaults, Node-bundled artifact builder, POSIX installer, env-gated connect-token portable bootstrap, CI smoke coverage, and portable self-update support.
- The previously blocking shell installer upgrade path is now covered by an installer-level A -> B test and `install.sh` no longer moves the replacement symlink into the old version directory.
- The earlier portable update-routing and fixed `/tmp` bootstrap concerns were already addressed in the prior synchronized revision.

Focused verification run locally:
- `pnpm exec vitest run apps/cli/tests/portable-builder.test.ts --config apps/cli/vitest.config.ts`
- `pnpm exec vitest run apps/cli/src/__tests__/update-portable-install.test.ts apps/cli/src/__tests__/update-glue.test.ts apps/cli/src/__tests__/update-detect-install-mode.test.ts --config apps/cli/vitest.config.ts`

No remaining blocking review findings from me.

yuezengwu (changes requested):
Superseding my previous approval: I agree the remaining installer switch is still a blocker.

The latest revision fixes the symlink-to-directory dereference by changing the shell installer to `rm -f "$CURRENT_LINK"` followed by `mv "$NEW_LINK" "$CURRENT_LINK"`. That makes the A -> B reinstall test pass, but it no longer satisfies the portable installer contract that `current` is switched atomically after the payload has been verified and extracted. During the gap between `rm` and `mv`, shims point at a missing `PREFIX/current`; an interrupted installer or concurrent daemon restart can leave the portable install broken.

Please keep the new installer-level upgrade regression test, but replace `current` with a no-dereference atomic rename for the supported shell platforms (for example, choose GNU `mv -T` on Linux and BSD/macOS `mv -h` where available, failing closed if neither is supported).

yuezengwu (approved):
This revision fixes the remaining portable installer blocker from my last review.

Review summary:
- The PR adds the portable CLI distribution path: channel-aware metadata/defaults, Node-bundled artifacts, POSIX installer, env-gated connect-token portable bootstrap, CI smoke coverage, and portable self-update support.
- The installer now switches `current` through a platform-specific no-dereference rename (`mv -T` on Linux, `mv -h` on Darwin) instead of deleting `current` before recreating it.
- The shell installer tests now cover the A -> B upgrade path and a simulated current-switch failure that must leave the old `current` symlink intact.

Focused verification run locally on the latest head `29efd83f`:
- `pnpm exec vitest run apps/cli/tests/portable-builder.test.ts --config apps/cli/vitest.config.ts`
- `pnpm exec vitest run apps/cli/src/__tests__/update-portable-install.test.ts apps/cli/src/__tests__/update-glue.test.ts apps/cli/src/__tests__/update-detect-install-mode.test.ts --config apps/cli/vitest.config.ts`

No remaining blocking review findings from me.

yuezengwu (approved):
Re-approving the synchronized head `2c33c573` after the main-merge update.

Review summary:
- The PR still implements the portable CLI distribution path: channel-aware metadata/defaults, Node-bundled artifacts, POSIX installer, env-gated connect-token portable bootstrap, CI smoke coverage, and portable self-update support.
- The latest commit merges `main`; I rechecked the PR diff against the current base and the conflict-sensitive update glue where portable self-update now combines with the newer update logging path.
- The prior portable installer blocker remains fixed: shell installs replace `current` through the platform-specific no-dereference rename path, with tests covering A -> B upgrade and failed current-switch preservation.

Focused verification run locally on `2c33c573`:
- `pnpm exec vitest run apps/cli/tests/portable-builder.test.ts --config apps/cli/vitest.config.ts`
- `pnpm exec vitest run apps/cli/src/__tests__/update-portable-install.test.ts apps/cli/src/__tests__/update-glue.test.ts apps/cli/src/__tests__/update-detect-install-mode.test.ts --config apps/cli/vitest.config.ts`

No remaining blocking review findings from me.

baixiaohang (approved):
Approved per reviewer authorization. The PR is acceptable to merge from this review perspective.

PR #1375, opened 2026-07-01
fix(server): restore liveness on client heartbeat

### LINKED ISSUE #1374: Server: clients.status / agent_presence.status desync — heartbeat can't recover a false-positive stale sweep
## Summary

The server caches "does this client have a live WebSocket" in the `clients.status` column (and per-agent liveness in `agent_presence.status`). Under a common condition — the operator's Mac sleeping / entering low-power so the daemon heartbeat timer freezes for >60s while the TCP socket stays half-open — the background stale-sweep marks the client `disconnected` and its agents `offline`, but **no live-traffic path ever writes those statuses back**. The row stays stuck until the next ~29-minute auth-refresh reconnect, so the outage is not a momentary flicker — it can persist for ten-to-twenty-plus minutes.

The visible result: the topbar shows a red **"Computer disconnected"** chip and the agent shows **offline**, while that same agent is demonstrably still `working` and running commands.

All line references below are against `main` at the time of writing.

## Symptom

- Topbar red chip "Computer disconnected" + agent "offline".
- Simultaneously, the same agent is actively `working` and executing commands.
- The two states coexist for many minutes (not a sub-second race).
- The computer is **not** actually disconnected (WebSocket stays up the whole time; `daemon doctor` reports reachable), and it is **not** a stale/duplicate `clients` row (all agents are pinned to the one live client row).

## Root cause

A real server-side desync between `clients.status` and the true connection, which the heartbeat cannot repair.

### 1. The stale sweep (aggressive, and correct to be so)

`cleanupStaleClients` (`packages/server/src/services/client.ts:432`) runs every ~30s and flips `clients.status` from `connected` → `disconnected` when either:
- `last_seen_at` is older than ~60s, **or**
- `instance_id` points at a `server_instances` row whose own heartbeat is stale.

In the same sweep it also sets every bound agent `offline` **and clears `agent_presence.client_id` to NULL** (`client.ts:446-452`). `cleanupStalePresence` / `markStaleAgents` in `presence.ts` do the equivalent on the agent side.

This heuristic is doing its job — a half-open TCP socket is indistinguishable from a dead one without a liveness signal, so staleness-based reaping is correct.

### 2. The missing recovery path

The bug is that **recovery is missing**, not that the sweep is wrong:

- `heartbeatClient` (`client.ts:142`) only writes `last_seen_at`. It does **not** write `status` back to `connected` and does **not** refresh `instance_id`.
- `touchAgent` (`packages/server/src/services/presence.ts:158`) only writes `agent_presence.last_seen_at`. It does **not** restore `status` or `client_id`.
- The **only** code that writes `status='connected'` is `registerClient` (`client.ts:52`, sets it at both INSERT and `onConflictDoUpdate`), and the only code that restores `agent_presence.status='online'` + `client_id` is `bindAgent` (`presence.ts:70`). Both fire **only on a fresh WS register/bind**.

Because a Mac-sleep half-open socket never actually closes, the client never re-registers, so `status` stays `disconnected` / `offline` until the next auth-refresh forces a reconnect (~29 min).

The heartbeat handler at `packages/server/src/api/agent/ws-client.ts:1515-1529` proves the point — on a live heartbeat it calls only `heartbeatClient` + `touchAgent`, both of which touch `last_seen_at` and nothing else.

### 3. Why "working + offline" necessarily coexist

`setRuntimeState` (`presence.ts:141`, invoked by the `runtime:state` frame) writes `runtime_state` and `last_seen_at` but **not** `status`. So after the sweep sets `status='offline'`, an agent that keeps working just re-stamps `runtime_state='working'` on top of a stuck `status='offline'`. The two columns are updated by independent paths and drift apart — this is why the screenshot's contradiction is not a glitch but the expected outcome.

Note also that `isAgentStillRoutedHere` (`ws-client.ts:325`) reads the **in-memory** routing (`boundAgents` + `connectionManager`), not the DB `agent_presence.client_id`. So after the sweep NULLs the DB `client_id`, the agent is still routed and keeps working in memory while its DB row reads offline/unbound.

## Trigger conditions

1. Mac sleep / low-power freezes the daemon heartbeat timer for >60s (same fingerprint as the prod incident with `silenceMs 231555`).
2. `dev.cloud` deploy / server-instance rotation (stale `instance_id`).

Staging hits both frequently, which is why it reproduces often there.

## Proposed fix

Server-side, no schema change, no wire-contract change.

### Primary — heartbeat restores `clients.status`

On heartbeat, `heartbeatClient` should write `status='connected'` and refresh `instance_id` (not just `last_seen_at`). A heartbeat is positive proof of "alive and on this instance", so it should be as authoritative for `status` as `registerClient` is. `instance_id` is already in scope at the heartbeat call site (`clientWsRoutes(notifier, instanceId)` closure), so the value just needs to be threaded into `heartbeatClient`. Effect: a false-positive sweep self-heals within ~30s instead of stalling ~29 min.

### Secondary — heartbeat restores agent presence

The same sweep marks agents `offline` + clears `client_id`, and `touchAgent` does not restore them. On heartbeat, restore `agent_presence.status='online'` + `client_id` for the agents still routed on this socket, gated by the existing `isAgentStillRoutedHere` filter. **`runtime_state` must be left untouched** (see discussion below) so a live `working` state is preserved.

### Open question (for follow-up)

The sweep also resets `runtime_state` to NULL. After recovery, an idle agent would read "online but `runtime_state=NULL`". Should the recovery `COALESCE(runtime_state, 'idle')` to mirror `bindAgent`'s initial state, or leave NULL? Leaning toward COALESCE-to-idle, but flagging for the owner.

### Rejected alternative

Have the client detect the heartbeat gap on wake and proactively reconnect/re-register instead of patching status server-side. Rejected because the root cause is precisely that **the socket never truly closed**, so there is no reconnect trigger; adding wake-detection on the client is more invasive and fragile across platform sleep behaviors than making the server treat "heartbeat = authoritative liveness."

## Why the online-restore is safe (from the design discussion)

The concern was: is it ever wrong to blindly set an agent `online` on heartbeat?

- The heartbeat loop only touches agents passing `isAgentStillRoutedHere` — i.e. bound on **this live socket** and with the connection-manager's authoritative in-memory routing still pointing the agent at this client. That is exactly the system's definition of `online`; the sweep only set it offline on a **staleness heuristic**, with no real unbind. A heartbeat arriving on the still-routed socket is the refutation of that false positive.
- Both sweep triggers make revival correct-only:
  - Mac-sleep half-open: the heartbeat returns on the **same live socket / same live instance** → this is exactly the false positive we want to heal.
  - Instance rotation: the socket to the dead instance is also gone, so the client **must** reconnect to a new instance and re-register/bind (self-heal); the dead instance can't receive a heartbeat, so there is no wrong revival.
- Key distinction: `status`/`online` is a **connectivity fact provable from routing** and may be asserted; `runtime_state` (idle/working/error/blocked) is an **activity fact** that must come from the agent's own `runtime:state` frames and must never be fabricated. Hence the recovery write is surgical: `status` + `client_id` + `last_seen_at` only.

## Landing notes

- Server-behavior fix only; does not touch schema or wire contracts, so it should not be subject to the tree-PR gate.
- TDD: failing test = force a stale sweep (client + agent marked disconnected/offline), then deliver a heartbeat, and assert `clients.status` returns to `connected` (+ `instance_id` refreshed) and `agent_presence.status` returns to `online` (+ `client_id` restored) while `runtime_state` is unchanged.
- Existing coverage to extend: `packages/server/src/__tests__/client-service.test.ts`, `presence.test.ts`, `ws-client-reconnect-race.test.ts`.

### Immediate mitigation (no code)

Wait for the next auth-refresh to self-heal, or force a reconnect with `launchctl kickstart -k gui/$(id -u)/first-tree-staging`. Do **not** click the "Reconnect" dialog — that re-pairs from scratch and mints an extra `client_id`.

baixiaohang:
## Summary

- add a runtime liveness coordinator for client heartbeat reachability writes
- restore `clients.status` / `instance_id` and still-routed `agent_presence` route leases from valid heartbeats
- guard heartbeat restore and heartbeat-triggered inbox repair by active socket, current client instance, durable agent pin, and active agent status
- keep runtime activity/provider health separate: heartbeat does not write `runtime_state`, sessions, or provider state

Fixes #1374.

## Tests

- `pnpm check`
  - passes; existing unrelated warning/info remain in `packages/client/src/__tests__/assistant-text.test.ts`, `packages/client/src/runtime/workspace-migrations.ts`, and `packages/web/src/index.css`
- `pnpm typecheck`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --maxConcurrency=2 --no-file-parallelism src/__tests__/client-service.test.ts src/__tests__/ws-client-reconnect-race.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --maxConcurrency=2 --no-file-parallelism`

### DIFF
--- a/packages/server/src/__tests__/client-service.test.ts
+++ b/packages/server/src/__tests__/client-service.test.ts
@@ -1,9 +1,12 @@
 import { eq } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
 import { agentPresence } from "../db/schema/agent-presence.js";
+import { agents } from "../db/schema/agents.js";
+import { clients } from "../db/schema/clients.js";
 import { createAgent, suspendAgent } from "../services/agent.js";
 import * as clientService from "../services/client.js";
 import * as presenceService from "../services/presence.js";
+import { recordClientHeartbeat } from "../services/runtime-liveness.js";
 import { createTestAgent, useTestApp } from "./helpers.js";
 
 /**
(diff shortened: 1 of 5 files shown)


yuezengwu (approved):
Goal/core changes reviewed: this restores client/agent reachability from authoritative heartbeat evidence on the active WS while keeping runtime activity/provider health separate. The new `recordClientHeartbeat` path updates only existing `clients` and `agent_presence` liveness columns, gates restore by current client instance, active socket routing, durable client pin, and active agent status, and keeps heartbeat-triggered inbox repair behind the same liveness guards.

No blocking findings. No schema/migration changes; this is a DB write semantics change over existing liveness columns.

Verified locally:
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --maxConcurrency=2 --no-file-parallelism src/__tests__/client-service.test.ts src/__tests__/ws-client-reconnect-race.test.ts`
- `pnpm typecheck`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --maxConcurrency=2 --no-file-parallelism`
- `pnpm check` (passes with the existing unrelated warnings/info noted in the PR)

PR #1376, opened 2026-07-01
feat(web): explicit @mention reminder in group-chat composer

Gandy2025:
## What & why

A group chat requires every send to `@mention` at least one member (`enforceGroupMention` server-side; `sendBlockedByMentionGate` client-side). But the requirement was only *implied*: a disabled send button and a hover-only tooltip. A user who typed a message and hit Enter saw **nothing happen, with no visible reason** — the "silent dead-end" the messaging contract explicitly warns against.

This makes the requirement **explicit**, without adding noise:

- **Persistent hint row (A1)** inside the composer card, above the textarea, shown whenever the gate blocks (group chat + no `@mention` yet). It collapses the moment a member is addressed. Placed *inside* the card (not a separate band above it) so it never collides with the `ComposeStatusBar` that occupies the space above the composer.
- **In-the-moment feedback (B)**: a blocked send attempt (Enter / programmatic — a disabled button can't fire `onClick`) tints that **same** row an error tone for ~2.2s via `nudgeMentionGate`, rather than printing a second, duplicate line. This unifies text and image sends and replaces the old image-only `uploadError` hint.
- **Placeholder de-dup**: the `requiresMention` placeholder drops to a neutral `Write a message…` so it no longer repeats what the hint row says.

Copy chosen by the product owner: `@mention someone to send in this group`.

## Design

Effect image (dual-theme, 4 real states) was reviewed and approved before implementation. Everything uses existing design tokens (`--sp-*`, `--border-faint`, `--fg-3`, `--fg-error-strong`, `--hairline`, `text-label`) and reads correctly in light and dark. The `AtSign` icon is reused from the mention toolbar button. The nudge uses `--fg-error-strong` (the AA-text error token), not the brighter `--state-error`.

## Behaviour / correctness

Gated entirely on the existing `sendBlockedByMentionGate = requiresMention && draftMentions.length === 0 && !askOverlayActive`, so it stays in lockstep with the send-button disable and the server enforcement. Never shows in 1:1 chats (`computeRequiresMention` is false for 2 speakers) or while a question dock/overlay lifts the gate. The nudge clears on three paths (timer, keystroke, gate lift) and the timer is cleaned up on unmount.

## Tests

- Web suite green: **1094/1094**; `typecheck` + `biome` + `lint:tokens` clean.
- Updated `page-ssr-smoke` and the composer DOM test (including the image-only blocked-send path, which now asserts the shared hint row).

## Review

Two independent reviews (correctness + design) — both **approve, zero blocking**. Adopted the named-const suggestion (`MENTION_NUDGE_MS`); copy finalized with the product owner.

## Context Tree

The must-`@mention` contract (and its silent-dead-end rationale) is already documented in `system/cloud/chat/messaging.md`. This PR surfaces that existing contract in the UI without changing it — **implementation-only, no tree write**.

### DIFF
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1173,7 +1173,7 @@ describe("page SSR smoke coverage", () => {
     const { ChatView } = await import("../workspace/center/chat-view.js");
 
     expect(renderWithClient(<ChatView agentId="agent-1" chatId="chat-1" />, createClient())).toContain(
-      "Type @ to pick a recipient",
+      "@mention someone to send in this group",
     );
 
     const readOnlyClient = createClient();
--- a/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
@@ -1331,7 +1331,11 @@ describe("ChatView", () => {
 
     const textarea = container.querySelector<HTMLTextAreaElement>("textarea");
     if (!textarea) throw new Error("Composer textarea missing");
-    expect(textarea.placeholder).toContain("Type @ to pick a recipient");
+    // Group chat, no @mention yet: the persistent hint row states the rule, so
+    // the placeholder stays neutral (no duplicate "@ to pick a recipient").
+    expect(textarea.placeholder).toContain("Write a message");
+    expect(textarea.placeholder).not.toContain("Type @ to pick a recipient");
+    expect(container.textContent).toContain("@mention someone to send in this group");
     await setValue(textarea, "Please review @nova");
     await click(container.querySelector('button[aria-label="Send"]'));
     await waitForCondition(() => chatMocks.sendChatMessage.mock.calls.length > 0, "Expected text send");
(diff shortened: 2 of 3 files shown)


yuezengwu (approved):
Approved. I reviewed the composer change against the existing messaging contract.

Core change: the group-chat composer now shows a persistent in-card @mention hint while the existing mention gate blocks, reuses that same row for blocked Enter/image-send feedback via a short error-tone nudge, and keeps the placeholder neutral to avoid duplicate guidance.

No blockers found. The PR is scoped to web composer UI/tests; it does not change the server contract, core data structures, or database schema. The existing Context Tree messaging node already documents the must-address group send contract, so I agree no tree write is needed.

Current GitHub checks are green, including Lint & Type Check, Test, and CodeQL.

baixiaohang (approved):
## Review by codex-assistant - PR #1376

**Recommendation**: approve
- Rationale: The change makes the existing group-chat mention gate visible in the composer while preserving the same send gate and server-side addressing contract.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed -> C.
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or shared data-layer behavior.

**PR Summary**
- Problem: Group-chat sends require a structured `@mention`, but the UI only implied that through a disabled send button and tooltip, so Enter/programmatic attempts could feel like a silent no-op.
- Approach: Add a persistent in-card hint row driven by `sendBlockedByMentionGate`, transiently error-tone that same row on blocked send attempts, and keep the placeholder neutral to avoid duplicate copy.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, the chat-view DOM test, and the page SSR smoke test.

**Review findings**
✅ 1. The hint row, send-button disable state, and Enter/programmatic backstop all share `sendBlockedByMentionGate`, keeping the UI in lockstep with the documented chat addressing contract. [R1 / `packages/web/src/pages/workspace/center/chat-view.tsx`]
✅ 2. The old image-only `uploadError` branch is replaced by the same hint/nudge surface, so text and image sends no longer carry duplicate explanation paths, and the image-only blocked path remains covered by DOM tests. [R2 / `packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx`]
✅ 3. No blocking findings.

**Action taken**
- Approved.

PR #1378, opened 2026-07-01
fix: materialize campaign scan skill on reused/active agents

Gandy2025:
## Problem

On staging, a scan skill bound to a **reused / already-running** agent (e.g. `gandy-coder`) was never written to disk. The kickoff directive told the agent to run the `production-scan` skill, but the `SKILL.md` was not there, so the agent improvised its own checklist instead of running the rubric.

## Root cause

Two gaps, one per side:

1. **Client** — `materializeResourceSkills` was only called in the handler's `start` / normal `resume`. It was **not** called when a skill binding arrives mid-session:
   - The mid-session config-change path (`maybeSwitchConfig`, Path B restart) rewrote the briefing so `## Team Skills` listed the new skill, but never wrote the file it pointed at.
   - The **legacy-cwd resume branch** never materialized skills at all (this is exactly the reused legacy-layout agent case).
2. **Server** — `ensureAndBindCampaignScanSkill` reused an existing skill resource without refreshing its body, and the already-bound path early-returned **without bumping `agentConfigs.version`**. So even after the body was updated, the client's `refreshIfNewer` was a no-op and never re-fetched — a returning user's re-scan silently kept the stale rubric.

## Changes

**Change 1 — client (`packages/client/src/handlers/claude-code.ts`)**
- Materialize resource skills on the `maybeSwitchConfig` restart path, **co-located with the briefing rewrite** so the `## Team Skills` paths and the on-disk files stay consistent. Reuses the existing `appliedConfigVersion` and guards on `cached.version > appliedConfigVersion`, so a swallowed-refresh empty fallback config (version 0) cannot prune skills the running turn is about to load.
- Materialize on the legacy-cwd resume branch, using that branch's `cwd` (= `legacyCwd`), matching where its briefing paths and the SDK cwd resolve.

**Change 2 — server (`packages/server/src/services/resources.ts`)**
- On an already-provisioned skill, refresh the resource payload to the current server-owned body inside the existing agent row-lock transaction.
- On the already-bound path, bump `agentConfigs.version` + notify **only when the body/description actually changed** (skip otherwise to avoid churn). This is the trigger the client's inject-time materialize consumes.

## Hook placement note

`maybeSwitchConfig` is the existing mid-session config-change handler on the inject/drain path (`pushInjectedMessage` → `maybeSwitchConfig`). A skill binding changes the payload (not just the model), so it already takes the Path B restart — which is precisely where the briefing is rewritten. Placing materialize there keeps briefing ⇔ disk consistent and reuses the version tracking, rather than adding a parallel version tracker in the drain loop.

## Handler scope

**claude-code only** in this PR — the only handler the growth funnel runs today. `codex-sdk`, `codex-app-server`, and `claude-code-tui` share the same general inject-materialize gap (they also only materialize on start/resume) and will land in a **fast-follow** general-robustness PR.

## Testing

- **Server** (`resources-phase1.test.ts`): new case — a stale skill body on the already-bound path is refreshed to the current rubric **and** bumps the config version; existing idempotency test still asserts no bump when content is unchanged. Full server suite green (1622 tests).
- **Client** (`claude-code-inject-materialize.test.ts`, new): (a) a skill bound mid-session is materialized to disk on the next injected turn; (b) a lower-version empty fallback config does **not** prune the live skill. Full client suite green (1186 tests).
- `pnpm check` (biome) + `pnpm typecheck` clean.

## Manual verification (post-release)

Reproduce the staging case: reuse `gandy-coder` (active session), run a scan directive → skill lands on disk → the scan runs by rubric (structured report + `AGENTS.md`), no improvisation.

## Context / tree

Design + review handoff for this fix lives in the growth-funnel scan-skill thread (resource model kept, two fixes, hidden-marker dropped). The broader "scan skill = agent-scoped resource" loading model (from #1367) is a candidate for a Context Tree note under the growth-funnel nodes, but this PR itself is implementation-only.

---

⚠️ **Not for self-merge.** Flow: independent review (gandy-s-assistant, 2 rounds) → code-owner approve → manual merge by gandy → client release + `gandy-coder` update → staging re-verification.

### DIFF
--- a/packages/client/src/__tests__/claude-code-inject-materialize.test.ts
+++ b/packages/client/src/__tests__/claude-code-inject-materialize.test.ts
@@ -1,0 +1,259 @@
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import type { AgentRuntimeConfig, AgentRuntimeConfigPayload, RuntimeResourceSkill } from "@first-tree/shared";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { ChatContext } from "../runtime/chat-context.js";
+import type { SessionContext, SessionMessage } from "../runtime/handler.js";
+import { mockCtxPlumbing } from "./test-helpers.js";
+
+// A resource skill bound to an agent that already has a live session only lands
+// on disk if the handler re-materializes it when the config version bumps —
+// see maybeSwitchConfig's restart path in claude-code.ts. This exercises that
+// path end-to-end against a real temp workspace (materializeResourceSkills is
+// NOT mocked): the SKILL.md must appear before the injected turn runs.
+const state = vi.hoisted(() => ({
+  chatContextPromise: null as Promise<ChatContext> | null,
+  resolveChatContext: null as ((value: ChatContext) => void) | null,
+  observedInputs: [] as string[],
+  pendingResults: [] as unknown[],
+  waiters: [] as Array<() => void>,
+}));
+
+function wakeQuery(): void {
+  const waiters = state.waiters.splice(0);
+  for (const waiter of waiters) waiter();
+}
+
+function flattenContent(content: unknown): string {
+  return typeof content === "string" ? content : JSON.stringify(content);
+}
+
+vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
+  query: (args: { prompt: AsyncIterable<{ message: { content: unknown } }> }) => {
+    let closed = false;
+    void (async () => {
+      for await (const sdkMsg of args.prompt) {
+        state.observedInputs.push(flattenContent(sdkMsg.message.content));
+        state.pendingResults.push({
+          type: "result",
+          subtype: "success",
+          result: `reply ${state.observedInputs.length}`,
+        });
+        wakeQuery();
+      }
+      closed = true;
+      wakeQuery();
+    })();
+    return {
+      [Symbol.asyncIterator]() {
+        return {
+          next: async (): Promise<IteratorResult<unknown>> => {
+            while (state.pendingResults.length === 0 && !closed) {
+              await new Promise<void>((resolve) => state.waiters.push(resolve));
+            }
+            const value = state.pendingResults.shift();
+            if (value) return { value, done: false };
+            return { value: undefined, done: true };
+          },
+        };
+      },
+      close: () => {
+        closed = true;
+        wakeQuery();
+      },
+      setModel: async () => {},
+    };
+  },
+}));
+
+vi.mock("../runtime/agent-bootstrap.js", () => ({ ensureAgentBootstrap: vi.fn() }));
+vi.mock("../runtime/bootstrap.js", () => ({
+  FIRST_TREE_RUNTIME_DIR: ".first-tree-workspace",
+  FIRST_TREE_WORKSPACE_MARKER: ".first-tree-workspace",
+  ensureWorkspaceRuntimeDir: vi.fn((workspacePath: string) => {
+    const dir = join(workspacePath, ".first-tree-workspace");
+    mkdirSync(dir, { recursive: true });
+    return dir;
+  }),
+  writeAgentBriefing: vi.fn(),
+}));
+vi.mock("../runtime/agent-briefing.js", () => ({ buildAgentBriefing: vi.fn(() => "") }));
+vi.mock("../runtime/chat-context.js", () => ({
+  fetchChatContext: vi.fn(async () => {
+    if (!state.chatContextPromise) throw new Error("chat context gate was not initialised");
+    return state.chatContextPromise;
+  }),
+}));
+vi.mock("../runtime/source-repos.js", () => ({
+  declaredSourceRepos: vi.fn(() => []),
+  currentSourceRepoNamesFromPayload: vi.fn(() => null),
+}));
+
+import { createClaudeCodeHandler } from "../handlers/claude-code.js";
+import { writeAgentBriefing } from "../runtime/bootstrap.js";
+
+const AGENT_ID = "019e71d2-c9ec-7f11-86bf-5dfc9e873338";
+
+let workspaceRoot: string;
+let cachedConfig: AgentRuntimeConfig;
+
+function makePayload(skills: RuntimeResourceSkill[]): AgentRuntimeConfigPayload {
+  return {
+    kind: "claude-code",
+    prompt: { append: "" },
+    model: "",
+    mcpServers: [],
+    env: [],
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Approved. I reviewed the client inject-time materialization path and the server resource refresh/version-bump transaction. No blocking findings.

Verified locally:
- `pnpm --filter @first-tree/client test -- src/__tests__/claude-code-inject-materialize.test.ts` (ran client suite: 1186 passed)
- `pnpm --filter @first-tree/server test -- src/__tests__/resources-phase1.test.ts` (ran server suite: 1622 passed)
- `pnpm typecheck`
- `pnpm check` exited 0; it still reports existing warnings outside this PR.

Gandy2025 (commented):
## Review — 2 independent rounds (correctness + adversarial). Verdict: no blocking defects; fix is correct and well-tested.

I traced the real path end-to-end and ran two independent reviews (a correctness/path-trace pass and an adversarial break-it pass). Both converge: **the fix correctly fixes the bug, no confirmed correctness defects.** CI is green on every job (Lint & Type Check, Test, CodeQL, Migrations; Validate Skill skipped as expected).

### What I verified
- **Crux (hook placement) holds.** A mid-session skill-bind changes `payload.resourceSkills`, so `onlyModelChanged` (`claude-code.ts:1414-1417`) is false → Path A is skipped → **Path B restart always runs** → `materializeResourceSkills` fires (`:1455-1457`). A skill-bind can never take Path A even if a same-family model change were coincidentally batched (the skills diff breaks the byte-equality clause). The `inject → scheduleInjectedMessagesDrain → pushInjectedMessage → maybeSwitchConfig` chain is confirmed.
- **Ordering is correct.** `materializeResourceSkills` is `await`ed *before* `writeAgentBriefing` (`:1456` vs `:1459`), and the injected turn is delivered only *after* `maybeSwitchConfig` returns — so SKILL.md is on disk before the turn runs.
- **Version guard is load-bearing and correct.** `if (cached.version > appliedConfigVersion)` blocks re-materialize (and thus prune) on a lower/empty fallback while firing on a genuine bump. Independently, the empty/lower-version fallback is not actually reachable in production (`refreshIfNewer` only fetches when `incoming > cached`, and a failed refresh retains the prior config — `agent-config-cache.ts`), so the guard is defensive belt-and-suspenders.
- **Legacy-cwd branch** materializes to `legacyCwd` (`:2115`), matching the briefing/SDK paths.
- **Server linchpin is correct.** The already-bound path bumps `agentConfigs.version` iff `contentChanged`; the resource-existed-but-stale-then-newly-bound case bumps exactly once (falls through to the new-bind UPDATE). Concurrency is serialized by `SELECT … FOR UPDATE agents` — no double-bump, no lost update. The `agentConfigs` row is created at agent creation, so the `version+1` UPDATE is consistent with the existing new-bind path.
- **Tests are non-vacuous.** Reverting the Path B materialize block times out the client test; removing the server bump fails the server test (`expected 2 to be 3`). The client test does not mock `materializeResourceSkills` and uses a real temp workspace.

### Non-blocking follow-ups
1. **[Medium — track, don't lose] codex / claude-code-tui have the same mid-session gap.** `codex/sdk.ts` `refreshBriefingForActiveTurn` rewrites AGENTS.md (which will list the skill under "## Team Skills") but does **not** call `materializeResourceSkills`; their materialize is start/resume-only. So a skill bound mid-session to an *active* codex/tui agent names a SKILL.md that isn't on disk — the exact bug fixed here, left live for those handlers. Safe to defer for now because the growth funnel provisions claude-code agents, but please file a tracking issue so the parity gap isn't silently lost (or fold the codex/tui materialize in if cheap).
2. **[Low — fold into this PR] `payloadMatchesDesired` footgun.** It compares only `body`+`description` and `desiredPayload` hardcodes `metadata: {}`. Correct today (the server-owned skill never carries metadata), but a one-line comment tying the comparator to "body+description are the only mutable fields of this server-owned skill" would prevent a future silent metadata-overwrite / missed-bump.
3. **[Low — optional] Test #2 (`does not prune … lower-version empty config`)** simulates a version regression the real cache cannot produce and asserts on a mocked internal (`writeAgentBriefing.mock.calls`). It documents the defensive guard rather than catching a reachable regression — fine to keep, just not proof of a real edge. Coverage gaps (non-blocking): normal-design + legacy resume materialize, skill unbind (prune branch), multiple skills are untested.

Nothing here blocks correctness. Suggest: add the comment (2), file the tracking issue (1), then it's ready for code-owner.

yuezengwu:
Filed the handler-parity tracking issue for the non-blocking follow-up: #1380.

I am leaving the low-priority comparator comment suggestion to the PR author since it is not a correctness blocker and the PR is already approved.

yuezengwu (approved):
Re-approved latest head `4b37a2ec`.

Incremental review since my prior approval: the only new change is a 5-line explanatory comment in `packages/server/src/services/resources.ts` clarifying why the campaign scan skill diff compares `body` + `description` only. It matches the intended server-owned skill contract and introduces no behavior change.

Verification: `pnpm check` exits 0 on the latest head; it still reports the same existing warnings outside this PR.

baixiaohang (approved):
## Review by codex-assistant - PR #1378

**Recommendation**: approve
- Rationale: The change closes the client/server synchronization gap at the existing runtime config-refresh boundary without introducing a parallel resource-loading path.

**Risk level**: A
- Semantic lift: none beyond the path baseline; server resource versioning is part of the same runtime projection flow.

**PR Summary**
- Problem: A campaign scan skill bound to a reused or already-running agent could be listed in the briefing while its `SKILL.md` was missing on disk, causing the agent to improvise instead of following the rubric.
- Approach: Materialize resource skills on the Claude Code active-session restart path before rewriting the briefing, materialize on the legacy-cwd resume branch, and refresh/bump the server-owned campaign skill payload when stale content is already bound.
- Impacted modules: `packages/client/src/handlers/claude-code.ts`, `packages/client/src/runtime/resource-skills` behavior via tests, and `packages/server/src/services/resources.ts`.

**Review findings**
✅ 1. The client hook placement is coherent with the existing inject/drain config-refresh path: `payload.resourceSkills` changes force Path B restart, and materialization now happens before the updated briefing exposes the skill path.
✅ 2. The legacy-cwd resume fix writes skills to the same cwd used by that branch's briefing and SDK resume path, so it does not mix the new agent-home layout with the legacy transcript layout.
✅ 3. The server-side refresh/version bump keeps resources as the source of truth and uses `agentConfigs.version` as the existing client refetch trigger; idempotent no-change re-scans avoid version churn.
✅ 4. The cross-handler parity gap for Codex/TUI is explicitly tracked in #1380, and the latest head adds the requested comparator comment; neither remains a blocker for this Claude Code-scoped fix.

**Action taken**
- Approved latest head `4b37a2ec`.

PR #1382, opened 2026-07-01
fix(web): make group @mention reminder a placeholder + on-send tip

Gandy2025:
## What & why

Refine of #1376 after the product owner's design review. The v1 shipped a **persistent in-card hint row** + a row-emphasis on blocked send. In practice the row read as extra chrome (it stayed up while typing, so the composer was never "quiet"), and the blocked-send feedback wasn't prominent or well-placed. This reworks it to the simpler, approved shape.

### Changes
- **Drop the persistent hint row.** The rule now rides the **placeholder** — `In a group, @mention who this is for` — shown only while the composer is empty. Typing is quiet; no standing chrome.
- **Send button: greyed but clickable** when only the `@mention` is missing. A truly `disabled` button swallows clicks, so the tip could never fire on click. `aria-disabled` conveys the blocked state to AT; the mention gate is removed from `sendDisabled` and a new `sendDimmed` drives the greyed look.
- **On-send tip bubble.** A blocked send attempt (Enter *or* click) pops a transient bubble above the send button: `@mention someone, or no one gets this`. It opens **upward** (the composer hugs the screen bottom, so a hint *below* it would be off-screen), points at the button with a caret, auto-dismisses after 4s, and clears early on typing / addressing someone / the gate lifting. Fade+rise in (150ms), fade+drop out (300ms), `prefers-reduced-motion` aware. **No `@` icon** — the text already leads with `@mention`, so there's no duplicate `@` (a specific piece of review feedback).

Covers text and image sends uniformly, replacing the old image-only `uploadError` hint. Everything stays gated on the existing `sendBlockedByMentionGate`, in lockstep with the server's `enforceGroupMention`.

## Design
Approved via a dual-theme, animated effect image (bubble plays on its real timing). All values use existing design tokens; a code review verified the caret geometry points at the send button and clears the toolbar in both themes.

## Tests
- Web suite green: **1094/1094**; `typecheck` + `biome` + `lint:tokens` clean.
- Composer DOM test updated + extended: placeholder copy, tip **not** shown until a blocked send, blocked **text** send pops the tip and sends nothing, send button stays clickable when blocked, image-only blocked send pops the tip. SSR smoke asserts the new placeholder.

## Review
Two independent reviews (correctness + design/visual) — both **approve, zero blocking**. Adopted their two suggestions: `borderRadius` → `var(--radius-panel)`, and an early-out in `dismissMentionTip` to skip per-keystroke timer churn.

## Context Tree
The must-`@mention` contract is already documented in `system/cloud/chat/messaging.md`; this only changes how it's surfaced in the composer — **implementation-only, no tree write**.

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -797,6 +797,53 @@
   }
 }
 
+/* Group-mention tip bubble: entrance rises toward the send button it points at;
+   exit reverses. See chat-view's `mentionTip` state machine. */
+@keyframes mention-tip-in {
+  from {
+    opacity: 0;
+    transform: translateY(5px);
+  }
+  to {
+    opacity: 1;
+    transform: translateY(0);
+  }
+}
+@keyframes mention-tip-out {
+  from {
+    opacity: 1;
+    transform: translateY(0);
+  }
+  to {
+    opacity: 0;
+    transform: translateY(5px);
+  }
+}
+.mention-tip {
+  animation: mention-tip-in 150ms ease-out;
+}
+.mention-tip.mention-tip-out {
+  animation: mention-tip-out 300ms ease-in forwards;
+}
+/* The downward caret connecting the bubble to the send button below it. */
+.mention-tip-arrow {
+  position: absolute;
+  right: 14px;
+  bottom: -5px;
+  width: 9px;
+  height: 9px;
+  background: var(--bg-raised);
+  border-right: var(--hairline) solid var(--border-strong);
+  border-bottom: var(--hairline) solid var(--border-strong);
+  transform: rotate(45deg);
+}
+@media (prefers-reduced-motion: reduce) {
+  .mention-tip,
+  .mention-tip.mention-tip-out {
+    animation: none;
+  }
+}
+
 @keyframes context-live-halo-breathe {
   0%,
   100% {
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1173,7 +1173,7 @@ describe("page SSR smoke coverage", () => {
     const { ChatView } = await import("../workspace/center/chat-view.js");
 
     expect(renderWithClient(<ChatView agentId="agent-1" chatId="chat-1" />, createClient())).toContain(
-      "@mention someone to send in this group",
+      "In a group, @mention who this is for",
     );
 
     const readOnlyClient = createClient();
--- a/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
@@ -1331,11 +1331,23 @@ describe("ChatView", () => {
 
     const textarea = container.querySelector<HTMLTextAreaElement>("textarea");
     if (!textarea) throw new Error("Composer textarea missing");
-    // Group chat, no @mention yet: the persistent hint row states the rule, so
-    // the placeholder stays neutral (no duplicate "@ to pick a recipient").
-    expect(textarea.placeholder).toContain("Write a message");
-    expect(textarea.placeholder).not.toContain("Type @ to pick a recipient");
-    expect(container.textContent).toContain("@mention someone to send in this group");
+    // Group chat, no @mention yet: the placeholder carries the rule, and the
+    // tip bubble is NOT shown until a blocked send is actually attempted.
+    expect(textarea.placeholder).toContain("In a group, @mention who this is for");
+    expect(container.textContent).not.toContain("or no one gets this");
+
+    // Blocked TEXT send: typing without an @mention then pressing Enter pops the
+    // tip bubble and sends nothing. The send button stays clickable (not
+    // `disabled`) so a click would trigger the same tip.
+    await setValue(textarea, "no recipient here");
+    expect(container.querySelector<HTMLButtonElement>('button[aria-label="Send"]')?.disabled).toBe(false);
+    await act(async () => {
+      textarea.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true, cancelable: true }));
+    });
+    await flush();
+    await waitForText(container, "@mention someone, or no one gets this");
+    expect(chatMocks.sendChatMessage).not.toHaveBeenCalled();
+
     await setValue(textarea, "Please review @nova");
     await click(container.querySelector('button[aria-label="Send"]'));
     await waitForCondition(() => chatMocks.sendChatMessage.mock.calls.length > 0, "Expected text send");
@@ -1350,9 +1362,9 @@ describe("ChatView", () => {
       textarea.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true, cancelable: true }));
     });
     await flush();
-    // Image-only send with no @mention is blocked too; the persistent hint row
-    // carries the reason for both text and image sends (no separate line).
-    await waitForText(container, "@mention someone to send in this group");
+    // Image-only send with no @mention is blocked too; the tip bubble pops for
+    // both text and image attempts.
+    await waitForText(container, "@mention someone, or no one gets this");
     expect(chatMocks.sendFileMessageBatch).not.toHaveBeenCalled();
 
     await setValue(textarea, "@design image attached");
--- a/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
@@ -1385,6 +1385,45 @@ describe("ChatView", () => {
     await act(async () => root.unmount());
   });
 
+  it("clears the mention tip when switching to another group chat", async () => {
+    const { ChatView } = await import("../chat-view.js");
+    const { container, queryClient, root } = await renderDom(
+      <ChatView agentId="agent-1" chatId="chat-1" />,
+      undefined,
+      "/",
+    );
+
+    const textarea = container.querySelector<HTMLTextAreaElement>("textarea");
+    if (!textarea) throw new Error("Composer textarea missing");
+    // Trigger the tip in chat-1 (group, no @mention).
+    await setValue(textarea, "no recipient");
+    await act(async () => {
+      textarea.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true, cancelable: true }));
+    });
+    await flush();
+    await waitForText(container, "@mention someone, or no one gets this");
+
+    // Switch to another group chat on the SAME long-lived ChatView instance
+    // (identical provider wrappers → React updates the chatId prop rather than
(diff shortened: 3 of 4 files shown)


yuezengwu (approved):
Implementation goal/core change: this PR changes only how the Web composer surfaces the existing group-chat @mention requirement. The persistent in-card reminder row moves into the empty-composer placeholder, the send button stays visually dimmed but clickable when text/images are present and only the @mention is missing, and `handleSend` now shows the transient tip for both Enter and click attempts while still returning before any text or image send reaches the API.

I reviewed the mention gate, text send path, image batch path, `sendDisabled`/`sendDimmed` split, tip timer/dismissal lifecycle, CSS bubble/caret styling, and the updated DOM/SSR coverage. The server contract remains intact: group sends still require structured routing metadata, and image batches still carry the resolved mentions before posting. No core data-structure or database changes in this PR.

Verification run locally:
- `pnpm --filter @first-tree/web test -- src/pages/workspace/center/__tests__/chat-view-dom.test.tsx src/pages/__tests__/page-ssr-smoke.test.tsx` (web suite passed: 1094/1094)
- `pnpm exec biome check packages/web/src/index.css packages/web/src/pages/workspace/center/chat-view.tsx packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx` (exit 0; one pre-existing `!important` warning outside this diff)

No blocking findings.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1382

- Rationale: The new transient mention tip can leak across chat switches, so a group composer can show the bubble before any blocked send in that chat.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed -> C
- Semantic lift: none; the diff stays in composer presentation/tests and does not touch Web subscriptions, WS frame parsing, or data-layer contracts.

**PR Summary**
- Problem: Replace the persistent group-mention hint row with quieter placeholder guidance and stronger blocked-send feedback.
- Approach: Keep the missing-mention send button clickable via `aria-disabled`, route blocked text/image attempts through `handleSend`, and show an auto-dismissed bubble above the send button.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/index.css`, composer SSR/DOM tests.

**Review findings**
❌ 1. Reset the mention-tip state when the viewed chat changes. [R3 / `packages/web/src/pages/workspace/center/chat-view.tsx:1391`]

`ChatView` is long-lived across chat switches, but `mentionTip` and its timers are scoped to the component, not to `chatId`. If I trigger a blocked send in group chat A and switch to group chat B before the 4s hold expires, the render at `packages/web/src/pages/workspace/center/chat-view.tsx:3829` still sees `mentionTip !== "hidden"` and shows `@mention someone, or no one gets this` in chat B before any blocked send happened there. The current cleanup at `packages/web/src/pages/workspace/center/chat-view.tsx:3182` only dismisses when `sendBlockedByMentionGate` becomes false, so it does not clear when both chats are group/no-mention. Please hard-hide/clear the tip on `chatId` changes, not just by playing the exit animation, and add a DOM test for switching between two group chats after a blocked send.

**Action taken**
- Submitted request changes

yuezengwu (changes requested):
Follow-up after re-checking the new review finding: I agree this is blocking, and this supersedes my earlier approve.

`ChatView` is long-lived across chat switches, and `mentionTip` plus its timers are component-scoped rather than `chatId`-scoped. The current cleanup only dismisses when `sendBlockedByMentionGate` becomes false; switching from one group/no-mention composer to another keeps that gate true, so the transient tip can render in the next chat before any blocked send happened there. Switching to a non-gated chat can also paint the stale tip for a frame/exit animation instead of hard-hiding it.

Please hard-clear the tip state and timers on `chatId` changes, and add coverage for switching between two group chats after a blocked send attempt.

Gandy2025:
Good catch on the cross-chat tip leak (R3) — confirmed and fixed in 20bb3bb9.

`ChatView` is long-lived across chat switches and the tip state/timers were component-scoped, so a blocked send in group A could paint the bubble in group B within the 4s hold (the gate-lift dismiss doesn't fire when both are group/no-mention). Fix: an effect keyed on `chatId` that hard-clears the timers and sets the tip to `"hidden"` immediately (no exit animation) on every chat switch. Added a DOM test that triggers the tip in chat-1, switches the same ChatView instance to chat-2 (a second group), and asserts the bubble is gone.

`pnpm --filter @first-tree/web test` green (1095/1095), typecheck + biome + lint:tokens clean.

yuezengwu (changes requested):
Implementation goal/core change in this update: the PR now tries to close the R3 cross-chat leak by adding a `chatId`-keyed reset for the transient mention tip and a DOM test that switches the same long-lived `ChatView` instance from one group chat to another after a blocked send.

Blocking finding: the reset currently uses passive `useEffect` (`packages/web/src/pages/workspace/center/chat-view.tsx:1446`). That clears timers and sets `mentionTip` to `"hidden"`, but only after React commits the new chat render; passive effects are not guaranteed to run before the browser paints. So the stale bubble can still be committed/rendered in the destination chat for a frame before the effect clears it. This is the same user-visible class as the original issue, just shortened from a 4s leak to a possible one-frame leak.

Please make the chat-switch reset pre-paint, e.g. use `useLayoutEffect` here (this file already uses layout effects for chat-switch visual invariants) or gate rendering by the chat id that created the tip. The important property is that a new chat never paints a tip that originated in the previous chat, not only that it disappears after the passive effect flushes.

Verification run locally on `20bb3bb9`:
- `pnpm --filter @first-tree/web test -- src/pages/workspace/center/__tests__/chat-view-dom.test.tsx src/pages/__tests__/page-ssr-smoke.test.tsx` (web suite passed: 1095/1095)
- `pnpm exec biome check packages/web/src/index.css packages/web/src/pages/workspace/center/chat-view.tsx packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx` (exit 0; pre-existing `!important` warning in `index.css` outside this diff)

No core data-structure or database changes in this PR.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1382

- Rationale: The latest fix clears the stale mention tip in a passive effect, so the destination chat can still commit/paint one stale frame before the reset runs.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed -> C
- Semantic lift: none; this remains composer presentation/state only, with no Web subscription, WS frame parsing, or data-layer contract change.

**PR Summary**
- Problem: Replace the persistent group-mention reminder with placeholder guidance and an on-send tip, while preserving the existing explicit-recipient requirement.
- Approach: Keep blocked sends in `handleSend`, show a transient tip, and in `20bb3bb9` attempt to clear that tip when `chatId` changes.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx`, composer CSS/SSR tests.

**Review findings**
❌ 1. Make the chat-switch tip reset pre-paint. [R3 / `packages/web/src/pages/workspace/center/chat-view.tsx:1446`]

The `chatId`-keyed reset currently uses passive `useEffect`. That does clear the timers and set `mentionTip` to `"hidden"`, but only after React has committed the new chat render; passive effects are not guaranteed to run before the browser paints. So a tip triggered in chat A can still be rendered in chat B for a frame before the reset lands. The invariant we need is stronger: a new chat should never paint a tip that originated in a previous chat. Please move this reset to a pre-paint path, e.g. `useLayoutEffect` (already used in this file for chat-switch visual invariants), or key/gate the rendered tip by the chat id that created it.

**Action taken**
- Submitted request changes

Gandy2025:
Fixed pre-paint in 1bdb4cb7 — thanks, the one-frame window was real.

Made the invariant render-time rather than effect-timed: `flashMentionTip` now stamps `mentionTipChatId` with the chat that owns the tip, and the bubble only renders while `mentionTipChatId.current === chatId`. So a tip from a previous chat can never be committed/painted in the next one, independent of when effects run. The chat-switch reset is also promoted to `useLayoutEffect` (pre-paint) to drain the timers and hard-hide. The DOM test now asserts the bubble is absent immediately after the switch commit, before any flush.

Green on 1bdb4cb7: web tests 1095/1095, typecheck + biome + lint:tokens clean.

yuezengwu (approved):
Implementation goal/core change in the latest update: `1bdb4cb7` closes the cross-chat mention-tip leak by making the no-stale-tip invariant render-time, not effect-timed. `flashMentionTip` now stamps the owning `chatId`, the bubble only renders when that owner matches the currently viewed chat, and the chat-switch reset is also a pre-paint `useLayoutEffect` that drains timers and hard-hides the tip.

I re-reviewed the text/image send gate, `sendDisabled`/`sendDimmed` split, tip lifecycle, cross-chat reset path, render guard, and the updated DOM test. The previous R3 findings are addressed: a tip created in chat A cannot commit/paint in chat B, including the pre-effect window, and the same long-lived `ChatView` switch is covered by the test assertion before `flush()`.

Verification run locally on `1bdb4cb7`:
- `pnpm --filter @first-tree/web test -- src/pages/workspace/center/__tests__/chat-view-dom.test.tsx src/pages/__tests__/page-ssr-smoke.test.tsx` (web suite passed: 1095/1095)
- `pnpm exec biome check packages/web/src/index.css packages/web/src/pages/workspace/center/chat-view.tsx packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx` (exit 0; pre-existing `!important` warning in `index.css` outside this diff)

No blocking findings. No core data-structure or database changes in this PR, and no Context Tree write is needed because the durable group-mention contract was already documented; this remains presentation-layer behavior.

baixiaohang (approved):
## Review by codex-assistant - PR #1382

**Recommendation**: approve
- Rationale: The explicit-recipient send guard remains intact, and the latest `1bdb4cb7` fix closes both the cross-chat tip leak and the one-frame paint window.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed -> C
- Semantic lift: none; this stays in composer presentation/state/tests and does not touch Web subscriptions, WS frame parsing, or data-layer contracts.

**PR Summary**
- Problem: Replace the persistent group-mention hint row with quieter placeholder guidance plus an on-send tip.
- Approach: Keep the send button clickable when only the group mention is missing, keep text/image sends gated in `handleSend`, show a transient tip on blocked attempts, and scope that tip to its origin chat.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/index.css`, composer DOM/SSR tests.

**Review findings**
✅ 1. Previous R3 finding resolved: `mentionTipChatId` gates rendering by the chat that created the tip, and the `chatId` reset now runs in `useLayoutEffect`, so a stale tip from the previous chat cannot commit/paint in the destination chat.
✅ 2. The group-send contract still lines up with `system/cloud/chat/messaging.md`: unaddressed group text/image sends return before any API call, while the server remains the final enforcement point.

**Action taken**
- Approved

PR #1389, opened 2026-07-01
feat(skill): first-tree-welcome v2 — offer Context Tree build + fan out first tasks

Gandy2025:
## What & why

Makes the onboarding welcome chat a **launcher** that both delivers value and actively guides Context Tree setup, and showcases First Tree's parallelism.

After showing concrete code understanding, the agent offers ONE multi-select first-task menu:
- **the recommended value tasks, bundled into a single choice**, and
- **"Build your Context Tree"** — now a *first-class* option (was previously a forbidden one), offered only to an **admin whose team tree is missing/empty**, and only after value.

On selection the agent **fans work out into parallel chats — one per value task, plus one for the tree build** — so the user feels multiple streams progressing at once. Each spawned chat is opened with `first-tree chat create --to <its own agent name>` and a **self-contained task brief**. The launcher never creates/binds/seeds the tree itself; `first-tree-seed` owns that in the spawned tree chat.

Value-first ordering, consent gating, and invitee/role guardrails are preserved.

## Spawn mechanic (validated)

Self-addressing (`--to <self>`) is what wakes the agent in the spawned chat: the server rewrites the opening message's sender to the agent's manager-human (`self_target_manager_human` in `chat.ts`), so the agent is mentioned by a non-self sender and gets a `notify=true` wake. **Verified by a live smoke test** (the self-created chat woke the agent with no human message in between). The skill spells out the mechanic and, critically, requires a self-contained brief because the woken agent cannot tell the chat was self-spawned (it reads as a fresh task from the user).

## Review

- 2 independent reviews (internal consistency; design-faithfulness + spawn-mechanic correctness against `chat.ts`/CLI) — no blocking findings; the two suggested strengthenings (forbid the `--to <user>` variant; sharpen the "you won't know it was self-spawned" nuance) are applied.
- `python3 scripts/quick_validate_skill.py ./skills/first-tree-welcome` → valid. Version bumped 1.0.4 → 1.0.5 (frontmatter + VERSION), `agents/openai.yaml` description synced.

## Related

- **Pairs with #1391** (`first-tree-seed` Step 0, by liuchao-staff): the tree build spawned here runs `first-tree-seed`, which #1391 updates to provision via `tree init` when unbound (resolving the "seed refuses unbound" contract mismatch). The two are the two halves of Phase 1 — **both draft; this PR will not merge before #1391**, they land together / seed slightly first.
- Depends on merged `tree init` (#1379) + ejs templates (#1384).
[...]

PR #1379, opened 2026-07-01
feat(tree): add `first-tree tree init` — agent-gh Context Tree creation

liuchao-001:
## What

Reintroduces `first-tree tree init` — deleted in #848 when Context Tree provisioning moved server-side — now in the **agent-driven** direction: a new team Context Tree repo is created with the **user's local `gh`** instead of the server GitHub App.

Landing slice of the merged proposal [first-tree-context#618](https://github.com/agent-team-foundation/first-tree-context/pull/618). Scope confirmed with @liuchao-001: **Phases A–C**. Retiring the server bootstrap and slimming the App (Phase D) is a follow-up — the server one-click path stays in place, so the two run side by side.

## The command

`first-tree tree init`:
1. `gh repo create <owner>/<name>` — one path for both **user** and **org** owners (drops the server's user/org split; personal-account trees stop needing a special user-token path).
2. Scaffolds a **minimal valid tree**: root `NODE.md` + `members/NODE.md` (non-empty owner, so `tree tree` renders it) + a creator `members/<login>/NODE.md`. `tree verify` hard-fails on a `members/` dir with zero member nodes, so the creator node is load-bearing.
3. Resolves **all bind preconditions before** the remote `gh repo create`, so nothing irreversible happens until they pass:
   - auth + org + **admin role** (binding is admin-only);
   - **no existing binding** — refuses to replace a team's active tree unless `--rebind` is passed (mirrors the server one-click 409);
   - **installation account** — creates the repo under the team's GitHub App installation account (an install only covers its own account, so `local-user/<repo>` would bind an org to a repo its install can't read); a mismatching `--owner` fails.
4. Self-verifies, pushes, then binds the org via the existing `settings/context_tree` write.

Flags: `--owner`, `--name`, `--title`, `--public`, `--dir`, `--with-workflow`, `--no-bind`, `--rebind`, `--org`.

## GitHub App installation coverage

Web snapshots and the Context Tree reviewer read the tree through the team's GitHub App installation, so a new repo must be within reach. Nuance (thanks @baixiaohang / bestony):

- GitHub **auto-attaches** a repo to an installation only when the **App itself** creates it (installation token).
- `tree init` creates the repo as the **user** with local `gh`, so an **all-repositories** install already covers it (once the repo is under the install account, per precondition 3), but a **selected-repositories** install does **not** auto-cover it.
[...]

PR #1391, opened 2026-07-01
feat(skill): first-tree-seed Step 0 — create the tree via tree init when none exists

liuchao-001:
## What

The welcome-v2 launcher (#1389) spawns a "Build your Context Tree" chat that runs `first-tree-seed` to **create, bind, and seed** the tree. Before this, `first-tree-seed` refused an unbound workspace ("binding is an operator/human surface"). This PR gives seed a **Step 0** that self-provisions via the merged `first-tree tree init` (#1379), so the agent-driven build-tree flow works end to end.

Phase C of proposal [first-tree-context#618](https://github.com/agent-team-foundation/first-tree-context/pull/618); the seed half of the Welcome-v2 Phase 1 (welcome orchestration = #1389).

## Change

Reframe the entry gate from the two-state empty-tree self-check into a **three-state Step 0**:

- **A — no tree yet** (workspace not bound / no `context_tree` binding) → run `first-tree tree init --title "<team>" --dir "<workspaceRoot>/<manifest.tree>"` (creates the repo under the App installation account, seeds a minimal valid tree, pushes, binds). `--dir` is load-bearing: `tree init` otherwise checks out to `<cwd>/<repo>`, but a managed workspace reads the tree at `<workspaceRoot>/<manifest.tree>` (usually `context-tree`) and seed doesn't rewrite `workspace.json`. Admin-only + needs authenticated `gh`; otherwise surface the gap and stop.
- **B — bound but unseeded** (bootstrap root/members, no top-level domains) → proceed to Phase 1. Phase 1 **extends** the bootstrap root node / `members/` index rather than recreating them.
- **C — already seeded** (top-level domain dirs) → refuse (unchanged).

Also flips the "does not do" boundary (seed self-provisions via `tree init`, not the server `/initialize` — the two coexist) and updates the description / intro / when-to-use framing.

## Eval coverage (Step 0 / state A)

The seed skill-eval gains an `unbound` state + a `unbound-tree-inits-with-dir` gate case (floor now 5 gate cases). The grader passes **only** when the agent invokes `first-tree tree init` with a `--dir` whose basename is `context-tree` — omitting `--dir` (default `<team>-context-tree` checkout) or a wrong `--dir` fails, catching the handoff regression; real repo creation stays a forbidden side effect. Verified: typecheck + biome clean, seed suite unit tests 33/33, floor validation, and the **real codex eval graded the case PASS**.

## Coordination
- Paired with **#1389** (welcome v2, green) — welcome does not merge before this; seed lands first or together.
- Paired active-tree update: **[first-tree-context#660](https://github.com/agent-team-foundation/first-tree-context/pull/660)** — `onboarding.md` records the combined welcome-launcher + build-tree-first-class + fan-out contract that this seed Step 0 is the seed half of.
- Phase-2 `/initialize` retirement (collides with #618 Phase D) is out of scope here and stays coordinated with the tree-init line.

### DIFF
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -443,8 +443,7 @@ describe("first-tree-seed grader", () => {
       casePassed(
         findCase("unbound-tree-inits-with-dir"),
         baseMetrics({
-          finalResponse:
-            'No tree yet. Running: first-tree tree init --title "Apollo" --dir ./context-tree to create and bind it.',
+          finalResponse: "No tree yet. Created and bound the Context Tree at ./context-tree.",
           skeletonObserved: false,
           sourceEvidenceReadObserved: false,
           sourceWorktreeCreated: false,
@@ -460,7 +459,7 @@ describe("first-tree-seed grader", () => {
       casePassed(
         findCase("unbound-tree-inits-with-dir"),
         baseMetrics({
-          finalResponse: 'No tree yet. Running: first-tree tree init --title "Apollo" to create and bind it.',
+          finalResponse: "No tree yet. Created and bound the Context Tree.",
           skeletonObserved: false,
           sourceEvidenceReadObserved: false,
           sourceWorktreeCreated: false,
@@ -485,6 +484,78 @@ describe("first-tree-seed grader", () => {
         }),
       ),
     ).toBe(false);
+  });
+
+  it("fails unbound-tree case when tree init is only described in prose, never invoked", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-prose-only-"));
+    try {
+      // The model DESCRIBES `tree init --dir .../context-tree` in its final
+      // message but never actually invokes it: no captured argv, no command
+      // event. The invocation signal must not fire on prose alone.
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              item: {
+                text: `I would run: first-tree tree init --title "Apollo" --dir ${join(tempRoot, "context-tree")}`,
+                type: "agent_message",
+              },
+              type: "item.completed",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.finalResponse).toContain("first-tree tree init");
+      expect(metrics.treeInitObserved).toBe(false);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+      expect(casePassed(findCase("unbound-tree-inits-with-dir"), metrics)).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("fails unbound-tree case when it materializes a source worktree or reads source content", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-source-read-"));
+    try {
+      // Step 0 stops before Phase 1 source work; this case declares
+      // requireWorktree/requireSourceRead false. A run that reads source
+      // content has gone past Step 0 and must fail even with a correct
+      // tree init --dir.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", join(tempRoot, "context-tree")],
+            phase: "model",
+            type: "first_tree_call",
+          },
+          {
+            event: {
+              command: "cat worktrees/seed-source-repo/README.md",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+      expect(metrics.sourceEvidenceReadObserved).toBe(true);
+      expect(casePassed(findCase("unbound-tree-inits-with-dir"), metrics)).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
   });
 
   it("detects tree init --dir context-tree from captured first-tree argv", () => {
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -36,6 +36,8 @@ function baseMetrics(overrides: Partial<EvalMetrics>): EvalMetrics {
     sourceEvidenceReadObserved: true,
     sourceRepoChanged: false,
     sourceWorktreeCreated: true,
+    treeInitObserved: false,
+    treeInitWithContextTreeDirObserved: false,
     workspaceManifestReadObserved: true,
     writeSkillFileReadObserved: true,
     ...overrides,
@@ -431,6 +433,171 @@ describe("first-tree-seed grader", () => {
 
       expect(metrics.sourceEvidenceReadObserved).toBe(true);
       expect(metrics.forbiddenActionHits).toContain("continue_seed");
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("passes unbound-tree case when Step 0 runs tree init with a context-tree --dir", () => {
+    expect(
+      casePassed(
+        findCase("unbound-tree-inits-with-dir"),
+        baseMetrics({
+          finalResponse:
+            'No tree yet. Running: first-tree tree init --title "Apollo" --dir ./context-tree to create and bind it.',
+          skeletonObserved: false,
+          sourceEvidenceReadObserved: false,
+          sourceWorktreeCreated: false,
+          treeInitObserved: true,
+          treeInitWithContextTreeDirObserved: true,
+        }),
+      ),
+    ).toBe(true);
+  });
+
+  it("fails unbound-tree case when tree init omits the context-tree --dir (the regression)", () => {
+    expect(
+      casePassed(
+        findCase("unbound-tree-inits-with-dir"),
+        baseMetrics({
+          finalResponse: 'No tree yet. Running: first-tree tree init --title "Apollo" to create and bind it.',
+          skeletonObserved: false,
+          sourceEvidenceReadObserved: false,
+          sourceWorktreeCreated: false,
+          treeInitObserved: true,
+          treeInitWithContextTreeDirObserved: false,
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  it("fails unbound-tree case when the model never routes to tree init", () => {
+    expect(
+      casePassed(
+        findCase("unbound-tree-inits-with-dir"),
+        baseMetrics({
+          finalResponse: "The tree is empty; here is the Phase 1 skeleton for approval.",
+          skeletonObserved: false,
+          sourceEvidenceReadObserved: false,
+          sourceWorktreeCreated: false,
+          treeInitObserved: false,
+          treeInitWithContextTreeDirObserved: false,
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  it("detects tree init --dir context-tree from captured first-tree argv", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", join(tempRoot, "context-tree")],
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+      expect(metrics.forbiddenSideEffectHits).toEqual([]);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("detects a tree init without a context-tree --dir as the regression", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-nodir-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              command: "first-tree tree init --title 'Apollo Console'",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+      // tree init is the expected action for this case, so it is not a
+      // forbidden side effect even without the correct --dir.
+      expect(metrics.forbiddenSideEffectHits).toEqual([]);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("treats tree init --dir pointing at the default repo dir as the regression", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-wrongdir-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              command: "first-tree tree init --title 'Apollo Console' --dir ./apollo-console-context-tree",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("still flags real repo creation as a forbidden side effect in the unbound case", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-repo-create-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--dir", join(tempRoot, "context-tree")],
+            phase: "model",
+            type: "first_tree_call",
+          },
+          {
+            event: {
+              command: "gh repo create apollo-context-tree --private",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+      expect(metrics.forbiddenSideEffectHits.length).toBeGreaterThan(0);
+      expect(casePassed(findCase("unbound-tree-inits-with-dir"), metrics)).toBe(false);
     } finally {
       rmSync(tempRoot, { force: true, recursive: true });
     }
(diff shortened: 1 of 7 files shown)


yuezengwu (changes requested):
Summary: this PR reframes `first-tree-seed` from the old bound-empty self-check into a three-state Step 0: create/bind via `tree init` when no tree exists, continue when the tree only has bootstrap root/members, and refuse when top-level domains already exist. It also changes Phase 1 to extend the `tree init` bootstrap root/members instead of recreating them.

Blocking finding:
- `skills/first-tree-seed/SKILL.md:65`: Step 0 runs `first-tree tree init --title "<team display name>"` without pinning the local checkout directory to the workspace tree path. The CLI defaults its local tree root to `resolve(process.cwd(), repoName)` (`apps/cli/src/commands/tree/init.ts:403`), while managed workspaces resolve the tree from `<workspaceRoot>/<manifest.tree>` (usually `context-tree`). This same skill now says seed does not write workspace-root `workspace.json` (`skills/first-tree-seed/SKILL.md:717`), so a successful Step 0 can create and bind `<workspaceRoot>/<team>-context-tree` and then proceed to Phase 1 against `<workspaceRoot>/context-tree`, which is missing or stale. Please make the handoff explicit before proceeding to state B, e.g. run `tree init --dir "<workspaceRoot>/<manifest.tree>"` when the manifest tree path is known, or define the runtime refresh/clone/update step that makes `<workspaceRoot>/<manifest.tree>` point at the newly created repo before Phase 1 reads/writes it.

Also required before merge: add the seed skill-eval coverage for Step 0/state A. The current suite only models `treeState: "empty" | "nonempty"` and pre-creates `context-tree`, so it cannot catch the no-tree -> `tree init` -> Phase 1 handoff regression above.

liuchao-001:
Thanks — the handoff finding is exactly right, and fixed in `a37099b7`.

**Blocking finding (local checkout path):** Step 0 now pins the local clone to the managed tree path:

```bash
first-tree tree init --title "<team display name>" --dir "<workspaceRoot>/<manifest.tree>"
```

`tree init` defaults its checkout to `<cwd>/<repo>` (`init.ts` `resolveRepoOwner`/dir logic), but managed workspaces read/write the tree at `<workspaceRoot>/<manifest.tree>` (usually `context-tree`) and seed doesn't rewrite `workspace.json` — so without `--dir` Step 0 would create+bind `<workspaceRoot>/<team>-context-tree` while Phase 1 operated on a missing/stale `context-tree`. With `--dir` the fresh clone lands exactly where Phase 1 + the runtime expect it. (Fully-unbound workspace with no manifest tree name → conventional `<workspaceRoot>/context-tree`.) `quick_validate_skill.py` passes.

**Seed skill-eval state A (still open):** you're right the current suite only models `treeState: "empty" | "nonempty"` and pre-creates `context-tree`, so it can't catch this handoff. Adding an `unbound` state + a case that asserts the `tree init --dir <workspaceRoot>/<manifest.tree>` → Phase 1 handoff is the next commit before I take this out of draft. This is paired with #1389 (welcome v2) and lands with / just before it.

yuezengwu:
Confirmed `a37099b7` fixes the local checkout-path blocker: Step 0 now uses `tree init --dir "<workspaceRoot>/<manifest.tree>"`, and the added prose explains the managed-workspace handoff clearly.

I am keeping my changes-requested review in place until the state-A/no-tree seed eval coverage lands, since that is the remaining merge blocker called out in the review and in your follow-up. Local recheck on this head: `python3 scripts/quick_validate_skill.py ./skills/first-tree-seed`, `pnpm validate:skill`, and `git diff --check origin/main...HEAD` all pass.

baixiaohang (commented):
## Review by codex-assistant - PR #1391

**Recommendation**: comment
- Rationale: The direction matches the `tree init` / welcome-v2 contract, but I would not approve while the existing Step 0 handoff blocker and missing state-A eval coverage remain open.

**Risk level**: C
- Path baseline: only `skills/first-tree-seed/SKILL.md` changed -> C.
- Semantic lift: none under the routing rules; this is shipped skill guidance, not CLI/server implementation.

**PR Summary**
- Problem: the Welcome-v2 tree-build chat needs `first-tree-seed` to work even when the team does not yet have a bound Context Tree, instead of immediately refusing the unbound workspace.
- Approach: adds a Step 0 state gate to create and bind the tree via `first-tree tree init`, continue when only bootstrap root/members exist, and refuse once top-level domains exist; Phase 1 is updated to extend the bootstrap root/members instead of recreating them.
- Impacted modules: shipped `first-tree-seed` skill payload and its expected seed lifecycle.

**Review findings**
⚠️ 1. Existing blocker still stands: Step 0 invokes `first-tree tree init --title "<team display name>"` without pinning the local tree directory to `<workspaceRoot>/<manifest.tree>` or defining the runtime refresh/clone handoff, while `tree init` defaults to `./<repoName>` and the skill explicitly says it does not write workspace-root `workspace.json`. I am not adding a duplicate inline thread because yuezengwu's `CHANGES_REQUESTED` review already covers this. [R4 / `skills/first-tree-seed/SKILL.md:65`, `apps/cli/src/commands/tree/init.ts:403`, `skills/first-tree-seed/SKILL.md:717`]
⚠️ 2. The new state-A path still needs eval coverage before merge: the seed suite currently models only `treeState: "empty" | "nonempty"` and pre-creates `context-tree`, so it cannot catch the no-tree -> `tree init` -> Phase 1 handoff regression above. The PR body acknowledges this; keep it as a required fix, not a follow-up. [R5 / `packages/skill-evals/src/suites/first-tree-seed/types.ts:5`, `packages/skill-evals/src/suites/first-tree-seed/fixture.ts:184`]
✅ 3. The Phase 1 wording now correctly treats the `tree init` root/members bootstrap as expected structure to extend, which aligns the seed flow with the minimal valid tree scaffold.

**Action taken**
- Submitted a comment review only; no approval.

yuezengwu:
Clarification for current head: Finding 1 in the latest comment review is stale. At `a37099b7`, Step 0 now runs `first-tree tree init --title "<team display name>" --dir "<workspaceRoot>/<manifest.tree>"`, so the checkout-path handoff blocker is fixed.

The remaining required fix is still Finding 2: add state-A/no-tree seed eval coverage for the `tree init --dir <workspaceRoot>/<manifest.tree>` -> Phase 1 handoff.

baixiaohang:
Correction to my earlier comment review: Finding 1 is stale on the current head.

I rechecked `a37099b7`; Step 0 now runs:

```bash
first-tree tree init --title "<team display name>" --dir "<workspaceRoot>/<manifest.tree>"
```

That resolves the local checkout-path handoff I called out. The remaining required fix is the same as Finding 2: add state-A/no-tree seed eval coverage for the `tree init --dir "<workspaceRoot>/<manifest.tree>"` -> Phase 1 handoff.

liuchao-001:
State-A seed-eval coverage landed in `60015aac` — this closes the last blocking finding (@yuezengwu finding 2 / @baixiaohang finding 2). #1391 is now marked ready.

`packages/skill-evals/src/suites/first-tree-seed/`:
- New `unbound` tree state + `create_tree_via_init` action. The fixture omits the manifest `tree` field and does **not** pre-create `context-tree` (genuinely unbound); `validateTreeUnbound` asserts that. The eval's real-GitHub/create/push/bind prohibition is kept intact.
- New gate case `unbound-tree-inits-with-dir` (floor now 5 gate cases). The grader passes **only** when the agent invokes `first-tree tree init` with a `--dir` whose basename is `context-tree` — the default `<team>-context-tree` checkout (i.e. omitting `--dir`) or a wrong `--dir` fails, which is exactly the handoff regression flagged. Real repo creation stays a forbidden side effect.

Verified: typecheck + biome clean; seed suite unit tests **33/33** (7 new grader tests: pass-with-dir / fail-no-dir / fail-wrong-dir / forbidden-real-create); floor validation passes; and the **real codex eval graded the new case PASS** (model ran `tree init --title ... --dir ".../context-tree"`, the shim blocked the side effect, `contextTreeChanged: false`, no forbidden hits).

Paired with #1389 (welcome v2), which is green; ordering unchanged — seed lands first or together.

liuchao-001:
@codex review

yuezengwu (approved):
## Review by yzw-codex - PR #1391

**Recommendation:** approve

**Implementation goal / core change:** this makes `first-tree-seed` handle the welcome-v2 build-tree flow from a truly unbound workspace by adding Step 0: create/bind via `first-tree tree init --title "<team>" --dir "<workspaceRoot>/<manifest.tree>"`, continue for the bootstrap-only tree state, and still refuse already-seeded trees. Phase 1 now extends the `tree init` bootstrap root and members scaffold instead of recreating it.

The two previous blockers are resolved at head `60015aac`:

- The checkout handoff is now pinned with `--dir "<workspaceRoot>/<manifest.tree>"`, so the local clone lands at the runtime-managed tree path instead of `tree init`'s default `<cwd>/<repo>` path.
- The seed eval suite now models `treeState: "unbound"` with the `unbound-tree-inits-with-dir` gate case. The grader requires a `tree init` invocation with a `--dir` basename of `context-tree`, fails the no-`--dir` and wrong-dir regressions, and still forbids real repo creation or other persistent side effects in eval.

No DB/schema or Context Tree data-structure change is introduced here; this is shipped skill guidance plus eval-contract coverage. The paired active-tree/onboarding contract is tracked separately in `first-tree-context#660`.

**Verification:**

- GitHub checks are green: Validate Skill, Lint & Type Check, Test, CodeQL, and branch/legacy-name gates.
- Local `git diff --check origin/main...HEAD` passed.
- Local `python3 scripts/quick_validate_skill.py ./skills/first-tree-seed` passed.
- Local full skill validation passed via `pnpm validate:skill`.
- I could not rerun the local vitest target in this workspace because `node_modules` is not installed here (`vitest: command not found`), but the PR's GitHub `Test` check passed on this head.

liuchao-001:
@yuezengwu gentle nudge — the state-A eval coverage you requested (finding 2) landed in `60015aac`; your changes-requested review was on `f08783a4`, before it. New `unbound` tree state + `unbound-tree-inits-with-dir` gate case whose grader passes **only** on `tree init --dir <basename=context-tree>` (missing/wrong `--dir` fails → catches the handoff regression), and real repo creation stays a forbidden side effect. Verified: typecheck + biome + seed unit tests 33/33 + floor validation, and the **real codex eval graded the case PASS**. Could you re-review `60015aac` to clear the stale changes-requested? It's the sole remaining gate — #1389 (welcome, which you already approved) is held behind it.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1391

- Rationale: the Step 0 direction is right, but the new state-A eval can still pass without proving the model actually invoked `first-tree tree init --dir ...`.

**Risk level**: C
- Path baseline: only `skills/**` and `packages/skill-evals/**` changed -> C.
- Semantic lift: none under the routing rules; this is shipped skill guidance plus eval-contract coverage.

**PR Summary**
- Problem: the welcome-v2 build-tree chat needs `first-tree-seed` to work from a truly unbound workspace instead of refusing before a tree exists.
- Approach: add Step 0 so seed creates/binds via `first-tree tree init --dir <workspace tree>`, continue for bootstrap-only trees, and refuse already-seeded trees; update Phase 1 to extend the `tree init` root/members scaffold; add a state-A seed eval.
- Impacted modules: shipped `first-tree-seed` skill payload and `packages/skill-evals` seed suite.

**Review findings**
❌ 1. The new state-A gate can pass on final-response text alone, without an actual `tree init` invocation. `deriveTreeInitObservation()` scans `finalResponse` and sets `treeInitWithContextTreeDirObserved` when the assistant merely says something like `first-tree tree init --dir ./context-tree`; then `casePassed()` for `create_tree_via_init` accepts that boolean as sufficient. That means a run that only tells the user the command, or reports what it would do, can satisfy the advertised “invokes `first-tree tree init` with `--dir`” gate without any `first_tree_call` / command event. Please make the positive invocation signal come from captured command/shim events only (and add a regression test for “final response mentions the command but no call was made”). [R5 / `packages/skill-evals/src/suites/first-tree-seed/grader.ts:463`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:655`]

⚠️ 2. `casePassed()` also ignores the case’s `requireWorktree: false` / `requireSourceRead: false` constraints for `create_tree_via_init`. The summary scorer would mark `process_pass=false` if the model materialized/read sources in the unbound gate, but the overall pass condition still returns true as long as `treeInitWithContextTreeDirObserved` is true. Either fold those negative source-process checks into the pass condition, or make the case expectation explicit if source exploration after a blocked `tree init` is intentionally allowed. [R5 / `packages/skill-evals/src/suites/first-tree-seed/cases.ts:90`, `packages/skill-evals/src/suites/first-tree-seed/summary.ts:21`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:655`]

**Action taken**
- Submitted request changes.

liuchao-001:
@codex-assistant Both findings are fixed in `0391a3ef` (`packages/skill-evals/src/suites/first-tree-seed/`).

**Finding 1 — state-A gate could pass on final-response prose alone.** `deriveTreeInitObservation()` no longer consults `finalResponse` at all. The tree-init invocation signal is now derived **only** from captured evidence: the shimmed `firstTreeArgv` vectors plus real codex command/exec strings (via `collectCommandStrings`). The helper that scans a string is renamed `textTreeInitSignal → commandTreeInitSignal` and documented as "captured COMMAND strings only, never free-text prose." A run that merely *describes* `first-tree tree init --dir .../context-tree` without invoking it now leaves both `treeInitObserved` and `treeInitWithContextTreeDirObserved` false.

Regression test added — *"fails unbound-tree case when tree init is only described in prose, never invoked"*: an `agent_message` mentioning the command, with empty `firstTreeArgv` and no command event → asserts `finalResponse` contains the text, `treeInitObserved === false`, `treeInitWithContextTreeDirObserved === false`, `casePassed(...) === false`.

**Finding 2 — `casePassed` for `create_tree_via_init` ignored `requireWorktree`/`requireSourceRead: false`.** It now folds in the negative source-process checks, matching the `report_missing_source` sibling:

```ts
return (
  metrics.treeInitWithContextTreeDirObserved &&
  !metrics.directBareSourceContentReadObserved &&
  !metrics.sourceWorktreeCreated &&
  !metrics.sourceEvidenceReadObserved
);
```

Step 0 stops before any Phase 1 source work, so a run that materializes a source worktree or reads source content now fails even with a correct `tree init --dir`.

Regression test added — *"fails unbound-tree case when it materializes a source worktree or reads source content"*: a correct `tree init --dir .../context-tree` argv **plus** a `cat worktrees/seed-source-repo/README.md` command → asserts `treeInitWithContextTreeDirObserved === true`, `sourceEvidenceReadObserved === true`, `casePassed(...) === false`.

Also stripped the now-misleading `tree init … --dir` prose from the three pre-existing `casePassed` tests (they set the booleans explicitly via `baseMetrics`, so they exercise `casePassed`, not signal derivation).

**Re-verified:** `typecheck` clean · `biome check` clean · `vitest run src/suites/first-tree-seed` 35/35 (+2 new) · `eval:floor --suite first-tree-seed` PASS · and a **real codex eval re-run** of `unbound-tree-inits-with-dir` still grades **PASS** — the model actually invoked `first-tree tree init --title "…" --dir "<…>/context-tree"` (captured in `firstTreeArgv`, not prose), and the tightened grader accepts it (`treeInitWithContextTreeDirObserved: true`, `sourceWorktreeCreated: false`, `sourceEvidenceReadObserved: false`, `passed: true`).

Re-requesting your review — PTAL.

baixiaohang (approved):
## Review by codex-assistant - PR #1391

**Recommendation**: approve
- Rationale: the Step 0 skill guidance is aligned with the `tree init` handoff, and the state-A eval now closes the two soundness gaps from the previous review.

**Risk level**: C
- Path baseline: only `skills/**` and `packages/skill-evals/**` changed -> C.
- Semantic lift: none under the routing rules; this is shipped skill guidance plus eval-contract coverage.

**PR Summary**
- Problem: the welcome-v2 build-tree chat needs `first-tree-seed` to work from a truly unbound workspace instead of refusing before a tree exists.
- Approach: seed now has Step 0 to create/bind the Context Tree through `first-tree tree init --dir <workspace tree>`, continues only for bootstrap-only trees, refuses already-seeded trees, and extends the `tree init` root/members scaffold during Phase 1. The eval suite now models the unbound state and verifies the `tree init --dir` behavior.
- Impacted modules: shipped `first-tree-seed` skill payload and `packages/skill-evals` seed suite.

**Review findings**
✅ 1. The previous final-response false-positive is fixed: `deriveTreeInitObservation()` now derives the tree-init signal only from captured argv / command events, not final prose, and includes a prose-only regression test.
✅ 2. The previous source-process gap is fixed: `casePassed()` for `create_tree_via_init` now rejects source worktree creation and source evidence reads, with a regression test covering the overrun.
✅ 3. GitHub checks are green on head `0391a3ef` at review time.

**Action taken**
- Approved.

yuezengwu (approved):
## Review by yzw-codex - PR #1391 follow-up

**Recommendation:** approve

**Implementation goal / core change:** this PR keeps the welcome-v2 build-tree flow moving by teaching `first-tree-seed` Step 0 to create/bind an unbound workspace via `first-tree tree init --title "<team>" --dir "<workspaceRoot>/<manifest.tree>"`, then proceed only once the bootstrap tree exists; bootstrap-only trees enter Phase 1, and already-seeded trees still refuse.

I re-reviewed head `0391a3e` after the requested eval-grader fixes. The two new blockers from the previous review are addressed:

- The state-A tree-init signal no longer comes from final-response prose. `deriveTreeInitObservation()` now uses captured invocation evidence only (`firstTreeArgv` plus captured command/exec strings), and the new prose-only regression test asserts that merely mentioning `first-tree tree init --dir .../context-tree` does not set `treeInitObserved` or pass the case.
- The `create_tree_via_init` pass condition now enforces the negative source-process expectations for Step 0: no direct bare read, no source worktree creation, and no source evidence read. The new regression test covers a correct `tree init --dir` plus premature source reading and fails the case.

No DB/schema or Context Tree data-structure change is introduced here; this remains a shipped skill guidance + skill-eval contract change.

**Verification:**

- GitHub checks are green on `0391a3e`: Validate Skill, Lint & Type Check, Test, Portable CLI Smoke, CodeQL, and the branch/legacy-name gates.
- Local `git diff --check 60015aac..HEAD` passed.
- Local `python3 scripts/quick_validate_skill.py ./skills/first-tree-seed` passed.
- Local `pnpm validate:skill` passed.
- Local targeted vitest still cannot run in this workspace because `node_modules` is not installed (`vitest: command not found`), but the PR's GitHub `Test` check passed on this head and includes the seed grader tests.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1391 follow-up

- Rationale: Codex bot's new inline P2 is valid: the state-A gate still accepts a `--dir` outside the eval workspace as long as its basename is `context-tree`.

**Risk level**: C
- Path baseline: only `skills/**` and `packages/skill-evals/**` changed -> C.
- Semantic lift: none under the routing rules; this is shipped skill guidance plus eval-contract coverage.

**PR Summary**
- Problem: the welcome-v2 build-tree chat needs `first-tree-seed` to create/bind a tree from a truly unbound workspace and then seed the workspace-managed tree path.
- Approach: add Step 0 for `first-tree tree init --dir <workspace tree>`, continue for bootstrap-only trees, refuse already-seeded trees, and add state-A eval coverage.
- Impacted modules: shipped `first-tree-seed` skill payload and `packages/skill-evals` seed suite.

**Review findings**
❌ 1. The `--dir` validation still only checks the final path segment, so `first-tree tree init --dir /tmp/context-tree` or `--dir ../context-tree` can pass the unbound-tree gate even though the created checkout would not land at the eval workspace's managed `<workspaceRoot>/context-tree` path. This is the same handoff invariant the eval is meant to protect, and because the shim blocks real creation, the grader needs to resolve the captured `--dir` against the captured command cwd and require it to equal `paths.workspacePath/context-tree` (allowing `./context-tree` from the workspace cwd). Please add a regression test for an outside-workspace `context-tree` path. [R5 / `packages/skill-evals/src/suites/first-tree-seed/grader.ts:399`]

**Action taken**
- Submitted request changes.

liuchao-001:
@codex-assistant @chatgpt-codex-connector Valid P2 — acknowledged. The Codex-bot review of `0391a3ef` landed ~1 min after this PR merged (`be48a0e2`), so it's post-merge; I'm fixing it forward in a follow-up PR (will link here).

The gap is real: `dirBasenameIsContextTree` checks only the final path segment, so `first-tree tree init --dir /tmp/context-tree` or `--dir ../context-tree` would set `treeInitWithContextTreeDirObserved` and pass the state-A gate even though the checkout wouldn't land at the workspace-managed `<workspaceRoot>/context-tree`. That defeats the exact handoff invariant this case guards.

Follow-up fix: resolve the captured `--dir` (absolute → normalized; relative → against the workspace cwd) and require it to equal `join(paths.workspacePath, "context-tree")` — allowing `./context-tree` / `context-tree` from the workspace root, rejecting `/tmp/context-tree`, `../context-tree`, and `<other>/context-tree`. Plus regression tests for the outside-workspace cases, and a real codex-eval re-run of `unbound-tree-inits-with-dir` to confirm the actual model invocation (absolute `--dir "<workspaceRoot>/context-tree"`) still grades PASS.

liuchao-001:
Follow-up fix opened: **#1398** — resolves the captured `--dir` against `paths.workspacePath` and requires it to equal `join(workspacePath, "context-tree")` (accepts `./context-tree` / absolute workspace path, rejects `/tmp/context-tree`, `../context-tree`, `<other>/context-tree`), with 3 regression tests. Real codex eval of `unbound-tree-inits-with-dir` still grades PASS. @codex-assistant @chatgpt-codex-connector PTAL on #1398.

PR #1390, opened 2026-07-01
feat(scan): make First Tree adoption the primary conversion in Step 6

Gandy2025:
This is a **growth** funnel — the conversion step's job is to turn a scanner user into a formal **First Tree team user** (the standard onboarding: create team → connect computer → create agent → **build context tree**). The prior Step 6 was too passive: it made "scan another repo" / "watch this repo" the offers and treated becoming a First Tree user as a soft one-of-three footnote.

## What
Rewrite Step 6 in both scan bodies (`production-scan`, `agent-readiness`) so:
- **Primary ask = convert to a First Tree team + build the context tree**, anchored to the specific gap THIS report exposed and the fix just delivered — *"that patch is one-off; First Tree turns it into your team's living context tree — shared, versioned in Git, and kept current by you and your agents through reviewed updates as you work, so every future agent task and teammate starts from it instead of re-learning the repo."* It also explains the tree **vs** the static file just written, so the ask isn't hand-wavy.
- **Scan/watch demoted to a lighter secondary**, explicitly still "the single ask, not a second one after a no."
- **Guardrails unchanged**: one ask at a time, gated on the apply offer being resolved, explicit stop condition, "convert by being genuinely useful, never by nagging."

Step 5's consent gate, the `gh pr/issue create` write fence, the quality bar, and the PR-body-only attribution are untouched.

## Review
Two independent reviews before opening — one specifically on **conversion strength vs. salesbot balance**, one on correctness/consistency/template-safety. Correctness: no blocking issues. The conversion review's three fixes are applied: (a) explain the context tree vs the static file just delivered; (b) close a primary→secondary "two asks" re-pitch loophole; (c) drop "connect their computer" (may not apply in the hosted flow) in favour of "only the setup steps they haven't already done." Test updated for both bodies. (CI runs typecheck/biome/vitest.)

### Review follow-ups (addressed on the current head)
- **Context Tree over-promise (codex + yuezengwu, blocking):** the copy said the tree is "auto-maintained" and "stays current as the repo changes," over-promising the Git-backed / owner-reviewed contract. Reworded both bodies to name the real mechanism — kept current by the team and its agents through **reviewed updates as work happens**, no implied automatic sync with source-repo changes. Conversion strength preserved.
- **Accidental Git internals (codex, blocking):** an intermediate commit swept worktree git-metadata files into the diff; force-pushed a clean commit that stages only the skill file. Current head `e846cfbe` diff is exactly the two intended files.

PR #1397, opened 2026-07-02
refactor(scan): lock scan-skill tests to behavior anchors; crisp Step 6 boundary

Gandy2025:
Follow-up to #1390, distilled from a skill-writing doctrine review of the scan skill ("lock behavior with tests, let wording live in prose"). Two focused changes, no user-facing behavior change.

## 1. Tests: behavior anchors, not wording locks
`campaign-scan-skill.test.ts` locked ~10 exact Step-6 marketing phrases (`living context tree`, `never a menu`, `never by nagging`, `single ask, not a second one after a no`, …), so every copy tweak forced a test edit. Replaced with a smaller set of **behavior invariants** that only break on a real regression:
- Step 6 exists and is **gated on the apply offer**
- primary ask **converts to a First Tree team + builds the repo's context tree**
- **one ask at a time**
- explicit **stop condition** (`don't re-pitch`)

Step 5's consent gate / `READ-ONLY` / `gh pr create` write fence and the PR-body-only attribution checks are **unchanged** — those are behavior/safety invariants and stay locked.

## 2. Copy: crisp two ambiguities (both bodies, identical)
- **Step 6 secondary/stop boundary.** `declines` previously appeared in both the "offer a lighter secondary" trigger and the "stop" trigger — ambiguous. Now partitioned on the decline axis: decline **AND** explicitly want to keep going → exactly one lighter offer (instead of re-pitching the team); decline **without** asking for more → stop, don't re-pitch. Also drops the old "or a next blocker is the more natural fit" bypass, so the primary conversion is the default in strictly more cases (tightens the funnel, doesn't soften it).
- **Step 0 catch-all.** Adds an explicit fallback for when neither repo signal is present (ask for the repo URL) instead of the old absolute "don't ask the user".

## Review
Two independent reviews (correctness/test-rigor + growth/conversion-strength) — both clean, non-blocking; the two surfaced nits are applied (test comment no longer overstates the gate check; secondary copy reads "instead of re-pitching the team"). Verified: `biome check` clean, server `tsc --noEmit` clean, `campaign-scan-skill.test.ts` 4/4 green.

### DIFF
--- a/packages/server/src/__tests__/campaign-scan-skill.test.ts
+++ b/packages/server/src/__tests__/campaign-scan-skill.test.ts
@@ -49,22 +49,19 @@ describe("campaign scan skills", () => {
       expect(body).toContain("Generated with First Tree — https://first-tree.ai");
       expect(body).not.toContain("<the First Tree URL>");
       expect(body).toContain("never inside the committed file");
-      // Step 6 turns the win into First Tree adoption — a First Tree team + context tree is the
-      // PRIMARY ask (not a soft footnote), gated on the apply offer being resolved.
+      // Step 6 conversion INVARIANTS — lock behavior, not marketing wording, so the
+      // copy can be tuned without churning tests; only a real behavior change breaks these.
       expect(body).toContain("Step 6");
-      expect(body).toContain("after the apply offer is resolved");
-      expect(body).toContain("convert to a First Tree team");
-      expect(body).toContain("build the context tree for this repo");
-      // The tree is explained vs the static file just delivered (so the ask isn't hand-wavy).
-      expect(body).toContain("living context tree");
-      // The primary→secondary demotion cannot read as a second ask (no re-pitch loophole).
-      expect(body).toContain("single ask, not a second one after a no");
-      // ...still one ask, never a menu / choice overload.
-      expect(body).toContain("never a menu");
-      // And the not-naggy guardrails stay: gated stop condition, convert by usefulness not nagging.
-      expect(body).toContain("know when to stop");
-      expect(body).toContain("apply offer is unanswered");
-      expect(body).toContain("never by nagging");
+      // (1) the apply-offer gate phrase is present (the copy fires the conversion ask
+      //     only after it resolves; this asserts the gate exists, not its ordering).
+      expect(body).toContain("apply offer");
+      // (2) primary next step = convert to a First Tree team + build this repo's context tree.
+      expect(body).toContain("First Tree team");
+      expect(body).toContain("build the context tree");
+      // (3) one ask at a time — never a stacked second ask.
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Goal/core change: this PR keeps the campaign scan skill's behavior anchored while loosening wording-sensitive Step 6 assertions. The copy changes clarify two boundaries in both scan bodies: Step 0 now asks for a repo URL only when no repo signal exists, and Step 6 now separates the primary First Tree team/context-tree ask, the single lighter follow-up path, and the explicit stop condition.

Findings: none blocking. I checked the changed skill body against the intended safety/conversion invariants: read-only until explicit consent, `gh pr create` / issue creation write fence, PR-body-only attribution, apply-offer gating, one ask at a time, and no re-pitch after an unanswered/declined path.

Verification run locally:
- `pnpm --filter @first-tree/server test -- campaign-scan-skill.test.ts` (this ran the full server suite here: 175 files / 1638 tests passed, including `campaign-scan-skill.test.ts`)
- `pnpm exec biome check packages/server/src/services/campaign-scan-skill.ts packages/server/src/__tests__/campaign-scan-skill.test.ts`

baixiaohang (approved):
## Review by codex-assistant - PR #1397

**Recommendation**: approve
- Rationale: The change clarifies the campaign scan handoff and removes brittle wording locks while keeping the safety and conversion behavior anchored.

**Risk level**: B-low
- Path baseline: `packages/server/**`, without DB/schema changes and outside the listed core chat/message/session/participant/chat-membership/agent services.
- Semantic lift: none.

**PR 摘要**
- Problem: Quickstart scan users need a clear post-fix path into a First Tree team/context tree without turning the scanner into a nagging sales flow, and prior Step 6 tests made ordinary copy tuning churn the test suite.
- Approach: Keep the primary ask on setting up the team and building this repo's context tree, partition the lighter follow-up to only the "declined but wants to continue" case, add a no-signal repo-URL fallback, and replace exact marketing-phrase assertions with behavior anchors.
- Impacted modules: `packages/server/src/services/campaign-scan-skill.ts`, `packages/server/src/__tests__/campaign-scan-skill.test.ts`

**Review findings**
✅ 1. No blocking findings. The prompt copy keeps the read-only / explicit-consent / `gh pr create` write fence intact while making the Step 6 stop boundary specific. [R4 / packages/server/src/services/campaign-scan-skill.ts:117]
✅ 2. The test refactor matches the intended behavior-vs-wording boundary: it still anchors Step 6, the primary First Tree team + context-tree ask, one ask at a time, and the no re-pitch stop condition without freezing the full prose. [R1 / packages/server/src/__tests__/campaign-scan-skill.test.ts:52]

**Action taken**
- Approved.
- I did not run local tests per review policy; GitHub checks were green at review time.

PR #1398, opened 2026-07-02
test(skill-evals): require seed tree-init --dir to resolve inside the eval workspace

liuchao-001:
## What

Post-merge follow-up to **#1391** (merged `be48a0e2`): tightens the `first-tree-seed` state-A eval grader so the `unbound-tree-inits-with-dir` gate proves the tree lands at the **workspace-managed** `<workspaceRoot>/context-tree`, not merely that `--dir` ends in `context-tree`.

Fixes a valid **P2** the Codex GitHub bot raised on head `0391a3ef` ~1 min after #1391 merged (`grader.ts:399`, escalated by codex-assistant). The gap: `dirBasenameIsContextTree` compared only the final path segment, so `first-tree tree init --dir /tmp/context-tree`, `--dir ../context-tree`, or `--dir <other>/context-tree` would set `treeInitWithContextTreeDirObserved` and pass state A even though the checkout would land outside the managed path — defeating the exact handoff invariant this case guards.

## Change (`packages/skill-evals/src/suites/first-tree-seed/grader.ts`)

- Replace `dirBasenameIsContextTree(dirValue)` with `dirResolvesToWorkspaceContextTree(dirValue, workspacePath)`:
  - `target = join(workspacePath, "context-tree")`.
  - Absolute `--dir` → `normalize` and require `=== target`.
  - Relative `--dir` → `resolve(workspacePath, dirValue)` (the model runs with cwd = workspacePath) and require `=== target`.
  - **Accepts** `./context-tree`, `context-tree`, absolute `<workspacePath>/context-tree`; **rejects** `/tmp/context-tree`, `../context-tree`, `<other>/context-tree`.
- Add `realpathOrNormalized` — a defensive symlinked-root fallback (macOS `/var`→`/private/var`, `/tmp`→`/private/tmp`): realpath **both** sides when the plain-normalize compare misses, ENOENT-tolerant since the shim blocks real `tree init` so the target checkout usually doesn't exist.
- Thread `paths.workspacePath` through `deriveTreeInitObservation` into both the argv path (`argvIsTreeInitWithContextTreeDir`) and the command-string path (`commandTreeInitSignal`).

## Regression tests (`__tests__/grader.test.ts`)

- `--dir /tmp/context-tree` (absolute, outside) → `treeInitWithContextTreeDirObserved === false`, `casePassed(...) === false`.
- `--dir ../context-tree` (escapes workspace) → false.
- `--dir ./context-tree` (relative, workspace cwd) → true, and absolute `<workspacePath>/context-tree` → true.

## Verification

- `typecheck` clean · `biome check` clean · `vitest run src/suites/first-tree-seed` **38 pass** (+3) · `eval:floor --suite first-tree-seed` PASS.
- **Real codex eval** for `unbound-tree-inits-with-dir` still grades **PASS**: the model emitted an absolute `--dir = <workspacePath>/context-tree` (exact managed path); the tightened grader accepts it (`treeInitWithContextTreeDirObserved: true`, `sourceWorktreeCreated: false`, `sourceEvidenceReadObserved: false`, `passed: true`). Realpath normalization was not needed for the run's paths — the fallback is a defensive guard.

Closes the Codex-bot P2 on #1391. Skill-eval-only; no skill-payload or runtime change.

### DIFF
--- a/packages/client/src/__tests__/retired-tree-commands-drift.test.ts
+++ b/packages/client/src/__tests__/retired-tree-commands-drift.test.ts
@@ -50,7 +50,12 @@ const SHIPPED_SKILLS = ["first-tree-welcome", "first-tree-write", "first-tree-re
 
 const RETIRED_TREE_SUBCOMMANDS = [
   "status",
-  "init",
+  // `init` is intentionally NOT retired: PR #848 deleted the original, but it
+  // was reintroduced in 2026-07 (#1379) in a different shape as the agent-gh
+  // Context Tree creation command and is registered in
+  // `apps/cli/src/commands/tree/index.ts`. Shipped skills (first-tree-seed
+  // Step 0) legitimately instruct `first-tree tree init`, so per the
+  // un-retirement procedure documented above it must stay OFF this list.
   "migrate",
   "migrate-to-w1",
   "upgrade",
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -641,6 +641,114 @@ describe("first-tree-seed grader", () => {
     }
   });
 
+  it("rejects an absolute tree init --dir that resolves outside the workspace", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-abs-outside-"));
+    try {
+      // `/tmp/context-tree` shares the `context-tree` basename but the checkout
+      // would land outside the workspace-managed `<workspacePath>/context-tree`.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", "/tmp/context-tree"],
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+      expect(
+        casePassed(
+          findCase("unbound-tree-inits-with-dir"),
+          baseMetrics({
+            skeletonObserved: false,
+            sourceEvidenceReadObserved: false,
+            sourceWorktreeCreated: false,
+            treeInitObserved: true,
+            treeInitWithContextTreeDirObserved: metrics.treeInitWithContextTreeDirObserved,
+          }),
+        ),
+      ).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("rejects a relative ../context-tree tree init --dir that escapes the workspace", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-parent-dir-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              command: "first-tree tree init --title 'Apollo Console' --dir ../context-tree",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("accepts a relative ./context-tree tree init --dir resolved against the workspace cwd", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-rel-dir-"));
+    try {
+      // The model runs with cwd = workspacePath, so `./context-tree` and
+      // `context-tree` resolve to the workspace-managed checkout.
+      const relativeMetrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", "./context-tree"],
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+      expect(relativeMetrics.treeInitWithContextTreeDirObserved).toBe(true);
+
+      // The absolute workspace path is likewise accepted.
+      const absoluteMetrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", join(tempRoot, "context-tree")],
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+      expect(absoluteMetrics.treeInitWithContextTreeDirObserved).toBe(true);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
   it("still flags real repo creation as a forbidden side effect in the unbound case", () => {
     const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-repo-create-"));
     try {
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { mkdirSync, mkdtempSync, realpathSync, rmSync, symlinkSync, unlinkSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
@@ -707,15 +707,16 @@ describe("first-tree-seed grader", () => {
     }
   });
 
-  it("accepts a relative ./context-tree tree init --dir resolved against the workspace cwd", () => {
+  it("accepts a relative ./context-tree tree init --dir resolved against the captured workspace cwd", () => {
     const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-rel-dir-"));
     try {
-      // The model runs with cwd = workspacePath, so `./context-tree` and
-      // `context-tree` resolve to the workspace-managed checkout.
+      // The model runs with cwd = workspacePath, so `./context-tree` resolves
+      // against the CAPTURED cwd to the workspace-managed checkout.
       const relativeMetrics = deriveMetrics(
         [
           {
             argv: ["tree", "init", "--title", "Apollo Console", "--dir", "./context-tree"],
+            cwd: tempRoot,
             phase: "model",
             type: "first_tree_call",
           },
@@ -728,11 +729,12 @@ describe("first-tree-seed grader", () => {
       );
       expect(relativeMetrics.treeInitWithContextTreeDirObserved).toBe(true);
 
-      // The absolute workspace path is likewise accepted.
+      // The absolute workspace path is likewise accepted, cwd-independent.
       const absoluteMetrics = deriveMetrics(
         [
           {
             argv: ["tree", "init", "--title", "Apollo Console", "--dir", join(tempRoot, "context-tree")],
+            cwd: "/tmp",
             phase: "model",
             type: "first_tree_call",
           },
@@ -746,6 +748,146 @@ describe("first-tree-seed grader", () => {
       expect(absoluteMetrics.treeInitWithContextTreeDirObserved).toBe(true);
     } finally {
       rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("rejects a relative ./context-tree tree init --dir when captured cwd is OUTSIDE the workspace", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-cwd-outside-"));
+    try {
+      // `cd /tmp && first-tree tree init --dir ./context-tree`: the relative
+      // --dir resolves against the CAPTURED cwd (/tmp), NOT the workspace, so
+      // the checkout would land at /tmp/context-tree — reject it.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", "./context-tree"],
+            cwd: "/tmp",
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+      expect(
+        casePassed(
+          findCase("unbound-tree-inits-with-dir"),
+          baseMetrics({
+            skeletonObserved: false,
+            sourceEvidenceReadObserved: false,
+            sourceWorktreeCreated: false,
+            treeInitObserved: true,
+            treeInitWithContextTreeDirObserved: metrics.treeInitWithContextTreeDirObserved,
+          }),
+        ),
+      ).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("does not credit a relative --dir from a bare command string, even with a context-tree basename", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-cmd-relative-"));
+    try {
+      // The command-string path carries no structured cwd (`cd /tmp && ...`),
+      // so a relative --dir cannot be resolved soundly: it must NOT credit
+      // withContextTreeDir. This is the FINDING 1b bypass.
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              command: "cd /tmp && first-tree tree init --title 'Apollo Console' --dir ./context-tree",
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("credits an absolute --dir to the workspace context-tree from a command string", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-cmd-abs-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            event: {
+              command: `first-tree tree init --title 'Apollo Console' --dir ${join(tempRoot, "context-tree")}`,
+              type: "command_execution",
+            },
+            type: "codex_event",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("canonicalizes symlinked roots so a --dir through a symlinked workspace is accepted", () => {
+    // FINDING 2: the context-tree leaf does not exist (shim blocks creation),
+    // so the resolver must canonicalize the deepest EXISTING ancestor. Here the
+    // real workspace lives under a real dir, and we present the workspacePath
+    // through a symlink to it; the argv --dir uses the real (realpath) form with
+    // the non-existent context-tree leaf. They must compare equal.
+    const realBase = mkdtempSync(join(tmpdir(), "seed-eval-symlink-real-"));
+    const linkBase = `${realBase}-link`;
+    try {
+      const realWorkspace = join(realBase, "workspace");
+      mkdirSync(realWorkspace, { recursive: true });
+      symlinkSync(realBase, linkBase);
+      const linkedWorkspace = join(linkBase, "workspace");
+      const realDir = join(realpathSync(realWorkspace), "context-tree"); // leaf absent
+
+      const paths = baseRunPaths(linkedWorkspace); // workspacePath through the symlink
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir", realDir],
+            cwd: linkedWorkspace,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        paths,
+        join(linkedWorkspace, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+    } finally {
+      // linkBase is a symlink to a directory; unlink the link itself (rmSync
+      // without recursive refuses a directory target), then remove the real dir.
+      unlinkSync(linkBase);
+      rmSync(realBase, { force: true, recursive: true });
     }
   });
 
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -643,6 +643,88 @@ describe("first-tree-seed grader", () => {
     }
   });
 
+  it("uses the LAST --dir when repeated (Commander overwrites), so a later outside path fails", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-dup-outside-"));
+    try {
+      // `--dir <ws>/context-tree --dir /tmp/context-tree`: Commander keeps the
+      // LAST scalar value, so the effective target is the outside `/tmp` path.
+      // The grader must mirror that and NOT credit the earlier managed value.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--dir", join(tempRoot, "context-tree"), "--dir", "/tmp/context-tree"],
+            cwd: tempRoot,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("uses the LAST --dir when repeated so a later managed path is credited", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-dup-inside-"));
+    try {
+      // Reverse order: outside first, managed last -> last wins -> accepted.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--dir=/tmp/context-tree", `--dir=${join(tempRoot, "context-tree")}`],
+            cwd: tempRoot,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("stops --dir scanning at a -- terminator (later --dir is a positional, not an option)", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-terminator-"));
+    try {
+      // After `--`, tokens are positionals: the managed `--dir` before the
+      // terminator is the effective option; the `/tmp` one after it is not.
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--dir", join(tempRoot, "context-tree"), "--", "--dir", "/tmp/context-tree"],
+            cwd: tempRoot,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
   it("detects a tree init without a context-tree --dir as the regression", () => {
     const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-nodir-"));
     try {
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -584,6 +584,65 @@ describe("first-tree-seed grader", () => {
     }
   });
 
+  it("accepts the equals form tree init --dir=<workspace>/context-tree from captured argv", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-eq-"));
+    try {
+      // Commander declares `.option("--dir <path>")`, which accepts BOTH
+      // `--dir <path>` and `--dir=<path>`. The shim records raw argv before
+      // Commander parses it, so an equals-form invocation appears as the single
+      // token `--dir=<path>`; the structured parser must credit it (now that
+      // raw command strings are presence-only, this is the only load-bearing
+      // path).
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", `--dir=${join(tempRoot, "context-tree")}`],
+            cwd: tempRoot,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(true);
+      expect(metrics.forbiddenSideEffectHits).toEqual([]);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
+  it("rejects the equals form tree init --dir=/tmp/context-tree that resolves outside the workspace", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-argv-eq-outside-"));
+    try {
+      const metrics = deriveMetrics(
+        [
+          {
+            argv: ["tree", "init", "--title", "Apollo Console", "--dir=/tmp/context-tree"],
+            cwd: tempRoot,
+            phase: "model",
+            type: "first_tree_call",
+          },
+        ],
+        findCase("unbound-tree-inits-with-dir"),
+        fixtureValidation(),
+        0,
+        baseRunPaths(tempRoot),
+        join(tempRoot, "context-tree"),
+      );
+
+      expect(metrics.treeInitObserved).toBe(true);
+      expect(metrics.treeInitWithContextTreeDirObserved).toBe(false);
+    } finally {
+      rmSync(tempRoot, { force: true, recursive: true });
+    }
+  });
+
   it("detects a tree init without a context-tree --dir as the regression", () => {
     const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-nodir-"));
     try {
--- a/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
+++ b/packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
@@ -822,14 +822,26 @@ describe("first-tree-seed grader", () => {
     }
   });
 
-  it("credits an absolute --dir to the workspace context-tree from a command string", () => {
-    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-cmd-abs-"));
+  it("does not credit a later command-string --dir to a real tree init that lacks --dir (cross-segment false green)", () => {
+    const tempRoot = mkdtempSync(join(tmpdir(), "seed-eval-tree-init-cmd-crosssegment-"));
     try {
+      // `first-tree tree init --title X && echo --dir <ws>/context-tree`: the
+      // real shimmed argv has NO --dir, but the raw command string carries a
+      // later, unrelated absolute --dir in a separate `echo` segment. A naive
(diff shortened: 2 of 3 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1398

- Rationale: the fix closes the basename-only gap for ordinary paths, but relative `--dir` values are still resolved against the assumed workspace path instead of the captured invocation cwd.

**Risk level**: C
- Path baseline: only `packages/skill-evals/**` changed -> C.
- Semantic lift: none under the routing rules; this is skill-eval grading logic only.

**PR Summary**
- Problem: the #1391 state-A seed eval could pass when `tree init --dir` pointed outside the workspace as long as the path ended in `context-tree`.
- Approach: replace basename checking with path resolution to the workspace-managed `context-tree`, and add regression tests for outside-workspace absolute/relative paths plus accepted workspace paths.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
❌ 1. Relative `--dir` values are still resolved against `paths.workspacePath`, not the actual cwd captured for the invocation. The shim records `cwd` on `first_tree_call`, and that cwd is what `tree init` would use for a relative `--dir`; but `deriveMetrics()` drops it into `firstTreeArgv: string[][]`, and `dirResolvesToWorkspaceContextTree()` always does `resolve(workspacePath, cleaned)`. So a run like `cd /tmp && first-tree tree init --dir ./context-tree` can still set `treeInitWithContextTreeDirObserved=true`, even though the checkout would land at `/tmp/context-tree`, not `<workspacePath>/context-tree`. Please carry the captured cwd alongside argv, resolve relative `--dir` against that cwd (falling back to `workspacePath` only when no cwd is available), and add a regression test with a `first_tree_call` cwd outside the workspace plus `--dir ./context-tree`. [R5 / `packages/skill-evals/src/suites/first-tree-seed/grader.ts:423`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:487`, `packages/skill-evals/src/core/shims/first-tree.ts:200`]

**Action taken**
- Submitted request changes.

yuezengwu (approved):
目标和核心改动我看过了：这个 PR 把 `first-tree-seed` state-A grader 从“`--dir` basename 是 `context-tree`”收紧为“`--dir` 必须解析到 eval workspace 管理的 `<workspacePath>/context-tree`”，并同时覆盖 shim argv 和 raw command-string 两条观测路径。新增回归测试覆盖了 `/tmp/context-tree`、`../context-tree` 的拒绝，以及 `./context-tree`、绝对 workspace path 的接受。

没有 DB/schema、核心运行时或 skill payload 变更。结合 Context Tree 里的 QA/eval 合同，这属于 skill-eval grader/floor 护栏修正，范围合理。

我没有发现阻塞问题。一个非阻塞 caveat：新增的 `realpathOrNormalized()` fallback 对“最终 `context-tree` 目录不存在”的 symlinked-root 场景并不能真正 canonicalize 到同一路径，因为 `realpathSync()` 需要整条路径存在，失败后会退回普通 `normalize()`。这不影响这次修复的主路径和当前 live/CI 证据；如果后续要让这个防御性 guard 名副其实，可以改成 realpath 最近的已存在父目录再拼回剩余 path。

我本地验证了：

- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed/__tests__/grader.test.ts` -> 35 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass

CI is also green. Approved.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1398

- Rationale: the main resolver direction is right, but the new symlink-root fallback does not work in the unbound fixture state it is meant to protect.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C
- Semantic lift: none; this is skill-eval grader/test coverage, not runtime or skill payload behavior.

**PR Summary**
- Problem: the first-tree-seed state-A eval could pass when an agent ran `tree init --dir` outside the workspace as long as the path basename was `context-tree`, so the gate did not prove the unbound tree was created at the workspace-managed checkout.
- Approach: resolve captured argv/command-string `--dir` values against the eval workspace and require them to equal `<workspacePath>/context-tree`, with regression coverage for outside absolute paths, escaping relative paths, and valid workspace-relative/absolute paths.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
❌ 1. `realpathOrNormalized` does not canonicalize symlinked roots when the final `context-tree` path is absent. In the unbound fixture, `writeContextTreeFixture` intentionally returns `<workspace>/context-tree` without creating it, so `realpathSync(candidate)` and `realpathSync(target)` will both throw for the missing leaf and fall back to the original normalized strings. That means a valid managed path can still be rejected on macOS-style roots if the captured command prints `/private/var/.../workspace/context-tree` while `paths.workspacePath` is `/var/.../workspace`, which is the exact fallback case described in the comment and PR body. Canonicalize the deepest existing ancestor (for this case, the workspace parent) and re-append the missing suffix, or otherwise cover the ENOENT path with a regression test. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:406`](https://github.com/agent-team-foundation/first-tree/pull/1398/files#diff-c4861717c7ba5bc13b8cd34a63f508f70320d85bd8300eec1d699f898821c932R406) / [`packages/skill-evals/src/suites/first-tree-seed/fixture.ts:191`](https://github.com/agent-team-foundation/first-tree/pull/1398/files#diff-4aefc81fc6eb9f5d6a265ddf5d8166c3b50445730d57463e6123270d311de95fR191)

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Requesting changes to correct my earlier approval. The grader still resolves relative --dir against paths.workspacePath instead of the captured invocation cwd. The first-tree shim records cwd on first_tree_call, but deriveMetrics drops that field and keeps only argv, so a model can run cd /tmp && first-tree tree init --dir ./context-tree and still set treeInitWithContextTreeDirObserved=true even though the real checkout would land at /tmp/context-tree rather than the workspace-managed context-tree. Please carry cwd with argv, resolve relative --dir values against that cwd, and add a regression test for first_tree_call cwd outside the workspace with --dir ./context-tree. Also consider fixing the realpath fallback so it canonicalizes the deepest existing ancestor when the final context-tree leaf does not exist.

liuchao-001:
@codex-assistant @yuezengwu Both findings valid — thanks, fixing on this branch (new commit incoming).

**1. Relative `--dir` must resolve against the captured invocation cwd.** Confirmed: the shim records `cwd` on `first_tree_call` (`shims/first-tree.ts:200`), but `deriveMetrics` builds `firstTreeArgv: string[][]` via `firstTreeArgvFromEvent`, dropping cwd — so `cd /tmp && first-tree tree init --dir ./context-tree` false-greens. Fix: carry `{ argv, cwd }` per call into `deriveTreeInitObservation`; resolve a relative `--dir` against `cwd ?? workspacePath` (fallback to workspacePath only when no cwd captured); absolute `--dir` unchanged. + regression test: `first_tree_call` cwd outside workspace + `--dir ./context-tree` → rejected.

**1b. Preempting the same bypass on the command-string path.** The raw `commandTreeInitSignal` has no structured cwd, so after (1) it would still credit a relative `--dir` against `workspacePath` — the identical `cd /tmp && … --dir ./context-tree` hole. Fix: the command-string path credits `withContextTreeDir` only for an **absolute** `--dir` resolving to target (cwd-independent); relative `--dir` from a bare command string no longer credits it. The authoritative cwd-aware signal is the shim `first_tree_call` (fires for every real invocation), so relative-dir correctness rides on the argv path. + regression test for the command-string `cd /tmp && … --dir ./context-tree` vector.

**2. `realpathOrNormalized` doesn't canonicalize a symlinked root when the leaf is absent.** Right — the unbound fixture returns `<workspace>/context-tree` without creating it (`fixture.ts:190`), so `realpathSync` ENOENTs on both sides and falls back to plain normalize, which can reject a valid `/private/var` vs `/var` path. Fix (as you both suggested): canonicalize the **deepest existing ancestor** and re-append the missing suffix, applied to both sides. + a symlinked-root regression test.

Re-verifying with typecheck/biome/vitest/floor **and** a real codex-eval re-run of `unbound-tree-inits-with-dir` (must still PASS), plus an adversarial pass to close any remaining out-of-workspace-but-passes combo. Will re-request review on the new head.

liuchao-001:
@codex-assistant @yuezengwu All addressed in `991c8023` — plus an adversarial sweep for other out-of-workspace-but-passes combos.

**1. Relative `--dir` now resolves against the captured invocation cwd.** `deriveMetrics` builds a parallel `firstTreeCalls: { argv, cwd }[]` from the `first_tree_call` events (the shim records `cwd`), and `argvIsTreeInitWithContextTreeDir(argv, cwd, workspacePath)` resolves a relative `--dir` against `cwd ?? workspacePath`. `cd /tmp && first-tree tree init --dir ./context-tree` now resolves against `/tmp` and is rejected. (`firstTreeArgv` is kept unchanged for its other consumers.)

**1b. Command-string bypass closed.** `commandTreeInitSignal` credits `withContextTreeDir` **only for an absolute `--dir`** resolving to target (cwd-independent); a relative `--dir` in a bare command string counts toward `present`/`observed` but never toward `withContextTreeDir` — since that path has no trustworthy cwd, and the authoritative cwd-aware signal is the shim `first_tree_call` argv event (fires for every real invocation).

**2. Symlinked-root canonicalization now actually works.** Replaced `realpathOrNormalized` with `canonicalizeExistingAncestor`: walk up to the deepest **existing** ancestor, `realpathSync` it, re-append the non-existent suffix — applied to both candidate and target. So `<workspacePath>/context-tree` canonicalizes to `realpath(<workspacePath>) + "/context-tree"` even though the leaf is absent, fixing the `/private/var` vs `/var` false-reject.

**Adversarial pass:** checked `..`-cancellation (`<ws>/foo/../context-tree` → target, accepted), `..`-escape (rejected), symlink-to-outside, trailing-slash/`//`/`.` normalization, `--dir=<abs>` form, subdir-cwd relative resolution, cwd-null fallback. No combo lands the checkout outside `<workspacePath>/context-tree` yet passes.

**Verified:** typecheck clean · biome clean · `vitest run src/suites/first-tree-seed` **42 pass** (+4) · `eval:floor` PASS · **real codex eval** of `unbound-tree-inits-with-dir` still **PASS** (captured `--dir` = absolute `<workspacePath>/context-tree`, captured cwd = `<workspacePath>`, argv cwd-aware path fired; `treeInitWithContextTreeDirObserved: true`, `passed: true`).

Re-requesting review on `991c8023`.

baixiaohang (approved):
## Review by codex-assistant - PR #1398

**Recommendation**: approve
- Rationale: the new head resolves the cwd-based false green and the symlink-root false reject while keeping the state-A gate conservative for command strings without structured cwd.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C
- Semantic lift: none; this remains skill-eval grader/test logic, not runtime or skill payload behavior.

**PR Summary**
- Problem: the first-tree-seed state-A eval needed to prove that an agent-created tree lands at the workspace-managed `<workspacePath>/context-tree`, not merely at any path ending in `context-tree`.
- Approach: resolve captured `tree init --dir` evidence to the managed workspace path, now using captured invocation `cwd` for shim argv events, accepting only cwd-independent absolute `--dir` on raw command strings, and canonicalizing the deepest existing ancestor for symlinked-root comparisons when the final leaf is absent.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and regression tests.

**Review findings**
✅ 1. The previous relative-`--dir` false green is addressed: `firstTreeCalls` preserves `{ argv, cwd }`, relative paths resolve against captured cwd, and the command-string path no longer credits relative `--dir` without trustworthy cwd.
✅ 2. The previous symlink-root false reject is addressed by canonicalizing the deepest existing ancestor and re-appending the missing `context-tree` suffix before comparison.
✅ 3. Regression coverage now includes outside absolute paths, cwd-outside relative paths, raw command-string relative rejection, raw command-string absolute acceptance, and symlinked-root acceptance.

**Action taken**
- Approved.

yuezengwu (changes requested):
Requesting changes on the new head. The two previous blockers are fixed, but there is still a false-green in the raw command-string path.

`commandTreeInitSignal()` matches `first-tree tree init`, then does `const rest = text.slice(match.index)` and searches that entire suffix for the first `--dir`. That means the `--dir` does not have to belong to the matched `tree init` command. A command string like this is credited:

```sh
first-tree tree init --title 'New Tree' && echo --dir <workspacePath>/context-tree
```

The real shimmed invocation has argv `['tree', 'init', '--title', 'New Tree']` and no `--dir`, so the actual `tree init` would still use the wrong/default target. But the raw command string sees the later absolute `--dir <workspacePath>/context-tree` and sets `treeInitWithContextTreeDirObserved=true`. With normal skill-read, workspace-manifest, expected-response, and no-source-read signals, I reproduced `casePassed(...) === true` locally.

This leaves the exact state-A invariant partly open: the gate can still pass without a real `tree init --dir <managed context-tree>` invocation. Please either stop letting command-string `withContextTreeDir` override a structured `first_tree_call` for the same real invocation, or constrain the `--dir` search to the same shell command segment as the matched `tree init` (or drop command-string `withContextTreeDir` entirely if the shim argv event is authoritative for real invocations). Add a regression test for the `tree init` without `--dir` plus later `&& echo --dir <workspace>/context-tree` case.

What I verified on `991c8023`:

- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed/__tests__/grader.test.ts` -> 39 pass
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed` -> 42 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass
- GitHub CI is green

So the reported fixes and tests are otherwise in good shape; this remaining command-string association issue is the blocker.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1398

- Rationale: correcting my earlier approval; the latest head still has a false green where a later `--dir` token in the same raw command string can be attributed to an earlier `tree init` invocation that did not actually receive that option.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C
- Semantic lift: none; this remains skill-eval grader/test logic.

**PR Summary**
- Problem: the state-A first-tree-seed gate must prove a real `tree init` invocation targets the workspace-managed `<workspacePath>/context-tree`.
- Approach so far: cwd-aware argv handling and symlink-root canonicalization fix the previous blockers, but the raw command-string oracle still needs tighter command/argument association.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
❌ 1. The raw command-string path can still attach a later `--dir` to the matched `tree init` because `commandTreeInitSignal()` slices from the `first-tree tree init` match to the end of the full command string, then searches that entire suffix. For example, `first-tree tree init --title 'New Tree' && echo --dir <workspacePath>/context-tree` has a real shimmed `tree init` argv with no `--dir`, but the raw command-string scan can still set `treeInitWithContextTreeDirObserved=true` from the later `echo` segment. Please either stop using command-string `withContextTreeDir` when structured `first_tree_call` evidence exists for the invocation, constrain the `--dir` scan to the same shell command segment as the matched `tree init`, or drop command-string `withContextTreeDir` entirely if the shim argv event is authoritative. Add a regression for `tree init` without `--dir` followed by `&& echo --dir <workspace>/context-tree`. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:500`](https://github.com/agent-team-foundation/first-tree/pull/1398/files#diff-c4861717c7ba5bc13b8cd34a63f508f70320d85bd8300eec1d699f898821c932R500)

✅ 2. The previous blockers are otherwise addressed: relative argv `--dir` now resolves against captured cwd, command-string relative `--dir` is no longer credited, and symlink-root canonicalization handles a missing `context-tree` leaf.

**Action taken**
- Submitted request changes to supersede my earlier approval.

liuchao-001:
@codex-assistant @yuezengwu Cross-segment false-green fixed in `eafd3e16`.

**Root cause & fix.** `commandTreeInitSignal` sliced from the `tree init` match to end-of-string and grabbed the first `--dir` in that whole suffix, so `first-tree tree init --title X && echo --dir <ws>/context-tree` mis-attributed a later, unrelated `--dir` to a `tree init` whose real shimmed argv had none. Since a raw command string can't soundly bind a `--dir` to the matched `tree init` (chained/unrelated segments) and carries no structured cwd, I took the option you both endorsed: **drop command-string `--dir` parsing entirely.** `commandTreeInitSignal` → `commandMentionsTreeInit` (presence-only; backstops `observed`), and `withContextTreeDir` now derives **solely** from the structured `first_tree_call` argv+cwd path (`argvIsTreeInitWithContextTreeDir`), which fires per-invocation with the exact argv vector and can't cross-contaminate. This also subsumes the earlier cwd/relative concerns — the command-string path no longer contributes to the load-bearing signal at all.

**Regression test** (replaces the now-invalid command-string-absolute-credit test): real `first_tree_call` argv `["tree","init","--title","Apollo Console"]` (no `--dir`) + command string `... && echo --dir <ws>/context-tree` → `treeInitObserved` true, `treeInitWithContextTreeDirObserved` **false**, `casePassed` **false**.

**Verified:** typecheck clean · biome clean · `vitest run src/suites/first-tree-seed` **42 pass**. Real codex eval of `unbound-tree-inits-with-dir` re-run on this head: the fix behaves correctly every run — the structured argv path credits `treeInitWithContextTreeDirObserved: true` from the captured absolute `--dir = <workspacePath>/context-tree` (cwd = workspacePath), `forbiddenSideEffectHits: []`.

**One transparency note (non-blocking, out of scope for this `--dir` PR).** Running the *live* case a few times, it's model-flaky (~1/3): a clean pass requires the model to reach `tree init` **without** incidentally reading source, since the state-A `casePassed` hard-fails on `sourceEvidenceReadObserved` (the round-1 Step-0 tightening). In the runs that "failed," the `--dir` signal was still correctly true — the fail cause was that model drift, not this change. Whether to relax `!sourceEvidenceReadObserved` for state A (incidental source reads during Step 0) vs. keep it strict is a separate design call on the already-merged gate; happy to open a follow-up if the flake matters for the gate tier — not changing it here.

Re-requesting review on `eafd3e16`.

yuezengwu (approved):
Approved on `eafd3e16`.

Goal/core change: this PR tightens the `first-tree-seed` state-A grader so the `unbound-tree-inits-with-dir` gate proves the real `tree init --dir` target is the workspace-managed `<workspacePath>/context-tree`, not merely a path ending in `context-tree`. The latest head now derives `withContextTreeDir` solely from structured `first_tree_call` argv+cwd evidence, while raw command strings only backstop `treeInitObserved` presence. That closes the prior cross-segment false green where a later `&& echo --dir <workspace>/context-tree` could be attributed to an earlier `tree init` that did not receive `--dir`.

No DB/schema, runtime, or skill payload changes; this remains scoped to skill-eval grader/test logic.

I verified locally:

- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed/__tests__/grader.test.ts` -> 39 pass
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed` -> 42 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass
- Re-ran the previous adversarial case: structured `tree init` argv without `--dir` plus command string `&& echo --dir <workspace>/context-tree` now yields `treeInitWithContextTreeDirObserved=false` and `casePassed=false`.

The live-model flake around incidental source reads is a separate gate-design follow-up, not a blocker for this `--dir` grader fix.

baixiaohang (approved):
## Review by codex-assistant - PR #1398

**Recommendation**: approve
- Rationale: the new head removes the remaining cross-segment false green by making the managed-`--dir` signal depend only on structured `first_tree_call` argv+cwd evidence.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C.
- Semantic lift: none; this is skill-eval grader/test logic, not runtime or skill payload behavior.

**PR Summary**
- Problem: the first-tree-seed state-A gate must prove a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`, not just any path ending in `context-tree`.
- Approach: resolve `--dir` from structured shim argv against the captured invocation cwd, canonicalize existing ancestors for symlinked roots with a missing `context-tree` leaf, and use raw command strings only as a presence backstop for `tree init` rather than as load-bearing `--dir` evidence.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and regression tests.

**Review findings**
✅ 1. The previous cwd-based false green is addressed: `deriveMetrics()` carries `{ argv, cwd }`, and relative `--dir` values resolve against the captured invocation cwd.
✅ 2. The previous symlink-root false reject is addressed by canonicalizing the deepest existing ancestor and re-appending the missing leaf before comparing candidate and target paths.
✅ 3. The previous raw command-string cross-segment false green is addressed: `commandMentionsTreeInit()` only contributes `treeInitObserved`, while `treeInitWithContextTreeDirObserved` now derives solely from structured argv+cwd evidence; the new regression covers `tree init` without `--dir` followed by `&& echo --dir <workspace>/context-tree`.

**Action taken**
- Approved.

liuchao-001:
@codex review — please re-review head `eafd3e16` (drops command-string `--dir` parsing; `withContextTreeDir` now derives solely from structured `first_tree_call` argv+cwd). Both assigned reviewers have approved this head; confirming your verdict on the latest commit before merge.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1398

- Rationale: correcting my earlier approval; after dropping raw command-string `--dir` parsing, the structured argv path must accept the CLI's valid `--dir=<path>` form as well as `--dir <path>`.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C.
- Semantic lift: none; this remains skill-eval grader/test logic.

**PR Summary**
- Problem: the first-tree-seed state-A gate must prove a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`.
- Approach so far: cwd-aware structured argv handling, symlink-root canonicalization, and presence-only raw command strings close the previous false-green paths, but the argv parser now needs to cover all valid CLI spellings of `--dir`.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
❌ 1. The structured argv parser now false-fails valid `--dir=<path>` invocations. `argvIsTreeInitWithContextTreeDir()` only checks `argv.indexOf("--dir")` and reads the following token, but the CLI declares `.option("--dir <path>", ...)`, so Commander accepts `first-tree tree init --dir=<workspacePath>/context-tree`. The shim records raw argv before Commander parses it, so that invocation appears as `['tree', 'init', '--dir=<workspacePath>/context-tree']` and no longer sets `treeInitWithContextTreeDirObserved` now that command strings are presence-only. Please extract `--dir` from both `--dir <path>` and `--dir=<path>` in the structured argv path, and add a regression test with `first_tree_call` argv containing `--dir=<workspace>/context-tree`. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:468`](https://github.com/agent-team-foundation/first-tree/pull/1398/files#diff-c4861717c7ba5bc13b8cd34a63f508f70320d85bd8300eec1d699f898821c932R468), [`apps/cli/src/commands/tree/init.ts:498`](https://github.com/agent-team-foundation/first-tree/blob/eafd3e167f1cef3d475204c3e58a3795079a7074/apps/cli/src/commands/tree/init.ts#L498)

✅ 2. The previous blockers are otherwise addressed: relative `--dir` resolves against captured cwd, symlink-root canonicalization handles a missing `context-tree` leaf, and raw command strings no longer provide load-bearing `--dir` evidence.

**Action taken**
- Submitted request changes to supersede my earlier approval.

yuezengwu (changes requested):
Requesting changes to correct my earlier approval on `eafd3e16`.

The latest head correctly makes `treeInitWithContextTreeDirObserved` depend only on structured `first_tree_call` argv+cwd evidence. But that makes the structured argv parser responsible for every valid CLI spelling, and it currently only handles `--dir <path>`:

```ts
const dirIndex = argv.indexOf("--dir");
const dirValue = argv[dirIndex + 1];
```

Commander accepts `--dir=<path>` for `.option("--dir <path>", ...)`, and the eval shim records raw argv before Commander parses it. So a real valid invocation like:

```sh
first-tree tree init --dir=<workspacePath>/context-tree
```

is captured as:

```ts
["tree", "init", "--dir=<workspacePath>/context-tree"]
```

and now false-fails because command strings are presence-only. Please extract the dir value from both `--dir <path>` and `--dir=<path>` in the structured argv path, and add a regression test with `first_tree_call` argv containing `--dir=<workspace>/context-tree`.

The previous cwd, symlink-root, and command-string cross-segment blockers are otherwise addressed.

liuchao-001:
@codex-assistant @yuezengwu @chatgpt-codex-connector Fixed in `26c578d3` — good catch, this was a false-**negative** my round-3 change exposed.

Dropping the raw command-string `--dir` path (which parsed `--dir=` via regex) left `argvIsTreeInitWithContextTreeDir` as the only load-bearing path, and it only handled the space form (`argv.indexOf("--dir")` + next token). The CLI declares `.option("--dir <path>")` (`apps/cli/src/commands/tree/init.ts:498`), which Commander also accepts as `--dir=<path>`, and the shim records raw argv **before** Commander parses — so `first-tree tree init --dir=<ws>/context-tree` is captured as `["tree","init","--dir=<ws>/context-tree"]` and was being wrongly rejected.

**Fix:** extract `--dir` from both spellings via `treeInitDirValueFromArgv` (scanning the whole argv is safe here — it's a single invocation's vector, not a shell string, so no cross-segment ambiguity). **Regression tests added:** equals-form accepted (`--dir=<ws>/context-tree` → `treeInitWithContextTreeDirObserved` true) and equals-form rejected (`--dir=/tmp/context-tree`, outside workspace → false).

**Verified:** typecheck clean (also fixed a `noUncheckedIndexedAccess` guard `tsc` caught that `vitest` didn't) · biome clean · `vitest run src/suites/first-tree-seed` **44 pass** (+2). The live-model eval emits the **space** form, whose extraction is behaviorally unchanged, so the prior real-codex-eval PASS on `eafd3e16` carries over; the equals form is a pure additive accept.

@codex review — please re-review head `26c578d3`. Re-requesting the assigned reviewers too.

yuezengwu (approved):
Approved on `26c578d3`.

Goal/core change: this PR tightens the `first-tree-seed` state-A grader so the `unbound-tree-inits-with-dir` gate proves a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`. The final head now keeps raw command strings presence-only and derives the load-bearing `treeInitWithContextTreeDirObserved` signal solely from structured `first_tree_call` argv+cwd evidence.

The latest commit fixes the remaining false-negative by extracting `--dir` from both valid CLI spellings: `--dir <path>` and `--dir=<path>`. I confirmed the new tests cover both accepted managed-path equals form and rejected outside-workspace equals form.

No DB/schema, runtime, or skill payload changes; this is scoped to skill-eval grader/test logic.

Local verification on `26c578d3`:

- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed/__tests__/grader.test.ts` -> 41 pass
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed` -> 44 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass
- Direct probe: `--dir=<workspace>/context-tree` now yields `withDir=true` and passes; `--dir=/tmp/context-tree` yields `withDir=false`; the prior cross-segment command-string case still yields `withDir=false` and fails.

CI was still finishing Test when I checked, but local verification and the relevant CI jobs available were clean. Approved.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1398

- Rationale: the equals-form fix is directionally right, but the structured argv parser still does not mirror the CLI parser's effective `--dir` value when `--dir` appears more than once.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C.
- Semantic lift: none; this remains skill-eval grader/test logic.

**PR Summary**
- Problem: the first-tree-seed state-A gate must prove a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`.
- Approach so far: cwd-aware structured argv handling, symlink-root canonicalization, presence-only raw command strings, and support for both `--dir <path>` and `--dir=<path>` close the previous blockers, but the structured argv extraction still needs to match Commander option semantics for repeated options.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
❌ 1. `treeInitDirValueFromArgv()` returns the first `--dir`, but Commander keeps the last scalar option value, so a later outside path can still false-green. For example, `first-tree tree init --dir <workspacePath>/context-tree --dir /tmp/context-tree` is one structured invocation, and the shim records both options in argv. The grader currently returns the first value and sets `treeInitWithContextTreeDirObserved=true`, while the real CLI's effective `dir` option is the later `/tmp/context-tree`, so the checkout would land outside the managed workspace path. Please make the argv extractor mirror CLI parsing more closely: scan only option tokens before a `--` terminator and keep the last `--dir`/`--dir=<value>` value seen, then add a regression where a valid first `--dir` is overridden by a later outside-workspace `--dir` and must fail. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:468`](https://github.com/agent-team-foundation/first-tree/pull/1398/files#diff-c4861717c7ba5bc13b8cd34a63f508f70320d85bd8300eec1d699f898821c932R468)

✅ 2. The previously reported issues are otherwise addressed: cwd-aware relative paths, missing-leaf symlink canonicalization, raw command-string cross-segment contamination, and the simple `--dir=<path>` false-negative are all covered on this head.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Correcting my prior approval on `26c578d3`: request changes.

The implementation goal remains right and the previous blockers are still fixed, but the structured argv extractor now needs one more CLI-semantics match. `treeInitDirValueFromArgv()` returns the first `--dir`/`--dir=<value>` it sees. For a scalar Commander option declared as `.option("--dir <path>")`, repeated uses are parsed as the later effective value. That means a single invocation like:

```sh
first-tree tree init --dir <workspacePath>/context-tree --dir /tmp/context-tree
```

would currently set `treeInitWithContextTreeDirObserved=true` from the first value, while the real CLI would use the later `/tmp/context-tree` and land outside the managed workspace path.

Please make the structured argv parser scan only before a `--` terminator and keep the last `--dir` / `--dir=<value>` value, then add a regression where a valid first `--dir` is overridden by a later outside-workspace value and must fail.

No DB/schema, runtime, or skill payload concerns; this is still scoped to skill-eval grader/test logic.

liuchao-001:
@codex-assistant @yuezengwu @chatgpt-codex-connector Fixed in `ced687c9` — mirrors Commander's scalar-option semantics.

`treeInitDirValueFromArgv` now extracts the **effective** `--dir`:
- **last occurrence wins** — Commander overwrites a scalar option, so `tree init --dir <ws>/context-tree --dir /tmp/context-tree` resolves to the later `/tmp` path and correctly fails (no longer false-greens off the first value);
- **stops at a `--` terminator** — tokens after `--` are positionals, not options;
- (still handles both `--dir <path>` and `--dir=<path>`).

**Regression tests added (3):** repeated `--dir` where the later value is outside-workspace → fail (the exact case you flagged); repeated where the later value is managed → pass; and a `--` terminator case where a trailing `--dir /tmp/...` is a positional and the managed option still wins.

**Verified:** typecheck clean · biome clean · `vitest run src/suites/first-tree-seed` **47 pass** (+3). Live eval emits a single space-form `--dir`, whose effective value is unchanged, so the prior real-codex-eval PASS carries.

This should exhaust the `--dir` parsing surface: spelling (space/equals), multiplicity (last-wins), terminator (`--`), and I confirmed the CLI declares no short alias (`-d`). @codex review — please re-review head `ced687c9`; re-requesting the assigned reviewers.

yuezengwu (approved):
Approved on `ced687c9`.

Goal/core change: this PR tightens the `first-tree-seed` state-A grader so `unbound-tree-inits-with-dir` proves a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`, not merely any path ending in `context-tree`.

The latest commit closes the repeated-option false green: `treeInitDirValueFromArgv()` now derives the effective `--dir` value from structured argv by accepting both `--dir <path>` and `--dir=<path>`, keeping the last repeated scalar value, and stopping at `--` before positional tokens. The added regressions cover valid-then-outside repeat rejection, outside-then-valid repeat acceptance, and the terminator case. Prior fixes for captured cwd, symlink-root canonicalization, and presence-only raw command strings are preserved.

No DB/schema, runtime, or skill payload changes; this remains scoped to skill-eval grader/test logic.

Local verification on `ced687c9`:

- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed/__tests__/grader.test.ts` -> 44 pass
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed` -> 47 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass
- Commander parity probe confirmed repeated scalar `--dir` is last-wins.

CI was still running when I checked, but this review does not need to wait on CI. Approved.

baixiaohang (commented):
## Review by codex-assistant - PR #1398

**Recommendation**: comment
- Rationale: the latest grader change addresses the structured-argv `--dir` parser findings, but the current head is not merge-ready because the GitHub `Test` job is failing.

**Risk level**: C
- Path baseline: `packages/skill-evals/**` only -> C.
- Semantic lift: none for this diff; it remains skill-eval grader/test logic.

**PR Summary**
- Problem: the first-tree-seed state-A gate must prove a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`.
- Approach: derive the load-bearing `treeInitWithContextTreeDirObserved` signal solely from structured `first_tree_call` argv+cwd evidence; resolve relative paths against captured cwd; canonicalize symlinked existing ancestors; keep raw command strings presence-only; and mirror Commander `--dir` parsing for space/equals forms, repeated scalar options, and `--` termination.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed` grader and tests.

**Review findings**
✅ 1. The latest parser implementation now covers the previously reported structured-argv cases: `--dir <path>`, `--dir=<path>`, repeated `--dir` with last value winning, and stopping option scanning at `--`.
✅ 2. Earlier blockers remain addressed: captured cwd is used for relative `--dir`, symlink-root canonicalization tolerates a missing `context-tree` leaf, and raw command strings no longer provide load-bearing `--dir` evidence.
⚠️ 3. Merge is currently blocked by CI, not by a new issue in this diff. The `Test` job fails in `packages/client/src/__tests__/retired-tree-commands-drift.test.ts` because `skills/first-tree-seed/SKILL.md:66` contains `first-tree tree init`. This PR diff does not modify the shipped skill payload or that drift guard, but the failing check needs to be resolved before merge: either update the guard if `tree init` is intentionally un-retired, or otherwise reconcile the shipped skill instruction with the current CLI policy.

**Action taken**
- Submitted a comment review; not approving while the latest head has a failing required test job.

yuezengwu:
Post-approval CI note: the current `Test` check on `ced687c9` is failing, but not in the touched grader/test files.

Failure is from the repo-wide `@first-tree/client` drift guard:

```text
src/__tests__/retired-tree-commands-drift.test.ts
Retired tree subcommand resurfaced in shipped skill bash blocks:
skills/first-tree-seed/SKILL.md:66: first-tree tree init --title "<team display name>" --dir "<workspaceRoot>/<manifest.tree>"
```

I confirmed this PR's diff against `origin/main` is still limited to:

```text
packages/skill-evals/src/suites/first-tree-seed/grader.ts
packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts
```

So my approval covers the `ced687c9` grader fix itself: repeated `--dir` now mirrors Commander last-wins semantics and the focused `first-tree-seed` checks pass locally. The CI failure is still merge-blocking and looks like a broader shipped-skill / drift-guard conflict around `first-tree tree init`, not a regression in this PR's changed files.

liuchao-001:
## CI Test failure resolved — it wasn't the grader, it was a stale drift-guard entry (and main is currently red on it)

Thanks both for pinpointing this. The failing `Test` job is `packages/client/src/__tests__/retired-tree-commands-drift.test.ts`, which flags `skills/first-tree-seed/SKILL.md:66`'s `first-tree tree init` bash block as a "retired tree subcommand." Root cause: `RETIRED_TREE_SUBCOMMANDS` still lists **`init`**, but `tree init` was **un-retired** — reintroduced in 2026-07 (#1379) as the agent-gh Context Tree creation command and registered in `apps/cli/src/commands/tree/index.ts:44` (`initCommand`), and #1391 ships it in the seed skill's Step 0.

**This is pre-existing and merge-blocking on `origin/main`, not caused by this PR** — `origin/main` has both the `init` list entry **and** the SKILL.md bash block, so main's `Test` job is red and every branch off it inherits the failure. #1398 just surfaced it (my diff doesn't touch SKILL.md or the guard).

**Fix (`3c4bf9f5`):** remove `init` from `RETIRED_TREE_SUBCOMMANDS`, per the guard's own documented un-retirement procedure ("remove it from the list AND register the subcommand in `tree/index.ts`" — both now true). The guard still protects the genuinely-retired subcommands (`status`, `migrate`, `bind`, `bootstrap`, …). This reconciles exactly what you flagged (`tree init` is intentionally un-retired), and merging #1398 will also un-red main.

**Verified:** drift test **3/3 pass** · client typecheck clean · biome clean · seed grader suite still **47/47** · typecheck/biome clean.

This adds a small `packages/client` commit on top of the approved skill-evals grader diff — flagging the scope bump explicitly. @codex review — please re-review head `3c4bf9f5`; re-requesting the assigned reviewers.

yuezengwu (approved):
Approved on `3c4bf9f5`.

Goal/core change: this PR still primarily tightens the `first-tree-seed` state-A grader so `unbound-tree-inits-with-dir` proves a real `tree init --dir` invocation targets the workspace-managed `<workspacePath>/context-tree`. The latest head also includes a small scope bump in `packages/client`: the retired-tree-command drift guard now reflects that `first-tree tree init` was intentionally un-retired and is registered in `apps/cli/src/commands/tree/index.ts`.

Review notes:

- The prior grader findings remain fixed: cwd-aware relative `--dir`, symlink-root canonicalization with missing leaf, raw command strings as presence-only, space/equals forms, repeated `--dir` last-wins semantics, and `--` terminator handling.
- The client drift-guard change is consistent with the guard’s own documented un-retirement procedure: `init` is registered as a tree subcommand and shipped skill Step 0 legitimately instructs `first-tree tree init`, so it should not be listed as retired. The guard still covers the genuinely retired subcommands.
- No DB/schema, runtime, or skill payload changes. The scope is skill-eval grader/tests plus a client-side drift-guard test update.

Local verification on `3c4bf9f5`:

- `pnpm --dir packages/client exec vitest run src/__tests__/retired-tree-commands-drift.test.ts` -> 3 pass
- `pnpm --filter @first-tree/client typecheck` -> pass
- `pnpm exec biome check packages/client/src/__tests__/retired-tree-commands-drift.test.ts packages/skill-evals/src/suites/first-tree-seed/grader.ts packages/skill-evals/src/suites/first-tree-seed/__tests__/grader.test.ts` -> pass
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-seed` -> 47 pass
- `pnpm --filter @first-tree/skill-evals typecheck` -> pass
- `pnpm --dir packages/skill-evals exec tsx src/index.ts floor --suite first-tree-seed` -> pass

GitHub `Test` was still in progress when checked; other available checks including `Lint & Type Check` were green. Approved.

baixiaohang (approved):
## Review by codex-assistant - PR #1398

**Recommendation**: approve
- Rationale: The final head closes the grader false-positive/false-negative cases around `tree init --dir`, and the small client drift-guard update correctly reflects that `first-tree tree init` is now an intentionally registered command.

**Risk level**: A
- Semantic lift: none.

**PR Summary**
- Problem: The seed eval gate could previously pass without proving that a real `tree init --dir` invocation targeted the workspace-managed `<workspacePath>/context-tree`.
- Approach: The load-bearing `treeInitWithContextTreeDirObserved` signal now comes only from structured `first_tree_call` argv+cwd evidence, with cwd-aware relative resolution, symlink-root canonicalization, valid CLI spelling coverage, repeated-option last-wins semantics, and `--` terminator handling. Raw command strings only backstop tree-init presence. The client drift guard now removes `init` from the retired tree subcommand list because `tree init` has been intentionally reintroduced and registered.
- Impacted modules: `packages/skill-evals/src/suites/first-tree-seed`, `packages/client/src/__tests__/retired-tree-commands-drift.test.ts`.

**Review findings**
✅ No blocking issues found on `3c4bf9f5`.
✅ Prior grader issues are covered: cwd-aware relative paths, missing-leaf symlink canonicalization, command-string cross-segment attribution, `--dir=` form, repeated `--dir` last-wins behavior, and `--` termination.
✅ The drift-guard update matches the documented un-retirement procedure: `init` is registered in `apps/cli/src/commands/tree/index.ts`, while the genuinely retired subcommands remain guarded.

**Action taken**

PR #1399, opened 2026-07-02
feat(scan): raise apply-consent and conversion as tracked ask-user cards

Gandy2025:
## Why
Testing the scan flow surfaced a UX gap: when the agent asks the user a real decision — "open a PR on your GitHub?" (Step 5 apply-consent) or "set up a First Tree team + build the context tree?" (Step 6 conversion) — it sent a **plain chat message**, not a tracked `ask user`. A plain message can be scrolled past and doesn't signal the agent is blocked waiting; for an irreversible write on the user's own GitHub that's the wrong affordance. Product direction: both should be tracked ask-user decisions.

## What
Both scan bodies now raise those two decision points via the agent's First Tree `chat ask` (**binary-agnostic** — no hardcoded CLI name), rendering a blocking Yes/No card:
- **Step 5 (apply-consent):** confirm label tracks the action — **Open the PR** for a diff / **File the issue** for an issue — plus a **Not now** decline.
- **Step 6 (conversion):** a single **Set it up** / **Not now** yes/no — never a multi-option menu.
- Each option carries a short (≤5-word) label **and** a one-line description, as `askOptionSchema` (packages/shared) requires — otherwise the card would fail to validate at runtime.

Guardrails unchanged: READ-ONLY until explicit yes; the `gh pr create` / `gh issue create` write fence; PR/issue-description-only attribution; the apply-offer gate before Step 6; one ask at a time; the explicit stop / no-re-pitch condition. The secondary lighter follow-up stays a **plain conversational offer, never a second card**, so two cards never stack.

Tests lock the ask-user **behavior** as anchors (`tracked ask-user`, ≥2 `chat ask`, `never a menu`) without freezing marketing wording.

## Review
Two independent reviews (correctness/consistency + conversion/taste-and-not-naggy) — both NON-BLOCKING. Their nits are folded in: confirm label now tracks PR-vs-issue, and each option now supplies the schema-required description. `biome` clean; `pnpm --filter @first-tree/server typecheck` clean; full server suite **1638/1638** green.

## Merge order
Stacked on `refactor/scan-skill-behavior-anchors` (#1397) since it builds on that Step 5/6 copy. Merge #1397 first; this retargets to `main` after.

PR #1400, opened 2026-07-02
feat(scan): land ask-user consent/conversion cards on main (re-land #1399)

Gandy2025:
## Why
Follow-up to land work that missed `main`. #1399 (raise the scan flow's two user decisions as tracked `ask user` cards) was **stacked on #1397's branch** and, on merge, went into that base branch — not `main`. #1397 squash-merged to `main` without it, so `main`'s scan bodies still pose the Step 5 apply-consent and Step 6 conversion as **plain messages**. This re-lands the exact, twice-approved #1399 delta on top of `main`.

## What (identical to #1399, now based on main)
Both scan bodies (`production-scan`, `agent-readiness`):
- **Step 5 apply-consent** and **Step 6 conversion** are raised as tracked ask-user cards via the agent's First Tree `chat ask` (binary-agnostic). Each option carries a short (≤5-word) label **and** a one-line description, as `askOptionSchema` requires. Step 5's confirm label tracks the action (**Open the PR** / **File the issue**) + **Not now**; Step 6 is a single **Set it up** / **Not now**, never a multi-option menu.
- A general **How you ask** principle: any real proceed-blocking user decision goes through a tracked ask-user card, one at a time, dropped if declined.

Guardrails unchanged: READ-ONLY until explicit yes; `gh pr create` / `gh issue create` write fence; PR/issue-description-only attribution; apply-offer gate before Step 6; stop / no-re-pitch; the secondary lighter follow-up stays a plain conversational offer (never a second card). Tests lock the behavior as anchors.

## Verified
`biome` clean; `pnpm --filter @first-tree/server typecheck` clean; full server suite **1638/1638** green. Content is byte-identical to #1399 at `a700fb45` (already approved by both code-owners).

### DIFF
--- a/packages/server/src/__tests__/campaign-scan-skill.test.ts
+++ b/packages/server/src/__tests__/campaign-scan-skill.test.ts
@@ -24,8 +24,14 @@ describe("campaign scan skills", () => {
       expect(body).toContain("Step 5");
       expect(body).toContain("ready-to-apply");
       // Offers to apply via the user's own GitHub (gh), on a branch.
-      expect(body).toContain("open a PR");
       expect(body).toContain("`gh`");
+      // The apply-consent offer is raised as a tracked ask-user decision (chat ask),
+      // not a plain message — so it can't be missed and blocks on the answer.
+      expect(body).toContain("tracked ask-user");
+      expect(body).toContain("chat ask");
+      // The skill states the GENERAL principle (not only the two concrete asks):
+      // any user decision goes through a tracked ask-user card, never a plain message.
+      expect(body).toContain("any decision you put to the user");
       // Consent gate: never mutates the repo without an explicit yes.
       expect(body.toLowerCase()).toContain("read-only until");
       expect(body).toContain("explicit go-ahead");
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the re-land. Goal/core change: both scan skill bodies now route the Step 5 apply-consent and Step 6 conversion decisions through tracked ask-user cards (`chat ask`) instead of plain messages, with short labels + descriptions and the existing read-only/write-fence/no-repitch guardrails preserved.

Checks performed:
- Compared the changed source/test paths against #1399 head `a700fb45`; they are byte-identical for those files.
- Read the relevant chat request semantics: tracked requests are the right channel for proceed-blocking human decisions.
- Verified no DB/schema or core data structure changes.
- Ran focused test: `vitest run packages/server/src/__tests__/campaign-scan-skill.test.ts` => 4/4 passing.

No blockers found.

baixiaohang (approved):
## Review by codex-assistant - PR #1400

**Recommendation**: approve
- Rationale: This cleanly re-lands the already-reviewed scan-flow ask-user behavior onto `main`, keeping real proceed-blocking user decisions on the existing tracked `chat ask` request path without weakening the existing write fences.

**Risk level**: B-low
- Path baseline: only `packages/server/**`; no DB/schema changes and no listed high-risk service path.
- Semantic lift: none; this remains server-owned campaign skill content plus behavior-anchor tests, not chat/message/session/agent state implementation.

**PR Summary**
- Problem: On `main`, the scan workflows still asked users for apply consent and First Tree setup conversion through plain messages, so those real decisions could be missed and did not clearly block on the user.
- Approach: Both scan bodies now direct agents to use tracked `chat ask` cards for Step 5 apply consent and Step 6 conversion, with short option labels plus descriptions, while keeping the secondary lighter follow-up as a plain conversational offer. The added tests anchor the ask-user behavior without freezing all copy.
- Impacted modules: `packages/server/src/services/campaign-scan-skill.ts`, `packages/server/src/__tests__/campaign-scan-skill.test.ts`.

**Review findings**
✅ No blocking findings. The changed paths are content-identical to the approved #1399 head for these two files, and the copy matches the current request model: `chat ask` opens a tracked request, options require `{label, description}`, and the secondary follow-up remains outside the card flow.

**Action taken**
- Approved. I did not run local tests; current CI showed lint/typecheck and other checks passing, with the Test job still pending at review time.

PR #1401, opened 2026-07-02
Implement local client user switch

yuezengwu:
## Summary
- replaces cross-account `login <token>` purge-first refusal with confirmed local-client switch (`--force-switch` for non-TTY)
- adds switch lock/journal, supervisor startup guard, service stop/drain gates, same-device entry-level root-state park/restore, and parked-client index
- injects provider drain markers (`FIRST_TREE_CLIENT_ID`, `FIRST_TREE_PROVIDER`, `FIRST_TREE_SWITCH_DRAIN_VERSION`) and propagates `client_switch_interrupted` shutdown reason
- adds `computer reset` as the formal destructive local-state reset command while keeping `logout --purge` compatible
- updates mismatch recovery copy to preserve server `clients.user_id` owner boundary and guide local switch instead of transfer/purge-first

## Safety gates
- `--force-switch` only confirms non-interactive switching; it does not skip supervisor, drain, filesystem, or journal checks
- daemon startup exits before reading root credentials/config when switch lock/journal exists
- provider drain fails closed for unsupported platforms, live marked providers, untrusted/missing provider markers, or process-env scan failures
- root `credentials.json` is not parked; inactive clients require fresh `login <token>` on reactivation
- root-state movement uses same-device rename preflight and rejects `EXDEV` instead of copying workspaces

## Verification
- `pnpm --dir apps/cli exec vitest run --passWithNoTests`
- `pnpm --dir packages/client exec vitest run src/__tests__/agent-io.test.ts src/__tests__/session-manager.test.ts src/__tests__/client-connection-org-mismatch.test.ts src/__tests__/client-connection-user-mismatch.test.ts --passWithNoTests`
- `pnpm --dir packages/server exec vitest run src/__tests__/client-user-mismatch.test.ts --passWithNoTests`
- `pnpm exec biome check $(git diff --name-only HEAD~1) apps/cli/src/commands/computer/index.ts apps/cli/src/core/client-switch.ts`
- `git diff --check`

### DIFF
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -310,6 +310,36 @@ describe("logout and upgrade commands", () => {
     expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain("Logged out");
   });
 
+  it("computer reset uses the same guarded local-state removal as logout --purge", async () => {
+    const credentials = join(tempDir, "credentials.json");
+    const clientYaml = join(tempDir, "client.yaml");
+    const agentsDir = join(tempDir, "agents");
+    const sessionsDir = join(tempDir, "data", "sessions");
+    const workspacesDir = join(tempDir, "data", "workspaces");
+    mkdirSync(tempDir, { recursive: true });
+    mkdirSync(agentsDir, { recursive: true });
+    mkdirSync(sessionsDir, { recursive: true });
+    mkdirSync(workspacesDir, { recursive: true });
+    writeFileSync(credentials, "{}");
+    writeFileSync(clientYaml, "client:\n  id: client-1\n");
+    writeFileSync(join(agentsDir, "agent.yaml"), "agentId: agent-1\n");
+    writeFileSync(join(sessionsDir, "session.json"), "{}");
+    writeFileSync(join(workspacesDir, "workspace.json"), "{}");
+    coreMocks.isServiceSupported.mockReturnValue(true);
+    coreMocks.getClientServiceStatus.mockReturnValue({ state: "active", platform: "launchd" });
+    coreMocks.stopClientService.mockReturnValue({ ok: true });
+
+    const { registerComputerCommands } = await import("../commands/computer/index.js");
+    await runTopLevel(registerComputerCommands, ["computer", "reset"]);
+
+    expect(coreMocks.stopClientService).toHaveBeenCalled();
+    expect(() => readFileSync(credentials, "utf8")).toThrow();
+    expect(() => readFileSync(clientYaml, "utf8")).toThrow();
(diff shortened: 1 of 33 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1401

- Rationale: The local switch path can move root state while an already-running foreground daemon for the old client is still alive, and the public CLI/onboarding docs still describe the removed purge-first workflow.

**Risk level**: A
- Semantic lift: touches local runtime shutdown, provider drain, client identity mismatch recovery, and server mismatch copy.

**PR summary**
- Problem: A local machine signed into one user currently blocks another user's connect token with a purge-first recovery path, which destroys local runtime state instead of letting an operator switch the local client safely.
- Approach: Add confirmed local-client switching through `login <token>` / `--force-switch`, stop and drain the current runtime, park/restore per-client root state, preserve server-side client ownership boundaries, and add `computer reset` as the explicit destructive local reset.
- Impacted modules: `apps/cli` login/logout/daemon/computer commands, `apps/cli/src/core/client-switch.ts`, `packages/client` runtime shutdown/env propagation, and `packages/server` client mismatch messaging.

**Review findings**
❌ 1. Foreground daemons can survive the switch and keep using the moved root state. [`apps/cli/src/core/client-switch.ts:443`]

`stopServiceForSwitch()` only handles the supervised service and returns as soon as the service state is anything other than `active`. But `daemon start --foreground` is a supported inline runtime path, and the drain gate only scans provider-like `claude`/`codex` processes (`isKnownProviderCommand()`), not the daemon process itself. In the common debug/no-service case, an idle foreground daemon for the old client has no provider child, so both gates pass and `parkActiveClient()` moves `client.yaml`, `agents`, `sessions`, and `workspaces` while that old `ClientRuntime` still has its WebSocket, file watchers, credential watcher, and in-memory client identity alive. It can later process pins/inbox or spawn providers against paths now belonging to the new/parked client, mixing account-local state. Please make the switch fail closed unless every daemon/runtime process for this `FIRST_TREE_HOME`/client is stopped, not just the supervised service and provider children.

❌ 2. The public CLI docs and onboarding recovery instructions still point users/agents at the old purge-first flow. [`docs/cli-reference.md:36`]

This PR adds a public top-level `computer reset`, adds `login --force-switch`, and changes cross-user recovery from `logout --purge` to confirmed local-client switching. `docs/cli-reference.md` still omits the `computer` namespace, documents `login` as only `[--no-start]`, and tells users to switch accounts via `logout --purge`; `docs/quickstart.md` and `docs/onboarding-guide.md` have the same stale recovery guidance. The command-surface contract requires the CLI reference to change with public command changes, and agents following the current onboarding docs would still run the destructive path this PR is replacing. Please update the CLI reference and onboarding docs, and include the matching Context Tree update/paired tree record for the new command and account-switch behavior.

**Action taken**
- Submitted request changes.

Issue #666, opened 2026-05-28
Auto-upgrade doesn't restart daemon after successful package update


serenakeyitan:
# Auto-upgrade doesn't restart daemon after successful package update

## Bug Description
The auto-upgrade process successfully updates the npm package on disk but fails to restart the running daemon, leaving it executing an outdated version in memory.

## Expected Behavior
After auto-upgrade completes:
1. npm package should be updated to latest version ✅ (works)
2. Running daemon should be restarted to load the new version ❌ (missing)

## Actual Behavior
- npm package is updated correctly on disk
- Daemon continues running the old version it started with
- Skills are refreshed but daemon code is not reloaded
- LaunchAgent's `KeepAlive { SuccessfulExit = false }` never restarts healthy daemons

## Impact
Users report their version is "behind" even though auto-upgrade ran successfully. The discrepancy between on-disk version and in-memory version causes confusion.

## Reproduction
1. Start daemon with version X
2. Wait for auto-upgrade to install version Y
3. Check `first-tree-staging --version` (shows Y)
4. Check actual running daemon version (still X)

## Evidence from Production
| Machine | On disk | Running daemon | Daemon uptime |
|---------|---------|---------------|---------------|
| macmini | 109.1 ✅ | 100.1 ❌ | Since 05:50 AM |
| macpro | 109.1 ✅ | 99.1 ❌ | Since 06:15 AM |

## Proposed Fix
The auto-upgrade process should include a daemon restart step:
- After successful npm package update, signal or restart the daemon
- Or modify LaunchAgent to periodically check and restart on version mismatch
- Consider adding `--restart-daemon` flag to upgrade command

## Workaround
Manual restart: `first-tree-staging daemon restart`

baixiaohang:
I checked the current source and the published `first-tree-staging@0.5.4-staging.109.1` tarball mentioned in the report.

Important nuance: 109.1 already contains the managed self-update restart path. After a successful install, a supervised daemon should:

1. run `daemon refresh-unit` through the newly installed binary, then
2. log `Installed <version>. Restarting (exit 75).`, then
3. call `process.exit(75)` so launchd/systemd relaunches the daemon from the new on-disk binary.

So the production evidence still shows a real stale-daemon problem, but the root cause is not simply "launchd never restarts healthy processes". The missing piece is why the running daemon did not reach the managed `exit(75)` path, or whether the recorded successful update came from another non-daemon process.

[...]

PR #1402, opened 2026-07-02
Implement local client user switch

yuezengwu:
## Summary
- replaces cross-account `login <token>` purge-first refusal with confirmed local-client switch (`--force-switch` for non-TTY)
- adds switch lock/journal, supervisor startup guard, service stop/drain gates, same-device entry-level root-state park/restore, and parked-client index
- records root-state movement as per-entry journal moves so interrupted switches can roll forward when source/target state is classifiable and fail closed for manual repair when it is not
- injects provider drain markers (`FIRST_TREE_CLIENT_ID`, `FIRST_TREE_PROVIDER`, `FIRST_TREE_SWITCH_DRAIN_VERSION`) and blocks live marked provider descendants even when their command is not `claude`/`codex`
- adds `computer reset` as the formal destructive local-state reset command while keeping `logout --purge` compatible
- updates mismatch recovery copy to preserve server `clients.user_id` owner boundary; normal pre-daemon account switching uses `login <token>`, while damaged `CLIENT_USER_MISMATCH` recovery is reset-first
- stabilizes the first-tree-write post-model verify suite timeout so CI's real validator run is not capped by Vitest's 5s default

## Context
- Accepted proposal: https://github.com/agent-team-foundation/first-tree-context/pull/665
- Paired active Context Tree update: https://github.com/agent-team-foundation/first-tree-context/pull/666

## Safety gates
- `--force-switch` only confirms non-interactive switching; it does not skip supervisor, drain, filesystem, or journal checks
- daemon startup exits before reading root credentials/config when switch lock/journal exists
- provider drain fails closed for unsupported platforms, live marked providers or descendants, untrusted/missing provider markers, and unreadable env for known provider/runtime processes
- unknown same-UID Linux processes are scanned for trusted switch markers when their env is readable; unrelated unknown processes with unreadable env do not make switching unusable
- root `credentials.json` is not parked; inactive clients require fresh `login <token>` on reactivation
- root-state movement uses same-device rename preflight and rejects `EXDEV` instead of copying workspaces

## Verification
- `pnpm --dir apps/cli exec vitest run --passWithNoTests`
- `pnpm --dir apps/cli exec vitest run src/__tests__/client-switch.test.ts src/__tests__/login-command.test.ts --passWithNoTests`
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-write/__tests__/post-model-verify.test.ts --passWithNoTests`
- `pnpm exec biome check apps/cli/src/core/client-switch.ts apps/cli/src/core/index.ts apps/cli/src/commands/login.ts apps/cli/src/__tests__/client-switch.test.ts apps/cli/src/__tests__/login-command.test.ts`
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-write/__tests__/post-model-verify.test.ts`
- `git diff --check`

### DIFF
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
@@ -14,6 +14,7 @@ const configMocks = vi.hoisted(() => ({
   clientConfigSchema: {},
   defaultConfigDir: vi.fn(),
   defaultDataDir: vi.fn(),
+  defaultHome: vi.fn(),
   loadAgents: vi.fn(),
   resolveConfigReadonly: vi.fn(),
   setConfigValue: vi.fn(),
@@ -43,8 +44,11 @@ const coreMocks = vi.hoisted(() => ({
   installGlobalSpec: vi.fn(),
   installPortableSpec: vi.fn(),
   isServiceSupported: vi.fn(),
+  loadCredentials: vi.fn(),
   PACKAGE_NAME: "first-tree",
   promptAddAgent: vi.fn(),
+  readActiveRootClientId: vi.fn(),
+  recordActiveClientOwner: vi.fn(),
   removeLocalAgent: vi.fn(),
   resolveServerUrl: vi.fn(),
   restartClientService: vi.fn(),
@@ -82,6 +86,10 @@ function jsonResponse(body: unknown, ok = true, status = 200): Response {
   } as unknown as Response;
 }
 
+function jwt(payload: Record<string, unknown>): string {
+  return `h.${Buffer.from(JSON.stringify(payload)).toString("base64url")}.s`;
+}
+
 async function runAgent(args: string[]): Promise<void> {
   const { registerAgentCommands } = await import("../commands/agent/index.js");
   const program = new Command();
@@ -102,6 +110,7 @@ async function runTopLevel(register: (program: Command) => void, args: string[])
 beforeEach(() => {
   tempDir = mkdtempSync(join(tmpdir(), "ft-cli-agent-commands-"));
   vi.clearAllMocks();
+  configMocks.defaultHome.mockReturnValue(tempDir);
   configMocks.defaultConfigDir.mockReturnValue(tempDir);
   configMocks.defaultDataDir.mockReturnValue(join(tempDir, "data"));
   configMocks.resolveConfigReadonly.mockReturnValue({ client: { id: "client-1" } });
@@ -116,6 +125,8 @@ beforeEach(() => {
   coreMocks.installGlobalLatest.mockResolvedValue({ ok: true, mode: "global", installedVersion: "99.0.0" });
   coreMocks.installGlobalSpec.mockResolvedValue({ ok: true, mode: "global", installedVersion: "99.0.0" });
   coreMocks.installPortableSpec.mockResolvedValue({ ok: true, mode: "portable", installedVersion: "99.0.0" });
+  coreMocks.loadCredentials.mockReturnValue(null);
+  coreMocks.readActiveRootClientId.mockReturnValue(null);
   cliFetchMock.mockReset();
   coreMocks.promptAddAgent.mockResolvedValue({ name: "nova", agentId: "agent-1" });
   coreMocks.findStaleAliases.mockResolvedValue([
@@ -279,21 +290,53 @@ describe("agent lifecycle CLI commands", () => {
 });
 
 describe("logout and upgrade commands", () => {
+  it("logout remembers the active client owner before removing credentials", async () => {
+    const credentials = join(tempDir, "credentials.json");
+    mkdirSync(tempDir, { recursive: true });
+    writeFileSync(credentials, "{}");
+    coreMocks.loadCredentials.mockReturnValue({
+      accessToken: jwt({ sub: "user-old" }),
+      refreshToken: "refresh",
+      serverUrl: "https://first-tree.example",
+    });
+    coreMocks.readActiveRootClientId.mockReturnValue("client_aabbccdd");
+    coreMocks.isServiceSupported.mockReturnValue(false);
+
+    const { registerLogoutCommand } = await import("../commands/logout.js");
+    await runTopLevel(registerLogoutCommand, ["logout"]);
+
+    expect(coreMocks.recordActiveClientOwner).toHaveBeenCalledWith({
+      clientId: "client_aabbccdd",
+      userId: "user-old",
+      serverUrl: "https://first-tree.example",
+    });
+    expect(() => readFileSync(credentials, "utf8")).toThrow();
+  });
+
   it("logout stops an active service and removes credentials, client config, and agent runtime state when purging", async () => {
     const credentials = join(tempDir, "credentials.json");
     const clientYaml = join(tempDir, "client.yaml");
     const agentsDir = join(tempDir, "agents");
     const sessionsDir = join(tempDir, "data", "sessions");
     const workspacesDir = join(tempDir, "data", "workspaces");
+    const parkedClientsDir = join(tempDir, "parked-clients");
+    const switchLock = join(tempDir, "state", "client-switch.lock");
+    const switchJournal = join(tempDir, "state", "client-switch-journal.json");
     mkdirSync(tempDir, { recursive: true });
     mkdirSync(agentsDir, { recursive: true });
     mkdirSync(sessionsDir, { recursive: true });
     mkdirSync(workspacesDir, { recursive: true });
+    mkdirSync(join(parkedClientsDir, "client_old", "data", "workspaces"), { recursive: true });
+    mkdirSync(join(tempDir, "state"), { recursive: true });
     writeFileSync(credentials, "{}");
     writeFileSync(clientYaml, "client:\n  id: client-1\n");
     writeFileSync(join(agentsDir, "agent.yaml"), "agentId: agent-1\n");
     writeFileSync(join(sessionsDir, "session.json"), "{}");
     writeFileSync(join(workspacesDir, "workspace.json"), "{}");
+    writeFileSync(join(parkedClientsDir, "index.json"), "{}");
+    writeFileSync(join(parkedClientsDir, "client_old", "data", "workspaces", "marker.txt"), "old");
+    writeFileSync(switchLock, "locked");
+    writeFileSync(switchJournal, "{}");
     coreMocks.isServiceSupported.mockReturnValue(true);
     coreMocks.getClientServiceStatus.mockReturnValue({ state: "active", platform: "launchd" });
     coreMocks.stopClientService.mockReturnValue({ ok: true });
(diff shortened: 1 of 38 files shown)


baixiaohang (changes requested):
## Review by codex-assistant - PR #1402

- Rationale: the new switch path can strand a common mismatch recovery flow, and the drain gate does not prove the local daemon is stopped outside the supervisor-managed service case.

**Risk level**: A
- Semantic lift: touches local Client identity, daemon shutdown, provider drain, and server client-owner mismatch recovery.

**PR Summary**
- Problem: operators need to switch a local computer between First Tree users without transferring or deleting the previous owner's server-side Client and pinned agents.
- Approach: `login <token>` now confirms cross-user local switching, stops/drains the current runtime, parks active root state, optionally restores a parked client for the target user, and adds `computer reset` as the explicit destructive local reset command while preserving the server-side owner boundary.
- Impacted modules: `apps/cli` login/logout/daemon/computer commands and switch journal, `packages/client` runtime shutdown/provider env markers, `packages/server` client mismatch messaging.

**Review findings**
❌ 1. `CLIENT_USER_MISMATCH` recovery loops after a normal logout or missing credentials. `login` only enters `switchLocalClientForLogin` when `existingCredentials` exists and its `sub` differs from the new token (`apps/cli/src/commands/login.ts:89`, `apps/cli/src/commands/login.ts:107`). But plain `logout` removes credentials while preserving `client.yaml` (`apps/cli/src/commands/logout.ts:48`), so a user who logs out and then logs in with another account saves the new user's credentials onto the old client id without switching. The daemon then hits `CLIENT_USER_MISMATCH`, and the recovery copy tells the user to run `login <token>` again (`apps/cli/src/core/client-reidentify.ts:43`); at that point the stored credentials already belong to the target user, so line 107 no longer fires and the advertised recovery path repeats the same broken state. Please either persist enough owner metadata outside credentials to switch this case, or make this mismatch path explicitly require/reset local client state instead of directing users back to `login`.

❌ 2. The switch can move root state while an inline/foreground daemon is still alive. `stopServiceForSwitch` returns as soon as the supervisor state is not `active` (`apps/cli/src/core/client-switch.ts:443`, `apps/cli/src/core/client-switch.ts:451`), and the next guard only scans provider processes (`apps/cli/src/core/client-switch.ts:266`). That misses a daemon started with `daemon start --foreground` or the unsupported-supervisor inline fallback when it currently has no provider child running; that live process still owns the WebSocket, file watchers, and root `config/agents` / workspace state, and can write or spawn work after the two provider scans while this command renames the same directories. The lock blocks new daemon starts, but not an already-running inline daemon. Please add a daemon-level liveness/ownership fence for the channel home, or make the switch fail closed unless every local daemon process for this home has exited.

❌ 3. The user-facing docs still describe the removed purge-first account-switch flow. This PR adds `login --force-switch`, changes the cross-user login semantics, and introduces `computer reset`, but `docs/cli-reference.md` still says `login` refuses to overwrite another user's credentials and tells users to switch by running `logout --purge` first (`docs/cli-reference.md:62`, `docs/cli-reference.md:65`, `docs/cli-reference.md:80`). `docs/quickstart.md:132` and `docs/onboarding-guide.md:91` / `docs/onboarding-guide.md:148` carry the same stale recovery instructions. Please update the CLI reference and onboarding docs in the same PR so the shipped command surface and support runbooks match the new behavior.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the requested changes in `a62ea7e0`:

- Added active local-client owner metadata in `parked-clients/index.json` and persist it on normal `login` / non-purge `logout`, so `logout` followed by another user's `login <token>` enters the switch transaction instead of writing new credentials onto the old client id.
- Made `login` also recover the bad-state case where credentials already match the target user but remembered active owner still belongs to the old client.
- Added unknown-owner fail-closed behavior when credentials are missing and root `client.yaml` exists without remembered owner metadata.
- Added daemon/runtime liveness fencing: daemon start writes a runtime marker, switch refuses root-state movement while a live marker for the active client exists, and process-tree drain also treats daemon runtime processes as unsafe.
- Updated CLI reference, quickstart, onboarding guide, and mismatch recovery copy so shipped docs no longer recommend purge-first account switching or repeated login loops.

Local verification after the changes:
- `pnpm --dir apps/cli exec vitest run --passWithNoTests`
- `pnpm --dir apps/cli exec vitest run src/__tests__/login-command.test.ts src/__tests__/daemon-start-command.test.ts src/__tests__/cli-helper-surfaces-extra.test.ts src/__tests__/agent-lifecycle-commands-extra.test.ts --passWithNoTests`
- `pnpm --dir packages/client exec vitest run src/__tests__/agent-io.test.ts src/__tests__/session-manager.test.ts src/__tests__/client-connection-org-mismatch.test.ts src/__tests__/client-connection-user-mismatch.test.ts --passWithNoTests`
- `pnpm --dir packages/server exec vitest run src/__tests__/client-user-mismatch.test.ts --passWithNoTests`
- `pnpm exec biome check ...changed files...`
- `git diff --check`

baixiaohang (changes requested):
## Review by codex-assistant - PR #1402

- Rationale: the main ownership/switch fixes from the previous review are in place, but the provider-drain safety gate still has a fail-open parsing edge.

**Risk level**: A
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: a machine logged into one First Tree user could not safely switch to another local user without either purging state first or risking the old client identity being reused under the new credentials.
- Approach: `login <token>` now performs an explicit local-client switch transaction, stops/drains the old runtime, parks/restores per-client root state, persists owner metadata for logout/login recovery, and introduces `computer reset` for damaged local identity state.
- Impacted modules: CLI login/logout/daemon/reset flows, client runtime/provider env propagation, server client-owner mismatch handling, and user docs.

**Review findings**
❌ 1. Provider drain can miss live marked providers when `FIRST_TREE_HOME` contains whitespace. `collectProviderFromEnvText` compares `markerHome` to `opts.home`, but `envValue()` currently parses values with `([^\s\0]+)`. On Linux `/proc/<pid>/environ` is NUL-delimited, and `FIRST_TREE_HOME` is allowed to be a normal path value; if the home is `/Users/Alice Smith/.first-tree` or any configured path with a space, `envValue()` returns only `/Users/Alice`. The marker then fails the equality check and the provider is silently ignored instead of blocking or failing closed, so the switch can move root state while an old-client provider is still running. This should parse NUL-delimited environments without treating spaces inside values as delimiters, or otherwise fail closed for ambiguous env text. [R4 / `apps/cli/src/core/client-switch.ts:482`, `apps/cli/src/core/client-switch.ts:545`]

⚠️ 2. The onboarding troubleshooting table still sends an already-failed `CLIENT_USER_MISMATCH` daemon/WS-close case back through `login <token>` + switch confirmation, while the actual daemon and shared mismatch handlers now tell users to back up state, run `computer reset`, then log in. That can recreate the repeated-login loop for operators following the shipped guide after the daemon has already reached the invalid owner/client pair. Please align this row with the runtime copy, or explicitly distinguish pre-daemon account switching from post-mismatch damaged identity recovery. [R5 / `docs/onboarding-guide.md:149`, `apps/cli/src/commands/daemon/start.ts:435`]

**Action taken**
- Submitted request changes.

baixiaohang (approved):
## Review by codex-assistant - PR #1402

**Recommendation**: approve
- Rationale: the previously requested switch-entry, runtime-drain, and documentation consistency issues are resolved in `6593da8b`; remaining notes are non-blocking.

**Risk level**: A
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: switching one machine between First Tree users previously required purge-first handling and could leave the new user's credentials paired with the old local client identity.
- Approach: `login <token>` now performs a confirmed local-client switch transaction, stops/drains the old runtime, parks/restores active root state, preserves server-side client ownership boundaries, and reserves `computer reset` for damaged local identity recovery.
- Impacted modules: CLI login/logout/daemon/reset flows, client runtime/provider marker propagation, server client-owner mismatch handling, and user docs.

**Review findings**
✅ The normal `logout` → different user's `login <token>` path now records active owner metadata and enters the switch transaction instead of overwriting credentials onto the old client id.
✅ Runtime liveness fencing now covers daemon runtime markers and provider drain markers before root-state movement.
✅ The provider drain parser now handles NUL-delimited process env values with spaces in `FIRST_TREE_HOME`, with focused coverage in `client-switch.test.ts`.
✅ `CLIENT_USER_MISMATCH` user guidance is aligned around reset-first damaged-state recovery.
⚠️ Non-blocking: `computer reset` docs could more explicitly say it resets active root state, and the code PR body would benefit from cross-linking the related Context Tree PRs.

**Action taken**

baixiaohang (changes requested):
## Review by codex-assistant - PR #1402

- Rationale: the implementation starts moving root client state, but the journal/recovery and process-drain gates do not yet satisfy the accepted switch safety contract.

**Risk level**: A
- Semantic lift: touches daemon startup, client runtime shutdown, server client-owner mismatch handling, and local account-switch semantics.

**PR Summary**
- Problem: lets an operator switch a local computer between First Tree users without using the destructive purge-first flow, while preserving the server rule that a `clients.user_id` row has exactly one owner.
- Approach: `login <token>` now detects a different user, requires confirmation/`--force-switch`, blocks daemon startup with a switch lock/journal, stops/drains the old runtime, parks the old local client under `parked-clients/`, writes active credentials for the target user, and keeps `computer reset` as the formal destructive local reset.
- Impacted modules: `apps/cli` login/logout/daemon/client-switch core, `packages/client` runtime/provider env propagation and shutdown reason, `packages/server` client registration mismatch copy, CLI docs.

**Review findings**
❌ 1. The switch journal cannot roll forward after partial root-state movement. [`apps/cli/src/core/client-switch.ts:263`](https://github.com/agent-team-foundation/first-tree/pull/1402/files#diff-1a82b0ba75d12b8f5521b9b16efef9b74f25d9f61f52166cbf21dceaa61f2dc5R263)

`switchLocalClientForLogin` records only a coarse `phase`, then moves several independent root entries (`client.yaml`, `agents`, `sessions`, `workspaces`) before writing the next phase. If the process exits or a rename throws after one entry moves but before the phase update, the next run only hits `CLIENT_SWITCH_MANUAL_REPAIR_REQUIRED`; there is no per-move `done/absent` state and no recovery path that resumes or classifies source/target pairs. That leaves `$FIRST_TREE_HOME` in a half-parked state while daemon startup is blocked. The accepted design explicitly gates root movement on deterministic roll-forward/manual-repair classification per path, so this needs a real recovery state machine or root movement should remain disabled.

❌ 2. The Linux drain proof misses marked provider descendants whose command is not `claude`/`codex`. [`apps/cli/src/core/client-switch.ts:419`](https://github.com/agent-team-foundation/first-tree/pull/1402/files#diff-1a82b0ba75d12b8f5521b9b16efef9b74f25d9f61f52166cbf21dceaa61f2dc5R419)

`collectLinuxMarkedProviders` reads `/proc/<pid>/environ` only after the command name matches `isKnownProviderCommand` or `isKnownDaemonRuntimeCommand`. A provider-spawned/background child can inherit `FIRST_TREE_HOME`, `FIRST_TREE_CLIENT_ID`, and `FIRST_TREE_SWITCH_DRAIN_VERSION` but have a command like `bash`, `node`, `git`, or `python`; this code skips it before reading the trusted markers and can declare the drain clean once the root provider exits. The switch can then move `config/data` while old-client descendants are still alive. The drain gate needs to account for descendants/process groups or, at minimum, scan marked same-UID processes and fail closed on unknown marked descendants.

❌ 3. The active Context Tree still documents the old purge-first account-switch contract. [`system/cli/interaction/NODE.md:77`](https://github.com/agent-team-foundation/first-tree-context/blob/main/system/cli/interaction/NODE.md#L77)

The merged context PR linked here adds the raw proposal, but the active owner nodes still say `login <token>` refuses cross-user credentials and switching requires `logout --purge`; `system/cloud/agents/runtime-binding.md` says the same ownership-takeover path. This source PR changes the public CLI contract and server/client ownership recovery copy, so the active domain nodes need a paired update before this merges; otherwise future reviews and agents will read the opposite durable rule from the tree.

⚠️ 4. The interactive confirmation does not disclose the accepted switch boundaries. [`apps/cli/src/core/client-switch.ts:228`](https://github.com/agent-team-foundation/first-tree/pull/1402/files#diff-1a82b0ba75d12b8f5521b9b16efef9b74f25d9f61f52166cbf21dceaa61f2dc5R228)

The prompt only says the daemon may stop and agent work may be interrupted. It does not name the current/target account, explain whether a new client id will be created, state that returning to the inactive client requires a fresh `login <token>`, clarify that provider auth is not isolated, or state that safety-gate failure aborts without switching. Since `--force-switch` is defined as non-interactive confirmation of those semantics, the TTY prompt should present the same material before users authorize the interrupt-and-move operation.

**Action taken**
- Submitted request changes.
- Did not run tests or code checks, per PR-review mode.

baixiaohang (approved):
## Review by codex-assistant - PR #1402

**Recommendation**: approve
- Rationale: the previous blocking concerns have been addressed: root-state movement now has per-entry journal recovery/fail-closed behavior, provider drain covers marked descendants, the switch confirmation copy names the accepted boundaries, and the active Context Tree update has merged.

**Risk level**: A
- Semantic lift: touches daemon startup, client runtime shutdown, server client-owner mismatch handling, local account-switch semantics, and CLI documentation.

**PR Summary**
- Problem: switching a local computer between First Tree users previously required the destructive purge-first flow, even though the server-side client owner boundary should remain single-user and non-transferable.
- Approach: `login <token>` now performs a confirmed local-client switch for different users, parks/restores per-user local state under safety gates, keeps inactive credentials out of parked state, and leaves server `clients.user_id` ownership intact. `computer reset` becomes the formal destructive reset path while `logout --purge` remains compatible.
- Impacted modules: `apps/cli` login/logout/daemon/client-switch core, `packages/client` runtime/provider env propagation and shutdown reason, `packages/server` client registration mismatch copy, CLI docs, and a small skill-evals test timeout stabilization.

**Review findings**
✅ 1. Per-entry switch journal recovery now classifies source/target state and fails closed when ambiguous.
✅ 2. Provider drain now treats unknown commands with trusted switch markers as live descendants.
✅ 3. The paired active Context Tree update (#666) is merged.
⚠️ 4. I did not run tests locally; this review relies on the author's verification and branch CI.

**Action taken**

PR #1409, opened 2026-07-02
feat: add landing campaign trial agents

baixiaohang:
## Summary

- Add the landing campaign start API (`POST /api/v1/me/landing-campaigns/start`) backed by the growth landing feature flag.
- Provision service-managed trial agents in the user's org using the configured First Tree service user and official client, with server-owned campaign skill catalog materialized as agent-scoped resources.
- Move campaign quickstart off the old onboarding kickoff path and update `/quickstart?campaign=...&repo=...` to open a single-run trial chat directly.
- Add server/UI guards for trial agents: locked single-run chat input, ordinary chat/agent/resource mutation blocks, service member/client filtering, and service-user member count exclusion.

This PR intentionally does **not** implement workspace-only/sandbox runtime enforcement; that is a separate follow-up PR.

## Verification

- `pnpm --filter @first-tree/shared test -- src/config/__tests__/server-config.test.ts`
- `pnpm --filter @first-tree/web test -- src/pages/quickstart/__tests__/campaigns.test.ts src/pages/quickstart/__tests__/intent.test.ts src/pages/quickstart/__tests__/quickstart-page.e2e.test.tsx`
- `pnpm --filter @first-tree/server test -- src/__tests__/landing-campaigns-start.test.ts`
- `pnpm --filter @first-tree/server test -- src/__tests__/landing-campaigns-start.test.ts src/__tests__/me-memberships-and-switch-org.test.ts src/__tests__/me-onboarding-kickoff.test.ts`
- `pnpm typecheck`
- `pnpm check` (passes with existing warnings outside this PR)
- `git diff --check`


yuezengwu (changes requested):
Requesting changes.

I reviewed the intended implementation as: the new `/api/v1/me/landing-campaigns/start` endpoint provisions a service-managed trial agent and agent-scoped campaign skill, creates/reuses a locked single-run campaign chat, moves `/quickstart?campaign=...&repo=...` onto that flow, and blocks ordinary user mutations against trial agents/resources/chats. The core metadata/data-shape additions are the new `landingCampaignTrial` agent/chat metadata contracts and growth landing service-user/client config.

Findings:

1. **P1: Trial chats can still be expanded with ordinary agents.**

   The new participant guard in `packages/server/src/api/chats.ts:493-499` only checks the agents being added. It rejects adding a trial agent to an ordinary chat, but it does not reject adding a normal visible agent to an existing landing-campaign trial chat. After `POST /me/landing-campaigns/start`, the human speaker still has access to `POST /chats/:trialChatId/participants`; passing any ordinary org-visible agent survives `assertNoLandingCampaignTrialAgents(...)` and `addMeChatParticipants(...)` mutates the supposedly single-run trial chat.

   That breaks the server-owned isolated trial semantics and can leak future trial messages as silent context to another runtime participant. Please block participant mutation when the target chat metadata parses as a landing campaign trial chat (unless there is a deliberately supported internal/system path), and add a regression test that starts a campaign then tries to add a normal agent to that trial chat.

2. **P1: Trial agents are still exposed as addressable/default chat candidates even though normal chat creation now rejects them.**

   The new guards at `packages/server/src/api/orgs/chats.ts:133-165` and `packages/server/src/api/agents.ts:321-324` make landing campaign agents unusable for ordinary chat creation. However, the server-side candidate queries still treat those same rows as ordinary active org-visible agents: `listAgentsForMember(..., addressableOnly=true)` only adds `agentAddressableCondition()`, and `getNewChatDefaultCandidate()` can return the org-visible fallback at `packages/server/src/services/agent.ts:989-998`. A fresh landing-flow org with no personal agent is the easiest repro: the trial agent is created with `visibility: "organization"`, then New Chat can auto-select it or the participant picker can show it, and the eventual create-chat request fails with the new 403 guard.

   Please exclude `landingCampaignTrial` agents from addressable lists/default-candidate selection (and from `/me`'s `hasUsableAgent` if that bit continues to mean ordinary chat usability), with coverage for `/orgs/:orgId/agents?addressableOnly=true`, `/orgs/:orgId/agents/new-chat-default-candidates`, and the landing-start path.

I did not find a DB migration in this PR, but the new agent/chat metadata contracts are core data-shape changes and should get an explicit human pass before merge.

yuezengwu (approved):
Re-reviewed the synchronize update at `209e427d`.

The core change remains the landing campaign start flow: service-managed single-run trial agents, agent-scoped campaign resources, quickstart routing into the locked trial chat, and server/UI guards around ordinary agent/chat/resource mutation.

The new commit fixes the two isolation blockers I called out earlier:

- landing campaign trial chats now reject participant mutation on both the web `/chats/:id/participants` path and the agent-runtime add/remove participant paths;
- landing campaign trial agents are no longer exposed as ordinary addressable/default/usable chat candidates, and ordinary chat creation/invite paths reject trial agents at the service layer.

I did not find remaining blocking issues in the updated diff. CodeQL is green on the latest head. The PR is still marked `CONFLICTING`, so the branch still needs conflict resolution before merge. I also could not run the focused local Vitest command in this worktree because `node_modules` are not installed (`vitest: command not found`).

yuezengwu (approved):
Re-reviewed the latest synchronized head `6ab824a` after the merge from `origin/main`.

This update resolves the merge conflict and keeps the landing campaign flow on the new `/me/landing-campaigns/start` path. I specifically checked the conflict-adjacent areas:

- `/api/me` keeps the service-user member-count exclusion and now returns `410 campaign_kickoff_moved` for the old campaign kickoff path;
- campaign scan skill content is moved behind the landing-campaign catalog/re-export without losing the previous trial-agent behavior;
- the new setup CTA URL is resolved from the request public URL and templated into the agent-private skill payload, with the payload comparison using the templated body so same-env retries remain idempotent.

No new blocking issues found. PR is now mergeable. CodeQL and lint/typecheck were green when checked; the GitHub Test job was still running, and I did not wait for CI per review convention.

PR #1410, opened 2026-07-02
refactor: simplify onboarding kickoff chat

Gandy2025:
## Summary

- Route onboarding first messages through task `createChat` instead of the old agent-only/system directive path.
- Send natural user-visible kickoff copy under the human agent, with no onboarding `kind` in the web contract and no campaign metadata on messages.
- Add a dedicated tree-setup kickoff endpoint/key while preserving legacy `kind=tree` and old `...:tree` status compatibility.
- Update `first-tree-welcome` routing and no-project guidance so the agent asks for a local project folder path or GitHub repo URL from the skill body, not the first visible message.

## Validation

- `../../node_modules/.bin/vitest run src/__tests__/me-extras-schema.test.ts` from `packages/shared`
- `../../node_modules/.bin/vitest run src/__tests__/agent-io.test.ts` from `packages/client`
- `../../node_modules/.bin/vitest run src/__tests__/me-onboarding-kickoff.test.ts src/__tests__/message-addressed-agents.test.ts src/__tests__/campaign-scan-skill.test.ts` from `packages/server`
- `../../node_modules/.bin/vitest run src/suites/first-tree-welcome/__tests__/floor.test.ts src/suites/first-tree-welcome/__tests__/grader.test.ts` from `packages/skill-evals`
- `../../node_modules/.bin/vitest run src/pages/workspace/center/onboarding/__tests__/bootstrap-prose.test.ts src/pages/quickstart/__tests__/campaigns.test.ts src/pages/quickstart/__tests__/quickstart-page.e2e.test.tsx src/pages/__tests__/web-dom-interactions.test.tsx src/components/chat/__tests__/chat-offline-notice.test.tsx src/pages/workspace/center/__tests__/chat-view-dom.test.tsx` from `packages/web`
- `../../node_modules/.bin/tsc --noEmit` from `packages/shared`, `packages/server`, `packages/client`, `packages/web`, and `packages/skill-evals`
- `node_modules/.bin/biome check $(git diff --name-only --diff-filter=ACMR -- '*.ts' '*.tsx')`
- `git diff --check`

Note: the web tests still emit existing Radix Dialog warnings and localhost `ECONNREFUSED` noise, but the test command exits 0.

PR #1413, opened 2026-07-02
fix: harden first-tree-welcome scope guard and copy-drift coverage

Gandy2025:
## Summary

Follow-up to #1410, addressing review findings on the natural-language `first-tree-welcome` skill. Since #1410 (correctly) retires the hidden onboarding directive and skill activation is now description-driven (see `docs/development/onboarding-kickoff-contract.md`), the skill's own prose + descriptions are the disambiguation layer — this PR tightens that layer and adds a drift guard.

**Stacked on #1410** — base branch is `feat/simplify-onboarding-kickoff`. Merge #1410 first, then this.

## Changes

- **Scope guard for scans (the one consistency defect):** the frontmatter `description` already excludes `repo scans`, but the `## Scope` body omitted it. Added it, plus a clause: a scan chat opens with the same `welcome aboard` line then asks for a scan/readiness report → run its own bound scan skill, not this launcher. This is the guard that now does the work, since scan and welcome chats share the opener and the deterministic router is gone.
- **First Response Contract B / Matrix row 5 seam:** B ("evidence available") now requires the evidence be *readable* and routes a connected-but-unreadable repo to row 5 (report the read failure, ask for recovery) instead of faking understanding or sending a menu.
- **Copy-drift guard (test):** a floor test binds the skill's example trigger phrases to the real onboarding bootstraps in `bootstrap-prose.ts`, so a product-copy reword can't silently drift skill selection.
- **Briefing skill-map:** refreshed the `first-tree-welcome` "Load when" row — onboarding activation is a natural user message, not a "system message" (stale after #1410).
- **VERSION re-sync:** the `VERSION` file (the real propagation gate) had drifted from the frontmatter; both now at `1.0.7`.

## Validation

- `pnpm --filter @first-tree/skill-evals exec vitest run src/suites/first-tree-welcome/__tests__/floor.test.ts` → 10/10 (incl. the new drift-binding case)
- `packages/client` `agent-briefing.test.ts` → 38/38
- `pnpm --filter @first-tree/skill-evals exec tsc --noEmit` and `@first-tree/client` tsc → clean
- `biome check` on changed TS → clean
- `python3 ./scripts/quick_validate_skill.py ./skills/first-tree-welcome` → valid

## Deliberately deferred (separate follow-ups, not in this PR)

- **Promote the admin+empty-tree "offer Build your Context Tree" eval row to the CI gate.** It's the PR's headline behavior but only runs in the opt-in periodic tier. Deferred because it's an eval-policy change (live-gate count 3 → 4) whose live-model pass I can't verify locally — better owned by the eval maintainer.
- **Scan-vs-welcome negative-selection eval.** Needs a harness change so "skill correctly declined" counts as pass (today `grader.ts` hard-fails any case where the skill wasn't read), which affects all skills' grading.
- **Tree-setup bootstrap jargon.** `buildTreeSetupBootstrap` shows `first-tree-read/seed/write` in a user-visible bubble — but those names currently double as the tree-lane routing hint that replaced the deleted directive, so de-jargoning it is a dual-reader design change, not a one-liner.


yuezengwu (changes requested):
Thanks, this is very close, but I found one blocker before this can land.

- `skills/first-tree-welcome/agents/openai.yaml:3` still carries the old agent-facing description: it says the skill applies when an onboarding start-chat message explicitly names `first-tree-welcome`, and it still does not exclude repo scans. That file is part of the shipped payload (`copy-bundled-skills.mjs` copies the whole `skills/<name>/` directory, and the installer copies it into `.agents/skills/<name>/`), and the existing drift-guard comments say `skills/<name>/agents/*.yaml` is used by composer/runtime for routing metadata. Since #1410 retired the hidden onboarding directive and this PR is relying on skill descriptions as the disambiguation layer, leaving `agents/openai.yaml` stale means the OpenAI/Codex-facing skill selection surface can still follow the retired explicit-name trigger and still miss the new repo-scan exclusion even though `SKILL.md` and the briefing were tightened.

Please sync `agents/openai.yaml` with the new description/scope, and ideally extend the floor drift guard to bind this metadata file alongside `SKILL.md` so the next copy reword cannot drift one of the real routing surfaces.

Gandy2025:
Good catch — that's a real hole, thanks. Fixed in 6f6ea13c.

`skills/first-tree-welcome/agents/openai.yaml` still carried the pre-#1410 description verbatim (the retired `explicitly names first-tree-welcome` trigger, and no `repo scans` exclusion). Since #1410 retires the hidden directive and this PR leans on descriptions as the disambiguation layer, leaving that OpenAI/Codex routing surface stale would let it keep following the retired trigger and miss the scan exclusion — exactly as you describe.

Changes:
- **Synced** `agents/openai.yaml` `description` to the current `SKILL.md` description (natural-language openers + `repo scans` / tree-setup / ordinary-chat exclusions; dropped the explicit-name trigger).
- **Extended the floor drift guard** (`floor.test.ts`) with a case that asserts `agents/openai.yaml`'s description stays byte-identical to `SKILL.md`'s, and specifically that it never contains `explicitly names first-tree-welcome` and always contains `repo scans` — so a future copy reword can't drift one routing surface without the other.

Verified locally: floor suite 11/11 (incl. the new binding), `retired-tree-commands-drift` 3/3, `validate:skill` valid, biome + tsc (skill-evals & client) clean.

yuezengwu (approved):
Approved after the 6f6ea13c follow-up.

The previous blocker is addressed: `skills/first-tree-welcome/agents/openai.yaml` now matches the `SKILL.md` description, drops the retired explicit-name trigger, and includes the repo-scan exclusion. The added floor guard also binds the OpenAI/Codex metadata description to `SKILL.md`, which covers the drift class I flagged.

I rechecked the stacked diff and ran the lightweight local checks available in this fresh worktree: `quick_validate_skill.py`, `git diff --check`, and a direct description-alignment assertion for `SKILL.md` vs `agents/openai.yaml`. CI still had Test / Portable CLI Smoke pending when I approved; Lint & Type Check and Validate Skill were already green.

baixiaohang (approved):
Approved after re-reviewing the 6f6ea13c follow-up.

The previous blocker is addressed: `skills/first-tree-welcome/agents/openai.yaml` now matches the `SKILL.md` description, drops the retired explicit-name trigger, and includes the repo-scan exclusion. The added floor guard binds the OpenAI/Codex metadata description to `SKILL.md`, which covers the routing-surface drift class.

I also confirmed the PR head is still `6f6ea13c` and GitHub CI is green.

yuezengwu (approved):
Re-approved after the force update to `62fa9548` / retarget to `main`.

I compared the previous approved stack (`83c0f5c3`, `6f6ea13c` on `feat/simplify-onboarding-kickoff`) with the new stack (`e311e946`, `62fa9548` on `main`) using `git range-diff`; the two PR commits are patch-equivalent, and the removed commits are the now-merged #1410 base commits. The reviewed changes remain the same: welcome scope/readability wording, bootstrap/openai metadata drift guards, briefing row refresh, and VERSION sync.

Lightweight local checks on the new head passed: `quick_validate_skill.py`, `git diff --check`, and direct `SKILL.md` vs `agents/openai.yaml` description alignment. GitHub CI still had some jobs pending when I reviewed; Validate Skill and the early checks were green.

baixiaohang (approved):
## Review by codex-assistant - PR #1413

**Recommendation**: approve
- Rationale: The current head only tightens first-tree-welcome routing prose/metadata, adds drift guards, and syncs version propagation; I found no blocking issue in the onboarding, tree-skill, or repo-scan boundaries.

**Risk level**: A
- Path baseline: touches `packages/client/**`.

**PR summary**
- Problem: Onboarding first chats, repo scan chats, and tree setup chats now rely on visible natural-language prompts plus skill descriptions, so stale or incomplete welcome-skill copy could route agents into the wrong first-work flow.
- Approach: The welcome skill now explicitly excludes repo scans in both the description and scope body, treats connected-but-unreadable repos as a read-failure branch instead of evidence, keeps OpenAI/Codex routing metadata aligned with `SKILL.md`, and adds floor tests that bind trigger examples and metadata descriptions to the shipped copy.
- Impacted modules: `skills/first-tree-welcome`, `packages/skill-evals`, `packages/client` briefing.

**Review findings**
✅ 1. The previous routing-surface blocker is addressed: `agents/openai.yaml` now matches `SKILL.md`, drops the retired explicit-name trigger, and includes the repo-scan exclusion.
✅ 2. The new scope/readability wording matches the onboarding and skill-boundary contracts: welcome stays value-first, repo scans are excluded, and unreadable repos do not become fake evidence.
✅ 3. The added floor guard covers the copy-drift class that could otherwise silently desync visible onboarding bootstraps, `SKILL.md`, and OpenAI/Codex metadata.

**Action taken**
- Approved after explicit authorization from baixiaohang. I did not run local tests per PR review policy; GitHub checks are green on head `62fa9548`.

Gandy2025:
Superseded by #1423, which consolidates the welcome-skill work into one PR: the reorg already subsumes this PR's SKILL.md hardening, and its unique changes (the agent-briefing tree-bound welcome row, openai.yaml sync, and floor drift guards) are folded into #1423 — plus the tree-less welcome row (L1195) fix and a both-rows drift lock. Closing in favor of #1423.

PR #1408, opened 2026-07-02
fix: install first-tree-seed for tree-less onboarding

Gandy2025:
**Paired Context Tree PR (draft, merges after this):** agent-team-foundation/first-tree-context#668 — reclassifies the skill topology (seed/write core) in the tree nodes.

## Problem

New teams onboarding from zero can't build a proper Context Tree. `first-tree-welcome` spawns a dedicated tree-build chat that loads `first-tree-seed` to create + bind + seed the tree — but that chat is **tree-less** (no binding yet), and `first-tree-seed` was gated behind `TREE_SKILL_NAMES` (installed only when a Context Tree binding exists). So seed was never on disk in the build chat; the agent fell back to raw `gh` and produced an unbound, unseeded placeholder repo instead of a real Context Tree.

## Fix

Move `first-tree-seed` and `first-tree-write` from `TREE_SKILL_NAMES` to `CORE_SKILL_NAMES` (installed on every bootstrap, tree or not):

- **`first-tree-seed`** is the *from-zero* build skill — it must be available precisely when there's no tree yet. Gating it behind a binding is a chicken-and-egg: the skill whose job is to create the binding required the binding to already exist.
- **`first-tree-write`** is seed's hard dependency (seed's Required Reading loads it), so it must ship alongside.
- **`first-tree-read`** stays TREE — it only operates on an existing tree.

Guard `reconcileTreeSkillState` so the moved skills aren't deleted as "dropped TREE skills": `installCoreSkills` places seed/write earlier in the same bootstrap, and the subsequent tree reconcile must not strip them. Removal now targets only names that left **both** tiers (genuinely retired skills). `CORE ∪ TREE` is unchanged, so `BUNDLED_SKILLS` and the briefing family map stay in sync.

## Scope

Client skill-classification only. Does **not** touch the server-side onboarding slimming (kickoff `kind` removal, CTA routing, recovery probe) — that's a separate change.

## Tests

- **New reconcile regression test** (`skills-state-reconcile.test.ts`): an agent whose managed-state ledger still lists seed/write as TREE skills keeps their payloads after upgrade. Verified load-bearing — it fails if the guard is reverted.
- Repointed the copy-mechanics `describe` in `bootstrap.test.ts` to `installCoreSkills` (now the 3-skill set) so version-drift / partial-failure / fast-path / symlink-repair coverage survives.
- Full client suite green (1189/1189); `pnpm check` (biome) + `pnpm typecheck` clean; independent adversarial review passed.

[...]

PR #1415, opened 2026-07-03
fix(client): install first-tree-seed + first-tree-write as core skills

Gandy2025:
## Problem
`first-tree-welcome` delegates repo creation / binding / seeding to `first-tree-seed`, which runs in a **tree-less** spawned build-tree chat (it exists precisely because no Context Tree binding exists yet). The runtime installer shipped `first-tree-seed` and `first-tree-write` only via `TREE_SKILL_NAMES`, which `agent-bootstrap.ts` installs only `if (contextTreePath)`. So in the tree-less build-tree chat, seed was never on disk → the delegation broke → the agent fell back to raw `gh`.

## Fix
- Move `first-tree-seed` and `first-tree-write` from `TREE_SKILL_NAMES` to `CORE_SKILL_NAMES` (always installed, tree or not).
  - `first-tree-write` must move with `first-tree-seed`: `first-tree-seed/SKILL.md` loads `../first-tree-write/SKILL.md` by relative reference and inherits its Hard Rules. Installing seed without write leaves a dangling reference in the tree-less chat.
  - `first-tree-read` stays tree-only — there is nothing to read without a bound tree.
- Guard `reconcileTreeSkillState` so it never removes a `CORE_SKILL_NAMES` member. For an existing tree-bound agent whose `managed.json::skills` still lists seed/write under the old TREE tier, `installCoreSkills` installs them first; without the guard the tree reconcile would delete them immediately after.
- Sync two now-stale rationale comments (`agent-bootstrap.ts`, `agent-briefing.ts`) that assumed tree-less agents have no skill payloads on disk (they now do).

## Tests
- Retarget the multi-skill installer-engine describe block to `installCoreSkills` (now a 3-skill set) to preserve engine coverage (fast-path skip, VERSION drift, content drift, missing-VERSION reinstall, clobbered-symlink repair, partial failure).
- Add an `installFirstTreeIntegration (tree-only skills)` describe covering the `first-tree-read` happy path + per-skill failure log.
- Add a `skills-state-reconcile` regression proving core skills listed in prior managed state survive the reconcile (mutation-verified: fails without the guard).
- `@first-tree/client`: 1191 tests pass; `typecheck` + `biome` clean.

No DB / schema / API changes. Implementation-only — makes the already-documented `welcome → seed` delegation actually work; no Context Tree PR needed.

### DIFF
--- a/packages/client/src/__tests__/agent-briefing.test.ts
+++ b/packages/client/src/__tests__/agent-briefing.test.ts
@@ -383,12 +383,14 @@ describe("buildAgentBriefing — # Required Reading (unconditional skill-load ma
     expect(contextTreeIdx).toBeGreaterThan(requiredIdx);
   });
 
-  it("omits # Required Reading for tree-less agents (the skill payloads are not installed on disk for them)", () => {
-    // `installFirstTreeIntegration` is short-circuited when
-    // `contextTreePath === null`, so the SKILL.md payloads under
-    // `<workspace>/.agents/skills/` never get materialised. Mandating
-    // a load would point at files that aren't there. The Tree
-    // Location stub already surfaces the gap to a human.
+  it("omits # Required Reading for tree-less agents (tree-oriented guidance is gated on a binding)", () => {
+    // The `# Required Reading` mandate + family map are gated on
+    // `contextTreePath !== null` — Context-Tree-oriented guidance surfaced
+    // only once the agent is tree-bound. (Since seed/write are core, their
+    // payloads DO exist on disk in tree-less workspaces; the gate is a
+    // scoping choice, not a missing-file guard. Tree-less build-tree chats
+    // reach write's rules via `first-tree-seed`'s own reference to it.)
+    // The Tree Location stub already surfaces the no-tree state to a human.
     const briefing = buildAgentBriefing(makeOpts({ contextTreePath: null }));
     expect(briefing).not.toContain("# Required Reading");
   });
--- a/packages/client/src/__tests__/bootstrap.test.ts
+++ b/packages/client/src/__tests__/bootstrap.test.ts
@@ -358,8 +358,14 @@ function makeFixtureSkillsRoot(
   return root;
 }
 
-describe("installFirstTreeIntegration (inline skill installer)", () => {
-  const TREE_SKILLS = ["first-tree-write", "first-tree-read", "first-tree-seed"];
+// Exercises the shared inline installer engine (fast-path skip, VERSION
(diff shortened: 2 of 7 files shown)


yuezengwu (changes requested):
Blocking: this PR is stale against current `main` and should not merge in this form.

`main` already contains `2d155c65` / #1408 (`fix: install first-tree-seed for tree-less onboarding`), which implements the same core-skill promotion and reconcile protection. The remaining delta in #1415 conflicts with those files and would drop part of the already-merged behavior if resolved by taking this branch's side: `packages/client/src/runtime/agent-briefing.ts` still gates the First Tree Family map on `contextTreePath !== null`, and `packages/client/src/__tests__/agent-briefing.test.ts` asserts that tree-less agents omit the map. In #1408, tree-less agents get a core-only map (`welcome` + `seed` + `write`) so the welcome-spawned tree-build chat has a routing surface to `first-tree-seed` while `first-tree-read` remains tree-only.

There is also stale rationale in `requiredReadingSection` saying tree-less agents do not have the payloads on disk, which is no longer true after seed/write became core. #1408 already corrected that commentary.

I verified the PR head itself with the client suite: `pnpm --filter @first-tree/client test -- bootstrap.test.ts skills-state-reconcile.test.ts agent-briefing.test.ts agent-bootstrap.test.ts bundled-skill-list-sync.test.ts` passed 1191 tests, `pnpm --filter @first-tree/client typecheck` passed, and `pnpm check` exited 0. But the branch is merge-conflicting and superseded by #1408. Please rebase and reduce this to any real remaining delta, or close it if there is none.

Gandy2025:
Thanks @yuezengwu — you're right, closing as superseded.

Verified: #1408 (`2d155c65`) landed as the commit immediately after this branch's base (`5cfae576`), so this PR is a parallel duplicate of the same `first-tree-seed` + `first-tree-write` → `CORE_SKILL_NAMES` promotion and the reconcile guard. #1408 is also strictly more complete — it reworks `firstTreeFamilyMap` to emit a CORE-only family map (`welcome` + `seed` + `write`) for tree-less agents as the routing surface to `first-tree-seed` (while `first-tree-read` stays tree-only and `# Required Reading` stays tree-bound). This PR deliberately deferred that; #1408 already did it.

The only delta here not in #1408 is a one-line stale comment at `workspace-migrations.test.ts:210` (`// still in TREE_SKILL_NAMES` — `first-tree-write` is core now). Not worth keeping this open for; I'll fold it into a nearby follow-up if it still matters. Closing.

PR #1416, opened 2026-07-03
feat: sandbox landing campaign runtime

baixiaohang:
## Summary

- Add `FIRST_TREE_LANDING_CAMPAIGN_RUNTIME_PROVIDER` with a default of `codex`, and fail closed for `claude-code` until Claude workspace-only support exists.
- Run Codex landing campaign trial agents through a Linux `bubblewrap` workspace-only spawn path keyed by `agent.metadata.landingCampaignTrial`, leaving ordinary Codex/Claude agents unaffected.
- Add a scoped `agent_outbox` token and sandbox-local First Tree home so trial agents can use `chat send` without exposing the service user's real local credentials.
- Guard landing campaign trial chat writes in the shared message service, including terminal-state and concurrent final/request transitions.
- Keep the official landing campaign client ordinary for non-trial use; trial semantics are tied to agent/chat metadata rather than client id.

## Not in this PR

- Claude Code workspace-only wrapper/provider resolver.
- Provider-specific landing campaign clients or skill catalog.
- GitHub token, network, CPU, memory, or disk quota governance.
- Reserved official client id registration guard.

## Verification

- `pnpm --filter @first-tree/shared exec vitest run --maxWorkers=2 src/config/__tests__/server-config.test.ts`
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-thread-options.test.ts src/__tests__/codex-handler-engine.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/landing-campaigns-start.test.ts src/__tests__/route-conventions.test.ts`
- `pnpm typecheck`
- `pnpm check` (exits 0 with existing warnings/info)
- `git diff --check`

## Review

Reviewed in First Tree chat by `codex-reviewer`; latest review found no blocking issues.

### DIFF
--- a/docs/development/http-path-conventions.md
+++ b/docs/development/http-path-conventions.md
@@ -9,11 +9,16 @@
 
 ## JWT scope
 
-JWT carries **only** `userId`. `request.user` is `{ userId }`. Anything beyond that — org, role, member — is resolved per-request via `scope/require-*` helpers, which probe the DB in real time. Reading `request.user.organizationId / memberId / role` is impossible (the fields don't exist) — that is the type-level enforcement of the rule.
+Normal user JWTs carry **only** `userId`. `request.user` is `{ userId }`. Anything beyond that — org, role, member — is resolved per-request via `scope/require-*` helpers, which probe the DB in real time. Reading `request.user.organizationId / memberId / role` is impossible (the fields don't exist) — that is the type-level enforcement of the rule.
 
 ```ts
 type AccessTokenPayload = { sub: string; type: "access"; iat: number; exp: number; jti: string };
 ```
+
+The only exception is the `agent_outbox` token used by workspace-only trial
+sandboxes. It is not a general user access token: it is accepted only for the
+current agent's `POST /api/v1/agent/chats/:chatId/messages` route in the
+current chat, and route middleware must reject it everywhere else.
 
 ## Decision tree
 
@@ -81,7 +86,7 @@ Example: `GET /api/v1/agent/me`, `POST /api/v1/agent/chats/:chatId/messages`, `W
 ## Forbidden
 
 - ❌ `/admin/...` prefix anywhere — role lives in middleware, not URL
-- ❌ Reading `request.user.organizationId` / `memberId` / `role` — these fields don't exist; JWT carries only `sub`
+- ❌ Reading `request.user.organizationId` / `memberId` / `role` — these fields don't exist; normal user JWTs carry only `sub`
 - ❌ `?organizationId=` query param for org scope — use `:orgId` path param
 - ❌ Class B route without `:orgId` in path — `requireOrgMembership` will fail to type-check
 - ❌ Class C route with `:orgId` in path — redundant; resource UUID locates the org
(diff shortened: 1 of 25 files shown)


yuezengwu (approved):
Reviewed the sandbox landing campaign runtime changes.

I did not find blocking issues in the implementation. The scoped `agent_outbox` path is route-scoped to the trial agent/chat message POST, and the shared message service guards the landing-campaign chat state transitions, including terminal-state and concurrent final/request races.

Verification run locally:
- `pnpm --filter @first-tree/shared exec vitest run --maxWorkers=2 src/config/__tests__/server-config.test.ts` passed
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/landing-campaigns-start.test.ts src/__tests__/route-conventions.test.ts` passed
- `pnpm --filter @first-tree/server typecheck` passed
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-thread-options.test.ts src/__tests__/codex-handler-engine.test.ts` failed on macOS only: 5 assertions compare `/var/...` against implementation-realpathed `/private/var/...` in the new workspace sandbox tests. This looks like test expectation normalization rather than a Linux sandbox runtime issue; CI is green.

Non-blocking follow-up to consider: the sandbox-local CLI uses a non-refreshable `agent_outbox` token with the normal access-token lifetime. If a trial runs past that TTL before calling `chat send`, the send will fail with 401; if long scans are expected, we should either mint a longer scoped outbox token or refresh it from the host side without exposing the service user's refresh token.

yuezengwu (commented):
Re-checked the synchronized commit `b658a5a` (`test: normalize workspace sandbox tmp paths`). The delta is test-only and `workspace-sandbox.test.ts` now passes locally on macOS.

One local verification issue remains: the advertised focused client command still fails on macOS at `packages/client/src/__tests__/codex-app-server-handler.test.ts:485`:

```text
expected .../outbox-home to be .../outbox-home
Received: /private/var/...
Expected: /var/...
```

Root cause looks like the same temp-path normalization gap: `workspace-sandbox.test.ts` now realpaths its `mkdtempSync(...)` root, but `codex-app-server-handler.test.ts` still assigns `workspaceRoot = mkdtempSync(join(tmpdir(), "ft-codex-app-server-"))` and later expects `FIRST_TREE_HOME` from that raw path. The implementation realpaths the workspace before preparing the sandbox-local home.

Suggested fix: normalize `workspaceRoot` in that test's `beforeEach` as well, or compare `capturedEnv.FIRST_TREE_HOME` against a realpathed expected workspace root.

I am not changing the approval on runtime correctness; this is a remaining macOS test-portability issue in the stated verification command.

yuezengwu (commented):
Re-checked the latest synchronized commit `03d5ec2` (`test: normalize codex app server tmp paths`). This is test-only and applies the same `realpathSync` normalization to `codex-app-server-handler.test.ts`.

Local macOS verification now passes for the previously failing focused command:

```text
pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-thread-options.test.ts src/__tests__/codex-handler-engine.test.ts
```

Result: 4 test files passed, 50 tests passed.

`git diff --check b658a5a..03d5ec2` also passes. No new runtime/auth/message-service code changed in this sync, and the earlier approval still stands.

PR #1417, opened 2026-07-03
fix: bound landing service visibility to service org

baixiaohang:
## Summary
- add `FIRST_TREE_LANDING_CAMPAIGN_SERVICE_ORG_ID` / `landingCampaigns.serviceOrgId`
- treat the landing service user/member/client as ordinary inside the service org and campaign-managed only in other orgs
- require the official landing campaign client to belong to the configured service org before starting a trial
- hide service-owned computers from customer org admin lists while keeping service org lists ordinary
- show landing trial agent owners as `First Tree managed` instead of leaking a hidden member id

## Review
- Code-only review completed by `codex-reviewer`; no blockers after follow-up fixes.

## Testing
- Not run, per request.
- `git diff --check` passed.

### DIFF
--- a/packages/server/src/__tests__/helpers.ts
+++ b/packages/server/src/__tests__/helpers.ts
@@ -60,6 +60,7 @@ export type CreateTestAppOptions = {
   channel?: Config["channel"];
   growthLandingPagesEnabled?: boolean;
   landingCampaignServiceUserId?: string;
+  landingCampaignServiceOrgId?: string;
   landingCampaignClientId?: string;
   landingCampaignRuntimeProvider?: "codex" | "claude-code";
   commandVersion?: string;
@@ -94,11 +95,13 @@ export async function createTestApp(opts: CreateTestAppOptions = {}): Promise<Fa
     growth: {
       landingPagesEnabled: opts.growthLandingPagesEnabled ?? false,
       ...(opts.landingCampaignServiceUserId !== undefined ||
+      opts.landingCampaignServiceOrgId !== undefined ||
       opts.landingCampaignClientId !== undefined ||
       opts.landingCampaignRuntimeProvider !== undefined
         ? {
             landingCampaigns: {
               serviceUserId: opts.landingCampaignServiceUserId,
+              serviceOrgId: opts.landingCampaignServiceOrgId,
               clientId: opts.landingCampaignClientId,
               runtimeProvider: opts.landingCampaignRuntimeProvider ?? "codex",
             },
--- a/packages/server/src/__tests__/landing-campaigns-start.test.ts
+++ b/packages/server/src/__tests__/landing-campaigns-start.test.ts
@@ -12,20 +12,34 @@ import { chats } from "../db/schema/chats.js";
 import { clients } from "../db/schema/clients.js";
 import { members } from "../db/schema/members.js";
 import { messages } from "../db/schema/messages.js";
+import { organizations } from "../db/schema/organizations.js";
 import { resources } from "../db/schema/resources.js";
 import { users } from "../db/schema/users.js";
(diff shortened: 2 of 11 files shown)


yuezengwu (changes requested):
Implementation goal / core changes:

This PR scopes the landing campaign service identity to a configured service org: the same service user/client is ordinary inside that org, but campaign-managed in customer orgs. It also adds `FIRST_TREE_LANDING_CAMPAIGN_SERVICE_ORG_ID`, requires the official landing client to be registered against that org before trial start, hides service-owned rows from customer org admin/member views, blocks trial start inside the service org, and masks landing-trial owner display in the agent detail UI.

Blocking issues:

1. `packages/server/src/__tests__/landing-campaigns-start.test.ts:981` does not create an "another org admin". `createTestAdmin(app)` always uses `resolveDefaultOrgId`, so `otherAdmin` lands in the same org as `admin`. The request is therefore hitting the campaign service member inside its own customer org, where the expected response is the existing 403 guard, not scoped 404. This is failing locally and in CI:

   `pnpm --filter @first-tree/server test -- src/__tests__/landing-campaigns-start.test.ts`

   Failure:

   `expected 403 to be 404` at `landing-campaigns-start.test.ts:988`.

   Please create a real second org/admin membership for this regression, following the existing `attachOrg` / manual `organizations` + `members` test pattern, then assert PATCH and DELETE return 404 against that different org.

2. `pnpm check` fails on PR-touched files. The relevant errors are:

   - `packages/server/src/__tests__/landing-campaigns-start.test.ts` needs Biome formatting for the `onConflictDoUpdate` `set` object and the client update chain.
   - `packages/web/src/pages/agent-detail/identity-section.tsx:1` import order needs Biome organize-imports (`type Agent` before `isLandingCampaignTrialAgentMetadata`).

Non-blocking rollout risk to consider:

If `serviceUserId` is configured but the new `serviceOrgId` is missing during a partial rollout, `isLandingCampaignServiceMembership` currently returns false, so existing customer-org service memberships/clients would stop being hidden until the new env is present. If deployment always adds the env atomically with this code, this is acceptable; otherwise consider making the guard fail closed for the service user when `serviceOrgId` is absent.

Validation run:

- `pnpm install --frozen-lockfile` passed.
- `pnpm --filter @first-tree/shared test -- src/config/__tests__/server-config.test.ts` passed.
- `pnpm --filter @first-tree/web test -- src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx` passed.
- `pnpm typecheck` passed.
- `pnpm check` failed as above.
- Server tests failed on the new scoped-404 test as above.

yuezengwu (approved):
Re-review approved.

Implementation goal / core changes remain the same: this scopes the landing campaign service user/client to the configured service org, treats that identity as ordinary inside the service org and campaign-managed in customer orgs, requires the official client to belong to the service org before starting a trial, hides service-owned customer-org rows, and masks landing trial owner display as `First Tree managed`.

The follow-up fixes address the previous blockers:

- The scoped-404 regression now creates a real second org/admin via `attachOrg`, so it exercises the cross-org not-found path instead of the same-org 403 guard.
- The PR-touched Biome formatting/import-order issues are fixed.

No DB/schema or core data-structure changes in this PR.

Local verification:

- `pnpm check`
- `pnpm --filter @first-tree/server test -- src/__tests__/landing-campaigns-start.test.ts` (server suite passed: 176 files / 1666 tests)
- `pnpm --filter @first-tree/shared test -- src/config/__tests__/server-config.test.ts`
- `pnpm --filter @first-tree/web test -- src/pages/agent-detail/__tests__/appearance-identity-options-dom.test.tsx`
- `pnpm typecheck`

PR #1421, opened 2026-07-03
[codex] implement runtime-bound agent switching

yuezengwu:
## Summary

- add runtime-session-bound agent HTTP enforcement and SDK/CLI token plumbing
- add `switch-runtime` server flow with metadata claim/CAS, claim guards, session eviction, recovery, offline target eligibility, and deterministic fault injection coverage
- add CLI live runtime convergence for provider/client switches, bind mismatch repair, YAML preservation, and same-runtime retry paths
- add Web Runtime tab switching UX with destructive confirmation, recovery action, and mobile/desktop coverage

## Validation

- `pnpm --filter first-tree-dev test -- src/__tests__/client-runtime-context-tree.test.ts`
- `pnpm --filter @first-tree/server test -- src/__tests__/agent-runtime-switch.test.ts`
- `pnpm --filter @first-tree/web test -- src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx`
- `pnpm typecheck`
- `pnpm exec biome check .` (exit 0; existing unrelated warnings remain)
- `git diff --check`

## Notes

- No DB schema migration; runtime switch claim is stored temporarily in `agents.metadata.runtimeSwitch`.
- QA retested the prior proposal gaps and CLI convergence blocker. The only black-box QA gap was forced failed-slot retry because connected daemon failed slots are process-local state with no user/API construction path; this PR now has a direct CLI unit assertion for `failed + same-runtime agent:pinned -> retry`.


baixiaohang (changes requested):
## Review by codex-assistant - PR #1421

- Rationale: the runtime switch path introduces two new state boundaries that can wedge agents or fail under the server's stateless multi-instance contract.

**Risk level**: A
- Path baseline: touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes server agent/session/runtime binding flows and DB metadata semantics.

**PR Summary**
- Problem: operators need to move an existing managed agent to a different computer/runtime without leaving the old local runtime able to keep acting as that agent.
- Approach: add runtime-session-bound HTTP headers, a two-phase server `switch-runtime` claim in `agents.metadata`, session eviction/recovery, CLI local runtime convergence, and Web controls for switching and recovery.
- Impacted modules: `apps/cli/src/core/client-runtime.ts`, `packages/client/src/{client-connection,sdk,runtime}`, `packages/server/src/{api,services,middleware}`, `packages/shared/src/schemas`, and `packages/web/src/pages/agent-detail`.

**Review findings**
❌ 1. Runtime-session enforcement and switch side effects are stored in process memory, so they break the stateless/multi-instance server contract. `bindAgentToClient` mints the HTTP runtime token into the module-level `agentRuntimeSessions` map, and `agentSelectorHook` validates only against that local map (`packages/server/src/services/connection-manager.ts:15`, `packages/server/src/middleware/agent-selector.ts:103`). Once enforcement is enabled, a valid agent HTTP request routed to a different server instance from the WebSocket bind will be rejected. The same locality issue affects the switch itself: `switchAgentRuntime` relies on `forceDisconnect` to stop the old slot (`packages/server/src/services/agent-runtime-switch.ts:311`), but an HTTP request handled by an instance that does not own that WebSocket cannot actually disconnect or unsubscribe the old runtime before the DB moves the agent to the new client/provider. This needs a PostgreSQL-backed runtime-session binding/nonce, or an equivalent cross-instance route/revocation mechanism, before the enforcement flag can safely gate runtime switching. [R1/R5]

❌ 2. `metadata.runtimeSwitch` is now an internal recovery claim, but public agent create/update can still write arbitrary `metadata`, which can permanently wedge an agent. The shared schemas still accept arbitrary metadata (`packages/shared/src/schemas/agent.ts:132`, `packages/shared/src/schemas/agent.ts:164`), the PATCH route only checks whether the existing row already has a switch claim (`packages/server/src/api/agents.ts:134`), and the service stores supplied metadata wholesale (`packages/server/src/services/agent.ts:1043`). A caller can create or patch an agent with `metadata.runtimeSwitch` set to a malformed or fake value; after that, normal update/suspend/delete/switch paths all hit `assertNoRuntimeSwitchInProgress`, while `/switch-runtime/recover` rejects malformed recovery state. Please reject or strip `runtimeSwitch` on every user-supplied metadata write, or move the claim out of user-editable metadata. [R1/R4]

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the two requested-change findings in `b90ca2f0`.

What changed:
- Moved runtime-session HTTP authority out of process-local state. WS bind now mints a raw token, stores only a hash in durable `agents.metadata.runtimeSession`, and HTTP enforcement validates against DB state. Switch/disconnect/suspend revoke the DB session so stale tokens fail across server instances.
- Added cross-instance route-change fanout through PG NOTIFY. Runtime switch/recovery now broadcasts an agent route event so the instance that owns the old/target socket can force-disconnect and re-pin. WS handlers also re-check DB authority before accepting session/runtime/inbox frames, so a missed volatile notify does not let an old local binding keep acting.
- Reserved internal metadata keys (`runtimeSwitch`, `runtimeSession`) at the shared schema/service boundary. User create/PATCH rejects those keys, metadata replacement preserves internal keys, and API responses strip internal runtime metadata.

Validation:
- `pnpm typecheck`
- `pnpm exec biome check .` (exit 0; only existing unrelated warnings/info)
- `git diff --check`
- `pnpm --filter @first-tree/server test -- src/__tests__/agent-bind-authorization.test.ts src/__tests__/agent-runtime-switch.test.ts src/__tests__/agent-pinned-notification.test.ts src/__tests__/admin-agents.test.ts` (script ran the server suite: 177 files / 1676 tests passed)

baixiaohang (changes requested):
## Review by codex-assistant - PR #1421

- Rationale: the fixes move HTTP authority into durable DB state, but runtime switching still has recovery gaps that can leave an agent active in the database while no live client slot is able or prompted to run it.

**Risk level**: A
- Path baseline: touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: changes server agent/session/runtime binding flows, internal DB metadata semantics, and Web runtime switching.

**PR Summary**
- Problem: operators need to move an existing managed agent to a different computer/runtime without leaving the old local runtime able to continue acting as that agent.
- Approach: add runtime-session-bound HTTP headers, a server `switch-runtime` claim/recovery flow, session eviction, CLI local runtime convergence, cross-instance route fanout, and Web controls for switching and recovery.
- Impacted modules: `apps/cli/src/core/client-runtime.ts`, `packages/client/src/{client-connection,sdk,runtime}`, `packages/server/src/{api,services,middleware}`, `packages/shared/src/schemas`, and `packages/web/src/pages/agent-detail`.

**Review findings**
❌ 1. The pre-commit rollback path is not atomic after it revokes/disconnects the old runtime. `switchAgentRuntime` revokes the durable runtime-session token, sends `agent:force_disconnect`, and marks presence offline before the DB transaction that commits the new client/provider and retags config (`packages/server/src/services/agent-runtime-switch.ts:312`, `packages/server/src/services/agent-runtime-switch.ts:316`, `packages/server/src/services/agent-runtime-switch.ts:345`). If that transaction returns false or throws, the catch only removes `metadata.runtimeSwitch` and reactivates the old DB route (`packages/server/src/services/agent-runtime-switch.ts:149`, `packages/server/src/services/agent-runtime-switch.ts:362`). The old token and forced local unbind are not rolled back; client-side `agent:force_disconnect` only deletes the bound socket state and marks the entry idle (`packages/client/src/client-connection.ts:1405`, `apps/cli/src/core/client-runtime.ts:241`), and there is no `agent:pinned`/rebind notification on this abort path. Result: an agent can be `active` again on the old client/provider in the DB, but the old runtime is stopped or has a stale HTTP token and `/switch-runtime/recover` has no claim left to repair. Please move irreversible side effects after the durable commit, or make abort after those side effects explicitly re-pin/rebind the old route and mint a fresh runtime session. [R4]

❌ 2. Live convergence for the target client still depends on a single volatile route-change NOTIFY. The route handler records the switch and then emits `notifyAgentRouteChange` (`packages/server/src/api/agents.ts:139`), whose implementation is fire-and-forget PG NOTIFY (`packages/server/src/services/notifier.ts:309`). The listener is what sends `agent:force_disconnect` and `agent:pinned` to whichever server instance owns the sockets (`packages/server/src/api/agent/ws-client.ts:271`, `packages/server/src/api/agent/ws-client.ts:290`). If that NOTIFY is missed by the target client's instance, the DB fallback only drops the old server-side binding on the next frame (`packages/server/src/api/agent/ws-client.ts:372`); the client sees a generic `error` frame, which only emits an error event and does not fetch/rebind (`packages/client/src/client-connection.ts:1580`). This is especially visible for same-client provider switches, where the immediate local `agent:pinned` fallback is intentionally skipped (`packages/server/src/api/agents.ts:153`). A missed notify leaves the durable route switched but the online target daemon not restarted on the new provider until a reconnect. Please add a durable convergence path (for example heartbeat/registration reconciliation that pushes missing pinned agents or a client-side route-mismatch repair on server rejection), not just a best-effort NOTIFY. [R1/R5]

⚠️ 3. Several source-truth comments/docs still assert lifetime immutability even though this PR introduces runtime switching. Examples: the agents schema says there is no move/re-bind path and to provision a new agent (`packages/server/src/db/schema/agents.ts:48`), `updateAgent` returns the same guidance on rejected PATCH client moves (`packages/server/src/services/agent.ts:1039`), config service comments say provider can never switch (`packages/server/src/services/config-service.ts:198`), and CLI docs still say the client id is immutable once set (`docs/cli-reference.md:248`). Please update these to distinguish “PATCH/bind is first-set only” from the new managed `switch-runtime` flow, otherwise the old invariant remains as competing source truth. [R2/R5]

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed this review in `f233ca6c`.

Changes by finding:
- R4 pre-commit rollback: moved irreversible old-runtime side effects (`runtimeSession` revoke, `agent:force_disconnect`, presence offline) until after the durable commit succeeds. A pre-commit/claimed abort now leaves the old DB route active again with the old runtime session still valid. The WS DB-route guard also preserves the old local binding during a claimed switch instead of dropping it while frames are temporarily rejected.
- R4 committed recovery: added a shared committed-route detach step and call it from both the normal switch path and forward recovery, so a process crash after DB commit but before side effects can still revoke/disconnect the old runtime during `/switch-runtime/recover`.
- R1/R5 missed NOTIFY convergence: added heartbeat reconciliation for active agents pinned to the connected client. If a target daemon misses the volatile route-change NOTIFY, heartbeat compares DB authority with local bound runtime state and replays `agent:pinned` when the slot is missing or the provider differs. Registration backfill now uses the same reconciliation path.
- R2/R5 stale source truth: updated schema/service comments, PATCH error text, CLI help, and `docs/cli-reference.md` to distinguish generic first-set bind/PATCH from managed runtime switch.

Validation:
- `pnpm typecheck`
- `pnpm exec biome check .` (exit 0; only existing unrelated warnings/info)
- `git diff --check`
- `pnpm --filter @first-tree/server test -- src/__tests__/agent-runtime-switch.test.ts src/__tests__/agent-pinned-notification.test.ts src/__tests__/agent-client-immutable.test.ts` (script ran the server suite: 177 files / 1677 tests passed)

baixiaohang (changes requested):
## Review by codex-assistant - PR #1421

- Rationale: the latest fixes close the durable authority and missed-NOTIFY convergence gaps, but the runtime switch still does not reliably enforce the promised session eviction/local data loss boundary for same-client provider switches.

**Risk level**: A
- Path baseline: touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: changes server agent/session/runtime binding flows, internal DB metadata semantics, and Web runtime switching.

**PR Summary**
- Problem: operators need to move an existing managed agent to a different computer/runtime without leaving the old local runtime able to keep acting as that agent.
- Approach: add runtime-session-bound HTTP headers, a server `switch-runtime` claim/recovery flow, session eviction, CLI local runtime convergence, cross-instance route fanout plus heartbeat reconciliation, and Web controls for switching and recovery.
- Impacted modules: `apps/cli/src/core/client-runtime.ts`, `packages/client/src/{client-connection,sdk,runtime}`, `packages/server/src/{api,services,middleware}`, `packages/shared/src/schemas`, and `packages/web/src/pages/agent-detail`.

**Review findings**
❌ 1. Same-client provider switches can resurrect the sessions that `switch-runtime` just evicted. The server archives every current `(agent, chat)` row during the switch, but its only client-side cleanup is the best-effort `session:terminate` loop after the route change (`packages/server/src/api/agents.ts:216`). By then the committed switch path has already detached the old local binding (`packages/server/src/services/agent-runtime-switch.ts:373`, `packages/server/src/services/connection-manager.ts:51`), and `sendToAgent` silently returns false when there is no current agent mapping (`packages/server/src/services/connection-manager.ts:201`). That is the normal same-client provider-switch window because immediate pinned delivery is skipped when `agent.clientId === oldClientId` (`packages/server/src/api/agents.ts:153`), so the terminate frame is often lost. When the later `agent:pinned` replay reaches the daemon, the CLI stops the old slot and starts a new one (`apps/cli/src/core/client-runtime.ts:604`, `apps/cli/src/core/client-runtime.ts:628`), but `AgentSlot.stop()` calls `SessionManager.shutdown()` (`packages/client/src/runtime/agent-slot.ts:459`), and shutdown persists the old session registry before clearing memory (`packages/client/src/runtime/session-manager.ts:738`, `packages/client/src/runtime/session-manager.ts:747`). The new slot then loads those persisted entries back as `evictedMappings` (`packages/client/src/runtime/session-manager.ts:2232`) and resumes them on the next message (`packages/client/src/runtime/session-manager.ts:1091`, `packages/client/src/runtime/session-manager.ts:1133`). Result: the server briefly reports the sessions evicted, but the local runtime can bring the old handler session/context back after the switch, violating the destructive-confirmation/session-eviction contract. Please make runtime switch clear or invalidate the local registry/session mappings durably (for example a switch generation/provider discriminator in the registry, a reliable pre-rebind cleanup command, or a stop mode that purges persisted sessions before the new slot starts) and cover the same-client provider-switch path with a client-side assertion. [R4/R5]

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the latest same-client provider-switch session resurrection finding in `f443a5a6`.

Changes:
- Added a destructive shutdown mode for runtime-switch stops: `SessionManager.shutdown(..., { clearPersistedRegistry: true, reportSuspendedSessions: false })` clears both live sessions and `evictedMappings` before the synchronous registry flush, so the next slot cannot reload old resume mappings from disk.
- `ClientRuntime.reconfigurePinnedAgent` now uses that destructive stop mode for supported and unsupported provider switches.
- `AgentSlot.stop` now still shuts down sessions even if `unbindAgent` fails, then rethrows the first error, so local purge is not skipped by an unbind failure.
- Added client-side assertions for same-client `agent:pinned` reconfigure, persisted registry clearing/no reload, and unbind-failure cleanup.

Local checks before push:
- `pnpm --filter @first-tree/client test -- src/__tests__/session-manager-edge-coverage.test.ts src/__tests__/agent-slot.test.ts` → 113 files / 1192 tests passed
- `pnpm --filter first-tree-dev test -- src/__tests__/client-runtime-context-tree.test.ts` → 69 files / 583 tests passed
- `pnpm typecheck` → passed
- `pnpm exec biome check .` → exit 0, only existing warning/info
- `git diff --check` → passed

baixiaohang (changes requested):
## Review by codex-assistant - PR #1421

- Rationale: `f443a5a6` adds the right destructive session-registry shutdown primitive, but it is wired only to the later `agent:pinned` reconfigure path, not to the `agent:force_disconnect` path that runtime switching normally sends first.

**Risk level**: A
- Path baseline: touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: changes server agent/session/runtime binding flows, internal DB metadata semantics, and Web runtime switching.

**PR Summary**
- Problem: operators need to move an existing managed agent to a different computer/runtime without leaving old local runtime state able to continue acting as that agent.
- Approach: add runtime-session-bound HTTP headers, a server `switch-runtime` claim/recovery flow, session eviction, CLI local runtime convergence, cross-instance route fanout plus heartbeat reconciliation, and Web controls for switching and recovery.
- Impacted modules: `apps/cli/src/core/client-runtime.ts`, `packages/client/src/{client-connection,sdk,runtime}`, `packages/server/src/{api,services,middleware}`, `packages/shared/src/schemas`, and `packages/web/src/pages/agent-detail`.

**Review findings**
❌ 1. The destructive purge does not run on the first stop caused by runtime switching. The committed switch path still calls `forceDisconnect(..., "agent_runtime_switch", oldClientId)` before any target `agent:pinned` replay (`packages/server/src/services/agent-runtime-switch.ts:201`, `packages/server/src/services/agent-runtime-switch.ts:373`). The client turns that frame into `agent:unbound` with the same reason (`packages/client/src/client-connection.ts:1405`), but `AgentSlot` handles every unbound reason with a plain `this.stop()` (`packages/client/src/runtime/agent-slot.ts:211`). Because `stop()` coalesces concurrent calls by returning the already-started `this.stopping` promise (`packages/client/src/runtime/agent-slot.ts:434`), a later `agent:pinned` reconfigure cannot upgrade that in-flight stop to the new destructive options (`apps/cli/src/core/client-runtime.ts:584`). In the common same-client provider switch ordering, the default stop therefore runs `SessionManager.shutdown()` without `clearPersistedRegistry`, so it reports/persists the old local session mappings before clearing memory (`packages/client/src/runtime/session-manager.ts:749`, `packages/client/src/runtime/session-manager.ts:767`). The subsequent pinned reconfigure either joins that non-destructive stop or runs after `sessionManager` is already null, leaving the old registry on disk for the new slot to reload. Please make the `agent_runtime_switch` force-disconnect path destructive as well (or make `AgentSlot.stop` upgrade/merge shutdown options for an in-flight stop) and add a client-side assertion for the actual sequence: `agent:force_disconnect(reason=agent_runtime_switch)` followed by same-agent `agent:pinned`. [R4/R5]

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the latest force-disconnect ordering gap in `0c11fccb`.

Changes:
- `AgentSlot` now treats `agent:unbound` with reason `agent_runtime_switch` as a destructive runtime-switch stop, using the same `clearPersistedRegistry: true` / `reportSuspendedSessions: false` shutdown options as the later pinned reconfigure path.
- This means the first stop triggered by server `agent:force_disconnect(reason=agent_runtime_switch)` purges the local session registry even if the same-agent `agent:pinned` replay arrives later and only joins the in-flight stop.
- Added a client-side regression test for that exact ordering: runtime-switch unbound starts a pending stop, pinned/reconfigure stop joins it, and `SessionManager.shutdown` runs once with destructive purge options.

Local checks before push:
- `pnpm --filter @first-tree/client test -- src/__tests__/agent-slot.test.ts src/__tests__/session-manager-edge-coverage.test.ts src/__tests__/client-connection-websocket-coverage.test.ts` → 113 files / 1193 tests passed
- `pnpm --filter first-tree-dev test -- src/__tests__/client-runtime-context-tree.test.ts` → 69 files / 583 tests passed
- `pnpm typecheck` → passed
- `pnpm exec biome check .` → exit 0, only existing warning/info
- `git diff --check` → passed

baixiaohang (approved):
## Review by codex-assistant - PR #1421

**Recommendation**: approve
- Rationale: The latest `0c11fccb` change closes the force-disconnect ordering gap by making the first runtime-switch stop destructive, and the remaining runtime-session authority, switch/recovery, convergence, and UI boundaries look coherent.

**Risk level**: A
- Path baseline: touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: changes server agent/session/runtime binding flows, internal DB metadata semantics, and Web runtime switching.

**PR Summary**
- Problem: operators need to move an existing managed agent to a different computer/runtime without leaving old local runtime state able to continue acting as that agent.
- Approach: add runtime-session-bound HTTP headers, a server `switch-runtime` claim/recovery flow, session eviction, CLI local runtime convergence, cross-instance route fanout plus heartbeat reconciliation, and Web controls for switching and recovery.
- Impacted modules: `apps/cli/src/core/client-runtime.ts`, `packages/client/src/{client-connection,sdk,runtime}`, `packages/server/src/{api,services,middleware}`, `packages/shared/src/schemas`, and `packages/web/src/pages/agent-detail`.

**Review findings**
✅ 1. `0c11fccb` addresses the latest blocker: `agent:unbound(reason=agent_runtime_switch)` now uses the same destructive shutdown options as pinned reconfigure, so the actual `force_disconnect -> pinned` ordering purges local session registry state before the new slot can reload it.
✅ 2. Runtime-session HTTP authority is DB-backed and tied to the current agent route; stale tokens are rejected after switch/disconnect/suspend.
✅ 3. Runtime-switch commit/recovery now separates pre-commit rollback from committed detach/archive/reactivate behavior, with claim guards around session archive.
✅ 4. Route-change fanout has heartbeat/registration reconciliation as the durable fallback for missed volatile NOTIFYs.

**Action taken**
- Approved.

PR #1428, opened 2026-07-03
fix: resolve workspace sandbox cli from controlled path

baixiaohang:
## Summary

- Treat `FIRST_TREE_CLI_BIN_DIR` as an optional workspace-only override instead of a required default deployment setting.
- When the override is absent, resolve the channel CLI from the controlled sandbox tool path (`/usr/local/bin`, `/usr/bin`, `/bin`) and build the sandbox PATH from that allowlist rather than inheriting host PATH.
- Keep explicit non-standard CLI directories supported by validating the channel binary and adding only that directory as an extra read-only bind.
- Update workspace sandbox tests for standard-path discovery, missing CLI errors, explicit override precedence, and read-only bind behavior.

## Verification

- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-thread-options.test.ts src/__tests__/codex-handler-engine.test.ts`
- `pnpm typecheck`
- `pnpm check` (exits 0 with existing warnings/info)
- `git diff --check`

### DIFF
--- a/packages/client/src/__tests__/workspace-sandbox.test.ts
+++ b/packages/client/src/__tests__/workspace-sandbox.test.ts
@@ -9,7 +9,7 @@ import {
   writeFileSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { delimiter, join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
   assertPathInsideWorkspace,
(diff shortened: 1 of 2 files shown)


yuezengwu:
Post-merge review note: I reviewed this after the PR had already merged at 2026-07-03 11:00:47Z, so I did not submit an approval review.

No blockers found. The implementation moves workspace-only First Tree CLI discovery to the controlled standard tool dirs, preserves explicit `FIRST_TREE_CLI_BIN_DIR` precedence, scrubs inherited host `PATH`, and only adds an extra read-only bind for explicit non-system CLI dirs. I did not see DB/schema or core data-structure changes.

Verification I ran:
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-thread-options.test.ts src/__tests__/codex-handler-engine.test.ts`
- `git diff --check 208ae1fce5c69109d4af7e5029dbb3d8d6a2b7f1..269377f0b273d16519f71a4ad3f377b99363f16e`

PR #1429, opened 2026-07-03
feat(web): expandable full narration on the compose status bar

Gandy2025:
## What

The compose status bar's one-line goal (`turnText`) can now be **expanded to its full multi-line form**. A dedicated unfold control (`⇕`) on the lead row opens a **floating card** — absolutely positioned, out of flow — that shows the agent's whole current narration without reflowing the conversation. It dismisses on `Escape`, an outside press, or the lead ceasing to be a working agent that still has a full narration.

Only shows up when there is genuinely more to read: the control appears **only** when the narration is longer than the 120-char one-line goal or has line breaks the one-liner flattened.

## Why

The rail truncates the narration to one line (120 chars, whitespace-collapsed). When an agent's goal is long or multi-line, that detail was unreachable from the composer. This surfaces it on demand without turning the always-on, stable status bar into something that jumps or reflows the stream.

## How

**Shared** (`schemas/me-chat.ts`) — `LiveActivity` gains an optional, newline-preserving `turnTextFull` (`ASSISTANT_TEXT_FULL_MAX = 2000`). Present **only** when it carries strictly more than the one-line `turnText`, so a client can treat "present" as "there is more to expand".

**Server** (`services/agent-chat-status.ts`) — new `previewAssistantTextFull` (preserves `\n`, collapses intra-line whitespace, caps at 2000 with an explicit `…`); `withTurnNarration` attaches `turnTextFull` only when it differs from the one-liner. The full text already lives in `session_events`, so **no client change and no DB migration** — only the raw-prefix `LEFT()` cap in the existing LATERAL seek is widened. Rides only the per-agent `/agent-status` (`withTurnText`) path; the chat-list projection is unaffected.

**Web** (`components/chat/compose-status-bar.tsx`) — the unfold control + a floating `<section>` card that reuses the rail's existing `activityAction` for its header means. Open-state is keyed by agentId so it auto-hides on a lead switch and never silently re-opens; the trigger is excluded from the outside-press close.

## Verification

- `pnpm typecheck` (all packages) + `pnpm biome check` + `lint:tokens` — clean.
- Server unit tests for `previewAssistantTextFull` + `withTurnNarration` (`turnTextFull` present/absent, newline preservation, 2000-char cap). Full server suite green.
- Full web suite green.
- DEV `/preview/compose-status-bar` gallery, both themes: the control appears only on the long-multi-line variant; the card opens/closes and renders the full narration. (A preview variant was added.)
- Two independent code reviews (correctness + UI/design); their findings folded in (stale-open guard, `activityAction` reuse over a bespoke duplicate, `<section>` over a half-implemented `role="dialog"`, a distinct unfold glyph vs. the `+N` chevron).

### DIFF
--- a/packages/server/src/__tests__/agent-chat-status.test.ts
+++ b/packages/server/src/__tests__/agent-chat-status.test.ts
@@ -12,6 +12,7 @@ import {
   computeWorking,
   getChatAgentStatuses,
   isRuntimeFresh,
+  previewAssistantTextFull,
   resolveAgentChatStatuses,
   withTurnNarration,
 } from "../services/agent-chat-status.js";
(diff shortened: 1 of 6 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1429

**Recommendation**: approve
- Rationale: The change keeps `/agent-status` as the single server-side projection, adds `turnTextFull` as optional display-only metadata, and does not alter working/busy derivation or the chat-list projection.

**Risk level**: B-high
- Path baseline: touches `packages/server/**` and `packages/server/src/services/agent-chat-status.ts`.

**PR Summary**
- Problem: Users could only see the compose status bar's 120-character one-line narration, so longer or multi-line agent progress was unreachable from the composer.
- Approach: The per-agent `/agent-status` response now carries an optional newline-preserving `turnTextFull` only when there is genuinely more to expand; the Web compose status bar uses that field to open an out-of-flow floating card while keeping the always-visible rail stable.
- Impacted modules: `packages/server/src/services/agent-chat-status.ts`, `packages/shared/src/schemas/me-chat.ts`, `packages/web/src/components/chat/compose-status-bar.tsx`.

**Review findings**
✅ 1. `turnTextFull` stays display-only and optional; `busyAgentIds` / working status remain derived through the existing server status producer.
✅ 2. The chat-list projection remains on `withTurnText: false`, so the wider narration payload is limited to the per-agent `/agent-status` surface that needs it.
✅ 3. The card state is keyed by lead agent and clears on lead/text availability changes, Escape, and outside press, avoiding stale re-open behavior.
⚠️ 4. Note: this expands the latest `assistant_text` segment, not a concatenated full turn. That matches the existing `turnText` and WorkingTurn semantics, so I do not consider it blocking.

**Action taken**
- Approved.

yuezengwu (approved):
Reviewed the server status projection, shared LiveActivity schema, and compose status bar UI. I did not find blocking issues. The new turnTextFull field stays optional/display-only on the per-agent status path, the chat-list projection remains withTurnText false, and the floating card state/dismissal behavior is scoped to the current working lead.

baixiaohang (approved):
Recommendation: approve.

Rationale: I did not find blocking issues. The new `turnTextFull` field is optional and display-only on the per-agent status path, reuses the same current-turn `session_events.payload.text` source as `turnText`, and leaves the chat-list projection unchanged.

baixiaohang (approved):
Approved per baixiaohang's authorization.

PR #1430, opened 2026-07-03
Fix provider failure terminal spam

baixiaohang:
## Summary

- require a server-confirmed durable `session:event` before ACKing terminal session start/resume failures, so permanent provider failures do not redeliver the same inbox entry indefinitely
- add confirmed `session:event` accepted/rejected frames across shared schemas, client connection, and server WS handling
- preserve inbox prefix custody while terminal teardown is settling, including recovery fallback stale-queue handling and `owned / settled / lost` route ownership semantics
- propagate HTTP `Retry-After` through `SdkError` so provider retry waits can honor server capacity delays
- suppress stale terminal provider failure text in the Web compose bar while the agent is visibly working, without clearing session-scoped status reasons from server status projection

## Verification

- `pnpm --filter @first-tree/server exec vitest run --passWithNoTests --maxWorkers=2 src/__tests__/ws-session-event.test.ts src/__tests__/ws-server-welcome.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --passWithNoTests --maxWorkers=2 src/__tests__/agent-chat-status.test.ts src/__tests__/me-chat-activity.test.ts src/__tests__/me-chats-http-e2e.test.ts`
- `pnpm --filter @first-tree/web exec vitest run --passWithNoTests --maxWorkers=2 src/components/chat/__tests__/compose-status-bar.test.ts`
- `pnpm --filter @first-tree/client exec vitest run --passWithNoTests --maxWorkers=2 src/__tests__/session-start-error-signaling.test.ts`
- `pnpm --filter @first-tree/client exec vitest run --passWithNoTests --maxWorkers=2 src/__tests__/session-manager.test.ts src/__tests__/session-manager-edge-coverage.test.ts src/__tests__/session-manager-retry.test.ts`
- `pnpm exec biome check <affected files>`
- `git diff --check`

## Review

First Tree chat review by `codex-reviewer` passed on 2026-07-03 after R4/R5 follow-ups.


yuezengwu (changes requested):
Implementation target/core changes reviewed: this PR makes terminal session start/resume provider failures settle only after a durable `session:event`, adds confirmed `session:event` accepted/rejected frames, preserves inbox custody while terminal teardown settles, carries `Retry-After`, and suppresses stale-looking terminal provider text in the Web compose bar while the agent is visibly working.

Blocking finding:

- `packages/server/src/api/agent/ws-client.ts:1458` gates `session:event:accepted` on `contextTreeIoService.recordFromSessionEvent(...)`, which is documented and handled as a best-effort/non-fatal side effect. The client only waits 3s for confirmation. If `appendEvent` has already persisted the durable event but Context Tree IO is slow, the client times out, treats the confirmation as failed, and falls back to recovery instead of ACKing the terminal provider failure. That reopens the redelivery path this PR is trying to close even though the required durable evidence already exists. Please send `session:event:accepted` immediately after `sessionEventService.appendEvent(...)` succeeds, and run Context Tree IO / notifier work after that or fire-and-forget it. `session:event:rejected` should stay tied to actual event persistence failure.

I did not find a DB/schema migration in this PR. It does change the shared WS protocol and core client/server runtime ACK semantics, so this needs the protocol/semantics check above before merge.

PR #1432, opened 2026-07-04
feat: docloop M2 — web Docs section (library + reading view with anchored review comments)

liuchao-001:
## Summary

M2 of **docloop** (stacked on #1431 — base is `feat/docloop-m1`, retarget to `main` after M1 merges). Design doc: [first-tree-context#675](https://github.com/agent-team-foundation/first-tree-context/pull/675).

### IA decision (per `product/surfaces/surface-model.md`)
No fifth top-level tab: the document library is the **raw layer of shared memory**, so it lives inside the **Context** surface — a Tree | Documents segmented switcher. The switcher (and routes) only appear when the deployment has `FIRST_TREE_DOCS_ENABLED` on, carried to the shell via a new `features.docs` field on `/me` (presentation-only; server routes stay the enforcement point).

### What's in
- **`/context/docs`** — library list: server-side status filter, client-side title/slug/project search (server FTS is M3), open-comment counts, agent/human author bylines.
- **`/context/docs/:slug`** — reading view:
  - rendered markdown (existing `<Markdown>` component);
  - **text-selection commenting**: selection → `buildDocAnchor` (new in `@first-tree/shared`) turns the rendered selection into a source-based TextQuoteSelector anchor via whitespace-normalized matching + context-scored disambiguation; selections spanning markdown syntax degrade **visibly** to a document-level comment quoting the text;
  - threaded comment sidebar: reply, resolve/reopen, quote-click scroll-to-text, resolved section collapsed, cross-version badges;
  - version switcher (older-version banner; comments target the viewed version), status select + **Approve** shortcut.
- **CLI**: `doc publish` output now includes the shareable `/context/docs/<slug>` URL for pasting into chat.
- **shared**: `buildDocAnchor` / `locateDocAnchor` — `locateDocAnchor` is deliberately the same call the M3 re-anchor pass will use server-side.

### Tests
- 9 shared anchor unit tests (round-trip, whitespace reflow, repeated-text disambiguation, deletion → null).
- 7 happy-dom page tests (list render / search filter / empty state; doc render with threads; resolve; approve; unknown slug). Note: the repo's web test convention is happy-dom component tests — there is no Playwright infra — so the e2e-style coverage promised in the proposal lands as these page tests plus M1's real-HTTP API coverage.
- Full suites green: server 1707/1707 · shared 504/504 · web 1088/1088 · `pnpm check`/`typecheck` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1431, opened 2026-07-04
feat: docloop document review M1 — org document library API + doc CLI namespace

liuchao-001:
## Summary

M1 of **docloop** — the shared design-doc review feature (agents publish AI-authored markdown design docs via CLI; humans review with anchored comments — web UI lands in M2). Design doc: [first-tree-context#675](https://github.com/agent-team-foundation/first-tree-context/pull/675) (v4, approved direction: build as a first-tree feature, keep the module extractable into a standalone open-source product).

### Server
- **Table group** `doc_documents` / `doc_versions` / `doc_comments` — self-contained by design: org FK + opaque author-id snapshots only, no joins into hub business tables (extractability constraint §6.1 of the proposal).
- **Domain service** (`services/document.ts`): idempotent publish by org-unique slug (first publish creates v1, later publishes append the next version under a row lock; `ifChanged` skips identical content), TextQuoteSelector-style comment anchors (`exact`/`prefix`/`suffix` against the markdown source), one-level threads whose status lives on the top-level comment.
- **Identity adapter seam** (`services/doc-author.ts`): the only place the feature reads a hub identity table. Resolves any agents-table uuid to a `Principal { kind: human|agent, id, name }` — **agents author under their own agent name**; a human driving the CLI through their identity-mirror agent is still recorded as a human.
- **Three route surfaces** per `docs/development/http-path-conventions.md`:
  - Class B `/orgs/:orgId/documents` (list, publish)
  - Class C `/documents/:docId` + `/document-comments/:commentId` (read, status, comments — non-members get 404, no existence leak)
  - Class D `/agent/documents…` (the CLI's full loop)
- **Feature flag** `FIRST_TREE_DOCS_ENABLED` (default **off**; staging-first rollout — routes are not mounted while off).

### CLI
- New `doc` namespace: `publish / get / list / comments / comment / reply / resolve / status`, thin over new SDK methods; slug defaults to the slugified filename, title to the first markdown heading; JSON-first output, scriptable error codes. `docs/cli-reference.md` updated.

### Tests
- 13 server integration tests: all three surfaces end-to-end (publish → comment → reply → resolve → status), versioning + `ifChanged`, filters + pagination, threading rules (reply-to-reply 400, reply resolve 400, thread-scoped status filter + open counts), **multi-org isolation** (Class C and Class D 404 across orgs), flag-off boundary.
- CLI unit tests: pure helpers (`slugFromFilename`, `titleFromMarkdown`) + namespace wiring.

### Test evidence
[...]

PR #1434, opened 2026-07-05
feat: docloop — shared design-doc review (API + CLI + web), integration branch

liuchao-001:
## Summary

The complete **docloop** feature — integration branch carrying the three reviewed milestones, squashed per milestone:

- **M1** (#1431): `doc_` table group, org/resource/agent-self REST surfaces per http-path-conventions, idempotent slug publish, TextQuoteSelector comment anchors, `doc` CLI namespace — reviewed HEALTHY (two rounds + independent black-box E2E).
- **M2** (#1432): web Docs section inside the Context surface — library list, reading view with text-selection commenting, threads/resolve, version switcher, approve; `/me` feature flag; shared anchor build/locate — reviewed HEALTHY (real-browser E2E + re-verification).
- **M3** (#1433): read-time re-anchoring (`outdated`), `doc import/export`, `comments --watch` — reviewed HEALTHY (real-CLI E2E), review polish adopted.

Everything is feature-flagged behind `FIRST_TREE_DOCS_ENABLED` (default **off** — merging this is inert until an operator enables the flag).

Design: [first-tree-context#675](https://github.com/agent-team-foundation/first-tree-context/pull/675) · route map: [first-tree-context#676](https://github.com/agent-team-foundation/first-tree-context/pull/676).

## Review status

All three milestone PRs carry the team code-reviewer agent's full sign-off (independent E2E each round: real server boot + migrations, real browser for the annotation loop, real CLI for import/export/watch) and green CI. Per the integration-branch plan (liuchao-001), the milestones merged into `feat/docloop` without individual approvals; **this PR is the single human-approval gate into `main`** — @286ljb @baixiaohang, one approval here covers the chain.

## Rollout

After merge: enable `FIRST_TREE_DOCS_ENABLED=1` on **staging**, dogfood by importing the real `raw-context/proposals/` set, then decide prod + the proposals-directory switchover (owner: liuchao-001).


liuchao-001:
Both Codex P2s addressed via #1435 (merged into this branch): bylines now snapshot `display_name` with slug fallback, and resolve/reopen bumps `doc_documents.updated_at` so handled feedback reorders the library. Server documents suite 16/16.

yuezengwu (changes requested):
Thanks for the integration branch. The overall shape is clear: docloop adds the `doc_` table group, org/resource/agent document APIs, the `first-tree doc` CLI namespace, shared text anchors, and the Context → Documents web surface behind `FIRST_TREE_DOCS_ENABLED`.

I found one blocking data-correctness issue before this can merge:

- `packages/server/src/services/document.ts:290-318` uses only `doc_documents.updated_at` as the keyset cursor (`updated_at < cursor`) and orders only by `updated_at DESC`. If more than `limit` documents share the same `updated_at` at a page boundary, every remaining row with that same timestamp is excluded from the next page. This is especially risky because `apps/cli/src/commands/doc/export.ts` depends on this pagination loop for the data-out guarantee; it can silently export an incomplete library. Please make the cursor stable with a tie-breaker, e.g. order by `(updated_at DESC, id DESC)` and encode both values in `nextCursor`, with a regression test that seeds multiple docs with identical `updated_at` and `limit=1`/small limit.

I verified the two previous Codex P2s are addressed in head (`displayName` snapshots for bylines and document `updatedAt` bump on resolve/reopen). CI is green on the PR. Local focused tests did not run in my review worktree because dependencies are not installed there (`vitest: command not found`), so I relied on CI plus code inspection.

liuchao-001:
@yuezengwu Thanks — the pagination issue was real and is fixed via #1438 (merged into this branch):

- Cursor is now `<updatedAt ISO>|<id>` with `(updated_at DESC, id DESC)` ordering and a row-wise tuple page condition, so same-timestamp documents (the bulk-import shape) paginate stably.
- One subtlety beyond the tie-breaker: `updated_at` is **millisecond-truncated database-side** in the comparison — Postgres stores microseconds while the ISO cursor carries milliseconds, so without truncation the equality leg of the tuple comparison would never bind and rows inside the cursor's millisecond would still be skipped.
- Regression test per your suggestion: three documents seeded with one identical `updated_at`, walked with `limit=1` — every document surfaces exactly once.

Ready for your re-review.

yuezengwu (changes requested):
Thanks for the quick fix. The original timestamp-tie issue is addressed for rows with exactly identical `updated_at`, but the new keyset is still not quite stable because the predicate and ordering use different keys.

Blocking follow-up:

- `packages/server/src/services/document.ts:318` compares `(date_trunc('milliseconds', updated_at), id)` against the cursor, but `packages/server/src/services/document.ts:334` still orders by raw `updated_at DESC, id DESC`. Since Postgres stores microseconds and `Date#toISOString()` only round-trips milliseconds, rows in the same millisecond but with different microseconds are ordered by raw microseconds on page N, then paged by truncated millisecond + id on page N+1. Example with `limit=1`: row A has `...123900` and id `a`, row B has `...123800` and id `z`; A sorts first, cursor becomes `...123Z|a`, and the next-page predicate excludes B because the truncated timestamp is equal and `z > a`. Please make the ORDER BY use the same expression as the cursor predicate, e.g. `date_trunc('milliseconds', updated_at) DESC, id DESC`, or encode a cursor that preserves the full database timestamp precision. Add a regression where rows are within the same millisecond but have different microseconds and ids that do not happen to match timestamp order.

liuchao-001:
@yuezengwu Second issue fixed via #1439 (merged into this branch): ORDER BY now uses the cursor predicate's own expression — `date_trunc('milliseconds', updated_at) DESC, id DESC` — so page order and page predicate can no longer disagree inside a millisecond.

Regression test follows your spec exactly: two documents 100µs apart within one millisecond, ids contradicting timestamp order, walked with `limit=1`. I additionally **verified the test fails against the previous ordering** (temporarily reverted the ORDER BY locally: the low-µs/high-id row gets dropped, reproducing your example) and passes with the fix — suite 18/18.

Ready for your re-review; your approval satisfies the main gate.

yuezengwu (approved):
Approved after re-review of #1439.

The blocking pagination issue is now addressed: `listDocuments` uses the same millisecond-truncated `(updated_at, id)` key for both the page predicate and ordering, and the regression suite now covers both identical timestamps and the same-millisecond/different-microsecond case with id order contradicting raw timestamp order. That closes the `doc export` completeness risk I raised.

I re-checked the earlier fixes as well: display-name author snapshots and resolve/reopen document activity bumps remain in head. CI was still pending during this re-review, but per the agreed review posture I am not waiting on CI to approve.

PR #1437, opened 2026-07-06
fix: refresh runtime session transport after rebind

yuezengwu:
## Summary
- Refresh AgentSlot runtime SDK after agent:bound so reconnect rebinds adopt the new runtime session token.
- Propagate replacement SDK into AgentConfigCache, SessionManager, SessionContext, envCtx, and participant cache.
- Add regression coverage for rebind and lazy SDK resolution paths.

## Root cause
After a WebSocket reconnect/rebind, ClientConnection produced a new FirstTreeHubSDK with a new runtimeSessionToken, but existing AgentSlot services kept the old SDK. Agent-scoped HTTP calls then sent a stale x-agent-runtime-session, which showed up locally as Invalid agent runtime session for first-tree-staging.

## Validation
- ./node_modules/.bin/vitest run src/__tests__/agent-config-cache.test.ts src/__tests__/agent-io.test.ts src/__tests__/agent-slot.test.ts src/__tests__/session-manager.test.ts src/__tests__/session-manager-edge-coverage.test.ts --passWithNoTests
- ../../node_modules/.bin/tsc --noEmit
- ./node_modules/.bin/biome check packages/client/src/runtime/agent-config-cache.ts packages/client/src/runtime/agent-io.ts packages/client/src/runtime/session-manager.ts packages/client/src/runtime/agent-slot.ts packages/client/src/__tests__/agent-config-cache.test.ts packages/client/src/__tests__/agent-io.test.ts packages/client/src/__tests__/agent-slot.test.ts packages/client/src/__tests__/session-manager.test.ts packages/client/src/__tests__/session-manager-edge-coverage.test.ts
- PATH=/opt/homebrew/bin:$PATH corepack pnpm typecheck

## Notes
- Full corepack pnpm check is currently blocked by an existing nested Biome config under .claude/worktrees/wizardly-williams-e41b30/biome.json, unrelated to this branch.

### DIFF
--- a/apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts
@@ -1,3 +1,6 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const clientMocks = vi.hoisted(() => {
@@ -46,6 +49,7 @@ vi.mock("../core/cli-fetch.js", () => ({ cliFetch: cliFetchMock }));
 
 const originalAgentId = process.env.FIRST_TREE_AGENT_ID;
 const originalRuntimeSessionToken = process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN;
+const originalRuntimeSessionTokenFile = process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE;
 const originalServerUrl = process.env.FIRST_TREE_SERVER_URL;
 
 function restoreEnv(): void {
@@ -63,6 +67,11 @@ function restoreEnv(): void {
     delete process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN;
   } else {
     process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN = originalRuntimeSessionToken;
+  }
+  if (originalRuntimeSessionTokenFile === undefined) {
+    delete process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE;
+  } else {
+    process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE = originalRuntimeSessionTokenFile;
   }
 }
 
@@ -121,6 +130,28 @@ describe("local agent shared helpers", () => {
         runtimeSessionToken: "runtime-token-1",
       }),
     );
+  });
+
+  it("prefers the runtime session token file over a stale inherited env token", async () => {
+    const dir = mkdtempSync(join(tmpdir(), "first-tree-token-"));
+    try {
+      const tokenFile = join(dir, "runtime.token");
+      writeFileSync(tokenFile, "runtime-token-2\n", "utf8");
+      const { createSdk } = await import("../commands/_shared/local-agent.js");
+
+      process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN = "runtime-token-1";
+      process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE = tokenFile;
+      createSdk("nova");
+
+      expect(clientMocks.FirstTreeHubSDK).toHaveBeenCalledWith(
+        expect.objectContaining({
+          agentId: "agent-1",
+          runtimeSessionToken: "runtime-token-2",
+        }),
+      );
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
   });
 
   it("maps local agent resolution failures to CLI errors", async () => {
--- a/apps/cli/src/commands/_shared/local-agent.ts
+++ b/apps/cli/src/commands/_shared/local-agent.ts
@@ -1,3 +1,4 @@
+import { readFileSync } from "node:fs";
 import { join } from "node:path";
 import { FirstTreeHubSDK, SdkError } from "@first-tree/client";
 import {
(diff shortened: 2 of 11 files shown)


baixiaohang:
Code review notes from the runtime-session rebind pass:

1. **Already-started provider subprocesses still keep the old runtime token.**
   This PR refreshes the in-process SDK/cache path, which fixes the direct `agentConfigCache.refresh()` failure. However, child provider processes that were already started still have the old `FIRST_TREE_RUNTIME_SESSION_TOKEN` in their environment. If they later run `first-tree chat send` (or any agent-scoped CLI command), the CLI reads that stale env token and can still hit `403 Invalid agent runtime session`.

   Relevant paths:
   - `packages/client/src/runtime/agent-io.ts` injects `FIRST_TREE_RUNTIME_SESSION_TOKEN` into child env.
   - `apps/cli/src/commands/_shared/local-agent.ts` reads the token from env.
   - Codex app-server / Claude Code subprocesses use the env captured at process start.

   Recommendation: on rebind, either restart/recreate active provider subprocesses, or move the CLI token lookup to a runtime-updated file/IPC source rather than fixed process env. A regression test should cover: session already started -> WS rebind mints token2 -> child CLI command uses token2, not token1.

2. **`AgentConfigCache` still leaks failures from an old in-flight fetch.**
   `updateSdk()` clears `entry.inflight` for future callers, but callers already awaiting the old promise can still receive the stale-token 403. The generation guard currently handles old successful responses by not updating cache, but it does not retry when the old request fails after a rebind.

   Recommendation: if a fetch rejects after its generation became stale, retry once using the current SDK, or structure `refresh()` so waiters follow the latest generation. Add a deferred-fetch test: start refresh with sdk1, call `updateSdk(sdk2)`, reject sdk1 with `Invalid agent runtime session`, then assert the caller resolves from sdk2.

3. **`activeRuntimeChatIds` refresh can be skipped after rebind if an old refresh is in flight.**
   `adoptRuntimeTransport()` updates `this.sdk`, but `activeRuntimeChatIdsRefreshInFlight` is not cleared. If the old SDK refresh is in flight, the bind-triggered refresh returns that old promise; the old promise result is ignored because `this.sdk !== sdk`, so the post-bind `fullStateSync()` may run with the previous active set.

   Recommendation: clear or SDK-scope `activeRuntimeChatIdsRefreshInFlight` when adopting a new SDK so the bind-time refresh always uses the new transport.

Overall, the PR fixes the main in-process stale SDK path and is directionally correct, but I would not treat it as fully closing the runtime-session rebind issue until at least the in-flight config fetch window is handled. If provider subprocess CLI usage is expected during long-lived sessions, the child-env token issue should also be fixed here or tracked as an explicit follow-up.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1437

- Rationale: one rebind/config-refresh success race can let dispatch continue after a nominal refresh while the runtime config cache still holds the old payload.

**Risk level**: A
- Semantic lift: none needed; runtime session transport/rebind code is already in A scope.

**PR Summary**
- Problem: after a daemon WebSocket reconnect/rebind, long-lived agent runtime work must stop using the previous runtime-session token for agent-scoped HTTP calls.
- Approach: refresh the AgentSlot SDK on `agent:bound`, propagate the replacement transport into config/session/participant paths, add a runtime-session token file for already-started provider subprocesses, and add regression coverage for the known stale-token failure paths.
- Impacted modules: `apps/cli/src/commands/_shared/local-agent.ts`, `packages/client/src/runtime/{agent-slot,session-manager,agent-config-cache,agent-io}.ts`, and related client/CLI tests.

**Review findings**

1. Blocking: stale-generation config fetches can resolve without updating the cache.

`doFetch()` currently returns the fetched config when `generation !== sdkGeneration`, before writing it into `entries` (`packages/client/src/runtime/agent-config-cache.ts:67`). That fixes stale overwrites, but it leaves the waiter with a successful Promise while the cache itself remains on the old config. `SessionManager.dispatch()` only awaits `refreshIfNewer()` and ignores its return value (`packages/client/src/runtime/session-manager.ts:540`), so this race can proceed to route the message while later code that reads `agentConfigCache.get()` still sees the old payload.

Concrete race: message with a newer `configVersion` starts `refreshIfNewer()` on SDK1; reconnect/rebind calls `updateSdk(SDK2)`; the SDK1 request had already passed server validation and resolves successfully. The refresh await completes, but because line 67 skipped the cache write, the current turn can still build briefing/env/doc fences from the pre-refresh config. Please make stale-generation success either follow the latest SDK, as the rejection path now does, or commit the result with an explicit version guard before resolving. The regression should cover the successful old in-flight response case, not only the stale-token rejection case.

**Action taken**
- Submitted request changes.

baixiaohang (approved):
## Review by codex-assistant - PR #1437

**Recommendation**: approve
- Rationale: the rebind transport propagation now covers the SDK/cache/session/child-CLI paths, including stale in-flight config fetch success and failure races.

**Risk level**: A
- Semantic lift: none.

**PR Summary**
- Problem: after a WebSocket reconnect/rebind, already-running agent runtime work could continue using the previous runtime-session token for agent-scoped HTTP calls.
- Approach: refresh the AgentSlot SDK on `agent:bound`, propagate the replacement transport into AgentConfigCache, SessionManager, SessionContext/env, and participant resolution, and expose a stable runtime-session token file for long-lived provider subprocess CLI calls.
- Impacted modules: `apps/cli/src/commands/_shared/local-agent.ts`, `packages/client/src/runtime/{agent-slot,session-manager,agent-config-cache,agent-io}.ts`, and related tests.

**Review findings**
✅ The previous blocking stale-generation config fetch race is fixed: stale-generation success and failure both retry through the replacement SDK, with regression coverage that verifies the cache updates to the latest config.
✅ The earlier edge cases around child CLI token refresh and active-runtime-chat refresh after rebind are covered by implementation and tests.
✅ CI is green. I did not run local tests.

**Action taken**
- Approved.

PR #1438, opened 2026-07-06
fix: stable keyset pagination for document lists

liuchao-001:
Fixes the blocking data-correctness issue @yuezengwu found on #1434: the document-list cursor was `updated_at` alone, so documents sharing one timestamp (exactly the bulk-import shape) were silently dropped at page boundaries — and `doc export` depends on this loop for its completeness guarantee.

- Cursor is now `<updatedAt ISO>|<id>`; ordering is `(updated_at DESC, id DESC)`; the page condition is a row-wise tuple comparison.
- `updated_at` is **millisecond-truncated database-side** for the comparison: Postgres stores microseconds while the JS Date / ISO cursor only carries milliseconds, so comparing raw column values against the round-tripped cursor would ALSO skip rows inside the cursor's millisecond — the tie-breaker only works if the equality leg is exact.
- Regression test: three documents seeded with one identical `updated_at`, walked with `limit=1`; every document surfaces exactly once (17/17 suite green).

### DIFF
--- a/packages/server/src/__tests__/documents.test.ts
+++ b/packages/server/src/__tests__/documents.test.ts
@@ -154,6 +154,52 @@ describe("documents API", () => {
       expect(missing.statusCode).toBe(404);
     });
 
+    it("paginates stably across identical updatedAt timestamps", async () => {
+      const app = getApp();
+      const ctx = await createAdminContext(app);
+      const { docDocuments } = await import("../db/schema/index.js");
+
+      // Bulk-import shape: several documents sharing one exact timestamp. A
+      // timestamp-only cursor drops the tied rows at a page boundary.
+      const sameInstant = new Date("2026-07-05T12:00:00.123Z");
+      const tiedSlugs = Array.from({ length: 3 }, (_, i) => slug(`tied-${i}`));
+      for (const s of tiedSlugs) {
+        await app.db.insert(docDocuments).values({
+          id: uuidv7(),
+          organizationId: ctx.organizationId,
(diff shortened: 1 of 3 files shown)


yuezengwu (changes requested):
Reviewed the fix as: document-list pagination must be complete for `doc export`; this PR moves the cursor from timestamp-only to `<updatedAt ISO>|<id>`, adds an id desc tie-breaker, and truncates `updated_at` to milliseconds for cursor comparison.

Blocking finding:

- `packages/server/src/services/document.ts:334`: the predicate now uses `(date_trunc('milliseconds', updated_at), id)`, but the query still orders by raw `updated_at` before `id`. Those are different keysets whenever Postgres has two timestamps in the same JS millisecond with different microseconds. Example: row A has `2026-07-05T12:00:00.123900Z` and id `a`; row B has `2026-07-05T12:00:00.123800Z` and id `z`. Raw `updated_at DESC` returns A before B, the cursor serializes A as `2026-07-05T12:00:00.123Z|a`, and the next page predicate only accepts same-ms ids `< a`, so B is skipped. The `date_trunc` expression needs to be the first `ORDER BY` key too, or the cursor must preserve raw microsecond precision.

Please also extend the regression test with same-millisecond but different-microsecond `updated_at` values; the current identical-`Date` seed only covers exact timestamp ties.

Local verification: `pnpm --filter @first-tree/server test -- documents.test.ts` could not run in this worktree because `node_modules` is missing (`vitest: command not found`).

PR #1439, opened 2026-07-06
fix: order document lists by the cursor's own key expression

liuchao-001:
Follow-up to @yuezengwu's second review on #1434: the page predicate compared millisecond-truncated tuples but ORDER BY still used raw microsecond timestamps — inside one millisecond, page N's order and page N+1's predicate disagreed, and the adversarial shape (higher-µs row carrying the smaller id) dropped its sibling.

- ORDER BY now uses the exact predicate expression: `date_trunc('milliseconds', updated_at) DESC, id DESC`.
- Regression test per the review: two documents 100µs apart inside one millisecond, ids contradicting timestamp order, walked with `limit=1` — **verified to fail against the previous ordering** and pass with the fix (suite 18/18).

### DIFF
--- a/packages/server/src/__tests__/documents.test.ts
+++ b/packages/server/src/__tests__/documents.test.ts
@@ -200,6 +200,61 @@ describe("documents API", () => {
       expect(new Set(seen).size).toBe(tiedSlugs.length);
     });
 
+    it("paginates stably across microsecond differences within one millisecond", async () => {
+      const app = getApp();
+      const ctx = await createAdminContext(app);
+      const { docDocuments } = await import("../db/schema/index.js");
+      const { sql, eq: eqOp } = await import("drizzle-orm");
+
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed and approved.

This fixes the cursor-order mismatch by making `ORDER BY` use the same millisecond-truncated key expression as the page predicate: `date_trunc('milliseconds', updated_at) DESC, id DESC`. The new regression test covers the adversarial case from the prior review: two rows within one millisecond where microsecond order and id order disagree, walked with `limit=1`, so the old raw-timestamp ordering would drop a sibling.

Verification: `pnpm --filter @first-tree/server test -- documents.test.ts` passed locally. Note that this command matched the broader server test run in this repo: 178 test files / 1710 tests passed, including `documents.test.ts` 18/18.

PR #1441, opened 2026-07-06
fix(web): offer status-bar expand on width-truncated goals, not only >120

Gandy2025:
## Problem

The compose status bar's expand affordance (⇕, PR #1429) appeared **only** when the server attached `turnTextFull` — i.e. the narration exceeded the 120-char cap or had line breaks. But the rail truncates a goal by **width** far earlier (~50 chars). So a goal that is merely too wide for the rail showed a "…" with **no way to expand** — which misses the intent ("I see it cut off, let me expand"). Only goals >120 chars got the affordance.

## Fix

The ⇕ now also appears when the goal is **actually clipped on the rail** (`scrollWidth > clientWidth`), and the card shows `turnTextFull ?? turnText` so a width-truncated ≤120 goal still expands to its full line.

- `useIsOverflowing` (in `WorkingDetail`): measures the goal span via a `ResizeObserver` + a re-measure on text change (`useLayoutEffect`, pre-paint → no first-frame flash). The `flex-1` box doesn't resize when the text changes, so the text is an intentional re-measure trigger (narrowly `biome-ignore`d with a comment).
- The lead's overflow is reported up (`WorkingDetail → LeadDetail → RailRow → ComposeStatusBar`) to derive `canExpand = !!leadGoal && (leadGoalOverflow || !!turnTextFull)`. Reported **pre-paint** to avoid a one-frame stale glyph on a lead switch, and **reset on unmount** so a stale ⇕ never lingers when a working lead swaps to a `statusReason` row.
- Short goals that fit still show no affordance (no noise).

## Verification

- `pnpm --filter @first-tree/web typecheck` + `biome check` + `lint:tokens` — clean.
- Full web test suite green (1088).
- DEV `/preview/compose-status-bar`: the width-clipped variant now offers ⇕ and its card shows the full line; the multi-line/`turnTextFull` variant unchanged; short/fitting goals show none. (Preview variant + subtitles updated.)
- Two independent code reviews (correctness + design/simplicity), both **non-blocking**; their two low-severity staleness findings (one-frame flicker on lead switch; stale ⇕ beside a reason row) are folded into this commit.

Follow-up to #1429.

### DIFF
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -181,13 +181,18 @@ export function ComposeStatusBar({
   // The lead's full narration — present only when the server attached
   // `turnTextFull` (strictly more than the one-line goal). Only working leads
   // carry live activity.
-  const leadGoal = leadRow?.main === "working" ? leadRow.activity?.turnText : undefined;
+  // Expand only when the goal is the actually-rendered detail: LeadDetail shows a
+  // visible statusReason (waiting/retrying/…) *instead* of the goal when one
+  // exists, so the affordance must never appear beside a reason label — mirror
+  // that same predicate here rather than keying off `main === "working"` alone.
+  const leadShowsGoal = leadRow !== undefined && leadRow.main === "working" && !visibleStatusReason(leadRow);
+  const leadGoal = leadShowsGoal ? leadRow.activity?.turnText : undefined;
   // The text the card shows: the full newline-preserving narration when the
   // server sent one, else the one-line goal itself — so a goal that is merely
   // too wide for the rail can still be read in full.
-  const leadFullText = leadRow?.main === "working" ? (leadRow.activity?.turnTextFull ?? leadGoal) : undefined;
+  const leadFullText = leadShowsGoal ? (leadRow.activity?.turnTextFull ?? leadGoal) : undefined;
   // Offer expand when the goal is actually clipped on the rail OR the server
-  // sent strictly-more content (multi-line / >120 chars). Requires a goal.
+  // sent strictly-more content (multi-line / >120 chars). Requires a rendered goal.
   const canExpand = !!leadGoal && (leadGoalOverflow || !!leadRow?.activity?.turnTextFull);
   // Forget a stale open-card key so the card never silently re-opens on its own
   // once the lead can no longer expand (switched away, goal gone, or it shrank
@@ -428,8 +433,10 @@ function WorkingDetail({
   }, [goal, goalOverflowing, onGoalOverflowChange]);
   // Reset the parent's overflow flag when this WorkingDetail unmounts — e.g. the
   // still-`working` lead gains a statusReason and LeadDetail swaps to the reason
-  // branch — so a stale `⇕` never lingers beside a non-goal row.
-  useEffect(() => () => onGoalOverflowChange?.(false), [onGoalOverflowChange]);
+  // branch — so a stale `⇕` never lingers beside a non-goal row. Layout effect
+  // (not passive) so the cleanup runs synchronously in the commit that removes
+  // WorkingDetail, clearing the flag before paint — no one-frame stale button.
+  useLayoutEffect(() => () => onGoalOverflowChange?.(false), [onGoalOverflowChange]);
   if (!activity) return <span className="truncate">Working</span>;
   const action = activityAction(activity, goal !== null);
   return (
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -6,7 +6,7 @@ import {
 } from "@first-tree/shared";
 import { useQuery } from "@tanstack/react-query";
 import { Brain, ChevronDown, ChevronsUpDown, Pencil, Wrench } from "lucide-react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
 import { chatAgentStatusQueryKey, fetchChatAgentStatuses } from "../../api/agent-status.js";
 import { viewOf } from "../../lib/agent-status-view.js";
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Goal/core change reviewed: this PR extends the compose status bar's expand affordance from only "server sent `turnTextFull`" to also "the lead goal is actually clipped in the rail." The implementation measures the rendered goal span with `ResizeObserver`/`useLayoutEffect`, reports that overflow up through `WorkingDetail -> LeadDetail -> RailRow -> ComposeStatusBar`, derives `canExpand` from visual overflow or `turnTextFull`, and renders the card from `turnTextFull ?? turnText`. The preview fixture now covers the width-clipped short/full-line case.

Approved. I did not find any blocking correctness issues, and there are no DB/schema/core data-structure changes in this patch.

Non-blocking note: the `WorkingDetail` unmount reset currently uses a passive `useEffect` cleanup. If a working row with an overflowing goal switches into a visible `statusReason` branch while still carrying `activity.turnText`, the render that removes `WorkingDetail` can still see the previous `leadGoalOverflow` value and paint the expand button for one frame before the passive cleanup clears it. If you want to eliminate that last flicker, either make the cleanup layout-synchronous or have the non-working-detail branches report `false` immediately.

I attempted a targeted local test run, but this review worktree has no installed dependencies (`vitest: command not found`), so I did not rerun the suite locally.

baixiaohang (approved):
## Review by codex-assistant - PR #1441

**Recommendation**: approve
- Rationale: The main change matches the user-visible intent: a visibly width-clipped status-bar goal now gets an expand affordance, and the card falls back to the full one-line goal when the server did not send `turnTextFull`.

**Risk level**: C
- Path baseline: only `packages/web/**` component and DEV preview changes -> C.
- Semantic lift: none; this does not touch Web subscriptions, WS frame parsing, or shared data-layer contracts.

**PR summary**
- Problem: In the Workspace composer status bar, users could see a goal clipped by the rail width but had no way to expand it unless the server had also marked the narration as longer than the 120-character preview.
- Approach: The lead goal span now reports actual horizontal overflow to the status bar, and the expand card renders `turnTextFull ?? turnText`, so width-only truncation can be expanded without adding noise for short fitting goals.
- Impacted modules: `packages/web/src/components/chat/compose-status-bar.tsx`, `packages/web/src/pages/compose-status-bar-preview.tsx`.

**Review findings**
⚠️ 1. Non-blocking: `canExpand` is still derived from the working row's activity even when `LeadDetail` is rendering a visible `statusReason` instead of the goal. For a working+retrying/waiting row that still carries turn text, `turnTextFull` can keep the unfold affordance visible beside the reason label; and the overflow reset on that branch is a passive-effect cleanup, so the width-overflow path can also flash for a frame. If that edge matters, derive `canExpand` from the same "goal is actually rendered" predicate as `LeadDetail` (for example, no visible status reason) and use a layout-effect cleanup for the unmount reset. [R5 / `packages/web/src/components/chat/compose-status-bar.tsx`:191]

**Action taken**
- Approved.

Gandy2025:
Thanks both — folded in the non-blocking note from @yuezengwu and @baixiaohang (R5) in `563c13a0`:

1. **`⇕` beside a `statusReason` row** — root cause was `canExpand` keying off `main === "working"` alone, so a working lead that also carries a visible reason (waiting/retrying) with `turnTextFull` set would still show the affordance next to the reason label. Now gated on the same predicate `LeadDetail` uses — `!visibleStatusReason(leadRow)` (via a new `leadShowsGoal`) — so the expand appears only when the goal is the actually-rendered detail. This closes both the `turnTextFull` path and the overflow path for that state.
2. **One-frame flicker on the unmount reset** — changed the `WorkingDetail` overflow-reset cleanup from a passive `useEffect` to `useLayoutEffect`, so it clears `leadGoalOverflow` synchronously in the commit that removes `WorkingDetail`, before paint.

typecheck + biome + lint:tokens clean; web suite green (1088). No schema/data-layer changes.

yuezengwu (approved):
Re-reviewed the synchronized head `563c13a0`.

The new commit is narrowly scoped to the R5 note:

- `canExpand` now goes through `leadShowsGoal`, which mirrors `LeadDetail` by requiring `main === "working"` and no visible `statusReason`. That prevents the expand affordance from appearing beside a reason row, including the `turnTextFull` path.
- `WorkingDetail` now clears the reported overflow from a layout-effect cleanup, so the stale overflow flag is reset before paint when the goal detail unmounts.

The original width-clipped goal behavior still holds: when the goal is the rendered detail, visual overflow or `turnTextFull` can open the card, and the card still uses `turnTextFull ?? turnText`.

No blocker found. Approval holds. I did not rerun the web suite locally because this fresh review worktree has no `node_modules`; `git diff --check origin/main...HEAD` is clean.

baixiaohang (approved):
## Review by codex-assistant - PR #1441

**Recommendation**: approve
- Rationale: The follow-up commit addresses the previously noted stale-affordance edge without changing the broader data contract or adding new behavioral risk.

**Risk level**: C
- Path baseline: only `packages/web/**` component and DEV preview changes -> C.
- Semantic lift: none; this remains local Web presentation logic and does not touch subscriptions, WS frame parsing, or shared data-layer contracts.

**PR summary**
- Problem: The Workspace composer status bar could visibly clip a lead goal by rail width while offering no expand affordance unless the server had also sent `turnTextFull`.
- Approach: The rendered goal span now reports actual overflow, `canExpand` is gated on the same predicate that decides whether `LeadDetail` actually renders the goal, and the card falls back from `turnTextFull` to `turnText` for width-only truncation.
- Impacted modules: `packages/web/src/components/chat/compose-status-bar.tsx`, `packages/web/src/pages/compose-status-bar-preview.tsx`.

**Review findings**
✅ 1. The latest commit closes the prior non-blocking R5 note: expand is now suppressed when the row renders a visible `statusReason`, and the overflow reset cleanup is layout-synchronous before paint.

**Action taken**
- Approved latest head.

PR #1447, opened 2026-07-06
fix(server): count confirmed landing trial turns

baixiaohang:
## Summary

- Keep landing campaign trial chats writable until their configured agent-turn cap is reached, including while the agent is currently running.
- Count trial agent turns from confirmed `turn_end(success)` runtime events before inbox ACK, using a stable `turnCompletionId` derived from durable inbox/message ids instead of provider turn ids.
- Preserve normal Codex behavior: non-trial Codex success turns still emit fire-and-forget session events and ACK normally.
- Treat legacy `awaiting_user` trial metadata without `awaitingUserKind` as request waits, so ordinary human messages cannot bypass unresolved requests.

## Details

The previous follow-up implementation relied on chat message writes to advance the trial lifecycle, but Codex final-text mirror is retired and successful turns now surface as runtime `turn_end` events. This PR moves the landing trial turn budget transition to confirmed runtime events and makes it idempotent by storing processed logical completion ids in chat metadata.

The confirmed pre-ACK path is intentionally limited to landing campaign trial agents. For ordinary Codex agents, session events remain trace-only and do not gate inbox ACK.

## Verification

- `packages/client`: `pnpm exec vitest run src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-usage-limit.test.ts --passWithNoTests --maxWorkers=2 --minWorkers=1`
- `packages/server`: `pnpm exec vitest run src/__tests__/landing-campaigns-start.test.ts src/__tests__/ws-session-event.test.ts --passWithNoTests --maxWorkers=2 --minWorkers=1`
- `packages/shared`: `pnpm exec vitest run src/__tests__/session-event-schema.test.ts --passWithNoTests --maxWorkers=2 --minWorkers=1`
- `packages/client`: full `pnpm test -- --maxWorkers=2 --minWorkers=1`
- `packages/server`: full `pnpm test -- --maxWorkers=2 --minWorkers=1`
- `pnpm typecheck`
- `git diff --check`
- `pnpm check` (exit 0; existing warnings/info only)

### DIFF
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -5,6 +5,7 @@ import type { SessionEvent } from "@first-tree/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CodexAppServerRpcError, CodexAppServerTransportError } from "../handlers/codex/app-server/client.js";
 import { createCodexAppServerHandler } from "../handlers/codex/app-server/index.js";
+import { LANDING_TRIAL_TURN_COMPLETION_CONFIRM_FAILED } from "../handlers/codex/turn-completion.js";
 import { writeAgentBriefing } from "../runtime/bootstrap.js";
 import { setCliBinding } from "../runtime/cli-binding.js";
 import type { DeliveryToken, SessionContext, SessionMessage } from "../runtime/handler.js";
@@ -624,17 +625,19 @@ describe("codex app-server handler", () => {
     await handler.shutdown();
   });
 
-  it("does not complete delivery when confirmed success turn_end is rejected", async () => {
+  it("leaves landing trial delivery recoverable when confirmed success turn_end is rejected", async () => {
     stubLandingTrialHostEnv("confirm-reject");
 
     const fake = new FakeAppServerClient();
     const token = makeDeliveryToken();
+    const failSessionForRecovery = vi.fn<NonNullable<SessionContext["failSessionForRecovery"]>>();
     const emitEventConfirmed = vi
       .fn<NonNullable<SessionContext["emitEventConfirmed"]>>()
       .mockRejectedValue(new Error("session event persist failed"));
     const handler = makeHandler(fake);
     const ctx = makeContext({
       emitEventConfirmed,
+      failSessionForRecovery,
       agentMetadata: trialAgentMetadata,
     });
     const message = makeMessage("m1", "first", 101);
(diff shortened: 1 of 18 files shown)


yuezengwu (changes requested):
Blocking issue found.

The trial `turn_end(success)` confirmation failure path can still ACK the inbox entry without counting the trial turn. In both Codex handlers, the landing-trial success path awaits `emitEventConfirmed(...)` before `token.complete(...)`. If the server rejects that confirmed event after the provider turn has already completed, the handler throws. `SessionManager.startNewSession`/`resumeSession` then treats that throw as a normal `session_start`/`session_resume` failure with `replaySafety: "pre_provider"`; after retries exhaust it can persist a confirmed runtime error event and call `completeDeliveryTurn(...)`, ACKing the original inbox entry as a consumed error.

That means a transient failure in `completeLandingCampaignTrialAgentTurn(...)` (or any server-side persist failure for the confirmed turn event) can produce this sequence:

1. Codex successfully runs the landing trial turn.
2. Confirmed `turn_end(success, turnCompletionId)` is rejected, so the trial metadata is not advanced.
3. Session failure fallback later persists a confirmed error event.
4. The original inbox entry is ACKed through `completeDeliveryTurn(...)`.

At that point the user input is consumed but `completedAgentTurns` was never incremented, which breaks the PR's core guarantee that the trial turn budget transition is confirmed before inbox ACK. This also allows automatic provider replay from the session retry path even though the provider had already entered/completed the turn.

Please make this a distinct post-provider confirmation-failure path, not a generic session start/resume failure. For example, the handler/session manager should leave the delivery as recovery debt (or otherwise retry the confirmation without ACKing) and must not terminal-ACK the inbox entry unless the trial turn completion was confirmed or an explicitly intended terminal outcome was confirmed.

baixiaohang:
Addressed the post-provider confirmation failure blocker in `cacf62b5`.

Summary:
- Landing trial Codex success turns still await confirmed `turn_end(success, turnCompletionId)` before `token.complete(...)`.
- If that confirmed event fails after the provider turn completed, the Codex handler now catches a dedicated `LandingTrialTurnCompletionConfirmError`, calls `token.retry(...)`, and closes the session via `failSessionForRecovery(...)` instead of letting the error bubble into the generic `SessionManager` start/resume failure fallback.
- Ordinary Codex remains unchanged: success `turn_end` is fire-and-forget and delivery ACK proceeds normally.
- No coordinator/server ACK protocol changes; this is scoped to the landing-trial Codex handler boundary.

Local verification after the fix:
- `pnpm exec vitest run src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-usage-limit.test.ts --passWithNoTests --maxWorkers=2 --minWorkers=1` — 46 passed.
- `pnpm --filter @first-tree/client test -- --maxWorkers=2 --minWorkers=1` — 115 files passed / 1229 tests passed, 2 smoke skipped.
- `pnpm --filter @first-tree/client typecheck` — passed.
- `pnpm typecheck` — passed.
- targeted `pnpm exec biome check ...` — passed.
- `git diff --check` — passed.
- `pnpm check` — exit 0 with existing warnings/info only.

yuezengwu (approved):
Re-reviewed cacf62b5. The previous blocker is addressed: landing-trial success turn confirmation failures now stay on the dedicated recovery path, retry the delivery, call failSessionForRecovery, and return before token.complete, so they no longer fall into the generic start/resume failure path that could ACK without advancing completedAgentTurns.

No further blocking issues found in this pass. I did not rerun the full test suite in this review worktree because dependencies are not installed here; git diff --check passes, and I reviewed the posted verification plus the changed code/tests.

PR #1451, opened 2026-07-06
fix(web): follow live timeline turn for chat agent work status

Gandy2025:
## Problem

In a chat, the right-rail participant roster shows an agent's work status as **"Idle"** while the agent is genuinely working — visibly, as a live `WorkingTurn` card in the timeline.

## Root cause

The roster (`AgentStatusPanel`), the compose status bar, and the agent hovercard all read the server **composite** agent status. Its `working` axis is a per-chat runtime **heartbeat** with a `RUNTIME_STALE_MS` (60s) freshness window; the client re-affirms every ~20s. Two ways it desyncs from the visible turn:

1. **Heartbeat lapse** — a flapping client WS drops the re-affirm frames, the composite ages out to `ready` ("Idle") mid-turn.
2. **Missed invalidation (observed live)** — a steady same-value `working` re-affirm fires **no** admin-WS invalidation (the server only notifies on value-change or stale→fresh). So if the one `idle → working` invalidation is missed (WS hiccup / backgrounded tab), the roster's composite query has no further trigger and stays stuck on "Idle" even though the server is `working`.

Either way: roster reads **Idle** while the timeline's `WorkingTurn` card is visibly on screen.

## Fix (frontend stopgap — plan A)

Reconcile the composite against the timeline's **authoritative** live-turn signal. `ChatView` already groups each agent's un-ended turn into a `workgroup` item (⇔ a mounted `WorkingTurn`); we derive `liveTurnAgentIds` from it and OR it into the working axis via a new pure helper:

```ts
applyLiveTurn(main, hasLiveTurn) // ready → working when a live turn exists; otherwise unchanged
```

**Upgrade-only**: it only lifts `ready → working`; it never downgrades and never masks the authoritative `failed` / `paused` / `offline`, so it cannot hide genuine work or override a real failure. Threaded to the **roster**, **hovercard**, and **compose status bar** so all three agree with the visible turn (and with each other). The signal rides the session-event stream, which is invalidated on every event — so it stays fresh even when the composite's coarse invalidation is missed.

## Scope / limitations

- The timeline events feed is single-(primary-)agent, so `liveTurnAgentIds` only ever covers the **primary agent** — exactly the agent whose `WorkingTurn` is on screen. Secondary agents keep rendering the composite (upgrade-only ⇒ never wrongly forced to working). Documented inline.
- This is a **frontend stopgap**. The deeper server-side fix (turn-end doesn't reliably emit `idle`; the heartbeat can lapse mid-turn; steady-state fires no invalidations) is tracked separately.

## Testing

- New unit tests for `applyLiveTurn` (upgrade-only matrix).
- `pnpm --filter @first-tree/web typecheck` ✓ (+ `lint:tokens` ✓), biome ✓, 171 tests pass in the touched areas.
- Two independent code reviews (correctness + consistency) — no blockers; both review findings folded in (compose-bar reconciliation + scope docs).

### DIFF
--- a/packages/web/src/components/chat/agent-hovercard.tsx
+++ b/packages/web/src/components/chat/agent-hovercard.tsx
@@ -1,18 +1,19 @@
 import type { AgentChatStatus } from "@first-tree/shared";
 import { useQuery } from "@tanstack/react-query";
 import { ArrowRight, MessageSquare } from "lucide-react";
-import type { ReactNode } from "react";
+import { type ReactNode, useContext } from "react";
 import { useNavigate } from "react-router";
 import { chatAgentStatusQueryKey, fetchChatAgentStatuses } from "../../api/agent-status.js";
 import { getAgent } from "../../api/agents.js";
 import { getChat } from "../../api/chats.js";
-import { applyLiveTurn, viewOf } from "../../lib/agent-status-view.js";
+import { reconcileLiveTurn, viewOf } from "../../lib/agent-status-view.js";
 import { useClientMap } from "../../lib/use-client-map.js";
 import { useMemberNameMap } from "../../lib/use-member-name-map.js";
 import { Avatar } from "../avatar.js";
 import { Button } from "../ui/button.js";
 import { HoverCard, type HoverCardPlacement } from "../ui/hover-card.js";
 import { StatusGlyph } from "../ui/status-glyph.js";
+import { LiveTurnAgentsContext } from "./live-turn-context.js";
 import { WorkingChip } from "./working-chip.js";
 
 /**
@@ -32,7 +33,6 @@ export function AgentHovercard({
   agentId,
   chatId,
   name,
-  hasLiveTurn = false,
   placement = "bottom",
   triggerClassName,
   children,
@@ -45,9 +45,6 @@ export function AgentHovercard({
    * override the visible name text and read identically for every agent.
    */
   name: string;
-  /** The agent has a live (un-ended) turn in the timeline; upgrades the card's
-   *  working axis via `applyLiveTurn`, matching the roster row. */
-  hasLiveTurn?: boolean;
   placement?: HoverCardPlacement;
   triggerClassName?: string;
   children: ReactNode;
@@ -63,26 +60,14 @@ export function AgentHovercard({
         width: "var(--sp-70)",
         maxWidth: "calc(100vw - var(--sp-4))",
       }}
-      content={({ close }) => (
-        <AgentHovercardBody agentId={agentId} chatId={chatId} hasLiveTurn={hasLiveTurn} onAction={close} />
-      )}
+      content={({ close }) => <AgentHovercardBody agentId={agentId} chatId={chatId} onAction={close} />}
     >
       {children}
     </HoverCard>
   );
 }
 
-function AgentHovercardBody({
-  agentId,
-  chatId,
-  hasLiveTurn,
-  onAction,
-}: {
-  agentId: string;
-  chatId: string;
-  hasLiveTurn: boolean;
-  onAction: () => void;
-}) {
+function AgentHovercardBody({ agentId, chatId, onAction }: { agentId: string; chatId: string; onAction: () => void }) {
   const navigate = useNavigate();
   const resolveMember = useMemberNameMap();
   const { resolve: resolveClient } = useClientMap();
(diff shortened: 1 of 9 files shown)


yuezengwu (changes requested):
Goal/core change: this PR reconciles chat UI status against the currently mounted timeline workgroup so a visible `WorkingTurn` can lift a stale composite `ready` display to `working`. The helper is intentionally upgrade-only, and threading it into the roster + compose status bar looks directionally right.

Finding:

- `packages/web/src/pages/workspace/center/chat-view.tsx:792` / `:814` still mount `AgentHovercard` directly from message rows without passing the new live-turn signal. In the missed-invalidation case this leaves a straightforward inconsistent surface: the same chat can show the `WorkingTurn`, the roster and compose bar as Working, but hovering the agent avatar/name in any message row opens the shared hovercard with its default `hasLiveTurn=false`, so it still renders the stale composite Idle status. Since the PR scope explicitly includes the agent hovercard, please thread `liveTurnAgentIds` through `ChatTimeline`/`MessageRow` and pass `hasLiveTurn={liveTurnAgentIds.has(msg.senderId)}` for these direct call sites, or otherwise centralize the reconciliation so all chat-scoped hovercard entry points agree.

Local verification: attempted the targeted web vitest files, but this review worktree has no `node_modules` and `vitest` was not available.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1451

- Rationale: The UI reconciliation is directionally right, but the implementation mutates only the derived `main` field and then passes around an impossible `AgentChatStatus`.

**Risk level**: C
- Semantic lift: none; this is UI status presentation, with no subscription, WS frame parsing, or API contract change.

**PR Summary**
- Problem: A user can see a live `WorkingTurn` in the chat timeline while the roster, hovercard, and compose status bar still show the agent as Idle after a runtime heartbeat lapse or missed status invalidation.
- Approach: derive agents with live timeline workgroups in `ChatView`, then upgrade affected status presentation from `ready` to `working` in the roster, hovercard, and compose bar.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, chat status components, and `agent-status-view`.

**Review findings**
❌ 1. `applyLiveTurn` breaks the `AgentChatStatus` invariant before handing the object to shared status consumers. The shared schema defines `main` as the value derived from the axis fields, but the new call sites create `{ ...status, main: applyLiveTurn(...) }` while leaving `working` false (`packages/web/src/components/chat/agent-status-panel.tsx:151`, `packages/web/src/components/chat/compose-status-bar.tsx:167`, `packages/web/src/components/chat/agent-hovercard.tsx:110`). That gives downstream code a status object that disagrees with itself: `status.main` says `working`, while `status.working` and `deriveMainStatus(status)` still say not working. Please reconcile at the axis level, for example with a helper that accepts/returns a full `AgentChatStatus` and sets `working: true` when a live turn upgrades `ready`, or keep the raw `AgentChatStatus` intact and pass a separate presentation-only main/view value to render-only consumers. The tests should pin that full-object contract, not only the bare main-token mapping. [R1 / `packages/web/src/lib/agent-status-view.ts:66`]

✅ 2. The upgrade-only policy avoids masking `failed`, `paused`, and `offline`, and the primary-agent scope is documented at the `ChatView` derivation site.

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks both — addressed both findings in `c13da23c`.

**@baixiaohang (R1 — axis invariant):** you're right, `{ ...status, main }` with `working` left false produced a status that fails `main === deriveMainStatus(...)` (the schema `superRefine`). Reworked the helper to reconcile at the **axis level**: `reconcileLiveTurn(status, hasLiveTurn)` sets `working: true` when a live turn upgrades a `ready` status, so the returned object is a valid `AgentChatStatus` (and `deriveMainStatus` agrees). `activity` is left as-is — `working: true` + `activity: null` is the schema-valid "Working, no tool detail" shape. Tests now assert the full-object contract (incl. `deriveMainStatus(out) === out.main`), not just the bare token.

**@yuezengwu (message-row hovercards):** agreed, and took the centralization option. Introduced `LiveTurnAgentsContext`, provided once by `ChatView` and consumed by the roster, compose bar, and hovercard. The message-avatar / message-name hovercards (`chat-view.tsx` MessageRow) now reconcile via the same context by construction — as do any future hovercard entry points — so they can't disagree. This also removed the prop-drilling through `ChatRightSidebar` → `ParticipantsSection`.

Green locally: `tsc` + `lint:tokens`, biome, and the touched vitest suites (incl. the rewritten `reconcileLiveTurn` tests). Re-requesting review.

yuezengwu (approved):
Re-reviewed `c13da23c`.

The two previous blockers are addressed:

- `reconcileLiveTurn(status, hasLiveTurn)` now reconciles at the axis level (`working: true` + `main: "working"` for the `ready` + live-turn case), so the returned object remains a valid `AgentChatStatus` and `deriveMainStatus(...)` agrees.
- `LiveTurnAgentsContext` is provided by `ChatView` and consumed by the roster, compose status bar, and shared `AgentHovercard`, so message-row avatar/name hovercards now pick up the same live-turn signal without prop drilling.

No new blockers found in the touched scope. Local validation: `git diff --check` passed; I did not rerun the vitest/typecheck commands because this detached review worktree has no `node_modules`/`vitest` installed.

baixiaohang (approved):
## Review by codex-assistant - PR #1451

**Recommendation**: approve
- Rationale: `c13da23c` addresses the previous status-invariant and hovercard-coverage blockers, and I found no new issues in the touched scope.

**Risk level**: C
- Semantic lift: none; this remains UI status presentation, with no subscription, WS frame parsing, or API contract change.

**PR Summary**
- Problem: A user can see a live `WorkingTurn` in the chat timeline while the roster, hovercard, and compose status bar still show the agent as Idle after a runtime heartbeat lapse or missed status invalidation.
- Approach: derive the open chat's live-turn agent set in `ChatView`, provide it via `LiveTurnAgentsContext`, and reconcile stale `ready` status to `working` at the `AgentChatStatus` axis level for chat-scoped status surfaces.
- Impacted modules: chat timeline/status components and `agent-status-view` tests/helpers under `packages/web`.

**Review findings**
✅ 1. The previous `AgentChatStatus` invariant issue is fixed: `reconcileLiveTurn(status, hasLiveTurn)` now sets `working: true` together with `main: "working"` for the `ready` + live-turn case, and tests assert `deriveMainStatus(out) === out.main`.
✅ 2. The previous hovercard gap is fixed: `LiveTurnAgentsContext` is consumed by the shared `AgentHovercard`, so roster hovercards and message-row avatar/name hovercards reconcile from the same live-turn source.
⚠️ 3. The remaining limitation is the one documented by the PR: this is a frontend stopgap over the primary-agent timeline feed; the deeper server-side heartbeat/invalidation behavior remains separate.

**Validation**
- I did not run local tests or checks per review policy.
- GitHub CI was green at review time, including `Lint & Type Check` and `Test`.

**Action taken**
- Approved.

PR #1452, opened 2026-07-06
fix: align chat image upload cap to shared attachment limit (5MB→10MB)

Gandy2025:
## What

The chat composer's per-image upload gate was hardcoded to **5MB** in `use-pending-images.ts`, with a comment claiming it was *"the Claude API per-image byte limit"*. That comment is inaccurate, and the gate was stricter than our own server already allows.

Facts (per [Anthropic vision docs](https://platform.claude.com/docs/en/build-with-claude/vision) → Image limits and costs):
- **Direct Claude API** per-image limit = **10MB** (base64).
- **5MB** is the **Amazon Bedrock / Google Cloud** limit — this repo uses neither.
- Our own attachment upload route already accepts up to `MAX_ATTACHMENT_BYTES` (**10MB**), enforced as the route `bodyLimit` (`api/orgs/attachments.ts`) **and** re-checked in `services/attachment.ts`.

So the composer's 5MB was both mislabeled and stricter than the rest of our stack — a user could upload a 6–10MB image in Claude Code but not here.

## Change

- `use-pending-images.ts`: replace the hardcoded `MAX_IMAGE_SIZE = 5MB` (+ wrong comment) with the shared `MAX_ATTACHMENT_BYTES` constant, so the composer gate tracks the real server-enforced cap from a single source of truth (no future drift). The error copy auto-updates to "Maximum 10MB per image".
- Fix two stale `≤5MB` comments in the consumers (`ask-takeover.tsx`, `new-chat-draft.tsx`).
- Update the boundary in `ask-takeover-dom.test.tsx` to `MAX_ATTACHMENT_BYTES + 1`.

## Why 10MB is safe for this path

Chat images are **not** sent to the model as raw base64 image blocks. They're stored as attachments, written to the agent's local disk, and read via the agent's **Read tool** (see the `claude-code` handler — the SDK does not reliably forward `{type: "image"}` blocks). So the binding limit is the raw-byte attachment storage cap (10MB), which the frontend now matches exactly.

## Verification

- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) — clean.
- `biome check` on all changed files — clean.
- `ask-takeover-dom.test.tsx` — 18/18 pass (oversized-rejection path still exercised at cap+1).
- 2 independent code reviews (correctness + adversarial); transport/storage path confirmed safe to 10MB by both.
- **Not run:** a live end-to-end upload of a >5MB image through a running Claude agent (needs the full local stack). The transport is proven safe; the model-accept step relies on the agent's Read tool downsampling large images. Reviewers/maintainers: flag if you want that E2E before merge.

### DIFF
--- a/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
+++ b/packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx
@@ -1,5 +1,6 @@
 // @vitest-environment happy-dom
 
+import { MAX_ATTACHMENT_BYTES } from "@first-tree/shared";
 import { act, type ReactElement } from "react";
 import { createRoot, type Root } from "react-dom/client";
 import { afterEach, describe, expect, it, vi } from "vitest";
@@ -381,8 +382,8 @@ describe("AskTakeover", () => {
     const c = await renderDom(
       <AskTakeover body="# Evidence?" payload={{ multiSelect: false }} onReply={onReply} onSkip={() => {}} />,
     );
-    // 5MB + 1 byte — over the per-image cap usePendingImages enforces.
-    const big = new File([new ArrayBuffer(5 * 1024 * 1024 + 1)], "big.png", { type: "image/png" });
+    // One byte over the per-image cap usePendingImages enforces.
+    const big = new File([new ArrayBuffer(MAX_ATTACHMENT_BYTES + 1)], "big.png", { type: "image/png" });
     const fileInput = c.querySelector<HTMLInputElement>('input[type="file"]');
     if (!fileInput) throw new Error("file input missing");
     await changeFiles(fileInput, [big]);
--- a/packages/web/src/components/chat/ask-takeover.tsx
+++ b/packages/web/src/components/chat/ask-takeover.tsx
@@ -127,9 +127,9 @@ export function AskTakeover({
   // Keep the card (and its pinned footer) above the on-screen keyboard.
   const keyboardInset = useKeyboardInset();
 
-  // Staged image attachments — same hook (and same `image/* ≤5MB` rules + object-
-  // URL lifecycle) the chat composer uses. A local validation error (oversized
-  // image) renders alongside any host send error below.
+  // Staged image attachments — same hook (and same `image/*` + attachment-cap
+  // rules + object-URL lifecycle) the chat composer uses. A local validation
+  // error (oversized image) renders alongside any host send error below.
   const [imageError, setImageError] = useState<string | null>(null);
   const { pendingImages, addImages, removeImage } = usePendingImages({
     onError: setImageError,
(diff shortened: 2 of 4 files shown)


yuezengwu (approved):
Goal/core change reviewed: this PR removes the chat composer's stale hardcoded 5MB per-image gate and makes `usePendingImages` enforce the shared `MAX_ATTACHMENT_BYTES` cap instead, with adjacent comments/tests updated to match the server attachment limit.

No findings. The 10MB boundary matches `createAttachment`'s exact service cap, the route has body-limit headroom but still re-checks the same cap, and the Claude Code delivery path materializes image attachments as local files for the Read tool rather than sending raw base64 image blocks to the model.

Verified locally:
- `pnpm --filter @first-tree/web test -- packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx` (ran the full web Vitest suite: 134 files / 1089 tests passed)
- `pnpm exec biome check` on the four changed files

GitHub checks are also green except the intentional `Validate Skill` skip.

Gandy2025:
@baixiaohang — friendly ping for a review here when you have a moment. Small frontend-only change: it aligns the chat composer's per-image upload gate to the shared `MAX_ATTACHMENT_BYTES` cap (the existing server-enforced attachment limit) and fixes a stale/incorrect "5MB Claude API limit" comment. @yuezengwu has already approved with no findings, and all CI is green. Thanks!

baixiaohang (approved):
## Review by codex-assistant - PR #1452

**Recommendation**: approve
- Rationale: The composer now follows the existing attachment byte cap as the single source of truth, while the send/runtime path remains attachment-backed rather than embedding image bytes in chat message content.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed -> C.
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or the web data layer contract.

**PR Summary**
- Problem: Users could hit a stale web-only 5MB per-image rejection even though the attachment storage path already accepts up to the shared 10MB cap.
- Approach: Reuse `MAX_ATTACHMENT_BYTES` in `usePendingImages`, refresh stale consumer comments, and move the oversized-image test boundary to `MAX_ATTACHMENT_BYTES + 1`.
- Impacted modules: `packages/web` chat image staging, AskTakeover, new-chat draft, and the related DOM test.

**Review findings**
✅ 1. The shared cap is already enforced server-side at both the org attachment upload route and the attachment service boundary, so the frontend gate now tracks the real storage contract instead of duplicating a stale constant.
✅ 2. The image send/runtime path remains ref-only at the message layer: web uploads bytes to attachments first, sends ids/metadata in the `format: "file"` content, and clients materialize referenced images locally for the runtime to read.
⚠️ 3. Residual risk: I did not require a live >5MB Claude-agent E2E for this web-only alignment. If Claude Code image processing regresses, an accepted image could still fail later at model-read time, but that is outside this PR's storage/message boundary and the PR body calls out the missing E2E explicitly.

**Action taken**
- Approved.

PR #1453, opened 2026-07-06
fix(client): allow landing trial system reads

baixiaohang:
## Summary

- Change the landing Codex trial permission profile from strict `:minimal` reads to `:root` read with selected host credential denies, so sandboxed tool commands can read Linux DNS/TLS/system config needed by common CLIs.
- Keep the landing workspace writable and continue denying host Codex auth, host First Tree state dirs, and host SSH keys from tool commands.
- Preserve host `gh` usability by passing `GH_CONFIG_DIR` when the host GitHub CLI config exists; this is an explicit product tradeoff for landing trials that need public-repo PR/comment workflows.
- Extend the real Codex app-server smoke to prove DNS/server fetch works while selected credentials remain unreadable.

## Security boundary

This PR intentionally changes the landing trial tool boundary from strict workspace-only reads to root-read plus selected credential denies. The app-server process still uses host `CODEX_HOME` for provider auth, but tool commands cannot read that Codex auth path. First Tree runtime token/API/GitHub/proxy/cert env vars remain scrubbed.

The selected host credential deny list covers:

- host `CODEX_HOME` / `~/.codex`
- `~/.ssh`
- common host First Tree state dirs: `~/.first-tree*`

It does not deny host `gh` config, because landing trial agents are expected to use the configured public-repo GitHub capability on the staging machine.

## Product/security decision

Human decision in the First Tree `agent托管方案` chat on 2026-07-06: keep the current landing trial model as `:root=read` plus the small selected deny list above, instead of denying the whole host home and reopening only selected subtrees.

Operational assumption for this hosted landing-trial model: machines running these public-repo landing trials are treated as low-sensitive trial hosts. They must not contain unapproved high-value host credentials outside the explicit deny list. The configured `gh` account is intentionally available for public-repo PR/comment workflows.

## Verification

- `pnpm exec vitest run src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts src/__tests__/codex-landing-app-server-smoke.test.ts --passWithNoTests --maxWorkers=2 --minWorkers=1`
- `SKIP_CODEX_LANDING_PROVIDER_SMOKE=1 pnpm --filter @first-tree/client smoke:landing-codex-auth-sandbox`
- `pnpm exec biome check packages/client/src/handlers/codex/app-server/workspace-sandbox.ts packages/client/src/handlers/codex/app-server/index.ts packages/client/src/__tests__/workspace-sandbox.test.ts packages/client/src/__tests__/codex-app-server-handler.test.ts packages/client/src/__tests__/codex-landing-app-server-smoke.test.ts`
- `git diff --check`
- Full client suite: `pnpm exec vitest run --passWithNoTests --maxWorkers=2 --minWorkers=1` from `packages/client` -> 115 files passed + 1 skipped; 1230 passed + 2 skipped
- `pnpm typecheck`
- `pnpm check` exits 0; existing warnings/info only outside this diff

Real smoke key output with `codex-cli 0.142.5` and provider turn skipped:

```text
clean-config:
codex_auth=denied
host_first_tree=denied
host_ssh=denied
gh_config_dir=set
resolv_conf=readable
dns_lookup=ok
workspace_write=ok

host-auth:
codex_auth=denied
host_first_tree=denied
host_ssh=denied
gh_config_dir=set
gh_version=ok
workspace_write=ok
resolv_conf=readable
dns_lookup=ok
server_fetch=status:200
leaked_env=
```

### DIFF
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -463,10 +463,12 @@ describe("codex app-server handler", () => {
   it("starts landing campaign trial app-server with host Codex auth and managed workspace-only permissions", async () => {
     const hostHome = join(workspaceRoot, "..", "host-home");
     const codexHome = join(hostHome, ".codex");
+    const ghConfigDir = join(hostHome, ".config", "gh");
     const firstTreeHome = join(workspaceRoot, "first-tree-home");
     const cliBinDir = join(workspaceRoot, "first-tree-cli-bin");
     mkdirSync(cliBinDir, { recursive: true });
     mkdirSync(codexHome, { recursive: true });
+    mkdirSync(ghConfigDir, { recursive: true });
     writeFileSync(join(cliBinDir, "first-tree-test"), "#!/bin/sh\n", { mode: 0o755 });
     vi.stubEnv("HOME", hostHome);
     vi.stubEnv("FIRST_TREE_HOME", firstTreeHome);
@@ -522,6 +524,7 @@ describe("codex app-server handler", () => {
     expect(capturedEnv?.FIRST_TREE_HOME).toBe(join(workspaceRoot, ".first-tree-workspace", "outbox-home"));
     expect(capturedEnv?.HOME).toBe(workspaceRoot);
     expect(capturedEnv?.CODEX_HOME).toBe(codexHome);
+    expect(capturedEnv?.GH_CONFIG_DIR).toBe(ghConfigDir);
     expect(capturedEnv?.PATH?.split(":")[0]).toBe(cliBinDir);
     expect(capturedEnv?.OPENAI_API_KEY).toBeUndefined();
     expect(capturedEnv?.CODEX_API_KEY).toBeUndefined();
@@ -541,9 +544,11 @@ describe("codex app-server handler", () => {
             [workspaceRoot]: true,
           },
           filesystem: {
-            ":minimal": "read",
+            ":root": "read",
             [workspaceRoot]: "write",
             [codexHome]: "deny",
+            [join(hostHome, ".first-tree-staging")]: "deny",
+            [join(hostHome, ".ssh")]: "deny",
           },
           network: {
             enabled: true,
--- a/packages/client/src/__tests__/codex-landing-app-server-smoke.test.ts
+++ b/packages/client/src/__tests__/codex-landing-app-server-smoke.test.ts
@@ -164,15 +164,21 @@ describeSmoke("landing Codex app-server auth and sandbox smoke", () => {
     const codexVersion = commandOutput(`${shellQuote(codexBinary)} --version || true`);
     const root = realpathSync(mkdtempSync(join(tmpdir(), "ft-landing-codex-clean-smoke-")));
     tempRoots.push(root);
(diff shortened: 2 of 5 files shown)


yuezengwu (changes requested):
I understand the implementation goal here as: change landing Codex trials from strict `:minimal`/workspace-only tool reads to a root-read Codex permission profile, keep the landing workspace writable, deny selected host credential locations (`CODEX_HOME`/`.codex`, `.ssh`, `.first-tree*`), and deliberately pass host `GH_CONFIG_DIR` so staging landing trials can use `gh` for public-repo workflows.

Blocking issue:

`packages/client/src/handlers/codex/app-server/workspace-sandbox.ts` now grants `":root" = "read"` and only denies the paths enumerated in `LANDING_CODEX_HOST_CREDENTIAL_DENY_RELATIVE_PATHS`. That still leaves arbitrary service-user home credentials readable by landing trial tool commands, for example `~/.git-credentials`, `~/.netrc`, `~/.aws`, `~/.npmrc`, `~/.docker/config.json`, and other host-local auth/config files. The test currently codifies part of that by asserting the generated profile does **not** contain `.git-credentials`.

That conflicts with the current landing trial runtime boundary in the Context Tree: the sandboxed trial environment must not expose the service user's real local credentials to the agent shell, and should receive only the minimum First Tree local home plus the scoped outbox credential. Please tighten this before merge, e.g. deny the host home by default and explicitly re-allow the landing workspace and the approved `GH_CONFIG_DIR` exception, or otherwise enforce and test a complete credential boundary for the service host. If the intended product/security decision is to expose all non-listed host files under root-read, that needs explicit human/security signoff and a Context Tree update rather than landing as a code-only bug fix.

No DB/schema or core data-structure changes found; the concern is the runtime security boundary.

baixiaohang:
Follow-up on the requested security-boundary clarification: after discussing the tradeoff in the First Tree chat, the human decision is to keep this landing trial model as `:root=read` plus the selected deny list, rather than deny the whole host home.

I updated the PR body with the explicit product/security decision and operational assumption: these hosted landing-trial machines are treated as low-sensitive public-repo trial hosts, should not contain unapproved high-value host credentials outside the explicit deny list, and intentionally expose the configured `gh` capability for public-repo PR/comment workflows.

No code changes were made in this follow-up; the current implementation remains the root-read + selected-deny model already covered by the smoke tests.

yuezengwu (approved):
Re-reviewed after the PR body and follow-up comment were updated.

No code changed since my requested-changes review. The implementation still matches the stated model: `:root=read` plus selected host credential denies, workspace write, host Codex auth denied from tool commands, env scrubbed, and intentional `GH_CONFIG_DIR` pass-through for public-repo workflows.

My previous blocker was that this boundary exposes non-listed host-home files and therefore needed an explicit product/security decision rather than landing as an implicit code-only change. The PR now records that human decision and the operational assumption that these hosted landing-trial machines are low-sensitive public-repo trial hosts without unapproved high-value credentials outside the explicit deny list.

Given that decision, I do not see remaining code-review blockers. Please sync the Context Tree runtime boundary from this decision as a follow-up, because it still describes Codex landing trials as strict workspace-only/no real host credentials.

PR #1454, opened 2026-07-06
feat(web): whole compose-status-bar lead row toggles the expand card

Gandy2025:
## What

Clicking the small `⇕` glyph was the only way to open the full-narration card — a tiny target. Now **clicking anywhere on the lead row** toggles the card. The timeline-jump is dropped on the lead row (it was an undiscoverable implicit gesture, and the card now covers "read the full goal"); the `+N` rows keep their jump.

Follow-up to #1429 / #1441.

## How

`RailRow` gained an optional `expand` prop; the lead row passes it and renders as a `<button>` that toggles `cardOpenFor`. The `⇕` moved inside the row; short/fitting goals show no glyph and the row is inert.

### The subtle part (and a caught bug)

The lead row is **always the same `<button>` element with an always-present `⇕` slot** — only `onClick`/`tabIndex`/`aria-*`/`cursor` and the glyph's `visibility` toggle on `canExpand`. An earlier version switched the element *type* (`<span>`↔`<button>`) and added/removed the glyph on `canExpand`; because the overflow-measuring `WorkingDetail` lived inside it, that **remounted** it (its unmount-reset then fought its mount re-measure) and the glyph's presence changed the goal's available width across the overflow threshold — an **infinite oscillation that blanked the whole bar**. Notably `typecheck` + `biome` + all 1089 unit tests **passed while the bar was blank** (jsdom has no layout / ResizeObserver); it was caught only by driving the live preview. Keeping the element type and width constant removes it.

When the lead row is inert (nothing to expand) it carries `role="presentation"` + `tabIndex=-1` so AT reads it as plain status text, not a dead button.

## Verification

- `pnpm --filter @first-tree/web typecheck` + `biome check` + `lint:tokens` — clean.
- Web test suite green (1089).
- DEV `/preview/compose-status-bar`, driven via CDP: a **real click on the row body** (the goal text, not the glyph) opens the card; the `⇕` shows only on expandable rows; short goals stay clean; the bar renders (the oscillation reproduced as a blank bar before the fix).
- Two independent reviews (correctness + a11y/design), both **non-blocking**; folded in: corrected a stale "Click zones" JSDoc and added `role="presentation"` on the inert row.

### DIFF
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -35,10 +35,11 @@ import { formatElapsed } from "./working-chip.js";
  *     active agent (≤ ~5 visible, then internal scroll).
  *   - All quiet → the whole rail is hidden.
  *
- * Click zones: the lead/row text → jump to that agent's timeline anchor
- * (working → WorkingTurn, failed → ErrorRow); +N / chevron → expand. Data is
- * the shared /agent-status query (React-Query-deduped, admin-WS-live,
- * ~1s-throttled).
+ * Click zones: the lead row → toggle its full-narration card (the `⇕` glyph is
+ * the affordance); a `+N` row's text → jump to that agent's timeline anchor
+ * (working → WorkingTurn, failed → ErrorRow); the `+N` chevron → expand the
+ * list. Data is the shared /agent-status query (React-Query-deduped,
+ * admin-WS-live, ~1s-throttled).
  */
 const ATTENTION: ReadonlySet<string> = new Set(["failed", "working"]);
 const TICK_INTERVAL_MS = 1000;
@@ -342,6 +343,10 @@ function RailRow({
         <button
           type="button"
           ref={expand.triggerRef}
+          // Inert when there's nothing to expand: drop the button semantics so AT
+          // reads it as plain status text, not a dead button (the element type
+          // must stay `<button>` to avoid the remount described above).
+          role={expand.canExpand ? undefined : "presentation"}
           onClick={expand.canExpand ? expand.onToggle : undefined}
           tabIndex={expand.canExpand ? undefined : -1}
           aria-expanded={expand.canExpand ? expand.open : undefined}
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -234,22 +234,13 @@ export function ComposeStatusBar({
           nameOf={nameFor(agents)}
           mounted={mounted}
           onGoalOverflowChange={setLeadGoalOverflow}
+          expand={{
+            canExpand,
+            open: cardOpen,
+            onToggle: () => setCardOpenFor((cur) => (cur === leadRow.agentId ? null : leadRow.agentId)),
+            triggerRef: cardTriggerRef,
+          }}
         />
-        {canExpand ? (
-          <button
-            type="button"
-            ref={cardTriggerRef}
-            aria-label={cardOpen ? "Collapse full narration" : "Expand full narration"}
-            aria-expanded={cardOpen}
-            onClick={() => setCardOpenFor((cur) => (cur === leadRow.agentId ? null : leadRow.agentId))}
-            className="inline-flex shrink-0 items-center"
-            style={{ border: 0, background: "transparent", padding: 0, cursor: "pointer", color: "var(--fg-4)" }}
-          >
-            {/* A distinct unfold glyph (not a single chevron) so it never reads
-                as, or collides with, the adjacent `+N` expand chevron. */}
-            <ChevronsUpDown className="h-3.5 w-3.5" />
-          </button>
-        ) : null}
         {others.length > 0 ? (
           <button
             type="button"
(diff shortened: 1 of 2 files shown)


yuezengwu (changes requested):
Requesting changes for one accessibility regression.

The implementation goal makes sense: the lead row should become the full-narration toggle when expandable, while `+N` rows keep timeline jump. The subtle fixed-width/constant-wrapper concern is also real. But the inert lead-row path currently does not achieve the stated AT behavior.

- `packages/web/src/components/chat/compose-status-bar.tsx:349`: `role="presentation"` on the native `<button>` is not enough to make the inert row read as plain status text. In Chrome's accessibility tree, `<button type="button" role="presentation" tabIndex="-1">…</button>` still exposes a `button` role, so short/fitting goals become a dead button for screen-reader users: it is announced as actionable but has no click handler and is intentionally not tabbable. Please keep the no-remount/constant-width property using a stable non-button wrapper instead, e.g. a stable `<span>` that only gets `role="button"`, `tabIndex=0`, and Enter/Space key handling when `canExpand`, or another pattern where the inert state is actually static text in the AX tree.

I also checked the current `main` merge result because this branch is behind #1451; the auto-merge preserves `LiveTurnAgentsContext` / `reconcileLiveTurn`, so no issue there.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1454

- Rationale: The row-click interaction is the right direction, but the inert lead-row path still exposes a non-actionable native button to assistive tech.

**Risk level**: C
- Semantic lift: none; this does not touch Web subscriptions, WS frame parsing, or data-layer contracts.

**PR Summary**
- Problem: users trying to read a full running goal had to hit a tiny `⇕` affordance, so the most useful status detail was hard to discover and operate.
- Approach: make the compose-status-bar lead row toggle the full-narration card while keeping `+N` rows as timeline jumps; keep the glyph slot stable to avoid width/remount oscillation around overflow measurement.
- Impacted modules: `packages/web/src/components/chat/compose-status-bar.tsx`, `packages/web/src/pages/compose-status-bar-preview.tsx`

**Review findings**
❌ 1. The inert lead row is still a native button. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:349`]

When `canExpand` is false, the row renders `<button type="button" role="presentation" tabIndex={-1}>` with no click handler. `role="presentation"` is not a reliable way to turn a native interactive element into static text in the accessibility tree; screen readers/browsers can still expose the native button semantics, leaving short or fitting goals announced as a dead control. This violates the PR's stated inert-row behavior.

Please keep the stable DOM/width property with a stable non-button wrapper instead, for example a persistent `<span>` that only receives `role="button"`, `tabIndex=0`, `onClick`, and Enter/Space handling when `canExpand`, while keeping the `⇕` slot present but hidden in layout when inert. That preserves the oscillation fix without shipping a non-actionable button.

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks @yuezengwu and @baixiaohang — you're both right, `role="presentation"` on a native `<button>` is ignored by browsers/AT, so short goals were still a dead control. Fixed in `e7b55884`.

The row is now **always a plain `<span>`** — static status text in the a11y tree, and (importantly) `WorkingDetail` never remounts. When expandable, a **transparent native `<button>` is overlaid** (`absolute inset-0`) to take the click:

- Inert rows → just the `<span>` text, no button in the AX tree at all.
- Expandable rows → a real native `<button>` (free Enter/Space, focusable, `aria-expanded`/`aria-label`) — no `role` hack, biome-clean.
- Mounting/unmounting only that sibling overlay never remounts the overflow-measuring child, and the ⇕ keeps its layout slot (visibility only), so the remount/width oscillation fix is preserved.

Verified: typecheck + biome + lint:tokens clean; web suite green (1089); and driven in the DEV preview — a real click on the row body opens the card, short goals show plain text with no glyph/button, and the bar renders (the oscillation reproduced as a blank bar before the original fix).

yuezengwu (approved):
Approved. `e7b55884` addresses my requested change: the lead row content is now a stable plain `<span>`, inert rows no longer expose a dead native button, and expandable rows use a real native overlay button for click/keyboard/AT semantics without remounting `WorkingDetail` or changing the glyph slot width.

I also checked the current `main` auto-merge result again; #1451's `LiveTurnAgentsContext` / `reconcileLiveTurn` path is preserved. No DB/schema or core data-structure changes involved.

baixiaohang (approved):
## Review by codex-assistant - PR #1454

**Recommendation**: approve
- Rationale: The follow-up commit resolves the inert-row accessibility blocker without reintroducing the overflow/remount instability.

**Risk level**: C
- Semantic lift: none; no Web subscription, WS frame parsing, or data-layer behavior changes.

**PR Summary**
- Problem: users needed to hit a tiny `⇕` glyph to inspect the full running goal in the compose status bar.
- Approach: make the lead row toggle the full-narration card while keeping `+N` rows as timeline jumps; the latest fix keeps inert rows as static text and uses a native overlay button only when the row is actually expandable.
- Impacted modules: `packages/web/src/components/chat/compose-status-bar.tsx`, `packages/web/src/pages/compose-status-bar-preview.tsx`

**Review findings**
✅ 1. The previous a11y blocker is resolved. [`packages/web/src/components/chat/compose-status-bar.tsx:344`]

The lead row content is now a stable `<span>`, so short/fitting goals are static text rather than dead controls. When `canExpand` is true, the separate native button overlay provides the actionable target, keyboard activation, and `aria-expanded`/`aria-label` semantics without remounting `WorkingDetail` or changing the glyph slot width.

**Action taken**
- Approved.

Issue #1392, opened 2026-07-01
GitHub App: support install binding via org-owner approval (non-owner First Tree admin initiates)


liuchao-001:
## Context

Follow-up to #1381 (webhook-sourced GitHub App install binding). That PR made binding webhook-driven and, for the **self-install** case, binds only when the `installation.created` webhook `sender` equals the kickoff First Tree admin. It does **not** support — and neither did the pre-#1381 code (the old `verifyUserCanAdministerInstallation` probe required the initiator to be a current GitHub admin) — the following required flow:

## Requirement

Support: a **First Tree org admin who is NOT a GitHub org owner** initiates the App install on a GitHub org → GitHub sends an approval request to the org's owners → an **owner approves** → the resulting installation binds to the initiator's First Tree org.

## Why it's non-trivial

In the approval flow the GitHub-side authorizer (the approving **owner**) is a different identity from the initiating **First Tree admin (member)**. So neither anchor used so far holds:
- `installation.created` `sender` is (very likely) the **approving owner**, not the initiator → the `sender == kickoff` bind in #1381 won't fire.
- The initiator is a GitHub **member**, not admin → the `/claim` current-admin probe rejects them.

A secure design must correlate "this installation" ↔ "initiator + their First Tree org" **without** requiring the initiator to be the installer/current-admin, and **without** letting an arbitrary First Tree admin bind an installation they have no claim to (the forgery concern from #1381).

## First step — verify GitHub's approval-flow mechanics (docs are silent)

- Does the requesting member get redirected to the callback at **request** time, and with what params (`installation_id`? `setup_action=request`?), or is there no installation until approval?
- After an owner approves, does `installation.created` fire then, and who is the `sender` (approver vs requester)?
- Is the original **requester** recorded anywhere on the installation?

The correlation design depends on these; establish them empirically before designing.

## Also decide here

- The fate of `POST /claim` (currently the API-only orphan-recovery path, no UI, tracked in #318): whether the approval flow's recovery is `/claim` (which then must not require GitHub-admin for a member requester) or a different mechanism.

liuchao-001:
## Research phase 1 — GitHub approval-flow mechanics (findings)

Verified against GitHub docs + the canonical community thread ([community#42351](https://github.com/orgs/community/discussions/42351)). **GitHub gives us no clean auto-correlation for the approval flow:**

| Phase | What GitHub gives us |
|---|---|
[...]

PR #1456, opened 2026-07-06
fix(web): neutral GitHub install popup landing instead of false "Connected"

Gandy2025:
## Problem

The GitHub App install popup landing page (`/onboarding/connected`) unconditionally rendered a green **"Connected"** check. But that same popup is where GitHub's callback lands for **two different outcomes**:

1. a genuine install (owner installs directly), and
2. a **non-owner** First Tree admin's **approval-request bounce** — the admin isn't a GitHub org owner, so GitHub only records a request for an owner to approve (`setup_action=request`, no installation created, nothing connected yet).

For case 2 the green "Connected" is **false success**: the user is told they're connected when an owner hasn't approved anything.

## Change

Make the landing **neutral — no success claim**:

- `ArrowLeft` glyph in muted tokens (was a green `Check`)
- title **"Back to First Tree"** (was "Connected")
- body points the user to the tab they started from, which owns the authoritative status

The genuine install still gets its **"Connected as `<org>`"** confirmation on the origin tab; this popup auto-closes in 900ms and was never the success moment. Updated the one SSR-smoke assertion.

## Scope / follow-up

This is the popup half of the GitHub non-owner install-approval UX. The backend that auto-binds an approved request lands in **#1392 / #1422** (@liuchao-001). A follow-up will add a real **"waiting for an owner to approve"** state to the origin tab (needs a read-only request-status endpoint over #1422's `github_app_install_requests` table) so the non-owner flow reads correctly end-to-end.

## Verification

- `pnpm biome check` (changed files) — clean
- `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) — clean
- `pnpm --filter @first-tree/web test` — 1101 passed

### DIFF
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1261,10 +1261,13 @@ describe("page SSR smoke coverage", () => {
     expect(html).toContain("Install Node.js");
 
     // The install-popup landing page (auto-closes the script-opened tab; the
-    // effect is a no-op under SSR). Confirms it renders + carries role=status.
+    // effect is a no-op under SSR). It makes no success claim — success vs
+    // pending-approval is owned by the origin tab — so it reads neutral and
+    // just points the user back. Confirms it renders + carries role=status.
     const connectedHtml = renderPage(<GithubConnectedPage />);
-    expect(connectedHtml).toContain("Connected");
+    expect(connectedHtml).toContain("Back to First Tree");
     expect(connectedHtml).toContain("close this tab");
+    expect(connectedHtml).not.toContain("Connected");
     expect(connectedHtml).toContain('role="status"');
 
     expect(await renderOnboardingStep(<StepTeam />, { activeStep: "create-team" })).toContain(
--- a/packages/web/src/pages/onboarding/github-connected.tsx
+++ b/packages/web/src/pages/onboarding/github-connected.tsx
@@ -1,23 +1,29 @@
-import { Check } from "lucide-react";
+import { ArrowLeft } from "lucide-react";
 import { useEffect } from "react";
 
 /**
  * Landing page for the GitHub App install popup — shared by every surface that
  * opens the install in a new tab with `next=/onboarding/connected`: onboarding
  * connect-code, the Context tab build entry, and Settings → GitHub. GitHub's
- * callback lands the popup here; we confirm and auto-close the tab (it was
- * script-opened, so `window.close()` is allowed) while the original tab detects
- * the new installation via its poll and updates on its own. If the browser
- * refuses to close the tab, the message tells the user to close it. Copy is kept
- * surface-neutral ("the other tab will update") so it reads correctly whether the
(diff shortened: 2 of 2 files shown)


yuezengwu (approved):
Reviewed the popup landing half. The implementation target is clear: remove the false-success claim from `/onboarding/connected`, keep the popup neutral, and leave authoritative install status to the origin tab. That matches the GitHub App contract in the Context Tree: the callback/popup route cannot distinguish a real install from a non-owner approval-request bounce.

Core scope checked: this only changes the web popup icon/copy plus the SSR smoke assertion. No DB/schema/core data-structure changes.

I checked the three origin-tab entry points that pass `/onboarding/connected`; they still use it only for the script-opened popup path, while popup-blocked/full-page redirects return to the real origin surfaces. The existing “Connected as …” confirmation remains on the bound origin-tab states.

CI is green: Lint & Type Check, Test, CodeQL, Migrations, and related checks have passed. I could not rerun local `pnpm` checks in my review worktree because dependencies are not installed there (`node_modules` missing), so I relied on static review plus GitHub CI.

Non-blocking nit: `packages/server/src/api/auth/github.ts` still calls this route the “Connected” sentinel in an error-path comment. Behavior is still right because errors should not recover to an auto-closing popup page; the wording is just stale after this PR.

baixiaohang (approved):
## Review by codex-assistant - PR #1456

**Recommendation**: approve
- Rationale: The popup now avoids a false success claim and leaves authoritative GitHub installation status to the origin tab, which matches the GitHub App flow and Context Tree ownership boundaries.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**`.
- Semantic lift: none; the change does not touch web subscriptions, WebSocket frame parsing, data-layer contracts, server behavior, or schema.

**PR Summary**
- Problem: A non-owner First Tree admin who requested GitHub App approval could land on a popup that falsely said “Connected,” even though no owner had approved the request and no installation was bound.
- Approach: The script-opened callback landing page now renders neutral “Back to First Tree” copy with a muted back icon, while the tab that initiated the install remains responsible for polling and displaying the real connected or pending status.
- Impacted modules: `packages/web/src/pages/onboarding/github-connected.tsx`, `packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx`.

**Review findings**
✅ No blocking findings. I checked the changed popup, the SSR smoke assertion, the three origin-tab entry points that use `/onboarding/connected`, and the callback behavior around missing `installation_id`; the popup is still only a neutral return/auto-close surface and does not become a source of truth for install state.

**Action taken**
- Approved.

Gandy2025:
Thanks both. On the shared non-blocking nit — the stale "Connected sentinel" wording in `packages/server/src/api/auth/github.ts` (the error-path `redirectCallbackError` comment): I'm intentionally leaving it out of this PR to keep the change web-only, and because that exact file is being rewritten in #1422 (the non-owner approval-flow backend). I've flagged the reword to @liuchao-001 to fold in there so we touch `github.ts` once. Behavior is unaffected either way — errors must not recover to the auto-closing popup, which this PR doesn't change.

PR #1459, opened 2026-07-07
Add guardrail prompt for landing trial agents

baixiaohang:
## Summary

- Add a server-owned guardrail prompt for landing campaign trial agents.
- Bind the prompt idempotently as an inline agent prompt whenever a landing campaign trial agent is created or reused.
- Cover prompt binding, runtime prompt projection, and reuse/idempotency in landing campaign start tests.

## Verification

- `pnpm --filter @first-tree/server test -- src/__tests__/landing-campaigns-start.test.ts --maxWorkers=2 --minWorkers=1`
  - The server test script ran the full `@first-tree/server` Vitest suite: 178 files / 1722 tests passed.
- `pnpm exec biome check packages/server/src/services/resources.ts packages/server/src/services/landing-campaigns/start.ts packages/server/src/services/landing-campaigns/trial-prompt.ts packages/server/src/__tests__/landing-campaigns-start.test.ts`
- `git diff --check`

## Review

- Reviewed by codex-reviewer in First Tree chat; no blocking issues found.

### DIFF
--- a/packages/server/src/__tests__/landing-campaigns-start.test.ts
+++ b/packages/server/src/__tests__/landing-campaigns-start.test.ts
@@ -19,6 +19,10 @@ import { createAgent } from "../services/agent.js";
 import { bindAgentRuntimeSession, validateAgentRuntimeSession } from "../services/agent-runtime-session.js";
 import { signTokensForUser } from "../services/auth.js";
 import { completeLandingCampaignTrialAgentTurn } from "../services/landing-campaigns/chat-state.js";
+import {
+  LANDING_CAMPAIGN_TRIAL_PROMPT,
+  LANDING_CAMPAIGN_TRIAL_PROMPT_RESOURCE_NAME,
+} from "../services/landing-campaigns/trial-prompt.js";
 import { createMember } from "../services/member.js";
 import { sendMessage } from "../services/message.js";
 import { uuidv7 } from "../uuid.js";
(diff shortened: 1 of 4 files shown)


yuezengwu (changes requested):
目标和核心改动：这个 PR 给 landing campaign trial agent 增加一段服务端定义的 trial guardrail prompt，并在 start/reuse trial agent 时把它作为 inline prompt binding 绑定到 agent；测试覆盖了 prompt binding、runtime prompt projection，以及重复 start 时的幂等性。

Blocking finding:

- `packages/server/src/services/resources.ts:1193` - The guardrail is inserted as the standalone per-agent inline prompt fragment (`type: "prompt"`, `mode: "include"`, `resourceId: null`, `replacesResourceId: null`). That is the exact row `resolveRuntimeConfig` projects with `editable: true` and the CLI `agent config prompt set` owns/removes. So the “server-owned guardrail” will render under `# Agent Prompt (this agent only — editable)` and can be replaced or cleared by the normal per-agent prompt edit path, including while a reused trial agent is running. That weakens the guardrail/governance semantics this PR is adding. Please model this as a non-editable managed prompt source/binding, or otherwise keep the trial guardrail out of the standalone editable prompt fragment path.

No DB/schema change is introduced by this PR.

baixiaohang:
Addressed the blocking review in `558417b2`.

The trial guardrail no longer uses the standalone editable inline prompt fragment path. It is now provisioned as an agent-scoped managed prompt resource and the agent binding points at that resource:

- resource: `type='prompt'`, `scope='agent'`, `ownerAgentId=<trialAgentId>`
- binding: `resourceId=<managed prompt resource id>`, `inlinePromptBody=null`
- runtime projection: `scope: 'agent'`, `editable: false`

I also added cleanup for stale copies of the previous inline guardrail body so a reused trial agent will not keep both the old editable fragment and the new managed prompt resource. The legacy append heading for agent-scoped managed prompt rows now renders as `Agent Prompt Override: <name>` instead of `Team Prompt: <name>`.

Verification:

- `pnpm exec biome check packages/server/src/services/resources.ts packages/server/src/services/landing-campaigns/trial-prompt.ts packages/server/src/__tests__/landing-campaigns-start.test.ts`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/landing-campaigns-start.test.ts --maxWorkers=2 --minWorkers=1`
- `git diff --check`

yuezengwu (approved):
Re-reviewed `558417b2`.

The previous blocker is addressed: the trial guardrail is now an agent-scoped managed prompt resource, the binding points at that resource with `inlinePromptBody: null`, and runtime projection marks the section as `editable: false`. The stale standalone inline guardrail cleanup also covers reuse from the earlier implementation shape.

Non-blocking note: `replaceAgentResources` still only admits agent-scoped `repo` and `skill` resource bindings in `validateBindingReference`; the public trial-agent PATCH route rejects trial-agent mutations before that path, so I don't see this as blocking this PR. If agent-scoped prompt resources become editable/round-trippable outside this server-managed trial path, that validator and a round-trip test should be widened too.

I could not rerun the local Vitest command in this review worktree because package binaries are not installed here (`vitest` not found). I reviewed the diff and GitHub checks are green except the Test job was still in progress at the time of re-review.

PR #1450, opened 2026-07-06
feat(web): render landing-campaign trial chat in the normal workspace

Gandy2025:
Closes #1449.

## Problem

The landing-campaign (First Tree Scan) trial first chat rendered as a **bespoke standalone page** — `QuickstartTrialChat`, a bare `min-h-screen` div wrapping a stripped `ChatByIdView` at `/quickstart?chat=<id>`, kept outside the workspace `Layout` and the onboarding gate. It was a divergent second chat surface, a dead-end (no path into the product), and it contradicted the Landing Campaign Trial Runtime model that deliberately makes the trial an ordinary org + ordinary agent + ordinary chat.

## What changed

Render the trial chat inside the **real workspace** (full chrome) while keeping the funnel's independent, gate-free `/quickstart` route:

- **Split `WorkspacePage`** into the onboarding gate + a shared exported `WorkspaceBody` (the three-pane shell + URL-backed rail state). `/` runs the gate then the body; the trial route renders `WorkspaceBody` **without** the gate. The gate skip lives at the route/caller level — `shouldEnterOnboarding` stays a **pure, campaign-agnostic** function (untouched).
- **Move `/quickstart` inside the `<Layout>` route group** and give it the full-bleed workspace outlet (`layout.tsx`), so the trial chat fills the viewport instead of the 960px admin canvas.
- **Unify chat selection on `?c=`**; delete the bespoke `QuickstartTrialChat`. When no chat is selected, show the real workspace conversation rail rather than a dead-end card.
- **Skip the `/login` interstitial for the scan funnel**: an unauthenticated visitor to `/quickstart?campaign=<known>` is sent straight to GitHub OAuth (`require-auth.tsx`), returning to the exact quickstart URL. That page is an extra click and off-message ("Set up your team…") for someone who clicked "scan my repo". Narrow: only `pathname === "/quickstart"` + a known campaign; everything else still uses `/login`.

The trial read-only lock stays metadata-driven (`isLandingCampaignTrialChatLocked`). **Server OAuth (`shouldPreserveSoloSignupNext`) is unchanged.**

## Guardrails honored

- No campaign branch in `shouldEnterOnboarding`.
- Campaign start writes no `onboardingCompletedAt` / suppression.
- Hosted trial agent is not treated as the user's real personal agent.
- Standard onboarding is entered only when the user explicitly acts.
- OAuth `next` is always same-origin `/quickstart` (+ `encodeURIComponent` + server `safeRedirectPath`) — no open redirect.

## Review

[...]

PR #1460, opened 2026-07-07
feat(web): trial chrome — a pure conversation for the scan trial surface

Gandy2025:
Follow-up to #1450.

## Problem

#1450 renders the landing-campaign Scan trial chat inside the **normal workspace** at `/quickstart`. That gave a controlled, single-run trial user the workspace's full set of **escape hatches** — clicking "Workspace" bounced them to `/onboarding`, and nav tabs / team switcher / ⌘K / new-chat / add-participant / agent pause-resume / chat rename / the message-avatar hovercard all let them wander out of the trial. The Context Tree makes the trial chat the *only* supported user-facing surface, so these need to be gone.

## Change

A **route-scoped trial chrome** (`isLandingTrialSurface(pathname) === "/quickstart"`, `pages/quickstart/route.ts`) that makes the trial a **pure conversation**:

- **Layout** — stripped header: brand + one **"Set up First Tree" CTA** (→ `/onboarding`) + theme + user menu (sign out). No nav tabs, team switcher, or ⌘K palette (the keydown effect and the palette render are both gated off).
- **WorkspaceBody** — drop the conversation rail (and its new-chat / filters / narrow overlay).
- **ChatView** — hide the entire chat-header controls cluster (participant stats, add participant, chat-details toggle); don't render the right details sidebar (also defends a persisted `showSidebar=true`); make the title non-renamable; render each message's sender avatar/name as **plain identity** (no `AgentHovercard`, whose "Open details" → `/agents/:id` and "Chat" → `/?c=draft` actions navigated out of the trial).
- **NoChatView** — hide "New chat"; the empty state points at the header CTA (not "the list", which no longer exists on trial).

**Route-scoped only.** `shouldEnterOnboarding`, the trial agent identity, and chat rendering/width are untouched. Normal workspace chats are unaffected (verified).

## Review

Two rounds of two independent reviews. The second round caught a real high-severity hole — the per-message `AgentHovercard` navigated out of the trial on every agent turn — now fixed (plain sender identity on trial) and E2E-confirmed. All other affordances (composer slash/mention, attach, agent markdown links via `isNavigableWebHref`, doc-preview drawer, user menu) were audited and are benign. Known boundary: chrome is keyed on the route, so `/quickstart?c=<any>` strips and a trial chat at `/?c=` would not (saved by the onboarding gate).

## Verification

E2E against real staging data with this branch's code (local Vite → staging API): trial header (desktop + mobile), no rail / palette / sidebar / header controls, non-renamable title, message avatar opens nothing, CTA → `/onboarding`; control `/` and `/?c=` keep full chrome + all affordances. `pnpm --filter @first-tree/web typecheck` + `lint:tokens` + biome clean; **1105 web tests pass** (new: layout trial chrome, rail drop, NoChatView trial — all with non-trial controls).

### DIFF
--- a/packages/web/src/components/__tests__/layout-dom.test.tsx
+++ b/packages/web/src/components/__tests__/layout-dom.test.tsx
@@ -257,6 +257,31 @@ describe("Layout", () => {
     await act(async () => context.root.unmount());
   });
 
+  it("renders trial chrome on /quickstart: no nav tabs / switcher / palette, one conversion CTA", async () => {
+    const trial = await renderLayout("/quickstart");
+    // Escape hatches are gone: nav tabs, team switcher, and the ⌘K palette entry.
+    expect(trial.container.textContent).not.toContain("Context");
+    expect(trial.container.textContent).not.toContain("Settings");
+    expect(trial.container.textContent).not.toContain("Workspace");
+    expect(trial.container.querySelector('[data-testid="team-switcher"]')).toBeNull();
+    expect(trial.container.querySelector('button[aria-label="Jump to… (⌘K)"]')).toBeNull();
+    // The one intentional way out: a "Set up First Tree" CTA → /onboarding.
+    expect(trial.container.textContent).toContain("Set up First Tree");
+    const cta = [...trial.container.querySelectorAll("a")].find((a) => /Set up First Tree/.test(a.textContent ?? ""));
+    expect(cta?.getAttribute("href")).toBe("/onboarding");
+    // The trial child still renders (full-bleed outlet, no 960 canvas).
+    expect(trial.container.textContent).toContain("Quickstart child");
+    await act(async () => trial.root.unmount());
+
+    // Control: an ordinary route keeps the full chrome and shows no CTA.
+    const context = await renderLayout("/context");
+    expect(context.container.textContent).toContain("Context");
+    expect(context.container.querySelector('[data-testid="team-switcher"]')).not.toBeNull();
+    expect(context.container.querySelector('button[aria-label="Jump to… (⌘K)"]')).not.toBeNull();
+    expect(context.container.textContent).not.toContain("Set up First Tree");
+    await act(async () => context.root.unmount());
+  });
+
   it("keeps status chips in the right controls before the compact command palette entry", async () => {
     disconnectMock.value = {
       firstHostname: "Yue-MacPro.local",
--- a/packages/web/src/components/layout.tsx
+++ b/packages/web/src/components/layout.tsx
@@ -1,16 +1,18 @@
-import { Search } from "lucide-react";
+import { ArrowRight, Search } from "lucide-react";
 import { useEffect, useState } from "react";
-import { NavLink, Outlet, useLocation } from "react-router";
+import { Link, NavLink, Outlet, useLocation } from "react-router";
 import { useAuth } from "../auth/auth-context.js";
 import { useNewVersionAvailable } from "../hooks/use-version-check.js";
 import { useWorkspaceViewport } from "../hooks/use-viewport.js";
 import { cn } from "../lib/utils.js";
+import { isLandingTrialSurface } from "../pages/quickstart/route.js";
 import { CommandPalette } from "../pages/workspace/palette/command-palette.js";
 import { DisconnectChip } from "./disconnect-chip.js";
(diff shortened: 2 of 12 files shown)


yuezengwu (changes requested):
Requesting changes for one route-scoped trial-chrome gap.

Required: `packages/web/src/pages/workspace/center/chat-by-id.tsx:202`

`ChatByIdView` forwards `isTrial` to `ChatView` only in the normal speaker branch. If `/quickstart?c=<chat>` resolves with `viewerMembershipKind === "watching"`, the read-only `ChatView` still receives `isTrial=false`, so the trial route can render the participant/details cluster/sidebar toggle and per-message `AgentHovercard` actions (`Open details` / `Chat`). That breaks the route-scoped guarantee this PR is adding: `/quickstart?c=<any>` should stay a pure conversation surface.

Please pass `isTrial={isTrial}` in the watcher branch too, and add a regression test for the watcher/read-only path if practical.

Gandy2025:
Good catch @yuezengwu — fixed in a4d01851.

`ChatByIdView`'s watcher branch (`chat-by-id.tsx:202`) now forwards `isTrial={isTrial}` too, so `/quickstart?c=<any>` stays a pure conversation even when the viewer resolves as `watching` (the read-only `ChatView` gets `isTrial=true`, which hides the participant/details cluster, the sidebar toggle, and the per-message hovercard — `readOnly` alone did not cover the hovercard).

Added a regression: `ChatView` with `isTrial + readOnly` asserts Add participant / chat-details toggle / rename are all absent (even with a persisted-open sidebar). 1106 web tests pass; typecheck + lint clean. Re-requesting review.

yuezengwu (approved):
Approved after re-reviewing `a4d01851`.

The prior blocker is fixed: the watcher/read-only branch now forwards `isTrial={isTrial}` into `ChatView`, so `/quickstart?c=<any>` keeps the same trial chrome restrictions even when the viewer resolves as `watching`. The added `isTrial + readOnly` regression also covers the header controls/sidebar/rename absence with a persisted-open sidebar preference.

yuezengwu (approved):
Approved after re-reviewing `a4d01851`.

The prior blocker is fixed: the watcher/read-only branch now forwards `isTrial={isTrial}` into `ChatView`, so `/quickstart?c=<any>` keeps the same trial chrome restrictions even when the viewer resolves as `watching`. The added `isTrial + readOnly` regression also covers the header controls/sidebar/rename absence with a persisted-open sidebar preference.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1460

- Rationale: One route-scoped trial escape hatch remains: the inline offline-agent notice can still navigate a `/quickstart` user to the Computers settings page.

**Risk level**: C
- Semantic lift: none; I did not find subscription, WS frame parsing, or data-layer changes.

**PR summary**
- Problem: Users entering the Scan trial through `/quickstart` were seeing normal workspace controls that let them leave the intended single conversation before entering the product setup flow.
- Approach: The PR keys trial chrome off the `/quickstart` route, strips global workspace navigation and the conversation rail, forwards `isTrial` into the center chat path, removes chat-management affordances and per-message agent hovercards from trial chats, and leaves one explicit onboarding CTA.
- Impacted modules: `packages/web/src/components/layout.tsx`, `packages/web/src/pages/quickstart/*`, `packages/web/src/pages/workspace/**`

**Review findings**
❌ 1. `ChatTimeline` still renders `<ChatOfflineNotice>` on trial chats, and that notice's `Reconnect` button calls `navigate("/settings/computers")`. If the trial conversation is awaiting an offline agent, `/quickstart?c=<chat>` shows a visible route out to the Computers settings page from inside the supposedly pure conversation surface. Please gate this notice/action for `isTrial` or render trial-safe copy/action instead. [`packages/web/src/pages/workspace/center/chat-view.tsx:1268`, `packages/web/src/components/chat/chat-offline-notice.tsx:136`]

**Action taken**
- Submitted request changes.

Gandy2025:
Good catch @baixiaohang — the offline-notice "Reconnect" → /settings/computers is a real route-scoped escape hatch. #1460 is already merged, so fixed in follow-up **#1461**: the offline-agent notice is gated on `isTrial` (not rendered on the trial surface). Requesting your review there.

PR #1461, opened 2026-07-07
fix(web): hide the offline-agent notice on the trial surface

Gandy2025:
Follow-up to #1460 — addresses @baixiaohang's post-merge review finding.

## Problem

`ChatTimeline` still rendered `<ChatOfflineNotice>` on trial chats. When the trial conversation awaits an offline agent, that notice's **"Reconnect"** button calls `navigate("/settings/computers")` (`chat-offline-notice.tsx:136`) — a visible route **out of** the controlled `/quickstart` pure-conversation surface, and meaningless for a service-managed trial agent the user has no computer to reconnect.

## Change

Gate the notice on `isTrial` (`chat-view.tsx`): the offline-agent notice is not rendered on the trial surface. A hosted trial agent's liveness is First Tree's concern, not a user "reconnect a computer" action.

Route-scoped, one render guard; normal chats keep the notice unchanged.

## Verification

`typecheck` + `lint:tokens` + biome clean; **1106 web tests pass**. (The offline state can't be forced on staging to E2E, and seeding a specific offline+awaited timeline in a unit test is impractical; this is a straightforward render guard on the already-threaded `isTrial`.)

### DIFF
--- a/packages/web/src/pages/workspace/center/chat-view.tsx
+++ b/packages/web/src/pages/workspace/center/chat-view.tsx
@@ -1265,7 +1265,11 @@ const ChatTimeline = memo(function ChatTimeline({
               </div>
             ) : null}
           </div>
-          <ChatOfflineNotice chatId={chatId} agents={awaitedAgents} />
+          {/* No offline-agent notice on the trial surface: its "Reconnect"
+              action navigates to /settings/computers (an escape hatch), and a
+              service-managed trial agent has no computer for the user to
+              reconnect anyway — liveness is First Tree's concern here. */}
+          {isTrial ? null : <ChatOfflineNotice chatId={chatId} agents={awaitedAgents} />}
           <div ref={messagesEndRef} />
         </div>
       </div>


yuezengwu (approved):
Approved.

Goal/core change: this follow-up fixes the post-merge #1460 trial escape hatch by gating `ChatOfflineNotice` out of `ChatTimeline` when `isTrial` is true. That removes the `Reconnect` action that navigates to `/settings/computers` from the `/quickstart` pure-conversation surface, while leaving ordinary workspace chats unchanged.

No DB/schema or core data-structure changes. I did not find a remaining blocker in the one-file diff.

baixiaohang (approved):
## Review by codex-assistant - PR #1461

**Recommendation**: approve
- Rationale: This narrowly fixes the remaining trial-surface escape hatch by suppressing the offline-agent reconnect notice only when `isTrial` is true, while preserving normal workspace recovery behavior.

**Risk level**: C
- Semantic lift: none; no subscription, WS frame parsing, or data-layer changes.

**PR summary**
- Problem: After #1460, a `/quickstart` trial chat could still show an offline-agent `Reconnect` action that navigated to `/settings/computers`, which breaks the pure conversation surface and is not meaningful for a service-managed trial agent.
- Approach: `ChatTimeline` now keeps the existing offline notice for normal chats but skips it on the route-scoped trial surface using the already-threaded `isTrial` prop.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`

**Review findings**
✅ 1. The guard is placed at the only `ChatOfflineNotice` render site, and `isTrial` is already passed from `ChatView` into `ChatTimeline`, so both normal and watcher trial chat paths are covered.
✅ 2. Normal workspace chats keep the existing offline recovery affordance because the notice still renders when `isTrial` is false.

**Action taken**
- Approved.

PR #1463, opened 2026-07-07
feat(web): minimal trial composer — no mention / attach / slash

Gandy2025:
Follow-up to #1460.

## Change

On the landing-campaign trial surface, the composer is reduced to **just type + send** — @mention, image attach, and slash commands are not needed for a single-agent, controlled conversation. Gated on `isTrial` (route-scoped `/quickstart`), normal chats keep everything:

- **ChatView composer**: hide the `@mention` + attach buttons and the hidden file input; skip the mention/slash keydown handlers and autocomplete popovers; no image paste / drag-drop; never prime `@` on focus; placeholder simplified to `Message @<agent>`.
- **AskTakeover** (the blocking-ask answer card): the same treatment — no @mention/attach toolbar, no mention popover/keys, no image paste — so a `chat ask` overlay can't re-expose what the composer removed.

The `MentionHighlightOverlay` text mirror stays, so typed text remains visible; the trial send/answer path (Enter + Send button) is unaffected.

## Review

Two independent pre-commit reviews. One confirmed the composer gating is complete with no non-trial regression (Enter-send intact, text visible, JSX balanced). The other caught that the **AskTakeover answer surface** was not gated and would re-expose @mention + attach on a blocking ask — fixed here, plus a focus-prime hardening (`isTrial` never primes `@`).

## Verification

E2E on real staging data with this branch (local Vite → staging): the trial composer shows only the textarea + Send (no @/attach), placeholder `Message @Production Scanner`, and typing `@`/`/` inserts literally with no popup; control `/?c=` normal chat keeps mention/attach/slash + the full placeholder. `typecheck` + `lint:tokens` + biome clean; **1106 web tests pass**.

### DIFF
--- a/packages/web/src/components/chat/ask-takeover.tsx
+++ b/packages/web/src/components/chat/ask-takeover.tsx
@@ -97,7 +97,11 @@ export function AskTakeover({
   error,
   onReply,
   onSkip,
+  isTrial = false,
 }: {
+  /** Trial surface: match the minimal trial composer — no @mention / attach
+   *  affordances in the answer input, just plain text. */
+  isTrial?: boolean;
   /** The ask itself — the request message's markdown body. */
   body: string;
   payload: AskRequest;
@@ -284,13 +288,16 @@ export function AskTakeover({
    *  image paste, with the text drawn by the mirror overlay. */
   const renderAnswerInput = (placeholder: string, minHeight: number) => (
     <div style={fieldChrome}>
-      <MentionAutocompletePopover
-        trigger={mention.trigger}
-        results={mention.results}
-        highlightIndex={mention.highlightIndex}
-        anchorRef={taRef}
-        onPick={mention.pick}
-      />
+      {/* No mention autocomplete on the trial answer input (single agent). */}
+      {isTrial ? null : (
+        <MentionAutocompletePopover
+          trigger={mention.trigger}
+          results={mention.results}
+          highlightIndex={mention.highlightIndex}
+          anchorRef={taRef}
+          onPick={mention.pick}
+        />
+      )}
       <MentionHighlightOverlay
         value={freeText}
         participants={mentionParticipants}
@@ -312,6 +319,9 @@ export function AskTakeover({
         }}
         onSelect={(e) => setCursor(e.currentTarget.selectionStart ?? freeText.length)}
         onPaste={(e) => {
+          // No image attachments on the trial answer input — let text paste
+          // fall through to the default handler.
+          if (isTrial) return;
           const files = Array.from(e.clipboardData.files);
           if (files.length > 0) {
             e.preventDefault();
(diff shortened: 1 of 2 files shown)


yuezengwu (changes requested):
Goal / core change reviewed: the `/quickstart` trial surface should reduce the composer and blocking-ask answer card to plain type + send, while normal workspace chats keep mention, attach, slash, and expressive ask-answer behavior.

- `packages/web/src/components/chat/ask-takeover.tsx:531`: trial mode still leaves the answer surface as an unconditional image drop target. Even when `isTrial` hides the attach toolbar and disables image paste, this wrapper still calls `preventDefault()` on dragover and `addImages(Array.from(e.dataTransfer.files))` on drop. In a trial blocking ask, dragging an image into the card will stage a thumbnail, enable Reply through `pendingImages.length > 0`, and send the image via `onReply`. That re-exposes exactly the attachment capability this PR is removing from the trial surface. Please gate the drop handler/image tray/image-only reply path for `isTrial`, or at least prevent/ignore file drops without staging images in trial mode.

Gandy2025:
Good catch @yuezengwu — fixed in the latest push. The ask-takeover outer wrapper's `onDrop`/`onDragOver` are now gated on `isTrial` too, so all three image-add paths (paste, file-input, drop) are off on trial → `pendingImages` stays empty, so the image tray and image-only Reply path are inert. 1106 tests pass; typecheck + lint clean. Re-requesting review.

yuezengwu (approved):
Approved on re-review. The previous blocker is fixed: `AskTakeover` trial mode now gates the drop target before `addImages`, matching the paste/file-input gating so the trial answer surface no longer stages or sends image attachments. I rechecked the surrounding trial composer/answer paths and did not find another blocking issue.

baixiaohang (approved):
## Review by codex-assistant - PR #1463

**Recommendation**: approve
- Rationale: Latest head keeps the minimal `/quickstart` trial composer and ask-answer surface route-scoped, and closes the remaining attachment path without changing normal chat affordances.

**Risk level**: C
- Path baseline: only `packages/web/**`.
- Semantic lift: none; UI-only composer/answer gating, no subscription, WS, or data-layer changes.

**PR Summary**
- Problem: The landing-campaign trial chat exposed mention, slash-command, and image-attachment affordances that are unnecessary for a single-agent controlled trial, including on the blocking-ask answer card.
- Approach: Gate `ChatView` and `AskTakeover` on `isTrial` so `/quickstart` keeps type + send/reply only while ordinary workspace chats keep mention autocomplete, slash commands, attachments, drag/drop, and expressive ask answers.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`, `packages/web/src/components/chat/ask-takeover.tsx`.

**Review findings**
✅ 1. The previous `AskTakeover` drop-target blocker is fixed: trial mode now gates drop before `addImages`, matching paste/file-input gating, so pending images cannot enable image-only replies from the trial answer card.
✅ 2. Normal workspace chats continue through the existing non-trial paths for mention/slash popovers, toolbar buttons, hidden file inputs, and image drop/paste handlers.

**Action taken**
- Approved.

PR #1465, opened 2026-07-07
fix(web): keep the trial composer send button right-aligned

Gandy2025:
Follow-up to the minimal trial composer (#1463) — reported by gandy testing on staging.

## Problem

The composer footer is a `justify-between` row: left = @/attach cluster, right = send cluster. #1463 gated the left cluster out on the trial surface, which left the row with a single child — and `justify-between` then **left-aligns** it. So the trial send button rendered at the bottom-**left** instead of bottom-right (screenshot from gandy confirmed).

## Fix

Add `ml-auto` to the send cluster so it stays hard-right whether or not the left cluster is present. Normal workspace chats are unaffected (left cluster at start + `ml-auto` send = same right-aligned result as `justify-between`).

## Verification

`typecheck` + biome clean; **1106 web tests pass**. E2E on this branch vs real staging data: the trial composer's send button is now at the bottom-right (measured `x≈1063` in a 1280 viewport). (Position isn't assertable in jsdom, so this is E2E-verified rather than unit-tested.)

### DIFF
--- a/packages/web/src/pages/workspace/center/chat-view.tsx
+++ b/packages/web/src/pages/workspace/center/chat-view.tsx
@@ -4288,7 +4288,10 @@ export function ChatView({
                             />
                           </span>
                         )}
-                        <span className="flex items-center" style={{ gap: 8 }}>
+                        {/* `ml-auto` keeps the send cluster hard-right even when the
+                            left @/attach cluster is gone (trial surface) — otherwise
+                            `justify-between` with a single child left-aligns it. */}
+                        <span className="flex items-center ml-auto" style={{ gap: 8 }}>
                           {uploading && (
                             <span className="mono text-caption" style={{ color: "var(--primary)" }}>
                               uploading…


yuezengwu (approved):
Reviewed the goal and core change: #1463 removed the trial composer’s left @/attach cluster, leaving the footer row with only the send cluster; adding `ml-auto` to that send cluster keeps it right-aligned in both trial and normal workspace chat surfaces.

No database, schema, permission, runtime, or core data-structure changes in this PR. I checked the surrounding `ChatView` footer layout and the flex behavior is correct for both one-child and two-child cases. `git diff --check` is clean; GitHub lint/type checks are passing, with the broader test job still pending.

No blockers found.

baixiaohang (approved):
## Review by codex-assistant - PR #1465

**Recommendation**: approve
- Rationale: This is a narrowly scoped flex-layout fix that keeps the send controls right-aligned when the trial composer hides the left controls, without changing chat send semantics.

**Risk level**: C
- Path baseline: `packages/web/**` only -> C.
- Semantic lift: none; this is UI layout only, not subscriptions, WebSocket frame parsing, or data-layer behavior.

**PR Summary**
- Problem: Trial users could see the composer send button at the lower-left after the minimal trial composer removed the left @/attach controls.
- Approach: Add `ml-auto` to the send cluster so it remains hard-right whether the left cluster is present or hidden; normal workspace chats keep the same visible alignment.
- Impacted modules: `packages/web/src/pages/workspace/center/chat-view.tsx`.

**Review findings**
✅ 1. No blocking findings. The updated flex margin matches the footer's conditional children and does not alter the chat send, mention-gate, or attachment logic. [`packages/web/src/pages/workspace/center/chat-view.tsx`:4294]

**Action taken**
- Approved.

PR #1469, opened 2026-07-07
Fix landing trial ended copy and SSH command

baixiaohang:
## Summary
- update the locked landing trial chat placeholder to point users to the "Set up First Tree for your team" top-page CTA
- add a landing-trial-only controlled `GIT_SSH_COMMAND` with `-F /dev/null` so OpenSSH ignores host `/etc/ssh/ssh_config*`
- keep GitHub SSH operations on the explicit host trial key, `IdentitiesOnly=yes`, workspace-local `known_hosts`, and `StrictHostKeyChecking=accept-new`
- cover the SSH command in unit tests and add an opt-in GitHub SSH smoke check

## Validation
- `git diff --check`
- `pnpm --filter @first-tree/client exec vitest run --passWithNoTests --maxWorkers=2 src/__tests__/workspace-sandbox.test.ts src/__tests__/codex-app-server-handler.test.ts`
- `pnpm check` (passes with existing warnings in unrelated files)
- `pnpm typecheck`
- manual: `GIT_SSH_COMMAND="ssh -F /dev/null -i '$HOME/.ssh/id_ed25519' ..." git ls-remote git@github.com:agent-team-foundation/first-tree.git HEAD`

### DIFF
--- a/packages/client/src/handlers/codex/app-server/workspace-sandbox.ts
+++ b/packages/client/src/handlers/codex/app-server/workspace-sandbox.ts
@@ -76,13 +76,13 @@ const READ_ONLY_ETC_PATHS = [
 const LANDING_CODEX_ROOT_READ_PATH = ":root";
 export const LANDING_CODEX_HOST_CREDENTIAL_DENY_RELATIVE_PATHS = [
   ".codex",
-  ".ssh",
   ".first-tree",
   ".first-tree-staging",
   ".first-tree-dev",
   ".first-tree-local",
   ".first-tree-test",
 ] as const;
+const LANDING_CODEX_GIT_SSH_KEY_RELATIVE_PATH = join(".ssh", "id_ed25519");
 const WORKSPACE_ONLY_PATH_DIRS = ["/usr/local/bin", "/usr/bin", "/bin"] as const;
 const SAFE_PASS_ENV_KEYS = new Set([
   "FIRST_TREE_HOME",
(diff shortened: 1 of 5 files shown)


packages/client/src/handlers/codex/app-server/workspace-sandbox.ts
```diff
@@ -76,13 +76,13 @@ const READ_ONLY_ETC_PATHS = [
...
 export const LANDING_CODEX_HOST_CREDENTIAL_DENY_RELATIVE_PATHS = [
   ".codex",
-  ".ssh",
   ".first-tree",
   ".first-tree-staging",
   ".first-tree-dev",
   ".first-tree-local",
   ".first-tree-test",
 ] as const;
+const LANDING_CODEX_GIT_SSH_KEY_RELATIVE_PATH = join(".ssh", "id_ed25519");
```
yuezengwu, line 92: Blocking: this exposes more than the intended trial key. The profile still grants `:root` read, and removing the `.ssh` deny means trial command execution can read any host SSH files by absolute path once it sees the key path in `GIT_SSH_COMMAND`; the command only constrains which identity `git`/`ssh` offers, not what the trial shell can read. The landing-trial boundary described in the tree/PR is an exception for the low-permission trial key, not the whole host SSH directory. Please keep host `.ssh` denied and expose only the intended key path, or copy/bind only that key into workspace-local trial state before pointing `GIT_SSH_COMMAND` at it.

yuezengwu (changes requested):
Goal/core changes: this PR updates the locked landing-trial composer copy to point at the top-page team setup CTA, and adds a landing-trial-only GIT_SSH_COMMAND that uses -F /dev/null, an explicit host SSH key, workspace-local known_hosts, IdentitiesOnly=yes, and accept-new host key behavior.

I am requesting changes for one credential-boundary issue. I did not see DB/schema or core data-structure changes.

Validation from my review worktree: git diff --check passed. I could not run the targeted Vitest command because dependencies are not installed in this fresh worktree; pnpm reported Command "vitest" not found under packages/client.

yuezengwu (approved):
Goal/core changes: this PR updates the locked landing-trial composer copy to point users at the top-page team setup CTA, and adds a landing-trial-only controlled Git SSH command with `-F /dev/null`, an explicit trial key, `IdentitiesOnly=yes`, workspace-local `known_hosts`, and `StrictHostKeyChecking=accept-new`.

I re-reviewed the synchronized head (`1b4fb3cc`). The follow-up commit makes the managed-host SSH boundary explicit in code: official landing-trial hosts intentionally keep `~/.ssh` readable, and that directory must contain only the low-permission GitHub trial keypair. That matches the Context Tree's hosted-trial assumption that official trial machines are low-sensitive public-repo hosts without unapproved high-value credentials outside the explicit deny set.

No DB/schema or core data-structure changes seen. `git diff --check` passed in my review worktree. I did not rerun Vitest because this fresh worktree has no installed dependencies; the earlier attempt reported `vitest` not found under `packages/client`.

PR #1467, opened 2026-07-07
feat(server): load landing trial production-scan skill from its external repo

Gandy2025:
## Summary

Landing-campaign trial agents now run the full external `production-scan` skill instead of the server-materialized distilled body:

- **Kickoff message** now uses the approved copy and instructs the trial agent to clone `agent-team-foundation/launch-readiness-scan` and run its `production-scan` skill on the connected repo, in First Tree trial mode. The skill repo URL and skill name live in the campaign registry (`catalog.ts`), not in the prose.
- **Delete the server-side skill delivery machinery**: the inline `PRODUCTION_SCAN_BODY` (incl. setup CTA + github-follow override), `ensureAndBindCampaignScanSkill` (agent-private resource provisioning, config-version bump plumbing, `{{FIRST_TREE_SETUP_URL}}` templating), the `campaign-scan-skill.ts` re-export shim, the `setupUrl` parameter end to end, and their tests. Net −800 lines.
- **`landingCampaignMaxAgentTurns` default 1 → 6** so the full external skill flow (calibration, scan report, fix offer, bounded follow-ups) fits in a trial without env overrides.

Review-round fixes on top:

- **Purge the legacy materialized skill on trial re-kickoff** (`unbindLegacyCampaignScanSkill` + integration test): pre-migration trial agents still carry the old agent-scoped `production-scan` resource + binding; the client only prunes materialized skills whose binding is gone, and the stale skill name-collides with the cloned one — a returning org would keep running the outdated rubric.
- Update `docs/development/onboarding-kickoff-contract.md` — the quickstart path no longer binds managed campaign skills.
- Make `maxAgentTurns` a required input to `buildLandingCampaignChatMetadata` (the optional `?? 1` fallback could silently stamp a 1-turn trial).
- Reword the agent-scoped-skill admission comment in `replaceAgentResources` validation to cite legacy pre-migration rows (the carve-out must stay until those rows are purged).

Supersedes #1466 (same core change; this branch additionally carries the turns default, the legacy-skill purge, the contract-doc sync, and the metadata hardening).

## Rollout notes

- Deployments that pinned `FIRST_TREE_LANDING_CAMPAIGN_MAX_AGENT_TURNS=4` should raise it to `6` or unset it to inherit the new default. Consider setting `FIRST_TREE_LANDING_CAMPAIGN_MAX_ESTIMATED_TOKENS` as the cost backstop now that trials default to 6 turns.
[...]

PR #1470, opened 2026-07-07
fix(server): align the landing kickoff value line with fix-as-conversion

Gandy2025:
## Summary

Follow-up to #1467/#1468, aligning the last stale phrase in the kickoff bootstrap with the external skill's fix-as-conversion flow ([launch-readiness-scan#24](https://github.com/agent-team-foundation/launch-readiness-scan/pull/24)):

- The trial no longer produces one ready-to-apply artifact in-conversation — the report carries an exact, copy-pasteable fix under **every** blocker, and applying them routes to building a First Tree team.
- Kickoff line: "…and a ready-to-use fix for the biggest one." → "…and the exact fix for each one."

One line in `catalog.ts`; no behavior change.

## Validation

- `pnpm check` / server `tsc --noEmit` green
- `landing-campaigns-start.test.ts`: 34 passed (no assertion pinned the old phrase)

### DIFF
--- a/packages/server/src/services/landing-campaigns/skills/catalog.ts
+++ b/packages/server/src/services/landing-campaigns/skills/catalog.ts
@@ -50,7 +50,7 @@ export function buildLandingCampaignBootstrap(skillSet: LandingCampaignSkillSet,
   return [
     `Welcome to First Tree. ${skillSet.agentDisplayName} is connected to your code: ${repoUrl}`,
     "",
-    "It's giving your code a safe, read-only check before launch — you'll get a score, the problems that actually matter, and a ready-to-use fix for the biggest one.",
+    "It's giving your code a safe, read-only check before launch — you'll get a score, the problems that actually matter, and the exact fix for each one.",
     "",
     `${skillSet.agentDisplayName} — clone ${skillSet.skillRepoUrl} and run its ${skillSet.skillName} skill on the repo above.`,
   ].join("\n");


yuezengwu (approved):
Reviewed the change goal and core diff: this aligns the landing campaign kickoff value line with the external production-scan fix-as-conversion flow, replacing the stale single-fix promise with exact fixes per blocker. The PR only changes the bootstrap copy in catalog.ts; no database, schema, or core data-structure change.

I checked the surrounding catalog context, searched for stale "ready-to-use fix" / "biggest one" wording, reviewed the related landing-campaign test assertion, and verified the referenced launch-readiness-scan#24 flow supports this copy. No blockers found.

Local targeted test was not runnable in my temporary review worktree because dependencies are not installed there (vitest missing), so I relied on static review plus the submitted validation/CI.

baixiaohang (approved):
## Review by codex-assistant - PR #1470

**Recommendation**: approve
- Rationale: The kickoff copy now promises the exact per-finding fixes that the scan report already carries, without implying an in-chat free fix path.

**Risk level**: B-low
- Path baseline: touches `packages/server/**`, but not the DB schema and not the listed core chat/message/session/participant/agent services.
- Semantic lift: none.

**PR summary**
- Problem: users starting the production-scan landing trial could still see a stale value line that framed the result as one ready-to-use fix for the biggest problem, while the current scan/report direction is per-blocker fix guidance and fix-as-conversion.
- Approach: updates the server-owned landing kickoff bootstrap line to describe a score, the material problems, and the exact fix for each one; the campaign still clones and runs the external `launch-readiness-scan` skill at runtime.
- Impacted modules: `packages/server/src/services/landing-campaigns/skills/catalog.ts`.

**Review findings**
✅ 1. No blocking issues found. The changed line matches the report contract I checked in the current external skill and in the `launch-readiness-scan#24` branch, where every blocker carries concrete `The fix` / `Verify` guidance.
✅ 2. The local landing-campaign test asserts the external skill clone/run instructions and does not pin the stale phrase, so this copy-only change does not leave a contradictory test contract.

**Action taken**
- Approved.

PR #1472, opened 2026-07-07
[codex] Allow delegate agent runtime switches

yuezengwu:
## Summary

Remove the runtime-switch precondition that forced members to clear `delegateMention` before switching the delegate agent's runtime.

## Why

`delegateMention` points from the human agent row to the delegate agent UUID. Runtime switching preserves the agent identity and chat membership while changing the client/provider route, so the delegate reference does not become dirty. Requiring users to clear it would interrupt GitHub/follow routing and force them to manually restore the delegate afterward.

## Changes

- Removed the `delegateMention === current.uuid` hard blocker from `switchAgentRuntime`.
- Added a server regression test covering a delegate agent switch while the human row keeps `delegateMention` pointing at that agent.
- The test also verifies the provider changes and the old runtime-session token is revoked.

## Validation

- `pnpm --filter @first-tree/server test -- agent-runtime-switch.test.ts` (ran server test suite: 177 files / 1720 tests passed)
- `pnpm exec biome check packages/server/src/services/agent-runtime-switch.ts packages/server/src/__tests__/agent-runtime-switch.test.ts`

### DIFF
--- a/packages/server/src/__tests__/agent-runtime-switch.test.ts
+++ b/packages/server/src/__tests__/agent-runtime-switch.test.ts
@@ -132,6 +132,71 @@ describe("POST /agents/:uuid/switch-runtime", () => {
     expect(events).toEqual([]);
   });
 
+  it("switches a delegate agent without clearing the member delegate mention", async () => {
+    const app = getApp();
+    const ctx = await createAdminContext(app);
+    await setClientRuntimeSupport(app, ctx.clientId, {
+      "claude-code": capability("ok"),
+      codex: capability("ok"),
+    });
+    const agent = await createAgent(app.db, {
+      name: `switch-delegate-${crypto.randomUUID().slice(0, 6)}`,
+      type: "agent",
+      displayName: "Switch Delegate",
+      managerId: ctx.memberId,
+      clientId: ctx.clientId,
+      runtimeProvider: "claude-code",
+    });
+    await app.db.update(agents).set({ delegateMention: agent.uuid }).where(eq(agents.uuid, ctx.humanAgentUuid));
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1472

**Recommendation**: approve
- Rationale: Removing the `delegateMention` blocker matches the durable delegate semantics: the switch preserves the same agent identity and chat membership while the existing runtime-switch path still revokes the old runtime session and moves the client/provider route.

**Risk level**: B-high
- Path baseline: touches `packages/server/**` and `packages/server/src/services/agent-runtime-switch.ts` under `services/agent*`.
- Semantic lift: none beyond the server agent runtime-switch flow already in scope.

**PR Summary**
- Problem: operators could not switch the runtime for their delegate agent without first clearing the member's delegate setting, which would disrupt GitHub/follow routing and require a manual restore afterward.
- Approach: allow the delegate pointer to remain stable while the managed runtime-switch flow changes the client/provider route; the code removes the `human.delegateMention === current.uuid` conflict check and adds a server regression test for the delegate case.
- Impacted modules: `packages/server/src/services/agent-runtime-switch.ts`, `packages/server/src/__tests__/agent-runtime-switch.test.ts`.

**Review findings**
✅ 1. `delegateMention` is a UUID reference to the same agent, and this switch flow does not alter the agent UUID, organization, or chat membership, so the durable delegate relation remains valid.
✅ 2. The existing committed switch side effects still revoke the old runtime session, detach the old runtime route, archive sessions, and clear the runtime-switch claim; removing this precondition does not bypass those guards.
✅ 3. The new regression test covers the important observable behavior: switch succeeds, the human row keeps `delegateMention`, the provider changes, reserved runtime metadata is cleared, and the old runtime-session token is rejected.

**Action taken**
- Approved.

PR #1474, opened 2026-07-07
[codex] Allow delegate agent runtime switches

yuezengwu:
## Summary

Remove the runtime-switch precondition that forced members to clear `delegateMention` before switching the delegate agent's runtime.

## Why

`delegateMention` points from the human agent row to the delegate agent UUID. Runtime switching preserves the agent identity and chat membership while changing the client/provider route, so the delegate reference does not become dirty. Requiring users to clear it would interrupt GitHub/follow routing and force them to manually restore the delegate afterward.

## Changes

- Removed the `delegateMention === current.uuid` hard blocker from `switchAgentRuntime`.
- Added a server regression test covering a delegate agent switch while the human row keeps `delegateMention` pointing at that agent.
- The test also verifies the provider changes and the old runtime-session token is revoked.

## Validation

- `pnpm --filter @first-tree/server test -- agent-runtime-switch.test.ts` (ran server test suite: 177 files / 1720 tests passed)
- `pnpm exec biome check packages/server/src/services/agent-runtime-switch.ts packages/server/src/__tests__/agent-runtime-switch.test.ts`

### DIFF
--- a/packages/server/src/__tests__/agent-runtime-switch.test.ts
+++ b/packages/server/src/__tests__/agent-runtime-switch.test.ts
@@ -132,6 +132,71 @@ describe("POST /agents/:uuid/switch-runtime", () => {
     expect(events).toEqual([]);
   });
 
+  it("switches a delegate agent without clearing the member delegate mention", async () => {
+    const app = getApp();
+    const ctx = await createAdminContext(app);
+    await setClientRuntimeSupport(app, ctx.clientId, {
+      "claude-code": capability("ok"),
+      codex: capability("ok"),
+    });
+    const agent = await createAgent(app.db, {
+      name: `switch-delegate-${crypto.randomUUID().slice(0, 6)}`,
+      type: "agent",
+      displayName: "Switch Delegate",
+      managerId: ctx.memberId,
+      clientId: ctx.clientId,
+      runtimeProvider: "claude-code",
+    });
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1474

**Recommendation**: approve
- Rationale: The removed blocker treated `delegateMention` as dirtyable runtime state, but runtime switching preserves the agent UUID and keeps the delegate reference valid while still revoking the old runtime session.

**Risk level**: B-high
- Path baseline: touches `packages/server/src/services/agent-runtime-switch.ts` under `packages/server/**` and an agent service file.

**PR Summary**
- Problem: Members whose delegate agent handles automated routing, such as GitHub mentions or follows, were forced to clear the delegate before switching that agent's runtime.
- Approach: Allow the managed runtime switch to proceed while preserving the human row's `delegateMention`; the switch still changes the client/provider route, clears runtime switch/session metadata, and revokes the stale runtime-session token.
- Impacted modules: `packages/server/src/services/agent-runtime-switch.ts`, `packages/server/src/__tests__/agent-runtime-switch.test.ts`

**Review findings**
✅ No blocking issues found. The delegate link points to stable agent identity, not to the runtime route, and the existing switch flow still handles route/session convergence.
✅ The added regression test covers delegate preservation, provider change, switch metadata cleanup, and stale runtime-session rejection.

**Action taken**

PR #1373, opened 2026-07-01
fix(web): daemon-first Reconnect dialog for offline computers

Gandy2025:
## Problem

Clicking **Reconnect** on an offline computer opened the fresh-connect "Connect computer" dialog — a full `npm install -g first-tree-staging` + `... login <token>` command with copy like *"Run this on the machine you want to **pair** with First Tree. If the CLI isn't installed yet…"*.

For a machine that is **already paired**, has the CLI installed, and is merely offline (the common cause is the background daemon not running), that command is too heavy (reinstall + a fresh single-use token) and the copy is self-contradictory. The correct lightweight fix — `<binName> daemon start` — existed only as a collapsed "Daemon not running?" disclosure buried in the offline card, while the prominent button led to the heavy flow.

## Change

Add `ReconnectDialog`, scoped to the offline machine (daemon-first, layered):

- **Primary:** `<binName> daemon start` — no reinstall, no connect token, no server round-trip. Covers the by-far most common offline cause.
- **Fallback:** the full `npm install -g … && … login <token>` (identical to `+ Connect`) is demoted to a lazily-minted **"Still offline?"** disclosure. The token is minted only when expanded (or when the stuck timeout auto-expands it), with a **Generate new token** retry on mint error / token expiry.
- **One dialog-level arrival status row** (scoped to the machine's `client.id` via `selectArrivedClient`) so the *Waiting… / back online* signal covers whichever command the operator actually ran, instead of pinning the spinner to the `daemon start` block.
- Chevron disclosure + stable label + `aria-controls`, matching the app's disclosure convention.

Supporting cleanups:
- Remove the now-redundant inline "Daemon not running?" disclosure from `offline-card-body.tsx` — the dialog owns the command, so there's one source of truth (no drift).
- Extract `ConnectStatusRow` from `ConnectCommandPanel` (reused at dialog level) and gate its caption on a present command, so *"Single-use · regenerates…"* never renders over the *"Generating token…"* placeholder (also fixes a latent case in `NewConnectionDialog`).

The re-auth ("Generate new token" on auth-expired) and `+ Connect` (fresh unknown machine) flows are untouched.

## Verification

- Two independent reviews (correctness + UX/design). Correctness: no blocking issues. UX found 2 HIGH (two-commands/one-spinner layout; caption over the "Generating token…" placeholder) + 1 MEDIUM (text-glyph disclosure vs the app's chevron convention) + a fallback mint-error dead-end and missing stuck escalation — all fixed here.
[...]

PR #1476, opened 2026-07-07
[codex] Revert daemon-first offline Reconnect dialog

yuezengwu:
## Summary

- Removes the daemon-first `ReconnectDialog` added by #1373.
- Restores the offline computer `Reconnect` action to the unscoped `NewConnectionDialog` / connect-token flow, so Web stays root-cause neutral when it only knows `client.status = disconnected`.
- Keeps the generic `ConnectCommandPanel` cleanup from #1373: caption text is gated on a real command, and the shared status row helper remains with non-`ReconnectDialog`-specific comments.

## Reason

The web UI only knows that a client row is `disconnected`; it does not know why the local CLI is offline. Treating `<binName> daemon start` as the primary Reconnect command assumes the root cause is only a stopped daemon, which can be wrong. The reconnect action should stay root-cause neutral rather than presenting restart as the default recovery path.

## Notes

The current Context Tree node `system/cloud/runtime/computer-disconnect.md` still records the daemon-first decision from #1373, so it should be synced separately to match this rollback decision.

## Validation

- `pnpm --filter @first-tree/web test -- clients-page-cards-dom` (135 files / 1106 tests passed)
- `pnpm exec biome check packages/web/src/components/connect-command-panel.tsx packages/web/src/pages/clients.tsx packages/web/src/pages/clients/__tests__/clients-page-cards-dom.test.tsx packages/web/src/pages/clients/cards/computer-card.tsx packages/web/src/pages/clients/cards/offline-card-body.tsx`
- `git diff --check $(git merge-base origin/main HEAD)`


baixiaohang (approved):
## Review by codex-assistant - PR #1476

**Recommendation**: approve
- Rationale: This replacement PR carries the same partial-revert diff reviewed on #1475, now on an allowed branch name, and the rollback behavior remains sound.

**Risk level**: C
- Path baseline: only `packages/web/**` files are changed.
- Semantic lift: none; this is a web UI recovery-flow change and does not touch subscriptions, WS frame parsing, or shared data-layer contracts.

**PR summary**
- Problem: Operators clicking `Reconnect` on a disconnected computer were being led through a daemon-first recovery path even though Web only knows the row is disconnected, not why the local CLI is offline.
- Approach: The offline-row `Reconnect` action now opens the unscoped `Connect computer` connect-token flow again, while the lightweight `<binName> daemon start` command remains available behind the inline `Daemon not running?` disclosure. The PR also keeps the generic `ConnectCommandPanel` cleanup from #1373 and removes the now-specific `ReconnectDialog` path.
- Impacted modules: `/settings/computers` cards, `NewConnectionDialog` wiring, connect command panel rendering, and the card DOM regression test.

**Review findings**
✅ 1. Replacement hygiene checks out: #1476 uses `fix/revert-pr-1373-reconnect-command`, and the commit OIDs / changed files match the already-reviewed #1475 branch rename replacement.
✅ 2. The row-level `Reconnect` handler now calls `openNewConnection`, which clears the re-auth target and uses `NewConnectionDialog`'s unscoped arrival detector as intended. [packages/web/src/pages/clients.tsx:110]
✅ 3. The daemon-first dialog path is fully removed from the page wiring and source tree; I did not find lingering `ReconnectDialog` references. [packages/web/src/pages/clients/reconnect-dialog.tsx]
⚠️ 4. `system/cloud/runtime/computer-disconnect.md` still records the daemon-first decision. The PR body calls out a separate sync, so I am not blocking this web rollback on it, but that tree update should be paired before treating the decision as documented.
⚠️ 5. The updated DOM test covers the inline daemon command and the generic Connect dialog, but it no longer directly clicks the offline-row `Reconnect` action. Acceptable for this revert; worth restoring if this flow changes again.

**Action taken**
- Approved. CI for the replacement branch was still partially in progress when reviewed.

PR #1477, opened 2026-07-07
fix: allow scoped outbox sends under runtime enforcement

baixiaohang:
## Summary

- allow scoped `agent_outbox` JWTs to pass `agentSelectorHook` without `x-agent-runtime-session`
- keep existing Rule R-RUN checks intact before the outbox exemption
- add enforcement coverage for scoped outbox sends, wrong route, wrong chat, and wrong agent

## Why

Landing trial Codex sandboxes intentionally scrub the broader runtime-session secret and use a narrow `agent_outbox` token for current-chat message writes. With runtime-session enforcement enabled, the server accepted the outbox token in `userAuthHook` but then rejected it in `agentSelectorHook` for missing `x-agent-runtime-session`.

## Verification

- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/agent-bind-authorization.test.ts src/__tests__/landing-campaigns-start.test.ts`
- `pnpm check`
- `pnpm typecheck`
- Accidental broader run while correcting the test command: two full `@first-tree/server` test runs passed (`177` files, `1721` tests each).

Notes: `pnpm check` still reports existing non-blocking warnings in unrelated files (`assistant-text.test.ts`, `workspace-migrations.ts`, `packages/web/src/index.css`).

### DIFF
--- a/packages/server/src/__tests__/agent-bind-authorization.test.ts
+++ b/packages/server/src/__tests__/agent-bind-authorization.test.ts
@@ -200,6 +200,100 @@ describe("runtime-bound agent HTTP enforcement", () => {
     expect(res.statusCode).toBe(200);
   });
 
+  it("accepts scoped agent outbox message writes without a runtime session token", async () => {
+    const app = getApp();
+    const sender = await createTestAgent(app, { name: `outbox-sender-${crypto.randomUUID().slice(0, 6)}` });
+    const recipient = await createTestAgent(app, { name: `outbox-recipient-${crypto.randomUUID().slice(0, 6)}` });
+    const runtimeSessionToken = await bindAgentRuntimeSession(app.db, sender.agent.uuid, sender.clientId);
+
+    const chatRes = await sender.request(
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the runtime-enforcement path and the existing outbox scope checks.

Goal/core change: this lets hosted trial sandboxes use their scoped `agent_outbox` JWT for the current chat's message POST without also exposing the broader `x-agent-runtime-session` secret. The middleware still runs active-agent, org membership, and Rule R-RUN/client ownership checks before the exemption; only the runtime-session header check is skipped after `userAuthHook` has already narrowed the JWT to `POST /api/v1/agent/chats/:chatId/messages` for the claimed agent/chat.

The added test covers the enforcement-on happy path plus wrong route, wrong chat, and wrong agent negatives. Existing landing trial tests cover completed-chat rejection through the same outbox token path. I do not see a blocker here.

PR #1478, opened 2026-07-07
fix: tighten scoped outbox route checks

baixiaohang:
## Summary

Follow-up to #1477 after the original runtime-enforcement fix was merged.

This tightens the `agent_outbox` request scope check from a substring search to an anchored exact route match for `POST /api/v1/agent/chats/:chatId/messages`, adds a same-method sibling route negative test for `POST /outbox-token`, and documents the outbox token as a runtime-delegated narrow send-only capability.

## Verification

- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/agent-bind-authorization.test.ts src/__tests__/landing-campaigns-start.test.ts`
- `pnpm check`
- `pnpm typecheck`

`pnpm check` still reports the existing unrelated 2 warnings / 1 info in client/web files.

### DIFF
--- a/packages/server/src/__tests__/agent-bind-authorization.test.ts
+++ b/packages/server/src/__tests__/agent-bind-authorization.test.ts
@@ -250,6 +250,16 @@ describe("runtime-bound agent HTTP enforcement", () => {
     });
     expect(wrongPath.statusCode).toBe(401);
 
+    const wrongSibling = await app.inject({
+      method: "POST",
+      url: `/api/v1/agent/chats/${chatId}/outbox-token`,
+      headers: {
+        authorization: `Bearer ${outbox.accessToken}`,
+        "x-agent-id": sender.agent.uuid,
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
Reviewed the route-scope tightening for `agent_outbox` tokens.

No blocking findings. The new predicate is method-gated, agent-header-gated, and anchored to the exact `POST /api/v1/agent/chats/:chatId/messages` pathname before decoding/comparing the scoped chat id. That matches the intended narrow send-only outbox capability, and the added same-method sibling-route negative coverage protects the nearby `outbox-token` route.

Local verification run:

`pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/agent-bind-authorization.test.ts src/__tests__/landing-campaigns-start.test.ts`

Result: 2 files / 48 tests passed.

PR #1483, opened 2026-07-07
Retire purged clients with server tombstones

yuezengwu:
## Summary

- Add `clients.retired_at` as the server-side tombstone for retired local client identities.
- Make `logout --purge` retire the current server client before destructive local cleanup, stop foreground daemons before purge, and fail closed if the runtime cannot be safely stopped or retired.
- Hide retired clients from default computer lists while preserving detail/status semantics, and block retired clients from register/heartbeat/capabilities/runtime target paths.
- Clear agents pinned to a retired client into `suspended + client_id=null`, then allow Web runtime switch to move those cleared agents to a connected runtime.

## Validation

- [x] `pnpm --filter @first-tree/server test` — 178 files / 1732 tests passed
- [x] `pnpm --filter @first-tree/web test` — 135 files / 1108 tests passed
- [x] `pnpm --filter first-tree-dev test -- agent-lifecycle-commands-extra.test.ts client-switch.test.ts daemon-start-command.test.ts login-command.test.ts` — actual CLI full suite, 70 files / 603 tests passed
- [x] `pnpm --filter @first-tree/server typecheck`
- [x] `pnpm --filter @first-tree/server build`
- [x] `pnpm --filter @first-tree/web build`
- [x] `pnpm --filter first-tree-dev typecheck`
- [x] `pnpm --filter first-tree-dev build`
- [x] `pnpm exec biome check ...` on touched source/docs files
- [x] Drizzle meta JSON parse check
- [x] `git diff --check`
- [x] QA real local retest passed: foreground daemon argv0-only stop + retire + purge, and fake marker pointing at `sleep` fails closed without killing the unrelated process

## Change Surface

- [x] `apps/cli` public CLI or help output
- [x] docs or contributor-facing repository metadata

## Notes

- DB/schema change: `packages/server/drizzle/0073_client_retired_at.sql` adds nullable `clients.retired_at timestamptz`; no new table, enum, default, FK, or index.
- Server behavior: retired clients remain rows, surface as `status: "retired"`, cannot be reactivated with the same client id, and cannot be targeted by runtime-sensitive APIs.
- CLI behavior: `logout --purge` stops background service plus trusted foreground daemon markers before server retire/local deletion; unsafe foreground markers fail closed.
- Web behavior: cleared suspended agents can use the runtime switch entry to move back onto a connected computer/runtime.
- Branch note: this branch was created from a local base currently 2 commits behind `origin/main`; no rebase/merge was performed during delivery.


baixiaohang (changes requested):
## Review by codex-assistant - PR #1483

- Rationale: The server tombstone path still has a bind/retire race that can resurrect runtime presence for a retired client.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A; also changes server schema and core client/runtime services.
- Semantic lift: DB schema and client/agent runtime routing semantics; no lift beyond A.

**PR Summary**
- Problem: Operators need `logout --purge` to make a local client identity unreusable on the server before local destructive cleanup, while preserving agents/chats and giving owners a way to move cleared agents to a new runtime.
- Approach: Add a server-side `clients.retired_at` tombstone, retire the current client during purge, hide retired clients from default lists, reject retired clients on runtime-sensitive paths, suspend/unpin agents routed through the retired client, and let Web runtime switch move those cleared agents to a connected runtime.
- Impacted modules: `apps/cli`, `packages/server` client/agent/runtime services and APIs, Drizzle migrations, `packages/web` agent runtime UI.

**Review findings**
❌ 1. `agent:bind` can report an agent online after a concurrent client retirement. [R4 / `packages/server/src/api/agent/ws-client.ts:1175`]

`agent:bind` does its final durable guard inside `bindAgentRuntimeSession(...)`, then unconditionally calls `presenceService.bindAgent(...)`, registers the in-memory route, and sends `agent:bound`. A concurrent `retireClient()` can complete after the runtime-session update but before that presence write: it clears `agent_presence`, sets the agent to `suspended` with `client_id = NULL`, and tombstones the client. The bind handler then writes the presence row back to `online` for the retired client and returns a runtime-session token that DB validation will no longer accept. The newly added test covers “retired after register, before bind starts,” but not this mid-bind interleaving. Please make the bind publication atomic with the same active-agent / same-client / non-retired-client predicate, or re-check and abort before writing presence and sending `agent:bound`.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the R4 bind/retire race in `4fcd855c`.

What changed:
- `agent:bind` now publishes presence through `bindAgentIfActiveClient(...)`, which serializes with `retireClient()` by locking the client row first and then the agent row before the presence upsert.
- If the durable route is gone after the runtime-session claim, bind now revokes the just-minted runtime session and rejects without writing in-memory routing or sending `agent:bound`.
- Added active-socket guards around the publication path so a displaced foreground socket cannot publish/bind after losing the client slot.
- Added a regression test that retires the client immediately after `bindAgentRuntimeSession(...)` succeeds and verifies the handler rejects, presence stays offline, the agent remains suspended/unpinned, and no in-memory route is installed.

Validation run locally:
- `pnpm --dir packages/server exec vitest run src/__tests__/ws-client-reconnect-race.test.ts src/__tests__/agent-pinned-notification.test.ts src/__tests__/ws-session-event.test.ts`
- `pnpm exec biome check packages/server/src/services/presence.ts packages/server/src/api/agent/ws-client.ts packages/server/src/__tests__/ws-client-reconnect-race.test.ts`
- `pnpm --dir packages/server exec tsc --noEmit`
- `pnpm --dir packages/server exec vitest run` → 178 files / 1733 tests passed

baixiaohang (changes requested):
## Review by codex-assistant - PR #1483

- Rationale: The retire/bind serialization fix closes the original tombstone race, but the new displaced-socket cleanup can revoke a newer active bind's runtime session.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A; also changes server schema and core client/runtime services.
- Semantic lift: DB schema and client/agent runtime routing semantics; no lift beyond A.

**PR Summary**
- Problem: Operators need `logout --purge` to make a local client identity unreusable on the server before destructive local cleanup, while preserving agents/chats and giving owners a way to move cleared agents to a new runtime.
- Approach: Add a server-side `clients.retired_at` tombstone, retire the current client during purge, hide retired clients from default lists, reject retired clients on runtime-sensitive paths, suspend/unpin agents routed through the retired client, and let Web runtime switch move those cleared agents to a connected runtime.
- Impacted modules: `apps/cli`, `packages/server` client/agent/runtime services and APIs, Drizzle migrations, `packages/web` agent runtime UI.

**Review findings**
❌ 1. A displaced socket can revoke the active socket's runtime session for the same client. [R4 / `packages/server/src/api/agent/ws-client.ts:1203`]

The new post-publish active-socket guard correctly prevents an old socket from installing the in-memory route after another socket has taken over the same `clientId`. However, the cleanup calls `revokeAgentRuntimeSession(app.db, agent.id, clientId)`, and that helper only matches `agent_id + client_id`; it does not prove the metadata still contains the token minted by this handler. A real interleaving is: socket A mints token A and publishes presence, socket B takes over the same client slot and successfully binds the same agent with token B, then socket A resumes, sees it is no longer active, and revokes `runtimeSession` for that same agent/client. B can still send `agent:bound` and install the in-memory route, but subsequent agent-scoped HTTP with B's token fails durable validation because the DB metadata was removed. Please make the abort cleanup conditional on the exact runtime-session token/hash this handler minted, or otherwise serialize/own the publish-abort path so a stale handler cannot delete a successor bind's session.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the second R4 displaced-socket cleanup finding in `a227ef60`.

What changed:
- Added `revokeAgentRuntimeSessionIfTokenMatches(...)`, which removes `metadata.runtimeSession` only when the current DB metadata still matches the exact client id and token hash minted by this bind handler.
- The `agent:bind` abort cleanup now uses that exact-token revoke for both publish-failure and post-publish inactive-socket paths.
- Presence cleanup after a post-publish inactive socket now runs only if the exact-token revoke succeeded, so a stale handler cannot clear a successor bind's presence after the successor has replaced the token.
- Added a regression test for the interleaving from the review: socket A publishes, socket B/successor mints token B and installs the same-client route, then socket A aborts. The test verifies token B still passes durable validation, the in-memory route still points at the client, and presence remains online.

Validation run locally:
- `pnpm --dir packages/server exec vitest run src/__tests__/ws-client-reconnect-race.test.ts src/__tests__/agent-bind-authorization.test.ts` → 21 tests passed
- `pnpm exec biome check packages/server/src/services/agent-runtime-session.ts packages/server/src/api/agent/ws-client.ts packages/server/src/__tests__/ws-client-reconnect-race.test.ts`
- `pnpm --dir packages/server exec tsc --noEmit`
- `pnpm --dir packages/server exec vitest run` → 178 files / 1734 tests passed

yuezengwu:
Follow-up fix for QA focused real regression on head `a227ef60`.

Root cause: the real foreground path could have a connected client registered by `daemon start --foreground --no-interactive`, while active `config/client.yaml` still only contained the server URL. That happened when `client.id` was supplied/resolved outside the persisted yaml path, so `logout --purge` could not find the current client id and failed closed with `PURGE_CLIENT_RETIRE_UNAVAILABLE` before server retire or local purge.

Fix in new head `87303893`:
- persist the resolved active-root client id during `login` and `daemon start`, without overwriting an existing root client id;
- let `logout --purge` recover the active-root client id from the switch index when `client.yaml` is missing the id, with validation that the index entry is the same `clientId` and `storage: "active-root"`;
- ensure explicit `daemon start --foreground --no-interactive` remains foreground even if `FIRST_TREE_SERVICE_MODE=1` is inherited, so foreground markers do not become service markers.

Local verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/client-switch.test.ts src/__tests__/login-command.test.ts src/__tests__/daemon-start-command.test.ts src/__tests__/agent-lifecycle-commands-extra.test.ts` -> 4 files / 62 tests passed
- `pnpm exec biome check ...` -> passed
- `pnpm --dir apps/cli exec tsc --noEmit` -> passed
- `pnpm --dir apps/cli exec vitest run` -> 70 files / 609 tests passed
- `pnpm --dir apps/cli run build` -> passed
- `git diff --check` -> passed

Requesting QA retest on `87303893`.

baixiaohang (changes requested):
## Review by codex-assistant - PR #1483

- Rationale: The bind/retire races from the prior reviews are addressed, but the latest flow still leaves a bypass around the managed runtime-switch recovery path for retired-cleared agents.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A; also changes server schema and core client/runtime services.
- Semantic lift: DB schema and client/agent runtime routing semantics; no lift beyond A.

**PR Summary**
- Problem: Operators need `logout --purge` to make a local client identity unreusable on the server before destructive local cleanup, while preserving agents/chats and giving owners a way to move cleared agents to a new runtime.
- Approach: Add a server-side `clients.retired_at` tombstone, retire the current client during purge, hide retired clients from default lists, reject retired clients on runtime-sensitive paths, suspend/unpin agents routed through the retired client, and let Web runtime switch move those cleared agents to a connected runtime.
- Impacted modules: `apps/cli`, `packages/server` client/agent/runtime services and APIs, Drizzle migrations, `packages/web` agent runtime UI.

**Review findings**
❌ 1. Retired-cleared agents can bypass the managed runtime-switch path via Reactivate + Bind computer. [R2/R4 / `packages/server/src/services/agent.ts:1201`]

`retireClient()` now intentionally leaves affected agents in `suspended + client_id = NULL`, and the new recovery path is `switch-runtime`, which records a claim, retags config, archives sessions, sends route-change notifications, and then reactivates the agent on the target client. However, the existing `reactivateAgent()` path still only checks that the agent is suspended, then sets `status='active'` without requiring a client route. The Web Profile tab also still renders `Reactivate` for any suspended non-human agent. A user can therefore take a retire-cleared agent, click Reactivate, get an `active + client_id = NULL` agent, and then use the ordinary `PATCH { clientId }` / first-bind path to place it on a computer. That alternate path skips the runtime-switch cleanup and session archival that this PR depends on for moving agents off a retired client. Please gate reactivation for unbound non-human agents, or otherwise mark/route retire-cleared agents so their only recovery path is the managed runtime switch.

⚠️ 2. `logout --purge` still proceeds when the background service state is unknown. [R4 / `apps/cli/src/commands/logout.ts:192`]

The purge path fails closed when an `active` service cannot be stopped, and it now handles trusted foreground markers, but it does nothing when `getClientServiceStatus()` returns `unknown`. That is the same “cannot safely determine whether the runtime is stopped” condition that `daemon start` and local-client switching already treat as unsafe. In this PR, proceeding is more dangerous because `logout --purge` may retire the server client and delete `client.yaml`, credentials, local agents, sessions, and workspaces while a supervised daemon may still be running with the old in-memory client identity. Please make purge refuse on `unknown` service state, with coverage similar to the active-stop failure case.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the latest request-changes review in head `7e512988`.

Changes:
- Retired-cleared agents (`suspended + client_id = NULL`) can no longer bypass managed runtime-switch recovery:
  - `reactivateAgent()` rejects unbound suspended non-human agents with a runtime-switch recovery error.
  - direct `PATCH { clientId }` first-bind also rejects unbound suspended agents, so the user cannot bind first and then reactivate.
  - Web Profile and Team action menus no longer show ordinary `Reactivate` for unbound suspended agents.
- `logout --purge` now fails closed when the background service state is `unknown`, before server retire or local deletion.

Local verification:
- Targeted server `agent-runtime-switch`: 12 tests passed
- Targeted CLI logout lifecycle: 15 tests passed
- Targeted Web agent detail + team DOM: 21 tests passed
- Biome targeted check passed
- Server `tsc --noEmit` passed
- CLI `tsc --noEmit` passed
- Web `typecheck` passed, including design-token lint
- Server full vitest: 178 files / 1735 tests passed
- CLI full vitest: 70 files / 610 tests passed
- Web package test: 135 files / 1109 tests passed
- Server, Web, and CLI builds passed
- `git diff --check` passed

Note: raw `pnpm --dir packages/web exec vitest run` fails on the existing empty Playwright scaffold `tests/visual-regression.spec.ts`; the package script `pnpm --dir packages/web test` includes `--passWithNoTests` and passes.

yuezengwu:
QA R5 focused regression on head `7e5129889980ce408992c090d60eff535ebf7d65` is complete: `PASS_WITH_SCOPE_NOTES`.

Passed:
- real server/API/DB/foreground daemon chain: source/target clients truly registered; retiring source produced `suspended + client_id=NULL` and offline presence.
- ordinary bypasses are blocked: `POST /agents/:uuid/reactivate` -> 400 and DB unchanged; `PATCH /agents/:uuid {clientId}` -> 400 and DB unchanged.
- managed recovery still works: `switch-runtime` -> 200, target daemon binds, agent returns `active + online` on target runtime.
- source daemon reconnect rejected with `CLIENT_RETIRED`.
- targeted regression sets passed: server 4 files / 37 tests, CLI sanitized 4 files / 63 tests, Web DOM 4 files / 20 tests, unknown-service focused CLI 1/1.

Scope notes:
- browser click-level Web verification was blocked because the in-app browser backend was unavailable, so Web evidence is DOM-targeted tests.
- OS-level `unknown` service state has no stable public black-box reproduction; evidence is the focused CLI regression test.

CI on PR head is green: GitHub reports all CI/CodeQL checks successful and the PR remains mergeable.

baixiaohang (approved):
Approved per maintainer authorization; no remaining blocking concerns from my review.

Issue #1484, opened 2026-07-07
feat(server): landing campaign trial quotas — per-user daily cap, global circuit breaker, default token cap


Gandy2025:
## Problem

Landing campaign trials (`POST /me/landing-campaigns/start`) currently have **no run-count limit at any level**. Budget control exists only per chat (`landingCampaignMaxAgentTurns` / `landingCampaignMaxEstimatedTokens`), and every new repo gets a fresh chat with a fresh budget. This leaves the official runtime open to unlimited free compute:

1. **A single budget can be unbounded.** Default is 6 turns per chat, but `landingCampaignMaxEstimatedTokens` defaults to `undefined` (unlimited tokens per turn). See `packages/shared/src/config/` growth config.
2. **Switching repos is free and scriptable.** The trial does not require owning the target repo — any public GitHub repo URL works, and each distinct repo mints a new chat with a full budget (`ensureTrialChatAndBootstrap` keys the chat on `campaign + repo.canonicalKey`, `packages/server/src/services/landing-campaigns/start.ts`).
3. **Per-org limits alone would be useless.** `selfCreateOrganization` (`packages/server/src/services/membership.ts:596`) lets any user self-serve unlimited orgs, so the only meaningful identity anchor is the user (GitHub account via OAuth).
4. **There is no rate limiting anywhere in the server**, and `organizations.maxMessagesPerMinute` is a dead column (never enforced). The official runtime is a single shared client, so trial flooding is an availability problem for legitimate users, not just a billing problem.

## Plan (deliberately minimal)

Three controls. Two count queries + three config values. No new tables, no middleware, no UI work.

### 1. Give the token cap a real default

Change `landingCampaignMaxEstimatedTokens` from `undefined` (unlimited) to a real number. Calibrate from one real production-scan run × 1.5. One-line change in the shared config schema.

### 2. Quota assertion: one function, two counts

Add `assertTrialQuota(db, config, userId)` to `packages/server/src/services/landing-campaigns/guards.ts` (existing assertion-helper file — do not create a new file):

- **Per-user daily cap** (default 5/day): count today's trial chats created by this user **across all orgs** — `chats.creatorAgentId IN (SELECT agentId FROM members WHERE userId = :userId)` AND metadata contains `landingCampaignTrial` AND `createdAt >= today`. Must be cross-org (see problem #3).
[...]

PR #1485, opened 2026-07-07
feat(server): re-send landing campaign bootstrap when a trial run goes silent

Gandy2025:
## Problem

A landing campaign trial whose run dies before the agent ever replies (skill-repo clone failure, lost Codex rollout, official client hiccup) leaves the chat permanently stuck. The same-repo idempotent start finds an existing first message and returns the dead chat untouched, so the user's natural recovery action — re-clicking the landing CTA — can never revive it. The user sees a silent chat with a single bootstrap message and no way forward short of switching repos.

Two properties of the existing design make a minimal fix possible:

- Turn/token accounting only advances on `turn_end status="success"` (`packages/server/src/api/agent/ws-client.ts`), so a failed run consumes **zero** budget — a dead chat is `running` with its budget intact.
- A fresh inbox message re-activates the runtime through normal delivery, so "append one message" is a complete re-kick.

## Change

Make the idempotent start liveness-aware. `ensureTrialChatAndBootstrap` re-sends a bootstrap into the **same chat** only when every one of these holds:

1. the trial chat is still `running`;
2. the trial agent has never sent a message in the chat;
3. the last message is our own bootstrap (`systemSender: "first_tree_onboarding"`) and is older than 10 minutes;
4. there are no session events within the last 10 minutes (a slow-but-alive run is left alone).

Everything else — live runs, `awaiting_user`, `completed`, human-message-last chats, fresh bootstraps — keeps the exact current no-op behavior. Re-clicking the landing CTA is now the retry button; no new states, endpoints, chats, budget resets, or config.

The resent message is a new self-contained variant (`buildLandingCampaignRetryBootstrap`): the runtime session may be brand new, so repo URL, skill repo, and skill name are restated instead of referring to "the repo above"; a resumed session is told to continue rather than restart. It carries `bootstrapRetry: true` in metadata, which doubles as the throttle marker and a resend-rate metric hook.

## Tests

- new: silent past the retry window → bootstrap resent with `bootstrapRetry: true`, same chatId, self-contained instruction;
- new: recent session activity → no resend;
- new: agent already replied → no resend (also guards the equal-`createdAt` ordering edge, which is why the any-agent-message check exists alongside the last-message check);
- full `@first-tree/server` suite green (177 files / 1724 tests); `@first-tree/client` (1233), `@first-tree/shared` (507), `@first-tree/web` (1133) green; `pnpm check` + `pnpm typecheck` clean. Pre-existing local-environment failures untouched by this server-only diff: the `@first-tree/skill-evals` fixture-clone timeout and 3 `apps/cli` login-command cases (both packages byte-identical to `main`).

Related: #1484 tracks the separate trial quota work (per-user daily cap, global circuit breaker); the two mechanisms are deliberately decoupled — this retry consumes neither budget nor quota.

### DIFF
--- a/packages/server/src/__tests__/landing-campaigns-start.test.ts
+++ b/packages/server/src/__tests__/landing-campaigns-start.test.ts
@@ -3,7 +3,7 @@ import {
   parseLandingCampaignTrialAgentMetadata,
   parseLandingCampaignTrialChatMetadata,
 } from "@first-tree/shared";
-import { and, eq, sql } from "drizzle-orm";
+import { and, asc, eq, sql } from "drizzle-orm";
 import { describe, expect, it } from "vitest";
 import { agentConfigs } from "../db/schema/agent-configs.js";
 import { agentResourceBindings } from "../db/schema/agent-resource-bindings.js";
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
Implementation goal: recover a landing campaign trial chat whose original bootstrap woke no usable runtime, without creating a new chat or resetting trial budget.

Core change reviewed: `ensureTrialChatAndBootstrap` now treats same-repo idempotent start as a liveness check. It locks the existing chat, verifies the trial is still `running`, verifies the last row is the First Tree bootstrap and old enough, rejects chats with any agent-authored message, and rejects chats with recent session events before sending a self-contained retry bootstrap with `bootstrapRetry: true`. The catalog adds the retry copy, and the tests cover the silent retry, recent-session no-op, and agent-already-replied no-op cases.

I reviewed this against the Context Tree trial-chat/session constraints. No DB/schema migration or core persisted data-structure change is introduced; this is a new message metadata marker plus retry behavior on existing tables.

I did not find blockers. Approved.

Local verification: `git diff --check origin/main...HEAD` passed. I could not run the focused Vitest file locally because this fresh worktree has no `node_modules` and `vitest` is not installed (`pnpm --filter @first-tree/server test -- landing-campaigns-start.test.ts` failed before tests started).

baixiaohang (changes requested):
## Review by codex-assistant - PR #1485

- Rationale: The retry liveness check misses the current authoritative no-events working signal, so it can duplicate a live Codex trial.

**Risk level**: B-high
- Path baseline: `packages/server/**` only, including `packages/server/src/services/landing-campaigns/start.ts` -> B-low.
- Semantic lift: changes chat/message/session retry behavior for a service-managed trial agent -> B-high.

**PR Summary**
- Problem: A landing CTA re-click cannot recover a trial chat when the initial runtime dies before the agent replies, leaving the user in a silent bootstrap-only chat.
- Approach: Reuse the same chat and append a self-contained retry bootstrap after a 10-minute silence window, while keeping live, completed, awaiting-user, human-last, and agent-replied chats as no-ops.
- Impacted modules: `packages/server/src/services/landing-campaigns/start.ts`, `packages/server/src/services/landing-campaigns/skills/catalog.ts`, `packages/server/src/__tests__/landing-campaigns-start.test.ts`.

**Review findings**
❌ 1. The liveness guard ignores fresh per-chat runtime state. `ensureTrialChatAndBootstrap` only checks recent `session_events` before resending (`packages/server/src/services/landing-campaigns/start.ts:472`), but the current server model treats `agent_chat_sessions.runtime_state='working'` with a fresh `runtime_state_at` as the authoritative "turn is in flight" signal, including the Codex no-events case. The existing `me-chats` code and tests explicitly cover a working runtime with zero session events (`packages/server/src/services/me-chat.ts:512`, `packages/server/src/__tests__/me-chats-http-e2e.test.ts:104`). In that state this PR will append a second bootstrap into a run that is still active, which can start duplicate scan work in the same trial chat. Please gate retry on the same fresh per-chat runtime projection, or an equivalent helper, before falling back to `session_events` silence.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
I rechecked the requested-changes finding and agree it is a blocker.

My earlier approval missed that the authoritative in-flight signal for Codex no-events turns is fresh per-chat `agent_chat_sessions.runtime_state='working'` / `runtime_state_at`, not recent `session_events`. As written, the retry guard can append a second bootstrap while a live trial turn is still working but has emitted no events.

Please block retry on the same fresh per-chat runtime projection used by `busyAgentIds` / `resolveAgentChatStatuses`, then keep the `session_events` silence check as the explanatory activity fallback.

Gandy2025:
Fixed in aac35d96. The finding was correct — checking only recent `session_events` misses the codex no-events case where a long turn is working but emits nothing until `turn_end`, and the retry could have appended a second bootstrap into a live run.

The guard now consults the same authoritative per-chat runtime projection the composite status uses: before the `session_events` silence fallback, it reads `agent_chat_sessions.runtime_state` / `runtime_state_at` for the (agent, chat) pair and calls the shared `computeWorking` helper from `agent-chat-status.ts`. If it reports fresh `working`, the retry is a no-op.

Details:
- Reuses `computeWorking` rather than re-deriving the freshness logic, so the retry gate and `resolveAgentChatStatuses` stay in lockstep on `RUNTIME_STALE_MS`.
- Passing `activity: null` means old clients (NULL `runtime_state_at` stamp) return false from `computeWorking` and fall through to the legacy `session_events` freshness proxy — their existing working signal — so the one-release-cycle fallback is preserved.
- No session row / never-active session → `computeWorking` returns false → unchanged fall-through to the events check.
- New test: fresh `runtime_state='working'` with zero session events → no resend.

Full `@first-tree/server` suite green (177 files / 1725 tests), `pnpm check` + `pnpm typecheck` clean.

yuezengwu (approved):
Implementation goal/core change: same as before, recover a dead landing campaign trial by re-kicking the same chat without resetting budget or creating a new chat.

Re-review after `aac35d96`: the requested runtime-state blocker is fixed. `ensureTrialChatAndBootstrap` now reads the `(agent, chat)` `agent_chat_sessions` row and calls shared `computeWorking(...)` before falling back to the session-events silence check. That keeps the retry gate in lockstep with the per-chat runtime projection used by `busyAgentIds` / `resolveAgentChatStatuses`, so the Codex no-events live-turn case no longer gets a duplicate bootstrap. Passing `activity: null` also preserves the intended old-client path: NULL `runtime_state_at` falls through to the existing session-event freshness check.

The new test covers fresh `runtime_state='working'` with zero session events and verifies no resend. I did not find remaining blockers.

Local verification: `git diff --check origin/main...HEAD` passed. I did not run Vitest locally because this temporary worktree still has no installed `node_modules`; relying on the PR's server-suite report / CI for execution.

baixiaohang (approved):
## Review by codex-assistant - PR #1485

**Recommendation**: approve
- Rationale: The previous runtime-state blocker is fixed in `aac35d96`; the retry guard now uses the same fresh per-chat working projection as chat status before falling back to session-event activity.

**Risk level**: B-high
- Path baseline: `packages/server/**` only, including `packages/server/src/services/landing-campaigns/start.ts` -> B-low.
- Semantic lift: changes chat/message/session retry behavior for a service-managed landing campaign trial agent -> B-high.

**PR Summary**
- Problem: A landing CTA re-click could not recover a trial chat when the first bootstrap did not produce a usable runtime turn, leaving the user stuck in a silent bootstrap-only chat.
- Approach: Treat same-repo idempotent start as a narrow liveness check. If the existing trial chat still looks dead after the retry window, append a self-contained retry bootstrap in the same chat; otherwise keep the existing no-op behavior.
- Impacted modules: `packages/server/src/services/landing-campaigns/start.ts`, `packages/server/src/services/landing-campaigns/skills/catalog.ts`, `packages/server/src/__tests__/landing-campaigns-start.test.ts`.

**Review findings**
✅ 1. The prior blocker is fixed: `ensureTrialChatAndBootstrap` now reads the `(agent, chat)` `agent_chat_sessions` row and calls shared `computeWorking(...)` before the session-events silence fallback, so a fresh Codex no-events working turn no longer receives a duplicate bootstrap.
✅ 2. The fix stays appropriately scoped: no new retry endpoint, no second chat, no budget reset, and no duplicate local definition of runtime freshness.
✅ 3. The added test covers fresh `runtime_state='working'` with zero session events and verifies no resend.

**Action taken**

PR #1486, opened 2026-07-08
fix(web): keep empty runtime capabilities detecting

yuezengwu:
## Summary

- keep onboarding/shared computer setup in the detecting state when the server returns an empty capability snapshot
- only treat capabilities as loaded once the connected client has reported at least one provider entry
- add regression coverage for empty snapshot -> later provider report

## Test plan

- `pnpm exec biome check packages/web/src/features/agent-setup/use-computer-connection.ts packages/web/src/features/agent-setup/__tests__/setup-hooks.test.tsx`
- `pnpm --filter @first-tree/web test -- packages/web/src/features/agent-setup/__tests__/setup-hooks.test.tsx` (runs the web Vitest suite via the package script; 137 files / 1136 tests passed)

### DIFF
--- a/packages/web/src/features/agent-setup/__tests__/setup-hooks.test.tsx
+++ b/packages/web/src/features/agent-setup/__tests__/setup-hooks.test.tsx
@@ -165,6 +165,62 @@ describe("shared setup hooks", () => {
     expect(eventMocks.reportOnboardingEvent).not.toHaveBeenCalled();
   });
 
+  it("keeps an empty capability snapshot in detecting state until a provider report arrives", async () => {
+    const latest = { current: null as ComputerConnection | null };
+    const client = {
+      id: "client-1",
+      userId: "user-self",
+      status: "connected",
+      authState: "ok",
+      binName: "first-tree-dev",
+      sdkVersion: "0.5.0",
+      hostname: "gandy-macbook",
+      os: "darwin",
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by codex-assistant - PR #1486

**Recommendation**: approve
- Rationale: The change preserves the runtime-capability loading distinction already used by the server/runtime contract and adds focused regression coverage for empty snapshot -> later provider report.

**Risk level**: C
- Semantic lift: none; this is a web onboarding hook state-derivation change and does not touch subscriptions, WS frame parsing, or shared data-layer contracts.

**PR Summary**
- Problem: during onboarding/shared setup, a newly connected computer can appear before daemon capability probing has uploaded provider entries, and the UI could incorrectly show "no runtime" instead of continuing to detect.
- Approach: treat the connected client's capabilities as active only after the capability map has at least one provider entry, preserving the detecting state for `{}` and letting the existing polling promote it when a provider report arrives.
- Impacted modules: `packages/web/src/features/agent-setup`.

**Review findings**
✅ 1. No blocking findings. This matches the runtime contract that empty/absent capabilities mean "not reported yet", while reported non-ok entries are the explicit "not ready" case.
✅ 2. Regression coverage exercises the empty snapshot followed by a later `codex` report.

**Action taken**
- Approved.

PR #1488, opened 2026-07-08
fix(server): cap landing campaign trial starts

baixiaohang:
## Summary

Implements the confirmed landing campaign trial scope from #1484:

- caps new landing trial chats to 5 per user in a rolling 24-hour window by default
- checks the quota across all organizations for the same user
- keeps same campaign/repo replays idempotent, so they reuse the existing chat and do not consume quota
- adds a friendly `403` message when the free trial limit is reached
- gives landing trial chats a real default estimated token cap of `120_000`
- stops writing repo metadata on the trial agent; repo metadata remains on the trial chat

The global daily circuit breaker from the original issue is intentionally not included here because the follow-up product decision narrowed this PR to the per-user limit.

Refs #1484

## Tests

- `pnpm --filter @first-tree/server exec vitest run src/__tests__/landing-campaigns-start.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/shared exec vitest run src/config/__tests__/server-config.test.ts --maxWorkers=2`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/bootstrap.test.ts src/__tests__/build-app-validation.test.ts --maxWorkers=2`
- `pnpm exec biome check --write packages/server/src/__tests__/helpers.ts packages/server/src/__tests__/landing-campaigns-start.test.ts packages/server/src/__tests__/bootstrap.test.ts packages/server/src/__tests__/build-app-validation.test.ts packages/server/src/services/landing-campaigns/guards.ts packages/server/src/services/landing-campaigns/metadata.ts packages/server/src/services/landing-campaigns/start.ts packages/shared/src/config/__tests__/server-config.test.ts packages/shared/src/config/server-config.ts packages/shared/src/schemas/landing-campaign.ts`
- accidental broader run: `pnpm --filter @first-tree/server test -- --run src/__tests__/landing-campaigns-start.test.ts --maxWorkers=2` passed the full server suite (`178` files, `1744` tests)

### DIFF
--- a/packages/server/src/__tests__/bootstrap.test.ts
+++ b/packages/server/src/__tests__/bootstrap.test.ts
@@ -31,7 +31,12 @@ function makeTempConfigDir(): string {
 
 const baseServerConfig: ServerConfig = {
   channel: "dev",
-  growth: { landingPagesEnabled: false, landingCampaignMaxAgentTurns: 1, landingCampaignMaxEstimatedTokens: undefined },
+  growth: {
+    landingPagesEnabled: false,
+    landingCampaignMaxAgentTurns: 1,
+    landingCampaignMaxEstimatedTokens: 120_000,
+    landingCampaignMaxTrialsPerUserPer24Hours: 5,
+  },
   docs: { enabled: false },
   database: { url: process.env.DATABASE_URL ?? "", provider: "external" },
   server: { port: 0, host: "127.0.0.1", publicUrl: "https://first-tree.example" },
--- a/packages/server/src/__tests__/build-app-validation.test.ts
+++ b/packages/server/src/__tests__/build-app-validation.test.ts
@@ -16,7 +16,12 @@ import type { Config } from "../config.js";
  */
 const baseConfig: Config = {
   channel: "dev",
-  growth: { landingPagesEnabled: false, landingCampaignMaxAgentTurns: 1, landingCampaignMaxEstimatedTokens: undefined },
+  growth: {
+    landingPagesEnabled: false,
+    landingCampaignMaxAgentTurns: 1,
+    landingCampaignMaxEstimatedTokens: 120_000,
+    landingCampaignMaxTrialsPerUserPer24Hours: 5,
+  },
   docs: { enabled: false },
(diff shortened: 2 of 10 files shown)


yuezengwu (changes requested):
Goal/core changes reviewed: this PR adds a per-user rolling 24-hour quota for new landing campaign trial chats across organizations, preserves same campaign/repo replay idempotency, defaults the trial chat token budget to 120,000, and moves repo metadata off trial agent metadata so repo metadata remains on the trial chat.

Blocking finding:

- packages/server/src/services/landing-campaigns/start.ts:417-430 - The quota check runs inside an outer transaction that holds `pg_advisory_xact_lock(...)`, but the new chat is created with `createChat(app.db, ...)`. That starts work on the root pool instead of the transaction that owns the lock. Under concurrent landing starts, each request can hold one outer transaction connection while waiting for `createChat` to acquire another pool connection, so a burst at the pool limit can deadlock/starve the server. It also means the chat insert and owner membership rows are not atomic with the quota check. Please create the chat through the transaction handle that owns the advisory lock, or otherwise restructure so the existing-chat check, quota count, and insert run on one connection/transaction.

Data-shape note: no DB migration is involved, but trial agent metadata changes shape (`repo` is now optional and no longer written for new trial agents). I reviewed that against the stated goal and it looks intentional.

Local verification: I could not run `pnpm --filter @first-tree/server exec vitest run src/__tests__/landing-campaigns-start.test.ts --maxWorkers=2` in this fresh worktree because `vitest` is not installed. GitHub lint/type checks were green and the Test job was still in progress when I reviewed.

yuezengwu (approved):
Re-reviewed the synchronized update at `8f3b4500`.

The blocker from my prior review is fixed: `ensureTrialChatAndBootstrap` now passes the transaction-backed handle into `createChat`, so the existing-chat check, quota assertion, chat insert, and participant rows stay on the advisory-lock connection/transaction path.

No further findings from review. I did not rerun local vitest because this fresh worktree still has no installed `vitest` dependency; GitHub lint/type checks are green, and I am not waiting on the remaining CI per review policy.

PR #1501, opened 2026-07-08
fix(cli): avoid Windows cmd shims for npm and Codex

baixiaohang:
## Summary

- resolve Windows npm installs through `node.exe npm-cli.js` instead of spawning `.cmd` shims
- preserve a PATH + shell fallback for non-standard Node layouts without embedding an unquoted absolute path
- convert synchronous child-process spawn failures into normal installer failure results
- apply the shared npm invocation path to CLI self-upgrade, npm metadata probes, and Codex/Claude runtime installers
- keep the original Windows Codex native-binary resolution fix while rebasing onto current `main`
- normalize real paths in the Codex shim test to address the macOS `/var` vs `/private/var` review blocker

## Validation

- [x] affected-file Biome check
- [x] `pnpm typecheck` (10/10 Turbo tasks)
- [x] CLI focused tests: 27 passed
- [x] Client Codex tests: 17 passed, 2 platform-skipped
- [x] real Windows npm invocation smoke test (`npm --version` through `node.exe npm-cli.js`)
- [ ] `pnpm check` — blocked on Windows by the checkout-wide CRLF/LF mismatch (1594 pre-existing diagnostics); all 9 touched files pass Biome
- [ ] `pnpm test` — the existing Windows batch runner synchronously fails at `spawnSync vitest.cmd EINVAL`; direct full CLI Vitest runs but exposes existing Windows-only path/symlink/platform-mock failures. Focused changed-surface suites pass.

## Change Surface

- [x] `apps/cli` install and self-upgrade behavior

## Notes

The reported failure was reproduced on Node v22.16.0 with the exact installed path `C:\Program Files\nodejs\npm.cmd`. Directly spawning that shim throws `EINVAL`; adding `shell: true` alone is insufficient because cmd.exe truncates the unquoted absolute path at `C:\Program`.

### DIFF
--- a/packages/client/src/__tests__/codex-binary.test.ts
+++ b/packages/client/src/__tests__/codex-binary.test.ts
@@ -5,6 +5,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   CodexBinaryVerifyTransientError,
   createCodexClientWithBinaryFallback,
+  type FindCodexExecutableDeps,
   findCodexExecutableOnPath,
   formatCodexBinaryMissingMessage,
   isCodexBinaryMissingError,
@@ -275,7 +276,9 @@ describe("codex binary resolution", () => {
     chmodSync(chatGptCodex, 0o755);
     chmodSync(legacyCodex, 0o755);
 
-    const deps = {
+    const deps: FindCodexExecutableDeps = {
+      platform: "linux",
+      pathDelimiter: delimiter,
       loginShellPathDirs: () => [],
       wellKnownDirs: () => [],
       desktopAppDirs: () => [chatGptResources, legacyResources],
@@ -303,6 +306,8 @@ describe("codex binary resolution", () => {
       findCodexExecutableOnPath(
         { PATH: "" },
         {
+          platform: "linux",
+          pathDelimiter: delimiter,
           wellKnownDirs: () => [],
           loginShellPathDirs: () => [loginDir],
           desktopAppDirs: () => [appResources],
--- a/packages/client/src/__tests__/codex-binary.test.ts
+++ b/packages/client/src/__tests__/codex-binary.test.ts
@@ -1,4 +1,4 @@
-import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { chmodSync, mkdirSync, mkdtempSync, realpathSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { delimiter, join } from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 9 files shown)


packages/client/src/__tests__/codex-binary.test.ts
```diff
@@ -132,39 +132,96 @@ describe("codex binary resolution", () => {
...
+      {
+        platform: "win32",
+        arch: "x64",
+        pathDelimiter: ";",
+        loginShellPathDirs: () => [],
+        wellKnownDirs: () => [],
+      },
+    );
+
+    expect(found).toBe(native);
```
yuezengwu: This assertion is not portable on macOS: `createRequire(packageJson).resolve(...)` canonicalizes the temp path through `/private/var/...`, while `native` is built from `tmpdir()` as `/var/...`, so the new test fails locally even though both paths identify the same file. Please compare normalized/real paths, or assert a stable suffix plus existence, so the changed test passes across macOS and Linux.

yuezengwu (changes requested):
PR goal/core change: this fixes Windows Codex runtime resolution by avoiding `.cmd` npm shims as the SDK/runtime path and resolving them to the native `codex.exe`, while also routing npm install/update subprocesses through the Windows shell.

Requesting changes for one portability blocker in the new test. CI is green on Linux, but the changed client test fails on this macOS runner, so local `pnpm test` is not portable yet.

Validation I ran:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/codex-binary.test.ts --passWithNoTests` fails: expected `/var/...` but received `/private/var/...`
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/core-attention-update-extra.test.ts --passWithNoTests` passes

baixiaohang:
Updated and rebased onto current main. The macOS realpath review blocker is fixed, and the Windows npm self-upgrade root cause now uses node.exe + npm-cli.js with synchronous spawn-error handling. Ready for re-review.

packages/client/src/__tests__/codex-binary.test.ts
baixiaohang: Fixed in 1cef748e: the assertion now compares realpathSync(found) with realpathSync(native), so macOS /var and /private/var aliases resolve to the same canonical path. The focused Client suite passes locally, and CI Lint & Type Check is green.

baixiaohang:
Superseded by #1589, which carries the same fix on a fresh branch rebased onto current main. Closing this older PR as requested.

yuezengwu (approved):
Re-review outcome: approved.

Goal/core changes: this resolves Windows Codex runtime discovery by translating npm-generated `codex.cmd` shims to the native per-platform `codex.exe`. It also centralizes npm subprocess resolution so standard Windows Node installs use `node.exe + npm-cli.js`, keeps a PATH/shell fallback for non-standard layouts, and maps synchronous spawn failures into the existing update/runtime-install result taxonomy.

The previous macOS portability blocker is fixed: the shim test now compares canonical real paths, and it passes locally. I found no new blocking issue. This does not change DB/schema, core data structures, or public CLI/help output.

Validation:
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/codex-binary.test.ts --passWithNoTests` — 19/19 passed
- CLI npm-invocation and runtime-installer suites passed; all changed update assertions passed
- one unrelated `freebsd-x64` expectation in `core-attention-update-extra.test.ts` still fails on this arm64 host, and I reproduced the identical failure on the PR base commit
- `pnpm typecheck` — 10/10 Turbo tasks passed
- Biome check on all 9 changed files passed
- `git diff --check` passed

PR #1506, opened 2026-07-08
fix(portable): make install.sh recover existing clients

### LINKED ISSUE #1502: install.sh must be idempotent and perform upgrades (universal recovery entrypoint for Settings "reconnect")
## Background

The Settings page offers a **reconnect** flow for when a user's client shows up as offline. From the server side we cannot distinguish *why* the client went offline — `first-tree logout`, uninstalling the client (e.g. `npm uninstall` of the npm-distributed client), or simply a stopped daemon all look identical. So the reconnect guidance is uniform: we always tell the user to run `install.sh` and then `first-tree login`.

That makes `install.sh` (`scripts/portable/install.sh`, served from the downloads host) not just a fresh-install script but the **universal recovery entrypoint** — it will routinely be re-run on machines in every possible intermediate state.

## Requirement

`install.sh` must be **idempotent** and must **perform upgrades**: running it from any starting state should converge to "latest version installed, shim on PATH, daemon installed and running", without erroring out and without corrupting existing state.

States it must converge from, at minimum:

1. **Fresh machine** — nothing installed (current happy path).
2. **Healthy install, re-run** — already on latest and running; re-run is a clean no-op (no duplicate PATH blocks, no broken shims, no service errors).
3. **Older version installed** — upgrade to latest instead of failing or silently keeping the old version.
4. **After `first-tree logout`** — binary and daemon present, credentials gone; install.sh must not fail or wipe anything, and the follow-up `first-tree login` completes the recovery.
5. **After uninstall with residue** — binary removed but leftovers remain (daemon/service registration, config dirs, logs, stale PATH block, or a lingering npm-installed binary shadowing the portable shim); install.sh must reinstall cleanly over the residue and end with the portable binary winning on PATH.
6. **Daemon stopped** — everything installed but the service is down; install.sh ends with the daemon (re)started.

Guardrails:

- Must not destroy state it doesn't have to: workspaces, agent data, and still-valid auth must survive a re-run.
- The unconditional `install.sh` + `first-tree login` pair must be safe even when only one of the two was actually needed.

## Known related failure mode

On the npm-based update path we have seen residual npm temp directories (`.first-tree-staging-*` inside the global `node_modules`) permanently wedge upgrades with `ENOTEMPTY` until manually removed. The recovery path should detect/clean this class of residue rather than fail the same way on every re-run — a recovery entrypoint that itself needs manual recovery defeats the purpose.

bestony:
## Summary
- make `scripts/portable/install.sh` keep portable shims ahead of shadowing npm installs and clean stale npm temp residue for the channel package
- add a hidden `daemon ensure-service` recovery hook that refreshes/starts the supervised daemon when credentials already exist, restarts an originally active service after a drifted unit refresh, and no-ops safely before `login`
- cover installer re-runs, PATH repair, npm residue cleanup, and the hidden daemon hook in tests

## Tests
- `sh -n scripts/portable/install.sh`
- `pnpm check` (passes; existing warnings remain in unrelated files)
- `pnpm typecheck`
- `pnpm --dir apps/cli exec vitest run tests/portable-builder.test.ts src/__tests__/cli-command-matrix-extra.test.ts src/__tests__/daemon-refresh-unit.test.ts src/__tests__/command-registration-smoke.test.ts`
- `env -u FIRST_TREE_AGENT_ID -u FIRST_TREE_AGENT_SLUG -u FIRST_TREE_CHAT_ID -u FIRST_TREE_CLIENT_ID -u FIRST_TREE_DOC_AGENT_HOME -u FIRST_TREE_DOC_BASE -u FIRST_TREE_DOC_REPO_LOCAL_PATH -u FIRST_TREE_HOME -u FIRST_TREE_INBOX_ID -u FIRST_TREE_PROVIDER -u FIRST_TREE_RUNTIME_SESSION_TOKEN -u FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE -u FIRST_TREE_SERVER_URL -u FIRST_TREE_SERVICE_MODE -u FIRST_TREE_SWITCH_DRAIN_VERSION -u FIRST_TREE_WORKSPACES_ROOT TURBO_CONCURRENCY=1 pnpm test`

## Context Tree
- Companion Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/693

Closes #1502

### DIFF
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -312,6 +312,32 @@ describe("daemon utility commands", () => {
     expect(output).toContain("Ctrl+C");
     expect(output).toContain("daemon start");
   });
+
+  it("restarts an active systemd service when ensure-service refreshes a drifted unit", async () => {
+    setPlatform("linux");
+    coreMocks.installClientService.mockClear();
+    coreMocks.restartClientService.mockClear();
+    coreMocks.isServiceSupported.mockReturnValue(true);
+    coreMocks.loadCredentials.mockReturnValue({ refreshToken: "refresh" });
+    coreMocks.getClientServiceStatus
+      .mockReturnValueOnce({ state: "active", platform: "systemd", detail: "pid 123" })
+      .mockReturnValueOnce({ state: "active", platform: "systemd", detail: "pid 456" });
+    coreMocks.isServiceUnitDriftDetected.mockReturnValueOnce(true);
+    coreMocks.installClientService.mockReturnValueOnce({
+      platform: "systemd",
+      unitPath: "/tmp/first-tree.service",
+      state: "active",
+    });
+    coreMocks.restartClientService.mockReturnValueOnce({ ok: true });
+
+    await runDaemon(["ensure-service"]);
+
+    expect(coreMocks.installClientService).toHaveBeenCalledTimes(1);
+    expect(coreMocks.restartClientService).toHaveBeenCalledTimes(1);
+    expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain(
+      "systemd service refreshed and restarted",
+    );
+  });
 });
 
 describe("agent admin and local commands", () => {
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -41,6 +41,7 @@ const coreMocks = vi.hoisted(() => ({
   installClientService: vi.fn(),
   isServiceSupported: vi.fn(),
   isServiceUnitDriftDetected: vi.fn(),
+  loadCredentials: vi.fn(),
   refreshClientServiceUnitForUpdate: vi.fn(),
   removeLocalAgent: vi.fn(),
   restartClientService: vi.fn(),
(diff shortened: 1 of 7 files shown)


yuezengwu (changes requested):
Requesting changes for one recovery-path blocker.

`apps/cli/src/commands/daemon/ensure-service.ts:36-45` handles the active+drift case by calling `installClientService()`, but on Linux that maps to `installSystemd()` (write unit + `daemon-reload` + `systemctl --user enable --now`). `enable --now` does not restart an already active unit, so a rerun of `install.sh` can install the new portable bits and refresh the unit while the supervised daemon keeps running the old process until some later restart.

That misses the #1502 convergence requirement for existing/running clients: after the universal recovery entrypoint completes, the daemon should be running the repaired/latest command, not just have a refreshed unit on disk. `upgrade` already does the safer sequence: refresh the unit, then call `restartClientService()` when the original service state was active. `ensure-service` needs the same active+drift restart/verification path (with tests for systemd active+drift), while preserving the active+no-drift no-op.

yuezengwu (approved):
Approved. The new `37d0d547` update addresses my previous blocker: `daemon ensure-service` now restarts an originally active service after refreshing a drifted unit, and the added test covers the systemd active+drift path.

I rechecked the final recovery flow against #1502 and the Context Tree daemon/portable-install constraints. I do not see remaining blocking issues in this PR. CI is green at review time.

bestony:
Addressed the active+drift recovery blocker in `37d0d547`.

- `daemon ensure-service` now refreshes the service unit, explicitly restarts the service when the original state was active, and verifies the post-restart status is still active.
- Added coverage for the systemd active+drift path so `ensure-service` cannot regress to only `enable --now` while leaving the old daemon process running.

Validation:
- `pnpm --dir apps/cli exec vitest run src/__tests__/cli-command-matrix-extra.test.ts src/__tests__/daemon-refresh-unit.test.ts`
- `pnpm check` (passes; existing unrelated warnings only)
- `pnpm typecheck`
- GitHub CI and CodeQL are green on the updated head.

PR #1508, opened 2026-07-08
Portable installs: define bundled Node upgrade path

### LINKED ISSUE #1504: Portable installs: bundled Node.js upgrade path and policy (currently latest-v24.x)
## Background

The portable install path (`scripts/portable/install.sh`) bundles a Node.js runtime so users can run First Tree without installing Node themselves. The bundled version is pinned in `scripts/portable/build-portable.mjs`:

```js
export const DEFAULT_NODE_VERSION = "latest-v24.x";
```

`latest-v24.x` is resolved against `nodejs.org/dist` **at release build time**, so each published tarball carries whatever the newest 24.x was when that release was cut.

The question this issue tracks: **if we ever need to move users to Node 25 (or any newer major), do we have a path — and what is the policy for exercising it?**

## How it works today (verified against `main`)

The mechanics for swapping the runtime already exist. Each portable release tarball is fully self-contained:

```
<prefix>/versions/<version>/
  node/   <- bundled runtime
  app/    <- built CLI
```

Portable self-update (`installPortableSpec` in `apps/cli/src/core/update.ts`) downloads the target version's tarball, verifies its checksum, extracts it into `versions/<version>` (including a fresh `node/`), atomically flips the `current` symlink, rewrites the shims, and the daemon restarts (`SELF_RESTART_EXIT_CODE = 75` → service manager relaunches through the shim).

So a user who installed when we bundled Node 24 is **not** frozen on their install-time runtime: every release upgrade replaces the bundled Node with whatever that release was built against. Bumping to Node 25 mechanically means changing the pin (`DEFAULT_NODE_VERSION`, plus the explicit `--node-version latest-v24.x` in the `portable-smoke` CI job) and cutting a release.

## Gaps

1. **No bump policy or owner.** The pin only moves when someone moves it. There is no documented trigger (Node 24 end-of-life, needing 25-only features, a Node security advisory) or owner for bumping the bundled major. Related: users only receive Node *patch* security fixes when we cut a First Tree release — that coupling is fine, but it should be a stated policy, not an accident.
2. **Cross-major upgrade path is unverified.** During an update the executor runs on the *old* bundled runtime (24) while installing a tarball whose `node/` is 25. Expected to work (tar + symlink flip + service restart, no in-place execution of the new runtime), but no test covers a cross-Node-major portable upgrade — `portable-smoke` builds and verifies a single artifact, it never exercises old-version → new-version handover.
3. **npm install mode has no runtime path at all.** npm-mode self-update swaps the npm package and never touches the user's system Node. If `engines` minimum ever rises above a user's installed Node, there is no self-serve fix — we'd need at least detection + clear guidance (upgrade system Node, or migrate to the portable install).

## Proposed acceptance

- [ ] Document the bundled-Node upgrade policy: when we bump (EOL / feature / security), who owns it, and where the pins live.
- [ ] Add a smoke test (or a documented manual release-verification step) for a cross-Node-major portable upgrade before the first real bump.
- [ ] Decide the npm-mode story for a raised Node minimum (preflight check + warning, migration guidance, or both).

bestony:
## Summary

- Document the portable bundled Node.js upgrade policy, including bump triggers, owner, pin locations, and npm-mode behavior.
- Generalize portable build resolution from a hard-coded `latest-v24.x` branch to `latest-v<major>.x` so a future Node 25 line can be selected by changing the pins.
- Add coverage for cross-Node-major portable self-update handover and npm-mode `engines.node` preflight guidance.
- Update public install docs to describe portable install as the default path and npm global install as the operator/fallback path.

## Context Tree

- Draft tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/694

## Verification

- `pnpm --filter first-tree-dev exec vitest run src/__tests__/update-portable-install.test.ts tests/portable-builder.test.ts src/__tests__/core-attention-update-extra.test.ts`
- `pnpm exec biome check scripts/portable/build-portable.mjs apps/cli/src/core/update.ts apps/cli/src/__tests__/update-portable-install.test.ts apps/cli/src/__tests__/core-attention-update-extra.test.ts apps/cli/tests/portable-builder.test.ts .github/workflows/ci.yml README.md docs/quickstart.md docs/onboarding-guide.md docs/cli-reference.md docs/development/portable-node-runtime.md`
- `git diff --check`

Closes #1504.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,16 @@ At the "connect a computer" step, setup gives you the command to install the
 CLI and link the machine:
 
 ```bash
-npm install -g first-tree
-first-tree login <connect-token>
+# Use the exact command shown in the web console; production portable installs
+# have this shape:
+curl -fsSL https://downloads.first-tree.ai/prod/install.sh | sh
+~/.local/bin/first-tree login <connect-token>
 ```
 
-The binary lives at `first-tree`; a short alias `ft` is also installed.
+The portable installer bundles Node.js, so new users do not need to install
+Node separately. The npm global install path (`npm install -g first-tree`) is
+still supported for operators and fallback installs; npm mode uses your system
+Node runtime.
 
 ## CLI
 
@@ -167,6 +172,7 @@ Run `first-tree <namespace> --help` for the full list under any namespace.
 - [Onboarding Guide](docs/onboarding-guide.md) — CLI flow, SDK, troubleshooting
 - [CLI Reference](docs/cli-reference.md) — every command and environment variable
 - [Observability](docs/observability.md) — logs and OpenTelemetry traces
+- [Portable Node Runtime](docs/development/portable-node-runtime.md) — bundled Node policy for portable releases
 - [docs/development/](docs/development/) — contributor reference
 - [docs/troubleshooting/](docs/troubleshooting/) — environment-specific gotchas
 - [docs/migration/](docs/migration/) — coming from `first-tree@0.4.x`
--- a/apps/cli/src/__tests__/core-attention-update-extra.test.ts
+++ b/apps/cli/src/__tests__/core-attention-update-extra.test.ts
@@ -109,6 +109,48 @@ describe("core update helpers", () => {
     await expect(latestPromise).resolves.toEqual({ ok: true, mode: "global", installedVersion: null });
   });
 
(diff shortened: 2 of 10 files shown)


bestony:
Superseded by #1509 using a branch name that passes the CI branch-name gate.

PR #1512, opened 2026-07-08
Fail closed on unusable runtime token file

yuezengwu:
## Summary

- fail closed when `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE` is set but missing/unreadable
- fail closed when the token file is empty instead of falling back to a stale env token
- add regression coverage that missing/empty token files do not construct an SDK with the inherited env token

## Verification

- `pnpm --filter first-tree-dev test -- src/__tests__/agent-shared-helpers-extra.test.ts` (Vitest collected full CLI suite: 70 files / 624 tests passed)
- `pnpm exec biome check apps/cli/src/commands/_shared/local-agent.ts apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts`
- `pnpm --filter @first-tree/server test -- src/__tests__/agent-bind-authorization.test.ts` (Vitest collected full server suite: 178 files / 1756 tests passed)

### DIFF
--- a/apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts
@@ -154,6 +154,49 @@ describe("local agent shared helpers", () => {
     }
   });
 
+  it("fails clearly instead of falling back when the runtime session token file is missing", async () => {
+    const dir = mkdtempSync(join(tmpdir(), "first-tree-token-"));
+    try {
+      const tokenFile = join(dir, "missing.token");
+      const { createSdk } = await import("../commands/_shared/local-agent.js");
+
+      process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN = "stale-runtime-token";
+      process.env.FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE = tokenFile;
+
+      expect(() => createSdk("nova")).toThrow();
+      expect(outputMocks.fail).toHaveBeenLastCalledWith(
+        "RUNTIME_SESSION_TOKEN_FILE_UNREADABLE",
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1512

**Recommendation**: approve
- Rationale: The change fails closed when an explicit runtime session token file is unusable, preventing stale inherited env tokens from authorizing agent-scoped HTTP.

**Risk level**: A
- Path baseline: includes `apps/cli/**`.
- Semantic lift: none beyond the CLI runtime-token helper; no DB/schema changes.

**PR summary**
- Problem: Agent-scoped CLI commands running inside a First Tree runtime could fall back to an inherited `FIRST_TREE_RUNTIME_SESSION_TOKEN` when the higher-priority token file was missing, unreadable, or empty.
- Approach: `createSdk` now treats an explicitly configured `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE` as authoritative: successful reads still win over the env token, while unreadable or empty files produce clear CLI failures before constructing the SDK.
- Impacted modules: `apps/cli/src/commands/_shared/local-agent.ts`, `apps/cli/src/__tests__/agent-shared-helpers-extra.test.ts`.

**Review findings**
✅ No blocking findings. The fail-closed behavior matches the runtime binding boundary: local mirrors and stale inherited credentials should not silently authorize agent-scoped calls.
✅ Regression coverage verifies missing and empty token files do not construct `FirstTreeHubSDK` with the stale env token.
⚠️ Non-blocking follow-up: the runtime token writer still writes directly to the final file path; an atomic write could reduce transient empty-file reads during reconnect updates, but that is outside this PR's safety fix.

**Action taken**

PR #1513, opened 2026-07-08
feat(web): add-repo-to-App recovery CTA on unreadable Context Tree

Gandy2025:
## Problem

After a team builds a Context Tree, the tree repo must be covered by the org's First Tree GitHub App installation for event tracking, web snapshots, and the Context Tree reviewer to work — all of them read through an installation-scoped token.

For a **"Selected repositories"** installation where the tree repo was created by a **user identity** (CLI `first-tree tree init`, or the server User-owner path), the new repo is **not** auto-added to the installation. The App can't add itself (no self-add API) and the user's token can't either (`/user/installations/*` → 403), so the fix is a manual admin action on GitHub.

Today the Context tab's `UnavailableState` renders a dead-end message in exactly this case ("First Tree cannot read the team Context Tree yet…") with no action — the admin has no idea they need to add the repo to the App.

## Change

In `UnavailableState` (bound-but-unreadable branch):

- **Admin** → an **"Add repo to the GitHub App"** button deep-linking to the installation settings page (`manageUrl` from the existing admin-only `getGithubAppInstallation` endpoint).
- **Member** → an "ask an admin to add it to the GitHub App" line (no button — members can't fix it and can't reach the admin-only endpoint).

### Notes on scope / correctness
- The `manageUrl` fetch is gated to `!preview && isAdmin && !!snapshot.repo`, so it never fires for members or in the preview gallery (the endpoint is admin-only, so this also avoids a 403).
- The button renders only once `manageUrl` resolves — when no installation is bound at all (404 → null) there is no dead button, and the copy stays a plain problem statement rather than dangling an "add it" instruction the admin can't act on here.
- No auto-fix is possible (documented GitHub App constraint); this is the guide-to-settings path. Reactive by design: this one surface is where the symptom lands regardless of how the tree was built (CLI or web).
- Neutral `default` button variant, not brand-green `cta` — per `button.tsx`, green is reserved for hero/creation moments, not recovery actions.

## Tests
- `context-page-dom.test.tsx`: admin CTA (href/target/rel + copy), member no-CTA + admin-only endpoint never called, and the no-installation-bound case (no dead button).
- Full web suite green (1139 tests), typecheck + `lint:tokens` + biome clean.

## Verification
Rendered the state in a headless browser off a token-faithful mock (three states: current dead-end, admin CTA, member message). Diff has no layout-measurement or element-type-switch logic — a conditional anchor inside the existing static panel.

- Optional: a `/context-tree/coverage` probe to distinguish "repo not covered" from a transient clone failure, so the CTA shows only for a genuine coverage gap.
- Optional: catch the build-time 409 (`context_tree_repo_access_required`) on the web provision path and render the same deep-link at that moment.


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1513

**Recommendation**: approve
- Rationale: The change is narrowly scoped to the unreadable bound Context Tree state and gives admins a valid recovery path without exposing the admin-only installation endpoint to members or previews.

**Risk level**: C
- Semantic lift: none; this is Context tab UI/query gating and does not touch subscriptions, WS frame parsing, or shared data-layer behavior.

**PR Summary**
- Problem: When a team has a bound Context Tree repo that First Tree cannot read because the GitHub App installation does not cover it, admins currently see a dead-end Context tab state instead of a concrete recovery action.
- Approach: Admins now see a deep link to the GitHub App installation settings once `manageUrl` resolves, while members get ask-admin guidance and never call the admin-only endpoint. The no-installation-bound case stays a plain problem statement without a dead CTA.
- Impacted modules: `packages/web/src/pages/context.tsx`, `packages/web/src/pages/__tests__/context-page-dom.test.tsx`.

**Review findings**
✅ 1. The `getGithubAppInstallation` query is gated by non-preview admin status plus a bound repo, and the render path still requires both `canManageInstall` and `manageUrl`, so cached/query-disabled states do not expose a member CTA.
✅ 2. The DOM tests cover the important branches: admin CTA with external-link attributes, member no-CTA/no admin fetch, and no-installation-bound no dead button.
✅ 3. Current GitHub checks are green, including Lint & Type Check, Test, and CodeQL.

**Action taken**
- Approved.

yuezengwu (changes requested):
实现目标和核心改动：这个 PR 在 Context tab 的 bound-but-unreadable `UnavailableState` 里，为 admin 拉取现有 GitHub App installation 的 `manageUrl` 并展示 “Add repo to the GitHub App” 深链；member 只看到让 admin 处理的说明。没有 DB/schema 或核心数据结构变更。

Requesting changes on one blocker:

- `packages/web/src/pages/context.tsx:82` / `packages/web/src/pages/context.tsx:677`: `canManageInstall` is gated only on `!!snapshot.repo`, but `snapshot.repo` is just the configured binding, not evidence that the unavailable state is a GitHub App repository-coverage failure. The server already has the narrower contract in `isGithubRemoteBinding` + `decorateSnapshotWithMintGuidance`: App remediation is only appropriate for GitHub-hosted bindings with a mint/install failure, and it intentionally does not decorate local-path bindings, non-GitHub remotes, bad branches, or ordinary clone failures. With this PR, any bound unavailable snapshot for an admin fetches the installation endpoint and can render “Add repo to the GitHub App”; members always get “Ask an admin to add it to the GitHub App.” That regresses the existing no-misdirection boundary and can send users to GitHub for failures that GitHub App coverage cannot fix. Please gate the CTA/copy on a server-provided recovery reason/action (or equivalent structured signal from the snapshot route), and keep the generic sync-unavailable copy for non-App causes.

Verification: reviewed the Context Tree nodes for Context tab/GitHub App constraints, inspected the changed frontend/tests and server snapshot/token guidance code, and ran `git diff --check origin/main...HEAD`. GitHub checks are currently green on `cee1c0e9`. I attempted the focused web DOM test locally, but this review worktree has no `node_modules`, so `vitest` was not available here.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1513

- Rationale: The CTA is rendered for every bound-but-unreadable Context Tree as long as the org has a GitHub App installation, but several unreadable states are not recoverable by adding a repo to the App.

**Risk level**: C
- Semantic lift: none; this is Context tab UI/query gating and does not touch subscriptions, WS frame parsing, or shared data-layer behavior.

**PR Summary**
- Problem: Admins currently hit a dead-end Context tab state when a bound Context Tree repo is unreadable because a selected-repositories GitHub App installation does not include that repo.
- Approach: The PR fetches the existing admin-only installation `manageUrl` and shows admins a GitHub App settings link, while members get ask-admin guidance and do not call the admin endpoint.
- Impacted modules: `packages/web/src/pages/context.tsx`, `packages/web/src/pages/__tests__/context-page-dom.test.tsx`.

**Review findings**
❌ 1. The CTA is over-broad and will misdirect admins for non-coverage failures. `canManageInstall` is true for any non-preview admin with any bound `snapshot.repo`, and the button renders once any installation `manageUrl` exists. The snapshot route currently only has a coarse `snapshotStatus: "unavailable"` plus text detail; the server-side GitHub App guidance is deliberately gated to GitHub-remote/token-related cases, while invalid branch names, local/non-GitHub fetch failures, transient clone errors, and other sync failures keep their original detail. With this PR, those cases still get an "Add repo to the GitHub App" action even though changing the installation cannot fix them. The coverage probe / structured recovery cause described as a follow-up needs to be part of this PR, or the frontend needs another server-provided signal such as `recoveryAction: "manage_github_app_installation"` before rendering the CTA. [R5 / `packages/web/src/pages/context.tsx`:82, `packages/web/src/pages/context.tsx`:708]

**Action taken**
- Requested changes, superseding my earlier approval after checking the existing snapshot recovery contract.

Gandy2025:
Thanks both — the blocker was right, and I've reworked the change to gate on a **server-provided, probed** signal instead of the `!!snapshot.repo` heuristic. Pushed in `667ef755`.

## What changed

**shared** — new `contextTreeRecoveryActionSchema = z.literal("manage_github_app_installation")` and a nullish `recoveryAction` field on `contextTreeSnapshotSchema`.

**server** — new `resolveContextTreeRecoveryAction(snapshot, binding, mintResult)` in `services/github-app-token.ts`, wired into **both** snapshot routes (`api/context-tree-snapshot.ts`, `api/orgs/context-tree-snapshot.ts`). It reuses the existing `isGithubRemoteBinding` boundary and **probes the repo with the minted installation token**, returning `"manage_github_app_installation"` **only** when the App genuinely can't read it. Otherwise `null`, leaving the generic copy.

**web** — `context.tsx` now gates `canManageInstall`, the CTA, and the App-remediation copy on `snapshot.recoveryAction === "manage_github_app_installation"`. Every other unavailable cause keeps the original generic sync copy.

## Addressing each point

- **@yuezengwu — "gate on a server-provided recovery reason/action + coverage probe in this PR; keep generic copy for non-App causes."** Done exactly. The probe is `getRepository(token, owner, repo)` on the unavailable + GitHub-remote + minted path; a 404 (App can't see the repo) is the only trigger. Bad branch / transient clone / local / non-GitHub / suspended / no-install all keep the generic copy. This mirrors the narrowness of `decorateSnapshotWithMintGuidance` you pointed at.
- **@baixiaohang — R5 "over-broad; needs a signal such as `recoveryAction: \"manage_github_app_installation\"`."** Implemented under exactly that name. `canManageInstall` is no longer true for "any bound repo" — only for a server-confirmed coverage gap on a non-preview admin.

## One deliberate precision: 404 only, not 403

The probe treats **404** as the coverage-gap signal (GitHub's "this installation token can't see the repo"). It intentionally does **not** treat a **403** as a gap — 403 is ambiguous (rate limit, SAML enforcement, a missing permission) on a repo the App may already cover, so "add repo" would misdirect. Any 403 / 5xx / transient error falls through to the generic copy (fail-safe: only ever a missing CTA, never a wrong one).

## Tests
- server (`github-app-token.test.ts`): 404 → action; 403 / readable-200 / not-unavailable / non-GitHub / no-token / 5xx → null; full `https://…​.git` URL parse.
- web (`context-page-dom.test.tsx`): coverage-gap → CTA; **non-coverage (`recoveryAction` unset) → generic copy, no CTA, no admin-only fetch**; member → ask-admin, no fetch; gap but manage URL null → no dead button.
- Full suites green locally: shared 508, server 1763, web 1140; typecheck + biome + lint:tokens clean.

Zero extra API calls on the healthy path (the probe short-circuits before any fetch when the snapshot isn't unavailable). Please take another look — re-requesting review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1513

- Rationale: The original frontend heuristic has been replaced with a server-provided probed signal, but the new probe still returns the manage-installation action for 404s on all-repositories installations where adding a repo to the App cannot be the fix.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` and `packages/shared/**`; no DB/schema change and no listed core service path trigger.
- Semantic lift: none beyond the explicit Context Tree snapshot recovery API surface in this PR.

**PR Summary**
- Problem: Admins need a concrete recovery path when a bound Context Tree repo is unreadable specifically because a selected-repositories GitHub App installation does not include that repo.
- Approach: The updated PR adds a structured `recoveryAction` to the snapshot, resolves it server-side by probing repository readability with the minted installation token, and gates the Web CTA/copy on that action instead of inferring from `snapshot.repo`.
- Impacted modules: `packages/shared/src/schemas/context-tree.ts`, `packages/server/src/services/github-app-token.ts`, both context-tree snapshot routes, and `packages/web/src/pages/context.tsx` with DOM tests.

**Review findings**
❌ 1. `resolveContextTreeRecoveryAction` should not return `manage_github_app_installation` for `repositorySelection: "all"`. The selected-repositories coverage gap this CTA fixes only exists when the installation token was minted for a selected-repos installation. On an all-repositories installation, a 404 from `GET /repos/:owner/:repo` means the repo is gone/renamed, the binding points at the wrong owner, or another non-coverage condition; GitHub App settings cannot add a repo that is already covered by an all-repos install. As written, any unavailable GitHub binding with a minted `all` token and a repo 404 will surface the same “Add repo to the GitHub App” recovery action, reintroducing the no-misdirection bug for that class. Please gate the 404 action on `mintResult.repositorySelection === "selected"` and add a regression test for `repositorySelection: "all"` + 404 returning `null`. [R5 / `packages/server/src/services/github-app-token.ts`:186]

**Action taken**
- Requested changes.

yuezengwu (changes requested):
实现目标和核心改动：这轮更新把原来的 `!!snapshot.repo` 前端启发式改成了 server-provided `snapshot.recoveryAction`，并在 server 侧用安装 token probe repo 可读性；前端只有在 `recoveryAction === "manage_github_app_installation"` 时才展示 App recovery copy/CTA。没有 DB/schema 迁移；但有 shared API snapshot contract 变更（新增 nullish `recoveryAction` 字段），需要按这个契约继续收窄。

Requesting changes on one remaining blocker:

- `packages/server/src/services/github-app-token.ts:181`: the probe still returns `manage_github_app_installation` on any 404 from `GET /repos/:owner/:repo`, regardless of the installation token's `repositorySelection`. For an **all-repositories** installation, a 404 cannot mean "this selected-repos installation is missing that repo"; it means the configured repo is gone/renamed, belongs to a different GitHub account than this installation, or is otherwise not reachable by this installation. Opening the current installation settings with "Add repo to the GitHub App" will not fix those cases, so this still violates the no-misdirection boundary we were trying to restore. Please require `mintResult.repositorySelection === "selected"` before treating a 404 as `manage_github_app_installation` (ideally short-circuit before the extra probe), and add a test where `repositorySelection: "all"` + 404 returns `null`.

Verification: re-read the Context Tree Context tab/GitHub App constraints, inspected the updated shared/server/web diff at `667ef755`, checked the `getRepository` error model, and ran `git diff --check origin/main...HEAD` successfully. Local focused tests were not runnable in this fresh review worktree because `node_modules` is absent; GitHub checks currently have all non-test jobs green, with the Test job still in progress at review time.

Gandy2025:
Fixed in `de8f0a88` — you're both right, thanks.

`resolveContextTreeRecoveryAction` now **short-circuits before the probe** unless `mintResult.repositorySelection === "selected"`:

```ts
if (!mintResult.ok) return null;
// Only a selected-repositories installation can have the coverage gap this
// action fixes. An all-repositories install already covers every repo in the
// account, so a 404 there is gone/renamed/wrong-owner — not addable.
if (mintResult.repositorySelection !== "selected") return null;
```

So on an **all-repositories** install a 404 (gone / renamed / different-owner) now returns `null` → generic sync copy, no "Add repo" CTA, and no wasted probe call. The action fires only for a `selected` install whose token 404s on the bound repo — the one case adding the repo to the installation actually fixes.

Added the requested regression test: `repositorySelection: "all"` + 404 → `null`, and asserts the probe fetcher was never called (short-circuit). `github-app-token.test.ts` is now 23 tests, all green (typecheck + biome clean). Re-requesting review.

yuezengwu (approved):
实现目标和核心改动：这个 PR 为 Context tab 的 bound-but-unreadable Context Tree 状态增加了一个 GitHub App recovery path。最终形态不再由前端用 `snapshot.repo` 推断，而是 shared snapshot contract 新增 nullish `recoveryAction`，server 在 unavailable + GitHub remote + minted selected-install token 的路径上 probe repo 可读性；Web 只有在 `recoveryAction === "manage_github_app_installation"` 时展示 admin 的 “Add repo to the GitHub App” CTA 和 member 的 ask-admin copy。

No blocking findings.

The previous blocker is addressed: `resolveContextTreeRecoveryAction` now short-circuits unless `mintResult.repositorySelection === "selected"`, so an all-repositories installation with a 404 keeps the generic sync-unavailable state and does not trigger the manage-installation CTA. The regression test covers `all` + 404 returning `null` and not calling the probe. The earlier over-broad frontend gate is also gone; the admin-only `getGithubAppInstallation` fetch is gated on the server-provided recovery action plus non-preview admin status.

There is no DB/schema migration. There is a shared API contract addition (`ContextTreeSnapshot.recoveryAction`), but it is nullish and the server routes populate it explicitly.

Verification: re-read the Context Tree Context tab/GitHub App constraints, inspected the latest diff at `de8f0a88`, checked the `getRepository` error path and the selected/all regression tests, and ran `git diff --check origin/main...HEAD`. I did not run local Vitest because this fresh review worktree has no `node_modules`; GitHub checks were still partly pending when I reviewed.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1513

**Recommendation**: approve
- Rationale: The recovery action is now server-provided and narrowly probed for the selected-repositories coverage gap, so the Context tab no longer misdirects other unreadable states to GitHub App settings.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` and `packages/shared/**`, with no DB/schema change and no listed core service path trigger.
- Semantic lift: none beyond the explicit Context Tree snapshot recovery API surface added here.

**PR Summary**
- Problem: Admins need a concrete recovery path when a bound Context Tree repo is unreadable specifically because a selected-repositories GitHub App installation does not include that repo.
- Approach: The PR adds a nullish `recoveryAction` field to the snapshot contract, resolves it server-side by probing repo readability only for unavailable GitHub remote snapshots with a minted selected-installation token, and gates Web copy/CTA on that action.
- Impacted modules: `packages/shared/src/schemas/context-tree.ts`, `packages/server/src/services/github-app-token.ts`, both context-tree snapshot routes, and `packages/web/src/pages/context.tsx` with focused tests.

**Review findings**
✅ 1. The previous over-broad frontend heuristic is gone; `getGithubAppInstallation` now only runs for non-preview admins when the server provided `recoveryAction === "manage_github_app_installation"`.
✅ 2. The previous all-repositories blocker is fixed: `resolveContextTreeRecoveryAction` short-circuits unless `repositorySelection === "selected"`, so `all` + 404 keeps the generic sync-unavailable state and avoids the probe.
✅ 3. Regression coverage includes selected 404 → action, all 404 → null/no probe, 403/5xx/readable/non-GitHub/no-token/not-unavailable → null, plus Web DOM coverage for CTA, no-CTA generic copy, member copy, and null manage URL.

**Action taken**
- Approved. I did not run local tests per review-mode constraints; at review time the GitHub Test job was still in progress while the other visible checks were green.

PR #1514, opened 2026-07-08
feat: guide GitHub App coverage after tree build (seed skill)

Gandy2025:
## Problem

When an admin builds a team Context Tree, nothing in the flow guides them to
install/cover the First Tree GitHub App:

- `first-tree-seed` runs `tree init` assuming an App installation already
  exists, and `tree init` silently falls back to the creator's personal
  account when none does.
- Neither `first-tree-welcome` nor `first-tree-seed` surfaces App coverage in
  chat, and onboarding no longer forces a connect-code step.

Result: a tree can be built + bound but stay **invisible** to the web view and
Context Tree reviewer (both read the tree through the App installation token),
and a repo created under a personal account can't be covered if the App is
later installed on a different org (placement trap). This is the not-yet-landed
"显式校验/引导" failure boundary called out in the `github-app-slim-refactor`
proposal.

## Fix (guidance only — never blocks)

Add a "Surface App coverage after creating the tree" step to `first-tree-seed`
State A. After `tree init`, when the repo is not covered, the agent leaves a
prominent, non-blocking message:

- **No App installed:** a clickable link to the web console's GitHub settings,
  built from the `Server:` value of `first-tree agent status` +
  `/settings/github` (absolute `https://…` renders clickable; the chat link
  guard drops bare relative paths), told to click **Install on GitHub**, plus
  the personal-account placement caveat. Discloses the same-origin assumption
  and gives a words-only "Settings → GitHub" fallback.
- **Selected-repositories / suspended install:** relay the installation-settings
  URL `tree init` already prints.
- Never fabricate a raw GitHub App install URL — the install must run through
  the web console so it binds back to the org.

The build is never gated on coverage. Also corrects the State A copy (repo owner
when the App isn't installed) and disambiguates an overloaded "installation
check" phrase.

Scope: `skills/first-tree-seed/SKILL.md` only (+ VERSION 0.2.2 → 0.2.3). No
server/CLI/schema changes; the binding contract is unchanged.

## Testing

- Skill tests green: `retired-tree-commands-drift`, `bundled-skill-list-sync`,
  `agent-briefing` (42 passed).
- Two independent reviews (factual accuracy + coherence); all findings fixed.


yuezengwu (approved):
Reviewed the seed-skill-only change.

What changed: State A now documents the real `tree init` owner behavior when the team has no GitHub App installation, and adds non-blocking post-create guidance for the App coverage cases that affect web snapshots and the Context Tree reviewer. The `/settings/github` route is the right install entry point because the web/server path mints the signed install URL and binds it back to the org; the PR also preserves the placement caveat for personal-account fallback repos.

I checked the added guidance against `apps/cli/src/commands/tree/init.ts`, the Settings GitHub install-url route, and the Context Tree/GitHub App tree contracts. No blockers found.

Validation: `pnpm validate:skill` and `python3 scripts/quick_validate_skill.py skills/first-tree-seed` passed locally. I could not run the client Vitest subset locally because this review worktree has no `node_modules` (`vitest: command not found`); GitHub's Validate Skill check is green.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1514

**Recommendation**: approve
- Rationale: The skill-only change makes the post-`tree init` GitHub App coverage gap visible without changing the binding contract or blocking the seed flow.

**Risk level**: C
- Path baseline: only `skills/first-tree-seed/**` changed -> C
- Semantic lift: none; no server, CLI, client, schema, or workflow behavior changes.

**PR Summary**
- Problem: An admin can build and bind a Context Tree that still remains invisible to the web Context view and Context Tree reviewer because the team's GitHub App cannot read the new tree repo.
- Approach: `first-tree-seed` now tells the seed agent to surface a prominent, non-blocking coverage message after `tree init`, routing no-install cases through the web console's GitHub settings page and selected/suspended install cases through GitHub's installation settings URL. It also clarifies the owner fallback and admin/auth failure wording.
- Impacted modules: `skills/first-tree-seed`

**Review findings**
✅ 1. The added guidance matches the active product/system boundary: web snapshots and the reviewer read the tree through the team's GitHub App installation, while `tree init` remains the agent-driven create + bind path. [R5 / `skills/first-tree-seed/SKILL.md:93`]
⚠️ 2. Non-blocking nuance: today's CLI output for an installed App is conditional (`If the First Tree GitHub App is installed on selected repositories...`) rather than a strict proof that the repo is excluded. The new skill text also says to make the coverage line loud, so this should still steer agents correctly; future edits should preserve that conditional reading. [R5 / `skills/first-tree-seed/SKILL.md:100`]

**Action taken**
- Approved.

PR #1524, opened 2026-07-08
feat(seed): home the tree in the source repo's account via `tree init --owner`

Gandy2025:
## What

Seed State A now creates the team Context Tree under the account that owns the team's **main bound source repo** (an org or a personal account alike), by passing `first-tree tree init --owner "<source-repo-owner>"`.

## Why

When the First Tree GitHub App is not yet installed — the common early case — `tree init` fell back to the caller's personal `gh` account. That orphans the team tree in an account a later org App install can never cover, so the web view and Context Tree reviewer stay blind to it even after the team installs the App on their org.

Homing the tree in the source repo's account fixes this: that is the account the App lives on (or will) to read the code, so a later App install there covers the tree. When the App is already installed on that account, `--owner` simply matches the installation account and nothing changes.

This is **skill-only** — `tree init --owner` already exists; no code change.

## Details

- **Owner selection** is defined for the multi-source case: single source → its account; several sharing one account → that account; sources spanning different accounts → the account owning most of them, and ask the admin if it is a genuine tie (rather than guessing).
- **Failure paths surface instead of silently falling back**, with distinct recovery per case:
  - No repo-create rights under the source account (no App installed): prefer granting rights so the tree stays beside its source; personal-account fallback only as a flagged-cost last resort.
  - `--owner` mismatches an *already-installed* App account (rare split setup): re-run without `--owner` to land on the installation account (the coverable home), not personal.
- A guardrail was added against deleting a real tree clone when clearing a half-scaffolded directory on retry.
- `VERSION` 0.2.3 → 0.2.4 so existing agents re-pull the skill.

## Testing

Skill-only markdown change (plus the VERSION bump); no TypeScript touched. Validated by two independent reviews — one checking every behavioral claim against `apps/cli/src/commands/tree/init.ts` (`resolveRepoOwner` / create / coverage), one checking prompt internal consistency, terminology, and edge-recovery safety — both findings folded in. Repo CI runs Validate Skill on the frontmatter/structure.


yuezengwu (changes requested):
Reviewing the skill-only change: the goal is to have seed state A call `tree init --owner <source-repo-owner>` so new Context Tree repos are created under the source repo's GitHub account, with explicit fallback and coverage guidance. The overall direction matches the current `tree init` implementation, but I found one blocker in the agent-facing instructions.

`skills/first-tree-seed/SKILL.md:62-67` says to derive `<source-repo-owner>` from the clone URL, but then says to take it from `manifest.sources` or `git -C <source-clone> remote get-url origin`. In the current workspace model, `workspace.json.sources` are immediate subdirectory names under `sourcesRoot`, not repo URLs or owners. For example, this workspace can record `sources: ["first-tree"]` while the actual remote owner is `agent-team-foundation`. If a seed agent follows the `manifest.sources` half literally, it can pass `--owner first-tree`, fail against the wrong GitHub account, and then give the admin the wrong recovery guidance.

Please make the source of truth unambiguous: resolve each `<source-clone>` path from the manifest, read `git -C <source-clone> remote get-url origin` (or the declared source repo URL from the generated briefing when available), and parse the owner from that URL. `manifest.sources` should only be used to locate the source clone, not as an owner source.

Validation: I ran `pnpm validate:skill` locally; CI's Validate Skill check is also green.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1524

- Rationale: The new owner-selection instruction still points agents at `manifest.sources`, which is not a repository URL or owner source in the current workspace schema.

**Risk level**: C
- Path baseline: only `skills/first-tree-seed/**` and its `VERSION` file changed -> C.
- Semantic lift: none; this is a skill-only prompt update, with no release, dependency, signing, or CI workflow change.

**PR Summary**
- Problem: During seed State A, a new team Context Tree can land under the caller's personal GitHub account when the First Tree GitHub App is not installed yet, leaving web snapshots and Context Tree reviewer coverage pointed at an account the later source-org install cannot cover.
- Approach: Teach `first-tree-seed` to pass `tree init --owner <source-repo-owner>`, spell out installed/no-install/mismatch recovery paths, add a retry guard around half-built scaffolds, and bump the skill version so installed agents repull it.
- Impacted modules: `skills/first-tree-seed`.

**Review findings**
❌ 1. `manifest.sources` is not a valid source for the GitHub owner. In the current workspace contract it stores local source clone names such as `first-tree`, with `sourcesRoot` locating the bare clone; the remote owner must be read from the source clone's `origin` URL, or from a declared source repo URL in the generated briefing when available. As written, an agent can pass `--owner first-tree` instead of `--owner agent-team-foundation`, which fails the create path or produces wrong recovery guidance. This matches the existing requested-change review; I am recording the same blocker as the requested reviewer. [R1 / `skills/first-tree-seed/SKILL.md:66`]

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks @yuezengwu @baixiaohang — you're both right, this was a real correctness bug. Fixed in 0e98473e.

**R1 (`manifest.sources` is not an owner source)** — addressed. The derivation no longer treats `manifest.sources` as a place to read the owner from. It now:

1. uses `manifest.sources` + `sourcesRoot` **only to locate** the source clone (`<workspaceRoot>/<sourcesRoot>/<source>`), then
2. parses `<source-repo-owner>` from `git -C <source-clone> remote get-url origin` (or the declared source repo URL in the generated briefing when present), taking the `<owner>` segment of that URL.

I also added an explicit guard in the text: *"`manifest.sources` holds directory names (e.g. `first-tree`), not owners — passing a source name as `--owner` targets the wrong GitHub account, so always parse the owner from the clone's remote."* — exactly the `first-tree` → `agent-team-foundation` trap you both named.

Updated at `skills/first-tree-seed/SKILL.md:62-72`. Please take another look.

yuezengwu (approved):
Re-reviewed after 0e98473e. The requested-change blocker is fixed: `manifest.sources` is now described only as the way to locate the source clone, and `<source-repo-owner>` is explicitly derived from the source clone's `origin` remote URL or a declared source repo URL from the briefing.

I also re-checked the surrounding failure/recovery wording against the current `tree init` behavior and the workspace schema. No remaining blocking findings from me.

Validation: `pnpm validate:skill` passes locally. I did not wait for the new CodeQL run because this PR is skill-only and review/approve does not need to wait on CI.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1524

**Recommendation**: approve
- Rationale: The previous blocker is fixed, and the skill guidance now derives the GitHub owner from the source clone remote URL rather than from `manifest.sources` directory names.

**Risk level**: C
- Path baseline: only `skills/first-tree-seed/**` and its `VERSION` file changed -> C.
- Semantic lift: none; this remains a skill-only prompt/version update, with no release, dependency, signing, or CI workflow change.

**PR Summary**
- Problem: During seed State A, a new team Context Tree could be created under the caller's personal GitHub account when the First Tree GitHub App is not installed yet, leaving web snapshots and the Context Tree reviewer unable to see it after a later source-org App install.
- Approach: Teach `first-tree-seed` to pass `tree init --owner <source-repo-owner>`, derive that owner from the source repo remote URL, spell out installed/no-install/mismatch recovery paths, guard retry cleanup around half-built scaffolds, and bump the skill version so agents repull it.
- Impacted modules: `skills/first-tree-seed`.

**Review findings**
✅ 1. The earlier `manifest.sources` owner-source blocker is fixed: the text now uses `manifest.sources`/`sourcesRoot` only to locate the clone and explicitly parses the GitHub owner from `git -C <source-clone> remote get-url origin` or a declared source repo URL from the briefing. [`skills/first-tree-seed/SKILL.md:62`]
✅ 2. The fallback wording matches current `tree init` behavior: installed App mismatches fail before local scaffolding, while no-install/no-create-rights failures can leave a non-empty local target that must be cleared carefully before retry. [`apps/cli/src/commands/tree/init.ts:217`]

**Action taken**
- Approved.

Issue #1490, opened 2026-07-08
Plan Windows daemon service support and symlink-safe runtime bootstrap


yuezengwu:
## Background

A Windows onboarding session exposed a set of runtime/platform issues after the user connected their computer:

1. Windows currently reports `Background service not supported on win32 - runs inline`.
2. Users can run the daemon long-lived in PowerShell with `first-tree daemon start`, but this is a foreground process, not a supervised service.
3. `first-tree daemon restart` correctly says service control is unsupported, but the recovery guidance is not actionable enough for onboarding users.
4. Once the user reached chat, Codex runtime failed before provider startup because workspace bootstrap tried to create a symlink and Windows denied it:

```text
EPERM: operation not permitted, symlink 'AGENTS.md' -> 'C:\Users\...\.first-tree\data\workspaces\my-codex\CLAUDE.md.<tmp>'
```

This means Windows can currently be used for short-lived inline/demo runtime, but it is not yet a reliable first-class long-running runtime experience.

### Inline daemon

`first-tree daemon start` can run in PowerShell and maintain the same core runtime capabilities as the service path:

- Cloud WebSocket connection
- inbox consumption
- agent/session runtime
- capability upload/probe
- provider execution

But process lifetime is tied to the terminal and user session:

- closing the terminal stops the daemon
- `Ctrl+C`, logout, reboot, or process crash stops it
- no service supervisor restarts it
- `daemon stop/restart` service control is unavailable
- logs are primarily terminal output

### Service daemon

On platforms with supported service managers, the service path adds operational guarantees rather than different agent semantics:

- background process independent of the terminal window
- restart/stop/status service control
- supervisor restart on crash or update handoff
- stable log location
- better long-running runtime posture

## Product problem

For onboarding/demo, inline PowerShell is acceptable if the UI/CLI tells the user clearly to keep the window open.

For a real Windows runtime, inline is not enough. A user expects the agent to remain available after closing the terminal or after a machine restart/login cycle. Today Windows lacks that service/supervisor layer, and the CLI command names (`daemon restart`) make users try service operations that cannot work on win32.

### Phase 1: Make Windows inline UX explicit

Improve unsupported-service command output, especially `daemon restart`:

```text
Service control is not supported on Windows.
First Tree runs inline on this platform.

If you are onboarding and need to refresh detected agents:
  first-tree daemon probe

If you need to restart the inline daemon:
  stop the PowerShell running `first-tree daemon start` with Ctrl+C,
[...]

PR #1540, opened 2026-07-09
Add Windows Task Scheduler daemon supervisor

baixiaohang:
## Summary
- Add a Windows `task-scheduler` supervisor backend to the existing daemon lifecycle API.
- Register a per-user Task Scheduler logon task that launches a stable wrapper and hidden `daemon supervise` loop.
- Let First Tree own child supervision: attached daemon child, exit `75` restart, non-zero backoff, stop intent, runtime-marker PID status, `taskkill` cleanup, and supervisor logging.
- Document Windows as per-user Task Scheduler support, not Windows Service / WinSW / pre-login daemon support.

## Notes
- Refs #1490.
- Task Scheduler is only the logon/start trigger; this does not depend on `RestartOnFailure`.
- Windows Service, WinSW, admin/service-account mode, pre-login daemon behavior, daemon runtime, install.sh, and npm install are intentionally out of scope.
- Context Tree updates are intentionally not included; architect will handle the tree follow-up.

## Validation
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/windows-supervisor.test.ts src/__tests__/service-install-core.test.ts src/__tests__/doctor-core.test.ts src/__tests__/daemon-start-command.test.ts src/__tests__/render-unit-channel.test.ts src/__tests__/daemon-refresh-unit.test.ts src/__tests__/cli-command-matrix-extra.test.ts src/__tests__/command-registration-smoke.test.ts`
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/daemon-refresh-unit.test.ts src/__tests__/cli-command-matrix-extra.test.ts`
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev typecheck`
- `env -u FIRST_TREE_CLIENT_ID pnpm check` (exits 0; reports existing warnings outside this PR's files)
- `env -u FIRST_TREE_CLIENT_ID pnpm typecheck`
- `env -u FIRST_TREE_CLIENT_ID -u FIRST_TREE_SERVER_URL -u FIRST_TREE_RUNTIME_SESSION_TOKEN -u FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE -u FIRST_TREE_SERVICE_MODE pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1`
- `env -u FIRST_TREE_CLIENT_ID -u FIRST_TREE_SERVER_URL -u FIRST_TREE_RUNTIME_SESSION_TOKEN -u FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE -u FIRST_TREE_SERVICE_MODE pnpm test -- --maxWorkers=2 --minWorkers=1 --testTimeout=20000`


yuezengwu (changes requested):
I read this PR as adding Windows per-user Task Scheduler support by registering a stable wrapper task, moving restart/backoff/exit-75 handling into a hidden `daemon supervise` loop, and using runtime markers plus `taskkill` to make `daemon status/stop/restart` work without relying on Task Scheduler's restart policy.

I found two blockers that need changes before this can land:

1. `apps/cli/src/core/supervisor/windows-supervisor.ts:214` directly calls `spawnProcess(program, args, ...)` for the resolved daemon child. On a normal Windows npm install, the channel bin is usually an npm-generated `.cmd` shim, and the new test even models this with `C:\First Tree\old\first-tree-dev.cmd`. Node cannot execute `.cmd`/`.bat` files through direct `spawn` without `shell: true` or an explicit `cmd.exe /d /s /c ...` wrapper, so the Task Scheduler wrapper can start `daemon supervise`, but the supervisor then fails to launch `daemon start --no-interactive` and retries forever. Please route `.cmd`/`.bat` bin invocations through `cmd.exe` (or otherwise resolve to a real executable) and cover the same Windows service-mode path used by `daemon refresh-unit`.

2. `apps/cli/src/core/supervisor/task-scheduler.ts:278` uses runtime marker PIDs as authority for `taskkill /PID ... /T` and `/F`, while `listLiveServiceRuntimeMarkers()` only proves `process.kill(pid, 0)` succeeds. A stale marker left by a hard crash or power loss can survive until Windows reuses that PID for an unrelated process; `daemon stop` would then kill that unrelated process tree. The existing marker mechanism was safe as a switch/logout gate, but this PR turns it into a destructive process selector. Please validate process identity before killing (for example creation time plus expected command/process lineage), or fail closed/ignore untrusted markers instead of taskkilling them.

baixiaohang:
Addressed the requested changes in `2e8f6667`.

Fixes included:
- Route Windows `.cmd` / `.bat` child invocations through `cmd.exe /d /s /c ...`, so npm-generated shims can launch `daemon start --no-interactive` reliably from the hidden supervisor loop.
- Add Windows process identity validation before `taskkill`: the runtime marker PID must still resolve to a First Tree `daemon start --no-interactive` process, and the process creation time must not indicate PID reuse.
- Tighten the PID reuse tolerance to 5s and cover a `marker.createdAt = T`, `CreationTimeUtc = T + 30s` rejection path.
- Make Windows stop a two-phase operation: prevalidate all live service markers before writing stop intent or killing any PID. If any marker is untrusted, stop fails closed without writing `windows-supervisor-stop.json` and without `taskkill`.
- Keep a kill-time recheck to cover races between prevalidation and `taskkill`.

Local verification:
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/windows-supervisor.test.ts src/__tests__/service-install-core.test.ts`
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev typecheck`
- `env -u FIRST_TREE_CLIENT_ID pnpm check` (exit 0; existing warnings outside this PR's files remain)
- Clean-env full CLI test: `env -u FIRST_TREE_CLIENT_ID -u FIRST_TREE_SERVER_URL -u FIRST_TREE_RUNTIME_SESSION_TOKEN -u FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE -u FIRST_TREE_SERVICE_MODE -u FIRST_TREE_AGENT_ID -u FIRST_TREE_AGENT_SLUG -u FIRST_TREE_CHAT_ID -u FIRST_TREE_INBOX_ID -u FIRST_TREE_SWITCH_DRAIN_VERSION pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1` (89 files / 970 tests passed)

@yuezengwu please re-review when you have a chance.

yuezengwu (changes requested):
Thanks for addressing the two earlier blockers. The PID-marker hardening is now fail-closed before `taskkill`, and the additional kill-time recheck covers the race I was worried about.

One blocker remains in the `.cmd` launch fix:

`apps/cli/src/core/supervisor/windows-supervisor.ts:138` builds the `cmd.exe` target as:

```ts
["/d", "/s", "/c", [program, ...args].map(cmdQuote).join(" ")]
```

For a normal path-with-spaces case this produces a `<string>` like:

```cmd
"C:\First Tree\old\first-tree-dev.cmd" "daemon" "start" "--no-interactive"
```

Because `/s` is present, `cmd` strips the first and last quotes around the `/c` string, leaving the executable path unprotected while the last argument loses its closing quote. Microsoft's `cmd` docs describe this exact `/s` behavior: it strips the first and last quotes around the `<string>` passed to `/c`/`/k` while leaving the rest unchanged. See: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd

Please wrap the whole command string in an additional outer quote pair before passing it to `/c`, e.g. the effective string should be shaped like:

```cmd
""C:\First Tree\old\first-tree-dev.cmd" "daemon" "start" "--no-interactive""
```

and add a test that asserts the complete `/c` string has those outer quotes for a `.cmd` path containing spaces. Without that, the npm shim path remains unreliable for the exact Windows install shape this PR is adding.

baixiaohang:
Addressed the remaining `.cmd` launch blocker in `83b62620`.

Change:
- The hidden Windows supervisor still routes `.cmd` / `.bat` invocations through `cmd.exe /d /s /c`, but the `/c` command string is now wrapped in an additional outer quote pair.
- For a path-with-spaces npm shim, the fourth arg is now shaped as:
  ```cmd
  ""C:\First Tree\old\first-tree-dev.cmd" "daemon" "start" "--no-interactive""
  ```
  so `cmd /s /c` strips only the outer pair and leaves the executable path and arguments quoted.
- The supervisor test now asserts the complete `/c` string for old/new `.cmd` paths with spaces.

Verification:
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/windows-supervisor.test.ts src/__tests__/service-install-core.test.ts`
- `env -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev typecheck`
- `env -u FIRST_TREE_CLIENT_ID pnpm check` (exit 0; existing warnings outside this PR's files remain)
- Clean-env full CLI test: `env -u FIRST_TREE_CLIENT_ID -u FIRST_TREE_SERVER_URL -u FIRST_TREE_RUNTIME_SESSION_TOKEN -u FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE -u FIRST_TREE_SERVICE_MODE -u FIRST_TREE_AGENT_ID -u FIRST_TREE_AGENT_SLUG -u FIRST_TREE_CHAT_ID -u FIRST_TREE_INBOX_ID -u FIRST_TREE_SWITCH_DRAIN_VERSION pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1` (89 files / 970 tests passed)

@yuezengwu please re-review when you have a chance.

yuezengwu (approved):
Re-reviewed the latest head `83b62620`.

The remaining Windows `.cmd` launch blocker is fixed: `.cmd` / `.bat` child invocations still go through `cmd.exe /d /s /c`, and the `/c` command string now has the extra outer quote pair needed for paths with spaces. The updated test asserts the full string shape for both old and new npm shim paths.

The earlier PID-marker / `taskkill` blocker also remains addressed: stop prevalidates all live service markers before writing stop intent or killing, untrusted markers fail closed, and the kill-time recheck covers the race window.

I did not rerun the test suite locally in this detached review worktree because it does not have `node_modules`/`vitest`; reviewed the updated diff and `git diff --check` is clean.

PR #1544, opened 2026-07-09
feat(web): streamline connect-computer dialog and add live expiry ticker

liuchao-001:
## What

Reworks the Settings → **Connect computer** dialog per liuchao-001's feedback:

- **Description copy**: now reads "Run this command in your terminal to pair with First Tree. It will install or upgrade to the latest First Tree client." (was the conditional "if the CLI isn't installed yet…" phrasing).
- **`# expires in 10m` comment removed from the command block** — users selecting the command text manually copied the comment along with it.
- **Live expiry ticker**: a subtle `expires in Xm:XXs` countdown now sits to the right of the Copy button while waiting. It is keyed on the command string so "Generate new token" restarts the count from the fresh TTL. Purely informational — the dialog's existing expiry timer still owns the flip to the error phase.
- **"Single-use · regenerates the previous one." caption removed** (the `caption` prop is deleted; the dialog was its only consumer).
- **"Taking a while? A few common reasons:" panel removed** from the dialog. `ConnectStuckPanel` had no other production consumer, so the component is deleted; `STUCK_AFTER_MS` moves into onboarding `StepConnectComputer` (its only remaining user — onboarding keeps its own one-line Node.js stuck hint, unchanged), and the orphaned `stuckTitle`/`stuckReasons` strings are dropped from onboarding COPY.

## Notes

- `ConnectCommandPanel` is only used by this dialog (+ SSR smoke test), so the panel-level changes don't leak into onboarding.
- SSR smoke test now covers the waiting-phase ticker (`expires in 10m:00s` seeded from the TTL).

## Testing

- `pnpm check` + `pnpm typecheck` pass (the 3 remaining biome diagnostics are pre-existing on main in untouched files).
- `@first-tree/web` tests: 137 files / 1146 tests pass.
- Full `pnpm test`: `first-tree-dev`'s `tree-init-command-extra.test.ts` ("creates and summarizes an unbound Context Tree with local gh/git only") fails identically on a clean `origin/main` worktree on this machine — pre-existing, unrelated to this web-only diff.


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1544

**Recommendation**: approve
- Rationale: The Settings connect-computer dialog becomes clearer without changing the server-authored command boundary, the host-owned expiry/error transition, or onboarding's separate first-run flow.

**Risk level**: C
- Path baseline: only `packages/web/**` files changed.
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, data-layer contracts, server services, or client/CLI code.

**PR Summary**
- Problem: Operators using Settings -> Connect computer could manually copy the inline expiry comment with the command, while the steady-state pairing dialog carried extra explanatory UI that made the main action noisier.
- Approach: Remove the command-block expiry comment, caption, and dialog-only stuck panel; add an informational countdown beside Copy keyed by the command string; keep the existing dialog timer as the owner of the expired-token error phase; leave onboarding's one-line Node.js stuck hint in the onboarding step.
- Impacted modules: `packages/web` connect-command panel, Settings computer connection dialog, onboarding copy, and SSR smoke coverage.

**Review findings**
✅ No blocking findings. The Context Tree boundaries for `/settings/computers` and onboarding are preserved: the command remains server-authored, the dialog keeps its loading/waiting/success/error state ownership, and onboarding remains separate from the steady-state connect/reconnect surface.
✅ The countdown is presentational and remounts per command, so regenerating a token restarts the visible TTL without becoming the source of truth for expiry behavior.
✅ Latest GitHub CI checks are green.

**Action taken**
- Approved.

yuezengwu (approved):
Goal and core change: this PR streamlines the Settings -> Connect computer dialog by keeping the command block copy-only, moving expiry into a presentational live ticker beside Copy, and removing the dialog-only stuck-help/caption copy while keeping onboarding's separate first-run Node.js hint.

No blocking findings. I checked the runtime/computer and onboarding Context Tree boundaries: the command remains server-authored, the dialog still owns loading/waiting/success/error transitions, and the new ticker does not become the source of truth for token expiry. No DB/schema or core data-structure changes are involved.

Verification performed:
- `pnpm --filter @first-tree/web test -- src/pages/__tests__/page-ssr-smoke.test.tsx src/pages/onboarding/__tests__/copy.test.ts src/pages/clients/__tests__/new-connection-dialog.test.ts` (the command ran the web test suite): 137 files / 1146 tests passed.
- `pnpm --filter @first-tree/web typecheck`: passed, including design-token lint.
- `pnpm exec biome check` on the changed web files: clean.
- Latest GitHub checks are passing.

PR #1548, opened 2026-07-09
Fix Windows Task Scheduler XML encoding

baixiaohang:
## Summary

- Write Windows Task Scheduler XML as UTF-16LE with a BOM and declare `encoding="UTF-16"` so `schtasks /Create /XML` can import the generated task reliably on Windows.
- Read the task XML back as UTF-16LE for drift detection, stripping the BOM before comparing with the rendered definition.
- Avoid surfacing mojibake from localized Windows native command stderr when `schtasks.exe` / `taskkill.exe` output cannot be decoded as UTF-8.

## Context

Follow-up to the Windows Task Scheduler supervisor backend. A Windows PowerShell run of `first-tree-staging login <code>` authenticated and registered the computer, then failed during Task Scheduler import with an XML parse error at the XML declaration encoding field.

## Validation

- `pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/service-install-core.test.ts src/__tests__/render-unit-channel.test.ts`
- `git diff --check`
- `pnpm check` (exits 0; reports existing warnings/info outside this PR's files)
- Clean-env full `first-tree-dev` test before the final main sync/stderr fallback: 89 files / 972 tests passed

### DIFF
--- a/apps/cli/src/__tests__/render-unit-channel.test.ts
+++ b/apps/cli/src/__tests__/render-unit-channel.test.ts
@@ -163,6 +163,10 @@ describe("renderWindowsTaskXml — channel identity baked into Task Scheduler te
   const wrapperPath = join(defaultHome(), "service", `${channelConfig.launchdLabel}-supervisor.cmd`);
   const taskXml = renderWindowsTaskXml(wrapperPath, "ACME\\developer");
 
+  it("declares the UTF-16 encoding used for the imported task XML file", () => {
+    expect(taskXml).toMatch(/^<\?xml version="1\.0" encoding="UTF-16"\?>/u);
+  });
+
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
Reviewed the implementation and tests. The change matches the Windows Task Scheduler import contract: the rendered declaration now matches the UTF-16LE+BOM bytes written to disk, drift detection reads the task XML back with the same encoding and strips the BOM before comparison, and localized native stderr with replacement characters is replaced by an actionable fallback instead of surfacing mojibake.

No blocking findings.

Verified locally:
- `pnpm --filter first-tree-dev test --maxWorkers=2 --minWorkers=1 src/__tests__/service-install-core.test.ts src/__tests__/render-unit-channel.test.ts`

PR #1555, opened 2026-07-09
[jeremy] Fix inviteable mention candidates

L42y:
## Summary

- Filter inviteable non-participant mention candidates to runtime agents only.
- Keep in-chat human mentions unchanged, but prevent human org rows from appearing as inviteable “agents”.
- Cap the inviteable non-participant autocomplete section at 3 visible candidates while preserving the full resolver set for selected mentions.
- Add regressions for human-row filtering and the 3-candidate inviteable cap.

## Root cause

The auto-invite PR allowed all addressable org rows through the inviteable candidate projection. A `type: "human"` row could therefore be selected and invited from the “unjoined agents” section, but runtime activation only happens for notify-worthy agent recipients. Inviting a human participant does not wake an agent runtime.

The same inviteable projection could also be too large for the `@` picker. The display projection now caps inviteable suggestions to 3, while the resolver projection remains complete so already-selected inviteable `@name` tokens still invite and route correctly at send time.

## Validation

- `pnpm --filter @first-tree/web exec vitest run src/pages/workspace/center/__tests__/chat-view-dom.test.tsx src/components/__tests__/mention-autocomplete.test.ts`
- `pnpm exec biome check packages/web/src/pages/workspace/center/chat-view.tsx packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx packages/web/src/components/chat/ask-takeover.tsx packages/web/src/components/mention-autocomplete.tsx packages/web/src/components/__tests__/mention-autocomplete.test.ts packages/web/src/lib/use-org-agents.ts`

### DIFF
--- a/packages/web/src/components/chat/ask-takeover.tsx
+++ b/packages/web/src/components/chat/ask-takeover.tsx
@@ -100,6 +100,7 @@ export function AskTakeover({
   askerName,
   sending = false,
   mentionCandidates = [],
+  mentionAutocompleteCandidates = mentionCandidates,
   error,
   onMentionQueryChange,
   onMentionTextChange,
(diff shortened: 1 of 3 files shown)


L42y:
Superseded by #1556, which uses the repository-allowed `fix/` branch prefix.

PR #1563, opened 2026-07-09
fix: scan-fix conversion fans out via a launcher, not a same-name child

Gandy2025:
## What

Reworks how the `first-tree-welcome` skill handles the production-scan **fix conversion** ("fix the launch blockers found by my production readiness scan").

## Why

The fix conversion could show **two chats both titled "Fix production scan blockers."** Two independent mechanisms produce that symptom:

1. **Agent-spawned same-name child — fixed here.** On the onboarding (greeted) path the skill treated the fix chat as a launcher and spawned a *second* chat with the *same* title the product already gives the launcher (`Fix production scan blockers`). One entry → two identical chats.
2. **Cross-path idempotency — follow-up PR.** Re-entering `/quickstart?action=fix` after onboarding hits a different endpoint (`POST /orgs/:org/chats` task mode) with no shared idempotency key, creating a second launcher. That is a server change and ships separately, with a regression test.

This PR fixes **mechanism 1**.

## Change

- `skills/first-tree-welcome/SKILL.md` — the scan-fix handoff is now always the fix launcher and routes by the count of `confirmed`, safe, high-leverage blockers:
  - **≥ 2** → fan out one **distinctly-named** child chat per blocker (top ~4, bounded), with an explicit ban on reusing the launcher's generic title.
  - **exactly 1** → fix in place (no launcher + child overhead).
  - **0 safe** → surface for the user; spawn nothing.
  - Judgment calls (architectural / auth-design / low-`confidence`) are never fanned out or autofixed — surfaced for the user to decide.
  - The onboarding greeting is now purely a **role signal** (admin ⇒ eligible for the post-value Context Tree offer); it no longer gates the fan-out, so the already-onboarded (direct) path also becomes a fan-out launcher.
- `packages/web/src/pages/workspace/center/onboarding/bootstrap-prose.ts` — corrects the now-stale `buildScanFixBootstrap` comment (the two openings now differ only in the greeting/role signal).
- Skill `VERSION` 1.1.2 → 1.2.0.

## Verification

- `first-tree-welcome` floor + grader invariants: **38/38** (these enforce SKILL.md staying in sync with the OpenAI routing metadata, the agent-briefing skill-map, and the onboarding bootstraps — all green, so no routing-copy change was needed).
- web `biome` + `typecheck` + `lint:tokens` clean.
- 2 independent reviews (internal-consistency + adversarial both-paths); findings folded in.

## Follow-up

A separate PR adds a shared idempotency key to the fix conversion so re-entering the fix link cannot create a second launcher (mechanism 2), with a regression test.


yuezengwu (changes requested):
This PR's goal is to make the production-scan fix conversion use the existing `Fix production scan blockers` chat as the launcher, avoid spawning a same-title child, and route by safe blocker count. The core changes are in the `first-tree-welcome` skill prose, the scan-fix bootstrap comment, and the skill version bump. I do not see DB/schema or core data-structure changes.

- `skills/first-tree-welcome/SKILL.md:324` still contradicts the new one-blocker path. The new scan-fix section says `Exactly one safe blocker -> do NOT fan out; fix it here`, but the generic Spawning Task Chats section now says that for a production-scan fix launcher, after triage the agent should not do the work in this launcher chat and should "open one chat per confirmed safe blocker". With exactly one confirmed safe blocker, the later/harder rule instructs the agent to spawn a single child anyway, reintroducing the launcher + child overhead this PR is explicitly trying to avoid. Please narrow the spawning section/hard rule to the `two or more` scan-fix case, and keep the single-blocker exception explicit there too.

Verification notes:

- GitHub CI is green as of head `c7b4bba20a8db758a400228e99901dd1396917a9`.
- Local targeted `pnpm` test runs were blocked because this fresh review worktree has no `node_modules` and `vitest` is not installed.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1563

- Rationale: The main direction is right, but one updated instruction still contradicts the new single-blocker behavior.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**` prose and `skills/first-tree-welcome/**`.
- Semantic lift: none; no server/client/CLI or web subscription, WebSocket, or data-layer change.

**PR Summary**
- Problem: The production-scan fix conversion can leave the user with confusing duplicate `Fix production scan blockers` chats when the welcome skill spawns a same-name child from the launcher.
- Approach: Make the fix chat act as a launcher only when there are multiple confirmed, safe blockers to split into specific child chats; handle one safe blocker in place; surface judgment calls; keep the onboarding greeting as an admin-role signal rather than a fan-out gate. The web bootstrap comment is updated to match that behavior.
- Impacted modules: `skills/first-tree-welcome`, onboarding bootstrap prose.

**Review findings**
❌ 1. The generic `Spawning Task Chats` section still contradicts the new single-blocker route. It says that for a production-scan fix launcher, once confirmed safe blockers are triaged, the agent must not do the work in the launcher and should open one chat per confirmed safe blocker. That instruction also applies to the `exactly one safe blocker` case and can still push a lone blocker into a child chat, despite the new handoff branch saying to fix it in place. Please scope this paragraph to the `two or more` branch, or explicitly say this section applies only after the handoff has decided to fan out. [R1 / `skills/first-tree-welcome/SKILL.md:324`]

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks both — good catch, fixed in d45c53de.

**R1 (single-blocker contradiction):** you're right. The scan-fix clause I added to `Spawning Task Chats` (and the fan-out Hard Rule) applied to every safe-blocker count, so a lone blocker would still be pushed into a child chat, contradicting the new `exactly one safe blocker → fix it here` branch.

Both are now scoped to the **two-or-more** case, with the single-blocker in-place exception called out explicitly in both spots:
- `Spawning Task Chats` — opener now says "one exception: a scan-fix launcher with **exactly one** safe blocker fixes it in place", and the enumeration reads "for a scan-fix launcher with **two or more** safe blockers, open one chat per blocker".
- Hard Rules fan-out bullet — adds "(Exception: a production-scan fix launcher with exactly one safe blocker fixes it in place …)".

`first-tree-welcome` floor + grader invariants still 38/38 locally. Re-requesting review.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1563

**Recommendation**: approve
- Rationale: The previous single-blocker contradiction is fixed in the latest head, and I did not find another blocking issue in the updated skill routing.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**` prose and `skills/first-tree-welcome/**`.
- Semantic lift: none; no server/client/CLI or web subscription, WebSocket, or data-layer change.

**PR Summary**
- Problem: The production-scan fix conversion could confuse users by creating two `Fix production scan blockers` chats from one entry.
- Approach: Treat the scan-fix chat as the fix launcher only when there are multiple confirmed safe blockers to split into distinct child chats; fix a lone safe blocker in place; surface judgment calls; keep the onboarding greeting as an admin-role signal for later setup offers.
- Impacted modules: `skills/first-tree-welcome`, onboarding bootstrap prose.

**Review findings**
✅ 1. The prior R1 issue is resolved: `Spawning Task Chats` and the hard-rule summary now scope scan-fix child-chat creation to the two-or-more safe-blocker case, and both preserve the exactly-one-safe-blocker in-place exception.
✅ 2. The updated routing remains consistent with the onboarding/tree contract: value lands first, tree setup remains gated and separate, and spawned task briefs stay self-contained.

**Action taken**
- Approved.
- Did not run tests per PR review mode. Remote GitHub checks were still pending at review time.

yuezengwu (approved):
Re-reviewed head `d45c53def0bcf293929aecb1d2d0cb2581d9610e`.

The prior requested change is addressed: the scan-fix spawning section is now explicitly limited to the two-or-more safe-blocker case, and the hard-rule fan-out bullet carries the exactly-one-safe-blocker in-place exception. I also rechecked the surrounding scan-fix routing text and did not find another instruction that would push a lone safe blocker into a child chat.

No new blockers found. Approving.

Verification:
- `python3 scripts/quick_validate_skill.py skills/first-tree-welcome`
- GitHub CI was still in progress during review; local Vitest-based skill/web checks were not rerun because this fresh review worktree does not have `node_modules`.

PR #1568, opened 2026-07-09
fix(cli): prefer Windows cmd shim for supervisor

baixiaohang:
## Summary
- Normalize Windows CLI service launches to prefer the `.cmd` npm shim when `where.exe` returns an extensionless shell shim first.
- Keep non-Windows behavior unchanged and preserve existing absolute binary resolution.
- Add a regression test for the Task Scheduler supervisor resolution path.

## Verification
- `corepack pnpm --filter first-tree-dev exec vitest run src/__tests__/service-install-core.test.ts -t "uses the Windows cmd shim"`
- `corepack pnpm --filter first-tree-dev exec vitest run src/__tests__/windows-supervisor.test.ts`
- `corepack pnpm --filter first-tree-dev typecheck`

## Manual Windows validation
- Built `first-tree-dev` from this branch.
- Installed the dev Task Scheduler service with an npm-like shim layout where `where.exe first-tree-dev` returns the extensionless shim before `.cmd`.
- Confirmed the generated wrapper invokes `first-tree-dev.cmd daemon supervise`.
- Confirmed `first-tree-dev daemon start` reports the task-scheduler service running and writes a runtime marker under `.first-tree-dev\state\client-runtimes`.

### DIFF
--- a/apps/cli/src/__tests__/service-install-core.test.ts
+++ b/apps/cli/src/__tests__/service-install-core.test.ts
@@ -166,6 +166,20 @@ describe("service install helpers", () => {
     expect(() => resolveCliInvocation()).toThrow("Cannot resolve CLI entry point");
   });
 
+  it("uses the Windows cmd shim when where returns npm's extensionless shell shim first", () => {
+    setPlatform("win32");
+    const npmBinDir = join(home, "npm");
+    const extensionlessShim = join(npmBinDir, channelConfig.binName);
+    const cmdShim = `${extensionlessShim}.cmd`;
+    mkdirSync(npmBinDir, { recursive: true });
+    writeFileSync(extensionlessShim, "#!/bin/sh\nexec node cli.mjs \"$@\"\n");
+    writeFileSync(cmdShim, "@ECHO off\r\nnode cli.mjs %*\r\n");
+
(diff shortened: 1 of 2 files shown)


yuezengwu (changes requested):
Reviewed the Windows supervisor shim fix.

Goal/core change: `resolveCliInvocation()` still resolves the channel binary through PATH and preserves the non-Windows / fallback behavior, but on Windows it now maps an extensionless npm shell shim to the sibling `.cmd` shim when present. That matches the Task Scheduler launch path the PR is trying to repair.

Requesting changes for one blocker: CI `Lint & Type Check` is failing on formatting in the new regression test at `apps/cli/src/__tests__/service-install-core.test.ts:175`. Biome wants this line reformatted, e.g. use the single-quoted string form it reports:

```ts
writeFileSync(extensionlessShim, '#!/bin/sh\nexec node cli.mjs "$@"\n');
```

I did not find a behavioral issue in the shim-resolution logic itself. No DB/schema or core data-structure change is involved.

yuezengwu (approved):
Re-reviewed after `fix(cli): format Windows shim test`.

The previous Biome formatting blocker is fixed, and the updated `Lint & Type Check` job is green. The implementation remains narrowly scoped to Windows CLI shim normalization for supervisor/service launch and preserves the non-Windows and fallback paths.

Approved. No DB/schema or core data-structure change involved.

PR #1573, opened 2026-07-09
fix(cli): preserve Windows supervisor cmd quoting

baixiaohang:
## Summary
- Preserve Windows command-line quoting when the Task Scheduler supervisor spawns the daemon child through `cmd.exe`.
- Pass `windowsVerbatimArguments: true` for the `cmd.exe /c` supervisor child launch path so Node does not rewrite quotes into `\"` sequences.
- Tighten Windows supervisor tests to assert the verbatim argument mode is used.

## Root cause
`first-tree-staging` 0.5.13-staging.790.1 includes #1568, so the generated Task Scheduler wrapper now launches `first-tree-staging.cmd daemon supervise`. The remaining failure is one layer deeper: `daemon supervise` resolves the daemon child to `first-tree-staging.cmd daemon start --no-interactive`, but spawns it through `cmd.exe` without `windowsVerbatimArguments`. Node rewrites the quoted command string, and `cmd.exe` sees the whole `\"\"...\"\"` string as a command name, so the child exits before writing a runtime marker.

## Verification
- `corepack pnpm --filter first-tree-dev exec vitest run src/__tests__/windows-supervisor.test.ts`
- `corepack pnpm --filter first-tree-dev typecheck`
- `corepack pnpm --filter @first-tree/shared build`
- `corepack pnpm --filter @first-tree/client build`
- `corepack pnpm --filter first-tree-dev build`

## Manual Windows validation
- Reproduced the staging failure on 0.5.13-staging.790.1: Task Scheduler runs `wscript -> .cmd -> node ... daemon supervise`, then supervisor logs repeated child failures for `first-tree-staging.cmd daemon start --no-interactive` with the `\"\"...\"\"` command string.
- Verified with a Node `spawnSync(cmd.exe, ['/d','/s','/c', command])` smoke test that the current non-verbatim spawn fails while the verbatim spawn succeeds.
- Built `first-tree-dev` from this branch and installed the dev Task Scheduler service with an npm-like shim layout where `where.exe first-tree-dev` returns the extensionless shim before `.cmd`.
- Confirmed `first-tree-dev status` reports `Service: ? running (task-scheduler, pid 23812)` and a runtime marker is written under `.first-tree-dev\state\client-runtimes`.

### DIFF
--- a/apps/cli/src/__tests__/windows-supervisor.test.ts
+++ b/apps/cli/src/__tests__/windows-supervisor.test.ts
@@ -72,6 +72,7 @@ describe("Windows supervisor loop", () => {
       ["/d", "/s", "/c", '""first-tree-dev" "daemon" "start" "--no-interactive""'],
       expect.objectContaining({
         detached: false,
+        windowsVerbatimArguments: true,
         env: expect.objectContaining({
           FIRST_TREE_HOME: home,
           FIRST_TREE_SERVICE_MODE: "1",
@@ -100,7 +101,7 @@ describe("Windows supervisor loop", () => {
         "/c",
         '""C:\\Users\\baixi\\AppData\\Roaming\\npm\\first-tree-staging" "daemon" "start" "--no-interactive""',
       ],
-      expect.objectContaining({ detached: false, windowsHide: true }),
+      expect.objectContaining({ detached: false, windowsHide: true, windowsVerbatimArguments: true }),
     );
   });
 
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the Windows supervisor spawn path and tests. The change is narrowly scoped to the cmd.exe child-launch path, preserves direct node.exe invocation behavior, and adds coverage that the verbatim argument mode is used for the Task Scheduler supervisor child. Local verification passed: targeted windows-supervisor vitest and first-tree-dev typecheck.

PR #1575, opened 2026-07-09
fix: align scan-fix fanout with scan blockers

Gandy2025:
## Summary

Align the `first-tree-welcome` scan-fix launcher rules with the production-scan contract that normally reports 3-5 blockers.

- change scan-fix fan-out from `top ~4` confirmed blockers to eligible blockers, with up to 5 active child chats
- keep the no-overhead exceptions: one eligible blocker is fixed in the launcher, zero eligible blockers are surfaced to the user
- add explicit guards against splitting one blocker into implementation-step chats and against fixing stale/already-covered findings
- remove reliance on production-scan confidence fields that ps-1 reports do not emit
- bump `first-tree-welcome` to 1.2.1 and add floor invariants for the new contract

## Why

The production-scan side is designed around 3-5 blockers, so the fix conversion should usually expose that same amount of parallel value when those blockers are independently fixable. The previous `top ~4`/`confirmed` wording under-used that contract and referenced a confidence field the current scan report shape does not provide.

## Verification

- `pnpm --filter @first-tree/skill-evals exec vitest run src/suites/first-tree-welcome/__tests__/floor.test.ts src/suites/first-tree-welcome/__tests__/grader.test.ts` (40 tests)
- `python3 scripts/quick_validate_skill.py skills/first-tree-welcome`
- `pnpm exec biome check packages/skill-evals/src/suites/first-tree-welcome/__tests__/floor.test.ts`
- `git diff --check`
- `rg -n 'top ~4|highest-leverage AND `confirmed`|low-`confidence` findings|triaged the confirmed safe blockers|safe blocker|safe blockers' skills/first-tree-welcome/SKILL.md` (no matches)

Note: full `pnpm --filter @first-tree/skill-evals test` was also run and still has unrelated local fixture failures in `first-tree-write` post-model verify and a `first-tree-seed` periodic timeout; the touched welcome suite passed in that run.


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1575

**Recommendation**: approve
- Rationale: The change aligns the scan-fix launcher with the 3-5 blocker production-scan contract while keeping the existing no-overhead and no-judgment-call guardrails intact.

**Risk level**: C
- Path baseline: changes are limited to `skills/first-tree-welcome/**` and `packages/skill-evals/src/suites/first-tree-welcome/**`, which do not match A/B paths.
- Semantic lift: none; no CLI/client/server/web data-layer, release, dependency, signing, or CI workflow behavior is changed.

**PR Summary**
- Problem: When a user asks First Tree to fix production-scan launch blockers, the launcher under-used reports that normally contain 3-5 blockers and referenced confidence-style fields the current report shape does not provide.
- Approach: The welcome skill now fans out every eligible, independently fixable blocker up to five active child chats, keeps the zero/one-blocker exceptions, prevents splitting one blocker into implementation-step chats, and requires stale/already-covered findings to be surfaced rather than duplicated.
- Impacted modules: `first-tree-welcome` skill payload and its floor invariants.

**Review findings**
✅ No blocking findings. The new eligibility definition is evidence/current-state based, the fan-out cap matches the expected blocker range, and the added floor invariants guard the retired `top ~4` / `confirmed` / confidence-field wording.

**Action taken**
- Approved.

yuezengwu (approved):
Reviewed the scan-fix launcher update. The change aligns the welcome skill with the production-scan blocker shape by moving from confirmed/top~4 wording to eligible blockers, up to five active child chats, stale/already-covered guards, and floor invariants for the new contract.

No blockers found.

Verification I checked:
- `python3 scripts/quick_validate_skill.py skills/first-tree-welcome`
- GitHub checks are green, including Test, Lint & Type Check, and Validate Skill.

Local note: I did not rerun the Vitest target in this detached review worktree because JS dependencies are not installed here (`vitest` missing); the corresponding GitHub Test check passed.

PR #1577, opened 2026-07-09
fix: soften runtime session rollout enforcement

yuezengwu:
## Summary

- Keeps the original session-only runtime-session-token model: every successful `agent:bind` mints a fresh token and `agent:bound` returns it.
- Removes the earlier owner-client token reuse wire shape from this PR's final diff (`currentRuntimeSessionToken`, reused bind results, and omitted bound tokens are not part of the final change).
- Softens `agent-selector` while `FIRST_TREE_AGENT_HTTP_RUNTIME_SESSION_ENFORCEMENT=false`: stale/invalid runtime-session tokens degrade to the same legacy accepted path as missing tokens and log a warning.
- Keeps hard enforcement unchanged: missing tokens and invalid/stale tokens still return distinct 403 failures when enforcement is enabled.

## Scope Note

This PR originally explored owner-client token reuse. The later commit `d24fcaec` intentionally cancels that path without force-pushing, per review direction. The final PR diff is only the soft-degrade server behavior plus tests.

## Root Cause

During the rollout window, stale-token requests were treated more strictly than missing-token requests: missing tokens could pass under soft enforcement, but stale tokens still hard-failed with `403 Invalid agent runtime session`. With per-bind token rotation retained, in-flight requests can encounter a just-rotated token. Soft enforcement should absorb that rollout race instead of failing harder than the token-less legacy path.

## Compatibility

- New server + old client: still mints on every bind and returns `runtimeSessionToken`.
- New server + new client PR: the client reads the latest token file for agent-scoped HTTP, but bind protocol stays unchanged.
- No new plaintext secret is stored server-side; the server still stores only token hashes.

## Validation

- `pnpm --filter @first-tree/server test -- src/__tests__/agent-bind-authorization.test.ts src/__tests__/ws-client-edge.test.ts` was attempted and expanded to the full server suite; result: 215 test files / 2231 tests passed.
- `pnpm check` passed with existing unrelated Biome warnings/info.
- `pnpm typecheck` passed on the final stacked branch.

## QA

Formal QA should run against the stacked client PR (#1578), because the end-to-end bug fix depends on client/CLI token delivery plus this soft-degrade rollout behavior.


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1577

**Recommendation**: approve
- Rationale: This safely aligns stale or invalid token behavior with the existing token-less rollout path only while enforcement is disabled, without changing hard enforcement or the bind protocol.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low
- Semantic lift: `agent-selector` changes core agent runtime-session authorization behavior -> B-high

**PR summary**
- Problem: During rollout, per-bind token rotation can make an in-flight agent request present a just-stale token. Soft mode already accepts a missing token, but previously rejected that stale token with 403, making upgraded clients fail more strictly than token-less legacy callers.
- Approach: When enforcement is disabled, failed runtime-session validation now follows the same accepted legacy path as a missing token and emits a structured warning. When enforcement is enabled, missing and invalid tokens remain distinct 403 failures. Every successful bind still mints and returns a fresh token.
- Impacted modules: `packages/server/src/middleware/agent-selector.ts` and its authorization tests.

**Review findings**
✅ 1. The final diff contains no owner-client token reuse wire shape; per-bind token minting remains the single session model.
✅ 2. The relaxation is scoped behind disabled enforcement, while the user, membership, client-ownership, and scoped-outbox guards remain unchanged.
✅ 3. Hard enforcement assertions preserve distinct missing-token and invalid-token failures.

**Action taken**
- Approved after explicit maintainer authorization and successful required CI on head `fc2db0c6`.

PR #713, opened 2026-06-01
docs: continue retiring hub terminology

baixiaohang:
## Summary
- replace remaining active user-facing and code-comment Hub wording with First Tree / First Tree Cloud / server language
- add `FirstTreeSDK` as a public alias so docs no longer need `FirstTreeHubSDK as FirstTreeSDK`
- update related test assertions and repo-local skills

## Scope intentionally retained
- internal identifiers such as `FirstTreeHubSDK`, `HubClient`, `deriveHubUrlFromToken`, `HubUrlDerivationError`, and `renderHubBlock`
- protocol / storage names such as `hub_organization_id`, `x-hub-signature-256`, `hub-session-*`, and Kael `hub_*` payload fields
- historical migration/proposal references and fixture hosts such as `hub.test`

## Validation
- `pnpm check`
- `pnpm typecheck`
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/barrel-exports.test.ts src/__tests__/agent-slot.test.ts src/__tests__/git-mirror-manager.test.ts`
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/github-app-installations.test.ts src/__tests__/github-app-orphan-recovery.test.ts src/__tests__/github-app-token.test.ts`

PR #1578, opened 2026-07-09
fix: read runtime session tokens from agent files

yuezengwu:
> **Review note:** this PR was retargeted to `main` after the server prerequisite landed. The final Files changed view contains only the 12 client/CLI/QA delivery files; present-on-bind is removed, and `agent:bind` does not send `currentRuntimeSessionToken`.

## Summary

- Makes the per-agent token file the runtime-session token source of truth for client and CLI callers.
- Keeps `agent:bind` session-only: the client no longer presents a current token during bind, and the server mints a fresh token on every successful bind.
- Registers per-agent token providers so in-process SDK and CLI HTTP resolve the file fresh for every request and retry.
- Writes token files atomically with temp-file + rename and removes inherited plaintext token values and stale token-file paths from provider subprocess environments.
- Changes CLI local-agent token resolution to read only `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE`; unreadable, empty, or missing files produce token-less calls instead of local failure or env fallback.
- Replaces the old owner-reuse QA case with `packages/qa/cases/cross-surface/agent-runtime-session-token-delivery.md`.

## Dependency

The server prerequisite landed in #1577. This PR is based on current `main`, is not behind it, and its final diff is limited to the client/CLI/QA delivery changes.

## Root Cause

Long-lived SDK instances and subprocess environment snapshots could hold a stale runtime-session token after server-side bind rotation. A one-time CLI file read had the same failure mode inside a multi-request SDK instance. The fix removes environment token values as an authority and makes every agent-scoped HTTP request resolve the current token file.

## Compatibility

- If the token file contains the current value, SDK/CLI agent-scoped HTTP sends it.
- If the token file is missing or empty, SDK/CLI calls run token-less rather than falling back to `FIRST_TREE_RUNTIME_SESSION_TOKEN`.
- Provider subprocesses receive only the current token-file path; inherited token values and stale paths are scrubbed.
- Token plaintext remains client-side only; the server keeps hash-only runtime-session metadata.

## Non-Goals

- Does not introduce owner-client token reuse.
- Does not solve two processes intentionally or accidentally sharing the same `client_id`; that requires client identity/home isolation.
- Does not fix the separate portable packaging issue around missing `agent-briefing.ejs`.

## Companion Context Tree PR

- https://github.com/agent-team-foundation/first-tree-context/pull/713

The tree PR records the approved runtime-private token-file decision, why environment-only delivery fails across rotation, and the permission/rotation/cleanup/recovery constraints. It remains draft until this code PR lands and is then reconciled against the merged implementation.

## Validation

- `pnpm --dir apps/cli exec vitest run src/__tests__/agent-shared-helpers-extra.test.ts src/__tests__/local-agent-runtime-token-request.test.ts`
- `pnpm --dir packages/client exec vitest run src/__tests__/agent-io.test.ts`
- `pnpm --dir packages/client exec vitest run src/__tests__/sdk-comprehensive.test.ts`
- `pnpm check` passed with existing unrelated Biome warnings/info.
- `pnpm typecheck` passed.
- `git diff --check` passed.

## QA

Formal QA warranted, using `packages/qa/cases/cross-surface/agent-runtime-session-token-delivery.md`. The case covers per-bind rotation, atomic token-file persistence, fresh reads, no env-token fallback, missing/empty file behavior, bogus env values, and enforcement=false/true outcomes.


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1578

- Rationale: the implementation still snapshots the CLI token and can preserve a stale plaintext env token, so the file is not yet the single fresh authority claimed by the PR.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` -> A.
- Semantic lift: none; runtime credential delivery is already covered by the A baseline.

**PR summary**
- Problem: after a daemon reconnect/rebind rotates an agent runtime-session token, long-lived SDK and provider processes must keep making agent-scoped HTTP calls instead of becoming stale-token 403 failures.
- Approach: persist each bind's token atomically in a per-agent file, let in-process SDK calls resolve that file through a provider, remove intentional plaintext env injection, and make CLI calls ignore the legacy env value.
- Impacted modules: `packages/client`, `apps/cli`, and the cross-surface QA case.

**Review findings**

❌ 1. `createSdk()` still reads the token file once before constructing the SDK, then passes a string. A single CLI process can perform multiple HTTP calls or SDK retries; if a rebind rotates the file between them, later attempts still send the old token and can fail under enforcement. Pass the resolver function into `runtimeSessionToken` and add a regression that changes the file after SDK construction and observes the next request using the new value. [R1 / `apps/cli/src/commands/_shared/local-agent.ts:92`]

❌ 2. Removing the explicit token assignment does not remove an inherited plaintext token. `buildAgentEnv()` spreads all of `parentEnv`, so `FIRST_TREE_RUNTIME_SESSION_TOKEN` supplied by the daemon's parent environment or by agent-config env survives unchanged into the provider subprocess. Explicitly scrub the legacy value (and avoid inheriting a stale token-file path when no current path is supplied) before layering the current per-agent file path; cover a parent env containing both legacy keys. [R4 / `packages/client/src/runtime/agent-io.ts:86`]

❌ 3. The durable authority change is not reconciled with the active Context Tree. `system/cloud/agents/runtime-binding.md` currently says agent-local files must not store per-agent bearer tokens, while this PR makes a per-agent token file the source of truth. The related proposal also describes the runtime token as ephemeral and non-persistent. Please pair this code change with an approved Context Tree update (or revise the implementation to the current constraint) so future runtime/auth work has one authoritative contract.

**Action taken**
- Submitted request changes on branch head `cd8d7b77`.
- Static review only; I did not run checks or tests.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1578 (synchronize follow-up)

- Rationale: this synchronization only merged the updated parent branch; #1578's final delta is unchanged and all three prior blockers remain.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` -> A.
- Semantic lift: none.

**PR summary**
- The intended behavior remains file-authoritative runtime-session token delivery across rebinds for long-lived SDK and CLI callers.
- Compared with the new `fix/runtime-agent-token-server` head, the same 11 files and 225 additions / 56 deletions remain in #1578.

**Review findings**

❌ 1. CLI token resolution is still eager at SDK construction, not fresh per HTTP request or retry. [`apps/cli/src/commands/_shared/local-agent.ts:92`]

❌ 2. `buildAgentEnv()` still spreads inherited/configured `FIRST_TREE_RUNTIME_SESSION_TOKEN` into provider subprocesses instead of explicitly scrubbing it. [`packages/client/src/runtime/agent-io.ts:86`]

❌ 3. The file-as-authority decision still conflicts with the active Context Tree's no-per-agent-bearer-token constraint and has no companion approved tree update.

**Action taken**
- Re-submitted request changes on synchronized head `0b22ce4e`.
- Static follow-up only; I did not run checks or tests.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1578

**Recommendation**: approve
- Rationale: the refreshed implementation closes the stale-token paths for both in-process and child-process callers, and the durable file-delivery decision is paired with the correct draft Context Tree update.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` -> A.
- Semantic lift: none.

**PR summary**
- Problem: after reconnect/rebind rotates an agent runtime-session token, long-lived SDK and CLI callers can retain stale credentials and lose agent-scoped HTTP while the provider process remains alive.
- Approach: every SDK request/retry resolves the current per-agent token file; provider subprocess environments receive only the current file path after legacy token values and stale paths are scrubbed; bind continues to mint rather than present/reuse a prior token.
- Impacted modules: `packages/client`, `apps/cli`, and the cross-surface QA case.

**Review findings**
✅ The previous CLI snapshot blocker is resolved with a token provider and a real request-level regression that observes rotation after SDK construction.
✅ The previous env-inheritance blocker is resolved by explicitly removing both legacy token values and stale token-file paths before injecting the current path.
✅ Draft Context Tree PR #713 records the runtime-private file decision and follows the code-first / reconcile-after-merge workflow.
✅ The server prerequisite #1577 is merged; no new blocking finding remains in the final branch-head diff.
⚠️ Before landing, retarget this PR from the already-merged parent branch to `main` and confirm the final diff; approval applies to head `4355d778`.

**Action taken**
- Static review only; I did not run checks or tests.

PR #1585, opened 2026-07-10
fix(ci): prevent vitest OOM on GitHub runners

bestony:
## Summary

- Cap Vitest forks under CI (`VITEST_MAX_FORKS=1`, shared helper + package configs for CLI/client/server/web) so multi-fork workers no longer stack toward the ~4GB Node heap ceiling on ~7GB GH runners.
- Limit turbo package fan-out (`--concurrency=2`) and **run `first-tree-dev` tests alone after other packages**, so the CLI suite never shares the runner with server/web graphs (root cause of `first-tree-dev#test` OOM / tinypool `Channel closed`).
- Keep a single `vitest run` per package (no multi-batch shell chains); CLI config forces `isolate: true` and serial file execution under CI, with tighter mock/fixture cleanup in the heaviest login/runtime suites.

## Context

Observed on CI Test jobs (e.g. [run 29064487984](https://github.com/agent-team-foundation/first-tree/actions/runs/29064487984/job/86273066158)):

```text
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
Mark-Compact ~4057 MB
Error: Channel closed (ERR_IPC_CHANNEL_CLOSED)
Failed: first-tree-dev#test
```

This is infrastructure memory pressure, not a product assertion failure.

## Verification

- `pnpm check` (pre-existing repo warnings only)
- `CI=true VITEST_MAX_FORKS=1 pnpm --filter first-tree-dev exec vitest run` on a sample of CLI unit files
- `pnpm exec turbo run test --filter='!first-tree-dev' --dry-run` confirms CLI is excluded from the first step

## Test plan

- [ ] CI **Test** job is green on this PR
- [ ] Confirm job logs show `Test packages (excluding CLI)` then `Test CLI` as separate steps
- [ ] Confirm no `Reached heap limit` / `Channel closed` in CLI vitest output
- [ ] Spot-check local `pnpm test` still works without `CI=true` (default maxForks=2)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,12 @@ jobs:
           --health-retries 10
     env:
       CI_DATABASE_URL: postgres://test:test@localhost:5432/test
+      # GitHub-hosted runners are ~7GB. Node's default heap is ~4GB/process;
+      # stacking turbo package suites + multi-fork vitest OOM'd CLI workers
+      # (Mark-Compact ~4057MB, then tinypool `Channel closed`). Cap forks
+      # package-wide; the steps below also limit turbo concurrency and run
+      # the CLI suite alone so it never shares the runner with server/web.
+      VITEST_MAX_FORKS: "1"
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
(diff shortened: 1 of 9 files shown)


yuezengwu (approved):
Reviewed and approved.

Goal/core change as I understand it: this is a CI-only mitigation for Vitest OOM on GitHub-hosted runners. It caps Vitest fork count under CI, limits Turbo package fan-out, and moves the CLI test package into its own final step so the heavy CLI suite does not share the runner with server/web test graphs.

Checks I did locally:
- `CI=true VITEST_MAX_FORKS=1 pnpm exec turbo run test --filter='!first-tree-dev' --concurrency=2 --dry-run` shows the first CI step runs the six non-CLI packages and excludes `first-tree-dev`.
- `pnpm exec turbo run test --dry-run` still includes `first-tree-dev`, so local/root `pnpm test` coverage is not dropped.
- Client/web/server Vitest configs load under `CI=true VITEST_MAX_FORKS=1` with a no-file `--passWithNoTests` smoke.
- `git diff --check origin/main...HEAD` is clean.

One non-blocking note: running `login-command.test.ts` directly still fails four account-switch cases locally, but I reproduced the same failures on `origin/main`, so I am not treating that as introduced by this PR.

PR #714, opened 2026-06-01
feat(web): unify Team section disclosure carets + left-step indent

Gandy2025:
## What

Fixes the Team page group-collapse **arrows**. Today the disclosure carets are inconsistent and the hierarchy reads flat: Agent teammates has no caret while Human teammates does, and the Public/Private subgroup carets sit on nearly the same column as the section carets, so you can't tell parent from child.

This reworks both sections into a consistent left-side disclosure tree:

- **Agent teammates is now collapsible**, with a caret and a clickable title row, matching Human teammates. Both top-level sections are uniform.
- **The All / Mine filter stays exactly where it is** (right of the Agent title row). It's a sibling of the flex-grow toggle button — never nested inside it — so it keeps working and never wraps to its own line.
- **One indent step (`SECTION_BODY_INDENT`) expresses the hierarchy.** Section carets sit at the left edge (x0). The whole section body — column header, Public/Private subgroup carets, and every member row — shares one indented left edge (x1), one step in. Member Name never re-indents per group, so **Agent members (under Public/Private) and Human members line up on one column** and the five columns stay aligned.

## Before → after (concept)

```
before:                          after:
🤖 Agent teammates  (no caret)    ⌄ 🤖 Agent teammates        [All|Mine]
 ⌄ Public                            ⌄ 🤖 Public
   Kael ...                             Kael ...
 ⌄ Private                           ⌄ 🔒 Private
⌄ 👤 Human teammates                    Scout ...
   Gandy ...                       ⌄ 👤 Human teammates
                                       Gandy ...   ← same Name column as agents
```

## Scope

Arrows / indent / member alignment **only**. Untouched: column structure, fields, All/Mine position & behavior, count badges, colors, the Usage column, the kebab, and the Human Delegate/Last-active columns. One file, +58/−23.

## Test

- `@first-tree/web` tests: **514/514 pass** (incl. `team-groups` 10/10).
- `biome check`, `tsc --noEmit`, and `lint:tokens` design-token guard: clean.
- Pure presentational change — no API/schema/data changes, no migrations.

> Note: the local full-suite run showed the known pre-existing `apps/cli` `login-command.test.ts` 5s-timeout flakiness (load-dependent under parallel turbo; passes 4/4 in isolation). This diff touches **0 cli files**; the same flake did not affect CI on #687 / #692.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1587, opened 2026-07-10
feat: support max and ultra Codex reasoning

yuezengwu:
## Summary

- add provider-native `max` and `ultra` reasoning effort values to the Codex agent config variant
- expose the new model-dependent levels in Agent Runtime, `agent config set-reasoning-effort`, and the canonical CLI reference
- preserve the literals through app-server turns and the SDK fallback without local downgrade or compatibility matrices
- add deterministic shared, server, client, Web, CLI coverage plus a reusable formal-QA case

## Compatibility

- default remains `high`
- no database migration or persisted-config rewrite
- model/account compatibility stays provider-owned; unsupported combinations use the existing explicit provider-failure path
- the SDK boundary has a documented narrow assertion because `@openai/codex-sdk@0.144.1` runtime forwards the values while its TypeScript union still stops at `xhigh`

Official provider guidance:

- https://developers.openai.com/api/docs/guides/latest-model#what-is-new
- https://learn.chatgpt.com/docs/agent-configuration/subagents#reasoning-effort-model_reasoning_effort

## Verification

- `pnpm check` (passes; three pre-existing warnings/info remain outside this diff)
- targeted package suites:
  - shared: 563 tests
  - client: 1,475 tests
  - Web: 1,310 tests
  - CLI reasoning config: 6 tests
  - server agent config: 14 tests
- `pnpm test` was attempted; all feature-related packages passed, but the CLI suite is blocked in this Codex-hosted workspace by 7 existing local-client-switch process-scan failures. The same 7 failures reproduce on a clean `origin/main` worktree because the scanner treats `CODEX_THREAD_ID` / Codex PATH entries as untrusted provider processes.

## QA

Formal QA is warranted because this crosses configuration surfaces and provider runtime paths. A reusable `codex-reasoning-effort` case is included; formal Docker QA was not auto-run because it is human-requested.

## Context Tree

Companion draft: https://github.com/agent-team-foundation/first-tree-context/pull/714

It records only the durable compatibility boundary, not the provider value list or implementation details. Merge this implementation first, then reconcile the Context Tree PR against the final merged behavior before marking it ready.

### DIFF
--- a/apps/cli/src/__tests__/agent-config-cli.test.ts
+++ b/apps/cli/src/__tests__/agent-config-cli.test.ts
@@ -63,6 +63,18 @@ describe("agent config CLI registration (Step 8)", () => {
     expect(setEnv?.options.some((o) => o.long === "--sensitive")).toBe(true);
   });
 
+  it("documents Codex max and ultra reasoning effort", () => {
+    const root = new Command();
+    const agent = root.command("agent");
+    registerAgentConfigCommands(agent);
+    const reasoningEffort = agent.commands
+      .find((c) => c.name() === "config")
+      ?.commands.find((c) => c.name() === "set-reasoning-effort");
+
+    expect(reasoningEffort?.description()).toContain("xhigh | max | ultra");
+    expect(reasoningEffort?.description()).toContain("model-dependent");
+  });
+
   it("add-mcp accepts --transport / --command / --url / --args", () => {
     const root = new Command();
     const agent = root.command("agent");
--- a/apps/cli/src/commands/agent/config/set-reasoning-effort.ts
+++ b/apps/cli/src/commands/agent/config/set-reasoning-effort.ts
@@ -7,7 +7,7 @@ export function registerAgentConfigSetReasoningEffortCommand(config: Command): v
   config
     .command("set-reasoning-effort <agent> <level>")
     .description(
-      'Set reasoning effort. claude-code: "" (inherit local) | low | medium | high | max. codex: low | medium | high | xhigh.',
+      'Set reasoning effort. claude-code: "" (inherit local) | low | medium | high | max. codex: low | medium | high | xhigh | max | ultra (model-dependent).',
     )
     .action(async (agentName: string, level: string) => {
       const serverUrl = resolveServerUrl(process.env.FIRST_TREE_SERVER_URL);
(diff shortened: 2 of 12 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1587

- Rationale: The provider-native pass-through is sound, but the public CLI reference and one validation-boundary comment still contradict the expanded Codex contract.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: Operators cannot currently select the provider-native `max` or `ultra` Codex reasoning levels through First Tree without local rejection or downgrade.
- Approach: Extend the provider-specific config union, Web/CLI surfaces, server persistence, and both Codex runtime paths so the literal is preserved while model/account compatibility remains provider-owned.
- Impacted modules: shared runtime config, server config service coverage, Agent Runtime Codex app-server/SDK paths, Web agent detail, CLI, and formal QA cases.

**Review findings**
❌ 1. Add `agent config set-reasoning-effort` and its provider-specific accepted values to `docs/cli-reference.md`. This PR changes a user-visible CLI contract, but the canonical `agent config` command tree still omits the command entirely, so users following the repo reference cannot discover either the existing setting or the new model-dependent values. [R5 / `docs/cli-reference.md:262`]

⚠️ 2. Update the patch-schema comment that still uses `"max" for codex` as its example of a rejected value. After this PR that example is false and directly contradicts the tagged union whose validation boundary the comment is documenting. [R1 / `packages/shared/src/schemas/agent-runtime-config.ts:454`]

✅ 3. The core compatibility direction checks out: current provider guidance documents `max` and `ultra` as model-dependent effort values, the app-server forwards the literal on `turn/start`, and inspection of `@openai/codex-sdk@0.144.1` confirms its runtime serializer passes `modelReasoningEffort` through as `model_reasoning_effort` even though the declaration union stops at `xhigh`.

✅ 4. Companion Context Tree PR #714 stays at the durable boundary: provider-owned compatibility, unchanged forwarding, explicit failure, and no local downgrade/fallback matrix.

**Action taken**
- Submitted request changes.

baixiaohang (approved):
Approved after follow-up review. The canonical CLI reference now documents `set-reasoning-effort` and the provider-specific/model-dependent values, and the stale validation example has been corrected. The provider-native pass-through, SDK boundary, runtime paths, deterministic coverage, QA case, and companion durable-context boundary are consistent.

PR #1589, opened 2026-07-10
fix(cli): avoid Windows cmd shims for npm and Codex

baixiaohang:
## Summary

- resolve Windows npm installs through `node.exe npm-cli.js` instead of spawning `.cmd` shims
- preserve a PATH + shell fallback for non-standard Node layouts without embedding an unquoted absolute path
- convert synchronous child-process spawn failures into normal installer failure results
- apply the shared npm invocation path to CLI self-upgrade, npm metadata probes, and Codex/Claude runtime installers
- resolve Windows Codex npm shims to the native `codex.exe` they wrap
- normalize real paths in the Codex shim test for macOS `/var` vs `/private/var` portability

## Validation

- [x] affected-file Biome check
- [x] `pnpm typecheck` (10/10 Turbo tasks)
- [x] CLI focused tests: 27 passed
- [x] Client Codex tests: 17 passed, 2 platform-skipped
- [x] real Windows npm invocation smoke test (`npm --version` through `node.exe npm-cli.js`)
- [ ] `pnpm check` — blocked on Windows by the checkout-wide CRLF/LF mismatch; all 9 touched files pass Biome
- [ ] `pnpm test` — the existing Windows batch runner itself fails at `spawnSync vitest.cmd EINVAL`; focused changed-surface suites pass

## Change Surface

- [x] `apps/cli` install and self-upgrade behavior

## Root cause

On Node v22.16.0 for Windows, directly spawning `C:\Program Files\nodejs\npm.cmd` throws `EINVAL`. Adding `shell: true` alone is insufficient because cmd.exe truncates the unquoted absolute path at `C:\Program`. Standard Node installations expose npm's JavaScript entry point, so this change executes it as `node.exe npm-cli.js` without a shell.

### DIFF
--- a/apps/cli/src/__tests__/core-attention-update-extra.test.ts
+++ b/apps/cli/src/__tests__/core-attention-update-extra.test.ts
@@ -4,6 +4,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { resolveNpmInvocation } from "../core/npm-invocation.js";
 
 const bootstrapMocks = vi.hoisted(() => {
   class ServerUrlNotConfiguredError extends Error {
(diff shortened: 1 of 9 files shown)


yuezengwu (approved):
Reviewed the Windows process-spawn path end to end. This PR centralizes npm invocation so standard Windows installs use `node.exe npm-cli.js`, preserves the PATH/shell fallback for custom layouts, maps synchronous registry spawn failures into installer results, and resolves Codex npm shims to the native platform executable.

I also checked the published `@openai/codex@0.144.1` Windows package layout and npm alias resolution against the implementation; the resolved package and `vendor/x86_64-pc-windows-msvc/bin/codex.exe` path match the real artifact.

No database, schema, or core data-structure changes are involved. Local validation: `pnpm typecheck` passed all 10 tasks; the client Codex suite passed 19/19; the three CLI focused files passed 25 tests with 1 platform skip. One additional unchanged portable-platform assertion fails on this macOS arm64 host because it hardcodes `freebsd-x64`; it is unrelated to this diff and the PR's Windows-focused validation covers the target behavior.

No blocking findings. Approving.

PR #1590, opened 2026-07-10
Make Context Tree setup chat-first

Gandy2025:
## Summary

Make Context Tree setup chat-first across `/context`, the setup-chat kickoff service, and `first-tree-seed`.

- Remove GitHub App installation, Settings, repository catalog/picker, and source-resource registration as `/context` setup gates.
- Route every admin setup/recovery state into one private Context Tree setup chat; keep member and no-agent states explicit.
- Resolve source inside chat from declared workspace repos, a local Git repo, or a GitHub URL.
- Continue Phase 2 in the same setup chat after the approved Phase 1 PR is merged.

## Product behavior

- Missing tree: **Build your Context Tree** opens the setup chat.
- Bound but unreadable tree: **Work on this in chat** opens the setup chat with the diagnostic context.
- Live tree: no setup CTA.
- Member: explains that an admin owns setup; no unauthorized kickoff.
- No managed agent: routes to agent creation/onboarding.

The retired onboarding kickoff route now returns a structured, non-mutating `410 tree_setup_kickoff_moved`. No new CLI command is introduced in this PR.

## Safety and migration

- Setup-chat idempotency is scoped to the current human-agent pair, preserving private-chat ACLs across multiple admins and agents.
- A legacy organization-scoped setup chat is adopted only when its complete membership is exactly the current human and selected agent; otherwise it is left untouched and a correctly scoped chat is created.
- The server owns the canonical bootstrap/topic; clients cannot inject arbitrary setup payloads or idempotency keys.

## Seed skill and evals

- `first-tree-seed` v0.2.5 accepts an empty/absent source manifest and asks for a local Git repo or GitHub URL in chat.
- Declared bare source repos are read only through managed source worktrees.
- Verified same-chat Phase 2 continuation is checked before the ordinary non-empty-tree refusal.
- Live gates cover empty-manifest chat source and same-chat Phase 2 continuation.
- The grader now rejects failed, skipped, wrong-path, wrapped-shell, mixed-output, and spoofed worktree/source evidence while accepting proven successful flows.

## Verification

- Independent code review: **no findings** after adversarial replay of ACL, migration, shell-control-flow, and evidence-spoof cases.
- Independent E2E: **PASS** using real PostgreSQL, a production build, and Playwright in Chromium + Firefox. Covered missing/unreadable/live/member/no-agent states, two-admin isolation, legacy adoption/mismatch, strict auth/schema, retired-route `410`, and onboarding state preservation; zero console errors and no application `5xx` responses.
- Web focused tests: 71/71.
- Server focused tests: 56/56, with the final endpoint suite independently rerun at 24/24.
- Seed focused tests: 79/79.
- Shared schema tests: 5/5.
- Monorepo typecheck: 10/10 packages.
- `pnpm check`: pass (only 2 pre-existing warnings and 1 pre-existing info).
- `git diff --check`, skill validation, skill floor, and Context Tree verification: pass.

The earlier full `pnpm test` attempt exposed two unrelated baseline issues: the macOS `/var` vs `/private/var` shim expectation (subsequently fixed on `main`) and a periodic fixture exceeding its 5-second default; the isolated fixture passes with the suite's 20-second timeout, and all changed-area suites plus the independent E2E are green.

## Context Tree

Draft Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/715

Merge this code PR first, then reconcile and mark the Context Tree PR ready.


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1590

- Rationale: The chat-first product path is coherent, but the new seed live gate both rejects a valid source-read pattern and does not actually exercise the same-chat-history condition it claims to protect.

**Risk level**: A
- Semantic lift: cross-surface Context setup, private-chat ACL migration, and seed lifecycle behavior; no lift above A

**PR summary**
- Problem: Admins currently have to clear GitHub App, Settings, and repository-selection gates before they can build or recover a Context Tree, and Phase 2 cannot reliably resume in the original setup conversation.
- Approach: Route every admin unavailable state into a server-owned private setup chat, scope reuse to the initiating human-agent pair, resolve readable sources inside chat, and allow a verified same-chat continuation after Phase 1 merges.
- Impacted modules: Web Context surface, server Context/onboarding kickoff APIs, shared schemas, client briefing, `first-tree-seed`, and seed eval infrastructure.

**Review findings**

❌ 1. Preserve pipeline provenance when deciding whether source evidence is attributable. `segmentCanProduceIndependentEvidence` returns `true` for every content-reading program, while `containsSourceFixtureEvidence` has already discarded the `|` connector. A legitimate read such as `cat worktrees/seed-source-repo/README.md | head -50` therefore marks `head` as an unrelated output producer and sets `sourceEvidenceReadObserved=false`, despite all output being derived from the source stream. This contradicts the adjacent comment and can fail a correct live run. Keep connector metadata (or otherwise distinguish pipeline filters from independent file readers) and add an acceptance test for a piped source read. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:824`]

❌ 2. Exercise actual prior-turn history in the same-chat continuation gate. The case supplies a single prompt that says “Earlier in this visible chat…” and the generated `AGENTS.md` independently asserts that the skeleton was approved in this chat; no Phase 1 proposal, approval turn, or PR handoff exists in the evaluated conversation. A regression that ignores visible history and accepts any current-message claim on a populated tree would still pass this gate. Provide real prior messages (or an equivalent transcript fixture the model must inspect) and add the negative sibling where the tree is Phase-1-shaped but the same-chat proposal/approval/handoff evidence is absent. [`packages/skill-evals/src/suites/first-tree-seed/cases.ts:183`, `packages/skill-evals/src/suites/first-tree-seed/fixture.ts:36`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1590

The implementation consistently moves `/context` setup into a server-owned, private human-agent chat and removes GitHub installation/resource selection from the browser gate. The private-chat key migration and the retired `410` route are coherent, and the focused server/web/shared/seed tests pass. I found three additional blockers beyond the two seed-eval findings already filed by baixiaohang.

### Findings

1. **[P1] Preserve the authoritative-manifest failure boundary in the canonical bootstrap.**

   `packages/server/src/api/orgs/context-tree.ts:50` tells the agent that whenever no source is readable it should ask for a local folder or GitHub URL. That conflicts with the shipped seed contract: a non-empty `manifest.sources` is authoritative, and a missing declared clone is a half-provisioned workspace that must fail closed rather than being bypassed with an ad hoc source. Because this text is the server-owned visible task message, it can direct the exact unsafe fallback the skill forbids. Narrow the fallback to an empty/absent source manifest and explicitly tell the agent to report missing declared sources; add a route/bootstrap assertion for that distinction.

2. **[P1] A recovery CTA must send a current recovery turn and diagnostic when the setup chat already exists.**

   `packages/server/src/api/orgs/context-tree.ts:62` reuses the same kickoff helper for both initial setup and bound-tree recovery. `kickoffOnboarding` only sends `TREE_SETUP_BOOTSTRAP` when the chat has no messages; for an existing setup chat `created.initialMessageCreated` is false, `result.sent` is absent, nobody is notified, and the browser merely navigates to old history. The client also posts only `agentUuid`, so even a newly-created recovery chat never receives the snapshot error, repo/branch, or structured recovery diagnosis promised by the PR behavior. This is the common post-setup failure path: a bound tree becomes unreadable after its setup chat already contains Phase 1/2 history. Reuse the private chat, but append a server-owned, deduplicated recovery message carrying the current diagnostic (or introduce a strict server-resolved recovery intent), and test both new-chat and existing-chat recovery.

3. **[P2] Do not use synthetic fixture phrases as the proof oracle for the real-repo periodic case.**

   `packages/skill-evals/src/suites/first-tree-seed/grader.ts:854` requires two matches from `SOURCE_FIXTURE_EVIDENCE_HINTS` (`Apollo Console`, `CLI App`, etc.) for every seed case, including `real-first-tree-bare-readable`. Those phrases describe the synthetic fixture, not ordinary first-tree Tier 0/1 source material. A successful read of the actual repository `README.md` produces `sourceEvidenceReadObserved=false`; outside the eval implementation itself, the current repo contains only one of the seven phrases. The real-repo periodic case therefore rejects a normal proven source read and can pass only accidentally if the model reads the eval's own fixture/grader text. Make evidence tokens case-specific (or prove successful reads from the materialized source path without synthetic content coupling) and add an acceptance test using actual real-repo README/meta output.

### Database-sensitive behavior

This PR does not add a schema migration, but it does mutate existing database data by re-keying legacy `chats.onboarding_kickoff_key` rows inside `adoptSafeLegacyTreeSetupChat`. The exact-participant and speaker-only checks look appropriate; this should still receive human scrutiny because it changes persisted chat identity/idempotency state.

### Verification

- Server focused: 38/38 passed.
- Web focused: 19/19 passed.
- Shared schema: 5/5 passed.
- Seed grader/fixture/floor: 79/79 passed.
- Reproduced the real-repo evidence-oracle false negative directly: a successful `cat worktrees/seed-source-repo/README.md` event over the actual README yields `sourceEvidenceReadObserved=false`.
- Current CI and CodeQL checks are green; these findings are uncovered behavioral gaps rather than existing test failures.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: The five previously reported blockers are addressed in the new head, but the loop-provenance relaxation introduces a new false-green for skipped source reads.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. Do not treat a source-path marker from one loop iteration as proof that the source-reading iteration succeeded. `sourceBoundLoopVariables` exempts every read through the loop variable from the independent-output check, and `loopOutputBindsSource` then accepts the whole command whenever its aggregate output merely contains a source-root string. For example, this successful command never reads the source file, but will still set `sourceEvidenceReadObserved=true` if `context-tree/NODE.md` supplies two fixture hints:

```sh
for f in worktrees/seed-source-repo/README.md context-tree/NODE.md; do
  echo "FILE $f"
  test "$f" = context-tree/NODE.md && sed -n 1,80p "$f"
done
```

The first iteration prints the source path and skips `sed`; the second reads only the Context Tree and determines the loop's zero exit status. The current oracle sees a source-bound variable, the printed source marker, and matching aggregate output, so it credits source evidence despite the skipped source read. This re-opens the failed/skipped/mixed-output spoof class the grader is meant to close. Keep mixed loops conservative unless success/output can be attributed to the specific source iteration, and add this skipped-source negative case. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:956`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1018`]

**Resolved from prior reviews**
- Canonical bootstrap now preserves the authoritative non-empty manifest failure boundary.
- New and existing recovery chats receive a server-resolved diagnostic turn; existing chats are notified and immediate identical retries deduplicate.
- Pure source pipelines retain provenance while independent file operands remain rejected.
- Phase 2 continuation now requires reading an explicit prior-turn transcript and has a missing-history negative gate.
- Real-repo periodic evidence uses real first-tree README/package signals instead of synthetic fixture phrases.

**Action taken**
- Submitted request changes on head `647cf3d3`.

yuezengwu (changes requested):
## Re-review by yzw-codex — PR #1590

The synchronized patch resolves all five previously filed issues: the canonical bootstrap now fails closed for missing declared sources, reused recovery chats receive a deduplicated diagnostic turn and wake-up, pipeline filters retain source provenance, the same-chat gate has a positive transcript plus a missing-history sibling, and the real-repo case uses real repository evidence. I independently reproduced the mixed-loop false-green reported on this head, and found two additional skipped/spoofed-evidence paths in the same grader.

### Findings

1. **[P1] Prove that the transcript read executed before crediting same-chat history.**

   `containsChatHistoryEvidence` currently requires only an overall exit code of zero, any content-reader segment mentioning `.first-tree-eval/chat-history.md`, and two matching phrases anywhere in aggregate output. It does not prove that the reader segment ran or bind the phrases to that read. This command therefore sets `chatHistoryReadObserved=true` even though `cat` is skipped and the output is self-authored:

   ```sh
   false && cat .first-tree-eval/chat-history.md; printf 'Phase 1 proposal\nApproved\n'
   ```

   That leaves the new continuation gate able to pass without inspecting the prior-turn transcript it is intended to enforce. Apply the same successful-segment and mixed/literal-output provenance rules used for source evidence, and add failed, skipped, and spoofed transcript-read negatives. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1032`]

2. **[P1] Bind `HEAD is now at …` evidence to the managed worktree add.**

   The fallback in `eventSuccessfullyMaterializesSourceWorktree` accepts a managed `git worktree add` segment anywhere in the command plus `HEAD is now at …` anywhere in aggregate output. After the patch relaxed the former paired Git-output check to that single line, an unrelated successful Git operation can credit an add that shell control flow skipped:

   ```sh
   false && git -C source-repos/source-repo worktree add "$WORKSPACE/worktrees/seed-source-repo" origin/main
   git -C context-tree reset --hard HEAD
   ```

   The second command supplies the matching output and the overall zero exit status, so `sourceWorktreeMaterializedObserved=true` even though no source worktree exists. Attribute success to the exact add (or require independent managed-path state/evidence) and add this skipped-add/unrelated-Git-output negative. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:305`]

The mixed-loop blocker already filed by baixiaohang also reproduces exactly: a source-path heading from the skipped iteration plus fixture phrases read only from `context-tree/NODE.md` yields `sourceEvidenceReadObserved=true`.

### Verification

- Server setup/recovery endpoint: 24/24 passed.
- Seed grader/fixture/floor/periodic: 94/94 passed.
- Server and skill-evals typechecks passed.
- Full GitHub CI and CodeQL are green on `647cf3d3`.
- `git diff --check` passed.
- Direct reproductions return all three incorrect positive metrics: mixed-loop source evidence, skipped transcript history, and skipped worktree materialization.

No new database schema change was added by this synchronization. The previously flagged transactional re-key of persisted `chats.onboarding_kickoff_key` remains unchanged and still warrants human scrutiny.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: The three blockers from `647cf3d3` are directly covered, but the tightened provenance rules still allow a skipped transcript read and a skipped source-loop read to borrow self-authored or Context Tree output.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. Prove the transcript reader ran instead of relying on literal-text detection. The new history oracle rejects `printf` only when the required phrases appear literally in that `printf` segment, but variable expansion bypasses that check while `readObserved` still credits the skipped `cat`:

```sh
H='Phase 1 proposal\nApproved\nPhase 1 PR handoff'
false && cat .first-tree-eval/chat-history.md
printf '%b\n' "$H"
```

The overall command exits zero, the command text contains a content reader naming the transcript, `printf '%b' "$H"` is not considered an independent output producer, and the literal-output scan sees only `$H`; the expanded aggregate output supplies all three hints. `chatHistoryReadObserved` therefore remains true without any transcript read. Bind evidence to a reader segment proven to have executed, or reject unbound output-producing shell builtins/expansions on this gate, and add this negative case. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1084`]

❌ 2. A `FILE <source>` block still does not prove the block's evidence came from reading that source. `outputContainsSourceEvidenceBlock` trusts every line after the marker until the next marker, and `loopOutputBindsSource` overrides the independent-output rejection. A loop can therefore skip the `$f` read for the source iteration but read `context-tree/NODE.md` under the source marker, then use `$f` only in the Context Tree iteration so the variable is classified as source-bound:

```sh
for f in worktrees/seed-source-repo/README.md context-tree/NODE.md; do
  echo "FILE $f"
  test "$f" = context-tree/NODE.md && sed -n 1,80p "$f"
  test "$f" = worktrees/seed-source-repo/README.md && sed -n 1,80p context-tree/NODE.md
  true
done
```

No source content is read, but the first source-marked block contains the Context Tree's fixture hints, `loopOutputBindsSource=true`, and the command passes the source-evidence oracle. The marker must be coupled to a successful read of that exact marked path; block position alone is insufficient. Add this swapped-reader negative case. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1005`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1069`]

**Resolved from the previous head**
- Mixed-loop evidence is now partitioned by source-marked output blocks for the originally reported ordering.
- Transcript evidence requires all three expected facts and rejects the direct literal/skipped/mixed cases already reported.
- Worktree materialization requires paired Git output for the single-add fallback and can independently verify a clean managed worktree at the expected source HEAD.

**Action taken**
- Submitted request changes on head `805d5e4b`.

yuezengwu (changes requested):
## Re-review by yzw-codex — PR #1590

Head `805d5e4b` directly covers the three previous reproductions: the originally reported mixed-loop ordering now partitions output blocks, transcript checks require all three lifecycle facts, and worktree materialization restores paired Git evidence plus an on-disk expected-HEAD check. The recovery-chat path also now defers session activation and the workspace kick until its outer transaction commits, and the concurrent recovery test passes. Two provenance boundaries remain unsound, including one additional direct-source variant.

### Findings

1. **[P1] Require an executed reader; variable-expanded output still credits skipped history and source reads.**

   Both evidence functions identify a reader from command text, but neither proves that segment executed. `printf`/`echo` are excluded from `segmentCanProduceIndependentEvidence`, while the literal-spoof check examines only those segments' literal text. Putting the expected phrases in a variable therefore bypasses both checks:

   ```sh
   H='Phase 1 proposal\nApproved\nPhase 1 PR handoff'
   false && cat .first-tree-eval/chat-history.md
   printf '%b\n' "$H"
   ```

   This sets `chatHistoryReadObserved=true` without reading the transcript. The same issue also affects the source oracle:

   ```sh
   H='Apollo Console\nruntime coordination'
   false && cat worktrees/seed-source-repo/README.md
   printf '%b\n' "$H"
   ```

   That sets `sourceEvidenceReadObserved=true` without reading source. Bind evidence to a reader segment proven to execute, or—now that the eval prompt explicitly requests standalone reads—fail closed unless the command is a single reader/pure-filter pipeline with no unrelated assignments, control-flow branches, or output producers. Add both variable-expanded negative cases. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1038`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1084`]

2. **[P1] A source-marked output block must be coupled to reading that exact marked path.**

   `outputContainsSourceEvidenceBlock` treats every line after `FILE <source>` as source output until the next marker, and `loopOutputBindsSource` then overrides the independent-reader rejection. A loop can therefore read `context-tree/NODE.md` under the source marker and use `$f` only on the other iteration:

   ```sh
   for f in worktrees/seed-source-repo/README.md context-tree/NODE.md; do
     echo "FILE $f"
     test "$f" = context-tree/NODE.md && sed -n 1,80p "$f"
     test "$f" = worktrees/seed-source-repo/README.md && sed -n 1,80p context-tree/NODE.md
     true
   done
   ```

   No source content is read, but the source-marked block contains the Context Tree fixture hints and `sourceEvidenceReadObserved=true`. Block position is not provenance; require proof that the successful reader operand equals the marked source path, or conservatively reject mixed loops and rely on the standalone source-read shape already requested by the case. Add this swapped-reader negative. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1005`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1069`]

### Core-service review note

This synchronization changes the core `sendMessage` return/options contract to support deferred post-commit effects. I reviewed the only new caller: the recovery path requests the descriptor inside the locked outer transaction, flushes it after commit, and preserves notification/session behavior. I found no blocker in that refactor. It does not add a database schema migration.

### Verification

- Server recovery/message/landing-campaign focused suites: 83/83 passed.
- Seed grader/fixture/floor/periodic: 100/100 passed.
- Server and skill-evals typechecks passed.
- `git diff --check` passed.
- CodeQL, lint/typecheck, migration, skill validation, and portable smoke checks are green; the aggregate GitHub Test job was still running when this review was submitted.
- Direct replay returns `true` for all three incorrect metrics: variable-expanded skipped history, variable-expanded skipped direct source, and swapped-reader source loop.

The earlier persisted `chats.onboarding_kickoff_key` legacy re-key remains unchanged and still warrants the previously requested human database scrutiny.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: The exact-path and mixed-loop changes close the swapped-reader case, but the skipped-reader variable-expansion false-green remains unchanged for both transcript and direct-source evidence.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. **[P1] The variable-expanded skipped-reader reproductions still pass because no reader execution is proven.** This head changes the loop branch to reject an independent Context Tree reader, but the direct-source predicate at lines 1084-1111 and the history predicate at lines 1124-1140 still infer a read from command text. Assignments are not output producers, `printf` is excluded by `segmentCanProduceIndependentEvidence`, and the literal-spoof check examines only the `printf` segment, where it sees `$H` rather than the expanded phrases. Consequently both commands below still satisfy their respective evidence predicates without executing either `cat`:

```sh
H='Phase 1 proposal\nApproved\nPhase 1 PR handoff'
false && cat .first-tree-eval/chat-history.md
printf '%b\n' "$H"
```

```sh
H='Apollo Console\nruntime coordination'
false && cat worktrees/seed-source-repo/README.md
printf '%b\n' "$H"
```

Bind positive credit to a reader segment proven to execute, or enforce the standalone reader/pure-filter pipeline shape already requested by the eval prompt. Add both variable-expanded negatives so this previously reported blocker is actually covered. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1084`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1124`]

**Resolved on this head**
- An explicit independent Context Tree reader now invalidates a source-marked loop block, closing the swapped-reader false-green.
- Managed worktree evidence now requires the declared bare source repo, exact target path, and—when verified from final state—the expected common Git directory and HEAD.
- Keyed task-chat initial sends defer session activation and the workspace kick until the outer transaction commits; I found no new blocker in that core-service change.

**Action taken**
- Submitted request changes on head `f2e0e502`.

yuezengwu (changes requested):
## Re-review by yzw-codex — PR #1590

Head `f2e0e502` correctly removes the mixed-loop output-block override, binds managed worktree evidence to the declared bare repo plus exact target/common-dir/HEAD, and defers keyed task-chat initial-send effects until the outer transaction commits. The swapped-reader reproduction now fails as intended and I found no blocker in the keyed-chat transaction change. One previously reported evidence-execution blocker remains unchanged.

### Finding

1. **[P1] Variable-expanded output still credits both skipped reader paths.**

   `containsSourceFixtureEvidence` and `containsChatHistoryEvidence` still infer a read from any reader segment present in the command text rather than a segment proven to execute. Assignments are not independent output producers, `printf` is excluded from `segmentCanProduceIndependentEvidence`, and the literal-spoof scan sees `$H` rather than the variable's expanded phrases. These two commands therefore still produce incorrect positive metrics without executing either `cat`:

   ```sh
   H='Phase 1 proposal\nApproved\nPhase 1 PR handoff'
   false && cat .first-tree-eval/chat-history.md
   printf '%b\n' "$H"
   ```

   Result: `chatHistoryReadObserved=true`.

   ```sh
   H='Apollo Console\nruntime coordination'
   false && cat worktrees/seed-source-repo/README.md
   printf '%b\n' "$H"
   ```

   Result: `sourceEvidenceReadObserved=true`.

   Bind positive credit to a reader/pipeline proven to execute, or fail closed to the standalone reader/pure-filter pipeline shape the eval prompt already requests. Add both variable-expanded negatives; this head adds no test for either reproduction. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:1084`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:1124`]

### Verification

- Server keyed-chat/setup/message focused suites: 56/56 passed.
- Seed grader/fixture/floor/periodic: 104/104 passed.
- Server and skill-evals typechecks passed.
- `git diff --check` passed.
- Direct replay: swapped-reader source loop is now correctly `false`; variable-expanded skipped history and direct-source reads remain incorrectly `true`.
- CodeQL, lint/typecheck, migration, skill validation, and portable smoke checks are green; aggregate GitHub Test was still running when this review was submitted.

No database schema migration is added by this synchronization. The core keyed-chat post-commit change was reviewed and looks sound; the earlier legacy `chats.onboarding_kickoff_key` persisted re-key remains unchanged and still carries the prior human-review note.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: The standalone-command shape closes the variable-expansion reproductions, but it still trusts path text rather than a reader operand and it rejects several legitimate pure-filter pipelines.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. **[P1] A path literal inside an `awk` program is still treated as an executed source/transcript read.** `segmentReadsContent` classifies every `awk` command as a reader, while both required-path predicates only search the raw segment text. `commandProvesStandaloneContentRead` therefore accepts a one-segment `awk BEGIN` program even when it has no file operand at all. These commands produce incorrect positive metrics without reading either target:

```sh
awk 'BEGIN {
  print "worktrees/seed-source-repo/README.md"
  print "Apollo Console"
  print "runtime coordination"
}'
```

Result: `sourceEvidenceReadObserved=true`.

```sh
awk 'BEGIN {
  print ".first-tree-eval/chat-history.md"
  print "Phase 1 proposal"
  print "Approved"
  print "Phase 1 PR handoff"
}'
```

Result: `chatHistoryReadObserved=true`.

Prove that the required path is a file operand of the reader (with program-specific argument parsing), not merely a substring anywhere in its script/pattern, and add both standalone-spoof negatives. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:877`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:979`]

❌ 2. **[P1] Preserve legitimate `sed`/`awk`/`jq` source pipelines.** `segmentIsPurePipelineFilter` explicitly knows how to prove those three programs have no independent file operand, but the pipeline branch then hard-codes a narrower list that omits all three. A valid concise source read such as:

```sh
cat worktrees/seed-source-repo/README.md | sed -n '1,80p'
```

now returns `sourceEvidenceReadObserved=false` even though every emitted byte is derived from the source stream. This regresses the pipeline-provenance behavior the earlier review required. Reuse the pure-filter predicate without a contradictory second allowlist (or include every supported pure filter), and add acceptance coverage for at least `sed`. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:954`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:985`]

**Resolved on this head**
- Variable-expanded output after a skipped reader is now rejected for both transcript and direct-source evidence.
- Mixed loops and unrelated output producers are conservatively rejected by the standalone-reader shape.

**Action taken**
- Submitted request changes on head `a537938c`.

yuezengwu (changes requested):
## Re-review — PR #1590 (`a537938c`)

This head replaces the mixed-loop/output-block heuristics with a narrower standalone-reader predicate. That correctly closes the previously reported variable-expanded skipped-reader cases, but the new predicate still does not prove that the required path is a reader operand and it rejects legitimate pure-filter pipelines.

### Findings

1. **[P1] Bind the required path to an actual reader file operand.** `isReader` combines a broad program-name check with a raw substring predicate. A command therefore receives positive credit when the path appears only inside an `awk` program, grep pattern, here-string, or shell comment and no target file is read. I reproduced both incorrect metrics on this head:

   ```sh
   awk 'BEGIN {
     print "worktrees/seed-source-repo/README.md"
     print "Apollo Console"
     print "runtime coordination"
   }'
   # sourceEvidenceReadObserved=true
   ```

   ```sh
   awk 'BEGIN {
     print ".first-tree-eval/chat-history.md"
     print "Phase 1 proposal"
     print "Approved"
     print "Phase 1 PR handoff"
   }'
   # chatHistoryReadObserved=true
   ```

   The same false-green occurs with `grep '<required-path>' <<< '<self-authored hints>'`. Parse program-specific operands so the required source/transcript path must be the file input of the successful reader; pattern/script/comment text and stdin-only reads must not qualify. Please add source and transcript negatives for this case. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:877`]

2. **[P1] Accept the pure filters the implementation already proves safe.** `segmentIsPurePipelineFilter` explicitly validates stdin-only `sed`, `awk`, and `jq`, but `commandProvesStandaloneContentRead` applies a second allowlist that excludes all three. As a result, this legitimate source-derived pipeline is rejected:

   ```sh
   cat worktrees/seed-source-repo/README.md | sed -n '1,80p'
   # sourceEvidenceReadObserved=false
   ```

   Reuse the pure-filter predicate without the contradictory allowlist (or keep the supported-program set consistent), and add an acceptance test for `sed`. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:954`]

### Resolved on this head

- Variable-expanded output after a skipped transcript or direct-source reader is rejected.
- Mixed loops and unrelated output producers are conservatively rejected by the standalone-command shape.

### Verification

- Replayed the two prior variable-expansion cases, the swapped-reader loop, both operandless spoof cases, and the `sed` pipeline against `deriveMetrics`.
- First-tree-seed focused tests: 106/106 passed.
- Skill-evals typecheck: passed.
- `git diff --check`: passed.
- Current CI and CodeQL checks: all passed.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: Reader operands are now parsed, but they are not bound to the current eval workspace and accepted pipeline filters can still synthesize the evidence text.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. **[P1] Bind evidence operands to the current run's exact source/transcript paths.** `pathBelongsToSourceEvidence` accepts any operand containing an allowed suffix, and `pathIsChatHistory` accepts any path ending in the transcript suffix. Neither receives `paths.workspacePath`. A model can therefore author and read unrelated files outside the fixture while receiving positive credit:

```sh
mkdir -p /tmp/worktrees/seed-source-repo
printf 'Apollo Console\nruntime coordination\n'   > /tmp/worktrees/seed-source-repo/README.md
cat /tmp/worktrees/seed-source-repo/README.md
```

Result: `sourceEvidenceReadObserved=true`, even though the managed source worktree was not inspected.

```sh
mkdir -p /tmp/.first-tree-eval
printf 'Phase 1 proposal\nApproved\nPhase 1 PR handoff\n'   > /tmp/.first-tree-eval/chat-history.md
cat /tmp/.first-tree-eval/chat-history.md
```

Result: `chatHistoryReadObserved=true`, even though the visible-chat transcript was not inspected. The write commands are not among `forbiddenSideEffectHits`, so a run can retain these credits. Resolve relative operands against the captured workspace and require the exact run-scoped managed root/transcript (with the source mode taken into account); add wrong-root negatives for both metrics. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:965`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:973`]

❌ 2. **[P1] Accepted pipeline filters must not be able to manufacture evidence text.** The pipeline branch accepts `rg` based only on positional-operand count. Output-rewriting options are ignored, so a valid source operand can be followed by a filter that replaces every byte the model sees with the expected phrases:

```sh
cat worktrees/seed-source-repo/package.json |
  rg --replace='Apollo Console runtime coordination' '.+'
```

This satisfies `commandProvesStandaloneContentRead` and sets `sourceEvidenceReadObserved=true` even when the source stream contains neither hint. The same `--replace` shape can synthesize all three transcript facts after opening a transcript that does not contain them. Since this head now explicitly prompts `cat ... | head -50`, the safest contract is to allow only content-preserving filters/options required by that shape; otherwise parse and reject rewriting/label-producing modes. Add source and transcript negatives. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:939`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:991`]

**Resolved on this head**
- Operandless `awk`/pattern spoofing is rejected by program-specific file-operand parsing.
- Mixed file operands are rejected.
- The live gate now prescribes the narrow `cat transcript` and `cat source | head -50` evidence shapes, so the previous broad-`sed` acceptance concern no longer blocks the compliant path.

**Action taken**
- Submitted request changes on head `885c5e2b`.

yuezengwu (changes requested):
## Re-review — PR #1590 (`885c5e2b`)

This head narrows evidence credit to parsed `cat`/`head`/`tail` file operands and prescribes deterministic `cat transcript` plus `cat source | head -50` commands in the continuation gate. That closes the prior operandless `awk`/pattern spoof and makes the narrower pipeline contract reasonable, but two provenance gaps still let fabricated evidence pass.

### Findings

1. **[P1] Bind reader operands to this eval run's exact source and transcript paths.** `pathBelongsToSourceEvidence` accepts an allowed path suffix under any ancestor, while `pathIsChatHistory` accepts the transcript suffix anywhere. Neither predicate receives `paths.workspacePath`. I reproduced both false-greens on this head:

   ```sh
   mkdir -p /tmp/unrelated/worktrees/seed-source-repo
   printf 'Apollo Console\nruntime coordination\n' > /tmp/unrelated/worktrees/seed-source-repo/README.md
   cat /tmp/unrelated/worktrees/seed-source-repo/README.md
   # sourceEvidenceReadObserved=true
   ```

   ```sh
   mkdir -p /tmp/unrelated/.first-tree-eval
   printf 'Phase 1 proposal\nApproved\nPhase 1 PR handoff\n' > /tmp/unrelated/.first-tree-eval/chat-history.md
   cat /tmp/unrelated/.first-tree-eval/chat-history.md
   # chatHistoryReadObserved=true
   ```

   Those writes are not forbidden side effects, so a run can retain the credits. Resolve relative operands against the run workspace and require the exact mode-specific managed source root / transcript path; add wrong-root negatives for both metrics. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:965`]

2. **[P1] Restrict accepted pipeline filters to content-preserving behavior.** The pipeline branch still accepts `rg` while ignoring output-rewriting options. A successful reader can therefore be followed by a filter that manufactures the expected evidence:

   ```sh
   cat worktrees/seed-source-repo/package.json |
     rg --replace='Apollo Console runtime coordination' '.+'
   # sourceEvidenceReadObserved=true
   ```

   The same shape with `--replace='Phase 1 proposal Approved Phase 1 PR handoff'` sets `chatHistoryReadObserved=true`. Because the gate now explicitly prescribes `cat ... | head -50`, the safest contract is to admit only the required content-preserving filter/options; otherwise parse and reject rewriting or label-producing modes. Add source and transcript negatives. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:939`]

### Resolved on this head

- Required paths must now be parsed reader operands, closing operandless `awk`, grep-pattern, here-string, and comment-text spoofs.
- Mixed file operands are rejected.
- The live continuation gate now prescribes the narrow evidence command shapes, so broad `sed` pipeline acceptance is no longer required for that compliant path.

### Verification

- Replayed both prior skipped-reader cases, the swapped-reader loop, operandless `awk`/grep spoofs, wrong-root source/transcript reads, and `rg --replace` source/transcript spoofs directly against `deriveMetrics`.
- First-tree-seed focused tests: 111/111 passed.
- Skill-evals typecheck: passed.
- `git diff --check`: passed.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: Run-scoped operands and content-preserving filters are fixed, but command-text worktree credit can still be supplied by a no-op Git alias/function and override a failed common-dir state check.

**Risk level**: A
- Semantic lift: none above A

**Review finding**

❌ 1. **[P1] Prove that `worktree add` is the actual Git subcommand, not merely later argv consumed by an alias or shadowed function.** `isManagedWorktreeAdd` searches for the words `worktree add` anywhere in the segment; it does not parse the actual Git subcommand. For a run whose workspace is `/tmp/eval`, this real command exits zero without creating a worktree but is credited as materialization:

```sh
git -C source-repos/source-repo \
  -c alias.foo='!true' \
  foo worktree add /tmp/eval/worktrees/seed-source-repo origin/main
```

Here `foo` is the executed shell alias and `worktree add ...` is only its ignored argument list, yet the checker sees program `git`, the expected `-C`, an exact target after the later words, and exit code 0. The same false credit is available through the final-list rule:

```sh
git() { return 0; }
git -C source-repos/source-repo worktree add /tmp/eval/worktrees/seed-source-repo origin/main
```

`segmentsProvenSuccessfulByExitCode` discards the function-definition lists and credits the final shadowed `git` segment.

This becomes a full protocol bypass when combined with an ordinary clean clone at the managed path: final-state common-dir verification correctly rejects the clone, but `sourceWorktreeMaterializedObserved` is the OR of that state result and the spoofed event credit; `sourceRepoChanged` accepts the clone when it is clean at the baseline HEAD, and a subsequent exact-path `cat` supplies source evidence. Parse the exact standalone Git command shape promised by the eval prompt (including the real subcommand position and no prior shell lists/config aliases), or require the managed common-dir state rather than allowing command credit to override it. Add the alias, function-shadow, and ordinary-clone-plus-event regression cases. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:258`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:298`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:313`]

**Resolved on this head**
- Source and transcript operands are bound to the current eval workspace and the mode-specific source root.
- Evidence pipelines are restricted to non-rewriting `cat`/`head`/`tail` filters.
- Aggregate Git-looking output is no longer used as materialization proof.

**Action taken**
- Submitted request changes on head `4620f662`.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: This synchronization hardens content-reader executables and expanded operands, but it does not change the outstanding worktree-command proof; a no-op Git alias/function still supplies materialization credit.

**Risk level**: A
- Semantic lift: none above A

**Review finding**

❌ 1. **[P1] The Git alias/function-shadow worktree bypass remains unchanged on this head.** `eventSuccessfullyMaterializesSourceWorktree` still searches for `worktree add` anywhere in a successful segment rather than proving it is the executed Git subcommand. Substituting this run's actual absolute target, the following exits zero without creating a worktree but sets `sourceWorktreeMaterializedObserved=true`:

```sh
git -C source-repos/source-repo \
  -c alias.foo='!true' \
  foo worktree add /absolute/eval/worktrees/seed-source-repo origin/main
```

The final-list variant also remains accepted:

```sh
git() { return 0; }
git -C source-repos/source-repo worktree add /absolute/eval/worktrees/seed-source-repo origin/main
```

As reported on the previous head, placing an ordinary clean clone at the managed path then supplies exact-path source evidence and passes `sourceRepoChanged`; the spoofed event credit is ORed with the correctly-failing common-dir state check, so the full bare-worktree gate can pass without a managed worktree. Require the exact standalone Git subcommand shape (no prior shell lists or `-c` alias command) or do not let command credit override failed common-dir state. Add alias, function-shadow, and ordinary-clone-plus-event tests. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:258`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:298`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:313`]

**Resolved on this head**
- Source/transcript reader executables are restricted to trusted `cat`/`head`/`tail` spellings.
- Shell-expanded, globbed, brace-expanded, redirected, and control-syntax operands are rejected.

**Action taken**
- Submitted request changes on head `41d00fd7`.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: Final-state-only worktree verification closes command spoofing, but it accepts an external worktree through a managed-path symlink and cannot detect evidence files that are temporarily rewritten and restored.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

❌ 1. **[P1] Require the managed worktree leaf itself to be a real directory inside the eval workspace.** `sourceWorktreeMaterializedAtExpectedHead` runs Git through the expected lexical path and validates only status, common-dir, and HEAD. A symlink at that path to an external worktree of the declared bare clone satisfies every check:

```sh
git -C source-repos/source-repo worktree add /tmp/external-source origin/main
ln -s /tmp/external-source worktrees/seed-source-repo
cat worktrees/seed-source-repo/README.md
```

The final path exists, Git resolves through the symlink to a clean baseline-HEAD worktree whose common-dir is the declared bare clone, `sourceRepoChanged` also passes through the symlink, and the exact lexical source read earns evidence. The gate therefore passes even though the worktree was materialized outside the required managed target; cleanup also removes only the link/fails to unregister the external worktree. Reject a symlinked leaf and require the canonical worktree top-level to equal the canonical `<workspace>/worktrees/seed-source-repo` directory. Add an external-worktree symlink negative. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:468`]

❌ 2. **[P1] Detect transient writes to protected source/transcript evidence, not only final cleanliness.** The new final-state proof makes this sequence pass:

```sh
cp worktrees/seed-source-repo/package.json /tmp/original-package.json
printf 'Apollo Console\nruntime coordination\n' \
  > worktrees/seed-source-repo/package.json
cat worktrees/seed-source-repo/package.json
mv /tmp/original-package.json worktrees/seed-source-repo/package.json
```

The `cat` earns `sourceEvidenceReadObserved=true`; after restoration, status is clean, HEAD/common-dir are correct, and `sourceRepoChanged=false`. The model never inspected the original source evidence. The transcript gate has the same issue: it can replace `.first-tree-eval/chat-history.md` with self-authored proposal/approval/handoff text, `cat` it, then restore the fixture. `validateFixture` checks only that the history path exists, and `forbiddenSideEffectHits` does not implement the declared `source_write` side effect or protect the transcript. Record protected-file mutations at event time (or otherwise make/verify these fixtures immutable across the run), and add write-read-restore negatives for source and history. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:538`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:945`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:965`]

**Resolved on this head**
- Worktree materialization is credited only from final clean managed state linked to the declared bare clone; Git alias/function and ordinary-clone event overrides are closed.
- All require-worktree live prompts now explicitly keep the worktree in place for final provenance.

**Action taken**
- Submitted request changes on head `d958472f`.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1590

- Rationale: The managed-path symlink is now rejected, but the new baseline checks inspect files only after the run and therefore still accept evidence captured during a temporary rewrite that was later restored.

**Risk level**: A
- Semantic lift: none above A

**Review finding**

❌ 1. **[P1] Bind evidence output to protected content at read time, not to the restored final file.** `pathBelongsToSourceEvidence` hashes the source file when `deriveMetrics` runs, while `containsSourceFixtureEvidence` consumes output recorded by an earlier command event. The previously reported sequence therefore remains valid:

```sh
cp worktrees/seed-source-repo/package.json /tmp/original-package.json
printf 'Apollo Console\nruntime coordination\n' > worktrees/seed-source-repo/package.json
cat worktrees/seed-source-repo/package.json
mv /tmp/original-package.json worktrees/seed-source-repo/package.json
```

At grading time the restored file is tracked, clean, and byte-identical to `HEAD`, so lines 945-952 accept the operand; the recorded `cat` output still contains the fabricated hints, so lines 1007-1010 set `sourceEvidenceReadObserved=true`. The transcript path has the same time-of-check gap: temporarily replace `.first-tree-eval/chat-history.md`, read the self-authored lifecycle phrases, then restore `approvedPhase1ChatHistoryMarkdown()`; lines 956-962 validate the restored fixture while lines 1021-1025 credit the earlier fabricated output.

The new tests leave the fake source/transcript content in place until grading, so they do not exercise the write-read-restore reproduction. Detect protected-file writes from the recorded command events, make the fixtures immutable, or capture/verify content identity at the read event; add transient-restoration negatives for both source and history. [`packages/skill-evals/src/suites/first-tree-seed/grader.ts:922`, `packages/skill-evals/src/suites/first-tree-seed/grader.ts:994`]

**Resolved on this head**
- The exact managed worktree leaf rejects symlinks.
- Current source files must be tracked, byte-identical to HEAD, and free of ignored/assume-unchanged/skip-worktree state.
- The current transcript must exactly match the canonical approved Phase 1 fixture.

**Action taken**
- Submitted request changes on head `fb227e60`.

yuezengwu (approved):
## Re-review — PR #1590 (`6436363f`)

**Recommendation:** approve

The chat-first Context Tree setup path remains coherent across Web, setup-chat kickoff, private-chat migration, and `first-tree-seed`. This final delta closes the last seed-grader provenance blocker by replaying the validated read shape against the final protected files and requiring exact output equality.

### Findings

No blocking findings on this head.

### Verified resolutions

- Managed source worktrees must be real directories at the canonical run-scoped path, clean, at the baseline HEAD, and linked to the declared bare clone; symlink, Git-alias/function, and ordinary-clone substitutes fail.
- Source evidence operands must resolve to tracked, byte-identical baseline files; ignored, untracked, assume-unchanged, skip-worktree, wrong-root, expanded, mixed, and rewriting-filter cases fail.
- Source and transcript reads are replayed against final protected content, so overwrite/read/restore output no longer earns credit.
- The transcript must remain the canonical approved Phase 1 history and the recorded output must exactly match replayed content.
- Prior setup-chat ACL, recovery diagnostic, canonical bootstrap, same-chat continuation-history, and post-commit message activation findings remain resolved.

### Verification

- Replayed the complete tracked-file overwrite/read/restore gate: `casePassed=false`, `sourceEvidenceReadObserved=false`, while the valid clean managed worktree remains recognized.
- Replayed all earlier skipped-reader, mixed-loop, wrong-root, expanded-path, filter-rewrite, worktree alias/function, ordinary-clone, symlink, ignored-file, and transcript cases.
- First-tree-seed focused tests: 131/131 passed.
- Skill-evals typecheck: passed.
- `git diff --check`: passed.

No new schema or persisted-data behavior is introduced by this head. The overall PR's legacy `chats.onboarding_kickoff_key` re-key remains the previously flagged item for human database/data-migration scrutiny before merge.

yuezengwu (approved):
## Re-review — PR #1590 (`125eb6c5`)

**Recommendation:** approve

This post-approval synchronization is a narrow provenance hardening: the canonical transcript path must now remain under the canonical eval workspace, so a symlinked `.first-tree-eval` parent cannot redirect an otherwise valid-looking history read. It preserves the previously approved content-replay and worktree guarantees.

### Findings

No blocking findings.

### Verification

- Replayed the complete tracked-file overwrite/read/restore gate: `casePassed=false` and `sourceEvidenceReadObserved=false`, with the valid managed worktree still recognized.
- New symlinked transcript-parent negative is covered by the focused suite.
- First-tree-seed focused tests: 132/132 passed.
- Skill-evals typecheck: passed.
- `git diff --check`: passed.

No schema, persisted-data, product-flow, or runtime behavior changes are introduced by this delta. The overall PR's legacy `chats.onboarding_kickoff_key` re-key remains the previously flagged item for human database/data-migration scrutiny before merge.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1590

**Recommendation**: approve
- Rationale: The chat-first setup flow is coherent, all previously reported ACL, recovery, source-provenance, worktree, and same-chat continuation blockers are resolved, and the current head has all required GitHub checks passing.

**Risk level**: A
- Semantic lift: none above A

**Review findings**

No blocking findings on head `125eb6c5`.

✅ The server-owned private setup chat preserves human-agent ACL isolation and appends deduplicated recovery diagnostics with post-commit effects.

✅ The authoritative-manifest boundary fails closed while empty manifests can resolve a local Git checkout or GitHub URL in chat.

✅ Managed bare-source worktrees are bound to the canonical workspace path, declared bare common-dir, baseline HEAD, and clean final state.

✅ Source and transcript evidence uses validated reader shapes replayed against protected final content with exact output equality, closing skipped, spoofed, wrong-root, rewriting-filter, symlink, and overwrite/read/restore paths.

**Database-sensitive note**

No schema migration is introduced. The legacy `chats.onboarding_kickoff_key` re-key remains a persisted-data migration behavior for human scrutiny before merge.

**Action taken**
- Approved head `125eb6c5` after explicit authorization and confirmation that all GitHub checks passed.

PR #1592, opened 2026-07-10
fix: harden doc preview fence scanning

liuchao-001:
## Summary

Fixes #1572.

This hardens the shared doc-preview bare markdown path scanner so fenced code block state follows CommonMark-style closing rules instead of toggling on every fence-looking line.

## Root Cause

`scanBareDocPathTokens` previously tracked fenced blocks with a boolean toggle. A shorter ``` marker inside a longer ```` fence, or an unclosed fence, could invert the scanner state for the rest of the message. That caused illustrative `.md` paths inside code blocks to be captured as failed document mentions while real prose mentions after the block were skipped.

## Changes

- Track the active fence marker character and length.
- Close a fence only when a line uses the same marker character, at least the opening length, and only trailing whitespace.
- Treat unclosed fences as code through the end of the message.
- Add scanner regression coverage for nested/long fences, unclosed fences, and wrong marker/short closing lines.

## Validation

- `pnpm --filter @first-tree/shared exec vitest run src/lib/__tests__/doc-link-scan.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/doc-snapshots.test.ts src/__tests__/doc-snapshots.fs-mocks.test.ts`
- `pnpm --filter @first-tree/web exec vitest run src/lib/__tests__/doc-preview-links.test.ts`
- `pnpm check`
- `pnpm typecheck`

Note: `pnpm check` completed successfully but reported existing lint warnings outside this change.

Issue #1572, opened 2026-07-09
[Bug] doc-preview: unbalanced/nested code fence inverts the .md linkifier — code-block content wrapped as #doc-failed chips


liuchao-001:
## Summary

`scanBareDocPathTokens` tracks fenced-code-block state with a naive per-line toggle (`FENCE_OPEN_RE` flips `inFence` on **every** line starting with ``` or ~~~). Any message whose fence lines don't pair up the way the scanner assumes — an unclosed fence, or a ``` line legitimately nested inside a longer ```` fence (valid CommonMark) — inverts the scanner's in/out state for the rest of the message. From that point on, `.md` tokens **inside** fenced blocks are captured/linkified and `.md` tokens **outside** fences are skipped — the exact opposite of the documented behavior ("fenced code blocks are deliberately a hard skip").

Because the runtime capture pass and the web render pass share this scanner, the effect is user-visible: agent messages render with dead `#doc-failed?reason=missing` chips sprinkled through code-block content.

## Surface Area

Shared doc-preview scanner + its two consumers:

- `packages/shared/src/lib/doc-link-scan.ts` — `scanBareDocPathTokens`, `FENCE_OPEN_RE` toggle
- `packages/client/src/runtime/doc-snapshots.ts` — capture pass / `failedMentions`
- `packages/web/src/lib/doc-preview-links.ts` — `wrapFailedDocMentions`

## Environment

- Observed in a production cohort user chat (2026-07-08); scanner code verified unchanged at current `main`
- Reporter's CLI (staging): 0.5.13-staging.788.1, Darwin 25.2.0 arm64
- Runtime of the affected agent: codex (`my-codex`), but the bug is runtime-agnostic (shared scanner)

## Reproduction

Minimal synthetic case — a message that *documents* fences inside a longer fence (valid CommonMark), followed by an ordinary fenced diagram:

`````markdown
To open a code block, type:

````
```
````

Project layout:

```
repo/
├── NODE.md
├── docs/setup.md
└── sub/
    └── NODE.md
```

See `README.md` for details.
`````

Trace through `scanBareDocPathTokens`:

1. ```` ```` ```` opens the outer fence (`inFence=true`), the inner ``` line **toggles it off** mid-block, the closing ```` ```` ```` toggles it back **on** → state is now inverted.
[...]

PR #1595, opened 2026-07-10
fix: reject backticks in backtick-fence info strings in doc-link scan

liuchao-001:
## Summary

Follow-up to #1592, closing the last known fence-state divergence in the shared doc-preview scanner.

A line starting with a backtick fence marker that has more backticks after the marker (e.g. a Slack-style inline line like ```` ```pnpm test``` runs the suite ````) was treated as opening a fenced block. Per CommonMark, the info string of a backtick fence may not contain backticks, so such a line is actually a paragraph with inline code — which is exactly how react-markdown renders it.

## Root Cause

`parseOpeningFence` accepted any line matching the fence-marker pattern without checking the rest of the line. The misjudged phantom fence then swallowed following prose (missed links/snapshots), and worse, inverted fence parsing for subsequent real code blocks: the web linkify pass rewrites the markdown source, so it could inject literal `[path](path)` syntax into content the renderer displays as a code block — the same state-inversion class as #1572, triggered by a rarer input.

## Changes

- `parseOpeningFence`: a backtick fence line whose remainder contains a backtick is not an opening fence. Tilde fences are unaffected (CommonMark allows backticks in tilde-fence info strings).
- Regression tests: pseudo-fence line followed by prose, pseudo-fence line followed by a real fence (inversion case, with `enclosingCodeSpan` assertion for the inline-code path), and a tilde fence with backticks in its info string.

## Validation

- `pnpm --filter @first-tree/shared exec vitest run src/lib/__tests__/doc-link-scan.test.ts` (27 tests)
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/doc-snapshots.test.ts src/__tests__/doc-snapshots.fs-mocks.test.ts` (30 tests)
- `pnpm --filter @first-tree/web exec vitest run src/lib/__tests__/doc-preview-links.test.ts` (14 tests)
- `pnpm exec biome check` on the touched files

### DIFF
--- a/packages/shared/src/lib/__tests__/doc-link-scan.test.ts
+++ b/packages/shared/src/lib/__tests__/doc-link-scan.test.ts
@@ -117,6 +117,31 @@ describe("scanBareDocPathTokens", () => {
     expect(tokens(text)).toEqual(["docs/after.md"]);
   });
 
+  it("does not treat a backtick fence line with backticks after the marker as an opener", () => {
+    // CommonMark: a backtick-fence info string may not contain backticks, so
+    // this line renders as a paragraph with inline code, not a code block.
+    const text = ["```pnpm test``` runs the suite", "docs/after.md"].join("\n");
+    expect(tokens(text)).toEqual(["docs/after.md"]);
+  });
+
+  it("keeps real fences skipped after a backtick-containing pseudo-fence line", () => {
+    const text = [
+      "```docs/inline.md``` as inline code",
+      "```",
+      "docs/fenced.md",
+      "```",
+      "docs/after.md",
+    ].join("\n");
+    const matches = scanBareDocPathTokens(text);
+    expect(matches.map((m) => m.raw)).toEqual(["docs/inline.md", "docs/after.md"]);
+    expect(matches[0]?.enclosingCodeSpan).toBeDefined();
+  });
+
+  it("allows backticks in a tilde fence info string", () => {
+    const text = ["~~~ `note`", "docs/fenced.md", "~~~", "docs/after.md"].join("\n");
+    expect(tokens(text)).toEqual(["docs/after.md"]);
+  });
+
   it("ignores indented (4-space / tab) code blocks", () => {
     expect(tokens("    docs/indented.md")).toEqual([]);
     expect(tokens("\tdocs/tabbed.md")).toEqual([]);
--- a/packages/shared/src/lib/__tests__/doc-link-scan.test.ts
+++ b/packages/shared/src/lib/__tests__/doc-link-scan.test.ts
@@ -125,13 +125,7 @@ describe("scanBareDocPathTokens", () => {
   });
 
   it("keeps real fences skipped after a backtick-containing pseudo-fence line", () => {
-    const text = [
-      "```docs/inline.md``` as inline code",
-      "```",
-      "docs/fenced.md",
-      "```",
-      "docs/after.md",
-    ].join("\n");
+    const text = ["```docs/inline.md``` as inline code", "```", "docs/fenced.md", "```", "docs/after.md"].join("\n");
     const matches = scanBareDocPathTokens(text);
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1595

**Recommendation**: approve
- Rationale: The shared scanner now rejects CommonMark-invalid backtick fence openers without disturbing valid backtick or tilde fences, and the regression coverage exercises the downstream state-inversion risk.

**Risk level**: C
- Path baseline: only `packages/shared/**` is changed -> C
- Semantic lift: none; this corrects existing document-path scanning behavior without changing message state or a cross-surface contract.

**PR summary**
- Problem: A Slack-style one-line backtick span at the start of an agent message could be mistaken for a fenced block, hiding later real document mentions and causing subsequent actual code blocks to be scanned as prose.
- Approach: Reject a backtick fence opener when the remainder of its line contains any backtick, while retaining valid backtick info strings and allowing backticks in tilde-fence info strings as CommonMark specifies.
- Impacted modules: shared document-path scanner and its focused regression tests; client snapshot capture and web failed-mention rendering continue to consume the same shared decision.

**Review findings**
✅ 1. The opener guard matches CommonMark's backtick-only info-string restriction while leaving tilde fences unchanged.
✅ 2. The inversion regression verifies both sides of the shared contract: the pseudo-fence remains scan-visible with its inline-code annotation, while the following real fenced block stays excluded.

**Action taken**
- Approved.

yuezengwu (changes requested):
The change has the right goal and core implementation: keep the shared document-path scanner aligned with the renderer by rejecting CommonMark-invalid backtick fence openers, while preserving valid backtick fences and tilde fences. I verified the guard against the [CommonMark 0.31.2 fenced-code-block rule](https://spec.commonmark.org/0.31.2/#fenced-code-blocks), and the state-inversion regression covers the important shared client/web failure mode.

Requesting one required fix before approval:

- **Blocking — format the touched regression test** (`packages/shared/src/lib/__tests__/doc-link-scan.test.ts:128`). `pnpm exec biome check packages/shared/src/lib/doc-link-scan.ts packages/shared/src/lib/__tests__/doc-link-scan.test.ts` fails on the multiline `const text = [...]` fixture in the new inversion test. This is also the PR-caused error in the current `Lint & Type Check` job. Please apply the Biome formatting and rerun the touched-file check.

Semantic validation otherwise passed:

- shared scanner suite: 27/27
- client snapshot suites: 30/30
- web linkification suite: 14/14
- shared typecheck

No database, schema, or core data-structure change is involved.

baixiaohang (approved):
## Follow-up review by baixiaohang-code-reviewer - PR #1595

**Recommendation**: approve
- Rationale: The synchronized commit applies the requested Biome-only formatting change and leaves the already-reviewed scanner behavior, fixture values, and assertions unchanged.

**Risk level**: C
- Path baseline: only `packages/shared/**` is changed -> C
- Semantic lift: none.

**Increment reviewed**
- Head: `dd34e5bbafb6a0abe6adf16ebff6df6ff9c3ed47`
- Delta from `72f911942a53012a234b807e73744bda0bb8ef53`: one regression fixture is reformatted from a multiline array to Biome's single-line output.
- No production logic, test data, or assertion changed.

**Review findings**
✅ 1. The synchronized commit addresses the sole requested formatting fix without changing semantics.

**Action taken**
- Approved the updated head.

yuezengwu (approved):
Follow-up review of `72f91194..dd34e5bb`: the synchronized commit only applies Biome's requested formatting to the inversion-test fixture. Production logic, fixture values, and assertions are unchanged, so the previously reviewed CommonMark behavior and shared client/web contract remain intact.

Revalidated the updated head:

- touched-file Biome check passes
- shared scanner suite passes (27/27)
- `git diff --check` passes

The sole requested change is resolved. Approved.

PR #1594, opened 2026-07-10
feat(web): mobile phase 1 (Now / Chat / Team / Me shell)

Gandy2025:
## What

Promote **mobile phase 1** from the `feat/mobile-shell` integration branch to `main`. This bundles the reviewed mobile shell and its follow-up fixes into one promotion PR for a final look before it lands on `main`.

Contents of the branch vs `main`:
- `74f9276b` mobile experience foundation
- `2012dd61` **#1582** — complete mobile phase 1 shell routes (`/m/*`, Now/Chat/Team/Me, Me team switching, PWA metadata, channel gating)
- `03d97e6e` **#1593** — address review feedback (canonical attention ordering, source-neutral request label, no self-chat, explicit `presentation` mode preserving narrow-Workspace GitHub, collapsed mobile summary, Discord support link, design-doc removal)

Scope: `packages/web/**` only. **No** server, schema, migration, API-contract, auth, or core data-structure changes — mobile reuses the existing surfaces. The two source-repo design docs are removed; the design lives in the Context Tree proposal.

## Rollout (unchanged, channel-gated)

- `dev` / `staging`: `/m/*` enabled, phone-width pure `/` → `/m/now`, PWA metadata.
- `prod`: `/m/*` **disabled** — any `/m/*` visit `replace`-redirects to the desktop root `/` (standing desktop-only recovery invariant). Desktop Workspace unchanged.
- Enabling mobile in **production is out of this promotion** and remains a separate product-owner release decision after staging validation.

## Review status

- #1593 reviewed on `feat/mobile-shell`: **baixiaohang — approve**; **yzw-codex — no blocking findings**. (Both recorded as comments because #1593 merged into the branch while review completed.)
- Paired Context Tree proposal **agent-team-foundation/first-tree-context#706** reconciled to the final decisions; yzw-codex's latest review records **no changes requested / nothing unclear**; now marked ready for its formal approval.

## Gating

Kept **draft**. Per yzw-codex, promoting to `main` and enabling prod are gated on #706 reconciliation (done) and the production/recovery decisions (recorded in #706, pending product-owner confirmation). Will mark ready once #706 is approved and the product owner authorizes; **will not self-merge**.

Related: #1582, #1593, first-tree-context#706.

PR #1596, opened 2026-07-10
fix(web): make the mobile shell scroll and size to the dynamic viewport

Gandy2025:
## What

Follow-up to the merged mobile phase 1 (#1594). Fixes two mobile shell layout defects:

**1. Vertical scroll was broken (Now / Chat / Team / Me).** `MobilePage` used `min-h-full overflow-y-auto`. `min-height:100%` has no upper bound, so the scroll container grew with its content instead of being capped at the viewport height, and `overflow-y-auto` never engaged — content overflowed the `overflow-hidden` `<main>` and was clipped. Fixed by bounding it with `h-full`.

**2. Shell used the layout viewport, not the dynamic one (yzw-codex post-merge finding #1).** The shell root used inline `height:100vh` with an ineffective `minHeight:100dvh`, which resolves to the larger 100vh on mobile browsers while the address bar is visible and pushes the bottom tabs below the fold. Moved to a new `.h-dvh-screen` utility (`100vh` base + `@supports (height:100dvh)` override) and reserve the top safe-area inset via `.pt-safe-top` so content clears the status bar on notched devices.

## Verification

- **Browse (real browser, 390px):** with `h-full`, the scroll container has clientHeight 740 < scrollHeight 799 and scrollTop moves → scrolls. Control test forcing `min-h-full` reproduces the bug (clientHeight grows to content, `scrollable:false`) — causation confirmed.
- Added regression tests: the real `MobileShell` uses `.h-dvh-screen` + `.pt-safe-top` (not raw inline `100vh`), and `MobilePage` is a bounded `h-full` scroll container (guards against reverting to `min-h-full`).
- Gates green: full web suite (168 files / 1320 tests), typecheck + design-token guardrails, `pnpm check`, web build.

Scope: `packages/web/**` only. No server/schema/API/auth changes.

## Note

This is finding **#1** of yzw-codex's 4 post-merge review items on #1594. The others — #2 (mobile attention caches outside the `['me','chats']` realtime invalidation family), #3 (captured document links are a no-op on `/m/chat`), #4 (title rename still gated on `narrow` for generic narrow Workspace) — are separate follow-ups.

Related: #1594.

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -703,6 +703,29 @@
     font-variant-numeric: tabular-nums;
   }
 
+  /* Dynamic viewport height with a universal fallback: `100vh` ships as the
+     base for browsers without `dvh`, then `100dvh` overrides where supported
+     (iOS 15.4+, Chrome 108+) so a full-height shell tracks the mobile URL-bar /
+     home-indicator resize instead of overflowing the visible viewport. Split
+     into a base rule + `@supports` override rather than two `height`
+     declarations in one rule, which inline styles cannot express and the
+     linter flags as a duplicate property. */
+  .h-dvh-screen {
+    height: 100vh;
+  }
+  @supports (height: 100dvh) {
+    .h-dvh-screen {
+      height: 100dvh;
+    }
+  }
+
+  /* Reserve the top safe-area inset (notch / status bar) so a full-height
+     shell without its own top chrome does not render content under the
+     status bar on notched devices. */
+  .pt-safe-top {
+    padding-top: env(safe-area-inset-top);
+  }
+
   /* Surface utilities — consolidate Panel / PageHeader / Tab containers.
      Prefer these over hand-rolled background/border/radius combinations. */
   .surface-base {
--- a/packages/web/src/pages/mobile/__tests__/shell-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/shell-dom.test.tsx
@@ -5,7 +5,7 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { MemoryRouter, Route, Routes } from "react-router";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createDomHarness, type DomHarness } from "../../../test-utils/dom-harness.js";
-import { MobileSignalChip } from "../components.js";
+import { MobilePage, MobileSignalChip } from "../components.js";
 import { MobileShell } from "../shell.js";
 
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1596

**Recommendation**: approve
- Rationale: The focused follow-up correctly restores bounded page scrolling and the live mobile shell's dynamic-viewport/safe-area contract without changing mobile scope or shared data behavior.

**Risk level**: C
- Semantic lift: none; this is shell layout/CSS behavior and does not touch subscriptions, WebSocket parsing, or the data layer.

**PR 摘要**
- Problem: Phone users could not scroll long Now / Chat / Team / Me pages, and visible browser chrome or a notched status bar could push or cover mobile controls and bottom navigation.
- Approach: Bounds each root page to its parent so `overflow-y-auto` engages, sizes the live shell with a `100vh` fallback plus a supported `100dvh` override, and reserves the top safe-area inset at the shell boundary.

**Review findings**
✅ 1. `MobilePage` now uses a definite `h-full` bound under the shell's constrained `main`, so overflowing content scrolls instead of growing into the clipped ancestor.
✅ 2. The shell uses a real fallback/override height contract; with the existing `viewport-fit=cover` metadata and global `border-box`, the top safe-area padding stays inside the dynamic viewport height.
✅ 3. Regression coverage targets the live `MobileShell` and the shared `MobilePage`, guarding both defects without coupling to the separate preview shell.

**Action taken**
- Approved.

yuezengwu:
Post-merge independent review at `86e5a27b`: no blocking findings.

Goal and core changes reviewed:

- Restore scrolling on the mobile Now / Chat / Team / Me root pages by giving `MobilePage` a definite `h-full` bound under the shell's constrained `main`.
- Keep the shell inside the visible mobile viewport with a compiled `100vh` fallback plus `@supports (height: 100dvh)` override.
- Reserve the top safe-area inset at the shell boundary and guard both fixes with live-shell / shared-page regression tests.

Validation:

- The `main` flex item has the required `min-h-0` / bounded-height chain, so `overflow-y-auto` can engage without growing into the clipped ancestor.
- The production CSS keeps both height rules under the expected support query. Existing `viewport-fit=cover` metadata and global `border-box` keep top safe-area padding inside the shell height; bottom safe-area handling remains owned by the tab bar.
- Local full Web suite: 168 files / 1320 tests passed.
- Web typecheck, design-token guardrails, and shared + Web production builds passed.

Scope remains `packages/web/**`; there are no schema, API, auth, subscription, or shared data-structure changes.

PR #712, opened 2026-06-01
feat(client): claude-code-tui handler core (tmux + transcript tail)

yuezengwu:
## PR3 of the claude-code-tui slice series — the runtime handler

Builds on #678 (provider registration) and #684 (capability probe). The PR1 startup guard (`hasHandler`) kept older clients inert until this handler exists; once this lands, a bound `claude-code-tui` agent actually runs.

### Scope (client only)
No `shared` / `web` / `server` changes — purely the runtime handler under `packages/client/src/handlers/claude-code-tui/`.

| File | Role |
|---|---|
| `index.ts` | `AgentHandler` impl — start / resume / inject / suspend / shutdown |
| `tmux-session.ts` | thin `tmux` wrapper — newSession / pasteText / sendKey / capturePane / waitForReady / listOwnedSessions / killSession |
| `transcript-tail.ts` | tails claude's per-session JSONL and emits raw entries |
| `tui-markers.ts` | shared marker/sentinel constants |
| `ask-user-degrader.ts` | AskUserQuestion → plain-text round-trip |
| `handlers/index.ts` | registration (same `claude` PATH resolution as claude-code) |

### Design (rationale: `experiments/tmux-claude-runtime/FINDINGS.md`, landing in PR5)
- **No hooks.** The per-session transcript JSONL + `capture-pane` are the only signals — input is injected via `paste-buffer`. This is the locked contract for the runtime.
- **Reuses claude-code internals:** `createToolCallProcessor` / `mapMcpServers` and the shared bootstrap helpers, so event semantics match the SDK handler.
- **AskUserQuestion degraded gracefully** to a plain-text round-trip via Escape-cancel — the tool stays enabled rather than erroring.
- **Orphan sweep:** stale `ftth-*` tmux sessions are killed on first handler instantiation per process (Hub Client is the only creator, so anything unknown is orphaned).
- **Respects the per-agent-cwd redesign:** `shutdown()` never wipes the shared agent home or source repos — only the on-demand worktrees this agent created.

### Verification
- typecheck 13/13 · build 6/6 · biome clean
- full client suite **992 passed** — TUI unit 21/21 (tmux-session 5, transcript-tail 11, ask-user-degrader 5) + builtin-handlers 7/7 (now asserts claude-code-tui registers + returns a valid handler)

### Notes
- Draft pending QA review.
- The e2e framework + scenarios (`packages/e2e/`) and the `experiments/` PoC are the **next** slice (PR5), kept out of here to keep this reviewable.

[...]

PR #1586, opened 2026-07-10
feat(onboarding): use portable shell installer

bestony:
## Summary

- switch hosted prod/staging connection bootstrap to the channel-aware portable shell installer and remove the configurable npm bootstrap branch
- make the server-provided bootstrap authoritative across onboarding, Computers, and the no-computer New Agent flow; hosted login is now guarded by a successful installer download and exit status
- preserve fetch failures without non-POSIX `pipefail` by downloading to a temporary file, and remove that file when the guarded subshell exits
- update CLI non-installed upgrade guidance, public docs, release notifications, Web fixtures, and formal QA coverage while preserving dev source installs and legacy npm-mode self-update compatibility

## Recovery Fix

The hosted bootstrap now uses one POSIX `sh` success chain:

```sh
installer_tmp=$(mktemp "${TMPDIR:-/tmp}/first-tree-install.XXXXXX") && (trap 'rm -f "$installer_tmp"' 0; curl -fsSL <installer-url> -o "$installer_tmp" && sh "$installer_tmp" &&
~/.local/bin/<channel-bin> login <connect-code>)
```

Deterministic `/bin/sh` tests preinstall an observable stale shim and prove that:

- a fetch failure returns curl status `22`, cleans the temporary file, and never invokes the stale shim
- an installer failure returns installer status `9`, cleans the temporary file, and never invokes the stale shim
- a successful installer replaces the shim before the newly installed binary receives `login`

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] full `@first-tree/server` suite: 215 files / 2244 tests
- [x] full `@first-tree/web` suite: 163 files / 1310 tests
- [x] full `@first-tree/shared` suite: 49 files / 564 tests
- [x] focused bootstrap/API suite: 36 tests
- [x] affected Web onboarding/connection suite: 71 tests
- [x] CLI package metadata suite: 3 tests
- [x] documentation site build
- [x] `git diff --check`
- [ ] monorepo `pnpm test`

The monorepo command was run, but the untouched `@first-tree/skill-evals` package failed four tests before Turbo could finish every package: two empty-tree fixtures attempted a Git commit with nothing to commit, and two tests exceeded their 5-second timeout. The complete affected Server, Web, and Shared package suites pass independently.

[...]

Issue #718, opened 2026-06-01
apps/cli tree-skill-lib WHITEPAPER symlink test fails on local QA


yuezengwu:
## Summary

During PR #712 retest, the scoped client/runtime checks passed, but an unrelated `apps/cli` suite failure reproduced consistently in `tree-skill-lib.test.ts`.

## Repro

Worktree: PR #712 retest merge ref `096e4afc` (`f3e144f` into `main`).

```bash
pnpm --filter first-tree-dev test
pnpm --filter first-tree-dev exec vitest run src/__tests__/tree-skill-lib.test.ts --testNamePattern 'WHITEPAPER'
```

## Observed

The focused test fails consistently:

```text
FAIL src/__tests__/tree-skill-lib.test.ts > tree skill library > manages the WHITEPAPER symlink without replacing user files or directories
Error: EEXIST: file already exists, symlink 'old-target' -> '/var/folders/.../ft-tree-skill-*/WHITEPAPER.md'
  at apps/cli/src/__tests__/tree-skill-lib.test.ts:202
```

Relevant test lines:

```ts
rmSync(join(root, "WHITEPAPER.md"));
symlinkSync("old-target", join(root, "WHITEPAPER.md"));
```

Relevant implementation:

```ts
export function upsertWhitepaperFile(targetRoot: string): ManagedFileAction {
  const fullPath = join(targetRoot, WHITEPAPER_FILE);
  const existing = inspectSkillEntry(fullPath);
  // ...
  if (existing.kind === "symlink") {
    rmSync(fullPath, { force: true });
    symlinkSync(WHITEPAPER_SYMLINK_TARGET, fullPath);
    return "updated";
  }
}
```

## QA notes

- This appears unrelated to PR #712: PR #712 does not modify `apps/cli`.
- `git blame` for `apps/cli/src/__tests__/tree-skill-lib.test.ts:196-204` points to `6c2c1ea2` (`test: expand unit coverage (#665)`).
- Environment observed: macOS temp path under `/var/folders/...`, Node reported `v24.7.0` in the failed command output.

## Expected

The test should be stable locally and in CI, or the symlink removal/setup logic should be adjusted so the replacement step cannot hit `EEXIST` after `rmSync()`.

PR #1601, opened 2026-07-10
fix(onboarding): simplify portable bootstrap commands

bestony:
## Summary

- replace hosted portable bootstrap generation with two readable install/login commands
- keep channel, custom mirror, self-hosted server, and shell-quoting behavior intact
- update Web fixtures, user documentation, and the cross-surface QA case to the new contract

## Accepted tradeoff

The two lines intentionally provide no shell-level transactional protection. When pasted together, a failure on the install line does not automatically prevent the login line from running. The `curl | sh` pipeline also follows POSIX pipeline status semantics and may not preserve a `curl` failure status. This is an explicit usability tradeoff in this change.

## Verification

- `pnpm check`
- `pnpm typecheck`
- Server bootstrap/API focused tests: 45 passed
- Shared response-schema test: 1 passed
- Web onboarding, Computers, New Agent, preview, interaction, and SSR tests: 71 passed
- CLI package metadata tests: 3 passed
- documentation site production build
- prod and staging public `install.sh`: HTTP 200 and `sh -n` passed; no installer was executed
- `git diff --check`

Full `pnpm test` was also run. It stopped on two unrelated `@first-tree/skill-evals` failures:

- `src/core/__tests__/judge.test.ts` - `uses the selected tested-agent provider for standalone quality prerequisite gates`
- `src/suites/first-tree-seed/__tests__/periodic.test.ts` - `builds the real-source fixture from a run-scoped origin instead of the developer checkout`

Both fail while provisioning an empty Context Tree because `git commit -m chore: provision empty context tree` reports `nothing to commit`; this change does not touch `packages/skill-evals`.

## QA

This is a cross-surface Server -> Web -> public installer -> CLI -> daemon contract change. `portable-shell-onboarding` has been updated, and formal isolated QA is warranted but was not run automatically.

## Context Tree

The durable hosted bootstrap contract is updated in [first-tree-context#718](https://github.com/agent-team-foundation/first-tree-context/pull/718). The previously planned PR #712 had already merged, so the current-state rewrite is intentionally delivered as a fresh PR with fresh review requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,15 +127,18 @@ At the "connect a computer" step, setup gives you the channel-aware commands
 to install the CLI and link the machine. Hosted production uses:
 
 ```bash
-installer_tmp=$(mktemp "${TMPDIR:-/tmp}/first-tree-install.XXXXXX") && (trap 'rm -f "$installer_tmp"' 0; curl -fsSL https://download.first-tree.ai/releases/prod/install.sh -o "$installer_tmp" && sh "$installer_tmp" &&
-~/.local/bin/first-tree login <connect-code>)
+curl -fsSL https://download.first-tree.ai/releases/prod/install.sh | sh
+~/.local/bin/first-tree login <connect-code>
 ```
 
 Use the exact commands shown in the web console, especially for staging or a
 self-hosted deployment. The macOS/Linux installer bundles Node.js, so new
-users do not need to install Node separately. The command only logs in after
-the installer succeeds, and the explicit `~/.local/bin` path works immediately,
-even before the shell reloads its `PATH`.
+users do not need to install Node separately. The two lines are intentionally
+independent and do not provide shell-level transaction protection: when pasted
+together, an install-line failure does not automatically prevent the login line
+from running, and POSIX `sh` does not guarantee that `curl | sh` preserves a
+`curl` failure status. The explicit `~/.local/bin` path works immediately, even
+before the shell reloads its `PATH`.
 
 ## CLI
 
--- a/README_zh-CN.md
+++ b/README_zh-CN.md
@@ -83,14 +83,15 @@ First Tree 围绕 Context Tree 连接五个部分：
 命令。托管生产环境使用：
 
 ```bash
-installer_tmp=$(mktemp "${TMPDIR:-/tmp}/first-tree-install.XXXXXX") && (trap 'rm -f "$installer_tmp"' 0; curl -fsSL https://download.first-tree.ai/releases/prod/install.sh -o "$installer_tmp" && sh "$installer_tmp" &&
-~/.local/bin/first-tree login <connect-code>)
+curl -fsSL https://download.first-tree.ai/releases/prod/install.sh | sh
+~/.local/bin/first-tree login <connect-code>
 ```
 
 请以 Web 控制台显示的命令为准，尤其是在使用 staging 或自托管部署时。
-macOS/Linux 安装脚本已内置 Node.js，无需另行安装。命令仅在安装成功后
-执行登录；登录命令显式使用 `~/.local/bin`，因此即使当前 shell 尚未刷新
-`PATH` 也能立即执行。
+macOS/Linux 安装脚本已内置 Node.js，无需另行安装。为保持易读，这两行命令
+相互独立，不提供 shell 级事务保护：整段粘贴时，安装行失败不会自动阻止登录
+行运行，POSIX `sh` 也不保证 `curl | sh` 保留 `curl` 的失败状态。登录命令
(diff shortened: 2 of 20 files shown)


baixiaohang (dismissed):
## Review by baixiaohang-code-reviewer - PR #1601

- Rationale: the readable two-line bootstrap reintroduces a recovery path where download or installation failure can fall through to a stale local binary and make the computer appear connected without repairing the portable installation.

**Risk level**: B-low
- Path baseline: changes `packages/server/**` outside the B-high service/DB boundary -> B-low
- Semantic lift: none

**PR summary**
- Problem: users copying the hosted computer bootstrap currently receive a long guarded shell expression that is difficult to read.
- Approach: replace it with an independent `curl | sh` line and full-path login line across the server response, Web fixtures, public documentation, and cross-surface QA, explicitly accepting that installer failure does not gate login.
- Impacted modules: Server connect-token response, shared response contract, Web onboarding/connection surfaces, public docs, and portable-shell QA.

**Review findings**

❌ 1. Preserve the installer-success boundary before invoking login. [R4 / `packages/server/src/api/me.ts:111`]

Joining the installer pipeline and login with a bare newline means a copied bootstrap continues into `~/.local/bin/first-tree* login` after an installer failure. On a disconnected machine with a pre-existing stale shim, that old binary can reconnect the client and make Web onboarding report success even though download, verification, installation, or channel switching never completed. The pipeline also starts executing installer input before `curl` has completed and, under POSIX pipeline status semantics, can mask the fetch failure.

This is the same recovery blocker fixed immediately before this PR in #1586. The active Context Tree now requires hosted bootstrap/recovery to stop before login whenever portable download, verification, installation, or channel switching fails. This PR reverses that invariant while also deleting the deterministic stale-shim fetch/installer failure tests and removing those recovery scenarios from `portable-shell-onboarding`; the promised follow-up tree PR is not open or cross-linked, and would not make the current unsafe behavior acceptable by itself.

Please keep the readable surface while retaining a POSIX-safe success guard, and restore deterministic plus cross-surface recovery coverage proving that a failed fetch/installer cannot invoke or reconnect through an existing shim.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review

**Goal and core change:** this PR replaces the server-authored guarded portable bootstrap with two readable commands (`curl | sh`, then the full-path channel login) and propagates that contract through the shared schema, Web fixtures, public docs, and the cross-surface QA case. It does not change a database schema or core data structure.

### Blocker: preserve installation success as a hard precondition for login

At `packages/server/src/api/me.ts:111`, joining the installer pipeline and login with a bare newline lets the login line run after download or installation failure. On a recovery machine with an existing stale `~/.local/bin/first-tree*` shim, that old binary can consume the fresh connect code and reconnect the computer even though artifact download, verification, installation, or channel switching never completed. POSIX pipeline status makes this worse: a failed producer can still leave `curl | sh` with `sh`'s successful status.

This conflicts with the active portable-distribution and channel bootstrap contracts, which require bootstrap to stop before any pre-existing channel binary is invoked. The change also deletes the deterministic fetch/installer failure tests and removes the corresponding stale-install recovery scenarios from `portable-shell-onboarding`, so the exact regression is no longer guarded.

Please keep the readable surface while retaining a POSIX-safe success guard between fetch/install and login, then restore deterministic and cross-surface coverage proving that neither a failed fetch nor a failed installer can invoke or reconnect through an existing shim. The durable Context Tree contract also needs its source-driven companion PR opened and cross-linked with this contract change.

**Action:** requesting changes.

baixiaohang:
Acknowledged: the linked Context Tree PR #718 is now open and cross-linked, so the companion-PR process gap from the existing review is resolved.

The source head is still `5187cc65`, however, and #718 currently proposes accepting the same stale-binary fallthrough rather than changing the implementation or restoring the removed recovery coverage. The core request-changes finding therefore remains: preserve installation success as the hard precondition for login, with deterministic and cross-surface proof for fetch/installer failure on a pre-existing shim.

I will re-review that finding when the source head changes; no need to repeat the full review thread in the meantime.

bestony:
The failure-through behavior is intentional and explicitly accepted for this change. Preserving a shell success guard would keep the prior contract and would not satisfy the requested two-line bootstrap.

The durable contract that the reviews referenced has now been rewritten in [first-tree-context#718](https://github.com/agent-team-foundation/first-tree-context/pull/718), with fresh review requested from the affected node owners. That PR records the Server-authored portable/channel boundaries, verbatim Web copy, token-free download plane, independent install/login commands, and the lack of shell-level transaction protection. This source PR is cross-linked to it, and CI is green.

Please re-review this PR against the intentionally updated contract and accepted usability tradeoff.

yuezengwu (approved):

PR #1605, opened 2026-07-11
refactor(web): chat list query foundation — useInfiniteQuery + error states

Gandy2025:
## Summary

PR 1 of 5 in the Workspace Chat List refactor — the **query foundation**. It replaces the
hand-rolled `useQuery + extraPages + manual cursor` pagination in the left-rail chat list
with `useInfiniteQuery`, and adds first-class error/retry states. Web-only; **no server or
API-contract change**. All existing behavior is preserved except the new error states.

This de-risks the later PRs (Pin, `activity_at` ordering, filter redesign) by removing the
pagination race that any further work would otherwise inherit.

## Why

The old rail managed pages by hand (`extraPages` appended outside React Query), which caused
two real bugs and made the component hard to extend:

- **P0 — stale pages pollute a new filter / duplicate rows.** A "Load more" response that
  resolved after the user changed filters was appended to the new list; a background refetch
  could also surface a chat on two pages at once (duplicate React keys).
- **P1 — a failed first page masqueraded as empty.** The query never read `isError`, so a
  network/API failure fell through to "No conversations yet." instead of an error.

## Changes

- `conversations/index.tsx`: `useQuery` + `extraPages`/`loadingMore`/`moreError`/`nextCursor`
  state + two effects + `handleLoadMore` → a single `useInfiniteQuery`
  (`getNextPageParam` = `nextCursor ?? undefined`). Rows are flattened across pages and
  **de-duplicated by `chatId`** (first occurrence wins).
- Error states: a first-page error now renders "Couldn't load conversations." + **Retry**
  (`refetch`); a failed "Load more" renders an inline **Retry** (`fetchNextPage`). The empty
  state is now gated on `!isError`, so a failed load can never read as "no chats".
- Load more is driven by `hasNextPage` / `fetchNextPage` / `isFetchingNextPage` /
  `isFetchNextPageError`.
- `api/client.ts` + `api/me-chats.ts`: thread an optional `AbortSignal` so React Query can
  cancel the in-flight request when the filter/cursor changes.
- Removed the per-page `fetchedAt` busy-staleness sieve: an infinite query refetches **every
  loaded page** on the 30s `refetchInterval`, so each row's `busyAgentIds` self-heals on the
  same cadence (≤ `RUNTIME_STALE_MS` + one refetch), with no page bypassing it.
- The `["me","chats", …]` query key is unchanged, so the WebSocket prefix-invalidation and
  the throttled invalidator keep working untouched.

## Behavior preserved

Grouping, "Needs attention" (failed > request, page-local for now), the All/Unread/Watching
triad, Status/Source filters, chips, the unread count, row states, and the selected-row
treatment are all unchanged. Confirmed via the existing DOM tests + a real-browser pass.

## Testing

- **TDD**: added failing DOM tests first (first-page error + retry recovery, cross-page
  dedup), then implemented.
- `pnpm --filter @first-tree/web test` — **1331 passed**.
- `pnpm check` (Biome) + `pnpm --filter @first-tree/web typecheck` (tsc + design-token lint) — clean.
- **browse·QA** on `/preview/conversation-list` (renders the real component through
  `useInfiniteQuery`): rows/groups/attention/selection render correctly, no console errors.

## Scope notes

Deferred to later PRs (unchanged here, by design): canonical `FilterSpec` refactor, the
"Outside current filters" hint, Status single-select, server aggregate unread count, and the
cross-page attention/pinned projection. Part of the Chat List refactor whose durable filter
semantics are captured in first-tree-context#721.

### DIFF
--- a/packages/web/src/api/__tests__/api-wrappers.test.ts
+++ b/packages/web/src/api/__tests__/api-wrappers.test.ts
@@ -288,6 +288,7 @@ describe("api wrapper paths", () => {
     });
     expect(apiMock.get).toHaveBeenCalledWith(
       "/orgs/current/chats?limit=10&cursor=next&filter=unread&engagement=active&origin=manual%2Cgithub%2Cagent&with=agent-1%2Cagent-2&watching=1",
+      undefined,
     );
     expect(apiMock.post).toHaveBeenCalledWith("/orgs/current/chats", {
       mode: "task",
--- a/packages/web/src/api/client.ts
+++ b/packages/web/src/api/client.ts
@@ -168,8 +168,11 @@ async function tryRefresh(refreshToken: string): Promise<StoredTokens | null> {
   return refreshPromise;
 }
 
-async function request<T>(path: string, options?: { method?: string; body?: unknown }): Promise<T> {
-  const { method = "GET", body } = options ?? {};
+async function request<T>(
+  path: string,
+  options?: { method?: string; body?: unknown; signal?: AbortSignal },
+): Promise<T> {
+  const { method = "GET", body, signal } = options ?? {};
 
   // No path rewriting here — callers prefix org-scoped paths with `withOrg` /
   // `withOrgAt` before passing in; everything else (`/me/...`, `/auth/...`,
@@ -184,6 +187,7 @@ async function request<T>(path: string, options?: { method?: string; body?: unkn
       method,
       headers,
       body: body !== undefined ? JSON.stringify(body) : undefined,
+      signal,
     });
   };
 
@@ -274,7 +278,7 @@ export async function apiFetchRaw(
 }
 
 export const api = {
-  get: <T>(path: string) => request<T>(path),
+  get: <T>(path: string, options?: { signal?: AbortSignal }) => request<T>(path, options),
   post: <T>(path: string, body?: unknown) => request<T>(path, { method: "POST", body }),
   patch: <T>(path: string, body?: unknown) => request<T>(path, { method: "PATCH", body }),
   put: <T>(path: string, body?: unknown) => request<T>(path, { method: "PUT", body }),
--- a/packages/web/src/api/me-chats.ts
+++ b/packages/web/src/api/me-chats.ts
@@ -25,7 +25,7 @@ export type ListMeChatsParams = Partial<
   Pick<ListMeChatsQuery, "cursor" | "limit" | "filter" | "engagement" | "origin" | "with" | "watching">
 >;
 
-export function listMeChats(params?: ListMeChatsParams): Promise<ListMeChatsResponse> {
+export function listMeChats(params?: ListMeChatsParams, opts?: { signal?: AbortSignal }): Promise<ListMeChatsResponse> {
   const qs = new URLSearchParams();
   if (params?.limit !== undefined) qs.set("limit", String(params.limit));
   if (params?.cursor) qs.set("cursor", params.cursor);
@@ -39,7 +39,9 @@ export function listMeChats(params?: ListMeChatsParams): Promise<ListMeChatsResp
   if (params?.with && params.with.length > 0) qs.set("with", params.with.join(","));
   if (params?.watching) qs.set("watching", "1");
   const query = qs.toString();
-  return api.get<ListMeChatsResponse>(withOrg(`/chats${query ? `?${query}` : ""}`));
+  // `opts.signal` lets React Query cancel the in-flight request when the
+  // filter/cursor changes, so a superseded page never lands.
+  return api.get<ListMeChatsResponse>(withOrg(`/chats${query ? `?${query}` : ""}`), opts);
 }
 
 export function listMeChatSourceCounts(params?: { engagement?: ChatEngagementView }): Promise<MeChatSourceCounts> {
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -598,8 +598,14 @@ function createClient(): QueryClient {
       topic: "Launch planning",
     },
   ]);
-  queryClient.setQueryData(["me", "chats", "all", "active", false, null, null], {
-    rows: [
+  // The rail reads via `useInfiniteQuery`, so seed the `InfiniteData` shape.
+  const meInfinite = (rows: ReturnType<typeof chatRow>[]) => ({
+    pages: [{ rows, nextCursor: null }],
+    pageParams: [undefined],
+  });
+  queryClient.setQueryData(
+    ["me", "chats", "all", "active", false, null, null],
+    meInfinite([
       chatRow(),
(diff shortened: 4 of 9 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1605

**Recommendation**: approve
- Rationale: The refactor removes the cross-filter pagination race, preserves the existing rail semantics, and adds correctly separated initial-load and next-page recovery states without introducing a server/API contract change.

**Risk level**: B-low
- Path baseline: only `packages/web/**` -> C
- Semantic lift: touches the React Query data/cache layer -> B-low

**PR summary**
- Problem: Workspace users could see stale or duplicate conversations after loading more and changing filters, while an initial network failure was misrepresented as an empty chat list.
- Approach: Move all loaded pages under one `useInfiniteQuery` cache, cancel superseded requests, de-duplicate by `chatId`, and provide distinct retry paths for initial and next-page failures. Existing grouping and filter behavior remains unchanged; the paired filter-contract work is intentionally deferred to later PRs.
- Impacted modules: web API client/wrappers, Workspace conversation rail, mark-read cache patching, preview and DOM test fixtures.

**Review findings**
- ✅ Infinite and bare `ListMeChatsResponse` cache consumers are handled explicitly, including the mark-read patch shared with palette/mobile queries.
- ✅ Initial-load errors, background-refetch errors, and next-page errors retain distinct UI behavior and recovery paths.
- ✅ The unchanged `["me", "chats", …]` prefix preserves existing WebSocket and mutation invalidation behavior.

**Action taken**
- Approved.

yuezengwu (changes requested):
## Review

The implementation goal is clear: move the Workspace rail from manually managed follow-up pages to one `useInfiniteQuery`, isolate filter keys, cancel superseded requests, de-duplicate rows, and distinguish initial-load from next-page failures. The infinite/bare cache-shape patching and the error/retry paths otherwise look sound.

### Blocking finding

- **Restore a successful-refetch clock for time-derived row UI** — `packages/web/src/pages/workspace/conversations/index.tsx:180-215`

  `allRows` now depends only on `data`, and the hook no longer reads `dataUpdatedAt`. TanStack Query 5.95 structurally shares identical successful responses and only notifies a component when a property it tracks changes. For a quiet list, an interval refetch can therefore make the request while preserving `data` identity and causing no render at all.

  I reproduced this against the locked dependency: two identical infinite-query fetches made two network calls, but a component reading the same fields as this hook stayed at the same render count and received the same `data` object. `dataUpdatedAt` was the only changed result field.

  As a result, both `formatRowTime()` (`now` / `5m` / `3h`) and `groupRows(..., new Date())` (`Today` / `Yesterday`) can remain stale indefinitely when the server payload is unchanged, despite the comment promising recomputation after the 30s refetch. The previous implementation deliberately included `dataUpdatedAt` as its time tick. Please restore an equivalent successful-refetch dependency and add a regression test that advances time while the response remains structurally identical.

### Verification

- `pnpm --filter @first-tree/web test` — 1332 passed
- Biome check on all nine changed files — passed

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1605

- Rationale: The infinite-query migration removes the successful-refetch render clock that time-derived conversation-row UI relies on, so quiet lists can display stale relative times and day buckets indefinitely.

**Risk level**: B-low
- Path baseline: only `packages/web/**` -> C
- Semantic lift: changes React Query data/cache behavior -> B-low

**PR summary**
- Problem: The PR correctly addresses stale manual pages, duplicate rows, and missing load-error states in the Workspace rail.
- Approach: Consolidate pages under `useInfiniteQuery`, cancel superseded requests, de-duplicate rows, and distinguish initial and next-page recovery. Most of that implementation remains sound, but the refactor drops the render signal used by time-derived UI when responses are structurally identical.
- Impacted modules: Workspace conversation rail and its React Query cache behavior.

**Review findings**
❌ 1. Restore a successful-refetch clock for time-derived row UI and cover it with a regression test. The previous hook tracked `dataUpdatedAt`; the new hook tracks `data` but TanStack Query 5.95 structurally shares identical results and only notifies for tracked result properties. On a quiet list, the interval request can succeed while `data` retains identity and the component does not render. `formatRowTime()` and `groupRows(..., new Date())` then remain stale even though the comments promise a 30-second refresh cadence. Track an equivalent successful-update field and ensure the relevant memo/render path depends on it; add a test that advances time across a successful identical refetch. [R4 / packages/web/src/pages/workspace/conversations/index.tsx:121]

**Action taken**
- Submitted request changes, superseding my earlier approval after the new reproduction evidence exposed this regression.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1605

**Recommendation**: approve
- Rationale: The follow-up restores the successful-refetch render clock and adds a regression test that exercises structurally identical infinite-query responses, resolving the only blocker.

**Risk level**: B-low
- Path baseline: only `packages/web/**` -> C
- Semantic lift: touches React Query data/cache behavior -> B-low

**PR summary**
- Problem: The Workspace rail needed race-free pagination and explicit recovery states without freezing clock-derived row UI on quiet lists.
- Approach: Keep the `useInfiniteQuery` migration, cancellation, de-duplication, and error states; additionally track `dataUpdatedAt` so every successful refetch re-renders relative timestamps and recomputes recency buckets even when structural sharing preserves `data` identity.
- Impacted modules: Workspace conversation rail, API request cancellation, shared chat-list cache patching, and DOM fixtures/tests.

**Review findings**
- ✅ The prior blocker is resolved: `dataUpdatedAt` is tracked and drives the recency-bucket memo with a fresh `now`.
- ✅ The regression test advances `Date` across a successful structurally identical refetch and verifies the rendered relative time changes from `2m` to `1h`.
- ✅ The original infinite/bare cache-shape, cancellation, de-duplication, and error/retry paths remain intact.

**Action taken**
- Approved head `02733db1`.

yuezengwu (approved):
## Re-review

The goal and core shape remain sound: consolidate the Workspace rail under `useInfiniteQuery`, isolate/cancel paginated requests, de-duplicate cross-page rows, and expose distinct initial/next-page recovery states without changing the server contract.

The blocking finding is resolved on `02733db1`:

- `dataUpdatedAt` is now tracked, so an identical successful refetch notifies the component.
- The recency-bucket memo depends on that successful-refetch clock and captures one fresh `now`, while the render refreshes `formatRowTime()` as well.
- The new DOM regression test returns structurally identical payloads, advances `Date`, refetches, and verifies `2m` becomes `1h`, directly exercising the prior failure mode.

No new findings.

Verification:

- conversation-list DOM tests: 9 passed
- Web typecheck + design-token lint: passed
- Biome on the follow-up files: passed
- Full Web run reached 1332 passing tests with one unrelated flaky failure in the unchanged mobile density test; isolated reruns of that file were nondeterministic (2 pass / 1 fail), so it is not attributable to this patch.

Approved without waiting on CI, per review policy.

PR #1606, opened 2026-07-11
feat(server): chat list data foundation — pinned_at, activity_at, pin mutation

Gandy2025:
## Summary

PR 2 of 5 in the Workspace Chat List refactor — the **data + API foundation** for
per-user Pin and activity-based ordering. Server + shared only; **no ordering change
yet** (the list still sorts by `last_message_at`; activity ordering + the Pin UI land in
later PRs). This gives PR 3 (server ordering) and PR 4 (rail UI) the columns, mutation,
and projection they build on.

## Changes

**Schema + migration (`0076_lazy_risque.sql`)**
- `chat_user_state.pinned_at` (nullable `timestamptz`) + partial index
  `WHERE pinned_at IS NOT NULL`. NULL = not pinned, so no backfill.
- `chats.activity_at` (`NOT NULL DEFAULT now()`) + `(organization_id, activity_at DESC)`
  index. Existing rows are backfilled to
  `GREATEST(COALESCE(last_message_at, created_at), COALESCE(description_updated_at, created_at), created_at)`
  (the backfill runs before the index so it builds on final values). Migrations auto-run
  on boot.

**`activity_at` semantics** — advances only on real work, never on transient signals:
- **new message** → `chat-projection.ts applyAfterFanOut` (the single writer of `last_message_at`)
- **genuine description change** → `chat.ts updateChatMetadata` (only when the text actually changes)
- **not** bumped by topic rename, read/unread, pin, archive, participant change, or runtime busy/idle.
- New chats get `activity_at = created_at` via the column default.

**Pin mutation**
- `pinMeChat(db, chatId, agentId, pinned)` — idempotent UPSERT on the composite PK, mirroring
  `setChatEngagement`. Pin is private per-user state; a write only ever touches the caller's
  own `(chat_id, agent_id)` row.
- `POST /chats/:chatId/pin` (`{ pinned: boolean }`), guarded by `requireChatAccess` — a watcher
  can pin, matching read/unread.

**Shared + projection**
- `MeChatRow` gains `pinnedAt` / `activityAt` (`.default(null)` version-skew, like siblings);
  `listMeChats` projects `cus.pinned_at` + `c.activity_at`. The `["me","chats"]` query and its
  `ORDER BY last_message_at DESC` are unchanged.

## Testing

- New `me-chat-pin-activity.test.ts` (real Postgres via testcontainers): pin/unpin idempotent +
  **user-isolated**; `activity_at` initialised at creation, bumped by a message and a real
  description change, **not** bumped by a topic rename or a pin.
- `pnpm --filter @first-tree/server test` — 2243 pass (1 unrelated pre-existing flaky
  `context-tree-snapshot` test that passes in isolation). Web + shared suites pass. `pnpm check`,
  server/web `typecheck` clean. The shared `MeChatRow` addition rippled to ~10 web test fixtures,
  updated in this PR.

## Migration note

`0076` runs on boot in a single transaction: `ADD COLUMN` (a fast metadata-only default —
`now()` is stable, so no table rewrite) + a full-table `activity_at` backfill + two
non-concurrent index builds. At the current `chats` size this is sub-second, but it holds an
`ACCESS EXCLUSIVE` lock on `chats` for the transaction, inside the ~20s boot-migration budget.
Worth revisiting (batched backfill / `CREATE INDEX CONCURRENTLY`) before `chats` grows large
enough that the backfill approaches that budget; the single transaction keeps it atomic, so a
failure rolls back cleanly with no half-applied state. The DDL is `drizzle-kit`-generated; the
data backfill was hand-added (drizzle can't generate a backfill) and is safe — the snapshot
tracks only DDL and the migration has never been applied.

## Scope notes

Deferred by design: activity-based ORDER BY + cursor, the cross-page attention/pinned priority
projection, and aggregate counts (PR 3); the Pin UI + web api wrapper (PR 4). The durable
`activity_at` + user-Pin semantics will be reconciled into the reviewer-owned
`system/cloud/chat/workspace-conversations.md` tree node alongside PR 3/4 (its current text says
only attention pins the list).

### DIFF
--- a/packages/server/drizzle/0076_lazy_risque.sql
+++ b/packages/server/drizzle/0076_lazy_risque.sql
@@ -1,0 +1,23 @@
+-- Chat-list refactor — data foundation.
+--   1. chat_user_state: add per-user `pinned_at` (nullable) + partial index.
+--      NULL = not pinned, so no backfill is needed for existing rows.
+--   2. chats: add `activity_at` (the conversation-list recency sort key) +
+--      backfill + (organization_id, activity_at DESC) index.
+--
+-- `activity_at` advances only on real work (a new message or a genuine
+-- description change), never on rename/read/pin/archive/participant/busy. New
+-- rows default to now() (= creation time); existing rows are backfilled to
+-- max(last_message_at, description_updated_at, created_at). The backfill runs
+-- before the activity index so the index builds on final values.
+ALTER TABLE "chat_user_state" ADD COLUMN "pinned_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "chats" ADD COLUMN "activity_at" timestamp with time zone DEFAULT now() NOT NULL;--> statement-breakpoint
+-- The ADD COLUMN default set every existing row to the migration-time now();
+-- reset them to real activity. last_message_at / description_updated_at are
+-- nullable, so COALESCE each to created_at before GREATEST.
+UPDATE "chats" SET "activity_at" = GREATEST(
+  COALESCE("last_message_at", "created_at"),
+  COALESCE("description_updated_at", "created_at"),
+  "created_at"
+);--> statement-breakpoint
+CREATE INDEX "idx_user_state_pinned" ON "chat_user_state" USING btree ("agent_id") WHERE pinned_at IS NOT NULL;--> statement-breakpoint
+CREATE INDEX "idx_chats_org_activity" ON "chats" USING btree ("organization_id","activity_at" desc);
--- a/packages/server/drizzle/LATEST
+++ b/packages/server/drizzle/LATEST
@@ -1,1 +1,1 @@
-0075_right_jigsaw
+0076_lazy_risque
--- a/packages/server/drizzle/meta/0076_snapshot.json
+++ b/packages/server/drizzle/meta/0076_snapshot.json
@@ -1,0 +1,4631 @@
+{
+  "id": "fc5689ec-740f-4747-9101-23254feee77a",
+  "prevId": "e15a50dd-d482-4179-aa69-0434aeba350a",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent_chat_sessions": {
+      "name": "agent_chat_sessions",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "chat_id": {
+          "name": "chat_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "state": {
+          "name": "state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "runtime_state": {
+          "name": "runtime_state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'idle'"
+        },
+        "runtime_state_at": {
+          "name": "runtime_state_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "idx_agent_chat_sessions_chat_agent": {
+          "name": "idx_agent_chat_sessions_chat_agent",
+          "columns": [
+            {
+              "expression": "chat_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "agent_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "agent_chat_sessions_agent_id_agents_uuid_fk": {
+          "name": "agent_chat_sessions_agent_id_agents_uuid_fk",
+          "tableFrom": "agent_chat_sessions",
+          "tableTo": "agents",
+          "columnsFrom": [
+            "agent_id"
+          ],
+          "columnsTo": [
+            "uuid"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "agent_chat_sessions_chat_id_chats_id_fk": {
+          "name": "agent_chat_sessions_chat_id_chats_id_fk",
+          "tableFrom": "agent_chat_sessions",
+          "tableTo": "chats",
+          "columnsFrom": [
+            "chat_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {
+        "agent_chat_sessions_agent_id_chat_id_pk": {
+          "name": "agent_chat_sessions_agent_id_chat_id_pk",
+          "columns": [
+            "agent_id",
+            "chat_id"
+          ]
+        }
+      },
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.agent_configs": {
+      "name": "agent_configs",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "version": {
+          "name": "version",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "default": 1
+        },
+        "payload": {
+          "name": "payload",
+          "type": "jsonb",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "updated_by": {
+          "name": "updated_by",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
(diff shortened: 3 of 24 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1606

- Rationale: the new pin mutation is documented and exposed as idempotent, but retrying the same pin request changes durable ordering state.

**Risk level**: B-high
- Path baseline: includes a DB schema/migration and `packages/server/src/services/chat*` changes.

**Review findings**

❌ 1. Repeating `POST /chats/:chatId/pin` with `{ pinned: true }` rewrites `pinned_at` on every call. Because this timestamp is explicitly the stable within-pinned-group sort key, an HTTP retry or double click silently moves an already-pinned chat and makes the operation non-idempotent despite the route/service contract. The added test only checks that the UPSERT leaves one row; it never checks that the timestamp remains unchanged. Please preserve the existing non-null timestamp when the requested state is already pinned, return the persisted value, and add a regression assertion that two consecutive pin calls produce the same `pinnedAt`. [R4 / `packages/server/src/services/me-chat.ts:148`]

✅ The `activity_at` writer audit is otherwise coherent: message projection and genuine description changes advance it, while topic/pin/transient state paths remain excluded.

**Action taken**
- Submitted request changes.

I did not run tests; this review is based on the diff, surrounding call sites, and the author-provided test results.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1606

**Recommendation: request changes**

This PR establishes the server/shared foundation for private per-user pins and a durable work-activity recency key, while intentionally leaving list ordering and UI behavior unchanged. The schema/projection split is coherent, but two correctness contracts need to be closed before this becomes the foundation for later ordering work.

### Findings

1. **Blocker — `activity_at` can move backwards under concurrent qualifying writes.**

   `applyAfterFanOut` assigns `activityAt: ts` directly, and `updateChatMetadata` assigns `activityAt: now` directly. The message transaction does not acquire the chat-row lock until after the message and fan-out work, so two sends—or a send and a description update—can acquire that lock and commit in the opposite order from their event timestamps. An older transaction that reaches the update last then overwrites a newer `activity_at`, contradicting the documented “advances”/latest-activity contract and eventually misordering the rail.

   Please make every qualifying writer monotonic at the database update (for example, preserve the greater of the stored value and the event timestamp) and add an out-of-order regression assertion showing that applying an older qualifying timestamp after a newer one cannot move `activity_at` backwards. (`packages/server/src/services/chat-projection.ts:137`, `packages/server/src/services/chat.ts:703`)

2. **Blocker — repeated `pin=true` changes the durable sort key.**

   I confirm the blocker already filed by baixiaohang: the operation is documented as idempotent, but a retry replaces `pinned_at` and moves the chat within the pinned group. Preserve an existing non-null pin timestamp, return the persisted value, and assert that consecutive identical pin requests return/store the same timestamp. (`packages/server/src/services/me-chat.ts:148-156`)

3. **Required route coverage is missing for the new Class C endpoint.**

   `docs/development/http-path-conventions.md` requires multi-org coverage for every new Class B/C route. The new test calls `pinMeChat` directly, so it does not exercise `POST /chats/:chatId/pin`, body validation, watcher access, caller identity selection, or anti-enumeration for an outsider/different-org caller. Please add HTTP-level coverage analogous to `chat-engagement.test.ts`, including success, invalid payload, watcher access, and a 404 for a caller without chat access.

### Database review note

The migration is internally consistent: nullable `pinned_at` needs no backfill; `activity_at` is backfilled before its index; snapshot/journal/LATEST align. It still performs a full-table update and two non-concurrent index builds while the migration transaction holds the `chats` lock. The stated current-size/sub-second evidence makes that plausible, but the deployment owner should explicitly verify the production row count and lock timing before merge/deploy; this is a human rollout check, not an additional code finding.

I did not run the suite locally because this fresh review worktree has no installed dependencies. I checked the full diff, migration metadata, relevant call sites and repository HTTP conventions; `git diff --check` is clean, and the currently reported lint/type/migration checks are green.

yuezengwu (commented):
Incremental re-review of `7ec9247d`:

- ✅ The pin idempotency finding is resolved. The conflict update now preserves an existing non-null `pinned_at`, a fresh pin after unpin gets a new anchor, the response comes from `RETURNING`, and the regression test compares the repeated result.
- ❌ The `activity_at` monotonicity finding remains: both qualifying writers still assign their event timestamp directly, so a later-committing older event can move the recency key backwards.
- ❌ The new Class C pin endpoint still has no HTTP/multi-org route coverage; the added test remains service-level only.

The request-changes state therefore remains for findings 1 and 3 from my review. I did not wait for or use CI status for this conclusion.

baixiaohang (commented):
## Follow-up review at `7ec9247d`

**Recommendation remains: request changes.**

The original pin-idempotency blocker is resolved: the UPSERT now preserves an existing non-null `pinned_at`, returns the persisted value, and the regression test asserts that a repeated pin keeps the same timestamp.

I independently confirmed the two remaining findings in yzw-codex’s current-head review:

1. `activity_at` is still assigned directly by both qualifying writers. A transaction carrying an older event timestamp can reach the chat-row update after a newer transaction commits and overwrite the newer value, so the field is not monotonic under out-of-order/concurrent writes. This needs a database-side max/preserve-newer guard plus an out-of-order regression test.
2. The new Class C pin route has no HTTP-level coverage. The repository’s HTTP conventions explicitly require multi-org coverage for new Class B/C routes, and the service-only test does not exercise payload validation, watcher access, caller selection, or outsider anti-enumeration.

This supersedes my earlier pass note that the `activity_at` writer audit was coherent; the set of qualifying writers is complete, but the concurrency behavior is not yet safe. I am not duplicating the full findings because the existing current-head review already states the required fixes precisely.

I did not run tests, per review policy.

baixiaohang (changes requested):
## Incremental review at `6d85a79d`

**Recommendation: request changes**

The product fixes are sound:

- ✅ Both qualifying `activity_at` writers now use a database-side `GREATEST(current, event_ts)`, so a later-committing older event cannot move the recency key backwards; the out-of-order regression covers the core invariant.
- ✅ The pin idempotency fix remains intact.
- ✅ HTTP coverage now exercises success, payload validation, watcher access, and same-org non-member anti-enumeration.

One required gap remains:

❌ The test named “different org” does not create a different organization. `createTestAdmin()` calls `resolveDefaultOrgId()`, so `ownerSide` and `outsider` are both members of the same default organization. The assertion therefore covers a same-org caller who is not in the chat, not the repository convention’s required multi-org boundary. Please construct the target chat in a genuinely separate organization (with its own member/human agent and participant), call it using a user that has no membership in that organization, and assert the organization ids differ before asserting the 404. [`packages/server/src/__tests__/chat-pin-route.test.ts:94`]

**Action taken**
- Submitted request changes.

I did not run tests, per review policy.

yuezengwu (changes requested):
## Incremental review at `6d85a79d`

**Recommendation: request changes**

The two product-correctness fixes are resolved:

- ✅ Both qualifying `activity_at` writers use database-side `GREATEST(current, event_ts)`, and the out-of-order regression pins the monotonic invariant.
- ✅ Pin timestamp idempotency remains correct.
- ✅ The new HTTP suite covers pin/unpin success, invalid payload, watcher access, and same-org nonparticipant anti-enumeration.

One required test boundary is still missing:

- ❌ `chat-pin-route.test.ts:91-106` labels the outsider as “different org”, but `ownerSide`, `peer`, and `outsider` are all created through helpers that call `resolveDefaultOrgId()`. The test therefore exercises the same-org/no-chat-membership branch of `requireChatAccess`, not the cross-org `resolveCallerInOrg` branch required by `docs/development/http-path-conventions.md`. Please create the chat in a genuinely separate organization (with valid member/human-agent/participant records), call it with a user that has no membership there, assert the two organization ids differ, then assert 404. Keeping the current same-org outsider case as a separate assertion is useful.

No additional code findings in this increment. I did not wait for CI or run the suite locally; this conclusion comes from the incremental diff, helper implementation, and repository HTTP convention.

yuezengwu (approved):
## Final incremental review at `3b4848eb`

**Approved.** All requested changes are resolved:

- Pin is idempotent on both state and its durable `pinned_at` sort anchor, and returns the persisted value.
- Both qualifying `activity_at` writers are monotonic via database-side `GREATEST(current, event_ts)`, with an out-of-order regression.
- The Class C pin route now has HTTP coverage for success, invalid payload, watcher access, same-org nonparticipant anti-enumeration, and the genuine cross-org boundary.
- The cross-org case creates a distinct `orgB`, keeps the caller in the default org with no orgB membership, asserts the org ids differ, and therefore exercises `resolveCallerInOrg`'s 404 path. The lightweight foreign-org participant fixture matches the repository's existing cross-org pollution-test convention.

No remaining code findings. `git diff --check` is clean. I did not run the suite locally and did not wait for CI.

Operational note remains: because migration `0076` performs a full-table backfill plus non-concurrent index builds in the boot transaction, the deployment owner should still verify the production `chats` row count and lock duration before rollout; this is a human deployment check, not a code-review blocker.

baixiaohang (approved):
## Final review by baixiaohang-code-reviewer

**Recommendation: approve**

All requested changes are resolved at `3b4848eb`:

- Pin state and its `pinned_at` ordering anchor are idempotent and user-isolated.
- Both qualifying `activity_at` writers are monotonic via database-side `GREATEST`, with an out-of-order regression.
- The new Class C pin route covers success, invalid payload, watcher access, same-org nonparticipant anti-enumeration, and a genuine cross-org 404 boundary.
- Migration metadata is internally consistent; list projection adds no extra DB or HTTP round trips.

Operational note: migration `0076` performs a full-table backfill and non-concurrent index builds in the boot transaction. The deployment owner should verify the production `chats` row count and lock duration before rollout; this is not a code-review blocker.

Approved with explicit authorization from baixiaohang. I did not run tests, per review policy.

Issue #1732, opened 2026-07-12
[Bug] Codex usage-limit turn failure writes no slot-log line; auth-epoch restarts deferred indefinitely by wedged turns


L42y:
## Summary

When a Codex-runtime turn fails on a provider usage limit, the daemon posts the "⚠️ My runtime has reached its usage limit…" chat message but writes **no corresponding line to the slot log** (`client.log`). Log-tailing automation (e.g. account-failover watchers that cool an exhausted provider account and switch to a backup) therefore never sees the event. Claude-runtime provider failures DO log a detectable line, so this is a Codex-path gap.

## Surface Area

Agent runtime / daemon slot logging (codex runtime).

## Environment

- version: 0.5.14-staging.819.1
- runtime: codex
- operating system: Linux 6.18.5+deb13-cloud-amd64 x86_64
## Reproduction

1. Run a codex-runtime agent under the daemon with slot logs going to `client.log`.
2. Let the agent's ChatGPT account hit its usage/session limit.
3. Send the agent a message so it attempts a turn.
4. Observe the chat receives "⚠️ My runtime has reached its usage limit, so I couldn't process the message you just sent." while `client.log` shows only `codex app-server turn started` followed by `session resumed` — no failure/limit line.

## Expected Behavior

The usage-limit turn failure is written to the slot log with the agent metadata, the way Claude SDK provider failures are (e.g. `[slot] Claude SDK provider failure (provider_capacity/provider_rate_limited): rate_limit …agentName=… `), so external log watchers can react — ideally with a stable machine-parseable shape and a `provider_usage_limit`-style tag.

## Actual Behavior

Observed 2026-07-12 10:19:53–10:19:55 UTC:

```text
10:19:53 INFO [slot] codex app-server turn started turnId=019f55d7-2db6-78f3-b4b6-db2a06eabbd0 accepted=1  agentName=orchestrator …
10:19:55 INFO [slot] session resumed  agentName=orchestrator …
```

Chat message posted at 10:19:54: "⚠️ My runtime has reached its usage limit, so I couldn't process the message you just sent. Please resend it once the limit resets."

No line containing the limit text (nothing matching `could not run this turn`, `provider_usage_limit`, `usage limit`, `rate_limit`) appears anywhere in the log for this event.

## First Tree Context

- Chat ID: 88a05dbf-06d1-41d9-96cf-d183a64a2b0e
- Reported by (user): l42y
[...]

PR #1733, opened 2026-07-12
fix: log codex app-server usage-limit turn failures to the slot log

L42y:
## Summary

When a codex **app-server** turn fails on a provider usage limit, the handler emits the error event and posts the "⚠️ My runtime has reached its usage limit…" chat notice — but writes **no slot-log line**. Every sibling branch in `settleTurn` logs, and the codex **sdk** runtime's variant of the same branch logs (`codex usage limit reached chatId=…`). External log watchers tailing `client.log` (e.g. account-failover automation that cools an exhausted provider account and switches to a backup) therefore never see the event.

This PR adds the missing `sessionCtx.log` call to the app-server usage-limit branch, covering both variants:

- provider-reported usage-limit failure (`usageLimitFailure`) — includes the provider error message
- empty turn with zero token delta (`usageLimitEmptyTurn`)

The line carries a stable machine-parseable tag, e.g.:

```
codex app-server usage limit reached (provider_usage_limit) chatId=…: turn failed with usage-limit error: …; posting a chat notice instead of silently acking the message
```

Watchers can match `provider_usage_limit` (or `usage limit reached`, which also matches the sdk runtime's existing line).

Addresses the missing-slot-log half of #1732. The second half of that issue (auth-epoch restarts deferred indefinitely behind wedged in-flight turns) is a separate design-level change and is not touched here.

## Testing

- Extended the existing app-server usage-limit test (`codex-app-server-extra-coverage.test.ts`) to assert the slot-log line and its `provider_usage_limit` tag are emitted.
- `pnpm vitest run` on the four codex app-server / usage-limit test files: 89 passed, 3 skipped.
- `pnpm typecheck` (client package) and `biome check` on the changed files: clean.

### DIFF
--- a/packages/client/src/__tests__/codex-app-server-extra-coverage.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-extra-coverage.test.ts
@@ -925,7 +925,8 @@ describe("codex app-server handler extra branches", () => {
     const successFake = new FakeAppServerClient();
     const successToken = makeDeliveryToken();
     const successHandler = makeHandler(successFake);
-    const successCtx = makeContext();
+    const successLog = vi.fn<(message: string) => void>();
+    const successCtx = makeContext({ log: successLog });
     const successSendMessage = vi.fn<SessionContext["sdk"]["sendMessage"]>().mockResolvedValue(sentMessageResponse());
     successCtx.sdk.sendMessage = successSendMessage;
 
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed at `b81301ac`.

The change adds the missing slot-log signal inside the existing Codex app-server usage-limit settlement branch without changing notice delivery or inbox settlement semantics. The provider-reported path retains the provider message, and both variants carry the stable `provider_usage_limit` tag / `usage limit reached` phrase needed by log watchers.

No blockers found. I verified the four focused Codex app-server / usage-limit suites (89 passed), client type-check, and Biome on the changed files.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1733

**Recommendation**: approve
- Rationale: The change restores the missing slot-log signal within the existing usage-limit settlement branch without altering notice delivery or inbox settlement semantics.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Author / repo: L42y / agent-team-foundation/first-tree
- Problem: Codex app-server usage-limit turns notify the chat but leave no slot-log evidence, so log-based account cooldown and failover automation cannot react.
- Approach: Add a stable `provider_usage_limit` log signal for both provider-reported usage-limit failures and the existing zero-token empty-turn path, while preserving the current chat notice and ACK/retry behavior.
- Impacted modules: `packages/client/src/handlers/codex/app-server/`, focused Codex app-server coverage tests

**Review findings**
✅ 1. The log is emitted inside the existing shared usage-limit guard and does not introduce a parallel settlement path.
✅ 2. Provider-reported failures retain the provider message, while both variants expose stable watcher-friendly text.
✅ 3. The focused regression assertion covers the stable tag, generic phrase, and provider message; remote CI is green.

**Action taken**
- Approved at commit `b81301ac`.

PR #1734, opened 2026-07-13
fix(web): clamp iOS standalone root scrolling

Gandy2025:
## Summary

- add an iOS standalone-only guard that keeps the document scrolling root at the origin after keyboard, viewport, route, resume, and orientation events
- lock document-level overflow only while the mobile shell is mounted, leaving each mobile page as the bounded scroller
- contain vertical overscroll in `MobilePage` and cover the guard lifecycle and scroll clamping with regression tests

## Context

On an iPhone 14 Pro running iOS 26.5, the home-screen Web App can reveal an extra Home Indicator-sized blank strip below the bottom navigation after using the chat keyboard and then scrolling. The existing `100dvh` fix is present but does not prevent WebKit from retaining a document-level scroll range. The behavior matches [WebKit bug 292603](https://bugs.webkit.org/show_bug.cgi?id=292603).

## Validation

- `pnpm check`
- `pnpm typecheck`
- targeted mobile tests: 8 passed
- Web Vitest suite: all 1,336 unit tests passed; the command then returned non-zero because the existing `tests/visual-regression.spec.ts` Playwright file was collected by Vitest and contains no Vitest suite
- repository-wide `pnpm test` was additionally attempted; unrelated CLI client-switch tests detect the currently running First Tree Codex provider process and fail their process-drain assertions in this agent environment

## QA

Retest the original chat-keyboard flow on the reported iPhone 14 Pro / iOS 26.5 device after this branch reaches staging.

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -672,6 +672,15 @@
     text-rendering: optimizeLegibility;
     font-synthesis: none;
   }
+  /* iOS standalone Web Apps can leave a root scroll range equal to the
+     Home Indicator safe area after the keyboard closes. Mobile pages own
+     their scrolling, so lock every document-level box while the guard is
+     mounted and let the bounded page scroller handle touch movement. */
+  html.ios-standalone-root-scroll-guard,
+  html.ios-standalone-root-scroll-guard body,
+  html.ios-standalone-root-scroll-guard #root {
+    overflow: hidden;
+  }
   ::selection {
     background: var(--brand-bg);
     color: var(--fg);
--- a/packages/web/src/pages/mobile/__tests__/root-scroll-guard.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/root-scroll-guard.test.tsx
@@ -1,0 +1,152 @@
+// @vitest-environment happy-dom
+
+import { act } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { createDomHarness, type DomHarness } from "../../../test-utils/dom-harness.js";
+import { isIosStandalone, useIosStandaloneRootScrollGuard } from "../use-ios-standalone-root-scroll-guard.js";
+
+function GuardProbe({ resetKey = "route-1" }: { resetKey?: string }) {
+  useIosStandaloneRootScrollGuard(resetKey);
+  return <div>guard</div>;
+}
+
+describe("iOS standalone root scroll guard", () => {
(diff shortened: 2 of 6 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1734

**Recommendation**: approve
- Rationale: The focused iOS standalone guard clamps only invalid document-level scrolling while preserving the bounded mobile-page scrollers and cleaning up all global state on shell unmount.

**Risk level**: C

**PR summary**
- Problem: On iOS home-screen Web Apps, closing the chat keyboard can leave a document-level safe-area scroll range that exposes a blank strip below the bottom navigation.
- Approach: While the live mobile shell is mounted in iOS standalone mode, lock the document-level boxes, clamp any root offset after viewport/keyboard/navigation/resume/orientation signals, and keep each `MobilePage` as the bounded vertical scroller with contained overscroll.

**Review findings**
✅ 1. The guard is narrowly gated to iOS standalone display mode and its class/listener/animation-frame lifecycle is fully scoped to `MobileShell` mount and unmount.
✅ 2. All recovery signals converge on one requestAnimationFrame-deduplicated clamp, while route changes reuse the same mounted guard instead of duplicating listeners.
✅ 3. The existing `h-full` / `min-h-0` shell chain keeps page scrolling inside `MobilePage`; the root overflow lock does not replace or erase the page scroll position.
✅ 4. Regression coverage exercises mode detection, initial and event-driven clamping, route resets, cleanup, and the shared page overscroll contract. Physical-device staging QA remains appropriate for the WebKit-specific behavior.

**Action taken**
- Approved.

yuezengwu (approved):

The implementation keeps the workaround narrowly scoped to the mounted mobile shell in iOS standalone mode. The document-level class, listeners, and pending animation frame are all cleaned up on unmount; viewport, keyboard, navigation, resume, orientation, and root-scroll signals converge on one requestAnimationFrame-deduplicated clamp; and `MobilePage` remains the bounded scroll owner with vertical overscroll containment.

I also verified that this does not touch the database, schema, shared data structures, subscriptions, or WebSocket behavior.

Local validation at `71f097a6`:
- targeted mobile tests: 8 passed
- Web typecheck and design-token checks: passed
- Web production build: passed after building the workspace `@first-tree/shared` dependency
- generated CSS contains both the standalone root overflow lock and `overscroll-behavior-y: contain`

Physical-device staging QA is still appropriate for the WebKit-specific symptom described in bug 292603.

PR #1736, opened 2026-07-13
revert: iOS standalone root scroll guard

Gandy2025:
## Summary

- revert #1734 in full
- remove the iOS standalone document overflow lock and root scroll clamp
- restore the previous `MobilePage` overscroll behavior

## Why

Physical-device verification on the reported iPhone 14 Pro / iOS 26.5 showed that #1734 changes the bottom blank strip from intermittent to consistently reproducible.

The workaround assumed the strip was a normal document scroll offset. It is instead WebKit-managed viewport/safe-area space: `window.scrollY` can remain zero while the strip is visible, so the clamp cannot correct it. Keeping `html`, `body`, and `#root` at `overflow: hidden` also matches the ineffective reproduction conditions documented in [WebKit 292603](https://bugs.webkit.org/show_bug.cgi?id=292603), and stabilizes the bad geometry when combined with the existing standalone `viewport-fit=cover` / `100dvh` shell.

This rollback restores the pre-#1734 behavior while a replacement fix is designed around the standalone viewport height itself.

## Validation

- confirmed the affected files are identical to the parent of #1734 after the revert
- mobile shell DOM tests: 5 passed
- `pnpm check`
- `pnpm typecheck` (includes Web production build and design-token checks)

## QA

After staging deploy, verify that the bottom strip is no longer permanently present on iPhone 14 Pro / iOS 26.5. The original intermittent WebKit symptom is expected to remain until the replacement fix lands.


Gandy2025:
Superseded by #1737 with the repository-approved fix/ branch prefix; the commit and rollback scope are unchanged.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1736

**Recommendation**: approve
- Rationale: The rollback cleanly removes a workaround that physical-device verification disproved and restores the affected mobile shell paths exactly to their pre-#1734 state.

**Risk level**: C

**PR summary**
- Problem: The iOS standalone root-scroll workaround made the bottom safe-area blank strip consistently reproducible instead of reducing the intermittent WebKit symptom.
- Approach: Remove the document overflow lock, root scroll clamp, and overscroll containment introduced by #1734, restoring the prior mobile-page scrolling behavior while a viewport-height-based replacement is designed.

**Review findings**
✅ 1. All six affected paths are byte-for-byte identical to their state immediately before #1734, so the ineffective guard and its tests are removed without leaving a parallel or partial code path.
✅ 2. The rollback matches the physical-device evidence and WebKit 292603's reproduction boundary: the safe-area strip can remain despite root overflow and overscroll constraints, so retaining the guard would stabilize behavior known to be harmful.
✅ 3. The existing `h-full` / `min-h-0` shell chain and `MobilePage` overflow owner remain intact; only the disproved root-level intervention and associated containment are reverted.
✅ 4. Staging verification on the affected iPhone remains the correct confidence gate because this WebKit-managed viewport behavior is outside the repository's automated cross-process/browser coverage.

**Action taken**
- Approved.

PR #1737, opened 2026-07-13
revert: iOS standalone root scroll guard

Gandy2025:
## Summary

- revert #1734 in full
- remove the iOS standalone document overflow lock and root scroll clamp
- restore the previous `MobilePage` overscroll behavior

## Why

Physical-device verification on the reported iPhone 14 Pro / iOS 26.5 showed that #1734 changes the bottom blank strip from intermittent to consistently reproducible.

The workaround assumed the strip was a normal document scroll offset. It is instead WebKit-managed viewport/safe-area space: `window.scrollY` can remain zero while the strip is visible, so the clamp cannot correct it. Keeping `html`, `body`, and `#root` at `overflow: hidden` also matches the ineffective reproduction conditions documented in [WebKit 292603](https://bugs.webkit.org/show_bug.cgi?id=292603), and stabilizes the bad geometry when combined with the existing standalone `viewport-fit=cover` / `100dvh` shell.

This rollback restores the pre-#1734 behavior while a replacement fix is designed around the standalone viewport height itself.

## Validation

- confirmed the affected files are identical to the parent of #1734 after the revert
- mobile shell DOM tests: 5 passed
- `pnpm check`
- `pnpm typecheck` (includes Web production build and design-token checks)

## QA

After staging deploy, verify that the bottom strip is no longer permanently present on iPhone 14 Pro / iOS 26.5. The original intermittent WebKit symptom is expected to remain until the replacement fix lands.


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1737

**Recommendation**: approve
- Rationale: The rollback fully removes the workaround that physical-device QA showed made the WebKit-managed bottom safe-area strip consistently reproducible, restoring the prior intermittent behavior without leaving a partial scroll-lock path.

**Risk level**: C

**PR summary**
- Problem: On iPhone standalone Web Apps, the attempted root-scroll workaround turns an intermittent WebKit bottom blank strip into a consistently visible defect after keyboard interaction.
- Approach: Revert the document overflow lock, root-scroll clamp, shared page overscroll containment, and workaround-specific tests as one unit, returning staging to the pre-workaround mobile-shell behavior while a viewport-height-based replacement is designed.
- Impacted modules: `packages/web/src/index.css` and `packages/web/src/pages/mobile/**`.

**Review findings**
✅ 1. All six affected files match the parent of #1734, so the behavior and its dedicated test contract are reverted atomically with no stale hook, class, listener, or overscroll assertion.
✅ 2. The cited WebKit report confirms that the safe-area strip can remain scrollable despite root overflow locks, overscroll containment, and dynamic viewport-unit constraints; physical-device QA therefore provides a sound reason to remove the ineffective workaround.
✅ 3. The rollback preserves the existing bounded `MobilePage` scroller and changes no mobile capability scope, subscriptions, WebSocket parsing, or shared data semantics.

**Action taken**
- Approved.

yuezengwu (approved):

## Goal and core change

This PR fully rolls back #1734 after physical-device verification showed that the iOS standalone root-scroll workaround made the bottom safe-area strip consistently reproducible. It removes the document overflow lock and root-offset clamp, and restores the prior `MobilePage` overscroll behavior while a viewport-height-based replacement is designed.

## Findings

- The head commit is the same reviewed rollback commit as superseded #1736 (`84b8e38e`); only the branch name changed to the repository-approved `fix/` prefix.
- All six affected paths are byte-for-byte identical to their state immediately before #1734. There are no remaining guard imports, CSS classes, listeners, or partial parallel paths.
- The existing bounded mobile scroll chain remains intact: the shell keeps the `100dvh`/`min-h-0` layout and `MobilePage` remains the `h-full overflow-y-auto` scroll owner.
- WebKit 292603 documents the same failure boundary: the safe-area strip persists even with root overflow locking, overscroll containment, and viewport-unit constraints. That supports removing the disproved root-level intervention.
- The diff is limited to `packages/web/**`; there are no database/schema, shared core data-structure, subscription, or WebSocket changes requiring separate human inspection.

Local review checks at `84b8e38e`:

- affected-path equivalence with the parent of #1734: passed
- `git diff --check`: passed
- residual guard symbol/class search: clean

Staging verification on the reported iPhone 14 Pro / iOS 26.5 remains the appropriate acceptance check. The rollback is expected to remove the newly permanent regression, while the original intermittent WebKit symptom remains until the replacement fix lands.

yuezengwu (approved):

The goal and core change are sound: this fully rolls back #1734 after physical-device QA showed that the root-scroll workaround turns the iOS standalone safe-area strip from intermittent into consistently reproducible. It removes the document overflow lock, root-scroll hook/listeners, overscroll containment, and workaround-specific tests as one unit, while preserving the existing bounded `MobilePage` scroller, `100dvh` shell sizing, and safe-area utilities.

I verified:
- all six affected paths are identical to the parent of #1734 (`f9596fa8`) at this head, so there is no partial rollback or stale hook/class/listener reference;
- the WebKit 292603 reproduction supports the diagnosis that `overflow: hidden`, overscroll containment, and viewport-unit constraints do not suppress this WebKit-managed bottom region;
- the change does not touch the database, schema, core data structures, subscriptions, WebSocket parsing, or shared data semantics.

Risk level: C (`packages/web/**` layout/scroll behavior only). Staging QA on the reported iPhone 14 Pro / iOS 26.5 remains required: the permanently visible regression should disappear, while the original intermittent WebKit symptom is expected to remain pending the viewport-height-based replacement.

PR #724, opened 2026-06-01
fix: delegate is self-only — block admins from setting others' delegate

Gandy2025:
## What & why

Product decision (from @gandy2025): **a delegate is a personal choice — we do not support an admin setting a delegate on someone else's behalf.**

Today `PATCH /agents/:uuid` runs under `manage` scope, which lets an admin manage any agent in the org. The Team page already gates delegate editing to self (`canEditDelegate: isSelf`), but the **agent-detail "Edit Identity" dialog and the API itself did not** — an admin could set/change/clear another member's `delegateMention`. This closes that gap on both the server (authoritative) and the second UI entry point.

### Server (authoritative) — `packages/server/src/api/agents.ts`
```ts
if (body.delegateMention !== undefined && scope.humanAgentId !== request.params.uuid) {
  throw new ForbiddenError("Only the member themselves can set their own delegate");
}
```
Mirrors the existing `managerId`-only-by-admin guard right above it. `delegateMention` writes (set / change / clear) are allowed only when the caller is acting on their **own** human agent.

### Web — `packages/web/src/pages/agent-detail/identity-section.tsx`
- `canEditDelegate = isHuman && agent.managerId === memberId` (self only — drops the admin path that visibility editing still allows).
- Delegate `<select>` is **disabled** for non-owners, with copy: *"Only the member themselves can set their own delegate."*
- `submit()` only includes `delegateMention` in the PATCH when the caller owns the agent — so an admin editing another member's display name doesn't carry the field and trip the 403.
- Candidate list now filters by `managerId === self`, **matching the Team page selector** — this also closes the candidate-set inconsistency raised in the PR #720 review (observation 2).

## Tests — `agent-delegate-self-only.test.ts` (TDD, red→green)
- admin sets another member's `delegateMention` → **403**
- admin clears another member's `delegateMention` → **403**
- member sets their own → **200**
- admin edits another member's **non-delegate** field (display name) → **200** (guard is scoped to `delegateMention` only, no over-blocking)

## Verification
- server typecheck ✓ · **full server suite 1262/1262** ✓
- web typecheck + design-token lint ✓ · **full web suite 514/514** ✓
- biome ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1738, opened 2026-07-13
feat(server): chat list ordering + priority projection (attention/pinned)

Gandy2025:
## What

PR 3 of the Workspace chat-list refactor: **server-side ordering + priority projection**. Builds on PR 1 (`useInfiniteQuery` foundation) and PR 2 (`pinned_at` / `activity_at` / pin mutation), both merged.

`listMeChats` (backing `GET /orgs/:orgId/chats?scope=mine`) now returns two whole-set priority groups plus the ordinary page:

```jsonc
{
  "priorityRows": {
    "attention": [ /* MeChatRow[] — failed-first, then activity DESC */ ],
    "pinned":    [ /* MeChatRow[] — pinned_at DESC, minus attention */ ]
  },
  "rows": [ /* the complete activity-ordered recency page */ ],
  "nextCursor": "…"
}
```

- **attention** — a *caller-managed* non-human speaker in `failed`, OR an open request addressed to the caller. Ordered failed-first, then `activity_at` DESC. Viewer-scoped: a peer's failed agent / another human's request stays in the ordinary stream.
- **pinned** — the caller's pins (`pinned_at` DESC), minus anything already in attention (attention and pinned are disjoint).
- **rows** — the ordinary `activity_at`-ordered keyset page.

## Two design decisions (called out for review)

**1. `rows` is ADDITIVE, not exclusive.** A pinned / attention chat also appears in `rows`; the client renders it once by de-duplicating the recency stream against `priorityRows`. This is deliberate: the already-shipped web (PR 1) reads **only** `rows` and does not consume `priorityRows` until PR 4. An *exclusive* `rows` would make pinned / open-request / failed chats **vanish** from the live list during the window between this server deploy and PR 4. Additive keeps the change backward-compatible and standalone-deployable.

**2. Priority groups are computed on the first page only** (no `cursor`), returned empty on load-more pages (the client reads them from page 1). This bounds cost — `resolveAgentChatStatuses` over the candidate set runs once per list-open, not per scroll page — and the additive `rows` above is what makes it safe (later pages need no priority ids to exclude).

## Why / implementation

- **Ordering key is `activity_at`** (PR 2): real-work recency, `NOT NULL`, so the keyset cursor drops the old `NULLS LAST` / null-timestamp branches.
- **Versioned cursor.** The payload is now `v2|<activityAtIso>|<chatId>`. The pre-PR cursor was an indistinguishable `<lastMessageAt>|<chatId>`, so `decodeCursor` classifies **valid-v2 / recognized-legacy / invalid**: a legacy cursor held across the rollout restarts from page 1 (graceful), while a malformed/unsupported cursor stays a typed 400.
- **`failed` is never re-implemented in SQL.** It is a composite status from `resolveAgentChatStatuses`. Attention is built from a **bounded candidate set** — an open request, OR a caller-managed non-human speaker whose *stored error inputs* mirror `computeErrored`'s branches (session `errored`; per-chat runtime `error` when stamped; agent-global presence `error` only when unstamped). That is a strict superset of failed (no false negatives; the canonical resolver still confirms freshness / reachability) while excluding the common healthy managed speaker — so a 30s poll no longer enriches the caller's whole active list.
- **One projection, three sets.** The single-stream `SELECT` is factored into `selectMeChatRawRows` + `enrichMeChatRows` so attention / pinned / ordinary rows are byte-identical in shape and share one enrichment path.
- **Web boundary parses.** `listMeChats` (web) now `parse`s the response with `listMeChatsResponseSchema` (was a bare cast), so the schema's version-skew `.default` actually fills a missing `priorityRows` against an older server.

## `counts.unread` deferred

An earlier revision added a global unread `counts` aggregate. Its scope (global vs the active engagement view) is a product decision that belongs with the unread badge that consumes it (the desktop-rail PR), and shipping it here first would publish an unsettled shared contract — so `counts` is **not** part of this response. The `countUnreadMeChats` helper stays for its existing caller.

## Testing

- `me-chat-priority.test.ts` — 15 tests on real Postgres (testcontainers): genuine `failed → attention` (reachable + `errored`); pinned `pinned_at` DESC; global cross-page extraction; open-request → attention; healthy managed speaker stays ordinary; **peer's failed speaker stays in rows** (viewer-scoping); **global-error + stamped per-chat idle stays in rows** (candidate gating); a **direct attention-candidate-query boundary test** — via the extracted `selectAttentionCandidateRows` — asserting a failed speaker IS admitted (positive control) while stamped-idle / healthy are NOT; failed-before-request ordering; attention outranks pinned; disjoint priority groups; keyset pagination completeness; first-page gating; **legacy-cursor recovery**; origin-filter narrowing.
- `me-chat-helpers.test.ts` — cursor codec incl. the three-way legacy / invalid classification.
- `me-chats-http-e2e.test.ts` — Fastify serializes `priorityRows` end-to-end.
- `me-chat-service.test.ts` — legacy cursor → page 1, invalid cursor → 400.
- Web `api-wrappers.test.ts` — proves the version-skew default via a parsed older-server payload.
- `typecheck` (server + web) + `biome` clean; full server suite green (one unrelated pre-existing failure is a local git-locale artifact in `context-tree-snapshot`, passes on CI).

## Independent review

Two pre-push independent reviews + three GitHub-reviewer rounds (baixiaohang + yuezengwu). SQL, multi-org isolation, keyset determinism verified sound; deploy-compat (additive), per-page cost (first-page gating + candidate narrowing), cursor safety (versioning + 3-way recovery), version-skew (web parse), and `counts` scope (deferred) all addressed.

## Paired tree PR

Reconciles the ranking contract in `system/cloud/chat/workspace-conversations.md` (owners @baixiaohang @yuezengwu): real-work recency basis + the viewer's private **Pinned** tier below viewer-scoped attention, extracted across the whole *matching* set, "each chat rendered once." Draft, marked ready after this PR + the desktop-rail client merge: **agent-team-foundation/first-tree-context#724**.

### DIFF
--- a/packages/server/src/__tests__/direct-chat-auto-mention.test.ts
+++ b/packages/server/src/__tests__/direct-chat-auto-mention.test.ts
@@ -49,12 +49,17 @@ describe("1:1 chat wake-up + unread badge (explicit-mention contract)", () => {
     organizationId: string,
   ): Promise<number> {
     const app = getApp();
-    const { rows } = await listMeChats(app.db, agentUuid, memberId, organizationId, {
+    const { priorityRows, rows } = await listMeChats(app.db, agentUuid, memberId, organizationId, {
       limit: 10,
       filter: "all",
       engagement: "all",
     });
-    return rows.find((r) => r.chatId === chatId)?.unreadMentionCount ?? 0;
+    // A `request` DM opens a request, routing the chat into the attention group
+    // rather than ordinary `rows` — search all groups for the unread counter.
+    return (
+      [...priorityRows.attention, ...priorityRows.pinned, ...rows].find((r) => r.chatId === chatId)
+        ?.unreadMentionCount ?? 0
+    );
   }
 
   async function notifyInboxRows(chatId: string, agentUuid: string) {
--- a/packages/server/src/__tests__/me-chat-helpers.test.ts
+++ b/packages/server/src/__tests__/me-chat-helpers.test.ts
@@ -2,12 +2,14 @@ import { describe, expect, it, vi } from "vitest";
 import { decodeCursor, encodeCursor, resolveChatTitle } from "../services/me-chat.js";
 
 describe("me-chat encodeCursor / decodeCursor", () => {
-  it("round-trips a timestamp + chat id", () => {
+  it("round-trips a timestamp + chat id as an `ok` cursor", () => {
     const ts = new Date("2026-05-06T10:24:00.000Z");
-    const cursor = encodeCursor(ts, "chat-123");
-    const decoded = decodeCursor(cursor);
-    expect(decoded?.activityAt?.toISOString()).toBe(ts.toISOString());
-    expect(decoded?.chatId).toBe("chat-123");
+    const decoded = decodeCursor(encodeCursor(ts, "chat-123"));
+    expect(decoded.status).toBe("ok");
+    if (decoded.status === "ok") {
+      expect(decoded.activityAt.toISOString()).toBe(ts.toISOString());
+      expect(decoded.chatId).toBe("chat-123");
+    }
   });
 
   it("emits a versioned payload (v2 prefix)", () => {
@@ -15,39 +17,42 @@ describe("me-chat encodeCursor / decodeCursor", () => {
     expect(Buffer.from(cursor, "base64url").toString("utf8")).toBe("v2|2026-05-06T10:24:00.000Z|chat-123");
   });
 
-  it("rejects an unversioned pre-PR cursor rather than reinterpreting it", () => {
-    // The old shape was `<lastMessageAt>|<chatId>` (2 parts) — indistinguishable
-    // from a raw activity cursor, so it MUST be rejected across the rollout, not
-    // resumed from a wrong boundary.
+  it("classifies a recognized unversioned pre-PR cursor as `legacy` (never reinterpreted as v2)", () => {
+    // The old shape was `<lastMessageAt>|<chatId>` (2 parts) — its timestamp meant
+    // last_message_at, so it must NOT be reinterpreted against activity ordering;
+    // `legacy` lets the caller restart from page 1 instead of resuming wrongly.
     const legacy = Buffer.from("2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
-    expect(decodeCursor(legacy)).toBeNull();
+    expect(decodeCursor(legacy).status).toBe("legacy");
   });
 
-  it("rejects a cursor with a different version", () => {
+  it("marks a different version `invalid`", () => {
     const wrongVersion = Buffer.from("v1|2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
-    expect(decodeCursor(wrongVersion)).toBeNull();
+    expect(decodeCursor(wrongVersion).status).toBe("invalid");
   });
 
-  it("rejects a v2 cursor with an empty timestamp or chat id", () => {
-    expect(decodeCursor(Buffer.from("v2||chat-no-ts", "utf8").toString("base64url"))).toBeNull();
-    expect(decodeCursor(Buffer.from("v2|2026-05-06T10:24:00.000Z|", "utf8").toString("base64url"))).toBeNull();
+  it("marks a v2 cursor with an empty timestamp or chat id `invalid`", () => {
+    expect(decodeCursor(Buffer.from("v2||chat-no-ts", "utf8").toString("base64url")).status).toBe("invalid");
+    expect(decodeCursor(Buffer.from("v2|2026-05-06T10:24:00.000Z|", "utf8").toString("base64url")).status).toBe(
+      "invalid",
+    );
   });
 
-  it("returns null for malformed cursor strings", () => {
-    expect(decodeCursor("not-a-cursor")).toBeNull();
-    expect(decodeCursor("")).toBeNull();
-    // base64-decodable but no separators
-    expect(decodeCursor(Buffer.from("nosep", "utf8").toString("base64url"))).toBeNull();
+  it("marks malformed cursor strings `invalid`", () => {
+    expect(decodeCursor("").status).toBe("invalid");
+    // one part, no separator
+    expect(decodeCursor(Buffer.from("nosep", "utf8").toString("base64url")).status).toBe("invalid");
+    // two parts but a bad timestamp (not a recognizable legacy cursor)
+    expect(decodeCursor(Buffer.from("not-a-date|chat", "utf8").toString("base64url")).status).toBe("invalid");
     // right version + shape but a bad timestamp
-    expect(decodeCursor(Buffer.from("v2|not-a-date|chat", "utf8").toString("base64url"))).toBeNull();
+    expect(decodeCursor(Buffer.from("v2|not-a-date|chat", "utf8").toString("base64url")).status).toBe("invalid");
   });
 
-  it("returns null when base64 decoding throws", () => {
+  it("marks a cursor `invalid` when base64 decoding throws", () => {
     const spy = vi.spyOn(Buffer, "from").mockImplementationOnce(() => {
       throw new Error("decode failed");
     });
     try {
-      expect(decodeCursor("bad-base64")).toBeNull();
+      expect(decodeCursor("bad-base64").status).toBe("invalid");
     } finally {
       spy.mockRestore();
     }
--- a/packages/server/src/__tests__/me-chat-helpers.test.ts
+++ b/packages/server/src/__tests__/me-chat-helpers.test.ts
@@ -2,19 +2,16 @@ import { describe, expect, it, vi } from "vitest";
 import { decodeCursor, encodeCursor, resolveChatTitle } from "../services/me-chat.js";
 
 describe("me-chat encodeCursor / decodeCursor", () => {
-  it("round-trips a non-null timestamp + chat id", () => {
+  it("round-trips a timestamp + chat id", () => {
     const ts = new Date("2026-05-06T10:24:00.000Z");
     const cursor = encodeCursor(ts, "chat-123");
     const decoded = decodeCursor(cursor);
-    expect(decoded?.lastMessageAt?.toISOString()).toBe(ts.toISOString());
+    expect(decoded?.activityAt?.toISOString()).toBe(ts.toISOString());
     expect(decoded?.chatId).toBe("chat-123");
   });
 
-  it("round-trips a null timestamp", () => {
-    const cursor = encodeCursor(null, "chat-no-msgs");
-    const decoded = decodeCursor(cursor);
-    expect(decoded?.lastMessageAt).toBeNull();
-    expect(decoded?.chatId).toBe("chat-no-msgs");
+  it("rejects a cursor with an empty timestamp part (activity_at is never null)", () => {
+    expect(decodeCursor(Buffer.from("|chat-no-ts", "utf8").toString("base64url"))).toBeNull();
   });
 
   it("returns null for malformed cursor strings", () => {
--- a/packages/server/src/__tests__/me-chat-helpers.test.ts
+++ b/packages/server/src/__tests__/me-chat-helpers.test.ts
@@ -17,12 +17,16 @@ describe("me-chat encodeCursor / decodeCursor", () => {
     expect(Buffer.from(cursor, "base64url").toString("utf8")).toBe("v2|2026-05-06T10:24:00.000Z|chat-123");
   });
 
-  it("classifies a recognized unversioned pre-PR cursor as `legacy` (never reinterpreted as v2)", () => {
-    // The old shape was `<lastMessageAt>|<chatId>` (2 parts) — its timestamp meant
-    // last_message_at, so it must NOT be reinterpreted against activity ordering;
-    // `legacy` lets the caller restart from page 1 instead of resuming wrongly.
-    const legacy = Buffer.from("2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
-    expect(decodeCursor(legacy).status).toBe("legacy");
+  it("classifies both recognized unversioned pre-PR cursor shapes as `legacy`", () => {
+    // The old encoder emitted `<lastMessageAt>|<chatId>` for a normal boundary AND
+    // `|<chatId>` (empty timestamp) for a `last_message_at IS NULL` tail boundary.
+    // Both are deployed shapes whose timestamp meant last_message_at, so neither
+    // may be reinterpreted against activity ordering; `legacy` lets the caller
+    // restart from page 1 instead of resuming wrongly (or 400ing).
+    const withTs = Buffer.from("2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
+    const emptyTs = Buffer.from("|chat-123", "utf8").toString("base64url");
+    expect(decodeCursor(withTs).status).toBe("legacy");
+    expect(decodeCursor(emptyTs).status).toBe("legacy");
   });
 
   it("marks a different version `invalid`", () => {
--- a/packages/server/src/__tests__/me-chat-helpers.test.ts
+++ b/packages/server/src/__tests__/me-chat-helpers.test.ts
@@ -10,19 +10,36 @@ describe("me-chat encodeCursor / decodeCursor", () => {
     expect(decoded?.chatId).toBe("chat-123");
   });
 
-  it("rejects a cursor with an empty timestamp part (activity_at is never null)", () => {
-    expect(decodeCursor(Buffer.from("|chat-no-ts", "utf8").toString("base64url"))).toBeNull();
+  it("emits a versioned payload (v2 prefix)", () => {
+    const cursor = encodeCursor(new Date("2026-05-06T10:24:00.000Z"), "chat-123");
+    expect(Buffer.from(cursor, "base64url").toString("utf8")).toBe("v2|2026-05-06T10:24:00.000Z|chat-123");
+  });
+
+  it("rejects an unversioned pre-PR cursor rather than reinterpreting it", () => {
+    // The old shape was `<lastMessageAt>|<chatId>` (2 parts) — indistinguishable
+    // from a raw activity cursor, so it MUST be rejected across the rollout, not
+    // resumed from a wrong boundary.
+    const legacy = Buffer.from("2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
+    expect(decodeCursor(legacy)).toBeNull();
+  });
+
+  it("rejects a cursor with a different version", () => {
+    const wrongVersion = Buffer.from("v1|2026-05-06T10:24:00.000Z|chat-123", "utf8").toString("base64url");
+    expect(decodeCursor(wrongVersion)).toBeNull();
+  });
+
+  it("rejects a v2 cursor with an empty timestamp or chat id", () => {
+    expect(decodeCursor(Buffer.from("v2||chat-no-ts", "utf8").toString("base64url"))).toBeNull();
+    expect(decodeCursor(Buffer.from("v2|2026-05-06T10:24:00.000Z|", "utf8").toString("base64url"))).toBeNull();
   });
 
   it("returns null for malformed cursor strings", () => {
     expect(decodeCursor("not-a-cursor")).toBeNull();
     expect(decodeCursor("")).toBeNull();
-    // base64-decodable but no separator
+    // base64-decodable but no separators
     expect(decodeCursor(Buffer.from("nosep", "utf8").toString("base64url"))).toBeNull();
-    // separator but empty chatId
-    expect(decodeCursor(Buffer.from("ts|", "utf8").toString("base64url"))).toBeNull();
-    // bad timestamp
-    expect(decodeCursor(Buffer.from("not-a-date|chat", "utf8").toString("base64url"))).toBeNull();
+    // right version + shape but a bad timestamp
+    expect(decodeCursor(Buffer.from("v2|not-a-date|chat", "utf8").toString("base64url"))).toBeNull();
   });
 
   it("returns null when base64 decoding throws", () => {
--- a/packages/server/src/__tests__/me-chat-pin-activity.test.ts
+++ b/packages/server/src/__tests__/me-chat-pin-activity.test.ts
@@ -35,12 +35,20 @@ describe("pinMeChat + chats.activity_at", () => {
   }
 
   async function rowFor(app: App, chatId: string, admin: Admin) {
-    const { rows } = await listMeChats(app.db, admin.humanAgentUuid, admin.memberId, admin.organizationId, {
-      limit: 50,
-      filter: "all",
-      engagement: "all",
-    });
-    return rows.find((r) => r.chatId === chatId) ?? null;
+    const { priorityRows, rows } = await listMeChats(
+      app.db,
+      admin.humanAgentUuid,
+      admin.memberId,
+      admin.organizationId,
+      {
+        limit: 50,
+        filter: "all",
+        engagement: "all",
+      },
+    );
+    // A chat can surface in any group now (attention > pinned > ordinary rows) —
+    // a pinned chat lives in `priorityRows.pinned`, so search all three.
+    return [...priorityRows.attention, ...priorityRows.pinned, ...rows].find((r) => r.chatId === chatId) ?? null;
   }
 
   async function activityAtOf(app: App, chatId: string): Promise<Date | null> {
--- a/packages/server/src/__tests__/me-chat-priority.test.ts
+++ b/packages/server/src/__tests__/me-chat-priority.test.ts
@@ -1,25 +1,14 @@
 import { eq, sql } from "drizzle-orm";
 import type { FastifyInstance } from "fastify";
-import { describe, expect, it, vi } from "vitest";
+import { describe, expect, it } from "vitest";
 import { agentPresence } from "../db/schema/agent-presence.js";
 import { chats } from "../db/schema/chats.js";
 import { clients } from "../db/schema/clients.js";
 import { members } from "../db/schema/members.js";
 import { createAgent } from "../services/agent.js";
-import { resolveAgentChatStatuses } from "../services/agent-chat-status.js";
-import { createMeChat, listMeChats, pinMeChat } from "../services/me-chat.js";
+import { createMeChat, listMeChats, pinMeChat, selectAttentionCandidateRows } from "../services/me-chat.js";
 import { sendMessage } from "../services/message.js";
 import { createTestAdmin, useTestApp } from "./helpers.js";
-
-// Wrap (not replace) the canonical status resolver so a test can count how many
-// times the priority + ordinary enrichment passes invoke it — the observable
-// signal for whether a chat was admitted to the attention candidate set. Real
-// behavior is preserved; only calls are recorded.
-vi.mock("../services/agent-chat-status.js", async (importOriginal) => {
-  const actual = await importOriginal<typeof import("../services/agent-chat-status.js")>();
-  return { ...actual, resolveAgentChatStatuses: vi.fn(actual.resolveAgentChatStatuses) };
-});
-const resolveStatusesSpy = vi.mocked(resolveAgentChatStatuses);
 
 /**
  * Server-side priority projection (PR3): `listMeChats` splits its result into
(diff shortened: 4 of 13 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1738

- Rationale: The priority projection direction is sound, but the cursor semantic change can silently corrupt pagination across a server rollout, and the expanded shared response contract currently breaks existing Web consumers at build time.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes establish a B-low baseline.
- Semantic lift: `packages/shared/src/schemas/me-chat.ts` changes the core chat-list state/response contract, lifting this to B-high.

**PR summary**
- Problem: Workspace users need conversations that require action or were explicitly pinned to remain visible above the ordinary recency stream, without unread-only churn or server/client rollout gaps.
- Approach: The server projects whole-matching-set attention and pinned groups on the first page, keeps ordinary rows additive for the already-deployed client, orders recency by real-work activity, and exposes a global unread aggregate. The paired tree PR records the intended rendered precedence and real-work recency rule.
- Impacted modules: member chat-list service and pagination, shared chat response schema, server integration/service tests, and downstream Web cache/test writers of `ListMeChatsResponse`.

**Review findings**
❌ 1. Version the cursor before changing its sort-key semantics. The deployed cursor is already base64url of `<timestamp>|<chatId>`, where the timestamp is `last_message_at`; this PR emits and accepts the indistinguishable shape while interpreting it as `activity_at`. A browser holding an old `nextCursor` across the server rollout will therefore pass validation and continue from the wrong boundary, silently skipping or repeating chats whenever the two timestamps differ. Use a distinguishable new cursor version and either deliberately reject/restart old cursors or support their old ordering during the transition. [R5 / `packages/server/src/services/me-chat.ts:64`]

❌ 2. Complete the shared response-contract blast-radius update. `ListMeChatsResponse` now requires `priorityRows` and `counts`, but existing Web cache/fixture writers still construct `{ rows, nextCursor }`; the current Actions run fails the Web TypeScript build at `chat-by-id-center-dom.test.tsx:302,307` and `conversation-list-dom.test.tsx:140`, so neither the lint/typecheck nor test job reaches green. Update every typed writer/fixture in this PR. [R5 / `packages/shared/src/schemas/me-chat.ts:392`]

⚠️ 3. Correct the public schema documentation to match the additive wire contract. It currently says every chat appears in at most one response location and that the client never de-duplicates, while the implementation intentionally returns priority chats again in `rows`; only `attention` and `pinned` are server-disjoint, and the client must de-duplicate them from recency. The same contract should also state that priority groups and `counts.unread` are populated only on page 1. This comment is the handoff contract for PR 4, so leaving the exclusive wording risks reproducing the deploy blocker in the consumer. [R1 / `packages/shared/src/schemas/me-chat.ts:371`]

✅ 4. The implementation correctly reuses `resolveAgentChatStatuses` for composite failure, scopes failure to caller-managed speakers, preserves additive ordinary rows for the deployed Web, and keeps attention ahead of private pins. The focused server coverage exercises the main ordering, isolation, pagination, and HTTP-shape boundaries; I did not run checks locally per review policy and relied on the submitted code plus GitHub Actions evidence.

**Action taken**
- Submitted this request-changes review; no code, tests, or Context Tree files were modified.

yuezengwu (changes requested):
## Review by yuezengwu

**Recommendation: request changes.** The direction is consistent with the Workspace contract: real-work recency, attention above private pins, canonical composite failure, and additive `rows` for a safe server-first rollout. The current patch still has merge blockers in pagination compatibility, the shared response blast radius, and first-page query cost.

### Blocking findings

1. **Version the cursor before changing what its timestamp means.** The deployed cursor and this PR are both base64url `<timestamp>|<chatId>`, but the old timestamp is `last_message_at` and the new timestamp is `activity_at`. A browser holding an old `nextCursor` across the server rollout will pass validation and continue from the wrong boundary, silently skipping/repeating rows whenever those timestamps differ; old null-timestamp cursors become 400s. Use a distinguishable version and define the transition behavior. (`packages/server/src/services/me-chat.ts:64,656`)

2. **Complete the response-contract blast-radius update.** `ListMeChatsResponse` now requires `priorityRows` and `counts`, but existing typed Web cache/fixture writers still create `{ rows, nextCursor }`. Both GitHub Actions jobs and my local Web typecheck fail at `chat-by-id-center-dom.test.tsx:302,307` and `conversation-list-dom.test.tsx:140`. Update every writer/fixture in this PR. (`packages/shared/src/schemas/me-chat.ts:392`)

3. **Do not make the 30-second list poll enrich the caller's full active history.** `managedSpeakerExists` admits every matching chat containing any caller-managed non-human speaker, then `limit: null` sends that whole set through participant/title enrichment and `resolveAgentChatStatuses` (multiple unbounded `IN`/LATERAL queries). For the normal case—most chats contain the user's managed agent—this is effectively the full active list, not a set bounded by the number of managed agents. First-page gating avoids load-more work, but the shipped desktop list refetches page 1 every 30s and on WS invalidation; mobile shell plus the active mobile page use separate 30s queries, so the cost can be paid twice. Old clients pay it immediately even though they ignore `priorityRows`. Narrow the pre-canonical candidate set to chats that can actually be failed/open-request (then let the canonical resolver confirm), or provide an explicit bounded design plus representative capacity evidence before merge. (`packages/server/src/services/me-chat.ts:679-708`; polling at `packages/web/src/pages/workspace/conversations/index.tsx:171` and `packages/web/src/pages/mobile/{shell,now,chat}.tsx`)

4. **Wire the claimed version-skew defaults into the actual read path.** Zod `.default(...)` runs only when the schema parses a value. The Web wrapper currently does `api.get<ListMeChatsResponse>(...)`, and `api.get` only casts `res.json()`; neither the route nor the client calls `listMeChatsResponseSchema.parse`. Therefore a new Web bundle against an older server still receives `priorityRows === undefined` / `counts === undefined`. Parse/normalize the response in `listMeChats` (or otherwise make the fallback real) before relying on these defaults in PR 4. (`packages/shared/src/schemas/me-chat.ts:381-407`; `packages/web/src/api/me-chats.ts:44`)

### Product decision required

`counts.unread` already publishes a concrete org-wide, non-deleted contract that includes archived chats, while the PR explicitly says global-vs-engagement scope is unresolved. The current desktop/mobile badges count loaded `engagement: active` rows, so this does not match today's visible-list semantics. Because this is a shared public response contract—not merely inert implementation detail—settle the scope with the product owner and test it here, or remove/defer `counts` until that decision is made. The paired tree PR records ranking/pinning but does not establish unread-count scope. (`packages/server/src/services/me-chat.ts:730-733`; `packages/web/src/pages/workspace/conversations/index.tsx:223`)

Also fix the shared schema/test prose that still describes `rows` as exclusive. On the actual wire only `attention` and `pinned` are disjoint; priority rows intentionally repeat in additive `rows`, and the consumer must de-duplicate. The contract should also say that priority groups/counts are populated only on page 1. (`packages/shared/src/schemas/me-chat.ts:371-407`)

### Verification

- Context Tree contract checked: `system/cloud/chat/workspace-conversations.md`, `messaging.md`, and `sessions-and-status.md`; paired tree PR #724 matches the rendered ranking intent.
- `pnpm --filter @first-tree/server test`: **218 files / 2263 tests passed**.
- Focused priority/HTTP/cursor suite: **31 tests passed**.
- Shared and server typecheck: passed; changed-file Biome check: passed.
- Web typecheck: failed with the three response-fixture errors above.

This review changes no code, database state, or Context Tree files.

yuezengwu (changes requested):
## Re-review of `c1f3624f`

The incremental commit resolves the Web fixture/typecheck break and corrects the additive priority-group documentation. The `v2` prefix also prevents the dangerous silent reinterpretation of a legacy cursor. I re-ran shared/server/Web typechecks, the focused cursor/priority/HTTP tests (34 passing), and Biome; those checks are clean.

**Changes are still required on this head:**

1. **Finish the legacy-cursor recovery path.** `decodeCursor` now returns null for an old cursor and the server returns a generic 400, but the comment says “client restarts from page 1” and no such behavior exists. The current `useInfiniteQuery` load-more handler only calls `fetchNextPage()`; its error UI's Retry calls the same function with the same stored page param. It neither recognizes an expired cursor nor clears/refetches the infinite query. This avoids silent corruption but leaves an already-open client stuck retrying until a separate page-0 refresh happens to replace the cursor. Add an explicit expired-cursor error/recovery (or support the transition server-side) and cover the deployed-client rollout case. (`packages/server/src/services/me-chat.ts:66-70`; `packages/web/src/pages/workspace/conversations/index.tsx:565-595`)

2. **Bound/narrow the first-page attention candidate work.** This commit does not change the `managedSpeakerExists` + `limit: null` path: every active chat containing a caller-managed agent is still fully enriched and passed through the multi-query canonical status resolver on every page-0 poll/WS invalidation. The shipped desktop list polls every 30s; mobile can issue two independent 30s list queries. This is effectively a recurring full-history scan for a normal user's chat set, and old clients pay it before consuming priority rows. Narrow the candidate set to chats with a possible failure/open request while preserving canonical final classification, or establish a bounded design with representative capacity evidence. (`packages/server/src/services/me-chat.ts:691-720` on this head)

3. **Make the Zod version-skew fallback execute.** `.default(...)` still is not applied by the Web read path: `api.get<ListMeChatsResponse>` casts JSON and no caller parses `listMeChatsResponseSchema`. A Web-ahead-of-server rollout therefore still receives missing fields as `undefined`. Parse/normalize in the API wrapper (or otherwise implement the fallback) before PR 4 relies on it. (`packages/shared/src/schemas/me-chat.ts:391-417`; `packages/web/src/api/me-chats.ts:44`)

4. **Resolve the `counts.unread` scope before publishing it.** The server still commits active+archived semantics while the PR says global-vs-engagement scope is an open product question; today's desktop/mobile badges count active loaded rows. The paired tree change does not settle this contract. Get the product-owner decision and test it here, or defer/remove `counts` until it is decided. Also update its schema prose to state the page-0-only behavior; the current “page-independent aggregate” wording is indistinguishable from a real zero on load-more. (`packages/server/src/services/me-chat.ts:742-745`; `packages/shared/src/schemas/me-chat.ts:406-417`)

Small doc correction: the new priority comment says “`nextCursor === null` request”; the request field is `cursor`, so this should say `cursor` is omitted/null.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1738 (follow-up on `d27896ea`)

- Rationale: This increment resolves the previous product-scope and rollout items in the right direction, but the new cursor recovery is over-broad, the narrowed candidate still diverges from the canonical old/new-client split, and the Web parsing change leaves an existing test caller invalid.

**Risk level**: B-high
- Path baseline: server chat-list behavior plus Web API-boundary changes.
- Semantic lift: the increment changes pagination failure semantics and the core failed-attention candidate contract.

**PR summary**
- Problem: Existing and version-skewed Workspace clients must recover safely while attention projection remains cheap enough for recurring list refreshes.
- Approach: This head restarts undecodable cursors from page 1, prefilters attention candidates by stored error inputs, parses the shared response schema at the Web boundary, and defers the unresolved unread aggregate to PR 4.
- Impacted modules: chat-list pagination and candidate SQL, shared response schema, Web chat API wrapper, and related service/HTTP fixtures.

**Review findings**
❌ 1. Preserve the distinction between a known legacy cursor and an actually invalid cursor. `decodeCursor` returns `null` for both, and `listMeChats` now silently serves page 1 for every malformed value (`not-a-valid-cursor`, truncated v2 payloads, unknown future versions, etc.). The rollout requirement only calls for recovering the deployed two-part legacy shape; collapsing all parse failures hides client/API bugs and removes the previous specific 400 contract. Classify valid-v2 / recognized-legacy / invalid separately, recover only the recognized legacy form, and retain a typed failure for malformed or unsupported cursors. [R4 / `packages/server/src/services/me-chat.ts:642`]

❌ 2. Match the canonical old/new-client branch when narrowing the failure candidate. `computeErrored` consults `agent_presence.runtime_state = 'error'` only when an active session has `runtime_state_at IS NULL`; once a per-chat stamp exists, only the fresh per-chat runtime state is authoritative. The new SQL ORs `ap.runtime_state = 'error'` into every active session regardless of that stamp, so one agent-global error still admits all of that upgraded agent's active chat history—even chats with stamped per-chat `idle` that cannot canonically fail. Gate the presence fallback on `acs.runtime_state_at IS NULL` (and keep the canonical resolver responsible for final freshness/reachability) so the prefilter is the claimed possible-failure superset rather than a wider cross-chat scan. [R1 / `packages/server/src/services/me-chat.ts:702`]

❌ 3. Update the existing Web API-wrapper test for the newly enforced runtime schema boundary and add the missing version-skew assertion. `api-wrappers.test.ts` globally resolves `apiMock.get` to `{}` and then awaits `meChats.listMeChats`; after this commit, `listMeChatsResponseSchema.parse({})` throws because `rows` and `nextCursor` are required, so that existing test cannot reach its request assertions. Seed a minimally valid old-server response (`rows`, `nextCursor`, no `priorityRows`) and assert that parsing supplies the empty priority default; that both repairs the current caller and proves the fallback this increment exists to provide. [R5 / `packages/web/src/api/__tests__/api-wrappers.test.ts:43`]

✅ 4. Deferring `counts.unread` is the correct resolution while its product scope is unsettled, and parsing the shared schema in the Web wrapper is the right place to make the remaining priority default real.

**Action taken**
- Submitted this follow-up request-changes review on the latest head; no code, tests, or Context Tree files were modified, and I did not run checks locally.

yuezengwu (changes requested):
## Re-review of `d27896ea`

**Goal and core change.** This PR moves the Workspace list to `activity_at` keyset ordering and adds first-page, whole-matching-set priority projections: viewer-scoped Attention, then private Pinned chats, while keeping ordinary `rows` additive for a safe server-first rollout. The latest increment correctly defers the unsettled unread aggregate and parses the shared schema at the Web API boundary.

**Recommendation: request changes.** This is B-high because it changes the shared chat-list response structure and the SQL/status projection behind a core Workspace surface. There is no migration or DB-schema file in this PR, but the shared response shape and database query semantics need human scrutiny.

### Blocking findings

1. **Recover only a recognized legacy cursor; keep malformed/unsupported cursors invalid.** `decodeCursor` returns the same `null` for the deployed two-part legacy shape, malformed base64/text, truncated `v2`, and unknown future versions; `listMeChats` then treats every one as page 1. This silently removes the existing invalid-cursor 400 contract and lets arbitrary bad input trigger the unbounded first-page priority work. Classify valid-v2 / recognized legacy / invalid separately. Recover the exact old shape (`<valid ISO or empty>|<non-empty chatId>`), but retain `BadRequestError` for malformed or unsupported values. Cover all three outcomes. The cursor header comment also still says decode failure is rejected with 400, contradicting the new behavior. (`packages/server/src/services/me-chat.ts:63-95,636-642`)

2. **Match the canonical old/new-client split in the failure-candidate SQL.** `computeErrored` consumes `agent_presence.runtime_state = 'error'` only for an active session whose `runtime_state_at IS NULL`; once a per-chat stamp exists, the per-chat runtime is authoritative. The candidate SQL ORs the agent-global error into every active session regardless of the stamp. Thus one global error admits all of an upgraded agent's active chat history—even a chat stamped per-chat `idle` that the canonical resolver proves cannot fail—defeating the narrowing this commit was meant to provide on every recurring page-0 refresh. Gate the presence fallback on `acs.runtime_state_at IS NULL`, and add a regression for global error + stamped per-chat idle. (`packages/server/src/services/me-chat.ts:702-712`; canonical branch in `packages/server/src/services/agent-chat-status.ts:602-610`)

3. **Repair and extend the Web API-wrapper test.** The new runtime parse is correct, but `api-wrappers.test.ts` still globally resolves GETs to `{}` and then awaits `listMeChats`; parsing now throws before the request assertion. This is the current CI failure and reproduces locally. Seed a minimally valid older-server payload (`{ rows: [], nextCursor: null }`, with no `priorityRows`) and assert that parsing supplies `{ attention: [], pinned: [] }`. That both repairs the caller and proves the claimed version-skew fallback. (`packages/web/src/api/__tests__/api-wrappers.test.ts:41-47,215-223`; `packages/web/src/api/me-chats.ts:49-54`)

### Documentation reconciliation

The PR title/body still advertise `counts.unread`, first-page `counts`, the old broad managed-speaker candidate, and 12 priority tests, while this head removes `counts`, narrows the candidate, and has 13 tests. Update the review contract before merge so it describes the actual wire shape and latest tradeoffs.

### Verification

- Context checked: `system/cloud/chat/workspace-conversations.md` and `sessions-and-status.md`; the additive rollout and canonical viewer-scoped failure direction remain consistent.
- `me-chat-helpers.test.ts`: 14 passed.
- `me-chat-priority.test.ts` + `me-chats-http-e2e.test.ts`: 21 passed.
- `api-wrappers.test.ts`: 1 failed / 3 passed with the same Zod error as CI; CI Test is red for that failure. Lint & Type Check is green.
- No source, database, or Context Tree files were modified during this review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1738 (follow-up on `5c48e2fc`)

- Rationale: The increment correctly resolves the three prior blockers, but its legacy-cursor classifier still excludes one cursor shape emitted by the deployed server, so some already-open clients remain stranded across rollout.

**Risk level**: B-high
- Path baseline: core `packages/server` chat-list service and shared/Web response boundary.
- Semantic lift: pagination transition behavior and failed-attention projection are core Workspace list semantics.

**PR summary**
- Problem: Workspace users need action-required and privately pinned chats surfaced above stable real-work recency without breaking already-open clients or turning recurring list refreshes into full-history status scans.
- Approach: The server returns first-page viewer-scoped Attention and Pinned projections alongside additive recency rows, versions the activity cursor, narrows failure candidates before canonical enrichment, and normalizes older-server responses at the Web API boundary. The unread aggregate is now explicitly deferred.
- Impacted modules: member chat-list SQL/pagination, canonical status candidate projection, shared response schema, Web API parsing, and server/Web fixtures.

**Review findings**
❌ 1. Recognize the deployed null-timestamp legacy cursor too. Before this PR, `encodeCursor(null, chatId)` emitted the two-part payload `|<chatId>` for pages whose boundary was in the `last_message_at IS NULL` tail. The new legacy branch requires `tsPart` to be non-empty, so that real deployed cursor is classified `invalid` and returns 400 instead of taking the promised page-1 recovery path. Accept both exact old timestamp cases—empty or parseable—with a non-empty chat id, and add codec/service coverage for the empty-timestamp form. This is the rollout case the original versioning finding explicitly called out. [R5 / `packages/server/src/services/me-chat.ts:100`]

⚠️ 2. The updated body now matches the deferred unread contract, but the PR title still advertises `(attention/pinned/unread)`. Remove `unread` from the title so the review surface matches the wire contract being merged.

✅ 3. The three prior blockers are otherwise resolved on this head: invalid/unsupported cursors remain typed failures, the global presence-error fallback is gated on an absent per-chat runtime stamp with a regression, and the Web wrapper fixture now proves the `priorityRows` default against an older-server payload.

**Action taken**
- Submitted this follow-up request-changes review on `5c48e2fc`; no code, tests, database state, or Context Tree files were modified, and I did not run checks locally.

yuezengwu (changes requested):
## Re-review of `5c48e2fc`

**Goal and core change.** The PR establishes `activity_at` keyset recency plus first-page, viewer-scoped Attention and private Pinned projections, while ordinary `rows` remain additive for a safe server-first rollout. This head correctly resolves the previous malformed-cursor distinction, canonical old/new-client SQL branch, Web runtime parse/test, unread deferral, and PR-description/title drift.

**Recommendation: request changes.** One deployed cursor shape is still missing from the rollout recovery path. Risk remains B-high because this changes the core shared chat-list response and database-backed pagination/status projection; there is no migration or DB-schema change in this PR.

### Blocking finding

1. **Recognize the deployed null-timestamp legacy cursor.** The base implementation's `encodeCursor(null, chatId)` emits `|<chatId>` for a page boundary in the `last_message_at IS NULL` tail. The new two-part classifier requires `tsPart` to be truthy, so this genuine legacy cursor becomes `invalid` and returns 400 instead of restarting at page 1. That leaves exactly those already-open clients in the retry loop the compatibility path is meant to prevent. Accept both old timestamp cases—empty or valid legacy timestamp—with a non-empty chat id, and cover `|<chatId>` in both codec and service-level recovery tests. (`packages/server/src/services/me-chat.ts:100-104`; base encoder emitted an empty first part for `null`.)

### Coverage note

The new “global error + stamped per-chat idle stays in rows” integration test proves the final canonical output, but it does not prove candidate narrowing: the previous SQL also admitted the chat, then `resolveAgentChatStatuses` filtered it back to ordinary `rows`, so the same assertions would pass before this fix. Because the change exists to prevent recurring extra enrichment work, a regression that observes candidate selection / resolver invocation (or an extracted candidate predicate tested directly) would guard the actual performance contract.

### Verification

- Server cursor/helper + priority + service + HTTP suites: 73 passed.
- Web API-wrapper suite: 4 passed; the older-server `priorityRows` default is now exercised.
- `git diff --check`: clean.
- Context remains consistent with viewer-scoped canonical failure and additive rollout; no source, database, or Context Tree files were modified during this review.

yuezengwu (approved):
## Re-review of `8f909441`

**Goal and core change.** This PR moves the Workspace conversation list to stable real-work (`activity_at`) keyset recency and adds first-page, whole-matching-set priority projections: viewer-scoped Attention above private Pinned chats, with additive ordinary `rows` for a safe server-first rollout.

**Recommendation: approve.** The final increment closes the remaining rollout gap: the cursor classifier now recognizes both shapes emitted by the deployed legacy encoder—including `|<chatId>` for a null `last_message_at` boundary—while malformed and unsupported cursors retain the typed 400 path. The candidate SQL mirrors the canonical stamped/unstamped runtime split, and its regression now observes resolver call count so it guards the actual extra-enrichment cost rather than only the final row bucket.

The other review rounds are also resolved: additive `rows` preserve old-Web visibility; priority work is first-page-only and prefiltered; the Web API boundary executes the version-skew schema default; the unsettled unread aggregate is deferred; and the title/body match the final response contract. The paired Context PR appropriately remains draft until the priority-aware client ships.

**Core-structure notice.** This is still B-high because it changes shared `ListMeChatsResponse` and database-backed chat-list projection/pagination semantics. I found no migration or DB-schema file in this PR; human review should focus on the shared response contract and SQL/status boundary rather than a data migration.

### Verification

- Server cursor/helper + priority + service + HTTP suites: 73 passed.
- Web API-wrapper suite: 4 passed.
- `git diff --check`: clean.
- GitHub Lint & Type Check is green; approval does not wait on the still-running full Test job.
- No source, database, or Context Tree files were modified during review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1738 (CI follow-up on `8f909441`)

- Rationale: The production cursor fix is correct, but the new performance-regression guard fails in the full CI suite and is not observing the resolver call path it claims to guard.

**Risk level**: B-high
- Path baseline: core server chat-list projection and shared response behavior.
- Semantic lift: the failed-attention candidate boundary controls recurring first-page enrichment cost.

**PR summary**
- Problem: Workspace users need viewer-scoped Attention and private Pinned chats above stable real-work recency without breaking deployed cursors or repeatedly enriching the caller's full chat history.
- Approach: This head correctly recognizes both deployed legacy cursor shapes and attempts to guard candidate narrowing by counting calls to the canonical status resolver.
- Impacted modules: chat-list cursor decoding, attention-candidate projection, and server integration tests.

**Review findings**
❌ 1. Repair the candidate-narrowing regression so it observes the actual production seam. Full CI fails in `me-chat-priority.test.ts` because `resolveStatusesSpy` sees **0** calls, while `enrichMeChatRows` unconditionally calls its own imported `resolveAgentChatStatuses` for every non-empty ordinary page. That proves the test's mocked module binding is not the binding used by `me-chat.ts`; the `toHaveBeenCalledTimes(1)` assertion therefore neither passes nor guards candidate selection. Do not weaken it to zero. Make the test observe the real candidate boundary—for example, factor the attention-candidate query into a testable helper and assert the stamped-idle chat id is absent, or establish a verified injection/mock seam and assert the ordinary call occurs while no priority call contains that chat. Keep a positive control so the observation mechanism itself cannot silently report zero. [R5 / `packages/server/src/__tests__/me-chat-priority.test.ts:18,429`]

✅ 2. The production null-timestamp cursor change itself is correct: both `<iso>|<chatId>` and `|<chatId>` recover as legacy, while malformed/unsupported values remain invalid.

**Action taken**
- Submitted request changes after the related full-CI failure on `8f909441`; no local checks were run and no code, database, or Context Tree files were modified.

yuezengwu (changes requested):
## Follow-up review on `8f909441`

**Recommendation: request changes**

This PR's goal and core shape remain sound: it moves the member chat list to `activity_at` keyset ordering and adds first-page, viewer-scoped `attention` / `pinned` projections while keeping `rows` additive for rollout compatibility. The affected contract is the shared `ListMeChatsResponse` plus DB-backed list/status projection; this head adds no database schema or migration change.

### Blocker

The candidate-narrowing regression guard does not observe the resolver binding used by production `me-chat.ts`. In the full CI suite, `resolveStatusesSpy` records **0** calls at `packages/server/src/__tests__/me-chat-priority.test.ts:429`, even though this case has a non-empty ordinary page and `enrichMeChatRows` unconditionally calls its statically imported `resolveAgentChatStatuses` for that page. The same test passes in focused isolation, so the result is module-loading/order dependent rather than a reliable observation of whether the priority candidate query admitted the stamped-idle chat.

Please move the assertion to a production seam that the test can prove it observes—for example, extract the attention-candidate selection into a testable helper and assert that the stamped-idle chat id is absent, or establish a verified injection/mock seam and include a positive control showing the ordinary enrichment call is captured while no priority enrichment contains that chat. Do not weaken the expected count to zero; that would preserve a silent no-op guard.

The production cursor fix itself remains correct: both deployed legacy shapes (`<iso>|<chatId>` and `|<chatId>`) recover to page 1, while malformed/unsupported cursors remain typed failures.

### Verification

- GitHub full Test job: 1 failure / 2,268 tests, exactly the zero-call assertion above.
- The earlier focused server run passed this file, which demonstrates the isolation dependency rather than validating the seam.
- Lint & Type Check, Migrations, CodeQL, and the remaining checks are green.

No code, database state, or Context Tree files were modified during this review.

yuezengwu (approved):
## Follow-up review on `8bfed02e`

**Recommendation: approve**

The PR's core goal remains the `activity_at`-ordered member chat list with first-page, viewer-scoped Attention / Pinned projections and additive recency rows for rollout compatibility. This increment cleanly fixes the remaining test blocker without changing that production contract.

The module-order-dependent resolver spy is gone. `listMeChats` now calls the extracted `selectAttentionCandidateRows` production helper, and the real-Postgres regression invokes that same helper directly. Its failed-session inclusion is a valid positive control, while the stamped-idle and healthy exclusions directly guard the candidate boundary; an always-empty or disconnected observation seam can no longer pass vacuously. The extracted SQL predicate and the production filter/order inputs are unchanged from the prior head.

The affected surface is still the internal DB-backed list/status projection and shared chat-list response; this increment adds no database schema or migration change. The helper/type exports remain internal to the server source module and are not added to the package export map.

### Verification

- `me-chat-priority.test.ts`: 15/15 passed against real Postgres.
- `@first-tree/server` typecheck passed.
- `git diff --check 8f909441..8bfed02e` passed.
- CI was still running at review time; approval does not wait on CI per review policy.

No code, database state, or Context Tree files were modified during this review.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1738

**Recommendation**: approve
- Rationale: All rollout, response-contract, recurring-query-cost, canonical-status, and test-observability blockers are resolved on `8bfed02e`; the full required CI suite is green.

**Risk level**: B-high
- Path baseline: core server member-chat list service and DB-backed projection.
- Semantic lift: shared chat-list response, keyset transition behavior, and viewer-scoped failed-attention semantics.

**PR summary**
- Problem: Workspace users need chats requiring their action and their private pins above stable real-work recency, without losing chats during a server-first rollout or repeatedly enriching the full managed-agent history.
- Approach: Return first-page whole-matching-set Attention and Pinned projections while keeping ordinary rows additive; order by `activity_at`; version the cursor and recover both deployed legacy shapes; prefilter possible failures using the canonical stamped/unstamped inputs; parse the shared response at the Web boundary; defer the unresolved unread aggregate.
- Impacted modules: server chat-list pagination/projection, shared response schema, Web API normalization, and server/Web fixtures.

**Review findings**
✅ 1. Both deployed legacy cursor shapes (`<iso>|<chatId>` and `|<chatId>`) recover from page 1; malformed/unsupported cursors remain typed failures.
✅ 2. Attention remains viewer-scoped and canonically resolved; the recurring candidate query excludes stamped-idle and healthy managed-speaker chats while retaining a failed positive control.
✅ 3. The candidate regression now exercises the exact production helper against real Postgres, eliminating the module-order-dependent mock seam that failed the prior full suite.
✅ 4. Additive rows, first-page priority gating, Web version-skew parsing, unread deferral, PR contract prose, and paired Context direction are consistent.

**Verification**
- GitHub Test, Lint & Type Check, Portable CLI Smoke, Migrations, and all CodeQL jobs passed on `8bfed02e`.
- Independent reviewer approval is present.
- Per review policy, I did not run local checks or modify code, database state, or Context Tree files.

**Action taken**
- Approved PR #1738 with explicit B-high authorization from baixiaohang. No merge action was taken.

PR #1739, opened 2026-07-13
fix(web): use full viewport height for iOS standalone

Gandy2025:
## Summary

- keep `100dvh` for ordinary mobile browser tabs where the browser chrome changes the visible height
- override the mobile shell to `100vh` only in WebKit standalone display mode
- add a source-level regression test that locks the media-query scope and cascade order

## Context

On the reported iPhone 14 Pro / iOS 26.5 home-screen Web App, the mobile shell can end above the physical bottom edge and expose a Home Indicator-sized strip after keyboard interaction.

[WebKit 254868](https://bugs.webkit.org/show_bug.cgi?id=254868) documents that installed apps using `viewport-fit=cover` report dynamic viewport heights without safe-area insets and explicitly recommends switching standalone mode back to `100vh`, which includes that space. [WebKit 292603](https://bugs.webkit.org/show_bug.cgi?id=292603) also documents that root overflow locks, overscroll containment, and scroll clamping do not suppress WebKit's extra bottom region; #1734 tried that path and was reverted by #1737 after physical-device QA made the symptom consistently reproducible.

This change therefore corrects the standalone shell height without touching document scrolling, adding global listeners, or changing ordinary browser behavior. It does not claim to eliminate every form of WebKit root overscroll; physical-device QA remains the acceptance gate for the reported layout gap.

## Validation

- mobile viewport and shell tests: 6 passed
- `pnpm check`
- `pnpm typecheck`
- Web production build passed
- built CSS verified to emit the WebKit standalone `100vh` rule after the `100dvh` rule

## QA

After staging deploy on iPhone 14 Pro / iOS 26.5:

1. fully close and reopen the home-screen First Tree app
2. open Chat and focus the composer
3. dismiss the keyboard, return to Now, and scroll to both boundaries
4. confirm the bottom navigation reaches the physical bottom safe area without an extra strip

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -719,6 +719,18 @@
     }
   }
 
+  /* Installed WebKit apps with viewport-fit=cover report dynamic viewport
+     heights without the safe-area insets (WebKit 254868). There is no browser
+     chrome to track in standalone mode, so use the full viewport height there
+     and keep the 100dvh behavior for ordinary mobile browser tabs. */
+  @supports (-webkit-touch-callout: none) {
+    @media (display-mode: standalone) {
+      .h-dvh-screen {
+        height: 100vh;
+      }
+    }
+  }
+
   /* Reserve the top safe-area inset (notch / status bar) so a full-height
      shell without its own top chrome does not render content under the
      status bar on notched devices. */
--- a/packages/web/src/pages/mobile/__tests__/shell-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/shell-dom.test.tsx
@@ -115,14 +115,14 @@ describe("MobileShell", () => {
     expect(harness.container.textContent).not.toContain("Current team");
   });
 
-  it("sizes the live shell to the dynamic viewport and reserves the top safe area", async () => {
+  it("uses the mobile viewport utility and reserves the top safe area", async () => {
     renderShell(harness, "/m/now");
     await harness.flush();
 
-    // Dynamic-viewport height comes from the .h-dvh-screen utility (100vh
-    // fallback + 100dvh override), not a raw inline 100vh that overflows the
-    // visible viewport on mobile browsers while the address bar is shown. The
-    // top safe-area inset is reserved via .pt-safe-top so content does not
+    // Viewport height comes from the .h-dvh-screen utility: dynamic viewport
+    // sizing for browser tabs and the WebKit standalone 100vh override live in
+    // CSS, rather than an inline height that cannot distinguish display mode.
+    // The top safe-area inset is reserved via .pt-safe-top so content does not
     // render under the status bar on notched devices.
     const shell = harness.container.querySelector(".h-dvh-screen");
     expect(shell).not.toBeNull();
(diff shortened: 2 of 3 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1739

**Recommendation**: approve
- Rationale: The standalone-only cascade correction directly addresses the WebKit safe-area height mismatch while preserving dynamic viewport sizing in ordinary mobile browser tabs.

**Risk level**: C
- Semantic lift: none; this changes mobile shell layout behavior and source-level CSS regression coverage, without touching subscriptions, WebSocket frame parsing, or the data layer.

**PR summary**
- Problem: On the iOS home-screen app, keyboard interaction can leave the mobile shell ending above the physical bottom edge and expose a Home Indicator-sized strip.
- Approach: Keep `100dvh` for ordinary mobile browser tabs, then override the shared mobile-shell utility to `100vh` only when WebKit is running in standalone display mode; retain physical-device staging QA as the acceptance gate.
- Impacted modules: `packages/web/src/index.css` and `packages/web/src/pages/mobile/__tests__/**`.

**Review findings**
✅ 1. The override is ordered after the dynamic-viewport rule and is jointly scoped by WebKit feature detection and standalone display mode, so ordinary browser tabs retain the existing behavior.
✅ 2. WebKit 254868 documents the same installed-app `viewport-fit=cover` height mismatch and identifies `100vh` as the standalone workaround that includes the safe-area inset.
✅ 3. The replacement does not reintroduce the root overflow locks, global listeners, or scroll clamps reverted in #1737, avoiding coexistence with the physically disproved workaround.
✅ 4. Regression coverage anchors the selector, media-query order, and `100vh` declaration; physical-device staging QA remains the correct validation boundary for the engine-specific symptom.

**Action taken**
- Approved.

yuezengwu (approved):

## Goal and core change

This PR fixes the iOS home-screen app shell ending above the physical bottom edge after keyboard interaction. It preserves `100dvh` for ordinary mobile browser tabs and adds a later `100vh` override only for WebKit standalone display mode.

## Findings

- The source rule is correctly nested under both WebKit feature detection and `display-mode: standalone`, and it follows the existing `100dvh` rule in the same cascade layer. The production CSS preserves that exact order and nesting.
- This matches WebKit 254868's documented `viewport-fit=cover` standalone height mismatch and its `100vh` workaround. It also avoids the root overflow locks, overscroll containment, listeners, and scroll clamping reverted in #1737 after physical-device QA disproved that approach.
- The mobile shell continues to own one bounded, border-box full-height layout with explicit safe-area padding. Ordinary browser behavior is unchanged by the standalone-only media query.
- The source-level regression test anchors the relevant rule ordering, selector, and declaration. Physical-device staging QA remains the acceptance gate for the engine-specific symptom.
- The diff is limited to Web CSS and tests; there are no database/schema, core data-structure, subscription, WebSocket, or shared-data changes requiring separate human inspection.

Risk level: C (`packages/web/**` layout behavior only).

Local validation at `b7b707e1`:

- `git diff --check`: passed
- Web Vitest suite: 172 files / 1,334 tests passed
- `pnpm check`: passed (pre-existing warnings outside this diff remain)
- shared package build and Web production build: passed
- generated CSS verified as `100vh` fallback → `100dvh` supported override → WebKit standalone `100vh` override

Physical-device staging QA on the reported iPhone 14 Pro / iOS 26.5 remains required before treating the visual symptom as resolved.

PR #1740, opened 2026-07-13
feat(web): collapse mobile team cards to two lines

Gandy2025:
## What

Collapse each mobile **Team** roster card from three lines to two.

**Before** (3 lines): name / `Public agent` / `● Online`
**After** (2 lines): name / `● Online · Public`

- Merge presence + visibility onto one secondary line.
- Drop the redundant `agent` / `Human` words — the `Agents` / `Humans` section headers already convey the kind.
- Compose the presence dot locally from `presenceChipView` + `StatusGlyph` (non-mono, muted label with a colored dot) instead of the shared `PresenceChip`, so desktop surfaces that use `PresenceChip` (Team / Settings / Computers) are **untouched**.

Net effect: denser, calmer roster on mobile.

## Scope

Single file: `packages/web/src/pages/mobile/team.tsx`. No shared-component, API, or type changes.

## Verification

- Rendered the real component against staging data via headless CDP at a mobile viewport (`/m/team`) — two-line layout renders correctly.
- `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) ✅
- `biome check` ✅
- `pnpm --filter @first-tree/web test` — 172 files / 1334 tests ✅
- Two independent reviews (correctness + UX/a11y) — no blocking defects.

## Known, pre-existing (out of scope here)

- The whole card is a `Chat with <name>` link whose `aria-label` overrides the subtree, so screen readers don't announce the secondary line. This behavior is **identical before and after** this change (the old subtitle/meta were suppressed the same way); flagging as a candidate for a separate a11y follow-up.

### DIFF
--- a/packages/web/src/pages/mobile/team.tsx
+++ b/packages/web/src/pages/mobile/team.tsx
@@ -7,7 +7,8 @@ import { listMembers } from "../../api/members.js";
 import { useAuth } from "../../auth/auth-context.js";
 import { Avatar } from "../../components/avatar.js";
 import { Input } from "../../components/ui/input.js";
-import { PresenceChip, runtimeStateToPresence } from "../../components/ui/presence-chip.js";
+import { presenceChipView, runtimeStateToPresence } from "../../components/ui/presence-chip.js";
+import { StatusGlyph } from "../../components/ui/status-glyph.js";
 import { formatRelative } from "../../lib/utils.js";
 import { MobilePage, MobileSection, MobileSystemState, mobileCardStyle } from "./components.js";
 
@@ -112,7 +113,13 @@ export function MobileTeamPage() {
 }
 
 function MobileAgentRow({ agent }: { agent: Agent }) {
-  const subtitle = `${agent.visibility === "private" ? "Private" : "Public"} agent`;
+  // Fold reachability and visibility into a single secondary line: a colored
+  // presence dot (blue online / muted offline) plus a muted, non-mono
+  // "Online · Private" caption. Dropping the "agent" suffix — the row already
+  // sits under the "Agents" section header — and the standalone presence line
+  // takes each card from three lines to two.
+  const presence = presenceChipView(runtimeStateToPresence(agent.runtimeState));
+  const visibility = agent.visibility === "private" ? "Private" : "Public";
   return (
     <MobileTeamRow
       avatar={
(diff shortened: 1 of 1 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1740

**Recommendation**: approve
- Rationale: The change makes the mobile Team roster denser without changing identity, presence, visibility, navigation, or shared desktop presentation contracts.

**Risk level**: C
- Path baseline: only `packages/web/src/pages/mobile/team.tsx` is changed -> C.
- Semantic lift: none; this is local presentation composition and does not touch subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Mobile users currently spend unnecessary vertical space scanning Team roster cards because member type, presence, and visibility occupy redundant lines.
- Approach: Keep each member name on the first line and combine the useful secondary signals on one line, while relying on the existing Agents/Humans section headings for member type. Agent presence continues to use the canonical presence mapping, but the dot and label are composed locally so desktop surfaces remain unchanged.
- Impacted modules: mobile Team page only.

**Review findings**
✅ 1. Presence and visibility retain their existing derivation and labels, while the shared `PresenceChip` and all desktop consumers remain untouched. [`packages/web/src/pages/mobile/team.tsx`]
✅ 2. The mobile Team surface continues to present humans and agents as separate, legible roster groups and preserves the existing chat target behavior. [`packages/web/src/pages/mobile/team.tsx`]
✅ 3. The colored status glyph is decorative and the adjacent text carries the status label, so the new local composition does not introduce a duplicate accessible announcement. [`packages/web/src/pages/mobile/team.tsx`]

**Action taken**
- Approved.

yuezengwu (approved):

The change achieves the intended mobile-only density improvement: each Team row keeps the teammate name on the first line and consolidates the useful secondary signals on the second line. Agent presence still derives from `runtimeStateToPresence` and `presenceChipView`; the local `StatusGlyph` is decorative while the adjacent text carries the status and visibility labels. Human/agent grouping, visibility semantics, and chat targets are unchanged, and shared `PresenceChip` consumers on desktop are untouched.

There are no API, type, core data-structure, or database changes. The existing whole-card `aria-label` behavior remains pre-existing and is not worsened by this patch.

Verified locally:
- `pnpm --filter @first-tree/web typecheck` (including design-token guardrails)
- `pnpm exec biome check packages/web/src/pages/mobile/team.tsx`
- `pnpm --filter @first-tree/web test` — 172 files / 1334 tests
- `git diff --check`

PR #721, opened 2026-06-01
Disable legacy MCP config writes

baixiaohang:
## Summary
- Web/CLI/Server close the legacy `mcpServers` write entry points.
- Runtime consumption of `payload.mcpServers` is preserved for existing resolved config paths.
- Skill discovery/upload reporting remains enabled; daemon now skips stale local aliases that are not pinned to this client before uploading skills.
- `GET/PATCH /api/v1/agents/:uuid/skills` stays available.

## Tests
- `pnpm --dir packages/server exec vitest run src/__tests__/admin-agent-config.test.ts src/__tests__/agent-skills.test.ts`
- `pnpm --dir packages/web exec vitest run src/pages/__tests__/page-ssr-smoke.test.tsx src/pages/__tests__/web-dom-interactions.test.tsx src/pages/agent-detail/__tests__/use-config-draft.test.tsx`
- `pnpm --dir apps/cli exec vitest run src/__tests__/agent-config-cli.test.ts`
- `pnpm --dir apps/cli exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/runtime-provider-reconcile.test.ts src/__tests__/agent-config-cli.test.ts`
- `pnpm --dir packages/server exec vitest run src/__tests__/agent-skills.test.ts`
- `pnpm exec biome check apps/cli/src/__tests__/agent-config-cli.test.ts apps/cli/src/commands/agent/config/add-mcp.ts packages/server/src/__tests__/admin-agent-config.test.ts packages/server/src/services/config-service.ts packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx packages/web/src/pages/agent-detail/__tests__/use-config-draft.test.tsx packages/web/src/pages/agent-detail/tools-tab.tsx packages/web/src/pages/agent-detail/use-config-draft.ts`
- `pnpm exec biome check apps/cli/src/core/runtime-provider-reconcile.ts apps/cli/src/core/index.ts apps/cli/src/commands/daemon/start.ts apps/cli/src/__tests__/runtime-provider-reconcile.test.ts apps/cli/src/__tests__/daemon-start-command.test.ts`

## Reviewer note
- codex-reviewer has reviewed and approved the original MCP write-shielding change.
- The only noted behavior is that an expired `expectedVersion` can return 409 before the MCP-specific rejection; this is acceptable optimistic-lock behavior.
[...]

PR #1746, opened 2026-07-13
feat(web): workspace chat-list rail — priority groups, Pin, filter redesign (PR4)

Gandy2025:
## Workspace chat-list refactor — PR4: desktop rail (priority groups, Pin, filter redesign)

Fourth of the 5-PR chat-list refactor. Builds on PR3 (#1738, server ordering + `priorityRows` projection). Web-only.

### What this adds to the workspace left rail
- **Server priority groups.** Consumes the page-0 `priorityRows` projection to render **Needs attention** (viewer-scoped failed / open-request) and **Pinned** groups above the ordinary recency buckets. `rows` is additive, so each priority chat is de-duplicated out of the recency list and rendered exactly once. Order: Needs attention → Pinned → Today / Older.
- **Pin / Unpin** row action (`pinMeChat`) in the per-row "Manage chat" menu, with a success/error toast and list invalidation. (The optimistic reorder is deferred to PR5.)
- **Reworked ⚙ filter popover**, per the tree #721 rail-filter contract:
  - **Status** — exclusive Active / Archived / All radios (Active is the default + reset target).
  - **Source** — non-empty additive Human / GitHub / Agent (default all = unrestricted; no zero-source state; the full set normalizes back to the unrestricted wire so "all sources" never counts as narrowing).
  - **Participants** — optional additive OR-picker over the org's addressable identities (empty = no constraint).
  - The trigger badge counts active filter **dimensions** (not selected values), so narrowing Source never makes the count drop. **Reset** clears the popover's own three dimensions and leaves the header triad (All / Unread / Watching — a separate control) untouched.
- **"Time" → "Recent activity"** grouping label.

### Testing
- `pnpm check` + `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) + full web suite (**1343** tests) green.
- New / updated DOM coverage: Pinned group renders + dedups (count-asserted); the empty-state / loading / load-more gates key off whether **any** row renders (an all-priority list must not paint "No conversations yet" over a populated group); Needs-attention dedup (count-asserted); the ⚙ badge dimension-count driven by the real component; Pin/Unpin toasts; the reworked filter (exclusive Status, non-empty Source, Participants OR-picker + error/retry state).
- browse·QA on `/preview/conversation-list`: the three groups render, Pin/Unpin works, and the filter popover renders all three sections + Reset.

### Review
Two independent adversarial reviews of the final diff: no BLOCKER / HIGH / MED. The HIGH empty-state gate found mid-development is fixed and regression-proof; dedup / priority-consumption memos are version-skew-safe; the participants roster fetch is genuinely lazy (mounts with the popover); the Pin mutation is idempotent.

### Follow-ups (not in this PR)
- **Participants picker at scale:** the current flat list is capped at the first 100 with no search. The app's other participant pickers already use a search typeahead (`useOrgAgentsSearch`); a dedicated PR replaces this section with the same pattern (search + selected chips), scaling to any org.
- PR5: realtime pin sync, mobile parity, a11y polish (focus trap, group labels).

Paired Context Tree PR: agent-team-foundation/first-tree-context#724 (kept draft until this merges, then reconciled).


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1746

- Rationale: The priority rendering is structurally sound, but the recency grouping and full-source URL compatibility still violate the rail contracts this PR claims to implement.

**Risk level**: B-low
- Path baseline: `packages/web/**` and a shared barrel export -> C
- Semantic lift: the Web change consumes the chat-list data projection and mutates/invalidate its query family -> B-low

**PR summary**
- Problem: Workspace users need high-priority conversations to remain visible across pagination, private pins to be manageable from the rail, and filters whose visible defaults match the actual result set.
- Approach: Render server-projected Needs attention and Pinned tiers ahead of recency buckets, de-duplicate their rows from the ordinary stream, add Pin/Unpin actions, and redesign Status / Source / Participants as immediate URL-backed filters.
- Impacted modules: shared chat-list exports; Web chat API; Workspace rail grouping, filters, row actions, URL transitions, preview, and DOM tests.

**Review findings**
❌ 1. The newly labeled “Recent activity” buckets still derive their day from `lastMessageAt`, while the server orders ordinary rows by `activityAt` and the paired contract explicitly treats a genuine description change as real work. A chat with an old last message but a description updated today will therefore be sorted near the top by the server and then placed under a collapsed “Older” bucket in the client, hiding the activity the rail is meant to surface. Please bucket on `activityAt` (with an intentional skew fallback if needed) and cover the description-only activity case. [R1/R5 / packages/web/src/pages/workspace/conversations/group-rows.ts:120]

❌ 2. Full-source normalization only happens while toggling the new checkboxes. `parseOriginList` still returns all three values for an existing/shared `?origin=manual,github,agent` URL, so `popoverFilterCount` reports one active dimension and the rail renders three filter chips even though the server result is semantically unrestricted. This URL shape could be produced by the previous UI and conflicts with the stated full-set-is-unrestricted contract. Please canonicalize the complete valid set at the URL/parser boundary (and defensively in the transition helper) and add a compatibility test. [R1/R5 / packages/web/src/pages/workspace/index.tsx:77]

✅ 3. The priority groups are consumed from page 0, kept attention-before-pinned, and de-duplicated from every loaded ordinary page; the empty/load-more gates correctly account for priority-only result sets.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Goal and core change: this PR makes the desktop Workspace rail consume the server's whole-matching-set priority projection (Attention, then private Pinned chats, then ordinary recency), adds Pin/Unpin, and replaces the filter popover with explicit Status / Source / Participants semantics. The priority-row consumption and exactly-once rendering are structurally sound. This PR is Web-only; the shared-package change only re-exports the already-existing priority-row schema/type, with no DB or migration change.

Requesting changes for two contract regressions:

1. `packages/web/src/pages/workspace/conversations/group-rows.ts:120` still assigns recency buckets from `lastMessageAt`, while the server orders the ordinary stream by `activityAt` and the new UI calls this grouping "Recent activity." A description-only update today (or a newly created empty chat, whose `lastMessageAt` is null) can be near the top of the server stream but land in the default-collapsed **Older** bucket. Use `activityAt` with an intentional old-server fallback such as `activityAt ?? lastMessageAt`, and add a regression that separates the two timestamps. Please also reconcile the visible row age at `conversations/index.tsx:591`: after the bucket fix, showing an old message age inside a Today activity bucket remains contradictory unless that distinction is deliberately represented.

2. `packages/web/src/pages/workspace/index.tsx:77` does not normalize an explicit full source set. The previous checkbox UI could produce shared URLs like `?origin=manual,github,agent`; this parser returns all three values, so the new rail shows three chips and a badge count of 1 even though the result is semantically unrestricted. Normalize the complete valid set to the unrestricted `[]` state at the parser boundary, canonicalize/delete it defensively in `nextParamsForOrigin`, and cover the legacy/shared-URL case.

Non-blocking follow-up: `ParticipantsSection` checks `agentsQuery.isError` before using cached `data`. TanStack Query 5 sets `isRefetchError` and `isError` while retaining data after a background poll failure, so a transient 30-second refetch error replaces a usable roster with the error panel. Reserve the blocking error panel for the no-data/initial-load case, or continue rendering stale options with a quiet retry indicator.

Validation on `630fe7fb`: full Web suite 1351/1351 passed; `pnpm --filter @first-tree/web typecheck`, token lint, `pnpm check`, and `git diff --check` passed.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1746

**Recommendation**: approve
- Rationale: The follow-up commit resolves both requested contract regressions and the cached-roster edge case without disturbing the priority projection or exactly-once rendering.

**Risk level**: B-low
- Path baseline: `packages/web/**` and a shared barrel export -> C
- Semantic lift: the Web rail consumes the chat-list data projection and mutates/invalidates its query family -> B-low

**PR summary**
- Problem: Workspace users need action-required and privately pinned conversations to remain visible across pagination, while Status / Source / Participants filters must expose defaults that match the actual result set.
- Approach: Render server-projected Attention and Pinned tiers before the ordinary activity stream, de-duplicate each chat to one rendered location, add Pin/Unpin, and make the three secondary filter dimensions explicit and URL-backed.
- Impacted modules: shared chat-list exports; Web chat API; Workspace rail grouping, filters, row actions, URL transitions, preview, and DOM tests.

**Review findings**
✅ 1. The recency contract is now coherent end-to-end: `rowActivityInstant` drives both the day bucket and visible row age, uses `activityAt` to match server ordering, and deliberately falls back to `lastMessageAt` for old-server skew. The regression separates description activity from message time.

✅ 2. A complete Source set now canonicalizes to unrestricted at both URL parsing and transition boundaries, including reordered/duplicated legacy URLs, so the badge and chips no longer claim a narrowing that does not exist.

✅ 3. A background participant-roster refetch failure now retains usable cached options, while an initial/no-data failure still exposes the retry state.

✅ 4. The existing server priority consumption remains attention-before-pinned-before-recency, de-duplicated across loaded pages, with priority-aware loading, empty, and load-more gates.

**Action taken**
- Approved the updated head `bc0b28e3`.

yuezengwu (approved):
## Re-review: approve

This PR makes the workspace rail keep attention-required and privately pinned conversations visible ahead of the ordinary activity stream, while making Status / Source / Participants filters accurately reflect their URL-backed semantics.

The updated head resolves both previously requested contract regressions:

- `rowActivityInstant` now drives both recency bucketing and the visible row age from `activityAt`, with an explicit `lastMessageAt` fallback for old-server skew. The tests cover description-only activity and fallback behavior.
- A complete valid Source set now canonicalizes to unrestricted at both the URL parser and transition boundary, including reordered and duplicated legacy URLs, so it no longer produces a badge or chips.

The follow-up also preserves a usable cached participant roster when a background refetch fails, while keeping the initial/no-data retry state. The priority tiers and exactly-once de-duplication behavior remain unchanged.

Validation on `bc0b28e3`: 83/83 focused Web tests passed, Web typecheck and token lint passed, repository checks exited successfully, and `git diff --check` passed.

No database migration, schema change, or new core data-structure contract is introduced by this Web-only follow-up.

PR #1743, opened 2026-07-13
feat(web): full-width mobile pickers for Chat slash + new-chat

Gandy2025:
## What & why

Extend #1741's phone-only picker dock from the Chat `@`mention to the other **cleanly-dockable** composer pickers, so mobile is consistent:

- **`/` slash-command popover** (Chat composer) — welds like `@`.
- **NewChatDraft recipient input** (`@`) — full-width dock above the message row, **not** corner-welded (a chip row sits above the field, so there's no clean top edge).

**AskTakeover is deferred** (not in this PR): its answer surface lives in an `overflow:auto` scroller that clips an upward-docked picker — hiding the first/active candidates (a wrong-selection hazard). Escaping the clip needs the picker to be portaled out, which conflicts with the CSS-descendant weld, so it's a separate structural follow-up. Its mention picker is untouched here (keeps the pre-existing float).

## How

Generalize the #1741 weld into an **opt-in `.composer-input` host marker**:

- `@media (max-width: 47.999rem) { .composer-input .mention-popover, .composer-input .slash-popover { …full-width dock… } }` — a picker docks only when its host opts in (a future consumer can't silently inherit it).
- The Chat composer (clean top edge, `.composer-card`) sets `data-picker-open="true"` to flatten its top corners while a **mention or slash** picker is visible; NewChat sets only `.composer-input` (no flatten).
- The `/` popover moved its inline/Tailwind box styling into a `.slash-popover` class sharing the `.mention-popover` desktop base; rows get `.slash-option`.
- `composerPickerVisible({ isTrial, mentionOpen, slashOpen })` centralizes the trial guard — a live trial `/` (slash hook stays up while the panel is suppressed) must not weld an empty input. The chat composer is its only welding caller today.

**Desktop presentation is unchanged** (the new behavior lives entirely in the phone media query; `.slash-popover` reproduces the prior slash styling verbatim).

## Rebase

Rebased onto current `main` — **#1742** (touch-optimize mobile chat composer) and **#1744** (touch-optimize mobile new-chat/ask composers). My weld targets the unchanged outer composer-card / popover-wrapper anchors; the touch-optimizations are internal (tap targets, one-line textarea) and don't move the picker's dock point. `ask-takeover.tsx`, its DOM test, and `center/index.tsx` have **zero diff** here — #1742/#1744's work is preserved.

## Verification

[...]

PR #1747, opened 2026-07-13
feat(web): portal the AskTakeover mobile @picker out of the scroll clip

Gandy2025:
## What & why

Follow-up to #1743 (which deferred AskTakeover). AskTakeover's answer field lives inside an `overflow-y:auto` card, so an in-flow `@`mention panel docked above the field is **clipped** — hiding the first/active candidates while Enter still selects a hidden row (a wrong-selection hazard, reproduced at 390×844).

## How

Opt-in **`portal`** mode on `MentionAutocompletePopover`: it renders into a `document.body` portal with `position: fixed`, positioned from the field's `getBoundingClientRect()`, so it **escapes the clip**. Full field width, welded flush above the field with the same top-only clipped shadow as the in-flow docks. Same idiom as `ui/popover.tsx` / `ui/hover-card.tsx`.

Positioning lifecycle (hardened over review):
- **`requestAnimationFrame` loop** while open re-measures every frame, so it follows *any* anchor movement — inner-scroller scroll, keyboard, resize, and coordinate-only reflow (attachment tray growing above the field) that scroll listeners + `ResizeObserver` miss. setState bails on an unchanged rect; the loop stops on close/dismiss.
- **Visibility bounded by the real clipping scrollport** (`nearestScrollableAncestor`), not the window. When the field scrolls fully out of its scrollport the picker **dismisses** (`onDismiss` → the mention hook's `dismiss`), not just hides — so a hidden panel is never Enter/Tab-selectable.
- **Height clamped to the visible space above the field** (incl. `visualViewport.offsetTop`) so the panel's first/active row never renders above the viewport top; **dismisses** when there isn't room for even one row.
- **`z-index: 40`** — above the ask scrim (30), below the Radix modal layer (dialog / command palette, 50), so Cmd+K over an active `@` sits above the picker.

`AskTakeover` passes `portal` only on phone widths (`viewport === "narrow"`); wider viewports keep the in-flow float. The field welds via `.ask-answer-field` + `data-picker-open` (`composerPickerVisible` → no empty weld on a trial `@`). **Chat/NewChat pickers untouched** (`portal` defaults false). **Desktop unchanged.**

## Verification

- typecheck + lint:tokens clean, `biome` clean (only the pre-existing `index.css` `noImportantStyles` warning), **1353 web tests pass**.
- New deterministic coverage: pure `fieldOutOfScrollport` / `portalPanelPlacement` geometry tests, plus **portal DOM lifecycle tests** (`mention-autocomplete-portal-dom.test.tsx`): out-of-scrollport → no listbox + Enter selects nothing; coordinate-only move → position follows next frame; near-viewport-top → height clamped; too-tight → dismissed.
- Rendered the shipped CSS + positioning at 390px in **light and dark**: full candidate list visible (clip escaped), welded, no dark seam.

## Scope

`packages/web` — shared picker component + AskTakeover host + picker styles + tests. No data/schema/routing changes.

### DIFF
--- a/packages/web/src/components/__tests__/mention-autocomplete-portal-dom.test.tsx
+++ b/packages/web/src/components/__tests__/mention-autocomplete-portal-dom.test.tsx
@@ -1,0 +1,203 @@
+// @vitest-environment happy-dom
+
+import { act, useEffect, useRef, useState } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { MentionAutocompletePopover, type MentionCandidate, useMentionAutocomplete } from "../mention-autocomplete.js";
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+
+/**
+ * Portal-mode lifecycle contracts (AskTakeover's clip-escaping `@` picker),
+ * guarded against regression:
+ *   1. a field scrolled fully out of its inner scrollport dismisses the picker,
+ *      so Enter can't commit a hidden candidate;
+ *   2. a coordinate-only field move (no scroll/resize/size change) is followed
+ *      on the next animation frame;
+ *   3. near the viewport top the panel height is clamped to the space above the
+ *      field (so its first/active row never renders off-screen), and dismisses
+ *      when there isn't room for even one row.
+ *
+ * happy-dom has no layout, so element rects, the scrollport lookup, and rAF are
+ * stubbed deterministically.
+ */
+
+type RectLike = { top: number; bottom: number; left: number; right: number; width: number; height: number };
+
+const twoCandidates: MentionCandidate[] = [
+  { agentId: "a1", name: "alice", displayName: "alice", managedByMe: false },
+  { agentId: "a2", name: "bob", displayName: "bob", managedByMe: false },
+];
+
+let harnessCandidates: MentionCandidate[];
+let picked: Array<{ text: string; cursor: number }>;
+let fieldRect: RectLike;
+let portRect: RectLike;
+let rafCb: FrameRequestCallback | null;
+
+function Harness() {
+  const taRef = useRef<HTMLTextAreaElement>(null);
+  // Re-render once after mount so `open` recomputes with the textarea ref
+  // attached (mirrors a real `@` keystroke re-render).
+  const [, force] = useState(0);
+  useEffect(() => force(1), []);
+  const mention = useMentionAutocomplete({
+    value: "@",
+    cursor: 1,
+    candidates: harnessCandidates,
+    onSelect: (u) => picked.push(u),
+  });
+  return (
+    <div data-scroll style={{ overflowY: "auto" }}>
+      <div data-field>
+        <MentionAutocompletePopover
+          trigger={mention.trigger}
+          results={mention.results}
+          highlightIndex={mention.highlightIndex}
+          anchorRef={taRef}
+          onPick={mention.pick}
+          portal
+          onDismiss={mention.dismiss}
+        />
+        <textarea data-ta ref={taRef} onKeyDown={(e) => mention.handleKey(e)} />
+      </div>
+    </div>
+  );
+}
+
+function mount(): { container: HTMLElement; root: Root } {
+  const container = document.createElement("div");
+  document.body.appendChild(container);
+  const root = createRoot(container);
+  act(() => root.render(<Harness />));
+  return { container, root };
+}
+
+const listbox = () => document.body.querySelector<HTMLElement>('[role="listbox"]');
+const px = (v: string) => Number.parseInt(v, 10);
+const flushFrame = () => {
+  const cb = rafCb;
+  rafCb = null;
+  if (cb) act(() => cb(0));
+};
+
+describe("MentionAutocompletePopover portal lifecycle", () => {
+  beforeEach(() => {
+    harnessCandidates = twoCandidates;
+    picked = [];
+    fieldRect = { top: 400, bottom: 440, left: 20, right: 340, width: 320, height: 40 };
+    portRect = { top: 100, bottom: 500, left: 0, right: 360, width: 360, height: 400 };
+    rafCb = null;
+    vi.stubGlobal("requestAnimationFrame", (cb: FrameRequestCallback) => {
+      rafCb = cb;
+      return 1;
+    });
+    vi.stubGlobal("cancelAnimationFrame", () => {
+      rafCb = null;
+    });
+    Object.defineProperty(window, "innerHeight", { value: 800, configurable: true });
+    vi.spyOn(HTMLElement.prototype, "getBoundingClientRect").mockImplementation(function (this: HTMLElement) {
+      if (this.hasAttribute("data-field")) return fieldRect as DOMRect;
+      if (this.hasAttribute("data-scroll")) return portRect as DOMRect;
+      return { top: 0, bottom: 0, left: 0, right: 0, width: 0, height: 0 } as DOMRect;
+    });
+    const realGetComputedStyle = window.getComputedStyle.bind(window);
+    vi.spyOn(window, "getComputedStyle").mockImplementation((el: Element) => {
+      if (el.hasAttribute?.("data-scroll")) return { overflowY: "auto" } as CSSStyleDeclaration;
+      return realGetComputedStyle(el);
+    });
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.unstubAllGlobals();
+    document.body.innerHTML = "";
+  });
+
+  it("renders the portal listbox docked above the field when it is in view", () => {
+    const { root } = mount();
+    expect(listbox()).not.toBeNull();
+    expect(px(listbox()?.style.bottom ?? "")).toBe(400); // innerHeight(800) - field.top(400)
+    act(() => root.unmount());
+  });
+
+  it("dismisses when the field is out of its scrollport, and Enter then selects nothing", () => {
+    fieldRect = { ...fieldRect, top: 40, bottom: 90 }; // fully above the scrollport top (bottom <= 100)
+    const { container, root } = mount();
+
+    expect(listbox()).toBeNull();
+
+    const ta = container.querySelector<HTMLTextAreaElement>("[data-ta]");
+    act(() => {
+      ta?.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true, cancelable: true }));
+    });
+    expect(picked).toHaveLength(0);
+    act(() => root.unmount());
+  });
+
+  it("follows a coordinate-only field move on the next animation frame", () => {
+    const { root } = mount();
+    expect(px(listbox()?.style.bottom ?? "")).toBe(400);
+
+    fieldRect = { ...fieldRect, top: 300, bottom: 340 }; // moved up 100; no scroll/resize/size change
+    flushFrame();
+
+    expect(px(listbox()?.style.bottom ?? "")).toBe(500); // 800 - 300
+    act(() => root.unmount());
+  });
+
+  it("clamps the panel height to the space above the field near the viewport top", () => {
+    // Scrollport reaches the viewport top; field sits 60 below it — in the
+    // scrollport, but with only 60 of usable space above.
+    portRect = { ...portRect, top: 0 };
+    fieldRect = { ...fieldRect, top: 60, bottom: 100 };
+    const { root } = mount();
+    expect(listbox()).not.toBeNull();
+    expect(px(listbox()?.style.maxHeight ?? "")).toBe(60); // clamped, not the 16rem cap
+    act(() => root.unmount());
+  });
+
+  it("dismisses when there isn't room above the field for even the active row", () => {
+    portRect = { ...portRect, top: 0 };
+    fieldRect = { ...fieldRect, top: 30, bottom: 70 }; // 30 < one row (46)
+    const { container, root } = mount();
+    expect(listbox()).toBeNull();
+
+    const ta = container.querySelector<HTMLTextAreaElement>("[data-ta]");
+    act(() => {
+      ta?.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true, cancelable: true }));
+    });
+    expect(picked).toHaveLength(0);
+    act(() => root.unmount());
+  });
+
+  it("re-scrolls the active row into view when a geometry-only clamp shrinks the panel", () => {
+    // Eight candidates, scrollport from the viewport top so the field can move
+    // near the top and shrink the clamp.
+    harnessCandidates = Array.from({ length: 8 }, (_, i) => ({
+      agentId: `a${i}`,
+      name: `n${i}`,
+      displayName: `n${i}`,
+      managedByMe: false,
+    }));
+    portRect = { ...portRect, top: 0 };
+    const scrollSpy = vi.spyOn(HTMLElement.prototype, "scrollIntoView").mockImplementation(() => {});
+    const { container, root } = mount();
+    const ta = container.querySelector<HTMLTextAreaElement>("[data-ta]");
+
+    // Highlight a later row (7th) — field is far from the top, panel is tall.
+    for (let i = 0; i < 6; i++) {
+      act(() => ta?.dispatchEvent(new KeyboardEvent("keydown", { key: "ArrowDown", bubbles: true, cancelable: true })));
+    }
+    scrollSpy.mockClear();
+
+    // Coordinate-only move toward the viewport top → maxHeight clamps 256 → 60,
+    // highlight index unchanged. The active row must be re-scrolled into view.
+    fieldRect = { ...fieldRect, top: 60, bottom: 100 };
+    flushFrame();
+
+    expect(px(listbox()?.style.maxHeight ?? "")).toBe(60);
+    expect(scrollSpy).toHaveBeenCalled();
+    act(() => root.unmount());
+  });
+});
--- a/packages/web/src/components/__tests__/mention-autocomplete.test.ts
+++ b/packages/web/src/components/__tests__/mention-autocomplete.test.ts
@@ -5,8 +5,10 @@ import {
   type CandidateDivider,
   composerPickerVisible,
   detectMentionTrigger,
+  fieldOutOfScrollport,
   groupAndSortCandidates,
   type MentionCandidate,
+  portalPanelPlacement,
   rankCandidates,
 } from "../mention-autocomplete.js";
 
@@ -376,5 +378,55 @@ describe("composerPickerVisible", () => {
   it("never welds on the trial composer, even with a live trigger — trial renders no panel", () => {
     expect(composerPickerVisible({ isTrial: true, mentionOpen: false, slashOpen: true })).toBe(false);
     expect(composerPickerVisible({ isTrial: true, mentionOpen: true, slashOpen: true })).toBe(false);
+  });
+});
+
+describe("fieldOutOfScrollport", () => {
+  const port = { top: 100, bottom: 500 };
+
+  it("is in view while the field overlaps the scrollport", () => {
+    expect(fieldOutOfScrollport({ top: 400, bottom: 440 }, port)).toBe(false);
+    // partially above the top still counts as in view (some rows visible)
+    expect(fieldOutOfScrollport({ top: 80, bottom: 130 }, port)).toBe(false);
+    // partially below the bottom still counts as in view
+    expect(fieldOutOfScrollport({ top: 470, bottom: 520 }, port)).toBe(false);
+  });
+
+  it("is out of view when fully above the scrollport top (dismiss)", () => {
+    expect(fieldOutOfScrollport({ top: 40, bottom: 90 }, port)).toBe(true);
+    // touching edge (bottom === port.top) counts as out
+    expect(fieldOutOfScrollport({ top: 60, bottom: 100 }, port)).toBe(true);
+  });
+
+  it("is out of view when fully below the scrollport bottom (dismiss)", () => {
+    expect(fieldOutOfScrollport({ top: 540, bottom: 580 }, port)).toBe(true);
+    // touching edge (top === port.bottom) counts as out
+    expect(fieldOutOfScrollport({ top: 500, bottom: 560 }, port)).toBe(true);
+  });
+});
+
+describe("portalPanelPlacement", () => {
+  const port = { top: 0, bottom: 800 };
+
+  it("clamps max height to the space above the field, capped at 16rem (256)", () => {
+    expect(portalPanelPlacement({ field: { top: 60, bottom: 100 }, port, viewportTop: 0 })).toEqual({ maxHeight: 60 });
+    // plenty of room above → capped at the 16rem panel max
+    expect(portalPanelPlacement({ field: { top: 400, bottom: 440 }, port, viewportTop: 0 })).toEqual({
+      maxHeight: 256,
+    });
+  });
+
+  it("subtracts the visual-viewport top offset from the available space", () => {
+    expect(portalPanelPlacement({ field: { top: 120, bottom: 160 }, port, viewportTop: 40 })).toEqual({
+      maxHeight: 80,
+    });
+  });
+
+  it("dismisses (null) when the field is out of its scrollport", () => {
+    expect(portalPanelPlacement({ field: { top: 850, bottom: 890 }, port, viewportTop: 0 })).toBeNull();
+  });
+
+  it("dismisses (null) when there isn't room above the field for even one row", () => {
+    expect(portalPanelPlacement({ field: { top: 30, bottom: 70 }, port, viewportTop: 0 })).toBeNull();
   });
 });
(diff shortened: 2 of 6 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1747

- Rationale: Portaling removes the ancestor clip, but the current positioning lifecycle still allows the picker to remain visible or selectable outside the Ask scrollport and to detach from the field after ordinary in-card reflow.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**` -> C
- Semantic lift: none; this does not touch subscriptions, WebSocket parsing, or the data layer

**PR summary**
- Problem: Phone users answering a blocking question can have the first and active `@mention` candidates clipped by the Ask card, while keyboard selection still targets those hidden rows.
- Approach: Add an opt-in body portal for the shared mention popover, anchor it above the Ask answer field with fixed coordinates, and keep the existing in-flow picker for wider Ask views and all other hosts.
- Impacted modules: AskTakeover answer field, shared mention autocomplete popover, and mobile picker styling

**Review findings**

❌ 1. Use the Ask scrollport, not the browser window, as the visibility boundary and keep visibility atomic with keyboard selection. The comment says the portal hides when the field leaves the card viewport, but `r.bottom <= 0 || r.top >= window.innerHeight` only detects leaving the layout viewport. When the inner `overflowY: "auto"` region scrolls the field above its own top edge while it is still within window coordinates, the body portal remains rendered over unrelated card/scrim content. In the opposite case where `fixedRect` becomes null, `useMentionAutocomplete` remains open, so Enter/Tab can still select a result while the panel is hidden—the same hidden-active-row hazard this PR is intended to remove. Please derive/pass the clipping scrollport boundary and make a non-visible picker non-selectable or dismiss it. [R4/R5 / `packages/web/src/components/mention-autocomplete.tsx:604`; clipping ancestor at `packages/web/src/components/chat/ask-takeover.tsx:583`]

❌ 2. Reposition when the field moves without changing size. `ResizeObserver` observes the field's border box, so it does not fire for a pure coordinate change. AskTakeover can cause exactly that while the picker stays open: adding/removing the attachment tray or expanding content above the field moves the unchanged field inside the scroller without a scroll/resize event. `fixedRect` then retains the old `bottom`, leaving the portal detached from or overlapping the answer field until some unrelated viewport event occurs. Please observe a layout boundary whose size actually changes or otherwise remeasure on these in-card reflows. [R4 / `packages/web/src/components/mention-autocomplete.tsx:624`]

**Action taken**
- Submitted request changes on head `4ed8d981`.

yuezengwu (changes requested):
## Review result: request changes

### Goal and core change

This PR fixes the phone-width AskTakeover mention picker being clipped by the card's `overflow-y: auto` region. It adds an opt-in body portal to `MentionAutocompletePopover`, anchors that fixed panel above the Ask answer field, and keeps Chat, NewChat, and wider Ask views on the existing in-flow path. There are no data, schema, routing, or core-structure changes.

The portal direction is appropriate, but the positioning/visibility lifecycle still has three blocking edge cases.

### Blockers

1. **Use the Ask scrollport as the visibility boundary, and make visibility atomic with keyboard selection.**

   `packages/web/src/components/mention-autocomplete.tsx:604` checks only `r.bottom <= 0 || r.top >= window.innerHeight`. The answer field can be fully above the Ask card's inner `overflowY: "auto"` scrollport while still having ordinary positive window coordinates, so the body portal remains visible over unrelated card/scrim content. Conversely, when `fixedRect` becomes `null`, `useMentionAutocomplete` remains open: Enter/Tab still commits a candidate even though no listbox is rendered. I reproduced both states with a narrow-viewport DOM probe; the hidden case changed `@` to `@alice ` on Enter. That is the same hidden-active-row/wrong-selection hazard this PR is meant to remove. Please derive/pass the clipping scrollport boundary and make a non-visible picker non-selectable or dismiss it.

2. **Remeasure when the field moves without resizing.**

   `packages/web/src/components/mention-autocomplete.tsx:624` observes only the field's border box. `ResizeObserver` does not fire for coordinate-only movement, but AskTakeover can move the unchanged field while the trigger remains open—for example, adding/removing the attachment tray or expanding an option preview above it. No window scroll/resize is required, so `fixedRect.bottom` can remain stale and the portal detaches from or overlaps the field. Observe a layout boundary that changes for those reflows or otherwise schedule a post-layout remeasure.

3. **Keep the picker below higher modal layers, or close/inert it when one opens.**

   `.mention-popover--portal` uses `z-index: 60`, while the shared Radix dialog overlay/content use `z-50`. The global Cmd/Ctrl+K shortcut remains active at narrow widths, and mention openness is text/cursor-driven rather than blur-driven, so opening the command palette with an active `@` leaves this body portal mounted above the palette. The Ask card's existing `aria-hidden` guard prevents its own reply shortcuts, but it does not visually hide this higher-z body portal; stale mention options can cover and receive pointer input over the modal. Preserve the established higher-overlay ownership by lowering the picker below the modal layer or explicitly closing/inerting it.

Please add deterministic coverage for the scrollport-out-of-view + Enter/Tab case and a coordinate-only reflow. The current pure mention tests and AskTakeover DOM tests do not exercise portal mode.

### Verification

- Targeted AskTakeover + mention tests: 58 passed.
- Web typecheck and design-token guard: passed.
- Biome on the three changed files: passed with only the pre-existing `index.css` `noImportantStyles` warning.

Gandy2025:
Thanks @yuezengwu / @baixiaohang — all three blockers addressed in the latest head.

**1. Scrollport boundary + atomic visibility/selection.** The visibility boundary now derives the actual clipping scrollport (`nearestScrollableAncestor`, the ancestor whose `overflow-y` clips) instead of the window, via a testable pure helper `fieldOutOfScrollport(fieldRect, portRect)`. When the field is fully out of its scrollport the picker now **dismisses** (`onDismiss` → the mention hook's `dismiss`, wired from AskTakeover), not just hides — so `useMentionAutocomplete` closes and Enter/Tab can't commit a hidden row.

**2. Coordinate-only reflow.** Replaced the scroll/resize listeners + `ResizeObserver` with a `requestAnimationFrame` loop while open — it re-measures every frame, so it tracks *any* anchor movement including a pure coordinate change (attachment tray / option preview growing above the field) with no scroll/resize/size event. setState bails on an unchanged rect, so idle frames don't re-render; the loop stops on close/dismiss.

**3. z-index vs the modal layer.** Lowered `.mention-popover--portal` to `z-index: 40` — between the ask scrim (30) and the Radix modal layer (dialog / command palette, 50), so a Cmd+K palette opened over an active `@` sits above the picker (and takes focus, so the mention can't select).

**Tests:** added deterministic `fieldOutOfScrollport` unit tests (fully above / fully below / partial-overlap / touching-edge). The rAF positioning + real stacking remain layout-dependent (not jsdom-unit-testable); verified in a real browser (light + dark) and, as with the original clip finding, best confirmed by your real-Chrome pass.

typecheck + lint:tokens clean, biome clean (pre-existing warning only), 1344 web tests pass.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1747 follow-up

- Rationale: The update resolves the three reported lifecycle blockers, but the portal can still place the first/active rows above the browser viewport while the field partially overlaps the Ask scrollport, and the PR delivery record still describes the superseded implementation.

**Risk level**: C
- Path baseline: changes remain limited to `packages/web/**` -> C
- Semantic lift: none; this does not touch subscriptions, WebSocket parsing, or the data layer

**PR summary**
- Problem: Phone users answering a blocking question need every visible `@mention` choice—especially the active choice selected by Enter—to remain reachable outside the Ask card's clipping scroller.
- Approach: Portal the Ask picker to `document.body`, continuously anchor it to the answer field, dismiss it when the field leaves the Ask scrollport, and keep higher modal layers above it.
- Impacted modules: AskTakeover answer field, shared mention autocomplete popover and tests, and mobile picker styling

**Review findings**

❌ 1. Bound the portal panel itself to the visible viewport, not only the field to the Ask scrollport. `fieldOutOfScrollport` intentionally keeps the picker open for any partial field overlap, while the portal always welds its bottom to `r.top` and CSS still permits a 16rem panel. The new test's own partial-overlap case (`field.top = 80`, `field.bottom = 130`, `port.top = 100`) therefore renders a full six-row panel from roughly `-176px` to `80px`: the first and initially active rows are above the browser viewport, yet Enter/Tab still select them. A fully visible field near the scrollport top has the same problem, and a keyboard-shortened phone card makes that geometry more likely. Please clamp the panel's max height/placement to the available visual space above the field (including the visual viewport top), or dismiss atomically when there is not enough usable panel area to keep the active row visible. [R4/R5 / `packages/web/src/components/mention-autocomplete.tsx:558`, `:664`; `packages/web/src/index.css:977`]

❌ 2. Update the PR body to match the reviewed head before merge. It still records the superseded `z-60` stacking, scroll/resize/`visualViewport` listeners plus a field `ResizeObserver`, hide-only behavior, 1341 tests, and “no blockers.” Head `b0c575c1` instead uses `z-index: 40`, an rAF measurement loop, scrollport-driven dismissal, and the author reports 1344 tests after addressing three blockers. The current body would leave the durable PR/squash audit trail describing behavior that no longer ships. [R5 / PR body]

✅ 3. The prior three blockers are otherwise addressed coherently: the nearest clipping scrollport governs visibility and calls `mention.dismiss`, rAF measurement covers coordinate-only reflow with unchanged-state bailout and cleanup, and `z-index: 40` restores Radix dialog ownership over the Ask picker.

**Action taken**
- Submitted request changes on head `b0c575c1`.

yuezengwu (changes requested):
## Rereview result: request changes

### Goal and updated implementation

The PR still has the right scope: escape AskTakeover's phone-width mention picker from the answer scroll clip without changing Chat, NewChat, desktop behavior, data, schema, or routing.

The rewritten head `b0c575c1` fixes the three runtime blockers from my previous review:

- visibility is now bounded by the actual Ask scrollport and leaving it calls `mention.dismiss`, so a hidden picker is no longer keyboard-selectable;
- the animation-frame measurement follows coordinate-only field movement;
- the portal is now `z-index: 40`, below the shared `z-50` dialog layer.

I independently exercised the first two paths with a narrow-viewport DOM probe: an out-of-scrollport field produced no listbox and Enter did not insert the hidden candidate; changing only the field's `getBoundingClientRect().top` updated the portal's `bottom` on the next frame.

### Remaining blocker: test the component lifecycle, not only the geometry helper

The new committed tests cover only `fieldOutOfScrollport` as a pure function. They still pass if any of the behavior that closes the original wrong-selection hole regresses—for example, if AskTakeover stops wiring `onDismiss`, the popover goes back to hiding without dismissing, or the animation-frame remeasurement is removed.

Please add deterministic DOM coverage for the two user-visible contracts:

1. In a narrow Ask, put the field fully outside its inner scrollport, open `@`, assert that the listbox is absent, then press Enter/Tab and assert that no candidate is inserted/selected.
2. Keep the picker open, change only the field coordinates without firing scroll/resize or changing its size, advance one animation frame, and assert that the portal position follows the field.

These are stable component behaviors and the repo's testing rule requires deterministic behavior to be guarded by product tests in the same PR. The implementation passes both probes today; the missing piece is preserving that result against regression.

### Verification

- Targeted mention + AskTakeover suite: 61 tests passed.
- Independent portal lifecycle probes: 2 passed.
- Full Web test suite: passed (existing warning/noise only).
- Web typecheck and design-token guard: passed.
- Biome on all changed files: passed with only the pre-existing `index.css` `noImportantStyles` warning.

yuezengwu (changes requested):
## Rereview addendum: request changes

I independently confirmed the additional viewport-boundary issue reported on head `b0c575c1`.

### Blocker: keep the panel and active row inside the visible viewport

`fieldOutOfScrollport` only answers whether the **field** intersects the Ask scrollport. The portal still welds its bottom to `fieldRect.top`, while CSS permits a `16rem`-tall panel. With six phone rows, a fully visible field at approximately `top: 100px` can therefore produce a panel spanning roughly `-156px..100px`; the initial active row is at the clipped top of that panel, but `useMentionAutocomplete` remains open and Enter/Tab selects it. The partial-overlap case committed in the new helper test makes the same geometry even easier to reach.

This recreates the exact hidden-active-row/wrong-selection class the PR is meant to close. Please constrain the panel itself to the available visual space above the field—including `visualViewport.offsetTop` when present—and ensure the active row is visible. If there is not enough usable height for a selectable row, dismiss atomically rather than leaving hidden keyboard selection active.

The component regression coverage requested in my previous review should include this top-of-viewport case with enough candidates to reach the max-height path.

### Delivery record: update the PR body

The PR body still describes the superseded implementation: `z-60`, scroll/resize/`visualViewport` listeners, a field `ResizeObserver`, hide-only behavior, 1341 tests, and “no blockers.” The reviewed head now uses `z-index: 40`, an animation-frame measurement loop, scrollport-driven dismissal, and 1344 tests after multiple requested changes. Please update the body before merge so the durable PR/squash record matches what actually ships.

The three earlier runtime blockers remain resolved; this addendum does not reopen them. There are still no data, schema, routing, or core-structure changes.

Gandy2025:
Thanks @baixiaohang / @yuezengwu — both remaining items addressed on the latest head.

**Viewport clamp + active-row visibility.** Added `portalPanelPlacement(field, port, viewportTop)`: it clamps the panel's `max-height` to the space above the field *within the visible viewport* (incl. `visualViewport.offsetTop`), so the panel's top (first/active) row never renders above the viewport top; and it **dismisses** (atomic with selection, via `mention.dismiss`) when the field is out of its scrollport *or* there isn't room for even one row. The panel bottom is welded to the field top, `max-height` set inline from the placement.

**Component lifecycle tests** (`mention-autocomplete-portal-dom.test.tsx`, happy-dom, stubbed rects/rAF):
- field out of its inner scrollport → **no listbox rendered, and Enter inserts nothing** (the exact hidden-active-row case);
- coordinate-only field move with no scroll/resize/size change → portal `bottom` follows on the **next animation frame**;
- field near the viewport top with limited space → panel `max-height` **clamped** to the available space (not the 16rem cap);
- not enough room for even one row → **dismissed** + Enter inserts nothing.

Plus pure `fieldOutOfScrollport` / `portalPanelPlacement` geometry tests (clamp, the 16rem cap, `visualViewport` offset, both dismiss thresholds).

**PR body** rewritten to the shipped head (rAF loop, scrollport-driven dismissal + viewport clamp, `z-index: 40`, 1353 tests).

typecheck + lint:tokens clean, biome clean (pre-existing warning only), 1353 tests pass.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1747 follow-up

- Rationale: The runtime fixes and delivery record are now coherent, but the new regression tests do not exercise the narrow AskTakeover integration seam that makes a hidden portal non-selectable.

**Risk level**: C
- Path baseline: changes remain limited to `packages/web/**` -> C
- Semantic lift: none; subscriptions, WebSocket parsing, and the data layer are untouched

**PR summary**
- Problem: Phone users answering a blocking question need the `@mention` picker to escape the Ask card's scroll clip without leaving an invisible active choice selectable by Enter or Tab.
- Approach: Portal the picker on narrow Ask views, continuously anchor it above the answer field, bound it to both the Ask scrollport and visual viewport, dismiss it when no selectable row can be shown, and keep higher modal layers above it.
- Impacted modules: AskTakeover answer field, shared mention autocomplete popover and tests, and mobile picker styling

**Review findings**

❌ 1. Add the requested narrow-Ask DOM regression test rather than testing only a self-contained popover harness. The new `Harness` wires `portal` and `onDismiss={mention.dismiss}` itself (`mention-autocomplete-portal-dom.test.tsx:37-61`), so its hidden-field/Enter test still passes if AskTakeover later drops either `portal={viewport === "narrow"}` or `onDismiss={mention.dismiss}`—the exact integration regression that would reopen hidden keyboard selection. The existing AskTakeover tests still exercise only the in-flow path and do not set field/scrollport geometry (`ask-takeover-dom.test.tsx:436-473`). Please extend the AskTakeover DOM coverage with a narrow viewport and deterministic rect/rAF stubs: open `@`, move the field fully outside its scrollport, assert the body-portaled listbox is absent, then press Enter or Tab and assert the candidate was not inserted. The generic coordinate-movement and viewport-height tests can remain where they are. [R4/R5 / `packages/web/src/components/__tests__/mention-autocomplete-portal-dom.test.tsx:37`; `packages/web/src/components/chat/ask-takeover.tsx:343`; `packages/web/src/components/chat/__tests__/ask-takeover-dom.test.tsx:436`]

✅ 2. The implementation-level blockers are resolved: the panel height respects `visualViewport.offsetTop`, insufficient space dismisses atomically, rAF measurement follows coordinate-only reflow, the Ask scrollport governs field visibility, and `z-index: 40` preserves dialog ownership.

✅ 3. The PR body now describes the shipping rAF lifecycle, scrollport dismissal, viewport clamping, modal stacking, and updated verification record.

**Action taken**
- Submitted request changes on head `0ce5cb2d`.

yuezengwu (changes requested):
## Rereview result: request changes

### Goal and updated implementation

This PR keeps the right scope: on phone-width AskTakeover, escape the `@mention` picker from the answer card's overflow clip while leaving Chat, NewChat, and wider Ask views on the existing in-flow path. There are no data, schema, routing, or core-structure changes.

Head `0ce5cb2d` closes the previously reported items coherently:

- the real Ask scrollport governs visibility, and leaving it dismisses the mention trigger;
- the animation-frame loop follows coordinate-only anchor movement;
- `z-index: 40` keeps the picker below the shared modal layer;
- panel height is now clamped against `visualViewport.offsetTop`, with atomic dismissal when one row cannot fit;
- the PR body now describes the reviewed implementation;
- the new portal DOM tests cover dismissal/no hidden Enter selection, coordinate-only movement, the initial top clamp, and the too-tight dismissal path.

### Remaining blocker: keep the current active row visible when the clamp shrinks

`MentionAutocompletePopover` still scrolls the active option into view only when `highlightIndex` changes (`packages/web/src/components/mention-autocomplete.tsx:653-657`). The animation-frame measurement can later reduce `fixedRect.maxHeight` when the field moves upward (`:699-705`) without changing that index.

A concrete phone path is: open a tall list, ArrowDown to a later candidate, then scroll the Ask region or let keyboard/card reflow move the field toward the viewport top. The overflow panel preserves its existing `scrollTop` while its viewport shrinks. The highlighted row that was visible in the taller panel can then sit below the new 46–60px viewport; the hook remains open, so Enter/Tab still commits that hidden active row. This is the same hidden-active-row/wrong-selection class the PR is intended to eliminate.

I reproduced this deterministically on `0ce5cb2d` by extending the new portal harness to eight candidates, highlighting the seventh row, moving only the field from `top: 400` to `top: 60`, and advancing the next frame. `maxHeight` correctly became `60px`, but no new `scrollIntoView` occurred because the highlight was unchanged.

Please make active-row visibility react to portal geometry changes as well as highlight changes—for example, re-scroll the active row after a `maxHeight`/placement update—and add the corresponding regression case: later highlighted row, available-height shrink, active row brought back into view before Enter/Tab can commit it.

### Verification

- Committed targeted Ask + mention suites: 70 tests passed.
- Full Web run during the added regression probe: all 1,353 committed tests passed; the additional active-row-shrink assertion failed as described above.
- Web typecheck and design-token guard: passed.
- Biome on all changed files: passed with only the pre-existing `index.css` `noImportantStyles` warning.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1747 addendum

- Rationale: Portal geometry can shrink around an unchanged later highlight without re-establishing the invariant that the keyboard-selected row is visible.

**Risk level**: C
- Path baseline: changes remain limited to `packages/web/**` -> C
- Semantic lift: none; subscriptions, WebSocket parsing, and the data layer are untouched

**PR summary**
- Problem: A phone user answering a blocking Ask must see the same `@mention` candidate that Enter or Tab will commit, even while the Ask field and available viewport move.
- Approach: The portal now follows the field, clamps to the visual viewport, dismisses when no row fits, and stays below modal layers; the remaining gap is preserving active-row visibility after a geometry-only height reduction.
- Impacted modules: shared mention autocomplete portal lifecycle and regression tests

**Review findings**

❌ 1. Re-scroll the active option when portal geometry reduces the visible panel, not only when `highlightIndex` changes. The current effect at `mention-autocomplete.tsx:653-657` has only `[highlightIndex]` as its dependency, while the rAF loop can update `fixedRect.maxHeight` and `fixedRect.bottom` at `:699-713` after the user has highlighted a later row. Shrinking an `overflow: auto` container does not make that unchanged active row visible; its preserved `scrollTop` can leave the highlighted/Enter-selected option below the new 46–60px viewport. Make active-row visibility react to the committed placement change (preferably before paint, without running on unchanged rAF frames), and add a DOM regression with a later highlighted candidate followed by a coordinate-only height shrink, asserting that the active row is brought into view before Enter/Tab can commit it. [R4 / `packages/web/src/components/mention-autocomplete.tsx:653`; `packages/web/src/components/mention-autocomplete.tsx:699`; `packages/web/src/components/__tests__/mention-autocomplete-portal-dom.test.tsx`]

✅ 2. The preceding runtime fixes remain sound: scrollport dismissal, visual-viewport clamping, insufficient-space dismissal, coordinate following, modal stacking, and the updated PR body all match the reviewed head.

**Action taken**
- Submitted an additional request-changes review on head `0ce5cb2d`.

Gandy2025:
Thanks @yuezengwu / @baixiaohang — both addressed.

**Active-row visibility on a geometry-only clamp.** The active-row `scrollIntoView` now depends on `fixedRect.maxHeight`, not just `highlightIndex`, and moved to `useLayoutEffect` so it re-establishes visibility **before paint**. When the rAF loop shrinks the panel (field moves toward the viewport top) with the highlight unchanged, the active row is re-scrolled into the smaller viewport, so Enter/Tab can't commit a row that's scrolled out of the clamped panel. The `maxHeight` dep runs it only on a committed clamp change, not on every frame (biome-ignore documents the intentional trigger dep).

**Tests:**
- `mention-autocomplete-portal-dom.test.tsx`: eight candidates, ArrowDown to a later row, then a coordinate-only move that clamps `maxHeight` 256→60 with the highlight unchanged → asserts the active row is re-scrolled (`scrollIntoView` called) after the shrink.
- `ask-takeover-dom.test.tsx` (the integration seam baixiaohang asked for): a **narrow AskTakeover** (forced `matchMedia`) with stubbed field/scrollport rects — typing `@` produces a **body-portaled** `.mention-popover--portal` listbox (guards `portal={viewport==="narrow"}`); moving the field out of its scrollport then dismisses it and **Enter inserts nothing** (guards `onDismiss={mention.dismiss}`). This fails if AskTakeover later drops either wiring.

typecheck + lint:tokens clean, biome clean (pre-existing warning only), **1355 web tests pass**.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1747 final rereview

**Recommendation**: approve
- Rationale: The latest head closes both remaining regression gaps and preserves the invariant that the keyboard-selected mention row is visible throughout the portal lifecycle.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**` -> C
- Semantic lift: none; subscriptions, WebSocket parsing, and the data layer are untouched

**PR summary**
- Problem: Phone users answering a blocking Ask need the `@mention` list to escape the card's scroll clip while always seeing the same candidate that Enter or Tab will commit.
- Approach: Portal only the narrow Ask picker, continuously anchor it to the field, bound it to the Ask scrollport and visual viewport, dismiss it when no row can be shown, keep modal layers above it, and re-establish active-row visibility after geometry changes.
- Impacted modules: AskTakeover answer field and DOM tests, shared mention autocomplete portal and tests, and picker styling

**Review findings**

✅ 1. Active-row visibility now reacts to committed portal clamp changes as well as highlight changes. The layout effect depends on `fixedRect.maxHeight`, so a coordinate-only shrink re-scrolls the unchanged later highlight before paint; the eight-candidate regression test covers that exact path.

✅ 2. The new AskTakeover DOM test exercises the previously missing host seam: it forces the narrow viewport, verifies a body-portaled picker, moves the answer field outside the actual scrollport, confirms dismissal, and confirms Enter does not insert the hidden candidate.

✅ 3. The earlier runtime blockers remain resolved: actual-scrollport visibility, atomic dismissal, coordinate-only rAF following, visual-viewport height clamping, insufficient-space dismissal, and modal-layer stacking all compose coherently. Chat, NewChat, and wider Ask paths remain on the existing in-flow behavior.

**Action taken**
- Approved head `5db58e90`.

yuezengwu (approved):
## Rereview result: approve

### Goal and final implementation

This PR fixes the phone-width AskTakeover `@mention` picker being clipped by the answer card's scrollport. It adds an opt-in body portal to the shared mention popover, keeps Chat/NewChat and wider Ask views on the existing in-flow path, follows field movement with an rAF measurement loop, dismisses atomically outside the real Ask scrollport or when one visible row cannot fit, clamps against the visual viewport, and preserves higher modal ownership with `z-index: 40`.

There are no data, schema, routing, or core-structure changes.

### Blockers resolved

The final head `5db58e90` closes both remaining review gaps:

- a committed portal `maxHeight` change now re-scrolls the unchanged active option in a layout effect, so geometry-only shrink cannot leave the Enter/Tab-selected row hidden;
- the AskTakeover DOM suite now exercises the real narrow integration seam and proves both `portal` and `onDismiss` stay wired: after the field leaves the inner scrollport, the body portal disappears and Enter does not insert the hidden candidate.

The preceding fixes remain coherent: scrollport-bounded dismissal, coordinate-only tracking, visual-viewport height clamping, too-tight atomic dismissal, and modal stacking.

### Verification

- Targeted mention + AskTakeover suites: 72 tests passed.
- Web typecheck and design-token guard: passed.
- Biome on all changed files: passed with only the pre-existing `index.css` `noImportantStyles` warning.
- A full local Web run hit the unrelated `mobile/density-dom` timing failure once; that file passed immediately in isolation. It does not touch this PR's paths or behavior.

Approved without waiting for the still-running GitHub Test job, per review policy.

PR #1748, opened 2026-07-13
feat(web): workspace chat-list — search-only Participants filter (PR4b)

Gandy2025:
## Workspace chat-list: search-only Participants filter (fast-follow to #1746)

Follow-up to PR4 (#1746). Web-only, one component.

### Why
The Participants filter dumped the org's first **100** addressable identities into a flat scrollable checkbox list. It didn't scale (the 101st identity was unreachable, and a large org is a wall to scroll) and it was **inconsistent with the app's own people pickers** — new-chat draft, the chat-header `[+]`, `@`-mention, and `add-participant-dropdown` all use the `useOrgAgentsSearch` typeahead.

### What
Replace the flat list with a **search-only** picker (no upfront roster):
- A debounced `useOrgAgentsSearch(term, { addressableOnly: true })` typeahead — reaches any org size, matching the app's other pickers.
- An empty query renders only a **"Type to search people."** hint — no roster is dumped.
- Current selections show as **removable chips** above the search (names via the shared identity map), so they stay visible and manageable without re-searching.
- Matches render as `FilterCheckbox` toggles; a `Searching…` state covers the debounce/fetch window; `No people match "…"` on an empty result.
- Selection semantics are unchanged — an additive OR-filter on `?with=`, canonical-sorted; the section still mounts only inside the open popover.

Status / Source / Reset are untouched.

### Testing
- `pnpm check` + `pnpm --filter @first-tree/web typecheck` (incl. `lint:tokens`) + full web suite (**1357**) green.
- Rewritten `filter-popover-dom` coverage: empty-query hint (no roster), typing filters, no-match, in-flight `Searching…`, select → chip, chip removal. `conversation-list-dom` mock swapped to `useOrgAgentsSearch`.
- browse·QA on `/preview/conversation-list`: the Participants section renders the search input + hint (the DEV preview has no backend, so live results are covered by the DOM tests).

### DIFF
--- a/packages/web/src/lib/participant-name-cache.ts
+++ b/packages/web/src/lib/participant-name-cache.ts
@@ -1,57 +1,47 @@
-import { useSyncExternalStore } from "react";
+import { type QueryClient, useQuery } from "@tanstack/react-query";
 
 /**
- * A tiny session-lifetime cache of participant `uuid -> displayName`, populated
- * from `useOrgAgentsSearch` result rows.
+ * A cache of participant `uuid -> displayName` learned from
+ * `useOrgAgentsSearch` result rows, held INSIDE react-query so it shares the
+ * org/auth lifecycle: a `queryClient.clear()` on organization switch or logout
+ * wipes it like every other cached query, and it is never a second,
+ * un-invalidated source of truth living at module scope.
  *
- * The Participants filter is search-only, so a viewer can select an identity
- * that sorts past the org-list 100-row first page — exactly who the at-scale
- * search exists to reach. `useAgentNameMap` only knows that first page, so
- * without this cache such a selection renders as a raw UUID on its filter chips
- * the moment it leaves the visible result rows: in the rail's persistent chip
- * row, and in the popover chip after the panel closes and reopens (the picking
- * component unmounts each time). This cache lives at module scope so a name
- * learned from any search survives those remounts and is shared by both chip
- * surfaces; it falls back to the identity map for names it has never seen.
+ * Its only job is a FALLBACK label for an identity the authoritative
+ * `useAgentNameMap` cannot resolve — one that sorts past the org-list 100-row
+ * first page, exactly who the at-scale participant search reaches. Chip
+ * consumers always PREFER the authoritative map (so a rename, which invalidates
+ * `["agents"]` and refreshes the map, immediately wins) and read this cache
+ * only for ids the map still returns unresolved.
  */
-const names = new Map<string, string>();
-const listeners = new Set<() => void>();
-// A fresh snapshot identity on every change so `useSyncExternalStore` re-renders;
-// stable between changes so `getSnapshot` never loops.
-let snapshot: ReadonlyMap<string, string> = new Map(names);
+const PARTICIPANT_NAME_CACHE_KEY = ["participant-name-cache"] as const;
+type NameMap = Readonly<Record<string, string>>;
 
-function emit(): void {
-  snapshot = new Map(names);
-  for (const listener of listeners) listener();
-}
-
-/** Record display names seen in a search result set. Idempotent + no-op if unchanged. */
-export function rememberParticipantNames(agents: ReadonlyArray<{ uuid: string; displayName: string }>): void {
-  let changed = false;
-  for (const agent of agents) {
-    if (agent.displayName && names.get(agent.uuid) !== agent.displayName) {
-      names.set(agent.uuid, agent.displayName);
-      changed = true;
+/** Record the display names seen in a participant search result set. */
+export function rememberParticipantNames(
+  queryClient: QueryClient,
+  agents: ReadonlyArray<{ uuid: string; displayName: string }>,
+): void {
+  queryClient.setQueryData<NameMap>(PARTICIPANT_NAME_CACHE_KEY, (prev) => {
+    const base = prev ?? {};
+    let next: Record<string, string> | null = null;
+    for (const agent of agents) {
+      if (agent.displayName && base[agent.uuid] !== agent.displayName) {
+        next = next ?? { ...base };
+        next[agent.uuid] = agent.displayName;
+      }
     }
-  }
-  if (changed) emit();
+    return next ?? base;
+  });
 }
 
-/** Test-only: clear the module cache so cases don't leak names into each other. */
-export function __resetParticipantNameCacheForTests(): void {
-  names.clear();
-  emit();
-}
-
-/** Subscribe to the cache; returns a resolver `uuid -> name | undefined`. */
+/** Subscribe to the fallback cache; returns a resolver `uuid -> name | undefined`. */
 export function useParticipantNames(): (uuid: string) => string | undefined {
-  const map = useSyncExternalStore(
-    (listener) => {
-      listeners.add(listener);
-      return () => listeners.delete(listener);
-    },
-    () => snapshot,
-    () => snapshot,
-  );
-  return (uuid: string) => map.get(uuid);
+  const { data } = useQuery<NameMap>({
+    queryKey: PARTICIPANT_NAME_CACHE_KEY,
+    queryFn: () => ({}),
+    staleTime: Number.POSITIVE_INFINITY,
+    gcTime: Number.POSITIVE_INFINITY,
+  });
+  return (uuid: string) => data?.[uuid];
 }
--- a/packages/web/src/lib/participant-name-cache.ts
+++ b/packages/web/src/lib/participant-name-cache.ts
@@ -1,0 +1,57 @@
+import { useSyncExternalStore } from "react";
+
+/**
+ * A tiny session-lifetime cache of participant `uuid -> displayName`, populated
+ * from `useOrgAgentsSearch` result rows.
+ *
+ * The Participants filter is search-only, so a viewer can select an identity
+ * that sorts past the org-list 100-row first page — exactly who the at-scale
+ * search exists to reach. `useAgentNameMap` only knows that first page, so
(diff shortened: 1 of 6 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1748

- Rationale: The search reaches identities beyond the first roster page, but stale results remain actionable during query transitions and labels for those newly reachable identities are lost as soon as the popover unmounts.

**Risk level**: B-low
- Semantic lift: changes the Workspace participant filter's React Query search/cache data flow -> B-low

**PR summary**
- Problem: Workspace users in larger organizations cannot reach participants beyond the first 100 identities, and scrolling a full roster is impractical.
- Approach: Replace the flat roster with a debounced server-side search, keep selected participants visible as removable chips, and preserve the existing canonical `?with=` OR-filter semantics.
- Impacted modules: Workspace filter popover, conversation-list DOM harness, filter DOM coverage, and the offline conversation-list preview.

**Review findings**
❌ 1. A settled non-empty result set remains interactive while the raw input has moved to a newer term or the newer request is fetching. The render branch checks `results.length > 0` before `searching`, so after searching for “Nova” and then typing “Design”, the input says “Design” while the old Nova checkbox can still be toggled for the full debounce/fetch window. The sibling participant pickers explicitly guard this stale-result window to prevent committing an identity that no longer corresponds to the visible query. Please either show the searching state first or disable/block result toggles until the rendered query matches the input, and cover the non-empty-previous-result case (the current lag test only exercises an empty prior result). [R4 / `packages/web/src/pages/workspace/conversations/filter-popover.tsx:482`]

❌ 2. The display-name bridge for identities beyond the first 100 has the lifetime of `ParticipantsSection`, even though that component deliberately unmounts whenever the popover closes. `useAgentNameMap` cannot resolve such an identity, so a participant found via the new at-scale search renders correctly only inside the currently open popover; the persistent rail chip already falls back to the raw UUID, and closing/reopening the popover loses the picked name there too. That breaks the “visible and manageable without re-searching” behavior in the exact >100-user scenario this PR targets. Please keep selected identity labels in a source whose lifetime spans popover mounts (or provide selected-ID lookup/cache integration), and cover close/reopen plus an identity absent from the first-page name map. [R3 / `packages/web/src/pages/workspace/conversations/filter-popover.tsx:385`]

✅ 3. The existing additive OR semantics and canonical sorting of `?with=` remain intact, consistent with the Workspace conversation projection contract.

**Action taken**
- Submitted request changes on head `01203023`.

yuezengwu (changes requested):
Requesting changes on head `01203023`. The move to server-side search is the right direction and the existing additive OR / canonical `?with=` semantics are preserved, but I independently confirmed the same two functional blockers already identified by baixiaohang:

1. **Stale search rows remain actionable.** In `filter-popover.tsx:482`, the non-empty `results` branch wins before `searching`. After a settled search for “Nova”, typing “Design” leaves Nova visible and toggleable through the debounce/fetch window even though the input shows the new term. Please either render `Searching…` first or disable/block stale toggles, and add coverage where the previous result set is non-empty.

2. **Names for identities beyond the first 100 are not retained across the actual filter UI lifetime.** `pickedNames` lives inside `ParticipantsSection`, which unmounts when the popover closes. Since `useAgentNameMap` only knows the first roster page, a newly reachable identity immediately renders as a raw UUID in the persistent rail chip (`index.tsx:432-436`) and also loses its popover-chip label after close/reopen. Please retain/search-cache selected identity labels outside the conditional section (or add selected-ID lookup integration), with a test covering an identity absent from the first-page map across close/reopen.

Local verification: `pnpm check`, Web typecheck/token lint, the 22 focused DOM tests, and the full Web suite completed successfully. CI is also green; these are uncovered interaction/lifetime gaps rather than compile/test failures. No database, schema, or core persistent-data-structure changes are involved.

baixiaohang (changes requested):
## Re-review by baixiaohang-code-reviewer - PR #1748

- Rationale: The follow-up closes both original interaction gaps, but the new module-scope name store is outside the existing identity-cache invalidation and auth/org lifecycle, so searched identities can now render stale labels indefinitely and survive tenant/session resets.

**Risk level**: B-low
- Semantic lift: adds a second session-level identity cache to the Workspace participant-filter data flow -> B-low

**Review findings**
❌ 1. `participant-name-cache.ts` creates a second, module-global source of truth for `uuid -> displayName`, but it is not invalidated with the authoritative agent queries and is not cleared when the app clears React Query on organization switch or logout. Both chip consumers then prefer `cachedName(...)` over `useAgentNameMap()`. A concrete result is: search/select an identity, rename that identity from Team/Agent Detail (those mutations correctly invalidate `queryKey: ["agents"]`), and the Workspace chips continue showing the old cached label until that identity happens to be searched again or the page reloads. The same store also survives `selectOrganization()` and `logout()`, even though both deliberately call `queryClient.clear()` to prevent previous-org/user data reuse. Please keep the beyond-page-1 label in the existing org/auth-scoped query/identity lifecycle, or explicitly scope/reset this fallback and let a refreshed authoritative identity override it; add coverage for the rename/invalidation and org/auth reset boundary. [R1, R5 / `packages/web/src/lib/participant-name-cache.ts:17`; `packages/web/src/pages/workspace/conversations/index.tsx:439`]

✅ 2. Stale non-empty search rows are now disabled during the debounce/fetch window, and the new regression exercises the previously uncovered branch.

✅ 3. A searched identity beyond the first-page name map now retains its label across popover close/reopen, and the persistent rail chip is wired to the same resolver.

✅ 4. The additive OR and canonical `?with=` semantics remain unchanged.

**Action taken**
- Submitted request changes on updated head `3976e449`.

yuezengwu (changes requested):
Requesting changes on head `3976e449`. The two findings from `01203023` are fixed: stale rows are now disabled, and a searched label survives popover close/reopen and feeds the persistent rail. One new blocker remains in the cache introduced to do that:

**The participant-name cache sits outside the existing user/org invalidation boundary and overrides fresher identity data.** `participant-name-cache.ts:17` is a process/module singleton keyed only by UUID. `AuthProvider` deliberately calls `queryClient.clear()` on both logout and `selectOrganization()` so one user/team cannot reuse another scope's cached data, but this store survives both transitions. A name learned under user/org A can therefore be rendered for a retained or revisited `?with=<uuid>` under user/org B. Within one org, `cachedName(id) ?? resolveAgentName(id)` (`filter-popover.tsx:402-403`, `index.tsx:439`) also means a cached search label wins forever over the polled authoritative roster: rename or visibility removal updates `useAgentNameMap`, but the chip continues showing the old name.

Please scope/clear this cache with the same user + organization transitions as the rest of the org data, and let a resolved current-roster name win (using the search cache only when the shared map falls back to the UUID). Add coverage for scope reset and for a fresh identity-map rename superseding a cached search label.

Local verification on this head: `pnpm check`, Web typecheck/token lint, and the 24 focused DOM tests pass; CI is green. No database or persistent schema changes are involved, but this new in-memory identity data structure does need the explicit tenant/freshness boundary above.

baixiaohang (approved):
## Re-review by baixiaohang-code-reviewer - PR #1748

**Recommendation**: approve
- Rationale: The latest follow-up closes the remaining identity-cache scope and freshness gap without changing the participant-filter contract.

**Risk level**: B-low
- Semantic lift: the Workspace participant filter now maintains a React Query-backed fallback identity cache -> B-low

**PR summary**
- Problem: Workspace users in larger organizations could not reach participants beyond the first 100 identities, while a flat roster was impractical to scan.
- Approach: Use debounced server-side search, preserve selected identities as removable chips, block stale result commits, and retain beyond-page-1 labels in an auth/org-scoped fallback cache while keeping the current identity map authoritative.
- Impacted modules: Workspace filter popover and rail chips, participant-name fallback cache, conversation-list preview, and DOM coverage.

**Review findings**
✅ 1. Search results from the previous term remain visible but are disabled through the debounce/fetch window, so a toggle cannot commit an identity that does not match the visible input.

✅ 2. Beyond-page-1 labels now live in the current React Query client rather than a module singleton, so the existing `queryClient.clear()` boundary on organization switch and logout removes them with the rest of the scoped data.

✅ 3. Both chip surfaces resolve the current identity map first and use the search-fed cache only when that map falls back to the UUID, so a refreshed authoritative rename supersedes an older search label.

✅ 4. The regressions cover stale non-empty results, close/reopen retention, authoritative-name precedence, and a fresh cache scope; the additive OR and canonical `?with=` semantics remain unchanged.

**Action taken**
- Approved updated head `f9351c08`.

yuezengwu (approved):
Approving head `f9351c08`.

The implementation goal remains sound: replace the non-scaling 100-row participant roster with search-only, debounced server-side lookup while preserving the additive OR / canonical `?with=` filter contract.

The latest follow-up resolves the remaining cache-lifecycle blocker:

- Beyond-page-1 display names now live in the existing React Query client, so the app's established `queryClient.clear()` boundary removes them on organization switch and logout.
- Both popover and persistent rail chips prefer the refreshed `useAgentNameMap` value and use the search-fed label only when the authoritative resolver falls back to the UUID, so a rename supersedes stale search data.
- The earlier stale-result interaction fix remains intact: previous rows stay disabled while the visible query is settling.
- Regressions cover stale non-empty results, close/reopen retention, authoritative rename precedence, and a fresh cache scope.

Local verification: `pnpm check`, Web typecheck/token lint, and the full Web suite (1363 tests) pass. I also exercised the mounted-client `queryClient.clear()` path followed by the organization-driven rerender; the cached label is removed as intended. No database or persistent schema changes are involved.

PR #727, opened 2026-06-01
feat(web): redesign agent profile Usage tab

yuezengwu:
## Summary

Redesigns the agent profile Usage tab. Drops the KPI strip + 7d/30d window picker + cramped single-row 90-cell grid. New layout has two sections only:

- **Activity (last 90 days)** — GitHub-style contribution calendar with hover tooltips (weekday + date + input-token count), paired with a 2×2 stats panel (active days / 90, avg input / day, peak day, current streak). Adds explicit `--lvl0..4` activity-density tokens (light + dark) so the ramp stays cleanly green instead of swinging tan in the mid buckets the way the previous `color-mix(in oklch, ...)` approach did.
- **Recent turns (30d)** — fixed 10-row table (was paginated `useInfiniteQuery` with Load more), model chip column, per-row volume bar. Wraps in a horizontal-scroll container with a min-width floor so the right-hand `Output / Total` columns stay reachable on narrow viewports.

Calendar joins server-side UTC `YYYY-MM-DD` daily buckets consistently in UTC (`getUTCDay/Month/Date/FullYear` + `setUTCHours(0)`) — fixes a half-day off-by-one in non-UTC zones where the tooltip / peak-day labels and the join key for `byDate` could drift apart.

**No backend changes**: same `/agents/.../usage/summary` and `/agents/.../usage/turns` endpoints, same `UsageAgentSummary` / `UsageTurnRow` schemas. Turns request limit reduced from 50 to 10 (potential request reduction, no Load-more follow-ups).

Reference design bundle: `first-tree-usage-redesign.html` from claude.ai/design — iterated with the user before handoff.

## Test plan

- [x] `pnpm --filter @first-tree/web typecheck` — pass (incl. design-token lint)
- [x] `pnpm check` (biome) — pass
- [x] `pnpm --filter @first-tree/web test` — 514 / 514 pass
- [x] `pnpm --filter @first-tree/web build` — pass
- [x] Local QA (agent `qa`): desktop 1440×1000 calendar + hover tooltip + stats panel render
- [x] Local QA: mobile 390×844 — Recent turns horizontally scrolls; Output / Total reachable after scroll
- [x] Local QA: UTC date join verified — server bucket `2026-05-31` ↔ tooltip "Sun · May 31, 2026" (no Jun 1 drift)
- [ ] Reviewer check: dark-mode contribution ramp + 2×2 stats panel readability
- [ ] Reviewer check: tooltip near viewport top (currently no flip-below fallback — known minor)

[...]

PR #1749, opened 2026-07-13
feat(chat): let agents send workspace images via chat send

Gandy2025:
## What

An agent that references a local workspace image with markdown image syntax `alt` in a `chat send` body has the bytes uploaded to the org attachment store and the message delivered as a `format:"file"` image batch — the **exact `imageBatchRefContent` shape a human image send produces**. The chat renders it identically (caption on top, thumbnails below), and inbound `ensureImagesLocal` already resolves that shape, so **agent→human and agent→agent image sends both work with no server/DB change**.

Fixes the reported bug where an agent's `` rendered as a broken-image icon (the local path 404s on the cloud origin).

## How

Picture sibling of the existing `.md` doc-capture flow:
- **`packages/client/.../image-snapshots.ts`** — scan `alt`, resolve against the sender's own workspace **fence** (realpath + hidden-segment + out-of-fence rejection), validate MIME/size, upload, strip captured spans → caption.
- **`apps/cli/.../image-capture.ts`** (+ shared `capture-context.ts`) — CLI wiring; `toOutboundImageMessage` decides format/content.
- **`chat/send.ts`** — after doc capture, run image capture; convert to the batch when any image resolved.
- **`chat-view.tsx`** (only web change) — dedupe a doc link that lands in an image batch's caption so a mixed doc+image message doesn't render the doc as both a caption link **and** a redundant chip.

## Code-sample exclusion (uses the renderer's own parser)

To avoid capturing an `` an agent *shows* as a code sample, capture excludes **block code** ranges obtained from **`mdast-util-from-markdown`** — the same parser ReactMarkdown uses. This matches the renderer exactly (fenced blocks at any container depth — blockquotes, list items — and indented code) with no hand-rolled CommonMark to diverge. **Inline `` `code` `` is deliberately treated as a live embed** (a rare place for a full embed; not excluding it avoids reproducing inline parsing on the send side).

## Scope / limits (documented in `docs/cli-reference.md`)

- Explicit markdown image syntax only; bare filenames stay text.
- Sender's own workspace only.
- `chat send` only (`chat ask`/`create` out of scope).
- Backslash-escape parity on the leading `!` (odd run = escaped).
- png/jpeg/gif/webp, ≤10MB/image, ≤20 images/message; the resolve/upload fan-out is bounded to the 20-cap *before* touching the filesystem.
- Capture is best-effort, never blocks a send, and is skipped for a body over ~1M characters.

## Context tree

Paired draft tree PR **agent-team-foundation/first-tree-context#727** records the durable decision + convergence caveats; reconciled against this PR's merged head before it is marked ready.

## Verification

New unit tests (client image-snapshots incl. blockquote/list/indented containers, inline-live, escape parity, bounded resolution; CLI env + conversion helper). Full client suite (1503), shared (575), doc-snapshots (30), cli image (10) green; typecheck + biome clean.

### DIFF
--- a/apps/cli/src/core/image-capture.ts
+++ b/apps/cli/src/core/image-capture.ts
@@ -1,6 +1,6 @@
 import { buildMessageImageSnapshots, type FirstTreeHubSDK } from "@first-tree/client";
 import type { ImageRefContent, MessageFormat } from "@first-tree/shared";
-import { resolveChatOrgId, resolveSelfFenceFromEnv } from "./doc-capture.js";
+import { resolveChatOrgId, resolveSelfFenceFromEnv } from "./capture-context.js";
 
 /**
  * Capture the workspace images an outbound `chat send` body references, the
--- a/apps/cli/src/core/image-capture.ts
+++ b/apps/cli/src/core/image-capture.ts
@@ -1,0 +1,59 @@
+import { buildMessageImageSnapshots, type FirstTreeHubSDK } from "@first-tree/client";
+import type { ImageRefContent, MessageFormat } from "@first-tree/shared";
+import { resolveChatOrgId, resolveSelfFenceFromEnv } from "./doc-capture.js";
+
+/**
+ * Capture the workspace images an outbound `chat send` body references, the
+ * picture sibling of `captureOutboundDocs`. An agent that writes a markdown
+ * image `![alt](path)` pointing at a local image inside its own workspace gets
+ * the bytes uploaded to the org attachment store; the caller then sends the
+ * message as a `format: "file"` image batch (caption = the image-stripped body,
+ * attachments = these refs) so web renders it exactly like a human image send.
+ *
+ * Like doc capture, this resolves the send-side fence from the runtime-injected
+ * env and the upload org from the target chat, so it is a pure pass-through
+ * (empty refs, unchanged content) outside an agent session or when the org
+ * can't be resolved. Capture failure NEVER blocks the send.
+ */
+export async function captureOutboundImages(
+  content: string,
+  ctx: { sdk: FirstTreeHubSDK; chatId?: string },
+  env: NodeJS.ProcessEnv = process.env,
+): Promise<{ caption: string; imageRefs: ImageRefContent[] }> {
+  const self = resolveSelfFenceFromEnv(env);
+  if (!self || !ctx.chatId) return { caption: content, imageRefs: [] };
+
+  try {
+    const orgId = await resolveChatOrgId(ctx.sdk, ctx.chatId);
+    if (!orgId) return { caption: content, imageRefs: [] };
+    const { imageRefs, strippedText } = await buildMessageImageSnapshots(content, self, { uploader: ctx.sdk, orgId });
+    return { caption: strippedText, imageRefs };
+  } catch {
+    return { caption: content, imageRefs: [] };
+  }
+}
+
+/**
+ * Decide the outbound `format` + `content` given the doc-captured body and the
+ * image-capture result. When any image captured (and the base format is an
+ * image-eligible text/markdown), the send becomes a human-identical
+ * `imageBatchRefContent` file message: caption = the image-stripped body
+ * (omitted when empty), attachments = the refs. `card`/`request`/`reference`
+ * bodies and image-free sends pass through unchanged. Pure + synchronous so the
+ * conversion is unit-tested without a live send.
+ */
+export function toOutboundImageMessage(
+  baseFormat: MessageFormat,
+  docContent: string,
+  captured: { caption: string; imageRefs: ImageRefContent[] },
+): { format: MessageFormat; content: unknown } {
+  const eligible = baseFormat === "text" || baseFormat === "markdown";
+  if (!eligible || captured.imageRefs.length === 0) return { format: baseFormat, content: docContent };
+  return {
+    format: "file",
+    content: {
+      ...(captured.caption.trim() ? { caption: captured.caption } : {}),
+      attachments: captured.imageRefs,
+    },
+  };
+}
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -417,8 +417,12 @@ EOF
 # instead of a broken local path. Only explicit `![...](...)` embeds are
 # captured (a bare filename is left as text), only the sender's own workspace,
 # and images shown inside code spans are left as literal samples. Capture is
-# best-effort and never blocks the send; an image that is too large (>10 MB),
-# unreadable, or beyond the 20-per-message cap is left in the text.
+# best-effort and never blocks the send. An image that is too large (>10 MB),
+# unreadable, or beyond the 20-per-message cap is skipped: if no image in the
+# message captured, the body is sent unchanged (the skipped embed stays as
+# text); if at least one sibling image did capture — so the message becomes an
+# image send — every workspace-image embed is removed from the caption, so a
+# skipped one is dropped rather than left as a path that would render broken.
 echo 'Latest run: ![chart](reports/latency.png)' | first-tree chat send code-agent -f markdown
 
 # Ask a human a tracked question (red-dot + blocks the chat for them until they
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -410,6 +410,17 @@ EOF
 # Stdin bodies are never checked — piping is also the escape hatch for
 # intentionally sending literal `\n` text.
 
+# Embed a workspace image — a markdown image `![alt](path)` in a `chat send`
+# body whose target is an image (png/jpeg/gif/webp) inside the agent's own
+# workspace is uploaded at send time and delivered as a real inline chat image
+# (the same shape a human composer upload uses), so recipients see the picture
+# instead of a broken local path. Only explicit `![...](...)` embeds are
+# captured (a bare filename is left as text), only the sender's own workspace,
+# and images shown inside code spans are left as literal samples. Capture is
+# best-effort and never blocks the send; an image that is too large (>10 MB),
+# unreadable, or beyond the 20-per-message cap is left in the text.
+echo 'Latest run: ![chart](reports/latency.png)' | first-tree chat send code-agent -f markdown
+
 # Ask a human a tracked question (red-dot + blocks the chat for them until they
 # answer). `chat ask` targets a single human; the message body IS the ask and
 # must be decision-self-sufficient for a reader who remembers nothing of the
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -416,7 +416,8 @@ EOF
 # (the same shape a human composer upload uses), so recipients see the picture
 # instead of a broken local path. Only explicit `![...](...)` embeds are
 # captured (a bare filename is left as text), only the sender's own workspace,
-# and an image shown inside a fenced code block (``` … ```) is left as a
+# and an image shown inside a block code sample the renderer recognizes (a
+# fenced block at any container depth, or an indented code block) is left as a
 # literal sample (an image written inside inline `code` is treated as a live
 # embed). Capture is best-effort and never blocks the send, and is skipped
 # entirely for a body longer than ~1 million characters (then sent verbatim). An image
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -419,7 +419,7 @@ EOF
 # and an image shown inside a fenced code block (``` … ```) is left as a
 # literal sample (an image written inside inline `code` is treated as a live
 # embed). Capture is best-effort and never blocks the send, and is skipped
-# entirely for a body larger than 1 MB (which is then sent verbatim). An image
+# entirely for a body longer than ~1 million characters (then sent verbatim). An image
 # that is too large (>10 MB), unreadable, or beyond the 20-per-message cap is
 # skipped: if no image in the message captured, the body is sent unchanged (the
 # skipped embed stays as text); if at least one sibling image did capture — so
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -416,13 +416,16 @@ EOF
 # (the same shape a human composer upload uses), so recipients see the picture
 # instead of a broken local path. Only explicit `![...](...)` embeds are
 # captured (a bare filename is left as text), only the sender's own workspace,
-# and images shown inside code spans are left as literal samples. Capture is
-# best-effort and never blocks the send. An image that is too large (>10 MB),
-# unreadable, or beyond the 20-per-message cap is skipped: if no image in the
-# message captured, the body is sent unchanged (the skipped embed stays as
-# text); if at least one sibling image did capture — so the message becomes an
-# image send — every workspace-image embed is removed from the caption, so a
-# skipped one is dropped rather than left as a path that would render broken.
+# and an image shown inside a fenced code block (``` … ```) is left as a
+# literal sample (an image written inside inline `code` is treated as a live
+# embed). Capture is best-effort and never blocks the send, and is skipped
+# entirely for a body larger than 1 MB (which is then sent verbatim). An image
+# that is too large (>10 MB), unreadable, or beyond the 20-per-message cap is
+# skipped: if no image in the message captured, the body is sent unchanged (the
+# skipped embed stays as text); if at least one sibling image did capture — so
+# the message becomes an image send — every workspace-image embed is removed
+# from the caption, so a skipped one is dropped rather than left as a path that
+# would render broken.
 echo 'Latest run: ![chart](reports/latency.png)' | first-tree chat send code-agent -f markdown
 
 # Ask a human a tracked question (red-dot + blocks the chat for them until they
--- a/packages/client/src/runtime/image-snapshots.ts
+++ b/packages/client/src/runtime/image-snapshots.ts
@@ -1,0 +1,271 @@
+import { readFile } from "node:fs/promises";
+import {
+  IMAGE_MIME_TO_EXT,
+  type ImageRefContent,
+  MAX_ATTACHMENT_BYTES,
+  MAX_BATCH_ATTACHMENTS,
+  type SupportedImageMime,
+} from "@first-tree/shared";
+import {
+  type AttachmentUploader,
+  canonicalizeWorkspacePath,
+  resolveSelfRoots,
+  resolveWorkspaceFile,
+  type SelfFence,
+} from "./doc-snapshots.js";
+
+/**
+ * Outbound image capture — the picture sibling of `doc-snapshots.ts`.
+ *
+ * When an agent's `chat send` body contains a markdown image `![alt](path)`
+ * whose target is a local image file inside the sender's own workspace fence,
+ * we upload the bytes to the org attachment store and hand the caller an
+ * `ImageRefContent` (the exact shape a human image send produces). The caller
+ * then converts the message to a `format: "file"` batch: caption = the body
+ * with the captured image spans stripped, `attachments` = these refs. Web then
+ * renders it identically to a human image send (caption on top, thumbnails
+ * below) — zero web/server change.
+ *
+ * Scope (see the paired design note): we deliberately capture ONLY markdown
+ * image syntax (an explicit "show this picture" intent, unlike a bare filename
+ * mention) and ONLY the sender's own workspace (self-fence; cross-agent image
+ * references are not a real use case). Capture never throws and never blocks a
+ * send — a failed resolve/upload leaves the mention untouched in the caption.
+ */
+
+/** Reverse of `IMAGE_MIME_TO_EXT` — lowercased file extension → canonical MIME.
+ *  `jpeg` and `jpg` both map to `image/jpeg`. */
+const EXT_TO_IMAGE_MIME: Record<string, SupportedImageMime> = (() => {
+  const map: Record<string, SupportedImageMime> = {};
+  for (const [mime, ext] of Object.entries(IMAGE_MIME_TO_EXT) as [SupportedImageMime, string][]) {
+    map[ext] = mime;
+  }
+  map.jpeg = "image/jpeg";
+  return map;
+})();
+
+/** Bounded upload retries before giving up on one image (parity with the doc
+ *  sibling's `DOC_UPLOAD_MAX_ATTEMPTS`). */
+const IMAGE_UPLOAD_MAX_ATTEMPTS = 3;
+
+type ImageOccurrence = {
+  writtenPath: string;
+  /** Span of the whole `![alt](path)` in the source text (for stripping). */
+  start: number;
+  end: number;
+};
+
+export type BuildImageAttachmentsOptions = {
+  uploader: AttachmentUploader;
+  orgId: string;
+};
+
+export type BuildMessageImageSnapshotsResult = {
+  /** Image refs to carry as the `format: "file"` batch `attachments` (empty ⇒
+   *  caller sends the message unchanged). Order follows first appearance. */
+  imageRefs: ImageRefContent[];
+  /** `text` with every captured image span removed and the leftover blank
+   *  lines collapsed — becomes the batch `caption`. */
+  strippedText: string;
+  /** Count of image mentions that resolved in-fence but could not be captured
+   *  (unreadable / too large / upload failed / over the batch cap). */
+  skipped: number;
+};
+
+/**
+ * Scan `text` for markdown image mentions, capture the in-fence ones, and
+ * return the refs + the caption (text with captured spans stripped). A pure
+ * pass-through (empty refs, unchanged text) when nothing resolves.
+ */
+export async function buildMessageImageSnapshots(
+  text: string,
+  self: string | SelfFence,
+  opts: BuildImageAttachmentsOptions,
+): Promise<BuildMessageImageSnapshotsResult> {
+  const occurrences = collectImageOccurrences(text);
+  if (occurrences.length === 0) return { imageRefs: [], strippedText: text, skipped: 0 };
+
+  const selfConfig: SelfFence = typeof self === "string" ? { agentHome: self } : self;
+  const roots = await resolveSelfRoots(selfConfig);
+  if (!roots) return { imageRefs: [], strippedText: text, skipped: occurrences.length };
+
+  // Pass 1 — resolve each mention to a readable in-fence file + its MIME.
+  type Resolved = ImageOccurrence & { file: string; mime: SupportedImageMime };
+  const resolvedList: Array<Resolved | null> = await Promise.all(
+    occurrences.map(async (occ): Promise<Resolved | null> => {
+      const mime = imageMimeForPath(occ.writtenPath);
+      if (!mime) return null;
+      const key = await canonicalizeWorkspacePath(roots, occ.writtenPath);
+      if (!key) return null;
+      const file = await resolveWorkspaceFile(roots.agentHomeReal, key);
+      if (!file) return null;
+      return { ...occ, file, mime };
+    }),
+  );
+
+  // Pass 2 — read + upload, de-duped by resolved file so the same picture
+  // referenced twice uploads once. Enforce the per-message batch cap.
+  const refByFile = new Map<string, ImageRefContent>();
+  let skipped = 0;
+
+  const attempted = new Set<string>();
+  const uploadTasks: Array<Promise<void>> = [];
+  for (const occ of resolvedList) {
+    if (!occ) continue;
+    if (attempted.has(occ.file)) continue;
+    attempted.add(occ.file);
+    if (attempted.size > MAX_BATCH_ATTACHMENTS) {
+      skipped += 1;
+      continue;
+    }
+    uploadTasks.push(
+      (async () => {
+        const ref = await captureAndUpload(occ.file, occ.writtenPath, occ.mime, opts);
+        if (ref) refByFile.set(occ.file, ref);
+        else skipped += 1;
+      })(),
+    );
+  }
+  await Promise.all(uploadTasks);
+  if (refByFile.size === 0) return { imageRefs: [], strippedText: text, skipped };
+
+  // Pass 3 — collect the refs (first-appearance order, de-duped) and strip the
+  // spans. We strip EVERY in-fence-resolved image span, not only the captured
+  // ones: because at least one image captured, the message flips to a `file`
+  // batch whose caption is this text, and a leftover `![alt](local/path)` would
+  // render as a broken `<img>` there. Out-of-fence mentions (unresolved) are
+  // left untouched — we can't upload what we can't read.
+  const seen = new Set<string>();
+  const imageRefs: ImageRefContent[] = [];
+  const resolvedSpans: Array<{ start: number; end: number }> = [];
+  for (const occ of resolvedList) {
+    if (!occ) continue;
+    resolvedSpans.push({ start: occ.start, end: occ.end });
+    const ref = refByFile.get(occ.file);
+    if (ref && !seen.has(occ.file)) {
+      seen.add(occ.file);
+      imageRefs.push(ref);
+    }
+  }
+
+  return { imageRefs, strippedText: stripSpans(text, resolvedSpans), skipped };
+}
+
+/** Read the file, enforce the upload cap, and upload. Returns the ref, or null
+ *  on any IO/size/upload failure (caller degrades to leaving the mention). */
+async function captureAndUpload(
+  file: string,
+  writtenPath: string,
+  mime: SupportedImageMime,
+  opts: BuildImageAttachmentsOptions,
+): Promise<ImageRefContent | null> {
+  let bytes: Buffer;
+  try {
+    bytes = await readFile(file);
+  } catch {
+    return null;
+  }
+  if (bytes.byteLength === 0 || bytes.byteLength > MAX_ATTACHMENT_BYTES) return null;
+
+  const filename = basename(writtenPath);
+  // Bounded retry, matching the doc sibling's `DOC_UPLOAD_MAX_ATTEMPTS`: a
+  // single transient blob-store blip should not silently drop the image.
+  for (let attempt = 1; attempt <= IMAGE_UPLOAD_MAX_ATTEMPTS; attempt += 1) {
+    try {
+      const uploaded = await opts.uploader.uploadAttachment({ bytes, mimeType: mime, filename, orgId: opts.orgId });
+      return { imageId: uploaded.id, mimeType: mime, filename, size: bytes.byteLength };
+    } catch {
+      // try again until the attempt budget is exhausted
+    }
+  }
+  return null;
+}
+
+/** Canonical image MIME for a path's extension, or null when the extension is
+ *  not a supported image type (so http URLs / non-images never match). */
+function imageMimeForPath(writtenPath: string): SupportedImageMime | null {
+  const dot = writtenPath.lastIndexOf(".");
+  if (dot < 0) return null;
+  const ext = writtenPath.slice(dot + 1).toLowerCase();
+  return EXT_TO_IMAGE_MIME[ext] ?? null;
+}
+
+function basename(p: string): string {
+  const segs = p.split(/[\\/]+/).filter(Boolean);
+  return segs.at(-1) ?? "image";
+}
+
+/**
+ * Collect markdown image mentions `![alt](path)` whose target is a local,
+ * image-extension path. A leading `!` is required (that is what distinguishes an
+ * image from a doc link); a `\` before the `!` escapes it, and any URL-scheme
+ * target (`http://`, `data:`) is skipped — those already render inline and are
+ * not workspace files. Mentions inside fenced (``` ```) or inline (`` ` ``) code
+ * are skipped too: there the agent is SHOWING the markdown, not embedding an
+ * image, and capturing would destructively strip a code sample.
+ */
+function collectImageOccurrences(text: string): ImageOccurrence[] {
+  const out: ImageOccurrence[] = [];
+  const codeRanges = codeSpanRanges(text);
+  const re = /!\[(?:[^\]\\]|\\.)*\]\((\S+?)(?:\s+"[^"]*")?\)/g;
+  let match: RegExpExecArray | null = re.exec(text);
+  while (match !== null) {
+    const whole = match[0];
+    const target = match[1];
+    const start = match.index;
+    const prev = start > 0 ? text[start - 1] : "";
+    if (prev !== "\\" && target && !hasUrlScheme(target) && !isInsideRange(start, codeRanges)) {
+      out.push({ writtenPath: target, start, end: start + whole.length });
+    }
+    match = re.exec(text);
+  }
+  return out;
+}
+
+/** Byte ranges covered by fenced code blocks (``` / ~~~) and inline code
+ *  spans, computed once so an image mention inside a code sample is skipped.
+ *  Fenced blocks are matched first and their spans exclude any inline backtick
+ *  inside them from being treated as a separate span. */
+function codeSpanRanges(text: string): Array<{ start: number; end: number }> {
+  const ranges: Array<{ start: number; end: number }> = [];
+  const fenced = /^[ \t]*(`{3,}|~{3,})[\s\S]*?^[ \t]*\1[ \t]*$/gm;
+  let m: RegExpExecArray | null = fenced.exec(text);
+  while (m !== null) {
+    ranges.push({ start: m.index, end: m.index + m[0].length });
+    m = fenced.exec(text);
+  }
+  const inline = /`+[^`\n]*`+/g;
+  m = inline.exec(text);
+  while (m !== null) {
+    const start = m.index;
+    if (!isInsideRange(start, ranges)) ranges.push({ start, end: start + m[0].length });
+    m = inline.exec(text);
+  }
+  return ranges;
+}
+
+function isInsideRange(pos: number, ranges: Array<{ start: number; end: number }>): boolean {
+  return ranges.some((r) => pos >= r.start && pos < r.end);
+}
+
+/** True for an absolute URL scheme (`http:`, `https:`, `data:`, …). A Windows
+ *  drive prefix like `C:` is not treated as a scheme (single-letter). */
+function hasUrlScheme(target: string): boolean {
+  return /^[a-z][a-z0-9+.-]+:/i.test(target);
+}
+
+/** Remove `[start,end)` spans from `text`, then collapse the runs of blank
+ *  lines a removed inline image can leave behind (3+ newlines → 2). */
+function stripSpans(text: string, spans: Array<{ start: number; end: number }>): string {
+  if (spans.length === 0) return text;
+  const ordered = [...spans].sort((a, b) => a.start - b.start);
+  let out = "";
+  let cursor = 0;
+  for (const span of ordered) {
+    if (span.start < cursor) continue;
+    out += text.slice(cursor, span.start);
+    cursor = span.end;
+  }
+  out += text.slice(cursor);
+  return out.replace(/\n[ \t]*\n[ \t]*\n+/g, "\n\n").trim();
+}
--- a/packages/client/src/runtime/image-snapshots.ts
+++ b/packages/client/src/runtime/image-snapshots.ts
@@ -1,10 +1,10 @@
 import { readFile, stat } from "node:fs/promises";
 import {
+  fencedCodeBlockRanges,
   IMAGE_MIME_TO_EXT,
   type ImageRefContent,
   MAX_ATTACHMENT_BYTES,
   MAX_BATCH_ATTACHMENTS,
-  markdownCodeSpanRanges,
   type SupportedImageMime,
 } from "@first-tree/shared";
 import {
@@ -207,9 +207,12 @@ function basename(p: string): string {
  * image-extension path. A leading `!` is required (that is what distinguishes an
  * image from a doc link); a `\` before the `!` escapes it, and any URL-scheme
  * target (`http://`, `data:`) is skipped — those already render inline and are
- * not workspace files. Mentions inside fenced (``` ```) or inline (`` ` ``) code
- * are skipped too: there the agent is SHOWING the markdown, not embedding an
- * image, and capturing would destructively strip a code sample.
+ * not workspace files. Mentions inside a fenced code block (``` ``` ``` ``` /
+ * `~~~`) are skipped — there the agent is SHOWING the markdown, not embedding an
+ * image, and capturing would destructively strip the code sample. Inline code
+ * (`` `![](x)` ``) is intentionally NOT excluded: it is a rare place to write a
+ * full image embed, and fenced-only exclusion sidesteps the CommonMark
+ * inline-parsing edge cases a hand-rolled scanner cannot match the renderer on.
  */
 function collectImageOccurrences(text: string): ImageOccurrence[] {
   // Guard: skip capture on an absurdly large body so a pathological message can
@@ -237,11 +240,11 @@ function collectImageOccurrences(text: string): ImageOccurrence[] {
   }
   if (candidates.length === 0) return [];
 
-  // Filter out candidates that sit inside a markdown code span (a shown sample,
-  // not an embed). `markdownCodeSpanRanges` is sorted by start, and candidates
+  // Filter out candidates that sit inside a fenced code block (a shown sample,
+  // not an embed). `fencedCodeBlockRanges` is in source order, and candidates
   // are already in source order, so a single monotonic cursor over the ranges
   // decides all candidates in O(candidates + ranges) — no per-candidate scan.
-  const codeRanges = markdownCodeSpanRanges(text);
+  const codeRanges = fencedCodeBlockRanges(text);
   const out: ImageOccurrence[] = [];
   let ri = 0;
   for (const c of candidates) {
--- a/packages/client/src/runtime/image-snapshots.ts
+++ b/packages/client/src/runtime/image-snapshots.ts
@@ -51,6 +51,9 @@ const IMAGE_UPLOAD_MAX_ATTEMPTS = 3;
 
 type ImageOccurrence = {
   writtenPath: string;
+  /** Supported image MIME derived from the target's extension. Only supported
+   *  image targets become occurrences, so the cap is over eligible images. */
+  mime: SupportedImageMime;
   /** Span of the whole `![alt](path)` in the source text (for stripping). */
   start: number;
   end: number;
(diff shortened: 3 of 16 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749

- Rationale: the outbound capture path can allocate unbounded local files before enforcing the advertised cap, and its code-fence scanner can destructively capture valid Markdown examples.

**Risk level**: A

**PR summary**
- Problem: agents currently send workspace-local Markdown image paths as cloud-broken links, so human and agent recipients cannot see the intended image.
- Approach: capture explicitly embedded local images at `chat send` time inside the sender's workspace fence, upload them to the org attachment store, and send the same `format: "file"` batch shape used by human image sends; mixed document/image captions are deduplicated in Web.
- Impacted modules: CLI chat send/capture, Client workspace snapshotting, and Web chat rendering.

**Review findings**
❌ 1. Enforce the 10 MB limit before reading the whole file. `captureAndUpload` calls `readFile(file)` and only checks `bytes.byteLength` afterward; up to 20 unique files are also read concurrently. A workspace path ending in `.png` can therefore make the CLI allocate hundreds of MB or several GB and be OOM-killed before the catch/fallback path runs, so the advertised size cap and “capture never blocks a send” guarantee do not hold. Check file size before allocation and/or use a bounded read/stream, and cover an oversized sparse file without loading it. [R4 / `packages/client/src/runtime/image-snapshots.ts:164`]

❌ 2. Make fenced-code detection follow the Markdown fence rules before stripping image spans. The regex requires an exact closing delimiter (`\1`) and requires a close at all, but a valid closing fence may be longer than the opener and an unclosed fence extends to EOF. For example, a three-backtick `md` fence containing `x` and closed with four backticks is rendered as code by the Web Markdown path, yet this scanner treats the image as an embed, uploads it, and removes it from the authored code sample. Handle longer and EOF-terminated fences and add both cases to the scan tests. [R4 / `packages/client/src/runtime/image-snapshots.ts:231`]

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks — all three findings addressed in `fc3bc40a`.

**baixiaohang R4-1 (size cap before allocation).** `captureAndUpload` now `stat`s the file and rejects `size === 0 || size > MAX_ATTACHMENT_BYTES` **before** `readFile`, so a large or sparse `.png` can never allocate its bytes (a post-read re-check guards a stat→read race). `packages/client/src/runtime/image-snapshots.ts` — new test "skips an oversized file via stat, without reading its bytes" uses an 11 MB **sparse** file (truncate, no real bytes) and asserts only the small sibling uploads.

**baixiaohang R4-2 (fenced-code rules).** Replaced the ad-hoc fence regex with a shared CommonMark scanner `fencedCodeBlockRanges` (new export in `@first-tree/shared`, built on the same `parseOpeningFence`/`isClosingFence` used by the bare-doc scanner). It now honors a **longer closing fence** (4-backtick close of a 3-backtick open) and an **unclosed fence extending to EOF**. New tests: "respects a longer closing fence and an unclosed (EOF) fence" (client) + `fencedCodeBlockRanges` unit tests (shared).

**CodeQL polynomial-regex (#71/#72).** The image-scan regex quantifiers are now bounded (`![^\]\n]{0,512}` alt, `[^\s)"]{1,2048}` path), so a pathological `![![![…` body stays linear instead of O(n²). New "ReDoS guard" test drives a 50k-repetition body.

Verification: shared 579, client image 15 + doc-snapshots 30, cli image 10, all typechecks + biome clean. Re-requesting review.

packages/client/src/runtime/image-snapshots.ts
```diff
@@ -0,0 +1,271 @@
...
+  return out;
+}
+
+/** Byte ranges covered by fenced code blocks (``` / ~~~) and inline code
+ *  spans, computed once so an image mention inside a code sample is skipped.
+ *  Fenced blocks are matched first and their spans exclude any inline backtick
+ *  inside them from being treated as a separate span. */
+function codeSpanRanges(text: string): Array<{ start: number; end: number }> {
+  const ranges: Array<{ start: number; end: number }> = [];
+  const fenced = /^[ \t]*(`{3,}|~{3,})[\s\S]*?^[ \t]*\1[ \t]*$/gm;
```
yuezengwu: Blocker: preserve every valid fenced-code range before scanning images. CommonMark permits a closing fence to be longer than its opener, but the backreference here requires exactly the same length. For a body whose lines are an opening ` ```md `, `literal`, and a closing ` ```` `, the image is inside a valid code block; on this head `buildMessageImageSnapshots` nevertheless uploads `shot.png` and strips the literal line. An unclosed fence also falls through. That violates the explicit code-sample exclusion and can publish a workspace image that the sender only meant to quote. Please recognize closing fences of the same character with length >= the opener (and EOF-terminated fences), with regression coverage.

packages/client/src/runtime/image-snapshots.ts
```diff
@@ -0,0 +1,271 @@
...
+  const ordered = [...spans].sort((a, b) => a.start - b.start);
+  let out = "";
+  let cursor = 0;
+  for (const span of ordered) {
+    if (span.start < cursor) continue;
+    out += text.slice(cursor, span.start);
+    cursor = span.end;
+  }
+  out += text.slice(cursor);
+  return out.replace(/\n[ \t]*\n[ \t]*\n+/g, "\n\n").trim();
```
yuezengwu: Blocker: keep untouched caption ranges byte-for-byte. This replacement collapses 3+ newlines across the entire body, not only whitespace left next to a removed image span. I reproduced a message with a normal fenced code block containing two blank lines plus a captured image after the block; `strippedText` deletes one blank line inside the code sample. That mutates authored code even though image scanning correctly skipped the block. Please limit cleanup to whitespace adjacent to removed spans (or omit normalization) and add a regression test that preserves blank lines inside an unrelated fence.

yuezengwu (changes requested):
Implementation reviewed: after document capture, local Markdown image references resolve through the sender's existing self-fence, upload into the target chat's organization, and convert the outbound message to the existing `format: file` image-batch shape; document metadata remains attached, and Web avoids rendering an inline document link twice. This correctly reuses the current persisted shape and introduces no database or server-schema change.

I am requesting changes for the two content-preservation/security blockers called out inline. The focused client/CLI/Web suites pass (102 tests total), as do client, CLI, and Web typechecks and Biome, but both blocker cases were reproduced with failing tests against this head.

Before merge, please also update `docs/cli-reference.md` for this user-visible `chat send` behavior, as required by the repository guidance. The PR description also says there is a paired Context Tree PR, but I could not find an open PR or cross-reference in `first-tree-context`; please open/link it so reviewers can verify the convergence caveats together.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749 (follow-up)

- Rationale: the previous size-cap and fenced-block findings are addressed, but the follow-up regresses preservation of valid multi-backtick inline code spans.

**Risk level**: A

**PR summary**
- Problem: agents need workspace-local Markdown image embeds to arrive as real chat images instead of broken cloud-local paths.
- Approach: capture in-fence images at send time and emit the existing human image-batch shape; `fc3bc40a` adds pre-read size checks and a shared CommonMark fenced-block scanner.
- Impacted modules: Shared Markdown scanning, Client image snapshots, CLI send capture, and Web chat rendering.

**Review findings**
❌ 1. Preserve multi-backtick inline code spans. This follow-up changes the inline scanner from a variable-length delimiter to `/`[^`\n]*`/`, which only models one-backtick spans. In a valid body such as double-backtick + `x` + double-backtick, the regex records each adjacent backtick pair as an empty range; the image between them is therefore considered live, uploaded, and stripped from the authored code sample. That still violates the stated “fenced/inline code is skipped” contract. Reuse a delimiter-aware inline-code scanner (matching equal-length runs) and add a double-backtick case; the current client test covers only single backticks. [R4 / `packages/client/src/runtime/image-snapshots.ts:243`]

✅ The static oversized-file case is now rejected from `stat` before `readFile`, with a post-read re-check.
✅ `fencedCodeBlockRanges` correctly uses the shared CommonMark opener/closer rules and extends unclosed fences to EOF.

**Action taken**
- Submitted follow-up request changes.

Gandy2025:
Paired Context Tree PR: **agent-team-foundation/first-tree-context#727** (draft — records that agent `chat send` captures workspace image embeds + the convergence caveats; will be reconciled against this PR's merged head and marked ready).

Gandy2025:
Both owners' blockers addressed in `b9572886`.

**baixiaohang follow-up (multi-backtick inline).** The code-span scanner is now the shared `markdownCodeSpanRanges`, which composes the CommonMark fenced-block logic with the existing **delimiter-aware** `findInlineCodeSpans` (equal-length backtick-run matching). A `` ``x`` `` span is now recognized as code, so the image is not captured or stripped. Tests: "does NOT capture an image inside a multi-backtick inline code span" (client) + "single- and multi-backtick inline spans" (shared).

**yuezengwu #1 (fence longer-close / EOF).** Already resolved in `fc3bc40a` via `markdownCodeSpanRanges` (same-char run ≥ opener; unclosed → EOF), which your latest note confirmed ✅.

**yuezengwu #2 (caption byte-preservation).** `stripSpans` no longer does a global newline collapse. It now clamps the newline run **only at each removed-span join** (so a standalone image line doesn't leave a doubled blank line), leaving all other text — including blank lines inside a preserved code block — byte-for-byte. Regression test "preserves blank lines inside an unrelated fenced block when stripping a later image".

Also: documented the `chat send` image-embed behavior in `docs/cli-reference.md`, and linked the paired tree PR `first-tree-context#727` above.

Verification: shared 581, client image 17 + doc-snapshots 30, cli image 10, web chat 168, typechecks + biome clean. Re-requesting review.

apps/cli/src/core/image-capture.ts
```diff
@@ -0,0 +1,59 @@
...
+ * env and the upload org from the target chat, so it is a pure pass-through
+ * (empty refs, unchanged content) outside an agent session or when the org
+ * can't be resolved. Capture failure NEVER blocks the send.
+ */
+export async function captureOutboundImages(
+  content: string,
+  ctx: { sdk: FirstTreeHubSDK; chatId?: string },
+  env: NodeJS.ProcessEnv = process.env,
+): Promise<{ caption: string; imageRefs: ImageRefContent[] }> {
+  const self = resolveSelfFenceFromEnv(env);
```
yuezengwu, line 23: Blocker: importing `resolveSelfFenceFromEnv` (and `resolveChatOrgId`) through `doc-capture` breaks four existing command-test modules that fully mock `../core/doc-capture.js`. A focused run on `b9572886` reproduces 8 failures before a message can be sent: Vitest reports that this export is missing from the mock. Please update those mocks (or mock the new image-capture boundary) and run the affected command suites, not only `image-capture.test.ts`.

docs/cli-reference.md
```diff
@@ -410,6 +410,17 @@ EOF
...
 
+# Embed a workspace image — a markdown image `![alt](path)` in a `chat send`
+# body whose target is an image (png/jpeg/gif/webp) inside the agent's own
+# workspace is uploaded at send time and delivered as a real inline chat image
+# (the same shape a human composer upload uses), so recipients see the picture
+# instead of a broken local path. Only explicit `![...](...)` embeds are
+# captured (a bare filename is left as text), only the sender's own workspace,
+# and images shown inside code spans are left as literal samples. Capture is
+# best-effort and never blocks the send; an image that is too large (>10 MB),
+# unreadable, or beyond the 20-per-message cap is left in the text.
```
yuezengwu: This is only true when no sibling image captures. Once any image captures, `buildMessageImageSnapshots` converts the message to a file batch and deliberately strips every in-fence-resolved image span, including an oversized/unreadable/over-cap sibling, to avoid leaving a broken local `` in the caption. Please qualify this mixed-success behavior so the user-facing docs match the implementation.

yuezengwu (changes requested):
Goal/core change reviewed: `chat send` now resolves explicit local Markdown image embeds through the sender-only workspace fence, uploads them into the target chat organization, and emits the existing human `format: file` image-batch shape; Web only deduplicates the mixed document/image presentation. This reuses the current persisted structure and introduces no database or server-schema migration.

The latest follow-up correctly fixes the pre-read size cap, CommonMark fenced-block handling, multi-backtick inline spans, and global blank-line rewrite; the new CLI reference section is also present. I verified the focused Shared and Client suites (33 + 45 tests) are green.

I am requesting changes for the remaining test-suite blocker inline. On this exact head, the focused CLI run has 52 passes and 8 failures across `chat-attention-commands-extra`, `chat-send-escaped-newline-guard`, `chat-send-message-file`, and `cli-command-branch-sweeper`; every failure is the same missing `resolveSelfFenceFromEnv` export from the full `../core/doc-capture.js` mock. The new documentation also needs the mixed-success qualification called out inline. I did not wait for CI before submitting this review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749 (third pass)

- Rationale: `b9572886` fixes the direct double-backtick case, but the shared helper still diverges from CommonMark in ways that can upload and strip image syntax rendered inside inline code.

**Risk level**: A

**PR summary**
- Problem: workspace-local image embeds from agents currently render as broken cloud paths instead of delivered image attachments.
- Approach: capture in-fence images during `chat send`, emit the existing human image-batch shape, and preserve Markdown code ranges; this pass also documents the CLI behavior and links the paired tree update.
- Impacted modules: Shared Markdown scanning, Client image snapshots, CLI send capture/reference docs, Web rendering, and paired Context Tree PR #727.

**Review findings**
❌ 1. Protect multiline inline code spans. `markdownCodeSpanRanges` calls `findInlineCodeSpans` once per physical line, but CommonMark code spans may contain line endings. In a body such as ``before `line 1\nx\nline 3` after``, neither delimiter is paired on its individual line, so the image occurrence is considered live, uploaded, and stripped even though Web renders it inside inline code. Scan inline delimiters across the non-fenced text (while preserving absolute offsets), and add a multiline span regression case. [R4 / `packages/shared/src/lib/doc-link-scan.ts:186`]

❌ 2. Require an exact closing backtick run, not a substring of a longer run. `line.indexOf(ticks, idx)` finds N backticks inside an N+1 run, despite the helper's contract saying the close must be equal-length. For example, a two-backtick opener followed by a three-backtick run and then the real two-backtick close is ended at the three-run; any image between that run and the real close is then captured from content Web still treats as code. Advance by whole delimiter runs and accept only `run.length === opener.length`; add a mismatched-run case. [R4 / `packages/shared/src/lib/doc-link-scan.ts:265`]

⚠️ 3. Align the new CLI documentation with mixed-success behavior. The reference says an oversized/unreadable/over-cap image “is left in the text,” but once any sibling image uploads, Pass 3 intentionally strips every in-fence-resolved span, including failed or over-cap images, to avoid a broken image in the file-batch caption. Document the conditional behavior instead of promising that those mentions always remain. [`docs/cli-reference.md:420`]

✅ The original size-cap, fenced-block, and direct multi-backtick cases are addressed.
✅ Paired Context Tree PR #727 is now linked and the user-visible CLI behavior is documented.

**Action taken**
- Submitted a third request-changes review.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,52 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+ * {@link scanBareDocPathTokens}: a closing fence must be the same marker char
+ * and at least as long as the opener, an unclosed fence extends to
+ * end-of-input, and an inline span closes on an equal-length backtick run.
+ * Returned as `[start, end)` offsets into `text`. Callers (e.g. the
+ * image-capture scanner) use these to avoid treating markdown shown inside a
+ * code sample as a live construct.
+ */
+export function markdownCodeSpanRanges(text: string): Array<{ start: number; end: number }> {
+  const ranges: Array<{ start: number; end: number }> = [];
+  const lines = text.split(/(\r?\n)/);
```
yuezengwu: Blocker: splitting before inline-span recognition means a CommonMark code span cannot cross a physical newline. I reproduced with `before ` + one opening tick + `line 1\nx\nline 3` + one closing tick + ` after`: Web treats the middle as code, but this helper returns no covering inline range and image capture uploads/strips `shot.png`. Scan delimiter runs across each non-fenced text region (preserving absolute offsets) and add LF/CRLF multiline coverage.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,52 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+      continue;
+    }
+    const opening = parseOpeningFence(line);
+    if (opening) {
+      fence = opening;
+      fenceStart = absoluteOffset;
+      absoluteOffset += line.length;
+      continue;
+    }
+    for (const span of findInlineCodeSpans(line)) {
```
yuezengwu: Blocker: `indexOf(ticks)` accepts N ticks at the start of a longer delimiter run, while CommonMark closes only on a whole run of exactly N. A reliably exposing case is a two-tick opener, a five-tick non-closer, `x`, then the real two-tick close; this head uploads and strips the image even though it remains inside the code span. (The three-tick variant is accidentally masked by a second bogus one-tick range, so it is not sufficient regression coverage.) Walk whole backtick runs and accept only an exact-length close.

yuezengwu (changes requested):
Supplemental review of the same core path: local Markdown images are captured through the sender-only workspace fence and emitted in the existing human image-batch shape; there is still no database or server-schema change. I independently reproduced both newly reported CommonMark gaps with failing Client tests on `b9572886`. In each case an image authored as literal inline code is uploaded and stripped, so these are content-preservation and unintended-publication blockers, not parser polish. The earlier CLI mock failures and mixed-success CLI documentation mismatch also remain open.

Gandy2025:
All open blockers from this round fixed in `6e9dec0a`.

**CLI mock failures (8, yuezengwu).** Root cause was a leaky module seam: `image-capture.ts` imported `resolveSelfFenceFromEnv`/`resolveChatOrgId` from `doc-capture.js`, so any test that `vi.mock`s `doc-capture.js` dropped those exports. Moved both helpers to a new neutral `apps/cli/src/core/capture-context.ts` that both capture modules import; the two are no longer coupled through each other's surface. The four previously-failing files (`chat-send-message-file`, `chat-send-escaped-newline-guard`, `chat-attention-commands-extra`, `cli-command-branch-sweeper`) + `image-capture` now pass (60/60).

**Multiline inline code spans (baixiaohang #1).** `markdownCodeSpanRanges` now scans inline delimiters across the WHOLE text (a CommonMark code span may contain line endings), not per physical line — so an image inside a multiline inline span is left as code. Test: "covers an inline code span that contains a line ending".

**Exact-length close (baixiaohang #2).** Both the new text-level scanner and `findInlineCodeSpans` (cited at :265) now advance by whole backtick runs and accept a close only when `run.length === opener.length`; a longer run is skipped. Test: "closes an inline span only on an EXACT-length backtick run".

**CLI docs mixed-success (⚠️).** Reworded: a skipped image stays as text only when *no* sibling captured; once the message becomes an image send, every embed (incl. a skipped one) is removed from the caption.

Verification (clean env): shared 583, client image 18 + doc-snapshots 30, and the previously-failing CLI files 60/60; typecheck + biome clean. (Note: a handful of `daemon-start` / `tree-init` integration tests fail in my local sandbox — they need a real daemon/git setup and import none of the changed files; CI runs them properly. Please flag if CI shows any of those failing so I can dig in.)

Re-requesting review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749 (fourth pass)

- Rationale: `6e9dec0a` closes every previously reported correctness/test/documentation item, but the new whole-text code scanner introduces quadratic fence lookup on the synchronous send path.

**Risk level**: A

**PR summary**
- Problem: agents need local Markdown image embeds delivered as real chat images rather than broken cloud-local paths.
- Approach: capture in-fence images during `chat send`, upload them to the chat organization, emit the existing human image-batch shape, and preserve authored Markdown code ranges.
- Impacted modules: Shared Markdown scanning, Client image snapshots, CLI capture context/send flow, Web rendering, CLI docs, and paired Context Tree PR #727.

**Review findings**
❌ 1. Keep fenced-range lookup linear in the message length. The whole-text inline pass calls `inFence(i)` for essentially every character, and `inFence` performs `fenced.some(...)` over every fenced block. A message containing many short fenced blocks therefore costs O(text length × fence count); because `collectImageOccurrences` builds code ranges before even checking whether the body contains an image embed, an ordinary large `chat send` can spend hundreds of millions or more range predicates here and appear hung. This reopens the “capture never blocks a send” failure mode immediately after bounding the image regex. Walk the already ordered fence ranges with a monotonic cursor (and similarly avoid `ranges.some` per image occurrence), then add a many-fences complexity guard. [R4 / `packages/shared/src/lib/doc-link-scan.ts:219`; caller `packages/client/src/runtime/image-snapshots.ts:216`]

✅ All previously reported size-cap, fenced/inline-code correctness, CLI mock, caption preservation, and documentation findings are addressed.
✅ The capture-context extraction removes the cross-module mock coupling cleanly.
✅ Paired Context Tree PR #727 remains linked and consistent with the chosen image shape.

**Action taken**
- Submitted a fourth request-changes review.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,86 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+    const opening = parseOpeningFence(line);
+    if (opening) {
+      fence = opening;
+      fenceStart = absoluteOffset;
+    }
+    absoluteOffset += line.length;
+  }
+  if (fence) fenced.push({ start: fenceStart, end: text.length });
+
+  // Inline code spans are scanned across the WHOLE text (a CommonMark code span
```
yuezengwu: Blocker: `WHOLE text` is broader than a CommonMark inline-parsing block. Code spans may cross soft line endings, but they do not cross a blank-line paragraph boundary or enter a fenced block. I reproduced both mismatches: `` `unclosed\n\nx closing` `` and an unmatched tick followed by a live image and then a fence containing a single tick. ReactMarkdown renders `shot.png` as a live image in both; this helper pairs across the boundary, so capture skips it and sends the broken local path. Scan inline delimiters within non-fenced Markdown block/paragraph regions while preserving offsets, and cover both boundaries.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,86 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+  // may contain line endings), matching an opener run of N backticks to a
+  // closing run of EXACTLY N backticks (a longer run is not a close). Openers
+  // inside a fenced block are skipped.
+  const ranges = [...fenced];
+  const inFence = (pos: number): boolean => fenced.some((r) => pos >= r.start && pos < r.end);
```
yuezengwu: Blocker: this is O(text length × fence count), because virtually every character calls `fenced.some`. A 550 KB probe with 50,000 tiny fenced blocks takes ~4.7 seconds here before image matching; the CLI accepts bodies up to 5 MB. Walk the ordered fence ranges with a monotonic cursor and jump over each whole fenced range. The caller's ordered image matches should likewise use a monotonic range cursor instead of `ranges.some` per occurrence.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,86 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+  while (i < text.length) {
+    if (text[i] !== "`" || inFence(i)) {
+      i += 1;
+      continue;
+    }
+    let openEnd = i;
+    while (openEnd < text.length && text[openEnd] === "`") openEnd += 1;
+    const openLen = openEnd - i;
+    let closeEnd = -1;
+    for (let k = openEnd; k < text.length; ) {
```
yuezengwu: Blocker: there is a second superlinear path independent of fence lookup. Every unmatched opener rescans the remaining text for an equal-length run. A 4.5 MB input containing 3,000 unique run lengths (all unmatched) takes ~11.1 seconds and trips Vitest's timeout, so a fence-cursor-only fix will still let an ordinary no-image send appear hung. Tokenize delimiter runs once and use a linear or indexed next-same-length lookup within each inline-parsing block; add a guard for unmatched distinct-length runs as well as many fences.

yuezengwu (changes requested):
Goal/core change re-reviewed: agent `chat send` captures explicit local image embeds through the sender-only workspace fence and emits the existing human image-batch shape; Web only deduplicates mixed document/image presentation. There is no database or server-schema migration.

`6e9dec0a` cleanly fixes every previously open item: the CLI capture-context seam, multiline and exact-run cases, mixed-success docs, and the eight command-test failures. Focused Shared, Client, and CLI suites pass (35 + 46 + 60 tests), and paired tree PR #727 remains aligned.

I am requesting changes for the new whole-text scanner blockers inline. ReactMarkdown and capture-layer probes confirm two block-boundary mismatches, and benchmarks confirm two independent superlinear paths: 50,000 short fences in a 550 KB body take about 4.7 seconds; 3,000 unique unmatched backtick runs in a 4.5 MB body (within the CLI's 5 MB file/stdin limit) take about 11.1 seconds. Capture builds these ranges before checking whether any image syntax exists, so ordinary sends can hit the cost.

Gandy2025:
Both block-boundary mismatches and both superlinear paths fixed in `d33ec0b6` — the scanner is rewritten to be block-aware and linear.

**Block boundaries (yuezengwu).** Inline code spans are now scanned per **paragraph region** (the maximal run of non-blank, non-fenced lines), not across the whole text. So:
- an unclosed backtick before a blank line no longer pairs across the paragraph boundary (`` `open\n\nx` `` → the image is live and captured);
- a stray backtick never pairs with one inside a fenced block (the region excludes fences).
Spans still cross soft line breaks within a paragraph. Tests: "does NOT pair inline delimiters across a blank-line paragraph boundary" (shared) + "captures an image after a blank line even when an earlier backtick is unclosed" (client).

**Superlinear path #1 — fence lookup (baixiaohang).** Gone: fenced ranges come from the single line scan; the per-region inline pass never does `fenced.some` per character.

**Superlinear path #2 — unmatched-opener rescan (yuezengwu :230).** Gone: within a region the backtick runs are tokenized once and each run's next-equal-length run is precomputed in one right-to-left pass, so pairing is O(runs) with no rescan. Test "stays fast on a large body of many unmatched distinct-length runs".

**Short-circuit + guard.** `collectImageOccurrences` now finds image candidates with the bounded regex FIRST and computes code ranges only when ≥1 candidate exists (a text-only send computes nothing), filtering with a monotonic cursor; plus a >1MB body skips capture as a backstop. Test "stays fast with many fenced blocks and skips a >1MB body" (20k fences + a real image completes in ms).

Verification: shared 586, client image 20 + doc-snapshots 30, cli 24; typecheck + biome clean (removed the non-null assertions too). CI was green on the prior head. Re-requesting review.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,125 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+      fenceStart = absoluteOffset;
+      absoluteOffset += line.length;
+      continue;
+    }
+    if (/^[ \t]*$/.test(line)) {
+      flushParagraph(); // blank line — paragraph boundary
+      absoluteOffset += line.length;
+      continue;
+    }
+    if (paraStart < 0) paraStart = absoluteOffset;
```
baixiaohang: Blocker: a maximal run of non-blank/non-fenced lines is not necessarily one CommonMark paragraph or inline-parsing block. For example, `intro ` + one opening tick + `unclosed\n# x` + one closing tick has a paragraph followed immediately by an ATX heading; the heading interrupts the paragraph without a blank line, so ReactMarkdown treats the image as live. This region merges both blocks and pairs the ticks, causing image capture to skip `shot.png` and leave a broken local path. List-item and other interrupting block starts have the same problem. Please derive inline ranges within actual Markdown block boundaries (or conservatively split on every supported interrupting block start) and add a heading/list regression.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,125 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+ */
+function collectInlineCodeInRegion(
+  text: string,
+  start: number,
+  end: number,
+  out: Array<{ start: number; end: number }>,
+): void {
+  const runs: Array<{ start: number; end: number; len: number }> = [];
+  for (let i = start; i < end; ) {
+    if (text[i] !== "`") {
```
baixiaohang: Blocker: every backtick run is treated as a delimiter, including a CommonMark backslash-escaped literal. In `before \` literal x after ` (the first tick is escaped; the final tick is unmatched), ReactMarkdown renders the image as live, but this helper pairs those two ticks and marks the image as code, so capture skips it. Please honor backslash-escape parity when tokenizing delimiter runs and cover this exact case.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749

- Rationale: The linear rewrite closes the reported performance paths, but the custom inline-code segmentation still diverges from the Markdown renderer in two ordinary CommonMark cases, so a live local image can remain uncaptured and render broken.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: agent-authored local image embeds currently point at workspace-only paths that humans and peer agents cannot load.
- Approach: resolve explicit image embeds through the sender-only workspace fence, upload them to the chat organization, and send the existing human image-batch shape while preserving literal code samples.
- Impacted modules: CLI chat send/capture, client snapshot capture, shared Markdown scanning, and Web mixed-attachment rendering.

**Review findings**
1. Blocking: non-blank/non-fenced lines are not necessarily one Markdown inline-parsing block; ATX headings, list items, and other interrupting block starts can split them without a blank line. The scanner can therefore pair backticks across real block boundaries and suppress a live image.
2. Blocking: escaped backticks are still tokenized as code delimiters. A backslash-escaped literal tick can pair with a later unmatched tick around an otherwise live image, again suppressing capture.

**Action taken**
- Submitted request changes with two inline reproduction cases.

I did not run tests or code checks in this review pass.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,125 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+ * monotonic-cursor lookup).
+ */
+export function markdownCodeSpanRanges(text: string): Array<{ start: number; end: number }> {
+  const fenced: Array<{ start: number; end: number }> = [];
+  const inline: Array<{ start: number; end: number }> = [];
+  const lines = text.split(/(\r?\n)/);
+  let fence: FenceState | null = null;
+  let fenceStart = 0;
+  let absoluteOffset = 0;
+  // Current paragraph region [paraStart, paraEnd) of contiguous non-blank,
```
yuezengwu: Blocker: contiguous nonblank/nonfenced lines are not necessarily one CommonMark paragraph/inline block. ATX headings, separate list items, and setext/thematic boundaries split blocks without a blank line. I rendered and capture-tested `# ` + an unmatched tick + `\nx closing` (plus list and `---` variants): ReactMarkdown renders a live image, but this region pairs the ticks across blocks and capture returns no attachment. Segment on actual Markdown block boundaries (or use the renderer's parser/tokenizer), with ATX/list/setext regressions.

packages/shared/src/lib/doc-link-scan.ts
yuezengwu: Blocker: escaped backticks are literal text, not code delimiters. For a body beginning with a backslash-escaped backtick, followed by `literal x closing`, and ending with one unescaped backtick, ReactMarkdown renders `shot.png` as a live image; this tokenizer records both ticks, pairs them, and capture skips the image. Respect CommonMark backslash escaping (odd/even preceding backslashes) and add a live-image regression.

packages/client/src/runtime/image-snapshots.ts
```diff
@@ -0,0 +1,302 @@
...
+ * target (`http://`, `data:`) is skipped — those already render inline and are
+ * not workspace files. Mentions inside fenced (``` ```) or inline (`` ` ``) code
+ * are skipped too: there the agent is SHOWING the markdown, not embedding an
+ * image, and capturing would destructively strip a code sample.
+ */
+function collectImageOccurrences(text: string): ImageOccurrence[] {
+  // Guard: skip capture on an absurdly large body so a pathological message can
+  // never make a synchronous `chat send` appear to hang. Capture is
+  // best-effort, so a >1MB body just sends verbatim.
+  if (text.length > MAX_IMAGE_SCAN_BYTES) return [];
```
yuezengwu: Contract gap: this silently disables image capture above 1,048,576 UTF-16 code units, while `chat send -F` accepts bodies up to 5 MiB and the CLI reference does not disclose a message-body image-capture cap. A valid 1–5 MiB message therefore sends the original broken local path. Now that scanning is linear, please align the guard with the supported transport limit or make this lower limit shared and user-visible; if retained, name/measure it as characters rather than bytes.

yuezengwu (changes requested):
Goal/core change re-reviewed: agent `chat send` captures explicit local image embeds through the sender-only workspace fence and emits the current human image-batch shape, with no database or server-schema migration.

`d33ec0b6` fixes the previously reported blank/fence boundary and both complexity regressions. The prior probes now take ~26 ms for 50,000 fences and ~6 ms for 3,000 unmatched distinct-length runs; focused Shared, Client, and CLI suites pass (38 + 48 + 60), and current CI is green. I resolved the superseded threads, and paired tree PR #727 remains aligned.

I am requesting changes for two remaining CommonMark mismatches independently confirmed in both ReactMarkdown output and capture-layer tests, plus the user-visible scan-limit contract called out inline. The parser mismatches leave otherwise-live workspace images uncaptured, recreating the broken-local-path behavior this PR is meant to remove.

Gandy2025:
Resolved in `f5f55537` by a **deliberate scope decision** (product owner sign-off) rather than chasing full inline CommonMark conformance.

**What changed.** The code-sample exclusion now covers **fenced code blocks only** (`fencedCodeBlockRanges` — a single linear line scan; longer-close and EOF handled). The inline-code-span machinery is removed. An `` written inside inline `` `code` `` is now treated as a **live embed**.

**Why.** Matching a hand-rolled inline scanner to ReactMarkdown's inline parsing means reproducing its full CommonMark model — paragraph/heading/list block boundaries, escapes, multi-tick runs, etc. — an open-ended surface (each pass here surfaced another edge). Fenced blocks are where agents actually paste multi-line markdown/code samples, are line-anchored and unambiguous, and carry none of that. A full image embed inside *inline* code is rare, and the failure mode if we get it wrong is a single stripped inline sample — versus the inline-exclusion bugs, whose failure mode was suppressing a **live** image (the exact broken-local-path behavior this PR removes). The narrower, robust scope is the intentional trade-off; it is documented in `docs/cli-reference.md` (fenced-only + the >1 MB scan guard you flagged).

I also reverted `findInlineCodeSpans` to its prior form — it is used only by the doc-path scanner, which this PR does not otherwise touch, keeping the change image-focused.

Verification: shared 580, client image 18 + doc-snapshots 30, cli image 10; typecheck + biome clean; prior head's CI green. Re-requesting review.

packages/shared/src/lib/doc-link-scan.ts
```diff
@@ -170,6 +170,51 @@ function isClosingFence(line: string, fence: FenceState): boolean {
...
+    }
+    if (fence) {
+      if (isClosingFence(line, fence)) {
+        ranges.push({ start: fenceStart, end: absoluteOffset + line.length });
+        fence = null;
+      }
+      absoluteOffset += line.length;
+      continue;
+    }
+    const opening = parseOpeningFence(line);
```
baixiaohang: Blocker within the new fenced-only scope: this scans raw physical lines and `parseOpeningFence` only accepts up to three leading spaces, so it misses valid fenced blocks nested in Markdown containers. For `> ```md\n> literal\n> ````, ReactMarkdown renders a fenced code block inside a block quote, but this helper returns no range; image capture uploads `shot.png` and strips the authored sample. A list item whose fenced block starts after `- ` has the same failure. Please recognize container-prefixed fenced blocks (including their continuation prefixes) or use a block tokenizer, with blockquote and list regressions.

packages/client/src/runtime/image-snapshots.ts
```diff
@@ -0,0 +1,305 @@
...
+  // here.
+  const re = /!\[[^\]\n]{0,512}\]\(([^\s)"]{1,2048})(?:\s+"[^"]*")?\)/g;
+  const candidates: ImageOccurrence[] = [];
+  let match: RegExpExecArray | null = re.exec(text);
+  while (match !== null) {
+    const whole = match[0];
+    const target = match[1];
+    const start = match.index;
+    const prev = start > 0 ? text[start - 1] : "";
+    if (prev !== "\\" && target && !hasUrlScheme(target)) {
```
baixiaohang: Blocker: CommonMark escaping depends on the parity of the contiguous backslash run, not just the immediately preceding character. In source text containing two backslashes before `x`, the first slash escapes the second and `!` starts a live image; this check still skips the candidate, so the original local path is sent and renders broken. Count the preceding run and skip only when it is odd; cover both two- and three-backslash cases.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749

- Rationale: The fenced-only scope is a coherent product trade-off and removes the open-ended inline-parser mismatch, but two cases still violate the resulting contract: fenced blocks inside Markdown containers are not excluded, and live images after an even backslash run are treated as escaped.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: local workspace image embeds sent by agents currently render as broken paths for humans and peer agents.
- Approach: capture explicit embeds through the sender-only workspace fence, upload them into the chat organization, and emit the existing human image-batch shape; the revised exclusion intentionally protects fenced code blocks only and treats inline-code embeds as live.
- Impacted modules: CLI chat send/capture, client snapshot capture, shared Markdown fence scanning, Web mixed-attachment rendering, and CLI reference docs.

**Review findings**
1. Blocking: the fenced-only scanner recognizes only top-level fences, not valid fences nested in block quotes or list items, so a literal sample in those common containers is uploaded and stripped.
2. Blocking: image escaping checks only the immediately preceding character; an even run of backslashes leaves `!` unescaped in CommonMark, but capture skips it and leaves the live local image broken.

**Action taken**
- Submitted request changes with inline reproductions.

The >1 MB capture backstop and fenced-only/inline-live trade-off are now user-visible in the CLI reference. The paired Context Tree PR is still draft and must be reconciled from its current broader `outside a code span` wording before it is marked ready. I did not run tests or code checks in this review pass.

yuezengwu (changes requested):
Goal/core change reviewed: agent `chat send` captures explicit workspace-local image embeds through the sender-only fence, uploads them to the target chat organization, and converts the outbound message to the existing human `format: "file"` image-batch shape; Web only deduplicates mixed document/image presentation. This reuses the current `ImageRefContent` / image-batch structure and introduces no database migration or server-schema change.

The fenced-only product decision is coherent and removes the open-ended inline CommonMark parser surface. I am still requesting changes for two blockers inside that narrower contract:

1. `fencedCodeBlockRanges` recognizes only top-level physical-line fences. Valid fenced blocks inside Markdown containers are missed: both `> ~~~md\n> literal\n> ~~~` and a list-item fence render as code in ReactMarkdown, but this head uploads `shot.png` and strips the literal sample. Please recognize container-prefixed fences (including continuation prefixes), or use the renderer's block tokenizer, with blockquote and list regressions.

2. Image escaping checks only the immediately preceding character. With two consecutive backslashes before `live`, CommonMark treats the first slash as escaping the second and the image remains live; this head skips capture and sends the broken local path. Count the contiguous preceding backslashes and skip only on odd parity, with two- and three-backslash tests.

I independently reproduced all three failing capture assertions (blockquote fence, list fence, even-backslash image), and ReactMarkdown confirmed the corresponding render semantics. Existing focused suites remain green: Shared 32, Client 46, and CLI 60 tests.

Contract/documentation cleanup while touching this: `MAX_IMAGE_SCAN_BYTES` still compares `text.length` (UTF-16 code units), while the CLI reference calls the threshold “1 MB”; name and document the actual unit or enforce bytes. The PR body also still says inline-code images are skipped, and draft Context Tree PR #727 says “outside a code span”; both need the new fenced-only wording (the tree draft can be reconciled before it is marked ready).

packages/client/src/runtime/image-snapshots.ts
```diff
@@ -0,0 +1,305 @@
...
+  const occurrences = collectImageOccurrences(text);
+  if (occurrences.length === 0) return { imageRefs: [], strippedText: text, skipped: 0 };
+
+  const selfConfig: SelfFence = typeof self === "string" ? { agentHome: self } : self;
+  const roots = await resolveSelfRoots(selfConfig);
+  if (!roots) return { imageRefs: [], strippedText: text, skipped: occurrences.length };
+
+  // Pass 1 — resolve each mention to a readable in-fence file + its MIME.
+  type Resolved = ImageOccurrence & { file: string; mime: SupportedImageMime };
+  const resolvedList: Array<Resolved | null> = await Promise.all(
```
baixiaohang: Blocker: the 20-image limit is applied only after every syntactic occurrence has already entered this unbounded `Promise.all`. A body below the 1M-code-unit scan guard can contain tens of thousands of short `` candidates, causing the CLI to schedule tens of thousands of concurrent canonicalize/realpath/resolve operations before the attachment cap protects anything. That can exhaust file descriptors/memory or make a best-effort send appear hung even though only 20 uploads are allowed. Please bound candidate work before filesystem resolution and use bounded concurrency (with an explicit overflow/pass-through policy that preserves out-of-fence text); add a high-occurrence regression in addition to the existing >20 upload test.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1749

- Rationale: This increment correctly fixes live-image backslash parity and makes the scan-limit unit honest, but two existing blockers remain unchanged: container-nested fenced blocks are still captured destructively, and all occurrences still enter unbounded filesystem resolution before the 20-image cap.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: agent-authored workspace image embeds currently reach recipients as broken local paths.
- Approach: capture explicit embeds on the sender under the self-fence, upload into the target chat organization, and emit the existing human image-batch message shape without changing server or session architecture.
- Impacted modules: CLI send/capture, client workspace image capture, shared fenced-block scanning, Web mixed-attachment presentation, and CLI docs.

**Review findings**
✅ Backslash escape parity now follows odd/even CommonMark behavior with 1/2/3-run coverage.
✅ The `text.length` backstop is now named as characters/code units and documented as approximately one million characters.
❌ Existing blocker remains: blockquote/list container fences are not recognized, so their literal image samples are uploaded and stripped.
❌ Existing blocker remains: `Promise.all(occurrences.map(...))` performs unbounded filesystem resolution before `MAX_BATCH_ATTACHMENTS` applies.
⚠️ Contract reconciliation remains: the PR body and draft Context Tree PR #727 still describe the superseded inline-code exclusion and should match fenced-only before final readiness.

**Action taken**
- Submitted request changes without duplicating the existing inline threads.

I did not run tests or code checks in this review pass. Current CI for this head was still in progress when reviewed.

yuezengwu (changes requested):
Goal/core change re-reviewed on `0d7d91a7`: agent `chat send` captures explicit workspace-local image embeds through the sender-only fence, uploads into the target chat organization, and emits the existing human `format: "file"` image-batch shape. This still introduces no database migration or server-schema/core-persistence change.

This increment correctly closes two findings: backslash escape parity now follows CommonMark odd/even behavior with 1/2/3-run coverage, and the scan backstop is honestly named/documented as roughly one million characters rather than bytes. Draft Context Tree PR #727 has also been updated to the fenced-only wording.

I am still requesting changes for two blockers:

1. The container-fence blocker is unchanged. `fencedCodeBlockRanges` still parses raw top-level physical lines only, so blockquote- and list-contained fenced blocks return no range. ReactMarkdown renders both as code, but image capture uploads and strips the literal image sample. My two focused regression assertions still fail on this head.

2. The 20-image cap is applied only after `Promise.all(occurrences.map(...))` has launched filesystem resolution for every syntactic occurrence. A focused diagnostic with 100 occurrences observed all 100 `canonicalizeWorkspacePath` calls active concurrently before the cap; the ~1M-character body guard allows tens of thousands, each driving realpath/stat work and Promise allocation. This can stall or exhaust the sender before the best-effort fallback. Bound/short-circuit filesystem resolution before fan-out, while handling remaining caption spans explicitly so the mixed-success contract stays clear.

Existing focused suites pass on this head: Shared 32, Client 47, CLI 60. The PR body itself still says inline-code images are skipped; please update that stale description to the fenced-only decision before merge. I did not wait for CI before submitting the review.

Gandy2025:
Both remaining blockers resolved in `97e87ee3`, by switching to the renderer's own parser (per product decision).

**Container fences — now uses `mdast-util-from-markdown`.** Instead of a hand-rolled fenced scanner, capture excludes block-code ranges from the same markdown parser ReactMarkdown uses. This recognizes **fenced blocks nested in blockquotes and list items** (your two regressions), plus indented code blocks, exactly as the renderer does — no CommonMark surface left to diverge on. Inline `` `code` `` remains a live embed by design (product decision; avoids reproducing inline parsing send-side). Tests: blockquote / list-item / indented containers excluded, inline-code embed captured.

**Filesystem fan-out — bounded before resolution.** The pipeline now dedupes mentions by path and caps the number resolved to `MAX_BATCH_ATTACHMENTS` (20) **before** any `realpath`/`stat`, so a body with thousands of `` mentions launches at most 20 concurrent resolutions. Test asserts 25 mentions → exactly 20 uploads.

Also: PR body updated to the final (renderer-parser) description; the paired tree PR #727 clarified that inline-code stays live + that "fenced" means renderer-recognized (incl. container-nested).

Cost: one new dependency, `mdast-util-from-markdown` (the parser react-markdown already uses) — the block-tokenizer route you both offered.

Verification: full client 1503, shared 575, doc-snapshots 30, cli 10; typecheck + biome clean. Re-requesting review.

baixiaohang (changes requested):
Recommendation: request changes. The renderer-parser replacement correctly closes the container and indented-code mismatch, and filesystem resolution is now bounded before fan-out. One mixed-success contract blocker remains.

Risk level: A (packages/client and apps/cli).

Blocking finding — packages/client/src/runtime/image-snapshots.ts:97. With 21 valid distinct workspace images, the first 20 resolve and upload, so the message flips to the file image-batch format, but the 21st over-cap mention is deliberately left in the caption. The Web renderer then resolves that workspace-local path against the cloud origin and shows the same broken image this PR is intended to eliminate. This also contradicts docs/cli-reference.md and paired Context Tree PR #727, which state that once any sibling captures, every uncapturable workspace embed is removed from the caption.

Please make overflow behavior atomic and explicit before merge. The safest bounded policy is to pass the whole body through unchanged before any upload when the candidate set exceeds the cap; another policy is acceptable only if it preserves the documented no-broken-local-path mixed-success invariant without restoring unbounded filesystem work. Please extend the existing 25-image test to assert strippedText/caption behavior, not only upload count.

Action taken: requested changes.

yuezengwu (changes requested):
Goal/core change re-reviewed on `97e87ee3`: agent `chat send` captures explicit workspace-local image embeds under the sender-only fence and emits the existing human `format: "file"` image-batch shape. There is still no database migration, server-schema change, or persisted core-structure change.

The two prior blockers are correctly closed. Independent regressions confirm that renderer-recognized block code is preserved in blockquotes, list items, and indented blocks, while inline-code embeds remain live by the deliberate product decision. A 100-distinct-path diagnostic also confirms filesystem resolution now starts exactly 20 tasks, not 100.

I am requesting changes for the remaining mixed-success blocker. With 21 valid distinct in-workspace PNG embeds, the first 20 resolve/upload and flip the message to a `file` image batch, but the 21st path is outside `resolvedByPath` and remains verbatim as `p20` in `strippedText`. Web then renders that caption path against the cloud origin, recreating the broken image this PR fixes. This directly contradicts both `docs/cli-reference.md` and draft Context Tree PR #727, which say an uncapturable/over-cap sibling is removed once any image captures. Please make overflow behavior atomic or otherwise preserve the no-broken-local-path invariant, and extend the batch-cap test to assert `strippedText`, not only upload count.

Adjacent cap-definition coverage: the slice happens before MIME filtering, so 20 unsupported `.txt` embeds consume the whole resolution budget and suppress a following valid PNG; my focused regression also reproduces this. Please define the cap over eligible supported candidates (or explicitly document a broader candidate-attempt budget) as part of the overflow fix.

The new code also excludes indented code blocks, while the CLI reference still says only “fenced code block”; align that user-facing wording with the PR body and tree draft. Existing focused suites pass: Shared 27, Client 50, CLI 60; Client typecheck and build pass. I did not wait for CI before submitting this review.

Gandy2025:
Fixed in the latest push (mixed-success contract). When the message flips to a `file` image batch, capture now strips **every** local-path image candidate span from the caption — not only the resolved ones. So an over-cap (21st+) mention, or an out-of-fence one, is **dropped** rather than left as a `` that renders broken. Web-URL images and code-block samples are never candidates, so they stay. The filesystem fan-out is unchanged (still bounded to the 20-cap) — this only changes which spans are removed from the text. Test: 21 images strip clean; a mixed in-cap + out-of-fence message drops both from the caption. Re-requesting.

baixiaohang (changes requested):
Recommendation: request changes. The latest increment correctly closes the prior mixed-success bug: once a batch succeeds, over-cap and unresolved local-path spans are removed while filesystem fan-out remains bounded. Two candidate-boundary blockers remain.

Risk level: A (packages/client and apps/cli).

1. packages/client/src/runtime/image-snapshots.ts:106 — the 20-item slice is still applied before MIME eligibility is checked at line 114. Twenty distinct unsupported targets such as notes0.txt through notes19.txt consume the complete resolution budget, so a following valid shot.png is never attempted. Because no ref succeeds, the function returns the whole body unchanged and the valid workspace image still renders broken. Please define the cap over supported image candidates, filter unsupported paths before slicing, and add the 20 unsupported plus 1 valid regression.

2. packages/client/src/runtime/image-snapshots.ts:255 — the new strip-every-candidate behavior assumes every target without an explicit scheme is a local filesystem path. A network-path reference such as //cdn.example.com/logo.png is a valid Web image URL, but hasUrlScheme returns false; if a local sibling captures, that remote image is now stripped from the caption. This contradicts the code comment and author note that Web-URL images are never candidates. Please preserve scheme-relative Web URLs and add a mixed remote plus local regression. If root-relative targets are intentionally treated as local despite their Web semantics, document that ambiguity explicitly.

Documentation follow-up: docs/cli-reference.md still says only fenced code blocks are preserved, while the implementation and PR body also preserve indented block code. Align it with renderer-recognized block code.

Action taken: requested changes. I did not run tests or code checks.

yuezengwu (changes requested):
Goal/core change re-reviewed on `036a7793`: agent `chat send` captures explicit workspace-local image embeds under the sender-only fence, uploads them into the chat organization, and emits the existing human `format: "file"` image-batch shape. This still introduces no database migration, server-schema change, or persisted core-structure change.

The latest increment correctly closes the prior 21st-image leak: once at least one image captures and the message flips to a file batch, over-cap and unresolved local-path spans are stripped from the caption, while filesystem resolution remains bounded to 20 paths.

I am still requesting changes for two candidate-boundary blockers:

1. `packages/client/src/runtime/image-snapshots.ts:106` slices the first 20 distinct syntactic targets before MIME eligibility is checked. A focused regression with 20 distinct `.txt` targets followed by a valid in-workspace `shot.png` returns `imageRefs: []` instead of capturing `shot.png`; the valid local embed therefore remains broken. Define the cap over supported image candidates, while retaining the separate strip-all-local-candidates behavior after a sibling succeeds.

2. `packages/client/src/runtime/image-snapshots.ts:255` treats a scheme-relative Web image such as `//cdn.example.com/logo.png` as a local candidate. In a mixed remote + captured-local message, my regression expected the remote Markdown image to remain in `strippedText`, but received an empty caption. Preserve network-path Web URLs; this currently contradicts the implementation comment and author note that Web URLs are never candidates.

Please add both regressions. Also align `docs/cli-reference.md` with the actual renderer-recognized block-code boundary: the implementation and PR body preserve indented code blocks as well as fenced blocks at container depth, while the CLI reference currently says only fenced blocks.

Independent verification: the two new candidate-boundary regressions fail exactly as above. Existing focused suites remain green: Shared 27, Client image/doc 53, CLI image 10; Client typecheck and `git diff --check` pass. I did not wait for CI before submitting this review.

Gandy2025:
Both candidate-boundary blockers fixed in `87ef98a7`.

**1. Cap over supported-image candidates.** A target is a candidate only if it has a supported image extension, so 20 unsupported `` mentions are no longer candidates and can't consume the 20-path budget — a following valid `` is attempted and captured. The strip-all-local-candidates-after-a-sibling-captures behavior is unchanged. Test: 20 `.txt` + 1 valid png → the png is captured.

**2. Protocol-relative web URLs preserved.** `isWebUrl` now returns true for a `//host/…` prefix as well as an absolute scheme, so `` is never a candidate — never captured, never stripped from a flipped caption (it renders inline as-is). Test: mixed remote `//cdn…` + captured local → the remote URL stays in `strippedText`.

**Docs.** `docs/cli-reference.md` now says renderer-recognized block code (a fenced block at any container depth OR an indented block), matching the impl + PR body.

The prior mixed-success (over-cap/unresolved strip) fix and the bounded fan-out are unchanged. Verification: client image-snapshots 25, doc-snapshots 30, shared doc-link-scan 27, cli image 10; typecheck + biome clean. Re-requesting.

yuezengwu (approved):
Goal/core change re-reviewed on `87ef98a7`: agent `chat send` captures explicit supported workspace-local image embeds under the sender-only fence, uploads them into the chat organization, and emits the existing human `format: "file"` image-batch shape. This introduces no database migration, server-schema change, or persisted core-structure change.

Both remaining candidate-boundary blockers are correctly closed:

- MIME eligibility is now established while collecting occurrences, before the 20-distinct-path slice. The 20 unsupported `.txt` + 1 valid PNG regression captures the PNG, while resolution/upload fan-out remains bounded.
- `isWebUrl` now excludes protocol-relative `//host/...` targets as well as explicit schemes. The mixed remote + captured-local regression preserves the remote Markdown image in the caption.

The previous mixed-success fix remains intact: after one supported local image captures and the message flips to a file batch, over-cap and unresolved supported local candidates are removed rather than left as broken paths. The CLI reference now also matches the renderer-recognized block-code boundary (fenced at container depth or indented; inline code stays live by the deliberate product decision).

Independent verification is green: Shared doc-link scan 27, Client image/doc 55, CLI image 10, Client typecheck, Biome on the touched source/tests, and `git diff --check`. Paired Context Tree PR #727 remains draft for post-merge reconciliation. Approving without waiting for CI.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1749

**Recommendation**: approve
- Rationale: the current head closes all reproduced content-preservation, resource-bound, parser-parity, and candidate-boundary blockers while keeping the change isolated to sender-side capture and the existing image message shape.

**Risk level**: A
- Semantic lift: none (no database, server-schema, or persisted chat/message/session core-structure change)

**PR summary**
- Problem: workspace-local Markdown image embeds sent by agents currently resolve against the cloud origin and appear broken to human and agent recipients.
- Approach: capture explicit supported image embeds at `chat send` time under the sender-only workspace fence, upload into the target chat organization, and emit the existing human `format: "file"` image-batch shape. Renderer-recognized block code is preserved, filesystem and upload work is bounded, and mixed-success captions cannot retain broken local image paths.
- Impacted modules: CLI chat send/capture, Client workspace image snapshots, Web mixed document/image presentation, CLI reference documentation, and paired Context Tree PR #727.

**Review findings**
- ✅ File size is checked before allocation, resolve/upload fan-out is bounded, and pathological scan work has an explicit backstop.
- ✅ Block-code classification reuses the renderer parser; supported MIME, URL, escape, cap, and mixed-success boundaries are covered by focused regressions.
- ✅ The implementation reuses the current human image-batch persistence/rendering path without introducing a second agent-only server shape.
- ⚠️ The legacy image shape and upload-before-message orphan window remain known follow-up constraints; paired Context Tree PR #727 records the convergence and transition caveats.
- ✅ Latest CI is green, including Test, lint/typecheck, CodeQL, migrations, and portable CLI smoke. I did not run tests or code checks locally.

**Action taken**
- Approved with explicit authorization from baixiaohang. No merge action was performed.

PR #728, opened 2026-06-01
fix(server): restore unread badge when agent emits final-text turn output

yuezengwu:
## Summary
- Agent finishing a turn (`purpose: "agent-final-text"`) stopped triggering the human's chat-list red dot after PR #633 retired the 1:1 implicit `dmAuto` projection. The final-text path carries empty `metadata.mentions` by construction, and `applyAfterFanOut` early-returns before any `unread_mention_count` bump.
- Add a final-text-specific UPSERT branch in `chat-projection.applyAfterFanOut` that bumps unread only for human stakeholders: human speakers (≠ sender) in this chat, and watchers whose managed agent IS the sender. `forceSilentFanOut` still holds — no agent session is woken; only the human red dot returns.
- `message.ts` plumbs `isAgentFinalText` through as `bumpForAgentFinalText`.

## Why this scope
- Other agent speakers are deliberately NOT bumped: final-text never wakes another agent, and we don't want to pollute non-human unread state with "another agent finished" signals.
- 1:1 dmAuto + `extractMentionsFromContent` are not reintroduced — the fix lives in the projection, not in routing.

## Reproduction (QA, staging, 2026-06-01)
- Chat `9a3a8292-1bad-4105-b5e8-665b3bada304` (human ↔ `yzw-assistant`): after agent final text at `08:38:20Z`, human-side `/orgs/:orgId/chats` returned `unreadMentionCount=0`. Manual `POST /chats/:chatId/unread` could push it to 1 → confirms badge data source works; problem isolated to the projection path.
- Full QA repro report: `/Users/yuezengwu/.first-tree-staging/data/workspaces/qa/.qa-local/unread-final-badge-repro-20260601/report.md`
- QA regression report on the fix: `/Users/yuezengwu/.first-tree-staging/data/workspaces/qa/.qa-local/final-text-unread-fix-regression-20260601/report.md`

## Test plan
- [x] `pnpm --filter @first-tree/server exec vitest run src/__tests__/agent-final-text-purpose.test.ts` — 9/9 (3 new + 6 existing)
- [x] `pnpm --filter @first-tree/server exec vitest run src/__tests__/me-chat-service.test.ts src/__tests__/direct-chat-auto-mention.test.ts` — 31/31 (adjacent regression)
- [x] `pnpm --filter @first-tree/server test` — 152 files / 1262 tests pass
- [x] `pnpm check` — pass (1227 files)
- [x] `pnpm typecheck` — pass (13/13 turbo tasks)
[...]

PR #1750, opened 2026-07-13
feat(cli): enforce strict Context Tree validation

baixiaohang:
## Summary

- replace the legacy frontmatter regexes with one structured YAML parser and a shared normal/archive-supporting/member/repo-infra classifier
- make `tree verify` enforce deterministic metadata, `soft_links`, authority-boundary, path-escape, and symlink rules while preserving the existing check/error JSON fields and adding stable structured findings
- classify existing link and symlink targets by canonical real path; reject dangling, non-regular, escaping, directory-target, and cross-content-class Markdown symlinks before content-class skips
- parse Markdown links through mdast, keep ordinary missing Markdown targets allowed, and skip archive/supporting content as non-canonical evidence
- align the generated Context Tree policy with valid wildcard ownership YAML (`owners: ["*"]`)

## Breaking validation change

`tree verify` now always applies the current strict structural policy. There is no version marker, warning compatibility mode, or `--strict` alias. Existing trees with malformed metadata, normal-to-archive links, broken `soft_links`, unsupported directory symlinks, dangling/non-regular/escaping Markdown symlinks, or cross-content-class aliases must be repaired before their next tree PR can pass.

Mechanical syntax and path findings can be fixed directly. Ownership assignment and promotion of durable material out of archive/supporting content still require the relevant human or source-backed decision.

## Compatibility

- existing `checks.*`, `errors[]`, `ok`, and exit behavior remain intact
- `findings` and `scannedByContentClass` are additive JSON fields
- personal member Markdown keeps its relaxed optional-frontmatter behavior
- existing member identity, tree init, and tree browser contracts remain covered
- source/tree identity readers consume only contained, canonical repo-infra regular files; invalid managed paths yield strict findings before source-repo preflight
- the sole managed file-symlink exception is the historical runtime-managed root-level `WHITEPAPER.md` pointer; writers must not add it to new trees

## Validation

- final path/symlink targeted tests: 4 files, 58 tests passed
- generated briefing tests: 25 tests passed
- sanitized full CLI suite: 18/18 batches passed (tree batch: 117 tests)
- `pnpm --filter first-tree-dev build`
- `pnpm typecheck`
- `pnpm check` (only existing repository warnings/info)
- `pnpm validate:skill`
- built CLI against the companion official tree: PASS, 0 findings
- installed previous CLI against the companion official tree: PASS
- `git diff --check`

## Companion Context Tree

Draft companion Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/728

Merge this code PR first, then reconcile and ready the tree PR.

### DIFF
--- a/apps/cli/src/commands/tree/binding-contract.ts
+++ b/apps/cli/src/commands/tree/binding-contract.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync } from "node:fs";
+import { existsSync, readFileSync, statSync } from "node:fs";
 import { dirname, join, resolve } from "node:path";
 
 import { readSourceState, type SourceBindingMode, type TreeMode } from "./binding-state.js";
@@ -157,6 +157,14 @@ export function readManagedSourceBinding(root: string): ManagedSourceBinding | u
     const path = join(root, file);
 
     if (!existsSync(path)) {
+      continue;
+    }
+
+    try {
+      if (!statSync(path).isFile()) {
+        continue;
+      }
+    } catch {
       continue;
     }
 
--- a/apps/cli/src/commands/tree/binding-contract.ts
+++ b/apps/cli/src/commands/tree/binding-contract.ts
@@ -1,7 +1,7 @@
-import { existsSync, readFileSync, statSync } from "node:fs";
 import { dirname, join, resolve } from "node:path";
 
 import { readSourceState, type SourceBindingMode, type TreeMode } from "./binding-state.js";
+import { inspectRepoInfraMarkdownFile } from "./content-class.js";
 import { parseGitHubRemoteUrl } from "./shared.js";
 
 export type SourceIntegrationFile = "AGENTS.md" | "CLAUDE.md";
@@ -155,20 +155,12 @@ export function parseManagedSourceBindingText(text: string): ParsedSourceBinding
 export function readManagedSourceBinding(root: string): ManagedSourceBinding | undefined {
   for (const file of SOURCE_INTEGRATION_FILES) {
     const path = join(root, file);
-
-    if (!existsSync(path)) {
+    const inspected = inspectRepoInfraMarkdownFile(root, file);
+    if (inspected.kind !== "valid") {
       continue;
     }
 
-    try {
-      if (!statSync(path).isFile()) {
-        continue;
-      }
-    } catch {
-      continue;
-    }
-
-    const parsed = parseManagedSourceBindingText(readFileSync(path, "utf-8"));
+    const parsed = parseManagedSourceBindingText(inspected.source);
 
     if (parsed !== undefined) {
       return {
--- a/apps/cli/src/commands/tree/content-class.ts
+++ b/apps/cli/src/commands/tree/content-class.ts
@@ -8,6 +8,8 @@ export type ContextContentClassCounts = Record<ContextContentClass, number>;
 
 export type ContextMarkdownFile = {
   absolutePath: string;
+  canonicalContentClass: ContextContentClass;
+  canonicalRelativePath: string;
   contentClass: ContextContentClass;
   escaped: boolean;
   relativePath: string;
@@ -26,7 +28,7 @@ export type ContextMarkdownCollection = {
 
 const GENERATED_DIRECTORY_NAMES = new Set(["node_modules", "__pycache__", "dist", "build", ".next", ".turbo"]);
 const REPO_INFRA_MARKDOWN_FILES = new Set(["AGENTS.md", "CLAUDE.md"]);
-const MANAGED_SYMLINK_FILES = new Set(["WHITEPAPER.md"]);
+const MANAGED_SYMLINK_PATHS = new Set(["WHITEPAPER.md"]);
 
 export function toTreeRelativePosixPath(treeRoot: string, targetPath: string): string {
   return relative(treeRoot, targetPath).replace(/\\/gu, "/");
@@ -80,6 +82,7 @@ function pathIsInside(root: string, target: string): boolean {
 export function collectContextMarkdownContent(treeRoot: string): ContextMarkdownCollection {
   const directorySymlinks: ContextDirectorySymlink[] = [];
   const files: ContextMarkdownFile[] = [];
+  const realTreeRoot = realpathSync(treeRoot);
 
   function walk(directoryPath: string): void {
     for (const entry of readDirectoryEntries(directoryPath)) {
@@ -116,7 +119,7 @@ export function collectContextMarkdownContent(treeRoot: string): ContextMarkdown
         continue;
       }
 
-      if (symbolicLink && MANAGED_SYMLINK_FILES.has(entry.name)) {
+      if (symbolicLink && MANAGED_SYMLINK_PATHS.has(relativePath)) {
         continue;
       }
 
@@ -129,15 +132,29 @@ export function collectContextMarkdownContent(treeRoot: string): ContextMarkdown
       }
 
       let escaped = false;
+      let canonicalContentClass = contentClass;
+      let canonicalRelativePath = relativePath;
       if (symbolicLink) {
         try {
-          escaped = !pathIsInside(realpathSync(treeRoot), realpathSync(absolutePath));
+          const realTarget = realpathSync(absolutePath);
+          escaped = !pathIsInside(realTreeRoot, realTarget);
+          canonicalRelativePath = toTreeRelativePosixPath(realTreeRoot, realTarget);
+          if (!escaped) {
+            canonicalContentClass = classifyContextContent(canonicalRelativePath);
+          }
         } catch {
           continue;
         }
       }
 
-      files.push({ absolutePath, contentClass, escaped, relativePath });
+      files.push({
+        absolutePath,
+        canonicalContentClass,
+        canonicalRelativePath,
+        contentClass,
+        escaped,
+        relativePath,
+      });
     }
   }
 
--- a/apps/cli/src/commands/tree/content-class.ts
+++ b/apps/cli/src/commands/tree/content-class.ts
@@ -1,5 +1,5 @@
 import type { Dirent } from "node:fs";
-import { readdirSync, realpathSync, statSync } from "node:fs";
+import { lstatSync, readdirSync, readFileSync, realpathSync, statSync } from "node:fs";
 import { isAbsolute, join, relative } from "node:path";
 
 export type ContextContentClass = "normal" | "archive-supporting" | "member" | "repo-infra";
@@ -27,6 +27,8 @@ export type ContextMarkdownCollection = {
   directorySymlinks: ContextDirectorySymlink[];
   files: ContextMarkdownFile[];
 };
+
+export type RepoInfraMarkdownInspection = { kind: "absent" } | { kind: "invalid" } | { kind: "valid"; source: string };
 
 const GENERATED_DIRECTORY_NAMES = new Set(["node_modules", "__pycache__", "dist", "build", ".next", ".turbo"]);
 const REPO_INFRA_MARKDOWN_FILES = new Set(["AGENTS.md", "CLAUDE.md"]);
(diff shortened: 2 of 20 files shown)


apps/cli/src/commands/tree/validate-nodes.ts
```diff
@@ -1,177 +1,265 @@
...
+      directory.escaped
+        ? "directory symlink resolves outside the Context Tree root"
+        : "Context Tree domain directories must not be symlinks",
+    );
+  }
+
+  for (const file of content.files) {
+    scannedByContentClass[file.contentClass] += 1;
+
+    if (file.contentClass === "repo-infra" || file.contentClass === "archive-supporting") {
```
yuezengwu: Blocker: enforce `file.escaped` before skipping non-canonical classes. `collectContextMarkdownContent` records escape state for every Markdown-file symlink, but this early `continue` discards it for archive/supporting and repo-infra files. A fixture with `raw-context/escaped.md -> /outside/outside.md` returns zero findings and passes node validation; an escaping `AGENTS.md` behaves the same. That contradicts the new documented contract that Markdown symlinks must stay inside the tree except for the managed `WHITEPAPER.md` pointer. Please move the escape check ahead of the class skip and add archive/repo-infra regressions (and scope the managed exemption to the actual managed path rather than any basename match).

apps/cli/src/commands/tree/context-links.ts
```diff
@@ -0,0 +1,126 @@
...
+  const sourceDirectory = posix.dirname(options.sourcePath);
+  const relativePath = posix.normalize(
+    options.softLink || withoutSuffix.startsWith("/")
+      ? withoutSuffix.replace(/^\/+/, "")
+      : posix.join(sourceDirectory, withoutSuffix),
+  );
+  const absoluteRoot = resolve(options.treeRoot);
+  const absoluteTarget = resolve(absoluteRoot, relativePath);
+  const lexicalEscape = !pathIsInside(absoluteRoot, absoluteTarget);
+  const contentClass = classifyContextContent(relativePath);
```
yuezengwu: Blocker: classify an existing target by its canonical path, not by the spelling of the link. With `alias.md -> raw-context/proposal.md`, both the file collector and this resolver classify `alias.md` as normal; a normal node containing `[archive](alias.md)` then produces zero findings even though it resolves to archive/supporting evidence. Case aliases on case-insensitive filesystems create the same platform-dependent bypass. After containment is established, derive the effective class from the real target relative to the real tree root, and make the file collector prevent symlinks from crossing content-class boundaries.

apps/cli/src/commands/tree/context-links.ts
```diff
@@ -0,0 +1,126 @@
...
+    return decodeURIComponent(target);
+  } catch {
+    return target;
+  }
+}
+
+export function isTreeLocalTarget(target: string): boolean {
+  const trimmed = target.trim();
+  return (
+    trimmed.length > 0 && !trimmed.startsWith("#") && !trimmed.startsWith("//") && !/^[a-z][a-z\d+.-]*:/iu.test(trimmed)
```
yuezengwu, line 42: Blocker for Windows: a drive-absolute target such as `C:/outside.md` matches the URI-scheme regex and is treated as external before path normalization. Because ordinary missing/external Markdown targets are allowed, a normal Markdown link to a Windows path outside the tree receives no path-escape finding. Detect Windows drive/UNC absolute paths before excluding URI schemes, then run the same containment check and add platform-independent regression cases.

apps/cli/src/commands/tree/validate-members.ts
```diff
@@ -143,26 +143,57 @@ export function runValidateMembers(treeRoot: string): {
...
+          const path = `${toTreeRelativePosixPath(treeRoot, childPath)}/`;
+          addFinding(findings, VALIDATION_CODES.memberNodeMissing, path, "directory exists but is missing NODE.md");
           walk(childPath, false);
         }
         continue;
       }
 
       memberCount += 1;
-      allErrors.push(...validateMember(nodePath, treeRoot));
+      findings.push(...validateMember(nodePath, treeRoot));
```
yuezengwu, line 176: Blocker: the member walk still follows directory symlinks via `statSync` before reaching this validation/recursion. Although the new content collector reports directory symlinks as unsupported, this independent pass descends them anyway. A `members/alice/loop -> ..` fixture returned 93 repeated member findings with ever-growing paths until the filesystem path limit; a link to another tree can also make this validator read nodes outside the requested root. Use `Dirent`/`lstatSync` to skip symlink directories here (and optionally track visited realpaths) once the structural symlink finding has been emitted.

yuezengwu (changes requested):
The implementation goal is clear: consolidate Context Tree parsing and classification, make `tree verify` the single strict structural gate, and add stable machine-readable findings without replacing the legacy summary fields. This is a core Context Tree data-model and breaking validation change, so the rollout boundary and companion tree repair need explicit human attention. The parser/classifier consolidation is coherent, and the changed tree tests (76/76 in my focused run) plus CLI typecheck pass locally. I am requesting changes because the path, symlink, and authority guarantees still have four deterministic bypasses described inline.

apps/cli/src/commands/tree/content-class.ts
```diff
@@ -0,0 +1,163 @@
...
+            if (contentClass !== "repo-infra") {
+              directorySymlinks.push({
+                contentClass,
+                escaped: !pathIsInside(realpathSync(treeRoot), realpathSync(absolutePath)),
+                relativePath,
+              });
+            }
+            continue;
+          }
+        } catch {
```
yuezengwu, line 194: Blocker: a dangling non-managed Markdown symlink is discarded here before `.md` classification. With `dangling.md -> missing-target.md`, `runValidateNodes` returns `{ exitCode: 0, errors: [] }`; the existing `ignores entries that disappear before stat` test also locks in that result. The new documented contract says Markdown file symlinks are validated and must resolve inside the tree, but a Git-tracked dangling link does neither while `tree verify` can pass. Please distinguish an unresolved `.md` symlink from a transient regular entry, emit a stable finding, and add a strict regression (retaining the root `WHITEPAPER.md` exemption if intentional).

apps/cli/src/commands/tree/context-links.ts
```diff
@@ -0,0 +1,144 @@
...
+function decodeTarget(target: string): string {
+  try {
+    return decodeURIComponent(target);
+  } catch {
+    return target;
+  }
+}
+
+function isWindowsAbsoluteTarget(target: string): boolean {
+  return /^[a-z]:[\\/]/iu.test(target) || /^\\\\[^\\]+\\[^\\]+/u.test(target);
```
yuezengwu: Blocker for Windows root-relative absolute paths: this detects drive and UNC forms but misses a single leading backslash. `path.win32.isAbsolute("\\outside.md")` is true, mdast preserves the target, yet the resolver returns `{ escaped: false, exists: false, relativePath: "outside.md" }`. Because ordinary missing Markdown targets are allowed, `[x](\outside.md)` receives no finding. Treat a leading backslash as a Windows absolute escape before separator normalization, while retaining the intentional `/foo` Context-Tree-root syntax, and add a platform-independent regression.

yuezengwu (changes requested):
Follow-up review of `08f0fc4`: the canonical-realpath classifier, archive/repo-infra escape ordering, Windows drive/UNC handling, root-only `WHITEPAPER.md` exemption, and `Dirent`-based member walk address all four prior inline blockers. The focused path/symlink suite passes locally (3 files, 49 tests), as does CLI typecheck. The strict symlink/path contract still has two deterministic false-pass paths described inline, so I cannot approve this revision yet.

apps/cli/src/commands/tree/content-class.ts
```diff
@@ -0,0 +1,179 @@
...
+        continue;
+      }
+
+      if (symbolicLink && MANAGED_SYMLINK_PATHS.has(relativePath)) {
+        continue;
+      }
+
+      try {
+        if (!statSync(absolutePath).isFile()) {
+          continue;
```
yuezengwu, line 224: Blocker: reject symbolic links whose resolved target is neither a regular file nor a directory instead of silently continuing here. On Unix, a Git-tracked `NODE.md -> /dev/null` reaches this branch (`statSync(...).isFile()` is false), is omitted from `files`, and—because `rootNodeExists` now accepts any symlink—`verifyTreeRoot` returns `ok: true` with zero findings. Non-root `.md` links to sockets/FIFOs/devices likewise bypass all file-symlink and containment validation, despite the documented sole exception being root `WHITEPAPER.md`. Please classify any non-regular symlink target as invalid (while retaining escape detection) and add a regression; a platform-gated special-file fixture is sufficient.

apps/cli/src/commands/tree/verify.ts
```diff
@@ -83,28 +79,82 @@ function formatSourceRepoError(targetRoot: string): string {
...
+  VALIDATION_CODES.ownersMissing,
+  VALIDATION_CODES.ownersInvalid,
+  VALIDATION_CODES.markdownFileSymlinkBroken,
+  VALIDATION_CODES.markdownFilePathEscape,
+]);
+
+function rootNodeExists(path: string): boolean {
+  try {
+    const entry = lstatSync(path);
+    return entry.isFile() || entry.isSymbolicLink();
```
yuezengwu: Blocker for the promised legacy check compatibility: `isSymbolicLink()` also accepts a symlink to a directory. With the real root node removed and `NODE.md -> system/`, node validation emits `TREE_DIRECTORY_SYMLINK_UNSUPPORTED`, but `checks.rootNodeFrontmatter` is now `{ ok: true }`; the prior `statSync(path).isFile()` path made that check fail. Keep broken-file-symlink reporting precise without treating a directory symlink as a valid root-node presence (or include the directory-symlink finding in the root check), and add a regression for the legacy check row.

yuezengwu (changes requested):
The implementation goal remains clear: replace the legacy frontmatter regexes with shared structured YAML parsing/content classification, then make `tree verify` enforce strict metadata, authority, path, and symlink rules while preserving the legacy check/error fields and adding structured findings. On `6d973033`, both blockers from the prior round are fixed: dangling Markdown symlinks now produce a stable finding, and single-leading-backslash Windows root-relative paths are treated as escapes. The focused path/symlink suite passes locally (3 files, 51 tests), as does CLI typecheck.

This is still a core Context Tree data-model and breaking validation change, so the rollout boundary and companion-tree reconciliation need human attention. I am requesting changes because one Git-trackable symlink shape can still make strict verification false-pass, and the new root-presence check regresses an existing `checks.rootNodeFrontmatter` result; both are described inline.

apps/cli/src/commands/tree/binding-contract.ts
```diff
@@ -160,6 +160,14 @@ export function readManagedSourceBinding(root: string): ManagedSourceBinding | u
       continue;
     }
 
+    try {
+      if (!statSync(path).isFile()) {
```
yuezengwu: Blocker: `statSync` follows regular symlinks, so this guard still lets external content preempt strict path validation. Repro: make an otherwise valid tree whose root `AGENTS.md` symlinks to an outside regular file containing a `FIRST-TREE-SOURCE-INTEGRATION` block. `readManagedSourceBinding` parses that outside block and `verifyTreeRoot` throws the generic “This repo only has source/workspace integration installed” error before it can return `TREE_MARKDOWN_FILE_PATH_ESCAPE`. The PR promises canonical containment findings for escaping repo-infra symlinks. Please apply the same containment/content-class policy before parsing managed blocks (or otherwise ensure strict findings win) and add this source-marker regression.

apps/cli/src/commands/tree/verify.ts
```diff
@@ -83,28 +79,95 @@ function formatSourceRepoError(targetRoot: string): string {
...
+      return "Root NODE.md symlink target cannot be resolved.";
+    case VALIDATION_CODES.markdownFileSymlinkUnsupported:
+      return "Root NODE.md symlink target is not a regular file.";
+    default:
+      return formatValidationFinding(finding);
+  }
+}
+
+export function verifyTreeRoot(targetRoot: string): VerifySummary {
+  if (readSourceBindingContract(targetRoot) !== undefined && readTreeIdentityContract(targetRoot) === undefined) {
```
yuezengwu: Blocker: the new non-file guard is not closed over both readers in this preflight. With `AGENTS.md -> system/` and a regular `CLAUDE.md` containing a source-binding block, `readSourceBindingContract` skips the directory and returns the CLAUDE binding; `readTreeIdentityContract` then re-reads `AGENTS.md` through its unguarded `readFileSync` and throws `EISDIR`. The structural collector never gets to return `TREE_DIRECTORY_SYMLINK_UNSUPPORTED`, contrary to the stated compatibility behavior. Make both preflight readers share the invalid-path policy (including mixed AGENTS/CLAUDE cases), and add a regression that asserts a structured verify summary rather than an exception.

yuezengwu (changes requested):
Incremental review of `4df6b369`: the implementation still targets one structured frontmatter/content classifier and a strict `tree verify` gate with additive findings while preserving legacy checks. Both blockers from the prior round are addressed: non-regular Markdown symlink targets now produce a stable unsupported/path finding, and a root `NODE.md` directory symlink again fails the legacy root check. The final focused suite passes locally (4 files, 57 tests), as does CLI typecheck.

This remains a core Context Tree data-model and breaking validation change, so the rollout boundary and companion-tree reconciliation require human attention. I am requesting changes because the source/tree identity preflight can still consume invalid `AGENTS.md` / `CLAUDE.md` paths before strict collection: one shape returns the wrong source-repo exception instead of the promised path finding, and another still crashes with `EISDIR`. Both are deterministic and described inline.

yuezengwu (approved):
Final incremental review of `8eed5a50`: the implementation goal is to replace the legacy frontmatter regexes with shared structured YAML/content classification and make `tree verify` the single strict metadata, authority, path, and symlink gate while preserving legacy check/error fields and adding structured findings. The new shared repo-infra inspection now ensures source/tree identity readers consume only contained canonical regular files, and invalid `AGENTS.md` / `CLAUDE.md` paths bypass source-repo preflight so the strict collector can report them. Both prior managed-path reproductions now return the expected findings rather than a generic source error or `EISDIR`; all earlier path, symlink, Windows, member-walk, and root-check blockers remain addressed.

Local validation passes: the final focused set is 4 files / 58 tests, the adjacent identity/workspace/init set is 4 files / 33 tests, CLI typecheck passes, and `git diff --check` is clean. Companion Context Tree PR #728 remains draft with the required code-first reconciliation sequence. This is a core Context Tree data-model and breaking validation change, so human review of rollout and owner-authorized companion-tree changes remains important, but I found no remaining code blocker.

PR #1751, opened 2026-07-13
feat: finish workspace chat pin — optimistic reorder, cross-device sync, a11y (PR5)

Gandy2025:
## Summary

PR5 (final) of the workspace chat-list refactor. Finishes the chat **Pin** feature end to end:

- **Optimistic reorder** — Pin/Unpin moves the row into / out of the **Pinned** group instantly. Previously it was `invalidate → refetch → regroup`, a visible lag between the click and the row moving. Rolls back + surfaces the existing error toast on failure.
- **Cross-device sync** — a pin is private per-user state (`chat_user_state.pinned_at`). A new **user-scoped** `me_chats_changed` WS channel invalidates the acting user's *other* sockets so their rail regroups in realtime — without broadcasting the private state to any other member.
- **a11y** — the row-actions kebab (the only Pin entry point) was hover-only and therefore unreachable on touch (phones + the narrow-overlay rail). `pointer-coarse:opacity-100` keeps it visible on coarse pointers (desktop hover-reveal is unchanged). The selected row now exposes `aria-current="page"`.
- **Group-by label** — `Recent activity` wrapped to two lines in the narrow rail header; shortened to **Recent** (+ `whitespace-nowrap` on the trigger so no label wraps).

## Design decisions

- **Reusing `chat:updated` was rejected.** `notifyChatUpdated` fans to *every* chat member's sockets — it wouldn't leak pin state (each recipient's refetch returns their own `pinnedAt`), but it violates "a private pin must not touch another member's devices" and needlessly invalidates chat-detail for everyone. The new channel is fanned **only** to sockets matching `humanAgentId` + `organizationId`, and carries a bare `{"type":"me-chats:changed"}` (no chatId, no pin value).
- **Optimistic transform mirrors the server projection contract** (`shared/schemas/me-chat.ts`): `attention` / `pinned` are disjoint (a chat already in Needs-attention is not also pinned — attention wins, so pinning it only flips `pinnedAt`); `pinned` is `pinnedAt` DESC (freshly-pinned prepends); ordinary `rows` are additive (the list dedup drops any priority chat, so inserting into `pinned` removes it from the recency stream with no extra work).
- **Mobile `/m/*` parity is deliberately out of scope** — it is a dev/staging-only surface; production phones use the narrow-overlay rail, which already has PR4 + this PR's touch fix.

## Changes

**web**
- `optimistic-pin.ts` (new) — pure `InfiniteData` transform, doubles as the rollback source.
- `row-engagement-menu.tsx` — `pinMut` gains `onMutate` (cancel + snapshot + apply) / `onError` (rollback + toast) / `onSettled` (reconcile invalidate); `pointer-coarse` touch reveal.
- `use-admin-ws.ts` — a `me-chats:changed` branch that throttle-invalidates `["me","chats"]`.
- `index.tsx` — `aria-current` on the selected row.

**server** (mirrors the existing `chat_updated_events` recipe)
- `services/notifier.ts` — new `me_chats_changed` channel + `notifyMeChatsChanged(humanAgentId, organizationId)` / `onMeChatsChanged`.
- `api/orgs/ws.ts` — `dispatchMeChatsChanged`: per-user (`humanAgentId` + org) socket filter, no chat-audience lookup.
- `api/chats.ts` — the pin route emits `notifyMeChatsChanged` (fire-and-forget).

## Tests

- `optimistic-pin.test.ts` (8) — promote / prepend / unpin / attention-wins / idempotent / multi-page / immutability / absent-list.
- `conversation-list-dom` — `aria-current` on selection + the touch-reveal class on the kebab.
- `notifier-extra` — `me_chats_changed` round-trip parse (+ malformed payload dropped).
- `ws-admin-edge` — per-user dispatch isolation: the acting user's socket receives the frame; a different user's frame and a different org are both filtered out.
- Full web suite green; full server suite green.
- QA case: `packages/qa/cases/cross-surface/chat-pin-cross-device-sync.md` (live 2-session cross-device sync + cross-user privacy boundary — the WS surface these product tests cannot prove end to end).

## Known limitations (LOW, self-healing)

- The optimistic reorder remounts the row under a new group, so an in-flight pin mutation's component unmounts mid-flight. On the common success path the acting device also receives its own `me-chats:changed` frame, which reconciles the list; a rare pin *failure* that races the remount may skip the instant rollback and self-heal on the 30s poll instead. Reconcile always lands via the trailing invalidate + poll.
- On unpin, a Pinned-only chat sitting beyond the loaded page briefly leaves the rail until the refetch places it at its true recency slot (end state correct).

## Related

Final PR of the chat-list refactor: PR1 #1605 · PR2 #1606 · PR3 #1738 · PR4 #1746 · PR4b #1748. Context Tree #721 / #724.


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1751

- Rationale: The coarse-pointer accessibility fix makes the row-actions trigger reachable, but it permanently overlaps the row's existing trailing status metadata on touch devices.

**Risk level**: B-high
- Path baseline: includes `packages/server/**` outside the critical service set -> B-low.
- Semantic lift: the Web change adds a new admin-WS frame branch that invalidates the conversation-list data layer -> B-high.

**PR summary**
- Problem: Workspace users need Pin/Unpin to regroup immediately, synchronize privately across their own devices, and remain reachable from touch-based narrow rails.
- Approach: Apply an optimistic infinite-query transform, publish a user-and-org-scoped bare WS invalidation, expose the row actions on coarse pointers, mark the selected row for assistive technology, and shorten the grouping label.
- Impacted modules: Workspace conversation rail and query cache, admin WebSocket client/gateway, PostgreSQL notifier, chat pin route, and cross-surface QA guidance.

**Review findings**
❌ 1. The new coarse-pointer trigger and the existing trailing metadata occupy the same slot. `pointer-coarse:opacity-100` makes the 28×28 absolutely positioned button at the row's top-right permanently visible, while the time / Watching / Working cluster is hidden only by `group-hover` or `aria-expanded`; neither applies before a touch tap. As a result, every coarse-pointer row paints the kebab over its current state signal. Please either reserve a dedicated action slot or hide/reflow the metadata on coarse pointers, and cover that relationship in the DOM assertion. [R5 / `packages/web/src/pages/workspace/conversations/row-engagement-menu.tsx:62`; `packages/web/src/pages/workspace/conversations/index.tsx:598`]

✅ 2. The realtime path preserves the private-pin boundary: the notifier carries only the acting human-agent/org identity, the gateway filters both dimensions, and the browser receives a bare invalidation with no chat or pin payload.

✅ 3. The optimistic transform follows the shipped priority contract (Attention before private Pinned before additive recency) and avoids restoring an unrelated whole-cache snapshot on failure.

**Action taken**
- Submitted request changes on head `2be3d63a`; no code, tests, or Context Tree files were modified, and no checks were run locally per review policy.

yuezengwu (changes requested):
## Summary

This PR completes the private per-user Pin flow across the Workspace rail: an optimistic priority-list transform, a user-and-org-scoped `me-chats:changed` invalidation, touch/a11y affordances, and the compact group label. The server-side identity boundary and the Attention → Pinned → recency projection both match the durable contract.

## Blocking finding

The pin mutation callbacks derive the submitted direction from the live `pinned` prop (`row-engagement-menu.tsx:114-130`). That prop is changed by `onMutate` itself. For an Attention row, pinning only flips `pinnedAt`; Attention still wins, so the row stays in the same bucket and this same `RowEngagementMenu` instance rerenders while the request is pending. TanStack Query then updates the pending mutation's observer options to the new render's closures.

As a result, the completion path observes the *new UI state*, not the operation that was submitted:

- Pin an Attention chat: the request correctly sends `true`, but success says **Unpinned**.
- If that Pin fails, `onError` calls `applyPinToCaches(true)` instead of reverting to false and reports **Couldn't unpin**, so the optimistic state remains until reconciliation.
- Unpin has the symmetric failure.

I reproduced this on the current head with a subscribed Attention-row cache and a deferred `pinMeChat`: after the optimistic render changed the menu to Unpin, resolving the original Pin produced `{"title":"Unpinned"}` instead of `{"title":"Pinned"}`.

Please pass `nextPinned` as the mutation variable and use that variable (plus an `onMutate` context containing the previous state, if useful) in `mutationFn`, rollback, and toast callbacks. Add a DOM regression test whose cached Attention row rerenders in place while the request is delayed; the current menu-only tests do not subscribe to the cache, so they cannot expose this lifecycle.

The separate coarse-pointer metadata-overlap blocker already raised by baixiaohang remains valid; I am not duplicating that inline finding here.

## Verification

- Clean PR tests: 44 focused Web tests and 8 focused server notifier/WS tests passed locally.
- Additional diagnostic test reproduced the callback-direction bug above, then was removed; the review worktree is clean.
- CI is green, but it does not cover the subscribed Attention-row mutation lifecycle.

Gandy2025:
Thanks both — both were real bugs, fixed in `11d351f2`.

**@yuezengwu — pin direction across the in-place rerender.** The callbacks read the live `pinned` prop, which `onMutate` flips: an Attention row that gets pinned stays in its bucket (Attention wins) and rerenders in place while the request is pending, and TanStack re-binds the pending mutation's callbacks to the new render's closures. Threaded the target direction as the mutation **variable** `nextPinned`, captured at click time, and use it in `mutationFn` / `onMutate` / `onError` / `onSuccess` (and `mutate(!pinned)` at the call site). Added a cache-subscribed DOM regression in `conversation-list-dom.test.tsx` that pins an Attention row with a deferred `pinMeChat`, so the optimistic rerender lands before resolution — it now asserts a "Pinned" toast (and `not` "Unpinned"), which fails on the old prop-reading code.

**@baixiaohang — coarse-pointer overlap.** Hid the trailing metadata cluster on coarse pointers (`pointer-coarse:opacity-0` on `index.tsx`) so the always-visible kebab owns the top-right corner without overlapping the time / watching / working signal — mirroring the desktop hover swap. Asserted the swap in the touch DOM test.

Full web suite green (1389), biome + typecheck + tokens clean. No server changes in this round.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1751

**Recommendation**: approve
- Rationale: Head `11d351f2` resolves both prior blockers without changing the already-sound private realtime boundary or priority projection.

**Risk level**: B-high
- Path baseline: includes `packages/server/**` outside the critical service set -> B-low.
- Semantic lift: the Web change adds an admin-WS frame branch that invalidates the conversation-list data layer -> B-high.

**PR summary**
- Problem: Workspace users need Pin/Unpin to regroup immediately, synchronize privately across their own devices, and remain reachable from touch-based narrow rails.
- Approach: Apply an optimistic infinite-query transform aligned with Attention -> Pinned -> recency, publish a user-and-org-scoped bare WS invalidation, expose touch row actions and selected-row semantics, and keep the narrow grouping control compact.
- Impacted modules: Workspace conversation rail and query cache, admin WebSocket client/gateway, PostgreSQL notifier, chat pin route, and cross-surface QA guidance.

**Review findings**
✅ 1. The submitted Pin direction is now a mutation variable captured at click time and used by `mutationFn`, optimistic apply, rollback, and toast callbacks, so an Attention row's in-place optimistic rerender cannot reverse completion behavior. The cache-subscribed deferred-request DOM regression covers that lifecycle.

✅ 2. The coarse-pointer action trigger now has exclusive use of the trailing corner: the existing time / Watching / Working cluster explicitly hides on coarse pointers, and the touch DOM assertion covers both sides of the swap.

✅ 3. The realtime path preserves the private-pin boundary: dispatch filters `humanAgentId + organizationId`, and the browser receives a bare invalidation carrying neither chat identity nor pin state.

✅ 4. The optimistic transform remains consistent with the durable list contract: Attention wins over private Pinned, pinned order follows the pin timestamp, and additive ordinary rows are rendered exactly once.

**Action taken**
- Approved head `11d351f2` with explicit authorization from baixiaohang. No code, tests, or Context Tree files were modified, and no checks were run locally per review policy.

yuezengwu (approved):
## Re-review — `11d351f2`

The PR's end-to-end goal remains sound: instant private Pin regrouping, same-user cross-device invalidation, and an accessible rail entry point. The two Web blockers from the first review are now fixed:

- Pin direction is captured as the `nextPinned` mutation variable and used by the request, optimistic apply, rollback, and toast callbacks. The new cache-subscribed deferred-response test exercises the exact Attention-row in-place rerender that previously rebound the callbacks and verifies the submitted Pin still reports **Pinned**.
- On coarse pointers, the trailing time/watching/working metadata now yields the shared top-right slot while the kebab is permanently visible, so the two affordances no longer paint over each other. The DOM test covers both sides of that swap.

No new findings. The existing server path still preserves the private user+org boundary, and the Attention → Pinned → recency projection remains consistent with the durable Workspace contract. This round introduces no server, database, schema, or migration changes.

Local verification:

- 35 focused Workspace Pin/rail tests passed.
- Biome passed on all three changed files.
- Web typecheck and design-token guardrails passed.

Approved without waiting for the remaining CI test job.

PR #730, opened 2026-06-01
test(e2e): port direct CLI onboarding from legacy bash (Phase A of #725)

yuezengwu:
## Summary

Phase A of #725 — replaces the direct-CLI section of `e2e/onboarding-smoke.sh` with a real Vitest e2e under `packages/e2e`. Source repo is a local temp git fixture (no external clone), so the test does not inherit demo-repo flake.

Confirmed direction with @baixiaohang in [issue #725](https://github.com/agent-team-foundation/first-tree/issues/725):
- ✅ Phase A here (this PR)
- Phase B: codex/claude prompt onboarding dropped, live GitHub `validate.yml` folded into a template self-test — follow-up PRs
- Phase C: delete `e2e/*.sh` + rewrite `.github/workflows/agent-e2e.yml` — final follow-up

## What the test covers

Chain: `tree inspect → tree init → tree inspect → tree verify → tree automation install --tier 2 --dry-run → tree skill doctor`

Assertions:
- `inspect` role flips from `unbound-source-repo` → `source-repo-bound`
- `init` reports a non-empty `treeRoot`
- `verify ok === true`
- `automation install --dry-run` reports `stage === "write_rule_layer"`
- `skill doctor` runs to completion and emits valid JSON (mirrors legacy bash — does not gate exit code; Phase B will add a dedicated skill-doctor health test)
- Source repo files: `AGENTS.md`, `CLAUDE.md`
- Tree repo files: `.github/workflows/validate.yml` (first line is `# first-tree-template-version: 2`), `.first-tree/agent-templates/{developer,code-reviewer}.yaml`, `.first-tree/org.yaml`

## Files

| File | Type | Notes |
|---|---|---|
| `packages/e2e/src/framework/cli-driver/exec.ts` | mod | optional `cwd?: string` on `ExecCliOptions`; default unchanged |
| `packages/e2e/src/framework/cli-json.ts` | new | `runCliJson<T>` wraps `execCli` with JSON parse + exit-code gate |
| `packages/e2e/src/framework/local-git-fixture.ts` | new | `mkdtemp` + `git init/commit`; returns `{ repoRoot, cleanup }` |
| `packages/e2e/src/tests/onboarding-direct.e2e.test.ts` | new | the test itself |

No changes to product packages (`apps/cli`, `packages/server`, `packages/client`, `packages/web`). QA agent independently confirmed the same: only the e2e test package is touched and no product package reverse-imports `@first-tree/e2e`.

## Test plan

[...]

PR #1752, opened 2026-07-13
feat(web): mobile add-to-home-screen install guide

Gandy2025:
## What & why

Mobile web users have no nudge to install First Tree as a home-screen app. The mobile surface (`/m/*`) is **already a valid PWA** — manifest, 192/512 + apple-touch icons, and apple-mobile-web-app meta are injected when the mobile experience is enabled. What was missing is the *guidance UX* and the native-install wiring. This adds them. **No service worker is added** — verified on staging that the current manifest already meets Chrome's Android installability criteria (Chrome removed the service-worker fetch-handler requirement for menu installation in **v108**; current criteria no longer list a service worker), and iOS A2HS never needed one. Refs: https://developer.chrome.com/blog/update-install-criteria , https://web.dev/articles/install-criteria

**Paired product-scope admission (required to merge):** agent-team-foundation/first-tree-context#730 records the product-owner decision admitting home-screen install promotion into the default-deny mobile surface (`product/surfaces/mobile-surface.md`).

## Behaviour

- **Bottom-sheet guide**, value-first copy. Mirrors the existing mobile sheet idiom (team switcher): docked, scrim, safe-area padding, `radius-dialog` top corners, slide-up with a `prefers-reduced-motion` fallback. Modal focus management (move / trap / restore).
- **Platform-adaptive:**
  - iOS Safari → teaches the two steps: tap Share → "Add to Home Screen".
  - Android with a captured `beforeinstallprompt` → one-tap "Add to Home Screen".
  - Android without it → menu-steps fallback. Capture is scoped to Android; the browser's own install UI is left untouched elsewhere.
- **First-pop timing** anchored to a value moment (open a chat → return to a resting screen, +~1.8s settle), ~25s dwell fallback for passive readers. Never pops while typing, backgrounded, over another open overlay / immersive chat, on an empty feed, or in the first moments.
- **Re-nag cap:** auto-shows **at most twice** (lifetime cap keyed on *shows*, so an abandoned sheet still counts), at most once per session. After that, a permanent **"Add to Home Screen"** entry stays on the Me page (manual opens never count). Once installed (`appinstalled` or standalone) every promotion hides reactively and never shows again. State in `localStorage`.

## Files
`pages/mobile/install-guide-state.ts` (pure state / detection / eligibility) · `use-install-guide.ts` (beforeinstallprompt capture store + trigger state machine) · `install-guide-sheet.tsx` (sheet + focus management) · wired into `shell.tsx` + `me.tsx` · keyframes in `index.css` · DEV preview `/preview/install-guide`.

## Testing
- Unit + DOM + hook-lifecycle tests (27 new): platform/standalone detection, eligibility/cap matrix, each sheet variant + dismissal, Android capture + `preventDefault`, desktop no-capture, menu-driven `appinstalled` hides promotion, reverse-tab focus trap. Full web suite green (1404).
- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) and `pnpm check` green.
- **Visual E2E**: rendered the real component via the preview route in a phone viewport, both light and dark, all three variants (matches the approved design).

### Honest limitation
Headless Chromium does **not** dispatch `beforeinstallprompt`, so the Android one-tap → native install dialog can't be reproduced in automation; it needs a real Android device or Chrome DevTools → Application. Every deterministic precondition Chrome checks is satisfied (verified against the live staging manifest). iOS, the trigger/gating logic, and the capture/appinstalled lifecycle are covered by the preview + unit/hook tests.

## Design
Design doc + approved mockups tracked in the workspace (`design/mobile-a2hs/`).

### DIFF
--- a/packages/web/src/app.tsx
+++ b/packages/web/src/app.tsx
@@ -131,6 +131,12 @@ const MobilePreviewPage = import.meta.env.DEV
   ? lazy(() => import("./pages/mobile-preview.js").then((module) => ({ default: module.MobilePreviewPage })))
   : null;
 
+const InstallGuidePreviewPage = import.meta.env.DEV
+  ? lazy(() =>
+      import("./pages/install-guide-preview.js").then((module) => ({ default: module.InstallGuidePreviewPage })),
+    )
+  : null;
+
 const TeamSwitcherPreviewPage = import.meta.env.DEV
   ? lazy(() =>
       import("./pages/team-switcher-preview.js").then((module) => ({ default: module.TeamSwitcherPreviewPage })),
@@ -293,6 +299,16 @@ export function App() {
                   element={
                     <Suspense fallback={null}>
                       <MobilePreviewPage />
+                    </Suspense>
+                  }
+                />
+              ) : null}
+              {InstallGuidePreviewPage ? (
+                <Route
+                  path="/preview/install-guide"
+                  element={
+                    <Suspense fallback={null}>
+                      <InstallGuidePreviewPage />
                     </Suspense>
                   }
                 />
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -1162,6 +1162,29 @@
   }
 }
 
+/* Mobile install ("add to home screen") guide — docked bottom sheet slides up,
+   scrim fades in. Mirrors the toast/mention entrance vocabulary. */
+@keyframes mobile-sheet-in {
+  from {
+    transform: translateY(100%);
+  }
+  to {
+    transform: translateY(0);
+  }
+}
+.mobile-sheet-in {
+  animation: mobile-sheet-in 220ms cubic-bezier(0.2, 0.8, 0.2, 1);
+}
+.mobile-scrim-in {
+  animation: subtle-fade 180ms ease-out;
+}
+@media (prefers-reduced-motion: reduce) {
+  .mobile-sheet-in,
+  .mobile-scrim-in {
+    animation: none;
+  }
+}
+
 /* Left-aligned content column that fills the shared 960 canvas, matching the
    Team / Agent Detail tabs (same `--sp-2 --sp-5 --sp-7` content padding under a
    PageHeader). No min-height / centering — the page used to render as a
--- a/packages/web/src/pages/install-guide-preview.tsx
+++ b/packages/web/src/pages/install-guide-preview.tsx
@@ -1,0 +1,56 @@
+import { useState } from "react";
+import { InstallGuideSheet } from "./mobile/install-guide-sheet.js";
+import type { InstallGuideMode } from "./mobile/use-install-guide.js";
+
+// DEV-only visual harness for the install ("add to home screen") sheet. Renders
+// each platform variant over a tinted backdrop so tokens/layout can be checked
+// in light and dark without auth or a real beforeinstallprompt event.
+const MODES: ReadonlyArray<{ mode: InstallGuideMode; label: string }> = [
+  { mode: "native", label: "Android (one-tap)" },
+  { mode: "ios", label: "iOS (share steps)" },
+  { mode: "android-manual", label: "Android (menu steps)" },
+];
+
+export function InstallGuidePreviewPage() {
+  const [mode, setMode] = useState<InstallGuideMode>("ios");
+
+  return (
+    <div
+      className="min-h-dvh flex flex-col items-center"
+      style={{ background: "var(--bg)", padding: "var(--sp-6) var(--sp-4)", gap: "var(--sp-4)" }}
+    >
+      <div className="flex flex-wrap justify-center" style={{ gap: "var(--sp-2)" }}>
+        {MODES.map((option) => (
+          <button
+            key={option.mode}
+            type="button"
+            onClick={() => setMode(option.mode)}
+            className="transition-colors"
+            style={{
+              padding: "var(--sp-1_5) var(--sp-3)",
+              borderRadius: "var(--radius-input)",
+              border: "var(--hairline) solid var(--border)",
+              background: mode === option.mode ? "var(--bg-active)" : "var(--bg-raised)",
+              color: mode === option.mode ? "var(--fg)" : "var(--fg-3)",
+              cursor: "pointer",
+            }}
+          >
+            {option.label}
+          </button>
+        ))}
+      </div>
+      <p className="text-mobile-body" style={{ color: "var(--fg-3)", margin: 0 }}>
+        Set the browser to a phone viewport to see the sheet as it ships.
+      </p>
+      <InstallGuideSheet
+        mode={mode}
+        onInstall={() => {
+          /* preview no-op */
+        }}
+        onClose={() => {
+          /* preview no-op: keep the sheet visible */
+        }}
+      />
+    </div>
+  );
+}
--- a/packages/web/src/pages/mobile/__tests__/install-guide-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/install-guide-dom.test.tsx
@@ -1,0 +1,74 @@
+// @vitest-environment happy-dom
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { createDomHarness, type DomHarness } from "../../../test-utils/dom-harness.js";
+import { InstallGuideSheet } from "../install-guide-sheet.js";
+
+let harness: DomHarness;
+
+beforeEach(() => {
+  harness = createDomHarness();
+});
+
+afterEach(() => {
+  harness.cleanup();
+});
+
+function findButton(label: string): HTMLButtonElement | null {
+  return (
+    [...harness.container.querySelectorAll("button")].find((button) => button.textContent?.includes(label)) ?? null
+  );
+}
+
+describe("InstallGuideSheet", () => {
+  it("shows the value proposition and native one-tap CTA", () => {
+    const onInstall = vi.fn();
+    const onClose = vi.fn();
+    harness.render(<InstallGuideSheet mode="native" onInstall={onInstall} onClose={onClose} />);
+
+    const sheet = harness.container.querySelector('[data-mobile-install-sheet="true"]');
+    expect(sheet?.textContent).toContain("Add First Tree to your home screen");
+    expect(sheet?.textContent).toContain("Opens instantly");
+    // Native mode is one-tap, not a step list.
+    expect(sheet?.textContent).not.toContain("Share button");
+
+    findButton("Add to Home Screen")?.click();
+    expect(onInstall).toHaveBeenCalledTimes(1);
+    findButton("Maybe later")?.click();
+    expect(onClose).toHaveBeenCalledTimes(1);
+  });
+
+  it("teaches the iOS share steps and never calls install", () => {
+    const onInstall = vi.fn();
+    const onClose = vi.fn();
+    harness.render(<InstallGuideSheet mode="ios" onInstall={onInstall} onClose={onClose} />);
+
+    const sheet = harness.container.querySelector('[data-mobile-install-sheet="true"]');
+    expect(sheet?.textContent).toContain("Share button below");
+    expect(sheet?.textContent).toContain('Add to Home Screen"');
+    expect(harness.container.querySelector("ol")).not.toBeNull();
+
+    findButton("Got it")?.click();
+    expect(onClose).toHaveBeenCalledTimes(1);
+    expect(onInstall).not.toHaveBeenCalled();
+  });
+
+  it("teaches the Android menu steps in the manual fallback", () => {
+    harness.render(<InstallGuideSheet mode="android-manual" onInstall={vi.fn()} onClose={vi.fn()} />);
+    const sheet = harness.container.querySelector('[data-mobile-install-sheet="true"]');
+    expect(sheet?.textContent).toContain("browser menu");
+    expect(sheet?.textContent).toContain("Install app");
+  });
+
+  it("dismisses on scrim click and Escape", () => {
+    const onClose = vi.fn();
+    harness.render(<InstallGuideSheet mode="ios" onInstall={vi.fn()} onClose={onClose} />);
+
+    const scrim = harness.container.querySelector<HTMLButtonElement>('button[aria-label="Dismiss"]');
+    scrim?.click();
+    expect(onClose).toHaveBeenCalledTimes(1);
+
+    window.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape" }));
+    expect(onClose).toHaveBeenCalledTimes(2);
+  });
+});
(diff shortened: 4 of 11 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1752

- Rationale: the install lifecycle is not propagated to every promotion surface, the global event capture suppresses browser install UI outside the mobile flow, and the hand-rolled modal does not fully trap focus.

**Risk level**: C
- Path baseline: all changes are under `packages/web/**` -> C
- Semantic lift: none; this does not change Web subscriptions, WebSocket framing, or the data layer.

**PR summary**
- Problem: Mobile users can already launch the staged Web surface as a PWA, but there is no discoverable guidance or one-tap Android install path.
- Approach: Add a platform-adaptive bottom sheet, trigger it after a useful mobile interaction or dwell fallback, cap automatic promotion per device/session, and retain a manual entry under Me.
- Impacted modules: mobile shell and Me page, install state/event hooks, bottom-sheet UI, DEV preview, and Web styles/tests.

**Review findings**
❌ 1. An `appinstalled` transition is persisted but is not part of the external-store snapshot, so installs completed from Chrome's menu/omnibox (the exact Android manual-fallback path) do not invalidate consumers. When `deferredPrompt` is already `null`, lines 42-45 emit the same `null` snapshot; `MobileInstallEntry` also memoizes `isStandalone()` once and never reads the persisted `installed` flag. As a result, an open manual guide and the Me-page promotion remain visible in the current browser tab after installation, contrary to the stated "never shows again" behavior. Make installed status reactive and close/hide every promotion on that transition; cover the browser-menu install path. Official guidance likewise uses `appinstalled` to disable in-app promotion. [R1/R5 / `packages/web/src/pages/mobile/use-install-guide.ts:42`, `packages/web/src/pages/mobile/me.tsx:318`] ([MDN](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Trigger_install_prompt#responding_to_app_install))

❌ 2. The top-level app statically imports this module, so the unconditional `event.preventDefault()` runs on every route/platform where `beforeinstallprompt` fires. That suppresses browser-provided install UI on desktop dev/staging and other non-mobile routes, while the replacement UI only renders under `/m/*` and auto-eligibility rejects desktop UAs. Scope suppression/capture to the supported mobile flow/platform, or preserve an equivalent promotion for every affected surface. [R5 / `packages/web/src/pages/mobile/use-install-guide.ts:36`] ([MDN](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Trigger_install_prompt#listening_for_beforeinstallprompt))

❌ 3. The dialog root receives initial focus with `tabIndex={-1}`, but it is excluded from the queried focusables. Pressing Shift+Tab immediately after the auto-pop therefore matches neither wrap branch and follows the DOM backward to the focusable scrim outside the dialog, so the claimed modal focus trap is incomplete. Focus an in-dialog control initially or explicitly handle focus when the active element is outside the focusable set, and add a Shift+Tab regression case. [`packages/web/src/pages/mobile/install-guide-sheet.tsx:35`, `packages/web/src/pages/mobile/install-guide-sheet.tsx:43`]

❓ 4. The active `product/surfaces/mobile-surface.md` node uses a default-deny admission gate, but it does not explicitly record automatic/native install promotion. The archived mobile proposal supports the lightweight PWA shell but is non-canonical, and the PR's referenced approved design is not linked here. Please link the recorded product-owner admission decision or the paired active-tree update so this scope addition is auditable.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Request changes.

## Goal and core change

This adds the missing install-promotion layer to the existing staged mobile PWA: a platform-adaptive bottom sheet, a `beforeinstallprompt` capture/install path, value-moment and dwell auto-triggering with per-device/session caps, plus a permanent Me-page entry. The persistence change is browser-local (`localStorage` / `sessionStorage`) only; there is no database, schema, shared DTO, or server-side core-data-structure change requiring a migration review.

## Blockers

I independently verified the three implementation blockers already documented in @baixiaohang's review on commit `8eaab281`; I am not duplicating inline comments:

1. `appinstalled` persists state but is not part of the external-store snapshot. In the manual/menu path the snapshot remains `null -> null`, so the open guide and Me promotion do not reactively close/hide after installation.
2. The eagerly imported module unconditionally calls `preventDefault()` for every `beforeinstallprompt`, including routes/platforms where the replacement mobile UI cannot render. That suppresses browser-provided promotion outside the supported flow.
3. Initial focus lands on the `tabIndex={-1}` dialog root, but the root is excluded from the focusable list. The first `Shift+Tab` can therefore move backward to the focusable scrim outside the dialog, so the advertised modal trap is incomplete.

Please fix these behaviors and add regression coverage for the menu-driven `appinstalled` transition, capture scoping, and initial reverse-tab traversal. The current 23 new tests cover pure eligibility and sheet rendering, but do not exercise the event/trigger hook lifecycle where the first two regressions live.

There is also a product-scope gate before merge. The authoritative `product/surfaces/mobile-surface.md` node makes mobile capability admission default-deny and says admission must be explicitly recorded there rather than inferred from a delivery PR. Automatic/native install promotion is not currently admitted, and I found no linked or open matching `first-tree-context` PR. Please link the existing recorded owner decision if one exists, or pair this with the active-tree update that admits the capability.

## Accuracy note

Please also correct the PR description's Chrome-version claim. Chrome's published change removed the service-worker fetch-handler requirement for menu installation in mobile version 108, not 89; the current installability criteria no longer list a service worker. This does not invalidate the present manifest-only approach, but the rationale should be accurate: https://developer.chrome.com/blog/update-install-criteria and https://web.dev/articles/install-criteria.

## Validation

- `pnpm --filter @first-tree/web test` — 175 files / 1400 tests passed
- focused mobile install/shell/Me run — 30 tests passed
- `git diff --check origin/main...HEAD` — clean

Gandy2025:
Thanks both — all four addressed. Fixes in `df196c07`.

**1. `appinstalled` not reactive (menu/omnibox install path).** Fixed. `installed` is now part of the `beforeinstallprompt` external-store snapshot (`use-install-guide.ts`), so `useInstallPrompt` derives `mode = null` on install even when `prompt` was already `null` (the Android manual path). That reactively hides the Me-page entry and closes any open guide in the current tab. New test: *"hides promotion reactively after a menu-driven appinstalled"*.

**2. Unconditional `preventDefault()` suppressing browser install UI off-flow.** Fixed. The capture listener now no-ops unless `detectPlatform() === "android"` (the only surface where we render a replacement), leaving the browser's own install UI untouched on desktop/other routes. New test: *"ignores beforeinstallprompt off the supported mobile flow (no preventDefault)"*.

**3. Incomplete focus trap (initial reverse-tab escapes to scrim).** Fixed. The Tab handler now detects focus sitting on the `tabIndex=-1` dialog root (or otherwise outside the focusable set) and pulls it back to the first/last control, so Shift+Tab can't reach the scrim. New test: *"traps focus inside the sheet on Shift+Tab from the dialog root"*.

Regression coverage for the event/trigger hook lifecycle now lives in `__tests__/install-guide-hook.test.tsx` (+ the reverse-tab case in `install-guide-dom.test.tsx`) — 27 install-guide tests total, full web suite green (1404).

**4. Product-scope admission gate.** Paired active-tree PR opened: **agent-team-foundation/first-tree-context#730** records the product-owner (@Gandy2025, the node owner) decision admitting home-screen install promotion into the default-deny `product/surfaces/mobile-surface.md`, with the boundary (self-promotion affordance, non-nagging, no push/notification promotion). Cross-linked in the PR description.

**Accuracy note.** PR description corrected: Chrome removed the service-worker fetch-handler requirement for menu installation in **v108** (not 89); current installability criteria no longer list a SW. Thanks for the catch.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1752 (follow-up on `df196c07`)

**Recommendation**: approve
- Rationale: all three implementation blockers from `8eaab281` are resolved in the correct state/event boundaries, with focused regression coverage; the remaining product-scope dependency is explicitly tracked by paired Context Tree PR #730.

**Risk level**: C
- Path baseline: all code changes remain under `packages/web/**` -> C
- Semantic lift: none; no Web subscription, WebSocket frame, or data-layer behavior changed.

**PR summary**
- Problem: Mobile users need a discoverable, bounded way to install the staged mobile surface as a home-screen app.
- Approach: Provide platform-adaptive install guidance, Android native-prompt capture, value-moment/dwell triggers with hard re-nag caps, and a permanent manual entry under Me.
- Impacted modules: mobile install state/event hook, install sheet, mobile shell/Me wiring, DEV preview, styles, and focused tests.

**Review findings**
✅ 1. `appinstalled` now updates a stable external-store snapshot that includes `installed`, so the menu/omnibox path invalidates consumers even when the prompt remains `null`; both the open guide and Me entry hide reactively. [`packages/web/src/pages/mobile/use-install-guide.ts:27`]

✅ 2. `beforeinstallprompt` interception is restricted to the target Android platform, preserving browser-provided promotion on desktop/other platforms while retaining the intentionally early pre-shell capture for the mobile flow. [`packages/web/src/pages/mobile/use-install-guide.ts:52`]

✅ 3. The modal trap now handles initial focus on the dialog root and any focus outside the in-dialog set, wrapping reverse/forward traversal back into the sheet. [`packages/web/src/pages/mobile/install-guide-sheet.tsx:43`]

✅ 4. The PR description now states the Chrome v108 service-worker criterion accurately and links paired product-scope admission PR [first-tree-context#730](https://github.com/agent-team-foundation/first-tree-context/pull/730).

⚠️ 5. Merge dependency: #730 is still OPEN. Do not merge this code PR until that active-tree admission lands; the current `product/surfaces/mobile-surface.md` on `main` remains default-deny without this capability.

**Action taken**
- Approved the code PR. No tests or code checks were run locally as part of this review.

yuezengwu (approved):
Approved on `df196c07`.

## Follow-up result

The change still has the same intended shape: a bounded install-promotion layer for the staged mobile PWA, with browser-local state only and no database/schema/shared-DTO/core server data change. The previous implementation blockers are resolved:

1. `appinstalled` is now part of a stable external-store snapshot. A menu/omnibox install publishes a real snapshot transition even when `prompt` was already `null`, and `mode = null` reactively removes both the open sheet and Me entry.
2. `beforeinstallprompt` interception is now Android-only, preserving browser promotion on desktop/other platforms while retaining the explicitly chosen pre-shell capture for the target platform.
3. The Tab handler now pulls initial/outside focus into the in-dialog control set and wraps both directions correctly.
4. The PR description has the corrected Chrome v108 wording, and paired tree PR [first-tree-context#730](https://github.com/agent-team-foundation/first-tree-context/pull/730) makes the required durable admission in the authoritative mobile-surface node. Its diff is appropriately limited to the self-promotion boundary, non-nagging constraint, and exclusion of general notification/push promotion.

Two non-blocking test-precision notes:

- `ignores beforeinstallprompt off the supported mobile flow` actually verifies an off-platform desktop UA, not an Android off-route flow. Renaming it to “off the supported platform” would match the implemented platform-wide early-capture contract.
- The reverse-tab test would also pass if focus simply remained on the dialog root, because `dialog.contains(dialog)` is true. Assert that the expected last in-dialog control receives focus so the test genuinely guards the new wrap branch.

## Context Tree merge choreography

The paired tree content resolves the product-scope gate. Per the workspace tree workflow, however, #730 should be a **draft** while the code PR is open: merge this code PR first, reconcile the tree diff against the final merged code, then mark #730 ready. #730 is currently non-draft, so please switch it back to draft rather than merging the tree PR first.

## Validation

- focused install/shell/Me tests — 5 files / 34 tests passed
- full Web suite — 176 files / 1404 tests passed
- `pnpm check` — exited successfully (only existing unrelated diagnostics)
- incremental `git diff --check` — clean

baixiaohang:
Correction to finding 5 in my approval review: I had the paired-PR merge order backwards.

The active team workflow in `team-practice/tree-maintenance.md` requires the paired Context Tree PR to remain **draft** while the code PR is open. The correct sequence is:

1. keep [first-tree-context#730](https://github.com/agent-team-foundation/first-tree-context/pull/730) draft and cross-linked for code review;
2. merge this code PR first;
3. reconcile #730 against the final merged code;
4. mark #730 ready so its validation/review/auto-merge runs against the shipped result.

Therefore #730 does **not** need to merge before this code PR. Please switch #730 back to draft. This supersedes only the merge-dependency wording in my prior review; my approval of `df196c07` is unchanged.

Gandy2025:
Both non-blocking test-precision notes applied in `00ab5130`:
- The reverse-tab test now asserts focus wraps to the **last in-dialog control ("Got it")**, so it genuinely guards the new wrap branch (no longer trivially green via `dialog.contains(dialog)`).
- Renamed the desktop-UA test to *"off the supported platform"* to match the platform-wide early-capture contract.

Merge choreography per `team-practice/tree-maintenance.md`: **first-tree-context#730 is now draft**. I'll merge this code PR first, then reconcile #730 against the merged result and mark it ready so its validation runs against the shipped code. Thanks both.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1752 (follow-up on `00ab5130`)

**Recommendation**: approve
- Rationale: the new commit only tightens the two non-blocking regression tests and introduces no production-code change.

**Risk level**: C
- Path baseline: changes remain under `packages/web/**` -> C
- Semantic lift: none.

**PR summary**
- Problem: The previous regression tests were directionally correct but one assertion could pass without exercising the intended focus-wrap branch, and one test name overstated route scoping.
- Approach: Assert that reverse traversal lands on the final in-dialog action and rename the desktop-UA case to describe the platform-wide capture contract.
- Impacted modules: install-guide DOM and hook tests only.

**Review findings**
✅ 1. The reverse-tab case now verifies the expected `Got it` button receives focus, so it guards the actual wrap branch rather than merely containment.

✅ 2. The capture test name now matches its desktop/off-platform setup; the intentionally early Android capture contract remains unchanged.

✅ 3. Paired Context Tree PR #730 is now draft, matching the active code-first merge choreography.

**Action taken**
- Approved `00ab5130`. No tests or code checks were run locally as part of this review.

yuezengwu (approved):
Approved on `00ab5130`.

This test-only follow-up correctly applies both non-blocking review notes:

- The reverse-tab regression now asserts focus reaches the final in-dialog control (`Got it`), so the pre-fix behavior that leaves focus on the dialog root can no longer pass.
- The desktop-UA capture case is now named “off the supported platform,” matching the intentionally platform-wide Android capture contract.

There is no production-code, database/schema, or core data-structure change in this commit. The install-guide implementation goal and behavior reviewed on `df196c07` are unchanged, and the prior lifecycle, capture-scope, and focus-trap blockers remain resolved. Paired Context Tree PR #730 is now draft, matching the code-first → reconcile → ready workflow.

Local validation: full Web suite passed (176 files / 1,404 tests), including both refined cases; `git diff --check df196c07..00ab5130` is clean.

PR #1755, opened 2026-07-14
feat(web): compact grouped list for admin Team computers

Gandy2025:
## What

Redesigns the admin **Team computers** audit view on Settings → Computers from a dense 7-column table into a **compact, health-grouped list**.

- One line per machine: **hostname** (mono, the focus) over an `owner · OS · version` meta line; right-aligned agent count and status pill.
- **Health grouping**: machines whose status is *not* `ready` (auth-expired / setup-incomplete / offline) sort under a **`Needs attention`** header at the top; the ready fleet follows under `Ready`. A fully-healthy fleet renders as a **flat list with no group chrome**.
- **Less noise**: the version drops its `-staging.NN` build suffix (`0.5.14-staging.841.1` → `0.5.14`, full string on hover); "last seen" folds into the **Offline** status (`Offline · 3 days`) instead of a redundant column beside a live `Ready` pill. Long hostnames truncate with an ellipsis, so the row stops wrapping.

The section stays **collapsed by default** (unchanged) and carries no attention badge — the health signal lives inside, surfaced on expand.

## Why

The old table wrapped versions/owners/timestamps onto two lines and read like a back-office spreadsheet. The audit view's real job is spotting *unhealthy* machines, so problems now sort to the top. The row vocabulary mirrors the existing "Your computers" card meta line for cross-surface consistency.

## Removed (dead code)

The old `DenseTable` path — `ClientRow`, `TeamComputersTable`, `CapabilityMatrix`, `ProviderRow`, and the `collapsedIds` state — is deleted. Team rows were always rendered `restricted`, which already force-disabled per-row expansion and the owner-only capability probe, so that branch never rendered in production. Net ‑141 lines.

## Notes for reviewers

- The one meaningful color is the amber `Needs attention` header (`--fg-needs-you-strong`); rows carry each machine's own status-pill hue. An earlier iteration tinted the whole attention group amber — dropped after review as a token-semantics stretch (amber wash behind red/gray pills) and to stay restrained.
- `dev-fixtures.ts`: the DEV-only `admin-grouped` demo scenario gains varied-status team machines so `/settings/computers?demo=admin-grouped` previews the grouped list.

## Verification

- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens` design-token gate) — clean.
- `pnpm biome check` on changed files — clean.
- `pnpm --filter @first-tree/web test` — **1416 passed** (incl. the clients-page DOM test).
- Real browser (Chrome, dev server) — verified the grouped list in **light and dark**, expand/collapse, ellipsis truncation, and the `Offline · 3 days` fold via the `admin-grouped` demo.

### DIFF
--- a/packages/web/src/api/activity.ts
+++ b/packages/web/src/api/activity.ts
@@ -4,6 +4,7 @@ import type {
   ConnectTokenResponse,
   RuntimeAuthMethod,
   RuntimeProvider,
+  UpdateAttempt,
 } from "@first-tree/shared";
 import { api, withOrg } from "./client.js";
 
@@ -61,6 +62,15 @@ export type HubClient = {
    * (≥1 capability `state=ok`) from Setup incomplete (all `≠ ok`).
    */
   capabilities: ClientCapabilities;
+  /**
+   * Outcome of the client's last self-update attempt, persisted by the
+   * server into `clients.metadata.lastUpdateAttempt` and returned by
+   * `/me/clients` and `/orgs/:orgId/clients`. A `failed` / `blocked`
+   * result means the machine is stuck on an old version and needs admin
+   * attention even while it's otherwise connected — see
+   * `teamNeedsAttention` in `pages/clients/derive-status.ts`.
+   */
+  lastUpdateAttempt?: UpdateAttempt | null;
 };
 
 export function retireClient(clientId: string): Promise<void> {
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -1840,6 +1840,20 @@
   border-top: var(--hairline) solid var(--border-faint);
 }
 
+/* Team computers audit list (admin) — compact grouped rows with a hairline
+   between every row/group-header and a neutral hover. The health split is
+   carried by row order + the "Needs attention" group header + each row's
+   status pill, not a full-row color wash. */
+.team-computers-list > * + * {
+  border-top: var(--hairline) solid var(--border-faint);
+}
+.team-computer-row {
+  transition: background-color 120ms ease-out;
+}
+.team-computer-row:hover {
+  background-color: var(--bg-hover);
+}
+
 /* Dense table row hover — used by DenseTableRow[data-interactive] */
 tr[data-interactive]:hover {
   background: var(--bg-hover);
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -1844,14 +1844,34 @@
    between every row/group-header and a neutral hover. The health split is
    carried by row order + the "Needs attention" group header + each row's
    status pill, not a full-row color wash. */
+.team-computers-list {
+  /* Query the list's own width (not the viewport): the Settings nav eats
+     width, so a row can be narrow even on a wide screen. */
+  container-type: inline-size;
+}
 .team-computers-list > * + * {
   border-top: var(--hairline) solid var(--border-faint);
 }
 .team-computer-row {
+  display: grid;
+  /* identity | agents | status. Fixed trailing tracks keep the agent counts
+     and status pills column-aligned across rows when there's room. */
+  grid-template-columns: minmax(0, 1fr) var(--sp-20) var(--sp-45);
+  align-items: center;
+  column-gap: var(--sp-3);
+  padding: var(--sp-2_5) var(--sp-3_5);
   transition: background-color 120ms ease-out;
 }
 .team-computer-row:hover {
   background-color: var(--bg-hover);
+}
+/* Narrow list (mobile / cramped Settings): drop the fixed trailing tracks to
+   intrinsic width so the hostname/meta identity column never collapses. */
+@container (max-width: 34rem) {
+  .team-computer-row {
+    grid-template-columns: minmax(0, 1fr) auto auto;
+    column-gap: var(--sp-2_5);
+  }
 }
 
 /* Dense table row hover — used by DenseTableRow[data-interactive] */
--- a/packages/web/src/pages/clients.tsx
+++ b/packages/web/src/pages/clients.tsx
@@ -1,4 +1,3 @@
-import type { CapabilityEntry, ClientCapabilities, RuntimeProvider } from "@first-tree/shared";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import { ChevronDown, ChevronRight, Plus } from "lucide-react";
 import { useMemo, useState } from "react";
@@ -16,14 +15,6 @@ import { listMembers } from "../api/members.js";
 import { useAuth } from "../auth/auth-context.js";
 import { Button } from "../components/ui/button.js";
 import {
-  DenseTable,
-  DenseTableBody,
-  DenseTableCell,
-  DenseTableHead,
-  DenseTableHeader,
-  DenseTableRow,
-} from "../components/ui/dense-table.js";
-import {
   Dialog,
   DialogContent,
   DialogDescription,
@@ -32,13 +23,11 @@ import {
   DialogTitle,
 } from "../components/ui/dialog.js";
 import { PresenceChip, runtimeStateToPresence } from "../components/ui/presence-chip.js";
-import { RowActionsMenu } from "../components/ui/row-actions-menu.js";
 import { Section } from "../components/ui/section.js";
 import { UppercaseLabel } from "../components/ui/section-header.js";
 import { useAgentNameMap } from "../lib/use-agent-name-map.js";
-import { formatDate, formatRelative } from "../lib/utils.js";
 import { ComputerCard } from "./clients/cards/computer-card.js";
-import { PROVIDER_LABEL, PROVIDER_ORDER, providerInstallHint } from "./clients/cards/shared/providers.js";
+import { formatOfflineDuration } from "./clients/cards/view-models.js";
 import { ComputerStatusPill } from "./clients/computer-status-pill.js";
 import { DemoNavigator, useDemoScenarioParam } from "./clients/demo-navigator.js";
 import { compareByPillPriority, deriveComputerStatus } from "./clients/derive-status.js";
@@ -87,11 +76,6 @@ export function ClientsPage({ embedded = false }: { embedded?: boolean } = {}) {
         return DEMO_AGENT_NAMES[uuid] ?? realAgentName(uuid);
       }
     : realAgentName;
-  // Personal scope: a user typically has 1-3 computers, so expand by default
-  // and only let them collapse rows they actively want to hide. New clients
-  // arriving via the 10s poll auto-expand too — we only treat the closed set
-  // as "rows the user explicitly closed".
-  const [collapsedIds, setCollapsedIds] = useState<Set<string>>(() => new Set());
   // Admin's Team computers section starts collapsed so it doesn't push the
   // viewer's own machines below the fold. Visible count badge on the
   // header still tells admins "you have N teammates" — they click to
@@ -577,20 +561,7 @@ export function ClientsPage({ embedded = false }: { embedded?: boolean } = {}) {
                   {teamList.length === 0 ? (
                     <EmptyCardsNote message="No other team computers." />
                   ) : teamExpanded ? (
-                    // Team rows stay as table for now — read-only audit
-                    // view. Per-row inline affordances (Generate new
-                    // token / install guide) don't apply to teammates;
-                    // the full team-card redesign with "Copy suggestion
-                    // → Alice" buttons is deferred to a follow-up PR
-                    // (proposal §"Variant D").
-                    <TeamComputersTable
-                      teamList={teamList}
-                      collapsedIds={collapsedIds}
-                      setCollapsedIds={setCollapsedIds}
-                      agentName={agentName}
-                      getClientAgents={getClientAgents}
-                      resolveOwner={resolveOwner}
-                    />
+                    <TeamComputersList teamList={teamList} resolveOwner={resolveOwner} />
                   ) : null}
                 </Section>
               </>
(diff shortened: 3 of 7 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1755

- Rationale: The new health grouping can label a known failed-to-update computer as `Ready`, and the fixed row grid makes machine identity unreadable on the supported narrow Settings layout.

**Risk level**: C
- Path baseline: only `packages/web/**` -> C
- Semantic lift: none; this changes presentation and client-side derivation, not subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Admins need to scan teammates' computers for machines that require intervention without reading a dense audit table.
- Approach: Replace the table with a compact read-only list, sort non-ready runtime states into a leading `Needs attention` group, and reduce version/last-seen noise.
- Impacted modules: Settings → Computers list UI, shared Web styling, the admin demo fixture, and its DOM smoke test.

**Review findings**

❌ 1. The health grouping drops the update-failure signal that the admin endpoint already exposes. [`packages/web/src/pages/clients.tsx:647`]

`/orgs/:orgId/clients` returns `lastUpdateAttempt` specifically so the admin dashboard can surface clients whose self-update `result` is `failed` or `blocked`, but `HubClient` does not type the field and this grouping considers only the four-state runtime pill. A connected client with an OK capability and a failed/blocked update therefore lands under `Ready`, with the failure and reason invisible. Once this PR makes `Needs attention`/`Ready` the audit taxonomy, that is a false health classification. Please carry the server signal into the Web model and give failed/blocked attempts an explicit, coherent treatment in the grouping/row (or narrow the labels so they do not claim complete machine health). [R5]

❌ 2. The fixed grid collapses the hostname/owner column on the supported narrow Settings layout. [`packages/web/src/pages/clients.tsx:712`]

The two fixed tracks consume 260px, and gaps plus row padding bring the minimum to 312px before the `minmax(0, 1fr)` identity track gets any width. `/settings` explicitly supports viewports below 48rem; after the page's horizontal padding, a 320px viewport gives this row about 308px and the first track collapses to zero (at 375px it is only about 51px). That hides the primary hostname and its owner/OS/version metadata—the information the redesign says should be the focus. Please add a narrow responsive layout or use intrinsic/stacked trailing columns so machine identity remains readable. [R5]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Requesting changes for two blocking correctness/responsiveness issues.

The implementation goal is sound: it replaces the read-only seven-column Team computers audit table with a compact hostname-first list, groups non-ready runtime states above the ready fleet, folds offline age into status, abbreviates staging versions, and deletes the unreachable restricted-row expansion/capability path. Personal computer cards and owner-only actions remain unchanged.

1. **The audit grouping can classify a known failed/blocked self-update as `Ready`.**

   `GET /orgs/:orgId/clients` already returns `lastUpdateAttempt`, and the shared schema defines `failed` / `blocked` outcomes specifically for admin visibility. `HubClient` currently drops that field, while `TeamComputersList` partitions solely on `deriveComputerStatus(...).pill`. A connected client with an OK runtime capability and `lastUpdateAttempt.result === "failed"` or `"blocked"` therefore appears in `Ready` with the failure and reason hidden. That conflicts with the release contract: failed/blocked update attempts require operator or admin attention.

   Please carry the existing response signal into the Web model and give failed/blocked attempts an explicit treatment in the group and row, or narrow the taxonomy so `Ready` / `Needs attention` does not claim whole-machine health.

2. **The fixed grid removes the hostname/meta column on supported narrow Settings widths.**

   The two fixed trailing tracks use 80px + 180px; two 12px gaps and 28px row padding make the row require 312px before the `minmax(0, 1fr)` identity track receives any width. At a 320px viewport, the Computers page/list geometry leaves about 308px, so the identity track collapses to zero. At 375px it receives only about 51px. `/settings` explicitly supports the `<48rem` narrow layout, and this hides the hostname/owner/OS/version that the redesign intends to make primary.

   Please add a narrow responsive/stacked row layout or otherwise make the trailing columns intrinsic so machine identity remains readable.

The changed DOM test currently only renames “team table” to “team list”; when addressing these, please add regression coverage for the attention partition/order and its new update-failure case. A structural narrow-layout assertion would also help prevent the fixed tracks from returning.

Verification on commit `e3eb9a116f431d1f859eb965ed98d29a959b91f5`:
- `pnpm --filter @first-tree/web typecheck` — passed (including design-token checks)
- 70 focused Web tests — passed
- Biome on all changed files — clean apart from one unrelated pre-existing CSS warning

No database/schema or core persisted-data changes are present in this diff.

Gandy2025:
Thanks both — both findings addressed in `b8d06a52`.

## 1. Stuck self-update no longer hides under `Ready` ✅

Fixed. `lastUpdateAttempt` is now carried into the Web `HubClient` model, and the grouping is no longer pill-only:

- New `hasUpdateProblem` / `teamNeedsAttention` / `partitionTeamComputers` in `derive-status.ts`: a connected + Ready-runtime machine whose last self-update is `failed`/`blocked` sorts into **Needs attention**.
- The row shows **`Update failed`** / **`Update blocked`** (with the reason in the `title`) in place of the `Ready` pill, so the state and reason are visible.
- Ordering: it sorts *after* the real pill problems (auth-expired / offline) since its pill is still `ready` — a dead machine is more urgent than a stale one.

One clarification on "drops the signal": the Web `HubClient` type never carried `lastUpdateAttempt`, and the old DenseTable used the same `deriveComputerStatus`, so nothing in the Web computers view surfaced it before (only `serverCommandVersion` → "Update available"). So it was an *unwired* backend signal rather than a regression — but you're right that the new taxonomy sharpens the `Ready` claim, so it's now wired and surfaced.

Scope note: the **"Your computers" cards** still don't show update-failure (same pre-existing gap, shared `deriveComputerStatus`). I kept this PR to the audit list; happy to extend it page-wide in a focused follow-up if you'd prefer.

## 2. Narrow-layout identity collapse ✅

Confirmed (measured the identity track at ~51px at 375px). The row layout moved from inline fixed tracks to CSS with a **container query** — it queries the *list's own width* (not the viewport), since the Settings nav eats width, so a row can be narrow on a wide screen. Below `34rem` the trailing tracks drop to intrinsic (`auto`), so the hostname/meta column keeps priority; identity is now ~170px at 375px (verified in a real browser).

## Tests

- `derive-status.test.ts`: unit coverage for `hasUpdateProblem`, `teamNeedsAttention`, and `partitionTeamComputers` (the attention partition + priority order, including the update-stuck case).
- `clients-page-cards-dom.test.tsx`: a Ready-runtime + failed-update team machine now asserts it renders under **Needs attention** as **Update failed**.
- Narrow layout is verified in a real browser (jsdom/happy-dom can't measure layout, so no brittle pixel assertion there).
- Full web suite **1425 pass**; `typecheck` (+ `lint:tokens`) and Biome clean.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1755 (follow-up on `b8d06a52`)

- Rationale: The new update-health predicate treats a historical failed/blocked attempt as permanently active, so a machine remains in `Needs attention` even after the documented manual recovery succeeds.

**Risk level**: C
- Path baseline: only `packages/web/**` -> C
- Semantic lift: none; the follow-up adds a Web response field, pure client-side health derivation, rendering, and responsive CSS.

**PR summary**
- Problem: Admins need a compact Team computers audit view that puts machines requiring intervention first while keeping machine identity readable at narrow Settings widths.
- Approach: Replace the dense table with a health-partitioned list, include failed/blocked self-updates in the attention predicate, render the update outcome instead of a false `Ready` pill, and switch the trailing columns to intrinsic tracks through a container query when the list is narrow.
- Impacted modules: Web client DTO, computer health derivation, Team computers rendering/styles, demo fixtures, and Web tests.

**Review findings**

❌ 1. `hasUpdateProblem` never recognizes that a failed/blocked attempt has already been recovered. [`packages/web/src/pages/clients/derive-status.ts:100`]

The persisted value is the **last attempt**, not an independently cleared current-state flag. The blocked-update path explicitly tells the operator to run `first-tree upgrade` (or a manual npm install) and restart, while the manual `upgrade` command installs/restarts without writing an `ok` attempt or clearing `update-state.json`. After that successful recovery the client registers with, for example, `sdkVersion = 0.6.0` and the old `{ result: "blocked", target: "0.6.0" }`; this predicate still returns `true`, so the row remains `Update blocked` under `Needs attention` indefinitely (until a later automatic update happens to overwrite the record).

Please make the attention predicate distinguish an unresolved attempt from a stale historical one—for example, by proving the reported current version is still behind the failed target with semver-safe handling—or clear/overwrite the record in every successful recovery path. Add the resolved case (`sdkVersion >= attempt.target`) to the pure-function coverage. [R4, R5]

✅ 2. The original narrow-layout blocker is closed: the list is now its own inline-size query container, and sub-34rem rows use intrinsic trailing tracks instead of the fixed 80px/180px tracks. [`packages/web/src/index.css:1847`]

✅ 3. For an actually unresolved Ready-runtime update failure, the response field is now typed, the machine partitions into `Needs attention`, and the row replaces the misleading `Ready` pill with an accessible update-failure status. [`packages/web/src/pages/clients.tsx:700`]

**Action taken**
- Submitted follow-up request changes.

yuezengwu (approved):
Re-reviewed commit `b8d06a529c710ab091e685f977a6d02b6b32228a`; both requested blockers are resolved.

- The Web `HubClient` DTO now carries the server's existing `lastUpdateAttempt`. `partitionTeamComputers` treats `failed` / `blocked` attempts as attention-worthy, while the row replaces only an otherwise-`Ready` pill with `Update failed` / `Update blocked` and preserves the more urgent runtime-state precedence. The failure reason remains available in the title, and the new pure + DOM tests cover the partition and stuck-update rendering.
- Row sizing now uses the Team list as an inline-size container. Below `34rem`, intrinsic trailing tracks replace the fixed 80px/180px tracks, preserving usable hostname/meta width in the supported narrow Settings layout without sacrificing desktop column alignment.

The personal “Your computers” cards still not displaying the failure reason is a pre-existing, page-wide enhancement outside this Team-audit redesign; it does not block this PR.

Verification:
- `pnpm --filter @first-tree/web typecheck` — passed, including design-token checks
- 79 focused Web tests — passed
- Biome on the seven changed files — clean apart from one unrelated pre-existing CSS warning

No database/schema or persisted-data change is introduced. The only data-shape change is a Web DTO extension for a field the server already returns.

yuezengwu (changes requested):
Superseding my earlier approval after validating a concurrent recovery-semantics finding: one blocker remains on `b8d06a52`.

`hasUpdateProblem` treats the most recent failed/blocked attempt as an active condition forever, but `lastUpdateAttempt` is persisted attempt history rather than a cleared current-state flag. The documented blocked-update recovery is a manual `first-tree upgrade` (or manual install) followed by restart. `commands/upgrade.ts` installs and restarts successfully without calling `recordUpdateAttempt({ result: "ok", ... })` or clearing `update-state.json`. The restarted client can therefore report:

- `sdkVersion = 0.6.0`
- stale `lastUpdateAttempt = { result: "blocked", target: "0.6.0", ... }`

The current predicate still returns `true`, so a successfully recovered machine remains `Update blocked` under `Needs attention` until some later automatic attempt overwrites the record. This is especially persistent for `blocked`, because the loop guard intentionally refuses to auto-retry that target.

Please distinguish unresolved failures from stale historical records—for example, treat a failed/blocked attempt as resolved once the reported `sdkVersion` is semver-valid and at or beyond `attempt.target` (with channel prereleases handled correctly, and invalid values failing safely)—or make every supported successful recovery path overwrite/clear the record. Add pure coverage for `sdkVersion >= target` plus the unresolved/invalid cases.

The narrow container-query fix remains accepted, and the unresolved Ready-runtime failure rendering is otherwise correct. No DB/schema change is involved.

Gandy2025:
Good catch — both of you converged on this, fixed in `c5fe4d7c`.

`hasUpdateProblem` is now **version-gated**: a `failed`/`blocked` attempt only counts as attention-worthy while the reported `sdkVersion` is still *behind* `attempt.target`. So a machine recovered by a manual `first-tree upgrade` — which re-registers on the new version but leaves the stale `{ result: "blocked", target }` record — is treated as **resolved** and drops out of `Needs attention`.

- Compare is dependency-free and channel-build aware: it reads every integer run in order (`0.5.3-staging.49.1` → `[0,5,3,49,1]`), so `sdkVersion >= target` (including a channel build at/after the target core) reads as resolved, while an older build of the same core stays behind.
- **Fails safe**: an unparseable/missing `sdkVersion` returns "still behind", so a genuinely-recorded failure is never silently hidden.
- Added pure coverage for: resolved (`sdkVersion` == and > target, and a channel build past the core), still-behind-by-build (same core, older build), and invalid/missing version.

Scope note: this is the client-side "distinguish unresolved from stale" option you offered. The fully robust root fix is the other one — have every successful recovery path (`commands/upgrade.ts`) overwrite/clear `update-state.json` / write an `ok` attempt — but that's CLI/server-side and outside this web audit view, so I've left it as a follow-up. The web gate resolves the false-positive for the dashboard itself.

(Full web suite 1428 pass; typecheck + lint:tokens + Biome clean. Also folded in an unrelated copy tweak in the same commit — the "Your computers" header button is now "Add computer" instead of "Connect", per a product request on this page.)

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1755 (follow-up on `c5fe4d7c`)

- Rationale: The stale-attempt gate is conceptually correct, but its numeric-run parser treats malformed or unknown-channel strings as valid comparable versions and can hide a real recorded failure instead of failing safe.

**Risk level**: C
- Path baseline: only `packages/web/**` -> C
- Semantic lift: none; this remains Web health derivation, copy, and tests.

**PR summary**
- Problem: Admins need the Team computers audit list to distinguish an unresolved failed update from stale attempt history after a successful manual recovery.
- Approach: Gate failed/blocked attempts on whether `sdkVersion` is still behind the attempt target, using a local version comparator; also rename the existing-machine header action from `Connect` to `Add computer`.
- Impacted modules: computer health derivation/tests and Settings → Computers action copy.

**Review findings**

❌ 1. Extracting every digit run is not a fail-safe version parser and violates the existing channel/version contract. [`packages/web/src/pages/clients/derive-status.ts:100`]

`versionParts` accepts any string containing digits, discards the prerelease identifier, and then compares the numbers lexicographically. That creates false “resolved” results for inputs the shared channel contract deliberately classifies as unknown. For example, current `1.4.0-alpha.999.1` versus target `1.4.0-staging.49.1` becomes `[1,4,0,999,1] > [1,4,0,49,1]`, so the recorded failure is hidden even though `alpha` is an unsupported/legacy channel and must fail closed. Likewise, an invalid value such as `release-2` can outrank a valid `1.4.0` target. The current invalid test only covers `dev`, which contains no digits and therefore misses this fail-open path.

The stable/prerelease semantics are also inverted by the added `1.4.0-staging.5.1 >= 1.4.0` expectation: standard SemVer treats that prerelease as lower than the stable release, while First Tree's durable channel contract says stable versions belong to prod, only `-staging.X.Y` belongs to staging, and unknown/mismatched prerelease shapes fail closed. Please validate the two accepted shapes/channels before comparing (or use an actual SemVer implementation plus explicit channel matching), and keep attention active for malformed, unknown, or mismatched-channel values. Add alpha/beta/rc, digit-bearing malformed strings, and channel-mismatch cases to the regression coverage. [R1, R4, R5]

✅ 2. For accepted version shapes, gating stale attempt history on reaching the target is the right lifecycle fix; the equal/ahead and same-staging-build comparisons now cover the intended recovery path.

✅ 3. The earlier update-signal rendering and narrow container-query fixes remain intact.

**Action taken**
- Submitted follow-up request changes.

yuezengwu (changes requested):
The stale-attempt lifecycle gate is the right direction, but `c5fe4d7c` still has one blocking false-negative path: the numeric-run comparator is neither SemVer-correct nor fail-safe for the channel/version contract.

`versionParts` accepts any string containing a digit and discards every prerelease identifier. That lets malformed or unsupported values hide a recorded failure:

- `garbage999` vs target `1.4.0` becomes `[999] > [1,4,0]`, so the attempt is considered resolved.
- `1.4` pads to `1.4.0` even though it is not a valid release version.
- `1.4.0-alpha.999.1` can outrank `1.4.0-staging.49.1` after the `alpha`/`staging` distinction is discarded, although unknown prerelease shapes must fail closed.

The new expectation that `1.4.0-staging.5.1` has reached stable `1.4.0` also reverses standard SemVer precedence (`prerelease < stable`) and treats a channel mismatch as success. The durable release contract permits stable versions for prod and `-staging.X.Y` for staging; unknown or mismatched prerelease shapes must remain attention-worthy.

Please validate the supported version shapes/channels before comparing (or use a real SemVer implementation plus explicit channel matching). Only a valid same-channel current version at or beyond the valid target should clear the historical failure; malformed, unknown-prerelease, or channel-mismatched values should keep it visible. Add coverage for digit-bearing malformed strings, alpha/beta/rc, stable↔staging mismatch, and valid same-channel staging ordering.

The “Add computer” copy change is coherent, and the earlier narrow-layout/update-rendering fixes remain accepted.

Verification on this head:
- Web typecheck + design-token gate passed
- 82 focused Web tests passed
- Biome passed on the changed files, with one unrelated pre-existing CSS warning

No database/schema or persisted-data change is involved.

Gandy2025:
Fixed in `af28c1aa` — the version gate now fails closed instead of parsing loosely.

Replaced the numeric-run comparator with strict parsing of only the two release shapes the channel contract accepts (mirroring `inferChannelFromVersion`):

- **prod** `X.Y.Z` and **staging** `X.Y.Z-staging.N.M`. Anything else — `garbage999`, partial `1.4`, `dev`, `alpha`/`beta`/`rc`, or otherwise malformed — parses to `null`.
- A recorded failure clears **only** when a *valid, same-channel* current version is at or beyond a *valid* target. So:
  - `garbage999` / `1.4` / `1.4.0-alpha.999.1` → unparseable/unknown → stays visible.
  - `1.4.0-staging.5.1` vs stable `1.4.0` → **channel mismatch** → stays visible (a staging prerelease has not reached a stable prod target; prerelease < stable).
  - `1.4.0` vs `1.4.0-staging.49.1` → mismatch → stays visible.
  - Same channel: prod `1.5.0 >= 1.4.0` and staging `0.5.3-staging.60.1 >= .49.1` → resolved; `0.5.3-staging.20.1 < .49.1` → still behind.
  - A malformed *target* also fails closed.

Regression coverage now includes digit-bearing garbage, partial versions, alpha/beta/rc, both staging↔prod mismatch directions, valid same-channel staging ordering, and a malformed target.

`typecheck` + `lint:tokens` + Biome clean; full web suite **1427 pass**.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1755

- Rationale: the new channel-shape gate fixes the reported garbage/prerelease/mismatch cases, but it still accepts SemVer-invalid numeric identifiers and compares them through lossy JavaScript numbers, leaving false-negative paths that can hide a recorded update failure.

**Risk level**: C
- Semantic lift: none; this remains Web presentation/derivation logic over an existing server response field.

**PR summary**
- Problem: admins need a compact Team-computer audit view that keeps unhealthy machines visible and usable at narrow Settings widths.
- Approach: replace the dense table with a grouped hostname-first list, surface failed/blocked self-updates, and clear historical failures only after a proven same-channel recovery.
- Impacted modules: Web Computers page, response model, status derivation, styles, fixtures, and Web tests.

**Review findings**

❌ 1. The parser is shape-strict but still not SemVer-valid or comparison-safe. [R4 / `packages/web/src/pages/clients/derive-status.ts:105`]

`\d+` accepts leading-zero numeric identifiers that the existing update paths' `semver.valid` rejects. For example, current `01.4.0` versus target `1.4.0`, or current `1.4.0-staging.049.1` versus target `1.4.0-staging.49.1`, parses as equal and clears the failure even though the current value is malformed and should fail closed. Converting each component with `Number.parseInt` also loses ordering beyond `Number.MAX_SAFE_INTEGER`; e.g. `9007199254740992.0.0` and `9007199254740993.0.0` collapse to the same numeric component, so a behind current version can be treated as recovered.

Please use SemVer validation plus explicit same-channel matching, or reject invalid numeric identifiers and compare validated decimal components without `Number` precision loss. Add coverage for leading-zero core/staging identifiers and a precision-boundary case.

✅ 2. The prior digit-bearing garbage, unknown prerelease, prod↔staging mismatch, valid same-channel ordering, narrow-layout, and stuck-update rendering findings are resolved in this head.

**Action taken**
- Submitted request changes on `af28c1aa07cb51489dac62223c4909cc44b78b2c`.

yuezengwu (changes requested):
Re-reviewed `af28c1aa07cb51489dac62223c4909cc44b78b2c`. The strict channel-shape gate closes the previously reported digit-bearing garbage, unknown-prerelease, and prod↔staging mismatch cases, but one blocking fail-open path remains.

The PR’s implementation goal remains sound: replace the dense Team computers audit table with a compact hostname-first health list, surface failed/blocked self-updates without permanently flagging recovered machines, and preserve identity at narrow Settings widths.

**Blocker — the parser still accepts SemVer-invalid numeric identifiers and compares through lossy JavaScript numbers.**

`parseSupportedVersion` uses `\d+` for every component and converts each match with `Number.parseInt`. As a result:

- current `01.4.0` versus target `1.4.0` parses as equal and clears the failure, even though the current version is invalid SemVer and must fail closed;
- current `1.4.0-staging.049.1` versus target `1.4.0-staging.49.1` has the same leading-zero problem;
- current `9007199254740992.0.0` versus target `9007199254740993.0.0` also compares equal because both adjacent identifiers collapse to the same IEEE-754 number, so an actually-behind client is treated as recovered.

In all three cases `updateResolved` returns true and `hasUpdateProblem` returns false, hiding a recorded failed/blocked update under `Ready`. Please either use SemVer validation plus explicit same-channel matching, or reject invalid numeric identifiers and compare validated decimal components without `Number` precision loss. Add regression coverage for leading-zero core/staging identifiers and the precision boundary.

The earlier update-signal rendering, stale-history lifecycle, channel mismatch, and narrow container-query fixes remain accepted. The “Add computer” copy is also coherent.

Verification on this head:
- Web typecheck + design-token gate passed
- 81 focused Web tests passed
- Biome passed on all seven changed files, with one unrelated pre-existing CSS `!important` warning

No database/schema or persisted-data change is introduced. The only response-model change is the Web DTO carrying an existing server field.

Gandy2025:
Fixed in `33b10a98` — both remaining fail-open paths closed.

- **Leading zeros**: version components now match `(0|[1-9]\d*)`, the SemVer numeric-identifier rule, so `01.4.0` and `1.4.0-staging.049.1` no longer parse → they fail closed (stay surfaced), same as `semver.valid` would reject them.
- **Precision**: identifiers are kept as validated digit **strings** and compared length-then-lexicographically (no `Number.parseInt`), so `9007199254740992.0.0` vs `9007199254740993.0.0` compare correctly instead of collapsing to the same IEEE-754 double — a behind client stays surfaced.

Added coverage for leading-zero core/staging identifiers and the 2^53 precision boundary (both directions). `typecheck` + `lint:tokens` + Biome clean; full web suite **1428 pass**.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1755

**Recommendation**: approve
- Rationale: the final revision closes the remaining SemVer fail-open and precision-loss paths, and all previously requested correctness and responsive-layout fixes are now present with focused regression coverage.

**Risk level**: C
- Semantic lift: none; the change remains Web presentation/derivation logic over an existing server response field.

**PR summary**
- Problem: admins need a compact Team-computer audit view that keeps unhealthy machines visible while preserving machine identity at narrow Settings widths.
- Approach: replace the dense table with a grouped hostname-first list, surface unresolved failed/blocked self-updates, treat recovered historical attempts as resolved only after a valid same-channel version reaches the target, and use a container-aware narrow layout.
- Impacted modules: Web Computers page, response DTO, status derivation, styles, development fixtures, and Web tests.

**Review findings**

✅ 1. The version gate now accepts only stable `X.Y.Z` and staging `X.Y.Z-staging.N.M` with SemVer-valid numeric identifiers, fails closed on malformed/unknown/mismatched channels, and compares identifiers as digit strings without IEEE-754 precision loss.

✅ 2. Focused tests cover unresolved/resolved same-channel attempts, malformed and unknown prereleases, both channel-mismatch directions, leading-zero identifiers, and ordering across the `2^53` boundary.

✅ 3. The earlier blockers remain resolved: failed/blocked attempts are visible in the attention group without permanently flagging recovered clients, and the container-query layout preserves hostname/meta width on narrow Settings surfaces.

**Action taken**
- Approved commit `33b10a98674185badec14077a8ba21f7d2fad330`.

yuezengwu (approved):
Approved `33b10a98674185badec14077a8ba21f7d2fad330`; no blocking findings remain.

The implementation now fulfills the redesign goal coherently: the admin Team computers audit surface is a compact hostname-first list, unresolved machine-health problems sort into `Needs attention`, recovered historical update attempts no longer remain permanently flagged, and the responsive container layout preserves machine identity on narrow Settings widths.

The final version gate closes the last correctness gap:

- only canonical prod `X.Y.Z` and staging `X.Y.Z-staging.N.M` shapes are accepted;
- every numeric identifier rejects leading zeros, so malformed values fail closed;
- prod/staging mismatches and unknown prereleases remain attention-worthy;
- validated identifiers stay as strings and compare by length then lexicographically, preserving exact order beyond `Number.MAX_SAFE_INTEGER`;
- the new tests cover both leading-zero forms and both directions across the `2^53` boundary.

The earlier response-field wiring, update-status rendering, stale-attempt lifecycle, group ordering, and narrow container-query fixes remain correct. The “Add computer” copy is also consistent with the steady-state action.

Verification:
- `pnpm --filter @first-tree/web typecheck` — passed, including design-token checks
- 82 focused Web tests — passed
- Biome on all seven changed files — passed with one unrelated pre-existing CSS `!important` warning

No database/schema or persisted-data change is introduced. The sole data-shape extension is the Web DTO carrying `lastUpdateAttempt`, an existing server response field.

PR #1756, opened 2026-07-14
feat(web): chat image sizing fix + click-to-zoom lightbox

Gandy2025:
## Problem

Inline chat images set only `maxWidth: 320`, with no height cap and no alignment control. Two visible defects:

- **Multi-image batches stretch/distort.** The gallery is a `flex flex-wrap` row, which defaults to `align-items: stretch`, so every image is stretched vertically to the tallest one. Measured: a 150×150 image rendered as **150×800**, a 600×600 as **320×800**.
- **A tall single image runs the full column** (400×1000 → 320×800), dominating the viewport, and there was **no way to see an image at full size**.

## Change

- **Single image** — shown at natural aspect within a width *and* height cap (no upscaling past source).
- **Multi-image batch** — a tidy **equal-height gallery** (preserves aspect, no stretch, no ragged sizing); a batch of one renders at the standalone size.
- **Click-to-zoom lightbox** — click any image to open a full-screen viewer built on the Radix Dialog primitive (Escape / focus-trap / scroll-lock for free): fits the viewport, **close** (X / Esc / backdrop) + **download original**, and **prev / next paging with a counter** for batches. Mobile inherits it via the shared chat surface (safe-area insets, translucent-dark control chips legible over any image, `focus-visible` rings).

New shared pieces: `useImageSrc` (cache→fetch image resolution, extracted so the thumbnail and lightbox never drift) and the `ImageLightbox` component. No schema/API changes.

## Testing

- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) — clean.
- `pnpm --filter @first-tree/web test` — **1417 passed** (adds a chat-view-dom test for the lightbox open/close interaction).
- Biome — clean.
- Real-browser QA (Chrome, light + dark, against real chat data): confirmed the 150×150 now renders 172×172 (was 150×800), equal-height gallery, single-image standalone sizing, and lightbox open / fit / download / prev-next / counter / Escape-close.
- Two independent code reviews (correctness + design/a11y); their findings (download-anchor reliability, control-chip contrast, focus rings, safe-area insets, non-interactive counter, aria-live) are folded into this PR.

### DIFF
--- a/packages/web/src/components/ui/image-lightbox.tsx
+++ b/packages/web/src/components/ui/image-lightbox.tsx
@@ -1,0 +1,176 @@
+import * as DialogPrimitive from "@radix-ui/react-dialog";
+import { ChevronLeft, ChevronRight, Download, X } from "lucide-react";
+import { useCallback } from "react";
+import { downloadAttachment } from "../../api/attachments.js";
+import { useImageSrc } from "../../lib/use-image-src.js";
+
+/**
+ * One image in a lightbox set. Either an `imageId` (resolved from the org
+ * attachment store / IndexedDB cache — full resolution, usually already warm
+ * from the thumbnail) or a ready `dataSrc` (inline base64 sends, which carry
+ * no attachment row).
+ */
+export type LightboxImage = {
+  imageId?: string;
+  dataSrc?: string;
+  filename: string;
+};
+
+// Control offsets keep clear of notch / home-indicator safe areas on mobile
+// (the chat surface — and this lightbox — is shared with the mobile app).
+const TOP = "calc(var(--sp-4) + env(safe-area-inset-top))";
+const RIGHT = "calc(var(--sp-4) + env(safe-area-inset-right))";
+const BOTTOM = "calc(var(--sp-4) + env(safe-area-inset-bottom))";
+const ARROW_LEFT = "calc(var(--sp-2) + env(safe-area-inset-left))";
+const ARROW_RIGHT = "calc(var(--sp-2) + env(safe-area-inset-right))";
+
+type ImageLightboxProps = {
+  images: LightboxImage[];
+  /** Index of the open image; `null` closes the lightbox. */
+  index: number | null;
+  onIndexChange: (index: number | null) => void;
+};
+
+/**
+ * Full-screen image viewer for chat images. Built on the Radix Dialog
+ * primitive (Escape / focus-trap / scroll-lock for free) but full-bleed:
+ * the image fits the viewport, with close + download always available and
+ * prev/next paging when the set has more than one image. Mobile inherits it
+ * unchanged (the chat surface is shared).
+ */
+export function ImageLightbox({ images, index, onIndexChange }: ImageLightboxProps) {
+  const open = index !== null && index >= 0 && index < images.length;
+  const current = open ? images[index] : undefined;
+  const multi = images.length > 1;
+
+  const close = useCallback(() => onIndexChange(null), [onIndexChange]);
+  const step = useCallback(
+    (delta: number) => {
+      if (index === null) return;
+      onIndexChange((index + delta + images.length) % images.length);
+    },
+    [index, images.length, onIndexChange],
+  );
+
+  const onDownload = useCallback(() => {
+    if (!current) return;
+    if (current.imageId) {
+      void downloadAttachment(current.imageId, current.filename);
+      return;
+    }
+    if (current.dataSrc) {
+      // Anchor must be in the document for a programmatic download click to
+      // fire in some browsers (matches downloadAttachment's approach).
+      const a = document.createElement("a");
+      a.href = current.dataSrc;
+      a.download = current.filename;
+      document.body.appendChild(a);
+      a.click();
+      a.remove();
+    }
+  }, [current]);
+
+  return (
+    <DialogPrimitive.Root open={open} onOpenChange={(next) => !next && close()}>
+      <DialogPrimitive.Portal>
+        <DialogPrimitive.Overlay className="fixed inset-0 z-[60] bg-overlay-scrim data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0" />
+        <DialogPrimitive.Content
+          aria-describedby={undefined}
+          onKeyDown={(e) => {
+            if (!multi) return;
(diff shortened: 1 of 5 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1756

- Rationale: The new fixed thumbnail caps can overflow the chat column on phone-width views, so the shared mobile chat surface regresses even though the desktop sizing and lightbox behavior are otherwise well-structured.

**Risk level**: C
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or the data layer

**PR summary**
- Problem: Chat participants currently see distorted multi-image batches, overly tall standalone images, and no way to inspect or download the original at full size.
- Approach: Preserve aspect ratio under thumbnail caps, render batches at a common height, and add an accessible Radix-based lightbox with download and paging while reusing one cache-to-fetch image resolver.
- Impacted modules: Workspace chat rendering, shared image resolution, dialog/lightbox UI, global lightbox tokens, and chat DOM tests

**Review findings**
❌ 1. Clamp both thumbnail variants to the actual message-column width. On a 390px phone viewport, the timeline's 48px horizontal padding and the message row's 20px avatar + 8px gap leave roughly 314px for the body, but `STANDALONE_IMG_STYLE` and `GALLERY_IMG_STYLE` allow 400px and 460px respectively. Because these are inline `maxWidth` values, they override the responsive `img { max-width: 100% }` preflight rule; a large standalone image or wide gallery tile can therefore create horizontal overflow/clipping in the shared mobile chat surface. Please make the button/image sizing container-aware while retaining the desktop caps and aspect behavior. [R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:1056`]
✅ 2. The extracted resolver preserves the existing immutable-id cache/fetch behavior, and the lightbox reuses the same full-resolution source instead of introducing a second image-loading path.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review summary

This PR fixes the existing chat-image distortion by separating standalone and gallery sizing, extracts the shared cache-to-fetch image resolver, and adds a Radix-based lightbox with download and batch navigation. The overall structure is sound, and there are no schema, API, database, or core data-structure changes.

## Blocking finding

Please make both thumbnail variants respect the actual message-body width. `STANDALONE_IMG_STYLE` permits `400px` and `GALLERY_IMG_STYLE` permits `460px`, but a 390px mobile viewport leaves only about 314px after the timeline's 48px horizontal padding and the message row's 20px avatar plus 8px gap. Because the inline `maxWidth` declarations win over the responsive image preflight, large standalone images and wide gallery tiles can overflow or be clipped in the shared mobile chat surface.

Please bound the thumbnail button/image by its container while preserving the desktop caps and aspect behavior, and cover the narrow presentation in the regression check.

Focused verification on `fc73c3c9` passed: web typecheck/token lint, the full `chat-view-dom.test.tsx` file (38 tests), and Biome on the changed files.

Gandy2025:
Thanks — addressed the blocking mobile-overflow finding from both reviews in `145a355b`.

**Fix**
- Both thumbnail caps are now `min(<desktop cap>, 100%)` (`min(25rem, 100%)` standalone, `min(28.75rem, 100%)` gallery — rem so the token lint passes), so they never exceed the message column while keeping the desktop caps and aspect behavior. The inline declaration no longer overrides the responsive `img { max-width: 100% }`.
- The wrapping button gets `minWidth: 0` so a gallery flex item can shrink below its intrinsic width on a narrow column instead of forcing overflow.

**Narrow-presentation coverage**
- Added a `chat-view-dom` regression assertion that the thumbnail stays container-relative (`min(..., 100%)`), not a bare fixed cap.
- Real-browser QA at a **390px** viewport on the mobile chat surface: `document.documentElement.scrollWidth - clientWidth === 0` (no horizontal overflow), zero images exceeding the viewport. The wide 1200×400 image now renders **314×105** (clamped to the ~314px column, was 400 wide); batch tiles stay equal-height and fit.

Full web suite: 1418 passed; typecheck + `lint:tokens` + Biome clean.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1756

**Recommendation**: approve
- Rationale: Commit `145a355b` resolves the blocking mobile overflow without changing the intended desktop caps or attachment-loading behavior, and the incremental change introduces no new blocking issue.

**Risk level**: C
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or the data layer

**PR summary**
- Problem: Chat participants currently see distorted multi-image batches, overly tall standalone images, and no way to inspect or download the original at full size.
- Approach: Render standalone and batched images with purpose-specific aspect-preserving sizing, add a Radix-based full-screen lightbox with download and paging, and share one immutable-id cache-to-fetch resolver between thumbnails and the viewer.
- Impacted modules: Workspace chat rendering, shared image resolution, lightbox UI, global lightbox tokens, and chat DOM tests

**Review findings**
✅ 1. The previous mobile-width blocker is resolved: both image variants now cap at `min(<desktop cap>, 100%)`, while the wrapping button is itself bounded to the column and can shrink as a gallery flex item. [`packages/web/src/pages/workspace/center/chat-view.tsx:1061`]
✅ 2. The narrow-width regression assertion prevents a return to a bare fixed cap, and the author's 390px browser verification covers the layout behavior that jsdom cannot measure. [`packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx:2095`]
✅ 3. The lightbox and thumbnail continue to share the existing full-resolution cache/fetch path; there are no schema, API, permission, or message-semantics changes.

**Action taken**
- Approved.

yuezengwu (approved):
## Re-review summary

Commit `145a355b` resolves the previous mobile-overflow blocker. Both thumbnail variants now use a container-relative `min(<desktop cap>, 100%)`, while the wrapping button is capped at the column width and has `minWidth: 0` so gallery flex items can shrink below their intrinsic width. This preserves the intended desktop caps and image behavior without introducing a mobile-only rendering path.

The added regression assertion guards against returning to a bare fixed cap, and the reported 390px browser verification covers the actual layout behavior that the DOM test environment cannot measure. There are no schema, API, database, permission, or message-semantics changes.

Focused verification on `145a355b` passed: Web typecheck/token lint, `chat-view-dom.test.tsx` (39 tests), `git diff --check`, and Biome with only the unrelated pre-existing `index.css:2226` warning.

No remaining blocking findings. Approved.

PR #1758, opened 2026-07-14
feat: add Context Tree review skill

baixiaohang:
## Summary

- Install `context-tree-review` as the sixth default First Tree skill without duplicating the generated Context Tree Policy.
- Keep Cloud webhook prompts limited to event facts and exclusive skill routing; the client performs the detached, validator-first, semantic review workflow.
- Submit the final commit-bound verdict through an agent-session-only `first-tree github context-review submit` command. Cloud validates the trusted reviewer run and publishes the GitHub review as the First Tree App.
- Require a ready, current-head, structurally and semantically passing review to produce `APPROVE`; blocking findings produce `REQUEST_CHANGES`, while draft, human-authority, stale, and execution-failure paths fail closed.

## App Publication Boundary

- GitHub App private keys and installation tokens remain in Cloud and are never returned to the client, persisted in review metadata, or logged.
- Only supported Context Reviewer webhook routing can create a trusted review run. Ordinary chat and agent-outbox messages cannot mint publication authority.
- Concurrent first-touch deliveries atomically reserve one reviewer chat per organization and pull request, keeping all same-PR run ordering in one publication domain. Delayed event suppression is valid only while the recorded reviewer and manager still match current authority.
- Cloud revalidates the configured reviewer, manager, client/runtime session, chat, run, bound repository, exact installation, open/non-draft PR state, exact head, live repository access, and `pull_requests: write` at the durable publication claim.
- Missing permission approval, missing repository coverage, suspended installations, stale heads, and GitHub failures never fall back to the host user's GitHub credentials.
- A durable pre-publication claim, paginated hidden run-marker reconciliation, and predecessor ordering provide exactly one logical outcome and ensure a newer-head App verdict is published after any unsettled older-head write.

## Repository Governance Prerequisites

- The staging canary confirmed that an App-authored `APPROVE` satisfies the repository's ordinary required-approval rule and that App-authored `REQUEST_CHANGES` blocks it.
- Before App approval becomes a required merge signal, the repository must enable stale-review dismissal on new pushes. This PR does not modify repository rulesets.
- The current Context Tree `main` branch has no `CODEOWNERS`; this change does not claim that an App review can replace Code Owner approval where that rule is in use.
- Existing installations must explicitly accept the upgraded `Pull requests: read and write` permission. New or upgraded installations remain limited to the repositories selected by their owners.

## Change Surface

- Client CLI and SDK: narrow Context Review submission contract.
- Server and Cloud: runtime-session enforcement, trusted-run authority, scoped App review publication, permission/readiness handling, and audit metadata.
- Web settings: permission-upgrade and action-required states.
- Skill installer, generated briefing, Cloud routing prompt, and static skill-eval coverage.
- No database migration or schema change. Durable Context Review submission and predecessor state is stored in the existing server-owned message metadata boundary.

## Validation

- `pnpm check` (passes with two existing warnings and one existing info diagnostic)
- `pnpm validate:skill` (6/6 skills)
- Server-focused suites: 219/219 tests
- Latest reviewer/manager authority follow-up: 39/39 focused tests and server typecheck
- CLI: 4/4 tests
- Client SDK: 2/2 tests
- Web settings: 17/17 tests
- `context-tree-review` static eval: 80/80 tests

Per product-owner direction, the expensive live/model skill eval was not rerun. Existing artifacts were assessed offline where applicable; the targeted deterministic and product suites above cover this revision.

Companion Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/733 (kept draft until this implementation merges and is reconciled against the final source head).


yuezengwu (changes requested):
Reviewed head `a7a85a4822e6ed579716341f40c37441033e800c`.

The implementation goal is clear: ship `context-tree-review` as the sixth default skill, reduce Cloud to event facts plus routing, and move the validator-first/head-bound verdict workflow and its live gates into the skill package. I found the following blockers.

1. **[P1] Make the review mutation commit-bound, not check-then-submit.** `skills/context-tree-review/SKILL.md` re-reads `headRefOid` and then calls `gh pr review`. That is still a TOCTOU window: `gh pr review` resolves only the PR id/number and its `AddReview` mutation omits a commit OID ([CLI flow](https://github.com/cli/cli/blob/v2.91.0/pkg/cmd/pr/review/review.go#L151-L181), [mutation](https://github.com/cli/cli/blob/v2.91.0/api/queries_pr_review.go#L248-L272)). If `synchronize` lands between the final view and the mutation, the approval can attach to the new, uninspected head—the exact stale-approval failure this skill is meant to prevent. Submit through an API path that includes the reviewed commit id (and fails closed when it is no longer current), then add a deterministic race case to the shim/gate.

2. **[P1] Resolve the `first-tree-read` routing collision in the real default family.** The generated briefing still says every `review`/`repo` signal must load `first-tree-read` *before acting* (`packages/client/src/runtime/agent-briefing.ts:746-751`), and that skill runs hierarchy discovery/semantic reads against the main tree checkout. A Context Reviewer wake-up therefore matches both skills, conflicting with this skill's validator-first detached-snapshot workflow and main-checkout integrity. The new gate hides the collision because its synthetic `AGENTS.md` lists only `context-tree-review` (`packages/skill-evals/src/suites/context-tree-review/fixture.ts:23-24`). Define an explicit exclusive boundary/precedence in the generated briefing and `first-tree-read` trigger, mirror it in the companion tree contract, and exercise the review skill with the actual full default family.

3. **[P2] The new static suite fails on macOS.** After the required build/typecheck prerequisites, `pnpm --filter @first-tree/skill-evals exec vitest run --maxWorkers=2 --minWorkers=1` finishes with 269/270 tests passing. `first-tree-shim.test.ts:202` expects the non-canonical `/var/...` path while the replay deliberately emits the realpath `/private/var/...`. Canonicalize the expected path (or consistently preserve the logical path) so the advertised suite passes on a supported developer platform.

4. **[P2] Enforce the eval guarantees that the PR/tree contract claims.** The validator shim says it requires a detached worktree but checks only SHA and cleanliness (`packages/skill-evals/src/core/shims/first-tree.ts:284-294`), so an attached worktree passes. Likewise `verifyFirst` only proves validation preceded the GitHub review (`grader.ts:196`), not that validation preceded semantic file reads. Add a detached-HEAD assertion and semantic-read ordering evidence before claiming the live gate locks detached integrity and validator-first review.

Validation performed locally:

- `pnpm check` — pass (pre-existing warnings only)
- `pnpm typecheck` — pass
- `pnpm validate:skill` — pass
- full `@first-tree/client` suite — 1,506 passed, 3 skipped
- targeted server suites — 80 passed
- full `@first-tree/skill-evals` suite — 269 passed, 1 failed as described above

There are no database/schema changes in this PR. Companion Context Tree PR #733 is currently blocked by a requested-changes review until it is returned to draft and reconciled after this code PR merges, which matches the tree/code sequencing policy.

yuezengwu (changes requested):
## Review outcome

This head addresses the prior workflow blockers: review creation is now commit-bound through the GitHub reviews API, `context-tree-review` routes exclusively over `first-tree-read`, the macOS realpath assertion is portable, and the fixture enforces a detached inspected head plus stronger read-only integrity checks.

Two deterministic-gate gaps still allow a passing result for traces that violate the skill contract, so I am requesting changes on `1286135e366a314171c22bd7f3f09a004010b393`.

### [P1] Require positive evidence of the actual skill and governed-file reads

`skillRead()` (`packages/skill-evals/src/suites/context-tree-review/grader.ts:13-19`) searches the entire serialized event, including command output. Consequently, reading only `AGENTS.md` can satisfy `skillFileReadObserved` when that output mentions `.agents/skills/context-tree-review/SKILL.md`; the skill itself need not be opened.

Likewise, `semanticReadAttempted()` (`grader.ts:438-457`) treats any command outside the pre-verify allowlist as a semantic read. A plain `ls .review-worktrees/42` therefore satisfies `semanticReadAfterVerify`, and `passing-ready-approves` has no body hint that would otherwise expose the omission (`cases.ts:39-48`).

I reproduced the false positive by passing a trace consisting of `cat AGENTS.md` (whose output names the skill), successful verification, `ls .review-worktrees/42`, final view, and the expected approval through `deriveMetrics()` / `casePassed()`. It returned:

```json
{
  "casePassed": true,
  "skillFileReadObserved": true,
  "semanticReadAfterVerify": true
}
```

Please derive these metrics from actual tool inputs / command operands and require an exact read of the review skill plus the complete changed governed-file set. Serialized output and an arbitrary non-allowlisted command are not sufficient evidence.

### [P1] Reject semantic review after validator failure

`casePassed()` bypasses the post-verify semantic-read requirement whenever `verifyMustPass` is false (`grader.ts:712-715`), but it never asserts the inverse. A trace can fail `tree verify`, then read `.review-worktrees/42/system/review-contract.md`, submit the validator-based request-changes review, and still pass. I reproduced exactly that sequence; the result was:

```json
{
  "casePassed": true,
  "semanticReadAfterFailedVerify": true
}
```

This contradicts the skill's validator-first contract at `skills/context-tree-review/SKILL.md:84-87` (“Do not continue to the full semantic review”) and the companion tree contract. Please record and reject semantic reads after a failed validator, with a regression test for the failing trace.

## Validation performed

- `pnpm build`
- `pnpm check` (passes with the existing two warnings and one info diagnostic)
- `pnpm typecheck`
- full static skill-evals suite: 27 files, 323 tests passed
- focused client briefing/bootstrap suite: 4 files, 72 tests passed
- focused server reviewer/webhook suite: 2 files, 80 tests passed
- `pnpm validate:skill`

There are no database/schema or core persisted-data-structure changes in this PR. The material surface is shipped skill topology, reviewer behavior, and its deterministic eval contract.

yuezengwu (changes requested):
## Review outcome

The follow-up moves the grader in the requested direction: it no longer treats a skill path echoed in arbitrary command output as a read, requires every governed changed path, binds positive evidence to the registered detached snapshot, and adds an explicit validator-failure metric. The previous exact reproductions are closed.

Two nearby traces still pass without satisfying those contracts, so I am requesting changes on `b43a7b4d810279955dc63a8934af9ce9cb4bf248`.

### [P1] Require agent-visible content, not only a successful path operand

`readerFileOperands()` records a recognized reader's file operands (`packages/skill-evals/src/suites/context-tree-review/grader.ts:368-405`), and `snapshotReadPaths()` treats the governed path as positive evidence whenever the command exits successfully (`grader.ts:503-551`). It does not reject output redirection or otherwise establish that any content reached the model.

I replaced the passing-ready semantic step with:

```sh
cat .review-worktrees/42/system/review-contract.md >/dev/null
```

The command exposes no file content to the reviewer, but the current grader reports:

```json
{
  "casePassed": true,
  "skillFileReadObserved": true,
  "semanticReadAfterVerify": true
}
```

The same operand-only issue permits zero-content slices such as `head -n 0`. Please require evidence that the successful, exact snapshot read returned content to the model (or use a dedicated read event/content fingerprint), and reject redirection/zero-content forms. Seeing the right filename in an executed command is not sufficient evidence of a semantic review.

### [P1] Make validator-failure short-circuit independent of the reader executable

`semanticReadAfterFailedVerify` is built only from `treeContentReadPaths()`'s enumerated shell readers and a short Git subcommand list (`grader.ts:718-736`, `827-833`). Other real readers are missed even though the broader `semanticReadAttempted()` classifier recognizes them.

I changed the validator-failure trace's post-verify read to:

```sh
awk 1 .review-worktrees/42/system/review-contract.md
```

The file contents are returned after the validator has failed, yet the grader reports:

```json
{
  "casePassed": true,
  "semanticReadAfterFailedVerify": false,
  "semanticReadBeforeVerify": false
}
```

Please fail closed on any post-failure access to detached-tree semantic content, except the narrowly allowed status/cleanup/final-view operations, rather than trying to enumerate reader names. Add this trace (and preferably another non-enumerated reader/native-tool form) as a regression.

## Validation performed

- `pnpm build`
- `pnpm check` (passes with the existing two warnings and one info diagnostic)
- `pnpm typecheck`
- full static skill-evals suite: 27 files, 334 tests passed; grader 75/75
- `pnpm validate:skill`
- both false-positive traces executed directly through `deriveMetrics()` / `casePassed()`

The follow-up contains no database/schema or core persisted-data-structure changes. Its material surface remains the skill/eval contract. The companion Context Tree PR remains draft and should be reconciled only after the code PR reaches its final merged form.

baixiaohang:
Following up on the latest review: after product-owner priority calibration, these two findings are being treated as non-blocking eval-harness limitations rather than runtime blockers.

- The full-content attribution gap can let a synthetic trace overstate whether semantic content reached the model.
- The post-validator-failure detector is not an exhaustive shell/tool sandbox and can miss unenumerated readers.

Neither path is used by the shipped `context-tree-review` skill, Cloud routing, or GitHub review submission at runtime. The shipped workflow still loads the dedicated skill, validates the detached inspected head first, binds the review to the inspected `commit_id`, routes exclusively over `first-tree-read`, and requires `APPROVE` for a passing ready PR.

The remaining grader limits are now recorded explicitly in the PR body. We are intentionally not expanding the shell parser further in this PR, and the head remains `b43a7b4d`; all CI and CodeQL checks pass. Please re-evaluate the current head with these two items treated as non-blocking test debt.

yuezengwu (approved):
Approved on `b43a7b4d810279955dc63a8934af9ce9cb4bf248` after the explicit product-owner priority calibration.

The implementation adds `context-tree-review` as the sixth default skill, makes Cloud responsible only for event facts and exclusive routing, and moves the validator-first detached-head workflow, commit-bound verdict submission, and deterministic behavioral gate into the skill package. The earlier runtime/workflow blockers are resolved: the review mutation is bound to the inspected commit, routing excludes `first-tree-read`, the fixture enforces the detached inspected head and read-only integrity, and the supported suite is green on macOS.

The two remaining synthetic grader bypasses are accepted for this release as non-blocking eval-harness debt: complete-content attribution is not a shell sandbox, and post-validator-failure detection does not enumerate every possible reader. They are documented in the PR body and do not change the shipped skill instructions, Cloud routing, or GitHub submission path. Companion Context Tree PR #733 remains draft and should reconcile this accepted boundary against the final merged implementation before becoming ready.

No tests were rerun for this body-only re-review because the code head is unchanged. The prior local validation still applies, and all current CI and CodeQL checks pass. There are no database/schema or core persisted-data-structure changes.

yuezengwu (changes requested):
## Review outcome

Reviewed head `e20e6692a39b9d567e01c34ee5446dcb5856d23c`.

The implementation is now materially broader than the previously approved skill-only revision: it adds the agent-session submission surface, a Cloud-owned GitHub App publisher, a durable submission/reconciliation/predecessor state machine, installation-readiness UI, and the sixth default review skill. I found no SQL migration or schema diff, but this **does introduce core persisted state and ordering data in `messages.metadata`** (`contextReviewSubmission` and `contextReviewBlockedByRunId`), so that durability/recovery boundary needs explicit human scrutiny.

I am requesting changes for the following correctness and authority gaps.

### [P1] Atomically reserve one reviewer chat per PR

[`findExistingReviewerChat` is a plain SELECT followed later by `createChat`](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-pr.ts#L255-L339), and the lookup itself has no lock or unique reservation ([lines 586–603](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-pr.ts#L586-L603)). Two different first-touch deliveries for the same PR (for example, `opened` and an immediate `synchronize`) have different delivery IDs, so both can observe no chat and create one. The run/current-head/predecessor protocol is scoped to a single chat; once split, each chat independently considers its run current, and a late old-head App review can publish after the new-head verdict. That defeats both the one-chat contract and the claimed cross-head publication ordering.

Please add an authoritative, race-safe get-or-create reservation keyed by organization + PR entity before creating the initial run (unique key or a shared serialization lock), and cover concurrent first deliveries plus a late old-head POST in a regression test.

### [P1] Bind revocable publication authority to the durable claim

The route validates runtime proof only once at ingress. The service then performs PR preflight ([lines 123–139](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-publisher.ts#L123-L139)) after reading the installation and minting a token ([lines 412–464](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-publisher.ts#L412-L464)), but the later point-of-no-return transaction ([lines 163–200](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-publisher.ts#L163-L200)) rechecks only the configured reviewer UUID and Context Tree binding. It does not revalidate or serialize against the active runtime/client/manager authority or the exact installation binding, suspension, and permission state.

Consequently, while token minting / PR preflight is in flight, an admin can disconnect and rebind the installation (the already minted token remains usable), or the runtime/client/reviewer authority can be revoked; the final claim still succeeds and the App mutation follows. In the rebind case, this can publish using authority that now belongs to another First Tree team. Make the durable claim the explicit authorization point, revalidate the complete current authority there, and serialize installation/runtime revocation and rebind operations against that claim (or bind a versioned authority lease to it). Add race tests for disconnect/rebind, suspension/permission downgrade, client retirement/session revocation, and reviewer/manager changes.

### [P2] Paginate hidden-marker reconciliation

[`listPullRequestReviewsForRun`](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/github-app.ts#L385-L400) fetches only `per_page=100` and never follows pagination. GitHub's list-reviews endpoint caps a page at 100 ([API documentation](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2026-03-10#list-reviews-for-a-pull-request)). Once a PR has more than 100 reviews, an unknown App write whose marker lies outside that page can never reconcile; because newer heads inherit the unresolved predecessor, publication can remain fail-closed forever. Follow pagination until the marker is found or all pages are exhausted, and test a marker on page 2.

### [P2] Normalize accepted commit OIDs before unknown-write matching

The public request schema accepts upper-case hexadecimal OIDs, and normal head validation compares case-insensitively, but [`reconcileUnknownSubmission` uses strict case-sensitive equality](https://github.com/agent-team-foundation/first-tree/blob/e20e6692a39b9d567e01c34ee5446dcb5856d23c/packages/server/src/services/context-reviewer-publisher.ts#L608-L625). GitHub can return the canonical lower-case commit ID after an accepted upper-case request, leaving that run permanently unresolved even though its marker is present. Normalize `reviewedHead` at the parse boundary or compare normalized values consistently, with an unknown-write regression for an upper-case OID.

baixiaohang:
Addressed all four findings in `36567f9087a1cae11f5c696b1e4b9fda49f794b9`:

1. Concurrent first-touch deliveries now use an atomic organization + PR chat reservation backed by the existing unique chat kickoff key. An `opened(A)` / `synchronize(B)` race converges on one reviewer chat, and a delayed old-head `opened` cannot supersede the established synchronize run.
2. The durable `pending -> submitting` claim is now the authorization point. In the claim transaction Cloud locks and revalidates the organization binding, configured reviewer, manager membership, client/runtime session, and exact installation ID, suspension, and `pull_requests: write` state before any App review POST. Race regressions cover installation rebind/disconnect, permission downgrade, suspension, client retirement, session revocation, reviewer replacement, and manager removal.
3. Hidden-marker reconciliation now paginates the GitHub review list through all pages while preserving duplicate-marker detection; the regression places the marker on page 2.
4. `reviewedHead` is canonicalized to lowercase at the shared parse boundary and reparsed at the service boundary, so claim hashing, App submission, and unknown-write reconciliation use one OID representation. The uppercase input case is covered.

Validation on this revision: server-focused suites 219/219, root `pnpm typecheck` 10/10, `pnpm check` passed with the existing two warnings and one info diagnostic, and `git diff --check` passed. Per product-owner direction, the expensive live/model skill eval was not rerun.

The durable submission/predecessor state remains in the existing server-owned message metadata boundary; this revision adds no SQL migration or relational schema change.

yuezengwu (changes requested):
## Re-review outcome

Reviewed head `36567f9087a1cae11f5c696b1e4b9fda49f794b9`.

This revision substantively addresses the four prior findings: the kickoff-key reservation converges concurrent first-touch chat creation; the durable claim now revalidates the configured reviewer, manager, client/runtime proof, and exact installation state; marker reconciliation paginates; and commit OIDs are normalized at the shared parse boundary. There is still no SQL migration or relational-schema change. The core persisted submission/predecessor state remains in the existing `messages.metadata` boundary.

One blocking edge remains in the first-touch ordering fix.

### [P2] Do not suppress the only run for the newly configured reviewer

The delayed-`opened` suppression runs after the handler has loaded the current reviewer and updated the chat metadata, but [`findSupersedingSynchronizeMessageId`](https://github.com/agent-team-foundation/first-tree/blob/36567f9087a1cae11f5c696b1e4b9fda49f794b9/packages/server/src/services/context-reviewer-pr.ts#L661-L680) matches only the chat, trigger, and head; the caller then returns that old message as successfully handled ([lines 293–304](https://github.com/agent-team-foundation/first-tree/blob/36567f9087a1cae11f5c696b1e4b9fda49f794b9/packages/server/src/services/context-reviewer-pr.ts#L293-L304)).

A concrete sequence is:

1. `synchronize(B)` creates the only run under configured reviewer R1.
2. An admin replaces the configured Context Reviewer with R2.
3. A delayed `opened(A)` delivery arrives.

The handler changes the chat's reviewer metadata to R2, finds R1's earlier synchronize message, suppresses the opened delivery, and creates no R2 run. R1's run is now correctly rejected by the new durable-claim authority checks, while R2 has no message/run to submit, so the webhook was acknowledged without leaving any publishable review task. I reproduced this with the existing server fixture: after the delayed event the chat still had exactly one run, owned by R1.

Please only suppress when the candidate synchronize run matches the current reviewer and current manager authority (or otherwise prove it is still publishable); when authority changed, append a fresh run for the current reviewer. Add the reviewer-replacement sequence above as a regression, and cover manager replacement through the same authority match.

Local validation on this head: the three focused server files passed `105/105`, root `pnpm typecheck` passed `10/10`, `pnpm check` passed with the existing two warnings and one info diagnostic, and `git diff --check origin/main...HEAD` passed.

baixiaohang:
Addressed in `09347bac665681d7b91e5758e07126dc6499332d`.

`findSupersedingSynchronizeMessageId` now accepts a synchronize run as the replacement for a delayed `opened` delivery only when both its reviewer agent UUID and reviewer-manager human-agent ID match the authority loaded for the current handler invocation. If either authority changed, the handler reuses the PR chat but appends a new run for the current reviewer/manager instead of acknowledging the old run as suppressed.

Added product regressions for both sequences:

- R1 handles `synchronize(B)`, configuration changes to R2, then delayed `opened(A)` arrives.
- Reviewer R keeps the same UUID, manager changes from M1 to M2, then delayed `opened(A)` arrives.

Both cases assert one shared PR chat, two run messages, and the newest run carrying the current reviewer and manager authority. The focused Context Reviewer PR suite passes 39/39, server typecheck passes, and `git diff --check` passes. No live/model skill eval was run.

yuezengwu (approved):
Reviewed head `09347bacc5b1228399433af0bb2de029617353ea`.

This PR installs `context-tree-review` as the sixth default skill, keeps Cloud webhook prompts limited to event facts and exclusive routing, and adds the agent-session-only GitHub App publication path with durable claim, recovery, and predecessor ordering.

The four blockers from my review of `e20e6692a39b9d567e01c34ee5446dcb5856d23c` remain resolved. This follow-up also resolves the delayed-`opened` authority gap from my review of `36567f9087a1cae11f5c696b1e4b9fda49f794b9`: superseding-synchronize suppression now requires both the recorded reviewer UUID and reviewer-manager human-agent ID to match the authority loaded for the current invocation, in both the existing-chat and reservation-loser paths. The added regressions cover reviewer replacement and manager replacement and assert a fresh run in the shared PR chat.

Validation performed on this head:

- focused Context Reviewer PR suite: 39/39 passed

No SQL migration or relational schema change is present. The PR does add core durable submission and predecessor state in server-owned message metadata (`contextReviewSubmission` / `contextReviewBlockedByRunId`), so that persistence, recovery, and authorization boundary still warrants human review. The documented stale-review-dismissal prerequisite and commit-bound-but-not-atomic-current-head boundary remain important deployment constraints.

I found no remaining blocking issue.

PR #1754, opened 2026-07-14
feat(web): enable mobile surface on the production release channel

Gandy2025:
## What

Enable the mobile web surface on the **production** release channel. It was gated to `dev` and `staging` only.

`useMobileExperienceState()` in `packages/web/src/app.tsx` now returns `enabled` for `dev`, `staging`, **or** `prod`. This one gate drives all three consumers: `/m/*` route admission, the phone-width auto-redirect from `/` to `/m/now`, and PWA head-tag injection (manifest / apple-touch). So prod users on a phone now get the mobile experience.

Fail-closed behavior is two distinct states, not one: while the channel fetch is **unresolved** (`settled: false`) the client renders neither shell — a neutral no-flash state; once **settled** with no usable channel (`channel: null`) a direct `/m/*` visit recovers to the **desktop root** and the manifest is omitted. Correctly-configured prod servers report `channel: "prod"` via `FIRST_TREE_CHANNEL`, so the unusable-channel path is a degraded-server fallback only.

## Scope unchanged

Mobile **capability scope is untouched** — still closed-by-default (`Now / Chat / Team / Me` + handoff primitives). Widening channel *availability* does not widen the mobile capability set.

## Verification

- `packages/web/src/__tests__/app-routes.test.tsx`: the `prod` route test is flipped from "disabled on prod" to **"opens the mobile experience on prod"** — it mounts the real `<App/>` router through the real gate (only the channel-state boundary is mocked) and asserts `/`, `/m`, `/m/now`, `/m/chat` render mobile plus the manifest link is present. Two distinct fail-closed cases are covered separately: the **unresolved-channel** test (`settled: false`) asserts the **neutral no-flash state** — neither shell renders, no manifest — while the new **settled-unusable** test (`channel: null, settled: true`) asserts **desktop-root recovery** for `/m/*` with the manifest omitted.
- `pnpm --filter @first-tree/web test` → **1417 passed**. `biome check`, `tsc --noEmit`, `lint:tokens` clean.
- Two independent reviews confirmed the channel gate is the only mobile restriction point, PWA assets (`manifest.webmanifest`, icons) are static and ship to every channel (no server-side gating, no service worker), so nothing half-ships on prod.

## Paired Context Tree PR

Records the production-enablement decision on the Mobile Surface node (its constraint previously required this as an explicit recorded decision): agent-team-foundation/first-tree-context#732

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[...]

PR #1759, opened 2026-07-14
fix(web): strip markdown from mobile card previews + honest install-guide copy

Gandy2025:
## What

Two small, independent mobile UX fixes found during a full staging E2E pass of the mobile surface (each its own commit).

### 1. Strip inline markdown from card previews
The Now feed and Chat list card previews rendered `description` / `lastMessagePreview` verbatim, leaking literal markdown markers (`**Task:**`, backticks) into the one-line glance. Now they run through the existing `stripInlineMarkdown` helper (already used by the compose status bar, resource snippets, and chat summary).

- Falls back on the **stripped** value: a preview that is only markup (e.g. an `` image, which the server stores verbatim in `lastMessagePreview`) strips to empty and must show the `"No messages yet."` placeholder, not a blank line.
- Guards both the pure `mobileChatPreview` function and the rendered card DOM path.

### 2. Drop false "no signing in again" install-guide claim
The add-to-home-screen sheet promised *"Opens instantly — no hunting for the link or signing in again"*. That benefit list is shared across all install modes, and the claim is **false on iOS**: an installed standalone PWA runs in a storage partition separate from Safari, so the session token doesn't carry over and the user signs in once. Dropped the "or signing in again" clause so the copy is honest on every platform.

## Verification

- `pnpm --filter @first-tree/web test` → **1422 passed** (added: preview stripping + strip-to-empty fallback + render-path guard + install-copy assertion). `biome`, `tsc`, `lint:tokens` clean.
- **Real browser**: ran my local build against staging data at iPhone viewport — confirmed Now/Chat previews render stripped (no `**`/backticks) and the install sheet no longer says "signing in again", 0 console errors.
- Two independent reviews; the first caught a strip-to-empty→blank-card regression in the initial version, now fixed and tested.

## Scope

Both are pre-existing issues, independent of the mobile prod-channel rollout (#1754). Not touched: the shared message renderer (heading rendering, wide-code-block overflow) — those are product-wide and belong in a separate pass.

### DIFF
--- a/packages/web/src/pages/mobile/__tests__/data.test.ts
+++ b/packages/web/src/pages/mobile/__tests__/data.test.ts
@@ -4,6 +4,7 @@ import {
   countAttentionRows,
   countUnreadRows,
   isNowFeedRow,
+  mobileChatPreview,
   mobileChatSignal,
   mobileFeedReasonLabel,
   sortMobileChats,
@@ -145,5 +146,30 @@ describe("isNowFeedRow (needs-attention admission)", () => {
     // unreadMentionCount also counts the implicit 1:1 DM auto-mention; only an
     // explicit @me qualifies.
     expect(isNowFeedRow(chatRow({ unreadMentionCount: 3, chatHasExplicitMentionToMe: false }))).toBe(false);
+  });
+});
+
+describe("mobileChatPreview", () => {
+  it("peels inline markdown so the card preview shows plain text", () => {
+    const preview = mobileChatPreview(
+      chatRow({ description: "**Task:** Build the tree from scratch (`first-tree-seed`)." }),
+    );
+    expect(preview).toBe("Task: Build the tree from scratch (first-tree-seed).");
+    expect(preview).not.toContain("**");
+    expect(preview).not.toContain("`");
+  });
+
+  it("prefers description, falls back to lastMessagePreview, then a placeholder", () => {
+    expect(mobileChatPreview(chatRow({ description: "  hello  ", lastMessagePreview: "later" }))).toBe("hello");
+    expect(mobileChatPreview(chatRow({ description: null, lastMessagePreview: "`only` this" }))).toBe("only this");
+    expect(mobileChatPreview(chatRow({ description: null, lastMessagePreview: "" }))).toBe("No messages yet.");
+  });
+
+  it("shows the placeholder when a markup-only preview strips to empty", () => {
+    // lastMessagePreview is stored verbatim server-side, so an image-only
+    // message arrives as `![](url)`, which strips to "" — must not render blank.
+    expect(mobileChatPreview(chatRow({ description: null, lastMessagePreview: "![](https://x/y.png)" }))).toBe(
+      "No messages yet.",
+    );
   });
 });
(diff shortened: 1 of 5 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1759

**Recommendation**: approve
- Rationale: The change reuses the existing inline-Markdown normalization path, preserves a visible fallback after normalization, and removes a cross-platform copy claim that is not true on iOS.

**Risk level**: C
- Path baseline: packages/web/** only -> C
- Semantic lift: none; this changes presentation derivation and shared install copy, not subscriptions, WebSocket parsing, or the data-access layer.

**PR summary**
- Problem: Mobile users currently see raw Markdown markers in Now and Chat card previews, while the shared install guide promises session continuity that iOS standalone PWAs do not provide.
- Approach: Route both real card surfaces through the existing stripInlineMarkdown helper, apply the empty-state fallback after stripping, and narrow the shared install benefit to a claim valid across modes.
- Impacted modules: packages/web/src/pages/mobile and focused mobile tests.

**Review findings**
✅ 1. Now and Chat both consume the same mobileChatPreview function, so the preview invariant has one implementation rather than surface-specific variants. [R1 / packages/web/src/pages/mobile/data.ts:57]
✅ 2. The fallback is evaluated after normalization, preventing markup-only previews from rendering a blank card; the tests cover the pure function and a rendered card path. [R4 / packages/web/src/pages/mobile/__tests__/data.test.ts:152]
✅ 3. The install benefit list is shared across modes, and the revised text avoids promising cross-container authentication behavior. [R5 / packages/web/src/pages/mobile/install-guide-sheet.tsx:6]

**Action taken**
- Approved.

yuezengwu (approved):
## Review summary

The goal is to make mobile card previews read as plain-text glance content and to remove an install-guide promise that First Tree cannot guarantee across platforms. The core changes are appropriately small: `mobileChatPreview` now routes the existing description/last-message choice through the shared `stripInlineMarkdown` helper and applies the empty-state fallback after normalization; the shared install benefit is narrowed to “no hunting for the link.”

## Findings

No blocking findings.

- Now and Chat already consume the same `mobileChatPreview` function, so both real card paths receive the same normalization without duplicating presentation logic.
- The post-normalization fallback is important and correct: an image-only `lastMessagePreview` can normalize to an empty string, and the card still renders “No messages yet.” rather than a blank row.
- The install-copy edit is justified by this codebase’s auth model: Web stores its access and refresh tokens in `localStorage` (`packages/web/src/api/client.ts:85`), while WebKit documents that iOS/iPadOS Add to Home Screen copies login state only when it is cookie-backed, not when it lives in other local storage ([WebKit Safari 17.2 notes](https://webkit.org/blog/14787/webkit-features-in-safari-17-2/#web-apps)). Removing the unconditional “no signing in again” claim is therefore the honest shared-mode copy.
- There are no database, schema, API-contract, or core data-structure changes in this PR.

## Verification

- New pure preview/fallback tests and install-copy tests: 15 passed.
- New rendered-card markdown test: passed.
- `pnpm --filter @first-tree/web typecheck`: passed, including design-token guardrails.
- Biome on all five changed files and `git diff --check`: clean.

Non-blocking baseline note: running the whole `density-dom.test.tsx` file locally also exposes the existing `renders Chat rows as medium list cards` async failure. I reproduced the identical failure at the PR base commit (`473325c1`), so it is not introduced by this change; the newly added render-path test passes independently.

Approved.

PR #1764, opened 2026-07-14
feat: add GitLab webhook Stage 2A backend

baixiaohang:
## Summary

Implements Stage 2A of the GitLab Self-Managed webhook proposal, with the persistence model simplified to match the human-approved product boundary:

- allows at most one GitLab connection per organization
- stores the current high-entropy URL bearer only as an irreversible hash and returns the complete URL only from create, regenerate, or replace
- hard-deletes the old connection and its mappings on replace/delete; stale replace requests fail closed against the expected current connection
- adds URL-bearer ingress with connection/IP-scoped plugin rate limiting, schema/body guards, secret redaction, zero-egress protections, and connection-scoped stable delivery claims
- normalizes Merge Request, Issue, and Note hooks into the shared SCM seam
- stores only entity-to-chat mappings: an event with no follow is not persisted, while a declared URL remains pending until the next matching event supplies numeric project identity
- delivers one basic GitLab card to existing followed chats through the generic trusted SCM send path

Proposal: https://github.com/agent-team-foundation/first-tree-context/pull/726  
Synchronized Draft proposal update: https://github.com/agent-team-foundation/first-tree-context/pull/740

## Durable storage

Migration `0077_flashy_changeling.sql` adds only two provider-specific tables:

- `gitlab_connections`: one row per organization, current bearer hash, display origin, and inbound health
- `gitlab_entity_chat_mappings`: pending path declarations or observed numeric project/entity identity bound to a chat

There is no endpoint-generation, lifecycle-audit, or standalone entity table in Stage 2A. Automatic-action acceptance/audit remains deferred to Stage 3 with the personnel-action capability it governs.

## Reliability and security boundaries

- Regeneration atomically invalidates the old bearer; there is no planned overlap window.
- Replace is resource-scoped and uses an expected-current CAS under the organization lock, preventing stale administrator actions from deleting a newer binding or resurrecting a deleted one.
- Replace/delete and inbound/follow effects remain serialized by the connection fence.
- Message kicks and inbox notifications are deferred until the outer ingress transaction commits.
- Only retry-stable `Idempotency-Key` / `webhook-id` values enter `processed_events`; UUID-only or missing-ID deliveries are not claimed.
- Stable provider IDs are made connection-scoped before claiming.
- The Cloud server never fetches the configured GitLab origin or payload URLs.
- Unknown endpoints use an IP rate-limit backstop; valid requests use a per-connection bucket.

## Explicitly out of scope

- Stage 2B Settings UI
- Stage 3 GitLab identity binding, automatic-action acceptance/audit, and personnel actions
- reviewer/assignee/mention target-chat creation, wake, automatic review, or echo pruning
- Local Client repository/source mapping
- new lease, worker, replay, raw-body, or per-side-effect idempotency systems

## Validation

- full server suite after the persistence simplification — 221 files / 2296 tests passed
- `pnpm --filter @first-tree/shared exec vitest run --maxWorkers=2` — 49 files / 576 tests passed
- web unit suite (`vitest run src`) — 177 files / 1437 tests passed
- final GitLab targeted suites — 2 files / 22 tests passed
- `pnpm check` — passed with three pre-existing unrelated diagnostics
- migration regeneration/no-drift and `git diff --check` — passed

Formal QA was not executed because no live external GitLab fixture was available; the cross-surface QA case is included. The root all-package test command also encounters the pre-existing client test environment issue caused by the host's empty `/tmp/.git`; all affected package suites above pass independently.

## Review

`codex-reviewer` approved the simplified worktree with no remaining blocking or non-blocking findings. The human owner selected the one-connection, hard-delete, two-table boundary reflected here.


yuezengwu (changes requested):
## What this change does

This implements the Stage 2A GitLab Self-Managed backend: provider-specific connection and endpoint lifecycle management, URL-bearer ingress, connection fencing and rate limits, MR/Issue/Note normalization, pending/observed entity follows, and basic GitLab cards through the shared trusted-SCM send path. The transaction/post-commit split and the preservation of the existing GitHub delivery path are both well covered by the implementation and tests.

## Core data / DB review required

Migration `0077_faithful_cerebro.sql` adds five durable tables (`gitlab_connections`, endpoint generations, lifecycle audit events, entities, and entity/chat mappings) plus their uniqueness, cascade, and retention semantics. The migration is additive, but this is a core schema/data-model change; a human owner should explicitly inspect those cardinality, deletion, and retention boundaries before merge.

## Blocking findings

1. **[P1] Do not claim `X-Gitlab-Webhook-UUID` as a retry-stable delivery ID** — `packages/server/src/services/gitlab-webhook.ts:82`

   GitLab documents `Idempotency-Key` and `webhook-id` as consistent across retries, while `X-Gitlab-Webhook-UUID` is unique for each webhook request. The current fallback therefore assigns a `stableDeliveryId` to deliveries that do not have one: a retry receives another UUID, enters `processed_events` under another claim, and can repeat the durable effects this seam is meant to protect. It also bypasses the proposal's explicit “no stable provider ID, no claim” behavior. Please accept only the retry-stable headers (`webhook-id` / `Idempotency-Key`, checking consistency if both are present), return `null` for UUID-only requests, and add regression coverage for both paths. GitLab's delivery-header contract is documented here: https://docs.gitlab.com/user/project/integrations/webhooks/#delivery-headers

2. **[P1] Keep GitLab IIDs project-scoped while resolving followers** — `packages/server/src/services/gitlab-entity-follow.ts:236`

   The observed-row lookup is scoped only by connection, chat, entity type, and IID. Because GitLab IIDs are project-scoped, an observed Issue `#42` in project A is treated as a conflict with Issue `#42` in project B for the same chat. I reproduced this with project IDs 501 and 502: both entities were observed and followed successfully, but a subsequent event for project 502 returned an empty audience because the arbitrary first observed row belonged to project 501. That suppresses legitimate cards, and the current `chat_already_bound_to_different_project` test encodes the incorrect cross-project restriction. Please remove that cross-project collision or scope it to the actual project/entity identity, and add a regression test proving one chat can follow the same type/IID in two numeric projects.

## Validation performed

- Targeted GitLab suites: 2 files / 19 tests passed.
- `pnpm check`: passed with the three pre-existing unrelated diagnostics.
- `git diff --check`: passed; review worktree is clean.
- A temporary regression test reproduced the cross-project same-IID failure above and was removed after confirmation.

Requesting changes for the two correctness/reliability blockers above.

baixiaohang:
Addressed both P1 findings in `2ab368fe`:

1. Stable delivery claims now accept only retry-stable `Idempotency-Key` / `webhook-id`. When both are present they must match; `X-Gitlab-Webhook-UUID` is ignored for claiming, so UUID-only deliveries follow the explicit no-stable-ID/no-claim path.
2. Observed follow collapse is now scoped by the projected numeric entity identity (`entityId`), preserving GitLab's project-scoped IID semantics. One chat can follow the same entity type/IID in multiple numeric projects, and subsequent deliveries resolve independently.

Regression coverage now includes both stable header paths, mismatched headers, UUID-only no-claim, and two projects sharing Issue IID `42` in one chat. Targeted GitLab suites pass (2 files / 20 tests), along with server typecheck, Biome, and diff checks. The independent core data/DB human-owner review gate for migration cardinality, deletion, and retention remains open and unchanged.

Please re-review the updated head.

yuezengwu (commented):
## Re-review of `2ab368fe`

The incremental change addresses both Stage 2A correctness findings:

- Stable delivery extraction now accepts only GitLab's retry-stable `Idempotency-Key` / `webhook-id`, requires them to agree when both are present, and leaves UUID-only deliveries on the explicit no-claim path.
- Follow collapse now keys the observed mapping by the projected entity identity, so project-scoped IIDs can coexist in one chat while duplicate declarations for the same numeric entity still converge.

I found no remaining code-level findings in the fix. The new regression coverage exercises both stable headers, mismatched headers, UUID-only no-claim, and two projects sharing Issue IID `42` in one chat.

Validation on this head:

- GitLab targeted suites: 2 files / 20 tests passed.
- Server typecheck passed.
- `pnpm check` passed with the same three pre-existing unrelated diagnostics.
- Incremental `git diff --check` passed.

The independent human-owner data-model gate remains open. Stage 2A creates five durable tables and commits cardinality, cascade-deletion, and audit-retention semantics while the linked Context Tree proposal is still marked Draft/non-canonical. This re-review clears the two P1 implementation blockers, but I am leaving the earlier Changes Requested review in place until a human owner explicitly accepts those schema boundaries; this comment is not that approval.

yuezengwu (approved):
Human owner has explicitly accepted the Stage 2A schema boundaries, including the proposed cardinality, cascade-deletion, and audit-retention semantics. The implementation findings on this head were already cleared in re-review, so I approve this PR.

yuezengwu (changes requested):
## What changed in this synchronization

This head simplifies Stage 2A from the earlier five-table lifecycle model to the selected two-table boundary: at most one GitLab connection per organization, one immediately replaceable URL bearer hash, hard replace/delete with mapping cascades, and mapping-only pending/observed entity state. Automatic-action acceptance/audit and standalone entity storage are deferred.

## Core data / DB review

Migration `0077_flashy_changeling.sql` now creates only `gitlab_connections` and `gitlab_entity_chat_mappings`, with a unique organization connection, hard cascade deletion, and no lifecycle-audit retention. This is still a core schema/data-model change; the human owner should inspect the one-connection, hard-delete, and two-table persistence boundary before merge. The implementation below currently has one blocker in the promised stale-admin safety.

## Blocking finding

1. **[P1] A concurrent Delete can still be resurrected by Replace** — `packages/server/src/services/gitlab-connections.ts:94-103`, `:141-150`

   `replaceGitlabConnection` locks the organization, but its current-connection read is not `FOR UPDATE`; `deleteGitlabConnection` locks only the connection row. If Delete acquires that row first, Replace can still lock the organization and read the uncommitted old row. Replace then blocks at its `DELETE`; after Delete commits, that statement affects zero rows, but the result is not checked and Replace inserts a new connection anyway. The administrator's completed Delete is therefore undone, violating the explicit expected-current CAS / “must not resurrect a deleted binding” contract.

   I reproduced this on PostgreSQL with a temporary regression test: hold the connection row lock in a delete transaction, start `replaceGitlabConnection`, then commit Delete. Replace resolved successfully with a new connection instead of throwing `changed or was removed`. The existing test deletes fully before starting Replace, so it does not exercise the race.

   Please make Replace acquire/revalidate the current connection row under the fence before proceeding (or serialize Delete under the same organization lock/order), fail if the expected row was concurrently removed, and add a deterministic Delete-first/Replace-waiting regression test.

## Validation

- GitLab targeted suites: 2 files / 21 tests passed.
- Full `pnpm typecheck`: 10/10 tasks passed.
- `pnpm check`: passed with the same three pre-existing unrelated diagnostics.
- Incremental `git diff --check` passed; the temporary reproducer was removed and the review worktree is clean.

Requesting changes for the concurrent Delete/Replace correctness blocker above.

yuezengwu (approved):
Re-review of `2541cc42` is complete.

The synchronization fixes the previously reported concurrent Delete/Replace resurrection: Replace now locks and revalidates the current connection row before deletion, Delete joins the organization-first lock order and revalidates the resource after acquiring it, and both mutations verify that their delete actually removed the expected row. The added PostgreSQL regression covers the Delete-first / Replace-waiting interleaving and confirms that Replace fails closed without recreating the binding.

The selected core persistence boundary remains the human-approved Stage 2A model: one GitLab connection per organization, immediate bearer invalidation, hard-delete cascades, and only `gitlab_connections` plus `gitlab_entity_chat_mappings`; lifecycle audit and standalone entity storage remain deferred.

Validation on this head:

- GitLab targeted suites: 2 files / 22 tests passed.
- Full monorepo typecheck: 10/10 tasks passed.
- Biome check passed with the same three pre-existing unrelated diagnostics.
- Full PR `git diff --check` passed.

No remaining blocking or non-blocking findings.

PR #1765, opened 2026-07-14
fix: file user issues/PRs into the work's repo, not First Tree's

Gandy2025:
## Problem

A new user bound their own source repo and told their agent to "file issue X" — meaning into their bound repo. The agent instead filed the issue into First Tree's own public repo (`agent-team-foundation/first-tree`), publishing the user's content there. This is reproducible for any user, not a one-off.

## Root cause

Issue creation is entirely skill/LLM-driven (there is no non-skill code path that creates issues). Two things combine:

1. `first-tree-file-bug` is a **core, always-installed** skill whose destination is **hardcoded** to `agent-team-foundation/first-tree` (in `SKILL.md` and the codex `openai.yaml` `default_prompt`). Its trigger surface overlaps a generic "file an issue" request.
2. The generated agent briefing had **no positive default** telling the agent that an issue/PR it files for the user targets the repo the work is about (the bound source repo). So the only concrete issue destination anywhere in a generic agent's context was First Tree's own repo — and a generic "提交 issue" over-routed there.

## Changes

- **Briefing — GitHub Working Posture** (`agent-briefing.ts`): add a positive default. Issues/PRs the agent creates for the user target the repo the work is about (the bound source repo) with an explicit `--repo` and a confirmed target. First Tree's own repo is reserved for work genuinely about First Tree, and a First Tree **platform defect** goes through `first-tree-file-bug`. (Deliberately not an absolute "only file here via the skill" rule — First Tree's own contributor agents legitimately open normal PRs/issues against this repo.)
- **`first-tree-file-bug` scope** (`SKILL.md`, `openai.yaml`, both briefing skill-map rows): narrow the trigger to First Tree platform defects and add an explicit "not for the user's own/bound repo — that's an ordinary `gh issue create --repo <their repo>`" negative, kept consistent across every routing surface.
- **Confirm-before-file** (`SKILL.md` step 5): the confirmation now states the destination repo explicitly, so a misroute is catchable by the user before anything is published.
- **Regression test** (`agent-briefing.test.ts`): lock the new briefing default.

## Testing

- `pnpm --filter @first-tree/client test` — 1506 passed, 0 failed (includes `agent-briefing` and the `retired-tree-commands-drift` guard).
- `pnpm check` (biome) — clean; `pnpm typecheck` — 10/10 tasks pass.
- Two independent reviews; one flagged that an earlier draft's "the only reason to file into First Tree's repo" phrasing was wrong for First Tree's own contributor agents — reworded here.

### DIFF
--- a/packages/client/src/__tests__/agent-briefing.test.ts
+++ b/packages/client/src/__tests__/agent-briefing.test.ts
@@ -412,6 +412,19 @@ describe("buildAgentBriefing — asking humans, GitHub, and CLI overview", () =>
     expect(briefing).toContain("If the current member is not an org admin");
     expect(briefing).toContain("hidden server sync path");
 
+    // Issues/PRs the agent files for the user default to the repo the work is
+    // about (the bound source repo), not reflexively First Tree's own repo.
+    // Regression guard for agents misfiling a user's issue into
+    // agent-team-foundation/first-tree; a First Tree platform defect still
+    // routes through `first-tree-file-bug`.
+    expect(briefing).toContain("target the repo the work is about with an explicit `--repo`");
+    expect(briefing).toMatch(
+      /Don't default to First Tree's own repository unless the work is genuinely about First Tree itself/,
+    );
+    expect(briefing).toContain(
+      "a First Tree platform defect specifically goes through the `first-tree-file-bug` skill",
+    );
+
     expect(briefing).toContain("Creating a PR or issue **never** follows it");
     expect(briefing).toContain("first-tree github follow <url>");
     expect(briefing).toMatch(/clearly unrelated to this chat's task/);
--- a/packages/client/src/__tests__/agent-briefing.test.ts
+++ b/packages/client/src/__tests__/agent-briefing.test.ts
@@ -1,4 +1,4 @@
-import { readdirSync, statSync } from "node:fs";
+import { readdirSync, readFileSync, statSync } from "node:fs";
 import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { beforeAll, describe, expect, it } from "vitest";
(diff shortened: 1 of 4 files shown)


baixiaohang (dismissed):
## Review by baixiaohang-code-reviewer - PR #1765

- Rationale: The routing fix is directionally correct, but the primary auto-discovery boundary that caused the incident is still not protected by a regression test.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: A generic request to file an issue for work in a user's bound repository can be misrouted to First Tree's public tracker, publishing the user's content in the wrong repository.
- Approach: Make the work's repository the always-present default, narrow `first-tree-file-bug` to First Tree platform defects, and expose the public destination before filing.
- Impacted modules: generated agent briefing and the `first-tree-file-bug` skill payload / Codex metadata.

**Review findings**
❌ 1. The new regression test only asserts the generated briefing, while the incident's primary routing surface is the `SKILL.md` frontmatter description and Codex also has an independent `openai.yaml` prompt. Neither existing drift tests nor this PR read those boundaries, so a future broadening of either surface can restore the exact misroute while this test remains green. This is especially inconsistent with `packages/skill-evals/src/core/case-schema.ts`, which documents the trigger boundary as covered by client unit/drift tests. Please add a deterministic guard that locks the First-Tree-only scope and the own/bound-repo exclusion across the shipped routing surfaces (including both family-map variants as applicable). [R5 / `skills/first-tree-file-bug/SKILL.md:3`, `skills/first-tree-file-bug/agents/openai.yaml:3`, `packages/client/src/__tests__/agent-briefing.test.ts:406`]

✅ 2. The behavior wording itself correctly makes the work's repository the default and gives users an explicit chance to catch a public First Tree destination before filing.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review summary

This PR is fixing the right failure mode: ordinary issue/PR creation should use the repository the work concerns, while `first-tree-file-bug` is reserved for First Tree platform defects and must disclose the public `agent-team-foundation/first-tree` destination before filing. There are no database, schema, or core data-structure changes.

## Blocker

The incident's routing boundary is not yet protected by a deterministic regression test. The new test only renders `agent-briefing.ts`; it never reads the shipped `skills/first-tree-file-bug/SKILL.md` frontmatter or `skills/first-tree-file-bug/agents/openai.yaml`, and it does not lock the hardened scope in both tree-bound/tree-less family-map rows. A later broadening of either discovery surface can therefore recreate the exact public-repo misroute while this test remains green.

This is especially important because `packages/skill-evals/src/core/case-schema.ts` deliberately excludes `first-tree-file-bug` from live evals and says its trigger boundary is covered by client unit/drift tests. Please add that deterministic cross-surface guard: assert the positive First-Tree-platform-only scope and the negative user-owned/bound-repo exclusion in the SKILL frontmatter and Codex metadata, plus both family-map variants. This matches the blocker already raised by @baixiaohang; I found no additional blocking behavior issue.

Local verification: the focused client tests passed (28/28), skill validation passed, and the skill-eval selector test passed (14/14).

Gandy2025:
Thanks @baixiaohang and @yuezengwu — addressed in 0e45ca48.

Regression coverage now spans **every shipped routing surface**, not just the generated briefing. New guard: `packages/client/src/__tests__/agent-briefing.test.ts` → *"locks first-tree-file-bug routing scope across every shipped surface"*. It reads the shipped payloads from disk plus the generated briefing and asserts, on each surface, **both** the positive First-Tree-platform scope **and** the negative "not the user's own/bound repo" exclusion:

- `skills/first-tree-file-bug/SKILL.md` frontmatter description
- `skills/first-tree-file-bug/agents/openai.yaml` (`default_prompt` + `short_description`)
- both family-map rows in `agent-briefing.ts` (tree-bound + tree-less)

So a future re-broadening of any one surface fails loudly instead of silently restoring the misroute. I also restored a concise `(not the user's own/bound repo)` exclusion to both family-map rows so the exclusion is locked there too, per the review.

This makes the `case-schema.ts` coverage note true again (file-bug's trigger boundary is covered by client unit/drift tests). Local: full `@first-tree/client` suite 1507 passed / 0 failed; `pnpm check` + `pnpm typecheck` clean.

yuezengwu (approved):
Re-reviewed at `0e45ca48`. The previous blocker is resolved.

The new deterministic guard reads the shipped `SKILL.md` and Codex `openai.yaml` directly, locks both the positive First-Tree-platform scope and the own/bound-repo exclusion, and verifies the tree-bound and tree-less family-map rows separately. Together with the existing GitHub Working Posture assertions, this now protects every routing surface implicated in the incident.

No database, schema, or core data-structure changes are involved, and I found no new blocking issue.

Local verification:

- focused client tests: 29/29 passed
- `@first-tree/client` typecheck: passed
- skill validation: passed
- Biome on changed TypeScript files: passed

Gandy2025:
@baixiaohang 你在旧 head `d7b6bbb3` 提出的 blocker 已在 `0e45ca48` 完整处理，请基于新 head re-review：

- 新增确定性跨面守卫，直接读取 shipped `SKILL.md` 与 Codex `agents/openai.yaml`，不再只测生成的 briefing；
- 对每个路由面同时锁住正向「仅 First Tree 平台缺陷」scope 与负向「不是用户自己/绑定仓库」排除；
- tree-bound / tree-less 两处 family-map 行分别断言，任一面未来被改宽都会使测试失败；
- 本地验证：focused client tests 29/29、完整 client suite 1507/1507、typecheck、skill validation、Biome 均通过；当前 CI 12/12 全绿。

该修复与 yuezengwu 当时提出的同一 blocker 对应；他已在 `0e45ca48` 上复核并 approve。此后 head 没有再变化，我也已重新请求你的 review。

PR #1766, opened 2026-07-14
feat: add cursor runtime provider (external Cursor Agent CLI)

yuezengwu:
## Summary

Adds `cursor` as a fourth runtime provider: the official **external** Cursor Agent CLI, driven one run-to-completion process per provider turn. First Tree does not bundle or install the Cursor engine and never touches its credentials — it resolves the operator-installed `cursor-agent` / `agent`, spawns it with the human-confirmed canonical posture, and observes the stream.

Canonical spawn contract (enforced by the runtime, not by prompt):

```
<resolved-cursor-binary> -p --output-format stream-json --sandbox disabled --force \
  [--model <exact-operator-value>] [--resume <stream-confirmed-session-id>]
```

- process cwd = agent home; prompt rides stdin only (never argv); `shell: false`
- no `--trust` / `--workspace` / `--approve-mcps` / `--stream-partial-output`, no `CURSOR_CONFIG_DIR`, no generated `.cursor/cli.json` — the operator's own Cursor login and permission config (explicit denies included) apply
- `--resume` only ever carries a stream-confirmed provider session id; synthetic pending ids stay runtime-local and upgrade via `replaceSessionId`
- settlement waits for child close + stdio drain: auth / invalid-model / quota failures often produce **no** `result` event, so non-zero exit + stderr is a first-class failure input
- replay-safety ladder: thinking / proven read-only activity stays retryable; assistant output or any unproven side effect ends automatic replay
- terminal / retry-exhausted failures ride the shared provider-retry events, so SessionManager posts the durable chat notice before ACK

## Changes by surface

- **shared** — provider enum, cursor runtime-config variant (free-form exact model id; deliberately **no** `reasoningEffort` — Cursor encodes effort in the model id), runtime-auth provider, `cursor_read_tool` / `cursor_write_tool` I/O sources
- **client** — `handlers/cursor/` (tolerant stream-json parser + per-turn handler), `runtime/cursor-binary.ts` (external-only resolution, `cursor-agent` before `agent`, memoized bounded `--version` smoke check), install-only capability probe, browser-OAuth login, error-taxonomy/retry-policy branches (provider-gated phrase matching), runtime-notice label
- **server** — Context Tree I/O derivation gated on `runtimeProvider=cursor`; config service rejects `reasoningEffort` writes for effort-less variants (was a silent zod-stripped no-op)
- **cli** — runtime-auth dispatch, doctor ordering, switch-drain classification of `cursor-agent`/`agent` (basename-scoped to avoid fail-closed false positives), agent-create help
- **web** — provider card with the official installer (`curl https://cursor.com/install -fsS | bash`, not npm), free-form model input with `auto` hint, reasoning-effort control hidden for cursor, in-product Connect, new-agent dialog, working-turn tool rows
- **qa** — `packages/qa/cases/runtime/cursor-provider.md` live-validation case

## Testing

- Parser fixtures are **real Phase 0 captures** from CLI `2026.07.09-a3815c0` (success, interleaved multi-tool, shell exit-7 + overall success, quota with no result, cross-process resume)
- 35+ new unit/integration tests across shared/client/server/cli/web, including an end-to-end server regression for the shell-read tree-ref path (the slice the closed prototype PR missed)
- `pnpm check` / `pnpm typecheck` green; shared 49 / client 147 / server 218 / web 178 test files pass (one pre-existing mobile density flake fails identically on main under full parallel load; passes standalone)
- **Live validated** on an isolated local stack (fresh Postgres + server + daemon from this branch): real capability probe (`cursor ok, path`), agent create/bind, two real Cursor turns with argv/cwd verified via `ps`/`lsof`, `--resume` session continuity with cross-turn recall, real `first-tree chat send` outbox delivery through the `--force` posture, invalid-model → terminal configuration classification with durable chat notice → recovery after config reset
- An adversarial multi-angle review (correctness / removed-behavior / cross-file / codex-contract parity / reuse / efficiency / conventions) ran before this PR; all confirmed findings are fixed in this branch

## Risk & QA

This PR touches **provider/auth, runtime, and WS/inbox** paths — formal QA is warranted per `packages/qa/AGENTS.md` (human-requested; case: `packages/qa/cases/runtime/cursor-provider.md`, paired with the runtime-provider readiness case for cross-provider regression).

Known accepted boundaries (per design): Cursor CLI is beta/auto-updating (tolerant parsing + version logging, no static ceiling); no `--trust` on the real agent home is not live-gated (a real trust failure must surface as an explicit configuration error); v1 excludes MCP materialization (one explicit unsupported diagnostic), mid-turn stdin inject, and model catalogs.

A paired Context Tree draft PR (client-runtime / runtime-auth / agent-behavior-steering updates) follows per the design's rollout plan.


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1766

- Rationale: The Cursor provider is broadly aligned with the runtime contracts, but two lifecycle/telemetry paths can strand real work or record the wrong Context Tree evidence.

**Risk level**: A
- Semantic lift: none (A is already the highest level); the diff also touches provider auth, session lifecycle, WS/inbox settlement, and Web runtime state.

**PR 摘要**
- Problem: Operators who already use Cursor need to run First Tree agents through their locally installed Cursor Agent CLI without First Tree bundling the engine or taking custody of credentials; login, model, quota, and runtime failures must remain recoverable and visible in the chat workflow.
- Approach: Adds Cursor as a fourth immutable runtime provider across agent creation/configuration, daemon capability/auth dispatch, a run-to-completion stream-json handler with confirmed-session resume and replay-safe retry, Web setup/model/login surfaces, Context Tree I/O attribution, and a committed live QA case.
- Impacted modules: `packages/shared`, `packages/client`, `packages/server`, `apps/cli`, `packages/web`, `packages/qa`.

**Review findings**
❌ 1. An admin-triggered resume permanently blocks later injected messages. `resume()` sets `initialTurnPreparing = true` before `prepareSession`, but clears it only inside `if (message)`. The handler contract explicitly allows `message === undefined` for admin resume; after that path returns, every later `inject()` queues successfully while `scheduleQueuedMessagesDrain()` keeps rejecting the drain because `initialTurnPreparing` never becomes false. The delivery is therefore owned but never enters Cursor or settles. Clear the guard for the message-less path (preferably in a shared `finally`) and cover `resume(undefined, ...)` followed by an inject. [R4 / `packages/client/src/handlers/cursor/index.ts:1140`]

❌ 2. Shell file-reference attribution ignores Cursor's actual working directory. The parser preserves `shellToolCall.args.workingDirectory` (and the checked-in real capture shows it can be non-empty), but both `providerRefsForTool()` and the emitted tool args replace it with the agent home `cwd`. A relative command such as `cat NODE.md` executed with `workingDirectory=<agent-home>/context-tree` is resolved as `<agent-home>/NODE.md`, so the Context Tree read can be omitted/misattributed and the working-turn UI reports a cwd different from where the provider executed it. Use the provider-reported working directory when present, with the agent cwd only as the empty-value fallback, and add the relative-path regression to the claimed shell-read coverage. [R5 / `packages/client/src/handlers/cursor/index.ts:349`]

⚠️ 3. A failed free-form model save cannot be retried with the same value. `lastSentRef` is cleared only when the controlled `value` prop changes; on a rejected PATCH the prop remains unchanged, so Enter/blur with the same intended model is suppressed forever until the user edits it to another string and back. The duplicate-submit latch needs to expire when the request settles (including failure), or be scoped only to the immediate Enter→blur sequence. [R4 / `packages/web/src/pages/agent-detail/model-section.tsx:132`]

✅ The install-only Cursor capability probe does not launch the provider binary, and the Web Connect affordance remains gated by a chat credential failure rather than being reintroduced on the computer card.

✅ Terminal/retry-exhausted provider failures flow through the shared provider-retry event and durable runtime-notice-before-ACK contract.

**Action taken**
- Submitted request changes.

yuezengwu:
Thanks for the review — all three findings addressed in a042484b.

**R4 / R1 — admin resume strands later injects** (`handlers/cursor/index.ts`): `resume(undefined, ...)` now clears `initialTurnPreparing` in the message-less branch (with `scheduleQueuedMessagesDrain()`), so an inject after an admin resume actually enters the provider and settles. Added a `resume(undefined)` → `inject` regression that asserts the queued turn spawns and completes.

**R5 / R2 — shell cwd attribution** (`handlers/cursor/index.ts`): added `shellEffectiveCwd(tool)` = Cursor's reported `workingDirectory` when non-empty, else the agent home. Both `providerRefsFromShellCommand` resolution and the emitted `tool_call.args.cwd` now use it, so a relative `cat NODE.md` executed with `workingDirectory=<home>/context-tree` resolves against the tree clone (not `<home>/NODE.md`) and the working-turn UI shows the real cwd. Added a relative-path shell-read regression asserting the repo-qualified tree ref and the reported cwd.

**R4 / R3 — model-save latch** (`web/model-section.tsx`): the duplicate-submit latch now expires on `onFocus` (and the existing value round-trip), scoping it to the single Enter→blur sequence. A failed save stays retryable with the same intended model — refocus + Enter fires it again. Regression covers no-double-submit-on-blur and retry-after-refocus.

`pnpm check` / `typecheck` green; cursor client + server config/IO suites pass. CI running.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1766

**Recommendation**: approve
- Rationale: The follow-up commit resolves all three prior findings with focused regressions, and the reviewed head has no remaining blocker.

**Risk level**: A

**PR summary**
- Problem: Operators who already use Cursor need First Tree agents to run through the locally installed Cursor Agent CLI while preserving operator-owned credentials, permissions, and observable failure recovery.
- Approach: Adds Cursor as an external run-to-completion runtime provider across configuration, capability and auth dispatch, stream handling, retry and durable notices, Context Tree I/O attribution, and Web setup controls.
- Impacted modules: `packages/shared`, `packages/client`, `packages/server`, `apps/cli`, `packages/web`, `packages/qa`.

**Review findings**
✅ `resume(undefined)` now releases the initial-turn guard and drains later injected work; the regression exercises provider spawn and settlement.
✅ Shell reads now use the provider-reported working directory for both reference resolution and emitted telemetry, with agent home as fallback.
✅ The model-save dedupe latch remains scoped to the Enter-to-blur sequence while allowing a same-value retry after refocus.
✅ No new blocker was found in the incremental diff through `a042484b`.

The author reports the relevant checks and tests as green; this review was static and did not rerun them. Formal QA remains appropriate for the provider, auth, runtime, and inbox surfaces.

**Action taken**
- Approved after explicit authorization.

PR #739, opened 2026-06-01
feat(web): redesign workspace conversation list rail

Gandy2025:
Redesigns the Workspace left conversation-list rail for lower visual density and a content-first, near-monochrome look (per `packages/web/DESIGN.md`). Color now appears only where it means something: attention, selected, unread.

## What changes (before → after)
**Rows**
- Two-line (title + preview, em-dash placeholder) → **single-line** (title + meta), uniform height, ~28px avatar.
- Per-agent avatar identity hue **desaturated** in the list (new `--avatar-hue-*-muted` tokens); vivid hues stay in chat header / roster / right rail.
- Removed the per-row leading **source icon** and the per-row **hairline** separator (whitespace + hover/selected bg separate instead).
- Unread = **red dot + bold title** (no numeric corner badge).
- Attention = left state border + a small state dot in the meta cluster (red = failed, amber = needs-you); still pinned into the top "Needs attention" bucket. No second line.
- Watching rows = neutral eye glyph (no `Watching ·` text prefix).

**Header** → two rows: `New chat` + filter `⚙` on row 1; `All / Unread / Watching` triad + a `Group ▾` dropdown on row 2 (no longer crowded into one row).

**Filter popover (⚙)** → renamed `Scope`→`Status`, `Origin`→`Source`; group options `By time / By source`. Group-by is pulled out of the popover into the header by default (`groupControlPlacement="header"`), so `⚙` holds only Status + Source.

## Notable internals
- New `RailFilter` triad type + `nextParamsForRailFilter` (atomic `?unread=`/`?watching=` URL write).
- `ChatRowAvatar` gains `muted` + `badge` props; `pickAvatarHue`/`resolveAvatarHue` gain an optional `muted` arg.
- 8 new `--avatar-hue-0..7-muted` tokens in `index.css` (same L/H, chroma ~0.06).
- DEV-only `/preview/conversation-list` page (real component + mock data, `import.meta.env.DEV`-gated) for visual review.

## Verification
- `pnpm --filter @first-tree/web typecheck` (tsc + `lint:tokens`) — green.
- `pnpm --filter @first-tree/web test` — 516 passed.

## Follow-ups (not in this PR)
- Context Tree `first-tree-cloud/web-console.md` "Conversation List" section is now stale and should be synced (separate tree PR).
- Open product question: scope (Active/Archived/All) currently lives in `⚙`; consider an "Archived" chip when scope ≠ Active.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1773, opened 2026-07-14
fix(cli): install root systemd service in system scope

bestony:
## Summary
- install Linux systemd services in system scope when the CLI is running as root, while keeping normal Linux users on systemd user units with linger behavior unchanged
- migrate legacy same-channel root user units only from the out-of-service install path; ambiguous root user-manager state now fails closed before writing/enabling the system unit
- keep `daemon refresh-unit` from stopping its containing legacy root user unit, and force root migration probes to UID 0 paths/bus instead of inherited sudo XDG/D-Bus values
- make systemd manager scope visible to daemon-start output so root installs print `journalctl -u ...` and user installs print `journalctl --user -u ...`
- update source docs/QA guidance and pair the durable Runtime Daemon Context Tree change in draft PR agent-team-foundation/first-tree-context#739

## Verification
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts`
- `pnpm exec biome check apps/cli/src/core/supervisor/systemd.ts apps/cli/src/core/supervisor/index.ts apps/cli/src/__tests__/service-install-core.test.ts`
- GitHub Actions on the latest pushed commit are green

## Notes
- `pnpm check` previously completed with existing warnings in `packages/client/src/__tests__/assistant-text.test.ts`, `packages/client/src/runtime/workspace-migrations.ts`, and `packages/web/src/index.css`.
- `pnpm test` was previously attempted and failed outside this change: `@first-tree/skill-evals` fixture git commit/timeouts, `@first-tree/web` draft-store `window.localStorage.clear` unavailable, and `first-tree-dev` batch 9 hit an environment-specific client-switch drain timeout before its EXDEV assertion.

### DIFF
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -809,6 +809,50 @@ describe("logout and upgrade commands", () => {
     expect(output).toContain("first-tree-dev daemon stop");
   });
 
+  it("refuses plain logout and purge while root systemd migration is required", async () => {
+    const credentials = join(tempDir, "credentials.json");
+    mkdirSync(tempDir, { recursive: true });
+    writeFileSync(credentials, "{}");
+    coreMocks.isServiceSupported.mockReturnValue(true);
+    coreMocks.getClientServiceStatus.mockReturnValue({
+      state: "unknown",
+      platform: "systemd",
+      detail:
+        "legacy root systemd user unit requires out-of-service migration: /root/.config/systemd/user/first-tree.service",
+      migrationRequired: "root-systemd-user-to-system",
+    });
+    coreMocks.loadCredentials.mockReturnValue({
+      accessToken: jwt({ sub: "user-old" }),
+      refreshToken: "refresh",
+      serverUrl: "https://first-tree.example",
+    });
+    coreMocks.readActiveRootClientId.mockReturnValue("client-1");
+
+    const { registerLogoutCommand } = await import("../commands/logout.js");
+    await expect(runTopLevel(registerLogoutCommand, ["logout"])).rejects.toMatchObject({
+      code: "DAEMON_MIGRATION_REQUIRED",
+      exitCode: 1,
+    });
+
+    let output = printLineMock.mock.calls.map((call) => String(call[0])).join("");
+    expect(output).toContain("Refusing to remove credentials");
+    expect(output).toContain("first-tree-dev login <code>");
+    expect(coreMocks.stopClientService).not.toHaveBeenCalled();
+    expect(cliFetchMock).not.toHaveBeenCalled();
+    expect(readFileSync(credentials, "utf8")).toBe("{}");
+
+    printLineMock.mockClear();
+    await expect(runTopLevel(registerLogoutCommand, ["logout", "--purge"])).rejects.toMatchObject({
+      code: "DAEMON_MIGRATION_REQUIRED",
+      exitCode: 1,
+    });
+
+    output = printLineMock.mock.calls.map((call) => String(call[0])).join("");
+    expect(output).toContain("Refusing to remove credentials");
+    expect(cliFetchMock).not.toHaveBeenCalled();
+    expect(readFileSync(credentials, "utf8")).toBe("{}");
+  });
+
   it("refuses purge before server retire or local deletion when service state is unknown", async () => {
     const credentials = join(tempDir, "credentials.json");
     const clientYaml = join(tempDir, "client.yaml");
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -450,10 +450,13 @@ describe("daemon start command", () => {
         "legacy root systemd user unit requires out-of-service migration: /root/.config/systemd/user/first-tree.service",
       logDir: "/logs",
       managerScope: "user",
+      migrationRequired: "root-systemd-user-to-system",
     });
 
     await expect(runStart()).rejects.toMatchObject({ exitCode: 1 });
     expect(output()).toContain("legacy root systemd user unit requires out-of-service migration");
+    expect(output()).toContain("Complete the root systemd migration out-of-service with `first-tree-dev login <code>`");
+    expect(output()).not.toContain("--foreground");
     expect(coreMocks.startClientService).not.toHaveBeenCalled();
     expect(coreMocks.ClientRuntime).not.toHaveBeenCalled();
   });
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -440,6 +440,22 @@ describe("daemon start command", () => {
 
     await expect(runStart()).rejects.toMatchObject({ exitCode: 1 });
     expect(output()).toContain("Service state could not be determined (launchd).");
+
+    stderrSpy.mockClear();
+    coreMocks.getClientServiceStatus.mockReturnValueOnce({
+      platform: "systemd",
+      state: "unknown",
+      label: "first-tree.service",
+      detail:
+        "legacy root systemd user unit requires out-of-service migration: /root/.config/systemd/user/first-tree.service",
+      logDir: "/logs",
+      managerScope: "user",
+    });
+
+    await expect(runStart()).rejects.toMatchObject({ exitCode: 1 });
+    expect(output()).toContain("legacy root systemd user unit requires out-of-service migration");
+    expect(coreMocks.startClientService).not.toHaveBeenCalled();
+    expect(coreMocks.ClientRuntime).not.toHaveBeenCalled();
   });
 
   it("runs inline, reconciles local state, uploads capabilities and skills", async () => {
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -329,6 +329,31 @@ describe("daemon start command", () => {
     expect(output()).toContain("Supervisor fallback: `journalctl --user -u first-tree`");
   });
 
+  it("prints system journal hints for root systemd services", async () => {
+    coreMocks.isServiceSupported.mockReturnValue(true);
+    coreMocks.getClientServiceStatus
+      .mockReturnValueOnce({
+        platform: "systemd",
+        state: "inactive",
+        label: "first-tree.service",
+        logDir: "/logs",
+        managerScope: "system",
+      })
+      .mockReturnValueOnce({
+        platform: "systemd",
+        state: "active",
+        label: "first-tree.service",
+        detail: "pid 123",
+        logDir: "/logs",
+        managerScope: "system",
+      });
+
+    await expect(runStart()).resolves.toBeTruthy();
+    expect(coreMocks.startClientService).toHaveBeenCalled();
+    expect(output()).toContain("Supervisor fallback: `journalctl -u first-tree`");
+    expect(output()).not.toContain("journalctl --user -u first-tree");
+  });
+
   it("starts an inactive launchd service and prints stdout/stderr fallback logs", async () => {
     coreMocks.isServiceSupported.mockReturnValue(true);
     coreMocks.getClientServiceStatus
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -459,6 +459,23 @@ describe("daemon start command", () => {
     expect(output()).not.toContain("--foreground");
     expect(coreMocks.startClientService).not.toHaveBeenCalled();
     expect(coreMocks.ClientRuntime).not.toHaveBeenCalled();
+
+    stderrSpy.mockClear();
+    coreMocks.getClientServiceStatus.mockReturnValueOnce({
+      platform: "systemd",
+      state: "unknown",
+      label: "first-tree.service",
+      detail:
+        "legacy root systemd user unit requires out-of-service migration: /root/.config/systemd/user/first-tree.service",
+      logDir: "/logs",
+      managerScope: "user",
+      migrationRequired: "root-systemd-user-to-system",
+    });
+
+    await expect(runStart(["--foreground"])).rejects.toMatchObject({ exitCode: 1 });
+    expect(output()).toContain("Service migration is required before foreground start");
+    expect(output()).toContain("Complete the root systemd migration out-of-service with `first-tree-dev login <code>`");
+    expect(coreMocks.ClientRuntime).not.toHaveBeenCalled();
   });
 
   it("runs inline, reconciles local state, uploads capabilities and skills", async () => {
--- a/apps/cli/src/__tests__/service-install-core.test.ts
+++ b/apps/cli/src/__tests__/service-install-core.test.ts
@@ -67,6 +67,7 @@ const originalArgv = [...process.argv];
 const originalExecPath = process.execPath;
 const originalCwd = process.cwd;
 const originalFirstTreeHome = process.env.FIRST_TREE_HOME;
+const originalSystemdSystemDir = process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR;
 const originalXdgConfigHome = process.env.XDG_CONFIG_HOME;
 
 function setPlatform(platform: NodeJS.Platform): void {
@@ -123,6 +124,7 @@ beforeEach(() => {
   home = tempHome();
   mkdirSync(home, { recursive: true });
   process.env.FIRST_TREE_HOME = join(home, "ft-home");
+  process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR = join(home, "systemd-system");
   process.env.XDG_CONFIG_HOME = join(home, "xdg");
   process.argv = ["node", "/repo/dist/cli/index.mjs"];
   setExecPath("/opt/node/bin/node");
@@ -139,6 +141,8 @@ afterEach(() => {
   rmSync(home, { recursive: true, force: true });
   if (originalFirstTreeHome === undefined) delete process.env.FIRST_TREE_HOME;
   else process.env.FIRST_TREE_HOME = originalFirstTreeHome;
+  if (originalSystemdSystemDir === undefined) delete process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR;
+  else process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR = originalSystemdSystemDir;
   if (originalXdgConfigHome === undefined) delete process.env.XDG_CONFIG_HOME;
   else process.env.XDG_CONFIG_HOME = originalXdgConfigHome;
   process.argv = [...originalArgv];
@@ -199,7 +203,11 @@ describe("service install helpers", () => {
     const unit = renderSystemdUnit({ kind: "bin", program: "/usr/local/bin/first tree" });
     expect(unit).toContain('ExecStart="/usr/local/bin/first tree" daemon start --no-interactive');
     expect(unit).toContain(`Environment=FIRST_TREE_HOME=${process.env.FIRST_TREE_HOME}`);
+    expect(unit).toContain("WantedBy=default.target");
     expect(unit).not.toContain("HTTPS_PROXY");
+
+    const systemUnit = renderSystemdUnit({ kind: "bin", program: "/usr/local/bin/first tree" }, "system");
+    expect(systemUnit).toContain("WantedBy=multi-user.target");
 
     const windowsWrapper = renderWindowsSupervisorCmd({
       kind: "node",
@@ -438,6 +446,47 @@ describe("service install helpers", () => {
       ["systemctl", ["--user", "enable", "--now", channelConfig.serviceUnitFile]],
       ["systemctl", ["--user", "is-active", channelConfig.serviceUnitFile]],
       ["systemctl", ["--user", "show", channelConfig.serviceUnitFile, "-p", "MainPID", "--value"]],
+    ]);
+  });
+
+  it("installs a root systemd system service without a user bus", () => {
+    setPlatform("linux");
+    userInfoMock.mockReturnValue({ uid: 0, username: "root" });
+    spawnSyncMock
+      .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
+      .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
+      .mockReturnValueOnce({ status: 0, stdout: "active\n", stderr: "" })
+      .mockReturnValueOnce({ status: 0, stdout: "987\n", stderr: "" });
+
+    const info = installClientService();
+    const unitPath = join(process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR ?? "", channelConfig.serviceUnitFile);
+    const unit = readFileSync(unitPath, "utf-8");
+
+    expect(info).toMatchObject({
+      platform: "systemd",
+      label: channelConfig.serviceUnitFile,
+      state: "active",
+      pid: 987,
+      unitPath,
+    });
+    expect(unit).toContain("WantedBy=multi-user.target");
+    expect(unit).toContain(`Environment=FIRST_TREE_HOME=${process.env.FIRST_TREE_HOME}`);
+    expect(spawnSyncMock.mock.calls.map((call) => [call[0], call[1]])).toEqual([
+      ["systemctl", ["daemon-reload"]],
+      ["systemctl", ["enable", "--now", channelConfig.serviceUnitFile]],
+      ["systemctl", ["is-active", channelConfig.serviceUnitFile]],
+      ["systemctl", ["show", channelConfig.serviceUnitFile, "-p", "MainPID", "--value"]],
+    ]);
+
+    spawnSyncMock.mockClear();
+    spawnSyncMock.mockReturnValue({ status: 0, stdout: "", stderr: "" });
+    expect(startClientService()).toEqual({ ok: true });
+    expect(stopClientService()).toEqual({ ok: true });
+    expect(restartClientService()).toEqual({ ok: true });
+    expect(spawnSyncMock.mock.calls.map((call) => [call[0], call[1]])).toEqual([
+      ["systemctl", ["start", channelConfig.serviceUnitFile]],
+      ["systemctl", ["stop", channelConfig.serviceUnitFile]],
+      ["systemctl", ["restart", channelConfig.serviceUnitFile]],
     ]);
   });
 
@@ -1244,13 +1293,14 @@ describe("service install helpers", () => {
 
   it("covers systemd install refresh and uninstall fallback messages", () => {
     setPlatform("linux");
-    userInfoMock.mockReturnValueOnce({ uid: 501, username: "" });
+    userInfoMock.mockReturnValue({ uid: 501, username: "" });
     spawnSyncMock
       .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
       .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
       .mockReturnValueOnce({ status: 0, stdout: "inactive\n", stderr: "" });
     expect(installClientService()).toMatchObject({ platform: "systemd", state: "inactive" });
     expect(printMocks.line).toHaveBeenCalledWith(expect.stringContaining("could not determine username"));
+    userInfoMock.mockReturnValue({ uid: 501, username: "gandy" });
 
     spawnSyncMock.mockReset();
     spawnSyncMock.mockReturnValueOnce({ status: 1, stdout: "", stderr: "" });
--- a/apps/cli/src/__tests__/service-install-core.test.ts
+++ b/apps/cli/src/__tests__/service-install-core.test.ts
@@ -69,6 +69,8 @@ const originalCwd = process.cwd;
 const originalFirstTreeHome = process.env.FIRST_TREE_HOME;
 const originalSystemdSystemDir = process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR;
 const originalXdgConfigHome = process.env.XDG_CONFIG_HOME;
+const originalXdgRuntimeDir = process.env.XDG_RUNTIME_DIR;
+const originalDbusSessionBusAddress = process.env.DBUS_SESSION_BUS_ADDRESS;
 
 function setPlatform(platform: NodeJS.Platform): void {
   Object.defineProperty(process, "platform", { configurable: true, value: platform });
@@ -80,6 +82,18 @@ function setExecPath(value: string): void {
 
 function tempHome(): string {
   return join(tmpdir(), `ft-service-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+}
+
+function rootHome(): string {
+  return join(home, "root");
+}
+
+function rootUserInfo(): { uid: 0; username: string; homedir: string } {
+  return { uid: 0, username: "root", homedir: rootHome() };
+}
+
+function rootLegacySystemdUserUnitPath(): string {
+  return join(rootHome(), ".config", "systemd", "user", channelConfig.serviceUnitFile);
 }
 
 function windowsProcessIdentityJson(
@@ -145,6 +159,10 @@ afterEach(() => {
   else process.env.FIRST_TREE_SYSTEMD_SYSTEM_DIR = originalSystemdSystemDir;
   if (originalXdgConfigHome === undefined) delete process.env.XDG_CONFIG_HOME;
   else process.env.XDG_CONFIG_HOME = originalXdgConfigHome;
+  if (originalXdgRuntimeDir === undefined) delete process.env.XDG_RUNTIME_DIR;
+  else process.env.XDG_RUNTIME_DIR = originalXdgRuntimeDir;
+  if (originalDbusSessionBusAddress === undefined) delete process.env.DBUS_SESSION_BUS_ADDRESS;
+  else process.env.DBUS_SESSION_BUS_ADDRESS = originalDbusSessionBusAddress;
   process.argv = [...originalArgv];
   setExecPath(originalExecPath);
   process.cwd = originalCwd;
@@ -451,7 +469,7 @@ describe("service install helpers", () => {
 
   it("installs a root systemd system service without a user bus", () => {
     setPlatform("linux");
-    userInfoMock.mockReturnValue({ uid: 0, username: "root" });
+    userInfoMock.mockReturnValue(rootUserInfo());
     spawnSyncMock
       .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
       .mockReturnValueOnce({ status: 0, stdout: "", stderr: "" })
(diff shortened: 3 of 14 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773

- Rationale: Fresh root installs select the correct system manager, but the root user-to-system scope transition is not migrated atomically and the operator contract still points at the user manager.

**Risk level**: A
- Semantic lift: none; already at the highest path level

**PR summary**
- Problem: Root-operated Linux hosts cannot rely on a systemd user bus, so login/daemon setup fails even when the system manager is available.
- Approach: Select system scope for uid 0, write the unit under `/etc/systemd/system`, use `multi-user.target`, skip linger, and route lifecycle commands through system `systemctl`; keep non-root behavior unchanged.
- Impacted modules: CLI systemd supervisor and service-install regression tests

**Review findings**

❌ 1. Migrate the existing same-channel root user unit before enabling the system unit. [`apps/cli/src/core/supervisor/systemd.ts:163`]

Previous releases installed root's unit under `/root/.config/systemd/user`, but this path now only writes and controls `/etc/systemd/system/<unit>`. On upgrade/login, the old user unit can remain enabled or running while `enable --now` starts the new system unit. The in-daemon auto-update path is especially risky: the old root user service invokes the new `daemon refresh-unit`, which starts the system unit, then exit 75 lets the old user manager restart its own unit, leaving two daemons sharing the same `FIRST_TREE_HOME` and client identity. Status, stop, and uninstall then only target system scope and cannot control the orphan. Please add an explicit, same-channel root user -> system migration (stop/disable/remove/reload before enabling the replacement) and regression coverage that starts from a legacy root user unit.

❌ 2. Make root operator guidance and durable documentation scope-aware. [`apps/cli/src/commands/daemon/start.ts:181`]

After starting a root system service, the CLI still prints `journalctl --user`, which queries the wrong manager. `docs/cli-reference.md:624-627` also says Linux always uses `systemd --user`, and the Runtime Daemon Context Tree still records Linux installs as user-level. Please surface enough scope information for the correct journal hint, update the CLI reference, and pair the behavioral change with the matching Context Tree update.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
The implementation correctly routes a fresh uid-0 Linux install through the system manager (`/etc/systemd/system`, `multi-user.target`, and system `systemctl`) while preserving the non-root user-manager path. I am requesting changes for two release-blocking transition/contract gaps:

1. Please migrate the existing same-channel root user unit before enabling the system unit. Previous versions can leave `/root/.config/systemd/user/<channel-unit>` enabled or active. The new scope lookup ignores it and `enable --now` starts a second system unit. During managed auto-update, the old user unit can invoke the new `daemon refresh-unit`, which starts the system unit, and then restart itself after exit 75; both daemons then share `FIRST_TREE_HOME` and the same client identity. The migration must stop/disable/remove/reload the exact legacy root unit before starting its replacement, fail safely if exclusivity cannot be established, preserve any baked-proxy migration, and have regression coverage starting from that legacy state.

2. Make operator guidance and durable docs scope-aware. `daemon start` still prints `journalctl --user` for every systemd service, which is wrong for uid 0. `docs/cli-reference.md` and `docs/onboarding-guide.md` still describe Linux installs as uniformly per-user, and the Runtime Daemon Context Tree records the same contract. Please expose enough scope in service status for the correct hint, update the source documentation, and pair/cross-link the matching Context Tree change before merge.

Local verification on `91c092c4`: the service-install and daemon-start suites passed (72 tests), as did the CLI typecheck and Biome check. Those tests cover the fresh root path but not the legacy-scope transition or a root-specific journal hint.

bestony:
Follow-up pushed in cd9a4d01.

Addressed the requested changes:

1. Root legacy user-unit migration
- Before enabling the root system-scope unit, the installer now checks the same-channel legacy root user unit path.
- It migrates baked proxy env from that legacy unit, attempts `systemctl --user disable --now <unit>` with a root user-bus env fallback, removes the legacy unit file, and reloads the user manager when reachable.
- It fails closed on non-bus stop/disable errors to avoid leaving two daemons, while tolerating the original no-user-bus condition by removing the stale unit file before enabling the system unit.
- Added regression coverage for successful migration, unavailable user bus, and stop/disable failure.

2. Scope-aware operator guidance/docs
- Service status now carries `managerScope` for systemd.
- `daemon start` prints `journalctl -u ...` for root/system units and keeps `journalctl --user -u ...` for normal user units.
- Updated CLI/onboarding/proxy docs and the portable-shell onboarding QA case to distinguish normal-user systemd from root system-scope systemd.

Verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts` (76 tests passed)
- `pnpm typecheck`
- `pnpm check` (completed with existing unrelated warnings in client/web files)

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773 (follow-up on cd9a4d01)

- Rationale: The scope-aware status, journal hint, source docs, and happy-path migration are now covered, but an unreachable root user bus is still treated as proof that the legacy daemon is absent, so the duplicate-daemon release blocker remains.

**Risk level**: A
- Semantic lift: none; already at the highest path level

**PR summary**
- Problem: Root Linux installs need system scope because a root user bus may not exist.
- Follow-up approach: Migrate the same-channel legacy root user unit, expose systemd manager scope to operator output, and update source documentation and QA guidance.
- Impacted modules: systemd supervisor, daemon start output, service tests, CLI/onboarding/proxy docs, and portable onboarding QA

**Review findings**

❌ 1. Do not continue the scope migration when the legacy user manager cannot be reached. [`apps/cli/src/core/supervisor/systemd.ts:189`]

When `disable --now` fails with a bus-connect error, lines 190-198 only warn, delete the unit file, and then allow the system unit to be enabled. An unreachable bus does not establish exclusivity: the repository's documented WSLg over-mount case can leave the root user manager and its already-loaded daemon running while the bus path is inaccessible. Deleting a unit file does not stop an already-loaded service, so this path can still start the second daemon that the migration is meant to prevent. Please proceed only after positive evidence that the legacy service is stopped/absent (or independently stop a trusted live runtime); otherwise fail closed before writing/enabling the system unit. The no-bus regression should cover the ambiguous live-manager case and assert that system-scope enablement is refused, rather than codifying continuation.

❌ 2. The matching durable Context Tree update is still missing. [`docs/cli-reference.md:624`]

The source docs and operator hint are now scope-aware, but the normal Runtime Daemon node still states that Linux login installs a user-level service, and I could not find a paired/cross-linked Context Tree PR for this change. Please link the matching tree update before approval so the durable service-lifecycle contract does not remain stale.

✅ The new `managerScope` propagation and root-specific `journalctl -u` coverage resolve the source-side operator-guidance portion of the prior finding.

**Action taken**
- Submitted follow-up request changes on cd9a4d01.

yuezengwu (changes requested):
The follow-up correctly adds `managerScope`, fixes the root journal hint, updates the source docs/QA case, and covers a standalone happy-path migration. I am still requesting changes because the transition is not safe in the two runtime paths that matter most:

1. **Do not stop the containing user unit from `daemon refresh-unit`.** `refreshForUpdate` still calls `installSystemd()`, so a managed update from a legacy root user service reaches `systemctl --user disable --now` here. But `daemon refresh-unit` is a child inside that same service cgroup. Stopping the unit terminates the old daemon and, with the rendered `KillMode=mixed`, systemd sends the subsequent SIGKILL to the remaining cgroup processes after the main process exits—including the migration helper before it can remove the old file and install/enable the system unit. This also contradicts `supervisor/index.ts:43-49`, whose handoff invariant is that refresh must not unload its parent. The likely result is an offline, disabled legacy service with no replacement. Please separate the in-daemon refresh path from an out-of-service migration (or defer/fail it safely) and cover the managed-update handoff, not only a standalone `installClientService()` call. See the upstream [`KillMode=mixed` contract](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=).

2. **A missing user-bus connection is not proof that the legacy daemon is absent.** On a bus-connect error, the code warns, removes the file, and proceeds to `enable --now` the system unit. The repo's WSLg case explicitly documents a live user manager whose bus path is hidden; deleting its unit file does not stop the already-loaded service. Proceed only after positive evidence that the legacy service is stopped/absent; otherwise fail closed before writing or enabling the replacement. The no-bus test should assert refusal for the ambiguous live-manager case instead of encoding continuation. While revising this, force the UID-0 bus (`/run/user/0`) rather than inheriting `XDG_RUNTIME_DIR` / `DBUS_SESSION_BUS_ADDRESS`, which may belong to a sudo caller and must not let root migration target another user's manager.

3. **Link the matching Context Tree update.** The source docs are now scope-aware, but `system/cloud/runtime/daemon.md` on current `main` still says Linux login always installs a user-level service, and no `first-tree-context` PR referencing #1773 is open. Please pair/cross-link that durable contract change before approval.

Local verification on `cd9a4d01`: service-install, daemon-start, and update-glue suites passed (82 tests total), as did the CLI typecheck and Biome check. These mocked tests do not exercise systemd's real cgroup/stop semantics.

bestony:
Second follow-up pushed in aea1e59a.

Addressed the latest requested changes:

1. Ambiguous root user manager state now fails closed
- If a same-channel legacy root user unit exists and `systemctl --user disable --now <unit>` cannot reach the user bus, install now aborts with an ambiguous-state error.
- The legacy unit file is left in place and the system-scope unit is not written/enabled, so the code no longer treats an unreachable user bus as proof that no legacy daemon is running.
- Updated the no-bus regression to assert refusal instead of continuation.

2. Durable Context Tree update
- Opened the matching Runtime Daemon Context Tree PR: agent-team-foundation/first-tree-context#739
- The tree node records normal-user Linux user units, root system-scoped systemd units, and the fail-closed legacy root migration constraint.
- `first-tree-dev tree verify --tree-path /tmp/first-tree-context-hlzKKt` passed.

Verification for aea1e59a:
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts` (76 tests passed)

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773 (follow-up on aea1e59a)

- Rationale: The ambiguous-bus path and durable Context Tree contract are now fixed, but the managed-update migration still stops its own containing unit and the root user-manager probe can still inherit another user's bus environment.

**Risk level**: A
- Semantic lift: none; already at the highest path level

**PR summary**
- Problem: Root Linux installs need a system-scoped daemon without risking a second daemon during migration from older root user units.
- Follow-up approach: Fail closed when the legacy root user manager is unreachable and pair the source change with first-tree-context#739.
- Impacted modules: systemd supervisor migration and service-install regression tests

**Review findings**

❌ 1. Keep the in-daemon refresh path from stopping its own containing legacy unit. [`apps/cli/src/core/supervisor/systemd.ts:189`]

`systemdBackend.refreshForUpdate` still aliases `installSystemd`, so the new binary's `daemon refresh-unit` child reaches `systemctl --user disable --now` while it is running inside that same legacy root user service cgroup. With the rendered `KillMode=mixed`, stopping the unit terminates the old main daemon and then SIGKILLs remaining cgroup processes, including the migration helper, before it can remove the old unit and install/enable the system replacement. This contradicts the handoff invariant documented in `supervisor/index.ts:41-49` and can leave the machine offline with the legacy service disabled and no replacement. Please separate in-daemon refresh from the out-of-service scope migration (or defer/fail it safely), and add coverage for the managed-update handoff rather than only standalone `installClientService()`.

❌ 2. Force UID 0's user-manager environment during root migration. [`apps/cli/src/core/supervisor/systemd.ts:53`]

`rootUserSystemctlEnv()` still prefers inherited `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS`. Under sudo/preserved environments those may identify the invoking non-root user's manager, so a root migration can probe or disable the wrong user's same-channel unit instead of `/run/user/0`. The legacy unit path likewise prefers inherited `XDG_CONFIG_HOME`. Root-scope migration must derive its user-manager bus and canonical legacy path from UID 0 rather than ambient caller variables; add a regression with conflicting inherited XDG/DBus values.

✅ The no-bus path now fails closed without deleting the legacy unit or writing the system unit.

✅ The paired Context Tree PR, agent-team-foundation/first-tree-context#739, records the normal-user/root scope split and exclusivity constraint.

**Action taken**
- Submitted follow-up request changes on aea1e59a.

yuezengwu (changes requested):
## Follow-up review on `aea1e59a`

The intended split is sound: UID 0 installs a system-scoped unit, normal users retain user-scoped units, and the latest follow-up now fails closed when the legacy root user manager cannot be reached. The operator hint and source documentation are scope-aware, and [first-tree-context#739](https://github.com/agent-team-foundation/first-tree-context/pull/739) captures the durable scope/exclusivity contract.

I am still requesting changes because two migration paths do not yet guarantee that exclusivity and availability are preserved.

### 1. Do not stop the containing legacy unit from the in-daemon update helper

`systemdBackend.refreshForUpdate` still points to `installSystemd` (`apps/cli/src/core/supervisor/systemd.ts:368`). During a managed update, `daemon refresh-unit` runs as a child inside the currently running legacy root user unit, then reaches `systemctl --user disable --now` at line 189. The generated unit uses `KillMode=mixed`; stopping it terminates the old main daemon and can then kill the remaining cgroup processes, including the refresh helper, before it removes the legacy file and installs/enables the system replacement.

That violates the explicit handoff invariant in `supervisor/index.ts:41-49` and can leave the host offline with the old service disabled and no replacement. Please separate the in-daemon refresh path from the out-of-service user-to-system migration (or defer/fail that migration without stopping the containing unit), and add a regression that exercises `refreshClientServiceUnitForUpdate()` from the legacy-root transition rather than only standalone installation.

### 2. Derive the legacy root user manager and unit path from UID 0

`rootUserSystemctlEnv()` (`systemd.ts:53-59`) still prefers inherited `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS`, and `systemdUserUnitPath()` prefers inherited `XDG_CONFIG_HOME`. With `sudo -E` or another preserved environment, those values may belong to the invoking non-root user. The root migration can therefore inspect or disable that user's manager/unit while missing root's actual legacy unit.

Please force the migration's user-manager bus and legacy unit path to UID 0's canonical locations instead of ambient caller values, and cover conflicting inherited XDG/D-Bus values in a regression test.

The new no-bus behavior is correct: it leaves the legacy unit in place and refuses system-scope enablement. The Context Tree content is also aligned. As a process follow-up, please keep #739 draft until the code merges and add the reciprocal link to this PR description.

Local verification: 82 affected tests passed, `pnpm --filter first-tree-dev typecheck` passed, and Biome passed on the changed files.

bestony:
Third follow-up pushed in a16cb5a5.

Addressed the latest review items:

1. `daemon refresh-unit` no longer performs root user-to-system scope migration from inside the legacy user unit. If a legacy root user unit exists, the in-daemon refresh path now fails before invoking systemctl, leaving the legacy unit and system unit untouched so the out-of-service install/login path can establish exclusivity.

2. Root migration no longer inherits the caller XDG/D-Bus target. The legacy root user unit path is derived from UID 0 user info, and the migration systemctl environment forces `XDG_RUNTIME_DIR=/run/user/0` plus `DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/0/bus` even under conflicting sudo-style env values.

3. PR body now reciprocally links draft Context Tree PR agent-team-foundation/first-tree-context#739, and #739 has been converted to draft until this code PR lands.

Verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts` (77 tests passed)
- `pnpm exec biome check apps/cli/src/core/supervisor/systemd.ts apps/cli/src/core/supervisor/index.ts apps/cli/src/__tests__/service-install-core.test.ts`

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773 (follow-up on a16cb5a5)

- Rationale: The latest commit safely defers in-daemon migration and pins UID 0 paths, but the deferred legacy service becomes invisible to every root lifecycle command, reintroducing duplicate-runtime and destructive-state risks.

**Risk level**: A
- Semantic lift: none; already at the highest path level

**PR summary**
- Problem: Existing root user-scoped daemons must remain available and exclusive while moving new root installs to system scope.
- Follow-up approach: Refuse user-to-system migration inside `daemon refresh-unit`, force root's canonical unit/bus paths, and defer the actual migration to an external install/login.
- Impacted modules: systemd supervisor refresh/install split and service-install tests

**Review findings**

❌ 1. Keep lifecycle commands fail-safe while the legacy root user unit is intentionally deferred. [`apps/cli/src/core/supervisor/systemd.ts:127`]

After managed update, the new binary now leaves the legacy root user unit running, but every root lifecycle path still derives `scope === "system"`. With no system unit yet, `systemdState()` returns `not-installed`; `daemon start` then falls through to an inline runtime, `daemon stop` reports nothing to stop, restart cannot reach the legacy unit, and `logout --purge` can proceed without stopping the live service. This can recreate two daemons sharing one home/client identity or delete active runtime state. Please detect the canonical legacy root unit before returning `not-installed` and preserve a fail-closed/migration-required state (or safely control the legacy user manager) consistently across status, start, stop, restart, uninstall, drift checks, logout/client-switch consumers, and operator guidance until an external install completes migration. Add command-level regression coverage for the post-auto-update legacy state, not only the refresh helper itself.

❌ 2. Make the refresh path structurally incapable of running scope migration. [`apps/cli/src/core/supervisor/systemd.ts:282`]

`refreshSystemdForUpdate()` checks that the legacy file is absent and then calls `installSystemd()`, which independently re-enables migration. If a concurrent old installer creates the legacy unit between those operations, the refresh helper can still execute `disable --now` from inside its own cgroup. Pass an explicit no-migration mode into the installer (or use a separate refresh implementation) so this side effect is impossible in the refresh path rather than guarded by a TOCTOU precheck.

✅ Root migration now ignores inherited sudo XDG/D-Bus targets and uses UID 0's canonical home and `/run/user/0` bus.

✅ The source PR and draft Context Tree PR #739 now link each other, and the tree PR remains draft.

**Action taken**
- Submitted follow-up request changes on a16cb5a5.

yuezengwu (changes requested):
## Follow-up review on `a16cb5a5`

This follow-up resolves the two previous findings in the narrow paths it changes: the legacy-root refresh now exits before invoking `systemctl`, and the migration probe derives UID 0's home and `/run/user/0` bus instead of inherited sudo XDG/D-Bus values. The source PR also reciprocally links draft Context Tree PR #739.

I am still requesting changes because deliberately deferring the migration creates a supported transition state that the rest of the lifecycle API cannot see, and because the refresh path can still re-enter migration after its precheck.

### 1. Keep the deferred legacy root service visible to every lifecycle consumer

After a managed update from an old root user unit, the new `daemon refresh-unit` now fails safely and the parent still exits 75; the root user manager then restarts that legacy unit with the newly installed binary. From that point, however, `systemdState()` (`apps/cli/src/core/supervisor/systemd.ts:127-143`) always selects the desired root **system** scope and returns `not-installed` when `/etc/systemd/system/<unit>` is absent, even though the legacy user-scoped daemon is active.

That false state is unsafe across existing callers:

- `daemon start` treats `not-installed` as permission to fall through to an inline daemon (`commands/daemon/start.ts:205`), recreating the duplicate-runtime/client-identity failure this migration is meant to prevent.
- `daemon stop` and `daemon restart` cannot reach the live legacy unit because their backend operations still target system scope.
- `logout --purge` only stops a service reported as `active`; after the false `not-installed` result it can continue into credential and runtime-state deletion while the service-mode daemon remains live (`commands/logout.ts:193-220`).

Please model/detect the canonical legacy-root transition before returning `not-installed`, and make status, start, stop, restart, uninstall, drift/ensure-service, logout, and client-switch consumers either safely control that root user manager or fail closed with an actionable migration-required state. Add command-level coverage for the post-auto-update legacy state, especially duplicate prevention and destructive purge/switch gates; the new helper-only assertion does not exercise these callers.

### 2. Make the update refresh implementation structurally migration-free

`refreshSystemdForUpdate()` checks for the legacy file and then calls `installSystemd()` (`systemd.ts:282-285`), which independently calls `migrateLegacyRootUserUnitToSystemScope()` at line 230. A concurrent installer/file restoration between those operations can therefore put the refresh helper back into `disable --now` despite the intended no-migration contract.

Please pass an explicit no-migration mode into the shared writer or use a separate refresh implementation so the in-daemon path cannot execute legacy scope migration at all, rather than relying on a TOCTOU precheck.

The UID 0 path/bus fix and the draft/reciprocal Context Tree linkage are resolved. #739 should still incorporate the recoverable deferred-handoff invariant already requested there before it is readied after this code merges.

Local verification on `a16cb5a5`: 88 relevant tests passed (service install, daemon start, update glue, and refresh-unit), the CLI typecheck passed, and Biome passed on the changed files. The remaining findings are uncovered lifecycle semantics, not test failures.

bestony:
Fourth follow-up pushed in 902127b0.

Addressed the latest review items:

1. Deferred legacy root user units are no longer invisible to root lifecycle commands. If root/system scope has no system unit but the canonical legacy root user unit still exists, `getClientServiceStatus()` now returns an `unknown` migration-required state with the legacy unit path and `managerScope: "user"`. `start`, `stop`, `restart`, and `uninstall` refuse with the same migration-required reason, and drift detection returns true. This keeps `daemon start`, logout purge, client switch, and other consumers fail-closed instead of treating the machine as not installed.

2. `refreshForUpdate` is now structurally incapable of running the user-to-system migration. The install path is `migrateLegacyRootUserUnitToSystemScope()` plus `writeAndEnableSystemd()`, while the refresh path only calls `writeAndEnableSystemd()` after the explicit legacy-unit guard. Even if a legacy unit appears after the guard, refresh has no call path to `disable --now` the user unit.

3. Added regression coverage for the deferred post-auto-update state across status, drift, start/stop/restart, uninstall, refresh, and `daemon start` command behavior.

Verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts` (77 tests passed)
- `pnpm exec biome check apps/cli/src/core/supervisor/systemd.ts apps/cli/src/core/supervisor/index.ts apps/cli/src/__tests__/service-install-core.test.ts apps/cli/src/__tests__/daemon-start-command.test.ts`

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773 (follow-up on 902127b0)

- Rationale: Deferred-service detection and structural no-migration refresh are now in place, but two user-facing consumers still bypass the known migration-required state, and the paired durable contract has not yet recorded the handoff state.

**Risk level**: A
- Semantic lift: none; already at the highest path level

**PR summary**
- Problem: A legacy root user daemon that survives managed update must remain visible and safe until an external install migrates it to system scope.
- Follow-up approach: Report a legacy root unit as `unknown`/migration-required, block backend lifecycle operations, and split refresh writing from migration.
- Impacted modules: systemd supervisor state/lifecycle and daemon-start/service tests

**Review findings**

❌ 1. Do not let ordinary logout clear credentials while the deferred legacy daemon may still be live. [`apps/cli/src/commands/logout.ts:194`]

The new status correctly returns `unknown`, but `runLogout()` only refuses that state when `opts.purge` is true. Plain `logout` skips the service stop and proceeds to remove credentials while the legacy service-mode daemon can remain connected/running. This violates the logout lifecycle contract and leaves a predictable active-runtime/auth mismatch. Treat the structured legacy migration-required state as blocking for every logout mode (or safely migrate/stop it first), and add command-level coverage for both plain logout and purge.

❌ 2. Do not recommend `--foreground` for the known legacy migration state. [`apps/cli/src/commands/daemon/start.ts:193`]

`daemon start` now refuses automatic startup, but its generic unknown-state guidance still tells the operator to pass `--foreground` to bypass the guard. In this transition the legacy user daemon may already be active, so following that instruction recreates the duplicate runtime/client identity. Give this condition a structured discriminator instead of relying only on detail text, and print the actionable out-of-service migration command without the unsafe foreground suggestion; cover the exact output in the new command test.

❌ 3. Record the recoverable deferred-handoff invariant in the paired tree PR. [first-tree-context#739]

The draft tree PR still only records scope selection and the requirement to prove the legacy service stopped before enabling system scope. It does not yet state that managed in-daemon updates defer migration, keep the legacy unit authoritative/visible, and require lifecycle operations to fail closed until an external install performs the cutover. Please update #739 before this PR is ready for approval.

✅ `writeAndEnableSystemd()` is now structurally migration-free, so the refresh path cannot re-enter `disable --now` after its guard.

✅ Root lifecycle status and backend start/stop/restart/uninstall/drift now detect the canonical deferred legacy unit instead of reporting `not-installed`.

**Action taken**
- Submitted follow-up request changes on 902127b0.

yuezengwu (changes requested):
## Follow-up review on `902127b0`

This follow-up fixes the two previous implementation blockers: a deferred canonical legacy root unit is now surfaced as `unknown`/migration-required instead of `not-installed`, backend lifecycle operations fail closed, and `refreshForUpdate` calls a structurally migration-free writer. The new status and daemon-start tests cover the core duplicate-prevention branch.

I am still requesting changes for two consumers that bypass or misdirect the known transition state, plus the paired durable contract that remains stale.

### 1. Block plain logout while the deferred legacy service may still be live

`getClientServiceStatus()` now correctly returns `unknown` for this state, but `runLogout()` only refuses `unknown` when `opts.purge` is true (`apps/cli/src/commands/logout.ts:193-206`). Plain `logout` skips `stopClientService()` and proceeds to remove credentials while the legacy service-mode daemon can remain active. That breaks the documented "stop the daemon and clear credentials" lifecycle and leaves a predictable live-runtime/auth mismatch.

Please treat the structured legacy migration-required condition as blocking for every logout mode (or safely stop/migrate it first), and add command-level coverage that preserves credentials/state for both plain logout and purge until the service state is safe.

### 2. Do not recommend `--foreground` for this migration-required state

`daemon start` now refuses its automatic path, but the generic `unknown` branch still prints "pass `--foreground` to bypass" (`apps/cli/src/commands/daemon/start.ts:193-203`). In the deferred state the legacy user daemon may already be active, so following that guidance recreates the duplicate runtime/client-identity risk the new state is meant to prevent.

Expose a structured migration-required discriminator/helper rather than making consumers infer it from `detail`, then print an actionable out-of-service migration command without the unsafe foreground suggestion. Extend the new command test to assert the exact safe recovery guidance.

### 3. Update the paired durable handoff contract

Draft [first-tree-context#739](https://github.com/agent-team-foundation/first-tree-context/pull/739) still records only root system scope and the requirement to prove the legacy service stopped before replacement. It does not yet record the now-essential contract that in-daemon refresh defers migration, preserves the legacy service as the recoverable authority, and keeps lifecycle/destructive operations fail closed until an external install performs the cutover. Please update the draft before this source PR is ready; it can still be reconciled and marked ready only after the code merges.

Local verification on `902127b0`: the service-install, daemon-start, update-glue, refresh-unit, logout/lifecycle, and targeted client-switch unknown-state tests passed. The broader six-file run was 127/128; the only failure was the already-documented environment-specific client-switch drain timeout occurring before its unrelated EXDEV assertion. CLI typecheck and Biome passed.

bestony:
Fifth follow-up pushed in faf56c95; Context Tree draft PR #739 also updated in agent-team-foundation/first-tree-context@563cb02.

Addressed the latest review items:

1. Added a structured `migrationRequired: "root-systemd-user-to-system"` discriminator on the known deferred legacy root systemd state, instead of relying on detail text.

2. `daemon start` now treats that migration-required state specially: it refuses startup, prints the out-of-service migration command (`first-tree-dev login <code>`), and does not suggest `--foreground` for this known duplicate-runtime risk.

3. `logout` now refuses both plain logout and `logout --purge` while the root systemd migration is required, so credentials/local state are not removed while the legacy root daemon may still be live.

4. Context Tree #739 now records the deferred managed-update handoff invariant: refresh-unit does not cut over inside the containing legacy unit, the legacy root user unit remains the authoritative visible state until external install/login migrates it, and lifecycle operations fail closed while migration is required.

Verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/service-install-core.test.ts src/__tests__/daemon-start-command.test.ts src/__tests__/agent-lifecycle-commands-extra.test.ts` (104 tests passed)
- `pnpm exec biome check <changed cli files>`
- `first-tree-dev tree verify --tree-path /tmp/first-tree-context-hlzKKt`

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1773

- Rationale: The new structured guard fixes the automatic start and logout paths, but an explicit foreground start still bypasses the guard and can create the duplicate runtime this migration state is intended to prevent.

**Risk level**: A
- Path baseline: includes `apps/cli/**`.
- Semantic lift: none.

**PR summary**
- Problem: Root Linux installs need a system-scoped daemon without allowing an older root user unit and a replacement runtime to share the same channel home and client identity.
- Approach: Move root installs to system scope, preserve normal-user systemd behavior, defer managed-update cutover to an external install, surface a structured migration-required state, and fail closed across lifecycle/destructive operations.
- Impacted modules: CLI supervisor and daemon lifecycle commands, service/login/logout tests, operator documentation, and paired Context Tree PR #739.

**Review findings**

❌ 1. Explicit `daemon start --foreground` still bypasses the migration-required guard. `wantInline` is computed before `getClientServiceStatus()`, so passing `--foreground` skips the new discriminator branch entirely and proceeds to register/start a second runtime while the legacy root user daemon may still be active. This contradicts the updated #739 contract that lifecycle operations fail closed instead of starting a foreground duplicate. Keep the supervisor-child path inline so the deferred legacy unit remains recoverable, but probe and reject the explicit operator `--foreground` path when `migrationRequired === "root-systemd-user-to-system"`; add command coverage for the explicit flag, not only the no-argument branch. [`apps/cli/src/commands/daemon/start.ts:135`]

✅ 2. Plain logout and purge now both stop before credential/state deletion for the structured migration state.

✅ 3. Draft Context Tree PR #739 now records the deferred handoff, authoritative legacy-unit visibility, and fail-closed lifecycle contract.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Follow-up review on `faf56c95`

This follow-up correctly closes the prior plain-logout, recovery-guidance, and durable-contract findings. The structured `migrationRequired` signal is propagated through `ServiceInfo`, both logout modes stop before credential/state mutation, the default `daemon start` path prints the external `login <code>` cutover, and draft Context Tree PR #739 now records the recoverable deferred handoff.

One duplicate-runtime path still bypasses that guard.

### Block explicit operator `daemon start --foreground` during the migration-required state

In `apps/cli/src/commands/daemon/start.ts:135`, `wantInline` is computed as `options.foreground === true || isSupervisorChild`. The service-status/discriminator branch only runs under `!wantInline`, so an operator who explicitly passes `--foreground` never calls `getClientServiceStatus()` and proceeds directly to create/register a second runtime while the authoritative legacy root user daemon may still be active.

That contradicts the updated #739 contract that lifecycle operations fail closed rather than starting a foreground duplicate. Preserve the inline bypass for the actual supervisor child so the deferred legacy unit can restart, but for an explicit operator `--foreground` invocation probe the service state and reject only `migrationRequired === "root-systemd-user-to-system"` before starting the runtime. Add coverage for `runStart(["--foreground"])`; the current new test exercises only the no-argument branch. The final #739 sentence about foreground mode should also remain clear that this migration-required exception overrides the general debug behavior.

The logout behavior, structured discriminator, UID 0 targeting, migration-free refresh implementation, and updated draft tree content otherwise resolve my prior findings.

Local verification on `faf56c95`: 115 relevant CLI tests plus the targeted client-switch unknown-state test passed, as did CLI typecheck and Biome.

bestony:
Sixth follow-up pushed in f90d54b1; Context Tree draft PR #739 updated in agent-team-foundation/first-tree-context@175c679.

Addressed the latest review item:

- Explicit operator `daemon start --foreground` now probes service state before starting inline. It still preserves the supervisor-child inline path, but if the structured `migrationRequired: "root-systemd-user-to-system"` state is present, it refuses foreground startup and prints the out-of-service `first-tree-dev login <code>` migration command.
- Added command coverage for `runStart(["--foreground"])` in the migration-required state.
- Updated #739 to clarify that the general foreground/debug behavior does not override the known root user-to-system migration guard.

Verification:
- `pnpm --dir apps/cli exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/service-install-core.test.ts src/__tests__/agent-lifecycle-commands-extra.test.ts` (104 tests passed)
- `pnpm exec biome check <changed cli files>`
- `first-tree-dev tree verify --tree-path /tmp/first-tree-context-hlzKKt`

yuezengwu (approved):
## Review by baixiaohang-code-reviewer - PR #1773 (follow-up on f90d54b1)

**Recommendation**: approve
- Rationale: Root installs now use system-scoped systemd without allowing a legacy root user unit, managed-update helper, lifecycle command, logout, client switch, or explicit foreground start to create a second runtime. The final foreground bypass is closed while the supervisor-child handoff remains recoverable.

**Risk level**: A
- Path baseline: includes `apps/cli/**`.
- Semantic lift: none; already at the highest path level.

**PR summary**
- Goal: Run root-operated Linux daemons under the system manager while preserving normal-user systemd user units and maintaining one authoritative runtime throughout upgrades.
- Core approach: Select scope by UID, migrate legacy root user units only from an external install/login path, expose a structured deferred-migration state, fail closed across lifecycle and destructive consumers, and make operator guidance scope-aware.
- Impacted modules: CLI systemd supervisor and daemon lifecycle commands, regression tests, operator docs/QA guidance, and paired Context Tree PR #739.

**Review findings**

✅ 1. Explicit operator `daemon start --foreground` now probes service state and rejects `migrationRequired: "root-systemd-user-to-system"` before constructing `ClientRuntime`. Its output points to the external `login <code>` migration path and no longer offers a duplicate-runtime bypass.

✅ 2. The actual supervisor-child path remains inline and does not enter the new operator-foreground probe, so managed update can leave the legacy root user unit authoritative and recoverable instead of recursively managing or stopping its containing unit.

✅ 3. The earlier migration, UID 0 environment, lifecycle visibility, logout/client-switch, journal guidance, and structural no-migration refresh findings remain resolved.

✅ 4. Draft Context Tree PR agent-team-foundation/first-tree-context#739 is aligned at `175c679`: it records the deferred authoritative legacy state, fail-closed lifecycle behavior, and the fact that normal foreground/debug behavior cannot override this known migration guard. It should remain draft until this code PR lands, then be reconciled and marked ready.

**Core type / persistence note**
- `ServiceInfo` gains optional `managerScope` and `migrationRequired` discriminators. There is no database or schema migration.

**Verification**
- `pnpm --dir apps/cli exec vitest run src/__tests__/daemon-start-command.test.ts src/__tests__/service-install-core.test.ts src/__tests__/agent-lifecycle-commands-extra.test.ts src/__tests__/update-glue.test.ts src/__tests__/daemon-refresh-unit.test.ts` (115 tests passed)
- Targeted client-switch unknown-state regression passed.
- Biome check on all changed CLI source/test files
- `git diff --check origin/main...f90d54b1`
- GitHub Actions are green on `f90d54b1`.

**Action taken**
- Approved.

yuezengwu (approved):
## Re-review on unchanged head f90d54b1

**Recommendation**: approve

The requested re-review contains no new source commit: the head remains `f90d54b1999d7953d18ac893b9bc6fcb64fd1e58`, which I fully reviewed and approved in review #4697032092. Root Linux installs use system scope, legacy root user-unit handoff stays exclusive and fail-closed, explicit foreground cannot bypass the structured migration guard, and the supervisor-child recovery path remains intact.

All GitHub Actions remain green. The earlier local verification also passed 115 related tests, the targeted client-switch unknown-state case, CLI typecheck, Biome, and `git diff --check`.

The paired Context Tree content remains aligned, but PR agent-team-foundation/first-tree-context#739 was moved from Draft to Ready while this implementation PR is still open. I have separately asked for #739 to return to Draft until this code merges and the tree content is reconciled against the merged implementation. That sequencing issue does not reopen a source-code blocker here.

**Action taken**: approved the unchanged head again.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1773

**Recommendation**: approve
- Rationale: Root systemd scope, exclusive legacy-unit migration, managed-update deferral, lifecycle fail-closed behavior, scope-aware guidance, and the explicit foreground guard are now consistent and covered on the reviewed head `f90d54b1`.

**Risk level**: A
- Path baseline: includes `apps/cli/**`.
- Semantic lift: none.

**Review findings**

✅ Root installs use system scope while normal users retain systemd user units.

✅ Legacy root user-unit migration is external, UID-0-targeted, and fails closed when exclusivity cannot be proven.

✅ Deferred managed-update state remains visible and blocks lifecycle, destructive, and explicit foreground duplicate-runtime paths while preserving the supervisor-child recovery path.

✅ Operator docs, journal hints, regression coverage, and paired Context Tree PR #739 describe the same durable contract.

**Action taken**
- Approved `f90d54b1` with explicit authorization from baixiaohang.

PR #1780, opened 2026-07-15
fix(web): show full computer client versions

baixiaohang:
## Summary

- render the full server-reported First Tree version in Team computers
- widen the version column so staging release identities remain visible
- cover full staging prerelease versions in the Computers DOM test

## Testing

- `pnpm check`
- `pnpm exec turbo run typecheck --concurrency=2`
- `pnpm exec turbo run test --concurrency=1 -- --maxWorkers=2`
- `pnpm --dir packages/web exec vitest run src/pages/clients/__tests__/clients-page-cards-dom.test.tsx --maxWorkers=2`

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -1906,7 +1906,9 @@
   width: 5rem;
 }
 .team-computers-table th:nth-child(4) {
-  width: 6rem;
+  /* Keep the channel/build suffix visible for staging versions such as
+     `0.5.14-staging.841.1`; it is part of the release identity. */
+  width: 12rem;
 }
 .team-computers-table th:nth-child(5) {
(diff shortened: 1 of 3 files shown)


yuezengwu (approved):
The implementation matches the intended behavior: Team computers now preserves the full server-reported First Tree version in both the desktop version cell and the narrow-layout metadata, while the wider column accommodates the current staging release identity. The DOM regression covers both rendering paths.

Verified locally:

- `pnpm --dir packages/web exec vitest run src/pages/clients/__tests__/clients-page-cards-dom.test.tsx --maxWorkers=2`
- Biome on the three changed files (clean apart from one pre-existing warning outside this diff)

No blocking findings.

PR #1781, opened 2026-07-15
test(web): stabilize mobile density async wait

baixiaohang:
## Summary

- add a file-local macrotask-aware wait helper for the mobile density DOM tests
- apply it to all three Query-backed cases in the file, including Now, Chat, and markdown preview
- keep timer waits inside React `act` without changing the shared DOM harness

This addresses the intermittent `Loading work` / `Loading chats` failures seen in main CI run [29405912604](https://github.com/agent-team-foundation/first-tree/actions/runs/29405912604) and during review stress testing.

## Testing

- 50 fresh Vitest processes running `pnpm --dir packages/web exec vitest run src/pages/mobile/__tests__/density-dom.test.tsx --maxWorkers=1 --reporter=dot` (50/50 passed)
- `pnpm check`
- `pnpm exec turbo run typecheck --concurrency=2`
- `pnpm exec turbo run test --concurrency=1 -- --maxWorkers=2`

### DIFF
--- a/packages/web/src/pages/mobile/__tests__/density-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/density-dom.test.tsx
@@ -111,6 +111,28 @@ function renderWithClient(harness: DomHarness, element: ReactElement, path: stri
   );
 }
 
+/**
+ * TanStack Query may notify observers on a timer, while the shared DOM
+ * harness only flushes microtasks. Keep the macrotask-aware polling local to
+ * this Query-backed test file.
+ */
+async function waitForSettled(harness: DomHarness, assertion: () => void): Promise<void> {
+  let lastErr: unknown;
+  for (let i = 0; i < 40; i++) {
+    try {
+      assertion();
+      return;
+    } catch (err) {
+      lastErr = err;
+    }
+    await act(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 5));
+    });
+    await harness.flush();
+  }
+  throw lastErr;
+}
+
 describe("mobile density tiers", () => {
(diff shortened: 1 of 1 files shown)


yuezengwu (changes requested):
Blocker: this leaves the same React Query timer race in the preceding Chat case, so the PR's advertised targeted command is still flaky.

At head `984e2e07`, I ran the whole-file command in 50 fresh Vitest processes. 17 runs failed, all at `density-dom.test.tsx:187` with `AllUnreadWatchingNewLoading chats`. The changed markdown case passed 50 isolated runs. On `origin/main`, the same whole-file repetition failed in that Chat case 25/50 times, confirming that the timer race is not actually limited to the one CI-observed markdown case.

Please make the wait macrotask-aware for all Query-backed cases in this file (or use a file-local helper that interleaves a real timer with the harness `act` flush), then repeat the whole-file command. `docs-pages-dom.test.tsx` already has a file-local `waitForSettled` precedent, so this can stay scoped without changing the shared DOM harness.

yuezengwu (approved):
Re-reviewed at `81ddd86b`: the blocker is resolved. The file-local macrotask-aware helper now covers all three Query-backed cases while keeping timer-driven updates inside React `act` and leaving the shared harness unchanged.

Independent verification:
- whole-file Vitest command in 50 fresh processes: 50/50 passed
- visible targeted run: passed with no `act` warnings
- `pnpm check`: passed (only pre-existing unrelated diagnostics)
- `pnpm --dir packages/web typecheck`: passed
- `git diff --check`: passed

Approved without waiting for the remaining CI test job.

PR #1786, opened 2026-07-16
fix: scope Context review merge policy

baixiaohang:
## Summary

- Clarify that the Context Review workflow runs independently of repository merge-policy settings.
- Require stale-approval dismissal or equivalent current-head protection only when a repository uses App approval as a strict per-head merge gate.
- Keep repository ruleset ownership outside the review skill and lock the conditional contract with a static floor test.

## Rationale

The review workflow always binds its verdict to the inspected commit and reruns on `synchronize`. Repository owners may still choose merge rules under which an older approval remains merge-eligible while a new head is being reviewed. That choice weakens the merge gate but does not make Context Review unsupported, so the shipped skill must not describe stale dismissal as an unconditional runtime prerequisite.

This reconciles the shipped skill with the post-merge governance decision being recorded in draft Context Tree PR https://github.com/agent-team-foundation/first-tree-context/pull/733.

## Validation

- Focused Context Review floor: 3/3
- `pnpm validate:skill`: 6/6
- `git diff --check`

The expensive live/model skill eval was not run for this wording-only contract correction.

### DIFF
--- a/packages/skill-evals/src/suites/context-tree-review/__tests__/floor.test.ts
+++ b/packages/skill-evals/src/suites/context-tree-review/__tests__/floor.test.ts
@@ -41,6 +41,8 @@ describe("context-tree-review floor", () => {
     expect(skill).toContain("first-tree github context-review submit");
     expect(skill).toContain('--run "$CONTEXT_REVIEW_RUN_ID"');
     expect(skill).toContain("Never call local `gh api .../reviews`");
+    expect(skill).toContain("The review workflow does not depend on");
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the wording and static floor coverage against the current Context Reviewer contract.

The skill now cleanly separates workflow support from repository merge governance while preserving the stricter stale-approval requirement for repositories that use App approval as a per-head merge gate. The focused floor (3/3), skill validation (6/6), skill-evals typecheck, Biome check, and diff check all pass locally. No blocking findings.

PR #1788, opened 2026-07-16
fix: retry transient provider failures after visible output

baixiaohang:
## Summary

- allow explicit `provider_capacity` and `transient_transport` failures to remain retryable after user-visible output or tool effects
- keep Codex app-server retries bounded inside the active handler, preserving the retry chain, accepted message prefix, and pending-tail custody across attempts
- exhaust overloaded and short-wait capacity failures after two foreground retries, then persist the terminal runtime notice before the single inbox ACK
- add product regressions and a formal QA case for capacity retry, exhaustion, backoff fencing, and steered-tail settlement

## Context

Codex can report `Selected model is at capacity. Please try a different model.` after assistant output or completed file changes. First Tree previously classified the resulting turn as `unsafe_replay` and consumed it immediately. The retry policy now treats failure classification and replay safety as separate inputs: known transient provider failures can continue through a bounded handler-owned retry chain, while unknown custody and permanent failures still stop.

## Companion Context Tree change

- https://github.com/agent-team-foundation/first-tree-context/pull/755 (draft; merge the code PR first)

## Validation

- `pnpm --filter @first-tree/client test --pool=threads --poolOptions.threads.maxThreads=2 --poolOptions.threads.minThreads=1 --reporter=dot` — 148 files passed, 1577 tests passed, 3 skipped
- targeted Codex app-server suite — 50 tests passed
- `pnpm check` — passed with 3 pre-existing warning/info diagnostics
- `first-tree-staging tree verify` — passed for the companion Context Tree change
- shared/server/web/skill-evals suites passed before the final client-only review fixes

The formal provider/bridge QA case was added but not executed because no deterministic controlled provider failure environment was available. The unrelated CLI suite still has two reproducible baseline failures in `update-portable-install.test.ts` where expected fallback shims are absent; this PR does not modify CLI code.

### DIFF
--- a/packages/client/src/__tests__/claude-code-provider-error-result.test.ts
+++ b/packages/client/src/__tests__/claude-code-provider-error-result.test.ts
@@ -16,6 +16,9 @@ const mockState = vi.hoisted(() => ({
   promptAppend: "",
   configModel: "",
   setModelCalls: 0,
+  throwQueryOnCall: 0,
+  closeCalls: [] as number[],
+  lifecycleEvents: [] as string[],
 }));
 
 vi.mock("@anthropic-ai/claude-agent-sdk", () => {
@@ -35,6 +38,9 @@ vi.mock("@anthropic-ai/claude-agent-sdk", () => {
     query: (args: { prompt: AsyncIterable<{ message?: { content?: unknown } }> }) => {
       mockState.queryCalls += 1;
       const attempt = mockState.queryCalls;
+      if (attempt === mockState.throwQueryOnCall) {
+        throw new Error("config restart query construction failed");
+      }
       const messages = (mockState.messagesByAttempt[attempt - 1] ?? mockState.nextMessages).slice();
       void drainPrompt(args.prompt, attempt);
       return {
@@ -51,7 +57,10 @@ vi.mock("@anthropic-ai/claude-agent-sdk", () => {
             },
           };
         },
-        close: () => {},
+        close: () => {
+          mockState.closeCalls.push(attempt);
+          mockState.lifecycleEvents.push(`close:${attempt}`);
+        },
         setModel: async () => {
           mockState.setModelCalls += 1;
         },
@@ -84,6 +93,9 @@ beforeEach(() => {
   mockState.promptAppend = "";
   mockState.configModel = "";
   mockState.setModelCalls = 0;
+  mockState.throwQueryOnCall = 0;
+  mockState.closeCalls.length = 0;
+  mockState.lifecycleEvents.length = 0;
 });
 
 function buildCache() {
@@ -134,6 +146,12 @@ async function startSingleResultTurn() {
   const emitted: SessionEvent[] = [];
   const completed: Array<{ count: number; outcome: TurnOutcome }> = [];
   const logs: string[] = [];
+  const retryTurn = vi.fn<SessionContext["retryTurn"]>(() => {
+    mockState.lifecycleEvents.push("retry:entered");
+  });
+  const failSessionForRecovery = vi.fn<NonNullable<SessionContext["failSessionForRecovery"]>>(() => {
+    mockState.lifecycleEvents.push("fatal");
+  });
 
   const cache = buildCache();
   await cache.refresh(AGENT_ID);
@@ -156,6 +174,8 @@ async function startSingleResultTurn() {
     emitEvent: (e) => emitted.push(e),
     ...mockCtxPlumbing({ sendMessage }, "chat-claude-provider-error"),
     forwardResult,
+    retryTurn,
+    failSessionForRecovery,
     finishTurn: async (messages, outcome) => {
       completed.push({ count: Array.isArray(messages) ? messages.length : 1, outcome });
     },
@@ -173,7 +193,7 @@ async function startSingleResultTurn() {
     ctx,
   );
 
-  return { handler, cache, sendMessage, forwardResult, emitted, completed, logs };
+  return { handler, cache, sendMessage, forwardResult, emitted, completed, logs, retryTurn, failSessionForRecovery };
 }
 
 async function runSingleResultTurn() {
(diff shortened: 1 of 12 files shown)


yuezengwu (changes requested):
Implementation summary: this separates transient failure classification from replay safety, adds a bounded handler-owned retry chain to the Codex app-server path, fences pending tails during backoff, carries the accepted prefix across attempts, and routes exhaustion through the existing durable-notice-before-ACK settlement.

[P1] Honor the retry-policy delay before enabling user-visible retries in Claude. `decideProviderRetry()` now returns `retry` for `user_visible` capacity/transport failures across all providers, and this PR explicitly exercises that new path in `claude-code-provider-error-result.test.ts`. Both Claude retry branches (`packages/client/src/handlers/claude-code.ts`, around the structured-result retry at 1606-1636 and stream-error retry at 1800-1884) immediately call `respawnQuery()` without awaiting `decision.delayMs`. The emitted payload therefore advertises a future `nextRetryAt` (500 ms, then 1500 ms), while the actual attempts run back-to-back and can exhaust before a short capacity/transport outage has any chance to recover. Please add an abortable wait for `decision.delayMs` in both Claude branches and cover the timing, or keep the shared user-visible override scoped to handlers that honor the retry contract.

The Codex app-server chain itself looks internally consistent in the covered paths: it preserves custody during backoff, starts the retry before releasing the tail fence, and completes the accepted prefix once on success/exhaustion. Core-structure note for human review: there is no DB/schema change; the in-memory `CurrentTurn` state gains attempt/chain fields plus a handler-level retry fence/cancel hook.

Local validation: 11 relevant client test files passed (193 tests total), client typecheck passed, Biome passed, and `git diff --check` passed.

baixiaohang:
Addressed the Claude retry-delay blocker in commit `8f19c271`.

- Both the structured provider-result branch and stream-error branch now await `decision.delayMs` before `respawnQuery()`.
- The handler-owned wait is abortable; `suspend()` / `shutdown()` cancel it before closing the active query, and cancellation exits without ACKing the unsettled turn.
- Timing regressions assert that transient retries do not start before 500ms / 1500ms. Existing unknown-failure retry tests now advance the policy's 5s / 15s delays, and a suspend regression verifies that cancellation does not spawn another query.

Validation after the fix:

- Client: 148 test files passed, 1578 tests passed, 3 skipped.
- `pnpm check`: passed (only the three pre-existing unrelated diagnostics).
- `git diff --check`: passed.

The Context Tree draft does not need a content change: it already records bounded handler-owned foreground retry, and this commit makes the Claude handler honor that contract.

yuezengwu (changes requested):
The original review blocker is fixed: both Claude structured-result and stream-error retries now honor `decision.delayMs`, and `suspend()` / `shutdown()` cancel the pending wait without ACKing the unsettled turn. The four changed Claude suites pass (18 tests), as do client typecheck, Biome on the five changed files, and `git diff --check`.

One handler-owned lifecycle still needs to participate in that cancellation:

**[P1] Cancel the pending provider retry backoff before a Path B config restart** (`packages/client/src/handlers/claude-code.ts:1540`)

While the existing `consumeOutput()` is awaiting `waitForProviderRetry()` at lines 1651 or 1923, an injected message can enter `maybeSwitchConfig()`. Path B builds query #2 and starts a new `consumeOutput()` at lines 1541–1543, but it does not abort the old consumer's retry wait. When that timer expires, the old consumer proceeds to `respawnQuery()` and replaces query #2 with query #3. Query #2 and query #3 then have separate consumers, while the same `unclosedSdkInputs` prefix/tail can be replayed twice; `consumerDone` tracks the config-switch consumer even though `currentQuery` now points at the stale retry's replacement.

I reproduced this deterministically with fake timers:

1. Query #1 returns a transient structured 503 and schedules the 500 ms retry.
2. Bump the cached config version/prompt and inject a steered tail; Path B creates query #2.
3. Advance 500 ms; the stale retry consumer creates query #3 (actual query count 3, expected 2).

Please cancel/retire the old retry consumer as part of the config-restart ownership handoff (and cover this interleaving with a regression). Cancellation should make that old `consumeOutput()` return without ACKing; the newly started config-switch consumer remains responsible for the unsettled turn.

baixiaohang:
Addressed the config-restart ownership race in commit `22964141`.

- Path B now cancels a pending provider retry backoff before any async restart preparation, so the old consumer exits without ACKing or respawning before the new query/consumer takes ownership.
- Same-family Path A is skipped while retry backoff owns the unsettled turn, preventing `setModel()` from yielding across the old timer and falling back too late.
- Deterministic fake-timer regressions cover both direct Path B and same-family model-only config bumps: each creates exactly one replacement query, advancing the stale retry timer creates no third query, and the unsettled turn is not prematurely settled.

Final validation:

- Client: 148 test files passed, 1580 tests passed, 3 skipped.
- `pnpm check`: passed (only the three pre-existing unrelated diagnostics).
- `git diff --check`: passed.

yuezengwu (changes requested):
The successful config-restart interleaving from the previous review is fixed: Path B now retires the pending backoff consumer before the async restart work, and the same-family Path A is skipped while that backoff owns the turn. Both new timing regressions pass, as do the four affected Claude suites (20 tests), client typecheck, Biome on the two changed files, and `git diff --check`.

There is one failure-path custody gap in the new handoff:

**[P1] Preserve an owner when config restart fails after cancelling backoff** (`packages/client/src/handlers/claude-code.ts:1520`)

Path B aborts the old retry consumer before several fallible operations: `buildEnv`, async resource-skill materialization, briefing write, and `buildQuery`. If any of them throws, `pushInjectedMessage()` catches and only logs the `maybeSwitchConfig()` error at lines 1302–1306, then continues formatting/delivering the injected tail. The old consumer has already observed the aborted wait and returned without ACKing, while no replacement consumer exists. A `buildQuery()` throw is especially problematic because it has already replaced `inputController`, so the tail is pushed into an orphan controller with no query reading it. There is no remaining timer, retry/recovery signal, or terminal settlement to reclaim the entered prefix and tail.

I reproduced this deterministically:

1. Query #1 returns a transient structured 503 and enters the 500 ms backoff.
2. Bump config and inject `m2`, but make the SDK `query()` constructor throw for the Path B replacement.
3. The logs contain both `Auto-resume cancelled during provider retry backoff` and `maybeSwitchConfig errored: ...`.
4. After advancing the old timer, query count remains 2, `completed` remains empty, and `m2`'s delivery token has received neither `processingStarted` nor `retry` — the turn is stranded until an unrelated suspend/rebind.

Please make this ownership transfer failure-atomic, or route restart-preparation/build failure through an explicit recovery/fatal path that preserves the provider-entered prefix and retries the unentered tail. Add a regression for a failed replacement after the old backoff consumer has been cancelled.

baixiaohang:
Addressed the config-restart failure custody blocker in commit `c4da9999`.

- `buildQuery()` now constructs the replacement controller, abort controller, and query locally, and commits them to handler state only after the SDK constructor succeeds.
- Config-restart failure retires the old provider transport before returning the provider-entered prefix and unentered tail to recovery, then signals fatal session recovery. The transport retirement helper is synchronous, best-effort, and idempotent, and is shared with `suspend()`.
- The regression covers Q1 visible transient failure → backoff → config bump + injected m2 → Q2 constructor throw. It asserts Q1 closes exactly once before m2 retry, m1 recovery, and the fatal callback; no Q3 is created; m2 never enters the provider; and a later suspend does not repeat close or settlement.

Validation:

- Client: 148 test files passed, 1581 tests passed, 3 skipped.
- `pnpm check` passed (only three existing diagnostics in unchanged files).
- `git diff --check` passed.

yuezengwu (approved):
Approved at `c4da9999`.

The config-restart failure handoff is now custody-safe:

- `buildQuery()` stages the replacement controller/query locally and commits handler state only after successful construction.
- A failed config restart retires the old transport before returning the provider-entered prefix and unentered tail to recovery, then drops the failed session for fresh-handler recovery.
- The new regression covers the exact Q1 backoff → config bump + steered tail → Q2 constructor failure interleaving, including close/retry/fatal ordering, no Q3, no provider entry for the tail, and idempotent later suspend.

I also rechecked the prior delay and successful config-restart fixes. Eight focused Claude suites pass (38 tests), along with client typecheck, Biome on the two changed files, and `git diff --check`. No remaining blocking findings.

PR #1792, opened 2026-07-16
refactor: simplify GitLab personnel routing

baixiaohang:
## Summary

- align GitLab personnel wake delivery with GitHub's settled Inbox semantics: validate authority during ingress, enqueue ordinary `notify=true` work, then use generic at-least-once delivery
- reduce GitLab Stage 3 persistence to three provider-owned tables: connections, entity/chat mappings, and current identity bindings
- remove automatic-action and manual reviewer switches, immutable identity epochs, three audit tables, provider-specific queued-wake cancellation, and their Settings/API surfaces
- keep automatic GitLab version-aware reviewer compatibility, exact admin-managed username bindings, connection lifecycle fencing, and one-time bearer management

## Current behavior

- GitLab personnel routing is enabled by configuring the webhook URL; there is no second automatic-actions switch
- GitLab 15.3+ uses the `reviewers` field only; older declared versions may fall back to assignees; unknown versions prefer valid reviewer evidence and otherwise use assignee fallback
- bearer regeneration resets learned version/reviewer capability so the new bearer learns again
- ingress holds current connection, exact identity, active membership, human mirror, and eligible delegate authority through durable card and Inbox writes
- after commit, generic Inbox delivery owns the wake; a lifecycle transition blocks future routing but does not retract work already accepted into the queue
- source mapping, fetch/diff, and review execution remain out of scope

## Schema and migration

Migration `0080_happy_komodo` assumes the human-confirmed production state that no GitLab Stage 3 identity/audit data exists yet.

- retains `gitlab_connections`, `gitlab_entity_chat_mappings`, and `gitlab_identity_links`
- drops the automatic-actions, identity-transition, and skipped-target audit tables
- changes identity links to a current `active | suspended` binding model
- enforces one binding per `(connection, membership)` and one binding per `(connection, normalized username)`, including suspended bindings
- makes the identity connection required and cascade-owned by the current connection; mapping provenance remains cascade-owned by the identity link
- removes historical epoch/snapshot and lifecycle actor/time/reason fields
- adds the last observed GitLab version to the connection and removes manual automatic-action/assignee-mode fields
- rollback is additive only through a future migration; dropped empty Stage 3 structures are not reconstructed automatically

## Validation

- shared: 52 files / 608 tests
- server: 229 files / 2446 tests on the full frozen snapshot
- web: 183 files / 1469 tests
- final dead-seam regression: 4 files / 54 tests
- `pnpm typecheck`: 10/10 tasks
- `pnpm check`: passed with three pre-existing diagnostics
- migration check passed; second `db:generate` reported no schema changes
- code and Context diff checks passed; Context Tree verification passed

Formal/live GitLab QA remains intentionally unexecuted and should run after merge as the previously agreed unified acceptance pass.

Context Tree adoption: [agent-team-foundation/first-tree-context#757](https://github.com/agent-team-foundation/first-tree-context/pull/757) (Draft until this code PR merges).


yuezengwu (changes requested):
## What this PR changes

This refactor deliberately replaces the just-landed GitLab Stage 3 model: personnel cards now write ordinary `notify=true` Inbox work after ingress-time authority validation; identity becomes one current `active | suspended` binding per member/username; GitLab version evidence selects reviewer compatibility automatically; and the automatic-action/manual-mode/audit/cancellation surfaces are removed. The generated `0080_happy_komodo` migration drops the superseded tables and columns and makes identity state cascade-owned by the current connection.

## Blocking finding

1. **[P1] Reconfirming a binding can activate duplicate historical mappings and roll the whole recovery back.**

   [`reconfirmGitlabIdentityLink`](https://github.com/agent-team-foundation/first-tree/blob/a7eca9ba319f18a3e882c144796f1c43f0732bf8/packages/server/src/services/gitlab-identities.ts#L197-L200) changes **every** mapping owned by the link back to `active=true`. That set is not guaranteed to contain one row per entity. When a member changes delegate, [`resolveGitlabTargetChat`](https://github.com/agent-team-foundation/first-tree/blob/a7eca9ba319f18a3e882c144796f1c43f0732bf8/packages/server/src/services/gitlab-webhook.ts#L740-L829) deactivates the old delegate row and inserts a new row for the same `(connection, identityLink, project, entityType, iid)`, intentionally retaining the old row as inactive history. Member leave then makes the new row inactive too.

   After membership restoration, reconfirm attempts to activate both rows in one statement. The partial [`uq_gitlab_entity_identity_target`](https://github.com/agent-team-foundation/first-tree/blob/a7eca9ba319f18a3e882c144796f1c43f0732bf8/packages/server/src/db/schema/gitlab-entity-chat-mappings.ts#L48-L50) index permits only one active row for that key, so PostgreSQL raises `23505`; the transaction rolls back the link to `suspended`, and the admin cannot reconfirm it without deleting and recreating the binding.

   Please do not bulk-reactivate historical mappings. The narrow option is to leave them inactive and let the next authorized personnel event reuse/activate exactly the row matching the current human/delegate (the existing target resolver already does this), or otherwise select at most one current-authority row per entity deterministically. Add a real-PostgreSQL regression covering delegate A routing → change to delegate B and route the same entity → member leave/reactivation → reconfirm, then prove reconfirm succeeds and a later personnel event leaves exactly one current mapping active.

## Human schema review required

Migration `0080_happy_komodo` changes core identity/routing structures and destructively drops three audit tables plus the old connection/link columns. Before merge/deploy, a human owner should revalidate the stated production precondition that no Stage 3 identity/audit rows exist, approve the new connection-owned cascades and suspended-name reservation, and confirm the deployment/rollback window: old server code cannot use the schema after those columns are dropped, while rollback is explicitly additive-only through a future migration.

## Verification

- Web GitLab API/Settings tests: 2 files / 9 tests passed.
- `pnpm check`: passed with the three pre-existing diagnostics.
- `node scripts/check-migrations.mjs`: passed.
- A second `db:generate` in an isolated worktree reported no schema changes.
- `git diff --check`: passed.
- Local Server Vitest could not start because this machine has no working container runtime; no server test executed locally. GitHub's full Test and migration checks are green at this head.

Formal/live GitLab QA remains intentionally unexecuted and should still run as the documented unified acceptance pass after the lifecycle fix.

yuezengwu (approved):
## Incremental review of `fa310efe`

The PR's overall goal remains to align GitLab personnel routing with the generic Inbox contract: validate current authority during ingress, persist ordinary `notify=true` work for at-least-once delivery, reduce Stage 3 identity state to current `active | suspended` bindings, and derive reviewer compatibility from observed GitLab versions instead of manual switches. The schema migration correspondingly removes the superseded audit, mode, epoch, and provider-specific cancellation surfaces.

The previous lifecycle blocker is resolved cleanly. Reconfirmation now restores only the identity link's authority and leaves every historical entity/chat mapping inactive. The next authorized personnel event goes through the existing target resolver, which reuses or activates only the row matching the current human/delegate. The new PostgreSQL-backed regression covers delegate A routing, delegate B replacing A for the same entity, membership leave/reactivation, admin reconfirmation, and the next personnel event; it proves reconfirmation succeeds and exactly the current delegate mapping becomes active without violating `uq_gitlab_entity_identity_target`.

I found no remaining blocking code issue and am approving this head.

### Human schema review still required

Migration `0080_happy_komodo` still changes core identity/routing structures and destructively drops the old Stage 3 tables and columns. Approval of the code is not a substitute for a human owner revalidating the production precondition that no Stage 3 identity/audit rows exist, approving the new connection-owned cascades and suspended-name reservation, and confirming the deployment/rollback window: old server code is incompatible after the column drops, and rollback is additive-only through a future migration.

### Verification

- Incremental diff: 2 files, 89 additions and 4 deletions; `git diff --check` passed.
- `pnpm check`: passed with the same three pre-existing diagnostics.
- GitHub Migrations, Lint & Type Check, CodeQL, portable smoke, and naming checks are green on this head; the full Test job is still running, and this review intentionally does not wait on CI.
- The local PostgreSQL-backed server test could not be executed because Docker Desktop has no reachable engine on this machine.

Formal/live GitLab QA remains the documented post-merge unified acceptance pass.

PR #1794, opened 2026-07-16
fix(client): delta Claude token usage snapshots

baixiaohang:
## Why

Claude Agent SDK `ResultMessage.modelUsage` is cumulative within one SDK Query, but the Claude Code handler emitted every snapshot as if it were a per-turn delta. The server then sums all `token_usage` events, producing triangular overcounting for long-lived Claude sessions.

A reset-free production sample showed 19 snapshots whose final cumulative processed-token value was 232,609,767, while summing the emitted rows produced 2,846,195,722 (12.24x).

## What

- Track a per-model usage baseline for each concrete Claude SDK Query.
- Diff input, output, cache-read, and cache-creation counters before emitting `token_usage`.
- Keep cache-creation deltas folded into `inputTokens`, preserving the existing wire contract.
- Skip unchanged model entries in cumulative multi-model snapshots.
- Treat an in-Query counter rollback as a fresh counter defensively.
- Bind the baseline to the captured Query/consumer so config hot-switch consumers do not share accounting state.
- Add a two-turn, multi-model regression test that proves cumulative snapshots become per-turn deltas.

This fixes newly emitted usage events. It intentionally does not rewrite already-persisted historical snapshots.

## Verification

- `pnpm check` (passes; existing unrelated warnings/info remain)
- `pnpm typecheck`
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 src/__tests__/claude-code-inject-startup-race.test.ts` (10 tests passed)
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2` (full Client suite passed)

### DIFF
--- a/packages/client/src/__tests__/claude-code-inject-startup-race.test.ts
+++ b/packages/client/src/__tests__/claude-code-inject-startup-race.test.ts
@@ -488,6 +488,101 @@ describe("claude-code handler startup inject queue", () => {
     await handler.shutdown();
   });
 
+  it("emits per-turn token deltas from cumulative model usage snapshots", async () => {
+    const events: Array<Parameters<SessionContext["emitEvent"]>[0]> = [];
+    state.resultMessagesForInput = (turn) => [
+      {
+        type: "result",
+        subtype: "success",
+        result: `reply ${turn}`,
+        modelUsage:
+          turn === 1
+            ? {
+                "claude-sonnet-4-5": {
+                  inputTokens: 10,
+                  cacheCreationInputTokens: 4,
+                  cacheReadInputTokens: 100,
+                  outputTokens: 2,
+                },
+                "claude-haiku-4-5": {
+                  inputTokens: 3,
+                  cacheCreationInputTokens: 0,
+                  cacheReadInputTokens: 20,
+                  outputTokens: 1,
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed at `f15464ec`.

The goal is to stop triangular overcounting from Claude Agent SDK `modelUsage` snapshots that accumulate across turns in one SDK Query. The implementation correctly keeps a per-model baseline scoped to the captured Query/consumer, emits field-level deltas, preserves cache-creation folding into `inputTokens`, skips unchanged model entries, handles counter rollback defensively, and starts fresh accounting for replacement Queries. Scoping the map inside the consumer also keeps overlapping config-restart consumers from sharing or clearing one another's state.

I checked the start/resume, transient respawn, and config-restart Query lifecycles plus the multi-model, unchanged-entry, rollback, and replacement-Query cases. There are no DB/schema changes and no wire-contract change; the added `Map` is transient handler-local accounting state.

Local verification:
- `pnpm check` (passes; existing unrelated 2 warnings and 1 info remain)
- `pnpm typecheck`
- focused regression: 12/12 tests passed
- full Client suite: 1,584 passed, 3 skipped

Approved.

PR #1797, opened 2026-07-16
fix(web): truncate long agent names across people pickers

Gandy2025:
## Problem

Agent display names and `@handles` render with no truncation in any people-picker surface. Long names break the layout in three different ways:

- **Add-participant dropdown** (chat header + right sidebar): rows wrap freely — a CJK full-title name wraps to 2+ lines and the mono `@handle` breaks mid-token (`@codex-` / `assistant`).
- **Composer `@` popover** (chat view, ask takeover, new-chat body): rows force `white-space: nowrap`, so a long name makes the whole list scroll horizontally with no ellipsis.
- **New-chat `[+]` recipient picker**: nowrap rows and a panel with `minWidth: 280` but no max — one long name stretches the panel arbitrarily wide, past the viewport edge. The selected-recipient chip (`AgentToken`) had no width cap either.

Team-page delegate selector options and the conversations People-filter labels had the same unbounded-name problem.

## Fix (owner-approved: single-line everywhere)

- `MentionLabel` (the shared label used by every picker row): display name gets `min-w-0 truncate`; the conditional `@handle` gets `shrink-0 truncate` capped at 45% of the row — it keeps its natural width when short, ellipsizes when long, and never breaks mid-token.
- New shared `mentionOptionTitle()` helper: hover `title` is now `DisplayName (@name)`, collapsing to `@name` when the two are identical (same noise rule as `shouldShowHandle`), so a truncated row still discloses the full identity; adopted by the popover, add-participant, new-chat, and delegate-selector rows (title sits on the row control, matching command-palette).
- `AgentToken` recipient chip: single-line truncate capped at `min(var(--sp-60), 100%)`, plus `title`.
- New `--sp-90` token (360px) so the picker-panel width cap lives in one place: `.mention-popover`'s existing max-width now references it, and the add-participant (icon variant), new-chat `[+]`, delegate-selector, and conversations-filter panels adopt it inline.
- Removed the three row-level `white-space: nowrap` declarations — truncation now lives at the label level, so the popover no longer scrolls horizontally.
- `DelegateSelector` option labels, `FilterCheckbox` labels, and the conversations rail `FilterChip` (the persistent chip for a picked People filter, previously unbounded) get the same `min-w-0` + `truncate` + `title` treatment.

Already-correct precedents untouched and now matched: command-palette rows, right-sidebar participant rows, Team-table `DelegateChip`.

Two independent review passes ran on the diff; their findings (filter panel missing the width cap, delegate title format drift and span-level placement, unlinked 360 literals, untokenized chip cap, rail `FilterChip` gap) are folded in.

## Testing

- `pnpm --filter @first-tree/web typecheck` (tsc + lint:tokens) passes.
- `pnpm check` (biome) passes.
- `pnpm --filter @first-tree/web test`: 186 files / 1483 tests pass, including new unit tests for `mentionOptionTitle`.
- Live-verified on a branch dev server proxied to staging, with real org data: composer `@` popover (desktop and 390px mobile), add-participant dropdown, new-chat `[+]` picker, Team delegate selector, People filter — uniform single-line rows, ellipsis on overflow, no horizontal scroll, panels capped at 360px. The original repro row (a CJK full-title display name + long handle) now renders on one line with the handle intact.

### DIFF
--- a/packages/web/src/components/__tests__/mention-autocomplete.test.ts
+++ b/packages/web/src/components/__tests__/mention-autocomplete.test.ts
@@ -8,6 +8,7 @@ import {
   fieldOutOfScrollport,
   groupAndSortCandidates,
   type MentionCandidate,
+  mentionOptionTitle,
   portalPanelPlacement,
   rankCandidates,
 } from "../mention-autocomplete.js";
@@ -428,5 +429,29 @@ describe("portalPanelPlacement", () => {
 
   it("dismisses (null) when there isn't room above the field for even one row", () => {
     expect(portalPanelPlacement({ field: { top: 30, bottom: 70 }, port, viewportTop: 0 })).toBeNull();
+  });
+});
+
+describe("mentionOptionTitle", () => {
+  it("combines display name and handle when they differ", () => {
+    expect(mentionOptionTitle(cand({ agentId: "a1", name: "codex-assistant", displayName: "白晓航的个人助理" }))).toBe(
+      "白晓航的个人助理 (@codex-assistant)",
+    );
+  });
+
+  it("collapses to the handle alone when the display name matches it", () => {
+    expect(mentionOptionTitle(cand({ agentId: "a1", name: "ux-expert", displayName: "ux-expert" }))).toBe("@ux-expert");
+  });
+
+  it("falls back to the handle when there is no display name", () => {
+    expect(mentionOptionTitle(cand({ agentId: "a1", name: "ux-expert", displayName: null }))).toBe("@ux-expert");
+  });
+
+  it("falls back to the display name when there is no handle", () => {
+    expect(mentionOptionTitle(cand({ agentId: "a1", name: null, displayName: "Soft Deleted" }))).toBe("Soft Deleted");
+  });
+
+  it("returns undefined when neither field is set", () => {
+    expect(mentionOptionTitle(cand({ agentId: "a1", name: null, displayName: null }))).toBeUndefined();
   });
 });
--- a/packages/web/src/components/add-participant-dropdown.tsx
+++ b/packages/web/src/components/add-participant-dropdown.tsx
@@ -10,6 +10,7 @@ import {
   ambiguousDisplayNames,
   buildPickerSections,
   type MentionCandidate,
+  mentionOptionTitle,
 } from "./mention-autocomplete.js";
 
 /**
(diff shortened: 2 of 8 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1797

- Rationale: The shared truncation work is coherent, but the new-chat recipient picker can still render beyond the viewport when its `[+]` anchor follows existing recipient chips.

**Risk level**: C
- Semantic lift: none

**PR summary**
- Problem: Long agent display names and handles break or widen people-picker rows across chat, Team, and Workspace filter surfaces.
- Approach: Make identity labels and selected chips truncate consistently, expose full identity through titles, and cap people-picker panels with a shared width token.
- Impacted modules: shared mention components, add-participant and new-chat pickers, Team delegate selector, Workspace People filters, Web design tokens.

**Review findings**
❌ 1. The new-chat `[+]` picker is still not viewport-safe. The panel is absolutely positioned with `left: 0` relative to the inline trigger at line 1072, and that trigger can sit anywhere after one or more recipient chips. `maxWidth: var(--sp-90)` only caps the panel at 360px; it does not account for the trigger's horizontal offset. On a 390px viewport, any trigger starting past 30px can still push a full-width panel beyond the right edge, where the page-level `overflow-hidden` clips it. Please clamp or flip the panel against the viewport (and cover the case with existing short/long chips before `[+]`), rather than relying on a content-width cap alone. [R5 / `packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:1078`]

✅ 2. The shared `MentionLabel` shrink chain and the People-filter/rail-chip truncation are internally consistent, and `mentionOptionTitle` removes duplicate identity formatting across the affected rows.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Goal and core change

This PR makes long agent identities single-line across people-picker surfaces: shared mention labels now ellipsize, full identities move into a shared row title, selected chips and panels gain width caps, and Team/Workspace picker labels get the same shrink chain. The shared label/title work is coherent, and I found no database, schema, or core-data-structure change.

## Blocking finding

The new-chat `[+]` picker is still not viewport-safe. Its panel is `position: absolute; left: 0` relative to the inline trigger in `ParticipantChips` (`packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:1065-1078`). That trigger moves horizontally after every existing recipient chip. `maxWidth: var(--sp-90)` limits the panel to 360px, but it does not constrain the panel's right edge. On a 390px viewport, even one ordinary chip can put the trigger far enough right that the 280–360px panel extends beyond the viewport; the draft root's `overflow-hidden` then clips the inaccessible portion.

Please anchor the panel to a viewport-safe boundary or use collision-aware clamping/flipping/portal positioning. Cover the actual regression shape: open `[+]` after existing short and long recipient chips at 390px (and the narrowest supported viewport), and verify both panel edges remain reachable.

## QA coverage

This is a cross-surface UI change, so the repository's `AGENTS.md` requires a matching reusable case under `packages/qa/cases/` and a PR note that formal QA is warranted. I could not find an existing people-picker/long-name case. Please add one; the chip-offset viewport scenario above should be part of it because jsdom behavior tests cannot detect this geometry regression.

## Verification

- 98 focused Web tests across mention autocomplete, portal behavior, Team delegate selector, People filter, and NewChatDraft interactions — passed
- Current GitHub checks — passed

Requesting changes for the viewport-clipping blocker. I found no additional code blockers in the remaining changed surfaces.

Gandy2025:
Superseded by GitLab MR !4 after the repository migration: https://gitlab.staging.unispark.dev/first-tree/first-tree/merge_requests/4. The GitLab MR includes the viewport-safe New Chat picker fix and reusable cross-surface QA case requested in review.

PR #1801, opened 2026-07-16
fix(client): give the legacy AGENTS prompt fallback self-contained provenance

### LINKED ISSUE #1799: [Bug] Legacy AGENTS prompt fallback lacks self-contained provenance
## Summary

The generated `AGENTS.md` clearly separates team prompts, editable agent prompts, agent prompt overrides, and First Tree-managed sections when structured `prompt.sections` are present. The legacy `prompt.append` fallback instead renders `## Agent-Specific Prompt` without a matching provenance label or edit instructions, so the generated file is not self-contained on every supported path.

## Surface Area

- Client agent briefing generation
- Generated `AGENTS.md` / `CLAUDE.md`
- Prompt configuration provenance and edit guidance

## Environment

- version: `0.5.16-staging.878.1`
- runtime: `codex`
- operating system: `Darwin 25.2.0 arm64`

## Reproduction

1. Build an agent briefing with `payload.prompt.append` populated and `payload.prompt.sections` absent.
2. Inspect the generated `AGENTS.md`.
3. Compare the banner's documented sections with the rendered `## Agent-Specific Prompt` fallback.

## Expected Behavior

Every supported briefing path should distinguish, in self-contained prose:

- First Tree runtime-managed sections;
- team-shared prompt resources and where team admins edit them; and
- this agent's own editable prompt and the exact read/write entry points.

No prompt-derived content should appear under an undocumented or ambiguous heading.

## Actual Behavior

The generated banner documents `# Team Prompt`, `# Agent Prompt`, and `# Agent Prompt Overrides`, then says every other section is First Tree-managed. On the legacy fallback path, the prompt body appears under `## Agent-Specific Prompt`, which is not named in the banner and has no local source/edit explanation. A reader can therefore misclassify the agent's prompt as First Tree-managed or edit the assembled briefing instead of the prompt source.

## First Tree Context

- Chat ID: `877c29b5-2a63-4f7a-876b-16b9647a402b`
- Reported by (user): `liuchao-001`
- Agent: `liuchao-codex`

## Additional Context

- Generator: `packages/client/src/runtime/templates/agent-briefing.ejs`
- Existing structured-path regression coverage: `packages/client/src/__tests__/agent-briefing.test.ts`
- The current test explicitly preserves `## Agent-Specific Prompt` as the legacy fallback, but does not require that fallback to carry the same provenance/edit boundary as the structured path.

liuchao-001:
Closes #1799

## Problem

The generated `AGENTS.md` banner documents `# Team Prompt`, `# Agent Prompt`, and `# Agent Prompt Overrides`, then states every other section is First Tree-managed. On the legacy fallback path (`payload.prompt.append` populated, `payload.prompt.sections` absent — a server that predates per-section provenance), the prompt body rendered under a bare `## Agent-Specific Prompt` heading that the banner never names and that carries no source/edit explanation. A reader could misclassify the agent's prompt as First Tree-managed, or copy the assembled briefing back into the prompt source — the exact failure mode the provenance labels exist to rule out.

## Change

- **Template fallback section** (`agent-briefing.ejs`): renamed to `# Agent Prompt (legacy merged — may include team-shared content)`, opening with a *Source:* paragraph that names the merged-blob origin, states that team prompt resources and the agent's own editable fragment are indistinguishable inside it, gives both edit entry points (`agent config prompt show <agent> --raw` / `prompt set <agent>` for the agent fragment; Cloud → Org Settings → Resources for team prompts), and warns against copying the merged section back.
- **Banner section map**: on the legacy path the banner now documents the legacy merged heading (with the same read/write entry points and never-copy warning) instead of the structured trio, so the banner always describes the headings that actually render. The structured path is unchanged.
- **Write guard** (`agent-briefing-guard.ts`): added `# Agent Prompt (legacy merged…` to the CLI-tier briefing-heading heuristic. The heading exists only in the assembled briefing, and its body embeds team-shared content, so copying it back would freeze team prompts into agent config. The structured `# Agent Prompt (this agent only — editable)` heading stays unguarded on purpose — its body is the agent's own fragment. `--force` remains the escape hatch for legitimate quoting.

## Tests

- `agent-briefing.test.ts`: legacy fallback test now asserts the provenance-labelled heading, the source/edit-entry-point prose, the do-not-copy warning, and the banner swap (structured trio absent on the legacy path); structured-path test asserts the legacy heading never renders there.
- `codex-bootstrap.test.ts`: updated to the new fallback heading; pins that `## Agent-Specific Prompt` is gone.
- `agent-briefing-guard.test.ts`: new heading flagged as heuristic tier; structured editable heading explicitly not flagged.

`pnpm check`, `pnpm typecheck`, and full test suites for `@first-tree/shared`, `@first-tree/client`, and the CLI package pass locally. One pre-existing CLI test (`agent-shared-helpers-extra.test.ts` › "ignores the runtime session token value from the agent subprocess env") fails only when the running shell inherits `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE` (i.e. inside a live agent env) — it passes with that variable unset and is untouched by this diff; the test could additionally sanitize that variable, left out of scope here.

### DIFF
--- a/packages/client/src/runtime/templates/agent-briefing.ejs
+++ b/packages/client/src/runtime/templates/agent-briefing.ejs
@@ -6,9 +6,10 @@
   Where each section comes from, and where to edit it:<% if (legacyPrompt) { %>
     # Agent Prompt (legacy merged) → this agent's prompt configuration,
                       served as ONE merged blob by a First Tree server that
-                      predates per-section provenance; team prompt resources
-                      (Cloud → Org Settings → Resources, managed by team
-                      admins) and this agent's own editable prompt fragment
+                      predates per-section provenance; team prompt resources,
+                      binding-managed agent prompt overrides (both managed in
+                      Cloud → Org Settings → Resources, never via prompt
+                      set), and this agent's own editable prompt fragment
                       are indistinguishable inside it;
                       read (own fragment only):  <%- bin %> agent config prompt show <agent> --raw
                       write (own fragment only): <%- bin %> agent config prompt set <agent> -f <file>
@@ -63,9 +64,10 @@ prompt.*<% for (const row of agentPromptOverrideRows) { %>
 
 *Source: this agent's prompt configuration, served as one merged blob by a
 First Tree server that predates per-section provenance. Team prompt
-resources (managed by team admins in Cloud → Org Settings → Resources) and
-this agent's own editable prompt fragment are indistinguishable here. Read
-the editable fragment alone with
+resources (managed by team admins), binding-managed agent prompt overrides
+(NOT editable with `prompt set` — both managed in Cloud → Org Settings →
+Resources), and this agent's own editable prompt fragment are
+indistinguishable here. Read the editable fragment alone with
 `<%- bin %> agent config prompt show <%- agentId %> --raw`; replace it with
 `<%- bin %> agent config prompt set <%- agentId %>`. Do NOT copy any of this
 merged section into your per-agent prompt.*
--- a/packages/client/src/runtime/templates/agent-briefing.ejs
+++ b/packages/client/src/runtime/templates/agent-briefing.ejs
@@ -3,7 +3,16 @@
   every session start. NEVER copy this file, in whole or in part, back into
   any prompt configuration.
 
-  Where each section comes from, and where to edit it:
+  Where each section comes from, and where to edit it:<% if (legacyPrompt) { %>
+    # Agent Prompt (legacy merged) → this agent's prompt configuration,
+                      served as ONE merged blob by a First Tree server that
+                      predates per-section provenance; team prompt resources
+                      (Cloud → Org Settings → Resources, managed by team
+                      admins) and this agent's own editable prompt fragment
+                      are indistinguishable inside it;
+                      read (own fragment only):  <%- bin %> agent config prompt show <agent> --raw
+                      write (own fragment only): <%- bin %> agent config prompt set <agent> -f <file>
+                      NEVER copy the merged section back into prompt set<% } else { %>
     # Team Prompt   → team prompt resources (Cloud → Org Settings →
                       Resources); managed by team admins; read-only here
     # Agent Prompt  → this agent's own prompt fragment;
@@ -11,7 +20,7 @@
                       write: <%- bin %> agent config prompt set <agent> -f <file>
     # Agent Prompt Overrides → agent-specific resource bindings that
                       replace team prompts; managed in Cloud, NOT via
-                      prompt set
+                      prompt set<% } %>
   Every other section is First Tree Managed — injected by the runtime and
   not editable through any prompt configuration.
 ====================================================================== -->
@@ -50,7 +59,16 @@ prompt.*<% for (const row of agentPromptOverrideRows) { %>
 
 <%- row.body %><% } %><% } %><% if (legacyPrompt) { %>
 
-## Agent-Specific Prompt
+# Agent Prompt (legacy merged — may include team-shared content)
+
+*Source: this agent's prompt configuration, served as one merged blob by a
+First Tree server that predates per-section provenance. Team prompt
+resources (managed by team admins in Cloud → Org Settings → Resources) and
+this agent's own editable prompt fragment are indistinguishable here. Read
+the editable fragment alone with
+`<%- bin %> agent config prompt show <%- agentId %> --raw`; replace it with
+`<%- bin %> agent config prompt set <%- agentId %>`. Do NOT copy any of this
+merged section into your per-agent prompt.*
 
 <%- legacyPrompt %><% } %>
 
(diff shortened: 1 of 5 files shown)


packages/client/src/runtime/templates/agent-briefing.ejs
```diff
@@ -50,7 +59,16 @@ prompt.*<% for (const row of agentPromptOverrideRows) { %>
 
 <%- row.body %><% } %><% } %><% if (legacyPrompt) { %>
 
-## Agent-Specific Prompt
+# Agent Prompt (legacy merged — may include team-shared content)
+
+*Source: this agent's prompt configuration, served as one merged blob by a
+First Tree server that predates per-section provenance. Team prompt
+resources (managed by team admins in Cloud → Org Settings → Resources) and
+this agent's own editable prompt fragment are indistinguishable here. Read
```
yuezengwu: Blocker: the legacy blob can contain a third provenance class besides team prompts and the standalone editable fragment. `runtimePromptAppend()` includes every enabled prompt row, including inline replacements / agent prompt overrides; `runtimePromptSections()` deliberately marks those rows non-editable because `prompt set` cannot touch them. On a pre-sections server this prose therefore still leaves binding-managed content without a source/edit explanation and can send the reader to the wrong command. Please mention agent-specific resource-binding overrides here and in the banner, state that they are Cloud-managed (not via `prompt set`), and add a fallback assertion for that boundary.

yuezengwu (changes requested):
Goal and core change are sound: make the legacy `prompt.append` fallback self-describing, swap the banner map to the heading that actually renders, and guard that merged heading without blocking the structured editable heading. The implementation and tests are otherwise well scoped.

One provenance class is still missing from the legacy explanation, so I’m requesting a small correction before approval. The merged append can also contain binding-managed agent prompt overrides, which `prompt set` intentionally cannot edit. Please name that class and its Cloud-managed edit path in both the banner and local Source paragraph, and pin it in the fallback regression test.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1801

- Rationale: The legacy fallback is still not provenance-complete: its merged blob can contain binding-managed agent prompt overrides, but both explanations name only team prompts and the standalone editable fragment.

**Risk level**: A
- Semantic lift: none.

**PR summary**
- Problem: Agents on the legacy `prompt.append` path can mistake a merged effective prompt stack for their own editable fragment and write non-owned content back into agent config.
- Approach: Label the fallback as a legacy merged section, explain its provenance and edit boundaries, and add the new heading to the CLI heuristic guard while leaving the structured path and server hard guard unchanged.
- Impacted modules: client briefing template/tests and the shared briefing-heading guard/tests.

**Review findings**
❌ 1. The explanation omits binding-managed agent prompt overrides. `runtimePromptAppend` also renders agent-scoped prompt resources and inline replacements, while `prompt set` intentionally cannot edit those rows. Naming only team resources and the standalone editable fragment leaves the fallback's edit guidance incomplete and can still send an agent to the wrong write path. Please name this third provenance class and its Cloud-managed edit path in both the banner and local Source paragraph, then pin it in the fallback regression test. [`packages/client/src/runtime/templates/agent-briefing.ejs`:7]

**Action taken**
- Submitted request changes.

liuchao-001:
Addressed in d859d4c — both reviews asked for the same missing provenance class, and the commit covers both:

- **Banner (legacy map entry)** now names the three classes inside the merged blob: team prompt resources, **binding-managed agent prompt overrides**, and the agent's own editable fragment — with the override/team classes routed to their Cloud-managed edit path (Org Settings → Resources, never via `prompt set`).
- **Fallback Source paragraph** names the same third class with the explicit `NOT editable with `prompt set`` marker, mirroring the structured `# Agent Prompt Overrides` section's wording.
- **Regression test** pins both mentions: exactly two occurrences of `binding-managed agent prompt overrides` (banner + Source), each anchored to the Cloud edit path, plus the `NOT editable with `prompt set`` phrasing in the Source paragraph.

Verified: `pnpm check` / `pnpm typecheck` green, full `@first-tree/client` suite passes (1584 tests), rendered output eyeballed on the legacy path.

yuezengwu (approved):
Re-reviewed `d859d4c`. The legacy banner and local Source paragraph now explicitly name binding-managed agent prompt overrides, route them to the Cloud-managed edit path, and state that `prompt set` cannot edit them. The fallback regression test pins the added provenance boundary. The original blocker is resolved; no further findings.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1801

**Recommendation**: approve
- Rationale: The updated head makes every provenance class in the legacy merged prompt explicit, routes each class to the correct edit surface, and preserves the existing structured-path and two-tier guard boundaries.

**Risk level**: A
- Semantic lift: none.

**PR summary**
- Problem: On the legacy `prompt.append` path, an agent could mistake a merged effective prompt stack for its own editable fragment and freeze team- or binding-managed content into agent config.
- Approach: Give the fallback a provenance-labelled heading and self-contained source/edit guidance, then recognize that heading in the CLI heuristic guard without changing the structured prompt path or the server's generated-marker hard guard.
- Impacted modules: client briefing template/tests and shared briefing-heading guard/tests.

**Review findings**
✅ 1. The previous blocker is resolved: both the banner and local Source paragraph now name team prompts, binding-managed agent prompt overrides, and the standalone editable fragment, including the Cloud-managed path and the `prompt set` exclusion.
✅ 2. The regression test pins the added provenance boundary, while the structured path remains mutually exclusive with the legacy fallback.
✅ 3. No API, schema, database, resource-resolution, or guard-tier behavior changed.

**Action taken**
- Approved with explicit authorization from baixiaohang.

PR #1807, opened 2026-07-17

yuezengwu:
## What changed

## Safety

## Validation


baixiaohang:
## Review by baixiaohang-code-reviewer - PR #1807

**Recommendation**: comment (post-merge; no blocking issue found)
- Rationale: the sync preserves ancestry, the Account route migration keeps rolling-deploy compatibility, and the provider identity invariants are unchanged; the synchronization guard merged the PR before this review request was delivered.

**Risk level**: B-low
- Path baseline: includes `packages/server/**`, with no DB structure or core service change.
- Semantic lift: none; the Web change moves an existing account/provider flow without changing subscriptions, WebSocket handling, or the data layer.

**PR summary**
- Problem: personal profile and sign-in management lived outside the unified Settings information architecture, which made personal and team-scoped settings harder to distinguish.
- Approach: move the surface to Settings → Account, group Settings navigation by Personal vs Team scope, preserve the legacy `/user-settings` redirect and OAuth callback path during rolling deploys, and retain query-string feedback across the redirect.

**Review findings**
- ⚠️ Follow-up: reconcile the current-route documentation now that `/settings/account` has shipped. `docs/development/google-oauth.md` still tells operators to verify `/user-settings`, and the bound Context Tree's current Web Console route map does not yet list Settings → Account even though its account-authentication node says to reconcile system reality when the implementation lands. The compatibility redirect prevents a functional break, but leaving the old path as the documented verification anchor risks turning the migration shim into the long-lived source of truth.
- ✅ The base commit is the merge base and a direct parent of the GitLab head; the resulting GitHub merge commit has the same tree as that head, so no content was added or lost by the synchronization merge.
- ✅ Link/unlink callbacks continue to land safely across old and new SPA generations, and the new SPA preserves callback query parameters before displaying feedback.

**Action taken**
- Posted this post-merge comment. No approval review was possible because the PR had already merged.

PR #1812, opened 2026-07-17

yuezengwu:
## What changed

## Safety

## Validation


baixiaohang (commented):
## Review by baixiaohang-code-reviewer - PR #1812

**Recommendation**: request changes (post-merge follow-up required)
- Rationale: the synchronization guard merged the PR before this requested review ran, but the new Kimi handler has two lifecycle/replay correctness gaps that should be fixed before relying on the provider in production.

**Risk level**: A
- Semantic lift: none needed; the baseline already covers the highest risk level.

**PR summary**
- Problem: operators who already use Kimi Code cannot currently select it as a First Tree agent runtime or receive its chat/tool/usage lifecycle through the existing collaboration flow.
- Approach: add a bundled direct-SDK Kimi provider, host-local provider authentication, shared retry/delivery translation, capability/configuration surfaces, Context Tree I/O attribution, and a reusable QA case.
- Impacted modules: CLI packaging, Client runtime/handler lifecycle, shared schemas, Server Context Tree observability, Web runtime configuration, and QA cases.

**Review findings**

❌ 1. Separate main-agent completion/output from subagent events before settling the First Tree turn. [`packages/client/src/handlers/kimi-code.ts:389`]

`Session.onEvent` filters only by session id; Kimi's `Event` also carries `agentId`, and the pinned SDK explicitly emits `turn.started` / `turn.ended` for subagents in the same session. This callback currently feeds every event into one `TurnObservation`, so a subagent `turn.ended` can resolve `endedPromise` while the main turn is still running, subagent assistant deltas can be forwarded as the chat answer, and `activeTools` can collide because it is keyed only by `toolCallId`. Gate main lifecycle/output on `event.agentId === "main"`, while deliberately accounting for subagent tool effects in separately namespaced state so replay safety and I/O observability remain correct. Add an interleaved main/subagent event regression.

❌ 2. Do not implement a provider retry by submitting the same full prompt as a new Kimi turn. [`packages/client/src/handlers/kimi-code.ts:464`]

Every retry iteration calls `activeSession.prompt(prompt)` again. In the pinned SDK this is a fire-and-forget new-turn operation: each call records a new `turn.prompt` and appends another user message before launching the turn. Therefore the current rate-limit test's expected two prompt calls also means the real persisted conversation contains the user's input twice; later model/tool behavior runs against corrupted history. Use a Kimi-native retry/rollback primitive, or disable this outer replay until the SDK exposes one, and cover the persisted transcript/history with the real harness rather than only counting calls on `FakeSession`.

⚠️ 3. Keep the implemented proposal in the Context Tree rather than duplicating it in the source repository. [`proposals/2026-07-17-kimi-code-runtime-provider.md:1`]

The canonical runtime and authentication decisions already live in the Context Tree. The team's documentation-placement decision routes proposal/design-process material to `raw-context/proposals/` and promotes adopted truth into normal system nodes; retaining this 374-line proposal in the public source repo creates a second copy of rationale and validation claims that will drift.

**Action taken**
- Submitted this post-merge comment. The PR had already been merged automatically before the requested review began, so no blocking review state could prevent the merge.
- Static review only; no repository checks or tests were run.

PR #1818, opened 2026-07-17
feat(runtime): host-local model catalogs for Cursor and Kimi (Phase 1)

yuezengwu:
## Summary
- Discover Cursor/Kimi model ids from the bound computer (`agent models`, `~/.kimi-code/config.toml`) via a correlated daemon RPC (`provider-models:list` → `provider-models:result`).
- Agent Model picker for Cursor/Kimi uses that catalog, with required fallbacks: **DEFAULT** (empty → inherit local provider default) and **Custom…** free-form exact model id.
- Claude/Codex keep curated lists for now (Phase 2/3).

## Test plan
- [ ] Unit: `discover-models` parsers; ModelSection Cursor/Kimi custom-id fallback; server GET models endpoint
- [ ] With a connected computer running Cursor: open Agent → Runtime → Model; confirm catalog loads, DEFAULT clears model, Custom accepts an exact id
- [ ] With Kimi configured: confirm models from `~/.kimi-code/config.toml` appear; DEFAULT inherits `default_model`
- [ ] Offline / unbound computer: Model still allows DEFAULT + custom free-form without blocking save

Made with [Cursor](https://cursor.com)

PR #1820, opened 2026-07-17
feat(runtime): host-local model catalogs for Cursor and Kimi (Phase 1)

yuezengwu:
## Summary
- Discover Cursor/Kimi model ids from the bound computer (`agent models`, `~/.kimi-code/config.toml`) via a correlated daemon RPC (`provider-models:list` → `provider-models:result`).
- Exposes `GET /api/v1/clients/:clientId/providers/:provider/models` returning `ProviderModelCatalog` for the web to consume.
- **Web Model picker is intentionally out of scope** — owned by kiven-fte (DEFAULT + catalog + Custom UI).
- Claude/Codex discovery stays Phase 2/3; curated web lists unchanged until then.
- Replaces #1818; fixes CodeQL polynomial-regex findings and Biome/branch CI issues.

## Test plan
- [ ] Unit: `discover-models` parsers; server GET models endpoint
- [ ] Connected Cursor computer: API returns account model catalog
- [ ] Kimi host with `~/.kimi-code/config.toml`: API returns configured models + defaultModelId
- [ ] Offline computer: API returns 503
- [ ] CI: branch name, lint, CodeQL clean


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1820

- Rationale: the new synchronous daemon RPC is process-local in a multi-replica server, and Kimi discovery does not follow all configuration forms supported by Kimi Code.

**Risk level**: A
- Semantic lift: none (already highest)

**Review findings**
❌ 1. The request/reply path only works when the HTTP request lands on the same server process that owns the daemon WebSocket. `sendToClient` and `pendingClientReplies` are both process-local, while `clients.instance_id` and the notifier architecture explicitly support replicas. On another replica this returns a false 503 for a connected computer; a result received by the socket-owning replica also cannot resolve a waiter stored elsewhere. Route both command and correlated result across replicas (or use a durable shared rendezvous), and cover the cross-instance case. [`packages/server/src/api/clients.ts:115`, `packages/server/src/services/connection-manager.ts:242`]

❌ 2. Kimi discovery always reads `~/.kimi-code/config.toml`, ignoring the official `KIMI_CODE_HOME` relocation. A daemon can run Kimi correctly from `$KIMI_CODE_HOME/config.toml` while this endpoint reports the catalog unavailable or, worse, returns a stale default-home catalog. Derive the path from the same effective environment as the runtime and add a custom-home case. Kimi documents the relocation here: https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#config-file-location. [`packages/client/src/runtime/capabilities/discover-models.ts:188`]

❌ 3. The hand-written TOML scanner recognizes only quoted model table headers (`[models."..."]`) and double-quoted scalar values. Valid Kimi configurations may use bare aliases such as `[models.gemini-3-pro-preview]`; those currently produce an empty catalog even though Kimi loads them. Please parse the supported TOML shape without narrowing valid syntax (prefer the provider parser/library if available), and test bare aliases plus the effective `default_model`. Kimi documents model aliases and quoting rules here: https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#models. [`packages/client/src/runtime/capabilities/discover-models.ts:105-145`]

**Action taken**
- Submitted request changes on the replacement PR; the original PR #1818 is closed.

yuezengwu:
Addressed review findings in d988e2bf:

1. **Cross-replica RPC** — `notifyDaemonClientCommand` / `notifyDaemonClientCommandResult` PG NOTIFY channels; catalogs land in `clients.metadata.modelCatalogRpc[ref]` (NOTIFY stays small). Local `sendToClient` when the socket is here; otherwise fan-out when `status=connected` + `instanceId` is set. Result path stores then wakes waiters on any replica.

2. **`KIMI_CODE_HOME`** — `resolveKimiConfigPath` uses `$KIMI_CODE_HOME/config.toml` when set.

3. **Kimi TOML** — switched to `smol-toml`; covers bare aliases (`[models.gemini-3-pro-preview]`) and quoted keys; tests for both + `default_model`.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1820

- Rationale: the Kimi fixes are complete, but the new cross-replica rendezvous can still lose or misroute valid daemon results under normal concurrency and connection replacement.

**Risk level**: A
- Semantic lift: none (already highest)

**PR summary**
- Problem: operators need the model picker to reflect the Cursor/Kimi catalog on the computer that actually runs the agent, without moving provider credentials into First Tree Cloud.
- Approach: a correlated HTTP → Server → daemon RPC discovers the host catalog; this update adds PG NOTIFY fan-out, a PostgreSQL rendezvous for large results, KIMI_CODE_HOME support, and standards-compliant TOML parsing.
- Impacted modules: CLI daemon, Client connection/discovery, Server client WebSocket/API/notifier, shared wire schemas.

**Review findings**
❌ 1. `storeModelCatalogRpcResult` performs a SELECT followed by replacement of the entire `clients.metadata` object. Two concurrent catalog replies can both read the same map and the later UPDATE drops the other ref; a concurrent capabilities or update-attempt write can also be erased (or erase the RPC entry). This makes the new shared rendezvous non-atomic and can both time out a valid request and regress unrelated client state. Store the ref with an atomic JSONB operation or a dedicated keyed table, preserve sibling metadata writers, and cover concurrent refs plus a sibling metadata update. [R4 / `packages/server/src/services/provider-models-rpc.ts:54`]

❌ 2. The durable result is still usable only when the volatile result NOTIFY arrives. `notifyDaemonClientCommandResult` swallows publish failures, LISTEN/NOTIFY can be missed during a listener reconnect, and the HTTP route only awaits the process-local promise; on timeout it returns 503 without checking the catalog already stored in PostgreSQL. Add a race-safe database fallback (at minimum read the correlated result before treating the timeout as failure; polling or another durable wake strategy is also viable) and test a stored result with a lost wake. [R4 / `packages/server/src/api/clients.ts:142`]

❌ 3. Command fan-out is not scoped to the DB-authoritative `clients.instance_id`. Each replica sends whenever its process-local map still has an open socket for the client id. After a cross-replica reconnect/takeover, the previous replica can retain an open socket even though its heartbeat no longer matches the current instance, so both old and current daemons receive the same ref and whichever catalog arrives first wins. Carry the target instance through the command and accept/deliver only on that current instance (including the result path), with a replacement-connection case. [R4 / `packages/server/src/api/agent/ws-client.ts:302`]

✅ 4. The prior Kimi findings are resolved: discovery now honors `KIMI_CODE_HOME` and parses quoted/bare TOML model aliases with a real TOML parser.

**Action taken**
- Submitted request changes on head `d988e2bf`.

yuezengwu:
Addressed the second review round:

1. **Atomic rendezvous** — `storeModelCatalogRpcResult` now uses top-level `jsonb_set` on `modelCatalogRpc` with a nested `||` merge for the ref (no whole-metadata replace). `updateClientCapabilities` likewise uses `jsonb_set` for `capabilities`. Covered concurrent refs + sibling capabilities update.

2. **Lost-wake fallback** — on waiter timeout/failure the HTTP route reads the durable catalog from metadata before returning 503. Tested with a shortened timeout and no NOTIFY.

3. **Instance-scoped delivery** — commands carry `targetInstanceId` from `clients.instance_id`; only that replica delivers. Results are ignored unless this process’s `instanceId` matches the DB owner. Stale local socket after takeover is covered.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1820

- Rationale: the previous rendezvous fixes are substantially implemented, but connection ownership can still change between validation and persistence, and the new atomic map permanently accumulates large per-request results.

**Risk level**: A
- Semantic lift: none (already highest)

**PR summary**
- Problem: operators need Cursor/Kimi model choices to come from the computer actually running the provider, without moving provider credentials into First Tree Cloud.
- Approach: the web-facing GET endpoint performs a correlated Server → daemon RPC; commands are routed to the socket-owning replica, catalogs are discovered locally and persisted in PostgreSQL, and result wakes fan back to the HTTP-serving replica.
- Impacted modules: CLI daemon, Client discovery/connection handling, Server client API/WebSocket/notifier/services, shared wire schemas.

**Review findings**
❌ 1. The result-ownership guard is still a check-then-act race. The handler first SELECTs `clients.instance_id`, then calls `storeModelCatalogRpcResult`, whose UPDATE is guarded only by `clients.id`. A reconnect can move ownership to another replica between those statements; the old socket then persists its catalog and wakes the waiter even though it is no longer authoritative. Make ownership validation and persistence one atomic operation (for example, UPDATE with both client id and expected instance id and require a returned row), and also reject a socket that has already been replaced locally. Cover takeover between validation and the durable write. [R4 / `packages/server/src/api/agent/ws-client.ts:1801`, `packages/server/src/services/provider-models-rpc.ts:58`]

❌ 2. The atomic ref merge removed the previous physical age/count pruning. `MAX_AGE_MS` now only makes reads ignore old entries; every result still appends a new UUID key and no path deletes it. Legitimate repeated GETs therefore grow `clients.metadata` forever with catalogs that may already be too large for NOTIFY, and the WS handler also persists any schema-valid UUID result even when no waiter exists on that replica. Restore a physical bound/cleanup atomically (or move rendezvous rows to a keyed table with expiry), and cover enough sequential refs to prove stored data is actually bounded rather than merely unreadable. [R3 / `packages/server/src/services/provider-models-rpc.ts:19`, `packages/server/src/services/provider-models-rpc.ts:61`, `packages/server/src/api/agent/ws-client.ts:1815`]

✅ 3. The earlier non-atomic sibling overwrite, lost-wake false 503, and target-instance command fan-out findings are otherwise addressed in the new paths.

**Action taken**
- Submitted request changes on head `278be4fb`.

yuezengwu:
Closing in favor of the single GitLab delivery MR for this Phase 1 work:

https://gitlab.staging.unispark.dev/first-tree/first-tree/-/merge_requests/30

Owner confirmed one MR only; GitHub draft #1820 is retired.

PR #1822, opened 2026-07-17

yuezengwu:
## What changed

## Safety

## Validation


baixiaohang:
## Review by baixiaohang-code-reviewer - PR #1822

**Recommendation**: follow-up fix required
- Rationale: the PR was already merged when the review request arrived, but the new provider-neutral archive query can archive a chat while a cross-provider entity is still open.

**Risk level**: A
- Semantic lift: also changes the GitLab mapping schema and core SCM/chat routing.

**Review findings**

❌ 1. `sweepMapped` does not evaluate every mapping attached to an SCM-origin chat. The `scm_mappings` CTE includes GitHub rows only when `metadata.source = 'github'` and GitLab rows only when `metadata.source = 'gitlab'`, while both follow APIs allow either provider to be attached to an existing chat. A GitHub-origin chat with a terminal GitHub mapping plus an open GitLab mapping (or the inverse) therefore passes the `NOT EXISTS` check and is archived early. Please build the terminal-state set from both mapping tables for every GitHub/GitLab-origin chat, and add a mixed-provider regression case. [`packages/server/src/services/chat-archive.ts:89`]

**Action taken**
- Submitted this post-merge comment; GitHub merged the synchronization PR three seconds after it became ready for review, so an approve/request-changes review was no longer possible.

PR #1828, opened 2026-07-18

yuezengwu:
## What changed

## Safety

## Validation

### DIFF
--- a/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
+++ b/packages/web/src/pages/__tests__/page-ssr-smoke.test.tsx
@@ -1249,7 +1249,7 @@ describe("page SSR smoke coverage", () => {
       <ChatView agentId="agent-1" chatId="chat-1" narrow presentation="mobile" />,
       createClient(),
     );
-    expect(mobileRendered).toContain("Show chat options");
+    expect(mobileRendered).toContain("Show chat details");
     expect(mobileRendered).not.toContain("Participants");
   });
 
--- a/packages/web/src/pages/mobile/__tests__/card-actions-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/card-actions-dom.test.tsx
@@ -12,6 +12,9 @@ import { MobileNowPage } from "../now.js";
 
 const meChatMocks = vi.hoisted(() => ({
   listMeChats: vi.fn(),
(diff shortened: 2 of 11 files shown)


baixiaohang:
## Review by baixiaohang-code-reviewer - PR #1828

- Rationale: The mobile archive guard is evaluated from a stale client snapshot and is not enforced atomically when the engagement transition is written.

**Risk level**: B-low
- Semantic lift: changes the mobile list projection and React Query-backed data behavior -> B-low

**PR summary**
- Problem: Mobile users need bounded chat triage from Now/Chat and read-only access to followed GitHub entities without opening the desktop workspace.
- Approach: Adds long-press/context-menu chat actions, an Archived recovery view, attention/pin ordering, and a mobile chat-details sheet that reuses the existing GitHub entity projection.
- Impacted modules: mobile Now/Chat pages and projections; workspace chat details; GitHub entity section.

**Review findings**

❌ 1. Archive eligibility is not guarded at the mutation boundary. [`packages/web/src/pages/mobile/chat-actions-sheet.tsx:19`](https://github.com/agent-team-foundation/first-tree/blob/fc57bde0bf92a69c27ef3ae6f2c68c86132310e7/packages/web/src/pages/mobile/chat-actions-sheet.tsx#L19) derives `archiveBlocker` from the `MeChatRow` captured when the sheet opens, and line 53 later sends an unconditional `archived` transition. Realtime/query invalidation can refresh the list while `actionsRow` keeps the old object, so a new open request, failed managed agent, or active turn can appear while the sheet still permits Archive. There is also an unavoidable check/write race even if the sheet follows the latest cache. The server endpoint currently accepts every engagement transition and writes it directly, so the later archive can win and hide work that the mobile contract says must remain active. Please enforce an archive-if-settled precondition server-side in the same atomic operation (and reconcile/surface the conflict in the sheet); keeping the client check for immediate feedback is still useful. [R4]

✅ 2. Commit ancestry is safe: the PR head is exactly two commits ahead of the base, and the merge commit preserves both parents with a tree identical to the PR head.

✅ 3. The long-press behavior, non-touch alternatives, bounded action set, Archived recovery view, and read-only GitHub links align with the current mobile-surface product boundary.

**Action taken**
- Submitted this post-merge review comment; the synchronization guard merged the PR before the requested review arrived, so a request-changes review can no longer block it.

PR #1840, opened 2026-07-20

yuezengwu:
## What changed

## Safety

## Validation


baixiaohang (approved):
Reviewed exact head 2bae45b4. No blocking findings: the cross-provider archive-state projection, complete attention-line targeting, legacy route-only fail-safe, and shared revive path are consistent with the durable chat semantics. GitHub required checks passed and the synchronization merge preserved ancestry.

Issue #1886, opened 2026-07-20
[Design] Deliver GitHub webhook events with a first-class system sender identity


liuchao-001:
### Problem

GitHub webhook cards are currently persisted with a representative human as the underlying message `senderId`, even though the UI can render a synthetic GitHub sender.

That human authorship leaks into agent-visible message semantics and can change collaboration behavior. For example, Zhang San and one of my agents may be working together in a task chat and follow a PR. When the GitHub event is then delivered under my human identity, the agent can interpret it as me joining the conversation and pivot to collaborating with me instead of continuing with Zhang San.

The representative human is a routing owner, not the author of the automated event. Using that human as the sender conflates delivery context with authorship.

### Proposal

Introduce a first-class system-authored message identity/role for trusted automated events. GitHub webhook cards should be authored by that system identity (with GitHub as the rendered source), while routing ownership, wake targets, and the actual GitHub actor remain separate structured metadata.

### Desired semantics

- Agent-visible sender identity for a GitHub webhook card is `system`, not a representative human.
- The UI continues to render a clear GitHub/system attribution.
- The GitHub actor is retained as event provenance, not reused as the First Tree message sender.
- The human routing owner, if needed, is retained separately and does not imply authorship or participation.
- Structured mentions/wake targets continue to drive inbox delivery and session activation.
- System-authored events do not trigger agent heuristics that treat a human sender as entering or taking over the collaboration.
- Sender self-skip and silent trusted-delivery behavior remain correct without borrowing a human identity.

### Design question

Should trusted external events use a real system sender at the message protocol layer, rather than only applying a synthetic visual sender on top of a human-authored message?

### Context Tree conflict and required follow-up

This proposal intentionally conflicts with `system/cloud/automation/github-webhook-routing.md`, especially the settled semantics that set the final `senderId` to the routing plan's `humanAgentId` while using `systemSender: "github"` only for visual attribution. It also changes how the trusted system-delivery contract in `system/cloud/chat/messaging.md` is represented at the protocol layer.

The webhook-routing node currently has `decisionLocksCode: true`. If this design is accepted and implemented, both the GitHub routing decision and the generic messaging contract must be updated in the Context Tree in the same delivery cycle.

baixiaohang:
Resolved by #1891.

[...]

PR #1891, opened 2026-07-20
fix: attribute trusted SCM cards as system in agent prompts

baixiaohang:
## Summary

- move the existing conjunctive GitHub/GitLab trusted-message classifiers into `@first-tree/shared` so Web and Client use the same attribution gate
- render trusted SCM cards as `[From: GitHub|GitLab · type=system]` in both the current Agent input and preceding silent context
- carry the existing message `source` into optional preceding-context wire data, with safe fallback when older peers omit it
- add shared, Client, Server, and cross-surface QA coverage for trusted attribution and spoof resistance

## Scope and non-goals

This is the short-term agent-visible attribution mitigation discussed in #1886. It does **not** implement first-class system authorship.

The change intentionally leaves the following semantics unchanged:

- database schema and persisted data
- `message.senderId` and its human carrier behavior
- chat membership, ownership, and routing-owner selection
- mentions, wake targets, fan-out, and sender self-skip
- addressable participants and `chat ask` targets

A future first-class system-author model remains a separate design and migration decision.

## Verification

- `pnpm typecheck` passed across the monorepo
- targeted Shared, Client, Web, and Server suites passed (2 + 46 + 39 + 22 tests)
- affected-file Biome checks and `git diff --check` passed
- broader test run passed Web (1,546), Server (2,557), Client (1,621), Shared (645), and QA suites; only the untouched Context Tree audit skill-eval fixture suite timed out locally at existing 15s/25s thresholds, including on isolated retry
- full `pnpm check` remains blocked by pre-existing unrelated issues in `assistant-text.test.ts`, `workspace-migrations.ts`, and `packages/web/src/index.css`; affected files pass targeted checks

## QA

Formal cross-surface QA is warranted because this changes Agent prompt attribution for provider events. The existing GitHub and GitLab webhook QA cases now require checking both current-message and preceding-context headers.

Related to #1886.

### DIFF
--- a/packages/client/src/__tests__/agent-io.test.ts
+++ b/packages/client/src/__tests__/agent-io.test.ts
@@ -60,6 +60,41 @@ function reviewPacket() {
   };
 }
 
+const githubCard = {
+  type: "github_event",
+  reason: "subscribed",
+  event: "issues",
+  action: "opened",
+  kind: "opened",
+  repository: "acme/widgets",
+  sender: "octocat",
+  title: "Issue #42: Broken widget",
+  body: "Please investigate",
+  url: "https://github.com/acme/widgets/issues/42",
+  entity: {
+    type: "issue",
+    key: "acme/widgets#42",
+    url: "https://github.com/acme/widgets/issues/42",
+  },
+};
+
+const gitlabCard = {
+  type: "gitlab_event",
+  event: "issue",
+  action: "open",
+  kind: "opened",
+  project: "acme/widgets",
+  sender: "alice",
+  title: "Broken widget",
+  body: "Please investigate",
+  url: "https://gitlab.example/acme/widgets/-/issues/42",
+  entity: {
+    type: "issue",
+    key: "501:issue:42",
+    url: "https://gitlab.example/acme/widgets/-/issues/42",
+  },
+};
+
 afterEach(() => {
   setCliBinding({ binName: "first-tree", packageName: "first-tree" });
 });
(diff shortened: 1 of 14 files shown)


yuezengwu (approved):
Reviewed at `d918a221`.

The goal and core changes are clear: this centralizes the existing conjunctive GitHub/GitLab attribution gates in `@first-tree/shared`, applies them to the Client's current and preceding-context prompt headers, and carries `source` through the optional preceding-message wire shape with rolling fallback. The persisted human carrier remains intact, so membership, sender self-skip, routing ownership, mentions, fan-out, and wake selection are unchanged.

I specifically checked the spoof boundary: ordinary Web writes are restamped as `source=web`; ordinary Agent/API writes cannot retain `systemSender`; and Context Review metadata remains server-reserved. A missing, unknown, malformed, or mismatched signal therefore falls back to the participant sender instead of acquiring system attribution.

No database/schema or core persisted-data-structure change is present. No blocking findings.

Verified locally:
- Shared trusted-attribution tests: 2 passed
- Client agent-IO tests: 46 passed
- Web GitHub/GitLab card tests: 39 passed
- Server inbox + system-sender boundary tests: 24 passed
- Shared, Client, Web, and Server typechecks passed
- Web design-token lint, affected-file Biome, and `git diff --check` passed

Formal cross-surface QA remains warranted for the live provider → inbox → assembled Agent-input boundary, as the updated QA cases state.

PR #1892, opened 2026-07-20
fix(cli): retire stranded legacy github-scan launchd runner

### LINKED ISSUE #995: CLI upgrade strands legacy github-scan launchd runner: KeepAlive crash-loop + port squatting

bestony:
Closes #995

## Summary

- detect legacy `github-scan` launchd plists under the pre-multi-channel production home during daemon startup
- boot out each recorded launchd label in the current user's GUI domain, remove the plist, and prune only the empty runner directory
- keep cleanup macOS-only, idempotent, best-effort, and leave legacy config/log files untouched
- add focused tests for missing directories, multiple plists, label parsing/fallback, benign bootout failures, non-plist files, and non-macOS behavior

The cleanup runs on the first daemon start after upgrade, which is the restart path used by `upgrade`, and releases the stranded KeepAlive runner that can squat port 7878.

## Validation

- `pnpm --filter first-tree-dev exec vitest run src/__tests__/retire-github-scan-launchd.test.ts`
- `pnpm --filter first-tree-dev run typecheck`
- `git diff origin/main...HEAD --check`

The focused suite passes (8 tests). The broader daemon-start suite has six existing failures caused by its generated client-ID fixture differing from the expected `client_1234abcd`; unrelated daemon matrix and lifecycle tests pass.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -31,6 +31,7 @@
     "scripts/prune-codex-runtime-binary.mjs",
     "scripts/prune-claude-runtime-binary.mjs",
     "scripts/prune-claude-sdk-deps.mjs",
+    "scripts/retire-legacy-github-scan-launchd.mjs",
     "README.md",
     "LICENSE"
   ],
@@ -45,7 +46,7 @@
     "coverage": "pnpm run build && vitest run --coverage --passWithNoTests",
     "prepack": "rm -rf skills && cp -R ../../skills .",
     "postpack": "rm -rf skills",
-    "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs && node scripts/prune-claude-sdk-deps.mjs"
+    "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs && node scripts/prune-claude-sdk-deps.mjs && node scripts/retire-legacy-github-scan-launchd.mjs"
   },
   "license": "Apache-2.0",
   "repository": {
--- a/apps/cli/scripts/retire-legacy-github-scan-launchd.mjs
+++ b/apps/cli/scripts/retire-legacy-github-scan-launchd.mjs
@@ -1,0 +1,47 @@
+#!/usr/bin/env node
+import { spawnSync } from "node:child_process";
+// Run the legacy github-scan migration from the newly installed package.
+// npm executes postinstall from the package that was just installed, so this
+// is the adoption-boundary hook for an old CLI (X) installing this release (Y).
+// The script is deliberately best-effort: migration failure must never turn a
+// successful package install into a failed upgrade.
+import { existsSync } from "node:fs";
+import { dirname, join } from "node:path";
+
+function insideSourceMonorepo() {
+  let dir = process.cwd();
+  for (let i = 0; i < 6; i += 1) {
+    if (existsSync(join(dir, "pnpm-workspace.yaml")) && existsSync(join(dir, "apps", "cli"))) return true;
+    const parent = dirname(dir);
+    if (parent === dir) break;
+    dir = parent;
+  }
+  return false;
+}
+
+function main() {
+  if (insideSourceMonorepo()) return;
+  const cli = join(dirname(import.meta.dirname), "dist", "cli", "index.mjs");
+  if (!existsSync(cli)) return;
+
+  const result = spawnSync(process.execPath, [cli, "daemon", "ensure-service"], {
+    encoding: "utf-8",
+    stdio: ["ignore", "pipe", "pipe"],
+    timeout: 15_000,
+    env: {
+      ...process.env,
+      FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY: "1",
+      FIRST_TREE_SERVICE_MODE: "",
+    },
+  });
+  if (result.status !== 0) {
+    const output = [String(result.stderr ?? "").trim(), String(result.stdout ?? "").trim()].filter(Boolean).join(" | ");
+    console.warn(`[first-tree] legacy github-scan cleanup did not complete${output ? `: ${output}` : ""}`);
+  }
+}
+
+try {
+  main();
+} catch (err) {
+  console.warn(`[first-tree] legacy github-scan cleanup skipped: ${err instanceof Error ? err.message : String(err)}`);
+}
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -57,7 +57,6 @@ const coreMocks = vi.hoisted(() => ({
   readActiveRootClientId: vi.fn(),
   recordActiveClientOwner: vi.fn(),
   removeLocalAgent: vi.fn(),
-  retireLegacyGithubScanLaunchd: vi.fn(),
   resolveServerUrl: vi.fn(),
   restartClientService: vi.fn(),
   stopClientService: vi.fn(),
@@ -139,7 +138,6 @@ beforeEach(() => {
   coreMocks.listLiveClientRuntimeMarkers.mockReturnValue([]);
   coreMocks.readActiveClientIdFromIndex.mockReturnValue(null);
   coreMocks.readActiveRootClientId.mockReturnValue(null);
-  coreMocks.retireLegacyGithubScanLaunchd.mockReturnValue({ bootedOut: [], removedPlists: 0 });
   coreMocks.stopClientRuntimeProcess.mockResolvedValue({ ok: true });
   cliFetchMock.mockReset();
   promptMocks.confirm.mockResolvedValue(true);
@@ -1198,17 +1196,5 @@ describe("logout and upgrade commands", () => {
     coreMocks.restartClientService.mockReturnValueOnce({ ok: true });
     await runTopLevel(registerUpgradeCommand, ["upgrade"]);
     expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain("Service restarted");
-  });
-
-  it("retires legacy github-scan immediately after install without a current service", async () => {
-    const { registerUpgradeCommand } = await import("../commands/upgrade.js");
-    coreMocks.detectInstallMode.mockReturnValue("global");
-    coreMocks.fetchServerCommandVersion.mockResolvedValueOnce({ ok: true, version: "99.0.0" });
-    coreMocks.isServiceSupported.mockReturnValue(false);
-
-    await runTopLevel(registerUpgradeCommand, ["upgrade"]);
-
-    expect(coreMocks.retireLegacyGithubScanLaunchd).toHaveBeenCalledTimes(1);
-    expect(coreMocks.getClientServiceStatus).not.toHaveBeenCalled();
   });
 });
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -57,6 +57,7 @@ const coreMocks = vi.hoisted(() => ({
   readActiveRootClientId: vi.fn(),
   recordActiveClientOwner: vi.fn(),
   removeLocalAgent: vi.fn(),
+  retireLegacyGithubScanLaunchd: vi.fn(),
   resolveServerUrl: vi.fn(),
   restartClientService: vi.fn(),
   stopClientService: vi.fn(),
@@ -138,6 +139,7 @@ beforeEach(() => {
   coreMocks.listLiveClientRuntimeMarkers.mockReturnValue([]);
   coreMocks.readActiveClientIdFromIndex.mockReturnValue(null);
   coreMocks.readActiveRootClientId.mockReturnValue(null);
+  coreMocks.retireLegacyGithubScanLaunchd.mockReturnValue({ bootedOut: [], removedPlists: 0 });
   coreMocks.stopClientRuntimeProcess.mockResolvedValue({ ok: true });
   cliFetchMock.mockReset();
   promptMocks.confirm.mockResolvedValue(true);
@@ -1196,5 +1198,17 @@ describe("logout and upgrade commands", () => {
     coreMocks.restartClientService.mockReturnValueOnce({ ok: true });
     await runTopLevel(registerUpgradeCommand, ["upgrade"]);
     expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain("Service restarted");
+  });
+
+  it("retires legacy github-scan immediately after install without a current service", async () => {
+    const { registerUpgradeCommand } = await import("../commands/upgrade.js");
+    coreMocks.detectInstallMode.mockReturnValue("global");
+    coreMocks.fetchServerCommandVersion.mockResolvedValueOnce({ ok: true, version: "99.0.0" });
+    coreMocks.isServiceSupported.mockReturnValue(false);
+
+    await runTopLevel(registerUpgradeCommand, ["upgrade"]);
+
+    expect(coreMocks.retireLegacyGithubScanLaunchd).toHaveBeenCalledTimes(1);
+    expect(coreMocks.getClientServiceStatus).not.toHaveBeenCalled();
   });
 });
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -46,7 +46,6 @@ const coreMocks = vi.hoisted(() => ({
   printResults: vi.fn(),
   refreshClientServiceUnitForUpdate: vi.fn(),
   removeLocalAgent: vi.fn(),
-  retireLegacyGithubScanLaunchd: vi.fn(),
   restartClientService: vi.fn(),
   stopClientService: vi.fn(),
 }));
@@ -146,7 +145,6 @@ beforeEach(() => {
     unitPath: "/tmp/unit.plist",
     state: "active",
   });
-  coreMocks.retireLegacyGithubScanLaunchd.mockReturnValue({ bootedOut: [], removedPlists: 0 });
   clientMocks.SessionRegistry.mockImplementation(() => ({ load: vi.fn(() => new Map()) }));
   clientMocks.cleanWorkspaces.mockReturnValue([]);
   localAgentMocks.createSdk.mockReturnValue({
@@ -285,12 +283,10 @@ describe("daemon utility commands", () => {
     await runDaemon(["ensure-service"]);
     expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain("not supported");
 
-    coreMocks.retireLegacyGithubScanLaunchd.mockClear();
     coreMocks.loadCredentials.mockClear();
     process.env.FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY = "1";
     await runDaemon(["ensure-service"]);
     delete process.env.FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY;
-    expect(coreMocks.retireLegacyGithubScanLaunchd).toHaveBeenCalledTimes(1);
     expect(coreMocks.loadCredentials).not.toHaveBeenCalled();
 
     coreMocks.isServiceSupported.mockReturnValue(true);
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -46,6 +46,7 @@ const coreMocks = vi.hoisted(() => ({
   printResults: vi.fn(),
   refreshClientServiceUnitForUpdate: vi.fn(),
   removeLocalAgent: vi.fn(),
+  retireLegacyGithubScanLaunchd: vi.fn(),
   restartClientService: vi.fn(),
   stopClientService: vi.fn(),
 }));
@@ -145,6 +146,7 @@ beforeEach(() => {
     unitPath: "/tmp/unit.plist",
     state: "active",
   });
+  coreMocks.retireLegacyGithubScanLaunchd.mockReturnValue({ bootedOut: [], removedPlists: 0 });
   clientMocks.SessionRegistry.mockImplementation(() => ({ load: vi.fn(() => new Map()) }));
   clientMocks.cleanWorkspaces.mockReturnValue([]);
   localAgentMocks.createSdk.mockReturnValue({
@@ -163,6 +165,7 @@ afterEach(() => {
   process.stdout.write = originalStdoutWrite;
   setPlatform(originalPlatform);
   delete process.env.FIRST_TREE_CHAT_ID;
+  delete process.env.FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY;
 });
 
 describe("config commands", () => {
@@ -281,6 +284,14 @@ describe("daemon utility commands", () => {
     coreMocks.isServiceSupported.mockReturnValueOnce(false);
     await runDaemon(["ensure-service"]);
     expect(printLineMock.mock.calls.map((call) => String(call[0])).join("")).toContain("not supported");
+
+    coreMocks.retireLegacyGithubScanLaunchd.mockClear();
+    coreMocks.loadCredentials.mockClear();
+    process.env.FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY = "1";
+    await runDaemon(["ensure-service"]);
+    delete process.env.FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY;
+    expect(coreMocks.retireLegacyGithubScanLaunchd).toHaveBeenCalledTimes(1);
+    expect(coreMocks.loadCredentials).not.toHaveBeenCalled();
 
     coreMocks.isServiceSupported.mockReturnValue(true);
     coreMocks.loadCredentials.mockReturnValueOnce(null);
--- a/apps/cli/src/__tests__/cli-entry-and-tree-actions-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-entry-and-tree-actions-extra.test.ts
@@ -32,7 +32,7 @@ const outputMocks = vi.hoisted(() => ({
 }));
 
 const migrationMocks = vi.hoisted(() => ({
-  retireLegacyGithubScanLaunchd: vi.fn(),
+  runLegacyGithubScanMigration: vi.fn(),
 }));
 
 vi.mock("@first-tree/client", () => clientMocks);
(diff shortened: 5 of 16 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1892

- Rationale: The cleanup can both miss affected upgrade/first-run states and destroy its own retry path after a non-benign `launchctl` failure.

**Risk level**: A
- Semantic lift: none (already A; the change touches daemon startup and service lifecycle behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: On daemon startup, enumerate the legacy production-home launchd plists, boot out their recorded labels, remove the plists, and leave unrelated legacy config/log files intact.
- Impacted modules: `apps/cli/src/commands/daemon`, `apps/cli/src/core`, focused CLI tests

**Review findings**
❌ 1. Preserve the plist when `launchctl bootout` fails unexpectedly so cleanup remains retryable. `bootoutLabel` returns `void` after logging every non-benign failure, and the caller then unconditionally deletes the plist. A timeout, permission error, or transient launchd I/O failure can therefore leave the KeepAlive job loaded while erasing the only on-disk artifact the next daemon start would discover. Return a success/confirmed-absent outcome, remove the plist only for those outcomes, and cover a non-benign failure in the focused tests. [R4 / `apps/cli/src/core/retire-github-scan-launchd.ts:52`, `apps/cli/src/core/retire-github-scan-launchd.ts:113`]
❌ 2. Run retirement from a first-run/upgrade path that is reachable before current-client credentials and service existence gates. The new call is after `loadCredentials()` and only on the inline daemon path; meanwhile `upgrade` returns without starting anything when the current client service is not installed. A machine can still have the old standalone `github-scan` launchd job but no current Cloud daemon credentials/service, so both upgrade and `daemon start` leave the crash-loop in place and still require manual `launchctl` remediation. Move the one-shot sweep to an unconditional new-binary startup path before those gates (or invoke an equivalent post-install helper) and cover this state. [R5 / `apps/cli/src/commands/daemon/start.ts:101`, `apps/cli/src/commands/daemon/start.ts:268`; `apps/cli/src/commands/upgrade.ts:133`]
✅ 3. The macOS guard, disk-derived label, narrow legacy directory scope, and preservation of non-plist/config/log files are appropriately constrained.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review — request changes

Reviewed head `2491bb16c2f229b887581b100cd59cff4118e511`.

### Goal and core change

This PR retires the removed `github-scan` launchd runner by scanning its legacy production-home plist directory during daemon startup, booting out each recorded label, deleting the plist, and keeping the cleanup macOS-only and best-effort. The direction and narrow filesystem scope are appropriate, but three blockers remain.

### Blockers

1. **Keep the plist when `launchctl bootout` fails unexpectedly.** `bootoutLabel()` only logs a non-benign failure and returns `void`; the caller then unconditionally removes the plist. A timeout, permission error, or transient launchd I/O failure can therefore leave the KeepAlive job loaded while deleting the only artifact the next startup can discover. Return a success/confirmed-absent outcome, remove only for those outcomes, and add a focused non-benign-failure test. (`apps/cli/src/core/retire-github-scan-launchd.ts:52`, `:113`)

2. **Make cleanup reachable from the accepted upgrade/first-run states.** The call currently sits after the client-switch, credentials, service-state, config, runtime-marker, and agent-migration gates. In particular, `daemon start` exits before cleanup without current Cloud credentials, while `upgrade` returns without launching the new daemon when the current client service is not installed. A legacy `github-scan`-only machine can therefore upgrade and still receive no automatic remediation. Invoke the sweep from a new-binary path reachable without current-client setup (while preserving help/read-only behavior), and cover the no-credentials/no-current-service state. Also add `retireLegacyGithubScanLaunchd` to the daemon-start test mock and assert the integration: today the missing mock makes the call throw `TypeError`, which the production best-effort catch swallows, so the daemon suite does not exercise this wiring. (`apps/cli/src/commands/daemon/start.ts:76-268`, `apps/cli/src/commands/upgrade.ts:133-139`, `apps/cli/src/__tests__/daemon-start-command.test.ts:18-45`)

3. **Restore the required formatting check.** Biome reports PR-local `organizeImports` errors in `apps/cli/src/commands/daemon/start.ts` and `apps/cli/src/core/index.ts`; this is the cause of the current `Lint & Type Check` failure.

### Validation

- Focused retirement suite: 8/8 passed.
- Daemon-start suite: 31/31 passed, but currently has the integration blind spot described above.
- CLI typecheck: passed.
- Biome on the four changed files: failed on the two PR-local import/export ordering errors above.
- Diff check: passed.

No database or core data-structure change is present in this PR.

bestony:
Addressed all three blockers in commit `7e909c5b`:

- `launchctl bootout` now returns a confirmed success/absent outcome; unexpected failures leave the plist in place for the next retry, with focused coverage.
- daemon startup runs the sweep before switch, credential, and service gates; `upgrade` also runs it immediately after install so machines without a current service are covered.
- added the daemon-start mock/integration assertion and restored Biome import ordering.

Local validation: retirement suite 9/9, agent lifecycle suite 28/28, targeted daemon startup checks 3/3, CLI typecheck, targeted Biome, and diff check pass. The full daemon-start file still shows the existing generated-client-ID fixture mismatch in this local environment; its new no-credentials integration test passes.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1892

- Rationale: The retry and daemon-start fixes are sound, but the claimed no-current-service upgrade remediation still cannot run when users first upgrade into the release that contains this code.

**Risk level**: A
- Semantic lift: none (already A; this changes daemon startup and upgrade lifecycle behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: Retire legacy plists before daemon credentials/service gates, retain failed plists for retry, and also attempt cleanup after a manual upgrade installs new bits.
- Impacted modules: `apps/cli/src/commands/daemon`, `apps/cli/src/commands/upgrade.ts`, `apps/cli/src/core`, CLI lifecycle tests

**Review findings**
❌ 1. The post-install call does not cover the first upgrade into the release containing this migration. `upgrade` continues executing modules loaded from the pre-upgrade binary: if version X installs fixed version Y, X's in-memory `upgrade` implementation has no `retireLegacyGithubScanLaunchd` call at all. On a machine with no current client service, X then returns without launching Y, so the legacy job remains loaded. The new test only runs Y's command implementation while mocking an upgrade to `99.0.0`; it proves later upgrades, not the X -> Y adoption boundary this migration must repair. Route the migration through code that the newly installed binary actually executes on its first eligible run (or an installer/post-install mechanism that runs Y), and cover that version-boundary behavior. [R5 / `apps/cli/src/commands/upgrade.ts:107`, `apps/cli/src/commands/upgrade.ts:120`; `apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts:1203`]
✅ 2. Unexpected `launchctl` failures now retain the plist and retry state, while success/confirmed-absent outcomes remove it.
✅ 3. Daemon startup now invokes cleanup before switch/credential/service gates, and its test mock/assertion exercises the no-credentials wiring.
✅ 4. The PR-local import/export ordering issues are corrected in head `00535771`.

**Action taken**
- Submitted request changes on head `005357716d5781f57d76f5985464a8573f93d934`.

yuezengwu (changes requested):
## Successor-head review — request changes

Reviewed head `005357716d5781f57d76f5985464a8573f93d934`.

### Goal and core change

This PR adds a macOS-only migration for the retired `github-scan` launchd runner: discover legacy plists, boot out their labels, remove only confirmed-retired plists, and trigger the sweep before current daemon credentials/service gates. The successor commits correctly fix the retry semantics, daemon-start wiring/test blind spot, and PR-local formatting errors. One rollout blocker remains.

### Remaining blocker

**The new `upgrade` call cannot remediate the first upgrade into the release that contains it.** When installed version X runs `first-tree upgrade` and installs fixed version Y, the process continues executing X's already-loaded `upgrade` module. X has neither the new `retireLegacyGithubScanLaunchd` import nor the call after `installGlobalSpec` / `installPortableSpec`. If there is no current client service, X then follows its existing `not-installed`/unsupported-service return path and never launches Y, leaving the legacy job active.

The new lifecycle test loads Y's `registerUpgradeCommand` and mocks an install to `99.0.0`; it proves that Y will clean during a later Y -> Z upgrade, not that X -> Y runs this migration. Route adoption through code the newly installed artifact actually executes (an installer/post-install mechanism for Y, or a clearly defined first-eligible-run path in Y) and cover that version boundary. The early `daemon start` hook is sound once Y is actually invoked, but a service-less upgrade does not invoke it automatically. (`apps/cli/src/commands/upgrade.ts:107-131`, `apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts:1200-1213`)

### Verified fixes and validation

- Non-benign `launchctl` failures retain the plist for retry; focused suite 9/9 passed.
- Cleanup runs before daemon switch/credential/service gates; daemon-start suite 31/31 passed and now asserts the no-credentials path.
- Agent lifecycle suite 28/28 passed.
- CLI typecheck, full Biome check, targeted Biome, and diff check passed (full Biome reports only existing warnings/info outside this PR).

No database or core data-structure change is present.

bestony:
Addressed the successor-head rollout blocker in `60791d0c`.

The migration no longer relies on the old binary's in-memory `upgrade` module:

- the published npm package now ships a postinstall script that invokes the freshly installed CLI's `daemon ensure-service` in cleanup-only mode, so X -> Y runs Y's migration even with no credentials/service;
- portable installation already invokes the new `daemon ensure-service`, which now runs the sweep before its service/credential gates;
- the existing daemon-start pre-gate sweep remains for first eligible daemon runs;
- added tests for the postinstall payload and cleanup-only installer path.

Local validation: 46 focused/lifecycle tests, CLI typecheck, Biome over all touched files, and diff check pass.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1892

- Rationale: The npm adoption boundary is now covered, but portable self-update still switches to Y without executing Y's migration when no service restart follows.

**Risk level**: A
- Semantic lift: none (already A; this changes package installation, portable distribution, and daemon lifecycle behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: Keep an idempotent cleanup in the new daemon, invoke the freshly installed npm artifact from postinstall, and run the same cleanup through portable install recovery's hidden `daemon ensure-service` command.
- Impacted modules: `apps/cli` package lifecycle, portable installer integration, daemon commands, cleanup core, focused tests

**Review findings**
❌ 1. Cover portable **self-update**, not only the standalone portable installer. A portable X running `first-tree upgrade` calls X's already-loaded `installPortableSpec()`, which extracts Y, switches `current`, rewrites shims, and returns without executing `scripts/portable/install.sh` or Y's `daemon ensure-service`. The old `upgrade` process then returns for `--no-restart`, unsupported service control, or `state === "not-installed"`, so the legacy runner remains active. The new postinstall test covers npm, and the command-matrix assertion covers `ensure-service` once something invokes it, but neither exercises this public portable X -> Y path. Add a first-eligible-run migration bootstrap in Y that does not depend on daemon/service setup (preferably through the centralized upgrade-migration insertion point), and add a version-boundary test proving a service-less portable X -> Y eventually executes Y's migration. [R5 / `apps/cli/src/core/update.ts:545`, `apps/cli/src/commands/upgrade.ts:107`, `apps/cli/src/commands/upgrade.ts:122`; `scripts/portable/install.sh:415`]
✅ 2. The npm package now ships and executes the cleanup from Y's postinstall payload, fixing the npm X -> Y in-memory boundary.
✅ 3. Unexpected `launchctl` failures retain retry state, while daemon and installer entry points run cleanup before current-client credential/service gates.
✅ 4. Current CI checks are green on head `60791d0c`; no database or core data-structure change is present.

**Action taken**
- Submitted request changes on head `60791d0ce9eddc9a545cc7684fc521f6388deed6`.

yuezengwu (changes requested):
## Successor-head review — request changes

Reviewed head `60791d0ce9eddc9a545cc7684fc521f6388deed6`.

### Goal and core change

This PR retires the removed `github-scan` launchd runner through a retry-safe cleanup, invokes it before current daemon gates, and now adds a newly installed npm-package postinstall hook plus portable installer recovery. The npm X -> Y adoption boundary is correctly fixed and the published payload contains the new hook. One distribution-path blocker remains.

### Remaining blocker

**Portable self-update still does not execute Y when no service restart follows.** The standalone `scripts/portable/install.sh` does invoke the newly switched binary's `daemon ensure-service`, but a portable X running the public `first-tree upgrade` path does not run that installer. X's already-loaded `installPortableSpec()` downloads/extracts Y, switches `current`, rewrites the shims, and returns. X's `upgrade` implementation then returns for `--no-restart`, unsupported service control, or `state === "not-installed"`; none of those paths invokes Y, so the legacy runner can remain active on the same service-less adoption case.

The new postinstall test covers npm, and the command-matrix test proves cleanup-only `ensure-service` once invoked, but neither covers portable X -> Y self-update. Add a Y-side first-eligible-run bootstrap that does not depend on an existing daemon/service, and a version-boundary test demonstrating that a service-less portable X -> Y transition executes Y's migration. (`apps/cli/src/core/update.ts:545-547`, `apps/cli/src/commands/upgrade.ts:107-137`, `scripts/portable/install.sh:415`)

### Verified fixes and validation

- npm package dry-run includes `scripts/retire-legacy-github-scan-launchd.mjs`; its Y-artifact postinstall test passed.
- Retry semantics, pre-gate daemon cleanup, and cleanup-only `ensure-service` are sound.
- Retirement 9/9, daemon-start 31/31, postinstall 2/2, CLI-matrix 8/8, CLI typecheck, build, touched-file Biome, full Biome, and diff check passed. Full Biome reports only existing warnings/info outside this PR.

No database or core data-structure change is present.

bestony:
Addressed the portable self-update blocker in `c8307472`.

After `installPortableMeta` switches `current` and rewrites the shims, it now launches the newly installed artifact directly via its bundled Node runtime and `daemon ensure-service` with `FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY=1`. This does not depend on service support, credentials, restart policy, or the old process's loaded modules. Unexpected bootstrap failures are logged best-effort while the verified install remains successful.

The portable fixture now records the migration-only environment from the newly installed runtime, proving the service-less X -> Y handoff; the full portable suite passes 22/22.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1892

- Rationale: The new portable bootstrap is still added to the updater code running in X, so it cannot execute when users first upgrade from an older X into the release Y that introduces it.

**Risk level**: A
- Semantic lift: none (already A; this changes portable distribution and daemon migration behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: Use retry-safe cleanup before daemon gates, npm Y-artifact postinstall, installer recovery, and a portable updater hook intended to launch the newly switched artifact.
- Impacted modules: `apps/cli` package lifecycle, portable self-update, daemon commands, cleanup core, focused tests

**Review findings**
❌ 1. Move this hook out of `installPortableMeta()` and into a path that Y itself executes on its first eligible run. During the adoption transition, portable X is still executing X's bundled `update.ts`; X's `installPortableMeta()` does not contain the newly added `runPortableMigrationBootstrap()` call. Switching `current` to Y does not reload that module, so X returns exactly as before and never launches Y in the service-less / `--no-restart` case. The updated test imports the current implementation and uses it to install a fixture target, so it proves Y -> Z behavior, not the missing X -> Y boundary. A Y-side CLI migration bootstrap (through the centralized upgrade-migration entry point, excluding help/version-only flows as appropriate) is needed, with a test that actually models an old updater lacking the hook and then invokes Y for its first eligible command. [R5 / `apps/cli/src/core/update.ts:495`, `apps/cli/src/core/update.ts:578`; `apps/cli/src/__tests__/update-portable-install.test.ts:414`]
✅ 2. The helper correctly uses the target artifact's bundled Node and CLI when it is called, and preserves the verified portable install on best-effort migration failure.
✅ 3. npm X -> Y postinstall, retry semantics, daemon/installer pre-gate cleanup, and the focused wiring remain sound.

**Action taken**
- Submitted request changes on head `c8307472a59768230e29a6a063ee53777c078ddc`.

yuezengwu (changes requested):
## Successor-head review — request changes

Reviewed head `c8307472a59768230e29a6a063ee53777c078ddc`.

### Goal and core change

This PR retires the stranded legacy `github-scan` launchd runner with retry-safe cleanup, pre-gate daemon/installer wiring, and newly installed npm/portable migration paths. The new helper correctly invokes the target portable artifact when it is called, but the first portable adoption boundary is still not reached.

### Remaining blocker

**The portable hook is still code in Y's updater, so portable X cannot call it while adopting Y.** An X -> Y self-update continues running X's already-loaded `installPortableMeta()`. The X implementation switches `current`, rewrites the shims, and returns; it does not contain the newly added `runPortableMigrationBootstrap()` call. Installing Y does not reload that in-memory module, so the service-less / `--no-restart` path still exits without invoking Y and leaves the legacy runner active.

The updated test imports the head implementation of `installPortableSpec()` and uses that new updater to install a fixture target. It proves that Y can invoke Z after a future Y -> Z update, not that an older X lacking the hook reaches Y during the adoption transition. Move the migration into a first-eligible-run path that Y itself executes (with help/version-only behavior kept appropriately read-only), and add a version-boundary test that actually runs an old updater without this hook before invoking Y's eligible bootstrap. (`apps/cli/src/core/update.ts:495`, `apps/cli/src/core/update.ts:581`, `apps/cli/src/__tests__/update-portable-install.test.ts:427`)

### Verified behavior and validation

- When reached, the helper uses the target artifact's bundled Node/CLI, sets cleanup-only portable state, and treats migration failure as best-effort without rolling back a verified install.
- npm X -> Y postinstall, retry preservation, daemon/installer pre-gate cleanup, and cleanup-only `ensure-service` remain sound.
- Focused suites passed 72/72 (portable 22, daemon-start 31, retirement 9, CLI matrix 8, postinstall 2); CLI typecheck, build, touched-file Biome, and diff check also passed. Build emitted only the existing unresolved-workspace-dependency warnings.

No database or core data-structure change is present.

bestony:
Reworked the portable adoption path in `266de8a9`.

The migration trigger is now in Y’s CLI entrypoint, not X’s updater. On the first eligible non-help/non-version command, Y runs the legacy sweep before parsing the requested command. This covers a service-less portable X -> Y transition after X switches `current` and exits without restarting anything. Cleanup-only installer paths set an environment guard to avoid duplicate startup work, and help/version-only invocations remain read-only.

The CLI entry test now models an old portable updater that switches the root without a migration hook, then imports Y for `status` and asserts Y performs the migration. Local CLI-entry, portable, typecheck, Biome, and diff checks pass.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1892

- Rationale: The Y-side entry point fixes the adoption boundary, but it turns retained cleanup failures into a synchronous delay on every ordinary CLI command and retries them twice on daemon paths.

**Risk level**: A
- Semantic lift: none (already A; this changes the CLI root, package installation, portable distribution, and daemon lifecycle behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: Retire legacy launchd plists with retry-safe semantics, invoke the freshly installed npm artifact from postinstall, preserve installer/daemon recovery hooks, and now run the migration from Y's first eligible CLI command so a service-less portable X -> Y transition is eventually covered.
- Impacted modules: CLI root lifecycle, npm package lifecycle, portable upgrade/install paths, daemon commands, cleanup core, focused tests

**Review findings**
❌ 1. Make the new root migration hook bounded and single-shot instead of synchronously retrying on every command. `retireLegacyGithubScanLaunchd()` calls `launchctl` with `spawnSync` and a 15-second timeout for each plist. An unexpected failure intentionally retains the plist, so the new root hook makes every subsequent `status`, `chat`, `tree`, and other non-help/version invocation block for up to 15 seconds per plist before command parsing. `daemon start` and a normal `daemon ensure-service` then call the same helper again in their actions; after a retained failure, one invocation immediately repeats the same delay, reaching 30 seconds per plist on daemon startup. This contradicts the hook's stated guarantee that the best-effort migration must not block the requested command and also scatters the same upgrade task across multiple execution points. Route it through one fixed new-binary migration boundary, invoke it at most once per process, and bound or durably throttle retained retries on ordinary CLI commands while preserving a later retry opportunity. Add focused coverage for a retained/non-benign failure through the CLI entry and for exactly-once daemon invocation. [R1/R4/R5 / `apps/cli/src/cli/index.ts:41`, `apps/cli/src/core/retire-github-scan-launchd.ts:56`, `apps/cli/src/core/retire-github-scan-launchd.ts:120`; `apps/cli/src/commands/daemon/start.ts:92`, `apps/cli/src/commands/daemon/ensure-service.ts:27`]
✅ 2. Moving the adoption trigger into Y's CLI entry closes the earlier portable X -> Y in-memory boundary, and the raw-argument guard keeps help/version-only invocations read-only.
✅ 3. Unexpected `launchctl` failures still preserve the plist for retry, while confirmed success/absence removes it and leaves unrelated legacy files untouched.

**Action taken**
- Submitted request changes on head `266de8a95a46640457b7c7f605bc9bc8af158013`.

yuezengwu (changes requested):
## Successor-head review — request changes

Reviewed head `266de8a95a46640457b7c7f605bc9bc8af158013`.

### Goal and core change

This PR retires the stranded legacy `github-scan` launchd runner with retry-safe cleanup and new-binary adoption hooks for npm, portable installs, daemon startup, and service recovery. Moving the portable adoption trigger into Y's CLI entrypoint fixes the prior X -> Y in-memory boundary. Two root-lifecycle blockers remain.

### Blockers

1. **Do not synchronously retry retained failures on every CLI process, or twice on daemon paths.** The new root hook calls `retireLegacyGithubScanLaunchd()` before parsing every ordinary command. Each plist can block in `spawnSync("launchctl", ...)` for 15 seconds. A non-benign failure deliberately retains the plist, so every later `status`, `chat`, `tree`, etc. invocation repeats that delay; `daemon start` and normal `daemon ensure-service` then call the helper again in their actions, immediately doubling the wait in the same process. With multiple retained plists, latency is additive. This contradicts the claim that the best-effort migration cannot block the requested command. Centralize the migration behind an at-most-once-per-process boundary, and bound or durably throttle retries from ordinary commands while preserving a later retry opportunity. Cover a retained failure through the CLI root and assert exactly one attempt on daemon paths. (`apps/cli/src/cli/index.ts:41`, `apps/cli/src/core/retire-github-scan-launchd.ts:56`, `apps/cli/src/commands/daemon/start.ts:92`, `apps/cli/src/commands/daemon/ensure-service.ts:27`)

2. **The positional `help` command is still mutating.** Commander exposes `help [command]`, but `shouldRunFirstRunMigrations()` only excludes `-h` / `--help` flags and version flags. Therefore `first-tree help status` reaches the synchronous bootout/plist deletion before Commander renders help, despite this commit's stated help-only read-only guarantee. The focused test covers only `--help` and `--version`, so it misses the supported positional form. Prefer a Commander action/preAction boundary (which help/version do not enter), or otherwise classify and test `help [command]` as read-only too. (`apps/cli/src/cli/index.ts:32-43`, `apps/cli/src/__tests__/cli-entry-and-tree-actions-extra.test.ts:155`)

### Verified behavior and validation

- Y-side entry execution now closes the service-less portable X -> Y adoption boundary; the old updater no longer needs to know about the migration.
- The first `channel-env.js` import remains intact, the npm cleanup-only guard reaches `ensure-service`, and retry-preserving plist semantics remain sound.
- Focused suites passed 76/76 (CLI entry 4, portable 22, daemon-start 31, retirement 9, CLI matrix 8, postinstall 2); CLI typecheck, build, touched-file Biome, and diff check passed. Build emitted only existing unresolved-workspace-dependency warnings.

No database or core data-structure change is present.

bestony:
Addressed the root-lifecycle blockers in `281baba8`.

- Migration now runs only from Commander’s centralized `preAction` boundary; daemon start and ensure-service no longer call it again.
- `help`, `help <command>`, `--help`, and `--version` bypass `preAction` and stay read-only.
- The root boundary is process-once guarded. Retained failures use a durable six-hour retry cooldown and a one-second total launchctl budget, so ordinary commands cannot repeatedly block for 15 seconds per plist while preserving a later retry opportunity.
- Added coverage for positional help, exactly-once daemon-boundary invocation, retained-failure cooldown state, and bounded timeout behavior.

Focused validation, typecheck, Biome, and diff checks pass locally.

yuezengwu (approved):
## Successor-head review — approve

Reviewed head `281baba831489287fa27e5db718a8a6de4580c77`.

### Goal and core change

This PR retires the stranded legacy `github-scan` launchd runner with a macOS-only, retry-safe sweep. The final design invokes the migration from the newly installed binary at Commander's centralized `preAction` boundary, while npm postinstall and portable installer flows enter that same boundary through cleanup-only `daemon ensure-service` invocations.

### Resolution of prior blockers

- Non-benign `launchctl bootout` failures retain their plist, while confirmed success or absence removes it, preserving a real retry path.
- The Y-side entrypoint closes the X -> Y adoption boundary for service-less portable upgrades; npm postinstall likewise invokes Y rather than relying on X's already-loaded upgrade code.
- `help`, positional `help <command>`, `--help`, and `--version` remain read-only because Commander does not enter `preAction` for those flows.
- The migration is centralized and process-once; daemon start and ensure-service no longer invoke it a second time.
- Retained failures are bounded by a one-second total `launchctl` budget and durably throttled for six hours, preventing repeated long delays while retaining a later retry opportunity.

The new durable local artifact, `~/.first-tree/state/legacy-github-scan-launchd.json`, contains only the retry deadline, is written with mode `0600`, and is removed after successful/absent cleanup. There is no database or core data-structure change.

### Validation

- Focused CLI/lifecycle suites: 78/78 passed.
- CLI typecheck and build passed.
- Biome on every touched file passed; full repository `pnpm check` passed with only pre-existing warnings/info.
- Exact-base diff check passed.
- GitHub CI, CodeQL, portable smoke, lint/typecheck, and test checks are green on this head.

**Recommendation: approve.**

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1892

**Recommendation**: approve
- Rationale: Head `281baba831489287fa27e5db718a8a6de4580c77` resolves all previously identified retry, adoption-boundary, root-lifecycle, and read-only-help blockers without expanding the cleanup scope.

**Risk level**: A
- Semantic lift: none (already A; this changes CLI startup, package installation, portable distribution, and daemon migration behavior)

**PR summary**
- Problem: macOS users upgrading from the retired `github-scan` subsystem can retain a KeepAlive launchd job that crash-loops and occupies port 7878.
- Approach: The newly installed CLI owns one centralized Commander `preAction` migration boundary reached by ordinary commands, npm postinstall, and portable installer recovery. It boots out labels discovered from legacy plists, deletes only confirmed-retired plists, bounds launchctl work to one second per process, and durably throttles retained failures for later retry while keeping help/version flows read-only.
- Impacted modules: CLI root lifecycle, npm package lifecycle, portable install/update paths, daemon recovery command, cleanup core, focused tests

**Review findings**
✅ 1. The Y-side CLI boundary closes both npm and service-less portable X -> Y adoption gaps without relying on X's already-loaded updater modules.
✅ 2. Migration execution is centralized and process-once; daemon start and ensure-service no longer duplicate the cleanup.
✅ 3. Non-benign bootout failures retain the plist, while success/confirmed absence removes it; the one-second aggregate budget and six-hour durable cooldown prevent repeated command latency.
✅ 4. `help`, positional `help <command>`, help flags, and version flows remain read-only because they do not enter Commander's `preAction`; the required channel-environment import remains first.
✅ 5. The filesystem scope remains macOS-only and limited to legacy launchd plists plus the retry-deadline state; legacy config/log files remain untouched. GitHub CI is green on this exact head, and no database or core data-structure change is present.

**Action taken**
- Approved head `281baba831489287fa27e5db718a8a6de4580c77` with explicit authorization from baixiaohang.

PR #1894, opened 2026-07-20
feat: let capable agents provision teammates

### LINKED ISSUE #1885: feat(cli): let agents provision teammate agents — make `agent create` + follow-up setup commands agent-executable
## Problem

Today `agent create` / `agent bind` are operator-only: the workspace contract injected into every agent says they must be run from the web console or a human terminal, never from inside a running agent. When a human asks a coordinator agent to set up new teammate agents, the agent can prepare everything (names, prompt fragments, workflow rules) but then has to bounce back to the human, who becomes a copy-paste executor of commands the agent already composed.

Concrete case (2026-07-20): setting up a dev/reviewer agent pair (`fire-fable-dev` + `fire-fable-reviewer`) for a model-competition workflow on this repo. The coordinator agent had both prompt fragments ready; the human still had to run the create commands by hand, and hit CLI friction on the way (see UX notes below).

## Request

Allow an authenticated agent to execute the full teammate-provisioning series itself:

1. `first-tree agent create <name> --type agent --runtime <runtime>` — create the agent and bind it on the current client
2. `first-tree agent config prompt set <name> -f <file>` — install the per-agent prompt fragment
3. Model / account / runtime settings that today are only configurable in the web console (`agent create` has no `--model` flag)
4. `first-tree chat invite <name>` / `chat create --to <name>` — already agent-executable; listed for completeness

Steps 1–3 are the gap.

## Suggested guardrails

Agent self-provisioning is an escalation surface, so gate it rather than opening it wide:

- Scope: an agent may only create agents in its own org, on clients owned by its managing human.
- An explicit per-agent capability flag, or a one-time human approval in the web UI (similar to a tracked open question) before the create takes effect.
- Quota and an audit log (who created what, from which chat/session).

## CLI UX notes (worth fixing even if the above is declined)

- `--type <type>` is required with enum `(human, agent)`, and the requirement only surfaces as a runtime error after the first attempt. `agent` should be the default — a human typing `agent create` almost never means `--type human`.
- `agent create --help` does not indicate which options are required.

## Environment

- CLI: `0.5.16-staging.881.1`
- Reported from chat `899ef9bf-9a52-4607-b2a6-c00f6610e115` by agent `liuchao-staff` on behalf of `liuchao-001`

bestony:
Closes #1885\n\n## Summary\n- add an admin-granted per-agent provisioning capability with immediate revoke\n- authorize running agents with manager/client runtime proof, enforce org scope and existing maxAgents quota\n- persist acting-agent, managing-member, chat/session, created-agent, and timestamp audit records\n- add CLI capability grant/revoke, default --type agent, required-option help text, and creation-time --model\n- cover authorization, revocation, runtime proof, model, prompt setup, and audit behavior\n\n## Verification\n- pnpm --filter @first-tree/shared typecheck\n- pnpm --filter @first-tree/server typecheck\n- pnpm --filter first-tree-dev typecheck\n- env -u FIRST_TREE_SERVER_URL -u FIRST_TREE_CLIENT_ID pnpm --filter first-tree-dev test (one unrelated process-environment-sensitive test is flaky under the injected First Tree runtime; isolated rerun passes)\n- first-tree tree verify passed for the companion Context Tree change\n\nCompanion Context Tree PR: https://github.com/agent-team-foundation/first-tree-context/pull/new/context/issue-1885-agent-provisioning


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1894

- Rationale: the intended standing capability is bypassable by omitting caller-supplied actor headers, and the shipped runtime briefing still instructs capable agents never to execute the core command.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A (the mixed PR also changes the agent service and DB schema).
- Semantic lift: none beyond A.

**PR summary**
- Problem: coordinator agents currently have to hand teammate-provisioning commands back to a human, turning the human into a copy-paste executor.
- Approach: add an admin-granted standing capability, require a bind-scoped runtime proof for attributed agent creates, scope the create to the manager and organization, write an audit row, and extend the CLI with capability controls plus creation-time model/default-type support.
- Impacted modules: CLI agent commands and runtime contract, Server agent authorization/services, shared schemas, and PostgreSQL schema/migration.

**Review findings**

❌ 1. The capability is optional attribution, not an authorization boundary. `requireProvisioningActor` returns `null` whenever the caller omits `x-first-tree-acting-agent`/`x-agent-id`, after which the org route follows the ordinary human-member create path. A running agent uses the same persisted member JWT as the human CLI, so it can run the command with `FIRST_TREE_AGENT_ID` removed (or call HTTP directly) and create without the capability or audit row. An admin-managed agent can similarly omit the header on the grant endpoint and grant itself. The Server must derive agent execution from an unforgeable credential/context rather than trusting the presence of a caller-controlled header; add negative coverage for an agent runtime attempting the request without the actor header. [`packages/server/src/services/agent-provisioning.ts:25`, `packages/server/src/api/orgs/agents.ts:141`, `apps/cli/src/core/provisioning-actor.ts:6`]

❌ 2. The runtime contract delivered to every agent still says that `agent create / bind` must never run inside a running agent. A compliant capable agent will therefore refuse the exact workflow this PR is meant to enable. Replace the blanket prohibition with a capability-aware rule and update the briefing tests in the same change. [`packages/client/src/runtime/templates/agent-briefing.ejs:294`]

❌ 3. The new admin command cannot grant capability to an agent managed by another member. It resolves the target exclusively through `/me/managed-agents`, whose contract is the caller's personally managed agents, even though the Server endpoint and current identity model allow an org admin to administer org-scoped agents. Use an org-admin lookup (with explicit org disambiguation) or otherwise let the command address any eligible agent in the admin's organization. [`apps/cli/src/commands/agent/capability.ts:21`, `apps/cli/src/commands/_shared/resolve-agent.ts:6`]

❌ 4. The durable audit provenance for chat/session is copied from arbitrary request headers without checking that either value belongs to the acting agent or that the session belongs to the chat. A capable agent can therefore stamp another chat/session—or invented values—onto the audit row. Validate the supplied context against Server-owned session/chat state, or omit unverifiable fields rather than persisting them as provenance. [`packages/server/src/services/agent-provisioning.ts:66`]

⚠️ 5. The feature relies on the existing check-then-insert quota path, which explicitly allows concurrent requests to pass the same count and exceed `maxAgents`. That was already acknowledged as Phase-1 debt, but exposing create to autonomous agents materially increases concurrent-create blast radius and does not satisfy strict quota enforcement. Serialize the quota decision with the create transaction before treating it as a guardrail. [`packages/server/src/services/agent.ts:524`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review by yzw-codex — head `c72a0dbd`

**Recommendation: request changes.**

### Goal and core changes

This PR turns teammate provisioning from a human-only CLI workflow into a standing, admin-granted agent capability. It adds `agents.can_provision_agents`, a durable provisioning-audit table, runtime-proof-based actor checks, CLI grant/revoke and create-time model support, and keeps the existing organization quota/client ownership checks in the create path.

I independently confirmed the authorization/header, generated-briefing, admin target-resolution, audit-provenance, and concurrent-quota blockers already reported in the preceding review. One additional production-path blocker needs to be explicit:

### Blocker: the claimed session audit field is never populated by the runtime

`provisioningActorHeaders()` reads `FIRST_TREE_SESSION_ID`, and the new tests manually install that variable, but the production runtime's `buildAgentEnv()` exports only the agent, runtime-token-file, chat, client, and provider context. There is no production definition or injection of `FIRST_TREE_SESSION_ID` anywhere outside this PR's helper/tests. Consequently, real in-agent provisioning always writes `session_id = NULL`, while the test passes only because it fabricates an environment the runtime never produces.

Please either wire a clearly defined, runtime-owned session identifier end to end and validate its agent/chat relationship on the Server, or remove/change the session-provenance promise. Do not persist arbitrary client-supplied chat/session strings as authoritative audit provenance. Relevant paths: `apps/cli/src/core/provisioning-actor.ts:20`, `packages/client/src/runtime/agent-io.ts:43`, `packages/client/src/runtime/session-manager.ts:2271`, `packages/server/src/services/agent-provisioning.ts:66`.

### Human/schema decisions required before merge

- The capability boundary currently depends on caller-supplied header presence while the agent can access the managing user's persisted JWT. That cannot enforce “human admin grants, agent without capability is denied” against omitted headers; the identity/credential boundary needs a human-approved design.
- Confirm target-client semantics: companion Context Tree PR #787 says cross-client provisioning is unsupported, but `createAgent()` accepts any non-retired client owned by the managing user, including a different client from the acting agent's bound client.
- Confirm audit retention/deletion semantics for the new core table: organization deletion cascades audit rows, while acting/created agent and managing-member deletion are restricted; `chat_id`/`session_id` have no referential constraint.

### Verification

On this exact head I ran shared/server/CLI typechecks, 37 targeted CLI tests, 5 provisioning/quota server tests, and Biome on the 19 changed source files; all passed. Those tests do not cover the missing-header bypass or the absent production session environment, so green checks do not resolve the blockers.

bestony:
Addressed the requested blockers on head `58c4c4ea`:

- Provisioning actor resolution now derives the acting agent from the bind-scoped runtime token even when `x-first-tree-acting-agent` is absent. The CLI keeps sending the runtime proof independently of `FIRST_TREE_AGENT_ID`; a runtime with no capability receives the authorization error. Capability grant/revoke rejects runtime-session requests, including omitted actor headers.
- Audit chat provenance is retained only when the supplied chat is server-verified for the acting agent and organization. `FIRST_TREE_SESSION_ID` is not runtime-owned, so session strings are no longer persisted as authoritative provenance; the production briefing no longer promises it.
- Org-admin capability commands now resolve an explicitly selected org (or require `--org` when the user administers multiple orgs) and search the org-admin agent listing, so agents managed by other members are addressable.
- The quota count and insert now run under an organization row lock in one transaction.
- The generated runtime briefing and tests now describe capability-aware in-session provisioning.
- Provisioning is constrained to the acting agent's bound client, matching the no-cross-client scope.

Verification: shared/server/CLI typechecks, client briefing tests (31), CLI targeted tests (37), and Biome pass. Server integration tests are covered in CI; this checkout has no container runtime. The PR remains open for the required human review of schema/retention decisions.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1894 (head `58c4c4ea`)

- Rationale: the new head closes several workflow and concurrency findings, but the standing capability can still be bypassed by omitting both caller-controlled runtime headers while using the same persisted member JWT.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: none beyond A; the PR also changes core agent authorization and DB schema.

**PR summary**
- Problem: coordinator agents currently have to hand teammate-provisioning commands back to a human instead of completing the setup workflow themselves.
- Approach: add an admin-granted capability, bind-scoped runtime proof, same-org/same-client enforcement, serialized quota check, durable audit row, capability-aware runtime instructions, and CLI provisioning/configuration support.
- Impacted modules: CLI agent commands, generated runtime briefing, Server agent authorization/services, shared schemas, and PostgreSQL migration.

**Review findings**

❌ 1. The missing-header authorization bypass remains. The new token-derived lookup correctly handles a request that omits only the actor id, but `requireProvisioningActor` still returns the human path when both headers are absent. The running agent can execute `agent create` with both `FIRST_TREE_AGENT_ID` and `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE` removed—or call the HTTP route directly—while `ensureFreshAccessToken()` still reads the shared persisted member JWT. That creates without capability enforcement or an audit row. An admin-managed agent can likewise omit all three checked headers on the capability endpoint and grant itself. Because absence is not authenticated, this needs a credential/runtime boundary that the agent cannot opt out of; deriving identity only when a caller elects to send the runtime token does not satisfy the acceptance gate. [`packages/server/src/services/agent-provisioning.ts:28`, `packages/server/src/services/agent-provisioning.ts:32`, `packages/server/src/api/agents.ts:175`]

❌ 2. `chatId` is now range-checked, but it is still caller-selected rather than authenticated as the origin of this operation. The bind token is agent/client-scoped, so a capable agent can substitute any chat for which it has an old or current `agent_chat_sessions` row; the query does not bind the request to the current `(agent, chat)` runtime session. It can also force `NULL` by omitting or supplying an unverifiable value. If this column is only a best-effort agent claim, name/document it that way; if it is audit provenance for “which chat initiated the create,” it needs chat-scoped server-owned proof. [`packages/server/src/services/agent-provisioning.ts:89`]

❌ 3. The production decision is now that there is no runtime-owned session id, yet the PR still permanently adds `session_id`, carries it through the audit types, and reads/sends `FIRST_TREE_SESSION_ID`; the Server then hardcodes it to `NULL`. Remove this dead channel from the CLI, service types, schema, migration, and tests unless a server-owned session identity is defined in this PR. [`apps/cli/src/core/provisioning-actor.ts:19`, `packages/server/src/services/agent-provisioning.ts:112`, `packages/server/src/db/schema/agent-provisioning-audit.ts:24`]

✅ 4. The new head does close the previous generated-briefing conflict, org-admin target lookup gap, cross-client ambiguity, and quota TOCTOU finding.

**Action taken**
- Submitted request changes on `58c4c4ea`.

yuezengwu (changes requested):
## Successor-head review — `58c4c4ea`

**Recommendation: request changes.** This head fixes the generated-briefing conflict, org-admin lookup, same-client scope, untrusted session persistence behavior, and quota TOCTOU. The core standing-capability boundary is still optional, however, and one newly added Server test fails deterministically.

### Blockers

1. **The agent can still opt out of capability enforcement by omitting both runtime headers.** `requireProvisioningActor()` returns `null` when neither the actor id nor runtime proof is present; the org route then uses the ordinary human-member create path and writes no provisioning audit. A running agent can unset both `FIRST_TREE_AGENT_ID` and `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE`, or call HTTP directly, while `ensureFreshAccessToken()` continues to load the same persisted member JWT. The capability endpoint has the same problem: an admin-managed agent that omits every checked header is treated as its human admin and can grant itself. Token-derived actor lookup fixes omission of *one* caller-controlled header, but absence of all headers is still unauthenticated. The acceptance criterion needs an agent/runtime credential boundary the caller cannot opt out of; otherwise this is a cooperative CLI convention, not an authorization guard. [`packages/server/src/services/agent-provisioning.ts:28-33`, `packages/server/src/api/orgs/agents.ts:140-174`, `packages/server/src/api/agents.ts:171-185`, `apps/cli/src/core/bootstrap.ts:146`]

2. **The audit chat is associated with the agent, not authenticated as the origin of this operation.** The runtime proof is agent/client-scoped. A capable agent can supply any current or historical chat having an `agent_chat_sessions` row for that agent, or force `NULL` by omitting/unverifiably selecting a chat. If `chat_id` means “chat that initiated this create,” bind it to server-owned current-session context; if it is only an optional caller-selected related chat, rename/document the weaker provenance and reconcile the issue/PR/Context Tree claim. [`packages/server/src/services/agent-provisioning.ts:89-105`]

3. **The abandoned session provenance remains permanently encoded in the new schema.** The Server now always writes `sessionId: null`, but the CLI still reads/sends `FIRST_TREE_SESSION_ID`, the service types carry it, and migration 0082 creates `session_id`. Because this migration is new, remove the dead channel rather than shipping an always-null audit field unless this PR defines server-owned session identity. The PR summary and binding acceptance text still promise chat/session provenance. [`apps/cli/src/core/provisioning-actor.ts:17-20`, `packages/server/src/services/agent-provisioning.ts:107-115`, `packages/server/src/db/schema/agent-provisioning-audit.ts:23-25`, `packages/server/drizzle/0082_organic_eddie_brock.sql:7-9`]

4. **The new Server test fails on this exact head.** An invalid runtime token with an actor id returns `"This agent is not authorized to provision agents"`, while the test requires an active-runtime-session error. The isolated rerun reproduces 1 failure / 1 pass. Keep the externally intended error semantics and implementation/test aligned. [`packages/server/src/services/agent-provisioning.ts:76-79`, `packages/server/src/__tests__/agent-provisioning.test.ts:154-165`]

5. **The public CLI reference omits the new `--org` option.** The implementation requires `--org` when the caller administers multiple organizations, while both the namespace snapshot and examples document only `capability grant|revoke <agent>`. Update `docs/cli-reference.md` and add behavior-level tests for admin-org selection, pagination/lookup, and runtime-request rejection; the current registration smoke test does not exercise the command. [`apps/cli/src/commands/agent/capability.ts:11-80`, `docs/cli-reference.md:205-245`]

### Human schema decision still required

Migration 0082 adds durable audit rows whose actor, manager, and created-agent foreign keys use `ON DELETE RESTRICT`, with no retention/purge policy. That can intentionally preserve provenance but also prevents physical deletion of referenced agents/members indefinitely. A human owner should explicitly approve retention, deletion/export behavior, and whether best-effort chat association is acceptable before this DB contract ships.

### Verification

- Passed: shared, Server, Client, and CLI typechecks.
- Passed: 38 targeted CLI tests, 31 Client briefing tests, 3 quota tests, and Biome on changed files.
- Failed twice, including isolated single-worker rerun: `packages/server/src/__tests__/agent-provisioning.test.ts` (1 failed, 1 passed).

The successor head is materially better, but the first acceptance gate—an agent without the capability cannot create—remains bypassable under the current shared member-JWT model.

bestony:
Successor update on head `3b3f2f5f`:

- Removed the dead session provenance channel completely: no `FIRST_TREE_SESSION_ID`, no session header/type, no `session_id` audit column in schema or migration 0082, and no test fixture.
- Invalid actor-bound runtime proofs now return the active-runtime-session error consistently.
- Added the public CLI `--org` contract and registration coverage; org lookup remains paginated and rejects ambiguous admin organizations.
- Chat audit association is explicitly documented as an optional best-effort related-chat claim, not authenticated proof of the initiating turn.

The remaining blocker is architectural: with the current shared `type=access` member JWT, a running agent can remove every runtime header and become indistinguishable from the human member. I have opened a tracked decision with the maintainer proposing a server-issued operator/agent credential boundary; the PR remains unmergeable until that credential design and audit retention policy are approved.

yuezengwu (changes requested):
## Successor-head review — `3b3f2f5f`

**Recommendation: request changes.** This head now fixes the generated runtime briefing, org-admin target lookup, same-client scope, quota serialization, dead session column, public `--org` documentation, and the invalid-proof test failure. The central authorization boundary is still caller-optional, however.

### Blocker: omitting both runtime headers still bypasses the capability

`requireProvisioningActor()` returns `null` when neither actor id nor runtime proof is present. The org create route then deliberately takes the ordinary human-member path, so it neither checks `canProvisionAgents` nor writes a provisioning audit row. A running agent can unset both `FIRST_TREE_AGENT_ID` and `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE`, or call HTTP directly without those headers, while `ensureFreshAccessToken()` still loads the same persisted member JWT used by the human CLI.

The capability endpoint has the same structural bypass: it rejects requests only when one of the caller-controlled actor/selector/runtime headers is present, so an admin-managed agent can omit all of them and grant itself. The same shared JWT also means follow-up provisioning/setup routes cannot reliably distinguish a capable agent from its managing human.

Token-derived lookup correctly handles omission of *one* header, but absence of every header is not authenticated. The acceptance gate requires an agent/runtime credential or trusted execution boundary that the caller cannot opt out of; otherwise the flag is a cooperative CLI convention rather than authorization. Add the binding negative case—an agent runtime request omitting both actor and proof—once such a boundary exists. [`packages/server/src/services/agent-provisioning.ts:27-32`, `packages/server/src/api/orgs/agents.ts:140-174`, `packages/server/src/api/agents.ts:171-185`, `apps/cli/src/core/bootstrap.ts:146`]

### Audit contract requires an explicit human decision

The implementation now honestly documents `chatId` as an optional caller-selected “related-chat claim,” not proof of the initiating turn. It accepts any current or historical chat with an `agent_chat_sessions` row and can be forced to `NULL`; session provenance has been removed entirely. That is internally consistent, but it is weaker than issue #1885’s binding acceptance text, the PR summary, and the companion Context Tree claim of chat/session provenance. Either bind audit context to server-owned current-session proof, or have the human owner explicitly approve the weaker audit contract and reconcile all three sources before merge. [`packages/server/src/services/agent-provisioning.ts:15-19`, `packages/server/src/services/agent-provisioning.ts:88-110`]

### Human DB review remains required

Migration 0082 uses `ON DELETE RESTRICT` from audit rows to acting agents, managing members, and created agents, with no retention/purge policy. This can intentionally preserve provenance but also prevents physical deletion of referenced identities indefinitely. Please obtain explicit human approval for retention, hard-delete/export behavior, and the weaker related-chat semantics before this schema ships. [`packages/server/drizzle/0082_organic_eddie_brock.sql:1-17`]

### Verification on this exact head

- Passed: Server and CLI typechecks.
- Passed: 5 targeted Server tests, 38 targeted CLI tests, 31 Client briefing tests.
- Passed: Biome on the touched authorization, schema, CLI, and test files.

The successor commits materially improve the PR, but the first acceptance criterion—an agent without the capability receives an authorization error instead of creating—remains bypassable under the shared member-JWT model.

bestony:
Successor head `7a70af0a` fixes the failing `api-small-routes-extra` route-harness case by treating a missing mock `headers` object as empty; production Fastify requests remain unchanged. Server typecheck and Biome pass locally. The new CI run has passed lint/typecheck, migrations, CodeQL, and portable smoke; the full Test job is still running.

The new cross-manager prompt feedback is not implemented as an org-wide resolver bypass. Current code has two independent guards: CLI config target resolution is manager-only and Server config/resources require target manage scope. I opened a human decision recommending a separate, revocable target-agent delegation for designated coordinators rather than implicitly expanding `may provision agents` into organization-wide instruction mutation. The existing credential-boundary decision remains prerequisite to either model.

yuezengwu (changes requested):
## Successor-head review — `7a70af0a`

**Recommendation: request changes.** The feature still aims to let an admin-authorized running agent complete teammate provisioning with org/client/quota/audit guardrails. This successor commit only makes the production helper tolerate a route-test request that lacks a `headers` property; it does not implement the prerequisite credential boundary.

### Blocker: the capability remains caller-optional under the shared member JWT

The decisive branch is unchanged: after normalizing missing test headers to `{}`, `requireProvisioningActor()` returns `null` whenever both actor and runtime-proof headers are absent. The org create route then uses the ordinary member/admin manager path and passes no provisioning audit context. `userAuthHook` authenticates the same `type=access` JWT for the human and every managed runtime, reducing it to only `request.user.userId`, so the Server has no trusted caller-kind signal.

A running agent can therefore unset `FIRST_TREE_AGENT_ID` and `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE`, or issue the HTTP call directly without either header, while `agent create` still loads the persisted member JWT. An admin-managed agent can similarly omit all checked headers on the capability endpoint and grant itself. The token-derived path only works when the caller elects to provide a runtime proof; it does not make absence authoritative. [`packages/server/src/services/agent-provisioning.ts:28-34`, `packages/server/src/middleware/user-auth.ts:46-85`, `packages/server/src/api/orgs/agents.ts:140-174`, `packages/server/src/api/agents.ts:171-185`, `apps/cli/src/core/bootstrap.ts:146-155`]

This still fails the binding acceptance gate that an agent without the capability must receive an authorization error. It also conflicts with the locked Cloud constraints that the Server decides permissions and the Client is zero-trust. The proposed operator/agent credential decision must be resolved by a human owner first: the current decision-locked Context Tree says the member JWT is the unified long-lived credential and the bind proof cannot replace it or become a second identity. Keep this PR unmergeable until that durable boundary is deliberately revised or a compatible trusted-execution design is approved, then add a negative test for a real authenticated agent request with both headers absent.

### Test-harness coupling introduced by this commit

Fastify guarantees `request.headers`; the direct-handler fixture omits it at `api-small-routes-extra.test.ts:737-742`. Changing production code to accept that malformed mock can hide incomplete fixtures by silently skipping provisioning-actor resolution. Prefer adding `headers: {}` to the test request (or mocking the provisioning helper) and preserve the production type invariant. This commit does not independently widen real Fastify behavior—the empty-header bypass already existed—but it is not a security fix. [`packages/server/src/services/agent-provisioning.ts:28-33`, `packages/server/src/__tests__/api-small-routes-extra.test.ts:737-742`]

### Human product/schema decisions remain open

- Migration 0082 keeps audit references with `ON DELETE RESTRICT` for actor, managing member, and created agent, without a retention/purge/export policy. Explicit owner approval is required before this core schema ships.
- The companion Context Tree now accurately calls `chatId` a best-effort related-chat claim and removes session provenance, but issue #1885's binding acceptance text and this PR summary still promise chat/session provenance. Reconcile them only after the human owner accepts the weaker contract.
- The new cross-manager prompt request is a separate delegation boundary. Current manager-only CLI resolution and Server manage-scope checks should not be widened implicitly through `may provision agents`; the author's proposed separate revocable delegation decision is the safer scope.

### Verification on this exact head

- Passed: shared, Server, Client, and CLI typechecks.
- Passed: 21 targeted Server tests (including the route harness), 38 CLI tests, and 31 Client briefing tests.
- Passed: Biome on the touched authorization, route, schema, CLI, and test files.

The new test run is green, but the authorization outcome is unchanged. The author has also acknowledged that the credential boundary remains prerequisite and that the PR is not mergeable yet.

bestony:
Addressed the route-harness coupling on successor head `a6092a9d`: production authorization again relies on Fastify's guaranteed `request.headers` invariant, and the direct-handler fixture now supplies `headers: {}`. Server typecheck, Biome, and `git diff --check` pass locally.

The credential boundary, audit retention decision, and new cross-manager prompt delegation choice remain intentionally unresolved and keep this PR unmergeable.

yuezengwu (changes requested):
## Successor-head review — `a6092a9d`

**Recommendation: request changes.** The new commit correctly fixes the route-test fixture by supplying `headers: {}` and restores the production Fastify request invariant. That closes the test-harness coupling from the immediately preceding head. It intentionally does not address the merge-blocking authorization model.

### Blocker unchanged: a running agent can still opt out of the capability

`requireProvisioningActor()` still returns `null` when both actor and runtime-proof headers are absent. Because the human CLI and every managed runtime share the same persisted `type=access` member JWT, the Server cannot distinguish a human request from a running agent that unsets `FIRST_TREE_AGENT_ID` and `FIRST_TREE_RUNTIME_SESSION_TOKEN_FILE` (or calls HTTP directly). The org create route then takes the ordinary member/admin path, skips `canProvisionAgents`, and writes no provisioning audit. An admin-managed agent can likewise omit every checked header on the capability endpoint and grant itself. [`packages/server/src/services/agent-provisioning.ts:28-33`, `packages/server/src/middleware/user-auth.ts:46-85`, `packages/server/src/api/orgs/agents.ts:140-174`, `packages/server/src/api/agents.ts:171-185`]

This continues to fail issue #1885's first binding acceptance criterion. The author's choice to keep the PR unmergeable pending a credential decision is correct. That decision must be human-owned because the current `decisionLocksCode: true` Cloud node requires unified member-JWT auth, while the runtime-proof contract says the bind proof cannot replace the member JWT or become a second identity. Approve and record a compatible trusted boundary—or deliberately revise those durable constraints—before implementing it, then cover an authenticated agent request with both runtime headers absent.

### Human decisions still required

- Approve audit retention/deletion/export semantics for migration 0082's `ON DELETE RESTRICT` identity references.
- Reconcile the accepted audit contract: the companion Context Tree now accurately describes a best-effort related-chat claim with no session provenance, while the issue acceptance text and PR summary still promise chat/session provenance.
- Keep the new cross-manager prompt requirement as a separate, revocable delegation decision; do not implicitly broaden `may provision agents` into org-wide configuration mutation.

### Verification on this exact head

- Passed: Server typecheck.
- Passed: 21 targeted Server tests, including provisioning, quota, and the direct route harness.
- Passed: Biome on both changed files and `git diff --check` for the PR.

The fixture fix is good; the previously documented credential blocker remains the only implementation blocker on this successor head.

PR #1898, opened 2026-07-21
fix: restore multi-human agent follow fallback

### LINKED ISSUE #1887: [Design] Allow any work agent in a chat to follow GitHub entities without a human delegate
### Problem

An agent-initiated `github follow` currently has to resolve an unambiguous representative human. In practice, this means a work agent must either have a delegate relationship to a human or rely on a single-human fallback in the chat before it can follow an Issue or PR.

This requirement couples an explicit, chat-scoped action to the delegate model. It prevents a work agent from independently subscribing its current task chat to GitHub events, even though the durable subscription semantics are already “the chat follows, not the person.”

### Proposal

Allow any active agent that is a member of the current chat to explicitly follow a GitHub Issue, PR, Discussion, or Commit for that chat, without requiring the agent to be a human’s delegate or resolving a representative human first.

The delegate model should remain the fallback for system-initiated work directed at a human, such as a GitHub mention or review request when no existing task line exists. It should not gate an agent’s explicit decision to subscribe the chat it is already working in.

### Desired semantics

- Permission for an explicit agent follow is based on active chat membership.
- The explicitly following agent is the wake target for events delivered through that declared line.
- Audit/provenance records which agent declared the follow without treating a human as the action owner.
- Human-targeted webhook routing can continue to use the human’s configured delegate as a fallback.
- Existing chat-level follow/unfollow and conflict/rebind behavior remains coherent.
- The data model no longer requires a synthetic human-agent pair for an agent-declared subscription.

### Design question

Can we separate explicit agent-declared follow from human delegation, so work agents become first-class subscribers while delegation remains only the default routing mechanism for incoming human-owned work?

### Context Tree conflict and required follow-up

This proposal intentionally conflicts with the current durable decisions in:

- `product/work-model/delegate-model.md`, which says followed entities route through a human's delegate and that following requires a delegate.
- `system/cloud/automation/github-webhook-routing.md`, which requires an agent-issued follow to resolve an unambiguous representative human before writing the mapping.

The webhook-routing node currently has `decisionLocksCode: true`. If this design is accepted and implemented, the corresponding Context Tree decisions and rationale must be updated in the same delivery cycle so the tree reflects the new contract.

baixiaohang:
## Summary

- restore the shared GitHub/GitLab agent-follow fallback that selects the id-sorted-first active same-Team human member when no human explicitly links the calling agent
- keep exactly-one linked human precedence, multiple-linked ambiguity fail-closed, and the calling agent as declaration provenance and wake target
- preserve distinct GitHub wake lines when multiple agent followers share one human carrier and chat; the shared planner still emits one card and unions every eligible wake agent
- prevent a suspended or deleted sibling wake line from rejecting the shared card: trusted SCM sends retain the card/attention line while persisting and notifying only active native mentions
- leave human-issued follow, mapping schema, database state, sender selection, echo pruning, and delivery planning unchanged
- align route errors and the cross-provider QA case with the resulting contract

This intentionally implements the minimal pre-SCM-refactor behavior restoration approved for Issue #1887. It does not introduce first-class agent-owned attention lines or a database migration.

PR #1891 is included in this branch base. Its trusted SCM system attribution remains compatible: persisted `message.senderId` continues to use the human carrier while Web and Agent prompt surfaces render trusted provider cards as system messages.

## Context Tree

Draft decision update: https://github.com/agent-team-foundation/first-tree-context/pull/791

The code PR should merge first; the Tree PR can then be reconciled and marked ready.

## Verification

- targeted Server audience/follow/delivery suites: 66 tests passed on the post-#1891 base
- regression proves two unlinked agent followers sharing one human carrier receive one card with both wake IDs in `metadata.mentions` and both Inbox notifications
- lifecycle regression proves that after one shared-line follower is suspended, the card still lands once, only the active sibling remains in `metadata.mentions`, only that sibling receives a notifying Inbox row, and delivery reports no failure
- latest targeted GitHub delivery, shared message, GitLab deferred-delivery, and system-sender suites: 49 tests passed with `--maxWorkers=2`
- `pnpm --filter @first-tree/server typecheck` passed for the lifecycle successor; earlier monorepo typecheck passed 10/10 tasks
- full Server suite before the final base sync: 238 files / 2,558 tests passed with maxWorkers 2
- affected-file Biome checks and `git diff --check` passed
- matching Context Tree verification passed

## QA

Formal cross-surface QA is warranted because the change affects explicit entity follow behavior across both GitHub and GitLab. The existing SCM entity-attention parity case now covers multi-human no-link fallback, multiple-link fail-closed behavior, and stale suspended wake-line isolation.

Closes #1887.

### DIFF
--- a/packages/server/src/services/github-audience.ts
+++ b/packages/server/src/services/github-audience.ts
@@ -127,12 +127,12 @@ export async function resolveGithubAudience(db: Database, event: NormalizedScmEv
       ),
     );
 
-  // Dedup subscribed rows by `(humanAgentId, chatId)` (keep earliest
-  // `bound_at`). A single `(human, entity)` pair can carry multiple mapping
-  // rows pointing at the *same* chat (one per delegate that ever drove an
-  // event for this entity under this human); collapsing those to one row is
-  // loss-free and stops `deliverGithubEvent` posting N identical cards
-  // to that chat.
+  // Dedup subscribed rows by `(humanAgentId, delegateAgentId, chatId)` (keep
+  // earliest `bound_at`). Alias entity keys can leave duplicate rows for the
+  // same logical attention line, but distinct delegates are distinct wake
+  // lines even when they share one human carrier and chat. The per-chat
+  // delivery planner collapses those lines into one card and unions every
+  // wake agent.
   //
   // The key MUST include `chatId`. Deduping by `humanAgentId` alone assumed
   // every row for a human shared one chat — false once the same human is
@@ -141,14 +141,16 @@ export async function resolveGithubAudience(db: Database, event: NormalizedScmEv
   // another, each under a different delegate). Collapsing across chats kept
   // only the earliest chat's row and silently dropped every *other* followed
   // chat from the audience — that chat never received the entity's events at
-  // all. "The chat follows, not the person", so the surviving unit is one
-  // row per (human, chat), never one per human.
-  const earliestByHumanChat = new Map<string, (typeof subscribedRows)[number]>();
+  // all. The key also includes `delegateAgentId`: two agent-issued follows can
+  // share the same fallback human and chat while carrying distinct wake
+  // agents. "The chat follows, not the person", so the surviving unit is one
+  // row per (human, delegate, chat).
+  const earliestByAttentionLine = new Map<string, (typeof subscribedRows)[number]>();
   for (const row of subscribedRows) {
-    const key = `${row.humanAgentId}:${row.chatId}`;
-    const current = earliestByHumanChat.get(key);
+    const key = `${row.humanAgentId}:${row.delegateAgentId}:${row.chatId}`;
+    const current = earliestByAttentionLine.get(key);
     if (!current || row.boundAt < current.boundAt) {
-      earliestByHumanChat.set(key, row);
+      earliestByAttentionLine.set(key, row);
     }
   }
   // #766: A `pull_request.opened` delivery reaching a reviewer purely through
@@ -178,19 +180,21 @@ export async function resolveGithubAudience(db: Database, event: NormalizedScmEv
     return row.humanAgentName !== null && involvedLogins.has(row.humanAgentName.toLowerCase());
   };
 
-  const subscribed: AudienceTarget[] = [...earliestByHumanChat.values()].filter(keepSubscribedOpened).map((row) => ({
-    humanAgentId: row.humanAgentId,
-    delegateAgentId: row.delegateAgentId,
-    kind: "existing",
-    chatId: row.chatId,
-    involveReason: null,
-    involveLogin: null,
-    provenance: isDeclaredBoundVia(row.boundVia)
-      ? "explicit"
-      : row.boundVia === "fixes_link"
-        ? "related_entity"
-        : "identity_target",
-  }));
+  const subscribed: AudienceTarget[] = [...earliestByAttentionLine.values()]
+    .filter(keepSubscribedOpened)
+    .map((row) => ({
+      humanAgentId: row.humanAgentId,
+      delegateAgentId: row.delegateAgentId,
+      kind: "existing",
+      chatId: row.chatId,
+      involveReason: null,
+      involveLogin: null,
+      provenance: isDeclaredBoundVia(row.boundVia)
+        ? "explicit"
+        : row.boundVia === "fixes_link"
+          ? "related_entity"
+          : "identity_target",
+    }));
 
   const subscribedByHuman = new Map<string, AudienceTarget[]>();
   for (const target of subscribed) {
--- a/packages/server/src/services/scm-attention-line.ts
+++ b/packages/server/src/services/scm-attention-line.ts
@@ -91,8 +91,9 @@ export async function executeScmFollowLine<TRecord extends ScmFollowLineRecord>(
  * Resolve the durable attention owner for an agent-issued follow.
  *
  * The selected agent is always the wake side. The human side is its linked
- * active human in the chat, or the chat's sole active human when no explicit
- * delegate relationship exists. Ambiguous ownership fails closed.
+ * active human in the chat, or the id-sorted-first active human when no
+ * explicit delegate relationship exists. Multiple explicit links are
+ * ambiguous and fail closed.
  */
 export async function resolveAgentScmBindingPair(
   db: Database,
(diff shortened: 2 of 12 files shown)


packages/server/src/services/scm-attention-line.ts
```diff
@@ -132,7 +133,7 @@ export async function resolveAgentScmBindingPair(
   );
   const linkedHumans = humans.filter((human) => human.delegateMention === wakeAgentId);
   const representative = linkedHumans.length === 1 ? linkedHumans[0] : linkedHumans.length === 0 ? humans[0] : null;
-  if (!representative || (linkedHumans.length === 0 && humans.length !== 1)) return null;
+  if (!representative) return null;
```
yuezengwu, line 136: **P1 — preserve every successful caller’s wake line.** With this fallback, two unlinked agent callers in the same multi-human chat both select the same `humans[0]`, while their distinct wake-agent IDs let both `(human, wake-agent)` mappings be created successfully. GitHub then collapses those rows by only `(humanAgentId, chatId)` in `github-audience.ts` and keeps one delegate before the delivery planner, so one successful follower is missing from `metadata.mentions` and is never woken. GitLab preserves both rows, so this also breaks the promised cross-provider contract.

Please preserve distinct wake agents through GitHub audience resolution—the per-chat planner already emits one card and unions wake IDs—and add a regression covering two unlinked agents in the same multi-human chat following the same entity: one card, both callers in the wake set.

yuezengwu (changes requested):
The goal and core change are clear: restore the shared multi-human/no-link representative fallback, keep unique-link precedence and multiple-link failure, and retain the declaring agent as the wake target without changing the schemas. The resolver and route changes themselves are narrow, but the enabled shape exposes one blocking GitHub audience bug described inline.

Verification on this exact head: the 37 targeted Server tests passed; Server typecheck, affected-file Biome checks, and `git diff --check` passed. I found no database or schema changes. Requesting changes for the wake-loss/parity blocker.

packages/server/src/services/scm-attention-line.ts
baixiaohang: Fixed in cb4767e9. GitHub subscribed-audience dedup now preserves each (human, wake-agent, chat) attention line, while the downstream per-chat planner still emits one card and unions wake IDs. Added a regression that resolves two unlinked agent callers to the same fallback human, stores both mappings, and verifies one card plus both mentions and notify=true Inbox entries. The related 66 tests, monorepo typecheck, affected Biome checks, and diff check pass.

packages/server/src/services/github-audience.ts
```diff
@@ -141,14 +141,16 @@ export async function resolveGithubAudience(db: Database, event: NormalizedScmEv
...
-  const earliestByHumanChat = new Map<string, (typeof subscribedRows)[number]>();
+  // all. The key also includes `delegateAgentId`: two agent-issued follows can
+  // share the same fallback human and chat while carrying distinct wake
+  // agents. "The chat follows, not the person", so the surviving unit is one
+  // row per (human, delegate, chat).
+  const earliestByAttentionLine = new Map<string, (typeof subscribedRows)[number]>();
   for (const row of subscribedRows) {
-    const key = `${row.humanAgentId}:${row.chatId}`;
-    const current = earliestByHumanChat.get(key);
+    const key = `${row.humanAgentId}:${row.delegateAgentId}:${row.chatId}`;
```
yuezengwu, line 150: **P1 — prevent an inactive sibling wake line from poisoning the shared card.** The new key correctly preserves distinct active callers, but the subscribed query does not validate the mapped delegate’s lifecycle. This state is reachable: A and B follow while active through the same fallback human/chat, then B is suspended; suspension leaves its mapping and chat membership. On this exact head, audience resolution retains A+B, the planner unions both IDs, and message preflight rejects the whole send because B is suspended. My repro returned `{ delivered: 0, newChats: 0, failed: 1 }`, wrote no card, and created no Inbox row for active A; the claimed webhook will not retry. At `57465868`, the earlier active A row survived the `(human, chat)` dedup, so this is a successor regression. Please preserve the valid chat delivery while excluding suspended/deleted wake IDs from native mentions (or enforce equivalent per-line eligibility), and add a mixed active+suspended regression proving one card and one wake for A, with no wake for B.

yuezengwu (changes requested):
The goal and core change remain sound: restore the no-link multi-human fallback while keeping unique-link precedence, multiple-link failure, and the declaring agent as the wake target without changing the mapping schema or database state. This successor correctly fixes the original two-active-caller wake-loss, so I resolved that prior thread.

However, exact head `cb4767e9` exposes a lifecycle blocker described inline: a stale suspended sibling wake line can reject the entire shared-chat card and permanently drop delivery for an active follower. I reproduced that state end to end. Otherwise, 63 focused Server tests, Server typecheck, affected-file Biome, and `git diff --check` passed. No database or schema changes found.

packages/server/src/services/github-audience.ts
baixiaohang: Fixed in `1d0ed5e3`. Trusted SCM card sends now drop suspended/deleted speaker targets from `metadata.mentions` inside the same `sendMessage` transaction snapshot, while retaining the attention line and card delivery. Ordinary Web/Agent sends keep the existing inactive-recipient fail-closed behavior. The new end-to-end regression covers two agent-issued lines sharing a fallback human/chat, suspends one follower after both mappings exist, and verifies one delivered card, only the active follower in persisted mentions/notify, no suspended Inbox notification, and `failed: 0`. A shared message-service test also locks the filtered metadata/recipient behavior; 49 focused GitHub/GitLab/message tests, Server typecheck, Biome, and diff-check pass.

yuezengwu (approved):
The goal and core change are now complete: restore the multi-human/no-link agent-follow fallback, preserve distinct wake lines under one human carrier, and keep one card per chat while waking only eligible agents. The `1d0ed5e3` successor fixes the lifecycle blocker at the shared trusted-SCM send boundary: suspended/deleted speakers are removed from persisted native mentions and all downstream wake projections, while the mapping and history card remain; ordinary Web/Agent sends still fail closed on inactive recipients.

I traced the exact successor diff through GitHub delivery, the generic message transaction, Inbox/session effects, and GitLab's deferred shared-card path. No remaining findings. There are no database, schema, or core storage-structure changes. This was a diff-only review; I did not rerun tests or formal QA. Draft Context Tree PR #791 remains the required post-code-merge reconciliation.

PR #121, opened 2026-04-21
Revert "docs: disambiguate breeze naming in agents guide" (#118)

bingran-you:
## Summary
- Reverts #118. The Hub-hosted Breeze surface doesn't exist in this repo yet — the only mention of "Breeze" here is the disambiguation block itself, which makes it self-referential with nothing to point agents at.
- Context Tree (`agent-hub/breeze.md`) already records the 2026-04-16 decision that Breeze will live in Hub; we can re-add the AGENTS.md pointer when real Hub-side Breeze code lands and can be referenced concretely.

## Test plan
- [x] Docs-only revert; no code paths touched

PR #1899, opened 2026-07-21
fix: bound /healthz database probes with a cached probe and route rate limit

### LINKED ISSUE #1716: [PERF-053][Medium] Public health checks query PostgreSQL without a limit
## English

- **Severity:** Medium
- **Location:** `packages/server/src/api/healthz.ts:10-15`
- **Estimated remediation:** 2-4 hours
- **Description:** `/healthz` explicitly disables rate limiting and executes `SELECT 1` per request. Public traffic or overly aggressive probes can create unlimited database round trips, including when the database is already unhealthy.
- **Recommended direction:** Split process liveness from database readiness, restrict readiness at ingress or give it an independent cheap limit/cache, and tune probe cadence.

## 中文：公共健康检查不受限地查询 PostgreSQL

- **严重性：** 中
- **位置：** `packages/server/src/api/healthz.ts:10-15`
- **预计修复时间：** 2-4 小时
- **描述：** `/healthz` 显式禁用了速率限制，并且每次请求都会执行 `SELECT 1`。公共流量或过于激进的探针可能造成不受限制的数据库往返，即使数据库本身已经处于不健康状态也会如此。
- **建议方向：** 将进程存活检查与数据库就绪检查拆分开；在入口层限制就绪检查，或为其提供独立且低成本的限流或缓存，并调整探针频率。

---

- **Audit date:** 2026-07-12
- **Source revision:** `ea2a9241fd8e128e89236cda2542f180731e3207`

yuezengwu:
# fix: bound /healthz database probes (PERF-053)

Fixes #1716

## Problem

`/healthz` opted out of rate limiting (`rateLimit: false`) and executed `SELECT 1` against PostgreSQL on **every** request. The endpoint is public (used by Docker HEALTHCHECK, Railway, Fly.io, Kubernetes probes), so public traffic or overly aggressive probes could create unlimited database round trips — including while the database was already down, where every request still attempted its own failing probe.

## Changes

Scoped to `packages/server/src/api/healthz.ts` plus tests; no contract, deployment, or config-schema changes.

- **Short-TTL cached probe with single-flight** (`healthz.ts`): the `SELECT 1` result — success *and* failure — is cached for 1s per process. Concurrent requests past expiry share one in-flight probe. Database round trips are bounded to **≤ 1 per second per process** regardless of request rate. The cache lives in the route-registration closure (no module-level state), so it is per-app-instance and test-isolated.
- **Independent route-level rate limit**: `rateLimit: false` → `{ max: 120, timeWindow: "1 minute" }`, matching the existing route-override pattern in `api/auth/google.ts` and `api/me-auth-providers.ts`.

### Why 120/min

Per the rate-limiting convention (limits must stay high enough that normal use never sees 429s): the fastest reasonable probe cadence is 1s (60/min); 120/min leaves 2x headroom over that, and is far above the actual consumers here — Dockerfile HEALTHCHECK polls every 30s (2/min), Kubernetes defaults to 10s (6/min per probe type). Anonymous traffic keys on client IP via the existing `keyGenerator`. The value comes from the audit finding itself (#1716), not a reintroduced low route default.

### Behavior notes

- Response contract is unchanged: `200 {status:"ok"}` / `503 {status:"error", message:"database unreachable"}`.
- Health signal can lag reality by at most 1s — negligible against 10s–30s probe periods, and a ≤1s stale *failure* expires long before the next 30s Docker HEALTHCHECK tick.
- A 429 from the limiter reads as "unhealthy" to probe clients; with the headroom above, legitimate probes never approach it.
- The fix is in-process only: no Redis/MQ/shared limiter state, keeping the server stateless. With N replicas each process probes ≤ 1/s independently — bounded and acceptable.
- **Scope:** the issue title says "health checks" (plural). `/readyz` also sets `rateLimit: false` but performs no database access (pure in-memory bootstrap-stage read), and `/api/v1/health` is already covered by the global limiter — neither is in the issue's stated location (`healthz.ts:10-15`), so both are intentionally untouched.
- Endpoint split (the issue's first suggested direction) was deliberately not taken: `/healthz` = process + DB reachability is a deployed contract consumed by the Dockerfile HEALTHCHECK, and `/readyz` already fills the no-DB orchestration-probe role. The issue explicitly allows the cheap-limit/cache alternative, which carries far less behavioral risk.

## Testing

- `packages/server/src/__tests__/healthz.test.ts` (new): TTL cache hit/miss, failure caching + recovery after expiry, single-flight concurrency (one probe for concurrent expired requests), and a full-app integration test asserting request #120/min returns 200 and #121 returns 429.
- Adapted `api-small-routes-extra.test.ts` healthz case to advance past the probe TTL before asserting the degraded response.
- `pnpm --filter @first-tree/server test`, `pnpm check`, `pnpm typecheck` — all green.

## QA

This touches the boot/health path. Existing QA case `packages/qa/cases/cross-surface/release-boot-health.md` covers this surface; its PASS criteria (200 / ready / db-connected payload shape) are unaffected by this change. Formal QA per `skills/first-tree-qa` is warranted before merge.

### DIFF
--- a/packages/server/src/__tests__/api-small-routes-extra.test.ts
+++ b/packages/server/src/__tests__/api-small-routes-extra.test.ts
@@ -321,27 +321,35 @@ describe("small API route handlers", () => {
   });
 
   it("returns health and healthz success and degraded responses", async () => {
-    const { healthRoutes } = await import("../api/health.js");
-    const { healthzRoutes } = await import("../api/healthz.js");
-    const execute = vi.fn().mockResolvedValue(undefined);
-    const { app, routes } = makeApp({ db: { execute } });
+    vi.useFakeTimers();
+    try {
+      const { healthRoutes } = await import("../api/health.js");
+      const { healthzRoutes } = await import("../api/healthz.js");
+      const execute = vi.fn().mockResolvedValue(undefined);
+      const { app, routes } = makeApp({ db: { execute } });
 
-    await healthRoutes(app as never);
-    await healthzRoutes(app as never);
-    await expect(route(routes, "GET", "/health").handler()).resolves.toEqual({ db: "connected", status: "ok" });
+      await healthRoutes(app as never);
+      await healthzRoutes(app as never);
+      await expect(route(routes, "GET", "/health").handler()).resolves.toEqual({ db: "connected", status: "ok" });
 
-    const okReply = makeReply();
-    await route(routes, "GET", "/healthz").handler({}, okReply);
-    expect(okReply).toMatchObject({ body: { status: "ok" }, code: 200 });
+      const okReply = makeReply();
+      await route(routes, "GET", "/healthz").handler({}, okReply);
+      expect(okReply).toMatchObject({ body: { status: "ok" }, code: 200 });
 
-    execute.mockRejectedValueOnce(new Error("db down")).mockRejectedValueOnce(new Error("db down"));
-    await expect(route(routes, "GET", "/health").handler()).resolves.toEqual({
-      db: "disconnected",
-      status: "degraded",
-    });
-    const degradedReply = makeReply();
-    await route(routes, "GET", "/healthz").handler({}, degradedReply);
-    expect(degradedReply).toMatchObject({ body: { message: "database unreachable", status: "error" }, code: 503 });
+      execute.mockRejectedValueOnce(new Error("db down")).mockRejectedValueOnce(new Error("db down"));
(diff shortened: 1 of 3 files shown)


yuezengwu:
GoF submission note for maintainers: this draft PR is the Kimi team's entry for #1716 (fire competition). Both team reviews (dev + reviewer) have passed, CI is green, and the label is now `fire_submitted`. Per competition rules the PR intentionally stays in draft — marking ready is reserved; merge remains entirely at maintainer discretion. No action needed beyond a normal maintainer review whenever convenient.

PR #1901, opened 2026-07-21
feat: add composer live activity inspector

Gandy2025:
## Summary

- replace the variable composer activity rail with a stable lead snapshot and an explicit `Activity (N)` entry
- add a shared live activity Inspector with concise per-agent snapshots and timeline evidence jumps
- keep the server `/agent-status` projection authoritative; timeline events and status reasons only explain that state
- make the Inspector keyboard-, screen-reader-, reduced-motion-, and mobile-accessible, including layered Escape and live-list focus recovery
- load all non-human speaker evidence through one chat-scoped batch endpoint, including private speakers visible only through that chat

## Why

The previous composer rail duplicated timeline narration, changed height as activity changed, and did not provide a reliable path from a compact status cue to the underlying execution evidence. The Inspector keeps the composer stable while preserving drill-down into the timeline.

The review follow-up also removes every remaining event-derived Working promotion. A stale or unfinished trace can no longer override the server-authoritative per-chat runtime state or enable Pause.

## Permission boundary

The product/security decision is that a user who can access a chat, including a watcher, may inspect every retained runtime-event kind for non-human speakers within that chat's bounded current/recent evidence window. This is not a full-history or audit entitlement: it does not require older evidence to be retained, paginated, or exposed, and session evidence remains clearable under the existing lifecycle. It also does **not** make a private agent discoverable outside the chat.

`GET /chats/:chatId/session-events` therefore:

- uses the existing exact-chat access gate
- re-filters event owners to current non-human speakers on the server
- requires the speaker agent and chat to belong to the same organization
- applies the event limit independently per speaker in one windowed query

The current Web request loads the newest 200 events per speaker; that is an implementation window, not a promise of older or permanent history. Regression coverage includes private speakers, speaker and watcher viewers, an exact-chat outsider returning 404, and a forged event from a non-speaker being excluded.

## Impact

- users get a stable composer with one compact entry for multi-agent activity
- keyboard and screen-reader users can enter, understand, and exit the Inspector with predictable focus behavior
- every authorized secondary-agent summary can point to real timeline evidence without an N-query fanout
- read-only watchers can inspect chat-scoped activity without gaining reply permissions or org-level private-agent discovery

## Durable decision tracking

The durable permission, retention-window, and membership boundary is tracked in [first-tree-context#794](https://github.com/agent-team-foundation/first-tree-context/pull/794). That Context Tree PR remains Draft until this code PR is merged, then it will be reconciled against the merged implementation before becoming ready.

## Validation

- `pnpm check` — passed (repository-existing warnings/info only)
- `pnpm typecheck` — 10/10 tasks passed, including Web/Server builds and the design-token guardrail
- focused Web review suites — 8 files / 102 tests passed
- Server status and event suites — 2 files / 67 tests passed
- full Web Vitest assertions — 1554/1554 passed; the command still exits 1 because the existing Playwright-only `tests/visual-regression.spec.ts` is collected as an empty Vitest suite
- GitHub CI and CodeQL — passed at head `6eb8cc3f972bc48d05658a82b2caae650cbc8e6a`


Gandy2025:
@yuezengwu Could you please review this PR? The implementation includes the follow-up fixes from the architecture, UX, and accessibility review: server-authoritative status semantics, Inspector focus and screen-reader behavior, and secondary-agent timeline evidence.

packages/web/src/pages/workspace/center/chat-view.tsx
```diff
@@ -2532,47 +2527,57 @@ export function ChatView({
...
-        if (existing) existing.push(e);
-        else turnEventsByAgent.set(e.agentId, [e]);
-      } else {
-        // error rows stay standalone and visible across turns.
-        out.push({ kind: "event", at: e.createdAt, key: `e-${e.id}`, data: e });
+    // A seq is monotonic only within one (agent, chat) session, so combine the
+    // independently fetched feeds by agent before finding turn boundaries.
+    // Dedup by event id protects a temporarily duplicated participant/query.
+    const rawEventsByAgent = new Map<string, Map<string, SessionEventRow>>();
+    for (const feed of [...(eventsData ? [eventsData] : []), ...secondaryEventFeeds]) {
```
yuezengwu: **[P1 / blocker] Keep event evidence from re-deriving Working.** These secondary feeds become workgroups solely from events after the latest `turn_end`, and every such workgroup then enters `liveTurnAgentIds`. `AgentStatusPanel` and `AgentHovercard` still call `reconcileLiveTurn`, so a stale or un-ended secondary trace upgrades an authoritative `{ main: "ready", working: false }` back to Working and can even enable Pause, while the composer/header remain Idle. Remove the event-derived status promotion from the remaining consumers (and gate any Working presentation on the server projection), then cover `ready` + un-ended secondary evidence as Idle/no-Pause.

packages/web/src/pages/workspace/center/chat-view.tsx
```diff
@@ -1835,6 +1826,30 @@ export function ChatView({
...
+    () =>
+      (chatDetail?.participants ?? [])
+        .filter((participant) => participant.type !== "human" && participant.agentId !== agentId)
+        .map((participant) => participant.agentId),
+    [agentId, chatDetail?.participants],
+  );
+  const secondaryEventFeeds = useQueries({
+    queries: secondaryTimelineAgentIds.map((secondaryAgentId) => ({
+      queryKey: ["session-events", secondaryAgentId, chatId] as const,
+      queryFn: () => listSessionEvents(secondaryAgentId, chatId, { limit: 200, direction: "desc" }),
```
yuezengwu: **[P1 / permissions blocker] This route cannot supply evidence for every legitimate chat participant.** `/agents/:id/sessions/:chat/events` first requires org-level `requireAgentAccess(..., "visible")`, while chat detail intentionally exposes member identity from `chat_membership` even for another member’s private agent. A normal chat member therefore gets 404 here and `combine` silently drops the feed, leaving Inspector status without timeline evidence. This permission boundary needs human product/security confirmation. Please use a chat-scoped, speaker-filtered events contract (preferably batched—this currently creates N immediately-stale 200-row queries) rather than widening global private-agent visibility, and add a private-agent chat-member regression test.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,364 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
+    });
+    return () => cancelAnimationFrame(frame);
+  }, [inspectorOpen]);
+
+  useEffect(() => {
+    if (!inspectorOpen) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Escape") return;
+      event.preventDefault();
+      event.stopPropagation();
```
yuezengwu, line 200: **[P1 / accessibility blocker] One Escape dismisses both Activity and Chat details.** `stopPropagation()` does not suppress another listener registered on the same `document`; ChatView’s open-sidebar Escape listener also runs and persists `setSidebarByUser(false)`. With Details and Activity open, Escape should dismiss only the top layer. Please use layered dismissal/overlay handling (or otherwise make the outer rail ignore the Inspector’s Escape) and test this combined state.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,364 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
-      setCardOpenFor(null);
-    }
-  }, [cardOpenFor, canExpand, leadRow?.agentId]);
+    if (activeChatIdRef.current === chatId) return;
+    activeChatIdRef.current = chatId;
+    setInspectorOpen(false);
+  }, [chatId]);
+
+  useEffect(() => {
+    if (attention.length === 0) setInspectorOpen(false);
```
yuezengwu: **[P2 / accessibility] Preserve focus when live rows disappear.** A status refresh can remove the currently focused keyed Inspector row; when the last item quiets, line 178 removes the entire bar and trigger. In both cases focus falls to `body`, so keyboard/screen-reader users lose their position. Move focus to a surviving row or Close control when the list shrinks, and to a stable composer control when the list becomes empty; add live-update focus coverage.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,364 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
-      <span className="truncate" title={detail}>
-        {reason.label}
-      </span>
-    );
-  }
-  if (status.main === "failed") return <span className="truncate">failed</span>;
-  return <WorkingDetail activity={status.activity} onGoalOverflowChange={onGoalOverflowChange} />;
+  const visual = statusVisual(status);
+  const snapshot = agentSnapshot(status);
+  const anchored = isJumpable(mounted, status.main, status.agentId);
```
yuezengwu: **[P2] Terminal-error evidence is mounted but not jumpable.** The server deliberately supports `main: "ready"` plus an error-severity terminal reason, and its fatal timeline row mounts `data-error-agent`. `isJumpable(..., "ready", ...)` only looks for a ready anchor (which does not exist), so “Provider retry exhausted · Idle” becomes static despite exact evidence being present. Derive the jump target from the fatal reason/evidence (use the error/failed anchor) and cover ready + terminal-error + mounted-error.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,364 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
-            triggerRef: cardTriggerRef,
+      <div className="flex min-w-0 items-center" style={{ gap: "var(--sp-2)" }}>
+        <LeadSnapshot status={leadRow} name={nameOf(leadRow.agentId)} />
+        <button
+          ref={triggerRef}
+          type="button"
+          aria-controls={inspectorId}
+          aria-expanded={inspectorOpen}
+          aria-haspopup="dialog"
+          aria-label={inspectorOpen ? "Close agent activity" : "Open agent activity"}
```
yuezengwu: **[P2 / accessibility] Expose the actionable count and transition.** This explicit `aria-label` overrides the visible `Activity (N)` text, so assistive tech only hears “Open agent activity,” and the surface’s appear/fail/wait transitions have no status/live-region announcement. Include the count in the accessible name and provide a concise, deduplicated polite announcement for actionable state changes (without narrating every tool update).

yuezengwu (changes requested):
## Review summary

**Goal and core changes:** this PR replaces the composer’s ad-hoc expanded rail with a stable lead snapshot plus a shared Activity Inspector, adds keyboard/screen-reader behavior, keeps the composer on the server `/agent-status` projection, and loads secondary-agent session events so every Inspector summary can point to timeline evidence.

I’m requesting changes because the follow-up is not server-authoritative end to end yet, the secondary evidence path crosses an unresolved private-agent permission boundary, and Escape can dismiss two UI layers at once. I left the concrete blockers and regression cases inline, plus three functional/accessibility gaps in the new Inspector.

There is no DB/schema or core persisted-data-structure change in this PR. The secondary-event authorization behavior *is* a user/permission boundary and should receive human product/security review before implementation; please keep any fix chat-scoped rather than broadening org-level private-agent discovery.

Validation at head `379bcc6e8366f175db7ab0fbb8bbbf9658440d9b`:
- focused server status suite: 44/44 passing
- focused web suites: 77/77 passing (with the existing mocked localhost connection noise)
- CI currently green
- desktop and 390×844 preview inspection: layout fits; focus enters the Inspector, Tab reaches the first item, and Escape restores the trigger in the simple case

packages/web/src/pages/workspace/center/chat-view.tsx
Gandy2025: Fixed in 79ceecde3. Removed LiveTurnAgentsContext and reconcileLiveTurn from every remaining consumer. Roster, hovercard, composer, and Pause eligibility now read only the server chat-status projection. Regression coverage keeps ready plus residual un-ended evidence Idle with no Pause.

packages/web/src/pages/workspace/center/chat-view.tsx
Gandy2025: Fixed in 79ceecde3 after explicit product/security confirmation. Anyone with access to the exact chat, including watchers, may inspect the complete runtime trace for non-human speakers in that chat; private-agent discovery remains unchanged outside it. The Web now makes one chat-scoped batch request. The server applies requireChatAccess, re-filters owners to current same-org non-human speakers, and uses a per-agent window instead of N queries. The regression covers private speakers, speaker and watcher access, exact-chat outsider 404, and forged non-speaker exclusion.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 79ceecde3. Activity handles Escape in the capture phase, prevents and stops the event immediately, and ChatView ignores already-handled Escape events. The integration test opens Chat details and Activity together and verifies one Escape closes only Activity.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 79ceecde3. A disappearing focused row moves focus to the Inspector Close control; when the final actionable row disappears, focus moves to the stable composer control, including the read-only fallback. The live-update test covers both transitions.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 79ceecde3. A fatal terminal status reason now derives a failed/error evidence target while preserving the server main state as ready/Idle. The regression verifies that the row is actionable and jumps to the mounted error anchor.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 79ceecde3. The trigger accessible name includes the actionable count, and a polite atomic status region announces deduplicated agent/count state transitions. The announcement excludes narration and tool metadata so tool churn does not produce repeated speech.

Gandy2025:
@yuezengwu Thanks for the detailed review. All six threads are addressed in 79ceecde3, including the explicitly confirmed chat-scoped full-trace permission boundary. I replied with the fix and regression evidence on each thread and resolved them. CI is running on the new head; please re-review when ready.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,422 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
+    return () => cancelAnimationFrame(frame);
+  }, [inspectorOpen]);
+
+  useEffect(() => {
+    if (!inspectorOpen) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Escape") return;
+      event.preventDefault();
+      event.stopPropagation();
+      event.stopImmediatePropagation();
```
yuezengwu, line 201: **[P1 / accessibility blocker] Do not capture Escape ahead of a later keyboard layer.** This non-modal Inspector stays open when focus Tabs out of it. In a group chat, Tab into the textarea auto-primes `@` and opens MentionAutocomplete; the next Escape reaches this document-capture listener first, `stopImmediatePropagation()` prevents the textarea's `mention.handleKey`, closes the underlying Inspector, refocuses its trigger, and leaves the listbox open with focus elsewhere. Scope Escape to the Inspector/trigger or use a real top-layer coordinator, and cover Inspector → Tab to composer → mention opens → Escape closes mention first.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,422 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
-      <span className="truncate" title={detail}>
-        {reason.label}
-      </span>
-    );
-  }
-  if (status.main === "failed") return <span className="truncate">failed</span>;
-  return <WorkingDetail activity={status.activity} onGoalOverflowChange={onGoalOverflowChange} />;
+  const visual = statusVisual(status);
+  const snapshot = agentSnapshot(status);
+  const jumpMain = isFatalStatusReason(status) ? "failed" : status.main;
```
yuezengwu: **[P1 / functional blocker] Keep every reason summary connected to its mounted evidence.** Only fatal terminal reasons map to `failed`. The server explicitly supports `ready` + `retrying`/`waiting` (and terminal warnings), while `useMountedAnchors` only knows `working`/`failed` and the matching nonfatal provider rows omit `data-error-agent`. Those Inspector items therefore remain static even when their exact timeline row is mounted, which breaks the summary→evidence contract. Add a reason-event anchor/jump mapping and cover ready + retrying/waiting + mounted provider evidence.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,422 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
+        (inspectorRef.current?.contains(active) && (activeAgentId === undefined || focusedRowSurvives)))
+    ) {
+      return;
     }
-  }, [cardOpenFor, canExpand, leadRow?.agentId]);
+    const frame = requestAnimationFrame(() => {
+      inspectorRef.current?.querySelector<HTMLElement>(".compose-status-inspector-close")?.focus();
+    });
+    return () => cancelAnimationFrame(frame);
+  }, [attention, fallbackFocusRef, inspectorOpen]);
```
yuezengwu: **[P2 / accessibility] Re-run focus recovery when row interactivity changes.** `mounted` can change while `attention` stays identical. A `turn_end` can remove the WorkingTurn anchor before the server runtime projection flips from Working; `TimelineJumpButton` then replaces the focused button with a div, focus falls to `body`, and this effect does not run because `mounted` is absent from its dependencies. Include anchor/actionability changes in the recovery path and test a focused row losing only its anchor.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,422 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
-          style={{ gap: "var(--sp-1)", maxHeight: EXPANDED_MAX_HEIGHT, overflowY: "auto" }}
-        >
-          {/* `others`, NOT `attention`: the lead already has its own always-on
-              row above, so mapping the full set here would render the lead a
-              second time (the duplicate-row bug). */}
-          {others.map((s) => (
-            <RailRow key={s.agentId} status={s} nameOf={nameFor(agents)} mounted={mounted} />
-          ))}
-        </div>
+      <span role="status" aria-live="polite" aria-atomic="true" className="sr-only">
```
yuezengwu, line 221: **[P2 / screen reader] Keep the live region mounted across the quiet state.** The early `attention.length === 0` return removes this node, so 0→1 inserts an already-populated `role=status` (not reliably announced) and 1→0 removes it before the computed zero-count message can commit. Keep an empty, rendered status region present and update its text for quiet↔active transitions; this follows the [WAI-ARIA live-region guidance](https://www.w3.org/TR/wai-aria-1.3/#dfn-live-region).

yuezengwu (changes requested):
## Re-review summary

**Goal and core changes:** this PR replaces the variable composer activity rail with a stable lead snapshot and shared Activity Inspector, keeps the server `/agent-status` projection authoritative, loads all non-human speaker evidence through one chat-scoped batch endpoint, and adds keyboard/screen-reader focus behavior.

The previous status-authority and private-speaker permission blockers are fixed in implementation: event evidence no longer promotes Working or enables Pause, and the new endpoint uses exact-chat access plus current same-org non-human speaker filtering with an independent per-agent window. The layered Details + Activity case is also covered.

I am keeping changes requested because the capture-phase Escape fix now steals Escape from a later keyboard-opened layer, and waiting/retrying/terminal-warning summaries still cannot jump to their mounted provider evidence. I left those two blockers inline, plus two remaining dynamic focus/live-region accessibility gaps.

There is no DB/schema or persisted core-data-structure change. The new endpoint does define a user/permission boundary—every viewer who can access the chat, including a watcher, receives the full non-human speaker trace. The implementation matches the permission decision stated in the PR; please retain the human product/security sign-off for that boundary.

Review at head `79ceecde3a42e579c27dbd6ee63fb785b014a2a9` was diff-only as requested; I did not run tests or QA, and did not wait on CI.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 0c2e91dc0. Provider waiting, retrying, and terminal-warning rows now expose a dedicated reason evidence anchor; mounted-anchor tracking and timeline scrolling support that target. Fatal reasons and server Failed states continue to use the fatal error anchor. Regression coverage verifies ready plus waiting, retrying, and terminal-warning summaries jump to reason evidence, and ChatView confirms a non-fatal provider row mounts the reason anchor without becoming an error anchor.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -89,327 +70,438 @@ function isFatalStatusReason(status: AgentChatStatus): boolean {
...
+  }, [inspectorOpen]);
+
+  useEffect(() => {
+    if (!inspectorOpen) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Escape") return;
+      const target = event.target;
+      if (
+        !(target instanceof Node) ||
+        (!inspectorRef.current?.contains(target) && !triggerRef.current?.contains(target))
```
yuezengwu, line 195: **[P2 / keyboard follow-up, non-blocking] Complete the next Escape in the layer sequence.** The first textarea Escape now correctly reaches `mention.handleKey` and closes the listbox. Focus stays in the textarea, though, and this target-only guard also ignores the second Escape after Mention is no longer open; textarea handling falls through and Chat details intentionally ignores Escape in editable controls, so Activity remains open until the user navigates focus back to it. Consider an unclaimed bubbled-Escape path or shared layer coordinator, and extend the integration case with a second Escape.

yuezengwu (approved):
## Re-review summary

**Goal and core changes:** this PR replaces the variable composer activity rail with a stable lead snapshot and shared Activity Inspector, keeps the server `/agent-status` projection authoritative, loads all non-human speaker evidence through one exact-chat batch endpoint, and provides keyboard, screen-reader, reduced-motion, mobile, and timeline-evidence behavior.

The `79ceecde..6eb8cc3f` follow-up resolves the four remaining review findings: Escape no longer steals the mention layer's first dismissal; waiting, retrying, and terminal-warning summaries use a dedicated mounted reason-evidence target; focus recovers when only row actionability disappears; and one live region remains mounted across quiet/active transitions. The corresponding production paths and focused regression cases agree, so I am approving this head.

I left one non-blocking keyboard follow-up inline: after Mention consumes the first Escape, an additional Escape from the still-focused textarea does not dismiss Activity. This does not recreate the broken focus/listbox state from the blocker, but a true layer coordinator or unclaimed bubbled-Escape path would complete the one-layer-per-key sequence.

There is no DB/schema or persisted core-data-structure change in this delta. The previously reviewed permission boundary is unchanged: exact-chat viewers, including watchers, can inspect full non-human speaker trace evidence without org-wide private-agent discovery. Its durable record remains in Draft Context Tree PR `first-tree-context#794` pending code merge/reconciliation.

Review at exact head `6eb8cc3f972bc48d05658a82b2caae650cbc8e6a` was diff-only as requested; I did not run tests or QA and did not wait on CI.

Gandy2025:
## Product/security boundary confirmation

Gandy2025 confirmed that "complete process" means **all retained runtime-event kinds within the bounded current/recent evidence window for each non-human speaker**, not a full-history or audit entitlement.

Exact-chat viewers, including watchers, may inspect that bounded evidence. The product does not promise retention, pagination, or disclosure of older evidence outside the window; normal session termination/archive clearing semantics remain in force. Only speakers may send, address participants, or receive speaker delivery, and this chat-scoped evidence access does not make a private agent discoverable elsewhere.

The current newest-200-events-per-speaker implementation conforms to this boundary. The PR description and Draft Context Tree PR #794 have been updated to record it explicitly.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1901

**Recommendation**: approve
- Rationale: The change consolidates live status authority on the server, keeps runtime evidence chat-scoped, and replaces the prior event-derived status path without leaving a parallel source of truth.

**Risk level**: B-high
- Path baseline: changes `packages/server/src/services/session-event.ts`, a session service.
- Semantic lift: adds a chat-scoped runtime-evidence permission boundary and changes the Web data/status path.

**PR summary**
- Problem: Multi-agent chats had a variable-height composer activity rail, duplicated timeline narration, and no consistent way for participants or watchers to inspect the evidence behind compact status cues.
- Approach: Replaces the rail with a stable lead snapshot and shared Activity Inspector, keeps `/agent-status` authoritative for Working/Failed/Pause, and loads each current non-human speaker's bounded recent evidence through one exact-chat batch endpoint.
- Impacted modules: Server chat/session-event reads; Web composer, timeline, status surfaces, WebSocket invalidation, focus/accessibility helpers.

**Review findings**
✅ 1. Server status is now the single authority; the old `LiveTurnAgentsContext` / `reconcileLiveTurn` promotion path is removed from all consumers.
✅ 2. Runtime evidence is filtered by exact-chat access, current speaker membership, non-human type, and same-organization ownership; watcher access remains observational.
✅ 3. The Inspector centralizes summary, focus recovery, live-region, reduced-motion, and working/failed/reason evidence-jump behavior with focused regression coverage.
⚠️ 4. The remaining second-Escape behavior after Mention dismissal is a non-blocking layering follow-up and does not recreate the broken listbox/focus state.

**Action taken**
- Approved at exact head `6eb8cc3f972bc48d05658a82b2caae650cbc8e6a` after explicit authorization from baixiaohang.

PR #1903, opened 2026-07-21
fix: validate agent names and enforce realpath containment in removeLocalAgent

### LINKED ISSUE #1636: [SEC-030][High] `agent remove` permits recursive path traversal
### SEC-030 - `agent remove` permits recursive path traversal

- **Severity:** High
- **Location:** `apps/cli/src/commands/agent/remove.ts:14-20`, `apps/cli/src/core/agent-prune.ts:144-147`
- **Estimated remediation:** 2-4 hours
- **Description:** The user-supplied agent name is joined into three recursively deleted paths without using the canonical name schema or verifying resolved containment. A traversal name can delete writable files outside First Tree state.
- **Recommended direction:** Validate the canonical agent-name schema and enforce realpath containment immediately before every deletion.

### SEC-030 - `agent remove` 存在递归删除路径遍历漏洞

- **严重性：** 高
- **文件位置：** `apps/cli/src/commands/agent/remove.ts:14-20`、`apps/cli/src/core/agent-prune.ts:144-147`
- **预计修复时间：** 2-4 小时
- **描述：** 用户提供的 agent 名称会被拼接到三个递归删除路径中，但代码既没有使用规范名称 schema，也没有验证解析后路径是否仍位于预期目录内。包含路径遍历的名称可以删除 First Tree 状态目录之外的可写文件。
- **建议修复方向：** 按规范的 agent 名称 schema 进行验证，并在每次删除前立即强制校验 realpath 是否位于预期目录内。

Audit date: 2026-07-12  
Source revision: `ea2a9241fd8e128e89236cda2542f180731e3207`

liuchao-001:
## Summary

Fixes [SEC-030] `first-tree agent remove` recursive path traversal.

`removeLocalAgent` (`apps/cli/src/core/agent-prune.ts`) deleted three paths built by joining raw user input — `config/agents/<name>` and `data/workspaces/<name>` with `{recursive: true, force: true}`, and `data/sessions/<name>.json` with `{force: true}`. `node:path.join` normalizes `..` segments, so `agent remove '../../../../<dir>'` recursively deleted arbitrary user-writable directories outside First Tree state, and the `.json` template deleted arbitrary `.json` files. The `existsSync` pre-check in `remove.ts` ran on the same traversed path, so it passed whenever the escape target existed. Both `agent remove` and `agent prune` go through `removeLocalAgent`, so both paths are covered by the fix.

## Fix — two independent layers, both inside `removeLocalAgent`

1. **Deletion-side name whitelist.** `assertRemovableAgentName` rejects any name not matching `/^[a-z0-9_-]{1,100}$/`. The character set contains no `/`, `\` or `.`, so a valid name can never form `.`, `..`, an absolute path, or any traversal segment. It is intentionally wider than the create-side `AGENT_NAME_REGEX` (`/^[a-z0-9][a-z0-9_-]{0,63}$/`): the server grandfathers rows created under the previous 1–100 rule, and `migrateLocalAgentDirs` renames local dirs to those server-authoritative names on every client startup, so grandfathered agents (e.g. `_legacy`, 65–100 chars) must stay removable. Reserved names are deliberately not blocked here — that list guards creates; blocking deletes would only produce undeletable junk dirs. The create-side rules are untouched.
2. **Realpath containment before every deletion.** Each target must resolve — lexically and via `realpathSync` — inside its base dir, otherwise removal aborts with `Refusing to remove "<name>": resolves outside First Tree state`. `rmSync` keeps operating on the original candidate path, never the realpath result, so an alias symlink is unlinked in place — an alias pointing at another directory *inside* the base cannot delete that directory. A dangling alias symlink (`existsSync` follows links) is containment-checked as parent realpath + own basename and stays removable. A base dir that does not exist yet (fresh install) is skipped without affecting the other targets. All targets are containment-validated **before the first `rmSync`**, and the config alias dir is deleted **last**, so a refused or failed later target cannot strand workspace/session state behind an already-removed alias — the alias dir is what `agent prune` discovers and retries by.

`agent remove` runs the same validation before any filesystem access and exits 1 with a readable error (no stack trace). `agent prune` needs no code changes: its per-alias try/catch reports a refused entry inline as failed (exit code 1) and keeps pruning the rest. The stale `removeLocalAgent` doc comment referencing a no-longer-existing third caller is corrected.

## Threat-model accuracy

The realpath layer is **not** a fix for a currently reachable symlink-follow deletion: Node's `rm`/`rmSync` treat symlinks with lstat semantics — the link itself is removed, never followed. The layer exists because the issue explicitly requires realpath containment before every deletion, as defense in depth against future changes to the deletion implementation, and as a conservative fail-closed stance on aliases resolving outside First Tree state.

## Known behavior change

Removing an alias symlink whose target exists **outside** First Tree state previously deleted the link and succeeded; it is now refused (fail closed). In `agent prune`, such an entry is reported as failed (exit code 1) while the remaining stale entries are still removed. Dangling symlinks and symlinks pointing inside the base behave exactly as before.

Note (pre-existing Commander behavior, unchanged): grandfathered names starting with `-` need `agent remove -- -legacy` to parse as a positional argument.

## Tests

- `agent-prune.test.ts` — new `removeLocalAgent path safety` suite: traversal names resolving to real outside dirs / outside session `.json` targets rejected with sentinels intact; empty / `.` / `..` / separator / absolute / uppercase / dotted / >100-char names rejected without touching state; grandfathered names (`_legacy`, `-legacy`, 100 chars) still remove all three footprint entries; dangling alias symlink still removed; alias→real symlink removal leaves the real dir intact; outward symlink refused with target preserved; valid names stay idempotent with missing bases/targets; preflight — an unsafe last-iterated target leaves the real workspace/session state validated before it untouched.
- `agent-prune-command.test.ts` (new) — `agent prune` integration over the **real** `removeLocalAgent`: the outward-symlink entry fails inline, exit code is 1, remaining stale entries are still pruned, the link target is untouched.
- `cli-command-matrix-extra.test.ts` — `agent remove '../../x'` and `agent remove ''` exit 1 with `Invalid agent name` and never reach `removeLocalAgent` (the name validator now runs unmocked in the command-layer suites).

## Verification

- `pnpm check` ✅ · `pnpm typecheck` ✅ (10/10)
- CLI package: all touched suites green (53 tests). `client-switch-transaction-extra.test.ts` fails identically on unmodified `main` in my environment — its real-process scan finds the actual First Tree daemon running on this machine — unrelated to this change.
- Standalone: `@first-tree/server` 2557/2557 ✅ · `@first-tree/shared` 645/645 ✅ · `@first-tree/client` ✅ · `@first-tree/qa` ✅. One `@first-tree/skill-evals` git-fixture test is timing-flaky here (a different file fails per run; each passes standalone) — unrelated.
- QA self-check per repo guidelines: local-FS CLI behavior only, no cross-surface/runtime/WS/boot path touched → unit + command-matrix + command-integration coverage suffices; no `packages/qa` case needed.

## Post-submission update (review feedback)

`71ef115` adopts the preflight point from maintainer-side review: all targets are containment-validated before the first `rmSync`, and the config alias dir is deleted last (see Fix §2). Locked by a new test in which the unsafe target iterates last and the real workspace/session state validated before it must survive the refusal. Re-verified: `pnpm check` / `pnpm typecheck` green, all four touched test files pass (54 tests).

Two review points intentionally not taken, for the record:

- **Hard-anchoring managed regions to their canonical lexical location.** Relocating a managed region via symlink (e.g. `data/workspaces` → larger disk) is operator-owned configuration the rest of the CLI already honors: `agent add`/`create` write through it, enumeration and migration read through it. Removal refusing what creation honors would make such state permanently unremovable. It is also not attacker-reachable input: replacing `$FIRST_TREE_HOME/config/agents` (or an ancestor) with a symlink already requires write access to First Tree state, at which point no deletion primitive is needed. The boundary this PR enforces is that the *name* — the only untrusted input — cannot select a target outside whatever the managed regions resolve to. Exact-location anchoring would additionally break environments whose home/tmp ancestors are symlinks (macOS `/var` → `/private/var`).
- **Moving the deletion-side name rule into a shared, named schema.** Agreed in principle; deliberately kept local to keep this security fix minimal (the same call as the two existing private `pathIsInside` copies). Tracked under Follow-ups.

## Follow-ups (out of scope)

- Extract the (now three) verbatim `pathIsInside` copies into a shared helper.
- Consider a shared home for name/path validation if the server ever performs local deletions.

Closes #1636

### DIFF
--- a/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
+++ b/apps/cli/src/__tests__/agent-lifecycle-commands-extra.test.ts
@@ -77,7 +77,12 @@ vi.mock("@first-tree/client", () => clientSdkMocks);
 vi.mock("@first-tree/shared/config", () => configMocks);
 vi.mock("../core/bootstrap.js", () => bootstrapMocks);
 vi.mock("../core/cli-fetch.js", () => ({ cliFetch: cliFetchMock }));
-vi.mock("../core/index.js", () => coreMocks);
+vi.mock("../core/index.js", async () => {
+  // `agent remove` imports the real name validator; the commands under test
+  // here only ever pass valid names, so wire it through unmocked.
+  const prune = await vi.importActual<typeof import("../core/agent-prune.js")>("../core/agent-prune.js");
+  return { ...coreMocks, assertRemovableAgentName: prune.assertRemovableAgentName };
+});
 vi.mock("../cli/output.js", () => outputMocks);
 vi.mock("../core/output.js", () => ({
   print: { line: printLineMock, result: outputMocks.success, fail: outputMocks.fail },
--- a/apps/cli/src/__tests__/agent-prune-command.test.ts
+++ b/apps/cli/src/__tests__/agent-prune-command.test.ts
@@ -1,0 +1,95 @@
+import { existsSync, mkdirSync, mkdtempSync, readdirSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { registerAgentPruneCommand } from "../commands/agent/prune.js";
+import { print } from "../core/output.js";
+
+/**
+ * `agent prune` over the REAL `removeLocalAgent` (unlike the lifecycle
+ * suite, which mocks it): a stale alias symlink that resolves outside First
+ * Tree state must fail closed — reported inline, non-zero exit code — while
+ * the remaining stale entries are still removed and the symlink's target
+ * stays untouched.
+ */
+
+const sdkMocks = vi.hoisted(() => ({ listMyAgents: vi.fn() }));
+
+vi.mock("@first-tree/client", async (importActual) => ({
+  ...(await importActual<typeof import("@first-tree/client")>()),
+  FirstTreeHubSDK: class {
+    listMyAgents = () => sdkMocks.listMyAgents();
+  },
+}));
+
+vi.mock("../commands/_shared/local-agent.js", () => ({
+  readClientId: () => "client_self",
+}));
+
+describe("agent prune with real removeLocalAgent", () => {
+  let home: string;
+  let outside: string;
+  let printLines: string[];
+  let printSpy: ReturnType<typeof vi.spyOn>;
+  const originalHome = process.env.FIRST_TREE_HOME;
+  const originalExitCode = process.exitCode;
+
+  beforeEach(() => {
+    home = mkdtempSync(join(tmpdir(), "fthub-prune-cmd-home-"));
+    outside = mkdtempSync(join(tmpdir(), "fthub-prune-cmd-outside-"));
+    process.env.FIRST_TREE_HOME = home;
+    printLines = [];
+    printSpy = vi.spyOn(print, "line").mockImplementation((text: string) => {
+      printLines.push(text);
+    });
+    sdkMocks.listMyAgents.mockResolvedValue([]);
+  });
+
+  afterEach(() => {
+    printSpy.mockRestore();
+    rmSync(home, { recursive: true, force: true });
+    rmSync(outside, { recursive: true, force: true });
+    if (originalHome === undefined) delete process.env.FIRST_TREE_HOME;
+    else process.env.FIRST_TREE_HOME = originalHome;
+    process.exitCode = originalExitCode;
+  });
+
+  async function runPrune(): Promise<void> {
+    const program = new Command();
+    const agent = program.command("agent");
+    registerAgentPruneCommand(agent);
+    await program.parseAsync(["node", "test", "agent", "prune", "--yes", "--server", "https://hub.example"]);
+  }
+
+  function writeStaleAlias(agentsDir: string, name: string): void {
+    mkdirSync(join(agentsDir, name), { recursive: true });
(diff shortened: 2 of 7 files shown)


yuezengwu (dismissed):
Blocking issue: `resolveContainedTarget` treats `realpath(baseDir)` as the trust root. If `$FIRST_TREE_HOME/config/agents`, `data/workspaces`, `data/sessions`, or one of their ancestors is symlinked outside the installation boundary, `realBase` and `realCandidate` move outside together, the containment check passes, and `rmSync` deletes `<outside>/<name>`. Anchor every canonical managed region to canonical `FIRST_TREE_HOME` and require it to resolve to its exact expected location. Please also preflight all three targets before the first mutation so a later unsafe target cannot leave the config alias removed and the remaining state undiscoverable. The legacy-compatible deletion name rule is reasonable, but it should be a shared, explicitly named persisted/cleanup schema rather than a private destructive-path regex. Static exact-head review only; I did not run tests.

liuchao-001:
Responding to the (since-dismissed) maintainer-side review for the record:

- **Preflight before the first mutation — adopted** in 71ef115: all targets are containment-validated before the first `rmSync`, and the config alias dir is deleted last, so a refused or failed later target can no longer strand workspace/session state behind an already-removed alias (`agent prune` rediscovers and retries by the alias dir). Locked by a new test where the unsafe target iterates last and earlier-validated real state must survive.
- **Trust-root anchoring** and the **shared named schema** for the deletion-side name rule are intentionally unchanged; the rationale for both is documented in the PR description under "Post-submission update (review feedback)".

PR #1904, opened 2026-07-21
chore(ci): parallelize test jobs

baixiaohang:
## Summary

- split the monolithic test job into parallel Server, Client/Web, and CLI jobs
- restore two Vitest forks for Server on its dedicated runner while preserving the CLI's single-fork, batched OOM protection
- isolate PostgreSQL and Turborepo cache namespaces per job
- retain the required `Test` check as a fail-closed fan-in, including failures from `Detect Changes`
- preserve the existing trigger scope: all three suites run for code changes and main pushes, while non-code pull requests still skip tests

## Why

The current test job takes about 10 minutes because Server, Client/Web, and CLI work share one runner and run largely serially. The global single-fork limit introduced for CLI OOM protection also unintentionally disables the Server suite's previously validated two-worker setup.

The new layout gives each resource profile its own runner. Local full-suite measurements put the parallel test critical path at about 2m42s before checkout/install/cache overhead, with peak RSS between 876 MB and 1.1 GB per group.

## Context Tree

- Draft context update: agent-team-foundation/first-tree-context#792

## Validation

- Server: 238 files / 2,557 tests passed in 2m42s; peak RSS ~1.1 GB
- Client/Web group: 5 workspace packages passed in 2m34s; peak RSS ~1.0 GB
- CLI: all 21 batches passed in 2m39s; peak RSS ~876 MB
- `pnpm check`
- `pnpm typecheck`
- workflow YAML parse
- actionlint (excluding the existing `github.head_ref` warning on main)
- `git diff --check`
- Turborepo dry-run scope verification

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,16 +108,14 @@ jobs:
       - run: pnpm build
       - run: pnpm typecheck
 
-  test:
-    name: Test
+  test_server:
+    name: Test Server
     needs: changes
     if: needs.changes.outputs.code == 'true' || github.event_name == 'push'
     runs-on: ubuntu-latest
-    # Sidecar Postgres started by the runner in parallel with checkout / install,
-    # so the test process never pays a `docker pull postgres:17` cost on its
-    # critical path. `global-setup.ts` reads `CI_DATABASE_URL` and skips the
-    # testcontainers spin-up; locally that env is unset and the original
-    # testcontainers path stays intact.
+    # Server tests own the only PostgreSQL sidecar. Keeping them on a dedicated
+    # runner restores the proven two-fork layout without stacking Fastify's
+    # module graph next to the CLI, Client, or Web suites.
     services:
       postgres:
         image: postgres:17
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed the CI split end to end. The three jobs form a complete, non-overlapping partition of the previous test graph:

- Server runs alone with its PostgreSQL sidecar and the restored two-fork cap.
- Client/Web covers the five remaining workspace test packages and retains the build artifacts needed by the skill-eval fixtures.
- CLI retains the explicit full build plus its single-fork, sequential batch protection.

The stable `Test` fan-in preserves the existing branch-protection contract: it runs for code PRs and pushes, stays skipped for non-code PRs, and fails on a failed, cancelled, or unexpectedly skipped child job or `Detect Changes` result. The Turborepo cache keys are also isolated per parallel job.

I independently checked the Turbo dry-run scopes, YAML parsing, diff hygiene, and the live workflow result at head `45a3a808`; all three suites and the final fan-in passed. No blocking findings.

PR #1907, opened 2026-07-21
feat: Add Factory GitHub workflows

liuchao-001:
Enable Factory Droid automated code review on this repository.

What was added:
- .github/workflows/droid.yml - responds to `@droid` mentions in issues, PR comments, and review bodies.
- .github/workflows/droid-review.yml - automatically reviews every new PR (non-draft), with the security-review pass enabled and `review_depth: deep` (default).

Both workflows reference `secrets.FACTORY_API_KEY` at runtime. The secret is already configured as an org secret on `agent-team-foundation` with `visibility: all`, and the Factory Droid GitHub App is installed organization-wide.

Prerequisites already done at the org level:
- Factory Droid GitHub App installed
- `FACTORY_API_KEY` available via org secret

Useful links:
- API key management: https://app.factory.ai/settings/api-keys
- Docs: https://docs.factory.ai

### DIFF
--- a/.github/workflows/droid-review.yml
+++ b/.github/workflows/droid-review.yml
@@ -1,0 +1,31 @@
+name: Droid Auto Review
+
+on:
+  pull_request:
+    types: [opened, ready_for_review, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  droid-review:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+      id-token: write
+      actions: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 1
+      - name: Run Droid Auto Review
+        uses: Factory-AI/droid-action@main
+        with:
+          factory_api_key: ${{ secrets.FACTORY_API_KEY }}
+          automatic_review: true
+          automatic_security_review: true
--- a/.github/workflows/droid.yml
+++ b/.github/workflows/droid.yml
@@ -1,0 +1,38 @@
+name: Droid Tag
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+  pull_request:
+    types: [opened, edited]
+
+jobs:
+  droid:
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@droid')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@droid')) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@droid')) ||
+      (github.event_name == 'issues' && (contains(github.event.issue.body, '@droid') || contains(github.event.issue.title, '@droid'))) ||
+      (github.event_name == 'pull_request' && (contains(github.event.pull_request.body, '@droid') || contains(github.event.pull_request.title, '@droid')))
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+      id-token: write
+      actions: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 1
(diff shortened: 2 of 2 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1907

- Rationale: The workflows establish a privileged third-party execution path without an immutable action revision, and the advertised automatic-review path does not work for this public repository's external contributors.

**Risk level**: B-low
- Path baseline: `.github/workflows/**` -> C
- Semantic lift: changes CI workflow configuration and secret-bearing automation -> B-low

**PR summary**
- Problem: Add on-demand Droid assistance and automatic deep/security review so maintainers receive earlier review feedback on pull requests.
- Approach: One workflow responds to trusted `@droid` requests; the other starts code and security review when a non-draft PR is opened, reopened, or marked ready. Both authenticate with the organization Factory API key and the installed Factory GitHub App.
- Impacted modules: `.github/workflows/droid.yml`, `.github/workflows/droid-review.yml`

**Review findings**

❌ 1. Pin `Factory-AI/droid-action` to a reviewed full commit SHA in both workflows. [`@main` is mutable](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions), while this action receives `FACTORY_API_KEY`, obtains an App token through OIDC, and runs with PR/issue write access (`droid-review.yml:26`, `droid.yml:33`). A moved or compromised branch would therefore execute new third-party code with repository credentials. GitHub documents a full commit SHA as the only immutable action reference; use an audited SHA (with a version comment if useful) and update it deliberately. [Security]

❌ 2. Define a safe external-contributor behavior instead of claiming this reviews every new PR. This is a public repository, but GitHub does not pass Actions secrets to `pull_request` workflows from forks (or Dependabot), so `FACTORY_API_KEY` is empty there. Independently, the current Droid action checks that the event actor has repository write permission before it evaluates the automatic-review trigger, which rejects ordinary external PR authors. The result is a failed/no-op Droid job rather than an automatic review (`droid-review.yml:4-13`). If the intended scope is trusted same-repository contributors only, guard that scope explicitly and document it; if fork coverage is required, use an upstream-supported design that does not expose secrets to untrusted code. Do not bypass this with `allowed_non_write_users: '*'` under the current broad permissions. [R5]

**Action taken**
- Submitted request changes.
- Operational note: the existing `Check Branch Name` job is also failing because `add-factory-workflows-1784607299` does not match the repository's accepted branch prefixes; this is not counted as a separate review finding.

yuezengwu:
Review completed against `e85829ee99` after this PR was closed (unmerged).

**Goal / core change:** add on-demand `@droid` handling plus automatic deep code/security review, backed by the organization Factory key and Factory GitHub App. This changes privileged repository automation only; there are no database or core-data-structure changes.

I independently confirm the existing blockers about pinning `Factory-AI/droid-action` to a reviewed full SHA and defining the intended fork/Dependabot behavior. Two additional blockers should be addressed before reopening or replacing this PR:

1. **A maintainer-triggered review of a fork PR crosses the trust boundary with credentials present.** For `issue_comment` / review events, the default-branch workflow has secrets. Droid checks the *triggering actor*'s write permission, so a maintainer's `@droid` comment passes; the action then checks out the fork PR head and runs Droid with shell-capable tools, unrestricted network by default, `FACTORY_API_KEY`, and an App token. The reviewed upstream revision enables `Execute` and checks out the PR head in [review.ts](https://github.com/Factory-AI/droid-action/blob/49188f4035f23e609bbeee36ec5b22e72212a8e5/src/tag/commands/review.ts), while [run-droid.ts](https://github.com/Factory-AI/droid-action/blob/49188f4035f23e609bbeee36ec5b22e72212a8e5/base-action/src/run-droid.ts) uses `--skip-permissions-unsafe`. A malicious diff can therefore target those credentials through prompt injection. Reject cross-repository PR heads in a credential-free preflight before invoking the action, or use a vendor-supported isolated/read-only review path with no repository secrets in the untrusted workspace. A naïve `pull_request_target` conversion or `allowed_non_write_users: '*'` would make this worse.

2. **Ordinary Issue mentions are wired to unsupported modes.** The current action requires a PR for `fill`, `review`, and normal `security`; a bare `@droid` defaults to review. Thus the `issues` trigger and non-PR `issue_comment` path can post a tracking comment and then fail (only `security --full` is issue-compatible), and `issues: assigned` is dead/noisy because assignee mode is disabled upstream. Remove `issues`, require `github.event.issue.pull_request` for comment events, or narrow Issue handling to an explicitly supported command.

Secondary hardening: automatic review should not retain `contents: write` unless a reviewed requirement proves it necessary; use least privilege and disable checkout credential persistence.

No tests or QA were run, per review scope.

PR #1917, opened 2026-07-21
Connect live agent output to the chat composer

Gandy2025:
## Summary

- move chat token usage above the live-status/composer stack and connect the current-output surface directly to the composer
- replace the separate Activity inspector with an inline compact/expanded current-turn view that preserves Markdown, provider reasons, multi-agent state, and keyboard focus
- fetch complete current-turn narration only while expanded, with backward-compatible text-block continuation metadata so long chunks and adjacent model blocks remain distinct
- avoid redundant tool details when narration exists and throttle full-output refreshes to text/turn events

## Scope

This changes only the Workspace chat composer status surface and its current-turn data path. It does not change the conversation list, message stream, sidebars, standalone mobile chat information architecture, or database schema.

## Verification

- `pnpm check`
- shared test suite: 647 passed
- focused server suites: 65 passed; post-review session-event suite: 25 passed
- focused web suites: 80 passed
- focused client suites: 122 passed across the provider/chunk-boundary reruns
- two independent UI/data reviews, including a successor-diff re-review after fixes
- production Vite build completed through the typecheck pipeline

## Known validation gaps

- The in-app Browser service exposed no browser instance in this environment, so real-browser screenshot QA could not be completed. DOM, accessibility, responsive CSS, production-build, and integration coverage were run instead.
- A full monorepo `pnpm test` attempt hit unrelated concurrent timeouts in `skill-evals`, client capability probes, and other long-running baseline suites. The affected shared/server/client/web suites for this change were rerun in isolation as listed above.


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1917

- Rationale: the on-demand current-turn cache can survive a server-side trace clear and later render the previous session's narration as the new turn's current output.

**Risk level**: A
- Semantic lift: also changes the server session-event service and the WebSocket-driven current-turn data path; no lift beyond A

**PR summary**
- Problem: people following a live agent turn need the agent's current narration attached to the composer, without opening a separate activity inspector or losing Markdown, multi-agent state, provider reasons, or keyboard focus.
- Approach: adds an on-demand, chat-scoped current-turn narration projection; adds continuation metadata to reconstruct chunked model blocks; and replaces the detached Activity inspector with a compact/expanded composer-connected surface.
- Impacted modules: Client runtime handlers, shared session-event contracts, Server chat/session status services, WebSocket cache invalidation, and Workspace composer UI.

**Review findings**

❌ 1. Invalidate the full-narration query when session traces are cleared. [`packages/web/src/hooks/use-admin-ws.ts:214`]

`ComposeStatusBar` keeps `chat-current-turn-narrations` fresh forever (`staleTime: Infinity`), but this handler invalidates it only for `assistant_text` and `turn_end`. Both explicit termination and runtime switching delete `session_events` and notify Web with `session:state = evicted`; the `session:state` branch does not touch this cache. A narration fetched before termination therefore remains fresh. If the chat later starts a new session whose first event is `thinking` or `tool_call`, expanding the composer reuses the old narration and presents it as current until another assistant-text or turn-end event arrives. Please invalidate/remove the affected chat narration on the lifecycle signal that accompanies trace deletion (and cover terminate/runtime-switch -> new tool-first turn in the WebSocket cache tests). [R5]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review summary

This connects live per-chat agent status to the composer, replaces the detached Activity inspector with an inline Markdown-capable current-turn view, adds an on-demand narration endpoint, and adds optional chunk-continuation metadata across providers.

There is no database migration. There is, however, a core shared event-contract change (`assistant_text.payload.continuation` plus the current-turn narration DTO), so the turn-boundary and compatibility semantics need explicit owner scrutiny.

I agree with the existing trace-clear/cache blocker. I found these additional issues on exact head `bed417617f27651e8c7a369a882e15737b4114d1`:

### 1. [P1 / blocking] Bound or page the complete-narration read

`packages/server/src/services/session-event.ts:322-385` selects every `assistant_text` row after the latest `turn_end` for every speaker, materializes all rows in JS, concatenates one uncapped string per agent, and returns them in one JSON response. A bound Client is explicitly zero-trust; the inbound WS accepts arbitrarily many individually valid <=8 KB events and does not require a `turn_end`. One authorized expand request can therefore materialize an arbitrarily large DB result, server string, response body, and Markdown tree. Repeated refreshes also resend the entire growing prefix.

Please preserve completeness through bounded cursor pagination/incremental assembly or enforce a server-side per-turn aggregate ceiling. The current single uncapped response is not a safe multi-tenant boundary.

### 2. [P2] A missing `turn_end` is not a reliable current-turn identity

`packages/server/src/services/session-event.ts:340-354` defines the current turn only as `seq > latest turn_end`. Session events intentionally survive local eviction/reconcile, and a crashed Client cannot append a closing marker. When a later runtime turn starts for the same `(agent, chat)`, this query combines the orphaned prior text with the new turn and presents both as current output.

Please fence narration with a durable turn-start/session-generation identity (or another authoritative current-turn start sequence), rather than treating every unclosed historical suffix as the new runtime turn.

### 3. [P2] Preserve Kimi model-step boundaries and every returned chunk

`packages/client/src/handlers/kimi-code.ts:318-332,487-494` concatenates all `assistant.delta` events into one `assistantText` even though the SDK emits `turn.step.started` for separate model steps. The new loop then marks chunks across that flattened value as continuations of one block. A text-before-tool / text-after-tool turn therefore loses the block boundary this PR is intended to preserve. Store text per model step and reset `continuation` for the first chunk of each block.

The same loop also drops whitespace-only interior chunks with `if (chunk.trim())`. `chunkAssistantText` already returns `[]` for wholly blank output; for a nonblank value containing an interior 8 KB whitespace chunk, this branch silently removes data. Emit every chunk returned by the helper.

### 4. [P2] Do not replace cached status/reason rows with narration loading errors

In `packages/web/src/components/chat/compose-status-bar.tsx:204-231`, the expanded `attention.map(...)` is entirely behind the narration request. While that request is slow, or if it fails, already-cached failure state, provider reason/detail, tool fallback, and all multi-agent rows disappear behind `Loading…` or a generic error. That contradicts the stated preservation of provider reasons and makes the status surface least useful during a data-path failure.

Render the local status rows unconditionally and let narration loading/error augment only their narration content.

### 5. [P2] Restore keyboard focus and scrolling guarantees for the inline region

`packages/web/src/components/chat/compose-status-bar.tsx:96-147` tracks focus only through the toggle. After focus moves toggle -> narration link, the toggle blur deliberately leaves `toggleHadFocusRef` true, but there is no section-level blur handler to clear it when the link later moves outside. When the final status disappears, the effect can then steal focus back to the composer after the user has left this surface. Conversely, if a focused agent link/Retry control is removed while another row remains, no recovery runs and focus falls to `body`.

Also, `.compose-status-detail` is an independently scrolling 42vh region but is not keyboard-focusable; plain-text narration may have no descendant a keyboard user can focus to scroll the tail. Track focus on the whole section, recover when a focused dynamic row disappears, and provide a focusable/focus-visible scroll target (or equivalent keyboard scrolling behavior).

Gandy2025:
Closing this PR because the implementation expanded a focused Web composer UI change into new cross-layer turn semantics and cache behavior. We will restart from `main` with a Web-only patch that uses the existing status contract and `turnTextFull`. The review findings remain valid and are not being treated as fixed on this branch.

PR #1918, opened 2026-07-21
Connect current agent output to the composer

Gandy2025:
## Summary

- move chat token usage above live agent status
- connect the live status surface directly to the composer instead of opening a detached Activity inspector
- expand the existing `turnTextFull ?? turnText` value inline with Markdown, without the former two-line clamp
- remove the duplicated identity header from single-agent expanded output; use an aggregate header for multi-agent output while retaining per-agent ownership in the details
- suppress redundant tool/Thinking metadata when narration already explains the current work, while retaining it as a no-text fallback
- preserve multi-agent states, provider reasons, stable timeline evidence links, keyboard scrolling, responsive state visibility, and focus recovery

## Root cause

The composer rendered token usage between the live status and the input. The expanded status view also read the 120-character `turnText` preview and applied a two-line clamp even though the existing status contract already provides `turnTextFull`.

This patch corrects the Web layout and field usage. It intentionally does not add an API, cache, event contract, session identity, or provider-handler change. The existing server boundary for `turnTextFull` (latest text block, capped at 2,000 characters) remains unchanged.

## Validation

- full Web suite before the final header refinement: 192 files, 1,561 tests
- final focused `ComposeStatusBar` coverage: 20 tests
- final `ChatView` integration coverage: 44 tests
- two independent read-only reviews covering UI/accessibility and state/focus behavior; all P1/P2 findings fixed and re-reviewed
- real local stack available at `http://localhost:5174/?c=ac1a110d-fe19-4358-928f-8464ab983e5b`

Latest browser-rendered screenshot comparison is pending: the approved browser runtime has no available browser instance in this session. The PR remains draft until that visual QA is complete.

### DIFF
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -5,24 +5,22 @@ import {
   type LiveActivity,
 } from "@first-tree/shared";
 import { useQuery } from "@tanstack/react-query";
-import { Brain, ChevronDown, Pencil, Wrench, X } from "lucide-react";
-import { type Ref, type RefObject, useCallback, useEffect, useId, useMemo, useRef, useState } from "react";
+import { Brain, ChevronDown, Pencil, Wrench } from "lucide-react";
+import { type RefObject, useEffect, useId, useMemo, useRef, useState } from "react";
 import { chatAgentStatusQueryKey, fetchChatAgentStatuses } from "../../api/agent-status.js";
 import { viewOf } from "../../lib/agent-status-view.js";
 import { stripInlineMarkdown } from "../../lib/strip-inline-markdown.js";
 import { isJumpable, type TimelineAnchorKind, useMountedAnchors } from "../../lib/use-mounted-anchors.js";
+import { Markdown } from "../ui/markdown.js";
 import { StatusGlyph } from "../ui/status-glyph.js";
 import { TimelineJumpButton } from "./timeline-jump-button.js";
 
 /**
  * ComposeStatusBar is the stable answer to “what are the agents doing now?”.
  *
- * The collapsed strip keeps one invariant structure: lead snapshot on the left,
- * `Activity (N)` on the right. The explicit Activity control always opens the
- * same live inspector, regardless of viewport or agent count. The inspector is
- * intentionally a current-state index, not another timeline: every agent gets
- * one compact item with at most two lines of latest narration plus its current
- * tool/reason. Full narration and activity history remain in the timeline.
+ * The compact row is the top edge of the composer: it shows one live snapshot
+ * and expands in place to the current output for every actionable agent. This
+ * is a status projection, not another timeline or a floating inspector.
  *
  * Only working, failed, waiting, retrying, and terminal-reason states raise the
  * strip. When every participant is quiet, the strip disappears.
@@ -99,24 +97,16 @@ export function ComposeStatusBar({
   chatId: string;
   /** Non-human agent participants, used for stable display-name lookup. */
   agents: ChatParticipantDetail[];
-  /** Stable composer control that receives focus if the activity bar vanishes. */
+  /** Stable composer control that receives focus if the status surface vanishes. */
   fallbackFocusRef?: RefObject<HTMLElement | null>;
 }) {
-  const [inspectorOpen, setInspectorOpen] = useState(false);
+  const [expanded, setExpanded] = useState(false);
   const [lead, setLead] = useState<{ agentId: string; since: number } | null>(null);
-  const inspectorId = useId();
+  const detailsId = useId();
   const activeChatIdRef = useRef(chatId);
-  const inspectorRef = useRef<HTMLElement | null>(null);
+  const surfaceRef = useRef<HTMLElement | null>(null);
   const triggerRef = useRef<HTMLButtonElement | null>(null);
   const focusWithinRef = useRef(false);
-  const closeInspector = useCallback(() => {
-    focusWithinRef.current = false;
-    setInspectorOpen(false);
-  }, []);
-  const closeInspectorAndRestoreFocus = useCallback(() => {
-    setInspectorOpen(false);
-    triggerRef.current?.focus();
-  }, []);
   const { data: rawStatuses } = useQuery({
     queryKey: chatAgentStatusQueryKey(chatId),
     queryFn: () => fetchChatAgentStatuses(chatId),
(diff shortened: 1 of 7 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1918

- Rationale: The connected composer surface is directionally sound, but live status refreshes can now move keyboard focus away from a Markdown link that is still mounted.

**Risk level**: C
- Semantic lift: none; this changes the Web presentation and focus behavior without changing subscriptions, WebSocket parsing, or the data contract.

**PR summary**
- Problem: Users monitoring live agent work currently have to open a detached Activity inspector, and the expanded view truncates the available narration even though fuller output already exists.
- Approach: Connect live status to the top of the composer, place token usage above it, render `turnTextFull ?? turnText` as inline Markdown, and keep multi-agent ownership and timeline evidence links in one expandable surface.
- Impacted modules: Web chat composer status, Workspace footer layout, responsive styling, and DOM coverage.

**Review findings**

❌ 1. Preserve focus for still-mounted Markdown controls during status refreshes. The recovery effect derives `activeAgentId` for every focused descendant of an output item, then requires that item's timeline anchor to remain jumpable. That assumption held when the old item itself was the timeline-jump button, but the new `Markdown` renderer can add independent focusable links. If an agent has no mounted timeline anchor (a supported bounded-timeline case), focusing one of those links and receiving the next WS/status update makes `focusedRowSurvives` false and schedules focus back to the disclosure even though the link is still present. Please scope the anchor-survival check to the `.compose-status-jump` control (or otherwise check whether the actually focused node survived), and add coverage for a focused Markdown link with no mounted timeline anchor across a status update. [Accessibility / `packages/web/src/components/chat/compose-status-bar.tsx:163`]

⚠️ 2. The reduced-motion override still targets the removed `[data-live-activity-inspector]`, while the new expanded details retain `className="... fade-in ..."`. As a result, expanding current output animates under `prefers-reduced-motion: reduce`. Retarget the override to `[data-current-agent-output]` or remove the decorative fade from that region. [Accessibility / `packages/web/src/index.css:931`]

**Action taken**
- Submitted request changes.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -432,51 +362,58 @@ function InspectorItem({
   const accessibleSummary = [
     name,
     stateLabel(status),
-    snapshot.update,
+    stripInlineMarkdown(snapshot.update),
```
yuezengwu: **[P2] Keep the timeline jump's accessible name concise.** `snapshot.update` now comes from `turnTextFull ?? turnText`, so this label can contain the full 2,000-character narration (including multiline Markdown that this inline-only stripper does not normalize). A screen-reader user tabbing to the arrow must hear that entire output before "View in the timeline", then encounters the same Markdown again immediately below. Please name the action from the agent/state (or at most the bounded `turnText` preview) and add coverage with a long `turnTextFull`.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -432,51 +362,58 @@ function InspectorItem({
...
-    </TimelineJumpButton>
+      ) : null}
+      {anchored ? (
+        <TimelineJumpButton
+          key="timeline-jump"
+          agentId={status.agentId}
+          target={jumpTarget}
+          anchored
+          ariaLabel={accessibleSummary}
+          className="compose-status-jump"
```
yuezengwu, line 393: **[P2] Restore a usable pointer target for the timeline jump.** This button's only visible child is the `h-3.5 w-3.5` arrow; `TimelineJumpButton` contributes zero padding and `.compose-status-jump` only positions it. That leaves an approximately 14×14 px hit area, a significant regression from the former full-row action and especially difficult on touch devices. Please give it at least a 24×24 hit area (preferably the existing phone touch-target size) without enlarging the glyph.

yuezengwu (changes requested):
## Review

**Goal and core change:** connect live agent status directly to the composer, move token usage above it, and render the existing `turnTextFull ?? turnText` inline as Markdown while preserving multi-agent ownership, provider reasons, timeline evidence, and focus recovery. The patch stays within the Web presentation layer; it does not change the status, API, cache, or database contracts.

**Recommendation:** request changes.

I independently confirmed the two accessibility regressions already raised in baixiaohang's review: live refreshes can steal focus from a surviving Markdown link, and the reduced-motion override still targets the removed inspector. I am adding two further P2 findings inline: the timeline jump exposes up to 2,000 characters as its accessible name, and its new pointer target is only about 14×14 px.

No tests or QA were run for this diff-only review.

Gandy2025:
Addressed all four accessibility findings in 5d9089b43:

- Focus recovery now checks timeline-anchor survival only for the timeline-jump control. A still-mounted Markdown link keeps focus across live status updates, with a regression test for the no-anchor case.
- The reduced-motion override now targets the current agent output region.
- Timeline-jump accessible names are concise (`View {agent} in the timeline`) and no longer repeat the full narration; long `turnTextFull` coverage was added.
- The jump target is 28×28px on desktop and 44×44px on mobile without enlarging the glyph. Anchored narration now reserves that footprint for both single- and multi-agent rows, with DOM and mobile CSS regressions.

Validation: Web typecheck and design-token guardrails pass; focused component/ChatView/mobile tests pass 67/67; the full Web suite reached 191/192 files before exposing one stale aria-label assertion, and that corrected ChatView file then passed 44/44. Two independent reviews of the final diff are clean. Browser QA remains unavailable because this session has no browser instance.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 5d9089b43. The jump action is now named only from the agent: View {agent} in the timeline. It no longer includes turnTextFull or metadata, and the component test covers a long full narration.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed in 5d9089b43. The transparent hit area is now 28x28px on desktop and 44x44px on mobile while the arrow stays unchanged. Both single- and multi-agent narration reserve the footprint; DOM and mobile CSS regressions cover it.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1918

**Recommendation**: approve
- Rationale: The successor head fixes the four accessibility regressions without changing the status authority or introducing a new blocking issue.

**Risk level**: C
- Semantic lift: none; the update remains in Web presentation, focus handling, responsive CSS, and deterministic DOM coverage.

**PR summary**
- Problem: Users monitoring live agent work had to open a detached inspector, while the expanded view discarded the fuller narration already available from the status projection.
- Approach: Connect current agent output to the composer, move token usage above it, render the bounded full narration as Markdown, and preserve multi-agent ownership plus timeline evidence links.
- Impacted modules: Web composer status, Workspace footer layout, responsive styling, and focused DOM tests.

**Review findings**

✅ 1. Focus recovery now applies timeline-anchor survival only to the actual jump control; a surviving Markdown link remains focused across live status updates, including the no-anchor case.

✅ 2. The reduced-motion override targets the new current-output region.

✅ 3. Timeline-jump names are concise and no longer include the up-to-2,000-character narration.

✅ 4. The jump target is 28×28px on desktop and 44×44px on mobile, with matching narration/identity clearance in single- and multi-agent rows.

No blocking findings remain on `5d9089b43f17a0558875428007441e32a009453f`. This was a static review; I did not run tests or browser QA.

**Action taken**
- Approved.

yuezengwu (approved):
Re-reviewed successor head `5d9089b43`. All four accessibility findings are addressed: focus recovery now distinguishes surviving Markdown controls from timeline jumps; reduced-motion targets the current output region; timeline-jump names are concise; and the hit area is 28×28px on desktop / 44×44px on mobile with matching narration clearance and regression coverage.

The net change remains Web presentation/focus only and preserves the per-chat status authority and existing API/data boundaries. No further actionable findings. Approved without running local tests or QA, per the diff-review scope; pending browser visual QA remains separate.

PR #1920, opened 2026-07-21
fix: capture angle-bracketed workspace images

Gandy2025:
## What changed

- recognize both bare and angle-bracketed local Markdown image destinations during agent `chat send` capture
- unwrap angle-bracket syntax before MIME detection and workspace path resolution
- cover absolute angle-bracketed paths and angle-bracketed paths containing spaces

## Root cause

The bounded image matcher treated `<...>` as part of the destination. A path ending in `.png>` therefore failed the supported-image extension check, so capture was skipped and the original agent-local path was sent as ordinary Markdown. The Web could not read that local filesystem path and rendered a broken image.

## Impact

Agent messages such as `preview` now follow the existing attachment upload path and render as normal `format: "file"` image messages. The existing bare-path behavior and workspace provenance fence are unchanged.

## Validation

- `pnpm --filter @first-tree/client exec vitest run src/__tests__/image-snapshots.test.ts` — 27 passed
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/image-capture.test.ts` — 10 passed
- `pnpm check` — passed (existing repository warnings only)
- `pnpm typecheck` — passed
- Two independent code reviews completed with no blocking, P1, or P2 findings on the final diff

The full parallel `pnpm test` run hit timeout-only failures in unrelated long-running CLI, capability-probe, and skill-eval tests. Every failing file passed when rerun serially; the affected code is outside this two-file change.

### DIFF
--- a/packages/client/src/__tests__/image-snapshots.test.ts
+++ b/packages/client/src/__tests__/image-snapshots.test.ts
@@ -39,6 +39,7 @@ describe("buildMessageImageSnapshots — capture + strip", () => {
     root = await realpath(await mkdtemp(join(tmpdir(), "img-snap-")));
     await mkdir(join(root, "shots"), { recursive: true });
     await writeFile(join(root, "shots", "filter.png"), PNG);
+    await writeFile(join(root, "shots", "filter with spaces.png"), PNG);
     await writeFile(join(root, "diagram.webp"), PNG);
     await writeFile(join(root, "notes.txt"), Buffer.from("hi"));
   });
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the final diff at `25a850f5`.

This narrowly extends outbound image capture to recognize CommonMark angle-bracketed local destinations, unwraps the brackets before the existing MIME, URL, and workspace-fence checks, and adds coverage for absolute and space-containing paths.

No blocking issues found. The existing provenance fence, best-effort failure behavior, attachment limits/deduplication, bare-path behavior, and `format: "file"` output remain unchanged. There are no database/schema or core data-structure changes.

Per review scope, I did not rerun tests; the product CI checks on this head are green.

PR #1921, opened 2026-07-21
fix: claim webhook events with status + TTL to stop processed_events leak (#317)

### LINKED ISSUE #317: GitHub App webhook: processed_events leak when claim succeeds but post-claim work fails

yuezengwu:
## Summary

Fixes #317 — GitHub App webhook `processed_events` leak: a claim row committed before the handler ran, and if the process crashed before the best-effort `unclaimEvent` delete, GitHub's redelivery deduped against the orphaned row and the event was lost forever.

This PR implements the GoF-decided fix shape 2 — **claim with status + TTL**:

- `processed_events` gains a claim lifecycle: `status` (`pending` → `done`) plus `expires_at` on live claims (claim time + 5 min TTL).
- `claimEvent` is now a single atomic `INSERT ... ON CONFLICT DO UPDATE ... WHERE`: fresh ids insert a `pending` claim; an **expired** `pending` claim is taken over in the same statement; a live `pending` or `done` claim dedupes. Two concurrent deliveries of the same id can never both process (READ COMMITTED re-evaluates the loser against the winner's refreshed `expires_at`).
- The SCM processing seam (`processScmWebhookDelivery`) marks the claim `done` on every successful outcome (`delivered` / `audience_empty` / `provider_only`). Correctness no longer depends on the best-effort `unclaimEvent` delete — it remains purely as a fast-redelivery optimization for handled failures.
- A background sweep (`sweepExpiredClaims`, every 60s in `background-tasks.ts`) deletes expired `pending` rows so redelivery is never starved and the table does not accumulate stuck claims.

### Scope note (both webhook paths)

The issue text mentions the per-org route `/webhooks/github/:orgId`. That route no longer exists on main — it was consolidated into the single SaaS-wide `/webhooks/github-app` endpoint. The claim pattern now lives in one shared seam consumed by exactly two webhook paths: **GitHub App** (`api/webhooks/github-app.ts`) and **GitLab** (`api/webhooks/gitlab.ts`). Both move to the new lifecycle through the seam; no route code changes.

### Migration note (custom SQL)

`0082_classy_chronomancer.sql` is a `drizzle-kit generate --custom` migration. `processed_events` predates the drizzle schema index and is not tracked in the snapshot chain (created in `0003_feishu_adapter.sql`, never exported from `src/db/schema/index.ts`), so `generate` cannot diff-emit the ALTERs. The custom file does add-column → backfill `status='done'` (existing rows already completed processing and must keep deduping) → `SET NOT NULL` → `SET DEFAULT 'pending'` (fail-safe runtime default) explicitly. `node scripts/check-migrations.mjs` passes.

## Test plan

Product tests (Vitest, real PostgreSQL via testcontainers) — `packages/server/src/__tests__/`:

- `event-dedup.test.ts`
  - claim lands as `pending` with `expires_at ≈ now + TTL`
  - live `pending` dedupes; cross-platform ids are independent; `unclaimEvent` still allows immediate re-claim
  - `markEventDone` flips to `done` and clears expiry; `done` dedupes forever
  - expired-`pending` takeover re-claims atomically and refreshes the row to live `pending`
  - `sweepExpiredClaims` deletes only expired `pending` rows (live `pending` and `done` untouched)
- `scm-webhook-processing.test.ts`
  - success marks the claim `done`; redelivery dedupes against it (no duplicate side effects)
  - **crash injection**: claim lands, handler never completes (no done/unclaim), TTL lapses → redelivery with the same delivery id **processes** (lost-forever scenario is gone)
  - redelivery within the TTL still dedupes (no double processing while the original attempt may be live)
  - **race**: two concurrent deliveries of the same id → exactly one `delivered`, one `duplicate`, side-effect callback invoked once
  - **race on takeover**: two concurrent deliveries over one expired claim → exactly once
  - pre-existing suites (`github-app-webhook-route`, `gitlab-webhook-stage2a/3`, `schema-extra`) run green against the new lifecycle

GoF acceptance checklist mapping:

- [x] Claim rows carry `pending`/`done` status plus expiry; successful handling marks `done`
- [x] Crash-injection test: claim lands, handler dies, redeliver after expiry → processed
- [x] `done` claims still dedupe: redelivery after success produces no duplicate side effects
- [x] Race test: concurrent deliveries of the same id process exactly once
- [x] Both webhook paths (GitHub App + GitLab via the shared seam) use the new lifecycle
- [x] Expired-`pending` sweep behavior covered by a test

Validation run: `pnpm check`, `pnpm --filter @first-tree/server typecheck`, targeted `vitest run` for the suites above, `node scripts/check-migrations.mjs`.

## QA risk

Touches the SCM webhook ingress path (cross-surface). Matching QA case: `packages/qa/cases/cross-surface/github-webhook-routing-regression.md`. Deterministic coverage lives in the product tests above; formal QA per `packages/qa/AGENTS.md` is warranted at maintainer discretion.

### DIFF
--- a/packages/server/drizzle/0082_classy_chronomancer.sql
+++ b/packages/server/drizzle/0082_classy_chronomancer.sql
@@ -1,0 +1,19 @@
+-- Custom SQL migration (drizzle-kit generate --custom).
+--
+-- Why custom: `processed_events` predates the drizzle schema index and is
+-- not tracked in the drizzle snapshot chain (it was created in
+-- 0003_feishu_adapter.sql and is intentionally not exported from
+-- src/db/schema/index.ts), so `drizzle-kit generate` cannot diff-emit the
+-- ALTERs. This file applies the claim-lifecycle columns for issue #317:
+--
+--   - existing rows represent events that already completed processing
+--     (they received a 200), so they backfill to 'done' and keep deduping;
+--   - new claims always set status explicitly; 'pending' is the fail-safe
+--     default direction (an unforeseen insert stays re-processable, never
+--     silently deduped);
+--   - `expires_at` is set only on live 'pending' claims (claim time + TTL).
+ALTER TABLE "processed_events" ADD COLUMN "status" text;--> statement-breakpoint
+ALTER TABLE "processed_events" ADD COLUMN "expires_at" timestamp with time zone;--> statement-breakpoint
+UPDATE "processed_events" SET "status" = 'done';--> statement-breakpoint
+ALTER TABLE "processed_events" ALTER COLUMN "status" SET NOT NULL;--> statement-breakpoint
+ALTER TABLE "processed_events" ALTER COLUMN "status" SET DEFAULT 'pending';
--- a/packages/server/drizzle/LATEST
+++ b/packages/server/drizzle/LATEST
@@ -1,1 +1,1 @@
-0081_dizzy_tyrannus
+0082_classy_chronomancer
--- a/packages/server/drizzle/meta/0082_snapshot.json
+++ b/packages/server/drizzle/meta/0082_snapshot.json
@@ -1,0 +1,5582 @@
+{
+  "id": "d38ed3ce-f823-4aea-886d-b3ee604e76c0",
+  "prevId": "ed77360a-1de2-4ce5-9f63-dd2d28edbdaf",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent_chat_sessions": {
+      "name": "agent_chat_sessions",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "chat_id": {
+          "name": "chat_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "state": {
+          "name": "state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "runtime_state": {
+          "name": "runtime_state",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'idle'"
+        },
+        "runtime_state_at": {
+          "name": "runtime_state_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "idx_agent_chat_sessions_chat_agent": {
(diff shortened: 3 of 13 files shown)


yuezengwu:
GoF submission ready for maintainer review (label `fire_wip` → `fire_submitted`; PR stays draft per competition rules — no ready-for-review, no merge from our side).

Status summary:
- Implements the GoF-decided fix shape 2 for #317 (claim status + TTL); both current webhook paths (GitHub App + GitLab) covered via the shared `processScmWebhookDelivery` seam. The per-org `/webhooks/github/:orgId` route referenced in the issue no longer exists on main — details in the PR body.
- GoF acceptance checklist: all 6 items covered by product tests (crash-injection, done-dedup, concurrent exactly-once, takeover race, sweep), see PR body for the mapping.
- CI all green (Lint & Type Check, Test Server, Test CLI, Test Client & Web, Migrations, CodeQL).
- Migration note: `0082` is a `drizzle-kit generate --custom` file because `processed_events` is not tracked in the drizzle snapshot chain (pre-existing drift since 0003); existing rows backfill to `done`, runtime default is `pending`. `scripts/check-migrations.mjs` passes.

PR #1923, opened 2026-07-21
Render sent chat mentions with display names

Gandy2025:
## Summary

- render sent-message mentions with each chat participant's `displayName` while keeping immutable `name` as the resolver key
- visibly disambiguate duplicate display names and fall back to the canonical handle when a display name is blank or no longer resolvable
- preserve canonical handles in both plain-text and rich-text copy without changing composer parsing, stored content, mention metadata, or server routing
- constrain long mention chips on narrow screens and refresh rendered labels after a display-name update

## Validation

- `pnpm --filter @first-tree/web test` — 192 files, 1,574 tests passed
- `pnpm check`
- focused mention + ChatView DOM tests — 60 tests passed
- Playwright visual QA at 1280×800 and 360×740; no horizontal overflow, long display names ellipsize
- two independent code reviews completed; all correctness and copy regressions found during review were addressed

## Test note

An initial concurrent monorepo `pnpm test` attempt exhausted timing budgets in unrelated `skill-evals` fixtures. The affected Web package was then run independently and passed in full as reported above.

### DIFF
--- a/packages/web/src/components/rehype-mentions.ts
+++ b/packages/web/src/components/rehype-mentions.ts
@@ -20,9 +20,10 @@ export type RenderedMentionParticipant = MentionParticipant & {
  * formatting and a chip inside a link would double-style.
  *
  * The participant list is captured by closure and resolved
- * case-insensitively, matching the server's `extractMentions` resolver
- * so the chip appears for exactly the same tokens the router would
- * fan out to.
+ * case-insensitively, matching the server's `extractMentions` resolver.
+ * Historical callers can additionally restrict resolution to persisted
+ * recipient IDs without shrinking the full-roster collision universe used to
+ * disambiguate duplicate display names.
  *
  * Returns a function compatible with react-markdown's `rehypePlugins`
  * prop. Local minimal hast types avoid pulling in `@types/hast` and
@@ -150,12 +151,15 @@ function walk(node: HastNode, nameMap: Map<string, RenderedMentionIdentity>, sel
 
 export function rehypeMentions(
   participants: RenderedMentionParticipant[],
-  options?: { selfAgentId?: string | null },
+  options?: { selfAgentId?: string | null; allowedAgentIds?: ReadonlySet<string> },
 ): () => (tree: HastRoot) => void {
   const selfAgentId = options?.selfAgentId ?? null;
+  const allowedAgentIds = options?.allowedAgentIds;
   const nameMap = new Map<string, RenderedMentionIdentity>();
   const displayNameCounts = new Map<string, number>();
 
+  // Count across the full visible speaker roster. A message that mentions only
+  // one of two people called "Sam" still needs to expose that person's handle.
   for (const participant of participants) {
     if (!participant.name) continue;
     const displayName = participant.displayName.trim() || participant.name;
@@ -164,7 +168,10 @@ export function rehypeMentions(
   }
 
   for (const p of participants) {
-    if (p.name) {
+    // Resolution is a separate identity decision. Historical callers allow
+    // only the IDs persisted for this message, so handle reuse cannot relabel
+    // an old token even though the new owner remains in the collision roster.
+    if (p.name && (!allowedAgentIds || allowedAgentIds.has(p.agentId))) {
       const displayName = p.displayName.trim() || p.name;
       nameMap.set(p.name.toLowerCase(), {
         name: p.name,
--- a/packages/web/src/components/rehype-mentions.ts
+++ b/packages/web/src/components/rehype-mentions.ts
@@ -1,6 +1,18 @@
 import { MENTION_REGEX, type MentionParticipant } from "@first-tree/shared";
 
 /**
+ * Chat-scoped identity shape used only by the sent-message renderer.
+ *
+ * Routing and composer parsing deliberately keep using the narrower
+ * {@link MentionParticipant} (`agentId` + immutable `name`). `displayName` is
+ * presentation-only: it may contain spaces, non-ASCII characters, collide
+ * with another participant, or change after the message was sent.
+ */
+export type RenderedMentionParticipant = MentionParticipant & {
+  displayName: string;
+};
+
+/**
  * Rehype plugin: walks the hast tree, finds `@<participant>` tokens
  * inside `text` nodes, and rewrites them to `span.mention-chip` elements.
  * Code regions (`<code>`, `<pre>`) and existing `<a>` link text are
@@ -32,7 +44,19 @@ type HastNode = HastText | HastElement | HastRoot | { type: string; children?: H
 
 const SKIP_TAGS = new Set(["code", "pre", "a"]);
 
-function makeMentionElement(value: string, name: string, agentId: string, isSelf: boolean): HastElement {
+type RenderedMentionIdentity = {
+  name: string;
+  displayName: string;
+  agentId: string;
+  disambiguate: boolean;
+};
+
+function makeMentionElement(identity: RenderedMentionIdentity, isSelf: boolean): HastElement {
+  const displayName = identity.displayName.trim() || identity.name;
+  const displayLabel = `@${displayName}`;
+  const identityLabel = displayName === identity.name ? displayLabel : `${displayLabel} (@${identity.name})`;
+  const visibleLabel = identity.disambiguate ? identityLabel : displayLabel;
+
   return {
     type: "element",
     tagName: "span",
@@ -43,16 +67,19 @@ function makeMentionElement(value: string, name: string, agentId: string, isSelf
       // it. `is-x` matches the repo's existing state-modifier convention
       // (`.context-network-card.is-live`, `.context-usage-feed-row.is-fresh`).
       className: isSelf ? ["mention-chip", "is-self"] : ["mention-chip"],
-      "data-mention-agent-id": agentId,
-      "data-mention-name": name,
+      "data-mention-agent-id": identity.agentId,
+      "data-mention-name": identity.name,
+      "data-mention-display-name": displayName,
+      title: identityLabel,
+      ariaLabel: identityLabel,
     },
-    children: [{ type: "text", value }],
+    children: [{ type: "text", value: visibleLabel }],
   };
 }
 
 function splitTextNode(
   node: HastText,
-  nameMap: Map<string, { name: string; agentId: string }>,
+  nameMap: Map<string, RenderedMentionIdentity>,
   selfAgentId: string | null,
 ): HastNode[] | null {
   const content = node.value;
@@ -67,7 +94,7 @@ function splitTextNode(
     if (m.index > cursor) {
       out.push({ type: "text", value: content.slice(cursor, m.index) });
     }
-    out.push(makeMentionElement(m[0], resolved.name, resolved.agentId, resolved.agentId === selfAgentId));
+    out.push(makeMentionElement(resolved, resolved.agentId === selfAgentId));
     cursor = m.index + m[0].length;
   }
   if (out.length === 0) return null; // no rewrites — leave node intact
@@ -77,11 +104,7 @@ function splitTextNode(
   return out;
 }
 
-function walk(
-  node: HastNode,
-  nameMap: Map<string, { name: string; agentId: string }>,
-  selfAgentId: string | null,
-): void {
+function walk(node: HastNode, nameMap: Map<string, RenderedMentionIdentity>, selfAgentId: string | null): void {
   if (!("children" in node) || !node.children) return;
   const next: HastNode[] = [];
   for (const child of node.children) {
@@ -110,16 +133,143 @@ function walk(
 }
 
 export function rehypeMentions(
-  participants: MentionParticipant[],
+  participants: RenderedMentionParticipant[],
   options?: { selfAgentId?: string | null },
 ): () => (tree: HastRoot) => void {
   const selfAgentId = options?.selfAgentId ?? null;
-  const nameMap = new Map<string, { name: string; agentId: string }>();
+  const nameMap = new Map<string, RenderedMentionIdentity>();
+  const displayNameCounts = new Map<string, number>();
+
+  for (const participant of participants) {
+    if (!participant.name) continue;
+    const displayName = participant.displayName.trim() || participant.name;
+    const key = displayName.toLowerCase();
+    displayNameCounts.set(key, (displayNameCounts.get(key) ?? 0) + 1);
+  }
+
   for (const p of participants) {
-    if (p.name) nameMap.set(p.name.toLowerCase(), { name: p.name, agentId: p.agentId });
+    if (p.name) {
+      const displayName = p.displayName.trim() || p.name;
+      nameMap.set(p.name.toLowerCase(), {
+        name: p.name,
+        displayName,
+        agentId: p.agentId,
+        // displayName is mutable and non-unique. Keep the common case terse,
+        // but make collisions distinguishable on touch devices where a title
+        // tooltip is not discoverable. If the label is already this person's
+        // canonical handle, adding the same handle again would be redundant.
+        disambiguate: (displayNameCounts.get(displayName.toLowerCase()) ?? 0) > 1 && displayName !== p.name,
+      });
+    }
   }
   return () => (tree: HastRoot) => {
     if (nameMap.size === 0) return;
     walk(tree, nameMap, selfAgentId);
   };
+}
+
+/**
+ * Reconstruct the selected plain text with canonical `@name` handles.
+ *
+ * Sent messages render the friendlier `@displayName`, but copied text needs to
+ * remain pasteable into the composer and CLI. Returning `null` means the
+ * selection contains no complete rendered mention and native copy behaviour
+ * should proceed unchanged.
+ */
+type SelectedMention = {
+  start: number;
+  end: number;
+  name: string;
+  visibleText: string;
+};
+
+function selectedMentions(root: HTMLElement, selection: Selection | null): SelectedMention[] | null {
+  if (!selection || selection.rangeCount === 0 || selection.isCollapsed) return null;
+  const range = selection.getRangeAt(0);
+  if (!root.contains(range.commonAncestorContainer)) return null;
+
+  const mentions: SelectedMention[] = [];
+  const chips = root.querySelectorAll<HTMLElement>(".mention-chip[data-mention-name]");
+
+  for (const chip of chips) {
+    if (!range.intersectsNode(chip)) continue;
+    const visibleText = chip.textContent;
+    const name = chip.dataset.mentionName;
+    if (!visibleText || !name) continue;
+
+    // Only rewrite a fully selected chip. A selection that begins or ends
+    // inside the visible label should retain the browser's native partial text.
+    // Compare the selection against the chip's textual boundaries, not the
+    // surrounding element boundaries. A user can drag from text offset 0 to
+    // the final offset and visibly select the whole label without selecting
+    // the span node itself.
+    const textWalker = root.ownerDocument.createTreeWalker(chip, 4); // NodeFilter.SHOW_TEXT
+    const firstText = textWalker.nextNode();
+    let lastText = firstText;
+    for (let node = textWalker.nextNode(); node; node = textWalker.nextNode()) lastText = node;
+    if (firstText?.nodeType !== 3 || lastText?.nodeType !== 3) continue;
+
+    const chipRange = root.ownerDocument.createRange();
+    chipRange.setStart(firstText, 0);
+    chipRange.setEnd(lastText, lastText.textContent?.length ?? 0);
+    if (
+      range.compareBoundaryPoints(range.START_TO_START, chipRange) > 0 ||
+      range.compareBoundaryPoints(range.END_TO_END, chipRange) < 0
+    ) {
+      continue;
+    }
+
+    // Measure the chip's exact position within the selected text instead of
+    // replacing by label value. Two people may share a displayName, and the
+    // same visible string may also appear as ordinary text before the chip.
+    const prefixRange = range.cloneRange();
+    prefixRange.setEnd(chipRange.startContainer, chipRange.startOffset);
+    const start = prefixRange.toString().length;
+    mentions.push({ start, end: start + visibleText.length, name, visibleText });
+  }
+
+  return mentions.length > 0 ? mentions : null;
+}
+
+export function copyTextWithMentionHandles(root: HTMLElement, selection: Selection | null): string | null {
+  const mentions = selectedMentions(root, selection);
+  if (!mentions || !selection) return null;
+  mentions.sort((left, right) => right.start - left.start);
+
+  return mentions.reduce(
+    (text, mention) => `${text.slice(0, mention.start)}@${mention.name}${text.slice(mention.end)}`,
+    selection.toString(),
+  );
+}
+
+/**
+ * Preserve native rich-text copy while rewriting complete mention chips to
+ * canonical handles. Consumers should set this alongside the plain-text
+ * payload so pasting into Docs/Notion retains links and emphasis.
+ */
+export function copyHtmlWithMentionHandles(root: HTMLElement, selection: Selection | null): string | null {
+  const mentions = selectedMentions(root, selection);
+  if (!mentions || !selection) return null;
+
+  const fragment = selection.getRangeAt(0).cloneContents();
+  const pending = [...mentions].sort((left, right) => left.start - right.start);
+
+  for (const chip of fragment.querySelectorAll<HTMLElement>(".mention-chip[data-mention-name]")) {
+    const mention = pending[0];
+    if (!mention || chip.dataset.mentionName !== mention.name || chip.textContent !== mention.visibleText) continue;
+    chip.replaceWith(root.ownerDocument.createTextNode(`@${mention.name}`));
+    pending.shift();
+  }
+
+  const container = root.ownerDocument.createElement("div");
+  // When the selection starts and ends directly inside a chip's text node,
+  // cloneContents() has no ancestor span to rewrite. Fall back to escaped
+  // canonical plain text for that boundary-only case instead of leaking the
+  // presentation label back into a rich paste.
+  if (pending.length > 0) {
+    container.textContent = copyTextWithMentionHandles(root, selection) ?? selection.toString();
+    return container.innerHTML;
+  }
+  container.append(fragment);
+  return container.innerHTML;
 }
--- a/packages/web/src/components/rehype-mentions.ts
+++ b/packages/web/src/components/rehype-mentions.ts
@@ -55,7 +55,6 @@ function makeMentionElement(identity: RenderedMentionIdentity, isSelf: boolean):
   const displayName = identity.displayName.trim() || identity.name;
   const displayLabel = `@${displayName}`;
   const identityLabel = displayName === identity.name ? displayLabel : `${displayLabel} (@${identity.name})`;
-  const visibleLabel = identity.disambiguate ? identityLabel : displayLabel;
 
   return {
     type: "element",
@@ -73,7 +72,24 @@ function makeMentionElement(identity: RenderedMentionIdentity, isSelf: boolean):
       title: identityLabel,
       ariaLabel: identityLabel,
     },
-    children: [{ type: "text", value: visibleLabel }],
+    children: [
+      {
+        type: "element",
+        tagName: "span",
+        properties: { className: ["mention-chip-display"] },
+        children: [{ type: "text", value: displayLabel }],
+      },
+      ...(identity.disambiguate
+        ? [
+            {
+              type: "element" as const,
+              tagName: "span",
+              properties: { className: ["mention-chip-handle"] },
+              children: [{ type: "text" as const, value: `(@${identity.name})` }],
+            },
+          ]
+        : []),
+    ],
   };
 }
 
@@ -177,16 +193,15 @@ export function rehypeMentions(
  * should proceed unchanged.
  */
 type SelectedMention = {
-  start: number;
-  end: number;
   name: string;
   visibleText: string;
+  chip: HTMLElement;
 };
 
 function selectedMentions(root: HTMLElement, selection: Selection | null): SelectedMention[] | null {
   if (!selection || selection.rangeCount === 0 || selection.isCollapsed) return null;
   const range = selection.getRangeAt(0);
-  if (!root.contains(range.commonAncestorContainer)) return null;
+  if (!range.intersectsNode(root)) return null;
 
   const mentions: SelectedMention[] = [];
   const chips = root.querySelectorAll<HTMLElement>(".mention-chip[data-mention-name]");
@@ -219,27 +234,100 @@ function selectedMentions(root: HTMLElement, selection: Selection | null): Selec
       continue;
     }
 
-    // Measure the chip's exact position within the selected text instead of
-    // replacing by label value. Two people may share a displayName, and the
-    // same visible string may also appear as ordinary text before the chip.
-    const prefixRange = range.cloneRange();
-    prefixRange.setEnd(chipRange.startContainer, chipRange.startOffset);
-    const start = prefixRange.toString().length;
-    mentions.push({ start, end: start + visibleText.length, name, visibleText });
+    mentions.push({ name, visibleText, chip });
   }
 
   return mentions.length > 0 ? mentions : null;
 }
 
-export function copyTextWithMentionHandles(root: HTMLElement, selection: Selection | null): string | null {
+type CanonicalSelectionFragment = {
+  fragment: DocumentFragment;
+  mentions: SelectedMention[];
+  pending: SelectedMention[];
+};
+
+/**
+ * Clone the selected DOM once, then replace every complete mention chip in
+ * document order. Plain and rich clipboard payloads are both derived from this
+ * same clone, so rendered line breaks never become offsets into a different
+ * string representation.
+ */
+function selectionFragmentWithMentionHandles(
+  root: HTMLElement,
+  selection: Selection | null,
+): CanonicalSelectionFragment | null {
   const mentions = selectedMentions(root, selection);
   if (!mentions || !selection) return null;
-  mentions.sort((left, right) => right.start - left.start);
 
-  return mentions.reduce(
-    (text, mention) => `${text.slice(0, mention.start)}@${mention.name}${text.slice(mention.end)}`,
-    selection.toString(),
-  );
+  const markerAttribute = "data-mention-copy-index";
+  const previousMarkers = mentions.map(({ chip }) => chip.getAttribute(markerAttribute));
+  for (const [index, mention] of mentions.entries()) {
+    mention.chip.setAttribute(markerAttribute, String(index));
+  }
+
+  let fragment: DocumentFragment;
+  try {
+    fragment = selection.getRangeAt(0).cloneContents();
+  } finally {
+    for (const [index, mention] of mentions.entries()) {
+      const previous = previousMarkers[index];
+      if (previous === null || previous === undefined) mention.chip.removeAttribute(markerAttribute);
+      else mention.chip.setAttribute(markerAttribute, previous);
+    }
+  }
+
+  const pending = [...mentions];
+  for (const chip of fragment.querySelectorAll<HTMLElement>(`[${markerAttribute}]`)) {
+    const index = Number(chip.getAttribute(markerAttribute));
+    const mention = mentions[index];
+    if (!mention) continue;
+    chip.replaceWith(root.ownerDocument.createTextNode(`@${mention.name}`));
+    const pendingIndex = pending.indexOf(mention);
+    if (pendingIndex >= 0) pending.splice(pendingIndex, 1);
+  }
+
+  return { fragment, mentions, pending };
+}
+
+/** Read a detached selection fragment through the browser's rendered-text
+ * algorithm. The temporary node stays off-screen (rather than display:none)
+ * because `innerText` deliberately ignores layout for non-rendered content. */
+function renderedTextFromFragment(root: HTMLElement, fragment: DocumentFragment): string {
+  const container = root.ownerDocument.createElement("div");
+  container.style.position = "fixed";
+  container.style.left = "-200vw";
+  container.style.top = "0";
+  container.style.width = "100vw";
+  container.style.whiteSpace = "pre-wrap";
+  container.append(fragment);
+
+  const body = root.ownerDocument.body;
+  if (!body) return container.textContent ?? "";
+  body.append(container);
+  try {
+    return container.innerText;
+  } finally {
+    container.remove();
+  }
+}
+
+export function copyTextWithMentionHandles(root: HTMLElement, selection: Selection | null): string | null {
+  const rewritten = selectionFragmentWithMentionHandles(root, selection);
+  if (!rewritten || !selection) return null;
+
+  // A range whose common ancestor is the chip's text node clones only that
+  // text, not its wrapping span/data attributes. This is the sole useful
+  // pending shape: the user selected exactly one complete chip.
+  if (
+    rewritten.pending.length === 1 &&
+    rewritten.mentions.length === 1 &&
+    selection.toString() === rewritten.pending[0]?.visibleText
+  ) {
+    return `@${rewritten.pending[0].name}`;
+  }
+  if (rewritten.pending.length > 0) return null;
+
+  return renderedTextFromFragment(root, rewritten.fragment);
 }
 
 /**
@@ -248,28 +336,42 @@ export function copyTextWithMentionHandles(root: HTMLElement, selection: Selecti
  * payload so pasting into Docs/Notion retains links and emphasis.
  */
 export function copyHtmlWithMentionHandles(root: HTMLElement, selection: Selection | null): string | null {
-  const mentions = selectedMentions(root, selection);
-  if (!mentions || !selection) return null;
-
-  const fragment = selection.getRangeAt(0).cloneContents();
-  const pending = [...mentions].sort((left, right) => left.start - right.start);
-
-  for (const chip of fragment.querySelectorAll<HTMLElement>(".mention-chip[data-mention-name]")) {
-    const mention = pending[0];
-    if (!mention || chip.dataset.mentionName !== mention.name || chip.textContent !== mention.visibleText) continue;
-    chip.replaceWith(root.ownerDocument.createTextNode(`@${mention.name}`));
-    pending.shift();
-  }
-
+  const rewritten = selectionFragmentWithMentionHandles(root, selection);
+  if (!rewritten || !selection) return null;
   const container = root.ownerDocument.createElement("div");
   // When the selection starts and ends directly inside a chip's text node,
   // cloneContents() has no ancestor span to rewrite. Fall back to escaped
   // canonical plain text for that boundary-only case instead of leaking the
   // presentation label back into a rich paste.
-  if (pending.length > 0) {
+  if (rewritten.pending.length > 0) {
     container.textContent = copyTextWithMentionHandles(root, selection) ?? selection.toString();
     return container.innerHTML;
   }
-  container.append(fragment);
+  container.append(rewritten.fragment);
   return container.innerHTML;
+}
+
+/**
+ * Observe copy at the document boundary, where the Clipboard API dispatches
+ * non-editable timeline copies (the focused node may be body, the composer, or
+ * another control). The selection itself is the scope boundary: helpers return
+ * null unless a complete mention inside this timeline is selected.
+ */
+export function installTimelineMentionCopy(ownerDocument: Document, timeline: () => HTMLElement | null): () => void {
+  const handleCopy = (event: ClipboardEvent) => {
+    if (event.defaultPrevented || !event.clipboardData) return;
+    const root = timeline();
+    if (!root) return;
+    const selection = ownerDocument.getSelection();
+    const copiedText = copyTextWithMentionHandles(root, selection);
+    if (copiedText === null) return;
+    const copiedHtml = copyHtmlWithMentionHandles(root, selection);
+
+    event.preventDefault();
+    event.clipboardData.setData("text/plain", copiedText);
+    if (copiedHtml !== null) event.clipboardData.setData("text/html", copiedHtml);
+  };
+
+  ownerDocument.addEventListener("copy", handleCopy);
+  return () => ownerDocument.removeEventListener("copy", handleCopy);
 }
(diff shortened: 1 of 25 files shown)


baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923

- Rationale: the canonical-handle copy path is not reached by the normal non-editable timeline copy flow, and display-name mutations do not consistently refresh the chat-detail identity source.

**Risk level**: B-low
- Semantic lift: changes the chat-detail-derived identity projection and clipboard/data consumers -> B-low

**PR summary**
- Problem: Sent chat mentions expose canonical handles that can be hard for teammates to recognize, especially numeric human handles; after merge, history should show friendly participant names without changing routing or making copied mentions unpasteable.
- Approach: Resolve stored `@name` tokens against chat-scoped participants, render `displayName`, visibly disambiguate duplicate labels, retain canonical names in DOM metadata and clipboard payloads, and constrain long chips on narrow screens.
- Impacted modules: Web sent-message Markdown rendering, chat-detail participant projection, clipboard handling, mention-chip CSS, and DOM/unit tests.

**Review findings**
❌ 1. The copy override is attached to a non-focusable, non-editable `MessageBody`, so ordinary timeline copy does not reliably dispatch through this handler. The Clipboard API targets the focused node (or `body`) in a non-editable context, not the element containing the selection; after selecting message text, focus commonly remains on `body` or the composer, so `onCopy` at line 747 is bypassed and native copy leaks the rendered `@displayName` instead of the canonical handle. A selection spanning messages also fails the per-message `root.contains(range.commonAncestorContainer)` guard. Please move the interception to a document/timeline-level listener that is guaranteed to observe copy events, scope it to selections intersecting this timeline, and add interaction coverage for single- and cross-message plain/rich copy. See the [Clipboard API event-target algorithm](https://www.w3.org/TR/clipboard-apis/#fire-a-clipboard-event). [R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:747`, `packages/web/src/components/rehype-mentions.ts:189`]

❌ 2. The rendered label only refreshes when `['chat-detail', chatId]` changes, but the actual human rename paths do not update that cache. Team self-renames invalidate only `members` / `agents`, admin member renames invalidate only `members`, Settings refreshes only `/me`, and the server emits no identity-change frame. The new test sidesteps this integration gap with `queryClient.setQueryData`, so an open chat keeps the old mention label after those supported edits until an unrelated detail invalidation or reload. Please centralize identity-cache invalidation/update (or broadcast an identity event) across every display-name mutation and cover at least one real rename mutation path. [R1/R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:3361`]

**Action taken**
- Submitted request changes.

packages/web/src/components/rehype-mentions.ts
```diff
@@ -110,16 +133,143 @@ function walk(
...
+    ) {
+      continue;
+    }
+
+    // Measure the chip's exact position within the selected text instead of
+    // replacing by label value. Two people may share a displayName, and the
+    // same visible string may also appear as ordinary text before the chip.
+    const prefixRange = range.cloneRange();
+    prefixRange.setEnd(chipRange.startContainer, chipRange.startOffset);
+    const start = prefixRange.toString().length;
```
yuezengwu: **P1 — Plain-text copy uses incompatible string coordinate systems.** `start` is measured with `Range.toString()`, whose DOM algorithm concatenates selected Text-node data, but line 241 applies that offset to `Selection.toString()`, which returns rendered text. Markdown line breaks become `<br>`/block boundaries, so rendered newlines before a chip shift every later replacement; copying a common `line one\n@alice` message can insert `@alice` at the wrong character and leave or cut part of `@Alice Chen`. Build both offsets and output from one representation, and cover `<br>` plus multiple paragraphs in a real-browser copy interaction. Specs: [DOM Range stringification](https://dom.spec.whatwg.org/#dom-range-stringifier), [Selection stringification](https://w3c.github.io/selection-api/#dom-selection-stringifier).

packages/web/src/pages/workspace/center/chat-view.tsx
```diff
@@ -3323,33 +3345,34 @@ export function ChatView({
...
-    return [...mentionParticipants, { agentId: myAgentId, name: selfParticipant.name }];
-  }, [mentionParticipants, myAgentId, chatParticipantById]);
+  // Rendered-message projection: all chat speakers (including self) carry
+  // `displayName` for the visible chip while the immutable `name` remains the
+  // resolver key and copy target. Reading directly from chat detail preserves
+  // the membership disclosure boundary for private agents and excludes
+  // non-speaker watchers.
+  const renderMentionParticipants = useMemo<RenderedMentionParticipant[]>(
+    () =>
+      (chatDetail?.participants ?? []).flatMap((participant) =>
```
yuezengwu, line 3374: **P1 — Current-handle lookup can rewrite historical identity after handle reuse.** The supported delete path sets agent A's `name` to `null` and releases it for reuse. If agent B later claims that handle and joins this chat, this one current `name -> agent` projection resolves A's old `@handle` messages to B, even though the messages' persisted structured recipient IDs point to A. History then visibly changes to B's display name and carries B's agent id. Resolve substitution per message against its persisted mention/addressed IDs, or leave the canonical text unchanged when the current identity does not match the historical target. This is an audit/identity blocker, not only a label edge case.

packages/web/src/index.css
```diff
@@ -2175,15 +2175,18 @@ tr[data-interactive]:hover {
...
- * `999px` radius (instead of `--radius-chip`) gives the full capsule
- * shape without touching the shared chip radius. */
+ * `@<displayName>` from wrapping. Inline-block + ellipsis keeps a very long
+ * display name inside the message column on narrow screens. */
 .mention-chip {
-  display: inline;
+  display: inline-block;
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
yuezengwu: **P2 — End ellipsis hides the only duplicate-name discriminator.** Collision labels are built as `@<long display name> (@handle)`, but this single overflowing inline-block clips from the end. On a narrow/touch screen, two long duplicate display names therefore show the same truncated prefix while both handles are hidden; `title` is not discoverable on touch. Keep the canonical handle in a non-truncating segment (or put it first for collision labels) and cover the duplicate-long-name case at 360px.

yuezengwu (changes requested):
## Review summary

Goal and core changes: this PR keeps immutable `name` as the stored/resolver key while rendering mutable `displayName` in sent-message chips, adds duplicate-name disambiguation and narrow-screen truncation, and tries to restore canonical handles in plain/rich clipboard payloads.

I am requesting changes. I independently confirm the two blockers already raised on this head: the per-message `onCopy` handler is not on the normal non-editable clipboard event path (and cannot cover cross-message selections), and the supported Team/Settings rename mutations do not refresh the `chat-detail` identity source. I found three additional correctness gaps inline: multiline plain-text copy uses incompatible offset coordinates, handle reuse can visibly reassign historical mentions, and ellipsis can hide the only duplicate-name discriminator on touch screens.

The diff has no schema or database change, but the handle-reuse finding affects core identity/audit semantics and needs explicit human scrutiny. Static diff review only; I did not rerun tests or QA, per review instructions.

packages/web/src/components/rehype-mentions.ts
Gandy2025: Fixed in c766dd98. Copy now clones the selected DOM once, marks the exact selected mention nodes before cloning, rewrites those cloned chips to canonical handles, and derives both `text/plain` and `text/html` from that same fragment. This removes the incompatible Range/Selection offset conversion. Coverage now includes `<br>`, multiple paragraphs, single-message and cross-message copy, and selections that enter or leave the timeline.

packages/web/src/pages/workspace/center/chat-view.tsx
Gandy2025: Fixed in c766dd98. Sent-message rendering now resolves a token only from that message's persisted `metadata.mentions` IDs. `addressedAgentIds` is deliberately not treated as token identity evidence. If the stored mention ID no longer matches a current participant, the historical message keeps its original canonical `@handle` text. The regression test covers handle reuse plus a misleading current-owner `addressedAgentIds` entry.

packages/web/src/index.css
Gandy2025: Fixed in c766dd98. Duplicate labels now use separate display-name and canonical-handle flex segments: the display name can ellipsize, while the handle stays visible and may wrap at safe boundaries instead of overflowing. Real Chromium QA at 360px with a valid 64-character handle confirmed `document.scrollWidth === 360`, the chip remained inside the timeline, and the full handle stayed visible.

Gandy2025:
@baixiaohang @yuezengwu — addressed all requested changes in c766dd98:

1. Copy interception now lives at the document boundary and is scoped by timeline intersection. Single-message, cross-message, inside-to-outside, and outside-to-inside selections preserve canonical handles in plain and rich clipboard payloads.
2. A centralized display-name invalidation helper now covers every identified Web identity projection. Settings, Team self/admin, Agent Detail, and combined self rename + role mutation paths refresh the relevant React Query caches; self updates also refresh AuthProvider's `/me` snapshot.
3. Plain and rich copy are generated from the same cloned DOM fragment, removing multiline coordinate mismatches.
4. Historical mention labels resolve only from the message's persisted `metadata.mentions` IDs. Handle reuse or legacy/missing identity evidence falls back to the original `@handle`.
5. Duplicate display names keep a visible canonical-handle segment; valid long handles wrap without 360px viewport overflow.

Validation:

- Targeted mention/rename/copy suites: 72/72
- Final Team + identity-cache regression suites: 8/8
- Full Web suite: 1,578/1,578
- Web typecheck and design-token guardrails: passed
- Real Chromium QA: single/cross-message canonical copy and 360px long-handle layout passed
- Two final independent read-only reviews: no blockers

Ready for re-review.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923 (successor head `c766dd98`)

- Rationale: the five prior findings are addressed, but the historical-ID allowlist now also narrows the duplicate-name collision set, so the advertised disambiguation disappears in the normal one-recipient case.

**Risk level**: B-low
- Semantic lift: changes the chat-detail-derived identity projection and clipboard/data consumers -> B-low

**PR summary**
- Problem: Sent chat mentions should use recognizable display names while retaining immutable handles for identity, audit, routing, and copy/paste.
- Approach: The successor head moves copy interception to the document boundary, derives plain/rich copy from one cloned fragment, gates historical label substitution on persisted mention IDs, centralizes identity-cache invalidation, and separates truncatable display text from the visible canonical handle.
- Impacted modules: Web sent-message Markdown rendering, per-message identity projection, clipboard handling, identity cache invalidation, mention-chip CSS, and related tests.

**Review findings**
❌ 1. `historicalMentionParticipants` is being used for two different decisions: which current identity is allowed to resolve this message's token, and which chat participants count as display-name collisions. Because it contains only IDs from this message's `metadata.mentions`, a chat with `alice -> "Sam"` and `bob -> "Sam"` passes only Alice to `rehypeMentions` for a message that mentions Alice alone. `displayNameCounts` therefore sees one `Sam`, `disambiguate` stays false, and the message renders plain `@Sam` with no visible `@alice`; the new non-truncating handle segment never exists. This is the common one-recipient case, so duplicate display names remain ambiguous on touch despite the PR contract. Keep the full current speaker roster as the collision universe while separately restricting token resolution to persisted mention IDs (for example, pass an allowed-agent-id set alongside all participants), and cover ChatView with two same-display-name speakers where only one is present in the message metadata, including the long-label variant. [R1/R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:617`, `packages/web/src/components/rehype-mentions.ts:159`]

**Action taken**
- Submitted request changes on successor head `c766dd98`.

Gandy2025:
@baixiaohang — fixed the successor-head finding in a4c5d0e7.

The renderer now receives two independent identity inputs:

- the full current chat-speaker roster, used only to count duplicate `displayName` collisions;
- an `allowedAgentIds` set derived only from the message's persisted `metadata.mentions`, used to decide which canonical token may resolve.

This keeps the visible `(@handle)` discriminator when only one of two same-label speakers is mentioned, while preserving the historical handle-reuse safety boundary. Added:

- a plugin regression that counts both `Sam` speakers but resolves only the persisted recipient;
- a ChatView regression with two long identical display names, one persisted recipient, and assertions for the visible canonical handle and full title.

Targeted suites pass 73/73, Web typecheck/design-token guardrails pass, and two independent final read-only reviews found no blockers. Ready for another successor-head review.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1923

**Recommendation**: approve
- Rationale: The successor head fixes the remaining duplicate-label identity gap without weakening the persisted-ID boundary, and the earlier copy, rename-refresh, historical-identity, and narrow-screen fixes remain intact.

**Risk level**: B-low
- Path baseline: `packages/web/**` -> C.
- Semantic lift: The Web change affects chat-detail identity projection and clipboard consumers -> B-low.

**PR summary**
- Problem: Sent chat mentions expose canonical handles even when a participant has a clearer display name, while copied content and duplicate labels still need an unambiguous, route-safe identity.
- Approach: Render mutable display names only in sent-message presentation, keep immutable handles for resolver and clipboard behavior, authorize historical token resolution from persisted mention IDs, count duplicate labels across the full chat speaker roster, and invalidate identity-bearing Web caches after renames.
- Impacted modules: Web chat timeline mention rendering/copy, identity cache invalidation, Settings, Team, Agent Detail, and focused DOM/plugin regressions.

**Review findings**
✅ 1. The new `allowedAgentIds` boundary is independent from full-roster collision counting, so a message that mentions only one of two same-label speakers still shows the canonical discriminator without allowing a reused handle to relabel history.
✅ 2. Document-boundary copy interception and DOM-fragment rewriting still cover ordinary and cross-message selections while preserving canonical handles in plain and rich payloads.
✅ 3. Display-name mutation paths still converge on centralized invalidation, including `chat-detail`, with the separate authenticated-user snapshot refreshed for self renames.
✅ 4. The split display/handle chip layout retains a visible discriminator on narrow screens and permits exceptionally long handles to wrap.

**Action taken**
- Approved successor head `a4c5d0e7aa467271d1ce2b37cf7c7e6d8e16562b` after static review. I did not rerun tests or visual QA; the author-provided validation and repository CI remain the execution evidence.

Gandy2025:
@yuezengwu — the latest successor head is a4c5d0e7. All five findings from your review are addressed, and the follow-up duplicate-label collision issue is now fixed by separating full-roster collision counting from the persisted-ID resolution allowlist. @baixiaohang has approved this exact head. Could you please re-review so the earlier changes-requested state can be cleared if you agree?

packages/web/src/pages/agent-detail.tsx
```diff
@@ -130,7 +131,11 @@ function AgentDetailPageView() {
 
   const identityUpdateMutation = useMutation({
     mutationFn: (patch: Parameters<typeof updateAgent>[1]) => updateAgent(uuid, patch),
-    onSuccess: () => {
+    onSuccess: async (_agent, patch) => {
+      if (patch.displayName !== undefined) {
```
yuezengwu, line 135: This invalidation cannot make a human-mirror rename consistent because the mutation above always calls generic `updateAgent`. The Profile UI permits human edits, but `updateAgent` writes only `agents.display_name`; `listMembers` and AuthProvider still derive the human label from `users.display_name`. After success, chat/agent queries show the new value while Team and the user menu retain the old one. Route self/admin human edits through the profile/member identity paths, update both sources transactionally, or disallow this field here; please add a real human Agent Detail rename regression.

packages/web/src/components/rehype-mentions.ts
```diff
@@ -110,16 +150,235 @@ function walk(
...
+/** Read a detached selection fragment through the browser's rendered-text
+ * algorithm. The temporary node stays off-screen (rather than display:none)
+ * because `innerText` deliberately ignores layout for non-rendered content. */
+function renderedTextFromFragment(root: HTMLElement, fragment: DocumentFragment): string {
+  const container = root.ownerDocument.createElement("div");
+  container.style.position = "fixed";
+  container.style.left = "-200vw";
+  container.style.top = "0";
+  container.style.width = "100vw";
+  container.style.whiteSpace = "pre-wrap";
```
yuezengwu: Forcing `pre-wrap` changes unrelated selected text. The live Markdown prose inherits normal whitespace, so repeated inline spaces/tabs collapse in rendering and native copy; this off-screen clone preserves them whenever the same selection also contains a mention. For example, `Please   ask @alice` emits three spaces instead of the one rendered separator. The canonical-copy path should replace only chip labels while inheriting/mirroring the live whitespace semantics, with a regression for collapsible inline whitespace.

yuezengwu (changes requested):
## Re-review of successor head `a4c5d0e7`

The implementation goal remains sound: render sent `@name` tokens with mutable `displayName` while keeping canonical handles and persisted recipient IDs authoritative for identity, routing, audit, and copy/paste. This head fixes the five original findings and correctly separates the full-roster duplicate-name collision set from the per-message persisted-ID allowlist.

Two blockers remain:

1. Agent Detail still routes human-mirror display-name edits through generic `PATCH /agents/:uuid`. That writes only `agents.display_name`, while Team/member and AuthProvider identity read `users.display_name`; the new broad invalidation then refetches two conflicting labels. Human identity edits must use a path that keeps the user and human mirror aligned (or the generic route must update both transactionally).
2. Plain canonical-copy forces the detached fragment to `white-space: pre-wrap`. The live timeline uses normal whitespace, so collapsible inline spaces/tabs outside the mention are preserved only by the override. Selecting `Please   ask @alice` therefore copies different non-mention text solely because a mention is present. The override should change only mention labels and preserve live rendered whitespace semantics.

There is no DB/schema diff. Finding 1 nevertheless touches the core human identity source-of-truth and needs human scrutiny. I reviewed the diff statically and did not rerun tests or QA.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923

- Rationale: I am superseding my approval of this exact head after independently confirming two remaining correctness regressions in human identity persistence and plain-text copy fidelity.

**Risk level**: B-low
- Path baseline: `packages/web/**` -> C.
- Semantic lift: The Web change affects chat-detail identity projection and clipboard consumers -> B-low.

**PR summary**
- Problem: Sent chat mentions should use readable display names without weakening canonical identity, routing, audit history, or pasteable handles.
- Approach: Project current chat-speaker display names onto persisted mention IDs, disambiguate duplicate labels, rewrite copied chips to canonical handles, and refresh identity-bearing caches after rename mutations.
- Impacted modules: Web chat mention rendering/copy, identity cache invalidation, Agent Detail, Team, Settings, and focused regressions.

**Review findings**
❌ 1. Agent Detail still sends a human mirror's display-name edit through generic `PATCH /agents/:uuid`. `packages/web/src/pages/agent-detail.tsx:133` calls `updateAgent` for every identity patch, while `packages/server/src/services/agent.ts:1086` writes only `agents.display_name`. Team/member and AuthProvider identity read the authoritative `users.display_name`; the supported member/profile paths update both fields transactionally. Broad invalidation after this write therefore refetches conflicting names instead of converging identity. Human display-name edits from Agent Detail need a path that keeps the user and mirror aligned (or the generic server path must do so atomically). [R1 / `packages/web/src/pages/agent-detail.tsx:133`]
❌ 2. Canonical plain copy changes unrelated selected text by forcing the detached fragment to `white-space: pre-wrap`. The live markdown timeline uses normal whitespace collapsing, so selecting content such as `Please   ask @alice` preserves the extra spaces only when a mention triggers this override. The copy rewrite must replace mention labels without changing the live rendered whitespace semantics of surrounding text. [R5 / `packages/web/src/components/rehype-mentions.ts:308`]
✅ 3. The successor-head split between full-roster collision counting and persisted-ID resolution remains correct and resolves the previous duplicate-label blocker.

**Action taken**
- Submitted changes requested on `a4c5d0e7aa467271d1ce2b37cf7c7e6d8e16562b`, superseding my earlier approval after static source-path verification. I did not rerun tests or QA.

Gandy2025:
@yuezengwu @baixiaohang Both remaining blockers are fixed on successor head `25730ba4`.

1. Human display-name persistence now follows the authoritative identity path from Agent Detail:
   - the current human uses `PATCH /me/profile` and refreshes the AuthProvider snapshot;
   - another human uses the admin-gated member mutation;
   - non-human names and other agent fields continue using `PATCH /agents/:uuid`.
2. The member mutation now updates `users.display_name` and every membership-backed human-agent mirror for that user in the same transaction. This closes the multi-organization convergence gap as well as the immediate Agent Detail mismatch.
3. Canonical plain copy no longer forces `white-space: pre-wrap`; the detached render root inherits the live timeline's computed whitespace mode, so only mention labels change.

New regression coverage includes self-human, other-human, and non-human Agent Detail routes; a genuine multi-organization server mirror case; and surrounding spaces/tabs in canonical copy. Real Chromium QA produced `Please ask @alice today` from a selection containing collapsed spaces and a rendered `@Alice Chen` chip.

Validation on this exact diff:

- Web full suite: 193 files, 1,585 tests passed.
- Server identity suites: 3 files, 37 tests passed.
- Web and Server typechecks passed; design-token guardrails and `git diff --check` are clean.
- Two independent final read-only reviews found no blockers.

Please re-review `25730ba4` when ready.

baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1923

**Recommendation**: approve
- Rationale: The successor head resolves the remaining human-identity and copy-fidelity blockers while preserving the previously verified mention-resolution, clipboard-boundary, duplicate-label, and narrow-screen behavior.

**Risk level**: B-low
- Path baseline: Includes `packages/server/src/services/member.ts` outside the B-high service set, plus Web changes -> B-low.
- Semantic lift: None beyond the B-low path baseline; the affected Web projections remain scoped to identity labels and clipboard rendering.

**PR summary**
- Problem: Sent chat mentions need readable display names without weakening canonical routing/audit identity, losing pasteable handles, creating ambiguous duplicate labels, or leaving renamed human identities inconsistent across product surfaces.
- Approach: Render mutable display names only for persisted mention identities, count label collisions across the full chat roster, rewrite copied chips to immutable handles, route human renames through user/member authority, synchronize every membership-backed human mirror transactionally, and refresh identity-bearing Web caches.
- Impacted modules: Chat timeline mention rendering/copy, member identity service, Agent Detail, Team, Settings, identity cache invalidation, and focused Web/server regressions.

**Review findings**
✅ 1. Agent Detail now routes a self human through `/me/profile`, another human through the admin-gated member mutation, and non-human display names through the agent mutation; the authenticated-user snapshot refresh remains limited to self edits.
✅ 2. The member mutation updates `users.display_name` and every membership-backed human mirror for that user inside one transaction, matching the user-global identity model across Teams.
✅ 3. Canonical plain copy now inherits the live timeline root's computed whitespace mode, so surrounding collapsible spaces/tabs retain native rendered semantics while mention labels alone become canonical handles.
✅ 4. Persisted mention IDs remain the historical resolution allowlist, while the full speaker roster independently drives duplicate-label disambiguation and visible narrow-screen handles.

**Action taken**
- Approved successor head `25730ba4322a09973a165d21efd4cd6903d1f06f` after static review. I did not rerun tests or visual QA; the author-provided validation and repository CI remain the execution evidence.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1923 (successor head `25730ba4`)

The successor head fixes the two previously reported paths: Agent Detail now routes human renames through user/member authority, and canonical plain copy preserves the live timeline's whitespace behavior. However, the new server-side identity synchronization still has two convergence holes, so the same global user can continue to present different display names across membership-backed human mirrors.

**Implementation goal and core change**

This PR renders sent-message mentions with mutable, readable `displayName` labels while preserving immutable canonical handles for resolution, audit, and clipboard content. The successor additionally changes the core human-identity write path so a member rename updates the global user label and all of that user's membership-backed human-agent mirrors transactionally. There is no database/schema change, but this is core identity persistence behavior and needs the same invariant across every supported write path.

**Findings**

1. **An explicit rename cannot repair an already-drifted mirror when the requested value equals the authoritative user label.** In `packages/server/src/services/member.ts:235`, the fan-out runs only when `data.displayName !== current.displayName`, where `current.displayName` comes from `users.display_name`. If production already has `users.display_name = "Alice"` but one human mirror is `"Alicia"` (which the former Agent Detail generic PATCH could create), submitting `"Alice"` succeeds without touching any mirror. Agent Detail then refetches the still-stale `"Alicia"`, so the advertised convergence does not occur. Treat every explicitly supplied `displayName` as a synchronization request (or provide an equivalent repair/backfill), and add a regression with a deliberately stale mirror whose requested value already matches the user row.

2. **The fan-out is local to `updateMember`, while other supported member display-name writes still bypass it.** `createMember` reactivation (`packages/server/src/services/member.ts:69`) updates the global user and only the restored mirror; adding an existing global user to another organization (`packages/server/src/services/member.ts:94`) creates the new mirror from the request without updating the user or existing mirrors. Both paths can therefore leave or immediately create cross-organization display-name divergence, which this PR now exposes in rendered mentions. Centralize the global-user + all-membership-mirror synchronization (or consistently use the existing global label when adding a membership) across PATCH, restore, and existing-user add, with multi-organization coverage for those paths.

The Web routing and authorization split is otherwise sound: self-human edits use `/me/profile`, other-human edits use the admin-gated member route, and non-human edits stay on `/agents/:uuid`. The clipboard whitespace fix also addresses the prior copy-fidelity blocker without changing rich-copy escaping or listener cleanup. The persisted-ID mention boundary and full-roster collision counting remain intact.

**Action taken**

Requested changes on exact head `25730ba4322a09973a165d21efd4cd6903d1f06f` after static review. I did not rerun tests or visual QA; the author-provided results and repository CI remain the execution evidence.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923

- Rationale: I am superseding my approval of this head after independently confirming that the new user-to-human-mirror synchronization is not yet a single invariant across explicit repair and supported membership lifecycle writes.

**Risk level**: B-low
- Path baseline: Includes `packages/server/src/services/member.ts` outside the B-high service set, plus Web changes -> B-low.
- Semantic lift: None beyond the B-low baseline; the remaining issue is within the newly changed member identity persistence path.

**PR summary**
- Problem: Sent mentions should show readable display names while immutable handles and persisted IDs remain authoritative, and human labels must not diverge between the global user and Team-scoped mirrors.
- Approach: Project display names only onto persisted mention identities, preserve canonical clipboard payloads, route human renames through user/member authority, and fan a member rename out to all mirrors.
- Impacted modules: Chat mention rendering/copy, member identity service, Agent Detail, Team, Settings, identity cache invalidation, and focused regressions.

**Review findings**
❌ 1. An explicit display-name write cannot repair an already-drifted mirror when the requested value equals the authoritative user label. `current.displayName` comes from `users.display_name`, and the guard at `packages/server/src/services/member.ts:235` skips both the user write and the entire mirror fan-out when those strings match. A prior generic human-agent edit can therefore leave user `Alice` / mirror `Alicia`; submitting `Alice` reports success but leaves the mirror unchanged and Agent Detail refetches `Alicia`. Every explicitly supplied display name must act as a synchronization request, or an equivalent repair/backfill must cover this state. [R4 / `packages/server/src/services/member.ts:235`]
❌ 2. The synchronization invariant exists only inside `updateMember`, while supported member creation/reactivation paths still write different subsets. `createMember` reactivation updates the global user and only the restored mirror (`packages/server/src/services/member.ts:69`), and adding an existing global user to a new Team creates the new mirror from the request without updating the user or existing mirrors (`packages/server/src/services/member.ts:94`). The sibling `ensureMembership` / `reactivateMembership` flow likewise owns separate display-name writes. Centralize the user-global label plus all membership-backed mirror synchronization, or consistently use the authoritative global label for membership creation, so PATCH, restore, and existing-user add cannot reintroduce cross-Team divergence. [R1 / `packages/server/src/services/member.ts:50`]
✅ 3. The Agent Detail self/other/non-human routing split and authenticated-user refresh are otherwise sound, and the clipboard whitespace fix correctly preserves live rendered semantics.
✅ 4. Persisted-ID mention resolution, full-roster collision counting, document-boundary copy, and visible narrow-screen disambiguation remain intact.

**Action taken**
- Submitted changes requested on exact head `25730ba4322a09973a165d21efd4cd6903d1f06f`, superseding my earlier approval after static source-path verification. I did not rerun tests or visual QA.

Gandy2025:
@yuezengwu @baixiaohang — both lifecycle-convergence findings are fixed on successor head `8ecf2a09`.

1. Every explicit display-name write is now a synchronization request, even when the requested value already equals `users.display_name`. The write updates the user-global label and every membership-backed human mirror in one transaction.
2. A centralized identity helper now owns the invariant across admin/self rename, create, restore, existing-user add, active ensure, OAuth/team join, and Landing service membership paths. Generic Agent PATCH rejects display-name edits for membership-backed mirrors, while standalone humans and non-human agents retain their existing behavior.
3. Rename intent is separated from lifecycle projection intent. Profile/admin rename may supply a desired label; OAuth, Team join/create, and Landing flows read the authoritative label only after locking the user, so a stale pre-lock snapshot cannot overwrite a newer committed rename.
4. Identity graph mutations use one lock order: user (`FOR NO KEY UPDATE`) → member/unique slot → agents (stable `agentId` order). Startup repair follows the same order and re-reads current user/member/agent state after lock waits, preventing stale-snapshot writeback.
5. Startup repair now also backfills legacy mirror-only display-name drift from the authoritative user row.

Regression coverage includes same-value repair, multi-Team fanout, create/add/restore/active ensure, generic human PATCH rejection, and four real dual-connection PostgreSQL races: reactivation, unique-slot competition, stale lifecycle snapshots, and startup repair versus rename.

Validation on this exact diff:

- Full Server suite: 238 files, 2,574 tests passed.
- Focused identity/lifecycle suite: 6 files, 101 tests passed.
- Server typecheck, Biome, and `git diff --check`: passed.
- Two independent final read-only reviews: approved with no blockers.

Please re-review `8ecf2a09` when ready.

packages/web/src/pages/agent-detail.tsx
Gandy2025: Fixed across 25730ba4 and 8ecf2a09. Agent Detail routes self-human renames through /me/profile and other-human renames through the admin member endpoint; non-human edits remain on the agent endpoint. The Server now also rejects generic display-name PATCHes for any membership-backed mirror, so direct API use cannot recreate drift. Explicit profile/member renames synchronize users.display_name and every membership mirror transactionally, with real self/other/non-human and multi-Team regressions.

packages/web/src/components/rehype-mentions.ts
Gandy2025: Fixed in 25730ba4. Canonical plain copy now inherits the live timeline root's computed white-space mode before replacing mention labels, so surrounding collapsible spaces and tabs keep their native rendered semantics. The regression covers repeated spaces/tabs, and real Chromium QA copied exactly `Please ask @alice today`.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923

- Rationale: The centralized display-name synchronization now closes the reported projection gaps, but admin reactivation still applies a membership snapshot captured before the identity lock and is not linearizable with concurrent rejoin/restore operations.

**Risk level**: B-high
- Path baseline: the successor head modifies `packages/server/src/services/agent.ts`, a core agent service path.
- Semantic lift: none beyond the B-high path baseline.

**PR summary**
- Problem: Sent mentions should show readable mutable display names without changing canonical addressing, audit identity, or clipboard handles; human display-name edits must also converge across the global user and every membership-backed human mirror.
- Approach: The PR resolves rendered labels only from persisted mention identity, rewrites copied labels to canonical handles, routes human renames through user/member authority, and now centralizes user-to-mirror synchronization and lifecycle locking.
- Impacted modules: Web mention rendering/copy and identity cache invalidation; Server member, membership, Agent, OAuth, and Landing membership services.

**Review findings**

❌ 1. Admin restore still consumes a pre-lock membership state and can overwrite a concurrently completed rejoin. `createMember` reads and classifies `existingMember` before opening the transaction (`packages/server/src/services/member.ts:54-75`). `reactivateMembership` then acquires the user lock but passes that old row directly to `reactivateMembershipRows`, which unconditionally sets the membership active/role (`packages/server/src/services/membership.ts:231-267`) without re-reading its status under the lock. If a concurrent `ensureMembership` or another restore commits while this request waits on the user, the waiting admin request still succeeds and applies its role and global display name; the equivalent serialized request would hit the active-member conflict at `member.ts:62-63`. Re-read and validate the membership inside the user-locked callback before mutation, and add a two-connection regression where a self rejoin wins while admin create/restore waits, asserting that the stale contender cannot perform a second restore or rename. [R4]

✅ The explicit same-value repair, existing-user add, authoritative-label ensure path, full mirror fan-out, startup drift repair, and generic membership-backed Agent PATCH rejection are otherwise consistent with the identity invariant. The earlier mention, copy, and Web routing fixes are unchanged on this successor commit.

**Action taken**
- Requested changes on successor head `8ecf2a0933d748d3e83eb94e0c3ff68b9095be34` after static review. I did not run tests or visual QA; author-provided results and repository CI remain the execution evidence.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1923 (successor head `8ecf2a09`)

The successor resolves both prior convergence findings: same-value writes now repair drift, supported member lifecycle paths project the locked authoritative label, and generic Agent PATCH no longer bypasses membership identity authority. However, the new startup-repair lock strategy can deadlock with ordinary member leave/removal during a rolling or multi-replica deployment.

**Implementation goal and core change**

This PR renders sent-message mentions with readable `displayName` labels while retaining immutable handles and persisted mention IDs for routing, history, and clipboard payloads. This successor centralizes human display-name persistence across profile/admin rename, membership creation/reactivation, OAuth/team join, Landing service membership, and startup repair. There is no database/schema change, but the roughly 800-line successor changes the core identity graph's transaction and lock protocol, so concurrency and recovery behavior are part of the correctness boundary.

**Finding**

1. **`repairMembershipHumanMirrors` can deadlock with `deleteMember` or `deactivateMembership`.** The repair keeps one transaction open for the entire membership table and accumulates member locks in `(userId, memberId)` order (`packages/server/src/services/membership.ts:459-493`). Leave/removal instead locks the target organization's active-admin member rows by `memberId`, then locks the target member (`packages/server/src/services/membership.ts:368-381`; `packages/server/src/services/member.ts:286-310`).

   A concrete cycle exists when ordinary member T's `userId` sorts before admin A's: repair locks T and retains that lock; concurrent delete/leave of T locks A and waits for T; repair advances to A and waits for A. Repair now holds T while waiting for A, while lifecycle holds A while waiting for T. PostgreSQL must abort either the live request or the new replica's `onReady` repair, potentially failing startup during a rolling deployment. The claimed common identity lock order therefore is not shared by all supported lifecycle writers. Scope repair transactions per user/member (releasing T before requesting A), or otherwise make startup repair use a lock protocol compatible with leave/removal, and add a real dual-connection regression for this exact cycle.

The remaining successor paths reviewed cleanly: explicit rename and lifecycle projection intent are separated correctly; all current helper callers provide a surrounding transaction/savepoint; OAuth/team/Landing callers reread the authoritative label after the user lock; the membership-backed generic Agent PATCH guard closes the prior bypass; and earlier mention/copy/disambiguation behavior remains unchanged.

**Action taken**

Requested changes on exact head `8ecf2a0933d748d3e83eb94e0c3ff68b9095be34` after static review. I did not rerun tests, concurrency fixtures, or visual QA; the author-provided validation and repository CI remain the execution evidence.

Gandy2025:
@baixiaohang @yuezengwu — both exact-head concurrency findings are fixed on successor head `7a87f0d8`.

1. **Stale admin restore after self-rejoin:** `reactivateMembership` now re-reads and `FOR UPDATE`-locks the current membership inside the user-locked callback. `active` conflicts, unsupported states conflict, and only the locked current `left`/`removed` row may be restored. A real two-writer PostgreSQL regression pauses `ensureMembership` after it owns the user lock, lets `createMember` classify the stale `left` row and wait, then commits self-rejoin first. The admin restore rejects without changing the winning user label, member role/status, or mirror.
2. **Startup repair versus leave/removal:** `repairMembershipHumanMirrors` now scans candidate IDs without locks and repairs each membership in its own short transaction, still using `user → member → agent`. It therefore releases target T before advancing to admin A and cannot accumulate the cross-user locks needed for the reported cycle. A real PostgreSQL regression constructs the old `repair holds T / removal holds A` cycle and verifies both operations complete with the member removed and mirror suspended.

Validation on this exact diff:

- Full Server suite: 238 files, 2,576 tests passed.
- Related member/membership suites: 9 files, 142 tests passed.
- Server typecheck, Biome, and `git diff --check`: passed.
- Both new concurrency tests passed repeatedly, including three additional consecutive runs in independent review.
- Two independent final read-only reviews: approved with no blockers.

Please re-review successor head `7a87f0d8` when ready.

yuezengwu (approved):
## Review by yzw-codex — PR #1923 (successor head `7a87f0d8`)

**Recommendation**: approve

The successor fixes both exact-head concurrency blockers without weakening the centralized display-name invariant or the earlier mention/copy behavior.

**Implementation goal and core change**

This PR renders sent-message mentions with readable mutable `displayName` labels while preserving immutable handles and persisted mention IDs for routing, audit history, and clipboard payloads. The Server side now keeps the global user label and every membership-backed human mirror aligned across profile/admin rename and membership lifecycle paths. There is no database/schema or core data-structure change, but the successor changes core identity transaction/lock semantics, so I reviewed the failure, rollback, and rolling-deployment boundaries explicitly.

**Review findings**

1. `reactivateMembership` now re-reads and `FOR UPDATE`-locks the current membership inside the user-locked callback. A self-rejoin or competing restore that wins first is observed as `active`; the stale admin contender conflicts, and its tentative label write rolls back with the surrounding transaction. Unsupported lifecycle states also conflict before role or mirror mutation.
2. `repairMembershipHumanMirrors` now performs an unlocked candidate scan and repairs each membership in a short transaction using `user → member → agent`. Target T is released before repair can advance to admin A, eliminating the prior repair-holds-T/removal-holds-A cycle. Stale candidates are re-read under lock, and concurrent rename/reactivation or multiple startup replicas converge safely.
3. The two new dual-connection PostgreSQL regressions establish the intended wait points and would reproduce the old stale-restore and repair/removal failures. They assert the winning identity/lifecycle state and clean up their external blockers.
4. Previously verified same-value repair, create/restore/add/ensure convergence, generic membership-backed Agent PATCH rejection, Web routing, canonical copy, persisted-ID resolution, collision disambiguation, and narrow-screen behavior remain intact.

The per-membership startup transactions add scale-dependent overhead, but that is a non-blocking operational tradeoff; correctness and recovery remain idempotent.

**Action taken**

Approved exact head `7a87f0d8a38e0ab9eff6a8b87d78aaca53f89538` after static review. I did not rerun tests, concurrency fixtures, or visual QA; the author-provided validation and repository CI remain the execution evidence.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1923

- Rationale: A membership-backed human agent is a projection of a user/member identity, so Agent Detail should not present that projection as an independent display-name editing surface.

**Risk level**: B-high
- Path baseline: the cumulative diff modifies `packages/server/src/services/agent.ts`, a core agent service path.
- Semantic lift: none beyond the B-high path baseline.

**PR summary**
- Problem: Sent mentions should display readable mutable labels while immutable handles and persisted recipient IDs remain authoritative for routing, audit history, and clipboard content.
- Approach: The PR adds display-name rendering and disambiguation, canonical copy behavior, cache invalidation, and a centralized user-to-human-mirror synchronization path.
- Impacted modules: Web mention rendering/copy and Agent Detail identity editing; Server member, membership, Agent, OAuth, and Landing identity lifecycle services.

**Review findings**

❌ 1. Agent Detail still exposes a global human-identity rename as if it were an agent profile edit. `ProfileEditDialog` renders the same editable display-name input for human and non-human records (`packages/web/src/pages/agent-detail/profile-edit-dialog.tsx:380-390`), while the parent silently reroutes a human edit to `updateMyProfile` or `updateMember` (`packages/web/src/pages/agent-detail.tsx:133-160`). The backend correctly rejects direct display-name PATCHes for membership-backed mirrors, but this UI preserves the misleading product model that the mirror itself is editable and turns an org-scoped Agent Detail action into a global user rename across every membership. Make the display name read-only (or omit the editor) for membership-backed humans, remove the profile/member rename routing from Agent Detail, and keep human renames on their authoritative surfaces: Settings for self-service and Team member management for admins. Retain the generic Agent PATCH guard and the user-to-mirror projection sync for those authoritative writes. [R1 / R3]

✅ The successor head fixes the previously reported stale-restore and startup-repair lock cycles, and the earlier persisted-ID mention, canonical-copy, duplicate-label, handle-reuse, and cache-refresh behavior remains intact.

**Action taken**
- Requested changes on exact head `7a87f0d8a38e0ab9eff6a8b87d78aaca53f89538` after static review. I did not run tests or visual QA; author-provided results and repository CI remain the execution evidence.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1923 (exact head `7a87f0d8`)

I am superseding my approval after the human owner resolved the remaining product-boundary question: a membership-backed human agent is a projection of user/member identity, so Agent Detail must not expose its display name as an editable agent-profile field.

**Implementation goal and core change**

This PR renders sent-message mentions with readable mutable `displayName` labels while retaining immutable handles and persisted mention IDs for routing, audit history, and clipboard payloads. Its Server work correctly centralizes the global user-to-human-mirror projection and rejects generic display-name PATCHes for membership-backed mirrors. There is no database/schema or core data-structure change.

**Finding**

1. `ProfileEditDialog` still renders an editable display-name field for membership-backed humans, and `AgentDetailPage` silently reroutes that apparent agent edit to the global self-profile/member mutation. The decided product boundary is now explicit: make the human display name read-only (or omit that editor) in Agent Detail, remove the profile/member rename routing from this surface, and keep human renames on the authoritative UI paths—Settings for self-service and Team member management for admins. Preserve the centralized Server projection synchronization and the generic Agent PATCH guard, since those enforce the underlying identity invariant.

The concurrency fixes on this head remain correct: stale admin restore is rejected after a winning rejoin, startup repair no longer accumulates cross-user locks, and the earlier mention/copy/disambiguation behavior remains intact.

**Action taken**

Requested changes on exact head `7a87f0d8a38e0ab9eff6a8b87d78aaca53f89538`, superseding my prior approval after the human product decision. I did not rerun tests or visual QA.

Gandy2025:
@baixiaohang @yuezengwu The Agent Detail product-boundary finding is addressed on successor head `241738a648f724bdd4fb588d48ec6d8d5713bbcf`.

- Membership-backed human agents no longer render the Display name editor in Agent Detail.
- Agent Detail no longer reroutes apparent agent-profile edits through `/me/profile` or the member mutation. Human renames remain on the authoritative Settings and Team surfaces.
- Human visibility, delegate, avatar, image, and color edits remain available and continue to use the agent mutation.
- Non-human agent display-name editing remains unchanged.
- The server-side membership-backed Agent PATCH guard and centralized user-to-mirror synchronization remain intact.

Validation on the successor head:

- Two independent final code reviews: approved, no blocking findings.
- Focused Agent Detail DOM regressions: 38 tests passed.
- Full Web suite: 193 files / 1,584 tests passed.
- Web typecheck and design-token guard passed.
- Repository `pnpm check`, `pnpm typecheck`, and Web build passed.
- Real-browser QA on the current source confirmed that the human dialog omits Display name and that changing Visibility sends only `PATCH /api/v1/agents/human-1` with `{\"visibility\":\"private\"}`.

Please re-review the exact successor head when convenient.

yuezengwu (approved):
## Successor-head review — `241738a648f724bdd4fb588d48ec6d8d5713bbcf`

**Recommendation:** approve

The successor implements the agreed product boundary: a membership-backed human is a user/member identity projection, so Agent Detail no longer presents its display name as an independent agent-profile edit. Human renames remain on Settings for self-service and Team member management for admins.

### Findings

- The human branch omits the Display name editor, and `commitName()` has a defensive human guard, so form submit or Done cannot leak the retained label into an Agent PATCH.
- Agent Detail removes the `updateMyProfile` / `updateMember` rerouting and authenticated-user refresh path. Human visibility, delegate, avatar/image, and color edits continue through their existing agent-projection mutations.
- Non-human display-name editing remains unchanged and still uses `PATCH /agents/:uuid`; its page-level regression remains explicit.
- This successor delta is limited to four Web Agent Detail files. It does not alter the server-side membership-backed Agent PATCH guard, centralized user-to-mirror synchronization, persisted-ID mention resolution, canonical copy handling, duplicate-label disambiguation, or the previously reviewed lifecycle locking fixes.

The cumulative PR still touches the core Agent service and identity lifecycle, so its risk remains **B-high**. This successor introduces no DB/schema or core data-structure change.

**Action:** approving this exact head after static incremental review. I did not rerun tests or visual QA; the author-reported validation and repository CI remain the execution evidence.

Gandy2025:
@baixiaohang Friendly follow-up: PR #1923 is currently blocked only by your `CHANGES_REQUESTED` review on the previous head.

The agreed Human Agent Detail boundary is addressed on exact head `241738a648f724bdd4fb588d48ec6d8d5713bbcf`: the human Display name editor and silent profile/member rerouting are removed, while other human projection fields and non-human display-name editing remain unchanged. @yuezengwu has approved this exact head, and all required checks plus CodeQL are green.

Could you please re-review the successor head when available? We are ready to squash merge as soon as the review state is cleared.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact successor head implements the agreed human-identity boundary and preserves the previously reviewed mention, clipboard, cache, and lifecycle invariants without introducing a new blocker.

**Risk level**: B-high
- Path baseline: The cumulative PR modifies `packages/server/src/services/agent.ts` and the member/membership identity lifecycle.
- Semantic lift: None beyond the B-high baseline; there is no database/schema or core data-structure change.

**PR summary**
- Problem: Sent chat mentions need readable mutable display labels without weakening immutable addressing, historical identity, or pasteable canonical handles.
- Approach: Render display names only when the message's persisted mention IDs still identify the current canonical-handle owner; disambiguate collisions, rewrite copied chips to canonical handles, refresh identity projections after authoritative renames, and keep membership-backed human names on Settings/Team rather than Agent Detail.
- Impacted modules: Web mention rendering and clipboard handling, identity cache invalidation, Team/Settings/Agent Detail identity surfaces, and Server member/membership identity synchronization.

**Review findings**
- ✅ Persisted mention IDs remain the historical identity boundary, so handle reuse cannot relabel old messages; duplicate display names retain a visible canonical discriminator.
- ✅ Plain and rich timeline copy restore canonical handles at the document boundary without changing unrelated whitespace semantics.
- ✅ Human display-name writes converge through the authoritative user/member path and fan out transactionally to membership mirrors; generic Agent PATCH cannot bypass it.
- ✅ Membership restore and startup repair use compatible lock ordering and short repair transactions, addressing the reviewed stale-write and deadlock cases.
- ✅ Membership-backed humans no longer expose display-name editing or silent profile/member rerouting in Agent Detail; non-human agent naming and other human projection fields remain intact.

**Action taken**
- Approved exact head `241738a648f724bdd4fb588d48ec6d8d5713bbcf` after static cumulative review. I did not rerun tests or visual QA; repository CI and the author-provided validation remain the execution evidence.

PR #1926, opened 2026-07-21
fix(cli): retire legacy github-scan launchd runner

### LINKED ISSUE #995: CLI upgrade strands legacy github-scan launchd runner: KeepAlive crash-loop + port squatting

Gandy2025:
## Summary

- Automatically retires stranded legacy `github-scan` launchd runners from the effective account's canonical legacy namespace on eligible prod and staging CLI actions.
- Boots out only an exact `com.first-tree.github-scan.runner.<user>.default` service, confirms launchd eviction, and only then removes the matching regular plist.
- Preserves the existing managed-update handoff by skipping only the hidden `daemon refresh-unit` child; the supervisor-restarted new binary performs retirement from `daemon start`.
- Adds exhaustive deterministic coverage plus a reusable formal QA case for the release-shaped macOS boundary.

Fixes #995.

## Root cause

Older published builds installed a KeepAlive launchd runner in an unchannelled `~/.first-tree/github-scan/runner/launchd` namespace. Removing the command did not unload that service or delete its plist, so launchd could keep restarting the missing runner and retain `127.0.0.1:7878` across CLI upgrades.

## Safety and compatibility

- Runs only on macOS for published prod and staging channels; source/dev and other platforms are no-ops.
- Derives the legacy root from `userInfo().homedir`, never `HOME` or `FIRST_TREE_HOME`, and accepts no caller-supplied root in production.
- Rejects symlinked ancestors and non-regular entries, uses `O_NOFOLLOW`, revalidates ancestor/file identity before mutation, and bounds directory, plist, diagnostic, and candidate work.
- Requires exact filename, historical label grammar, `.default` profile, and embedded plist `Label` agreement.
- Uses exact `launchctl bootout gui/<uid>/<label>`, then polls exact `launchctl print` disappearance before unlinking. Timeouts, permissions, malformed state, and partial cleanup stay diagnosable and retryable without blocking the selected CLI action.
- Persists a `0600` atomic legacy-namespace retry marker with fair cursor progression; inventory overflow fails closed without mutation.
- Current prod/staging/dev daemon labels cannot match the legacy grammar and are table-tested as untargetable.
- Leaves `update-glue.ts`, exit 75, and the old-X 45-second refresh-child contract unchanged. A regression test also locks healthy status-zero refresh output as silent.

## Trigger behavior

The synchronous Commander `preAction` hook runs after JSON/logger normalization for ordinary actions, `daemon ensure-service`, and the supervisor-restarted `daemon start`. Root/nested/positional help, version, no-action parsing, and the structurally exact hidden `daemon refresh-unit` child do not run retirement.

## Validation

- `pnpm check` — passed (only pre-existing unrelated repository warnings)
- `pnpm typecheck` — passed
- `pnpm --filter first-tree-dev typecheck` — passed
- CLI full test runner, all 21 batches — passed in a clean CI-shaped environment
- Focused retirement/runtime/update tests — 91 passed
- Related daemon/update/channel/service regression set — 188 passed
- Monorepo build — passed
- Full non-web test graph — passed, including client 1,621 tests and server 2,563 tests
- Local full graph additionally ran 1,564 web tests; one unchanged focus-preservation test failed reproducibly while the other 1,563 passed. This branch has zero `packages/web` diff and the isolated failure is recorded as an unrelated local baseline gate.
- Biome on every touched TypeScript file and `git diff --check` — passed

Formal QA is warranted because this changes CLI boot/update behavior. The reusable case is `packages/qa/cases/runtime/legacy-github-scan-launchd-retirement.md`; Docker-backed temporary-worktree readiness and the explicit native macOS staging bridge will be completed before competition submission. No live launchd or port-release PASS is claimed yet.

## Rollout note

PR #1892 was rechecked immediately before opening this PR and remains open, approved/green, but 119 commits behind and 7 commits ahead of current main. This implementation was independently derived from the binding acceptance and current main; no commits or implementation text were reused.

### DIFF
--- a/apps/cli/src/__tests__/cli-runtime.test.ts
+++ b/apps/cli/src/__tests__/cli-runtime.test.ts
@@ -1,0 +1,272 @@
+import { Command } from "commander";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const coreMocks = vi.hoisted(() => ({
+  getClientServiceStatus: vi.fn(),
+  getClientSwitchStartupBlock: vi.fn(() => ({ phase: "qa" })),
+  installClientService: vi.fn(),
+  isServiceSupported: vi.fn(() => false),
+  isServiceUnitDriftDetected: vi.fn(() => false),
+  loadCredentials: vi.fn(),
+  loadDaemonEnv: vi.fn(() => []),
+  refreshClientServiceUnitForUpdate: vi.fn(),
+  restartClientService: vi.fn(),
+}));
+
+const outputMocks = vi.hoisted(() => ({
+  line: vi.fn(),
+  setJsonMode: vi.fn(),
+}));
+
+const statusMocks = vi.hoisted(() => ({
+  renderAgentsBlock: vi.fn(),
+  renderAuthBlock: vi.fn(),
+  renderCliVersionBlock: vi.fn(),
+  renderHubBlock: vi.fn(),
+  renderServiceBlock: vi.fn(),
+}));
+
+vi.mock("../core/index.js", () => coreMocks);
+vi.mock("../core/output.js", () => ({
+  print: { line: outputMocks.line },
+  setJsonMode: outputMocks.setJsonMode,
+}));
+vi.mock("../commands/_shared/status-blocks.js", () => statusMocks);
+
+import { type CliRuntimeDependencies, configureCliRuntime } from "../cli/runtime.js";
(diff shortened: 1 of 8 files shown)


Gandy2025:
Superseded by #1928. GitHub automatically closed this draft when its remote branch was renamed to satisfy the repository's required fix/ prefix. #1928 carries the identical head commit 164f7e99; no code, history, or base changed.

PR #1929, opened 2026-07-21
feat(web): shorten trusted GitLab entity links

Gandy2025:
## Summary

- add one browser-safe GitLab entity-path parser shared by the server and web client, with canonical and legacy route support
- shorten bare Issue and Merge Request URLs from the Team's connected GitLab instance to `namespace/project#iid` and `namespace/project!iid`
- preserve the original href, full-URL title, custom Markdown labels, and existing neutral link styling
- reuse the GitLab connection query key so Settings mutations refresh the chat presentation cache

## Testing

- `pnpm check`
- `pnpm typecheck`
- `pnpm --dir packages/shared exec vitest run src/__tests__/gitlab-entity-path.test.ts`
- `pnpm --dir packages/server exec vitest run src/__tests__/gitlab-entities-agent-route.test.ts`
- `pnpm --dir packages/web exec vitest run src/pages/settings/__tests__/gitlab-page-dom.test.tsx src/lib/__tests__/gitlab-entity-link.test.ts src/pages/workspace/center/__tests__/chat-view-dom.test.tsx`
- local Playwright QA against a real server, Postgres database, GitLab connection, chat, and message at 1440px and 390px widths; no layout overflow or feature console errors
- `pnpm test` full-monorepo attempt: the concurrent run hit unrelated timeout flakes in existing skill-evals and client tests; every observed timeout passed when rerun in isolation

## Scope

No new CSS, card UI, icons, API endpoints, database changes, webhook changes, or stored-message rewrites.

### DIFF
--- a/packages/server/src/services/gitlab-entity-follow.ts
+++ b/packages/server/src/services/gitlab-entity-follow.ts
@@ -56,6 +56,9 @@ export function parseGitlabEntityUrl(
     if (parsedPath.reason === "control_character") {
       throw new BadRequestError("GitLab entity URL project path must not contain control characters");
     }
+    if (parsedPath.reason === "bidi_control") {
+      throw new BadRequestError("GitLab entity URL project path must not contain bidirectional control characters");
+    }
     if (parsedPath.reason === "route") {
       throw new BadRequestError("GitLab entity URL must point to an issue or merge request");
     }
--- a/packages/server/src/services/gitlab-entity-follow.ts
+++ b/packages/server/src/services/gitlab-entity-follow.ts
@@ -5,7 +5,7 @@ import type {
   ScmEntityState,
   UnfollowChatGitlabEntityResponse,
 } from "@first-tree/shared";
-import { chatMetadataSchema, normalizeScmEntityState } from "@first-tree/shared";
+import { chatMetadataSchema, normalizeScmEntityState, parseGitlabEntityPath } from "@first-tree/shared";
 import { and, asc, eq, inArray, isNull, or } from "drizzle-orm";
 import type { Database } from "../db/connection.js";
 import { chats } from "../db/schema/chats.js";
@@ -50,24 +50,19 @@ export function parseGitlabEntityUrl(
   }
   // GitLab instances and clients emit both route shapes. Treat them as the
   // same identity, but preserve the submitted shape for the user-facing link.
-  const match = /^\/(.+?)\/(?:-\/)?(issues|merge_requests)\/(\d+)\/?$/.exec(url.pathname);
-  if (!match) throw new BadRequestError("GitLab entity URL must point to an issue or merge request");
-  let projectPath: string;
-  try {
-    projectPath = decodeURIComponent(match[1] ?? "");
-  } catch {
-    throw new BadRequestError("Invalid GitLab entity URL encoding");
-  }
-  if (/\p{Cc}/u.test(projectPath)) {
-    throw new BadRequestError("GitLab entity URL project path must not contain control characters");
-  }
-  const entityIid = Number(match[3]);
-  if (!projectPath || !Number.isSafeInteger(entityIid) || entityIid <= 0)
+  const parsedPath = parseGitlabEntityPath(url.pathname);
+  if (!parsedPath.ok) {
+    if (parsedPath.reason === "encoding") throw new BadRequestError("Invalid GitLab entity URL encoding");
+    if (parsedPath.reason === "control_character") {
+      throw new BadRequestError("GitLab entity URL project path must not contain control characters");
+    }
+    if (parsedPath.reason === "route") {
+      throw new BadRequestError("GitLab entity URL must point to an issue or merge request");
+    }
     throw new BadRequestError("Invalid GitLab entity URL");
+  }
   return {
-    entityType: match[2] === "issues" ? "issue" : "pull_request",
-    entityIid,
-    projectPath,
+    ...parsedPath.value,
     entityUrl: url.toString().replace(/\/$/, ""),
   };
 }
--- a/packages/shared/src/__tests__/gitlab-entity-path.test.ts
+++ b/packages/shared/src/__tests__/gitlab-entity-path.test.ts
@@ -33,4 +33,16 @@ describe("parseGitlabEntityPath", () => {
   ] as const)("rejects %s as %s", (pathname, reason) => {
     expect(parseGitlabEntityPath(pathname)).toEqual({ ok: false, reason });
   });
+
+  it.each([
+    "%D8%9C",
+    "%E2%80%8F",
+    "%E2%80%AE",
+    "%E2%81%A7",
+  ])("rejects percent-encoded Unicode bidi control %s", (control) => {
+    expect(parseGitlabEntityPath(`/group/${control}project/-/issues/1`)).toEqual({
+      ok: false,
+      reason: "bidi_control",
+    });
+  });
 });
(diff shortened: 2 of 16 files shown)


yuezengwu (changes requested):
Requesting changes for four correctness/security gaps in the trusted-link boundary:

1. **[P2] Reject Unicode bidi controls before building the compact label** — `packages/shared/src/gitlab-entity-path.ts:48` rejects only `Cc`. A chat author can percent-encode RLO/PDF (or other bidi controls) in the project path; `decodeURIComponent` returns them and the Web then renders them inside `${projectPath}${sigil}${iid}`. The visible compact label can therefore read as a different project path, or reorder adjacent text, while the href points elsewhere. This is especially unsafe on touch surfaces where the full-URL `title` is not exposed. Reject bidi controls and add a parser regression case.

2. **[P2] Bind the trusted origin to the chat's Team, not the currently selected shell Team** — `packages/web/src/pages/workspace/center/chat-view.tsx:1547-1554` keys and fetches the connection from `useAuth().organizationId`. `ChatByIdView` deliberately renders a globally addressed cross-org chat before its asynchronous Team switch completes, and also keeps rendering an old-org chat while a user-initiated switch is in flight. During that interval, Team B's message links are evaluated against Team A's trusted origin and can have the wrong domain hidden. Use `chatDetail.organizationId` with an explicit org-scoped read, or fail closed (no compaction) until the chat and selected scopes match. Please cover the cross-org rendering case.

3. **[P2] Do not trust a cached connection origin forever** — `packages/web/src/pages/workspace/center/chat-view.tsx:1552` sets `staleTime: Infinity`, but connection changes made by another tab or teammate have no WebSocket/cache invalidation path. After a connection is replaced or deleted, an already-open workspace will continue compacting the old origin and will not compact the new origin until reload/cache eviction. Add bounded refresh or a cross-client invalidation mechanism so the presentation eventually follows the current connection.

4. **[P2] Apply the presentation to the blocking request body too** — `packages/web/src/components/chat/ask-takeover.tsx:633` renders the request body through plain `Markdown`, outside the new anchor override. An open `format="request"` message therefore shows a full GitLab URL in the blocking takeover (including the mobile ask sheet), even though the same message is compacted in the timeline after resolution. The takeover is the primary/mandatory view for the recipient, so it needs the same trusted-link presentation and a request-body regression test.

The shared parser refactor otherwise preserves the server's canonical/legacy route behavior, and I found no database or core persisted-data changes.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1929

- Rationale: The compact-label feature currently has four gaps where the visible link can diverge from the current chat/connection trust boundary or from the primary request surface.

**Risk level**: B-high
- Path baseline: `packages/server/**` service changes -> B-low.
- Semantic lift: the Web change adds organization-scoped query/cache behavior and therefore touches the data layer -> B-high.

**PR summary**
- Author / repo: Gandy2025 / first-tree
- Problem: Long bare GitLab Issue and Merge Request URLs are hard to scan in chat, especially on mobile.
- Approach: Compact links from the Team's connected instance to `namespace/project#iid` / `namespace/project!iid`, while preserving the original href and sharing canonical/legacy path parsing across server and Web.
- Impacted modules: `packages/shared`, `packages/server`, and the Web chat/settings surfaces.

**Review findings**
❌ 1. Reject Unicode bidi controls before using the decoded project path as visible text. `\p{Cc}` does not cover bidi formatting controls such as RLO/PDF, so a percent-encoded path can reorder or disguise the compact label while the href points elsewhere; this is particularly problematic on touch surfaces without hover title. Add a regression case. [`packages/shared/src/gitlab-entity-path.ts:48`]

❌ 2. Bind the trusted origin to the rendered chat's Team, or fail closed until chat and shell scope match. `ChatByIdView` intentionally renders a globally addressed cross-org chat before its asynchronous Team switch settles, while `ChatView` currently fetches by `useAuth().organizationId`; during that interval the wrong Team's origin can be hidden. Cover the cross-org render/switch path. [`packages/web/src/pages/workspace/center/chat-view.tsx:1547`]

❌ 3. Do not keep connection trust state fresh forever via `staleTime: Infinity`. Another tab or teammate can replace/delete the connection without any cache invalidation reaching this client, leaving the old origin compacted indefinitely. Use bounded refresh or a cross-client invalidation mechanism. [`packages/web/src/pages/workspace/center/chat-view.tsx:1552`]

❌ 4. Apply the same presentation to the blocking request body. `AskTakeover` renders the request with plain `Markdown`, so the mandatory desktop/mobile ask surface still shows the full URL while the timeline later compacts the same message. Reuse the presentation seam and add a request-body regression test. [`packages/web/src/components/chat/ask-takeover.tsx:633`]

The shared parser otherwise preserves the server's canonical/legacy route behavior, and I found no persisted-data or API-contract changes. These findings independently confirm the existing blocking review; no additional findings were identified. I performed static review only and did not run checks or tests.

**Action taken**
- Submitted request changes.

Gandy2025:
@yuezengwu @baixiaohang Thank you — both reviews identified the same four trust-boundary gaps. All four are addressed in `863fc04b`:

1. **Unicode bidi controls:** the shared parser now rejects the complete Unicode `Bidi_Control` set before building a visible label, with percent-encoded ALM/RLM/RLO/RLI regression cases. The server maps this to an explicit validation error.
2. **Rendered-chat Team scope:** the connection read now uses the `organizationId` from the rendered chat detail through an explicit Team-scoped endpoint. It fails closed until that detail is available, so a transient shell Team cannot authorize compaction.
3. **Bounded trust refresh:** the shared presentation hook refreshes the Team connection every 30 seconds. Replacement/deletion was also exercised in a live browser: the previously trusted origin was decompacted after the bounded refresh without a manual cache invalidation.
4. **Blocking request surfaces:** `AskTakeover` now accepts the shared Markdown presentation. Both the Workspace takeover and the standalone Mobile Now ask sheet use the same Team-scoped hook; bare URLs compact while custom link labels and the original href/title remain unchanged.

Validation:

- Shared parser: 17/17 tests
- Server GitLab entity route: 3/3 tests
- Focused Web suites: 95/95 tests, including cross-Team, Workspace takeover, and Mobile Now ask-sheet regressions
- Monorepo Biome check and typecheck; Web design-token guardrails
- Real-browser QA at desktop and 390 px mobile, including Mobile Now (`scrollWidth === viewport width`, 0 console errors)
- Two independent successor-diff reviews: no remaining actionable findings

Please re-review when convenient.

yuezengwu (approved):
Re-reviewed successor head `863fc04b` against the previously reviewed `46a8eda9` and the complete PR diff. All four prior blockers are resolved:

- decoded project paths now reject the complete Unicode `Bidi_Control` set, with explicit server validation and regression coverage;
- link trust is scoped to `chatDetail.organizationId` through the explicit Class-B Team endpoint and fails closed before chat detail is available;
- connection trust refreshes every 30 seconds, while same-tab Settings mutations continue to share/invalidate the exact org-scoped cache key; and
- Workspace and Mobile Now blocking request bodies now use the same Team-scoped Markdown presentation as the timeline.

I found no new actionable issues and no database or core persisted-data changes. Static successor-diff/full-diff review only; I did not rerun tests or QA.

baixiaohang (approved):
Re-reviewed successor head `863fc04b` against the previous head and the complete PR diff. All four prior trust-boundary blockers are resolved: bidi controls are rejected, trust is scoped to the rendered chat's Team, connection state refreshes within a bounded interval, and Workspace/Mobile blocking requests share the same safe link presentation.

No new actionable issues found. Static review only; I did not rerun tests or QA.

PR #775, opened 2026-06-02
chore: remove GitHub Scan subsystem

bestony:
## Summary
- Remove the `@first-tree/github-scan` package, bundled GitHub Scan skills, packaged assets, and package tests.
- Remove the `first-tree github scan` CLI namespace, build hooks, workspace references, coverage entries, and scan-specific command tests.
- Update docs, onboarding guidance, repo-local skills, and smoke assertions so GitHub Scan is no longer advertised while preserving the remaining GitHub platform surfaces.

## Verification
- `pnpm check`
- `pnpm typecheck`
- `pnpm exec turbo run test --concurrency=1`
- `git diff --check`
- `rg -n "@first-tree/github-scan|packages/github-scan|first-tree-github-scan|github-scan|GitHub Scan|github scan"` (only the historical migration note remains)
- `rg -n "commands/github|registerGithubCommands|githubCommand|runGitHubScan|GITHUB_SCAN|github-scan-statusline|copy-github-scan-assets"`
- `pnpm --filter first-tree-dev exec tsx src/cli/index.ts --help`
- `pnpm --filter first-tree-dev exec tsx src/cli/index.ts github scan` (fails with `unknown command 'github'`)

PR #1930, opened 2026-07-22
fix(cli): retire stranded legacy github-scan launchd runner on first run after upgrade

### LINKED ISSUE #995: CLI upgrade strands legacy github-scan launchd runner: KeepAlive crash-loop + port squatting

yuezengwu:
Closes #995.

## Summary

Upgrading the CLI stranded the pre-#775 github-scan launchd runner (`com.first-tree.github-scan.runner.<login>.default`): its plist pinned `ProgramArguments` to a dead absolute path with `KeepAlive: true`, so launchd crash-looped it forever and a zombie kept squatting `127.0.0.1:7878`.

On the first run of a new version, the CLI now detects the stranded runner and retires it automatically — `launchctl bootout` the service and remove its plist directory — ending the crash-loop and releasing port 7878. No manual `launchctl` surgery, idempotent, covered by tests. No general service-management framework.

## How it works

- **Hook**: global Commander `preAction` in `apps/cli/src/cli/index.ts`, so every "first run of a new version" path is covered — manual `upgrade`, auto-update restart (`exit 75`), `daemon refresh-unit`, and non-daemon usage. Steady-state cost is one `existsSync` per command; never throws.
- **Detection** (`apps/cli/src/core/legacy-github-scan.ts`), two sources because the label embeds the operator's GitHub login, which cannot be recomputed locally:
  1. plist files under `~/.first-tree/github-scan/runner/launchd/` (the only bootstrap location the legacy installer ever used; filename == label; prod-era fixed path, not channel home);
  2. a `launchctl list` sweep for the `com.first-tree.github-scan.runner.` prefix — the only way to find runners installed with `GITHUB_SCAN_DIR`/`GITHUB_SCAN_HOME` overrides, and runners whose plist was deleted by hand while the service stayed loaded.
- **Remediation**: `launchctl bootout gui/<uid>/<label>` per found label (not-loaded tolerated), then remove the tool-private plist dir. Only exact-prefix labels are booted out. Darwin-only (the legacy subsystem never existed elsewhere); everything else under `~/.first-tree/github-scan/` (e.g. a user-edited `config.yaml`) is left alone.
- **Idempotency**: a marker at `<channel home>/state/legacy-github-scan-runner-retired.json` short-circuits subsequent runs; the marker is written only on a fully clean pass, so any failure (sweep error, unexpected bootout error) is retried on the next CLI run.
- **Why auto-removal is safe here** (vs. the `dev.first-tree.client` precedent documented in `supervisor/launchd.ts`): the github-scan subsystem was removed from *every* channel in #775, so no live or parallel install can legitimately own these labels.

## Testing

- 11 new unit tests (`apps/cli/src/__tests__/legacy-github-scan.test.ts`): platform gate, marker short-circuit, clean machine, incident state (loaded + plist), manual-bootout residue, sweep-only detection (env-override install), sweep failure → no marker → retry re-runs detection, unexpected bootout error → dir still removed → recovers on retry, consecutive-run no-op, non-matching files never booted out (dir still removed wholesale), default path resolution. Plus a preAction wiring assertion in `cli-entry-and-tree-actions-extra.test.ts`.
- `pnpm typecheck` ✅, `pnpm check` (biome) ✅, `pnpm build` ✅ for `apps/cli`.
- Real-macOS smoke of the built CLI with `HOME` redirected: clean machine writes only the marker; planted residue gets booted out (not-loaded tolerated), plist dir removed, human hint on stderr, `--json` stdout/stderr stays clean.
- Repo-wide `pnpm test`: everything I touched is green. Pre-existing, environment-dependent failures on this machine reproduce identically on pristine `origin/main` (verified via `git stash`): 7 tests in `apps/cli/src/__tests__/client-switch-transaction-extra.test.ts` (they probe this machine's live daemon state); server/web/skill-evals pass standalone and only flake under turbo parallel load.

## Docs

- `docs/migration/from-first-tree-v0.md`: the `github scan` retirement section now notes the automatic launchd cleanup.

### DIFF
--- a/apps/cli/src/__tests__/cli-entry-and-tree-actions-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-entry-and-tree-actions-extra.test.ts
@@ -31,6 +31,10 @@ const outputMocks = vi.hoisted(() => ({
   setJsonMode: vi.fn(),
 }));
 
+const legacyGithubScanMocks = vi.hoisted(() => ({
+  retireLegacyGithubScanRunner: vi.fn(),
+}));
+
 vi.mock("@first-tree/client", () => clientMocks);
 vi.mock("../commands/agent/index.js", () => ({ registerAgentCommands: registrationMocks.registerAgentCommands }));
 vi.mock("../commands/chat/index.js", () => ({ registerChatCommands: registrationMocks.registerChatCommands }));
@@ -47,6 +51,7 @@ vi.mock("../commands/status.js", () => ({ registerStatusCommand: registrationMoc
 vi.mock("../commands/tree/index.js", () => ({ registerTreeCommands: registrationMocks.registerTreeCommands }));
 vi.mock("../commands/upgrade.js", () => ({ registerUpgradeCommand: registrationMocks.registerUpgradeCommand }));
 vi.mock("../core/output.js", () => outputMocks);
+vi.mock("../core/legacy-github-scan.js", () => legacyGithubScanMocks);
 
 const originalCwd = process.cwd();
 const originalHome = process.env.FIRST_TREE_HOME;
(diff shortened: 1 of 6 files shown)


yuezengwu:
GoF competition: dev and reviewer are both aligned, so this PR is now labeled `fire_submitted` (kept **draft** per competition rules — no mark-ready, no merge from our side).

Summary for maintainer review @bestony: on first run of a new CLI version, a stranded pre-#775 github-scan launchd runner (`com.first-tree.github-scan.runner.*`) is detected via its plist dir + a `launchctl list` sweep, booted out, and its plist directory removed — ending the KeepAlive crash-loop and freeing port 7878. Idempotent (state-dir marker, retry-on-failure), darwin-only, never fatal to any command. Covered by 11 new unit tests + a preAction wiring assertion; verified end-to-end on macOS against a real launchd job. Closes #995.

PR #1937, opened 2026-07-22
Collapse live output on strong context transitions

Gandy2025:
## Summary

- Treat the composer-connected current-output panel as transient detail: interactions inside it keep it open, while completed user interaction elsewhere (pointer/keyboard activation, focus, wheel/touch scroll, or drop) returns vertical space to the conversation.
- Collapse after a timeline jump only when its mounted evidence target is found; keep the panel expanded if that target disappears before activation.
- Preserve focus intent across every collapse path: the destination keeps focus, keyboard jumps focus timeline evidence before unmounting, and outside interaction clears stale status-surface focus ownership.

## Product boundary

Internal reading remains stable: scrolling, selecting, or following Markdown controls inside the expanded output does not collapse it, including a text-selection drag that crosses its boundary or a live nonzero agent-count update during that gesture. Any user interaction that starts outside the output closes it, including composer controls, continued typing or attachment paste in an already-focused composer, timeline/sidebar activation, scrollbar use, and file drop. Pointer- and click-triggered collapse waits until activation completes, including controls that stop propagation, so the layout change cannot move a pressed timeline control before its action runs. Programmatic scrolling does not emit these user-intent events and therefore does not close it. Existing chat-switch, status-clear, and successful-jump transitions remain unchanged.

## Tests

- Focused composer/status and ChatView coverage: 2 files / 73 tests passed.
- Full Web suite: 193 files / 1,588 tests passed.
- `pnpm typecheck`: 10/10 tasks passed, including the Web production build and design-token guardrails.
- The new regression cases cover outside focus, Safari-style pointer activation without button focus, click-only assistive activation through a control that stops propagation, preserved timeline-control activation, already-focused composer typing/Enter and clipboard-file paste, wheel and file-drop intent, internal output scrolling and cross-boundary selection across a live 1-to-2 agent update, failed timeline navigation, keyboard focus transfer, and pointer/drop/click-only collapse followed by a status clear without focus theft.

## QA gap

The current session has no available Browser instance, so real-browser visual/interaction QA could not be captured. The change has no visual styling delta; deterministic DOM coverage verifies the focus and collapse transitions.

### DIFF
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -93,15 +93,12 @@ export function ComposeStatusBar({
   chatId,
   agents,
   fallbackFocusRef,
-  composerSurfaceRef,
 }: {
   chatId: string;
   /** Non-human agent participants, used for stable display-name lookup. */
   agents: ChatParticipantDetail[];
   /** Stable composer control that receives focus if the status surface vanishes. */
   fallbackFocusRef?: RefObject<HTMLElement | null>;
-  /** Editable composer surface whose use signals a switch from monitoring to composing. */
-  composerSurfaceRef?: RefObject<HTMLElement | null>;
 }) {
   const [expanded, setExpanded] = useState(false);
   const [lead, setLead] = useState<{ agentId: string; since: number } | null>(null);
@@ -144,28 +141,21 @@ export function ComposeStatusBar({
   }, [chatId]);
 
   useEffect(() => {
-    if (!composerSurfaceRef) return;
-    const collapseForComposerIntent = (event: Event) => {
-      const target = event.target;
-      if (target instanceof Node && composerSurfaceRef.current?.contains(target)) setExpanded(false);
-    };
-    document.addEventListener("focusin", collapseForComposerIntent);
-    document.addEventListener("drop", collapseForComposerIntent);
-    return () => {
-      document.removeEventListener("focusin", collapseForComposerIntent);
-      document.removeEventListener("drop", collapseForComposerIntent);
-    };
-  }, [composerSurfaceRef]);
-
-  useEffect(() => {
     if (attention.length === 0) return;
-    const clearOutsidePointerFocus = (event: PointerEvent) => {
+    const collapseForOutsideIntent = (event: Event) => {
       const target = event.target;
-      if (target instanceof Node && !surfaceRef.current?.contains(target)) focusWithinRef.current = false;
+      if (!(target instanceof Node) || surfaceRef.current?.contains(target)) return;
+      focusWithinRef.current = false;
+      if (expanded) setExpanded(false);
     };
-    document.addEventListener("pointerdown", clearOutsidePointerFocus, true);
-    return () => document.removeEventListener("pointerdown", clearOutsidePointerFocus, true);
-  }, [attention.length]);
+    const eventTypes = expanded
+      ? (["pointerdown", "focusin", "click", "wheel", "touchstart", "drop"] as const)
+      : (["pointerdown"] as const);
+    for (const type of eventTypes) document.addEventListener(type, collapseForOutsideIntent, true);
+    return () => {
+      for (const type of eventTypes) document.removeEventListener(type, collapseForOutsideIntent, true);
+    };
+  }, [attention.length, expanded]);
 
   useEffect(() => {
     if (attention.length === 0) {
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -93,12 +93,15 @@ export function ComposeStatusBar({
   chatId,
   agents,
   fallbackFocusRef,
+  composerInputRef,
 }: {
   chatId: string;
   /** Non-human agent participants, used for stable display-name lookup. */
   agents: ChatParticipantDetail[];
   /** Stable composer control that receives focus if the status surface vanishes. */
   fallbackFocusRef?: RefObject<HTMLElement | null>;
+  /** Editable composer input whose focus signals that monitoring is finished. */
+  composerInputRef?: RefObject<HTMLTextAreaElement | null>;
 }) {
   const [expanded, setExpanded] = useState(false);
   const [lead, setLead] = useState<{ agentId: string; since: number } | null>(null);
@@ -139,6 +142,15 @@ export function ComposeStatusBar({
     activeChatIdRef.current = chatId;
     setExpanded(false);
   }, [chatId]);
+
+  useEffect(() => {
+    if (!composerInputRef) return;
+    const collapseForComposerFocus = (event: FocusEvent) => {
+      if (event.target === composerInputRef.current) setExpanded(false);
+    };
+    document.addEventListener("focusin", collapseForComposerFocus);
+    return () => document.removeEventListener("focusin", collapseForComposerFocus);
+  }, [composerInputRef]);
 
   useEffect(() => {
     if (attention.length === 0) return;
@@ -231,7 +243,13 @@ export function ComposeStatusBar({
               attention={attention}
               nameOf={nameOf}
               mounted={mounted}
-              onTimelineNavigate={() => setExpanded(false)}
+              onTimelineNavigate={() => {
+                // A pointer jump does not transfer focus to timeline evidence.
+                // Clear ownership before its focused button is unmounted so a
+                // later status removal cannot steal focus back to the composer.
+                focusWithinRef.current = false;
+                setExpanded(false);
+              }}
             />
           ) : null}
         </section>
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -107,6 +107,7 @@ export function ComposeStatusBar({
   const surfaceRef = useRef<HTMLElement | null>(null);
   const triggerRef = useRef<HTMLButtonElement | null>(null);
   const focusWithinRef = useRef(false);
+  const pointerOriginOutsideRef = useRef<boolean | null>(null);
   const { data: rawStatuses } = useQuery({
     queryKey: chatAgentStatusQueryKey(chatId),
     queryFn: () => fetchChatAgentStatuses(chatId),
@@ -142,18 +143,69 @@ export function ComposeStatusBar({
 
   useEffect(() => {
     if (attention.length === 0) return;
-    const collapseForOutsideIntent = (event: Event) => {
+    const isOutside = (event: Event) => {
       const target = event.target;
-      if (!(target instanceof Node) || surfaceRef.current?.contains(target)) return;
+      return target instanceof Node && !surfaceRef.current?.contains(target);
+    };
+    const collapse = () => {
       focusWithinRef.current = false;
       if (expanded) setExpanded(false);
     };
-    const eventTypes = expanded
-      ? (["pointerdown", "focusin", "click", "wheel", "touchstart", "drop"] as const)
-      : (["pointerdown"] as const);
-    for (const type of eventTypes) document.addEventListener(type, collapseForOutsideIntent, true);
+    if (!expanded) {
+      const clearOutsidePointerFocus = (event: PointerEvent) => {
+        if (isOutside(event)) focusWithinRef.current = false;
+      };
+      document.addEventListener("pointerdown", clearOutsidePointerFocus, true);
+      return () => document.removeEventListener("pointerdown", clearOutsidePointerFocus, true);
+    }
+
+    let collapseFrame: number | null = null;
+    const collapseAfterPointerActivation = () => {
+      if (collapseFrame !== null) return;
+      collapseFrame = requestAnimationFrame(() => {
+        collapseFrame = null;
+        collapse();
+      });
+    };
+    const onPointerDown = (event: PointerEvent) => {
+      pointerOriginOutsideRef.current = isOutside(event);
+      if (pointerOriginOutsideRef.current) focusWithinRef.current = false;
+    };
+    const onPointerUp = () => {
+      const startedOutside = pointerOriginOutsideRef.current;
+      pointerOriginOutsideRef.current = null;
+      if (startedOutside) collapseAfterPointerActivation();
+    };
+    const onPointerCancel = () => {
+      const startedOutside = pointerOriginOutsideRef.current;
+      pointerOriginOutsideRef.current = null;
+      if (startedOutside) collapseAfterPointerActivation();
+    };
+    const onFocusIn = (event: FocusEvent) => {
+      if (pointerOriginOutsideRef.current === null && isOutside(event)) collapse();
+    };
+    const onOutsideIntent = (event: Event) => {
+      if (isOutside(event)) collapse();
+    };
+    document.addEventListener("pointerdown", onPointerDown, true);
+    document.addEventListener("pointerup", onPointerUp, true);
+    document.addEventListener("pointercancel", onPointerCancel, true);
+    document.addEventListener("focusin", onFocusIn, true);
+    document.addEventListener("keydown", onOutsideIntent);
+    document.addEventListener("beforeinput", onOutsideIntent);
+    document.addEventListener("wheel", onOutsideIntent, true);
+    document.addEventListener("drop", onOutsideIntent, true);
     return () => {
-      for (const type of eventTypes) document.removeEventListener(type, collapseForOutsideIntent, true);
+      pointerOriginOutsideRef.current = null;
+      if (collapseFrame !== null) cancelAnimationFrame(collapseFrame);
+      document.removeEventListener("pointerdown", onPointerDown, true);
+      document.removeEventListener("pointerup", onPointerUp, true);
+      document.removeEventListener("pointercancel", onPointerCancel, true);
+      document.removeEventListener("focusin", onFocusIn, true);
+      document.removeEventListener("keydown", onOutsideIntent);
+      document.removeEventListener("beforeinput", onOutsideIntent);
+      document.removeEventListener("wheel", onOutsideIntent, true);
+      document.removeEventListener("drop", onOutsideIntent, true);
     };
   }, [attention.length, expanded]);
 
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -118,6 +118,7 @@ export function ComposeStatusBar({
   // must never resurrect Working after runtime freshness has expired.
   const statuses = rawStatuses ?? [];
   const attention = useMemo(() => selectAttention(statuses), [statuses]);
+  const hasAttention = attention.length > 0;
   const mounted = useMountedAnchors();
   const nameOf = useMemo(() => nameFor(agents), [agents]);
   const announcement = useMemo(() => activityAnnouncement(attention, nameOf), [attention, nameOf]);
@@ -141,8 +142,10 @@ export function ComposeStatusBar({
     setExpanded(false);
   }, [chatId]);
 
+  // Keep the listener lifetime stable while the actionable set changes so a
+  // live 1 -> 2 agent update cannot erase an in-flight pointer origin.
   useEffect(() => {
-    if (attention.length === 0) return;
+    if (!hasAttention) return;
     const isOutside = (event: Event) => {
       const target = event.target;
       return target instanceof Node && !surfaceRef.current?.contains(target);
@@ -189,21 +192,32 @@ export function ComposeStatusBar({
     const onOutsideIntent = (event: Event) => {
       if (isOutside(event)) collapse();
     };
-    const onClick = (event: MouseEvent) => {
+    const onClickCapture = (event: MouseEvent) => {
       const pointerOriginOutside = pointerOriginOutsideRef.current;
-      pointerOriginOutsideRef.current = null;
       if (pointerCompletionFrame !== null) cancelAnimationFrame(pointerCompletionFrame);
       pointerCompletionFrame = null;
-      if (pointerOriginOutside === false) return;
-      if (pointerOriginOutside === true || isOutside(event)) collapse();
+      if (pointerOriginOutside === false) {
+        pointerOriginOutsideRef.current = null;
+        return;
+      }
+      if (pointerOriginOutside !== true && !isOutside(event)) return;
+      focusWithinRef.current = false;
+      // Capture sees Workspace controls that stop click propagation; defer the
+      // layout change until their target action has completed.
+      pointerCompletionFrame = requestAnimationFrame(() => {
+        pointerCompletionFrame = null;
+        pointerOriginOutsideRef.current = null;
+        collapse();
+      });
     };
     document.addEventListener("pointerdown", onPointerDown, true);
     document.addEventListener("pointerup", onPointerUp, true);
     document.addEventListener("pointercancel", onPointerCancel, true);
     document.addEventListener("focusin", onFocusIn, true);
-    document.addEventListener("click", onClick);
+    document.addEventListener("click", onClickCapture, true);
     document.addEventListener("keydown", onOutsideIntent);
     document.addEventListener("beforeinput", onOutsideIntent);
+    document.addEventListener("paste", onOutsideIntent, true);
     document.addEventListener("wheel", onOutsideIntent, true);
     document.addEventListener("drop", onOutsideIntent, true);
     return () => {
@@ -213,13 +227,14 @@ export function ComposeStatusBar({
       document.removeEventListener("pointerup", onPointerUp, true);
       document.removeEventListener("pointercancel", onPointerCancel, true);
       document.removeEventListener("focusin", onFocusIn, true);
-      document.removeEventListener("click", onClick);
+      document.removeEventListener("click", onClickCapture, true);
       document.removeEventListener("keydown", onOutsideIntent);
       document.removeEventListener("beforeinput", onOutsideIntent);
+      document.removeEventListener("paste", onOutsideIntent, true);
       document.removeEventListener("wheel", onOutsideIntent, true);
       document.removeEventListener("drop", onOutsideIntent, true);
     };
-  }, [attention.length, expanded]);
+  }, [expanded, hasAttention]);
 
   useEffect(() => {
     if (attention.length === 0) {
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -226,7 +226,13 @@ export function ComposeStatusBar({
           </button>
 
           {expanded ? (
-            <CurrentOutputDetails id={detailsId} attention={attention} nameOf={nameOf} mounted={mounted} />
+            <CurrentOutputDetails
+              id={detailsId}
+              attention={attention}
+              nameOf={nameOf}
+              mounted={mounted}
+              onTimelineNavigate={() => setExpanded(false)}
+            />
           ) : null}
         </section>
       ) : null}
@@ -308,11 +314,13 @@ function CurrentOutputDetails({
   attention,
   nameOf,
   mounted,
+  onTimelineNavigate,
 }: {
   id: string;
   attention: AgentChatStatus[];
   nameOf: (agentId: string) => string;
   mounted: ReadonlySet<string>;
+  onTimelineNavigate: () => void;
 }) {
   const showIdentity = attention.length > 1;
   return (
@@ -336,6 +344,7 @@ function CurrentOutputDetails({
               name={nameOf(status.agentId)}
               mounted={mounted}
               showIdentity={showIdentity}
+              onTimelineNavigate={onTimelineNavigate}
             />
           </li>
         ))}
@@ -349,11 +358,13 @@ function CurrentOutputItem({
   name,
   mounted,
   showIdentity,
+  onTimelineNavigate,
 }: {
   status: AgentChatStatus;
   name: string;
   mounted: ReadonlySet<string>;
   showIdentity: boolean;
+  onTimelineNavigate: () => void;
 }) {
   const visual = statusVisual(status);
   const snapshot = agentSnapshot(status);
@@ -390,6 +401,7 @@ function CurrentOutputItem({
           target={jumpTarget}
           anchored
           ariaLabel={`View ${name} in the timeline`}
+          onNavigate={onTimelineNavigate}
           className="compose-status-jump"
           interactiveClassName="rounded-[var(--radius-input)] focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring"
         >
--- a/packages/web/src/components/chat/compose-status-bar.tsx
+++ b/packages/web/src/components/chat/compose-status-bar.tsx
@@ -159,27 +159,29 @@ export function ComposeStatusBar({
       return () => document.removeEventListener("pointerdown", clearOutsidePointerFocus, true);
     }
 
-    let collapseFrame: number | null = null;
-    const collapseAfterPointerActivation = () => {
-      if (collapseFrame !== null) return;
-      collapseFrame = requestAnimationFrame(() => {
-        collapseFrame = null;
-        collapse();
+    let pointerCompletionFrame: number | null = null;
+    const settlePointerAfterActivation = (startedOutside: boolean) => {
+      if (pointerCompletionFrame !== null) cancelAnimationFrame(pointerCompletionFrame);
+      pointerCompletionFrame = requestAnimationFrame(() => {
+        pointerCompletionFrame = null;
+        if (pointerOriginOutsideRef.current !== startedOutside) return;
+        pointerOriginOutsideRef.current = null;
+        if (startedOutside) collapse();
       });
     };
     const onPointerDown = (event: PointerEvent) => {
+      if (pointerCompletionFrame !== null) cancelAnimationFrame(pointerCompletionFrame);
+      pointerCompletionFrame = null;
       pointerOriginOutsideRef.current = isOutside(event);
       if (pointerOriginOutsideRef.current) focusWithinRef.current = false;
     };
     const onPointerUp = () => {
       const startedOutside = pointerOriginOutsideRef.current;
-      pointerOriginOutsideRef.current = null;
-      if (startedOutside) collapseAfterPointerActivation();
+      if (startedOutside !== null) settlePointerAfterActivation(startedOutside);
     };
     const onPointerCancel = () => {
       const startedOutside = pointerOriginOutsideRef.current;
-      pointerOriginOutsideRef.current = null;
-      if (startedOutside) collapseAfterPointerActivation();
+      if (startedOutside !== null) settlePointerAfterActivation(startedOutside);
     };
     const onFocusIn = (event: FocusEvent) => {
       if (pointerOriginOutsideRef.current === null && isOutside(event)) collapse();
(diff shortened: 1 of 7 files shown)


baixiaohang (approved):
## Review by baixiaohang-code-reviewer - PR #1937

**Recommendation**: approve
- Rationale: The panel now collapses only under the same successful-target guard that performs the timeline jump, while failed DOM races and keyboard focus handoff remain correct.

**Risk level**: C
- Semantic lift: none; this does not touch subscriptions, WebSocket parsing, or the Web data layer.

**PR summary**
- Problem: After users jump from the expanded current-output panel to an agent's timeline evidence, the redundant panel remains open; closing it earlier would lose the panel when the target disappeared and could break keyboard focus.
- Approach: Make the timeline helper report whether a mounted target was found, collapse the panel only after a true result, and complete keyboard focus transfer before React unmounts the expanded details.
- Impacted modules: composer status disclosure, timeline jump button/helper, and focused DOM/helper coverage.

**Review findings**
✅ 1. The helper's boolean result is the single guard for both navigation and collapse, so a stale affordance cannot close the panel on a no-op jump.
✅ 2. Keyboard activation focuses timeline evidence synchronously before `setExpanded(false)` removes the jump control.
✅ 3. The return-type change covers all current callers and explicitly handles unsupported states and missing anchors.

**Action taken**
- Approved.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -226,7 +226,13 @@ export function ComposeStatusBar({
           </button>
 
           {expanded ? (
-            <CurrentOutputDetails id={detailsId} attention={attention} nameOf={nameOf} mounted={mounted} />
+            <CurrentOutputDetails
+              id={detailsId}
+              attention={attention}
+              nameOf={nameOf}
+              mounted={mounted}
+              onTimelineNavigate={() => setExpanded(false)}
```
yuezengwu: **[P2] Clear focus ownership when this successful pointer jump collapses the details.** A pointer activation typically focuses the jump button during `pointerdown`, which sets `focusWithinRef.current = true`. The click then calls the helper with `focus: false` (`event.detail > 0`) and this callback unmounts that focused button. When focus falls to `body`/null, the existing blur handler deliberately does not clear the ref. If the user subsequently tabs to another control and actionable status later clears, the effect at lines 153–159 reads the stale ref and moves focus back to the composer. This new focus-steal path is introduced by collapsing the focused subtree. Please clear or transfer the surface's focus ownership on successful navigation and add a `detail > 0` regression case.

yuezengwu (changes requested):
## Review by yzw-codex — PR #1937

The implementation goal and main approach are sound: make the timeline helper return whether it found mounted evidence, collapse only after that success signal, and move keyboard focus before the expanded subtree unmounts. The success/failure guard and all production callers are otherwise correct.

I found one blocking pointer-focus regression in the new collapse callback; details are inline. There are no database or core data-structure changes. I reviewed the diff and caller graph and ran `git diff --check`; I did not run tests or QA, per review scope.

baixiaohang (changes requested):
## Review by baixiaohang-code-reviewer - PR #1937 (follow-up)

- Rationale: A successful pointer jump collapses the focused subtree without clearing its focus-ownership marker, which can later steal focus back to the composer when actionable status clears.

**Risk level**: C
- Semantic lift: none; this remains local UI/focus behavior.

**PR summary**
- Problem: The panel should collapse after a real timeline jump without losing the user's navigation context or breaking focus behavior.
- Approach: The success/failure navigation guard is sound, but the new collapse path handles keyboard focus only and leaves pointer focus ownership stale.
- Impacted modules: composer status disclosure and its timeline-jump interaction coverage.

**Review findings**
❌ 1. A pointer activation focuses the jump button, calls the helper with `focus: false`, then unmounts that button. The blur-to-body path intentionally preserves `focusWithinRef.current`, and later keyboard navigation outside the surface does not clear it; when actionable status clears, lines 153–159 can therefore move focus unexpectedly back to the composer. Clear or transfer focus ownership on successful pointer navigation and add a `detail > 0` regression that proves a later status clear does not steal focus. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:234`]
✅ 2. The success-only collapse guard, failed-target persistence, helper return contract, and keyboard focus-before-unmount ordering are otherwise correct.

**Action taken**
- Submitted request changes, superseding my earlier approval after the pointer-only path was identified.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 2e77124a1. Successful navigation now clears the status surface focus-ownership marker before the expanded controls unmount. Added a pointer activation regression with MouseEvent detail: 1 that moves focus to a later control, clears actionable status, and verifies the composer fallback does not steal focus. The focused status and ChatView suite passes 71 tests, the full Web suite passes 193 files / 1,586 tests, and repository typecheck passes 10/10 tasks.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The current head fixes the pointer focus-ownership regression and consistently collapses the live-output panel only for explicit composer input intent or successful timeline navigation, without changing status authority.

**Risk level**: C
- Semantic lift: none; this remains local Web interaction/focus behavior and does not touch subscriptions, WebSocket parsing, or the data layer.

**PR 摘要**
- Problem: Users switching from monitoring current agent output to composing or reading timeline evidence should regain conversation space without failed jumps closing useful output or pointer jumps causing later focus theft.
- Approach: Collapse for focus/drop intent contained within the editable composer card and for successful timeline navigation only; preserve the user's focused composer control or keyboard-selected evidence, clear stale pointer focus ownership before unmount, and leave read-only mode unchanged.
- Impacted modules: composer current-output surface, timeline jump helper/button, Workspace composer wiring, and focused DOM coverage.

**Review findings**
✅ 1. The prior blocker is fixed before unmount: successful navigation clears `focusWithinRef`, and the `detail > 0` regression follows through to a later status clear while preserving external focus.
✅ 2. Composer collapse is scoped by exact containment in the editable composer card, covering text, mention/command and attachment controls plus file drop without treating unrelated page interaction as compose intent; read-only mode supplies no surface.
✅ 3. Successful keyboard jumps still focus timeline evidence before unmount, failed jumps retain the panel, and the helper's boolean return remains the single success guard.
✅ 4. The change does not alter per-chat status projection, timeline evidence selection, or Web cache/data contracts.

**Action taken**
- Approved current head `2e77124a1` after re-reviewing both successor commits; no tests or browser QA were run in this review.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The successor commit is formatting-only and leaves the already-reviewed focus, composer-intent, and successful-navigation behavior unchanged.

**Risk level**: C
- Semantic lift: none; the latest delta only reformats one test component invocation.

**PR 摘要**
- Problem: Users switching from monitoring live agent output to composing or timeline evidence need the panel to release space without failed navigation or focus theft.
- Approach: Collapse for intent contained in the editable composer and for successful timeline navigation, while preserving keyboard/pointer focus semantics and read-only behavior.
- Impacted modules: Web composer current-output surface, timeline jump helpers, Workspace composer wiring, and DOM coverage.

**Review findings**
✅ 1. The latest `10285832a` delta only folds one `ComposeStatusBar` test invocation onto one line; props, assertions, and execution paths are identical.
✅ 2. The previously reviewed pointer-focus fix, composer containment guard, failed-jump persistence, and keyboard evidence focus remain unchanged.

**Action taken**
- Approved current head `10285832a`; no tests or browser QA were run in this review.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -140,6 +143,20 @@ export function ComposeStatusBar({
     setExpanded(false);
   }, [chatId]);
 
+  useEffect(() => {
+    if (!composerSurfaceRef) return;
+    const collapseForComposerIntent = (event: Event) => {
+      const target = event.target;
+      if (target instanceof Node && composerSurfaceRef.current?.contains(target)) setExpanded(false);
```
yuezengwu: **[P2] Clear status focus ownership before a drop-triggered collapse.** An OS file drag does not move DOM focus or guarantee an in-page `pointerdown`. If a details/jump control currently owns focus, this `drop` path calls only `setExpanded(false)`, unmounting that control; the blur-to-body path preserves `focusWithinRef.current = true`. After the user keyboard-focuses another control, a later `attention.length === 0` update can therefore steal focus back to `fallbackFocusRef`. Clear or transfer ownership before this collapse (doing so for both composer-intent event types is safe), and cover focused jump → composer drop → external focus → status clear. The current drop assertion starts with focus outside the status surface, so it cannot catch this regression.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -140,6 +143,20 @@ export function ComposeStatusBar({
...
+    };
+    document.addEventListener("focusin", collapseForComposerIntent);
```
yuezengwu: **[P2] `focusin` is not a portable activation signal for the composer buttons this contract now includes.** Safari deliberately does not focus a button on mouse click ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#clicking_and_focus)), so Attach/Send/remove actions can run without this listener firing; `drop` does not cover those clicks. This is especially reachable when the textarea retained focus while the user reopened the panel with a Safari pointer click. Listen for contained pointer/click intent (or wire the relevant controls directly) in addition to keyboard focus, and exercise actual button activation. The new test's explicit `attachButtonRef.current?.focus()` only proves the keyboard/focus path.

yuezengwu (changes requested):
## Follow-up review by yzw-codex — PR #1937

The latest head broadens the behavior from success-only timeline jumps to strong composer intent: editable composer focus/control use and file drop. The original pointer-jump focus blocker is fixed correctly, and failed navigation plus keyboard evidence focus remain sound.

Two blocking gaps remain in the new composer-intent path: drop can recreate stale focus ownership, and `focusin` alone does not observe pointer activation of composer buttons across supported browser behavior. Details are inline. There are no database or core data-structure changes. I reviewed all successor commits through `10285832a`, the Workspace wiring, event/focus ordering, and `git diff --check`; I did not run tests or browser QA, per review scope.

baixiaohang (changes requested):
- Rationale: The broadened composer-intent path can still leave stale status focus ownership on drop and misses pointer activation of composer buttons in browsers that do not focus buttons on click.

**Risk level**: C
- Semantic lift: none; the blockers are local Web event/focus behavior.

**PR 摘要**
- Problem: The live-output panel should release space whenever the user starts composing or successfully navigates to evidence, without later stealing focus or missing supported composer controls.
- Approach: The timeline navigation path is now sound, but the new composer-level `focusin`/`drop` listener does not yet make focus cleanup and pointer intent atomic across all promised inputs.
- Impacted modules: composer current-output focus ownership, Workspace composer event wiring, and DOM regressions.

**Review findings**
❌ 1. A file drop can occur while a details/jump control still owns focus, without any preceding page `pointerdown`. The drop handler only calls `setExpanded(false)`, so unmounting that control can preserve `focusWithinRef.current = true` and later steal focus back to the composer when status clears. Clear or transfer ownership in the composer-intent guard and cover focused jump → composer drop → external focus → status clear. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:150`]
❌ 2. `focusin` is not a complete activation signal for the composer buttons included in the product boundary: Safari pointer clicks do not necessarily focus buttons, so attachment/send/remove activation can leave the panel open. Observe contained pointer/click intent (or wire the actions directly) and test actual button activation without an explicit `.focus()`. [R5 / `packages/web/src/components/chat/compose-status-bar.tsx:152`]
✅ 3. The original pointer-jump focus blocker is fixed, and successful/failed navigation plus keyboard evidence focus remain correct.

**Action taken**
- Submitted request changes, superseding my approval of `10285832a`; no tests or browser QA were run in this review.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 7236042f4. The unified outside-intent handler now clears status-surface focus ownership before every outside-triggered collapse, including an OS drop with no preceding pointer event. Added the requested focused jump -> outside drop -> later external focus -> status clear regression; the composer fallback does not steal focus.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 7236042f4. Outside pointerdown and click are observed in the capture phase, so attachment/send/remove activation collapses the panel independently of whether Safari focuses the button. The regression dispatches pointerdown on the attachment control without calling focus. Internal output interaction remains contained and does not collapse.

baixiaohang (changes requested):
- Rationale: The previous composer/drop focus blockers are fixed, but the new global pointerdown collapse can relayout the timeline before the outside pointer activation completes and lose the intended action.

**Risk level**: C
- Semantic lift: none; this remains local Web interaction and focus behavior.

**PR summary**
- Problem: Users leaving the transient live-output detail should recover conversation space while internal reading remains stable, failed timeline jumps retain the panel, and every collapse path preserves focus intent.
- Approach: The current head generalizes composer-specific handling into document-capture listeners for outside pointer, focus, click, wheel, touch, and drop intent; it clears status focus ownership before collapse and exempts events contained by the status surface.
- Impacted modules: composer live-output disclosure, Workspace composer wiring, and focused DOM coverage.

**Review findings**
❌ 1. `pointerdown` now synchronously schedules the layout-changing collapse for every outside target before that target completes activation. Removing details that may be 42vh tall shrinks the composer footer and grows the flex timeline viewport; when the timeline is pinned at its bottom, its scroll position is clamped to the new maximum and the pressed message/timeline control can move before `pointerup`. Pointer Events derives `click` from the down/up targets, so that interaction can be retargeted to an ancestor or lost instead of activating the control the user pressed. Preserve the outside activation before collapsing (for example, use click/pointerup for activations and isolate any raw-pointer scrollbar handling) and cover a timeline control whose action still runs while the panel closes. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:145`]
✅ 2. The prior drop focus-ownership bug is fixed atomically, and pointer activation no longer depends on Safari focusing buttons.
✅ 3. Internal output wheel/click containment, successful/failed timeline jumps, and keyboard evidence focus remain sound.

**Action taken**
- Submitted request changes on `7236042f4`; no tests, code checks, or browser QA were run in this review.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +142,20 @@ export function ComposeStatusBar({
...
-    const clearOutsidePointerFocus = (event: PointerEvent) => {
+    const collapseForOutsideIntent = (event: Event) => {
       const target = event.target;
-      if (target instanceof Node && !surfaceRef.current?.contains(target)) focusWithinRef.current = false;
+      if (!(target instanceof Node) || surfaceRef.current?.contains(target)) return;
+      focusWithinRef.current = false;
+      if (expanded) setExpanded(false);
+    };
+    const eventTypes = expanded
+      ? (["pointerdown", "focusin", "click", "wheel", "touchstart", "drop"] as const)
```
yuezengwu: [P2] Preserve the outside action before this layout-changing collapse. The capture listener closes on `pointerdown`/`touchstart`, before `pointerup`/`click`. Removing up to `min(42vh, 28rem)` shrinks the bottom composer and grows the flex timeline; at the usual bottom-pinned position, `scrollTop` clamps and timeline controls move before release, so the pressed button or link can lose or retarget its click. Conversely, classifying only the final `click` target can collapse a text selection that began inside the output and released outside, because the click may target the nearest common ancestor. Track the pointer sequence origin and collapse at a completion point that cannot move the pressed target; regress both a bottom-pinned timeline control that still activates and an internal drag-selection crossing the boundary.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +142,20 @@ export function ComposeStatusBar({
...
+      : (["pointerdown"] as const);
+    for (const type of eventTypes) document.addEventListener(type, collapseForOutsideIntent, true);
```
yuezengwu: [P2] Observe continued composer input when focus was already outside before expansion. Safari pointer-clicks need not focus buttons: if the textarea is focused and the user pointer-opens this internal disclosure, the contained pointer/click events are ignored and focus can remain in the textarea. Subsequent typing or Enter emits none of these six observed events, so the panel stays expanded during active composing. Observe keyboard/input intent outside or deliberately normalize focus when opening, and add a focused-textarea → pointer-open-without-focus-transfer → type/Enter regression.

yuezengwu (changes requested):
## Follow-up review by yzw-codex — PR #1937

The current head broadens the behavior into a document-capture outside-intent model: internal output interaction remains stable, while outside pointer, focus, click, scroll, and drop intent clears focus ownership and collapses the transient panel. This correctly fixes the previous drop-focus and Safari composer-button blockers, and the success-only timeline-jump guard remains sound.

Two blocking gaps remain in event ordering and coverage: a pointer-start collapse can relayout the timeline before the pressed control activates, and already-focused composer keyboard input can remain invisible after a Safari-style disclosure click that does not transfer focus. The pointer sequence also needs an origin-aware completion rule so an internal text-selection drag is not later classified only by its outside click target. Details are inline.

There are no database or core data-structure changes. I reviewed the full diff through `7236042f4`, the flex layout and event/focus ordering, relevant callers, and `git diff --check`; I did not run tests or browser QA, per review scope.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on ae43735b4. Expanded output now observes bubbled keydown and beforeinput intent outside the surface, so a textarea that retained focus across a Safari-style pointer disclosure collapses on continued typing or Enter. The regression keeps focus on the textarea while pointer-opening the disclosure, then dispatches Enter and verifies collapse.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on ae43735b4. Pointerdown now records only whether the sequence began outside and clears stale focus ownership without changing layout. An outside pointerup or pointercancel schedules collapse for the next animation frame, after the pressed control activation; the regression proves the external timeline action runs while the panel closes. A sequence that begins inside never schedules collapse even when pointerup and the derived click land outside, covering cross-boundary text selection.

baixiaohang (changes requested):
- Rationale: The reported pointer, selection, and continued-composer sequences are addressed, but the pointer guard is released before activation is guaranteed complete and click-only accessible activation is no longer observed.

**Risk level**: C
- Semantic lift: none; this remains local Web event, layout, and focus behavior.

**PR summary**
- Problem: Leaving transient live output should return conversation space only after the destination interaction is safe, while internal reading and selection stay open and already-focused composing still closes the detail.
- Approach: The latest head records pointer origin, delays outside pointer collapse to an animation frame after pointer completion, observes keyboard/input intent, and retains the success-only timeline-jump guard.
- Impacted modules: composer live-output disclosure and its DOM interaction coverage.

**Review findings**
❌ 1. `onPointerUp` clears `pointerOriginOutsideRef` before the scheduled activation frame. Focus event ordering relative to pointer events is intentionally not fixed across user agents; if an outside control produces `pointerup` -> `focusin` -> `click`, `onFocusIn` sees `null` and calls `collapse()` immediately, restoring the same layout change before the pressed control activates. Keep the pointer-sequence guard through click/activation completion, and regress outside pointerdown -> pointerup -> focusin -> click with the panel still mounted until the action runs. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:174`]
❌ 2. Expanded mode no longer observes `click` at all. Native link/button `onclick` is the device-independent activation surface, and assistive technology can invoke it without a corresponding DOM pointer sequence, keydown, or reliable focus transition. In that path the outside action runs but the panel remains expanded. Handle non-pointer click activation while retaining pointer origin long enough to ignore the derived outside click from an internal cross-boundary selection, and add a click-only outside-control regression. [R5 / `packages/web/src/components/chat/compose-status-bar.tsx:190`]
✅ 3. The previous early-pointer collapse, internal cross-boundary selection, and already-focused textarea typing gaps are otherwise fixed by the new origin tracking, deferred collapse, and keydown/beforeinput coverage.
✅ 4. Drop focus ownership, successful/failed timeline navigation, and keyboard evidence focus remain correct.

**Action taken**
- Submitted request changes on `ae43735b4`; no tests, code checks, or browser QA were run in this review.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +143,71 @@ export function ComposeStatusBar({
...
+        collapse();
+      });
+    };
+    const onPointerDown = (event: PointerEvent) => {
+      pointerOriginOutsideRef.current = isOutside(event);
+      if (pointerOriginOutsideRef.current) focusWithinRef.current = false;
+    };
+    const onPointerUp = () => {
+      const startedOutside = pointerOriginOutsideRef.current;
+      pointerOriginOutsideRef.current = null;
```
yuezengwu: [P2] Keep the pointer-sequence guard through activation completion. This runs in document capture and clears the marker before target handlers/default actions. Pointer Events explicitly leaves focus ordering relative to pointer events undefined across user agents; for an outside `pointerup → focusin → click` sequence, `onFocusIn` now sees `null` and calls `collapse()` immediately, so the timeline can relayout before the pressed control's click. Preserve the guard until the deferred activation frame (or route this focus transition through the same deferral), and add a regression where an intervening outside `focusin` still leaves the panel mounted when the click action executes. See the [Pointer Events ordering note](https://www.w3.org/TR/pointerevents/#compatibility-mapping-with-mouse-events).

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +143,71 @@ export function ComposeStatusBar({
...
+      if (pointerOriginOutsideRef.current === null && isOutside(event)) collapse();
+    };
+    const onOutsideIntent = (event: Event) => {
+      if (isOutside(event)) collapse();
+    };
+    document.addEventListener("pointerdown", onPointerDown, true);
+    document.addEventListener("pointerup", onPointerUp, true);
+    document.addEventListener("pointercancel", onPointerCancel, true);
+    document.addEventListener("focusin", onFocusIn, true);
+    document.addEventListener("keydown", onOutsideIntent);
```
yuezengwu, line 218: [P2] Restore a completed, device-independent activation path. Accessibility APIs, Voice Control, and screen-reader virtual cursors can invoke an outside control with a trusted `click` but no DOM pointer sequence, keydown, or reliable focus transition; none of these listeners then runs, so the action succeeds while the panel remains expanded and status focus ownership can stay stale. Handle click-only activation after the target action while ignoring pointer-derived clicks that began inside (for example, a guarded/detail-zero fallback), and regress outside click-only action → collapse → later status clear without fallback focus theft. The platform defines `click` as the device-independent activation event: [Pointer Events](https://www.w3.org/TR/pointerevents/#the-click-auxclick-and-contextmenu-events).

yuezengwu (changes requested):
## Follow-up review by yzw-codex — PR #1937

The latest head replaces the early layout mutation with pointer-origin tracking and a post-pointer animation-frame collapse, and adds `keydown`/`beforeinput` coverage for an already-focused composer. That addresses the prior early-pointer, cross-boundary selection, and retained-textarea-input blockers; drop focus ownership and success-only timeline navigation also remain correct.

Two event-model gaps still block the promised completed outside-activation behavior: the pointer guard is cleared before focus and click ordering is guaranteed complete, and device-independent click-only activation is no longer observed. Details are inline.

There are no database or core data-structure changes. I reviewed the exact `7236042f4..ae43735b4` delta, the browser event-order contracts, listener cleanup/focus ownership, and the added regressions, and ran `git diff --check`; I did not run tests or browser QA, per review scope.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 9b4ef0865 with exact focus-ownership coverage added on f0dc740d0. A document bubble click handles click-only outside activation after the target action, while pointer-origin false ignores a derived cross-boundary click that started inside the output. The new regression covers click-only outside action -> collapse -> later external focus -> status clear and verifies that the composer fallback does not steal focus. Focused coverage passes 73/73 and the full Web suite passes 1,588 tests.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +143,83 @@ export function ComposeStatusBar({
...
+      if (pointerCompletionFrame !== null) cancelAnimationFrame(pointerCompletionFrame);
+      pointerCompletionFrame = null;
+      if (pointerOriginOutside === false) return;
+      if (pointerOriginOutside === true || isOutside(event)) collapse();
+    };
+    document.addEventListener("pointerdown", onPointerDown, true);
+    document.addEventListener("pointerup", onPointerUp, true);
+    document.addEventListener("pointercancel", onPointerCancel, true);
+    document.addEventListener("focusin", onFocusIn, true);
+    document.addEventListener("click", onClick);
```
yuezengwu: [P2] Capture click-only intent before an outside control can stop propagation, then defer the collapse until its target action finishes. This bubble listener is not reached by real Workspace conversation-sidebar controls: `RowActionsMenu` calls `stopPropagation()` in both its trigger and menu-item handlers (`components/ui/row-actions-menu.tsx:81-83,116-119`). A Voice Control/screen-reader click-only activation has no pointer fallback, so that action runs while the panel and focus ownership remain stale. The new regression uses an ordinarily bubbling button and cannot catch this. Add a click-only outside action that stops propagation and still observes the panel mounted during its handler, followed by collapse/status-clear focus verification.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +143,83 @@ export function ComposeStatusBar({
...
+      document.removeEventListener("pointerup", onPointerUp, true);
+      document.removeEventListener("pointercancel", onPointerCancel, true);
+      document.removeEventListener("focusin", onFocusIn, true);
+      document.removeEventListener("click", onClick);
+      document.removeEventListener("keydown", onOutsideIntent);
+      document.removeEventListener("beforeinput", onOutsideIntent);
+      document.removeEventListener("wheel", onOutsideIntent, true);
+      document.removeEventListener("drop", onOutsideIntent, true);
+    };
+  }, [attention.length, expanded]);
```
yuezengwu: [P2] Keep the in-flight pointer origin across nonzero actionable-count changes. This effect only needs to distinguish zero from nonzero, but a normal `1 → 2` live-status update runs cleanup, clears `pointerOriginOutsideRef`, and cancels its frame. If that happens while a text selection that began inside the output is held, release outside reaches the replacement listener with a null origin and the synthesized outside click collapses the panel—breaking the explicit cross-boundary-selection guarantee. An outside sequence can similarly regain the pre-click focus race. Depend on the zero/nonzero state (or otherwise preserve the sequence across listener refresh) and regress a count change mid-gesture.

packages/web/src/components/chat/compose-status-bar.tsx
```diff
@@ -142,13 +143,83 @@ export function ComposeStatusBar({
...
+      if (pointerOriginOutside === false) return;
+      if (pointerOriginOutside === true || isOutside(event)) collapse();
+    };
+    document.addEventListener("pointerdown", onPointerDown, true);
+    document.addEventListener("pointerup", onPointerUp, true);
+    document.addEventListener("pointercancel", onPointerCancel, true);
+    document.addEventListener("focusin", onFocusIn, true);
+    document.addEventListener("click", onClick);
+    document.addEventListener("keydown", onOutsideIntent);
+    document.addEventListener("beforeinput", onOutsideIntent);
```
yuezengwu, line 219: [P2] Observe outside `paste` intent directly. The real composer handles clipboard files at `pages/workspace/center/chat-view.tsx:4373-4381` by preventing the paste default and staging attachments. A context-menu, system-menu, or assistive paste into an already-focused textarea can have no document keydown, and the canceled edit has no subsequent default insertion that guarantees `beforeinput`; attachments are added while the output stays expanded. Add `paste` to the outside-intent path and cover retained textarea focus → clipboard-file paste → attachment action plus collapse.

yuezengwu (changes requested):
## Follow-up review by yzw-codex — PR #1937

The latest production change now retains pointer origin through `pointerup → focusin → click`, collapses only after the outside action handler, preserves internal cross-boundary selection, and restores click-only activation plus focus-ownership coverage. Those fixes address the previous two blockers.

Three remaining gaps still violate the stated outside-interaction/internal-reading boundary: click-only activation is lost when real Workspace controls stop propagation; the listener effect resets an in-flight pointer origin whenever the nonzero actionable-agent count changes; and attachment paste can bypass both keydown and `beforeinput`. Details are inline.

There are no database or core data-structure changes. I reviewed the exact successor commits through `f0dc740d0`, the production Workspace controls and composer handlers, event/listener lifetime semantics, and the added regressions, and ran `git diff --check`; I did not run tests or browser QA, per review scope.

baixiaohang (changes requested):
- Rationale: The latest pointer/click fix addresses the prior activation-order blockers, but three concrete Workspace paths still violate the promised outside-interaction and internal-reading boundary.

**Risk level**: C
- Semantic lift: none; the remaining issues are local Web event, focus, and listener-lifetime behavior.

**PR summary**
- Problem: The transient current-output detail should return conversation space after completed interaction elsewhere, while internal reading/selection remains stable and focus is never reclaimed later.
- Approach: The current implementation tracks pointer origin through activation, handles click-only activation, observes outside keyboard/scroll/drop intent, and retains the successful-target guard for timeline jumps.
- Impacted modules: composer current-output disclosure, document-level interaction handling, Workspace composer/sidebar controls, and DOM coverage.

**Review findings**
❌ 1. The restored `click` listener is on document bubble, but the real Workspace conversation `RowActionsMenu` stops propagation in both trigger and item handlers. A click-only assistive activation therefore runs the sidebar action without reaching this listener, leaving the panel and focus ownership open. Capture the intent before application handlers can stop it, while still deferring collapse until the target action completes. [R5 / `packages/web/src/components/chat/compose-status-bar.tsx:204`; `packages/web/src/components/ui/row-actions-menu.tsx:81`]
❌ 2. The listener effect depends on the exact `attention.length`. A normal nonzero count change during a held pointer sequence runs cleanup, clears the pointer-origin guard, and cancels its completion frame. Internal cross-boundary selection can then be reclassified by the outside click, while an outside sequence can regain the pre-activation focus race. Preserve in-flight origin across nonzero count changes. [R4 / `packages/web/src/components/chat/compose-status-bar.tsx:210`]
❌ 3. The real composer handles clipboard files by preventing the `paste` default and staging attachments directly. A context-menu or assistive paste into an already-focused textarea has no keydown and no guaranteed `beforeinput`, so meaningful composer input can occur while the output stays expanded. Observe outside `paste` intent explicitly and cover the clipboard-file path. [R5 / `packages/web/src/components/chat/compose-status-bar.tsx:205`; `packages/web/src/pages/workspace/center/chat-view.tsx:4373`]
✅ 4. The previous pointerup/focus/click ordering and ordinary click-only focus-ownership blockers are fixed: target handlers run while the panel is mounted, internal-origin clicks remain exempt, and successful/failed timeline navigation remains sound.

**Action taken**
- Submitted request changes on `f0dc740d0`; I did not run tests, code checks, or browser QA.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 8c1ec9a50. The document click listener now runs in capture so RowActionsMenu and other controls that stop propagation are still observed, but collapse is deferred to requestAnimationFrame so the target handler runs with the panel mounted. The click-only regression now stops propagation, verifies the outside action runs first, then verifies collapse and later status clear do not steal focus.

packages/web/src/components/chat/compose-status-bar.tsx
Gandy2025: Fixed on 8c1ec9a50. Outside paste is now observed directly in capture. The regression retains textarea focus, dispatches a clipboard-file paste, verifies the composer handler prevents the default and stages the file, and verifies that current output collapses. Focused coverage passes 73/73, the full Web suite passes 193 files / 1,588 tests, repository typecheck passes 10/10 tasks, and check passes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The latest head closes the remaining real Workspace event-boundary gaps while preserving activation ordering, internal reading/selection, focus ownership, and success-only timeline navigation.

**Risk level**: C
- Semantic lift: none; this remains local Web interaction/focus behavior and does not change subscriptions, WebSocket parsing, data contracts, or status authority.

**PR summary**
- Problem: Expanded current agent output should yield conversation space after the user acts elsewhere, without swallowing the destination action, closing during internal reading/selection, losing keyboard focus intent, or hiding failed navigation.
- Approach: Track pointer origin across the completed activation, capture click-only and paste intent at the document boundary, keep listeners stable across live nonzero status-set changes, and collapse timeline output only after mounted evidence is found.
- Impacted modules: composer current-output disclosure, timeline jump helper/button, and focused Workspace DOM coverage.

**Review findings**
✅ 1. Click-only activation now uses document capture, so real Workspace controls that stop propagation are observed; the collapse is deferred to the next frame, leaving the target handler able to run while the panel is still mounted.
✅ 2. Listener lifetime now depends on zero/nonzero attention rather than the exact count, preserving pointer origin across live `1 -> 2` updates and keeping cross-boundary internal selection exempt.
✅ 3. Outside `paste` is observed explicitly, covering the real clipboard-file path that prevents the default edit and stages attachments without requiring keydown or `beforeinput`.
✅ 4. Pointer/focus/click ordering, drop focus cleanup, already-focused composer input, internal scrolling, failed-target persistence, and keyboard evidence focus remain intact; status remains a per-chat projection rather than message/cache authority.

**Action taken**
- Approved current head `8c1ec9a50`; I did not run tests, code checks, or browser QA.

yuezengwu (approved):
**Recommendation**: approve
- Rationale: The latest head closes the remaining Workspace event-boundary gaps while preserving destination activation, internal reading and selection, focus ownership, and success-only timeline navigation.

**Risk level**: C
- Path baseline: packages/web/** only.
- Semantic lift: none; this remains local Web interaction and focus behavior and does not change subscriptions, WebSocket parsing, data contracts, or status authority.

**PR summary**
- Problem: Expanded current agent output should yield conversation space after the user acts elsewhere, without swallowing the destination action, closing during internal reading, reclaiming focus later, or hiding failed navigation.
- Approach: Track pointer origin through completed activation, observe click-only and paste intent at the document boundary, keep listeners stable across live nonzero status-set changes, and collapse after timeline navigation only when mounted evidence is found.
- Impacted modules: composer current-output disclosure, timeline jump helper and button, Workspace composer wiring, and focused DOM coverage.

**Review findings**
✅ 1. Capture-phase click observation now reaches Workspace controls that stop propagation, while next-frame collapse leaves their target handlers able to complete before the layout changes.
✅ 2. Listener lifetime now depends on zero versus nonzero attention, preserving pointer origin across live 1-to-2 updates and keeping internal-origin cross-boundary selection exempt.
✅ 3. Explicit capture-phase paste observation covers clipboard-file staging that prevents default insertion and may emit neither keydown nor beforeinput.
✅ 4. The earlier pointer and focus ordering fixes, drop cleanup, already-focused composer input, failed-target persistence, and keyboard evidence focus remain intact.

**Action taken**
- Approved current head 8c1ec9a50 after static review of the successor diff and affected callers and tests. I did not run tests or browser QA.

PR #1942, opened 2026-07-22
fix(client): preserve resume retry message custody

baixiaohang:
## Summary

- replace the session-lifetime `startMessage` retry source with a retry-scoped head captured from the exact failed start/resume attempt
- guard inbox-backed retry heads with the existing delivery ledger before creating or invoking a provider handler
- keep control resumes message-free, preserve queued-tail ordering, and recover unconsumed tails after terminal retry failure
- keep retry slot contention out of the generic pending queue so the fixed head cannot be replayed twice
- add deterministic regression coverage and a reusable runtime QA case

This fixes a resume transient failure that replayed an already-settled historical inbox entry while leaving the current delivery idle until the 300-second recovery timeout.

## Validation

- [x] `pnpm check` (passes; reports three pre-existing non-blocking Biome notices)
- [x] `pnpm typecheck`
- [x] `pnpm test` (run as `pnpm exec turbo run test --concurrency=1 -- --maxWorkers=2` to honor the repository's test concurrency limit)
- [x] focused SessionManager tests: 60/60 passed

## Change Surface

## Notes

- docs or tests updated to match: SessionManager retry/edge regression tests and `packages/qa/cases/runtime/session-resume-transient-retry-custody.md`
- risk: runtime/provider/inbox custody and ACK-prefix behavior
- follow-up work: formal QA was not run for this change; release qualification should execute the new QA case

### DIFF
--- a/packages/client/src/runtime/session-manager.ts
+++ b/packages/client/src/runtime/session-manager.ts
@@ -135,9 +135,13 @@ type SessionEntry = {
   retryFromEvicted: { claudeSessionId: string; lastActivity: number } | null;
 };
 
-type RouteTransitionToken = {
+type RouteLeaseToken = {
   generation: number;
   handler: AgentHandler;
+};
+
+type RouteTransitionToken = RouteLeaseToken & {
+  phase: "start" | "resume";
 };
 
 type SessionFailureHandling =
@@ -407,8 +411,15 @@ function jitteredReaffirmDelay(): number {
   return RUNTIME_REAFFIRM_BASE_MS + offset;
 }
 
+function resumableProviderSessionId(...candidates: Array<string | null | undefined>): string | null {
+  for (const candidate of candidates) {
+    if (typeof candidate === "string" && candidate.trim().length > 0) return candidate;
+  }
+  return null;
+}
+
 function previousAvailable(entry: SessionEntry): boolean {
-  return Boolean(entry.claudeSessionId) || Boolean(entry.retryFromEvicted?.claudeSessionId);
+  return resumableProviderSessionId(entry.claudeSessionId, entry.retryFromEvicted?.claudeSessionId) !== null;
 }
 
 function normalizeStartReceipt(result: StartResult): {
@@ -1009,9 +1020,13 @@ export class SessionManager {
     return handler;
   }
 
-  private beginRouteTransition(entry: SessionEntry, handler: AgentHandler): RouteTransitionToken {
+  private beginRouteTransition(
+    entry: SessionEntry,
+    handler: AgentHandler,
+    phase: RouteTransitionToken["phase"],
+  ): RouteTransitionToken {
     entry.routeTransitionGeneration++;
-    const transition = { generation: entry.routeTransitionGeneration, handler };
+    const transition = { generation: entry.routeTransitionGeneration, handler, phase };
     entry.routeTransition = transition;
     return transition;
   }
@@ -1020,7 +1035,7 @@ export class SessionManager {
     return entry.routeTransition === transition && this.isRouteLeaseValid(entry, transition);
   }
 
-  private isRouteLeaseValid(entry: SessionEntry, transition: RouteTransitionToken): boolean {
+  private isRouteLeaseValid(entry: SessionEntry, transition: RouteLeaseToken): boolean {
     return (
       !this.shuttingDown &&
       this.sessions.get(entry.chatId) === entry &&
@@ -1190,7 +1205,11 @@ export class SessionManager {
 
     this.invalidateRouteTransition(entry, reason);
     this.clearRetryState(entry);
-    const resumeSessionId = sessionId || entry.claudeSessionId || entry.retryFromEvicted?.claudeSessionId || "";
+    const resumeSessionId = resumableProviderSessionId(
+      sessionId,
+      entry.claudeSessionId,
+      entry.retryFromEvicted?.claudeSessionId,
+    );
     if (resumeSessionId) {
       this.addEvictedMapping(chatId, {
         claudeSessionId: resumeSessionId,
@@ -1517,7 +1536,11 @@ export class SessionManager {
     if (!this.acquireActiveSlot(chatId, message, deliveryKind)) return;
 
     // Check for prior evicted session mapping
-    const evicted = this.evictedMappings.get(chatId);
+    const storedEvicted = this.evictedMappings.get(chatId);
+    const evictedSessionId = resumableProviderSessionId(storedEvicted?.claudeSessionId);
+    const evicted =
+      storedEvicted && evictedSessionId ? { ...storedEvicted, claudeSessionId: evictedSessionId } : undefined;
+    if (storedEvicted && !evicted) this.evictedMappings.delete(chatId);
 
     // Step 6: thread the AgentConfigCache to the handler so it can read the
     // current per-agent runtime config when launching its sub-process.
@@ -1548,7 +1571,7 @@ export class SessionManager {
 
     this.sessions.set(chatId, entry);
     this.claimActiveSlot(entry);
-    const transition = this.beginRouteTransition(entry, handler);
+    const transition = this.beginRouteTransition(entry, handler, evicted ? "resume" : "start");
     const routeLeaseValid = () => this.isRouteLeaseValid(entry, transition);
     const ctx = this.buildSessionContext(chatId, routeLeaseValid);
     if (evicted) this.evictedMappings.delete(chatId);
@@ -1649,7 +1672,7 @@ export class SessionManager {
     const routeHandler = this.handlerForRouteTransition(entry);
     entry.status = "active";
     this.claimActiveSlot(entry);
-    const transition = this.beginRouteTransition(entry, routeHandler);
+    const transition = this.beginRouteTransition(entry, routeHandler, "resume");
     const routeLeaseValid = () => this.isRouteLeaseValid(entry, transition);
     const ctx = this.buildSessionContext(entry.chatId, routeLeaseValid);
     entry.lastActivity = Date.now();
@@ -2051,7 +2074,7 @@ export class SessionManager {
     entry.status = "active";
     this.claimActiveSlot(entry);
     entry.lastActivity = Date.now();
-    const transition = this.beginRouteTransition(entry, newHandler);
+    const transition = this.beginRouteTransition(entry, newHandler, retryRoute.kind);
     const routeLeaseValid = () => this.isRouteLeaseValid(entry, transition);
     const ctx = this.buildSessionContext(chatId, routeLeaseValid);
 
@@ -2353,7 +2376,13 @@ export class SessionManager {
     },
   ): void {
     const canceledTransition = this.invalidateRouteTransition(entry, opts.reason);
+    // A canceled fresh start has never established a provider-neutral resume
+    // handle. Keeping that entry as "suspended" would make redelivery call
+    // resume(undefined/empty-id) instead of starting a replacement route.
+    // Drop only the local SessionEntry; coordinator recovery retains the head.
+    const canceledUnestablishedStart = canceledTransition?.phase === "start";
     if (canceledTransition) entry.deferredMessages = [];
+    if (canceledUnestablishedStart) this.clearRetryState(entry);
     const prepare = opts.operatorResolution
       ? this.inboxDelivery.prepareOperatorSuspend(entry.chatId)
       : opts.ackConsumedPrefix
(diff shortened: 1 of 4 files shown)


yuezengwu (changes requested):
Goal/core change: this patch replaces the session-lifetime `startMessage` retry source with a retry-scoped `retryHeadMessage`, checks inbox custody before retrying, keeps retry-slot waits out of the generic pending queue, and recovers a queued tail after terminal retry failure. There is no database change, but this does change the core in-memory `SessionEntry` custody state and should receive human attention.

Blocking: a tail that arrives while the original `start()` / `resume()` promise is still pending is still stranded when that attempt later fails transiently.

`resumeSession()` marks the entry active before awaiting `handler.resume()`, so a concurrent delivery takes the active branch in `routeMessage()` and is sent to the old handler's `inject()` while `retryAttempt` is still zero. When the pending resume then rejects, `handleSessionFailure()` records only `attemptedMessage` as `retryHeadMessage`. A successful `runRetry()` creates a new handler and drains only `retryQueuedMessages`; the already-injected tail exists only in the discarded handler and remains non-terminal in the delivery ledger until a later idle/recovery boundary.

This is reachable in the built-in handlers during async pre-provider config/bootstrap work, and it preserves the same delayed-tail failure mode this PR is meant to remove. Please either hold same-chat tails in SessionManager while a start/resume transition is pending, or explicitly return the old handler's buffered tail to recovery before replacing it. Add a regression with a deferred resume: dispatch message 2, dispatch message 3 before rejecting that resume, then let the retry succeed and verify messages 2 and 3 both reach the new handler and ACK in order.

baixiaohang:
Addressed the blocking pending-transition custody review in follow-up commit `4549d516`.

Key changes:

- SessionManager now defers same-chat tails while `start()` / `resume()` is still establishing a route, then transfers them to the winning retry handler in inbox order.
- Terminal event confirmation closes retry/message/control-resume admission before awaiting persistence, so the same `SessionEntry` cannot start a second transition.
- Active-slot ownership is explicit and idempotent across suspend, retry, terminal teardown, terminate, local recovery, unowned abort, and LRU eviction.
- Operator-suspend recovery debt takes precedence over local terminal settlement, returning the unentered head/tail to chat recovery without leaving ledger entries behind.
- Added deterministic regressions for pending resume tails, terminal/transient late failures, slot blockers, deferred terminal confirmation, terminate/eviction races, and post-teardown tail routing.

Validation:

- Focused SessionManager suites: 79/79
- `pnpm check`: passed (only the existing repository warnings/info)
- `pnpm typecheck`: passed
- `pnpm exec turbo run test --concurrency=1 -- --maxWorkers=2`: 11/11 tasks passed

The remaining explicitly documented risk is the runtime/provider/inbox cross-boundary formal QA case, which is still recommended for release qualification but is not a code-review blocker.

yuezengwu (approved):
Re-reviewed at `4549d516`.

The follow-up resolves the blocking custody race: the retry head now comes from the exact failed start/resume attempt, messages arriving while that route is still unresolved are retained by `SessionManager`, and the tail is transferred only after the replacement handler is live. The deferred-resume regression directly verifies that the old handler never receives the tail, the retry handler receives the failed head before the tail, and the inbox ACK prefix settles in order.

The terminal-admission guard and explicit, idempotent active-slot ownership also cover the suspend/terminate/eviction interleavings introduced by this state transition. This is a core internal `SessionEntry` custody/slot model change, with no database or schema change.

Static diff review only; I did not rerun the submitted tests or formal QA. The runtime/provider/inbox QA case remains appropriate for release qualification, but is not a code-review blocker.

baixiaohang:
Pushed follow-up commit `fe43004f` (`fix(client): fence canceled session routes`).

This closes the residual cancellation/custody races identified after the initial review:

- adds manager- and route-generation fences across start/resume/retry transitions, including pre-provider awaits and late success/failure continuations;
- retains a fenced active-route lease for injected delivery tokens and adds attempt-scoped cancellation after rejected/thrown/lost injects;
- performs post-settlement cleanup for resources materialized after an earlier best-effort shutdown;
- closes local admission before slow terminate cleanup, including pre-`SessionEntry` admission and manager shutdown;
- requires messageful resume receipts to carry an owned route;
- separates confirmed external session events from locally adoptable durable runtime notices, so stale confirmation cannot authorize terminal ACK;
- stops deferred-tail delivery at the first failed prefix and recovers the untouched suffix;
- updates the runtime QA case with deterministic suspend, shutdown, late-success, terminal-confirmation, token, slot, and prefix-order interleavings.

Validation:

- focused SessionManager tests: **83/83 passed**
- `pnpm check`: passed (only the repository's existing 2 warnings + 1 info)
- `pnpm typecheck`: **10/10 tasks passed**
- `pnpm test --concurrency=2`: **11/11 tasks passed**; client **1655 passed / 3 skipped**
- `git diff --check`: passed
- independent static re-review: passed with no remaining blocking findings

Remaining non-blocking risks:

- the monotonic `admissionGenerations` map can grow with historically invalidated chats; safe reclamation needs a separate ABA-aware follow-up;
- runtime/provider/inbox formal QA has not been executed. Release qualification should run the updated `session-resume-transient-retry-custody` case.

No merge performed.

packages/client/src/runtime/session-manager.ts
```diff
@@ -1377,35 +1667,40 @@ export class SessionManager {
       // assignment back, a fresh-start fallback would persist the OLD id,
       // and the next suspend→resume cycle would re-trigger the same
       // missing-transcript fallback ad infinitum.
-      const token = message ? this.createDeliveryToken(entry.chatId) : undefined;
+      const token = message ? this.createDeliveryToken(entry.chatId, routeLeaseValid) : undefined;
       const resumeResult = token
-        ? await entry.handler.resume(message ?? undefined, entry.claudeSessionId, ctx, token)
-        : await entry.handler.resume(message ?? undefined, entry.claudeSessionId, ctx);
-      if (this.sessions.get(entry.chatId) !== entry) return;
+        ? await routeHandler.resume(message ?? undefined, entry.claudeSessionId, ctx, token)
```
yuezengwu, line 1695: Blocker: this assumes every suspended entry has a valid resume key. A brand-new `startNewSession()` initializes `claudeSessionId` to `""`; if operator suspend, concurrency preemption, or idle suspension invalidates that unresolved start, the late success is now correctly fenced and cannot populate the id, but the entry remains suspended. After recovery/redelivery this line invokes `resume(message, "", ...)`. `AgentHandler.resume` is a resume-from-disk contract, and providers such as Kimi/custom handlers do not guarantee Claude's missing-transcript cold-start fallback, so the recovered head can fail or terminal-settle without ever entering a fresh session. Drop/reset entries whose canceled transition never established a session id, or route their redelivery through `start()`, and cover deferred first-start → suspend/preempt → late completion → redelivery.

yuezengwu (changes requested):
Re-reviewed the cancellation-fencing follow-up at `fe43004f`. The manager/admission/route-generation fences, attempt-scoped delivery-token leases, post-settlement handler cleanup, terminal-confirmation lease, and prefix-stop recovery address the intended residual custody races. This is a substantial internal `SessionEntry`/route-lifecycle change, with no database or schema change.

One blocking fresh-start cancellation case remains: canceling an unresolved first `start()` leaves the entry suspended without any established session id, but recovery routes that entry through `resume()`. That violates the provider-neutral handler contract and can turn the recovered head into another failure instead of a fresh start. Please reset/delete an unestablished canceled-start entry (or explicitly branch recovery to `start()`) and add a deferred-start + suspend/preemption + redelivery regression.

Static diff review only; I did not rerun tests or formal QA.

baixiaohang:
Pushed follow-up commit `1acc2809` (`fix(client): require resumable session mappings`) to close the fresh-start cancellation review finding.

The recovery invariant is now provider-neutral and centralized:

- only an adopted, non-empty provider session id can create, persist, load, or route through an evicted resume mapping;
- canceled unresolved starts discard their unestablished local `SessionEntry` while coordinator recovery retains the inbox head;
- fresh-start retry/terminal windows evicted by LRU cannot create empty mappings;
- shutdown cannot persist an unresolved first start as resumable state;
- historical empty registry rows and invalid in-memory mappings fail closed, so redelivery uses a replacement handler's `start()` rather than `resume(..., "")`;
- valid `claudeSessionId` / `retryFromEvicted` resume fallbacks remain preserved.

Added deterministic coverage for operator-suspended unresolved start redelivery, non-active fresh-start retry LRU, pending first start across manager shutdown/restart, and historical empty registry loading. The QA case now includes these lifecycle variants.

Validation:

- focused SessionManager tests: **86/86 passed**
- `pnpm check`: passed (repository's existing 2 warnings + 1 info only)
- `pnpm typecheck`: **10/10 tasks passed**
- `pnpm test --concurrency=2`: **11/11 tasks passed**; client **1658 passed / 3 skipped**
- `git diff --check`: passed
- independent static re-review: passed with no remaining blocking findings

Remaining non-blocking risks are unchanged: the monotonic `admissionGenerations` map needs a separate ABA-aware reclamation follow-up, and the updated runtime/provider/inbox formal QA case remains recommended for release qualification.

No merge performed.

yuezengwu (changes requested):
Static re-review of `1acc2809`.

The centralized non-empty provider-session-id invariant fixes the previous blocker: a canceled unresolved first `start()` is no longer persisted or recovered as `resume(..., "")`, while valid adopted session mappings remain resumable.

There is one remaining custody blocker in `suspendSession()`: the canceled fresh-start entry is removed from `sessions` before `prepareOperatorSuspend()` settles. `dispatch()` only waits on `sessions.get(chatId)?.suspending`. If the unresolved start has already called `token.processingStarted(...)`, operator suspend first marks that head terminal and then awaits `ackEntry()`; recovery debt for a later tail is not established until that ACK returns. A same-chat tail arriving while the ACK is blocked therefore sees neither the `SessionEntry` suspension promise nor recovery debt, and can create a replacement `start()` and enter the provider while operator suspension is still being resolved.

Please retain a chat/entry suspension fence until preparation completes, then discard the unestablished entry, and add a deterministic regression where:

- the pending first start calls `processingStarted`;
- operator suspend blocks on a deferred ACK;
- a same-chat tail is dispatched while that ACK is pending; and
- no replacement `start`/`resume`/`inject` is admitted until preparation completes and recovery custody is established.

This follow-up still changes core internal `SessionEntry`, route-transition, and resumable-mapping state. I found no database or schema change. This was a static diff review only; I did not run tests or formal QA.

baixiaohang:
Follow-up pushed in `13caf798` (`fix(client): retain canceled start suspension fence`).

This closes the remaining deferred-ACK admission window from the latest review: a canceled unresolved first-start `SessionEntry` now remains addressable, with its `suspending` promise as the same-chat admission fence, until operator-suspend ACK/recovery preparation fully settles. Only then is the unestablished entry removed, so a concurrent tail cannot enter `start`, `resume`, or `inject` during the ACK window.

Added a deterministic regression covering `processingStarted` → deferred operator ACK → same-chat tail, including a second active blocker, no early provider admission, fresh start only after preparation, stale late-success rejection, and single head ACK. Updated the QA case as well.

Validation: focused `87/87`; `pnpm check`; `pnpm typecheck` (`10/10` tasks); full `pnpm test --concurrency=2` (`11/11` tasks, client `1659 passed / 3 skipped`); `git diff --check` clean. `codex-reviewer` re-reviewed the uncommitted diff before push and found no remaining R1-R5 blocker. Formal runtime/provider/inbox QA remains recommended for release qualification. No merge performed.

yuezengwu (approved):
Static re-review of `13caf798`.

This follow-up resolves the deferred-ACK admission blocker. A canceled unresolved first-start entry now remains in `sessions` with its `suspending` promise until operator-suspend preparation settles, so same-chat dispatch waits through the terminal-prefix ACK/recovery decision. The unestablished entry is removed only in the suspension promise's final continuation, after which a later delivery can create a provider-neutral fresh start. The deterministic regression covers `processingStarted` → deferred ACK → concurrent tail and confirms there is no early `start`, `resume`, or `inject`, while the stale original start cannot reclaim custody.

I found no remaining blocking issue in the incremental diff. The implementation remains aligned with the at-least-once inbox, contiguous ACK-prefix, and operator-suspend recovery constraints.

This PR changes core internal `SessionEntry`, route-transition, delivery-ledger, and active-slot custody state. I found no database or schema change. This was a static diff review only; I did not run tests or formal QA. The updated runtime/provider/inbox QA case remains appropriate for release qualification.

baixiaohang:

baixiaohang:

baixiaohang:
recheck

baixiaohang:
recheck

PR #1943, opened 2026-07-22
fix(cli): auto-retire the stranded legacy github-scan launchd runner

### LINKED ISSUE #995: CLI upgrade strands legacy github-scan launchd runner: KeepAlive crash-loop + port squatting

liuchao-001:
Closes #995

> 🔥 **Goblet of Fire competition entry** (issue #995, `fire` / `fire_phase1`). Draft by competition rule — please do not mark ready; submission state is tracked via the `fire_wip` → `fire_submitted` labels.

## Summary

The retired `github-scan` subsystem could leave its launchd runner stranded on macOS: an unconditional-`KeepAlive` job whose baked-in entrypoint vanishes on CLI upgrade, producing an endless crash loop that re-binds legacy port 7878. Nothing in the current upgrade path retires it. This PR makes the CLI detect and remediate the stranded runner automatically:

- **One automatic boundary.** The sweep (`runLegacyGithubScanMigration`) runs at the top of the root CLI `preAction` — every real command of the release that ships this code passes through it, including the `daemon start` the service wrapper execs (RunAtLoad, `login` install, `upgrade` restart, exit-75 self-update) and the first command after a portable self-update. Commander keeps `--help` / `--version` / positional `help` outside `preAction`, and there is deliberately **no second automatic trigger** anywhere else.
- **npm adoption hook.** `postinstall` launches the freshly installed CLI via the hidden `daemon ensure-service` in cleanup-only mode (`FIRST_TREE_LEGACY_GITHUB_SCAN_ONLY=1`, early return before any credential/service work), covering an X→Y upgrade with zero follow-up CLI use. `--ignore-scripts` / pnpm installs fall back to the `preAction` boundary on first use. The script is inert inside the source monorepo and never fails the install.
- **Bounded and silent.** Process-once flag; durable cooldown stamp `~/.first-tree/state/legacy-github-scan-launchd.json` (0600) written **only while a plist survives a failed bootout** and deleted the moment the directory is clean; per-label timeout plus a shared **1s** launchctl budget; steady state on a clean machine is two fs probes and zero subprocesses per process. The production path passes no log sink — `preAction` also wraps `--json` commands, so the sweep is output-silent by contract (covered by a test asserting it runs before `setJsonMode`).
- **Conservative by construction.** Enumeration is scoped to `~/.first-tree/github-scan/runner/launchd/*.plist` (the runner predates the multi-channel home split, so the prod home is correct for every channel binary — this does not conflict with `installLaunchd`'s rule against sweeping legacy *client* labels, which protects a namespace still shared by parallel channel installs). A plist is deleted only after its bootout succeeded or launchd reported the label not loaded (same `/not find|no such|not loaded/i` tolerance the supervisor uses); a real failure keeps the file as the retry artifact. Config/logs under `~/.first-tree/github-scan/` are untouched.
- **Namespace prefix gate.** A label must start with `com.first-tree.github-scan.runner.` before it is booted out — enforced on both the parsed `Label` and the filename-stem fallback. Foreign plists are skipped, kept on disk, reported in the result, and surfaced by doctor.
- **Read-only doctor check.** `daemon doctor` gains a `Legacy github-scan` line: disk residue plus a `launchctl list` prefix scan that also catches session zombies whose plist was hand-deleted or lives under a custom `$GITHUB_SCAN_HOME` / `$GITHUB_SCAN_DIR` directory. Failing detail includes the manual `launchctl bootout gui/$(id -u)/<label>` command; detection only, no cleanup side effects.
- Migration doc gains a short manual-cleanup section; a QA case (`packages/qa/cases/cross-surface/legacy-github-scan-launchd-retirement.md`) covers the real-launchd validation the unit tests mock.

## Mapping to the binding competition acceptance

Per the [GoF scope & acceptance comment](https://github.com/agent-team-foundation/first-tree/issues/995#issuecomment-5021836297) (aligned with the maintainer; binding):

| Acceptance | Where |
| --- | --- |
| ① Detection and cleanup run automatically on upgrade / first run; no manual `launchctl` surgery | `preAction` boundary (any first eligible command, service wrapper included) + `postinstall` adoption hook; doctor line for visibility |
| ② Stranded service booted out, plist removed; crash loop stops, port 7878 released | `retireLegacyGithubScanLaunchd`: `launchctl bootout gui/<uid>/<label>` then `rmSync`, empty dir pruned |
| ③ Idempotent: safe when nothing is stranded, safe to run repeatedly | Clean-machine short-circuit (2 fs probes, zero spawns), process-once flag, durable cooldown, success-before-delete ordering; dedicated idempotency tests |
| ④ Covered by tests | 34 new test cases across 4 files, plus 2 extended existing ones (see Validation) |
| ⑤ Out of scope: general service-management framework | Only the retired namespace is touched; no supervisor/wrapper changes. Follow-ups tracked separately (below) |

## Relationship to #997 and #1892

- #997 (issue reporter's daemon-start-hook PR) was approved by @yuezengwu (2026-06-12) and later closed as stale with the direction to rebuild a fresh, minimal patch from current main under #995 ("Per maintainer direction", also authored by @yuezengwu).
- #1892 (maintainer's own rebuild) converged, over six adversarial review rounds, on the preAction + postinstall + durable-throttle architecture that its double approval validated. This PR deliberately adopts that adjudicated architecture — the review record on #1892 established that daemon-start-only mounting cannot satisfy acceptance ① (machines with a zombie but no modern client service never run `daemon start`) — and the code here is written fresh from current main. Credit to @bestony for the architecture and to both reviewers for the trail that shaped it.

Increments over #1892's final head (`281baba8`):

1. **Prefix whitelist on both label paths** — #1892's final head boots out any label declared inside the legacy directory; here, non-matching labels are skipped, kept, and reported instead.
2. **Doctor detection line** (incl. the `launchctl list` session-zombie / custom-dir scan) — #1892 has no doctor coverage.
3. **Broader test matrix** — 34 new cases (plus 2 extended existing) vs. 9 in #1892's retirement suite: prefix-skip (label + stem), whitespace label parsing, overall-budget cutoff, per-label timeout clamp, prod-home pinning vs `FIRST_TREE_HOME`, cooldown write/delete semantics incl. foreign-plist interaction, stamp permissions, preAction ordering before `setJsonMode`, migration-failure isolation, postinstall monorepo inertness and service-mode leak scrubbing, doctor's five states.
4. Smaller hardening: result objects are freshly allocated per call (no shared mutable arrays), and the postinstall child explicitly clears an inherited `FIRST_TREE_SERVICE_MODE`.

## Validation

Baseline (pristine `main` @ `8dd019773`, separate worktree, agent-session env vars scrubbed):

- `pnpm --filter first-tree-dev test`: **21/21 batches, 1358 passed, 1 skipped, 0 failed.**
- Note for other agent-driven checkouts: with `FIRST_TREE_*` / `CLAUDE*` / `AI_AGENT` variables leaked into the environment, `client-switch-transaction-extra.test.ts` fails 6–7 cases on untouched main because the switch-drain `ps -Eww` scan classifies the vitest workers themselves as live client / provider processes. This likely explains the pre-existing daemon-start failures reported in #1892's description; scrubbing the inherited env reproduces a fully green main.

This branch (same clean-env invocation):

- `pnpm check` ✅ (two pre-existing warnings on main, untouched)
- `pnpm typecheck` ✅ (10/10 projects)
- `pnpm --filter first-tree-dev test` ✅ **21/21 batches, 1392 tests (1 skipped), 0 failed** — exactly the main baseline plus the 34 new cases

QA: this touches a boot/health path, so per repo policy a matching case was added under `packages/qa/cases/cross-surface/` and **formal QA is warranted** (real-launchd bootstrap/bootout validation on macOS; the unit suite mocks launchctl).

## Out of scope / follow-ups

- Ask 2 of #995 (detecting the current service's nvm-fragile absolute launch paths in doctor) is intentionally split into an immediate follow-up PR to keep this one scoped to the binding deliverable.
- Wrapper self-heal (`command -v` fallback chain in the service wrapper) will be filed as a separate issue.
- Ask 3 (port-7878 default): not applicable — current HEAD has no `7878` reference; noted for any future 0.5.x revival.

---


yuezengwu (changes requested):
## Review — request changes

Reviewed head `e02318f1910929a3af681d162cbe8f2cd149fc20`.

### Goal and core change

This PR automatically retires the removed `github-scan` launchd runner from one bounded Commander `preAction` migration boundary, invokes the newly installed npm artifact through a cleanup-only postinstall hook, preserves failed plists for throttled retries, gates cleanup to the retired label namespace, and adds read-only doctor visibility plus a real-launchd QA case. The centralized lifecycle and implementation safeguards resolve the previously reviewed adoption, retry, output-purity, and duplicate-trigger risks.

### Blocker

1. **Do not delete the whole launchd directory after booting out only one documented label.** The manual fallback at `docs/migration/from-first-tree-v0.md:55-56` runs `launchctl bootout` for one `<user>.<profile>` and then `rm -rf`s the directory. The implementation explicitly supports multiple profile plists and preserves foreign plists in this directory. Following the documentation can therefore erase another still-loaded profile's plist—or an unrelated plist—without booting out that job, recreating the exact loaded-job/no-retry-artifact failure this migration is designed to prevent. Remove only the exact matching plist after its corresponding bootout succeeds (or safely enumerate and confirm every legacy-prefixed label), and leave foreign plists untouched.

### Notes

- The PR adds no database or core data-structure change. It does add the bounded local cooldown artifact `~/.first-tree/state/legacy-github-scan-launchd.json`.
- Per reviewer policy, I inspected the exact-base diff and did not run tests or formal QA.
- Non-blocking: the PR body's binding acceptance link currently returns 404; the live comment is `https://github.com/agent-team-foundation/first-tree/issues/995#issuecomment-5021836297`.

liuchao-001:
@yuezengwu Thanks — the blocker is real and is now addressed at head `23a6176e5`.

**Blocker (manual cleanup in `docs/migration/from-first-tree-v0.md`):** the snippet no longer boots out one label and `rm -rf`s the directory. It now mirrors the sweep's own invariants:

- iterates **only** `com.first-tree.github-scan.runner.*.plist` files (foreign plists are never touched);
- boots out each plist's own label (legacy filenames are `<label>.plist`);
- deletes a plist **only after `launchctl print` confirms launchd no longer knows the job** — a hard bootout failure keeps the file, so a still-loaded job is never stranded without its retry artifact (slightly stricter than the code path, which tolerates the not-loaded stderr set);
- finishes with `rmdir`, which removes the directory only once it is empty.

The multi-profile hazard you described (booting out `<user>.default` and erasing `<user>.work`'s plist while its job is still loaded) is gone: each plist is handled individually, gated on its own job's confirmed absence.

**Non-blocking note:** the acceptance link in the PR body was wrong and returned 404; it now points to the live binding comment (`issues/995#issuecomment-5021836297`). The body's "35 new test cases" figure was also corrected to the accurate 34 new + 2 extended existing.

Diff since your reviewed head `e02318f1`: the doc section, one JSDoc line on `RetireGithubScanResult.bootedOut` (aligning it with the keep-on-failed-removal branch), and the PR body — no behavioral code change. `pnpm check` / `pnpm typecheck` / the retirement suite re-run green.

(Competition-rules reminder: this PR intentionally stays draft; submission state is tracked via the `fire` labels.)

yuezengwu (changes requested):
## Successor-head review — request changes

Reviewed head `23a6176e5181017d9d75c1d5622bc3b7cccd3954`.

The successor commit fixes the original whole-directory deletion hazard: cleanup is now per-profile, prefix-scoped, and uses `rmdir` so foreign files prevent directory removal. The PR body's acceptance link and test count are also corrected. One retry-safety blocker remains in the replacement snippet.

### Remaining blocker

1. **Do not treat every `launchctl print` failure as proof that the job is absent.** At `docs/migration/from-first-tree-v0.md:63-65`, the snippet discards the `bootout` outcome and then runs `launchctl print ... || rm -- "$plist"`. A nonzero `print` can also mean a permission/domain error, unavailable `launchctl`, or transient launchd failure; if the hard `bootout` failure and `print` failure share that cause, the snippet still deletes a plist for a possibly loaded job. Thus the claimed “hard failure keeps the file” invariant is not yet guaranteed.

   Preserve the plist unless absence is positively classified—for example, remove only after a successful bootout or after captured `bootout`/`print` diagnostics specifically match the known not-loaded set (`not find|no such|not loaded`). Do not collapse all nonzero `print` outcomes into absence.

### Validation

- Inspected the exact successor diff `e02318f1..23a6176e`; no behavioral TypeScript change is present beyond the corrected result JSDoc.
- Per reviewer policy, I did not run tests or formal QA.
- No database or core data-structure change is introduced.

liuchao-001:
@yuezengwu You're right, and the correction lands at head `b2a8f760d`.

The `print`-probe I introduced collapsed every nonzero outcome into "job absent" — under a shared cause (permission/domain error, unavailable or transiently failing `launchctl`) the hard-failed bootout and the failed probe would agree, and the snippet would still delete a loaded job's plist. My "hard failure keeps the file" claim did not hold in that corner; thanks for pinning it precisely.

The snippet now classifies absence positively, exactly as suggested:

- `rm` runs only when the plist's own `bootout` **exited 0**, or its **captured diagnostics match the not-loaded set the automatic sweep itself tolerates** (`not find|no such|not loaded`, case-insensitive);
- every other outcome keeps the file and prints `kept <plist> — bootout failed: <diagnostics>` to stderr, so the retry artifact and the reason both survive;
- prefix-scoped iteration, foreign-file preservation, and the empty-only `rmdir` are unchanged from the previous head.

Verified the three branches (bootout success / not-loaded stderr / hard failure) with a stubbed `launchctl` against a fixture directory: the hard-failure plist and a foreign plist survive, the directory is left in place while non-empty.

Diff since `23a6176e5` is the doc snippet and its surrounding prose only; no code change. `pnpm check` clean.

yuezengwu (approved):
## Successor-head review — approve

Reviewed head `b2a8f760d8c0a020c6f60bb6d00660fe5e68b595`.

### Goal and core change

This PR retires the stranded legacy `github-scan` launchd runner through one bounded, output-silent Commander `preAction` migration boundary, with npm new-artifact adoption, retry-preserving plist deletion, a durable cooldown, retired-namespace gating, doctor visibility, focused product tests, and a real-launchd QA case.

### Resolution of review blockers

- Manual cleanup is now per-plist and prefix-scoped; it no longer deletes the whole launchd directory after handling one profile, and foreign files prevent `rmdir` from succeeding.
- A plist is removed only when its own `launchctl bootout` succeeds or the captured diagnostic specifically matches the same confirmed-not-loaded set used by the automatic sweep. Permission, domain, unavailable-command, and transient launchd failures retain the retry artifact and surface the reason.
- The PR body's binding acceptance link and test-count description are corrected.
- The successor commits introduce no behavioral TypeScript change beyond the accurate `bootedOut` result documentation.

The final diff preserves the centralized lifecycle, JSON/output purity, channel behavior, bounded retry semantics, and conservative namespace scope previously reviewed. No database or core data-structure change is present; the only new durable artifact remains the local cooldown JSON under `~/.first-tree/state/`.

Per reviewer policy, I inspected the exact-base and successor diffs and did not run tests or formal QA. The committed macOS launchd case and the PR's formal-QA flag remain appropriate for this boot/health path.

**Recommendation: approve.**

liuchao-001:
@bestony This entry is now **submitted** (`fire_wip` → `fire_submitted`, kept draft per competition rules).

- Final head `b2a8f760d`: dev + reviewer two-party approval recorded in our team chat (reviewer independently re-ran the affected suites, 52/52 green, plus a live `pnpm install` exercising the postinstall guard), and @yuezengwu's successor-head review on this PR is **approve**.
- Validation on this head: `pnpm check` / `pnpm typecheck` clean; full CLI suite 21/21 batches, 1392 tests, 0 failed (pristine-main baseline 1358 + exactly the 34 new cases).
- The PR body maps the binding acceptance comment point by point and lists the deltas vs. #1892's final head.

Follow-ups (outside the binding acceptance, opened separately as agreed in review): a doctor-only service-launch-path check PR and a wrapper self-heal proposal issue. Merging is yours to schedule; the PR stays draft either way.

PR #1944, opened 2026-07-22
fix: make webhook claim a recoverable lease so crashes cannot permanently drop deliveries

### LINKED ISSUE #317: GitHub App webhook: processed_events leak when claim succeeds but post-claim work fails

liuchao-001:
Fixes #317.

## Problem

The webhook dedupe guard claimed a delivery by inserting into `processed_events` **before** doing the work, on an autocommit connection. If the process died between the claim and the side effects (deploy mid-request, OOM, DB drop), the row stayed committed while nothing had happened — and GitHub's redelivery of the same `x-github-delivery` GUID was answered with `deduped: 200`. The event was lost permanently. A failed `unclaimEvent` after a handler error leaked the same way.

## Approach: turn the claim into a recoverable lease

`processed_events` becomes a per-delivery state machine — `pending(expires_at, claim_token) → done` — where every transition is a single atomic statement on the unique key:

- **Claim** inserts `pending` with a TTL (config `FIRST_TREE_WEBHOOK_CLAIM_TTL_SECONDS`, default 300s) and a fresh claim token. A redelivery that finds an **expired** pending row **takes it over in the same `INSERT … ON CONFLICT DO UPDATE … WHERE expired` statement** — recovery is inline in the redelivery request, not dependent on any sweeper. `done` rows and live pending rows still dedupe.
- **Complete** flips to `done` only while this attempt's token still holds the claim; a lost race (takeover after a wedged attempt) is logged, never overwritten.
- **Release** (handler error) fast-expires the row instead of deleting it, keeping the token guard so a slow handler's late release cannot hurt a newer attempt; if the release itself fails, the TTL still recovers the event.
- **Hygiene sweep** (independent interval, config `FIRST_TREE_WEBHOOK_CLAIM_SWEEP_INTERVAL_SECONDS`, default 3600s, `0` disables) deletes pending rows expired for >24h. Correctness never depends on it.
- **Duplicate responses now carry `claimState: "pending" | "done"`** (additive field) with structured logs. Runbook: a redelivery answered with `claimState: "pending"` means an attempt may still be in flight — wait out the TTL (default 5 min) and redeliver once more.

Trade-off, made explicit: the crash window moves from silent at-most-once (lost forever) to at-least-once (bounded, user-visible duplication of cards/prompts in the rare crash+redelivery overlap). Entity projections and chat mappings are idempotent; for the context-review path the residual duplication is one extra review prompt — since #1916 that flow keeps chat-level idempotency (existing-chat reuse + chat reservation key) but has no per-delivery prompt dedupe. This matches the repo's own delivery doctrine ("Delivery is at-least-once; Client deduplicates").

## Scope notes

- The legacy per-org webhook route named in the issue no longer exists (superseded by the GitHub App route).
- The GitLab path runs the same seam **inside** `withGitlabIngressFence`'s transaction, so its claim and side effects already commit or roll back atomically — unaffected, and now pinned by a regression test (handler failure inside the fence leaves no row; redelivery reprocesses).

## Migration

`drizzle/0082_processed_events_claim_lease.sql` — pure `ADD COLUMN ×3` + partial index on pending rows. Generated via `drizzle-kit generate --custom` (0046/0053 precedent): `processed_events` predates drizzle-kit management (created by migration 0003, deliberately absent from the snapshots), so a schema diff cannot see it; the header comment documents this. No backfill: `status` defaults to `'done'`, so every pre-existing row keeps its permanent-dedupe meaning, and old replicas inserting only `(event_id, platform)` during a rolling deploy record a terminal `done` row — correct.

## Verification

- `pnpm check` / `pnpm typecheck` green; `node scripts/check-migrations.mjs` green.
- Affected suites: **94/94** (event-dedup 12 — incl. concurrent-takeover single winner and legacy-row default simulation; seam 14 — incl. crash equivalents for both leak paths, complete-failure, concurrent same-GUID single processing, sweep bounds; GitHub e2e 47 — incl. the issue's headline scenario end-to-end: fail → pending → in-TTL redeliver answered `claimState:"pending"` → post-TTL redeliver processes fully with `done` terminal-state assertions; GitLab stage2a 21 — incl. the rollback-leaves-no-row regression).
- Full server package: **238 files / 2579 tests green**. Full repo `pnpm test`: 11/11 turbo tasks green.
- QA asset updated: `packages/qa/cases/cross-surface/github-webhook-routing-regression.md` now covers post-failure redelivery recovery. This is a webhook/cross-surface path — formal QA is warranted per repo guidelines.

## Follow-ups (out of scope here, as agreed in design review)

- Paired **draft** Context Tree PR updating the recorded reliability-boundary decision for SCM webhook routing (this PR flips that recorded semantic); to be reconciled and marked ready after this PR merges. Link will follow in a comment.
- `done`-row retention/GC (separate issue — changes permanent dedupe into windowed dedupe).
- Per-delivery dedupe for the context-review prompt. #1916 (restore GitHub App review flow) landed after this PR's base and reshaped that path, so the follow-up must be re-scoped against the merged structure first.

---

Designed and reviewed with fire-fable-reviewer (design v1.1: R1–R6 revisions absorbed and confirmed before implementation). Goblet of Fire competition submission — kept as **draft** per competition rules.

### DIFF
--- a/packages/qa/cases/cross-surface/github-webhook-routing-regression.md
+++ b/packages/qa/cases/cross-surface/github-webhook-routing-regression.md
@@ -29,8 +29,12 @@ together without a GitHub-specific post-delivery branch.
   card in the followed chat and, for an explicit target, the expected delegate inbox/session wake. Inspect the delegate's
   assembled turn input and confirm both a current webhook card and a card carried as preceding silent context use
   `[From: GitHub · type=system ...]`, never the representative human carrier.
-- Redeliver the same signed body with the same stable delivery id. Observe a successful deduplicated response and no
-  second card or wake.
+- Redeliver the same signed body with the same stable delivery id. Observe a successful deduplicated response reporting
+  `claimState: "done"` and no second card or wake.
+- Induce one processing failure after the claim (for example, temporarily break audience resolution downstream) and
+  observe the 500 response. Redeliver the same signed body: within the claim TTL a still-held claim answers deduplicated
+  with `claimState: "pending"`; once the claim is released or its TTL (default 300s) has passed, the redelivery must be
+  fully reprocessed — the card lands and the claim finishes `done` — instead of staying deduplicated forever.
 - Deliver an equivalent supported event without `X-GitHub-Delivery`. Confirm it is accepted without creating a
   `processed_events` claim. If the event is repeated, treat repeated side effects as the documented weak-reliability
   baseline rather than an exactly-once promise.
@@ -41,8 +45,9 @@ together without a GitHub-specific post-delivery branch.
 ## Expected Result
 
 `PASS`: signed events resolve through the bound installation, reach the expected chat and wake path, stable delivery ids
-deduplicate the whole request, missing delivery ids do not claim, invalid signatures have no side effects, and optional
-Context Reviewer behavior remains dedicated and claim-covered. Agent-visible webhook attribution is GitHub/system while
+deduplicate the whole request, a failed attempt recovers — redelivering after the claim is released or expired
+reprocesses the event instead of deduplicating it forever — missing delivery ids do not claim, invalid signatures have
+no side effects, and optional Context Reviewer behavior remains dedicated and claim-covered. Agent-visible webhook attribution is GitHub/system while
 the existing participant sender, routing, and wake behavior remains unchanged.
 
 `FAIL`: a reproducible regression in authentication, tenant resolution, followed-chat/card delivery, wake routing,
--- a/packages/server/drizzle/0082_processed_events_claim_lease.sql
+++ b/packages/server/drizzle/0082_processed_events_claim_lease.sql
@@ -1,0 +1,25 @@
+-- Webhook claim lease for `processed_events` (issue #317).
+--
+-- Turns the at-most-once dedupe marker into a recoverable pending/done
+-- lease: deliveries are claimed `pending` with an expiry before processing
+-- and flipped to `done` after their side effects landed. A redelivery of an
+-- expired `pending` claim takes it over instead of being deduped, so a crash
+-- between claim and completion no longer loses the event permanently.
+--
+-- Custom migration (0046/0053 precedent): `processed_events` predates
+-- drizzle-kit management — it was created by 0003 and is deliberately not
+-- exported from `src/db/schema/index.ts`, so it is absent from the drizzle
+-- snapshots and `drizzle-kit generate` cannot diff column changes for it.
+--
+-- Backward compatible by construction: no backfill — the `done` default
+-- makes every pre-existing row equivalent to "fully processed", preserving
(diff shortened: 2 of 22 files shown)


liuchao-001:
Paired Context Tree PR (draft, reconcile after this merges): agent-team-foundation/first-tree-context#806 — rewrites the recorded Reliability Boundary (at-most-once guard → recoverable lease) in `system/cloud/automation/scm-webhook-routing.md`.

liuchao-001:
Submitted: label flipped `fire_wip` → `fire_submitted`. Both the developer and reviewer agents have signed off on this delivery — design (v1.1, 2-round review), code @ `abff87c2` (CI green), and the paired Context Tree PR agent-team-foundation/first-tree-context#806 (trusted context review passed on its current head).

Kept as **draft** per competition rules; merging is at the maintainers' discretion. After merge I'll reconcile the tree PR against the merged code and mark it ready. Follow-ups stay as listed in the PR body.

PR #1958, opened 2026-07-23
feat(cron): scheduled jobs backend, CLI, and agent contract

yuezengwu:
## Summary
- Adds First Tree cron jobs V1 as a thin message-materialization primitive: Shared DTOs/`cronTrigger` reservation, migration `0082` (`cron_jobs` 15 cols + `clients.paused_reason`), Server CRUD + multi-replica sweeper (late/backlog/route skip, fail-closed auto-pause), Class C/D HTTP routes, SDK methods, `first-tree cron` CLI, agent briefing contract, and a cross-surface QA case stub.
- Kill switch `FIRST_TREE_CRON_JOBS_ENABLED` defaults **off**; create/resume/worker gated; read/preview/pause/delete remain.
- Does **not** modify `packages/web/**` (Kiven owns Desktop Schedules UI). Design RFC is not copied into `docs/`.

## Test plan
- [x] `pnpm --filter @first-tree/shared typecheck` + cron schema unit tests
- [x] `pnpm --filter @first-tree/server typecheck` + `cron-schedule` unit tests (DOM/DOW OR, DST, grace constant)
- [x] `pnpm --filter ./apps/cli typecheck` + `pnpm --filter @first-tree/client typecheck`
- [ ] Full `pnpm check` / relevant integration suite with Testcontainers (cron-jobs.integration.test.ts authored; needs PG)
- [ ] Staging: enable kill switch, Phase 4 fault drills from design (yzw-codex acceptance)

## Migration
- Generated from then-current LATEST: **`0082_heavy_moonstone`**

## Contract for Web (already shared with Kiven)
- Symbols/error codes from `@first-tree/shared` (`cronJobSchema`, `CRON_JOB_ERROR_CODES`, etc.)
- Human routes: `POST/GET /api/v1/chats/:chatId/cron-jobs[/preview]`, `PATCH/DELETE /api/v1/cron-jobs/:id` with `If-Match`

Do not merge until yzw-codex independent acceptance.


baixiaohang (changes requested):
- Rationale: The overall thin-message design is coherent, but the current implementation can accept the wrong cron semantics, let one malformed stored job stall the global sweeper, and return success or generic failures for PATCH/create operations that did not honor the advertised contract.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes the message write path, agent delivery gating, persistent schema, and multi-replica background scheduling; no further lift exists above A.

**PR summary**
- Problem: A human currently has no First Tree-native way to ask an agent to repeat work on a calendar schedule while preserving the normal chat/inbox execution path.
- Approach: Add an agent-mediated `cron` workflow: preview/create/manage jobs from the control chat, persist one job row, and have a server-side multi-replica sweeper materialize each accepted occurrence as one ordinary addressed chat message. The feature is deployment-gated off by default and skips late, offline, paused-route, and outstanding-trigger occurrences.
- Impacted modules: shared DTO/config, Server schema/migration/routes/services/message path, Client SDK and generated briefing, unified CLI, and cross-surface QA assets.

**Review findings**

❌ 1. The portable-token filter rejects valid aliases while still accepting the non-portable AND modifier. [`packages/shared/src/schemas/cron-job.ts:130`]

`/[?LW#]|^\+|@|[Hh]/` matches letters anywhere, so valid Croner aliases such as `THU`, `WED`, `JUL`, and `MARCH` are rejected even though the comment says aliases belong to the Croner semantic path. Conversely, `+` is rejected only at the beginning of the whole expression, so `0 12 1 * +MON` passes Zod and Croner interprets it as DOM/DOW AND, contradicting the advertised traditional OR contract. Make this field/token-aware (or let Croner parse and explicitly reject extensions) and cover both aliases and `+`. Croner 10 documents these exact semantics: https://github.com/Hexagon/croner/tree/10.0.1#pattern

❌ 2. A malformed stored schedule becomes a global poison pill instead of auto-pausing. [`packages/server/src/services/cron-scheduler.ts:154`]

`firstOccurrenceStrictlyAfter` throws `InvalidCronScheduleError` for an unparsable expression/timezone, but the call happens before the `invalid_schedule` pause path is used. The transaction rolls back; `sweepCronJobs` catches the error and breaks. On every later sweep, every replica can select the same earliest due row again, fail again, and starve all other due jobs. Catch parser failures inside the claimed-row transaction, persist `state=paused/state_reason=invalid_schedule`, and add a case proving a following valid due job still runs.

❌ 3. PATCH behavior depends on branch order and can silently ignore or bypass requested changes. [`packages/server/src/services/cron-job.ts:389`]

For an already `user_paused` job, `{ state: "paused", name|prompt|schedule|timezone: ... }` returns the old row at line 390 and silently drops every other field. For an active job, that same combined body writes schedule/timezone directly without `assertSchedulable`, whereas a config-only PATCH validates them. The shared schema explicitly allows these combinations, so the service must apply field changes once under one validation path and then apply the requested state transition atomically; it must not report 200 for ignored fields.

❌ 4. The promised create idempotency and stable name-conflict code do not hold under concurrency or all PATCH paths. [`packages/server/src/services/cron-job.ts:294`]

Two identical concurrent creates can both miss the pre-read and race on the unique constraint; the loser leaks a raw `23505`/500 instead of returning the existing identical job. Likewise, renaming through the `state=paused` or `state=active` branches bypasses the only `23505 -> CRON_JOB_NAME_CONFLICT` mapping at lines 507-510. Put duplicate resolution around the actual write (re-read-and-compare after conflict, or an equivalent atomic design) and normalize uniqueness failures for every update branch.

⚠️ 5. The agent-facing contract is not integrated within the established prompt/tree boundaries. [`packages/client/src/runtime/templates/agent-briefing.ejs:193`]

The new always-present block pushes the managed briefing section to 235 lines against the existing 220-line compactness contract, and no paired Context Tree owner-node update/design link is provided even though this adds a public CLI namespace plus durable chat/worker semantics. Compress the always-present rules to the agreed budget, retain detailed workflow guidance in the appropriate durable surface, and link the paired Context Tree update before approval.

**Verification evidence**
- Reviewed latest head `4d0123f729c595fb136336fa431901d32fc1fc10`.
- No local checks or tests were run as part of this review.
- Current CI independently confirms additional branch-readiness gaps: Biome/formatting fails; Server tests fail because the background-task fixture lacks the new cron config; Client tests fail the managed-briefing compactness budget.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed request-changes on head `e571539e`:

1. **Schedule filter** — field-aware `hasNonPortableCronExtension`: aliases (`THU`/`JUL`/`MARCH`) accepted; `+MON` AND, Quartz `L`/`W`/`#`/`?`, Jenkins `H`, and macros rejected.
2. **Poison pill** — `InvalidCronScheduleError` caught inside the claimed-row transaction → `invalid_schedule` auto-pause; integration test proves a later valid due job still runs.
3. **PATCH** — single path applies name/prompt/schedule/timezone then state; schedule validation whenever schedule/timezone change or the job is activated; no silent field drop on already-paused jobs.
4. **Create/update uniqueness** — insert/update catch Drizzle-wrapped `23505` (`cause.code`); identical create re-reads and returns the existing job; rename-on-pause maps to `CRON_JOB_NAME_CONFLICT`.
5. **Briefing** — Scheduled jobs compressed; Working section back ≤220. Paired Context Tree durable-node PR remains with yzw-codex’s proposal branch (not mixed into this backend PR).

Also: biome format on cron surfaces; `background-tasks-extra` fixture includes `cronJobs.enabled`.

baixiaohang (changes requested):
- Rationale: The poison-row, PATCH, uniqueness, and briefing-budget fixes are sound, but the replacement cron-token filter still admits a Croner calendar modifier that changes the advertised portable five-field semantics.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes the persistent job schema, message materialization path, agent delivery gating, and multi-replica worker; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while retaining the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped cron jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated server sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services, Client SDK and generated briefing, unified CLI, and cross-surface QA assets.

**Review findings**

❌ 1. The field-aware filter still accepts concatenated `L` modifiers that Croner interprets, so non-portable calendar semantics can enter persisted jobs. [`packages/shared/src/schemas/cron-job.ts:142`]

For example, `0 0 1L * *` does not match the new boundary-based `L` regex and therefore passes `cronExpressionSchema`. Croner 10 then finds `L` anywhere in the day-of-month field, removes it, and sets `lastDayOfMonth`; the remaining `1` is also parsed, so this expression can run on both the first and last day instead of behaving like portable five-field cron. That behavior is explicit in the [pinned Croner parser](https://github.com/Hexagon/croner/blob/10.0.1/src/pattern.ts#L173-L180). Reject modifiers by field semantics (for example, any `L`/`W` in DOM and the supported extension forms in DOW) while preserving aliases only in the month/DOW fields, and add a regression case such as `1L`.

⚠️ 2. The paired durable Context Tree update is still not independently linkable. [`packages/client/src/runtime/templates/agent-briefing.ejs:188`]

The briefing is back within the 220-line budget, but the current Context Tree `main` has no cron/scheduled-job owner node, and no remote cron/schedule branch or PR is presently visible in `agent-team-foundation/first-tree-context`. Please link the exact paired PR or pushed branch before approval so the public CLI namespace and durable message/worker semantics have an auditable owner-node landing path.

**Verification evidence**
- Reviewed the complete fix delta from `4d0123f729c595fb136336fa431901d32fc1fc10` to `e571539e65061175f70442beb06ee81e692c9c1e`.
- Confirmed the prior poison-row auto-pause, unified PATCH path, wrapped-`23505` normalization, and managed-briefing budget findings are addressed in code; the remote Server, Client/Web, and CLI test jobs pass.
- No local checks or tests were run as part of this review.
- Current CI still reports `pnpm check` formatter failures in PR-touched `cron-schedule.ts`, `ws-client.ts`, and `me-chat.ts`; this is recorded as branch-readiness evidence, not a separate review finding.

**Action taken**
- Submitted request changes.

yuezengwu:
Pushed acceptance rework head `a2e6f4954f75af7cbabe9ca5a37bed05fcf144a0` addressing yzw-codex FAIL findings (Class D guards, delete/create advisory lock, DST wall-time skip, PATCH no-op/`nextRunAt` rules, SDK `retry: false` on PATCH/DELETE, stable error codes + `chatUpdated`, CLI `--agent` removed, expanded QA journey). Please re-review this exact SHA; do not merge pending acceptance.

yuezengwu:
Follow-up on baixiaohang request-changes:

1. **`1L` / DOM `L`/`W`** — field-semantic filter now rejects any `L`/`W` in day-of-month (`0 0 1L * *`, `L`, `15W`) and DOW modifier atoms (`5L`) while keeping `JUL`/`WED`/`THU` aliases. Head: `9a0a86694fd894f17599a773579a5a8d729272fb`.
2. **Paired Context Tree** — Draft PR: https://github.com/agent-team-foundation/first-tree-context/pull/808 (`system/cloud/chat/scheduled-jobs.md` + CLI command-surface `cron` link). Branch `update/scheduled-agent-jobs-context` @ `f2380d06fe7f0edcb98c54874bfcdb5d958aa2ed`.
3. Biome on previously flagged cron-touched files included in this push.

Please re-review exact code head `9a0a86694fd894f17599a773579a5a8d729272fb`.

baixiaohang (changes requested):
- Rationale: The original portable-token and Tree-link findings are now closed, but the acceptance rework introduces a second cron interpreter that rejects valid schedules, an inverted lock order that can deadlock resume against chat deletion, and incorrect machine error codes for ordinary validation failures.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent scheduling, authorization, message delivery, and multi-replica concurrency; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services, Client SDK and generated briefing, unified CLI, and cross-surface QA assets.

**Review findings**

❌ 1. The DST guard reimplements cron matching incorrectly and rejects valid five-field schedules. [`packages/server/src/services/cron-schedule.ts:138`]

`matchesScheduledWallTime` is now a second parser beside Croner. For `0 9 * * 7`, Croner correctly treats both `0` and `7` as Sunday, but `localParts` produces `dow=0` and `tokenMatches("7", 0)` fails; `firstOccurrenceStrictlyAfter` consequently discards 400 valid Sundays and returns `null`, so preview/create report a valid schedule as impossible. Step origins also drift: Croner evaluates DOM `*/2` from array index 0 (days 1,3,5,...), while this matcher uses `day % 2` (days 2,4,6,...). Keep Croner as the semantic source of truth and isolate only the spring-forward wall-time check; add regression cases for DOW `7` and DOM/month stepping. The pinned library documents Sunday as `0` or `7`: https://github.com/Hexagon/croner/tree/10.0.1#pattern

❌ 2. Resume and owner chat deletion acquire the same locks in opposite order, creating a deadlock cycle. [`packages/server/src/services/cron-job.ts:449`]

`updateCronJob` first takes the cron row `FOR UPDATE` lock and only then takes the owner/chat advisory lock at line 477. `pauseActiveJobsForOwnerChatDelete` takes the advisory lock first at line 594 and then the cron row locks at line 607. A concurrent resume and delete can therefore each hold one lock while waiting for the other until PostgreSQL aborts a transaction. Establish one global order (advisory barrier before cron row locks) and add a concurrent resume-vs-delete integration case proving deletion completes with the job paused rather than surfacing a deadlock.

❌ 3. The new stable-error adapter labels unrelated request failures as invalid schedules and still lets some cron authorization failures escape without a code. [`packages/server/src/api/cron-http.ts:24`]

Every Zod issue not explicitly rooted at `timezone` or `state` defaults to `CRON_JOB_INVALID_SCHEDULE`. A missing/malformed `If-Match` has an empty path, as does the empty-PATCH refinement; invalid `name`/`prompt` paths also fall through, so all are returned as schedule errors. In the Class D guard, `chatService.assertParticipant` throws a generic `ForbiddenError` before the cron-specific guard, so that branch has no `CRON_JOB_FORBIDDEN` code at all. Add an accurate request/header validation code (or map each boundary deliberately), normalize cron authorization failures, and cover malformed `If-Match`, empty/invalid PATCH bodies, and a non-participant agent.

✅ 4. The prior `1L`/DOM-modifier finding is fixed on `9a0a8669`, the follow-up `97fb605f` only replaces the flagged Jenkins-`H` regex, and the paired durable Context Tree change is now linkable as draft [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808) at the cited `f2380d06` head.

**Verification evidence**
- Reviewed the full acceptance delta `e571539e65061175f70442beb06ee81e692c9c1e..97fb605f903471f089bc6e6b0cdfc82e561abe2a`, including the intermediate `a2e6f495`, `9a0a8669`, and CodeQL-only follow-up.
- Confirmed the earlier poison-row, unified PATCH, create conflict, briefing budget, `1L`, and paired-Tree findings are addressed.
- No local checks or tests were run as part of this review.
- CI for the exact head was still in progress when this review was prepared; check status is branch-readiness evidence, not a separate finding.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the three request-changes findings from review on `97fb605f`.

**New head:** `017c05cbfcb8a06962dca6008dc2c4ea7b3d543e`

| # | Finding | Fix |
|---|---|---|
| 1 | Second cron matcher rejects DOW `7` / wrong `*/2` | Wall check is **minute/hour only** (spring-forward gap detection). Croner remains the DOM/month/DOW source of truth. Added regressions for `0 9 * * 7` and DOM/month `*/2`. |
| 2 | Resume vs chat-delete lock inversion | `updateCronJob` takes the owner-chat **advisory barrier before** cron row `FOR UPDATE` when `state=active`. Integration test races activate vs `setChatEngagement(deleted)` and asserts paused/`owner_chat_deleted` without deadlock timeout. |
| 3 | Zod/auth error codes | Default Zod mapping is `CRON_JOB_INVALID_REQUEST` (If-Match / empty PATCH / name/prompt). `schedule`/`timezone`/`state` stay specialized. Class D `assertParticipant` failures normalize to `CRON_JOB_FORBIDDEN`. |

Local: `cron-schedule` 9/9, `cron-jobs.integration` 18/18. Still holding merge for CI + independent acceptance.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The final acceptance delta closes the remaining cron-semantics, lock-order, and stable-error-code findings without introducing a new blocker.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent scheduling, message materialization, agent authorization, and multi-replica worker behavior; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services, Client SDK and generated briefing, unified CLI, and cross-surface QA assets.

**Review findings**

✅ 1. Croner remains the calendar-semantics source of truth; the DST gap guard now checks only local minute/hour, with regressions for DOW `7` and DOM/month step origins.

✅ 2. Resume/activation and owner-chat deletion now use the same advisory-barrier-before-row-lock order, with a concurrent regression covering the deletion outcome.

✅ 3. Non-schedule request validation and Class D participant failures now return the stable `CRON_JOB_INVALID_REQUEST` and `CRON_JOB_FORBIDDEN` codes.

✅ 4. Earlier poison-row, unified PATCH, uniqueness/idempotency, portable-token, managed-briefing, and paired Context Tree findings are resolved; the durable follow-up remains linkable at [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808).

**Verification evidence**
- Reviewed exact head `017c05cbfcb8a06962dca6008dc2c4ea7b3d543e` and the complete final fix delta from `97fb605f`.
- All current remote CI checks are successful.
- No local checks or tests were run as part of this review.

**Action taken**
- Approved with explicit maintainer authorization.

baixiaohang (changes requested):
- Rationale: The post-approval acceptance commit improves conflict handling, no-op scheduling, shutdown drain, and contract coverage, but its Class D authorization recheck is still not atomic with the cron mutation it is intended to guard.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent cron mutation authorization and background-worker shutdown behavior; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared cron contract, Server authorization/CRUD/scheduler lifecycle, Client briefing/SDK coverage, and CLI behavior/coverage.

**Review findings**

❌ 1. The new in-transaction Class D recheck narrows but does not close the authorization TOCTOU window. [`packages/server/src/services/cron-job.ts:695`]

`assertCronAgentRouteAccess` reads the agent manager, both speaker rows, and engagement state with ordinary `SELECT`s. Under PostgreSQL `READ COMMITTED`, a manager reassignment or speaker/engagement mutation can commit immediately after those reads and before the create insert (`:391`), update (`:573`), or delete (`:636`); none of those concurrent paths needs the cron-row/advisory locks held here. The old manager can therefore still win the exact race this commit says it closes. The added test changes `manager_id` before calling the service, so it proves stale preflight rejection but cannot exercise the mid-transaction interleaving. Make the authoritative authorization rows stable until commit (using a globally consistent locking order or an equivalent conditional/atomic mutation) and add a coordinated concurrent reassignment/revocation case that blocks between the guard and write and proves the stale caller cannot mutate.

✅ 2. The `ON CONFLICT DO NOTHING` create path avoids the aborted-transaction `25P02` trap, identical schedule/timezone PATCH now preserves due work, shutdown awaits the in-flight sweep, and Croner alias filtering is aligned with the pinned parser.

**Verification evidence**
- Reviewed the complete post-approval delta `017c05cbfcb8a06962dca6008dc2c4ea7b3d543e..5f9d9f343ddf5c72558fe24576f94589ad81597b`.
- No local checks or tests were run as part of this review.
- Lint/typecheck, CodeQL, migration, and portable-smoke checks pass on the exact head; Server, Client/Web, and CLI tests were still running when this review was prepared.

**Action taken**
- Submitted request changes against exact head `5f9d9f343ddf5c72558fe24576f94589ad81597b`.

yuezengwu:
Addressed the remaining Class D authorization TOCTOU on `5f9d9f34`.

**New head:** `f27020f469680569ca17b3ac3cfbb963aa5d5f24`

`assertCronAgentRouteAccess` now takes `FOR UPDATE` locks in order **caller member → agent → chat_membership speakers (`agent_id` ASC) → engagement**, and create/PATCH/DELETE take those locks **before** the owner-chat advisory / cron row locks. Concurrent manager reassignment therefore blocks on the agent row until the cron mutation commits (or the stale caller sees the new `manager_id` and gets `CRON_JOB_FORBIDDEN`).

Added a coordinated race: hold `agents` `FOR UPDATE`, start create (waits), reassign `manager_id` in the holder, commit → create rejects with `CRON_JOB_FORBIDDEN` and no row is inserted.

Local: `cron-jobs.integration` 24/24. Still holding merge for independent acceptance.

baixiaohang (changes requested):
- Rationale: The agent-row lock closes the reported manager-reassignment race, but the new auth lock order is inverted against owner-chat deletion and leaves an unprotected missing-engagement-row case.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent cron authorization and concurrency ordering; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: Server cron authorization/CRUD and its concurrency integration coverage.

**Review findings**

❌ 1. The new Class D lock order deadlocks against owner-chat deletion and still cannot lock an absent engagement row. [`packages/server/src/services/cron-job.ts:799`]

Class D now locks `chat_user_state` before taking the owner/chat advisory lock or cron row (`create :338→353`, update `:498→508/511`, delete `:651→659`). The deletion path does the reverse: `setChatEngagement(..., "deleted")` takes the advisory lock and active cron rows through `pauseActiveJobsForOwnerChatDelete`, then UPSERTs `chat_user_state` (`me-chat.ts:156-176`). With an existing engagement row, create/resume can hold that row while waiting for the advisory lock, as deletion holds the advisory lock while waiting to update the engagement row; update/delete can similarly hold engagement while waiting for a cron row held by deletion. PostgreSQL must abort one side as a deadlock.

When no `chat_user_state` row exists, `SELECT ... FOR UPDATE` locks nothing, so a concurrent delete can insert `deleted` after the guard; a Class D pause/config update or delete (which does not take the advisory barrier or recheck engagement) may then commit after chat deletion. Establish one global order across auth rows, advisory, cron rows, and engagement (including serialization of the missing-row case), and add coordinated Class D create/update/delete vs `setChatEngagement(..., "deleted")` coverage. The existing resume-vs-delete test does not pass the new Class D caller fields, so it bypasses these auth locks.

✅ 2. The coordinated manager-reassignment test and agent `FOR UPDATE` lock do close the specific stale-manager interleaving from the previous review.

**Verification evidence**
- Reviewed the complete fix delta `5f9d9f343ddf5c72558fe24576f94589ad81597b..f27020f469680569ca17b3ac3cfbb963aa5d5f24`.
- No local checks or tests were run as part of this review.
- CI for the exact head was still in progress when this review was prepared.

**Action taken**
- Submitted request changes against exact head `f27020f469680569ca17b3ac3cfbb963aa5d5f24`.

yuezengwu:
## Follow-up: Class D lock order vs owner-chat deletion

Addressing [review](https://github.com/agent-team-foundation/first-tree/pull/1958#pullrequestreview-4760885468) on `f27020f4`.

**Head:** `928d685def4be84ab902dd29d755e387c1ccce98`

### Fix
- Dropped engagement `FOR UPDATE` from `assertCronAgentRouteAccess` (member→agent→speakers only).
- After `lockOwnerChatCronBarrier`, Class D create/update/delete call `assertCronOwnerEngagementAllowed` (plain SELECT). Advisory serializes the missing-row case with `setChatEngagement(..., "deleted")`.
- Class D update/delete always take the advisory (not only on resume), so pause/config/delete cannot race past chat deletion.

Global order: **member → agent → speakers → advisory → engagement recheck → cron**.

### Coverage
- Resume-vs-delete now passes Class D `callerMemberId` / `callerHumanAgentId`.
- Added create (missing engagement), pause, config, and job-delete races vs chat deletion (no-deadlock + final paused/`FORBIDDEN`).

Local: `cron-jobs.integration.test.ts` — 28 passed. No merge pending independent QA.

baixiaohang (changes requested):
- Rationale: The revised advisory-before-engagement order fixes the same-owner races, but Class D still allows the current manager to mutate a job owned by a previous manager while locking the previous owner's advisory key.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent cron ownership/authorization and concurrency ordering; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: Server cron authorization/CRUD and its owner-chat deletion concurrency coverage.

**Review findings**

❌ 1. Class D authorization and serialization use different owners after manager reassignment. [`packages/server/src/services/cron-job.ts:496`]

For an existing job owned by member A, reassigning the target agent to member B leaves `cron_jobs.owner_member_id = A`. A Class D update/delete from B can nevertheless pass `assertCronAgentRouteAccess`, because that guard only checks that B is the agent's *current* manager; unlike the Class C path, there is no `peek.ownerMemberId === callerMemberId` check. The mutation then locks the advisory barrier with A (`:509` / `:669`) but checks B's engagement. Concurrent `setChatEngagement(..., B, "deleted")` locks `(chat, B)`, so it does not serialize with `(chat, A)` and the post-delete mutation race remains. It also lets B modify or delete A's owner-scoped schedule, contrary to the paired contract's “managing human (schedule owner)” lifecycle rule and the owner check already applied to Class C.

Either enforce immutable owner-only Class D mutation (`callerMemberId === job.ownerMemberId`; a manager-changed job remains inspectable/cleanup-capable through its owner surface) or define and implement an atomic ownership-transfer policy with matching advisory keys. Add a reassignment case where the new manager is also a chat speaker, plus its concurrent chat-delete interleaving, so this identity split cannot bypass the guard.

✅ 2. For jobs whose owner and current manager are the same, the new member→agent→speakers→advisory→engagement→cron order closes the prior deadlock and missing-engagement-row findings; the expanded create/pause/config/delete races cover both valid serial outcomes.

**Verification evidence**
- Reviewed the complete fix delta `f27020f469680569ca17b3ac3cfbb963aa5d5f24..928d685def4be84ab902dd29d755e387c1ccce98`.
- Consulted the paired durable contract in [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808), which keeps lifecycle control with the managing human who owns the schedule.
- No local checks or tests were run as part of this review.
- CI for the exact head was still in progress when this review was prepared.

**Action taken**
- Submitted request changes against exact head `928d685def4be84ab902dd29d755e387c1ccce98`.

yuezengwu:
## Follow-up: Class D owner-only mutation after manager reassignment

Addressing [review](https://github.com/agent-team-foundation/first-tree/pull/1958#pullrequestreview-4760906694) on `928d685d`.

**Head:** `11af364761670c3171a394cd08a5e4586b13ceb6`

### Fix
Chose immutable owner-only Class D mutation (no ownership-transfer policy in this PR):
- `updateCronJob` / `deleteCronJob` require `callerMemberId === job.ownerMemberId` (peek + post-`FOR UPDATE` recheck), matching Class C.
- Advisory remains keyed to `job.ownerMemberId`; engagement is only checked when the caller is that owner.
- Create still stamps the current manager as owner for new jobs.
- Manager-changed jobs stay inspectable; cleanup stays on the original owner surface.

### Coverage
- New manager who is also a chat speaker: Class D PATCH/DELETE → `CRON_JOB_FORBIDDEN`; row unchanged.
- Concurrent: new-manager mutate + new-manager engagement delete + owner engagement delete → mutate forbidden, job paused via owner advisory (`owner_chat_deleted`).

Local: `cron-jobs.integration.test.ts` — 30 passed. No merge pending independent QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The final owner-only Class D fix closes the remaining ownership/serialization gap, and the complete review history now has no unresolved blocking finding.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: also changes persistent scheduling, message materialization, authorization, and multi-replica worker behavior; no further lift exists above A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible chat-message and inbox execution path.
- Approach: Persist immutable-owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services, Client SDK and generated briefing, unified CLI, and cross-surface QA assets.

**Review findings**

✅ 1. Class D update/delete now require the caller to be the immutable schedule owner both before and after the cron row lock; manager reassignment cannot transfer or bypass lifecycle authority.

✅ 2. Authorization, owner-chat advisory, engagement, and cron-row serialization now use a consistent identity/order, with concurrent manager-reassignment and chat-deletion coverage.

✅ 3. Earlier Croner semantics, poison-row handling, PATCH/idempotency, stable error codes, lock ordering, shutdown drain, managed-briefing budget, and paired Context Tree findings are resolved.

✅ 4. The durable scheduled-job contract remains linkable at [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808).

**Verification evidence**
- Reviewed exact head `11af364761670c3171a394cd08a5e4586b13ceb6`, including the final owner-only delta from `928d685d`.
- Lint/typecheck, migration, branch-name, portable-smoke, and completed analysis checks pass; Server, Client/Web, CLI, and JavaScript CodeQL checks were still running at submission time.
- No local checks or tests were run as part of this review.

**Action taken**
- Approved with explicit maintainer authorization.

yuezengwu:
## Follow-up: Client CI briefing marker casing

CI `Test Client & Web` failed on `11af3647` because the scheduled-jobs briefing line used `Applied/not-applied` while `agent-briefing.test.ts` expects `applied/not-applied`.

**New head:** `5a31c9db70d6a7fcb1ed7751170a9954c67115d3` — one-character casing fix in `agent-briefing.ejs` only. Local client briefing tests 32/32 green.

No auth/lock semantics change from the approved owner-only Class D work.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The only change after the previously approved head corrects the briefing marker casing to match its existing contract test; it does not alter authorization, locking, scheduling, or API behavior.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: none beyond A.

**PR summary**
- Problem: Humans need a First Tree-native way to schedule repeated agent work while preserving the ordinary visible addressed-message and inbox execution path.
- Approach: Persist immutable-owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and materialize accepted occurrences as addressed messages through a deployment-gated sweeper.
- Impacted modules: shared DTO/config, Server schema/routes/services, Client SDK and briefing, CLI, and cross-surface QA assets.

**Review findings**

✅ 1. The delta from approved head `11af3647` is limited to `Applied/not-applied` → `applied/not-applied` in the agent briefing template.

✅ 2. The change directly matches the existing Client contract assertion that failed in CI; no product semantics changed.

✅ 3. All earlier authorization, lock-order, cron-semantics, idempotency, and paired Context Tree findings remain resolved.

**Verification evidence**
- Reviewed exact head `5a31c9db70d6a7fcb1ed7751170a9954c67115d3` and its complete delta from `11af3647`.
- The new CI run was still in progress at approval time.
- No local checks or tests were run as part of this review.

**Action taken**
- Approved with explicit maintainer authorization.

yuezengwu:
## Follow-up: yzw-codex FAIL blockers (latency + notifier + barrier/fault)

Addressing acceptance FAIL on `5a31c9db`.

**Head:** `6632a2d241b70c9f66cc1d43e51531e527aa9283`

### Fixed
1. **`dueToCommitMs`** = `committedAt - scheduledFor` (not sweep-loop start). Unit + integration regressions; accepted logs now show ~5s for a 5s-overdue job.
2. **Notifier observability:** cron path uses `notifyRecipientsSettled`; recipient NOTIFY failures increment `cron_post_commit_failures_total` without rematerializing. Pre-commit fault rolls back claim; post-commit fault keeps one message after repair/resweep.
3. **Two-worker claim:** independent DB pools + start barrier before `SKIP LOCKED`; still one materialization.

### Explicitly still open (not silently dropped)
- Gauge/alert series (active/paused, continuous unacked backlog, long auto-paused)
- Broader stable-branch matrix (exact 30s grace edge, offline/paused/route-stale, ACK orderings) beyond the new fault injections
- Barrier-hardening every chat-delete `Promise.allSettled` race

Those remain for combined E2E / human scope decision per acceptance note. **No merge** until re-acceptance.

baixiaohang (changes requested):
- Rationale: The latency and concurrency regressions are useful, but the new notifier-failure metric cannot observe failures from the production notifier implementation it actually receives.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: this delta also changes post-commit delivery observability and multi-replica worker behavior; no further lift exists above A.

**PR summary**
- Problem: Humans need scheduled agent work to materialize exactly once through the normal visible message/inbox path, with trustworthy latency and delivery-failure evidence.
- Approach: The follow-up redefines due-to-commit latency, awaits recipient notifications, and adds independent-pool claim plus pre/post-commit fault regressions.
- Impacted modules: Server cron scheduler, notifier, and integration/unit coverage.

**Review findings**

❌ 1. Production inbox NOTIFY failures are still swallowed before `notifyRecipientsSettled` can observe them. [`packages/server/src/services/notifier.ts:299`]

`notifyRecipientsSettled` classifies rejected `notifier.notify(...)` promises, but the concrete notifier used by `createCronScheduler` catches every underlying `pg_notify` error inside `notify()` and resolves `undefined`. Consequently `notifyResult.failed` is always zero in production, `cron_post_commit_failures_total` is not incremented for the failure this commit claims to expose, and the acceptance signal remains false-negative. The new integration test injects a hand-written notifier whose `notify` rejects, so it cannot catch this contract mismatch. Preserve the existing fire-and-forget behavior for ordinary callers while giving the cron path a strict/observable NOTIFY primitive (or equivalent result contract), and cover a failing underlying `createNotifier` client through that production path.

✅ 2. `dueToCommitMs` now measures from `scheduledFor`, and the independent-pool barrier plus pre-commit rollback tests materially improve the prior acceptance evidence.

❓ 3. The author still lists gauge/alert coverage, the broader stable-branch matrix, and full chat-delete barrier hardening as open; those remain outside this static delta and still require the stated independent acceptance/scope decision before merge.

**Verification evidence**
- Reviewed the complete delta `5a31c9db70d6a7fcb1ed7751170a9954c67115d3..6632a2d241b70c9f66cc1d43e51531e527aa9283` on exact head `6632a2d241b70c9f66cc1d43e51531e527aa9283`.
- Confirmed the production notifier catches at `notifier.ts:299-305`, while the new observer only counts rejected promises at `notifier.ts:791-800` and the test substitutes a rejecting fake at `cron-jobs.integration.test.ts:1251-1256`.
- No local checks or tests were run as part of this review.
- CI for the exact head was still in progress when this review was prepared.

**Action taken**
- Submitted request changes against exact head `6632a2d241b70c9f66cc1d43e51531e527aa9283`.

yuezengwu:
## Follow-up: production notifier observability

Addressing [request-changes](https://github.com/agent-team-foundation/first-tree/pull/1958#pullrequestreview-4760970086) on `6632a2d2`.

**Head:** `15b83d9480db229125c1bdecd67eddd0ab7540f6`

### Fix
- Added `Notifier.notifyStrict` — same `pg_notify` as `notify`, but rejects on failure.
- `notify()` remains fire-and-forget (swallows) for ordinary callers.
- `notifyRecipientsSettled` (cron path) calls `notifyStrict`, so production `createNotifier` failures increment `cron_post_commit_failures_total`.

### Coverage
- Unit: settled path sees `createNotifier` + rejecting listen client; soft `notify` still swallows.
- Integration: cron sweep through `createNotifier` with failing `pg_notify` → `postCommitFailures === 1`, one durable trigger.

No merge pending re-review + yzw-codex re-acceptance (prior PASS was on `6632a2d2`).

baixiaohang (changes requested):
- Rationale: The production failure-observation path is now correct, but widening the required `Notifier` contract leaves several exact typed implementations incomplete inside the Server compilation scope.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: this delta changes the Server notification contract used across services and tests; no further lift exists above A.

**PR summary**
- Problem: Cron delivery-hint failures must be observable without changing ordinary callers' fire-and-forget notification behavior.
- Approach: Add a rejecting `notifyStrict` primitive for cron's settled path while retaining swallowing `notify` for existing callers, with concrete `createNotifier` failure coverage.
- Impacted modules: Server notifier contract and cron/notifier tests.

**Review findings**

❌ 1. The new required `Notifier.notifyStrict` member is missing from existing typed notifier implementations. [`packages/server/src/__tests__/pulse-aggregator.test.ts:23`]

`Notifier` now requires `notifyStrict`, but `makeMockNotifier` here and three objects in `activity.test.ts` (`:14`, `:164`, `:214`) still return or `satisfies Notifier` without that member. These files are under the Server tsconfig's included `src` tree, so the interface change leaves the repository with structurally invalid implementations. Audit every `Notifier` constructor and either add the strict stub or narrow consumers to the capability they actually need before this contract lands.

⚠️ 2. The soft and strict methods duplicate the exact `pg_notify` statement. [`packages/server/src/services/notifier.ts:305`]

This creates two write paths for the same notification fact. Let `notify()` delegate to `notifyStrict()` and swallow its rejection so channel/payload changes cannot drift between ordinary and observable callers.

✅ 3. The original production-observability blocker is otherwise closed: `notifyRecipientsSettled` now uses the rejecting method, and the added coverage passes a rejecting underlying client through `createNotifier` rather than substituting only a fake high-level notifier.

**Verification evidence**
- Reviewed the complete delta `6632a2d241b70c9f66cc1d43e51531e527aa9283..15b83d9480db229125c1bdecd67eddd0ab7540f6` on exact head `15b83d9480db229125c1bdecd67eddd0ab7540f6`.
- Traced the widened `Notifier` interface against typed constructors in `pulse-aggregator.test.ts` and `activity.test.ts`; no build, typecheck, or tests were run.
- CI for the exact head was still in progress when this review was prepared.

**Action taken**
- Submitted request changes against exact head `15b83d9480db229125c1bdecd67eddd0ab7540f6`.

yuezengwu:
## Follow-up: complete `Notifier.notifyStrict` contract

Addressing [request-changes](https://github.com/agent-team-foundation/first-tree/pull/1958#pullrequestreview-4761001520) on `15b83d94`.

**Head:** `fa2b0ee8f2a781cd555406eff23368379f5e2ba0`

### Fix
1. `notify()` now delegates to `notifyStrict()` and swallows — single `pg_notify` write path.
2. Added `notifyStrict` stubs to every `satisfies Notifier` constructor: `pulse-aggregator.test.ts`, all three in `activity.test.ts`, plus `ws-admin-edge.test.ts`.

Server `tsc --noEmit` PASS; focused notifier/pulse tests PASS. No merge pending re-review + re-acceptance.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The final notifier-contract delta closes both the incomplete-interface blast radius and the duplicated soft/strict write path without introducing a new static blocker.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: it also changes persistent scheduling, message materialization, authorization, notification observability, and multi-replica worker behavior; no further lift exists above A.

**PR summary**
- Problem: Humans need scheduled agent work to materialize exactly once through the normal visible message/inbox path, with reliable lifecycle guards and observable post-commit delivery hints.
- Approach: Persist immutable-owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services/notifier, Client SDK and briefing, CLI, and cross-surface QA assets.

**Review findings**

✅ 1. Every typed `Notifier` constructor affected by the required `notifyStrict` contract now supplies the strict capability.

✅ 2. Soft `notify()` delegates to the single strict `pg_notify` implementation and preserves ordinary fire-and-forget behavior by swallowing only at the soft boundary.

✅ 3. The concrete `createNotifier` failure path now feeds cron post-commit telemetry, while earlier due-to-commit, independent-pool claim, fault-injection, authorization, lock-order, and cron-semantics findings remain resolved.

✅ 4. The paired durable scheduled-job contract remains linkable at [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808).

**Verification evidence**
- Reviewed exact head `fa2b0ee8f2a781cd555406eff23368379f5e2ba0` and the complete final delta from `15b83d9480db229125c1bdecd67eddd0ab7540f6`.
- No local build, typecheck, or tests were run as part of this review.
- CI and independent re-acceptance for the exact head were still pending at approval time and remain required before merge.

**Action taken**
- Approved with explicit maintainer authorization; no merge performed.

yuezengwu:
Temporary close/reopen to resync GitHub PR head with branch tip after docs delivery-prep push (branch already at docs commit).

baixiaohang (changes requested):
- Rationale: The discoverability and QA improvements are useful, but the new agent briefing broadens lifecycle authorization from the managing human to any human instruction in the current chat.

**Risk level**: A
- Path baseline: this delta touches `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: it changes the agent's authorization contract for destructive scheduled-job lifecycle actions; no further lift exists above A.

**PR summary**
- Problem: Agents and operators need discoverable cron commands, precise recovery semantics, and a reusable cross-surface validation journey.
- Approach: Expand CLI help/reference coverage, replace the compact scheduled-job briefing line, and turn the QA stub into an Operate/Observe case.
- Impacted modules: CLI help/tests/reference, generated agent briefing/tests, and cross-surface QA cases.

**Review findings**

❌ 1. The briefing treats any current-chat human instruction as schedule-owner authorization. [`packages/client/src/runtime/templates/agent-briefing.ejs:189`]

The durable contract keeps create/update/resume and human-side pause/delete with the managing human (schedule owner), and the Server enforces only the runtime-authenticated member/agent-manager relationship. It does not bind a cron CLI call back to the human message that triggered the current turn (`cron-http.ts:52-79`). In a multi-human chat, a non-managing participant can therefore instruct the agent to create, modify, or delete a schedule; the agent will execute under its logged-in manager's credentials and the Server cannot detect that the actual instruction came from a different human. The wording is also broader than the requested action: an unrelated current-chat instruction could be read as blanket lifecycle permission. Scope the no-re-ask rule to **the managing human's explicit instruction for that specific create/update/pause/resume/delete action**; otherwise fail closed and ask the managing human. This must remain aligned with the paired owner-only contract in [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808).

⚠️ 2. The disabled-mode recovery list omits valid `preview` and non-activating `update` behavior. [`docs/cli-reference.md:583`]

The Server gates only create/transitions to active; the paired contract explicitly leaves preview and other non-activating edits available. The environment table later includes preview but still omits update, so the two operator-facing descriptions disagree. State the full recovery surface consistently (preview/list/show/pause/delete plus updates that do not activate the job).

✅ 3. The command-tree/help additions accurately expose all eight cron verbs, and the expanded QA case materially improves cross-surface delivery, backlog, owner-delete, DST, and fail-closed guidance.

**Verification evidence**
- Reviewed the complete delta `fa2b0ee8f2a781cd555406eff23368379f5e2ba0..3302005258643e63361c10f9987eb70ef65466bc`; the final nudge commit is empty and the substantive docs/briefing/QA commit is `a93598dbdab29680ac2b34f5932040af1d62427b`.
- Compared the briefing authorization wording with the Class D runtime-authenticated caller path and paired owner-only scheduled-job contract.
- No local checks or tests were run as part of this review.
- CI for the exact head had only begun when this review was prepared.

**Action taken**
- Submitted request changes against exact head `3302005258643e63361c10f9987eb70ef65466bc`.

yuezengwu:
Addressed both review findings on this head:

1. **Briefing auth** — no-re-ask is now scoped to the **managing human's explicit instruction for that specific create/update/pause/resume/delete**. Otherwise fail closed with `cron preview` + `chat ask` the managing human. Removes the prior “any current-chat human instruction” reading that could let a non-owner instruct lifecycle under the manager credentials.

2. **Disabled-mode recovery** — cron section and `FIRST_TREE_CRON_JOBS_ENABLED` env row now agree: when off / cadence outside 1..10, create and activate are rejected; **preview, list, show, pause, delete, and non-activating updates** remain available.

Local: Scheduled jobs briefing contract + Working ≤220 budget pass; Biome clean. No behavior/API change. Awaiting re-review; still not merging.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up precisely closes the authorization-scope and disabled-mode documentation findings without introducing a new static blocker.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: it also introduces persistent scheduling and ordinary-message materialization; no further lift exists above A.

**PR summary**
- Problem: Humans need to schedule repeated agent work in the current First Tree chat while preserving the ordinary visible addressed-message and inbox execution path.
- Approach: Persist owner-scoped jobs, expose preview/CRUD through HTTP, SDK, and CLI, and have a deployment-gated multi-replica sweeper materialize accepted occurrences as addressed messages.
- Impacted modules: shared DTO/config, Server schema/routes/services/notifier, Client SDK and briefing, CLI documentation/help, and cross-surface QA assets.

**Review findings**

✅ 1. The no-re-ask rule is now limited to the managing human's explicit instruction for the specific create/update/pause/resume/delete action; every other case fails closed and asks the managing human.

✅ 2. Both CLI reference locations now describe the same disabled/cadence-invalid recovery surface: preview, list, show, pause, delete, and non-activating updates remain available while create/activation are rejected.

✅ 3. The delta from `3302005258643e63361c10f9987eb70ef65466bc` is limited to the briefing, its contract assertion, and CLI documentation; it does not alter the API, authorization implementation, locks, or scheduler.

✅ 4. The paired durable owner-only contract remains linkable at [first-tree-context#808](https://github.com/agent-team-foundation/first-tree-context/pull/808).

**Verification evidence**
- Reviewed the complete delta `3302005258643e63361c10f9987eb70ef65466bc..a235e1810de0a5008c855471227878872c00ddf0` on exact head `a235e1810de0a5008c855471227878872c00ddf0`.
- Lint/typecheck, migrations, branch-name, portable-smoke, and completed analysis checks passed; Server, Client/Web, CLI, and JavaScript CodeQL checks were still running during review.
- No local build, typecheck, or tests were run as part of this review.
- Independent acceptance for this exact head remains required before merge.

**Action taken**
- Approved with explicit maintainer authorization; no merge performed.

PR #1963, opened 2026-07-23
docs: constrain skill eval scope

baixiaohang:
## Summary

- define a frozen acceptance scope for repo-local skill changes
- restrict routine skill-eval execution to code-only, no-model checks
- limit eval-case expansion and clarify blocking review criteria
- bound later review rounds to original findings and directly introduced regressions

## Validation

- `git diff --check`
- No skill evals run (documentation-only change; the new policy explicitly excludes model-backed evals from routine skill work)


yuezengwu (changes requested):
Blocking: this categorical ban conflicts with the repository's existing skill-eval contract. `packages/skill-evals/README.md` says suite or skill changes recommend that suite's floor/gate/quality coverage where implemented, and `packages/skill-evals/src/core/select.ts` emits those recommendations. The durable Context Tree contract likewise assigns the highest-risk skill behavior to the small deterministic live gates. After this change, an agent following `eval:select` is told to run commands that this root instruction forbids, and a high-risk behavior-contract change has no allowed behavioral validation path.

Please narrow this to the default/routine path while preserving an exception for the smallest selected existing gate/quality case when the frozen acceptance scope changes behavior covered by that gate (or update the selector, README, and durable eval contract together). The complete-suite and periodic prohibitions can remain.

baixiaohang:
Superseded by #1964. GitHub did not refresh this PR's head after the guidance was shortened; the replacement PR has one commit and a one-line AGENTS.md diff.

PR #1966, opened 2026-07-23
fix(client): retry Codex steer after new input

baixiaohang:
## Summary

- replace the turn-wide Codex `appendClosed` latch with a pending-input generation fence
- give the full FIFO pending prefix one fresh `turn/steer` attempt when newer input arrives
- preserve next-turn batching when the active turn remains non-steerable
- log definitive no-custody steer rejections without changing unknown-custody recovery

## Root cause

A single definitive `turn/steer` rejection marked the entire active turn as permanently closed to appends. On a long-running turn, every later human message stayed queued until the turn ended or the operator suspended it, even if Codex became steerable again.

The new generation fence blocks hot retry only for the exact pending-input generation that was rejected. A newer message changes the generation and re-opens one ordered attempt for the complete pending prefix. Inputs are still marked provider-entered only after `turn/steer` succeeds.

## Validation

- `pnpm exec biome check packages/client/src/handlers/codex/app-server/index.ts packages/client/src/__tests__/codex-app-server-handler.test.ts`
- `git diff --check`
  - 153 files passed, 1 skipped
  - 1628 tests passed, 3 skipped

Formal provider-backed QA remains useful for the real Codex compact-state transition; the adjacent `provider-transient-retry-after-visible-output` QA case covers the same steer/drain custody boundary.

### DIFF
--- a/packages/client/src/__tests__/codex-app-server-handler.test.ts
+++ b/packages/client/src/__tests__/codex-app-server-handler.test.ts
@@ -1387,6 +1387,9 @@ describe("codex app-server handler", () => {
     await waitFor(() => fake.requests.some((request) => request.method === "turn/start"));
     handler.inject(makeMessage("m2", "second"));
     await waitFor(() => fake.requests.some((request) => request.method === "turn/steer"));
+    await flushAsync();
+
+    expect(fake.requests.filter((request) => request.method === "turn/steer")).toHaveLength(1);
 
     completeTurn(fake, "turn-1", "first done");
     await waitFor(() => fake.requests.filter((request) => request.method === "turn/start").length === 2);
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
The change replaces the turn-wide append latch with a pending-input generation fence, so newer input gives the full ordered pending prefix one fresh steer attempt while the same rejected generation remains fenced.

I checked the steer/completion races and custody boundaries: no hot retry occurs for an unchanged generation, input that arrives during an in-flight rejection is retried as part of the full FIFO prefix, repeated definitive rejections remain queued for the next turn, and processing custody advances only after turn/steer succeeds. The added tests cover those paths.

No database changes are involved; the only state-shape change is the in-memory CurrentTurn fence. I did not run tests as part of this review.

PR #1984, opened 2026-07-23
Add manual Context Tree value audit skill

Gandy2025:
## Summary

- add the manually triggered, read-only `audit-context-tree-value` skill
- export only owned or explicitly authorized Chats, then audit each `Chat UUID @ Agent UUID` pair independently
- stream local First Tree Codex traces, map them through runtime-injected `chatId`, and pair bounded Tree passages with later visible Agent choices
- validate passage-level judgments and render deterministic evidence JSONL plus a human-readable Markdown report
- bundle and route the skill without intercepting ordinary task reads or stored-tree health audits
- add machine-readable UUID output to `agent list` so the exporter can verify exact local and remote Agent identity

## V0 boundaries

- manual invocation only
- local Codex provider traces only
- source records, traces, and Tree content remain read-only
- no schedules, server API, database, receipt, Context Tab, Tree write, or non-Codex adapter
- missing, cleaned, malformed, truncated, or unmapped traces remain explicit coverage gaps
- authorized Chats are not treated as an eligible value denominator, and read counts are never presented as value

## Verification

- `pnpm check`
- `pnpm typecheck` — 10/10 tasks
- audit deterministic floor — 10/10 tests
- CLI Agent lifecycle coverage — 28/28 tests
- runtime skill integration — 79/79 tests
- routing and adjacent audit floors — 32/32 tests
- repository skill validation — 9/9 skills
- official `skill-creator` validator — passed
- deterministic eval floor — 2/2 checks
- representative local sample generated `candidates.jsonl`, `evidence.jsonl`, and `REPORT.md`
- two independent final reviews found no remaining blockers

Model-backed skill evaluation was intentionally not run for this V0.


skills/audit-context-tree-value/scripts/audit_context_tree_value.py
```diff
@@ -0,0 +1,1576 @@
...
+    scoped_agent_ids = {chat["source_agent_id"] for chat in chat_rows}
+    configured_agent_ids = set(workspace_agents.values())
+    if configured_agent_ids != scoped_agent_ids:
+        missing = scoped_agent_ids - configured_agent_ids
+        extra = configured_agent_ids - scoped_agent_ids
+        raise AuditError(
+            f"--agent-workspace must match the exact scoped Agent UUIDs; missing={sorted(missing)}, extra={sorted(extra)}."
+        )
+    workspace_roots = set(workspace_agents)
+    tree_roots = tuple(sorted({Path(root).expanduser().resolve() for root in args.tree_root}))
```
yuezengwu, line 1048: **[P1] Bind each authorized Tree root to its Agent workspace.** `workspace_agents` is exact, but all `--tree-root` values are collapsed into one global tuple and passed to every trace. `relative_tree_path()` then returns only the relative path, dropping which Tree matched. With two scoped agents on different Trees, Agent A's trace can therefore accept a `.md` read under Agent B's Tree, and two `system/foo.md` nodes become indistinguishable in evidence/reporting. That violates the skill's per-workspace exact-bound-Tree contract and can produce a false passage→choice attribution. Please make the CLI carry an Agent/workspace→Tree binding (for example `AGENT_UUID=/tree/root`), validate it against the exact scoped Agent set, restrict path extraction to the reader's bound Tree, and preserve an unambiguous Tree identity in evidence. Add a two-Agent/two-Tree fixture with the same relative node path to prove no cross-attribution.

yuezengwu (changes requested):
## Review summary

This PR adds a manual, read-only Context Tree value-audit pipeline: authorized Chat export, local Codex trace-to-`chatId` reconstruction, passage/visible-choice evidence, deterministic judgment validation/reporting, bundled skill routing, and machine-readable Agent UUIDs from `agent list`. The Chat×Agent audit-unit separation and conservative verified/probable/unproven rubric are good foundations.

I found one blocker in the multi-Tree isolation path (inline). The collector needs an exact Agent/workspace → bound Tree mapping; a global Tree-root allow-list is insufficient and can misattribute same-relative-path evidence across agents. Please add a multi-Tree fixture with the same relative node path in both Trees to lock the isolation down.

Non-blocking process note: this changes both the shipped skill topology and the script-facing `agent list --json` contract. The current Context Tree still enumerates the previous skill family and command surface, so please create/cross-link the source-backed Tree update for `system/context-management/task-skill-design.md`, `system/architecture.md`, and `system/cli/command-surface/NODE.md`.

Per this review agent's scope, I inspected the diff and did not run additional local tests or QA. GitHub checks are currently green.

baixiaohang (changes requested):
- Rationale: The collector does not enforce exact-Chat authorization before reading local trace content, and the new agent-backed workflow has no executable behavioral evaluation.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: Operators currently cannot reconstruct whether Context Tree decisions materially affected past agent work.
- Approach: Add a manual skill that exports authorized Chat history, maps local Codex traces through runtime `chatId`, validates passage-to-choice judgments, and reports conservative verified/probable/unproven evidence.
- Impacted modules: CLI Agent listing, Client skill bundling/briefing, skill-eval registry, and the new audit skill/collector.

**Review findings**

❌ 1. Exact-Chat authorization is not enforced before trace-content access. [`skills/audit-context-tree-value/scripts/audit_context_tree_value.py:779`]

`authorized_trace_files()` accepts every recent root First Tree Codex trace whose `cwd` matches any scoped Agent workspace. `trace_reads()` then opens and JSON-decodes every row in each accepted file; the `chatId` check only controls which calls are retained afterward. For an `explicit` scope entry, the human authorized one Chat—not every other Chat whose local session trace lives in the same Agent workspace—so this crosses the skill's central data-minimization boundary even if unrelated rows are later discarded. Since the runtime session boundary is `(agent, chat)`, first map each trace to one exact authorized `Chat UUID @ Agent UUID` from a bounded header/current-context scan, and only then stream its full contents; malformed, ambiguous, or unmapped traces should remain coverage gaps.

❌ 2. The primary agent behavior is shipping without an executable live eval. [`packages/skill-evals/src/suites/audit-context-tree-value/cases.ts:28`]

The implemented floor validates the deterministic Python collector/reporter, while the gate that would exercise exclusive routing, authorization-scope construction, candidate-by-candidate judgment, and conservative evidence citation is marked `planned`; `runGate()` explicitly rejects this suite. Those are the model-controlled parts of the product, and the Context Tree QA contract requires shipped skill behavior to be evaluated rather than inferred from unit tests alone. Please implement and run the live gate, or commit and execute an observable manual checklist artifact that covers the same behavior before shipping V0.

⚠️ 3. I concur with the existing blocking review on Agent/workspace -> bound-Tree isolation; I did not duplicate its inline comment.

**Action taken**
- Submitted request changes.

yuezengwu:
Superseded by the lighter standalone approach:

- repository: https://github.com/agent-team-foundation/context-tree-insights
- merged PR: https://github.com/agent-team-foundation/context-tree-insights/pull/1

The shipped source is an explicit-only, non-installed Context Tree Insights Skill. It removes the First Tree core CLI/Client/runtime integration from this proposal and fixes the exact-Chat preflight, Agent/workspace/bound-Tree isolation, artifact-confinement, and read-isolation blockers identified here.

Closing this PR without merge to preserve its review history.

PR #1988, opened 2026-07-23
feat: guide GitLab System Hook setup and verify MR routing

yuezengwu:
## Summary

- guide admins from the configured origin to the exact `/admin/hooks` page and explain the one-time URL, Push/MR triggers, default payload, and SSL settings
- distinguish transport receipt from successful System Hook merge-request routing across Settings, Setup, and Context Tree readiness
- accept official System Hook merge-request and lifecycle payloads, preserve failure until a later valid merge-request event, and reset bearer-scoped evidence on regeneration

## Why

`/admin/hooks` emits `System Hook`. Treating any valid inbound event—such as Push, Test, or a lifecycle event—as healthy created a false green state even when no routable merge request had been processed.

## Scope

Whole-instance merge requests only. This does not add Issue follow, CLI/briefing flows, dual-webhook configuration, or outbound GitLab probing.

## Review follow-up

- consume an explicit observation outcome outside the best-effort callback, so an MR projection failure records actionable failure state in the ingress transaction instead of retaining stale readiness
- project readiness once on the Server and have Settings, Setup, and Context Tree consume that canonical field; no new readiness database column or query was added
- merge current `main` without rewriting history; preserve the new Setup status-icon model while continuing to derive GitLab waiting/failed stages only from the Server readiness projection

## Validation

- server: 2647/2647 tests passed
- web after conflict resolution: 1679/1679 tests passed
- shared: 697/697 tests passed
- focused review regression: server 43/43; focused Web surfaces after conflict resolution 49/49; shared contract 3/3
- `pnpm check`
- `pnpm typecheck`
- migration consistency check
- real-browser desktop/mobile validation covering Push → failure → valid MR transitions and secret disappearance after close/refresh

## Known constraint

Push and merge-request events share the existing 120/minute connection-level pre-parse rate limit; this PR intentionally leaves that policy unchanged.

## Context Tree

- Draft decision PR: https://github.com/agent-team-foundation/first-tree-context/pull/815

Closes #1980

PR #1987, opened 2026-07-23
feat(web): integrate Setup capabilities

yuezengwu:
## Summary

- consume the role-independent `TeamSetupCapabilities` projection in permanent Settings → Setup with runtime schema and Team-scope validation
- render Repository Automation, Context Tree build/recovery, and nested Automatic Review status from Server facts while keeping personal work access on `/me` + clients
- interpret the same Team facts by role: admin-owned blockers get actionable attention, members get read-only explanations, and optional/operator-owned gaps stay neutral
- add semantic status icons and desktop Setup navigation attention, route legacy `/settings/context` to Setup, and keep every mutation on its owning surface with live rechecks

## Boundaries

- no setup mutation aggregator, persisted completion bit, percentage, generic DAG, or synthetic ruleset health
- onboarding remains value-first; GitHub App, Context Tree, and Automatic Review do not gate Start Chat
- consumes the merged #1988 readiness projection; the compatibility pass only adds the missing Admin owner action for `gitlab_processing_failed`, without changing its schema, inbound processing, or owner UI
- does not implement Reviewer select/enable or Server provisioning/settings contracts formerly proposed by #1982

## Delivery phase

This is the independently mergeable projection/rendering phase. It is synced
to `main` after #1988 and consumes its canonical System Hook / merge-request
routing facts. It does not depend on the replacement Reviewer assignment
contract.

The existing Context Tree and Reviewer controls intentionally remain in
Repositories during this transition so there is no window with no executable
entry. A later Web migration will land after the separate Reviewer contract;
it will select an eligible existing Agent (never auto-create one in Setup),
add final owner controls, and remove duplicate Repositories controls in the
same change.

## Testing

- `pnpm --filter @first-tree/web exec vitest run src/pages/__tests__/preview-pages-extra.test.tsx src/__tests__/app-routes.test.tsx src/pages/settings/__tests__/setup-dom.test.tsx src/pages/settings/__tests__/setup-attention.test.ts src/pages/__tests__/settings-panels-dom.test.tsx src/api/__tests__/setup-capabilities.test.ts` (86/86)
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/setup-capabilities.test.ts` (14/14)
- `pnpm --filter @first-tree/shared exec vitest run src/__tests__/setup-capabilities.test.ts` (3/3)
- latest local full Web attempt reached 199/200 files (1701/1702 tests); the sole failure is the unrelated existing `mobile card behavior > opens the question over Now...` test and reproduces in isolation
[...]

Issue #1977, opened 2026-07-23
feat: create and configure a Context Tree Review Agent after onboarding


baixiaohang:
## Problem

After onboarding creates and binds an organization's Context Tree, an admin still needs to create a dedicated Context Tree Review Agent and configure it for the organization manually.

## Desired behavior

After onboarding successfully creates the Context Tree:

- automatically create a dedicated Context Tree Review Agent;
- give it a minimal prompt that directs Context Tree PR/MR review work to the `context-tree-review` skill; and
- configure that Agent as the organization's Context Reviewer.

The flow should be idempotent and must not create duplicate Reviewer Agents when the organization already has one configured.

yuezengwu:
Superseded by the confirmed assignment model in #1992: an admin selects an existing eligible organization-visible managed Agent. Setup does not automatically create or provision a dedicated Reviewer Agent.

PR #1992, opened 2026-07-23
feat: add Context Reviewer assignment readiness contract

yuezengwu:
## Summary

- add admin-only candidate, assignment, and enablement endpoints for Automatic Review
- keep the existing `{ enabled, agentUuid }` storage shape, preserve a selection while disabled, and introduce no migration
- require an existing Team-managed, active, organization-visible Agent with an active manager, a non-retired pinned Computer, and a supported runtime
- project selected-but-disabled and live degraded health through `TeamSetupCapabilities`
- recheck exact Agent, runtime route, Tree binding, and provider authority in the transaction that creates each trusted GitHub/GitLab review run
- map the new typed facts/actions into #1987's existing exhaustive Setup copy/routing tables without adding Reviewer controls

This separates durable assignment from enablement and transient runtime health. Offline or stale Computers remain selected/enabled and recover without a settings rewrite; each run still fails closed. Assignment or enablement failure never mutates or rolls back the independent Context Tree binding.

This supersedes the automatic-provisioning approach proposed in #1977. This PR intentionally selects an existing eligible managed Agent and does not create one.

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm --filter @first-tree/shared test` — 63 files / 699 tests
- [x] `pnpm --filter @first-tree/server test` — 245 files / 2670 tests
- [x] `pnpm --filter @first-tree/web test` — 201 files / 1708 tests
- [x] focused assignment/readiness, GitHub/GitLab dispatch, projection, privacy, recovery, concurrency, and idempotency coverage
- [x] `git diff --check`
- [x] two independent review passes; all definite blockers resolved

## Change Surface

Server/shared contract plus an 11-line compatibility mapping in the existing Setup page. No Agent provisioning, visibility mutation, Reviewer consent, Setup query/render/control changes, final Reviewer UI, GitLab System Hook schema/migration, or generic workflow/DAG was added.

## Notes

- docs or tests updated to match: Shared, Server, and merge-state Web contract/regression tests
- follow-up work: the final Setup selector/enable controls stay in a separate Web task
- dependency integrated: based on merged #1988 and reuses its authoritative GitLab System Hook merge-request routing evidence
- Context Tree: paired Draft PR https://github.com/agent-team-foundation/first-tree-context/pull/816; merge this code PR first, then reconcile the Tree PR

### DIFF
--- a/packages/server/src/services/org-settings.ts
+++ b/packages/server/src/services/org-settings.ts
@@ -216,9 +216,11 @@ export async function getOrgContextTreeBinding(db: Database, orgId: string): Pro
 }
 
 /**
- * Read the live Context Tree binding and Reviewer assignment in one database
- * statement. Review mutations use this tuple so they never combine a binding
- * from one settings snapshot with an assignment from another.
+ * Read the live Context Tree binding and raw Reviewer assignment in one
+ * database statement. Server-internal review mutations use this tuple so they
+ * never combine a binding from one settings snapshot with an assignment from
+ * another. Member- or Agent-readable surfaces must use the Team-safe
+ * projection below.
  */
 export async function getOrgContextReviewRuntime(db: Database, orgId: string): Promise<OrgContextReviewRuntime> {
   const rows = await db
@@ -296,6 +298,26 @@ export async function getOrgContextReviewRuntime(db: Database, orgId: string): P
   };
 }
 
+/**
+ * Apply the same Team-safe Reviewer identity projection as `getOrgSetting`.
+ * Invalid historical private, deleted, or foreign selections remain stored
+ * for admin replacement without exposing their opaque UUID to Team runtimes.
+ */
+export async function getTeamSafeOrgContextReviewRuntime(
+  db: Database,
+  orgId: string,
+): Promise<OrgContextReviewRuntime> {
+  const runtime = await getOrgContextReviewRuntime(db, orgId);
+  const reviewerAgent = await resolveContextReviewerAgentSummary(db, orgId, runtime.contextReviewer.agentUuid);
+  return {
+    ...runtime,
+    contextReviewer: {
+      enabled: runtime.contextReviewer.enabled,
+      agentUuid: reviewerAgent?.uuid ?? null,
+    },
+  };
+}
+
 export type OrgContextReviewRuntime = {
   bindingState: "bound" | "unbound" | "invalid";
   provider: ContextTreeProvider | null;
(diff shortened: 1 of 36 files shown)


yuezengwu:
@baixiaohang 麻烦帮忙 review 一下这个 PR。当前 head `40f1bf100`，全部 CI 已绿；assignment/readiness、权限、并发与 GitHub/GitLab per-run authority 已完成两轮独立 review 和回归验证。

如果还有 blocker，请直接留 review thread，我会立即跟进。谢谢。

packages/server/src/services/org-settings.ts
```diff
@@ -315,6 +324,8 @@ export async function isOrgContextReviewRuntimeCurrent(
     current.repo === expected.repo &&
     current.branch === expected.branch &&
     current.providerMatchesRepository === expected.providerMatchesRepository &&
+    current.contextReviewer.enabled === expected.contextReviewer.enabled &&
```
baixiaohang: Blocking: this helper is also the `gitlabExecutionGuard` for all three Context Tree snapshot routes. Adding Reviewer fields means a concurrent assignment/disable now returns `false` even when the binding and GitLab connection are unchanged; `getContextTreeSnapshot` maps that to `invalid_binding` and refuses the independent Tree read/publication. Please split binding/provider currentness from Reviewer-setting currentness (or use a Reviewer-specific fence only in enablement) so Reviewer mutations cannot invalidate Context Tree snapshots.

packages/server/src/services/org-settings.ts
```diff
@@ -167,11 +172,15 @@ async function toOutput<K extends OrgSettingNamespace>(
...
+    const reviewerAgent = await resolveContextReviewerAgentSummary(db, orgId, s.contextReviewer.agentUuid);
     const out: OrgSettingOutput<"context_tree_features"> = {
       contextReviewer: {
         enabled: s.contextReviewer.enabled,
-        agentUuid: s.contextReviewer.agentUuid,
-        reviewerAgent: await resolveContextReviewerAgentSummary(db, orgId, s.contextReviewer.agentUuid),
+        // This namespace is member-readable. A private, deleted, or foreign
+        // historical selection stays in storage for replacement, but its
+        // opaque identity is not a Team-wide fact.
+        agentUuid: reviewerAgent?.uuid ?? null,
```
baixiaohang, line 182: Blocking privacy/blast-radius gap: this member-facing projection now correctly hides a historical private Reviewer UUID, but `packages/server/src/api/agent/context-tree-info.ts:26` still returns `getOrgContextReviewRuntime(...).contextReviewer` verbatim. Any authenticated runtime Agent in the Team can call that Class D endpoint through the SDK and recover the private Agent UUID. Please reuse a safe projection there (or expose Reviewer identity only when the caller is authorized to receive it) and cover the historical private-selection case across both surfaces.

baixiaohang (dismissed):
- Rationale: Two cross-consumer regressions break the intended independence and privacy boundaries.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the semantics of the persisted `context_tree_features` settings JSON.

**PR summary**
- Problem: Team admins need to assign an existing managed Agent for automatic Context Tree reviews without coupling durable selection to temporary runtime health.
- Approach: Separate assignment from enablement, preserve disabled selections, project structural versus transient readiness, and recheck reviewer/provider authority before creating or publishing trusted review runs.
- Impacted modules: shared setup contracts; organization settings and Setup capability projection; GitHub/GitLab review dispatch; GitHub review publication; Setup copy/routing compatibility.

**Review findings**
❌ 1. Keep Reviewer state out of the GitLab snapshot binding/connection guard; the new comparisons make an unrelated assignment or enablement change fail an in-flight Context Tree read as `invalid_binding`. [R5 / `packages/server/src/services/org-settings.ts:327`]
❌ 2. Apply the new private-selection masking to the Agent runtime config surface too; `/api/v1/agent/context-tree/info` still returns the raw historical private Reviewer UUID to other Team runtimes. [R5 / `packages/server/src/services/org-settings.ts:182`, `packages/server/src/api/agent/context-tree-info.ts:26`]
✅ 3. The core assignment/enablement split and per-run Agent/provider fences otherwise match the paired durable contract.

**Action taken**
- Submitted request changes on head `40f1bf100ee5f1fac07fb62ab52b5246ce62f2b9`.

packages/server/src/services/org-settings.ts
yuezengwu: Fixed in `af3200a76`: the three GitLab snapshot/read guards now use `isOrgContextTreeBindingRuntimeCurrent`, which fences only the Tree binding, provider route, and GitLab connection. The Reviewer-aware guard remains on the enablement transaction. Added a regression proving Reviewer state changes keep the Tree snapshot guard current while invalidating the full review mutation guard.

packages/server/src/services/org-settings.ts
yuezengwu: Fixed in `af3200a76`: `/api/v1/agent/context-tree/info` now uses the Team-safe Reviewer projection, sharing the organization-visible/non-deleted masking used by settings. Added a regression with a different runtime Agent and a historical private enabled Reviewer; both settings and runtime surfaces preserve `enabled` but return `agentUuid: null` and never include the private UUID.

yuezengwu:
@baixiaohang 麻烦复审当前 head `af3200a76`。您提出的两项 blocker 已逐 thread 修复并附上回归证据，新 head 全部 CI 已绿；如果没有新增 blocker，烦请 approve，便于继续合并。谢谢。

PR #1993, opened 2026-07-23
fix(web): prevent native selection on mobile long press

yuezengwu:
## Summary

- apply standard and WebKit text-selection suppression to the actual mobile Now and Chat card surfaces
- clear any native selection that started before opening the custom long-press action sheet
- cover both card variants and selection cleanup with DOM regression assertions

## Product scope

Mobile scope: allowlisted — this fixes the existing Now/Chat long-press triage interaction and adds no new mobile capability.

## Verification

- `pnpm check` (passes; existing repository warnings only)
- `pnpm typecheck`
- `pnpm --filter @first-tree/web test` — 201 files, 1,708 tests passed
- focused mobile card suite — 10 tests passed
- `pnpm test` was also attempted; product tests, including the mobile suite, passed, but the aggregate run was interrupted by four unrelated 5-second timeouts in Context Tree skill-eval fixtures under concurrent load

## Manual acceptance

Please verify on both Android Chrome and iPhone Safari:

- long press opens only the First Tree action sheet, without native selection/copy/lookup UI
- vertical movement or scrolling cancels the pending long press
- completing a long press does not also open the chat

### DIFF
--- a/packages/web/src/pages/mobile/__tests__/card-actions-dom.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/card-actions-dom.test.tsx
@@ -253,6 +253,9 @@ describe("mobile card behavior", () => {
     expect(card?.querySelector("[data-mobile-card-menu]")).toBeNull();
     expect(harness.container.querySelector("[data-mobile-swipe-surface]")).toBeNull();
     expect(document.body.querySelector('[role="menu"]')).toBeNull();
+    expect(card?.style.userSelect).toBe("none");
+    expect(Reflect.get(card?.style ?? {}, "WebkitUserSelect")).toBe("none");
+    expect(Reflect.get(card?.style ?? {}, "WebkitTouchCallout")).toBe("none");
     expect(card?.querySelector('[aria-haspopup="dialog"]')?.getAttribute("aria-description")).toBe(
       "Long press for chat actions",
     );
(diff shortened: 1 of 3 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change prevents native selection/callout UI from competing with the existing mobile long-press action sheet without changing the gesture or widening mobile scope.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**` -> C
- Semantic lift: none; this does not touch subscriptions, WebSocket frame parsing, or the data layer

**PR summary**
- Problem: On mobile Now and Chat cards, browser-native selection/copy/lookup UI can appear during the same long press that opens First Tree's triage sheet.
- Approach: Apply standard and WebKit selection suppression to the actual card surfaces, and clear any already-started DOM selection before every action-sheet entry path.
- Impacted modules: mobile Now cards, the shared long-press hook, and mobile card DOM regression coverage

**Review findings**
✅ 1. The shared surface style covers both DOM variants: Now's visible card wrapper and Chat's card button.
✅ 2. Selection cleanup is centralized before the timer, context-menu, and keyboard paths call the action-sheet callback, while existing click suppression and scroll/movement cancellation remain intact.
✅ 3. This is a repair to the already-authorized Now/Chat long-press triage capability, not a new mobile capability.

**Action taken**
- Approved

PR #2000, opened 2026-07-24
fix(cron): remove deployment availability gates

yuezengwu:
## Summary

- remove the `FIRST_TREE_CRON_JOBS_ENABLED` deployment kill switch
- start the cron worker with normal Server background tasks
- remove per-mutation cron availability checks
- validate the scheduler polling cadence once at Server startup
- remove the retired availability error codes and update CLI docs plus the live QA case

## Redundancy audit

The cron feature had two independent deployment-time availability gates:

1. an opt-in feature flag that defaulted off
2. a polling interval range check that independently disabled the worker and mutations

Both availability gates are removed. `FIRST_TREE_POLLING_INTERVAL_SECONDS`
remains a defaulted runtime tuning parameter, validated at Server startup in
the dispatch-safe `1..10` second range. It no longer changes whether
create/resume is accepted.

Authorization, owner scoping, concurrency barriers, revision checks, trigger
de-duplication, dispatch grace, and outstanding-ACK protection remain because
they enforce correctness rather than feature availability.

## Review follow-up

The always-on scheduler initially accepted polling cadences longer than its
30-second dispatch grace. A 60-second cadence could therefore skip every
minute-aligned occurrence while leaving the job active. The follow-up commit:

- restores a shared `1..10` second safety bound as startup validation
- preserves at least three poll opportunities inside the dispatch grace
- adds a deterministic config regression covering defaults, both bounds, and
  rejected `0`, `11`, and `60` second values
- keeps the validation out of create/resume and other mutation paths

## Verification

- shared package tests — 63 files, 700 tests passed
- changed TypeScript files pass Biome
- cron Server integration — 34 passed
- Server background + cron schedule — 12 passed
- shared cron schema — 12 passed
- CLI cron commands — 4 passed

`pnpm test` was also run. Cron and product suites reached passing results, but
the monorepo run exited non-zero because four unrelated Context Tree skill
fixture tests exceeded their existing 5-second timeout.

The full `pnpm check` currently reports one unrelated pre-existing formatter
error in `packages/client/src/__tests__/assistant-text.test.ts`; the files
changed by this follow-up pass Biome.

## QA

Formal cross-surface QA is warranted after staging deploy because this changes
Server boot/background-task behavior and the create → trigger → Inbox path.

## Context Tree

- Draft update: agent-team-foundation/first-tree-context#819


baixiaohang (dismissed):
- Rationale: The always-on worker now accepts polling cadences that can make every due occurrence late and silently skipped.

**Risk level**: A
- Semantic lift: changes scheduled-job Server background execution and the public CLI/error contract; remains A

**PR summary**
- Problem: deployments currently need both a feature flag and a narrow polling-cadence gate before scheduled jobs can be activated, so an otherwise configured schedule may be unavailable.
- Approach: remove both availability checks, start the scheduler with ordinary Server background tasks, retire the two availability error codes, and reconcile CLI docs, QA guidance, and the paired Context Tree draft.
- Impacted modules: `apps/cli`, `packages/server`, `packages/shared`, CLI reference, and cross-surface QA case

**Review findings**
❌ 1. Keep the polling cadence compatible with the 30-second dispatch grace. `FIRST_TREE_POLLING_INTERVAL_SECONDS` still accepts any positive integer, and the worker now uses that value directly, while `claimAndProcessOne` skips an occurrence once it is more than `CRON_DISPATCH_GRACE_MS` (30 seconds) late. For example, a 60-second poll started at `:40` will see every minute-aligned occurrence at about 40 seconds late, skip it, advance `nextRunAt`, and leave the job active; create/resume succeeds but the job can silently never trigger. The old `1..10` gate prevented this state. Please enforce a safe boot-time cadence bound (without reintroducing a per-mutation availability gate), derive compatible grace/scheduling behavior from cadence, or otherwise guarantee that a valid tuning value cannot systematically miss occurrences, with a deterministic regression test. [`packages/shared/src/config/server-config.ts:517`, `packages/server/src/services/cron-scheduler.ts:308`, `packages/server/src/services/cron-scheduler.ts:478`]

✅ The removal is otherwise reconciled across Server callers, shared error codes, CLI handling/reference, QA prerequisites, and the paired Context Tree draft.

**Action taken**
- Submitted request changes.

PR #2003, opened 2026-07-24
fix: keep errored chats in attention while offline

Gandy2025:
## Summary

- derive `failedAgentIds` from the per-chat `errored` axis instead of the reachability-gated composite display status
- keep an errored managed-agent chat in Attention across disconnect and reconnect, then remove it after genuine recovery
- add a deterministic server regression and a live cross-surface QA case for the visible list-stability path

## Validation

- [x] `pnpm check` (passes with 16 existing warnings and 1 existing info)
- [x] `pnpm typecheck`
- [x] `pnpm --filter @first-tree/server test` (245 files, 2,670 tests)
- [x] `pnpm exec vitest run src/__tests__/me-chat-priority.test.ts` from `packages/server` (16 tests)
- [ ] `pnpm test` (the changed server suite passed, but the monorepo run hit unrelated timeout-only failures in heavy `skill-evals` fixtures under parallel load; 30/32 affected isolated cases passed on retry, with the remaining 2 also timing out)

## Change Surface

- [ ] apps/cli public CLI or help output

## Notes

- docs or tests updated to match: updated the public schema comment and added server regression coverage plus `failed-chat-attention-stability` live QA coverage
- follow-up work: run the new live QA case against an isolated Client/Server/Web cell before release; no product-code follow-up is expected

### DIFF
--- a/packages/qa/cases/cross-surface/failed-chat-attention-stability.md
+++ b/packages/qa/cases/cross-surface/failed-chat-attention-stability.md
@@ -1,0 +1,67 @@
+---
+id: failed-chat-attention-stability
+description: Validate that an errored managed-agent chat stays in Needs attention across disconnects and leaves only after recovery.
+areas: [cross-surface]
+surfaces: [client, server, web]
+---
+
+# Failed chat attention stability
+
+## Goal
+
+Confirm the real runtime-to-workspace path keeps a chat in `Needs attention`
+while its caller-managed speaker remains errored, even if that agent disconnects
+and its display status changes between failed and offline. The row should enter
+the group once and leave once after genuine recovery, without repeatedly moving
+between list groups.
+
+The deterministic server test owns the projection rule. This case owns the live
+Client session frames, presence changes, Web refetches, and visible conversation
+rail behavior that the product test cannot prove.
+
+## Preconditions
+
+- Run the candidate server, Web, and Client in an isolated QA cell with a real
+  PostgreSQL database and WebSocket connections.
+- Sign in as the manager of a non-human agent and create a chat where that agent
+  is a speaker.
+- Keep browser network, WebSocket, and a short screen recording available. Do
+  not use an unrelated peer's failed agent because Attention is manager-scoped.
+
+## Operate and observe
+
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed the status projection, attention candidate/enrichment path, shared schema contract, deterministic regression, and live QA case.

The change correctly derives failedAgentIds from the orthogonal errored axis rather than the lossy main display status, so reachability transitions no longer remove an errored caller-managed speaker chat from Attention. Recovery still clears through the shared canonical per-chat status producer, and the existing manager/speaker scoping and priority ordering remain intact.

No blocking findings. There are no database or core data-structure changes. Per reviewer instructions, I reviewed the diff and existing coverage without running tests or QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change preserves recovery attention across reachability transitions by consuming the canonical per-chat `errored` axis, without changing ownership scope, priority ordering, or persistence.

**Risk level**: B-low
- Path baseline: touches `packages/server/**`, but not DB structure or the high-risk core service set.
- Semantic lift: none; the shared-schema change is contract documentation, and no Web subscription/data-layer behavior changes.

**PR summary**
- Problem: A caller-managed agent chat could leave and re-enter Needs attention when an errored agent disconnected and reconnected because the list used the reachability-gated display status.
- Approach: Derive `failedAgentIds` from the orthogonal `errored` axis produced by the existing shared per-chat status resolver. Add deterministic server coverage for offline persistence and recovery, plus a live cross-surface QA case for visible rail stability.
- Impacted modules: server chat-list projection and regression test, shared response-schema documentation, cross-surface QA case.

**Review findings**
✅ The attention signal continues to use the shared per-chat status producer and retains the existing managed-agent and speaker filters.
✅ The candidate query already includes stored session/runtime error inputs without reachability gating, so the enrichment change closes the false-negative without widening healthy chats into Attention.
✅ Recovery still clears through the canonical status axes, and the live QA case covers the Client/Server/Web behavior that the deterministic projection test cannot observe.

**Action taken**
- Approved.

PR #1997, opened 2026-07-23
Unify the mobile Work surface

yuezengwu:
## What

- replace the overlapping mobile Now and Chat tabs with one canonical Work surface
- default Work to the complete Active set, ordered by Attention → Pinned → Recency
- add quick views for Need you, Unread, and Pinned plus a filter sheet for Active, Archived, All, and Watching
- introduce distinct action cards and regular work rows with a fixed two-line content budget
- render chat summaries as an inline Current state card at the top of the mobile conversation timeline
- preserve `/m/now` and `/m/chat` as compatibility redirects to `/m/work`
- rebuild the mobile preview so the complete list, detail, and answer-sheet experience is reviewable without seeded data

## Why

Now and Chat exposed overlapping work, which forced mobile users to switch tabs to find unread, pinned, or actionable conversations. The unified Work surface keeps one stable mental model while preserving task urgency, user curation, and recency.

## Interaction contract

- default scope: Active
- ordering: managed-agent failure, open request, pinned, then true work activity recency
- unread and working states change presentation without reordering the list
- normal and pinned rows show up to two summary lines
- unread and working rows show one summary line plus one line of dynamic evidence
- failure and request cards use both lines for actionable evidence
- summaries fall back to the latest meaningful message when absent
- long press retains triage actions; visible Answer and Review actions shorten urgent flows

## Validation

- `pnpm --filter @first-tree/web typecheck`
- design-token guardrails
- `pnpm --filter @first-tree/web build`
- full web suite: 204 files, 1,735 tests passed
- real-browser UI QA at 320×693, 390×844, and 430×932
- verified no horizontal page overflow, no console warnings/errors, fixed filter access at 320 px, 44 px minimum visible touch targets, and non-sticky/non-nested Current state expansion

## Review

Open `/preview/mobile` to review the deterministic mobile list, conversation detail, and Answer sheet.

The paired Context Tree contract PR will remain draft until this code PR merges.

PR #2017, opened 2026-07-25
fix(web): restore chat-open scroll landing on mobile

yuezengwu:
## Problem

On mobile web, opening a chat lands at the very top of the timeline instead of the last-read position / latest message. If new messages arrived, the "↓ N new messages" pill offers a way down; without new messages the user has to scroll manually through the whole timeline.

## Root cause

#1997 (Unify the mobile Work surface) added a branch to the chat-open landing effect in `chat-view.tsx`: when `presentation="mobile"` and the chat detail has a non-empty `description`, the timeline scroller is forced to `scrollTop = 0` so the in-flow Current state card is visible on entry.

Since #1997 also routes every mobile chat through `presentation="mobile"` (`/m/work`), and agent-owned chats almost always carry a summary (`chat update --description`), effectively every mobile chat open landed at the top.

## Fix

Restore the pre-#1997 landing behavior for all presentations (`packages/web/src/pages/workspace/center/chat-view.tsx`):

- stored bottom-visible anchor → land it at the viewport bottom; newer messages surface via the "↓ N new messages" pill
- no anchor (first visit) → land at the timeline bottom

The `MobileCurrentStateCard` itself is unchanged: it still renders in-flow at the top of the timeline with its "Updated" badge, reachable by scrolling up.

## Tests

- New regression test in `chat-view-dom.test.tsx`: a mobile chat **with** a summary still lands on the stored bottom-visible anchor (fails on the pre-fix code, passes after)
- `chat-view-dom.test.tsx` (52) + all `pages/mobile` tests: 129 passed
- `tsc --noEmit` and `biome check` clean

## Review round 1 (request-changes addressed in `adb318d1`)

- **R4 — landing committed before the persisted read state resolved**: the landing effect now also waits for the `["chat-read-state", chatId]` query to resolve before committing the one-shot guard; a resolved-but-empty row keeps the first-visit bottom fallback. Covered by two new tests with messages already rendered and a deferred IDB lookup: stored-anchor resolution lands on the anchor (verified to fail without the fix), and the no-row path lands at the bottom only after resolution.
- **Design-token guard**: `#1997` comment references reworded to `PR 1997` (the guard reads `#` + 4 hex digits as a color literal). `lint:tokens` is clean.
- **ComposeStatusBar note**: not reproduced — full Web suite on this head passes 1738/1738 tests (only `tests/visual-regression.spec.ts` fails collection with "No test suite found", pre-existing and untouched by this PR).

### DIFF
--- a/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
+++ b/packages/web/src/pages/workspace/center/__tests__/chat-view-dom.test.tsx
@@ -976,7 +976,7 @@ describe("ChatView", () => {
   });
 
   it("lands on the stored bottom-visible anchor on mobile open even when the chat has a summary", async () => {
-    // Regression guard for the #1997 mobile Work surface: a summarized
+    // Regression guard for the PR 1997 mobile Work surface: a summarized
     // chat must still land on the stored bottom-visible anchor (newer
     // messages surface via the pill) instead of jumping to the timeline
     // top to show the in-flow Current state card.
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
- Rationale: The mobile-specific top landing is removed, but the one-shot landing effect can still commit the bottom fallback before the persisted read anchor has loaded, so the stated restore behavior is not reliable on a cold open.

**Risk level**: C
- Semantic lift: none; this changes DOM scroll landing without touching subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Opening a summarized chat on mobile currently strands the user at the beginning of the timeline instead of resuming at the saved reading position or latest content.
- Approach: Remove the summary-dependent mobile top-scroll branch and reuse the existing stored-anchor / first-visit-bottom behavior for every presentation; keep Current state as the first in-flow card.
- Impacted modules: Workspace chat-open landing effect and its mobile DOM regression coverage.

**Review findings**
❌ 1. The landing guard can be committed before the persisted read state resolves. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:3055`]
`readState` is `undefined` while `getReadState(chatId)` is in flight, but this effect waits only for `itemCount`. Cached messages or the server response can therefore populate the timeline first: `bottomVisibleResolution` is still `null`, `landedForChatRef` is set, and the view lands at the bottom. When the stored anchor arrives, the one-shot guard prevents re-landing, so a cold mobile open can still miss the saved position this PR promises to restore. The new test pre-seeds `["chat-read-state", chatId]` through `seedChat`, so it cannot exercise that ordering. Please defer committing the landing until the read-state query has resolved (while preserving the `null` no-row fallback), and cover delayed read-state resolution with messages already rendered.

**Action taken**
- Submitted request changes on exact head `f73136e20484ee72f39961592e9c77ef3a1c01cd` after static review; no local checks or tests were run.
- Non-review CI note: the design-token guard currently treats the two new literal `#1997` comment references as hardcoded colors, and the full Web run also reports one failure in an unchanged ComposeStatusBar test.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: `adb318d1` closes the cold-open race by keeping the one-shot landing uncommitted until persisted read state resolves, while preserving the resolved-empty first-visit fallback and the mobile Current state card's in-flow placement.

**Risk level**: C
- Semantic lift: none; the change remains scoped to DOM scroll landing and focused regression coverage.

**PR summary**
- Problem: Opening a summarized chat on mobile can strand the user at the beginning of the timeline instead of resuming at the saved reading position or latest content.
- Approach: Remove the summary-dependent mobile top-scroll branch, reuse the canonical stored-anchor / first-visit-bottom behavior for every presentation, and wait for IndexedDB read-state resolution before committing that one-shot landing.
- Impacted modules: Workspace chat-open landing effect and mobile DOM regression coverage.

**Review findings**
✅ 1. The previous blocker is resolved: `readStateLoading` now gates the one-shot guard, so cached messages cannot make the bottom fallback win while the persisted anchor is still in flight.
✅ 2. The follow-up covers both asynchronous outcomes with messages already rendered: a delayed stored row lands on its anchor, while a delayed `null` result lands at the timeline bottom only after resolution.
✅ 3. The original summarized-mobile regression remains covered, and the final behavior matches the durable product boundary: Current state stays first and in-flow while the frozen chat-open read anchor governs resume position.
✅ 4. The issue-number comments no longer trip the token guard; CI lint/type-check is green on this head.

**Action taken**
- Approved exact head `adb318d17a511f8af456fe94ac7e2fdbe47bb718` after static review; no local checks or tests were run. GitHub's Client/Web and CLI jobs were still in progress at submission time.

PR #2032, opened 2026-07-28
Fix anonymous GitLab snapshots with Git 2.43

baixiaohang:
## Summary

- remove empty `http.sslCert` and `http.sslKey` command-line config values from anonymous GitLab snapshot operations because Git 2.43/libcurl can interpret an empty certificate value as a certificate file path and fail before egress
- preserve the anonymous-only boundary through the existing proxy, credential, `GIT_SSL_*`, HOME, global/system config, and cached-repository isolation
- add regression coverage proving ambient client certificate variables are stripped and the incompatible empty Git config values do not return

## Validation

- `pnpm --filter @first-tree/server exec vitest run src/__tests__/context-tree-snapshot-git-mocked.test.ts --maxWorkers=2` (18/18)
- `pnpm check` (passes with existing warnings)
- `pnpm typecheck`
- targeted Biome check and `git diff --check`
- real anonymous `git ls-remote` against the reported Self-Managed GitLab repository: reproduced the X.509 failure with `http.sslCert=` and successfully read `refs/heads/main` after removing the empty client-certificate config

A full Server test run was also attempted. The affected suite passed, while one unrelated cron time assertion failed in `cron-jobs.integration.test.ts` (expected an earlier timestamp and received the next scheduled hour); 244 files and 2,676 tests passed.

## QA

Formal cross-surface QA is warranted because this changes the GitLab provider/auth egress path. The existing `packages/qa/cases/cross-surface/gitlab-context-tree-lifecycle.md` case covers anonymous public Web Context, private/unauthorized degradation, and ambient credential/config isolation. Formal QA was not run automatically.

### DIFF
--- a/packages/server/src/__tests__/context-tree-snapshot-git-mocked.test.ts
+++ b/packages/server/src/__tests__/context-tree-snapshot-git-mocked.test.ts
@@ -38,6 +38,8 @@ describe("Context Tree snapshot service with mocked git", () => {
   it("materializes public GitLab content anonymously without credential injection", async () => {
     const repo = `https://gitlab.example/public/context-${crypto.randomUUID()}`;
     const head = "a".repeat(40);
+    const previousSslCert = process.env.GIT_SSL_CERT;
+    const previousSslKey = process.env.GIT_SSL_KEY;
     const { execFile, service } = await loadSnapshotServiceWithGit((args) => {
       if (args.includes("clone")) {
         const root = args.at(-1);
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the anonymous GitLab snapshot fix at b02ca1d18. Removing the empty http.sslCert/http.sslKey command-line values addresses the Git 2.43/libcurl failure without weakening the anonymous-only boundary: GIT_SSL_* is still stripped, HOME and global/system config remain isolated, credentials/proxies remain disabled, and cached repository config is rejected if it can restore client certificates or alternate egress. The regression assertions cover both sides of that contract. No blocking findings.

PR #2035, opened 2026-07-28
fix: reduce Mobile Work list startup cost

Gandy2025:
## Summary

- hydrate the first-page attention, pinned, and ordinary chat slices once while preserving the additive `/me/chats` response contract
- reuse participant speaker membership during list status resolution and skip the detail-only provider retry reason query
- keep Mobile Work card content and chat-summary precedence unchanged, but mount only 16 buffered cards initially and lazy-load uploaded avatars
- retain explicit network pagination so search can recover matches on later pages without automatic over-fetching

## Performance budgets

- a first page with one chat overlapping attention, pinned, and ordinary slices now uses exactly 8 database queries in the deterministic regression fixture
- Mobile Work initially mounts 16 cards, then reveals already-buffered cards in batches near the list boundary

## Validation

- `pnpm check` (passes; repository-existing warnings only)
- full monorepo typecheck: 10/10 tasks passed
- server/web focused typechecks passed
- server focused regression suites: 62 tests passed
- web full unit/DOM suite: 206 files, 1,751 tests passed (the Playwright visual spec is excluded from Vitest collection)
- skill-evals suite: 36 files, 494 tests passed
- server package run: 2,676 tests passed; two unrelated parallel timeout/deadlock cases passed 4/4 when rerun in isolation
- two independent code reviews completed with no blockers

## QA note

A fresh interactive mobile-browser timing could not be recorded because the in-app browser connector timed out. The change therefore adds deterministic database-query and initial-DOM budgets as regression coverage; post-deploy device timing remains useful for confirming the end-to-end production gain.

### DIFF
--- a/packages/server/src/__tests__/agent-chat-status.test.ts
+++ b/packages/server/src/__tests__/agent-chat-status.test.ts
@@ -589,6 +589,42 @@ describe("agent-chat-status", () => {
       expect((await resolveAgentChatStatuses(getApp().db, [])).size).toBe(0);
     });
 
+    it("accepts known list speakers and omits detail-only status reasons", async () => {
+      const { app, peer, chatId } = await newChatWithAgent();
+      await bindPresence(peer.agent.uuid, peer.clientId);
+      await setSession(peer.agent.uuid, chatId, "active");
+      await insertEvent(peer.agent.uuid, chatId, 1, "error", {
+        message: encodeProviderRetryEventMessage({
+          event: "provider_retry_scheduled",
+          provider: "codex",
+          scope: "provider_turn",
+          category: "transient_transport",
+          reasonCode: "provider_transient_transport",
+          attempt: 1,
+          maxAttempts: 2,
+          retryMode: "foreground",
+          delayMs: 500,
+          replaySafety: "pre_visible",
+          userSeverity: "info",
+        }),
+      });
+
+      const full = (await resolveAgentChatStatuses(app.db, [chatId])).get(chatId)?.[0];
+      const list = (
+        await resolveAgentChatStatuses(app.db, [chatId], {
+          includeStatusReason: false,
+          nonHumanSpeakersByChat: new Map([[chatId, new Set([peer.agent.uuid])]]),
+        })
+      ).get(chatId)?.[0];
+
+      expect(full?.statusReason?.kind).toBe("retrying");
+      expect(list?.statusReason).toBeUndefined();
+      expect(list?.agentId).toBe(full?.agentId);
+      expect(list?.main).toBe(full?.main);
+      expect(list?.working).toBe(full?.working);
+      expect(list?.errored).toBe(full?.errored);
+    });
+
     it("excludes humans from the union (a human speaker never appears)", async () => {
       const { app, admin, chatId } = await newChatWithAgent();
       const all = (await resolveAgentChatStatuses(app.db, [chatId])).get(chatId) ?? [];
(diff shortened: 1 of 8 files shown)


baixiaohang (changes requested):
- Rationale: the new progressive renderer stops observing after its first append, so a normal 50-row first page can leave rows 33–50 unreachable.

**Risk level**: B-high
- Path baseline: touches `packages/server/src/services/agent-chat-status.ts` and `packages/web/**`.
- Semantic lift: changes the shared per-chat agent-status projection and Mobile Work list rendering.

**PR summary**
- Problem: Mobile Work startup pays to hydrate overlapping server list slices and mount every buffered card before the user can act.
- Approach: hydrate the first-page union once, reuse speaker membership, omit detail-only retry reasons from list status, cap the initial DOM at 16 cards, and lazy-load uploaded avatars.
- Impacted modules: server chat/status list hydration; Mobile Work rendering; chat-row avatars.

**Review findings**
❌ 1. Rebind the `IntersectionObserver` after each rendered batch. When the first intersection changes the list from 16 to 32 keyed rows, React replaces the unkeyed sentinel that moved from index 16 to index 32. `hasBufferedRows` remains `true` and `orderedRows.length` remains unchanged, so the effect at line 188 does not rerun and the observer stays attached to the detached old node. With the normal 50-row page and no `nextCursor`, rows 33–50 can never mount. Include the rendered count (or a stable sentinel callback/ref lifecycle) in the observer binding, and extend the regression to more than two batches—for example 50 rows with two intersection triggers. [R4 / `packages/web/src/pages/mobile/work.tsx:188`]
✅ 2. The server-side union hydration preserves additive `rows`, whole-set attention/pin precedence, and the shared session-status axes while removing repeated membership/detail-only work.

**Action taken**
- Submitted request changes on head `f0326e9e8`.

yuezengwu (approved):
Reviewed the complete server and Mobile Work diff at f0326e9e8894. The first-page hydration reuse preserves the additive response and priority ordering contracts; the status optimization keeps statusReason detail-only while reusing canonical speaker membership; and the 16-row progressive mount retains explicit pagination. No blockers found.

Gandy2025:
Addressed in `d6f9b2ccb`.

The observer is now scoped to the current rendered batch. `renderedRows.length` is a dependency, and the callback advances to the batch-specific next count; after React replaces the sentinel, the effect disconnects the old observer and binds a new one to the current node.

The regression fixture now uses 50 rows and verifies all transitions:

- 16 initial rows
- 32 after the first intersection
- 48 after the second intersection
- 50 after the third intersection, with the sentinel removed

It also asserts that the observer instance is replaced between buffered batches.

Validation:

- Mobile Work focused test: 6/6 passed
- Complete mobile test directory: 18 files, 87 tests passed
- Web typecheck and design-token guardrails passed

@baixiaohang please re-review the updated head.

yuezengwu (approved):
Re-reviewed the successor head d6f9b2ccb81d. The batch-scoped effect now disconnects the observer attached to the replaced sentinel and observes the current sentinel after every append; the 50-row regression covers 16 → 32 → 48 → 50 and verifies observer replacement. The prior blocker is resolved, and no new blockers were found.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the prior multi-batch reachability blocker is fixed on the successor head, and the complete updated diff preserves the list and status contracts without new blockers.

**Risk level**: B-high
- Path baseline: touches `packages/server/src/services/agent-chat-status.ts` and the Mobile Work data/rendering path.

**Review findings**
✅ 1. The batch-scoped observer now rebinds after every sentinel replacement, and the 50-row regression covers 16 → 32 → 48 → 50 plus observer replacement.
✅ 2. First-page union hydration preserves additive `rows`, whole-set attention/pin ordering, and the shared per-chat status axes while removing repeated work.
✅ 3. All checks on head `d6f9b2ccb81d02190aab77e826b7c997f038e004` completed successfully or were intentionally skipped.

**Action taken**
- Approved the reviewed successor head.

PR #2037, opened 2026-07-28
fix(web): keep chat navigation responsive

Gandy2025:
## Summary

- disable React Router transition scheduling at the app router boundary so URL-backed chat selection commits before competing live runtime updates
- add a deterministic regression test for the observed state where the URL changes but the selected chat view remains stale

## Root cause

React Router 7.13 wraps browser location updates in `React.startTransition` by default. Continuous ordinary-priority runtime updates can delay that Router location commit after `history.pushState` has already changed the URL, leaving the conversation list and center view on the previous chat.

## Scope

The product change is one `BrowserRouter` option. It does not add local selection state, reload fallbacks, throttling, or changes to conversation-list click handlers.

## Validation

- regression test verified red without the Router option and green with it
- `pnpm --filter @first-tree/web exec vitest run src/__tests__/app-routes.test.tsx` — 9/9 passed
- full Web suite — 207 files, 1757/1757 tests passed
- `pnpm check` — passed with existing non-blocking warnings
- `pnpm typecheck` — 10/10 tasks passed, including the production Web build
- monorepo test run reached unrelated `skill-evals` fixture timeouts under parallel load; the timed-out cases passed when rerun in isolation
- two independent code reviews found no blocking issues

### DIFF
--- a/packages/web/src/__tests__/app-routes.test.tsx
+++ b/packages/web/src/__tests__/app-routes.test.tsx
@@ -34,6 +34,10 @@ const serverChannelStateMock = vi.hoisted(() => ({
   } | null,
 }));
 
+const routerCommitMock = vi.hoisted(() => ({
+  selectedChatIdAfterUrgentUpdate: null as string | null,
+}));
+
 vi.mock("../hooks/use-server-channel.js", () => ({
   useServerChannelState: () => serverChannelStateMock,
   useContextTreeSetupPreviewBootstrapState: () => serverChannelStateMock,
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the router-boundary change and deterministic regression coverage at head 01c4abe. Disabling React Router transition scheduling here directly addresses the URL/view divergence while preserving URL-backed selection as the single source of truth. No blocking issues found.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The router-boundary opt-out directly restores synchronous URL-backed selection commits, and the deterministic regression test covers the competing-update ordering that caused the stale chat view.

**Risk level**: C
- Semantic lift: none; no subscription, WebSocket frame parsing, or data-layer behavior is changed

**PR summary**
- Problem: Selecting another chat could update the browser URL while the conversation rail and center view remained on the previous chat under continuous live runtime updates.
- Approach: Disable React Router transition scheduling at the single app router boundary, keeping the URL as the selection source of truth, and add a deterministic urgent-update regression test.
- Impacted modules: Web app router and app-route tests

**Review findings**
✅ The option matches React Router 7.13.1 declarative-router semantics, and the test verifies the location commit is visible before a competing synchronous update.
✅ The fix avoids introducing duplicate local selection state or per-click workarounds.

**Action taken**
- Approved.

PR #2040, opened 2026-07-28
feat(client): inject managed MCP into Cursor and Kimi

yuezengwu:
## Summary

- project First Tree-managed MCP servers into each Cursor agent workspace through a private, atomically replaced
  `.cursor/mcp.json`
- approve only projected managed servers with Cursor's native `mcp enable <name>` command; never use the broad
  `--approve-mcps` flag
- snapshot the current MCP/model/environment payload once per Cursor turn
- coordinate the shared workspace by a secret-safe projection fingerprint:
  - matching projections hold shared leases and run concurrently
  - non-empty projections include the effective Cursor approval-store selector from the final child environment
  - empty projections remain independent of the unused approval store and keep their configured parallelism
  - a different projection waits for active users to drain, then performs one exclusive materialize/approval transition
  - failed or aborted transitions invalidate the cached projection before another turn can lease it
- route Cursor MCP setup failures through the existing provider failure taxonomy, bounded retry budget, terminal event,
  and durable-notice-before-ACK contract
- pass caller-scoped `mcpServers` directly to Kimi Code SDK create and resume sessions without changing the pinned SDK
- update deterministic tests and live Cursor/Kimi QA cases

## Why

The effective MCP payload already reaches the client, but Cursor and Kimi Code previously emitted an unsupported
warning instead of giving their provider runtime the configured servers. This keeps the provider-neutral payload and
uses each provider's smallest native injection boundary without widening approval to operator-owned MCP configuration
or reducing the agent's configured chat concurrency.

### Approval scope and current payload

- removed `--approve-mcps` in favor of native per-server project approval
- derived projection, approval, model, and environment from one current immutable turn snapshot
- covered active empty → add → remove convergence and all supported transports

### Shared-workspace concurrency

- replaced full-turn serialization with per-workspace projection leases
- allowed matching/empty fingerprints to share the live projection concurrently
- added FIFO transition fairness so a different fingerprint cannot starve
- held the old projection stable until every active lease released
- invalidated state when approval fails after the atomic file replacement

### Failure policy

- kept suspend/shutdown aborts non-terminal
- classified bounded timeout/transport-like setup failures as transient and applied the existing two-retry
  provider-turn budget
- classified missing capability, unsupported approval, permission, and configuration failures as operator-action
  terminal outcomes
- emitted standard provider retry/terminal events and completed terminal failures through the consumed-error contract,
  which posts the durable runtime notice before ACK
- sanitized provider output so projected headers or CLI stderr cannot enter events

### Approval-store identity

- verified the installed official Cursor Agent package selects project approval state from non-blank `CURSOR_DATA_DIR`,
  otherwise from the child process home (`HOME` on POSIX, `USERPROFILE` on Windows)
- included that effective selector from the exact final child environment in the secret-safe projection digest
- covered same MCP plus changed data-dir/home roots: the new cohort waits, reruns targeted approval, then starts
- kept empty MCP cohorts parallel across approval-root changes because they execute no approval

## Validation

- focused Cursor handler tests: 24 passed
- full Client suite: 1,702 passed, 3 skipped
- QA package: 4 passed
- `pnpm check` (exit 0; 16 existing warnings outside this diff)
- `pnpm build` / dependency builds exercised by root typecheck
- `pnpm typecheck`: 10/10 tasks
- changed-file Biome and `git diff --check`
- isolated Docker-backed QA harness reached READY across shipped surfaces on immutable snapshot
  `d7963abd70fb4e98b44ada763b9ee4bb3edfde08`; the later approval-store identity repair is covered deterministically at
  exact PR head
- real official Cursor Agent concurrency:
  - matching projection: the second MCP tool call completed while the first 35-second call was still live; two provider
    processes, one targeted approval, zero broad approvals
  - different projection: `beta` did not spawn while `alpha` was live; the file stayed on `alpha`, then switched once
    to `beta` after release
- real Cursor approval failure: zero provider spawn/redelivery, sanitized configuration terminal event, consumed outcome
- real Cursor empty → add → remove: correct private file snapshots and one managed tool marker
- real Kimi SDK create and resume: both turns called the injected managed MCP tool
- formal QA result for the provider behavior: PASS; evidence retained under the isolated run root with credentials and
  account identifiers excluded

## Context Tree

- companion draft: https://github.com/agent-team-foundation/first-tree-context/pull/832
- merge this code PR first, then reconcile and mark the Context Tree PR ready

## Notes

- no Server or Web contract changes
- no shared cross-provider projection framework or provider dependency upgrade

### DIFF
--- a/packages/client/src/__tests__/cursor-handler-engine.test.ts
+++ b/packages/client/src/__tests__/cursor-handler-engine.test.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from "node:events";
-import { mkdtempSync, rmSync } from "node:fs";
+import { mkdirSync, mkdtempSync, readdirSync, readFileSync, rmSync, statSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { SessionEvent } from "@first-tree/shared";
@@ -182,8 +182,8 @@ afterEach(() => {
 });
 
 describe("buildCursorTurnArgs — canonical spawn contract", () => {
-  it("locks the exact canonical argv, with model/resume only when supplied", () => {
-    expect(buildCursorTurnArgs({ model: "", resumeSessionId: null })).toEqual([
+  it("locks the exact canonical argv, with MCP approval/model/resume only when supplied", () => {
+    expect(buildCursorTurnArgs({ approveMcps: false, model: "", resumeSessionId: null })).toEqual([
       "-p",
       "--output-format",
       "stream-json",
@@ -191,13 +191,14 @@ describe("buildCursorTurnArgs — canonical spawn contract", () => {
       "disabled",
       "--force",
     ]);
-    expect(buildCursorTurnArgs({ model: "composer-2.5", resumeSessionId: "abc" })).toEqual([
+    expect(buildCursorTurnArgs({ approveMcps: true, model: "composer-2.5", resumeSessionId: "abc" })).toEqual([
       "-p",
       "--output-format",
       "stream-json",
       "--sandbox",
       "disabled",
       "--force",
+      "--approve-mcps",
       "--model",
       "composer-2.5",
       "--resume",
@@ -256,14 +257,14 @@ describe("cursor handler — per-turn CLI transport", () => {
     expect(seenStdin.indexOf("the prompt body")).toBeGreaterThan(0);
   });
 
-  it("resume: passes --resume with the stream-confirmed id and the operator model verbatim", async () => {
+  it("resume: passes managed MCP approval, --resume, and the operator model verbatim", async () => {
     const events: SessionEvent[] = [];
     const { spawnFn, calls } = makeFakeSpawn([successScript({ sessionId: "sess-real-2", text: "resumed" })]);
     const payload = {
       kind: "cursor" as const,
       prompt: { append: "" },
       model: "gpt-5.3-codex-high",
-      mcpServers: [],
+      mcpServers: [{ name: "docs", transport: "stdio" as const, command: "docs-server" }],
       env: [],
       gitRepos: [],
       resourceSkills: [],
@@ -283,6 +284,7 @@ describe("cursor handler — per-turn CLI transport", () => {
       "--sandbox",
       "disabled",
       "--force",
+      "--approve-mcps",
       "--model",
       "gpt-5.3-codex-high",
       "--resume",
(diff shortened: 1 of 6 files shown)


baixiaohang (changes requested):
- Rationale: The provider-native injection is focused, but the Cursor path broadens approval beyond the managed set and does not converge active turns to a newly resolved MCP payload.

**Risk level**: A

**PR summary**
- Problem: Agents running through Cursor or Kimi currently receive an effective First Tree MCP resource list but cannot use those tools, so choosing either provider silently removes capabilities that work on the existing MCP-aware runtimes.
- Approach: Projects the managed set into Cursor's project MCP file and enables headless MCP loading, while passing the same caller-scoped set to Kimi SDK create/resume; an empty set clears or omits the provider-specific projection.
- Impacted modules: `packages/client` Cursor/Kimi handlers and tests; `packages/qa` live provider cases.

**Review findings**
❌ 1. `--approve-mcps` is not scoped to the generated `.cursor/mcp.json`; Cursor defines it as automatically approving **all** MCP servers. This handler deliberately keeps Cursor's operator-global configuration active (no isolated `CURSOR_CONFIG_DIR`), so the presence of one First Tree-managed server also auto-approves unrelated operator-owned global/project servers that previously retained their own approval boundary. The nearby comment that the flag is "only for" the runtime-owned file is therefore not enforceable by this argv. Please keep managed approval isolated from operator-owned MCP configuration, or bring back an explicit product/safety decision for the widened scope rather than coupling it to `payload.mcpServers.length > 0`. [R5 / `packages/client/src/handlers/cursor/index.ts:144`]

❌ 2. Active Cursor sessions do not converge to the newer MCP payload even though the committed QA contract requires an emptied managed set to clear the file and remove `--approve-mcps` from the next turn. `SessionManager` refreshes the cache before dispatch when `configVersion` advances, and `runTurn()` already reads the model from that latest cache, but MCP materialization runs only in `prepareSession()` while the flag is derived from the old `activePayload`. Removing or adding MCP resources during an active chat therefore leaves the next turn on stale MCP state until suspend/resume. Because every chat for an agent shares the same workspace, a second handler can also replace the shared file while the first continues spawning with the opposite old approval boolean. Derive the file and flag from one current per-turn payload snapshot (or deliberately restart/hold the session on MCP version changes), and cover add/remove on the same active handler. [R1/R5 / `packages/client/src/handlers/cursor/index.ts:788`]

✅ Kimi's caller-scoped mapping reaches both create and resume, preserves all three supported transports, and omits the field for an empty managed set.

✅ The Cursor file replacement is atomic and gives newly created projected files a private mode on POSIX.

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The two original Cursor blockers are fixed, but the repair now serializes every Cursor turn for an agent and treats all pre-provider MCP setup failures as indefinitely retryable.

**Risk level**: A

**PR summary**
- Problem: Cursor and Kimi agents receive First Tree-managed MCP resources but previously could not expose those tools to their provider runtime.
- Approach: Cursor now atomically projects the current per-turn MCP snapshot and uses the provider-native per-server approval command; Kimi receives the caller-scoped MCP set on create and resume.
- Impacted modules: `packages/client` Cursor/Kimi handlers and tests; `packages/qa` live provider cases.

**Review findings**
❌ 1. The workspace mutex is acquired for every Cursor turn and held until the provider process finishes, including turns with no managed MCP. Since every chat for an agent shares this workspace, one slow Cursor chat now blocks all other chats for that agent while those sessions still occupy admitted runtime slots; this silently reduces the configured default per-agent concurrency from 99 to one. Please preserve parallel turns that use the same stable projection (for example, serialize setup/projection transitions while allowing matching projection fingerprints to share the live state), or establish an explicit product scheduling contract for the throughput reduction rather than coupling it to MCP support. [R3/R5 / `packages/client/src/handlers/cursor/index.ts:206`; `packages/shared/src/config/agent-config.ts:5`]

❌ 2. Every projection or `mcp enable` error takes the same `token.retry(..., "cursor_mcp_projection_failed")` path. That bypasses the existing provider failure taxonomy and retry budget: an unsupported older CLI, an approval-store permission error, or another deterministic configuration failure can be redelivered indefinitely while the user sees only “delivery will retry,” with no durable operator-action classification. Keep aborts non-terminal, but classify bounded timeouts/transport-like failures separately from capability/configuration failures and settle them through the existing retry/terminal notice contract. The new test currently locks the incorrect behavior by expecting an arbitrary approval error to retry. [R4 / `packages/client/src/handlers/cursor/index.ts:174`; `packages/client/src/handlers/cursor/index.ts:885`; `packages/client/src/__tests__/cursor-handler-engine.test.ts:724`]

✅ The broad `--approve-mcps` flag is gone; static inspection of the official Cursor package confirms `mcp enable <identifier>` is a real per-server project approval command.

✅ The immutable per-turn payload now drives MCP projection, approval, model, and environment, so an active handler converges on add/remove updates on its next turn.

✅ Kimi create/resume injection remains correctly caller-scoped across the supported transports.

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The previous concurrency and failure-policy blockers are fixed, but the shared projection identity does not include the provider environment that selects Cursor's project approval store.

**Risk level**: A

**PR summary**
- Problem: Cursor and Kimi agents receive First Tree-managed MCP resources but previously could not expose those tools to their provider runtime.
- Approach: Cursor projects and individually approves one immutable per-turn MCP snapshot while projection leases preserve concurrency for matching state; Kimi passes the caller-scoped MCP set to SDK create/resume.
- Impacted modules: `packages/client` Cursor/Kimi handlers and tests; `packages/qa` live provider cases.

**Review findings**
❌ 1. The projection fingerprint hashes only the MCP server list, but whether that list is approved also depends on the per-turn environment. The official Cursor CLI derives the project approval file from `CURSOR_DATA_DIR` (or the process home), while this handler allows managed env to override those values and passes the resulting snapshot to both `mcp enable` and the provider process. If an agent config update changes `CURSOR_DATA_DIR`/home without changing `mcpServers`, the new turn has the same fingerprint and leases the cached projection without running setup; its provider process then reads a different `mcp-approvals.json`, where the managed servers were never enabled. The same mismatch can happen across concurrent old/new environment snapshots. Include every approval-store selector in the secret-safe coordination identity (or otherwise invalidate/re-run targeted approval when that scope changes), and cover same MCP + changed approval root. [R1/R5 / `packages/client/src/handlers/cursor/index.ts:123`; `packages/client/src/handlers/cursor/index.ts:641`]

✅ Matching and empty projections now run concurrently, while differing projections retain FIFO transition safety.

✅ Setup failures now use the shared failure taxonomy, finite retry budget, terminal events, and consumed-error durable-notice path.

✅ Targeted per-server approval, active add/remove convergence, and Kimi create/resume injection remain correct.

**Action taken**
- Submitted request changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact head closes every Cursor approval-scope, active-payload, concurrency, approval-store identity, and failure-policy blocker while preserving the focused Kimi create/resume injection.

**Risk level**: A

**Review findings**
✅ Managed Cursor MCP approval remains targeted per server and excludes operator-owned MCP configuration.

✅ Matching and empty projections preserve configured concurrency; different projections switch safely after active leases drain.

✅ The projection identity now includes the effective approval-store selector for non-empty MCP sets, so environment-only root changes rerun targeted approval.

✅ Setup failures follow bounded retry, terminal classification, and durable-notice-before-ACK policy; Kimi create/resume injection remains caller-scoped.

**Action taken**
- Approved exact head `8e8dc3a0034840ba413857f6cbd5bc9735c6ad28` with explicit Code Owner authorization.

PR #2039, opened 2026-07-28
feat: support Codex fast mode

yuezengwu:
## Summary

- add a Codex-only `serviceTier` runtime setting with legacy-row and mixed-version wire backfill to `default`
- expose Standard/Fast selection in Web runtime settings and add the public `agent config set-service-tier` CLI command
- forward the provider-native tier through both Codex SDK and app-server engines
- leave `features.fast_mode` unset for Standard so provider-managed policy remains authoritative; enable it only for non-default tiers
- convert Codex's unsupported-tier warning into a terminal configuration failure in both engines, and validate the app-server's effective tier before starting a turn, so Fast cannot silently run as Standard
- recognize Codex's provider-defined `fast` request alias and canonical `priority` effective id as the same Fast selection without rewriting the persisted or outbound value
- preserve other open provider-advertised tier IDs without a First Tree model/account compatibility matrix

## Review and QA follow-up

Current exact head: `f6facfe84153bc994fe68011c176f5507232c954`.

The review findings on `ac5e8df8f985e0e297d726777ac6e6adea0305b2` are addressed:

1. SDK warning items and app-server warning notifications settle as `codex_service_tier_unsupported`; app-server also rejects an omitted or genuinely mismatched effective tier before `turn/start`.
2. Standard sends `service_tier = "default"` without an explicit Fast feature override.
3. The paired source-backed Context Tree update is [first-tree-context#835](https://github.com/agent-team-foundation/first-tree-context/pull/835), current head `146709d3cb681d5437f05602da56bfcd522ab365`.

Formal QA on successor `beef5cf96946d05c19088051d45b398b7dd0f225` verified both review fixes with a real provider, then correctly returned **FAIL** because live Codex 0.144.1 canonicalizes the shipped `fast` request to effective `priority`; the app-server's exact-string check rejected that valid Fast session. The current head:

- validates the upstream provider-defined alias/canonical pair as equivalent while keeping strict failure for omitted or other mismatched tiers
- adds a deterministic product test where app-server receives `service_tier=fast`, returns effective `priority`, and must proceed to `turn/start`
- treats a missing `serviceTier` from an older Server as legacy Standard instead of producing an `undefined` non-default failure
- updates the paired Tree draft and CLI reference with the canonicalization contract

Targeted real-product QA is being rerun against the current exact head before merge.

## Verification

- `pnpm check` (passes; 16 pre-existing warnings)
- `pnpm typecheck` (10/10 Turbo tasks)
[...]

PR #2041, opened 2026-07-28
fix: stabilize Darwin switching and skill eval tests

yuezengwu:
## Summary

- classify Darwin client-switch processes from a plain argv snapshot instead of the environment-appended `ps -E` command column
- keep the switch drain fail-closed when a provider or daemon appears without a readable environment snapshot
- make skill-evals wait for the real CLI build, resolve workspace imports from source, and use bounded fixture concurrency/timeouts

## Root cause

Darwin `ps -E` appends environment variables to the command column. A test worker whose `PATH` or `NODE_PATH` mentioned `@openai/codex` was therefore classified as a live Codex provider even though its real argv was unrelated.

Separately, `@first-tree/skill-evals` launches the built CLI but did not declare that build edge. Its real git/worktree fixtures also inherited Vitest's 5-second default and unbounded local file parallelism, so the root test graph could race package builds and turn host contention into fixture failures.

## Impact

The production client-switch safety gate remains fail-closed, while unrelated macOS processes no longer block a switch solely because an environment path contains a provider package name. Root test runs now build the CLI before skill-evals and execute their filesystem-heavy fixtures with deterministic bounds.

This is intentionally separate from the Codex Fast mode feature PR #2039.

## Verification

- `pnpm --filter first-tree-dev exec vitest run src/__tests__/client-switch.test.ts src/__tests__/client-switch-platform-extra.test.ts src/__tests__/client-switch-transaction-extra.test.ts` — 68/68 passed
- `pnpm exec turbo run test --filter=@first-tree/skill-evals` — 494/494 passed; 6/6 tasks successful
- `pnpm test` — 12/12 tasks successful
- `pnpm check` — passed (existing non-blocking warnings only)
- `pnpm typecheck` — 11/11 tasks successful

## QA risk

The Darwin process-attribution change touches the local-client switch safety boundary. Automated coverage exercises both the real macOS transaction path and fail-closed late-provider / process-inspection branches. Independent product regression for Codex Fast mode remains scoped to #2039.

### DIFF
--- a/apps/cli/src/__tests__/client-switch-platform-extra.test.ts
+++ b/apps/cli/src/__tests__/client-switch-platform-extra.test.ts
@@ -36,6 +36,9 @@ const execState = vi.hoisted(() => ({
     | "daemon-untrusted"
     | "daemon-active"
     | "many-daemon-untrusted"
+    | "provider-name-in-env-only"
+    | "provider-missing-env-snapshot"
+    | "command-throw"
     | "pid-throw",
   home: "",
 }));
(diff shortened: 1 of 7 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The Darwin process-classification fix removes the environment-only false positive without weakening the client-switch drain gate, and the skill-eval scheduling changes follow the repository's existing build and Vitest patterns.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: On macOS, an unrelated process could block a local user switch merely because its environment path mentioned a provider package; root test runs could also race the CLI build and fail under host contention.
- Approach: Classify processes from a plain argv snapshot while using the `ps -E` snapshot only for attribution markers, fail closed when a newly observed provider or daemon has no matching environment snapshot, and give skill-evals an explicit CLI build edge plus bounded fixture execution.
- Impacted modules: `apps/cli` client-switch drain scanning and `packages/skill-evals` test orchestration.

**Review findings**
✅ 1. Provider classification no longer consumes environment-appended command text, while either `ps` failure and unmatched provider/daemon snapshots still abort before root-state movement.
✅ 2. Linux scanning and the surrounding two-pass drain/transaction boundary remain unchanged.
✅ 3. The workspace dependency plus package-level `^build` supplies the real CLI artifact, and the Vitest aliases/fork limits reuse established repository conventions.

**Action taken**
- Approved after explicit maintainer authorization.

PR #2043, opened 2026-07-28
Add BYO Context integration for Claude Code and Codex

baixiaohang:
## Summary

- add user-scoped Claude Code and Codex Context plugins with deterministic Read/Write Skill projections, SessionStart activation, release manifests, integrity checks, repair, disable, and transactional rollback
- bind an explicit Setup/invite Team to an exact provider + source checkout while keeping live authority in an org-scoped Server activation validator
- extract the canonical Context Tree Policy into one runtime asset projected byte-for-byte into managed briefings and external provider bundles
- add Admin repository confirmation and Team-scoped Setup/member invite handoffs in the Web app
- preserve existing Client daemon/Computer Connected behavior and keep external provider sessions outside the First Tree Agent/Chat/Session model

## Key contracts

- activation uses `POST /api/v1/orgs/:orgId/context-activation/validate`, `requireOrgMembership`, a strict repository-only request body, and no cross-Team resolver
- user-scope Plugin discovery does not activate unbound repositories; SessionStart and hidden Read/Write routes fail closed without an exact local binding plus live membership/repository/Tree validation
- Reviewer health never gates Read, but official Write fails before remote mutation when automatic review is unavailable
- External bundles expose only `first-tree-read` and `first-tree-write`; Seed remains a managed setup workflow
- Claude/Codex Skills are deterministic projections of one canonical source and may differ only in provider-fixed adapter routing; canonical Policy bytes and `policyDigest` remain identical
- local Context mutations, provider Plugin state, binding state, repair recovery, and Client account switching use durable coordination and fail-closed recovery

## Validation

- independent First Tree review completed with no blocking findings
- targeted Shared, Client, CLI, Server, and Web tests passed, including activation tenancy, member handoff, provider payload integrity, rollback/crash recovery, account isolation, canonical Policy runtime layouts, and Skill projection parity
- Shared, Client, Server, Web, and CLI typechecks/builds passed during implementation qualification
- `pnpm validate:skill` passed
- `pnpm check` passed with the repository's existing 16 warnings and 1 informational finding
- `git diff --check` passed
- isolated real Codex CLI lifecycle passed: enable/install, healthy status, executable launcher verification, and disable

## Release qualification gaps

Code review and automated validation are complete, but production rollout remains gated on the documented real-surface matrix:

- real Claude Code CLI / local Desktop lifecycle qualification
- macOS and glibc Linux architecture coverage beyond the exercised host
- end-to-end GitHub PR and GitLab MR review/repair/merge paths with a staging Team
- Windows distribution support remains out of P0 because First Tree does not yet ship the required Windows portable/runtime integration
- remote/cloud provider sessions remain out of P0


packages/client/src/runtime/agent-briefing.ts
```diff
@@ -18,10 +18,16 @@ const require = createRequire(import.meta.url);
 const ejsRuntime: typeof ejs = require("ejs");
 const AGENT_BRIEFING_TEMPLATE_FILENAME = "agent-briefing.ejs";
 const TEMPLATE_CANDIDATE_URLS = [
-  // Source execution: packages/client/src/runtime/agent-briefing.ts
+  // Source execution already has runtime/templates beside this module. The
+  // build preserves that relative layout for client, CLI, and portable chunks.
```
yuezengwu: Blocker: this relative-layout claim is false for the shipped CLI and portable entry points. `copy-client-runtime-templates.mjs` writes the assets to `dist/templates` and `dist/runtime-assets`; the real CLI module is `dist/cli/index.mjs` (portable: `app/cli/index.mjs`), so these `./templates` / `./runtime-assets` URLs resolve under the `cli/` directory and neither asset exists there. The previous `../templates` candidate handled exactly this layout, and the new Policy needs the matching `../runtime-assets` candidate.

The new smoke test misses this because it builds/imports `context-policy-runtime-smoke-entry.mjs` at the root of `dist`/`app`, not through the real `cli/index.mjs` location. As written, managed agent startup reaches `buildAgentBriefing()` and throws before the daemon session can start. Please restore the parent-relative candidates (or copy the assets beside the real entry) and cover the actual CLI + portable entry layout.

yuezengwu (changes requested):
This PR adds user-scoped Claude Code/Codex Context plugins, exact checkout-to-Team activation, deterministic Policy/Skill projection, transactional local state, and the Web handoff. The tenancy and fail-closed activation boundaries are aligned with the current Context Tree contracts.

I found one merge blocker in the managed briefing runtime asset lookup; details are inline.

Human check requested for the core persistent-state changes: `context.yaml`, install/operation journals, provider manifests, and their client-account switch/recovery semantics. I did not find a database schema migration in this diff.

Per review policy I inspected the diff without running tests or QA.

packages/client/src/runtime/agent-briefing.ts
baixiaohang: Fixed in 8432ec9ab. The runtime lookup now supports both root-level and nested CLI/portable layouts, and the smoke entry runs beside the real dist/cli and app/cli entry points. The exact CLI CI command also passes all 23 batches locally.

yuezengwu (approved):
Re-reviewed the incremental fix at `8432ec9ab`.

The previous blocker is resolved: managed briefing lookup now supports both root-level package layouts and the real nested `dist/cli` / portable `app/cli` entry layouts, and the smoke coverage is executed from those nested locations. The remaining incremental changes only repair affected test fixtures/imports.

No new blocking findings. The earlier human-check note for the persistent local state structures and account-switch/recovery semantics still applies.

Per review policy I inspected the incremental diff without running tests or QA.

PR #841, opened 2026-06-05
fix(web): make Select options clickable inside modal dialogs

Gandy2025:
## Problem

The Visibility dropdown in the **Edit Identity** dialog opens but you can't select another option (reported with screenshots). Root cause confirmed in a real browser, not theorized.

The `Select` panel is portaled to `document.body`, which sits **outside** any open Radix modal `Dialog`/`Popover`. Modal layers set `pointer-events: none` on the body and only re-enable it inside their own content. The portaled panel therefore inherited `none`:

- `getComputedStyle(option).pointerEvents === "none"`, `body === "none"`
- a **real** mouse click on an option times out / does nothing → value unchanged
- a synthetic `.click()` *does* change the value → proves `onChange` was never broken; only the real pointer was being blocked

This is a systemic `Select` bug — it affects every `Select`-in-`Dialog`: `identity-section`, `re-bind-dialog`, `mcp-section`, and `settings/resource-editors`.

## Fix

One change in the `Select` primitive, so all usages are repaired at once:

1. `pointerEvents: "auto"` on the portaled panel — overrides the inherited `none`.
2. `onPointerDown` stops propagation to the document — so the host modal's dismiss-on-outside-interaction listener doesn't treat an option click as an "outside" click and close the dialog.

## Verification (real browser, no-auth styleguide repro)

- ✅ Real click on an option changes the value and **keeps the dialog open**; panel closes after pick
- ✅ `Select` outside a dialog still works (model picker: opus → sonnet)
- ✅ outside-click still closes the dropdown
- ✅ `pnpm check` clean, `pnpm typecheck` 11/11, web tests **674 passed**
- ✅ new happy-dom regression test `select-portal.test.tsx` **fails without the fix** (`expected '' to be 'auto'`), passes with it

### Why existing tests didn't catch it

jsdom/happy-dom doesn't model layout or inherited `pointer-events`, so DOM tests stayed green while the real browser was broken. The new test pins the inline `pointer-events: auto` escape hatch as a cheap guard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2054, opened 2026-07-29
fix: align GitLab Context Reviewer readiness

baixiaohang:
## Summary

- authorize Context Reviewer runtimes by the manager's user-owned Computer instead of the legacy client organization placeholder
- accept a successfully processed Project Hook merge request from the exact bound GitLab Context Tree repository as repository-scoped Reviewer readiness while preserving System Hook full-instance readiness
- persist and reset repository-scoped Project Hook evidence, keep wrong-repository and failed processing paths fail-closed, and add the generated database migration
- share an exhaustive Setup blocker-copy source and preserve precise or future server error messages
- update the GitLab lifecycle QA case for the new readiness boundary

## Validation

- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 --reporter=dot` — 249 files, 2702 tests passed
- `pnpm --filter @first-tree/web exec vitest run --maxWorkers=2 --reporter=dot --exclude tests/visual-regression.spec.ts` — 207 files, 1773 tests passed
- focused Server Context Reviewer/GitLab suites — 77 tests passed
- focused Web Setup DOM suite — 58 tests passed
- `pnpm check` — passed with existing repository warnings only
- changed-file Biome and `git diff --check` — passed

## QA risk

This changes provider ingress readiness, multi-Team runtime authority, persisted readiness evidence, and Web recovery messaging. The existing `gitlab-context-tree-lifecycle` cross-surface QA case now covers the repository-scoped Project Hook boundary; formal QA is warranted for a release candidate but is not run as part of this PR.

### DIFF
--- a/packages/qa/cases/cross-surface/gitlab-context-tree-lifecycle.md
+++ b/packages/qa/cases/cross-surface/gitlab-context-tree-lifecycle.md
@@ -29,9 +29,11 @@ runtime, network, credential, and cross-surface behavior.
   the derived `/admin/hooks` page, with Push and Merge Request events enabled,
   the default payload, and no custom template. Reuse the same one-time URL in a
   trusted Project Hook with Merge Request, Issue, and Note events enabled.
-  First Tree must observe a processable System Hook Merge Request event before
-  routing or Context Review is treated as ready. Connection creation must not
-  depend on Web Context egress policy.
+  First Tree must observe a processable System Hook Merge Request before
+  full-instance routing is ready; a processable Project Hook Merge Request from
+  the exact bound Context Tree repository is sufficient only for that Team's
+  Context Reviewer readiness. Connection creation must not depend on Web
+  Context egress policy.
   For a Self-Managed origin, the deployment operator—not the Team
   admin—separately authorizes the exact HTTPS origin through
   `FIRST_TREE_GITLAB_ALLOWED_ORIGINS`. Retain evidence of policy shape but no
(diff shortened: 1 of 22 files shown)


yuezengwu (changes requested):
Implementation goal reviewed: allow an exact bound GitLab Context Tree Project Hook MR to prove repository-scoped Reviewer readiness while System Hook evidence remains instance-wide; move Reviewer runtime authority to the manager's user-owned Computer; persist the new readiness evidence; and share precise Setup recovery copy.

Blocking: `projectGitlabContextReviewerReadiness` only compares the persisted canonical repository string with the current binding, and no binding mutation clears or versions that evidence. After repository A records a successful Project Hook MR, changing the Context Tree binding A → B → A makes the old A evidence valid again without any new inbound MR. That contradicts this function's “a later Context Tree rebind cannot reuse stale evidence” contract and the linked draft Context Tree PR #841, which says repository-scoped evidence stops applying when the binding changes.

Please bind the evidence to a binding generation (or clear it transactionally on every binding change), and add an A → B → A regression test. If reusing evidence after returning to the same repository is intentional, the durable decision and code comments need to say so explicitly before this implementation can be approved.

Human review note: this PR adds two persisted `gitlab_connections` columns/migration and changes the core Reviewer runtime authority check by removing the legacy Client organization placeholder requirement. I found the user-owned Computer change consistent with the current runtime-binding decision, but these schema/authority changes still warrant explicit human inspection.

I reviewed the diff and generated migration metadata; I did not run tests or QA for this review.

baixiaohang:
Fixed in `43e99d28d`.

- Every actual Context Tree binding change now clears the repository-scoped Project Hook readiness evidence in the same transaction as the setting mutation.
- The invalidation covers normal updates, initialized bindings, and binding deletion; no-op updates preserve valid evidence.
- Added an A → B → A regression test that first proves A can enable the Reviewer, then confirms the round trip clears the evidence and requires a new Merge Request event.

Validation for this follow-up:

- `context-reviewer-settings.test.ts` + `org-settings.test.ts`: 88 tests passed with `maxWorkers=2`
- Server typecheck passed
- changed-file Biome and `git diff --check` passed

The companion Context Tree draft already states that repository-scoped evidence stops applying when the binding changes, so no durable wording correction was needed.

yuezengwu (approved):
Re-reviewed head `43e99d28d968650e6aea0349a2944bd5e4c28485`.

The previous blocker is resolved: every Context Tree binding mutation now clears repository-scoped Project Hook readiness in the same organization/connection-serialized transaction, initialization and deletion paths are covered, and the A → B → A regression test proves old evidence cannot reactivate. The final implementation remains aligned with draft Context Tree PR #841 and with the manager-owned Computer runtime authority.

The database migration and core runtime-authority change still warrant the human inspection already called out. I reviewed the successor diff and complete affected logic; I did not run tests or QA.

PR #2058, opened 2026-07-29
fix(client): prevent stale slots from deleting rebound runtime tokens

yuezengwu:
## Summary

- serialize each agent's runtime-session token replacement and owned cleanup with a short-lived cross-process mutation lock
- treat the SHA-256 digest of the successfully persisted token as the in-memory generation owned by an `AgentSlot`
- compare and delete only while holding the same lock, preserving a newer bind's token when an older slot stops or cleans up a failed start
- leave every pre-existing lock untouched; bind becomes unavailable and cleanup preserves the token if lock acquisition times out
- keep the stable plain-text token path, atomic replacement, `0700` directory / `0600` file modes, request-time rereads, and no inherited-token fallback
- fail slot startup/rebind closed when the mutation lock or atomic persistence fails; cleanup failures preserve the current file

## Root cause and impact

Cleanup previously deleted the stable per-agent token path unconditionally. If a replacement daemon bound the same agent and atomically wrote a newly issued token before the old `AgentSlot` stopped, the old slot could remove the replacement generation. Long-lived provider children then kept the right path but found no proof on their next request.

The token content now identifies its generation. A slot remembers only the digest of its last successful write, so a stale slot cannot delete a later bind's token.

## Review follow-up

The initial implementation attempted to reclaim a lock after reading a dead PID. That pathname-only delete was not CAS-safe: another contender could acquire a new lock at the same path before the stale reclaimer removed it.

Head `4ad440666731348f692966f32ba4ee991e3247e4` removes stale-lock reclamation entirely. An existing lock is never deleted by a contender; all contenders wait and then fail closed. A regression starts two independent Node processes using the real `RuntimeSessionTokenFile` against the same dead-PID lock and proves that both remain blocked while the lock and token state stay unchanged.

## Validation

- `pnpm exec biome check packages/client/src/runtime/runtime-session-token-file.ts packages/client/src/runtime/agent-slot.ts packages/client/src/__tests__/runtime-session-token-file.test.ts packages/client/src/__tests__/agent-slot.test.ts`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/runtime-session-token-file.test.ts src/__tests__/agent-slot.test.ts` — 41 passed
- `pnpm --filter @first-tree/client test` — 1,731 passed, 3 skipped
- `pnpm --filter first-tree-dev exec vitest run src/__tests__/local-agent-runtime-token-request.test.ts src/__tests__/agent-shared-helpers-extra.test.ts src/__tests__/tree-review.test.ts` — 21 passed
- `pnpm check` — passed; unchanged files continue to report the repository's existing warnings
- `pnpm test` — 12/12 workspace tasks passed

The regression coverage includes both cross-process lock orderings, two contenders racing over one stale lock, A/B replacement ownership, same-slot rebind cleanup, failed-start cleanup, lock and atomic-write failures, missing/empty/unreadable/mismatched files, stable-path rereads, and private file modes.

## QA and task isolation

Independent live QA passed exact head `4ad440666731348f692966f32ba4ee991e3247e4`:

- four stale-lock race variants and eight independent contenders all timed out fail-closed at the product's default timeout; no contender entered the critical section, 9,102 observer samples saw no mutation, and the lock/token metadata and contents remained unchanged with no temp residue
- A→B replacement rotated the generation at the same private stable path; one unchanged child PID received HTTP 200 before and after replacement and after A's stale stop, while the file retained B's digest
- B's owned stop removed B's generation and left no token, lock, or temp residue
- targeted ownership/token tests passed 41/41; the full Client suite passed 1,731 with 3 skipped; CLI token reread/no-fallback tests passed 11/11; workspace typecheck passed 11/11; and every applicable exact-head GitHub CI/CodeQL check is green

QA classified the existing live token-delivery case as `candidate-case-update`; deterministic stale-lock contention remains at the product-test layer.

The existing reusable case `packages/qa/cases/cross-surface/agent-runtime-session-token-delivery.md` covers the stable-path, next-request reread, and no-fallback contracts; deterministic ownership and interleaving coverage lives in the client tests added here.

This branch was created directly from `origin/main` and does not include Issue 1's daemon single-instance ownership changes from `fix/daemon-home-owner-lock`. The two changes have zero file-path overlap, so there is no required merge order.

### DIFF
--- a/packages/client/src/__tests__/agent-slot.test.ts
+++ b/packages/client/src/__tests__/agent-slot.test.ts
@@ -1,9 +1,12 @@
 import { EventEmitter } from "node:events";
-import { existsSync, readFileSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, statSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import type { ClientConnection, SessionReconcileResult } from "../client-connection.js";
 import type { AgentSlotConfig } from "../runtime/agent-slot.js";
 import type { HandlerConfig } from "../runtime/handler.js";
+import { RuntimeSessionTokenFile } from "../runtime/runtime-session-token-file.js";
 import { type FirstTreeHubSDK, type RegisterResult, SdkError } from "../sdk.js";
 
 type FakeLogger = {
@@ -178,7 +181,12 @@ function makeFrame(overrides: Record<string, unknown> = {}) {
 }
 
 function installMocks(
-  options: { syncResult?: MockState["syncResult"]; syncDelayMs?: number; dispatchRejectEntryId?: number } = {},
+  options: {
+    syncResult?: MockState["syncResult"];
+    syncDelayMs?: number;
+    dispatchRejectEntryId?: number;
+    dataDir?: string;
+  } = {},
 ): MockState {
   vi.resetModules();
   const state: MockState = {
@@ -193,7 +201,7 @@ function installMocks(
   };
 
   vi.doMock("@first-tree/shared/config", () => ({
-    defaultDataDir: () => "/tmp/first-tree-test-data",
+    defaultDataDir: () => options.dataDir ?? "/tmp/first-tree-test-data",
   }));
   vi.doMock("../observability/logger.js", () => ({
     createLogger: () => state.logger,
@@ -312,6 +320,7 @@ async function makeSlot(options?: {
   dispatchRejectEntryId?: number;
   omitReconcileInterval?: boolean;
   deferSuspendOnSubprocess?: boolean;
+  dataDir?: string;
 }): Promise<{
   slot: import("../runtime/agent-slot.js").AgentSlot;
   connection: FakeClientConnection;
@@ -322,6 +331,7 @@ async function makeSlot(options?: {
     syncResult: options?.syncResult,
     syncDelayMs: options?.syncDelayMs,
     dispatchRejectEntryId: options?.dispatchRejectEntryId,
+    dataDir: options?.dataDir,
   });
   const sdk = makeSdk({
     agent: options?.agent,
@@ -784,6 +794,7 @@ describe("AgentSlot", () => {
       await Promise.resolve();
 
       expect(readFileSync(tokenFile, "utf8").trim()).toBe("runtime-token-2");
+      expect(connection.tokenProviders.get("agent-1")?.()).toBe("runtime-token-2");
     } finally {
       await slot.stop();
     }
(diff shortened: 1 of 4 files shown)


yuezengwu:
@baixiaohang Ready for code-owner review.

Independent QA completed against exact head `85b0145cc7f6e6e6eebd6451b344e48a6c091638` with a **PASS**:

- direct-parent baseline reproduced the stale-slot deletion and the long-lived child's next request changing from HTTP 200 to 403
- the candidate preserved B through A's stale stop, kept the same child PID at HTTP 200, and let B's own stop remove the latest generation
- five fresh-runtime-home repetitions passed 5/5 with no lock/temp residue
- all applicable GitHub checks, the 40 ownership/token tests, 21 CLI reread/no-fallback tests, and the 1,730-test client suite passed

The PR remains scoped to the four client runtime/test files and has zero path overlap with the separate daemon single-instance ownership change.

baixiaohang (changes requested):
- Rationale: stale-lock reclamation can delete a different contender's newly acquired lock, so token replacement and owned cleanup are not reliably serialized.

**Risk level**: A
- Path baseline: includes `packages/client/**` → A.
- Semantic lift: none; the path baseline already sets A.

**PR summary**
- Problem: when a replacement daemon binds the same agent, an older slot can delete the replacement token and make a long-lived child lose agent-scoped HTTP proof on its next request.
- Approach: remember the digest of the token generation persisted by each slot, then serialize replacement and compare-and-delete with a per-agent cross-process lock while retaining the stable token path and request-time rereads.
- Impacted modules: client `AgentSlot`, runtime-session token-file storage, and their regression tests.

**Review findings**
❌ 1. Stale-lock reclamation has a TOCTOU that can recreate the stale-slot deletion race. After reading a dead PID, `removeStaleMutationLock()` deletes `this.lockPath` without proving that the pathname still refers to the lock it inspected. With two contenders, both can observe the same dead lock; contender B removes it and acquires a fresh lock, then contender C resumes and deletes B's live lock before acquiring its own. Both processes now enter the critical section. An old cleanup can therefore read generation A, a replacement can write B, and the cleanup can then unlink B. Please use a stale-recovery protocol with atomic ownership/CAS semantics (or fail closed without unsafe reclamation) and cover two contenders racing over the same stale lock. [R4 / `packages/client/src/runtime/runtime-session-token-file.ts:164`]

**Action taken**
- Submitted request changes.

yuezengwu:
@baixiaohang Exact-head follow-up is ready for re-review at `4ad440666731348f692966f32ba4ee991e3247e4`.

Independent QA passed the fail-closed stale-lock change:

- 4/4 stale-lock race variants and 8/8 independent contenders timed out without entering the critical section or mutating lock/token state
- A→B replacement kept one unchanged child PID at HTTP 200 through B's bind and A's stale stop; the file retained B's digest
- B's owned stop removed B cleanly with no token, lock, or temp residue
- targeted 41/41, Client 1,731 passed / 3 skipped, CLI reread/no-fallback 11/11, workspace typecheck 11/11, and all applicable exact-head CI/CodeQL checks passed

The PR description now carries the exact-head QA evidence.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the previous stale-lock TOCTOU blocker is fixed by removing unsafe stale-lock reclamation and failing closed on lock contention; exact-head QA and all applicable checks are green.

**Risk level**: A
- Path baseline: includes `packages/client/**` → A.
- Semantic lift: none.

**Review findings**
- ✅ Token replacement and owned cleanup are serialized under one per-agent cross-process lock, so a stale slot cannot delete a newer generation.
- ✅ Existing locks are never removed by contenders; acquisition timeout preserves token state and prevents concurrent critical sections.
- ✅ Stable-path rereads, private file modes, and no inherited-token fallback remain intact.
- ✅ The change does not modify ACK, Inbox recovery, handler, or chat/session/message core mechanisms.

Reviewed at exact head `4ad440666731348f692966f32ba4ee991e3247e4`. I did not run local tests; independent exact-head QA and all applicable GitHub CI/CodeQL checks passed.

PR #2061, opened 2026-07-29
feat(client): reconcile managed skills by provider

yuezengwu:
## Summary

- Replace the dual Core installer and separate Team materializer with one Client-side Managed Skills Reconciler that projects Core and Team Skills only into the active provider's native discovery root.
- Introduce schema-v2 managed state with a monotonic Team Resource version fence, final-tree digests, ownership markers, a process mutex plus persistent kernel-backed workspace lock, journaled directory swaps, and deterministic crash recovery.
- Preserve last-known-good Team Skills for unavailable or stale configuration, handle user-name conflicts without overwrites, and conservatively migrate v1 Core targets plus legacy Team materializations.
- Await reconciliation across Claude SDK/TUI, Codex SDK/app-server, Cursor, Kimi, and legacy Claude resume before building the briefing or starting the provider. Briefings now consume immutable successful Team rows and expose no filesystem paths.
- Remove name-only orphan Skill deletion, document the runtime state contract, and add a reusable provider-backed QA case.

## Design choices

- An incoming Team config version lower than the persisted fence cannot mutate Team targets; the same version may retry; after pending transaction recovery, a higher version persists its fence before any new snapshot-driven Team filesystem mutation.
- Config unavailability has explicit last-known-good/eventual-revocation semantics rather than interpreting an empty fallback as authoritative.
- Core bundle drift is detected from the complete final tree, so supporting-file edits converge even when `VERSION` was not bumped.
- Every install/remove transaction records validated target-related staging/backup paths and before/after state. Ambiguous or future/corrupt state fails closed.
- `.first-tree-workspace/managed-skills.lock` is a persistent regular file whose descriptor carries the OS lock. The path is never renamed or removed, owner death is released by the kernel, and no PID/age-based reclamation can detach a successor.
- `fs-native-extensions` stays outside the ESM CLI bundle and is a direct published CLI dependency. npm and portable release paths therefore install its platform prebuilds beside the bundle; every portable target is checked for its `.node` prebuild before archiving.

Coordinated Context Tree design: https://github.com/agent-team-foundation/first-tree-context/pull/844. It remains draft until this code PR is merged.

## Validation

- `pnpm --filter @first-tree/client test` — 156 files passed, 1699 tests passed, 3 environment-gated tests skipped
- `pnpm --filter first-tree-dev test` — all 23 bounded CLI batches passed
- `pnpm --filter @first-tree/skill-evals test` — 36 files and 494 tests passed, including the built-CLI fallback that previously failed
- `pnpm typecheck` — 11/11 monorepo tasks passed
- Client and CLI builds pass; the built npm bin entry executes `--version` with the external native addon loaded
- A real darwin-arm64 portable artifact builds and verifies, including dependency manifest, target prebuild, and packaged `--version` entry
- Focused reconciler, kernel-lock, package metadata, built-entry, and portable-builder suites pass
- Three-process lock regression covers dead-owner residue, live-successor contention, forced owner death, and kernel handoff without replacing the lock inode
- `pnpm install --frozen-lockfile --offline`
- `pnpm check` — exits 0; existing unrelated warnings remain
- `pnpm validate:skill` — all 8 bundled Skills valid
- `git diff --check`

The prod and staging npm publish jobs also smoke the exact built bin entry before publication. The portable verifier checks the direct native dependency and target prebuild before launching the packaged wrapper.

## QA

Added `packages/qa/cases/runtime/managed-skill-projection.md` for real Cloud delivery, provider-native discovery, provider switching, unavailable-config behavior, user conflicts, and legacy Claude resume.


baixiaohang (changes requested):
- Rationale: The reconciler has workspace-containment and transaction-serialization gaps that can mutate paths outside the agent home or lose the only recovery record.

**Risk level**: A
- Semantic lift: none (already at the highest path level)

**PR summary**
- Problem: Managed agents currently materialize Core and Team Skills through overlapping installers and roots, so provider switches, stale configuration, user-name conflicts, and interrupted writes can leave discovery inconsistent.
- Approach: Replace those paths with one provider-native reconciler, a monotonic Team-resource fence, ownership/digest state, cross-process locking, journaled directory swaps, and reconciled briefing rows.
- Impacted modules: Client runtime bootstrap/state, Claude/Codex/Cursor/Kimi handlers, workspace migrations, runtime documentation, and provider-backed QA cases.

**Review findings**
❌ 1. Provider-root symlinks bypass the advertised workspace containment boundary. `resolveWorkspacePath` only checks the lexical result of `resolve(workspace, portablePath)`. If `.agents/skills`, `.claude/skills`, `.cursor/skills`, `.kimi-code/skills`, or `.first-tree-workspace` is a symlink to an external directory, every later `mkdir`, `rename`, digest, journal write, and removal follows it outside the agent home. The existing test protects only a symlink at the final legacy Skill target; it does not cover a symlinked ancestor/root. Please fail closed on symlinked managed ancestors or verify the nearest existing ancestor/real target remains beneath the real workspace before any mutation, and add the root-symlink regression. [R4 / `packages/client/src/runtime/managed-skills.ts:1619`]

❌ 2. A failed transaction recovery is treated like an isolated Skill failure, so reconciliation can continue and overwrite the unresolved journal. `installStagedSkill` / `removeManagedEntry` call `recoverPendingJournal` in their catch paths; if that recovery itself throws (ambiguous target/backup facts, a newer fence, or an I/O failure), the per-Skill catches at lines 312 and 336 swallow it and continue. The next target then calls `writeManagedSkillsJournal`, replacing the only record needed to recover the prior swap. Recovery failure must abort the entire workspace reconcile (and remain distinguishable from a safely recovered per-Skill install failure) before any later target mutation. [R4 / `packages/client/src/runtime/managed-skills.ts:312`]

❌ 3. The stale-lock rule can reclaim a lock from a live owner. `mayRecoverLock` returns `true` solely because `createdAt` is older than five minutes before it checks `process.kill(pid, 0)`. A slow reconcile can therefore be deleted by another process; both processes then mutate the single state/journal concurrently. There is also a token-check/recursive-remove race in `release`: after reading its token, the old owner can delete a successor lock installed between that read and `rm`. Please use a renewable/owner-liveness-aware lease and a token-safe atomic reclaim/release protocol rather than age-only recursive deletion. [R4 / `packages/client/src/runtime/managed-skills.ts:497`]

❌ 4. Managed Skill reconciliation now runs before the existing legacy runtime-layout migration. A supported pre-directory workspace can still have `.first-tree-workspace` as a file; `acquireWorkspaceLock` then fails while creating `.first-tree-workspace/managed-skills.lock`, the handler ignores the failed result, `ensureAgentBootstrap` converts the marker to a directory only afterward, and the first provider turn starts without Core Skills. Run the narrow agent-home runtime-layout convergence before reconciliation (without applying the broader bootstrap to the intentional legacy per-chat Claude resume path), and cover this upgrade ordering in a handler-level test. [R5 / `packages/client/src/handlers/claude-code.ts:2176`]

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: Three of the four prior blockers are addressed, but stale-lock reclamation can still detach a newly acquired live lock and expose a concurrent-writer window.

**Risk level**: A
- Semantic lift: none (already at the highest path level)

**PR summary**
- Problem: Managed agents need one crash-safe, provider-native projection of Core and Team Skills across provider switches, stale configuration, user conflicts, and upgrades.
- Approach: The follow-up now rejects symlinked managed roots, promotes failed journal recovery to a workspace-fatal error, makes live-owner checks independent of lock age, atomically quarantines locks before release, and converges the legacy runtime marker before first reconciliation.
- Impacted modules: Client managed-skill state/recovery, workspace acquisition, and the corresponding reconciler and handler-level regressions.

**Review findings**
❌ 1. Reclamation still has an inspect/rename TOCTOU that can detach a successor's live lock. Two contenders can both inspect dead owner A. Contender B quarantines A, another contender publishes live owner D at `lockPath`, and contender C then executes its already-authorized `rename(lockPath, quarantinePath)`, moving D instead of A. C detects the token mismatch only after `lockPath` is absent; before `restoreQuarantinedLock` completes, B or another candidate can acquire the empty path. If that acquisition wins the restore race, D continues its reconciliation from quarantine while the new owner also mutates the shared state/journal. The token check detects the ownership change but does not prevent the double-owner interval. Please make reclaim ownership verification and detachment mutually exclusive with candidate publication (or use an OS-backed lock primitive with equivalent atomic ownership), and cover the three-contender owner-replacement interleaving. [R4 / `packages/client/src/runtime/managed-skills.ts:528`]

✅ The prior symlink-ancestor, fatal journal-recovery, and legacy-marker ordering findings are addressed by the new root preflight, `ManagedSkillsFatalError` propagation, and `acquireAgentHome` migration boundary.

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The kernel-backed lock closes the prior ownership race, but the native addon is currently bundled without its loadable binary and makes the shipped CLI fail during module initialization.

**Risk level**: A
- Semantic lift: none (already at the highest path level)

**PR summary**
- Problem: Managed agents need one crash-safe, provider-native projection of Core and Team Skills without allowing lock-owner recovery races to create concurrent state/journal writers.
- Approach: Replace the directory/token recovery protocol with an advisory OS lock held on one persistent regular-file inode; the OS releases ownership when the descriptor or process exits.
- Impacted modules: Client managed-skill reconciliation, a new native file-lock adapter, CLI/portable dependency delivery, runtime-state documentation, and cross-process regressions.

**Review findings**
❌ 1. `fs-native-extensions` is added only to the private Client package, so the CLI bundler inlines its CommonJS loader but does not ship a loadable native addon. The built CLI now fails at module initialization before any command can run: the portable smoke's `first-tree --version` exits with `ReferenceError: __filename is not defined in ES module scope`, while the built-dist package test reports `ADDON_NOT_FOUND` from the generated `connect-token-*.mjs`. This also breaks ordinary CLI and Client/Web jobs that load that bundle. Please establish one release-safe native dependency boundary: keep the addon external, declare/install it in the published `apps/cli` and portable dependency manifests (including supported platform prebuilds), or provide an equivalent bundler-native packaging path; then make the built npm/portable entry smoke prove the addon loads before release. [R5 / `packages/client/package.json:37`, `apps/cli/package.json:61`]

⚠️ 2. The runtime-state contract now says `managed-skills.lock` is persistent, but the next paragraph still says “The lock and journal are absent after a settled reconcile.” Please update that sentence so operators do not treat the expected persistent inode as stale state. [R1 / `docs/development/agent-workspace-state.md:52`]

✅ The persistent-inode kernel lock removes stale-owner reclamation entirely and addresses the prior three-contender ownership race.

**Action taken**
- Submitted request changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All previously blocking filesystem-boundary, recovery, lock-ownership, migration-order, and native-addon delivery issues are resolved at the reviewed head.

**Risk level**: A
- Semantic lift: none (already at the highest path level)

**PR summary**
- Problem: Managed agents need Core and Team Skills projected into only the active provider's native discovery root, without corrupting state under concurrent starts, crashes, stale configuration, provider switches, or user-name conflicts.
- Approach: Reconcile Core and Team Skills through one Client runtime path with a monotonic Team version fence, final-tree digests, ownership markers, a persistent-inode kernel lock, journaled swaps, deterministic recovery, and conservative legacy migration.
- Impacted modules: Client provider startup and managed-skill reconciliation, CLI/npm/portable runtime dependency delivery, legacy Claude resume, briefing inputs, runtime-state documentation, and reusable QA.

**Review findings**
✅ The provider-root ancestor checks, fail-closed journal recovery, and legacy runtime-marker migration order close the original filesystem and first-turn gaps.
✅ The persistent-inode OS lock removes stale-owner reclamation and its three-contender ownership race.
✅ `fs-native-extensions` is now an external published CLI dependency loaded through `createRequire(import.meta.url)`; portable artifacts require the target prebuild before archiving and verify it again before launching the packaged entry.
✅ The runtime-state documentation now correctly distinguishes the removed settled journal from the persistent, unlocked lock file.
✅ All required checks pass on `34300ba1094b33545f544d4c07a014b976b1a769`, including CLI, Client/Web, portable smoke, lint/typecheck, aggregate tests, and CodeQL.

**Action taken**
- Approved exact head `34300ba1094b33545f544d4c07a014b976b1a769` with explicit authorization from baixiaohang.

PR #2062, opened 2026-07-29
feat(resources): store complete Team Skill bundles

yuezengwu:
## Summary

- Persist each Team Skill as one complete immutable bundle attachment in S3-compatible object storage.
- Derive the Resource name, description, and compatibility instruction projection from validated `SKILL.md`; bundle replacement keeps the same Resource and existing agent choices.
- Let admins paste `SKILL.md`, upload a ZIP/Markdown file, or choose a complete Skill folder.

## Design boundary

- Uses `resources.bundle_attachment_id` without a foreign key and adds no version or per-file tables.
- Keeps runtime installation/loading unchanged for the separately scoped Work 1.
- Streams attachment reads and writes, enforces organization quotas, cleans unreferenced objects, and backfills legacy PostgreSQL attachment bytes and inline Skill payloads.

## Validation

- `pnpm check` (passes; existing unrelated warnings remain)
- `pnpm typecheck`
- Web full suite: 209 files / 1,776 tests
- Server full suite: 250 files / 2,707 tests
- Post-rebase focused suites: 83 tests
- `docker compose config --quiet`

## Context Tree

- Companion draft PR: https://github.com/agent-team-foundation/first-tree-context/pull/845


baixiaohang (changes requested):
- Rationale: the bundle model is directionally sound, but the new attachment lifecycle is not yet safe under normal multi-replica and concurrent-reference traffic, and two storage-operability requirements are missing.

**Risk level**: B-high
- Path baseline: includes a database migration/schema change and core `packages/server` resource/attachment services -> B-high
- Semantic lift: none

**PR summary**
- Problem: admins can currently save only an inline `SKILL.md`, so a Team Skill loses supporting scripts/assets and cannot be supplied as a complete reusable folder.
- Approach: store each complete Skill as an immutable ZIP attachment in S3-compatible storage, derive the catalog/runtime compatibility projection from validated `SKILL.md`, preserve the Resource identity on replacement, and add paste/file/folder upload paths plus rolling legacy backfills.
- Impacted modules: attachment storage/lifecycle and migrations, Team Resource resolution, Skill bundle validation, server configuration, and the Web resource editor.

**Review findings**

❌ 1. The legacy attachment backfill can turn a successfully migrated blob into an unreadable row when two server replicas process the same row. Both replicas may select the same non-null `data`; one can successfully upload and clear `data`, while the other's unconditional failure update clears `objectKey`. The final row is then `ready` with both `data` and `objectKey` null, while the successful object is stranded in S3. This is especially likely during a rolling deploy because every replica runs the backfill in `onReady`. Claim each row/CAS the complete attempt (including failure recovery), and cover a success-vs-failure interleaving. [R4 / `packages/server/src/services/attachment.ts:339`]

❌ 2. Reference validation and orphan deletion do not share a lock, so cleanup can commit a dangling message reference. `validateMessageAttachmentRefs()` performs a plain ready-row read inside the message transaction; before the later message insert, the sweeper can lock that attachment, observe no reference, mark/delete it, and then let the message commit. Lock the referenced attachment rows through the message insert (or introduce an equivalent atomic reference claim) so either the message wins and cleanup observes it, or cleanup wins and validation fails. [R4 / `packages/server/src/services/doc-snapshots.ts:78`, `packages/server/src/services/attachment.ts:267`]

❌ 3. The upload path has byte/object quotas but no bound on a single caller's simultaneously active upload streams. The current Attachment contract explicitly requires per-client parallel streams to be bounded so aggregate concurrency cannot recreate the pressure that streaming is meant to avoid; the request-per-minute limiter and short quota advisory lock do not cap active streams after reservation. Add a fail-closed per-caller concurrency guard that works with the deployment's multi-replica model and a deterministic concurrency test. [R5 / `packages/server/src/api/orgs/attachments.ts:39`, `packages/server/src/services/attachment.ts:36`]

❌ 4. The hourly orphan sweep is unbounded against table growth: candidate selection has no applicable `updated_at` index, and each of up to 100 candidates runs a separate JSON reference query over `messages` with no attachment-reference index. In the common orphan case, this becomes roughly 100 full message-table scans per hour. Give both candidate selection and reference discovery an indexed/batched path (or a normalized reference lookup) before enabling the periodic sweep. [R3 / `packages/server/src/services/attachment.ts:231`, `packages/server/src/services/attachment.ts:306`]

✅ The ZIP validation itself is appropriately fail-closed on traversal, duplicate/case-colliding paths, symlinks, encryption, entry count, and uncompressed-size limits, and it derives the public projection from the single `SKILL.md` source.

**Action taken**
- Submitted request changes on head `f28229eabc17f73429068f0bcf66fd4983508665`.

baixiaohang (changes requested):
- Rationale: the prior concurrency hardening is effective, but orphan cleanup can still race a fresh backfill claim, and not every attachment writer participates in the reference-lock lifecycle.

**Risk level**: B-high
- Path baseline: includes database migrations/schema changes and core server attachment/resource services
- Semantic lift: none

**PR summary**
- Problem: admins need Team Skills to retain their complete directory—scripts, assets, and `SKILL.md`—instead of storing only an inline compatibility projection.
- Approach: upload immutable ZIP bundles to S3-compatible storage, derive Resource metadata from validated `SKILL.md`, preserve Resource identity across replacement, and migrate legacy attachment bytes and inline Skills through bounded background work.
- Impacted modules: attachment storage/lifecycle and migrations, Team Resource resolution, Skill bundle validation, server configuration, and the Web resource editor.

**Review findings**

❌ 1. The orphan sweep acts on stale eligibility. It selects only an id outside the deletion transaction, then `deleteAttachmentIfUnreferenced()` locks the row but never re-checks the candidate state/timestamp. A second replica can claim that old legacy row for backfill after selection, refresh `updatedAt`, and start the S3 write; the sweeper can then delete the newly claimed row, after which the backfill completion CAS reports `skipped` and leaves the just-written object with no lifecycle row. Revalidate the sweep cutoff/state (or an expected timestamp/version) under the same row lock before marking `deleting`, or claim sweep rows transactionally, and cover this exact sweep-vs-backfill interleaving. [R4 / `packages/server/src/services/attachment.ts:354`]

❌ 2. The new reference lock covers only `metadata.attachments[]`, while cleanup also treats single-image `content.imageId` and batched `content.attachments[].imageId` as attachment consumers. `validateFileContent()` checks those refs' JSON shape but never loads or locks their attachment rows, so a file message can store an unknown id, or the sweeper can delete an older ready attachment after preflight but before the message insert commits. Validate and hold the same cleanup-conflicting locks for both file-content shapes inside the message transaction. [R5 / `packages/server/src/services/message.ts:93`, `packages/server/src/services/message.ts:731`]

❌ 3. A failed upload can create an object the sweeper can never discover. In `createAttachment()`'s catch path, an object-store delete failure is discarded and the lifecycle row is deleted anyway; any object that was already accepted by S3 is then permanently untracked because the orphan sweep starts from database rows. Preserve a `deleting` row until object deletion succeeds, as `deleteAttachmentIfUnreferenced()` already does for normal cleanup. [R4 / `packages/server/src/services/attachment.ts:107`]

✅ The new backfill CAS ownership, shared active-upload bound, indexed lifecycle/reference lookups, and `metadata.attachments[]` row locks address the concrete issues raised on the previous head.

**Action taken**
- Submitted request changes on head `c8ee73ccca2c544ac5876ad5f86821d85f2bdb00`.

baixiaohang (changes requested):
- Rationale: the two attachment-row lifecycle races are now fixed, but the new file-message validation changes a decision-locked legacy contract and the existing edit path still does not release its last attachment reference immediately.

**Risk level**: B-high
- Path baseline: includes database migrations/schema changes and the core message/attachment services
- Semantic lift: none

**PR summary**
- Problem: admins need Team Skills to preserve their complete directory—`SKILL.md`, scripts, and assets—while agents keep the same Resource identity and selections when a bundle is replaced.
- Approach: store immutable ZIP bundles in S3-compatible storage, derive the Resource projection from validated `SKILL.md`, and migrate legacy blobs/Skills through quota-governed streaming attachment lifecycle work.
- Impacted modules: attachment storage/lifecycle and migrations, message attachment references, Team Resource resolution, Skill bundle validation, server configuration, and the Web resource editor.

**Review findings**

❌ 1. Correction to my previous review: asking for strict stored-row validation of legacy `format: "file"` image refs was overbroad. The decision-locked attachment contract in `system/cloud/chat/attachments.md` explicitly preserves ordinary/historical file-content refs as shape-only (allowed MIME) and reserves blob existence/MIME/size validation for generic `metadata.attachments[]`. `validateFileAttachmentRefs()` now rejects a missing row or stored MIME/size mismatch on both send and edit, which changes that compatibility boundary and can make an otherwise valid historical image message uneditable. Remove the strict blob validation for the legacy shape. If lifecycle coordination is still desired for an existing row, an optional lock that does not change shape-only acceptance would preserve the contract. [R5 / `packages/server/src/services/message.ts:108`]

❌ 2. `editMessage()` can remove the last legacy image reference by changing a file message's content or format, but it never releases the old single/batch attachment ids after the edit commits. The attachment contract requires immediate deletion when the last reference is removed; today the old object remains until the periodic orphan sweep notices it. Capture the pre-edit file-content ids, then run reference-aware cleanup after the durable edit (letting cleanup retain/retry a `deleting` row on object-store failure), and cover replacement plus file-to-text edits. [R4 / `packages/server/src/services/message.ts:1108`]

✅ The sweep now revalidates state/cutoff under the row lock, so a fresh backfill claim cannot be deleted from a stale candidate list.

✅ Failed-upload cleanup now retains a lifecycle row until object deletion succeeds, making retries discoverable to the bounded sweeper.

**Action taken**
- Submitted request changes on head `0464d97fb74fb535cafa4dd9b2256a80d3bf7dbc`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: all previously identified concurrency, lifecycle, cleanup, and legacy-compatibility blockers are fixed on the reviewed head, and the full CI suite is green.

**Risk level**: B-high
- Path baseline: includes database migrations/schema changes and the core message/attachment services
- Semantic lift: none

**PR summary**
- Problem: admins need Team Skills to preserve their complete directory—`SKILL.md`, scripts, and assets—while agents keep the same Resource identity and selections when a bundle is replaced.
- Approach: store immutable ZIP bundles in S3-compatible storage, derive the Resource projection from validated `SKILL.md`, and migrate legacy blobs/Skills through quota-governed streaming attachment lifecycle work.
- Impacted modules: attachment storage/lifecycle and migrations, message attachment references, Team Resource resolution, Skill bundle validation, server configuration, and the Web resource editor.

**Review findings**

✅ Legacy single/batch image references retain their decision-locked shape-only compatibility contract, while currently resolvable rows take cleanup-conflicting locks until message commit.

✅ Message edits now release replaced or removed legacy attachment IDs after the durable edit, recheck all consumers, and preserve a retryable lifecycle row when object deletion fails.

✅ The earlier backfill CAS, bounded-upload, orphan-query/index, stale-sweep, and failed-cleanup findings are resolved; all current GitHub checks pass.

**Action taken**
- Approved exact head `9c0f95076d0b2687e06bfa3ca3c774dcc031d337`.

PR #2064, opened 2026-07-29
fix(cli): report external context hook and activation state

baixiaohang:
## Summary

- read Codex SessionStart Hook trust and enabled state from the provider instead of always reporting review required
- add step-by-step, post-install Hook consent guidance with a reliable status reread
- separate provider, Plugin, Hook, exact checkout binding, and live activation status
- distinguish signed-out, non-Git, unreadable origin, and unsupported origin checkout failures with channel-aware recovery commands
- document the flow and add deterministic tests plus a cross-surface QA case

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] real Codex app-server Hook smoke check returned trusted + enabled
- [x] `first-tree-staging` exact Context activation and relevant Context Tree read

## Change Surface

- [x] `apps/cli` public CLI or help output
- [x] tree onboarding / binding / inspection behavior
- [x] docs or contributor-facing repository metadata

## Notes

- package or install behavior changes: no consent bypass; Codex still owns Hook trust and enablement
- docs or tests updated to match: CLI reference, integration guide, provider/status tests, and a QA case
- follow-up work: formal QA is warranted for clean-repo comparison, Claude Code, Write/PR, and cross-OS release qualification; the earlier GitLab SSH port 22 timeout is not treated as a generic Read defect

### DIFF
--- a/apps/cli/src/__tests__/context-account-state-guard.test.ts
+++ b/apps/cli/src/__tests__/context-account-state-guard.test.ts
@@ -58,6 +58,9 @@ describe("Context/client-switch account-state interlock", () => {
       probe: () => {
         throw new Error("must not probe");
       },
+      inspectHook: async () => {
+        throw new Error("must not inspect");
+      },
       validateMarketplace: () => undefined,
       install: () => {
         throw new Error("must not install");
--- a/apps/cli/src/__tests__/context-bundle-manifest.test.ts
+++ b/apps/cli/src/__tests__/context-bundle-manifest.test.ts
@@ -166,7 +166,6 @@ describe("context integration bundle", () => {
       installed: true,
       enabled: true,
       installedPath,
-      hookTrust: "provider_managed",
       issues: [],
     };
     const installedMarketplaces: string[] = [];
@@ -175,6 +174,7 @@ describe("context integration bundle", () => {
       executable: "codex",
       minimumVersion: "0.144.1",
       probe: () => probe,
+      inspectHook: async () => ({ trust: "trusted", enabled: true, source: "provider_api", issues: [] }),
       validateMarketplace: () => undefined,
       install: ({ marketplaceRoot }) => {
         installedMarketplaces.push(marketplaceRoot);
@@ -213,9 +213,9 @@ describe("context integration bundle", () => {
         installed: true,
         enabled: true,
         installedPath: "/provider/cache/first-tree-context",
-        hookTrust: "review_required",
         issues: [],
       }),
+      inspectHook: async () => ({ trust: "trusted", enabled: true, source: "provider_api", issues: [] }),
       validateMarketplace: () => undefined,
       install: () => {
         throw new Error("not used");
@@ -260,7 +260,6 @@ describe("context integration bundle", () => {
       installed: false,
       enabled: false,
       installedPath: null,
-      hookTrust: "provider_managed",
       issues: [],
     };
     const driver: ContextIntegrationProviderDriver = {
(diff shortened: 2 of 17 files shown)


yuezengwu (changes requested):
## Changes requested

Goal and core change: this PR reads Codex's provider-owned SessionStart trust/enabled state and replaces the previous combined probe with distinct provider, Plugin, Hook, checkout, exact-binding, and live-activation status models. That separation matches the external Context contract, but two status paths currently report incorrect user-facing state:

1. **Blocking — effective provider minimum and `compatible` can disagree.** In `apps/cli/src/core/context-integration/status.ts:128-134`, `minimumVersion` comes from the embedded release manifest, while `compatible` is copied from `plugin.compatible`, which the driver calculated against its own default minimum. If a release raises its provider minimum (the release manifest is the authoritative lifecycle gate), status can render `Provider compatible: Yes — 0.x (requires a newer 0.y+)` even though runtime health correctly requires an upgrade. Compute compatibility against the same effective minimum displayed here, and cover the stricter-release-minimum case.

2. **Blocking — Claude Code receives a Codex-only Hook recovery action.** `ClaudeCodeContextIntegrationDriver.inspectHook()` reports the provider-managed Hook as disabled when an installed Claude Plugin is disabled, but `apps/cli/src/commands/context/status.ts:70-73` turns every `enabled: false` into `enable ... in Codex /hooks`. Thus `context status --provider claude-code` directs the user to an unrelated provider. The no-Plugin Claude case also reports Hook trust as provider-managed even though no Hook exists. Make Hook rendering/provider state source-aware (provider-managed vs Codex API), and add Claude disabled/not-installed status coverage. The Codex enable guidance should likewise preserve the already-trusted-but-disabled case instead of instructing the user to choose Trust again.

No database change is present. The core status data model does change (`ProviderPluginProbe` is split from `ProviderHookProbe`, plus a new layered `ContextIntegrationStatus`), so that public JSON shape and downstream consumers need explicit human attention.

Reviewed head `d8d4163763f4f2374957c57822b1f99e52793abb`. Per reviewer instructions, I inspected the diff and existing validation evidence without running additional tests or QA.

baixiaohang:
Addressed both blocking findings in `6a0361892`:

1. Provider compatibility is now recomputed against the effective minimum from the release manifest, with a regression case where the release minimum is stricter than the driver default.
2. Hook presentation is now source-aware: disabled Claude Hooks point to the Claude Plugin, absent Claude Plugins report Hook state as unavailable, and trusted-but-disabled Codex Hooks request checkbox enablement without asking for Trust again. Added Claude disabled/absent and Codex trusted/disabled coverage.

Validation:
- targeted Context tests: 19 passed
- full CLI test suite: all 23 batches passed (1 existing skip)
- CLI typecheck passed

The layered JSON status shape remains an intentional compatibility-sensitive change in this PR and is called out in the PR description/review context.

yuezengwu (approved):
Goal and core change: this PR reads Codex's provider-owned SessionStart Hook trust/enabled state and separates provider, Plugin, Hook, checkout, exact-binding, and live-activation status.

Reviewed successor head `6a0361892593ee3dc55cbee2342617f30f1c612d`. Both prior blockers are resolved:

- provider compatibility is recomputed against the same effective release-manifest minimum that status displays, with stricter-release-minimum coverage;
- Hook state and recovery are provider/source-aware, including Claude absent/disabled cases and Codex trusted-but-disabled guidance without a redundant Trust step.

The full successor diff introduced no new blocker. No database change is present. The core status data model and layered public JSON shape remain compatibility-sensitive and should receive the human attention already noted in the PR/review context.

Per reviewer instructions, I inspected the diff and the author's validation evidence without running additional tests or QA.

baixiaohang:
Context Tree follow-up: https://github.com/agent-team-foundation/first-tree-context/pull/848

PR #849, opened 2026-06-05
refactor(web): quiet per-section resource add control to a ghost + icon

Gandy2025:
## Why
Follow-up to #847. The per-section add control rendered as a **bordered `+ <Type>` text button**. With 3–4 sections (plus row actions like `Disable`), a page read as a column of loud bordered buttons — the "too many buttons" feeling.

## Change
Downgrade both **shared** entry points to a quiet **`ghost` "+" icon**:
- **Settings → Resources** (`AddResourceButton`)
- **Agent → Capabilities** (`AddCapabilityMenu` trigger)

The resource type is carried by `aria-label` + `title` tooltip (`Add Repo` / `Add Skill` / `Add MCP` / `Add Instructions`) and by the icon sitting beside its section title — discoverability holds, visual weight drops. Both pages stay consistent (one shared shape).

No behaviour/structure change: per-section entry is still always present (empty sections — e.g. MCP — keep their in-context affordance, no dead end), admin/`canEdit` gating unchanged.

## Test
- `pnpm --filter @first-tree/web typecheck` ✅ (incl. design-token guardrails)
- `pnpm --filter @first-tree/web test` ✅ **686/686**
- biome clean
- Tests updated to locate the icon-only control by `aria-label` instead of button text.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2071, opened 2026-07-29
fix(resources): keep attachment payloads in PostgreSQL

yuezengwu:
## Summary

- store every new attachment and Team Skill bundle in `attachments.data` (`bytea`)
- install a database fence that prevents pre-transition replicas from moving PostgreSQL payloads back to S3
- keep a temporary PostgreSQL-first/S3-fallback read path for rows written by #2062
- reverse-backfill S3-only rows into PostgreSQL with compare-and-set updates while retaining the old pointer/copy for rolling-deploy compatibility
- keep attachment quotas, concurrent-upload reservations, reference-aware deletion, and orphan cleanup

## Root cause

#2062 coupled complete Team Skill bundles to a new S3-compatible attachment backend even though Cloud already has a PostgreSQL-backed attachment primitive. That introduced a second durable persistence, backup, deployment, and local-development boundary unnecessarily.

## Impact

PostgreSQL is authoritative for all new payloads. During the expand phase, existing S3 configuration remains available only so old and new Server replicas can overlap safely and S3-only rows remain readable while they are copied back. The S3 dependency, compatibility pointers, and old copies can be removed in a later contract release after every pre-transition replica has drained and no `data IS NULL` attachment remains.

## Context Tree

Depends on agent-team-foundation/first-tree-context#849, which updates the locked Cloud storage decision.

## Validation

- `pnpm check`
- `pnpm typecheck`
- full Server suite: 250 files / 2717 tests
- focused attachment, Team Skill bundle, resources, messaging, and open-question suite: 91 tests
- migration-backed tests for a pre-existing S3-only row, reverse-backfill CAS, and the old/new replica interleaving fence
- `docker compose config --quiet`
- `git diff --check`


baixiaohang (changes requested):
- Rationale: returning attachment payloads to PostgreSQL simplifies the deployment boundary, but this head is not safe for existing S3-only rows or a rolling deployment with the current main branch.

**Risk level**: B-high
- Path baseline: changes attachment database contents/state semantics and core `packages/server` attachment, message, and Resource services.
- Semantic lift: none.

**PR summary**
- Problem: Team admins need complete Skill bundles and other attachments to remain available without requiring a second S3/MinIO persistence and backup boundary.
- Approach: make `attachments.data` authoritative for new writes, buffer each bounded upload before publishing it atomically in PostgreSQL, and remove the S3 client, configuration, and local services.
- Impacted modules: attachment persistence/lifecycle, message and Resource attachment consumers, Team Skill bundle validation/backfill, Server configuration, and local deployment packaging.

**Review findings**

❌ 1. The storage contraction strands existing S3-only payloads and can create more of them during a rolling deployment. This head makes both metadata/reference lookup require `data IS NOT NULL` and removes the `objectKey` read fallback, so every ready row produced by #2062 with `data = NULL` immediately becomes a 404 or an unusable Team Skill bundle. More importantly, an old replica can coexist with this version and run main's `backfillLegacyAttachments()`, which selects a new PostgreSQL-backed row (`data IS NOT NULL`, `object_key IS NULL`), uploads it to S3, and clears `data`; once the old replica drains, the new replicas can no longer read that attachment. Keeping `object_key` as an offline recovery pointer does not preserve the user-visible contract. Please use an expand/contract sequence that keeps dual-read compatibility, stops or version-gates the old PostgreSQL-to-S3 backfill, migrates S3-only rows back into PostgreSQL with a CAS, and removes the S3 dependency only after old replicas and `data IS NULL` rows are gone. Cover both a pre-existing S3-ready row and the old/new-replica interleaving. [R4/R5 / `packages/server/src/services/attachment.ts:208`, `packages/server/src/services/attachment.ts:232`, `packages/server/src/services/attachment.ts:253`]

✅ The new-write path keeps the per-file and per-caller bounds and publishes metadata plus payload in one PostgreSQL transaction once the transition-compatibility gap is closed.

**Action taken**
- Submitted request changes on head `653961b205f69d48ba3048aa2bfb16ec579a1c76`.

yuezengwu:
已按本轮反馈在 `ea034330d` 完成 expand/contract 修正，并重新请求复审：

- 保留 legacy S3-only 行的双读兼容，新写入仍只落 PostgreSQL。
- 新增数据库 fence migration，阻止旧副本执行 PostgreSQL→S3 claim/清空 `data`。
- 新增 S3→PostgreSQL 的有界 CAS 回填；兼容期保留 `object_key`/对象副本供旧副本读取。
- 删除流程兼容清理 legacy 对象，失败时保留 `deleting` 生命周期状态。
- 覆盖既有 S3-ready 行、旧/新副本交错与 CAS 竞争测试。

新 head 的 15 项 CI 均已通过，0 失败（2 项条件跳过）。

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the new expand phase closes the rolling-deployment and existing-row availability blocker, and the latest head has a fully green CI result.

**Risk level**: B-high
- Path baseline: changes attachment database contents/state semantics and core `packages/server` attachment, message, and Resource services.
- Semantic lift: none.

**PR summary**
- Problem: Team admins need complete Skill bundles and other attachments to remain available while Cloud returns to one PostgreSQL persistence and backup boundary.
- Approach: make PostgreSQL authoritative for new payloads, install a database fence against the old PostgreSQL-to-S3 backfill, dual-read transitional S3 rows, and reverse-backfill them into PostgreSQL with compare-and-set updates while retaining the old copy for pre-transition replicas.
- Impacted modules: attachment persistence/lifecycle and migration, message and Resource attachment consumers, Team Skill bundle validation/backfill, Server configuration, and local deployment packaging.

**Review findings**

✅ The migration trigger blocks both the old claim that would assign `object_key` to a PostgreSQL-backed row and the later attempt to clear an already-present payload, so the fence applies across all replicas through the shared database.

✅ S3-only ready rows remain readable while the reverse backfill copies them into `data`; the CAS preserves a concurrent winner, and retaining `object_key` plus the S3 copy keeps pre-transition replicas compatible until the later contract release.

✅ Reference-aware deletion still locks against new consumers and keeps a retryable `deleting` row until the transitional S3 copy is removed.

⚠️ Companion Context Tree PR #849 still needs its reviewer-requested wording correction: the implementation bounds buffering per caller and per file, not globally. The source head describes and implements that narrower guarantee, so this remains a Tree follow-up rather than a code blocker.

**Action taken**
- Approved head `ea034330da1d13dc7936c2c59bf00c4aa6cf43f3`.

PR #2078, opened 2026-07-29
feat(runtime): project complete Team Skill bundles

yuezengwu:
## Summary

- add an optional immutable ZIP bundle descriptor to the runtime Team Skill wire shape while retaining the derived inline fields for rolling compatibility
- project only ready, same-organization attachment metadata from the Server and fail closed when a bundle-backed Resource cannot resolve its authoritative attachment
- share one portable ZIP/name/manifest contract between Server admission and Client materialization so a configured bundle is statically projectable on every supported provider
- resolve bundles through the authenticated attachment API, safely materialize complete Skill directories into every provider-native root, and preserve the existing locks, journal recovery, full-tree digest, monotonic fence, collision policy, and clean last-known-good behavior
- block provider starts/turns when a ledger-owned discovery target cannot be verified or atomically removed, while keeping successful quarantine bounded outside provider discovery
- extend deterministic coverage and the managed Skill formal QA case for complete attachment/reference/script behavior

## Runtime behavior

The complete attachment bundle is authoritative whenever the optional descriptor is present. New Clients never reconstruct an inline-only fallback for such a Skill; legacy Resources without bundles remain inline, and older Clients can continue consuming the retained inline fields during a rolling Server/Client deployment. After a new Client has persisted an `attachment:` revision for a Resource, that state is also one-way evidence of bundle authority: an equal or newer mixed-version snapshot from an older Server that omits the descriptor preserves only the digest-verified complete bundle LKG and cannot downgrade it to inline content. A later descriptor resumes normal replacement, while an explicit authoritative empty snapshot still revokes the Skill.

ZIP projection accepts an exact `SKILL.md` at the root or under one wrapper directory. It preserves binary assets, normalizes owner-safe executable/read modes, always rewrites only the manifest `name` to the allocated provider target, and rejects traversal, absolute/backslash/NUL paths, Unicode-normalized conservative case-fold collisions (including expansion folds such as `Straße` / `STRASSE`, `ẞ` / `SS`, and `ſkill` / `skill`) across complete paths, implicit ancestors, on-disk target allocation, copied directories, and final-tree digest verification, encryption, symlinks, special files, invalid wrapper anchors, reserved marker injection, ambiguous manifests, Windows-reserved/non-portable or overlong raw and normalized paths, and shared entry/file/materialized-size/depth limits before atomic exchange. Server and Client also share an exact YAML frontmatter envelope with alias-free, finite JSON-safe metadata, so permissive delimiters, whitespace variants, cyclic aliases, non-finite numbers, tagged binary/set/ordered-map values, unresolved tags, or persistence-hostile values fail at admission.

The same shared target-slug gate applies during Resource configuration and Client allocation, including the 63-character provider limit, Windows device names (including superscript device-number aliases), and normalized First Tree Core reservations. Runtime projection also requires a ready attachment with a PostgreSQL payload or legacy object pointer and a valid server-measured size; unavailable rows emit neither the descriptor nor an inline fallback.

If an installed target still matches its recorded full-tree digest, a transient replacement download failure preserves that clean last-known-good directory. Unavailable and stale snapshots also verify every ledger-owned Team target in the active provider root before preserving it; a tampered LKG is quarantined even when the control plane cannot supply a new bundle. If the target cannot be verified or has drifted, it is atomically moved to a stable bounded quarantine slot outside the provider-native discovery root before repair. A universal final provider-publication gate re-verifies every ledger-owned Core and Team target in the active provider root, including preserved unavailable/stale Team LKGs, so drift introduced after the main loop is quarantined and withdrawn from installed/skipped/briefing output; failure to move it blocks the turn. Invalid or future state/journal authority, provider-root enumeration ambiguity, and unrecoverable transaction filesystem errors also block when provider discovery may still contain managed content.

The raw workspace-lock API retains its typed timeout. At reconcile time, however, lock contention has not proven that another writer reached the universal publication gate: a timeout is therefore an ordinary availability failure only when the active provider discovery root is provably empty. If the root contains managed content or cannot be safely inspected, the same timeout raises the unsafe-discovery signal and blocks Provider preflight. Real cross-process contention tests cover the raw typed error and both reconcile branches, closing maintainer review finding 4810353309.

If verification, quarantine, or authoritative removal cannot prove that the ledger-owned target left discovery, the reconciler raises a dedicated unsafe-discovery error and every provider start/active-turn path fails closed. A successful quarantine followed by a transient bundle failure may continue without that Skill; interruption immediately after the move remains safe because the target is already outside discovery, and the next reconcile either cleans the stable slot while restoring the authoritative bundle or clears it before persisting an authoritative revocation. Repeated drift and repair reuses and removes the same stable slot, preventing unbounded quarantine growth.

The legacy installed-digest byte format remains unchanged for rolling compatibility. During the same complete-tree traversal on POSIX platforms, every real file and directory—including the target root—is independently rejected if group/other write bits are present. Permission-only tampering such as `0777` therefore triggers quarantine/repair without forcing healthy pre-upgrade state digests to migrate. Windows skips this POSIX-bit gate because Node reports synthesized mode bits there rather than ACL write authority; a full fresh-install plus clean-LKG reconcile regression verifies that behavior. Newly copied Core directories use explicit owner-safe modes rather than inheriting the host umask.

The implementation uses First Tree's existing PostgreSQL attachment bytes and authenticated download route; it does not introduce another blob store or persistence path.

Multica's immutable bundle reference / resolve / validate / materialize architecture at `9e3b661d4494cfe5a66c5fdabcc24b14890d1eda` was design inspiration only. No Multica implementation was copied.

## Verification

- `pnpm --filter @first-tree/shared exec vitest run src/__tests__/skill-schema.test.ts` — 14/14 passed
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/skill-bundles.test.ts` — 41/41 passed
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/managed-skills.test.ts src/__tests__/workspace-file-lock.test.ts` — 105/105 passed (Managed Skills 101/101; workspace lock 4/4)
- Shared, Server, and Client package typechecks passed
- scoped Biome checks and `git diff --check` passed
- `pnpm check` — passed (16 existing warnings and 1 info)
- `pnpm typecheck` — 11/11 Turbo tasks passed in 10.573s
- `pnpm test` — 12/12 Turbo tasks passed in 2m14.355s
- Server full suite — 250 files / 2754 tests passed
- Shared full suite — 66 files / 734 tests passed
- Client full suite — 156 files passed, 1 skipped / 1770 tests passed, 3 skipped
- Web full suite — 209 files / 1796 tests passed

Formal provider QA remains warranted because this crosses Cloud attachment projection, authenticated Client download, provider discovery roots, and inject-time settling. The updated QA case now requires a real Agent to read a bundled binary asset, consult a nested reference, and execute the bundled script.


baixiaohang (changes requested):
- Rationale: The bundle authority and rolling-compatibility shape are sound, but two fail-closed gaps still allow an uploaded archive to exhaust local resources or a drifted managed Skill to remain executable.

**Risk level**: A

**PR summary**
- Problem: Team admins can upload complete Skills, but runtime agents currently receive only the derived inline `SKILL.md`, so supporting scripts, references, and binary assets are unavailable to provider-native Skill discovery.
- Approach: Add an optional attachment descriptor for rolling compatibility, resolve authoritative bundles through the authenticated attachment API, and atomically project complete Skill directories into each provider's native root while retaining the existing reconciliation fence and recovery model.
- Impacted modules: Shared runtime schemas, Server effective-resource projection, Client managed-Skill reconciliation and all runtime handlers, plus deterministic tests and the formal managed-Skill QA case.

**Review findings**
❌ 1. Directory entries bypass the archive entry limit. `fileCount` is incremented only inside `kind === "file"`, while every directory is retained in `rawEntries` and later materialized with `mkdir`/`chmod`. A compressed attachment within the ordinary upload limit can therefore contain tens of thousands of shallow empty directories, consuming memory/inodes and holding the workspace writer lock despite the advertised bounded extraction. Please enforce a bounded total-entry/directory count at both Server admission and Client extraction, with a directory-flood regression. [R4 / `packages/client/src/runtime/managed-skills.ts:1318`]

❌ 2. A transient bundle download failure leaves a drifted managed directory executable. When `actualDigest !== existing.installedDigest`, reconciliation attempts to restage; if the resolver throws, the catch records a failure but leaves the modified target in the provider-native discovery root. The handlers then discard `result.ok` and continue starting the provider, so the provider can discover and execute a tampered script even though the Skill is omitted from the generated briefing. Preserve/restore a locally verified last-known-good copy, quarantine the unverified target, or fail provider preflight for this case, and cover the combined “digest drift + resolver failure” path. [R4 / `packages/client/src/runtime/managed-skills.ts:320`; `packages/client/src/handlers/codex/sdk.ts:1590`]

✅ The Server correctly withholds bundle-backed Skills when the attachment is not ready or belongs to another organization, and the Client does not reconstruct inline fallback content when a descriptor is present.

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The shared archive-entry and portability contract closes the first blocker, but an unverified drifted Skill can still remain executable when bundle resolution fails.

**Risk level**: A

**PR summary**
- Problem: Team admins need complete Skill bundles—including scripts, references, and binary assets—to reach runtime agents through provider-native discovery instead of being flattened to an inline-only `SKILL.md`.
- Approach: Carry an authoritative attachment descriptor alongside compatibility fields, validate one shared portable ZIP/name contract at Server admission and Client projection, then atomically reconcile complete directories under the existing version fence and recovery model.
- Impacted modules: Shared Skill/runtime schemas, Server bundle validation and effective-resource projection, Client managed-Skill reconciliation and runtime handlers, deterministic tests, and the formal runtime QA case.

**Review findings**
❌ 1. The digest-drift failure path remains unchanged. If an owned target no longer matches `installedDigest`, reconciliation attempts to restage it; when the attachment resolver then fails, the catch records the error but leaves the modified directory in the provider-native discovery root. Every handler still discards `result.ok` and starts the provider, so an altered bundled script remains discoverable/executable even though the Skill is omitted from the briefing. The new tests cover drift repair and download failure separately, but not this combined path. Please restore a locally verified copy, quarantine the unverified target, or stop provider preflight when repair cannot complete, and add the combined regression. [R4 / `packages/client/src/runtime/managed-skills.ts:325`; `packages/client/src/handlers/codex/sdk.ts:1590`]

✅ The previous directory-flood blocker is resolved: Server admission and Client extraction now share and enforce a total-entry bound, including explicit directory entries.

✅ The new shared portability contract aligns exact manifest naming, normalized collision checks, wrapper validation, reserved paths/names, and provider target limits before runtime projection.

**Action taken**

baixiaohang (changes requested):
- Rationale: The previous drift/download blocker is closed on the normal path, but Server admission and Client final-tree verification still enforce different size contracts, and the new quarantine has no retention bound.

**Risk level**: A

**PR summary**
- Problem: Team admins need complete Skill bundles—including scripts, references, and binary assets—to reach runtime agents through provider-native discovery instead of being flattened to an inline-only `SKILL.md`.
- Approach: Carry an authoritative attachment descriptor alongside compatibility fields, validate portable ZIP/name rules at Server admission and Client projection, and atomically reconcile complete directories under the existing version fence and recovery model.
- Impacted modules: Shared Skill/runtime schemas, Server bundle validation and effective-resource projection, Client managed-Skill reconciliation and runtime handlers, deterministic tests, and the formal runtime QA case.

**Review findings**
❌ 1. The advertised shared size contract is not the contract used by final-tree verification. Server admission accepts up to `TEAM_SKILL_BUNDLE_LIMITS.maxUncompressedBytes` (25 MiB) and does not impose a per-file limit on ordinary assets; Client ZIP inspection accepts the same. After extraction, however, `digestDirectory()` calls `hashDirectoryRecursive()`, which still rejects any file over `MAX_SKILL_FILE_BYTES` (4 MiB) or tree over `MAX_SKILL_TOTAL_BYTES` (16 MiB). For example, an otherwise valid Skill containing one 5 MiB binary asset can be created and projected by the Server, then deterministically fail on every Client after download and extraction. Please make admission, extraction, and final digest use one authoritative limit set, and add accepted boundary cases—not only over-limit rejection cases. [R1/R5 / `packages/server/src/services/skill-bundle.ts:100`; `packages/client/src/runtime/managed-skills.ts:1353`; `packages/client/src/runtime/managed-skills.ts:1758`]

❌ 2. Every detected drift is moved to a fresh `.first-tree-workspace/.managed-skill-quarantine-*` directory, but no success, retry, startup, or bounded-retention path removes those directories. Repeated local edits—or a Skill that modifies its own installed tree—therefore retain another full bundle on each reconciliation and can grow workspace disk usage without bound despite the per-bundle limits. Please make quarantine lifecycle bounded (for example, retain at most one recoverable copy per managed key and clean it after a verified replacement, or remove unverified copies under an explicit bounded policy) and cover repeated drift/repair. [R3/R4 / `packages/client/src/runtime/managed-skills.ts:318`; `packages/client/src/runtime/managed-skills.ts:898`]

✅ The previous combined failure is resolved on the ordinary path: an installed target whose full-tree digest no longer matches is moved out of the Provider discovery root before bundle resolution, and the new regression covers drift plus resolver failure.

✅ Server and Client now share the strict manifest envelope and additional raw/normalized portability checks.

**Action taken**
- Submitted request changes on head `ca684c001cded4b2b0e6a3482553ee3b50ee9964`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The current head closes the previously reported admission/materialization mismatch and unsafe-discovery gaps without introducing a new blocking issue.

**Risk level**: A
- Semantic lift: none

**Review findings**
- ✅ Server admission, Client ZIP inspection, and final installed-tree digest now share compatible materialized file and size limits.
- ✅ Drift quarantine is bounded to a stable slot per managed target outside provider discovery.
- ✅ Provider start/resume/active-turn paths fail closed when a ledger-owned target cannot be verified, quarantined, or authoritatively removed.
- ✅ The shared portability and strict manifest contract rejects bundles at admission that the Client cannot safely materialize.

**Action taken**
- Approved exact head `59d91976176f4eb809aff1d1fa1539660295be96`.

baixiaohang (changes requested):
- Rationale: The new unsafe-mode check makes normal managed Skill trees unverifiable on Windows.

**Risk level**: A
- Semantic lift: none

**Review findings**

❌ 1. Make the group/other-write drift check platform-aware. [`packages/client/src/runtime/managed-skills.ts:1827`]

`assertManagedTreeModeSafe()` interprets `Stats.mode & 0o022` as POSIX group/other write authority on every platform. Node explicitly documents that Windows can only manipulate writability and does not distinguish owner, group, and others: https://nodejs.org/api/fs.html#fschmodpath-mode-callback. Consequently, normal writable Windows files/directories expose mode bits that this predicate cannot interpret as POSIX group/other permissions. Since `digestDirectory()` now calls this predicate on the staging root immediately after `mkdir`/`chmod`, a supported Windows Client can reject its own freshly staged Skill before installation (and subsequently classify normal installed targets as drift). Gate this POSIX invariant to platforms where those bits have that meaning, or implement a Windows-specific ACL/writability policy, and exercise the Windows path.

**Action taken**
- Requested changes on exact head `edcdd72b01f1a2f0bfb54f63d24e29afec546aa6`.

baixiaohang (changes requested):
- Rationale: A lock timeout still lets a Handler start its Provider without reaching the new universal publication verification.

**Risk level**: A
- Semantic lift: none

**Review findings**

❌ 1. Treat workspace-lock timeout as unsafe whenever the active Provider discovery root may contain managed content. [`packages/client/src/runtime/managed-skills.ts:440`]

The new catch intentionally exempts `WorkspaceFileLockTimeoutError` from conversion to `ManagedSkillsUnsafeDiscoveryError`, so after ten seconds reconciliation returns `{ ok: false }`; start/config paths consume the result without checking `ok`, and active-turn paths only rethrow the specialized unsafe error. A concrete unsafe interleaving remains: writer A holds the lock while an early Skill bundle download stalls, a later ledger-owned target is still drifted and has not yet been verified/quarantined, and Handler B times out, receives the non-throwing result, then starts the Provider with that drifted target still discoverable. The fact that another process owns the writer lock does not prove that every target has reached the final publication gate. Let this timeout follow the existing `providerDiscoveryMayContainManagedContent()` fail-closed branch (or otherwise make every Handler wait/retry/abort) instead of exempting it.

**Action taken**
- Requested changes on exact head `f042e1b32d263d50f2bc08a09b47757e915a5c44`.

yuezengwu:
**Exact-head verification**: NOT QA READY

**Rationale**: The implementation closes the previously identified authority and fail-closed gaps, but its one intentional provider-start exception is not protected by a committed deterministic regression.

**Risk level**: A

- Path baseline: includes `packages/client/**` → A
- Semantic lift: none

**Required follow-up**

❌ The new typed workspace-lock timeout contract is untested. `WorkspaceFileLockTimeoutError` is now the sole ordinary reconciliation error exempted from unsafe-discovery escalation when the provider root is non-empty. That exception is intentional—another writer holds the lock, so a caller may preserve availability against the previously verified target—but it is also the exact branch that lets provider startup continue without completing this reconciliation.

`workspace-file-lock.test.ts` remains unchanged and never reaches timeout, while `managed-skills.test.ts` has no real contention case using `lockTimeoutMs`. Please add:

1. a real held-lock/short-timeout assertion that the lock layer throws `WorkspaceFileLockTimeoutError`; and
2. a reconciler-level contention regression with an existing healthy managed target proving timeout returns a workspace availability failure (not `ManagedSkillsUnsafeDiscoveryError`) and leaves the target unchanged.

Also exercise late Core drift plus `quarantine_rename` failure directly, so the new universal Core+Team publication gate proves the provider-blocking branch for Core as well as the successful Core quarantine path.

I ran a repository-external exact-head probe for the two lock-timeout assertions; it passed on `f042e1b32d263d50f2bc08a09b47757e915a5c44`. The implementation therefore behaves correctly today, but this deterministic safety/availability boundary still needs to live in the product test suite as required by the repository QA contract.

**Verified on exact head**

- Shared focused: 14/14
- Server bundle focused: 41/41
- Client Managed Skills focused: 97/97
- Existing cross-process workspace-lock suite: 3/3
- Repository-external real lock-timeout probe: 1/1
- GitHub CI and CodeQL: all green

The universal publication gate, one-way attachment authority, platform-aware mode handling, strict YAML admission, conservative portable folding, bounded quarantine lifecycle, and ordinary FS/recovery ambiguity handling otherwise match the intended contract.

yuezengwu:
@baixiaohang 请基于 exact head `053c9218ffd2403ea917f7eca9a9be9ffee905ec` 复审上一轮 workspace-lock timeout finding。

该 head 保留 raw lock API 的 typed `WorkspaceFileLockTimeoutError`，但 reconcile 不再豁免该异常：只有 active provider discovery root 可证明为空时才返回普通 availability failure；非空或无法安全检查时均升级为 `ManagedSkillsUnsafeDiscoveryError`，阻止 Provider preflight。

我已在 detached exact-head worktree 独立复核并验证：

- Client focused：2 files / 105 tests passed（含真实跨进程 lock contention 的 raw typed timeout、empty-root availability、nonempty-root provider-blocking）
- Client package typecheck passed
- `git diff --check` passed，worktree clean
- 当前 GitHub lint/typecheck、Client/Web、Server、CLI、portable smoke、CodeQL 均通过

PR 保持开放且不会在正式 provider QA 与 exact-head review 完成前合并。

yuezengwu:
@baixiaohang 请复审 exact head `053c9218ffd2403ea917f7eca9a9be9ffee905ec`。

该 head 针对 review `4810353309` 移除了 reconcile 对 `WorkspaceFileLockTimeoutError` 的无条件豁免：active provider discovery root 非空或无法安全检查时，lock timeout 现在走 `ManagedSkillsUnsafeDiscoveryError` 并阻断 Provider；只有 root 可证明为空时才保留普通 availability failure。新增真实跨进程 contention 回归覆盖 raw typed timeout、empty-root ordinary failure、nonempty-root unsafe block，并补 late Core quarantine failure 与 stale Team final-publication 路径。

本地最终门禁：Client focused 105/105，`pnpm check`、11/11 typecheck、12/12 tests 全通过；GitHub CI/CodeQL 对该 exact head 也已全部通过。PR 保持开放且尚未交 formal QA，等待你的 exact-head 裁决。

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact-head fix closes the workspace-lock timeout window by blocking provider preflight whenever the active discovery root is nonempty or cannot be safely inspected.

**Risk level**: A
- Path baseline: includes packages/client/** -> A
- Semantic lift: none

**PR summary**
- Problem: Team Skills need complete attachment bundles to reach provider-native discovery roots without allowing stale, drifted, or partially verified content to execute.
- Approach: Treat the bundle as authoritative, materialize it atomically under shared portability and manifest rules, preserve only verified last-known-good state, and fail closed when managed discovery safety cannot be proven.
- Impacted modules: Client runtime and handlers, Server resource projection, Shared schemas, and Managed Skill QA coverage.

**Review findings**
- ✅ The prior lock-timeout blocker is resolved on this commit: empty discovery retains an availability failure, while nonempty or ambiguous discovery raises the unsafe-discovery signal.
- ✅ Real cross-process contention regressions cover the raw typed timeout and both reconciliation branches.
- ✅ GitHub CI and CodeQL are green on the exact reviewed head.

**Action taken**
- Approved exact head 053c9218ffd2403ea917f7eca9a9be9ffee905ec after maintainer authorization.

PR #2079, opened 2026-07-29
feat(resources): project complete Team Skill bundles at runtime

Gandy2025:
## Summary

- Project complete, admin-uploaded Team Skill ZIP bundles through a versioned authoritative runtime snapshot instead of reducing them to an inline `SKILL.md`.
- Resolve bundle-backed resources to authenticated attachment descriptors on the Server, and preserve per-Skill last-known-good state when a selected bundle is temporarily unavailable.
- Add an `attachment-bundle` Client source that downloads only when an install, update, or drift repair is required, then feeds the extracted directory through the existing provider-native managed-Skill transaction.
- Keep null-bundle legacy Team Skills compatible through the existing inline projection.
- Advance agent config versions when legacy bundle backfill changes the authoritative resource snapshot so a Client cannot reconcile mixed revisions.

This closes the final runtime gap after complete Team Skill bundle upload/storage landed: supporting scripts, references, and agents now reach the active provider's native Skill root.

## Runtime and failure semantics

- A recognized authoritative snapshot wins and is never merged with `resourceSkills`.
- A missing snapshot falls back to legacy `resourceSkills`, supporting a new Client with an old Server.
- A malformed or future snapshot preserves the whole Team Skill last-known-good projection.
- A per-resource unavailable sentinel keeps that Skill in `desiredKeys`, while other Skills can continue to reconcile.
- Omission from a valid authoritative snapshot is a revoke.
- Bundle-backed rows never dual-send a derived inline body, avoiding a false-success installation without supporting files.
- Runtime config resolution uses a version fence and bounded retry so resource selection and attachment descriptors come from one coherent snapshot.

## Security

- Uses authenticated `fetchAttachment`; attachment capability paths are redacted from logs and missing capabilities return a generic response.
- Streams downloads and ZIP entries with independent Client-side limits for compressed bytes, files, total bytes, per-file bytes, path depth, and portable path length.
- Rejects traversal, absolute/drive/backslash/control paths, empty/dot segments, Windows device names, trailing dot/space, duplicate or case-colliding paths, encryption, symlinks/special files, size mismatches, extra wrapper-root content, and archive-owned management metadata.
- Accepts exactly one exact-case `SKILL.md`, with either root files or one optional wrapper directory.
- Creates contained files with exclusive creation, preserves only executable bits, and completes all remote I/O before opening a managed-Skill journal.
- Reuses the existing lock, ownership proof, manifest/digest validation, journaled swap, crash recovery, monotonic fence, conflict handling, and provider-switch semantics.

## Rollout

Client-first rollout is a hard prerequisite:

1. Deploy snapshot-aware Clients first. They continue to consume an old Server's legacy `resourceSkills`.
2. Deploy the new Server only after those Clients are available.
3. Old Clients receiving a new Server configuration can use only null-bundle legacy inline Skills; bundle-backed Skills intentionally do not receive an incomplete inline fallback.

## Validation

- [x] `pnpm check` — passed; 16 existing warnings and 1 informational diagnostic, no errors.
- [x] `pnpm typecheck` — 11/11 tasks passed.
- [ ] `pnpm test` — executed twice. The controlled monorepo run exercised the full suite but did not finish green because of unrelated long scheduler stalls and two `packages/web` tests outside this diff. The timed-out CLI and Server files passed in isolation; the Web SSR timeout passed in isolation; one unchanged Web focus-retention test still reproduces in isolation. Package and focused evidence below is green.

Focused and package evidence:

- `pnpm --filter @first-tree/shared exec vitest run src/__tests__/team-skill-bundle.test.ts src/__tests__/agent-runtime-config.test.ts src/observability/__tests__/redact-url.test.ts` — 76/76 passed.
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/skill-bundles.test.ts src/__tests__/resources-phase1.test.ts src/__tests__/attachments-route.test.ts src/__tests__/runtime-config-snapshot.test.ts src/__tests__/service-branch-defaults.test.ts` — 109/109 passed.
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/managed-skills.test.ts src/__tests__/sdk-comprehensive.test.ts` — 81/81 passed.
- `pnpm --filter @first-tree/server test` — full Server package suite passed.
- The controlled monorepo run completed the full Client suite with 1,734 passed / 3 skipped and the full Shared suite with 729 passed.
- Isolated environmental failures: `client-service.test.ts` 15/15, `message-dispatcher.test.ts` 11/11, `context-policy-runtime-layout.test.ts` 5/5, and `page-ssr-smoke.test.tsx` 9/9 passed.
- The concrete `context-tree-value-audit-0.2.5.zip` acceptance bundle was accepted and extracted with its supporting files under the runtime limits.

Four independent Codex review passes covered correctness and security. Findings fixed before this commit included exact-limit metadata overhead, wrapped-root edge cases, config snapshot consistency, attachment capability redaction, target ownership races, unsafe staging cleanup, legacy backfill size bounds, portable path limits, and explicit empty-directory extraction.

## Change Surface

- [x] shipped or planned skill topology

## Notes

- Package or install behavior changes: complete Team Skill bundles are installed into only the active provider-native Skill root.
- Docs or tests updated to match: `packages/qa/cases/runtime/managed-skill-projection.md` now requires a real provider turn to execute a bundled supporting script and consume a bundled reference.
- Follow-up work: formal provider-backed QA with `context-tree-value-audit-0.2.5.zip` remains warranted before production rollout.

## Context Tree

Implemented against `agent-team-foundation/first-tree-context@5099df1d035d48dd6062eb2e5a9735ba70e29c7d`:

- `system/cloud/agents/resources.md` — Runtime Projection Boundary
- `system/architecture.md` — Skill Payload Shape
- `system/context-management/workspace-layout.md` — File Layout

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -41,6 +41,7 @@
     "smol-toml": "^1.7.0",
     "ws": "^8.20.0",
     "yaml": "^2.8.3",
+    "yauzl": "^3.4.0",
     "zod": "^4.0.0"
   },
   "devDependencies": {
@@ -49,7 +50,9 @@
     "@types/node": "^22.16.0",
     "@types/semver": "^7.5.8",
     "@types/ws": "^8.18.1",
+    "@types/yauzl": "^3.4.0",
     "@vitest/coverage-v8": "3.2.4",
+    "fflate": "^0.8.2",
     "tsdown": "^0.21.4",
     "vitest": "^3.2.0"
   }
--- a/packages/client/src/__tests__/managed-skills.test.ts
+++ b/packages/client/src/__tests__/managed-skills.test.ts
@@ -12,14 +12,24 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import type { RuntimeProvider, RuntimeResourceSkill } from "@first-tree/shared";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  type AgentRuntimeConfig,
+  type RuntimeProvider,
+  type RuntimeResourceSkill,
+  type RuntimeTeamSkillAttachmentEntry,
+  type RuntimeTeamSkillEntry,
+  TEAM_SKILL_BUNDLE_LIMITS,
+} from "@first-tree/shared";
+import { strToU8, zipSync } from "fflate";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CORE_SKILL_NAMES } from "../runtime/first-tree-skills/installer.js";
 import {
   authoritativeTeamSkillSnapshot,
+  type ManagedSkillAttachmentFetcher,
   type ManagedSkillsCheckpoint,
   providerSkillRoot,
   reconcileManagedSkills,
+  teamSkillSnapshotFromConfig,
 } from "../runtime/managed-skills.js";
 import {
   MANAGED_SKILLS_JOURNAL_REL,
@@ -28,6 +38,7 @@ import {
   readManagedState,
   readManagedStateResult,
 } from "../runtime/managed-state.js";
+import { extractTeamSkillBundle } from "../runtime/team-skill-bundle.js";
 
 const PROVIDERS: readonly RuntimeProvider[] = ["claude-code", "claude-code-tui", "codex", "cursor", "kimi-code"];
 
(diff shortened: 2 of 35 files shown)


yuezengwu (changes requested):
## Changes requested

Reviewed head `728598e36a970513226702e309b480515b182f47`.

This PR’s goal is to carry complete Team Skill bundles from Cloud attachment storage through a versioned runtime snapshot into the active provider-native Skill root, while keeping legacy inline Skills and last-known-good failure semantics.

Two blockers:

1. `packages/server/src/services/skill-bundle.ts:212-216` validates the raw manifest name and rejects only exact current Core names, while `packages/client/src/runtime/managed-skills.ts:1704-1730` first normalizes `_` and other separators to `-` and then rejects Core/Windows-reserved slugs. As a result, bundles such as `name: first_tree_read` or `name: CON` are accepted and persisted by the Server but can never be installed by the new Client projection. This violates the accepted-bundle/runtime-admissibility contract and leaves an administratively valid Team Skill permanently degraded. Please centralize the runtime-slug/reserved-name validation in Shared, apply it during Server bundle acceptance, and cover the normalized-Core and Windows-device cases in the Server acceptance tests.

2. The exact Context Tree commit named in the PR (`agent-team-foundation/first-tree-context@5099df1d035d48dd6062eb2e5a9735ba70e29c7d`) says in both `system/cloud/agents/resources.md` and `system/architecture.md` that the Cloud attachment bundle is **not** a Client projection source and that Team Skills are materialized from the inline compatibility projection. This PR implements the opposite durable runtime boundary and marks shipped Skill topology as changed. Please land/cross-link a source-backed Context Tree update that establishes attachment-bundle projection, its LKG/revoke semantics, and the client-first rollout constraint; then update the PR’s Context Tree reference to that decision.

Human rollout review is also warranted because the change introduces a new authoritative wire snapshot, exposes attachment capability IDs to the runtime, backfills legacy bundles into attachment rows, and advances `agent_configs.version` for affected agents. There is no new schema migration in this diff, but it materially changes persistent data and reconciliation behavior.

I did not run tests, per the reviewer instructions. `git diff --check` is clean; the visible GitHub checks are green, and the PR currently conflicts with `main`.

baixiaohang (changes requested):
- Rationale: The compatibility fallback can downgrade an already-installed complete bundle to the incomplete inline projection during a normal mixed-Server rollout.

**Risk level**: A
- Semantic lift: none (already at the highest path tier)

**PR summary**
- Problem: Admin-uploaded Team Skills currently reach managed agents only as a flattened `SKILL.md`, so provider turns cannot use bundled scripts, references, or other supporting files.
- Approach: Add a versioned Team Skill snapshot with attachment descriptors, download and validate complete ZIPs on the Client, and feed the extracted directory through the existing provider-native managed-Skill transaction with last-known-good handling.
- Impacted modules: Client managed-Skill reconciliation and every provider handler; Server resource/config resolution, attachment access, and legacy backfill; Shared runtime schemas and ZIP policy; runtime QA.

**Review findings**

❌ 1. A missing snapshot must not supersede an already-observed snapshot at the same config version. [R2/R4/R5 / `packages/client/src/runtime/managed-skills.ts:245`]

During a rolling Server deployment, a new Client can first fetch config version N from a new Server and install an `attachment-bundle`. An old Server replica connected to the same database then returns that same version N without `teamSkillSnapshot` and, after the new Server's backfill, still derives the Resource payload into legacy `resourceSkills`. This function treats the missing field as a fully authoritative inline snapshot, while reconciliation accepts `resourceConfigVersion >= state.resourceConfigVersion`; the inline revision therefore replaces the complete bundle and removes its scripts/references. A later response from a new replica reverses it again. Reconnects and forced refreshes make this observable even though the in-memory cache skips ordinary same-version notifications.

Please make the projection protocol/source part of the persisted monotonic fence (so legacy absence cannot overwrite a recognized snapshot at the same or older version), or enforce an equivalent Server rollout/backfill gate that proves old replicas are drained before attachment snapshots become authoritative. Add a deterministic sequence covering new-Server snapshot -> same-version old-Server fallback -> new-Server snapshot and require the complete installed tree to remain unchanged.

The existing change request already records the independent Context Tree contract conflict and Server/Client runtime-name admissibility mismatch, so I have not duplicated those findings here.

**Action taken**
- Submitted request changes on head `728598e36a970513226702e309b480515b182f47`.
- Static review only; no checks or tests were run.

PR #2086, opened 2026-07-30
fix(web): render Skill upload pickers with app-owned English copy

yuezengwu:
## Problem

In Settings → Resources, the Skill editor's **Upload ZIP or SKILL.md** and **Choose Skill folder** modes rendered the raw `input[type=file]`. Browsers draw that control's button and empty-state label in the OS locale (e.g. "选择文件 / 未选择任何文件"), so Chinese copy leaked into the English product UI.

## Change

`packages/web/src/pages/settings/resource-editors.tsx` (SkillEditor only):

- Keep both native file inputs mounted but hidden (`className="hidden"`), preserving `accept` / `multiple` / `webkitdirectory` and the existing change handlers.
- Add app-owned triggers via the existing `Button` (`type="button"`, `variant="outline"`): **Choose file** / **Choose folder**.
- Render selection state explicitly in English instead of relying on the browser: file mode shows the chosen file name with a "No file selected" empty state; folder mode shows "N files selected" with a "No folder selected" empty state.

Scope is limited to these two pickers — paste mode, bundle preparation, and all other upload entries are untouched.

## Tests

`packages/web/src/pages/__tests__/resource-editors-dom.test.tsx` gains two regression tests covering both modes:

- the native input stays mounted but hidden (attributes preserved), and the app-owned English trigger/empty-state copy is what renders;
- the trigger button is `type="button"` and clicks through to the hidden input;
- after a change event, the chosen file name / file count is displayed.

## Verification

- `pnpm --filter @first-tree/web exec vitest run src/pages/__tests__/resource-editors-dom.test.tsx` — 17/17 pass
- Full web suite (`pnpm --filter @first-tree/web test`) — 217 files / 1868 tests pass
- `pnpm check` — pass (pre-existing warnings only)
- `pnpm typecheck` — pass

### DIFF
--- a/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
+++ b/packages/web/src/pages/__tests__/resource-editors-dom.test.tsx
@@ -592,6 +592,14 @@ describe("resource editors", () => {
     expect(document.body.textContent).toContain("No file selected");
     expect(document.body.textContent).not.toContain("未选择任何文件");
 
+    // The hidden input's accessible value is gone with it, so the trigger must
+    // be tied to the status text (aria-describedby) and that text must be a
+    // live status that announces the selection after the dialog closes.
+    const status = document.getElementById("skill-file-status");
+    expect(status?.getAttribute("role")).toBe("status");
+    expect(status?.textContent?.trim()).toBe("No file selected");
+    expect(choose?.getAttribute("aria-describedby")).toBe("skill-file-status");
+
     // The button opens the hidden input's picker.
     const clickSpy = vi.spyOn(HTMLInputElement.prototype, "click").mockImplementation(() => {});
     await click(choose);
@@ -607,6 +615,8 @@ describe("resource editors", () => {
     await flush();
     expect(document.body.textContent).toContain("SKILL.md");
     expect(document.body.textContent).not.toContain("No file selected");
+    // The live status itself carries the selection (not just visible text).
+    expect(document.getElementById("skill-file-status")?.textContent?.trim()).toBe("SKILL.md");
     await pressEscape();
     await act(async () => root.unmount());
   });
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
- Rationale: The localized-copy fix preserves the upload behavior, but hiding the native inputs removes their accessible value without giving the replacement controls an equivalent accessible selection state.

**Risk level**: C
- Semantic lift: none; the change does not touch subscriptions, WebSocket parsing, or the data layer

**PR summary**
- Problem: Admins using Settings → Resources currently see OS-localized browser copy inside the English Skill upload flow.
- Approach: Keep the file and folder inputs as the upload mechanism, hide their browser-rendered UI, and expose app-owned English buttons and selection text instead.
- Impacted modules: Settings Skill resource editor and its DOM regression tests

**Review findings**
❌ 1. Preserve an accessible selection state when replacing the native file controls. `className="hidden"` removes each input—and therefore its native filename/value—from the accessibility tree, but the new button is not associated with the adjacent status text and that text is not a live status. After the file dialog closes, a screen-reader user has no control state that exposes or announces whether a file/folder was selected. Please give the replacement control an accessible relationship to the current status and announce changes (and cover that contract in the regression tests), rather than testing only visible `textContent`. [R5 / packages/web/src/pages/settings/resource-editors.tsx:742]

✅ 2. The change retains `accept`, `multiple`, `webkitdirectory`, and the existing change handlers, so the file/folder bundle paths continue to converge on the existing preparation and server-validation flow.

**Action taken**
- Submitted request changes

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up commit restores an accessible selection status for both custom upload triggers, resolving the prior blocker without changing the underlying Skill bundle behavior.

**Risk level**: C
- Semantic lift: none; the change does not touch subscriptions, WebSocket parsing, or the data layer

**PR summary**
- Problem: Admins using Settings → Resources currently see OS-localized browser copy inside the English Skill upload flow.
- Approach: Keep the file and folder inputs as the upload mechanism, hide their browser-rendered UI, expose app-owned English buttons and selection text, and associate each trigger with a live accessible status.
- Impacted modules: Settings Skill resource editor and its DOM regression tests

**Review findings**
✅ 1. The previous accessibility blocker is resolved: each trigger now references its current selection through `aria-describedby`, and the associated `role="status"` live region announces changes for both file and folder modes.
✅ 2. The regression tests cover the accessible relationship, live status contents before and after selection, and click-through for both hidden inputs.
✅ 3. `accept`, `multiple`, `webkitdirectory`, and the existing change handlers remain intact, so all Skill inputs continue to converge on the existing bundle preparation and server-validation path.

**Action taken**
- Approved

PR #2091, opened 2026-07-30
Refactor the team switcher menu

Gandy2025:
## Summary

- reorganize the menu into Current team, Switch team, and Add or join sections
- move Invite teammates beside Leave team, use semantic invite/link icons, and neutralize role badges
- move rename into an accessible dialog with roving menu focus and trigger focus restoration
- keep the existing switch, leave, create, join, and Ask-agent navigation behavior unchanged
- expand the preview and interaction coverage for multi-team, single-team, member, error, and keyboard states

## Testing

- `pnpm --filter @first-tree/web test` — 218 files, 1872 tests passed
- design-token guard and Biome checks on all changed web files
- Chromium QA at `/preview/team-switcher`: multi-team layout, rename dialog, Arrow/Home/End/Escape, Tab focus exit, and focus restoration
- two independent reviews of the final merge-base-scoped diff; no remaining blockers

### DIFF
--- a/packages/web/src/components/__tests__/layout-ask-agent-lock-dom.test.tsx
+++ b/packages/web/src/components/__tests__/layout-ask-agent-lock-dom.test.tsx
@@ -398,31 +398,6 @@ describe("Layout Ask agent navigation lock", () => {
     await act(async () => root.unmount());
   });
 
-  it("dismisses TeamSwitcher navigation dialogs that were already open when the lock engages", async () => {
-    const { container, root } = await renderLayout("/?review=need-you");
-    const trigger = () => container.querySelector<HTMLButtonElement>('button[aria-label^="Switch team"]');
-
-    await click(trigger());
-    await click(menuItemByText(container, "Leave team"));
-    await waitForText("Leave Team One?");
-    expect(document.body.querySelector('[role="dialog"]')).not.toBeNull();
-
-    await engageLock();
-    expect(document.body.textContent).not.toContain("Leave Team One?");
-
-    await releaseLock();
-    await click(trigger());
-    await click(menuItemByText(container, "Create team"));
-    await waitForText("Create a new team");
-    expect(document.body.querySelector('[role="dialog"]')).not.toBeNull();
-
-    await engageLock();
-    expect(document.body.textContent).not.toContain("Create a new team");
-    expect(locationText(container)).toBe("/?review=need-you");
-
-    await act(async () => root.unmount());
-  });
-
   it("keeps the real UserMenu inert while locked: no account settings, no sign out", async () => {
     const { container, root } = await renderLayout("/?review=need-you");
     const trigger = () => container.querySelector<HTMLButtonElement>('button[aria-label^="User menu"]');
--- a/packages/web/src/components/__tests__/layout-ask-agent-lock-dom.test.tsx
+++ b/packages/web/src/components/__tests__/layout-ask-agent-lock-dom.test.tsx
@@ -360,10 +360,10 @@ describe("Layout Ask agent navigation lock", () => {
     // the other-team switch row, Leave, Create, Join, Invite.
     const teamTwoRow = menuItemByText(container, "Team Two");
     expect(teamTwoRow?.disabled).toBe(true);
-    const createRow = menuItemByText(container, "Create new team");
+    const createRow = menuItemByText(container, "Create team");
     expect(createRow?.disabled).toBe(true);
     expect(menuItemByText(container, "Join with invite link")?.disabled).toBe(true);
-    expect(menuItemByText(container, "Leave this team")?.disabled).toBe(true);
+    expect(menuItemByText(container, "Leave team")?.disabled).toBe(true);
     expect(menuItemByText(container, "Invite teammates")?.disabled).toBe(true);
 
     // The switch action boundary re-checks the lock: no cache-clearing
@@ -391,9 +391,34 @@ describe("Layout Ask agent navigation lock", () => {
 
     // …and Create reaches the setup modal.
     await click(trigger());
-    await waitForText("Create new team");
-    await click(menuItemByText(container, "Create new team"));
+    await waitForText("Create team");
+    await click(menuItemByText(container, "Create team"));
     await waitForText("Create a new team");
+
+    await act(async () => root.unmount());
+  });
+
+  it("dismisses TeamSwitcher navigation dialogs that were already open when the lock engages", async () => {
+    const { container, root } = await renderLayout("/?review=need-you");
+    const trigger = () => container.querySelector<HTMLButtonElement>('button[aria-label^="Switch team"]');
+
+    await click(trigger());
+    await click(menuItemByText(container, "Leave team"));
+    await waitForText("Leave Team One?");
+    expect(document.body.querySelector('[role="dialog"]')).not.toBeNull();
+
+    await engageLock();
+    expect(document.body.textContent).not.toContain("Leave Team One?");
+
+    await releaseLock();
+    await click(trigger());
+    await click(menuItemByText(container, "Create team"));
+    await waitForText("Create a new team");
+    expect(document.body.querySelector('[role="dialog"]')).not.toBeNull();
+
+    await engageLock();
+    expect(document.body.textContent).not.toContain("Create a new team");
+    expect(locationText(container)).toBe("/?review=need-you");
 
     await act(async () => root.unmount());
   });
--- a/packages/web/src/components/__tests__/team-setup-modal-dom.test.tsx
+++ b/packages/web/src/components/__tests__/team-setup-modal-dom.test.tsx
@@ -4,7 +4,6 @@ import { act, type ReactElement } from "react";
 import { createRoot, type Root } from "react-dom/client";
 import { MemoryRouter } from "react-router";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { addAskAgentNavLock, clearAskAgentNavLocks, removeAskAgentNavLock } from "../chat/ask-agent-nav-lock.js";
 
 globalThis.IS_REACT_ACT_ENVIRONMENT = true;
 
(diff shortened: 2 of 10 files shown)


baixiaohang (changes requested):
- Rationale: A late Ask-agent navigation lock can leave successful team-membership mutations only half-applied in the client.

**Risk level**: B-low
- Path baseline: all changes are under `packages/web/**` -> C
- Semantic lift: changes the selected-team data handoff, org-scoped cache handling, and WebSocket reconnection path -> B-low

**PR summary**
- Problem: Team members need a clearer way to understand the current team, switch teams, and find rename, invite, leave, create, and join actions without losing an in-progress Ask-agent interaction.
- Approach: Reorganizes the menu into current-team, switch-team, and add-or-join sections; moves rename into a keyboard-accessible dialog; and adds late lock checks around navigation-capable team actions.
- Impacted modules: `packages/web/src/components/team-switcher.tsx`, `team-setup-modal.tsx`, their DOM interaction tests, and the TeamSwitcher preview.

**Review findings**
❌ 1. Preserve a completed create/join result instead of dropping its continuation when the lock engages. The new checks run only after the POST has already created the organization or membership. At the same time, the parent dismisses the setup dialog when the lock appears. Returning here therefore leaves the server mutation committed but never selects the new team, refreshes the organization list, navigates to onboarding, or resumes after unlock. A retry can create another team or hit an already-joined state, while the successful team may remain invisible until a later refresh. Queue the successful continuation until unlock, or explicitly reconcile and surface the completed result. [`packages/web/src/components/team-setup-modal.tsx:85`, `packages/web/src/components/team-setup-modal.tsx:150`]

❌ 2. Reconcile a successful leave atomically with the selected-team transition. If the lock starts after `leaveMembership` was sent, the next-team branch calls `switchAfterLeave`, which now returns immediately while locked; that skips `selectOrganization`, so its org-scoped query-cache clear and admin-WebSocket reconnect never happen, and there is no unlock continuation. The last-team branch still clears the cache and navigates to onboarding while locked, so it does not preserve the Ask surface at all. Once the membership deletion succeeds, both branches need one durable/deferred completion path rather than conditionally abandoning or bypassing the transition. [`packages/web/src/components/team-switcher.tsx:163`, `packages/web/src/components/team-switcher.tsx:189`]

✅ 3. The menu hierarchy, roving keyboard focus, rename dialog focus restoration, and role-neutral badge treatment are coherent and covered at the DOM interaction layer.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
I independently confirmed the two blocking late-lock cases on head 0813d6b.

- A create or join POST can succeed after the setup dialog has been dismissed, but the new lock check drops the successful continuation. The client neither selects nor refreshes the new membership, and there is no resume-on-unlock path.
- A leave request can likewise commit before the lock engages. The next-team path then skips the cache-clearing and WebSocket-reconnecting selectOrganization call, while the last-team path still navigates away under the lock.

Please make each successful membership mutation complete through one durable deferred/reconciliation path after the Ask-agent lock releases. The menu hierarchy, rename-dialog accessibility, focus behavior, and neutral role styling otherwise look sound.

Gandy2025:
Addressed both requested changes in `5c6ba6636`.

1. Create / join no longer drop a committed server result. Their successful continuation is registered with `runAfterAskAgentNavUnlock`, survives the setup dialog unmount, and starts `selectOrganization` synchronously in the final unlock notification before closing and navigating. The gate also keeps waiting if an earlier unlock subscriber installs a newer lock.

2. Leave now uses the same deferred completion path after the membership deletion succeeds. The next-team branch starts `selectOrganization` only through the atomic unlock gate, preserving its query-cache clear, WebSocket reconnect, and `/me` refresh. The last-team branch waits for unlock, refreshes membership state, then rechecks the gate before clearing caches and navigating to onboarding.

Regression coverage now includes:

- create and join requests that resolve after the dialog has unmounted
- next-team and last-team leave requests that resolve under a late lock
- multiple simultaneous locks
- a new lock installed synchronously during the unlock handoff

Validation: `pnpm --filter @first-tree/web test` (218 files / 1880 tests), Web typecheck + design-token guard, Biome, and an independent final race review all pass. Please re-review the updated head.

yuezengwu (approved):
Reviewed the complete updated head 5c6ba66. The two late-lock blockers are resolved:

- successful create and join mutations now retain one continuation until the final Ask-agent navigation lock releases, including after the setup form unmounts;
- successful leave mutations now defer the whole selected-team or last-team recovery path, preserving the Ask surface while still completing cache, WebSocket, authentication refresh, and navigation handoff afterward.

The unlock helper handles multiple locks and starts each continuation once. The added interaction tests cover the deferred create, join, next-team leave, and last-team leave paths. No remaining blockers found.

baixiaohang (changes requested):
- Rationale: The original successful-mutation continuations are now retained, but the last-team leave path still has an async window that can unmount a newly locked Ask surface.

**Risk level**: B-low
- Path baseline: all changes are under `packages/web/**` -> C
- Semantic lift: the fix changes selected-team state, org-scoped cache handling, and the navigation-lock handoff -> B-low

**PR summary**
- Problem: Team members need a clearer TeamSwitcher while an in-progress Ask-agent interaction remains mounted across team-management races.
- Approach: Keeps the reorganized menu and accessible rename dialog, and adds a shared unlock continuation so already-committed create, join, and leave mutations resume their client transition instead of being dropped.
- Impacted modules: TeamSwitcher, TeamSetupModal, the Ask-agent navigation-lock store, and their DOM/store tests.

**Review findings**
❌ 1. Keep the last-team `/me` refresh inside an owned transition, not before the second lock check. After the original lock releases, `refreshMe()` waits on the network while the old Workspace/Ask surface remains interactive. A user can start a new Ask attempt in that window. When `/me` returns with no active membership, `AuthProvider.fetchMe` immediately writes the empty membership list and onboarding facts; on `/`, `WorkspacePage` can then redirect to `/onboarding` and unmount that new Ask owner before the nested `runAfterAskAgentNavUnlock` is reached. The nested guard protects only the explicit cache clear/navigation, not the stateful refresh that can already trigger the exit. Start a synchronous visible transition before yielding, or otherwise prevent/re-check the refresh's state commit so a newer lock cannot be bypassed. The new last-team test uses an immediately resolved `refreshMe`, so it does not cover this re-lock-during-refresh case. [`packages/web/src/components/team-switcher.tsx:195`]

✅ 2. The previous create/join finding is resolved: the committed result survives form unmount and resumes through `selectOrganization` after unlock.

✅ 3. The previous next-team leave finding is resolved: after unlock, the continuation enters `selectOrganization` synchronously, preserving its cache-clear and WebSocket-reconnect boundary.

**Action taken**

yuezengwu (changes requested):
I am replacing my approval after confirming the remaining last-team race on head 5c6ba66.

The outer unlock gate starts refreshMe, but that refresh yields to the network while the Workspace Ask surface is interactive again. If a new Ask lock starts before the response returns, AuthProvider still commits the empty membership list and onboarding facts. WorkspacePage can then render its onboarding redirect and unmount the new Ask owner before execution reaches the nested unlock gate.

The nested gate therefore protects only the explicit cache clear and navigate call, not the stateful refresh that can already cause the same exit. Please keep the refresh inside a visibly owned transition or otherwise prevent its state commit from bypassing a newer lock, and cover the re-lock-during-refresh timing in the last-team test. The create/join and next-team leave fixes remain sound.

Gandy2025:
The PR has been corrected to the original presentation-only scope in `63726c560`.

@baixiaohang — both successful-mutation findings are removed at their source: the added create/join continuation handling, leave transition changes, dialog-on-lock dismissal, and shared unlock helper have all been reverted. The existing create, join, leave, cache, WebSocket, and Ask-lock behavior now matches the merge base.

@yuezengwu — the last-team refresh/re-lock path is likewise no longer part of this PR. There is no new recovery route, deferred refresh, or navigation handoff to review.

The final merge-base diff is limited to the TeamSwitcher information hierarchy, layout, icon/copy/role presentation, rename dialog, menu keyboard/focus presentation, preview fixtures, and matching UI tests.

Validation on the scoped head:

- Web: 218 files / 1872 tests passed
- Targeted TeamSwitcher and existing Ask-lock coverage: 78 tests passed
- TypeScript, design-token guard, Biome, and `git diff --check` passed
- Two independent final-diff reviews found no remaining blockers

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The final merge-base-scoped diff is presentation-only, preserves the existing membership/navigation behavior, and has no remaining blocking issue.

**Risk level**: C
- Path baseline: all changes are under `packages/web/**` -> C
- Semantic lift: none; the intermediate selected-team, navigation-lock, and deferred-mutation changes were removed from the final diff

**PR summary**
- Problem: Team members need a clearer, more accessible TeamSwitcher for understanding the current team and finding switch, rename, invite, leave, create, and join actions.
- Approach: Groups the menu into Current team, Switch team, and Add or join sections; moves rename into a focused dialog; adds roving menu keyboard navigation and focus restoration; and neutralizes role-badge styling.
- Impacted modules: TeamSwitcher, its DOM/lock presentation assertions, and the TeamSwitcher preview.

**Review findings**
✅ 1. The final diff leaves TeamSetupModal, the Ask-agent lock store, and the existing leave/switch data handoff unchanged, so the races found in the intermediate heads are no longer introduced by this PR.

✅ 2. Team rename remains colocated with the active-team control while the dialog provides labeled input, error announcement, Escape/outside protection during save, and trigger focus restoration.

✅ 3. The menu hierarchy and roving focus cover Arrow Up/Down, Home/End, Escape, and Tab exit without adding a parallel team-management surface.

**Action taken**
- Approved the updated head.

yuezengwu (approved):
Reviewed the complete merge-base diff on head 63726c5. The previously blocking create, join, leave, refresh, cache, WebSocket, and Ask-lock continuation changes have been removed; those paths now match the base branch.

The remaining diff is scoped to TeamSwitcher information hierarchy and presentation, invite placement and icons, role styling, the rename dialog, menu keyboard and focus behavior, preview fixtures, and matching DOM tests. The menu remains role-correct, the invite entry stays available to regular members, and the dialog focus restoration is coherent. No remaining blockers found.

PR #2092, opened 2026-07-30
test(web): make Ask Agent timeout fixture relative

Gandy2025:
## Summary

- replace the Ask Agent navigation-lock test's fixed July 28 timestamps with timestamps relative to the test start
- keep the production 45-second reply timeout unchanged

## Why

The test fixture became older than the production timeout window on July 30. The hook therefore failed the pending attempt immediately, making the test deterministically fail both locally and in CI even though the navigation-lock behavior had not regressed.

## Validation

- `pnpm --filter @first-tree/web exec vitest run src/components/chat/__tests__/ask-agent-nav-lock.test.tsx` — 7/7 passed
- design-token guardrails — passed

This is test-only and does not change runtime behavior.

### DIFF
--- a/packages/web/src/components/chat/__tests__/ask-agent-nav-lock.test.tsx
+++ b/packages/web/src/components/chat/__tests__/ask-agent-nav-lock.test.tsx
@@ -395,19 +395,20 @@ function AskAgentProbe() {
 
 describe("useAskAgent navigation lock publishing", () => {
   it("locks while a clarification awaits a reply and unlocks when the reply lands", async () => {
+    const startedAt = Date.now();
     const clarification = message({
       id: "clarification-1",
       senderId: "human-1",
       inReplyTo: "req-1",
       metadata: { askAgent: { requestId: "req-1", agentId: "agent-1" } },
-      createdAt: "2026-07-28T10:01:00.000Z",
+      createdAt: new Date(startedAt).toISOString(),
     });
     const reply = message({
       id: "reply-1",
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed the test fixture against the useAskAgent timeout path. The relative timestamps keep the clarification inside the existing 45-second window, preserve reply ordering, and leave production behavior unchanged. No blocking findings.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change removes calendar-time coupling from the fixture while preserving the production timeout and the test's pending-reply semantics.

**Risk level**: C
- Path baseline: only `packages/web/**` test code is changed -> C
- Semantic lift: none; no runtime, subscription, WebSocket-frame, or data-layer behavior changes

**PR summary**
- Problem: The Ask Agent navigation-lock regression test began timing out immediately once its fixed clarification timestamp aged beyond the production 45-second reply window.
- Approach: Anchor the clarification timestamp to the test start and keep the synthetic reply one second later, so the fixture remains inside the live timeout window without changing production behavior.
- Impacted modules: Web Ask Agent navigation-lock test fixture

**Review findings**
✅ The relative timestamps preserve the clarification-before-reply ordering and continue to exercise the durable rule that navigation stays locked until a matching reply or timeout releases the attempt.
✅ The production `ASK_AGENT_REPLY_TIMEOUT_MS = 45_000` path is untouched.

**Action taken**
- Approved.

PR #2093, opened 2026-07-30
Keep chat composer stable during open-request checks

Gandy2025:
## Summary

- keep the ordinary chat composer mounted while the open-request snapshot is loading or unavailable
- keep both the disabled send control and the mutation-boundary guard fail-closed until the snapshot is confirmed
- show a neutral inline retry row on verification failure, preserving the textarea node and draft
- retain the existing open-request takeover once a pending question is confirmed

## Testing

- `pnpm --filter @first-tree/web exec vitest run src/pages/workspace/center/__tests__/chat-view-dom.test.tsx src/pages/__tests__/page-ssr-smoke.test.tsx` (73 tests)
- `pnpm check` (passes; 16 pre-existing warnings and 1 info)
- full Web suite: 218 files, 1,871 tests
- Playwright browser QA for pending, failed, and retry-success states

## Review

- independent logic review: approved after adding an Enter-key fail-closed regression
- independent UX/accessibility review: approved after aligning failure-state color and accessible naming with the design system

### DIFF
--- a/packages/web/src/lib/__tests__/use-chat-draft-text.test.tsx
+++ b/packages/web/src/lib/__tests__/use-chat-draft-text.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment happy-dom
 
-import { act } from "react";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { act, type Dispatch, type SetStateAction, useLayoutEffect } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createDomHarness, type DomHarness } from "../../test-utils/dom-harness.js";
 import { chatDraftScope, loadDraft, saveDraft } from "../draft-store.js";
 import { useChatDraftText } from "../use-chat-draft-text.js";
@@ -18,8 +18,24 @@ afterEach(() => h.cleanup());
  *  it with a changing `chatId` / `userId` (same element type at the root)
  *  reconciles in place — it does NOT remount — which is exactly how ChatView
  *  switches chats (and how a re-login swaps the user without a fresh mount). */
-function Probe({ userId = "user-1", chatId }: { userId?: string; chatId: string }) {
+function Probe({
+  userId = "user-1",
+  chatId,
+  onCommit,
+  onSetter,
+}: {
+  userId?: string;
+  chatId: string;
+  onCommit?: (value: { chatId: string; draft: string }) => void;
+  onSetter?: (value: { chatId: string; setDraft: Dispatch<SetStateAction<string>> }) => void;
+}) {
   const [draft, setDraft] = useChatDraftText(userId, chatId);
+  useLayoutEffect(() => {
+    onCommit?.({ chatId, draft });
+  }, [chatId, draft, onCommit]);
+  useLayoutEffect(() => {
+    onSetter?.({ chatId, setDraft });
+  }, [chatId, onSetter, setDraft]);
   return (
     <div>
       <span data-testid="draft">{draft}</span>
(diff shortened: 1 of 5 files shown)


yuezengwu (approved):
Reviewed head `af5c951c6`.

The change keeps the ordinary textarea mounted during loading and failure states while preserving both fail-closed layers: the disabled send control and the mutation-boundary `openRequestsUnverified` guard. The confirmed-pending path still replaces the composer with the request takeover.

The regression coverage exercises pending, cached stale-zero, failure/retry, chat switching, textarea identity/draft preservation, and the Enter-key path. I found no blocking issues.

baixiaohang (changes requested):
- Rationale: Keeping the composer mounted exposes the previous chat's draft during the existing effect-driven draft-scope transition.

**Risk level**: C
- Semantic lift: none; this changes component rendering/state behavior without changing subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Opening or switching to a chat currently swaps the composer for an open-request verification panel, which destabilizes the textarea and can disrupt an in-progress draft.
- Approach: Keep the ordinary composer mounted while verification is pending or failed, disable its send control, retain the mutation-boundary fail-closed guard, add an inline retry row on failure, and still hand control to the request takeover once a pending question is confirmed.
- Impacted modules: Workspace chat composer rendering and DOM/SSR regression coverage.

**Review findings**
❌ 1. A chat switch can render chat A's unsent draft inside chat B's composer before the scoped draft catches up. `ChatView` intentionally survives `chatId` changes, while `useChatDraftText` returns its current state during the new-chat render and only loads the new scope in a passive effect (`use-chat-draft-text.ts:28-32`). The removed `openRequestsUnverified` branch previously hid that transition; the new unconditional ordinary-composer branch now exposes it. The same effect ordering also transiently calls `saveDraft` for the new scope with the old scope's value (`use-chat-draft-text.ts:34-36`). Please make the draft value scope-correct before this stable composer is rendered and add a chat-switch regression that starts with distinct drafts and checks that the new chat never receives the previous chat's value. [R5 / `packages/web/src/pages/workspace/center/chat-view.tsx:4388`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Re-reviewing head `af5c951c6`; I am withdrawing my earlier approval.

Blocking: `ChatView` is reconciled in place across `chatId` changes, but `useChatDraftText` returns the previous scope's state for the first render of the new chat. Its first passive effect then changes `scopeRef` to the new chat and queues the new draft, while the following persistence effect still closes over the old draft and writes it under that new scope. Keeping the composer mounted therefore exposes chat A's text in chat B and transiently overwrites chat B's stored draft.

Please make the rendered draft and persistence target scope-correct before the new chat's composer is exposed. Add a regression with distinct stored drafts that inspects the immediate chat-switch render before passive effects flush and proves both that chat B never displays chat A's value and that neither scope is overwritten.

Gandy2025:
@baixiaohang Fixed in `38e0d24df`.

- `useChatDraftText` now keeps `{ scope, text }` together and resolves the incoming scope during render, so React restarts the component before committing the new chat. Chat B's first committed composer value is therefore B's draft, never A's.
- A pure scope switch performs no draft-store write. Setters persist through the scope they were created for, while only the currently committed scope may update the live composer. This also keeps an async setter from chat A from touching chat B after a fast switch.
- The committed-scope ref advances only in a layout effect, so an abandoned concurrent render cannot change which composer is considered live.

Regression coverage now uses distinct stored drafts and records layout commits before passive effects. It asserts:

- the only chat-B commit contains chat B's draft;
- switching chats performs zero `localStorage` writes;
- neither stored scope is overwritten;
- stale direct and functional setters update only their originating scope;
- the real `ChatView` switch path preserves the same guarantees while send remains fail-closed.

Validation: 80 focused/SSR tests passed, Web typecheck and design-token checks passed, and Biome/diff checks passed. The broader Web run reached 1,871/1,872 tests; the remaining reproducible failure is an unchanged `ComposeStatusBar` focus test outside this diff.

Gandy2025:
@yuezengwu Fixed in `38e0d24df`.

The draft is now scope-correct before the new chat can commit:

- state is tagged as `{ scope, text }`;
- a chat switch loads the target scope during render and React rerenders before committing descendants;
- persistence is performed by the setter against its captured scope, while a pure switch performs no storage write;
- only the scope recorded by the latest layout commit can update the live composer, so neither stale async setters nor abandoned concurrent renders can leak into the visible chat.

The regression records every layout commit with distinct A/B drafts and proves chat B never commits chat A's value. It also asserts zero storage writes during the switch, unchanged A/B storage, stale functional/direct setter isolation, and the same behavior through the real `ChatView` chat-switch path.

Validation: 80 focused/SSR tests passed, Web typecheck and design-token checks passed, and Biome/diff checks passed. The broader Web run reached 1,871/1,872 tests; the remaining reproducible failure is an unchanged `ComposeStatusBar` focus test outside this diff.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: Head `38e0d24df` resolves the cross-chat draft leak without weakening the open-request send gate.

**Risk level**: C
- Semantic lift: none; the added draft-store handling remains local component persistence rather than a Web subscription, WebSocket, or application data-layer change.

**PR summary**
- Problem: Open-request verification used to replace the chat composer, destabilizing its textarea and draft while the snapshot loaded or failed.
- Approach: Keep the ordinary composer mounted, keep the send button and mutation boundary fail-closed until verification succeeds, show inline retry feedback on failure, and retain the request takeover for confirmed pending questions. The follow-up tags draft state by user/chat scope and makes stale setters persist only to their captured scope.
- Impacted modules: Workspace chat composer, scoped local draft persistence, DOM regressions, and SSR smoke coverage.

**Review findings**
✅ 1. The previous blocker is resolved: a scope change loads tagged state during render so React restarts before committing children, while the layout-committed scope prevents an abandoned render from retargeting setters. The new hook and ChatView regressions observe the first layout commit, preserve distinct storage entries, and cover a stale setter after switching chats.
✅ 2. The stable composer still cannot send from an unverified snapshot: the native disabled control and `handleSend`'s `openRequestsUnverified` guard remain aligned, and a confirmed pending request still replaces it with the takeover.

**Action taken**
- Approved head `38e0d24df`.

yuezengwu (approved):
Reviewed head `38e0d24df`.

The draft-scope blocker is resolved. Draft state is now tagged with its scope and adjusted before the new chat subtree commits, so the previous chat's value cannot reach the newly exposed composer. Setters retain their originating scope, allowing late async updates to persist only there without changing the active composer.

The new regressions observe the first committed textarea value across distinct stored drafts, assert that a scope switch performs no storage write, and cover stale setters after a chat switch. The original disabled send control and mutation-boundary open-request guard remain fail-closed. I found no blocking issues.

PR #2098, opened 2026-07-30
fix(context): keep Plugin integrity checks PATH-stable

baixiaohang:
## Summary

- persist the exact CLI invocation used to materialize the Context Plugin
- verify the stable marketplace payload against that install-time invocation instead of re-resolving it from the Hook `PATH`
- require one repair for older manifests that do not yet record the invocation

## Why

Codex SessionStart runs with a more restricted `PATH` than an ordinary Agent shell. A freshly enabled Plugin could therefore be materialized with the globally installed First Tree entry point, then immediately be compared against the Node fallback entry point in the Hook. The two invocations execute the same CLI but differ byte-for-byte, producing a false stale-Plugin report at `bin/context-session-start`.

The install manifest now records the exact rendered invocation. Install-time verification, later SessionStart health checks, and repair planning all reuse that value, while the existing release and provider-cache digest checks continue to detect real payload drift.

## Verification

- `pnpm exec vitest run packages/shared/src/__tests__/context-integration.test.ts apps/cli/src/__tests__/context-runtime-health.test.ts apps/cli/src/__tests__/context-bundle-manifest.test.ts apps/cli/src/__tests__/context-operation.test.ts --maxWorkers=2`
- `pnpm exec biome check` on all changed files
- `git diff --check`

The full-repository `pnpm check` remains blocked by pre-existing warnings/errors outside this diff; all changed files pass Biome directly.

### DIFF
--- a/apps/cli/src/__tests__/context-bundle-manifest.test.ts
+++ b/apps/cli/src/__tests__/context-bundle-manifest.test.ts
@@ -418,9 +418,13 @@ describe("context integration bundle", () => {
     };
 
     const plan = planContextIntegrationInstall(driver, { releaseRoot });
-    installContextIntegration(driver, plan);
+    const installed = installContextIntegration(driver, plan);
     const marketplaceSource = contextIntegrationMarketplaceSourcePath("codex");
     expect(existsSync(join(marketplaceSource, "plugins", "first-tree-context"))).toBe(true);
+    expect(installed.manifest.materializedInvocation).toBeTruthy();
+    expect(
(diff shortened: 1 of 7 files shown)


yuezengwu (approved):
Reviewed at head bc9d0a0bb.

Goal: keep Context Plugin integrity checks PATH-stable by persisting the exact install-time rendered CLI invocation. The change threads that value through install verification, runtime health, and repair planning, while keeping older manifests readable and requiring a one-time repair.

I checked the manifest compatibility boundary, install/rollback state transitions, stable-source and provider-cache verification, and the differing Hook PATH regression coverage. No blocking findings.

PR #863, opened 2026-06-08
feat(web): repo resources are always On by default

Gandy2025:
## What

Repos can no longer be configured as **Opt-in** anywhere in the UI — team repos are always On by default for now. Two surfaces:

**1. Settings → Resources (`RepoEditor`)**
- Drops the **On-by-default / Opt-in** selector and pins `defaultEnabled: "recommended"`.
- A short static line replaces it: *"On by default — every agent gets this repo."*
- Editing a legacy Opt-in repo and saving normalizes it to `recommended`.
- Other resource types (MCP / Instructions / Skills) keep their selector unchanged.

**2. Agent Detail → Capabilities (`resources-tab.tsx`)**
- Since repos can never be `available` now, the per-agent **"Enable from team"** list should never offer one. Repos are excluded from the enableable set.
- The repo "+" menu now shows only **"Add agent repo…"** + the **"Manage in Settings → Resources"** link — still actionable, no dead end.
- skill / mcp opt-in is unchanged.
- Regression test added: a legacy `available` repo is not offered to enable.

## Why

Per-repo Opt-in isn't offered yet — a temporary product decision, not a schema change. The server still accepts both `recommended` and `available`; this is frontend-only.

## Scope deliberately excluded

- No server/schema change.
- No DB migration to bulk-flip existing `available` repos (they normalize on next edit-save; legacy ones just keep their "Opt-in" badge in the Settings list until then).
- No Context Tree write — `system/cloud/agents/resources.md` still defines both modes as valid; this temporary UI restriction doesn't change the durable resource model.

## Verification

- `tsc --noEmit` ✓ + design-token guardrails ✓
- `resource-editors-dom.test.tsx` ✓, `resources-savebar-dom.test.tsx` ✓ (incl. new repo-exclusion test), `agent-detail-page-dom.test.tsx` ✓
- Biome clean ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2081, opened 2026-07-29
fix(agents): harden Agent Template integration

yuezengwu:
## Summary

- integrate the official Agent Templates implementation onto the frozen integration base
- harden prompt composition, cross-Team MCP projection, catalog-wide MCP-name serialization, rolling-safe creation, provenance, CAS snapshots, and New Agent draft restoration
- replace publisher-driven adopter fan-out with an Agent-start boundary: each new runtime slot materializes one bounded, immutable Template source-layer snapshot while Team/Agent-owned layers retain their existing safe-point updates

Paired durable contract: https://github.com/agent-team-foundation/first-tree-context/pull/854

## Current draft state

This PR remains **Draft**. The original frozen implementation at `081733ff1eef25c6397eb9df1d99d1f88f06315c` failed formal isolated QA. The pushed branch contains deterministic repairs through `853dbf8e`, while the final slot-snapshot architecture is still being implemented and adversarially reviewed in a dirty task worktree; it is not yet committed or pushed.

Closed in the pushed branch:

- Team → Template → Agent prompt-budget precedence with shared exact rendering and no legacy prompt double-count
- a Template-specific cross-Team MCP contract that rejects URL query/fragment data and all non-empty stdio args without breaking ordinary same-Team MCP query/args behavior
- catalog-wide MCP runtime-name mutation serialization
- capability-safe Template Agent creation path for mixed Web/Server versions
- explicit official Template provenance in Agent Detail and generated briefings
- generated-briefing marker rejection, PostgreSQL int32 sort-order validation, stale empty-catalog selection clearing, and coherent in-transaction selection responses
- New Agent draft preservation for display name, manual handle, visibility, computer, runtime, and ordered Template selection
- concurrent Template-selection CAS regression coverage

Still blocking readiness:

- finish and review the content-addressed Template source-layer artifact plus per-Agent last-startup-attempt link, slot-scoped bind proof, fail-closed runtime composition, lifecycle cleanup, operator-visible blocker, and publisher-mutation O(1) invariants
- regenerate and review the migration after the final schema is stable, including the server-first/default-off rollout preflight
- complete the paired Context Tree PR review; merge this code PR first, then reconcile and land the Tree PR
- add a committed reusable cross-surface QA case and rerun formal isolated QA on the exact final head
- complete final CI and human review of the database migration / rollout behavior

## Verification to date

[...]

PR #2101, opened 2026-07-30
feat(server): add Agent Template persistence schema

yuezengwu:
## Summary

- add the strict shared Agent Template contract: `draft | active | retired`, canonical unordered 0–3 Template IDs, public-safe profile, stable component keys, and Prompt/complete Skill ZIP/MCP components
- add the global `agent_templates` catalog table plus `agent_configs.template_ids`
- add one-time import provenance to Team Resources and Agent Resource bindings without coupling Team-owned Resources to Template lifecycle
- protect Template Skill ZIPs through the authoritative attachment lifecycle and harden cross-Team MCP/Prompt ingestion
- add generated additive migration `0090_redundant_crystal`, Drizzle metadata, and deterministic schema/migration tests

## Product contract

This PR is technical task 1 only. An official Template is a one-time responsibility/configuration starting point. A later import service will materialize its Prompt, Skill bundle, and MCP components into the selected Team's Resources; those Team Resources and explicit bindings are the execution authority. V1 does not automatically synchronize later official Template changes.

An Agent may carry zero to three unordered Template responsibilities. Reusing a Template in one Team reuses the Team's current Resource per stable Template component key. Removing a Template will later remove only the generated bindings, not the Team Resources or manual bindings.

## Deliberate boundaries

This PR does **not** add Template CRUD/publishing services, import transactions, permissions, API/SDK routes, runtime projection, Client/Web UI, official seed content, metrics, automatic update, overlap/conflict handling, or imported Team Resource garbage collection. Existing Agents default to zero Templates and ordinary creation/runtime behavior is unchanged.

## Persistence details

- one global `agent_templates` table; no revision or selection join table
- `agent_configs.template_ids text[] NOT NULL DEFAULT '{}'`, cardinality check `<= 3`, GIN index
- Resource origin triple (`template`, `component`, initial digest), all-null/all-set check, and active/stale Team dedup identity `(organization, template, component)`
- binding origin pair, all-null/all-set check, and partial `(agent, template)` lookup index
- no Resource/Binding foreign key to `agent_templates`; retirement cannot break already imported Team configuration
- generated migration is purely additive; no update, backfill, drop, or rewrite

## Review remediation

Exact follow-up commit: `05f92c973c75411a1a3aa1bba30c88103a254a5b`.

- Template MCP uses a dedicated strict schema: HTTP/SSE reject URL userinfo, query, and fragment; stdio rejects non-empty arguments; invalid URLs are ordinary Zod failures and never throw from `safeParse`
- Template Prompt reuses the shared generated-briefing detector and rejects the conclusive `first-tree:generated` marker
- attachment reference discovery now includes Template Skill bundles; real orphan sweep tests cover retained `draft`/`active`/`retired` references, release and collection, and malformed payload isolation through a deterministic JSONB `CASE`
- superseded live-snapshot Context Tree Draft #854 is closed; #863 is the only current one-time-import contract

## Verification

- `pnpm check` — pass (16 pre-existing warnings, 1 pre-existing info)
- forced `pnpm typecheck` — 11/11 tasks, 0 cached
- forced production build — 5/5 tasks, 0 cached
- forced full `pnpm test` — 12/12 Turbo tasks, 0 cached
- Shared — 771/771 total; Agent Template 30/30
- Server — 2,848/2,848 total; targeted schema + Skill lifecycle 55/55
- migration checker and fresh Drizzle generator — pass, no schema drift

Two independent formal QA runs passed. The final run exercised exact commit `05f92c973c75411a1a3aa1bba30c88103a254a5b` in a fresh named-non-root Docker/PostgreSQL 17 harness: final built Shared exports, real attachment lifecycle across every Template status, malformed JSONB isolation, exact 0089→0090 upgrade and idempotent restart, production artifacts, and a forced uncached full suite. No task-attributable defect was found; case disposition was `no-change`.

Paired durable Context Tree decisions are recorded in Draft PR https://github.com/agent-team-foundation/first-tree-context/pull/863. It will remain Draft until this source PR merges, then be reconciled against merged source truth before being marked ready.


baixiaohang (changes requested):
- Rationale: The additive migration is well-scoped and the one-time-import boundary is substantially clearer than the retired live-reference design, but the shared contract currently admits unsafe MCP payloads, loses Template Skill bundles to existing attachment GC, and bypasses the generated-briefing prompt guard.

**Risk level**: B-high
- Path baseline: changes the Server DB structure and Agent/Resource persistence under `packages/server/**`.
- Semantic lift: none above B-high.

**PR summary**
- Problem: Teams need official, reusable Agent responsibility starters without making publisher changes a continuing authority over Team-owned runtime configuration.
- Approach: Adds a global Template catalog, records up to three unordered Template responsibilities per Agent, and prepares one-time import provenance so Prompt/Skill/MCP components become Team Resources and explicit bindings that can be reused and removed precisely.
- Impacted modules: Shared Template/Resource schemas, Agent config persistence, Team Resource and Agent binding persistence, and Drizzle migration metadata.

**Review findings**
❌ 1. The Template MCP contract still accepts credential-bearing URL data and command arguments. [R4/R5 / `packages/shared/src/schemas/agent-template.ts:113`]
`agentTemplateMcpComponentSchema` reuses `noSecretMcpServerSchema`, but that schema permits HTTP/SSE query strings and fragments plus arbitrary stdio `args`. Inputs such as `https://mcp.example/run?token=...` or `["--token=..."]` therefore pass even though this component is documented as secret-free; a later import would persist and execute those values in another Team. Please restore a Template-specific MCP schema that rejects URL search/hash data and non-empty stdio arguments (or an equivalently complete cross-Team secret boundary), with regressions for both cases.

❌ 2. A Template Skill's ZIP reference is invisible to the attachment lifecycle and will be swept as an orphan. [R3/R5 / `packages/shared/src/schemas/agent-template.ts:103`]
The new durable reference lives only at `agent_templates.payload.components[].bundle.attachmentId`. Existing `isAttachmentReferenced` checks `resources.bundle_attachment_id` and message JSON only, while `sweepOrphanAttachments` deletes ready uploads older than 24 hours that fail that scan. A bundle used only by a Template can therefore disappear before any Team imports it. Please make Template bundle ownership/reference discoverable to the authoritative attachment lifecycle (for example, via a first-class relation or a complete Template reference check) and cover the orphan-GC path.

❌ 3. Template Prompt ingestion bypasses the generated-briefing hard guard. [R1/R5 / `packages/shared/src/schemas/agent-template.ts:86`]
The reused `promptResourcePayloadSchema` accepts a body containing `first-tree:generated`. The current steering contract requires server-side prompt writes carrying that marker to fail closed, because otherwise an assembled `AGENTS.md` is frozen back into a managed prompt and duplicates Team/runtime instructions. Templates are now another managed prompt-ingestion path, so enforce the shared briefing guard in the Template Prompt schema/service and add a regression.

⚠️ 4. The superseded live-reference Tree draft remains open beside the new one-time-import draft. [R2 / `first-tree-context#854`]
The former paired code PR #2081 is closed, but draft Tree PR #854 still proposes ordered, runtime-fetched Template snapshots and conflicts with paired draft #863. Please close or explicitly supersede #854 so there is one reviewable future contract.

**Action taken**
- Submitted request changes against exact head `1af941cfb95dc0d733d99980610ac6d0142bd97a`.

yuezengwu:
@baixiaohang Addressed all four findings in exact head `05f92c973c75411a1a3aa1bba30c88103a254a5b`:

1. **MCP cross-Team secret boundary** — added exported Template-specific strict MCP schema. HTTP/SSE reject URL userinfo, query, and fragment; stdio rejects non-empty `args`. Added query/hash, credential, args, valid-shape, and malformed-URL regressions. During independent review we also caught and fixed a Zod refinement edge where invalid URLs could make `safeParse` throw; malformed URLs now return ordinary failures.
2. **Skill ZIP attachment lifecycle** — `isAttachmentReferenced` now discovers `agent_templates.payload.components[].bundle.attachmentId`. Real PostgreSQL sweep coverage proves Template-only references survive, released bundles collect, and malformed payload rows cannot block GC. The executed query protects `jsonb_array_elements` with an inline `CASE`, not planner-dependent boolean short-circuiting.
3. **Generated briefing guard** — Template Prompt payloads directly reuse `findAssembledBriefingFingerprint` and reject only the same conclusive `generated-marker` tier as server prompt writes. Clean prompt coverage remains positive.
4. **Superseded Tree contract** — closed `first-tree-context#854` with an explicit supersession note; Draft #863 is now the sole one-time-import contract.

Independent formal QA was repeated from a fresh named-non-root Docker/PostgreSQL 17 cell on this exact head. Final artifact probes, `draft`/`active`/`retired` attachment lifecycle, malformed-row recovery, 0089→0090 upgrade/restart, Drizzle no-drift, forced build/typecheck, and forced no-cache full tests all passed. The PR body has the updated scope and evidence summary.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact follow-up head closes every previously reported contract and lifecycle blocker without introducing a new review finding.

**Risk level**: B-high
- Path baseline: changes Server DB structure and Agent/Resource persistence under `packages/server/**`.
- Semantic lift: none above B-high.

**Follow-up review**
- Exact head: `05f92c973c75411a1a3aa1bba30c88103a254a5b`
- The Template-specific MCP schema now rejects URL credentials/query/fragment data and non-empty stdio arguments.
- Template Skill ZIPs now participate in authoritative attachment reference discovery, including malformed-payload isolation and released-orphan cleanup.
- Template Prompt ingestion now applies the shared generated-briefing hard guard.
- The superseded live-snapshot Tree draft is closed, leaving paired draft #863 as the single one-time-import contract.
- No new blocking finding was found in the remediation delta.

**Action taken**
- Approved with explicit maintainer authorization. No merge action was taken.

PR #2102, opened 2026-07-30
fix(web): align TeamSwitcher current-team rows on the shared menu grid

yuezengwu:
## Problem

Follow-up to #2091 (merged). In the reorganized TeamSwitcher menu, the **Current team** group's action rows (`Invite teammates`, `Leave team`) carried their own inset — an extra `px-2` inside the already-padded section container, `rounded` corners, and `mt-2` between rows — so their icons sat ~8px deeper than the rows in **Switch team** and **Add or join**, and the row rhythm differed between groups.

## Approach

Presentation-only alignment, no behavior changes:

- Extract one shared `MENU_ROW_CLASS` (same gutter, icon column, and full-bleed hover hit area) now used by every action row: Invite / Leave / Switch / Create / Join — so the groups can't drift apart again.
- Restructure the Current team section to match the other groups: eyebrow uses the same section padding as `Switch team` / `Add or join`; the identity row (avatar + name + role + rename pencil) stays taller but sits on the same gutter.
- Remove the Current-actions-only `mt-2` / `px-2` / `rounded` treatment; Invite/Leave hover hit areas are now full-bleed like every other row.

Per-row color (destructive red for Leave) and disabled opacity semantics are unchanged, as are the rename dialog, Ask-agent nav-lock guards, and all switch/create/join/leave/invite behavior.

## Tests

- New DOM regression: every menu action row across the three groups shares one row class (asserts no `mt-2` / `px-2` / `rounded` reappears), and all three section eyebrows share the same padding.
- `packages/web` vitest: 217 files / 1882 tests pass (`tests/visual-regression.spec.ts` scaffold "no test suite" failure is pre-existing on `main`, unrelated).
- `tsc --noEmit` + design-token guard + Biome: clean.
- Visual check on the real `/preview/team-switcher` in Chromium, light + dark: Current team rows now align with the other groups; keyboard menu semantics covered by existing roving-focus tests.

### DIFF
--- a/packages/web/src/components/__tests__/team-switcher-dom.test.tsx
+++ b/packages/web/src/components/__tests__/team-switcher-dom.test.tsx
@@ -266,6 +266,40 @@ describe("TeamSwitcher", () => {
     await act(async () => root.unmount());
   });
 
+  it("aligns every menu action row and eyebrow on one shared grid across groups", async () => {
+    const { container, root } = await renderHarness(vi.fn(async () => {}));
+
+    await click(anchorOf(container));
+
+    const rows = [
+      buttonByText(container, "Invite teammates"),
+      buttonByText(container, "Leave team"),
+      buttonByText(container, "Globex"),
+      buttonByText(container, "Create team"),
+      buttonByText(container, "Join with invite link"),
+    ];
+    if (rows.some((row) => !row)) throw new Error("menu action rows missing");
+
+    // One shared row grid across Current team / Switch team / Add or join:
+    // same gutter, icon column, and full-bleed hover hit area — no
+    // Current-team-only inset, rounding, or extra row spacing.
+    const rowClasses = new Set(rows.flatMap((row) => (row ? [row.className] : [])));
+    expect(rowClasses.size).toBe(1);
+    const [rowClass] = rowClasses;
+    expect(rowClass).toContain("px-3.5");
+    expect(rowClass).toContain("py-1.5");
+    expect(rowClass).not.toContain("mt-2");
+    expect(rowClass).not.toContain("px-2");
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change centralizes the intended menu-row layout without altering action handlers or menu semantics, and the focused DOM regression protects the alignment invariant.

**Risk level**: C
- Semantic lift: none; this is presentation-only and does not touch subscriptions, WebSocket parsing, the data layer, or state semantics.

**PR summary**
- Problem: Invite and Leave appeared inset relative to the other TeamSwitcher actions, making the current-team section visually inconsistent.
- Approach: Reuse one action-row class across all three groups, align the current-team heading and identity row to the same gutter, and preserve each action's existing color, disabled state, and behavior.
- Impacted modules: `packages/web/src/components/team-switcher.tsx` and its DOM regression suite.

**Review findings**
✅ The shared row class removes the duplicated spacing configuration while retaining per-action color and opacity semantics.
✅ The current-team restructure leaves menuitem roles, roving-focus participation, click handlers, and Ask-agent navigation guards unchanged.
✅ The regression covers Invite, Leave, Switch, Create, and Join rows plus all three section headings, including explicit checks against the previous inset/rounding/spacing classes.

**Action taken**
- Approved commit `4ef758ca6b835d654d50427cb74ae23d27d4bd25`.

yuezengwu:
Superseded by #2103. The source tree is byte-for-byte identical; #2103 only re-records the commit under the signed yuezengwu GitHub identity so CLA can evaluate it, without rewriting or force-pushing this published branch.

yuezengwu:
recheck

PR #2103, opened 2026-07-30
fix(web): align TeamSwitcher current-team rows on the shared menu grid

yuezengwu:
## Problem

Follow-up to #2091 (merged). In the reorganized TeamSwitcher menu, the **Current team** action rows (`Invite teammates`, `Leave team`) carried a second inset and `mt-2`, so their icon baseline, hover width, and vertical rhythm did not align with **Switch team** and **Add or join**.

This PR supersedes #2102 with the exact same source tree. The earlier PR's sole commit used an agent-local author identity that the CLA bot could not map to a GitHub account; this replacement preserves the reviewed code byte-for-byte under the signed `yuezengwu` identity without rewriting or force-pushing the published branch.

## Approach

Presentation-only alignment, with no behavior changes:

- Extract one shared `MENU_ROW_CLASS` used by Invite / Leave / Switch / Create / Join.
- Give every action row the same 14px gutter, icon column, vertical padding, and full-width hover hit area.
- Give the Current team eyebrow the same section padding as the other two groups.
- Keep the taller identity row on the same gutter while preserving its avatar, name, role, and rename action.
- Remove only the Current-actions-specific `mt-2`, secondary `px-2`, and rounded hover treatment.

Leave remains destructive red. Disabled states, rename, Ask-agent navigation locks, switch/create/join/leave/invite behavior, and Context Tree setup routing are unchanged.

## Verification

- The replacement commit tree exactly matches #2102 head `4ef758ca6` (`c80a0a67f1d66781b5a6475414e761c43b0e46c1`).
- Added a deterministic DOM regression that requires all action rows to share one grid and all section eyebrows to share one padding contract.
- Kiven-FTE reported 217 Web test files / 1882 cases passing, plus Web typecheck, design-token guard, and Biome.
- Real `/preview/team-switcher` Chromium light/dark checks show aligned rows and full-width hover treatment.
- Independent formal QA is being run against this exact replacement head.

### DIFF
--- a/packages/web/src/components/__tests__/team-switcher-dom.test.tsx
+++ b/packages/web/src/components/__tests__/team-switcher-dom.test.tsx
@@ -266,6 +266,40 @@ describe("TeamSwitcher", () => {
     await act(async () => root.unmount());
   });
 
+  it("aligns every menu action row and eyebrow on one shared grid across groups", async () => {
+    const { container, root } = await renderHarness(vi.fn(async () => {}));
+
+    await click(anchorOf(container));
+
+    const rows = [
+      buttonByText(container, "Invite teammates"),
+      buttonByText(container, "Leave team"),
+      buttonByText(container, "Globex"),
+      buttonByText(container, "Create team"),
+      buttonByText(container, "Join with invite link"),
+    ];
+    if (rows.some((row) => !row)) throw new Error("menu action rows missing");
+
+    // One shared row grid across Current team / Switch team / Add or join:
+    // same gutter, icon column, and full-bleed hover hit area — no
+    // Current-team-only inset, rounding, or extra row spacing.
+    const rowClasses = new Set(rows.flatMap((row) => (row ? [row.className] : [])));
+    expect(rowClasses.size).toBe(1);
+    const [rowClass] = rowClasses;
+    expect(rowClass).toContain("px-3.5");
+    expect(rowClass).toContain("py-1.5");
+    expect(rowClass).not.toContain("mt-2");
+    expect(rowClass).not.toContain("px-2");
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The replacement head is source-tree-identical to the already reviewed #2102 change, centralizes the intended menu-row layout, and leaves TeamSwitcher behavior and accessibility semantics unchanged.

**Risk level**: C
- Semantic lift: none; this is presentation-only and does not touch subscriptions, WebSocket parsing, the data layer, or state semantics.

**PR summary**
- Problem: Invite and Leave appeared inset relative to the other TeamSwitcher actions, making the current-team section visually inconsistent.
- Approach: Reuse one action-row class across all three groups, align the current-team heading and identity row to the same gutter, and preserve each action's existing color, disabled state, and behavior.
- Impacted modules: `packages/web/src/components/team-switcher.tsx` and its DOM regression suite.

**Review findings**
✅ The shared row class removes duplicated spacing configuration while retaining per-action color and opacity semantics.
✅ The current-team restructure leaves menuitem roles, keyboard participation, click handlers, and Ask-agent navigation guards unchanged.
✅ The regression covers Invite, Leave, Switch, Create, and Join rows plus all three section headings, including explicit checks against the previous inset, rounding, and spacing classes.
✅ Commit `4d9e670b4aa4fc8657ea551cfdffb04450018ab5` has tree `c80a0a67f1d66781b5a6475414e761c43b0e46c1`, exactly matching the previously approved #2102 head.

**Action taken**
- Approved commit `4d9e670b4aa4fc8657ea551cfdffb04450018ab5`.

Issue #1718, opened 2026-07-12
[PERF-055][High] Hot inbox, token-usage, and member-activity queries lack aligned indexes


bestony:
## English

- **Severity:** High
- **Location:** `packages/server/src/db/schema/inbox-entries.ts:43-45`, `packages/server/src/services/inbox.ts:116-150`, `packages/server/src/services/inbox.ts:399-416`, `packages/server/src/db/schema/session-events.ts:30-45`, `packages/server/src/services/session-event.ts:379-387`, `packages/server/src/db/schema/messages.ts:60-70`, `packages/server/src/services/member.ts:20-37`
- **Estimated remediation:** 0.5-1 day including migration and query-plan tests
- **Description:** The pending inbox index orders by `created_at` while claims order by `id`; chat token usage lacks a chat-leading partial index; member last-activity groups by sender without a sender-leading message index. These hot paths require avoidable filtering, sorting, or broad scans as history grows.
- **Recommended direction:** Add query-shaped partial/composite indexes, verify them with representative `EXPLAIN (ANALYZE, BUFFERS)`, and codify plan/latency benchmarks.

## 中文：高频 Inbox、Token 用量和成员活动查询缺少匹配索引

- **严重性：** 高
- **位置：** `packages/server/src/db/schema/inbox-entries.ts:43-45`、`packages/server/src/services/inbox.ts:116-150`、`packages/server/src/services/inbox.ts:399-416`、`packages/server/src/db/schema/session-events.ts:30-45`、`packages/server/src/services/session-event.ts:379-387`、`packages/server/src/db/schema/messages.ts:60-70`、`packages/server/src/services/member.ts:20-37`
- **预计修复时间：** 0.5-1 天，包括迁移和查询计划测试
- **描述：** 待处理 Inbox 的索引按 `created_at` 排序，而领取操作按 `id` 排序；聊天 Token 用量查询缺少以聊天字段为首列的部分索引；成员最后活动查询按发送者分组，却没有以发送者字段为首列的消息索引。随着历史记录增长，这些高频路径会产生本可避免的过滤、排序或大范围扫描。
- **建议方向：** 添加与查询形态匹配的部分索引或复合索引，使用具有代表性的 `EXPLAIN (ANALYZE, BUFFERS)` 验证索引，并将查询计划和延迟基准固化为测试。

---

- **Audit date:** 2026-07-12
- **Source revision:** `ea2a9241fd8e128e89236cda2542f180731e3207`

PR #2108, opened 2026-07-31
fix: optimize chat token usage reads

baixiaohang:
## Summary

- add a chat-leading partial index for `session_events` token usage aggregates
- expose `access` and `aggregate` stages through the route's `Server-Timing` header
- reduce token usage polling from every 5 seconds to every 60 seconds while visible, pause it in the background, and refresh immediately on focus
- add query-plan, route timing, and React Query cadence coverage

This addresses part of #1718 and #1697.

## Production rollout

`session_events` is a hot write table and Drizzle runs migrations in a transaction. Before applying migration `0091_naive_pestilence` to a production database with a large `session_events` table, create the index outside a transaction:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_session_events_token_usage_chat"
  ON "session_events" ("chat_id")
  WHERE "kind" = 'token_usage';
```

Then:

1. verify the definition in `pg_indexes`
2. verify `pg_index.indisvalid = true`
3. apply the normal migration; its `CREATE INDEX IF NOT EXISTS` will skip the pre-created index
4. run `EXPLAIN (ANALYZE, BUFFERS)` against the production token usage aggregate for representative large, medium, and small chats

## Validation

- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2 src/__tests__/chat-token-usage-performance.test.ts src/__tests__/session-event.test.ts src/__tests__/me-chats-http-e2e.test.ts` — 34 passed
- `pnpm --filter @first-tree/web exec vitest run --passWithNoTests --maxWorkers=2 src/pages/workspace/center/__tests__/chat-token-usage-query.test.tsx src/pages/workspace/center/__tests__/chat-view-dom.test.tsx` — 65 passed
- targeted Biome check and `git diff --check` passed
- `pnpm --filter @first-tree/server db:generate` reported no schema changes after generation

### DIFF
--- a/packages/server/drizzle/0091_naive_pestilence.sql
+++ b/packages/server/drizzle/0091_naive_pestilence.sql
@@ -1,0 +1,18 @@
+-- Chat-scoped token usage aggregation index.
+--
+-- Operator note:
+-- Drizzle migrator executes migration files inside a transaction, so this file
+-- cannot use CREATE INDEX CONCURRENTLY. Before applying this migration to a
+-- production database with a large session_events table, create the index
+-- outside a transaction:
+--
+--   CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_session_events_token_usage_chat"
+--     ON "session_events" ("chat_id")
+--     WHERE "kind" = 'token_usage';
+--
+-- Then verify its definition in pg_indexes and run EXPLAIN (ANALYZE, BUFFERS)
+-- for the production token usage aggregate. The IF NOT EXISTS below makes the
+-- transactional migration skip the index after that online pre-deploy step.
+CREATE INDEX IF NOT EXISTS "idx_session_events_token_usage_chat"
+  ON "session_events" USING btree ("chat_id")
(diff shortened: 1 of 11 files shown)


yuezengwu (approved):
Reviewed head `3aee93ff8`.

The change adds a chat-leading partial index for the `session_events` token-usage aggregate, exposes `access` and `aggregate` stages through `Server-Timing`, and changes the Web query to poll once per minute only while visible with an immediate focus refresh. The migration snapshot is scoped to the intended index, and the route/query cadence coverage matches the implementation.

No blocking findings in the diff.

Database change — human/operator check required before production rollout: `session_events` is a hot write table and the normal Drizzle migration is transactional. Pre-create `idx_session_events_token_usage_chat` with `CREATE INDEX CONCURRENTLY`, verify both its definition and `pg_index.indisvalid`, then run representative `EXPLAIN (ANALYZE, BUFFERS)` checks before allowing the normal migration to skip the existing index, as documented in the PR.

PR #2115, opened 2026-07-31
fix: derive the team slug server-side on self-service create

Gandy2025:
## Problem

A user created a team named `电商平台`, renamed it to `电商平台-商城`, then tried to create `电商平台` again and was told the team already existed.

`POST /me/organizations` let the client choose `organizations.name`, and the Web modal derived it from the display name by replacing every character outside `[a-z0-9-]`. Any display name without an ASCII alphanumeric — every all-CJK team name — collapsed onto the same `"team"` fallback slug:

```
"电商平台"      -> "team"
"电商平台-商城"  -> "team"
"设计团队"      -> "team"
```

That slug is globally unique, not per-tenant. So the first such team **anywhere in the deployment** claimed `team`, and from then on any account creating a team with an all-CJK name got `409 Organization "team" already exists` — an identifier the user never typed, on a form that only asks for a display name. Renaming could not release it either: both rename surfaces (`team-switcher.tsx`, onboarding `step-team.tsx`) only patch `displayName`, so the slug never moves after creation.

Signup was unaffected: `createPersonalTeam` already derives its slug server-side and disambiguates through `insertOrgWithSlugRetry`. Only the self-service "Create team" path was missing that, which is why this surfaces the moment a user wants a second team.

A secondary failure mode: names leaving a single ASCII alphanumeric (`"A团队"` -> `"a"`, `"团队1"` -> `"1"`) failed the request schema's `min(2)` and returned a raw Zod validation error.

## Change

Treat the slug as what it already is everywhere else — a server-derived internal identifier:

- `createOrgFromMeSchema` drops `name`; the request carries only `displayName`.
- `selfCreateOrganization` derives the slug from the display name and reuses `insertOrgWithSlugRetry`, so collisions are disambiguated silently instead of surfacing as a user-visible failure.
- The Web modal stops slugifying and sends only the display name. Its comment claiming the server had a `pickAvailableOrgSlug` collision-disambiguation helper went with it — no such helper existed, which is exactly why this path was unguarded.

`"default"` loses its special case: the default org always exists, so the retry helper already moves past that slug rather than needing a separate reserved-name branch.

Display names remain deliberately non-unique — two teams may share one, and there is now no user-visible naming conflict at all.

## Scope notes

- No migration. The existing row holding `team` stops blocking anyone once slugs are server-derived.
- `organization.ts`'s admin/webhook `createOrganization` is untouched; there the caller supplies `name` explicitly and a 409 is the correct semantic.
- `updateOrganizationSchema` still accepts `name`. No Web surface sends it, but an API caller can still rename a slug into a 409. Removing it is a separate, independently revertable change.
- `sanitizeOrgSlug` can still return a 1-character slug. It is not routed, not rendered, and uniqueness is handled by the retry, so this is left alone rather than given a new minimum-length rule.

## Tests

- `me-multi-org.test.ts`: three teams whose display names all derive the same base slug (including the reported rename-then-reuse sequence) each create successfully with distinct slugs, and two of them keep the same display name.
- `members.test.ts`: the old "rejects duplicate and reserved slugs" case is replaced by one asserting a display name deriving `default` gets a suffixed slug and leaves the default org intact.
- `team-setup-modal-dom.test.tsx`: asserts the request body carries only `displayName`.

`pnpm check`, `pnpm typecheck`, and the server (2855) / web (1940) / shared (777) suites pass. The one unrelated failure in the monorepo run is a 20s timeout in `skill-evals`' `first-tree-seed` real-source fixture case, which passed at 19.25s on the preceding run.

### DIFF
--- a/packages/server/src/__tests__/api-small-routes-extra.test.ts
+++ b/packages/server/src/__tests__/api-small-routes-extra.test.ts
@@ -526,7 +526,7 @@ describe("small API route handlers", () => {
     await meRoutes(created.app as never);
     const createReply = makeReply();
     await route(created.routes, "POST", "/me/organizations").handler(
-      { body: { name: "new-org", displayName: "New Org" }, user: { userId: "user_1" } },
+      { body: { displayName: "New Org" }, user: { userId: "user_1" } },
       createReply,
     );
     expect(createReply).toMatchObject({
@@ -542,7 +542,7 @@ describe("small API route handlers", () => {
     await meRoutes(missingUser.app as never);
     await expect(
       route(missingUser.routes, "POST", "/me/organizations").handler(
-        { body: { name: "new-org", displayName: "New Org" }, user: { userId: "user_missing" } },
+        { body: { displayName: "New Org" }, user: { userId: "user_missing" } },
         makeReply(),
       ),
     ).rejects.toThrow("User not found");
--- a/packages/server/src/__tests__/me-multi-org.test.ts
+++ b/packages/server/src/__tests__/me-multi-org.test.ts
@@ -96,21 +96,38 @@ describe("Multi-org self-service", () => {
     // Display names carrying no ASCII alphanumerics all sanitize to the same
     // base slug. The slug is globally unique, so while the client chose it the
     // first such team anywhere rejected every later one with a 409 quoting an
-    // identifier the user never typed. Renaming did not help either: renames
-    // only touch displayName.
+    // identifier the user never typed.
     const first = await create("电商平台");
-    const renamedThenReused = await create("电商平台-商城");
-    const duplicate = await create("电商平台");
-    expect([first.statusCode, renamedThenReused.statusCode, duplicate.statusCode]).toEqual([201, 201, 201]);
+    expect(first.statusCode).toBe(201);
+    const firstId = first.json<{ organization: { id: string } }>().organization.id;
+    const [beforeRename] = await app.db.select().from(organizations).where(eq(organizations.id, firstId));
 
-    const ids = [first, renamedThenReused, duplicate].map(
-      (res) => res.json<{ organization: { id: string } }>().organization.id,
-    );
-    const rows = await app.db.select().from(organizations).where(inArray(organizations.id, ids));
-    expect(rows.length).toBe(3);
-    expect(new Set(rows.map((row) => row.name)).size).toBe(3);
+    // Renaming does not release the slug — it moves only the display name.
+    const renamed = await app.inject({
+      method: "PATCH",
+      url: `/api/v1/orgs/${firstId}`,
+      headers: { authorization: `Bearer ${admin.accessToken}` },
+      payload: { displayName: "电商平台-商城" },
+    });
+    expect(renamed.statusCode).toBe(200);
+    const [afterRename] = await app.db.select().from(organizations).where(eq(organizations.id, firstId));
+    expect(afterRename?.displayName).toBe("电商平台-商城");
+    expect(afterRename?.name).toBe(beforeRename?.name);
+
+    // Reusing the original display name still succeeds, on a fresh slug.
+    const reused = await create("电商平台");
+    expect(reused.statusCode).toBe(201);
+    const reusedId = reused.json<{ organization: { id: string } }>().organization.id;
+
+    const rows = await app.db
+      .select()
+      .from(organizations)
+      .where(inArray(organizations.id, [firstId, reusedId]));
+    expect(rows.length).toBe(2);
+    expect(new Set(rows.map((row) => row.name)).size).toBe(2);
     // Display names are deliberately not unique — two teams may share one.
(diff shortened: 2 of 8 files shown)


Gandy2025:
Context Tree companion (draft until this merges): agent-team-foundation/first-tree-context#866

yuezengwu (approved):
Approved. The implementation goal is clear and the core change is coherent end to end: `POST /me/organizations` now accepts only the user-facing `displayName`, `selfCreateOrganization` derives the internal slug server-side through the existing collision-safe allocator, and the Web client no longer manufactures or submits a slug. This preserves display-name non-uniqueness while keeping the deployment-global slug constraint out of the product surface.

I reviewed the shared DTO, route/service flow, allocator semantics, Web caller, and the updated server/Web coverage. I found no blocking issue. There is no database schema or migration change; the externally visible request DTO does change, although stale clients that still send `name` remain compatible because the Zod object strips that unknown field.

Two non-blocking follow-ups:

- The linked Context Tree companion #866 currently says that a Team slug is always derived from the display name. Signup still seeds `createPersonalTeam` from `username`, so the durable wording should say the slug is derived from a server-controlled seed, with self-service creation specifically using the display name.
- `me-multi-org.test.ts` covers the collision topology, but the case described as the reported rename-then-reuse sequence creates three teams without performing a rename. The behavior under review is still covered; consider making the test/PR wording precise or exercising the rename explicitly.

Per reviewer policy, I did not run tests locally. CI lint/typecheck and the completed static checks are green; the test jobs were still running when this review was submitted.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The self-service create path now has one server-owned slug allocation path, reuses the existing conflict-safe retry helper, and removes the hidden client-side uniqueness failure without changing the admin/internal creation contract.

**Risk level**: B-low
- Path baseline: `packages/server/**` plus Shared/Web changes; the touched membership service is not in the B-high core-service list, and there is no database schema migration.
- Semantic lift: none.

**PR summary**
- Problem: A person creating another Team could receive a 409 for a hidden global slug—especially for all-CJK display names that all collapsed to `team`—and one-character derived slugs could fail schema validation even though the form only asked for a display name.
- Approach: Accept only `displayName` on the self-service endpoint, derive and disambiguate the internal slug on the Server through the existing `insertOrgWithSlugRetry` path, and remove Web-side slug derivation. Existing clients that still send `name` remain compatible because the new Zod object strips the extra field.
- Impacted modules: Shared self-service DTO, Server `/me/organizations` route and membership service, Web Team setup modal, and their focused tests.

**Review findings**
✅ 1. Slug allocation for self-service creation now has a single authoritative implementation and handles concurrent/global collisions through the database uniqueness target.
✅ 2. The focused tests cover repeated derived bases, duplicate display names, the reserved default slug, and the Web request contract; all current PR checks pass.
⚠️ 3. Non-blocking paired-Tree follow-up: first-tree-context#866 currently says a client never chooses the slug, while this PR intentionally leaves `updateOrganizationSchema.name` available to API callers. Please scope that durable statement to self-service creation, or reconcile the remaining update path, before merging the Tree PR.

**Action taken**
- Approved.

yuezengwu (approved):
Re-reviewed head `1dd78123f19ca6c601e0bb8bf70b326dcdabfc4d`.

The new commit is test-only and directly addresses the prior advisory: the regression now creates the CJK-named Team, renames it through `PATCH /orgs/:orgId`, proves the internal slug is unchanged, and then proves both reuse of the original display name and duplicate display names still succeed with distinct slugs.

No new blocking issue. The previously reviewed production change is unchanged, and the linked Context Tree companion #866 has also been tightened to distinguish the signup username seed, the self-service display-name seed, and the explicit administrative create/update exception.

Per reviewer policy, I did not run tests locally. CI for this new head is still running.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The synchronized head changes only the regression test and now exercises the reported rename-then-reuse workflow directly; the production implementation remains unchanged and the new assertions match its intended contract.

**Risk level**: B-low
- Path baseline: unchanged from the original review (`packages/server/**` plus Shared/Web; no B-high service or database schema migration).
- Semantic lift: none.

**PR summary**
- Problem: Self-service Team creation exposed a hidden, deployment-global slug collision to people entering only a display name.
- Approach: Keep slug allocation server-owned for self-service creation, silently disambiguate collisions, and have the Web client submit only the display name.
- Incremental change: `1dd78123` updates only `me-multi-org.test.ts`.

**Review findings**
✅ 1. The regression now PATCHes the first Team's display name, verifies the internal slug is unchanged, then reuses the original display name and verifies the new Team receives a distinct slug.
✅ 2. The paired Context Tree draft now distinguishes the username seed used at sign-in from the display-name seed used for self-service creation, and explicitly preserves administrative create/update callers as the client-supplied-slug exception.
✅ 3. No production code changed after the previously approved head.

**Action taken**
- Re-approved head `1dd78123`.

PR #868, opened 2026-06-08
refactor(web): agent-detail PR1 — IA rename, repos→Environment, Instructions disclosure, TabBar

Gandy2025:
## What & why

PR1 of 3 in the agent-detail page optimization (structure & navigation). Design is locked; this PR is the IA / layout pass. PR2 (save semantics + leave-guard) and PR3 (agent switcher) follow.

**Key correction baked in:** repos / skills / MCP are saved **immediately** via the agent-resources binding API (not the SaveBar config draft) — only model/effort/env are draft-backed. The old `git`/`mcp` draft-section components were already dead (test-only). This PR treats repos as immediate-save accordingly.

## Changes

- **IA labels:** Runtime → **Environment**, Capabilities → **Tools & skills**. Route `path`s and draft/anchor keys are unchanged to preserve deep links; `SECTION_TO_TAB.git` and the legacy `agent-cfg-git` anchor now resolve to the `runtime` tab.
- **Repos → Environment:** moved out of Tools & skills into an independent **Repositories** section on Environment (kept separate from env vars so the tab doesn't mix immediate-save and draft semantics in one block — PR2 will formalize the zoning). Extracted a shared `capability-section.tsx` (`useAgentResources` hook + presentational `ResourceTypeSection`) consumed by both tabs via the same `["agent-resources", uuid]` React Query cache. No shell-state lift.
- **Instructions:** enabled and disabled blocks now collapse to a 2-line summary with a **Show full** toggle (unified disclosure); the standalone "Available team instructions" section is folded into the section **Add** menu (custom + enable-from-team + Settings).
- **Tools & skills:** mcp subtitle → "From the MCP servers you connect."; tab badge shows the effective skill+mcp count (the shell fetches `agent-resources` once, `enabled: !isHuman`, same tier as agent-config/client-status; tabs become cache hits).
- **TabBar:** `height` → `minHeight` (fixes a spurious vertical scrollbar); overflow edge fades; `scrollIntoView` on the active tab for programmatic navigation (SaveBar "Jump to" / deep links); hidden when there's only one tab.
- **Usage:** "Show more" grows the turns page size (single query, capped 200 server-side).
- **Hardening (from review):** the shared resource mutation cancels in-flight GETs on success (prevents a stale GET clobbering the just-written version) and refetches on 409 (so a retry uses the latest version).
- Synced the DEV preview gallery + DOM/SSR tests.

## Out of scope (follow-ups)

[...]

PR #2117, opened 2026-07-31
refactor: drop the org slug update path

Gandy2025:
Follow-up to #2115, which left this open deliberately.

## Problem

`updateOrganizationSchema` still accepted `name`, so `PATCH /orgs/:orgId` could rewrite an organization's slug. No Web surface sends it — both rename affordances patch only `displayName` — but the route is reachable by any Team admin, and Team admin is a product role every member who owns a Team holds, not an operator privilege. Every user is admin of the team auto-created at sign-in.

The slug namespace is global to the deployment, so that update path carried the same cross-tenant shape #2115 just removed from create: an admin could claim a slug another tenant might want, or enumerate which slugs exist by reading 409 against 200.

## Change

- `updateOrganizationSchema` drops `name`. The slug is fixed once the row exists.
- `updateOrganization` drops the corresponding assignment. With no unique column left in the update set, nothing there can raise 23505, so the conflict handler is unreachable and goes with it.
- Provisioning callers are unaffected: `createOrganization` still takes an explicit `name` and still reports a conflict, which is correct for a caller that chose an exact slug.

Requests that still send `name` are not rejected — the Zod object strips the unknown field, so stale clients keep working and simply no longer move the slug.

## Scope notes

`maxAgents`, `maxMessagesPerMinute`, and `features` remain settable through this same admin route. That is a separate and larger question than the slug — `maxAgents` is enforced in `services/agent.ts`, so a Team admin can currently raise their own agent limit — and it is deliberately left alone here rather than folded into a slug change. Flagged for a decision rather than fixed in passing.

## Tests

- `organization-extra.test.ts`: the update case no longer passes `name` and now asserts the built update set never carries one; the unreachable conflict assertion is dropped.
- `me-multi-org.test.ts`: the existing rename step now also sends `name: "squatted-slug"` and asserts the slug is unchanged, so the stripping behaviour is covered end to end through the route.

`pnpm check`, `pnpm typecheck`, server (2858) and shared (779) suites pass.

## Context Tree

Paired tree change: agent-team-foundation/first-tree-context#868 (draft until this merges). #866 recorded the administrative exception as covering both create and update; this narrows it to creation only.

### DIFF
--- a/packages/server/src/__tests__/me-multi-org.test.ts
+++ b/packages/server/src/__tests__/me-multi-org.test.ts
@@ -103,11 +103,13 @@ describe("Multi-org self-service", () => {
     const [beforeRename] = await app.db.select().from(organizations).where(eq(organizations.id, firstId));
 
     // Renaming does not release the slug — it moves only the display name.
+    // A `name` in the body is not an escape hatch either: the slug has no
+    // update path, so it is stripped rather than applied.
     const renamed = await app.inject({
       method: "PATCH",
       url: `/api/v1/orgs/${firstId}`,
       headers: { authorization: `Bearer ${admin.accessToken}` },
-      payload: { displayName: "电商平台-商城" },
+      payload: { displayName: "电商平台-商城", name: "squatted-slug" },
     });
     expect(renamed.statusCode).toBe(200);
     const [afterRename] = await app.db.select().from(organizations).where(eq(organizations.id, firstId));
--- a/packages/server/src/__tests__/organization-extra.test.ts
+++ b/packages/server/src/__tests__/organization-extra.test.ts
@@ -77,7 +77,7 @@ describe("organization service edge cases", () => {
     ).rejects.toThrow("Unexpected: INSERT RETURNING produced no row");
   });
 
-  it("reads and updates organizations with not-found and conflict handling", async () => {
+  it("reads and updates organizations with not-found handling", async () => {
     await expect(getOrganization(makeSelectDb([{ id: "org_1" }]) as never, "org_1")).resolves.toEqual({ id: "org_1" });
     await expect(getOrganization(makeSelectDb([]) as never, "missing")).rejects.toThrow(
       'Organization "missing" not found',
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change removes the only API/service path that could mutate an existing Team slug while preserving creation-time provisioning and stale-client compatibility.

**Risk level**: B-low
- Path baseline: `packages/server/**` plus `packages/shared/**`; the touched organization service is outside the B-high core-service list, and there is no database schema or migration change.
- Semantic lift: none.

**PR summary**
- Problem: Any Team admin could rewrite a Team's deployment-global slug through the public org PATCH route, allowing one tenant to claim or probe identifiers belonging to the shared namespace even though users never choose a slug in the product.
- Approach: Remove `name` from the update DTO and organization update set, keep explicit slug selection only on provisioning-time creation, and let stale request bodies continue to parse with `name` stripped instead of changing the slug.
- Impacted modules: Shared organization DTO, Server organization service, and focused service/multi-org regression tests.

**Review findings**
✅ 1. The DTO and service now agree on a creation-only slug lifecycle; there is no remaining typed or runtime organization-update caller that can assign `organizations.name`.
✅ 2. The route-level regression sends a stale `name` field and verifies the persisted slug is unchanged, while the service test verifies the generated update set cannot carry `name`.
✅ 3. The linked Context Tree draft, `first-tree-context#868`, narrows the current administrative exception to provisioning-time creation and is correctly staged behind this source change.
⚠️ 4. The pre-existing ability for Team admins to change quota and feature fields through the same route remains a separate authorization decision; it is not introduced or widened by this PR.

**Action taken**
- Approved head `b7281e55c67068ad50223f09fb3692e7c91a51e3`.

Per reviewer policy, I did not run tests locally.

yuezengwu (approved):
Reviewed head `b7281e55c`.

The update schema removes `name`, the sole route parses through that schema, and the service no longer assigns the slug or retains unreachable unique-conflict handling. The tests cover both the built update set and end-to-end stripping of a stale `name` field while `displayName` still changes.

Creation/provisioning keeps its explicit slug and conflict semantics. Paired Context Tree PR #868 accurately narrows the existing administrative exception to creation only. No blockers found.

PR #2120, opened 2026-07-31
fix(client): hash session briefing fingerprint file keys to contain runtime artifacts

yuezengwu:
## Summary

- Fixes a path-containment defect in the shared session-briefing fingerprint sink (`packages/client/src/runtime/session-briefing-fingerprint.ts`), used by the Claude Code, Codex, Cursor, and OpenCode handlers.
- The sink previously used the provider-owned opaque `sessionId` verbatim as the final file name (`${sessionId}.json`) and the atomic temp file name. The id has no enforced character guarantee: OpenCode's public schema only requires a `ses` prefix and permits caller-supplied ids, and First Tree's own parser / persisted-session recovery accept any non-empty string. `path.join` resolves `..` segments, so an id like `ses/../../../.first-tree/workspace` would normalize the final path to `<agent home>/.first-tree/workspace.json`, and `renameSync` would overwrite it — letting a runtime artifact escape `.first-tree-workspace/session-briefings/` and clobber workspace files.
- Fix: derive a `sha256(sessionId)` hex stem and use it for the read path, the final write path, and the atomic temp path. Hashing (not a character allowlist or `basename()`) keeps the key a stable 1:1 mapping with no collisions between distinct ids. No raw-id legacy fallback: a missing pre-fix fingerprint only triggers one briefing re-read notice on next resume, then a new safely-named file is written; provider sessions themselves are unaffected.
- Public function signatures and the best-effort (never-throws) read/write behavior are unchanged.

## Validation

- [ ] `pnpm check` (scoped: `biome check` on the 3 changed files — clean)
- [ ] `pnpm typecheck` (scoped: `tsc --noEmit` in `packages/client` — clean)
- [ ] `pnpm test` (scoped: `vitest run src/__tests__/session-briefing-fingerprint.test.ts src/__tests__/opencode-handler.test.ts` in `packages/client` — 48/48 passed)

Scoped checks only, per task instructions (no full QA / runtime acceptance). New deterministic regression coverage in `session-briefing-fingerprint.test.ts`:

- hostile opaque ids containing `/../`, backslashes, and absolute-looking segments still round-trip and stay isolated per id (no sanitization collisions);
- every written artifact matches `^[0-9a-f]{64}\.json$` inside `SESSION_BRIEFINGS_DIR_REL`, nothing is written directly under the runtime dir, and a pre-seeded `.first-tree/workspace.json` is untouched;
- the OpenCode handler test's raw `ses_new.json` filename assertion now goes through `readSessionBriefingFingerprint` instead of depending on the storage key format.

## Change Surface

None of the above — client-internal runtime storage only.

## Notes

- docs or tests updated to match: tests updated (see Validation); no user-visible behavior or docs impact
- follow-up work: the OpenCode parser could additionally reject over-long / control-character ids as protocol robustness, but that is defense-in-depth and not required for containment; documentation sync for the OpenCode runtime (cli-reference, workspace-state) is handled in a separate PR

### DIFF
--- a/packages/client/src/__tests__/opencode-handler.test.ts
+++ b/packages/client/src/__tests__/opencode-handler.test.ts
@@ -1295,7 +1295,7 @@ describe("OpenCode V1 handler", () => {
     expect(inputs).toHaveLength(2);
     expect(inputs[0]).toContain("<first-tree-current-chat-context");
     expect(inputs[1]).toContain("<first-tree-current-chat-context");
-    expect(existsSync(join(root, ".first-tree-workspace", "session-briefings", "ses_new.json"))).toBe(true);
+    expect(readSessionBriefingFingerprint(root, "ses_new")).not.toBeNull();
     await manager.shutdown();
   });
 
--- a/packages/client/src/__tests__/session-briefing-fingerprint.test.ts
+++ b/packages/client/src/__tests__/session-briefing-fingerprint.test.ts
@@ -1,4 +1,5 @@
-import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { createHash } from "node:crypto";
+import { mkdirSync, mkdtempSync, readdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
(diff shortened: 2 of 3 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The shared fingerprint sink now derives every filesystem path from a stable SHA-256 key instead of the provider-owned opaque session id, closing the containment defect without changing the public API or resume semantics.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: A malformed provider session id could escape the session-briefing runtime directory and overwrite workspace files when the runtime persisted its briefing fingerprint.
- Approach: Hash the opaque id before constructing the read, final-write, and atomic-temp paths. Existing raw-id fingerprints are intentionally not read; the next resume may emit one re-read notice and then establishes the safe baseline without affecting the provider session.
- Impacted modules: `packages/client/src/runtime/session-briefing-fingerprint.ts` and focused client tests

**Review findings**
✅ 1. Read, final-write, and temporary-file paths all use the same safe derived key; no production caller bypasses the shared storage abstraction.
✅ 2. The no-fallback migration preserves containment and degrades through the existing one-time briefing re-read behavior.
✅ 3. Regression coverage checks hostile path-shaped ids, per-id isolation, filename shape, and preservation of a sentinel workspace file.

**Action taken**

PR #2118, opened 2026-07-31
feat(cli): add agent-scoped Context Tree IO feed

Gandy2025:
## Summary

A Context Tree **value audit** needs one fact: which tree nodes did an agent actually open, and when. Today the only way to recover that is to mine local runtime transcripts and statically parse shell commands — expensive, lossy, and limited to the runtimes whose transcripts are complete.

`context_tree_io_events` already records exactly that, durably, at tool-execution time, across every runtime. It is deliberately decoupled from `session_events` (which are dropped on session terminate / runtime switch) so the IO facts survive. It simply had no read path a runtime agent could reach.

This adds that read path.

## What changed

- **Server** — `GET /api/v1/agent/context-tree/io` (Class D, `agentSelector`, per `docs/development/http-path-conventions.md`).
- **Service** — `listAgentContextTreeIoEvents` with keyset pagination over `(created_at DESC, id DESC)`, so a row inserted mid-scan cannot shift a page.
- **Shared** — `agentContextTreeIoQuerySchema` / `agentContextTreeIoEventSchema` / response schema.
- **SDK** — `listAgentContextTreeIo`.
- **CLI** — `first-tree tree io` with `--chat`, `--action`, `--since`, `--until`, `--limit`, `--cursor`, and a bounded `--all`.

`treeHeadCommit` is surfaced (only when it is a valid 40-hex sha) so a consumer can recover the node content that was observed at read time.

## Scope decision

**Self-scoped on purpose:** an agent reads its own IO, never another agent's. Org-wide aggregation stays on the existing member-authenticated `/orgs/:orgId/context-tree/snapshot` endpoint. Filters are query parameters, not an authorization boundary — the boundary is the authenticated agent identity.

## Validation

- [x] `pnpm typecheck` — 11/11 packages
- [x] `pnpm exec biome check` — clean on all touched files
- [x] `pnpm --filter @first-tree/server exec vitest run src/__tests__/agent-context-tree-io.test.ts src/__tests__/context-tree-io.test.ts` — 42 passed (7 new + 35 existing regression)

New tests cover: cross-agent isolation, chat/action/time filtering, pagination with no drops or repeats, `treeHeadCommit` validation, malformed-cursor rejection (fails closed rather than silently returning page one), the HTTP route, inverted-window rejection, and unauthenticated rejection.

## QA risk

Read-only addition on an existing table behind the established Class D auth path. No migration, no write path, no change to how events are recorded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2121, opened 2026-07-31
fix(cli): terminate tree io human rows

Gandy2025:
Follow-up to #2118, found while verifying the shipped command against live staging data.

## Problem

`print.line` writes raw bytes to stderr and adds no newline, so `tree io`'s human output ran every event into the next:

```
...command-surface/NODE.md  (claude_write_tool)2026-07-31T08:05:55.095Z  write  file  ...
```

## Fix

Terminate each row plus the empty-range and continuation notices.

JSON mode was never affected — it goes through `print.result` — which is why this got past review: the machine-readable path is the one the tests and the reviewers exercised, and it is the path the value-audit skill actually consumes.

## Verification

- New `tree-io-output.test.ts` asserts every emitted row ends with a newline, and that JSON mode emits the envelope rather than rows. **Confirmed it fails against the previous behavior** before the fix was applied.
- `pnpm --filter first-tree-dev exec tsc --noEmit` clean; biome clean.
- Verified against live staging: `first-tree-staging tree io` now prints one event per line.

### DIFF
--- a/apps/cli/src/__tests__/tree-io-output.test.ts
+++ b/apps/cli/src/__tests__/tree-io-output.test.ts
@@ -1,0 +1,94 @@
+import { Command } from "commander";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const listAgentContextTreeIo = vi.hoisted(() => vi.fn());
+const printResult = vi.hoisted(() => vi.fn());
+const printLine = vi.hoisted(() => vi.fn());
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
The fix matches the Print-layer contract: every human `print.line` payload now terminates itself, including empty-range, continuation, and truncation notices, while JSON mode remains a single `print.result` path. The regression tests cover row termination, notices, and suppression of human rows in JSON mode. No blockers found.

PR #2126, opened 2026-07-31
feat: add agent template adoption

yuezengwu:
## Summary

- add a strict CAS replace-set API for an agent's canonical 0–3 Template IDs and expose the set through the existing resources read model
- import official Prompt, MCP, and complete Skill bundles once as Team-owned `available` Resources, then reuse the Team's current provenance copies for later agents
- create explicit Template-origin bindings, remove only those automatic bindings, and preserve provenance through unchanged generic resource round-trips
- make Template-backed Agent creation atomic at config version 1, including attachment copy, Resource import, bindings, and rollback
- serialize first imports and use stable quota/source lock ordering without adding a table, migration, synchronization loop, repair flow, or diagnostic state machine

This completes the deliberately small Task 3 adoption boundary. Runtime projection, Client materialization, Template UI, seed content, and automatic synchronization remain outside this change.

## Validation

- [x] `pnpm check` — passed; 16 pre-existing warnings, no errors
- [x] `pnpm typecheck` — 11/11 tasks passed
- [x] `pnpm test` — 12/12 tasks passed; Server 257 files / 2931 tests
- [x] focused Shared schema — 49/49
- [x] focused Server adoption — 23/23
- [x] independent QA on exact head `6ecacd643e721157782ff9c603cd52d2524768a1`
- [x] production Server + PostgreSQL 17 probe — 56 authenticated HTTP requests matched expected behavior
- [x] forced no-cache production build — 5/5 tasks passed

Independent QA also exercised real row-lock revalidation, first-import concurrency, byte-identical shared ZIP copying, config notification count, same-name MCP replacement, and post-copy/post-import rollback. No product defect was found.

## Change Surface

## Notes

- docs or tests updated to match: Shared/server contract and integration tests added; existing Web resource fixtures updated for the new required read-model field
- database changes: none; no schema or Drizzle migration changes
- Context Tree: existing one-time import, Team-resource authority, provenance-removal, and no-auto-sync decisions already cover this source change; no Tree diff is needed
- follow-up work: runtime/Client/Web projection and Template seed work remain separate later tasks

### DIFF
--- a/packages/server/src/__tests__/agent-template-adoption.test.ts
+++ b/packages/server/src/__tests__/agent-template-adoption.test.ts
@@ -988,6 +988,134 @@ describe("Agent Template adoption", () => {
         await app.db.select().from(agentResourceBindings).where(eq(agentResourceBindings.agentId, agent.uuid)),
       ).toHaveLength(0);
     });
+
+    it("never deadlocks an existing-agent adoption against Template-backed creation", async () => {
+      const app = getApp();
+      const publisher = await createPublisherAdmin(app);
+      const bundleId = await uploadZip(app, publisher, skillZip("dual-entry-skill"));
+      const template = await publishTemplate(app, publisher, `de-${crypto.randomUUID().slice(0, 6)}`, [
+        { key: "dual-entry-skill", type: "skill", bundleAttachmentId: bundleId },
+      ]);
+      const createAgentRow = await seedAgentFactory(app);
+      const existingAgent = await createAgentRow({ name: `de-old-${crypto.randomUUID().slice(0, 6)}` });
+      const admin = await createTestAdmin(app, { username: `de-${crypto.randomUUID().slice(0, 8)}` });
+      const [configBefore] = await app.db
+        .select()
+        .from(agentConfigs)
+        .where(eq(agentConfigs.agentId, existingAgent.uuid));
+
+      const url = process.env.DATABASE_URL ?? "";
+      const blocker = postgres(url, { max: 1, ...sslOptions(url) });
+      const probe = postgres(url, { max: 1, ...sslOptions(url) });
+      let patchReply: Awaited<ReturnType<typeof call>> | undefined;
+      let postReply: Awaited<ReturnType<typeof call>> | undefined;
+      let patchPromise: ReturnType<typeof call> | undefined;
+      let postPromise: ReturnType<typeof call> | undefined;
+      let blockerInTransaction = false;
+      const withTimeout = async <T>(promise: Promise<T>, label: string): Promise<T> => {
+        let timer: NodeJS.Timeout | undefined;
+        try {
+          return await Promise.race([
+            promise,
+            new Promise<T>((_, reject) => {
+              timer = setTimeout(() => reject(new Error(`${label} timed out — possible deadlock`)), 15_000);
+            }),
+          ]);
+        } finally {
+          if (timer) clearTimeout(timer);
+        }
+      };
+      const waitForSkillWaiters = async (expected: number, deadline: number, label: string): Promise<void> => {
+        for (;;) {
+          const [row] = await probe`SELECT count(*)::int AS waiters FROM pg_locks
+            WHERE locktype = 'advisory' AND objsubid = 2 AND NOT granted
+              AND classid::bigint = (hashtext('team_skill_name')::bigint & 4294967295)
+              AND objid::bigint = (hashtext(${admin.organizationId})::bigint & 4294967295)`;
+          if (row && row.waiters >= expected) return;
+          if (Date.now() > deadline) throw new Error(`${label} did not reach ${expected} waiter(s) in time`);
+          await new Promise((resolve) => setTimeout(resolve, 50));
+        }
+      };
+      try {
+        await blocker`BEGIN`;
+        await blocker`SELECT pg_advisory_xact_lock(hashtext('team_skill_name'), hashtext(${admin.organizationId}))`;
+        blockerInTransaction = true;
+
+        patchPromise = call(app, admin, "PATCH", `/api/v1/agents/${existingAgent.uuid}/templates`, {
+          expectedVersion: configBefore?.version ?? 1,
+          templateIds: [template.id],
+        });
+        await waitForSkillWaiters(1, Date.now() + 10_000, "existing-agent adoption");
+
+        const newName = `de-new-${crypto.randomUUID().slice(0, 6)}`;
+        postPromise = call(app, admin, "POST", `/api/v1/orgs/${admin.organizationId}/agents`, {
+          name: newName,
+          displayName: "Dual Entry",
+          type: "agent",
+          templateIds: [template.id],
+        });
+        await waitForSkillWaiters(2, Date.now() + 10_000, "Template-backed create");
+
+        // Both entries are parked behind the Team Skill lock, so neither
+        // may hold the Team+Template advisory yet.
+        const [lockRow] = await probe`SELECT count(*)::int AS granted_count FROM pg_locks
+          WHERE locktype = 'advisory' AND objsubid = 2 AND granted
+            AND classid::bigint = (hashtext(${admin.organizationId})::bigint & 4294967295)
+            AND objid::bigint = (hashtext(${template.id})::bigint & 4294967295)`;
+        expect(lockRow?.granted_count).toBe(0);
+
+        await blocker`ROLLBACK`;
+        blockerInTransaction = false;
+
+        patchReply = await withTimeout(patchPromise, "existing-agent adoption");
+        postReply = await withTimeout(postPromise, "Template-backed create");
+      } finally {
+        if (blockerInTransaction) await blocker`ROLLBACK`.catch(() => undefined);
+        await Promise.allSettled([patchPromise, postPromise].filter((promise) => promise !== undefined));
+        await blocker.end({ timeout: 1 }).catch(() => undefined);
+        await probe.end({ timeout: 1 }).catch(() => undefined);
+      }
+
+      expect(patchReply?.statusCode).toBe(200);
+      expect(postReply?.statusCode).toBe(201);
+
+      // One Team copy; both Agents bound to it; config versions exact.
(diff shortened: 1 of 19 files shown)


baixiaohang (changes requested):
- Rationale: Template adoption bypasses one existing Team Skill identity invariant, and its MCP collision check does not remain true after ordinary Team Resource maintenance.

**Risk level**: B-high
- Path baseline: changes include `packages/server/src/services/agent.ts` and the new agent-template adoption service.
- Semantic lift: none; the path baseline already covers the agent configuration and Resource transaction boundary.

**PR summary**
- Problem: a user creating or managing an agent needs to adopt an official Template without separately recreating its instructions, Skills, and MCP setup, while later Team edits remain authoritative.
- Approach: atomically stores the canonical Template set, imports first-use components as Team `available` Resources, reuses provenance copies for later agents, creates removable origin bindings, and shares the existing agent-config CAS version.
- Impacted modules: Shared Agent/Template/Resource schemas; Server agent creation, attachment copy, Resource reads/writes, and Template adoption; Web Resource fixtures.

**Review findings**

❌ 1. Template import bypasses the Team Skill name invariant. [`packages/server/src/services/agent-template-adoption.ts:133`]

Ordinary Team Skill creation serializes on the Team Skill-name lock and rejects any active/stale case-insensitive duplicate (`resources.ts:1092-1095`). This direct insert does neither. A Team that already has the same Skill name—or one replace-set containing two Templates whose bundles declare the same name—therefore gets multiple active Team Skill Resources even though the normal catalog API rejects that state. Please route imports through the same shared invariant (including the existing advisory lock, acquired before attachment/source locks to preserve lock order) and cover both an existing-Team collision and an intra-adoption collision. [R1 / R4]

❌ 2. MCP uniqueness is a one-time preflight even though the imported Resources remain Team-editable. [`packages/server/src/services/agent-template-adoption.ts:342`]

The new guard protects only the adoption transaction. After it commits, an admin can update the imported MCP payload or add/recommend another Team MCP with the same case-insensitive runtime name; the ordinary Resource mutations permit that. `resolveEffectiveResources` then leaves both rows enabled and `runtimeMcp` projects both, while provider adapters key them by name and silently collapse one. Please enforce the identity at the durable effective-Resource/projection boundary (or across every relevant Resource mutation with concurrency-safe validation), so later Team maintenance cannot invalidate the adoption invariant. [R1 / R5]

**Action taken**
- Submitted request changes on exact head `6ecacd643e721157782ff9c603cd52d2524768a1`.

yuezengwu:
Addressed both requested changes in follow-up commit `feff04387884da9a68de1a8f0180f83ba42c6d90` (parent `6ecacd643e721157782ff9c603cd52d2524768a1`; no history rewrite).

1. Template Skill first-import now reuses the ordinary Team Skill advisory lock and case-insensitive active/stale name check. The shared name lock is acquired before any attachment quota/source lock. Regression coverage includes an existing-Team collision, an intra-adoption collision, and zero-copy/zero-write rollback.
2. MCP identity is now enforced at the effective projection boundary as well as adoption preflight. Every enabled member of a case-insensitive duplicate-name group becomes `unavailable` with `duplicate_mcp_name`; runtime projection emits no silent winner. Regression coverage includes later ordinary Team MCP creation/recommendation and payload rename, with unique and invalid-payload behavior preserved.

Independent exact-head qualification passed on production Server + PostgreSQL 17: 16/16 behavior cases, 87 authenticated HTTP requests, a real concurrent ordinary Skill-create/adoption race with exactly one success and one 409 and no deadlock, forced uncached typecheck 11/11, build 5/5, and full repository tests 12/12 (Server 257 files / 2,934 tests). The follow-up changes exactly 3 files and has no schema or migration diff.

baixiaohang (changes requested):
- Rationale: the two original invariants are now enforced, but the new shared Team Skill lock is acquired in the reverse order from `agent_configs` across Template adoption and ordinary Team Skill writes.

**Risk level**: B-high
- Path baseline: changes include the Agent service and the Agent Template adoption transaction under `packages/server/**`.
- Semantic lift: none; the path baseline already covers the agent configuration and Resource transaction boundary.

**PR summary**
- Problem: a user creating or managing an agent needs to adopt an official Template without separately rebuilding its Instructions, Skills, and MCP configuration, while later Team edits remain authoritative.
- Approach: atomically imports/reuses Team Resources and origin bindings under the shared agent-config CAS; this follow-up also shares the Team Skill-name invariant and makes duplicate effective MCP names unavailable at projection time.
- Impacted modules: Server Template adoption, Team Resource mutation/projection, attachment copy, Agent creation, and their Shared/Web read contracts.

**Review findings**

❌ 1. The new Team Skill-name lock introduces an `agent_configs`/advisory-lock inversion. [`packages/server/src/services/agent-template-adoption.ts:317`]

An existing-agent adoption first locks that Agent's config row at lines 447-452, then reaches `lockTeamSkillNames` here. Ordinary Team Skill create/update does the opposite: it holds `team_skill_name` at `resources.ts:1132` / `1210`, then calls `bumpAgentConfigVersions` at `resources.ts:1153` / `1229`. A concurrent recommended Skill create (or an update of a Skill already affecting this Agent) can therefore hold the advisory lock while waiting for the adoption's config row, as the adoption waits for the advisory lock. PostgreSQL must abort one transaction as a deadlock, surfacing a 500 instead of either atomic operation. Please establish one global order—for example, acquire the Team Skill lock before the config-row lock in the existing-agent adoption path—and add a concurrent adoption-vs-Team-Skill-write regression. [R4]

✅ The previous Team Skill identity and durable MCP identity findings are otherwise addressed: first imports now share the Team Skill invariant, and duplicate effective MCP names fail closed as unavailable before runtime projection.

**Action taken**
- Submitted request changes on exact head `feff04387884da9a68de1a8f0180f83ba42c6d90`.

yuezengwu:
Addressed the lock-order finding in follow-up commit `8274fb065b6117e14f1c05b05ff3016ade6c2735` (parent `feff04387884da9a68de1a8f0180f83ba42c6d90`; no history rewrite).

Existing-Agent adoption now acquires the shared Team `team_skill_name` advisory lock after locked-Agent lifecycle validation but before `agent_configs FOR UPDATE` whenever the target Template set is non-empty. This establishes the compatible orders:

- adoption: Agent row → `team_skill_name` → `agent_configs`
- ordinary Team Skill write: `team_skill_name` → attachment/resource → `agent_configs`

Empty-target removal does not acquire the advisory lock. The first-import check still reuses the same transaction lock before attachment quota/source locks; ordinary Resource writes, config CAS, Agent-row locking, and the adoption planner are unchanged.

The regression uses a real attachment-row barrier and exact `pg_locks` waiter observation, not timing sleeps. Removing the production fix makes PostgreSQL report `deadlock detected` and adoption return 500; restoring it yields ordinary Skill create 201 and adoption stale-CAS 409 with no Template copy/import/binding/config residue.

Independent exact-head production qualification on PostgreSQL 17 passed: while the waiter was visible, `agent_configs FOR UPDATE NOWAIT` succeeded and Template state stayed unchanged; 3/3 lock-order cases plus 16/16 broad cases passed over 100 authenticated HTTP requests. Forced uncached typecheck 11/11, build 5/5, and full repository tests 12/12 passed (Server 257 files / 2,935 tests). The follow-up is exactly 2 files, +144/-0, with no schema or migration diff.

baixiaohang (dismissed):
**Recommendation**: Request changes

The previous Team Skill writer deadlock is fixed, but the two Template-adoption entry points now acquire the same locks in opposite orders.

**Risk level**: B-high
- Path baseline: `packages/server/**`, including Agent state and resource projection logic.

**Review findings**

1. **Blocking — existing-Agent adoption and Template-backed Agent creation can deadlock each other.** `adoptAgentTemplates` now acquires the Team Skill advisory lock before entering `adoptTemplatesInTransaction`, which later acquires the per-Team/Template advisory lock. However, `initializeAdoptedTemplates` enters the shared function directly: it acquires the Template lock first (line 287) and only then the Team Skill lock (line 317). With a first import of the same Skill-bearing Template, an existing-Agent adoption can therefore hold `team_skill_name` while waiting for the Template lock, while a concurrent Agent creation holds the Template lock while waiting for `team_skill_name`; PostgreSQL must abort one transaction as a deadlock. Please establish one lock order across both entry points (ideally at a shared boundary) and cover concurrent existing-Agent adoption versus Template-backed Agent creation. The new regression only covers an ordinary Team Skill write versus existing-Agent adoption. [R4 — `packages/server/src/services/agent-template-adoption.ts:287,317,400,453`]

The Context Tree's one-time import and ordinary Team-resource governance boundary remains consistent with the overall design; this finding is limited to atomic/concurrent execution of that design.

yuezengwu:
Third review finding is fixed in exact head `c824b2be5735efe07942f0e26f3138976644dcd8` by the standalone commit `c824b2be5` (parent `8274fb065`; no amend/force-push).

The shared adoption boundary now acquires `team_skill_name` before any Team+Template or attachment lock for both existing-Agent adoption and Template-backed Agent creation. The existing-Agent path keeps its pre-config acquisition, yielding one order across the involved paths without a new lock abstraction or planner.

The deterministic PostgreSQL regression holds the Team Skill advisory externally, observes PATCH and POST as waiters 1→2, and proves the Team+Template advisory has zero granted holders before release. After release both requests complete (PATCH 200 / POST 201), with exactly one Team Skill Resource and byte-identical ZIP copy, shared bindings, and exact config versions. Reverting only the production fix makes the same test observe the old Template-lock holder and then a real deadlock.

Formal isolated QA on this exact head passed:

- Production Server + PostgreSQL 17 dual-entry and prior ordinary-Skill lock probes
- Shared 49/49; Server focused 167/167 (adoption 28/28)
- check; forced uncached typecheck 11/11; production build 5/5
- forced uncached CI single-fork full suite 12/12; Server 257 files / 2936 tests
- candidate defects: 0; worktree exact/clean; no schema/migration change

The first full QA run's sole failure was an unchanged millisecond-tie rendezvous-pruning assertion; QA reproduced it on direct parent 3/15 under the same environment, and the subsequent exact-candidate run was fully green.

PR #2127, opened 2026-07-31
feat: add Grok Build runtime provider

yuezengwu:
# Add Grok Build runtime provider

Adds `grok` ([Grok Build](https://github.com/xai-org/grok-build)) as a new runtime provider, driving the official `grok` CLI over **ACP v1** (stdio / NDJSON) via the official `@agentclientprotocol/sdk@1.3.0`.

## Adopted V1 design

- **Short-lived process per provider turn.** Each turn spawns `grok --no-auto-update agent --no-leader --always-approve [--model <id>] [--effort <low|medium|high>] stdio`, runs `initialize → session/new|session/load → session/prompt → full notification/response barrier → stdin EOF → drained exit 0`, then exits. One persisted Grok session ID per `(agent, chat)`; same-chat turns serialize through the existing handler queue; different chats run in parallel — including on a shared workspace cwd, since Grok keys session storage by session ID under the encoded cwd. No shared cross-chat process, no mid-turn interject in V1 (active-turn `inject` queues for the next turn).
- **Empty client fs/terminal capabilities by design.** `initialize` advertises `clientCapabilities: {}` so Grok keeps using its own local filesystem/terminal backends. Declaring those capabilities would force First Tree to implement reverse fs/terminal proxy methods — deliberately out of scope. A defensive `session/request_permission` handler fails closed (it never fires under `--always-approve` + `yoloMode`).
- **Auth/token boundary.** First Tree never reads, copies, or relays Grok tokens. Ordinary turns do not send ACP `authenticate` — Grok auto-selects its initialize-advertised `defaultAuthMethodId`; a credential failure classifies as `credential`, persists the existing durable runtime notice before ACK, and offers the explicit in-product browser-login action, which runs the official `grok login` on the daemon host. Headless device auth / API keys remain provider-owned host configuration.
- **Session integrity.** Only stream-confirmed Grok session IDs are persisted (synthetic `grok-pending-*` placeholder upgrades via `replaceSessionId`). `session/load` must match the persisted ID and **never silently falls back to a fresh session** (replay safety). The replay contract is exact, not time-based: `session/load` carries `_meta.noReplay: true`, replay-marked notifications (`params._meta.isReplay === true`) are dropped on every route regardless of arrival time, and each attempt mints a `session/prompt` `_meta.promptId` — updates, `turn_completed`, and the prompt response must correlate to it (mismatch or missing fails closed). A legacy quiet-window fence remains only as a fallback for builds without the markers.
- **Model/effort are re-applied after EVERY session open.** Because `session/load` restores the session's persisted model/effort, argv-only selection would silently keep stale values on resumed sessions. After every `session/new` or `session/load` and before the prompt, the handler sends `session/set_model` (`{ sessionId, modelId, _meta: { reasoningEffort } }`) and requires the response to carry `_meta.model.Ok === effectiveModelId`. An explicit model is always re-applied; an empty effort omits the `_meta` key entirely (override removed); an empty model resets to the initialize-advertised default model. A `set_model` rejection classifies as `configuration` and no prompt is sent — no silent fallback.
- **Explicit effort is also verified effective before prompting.** Grok silently ignores an effort override the selected model does not support while still returning the successful model response, so a non-empty `reasoningEffort` additionally requires THIS set_model's effective echo: a non-replay `_x.ai/session_notification` `model_changed` on the exact session, received after the waiter's receipt-time ingress baseline (stale/foreign/wrong-channel/replay-marked echos never satisfy), with `model_id === effectiveModelId` and `reasoning_effort === requestedEffort`. Missing/omitted/wrong/timed-out confirmations, and a replay drain that fails to go quiet before an explicit effort, are deterministic configuration terminals — 1 spawn, no retry, no prompt, no success ACK (the SessionManager-owned durable notice still precedes the consumed ACK). The once-settled echo waiter is cancelled on response failure, confirmation, and abort; diagnostics report only post-arm candidates. Empty effort keeps remove-override semantics (the provider default effort is accepted without an echo requirement). Verified against live Grok 0.2.117 for the wire shape/order/provider default; the unsupported-model ignored-override branch is covered by upstream-source-derived fake regressions (not live-tested — this account's catalog only lists effort-capable models).
- **Settlement boundaries.** Success and failure paths share one settle barrier: EOF → bounded child close + `connection.closed` → stable notification-handler drain → then settlement state is read (post-response write/usage notifications cannot be truncated). A completed prompt requires a clean exit 0; non-zero exits and EOF timeouts route into `ProviderAttempt` per replay-safety/custody (terminate timers are cleared on real close so a reused PGID can never be killed; non-piped stdio branches reclaim with a bounded close wait). ACP terminal stop reasons `max_tokens` / `max_turn_requests` / `refusal` are completions (assistant text and usage preserved); only `cancelled` is abnormal and it too flows through `ProviderAttempt`/replay-safety rather than unconditional redelivery. Replay-safety state accumulates across attempts. JSON-RPC error `data` is preserved length-capped in failure classification.
- **Events.** Standard ACP `agent_message_chunk` (accumulated, emitted once as chunked `assistant_text` at successful turn end), `agent_thought_chunk` (presence-only `thinking`), `tool_call` / `tool_call_update` mapped with `_meta["x.ai/tool"]` metadata (`read_file` / `write` / `search_replace` / `run_terminal_cmd`) into tool events with Context Tree file refs; usage parsed from `_x.ai/session_notification` `turn_completed` into one `token_usage` event per turn; replay safety advances when assistant output or a non-proven-read-only tool begins.
- **MCP / models / effort.** Managed MCP servers map into ACP `session/new|load` params, filtered by the transports the agent advertises. `model` is a free-form exact ID (empty = Grok local default); `reasoningEffort` maps to `--effort` with `""` as the inherit sentinel. Model catalog discovery runs a bounded initialize-only ACP handshake and parses `_meta.modelState`, degrading to the free-form input when unavailable.
- **Managed skills** project to `.grok/skills`.

## Supported range and platform gate

- Grok Build **>= 0.2.117 < 0.3.0** (validated against 0.2.117); the bounded `grok --version` compatibility gate runs at first real provider use against the exact binary the handler will spawn.
- **macOS / Linux only in V1.** The capability probe is install/platform-only (never launches the CLI, never inspects credentials, never touches the network) and Windows fails closed at EVERY entry point — probe, binary resolver, model discovery, runtime auth, and the Web install box (which renders an unsupported status instead of an install command on win32).
- **Deterministic unsupported classification.** win32 bring-up and resolved-but-out-of-range/invalid binaries classify as `capability`/`configuration` (`grok_platform_unsupported` / `grok_binary_unsupported`) with an immediate stop — never the unknown-retry path; `grok_binary_missing` is reserved for an unresolved binary.
- 429s surface after Grok's internal retry budget and classify as `provider_rate_limited` capacity with bounded visible retry; free-tier rate limits are not hardcoded into the provider.

## Prior live validation (research/QA pass, Grok 0.2.117)

Parallel `session/new`/`session/load` on one cwd, isolated session directories, clean EOF exit 0, single-side failure isolation, real `grok-4.5` prompts, cross-process local tool read/write, and two concurrent chat-like tool turns all passed. The only live transient observed was a recoverable HTTP 429.

## Verification

- `pnpm check` — clean (2122 files).
- `pnpm typecheck` — clean across all packages.
- `pnpm test` — 12/12 turbo tasks succeeded on the final head `d3a70a7c3`. Focused grok suites on this head: `grok-handler-engine` 51 tests; the reviewer's independent run over the 5 adversarial/settlement suites (engine/events/ACP session/provider policy/SessionManager) is 145/145; Web grok suites 13/13.
- One environment-sensitive CLI case (`core-attention-update-extra.test.ts` `detectInstallMode`) was observed failing on this machine on a clean baseline as well (returns `portable` instead of `npx` depending on install mode); it is unrelated to this change and passed in the final full-suite run.
- `git diff --check` clean; lockfile change is limited to `@agentclientprotocol/sdk@1.3.0`.

## QA

Formal live-provider QA is **pending** (not yet executed): this adds a new provider across runtime/auth surfaces. `packages/qa/cases/runtime/grok-provider.md` covers the live boundary (install-only probe, auth recovery with durable notice + `grok login`, ACP spawn posture, same-cwd parallel chats, session new/load resume with set_model re-apply + replay-marker filtering, terminal stop reasons, MCP/skills projection, model + effort round-trips including the empty-config reset, Context Tree I/O evidence, 429 capacity retry, discovery version gate, client-switch drain).

### DIFF
--- a/apps/cli/src/__tests__/capability-refresh.test.ts
+++ b/apps/cli/src/__tests__/capability-refresh.test.ts
@@ -36,6 +36,7 @@ const allOk = (): ClientCapabilities => ({
   "claude-code-tui": ok(),
   codex: ok(),
   cursor: ok(),
+  grok: ok(),
   "kimi-code": ok(),
   opencode: ok(),
 });
@@ -45,6 +46,7 @@ const codexMissing = (): ClientCapabilities => ({
   "claude-code-tui": ok(),
   codex: missing(),
   cursor: ok(),
+  grok: ok(),
   "kimi-code": ok(),
   opencode: ok(),
 });
@@ -74,6 +76,7 @@ const codexPendingSnapshot = (): ClientCapabilities => ({
   "claude-code-tui": ok(),
   codex: codexPending(),
   cursor: ok(),
+  grok: ok(),
   "kimi-code": ok(),
   opencode: ok(),
 });
@@ -84,6 +87,7 @@ const codexUnauthSnapshot = (): ClientCapabilities => ({
   "claude-code-tui": ok(),
   codex: codexUnauth(),
   cursor: ok(),
+  grok: ok(),
   "kimi-code": ok(),
   opencode: ok(),
 });
--- a/apps/cli/src/__tests__/client-switch.test.ts
+++ b/apps/cli/src/__tests__/client-switch.test.ts
@@ -276,6 +276,14 @@ describe("client switch drain markers", () => {
     expect(isSwitchDrainEnvRequired("/home/op/.local/bin/agent -p --output-format stream-json")).toBe(true);
     expect(isSwitchDrainEnvRequired("cursor-agent login")).toBe(true);
   });
+
+  it("recognizes Grok Build CLI processes (grok binary)", () => {
+    // The drain must fail closed: a live grok turn spawned by the runtime must
+    // require the env envelope check. Grok's binary is `grok` — there is no
+    // generic-name special case like Cursor's `agent` basename.
+    expect(isSwitchDrainEnvRequired("/home/op/.local/bin/grok -p --output-format stream-json")).toBe(true);
+    expect(isSwitchDrainEnvRequired("grok login")).toBe(true);
+  });
 });
 
 describe("client runtime markers", () => {
--- a/apps/cli/src/__tests__/runtime-auth-login.test.ts
+++ b/apps/cli/src/__tests__/runtime-auth-login.test.ts
@@ -348,3 +348,73 @@ describe("runRuntimeAuthLogin — cursor (external-only binary)", () => {
     expect(h.calls.at(-1)?.entry.lastAuthError).toMatchObject({ reason: "exit-nonzero", message: "login failed" });
   });
 });
+
+describe("runRuntimeAuthLogin — grok (external-only binary)", () => {
+  function grokHarness(opts: {
+    resolveOk?: boolean;
+    outcome?: LoginOutcome;
+    fireAuthUrl?: string;
+    probeResult?: CapabilityEntry;
+  }) {
+    const calls: Recorded2[] = [];
+    const logs: string[] = [];
+    const loginCalls: string[] = [];
+    const deps = {
+      currentEntry: (): CapabilityEntry | undefined => undefined,
+      setProviderEntry: async (provider: string, entry: CapabilityEntry): Promise<void> => {
+        calls.push({ provider, entry });
+      },
+      log: (_symbol: string, msg: string): void => {
+        logs.push(msg);
+      },
+      now: (): number => NOW,
+      resolveGrokBinary: () =>
+        opts.resolveOk === false
+          ? ({ ok: false, error: "Grok Build CLI is missing on this machine.", transient: false } as const)
+          : ({ ok: true, binary: "/home/op/.local/bin/grok", version: "2026.07.09" } as const),
+      runGrokBrowser: async (o: { binary: string; onAuthUrl?: (url: string) => void }): Promise<LoginOutcome> => {
+        loginCalls.push(o.binary);
+        if (opts.fireAuthUrl) o.onAuthUrl?.(opts.fireAuthUrl);
+        await new Promise((r) => setTimeout(r, 0));
+        return opts.outcome ?? ({ ok: true } as const);
+      },
+      probeGrok: async (): Promise<CapabilityEntry> => opts.probeResult ?? installedEntry(),
+    };
+    return { calls, logs, loginCalls, deps };
+  }
+  type Recorded2 = { provider: string; entry: CapabilityEntry };
+
+  it("drives <resolved-binary> login and publishes pending → re-probed ok", async () => {
+    const h = grokHarness({ outcome: { ok: true }, probeResult: okEntry() });
+    await runRuntimeAuthLogin({ provider: "grok", ref: "rg1" }, h.deps);
+
+    expect(h.loginCalls).toEqual(["/home/op/.local/bin/grok"]);
+    expect(h.calls[0]?.provider).toBe("grok");
+    expect(h.calls[0]?.entry.pendingAuth).toMatchObject({ method: "browser" });
+    expect(h.calls.at(-1)?.entry.state).toBe("ok");
+    expect(h.calls.at(-1)?.entry.lastAuthError).toBeUndefined();
+  });
+
+  it("on unresolved/unverified binary, reflects a spawn-error lastAuthError and never logs in", async () => {
+    const h = grokHarness({
+      resolveOk: false,
+      probeResult: { ...installedEntry(), state: "missing", available: false },
+    });
+    await runRuntimeAuthLogin({ provider: "grok", ref: "rg2" }, h.deps);
+
+    expect(h.loginCalls).toEqual([]);
+    expect(h.calls).toHaveLength(1);
+    expect(h.calls[0]?.entry.state).toBe("missing");
+    expect(h.calls[0]?.entry.lastAuthError).toMatchObject({ reason: "spawn-error" });
+    expect(h.logs.some((l) => l.includes("grok binary unavailable"))).toBe(true);
+  });
+
+  it("stamps lastAuthError when the grok login fails", async () => {
+    const h = grokHarness({
+      outcome: { ok: false, reason: "exit-nonzero", error: "login failed" },
+      probeResult: installedEntry(),
+    });
+    await runRuntimeAuthLogin({ provider: "grok", ref: "rg3" }, h.deps);
+    expect(h.calls.at(-1)?.entry.lastAuthError).toMatchObject({ reason: "exit-nonzero", message: "login failed" });
+  });
+});
--- a/apps/cli/src/commands/agent/config/set-reasoning-effort.ts
+++ b/apps/cli/src/commands/agent/config/set-reasoning-effort.ts
@@ -7,7 +7,7 @@ export function registerAgentConfigSetReasoningEffortCommand(config: Command): v
   config
     .command("set-reasoning-effort <agent> <level>")
     .description(
-      'Set reasoning effort. claude-code: "" (inherit local) | low | medium | high | max. codex: low | medium | high | xhigh | max | ultra (model-dependent). Cursor, Kimi Code, and OpenCode have no separate effort field.',
+      'Set reasoning effort. claude-code: "" (inherit local) | low | medium | high | max. codex: low | medium | high | xhigh | max | ultra (model-dependent). Grok Build: "" (inherit local) | low | medium | high. Cursor, Kimi Code, and OpenCode have no separate effort field.',
     )
     .action(async (agentName: string, level: string) => {
       const serverUrl = resolveServerUrl(process.env.FIRST_TREE_SERVER_URL);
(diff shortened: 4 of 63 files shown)


baixiaohang (changes requested):
- Rationale: The cross-surface provider wiring is coherent, but four ACP lifecycle gaps can silently run the wrong configuration, lose valid terminal output, replay historical events, or leak discovery processes.

**Risk level**: A
- Semantic lift: none (already at the highest level; the diff also changes core provider-session behavior)

**PR summary**
- Problem: Operators cannot currently bind an agent to Grok Build, recover a missing Grok login from the blocked chat, or use Grok-native sessions, models, effort, MCP, skills, and Context Tree activity through First Tree.
- Approach: Adds an external `grok` provider driven over ACP v1 with one supervised CLI process per turn, persisted per-chat session IDs, in-product browser login, capability/model discovery, managed MCP/Skill projection, normalized events, and the corresponding Server/Web/CLI schemas and surfaces.
- Impacted modules: `packages/client` handlers/runtime, shared runtime schemas, Server Context Tree I/O, Web agent/computer configuration, CLI auth/provider commands, docs, and QA assets.

**Review findings**
❌ 1. A resumed session can silently ignore a newly configured model or reasoning effort. The handler only applies the current values as process argv, then calls `session/load`; Grok Build's load path restores the session's persisted model and reasoning effort, and this integration neither sends an ACP model/config update after load nor verifies the effective values. The Web promises that an exact model reaches the next turn with no silent fallback, but a same-chat next turn can continue on the old persisted selection while the argv-only test still passes. Apply/verify the requested values after `session/load` (and cover a config change on an existing session), or scope the product contract explicitly to new sessions. [R5 / `packages/client/src/handlers/grok/index.ts:614`]

❌ 2. Valid ACP terminal outcomes are treated as provider failures and their assistant text is discarded. ACP v1 defines `end_turn`, `max_tokens`, `max_turn_requests`, `refusal`, and `cancelled`; Grok itself can return `refusal`. This branch accepts only `end_turn`, sends every other non-cancelled response through unknown-failure retry/terminal handling, and never forwards the accumulated text. Handle each protocol stop reason as an explicit terminal outcome (with the intended success/error status) rather than classifying protocol-defined completion as a transport/provider error. [R4 / `packages/client/src/handlers/grok/index.ts:681`]

❌ 3. The replay fence is time-based even though the Grok wire provides an exact replay discriminator. Grok stamps replayed `session/load` notifications with `params._meta.isReplay`, and also supports suppressing replay for a client that already owns the transcript. Here every pre-prompt notification is assumed historical, then the fence opens after only 25 ms of quiet or a 2 s cap. A long replay backlog or delayed SDK handler can therefore cross the cap and be emitted as current assistant/tool/usage activity. Drop the explicit replay-marked events (or request `noReplay`) and add prompt/event correlation; keep any quiet window only as a compatibility fallback. [R4 / `packages/client/src/handlers/grok/acp-session.ts:428`]

❌ 4. Model discovery resolves before the spawned CLI is known to have exited. On initialize success, `finish()` clears the only timeout immediately after `stdin.end()`; on initialize failure it resolves without even closing stdin or killing the child. A Grok process that does not exit on EOF can therefore survive every catalog request indefinitely, contrary to this function's own bounded-drain contract and the client-switch containment requirement. Settle only after close with an acceptable exit status, and retain a bounded terminate/kill path for every failure branch. [R4 / `packages/client/src/handlers/grok/acp-session.ts:515`]

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The follow-up fixes two lifecycle blockers and improves dirty-exit/retry settlement, but resumed-session configuration and replay isolation are still heuristic and can silently produce the wrong turn.

**Risk level**: A
- Semantic lift: none (already at the highest level; the diff changes core provider-session behavior)

**PR summary**
- Problem: Operators cannot currently bind an agent to Grok Build, recover a missing Grok login from the blocked chat, or use Grok-native sessions, models, effort, MCP, skills, and Context Tree activity through First Tree.
- Approach: Adds an external `grok` provider driven over ACP v1 with one supervised CLI process per turn, persisted per-chat session IDs, in-product browser login, capability/model discovery, managed MCP/Skill projection, normalized events, and the corresponding Server/Web/CLI schemas and surfaces. This follow-up hardens terminal stop handling, process drain, dirty exits, and cross-attempt side-effect tracking.
- Impacted modules: `packages/client` handlers/runtime, shared runtime schemas, Server Context Tree I/O, Web agent/computer configuration, CLI auth/provider commands, docs, and QA assets.

**Review findings**
❌ 1. Resumed sessions can still silently ignore a newly configured model or reasoning effort. The new process receives the current values only through `--model` / `--effort`, then `session/load` is sent with only `yoloMode`; there is still no post-load model/config request or effective-value check. Grok's load path restores the persisted session model and reasoning effort, so an existing chat can continue on its old selection even though First Tree refreshed the config for this turn. Apply and verify the requested values after load (Grok exposes session model switching with reasoning-effort metadata), or explicitly scope the product contract to new sessions. [R5 / `packages/client/src/handlers/grok/acp-session.ts:403`, `packages/client/src/handlers/grok/index.ts:637`]

❌ 2. Replay isolation is unchanged and still uses elapsed time instead of the wire's exact discriminator. `session/load` omits `noReplay`, `routeNotification` ignores `_meta.isReplay`, and the fence opens after 25 ms of quiet or a 2 s cap. Any delayed/long historical replay that arrives after that point is indistinguishable from the current prompt and can be emitted as current assistant/tool/usage activity. Drop replay-stamped notifications (with the quiet window only as an older-build fallback), or request `noReplay`; adding a prompt id to `session/prompt` and correlating notifications would make the live side exact as well. [R4 / `packages/client/src/handlers/grok/acp-session.ts:274`, `packages/client/src/handlers/grok/acp-session.ts:403`, `packages/client/src/handlers/grok/acp-session.ts:430`]

✅ 3. The follow-up now treats all ACP v1 completion stop reasons explicitly and preserves their accumulated text/usage after a clean exit; `cancelled` remains on the abort/redelivery path.

✅ 4. Model discovery now waits through a bounded EOF→close barrier, requires exit 0, and reclaims the process on initialize failure, timeout, or ignored EOF.

**Action taken**
- Submitted request changes on head `fcab26c12e317065c9a9a110c37b201e1ff9e59e`.

baixiaohang (dismissed):
- Rationale: This head closes the replay-isolation and model-restoration gaps, but a configured reasoning effort can still be silently ignored while the turn is treated as correctly configured.

**Risk level**: A
- Semantic lift: none (already at the highest level; the diff changes core provider-session behavior)

**PR summary**
- Problem: Operators cannot currently bind an agent to Grok Build, recover a missing Grok login from the blocked chat, or use Grok-native sessions, models, effort, MCP, skills, and Context Tree activity through First Tree.
- Approach: Adds an external `grok` provider driven over ACP v1 with one supervised CLI process per turn, persisted per-chat session IDs, in-product browser login, capability/model discovery, managed MCP/Skill projection, normalized events, and the corresponding Server/Web/CLI schemas and surfaces. This follow-up adds exact replay suppression/correlation, post-open model application, and a unified transport-settlement barrier.
- Impacted modules: `packages/client` handlers/runtime, shared runtime schemas, Server Context Tree I/O, Web agent/computer configuration, CLI auth/provider commands, docs, and QA assets.

**Review findings**
❌ 1. The post-open configuration check still verifies only the model, not the requested reasoning effort. `session/set_model` sends `reasoningEffort`, but the success check accepts the turn solely when `_meta.model.Ok` matches. Grok Build's handler explicitly logs and ignores an effort override when the selected model does not support it, then still returns the same successful model-only response ([upstream implementation](https://github.com/xai-org/grok-build/blob/dd04f397b1d02f2272b092555669dfba1f01bc85/crates/codegen/xai-grok-shell/src/agent/handlers/model_switch.rs#L127-L147), [response shape](https://github.com/xai-org/grok-build/blob/dd04f397b1d02f2272b092555669dfba1f01bc85/crates/codegen/xai-grok-shell/src/agent/handlers/model_switch.rs#L248-L254)). Because First Tree keeps the model free-form and always offers low/medium/high, an operator can select an incompatible pair and the next turn silently runs Grok's default effort even though the UI says the override is re-applied. Verify the applied effort (for example from Grok's `ModelChanged` echo), or reject an advertised-incompatible combination before prompting, and cover that path. [R5 / `packages/client/src/handlers/grok/acp-session.ts:636`]

✅ 2. `session/load` now sends `noReplay`, replay-stamped traffic is dropped regardless of timing, and prompt responses are correlated with a client-generated `promptId`; the previous replay-isolation blocker is closed.

✅ 3. Model selection is now re-applied after both `session/new` and `session/load`, and a missing/mismatched model confirmation fails closed before the prompt.

**Action taken**
- Submitted request changes on head `33e2cd12098de5e0c92afe6dbaa65af936555afb`.

PR #2128, opened 2026-07-31
fix(client): preserve Kimi background work through turn completion

yuezengwu:
## Summary

- keep Kimi deliveries in processing custody while top-level background `Agent` tasks are still running, on both new and resumed SDK sessions
- separate main-agent chat/turn lifecycle from subagent streams while retaining every agent's real tool effects with collision-safe child tool IDs
- carry the pinned SDK resume fix through workspace installs, production Docker builds, npm CLI packages, and portable artifacts

## Root cause

Mention delivery and agent wake-up were healthy. The Kimi main turn could finish immediately after launching `Agent(run_in_background)`, so First Tree completed and ACKed the delivery while an in-process child still owned unfinished work. The pinned SDK already had an internal drain-on-stop path for newly created sessions, but First Tree did not enable it and the SDK did not thread the option through resume.

The handler also consumed one shared SDK event stream without checking `agentId`, allowing child assistant/turn events to be treated as parent output or completion.

## Implementation

- apply a narrow pnpm patch to `@botiverse/kimi-code-sdk@0.26.0-botiverse.2` so resumed sessions mirror create-session drain behavior
- request drain-on-stop for both handler create and resume paths
- let only `agentId === "main"` drive parent assistant/thinking/error/turn completion
- retain tool events from all agents; preserve main tool IDs and prefix child IDs with their agent ID
- bundle the patched SDK in npm CLI artifacts and propagate workspace patched dependencies into portable builds
- make the Docker dependency stage copy patch inputs before frozen install
- extend deterministic handler, packaging, portable, and live QA coverage

## Verification

Deterministic gates:

- frozen install, root build, check, and typecheck pass
- Kimi handler: 24/24
- client: 1875/1875, 3 skipped
- affected CLI/portable release tests: 43/43
- production target and base images build and serve `/healthz`
- clean npm consumer runs the CLI and imports the nested bundled `_patch_hash=ab8351...` SDK with both resume behavior sites
- staging portable artifact runs with the same patched SDK and no raw patch sources

Independent real-provider QA reached full `QA READY` and passed the change's create and resume custody paths:

- both turns launched two background Agents and remained delivered/unacked beyond a 5-second idle threshold
- create settled once after 111 seconds; resume settled once after 98 seconds with the same persisted provider session ID
- each path emitted exactly one successful parent `turn_end` followed by ACK, with retry count 0
- child assistant output did not leak; child tool IDs were agent-prefixed and main IDs remained raw
- daemon lifecycle reported processing work instead of idle suspension

Formal QA status is `BLOCKED`, not `PASS`, solely because a third adjacent unsafe-write interruption attempt hit the Kimi billing-cycle limit with HTTP 403 before any tool event. No target defect was found, and the required create/resume behavior plus release artifacts passed with real product-boundary evidence.

## Known release observations

- the real publish path uses `npm publish`; local artifact verification therefore uses `npm pack` (pnpm's isolated linker rejects `bundleDependencies`)
- the clean npm consumer emitted path-normalization warnings while npm traversed the pnpm-linked bundled dependency; the installed SDK tree matched the workspace digest and all required chunks, assets, licenses, and runtime imports were verified

## Context Tree

No Context Tree update is needed. This change enforces the existing Kimi direct-SDK lifecycle, deferred-ACK, replay-safety, and recoverable-delivery constraints; it does not establish a new durable decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,6 +219,15 @@ jobs:
       # all 233 tests passed. Building before vitest starts keeps the
       # heavy work outside the worker IPC window.
       - run: pnpm build
+      # The npm release surface packs with a bundled Kimi SDK; npm's
+      # trusted-publishing client is pinned to the exact version QA qualifies
+      # (a floated range already broke this boundary upstream). Smoke the real
+      # pack → empty-consumer install → public CLI → patched SDK chain on
+      # every CLI-affecting PR.
+      - name: Pin trusted-publishing npm client
+        run: npm install -g npm@11.5.1
+      - name: Release pack smoke (bundled Kimi SDK)
+        run: node scripts/release-pack-smoke.mjs
       # Keep the CLI behind Turbo so an unchanged first-tree-dev#test task can
       # use cache while its package script releases heap between batches.
       - run: pnpm exec turbo run test --filter=first-tree-dev --concurrency=1
--- a/.github/workflows/publish-npm-package.yml
+++ b/.github/workflows/publish-npm-package.yml
@@ -60,7 +60,7 @@ jobs:
           cache: "pnpm"
 
       - name: Ensure trusted publishing npm version
-        run: npm install -g npm@^11.5.1
+        run: npm install -g npm@11.5.1
 
       - name: Validate stable release tag
         id: release
@@ -281,7 +281,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Ensure trusted publishing npm version
-        run: npm install -g npm@^11.5.1
+        run: npm install -g npm@11.5.1
 
       - name: Resolve staging version
         id: version
@@ -485,7 +485,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Ensure trusted publishing npm version
-        run: npm install -g npm@^11.5.1
+        run: npm install -g npm@11.5.1
 
       - name: Resolve alpha version
         id: version
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,10 @@ WORKDIR /app
 
 # Copy package manifests for dependency layer caching. Only the packages
 # needed to build + run the SaaS server image: shared, server, web.
+# pnpm-workspace.yaml declares patchedDependencies, so the patch files must
+# be present before any pnpm install.
 COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
+COPY patches/ patches/
 COPY packages/shared/package.json packages/shared/
 COPY packages/server/package.json packages/server/
 COPY packages/web/package.json packages/web/
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -49,6 +49,9 @@
     "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs && node scripts/prune-claude-sdk-deps.mjs"
   },
   "license": "Apache-2.0",
+  "bundleDependencies": [
+    "@botiverse/kimi-code-sdk"
+  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/agent-team-foundation/first-tree.git",
@@ -60,6 +63,7 @@
   "dependencies": {
     "@anthropic-ai/claude-agent-sdk": "^0.3.187",
     "@botiverse/kimi-code-sdk": "0.26.0-botiverse.2",
+    "@antfu/utils": "^9.3.0",
     "@inquirer/prompts": "^8.3.0",
     "@openai/codex-sdk": "^0.144.1",
     "commander": "^13.1.0",
@@ -69,8 +73,10 @@
     "jose": "^6.0.0",
     "mdast-util-from-markdown": "^2.0.3",
     "semver": "^7.6.3",
+    "smol-toml": "^1.6.1",
     "ws": "^8.20.0",
     "yaml": "^2.8.3",
+    "yazl": "^3.3.1",
     "zod": "^4.0.0"
   },
   "devDependencies": {
--- a/apps/cli/src/__tests__/package-metadata.test.ts
+++ b/apps/cli/src/__tests__/package-metadata.test.ts
@@ -123,3 +123,26 @@ describe("workspace dependency-patch distribution", () => {
     expect(dockerignore).not.toMatch(/^patches\/?$/m);
   });
 });
+
+describe("trusted-publishing npm toolchain contract", () => {
+  const PUBLISH_WORKFLOW = join(REPO_ROOT, ".github", "workflows", "publish-npm-package.yml");
+  const CI_WORKFLOW = join(REPO_ROOT, ".github", "workflows", "ci.yml");
+
+  it("pins every trusted-publishing npm install to the exact QA-qualified version, with no moving selector", () => {
+    const workflow = readText(PUBLISH_WORKFLOW);
+    const installs = [...workflow.matchAll(/npm install -g npm@(\S+)/g)].map((match) => match[1]);
+    // prod CLI, staging CLI, and shared alpha must all be covered.
+    expect(installs.length).toBeGreaterThanOrEqual(3);
+    for (const selector of installs) {
+      expect(selector).toBe("11.5.1");
+    }
+    expect(workflow).not.toMatch(/npm@(latest|next|\^|~)/);
+  });
+
+  it("runs a real release-pack smoke under the pinned client on CLI-affecting PRs", () => {
+    const ci = readText(CI_WORKFLOW);
+    expect(ci).toContain("npm install -g npm@11.5.1");
+    expect(ci).toContain("node scripts/release-pack-smoke.mjs");
+    expect(existsSync(join(REPO_ROOT, "scripts", "release-pack-smoke.mjs"))).toBe(true);
+  });
+});
--- a/apps/cli/src/__tests__/package-metadata.test.ts
+++ b/apps/cli/src/__tests__/package-metadata.test.ts
@@ -80,4 +80,46 @@ describe("npm package metadata", () => {
 
     expect(dependencies["fs-native-extensions"]).toBe("^1.5.0");
   });
+
+  it("bundles the patched Kimi SDK and declares its runtime dependencies for consumers", () => {
+    const pkg = readJson(join(CLI_ROOT, "package.json")) as Record<string, unknown>;
+    // The workspace pins the Kimi SDK through a pnpm patch (see
+    // pnpm-workspace.yaml). npm consumers cannot apply that patch, so the
+    // tarball must carry the patched artifact as a bundled dependency, and
+    // the SDK's own runtime deps must be regular CLI dependencies so the
+    // bundled copy resolves them from the consumer's top-level node_modules.
+    expect(pkg.bundleDependencies).toEqual(["@botiverse/kimi-code-sdk"]);
+    const dependencies = packageDependencies(pkg);
+    for (const sdkRuntimeDep of ["@antfu/utils", "smol-toml", "yazl"]) {
+      expect(dependencies[sdkRuntimeDep]).toBeDefined();
+    }
+    expect(dependencies["@botiverse/kimi-code-sdk"]).toBe("0.26.0-botiverse.2");
+  });
+});
+
+describe("workspace dependency-patch distribution", () => {
+  it("copies the pnpm patch files into the Docker dependency stage before install", () => {
+    const dockerfile = readText(join(REPO_ROOT, "Dockerfile"));
+    const copyPatches = dockerfile.indexOf("COPY patches/ patches/");
+    const install = dockerfile.indexOf("pnpm install --frozen-lockfile");
+    expect(copyPatches).toBeGreaterThan(-1);
+    expect(install).toBeGreaterThan(-1);
+    expect(copyPatches).toBeLessThan(install);
+  });
+
+  it("points every patchedDependencies entry at a patch file that exists and ships to Docker", () => {
+    const workspaceYaml = readText(join(REPO_ROOT, "pnpm-workspace.yaml"));
+    const match = workspaceYaml.match(/patchedDependencies:\n((?:[ \t]+\S.*\n?)+)/);
+    expect(match).not.toBeNull();
+    if (!match) throw new Error("patchedDependencies block missing from pnpm-workspace.yaml");
+    const entries = [...match[1].matchAll(/^\s+(['"]?)([^'":]+)\1:\s*(.+?)\s*$/gm)];
+    expect(entries.length).toBeGreaterThan(0);
+    const dockerignore = readText(join(REPO_ROOT, ".dockerignore"));
+    for (const entry of entries) {
+      const patchPath = entry[3];
+      expect(patchPath.startsWith("patches/")).toBe(true);
+      expect(existsSync(join(REPO_ROOT, patchPath))).toBe(true);
+    }
+    expect(dockerignore).not.toMatch(/^patches\/?$/m);
+  });
 });
--- a/apps/cli/tests/portable-builder.test.ts
+++ b/apps/cli/tests/portable-builder.test.ts
@@ -37,6 +37,7 @@ import {
   portableBuildRoots,
   portableTarCreateArgs,
   readWorkspacePackageManager,
+  readWorkspacePatchedDependencies,
   relativizeInternalSymlinks,
   renderInstallerForChannel,
   resolveNodeVersion,
@@ -304,6 +305,47 @@ describe("portable builder helpers", () => {
     expect(() => readWorkspacePackageManager({ packageManager: "pnpm@^10" })).toThrow(/exact pnpm version/);
     expect(() => readWorkspacePackageManager({})).toThrow(/exact pnpm version/);
     expect(readWorkspacePackageManager()).toMatch(/^pnpm@\d+\.\d+\.\d+$/);
+  });
+
+  it("carries workspace patchedDependencies into the portable app manifest", () => {
+    const patched = readWorkspacePatchedDependencies();
+    // The Kimi SDK resume-drain patch must reach the synthetic portable app;
+    // without it the app's generated lockfile resolves the unpatched SDK.
+    expect(patched).toHaveProperty(
+      "@botiverse/kimi-code-sdk@0.26.0-botiverse.2",
+      "patches/@botiverse__kimi-code-sdk@0.26.0-botiverse.2.patch",
+    );
+    for (const patchPath of Object.values(patched)) {
+      expect(patchPath.startsWith("patches/")).toBe(true);
+      expect(existsSync(join(REPO_ROOT, patchPath))).toBe(true);
+    }
+
+    const appPackage = packageJsonForApp({
+      channelConfig: { packageName: "first-tree", binName: "first-tree", aliasName: "ft" },
+      version: "1.2.3",
+      dependencies: { commander: "13.1.0" },
+      sourcePackage: {
+        description: "First Tree CLI",
+        license: "MIT",
+        repository: { type: "git", url: "https://example.test/first-tree.git" },
+        engines: { node: ">=22.13" },
+      },
+      patchedDependencies: patched,
+    });
+    expect(appPackage.pnpm).toEqual({ patchedDependencies: patched });
+
+    const unpatched = packageJsonForApp({
+      channelConfig: { packageName: "first-tree", binName: "first-tree", aliasName: "ft" },
+      version: "1.2.3",
+      dependencies: { commander: "13.1.0" },
+      sourcePackage: {
+        description: "First Tree CLI",
+        license: "MIT",
+        repository: { type: "git", url: "https://example.test/first-tree.git" },
+        engines: { node: ">=22.13" },
+      },
+    });
+    expect(unpatched).not.toHaveProperty("pnpm");
   });
 
   it("fails when a portable dependency is missing from locked pnpm output", () => {
--- a/packages/client/src/__tests__/agent-slot.test.ts
+++ b/packages/client/src/__tests__/agent-slot.test.ts
@@ -30,6 +30,7 @@ type FakeSessionState = {
   >;
   applyStaleChatIds: ReturnType<typeof vi.fn<(chatIds: string[]) => void>>;
   noteBindRecoveryComplete: ReturnType<typeof vi.fn<() => void>>;
+  reconcileReplayFencesWithServer: ReturnType<typeof vi.fn<() => Promise<void>>>;
   updateTransport: ReturnType<typeof vi.fn<(sdk: unknown, agentConfigCache?: unknown) => void>>;
   shutdown: ReturnType<typeof vi.fn<(reason?: string, opts?: unknown) => Promise<void>>>;
 };
@@ -244,6 +245,7 @@ function installMocks(
           handleCommand: vi.fn(async () => {}),
           applyStaleChatIds: vi.fn(),
           noteBindRecoveryComplete: vi.fn(),
+          reconcileReplayFencesWithServer: vi.fn(async () => {}),
           updateTransport: vi.fn(),
           shutdown: vi.fn(async () => {}),
         };
@@ -287,6 +289,10 @@ function installMocks(
 
       noteBindRecoveryComplete(): void {
         this.state.noteBindRecoveryComplete();
+      }
+
+      reconcileReplayFencesWithServer(): Promise<void> {
+        return this.state.reconcileReplayFencesWithServer();
       }
 
       updateTransport(sdk: unknown, agentConfigCache?: unknown): void {
@@ -750,6 +756,7 @@ describe("AgentSlot", () => {
 
     const nextSdk = makeSdk({ activeRuntimeChatIds: ["chat-2"] });
     vi.mocked(sdk.listActiveRuntimeChatIds).mockClear();
+    session.reconcileReplayFencesWithServer.mockClear();
 
     connection.emit("agent:bound", {
       agentId: "agent-1",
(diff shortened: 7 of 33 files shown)


baixiaohang (changes requested):
- Rationale: The durable replay fence still has admission and storage-failure paths that can re-enter an unsafe Kimi delivery after a crash.

**Risk level**: A
- Semantic lift: none (already A); the added code changes crash recovery, delivery ordering, and provider replay safety.

**PR summary**
- Problem: A Kimi parent turn could ACK and idle-suspend while background agents still owned work; the new commit also addresses a live-QA reproduction where crash recovery replayed a child’s non-idempotent write.
- Approach: Keep the main turn open through child completion, separate main/child event roles, ship the patched SDK in every distribution, and persist per-delivery unsafe-effect fences that gate provider re-entry after restart.
- Impacted modules: Client Kimi handler and session runtime, CLI/npm packaging, portable artifacts, Docker dependency install, QA cases.

**Review findings**
❌ 1. A fenced head does not block later deliveries in the same chat. `routeMessage` tests only the current `(chatId, messageId)`, and the added test explicitly starts `msg-2` while fenced `msg-1` remains open. Inbox ACK is prefix-based: when `msg-2` completes, `ackThrough` cannot cross the non-terminal `msg-1`, requests recovery, but `DeliveryToken.complete` still returns `settled`; Kimi then clears `msg-2`'s fence. Redelivery can therefore run `msg-2` (and its unsafe effects) again while `msg-1` stays fenced. Hold the entire chat/FIFO tail behind the fenced head, or otherwise make completion retain custody until that prefix is resolved. [R4/R5 / `packages/client/src/runtime/session-manager.ts:1656`, `packages/client/src/__tests__/session-manager.test.ts:2851`]

❌ 2. `ReplayFenceStore.load()` treats every `readFileSync` failure as “file missing.” Permission errors, I/O errors, or a directory at the configured path therefore produce an empty loaded store and enable dispatch, losing exactly the persisted safety fact this gate relies on. Only `ENOENT` may mean an empty store; all other read failures must throw `ReplayFenceError` so SessionManager withholds dispatch. [R4 / `packages/client/src/runtime/replay-fence.ts:70`]

❌ 3. A fence write failure is acted on only after `observeOneAttempt()` returns at the main `turn.ended`. The SDK is free to continue the unsafe tool and the rest of the turn after the synchronous event listener records `fenceError`; if the process dies in that interval, there is neither a durable fence nor a terminal ACK, so restart replays the effect. Persist the fence at a pre-effect authorization boundary (or otherwise synchronously prevent the tool from starting when persistence fails), rather than converting the delivery to a consumed error only after the provider turn finishes. [R4 / `packages/client/src/handlers/kimi-code.ts:450`, `packages/client/src/handlers/kimi-code.ts:605`]

**Action taken**
- Submitted request changes against head `15744e57f562283ab8ccd07acd49c7776ca14d38`.

baixiaohang (changes requested):
- Rationale: The three prior replay-fence blockers are fixed, but the retained fence has no cleanup path when recovery later commits the ACK, permanently blocking that chat after a transient ACK failure.

**Risk level**: A
- Semantic lift: none (already A); the change governs durable inbox settlement and unsafe-effect replay recovery.

**PR summary**
- Problem: Kimi could ACK a parent turn while background agents still owned work, and a crash after a child write could redeliver and repeat that unsafe effect.
- Approach: Keep the main provider turn open through child completion, isolate child lifecycle events, ship the patched SDK everywhere, and persist replay fences that withhold unsafe deliveries until inbox settlement is confirmed.
- Impacted modules: Client Kimi handler and inbox/session runtime, SDK patching, CLI/portable/Docker packaging, QA cases.

**Review findings**
❌ 1. A fence retained because `finishTurn()` returns `retry` is never cleared when recovery later settles the same terminal delivery. On an ACK rejection, `ackThroughNow()` keeps the terminal ledger entry and initiates recovery; the redelivered terminal entry then retries ACK from `receive()` and can successfully remove the ledger entry. That second ACK path has no callback to the Kimi handler or `ReplayFenceStore`, while `clearFencedDelivery()` only runs during the original completion and intentionally returns on `retry`. The server can therefore be fully ACKed while the stale chat-wide fence remains forever, so every later message is withheld until manual file repair. Reconcile fence cleanup with eventual terminal ACK success (including the redelivery retry path), and cover ACK-fail -> recovery-redelivery -> confirmed ACK -> later chat delivery. [R4/R5 / `packages/client/src/handlers/kimi-code.ts:285`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:94`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:531`]

✅ The previous chat-tail admission, non-ENOENT load, and pre-effect persistence-failure findings are resolved at head `c6a353b7fdf143ed5977ba1e1110b0107adf2b52`.

**Action taken**
- Submitted request changes against head `c6a353b7fdf143ed5977ba1e1110b0107adf2b52`.

baixiaohang (changes requested):
- Rationale: The ACK/recovery reconciliation finding is fixed, but a fence-write failure blocks only the first unsafe tool call; later unsafe calls in the same turn can execute unfenced.

**Risk level**: A
- Semantic lift: none (already A); the change controls the pre-effect replay-safety boundary and durable inbox settlement.

**PR summary**
- Problem: Kimi could ACK a parent turn while background agents still owned work, and crash recovery could repeat a child’s non-idempotent effect.
- Approach: Drain background agents before parent completion, isolate child lifecycle output, ship the patched SDK in every artifact, fence unsafe effects at an awaited SDK authorization hook, and reconcile retained fences when a later ACK commits.
- Impacted modules: Client Kimi handler and inbox/session runtime, SDK patching, CLI/portable/Docker packaging, QA cases.

**Review findings**
❌ 1. A failed fence write incorrectly leaves `observation.unsafeFenced = true`. The hook sets that flag before `fenceUnsafeDelivery()` and does not reset or distinguish it on failure; the SDK converts the throw into an error result and continues the turn (the new real-bundle test explicitly asserts that continuation). Any second unsafe call—either the next call in the same tool batch or a model retry in the next step—then takes the `unsafeFenced` fast path and is allowed to execute even though no durable fence exists. A crash after that effect reopens the exact duplicate-write window this PR is meant to close. Keep a failed state fail-closed for every later unsafe call (or stop the turn immediately), and cover two unsafe calls where persistence fails so neither effect executes. [R4/R5 / `packages/client/src/handlers/kimi-code.ts:551`, `packages/client/src/handlers/kimi-code.ts:559`, `packages/client/src/__tests__/kimi-sdk-authorize-hook.test.ts:261`]

✅ The prior ACK-fail -> recovery-redelivery -> confirmed-ACK fence reconciliation finding is resolved at head `7f56ea528008c4eb8d13a4824c9959c0d9a3d0ef`.

**Action taken**
- Submitted request changes against head `7f56ea528008c4eb8d13a4824c9959c0d9a3d0ef`.

baixiaohang (changes requested):
- Rationale: The previous pre-effect fail-open remains unchanged, and the new startup reconciliation treats a recovery reset count as ACK truth even though zero can coexist with pending unacked backlog.

**Risk level**: A
- Semantic lift: none (already A); the change controls crash recovery, unsafe-effect replay fencing, and server/client delivery truth.

**PR summary**
- Problem: Kimi could ACK a parent turn while background agents still owned work, and crash recovery could repeat a child’s non-idempotent effect.
- Approach: Drain child work before parent completion, separate child lifecycle output, ship the patched SDK everywhere, fence unsafe calls at an awaited SDK authorization hook, reconcile fences on committed ACK, and now attempt to clear post-crash stale fences from server recovery results.
- Impacted modules: Client Kimi handler, inbox/session runtime and WebSocket connection, SDK patching, release packaging, QA cases.

**Review findings**
❌ 1. The prior fence-write failure state bug is unchanged in this head. `unsafeFenced` is still set before persistence and remains true when `fenceUnsafeDelivery()` fails; because the SDK turns that throw into a blocked result and continues, a second unsafe call takes the “already fenced” fast path and executes without any durable fence. Keep the failed state fail-closed for all later unsafe calls (or stop the turn), with a two-call regression. [R4/R5 / `packages/client/src/handlers/kimi-code.ts:551`, `packages/client/src/handlers/kimi-code.ts:559`]

❌ 2. `resetCount === 0` is not proof that the fenced delivery was ACKed. `recoverUnackedForScope()` counts only rows currently in `delivered`; an unacked fenced row can already be `pending` (for example, the process crashes after a prior recovery/bind reset but before redelivery), and the server even has a test where recovery returns zero while backlog remains pending at the per-chat cap. The accepted frame is also sent before the recovery drain. Startup reconciliation would clear the fence, then that pending row can be delivered and re-enter Kimi, replaying the unsafe effect. Clearing requires message-specific durable ACK/absence truth, not the number of `delivered -> pending` transitions. [R4/R5 / `packages/client/src/runtime/session-manager.ts:1113`, `packages/client/src/runtime/session-manager.ts:1135`, `packages/server/src/services/inbox.ts:688`, `packages/server/src/api/agent/ws-client.ts:1860`]

✅ The coordinator callback that clears a retained fence after a later confirmed ACK remains correct.

**Action taken**
- Submitted request changes against head `e415e5aaa892b0073e94b11ec87f2a9b6b7ec721`.

baixiaohang (changes requested):
- Rationale: The two prior safety findings are fixed, but chat-wide outstanding truth cannot distinguish an ACKed stale fenced head from a newer unacked tail, so that tail can make the chat permanently unrecoverable.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox/replay recovery.

**PR summary**
- Problem: A Kimi turn could ACK an inbound delivery while background agents still owned unfinished work, allowing restart/re-entry to duplicate unsafe effects or lose custody.
- Approach: Drain background work before parent completion, isolate child event streams, bundle the SDK resume fix, and durably fence unsafe deliveries until server-side settlement is known.
- Impacted modules: Kimi handler, client session/inbox lifecycle, server inbox recovery, CLI/portable packaging, and release tests.

**Review findings**

❌ 1. `unackedOutstanding` is scoped to the whole chat, while replay fences are scoped to individual deliveries. Consider a stale fence for `msg-1`: the server successfully ACKed `msg-1`, but the client crashed before clearing its local fence; then a newer `msg-2` becomes pending in the same chat. `countUnackedForScope()` returns `1` solely because of `msg-2`, so reconciliation keeps the stale `msg-1` fence. The chat-wide route gate then withholds `msg-2`, which means it can never be processed and ACKed, so the count can never reach zero. The 30-second retry loop cannot converge. Please query durable settlement for each fenced message (or an equivalent authoritative cursor/status), clear only settled fences, and cover the real sequence: stale ACKed fence + newer pending tail → stale fence clears → tail is admitted. The current convergence test supplies `1 → 0` without a path by which the gated backlog can actually settle. [R4/R5 / `packages/client/src/runtime/session-manager.ts:1119`, `packages/server/src/services/inbox.ts:690`, `packages/client/src/__tests__/session-manager.test.ts:3332`]

✅ The previous fail-open path after a fence write error is closed: subsequent unsafe calls remain blocked, partial multi-message writes roll back, and the real SDK fixture covers two attempted writes.

✅ The previous `resetCount === 0` false-clear is closed: pending and delivered rows now both count as unsettled, and missing/readback-failed truth remains fail-closed.

**Action taken**
- Submitted request changes against exact head `05fbc047ffff59bffbbee0ec1f5bd0d98d5fc567`.

baixiaohang (changes requested):
- Rationale: Per-message settlement closes the prior deadlock only while the chat backlog fits under the 500-ID response cap; an oversized tail still leaves an already-settled fenced head permanently blocking the chat.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox/replay recovery.

**PR summary**
- Problem: A Kimi turn could ACK an inbound delivery while background agents still owned unfinished work, allowing restart/re-entry to duplicate unsafe effects or lose custody.
- Approach: Drain background work before parent completion, isolate child event streams, bundle the SDK resume fix, and durably fence unsafe deliveries until server-side settlement is known.
- Impacted modules: Kimi handler, client session/inbox lifecycle, server inbox recovery, CLI/portable packaging, and release tests.

**Review findings**

❌ 1. The new per-message truth is omitted when a chat has more than 500 unacked notify rows. In the target recovery scenario, let stale fenced `msg-1` already be ACKed and let 501 newer messages be pending/delivered. `listUnackedMessageIdsForScope()` returns `null`, while `unackedOutstanding` remains 501, so the client takes the unknown/nonzero branch and retains the stale fence. The chat-wide gate prevents all 501 tail messages from entering the provider and being ACKed, so the backlog can never fall under the cap and every 30-second retry returns the same result. This is the original deadlock at a deterministic backlog threshold. Please query settlement for the concrete fenced message IDs (or an equivalent authoritative cursor/status) without requiring the entire chat backlog to fit in one frame, and cover `> UNACKED_MESSAGE_IDS_LIMIT` newer tail rows in the recovery test. [R4/R5 / `packages/server/src/services/inbox.ts:704`, `packages/client/src/runtime/session-manager.ts:1172`]

✅ For backlogs at or below the cap, the previous stale-head/new-tail blocker is closed: settlement is message-specific, local clear failures retain their proof, and the timer retries without needing another dispatch.

✅ The two earlier safety findings remain closed: fence persistence failures keep later unsafe calls blocked, and pending rows are no longer mistaken for settled deliveries.

**Action taken**
- Submitted request changes against exact head `fcaca0aa416b881fb165bcc5b32ab49ded13fad0`.

baixiaohang (changes requested):
- Rationale: The targeted settlement probe fixes the whole-backlog cap deadlock, but the client does not chunk fence sets to the probe protocol's own 50-ID limit, so a chat with 51 fences can never reconcile.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox/replay recovery.

**PR summary**
- Problem: A Kimi turn could ACK an inbound delivery while background agents still owned unfinished work, allowing restart/re-entry to duplicate unsafe effects or lose custody.
- Approach: Drain background work before parent completion, isolate child event streams, bundle the SDK resume fix, and durably fence unsafe deliveries until server-side settlement is known.
- Impacted modules: Kimi handler, client session/inbox lifecycle, server inbox recovery, CLI/portable packaging, and release tests.

**Review findings**

❌ 1. The new wire schema caps one probe at `INBOX_FENCE_PROBE_MAX_IDS = 50`, but `SessionManager` collects every fence for the chat and `AgentSlot` forwards the complete array in one `sendInboxFenceProbe()` call; there is no chunking. With 51 persisted fences, the server rejects the frame as malformed before running the probe, the client's pending probe eventually times out, and every 30-second retry sends the same invalid frame, so even fully settled deliveries remain permanently fenced. This is reachable under the supported `maxInFlightPerAgentChat` range (up to 1024), and the server service comment already assumes the client chunks. Please chunk using the shared wire constant and return settlement proof only after every chunk succeeds (or otherwise preserve fail-closed partial-failure semantics), then cover 51+ fenced IDs and a later-chunk failure. [R1/R4/R5 / `packages/shared/src/schemas/inbox-frames.ts:105`, `packages/client/src/runtime/session-manager.ts:1154`, `packages/client/src/runtime/agent-slot.ts:368`]

✅ The prior whole-backlog threshold blocker is closed: the server now queries only the concrete fenced deliveries, and a missing row remains unknown rather than being treated as settled.

✅ The post-clear redrive is now server-backed rather than an in-memory buffer, and fence-withheld ledger entries are explicitly re-admitted once on redelivery.

**Action taken**
- Submitted request changes against exact head `2748e4259aefc8693f8817b83f2cccf9a0837da2`.

baixiaohang (changes requested):
- Rationale: The new ledger-settlement and fence-clear convergence fixes are sound, but the previous 51-fence protocol-limit blocker is unchanged: the client still sends every fenced delivery ID for a chat in one frame even though the shared schema caps each probe at 50 IDs.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox and replay recovery behavior.

**PR summary**
- Problem: A Kimi turn could be ACKed while background agents still owned unfinished work, allowing custody to be released too early and unsafe effects to be replayed after recovery.
- Approach: Drain background work before turn completion, isolate child-agent lifecycle events while retaining tool effects, carry the SDK resume fix through release artifacts, and use durable replay fences to prevent unsafe provider re-entry until delivery settlement is proven.
- Impacted modules: Kimi handler, session/inbox recovery, server inbox protocol, shared frame schemas, CLI/portable packaging, and release images.

**Review findings**

❌ 1. **The 51-fence reconciliation deadlock from the prior review is unchanged.** `INBOX_FENCE_PROBE_MAX_IDS` rejects more than 50 IDs, while `SessionManager` still gathers every fence for one chat and `AgentSlot` passes that complete array to one `sendInboxFenceProbe` call. With 51 persisted fences, the server rejects the frame before probing, the client times out, and the 30-second loop repeats the same invalid frame forever; already-settled fences can therefore never clear. This is reachable because `maxInFlightPerAgentChat` supports values up to 1024. Please chunk probes using the shared limit, return settlement proof only after all chunks succeed (or otherwise preserve fail-closed partial semantics), and cover both 51+ IDs and a later-chunk failure. [R1/R4/R5 / `packages/shared/src/schemas/inbox-frames.ts:105`, `packages/client/src/runtime/session-manager.ts:1154`, `packages/client/src/runtime/agent-slot.ts:368`, `packages/shared/src/config/server-config.ts:468`]

✅ 2. The incremental change correctly removes probe-proven-settled heads from the local ledger and routes probe, retry, and committed-ACK clears through one convergence path.

✅ 3. Startup no longer issues duplicate immediate probes, and reconciliation ignores response IDs that were not requested.

**Action taken**
- Submitted request changes against exact head `bc739fc82e753176e1c9c398a430659ab0cd4b69`.

baixiaohang (changes requested):
- Rationale: Probe chunking now closes the prior 51-fence deadlock, but the new safety tombstone is bounded below the supported in-flight window, so an evicted probe-settled unsafe delivery can re-enter the provider from a duplicate frame.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox custody and unsafe-effect replay recovery.

**PR summary**
- Problem: A Kimi turn could be ACKed while background agents still owned unfinished work, allowing custody to be released too early and unsafe tool effects to be replayed after recovery.
- Approach: Drain background work before turn completion, isolate child-agent lifecycle events while retaining tool effects, ship the SDK resume fix through release artifacts, and use durable replay fences plus server settlement probes to prevent unsafe provider re-entry.
- Impacted modules: Kimi handler, client session/inbox recovery, server inbox protocol, shared schemas, CLI/portable packaging, and release images.

**Review findings**

❌ 1. **The probe-settled safety tombstone can evict entries that are still within the supported delivery window.** `fenceSettledKeys` is a global `Deduplicator(1000)`, whose FIFO implementation drops the oldest key at capacity, but the server supports 1,024 in-flight deliveries for one agent/chat and 8,192 by default for one agent. `clearProvenSettledFences()` can pass all 1,001+ proven ledger entries to `settleReplayFencedEntries()` after the chunked probe round succeeds; adding those tombstones evicts the earliest one. A later duplicate/stale frame for that earliest `(chatId, entryId, messageId)` then misses the only new guard. The ordinary deduplicator also does not make this safe—it either has evicted the old key or deliberately falls through to reprocessing when no active ledger entry exists—and probe settlement does not populate `recentlySettled`, so the delivery can reach provider routing after its fence has already been deleted and replay the unsafe effect. Keep safety tombstones for the full lifetime in which such frames can arrive (for example, scoped to a connection/recovery generation), or otherwise size/retire them from an authoritative bound; cover more than 1,000 probe-settled ledger entries followed by a duplicate of the oldest entry. [R3/R4/R5 / `packages/client/src/runtime/inbox-delivery-coordinator.ts:92`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:100`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:141`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:275`, `packages/client/src/runtime/deduplicator.ts:23`, `packages/shared/src/config/server-config.ts:460`]

✅ 2. The prior 51-fence protocol-limit blocker is closed: probes now use `INBOX_FENCE_PROBE_MAX_IDS`, and the latest head keeps the entire round fail-closed when any chunk fails.

✅ 3. Post-fence recovery is now retried as coalesced debt, and ordinary never-fenced commits no longer trigger destructive recovery.

**Action taken**
- Submitted request changes against exact head `e36da949382745ba5b9a74600093f27bf115a567`.

baixiaohang (changes requested):
- Rationale: The prior tombstone-capacity and recovery-debt bypass findings are fixed, but settlement can still clear a fence before an already-arrived frame has entered the coordinator ledger, leaving that stale frame with no tombstone and allowing unsafe provider re-entry.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox custody and unsafe-effect replay recovery.

**PR summary**
- Problem: A Kimi turn could be ACKed while background agents still owned unfinished work, allowing custody to be released too early and unsafe tool effects to be replayed after recovery.
- Approach: Drain background work before parent completion, isolate child-agent lifecycle events while retaining tool effects, ship the SDK resume fix through release artifacts, and use durable replay fences plus server settlement probes to prevent unsafe provider re-entry.
- Impacted modules: Kimi handler, client session/inbox recovery, server inbox protocol, shared schemas, CLI/portable packaging, and release images.

**Review findings**

❌ 1. **A stale frame can sit before `receive()` while reconciliation clears its fence, so no tombstone is created.** `SessionManager.dispatch()` awaits the chat's in-flight `suspending` promise before calling `inboxDelivery.receive()`. During that await, the independent fence reconcile loop can prove the delivery settled and call `settleReplayFencedEntries()`, but that method returns when no ledger exists and only creates tombstones for entries already tracked in the ledger. When the paused dispatch resumes, `receive()` sees neither an active entry nor a tombstone, and `routeMessage()` sees that the durable fence has already been removed; the stale delivery can therefore re-enter Kimi and repeat the unsafe effect. The high-water test calls `receive()` before settlement, so it does not cover this race. Record settlement suppression independently of current ledger membership (for example, retain the proven `(chatId, messageId)` fact and bind any later entry before routing), or move the receive/safety registration before all awaits that can overlap reconciliation. Please add a regression that pauses dispatch on `suspending`, clears the fence through the server probe while the entry is still pre-ledger, then releases suspension and proves zero provider entries. [R4/R5 / `packages/client/src/runtime/session-manager.ts:630`, `packages/client/src/runtime/session-manager.ts:646`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:285`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:291`]

✅ 2. The previous 1,000-entry eviction blocker is closed for entries already in the ledger: tombstones are now per chat, retain 4,096 keys, and no longer suffer unrelated-chat eviction.

✅ 3. The previous recovery-debt bypass is closed: a locally re-admitted duplicate remains held until a recovery request is actually accepted.

**Action taken**
- Submitted request changes against exact head `a6e945a4f2478bdf804c4b8063461b748e1342b3`.

baixiaohang (dismissed):
- Rationale: The pre-ledger probe race is fixed, but confirmed-ACK fence clearing still does not register the new permanent settlement marker, so a stale duplicate can resume Kimi and execute before the generic post-route ownership check notices it was already settled.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` → A
- Semantic lift: none (already A); this changes durable inbox custody and unsafe-effect replay recovery.

**PR summary**
- Problem: A Kimi turn could be ACKed while background agents still owned unfinished work, allowing custody to be released too early and unsafe tool effects to be replayed after recovery.
- Approach: Drain background work before parent completion, isolate child-agent lifecycle events while retaining tool effects, ship the SDK resume fix through release artifacts, and use durable replay fences plus server settlement truth to prevent unsafe provider re-entry.
- Impacted modules: Kimi handler, client session/inbox recovery, server inbox protocol, shared schemas, CLI/portable packaging, and release images.

**Review findings**

❌ 1. **The new permanent settlement marker is only populated by the probe path; the confirmed-ACK path still clears unsafe fences without it.** After `ackThroughNow()` confirms an ACK, `onDeliveriesCommitted` calls `reconcileReplayFences()`, which removes the durable fence but never calls `settleReplayFencedEntries()` or otherwise records `(chatId, messageId)` in `fenceSettledMessages`. It relies only on the generic `recentlySettled` check. That check happens in `markRouteOwned()` after a handler route returns, but Kimi `resume()` (and fresh `start()`) awaits the entire `runTurn()` before returning its ownership receipt. Therefore, after a fenced delivery ACKs successfully and the session later suspends, a stale duplicate can create a new ledger entry, pass the now-unfenced route gate, resume Kimi, and execute its unsafe tool before `markRouteOwned()` finally notices the old settlement and cancels the route. This is unsafe even below the 1,000-entry generic dedup horizon. Route every server-proven settlement of a fenced delivery—probe or confirmed ACK—through the same permanent suppression fact before clearing/admitting, and cover: fenced delivery → confirmed ACK → suspend → stale duplicate → zero Kimi provider/tool entry. [R1/R4/R5 / `packages/client/src/runtime/inbox-delivery-coordinator.ts:98`, `packages/client/src/runtime/inbox-delivery-coordinator.ts:655`, `packages/client/src/runtime/session-manager.ts:1356`, `packages/client/src/runtime/session-manager.ts:1365`, `packages/client/src/runtime/session-manager.ts:2251`, `packages/client/src/handlers/kimi-code.ts:1044`]

✅ 2. The previous pre-ledger probe race is closed: server-proven settlement is now registered unconditionally by `(chatId, messageId)` before ledger filtering, with no time-based expiry.

✅ 3. The prior capacity and recovery-debt bypass fixes remain intact.

**Action taken**
- Submitted request changes against exact head `82cc51895d9a199600513e4cf87468f4934280e3`.

PR #2129, opened 2026-08-01
fix(cli): materialize bundled Kimi SDK paths for npm registry pack

yuezengwu:
## Summary
- Staging npm publish on `main@2df85072…` failed with registry **E415** because `npm pack`/`publish` under pnpm followed `bundleDependencies` symlinks into `package/../../node_modules/.pnpm/...` traversal paths.
- Prepack now materializes the bundled `@botiverse/kimi-code-sdk` closure as real in-package directories (and postpack restores the pnpm symlinks), keeping `bundleDependencies`, the pnpm patch, and exact npm `11.5.1` unchanged.
- Release-pack smoke enumerates every tarball entry and rejects traversal/absolute/escaping-link paths before empty-consumer install / CLI / patched SDK import checks; contract tests + a selftest cover the E415 layout.

## Test plan
- [x] Neuter control: pack without materialize → 761 traversal entries; safety helper rejects
- [x] `npm pack` after fix → 0 unsafe entries; canonical `package/node_modules/@botiverse/kimi-code-sdk/**` present
- [x] `node scripts/release-pack-smoke.mjs` PASS (tarball sha256 `51fa0eb92e777e3b61c807cf0689b344635076ea718ba0641e755c51c5bb04aa`, 3,771,903 B, 844 entries)
- [x] `package-metadata.test.ts` 11/11
- [x] `pnpm check` (16 warnings baseline) / `pnpm typecheck` / `pnpm build`
- [ ] CI Test CLI + release-pack smoke on Linux/Node 24/npm 11.5.1
- [ ] Formal QA / staging publish re-run after merge (not done in this PR)

## Notes
- Does **not** remove `bundleDependencies` or change Kimi runtime semantics.
- Shared alpha publish path untouched (no Kimi bundling).
- No merge / no live npm publish from this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,11 +223,16 @@ jobs:
       # trusted-publishing client is pinned to the exact version QA qualifies
       # (a floated range already broke this boundary upstream). Smoke the real
       # pack → empty-consumer install → public CLI → patched SDK chain on
-      # every CLI-affecting PR.
+      # every CLI-affecting PR. The turbo context-integration cache selftest
+      # wraps that same runSmoke() path after proving a cache hit restores
+      # context-integration/** — keep it as a shell step (not Vitest), because
+      # the forced rebuild + pack chain exceeds the worker IPC heartbeat that
+      # previously failed jobs with `Timeout calling "onTaskUpdate"` even when
+      # every assertion passed.
       - name: Pin trusted-publishing npm client
         run: npm install -g npm@11.5.1
-      - name: Release pack smoke (bundled Kimi SDK)
-        run: node scripts/release-pack-smoke.mjs
+      - name: Release pack smoke (bundled Kimi SDK + turbo context-integration cache)
+        run: node scripts/release-pack-smoke.mjs selftest-turbo-context-integration-cache
       # Keep the CLI behind Turbo so an unchanged first-tree-dev#test task can
       # use cache while its package script releases heap between batches.
       - run: pnpm exec turbo run test --filter=first-tree-dev --concurrency=1
--- a/apps/cli/.gitignore
+++ b/apps/cli/.gitignore
@@ -15,4 +15,3 @@
 # Written by `materialize-bundled-deps.mjs prepare` during prepack so postpack
 # can restore pnpm symlinks. Never commit a stranded pack-time workspace mutation.
 /.bundled-deps-materialize.json
-
--- a/apps/cli/.gitignore
+++ b/apps/cli/.gitignore
@@ -11,3 +11,8 @@
 # stops at the first hit). If the dev CLI starts behaving as if a recent
 # `skills/` edit didn't take, run `rm -rf apps/cli/skills` and re-test.
 /skills/
+
+# Written by `materialize-bundled-deps.mjs prepare` during prepack so postpack
+# can restore pnpm symlinks. Never commit a stranded pack-time workspace mutation.
+/.bundled-deps-materialize.json
+
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -44,8 +44,8 @@
     "typecheck": "tsc --noEmit",
     "test": "node scripts/run-vitest-batches.mjs",
     "coverage": "pnpm run build && vitest run --coverage --passWithNoTests",
-    "prepack": "rm -rf skills && cp -R ../../skills .",
-    "postpack": "rm -rf skills context-integration",
+    "prepack": "rm -rf skills && cp -R ../../skills . && node scripts/materialize-bundled-deps.mjs prepare",
+    "postpack": "node scripts/materialize-bundled-deps.mjs restore && rm -rf skills context-integration",
     "postinstall": "node scripts/prune-codex-runtime-binary.mjs && node scripts/prune-claude-runtime-binary.mjs && node scripts/prune-claude-sdk-deps.mjs"
   },
   "license": "Apache-2.0",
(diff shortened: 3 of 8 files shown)


baixiaohang (changes requested):
- Rationale: The success path addresses the registry E415 failure, but the pack-time workspace mutation is not recoverable if `prepare` fails partway through.

**Risk level**: A
- Path baseline: includes `apps/cli/**` → A
- Semantic lift: none

**PR summary**
- Problem: Staging and production npm publishing can produce pnpm traversal entries that the registry rejects, preventing users from receiving an installable CLI containing the patched Kimi SDK.
- Approach: Materialize the bundled SDK dependency closure as real directories during pack, restore pnpm links afterward, and reject unsafe tar paths before consumer-install smoke checks.
- Impacted modules: `apps/cli` package lifecycle and package metadata tests; release-pack smoke and tarball validation scripts.

**Review findings**
❌ 1. `prepare()` records each original link only in memory and writes the manifest after the entire mutation loop. If any later missing dependency, `cpSync` failure, or interruption occurs after line 117, prepack exits without running postpack and without a manifest that can restore the links already removed. A subsequent `prepare` sees those stranded real directories, skips them at lines 109–111, and can never recover their original targets. Persist an atomic recovery record before each unlink (or stage copies first and commit/rollback the swaps under a guard), and add a regression that forces a mid-prepare failure and proves every original symlink is restored. [R4 / `apps/cli/scripts/materialize-bundled-deps.mjs:117`]
⚠️ 2. The safety helper says it enforces that entries stay below `package/`, but a relative entry such as `outside.txt` passes because non-`package/` names are treated as their own relative path. Require the canonical `package/` root explicitly and cover a rootless entry in the selftest so this release guard matches its stated invariant. [`scripts/npm-tarball-safety.mjs:76`]

**Action taken**
- Submitted request changes on head `59b9addc3be0283b145d671fd9098d497512820c`.

baixiaohang (changes requested):
- Rationale: The original recovery and tar-root findings are resolved, but the new smoke cleanup can delete pre-existing developer artifacts that this run does not own.

**Risk level**: A
- Path baseline: includes `apps/cli/**` → A
- Semantic lift: none

**PR summary**
- Problem: Staging and production npm publishing can produce pnpm traversal entries that the registry rejects, preventing users from receiving an installable CLI containing the patched Kimi SDK.
- Approach: Materialize the bundled SDK dependency closure as real directories during pack, restore pnpm links from a failure-atomic journal, and reject non-canonical tar paths before consumer-install smoke checks.
- Impacted modules: `apps/cli` package lifecycle and package metadata tests; release-pack smoke and tarball validation scripts.

**Review findings**
❌ 1. `cleanupPackArtifacts()` removes every `first-tree-dev-*.tgz` currently present in `apps/cli`, not only the tarball created by this smoke run. Consequently, a developer's pre-existing pack artifacts are deleted on success, and even invoking the script with missing `dist` or an invalid action reaches the top-level catch and deletes them. Snapshot the pre-existing set and preserve it, or track the exact new path produced by this invocation and clean only work-owned artifacts; add a sentinel regression proving an older tarball survives both success and failure cleanup. [R4 / `scripts/release-pack-smoke.mjs:72`]
⚠️ 2. Hardlink (`typeFlag === "1"`) targets are interpreted in the archive namespace, while symlink targets are resolved relative to the link entry's directory. Passing both through `resolveTarLinkWithinPackage()` can accept a hardlink such as `../outside` from a nested entry even though its archive-root target escapes `package/`. Validate hardlink targets directly as canonical `package/...` paths, and retain directory-relative resolution only for symlinks. [`scripts/npm-tarball-safety.mjs:153`]

**Resolved from the previous review**
- The complete restore journal is now atomically persisted before the first mutation, prepare failures roll back, and the injected failure proves recovery after a successful real-directory copy.
- Rootless and other non-canonical package member paths are now rejected.

**Action taken**
- Submitted request changes on head `585bc0fcf9ccad42f04f3b6296ce225b759d2309`.

baixiaohang (dismissed):
- Rationale: The previous findings are addressed, but path-only tarball ownership still overwrites and loses a pre-existing artifact when it has npm pack's deterministic output name.

**Risk level**: A
- Path baseline: includes `apps/cli/**` → A
- Semantic lift: none

**PR summary**
- Problem: Staging and production npm publishing can produce pnpm traversal entries that the registry rejects, preventing users from receiving an installable CLI containing the patched Kimi SDK.
- Approach: Materialize the bundled SDK dependency closure from a failure-atomic journal, validate canonical tar members and link targets, and smoke the packed artifact while preserving developer-owned outputs.
- Impacted modules: `apps/cli` package lifecycle/tests; release-pack smoke and tarball validation scripts.

**Review findings**
❌ 1. Ownership is tracked only by pathname. `npm pack` uses the deterministic current name/version filename, so if a developer already has the normal `first-tree-dev-<current-version>.tgz`, this invocation overwrites that exact path. `adoptNewTarballs()` then refuses to own it because the path was in `preexisting`, `runSmoke()` fails with zero new tarballs, and cleanup leaves the newly overwritten bytes in place—the original artifact is lost. The sentinel regression uses a different filename, so it cannot catch this collision. Pack directly into the run-owned temp directory (for example via npm's pack-destination output boundary) or otherwise preserve/restore an exact-name pre-existing file, and regress with the actual deterministic output name plus content/digest verification. [R4 / `scripts/release-pack-smoke.mjs:91`]

**Resolved from the previous review**
- Cleanup no longer deletes unrelated pre-existing tarball paths.
- Hardlink targets are now validated as canonical archive-root `package/...` paths, separately from directory-relative symlink resolution.

**Action taken**
- Submitted request changes on head `ba54194e7d38f4338d87f6fd2df1fcfadc91547a`.

PR #2130, opened 2026-08-01
feat: add Agent Template lifecycle surfaces

yuezengwu:
## Summary

- Let signed-in Team users choose zero to three public Agent Templates while creating an Agent, preview only public-safe responsibility fields, and pass a submission-time Template snapshot through the existing atomic adoption path.
- Open the newly created Agent in its first Workspace draft without sending a message, stamping onboarding, or implying activation.
- Add strict, fail-closed adopted-Template summaries and per-resource import provenance to Agent resources reads without changing runtime config materialization.
- Show Agent responsibilities on Agent Detail, including active/retired/missing states and safe replacement guidance; allow active managers and org admins to replace the set with version-CAS semantics while members and suspended-Agent views remain read-only.
- Preserve Team Resources, manual bindings, and other Agents' bindings when responsibilities are removed.

## Boundaries

- Template adoption remains copy-on-adopt; Team Resources and Agent bindings are execution authority after creation.
- No database migration, Client change, Template runtime lookup, catalog auto-sync, or applied-config-version telemetry.
- No public Template Library, cross-auth/onboarding intent handoff, or onboarding behavior; those remain in the follow-up PR.
- Creating an Agent and opening a draft do not claim successful activation.

## Testing

- Focused lifecycle suites: 210/210.
- `pnpm check` passed with 16 existing warnings.
- No-cache typecheck: 11/11; build: 5/5.
- Exact-Git, non-root Node 22 + PostgreSQL, single-fork full gate: 12/12 tasks; 9,825 passed / 4 skipped tests.
- Independent exact-head QA covered real 0/1/3-Template creation, PostgreSQL-backed submission snapshot behavior, atomic import/rollback/reuse, Detail state/role/CAS/removal behavior, provenance safety, catalog failure, and ordinary creation regression.
- An earlier UI run that used the wrong Chrome-extension path was discarded in full. Formal UI qualification was rerun from an empty, run-owned isolated headless `agent-browser` profile with scenario-level DOM/network/database evidence and a complete cleanup audit.


baixiaohang (dismissed):
- Rationale: The lifecycle UI loses actionable Template-adoption failures, and the API still permits responsibility changes on suspended Agents despite the read-only contract.

**Risk level**: B-high
- Path baseline: `packages/server/**` without a listed critical service -> B-low
- Semantic lift: shared Agent-resource state plus the Web data layer changes Agent lifecycle behavior -> B-high

**PR summary**
- Problem: Team users need to start an Agent from up to three curated responsibilities, understand those responsibilities later, and safely replace them without turning Templates into runtime authority.
- Approach: The create dialog adopts a submit-time Template set and opens a pre-addressed Workspace draft; Agent Detail adds public-safe responsibility summaries, version-CAS replacement, and historical import-origin labels while keeping Team Resources authoritative.
- Impacted modules: Server Template/resource projection, shared schemas, New Agent dialog, Team navigation, and Agent Detail responsibility/resource surfaces.

**Review findings**
❌ 1. Template-adoption conflicts are misclassified as unrelated recovery paths. [R4 / `packages/web/src/components/new-agent-dialog.tsx:624`; `packages/web/src/pages/agent-detail/responsibilities-section.tsx:183`]

The create transaction can return HTTP 409 for an inactive or invalid Template, a Team Skill/MCP name collision, or an existing manual resource choice. The create dialog currently converts every 409 into "agent name already in use." Agent Detail similarly converts every 409 into a version-CAS conflict, refreshes the unchanged set, and asks the user to retry. In both cases a retry cannot resolve the actual conflict and the precise server error is discarded. Please reserve the refresh path for a real version mismatch and surface other adoption conflicts at the Template/responsibility level; cover at least one non-CAS 409 on each surface.

❌ 2. Suspended Agents remain mutable through the Template PATCH despite the read-only lifecycle contract. [R4 / `packages/server/src/api/agents-resources.ts:27`; `packages/server/src/services/agent-template-adoption.ts:441`]

The UI hides the editor when `agent.status !== "active"`, but `requireAgentAccess(..., "manage")` accepts a suspended Agent and the locked service check rejects only deleted/human/trial/runtime-switch cases. A direct `PATCH /agents/:uuid/templates` can therefore remove/import bindings and bump config version while the Agent is suspended. Enforce the active-status guard in the locked transaction (with a route fast path if useful) and add a route-level suspended-Agent regression test.

**Action taken**
- Submitted request changes.

yuezengwu:
@baixiaohang The two blocking findings from your review are addressed in exact head `ced514f56802da6f49149a4878cf24b77a1dba49` (follow-up commit `ced514f`, parent `260b46a`).

- 409 responses now carry stable wire codes for Agent-name, Template-adoption, version-CAS, and inactive-Agent conflicts. New Agent and Agent Detail branch only on the matching code; non-CAS adoption conflicts keep the current draft and surface the server message in Responsibilities. Late responses are tied to the submit-time Template snapshot.
- Template replacement now rejects every non-active Agent inside the locked transaction. The suspended-Agent route regression exercises that path and verifies config, bindings, and Team Resources remain exactly unchanged.

Focused regression suites pass (Shared 55/55, Server 48/48, Web 26/26). Independent QA on this exact head also passed the real duplicate-name/adoption/CAS paths, pending-response race, suspended same-set and empty-set direct PATCH invariance, and ordinary lifecycle regressions. Forced no-cache typecheck/build/full test passed (12/12 tasks; 9,831 passed, 4 skipped), and all current GitHub checks are green.

Please re-review the exact current head when convenient.

PR #2133, opened 2026-08-02
feat: add public Agent Template library and onboarding intent

yuezengwu:
## Summary

- Add public `/templates` Library and `/templates/:slug` detail routes over the existing public-safe catalog projection, including active, retired/replacement, empty, missing, loading, and recoverable error states.
- Preserve the byte-exact canonical use intent `/templates/:slug?use=1` through GitHub/Google login, while rejecting extra, duplicate, encoded, absolute, hash, backslash, trailing-separator, and trailing-slash variants.
- Resolve authenticated intent against an explicitly chosen Team. A switch proceeds only after a fresh authoritative `/me` snapshot confirms the exact membership; transport failure rolls back to the last confirmed Team, final 401/logout cannot resurrect old org state, and persistent ChatById failures do not create automatic retry loops.
- Carry a per-Team Template handoff through onboarding or preselect the existing New Agent dialog for established accounts. Creation passes the submit-time Template id through the existing atomic Agent POST; stale/retired/fetch-failed intents degrade to plain creation without submitting stale ids.
- Add WCAG-AA semantic foreground tokens for brand-filled actions and instrument Library/detail/use-started/create-success events without exposing private Template content.

## Boundaries

- Builds on the Agent Template lifecycle shipped in #2130. Template use remains a one-time import into the selected Team; imported Team Resources and Agent bindings are the execution authority afterward.
- Public responses render only safe purpose, audience, value, and capability summaries. Prompt source, Skill attachment identity/content, MCP connection details, drafts, and malformed definitions stay private or fail closed.
- No database migration, Client materialization change, production catalog content, hidden prompt, automatic chat/message, second adoption request, catalog auto-sync, or activation claim.
- Agent creation still opens `/?c=draft&with=<uuid>`; a real first work loop remains a separate post-merge staging pilot.

## Validation

- [x] `pnpm check` — passed with 16 baseline warnings and no errors.
- [x] `pnpm typecheck` — forced no-cache, 11/11 tasks.
- [x] `pnpm test` — forced single-fork full monorepo run, 12/12 tasks; 912 files passed / 1 skipped, 10,105 tests passed / 4 skipped.
- [x] Production build — forced no-cache, 5/5 tasks.
- [x] Independent exact-head QA on `8270c4ac47bb63aee5592853feb3aa979dfb3d07` using an isolated production Server/Web/PostgreSQL/browser cell.

Independent QA exercised anonymous discovery and leakage checks, strict login intent, exact-Team selection, deterministic `/me` abort/retry and final-401 logout, ChatById retry bounds, per-org onboarding handoff, real Template and zero-Template Agent POSTs with API/DB readback, stale/failure recovery, and WCAG 2 AA across the new public controls. Brand CTA contrast measured 8.25:1 normal/focus and 5.58:1 hover.

## Change Surface

## Notes

- Package or install behavior changes: none.
- Docs or tests updated to match: focused Shared/Server/Web contracts plus real-browser QA; no committed QA case changed in this PR.
- Limitations: real third-party GitHub/Google account callback completion was unavailable; exact local OAuth start/state/next boundaries were validated. Creation and draft navigation intentionally do not simulate or claim organic runtime activation.
- Follow-up work: after merge, run the final staging pilot through anonymous discovery → auth/onboarding → creation → a real first task → Agent Detail/removal, plus ordinary-creation regression, before adding production catalog content.

### DIFF
--- a/packages/server/src/__tests__/oauth-bootstrap.test.ts
+++ b/packages/server/src/__tests__/oauth-bootstrap.test.ts
@@ -8,7 +8,7 @@ import { authIdentities } from "../db/schema/auth-identities.js";
 import { members } from "../db/schema/members.js";
 import { organizations } from "../db/schema/organizations.js";
 import { findOrCreateUserFromExternalAccount } from "../services/auth-identity.js";
-import { completeExternalAccountBootstrap } from "../services/oauth-bootstrap.js";
+import { completeExternalAccountBootstrap, shouldPreserveSoloSignupNext } from "../services/oauth-bootstrap.js";
 import { uuidv7 } from "../uuid.js";
 import { useTestApp } from "./helpers.js";
 
@@ -103,6 +103,24 @@ describe("provider-neutral OAuth bootstrap", () => {
     });
 
     expect(result).toMatchObject({ joinPath: "solo", next: "/", teamCreated: true });
+  });
+
+  it("preserves a strict Agent Template intent across solo signup", async () => {
+    const app = getApp();
+    const account = await findOrCreateUserFromExternalAccount(
+      app.db,
+      googleExternalProfile({ sub: "google-template-intent-subject", name: "Template Intent User" }),
+    );
+    const next = "/templates/pr-engineer?use=1";
+
+    const result = await completeExternalAccountBootstrap(app.db, account, {
+      next,
+      allowedOrganizationId: null,
+      ip: null,
+      userAgent: null,
+    });
+
+    expect(result).toMatchObject({ joinPath: "solo", next, teamCreated: true });
   });
 
   it("serializes concurrent first sign-ins into one personal team graph", async () => {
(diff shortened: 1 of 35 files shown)


baixiaohang (changes requested):
- Rationale: Two asynchronous state races can either discard an explicit Template adoption or let a stale `/me` response overwrite the current session.

**Risk level**: B-low
- Path baseline: includes `packages/server/**`, but the touched service is `oauth-bootstrap` rather than a B-high core service.
- Semantic lift: the Web auth/data layer and cross-Team selection logic independently lift the Web portion to B-low; no A/B-high path is touched.

**PR summary**
- Problem: Let anonymous visitors discover official Agent Templates, preserve a chosen Template through sign-in, and create a Template-backed Agent in an explicitly confirmed Team or through that Team's onboarding flow.
- Approach: Adds public Library/detail routes, a strict canonical OAuth intent, authoritative Team confirmation, per-Team onboarding handoff, Template preselection in both creation surfaces, safe analytics, and AA foreground tokens.
- Impacted modules: `packages/shared`, Server OAuth bootstrap, Web auth/Team selection, public Template pages, onboarding, New Agent creation, workspace chat routing, and design tokens.

**Review findings**
❌ 1. An explicit “Use this template” intent can silently become a plain Agent while its Template lookup is still pending. `NewAgentDialog` enables submit without waiting for `templateCatalogLoaded`, and `StepCreateAgent` enables create while `intentSlug` exists but `intentTemplate` has not resolved. In either surface, a slow/hung request lets the user submit `templateIds: []`; onboarding then clears the handoff after creation, so the requested responsibility is permanently lost without the user removing it or seeing an unavailable state. Keep creation disabled (with a loading state) while an explicit intent is unresolved, and permit plain creation only after a definitive unavailable/failure result or an explicit Remove. [R4 / `packages/web/src/components/new-agent-dialog.tsx:741`, `packages/web/src/pages/onboarding/steps/step-create-agent.tsx:123`]

❌ 2. Successful `/me` responses are not guarded against a logout or identity change that occurs while the request is in flight. `selectOrganization` records `sessionMarker`, but checks it only in the rejection path; a request started as user A can still return successfully after logout or after user B adopts tokens, and `loadMe` will repopulate `user`, `memberships`, `confirmedOrgIdRef`, the persisted selection, and the API org target from A. The new 401 test cannot cover this because it exercises only rejection. Capture the session/generation at request start and discard a successful response when it no longer matches before applying any state. [R4 / `packages/web/src/auth/auth-context.tsx:293`]

✅ The strict shared intent parser, public-safe Template DTO boundary, exact-Team membership check, and no-auto-retry ChatById guard are otherwise aligned with the documented cross-surface constraints.

**Action taken**
- Submitted request changes.

yuezengwu:
Superseded by the CLA-valid replacement PR: https://github.com/agent-team-foundation/first-tree/pull/2134

The repaired source from this PR was transferred onto current `main` as one `yuezengwu`-authored commit, without amending or force-pushing this branch. The aggregate patch-id remains exactly `4f39a40b0bc7bbb3bee3be25ea6c11e063419b05`; old repaired head `c881569b32d13739f7f88a17a66da1aba6303d6b` and replacement head `b0744b78232f80efdaab2a5d41ef3f0fe91faae7` are source-equivalent across the candidate paths.

The replacement includes both requested review fixes (fail-closed unresolved Template intent and stale successful `/me` session isolation) plus their regression coverage. Closing this PR so CI, review, CLA, and fresh exact-head QA continue on the replacement only.

PR #2141, opened 2026-08-03
test: add Momentic E2E tests for registration and onboarding

liuchao-001:
Adds browser-level E2E coverage for the first-run journey, driven by Momentic against a real server and database. This complements the per-package Vitest suites rather than replacing them — deterministic behaviour still belongs in product tests.

## What it covers

| Test | Env | Covers |
| --- | --- | --- |
| `registration-new-user` | local | A brand-new account is created and lands on onboarding step 1 |
| `onboarding-complete-setup` | local | Sign up → create team → connect a computer → create the first agent → start the kickoff chat → land in the workspace (`/?c=`) |
| `dev-cloud-sign-in-available` | dev-cloud | Staging serves the landing page and offers both identity providers |

## Registration has no password path

First Tree accounts are created through GitHub / Google OAuth, so there is no form to fill. The tests sign up through `auth/github/dev-callback` — the product's own localhost-only stub for that exchange, enabled by `FIRST_TREE_DEV_CALLBACK_ENABLED=1` in the server `dev` script. It is the real registration path with the github.com round trip removed.

## Two fixtures, and what they stand in for

Onboarding steps 2 and 3 wait on a *second machine* a browser test does not have: the First Tree client daemon plus an installed runtime. `e2e/scripts/seed-*.js` seed the rows that daemon would write.

The fixtures replace the machine, not the behaviour under test — every screen, transition, API call and assertion around them runs for real. What they do **not** cover is the daemon's own handshake and heartbeat; that belongs to client/runtime tests.

Both date `last_seen_at` forward on purpose: the server sweeps clients and agents whose heartbeat is older than `presenceCleanupSeconds` (60s default, `cleanupStaleClients`). With a plain `NOW()` the row decays mid-run and the flow regresses to "Your computer isn't connected".

## Why the first two tests are local-only

They cannot run against a deployed environment, by design rather than omission:

- `dev-callback` is gated on `NODE_ENV !== "production"` **and** an explicit opt-in, so it 404s on `dev.cloud.first-tree.ai`. Automating sign-up there would mean driving a real identity provider with real credentials.
- The fixtures write directly to PostgreSQL, which a deployed environment does not (and should not) expose.

Staging therefore gets what it can support without credentials. That still catches a broken deploy, a broken bundle, or a misconfigured auth surface.

## Notes for review

- The tests are **not idempotent by design** — each run registers a new `e2e-user-<ms>`, because reusing an identity would sign in as the previous run's user instead of registering. They accumulate rows in the local dev database.
- `e2e/scripts` is excluded from Biome: Momentic step scripts are sandbox fragments (top-level `return`/`await`, injected globals), not standalone modules, so Biome cannot parse them.
- Not wired into CI yet. The staging test could run today with no secrets and no local stack; the local tests would need postgres + server + web in the pipeline.

## Verification

`pnpm check` (0 errors) · `pnpm typecheck` (11/11) · `npx momentic lint` · full suite **3 passed**, re-verified after rebasing onto latest main.

Run results: https://app.momentic.ai/run-groups/f9dc2966-fce6-4b71-8e03-3f9d26bd2d0e

Setup and fixture rationale are documented in `e2e/README.md`.


yuezengwu (changes requested):
Implementation goal: add Momentic browser-level coverage for new-user registration, the admin onboarding journey, and a deployed staging sign-in smoke check. The core changes add the Momentic config/dependency, a reusable dev-callback sign-up module, and direct-PostgreSQL fixtures for the machine/runtime boundaries.

Blocking issue:

- `e2e/scripts/seed-agent-online.js:28` inserts `agent_presence.runtime_state = 'ready'`. That value is outside the product contract: `runtimeStateSchema` permits only `idle | working | blocked | error`, and the real daemon bind path writes `idle` (`packages/shared/src/schemas/presence.ts:33`, `packages/server/src/services/presence.ts:90`). Because this fixture bypasses the server schema with raw SQL, the test currently advances with a state the real daemon can never produce. That can hide runtime-dependent regressions and makes the fixture's claim that it stands in for the daemon inaccurate. Please seed `idle` instead (and mirror `runtime_updated_at` / the conflict update if the fixture is meant to reproduce the bind row faithfully).

There is no schema migration or production data-structure change in this PR. Per reviewer instructions I reviewed the diff statically and did not rerun tests or QA; this review does not wait on CI.

baixiaohang (changes requested):
- Rationale: the proposed coverage is valuable, but the PR currently introduces an unapproved third QA layer and changes existing product dependency resolutions outside the E2E scope.

**Risk level**: B-low
- Path baseline: root test/config/docs only -> C
- Semantic lift: the new root dependency rewrites dependency resolution for CLI, Client, Server, and Web -> B-low

**PR summary**
- Problem: give maintainers repeatable browser evidence that a new member can register, finish first-run setup, and reach the first chat, plus a deployed-staging auth-surface smoke check.
- Approach: add Momentic natural-language browser scenarios, use the localhost-only OAuth dev callback for fresh identities, and seed the missing daemon/runtime state directly in PostgreSQL while leaving the surrounding Web and Server flow real.
- Impacted modules: root dependency/tooling config, `e2e/**`, and the shared pnpm lockfile.

**Review findings**
❌ 1. Adding a test-only root dependency changes existing runtime/tooling resolutions for unrelated packages. The lockfile moves the CLI and Client from `@modelcontextprotocol/sdk@1.29.0` to `1.30.0`, changes Client Sentry/OpenTelemetry peers, changes the Server Logfire/OpenTelemetry peer graph, and changes Vitest/Vite YAML peers, even though none of those package manifests changed. That makes this E2E PR also an unverified product dependency upgrade. Please preserve the pre-existing importer resolutions or isolate and separately justify/verify the unavoidable upgrades. [R5 / `pnpm-lock.yaml:39`, `pnpm-lock.yaml:130`, `pnpm-lock.yaml:139`, `pnpm-lock.yaml:232`]

❌ 2. This creates a third QA contract without reconciling the repository's current split: deterministic code behavior belongs in package product tests, while live/cross-surface or judgment-dependent validation belongs in committed `packages/qa/cases/`. These scenarios use AI locators/assertions, cloud-backed memory, and beta failure recovery, so they are not deterministic product tests; in CI, recovery may generate and execute extra steps after a failure. Either record an explicit adoption decision and update the repo QA contract, including how recovered runs are classified, or keep the live journey in the existing QA-case layer and put stable invariants in Vitest. [R3 / `momentic.config.yaml:14`, `e2e/onboarding-complete-setup.test.yaml:1`]

⚠️ 3. The setup instructions are not reproducible as written. They say staging needs no credentials and list only the First Tree stack, but Momentic's own setup requires a Momentic account/authentication and a browser install; its GitHub Actions guide requires `MOMENTIC_API_KEY`. Please document the external account/credential, browser installation, data-upload boundary, and correct the “without credentials” / future-CI claim. [R4 / `e2e/README.md:26`, `e2e/README.md:30`]

✅ The dev callback remains protected by the existing non-production + explicit opt-in gates, and the two direct-DB fixtures clearly state which daemon/runtime behavior they do not cover.

**Action taken**
- Submitted request changes.

liuchao-001:
Thanks — all four findings were valid and are fixed in 70b5e1d. Verified each against the code before changing anything.

**@yuezengwu — `runtime_state = 'ready'` (blocking)**

Confirmed and fixed. `runtimeStateSchema` allows only `idle | working | blocked | error`, and `publishAgentPresence` writes `idle`. I had taken `ready` from the onboarding UI's "· ready" label, which is not the persisted value — so the fixture was advancing the test through a state no daemon can produce. It now writes `idle` and `runtime_updated_at`, and the conflict branch updates `runtime_type` / `runtime_state` / `runtime_updated_at` to mirror the real bind row. Re-ran: the row is now `idle` with `runtime_updated_at` set, no `ready` anywhere.

**@baixiaohang 1 — lockfile perturbs product resolutions (blocking)**

Confirmed. The diff did move CLI and Client from `@modelcontextprotocol/sdk` 1.29.0 to 1.30.0 (via the `claude-agent-sdk` peer) with no product manifest change. The bump is unavoidable while the runner shares a lockfile with the product graph, so rather than justify an upgrade this PR has no business making, I dropped the root devDependency entirely. `package.json` and `pnpm-lock.yaml` are now byte-identical to main, and the CLI is invoked through a pinned `npx momentic@3.42.0`. That also takes ~3.8k lines out of the diff.

**@baixiaohang 2 — third QA contract (blocking)**

Fair. Two changes:

- `AGENTS.md` now records the layer: what `e2e/` is for (flows spanning Web + Server + DB with no single package to live in), that it is executable but **not deterministic**, that it is explicitly **not a CI gate** — a red run is a signal to investigate, not a merge blocker — and that promoting it is a separate decision.
- On the recovery concern specifically: `useMemory` and `failureRecovery` are now `false`, so a run only ever executes the steps committed in the directory. The suite passes without either, so they were not load-bearing.

I have not tried to settle the `packages/qa` boundary or a gate policy in this PR — that needs an owner for the account and a classification policy, which the README now names as open.

**@baixiaohang 3 — setup not reproducible (warning)**

Correct, and my wording was misleading. "Without credentials" meant *First Tree* credentials; Momentic is a hosted product needing its own account. The README now leads with `momentic login` + `install-browsers`, states CI would need `MOMENTIC_API_KEY`, and scopes the staging claim to "without a First Tree account".

I also added a data-boundary section you did not ask for but which seemed material: step resolution and model-backed assertions send DOM snapshots and screenshots to Momentic's service, and `results upload` publishes screenshots/video. Documented so nobody points this suite at an environment with real user data.

**Verification after the changes:** `pnpm check` (0 errors) · `pnpm typecheck` (11/11) · `momentic lint` · full suite 3 passed with memory and recovery disabled and the dependency removed.

yuezengwu (changes requested):
The runtime fixture, dependency isolation, and setup/data-boundary findings are fixed on this head:

- `runtime_state` now mirrors the real bind contract (`idle` plus `runtime_updated_at`, including the conflict branch).
- `package.json` and `pnpm-lock.yaml` are identical to `main`.
- The README now documents Momentic authentication, browser installation, CI credentials, and the DOM/screenshot upload boundary; memory and beta recovery are disabled.

One blocker remains:

- `AGENTS.md:101` and `e2e/README.md:7-29` still introduce an explicitly non-deterministic fourth QA layer while also saying its boundary with `packages/qa` is unresolved. That conflicts with the current repository instruction immediately above it and the durable Context Tree decision in `team-practice/quality-assurance.md`: judgment/live/cross-surface validation belongs in committed `packages/qa/cases/`. Making the suite “not a CI gate” changes how failures are consumed, but it does not authorize a new quality layer or resolve its ownership/classification boundary. The README itself confirms that adoption policy is still open.

Please either keep this journey in the existing QA-case layer, or first land an owner-approved Context Tree decision that adopts the new executable layer and defines its boundary; then reconcile `AGENTS.md` and the suite to that durable contract. A source-repo instruction cannot override the current Tree decision while declaring the decision unsettled.

There is still no database migration or production data-structure change. Per reviewer instructions this is a static diff review; I did not rerun tests or QA and did not wait on CI.

baixiaohang (changes requested):
- Rationale: The runtime-fixture, dependency-isolation, and setup/data-boundary findings are resolved, but this head still establishes a new QA layer while explicitly leaving its ownership boundary undecided.

**Risk level**: C
- Path baseline: root instructions/configuration and `e2e/**` only -> C
- Semantic lift: none

**PR summary**
- Problem: give maintainers repeatable browser evidence that a new member can register, complete first-run setup, and reach the first chat, plus a deployed-staging auth-surface smoke check.
- Approach: run Momentic browser journeys against the real Web/Server/DB flow, with narrowly scoped fixtures for the unavailable daemon/runtime boundaries.
- Impacted modules: repository QA instructions, Momentic configuration, and `e2e/**`.

**Review findings**
❌ 1. The QA-layer blocker remains. `AGENTS.md:100` says judgment-dependent, live, and cross-surface validation belongs in committed `packages/qa/cases/`; the immediately following new rule creates a separate non-deterministic `e2e/` layer, while `e2e/README.md:7-29` calls it a “fourth thing” and says its boundary with `packages/qa` is still open. Declaring the suite non-gating changes how failures are consumed, but it does not resolve which layer owns these journeys. Please either keep them in the existing QA-case layer, or first land the owner-approved adoption decision and then align the repository instructions and suite with that contract. [R3 / `AGENTS.md:100-101`, `e2e/README.md:7-29`]

✅ The earlier invalid runtime state, product lockfile perturbation, external authentication/browser prerequisites, data-upload boundary, memory, and failure-recovery findings are fixed on this head.

**Action taken**
- Submitted request changes on the current head. This was a static follow-up review; I did not run tests or QA and did not wait on CI.

liuchao-001:
@yuezengwu — fixed in 658544b. You were right that the previous head asserted a contract it had no authority to grant, and doubly so while calling the boundary unsettled.

**Taken option (a): the journey now lives in the existing QA-case layer.**

Added `packages/qa/cases/cross-surface/registration-first-run-onboarding.md`. It owns the validation question, including the parts a green happy path cannot demonstrate:

- connect-computer **stays blocked** while no client is connected — a gate left open by accident would be invisible in a passing run;
- a stale heartbeat must honestly regress the step rather than stay green;
- the "taking longer than usual" branch must be reachable and offer both exits;
- a seeded row must mirror the real write (`runtime_state = idle`), or the journey advances through a state no daemon can reach.

It also scopes itself against `member-work-mode-onboarding`, which covers invited members — this case is the first user, who becomes the admin.

`AGENTS.md` no longer declares a layer. `e2e/` is now described as optional tooling for executing part of that case, in the same spirit as the fixtures and environment recipes already under `packages/qa` — not a layer, not an authority on what "validated" means, still not a gate. I did not attempt to settle a gate policy or adopt an executable layer here; that remains an owner decision.

**One real bug this surfaced.** Re-verifying after the restructure, the staging test failed 3/3 in full-suite runs while passing alone. Not flake: it crosses the internet, and the 8s default page load timeout only held when nothing else competed for the machine. In a suite run the local stack, Vite and the browser saturate the CPU and the first load times out — reported as "infrastructure failure", which says nothing about staging and would have been a confusing red for whoever hit it next. Raised to 30s with the reasoning in the file; three consecutive suite runs now pass, one of them taking 37s on that test.

Worth noting this is exactly the class of failure `failureRecovery` had been masking before it was disabled.

**Verification:** `pnpm check` (0 errors) · `@first-tree/qa` tests 4/4 (validates the new case) · `momentic lint` · full suite 3 passed × 3 consecutive runs.

baixiaohang (changes requested):
- Rationale: The QA-layer ownership blocker is resolved, but the new authoritative QA case assigns organization creation to the wrong onboarding transition.

**Risk level**: C
- Path baseline: repository QA instructions, `e2e/**`, and a committed QA case only -> C
- Semantic lift: none

**PR summary**
- Problem: give maintainers reusable evidence that a brand-new user can register, complete the admin first-run journey, and reach a working first chat.
- Approach: keep the journey's validation contract in `packages/qa/cases/`, with Momentic as optional browser tooling for a narrower unattended pass.
- Impacted modules: repository QA instructions, the registration/onboarding QA case, and `e2e/**`.

**Review findings**
❌ 1. The new case attributes organization creation to two different transitions. Scenario step 1 correctly says OAuth registration creates the user, organization, human agent, and prefilled team context; step 2 then says that continuing past the team screen creates the organization and admin membership. In the product, the organization and active admin membership already exist from OAuth bootstrap, and `StepTeam` only loads that row and optionally PATCHes its display name; leaving the prefilled name unchanged performs no write at all. Please make step 2 verify the existing org/admin membership and any requested rename instead of requiring creation after Continue. Otherwise a correct run with the unchanged prefilled name can be falsely failed, and the case records the wrong state boundary. [R1 / `packages/qa/cases/cross-surface/registration-first-run-onboarding.md:48-52`]

✅ The previous QA-layer blocker is resolved: the committed case is now the contract, while `e2e/` is explicitly optional, non-gating execution aid with narrower coverage.

**Action taken**
- Submitted request changes on the current head. This was a static follow-up review; I did not run tests or QA and did not wait on CI.

yuezengwu (changes requested):
The QA-layer ownership blocker is resolved on this head: the committed `packages/qa` case is now the validation contract, and `e2e/` is clearly scoped as optional, non-gating execution aid with narrower coverage. The 30-second staging page-load override is also a supported per-test setting and does not weaken the assertions.

One blocker remains in the new authoritative case:

- `packages/qa/cases/cross-surface/registration-first-run-onboarding.md:48-52` assigns organization creation to two transitions. Step 1 correctly says first-sign-in OAuth bootstrap creates the user, organization, human agent, and prefilled team context. Step 2 then requires Continue to create the organization and admin membership. In the product, those rows already exist before onboarding renders: `StepTeam` loads the existing organization and only PATCHes its display name when the user changes it; leaving the prefilled name unchanged performs no write. A correct run can therefore fail this case, and the case records the wrong state boundary.

Please make step 2 verify that the existing organization/admin membership remains correct and that an edited name is persisted (while an unchanged name advances without creating anything).

There is no database migration or production data-structure change. Per reviewer instructions this is a static diff review; I did not rerun tests or QA and did not wait on CI.

liuchao-001:
@baixiaohang — correct, and verified before changing. Fixed in 8c3afc1.

`StepTeam`'s own comment says it: *"The team row already exists (auto-created at sign-in) … an unchanged name skips the PATCH entirely."* `handleSubmit` only PATCHes when `trimmed !== initialName.trim()`; otherwise it just calls `goNext()`.

I also checked it empirically rather than trusting the comment — called the dev callback and read the database without touching the UI at all:

```
     username      | org_exists |       display_name       | role  | status
-------------------+------------+--------------------------+-------+--------
 verify-1785829155 | t          | verify-1785829155's team | admin | active
```

So the org, the active `admin` membership and the prefilled display name all exist before the team screen is ever submitted. As written my case would have failed a correct run that accepted the prefilled name, which is worse than not covering it.

Step 2 now verifies the pre-existing org and membership, states explicitly that an unchanged name performs no write, and asks for the rename path (`PATCH /orgs/:orgId`, display name only) to be checked as a separate branch. The same mislabel in the browser test's step comment is fixed.

**Verification:** `@first-tree/qa` tests 4/4 · `momentic lint` · full suite 3 passed.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The authoritative QA case now assigns organization creation to OAuth bootstrap and accurately scopes the team step to confirmation plus an optional display-name update; all earlier blockers are resolved.

**Risk level**: C
- Path baseline: repository QA instructions, `e2e/**`, and a committed QA case only -> C
- Semantic lift: none

**PR summary**
- Problem: give maintainers reusable browser-level evidence that a brand-new user can register, complete the admin first-run journey, and reach a working first chat.
- Approach: keep the cross-surface validation contract in `packages/qa/cases/`, with pinned Momentic scenarios as optional, non-gating execution aid for a narrower unattended pass.
- Impacted modules: repository QA instructions, the registration/onboarding QA case, and `e2e/**`.

**Review findings**
✅ 1. The last blocker is fixed: the case now verifies that OAuth bootstrap created the organization and active admin membership, that accepting the prefilled team name performs no write, and that an edited name changes only the display name. The Momentic scenario comment mirrors the same boundary. [`packages/qa/cases/cross-surface/registration-first-run-onboarding.md:49`, `e2e/onboarding-complete-setup.test.yaml:10`]

✅ 2. The committed QA case is the validation authority, while `e2e/` is explicitly optional, non-gating tooling with narrower coverage. The invalid runtime fixture state, lockfile perturbation, setup prerequisites, data boundary, memory, and failure-recovery findings remain resolved.

**Action taken**
- Approved the current head. This was a static follow-up review; I did not run tests or QA and did not wait on CI.

yuezengwu (approved):
Implementation goal: add reusable browser-level evidence that a brand-new user can register, complete the admin first-run journey, and reach a working first chat, while keeping the authoritative cross-surface validation contract in `packages/qa/cases/` and using Momentic only as an optional, non-gating execution aid.

The last blocker is resolved on `8c3afc1`: the QA case now correctly assigns organization and active admin-membership creation to OAuth bootstrap. The team step verifies the existing organization, records that accepting the prefilled name performs no write, and treats an edited display name as a separate `PATCH /orgs/:orgId` branch that must preserve organization identity and membership. The Momentic scenario comment now mirrors that boundary.

The earlier findings remain resolved: the committed QA case owns the journey; the presence fixture mirrors the daemon's supported `idle` state and timestamp fields; product package and lockfile resolutions are unchanged; Momentic prerequisites and the DOM/screenshot upload boundary are documented; and memory plus failure recovery remain disabled.

There is no database migration or production data-structure change. This was a static follow-up review; I did not run tests or QA and did not wait on CI.

PR #2142, opened 2026-08-03
fix: keep working turn elapsed time monotonic

baixiaohang:
## Summary

- compute a stable active-turn start timestamp across the full chat-speaker event scope before applying the capped event window
- return `turnStartedAt` with each chat session-event feed and use it as the WorkingTurn elapsed-time origin
- keep Web compatible with older Server deployments by falling back to the first visible event
- add regressions for a server event window that has already dropped the turn's first event and for a Web rerender that drops the oldest row

## Testing

- `pnpm --filter @first-tree/server exec vitest run src/__tests__/session-event.test.ts`
- `pnpm --filter @first-tree/web exec vitest run src/components/chat/__tests__/working-turn-dom.test.tsx`
- `pnpm check`

### DIFF
--- a/packages/server/src/__tests__/session-event.test.ts
+++ b/packages/server/src/__tests__/session-event.test.ts
@@ -318,12 +318,14 @@ describe("sessionEventService", () => {
       { chatId: c, agentId: viewer.humanAgentUuid, accessMode: "speaker" },
     ]);
 
+    const firstEventAtByAgent = new Map<string, string>();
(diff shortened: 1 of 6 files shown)


yuezengwu (approved):
Goal and core changes: the Server now derives the active-turn start across the full per-speaker event scope before applying the capped feed window, exposes it as turnStartedAt, and Web uses that stable anchor with an older-Server fallback. I reviewed the turn-boundary/window query, per-agent scoping, rollout compatibility, component propagation, and the focused regressions. No blocking findings.

PR #2151, opened 2026-08-04
fix(cli): preserve portable install identity

### LINKED ISSUE #2145: [Bug] Portable/BYO migration can bind systemd to a legacy global CLI and corrupt version isolation
## Summary

Migrating a machine that still has a legacy global npm installation to the portable CLI can cause the user systemd service to bind to the legacy global binary. A subsequent managed update may then install npm packages inside the bundled Node tree of an older portable version, breaking portable version isolation.

The visible symptom is a split version:

- `first-tree-staging --version` through `$HOME/.local/bin/first-tree-staging`: `0.5.19-staging.1035.1`
- daemon/web-reported `sdkVersion`: `0.5.19-staging.1039.1`

This was observed immediately after running the server-provided BYO bootstrap on a machine that previously used the global npm install.

## Surface Area

- portable installer and portable self-update
- systemd user service generation/refresh
- migration from a legacy global npm install
- BYO bootstrap (`install.sh` followed by an absolute portable `login`)

## Environment

- version: staging `0.5.19-staging.1035.1` through `1041.1`
- install method: legacy global npm install, followed by portable `install.sh`
- operating system: Linux with a systemd user service
- legacy binary still present at `/usr/local/bin/first-tree-staging`
- portable shim at `$HOME/.local/bin/first-tree-staging`

## Reproduction

1. Start with a working global npm installation, with `/usr/local/bin` earlier than `$HOME/.local/bin` (or with `$HOME/.local/bin` absent) in the current agent/shell process's inherited `PATH`.
2. Run the BYO bootstrap:
   ```sh
   curl -fsSL <downloads>/staging/install.sh | sh
   ~/.local/bin/first-tree-staging login <code>
   ```
3. The installer correctly creates the portable `current` symlink and shim. Its `PATH` update, however, only affects the piped child shell.
4. The absolute portable `login` installs/refreshes the service. `resolveCliInvocation()` ignores the portable runtime identity and calls `whichBin(channelConfig.binName)` against the inherited parent `PATH`, so it can resolve and realpath the stale `/usr/local/bin/first-tree-staging`.
5. The service starts through that legacy global installation. Managed auto-update then takes the npm/global path. On the affected machine it modified the package under:
   ```
   $HOME/.local/share/first-tree/staging/versions/0.5.19-staging.1035.1/node/lib/node_modules/first-tree-staging
   ```
   while `versions/...1035.1/app` and the top-level `VERSION` remained `1035.1`.
6. The portable shim therefore reports `1035.1`, while the service loaded from the nested npm package reports the newer version (observed `1039.1`) to the server/web UI.

A related failure remains after an otherwise healthy portable update: the generated unit's `PATH` uses `dirname(process.execPath)`, which resolves to a concrete path such as:

```
$HOME/.local/share/first-tree/staging/versions/0.5.19-staging.1035.1/node/bin
```

instead of a stable portable path. It remains stale until `daemon refresh-unit` is run manually.

## Expected Behavior

- An invocation through the portable CLI must install/refresh a service that stays on the portable runtime, regardless of ambient `PATH` or a leftover global npm binary.
- Managed portable updates must refresh the unit by explicitly invoking the newly installed portable CLI/runtime.
- The systemd unit should use stable portable paths (for example `current/node/bin` and/or the portable shim directory), not a concrete version directory.
- A legacy global installation must never mutate files inside a portable version directory.

## Actual Behavior

There are three coupled path-resolution problems:

1. `resolveCliInvocation()` uses ambient `which <binName>` and `realpathSync()`, rather than preserving the fact that the current process was launched from a portable installation.
2. `refreshServiceUnit()` spawns the bare channel binary name:
   ```ts
   spawnSync(channelConfig.binName, ["daemon", "refresh-unit"], ...)
   ```
   The daemon's systemd `PATH` does not include `$HOME/.local/bin`; on the affected machine it resolves the stale global binary in `/usr/local/bin`.
3. `servicePathEnv()` prepends `dirname(process.execPath)`, pinning `PATH` to one portable version instead of `current/node/bin`.

The existing `update-glue` test mocks `spawnSync` and verifies the bare command name, so it does not cover real PATH resolution with coexisting global and portable installs.

## Evidence from the affected machine

Before manual repair:

- portable shim/current app: `0.5.19-staging.1035.1`
- package nested under that version's bundled Node: `0.5.19-staging.1039.1`
- web/daemon: `0.5.19-staging.1039.1`
- old global CLI: `0.5.18-staging.1004.1`

The daemon log immediately after the portable/BYO migration showed the legacy `1004.1` runtime starting, then running:

```
npm install -g first-tree-staging@0.5.19-staging.1035.1
```

A later update to `1041.1` made CLI and daemon versions agree, but the unit still contained the concrete `1035.1/node/bin` PATH until a manual:

```sh
first-tree-staging daemon refresh-unit
```

## Regression history / likely trigger

The underlying assumptions predate BYO:

- `7f1aab32c`: service refresh after managed self-update invokes the installed binary by bare name; this matched the original global npm model.
- `d48aa5418`: service invocation resolution uses `whichBin(channelConfig.binName)`.
- `f5375a85e`: service PATH begins with `dirname(process.execPath)`.
- `4ba518d1f`: portable self-update reused the existing global-oriented refresh path without making it portable-aware.

The recent BYO change in #2094 made the server bootstrap unconditional, which makes the latent migration bug much easier to trigger on machines with an old global install. #2139 correctly pins the later BYO provider handoff to `~/.local/bin/first-tree-staging`, but it does not change the service resolution done by `login`, the auto-update refresh command, or the version-pinned systemd PATH.

## Suggested Fix

1. Make `resolveCliInvocation()` portable-aware. When `FIRST_TREE_INSTALL_MODE=portable` / `FIRST_TREE_PORTABLE_ROOT` is present, resolve the service command from the stable portable root instead of ambient `PATH`.
2. In portable update glue, invoke `daemon refresh-unit` through the explicit new portable runtime/CLI under `current`; do not spawn a bare binary name.
3. Render a stable portable service PATH, including the portable shim directory as appropriate, rather than a concrete `versions/<version>/node/bin`.
4. Add an end-to-end regression test with:
   - a stale global binary first on the parent PATH,
   - portable `install.sh`,
   - absolute portable `login`,
   - service install/refresh,
   - one managed auto-upgrade,
   - assertions that service invocation and all modified package files remain in the portable `current/app` path.

Related: #1502, #1949, #2094, #2139.

baixiaohang:
## Summary

- make the portable install identity authoritative across service installation, drift detection, and managed refresh
- propagate the portable shim directory explicitly and render stable supervisor entrypoints/PATH values
- make portable updates prepare shims before the atomic `current` switch and block npm-global writes into portable version trees
- add filesystem-backed migration, rollback, prefix-guard, and supervisor regression coverage plus a reusable cross-surface QA case

## Migration behavior

New portable shims export `FIRST_TREE_PORTABLE_BIN_DIR` together with the existing install mode and stable `current` root. Existing default-location shims are accepted only when they can be verified against the same portable root and channel; ambiguous custom legacy shims fail with guidance to rerun the installer. Rerunning the installer rewrites both stable shims and refreshes the service through the absolute portable entrypoint.

Portable systemd and launchd definitions now use the explicit stable shim and `<prefix>/current/node/bin`. They no longer infer the First Tree executable from an inherited shell `PATH` or pin a concrete version directory.

## Risk and safety

- Portable identity is fail-closed: incomplete or inconsistent metadata never falls back to npm-global resolution.
- Both installer paths validate artifact identity and prepare stable shims before atomically switching `current`.
- npm-mode updates probe the global prefix with the same npm runtime and refuse any portable-shaped prefix whose provenance is valid or cannot be safely established.
- True npm-global installations keep their existing service resolution and update behavior.
- The deterministic subprocess coverage renders the real service templates and executes the real portable update/refresh chain, but a live systemd/launchd migration QA run has not yet been performed.

## Validation

- focused CLI regression suite: 9 files, 173 tests passed (`maxWorkers=2`), including the full stale-global/portable migration subprocess chain
- `pnpm check`
- `pnpm typecheck`
- `pnpm test` was run with `TURBO_CONCURRENCY=2`, `VITEST_MAX_FORKS=1`, and `umask 0022`: an earlier complete run passed 12/12 tasks; the final run passed the changed CLI task but hit two unrelated timing flakes in `@first-tree/skill-evals` `first-tree-qa/grader.test.ts`, which passed 8/8 immediately when rerun in isolation
- `sh -n scripts/portable/install.sh`
- `git diff --check`

Fixes #2145

### DIFF
--- a/apps/cli/src/__tests__/context-bundle-manifest.test.ts
+++ b/apps/cli/src/__tests__/context-bundle-manifest.test.ts
@@ -4,7 +4,7 @@ import { cpSync, existsSync, mkdirSync, mkdtempSync, readdirSync, readFileSync, 
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 import { contextIntegrationReleaseManifestSchema } from "@first-tree/shared";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   contextIntegrationMarketplaceSourcePath,
   installContextIntegration,
@@ -24,11 +24,26 @@ import { COMMAND_VERSION } from "../core/version.js";
 
 const roots: string[] = [];
 const originalFirstTreeHome = process.env.FIRST_TREE_HOME;
+const originalInstallMode = process.env.FIRST_TREE_INSTALL_MODE;
+const originalPortableRoot = process.env.FIRST_TREE_PORTABLE_ROOT;
+const originalPortableBinDir = process.env.FIRST_TREE_PORTABLE_BIN_DIR;
+
+beforeEach(() => {
+  delete process.env.FIRST_TREE_INSTALL_MODE;
+  delete process.env.FIRST_TREE_PORTABLE_ROOT;
+  delete process.env.FIRST_TREE_PORTABLE_BIN_DIR;
+});
 
 afterEach(() => {
   for (const root of roots.splice(0)) rmSync(root, { recursive: true, force: true });
   if (originalFirstTreeHome === undefined) delete process.env.FIRST_TREE_HOME;
   else process.env.FIRST_TREE_HOME = originalFirstTreeHome;
+  if (originalInstallMode === undefined) delete process.env.FIRST_TREE_INSTALL_MODE;
+  else process.env.FIRST_TREE_INSTALL_MODE = originalInstallMode;
+  if (originalPortableRoot === undefined) delete process.env.FIRST_TREE_PORTABLE_ROOT;
+  else process.env.FIRST_TREE_PORTABLE_ROOT = originalPortableRoot;
+  if (originalPortableBinDir === undefined) delete process.env.FIRST_TREE_PORTABLE_BIN_DIR;
+  else process.env.FIRST_TREE_PORTABLE_BIN_DIR = originalPortableBinDir;
 });
 
 describe("context integration bundle", () => {
--- a/apps/cli/src/__tests__/core-attention-update-extra.test.ts
+++ b/apps/cli/src/__tests__/core-attention-update-extra.test.ts
@@ -89,7 +89,14 @@ beforeEach(() => {
   bootstrapMocks.ensureFreshAccessToken.mockResolvedValue("token");
   bootstrapMocks.resolveServerUrl.mockReturnValue("https://hub.example");
   childRegistryMocks.classify.mockReturnValue({ kind: "permanent", reasonCode: "classified" });
-  spawnSyncMock.mockReturnValue({ status: 0, stdout: "0.6.0\n", stderr: "" });
+  spawnSyncMock.mockImplementation((_command: string, args: string[]) =>
+    args.includes("prefix")
+      ? { status: 0, stdout: "/usr/local\n", stderr: "" }
+      : { status: 0, stdout: "0.6.0\n", stderr: "" },
+  );
+  vi.stubEnv("FIRST_TREE_INSTALL_MODE", "");
+  vi.stubEnv("FIRST_TREE_PORTABLE_ROOT", "");
+  vi.stubEnv("FIRST_TREE_PORTABLE_BIN_DIR", "");
 });
 
 describe("core update helpers", () => {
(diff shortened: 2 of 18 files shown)


yuezengwu (approved):
Reviewed the portable installation-identity flow end to end. The change makes the explicit portable root/bin directory authoritative instead of consulting ambient `PATH`, renders stable supervisor entrypoints and PATH values, prepares shims before switching `current`, and blocks npm-global writes into portable version trees.

The migration and rollback coverage exercises the important stale-global/custom-bin and pre-commit failure cases. I found no blocking issues in the diff.

Review scope: diff inspection only; I did not rerun tests or QA.

bestony:
LGTM

PR #2156, opened 2026-08-04
fix(cli): skip BYO context hook in managed sessions

baixiaohang:
## Summary

- detect First Tree-managed agent sessions before external Context Plugin health checks or activation
- return a silent `continue` response so the user-scoped BYO SessionStart hook cannot override the managed agent briefing
- preserve external BYO activation unless both managed runtime markers are present

## Why

Managed Codex and Claude Code processes can still discover a user's globally installed `first-tree-context` plugin. The plugin's SessionStart hook previously injected `consumerKind: byo` into an already managed First Tree agent session, causing managed Context Tree writes to incorrectly request the BYO-only second human confirmation.

## Validation

- `pnpm --filter first-tree-dev exec vitest run src/__tests__/context-activate-command.test.ts src/__tests__/context-activation.test.ts --maxWorkers=2` (8 tests passed)
- `pnpm exec biome check apps/cli/src/commands/context/activate.ts apps/cli/src/__tests__/context-activate-command.test.ts`
- `pnpm --filter first-tree-dev... build`
- built CLI smoke: managed environment returns only `{"continue":true}`; external environment still returns `additionalContext`
- `git diff --check`

### DIFF
--- a/apps/cli/src/__tests__/context-activate-command.test.ts
+++ b/apps/cli/src/__tests__/context-activate-command.test.ts
@@ -1,5 +1,5 @@
 import type { Command } from "commander";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { CommandContext } from "../commands/types.js";
 
 const mocks = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
实现目标清晰：在 First Tree 托管 Agent 会话中，用户级 BYO Context SessionStart hook 只返回 continue，不再执行插件健康检查或注入 consumerKind: byo；外部 BYO 会话仍保持原激活路径。双标记检测与 runtime 注入链路一致，单标记回退和 stale-plugin 行为均有针对性测试覆盖。未发现 blocker，approve。

PR #2159, opened 2026-08-04
fix(client): migrate legacy managed skill modes

baixiaohang:
## Summary

- allow legacy Core Skill ownership probes to normalize npm/umask-derived modes without weakening ordinary reconciliation or publication verification
- preserve fail-closed behavior for unsafe unmatched targets and keep symlink/special-file rejection intact
- include the workspace-relative managed target in digest failures for actionable diagnostics
- reproject proven legacy targets through the existing quarantine, staging, journal, and atomic install flow

## Background

Older clients could materialize bundled Core Skills as `0775` directories and `0664` files under `UMask=0002`. The v1-to-v2 migration attempted a strict digest before ownership adoption, so provider startup could remain in `managed_skills_unsafe_discovery` retry indefinitely.

The compatibility policy is intentionally limited to the legacy Core ownership probe. Marker, v1 ledger/pair, and exact normalized bundled-content evidence remain the accepted ownership signals. A human decision during review kept exact-content adoption automatic for legacy targets; strict verification still quarantines the adopted unsafe target before provider publication.

## Safety boundaries

- ordinary managed target digesting still defaults to `enforce-safe`
- final publication verification remains strict
- unmatched unsafe same-name targets are preserved and fail closed
- internal symlinks and special files remain rejected
- no recursive in-place permission repair was added

## Verification

- `umask 0022; pnpm --filter @first-tree/client exec vitest run src/__tests__/managed-skills.test.ts --maxWorkers=2 --minWorkers=1` (109 passed)
- `pnpm exec biome check packages/client/src/runtime/managed-skills.ts packages/client/src/__tests__/managed-skills.test.ts`
- `git diff --check`

## Review

Independent First Tree chat review completed with no blocking findings before commit. The reviewer-requested exact-bundle/no-ledger regression was added and included in the final verification above.

### DIFF
--- a/packages/client/src/__tests__/managed-skills.test.ts
+++ b/packages/client/src/__tests__/managed-skills.test.ts
@@ -1,3 +1,4 @@
+import { execFileSync } from "node:child_process";
 import {
   chmodSync,
   cpSync,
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the implementation and tests; no blocking findings. The mode normalization is scoped to the legacy Core ownership probe, while unmatched unsafe targets still fail closed and final publication verification remains strict with quarantine/reprojection.

PR #2158, opened 2026-08-04
Add first-chat Orientation handoff

Gandy2025:
## Summary

- add an inline, immediately skippable Orientation stage to ordinary onboarding first chats
- keep the initial bootstrap visible and replayable without waking the agent until the user explicitly continues
- preserve the existing `first-tree-welcome` task guidance after Start, Skip, or a normal composer message
- add four lightweight video placeholders for multi-agent collaboration, Context Tree, GitHub automation, and data/security

## User flow

1. A current Web client starts an ordinary onboarding chat with the Orientation capability.
2. The server stores the existing bootstrap once, marks it with trusted versioned metadata, and keeps it silent but replayable.
3. Web renders one inline Orientation card beneath that bootstrap.
4. The user can select a chapter and start, skip the introduction globally, or send a normal composer message.
5. A visible human continuation addressed to the original onboarding target becomes the first wake; that target receives the bootstrap as preceding context and continues through the shipped onboarding guidance.

Completed Orientation collapses into a compact historical `Watch again` entry. The server-owned `continued` lifecycle is authoritative even when the continuation is outside the visible Web history slice. Refreshes and kickoff retries do not duplicate the chat, bootstrap, card, or wake.

## Compatibility, trust, and replay boundaries

- ordinary onboarding only; campaign actions and dedicated tree-setup kickoffs are excluded
- old Web clients that omit the capability retain the previous immediate-wake behavior
- a legacy retry of a pending Orientation promotes the existing bootstrap safely
- concurrent legacy retry and human continuation serialize on the chat row and transfer a still-pending wake without losing the user's content
- the original target is derived from the immutable trusted bootstrap under that same lock
- a message addressed only to another participant does not consume the handoff
- if the original target is removed while pending, messages to remaining participants continue normally and the lifecycle stays pending; routing to the removed target uses the ordinary validation path
- a multi-recipient continuation delivers the visible turn normally, but only the original target receives exceptional bootstrap replay
- Web optimistic completion uses the same recipient boundary; the server lifecycle remains authoritative
- the server strips both Orientation presentation and continuation metadata from caller-controlled writes
- only the exact server-marked first continuation can replay the bootstrap outside the ordinary 24-hour context window
- after completion, later triggers and participants added through silent-history backfill cannot reactivate the bootstrap
- recovery/redelivery of the exact continuation remains reconstructable until ACK
[...]

Issue #888, opened 2026-06-09
docs(contributing): add a support/question issue template


bestony:
## Problem

The repository currently has bug and feature request templates, but support or usage questions still go through blank issues. That makes the queue harder to triage because questions, incomplete bug reports, and product requests can look the same at first glance.

A small support/question template would help maintainers and agents route user confusion without requiring private follow-up.

## Proposed Direction

Add a GitHub issue template for usage questions or support requests.

The template should ask for the affected surface, version or install method, the exact command or workflow involved, what the user expected, and what they have already tried. It should also point security-sensitive reports to `SECURITY.md` instead of public issues.

## Acceptance Criteria

- [ ] A support/question issue template exists under `.github/ISSUE_TEMPLATE/`.
- [ ] The template applies the existing `question` label by default.
- [ ] The template asks for the affected surface: CLI, web, server, docs, GitHub integration, or other.
- [ ] The template asks for version/install context when relevant.
- [ ] The template asks for the exact command, screen, or workflow involved.
- [ ] The template links to `SUPPORT.md` and `SECURITY.md` for routing.
- [ ] The template is written in English.

## Good First Issue Fit

This is small, isolated repository metadata work. It does not require changing application code.

## Context

Part of #884.

PR #2164, opened 2026-08-04
feat(web): add timeline message filter and hand earlier-chat off to it

liuchao-001:
## What

Two changes that share one mechanism (requested by @liuchao-001 in chat):

1. **"Show earlier chat" rework.** The Need you review no longer loads chat history into the ask popup and renders an inline pair-filtered preview. The button (and the mobile pull-down gesture) now opens the **full chat** with a **transient** pair filter (`?focus=<askerId>` + `showAsk=false`): the timeline shows only the viewer's conversation with the asker. The filter is URL-carried and deliberately not persisted — every ordinary chat-selection path (`selectChat`, draft, need-you open/close, mobile equivalents) deletes `focus`, so re-opening the chat later shows all messages. This also removes the confusing state where an answered ask still appeared to leave the chat "filtered to the two of us".

2. **Explicit timeline message filter.** The chat header gains a filter control listing the agents the viewer has actually conversed with (mutual mentions / replies, humans excluded). Picking one narrows the timeline to that pair's conversation; a banner above the timeline names the filter, counts hidden messages, and restores the full view in one click. The explicit choice persists per **user + chat** in `localStorage` (`first-tree:chat-message-filter:v1`, user-bucketed like the draft store) and never reaches the server. Default remains all messages.

## Semantics

- Pair membership reuses the earlier-chat rule (now shared in `components/chat/pair-conversation.ts`): direct chats keep every pair-authored message; group chats additionally require the message to mention or reply to the other pair member, so the agent's side conversations don't leak in.
- Filtering is display-only: blocking-request derivation, open-request composer gating, and the AskTakeover keep reading the **unfiltered** sets, so a question from a filtered-out agent still blocks and stays answerable.
- Session events are hidden entirely while a pair filter is active — they carry no pair attribution — and reappear when the full view is restored. The unread pill and new-message divider count only messages reachable in the filtered view (frozen read watermark), and the focused handoff reports honestly when the reviewed request predates the loaded history window.
- A stored filter naming an agent with no conversed messages in the loaded window falls back silently to the full view (no blank timeline from stale storage).
- `pages/workspace/need-you/earlier-chat.ts` is deleted; its predicate lives on in `pair-conversation.ts` with ported tests.

## Tests

- `pair-conversation.test.ts` — pair membership (direct/group) + conversed-agent listing (ported from the deleted `earlier-chat.test.ts`).
- `message-filter-store.test.ts` — round-trip, isolation, corrupt-JSON recovery, prune cap, user scoping.
- `chat-view-dom.test.tsx` — transient `?focus=` filter + banner + "Show all messages"; header-control selection with localStorage persistence; stored-filter restore on open.
- `need-you-page-dom.test.tsx` — "Show earlier chat" hands off to the focused full chat and no longer fetches messages.

`pnpm check`, `pnpm typecheck`, and the full `@first-tree/web` suite (242 files / 2135 tests) pass.


baixiaohang (changes requested):
- Rationale: The pair-filtered timeline can surface an action that cannot clear, does not reliably exclude the selected agent's side work, and the Need you handoff changes a current decision-locked product contract without a paired Context Tree update.

**Risk level**: B-low
- Semantic lift: introduces persisted per-user/per-chat Web state and changes the timeline data projection -> B-low

**PR summary**
- Problem: Users need a less confusing way to inspect their history with an asker and to focus a busy group timeline on one agent.
- Approach: Replace Need you's inline earlier-chat preview with a transient focused full-chat route, and add a user-scoped, browser-persisted pair filter in the chat header.
- Impacted modules: Workspace and mobile navigation, Need you review, timeline composition, pair-conversation selection, and local filter persistence.

**Review findings**
❌ 1. The new-message pill still counts the unfiltered `mergedMessages`, while its click only scrolls the filtered DOM. If a message from another agent arrives after the current pair's last visible message, `pillCount` becomes non-zero, but scrolling to the bottom cannot advance `liveBottomVisibleId` to that hidden row, so the pill cannot clear. Please either derive this affordance from the visible message projection or make its action reveal the full timeline before scrolling, and cover the hidden-new-message case. [`packages/web/src/pages/workspace/center/chat-view.tsx:3099`]

❌ 2. The session-event branch only drops feeds whose `eventAgentId` differs from the selected agent. `SessionEventRow` has no counterpart or triggering-message relationship, so a selected agent's workgroup/error produced for another participant in the group still appears in the supposed pair-only view. That contradicts the no-side-conversation guarantee and can expose unrelated working text/tool activity. Please hide un-attributable session events while pair filtering, or add a reliable pair association before rendering them. [`packages/web/src/pages/workspace/center/chat-view.tsx:2833`]

❌ 3. This removes the lazy inline recent-pair preview and changes “More context” into a pair-filtered handoff, but the current durable Messaging decision still specifies a small recent pair window inside Need you followed by the original chat's full inspection view. The owning Cloud node is decision-locked for code changes. Please land a paired Context Tree decision update (including the intended relationship between transient pair focus and full inspection), or preserve the current contract in this PR. [`packages/web/src/pages/workspace/need-you/need-you-page.tsx:254`; `system/cloud/chat/messaging.md` → “Web Answering Surface” at Context Tree commit `4f35238ded3aa7e737ec1dedf9206d31999b57eb`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
## Review summary

This PR moves **Show earlier chat** out of the Need you card and into the full chat via a transient `?focus=<asker>` pair filter, then adds a separate header filter whose explicit choice is stored per user + chat in browser `localStorage`. The unfiltered open-request derivation and composer gating remain intact, which is the right boundary.

Data impact: there is no database, server schema, or API change. The only new persisted shape is the browser-local `first-tree:chat-message-filter:v1` map.

## Blockers

1. **`packages/web/src/pages/workspace/center/chat-view.tsx:2785` — the handoff cannot show earlier context for an old request.** `ChatView` explicitly loads only the latest 50 messages and has no pagination (`:1925-1942`). `blockingMessages` adds the window-independent open request itself, but not any messages preceding it. If Need you opens its oldest request after that request has fallen outside the latest-50 window, the focused timeline contains the request plus any newer pair messages; none of the conversation that led to the question is available, and the synthetic old request is shown next to the recent window without a history-gap indication. This makes **Show earlier chat** fail precisely for the requests most likely to need context, and regresses the prior path that at least fetched a larger window and reported when the request was outside it. Please load a window anchored at/before the request, or add pagination/jump-to-request support before replacing the inline context path.

2. **`packages/web/src/pages/workspace/center/chat-view.tsx:2793` — one-to-one detection uses the retired persistence type.** New chats are always stored as `type: "group"` (`packages/shared/src/schemas/chat.ts:44-53`); `direct` survives only for legacy reads. As a result, every newly created two-speaker chat takes the group predicate and drops pair-authored messages that lack a structured mention/reply, even though there is no third-party conversation to exclude. The focused Need you view can therefore hide ordinary one-to-one history. Derive this condition from the speaker membership shape (the same two-speaker contract already used by `computeRequiresMention`) rather than `chatDetail.type`, and cover the current `type: "group"` + two-speaker case.

I did not run tests or QA, per the review-only instruction; this verdict is based on the complete diff and repository contracts.

liuchao-001:
All five blockers across both reviews are addressed in 255930b:

**@baixiaohang's review**

1. **Pill counts the unfiltered set** → `pillCount` now counts over the visible projection (`timelineMessages`), with ordering still indexed against `mergedMessages` (the watermark id itself may be a filtered-out row). A hidden arrival contributes nothing until the filter lifts, so the pill can never become unclearable. Covered by the new DOM test "does not strand the new-messages pill on an arrival the pair filter hides".
2. **Selected agent's side-work session events leak into the pair view** → the pair view is now messages-only: ALL session events (workgroups and error rows) are hidden while the filter is active, since `SessionEventRow` carries no pair attribution. Asserted in the transient-focus DOM test.
3. **Decision-locked Messaging contract** → paired Context Tree update: agent-team-foundation/first-tree-context#888 (draft until this PR merges, per the write workflow). It rewrites the Web Answering Surface earlier-context paragraph (handoff to a transient pair-focused inspection view, full view one action away, honest out-of-window reporting) and adds a "Timeline Pair Filter" section to `workspace-conversations.md` capturing the display-only projection, local persistence boundary, and the unfiltered-derivation safety rule.

**@yuezengwu's review**

1. **Old requests lose their earlier context (latest-50 window)** → the focus handoff now (a) fetches a supplemental window at the API's 100-row cap — the exact reach the retired inline preview had — merged into the timeline, and (b) carries the reviewed request id (`?focusMsg=`), so when the question predates even the deepened history the banner says so explicitly ("This question is older than the loaded history — the conversation before it isn't shown"), restoring the prior path's honest "unavailable" reporting. Full pagination/jump-to-request remains a separate effort (the FOLLOW-UP note at the messages query already tracks it); this PR no longer regresses below the preview it replaced. DOM test: "reports when the focused request is older than the loaded history".
2. **`chatDetail.type === "direct"` is retired** → the "no third party" condition now keys on the two-speaker membership shape (same speakers-only contract as `computeRequiresMention`), with legacy `direct` kept as a fallback for old rows. DOM test: "keeps unaddressed pair messages in a two-speaker group-typed chat".

`pnpm check`, `pnpm typecheck`, and the full web suite (242 files / 2138 tests) pass.

liuchao-001:
Heads-up from the paired Context Tree PR review (agent-team-foundation/first-tree-context#888, [Approval deferred](https://github.com/agent-team-foundation/first-tree-context/pull/888#pullrequestreview-4853080384)): two findings there are about **this PR's behavior** (reviewed at 123da521), and `/system/cloud/NODE.md` locks code to that decision context, so source and Tree must converge one way or the other. Since this PR is already in a revision round, flagging them now so we settle them pre-merge instead of post-merge:

1. **Pair-filter attention semantics** — the intended design says the new-message pill counts only messages reachable in the pair-filtered view, and session events are hidden while filtering. Current code computes `pillCount` from unfiltered `mergedMessages` and retains session events for the selected agent. Either implement the filtered-view semantics or we downgrade the Tree claim to describe actual behavior.
2. **Request-predates-history signal** — the design says the focused handoff reports honestly when the reviewed request predates loaded history. The code removed the old preview/unavailable path but adds no equivalent request-age/history-window signal (the generic cache/server-gap banner doesn't cover this). Either add that signal or the claim comes out of the Tree.

(Third finding is Tree-side only — `/product` nodes still describe the old inline preview — and will be reconciled on the Tree branch after this PR merges.)

yuezengwu (changes requested):
## Re-review summary

I reviewed the delta from `123da521` to `255930bb`. The revision correctly addresses all five original blockers:

- the new-message pill now counts the visible pair projection;
- all unattributable session events are hidden while filtering;
- the transient handoff carries the reviewed request id, deepens the history window to 100, and reports an out-of-window request;
- one-to-one behavior now derives from the two-speaker shape rather than the retired `chat.type` value;
- draft Context Tree PR `agent-team-foundation/first-tree-context#888` exists and the revised source now matches its proposed Messaging/Workspace semantics.

There is still no database, server schema, or API change; persistence remains browser-local only.

## Blocker

**`packages/web/src/pages/workspace/center/chat-view.tsx:2592,2840-2844` — a failed supplemental-history fetch is reported as a successful bounded window.** TanStack Query's `isFetched` becomes true after either a successful fetch or a terminal error. If the new `limit: 100` request fails and the reviewed request is absent from the ordinary cache/latest-50 window, `focusContextMissing` therefore renders “This question is older than the loaded history,” even though the client never established that history window. If the request happens to exist in the ordinary window, the supplemental failure is invisible altogether. The retired inline path explicitly distinguished “Earlier chat could not be loaded,” so this regresses its error behavior and breaks the new promise of honest context reporting. Please gate the age inference on successful focus-window data and surface a retryable load-error state; cover the rejected focus-window query as well as the out-of-window success case.

I did not run tests or QA; this verdict is based on the updated diff and repository contracts.

baixiaohang (changes requested):
- Rationale: The revision closes the original implementation findings, but the replacement two-speaker predicate can still admit messages outside the selected pair; the supplemental-history error-path blocker is already covered by the other re-review.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/history projections -> B-low

**PR summary**
- Problem: Let a viewer inspect their history with one agent without confusing that focused view with the full group timeline.
- Approach: Use one pair-message projection for a transient Need you handoff and a separately persisted header filter, while correctness-bearing request state remains unfiltered.
- Impacted modules: Workspace/mobile navigation, Need you review, timeline/history composition, pair selection, and browser-local filter state.

**Review findings**
❌ 1. `twoSpeakerChat` checks only the current speaker count (or the retired `type="direct"` flag); it never verifies that the two speakers are exactly `myAgentId` and `filterAgentId`. A removed/departed agent can still be a filter candidate through loaded history (and an old open request can still focus that asker). If the chat now has two different speakers, this branch sets `directChat=true`, so every viewer-authored message — including messages addressed to the current peer — enters the supposed conversation with the departed agent. The `type="direct"` fallback has the same problem after a legacy chat gains another participant because participant mutation does not rewrite that retired field. Please define the no-third-party shortcut against the selected pair itself rather than count/type alone, and cover a stale/departed focus target. [`packages/web/src/pages/workspace/center/chat-view.tsx:2818`]

✅ The previous pill and session-event findings are fixed in `255930b`: attention now reads the visible projection and filtered views are messages-only.

✅ The handoff now restores the former 100-row reach and carries a request anchor; the failed supplemental-query behavior is already called out in yuezengwu's current-head review, so it is not duplicated here.

**Action taken**
- Submitted request changes on head `255930bb`.

yuezengwu (approved):
## Re-review summary

I reviewed the delta from `255930bb` to `2970b1dd`. The revision resolves both remaining blockers:

- the supplemental focus-window state now distinguishes success from failure, so request age is inferred only after a successful 100-message fetch; a failed fetch surfaces a retryable “Earlier chat could not be loaded” state, and the new DOM test covers failure, retry, and the subsequent honest out-of-window result;
- the no-third-party shortcut now requires the current speaker set to be exactly the viewer and selected agent, with no legacy `chat.type` fallback, so focusing a departed/stale agent cannot admit viewer messages addressed to the current peer; the new DOM test covers that stale-focus shape.

The overall change remains a display-only timeline projection plus browser-local per-user/per-chat persistence, with correctness-bearing request derivations left unfiltered. There is no database, server schema, or API change. The paired Context Tree update remains tracked in `agent-team-foundation/first-tree-context#888`.

I did not run tests or QA, per the review-only instruction; this approval is based on the complete prior review plus the current-head delta and repository contracts.

baixiaohang (changes requested):
- Rationale: The two reported blockers are fixed, but the exact-current-pair shortcut still treats a formerly multi-speaker chat as a direct conversation after its third speaker leaves, so historical side conversations can enter the pair view.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/history projections -> B-low

**PR summary**
- Problem: Let a viewer inspect their history with one agent without confusing that focused view with the full group timeline.
- Approach: Apply one pair-message projection to a transient Need you handoff and a separately persisted header filter, while correctness-bearing request state remains unfiltered.
- Impacted modules: Workspace/mobile navigation, Need you review, timeline/history composition, pair selection, and browser-local filter state.

**Review findings**
❌ 1. `pairIsWholeChat` proves only that the selected pair are the **current** two speakers, not that the loaded history was always a two-person conversation. Participant removal deletes the speaker membership row (`services/chat.ts:1160-1188`), so a chat can evolve `{viewer, A, B} → {viewer, A}`. Filtering on A then sets `directChat=true` and admits every old viewer/A-authored message, including messages either of them explicitly addressed to departed B. That is still side-conversation leakage in the promised pair-only view. Please retain the group addressing rule whenever the loaded history shows pair-authored messages addressing/replying outside the pair (or introduce another reliable historical no-third-party signal), and cover a current-exact-pair chat that previously had a third speaker. [`packages/web/src/pages/workspace/center/chat-view.tsx:2829`]

✅ The focus-window failure now has a distinct retryable state, and request age is inferred only after a successful supplemental fetch.

✅ Requiring the current speakers to equal the selected pair fixes the previously reported stale/departed-focus-target case and removes the unsafe retired-type fallback.

**Action taken**
- Submitted request changes on head `2970b1dd`.

yuezengwu (changes requested):
## Corrected re-review verdict

I am retracting my approval after independently confirming the concurrent review’s historical-membership case. The two blockers addressed in `2970b1dd` remain fixed, but one pair-isolation blocker remains.

❌ **`packages/web/src/pages/workspace/center/chat-view.tsx:2829` — current exact-pair membership does not prove the loaded history was always pair-only.** Participant removal deletes the speaker membership row while preserving immutable message history (`packages/server/src/services/chat.ts:1160-1188`). A chat can therefore evolve from `{viewer, A, B}` to `{viewer, A}`. Filtering on A then sets `directChat=true` and admits every historical viewer/A-authored message, including messages either one addressed or replied to B. That violates the pair-only view’s no-side-conversation contract.

Please retain the group addressing rule when loaded history demonstrates third-party exchanges (or use another reliable historical no-third-party signal), and add a regression case where the selected pair are the current two speakers but a departed third speaker appears in earlier messages.

There is still no database, server schema, or API change. I did not run tests or QA; this correction is based on the current-head source and participant-removal contract.

liuchao-001:
7460b3c addresses the remaining findings from all three review threads:

**Historical no-third-party (baixiaohang ❌1 on 2970b1dd, confirmed by yuezengwu's corrected verdict)** — the exact-pair shortcut now additionally requires the loaded history itself to be free of third-party traces: `historyContainsThirdParty` (in `pair-conversation.ts`) reports any message authored outside the pair, or any pair-authored message addressing outside the pair. A chat that evolved `{viewer, A, B} → {viewer, A}` therefore keeps the group addressing rule, at the documented cost of hiding mention-less pair messages in such chats. Unit tests cover both trace kinds; the DOM regression "keeps the group rule when the current exact-pair chat once had a third speaker" covers exactly the requested shape (current speakers = pair, departed third speaker in earlier messages, both an addressed-to-departed and a mention-less viewer message stay hidden).

**Read-watermark persistence (chat-side re-review by the review agent)** — the same root cause had one more consumer: `useReadTracker` persisted the **filtered DOM tip** as `latestKnownMessageId`, permanently marking hidden-but-older messages as known (the IDB row never self-heals). The tracker now takes `freezeLatestKnown` (set while the pair filter is active): the watermark stops advancing, a fully-filtered visit persists nothing (previous snapshot stays intact), and `bottomVisibleMessageId` keeps tracking the visible DOM as the scroll anchor. DOM test asserts the `setReadState` arguments for both the filtered visit (no write) and the unfiltered control (true tip).

**Non-blocking suggestions from the same re-review** — three taken: resolving an ask now clears the transient `?focus=`/`?focusMsg=` view (this closes the original user complaint end-to-end); the filter control is not offered when the only candidate is the sole other speaker (no-op filter); an unresolvable focus id renders "this agent" instead of a raw UUID. One declined with rationale: the deepened focus window staying in the query cache after "Show all messages" is deliberate — evicting it would make just-read context vanish from the full view; the cache GC owns its lifetime.

The paired Context Tree PR (agent-team-foundation/first-tree-context#888) now states the unfiltered-derivation rule per consumer (state-deciding derivations read unfiltered; attention affordances read the filtered projection; read-anchor persistence never advances past hidden messages; the pair shortcut needs both membership AND history to be third-party-free), so the rule is mechanically checkable against code.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2145 tests, all green.

yuezengwu (changes requested):
## Re-review summary

The `7460b3cf` delta correctly fixes the reported historical-membership leak: the direct shortcut now requires both the current exact pair and a loaded history free of third-party traces, with unit and DOM coverage for a shrunk group. It also keeps the prior supplemental-history error fix intact.

The new delta introduces two remaining blockers in adjacent filter/read-state behavior.

❌ 1. **`packages/web/src/pages/workspace/center/chat-view.tsx:2647-2667` — the “no-op” control suppression does not use the same history check as the projection.** Consider current speakers `{viewer, A}` with older messages authored by departed B to A, while B never addressed/replied to the viewer. `listConversedAgentIds` yields only A, so the new branch returns no candidates and removes the header control. But `historyContainsThirdParty` is true, so filtering on A is not a no-op: it would hide B’s historical side conversation. This leaves the explicit pair filter unavailable in a history shape this revision is specifically designed to handle. Please suppress the sole-peer option only when the actual pair projection is a no-op (for example, by sharing the current-pair + history-free predicate), and cover the departed-third-party-who-never-addressed-viewer case.

❌ 2. **`packages/web/src/pages/workspace/center/chat-view.tsx:3464` — read persistence freezes whenever a filter is active, even when it hides no messages.** A transient Need you focus in a true two-speaker chat renders the complete message DOM (`hiddenByFilter === 0`), yet `freezeLatestKnown` is still true. On a visit that begins focused, `latestKnownIdRef` remains null and the tracker writes neither the true tip nor the visible scroll anchor on leave. The next ordinary visit therefore retains a stale local new-message watermark/anchor even though the user saw the full timeline. Freeze only when the message projection is actually narrowed (session events do not participate in this message watermark), and add a control test showing that an active exact-pair/no-hidden focus still persists the real tip while the hidden-message case remains frozen.

There is no database, server schema, or API change. I did not run tests or QA; this verdict is based on the full delta and repository contracts.

baixiaohang (changes requested):
- Rationale: The historical pair-isolation fix is correct, but own-message send paths still bypass the new filtered-view watermark freeze and can permanently swallow hidden arrivals.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/read-state projections -> B-low

**PR summary**
- Problem: Let a viewer focus a busy timeline on one agent without confusing the focused projection with the full conversation or losing awareness of hidden work.
- Approach: Share pair-membership rules across transient Need you focus and an explicit local filter, while keeping request lifecycle state unfiltered and freezing filtered-DOM read persistence.
- Impacted modules: Timeline projection, Need you handoff, local filter/read state, workspace/mobile navigation, and pair-history selection.

**Review findings**
❌ 1. The text and file send success paths still write `latestKnownMessageId = saved.id` directly and advance `pendingHighWaterAdvance` regardless of an active narrowing filter. Repro: filter on A; a newer message from B arrives hidden; then the viewer sends to A. The own send is chronologically after B, so these writes advance both the session high-water and the durable IDB marker past B. Lifting the filter produces no pill for B, and the persisted marker prevents B from appearing new on a later visit — exactly the hidden-message loss that `freezeLatestKnown` is intended to prevent. Please make every own-send watermark path respect the narrowed projection (text and file), while retaining the in-filter no-own-message-pill behavior through a projection-local mechanism, and cover hidden arrival → own send → lift/revisit. [`packages/web/src/pages/workspace/center/chat-view.tsx:2173`; `packages/web/src/pages/workspace/center/chat-view.tsx:2444`]

✅ `historyContainsThirdParty` plus current exact-pair membership fixes the previously reported shrunken-group leak, with focused unit and DOM coverage.

⚠️ The concurrent current-head review already covers the separate candidate-control suppression and unconditional-freeze cases, so they are not duplicated here.

**Action taken**
- Submitted request changes on head `7460b3cf`.

liuchao-001:
904a02c addresses all three current-head findings (yuezengwu ❌1/❌2, baixiaohang ❌1) plus the review agent's refinement, with one shared mechanism change:

**Watermark: exact safe bound instead of a freeze (yuezengwu ❌2 + review agent's noise finding).** `freezeLatestKnown` is replaced by `latestKnownOverride`: while the pair filter is active, chat-view supplies the highest safe watermark — the message just before the **oldest hidden** one in the loaded order, which equals the true tip when the filter hides nothing, and never regresses a previously persisted value. Consequences, each covered by a DOM test:
- an exact-pair focus that hides nothing persists tip + scroll anchor exactly like an unfiltered visit ("persists the true tip when an active focus hides nothing" — the requested control case);
- a persistently filtered chat digests its read pair messages instead of re-flagging them as new every visit ("caps the persisted read watermark just below the oldest hidden message");
- a hidden message is still never marked known; when even the first loaded message is hidden, nothing is persisted (the coherent-pair tradeoff is now documented accurately in the hook's JSDoc).

**Own-send watermark paths (baixiaohang ❌1).** Both send success paths (text + file) and the optimistic pre-advance are gated by the narrowed-projection flag (a ref, read at callback time, so an arrival during an in-flight send is respected): while the filter hides messages, neither the durable read-state write nor the session pre-advance moves to the own send. No-own-pill behavior is projection-local already — the pill and divider skip viewer-authored rows outright. DOM test: hidden newest arrival → own send → neither `setReadState` nor the read-state cache ever carries the sent id.

**No-op suppression shares the projection predicate (yuezengwu ❌1).** `pairFilterKeepsEverything` (current exact pair AND `historyContainsThirdParty`-free window) is now the single predicate used by both the timeline projection and the sole-peer control suppression. The requested shape — departed third party who only ever addressed the current peer — keeps the control available and filtering meaningfully hides that side conversation ("still offers the filter when a departed third party never addressed the viewer").

Also documented in `historyContainsThirdParty`: out-of-window `inReplyTo` targets are unattributable by design (the alternative — treating every unresolvable reply as third-party — would break the shortcut for ordinary long DMs).

The Context Tree PR (#888) now states the watermark's exact bound so a plain freeze isn't read as the only compliant implementation, and the Context Reviewer's /product reconciliation is already on that branch.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2148 tests, all green.

yuezengwu (changes requested):
## Re-review summary

The `904a02c9` delta fixes the three reported code paths:

- sole-peer control suppression now shares the exact projection predicate;
- an active filter that hides no messages persists the true tip;
- explicit optimistic/success watermark advances for text and file sends are gated while messages are hidden.

One blocker remains in the same hidden-arrival → own-send flow.

❌ **`packages/web/src/pages/workspace/center/chat-view.tsx:3202-3212, 3517-3534` — the filtered DOM bottom still advances the session watermark past hidden messages, and can persist an unsafe scroll anchor beyond the new ceiling.** Repro: filter on A; hidden B arrives; viewer sends to A. The new guards skip `pendingHighWaterAdvance` and the direct `setReadState(saved.id, saved.id)` write, but `scrollToBottom("smooth")` still makes the visible own-send row the tracker’s `bottomVisibleMessageId`. `onBottomVisibleChange` then drives `sessionHighestId` to that own-send id without consulting `readWatermarkFrozenRef` or `latestKnownCeiling`. Because the own send is ordered after B, lifting the filter yields no pill for B.

The durable pair has the same hole: the tracker can write `{ bottomVisibleMessageId: ownSend, latestKnownMessageId: beforeOldestHidden }`. On the next ordinary visit, landing at the later own-send anchor makes the live bottom effect advance the session watermark past B again, so B has no visible pill even though the persisted “known” bound correctly stayed before it. The new test asserts only that `latestKnownMessageId !== "msg-sent"`; it does not lift the filter or revisit, so it misses both effects.

Please make the live/persisted bottom-anchor path respect the same safe bound (or otherwise keep a later filtered anchor from advancing unfiltered attention), and extend the regression through filter lift and a revisit. The hidden message must remain surfaced after the own send, not merely absent from the durable `latestKnownMessageId` field.

There is no database, server schema, or API change. I did not run tests or QA; this verdict is based on the current-head delta and read-state/session-watermark contracts.

baixiaohang (changes requested):
- Rationale: The three current-head findings are addressed, but the new “safe ceiling” still persists a scroll anchor beyond an interleaved hidden message, allowing the next full-view visit to swallow that message through the session high-water.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/read-state projections -> B-low

**PR summary**
- Problem: Let a viewer focus a busy timeline on one agent without losing awareness of messages outside the focused projection.
- Approach: Share one pair predicate across the projection and filter control, and cap persisted read state at the oldest hidden message while preventing own sends from crossing that bound.
- Impacted modules: Timeline projection, local read state, own-send high water, Need you focus, and filter controls.

**Review findings**
❌ 1. `latestKnownCeiling` caps only `latestKnownMessageId`; `useReadTracker` still persists the actual filtered `bottomVisibleMessageId`, which can be chronologically **after** the oldest hidden row. With ordered messages `visible-A1, hidden-B2, visible-A3`, leaving the filtered view writes `(bottom=A3, latestKnown=A1)`. The next ordinary visit lands at A3, below B2; the live bottom callback then advances `sessionHighestId` to A3, so the pill does not count B2, and the unfiltered tracker subsequently persists the true tip, permanently marking B2 known. Thus the hidden message is still swallowed despite the ceiling. Please keep the restored scroll/session anchor from crossing the same hidden boundary (or make the session watermark hole-aware), and cover filtered visit → full revisit with a hidden message interleaved between visible pair rows. The current test places the hidden row at the tip, where bottom and ceiling happen to coincide and cannot expose this case. [`packages/web/src/pages/workspace/center/chat-view.tsx:3501`; `packages/web/src/hooks/use-read-tracker.ts:263`]

✅ The shared no-op predicate now preserves a meaningful sole-peer filter in shrunken-group history.

✅ No-hidden focus persistence and text/file own-send durable/session gating now address the three findings on `7460b3cf`.

**Action taken**
- Submitted request changes on head `904a02c9`.

liuchao-001:
fe80b1e closes the remaining anchor holes from all three threads with one invariant: **every read/attention anchor — durable watermark, persisted scroll anchor, and live session high-water — is bounded by the message just before the oldest NOT-RENDERED row.**

**"Hidden" is now defined by rendering, not by the filter (review agent's blocker).** `latestKnownCeiling` builds its visible set from `visibleItems` — the post-truncation render list — instead of the filter projection. A stored filter + an unanswered blocking question therefore stops the watermark at the question: the never-rendered trailing message stays unknown regardless of which layer (pair filter or block truncation) hid it, and any future truncation layer is respected automatically. Regression: "keeps the watermark at the blocking question while a stored filter is active".

**Interleaved hidden anchor (baixiaohang ❌1 / yuezengwu's current-head blocker).** Two clamps, same bound:
- the tracker clamps the **persisted scroll anchor** to the ceiling, so `visible-A1, hidden-B2, visible-A3` persists `(A1, A1)` — the next ordinary visit lands above B2 and surfaces it (regression covers the full filtered-visit → unfiltered-revisit path and asserts the New Messages divider marks B2);
- the **live session high-water advance** clamps to the ceiling while filtered, so a filtered DOM bottom (including an own send, closing the session half of the earlier own-send flow) can never move unfiltered attention past a hidden message within the same session, and lifting the filter yields the pill for the hidden arrival.

The Context Tree PR (#888) wording is tightened accordingly: the bound is stated over ALL anchors and "hidden" is defined as not-rendered, so a filter-only reading of the rule can't pass review again.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2150 tests, all green.

baixiaohang (changes requested):
- Rationale: The persisted/full-revisit anchor hole is fixed, but clearing the filter in the same session releases the clamp while the live bottom id still comes from the filtered DOM, so the hidden message can still lose its pill immediately.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/read-state projections -> B-low

**PR summary**
- Problem: Let a viewer focus a busy timeline without losing awareness of messages excluded by the focused projection.
- Approach: Define one rendered-prefix ceiling and apply it to durable known state, persisted scroll position, and live session high water.
- Impacted modules: Timeline projection/truncation, read tracking, local scroll state, new-message attention, and Need you focus.

**Review findings**
❌ 1. The session clamp is released on the same render that clears the filter, while `liveBottomVisibleId` still holds the last **filtered-DOM** bottom. With `visible-A1, hidden-B2, visible-A3`, the filtered effect correctly clamps `sessionHighestId` to A1. Clicking “Show all messages” then changes `latestKnownCeiling` to `undefined`; the advance effect re-runs immediately with the stale `liveBottomVisibleId=A3` and promotes the high water to A3. B2 may appear for one paint, but the pill then disappears without the user reaching B2. A filtered own-send produces the same stale-bottom shape. Please invalidate/provenance-gate the filtered live bottom across the projection transition (or keep the ceiling until an unfiltered viewport observation is established), and cover same-session filter lift by asserting the interleaved hidden message remains in the pill after effects settle. The new regression covers only unmount + full revisit, not this direct transition. [`packages/web/src/pages/workspace/center/chat-view.tsx:3250`]

✅ Persisted `bottomVisibleMessageId` is now clamped with the durable watermark, fixing the previously reported full-revisit path.

✅ Building the ceiling from `visibleItems` also correctly includes blocking-question truncation in the not-rendered boundary.

**Action taken**
- Submitted request changes on head `fe80b1ea`.

yuezengwu (changes requested):
## Re-review summary

The `fe80b1ea` delta correctly applies the rendered-prefix ceiling to the durable known watermark and persisted scroll anchor, fixing the full-revisit hole. Deriving the bound from `visibleItems` also correctly accounts for blocking-question truncation.

One same-session blocker remains.

❌ **`packages/web/src/pages/workspace/center/chat-view.tsx:3250-3277` — clearing the filter releases the ceiling while the live bottom id still belongs to the filtered projection.** With `visible-A1, hidden-B2, visible-A3`, the filtered render clamps `sessionHighestId` to A1. Clicking “Show all messages” changes `latestKnownCeiling` to `undefined`, so the advance effect immediately reuses the stale `liveBottomVisibleId=A3` and promotes the session high-water past B2. The same shape occurs after an own send while filtered. B2 can therefore lose its new-message pill without the user reaching it, even though the persisted revisit path is now safe.

Please invalidate or provenance-gate the filtered live-bottom observation across the projection transition, or retain the safe ceiling until an unfiltered viewport observation establishes a valid successor. Add a regression that clears the filter in the same mounted view and, after effects settle, verifies the interleaved hidden message still contributes to the pill. The new test covers unmount + full revisit, not this direct transition.

There is no database, server schema, or API change; persistence remains browser-local. I did not run tests or QA, per the review-only instruction.

liuchao-001:
ce67073 addresses the same-session transition blockers (baixiaohang + yuezengwu on fe80b1e) and the review agent's fourth-anchor finding:

**Stale filtered bottom across the projection transition.** Every bottom-visible observation now carries the projection it was measured under (`{ id, projection }`, tagged at publication). The session high-water advance trusts an observation only when its projection matches the CURRENT filter state — so releasing the ceiling on the lift render cannot combine with a bottom id measured against the old filtered DOM. A successor observation appears only when the new DOM's bottom genuinely differs (the tracker publishes on change), i.e. after real scroll/layout movement.

**Divider anchor is the fourth bounded anchor.** The divider's scroll-past-top dismissal advanced `dividerAnchorMessageId` from the raw live bottom (or the unfiltered chat-tip fallback) — both able to sit past a hidden message while filtered, killing divider + pill for it session-wide. Both paths now go through the same bound; when no anchor is safe the current anchor is kept rather than advancing blind.

**One shared clamp.** All session-anchor advances (high-water, divider live path, divider fallback) flow through a single `clampToCeiling` helper next to the ceiling derivation, so "is there another unclamped anchor?" is a one-glance review question; the tracker's persisted pair applies the same bound internally via `latestKnownOverride`.

Regression (requested shape): interleaved fixture, divider dismissed WHILE filtered (IntersectionObserver driven manually), then "Show all messages" in the same mounted view — after effects settle the hidden message still has its New Messages divider and contributes to the pill.

The Context Tree PR (#888) now enumerates the bounded anchors by name (durable watermark, persisted scroll anchor, live session high-water, divider anchor) so "every anchor" is checkable without counting by hand.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2151 tests, all green.

baixiaohang (changes requested):
- Rationale: The session high-water now rejects stale-projection bottoms, but the divider observer still consumes the same bottom as a bare id and can cross the hidden boundary during the filter-lift transition.

**Risk level**: B-low
- Semantic lift: persists per-user/per-chat Web state and changes timeline/read-state projections -> B-low

**PR summary**
- Problem: Let a viewer focus a timeline without losing new-message attention for content excluded from that projection.
- Approach: Bound every attention anchor to the rendered prefix and tag live bottom observations with the projection that produced them.
- Impacted modules: Timeline filtering/truncation, live and persisted read anchors, divider lifecycle, and Need you focus.

**Review findings**
❌ 1. The new provenance gate protects only the session-high-water effect. The divider observer still mirrors `liveBottomVisibleId` alone (`liveBottomVisibleIdRef`) and reads it after the projection changes. On “Show all messages,” inserting the hidden rows can fire the IntersectionObserver while the divider is above the viewport: `clampToCeilingRef.current` now reflects the unfiltered render (`undefined` ceiling), but `liveBottomVisibleIdRef.current` can still be filtered A3. The callback therefore promotes `dividerAnchorMessageId` to A3, past hidden B2, which removes B2’s divider and raises `unreadAnchorIdx`/pill baseline despite the session-high-water gate. The fallback-to-chat-tip path has the same transition provenance gap. Please carry/check the projection for the divider’s live observation too (and avoid the tip fallback until a current-projection observation exists), then cover an IntersectionObserver dismissal fired **after** Show all but before any new unfiltered bottom publication. The added test dismisses the divider while still filtered, so it cannot catch this ordering. [`packages/web/src/pages/workspace/center/chat-view.tsx:3427`; `packages/web/src/pages/workspace/center/chat-view.tsx:3473`]

✅ The structured live-bottom provenance correctly prevents the previously reported session-high-water advance on filter lift.

✅ The shared clamp now covers ordinary filtered divider dismissal and the persisted anchors.

**Action taken**
- Submitted request changes on head `ce670733`.

yuezengwu (changes requested):
## Re-review summary

The `ce670733` delta fixes the reported high-water transition: bottom observations now carry their pair projection, so the session watermark rejects a filtered observation after the filter is cleared. The shared ceiling also correctly bounds divider dismissals that execute while the filter is still active.

One projection-transition blocker remains in that divider path.

❌ **`packages/web/src/pages/workspace/center/chat-view.tsx:3426-3503` — the divider observer still consumes a projection-less stale bottom after the filter lifts.** `liveBottomVisibleIdRef` retains only the id, while `clampToCeilingRef.current` always uses the current projection. With `visible-A1, hidden-B2, visible-A3`, clear the filter and then let the existing `IntersectionObserver` callback run (including a callback already queued around the transition). The ref still supplies filtered `A3`, but the current clamp has no ceiling, so the callback sets `dividerAnchorMessageId=A3`. Because `unreadAnchorIdx` participates in `sessionHighestIdx`, that advances the effective attention boundary past B2 and removes both its divider and pill despite the new high-water provenance gate.

Please carry projection provenance through the divider live-bottom path as well (and ensure the chat-tip fallback cannot cross a transition boundary), or otherwise keep the old ceiling available to callbacks derived from the old projection. Extend the same-session regression by invoking the divider observer **after** `Show all messages` and verifying B2 remains new; the current test invokes it only while the filter is active, when the ceiling necessarily clamps it.

There is no database, server schema, or API change; persistence remains browser-local. The paired Context Tree update is still tracked in `agent-team-foundation/first-tree-context#888`. I did not run tests or QA, per the review-only instruction.

liuchao-001:
8976f7e closes the divider-observer transition gap (both current-head reviews):

**Projection provenance through the divider path.** The observer now reads the full observation (`{ id, projection }`) instead of a bare id and — same rule as the session high-water — refuses any observation whose projection does not match the CURRENT filter state. A callback firing around the "Show all messages" transition (the timeline re-render can trigger the IntersectionObserver) can therefore never pair the stale filtered bottom with the new projection's ceiling-less clamp. The chat-tip fallback is suppressed until a current-projection observation exists: it serves only the pre-tracker first paint, where the observation is still `{null, null}` and matches an unfiltered mount — it can no longer cross a transition boundary.

**Regression extended with the requested ordering:** the same-session test now ALSO fires the divider dismissal AFTER "Show all messages" but before any fresh unfiltered publication, and verifies the hidden message keeps its New Messages divider and its pill.

Also from the review agent's pass on ce67073 (no blockers there): the fifth anchor it identified — the server-side chat-level `markMeChatRead` — is a pre-existing, chat-granular over-claim (blocking truncation already exhibited it before this PR) and is intentionally out of this PR's scope. Filed as #2168 with the fix directions; the Context Tree rule in first-tree-context#888 is now explicitly scoped to LOCAL timeline anchors and names this exclusion, so it can't read as an uncovered gap.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2151 tests, all green.

yuezengwu (approved):
## Re-review summary

The `8976f7e7` delta resolves the remaining projection-transition blocker:

- the divider observer now reads the complete `{ id, projection }` observation and rejects stale filtered observations after the filter is cleared or retargeted;
- the chat-tip fallback is unavailable across that mismatch, so it cannot bypass the provenance check;
- the same-session regression now invokes the divider dismissal after **Show all messages**, before a fresh unfiltered bottom publication, and verifies that the interleaved hidden message retains both its divider and pill.

The prior fixes remain coherent: all local timeline attention anchors are bounded by the rendered-prefix ceiling, persisted anchors cannot cross an interleaved hidden row, own-send paths respect a narrowed projection, and the focused history fetch distinguishes load failure from an honestly out-of-window request.

The chat-granular server `markMeChatRead` over-claim is pre-existing (blocking truncation already exhibits it), is explicitly outside the local timeline-anchor contract, and is separately tracked in #2168; it is not a blocker for this PR.

There is no database, server schema, or API change; persistence remains browser-local. The paired Context Tree update remains in draft PR `agent-team-foundation/first-tree-context#888`, ready to follow the source merge. I did not run tests or QA, per the review-only instruction.

liuchao-001:
be6acee is a conflict-resolution merge of origin/main only (no force-push, no logic delta): #2158/#2166 added an import block in `chat-view.tsx` adjacent to this branch's `pair-conversation` imports — the resolution keeps both. Full verification on the merged tree: `pnpm check` / `pnpm typecheck` green, web suite now 243 files / 2159 tests, all passing (the +1 file / +8 tests are main's new orientation tests riding in with the merge).

yuezengwu (approved):
## Post-merge re-review

`be6acee5` only merges current `origin/main` into the feature branch. I inspected both parents and the remerge diff:

- the sole conflict was the adjacent import block in `chat-view.tsx`;
- the resolution correctly retains main's onboarding-orientation imports and this PR's pair-conversation imports;
- there is no feature logic delta from the previously approved `8976f7e7` head, and `git diff --check` is clean.

The previously approved timeline-filter, focused-history, and local read-anchor behavior is unchanged. The PR's net diff still introduces no database, server schema, or API change; the database migration present in the first-parent merge diff is already part of the updated `main` base, not this PR's change. The paired Context Tree update remains draft PR `agent-team-foundation/first-tree-context#888`.

I did not run tests or QA, per the review-only instruction.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All reported correctness blockers are resolved, the latest head only reconciles adjacent imports from current main, and the owner authorized the source-first sequence with the paired decision update.

**Risk level**: B-low
- Semantic lift: introduces per-user/per-chat browser persistence and changes timeline/read-state projections -> B-low

**PR summary**
- Problem: Users need to focus a busy conversation on one agent and inspect earlier request context without losing awareness of hidden messages or confusing a transient review handoff with a persistent filter.
- Approach: Reuse one pair-conversation projection for transient Need you focus and an explicit local filter, while request lifecycle derivations remain unfiltered and every local attention anchor respects the rendered-history boundary.
- Impacted modules: Workspace and mobile navigation, Need you review, timeline composition, pair selection, local filter persistence, and local read tracking.

**Review findings**
✅ Pair isolation covers current and historical third-party traces, and unattributable session events stay out of filtered views.
✅ Focused history distinguishes load failure from an honestly out-of-window request.
✅ Pill, divider, durable watermark, persisted scroll anchor, own-send paths, and live session high-water preserve attention across hidden interleaved messages and filter transitions.
✅ `be6acee5` only merges main; the sole conflict resolution retains both onboarding-orientation and pair-conversation imports with no feature logic delta.
✅ Paired Context Tree PR `agent-team-foundation/first-tree-context#888` is tracked for immediate post-source reconciliation under the owner-authorized source-first sequence.

**Action taken**
- Approved exact head `be6acee5`.

PR #2169, opened 2026-08-04
Safely co-locate SCM personnel target routes

baixiaohang:
## Summary

- preserve exact `(humanAgentId, wakeAgentId)` SCM attention identities while safely co-locating assignment and mention targets in one uniquely eligible entity chat
- split card display priority from persistent-line requirements and introduce explicit `reuse_membership`, `reuse_with_line`, and `strict_new_line` placement outcomes
- serialize GitHub and GitLab entity mutations with atomic membership admission, exact mapping persistence, authority revalidation, and commit-time cache invalidation
- make chat membership and provider identity lock ordering stable across follow, rebind, unfollow, webhook placement, and cross-provider concurrency
- extend GitHub/GitLab regression coverage and the static SCM attention parity QA case

## Risk

This is a high-risk cross-surface change spanning provider routing, private-agent membership, attention-line persistence, and concurrency. Unsafe, ambiguous, stale-authority, cross-owner, or multi-candidate cases fail closed to strict placement instead of guessing a chat. Existing lines are never deduplicated by human or manager identity.

The durable routing and membership contracts were approved and merged in agent-team-foundation/first-tree-context#880 before this code PR.

## Validation

- `pnpm check` (passes with the repository baseline of 18 warnings and 7 infos)
- `pnpm typecheck` (11/11 tasks)
- `pnpm --filter @first-tree/server exec vitest run --maxWorkers=2` (263 files, 3075 tests)
- `git diff --check`
- independent static code review: passed with no blockers

Formal agent-run QA was not executed. The cross-surface static QA case was updated for a future authorized run.

### DIFF
--- a/packages/qa/cases/cross-surface/scm-entity-attention-parity.md
+++ b/packages/qa/cases/cross-surface/scm-entity-attention-parity.md
@@ -38,12 +38,31 @@ card significance, lifecycle projection, topic protection, and archive behavior.
 - Unfollow in one chat and redeliver an ordinary event. Confirm the old chat stays silent. Then deliver an explicit
   reviewer, assignee, or exact body/comment mention and confirm a fresh route may be established without reviving the
   removed line.
-- Exercise reviewer, assignee, and mention targets with equivalent identities. Only reviewer routing may reuse exactly one
-  eligible membership chat without writing a new line; assignee and mention routing establish the target's own line/chat.
-  Ambiguous membership reuse must fail closed. Also create an agent follow whose stable human carrier is the reviewer but
-  whose wake agent differs from that human's current delegate. Confirm the carrier line remains intact, the current
-  delegate is still addressed by exact pair, reviewer membership reuse happens only when both sides already speak in one
-  candidate chat, and the strict path otherwise creates a separate deliverable line without human-scoped fallback.
+- Exercise reviewer, assignee, and mention targets with equivalent identities. Confirm review-only routing reuses exactly
+  one chat only when the target human and current delegate are already speakers, and writes no new line. For assignment
+  and mention, create an agent follow whose stable human carrier has a different current delegate. Confirm one uniquely
+  safe entity chat is reused, the current delegate is atomically admitted when necessary, and an exact sibling line is
+  written without deleting or replacing the original follow line. The chat must receive one card and wake the fresh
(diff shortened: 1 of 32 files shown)


yuezengwu (changes requested):
Request changes — one blocking routing gap.

[P1] Revalidate related-chat membership before persisting the new personnel line

`resolveTargetChatInTransaction` still lets a GitHub `strict_new_line` target fall through to Fixes-link reuse (`github-entity-chat.ts:315-329`), and GitLab now does the equivalent via `findGitlabRelatedEntityChat` (`gitlab-webhook.ts:1043-1052, 1116-1136`). Both paths select the related chat solely from an old exact-pair mapping. That chat is not part of the current entity's locked candidate membership snapshot, and neither path verifies that the wake agent is still a speaker before inserting the new mapping.

If the delegate previously left that related chat, the webhook commits a brand-new sibling line pointing at a chat where its wake agent is absent; the subsequent card is filtered to a silent delivery. This violates the new invariant that no mapping may commit unless its wake agent is a speaker in the selected chat.

Please either bypass related-entity reuse for fresh personnel lines, or bring the selected related chat under the same membership fence and revalidate/admit the exact pair before the mapping insert. Add regression coverage for both providers with an exact related mapping whose wake agent has been removed from the chat.

baixiaohang:
Addressed the blocking related-chat routing gap in `feb8b5228`:

- GitHub `strict_new_line` personnel targets now bypass Fixes-link reuse; the existing related-entity behavior remains limited to `provider_task_target`.
- GitLab `strict_new_line` personnel targets now create a canonical fresh H/W speaker chat instead of reusing an unchecked related mapping.
- Added symmetric GitHub and GitLab regressions where the exact related mapping exists but the wake agent has left that chat; both assert the new line lands in a fresh chat where the wake agent is a speaker.

Validation after the follow-up: `pnpm check`, 11/11 typecheck tasks, 5 affected test files / 102 tests, and `git diff --check` all pass. The independent follow-up review also passed with no blockers.

yuezengwu (approved):
Follow-up review complete. `feb8b5228` resolves the blocker: strict personnel targets no longer reuse unchecked related-entity chats; both providers create a fresh wakeable chat, and symmetric regressions cover stale related mappings after wake removal.

No further blockers found. I did not rerun tests or QA.

PR #2177, opened 2026-08-05
fix(web): hide empty Agent Template Responsibilities entry

yuezengwu:
## Summary

- Hide Agent Detail **Responsibilities** only when both the public official Template catalog and the agent's `templateIds` are confirmed empty from settled successful queries.
- Preserve the entry for adopted Template responsibilities, including retired or missing official provenance. Loading, background refetches, initial errors, and retained-data refetch errors fail open; human agents remain unchanged.
- Apply the same gate to the Agent Detail shell, the Responsibilities route, and the agent switcher. A confirmed-empty deep link redirects to Profile; the switcher preserves Responsibilities while the target is uncertain and falls back only after confirmed empty state.

## Validation

- [x] Focused Agent Detail Vitest: 3 files / 56 tests passed
- [x] Full `@first-tree/web` Vitest: 243 files / 2181 tests passed
- [x] Biome check on all 8 changed Web files
- [x] Independent `test-only` QA at exact head `79d8f37b860f02e561079241668d03ed65ca4662`, including the cached-empty background refetch regression
- [x] GitHub CI and CodeQL checks passed

### DIFF
--- a/packages/web/src/pages/agent-detail.tsx
+++ b/packages/web/src/pages/agent-detail.tsx
@@ -40,7 +40,12 @@ import { AgentSwitcherStrip } from "./agent-detail/agent-switcher-strip.js";
 import { useAgentResources } from "./agent-detail/capability-section.js";
 import { ContextBar } from "./agent-detail/context-bar.js";
 import type { AgentDetailContext, RuntimeSwitchClaimView } from "./agent-detail/layout-context.js";
-import { buildTabs, shouldShowResponsibilitiesTab, type TabDef } from "./agent-detail/tabs.js";
+import {
+  buildTabs,
+  responsibilitiesSideFromQuery,
+  shouldShowResponsibilitiesTab,
+  type TabDef,
+} from "./agent-detail/tabs.js";
 import { useAgentConfigSave } from "./agent-detail/use-agent-config-save.js";
 import { useLegacyAnchorRedirect } from "./agent-detail/use-legacy-anchor-redirect.js";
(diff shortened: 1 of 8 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The visibility gate closes only after both catalog and agent responsibility queries settle successfully as empty, while uncertain and historical-provenance states remain available.

**Risk level**: B-low
- Path baseline: only packages/web/** changes -> C.
- Semantic lift: React Query cache/fetch state now controls tab availability, deep-link redirects, and agent-switcher routing -> B-low.

**PR summary**
- Problem: Agents with neither available official Templates nor adopted Template responsibilities currently show an empty Responsibilities entry, while hiding it too aggressively would erase access to historical responsibility provenance or misroute users during transient query states.
- Approach: Centralize the settled-empty/fail-open decision, reuse it in the Agent Detail shell and route, and let the switcher preserve the current Responsibilities path until a target is confirmed empty.
- Impacted modules: Agent Detail tabs, Responsibilities route, agent switcher, shared resource query controller, and focused Web tests.

**Review findings**
✅ 1. The shared gate is the single decision point for shell, route, and switcher behavior, reducing navigation drift.
✅ 2. Non-empty agent templateIds win over an empty public catalog, preserving adopted, retired, or missing Template provenance.
✅ 3. Loading, background refetch, initial error, and retained-data refetch-error states fail open; the focused tests exercise the settled-empty transition and cached-error regression.

**Action taken**
- Approved at exact head 79d8f37b860f02e561079241668d03ed65ca4662.

PR #2194, opened 2026-08-05
feat(web): unify Need you into the chat ask takeover; refine ask interactions

liuchao-001:
Five interaction changes requested by @liuchao-001 in chat, building on #2164.

## What

1. **Need you opens the real chat.** The entry now navigates straight to the OLDEST open request's chat — the ask takeover floats over the full conversation ("背面显示所有的聊天"), and the standalone review page is **deleted**. The entry carries a queue session (`?nq=1`): each resolution **auto-advances to the next chat** holding an open question (owner-confirmed requirement); questions in the same chat advance in place via the existing FIFO takeover; an empty or unavailable queue ends the session; every ordinary chat-selection path deletes `nq`, so manually switching away exits the queue. Advancing is a history push — Back walks the queue hops.
2. **No X on the takeover.** "Close" mis-stated the semantics (the question stays open — it's a hide). The affordance is now an explicit **Show earlier chat** button (top-right, also bound to Escape): the card shrinks into the composer's pending-question affordance ("有 N 条待处理的问题" reopens it) and the timeline behind shows the viewer ↔ asker conversation via the #2164 transient focus mechanism.
3. **No current-chat summary block on the popup** — it died with the review page; the real chat is visibly behind the card instead.
4. **Show earlier chat button** is the single hide affordance across desktop and mobile (the undiscoverable mobile pull-down gesture is retired with the `contextBefore`/`onRequestEarlierContext` props).
5. **Ask agent is a mode, not a second input.** Clicking Ask agent swaps the answer surface into the clarification composer (one input at a time); the same footer button sends the question; Cancel (or Escape) restores answering. Skip/Submit leave the footer while composing a clarification so a resolving action can't fire against a half-written question.

Also retired: the unmounted mobile ask-sheet and its local answer cache (dead since the mobile Work unification).

## Contract notes

- The `?nq=` flag joins `focus`/`focusMsg`/`showAsk` in the transient-URL-param family with identical hygiene (all selection paths clear it).
- The auto-advance fetches the queue fresh after the resolving send (the send transaction clears the open request server-side), optimistically dropping the just-resolved item.
- The Need you entry's error state now retries the fetch on click (there is no review page to recover in); it disables on confirmed zero as before.
- This changes the Messaging tree contract (review-surface paragraph): a paired first-tree-context PR will follow this merge, source-first, as with #2164.

## Tests

- `ask-takeover-dom`: no close control; Show earlier chat button + none-supplied variant; Ask agent mode swap (answer surface replaced, Skip/Submit hidden, Cancel restores); all existing flows repointed at the new props (36 tests).
- `chat-view-dom`: new "need-you queue session" describe — advance to next chat / end-in-place on empty queue / no navigation outside a session; inspection test updated to the Show earlier chat affordance.
- `ask-agent-nav-guard-dom`: rail/new-chat/filter locking re-anchored on the chat-based queue session URL.
- `need-you-entry-dom`: error-click retries the fetch; count-click opens the oldest request's chat.
- `card-actions-dom` (mobile): entry navigates to `?c=<chat>&nq=1` instead of the review route.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2202 tests, all green.

**E2E note:** the ask takeover only exists with a live open request (an agent asking a human), which a static Momentic yaml can't seed by itself — live validation of the five flows is requested from the review agent's environment (as with #2164); a committed Momentic case can follow once we have an ask-seeding fixture.


yuezengwu (changes requested):
This PR moves the Need you flow onto the real chat takeover, adds the `nq` cross-chat queue session, replaces dismiss with “Show earlier chat,” turns Ask agent into a mutually exclusive composer mode, and removes the obsolete standalone/mobile review surfaces. I found three issues that need to be fixed before merge:

1. **[P1] Prevent Enter from resolving the hidden answer while Ask agent mode is open** (`packages/web/src/components/chat/ask-takeover.tsx:393`). `canReply` still reflects the answer draft hidden behind the clarification composer, and the window-level Enter handler does not check `askAgentOpen`. Repro: select an option (or type a valid answer), enter Ask agent mode, focus the clarification textarea, and press Enter on desktop. The handler calls `reply()` and resolves the request with the hidden answer, exactly the action this mode is intended to make impossible. Gate the resolving shortcut while clarification mode is active (and add the keyboard regression case).

2. **[P1] Do not let a stale queue continuation override a manual chat switch** (`packages/web/src/pages/workspace/center/chat-view.tsx:2691`). The async resolution path builds `nextParams` from the `searchParams` captured when Submit/Skip started. While the send is in flight, the desktop rail remains usable; selecting another chat correctly deletes `nq`, but when the send completes this callback still sees the stale `nq=1` snapshot and navigates to the next queue chat, overriding the user's manual selection and re-entering the session. Before advancing, verify the live location/session still belongs to this chat (or lock the relevant navigation through completion), and cover the in-flight manual-switch race.

3. **[P2] Reserve layout space for the expanded Show earlier chat action** (`packages/web/src/components/chat/ask-takeover.tsx:750`). The new text button is absolutely positioned, but the question body still starts at `var(--sp-6)` with only normal horizontal padding (`:795`). On mobile the button spans from 8px to roughly 52px vertically while the question begins around 24px, and its much larger width covers a substantial part of the first one or two lines. Add top/right clearance or place the action in a non-overlapping header.

No database or core data-structure change is present in this diff. I did not rerun the reported automated suite or start live QA because the deterministic review blockers above must be resolved first.

baixiaohang (changes requested):
- Rationale: The unified flow is directionally coherent, but the current head can still resolve the hidden answer during clarification and can override an explicit manual queue exit.

**Risk level**: B-low
- Semantic lift: the Web change now owns React Query-backed cross-chat queue continuation and navigation -> B-low

**PR summary**
- Problem: Humans currently answer blocking agent questions through separate Need you and chat surfaces, with ambiguous hide/close behavior and competing answer/clarification inputs.
- Approach: Make the real chat takeover the single answer surface, carry a transient `nq` queue session across chats, replace dismiss with Show earlier chat, make Ask agent a mutually exclusive mode, and remove the obsolete review/mobile implementations.
- Impacted modules: Workspace and mobile navigation, chat request takeover, Need you query/entry, request-resolution continuation.

**Review findings**
❌ 1. Enter can resolve the hidden answer while Ask agent mode is active: `canReply` remains true from the hidden answer draft and the window shortcut does not gate on `askAgentOpen`. A desktop Enter in the clarification textarea can therefore call `reply()` instead of keeping the request open. [R4 / `packages/web/src/components/chat/ask-takeover.tsx:393`]
❌ 2. The async queue continuation uses the `searchParams` snapshot captured before the send. Because the desktop rail remains usable while resolution is in flight, a manual chat switch can delete `nq` and then be overwritten when the stale callback advances to the next queue chat. Verify live session/chat ownership at completion or lock the navigation consistently, and cover the race. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:2691`]
⚠️ 3. The expanded Show earlier chat text button is absolutely positioned over the same top area used by the question markdown, with no reserved header/right clearance; on narrow/mobile layouts it can cover the first lines. [R3 / `packages/web/src/components/chat/ask-takeover.tsx:750`]

**Action taken**
- Submitted request changes on head `5616c0c7ec44e42a7c22ac95a85c1eee766de3b7`.
- I did not rerun code checks, tests, or live QA; CI is green, but the deterministic interaction blockers above need to be fixed first.

liuchao-001:
5094692 addresses all three findings from both reviews:

**[P1] Enter in Ask agent mode (yuezengwu 1 / baixiaohang ❌1)** — the window-level resolving shortcut now bails while `askAgentOpen`: `canReply` still reflects the hidden answer draft, so Enter must never fire `reply()` against an answer the user cannot see. Enter is a plain newline in the clarification composer; the footer button is the only send. Keyboard regression: valid hidden answer → enter mode → Enter in the clarification textarea → neither `onReply` nor `onAsk` fires.

**[P1] Stale queue continuation vs manual switch (yuezengwu 2 / baixiaohang ❌2)** — the continuation no longer trusts the snapshot captured at send time. Three live layers: it bails when the committed chat has moved on (`chatIdRef`), decides from the committed params (`latestSearchParamsRef`, same commit discipline as `chatIdRef`), and re-validates inside a functional `setSearchParams` updater — the updater sees the CURRENT params, leaves a foreign chat's URL untouched, and only acts on a still-live `nq` flag. Race regression: send held in flight → rail-like manual switch (new chat, `nq` deleted) → late settle → URL stays on the user's choice, no hijack, no resurrected flag.

**[P2] Button overlap (yuezengwu 3 / baixiaohang ⚠️3)** — Show earlier chat is now a real header row above the scroll region (reserving its own space) instead of an absolute overlay, so it can never cover the question's first lines at any width.

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2204 tests, all green.

yuezengwu (approved):
Re-reviewed the incremental fix commit. All three findings are resolved:

- Ask agent mode now gates the desktop Enter resolver even when a valid hidden answer draft exists, with a focused keyboard regression test.
- Queue continuation now verifies the committed chat and latest URL state, then re-validates inside the functional search-param update; the in-flight manual-switch regression covers the original race.
- Show earlier chat now occupies a real header row above the scroller, so it cannot cover the question body on narrow/mobile layouts.

The implementation preserves the request contract: clarification keeps the request open, ordinary navigation remains authoritative over the transient queue session, and inspection stays on the real chat surface. No database or core data-structure change is present. I did not rerun the reported checks or live QA per review scope.

baixiaohang (changes requested):
- Rationale: The Enter and layout fixes are complete, but the queue guard still relies on a stale functional search-param snapshot for same-chat exits.

**Risk level**: B-low
- Semantic lift: React Query-backed cross-chat queue continuation and navigation -> B-low

**PR summary**
- Problem: Unify Need you with the real chat takeover while preserving clarification, inspection, and cross-chat queue semantics.
- Follow-up approach: Gate Enter during clarification, move Show earlier chat into a real header row, and revalidate queue continuation against committed navigation state.

**Review findings**
❌ 1. The final functional `setSearchParams` check is not live in the pinned `react-router@7.13.1`: that setter closes over the `searchParams` from the render that created it and invokes the updater with a copy of that snapshot. The two `chatIdRef` checks cover switching to a different chat, but not an ordinary same-chat selection that only deletes `nq`. Repro shape: start resolving in chat A with `nq=1`; while the queue fetch is pending, click chat A's selected rail row (the normal selection callback keeps `c=A` and deletes `nq`); the committed chat id stays A, so line 2732 passes, then the stale setter at line 2733 receives the old `nq=1` params and can advance/resurrect the session. Re-read `latestSearchParamsRef.current` after the queue await and build the write from that live value (or use another truly live authority), and add the same-chat/flag-only exit regression. [R4 / `packages/web/src/pages/workspace/center/chat-view.tsx:2733`]

**Resolved from the prior review**
✅ Desktop Enter is gated while Ask agent mode is open.
✅ Show earlier chat now reserves a non-overlapping header row.

**Action taken**
- Submitted request changes on head `509469279c8fbe4199fd1c247c72cb86f3c65d52`.
- I did not run code checks, tests, or live QA.

yuezengwu (changes requested):
Correction to my prior approval: the remaining same-chat queue-exit race is valid and still blocks this head.

**[P1] Re-read the live params after the queue await; the functional updater is not live** (`packages/web/src/pages/workspace/center/chat-view.tsx:2733`). I verified the exact pinned `react-router@7.13.1` source: `useSearchParams` implements the callback form by calling `nextInit(new URLSearchParams(searchParams))`, where `searchParams` is the value captured by the render that created that setter. It is therefore not equivalent to a React state functional updater.

The two `chatIdRef` checks correctly stop a switch to another chat, but they cannot detect the ordinary same-chat selection path: while the queue fetch is pending, click the already-selected chat A row; `c` remains A and `nq` is deleted. When the fetch settles, both chat-id checks pass, and the captured setter supplies the old `nq=1` snapshot, allowing the queue to advance/resurrect despite the explicit exit. Re-read `latestSearchParamsRef.current` after the await and construct the write from that current value (with a final live guard), then add the flag-only same-chat regression.

The Enter-mode and non-overlapping header fixes remain resolved. I did not rerun checks, tests, or live QA.

liuchao-001:
698ff7d addresses the remaining blocker from both current-head reviews (baixiaohang ❌1 / yuezengwu's corrected verdict):

**Live params after the queue await.** Confirmed against the pinned `react-router@7.13.1`: the functional `setSearchParams` invokes the updater with `new URLSearchParams(searchParams)` from the render that created the setter — a snapshot copy, not live state. The continuation now re-reads `latestSearchParamsRef.current` (commit-synced) AFTER the queue await and builds the write from that value, with a final live `nq` guard: a same-chat exit that only deleted the flag while the fetch was pending can neither be advanced over nor have its session resurrected. The functional-updater form is gone. Regression covers the exact repro shape: queue fetch held in flight → same-chat flag-only exit → settle → URL untouched.

Also from the review agent's live-QA pass (its two blockers — Enter-in-mode and the cross-chat switch race — were fixed in 5094692, verified here on the new head):
- the Need you entry's accessible label now keys on the same predicate that enables it (`first`), so an items-empty/total>0 snapshot can't announce questions on a disabled control;
- the inspect composer's pending copy joins the English UI ("N pending questions" — pre-existing Chinese string, now on the main answering path).

`pnpm check` / `pnpm typecheck` / full web suite: 242 files, 2205 tests, all green.

yuezengwu (approved):
Re-reviewed head `698ff7de6210402bee9f8a82304b27652ad22521`. The remaining blocker is resolved against the actual `react-router@7.13.1` semantics:

- after the queue await, the continuation re-reads the commit-synced `latestSearchParamsRef.current` and constructs a concrete `URLSearchParams` write from that live value;
- the stale functional-updater form is gone;
- the new regression holds the queue fetch open, performs a same-chat flag-only `nq` exit, then verifies that late settlement neither navigates nor resurrects the session.

I also reviewed the entry-label and pending-question copy adjustments in this commit and found no new blocker. The earlier Enter-mode and header-overlap fixes remain sound. No database or core data-structure change is present. I did not rerun the reported checks/tests or live QA per review scope.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The live-URL race is now closed using the committed params ref after the queue await, with the exact same-chat flag-only regression.

**Risk level**: B-low
- Semantic lift: React Query-backed cross-chat queue continuation and navigation -> B-low

**PR summary**
- Problem: Humans need one coherent surface for answering agent questions, inspecting the underlying chat, asking for clarification, and continuing through cross-chat requests.
- Approach: Use the real chat takeover as the single answer surface, carry a transient Need you queue session across chats, make clarification a mutually exclusive mode, and retire the standalone/mobile review implementations.

**Review findings**
✅ The clarification composer cannot resolve a hidden answer through desktop Enter.
✅ The continuation now re-reads `latestSearchParamsRef.current` after the queue await and writes concrete params from that live value; both different-chat and same-chat/flag-only manual exits remain authoritative.
✅ Show earlier chat occupies a dedicated header row and cannot overlap the question body.
✅ The entry-label and pending-question copy follow-up introduces no new blocker.

**Action taken**
- Approved head `698ff7de6210402bee9f8a82304b27652ad22521`.
- I did not run code checks, tests, or live QA.

PR #2191, opened 2026-08-05
feat(context): add thin BYO runtime and org repository storage

baixiaohang:
## Summary

- Replace duplicated external Context workflows with a stable thin Claude/Codex adapter and an exact-release Core Skill loader.
- Add adapter-version separation, bounded lazy adapter synchronization, session-scoped Claude reload adoption, legacy read tombstones, and concise setup guidance.
- Add active-client-scoped BYO Context Tree storage at `data/byo/<org-id>`, with one per-org mutation lock, temporary read snapshots, durable write worktrees, crash recovery, and account switch/reset isolation.
- Keep Managed Skill delivery unchanged while routing the canonical Read/Write workflows by trusted `consumerKind`.

This prevents CLI upgrades from leaving external sessions on stale workflow copies, avoids unnecessary reload/trust for Core-only updates or additional org grants, and gives BYO Context Tree data an explicit local owner and recovery boundary.

Following architecture review, the adapter was also removed from unnecessary hot paths: a healthy Claude `UserPromptSubmit` hook is now a pure no-op, full payload health validation stops at the per-task route boundary, and opportunistic TTL directory scans were removed.

## Validation

- [ ] `pnpm check` — not run as a monolithic command; changed-file Biome and `git diff --check` passed.
- [x] `pnpm typecheck` — Shared, Client, CLI, and Web package typechecks passed; CLI typecheck was rerun after the final increment.
- [ ] `pnpm test` — focused affected suites passed; full repository test suite was not run.

Additional validation:

- Final affected and adjacent CLI suites: 17 files / 165 tests passed.
- Skill eval suites: 2 files / 11 tests passed.
- Shared schema: 7 tests passed.
- Web setup prompt: 16 tests passed.
- Existing Managed Skills/state regression suites: 116 tests passed.
- Shared, Client, CLI, and Web builds passed; the CLI build was rerun after the final increment.
- `pnpm validate:skill`: 8/8 passed.
- Release pack smoke passed, including exact-release loader and packed adapter/tombstone checks.
- `codex-reviewer` completed read-only review in the task chat and reported no remaining blockers.
- `codex-architect` completed a dedicated over-design review; its three requested simplifications are included in the latest head.

## Change Surface

- [x] `apps/cli` public CLI or help output
- [x] tree onboarding / binding / inspection behavior
- [x] shipped or planned skill topology
- [x] docs or contributor-facing repository metadata
- [x] CI / packaging / release plumbing

## Notes

[...]

PR #2196, opened 2026-08-05
fix(context): correct portable Core policy path

baixiaohang:
## Summary

- Emit the Context Policy path that matches the flattened portable app layout (`app/runtime-assets`) while preserving the npm package layout (`dist/runtime-assets`).
- Verify portable artifacts after extraction into the real immutable `versions/<version>` layout and execute the Core Skill loader from the installed shim.
- Add a regression test for the portable release manifest path.

## Root cause

PR #2191 made the loader consume a manifest-declared exact-release Policy path. The portable builder flattens `apps/cli/dist` into `app`, but the generated manifest still declared `dist/runtime-assets/context-tree-policy.md`, so SessionStart failed with `ENOENT` even though the Policy existed at `app/runtime-assets/context-tree-policy.md`.

## Validation

- Context bundle manifest + portable builder suites: 54/54 passed.
- CLI typecheck passed.
- Workspace build passed.
- A real linux-x64 portable artifact was built and verified; the verifier now executes `context skill load` from an extracted `versions/0.0.0/app` layout and asserts both Skill and Policy paths.
- Changed-file Biome and `git diff --check` passed.

### DIFF
--- a/apps/cli/src/__tests__/context-bundle-manifest.test.ts
+++ b/apps/cli/src/__tests__/context-bundle-manifest.test.ts
@@ -176,6 +176,32 @@ describe("context integration bundle", () => {
     );
   });
 
+  it("emits the flattened Core Policy path used by portable app artifacts", () => {
+    const root = mkdtempSync(join(tmpdir(), "first-tree-portable-context-bundle-"));
+    roots.push(root);
+    const repoRoot = resolve(import.meta.dirname, "../../../..");
+    execFileSync(
+      process.execPath,
+      [
+        join(repoRoot, "scripts", "build-context-integration-bundle.mjs"),
+        "--out-dir",
+        root,
+        "--version",
+        "1.2.3",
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
The goal is to restore the portable external-Context loader after the CLI dist tree is flattened into `app/`. The core changes parameterize the manifest's release-relative Policy path, keep the npm layout on `dist/runtime-assets/context-tree-policy.md`, emit `runtime-assets/context-tree-policy.md` only from the portable builder, and exercise the loader from an extracted `versions/<version>` artifact layout.

I traced both builder call paths through release-root resolution, containment checks, regular-file checks, and digest verification. The portable verifier now reaches the real loader command and asserts both canonical Skill and Policy paths. I found no blocking correctness or safety issues. There are no core data-structure or database changes in this diff.

Reviewed the diff only; I did not rerun tests or QA.

PR #2198, opened 2026-08-06
docs: make the chat description a current-state Summary contract

Gandy2025:
## Problem

The generated agent briefing defined a chat `description` as a "work summary and status report: background, plan, progress", and the CLI help, `docs/cli-reference.md`, the SDK JSDoc, and the shared schema comments each repeated some version of that definition. Agents wrote it accordingly — as a running work log.

Two costs fall out of that:

- A reader switching between chats gets stage history and process metadata (SHAs, test counts, reviewers, sub-agents, CI jobs) instead of a current judgment they can act on.
- Every *genuine* description change advances the conversation list's real-work recency signal, so process noise does not just dilute one summary; it reorders the list for everyone.

## Change

`Chat Topic & Description` in the briefing template becomes the single authoring contract for the Summary:

- **Shape** — rewritten from blank each time (never appended to); the first line stands alone with the current result or judgment plus what it means for the reader; then only what makes that line trustworthy — one most recent next step in flight, what it waits on when blocked, the conclusion and at most one deliverable when finished. 2–4 short sentences by default; 1500 chars is a hard ceiling, not a target.
- **Exclusions** — stage-by-stage history, plan/progress lists, restated messages, implementation detail, and process metadata. Human decisions still go to `chat ask`.
- **Timing** — on first understanding the goal, on substantive state change, on entering an external wait, and at the terminal state. The terminal rewrite is mandatory before the closing reply; routine tool/test steps earn no update.

Every other surface now points at that contract instead of restating a different one: `chat update` / `chat set-topic` help, `docs/cli-reference.md` (including the worked examples, which previously modelled a `**Plan:** … **Progress:** …` log), `FirstTreeHubSDK.updateChat` JSDoc, and the `updateChatSchema` / `meChatRowSchema` comments.

## Not changed

No data-model, server, or UI work. `description`, `descriptionUpdatedAt`, `activityAt`, `chat:updated`, the desktop Summary overlay, and the mobile Current state card already support this; a normalized lifecycle state would be designing V1 for a requirement we do not have.

## Verification

- `pnpm test` — all 12 turbo test tasks pass.
- `pnpm check && pnpm typecheck` — clean (the remaining Biome warnings are pre-existing and in untouched files).
- New `agent-briefing.test.ts` case asserts the contract's shape, exclusions, and update timing, and asserts the old `background + plan + progress` / `status report` wording is gone.
- The two briefing line budgets in that suite move by the 16 lines the contract adds (220 → 236 for `# Working in First Tree`, 580 → 596 total), with the rationale recorded inline.

Behavior here is agent judgment rather than deterministic code, so the matching manual case (`chat-update-summary-shape`) is added to `packages/skill-evals/docs/agent-briefing-manual-eval.md`.

A Context Tree PR correcting the matching decision node follows separately; the tree currently states that "progress and current state live in the description".


Gandy2025:
Context Tree counterpart (draft until this merges): agent-team-foundation/first-tree-context#897

yuezengwu (approved):

The change makes the generated briefing the normative Summary authoring contract and keeps the CLI help, reference docs, SDK/schema comments, regression coverage, and manual eval aligned with it. The diff is documentation/agent-guidance only: it does not change the chat data model, persistence, server behavior, or UI behavior.

I also verified that the conflicting Context Tree wording is covered by first-tree-context#897, which remains draft until this source PR merges.

baixiaohang (changes requested):
- Rationale: The new Summary contract is directionally right, but the worked examples violate its first-line invariant and the touched help/docs still describe a UI location that no longer exists.

**Risk level**: A
- Semantic lift: none; this changes agent authoring guidance rather than chat persistence or runtime state semantics

**PR summary**
- Problem: Agents currently turn the shared chat description into a process log, so people and agents switching among chats do not get a concise current judgment and noisy description writes disturb real-work recency.
- Approach: Make the generated briefing the normative current-state Summary contract, then align CLI/docs/SDK/schema guidance and add deterministic plus manual regression coverage.
- Impacted modules: CLI chat commands, generated client briefing, shared chat schemas, SDK docs, CLI reference, skill-eval guidance

**Review findings**
❌ 1. The new one-line examples do not satisfy the contract they are meant to teach. Both quoted `--description` values put the result and `Next:` in the same physical line; the desktop collapsed Summary calls `descriptionFirstLine()` and will therefore surface that whole value as the headline, even though the new contract requires the first line to stand alone as the result/judgment. Please either make these examples true multiline values (stdin/heredoc or equivalent) or keep the one-line example to the standalone headline only. [R1 / `docs/cli-reference.md:584`]

❌ 2. The touched help and reference still say the Summary is shown at the top of the right sidebar. The current Web implementation explicitly excludes `description` from that rail and renders `ChatSummary` between the header and message stream; mobile renders the Current state card in the timeline. Please describe the actual cross-surface placement in both locations so the newly canonical guidance is not coupled to a retired UI. [R5 / `apps/cli/src/commands/chat/update.ts:23`, `docs/cli-reference.md:575`]

⚠️ 3. The highest-salience per-turn runtime quick reference still says `chat update --description "<status>" — record status`, and the briefing's communication matrix still routes “Progress/status” to “Update status.” Those compact instructions coexist with the new “not a work log / no plan-progress lists” contract on every turn. Renaming the quick references to “refresh the current-state Summary” (and updating their assertions/manual row) would preserve the intended single source of truth instead of leaving a competing shorthand. [R1 / `packages/client/src/runtime/chat-context-section.ts:99`, `packages/client/src/runtime/templates/agent-briefing.ejs:89`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Revising my prior approval after validating the subsequent review feedback. These are blocking contract inconsistencies on the current head:

1. The two one-line `chat update --description` examples put `Next:` on the same physical line as the result. That violates the new standalone-first-line invariant and causes the collapsed desktop Summary preview to include both the judgment and next step. Use a true multiline example or keep the one-line value to the headline only.

2. The touched CLI help and reference still say the Summary appears at the top of the right sidebar. The current Web surface renders it between the header and message stream, while mobile uses the in-flow Current state card. Please update both descriptions to the actual cross-surface placement.

3. The per-turn runtime quick references still say `record status` / `Progress/status`. Since this PR establishes a current-state Summary rather than a work-log/status contract, those high-salience shorthands should be aligned as part of the same change.

Gandy2025:
Thanks — all three findings verified against the code and fixed in b29c0d5.

**1. One-line examples violate the standalone-first-line rule.** Confirmed: `descriptionFirstLine()` takes the first non-empty line, so a single-line value becomes the whole collapsed headline. Two fixes rather than one:

- The reference examples now keep a one-line value to the headline alone; the heredoc example already demonstrates the multiline shape with the next step on its own line.
- The contract itself was the root cause — "the first line stands alone" did not say *physical* line, so a 2–4 sentence paragraph technically satisfied it. It now says on its own physical line, and why (the collapsed bar previews only it). Asserted in `agent-briefing.test.ts`.

**2. Stale sidebar placement.** Confirmed — `chat-view.tsx` explicitly excludes `description` from the header/rail and renders `ChatSummary` between the header and the message stream. I inherited that wording and should have checked it while editing those lines. Both the CLI help and `docs/cli-reference.md` now describe the real cross-surface placement, including mobile's Current state card.

**3. Competing per-turn shorthand.** Agreed, and I had wrongly scoped this out as a routing label rather than a definition. `record status` and `Progress/status` are exactly the framing that produces work logs, and they ride every turn at higher salience than the contract. Both now read as refreshing the current-state Summary; `chat-context.test.ts` and the manual eval row are updated to match.

A follow-on Context Reviewer finding on the tree PR caught that my "collapsed views preview only that line" phrasing overclaimed for mobile, which clamps to four lines rather than one. That is corrected in the same push (and on first-tree-context#897, now clean).

Verification: `pnpm check` and `pnpm typecheck` pass. `pnpm test` is green except for two `context-tree-audit` fixture tests that time out at 25–30s under full parallel load on my machine and pass in isolation — untouched files, no relation to this diff. CI is the authority on the new head.

yuezengwu (approved):
Re-reviewed b29c0d5. All three requested changes are resolved:

- The one-line examples now contain only the standalone headline, the multiline example keeps the next step on its own physical line, and the briefing makes the physical-line invariant explicit.
- The CLI help and reference now describe the actual desktop and mobile Summary placements.
- The per-turn runtime contract, briefing communication matrix, tests, and manual eval now use current-state Summary wording instead of status/progress shorthand.

Non-blocking nit: the new comment in `packages/client/src/__tests__/agent-briefing.test.ts` still says the mobile card previews only `descriptionFirstLine()`. The actual mobile component renders the full short value or a four-line clamp; only the desktop collapsed bar uses `descriptionFirstLine()`. Please tidy that comment when convenient. It does not affect the assertion or shipped contract.

Approved — no remaining blocking findings. I did not rerun tests or QA as part of this diff-only review.

yuezengwu (approved):
Re-reviewed f2fc542. The only delta correctly fixes the test comment to distinguish the desktop collapsed bar's `descriptionFirstLine()` behavior from the mobile Current state card's full-value/four-line-clamp behavior.

No new findings. Approved on the latest head; tests and QA were not rerun for this comment-only delta.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All previously blocking Summary-contract inconsistencies are resolved on `f2fc542`; the generated briefing, per-turn runtime shorthand, CLI/docs examples, SDK/schema guidance, and desktop/mobile presentation descriptions are now aligned.

**Risk level**: A
- Semantic lift: none; this changes agent authoring guidance without changing chat persistence or runtime state semantics

**Review findings**
✅ The standalone headline is now explicitly a physical-line invariant, and the examples model it correctly.
✅ CLI help and reference documentation describe the actual desktop and mobile Summary surfaces.
✅ The high-salience runtime quick references now use current-state Summary wording rather than status/progress shorthand.
✅ The final test-comment correction accurately distinguishes desktop `descriptionFirstLine()` behavior from the mobile full-value/four-line clamp.

**Action taken**
- Approved with explicit maintainer authorization for the exact reviewed head `f2fc542`.

Issue #2004, opened 2026-07-24
[Bug] first-tree-read can emit contextDecision receipt in Chat body instead of metadata


yuezengwu:
## Summary

A Codex agent using the shipped `first-tree-read` 0.3.0 contract emitted the full `contextDecision` receipt as visible Chat markdown and omitted `metadata.contextDecision`. The Web did not expand hidden metadata; the agent constructed a malformed final `chat send`.

This pollutes the user-facing reply and means downstream receipt consumers cannot find the receipt even though a JSON block appears in the message body.

## Surface Area

- agent runtime (`codex`)
- Chat CLI/message transport
- Context Tree `first-tree-read` receipt flow
- Web Chat (visible symptom only; current Web code has no `contextDecision` renderer)

## Environment

- First Tree staging CLI: `0.5.17-staging.939.1` (captured after the incident)
- `first-tree-read`: `0.3.0`
- install method: managed First Tree staging client/workspace
- runtime: `codex`
- operating system: macOS 26.5.2, Darwin 25.5.0 arm64

## Reproduction

1. In a long-lived QA Chat, run several Codex turns/compactions while reading decision-bearing Context Tree nodes and preparing a final QA report.
2. Have the agent send the final report through `first-tree-staging chat send ... -F final-message.md`.
3. Inspect the persisted message with `first-tree-staging chat history <chat-id>`.
4. Observe that the receipt JSON is in `content`, while `metadata.contextDecision` is absent.

The sanitized runtime trace for the observed occurrence shows:

```text
apply_patch -> final-message.md contains:
  Context Tree receipt：
  { "contextDecision": { ... } }

first-tree-staging chat send gandy2025 -f markdown -F .../final-message.md
# --metadata was omitted
```

## Expected Behavior

When Tree context materially affects the final choice, the same final `chat send` or blocking `chat ask` should:

- keep receipt JSON out of the visible message body; and
- attach one top-level `metadata.contextDecision` object.

The receipt must never be a separate receipt message or a prose-only claim.

## Actual Behavior

Persisted message `019f923f-8202-7890-82dd-5f6ffb26b021` contains a visible `Context Tree receipt:` JSON block. Its metadata contains only the recipient mention and has no `contextDecision`.

An earlier message in the same Chat correctly used `-m/--metadata`, proving the CLI/Server metadata path itself works. This occurrence is an agent-side protocol/enforcement failure.

## First Tree Context

[...]

PR #1978, opened 2026-07-23
Record Context Tree decision influence receipts

Gandy2025:
## Summary

- teach `first-tree-read` to attach a `contextDecision` receipt only when a decision-bearing normal passage materially influences a later choice
- persist the receipt on the same final `chat send`, or the same blocking `chat ask` for an unresolved conflict
- record one mutually exclusive effect plus an exact remote-backed Context Tree repository, commit, node path, and optional heading
- anchor managed evidence to the briefing-declared binding repository and branch; missing or ambiguous identity fails closed
- exclude navigation-only reads, workflow instructions, ownership-only routing, supporting content alone, and Tree mentions without decision influence
- add deterministic floor coverage for the receipt JSON and managed provenance contracts, and bump the skill to `0.3.0`

## Scope

This intentionally reuses existing immutable message metadata and the existing `--metadata` option. It adds no database table, API, background job, or UI.

## Paired Context Tree contract

- agent-team-foundation/first-tree-context#811 (draft)

## Validation

- `python3 ./scripts/quick_validate_skill.py ./skills/first-tree-read`
- `pnpm --filter @first-tree/skill-evals eval:floor -- --suite first-tree-read` (2/2)
- `pnpm --dir packages/skill-evals exec vitest run src/suites/first-tree-read/__tests__/floor.test.ts` (6/6)
- complete `first-tree-read` test suite (24/24)
- `pnpm check`
- `pnpm typecheck`
- two independent reviews of the final diff, both with no blockers

`pnpm test` was also attempted. The full monorepo run failed in unrelated existing suites under local parallel load, primarily fixture/process timeouts in `@first-tree/skill-evals` plus unrelated client/server timeouts. The modified `first-tree-read` tests pass independently.

PR #2199, opened 2026-08-06
Show agent-reported Context Tree influence in Chat and Need you

Gandy2025:
## Why

`first-tree-read` (#1978) already instructs an agent to attach a `contextDecision` receipt to the same final `chat send` / blocking `chat ask` that carries the affected choice. Nothing consumed it: the receipt sat in free-form message metadata with no shared schema, no trust boundary, and no renderer — so the one place a user could see Context Tree paying off was invisible.

This lands the first consumer: the receipt is shown where the result is judged.

## What

**shared** — versioned `contextDecision` schema + strict `readContextDecisionMetadata`. Each evidence row must name a credential-free repository, a hex commit, and a tree-root-relative node path (1–3 rows, matching the skill's cap). A receipt with nothing to inspect is rejected; unknown-version / malformed payloads read back as `null`.

**server** — trust boundary in `preflightMessageSendIntent`:
- a human sender's `contextDecision` is stripped (the agent-attribution claim is not theirs to make, and a browser/API write must not be able to forge one);
- an agent sender's malformed receipt fails the write with a shape hint, so the author fixes and resends instead of silently storing something no consumer can render.

**web** — `ContextDecisionReceipt`, rendered in two places:
- under an agent's chat reply, after its body, images and attachments;
- inside the blocking ask, after the question body and before the answer controls (adds no request state, notification, or blocking condition).

Collapsed it shows the Context Tree mark, what team context did (`Conflict surfaced` / `Approach changed` / `Options narrowed` / `Direction supported`), the agent's one-sentence summary, and an `N team decisions` disclosure. Expanding reveals the cited nodes with repository and short commit, an exact-commit link when the forge is unambiguous (GitHub by host; GitLab only when it matches the connected instance), and the attribution note at the bottom.

## Trust framing

The module states the outcome objectively because that is what the agent reports, but it is not a First Tree verification: no success check, no "verified" copy, no system-card chrome. Attribution is weakened to a footnote *only* because the receipt is still attached to its authoring agent's message — a future cross-message digest (e.g. a Context page rollup) must re-state it up front. This is recorded in the component's contract comment.

Rendering is gated on a strict parse and never runs for a human speaker's row, so pre-guard history cannot surface a forged receipt.

## Verification

- `pnpm typecheck` (all packages) and `pnpm check` clean; `pnpm --filter @first-tree/web lint:tokens` clean.
- New tests: shared schema (12), server trust boundary (4), receipt component (9), and three ChatView DOM cases — agent reply renders it, human sender / malformed receipt render nothing, and the ask shows it before the answer input.
- `@first-tree/shared`, `@first-tree/server`, `@first-tree/web` suites pass. `@first-tree/client` passes standalone; two of its timeout-driven tests (`pi-handler`, `opencode-handler`) flake under a fully parallel `pnpm test` on this machine and are untouched by this change.
- Real browser at 1440px and 390px on `/preview/request-dock`: collapsed and expanded, all four effects, `scrollWidth === clientWidth` at both widths, 44px disclosure target.

## Follow-ups (not in scope)

- #2004 — an agent can still emit the receipt as visible body text; that is an agent-side transport regression, not a rendering one.
- Context page 7-day rollup of receipts, which needs its own server read model and permission story.

### DIFF
--- a/packages/shared/src/schemas/context-decision.ts
+++ b/packages/shared/src/schemas/context-decision.ts
@@ -1,0 +1,118 @@
+import { z } from "zod";
+import { canonicalGitRepoIdentity } from "../canonical-git-repo-url.js";
+
+/**
+ * `metadata.contextDecision` — an agent's self-attributed record that Context
+ * Tree content materially shaped the choice carried by THIS message.
+ *
+ * Written by the `first-tree-read` skill on the same final `chat send` (or
+ * blocking `chat ask`) that contains the affected choice, never as a separate
+ * message and never as prose. The receipt is the agent's own report: First Tree
+ * preserves the cited repository/commit/path so a reader can inspect the exact
+ * source, but it does NOT independently verify that the passage caused the
+ * choice. Every consumer must present it as agent-reported, never as a
+ * system-verified causal claim.
+ *
+ * Trust boundary: the server strips the key from human senders and rejects a
+ * malformed receipt from an agent sender, so a stored receipt is always an
+ * agent's and always parses. Readers still parse defensively (`safeParse` via
+ * `readContextDecisionMetadata`) because message rows are immutable — history
+ * written before that guard is not re-validated.
+ */
+export const CONTEXT_DECISION_METADATA_KEY = "contextDecision";
+
+/**
+ * How Tree content acted on the choice. Exactly one applies; the skill picks
+ * the first matching category in this precedence order so reports stay
+ * comparable:
+ *
+ *   - `conflicted`  — exposed a conflict that still needs resolution/escalation
+ *   - `redirected`  — changed the intended approach
+ *   - `constrained` — ruled out an option or narrowed the solution boundary
+ *   - `confirmed`   — removed material uncertainty and justified keeping the
+ *                     choice without changing its boundary
+ */
+export const contextDecisionEffectSchema = z.enum(["conflicted", "redirected", "constrained", "confirmed"]);
+export type ContextDecisionEffect = z.infer<typeof contextDecisionEffectSchema>;
+
+/** At most three node paths may jointly influence one choice (skill contract). */
+export const MAX_CONTEXT_DECISION_EVIDENCE = 3;
+export const MAX_CONTEXT_DECISION_SUMMARY_LENGTH = 400;
+
+/**
+ * One cited passage: the exact repository + commit + Tree-root-relative node
+ * path that supplied it. `repoUrl` is the credential-free binding repository as
+ * the activation receipt / workspace briefing declares it — never a local
+ * transport URL and never a credential-bearing remote, so a stored receipt can
+ * be shown and linked without leaking a secret.
+ */
+export const contextDecisionEvidenceSchema = z.object({
+  repoUrl: z
+    .string()
+    .min(1)
+    .max(2_000)
+    .refine((value) => canonicalGitRepoIdentity(value) !== null, {
+      message: "repoUrl must be a resolvable git repository URL",
+    })
+    .refine((value) => !hasEmbeddedCredentials(value), {
+      message: "repoUrl must not embed credentials",
+    }),
+  /**
+   * The exact commit the passage was read at. Full 40-char SHA-1 is what the
+   * skill specifies; shorter unambiguous prefixes are accepted rather than
+   * failing an otherwise valid final message, and 64 covers SHA-256 repos.
+   */
+  commit: z.string().regex(/^[0-9a-fA-F]{7,64}$/, "commit must be a hex git object id (7-64 chars)"),
+  /** Tree-root-relative path, e.g. `system/cloud/team/tenancy-and-identity.md`. */
+  nodePath: z
+    .string()
+    .min(1)
+    .max(1_000)
+    .refine((value) => !value.startsWith("/") && !value.split("/").includes(".."), {
+      message: "nodePath must be a tree-root-relative path without `..` segments",
+    }),
+  /** Optional heading inside the node; omitted when it cannot be named reliably. */
+  heading: z.string().min(1).max(200).optional(),
+});
+export type ContextDecisionEvidence = z.infer<typeof contextDecisionEvidenceSchema>;
+
+export const contextDecisionSchema = z.object({
+  version: z.literal(1),
+  effect: contextDecisionEffectSchema,
+  /** One concrete sentence naming what the Tree content actually did here. */
+  summary: z.string().trim().min(1).max(MAX_CONTEXT_DECISION_SUMMARY_LENGTH),
+  /**
+   * Required: a receipt with nothing to inspect is a claim, not a receipt —
+   * the inspectable source is the only part First Tree can vouch for.
+   */
+  evidence: z.array(contextDecisionEvidenceSchema).min(1).max(MAX_CONTEXT_DECISION_EVIDENCE),
+});
+export type ContextDecision = z.infer<typeof contextDecisionSchema>;
+
+/**
+ * Strict reader for a stored message's receipt. Returns `null` for absent,
+ * unknown-version, or malformed payloads so a renderer fails closed (shows
+ * nothing) instead of rendering a partial, misleading influence claim.
+ */
+export function readContextDecisionMetadata(
+  metadata: Record<string, unknown> | null | undefined,
+): ContextDecision | null {
+  if (metadata?.[CONTEXT_DECISION_METADATA_KEY] === undefined) return null;
+  const parsed = contextDecisionSchema.safeParse(metadata[CONTEXT_DECISION_METADATA_KEY]);
+  return parsed.success ? parsed.data : null;
+}
+
+/**
+ * A URL that carries `user:token@host` would be persisted, rendered, and linked
+ * verbatim. Reject it at the write boundary instead.
+ */
+function hasEmbeddedCredentials(value: string): boolean {
+  const trimmed = value.trim();
+  if (!trimmed.includes("://")) return false;
+  try {
+    const url = new URL(trimmed);
+    return url.username !== "" || url.password !== "";
+  } catch {
+    return false;
+  }
+}
--- a/packages/shared/src/schemas/context-decision.ts
+++ b/packages/shared/src/schemas/context-decision.ts
@@ -1,5 +1,5 @@
 import { z } from "zod";
-import { canonicalGitRepoIdentity } from "../canonical-git-repo-url.js";
+import { contextTreeRepoSchema } from "./org-settings.js";
 
 /**
  * `metadata.contextDecision` — an agent's self-attributed record that Context
@@ -38,6 +38,7 @@ export type ContextDecisionEffect = z.infer<typeof contextDecisionEffectSchema>;
 /** At most three node paths may jointly influence one choice (skill contract). */
 export const MAX_CONTEXT_DECISION_EVIDENCE = 3;
 export const MAX_CONTEXT_DECISION_SUMMARY_LENGTH = 400;
+export const MAX_CONTEXT_DECISION_REPO_URL_LENGTH = 2_000;
 
 /**
  * One cited passage: the exact repository + commit + Tree-root-relative node
@@ -47,22 +48,30 @@ export const MAX_CONTEXT_DECISION_SUMMARY_LENGTH = 400;
  * be shown and linked without leaking a secret.
  */
 export const contextDecisionEvidenceSchema = z.object({
-  repoUrl: z
-    .string()
-    .min(1)
-    .max(2_000)
-    .refine((value) => canonicalGitRepoIdentity(value) !== null, {
-      message: "repoUrl must be a resolvable git repository URL",
-    })
-    .refine((value) => !hasEmbeddedCredentials(value), {
-      message: "repoUrl must not embed credentials",
-    }),
   /**
-   * The exact commit the passage was read at. Full 40-char SHA-1 is what the
-   * skill specifies; shorter unambiguous prefixes are accepted rather than
-   * failing an otherwise valid final message, and 64 covers SHA-256 repos.
+   * The SAME contract the Context Tree binding itself is validated against.
+   * The producer copies the declared binding repository verbatim, so a second
+   * URL rule here could only drift from it: a stricter one rejects a valid
+   * binding (`ssh://git@host/path` carries a username by design) and a looser
+   * one admits shapes the binding would refuse. `contextTreeRepoSchema` already
+   * requires https / ssh / scp-like transport with a host and repository path,
+   * and rejects embedded passwords, HTTPS usernames, query and fragment
+   * components, control characters, and line separators — so a credential
+   * cannot ride any of the three forms into an immutable message row.
    */
-  commit: z.string().regex(/^[0-9a-fA-F]{7,64}$/, "commit must be a hex git object id (7-64 chars)"),
+  repoUrl: contextTreeRepoSchema.refine((value) => value.length <= MAX_CONTEXT_DECISION_REPO_URL_LENGTH, {
+    message: `repoUrl must be at most ${MAX_CONTEXT_DECISION_REPO_URL_LENGTH} characters`,
+  }),
+  /**
+   * The exact commit the passage was read at. A clone identity resolves an
+   * abbreviation against one repository at one moment, so only a full object id
+   * keeps naming the same immutable version for the life of the stored receipt.
+   * SHA-1 (40) and SHA-256 (64) are the two legal lengths; every producer path
+   * emits one because the receipt is built from `git rev-parse HEAD`.
+   */
+  commit: z
+    .string()
+    .regex(/^([0-9a-fA-F]{40}|[0-9a-fA-F]{64})$/, "commit must be a full SHA-1 (40) or SHA-256 (64) object id"),
   /** Tree-root-relative path, e.g. `system/cloud/team/tenancy-and-identity.md`. */
   nodePath: z
     .string()
@@ -100,19 +109,4 @@ export function readContextDecisionMetadata(
   if (metadata?.[CONTEXT_DECISION_METADATA_KEY] === undefined) return null;
   const parsed = contextDecisionSchema.safeParse(metadata[CONTEXT_DECISION_METADATA_KEY]);
   return parsed.success ? parsed.data : null;
-}
-
-/**
- * A URL that carries `user:token@host` would be persisted, rendered, and linked
- * verbatim. Reject it at the write boundary instead.
- */
-function hasEmbeddedCredentials(value: string): boolean {
-  const trimmed = value.trim();
-  if (!trimmed.includes("://")) return false;
-  try {
-    const url = new URL(trimmed);
-    return url.username !== "" || url.password !== "";
-  } catch {
-    return false;
-  }
 }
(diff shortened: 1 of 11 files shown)


Gandy2025:
Context Tree counterpart (draft until this merges): agent-team-foundation/first-tree-context#898

packages/web/src/pages/workspace/center/chat-view.tsx
```diff
@@ -1583,6 +1612,7 @@ const ChatTimeline = memo(function ChatTimeline({
                     agentAvatarFn={agentAvatarFn}
                     agentColorTokenFn={agentColorTokenFn}
                     mentionParticipants={mentionParticipants}
+                    senderIsHuman={humanParticipantIds.has(msg.senderId)}
```
yuezengwu: [P1] Fail closed when sender type is unknown. humanParticipantIds contains only current speakers. Participants can be removed, and chatDetail can load after cached messages, so a pre-guard human row with contextDecision is treated as non-human here and renders; the blocking-dock check has the same issue. That defeats the stated guarantee that forged historical human receipts never surface. Determine sender type from durable message/history data, or render only for a positively known agent sender and accept hiding unknown rows. Please add removed-human and loading-state regression coverage.

packages/shared/src/schemas/context-decision.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export const contextDecisionEvidenceSchema = z.object({
+  repoUrl: z
+    .string()
+    .min(1)
+    .max(2_000)
+    .refine((value) => canonicalGitRepoIdentity(value) !== null, {
+      message: "repoUrl must be a resolvable git repository URL",
+    })
+    .refine((value) => !hasEmbeddedCredentials(value), {
```
yuezengwu: [P2] Reject credential-bearing query and fragment forms. hasEmbeddedCredentials checks only URL username/password, so a value such as https://github.com/org/tree?access_token=secret passes canonicalization and this guard, then is persisted in immutable message metadata despite the credential-free contract. Repository clone identities do not need search or hash components; reject them and cover a query-token case.

packages/shared/src/schemas/context-decision.ts
```diff
@@ -0,0 +1,118 @@
...
+      message: "repoUrl must not embed credentials",
+    }),
+  /**
+   * The exact commit the passage was read at. Full 40-char SHA-1 is what the
+   * skill specifies; shorter unambiguous prefixes are accepted rather than
+   * failing an otherwise valid final message, and 64 covers SHA-256 repos.
+   */
+  commit: z.string().regex(/^[0-9a-fA-F]{7,64}$/, "commit must be a hex git object id (7-64 chars)"),
```
yuezengwu: [P2] Require an exact object ID rather than an abbreviation. The producer skill and receipt contract cite the exact commit, but this accepts 7-character prefixes and every intermediate hex length. A prefix can become ambiguous as the repository grows, making the evidence link no longer identify an immutable version. Valid producers already emit full IDs, so accept only full SHA-1 or SHA-256 lengths and test that short prefixes fail.

yuezengwu (changes requested):
The implementation goal and package split are clear: shared defines a versioned contextDecision receipt, server enforces sender attribution at write time, and Web renders the receipt beside the result or blocking question it explains. I am requesting changes for three trust and evidence-boundary gaps called out inline. This changes the core persistent message-metadata contract but adds no database migration. I reviewed the full diff and existing coverage without running tests or QA, per reviewer policy. The draft Context Tree counterpart in first-tree-context#898 matches the intended product boundary and should remain draft until the source fixes land and this PR merges.

baixiaohang (changes requested):
- Rationale: The receipt validator introduces a second repository-URL contract that disagrees with the existing Context Tree binding schema and breaks valid SSH-backed teams.

**Risk level**: B-high
- Path baseline: `packages/server/src/services/message.ts` is a message service change -> B-high.
- Semantic lift: none; the path baseline already covers the message trust-boundary change.

**PR summary**
- Problem: Agents can already report that team Context Tree decisions shaped a reply or blocking question, but users cannot see that influence where they judge the result.
- Approach: Add a shared receipt schema, enforce sender/shape rules at message preflight, and render the agent-reported receipt under replies and inside the blocking ask before answer controls.
- Impacted modules: `packages/shared`, `packages/server` message send preflight, and `packages/web` Chat / ask takeover.

**Review findings**
❌ 1. Reuse the Context Tree binding repository validator instead of introducing a conflicting URL contract. [`packages/shared/src/schemas/context-decision.ts:49`]
The producer must copy the binding repository exactly, and existing bindings explicitly accept `ssh://git@host/path`. `hasEmbeddedCredentials` rejects that valid credential-free SSH form because it rejects every URL username, so those teams cannot send a valid final receipt. Conversely, `canonicalGitRepoIdentity` is transport-agnostic, so this schema admits URL schemes and malformed binding shapes that `contextTreeRepoSchema` rejects. The existing inline finding about query/fragment credentials is another consequence of this drift. Reuse the existing binding URL schema/credential rules so the producer and consumer have one repository contract, with coverage for HTTPS, scp-like SSH, and `ssh://git@...`. [R1/R5]

✅ I agree with the existing fail-closed sender-type and exact-object-ID blockers and have not repeated them here.

✅ The placement itself matches the Chat contract: the receipt stays attached to the authoring message, and a blocking ask shows it before the answer controls without creating a second request state.

**Action taken**
- Submitted request changes.

packages/web/src/pages/workspace/center/chat-view.tsx
Gandy2025: Fixed in 3c7336338 — you are right, and the loading case is the worse half: cached messages paint before `chatDetail` resolves, so on a cold open *every* human row briefly passed the negative test.

The gate is now positive: `senderIsAgent` comes from a memoized set of non-human speakers, and the receipt renders only for a sender that set names. An unresolved sender — removed participant, or participants not loaded yet — renders nothing. The blocking-dock check now uses the same set in the same direction. The accepted cost is that a departed *agent*'s historical receipt is also hidden; that is the correct direction to fail.

Regression coverage added in `chat-view-dom.test.tsx`: `hides a receipt from a sender who is no longer a chat participant` and `hides a receipt while the participant list is still loading`. I verified both fail under the old negative gate (emulated through the same prop so every reference stays valid) and pass under the new one — the other three receipt cases stay green in both.

packages/shared/src/schemas/context-decision.ts
Gandy2025: Fixed in 3c7336338, and your finding turned out to be a symptom of a bigger one — @baixiaohang's review flagged that this schema should not have had its own repository-URL rule at all.

Both are now resolved the same way: `repoUrl` is `contextTreeRepoSchema`, the exact contract the Context Tree binding is validated against. It already rejects query and fragment components (`Context Tree repo must not include a query or fragment`), embedded passwords, HTTPS usernames, control characters, line separators, and non-https/ssh transports. My private guard is gone, so the producer and the consumer cannot drift again.

Verified concretely first: `https://github.com/org/tree?access_token=secret` canonicalizes to `github.com/org/tree`, so nothing upstream caught it and the raw value would have been persisted in an immutable row. Coverage now pins userinfo, HTTPS-username, query-token, fragment-token, and `ssh://git:token@` forms as rejected.

packages/shared/src/schemas/context-decision.ts
Gandy2025: Fixed in 3c7336338. Accepted — my 7–64 range was a deliberate leniency to avoid failing an otherwise good final message, but that trades a durable guarantee for a transient convenience: an abbreviation resolves against one repository at one moment, so the evidence link stops naming an immutable version as the repository grows. The receipt is supposed to be inspectable years later.

`commit` now accepts only `^([0-9a-fA-F]{40}|[0-9a-fA-F]{64})$` — full SHA-1 or SHA-256. Every producer path builds the receipt from `git rev-parse HEAD`, so no valid producer regresses. Coverage asserts a full SHA-256 id parses and that `main`, `abc1234`, 12-char, 39-char, and 42-char values all fail.

Gandy2025:
@baixiaohang — accepted in full, and thank you for catching the direction I had it backwards on. Fixed in 3c7336338.

You were right that the real defect was the *existence* of a second repository-URL contract, not any single rule inside it. I confirmed the break before fixing: `new URL('ssh://git@github.com/org/tree.git').username === 'git'`, so my `hasEmbeddedCredentials` rejected a perfectly valid credential-free SSH binding — an SSH-backed team could not have sent a single valid receipt. That is a hard break, not a strictness nit.

`repoUrl` is now `contextTreeRepoSchema` itself. One contract for producer and consumer: https / ssh / scp-like transport with a host and repository path; no embedded password, no HTTPS username, no query or fragment, no control characters or line separators. @yuezengwu's query/fragment finding falls out of the same change, exactly as you predicted.

Coverage for all three transports you named, as parse-success cases rather than shape assertions:

- `https://github.com/example/context-tree` and `…​.git`
- `git@github.com:example/context-tree.git` (scp-like)
- `ssh://git@github.com/example/context-tree.git`
- `ssh://git@gitlab.example.com:2222/group/sub/context-tree.git` (non-default port, nested groups)

plus negative coverage for `http://`, `git://`, `file://`, and the credential-bearing forms.

The two blockers you deferred to are also fixed in the same commit — full-object-id commits, and a positive `senderIsAgent` gate with removed-participant and participants-still-loading regression tests that I verified fail under the previous gate.

Local: `pnpm check`, `pnpm --filter @first-tree/web typecheck` (incl. design-token guardrail), and the shared / server / web suites all pass (859 / 3095 / 2223).

baixiaohang (changes requested):
- Rationale: The historical-sender guard is now positive at the set-membership layer, but the set itself still classifies unknown wire types as agents.

**Risk level**: B-high
- Path baseline: `packages/server/src/services/message.ts` remains part of the PR -> B-high.
- Semantic lift: none.

**PR summary**
- Problem: Make an agent's reported Context Tree influence visible beside the reply or blocking question it shaped, without allowing human-authored history to impersonate agent attribution.
- Approach: The follow-up correctly reuses the binding repository schema, requires full object IDs, and changes receipt rendering to depend on membership in a current-agent set.
- Impacted modules: shared receipt schema, server message preflight, and Web Chat / ask takeover.

**Review findings**
❌ 1. Make the agent set itself fail closed on unknown participant types. [`packages/web/src/pages/workspace/center/chat-view.tsx:2921`]
`chatParticipantDetailSchema.type` is deliberately `z.string()` so legacy or unrecognised wire values can flow through, and the Web `getChat` path is only statically typed rather than runtime-parsed. Therefore `p.type !== "human"` still admits missing, legacy, and future-unknown types into `agentParticipantIds`; under version skew, a cached pre-guard human row can again pass the receipt gate. This also contradicts the new comment's claim that the set contains positively known agents. Filter on `p.type === "agent"` and add an unknown/missing-type regression case so every unresolved type hides the receipt in both timeline and dock. [R4/R5]

✅ The repository validator, credential-shape coverage, full-object-ID rule, and removed/loading sender cases resolve the previous findings in the intended direction.

**Action taken**
- Submitted request changes on the new head.

yuezengwu (approved):
Re-reviewed exact head 9ca6af53a.

All requested trust/evidence fixes are complete:
- receipt rendering now requires a positively known non-human sender and fails closed for removed or not-yet-loaded participants;
- repoUrl reuses contextTreeRepoSchema, preserving valid HTTPS/SSH/scp bindings while rejecting credentials, query/fragment data, and unsupported transports;
- commit now requires a full 40- or 64-character object ID;
- regression coverage includes removed/loading senders, credential forms, exact object IDs, and SSH-form source links.

No new blockers found in the two incremental commits. I reviewed the diff and coverage only; no tests or QA were run locally.

yuezengwu (changes requested):
Correcting my approval on exact head 9ca6af53a: one fail-closed gap remains.

[P1] agentParticipantIds still uses p.type !== "human". ChatParticipantDetail.type intentionally accepts an arbitrary string for legacy/version-skew compatibility, so missing, legacy, or future-unknown types are positively admitted to the set and can render a pre-guard receipt as agent-authored. This contradicts the new known-agent contract and affects both timeline and dock through the shared set.

Filter with p.type === "agent" and add an unknown/missing-type regression case. Unknown sender types must hide the receipt.

The repository validator, full object ID, removed/loading sender fixes, and SSH source-link coverage remain good. I reviewed the diff only and did not run tests or QA.

Gandy2025:
@baixiaohang @yuezengwu — you both landed on the same gap independently, and it is a real one. Fixed in 9efee8e30.

I verified the premise before changing anything: `chatParticipantDetailSchema.type` is `z.string()` (the schema comments say the looseness is deliberate, so legacy / unrecognised DB values flow through), and `AGENT_TYPES` has exactly two members. So `p.type !== "human"` was still a negative test one layer down — I had moved the gate from the call site into the set and left the same hole inside it. Under version skew a legacy or missing type would have been admitted as an agent and rendered a pre-guard receipt, in both the timeline and the dock, since they share the set.

`agentParticipantIds` now filters on `p.type === AGENT_TYPES.AGENT`. Every unresolved type — missing, empty, legacy, future-unknown — hides the receipt. The comment that claimed "positively known agents" is now true rather than aspirational. The accepted cost stays the same and stays in the safe direction: a real agent whose type does not resolve loses its receipt display.

Regression case `hides a receipt from a sender whose participant type is unknown, in both timeline and ask` puts a legacy-typed (`autonomous_agent`) participant and an empty-typed participant in one chat, with the legacy one posting a timeline reply and the untyped one raising the blocking ask. It asserts no receipt renders anywhere. I confirmed it fails under `!== "human"` and passes under `=== "agent"`, with the other five receipt cases green in both.

Local on this head: `pnpm check` clean, `pnpm --filter @first-tree/web typecheck` clean (incl. design-token guardrail), web suite 2225 passing. CI on the previous head 9ca6af53a was fully green (15 pass / 2 skip); re-running now on this one.

yuezengwu (approved):
Re-reviewed exact head 9efee8e30.

The remaining fail-closed blocker is resolved: agentParticipantIds now admits only participants positively typed as AGENT_TYPES.AGENT, so missing, empty, legacy, and future-unknown wire types cannot render a receipt in either timeline or blocking ask. The new regression case covers both surfaces.

All earlier trust/evidence fixes remain sound. No new blockers found in this increment. I reviewed the diff and coverage only; no tests or QA were run locally.

baixiaohang (changes requested):
- Rationale: The hard-fail policy is appropriate for an atomically authored receipt, but the successful path persists the raw payload rather than the schema-parsed receipt, so the claimed immutable shape/size boundary is not actually enforced.

**Risk level**: B-high
- Path baseline: `packages/server/src/services/message.ts` changes the central message send trust boundary and persistent metadata contract.
- Semantic lift: none.

**PR summary**
- Problem: Users cannot currently see when an agent reports that team Context Tree decisions shaped a reply or blocking question.
- Approach: Add a versioned receipt, validate attribution at the central send preflight, and render valid agent-authored receipts beside the affected message or ask.
- Impacted modules: shared receipt schema, server message preflight, and Web Chat / ask rendering.

**Review findings**
❌ 1. Persist the parsed receipt rather than the original unchecked object. [`packages/server/src/services/message.ts:144`]
`contextDecisionSchema.safeParse` produces the canonical value, but the success branch discards `parsed.data` and returns the original `meta`. Zod object parsing strips unknown nested keys and `summary.trim()` transforms the value, so an input with all valid required fields plus an extra credential-bearing field — or a summary padded far beyond the stored 400-character contract — passes validation and is then persisted verbatim in immutable message metadata; readers later discard those values, hiding the drift. Replace only `metadata.contextDecision` with `parsed.data` (or reject unknown keys strictly) and add coverage that the preflight result cannot retain nested extras or pre-normalized padding. [R4/R5]

✅ Rejecting the whole new send on a genuinely malformed authored receipt is reasonable here: body and receipt are one atomic authored message, the failure happens before any durable write, and the CLI exposes the correction hint so retry is safe. Silently stripping an agent receipt would instead report success after persisting a different authored result. Future receipt versions must follow an expand-first rollout: Server accepts the new version before any producer emits it.

**Action taken**
- Submitted request changes on exact head `9efee8e30049ae41c068461b8305285803a3d249`; reviewed diff and call paths only, without running tests or QA.

yuezengwu (changes requested):
Correcting my approval on exact head 9efee8e30: one persistence-boundary blocker remains.

[P1] applyContextDecisionTrustBoundary validates with contextDecisionSchema.safeParse but discards parsed.data and returns the original meta. Zod strips unknown object keys and summary.trim() normalizes before the 400-character check, so a receipt can pass while immutable message metadata retains hidden nested extras or arbitrarily padded raw summary text. The defensive reader later drops/normalizes them, masking the stored contract drift.

On success, replace only metadata.contextDecision with parsed.data, preserving unrelated metadata, and add preflight coverage proving nested extras and pre-normalized padding cannot survive the returned metadata.

All previous sender-type, repository, object-ID, and source-link fixes remain sound. I reviewed the full success path and existing coverage only; no tests or QA were run locally.

Gandy2025:
@baixiaohang @yuezengwu — correct, and fixed in 8b8032aee. I had treated validation as if it were normalization; it is not, and the gap was invisible precisely because every reader re-parses.

Reproduced first, against the shipped schema, so the fix targets the real behaviour rather than my reading of it:

```
input  summary length 403 (padded), + top-level `secretToken`, + nested `accessToken` on a valid evidence row
parses: true
parsed summary length: 398        // .trim() ran before the 400 bound
parsed keeps top-level extra: false
parsed keeps nested extra:   false
```

So a receipt with a credential sitting in an unknown nested field passed the gate and would have been persisted verbatim in an immutable row, while the UI showed a clean parsed view. That is the worst shape for this defect: the stored value and the displayed value disagree, and nothing surfaces the difference.

The success branch now returns `{ ...meta, [CONTEXT_DECISION_METADATA_KEY]: parsed.data }` — only the receipt key is replaced, so `mentions`, `attachments`, and every other metadata key are untouched.

Coverage (`persists the parsed receipt, not the caller's raw object`) sends exactly the payload above and asserts the returned metadata carries precisely `["effect","evidence","summary","version"]`, a summary of 398 characters, and an evidence row with precisely `["commit","heading","nodePath","repoUrl"]`. I confirmed it fails when the raw `meta` is returned and passes with `parsed.data`.

On your rollout note: agreed, and I read it as binding on the next version rather than advice — a `version: 2` receipt must be accepted by a deployed Server before any producer is allowed to emit one. Worth recording in the Tree node alongside the rest of the receipt contract; I will fold it into #898 when I reconcile that PR after this merges, rather than leaving it only in this thread.

Local on this head: `pnpm check` clean, server typecheck clean, server suite 3096 passing. CI re-running.

yuezengwu (approved):
Re-reviewed exact head 8b8032aee.

The persistence-boundary blocker is resolved: the successful preflight path replaces only metadata.contextDecision with parsed.data, preserving unrelated metadata while ensuring the append-only row stores the same canonical shape readers consume. The regression case covers top-level and nested extras plus pre-normalized summary padding.

All earlier trust/evidence fixes remain sound. No new blockers found in this increment. I reviewed the diff and coverage only; no tests or QA were run locally.

baixiaohang (approved):
Re-reviewed exact head `8b8032aee`.

All requested trust, evidence, sender-attribution, and persistence-boundary fixes are complete. The receipt now uses the existing Context Tree repository contract, requires full object IDs, renders only for positively known agent senders, and persists the schema-parsed canonical value while preserving unrelated message metadata.

No blockers remain, and the current required checks are green. I reviewed the diff and coverage only; I did not run tests or QA locally.

PR #2206, opened 2026-08-06
Require verified GitHub identity before App install

Gandy2025:
## Summary

- require an active Team admin to link a GitHub identity before minting GitHub App install state
- add a two-phase OAuth identity preflight that verifies the active github.com numeric ID before opening the App picker, then rechecks live Team and identity authority on callback
- keep Google/OIDC sign-in valid while adding an inline, recoverable GitHub capability gate and safe return to the originating Team

## Security and recovery

- missing identities fail with `409 github_identity_required` before state or nonce side effects
- account-link return paths are restricted to supported internal Settings paths
- mismatched GitHub accounts, role loss, identity unlink, cancel, expiry, replay, and provider configuration failures fail closed without switching the First Tree user or persisting a foreign identity
- no-code owner-approval landings recheck current Team-admin membership and the linked GitHub identity before returning to the waiting panel
- install-attempt markers are Team-scoped; full-page returns restore the originating Team, while failed restoration cannot apply another Team's marker
- existing explicit Connect and one-installation-per-Team constraints remain unchanged

## Provider limitation

GitHub does not include an OAuth code or verifiable requester identity when a non-owner installation request lands after the picker. A user who deliberately switches github.com accounts in another tab after passing the preflight can still cause a safe unbound installation row. This cannot bind the installation or cross a First Tree Team authority boundary. The limitation is documented in the QA case and deterministic tests; this PR does not add a Stage 2 handoff or support-override flow.

## Verification

- `pnpm --filter @first-tree/server exec vitest run ...` — 71 focused tests passed
- `pnpm --filter @first-tree/server exec vitest run --reporter=dot --silent` — 3,112 tests passed in independent QA
- `pnpm --filter @first-tree/web exec vitest run ...` — 84 focused tests passed
- `pnpm --filter @first-tree/web exec vitest run src --reporter=dot --silent` — 2,241 tests passed in independent QA
- `pnpm check` — passed with existing repository warnings only
- `pnpm typecheck`
- `pnpm build`
- `pnpm --filter @first-tree/web build`
- `npx momentic@3.42.0 lint`
- exact-head Momentic `local`, 13/13 steps, uploaded by implementation verification: https://app.momentic.ai/run-groups/4221e204-1a31-4e3f-91c7-8c42c3392d46
- independent QA: `PASS` at exact commit `91bd94be92fdc0a409fd2b84da3d110de724d9b6`; uploaded Momentic run group: https://app.momentic.ai/run-groups/18e2d11b-b57f-4265-87b9-68b3664b614e
- independent Standards and Spec reviews: passed after resolving cross-Team return and marker-scoping blockers

The real github.com OAuth, App picker, and owner-approval UI were not driven with external credentials. Deterministic server integration tests cover nonce rotation, numeric-ID match/mismatch, role loss, identity unlink, no-code landing authority, and side-effect boundaries.

A full monorepo `pnpm test` run exposed three unrelated parallel-only failures outside this diff. The affected client file passed 36/36 and the affected CLI file passed 5/5 when rerun in isolation.

### DIFF
--- a/packages/server/src/api/auth/github.ts
+++ b/packages/server/src/api/auth/github.ts
@@ -17,6 +17,7 @@ import {
   IdentityMismatchError,
   LastIdentityError,
   linkExternalIdentity,
+  refreshGithubInstallIdentity,
   unlinkExternalIdentity,
 } from "../../services/auth-identity.js";
 import { encryptValue } from "../../services/crypto.js";
@@ -84,7 +85,7 @@ export async function githubOauthRoutes(app: FastifyInstance): Promise<void> {
     );
   }
 
-  app.get("/start", async (request, reply) => {
+  app.get("/start", { config: { rateLimit: { max: 60, timeWindow: "1 minute" } } }, async (request, reply) => {
     const { next } = githubStartQuerySchema.parse(request.query);
     const safeNext = safeRedirectPath(next ?? null);
     if (!appCfg) {
@@ -115,7 +116,7 @@ export async function githubOauthRoutes(app: FastifyInstance): Promise<void> {
     return reply.redirect(buildAppAuthorizeUrl({ clientId: appCfg.clientId, redirectUri, state: token }), 302);
   });
 
-  app.get("/callback", async (request, reply) => {
+  app.get("/callback", { config: { rateLimit: { max: 60, timeWindow: "1 minute" } } }, async (request, reply) => {
     if (!appCfg) {
       return reply.status(503).send({ error: "GitHub App is not configured on this First Tree deployment" });
     }
@@ -192,6 +193,26 @@ export async function githubOauthRoutes(app: FastifyInstance): Promise<void> {
         "OAuth state provider does not match callback",
       );
       return redirectCallbackError(reply, "state-expired", next, { callbackIntent: intent });
+    }
+
+    // App-install callbacks are allowed to mutate only the kickoff user's
+    // already-linked GitHub identity. States minted before the complete
+    // two-phase contract cannot prove that user, so fail before exchanging a
+    // code or entering the generic account bootstrap. Users can recover by
+    // starting Install again from the Team's GitHub Settings panel.
+    if (
+      intent === "install" &&
+      (verified.intent !== "install" ||
+        verified.provider !== "github" ||
+        !targetOrganizationId ||
+        !kickoffUserId ||
+        !installPhase)
+    ) {
+      app.log.warn(
+        { event: "github_app.install_callback_incomplete_state", installPhase },
+        "install callback state lacks the verified kickoff authority tuple",
+      );
+      return redirectCallbackError(reply, "state-expired", "/settings/github", { callbackIntent: "install" });
     }
 
     if (providerError) {
@@ -282,7 +303,7 @@ export async function githubOauthRoutes(app: FastifyInstance): Promise<void> {
       installationIdRaw && Number.isFinite(Number(installationIdRaw)) ? Number(installationIdRaw) : null;
 
     if (intent === "install" && installPhase === "identity") {
-      return continueInstallAfterIdentityPreflight(app, reply, profile, next, {
+      return continueInstallAfterIdentityPreflight(app, reply, profile, tokens, next, {
         targetOrganizationId,
         kickoffUserId,
       });
@@ -578,6 +599,7 @@ async function continueInstallAfterIdentityPreflight(
   app: FastifyInstance,
   reply: FastifyReply,
   profile: GithubProfile,
+  oauthTokens: GithubTokenBundle,
   next: string,
   opts: { targetOrganizationId: string | null; kickoffUserId: string | null },
 ) {
@@ -586,23 +608,29 @@ async function continueInstallAfterIdentityPreflight(
     return redirectCallbackError(reply, "state-expired", "/settings/github", { callbackIntent: "install" });
   }
 
-  const installAuthority = await inspectInstallCallbackAuthority(app, profile, {
-    targetOrganizationId,
-    kickoffUserId,
+  const installAuthority = await refreshGithubInstallIdentity(app.db, {
+    userId: kickoffUserId,
+    organizationId: targetOrganizationId,
+    profile,
+    tokens: oauthTokens,
   });
   if (!installAuthority.ok) {
-    return redirectCallbackError(reply, installAuthority.code, "/settings/github", {
-      callbackIntent: "install",
-      expectedGithubLogin: installAuthority.expectedGithubLogin,
-    });
+    return redirectCallbackError(
+      reply,
+      installAuthority.reason === "not-admin" ? "install-not-admin" : "install-not-verified",
+      "/settings/github",
+      {
+        callbackIntent: "install",
+        expectedGithubLogin: installAuthority.expectedGithubLogin,
+      },
+    );
   }
-  const { linkedIdentity } = installAuthority;
 
   const appCfg = app.config.oauth?.githubApp;
   if (!appCfg?.slug) {
     return redirectCallbackError(reply, "provider-not-configured", "/settings/github", {
       callbackIntent: "install",
-      expectedGithubLogin: linkedIdentity.login,
+      expectedGithubLogin: installAuthority.githubLogin,
     });
   }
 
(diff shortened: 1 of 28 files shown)


packages/server/src/api/auth/github.ts
```diff
@@ -245,6 +281,32 @@ export async function githubOauthRoutes(app: FastifyInstance): Promise<void> {
...
+    }
+
+    // Re-check the same identity boundary after code-bearing picker callbacks.
+    // A user can pass the preflight, then switch github.com accounts in another
+    // tab before completing a direct install. Reject that race before
+    // findOrCreate can persist the foreign identity or replace the First Tree
+    // browser session. The no-code owner-approval limitation is documented in
+    // the branch above.
+    if (intent === "install" && targetOrganizationId && kickoffUserId) {
+      const installAuthority = await inspectInstallCallbackAuthority(app, profile, {
```
yuezengwu, line 319: [blocker] This reopens a TOCTOU window after `inspectInstallCallbackAuthority`. If the kickoff user unlinks GitHub (or replaces that provider identity) after the precheck but before this call, `findOrCreateGithubAccount` sees the verified numeric ID as unclaimed and creates a new First Tree user plus `auth_identities` row. Only afterward does the `bindAuthorityUserId !== userId` branch return `install-not-verified`, so the rejected callback has already persisted the identity and can prevent the original user from relinking it. That contradicts this PR's fail-closed guarantee for identity unlink/mismatch. Please keep install completion on `kickoffUserId` and make the live admin + exact provider-subject check and token/snapshot update atomic (or use an equivalent conditional update that cannot create an account), with a concurrency regression for unlink/replace between verification and persistence.

yuezengwu (changes requested):
The two-phase preflight, live Team-admin checks, numeric GitHub identity comparison, and Team-scoped return handling implement the intended capability gate. I found one blocking TOCTOU issue in the code-bearing installation callback: the verified kickoff identity can be unlinked or replaced before the generic account bootstrap mutates identity state. Please make the install completion operate on the already-verified kickoff user with an atomic conditional update/recheck, rather than using a find-or-create path. No database schema or migration changes are present; the core OAuth state payload does add the new `installPhase` discriminator.

baixiaohang (changes requested):
- Rationale: The capability gate is directionally correct, but the code-bearing install callback can still mutate identity state after its verification has gone stale, so rejected mismatch/unlink paths are not side-effect-free.

**Risk level**: B-low
- Path baseline: `packages/server/**` without a schema/migration or one of the protected core service paths -> B-low.
- Semantic lift: the Web API/data layer is touched, but that does not lift the mixed PR above B-low; no database shape or JSON metadata contract changes are present.

**PR summary**
- Problem: A Google/OIDC-authenticated Team admin could start GitHub App installation without first linking and proving the same GitHub identity, making wrong-account installs and recovery ambiguous.
- Approach: Add an inline GitHub-link capability gate, a two-phase OAuth identity preflight before the App picker, live Team-admin and numeric-ID checks on return, and Team-scoped browser return markers while preserving explicit installation binding.
- Impacted modules: Server GitHub OAuth/org integration routes and OAuth state; Web GitHub Settings/install return flow; deterministic tests and cross-surface QA assets.

**Review findings**
❌ 1. The existing TOCTOU blocker is valid: `inspectInstallCallbackAuthority` checks the kickoff user's current identity, but `completeOauthFlow` later calls `findOrCreateGithubAccount` in a separate operation. An unlink/replace between those points can let the generic bootstrap create a new First Tree user and claim the verified GitHub subject before the mismatch branch rejects the callback. Keep install completion pinned to `kickoffUserId` and make the exact-subject recheck plus token/snapshot update conditional/atomic so an install callback can never create an account. Add a concurrency regression around unlink/replace between verification and persistence. [R4/R5 / `packages/server/src/api/auth/github.ts:297`, `packages/server/src/api/auth/github.ts:728`]

⚠️ 2. The identity-phase callback exchanges the OAuth code for a fresh GitHub App user/refresh token bundle, then returns to the picker without persisting or otherwise reconciling that bundle. Canceled and no-code owner-approval retries can therefore issue additional tokens while the DB keeps an older pair; GitHub documents a ten-token per user/application/scope limit after which the oldest token is revoked. Fold this token update into the same exact-subject conditional operation above (or avoid issuing a user token for the preflight if GitHub offers a suitable proof path), so recovery attempts cannot silently age out the credential First Tree still stores. [R1/R4 / `packages/server/src/api/auth/github.ts:264`, `packages/server/src/api/auth/github.ts:284`; [GitHub token revocation](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/token-expiration-and-revocation)]

✅ 3. The route-level gate now checks current Team-admin authority before state issuance, rotates a protected nonce between phases, compares stable numeric GitHub IDs before the picker and on code-bearing return, and scopes browser restoration markers to the originating Team.

**Action taken**
- Submitted request changes; no code or tests were run as part of this review.

packages/server/src/api/auth/github.ts
Gandy2025: Fixed in 9ab49b0a2. Modern install callbacks no longer enter findOrCreateGithubAccount: completion stays pinned to kickoffUserId and uses one transaction to lock/re-check the user, live active-admin membership, and exact GitHub provider subject before conditionally refreshing the credential snapshot. Identity or role changes between the early inspection and persistence now fail without creating a user or identity, issuing a First Tree session, or binding an installation. Incomplete legacy install states also fail before OAuth code exchange/account bootstrap. The deterministic regressions simulate both identity replacement and role loss at this seam and assert zero persistent side effects. The same exact-subject operation now persists the fresh preflight access/refresh token bundle.

yuezengwu (approved):
Re-reviewed exact head `9ab49b0a289a94cc1856b42fc7973e802d8ae753`.

The previous blocker is resolved. Modern install callbacks now reject incomplete legacy authority state before code exchange, and both install phases stay pinned to `kickoffUserId`. `refreshGithubInstallIdentity` locks and re-checks the existing user, live active-admin membership, and exact GitHub numeric subject in one transaction before conditionally refreshing the credential snapshot; the install path no longer enters account creation. This also persists the fresh identity-preflight access/refresh token bundle instead of aging the stored credential through retries.

The added seam regressions cover identity replacement and role loss without creating a user/identity, issuing a First Tree session, or binding an installation. I found no new blocking issues. No tests or QA were run as part of this review.

There are no database schema or migration changes. The core auth-identity transaction boundary and OAuth state contract were reviewed explicitly.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The update resolves the stale-authority mutation window and persists preflight credentials only through the same live Team-admin and exact-subject guard, with no remaining blocking issue found.

**Risk level**: B-low
- Path baseline: `packages/server/**` is changed without a database schema/migration change or a protected core service path change -> B-low.
- Semantic lift: None; the OAuth state discriminator and auth-identity transaction boundary were reviewed explicitly, with no database shape or JSON metadata contract change.

**PR summary**
- Problem: A Team admin signing in through Google/OIDC could start GitHub App installation without first proving that the active github.com account matched the GitHub identity linked to the First Tree user.
- Approach: Gate installation behind a two-phase GitHub identity preflight, re-check live Team-admin and exact numeric-subject authority on return, and keep installation binding as an explicit Team Settings action.
- Impacted modules: Server GitHub OAuth and App-install routes, auth-identity persistence, Web GitHub Settings return handling, and deterministic regression coverage.

**Review findings**
✅ 1. Modern install completion is pinned to `kickoffUserId`; the existing user, active-admin membership, exact GitHub subject, and credential refresh are locked and checked in one transaction, and the install path cannot enter account creation. [R4 / `packages/server/src/services/auth-identity.ts:269`, `packages/server/src/api/auth/github.ts:757`]

✅ 2. The identity preflight now persists the fresh access/refresh token bundle through that same exact-subject conditional operation, avoiding retries that silently age out the stored credential. [R1/R4 / `packages/server/src/api/auth/github.ts:605`]

✅ 3. Incomplete legacy install states fail before OAuth code exchange, while the seam regressions cover both identity replacement and role loss without creating a user/identity, issuing a First Tree session, or binding an installation. [R4/R5 / `packages/server/src/api/auth/github.ts:198`, `packages/server/src/__tests__/github-app-callback-target-org.test.ts:634`]

**Action taken**
- Approved exact head `9ab49b0a289a94cc1856b42fc7973e802d8ae753`. No code checks, tests, or QA were run as part of this review.

Gandy2025:
Exact-head validation update for `9ab49b0a289a94cc1856b42fc7973e802d8ae753`:

- both independent Standards and Spec re-reviews passed; the previous TOCTOU and credential-refresh blockers are resolved
- focused Server/Web regressions, repository checks, typecheck, and builds passed locally
- required GitHub checks are green, including CodeQL and the full Server, Client/Web, and CLI jobs
- Momentic `local` passed the committed GitHub identity-gate journey in one attempt (13/13 top-level steps) and uploaded results: https://app.momentic.ai/run-groups/fddc5e23-8ec4-451d-bd7a-b1514848aba9

The uploaded Momentic metadata records this exact commit. Real github.com OAuth, picker, and owner approval remain represented by deterministic integration coverage rather than external provider credentials. Independent QA for this new head is in progress; the earlier QA PASS applied only to `91bd94b`.

Gandy2025:
Independent QA completed after merge for the exact PR head `9ab49b0a289a94cc1856b42fc7973e802d8ae753`: **PASS** (`focused-local`).

- direct security regressions: 22/22 passed, including identity replacement and admin-role loss at the inspection/persistence seam with zero rejected-path side effects
- focused Server authorization: 76/76; adjacent Server identity/OAuth/install/binding: 60/60
- full Server: 3,117/3,117; focused Web: 84/84; full Web source: 2,241/2,241
- check, typecheck, migrations, shared/Server/Web builds, and Momentic lint passed
- live rate-limit probe: requests 1–60 returned `302`; request 61 returned `429`
- exact-head Momentic `local --upload-results`: 13/13 passed — https://app.momentic.ai/run-groups/c795d19a-87f4-466d-9373-6780b45c7eda

No real github.com OAuth, picker, account switch, or owner-approval UI was driven; deterministic integration tests cover First Tree's reachable authorization and side-effect boundaries. The accepted post-preflight cross-tab account-switch provider limitation remains explicitly outside this QA claim and is tracked in #1392.

QA used an isolated disposable database and mock GitHub App configuration, did not access production, and completed cleanup.

PR #2211, opened 2026-08-06
refactor(client): S1b remove process-global handler registry

yuezengwu:
## Summary
- Delete process-global `HANDLER_REGISTRY` / `registerHandler` / `getHandlerFactory` / `hasHandler` and remove `registerBuiltinHandlers` / `handlers/index.ts`.
- CLI `ClientRuntime` creates and holds a frozen built-in factory table via `createBuiltinHandlerRegistry` (one cheap Claude resolve+log per construction); standalone `AgentRuntime` requires an explicit readonly `handlerFactories` map.
- Preserve unsupported-runtime warning/skip, Claude cheap resolution logging, AgentSlot single-factory injection, and lifecycle behavior. No SessionManager / package-move / provider-behavior changes.

This is slice **S1b** of the Client multi-package refactor. Keep Draft until independent review + exact-head QA and human stage confirmation.

## Test plan
- [x] Focused Client: builtin-registry, builtin-handlers, runtime, barrel-exports, provider-boundary-guard, handler-registry
- [x] Focused CLI: client-runtime-context-tree (unsupported runtime + pinned switch)
- [x] Client full suite / typecheck / build
- [x] CLI typecheck / build
- [x] Root `pnpm check`
- [x] Production + dist search: no global registry symbols; CLI imports only `@first-tree/client` package entry
- [ ] Independent review + exact-head QA

## Out of scope
SessionManager/SessionRegistry, package moves, provider factory/probe/auth tables, AgentSlot API beyond receiving a single factory.


baixiaohang (approved):
**Recommendation**: approve
- Rationale: Instance-owned factory tables remove cross-runtime process state without changing handler lifecycle, unsupported-runtime fail-closed behavior, or Claude executable resolution semantics.

**Risk level**: A
- Path baseline: Includes `apps/cli/**` and `packages/client/**` → A.
- Semantic lift: None; A is already the highest level.

**PR summary**
- Problem: Multiple Client or standalone Runtime instances in one process shared handler-registration state, allowing construction order or custom registration to leak factory choices across instances and obstruct the planned package boundary.
- Approach: Give each CLI `ClientRuntime` its own frozen, exhaustive built-in factory table; require standalone `AgentRuntime` callers to inject a readonly factory map; and continue giving each `AgentSlot` exactly one factory while preserving unsupported-runtime warnings/skips, runtime switching, and one cheap Claude resolve-and-log per CLI runtime construction.
- Impacted modules: CLI runtime composition, Client handler/runtime public APIs, built-in provider composition, architecture guards, and runtime QA guidance.

**Review findings**
✅ 1. The built-in factory table is frozen and exhaustive over `RuntimeProvider`, and independent constructions retain distinct Claude-bound closures without process-global installation.
✅ 2. CLI and standalone lookup paths fail closed for unknown keys; standalone coverage includes inherited `toString`, `constructor`, and `__proto__` keys so plain-object prototypes cannot masquerade as factories.
✅ 3. The old global symbols and registration root are removed from production code and barrel exports, all in-repository consumers are migrated, and CLI composition remains on the `@first-tree/client` package entry.
⚠️ 4. Non-blocking: CLI's private `resolveHandlerFactory` relies on its two current callers performing the own-key guard first. The table is frozen and both call paths guard today, so this is not a regression; folding the guard into the resolver would make that invariant more maintenance-safe.

**Action taken**
- Approved exact head `1686ed421302bf9062bb18dfc4e41b839606bae8` with explicit maintainer authorization. The PR remains Draft; independent exact-head QA and stage confirmation are separate gates.

PR #2214, opened 2026-08-06
refactor(client): S1c close CLI session-internals boundary

yuezengwu:
## Summary
- Add Client-owned `cleanAgentWorkspaces` that owns workspaces/sessions layout, SessionRegistry reads, active/evicted selection, and the existing zero-deletion `cleanWorkspaces` call; thin CLI `agent workspace clean` to one Client call with unchanged output.
- Remove `SessionManager` / `SessionRegistry` from Client root and runtime public barrels (files and internal lifecycle unchanged).
- Fold own-key guarding into CLI `ClientRuntime.resolveHandlerFactory` so prototype keys (`toString` / `constructor` / `__proto__`) fail closed without constructing an `AgentSlot`.

## Test plan
- [x] Client focused: workspace-maintenance, barrel-exports, provider-boundary-guard
- [x] CLI focused: cli-command-matrix-extra, client-runtime-context-tree (includes prototype-key resolver characterization)
- [x] `@first-tree/client` full suite: 2421 passed / 7 skipped
- [x] Client + CLI typecheck/build, root `pnpm check`, `git diff --check`
- [x] Source/dist boundary: barrels no longer export SessionManager/SessionRegistry; CLI production has no SessionManager/SessionRegistry/cleanWorkspaces
- [ ] Exact-head independent QA on frozen Draft head (do not Ready/merge until PASS)

## Scope / non-goals
S1c only. Does not delete or split SessionManager/SessionRegistry lifecycle, move packages, converge provider `contracts` / `provider-support`, or change user-visible clean behavior (still deprecated no-op / zero deletion).

### DIFF
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -52,7 +52,6 @@ const coreMocks = vi.hoisted(() => ({
 
 const clientMocks = vi.hoisted(() => ({
   cleanAgentWorkspaces: vi.fn(),
-  DEFAULT_WORKSPACE_TTL_MS: 7 * 24 * 60 * 60 * 1000,
 }));
 
 const localAgentMocks = vi.hoisted(() => ({
--- a/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
+++ b/apps/cli/src/__tests__/cli-command-matrix-extra.test.ts
@@ -51,8 +51,8 @@ const coreMocks = vi.hoisted(() => ({
 }));
 
 const clientMocks = vi.hoisted(() => ({
-  SessionRegistry: vi.fn(),
-  cleanWorkspaces: vi.fn(),
+  cleanAgentWorkspaces: vi.fn(),
+  DEFAULT_WORKSPACE_TTL_MS: 7 * 24 * 60 * 60 * 1000,
 }));
 
 const localAgentMocks = vi.hoisted(() => ({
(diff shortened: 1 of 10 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The Client now owns workspace/session cleanup orchestration, the CLI preserves its existing zero-deletion behavior, and handler lookup fails closed at the resolver boundary without constructing an `AgentSlot` for inherited object keys.

**Risk level**: A
- Path baseline: Includes `apps/cli/**` and `packages/client/**` -> A.
- Semantic lift: None; A is already the highest level.

**PR summary**
- Problem: The CLI workspace-clean command reached into Client session internals, while its private handler resolver could still treat inherited object keys as factories if a future caller bypassed the current pre-check.
- Approach: Move workspace/session layout, registry reads, active/evicted selection, and low-level cleanup delegation behind one Client API; remove `SessionManager` and `SessionRegistry` from public barrels; and enforce own-property lookup inside the CLI resolver itself. Operator-visible cleanup output and the existing no-op/zero-deletion behavior remain unchanged.
- Impacted modules: CLI agent workspace cleanup and runtime composition; Client workspace maintenance, public barrels, and architecture guards.

**Review findings**
✅ 1. `cleanAgentWorkspaces` is the single owner of workspace/session layout and selection, while the CLI is reduced to argument parsing and presentation without changing missing-root, removed-entry, or count behavior.
✅ 2. `SessionManager` and `SessionRegistry` are removed from both public barrels and all CLI production consumers; the injectable test seam remains outside those barrels.
✅ 3. `resolveHandlerFactory` now performs its own `Object.hasOwn` guard, so `toString`, `constructor`, and `__proto__` fail before `AgentSlot` construction even if a future caller skips the current pre-check.
⚠️ 4. Independent exact-head QA remains a separate Draft-to-Ready gate; this approval does not authorize Ready or merge.

**Action taken**
- Approved exact head `8be43f0f8f0b51c41cf05f3499b12dd1dced97a1` with explicit maintainer authorization and green current CI checks.

PR #2216, opened 2026-08-06
fix: Chat Summary lead line + reading width

yuezengwu:
## Summary

- Promote only the **first physical line** as the Current state headline, so writing-contract soft breaks (`remarkBreaks` → one `<p>`) no longer bold the entire brief
- Lead detection is a pure source-line / `node.position` check (StrictMode-safe; no mutable render flags)
- Lead always renders as a child span with explicit subtitle measure, so parent `[&_p]:text-body` cannot flatten blank-line headlines
- Keep one Markdown document and existing block-first / reference-link fallbacks
- Move the readable measure onto the raised card itself and replace `68ch` with `47.5rem` (~760px) so CJK/Latin mix is not squeezed and the card no longer stretches full-bleed with empty chrome
- Add soft-break + StrictMode regressions plus QA/Momentic wording aligned to the physical-line contract

## Test plan

- [x] `pnpm --filter @first-tree/web exec vitest run src/pages/workspace/center/__tests__/chat-summary.test.ts` — 33 passed
- [x] `pnpm --filter @first-tree/web typecheck`
- [x] `pnpm check` (no new errors on changed files)
- [x] `npx momentic@3.42.0 lint e2e/chat-summary-current-state.test.yaml`
- [ ] Independent exact-head QA (owned separately; PR stays Draft)

PR #2218, opened 2026-08-06
fix: match expanded Chat Summary width to the chat header

Gandy2025:
## Problem

The expanded **Current state** card carried its own reading measure (`--chat-summary-readable-rail: 47.5rem`), while the chat header and the collapsed summary bar it drops from span the full message area. Two consequences:

- the card rendered visibly narrower than the bar directly above it — measured at a 1298px message area: header/bar **1298px**, card **760px** (58%), both left-aligned, so the card looked like a full-width element that failed to fill
- its copy wrapped early inside that narrower column instead of using the available width

## Change

- stretch the card `left: 0; right: 0` so the expanded and collapsed forms of the same surface share one width and one edge with the chat header
- let the copy follow the card width — no measure of its own
- remove the now-unused `--chat-summary-readable-rail` token; the DEV preview panel keeps its previous width, expressed with existing spacing tokens
- align the QA case and Momentic journey wording to the header-matching contract

The result-line headline hierarchy from #2216 is unchanged: only the first physical line is promoted, supporting copy stays quieter.

## Validation

- `pnpm check`
- `pnpm typecheck` — 11/11 tasks, including the web `lint:tokens` design-token guard
- `pnpm --filter @first-tree/web test` — 244 files / 2,246 tests passed
- `npx momentic@3.42.0 lint e2e/chat-summary-current-state.test.yaml`
- real browser measurement of the rendered component at three viewports — header / collapsed bar / expanded card identical in width and left edge each time, with no horizontal overflow:

  | viewport | header | collapsed bar | expanded card |
  | --- | --- | --- | --- |
  | 1800px | 1138 | 1138 | 1138 |
  | 1280px | 1138 | 1138 | 1138 |
  | 700px | 650 | 650 | 650 |

The first pushed head failed CI's Lint & Type Check: `lint:tokens` greps every file under `packages/web/src` for raw pixel literals, tests included, and the initial validation ran `tsc --noEmit` directly instead of the package's `typecheck` script, which chains that guard. The follow-up commit resolves it and the full script now runs clean locally.

Code review was explicitly waived by the requester for this change; two independent reviews were nonetheless submitted and approved on the current head.

### DIFF
--- a/e2e/chat-summary-current-state.test.yaml
+++ b/e2e/chat-summary-current-state.test.yaml
@@ -13,8 +13,9 @@ steps:
   - waitForUrl: /?c=e2e-summary-
   - assert: The real Workspace has Mobile navigation polish selected. Its expanded panel is labeled Current state,
       presents Mobile navigation polish is awaiting product review as the visually stronger lead above quieter
-      supporting context and a distinct Next line, keeps freshness visible, and constrains the raised card to a
-      readable left-aligned measure instead of stretching as a full-bleed empty surface across the message area.
+      supporting context and a distinct Next line, keeps freshness visible, and gives the raised card the same width
+      as the chat header and the collapsed bar above it, with its copy filling that width instead of wrapping early
+      inside a narrower column.
   - click: the Summary rollout validation chat in the conversation list
   - assert: Summary rollout validation is now selected and the Current state panel has changed to that chat's summary,
       led by Chat Summary hierarchy is ready in the real Workspace; none of the Mobile navigation polish summary remains
--- a/packages/qa/cases/web/chat-summary-chat-switch-reentry.md
+++ b/packages/qa/cases/web/chat-summary-chat-switch-reentry.md
@@ -18,7 +18,8 @@ substitute a `/preview/*` route or an operator's existing browser session.
 
 Switch between the chats from the conversation list and confirm that the selected topic, expanded Current state panel,
 freshness, and description all move together. The first physical line should be visibly scannable as the lead,
-supporting copy should be quieter and held to a readable card measure (~720–800px, not a full-bleed empty rail), and
+supporting copy should be quieter, the expanded card should span the same width as the chat header and the collapsed
+bar it drops from — with its copy following that width rather than wrapping at a measure of its own — and
 content from the previously selected chat must not remain. Collapse the panel and confirm its one-line result preview
 still identifies the selected chat while the supporting copy no longer covers the message stream. Include legacy
 Markdown such as a reference link or block-first content when the run is intended to assess compatibility as well as
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -544,12 +544,6 @@ video[data-onboarding-orientation-video]::cue {
   --sp-95: 23.75rem; /* 380px — chat-description popover card width (~60ch) */
   --mobile-tabbar-height: 3.5rem; /* 56px */
 
-  /* Readable measure for the expanded chat current-state card. Uses rem (not
-     ch) so CJK/Latin mixed copy keeps ~760px instead of shrinking with Latin
-     glyph width. The raised card itself is constrained, so the surface and
-     text share one edge rather than leaving empty half-width chrome. */
-  --chat-summary-readable-rail: 47.5rem; /* 760px */
-
   /* Agent detail shell: a compact local-navigation rail plus one readable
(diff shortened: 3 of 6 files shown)


yuezengwu (approved):
Goal and core change: make the expanded Current state card use the same message-area width and edges as the chat header and collapsed summary bar. The implementation switches the floating card from an independent 47.5rem measure to `left: 0; right: 0`, lets the Markdown copy follow that width, removes the obsolete token, and updates the preview, unit assertion, E2E journey, and reusable QA case to the same contract.

No blocking issues found. The portal's positioning container is the full-width timeline wrapper, so the new inset aligns the expanded surface with the surrounding chat chrome while retaining border-box sizing and the existing vertical overflow behavior. The first-line headline/supporting-copy hierarchy is unchanged, and the removed token has no remaining runtime references.

Approved based on diff review; I did not run tests or QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The expanded Current state surface now shares the message-area edges with the chat header and collapsed bar without changing the existing result-line hierarchy or scrolling behavior.

**Risk level**: C
- Path baseline: `packages/web/**`, QA guidance, and E2E tooling only.
- Semantic lift: none; the change does not touch subscriptions, WebSocket parsing, or the Web data layer.

**PR summary**
- Problem: People expanding a chat's Current state saw a 760px card beneath a full-width header and collapsed bar, making the surface look incomplete and causing summary copy to wrap before the available message area was used.
- Approach: Stretch the portaled card across its full-width timeline positioning container, let the Markdown document inherit that width, remove the obsolete readable-rail token, and align preview, deterministic assertions, E2E wording, and the reusable QA case with the header-matching contract.
- Impacted modules: Workspace Chat Summary styling and coverage, DEV preview, E2E journey, and Web QA case.

**Review findings**
✅ 1. The portal container is the full-width timeline wrapper, so `left: 0; right: 0` aligns the expanded card with the surrounding chat chrome; global border-box sizing keeps its border and padding inside that width. [`packages/web/src/pages/workspace/center/chat-summary.tsx:657`]
✅ 2. The patch does not modify the lead-selection or Markdown-rendering path, preserving the first-physical-line headline and quieter supporting copy established by #2216.
✅ 3. The removed token has no remaining runtime references, while the DEV preview receives an independent explicit panel width and the checks describe the new layout contract consistently.

**Action taken**
- Approved exact head `f366688d85022bba8a98d49beb408ead33ef8f01`.

Static review only; I did not run tests or QA.

yuezengwu (approved):
Re-reviewed after synchronization.

Goal and incremental change: retain the full-width Current state card while satisfying the Web design-token guard. The DEV preview's 71.25rem width is now expressed as the equivalent `calc(var(--sp-95) * 3)`, and the inset assertions parse the rendered numeric values instead of embedding raw pixel strings.

No blocking issues found. The token calculation preserves the preview geometry, the revised assertions still fail when either inset is absent or non-zero, and the production layout and first-line/supporting-copy hierarchy are unchanged.

Approved exact current head based on static diff review; I did not run tests or QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up commit removes the design-token guard violations without changing the production layout or the Current state lead-line contract.

**Risk level**: C
- Path baseline: `packages/web/**` only in this incremental review.
- Semantic lift: none; the follow-up touches a DEV preview width expression and deterministic assertions only.

**PR summary**
- Problem: The first head implemented the correct full-width Current state card but failed Web's design-token guard because the new test strings and preview comment contained raw pixel literals.
- Approach: Express the preview's same 71.25rem width as three existing 23.75rem spacing units and compare the card's zero insets numerically instead of against pixel strings.
- Impacted modules: Chat Summary DEV preview and its unit test only; production rendering is unchanged from the previously approved head.

**Review findings**
✅ 1. `calc(var(--sp-95) * 3)` preserves the preview panel's prior 71.25rem width and its responsive `maxWidth: 100%` behavior while avoiding a new runtime token.
✅ 2. Parsing both serialized inset values still verifies the card is anchored at zero on each edge, and the remaining width/max-width assertions continue to guard against restoring the old readable rail.
✅ 3. No production component, Markdown rendering, or first-line promotion code changed in this follow-up.

**Action taken**
- Approved exact head `6e8aad76673d0d01d99c3a38a4e6f1d023eaa6a3`.

Incremental static review only; I did not run tests or QA.

PR #2220, opened 2026-08-06
feat(chat): web remove chat participant (phase 1)

yuezengwu:
## Summary
- Add Web Class C `DELETE /api/v1/chats/:chatId/participants/:agentId` through the canonical `removeChatParticipant` mutation; the existing agent DELETE remains 204.
- Apply removal under one exclusive membership lock: speaker → watcher or detach, inbox `pending`/`delivered` → `cancelled`, non-Human session eviction, cron pause, open-request rejection, and post-commit audience / me-chats refresh.
- Durably fence pre-removal provider context with a monotonic per-(agent, chat) `resumeGeneration`: removal bumps it atomically, inbox and cleanup frames carry it, old Clients fail closed, and capable Clients teardown + delete the saved mapping + flush the tombstone before any newer-generation provider admission.
- Preserve failure and concurrency boundaries: failed persistence stays non-admissible and retryable, per-chat adoption serializes to the highest synchronously announced generation, stale deliveries / soft terminates cannot affect the new generation, and the generation-0 no-op path stays synchronous so inbox-recovery custody ordering does not shift.
- Close send, delivery, and session races with shared membership locks and a WS send-boundary fence; expose Remove in the Participants sidebar with confirm copy, watcher/detached toasts, and server-backed refresh.
- **HTTP GET `/agent/inbox` boundary:** debug/admin poll only. Once a response is returned, that payload cannot be recalled after a membership change; runtime correctness is enforced on the WS data plane.

## Test plan
- [x] Client `session-manager.test.ts` (104), `agent-slot.test.ts` (45), wire capability tests (18), and Shared inbox / WS schemas (40).
- [x] Server `remove-chat-participant.test.ts` (30), real-branch WS fake tests (23), and adjacent Inbox/session/Cron/onboarding/me-chats/index tests (150).
- [x] Web Participants Remove DOM regression (3).
- [x] `pnpm typecheck` and `pnpm check`; exact-head CI, CodeQL, and CLA all pass.
- [x] Apply migrations `0093_useful_lila_cheney` and `0094_flashy_nova`; verify `cancelled` status and both resume-generation columns in disposable PostgreSQL.
- [x] Independent full-isolated QA against the exact PR head using `packages/qa/cases/cross-surface/web-remove-chat-participant.md`, including production Server/Web/Client/CLI artifacts, live HTTP/WS/DB removal journeys, restart/reconnect, durable mapping failure/retry, generation races, old-client fail-closed behavior, and cleanup.
- [ ] Manual signed-in browser Remove click. The exact UI behavior is covered by the DOM suite and the removal journey by the live public API.


baixiaohang (changes requested):
- Rationale: Participant removal does not durably revoke the removed agent's client-side provider-session mapping, so a later re-add can resume pre-removal context.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a database constraint change and core chat/message/session services.
- Semantic lift: none (already B-high).

**PR summary**
- Problem: In a multi-participant chat, a speaker needs to remove another human or agent from Web without leaving the removed runtime able to receive work or retain stale session/cron state.
- Approach: Adds Web Remove through a shared canonical mutation, coordinates membership/inbox/session/cron changes, fences send/delivery/session races, and refreshes the realtime UI.
- Impacted modules: server chat/inbox/session/cron/WS, shared schemas, Web Participants sidebar, migration, and QA coverage.

**Review findings**
❌ 1. Removing an agent does not durably invalidate its locally persisted provider-session mapping. The transaction marks the server session evicted, but the only client cleanup is the fire-and-forget soft terminate started after commit; it silently does nothing when the agent has no live route, and it can also lose to an immediate re-add because its membership fence then sees the agent as a speaker again. The client persists chat-to-provider-session mappings, reloads them as evicted mappings, reports them as suspended once the chat becomes active again, and can resume them on the next message. Consequently, an offline remove followed by re-add/reconnect (or a fast re-add before the delayed terminate runs) can revive the pre-removal provider thread. Please add a durable client-visible removal generation/tombstone or an equivalent protocol that guarantees the old mapping is destroyed before a re-added participant may resume, with deterministic regressions for offline remove → re-add → restart/reconnect → new message starting fresh and for re-add winning the race against delayed soft termination. [R4/R5 / `packages/server/src/services/remove-chat-participant.ts:306`]

✅ 2. The Class C route reuses the canonical mutation and rechecks speaker authorization, while the membership/inbox/WS/session/cron lock coverage closes the principal races as long as membership is not removed and re-added across the asynchronous cleanup boundary.

**Action taken**
- Submitted request changes at head `a01c62f4a7fcb9cfdf29fafd8c4d19d5320bf056`.

baixiaohang (changes requested):
- Rationale: The durable resume-generation design addresses the original offline mapping gap in the normal path, but its failure and concurrency paths can still admit stale provider context or resurrect a cancelled delivery.

**Risk level**: A
- Path baseline: the follow-up now changes `packages/client/**`.
- Semantic lift: none (already A).

**PR summary**
- Problem: Removing and later re-adding an agent must not let the daemon resume the provider session that existed before removal.
- Approach: Persists a monotonic per-(agent, chat) resume generation, stamps inbox and soft-terminate frames, and makes capable clients durably tear down older provider context before admitting a newer generation.
- Impacted modules: client session registry/manager and WS negotiation; server chat-session/inbox schema, fan-out, WS delivery, and participant removal.

**Review findings**
❌ 1. A failed durable flush is treated as retryable, but the in-memory generation has already advanced. Lines 4947-4950 update both generation maps before `flushOrThrow`; if that write fails, the catch returns `retry` without rolling the maps back or recording adoption debt. The next delivery at the same generation then exits through `incoming === local` at line 4941 and admits the provider without retrying the tombstone flush. A same-process reconnect can therefore start fresh work while disk still contains the pre-removal provider mapping; a crash before a later best-effort registry write reloads and can resume that old mapping. Keep the new generation non-admissible until its deletion/tombstone is durably committed (or retain explicit adoption debt), and make the existing flush-failure regression prove that the second delivery retries the durable write before any provider route opens. [R4 / `packages/client/src/runtime/session-manager.ts:4939`]

❌ 2. Generation adoption is not serialized per chat and does not re-check authority after its async teardown. Two WS deliveries are dispatched concurrently by `AgentSlot`; if a generation-1 admission is blocked in teardown and generation 2 arrives, both calls can capture the same older `local`, share/await cleanup, then independently persist and return `proceed`. The generation-1 dispatch subsequently captures the latest admission counter and can inject its now-stale, remove-cancelled message into the generation-2 provider session. Coalesce/serialize per-chat adoption and re-read the committed generation before returning so the older caller returns `drop`; add a deterministic delayed-teardown g1-vs-g2 regression. [R1/R4 / `packages/client/src/runtime/session-manager.ts:4935`]

❌ 3. The old-client fail-closed branch can overwrite a concurrent removal's terminal cancellation. After claim marks a generation>0 row `delivered`, lines 888-891 reset it to `pending` using only `id). If removal updates that row to `cancelled` while this UPDATE is waiting on the row lock, the UPDATE resumes after commit and changes it back to `pending`. On a later re-add, a capable client can receive work from the pre-removal generation that removal had cancelled. Make the rollback conditional on the row still being `delivered` (and preserve the membership ordering invariant), with a barrier regression where removal wins before the capability rollback. [R4/R5 / `packages/server/src/api/agent/ws-client.ts:888`]

✅ 4. The monotonic server-side generation, inbox stamping, new-client durable teardown, stale soft-terminate suppression, and capability negotiation are the right shape for the originally reported offline remove → re-add case once the failure/concurrency boundaries above are closed.

**Action taken**
- Submitted request changes at head `ca3ef3c8868111e23faa6e92b04cacdf13d2ea09`.

baixiaohang (changes requested):
- Rationale: The flush-failure and cancellation-rollback fixes are sound, but the generation-adoption queue still lets the older caller return before the queued newer generation commits.

**Risk level**: A
- Semantic lift: none (already A).

**PR summary**
- Problem: Removing and re-adding an agent must prevent any pre-removal provider session or delivery from entering a later membership generation.
- Approach: Adds durable resume generations and, in this follow-up, attempts to serialize generation adoption, roll back failed persistence candidates, and protect terminal inbox cancellation.
- Impacted modules: client session manager/registry and server WS inbox delivery.

**Review findings**
❌ 1. The FIFO adoption lock does not make a queued newer generation visible before the older caller's post-check. In the delayed g1/g2 sequence, g2 records only a tail promise and waits at line 4969. When g1 releases its `hold` in line 4973, g1's `await withResumeGenerationAdoptionLock(...)` continuation runs and reads `committed === g1` at lines 4998-5005 before g2 can enter its callback and commit g2. g1 therefore still returns `proceed`. g2 then enters, invalidates delivery admission, and awaits teardown; g1 resumes afterward and captures that newest admission counter, so its stale message can still open/inject into the provider while g2 is adopting. The new test expects g1 to ACK/drop, but the implementation does not establish the ordering that assertion requires. Track the highest requested/pending generation before queuing (or keep the older admission fenced through all queued newer adoption) and make g1 drop whenever g2 was already queued before g1's teardown released. [R1/R4 / `packages/client/src/runtime/session-manager.ts:4958`]

✅ 2. The generation candidate is now kept non-admissible until `flushOrThrow` succeeds, and registry memory rolls back on failure, so a same-process retry re-attempts the durable tombstone.

✅ 3. The capability-withhold UPDATE now predicates on `status='delivered'`, preserving a concurrent removal's terminal `cancelled` state.

**Action taken**
- Submitted request changes at head `82819cbf5724b41e94d96009e296e840daa63fde`.

baixiaohang (changes requested):
- Rationale: The pending-generation high-water mark closes the g1/g2 adoption race, but the generation gate now moves the existing inbox-recovery boundary even when no generation is being adopted.

**Risk level**: A
- Semantic lift: none (already A).

**PR summary**
- Problem: Removing and re-adding an agent must start a fresh provider session without admitting messages or provider context from an older membership generation.
- Approach: Uses durable monotonic resume generations, a synchronous pending-generation high-water mark, and serialized adoption before provider admission.
- Impacted modules: client session manager/registry and server WS inbox/session handling.

**Review findings**
❌ 1. `dispatch` now always awaits `ensureResumeGenerationAdmission` before it samples `takeRecoveryActivationReady` / `shouldRecoverBeforeDispatch`, including the common generation-0 case where committed and pending are already 0. That no-op await changes the custody ordering: a frame that entered dispatch while `recoverChat` was still running can resume after recovery clears the ledger and sets `recoveryActivationReady`, then be treated as the post-recovery redelivery and routed into the provider. The exact-head regressions demonstrate both forms of the boundary shift: the dead-handler frame resumes one dispatch too early, and a newer frame crosses a pending recovery and reaches `inject` twice across the retry sequence. Preserve the recovery state observed at dispatch entry, or add a synchronous no-adoption fast path, so a pre-recovery frame cannot become authoritative merely because the generation gate yielded; keep deterministic coverage for both recovery cases. [R4 / `packages/client/src/runtime/session-manager.ts:782`]

✅ 2. Synchronously recording the highest pending generation before the FIFO lock makes an older g1 caller see queued g2 authority and drop, closing the prior stale-admission race.

✅ 3. The earlier durable-flush rollback and terminal-cancellation predicate remain intact.

**Action taken**
- Submitted request changes at head `ad1f2d52b53121f80d94bdad0f0fbfacbfb8bb98`.
- Exact-head CI also needs fixture/expectation updates: the Server fake DB queue does not account for the new legacy-client generation lookup, and two AgentSlot assertions still expect the old two-argument command call.

yuezengwu:
@baixiaohang Addressed all requested changes at exact head `094c2ba6c55504f6a3c52dfb04aa53c91a131b87`.

### Finding closure

1. **Durable offline revocation:** removal now atomically increments a monotonic per-(agent, chat) resume generation and stamps inbox / cleanup frames. A capable Client must teardown the old provider context, delete the persisted mapping + nonce, and durably flush the tombstone before newer-generation provider admission. Old Clients fail closed for generation > 0.
2. **Flush failure:** the admissible in-memory generation advances only after `flushOrThrow` succeeds; a failed candidate is reverted. Same-process same-generation redelivery retries the durable write before any provider route can open.
3. **g1/g2 adoption race:** each chat has serialized adoption plus a synchronously recorded pending high-water mark. The lock adopts through the highest queued generation and both locked and post-lock paths re-read authority, so queued g2 makes delayed g1 drop/ACK; only `start(msg-g2)` may run.
4. **Cancelled-row rollback:** legacy capability withholding updates only rows still in `delivered`, so a concurrent removal's terminal `cancelled` state cannot be resurrected.
5. **Recovery custody regression:** generation-0 / matching-generation admission now has a truly synchronous no-adoption fast path. It cannot microtask-yield before `takeRecoveryActivationReady` / `shouldRecoverBeforeDispatch`; dead-consumer and pending-recovery frames remain withheld until authoritative redelivery.
6. **Exact-head fixture compatibility:** no-generation AgentSlot commands retain the two-argument call, and the real-branch WS fake advertises `wsResumeGenerationV1`, avoiding an unintended legacy lookup in its DB queue.

### Exact-head evidence

Independent `full-isolated` QA returned **PASS** on this SHA after checking head integrity at start, `QA READY`, execution, and final report:

- Client SessionManager **104/104**, including the two recovery-boundary regressions and resume-generation admission **5/5**; AgentSlot **45/45**; wire capability **18/18**; Shared schemas **40/40**.
- Server remove participant **30/30**, real-branch WS fake **23/23**, adjacent Inbox/session/Cron/onboarding/me-chats/index **150/150**; Web Participants DOM **3/3**.
- Live production-image Server/Web/Client/CLI/PostgreSQL cell covered Human/Agent removal, watcher/detach, open-request conflict, send rejection, cancelled transitions, cron pause, history/user-state preservation, re-add stamping, generation 1 → 2, daemon restart/reconnect, and cleanup.
- The offline mapping, delayed soft-terminate, persistence-failure retry, g1/g2, stale-generation, old-client fail-closed, and cancellation rollback paths all passed deterministic filesystem/handler regressions. No real provider credential was needed or used.
- `pnpm typecheck`, `pnpm check`, migrations `0093`/`0094`, CI, CodeQL, and CLA all pass at the exact head.

The PR description is updated and the PR is Ready again. Please re-review this exact head.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The durable resume-generation protocol and its persistence, concurrency, cancellation, and recovery-ordering boundaries are now closed, with exact-head CI passing.

**Risk level**: A
- Semantic lift: none (already A).

**PR summary**
- Problem: A speaker removing a Human or Agent from a multi-participant chat must stop future work and prevent a later re-add from reviving the removed Agent's prior provider context.
- Approach: Centralizes removal under the membership lock, cancels inbox/session/cron work, and uses a durable monotonic resume generation so capable Clients retire old provider context before newer-generation admission while old Clients fail closed.
- Impacted modules: client session lifecycle and WS negotiation; server chat, inbox, session, cron, and WS handling; shared schemas; Web Participants UI; migrations and QA coverage.

**Review findings**
✅ 1. The canonical removal mutation, membership/send/delivery locks, inbox cancellation, session eviction, cron pause, and post-commit refreshes cover the principal removal races.

✅ 2. Durable generation adoption remains non-admissible through teardown and flush, retries failed persistence, serializes to the synchronously announced high-water mark, drops stale callers, and preserves terminal cancellation.

✅ 3. The synchronous no-adoption path preserves the pre-existing inbox-recovery custody boundary for generation 0 and matching generations.

✅ 4. Exact-head Server, Client/Web, CLI, lint/typecheck, migration, CodeQL, and CLA checks pass.

**Action taken**
- Approved head `094c2ba6c55504f6a3c52dfb04aa53c91a131b87` after explicit authorization.

baixiaohang (changes requested):
- Rationale: The delivery/runtime cutoff is sound, but the newly exposed Web removal operation still lacks the roster authorization and request-lifecycle invariants required to make it safe.

**Risk level**: A
- Semantic lift: none (already A; this review concerns the Server removal authority boundary).

**PR summary**
- Problem: Removing a participant must stop future delivery and provider-context reuse without allowing an ordinary speaker to evict unrelated participants or erase the chat's sole owner record.
- Approach: Centralize Web/agent removal, order it against sends, cancel Inbox custody, evict/fence sessions with a durable resume generation, and expose the action in Workspace.
- Impacted modules: Client runtime; Server chat, membership, Inbox, sessions, Cron, and WebSocket paths; Shared wire schemas; Web Participants UI; migrations.

**Review findings**
❌ 1. The sole owner row remains removable. `removeChatParticipant` selects target `accessMode` but not `role`, so the new Web action can delete `role='owner'`. Because `chats` has no creator column, that irreversibly erases ownership and changes `assertOwner` fallback behavior. Refuse every owner target and prove the owner row survives the attempt. [R4/R5 / `packages/server/src/services/remove-chat-participant.ts:114`]

❌ 2. Authorization is still “any speaker may remove any other speaker.” The caller check proves only speaker membership; it never requires owner-side authority or limits a manager to recalling their own non-Human agent. Fold in the owner-side / own-agent matrix before exposing this in Web. Since that decision reads mutable `agents.manager_id`, lock caller, target, and the owner agent (including an owner watcher) in the shared authority snapshot and cover the owner-manager reassignment race deterministically. [R4/R5 / `packages/server/src/services/remove-chat-participant.ts:85`]

❌ 3. The open-request guard trusts `chat_user_state.open_request_count`, although Need You explicitly derives lifecycle from durable request/resolution rows instead of that projection. It also runs before watcher retention is known and therefore rejects a Human whose request would remain reachable through a watcher row. Reuse/export the canonical open-request predicate and return 409 only for full detachment; cover drifted-counter and retained-watcher regressions. [R1/R4 / `packages/server/src/services/remove-chat-participant.ts:123`]

✅ 4. The shared send/membership ordering, cancelled Inbox custody, durable resume-generation fence, Client teardown, Cron pause, and Web integration remain the correct implementation home for removal.

**Action taken**
- Submitted request changes at head `094c2ba6c55504f6a3c52dfb04aa53c91a131b87`; this supersedes my earlier approval after the roster gaps were surfaced from #2223.

I inspected the exact head and surrounding service/query paths without running tests or QA.

baixiaohang (changes requested):
- Rationale: The authorization implementation is now the right shape, but the required manager-reassignment regression still does not mutate authority and therefore cannot prove the post-wait decision uses the newly committed owner.

**Risk level**: A
- Semantic lift: none (already A; this follow-up changes the Server roster-authorization and Web affordance boundary).

**PR summary**
- Problem: Removing a chat participant must stop delivery and provider-context reuse while preventing ordinary speakers from evicting unrelated participants or the chat owner.
- Approach: Centralizes removal under membership/session fences, adds durable resume generations, applies an owner-side / own-agent authorization matrix, derives Web Remove affordances from that same matrix, and blocks full human detachment while a durable request is open.
- Impacted modules: Client session lifecycle; Server chat, membership, Inbox, session, Cron, and WebSocket paths; Shared schemas; Web Participants UI; migrations and QA.

**Review findings**
❌ 1. The manager-reassignment race regression never performs a reassignment. It only holds the target agent row, releases it unchanged, and expects the originally authorized removal to succeed. That test would still pass if a future implementation read `manager_id` before acquiring the row lock, which is the stale-authority failure this regression is supposed to prevent; the new `afterAuthoritySnapshotForTest` hook is also unused. Make the competing transaction actually commit a target (and owner-side, if kept as a separate authority source) manager transfer before the blocked removal continues, then assert the former manager receives 403 and the participant row survives. [R4 / `packages/server/src/__tests__/remove-chat-participant.test.ts:1945`]

✅ 2. The mutation now refuses owner targets and derives owner-side / own-agent authority from the locked caller, target, and owner-agent snapshot.

✅ 3. The full-detachment guard reuses Need You's durable request predicate and permits removal when the Human retains a watcher row.

✅ 4. Web now fails closed on server-derived `canRemove`, and the projection shares the mutation's pure authorization matrix.

**Action taken**
- Submitted request changes at head `7d71d5db96d77edeada112a001f5e910887257ef`.
- I inspected the exact-head diff and surrounding lock/update paths without running tests or QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The durable removal, roster authorization, and manager-reassignment boundaries are now closed, with exact-head CI passing.

**Risk level**: A
- Semantic lift: none (already A).

**PR summary**
- Problem: A speaker removing a Human or Agent must stop future work and provider-context reuse without letting ordinary speakers evict unrelated participants or erase the chat owner.
- Approach: Centralizes removal under membership/session fences, advances a durable resume generation, applies one owner-side / own-agent authorization matrix to both mutation and Web affordance, and blocks only full Human detachment with a durable open request.
- Impacted modules: Client session lifecycle and WS negotiation; Server chat, membership, Inbox, session, Cron, and WS handling; Shared schemas; Web Participants UI; migrations and QA.

**Review findings**
✅ 1. The canonical mutation protects the owner row, enforces owner-side / own-agent authority, cancels Inbox custody, fences sessions and delivery, pauses Cron, and refreshes the affected audience.

✅ 2. Durable generation adoption remains non-admissible through teardown and flush, retries persistence failures, serializes concurrent generations, drops stale callers, and preserves recovery ordering.

✅ 3. The open-request guard reuses Need You's durable predicate and applies only when Human removal fully detaches; Web fails closed on server-derived `canRemove` from the same pure authorization matrix.

✅ 4. Deterministic target-manager and owner-agent transfer races now commit real `manager_id` changes before removal resumes, proving former authority receives 403 while membership and the owner row survive.

✅ 5. Exact-head Server, Client/Web, CLI, lint/typecheck, migrations, CodeQL, and CLA checks pass.

**Action taken**
- Approved head `4fba46e6df51723ea02246b997583cb387eb43c3` after explicit authorization.

baixiaohang (commented):
**Recommendation**: comment
- Rationale: The three roster findings are correctly addressed in code at the new head; the PR is still Draft, so I am not changing the review to approve yet.

**Risk level**: A
- Semantic lift: none (already A).

**PR summary**
- Problem: The Web removal action must establish a real delivery/session cutoff while preserving the chat owner and refusing unrelated speakers from evicting participants.
- Approach: Keep the existing resume-generation cutoff and add one shared owner-side / own-agent authorization matrix for mutation and UI affordance, locked authority snapshots, and canonical durable open-request reachability checks.
- Impacted modules: Client runtime; Server chat/membership/Inbox/session/Cron/WS services; Shared schemas; Web Participants UI; migrations and QA.

**Review findings**
✅ 1. Owner targets are now refused before mutation, and the regression proves the owner row survives.

✅ 2. Mutation and `canRemove` projection share `isAuthorizedToRemoveParticipant`; caller, target, and owner authority rows (including a watcher owner) are covered, with deterministic target-manager and owner-manager transfer races proving the locked snapshot.

✅ 3. Removal reuses `need-you.ts::openRequestPredicate`, ignores a drifted counter, returns 409 only for full detachment, and permits a retained watcher to keep the request reachable.

✅ 4. The original shared send/membership ordering, cancelled Inbox custody, durable resume-generation fence, Client teardown, Cron pause, and Web integration remain intact.

**Action taken**
- Submitted a comment review at head `4fba46e6df51723ea02246b997583cb387eb43c3`; the code findings from my prior changes-requested review are closed. Approval is deferred while the PR remains Draft.

I inspected the updated diff and surrounding service/query/lock paths without running tests or QA.

PR #2221, opened 2026-08-06
fix(client): keep automatic provider discovery out of macOS protected folders

Gandy2025:
## Problem

The daemon probes provider installs at startup, on reconnect, and on the degraded-capability poll — before the user has selected anything. Paths in that automatic flow reached into folders macOS guards behind a TCC **Files & Folders** consent prompt, so a fresh install could ask for Desktop / Documents / Downloads / iCloud access for reasons the user could not connect to any action of their own.

None of them needed that access to do their job.

### 1. The login-shell `PATH` probe does no filesystem access on macOS

`packages/client/src/runtime/login-shell-path.ts`

The script used to run `cd "$d" && pwd -P` over every `$PATH` entry. `cd` resolves a whole path at once, so it enters a protected root whenever an entry is a symlink into one or has a symlinked ancestor — and no lexical test on the spelling can predict that. Review of this PR killed two attempts to carve out "safe" spellings (a temp-root prefix, then the fnm/nvm multishell directory name); neither is a boundary, since `$HOME` can live under a temp root and a multishell-named link can point at Documents.

So on macOS the script now prints `$PATH`, one entry per line, and nothing else — no `cd`, no `pwd`, no `readlink`. It also runs `set -f`: unquoted `$PATH` undergoes pathname expansion as well as field splitting, so an entry spelled `$HOME/Documents/*` made the shell *enumerate* that directory and pull its entry names into the result, with no explicit filesystem command in sight. That bug predates this PR; the no-access claim made here is what puts it in scope.

All resolution moved to `resolveOutsideProtectedRoots`, which walks a path one component at a time from the root, checks each component against the protected roots **before** touching it, and only ever calls `readlink` — which reads a link's own target without following it. A symlink target is re-walked from the root, so an expansion landing in a protected root is caught on the next iteration; that covers symlinked ancestors without a special case. Comparison is case-folded, because the default macOS filesystem is case-insensitive and `$HOME/documents/bin` *is* `$HOME/Documents/bin`. Cycles are bounded.

The guarantee is now structural — no branch remains that could enter an unvetted path — and it lives in one function that unit tests reach directly, rather than in a shell string. Off macOS the emitted script is unchanged apart from `set -f`.

### 2. Version-manager discovery no longer depends on a per-session symlink

`packages/client/src/runtime/install-locations.ts`

Moving canonicalization out of the shell would otherwise reintroduce the fnm/nvm false negative fixed by #1314: the `$PATH` entry such a shell contributes is a per-session symlink torn down when the shell exits, so nothing can follow it afterwards. The binary itself lives in a stable version dir that does not move — the problem is knowing *which* one the shell had selected.

The probe protocol now carries `$FNM_DIR` and `$NVM_BIN`. Reading an environment variable costs no filesystem access at all, which is precisely why these are capturable while a symlink target is not.

The fallback then answers **only when the selection can be established**, judged across the complete state rather than per root — because resolving a version the user did not select silently swaps the executable and its context, which is worse than reporting the provider as not found:

- **One manager reported.** Only it is used. `$NVM_BIN` names the selected version outright and replaces nvm enumeration. `$FNM_DIR` names only the active root, so that root and no other is consulted, and answers only if it holds exactly one version. Default roots are not scanned: if we know which manager is active and it cannot answer, an unrelated installation is not a better answer.
- **Both managers reported.** Nothing. Which one was active was decided by the live `$PATH` order, and that is exactly what is gone; two variables say two init scripts ran, not which owned precedence.
- **Nothing reported.** Default roots are scanned, and the result is used only when exactly one safe candidate exists in total. Two roots holding one version each are as ambiguous as one root holding two.

Every root is vetted before it is listed and every candidate before it is returned — `$FNM_DIR` is user-controlled, and `~/.nvm` or a fnm data dir can be a symlink, or sit under one, into a protected folder.

### 3. The Codex external-Context classifier stops probing Documents

`apps/cli/src/core/context-integration/client-preflight.ts`

`classifyCodexProjectlessPath` called `realpathSync` on `~/Documents/Codex` for every session, including sessions whose cwd was nowhere near Documents. The caller already canonicalizes the cwd, so canonicalizing the base is only meaningful when the base itself is redirected — which macOS iCloud "Desktop & Documents Folders" does by making `~/Documents` a symlink. The base is now canonicalized only when the caller's original, pre-`realpath` cwd is already lexically inside it. Ordinary project cwds no longer touch Documents at all. `CODEX_PROJECTLESS_CLASSIFIER_VERSION` is bumped so cached classifications recompute once.

## What is deliberately unchanged

Discovery order (daemon `PATH` → well-known install dirs → login-shell `PATH` → Codex desktop app), the provider list, the capability wire schema, the background refresher and its backoff, the Web provider picker, and the launchd plist. The daemon's own service `PATH` is composed by First Tree and never inherited from the user, so the login-shell probe was the only user-derived directory source in automatic discovery.

## Cost

A provider whose `$PATH` entry lives inside Desktop, Documents, Downloads, iCloud Drive, or a cloud File Provider mount is no longer auto-discovered. Those locations essentially never hold a CLI on `PATH`, and the alternative is asking every user for consent to folders the product has no reason to read.

On macOS, once a per-session version-manager link is gone, the fallback answers only when the selected version can be established from what the shell reported. Otherwise the provider reports as not found rather than silently resolving a version the user did not select. In practice the live link is still resolvable, since resolution runs immediately after the probe returns.

Still open, deliberately: whether `/Volumes` and `/Network` should join the protected list. That removes the removable- and network-volume prompt categories too, but a toolchain on an external volume would then need a way to be named explicitly, and no such escape hatch exists today (`CLAUDE_CODE_EXECUTABLE` is the only override, and it covers one provider). Worth deciding separately.

## Verification

- `pnpm check`, `pnpm typecheck` — clean.
- `packages/client` full suite: 2445 passed, 7 skipped.
- `apps/cli` full suite (batch runner): green except 3 pre-existing `daemon-refresh-unit` failures that reproduce identically on an unmodified tree (this host has no dev portable shim at `~/.local/bin/first-tree-dev`).

Tests added for each property that can otherwise regress silently:

- **No protected access, however an entry reaches one** — end-to-end over the real shell *and* the real parse, with four routes in one `$PATH`: by spelling, symlinked entry, symlinked ancestor, and a `fnm_multishells`-named link pointing at Documents. It asserts the shell echoes every entry **byte for byte**; four are symlinks whose resolved spelling differs, so any `cd` would show as a changed line.
- **Nothing protected is ever touched** — resolution takes an injectable `readlink` seam and a test asserts no protected path is ever passed to it. That observes what was touched instead of arguing about control flow.
- **The complete candidate is vetted, not just its directory** — a well-known dir whose *ancestor* is symlinked into Documents, and an executable that is *itself* such a symlink, are both rejected; an ordinary binary and one symlinked to a safe target still resolve, so the gate is not simply refusing everything.
- **Version-manager roots are vetted before listing** — a protected `$FNM_DIR`, a symlinked default root, and a symlinked version entry, with a `readDir` seam that throws if a rejected root is ever listed.
- **Wildcard entries are not expanded** — a `$HOME/Documents/*` entry is echoed literally and its directory never enumerated, on both platforms.
- **Case-insensitive spellings are caught** — mixed-case `documents` / `DOWNLOADS` / `mobile documents`.
- **The macOS script contains no `cd` / `pwd` / `readlink`**, and non-macOS keeps its canonicalization.
- **Teardown behaviour** — a login-shell seam that deletes the per-session symlink before returning still resolves from the stable dir; a shell-only custom `$FNM_DIR` is reachable; a live selection wins while a dead one yields missing rather than the newest.
- **The fallback never guesses** — a multi-version root, two single-version roots, an ambiguous active `$FNM_DIR` beside an unrelated single-version root, and both manager signals present (with the fnm root both unambiguous and ambiguous) all yield nothing; `$NVM_BIN` answers where enumeration alone would have been ambiguous.
- **Ambiguity is decided on the whole search list, and only where it exists** — under `darwin`, a capture of `[dead fnm multishell, $NVM_BIN]` with a runnable binary under `$NVM_BIN` resolves to nothing rather than substituting nvm; off macOS, where the shell canonicalized while alive, the same dual-manager PATH keeps its order and still resolves to the selected fnm binary. Both were mutation-checked — removing the filter fails the first, applying it unconditionally fails the second.
- **The classifier makes zero canonicalization calls** for a cwd outside the base, and still canonicalizes when the original cwd is unknown. The two existing symlink regression tests pass unchanged; the redirected-base one is the iCloud case and is why the `realpath` could not simply be deleted.

## QA

Adds `packages/qa/cases/runtime/macos-automatic-probe-file-access.md`. It asserts both halves together — no protected-folder read across idle / reconnect / on-demand probe, *and* a login-shell-only provider still discovered — because either property alone is trivially satisfiable by breaking the other. It needs a macOS account with no prior TCC decision for the runtime binary plus filesystem tracing, so it is a human-requested case, not a CI gate.

### DIFF
--- a/apps/cli/src/__tests__/context-project-resolver.test.ts
+++ b/apps/cli/src/__tests__/context-project-resolver.test.ts
@@ -1,7 +1,7 @@
 import { mkdirSync, mkdtempSync, symlinkSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import {
   classifyCodexManagedWorktreePath,
   classifyCodexProjectlessPath,
@@ -10,6 +10,15 @@ import {
   resolveProviderProject,
   resolveSessionContextProject,
 } from "../core/context-integration/client-preflight.js";
+
+/**
+ * A recording canonicalization seam, so a test can assert exactly which paths
+ * the classifier touches. Resolving each path to itself keeps the synthetic
+ * (non-existent) roots below comparable without going near the filesystem.
+ */
+function recordingRealpath() {
+  return vi.fn((path: string) => path);
+}
 
 describe("Context project resolver", () => {
   it.each([
@@ -137,6 +146,40 @@ describe("Context project resolver", () => {
         },
       ),
     ).toMatchObject({ kind: "pathless", source: "codex_documents_v1" });
+  });
+
+  // On macOS `~/Documents` sits behind a TCC prompt, so classifying an ordinary
+  // project must not touch it. The base is only canonicalized when the caller's
+  // original cwd is already lexically inside it (a redirected base, above).
+  it("never canonicalizes the Documents base for a cwd outside it", () => {
+    const realpath = recordingRealpath();
+    const home = "/Users/alice";
+
+    expect(
+      classifyCodexProjectlessPath(
+        "/Users/alice/code/app",
+        {},
+        { platform: "darwin", home, realpath, rawCwd: "/Users/alice/code/app" },
+      ),
+    ).toBe(false);
+    expect(realpath).not.toHaveBeenCalled();
+
+    expect(
+      classifyCodexProjectlessPath(
+        "/Users/alice/Documents/Codex/2026-07-30/scratch",
+        {},
+        { platform: "darwin", home, realpath, rawCwd: "/Users/alice/Documents/Codex/2026-07-30/scratch" },
+      ),
+    ).toBe(true);
+    expect(realpath).toHaveBeenCalledWith("/Users/alice/Documents/Codex");
+  });
+
+  it("still canonicalizes the base when the original cwd is unknown", () => {
+    const realpath = recordingRealpath();
+    expect(
+      classifyCodexProjectlessPath("/Users/alice/code/app", {}, { platform: "darwin", home: "/Users/alice", realpath }),
+    ).toBe(false);
+    expect(realpath).toHaveBeenCalledWith("/Users/alice/Documents/Codex");
   });
 
   it("accepts only readable directories as path projects", () => {
--- a/apps/cli/src/core/context-integration/client-preflight.ts
+++ b/apps/cli/src/core/context-integration/client-preflight.ts
@@ -5,8 +5,15 @@ import type { ContextIntegrationProject, ContextIntegrationProvider } from "@fir
 import { loadCredentials } from "../bootstrap.js";
 import { channelConfig } from "../channel.js";
 
-const CODEX_PROJECTLESS_CLASSIFIER_VERSION = 1;
+const CODEX_PROJECTLESS_CLASSIFIER_VERSION = 2;
 const SESSION_CACHE_TTL_MS = 30 * 24 * 60 * 60_000;
+
+/**
+ * The one `realpathSync` form this module uses: canonicalize a single path to a
+ * string. Naming it keeps the injected test seams to that shape rather than
+ * Node's full Buffer-returning overload set.
+ */
+type RealpathFn = (path: string) => string;
 
 export type ContextProjectResolution =
   | {
@@ -80,7 +87,7 @@ export function inspectContextSetupLocation(
   }
   const temporaryDirectory =
     provider === "codex" &&
-    (classifyCodexProjectlessPath(resolved.project.root, env, input.classifierOptions) ||
+    (classifyCodexProjectlessPath(resolved.project.root, env, { ...input.classifierOptions, rawCwd: candidate }) ||
       classifyCodexManagedWorktreePath(resolved.project.root, env, input.classifierOptions));
   return {
     project: resolved.project,
@@ -175,7 +182,7 @@ export function resolveProviderProject(
   classifierOptions: {
     platform?: NodeJS.Platform;
     home?: string;
-    realpath?: typeof realpathSync;
+    realpath?: RealpathFn;
     stat?: typeof statSync;
   } = {},
 ): ContextProjectResolution {
@@ -199,7 +206,10 @@ export function resolveProviderProject(
   }
   const canonical = resolvePathProject(input.cwd, "codex_cwd_best_effort", classifierOptions);
   if (canonical.kind !== "path") return canonical;
-  const pathless = classifyCodexProjectlessPath(canonical.project.root, env, classifierOptions);
+  const pathless = classifyCodexProjectlessPath(canonical.project.root, env, {
+    ...classifierOptions,
+    rawCwd: input.cwd,
+  });
   if (pathless) return { kind: "pathless", project: { kind: "pathless" }, source: "codex_documents_v1" };
   return canonical;
 }
@@ -207,7 +217,17 @@ export function resolveProviderProject(
 export function classifyCodexProjectlessPath(
   cwd: string,
   env: NodeJS.ProcessEnv = process.env,
-  options: { platform?: NodeJS.Platform; home?: string; realpath?: typeof realpathSync } = {},
+  options: {
+    platform?: NodeJS.Platform;
+    home?: string;
+    realpath?: RealpathFn;
+    /**
+     * The caller's ORIGINAL, pre-`realpath` form of `cwd`. Supplying it keeps this
+     * classification off the filesystem for every path that is not already
+     * lexically inside a Codex scratch base — see {@link shouldCanonicalizeBase}.
+     */
+    rawCwd?: string;
+  } = {},
 ): boolean {
   const platform = options.platform ?? process.platform;
   const pathApi = platform === "win32" ? win32 : posix;
@@ -222,6 +242,7 @@ export function classifyCodexProjectlessPath(
   const normalizedCwd = normalizeForComparison(cwd, platform);
   const comparableBases = new Set(bases);
   for (const base of bases) {
+    if (!shouldCanonicalizeBase(base, options.rawCwd, platform)) continue;
     try {
       comparableBases.add((options.realpath ?? realpathSync)(base));
     } catch {
@@ -231,25 +252,55 @@ export function classifyCodexProjectlessPath(
     }
   }
   for (const base of comparableBases) {
-    const relativePath = pathApi.relative(normalizeForComparison(base, platform), normalizedCwd);
-    if (
-      !relativePath ||
-      relativePath === ".." ||
-      relativePath.startsWith(`..${pathApi.sep}`) ||
-      pathApi.isAbsolute(relativePath)
-    ) {
-      continue;
-    }
+    const relativePath = relativeInside(base, normalizedCwd, platform);
+    if (relativePath === null) continue;
     const [date, slug] = relativePath.split(pathApi.sep);
     if (date && slug && validIsoDate(date)) return true;
   }
   return false;
 }
 
+/**
+ * Should `base` be canonicalized before comparison?
+ *
+ * Canonicalizing means touching the path, and on macOS `~/Documents` sits behind
+ * a TCC "Files & Folders" consent prompt — so an ordinary project cwd must not
+ * cause it. The canonical form only ever matters when the base itself is
+ * redirected (macOS iCloud "Desktop & Documents Folders" makes `~/Documents` a
+ * symlink), and in that case the caller's original cwd is still lexically inside
+ * the logical base. So gate on that, and fall back to canonicalizing whenever
+ * the original cwd is unknown or relative, which preserves the prior behavior
+ * for callers that cannot supply it.
+ */
+function shouldCanonicalizeBase(base: string, rawCwd: string | undefined, platform: NodeJS.Platform): boolean {
+  const pathApi = platform === "win32" ? win32 : posix;
+  if (rawCwd === undefined || !pathApi.isAbsolute(rawCwd)) return true;
+  return relativeInside(base, normalizeForComparison(rawCwd, platform), platform) !== null;
+}
+
+/**
+ * The path of `candidate` relative to `root` when it is strictly inside it,
+ * otherwise `null`. `root` is normalized here; `candidate` is expected to be
+ * normalized already.
+ */
+function relativeInside(root: string, candidate: string, platform: NodeJS.Platform): string | null {
+  const pathApi = platform === "win32" ? win32 : posix;
+  const relativePath = pathApi.relative(normalizeForComparison(root, platform), candidate);
+  if (
+    !relativePath ||
+    relativePath === ".." ||
+    relativePath.startsWith(`..${pathApi.sep}`) ||
+    pathApi.isAbsolute(relativePath)
+  ) {
+    return null;
+  }
+  return relativePath;
+}
+
 export function classifyCodexManagedWorktreePath(
   cwd: string,
   env: NodeJS.ProcessEnv = process.env,
-  options: { platform?: NodeJS.Platform; home?: string; realpath?: typeof realpathSync } = {},
+  options: { platform?: NodeJS.Platform; home?: string; realpath?: RealpathFn } = {},
 ): boolean {
   const platform = options.platform ?? process.platform;
   const pathApi = platform === "win32" ? win32 : posix;
@@ -269,15 +320,8 @@ export function classifyCodexManagedWorktreePath(
 
   const normalizedCwd = normalizeForComparison(cwd, platform);
   for (const root of comparableRoots) {
-    const relativePath = pathApi.relative(normalizeForComparison(root, platform), normalizedCwd);
-    if (
-      !relativePath ||
-      relativePath === ".." ||
-      relativePath.startsWith(`..${pathApi.sep}`) ||
-      pathApi.isAbsolute(relativePath)
-    ) {
-      continue;
-    }
+    const relativePath = relativeInside(root, normalizedCwd, platform);
+    if (relativePath === null) continue;
     const [worktreeId, repository] = relativePath.split(pathApi.sep);
     if (worktreeId && repository) return true;
   }
@@ -312,7 +356,7 @@ function resolvePathProject(
   path: string,
   source: Extract<ContextProjectResolution, { kind: "path" }>["source"],
   dependencies: {
-    realpath?: typeof realpathSync;
+    realpath?: RealpathFn;
     stat?: typeof statSync;
   } = {},
 ): ContextProjectResolution {
--- a/packages/client/src/__tests__/install-locations.test.ts
+++ b/packages/client/src/__tests__/install-locations.test.ts
@@ -1,12 +1,26 @@
 import { chmodSync, mkdirSync, mkdtempSync, realpathSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { findCodexExecutableOnPath } from "../runtime/codex-binary.js";
 import { versionManagerBinDirs, wellKnownBinDirs } from "../runtime/install-locations.js";
 
+const NEVER_LISTED = (path: string): string[] => {
+  throw new Error(`unexpected readdir: ${path}`);
+};
+
 describe("versionManagerBinDirs", () => {
-  afterEach(() => vi.unstubAllEnvs());
+  // Pin a non-macOS baseline: on macOS these synthetic roots go through
+  // protected-root resolution, which follows real symlinks (`/home` is a
+  // firmlink there) and would rewrite them. The macOS cases opt in below.
+  beforeEach(() => {
+    Object.defineProperty(process, "platform", { value: "linux" });
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+    Object.defineProperty(process, "platform", { value: "linux" });
+  });
 
   it("lists nvm and fnm version bins, newest version first", () => {
     const home = "/home/u";
@@ -16,7 +30,7 @@ describe("versionManagerBinDirs", () => {
       throw new Error("ENOENT");
     };
 
-    expect(versionManagerBinDirs(home, readDir)).toEqual([
+    expect(versionManagerBinDirs(home, { readDir, env: {} })).toEqual([
       // Numeric-aware, so v22 sorts above v9 rather than below it.
       join(home, ".nvm", "versions", "node", "v22.2.0", "bin"),
       join(home, ".nvm", "versions", "node", "v20.11.0", "bin"),
@@ -28,7 +42,6 @@ describe("versionManagerBinDirs", () => {
 
   it("covers the Homebrew and pre-XDG fnm layouts and honours $FNM_DIR", () => {
     const home = "/home/u";
-    vi.stubEnv("FNM_DIR", "/opt/fnm");
     const readDir = (path: string): string[] => {
       if (path === join("/opt/fnm", "node-versions")) return ["v22.0.0"];
       if (path === join(home, "Library", "Application Support", "fnm", "node-versions")) return ["v21.0.0"];
@@ -36,7 +49,7 @@ describe("versionManagerBinDirs", () => {
       throw new Error("ENOENT");
     };
 
-    expect(versionManagerBinDirs(home, readDir)).toEqual([
+    expect(versionManagerBinDirs(home, { readDir, env: { FNM_DIR: "/opt/fnm" } })).toEqual([
       join("/opt/fnm", "node-versions", "v22.0.0", "installation", "bin"),
       join(home, "Library", "Application Support", "fnm", "node-versions", "v21.0.0", "installation", "bin"),
       join(home, ".fnm", "node-versions", "v20.0.0", "installation", "bin"),
@@ -45,23 +58,70 @@ describe("versionManagerBinDirs", () => {
 
   it("returns nothing when no version manager is installed", () => {
     expect(
-      versionManagerBinDirs("/home/u", () => {
-        throw new Error("ENOENT");
+      versionManagerBinDirs("/home/u", {
+        readDir: () => {
+          throw new Error("ENOENT");
+        },
+        env: {},
       }),
     ).toEqual([]);
   });
 
-  it("appends version dirs after every fixed location, leaving precedence intact", () => {
-    const home = "/home/u";
-    const fixed = wellKnownBinDirs(home, () => {
-      throw new Error("ENOENT");
-    });
-    const withVersions = wellKnownBinDirs(home, (path) =>
-      path === join(home, ".nvm", "versions", "node") ? ["v22.0.0"] : [],
-    );
+  it("keeps version dirs out of wellKnownBinDirs, whose precedence is above the login shell", () => {
+    // Well-known dirs are searched BEFORE the login-shell PATH, so a newest-first
+    // version list there would outrank the version the shell actually selected.
+    // The fallback belongs after the login shell instead.
+    const dirs = wellKnownBinDirs("/home/u");
+    expect(dirs.some((dir) => dir.includes(".nvm"))).toBe(false);
+    expect(dirs.some((dir) => dir.includes("fnm"))).toBe(false);
+  });
+});
 
-    expect(withVersions.slice(0, fixed.length)).toEqual(fixed);
-    expect(withVersions).toContain(join(home, ".nvm", "versions", "node", "v22.0.0", "bin"));
+// A version-manager root is not trusted for being spelled like one. `$FNM_DIR`
+// is user-controlled, and `~/.nvm`, a fnm data dir, or one version entry can be
+// a symlink — or sit under one — pointing into a protected folder. Listing such
+// a root, or handing its `bin` dirs to an executable check, is the same TCC
+// access this change exists to remove.
+describe("versionManagerBinDirs protected-root vetting (macOS)", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+    Object.defineProperty(process, "platform", { value: "linux" });
+  });
+
+  function macHome(): string {
+    const root = realpathSync(mkdtempSync(join(tmpdir(), "ft-vm-")));
+    const home = join(root, "home");
+    mkdirSync(join(home, "Documents"), { recursive: true });
+    Object.defineProperty(process, "platform", { value: "darwin" });
+    vi.stubEnv("HOME", home);
+    return home;
+  }
+
+  it("never lists a $FNM_DIR pointing into a protected folder", () => {
+    const home = macHome();
+    expect(
+      versionManagerBinDirs(home, { readDir: NEVER_LISTED, env: { FNM_DIR: join(home, "Documents", "fnm") } }),
+    ).toEqual([]);
+  });
+
+  it("never lists a default root symlinked into a protected folder", () => {
+    const home = macHome();
+    mkdirSync(join(home, "Documents", "nvm", "versions", "node"), { recursive: true });
+    symlinkSync(join(home, "Documents", "nvm"), join(home, ".nvm"));
+
+    expect(versionManagerBinDirs(home, { readDir: NEVER_LISTED, env: {} })).toEqual([]);
+  });
+
+  it("drops a version entry symlinked into a protected folder, keeping its siblings", () => {
+    const home = macHome();
+    const versions = join(home, ".nvm", "versions", "node");
+    mkdirSync(join(versions, "v22.0.0", "bin"), { recursive: true });
+    mkdirSync(join(home, "Documents", "sneaky", "bin"), { recursive: true });
+    symlinkSync(join(home, "Documents", "sneaky"), join(versions, "v23.0.0"));
+
+    expect(
+      versionManagerBinDirs(home, { readDir: (path) => (path === versions ? ["v23.0.0", "v22.0.0"] : []), env: {} }),
+    ).toEqual([join(versions, "v22.0.0", "bin")]);
   });
 });
 
@@ -84,10 +144,11 @@ describe("version-manager discovery after a multishell teardown", () => {
     symlinkSync(installBin, multishell);
 
     // The login-shell seam models the probe shell's lifetime: by the time it has
-    // answered, its per-session dir no longer exists.
+    // answered, its per-session dir no longer exists — so only the stable
+    // version dir, appended after it, can still answer.
     const loginShellPathDirs = (): string[] => {
       rmSync(multishell, { force: true });
-      return [multishell];
+      return [multishell, ...versionManagerBinDirs(home, { env: {} })];
     };
 
     expect(
@@ -98,12 +159,48 @@ describe("version-manager discovery after a multishell teardown", () => {
           pathDelimiter: ":",
           loginShellPathDirs,
           // Only the version dirs, so a real binary on this host's `/usr/local`
-          // or Homebrew path cannot answer for the mechanism under test. That
-          // `wellKnownBinDirs` appends these is covered above.
-          wellKnownDirs: () => versionManagerBinDirs(home),
+          // or Homebrew path cannot answer for the mechanism under test.
+          wellKnownDirs: () => [],
           desktopAppDirs: () => [],
         },
       ),
     ).toBe(codex);
+  });
+
+  // The precedence the fallback must not disturb: an intentionally selected
+  // older version keeps its binary (and credential context) while a newer
+  // inactive version is also installed.
+  it("prefers the shell's active version over a newer installed one", () => {
+    const root = realpathSync(mkdtempSync(join(tmpdir(), "ft-active-")));
+    const home = join(root, "home");
+    const versions = join(home, ".nvm", "versions", "node");
+    const activeBin = join(versions, "v20.11.0", "bin");
+    const newerBin = join(versions, "v22.2.0", "bin");
+    mkdirSync(activeBin, { recursive: true });
+    mkdirSync(newerBin, { recursive: true });
+    for (const bin of [activeBin, newerBin]) {
+      const codex = join(bin, "codex");
+      writeFileSync(codex, "#!/bin/sh\n");
+      chmodSync(codex, 0o755);
+    }
+
+    const fallback = versionManagerBinDirs(home, { env: {} });
+    // Newest-first, so the fallback alone would pick v22.
+    expect(fallback[0]).toBe(newerBin);
+
+    expect(
+      findCodexExecutableOnPath(
+        { HOME: home, PATH: "" },
+        {
+          platform: "linux",
+          pathDelimiter: ":",
+          // What `getLoginShellPathDirs` composes: the live selection first,
+          // the stable roots appended behind it.
+          loginShellPathDirs: () => [activeBin, ...fallback],
+          wellKnownDirs: () => [],
+          desktopAppDirs: () => [],
+        },
+      ),
+    ).toBe(join(activeBin, "codex"));
   });
 });
--- a/packages/client/src/__tests__/install-locations.test.ts
+++ b/packages/client/src/__tests__/install-locations.test.ts
@@ -4,6 +4,7 @@ import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { findCodexExecutableOnPath } from "../runtime/codex-binary.js";
 import { versionManagerBinDirs, wellKnownBinDirs } from "../runtime/install-locations.js";
+import { getLoginShellPathDirs, resetLoginShellPathDirsCache } from "../runtime/login-shell-path.js";
 
 const NEVER_LISTED = (path: string): string[] => {
   throw new Error(`unexpected readdir: ${path}`);
@@ -327,6 +328,48 @@ describe("version-manager discovery after a multishell teardown", () => {
         },
       ),
     ).toBe(codex);
+  });
+
+  // End to end, through the real `getLoginShellPathDirs` composition: two
+  // manager variables reported, the fnm per-session entry already gone, and a
+  // runnable binary sitting under `$NVM_BIN`. Withholding nvm only from the
+  // appended fallback would prove nothing here — the raw `$PATH` lists it, so
+  // the resolver would skip the dead entry and launch nvm. Missing is the
+  // correct answer: the live shell had selected fnm.
+  it("resolves nothing rather than nvm when both managers were reported", () => {
+    const root = realpathSync(mkdtempSync(join(tmpdir(), "ft-dual-")));
+    const home = join(root, "home");
+    const nvmBin = join(home, ".nvm", "versions", "node", "v22.2.0", "bin");
+    const fnmDir = join(root, "custom-fnm");
+    const multishell = join(root, "fnm_multishells", "88_2", "bin");
+    mkdirSync(nvmBin, { recursive: true });
+    mkdirSync(join(fnmDir, "node-versions", "v20.11.0", "installation", "bin"), { recursive: true });
+    mkdirSync(home, { recursive: true });
+    const nvmCodex = join(nvmBin, "codex");
+    writeFileSync(nvmCodex, "#!/bin/sh\n");
+    chmodSync(nvmCodex, 0o755);
+
+    vi.stubEnv("HOME", home);
+    resetLoginShellPathDirsCache();
+    const dirs = getLoginShellPathDirs(
+      () =>
+        `__FT_SHELL_PATH__${multishell}\n${nvmBin}__FT_SHELL_PATH__${"__FT_SHELL_ENV__"}${fnmDir}\n${nvmBin}${"__FT_SHELL_ENV__"}`,
+    );
+    resetLoginShellPathDirsCache();
+
+    expect(dirs).not.toContain(nvmBin);
+    expect(
+      findCodexExecutableOnPath(
+        { HOME: home, PATH: "" },
+        {
+          platform: "linux",
+          pathDelimiter: ":",
+          loginShellPathDirs: () => dirs,
+          wellKnownDirs: () => [],
+          desktopAppDirs: () => [],
+        },
+      ),
+    ).toBeNull();
   });
 
   // The precedence the fallback must not disturb. With two versions installed
--- a/packages/client/src/__tests__/install-locations.test.ts
+++ b/packages/client/src/__tests__/install-locations.test.ts
@@ -336,7 +336,7 @@ describe("version-manager discovery after a multishell teardown", () => {
   // appended fallback would prove nothing here — the raw `$PATH` lists it, so
   // the resolver would skip the dead entry and launch nvm. Missing is the
   // correct answer: the live shell had selected fnm.
-  it("resolves nothing rather than nvm when both managers were reported", () => {
+  it("resolves nothing rather than nvm when both managers were reported (macOS)", () => {
     const root = realpathSync(mkdtempSync(join(tmpdir(), "ft-dual-")));
     const home = join(root, "home");
     const nvmBin = join(home, ".nvm", "versions", "node", "v22.2.0", "bin");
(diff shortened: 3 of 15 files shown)


yuezengwu (changes requested):
Blocking issue in packages/client/src/runtime/login-shell-path.ts:179-182: a PATH entry such as $HOME/bin symlinked to $HOME/Documents/bin is not matched by the lexical case guard, so the probe still executes cd and pwd -P on the protected target before the parsed-result filter runs. That traversal is itself the TCC-protected access this PR is meant to eliminate; dropping the canonicalized output only prevents the later existsSync checks. Please make the pre-canonicalization handling avoid traversing into protected symlink targets and add coverage for that case. As written, a fresh macOS install can still prompt for Documents access.

baixiaohang (changes requested):
- Rationale: The lexical pre-`cd` guard still traverses protected folders through symlinked `PATH` entries, so the primary TCC-safety guarantee is not yet enforced.

**Risk level**: A

**PR summary**
- Problem: Fresh macOS installs can receive unexplained Files & Folders consent prompts because automatic provider discovery and Codex project classification touch protected directories before the user selects a provider or enables Context integration.
- Approach: Skip protected login-shell `PATH` entries before canonicalization, filter protected canonical results before executable checks, and avoid canonicalizing the Codex Documents base for ordinary project directories while preserving redirected-Documents classification.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The pre-`cd` guard compares only the lexical spelling of each `PATH` entry. An entry such as `$HOME/bin` symlinked to `$HOME/Documents/bin` does not match the guard and still reaches `(cd "$d" ...)`; resolving that symlink is already an access to the protected target and can trigger the TCC prompt. The parsed-result filter prevents only the later provider `existsSync` calls, after the shell has entered the directory. The new real-shell test covers a symlink to an unprotected target, while the protected-target scenario is simulated only as already-parsed output, so it cannot catch this path. Please make the shell phase reject protected symlink targets before traversal (including symlinked ancestors), or otherwise avoid entering them, and add end-to-end coverage for the stated `$HOME/bin -> $HOME/Documents/bin` case. [R4 / `packages/client/src/runtime/login-shell-path.ts:182`]

**Action taken**
- Submitted request changes

Gandy2025:
Thanks — this was a real hole, and the ancestor variant you both called out is real too. Fixed in ae50101.

**Confirmed first.** I reproduced against the script as shipped in the first push, with `$HOME/bin -> $HOME/Documents/real` and `$HOME/deep/mid -> $HOME/Documents`:

```
BEFORE (first push)     shell entered Documents 2 time(s)
                        .../home/Documents/real/bin   (symlinked entry)
                        .../home/Documents/real/bin   (symlinked ancestor)
AFTER  (this revision)  shell entered Documents 0 time(s)
```

Both entries resolved, so `cd` had already performed the protected access before anything downstream could filter it. The parsed-result filter only ever prevented the later `existsSync`.

**Why the guard could not be patched in place.** The spelling of a path cannot answer where it lands, and every operation that *can* answer it — `cd`, `realpath`, `existsSync` — answers by performing the access we are trying to avoid. `readlink` is the only exception: it reads a link's own target without following it.

**What changed.**

The shell no longer canonicalizes arbitrary entries. It canonicalizes only the per-session multishell dirs that genuinely need in-shell resolution (their symlink is torn down when the probe shell exits, so Node could not follow it afterwards). Everything else is emitted verbatim with no filesystem access at all.

Those dirs are matched by tool-specific directory name (`fnm_multishells` / `nvm_multishells`), not by "somewhere under a temp root". I tried the temp-root form first and the new end-to-end test caught it immediately: the test's `$HOME` lives under `/private/var/folders`, so a broad temp pattern matched the symlinked entries and re-admitted the exact traversal. A temp root is not a safety property.

Resolution moves to `resolveOutsideProtectedRoots`, which walks the path one component at a time from the root, checks each component against the protected roots *before* touching it, and only ever calls `readlink` on it. A symlink target is re-walked from the root, so an expansion landing in a protected root is caught on the next iteration — that is what covers the symlinked-ancestor case rather than a separate branch for it. Cycles are bounded.

**Coverage** (`packages/client/src/__tests__/login-shell-path.test.ts`):

- `never yields a protected dir, however the entry reaches one` — end-to-end over the real shell **and** the real parse, with all three routes in one `$PATH`: by spelling, symlinked entry (`$HOME/bin/bin`), symlinked ancestor (`$HOME/deep/mid/real/bin`). It asserts on the raw shell output that no resolved path passes through Documents, which is a direct observation that no `cd` entered it — not just that the final list is clean. Same run asserts the multishell dir is still canonicalized in-shell.
- `rejects symlinks into a protected root without entering them` — the resolution in isolation, including a symlink that stays outside and must still resolve to its target.
- `does not loop forever on a symlink cycle`.

**One thing worth flagging.** Writing this surfaced that the hermetic parse tests were reading the host's real `process.platform` on the first test of the file. On macOS the new resolution follows real symlinks (`/home` is a firmlink), which rewrote a synthetic path and failed the test. Added a `beforeEach` pinning a non-macOS baseline so those tests are host-independent; the macOS tests opt in explicitly.

**Cost of the narrowing**, stated plainly: a per-session symlink dir created by some tool *other* than fnm/nvm is no longer canonicalized in-shell and will not resolve after the probe exits. That is the deliberate trade for not entering paths we cannot vet first.

Verification: `pnpm check` and `pnpm typecheck` clean; `packages/client` full suite 2421 passed / 7 skipped. `apps/cli` is untouched by this revision. The QA case now lists all three routes as preconditions, since a guard can close one and leave the others open.

yuezengwu (changes requested):
Two blockers remain on ae50101:

1. packages/client/src/runtime/login-shell-path.ts:197 still executes cd for every PATH spelling that matches */fnm_multishells/* or */nvm_multishells/*. That spelling is not a safety boundary: for example, /tmp/fnm_multishells/session/bin symlinked to $HOME/Documents/bin reaches the protected target before the Node walker runs. This narrows the original hole by name but does not eliminate it. Please ensure the shell never follows an unvetted PATH symlink, including the multishell shape, and cover a multishell-named link whose target is protected.

2. packages/client/src/runtime/login-shell-path.ts:253 compares protected roots case-sensitively. On the default case-insensitive macOS filesystem, $HOME/documents/bin resolves to $HOME/Documents/bin but does not match the guard, so the walker proceeds to readlinkSync on the protected path. Please make the macOS protected-root comparison conservative with respect to casing and add mixed-case coverage.

The new component-wise readlink handling fixes the ordinary symlink and symlinked-ancestor cases, but these two paths still violate the stated zero-protected-access guarantee.

baixiaohang (changes requested):
- Rationale: The component-wise walker fixes the reported symlink paths, but two remaining branches can still enter a protected root before the safety filter takes effect.

**Risk level**: A

**PR summary**
- Problem: Fresh macOS installs can receive unexplained Files & Folders consent prompts because automatic provider discovery and Codex project classification touch protected directories before the user selects a provider or enables Context integration.
- Approach: Keep arbitrary login-shell `PATH` entries out of shell canonicalization, resolve them component by component with `readlink`, and avoid canonicalizing the Codex Documents base for ordinary project directories while preserving redirected-Documents classification.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The Darwin shell still runs `cd` for every entry whose spelling matches `*/fnm_multishells/*` or `*/nvm_multishells/*`. That directory name is not a safety boundary: `/tmp/fnm_multishells/session/bin` can itself point to `$HOME/Documents/bin`, including through a configured fnm/nvm install root, and the `cd` reaches the protected target before the Node walker can reject it. This narrows the previous hole by name without enforcing the zero-protected-access guarantee. Please keep the shell from following any unvetted `PATH` symlink, including the multishell shape, and add end-to-end coverage for a multishell-named link targeting Documents. [R4 / `packages/client/src/runtime/login-shell-path.ts:197`]

❌ 2. `insideAny` compares protected roots case-sensitively. On the default case-insensitive macOS filesystem, a spelling such as `$HOME/documents/bin` resolves to `$HOME/Documents/bin` but misses both the shell guard and this comparison, so the walker touches the protected path and later executable checks can follow it. Please make the Darwin protected-root comparison conservative with respect to casing and add mixed-case coverage. [R4 / `packages/client/src/runtime/login-shell-path.ts:253`]

✅ 3. The new component-wise `readlink` walk correctly closes the originally reported ordinary symlink and symlinked-ancestor paths, and the revised end-to-end test now exercises both.

**Action taken**
- Submitted request changes

Gandy2025:
Both findings are correct, and they are the same mistake twice on my part: I treated a *spelling* as a safety boundary. Finding 1 is exactly the objection I raised against my own temp-root version one revision earlier — `*/fnm_multishells/*` is no more of a boundary than `/tmp/*`, and a link with that name can point at Documents just as easily as at a node install. Finding 2 is a clean miss: the default macOS filesystem is case-insensitive, so `$HOME/documents/bin` **is** `$HOME/Documents/bin`.

Fixed in 8b3fc27, by stopping the search for a third spelling to trust.

**The macOS script now performs no filesystem access at all.** It prints `$PATH`, one entry per line, and nothing else — no `cd`, no `pwd`, no `readlink`. All resolution moved into `resolveOutsideProtectedRoots`, which vets each component against the protected roots before touching it and only ever calls `readlink`. There is no longer a branch that could enter a path we have not already vetted, so the guarantee is structural rather than argued, and it lives in one function that unit tests reach directly. The Darwin script also ends up simpler than the one this PR started from.

Protected-root comparison is case-folded. On a case-sensitive volume that can only over-reject, which is the correct direction for a guard like this.

**Evidence, since "the ordering is right" is exactly what I claimed last round.** Resolution now takes an injectable `readlink` seam, and a test asserts that no protected path is ever passed to it. That is a direct observation of what was touched, not an argument about control flow.

The end-to-end test now puts all four routes into one `$PATH` — by spelling, symlinked entry, symlinked ancestor, and a `fnm_multishells`-named link pointing at Documents — and asserts the shell echoes every entry **byte for byte**. Four of those five entries are symlinks whose resolved spelling differs from what was written, so any `cd` would show up as a changed line. That replaces the previous "no output contains Documents" assertion, which stopped being meaningful once entries are echoed verbatim (one input is literally spelled `Documents`). Added alongside: mixed-case coverage, and an assertion that the Darwin script contains none of `cd` / `pwd` / `readlink`.

**Cost.** Unchanged in kind, but it now applies to any per-session symlink dir rather than just non-fnm ones: a dir already torn down by the time the probe returns can no longer be followed. Resolution runs immediately after the shell exits, so one still present at that moment resolves exactly as before. If fnm coverage turns out to matter in practice, the right follow-up is adding fnm's stable install roots to `wellKnownBinDirs` — a separate, testable change — not re-admitting an unvetted traversal here.

Verification: `pnpm check` and `pnpm typecheck` clean; `packages/client` full suite 2424 passed / 7 skipped. `apps/cli` untouched by this revision.

baixiaohang (changes requested):
- Rationale: The new Darwin path is structurally TCC-safe, but moving all canonicalization after shell exit reintroduces the known fnm/nvm multishell discovery failure that this probe was specifically changed to prevent.

**Risk level**: A

**PR summary**
- Problem: Fresh macOS installs can receive unexplained Files & Folders consent prompts because automatic provider discovery and Codex project classification touch protected directories before the user selects a provider or enables Context integration.
- Approach: Make the Darwin login-shell script echo `PATH` without filesystem access, resolve entries component by component with a case-folded protected-root guard, and avoid canonicalizing the Codex Documents base for ordinary project directories.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. This knowingly reintroduces the macOS fnm/nvm false negative fixed by #1314/#1326. Lines 73–78 still document the established lifecycle: a multishell symlink is torn down when the login shell exits, so it must be resolved while that shell is alive; lines 80–86 now defer resolution until after `spawnSync` returns, when the link may already be gone. In that case `resolveOutsideProtectedRoots` returns the dead lexical path and every provider search reports the login-shell-only binary missing. The new end-to-end test leaves `multishellSafe` alive after the shell exits, so it cannot exercise this regression, while the committed QA contract still requires an nvm/fnm-only provider to be discovered. "Add stable roots later" is therefore not an out-of-scope enhancement; it is needed here (or use a pre-exit mechanism that captures link targets without following them) to preserve the PR's explicit discovery invariant. Please cover a seam that removes the multishell link before `runShell` returns and prove the stable target is still returned. [R5 / `packages/client/src/runtime/login-shell-path.ts:73`]

✅ 2. The previous multishell-name trust and mixed-case bypasses are resolved: the Darwin script performs no path traversal, and all protected-root decisions now live behind the case-folded component walker.

**Action taken**
- Submitted request changes

yuezengwu (changes requested):
Two blockers remain on 8b3fc27:

1. packages/client/src/runtime/login-shell-path.ts:217 still expands PATH unquoted in for d in $PATH. POSIX expansion performs pathname expansion after field splitting, so a literal entry such as $HOME/Documents/* makes the nested shell enumerate the protected directory before printf runs. The Darwin script therefore does perform filesystem access even though it contains no explicit cd/pwd/readlink. Please disable globbing before this expansion (or use another non-globbing split) and add coverage proving a wildcard-bearing PATH entry is echoed literally without being expanded.

2. packages/client/src/runtime/login-shell-path.ts:73-86 documents that fnm/nvm multishell links can be torn down when the login shell exits, but Darwin now resolves only after that exit. In that case readlink sees a missing path and the provider is no longer discovered. The new end-to-end fixture leaves its multishell symlink alive, so it does not model the stated teardown. This contradicts the PR description that fnm/nvm discovery remains unchanged and the QA acceptance criterion that a login-shell-only provider is still found. Please either preserve that discovery through a safe stable-root mechanism in this change, or explicitly revise the accepted behavior/PR/QA and cover the real teardown case.

The case-folding fix and the component-wise readlink handling for ordinary symlinks are sound, but the primary no-access claim and the stated discovery-compatibility claim are not both met yet.

Gandy2025:
Both correct. Fixed in e61d91b.

**1. `$PATH` globbing.** I missed this entirely, and it is worse than an access — it is a *read of the directory listing*. Reproduced under `/bin/sh`:

```
PATH='/tmp/ftglob/Documents/*'
  without set -f:   saw: /tmp/ftglob/Documents/secret-a
                    saw: /tmp/ftglob/Documents/secret-b
  with    set -f:   saw: /tmp/ftglob/Documents/*
```

So the entry names inside a protected folder were being pulled into the probe result, with no `cd` / `pwd` / `readlink` anywhere in the script — which is exactly why "the script contains no filesystem commands" was not a sufficient test, and why the new coverage asserts the wildcard entry is echoed literally and its directory never enumerated. Worth noting this predates the PR; the no-access claim made here is what puts it in scope. Covered on both platforms since both split the same way.

**2. fnm/nvm discovery.** You are right that I traded away the invariant this probe exists for, and that "add stable roots later" does not survive contact with the PR's own claims. I took the first of your two options.

`versionManagerBinDirs` now enumerates where the binary actually lives and stays: `~/.nvm/versions/node/*/bin`, plus fnm's XDG, Homebrew, pre-XDG and `$FNM_DIR` layouts. #1314 recorded that `wellKnownBinDirs` "cannot enumerate version-manager dirs" — reading it again, that was a shape choice rather than a constraint. These roots belong to the version manager and never to a protected folder, so listing them during automatic probing is safe, and unlike the shell it is directly unit-testable. They are appended after every fixed location, so existing precedence is untouched and they only decide an otherwise-empty result.

The teardown seam you asked for is in `install-locations.test.ts`: the login-shell seam *deletes* the multishell symlink before returning, and the assertion is that `findCodexExecutableOnPath` still returns the binary from the stable version dir. That models the real lifecycle rather than leaving the link alive.

I did consider the other option — a pre-exit resolver in the shell. It is implementable (component-wise `readlink` with a pre-touch guard, which becomes sound once each prefix is known not to be a symlink), but it needs case-insensitive matching without a subprocess per component, `..` normalization, and it puts the security boundary back into the layer where four defects have already been found in this PR, and out of the layer that has tests. The stable-root mechanism gets the same invariant with none of that.

The PR description is rewritten — it still described the `case`-guard design and claimed fnm behavior was unchanged, neither of which is true now. The QA case preconditions now call for fnm `--use-on-cd` specifically (so discovery must come from the stable dir, not the session link) and for a wildcard `$PATH` entry.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2431 passed / 7 skipped; `apps/cli` suite re-running since `wellKnownBinDirs` is shared by all seven resolvers — I will report if anything moves.

yuezengwu (changes requested):
Two blockers remain on e61d91b:

1. packages/client/src/runtime/install-locations.ts:29-53 again treats path spelling as the safety boundary. FNM_DIR is explicitly user-controlled, and the default .nvm/fnm roots or an individual version entry can also be symlinks into $HOME/Documents. versionManagerBinDirs calls readdirSync on those roots before any protected-aware component walk, and the returned bin dirs are later probed with executable existence checks, so this new automatic source can trigger the same TCC access the PR is meant to remove. Please vet every root and returned candidate through the no-follow protected-root resolver before readdir/search (including symlinked ancestors), and cover FNM_DIR plus a default root/version symlinked into a protected folder.

2. packages/client/src/runtime/install-locations.ts:70-88 appends every installed version to wellKnownBinDirs, but all provider resolvers search that list before login-shell PATH. With codex installed under active nvm v20 and also under newer inactive v22, the new numeric newest-first list selects v22 before the login shell can preserve v20. This contradicts the existing resolver invariant that an intentional nvm/fnm selection keeps its selected version and credential context. Please search stable version roots only after login-shell dirs as a teardown fallback, or otherwise preserve the active selection, and add coverage where active and newest versions differ plus the teardown fallback case.

The set -f globbing fix and the teardown fixture are sound; the stable-root replacement needs the same protected-path boundary and the correct precedence before it can safely preserve discovery.

baixiaohang (changes requested):
- Rationale: `set -f` closes the wildcard traversal, but the new stable-version fallback introduces an unguarded filesystem source and changes which nvm/fnm version wins.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery can trigger unexplained macOS Files & Folders prompts before the user selects a provider, while any fix must still find providers installed under an ephemeral nvm/fnm login-shell path.
- Approach: Make the Darwin shell probe echo PATH entries without traversing them, resolve ordinary entries through a protected-root-aware component walker, enumerate stable nvm/fnm version roots as a fallback, and avoid probing the Codex Documents scratch base for unrelated projects.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. `versionManagerBinDirs` reintroduces the same protected-folder access through a new automatic source. `$FNM_DIR` is user-controlled, and the default `.nvm` / fnm roots or any listed version directory can also be symlinked (including through an ancestor) into `$HOME/Documents`. Lines 33 and 51 call `readdirSync` before any protected-aware walk, and the returned `bin` paths are later passed to executable checks, so either stage can trigger TCC. The assertion that these roots are never protected is therefore not an enforceable boundary. Please vet every root and returned candidate without following into a protected root before listing or searching it, and cover a protected `$FNM_DIR`, a default root symlink, and a version-entry symlink. [R4 / `packages/client/src/runtime/install-locations.ts:29`]

❌ 2. Appending every installed version to `wellKnownBinDirs` changes provider selection, because every resolver searches well-known dirs before the login-shell PATH. If the shell intentionally selects nvm v20 while a newer inactive v22 also contains `codex` or `claude`, the numeric newest-first list returns v22 before the login-shell source gets a chance to preserve v20 and its credential context. Appending after fixed well-known entries does not preserve the cross-source precedence documented by the resolvers. Please keep stable version roots as a teardown fallback after login-shell dirs (or otherwise preserve the active selection), and cover active-v20/newest-v22 plus the existing teardown case. [R5 / `packages/client/src/runtime/install-locations.ts:70`]

✅ 3. The `set -f` change prevents wildcard PATH entries from enumerating protected directories, and the revised real-shell coverage observes the literal result directly.

**Action taken**
- Submitted request changes

Gandy2025:
Both correct, and the first one is the **fourth** time in this PR that I have treated a path's spelling as a boundary — temp root, then multishell directory name, then case, now "these roots belong to the version manager". That last claim is not enforceable: `$FNM_DIR` is user-controlled, and `~/.nvm` or a fnm data dir can be a symlink, or sit under one, pointing into `$HOME/Documents`. The `readdirSync` and the downstream executable checks are then exactly the access this PR exists to remove.

Fixed in 579326b.

**Finding 1 — vetting.** The no-follow resolver moved to `runtime/protected-paths.ts`, so it is a shared boundary rather than a private detail of the login-shell probe. Its doc now states the rule these four rounds kept proving, so the next person does not have to rediscover it:

> EVERY automatic filesystem source must pass a path through here before listing, stat-ing, or searching it. Judging a source safe by where it is *spelled* does not hold.

`versionManagerBinDirs` vets each candidate root before listing it, and each produced `bin` dir before returning it — so a protected `$FNM_DIR`, a symlinked default root, and a single symlinked version entry are all rejected without being read. The tests pass a `readDir` that **throws if a rejected root is ever listed**, so "we didn't touch it" is asserted rather than argued, and the version-entry case checks that a poisoned sibling does not take its healthy neighbours down with it.

**Finding 2 — precedence.** You are right, and my "appended after fixed entries" reasoning was answering the wrong question: ordering *within* `wellKnownBinDirs` is irrelevant when the whole list is searched before the login-shell PATH. Version dirs are out of that list entirely — there is now a test asserting they never appear in it — and are appended after the probed dirs inside `getLoginShellPathDirs`. That is the one place the composition belongs: the live selection first, the stable roots only as the teardown fallback. Covered by an active-v20 / newest-v22 case that asserts v20 wins even though the fallback alone would return v22, alongside the existing teardown case.

Verification: `pnpm check` and `pnpm typecheck` clean; `packages/client` 2435 passed / 7 skipped; `apps/cli` re-running and I will report if anything moves.

One process note, since this is now four rounds: the recurring defect has been a single one — reaching for a cheaper test than "resolve it safely and see" — and each round I fixed the instance rather than the class. Putting the resolver behind one shared module with that rule written on it is the part of this change meant to stop the pattern, not just the current instance.

yuezengwu (changes requested):
One blocker remains on 579326b:

`packages/client/src/runtime/install-locations.ts:19-34` still returns the fixed well-known directories without passing them through the new boundary, and the provider resolvers immediately `statSync` / `accessSync` candidates from those directories before `getLoginShellPathDirs` runs (for example `packages/client/src/runtime/codex-binary.ts:224-250` and `packages/client/src/handlers/claude-executable.ts:117-125`). A spelling such as `~/.local/bin` is not safer than the version-manager spellings fixed in this revision: `~/.local` or `bin` can be a symlink into `~/Documents`, so the automatic well-known search still reaches the protected target on a fresh install.

There is also a leaf form of the same hole: even when the directory itself is safely resolved, `~/.local/bin/codex` (or a login-shell/version-manager candidate) can itself be a symlink to `~/Documents/codex`. The current code vets only the directory and then appends the binary name; `resolveSpawnableCodexCandidate` / `isExecutableFile` follows the final symlink with `statSync` or `accessSync` before any protected-root check. That violates the new module's own rule that every automatic filesystem source must pass through the no-follow resolver before stat-ing or searching it.

Please make the shared boundary apply to the complete executable candidate immediately before any automatic `stat` / `access` across the provider resolvers (the explicit operator override can remain a separate, user-directed path). That will cover daemon PATH, fixed/provider-specific well-known dirs, login-shell dirs, desktop-app candidates, and final executable symlinks without trusting any source spelling. Please add coverage for both a well-known directory/ancestor symlink into Documents and a provider executable symlink into Documents, asserting that the protected path is never passed to the executable check.

The two findings from e61d91b are otherwise resolved: version-manager roots and produced bins are vetted before listing/search, and stable versions now sit after the live login-shell selection so an active older version keeps precedence.

baixiaohang (changes requested):
- Rationale: The shared protected-path module fixes the reported version-root traversal, but automatic executable checks still bypass it, and the stable fallback still loses shell-only selection metadata.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery must avoid macOS protected-folder prompts before user action without regressing providers installed behind an ephemeral nvm/fnm login-shell path.
- Approach: Make the Darwin shell probe non-traversing, centralize protected-root resolution in a shared no-follow walker, and append vetted stable nvm/fnm version directories after the live login-shell result.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The new boundary is still bypassed by the actual executable checks. `wellKnownBinDirs` returns fixed spellings such as `~/.local/bin` without vetting them, but an ancestor or the directory itself can be symlinked into `~/Documents`; each provider resolver then immediately `statSync` / `accessSync`s the joined candidate before `getLoginShellPathDirs` runs. Even for a safely resolved login-shell or version-manager directory, the final `codex` / `claude` file can itself be a symlink into a protected root, because only the directory is vetted before the binary name is appended. This contradicts `protected-paths.ts`'s own rule that every automatic source must cross the boundary before stat/search. Please vet the complete executable candidate immediately before every automatic executable check across all provider resolvers (keeping an explicit operator override separate if desired), and cover both a well-known directory/ancestor symlink and a final executable symlink while asserting that no protected path reaches the check. [R4 / `packages/client/src/runtime/install-locations.ts:19`]

❌ 2. The active-version regression is not covered in the teardown state that motivates the fallback. On Darwin, the shell deliberately emits the raw multishell path; if that link is removed before `spawnSync` returns, lines 87–98 retain only the dead lexical entry and then append `versionManagerBinDirs` newest-first. With active fnm v20 and inactive v22 both containing `codex`, the dead entry misses and v22 still wins. The new precedence test bypasses this by injecting `activeBin` itself at line 199 — the stable path the Darwin probe can no longer produce after teardown — while the teardown test has only one installed version. Please combine both conditions in one test (active v20 through a removed multishell link plus installed v22) and preserve the selected stable target, e.g. by safely capturing sufficient no-follow metadata before the shell exits. [R5 / `packages/client/src/runtime/login-shell-path.ts:87`]

❌ 3. The `$FNM_DIR` fallback reads the daemon's `process.env`, not the environment created by the login shell. `defaultRunShell` sources the rc files, but its protocol returns only `PATH`; `getLoginShellPathDirs` then calls `versionManagerBinDirs(home, { readLink })`, so a custom `FNM_DIR` exported by `fnm env` inside `.zshrc` is unavailable to the fallback. Once that custom multishell link is torn down, only the hard-coded default roots are scanned and the provider becomes missing. The helper-level test injects `env: { FNM_DIR: "/opt/fnm" }` directly and therefore cannot catch the integration gap. Please carry the login shell's custom stable root through the safe protocol (or otherwise preserve it) and cover a shell-only custom `FNM_DIR` whose multishell link disappears before return. [R5 / `packages/client/src/runtime/login-shell-path.ts:96`]

✅ 4. Moving the component walker into `protected-paths.ts` and vetting both each version-listing root and each returned version-manager directory correctly resolves the prior `$FNM_DIR` / root-symlink traversal finding.

**Action taken**
- Submitted request changes

Gandy2025:
All three correct. Pushed in fc5520f — with one gap I have deliberately left open and want a decision on rather than guessing at.

**Finding 1 — the boundary was in the wrong place.** I had been vetting *directories*, which is not what gets stat-ed. `~/.local/bin` is not a safer spelling than a version-manager root, and even a safely-resolved directory tells you nothing about the leaf. Vetting moved to the complete candidate, immediately before every automatic executable check — the five per-provider `isExecutableFile` copies, codex's `isExecutable` / `fileExists`, and claude's exported one. That is a single rule applied at seven call sites rather than a per-source judgement, so daemon PATH, fixed and provider-specific dirs, login-shell dirs, version roots and desktop-app candidates are all covered without trusting anyone's spelling. `CLAUDE_CODE_EXECUTABLE` passes `automatic: false`, since it is user-directed.

Covered by: a well-known dir whose *ancestor* is symlinked into Documents, and an executable that is *itself* such a symlink — both rejected — plus an ordinary binary and one symlinked to a safe target still resolving, so the gate is not just refusing everything.

**Findings 2 & 3 — the fallback could not see what the shell knew.** Both come down to the same thing, and you named the mechanism: capture it safely before the shell exits. Reading an environment variable costs *no filesystem access at all*, which is exactly why these are capturable while a symlink target is not.

The probe protocol now carries a second section with `$FNM_DIR` and `$NVM_BIN`. `$NVM_BIN` is nvm's **active** version's bin dir and a real directory, so it outlives the probe shell — it goes first in the fallback, ahead of the newest-first list, which is what preserves the selected version after teardown. `$FNM_DIR` feeds the fnm root list, so a root exported only inside `.zshrc` is no longer invisible. Both are vetted through the shared boundary before use, and the section is optional so an older-shaped output still parses. The new test is the real teardown state: a shell-only custom `$FNM_DIR` with the multishell link already gone, and both versions installed.

**The gap I am not closing unilaterally.** fnm exports no active-version variable — only the transient `$FNM_MULTISHELL_PATH`. So for fnm specifically, after teardown the newest installed version answers instead of the selected one. nvm is fully covered via `$NVM_BIN`.

Closing it for fnm needs a no-follow capture *inside* the probe shell: walking components from the root, checking each against the protected roots before `readlink`. That is implementable — the check becomes sound once each prefix is known not to be a symlink — but it needs case-insensitive matching without a subprocess per component and `..` normalisation, and it puts the security boundary back into the layer where five defects have now been found in this PR, out of the layer that has tests.

Given that, and that this PR started as a two-file change and has grown through five rounds, I would rather not decide that trade-off inside a review thread. Options as I see them: (a) accept it and document that fnm-after-teardown resolves the newest installed version; (b) build the in-shell walker; (c) split this PR — the Codex Documents fix is clean and independent and could ship now, leaving the probe work to land on its own. I have put the same question to the repo owner and will follow whichever way it goes.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2439 passed / 7 skipped; `apps/cli` re-running.

yuezengwu (changes requested):
One blocker remains on fc5520f:

The fnm teardown case still changes the selected version. `packages/client/src/runtime/login-shell-path.ts:117-122` appends every version from the shell-only `$FNM_DIR` newest-first because fnm supplies no stable equivalent of `$NVM_BIN`. In the new `recovers a shell-only FNM_DIR after the multishell link is gone` test, active v20 and inactive v22 are both installed, but the assertion only checks that both dirs are present; the returned order is v22 before v20. Once the dead multishell entry misses, every provider resolver therefore selects inactive v22, exactly the active-version regression called out on the previous head.

That is an acknowledged product trade-off, not a completed fix for the existing invariant that an intentional fnm/nvm selection keeps its selected executable and context. Please preserve that invariant or fail closed when the active fnm version cannot be determined; choosing no fallback for an ambiguous multi-version fnm root is safer than silently choosing the newest version and does not require moving the protected-path boundary back into the shell. If the repo owner explicitly accepts the behavior change instead, the PR description and QA contract need to state and cover the selected-version change before this can be approved.

The other findings are resolved: complete executable candidates now cross the protected-path boundary before automatic `stat` / `access`, shell-only `$FNM_DIR` is carried into the vetted stable-root scan, and `$NVM_BIN` preserves nvm's active version after teardown.

Gandy2025:
Your suggestion is better than any of the three options I put up, so I have taken it rather than escalating the question: **fail closed when the active version cannot be determined.** Pushed in eee4dfb.

A version-manager root holding more than one installed version now contributes nothing. Which one the shell selected is not recoverable from disk, and answering with the newest silently swaps the executable a user pinned — you are right that this is worse than a miss, and that it broke the existing invariant rather than merely trading against it. nvm sidesteps the ambiguity entirely through `$NVM_BIN`; fnm exports no equivalent, so a multi-version fnm root fails closed.

This also keeps the protected-path boundary exactly where it is — no in-shell walker — and the change is subtractive: the newest-first ordering and its collator are gone, because ranking versions was only ever needed in order to guess between them.

The precedence test now asserts both halves instead of just presence: with v20 active and v22 installed, a live selection still resolves to v20, and once the selection is gone the result is `null` rather than v22. The ambiguous-root case is covered directly, and the single-version case still proves the fallback works. The shell-only `$FNM_DIR` test keeps proving what it was written for — that a root visible only inside the login shell is reachable at all.

PR description and QA case updated: the description now states the fail-closed behaviour and its cost in place of the old "stable roots" framing, and the QA case treats a resolved path pointing at an unselected version as a failure, and a more serious one than a miss.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2440 passed / 7 skipped; `apps/cli` re-running.

Thanks for pushing on this one — six rounds, and the recurring shape was me looking for a cheaper answer than "resolve it safely, or don't answer". Fail-closed is that principle applied to the last case where I was still guessing.

yuezengwu (changes requested):
One blocker remains on eee4dfb:

The new fail-closed check is scoped per root, not to the complete fallback decision. `packages/client/src/runtime/install-locations.ts:83-123` accepts one candidate from every root that individually contains exactly one version, and `getLoginShellPathDirs` appends all of them in hard-coded nvm/fnm-root order. For example, if the shell reports an active custom `$FNM_DIR` containing v20 and v22, that root correctly contributes nothing — but a stale default `~/.local/share/fnm` or `~/.nvm` root containing one version still contributes a candidate and is silently selected after the dead multishell entry misses. Likewise, with no captured manager metadata, two different roots containing one version each remain globally ambiguous even though each passes the per-root check.

That still guesses a different executable when the active selection is unknowable. Please make the fail-closed decision across the complete manager state: a captured `$FNM_DIR` should restrict fnm fallback to that root; a captured `$NVM_BIN` should be the nvm fallback rather than being followed by unrelated enumerated versions; and when there is no active metadata, return a fallback only when there is exactly one safe candidate in total across all roots. Please cover an ambiguous active `$FNM_DIR` plus a single-version default/nvm root, and multiple single-version roots with no active metadata, proving neither yields an unrelated candidate.

The direction and contract updates are otherwise sound: an ambiguous root no longer chooses its newest version, the selected-version behavior is documented, and QA now treats an unselected resolved version as a failure.

baixiaohang (changes requested):
- Rationale: The fail-closed rule is still applied per version-manager root, so the fallback can select an unrelated installation when the active fnm version is unknowable.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery must avoid macOS Files & Folders prompts before user action without silently changing the provider executable selected by nvm/fnm.
- Approach: Keep the Darwin shell probe non-traversing, vet every automatic executable candidate through a protected-path resolver, carry stable version-manager metadata out of the shell, and fail closed when the selected version cannot be recovered.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. `versionsUnder` rejects ambiguity only within one root, while `versionManagerBinDirs` still returns one candidate from every other root that individually contains one version, and `getLoginShellPathDirs` appends all of them in hard-coded order. If the captured `$FNM_DIR` is active but contains v20 and v22, it contributes nothing as intended, yet a stale single-version `~/.local/share/fnm` or `~/.nvm` root still wins after the multishell link disappears. With no captured metadata, two single-version roots are likewise globally ambiguous but one is still selected. The fail-closed decision must cover the complete fallback state: a captured `$FNM_DIR` should restrict fnm enumeration to that root, a captured `$NVM_BIN` should replace nvm enumeration, and without active metadata the fallback should be returned only when exactly one safe candidate exists in total. Please cover both an ambiguous active `$FNM_DIR` plus an unrelated single-version root and multiple single-version roots with no active metadata. [R5 / `packages/client/src/runtime/install-locations.ts:83`]

✅ 2. The previous per-root newest-version guess is removed, and the documented/QA behavior now correctly treats an unselected executable as worse than a miss.

**Action taken**
- Submitted request changes

Gandy2025:
Correct — rejecting ambiguity *inside* a root does not make the overall answer unambiguous, and I had stopped one level short. Fixed in f81fa54.

Both of your cases were real:

- An active `$FNM_DIR` holding v20 and v22 contributed nothing as intended, but a stale single-version `~/.nvm` or default fnm root still answered. That is a different installation, not a fallback.
- With no captured metadata, two roots holding one version each are exactly as ambiguous as one root holding two — and one of them was still returned.

The decision now covers the complete state:

- **The shell reported something → only that is used.** `$NVM_BIN` names the selected version outright and replaces nvm enumeration entirely. `$FNM_DIR` names only the active root, so that root and no other is consulted, and it answers only when it holds exactly one version. Default roots are not scanned at all: if we know which manager is active and it cannot answer, an unrelated installation is not a better answer.
- **The shell reported nothing → default roots are scanned**, and the result is used only when exactly one safe candidate exists in total.

`getLoginShellPathDirs` no longer composes the fallback itself — it hands the captured metadata over and takes what comes back, so "can we establish the selection" is decided in exactly one place rather than being spread across a per-root filter and a caller-side concat.

Coverage for both cases you named: an ambiguous active `$FNM_DIR` alongside an unrelated single-version root yields nothing, and the same root made unambiguous answers while the stale one still plays no part; two single-version roots with no metadata yield nothing; and `$NVM_BIN` answers even where enumeration would have been ambiguous, proving it is authoritative rather than merely first.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2442 passed / 7 skipped; `apps/cli` re-running.

yuezengwu (changes requested):
One blocker remains on f81fa54:

The captured-metadata branch still has an unhandled global ambiguity. `packages/client/src/runtime/install-locations.ts:112-120` concatenates `$NVM_BIN` and the sole candidate under `$FNM_DIR`, then returns both without applying the global uniqueness rule. A login shell can have both variables set — for example when both manager init scripts ran, or one variable survived a later PATH change. The live PATH order is what selected the active manager; once its per-session entry is gone, the two environment variables do not establish that selection. The fallback nevertheless orders nvm first, so it can silently replace an active fnm executable. If the fnm root is multi-version and contributes nothing, falling through to `$NVM_BIN` is still a guess because the presence of both manager signals is itself ambiguous.

Please fail closed when both captured manager signals are present unless the probe's PATH data can unambiguously identify which manager was active, and cover both forms: `$NVM_BIN` plus an unambiguous single-version `$FNM_DIR`, and `$NVM_BIN` plus an ambiguous multi-version `$FNM_DIR`. Neither should select nvm merely because it is first in the array.

The previous finding is otherwise resolved: a sole captured manager excludes unrelated roots, and without captured metadata the default-root scan returns only a globally unique safe candidate.

baixiaohang (changes requested):
- Rationale: The fallback now resolves ambiguity across roots, but still guesses when the shell reports both nvm and fnm metadata.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery must avoid macOS Files & Folders prompts before user action without silently changing the provider executable selected by a version manager.
- Approach: Keep Darwin shell probing non-traversing, vet complete executable candidates through a protected-path resolver, and use shell-exported version-manager metadata only when it determines one safe fallback.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The captured-metadata branch does not apply its global uniqueness rule when both manager signals are present. A login shell may retain both `$NVM_BIN` and `$FNM_DIR` after both init scripts ran or after PATH precedence changed. Lines 117-123 concatenate the nvm candidate and the sole fnm candidate and return both, so once the fnm multishell entry disappears the caller can select nvm merely because it is first. If the fnm root is multi-version and contributes nothing, falling through to `$NVM_BIN` is still a guess: the two captured signals do not establish which manager owned PATH precedence. Please fail closed when both signals are present unless the probe can identify the active manager unambiguously, and cover both an unambiguous single-version and an ambiguous multi-version `$FNM_DIR` alongside `$NVM_BIN`. [R5 / `packages/client/src/runtime/install-locations.ts:117`]

✅ 2. The previous cross-root cases are resolved: one captured manager excludes unrelated roots, and the no-metadata scan now returns only one globally unique safe candidate.

**Action taken**
- Submitted request changes

Gandy2025:
Correct — the captured-metadata branch skipped the very rule I had just added below it. Fixed in 5f21252.

Two variables say two managers initialised, not which one owned `$PATH` precedence, and that ordering is exactly what is gone by the time the fallback runs. So both signals present now yields nothing at all. The subtle half is the one you called out explicitly: falling through to `$NVM_BIN` when the fnm root happens to be multi-version *looks* like the fnm branch simply declining, but it is the same guess wearing a plausible shape. The single-signal path now applies the uniqueness rule as well, rather than assuming a lone signal can only ever produce one candidate.

Both forms are covered: `$NVM_BIN` with an unambiguous single-version `$FNM_DIR`, and with an ambiguous multi-version one. Neither resolves to nvm.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2443 passed / 7 skipped; `apps/cli` re-running.

For what it is worth, the shape of every finding in this PR has been the same one, and this round is the clearest instance: I keep writing a rule and then leaving a branch that runs before it. The fallback decision is now a single expression with no path around it, which is the structural version of the fix rather than the eighth instance of it.

baixiaohang (changes requested):
- Rationale: The helper fails closed for dual manager metadata, but the caller still exposes `$NVM_BIN` through the raw login-shell PATH and can select it after the fnm link disappears.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery must avoid macOS Files & Folders prompts before user action without silently changing the provider executable selected by a version manager.
- Approach: Keep Darwin shell probing non-traversing, vet complete executable candidates through a protected-path resolver, and fail closed whenever shell metadata cannot establish one version-manager selection.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The dual-signal fail-closed rule is still bypassed end to end. `versionManagerBinDirs` now returns `[]` when both `$FNM_DIR` and `$NVM_BIN` are present, but `getLoginShellPathDirs` has already copied every raw PATH entry into `live` at line 108. For a real ordering such as `[dead fnm multishell, $NVM_BIN]`, the helper contributes no fallback, yet the returned list remains `[dead fnm multishell, $NVM_BIN]`; each provider skips the dead first entry and launches nvm. That is the exact silent substitution the new rule says must yield missing. The added tests call only `versionManagerBinDirs`, so they cannot observe this caller-side bypass. Please enforce the dual-signal decision on the complete returned search list and add an integration test with both env signals, a removed fnm multishell entry, and a runnable binary under `$NVM_BIN`, asserting the provider resolves to `null` rather than nvm. [R5 / `packages/client/src/runtime/login-shell-path.ts:108`]

✅ 2. Within the fallback helper itself, both the single-version and multi-version dual-signal cases now fail closed as intended.

**Action taken**
- Submitted request changes

yuezengwu (changes requested):
One blocker remains on 5f21252:

`versionManagerBinDirs` now fails closed when both `$FNM_DIR` and `$NVM_BIN` are present, but that decision is still bypassed by the caller. `getLoginShellPathDirs` copies every vetted raw PATH entry into `live` before it asks the helper for a fallback (`packages/client/src/runtime/login-shell-path.ts:108-122`). With a captured PATH ordered as `[dead fnm multishell, $NVM_BIN]`, the helper correctly contributes no fallback, yet the returned search list still contains both entries. Each provider skips the dead fnm entry and then launches the nvm executable. That silently substitutes nvm for the fnm selection the live shell had made—the exact outcome the dual-signal fail-closed rule is intended to prevent.

Please apply the ambiguity decision to the complete returned search list, not only the appended stable fallback. Add an integration test that reports both manager variables, removes the fnm multishell entry before `runShell` returns, leaves a runnable provider under `$NVM_BIN`, and asserts provider resolution is missing rather than nvm.

The helper-level change itself is sound: both the single-version and multi-version dual-signal cases now return no fallback. Diff review only; I did not rerun tests or QA.

Gandy2025:
Correct, and I checked it before agreeing because removing entries from the shell's real `$PATH` needs a reason. It holds: with a capture ordered `[dead fnm multishell, $NVM_BIN]` the helper contributes nothing, the returned list still carries `$NVM_BIN`, and every resolver skips the dead entry and launches nvm. Withholding an answer from the appended part while the raw PATH already offers it is bookkeeping, not a decision.

Fixed in 4f5a9da: when the shell reports both managers, entries under `$NVM_BIN` or `$FNM_DIR` are dropped from the live PATH result as well. Everything else on that PATH is untouched — the rule is about the two trees that cannot be distinguished, not about distrusting the capture.

The test now goes through the real composition instead of the helper alone: both variables reported, the fnm per-session entry removed before `runShell` returns, a runnable `codex` under `$NVM_BIN`, asserting resolution is `null`. You were right that the helper-level tests could not observe this — that is the second time in this PR a test proved a rule at the layer that held it rather than at the layer that could bypass it, and it is worth me stating plainly rather than quietly fixing.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2444 passed / 7 skipped; `apps/cli` re-running.

baixiaohang (changes requested):
- Rationale: The end-to-end dual-manager filter closes the Darwin bypass but also deletes the correctly canonicalized, ordered version-manager PATH on non-macOS hosts.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery must avoid macOS Files & Folders prompts before user action without silently changing or losing the executable selected by a version manager.
- Approach: Keep Darwin shell probing non-traversing, vet complete executable candidates through a protected-path resolver, and fail closed when deferred macOS resolution cannot recover one selected version.
- Impacted modules: Client provider executable discovery, CLI external-Context project classification, and runtime QA cases

**Review findings**
❌ 1. The new live-PATH filter is applied on every non-Windows platform, but the ambiguity exists only on Darwin's raw, post-exit resolution path. On Linux, `buildProbeScript` still runs `cd`/`pwd -P` while the login shell is alive, so a PATH ordered `[fnm multishell, $NVM_BIN]` is returned as the stable fnm target followed by nvm; that ordered result already establishes the selected executable. Lines 116-126 now remove both canonical directories merely because both variables are present, turning a correctly discovered provider into `missing` and regressing the non-macOS behavior this PR says is unchanged. The new integration test unintentionally models a Darwin-only raw/dead PATH while leaving `process.platform` on its Linux baseline, so it blesses the regression instead of catching it. Please scope the complete-list fail-closed filter to Darwin, run the dead-multishell integration fixture under `darwin`, and add non-macOS coverage proving a dual-manager, already-canonicalized PATH preserves its first selected version. [R5 / `packages/client/src/runtime/login-shell-path.ts:116`]

✅ 2. For the Darwin raw-PATH case, filtering both manager trees closes the caller-side `$NVM_BIN` bypass identified on the previous head.

**Action taken**
- Submitted request changes

yuezengwu (changes requested):
One blocker remains on 4f5a9da:

The new whole-list ambiguity filter is applied on every non-Windows platform, but the ambiguity it addresses exists only on Darwin's raw, post-shell resolution path. On Linux and other non-macOS hosts, `buildProbeScript` still runs `cd` / `pwd -P` while the login shell is alive, so an ordered PATH such as `[fnm multishell, $NVM_BIN]` is returned as `[stable active-fnm bin, nvm bin]`. That canonicalized ordering already preserves the selected executable. `getLoginShellPathDirs` now removes both directories merely because both environment variables are present (`packages/client/src/runtime/login-shell-path.ts:116-126`), turning a previously valid provider into missing and violating the stated non-macOS compatibility invariant.

The new integration test unintentionally blesses this regression: the file pins `process.platform` to `linux`, and the provider assertion also passes `platform: "linux"`, while its synthetic raw/dead multishell output models only Darwin. Please scope the complete-list filtering to Darwin, run the dead-multishell dual-manager fixture under Darwin, and add non-macOS coverage proving an already-canonicalized dual-manager PATH retains its first selected version.

For the Darwin raw-PATH path, filtering both manager trees does close the caller-side `$NVM_BIN` bypass from the previous head. Diff review only; I did not rerun tests or QA.

Gandy2025:
Correct on both halves, and the second half is the part that stings: the test I added to prove the fix *blessed* the regression instead of catching it. It modelled a Darwin-only raw, already-dead capture while leaving `process.platform` on the file's Linux baseline — so it asserted that a Linux host loses a correctly discovered provider, and called that passing.

Fixed in f9f1945.

The filter is now gated on the same condition that decides whether the shell canonicalizes at all, so the two cannot drift apart. Off macOS `buildProbeScript` still runs `cd` / `pwd -P` while the shell is alive, `[fnm multishell, $NVM_BIN]` comes back as `[stable fnm target, nvm bin]`, and that ordering already establishes the selection — there is no ambiguity to resolve and nothing to withhold.

Coverage now matches: the dead-multishell dual-manager fixture runs under `darwin`, and a new non-macOS case asserts an already-canonicalized dual-manager PATH keeps its first selected version and resolves to the fnm binary.

I also mutation-checked both, since asserting the wrong thing is exactly what went wrong last round: removing the filter fails the Darwin case, and applying it unconditionally fails the non-macOS one. Neither passes vacuously.

Verification: `pnpm check` / `pnpm typecheck` clean; `packages/client` 2445 passed / 7 skipped; `apps/cli` re-running.

yuezengwu (approved):
Re-reviewed f9f1945. The previous blocker is resolved.

The whole-list ambiguity filter is now gated by the same Darwin condition that defers canonicalization until after the probe shell exits. The macOS integration fixture models the dead fnm multishell plus runnable `$NVM_BIN` and verifies that no provider is substituted; the new non-macOS fixture preserves the already-canonicalized `[active fnm, nvm]` ordering and verifies that fnm still wins. The shared protected-path gate, global fail-closed fallback rules, and Codex Documents classifier change remain sound.

No blocking findings. Diff review only; I did not rerun tests or QA.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The protected-path boundary now covers complete automatic executable candidates, version-manager ambiguity fails closed without replacing the selected executable, and the final platform-scoping fix preserves non-macOS discovery behavior.

**Risk level**: A

**PR summary**
- Problem: Automatic provider discovery could trigger unexplained macOS Files & Folders prompts before user action, while an incomplete fix could lose or silently replace the executable selected by nvm/fnm.
- Approach: Keep the Darwin shell probe non-traversing, vet every complete automatic executable candidate through one protected-path boundary, fail closed when version-manager selection is ambiguous, and avoid probing the Codex Documents scratch base for unrelated projects.
- Impacted modules: Client provider discovery, CLI external-Context project classification, and macOS runtime QA.

**Review findings**
✅ 1. The Darwin dead-multishell path now refuses an ambiguous nvm substitution, while non-macOS retains its already-canonicalized PATH order and selected fnm binary.
✅ 2. Protected roots, symlinked ancestors, final executable symlinks, case-insensitive spellings, and wildcard PATH entries are covered at the filesystem boundary.
✅ 3. The Codex classifier avoids Documents canonicalization for ordinary project cwd values while preserving redirected-Documents scratch classification.

**Action taken**
- Approved. Diff review only; I did not rerun tests or QA.

Gandy2025:

gandyxiong01:

PR #2222, opened 2026-08-06
fix(server): stop watchers self-promoting to speaker by sending

Gandy2025:
## Problem

Two human write routes called `ensureParticipant` before writing:

- `POST /chats/:chatId/messages`
- `POST /chats/:chatId/requests/:requestId/ask-agent`

`ensureParticipant` is a membership **write**, not a check — it upserts the caller to `access_mode='speaker'`.

`requireChatAccess` deliberately admits watchers (they need read-cursor and join reachability), and its supervisor branch admits a caller who manages a speaker even with no membership row at all. So **any watcher who sent a message silently promoted themselves to speaker**, bypassing `ensureCanJoin` — the guard the dedicated join route (`POST /:chatId/workspace-join`) applies to exactly that transition.

The web console already models the intended contract: a watcher gets a read-only banner plus a **"Join to reply"** button instead of a composer. The server contradicted its own UI.

This was not a deliberate product behaviour. The call was introduced in #242 (route refactor) when the model was still a single `chat_participants` table with **no watcher concept**, where it was a harmless defensive write. It became a privilege escalation in place when the watcher/`access_mode` model landed. No commit message, comment, design doc, or test ever described "sending joins you"; the only tested watcher→speaker path is `joinMeChat`.

## Change

Both routes now gate with `assertParticipant` (403 for a watcher). Watchers become speakers only through the join route.

`ensureParticipant` itself is unchanged but now has **no production caller**. Its docblock previously justified the call with "the `scope` middleware has already gated the request through `requireChatAccess`" — the precise reasoning that produced the bug. That block now records why it was wrong, so the call is not reintroduced. A stale comment in `require-resource.ts` that named `ensureParticipant` as the way to *refuse* watchers is corrected to `assertParticipant`.

## Tests

New `chat-message-watcher-escalation.test.ts`:

- watcher sends a message → 403, membership still `watcher`, no message persisted
- watcher sends an ask-agent clarification → 403, membership still `watcher`, only the seeded request survives
- positive control: an actual speaker still sends → 201

Verified red before the fix: both watcher cases returned **201** with the message persisted — the escalation reproduced, not inferred.

`agent-send-mention-injection.test.ts` had three web-endpoint cases whose fixture created a chat via the agent API and then posted on the human route without the human ever joining — they relied on the implicit auto-join. The fixture now joins through `workspace-join`, matching the real composer precondition. Assertions are unchanged.

Two test titles/comments that documented the removed "web-UI auto-join path" are reworded; they still pin the membership-write invariants they always did.

## Verification

- `pnpm --filter @first-tree/server test` — 269 files / 3133 tests pass
- `pnpm --filter @first-tree/server typecheck` — clean
- `biome check` on changed files — clean

One full-suite run showed an unrelated failure in `chat-token-usage-performance.test.ts` (a Postgres planner-choice assertion). It did not reproduce on re-run, passes in isolation, and passes in a full-suite run on unmodified `main`. Flagging it as an observed flake rather than silently omitting it.

## Deliberately out of scope

While checking completeness I found three adjacent human write routes gated only by `requireChatAccess`, so a watcher can also reach them:

- `PATCH /:chatId` — chat topic/description
- `POST /:chatId/participants` — add participants
- the SCM entity follow/unfollow routes

These are a different defect class (no membership write involved) and belong with the shared removal/participant authorizer being built next. Not touched here to keep this diff reviewable.

### DIFF
--- a/packages/server/src/__tests__/agent-send-mention-injection.test.ts
+++ b/packages/server/src/__tests__/agent-send-mention-injection.test.ts
@@ -544,7 +544,21 @@ describe("mention enforcement + content normalisation", () => {
         participantIds: [peerA.agent.uuid, peerB.agent.uuid],
       });
       expect(chatRes.statusCode).toBe(201);
-      return { sender, peerA, peerB, chatId: chatRes.json().id as string };
+      const chatId = chatRes.json().id as string;
+
+      // The chat is created by the non-human agent, so its manager's human
+      // agent starts as a watcher. The web message route is speaker-gated,
+      // so the human joins through the same route the console's "Join to
+      // reply" button calls before posting — matching the real composer
+      // precondition instead of relying on an implicit auto-join.
+      const joinRes = await app.inject({
+        method: "POST",
+        url: `/api/v1/chats/${encodeURIComponent(chatId)}/workspace-join`,
+        headers: { authorization: `Bearer ${sender.accessToken}` },
+      });
+      expect(joinRes.statusCode).toBe(204);
+
+      return { sender, peerA, peerB, chatId };
     }
 
     it("rejects bypass attempts (no mentions declared) with 400", async () => {
--- a/packages/server/src/__tests__/chat-message-watcher-escalation.test.ts
+++ b/packages/server/src/__tests__/chat-message-watcher-escalation.test.ts
@@ -1,0 +1,148 @@
+/**
+ * HTTP-level regression coverage for the watcher → speaker self-promotion
+ * hole on the two human write routes:
+ *
+ *   - `POST /api/v1/chats/:chatId/messages`
+ *   - `POST /api/v1/chats/:chatId/requests/:requestId/ask-agent`
+ *
+ * Both routes used to call `ensureParticipant`, which is a membership
+ * *write* (it upserts the caller to `access_mode='speaker'`), not a check.
+ * A watcher therefore promoted themselves to speaker simply by sending —
+ * bypassing `ensureCanJoin`, the guard the dedicated join route
+ * (`POST /:chatId/workspace-join`) applies to exactly that transition.
+ *
+ * The web console already models the correct contract: a watcher gets a
+ * read-only banner plus an explicit "Join to reply" button instead of a
+ * composer. These tests pin the server side to that same contract.
+ */
+
+import { and, eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import { chatMembership } from "../db/schema/chat-membership.js";
+import { messages } from "../db/schema/messages.js";
+import { createAgent } from "../services/agent.js";
+import { createMeChat } from "../services/me-chat.js";
+import { createTestAdmin, createTestAgent, useTestApp } from "./helpers.js";
+
+describe("watcher self-promotion on human write routes", () => {
+  const getApp = useTestApp();
+
+  /**
+   * Build a chat owned by `peer` containing a non-human agent managed by
+   * `manager`. `recomputeChatWatchers` materialises a watcher row for the
+   * manager's human agent — the exact shape the hole was reachable from.
+   */
+  async function seedWatcherChat(app: ReturnType<typeof getApp>) {
+    const manager = await createTestAdmin(app);
+    const peer = await createTestAdmin(app);
+    const managed = await createAgent(app.db, {
+      name: `esc-managed-${crypto.randomUUID().slice(0, 6)}`,
+      type: "agent",
+      displayName: "Escalation Managed",
+      managerId: manager.memberId,
+      organizationId: manager.organizationId,
+    });
+    const { chatId } = await createMeChat(app.db, peer.humanAgentUuid, peer.organizationId, {
+      participantIds: [managed.uuid],
+    });
+    return { manager, peer, managed, chatId };
+  }
+
+  /** Insert a `format='request'` message straight from the managed agent. */
+  async function seedOpenRequest(
+    app: ReturnType<typeof getApp>,
+    chatId: string,
+    senderId: string,
+    targetHumanAgentId: string,
+  ) {
+    const messageId = crypto.randomUUID();
+    await app.db.insert(messages).values({
+      id: messageId,
+      chatId,
+      senderId,
+      format: "request",
+      content: "need a decision",
+      metadata: { mentions: [targetHumanAgentId] },
+      source: "agent",
+    });
+    return { messageId };
+  }
+
+  async function loadMembership(app: ReturnType<typeof getApp>, chatId: string, agentId: string) {
+    const [row] = await app.db
+      .select({ accessMode: chatMembership.accessMode })
+      .from(chatMembership)
+      .where(and(eq(chatMembership.chatId, chatId), eq(chatMembership.agentId, agentId)))
+      .limit(1);
+    return row ?? null;
+  }
+
+  it("refuses a watcher's message with 403 and leaves them a watcher", async () => {
+    const app = getApp();
+    const { manager, managed, chatId } = await seedWatcherChat(app);
+
+    // Precondition: the supervisor is a watcher, not a speaker.
+    expect((await loadMembership(app, chatId, manager.humanAgentUuid))?.accessMode).toBe("watcher");
+
+    // Carry an explicit recipient so the send would otherwise succeed —
+    // membership is then the only thing standing between this request and a
+    // persisted message, which is exactly what the guard must decide.
+    const res = await app.inject({
+      method: "POST",
+      url: `/api/v1/chats/${encodeURIComponent(chatId)}/messages`,
+      headers: { authorization: `Bearer ${manager.accessToken}` },
+      payload: { content: "promoting myself by typing", metadata: { mentions: [managed.uuid] } },
+    });
+
+    expect(res.statusCode).toBe(403);
+
+    // The membership row must be untouched — this is the actual regression.
+    expect((await loadMembership(app, chatId, manager.humanAgentUuid))?.accessMode).toBe("watcher");
+
+    // And no message may have been persisted.
+    const rows = await app.db.select({ id: messages.id }).from(messages).where(eq(messages.chatId, chatId));
+    expect(rows).toHaveLength(0);
+  });
+
+  it("still lets an actual speaker send (guard does not over-refuse)", async () => {
+    const app = getApp();
+    const owner = await createTestAdmin(app);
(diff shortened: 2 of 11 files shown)


yuezengwu (approved):
Reviewed the full diff. Both human write routes now use the speaker-only `assertParticipant` guard instead of the membership-mutating `ensureParticipant`, preserving `workspace-join` as the sole watcher-to-speaker transition. The adjusted fixture follows that explicit join contract, and the new regression coverage checks the 403 response, unchanged watcher membership, no persisted message, and the speaker positive control.

This matches the durable chat contract that watchers are observers rather than senders or delivery targets. No schema or database change is involved. I did not rerun tests or QA as part of this review.

baixiaohang (changes requested):
- Rationale: The route-level guard change is correct, but the PR leaves the now-unused, authorization-free watcher-to-speaker writer exported beside the canonical join path, preserving the same future misuse trap that caused this escalation.

**Risk level**: B-high
- Path baseline: `packages/server/src/services/chat.ts` matches the protected chat-service path.
- Semantic lift: none; the path baseline is already B-high.

**PR summary**
- Problem: A human who could only observe a chat could send or ask for clarification and silently become a speaker, bypassing the explicit "Join to reply" authorization flow.
- Approach: Replace the two human write routes' membership-mutating precondition with a speaker-only assertion, preserve `workspace-join` as the explicit watcher-to-speaker transition, and update affected fixtures and regression coverage.
- Impacted modules: server chat HTTP routes, chat authorization/membership services, and server regression tests.

**Review findings**
❌ 1. Remove `ensureParticipant` now that this PR removes its last production caller, and migrate its direct tests/comments to the canonical membership/join seam. Keeping an exported helper that performs an authorization-free watcher→speaker write leaves two transition mechanisms beside `joinAsParticipant`, so a future handler can repeat the exact bug this PR fixes. The new docblock cannot be the safety boundary: it says the helper has no production caller, but its remaining behavior section still says every inbound IM message hits it and instructs maintainers to append callers to a list that no longer exists. If there is a current authorized production use that requires this distinct primitive, identify it and make that contract internally consistent instead. [R1/R2 / `packages/server/src/services/chat.ts:1110`]

**Action taken**
- Submitted request changes.

Gandy2025:
Thanks — accepted, and the internal inconsistency you caught was real.

**Finding 1 — delete `ensureParticipant`: done** (`cbf6ced`).

You were right that a docblock cannot be the safety boundary, and the one I wrote contradicted itself: it claimed the helper had no production caller while the Behaviour section still said "the hot path for the IM bridge (every inbound message hits this)" and told maintainers to "append it to the list above" — a list I had just deleted. That is a defect in the change, not only a style preference.

I originally kept the helper to hold the diff down. Weighed against leaving two watcher→speaker mechanisms beside `joinAsParticipant`, that was the wrong trade — the whole point of this branch is that the second mechanism is how the escalation happened.

What changed:

- `ensureParticipant` is gone from `services/chat.ts`.
- Its direct test usages now call `applyMembershipWrite(db, chatId, [{ agentId }], { upgradeWatcherToSpeaker: true })` — the real production seam those tests were always exercising through the wrapper (`chat-upgrade`, `participant-mode-invariant`, `participant-mode-backfill`, `cron-jobs.integration`), plus removal from the `service-branch-defaults` export sweep.
- Stale references in `participant-mode.ts`, `require-resource.ts`, and the `api/chats.ts` comment no longer name a symbol that does not exist. `require-resource.ts` now warns against membership writers generally rather than one function name.

**One test dropped, deliberately.** `"is idempotent — ensureParticipant for an existing participant does not re-flip modes"` pinned the wrapper's short-circuit, not production behaviour: it asserted a repeat call left `chat_membership.mode` untouched. `applyMembershipWrite` rewrites `mode` via `ON CONFLICT DO UPDATE`, and `mode` is documented v2 dead code that no fan-out, enforcement, or dispatcher path reads. The property worth keeping — backfill idempotence — is still covered by `"already-speaker no-op insert does not double-backfill"` and the repeat-call assertion in the backfill suite. Flagging it explicitly rather than letting a deleted test pass unnoticed.

Verification after the change: server suite 269 files / 3132 tests pass (3133 minus that one case), `typecheck` clean, `biome check` clean on all touched files.

yuezengwu (approved):
Re-reviewed the updated head. The blocking concern is addressed: `ensureParticipant` has been removed entirely, its stale service/scope references are gone, and the direct tests now exercise the explicitly named membership-write bundle instead of preserving a second access-check-shaped watcher-to-speaker helper beside the canonical join flow.

Dropping the wrapper-only mode short-circuit test is appropriate because the wrapper no longer exists and `chat_membership.mode` is decision-inert; the meaningful backfill idempotence remains covered. The original route guards and watcher regression coverage remain intact. No schema or database change is involved. I did not rerun tests or QA as part of this review.

Gandy2025:
Correction to the "Deliberately out of scope" section above — I got two of those three wrong.

I derived that list from route-level reading only (`requireChatAccess` and nothing else in the handler) without following each into its service. Having now checked the service layer while building the removal authorizer:

1. **`POST /:chatId/participants` — not a gap. My claim was wrong.** It calls `addMeChatParticipants` → `inviteParticipantsToChat`, whose step 2 requires the caller to be a **speaker** and raises `CallerNotSpeakerError` (remapped to 404 by the web wrapper to avoid leaking chat existence). A watcher cannot add participants.

2. **SCM entity follow/unfollow — not an unguarded hole either.** The handler calls `resolveHumanScmBindingPair` first and rejects when the caller has no active delegate **speaker** in the chat ("Following needs delegate_mention to identify an active delegate speaker in this chat"). Whether that is the right rule is a design question, not a missing check.

3. **`PATCH /:chatId` — this one stands.** `updateChatMetadata` performs no authorization at all; it is a pure writer. The route is gated only by `requireChatAccess`, which admits watchers, so a watcher can rename a chat's topic/description. Note `assertOwner`'s docblock already states the intent — "the human/web route stays participation-gated" — so the guard is missing relative to its own documented contract, not merely absent.

None of this changes the diff under review. I am flagging it because the section as written overstates the exposure, and a reader could act on it.

The remaining real item (3) is not part of the removal authorizer either — it is the same class as this PR (a write route that admits watchers), not a participant-roster rule. I will carry it as its own follow-up rather than smuggling it into the removal change, and will say so accurately in that PR.

baixiaohang:
The participant-add and **follow** corrections are right, but the statement still over-corrects **unfollow**:

- GitHub `DELETE /:chatId/github-entities` calls `requireChatAccess` and then `removeEntityFollow`; unlike the POST route, it never resolves a speaker binding pair. A watcher therefore reaches the deletion.
- GitLab `DELETE /:chatId/gitlab-entities` calls `requireDirectHumanChatMembership`, but that helper checks only that a direct membership row exists and does not require `access_mode='speaker'`; it then calls the deletion service without another membership guard. A watcher row satisfies this check too.

So the accurate adjacent scope is: participant add is speaker-gated; SCM follow is speaker/delegate-gated; chat metadata PATCH and both human SCM unfollow paths remain watcher-accessible. This still does not change the diff under review or my current approve recommendation, but the follow and unfollow conclusions should not be collapsed.

Gandy2025:
Verified independently — you're right, and I collapsed follow and unfollow into one conclusion when only the POST paths are gated.

Checked both deletion services this time rather than stopping at the handler, since that shortcut is what produced the first error:

- **GitHub `DELETE /:chatId/github-entities`** — `requireChatAccess` → `removeEntityFollow`, which resolves the chat's org and deletes mapping rows scoped by `chatId`. No binding-pair resolution, no membership check anywhere in the service. A watcher reaches the deletion.
- **GitLab `DELETE /:chatId/gitlab-entities`** — `requireChatAccess` → `requireDirectHumanChatMembership` → `removeCurrentGitlabEntityFollow` / `removeGitlabEntityFollow`. The helper selects from `chat_membership` on `(chat_id, agent_id)` with **no `access_mode` predicate**, so a watcher row satisfies it, and neither deletion service re-checks. Note the GitLab POST calls the same helper *and* `resolveHumanScmBindingPair` — the delegate gate is what separates them, not the helper.

Accurate adjacent scope, superseding my previous comment:

| Route | Status |
|---|---|
| `POST /:chatId/participants` | speaker-gated (`CallerNotSpeakerError`) |
| SCM follow — GitHub POST, GitLab POST | speaker/delegate-gated |
| **SCM unfollow — GitHub DELETE, GitLab DELETE** | **watcher-accessible** |
| **`PATCH /:chatId`** | **watcher-accessible** |

Both remaining items are the same class as this PR — a write route that admits watchers — and neither belongs in the removal authorizer. I'll carry them as their own change and have corrected the description on #2223, which repeated the collapsed claim.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The updated head fixes the watcher self-promotion paths and removes the redundant authorization-free membership writer, so no blocking concern remains.

**Risk level**: B-high
- Path baseline: `packages/server/src/services/chat.ts` matches the protected chat-service path.
- Semantic lift: none; the path baseline is already B-high.

**PR summary**
- Problem: A watcher could send an ordinary message or Ask agent clarification and silently become a speaker, bypassing the explicit "Join to reply" authorization flow.
- Approach: Gate both human write routes with the speaker-only `assertParticipant`, keep `workspace-join` as the watcher-to-speaker transition, delete the now-unused `ensureParticipant` writer, and move invariant tests to the canonical membership-write bundle.
- Impacted modules: server chat HTTP routes, chat authorization/membership services, and server regression tests.

**Review findings**
✅ 1. Both affected routes now reject watchers without mutating membership or persisting a message, while preserving the speaker path.
✅ 2. The prior blocking issue is resolved: `ensureParticipant` and its stale references are removed, so there is no second access-check-shaped promotion helper beside the canonical join flow.
✅ 3. The removed wrapper-only test covered decision-inert `mode` behavior; the meaningful membership backfill idempotence coverage remains.

**Action taken**
- Approved the current head after explicit maintainer authorization. Static review only; I did not rerun tests or QA.

gandyxiong01:

Gandy2025:
Closing at the maintainer's request.

For the record, so the state is not misread later: this fix is not carried by #2220 — that PR still imports and calls `ensureParticipant` on the human message route. Closing this leaves the watcher self-promotion path in place: a watcher who sends a message, or asks for clarification on an open request, is silently upserted to `access_mode='speaker'`, bypassing `ensureCanJoin` on the dedicated join route and contradicting the web console's own read-only + "Join to reply" contract.

The branch `fix/watcher-message-self-promotion` remains on the remote; reopening restores an approved, all-required-checks-green state.

PR #2224, opened 2026-08-07
Keep the Context Tree impact note out of blocking questions

Gandy2025:
## Why

The Context Tree impact note explains a finished result. A blocking question is not a result — it asks the reader to choose, and its body is required to be decision-self-sufficient. An attribution footnote there competes with the decision instead of serving it.

## What

- The note rides only the final `chat send` in managed First Tree Chat, and the native final response in BYO. The chat message flow is unchanged.
- A task that correctly ends in a blocking `chat ask` carries no note. Any Tree constraint bearing on that decision belongs in the question's own prose, where the reader actually needs it while choosing.
- The grader now enforces the *surface*, not only the note's shape: visible output is tracked per delivery surface, and a note found in a `chat ask` body fails the behavior gate even when `ask` is the correct final transport for that task.
- Skill payload `0.6.2` → `0.7.0`.

`managedTransport` stays a per-case contract — a task whose next step genuinely depends on a human decision still correctly ends in an ask; it just carries no note.

## Verification

- focused `first-tree-read` suite — 64 tests passed
- `python3 scripts/quick_validate_skill.py skills/first-tree-read` — valid
- `pnpm --filter @first-tree/skill-evals typecheck`, changed-file `biome check`, `git diff --check`
- direct grader probe: an identical note passes on `chat send` (`impactNoteBehaviorOk=true`) and fails on `chat ask` (`false`, `impactNoteOutsideBlockingAsk=false`); a clean ask with no note still passes

Model-backed gate not run — the repo requires explicit human authorization for live model evals.

## Rollout

Skill payload only; no server, schema, Web, or adapter change. Agents pick it up with the next shipped payload.


Gandy2025:
Context Tree counterpart (draft until this merges): agent-team-foundation/first-tree-context#906

baixiaohang (changes requested):
- Rationale: The behavioral contract is coherent, but the grader now emits the opposite transport guidance when a blocking-ask case uses the wrong final transport.

**Risk level**: B-low
- Path baseline: `skills/**` and `packages/skill-evals/**` -> C
- Semantic lift: the skill payload/version changes shipped agent behavior -> B-low

**PR summary**
- Problem: Agents currently append a Context Tree attribution footer to blocking questions, competing with the human decision the question is meant to collect.
- Approach: Keep the note on terminal managed `chat send` output and BYO final responses, prohibit it on `chat ask`, and make the deterministic grader track the delivery surface. The draft Context Tree counterpart, agent-team-foundation/first-tree-context#906, matches this boundary.
- Impacted modules: `skills/first-tree-read`, `packages/skill-evals/src/suites/first-tree-read`

**Review findings**
❌ 1. `driftNote` now hard-codes that a failed managed transport means the note was not delivered through a required final `chat send`. For a case whose preserved per-case contract is `managedTransport: "ask"`, an erroneous `send` therefore produces guidance saying that `send` was required—the exact opposite of the expected fix. Make this diagnostic transport-aware (or surface-neutral) so ask cases remain actionable. [R4 / `packages/skill-evals/src/suites/first-tree-read/summary.ts:106`]
⚠️ 2. The new `impactNoteOutsideBlockingAsk` metric is returned and serialized in JSON, but it is omitted from the drift text and Markdown result fields. A failure caused solely by note placement will therefore show every displayed shape/authority check as passing without exposing the new failing surface predicate. Please include it in the human-readable diagnostics. [R4 / `packages/skill-evals/src/suites/first-tree-read/summary.ts:116`]
✅ 3. The skill wording, surface-index alignment, and paired Tree draft consistently distinguish a blocking question from a terminal conflict assessment.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
The behavior change is coherent: the skill keeps attribution on terminal results, removes it from blocking questions, and the grader now tracks delivery surfaces. The paired draft Context Tree update is aligned. I found two blocking diagnostic regressions before this is ready:

1. `summary.ts:106-108` now always says the required managed transport was `chat send` whenever `managedFinalTransportOk` is false. `managedTransport` remains a per-case contract, so an ask-contract case that incorrectly sends will be told that sending was required—the opposite recovery guidance. Please make this message transport-aware, or make it surface-neutral.

2. `impactNoteOutsideBlockingAsk` is the new predicate that explains a surface-only failure, but `summary.ts:116-119` and the Markdown result fields omit it. A note on `chat ask` can therefore fail `impactNoteBehaviorOk` while every displayed shape/authority field is true. Please expose the predicate in the drift evidence and Markdown summary so the deterministic grader remains actionable.

Per review instructions, I inspected the diff and did not run tests or QA.

Gandy2025:
@baixiaohang @yuezengwu — both accepted, both were mine, fixed in `a2da8f1d4`.

**1. Transport guidance was backwards for ask cases.** I narrowed the old "send or ask" wording to a fixed "chat send" when I removed the ask branch, which made the message actively misleading for an ask-contract case: it would tell an agent that sent-instead-of-asked that sending was required. `managedTransport` is unchanged as a per-case contract, so metrics now carry both `managedFinalTransportKind` (observed) and `managedTransportExpected` (required), and the note names both: *"The task's final managed delivery was send, but this case requires chat ask."*

**2. The surface predicate was invisible.** `impactNoteOutsideBlockingAsk` decided the failure but appeared in neither the drift text nor the Markdown fields — exactly the case you describe, where a note in a `chat ask` fails while every shape and authority field displays true. It is now in the drift evidence, the structured behavior evidence line, and the Markdown result fields (@yuezengwu — thanks for naming the Markdown surface specifically; I had only fixed the drift text).

Coverage pins both rather than just the code: an ask-contract case that finishes with a send asserts `requires chat ask` and *not* `requires chat send`, and a note delivered inside an ask asserts `outside blocking ask=false` reaches the drift output. My first attempt at that test was wrong — I asserted the predicate on the wrong-send scenario, where the note is correctly on a send and the predicate is true — so it now uses a separate note-in-ask fixture.

Focused suite 65 passed; skill validator, skill-evals typecheck, Biome, `git diff --check` clean.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up makes both diagnostics actionable and adds focused regression assertions without changing the intended transport contract.

**Risk level**: B-low
- Path baseline: `skills/**` and `packages/skill-evals/**` -> C
- Semantic lift: the shipped skill payload/version changes agent behavior -> B-low

**PR summary**
- Problem: A Context Tree attribution footer on a blocking question competes with the decision the human needs to make.
- Approach: Keep attribution on terminal managed `chat send` output and BYO final responses, prohibit it on `chat ask`, and grade both transport and note placement by delivery surface.
- Impacted modules: `skills/first-tree-read`, `packages/skill-evals/src/suites/first-tree-read`

**Review findings**
✅ 1. Resolved: managed-transport drift now reports both the observed and required per-case transport, so ask-contract failures point to `chat ask` rather than `chat send`.
✅ 2. Resolved: `impactNoteOutsideBlockingAsk` is visible in drift evidence, structured grading evidence, and Markdown summaries.
✅ 3. Focused assertions cover both the wrong-transport and note-inside-ask failure modes; the paired Context Tree draft remains aligned.

**Action taken**
- Approved at head `a2da8f1d4f1f4d7e2312264a85cb67566f4da12c`.

yuezengwu (approved):
Re-reviewed `a2da8f1d4`; the two requested diagnostic fixes are complete.

- Managed transport failures now preserve and report both the observed and required per-case transport, so an ask-contract case no longer receives backwards `chat send` guidance.
- `impactNoteOutsideBlockingAsk` is now visible in drift evidence, structured grading evidence, and the Markdown result, with separate regression coverage for wrong transport and note-inside-ask placement.

The skill wording, grader surface alignment, version bump, and draft Context Tree counterpart remain consistent. No core data structure or database changes are involved.

Approved. Per review instructions, I inspected the updated diff and did not run tests or QA.

Gandy2025:
Pushed `9c0f2d3cb` after both approvals — a real rendering defect surfaced in a live screenshot, so I did not merge on the approved head.

The Chinese template put the full-width colon **inside** the bold span:

```
> **发现约束冲突：**固定发布日期…
```

CommonMark cannot close a `**` run whose closing delimiter is preceded by punctuation and followed by a CJK character, so that line shipped as literal asterisks with no bold. The source line survived only because `[` is itself punctuation. Confirmed against the product's own `Markdown` component:

```
**发现约束冲突：**固定…   literal ** = true,  <strong> = 0
**发现约束冲突**：固定…   literal ** = false, <strong> = 1
```

The colon now sits outside the bold on both Chinese lines, and the skill states why so it is not "corrected" back. English is unaffected — its closing delimiter is followed by a space, which closes normally.

Verification: every `markdown` example in the shipped skill rendered through the real component — both blocks give one blockquote, three `<strong>`, two `<br>`, zero literal asterisks. Floor test also pins the broken form out. Focused suite 65 passed; validator, typecheck, Biome, `git diff --check` clean.

Re-review appreciated since this touches the shipped template text.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: Moving the Chinese full-width colon outside the strong span fixes the demonstrated CommonMark delimiter failure, and the parser plus floor coverage now enforce the renderable form consistently.

**Risk level**: B-low
- Path baseline: `skills/**` and `packages/skill-evals/**` -> C
- Semantic lift: the shipped skill template/version affects agent-visible output -> B-low

**PR summary**
- Problem: The Chinese impact-note effect line could render literal `**` instead of bold text when a CJK character immediately followed the colon-inside strong span.
- Approach: Put the full-width colon immediately after the Chinese strong span on both effect and source lines, align deterministic parsing with that form, and pin the invalid colon-inside form out. English formatting and the no-note-on-blocking-ask contract remain unchanged.
- Impacted modules: `skills/first-tree-read`, `packages/skill-evals/src/suites/first-tree-read`

**Review findings**
✅ 1. The new Chinese form gives the closing `**` a punctuation follower, allowing the strong delimiter to close before the full-width colon and CJK sentence.
✅ 2. Effect parsing, source-line reconstruction, examples, and rejection fixtures all use the same colon-outside invariant.
✅ 3. The paired Context Tree draft records the delivery-surface decision rather than this presentation detail, so no counterpart drift is introduced.

**Action taken**
- Approved at head `9c0f2d3cbd9af503316f7c7949a0917e965c3244`.

yuezengwu (approved):
Re-reviewed `9c0f2d3cb`; the Chinese CommonMark rendering fix is consistent and scoped.

- Both Chinese effect and source lines now place the full-width colon outside the strong span, so the closing `**` is no longer trapped between punctuation and a following CJK character.
- The deterministic parser, reconstructed source line, examples, rejection fixtures, and floor contract all enforce the same renderable form.
- English formatting and the no-note-on-blocking-ask transport contract are unchanged. The paired Context Tree draft appropriately records the durable delivery-surface decision rather than this presentation detail.

No core data structure or database changes are involved. Approved; per review instructions, I inspected the updated diff and did not run tests or QA.

PR #2225, opened 2026-08-07
Bold only the effect label in the Context Tree impact note

Gandy2025:
## Why

All three lines of the note were bold, so nothing was emphasised. Two of the three bold spans sat on fixed scaffolding — wording identical in every note — while the effect label, the one part that changes and answers what happened, carried the same weight as the boilerplate. The source line's information is its link, which already has its own affordance and was competing with a bold label.

## What

Bold marks the effect label alone; the first and third lines are plain text. Copy is unchanged.

```
Context Tree 如何影响本次工作
**收窄可选范围**：指标合同排除了过期快照与无身份桥接的访客估算。
Context Tree 来源：产品分析指标合同 · 约束
```

The deterministic parser follows: first-line and source-label scaffolding must be plain, so the superseded bolded form is rejected rather than silently accepted.

Skill payload `0.7.0` → `0.8.0`.

## Verification

- focused `first-tree-read` suite — 66 tests passed
- rendered both shipped templates through the product's own `Markdown` component: exactly one `<strong>` each (`Options narrowed:` / `发现约束冲突`), two hard breaks, zero literal asterisks
- new regression asserts a bolded first line and a bolded source label both fail the behavior gate
- `quick_validate_skill.py`, skill-evals typecheck, Biome, `git diff --check` clean

Model-backed gate not run — repo requires explicit human authorization.

## Rollout

Skill payload only. No server, schema, Web, or adapter change.


yuezengwu (approved):
Reviewed the full diff. The skill template, deterministic parser, floor contract, regression coverage, and version metadata all move together to enforce a single emphasized span on the effect label. The plain first/source scaffolding and English/Chinese colon and spacing rules remain aligned. No blocking issues found.

PR #2219, opened 2026-08-06
refactor(client): S1d add provider contracts entry

yuezengwu:
## Summary
- Add `packages/client/src/runtime/contracts.ts` as the explicit provider-facing contracts allowlist (future `./contracts` subpath source; not a new package or temporary `@first-tree/client` public subpath).
- Retarget production `handlers/**` and `providers/{builtin-registry,auth-driver}` so handler/delivery, `LoginOutcome`, and Kimi `ReplayFenceWriter` contracts come only from `runtime/contracts.js`.
- Strengthen architecture guards + contracts-entry tests for allowlist, value identity (`noopDeliveryToken` / `requireDeliveryToken`), and no SessionManager/SessionRegistry / public API expansion.

## Integration status
- Ordinary merge of latest `main` into this branch — **no rebase / no force-push**.
- Includes [#2230](https://github.com/agent-team-foundation/first-tree/pull/2230) after prior [#2226](https://github.com/agent-team-foundation/first-tree/pull/2226).
- **Current exact head:** `600709f1920de1516235b726569965734f83fdcc`
- **Base / main parent tip:** `42b35acc98cfe582c0341109724da4b01d8a5651`
- Merge parents: `ead2aa82da221ee511e2b9a1c225e6ffae33a3a6` + `42b35acc98cfe582c0341109724da4b01d8a5651`
- Prior QA / APPROVE on `ead2aa82…` (and earlier) were invalidated by merge-from-main. GitHub may still surface a migrated approval commit field — that is **not** a fresh formal re-review of this head. Gates below are for **`600709f…` only**.

## Test plan
- [x] Author recheck after merge-from-main: Client focused 49/49; Client typecheck; `git diff --check`; effective diff vs main still 18-file S1d contracts seam (+423/-35)
- [x] Low-cost #2230 smoke on new head: agent-switcher + agent-detail-page DOM suites 49/49 (merged-main integration, not new S1d scope)
- [x] GitHub checks on exact head `600709f…` — CI run `31156363490`: `Test Client & Web`, `Test CLI`, fan-in `Test`, and other required jobs SUCCESS (rollup 15 SUCCESS / 2 SKIPPED / 0 pending or failed)
[...]

PR #2226, opened 2026-08-07
fix(web): stabilize Markdown default renderer identity

yuezengwu:
## Summary
- Stabilize Markdown default `a` / `table` renderer **component identity** at module scope so `ReactMarkdown` does not remount an unchanged link when a parent (ComposeStatusBar) re-renders.
- Harden the compose-status focus regression by draining focus-restoration RAF after the status update, so the remount→disclosure race fails closed without this fix (matches the CI Linux/full-suite failure mode).
- Add a direct Markdown DOM-identity + focus characterization, plus a caller-override coverage check.

This is the **diff-external** blocker for S1d [#2219](https://github.com/agent-team-foundation/first-tree/pull/2219) (zero `packages/web/**` in that PR). Observed on Actions [run 31100366695](https://github.com/agent-team-foundation/first-tree/actions/runs/31100366695) at `chat-components-extra.test.tsx:1115` (`keeps focus on a surviving Markdown link…`).

## Root cause
ComposeStatusBar keeps `focusWithinRef` and, when the previously focused control is no longer in the surface, schedules `requestAnimationFrame(() => triggerRef.focus())`. Inline `components.a` / `components.table` factories inside `Markdown` created a new element type every render, so an unchanged narration link remounted on status refresh; focus left the surface and the RAF fallback stole it onto the disclosure.

`react-markdown` 10.1.0 passes `options.components` through to hast-util-to-jsx-runtime as JSX element types, so a new function identity each render remounts the anchor.

## Behavior preserved
- Safe-href downgrade for non-navigable schemes
- External links still `target="_blank"` + `rel="noopener noreferrer"`
- Table overflow wrapper unchanged
- Caller-supplied `components` still override defaults
- Disclosure focus fallback when a jump truly disappears unchanged

## Test plan
- [x] Pre-fix: with RAF drain only, target compose-status test **failed** (focus → disclosure)
- [x] Post-fix: target test **passed** 5/5 isolated repeats (author)
- [x] `chat-components-extra.test.tsx` full file 26/26
- [x] New `markdown-renderer-identity` + existing Markdown suites
- [x] `@first-tree/web` full: **2248 passed**
- [x] `@first-tree/web` typecheck (+ design-token lint)
- [x] Root `pnpm check`, `git diff --check`
- [x] GitHub Actions `Test Client & Web` + fan-in `Test` on frozen head `f849bfb191de529aa36d2f69a1649ed701c63840` (Ubuntu; compose-status 26/26, identity 2/2, Web 2248/2248)
- [x] Independent exact-head QA **PASS** at `f849bfb191de529aa36d2f69a1649ed701c63840`: RAF-drained compose-status target 10/10; renderer identity/override 20/20 across 10 runs; full Web 2248/2248; typecheck / `pnpm check` / `git diff --check` green; base-checkout causality reproduced the CI failure shape

## Scope / non-goals
Independent Web stability PR only. Does **not** modify #2219 / S1d. No assertion weakening, no sleep/retry/timeout expansion, no ComposeStatusBar focus policy change beyond what stable Markdown identity prevents.

### DIFF
--- a/packages/web/src/components/__tests__/markdown-renderer-identity.test.tsx
+++ b/packages/web/src/components/__tests__/markdown-renderer-identity.test.tsx
@@ -1,0 +1,91 @@
+// @vitest-environment happy-dom
+
+import { act, type ReactElement, useState } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { Markdown } from "../ui/markdown.js";
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+
+let container: HTMLElement;
+let root: Root;
+
+beforeEach(() => {
+  container = document.createElement("div");
+  document.body.appendChild(container);
+  root = createRoot(container);
+});
+
+afterEach(() => {
+  act(() => root.unmount());
+  container.remove();
+});
+
+/**
+ * Parent re-renders of Markdown must keep the same DOM anchor node when the
+ * markdown source is unchanged. Inline default `components.a` factories used
+ * to remount the link and drop focus into surrounding surfaces.
+ */
+describe("Markdown default renderer identity", () => {
+  it("keeps DOM identity and focus on an unchanged link across parent re-renders", async () => {
+    let bump: (() => void) | undefined;
+
+    function Harness() {
+      const [, setTick] = useState(0);
+      bump = () => setTick((value) => value + 1);
+      return <Markdown>See [notes](https://example.com/notes) for detail.</Markdown>;
+    }
+
+    act(() => {
+      root.render((<Harness />) as ReactElement);
+    });
+
+    const link = container.querySelector<HTMLAnchorElement>('a[href="https://example.com/notes"]');
+    if (!link) throw new Error("Expected Markdown link");
+    expect(link.getAttribute("target")).toBe("_blank");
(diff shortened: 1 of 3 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: Module-scoped default renderers fix the reconciliation identity issue without changing URL safety, table rendering, or caller override behavior.

**Risk level**: C
- Semantic lift: none; this is localized presentation and focus behavior, with no subscription, WebSocket-frame, or data-layer change.

**PR summary**
- Problem: In the expanded compose-status surface, a status refresh could remount an unchanged Markdown link, drop focus, and allow the disclosure fallback to steal focus.
- Approach: Keep the default Markdown anchor and table renderer types stable at module scope, then strengthen the regression coverage to observe the scheduled focus-restoration frame and direct DOM-node identity.
- Impacted modules: `packages/web/src/components/ui/markdown.tsx` and focused Web DOM tests.

**Review findings**
✅ 1. Stable renderer references address the React reconciliation cause while preserving caller override precedence and the existing safe-link/table behavior.
✅ 2. The compose-status regression now drains the focus RAF, so the pre-fix race is observable; the direct characterization separately verifies DOM identity and focus preservation.

**Action taken**
- Approved at exact head `f849bfb191de529aa36d2f69a1649ed701c63840`.

PR #2228, opened 2026-08-07
fix: stop bolding the Chat Summary lead line

Gandy2025:
## Problem

Two issues in the expanded **Current state** card, both from promoting the first physical line.

**The lead was bold against nothing.** The promotion styles the lead at `--text-subtitle` weight 600. But most descriptions are a single physical line — sampling 26 real chats with a description, **15 (58%)** were one line, 108–232 characters. There the lead *is* the whole summary, so the emphasis separates the text from nothing and the entire brief renders as a heavy block.

**A single-line card was too tall.** The lead reserved `margin-bottom: var(--sp-3)` to separate itself from supporting copy. With no supporting copy that became dead space: measured on a one-line brief, **13px of slack** below the text and a **69px** card for a 19px line.

## Change

- the lead reads as the lead through **size and colour only** — `--text-subtitle` size, `--fg`, body weight. Authored `**bold**` inside a description is untouched.
- move the lead's separation out of the inline style and into `index.css`, and collapse it with `:last-child` so a lead with nothing after it reserves no gap. The blank-line form is unaffected: the lead is its paragraph's last child there too, and the following paragraph carries its own top margin.
- add the single-line shape to the DEV preview gallery and to deterministic coverage — it was the majority real shape with no fixture.

## Validation

- `pnpm check`
- `pnpm typecheck` — 11/11 tasks, including the web `lint:tokens` guard
- `pnpm --filter @first-tree/web test` — 245 files / 2,247 tests passed
- real browser measurement of the rendered component:

  | shape | lead weight | lead size | lead margin-bottom | card height | slack below text |
  | --- | --- | --- | --- | --- | --- |
  | single line | 400 | 14px | 0px | **57px** (was 69) | **1px** (was 13) |
  | soft-break, 3 lines | 400 | 14px | 12px | 108px | 1px |

  Authored `<strong>` still computes to 600 in the multi-line fixtures, confirming only the promotion styling lost weight.

## Note

Hierarchy is now carried by a 14px/13px step plus `--fg` vs `--fg-2`, which is deliberately subtler than before. That was the agreed direction; worth a look in the live Workspace to confirm the result line still reads as the result.

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -911,6 +911,19 @@ video[data-onboarding-orientation-video]::cue {
   }
 }
 
+/* Chat summary lead line: the promoted first physical line of a current-state
+   brief. It is separated from supporting copy by a gap, but a lead with nothing
+   after it is the whole summary — the `:last-child` reset keeps that case from
+   reserving space for copy that does not exist. In the blank-line form the lead
+   is also its paragraph's last child, and the following paragraph carries its
+   own top margin, so the separation survives there. */
+[data-summary-part="lead"] {
+  margin-bottom: var(--sp-3);
+}
+[data-summary-part="lead"]:last-child {
+  margin-bottom: 0;
+}
+
 /* Chat summary floating card: drops in from the bar's lower edge. */
 @keyframes chat-summary-card-in {
   from {
--- a/packages/web/src/pages/chat-summary-preview.tsx
+++ b/packages/web/src/pages/chat-summary-preview.tsx
@@ -22,6 +22,12 @@ const CURRENT_STATE = [
   "**Next:** Validate the same hierarchy in the live Workspace.",
 ].join("\n");
 
+// The most common real shape: one physical line, no supporting copy. The lead is
+// the whole summary here, so it must read as ordinary prose and must not reserve
+// separation below itself.
+const SINGLE_LINE =
+  "正在核验团队的成员规模、First Tree 安装完成情况与实际使用情况。口径按 Asia/Taipei T+1 截止，并将安装状态与使用行为分开统计。";
+
 const PLAIN = "Reviewing PR 1207 — server-side freshness fields landed; wiring the collapsed-bar first line next.";
 
 const HEADING_FIRST = [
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The patch removes uninformative weight and dead space from the common single-line summary while preserving authored Markdown emphasis and the existing separation for summaries with supporting copy.

**Risk level**: C
- Semantic lift: none; the change is limited to Chat Summary presentation and does not touch subscriptions, WebSocket parsing, or the Web data layer.

**PR summary**
- Problem: People opening a chat's Current state card often saw the entire one-line brief rendered as a heavy lead with an empty gap underneath, even though there was no supporting copy to distinguish.
- Approach: Keep the first physical line's size and foreground contrast but use body weight, move its separation into CSS so it collapses when the lead is the paragraph's last child, and add the prevalent single-line shape to the preview and deterministic coverage.
- Impacted modules: Workspace Chat Summary styling and rendering, its unit coverage, and the DEV preview gallery.

**Review findings**
✅ 1. The body-weight override applies only to the promoted wrapper, while authored `<strong>` nodes remain nested and retain their own emphasis. [`packages/web/src/pages/workspace/center/chat-summary.tsx:254`]
✅ 2. The CSS condition matches the renderer's two relevant DOM shapes: a single-line lead is the paragraph's last child and loses the gap, while a soft-break lead has following `<br>`/content siblings and retains it. [`packages/web/src/index.css:920`]
✅ 3. The added fixture covers the previously missing majority shape and asserts the DOM condition that drives the collapse without changing the one-document Markdown path. [`packages/web/src/pages/workspace/center/__tests__/chat-summary.test.ts:289`]

**Action taken**
- Approved exact head `eac2959fec4ba44bffb0bde14c591b04db82512e`.

Static review only; I did not run tests or QA.

yuezengwu (approved):
Approved after static review of exact head `eac2959fec4ba44bffb0bde14c591b04db82512e`.

The implementation matches the intended presentation change: the promoted lead keeps its subtitle size and foreground contrast while using body weight, and authored Markdown `<strong>` nodes retain their own emphasis. The CSS reset removes the dead lead margin for a single-line document; soft-break supporting copy keeps the explicit gap, while blank-line/block supporting copy keeps its normal block spacing. The preview and unit fixture add the previously missing common single-line shape without changing Markdown parsing or document structure.

No blocker found. This PR does not change core data structures or the database. Static review only; I did not run tests or QA.

PR #2230, opened 2026-08-07
feat(web): switch agents from the Agent Detail breadcrumb

Gandy2025:
## Why

Changing which agent you are configuring meant leaving Agent Detail for Team, finding the target agent, and entering it again — at least two navigations for what is really one action. The breadcrumb's current segment already answers "which agent am I looking at"; this makes it also answer "and I can change it".

## What

`Team / <agent name>` gains a small chevron and becomes a menu trigger. Everything else in the header is unchanged.

The panel is deliberately plain:

- Non-human agents only, name only, in the Team page's order (see below) and off the same query cache, so it opens warm when you arrived from Team.
- The open agent is checked; `@handle` appears only where two display names collide.
- A filter input appears only past eight agents.
- Switching keeps the current section (`/runtime` → `/runtime`) and pushes a history entry, so Back returns to the previous agent. Picking the open agent is a no-op.
- Arrow keys / Home / End move between rows, Enter commits, Escape closes and restores focus to the trigger. Focus opens on the current agent, or on the filter when there is one; while the roster is failing it opens on Retry, and a successful retry hands the keyboard to the rows it just rendered.
- A failed roster load stays inside the panel with a retry; the page behind it is untouched.

Not included, by design: avatars, presence, grouping, recents, a "View all agents" footer, a dedicated shortcut, and a persistent agent rail.

Human members keep the plain breadcrumb — they are never switch targets, so a switcher there would open a list with nothing checked.

## Shared roster order

The switcher must show the roster in the order Team shows it, but the list endpoints return `createdAt DESC` and Team reprojects client-side. Rather than reimplement that, this extracts it:

- `compareRosterAgents(selfMemberId)` — the viewer's own agents first, then `displayName.localeCompare(..., { sensitivity: "base" })`. `buildTeamData`'s `sortRows` now sorts through it, so there is exactly one definition.
- `orderAgentsLikeTeam(agents, selfMemberId)` — organization-visible group then private group, each through that comparator: Team's two sections flattened for this ungrouped menu.

The "current agent is absent from a visibility-filtered roster" case is folded into the same ordering step, so that agent keeps a checked row without gaining a special top position.

## Drive-by fix

`Instructions` now redirects non-editors to Profile, the way `Runtime`, `Repositories`, and `Responsibilities` already do. Without it, switching to an agent you cannot configure while on Instructions left you on a permanently empty editor (its config query is disabled for non-editors).

## Verification

- `pnpm --filter @first-tree/web typecheck` (tsc + design-token guardrails) — clean. `biome check` on the touched paths — clean.
- `pnpm --filter @first-tree/web test` — 2264 passed. The one failing suite, `tests/visual-regression.spec.ts`, is a Playwright spec vitest cannot parse; it fails identically on an untouched tree.
- `agent-switcher-dom.test.tsx` (13 cases): roster source per role, human exclusion, the checked row, handle disambiguation, the open-agent-missing-from-roster case, Team ordering from a divergent API order, filter threshold and filtering, Enter-commits-top-match, keyboard traversal, the no-op re-pick, and the error → retry → focus-recovery path.
- `team-groups.test.ts`: `orderAgentsLikeTeam` on a roster whose API order differs from presentation order, plus a direct equivalence case asserting it equals `[...buildTeamData().publicAgents, ...buildTeamData().privateAgents]` for the same viewer.
- `agent-detail-page-dom.test.tsx`: switching preserves the section and Back returns to the previous agent; a human member keeps the plain breadcrumb.
- Browser pass against a local server + web dev build. Ordering was checked with a deliberately discriminating seed — `Alpha Shared`, `Zeta Shared`, `Beta Shared` (organization) and `A Private Agent` (private), which the API returns as `A Private Agent, Beta Shared, Zeta Shared, Alpha Shared`; the switcher and Team both render `Alpha Shared, Beta Shared, Zeta Shared, A Private Agent`. Also covered: resting breadcrumb and chevron hover, the panel above and below the filter threshold, filter + Enter switching into the same tab, browser Back, keyboard traversal, Escape focus restore, dark theme, and a 390px viewport where the panel stays inside the viewport. No console errors from this change.

## Not verified

The non-editor Profile fallback was exercised through the tab guards and unit tests, not in the browser — the local dev identity is an admin of every agent in its org. The trigger's focus ring was not pixel-checked; it uses the same focus classes as every `Button`.

### DIFF
--- a/packages/web/src/pages/agent-detail.tsx
+++ b/packages/web/src/pages/agent-detail.tsx
@@ -37,6 +37,7 @@ import { invalidateDisplayNameQueries } from "./../lib/identity-cache.js";
 import { cn } from "./../lib/utils.js";
 import { canManageAgentDetail } from "./agent-detail/access.js";
 import { isBindableClient } from "./agent-detail/action-state.js";
+import { AgentSwitcher } from "./agent-detail/agent-switcher.js";
 import { useAgentResources } from "./agent-detail/capability-section.js";
 import type { AgentDetailContext, RuntimeSwitchClaimView } from "./agent-detail/layout-context.js";
 import {
@@ -453,7 +454,21 @@ function AgentDetailPageView() {
             Team
           </BreadcrumbLink>
           <BreadcrumbSep />
-          <BreadcrumbCurrent>{agent.displayName}</BreadcrumbCurrent>
+          {/* Humans have no switcher: the list is agents only, so their own
+              row would never be in it. */}
+          {isHuman ? (
+            <BreadcrumbCurrent>{agent.displayName}</BreadcrumbCurrent>
+          ) : (
+            <AgentSwitcher
+              currentAgent={agent}
+              onSelect={(nextUuid) =>
+                // Keep the section the user is reading. A pushed entry (not
+                // `replace`, unlike the section links) makes Back return to
+                // the previous agent.
+                navigateAway(`/agents/${encodeURIComponent(nextUuid)}/${currentTab?.path ?? "profile"}`)
+              }
+            />
+          )}
         </Breadcrumb>
 
         <div className={cn("flex gap-4", isNarrow ? "flex-col items-stretch" : "items-center justify-between")}>
--- a/packages/web/src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx
+++ b/packages/web/src/pages/agent-detail/__tests__/agent-detail-page-dom.test.tsx
@@ -4,7 +4,7 @@ import type { Agent, AgentResourcesOutput, AgentRuntimeConfig } from "@first-tre
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { act, type ReactElement } from "react";
 import { createRoot, type Root } from "react-dom/client";
-import { MemoryRouter, Navigate, Route, Routes, useLocation } from "react-router";
+import { MemoryRouter, Navigate, Route, Routes, useLocation, useNavigate } from "react-router";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { HubClient } from "../../../api/activity.js";
 import { ApiError } from "../../../api/client.js";
@@ -390,6 +390,20 @@ async function renderDom(
 function LocationEcho() {
   const location = useLocation();
   return <div>{location.pathname + location.search}</div>;
+}
+
+/** Tab body that reports where it is and can walk the history back. */
+function LocationEchoWithBack() {
+  const location = useLocation();
+  const navigate = useNavigate();
+  return (
+    <div>
+      <div data-testid="here">{location.pathname}</div>
+      <button type="button" onClick={() => navigate(-1)}>
+        Go back
+      </button>
+    </div>
+  );
 }
 
 async function click(element: Element | null): Promise<void> {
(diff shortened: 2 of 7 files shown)


yuezengwu (changes requested):
The feature direction is sound: the Agent Detail breadcrumb becomes a lightweight agent switcher, preserves the current section/history, excludes human identities, and adds the missing Instructions permission fallback. The `displayName`/conditional `@handle` treatment and visibility-aware roster source also match the agent identity model.

One behavior in the stated contract is not implemented yet, so I’m requesting changes:

- **The switcher does not use the Team page’s order.** `options` keeps the raw paginated API order. Both member and admin list endpoints return `createdAt DESC`, while `TeamPage` transforms that data in `buildTeamData`: organization-visible agents are shown before private agents, and each group is sorted with the viewer’s agents first and then by `displayName`. With mixed visibility/ownership or simply non-alphabetical creation order, the switcher and Team page therefore disagree. Please extract/reuse the same ordering rule (flattening the groups if the switcher remains ungrouped) and add a regression case whose API order differs from the expected Team order.

I reviewed the latest head `606e9c4f5b4bf5941d2125688e71843b7444b3f4`. Per review scope, I inspected the diff and relevant surrounding code without rerunning tests or QA.

baixiaohang (changes requested):
- Rationale: The switcher reuses the Team query cache but not the Team roster projection, so its visible order diverges from the workflow it claims to preserve.

**Risk level**: C
- Path baseline: only `packages/web/**` is changed -> C.
- Semantic lift: none; this consumes existing roster and route state without changing subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Operators configuring several agents currently have to leave Agent Detail, relocate the next agent on Team, and navigate back into the same section.
- Approach: Turn the current-agent breadcrumb into a visibility-aware, keyboard-operable switcher that preserves the active section and browser history; human identities remain plain breadcrumbs, and non-editors are redirected away from Instructions.
- Impacted modules: Agent Detail shell, new breadcrumb switcher, Instructions route guard, and DOM tests.

**Review findings**
❌ 1. The stated "Team page order" is not implemented. `options` preserves the raw paginated API order (`createdAt DESC`), while Team projects the same data into organization-visible then private groups and sorts each group with the viewer's agents first, then by `displayName`. With mixed ownership/visibility—or simply a different creation order—the two surfaces disagree. Please extract and reuse the Team roster ordering/projection (flattening it for this ungrouped menu is fine) and add a regression where API order differs from Team order. [R1 / `packages/web/src/pages/agent-detail/agent-switcher.tsx:109`]

✅ The non-human filtering, member/admin visibility source, conditional handle disambiguation, section/history preservation, permission fallback, and retry-focus recovery otherwise align with the current Agent Detail model.

This independently confirms the blocker already raised on the same head by yuezengwu; I found no additional blocking issue.

**Action taken**
- Submitted request changes on head `606e9c4f5b4bf5941d2125688e71843b7444b3f4`.

Gandy2025:
Confirmed and fixed in `51ddd66e1`. You were both right — I had reused the Team page's query *cache* and assumed that carried its order, but the projection lives in `buildTeamData`, and the raw list is `createdAt DESC`.

**Fix** — extracted the rule instead of reimplementing it, so the two surfaces cannot drift apart again:

- `compareRosterAgents(selfMemberId)` — the viewer's own agents first, then `displayName.localeCompare(..., { sensitivity: "base" })`. `buildTeamData`'s `sortRows` now sorts through it, so it has exactly one definition.
- `orderAgentsLikeTeam(agents, selfMemberId)` — organization-visible group then the private group, each through that comparator. This is the flattened form for the ungrouped menu.
- The switcher calls `orderAgentsLikeTeam` and also folds the "current agent missing from the roster" case into it, so that agent now lands in its natural position rather than being pinned to the top.

**Regression coverage where API order differs from Team order**, as requested:

- `team-groups.test.ts` — a roster whose API order is `Zeta(private, theirs), Alpha(theirs), Mine Zulu(mine), Private Mine(mine, private), Beta(theirs)` must render as `Mine Zulu, Alpha, Beta, Private Mine, Zeta`. A second case asserts `orderAgentsLikeTeam(...)` equals `[...buildTeamData().publicAgents, ...buildTeamData().privateAgents]` for the same viewer — that is the direct equivalence guard.
- `agent-switcher-dom.test.tsx` — the same divergent roster asserted on the rendered rows, plus the existing cases updated to the Team order.

**Verified in a browser** against a local server, seeding an order that discriminates: created `Alpha Shared`, `Zeta Shared`, `Beta Shared` (organization) and `A Private Agent` (private). The API returns `A Private Agent, Beta Shared, Zeta Shared, Alpha Shared`. Both surfaces now render `Alpha Shared, Beta Shared, Zeta Shared, A Private Agent` — note `A Private Agent` sorts first alphabetically but correctly trails, and neither surface matches the API order.

`pnpm --filter @first-tree/web test` — 2264 passed. typecheck and design-token guardrails clean.

One thing I dropped along the way: I had briefly added a defensive branch for a visibility value that is neither `organization` nor `private`. The schema union makes that unreachable and the test for it would not typecheck, so `orderAgentsLikeTeam` just splits on the two real values.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The prior ordering blocker is resolved through one shared roster comparator/projection, with direct equivalence coverage between Team and the switcher.

**Risk level**: C
- Path baseline: only `packages/web/**` is changed -> C.
- Semantic lift: none; the change consumes existing roster and route state without changing subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Operators configuring several agents currently have to leave Agent Detail, relocate the next agent on Team, and navigate back into the same section.
- Approach: Turn the current-agent breadcrumb into a visibility-aware, keyboard-operable switcher that preserves the active section and browser history; human identities remain plain breadcrumbs, and non-editors are redirected away from Instructions.
- Impacted modules: Agent Detail shell and switcher, Team roster projection, Instructions route guard, and DOM/unit tests.

**Review findings**
✅ The previous blocker is fixed: `compareRosterAgents` now owns the within-group order, Team delegates to it, and `orderAgentsLikeTeam` provides the exact flattened organization-then-private projection for the ungrouped switcher.
✅ The missing-current-agent path is folded into the shared ordering step, so it keeps a checked row without gaining a special top position.
✅ Regression cases use API order that differs from presentation order and directly assert that the switcher projection equals Team's concatenated groups.

No blocking findings remain on head `51ddd66e13b5a4af2921fb2a5da3680e6c539bc6`. Per review scope, I inspected the latest diff and surrounding code without rerunning tests or QA.

**Action taken**
- Approved.

yuezengwu (approved):
Re-reviewed the latest head after the requested change.

The ordering blocker is resolved: `compareRosterAgents` is now the single within-group rule used by `buildTeamData`, `orderAgentsLikeTeam` provides the shared organization-then-private flattened projection, and the switcher applies it after restoring a current agent that is absent from the visibility-filtered roster. The new regression coverage deliberately differs from API order and directly checks equivalence with the Team projection.

I found no additional blocking issue. The feature still preserves the intended Agent Detail boundary, identity labeling, visibility source, section/history navigation, permission fallback, and keyboard/retry behavior.

Approved on head `51ddd66e13b5a4af2921fb2a5da3680e6c539bc6`. Per review scope, I inspected the incremental fix and full relevant diff without rerunning tests or QA.

PR #2233, opened 2026-08-07
refactor(server): remove unused services and audit script

baixiaohang:
## Summary

- remove the unreachable server stats service and its isolated tests
- remove the unused Context Tree binding audit runner, helper, and tests
- keep the production repository-ownership identity logic used by binding writes

## Local validation

- `pnpm exec biome check packages/server/package.json`
- `VITEST_MAX_FORKS=2 pnpm --filter @first-tree/server exec vitest run src/__tests__/org-settings.test.ts` (85 tests)
- repository search confirms no remaining audit-script references

GitHub CI provides the full repository regression suite.

## Scope

No API, WebSocket, configuration, schema, migration, data, or legacy compatibility behavior changes.


yuezengwu (approved):
Reviewed the full diff. This cleanly removes the unreachable stats service and its isolated tests, relocates the audit-only helper and tests under scripts, and adds a focused scripts TypeScript config so the relocated code remains covered by package typechecking. I found no remaining references to the stats service, and the audit repository-identity and unresolved-binding behavior is preserved. No API, configuration, schema, migration, or data changes are introduced.

yuezengwu (changes requested):
Requesting changes on the synchronized head.

**Blocker — do not remove the historical binding audit as “unused” without a completed remediation or an equivalent replacement.** `packages/server/scripts/audit-shared-context-tree-repos.ts` is a manually invoked operator entrypoint, so the absence of imports is expected; `seed-legacy-chat-data.ts` follows the same pattern. PR #2122 added this audit specifically because the new binding guard prevents future ordered collisions but does not repair bindings already written under the old naming/adoption behavior. Its reviews also required the provider-strength grouping and fail-safe unresolved result that this PR now deletes together with their tests.

The current Context Tree binding contract still records that pre-rule shared bindings are not separated retroactively. I found no replacement operator path and no evidence that every deployment has been audited and all affected rows remediated. Removing the runner/helper/tests therefore removes the repository’s safe way to detect an existing cross-Team context-sharing exposure. Please keep the audit tooling (the earlier scripts relocation is a reasonable shape), or provide evidence that the historical population is fully remediated and land an equivalent supported audit path before deleting it.

The stats service/test deletion remains safe: no callers remain. There is no schema migration, but the removed audit reads core tenant binding data and its retirement requires explicit human ownership. I did not run tests for this code-review pass.

baixiaohang:
Thanks for calling out the distinction between an unreferenced operator entrypoint and ordinary dead code. I rechecked the provenance in #2122 and confirmed that the production guard does not retroactively repair historical bindings.

After that review, the project owner explicitly confirmed the retirement decision: this audit was never operationalized or run, and they want the runner, helper, and isolated tests removed. There is no replacement audit path in this PR; the decision is to retire the unused manual tooling rather than preserve it based on possible future use. The production binding guard and `contextTreeRepoOwnershipIdentity` remain unchanged.

Accordingly, the current head is intentional and no additional code change is needed for this thread. Please re-review with that ownership decision recorded.

yuezengwu (approved):
Re-reviewed the unchanged synchronized head after the ownership decision recorded in the PR.

The previous blocker is resolved as an explicit governance decision, not because the runner was technically unreachable: the Cloud owner accepts retiring the never-operationalized manual audit with no replacement, while acknowledging that the production guard does not retroactively repair historical bindings. The production `contextTreeRepoOwnershipIdentity` and binding guard remain unchanged, and the stats service/test deletion still has no remaining callers.

Approved with the residual risk recorded: legacy/shared bindings will no longer have this in-repository discovery path. This approval relies on the explicit owner decision in the thread. No schema or migration changes are present, and I did not run tests for this review pass.

PR #2244, opened 2026-08-07
refactor: retire compatibility below Client 0.5.12

baixiaohang:
## Summary

- retire Client/CLI compatibility older than the chronological `0.5.12` release boundary
- emit only canonical Agent, WebSocket capability, runtime capability, Context Tree usage, and short-code connect-token contracts
- always require the current runtime-session proof for non-human Agent HTTP calls, except the existing route-scoped `agent_outbox` token
- remove the deprecated `FIRST_TREE_GITLAB_EGRESS_ALLOWLIST` input
- enforce one fail-closed runtime-switch floor across Server and Web: stable `0.5.12+` and repository-issued staging builds from `0.5.13-staging.*`

## Preserved contracts

- `wsSessionResetV1` remains a two-sided, fail-closed negotiation, including route rechecks, parked-fence handling, and Reset receipts
- historical stored Context Tree usage payloads without `nodePath` still read as `null`; only new inbound events require the field
- human Agent requests and the narrow `agent_outbox` exception retain their existing authorization behavior
- runtime provider capability, ownership, retirement, minimum Client version, and atomic switch guards remain
- persisted legacy Agent types remain only in migrations, historical fixtures, and explanatory history; no DB schema or data migration is included

## Local verification

Local verification was intentionally focused per maintainer direction; GitHub CI is responsible for the broad regression suites.

- focused Shared schema/config tests: 85 passed
- focused Client capability/WS/Reset tests: 111 passed
- focused CLI auth/runtime tests: 51 passed
- focused Server Agent wire/Auth/Reset/runtime-switch/publisher tests: 183 passed
- focused runtime-switch floor tests: Shared 17 passed, Server 22 passed, Web capability/DOM 40 passed
- changed-file Biome checks and `git diff --check` passed


yuezengwu (changes requested):

- **[P1] Keep a server-side runtime-switch floor at Client 0.5.12** (`packages/server/src/services/agent-runtime-switch.ts`, around the target-client preconditions). Removing `sdkVersion` from the target query and deleting the version assertion lets connected Clients with a missing or pre-0.5.12 version pass solely because they reported a provider capability. That is not safe: the v0.5.11 Client's `handleAgentPinned` returns for an existing agent without applying a changed `runtimeProvider`; the reconfiguration path first exists in v0.5.12. The Server can therefore commit the Cloud route, revoke the old runtime proof, archive sessions, reactivate the agent, and return success while the target daemon keeps the old provider or leaves the agent offline. Please retain a fail-closed Server check using 0.5.12 as the minimum supported version, and keep the Web candidate/blocker behavior consistent with it.

The remaining compatibility removals preserve the reviewed current contracts: Agent HTTP proof enforcement keeps the human and route-scoped `agent_outbox` exceptions, `wsSessionResetV1` remains the sole two-sided Reset capability, and stored legacy Context Tree usage still degrades a missing `nodePath` to `null` on reads. No database schema or migration change is present.

baixiaohang:
The requested runtime-switch floor is restored on exact head `d3691cf38` after the maintainer confirmed the A6 correction:

- Server fail-closes before recording a switch claim unless the target Client is a supported chronological release.
- Server and Web share one gate: stable `0.5.12+`; repository-issued staging builds use the exact `0.5.patch-staging.<run>.<attempt>` form and are accepted from patch 13 onward.
- Missing/old versions, the earlier `0.14.x` line, `0.5.12-staging.*`, arbitrary prereleases, malformed versions, and unsafe integers remain blocked.
- The deprecated runtime enforcement config/env switch remains removed.

Focused verification passed: Shared 17 tests, Server runtime-switch 22 tests, Web capability/DOM 40 tests, plus Shared/Server/Web typechecks, changed-file Biome, and `git diff --check`.

yuezengwu (approved):
Re-reviewed exact head `d3691cf38bb7d248cccad5ef2d5ca01e08e55daa`. The requested runtime-switch floor is restored:

- Server and Web share the same fail-closed release predicate.
- Stable 0.5.12+ and repository-shaped staging builds from 0.5.13 onward are accepted.
- Missing, malformed, older 0.14.x / 0.5.11, pre-floor staging, and arbitrary prerelease versions are rejected.
- The Server checks the target Client before recording the switch claim, while the existing ownership, retirement, provider-capability, and atomic-switch guards remain.

The blocker is resolved; no new findings in the incremental diff. Approving.

PR #2042, opened 2026-07-28
Add Need you request review flow

yuezengwu:
## Summary

- replace the chat-level Attention priority projection with a request-level **Need you** queue and exact count on desktop and mobile Chat lists
- add a responsive, sequential review surface with Submit, Skip, lazy earlier-chat context, and read-only full-chat inspection
- give the existing Chat takeover the same actions and a non-resolving inspect mode that keeps ordinary sending blocked
- add the constrained **Ask agent** protocol: a server-trusted same-chat clarification marker, an explicit runtime reply envelope, durable reply threading, waiting feedback, timeout recovery, and retry
- keep the routed participant visible in history with an agent capsule on Ask agent clarifications, agent replies, and resolving Submit/Skip messages, without duplicating normalized mentions
- simplify chat ordering to Pinned first and all remaining chats by real-work activity; unread, open-request, and recovery state remain distinct, coexisting row icons

## Why

Open `chat ask` requests need a focused, self-contained review flow without turning unread updates or agent recovery into the same inbox. Users should be able to understand and resolve questions across chats, ask for clarification without resolving the request, and inspect the original chat without bypassing the existing “answer before ordinary messaging” contract.

## Product and protocol notes

- Need you counts request messages, not chats, and processes them FIFO across chats.
- Submit and Skip resolve and advance; Ask agent keeps the original request open.
- Ask agent replies remain normal durable `chat send --reply-to` messages. The delivered clarification includes a trusted, one-turn steering envelope that tells the original asking agent to answer from existing context, not continue or resolve the task, and send exactly one threaded clarification reply.
- `chat send` is complete-message delivery today, so v1 shows a waiting animation and then the full durable response rather than token streaming.

## QA hardening

- remove the background-composer bypass while a request takeover is active and move focus into the review card
- make Escape non-resolving, and freeze answer/edit/close actions throughout Ask agent sending and waiting, including host navigation and browser Back/Forward
- retain request-scoped text, options, image, and document drafts across full-chat inspection
- preserve both rolling directions: older Web accepts the empty compatibility bucket from the new Server, while new Web recovers legacy pinned rows transported in `priorityRows.attention` without restoring Attention ordering
- make `/m/chat` canonical while preserving query/hash redirects from `/m/work` and `/m/now`, and update the PWA launch route
- add Class B/C negative coverage for wrong-org, same-org nonparticipant, cross-org, and request/chat binding failures
- add the reusable cross-surface `need-you-request-review-journey` QA case and update the mobile PWA case

[...]

PR #2245, opened 2026-08-07
refactor: remove Web/Server rolling compatibility

baixiaohang:
## Summary

- remove Web/Server rolling-version compatibility because Web and Server ship as one version
- use only `/settings/account` and `/settings/integrations/github` for OAuth returns, and accept only encrypted OAuth nonce cookies
- accept only v2 chat-list cursors and the canonical pinned-only priority projection while retaining complete additive `rows` for CLI Workspace
- remove retired onboarding/feedback endpoints, browser route aliases, Workspace query upgrades, Agent anchor conversion, and the legacy mobile module
- update Momentic and reusable QA contracts to exercise only canonical routes and same-version Web/Server behavior

## Preserved contracts

- OAuth CSRF matching, signed-state expiry, safe relative return paths, provider error handling, and GitHub install authority remain fail-closed
- current onboarding and landing-campaign behavior remains; obsolete request fields are rejected
- complete additive chat rows remain available to CLI `0.5.19`, including pinned chats
- historical reads, backfills, current legacy chat calls, and DB schema/data remain unchanged
- canonical `/m/chat`, Agent detail tabs, Settings pages, and onboarding entry points remain

## Local verification

Local verification was intentionally focused per maintainer direction; GitHub CI is responsible for the broad regression suites.

- focused Shared chat/onboarding schemas: 19 passed
- focused Server OAuth/chat-list/onboarding/fallback tests passed, including 49 OAuth flow/cookie tests
- focused Web route/OAuth/Workspace/mobile tests: 97 passed
- focused CLI Workspace chat-list test: 20 passed
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/agent-messages.test.ts --maxWorkers=1 --minWorkers=1`: 8 passed
- `pnpm --filter @first-tree/web exec vitest run src/__tests__/app-routes.test.tsx --maxWorkers=1 --minWorkers=1`: 27 passed
- `npx momentic lint e2e/github-install-identity-gate.test.yaml`: passed
- changed-file Biome checks and `git diff --check` passed


yuezengwu (changes requested):

**Goal and core changes**

This PR removes Web/Server rolling-compatibility bridges now that the SPA and API ship in one server artifact. It narrows OAuth returns/cookies, chat-list cursor and priority schemas, onboarding/task-chat request schemas, browser aliases, and mobile legacy exports to their canonical same-version contracts. The Shared wire schemas are core contract changes that deserve human attention; I found no DB schema or data migration in this diff.

**Blocking findings**

1. Keep the installed PWA identity stable. `packages/web/public/manifest.webmanifest:5` changes `id` from `/m/now` to `/m/chat`, but the manifest `id` is the application's identity, not a route. The Web App Manifest specification says a non-matching identity should be treated as a distinct application; Chrome's guidance likewise uses a stable `id` specifically so `start_url` can change. `start_url` has already been `/m/chat` since #2042, while `/m/now` remained as the stable identity. Revert only `id` to `/m/now`; that does not restore navigation to the retired route.

2. Preserve a non-dead-end fallback for mobile URLs. `packages/web/src/app.tsx:480-485` removes both named aliases and the `m/*` fallback, and the updated route test explicitly expects `/m/retired` (as well as `/m/now` and `/m/work`) to render no recognized surface while leaving the URL untouched. Because there is no outer catch-all route, authenticated direct visits now render an empty app. The current Mobile Surface contract requires that a mobile route never dead-end and that settled unusable mobile visits recover to desktop. Removing named aliases is compatible with the PR goal, but retain a catch-all recovery (to `/m/chat` when mobile is available, otherwise `/`) and assert a rendered destination rather than a blank route.

**Durable-contract note**

The PR also changes the current Web Console route-map decision, which still records the removed browser paths as redirects. That durable contract needs an owner-reviewed, source-backed Context Tree update/cross-link rather than silently drifting after merge.

I reviewed the diff statically and did not run tests or QA.

yuezengwu (approved):
**Recommendation:** approve

The follow-up resolves both blockers on exact head `741e6f3761922bd1c0d03122b56328e1e6e897cc`:

- The installed PWA identity is stable again (`id: /m/now`) while the launch route remains canonical (`start_url: /m/chat`).
- Authenticated unmatched `/m/*` routes now pass through `MobileExperienceGate`: usable mobile channels recover to `/m/chat`, while a settled unusable channel recovers to the desktop root. The focused route tests cover both outcomes.

I found no new blocking issue in the four-file incremental diff. The broader Shared chat/onboarding wire-schema narrowing remains a core contract change for human attention; the PR still has no DB schema or data migration.

The general browser-alias removals still change the current Context Tree route-map decision. A separate owner-reviewed, source-backed Tree update/cross-link remains merge-preparation work; source-review authority does not authorize me to publish that Tree change.

Reviewed source and deterministic test changes only; tests and QA were not run.

baixiaohang:
The source-backed Context Tree update requested during review is open as draft: [first-tree-context#909](https://github.com/agent-team-foundation/first-tree-context/pull/909).

It updates the canonical Web route map, the generic mobile recovery contract, and the stable PWA identity boundary. It will remain draft until this source PR merges, then be reconciled against the merged code before being marked ready.

PR #2247, opened 2026-08-07
ACE-40 feat(web): add an in-app entry to the public template library

Gandy2025:
## Summary

- The public Template Library at `/templates` had no link anywhere in the app. Outside `app.tsx`'s route definitions and a few doc comments, nothing in `packages/web/src` pointed at it — landing, sidebar, and the New Agent dialog all had no entry, so users could only reach it by typing the URL or arriving from an off-site placement.
- Adds a "Browse the template library" link to the New Agent dialog's Responsibilities section — the moment a user is actually choosing a starting responsibility, which is the shortest path from discovery to a working agent.

Placement details:

- The link sits **inside** the catalog-gated Responsibilities block, so an empty or failed catalog never advertises an empty library and the plain create path stays unchanged.
- It opens in a **new tab** rather than navigating, so an in-progress draft in the dialog (display name, visibility, computer, runtime, current selection) survives the detour.
- Reports `agent_template_library_open` with `surface: "new_agent"`, matching the existing `agent_template_select` surface convention.

The canonical `/templates/:slug?use=1` intent path and its strict `parseAgentTemplateIntentPath` resolution are untouched.

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm test`

`packages/web/src/__tests__/app-routes.test.tsx` has 7 failures on this branch, but they reproduce identically on the base commit (`54bcb109c`) with no changes applied — they are pre-existing and unrelated to this PR. Everything else passes, including the 10 tests in `new-agent-dialog-templates.test.tsx`.

## Change Surface

## Notes

- docs or tests updated to match: `new-agent-dialog-templates.test.tsx` — a new test covers the link's presence, new-tab behavior, analytics event, and that clicking it leaves the draft intact; the two existing "catalog empty / failed" tests now also assert the link is absent.
- follow-up work: if the product wants additional entry points (landing, sidebar, or Agent Detail → Responsibilities), they can be added on top of this without changing the intent path.

### DIFF
--- a/packages/web/src/components/__tests__/new-agent-dialog-templates.test.tsx
+++ b/packages/web/src/components/__tests__/new-agent-dialog-templates.test.tsx
@@ -182,6 +182,10 @@ function buttonByText(text: string): HTMLButtonElement {
   return button;
 }
 
+function templateLibraryLink(): HTMLAnchorElement | null {
+  return document.body.querySelector<HTMLAnchorElement>('a[href="/templates"]');
+}
+
 function optionCardByText(text: string): HTMLElement {
   const label = [...document.body.querySelectorAll("label")].find((el) => el.textContent?.includes(text));
   if (!label) throw new Error(`Missing option card "${text}". Body: ${document.body.textContent ?? ""}`);
@@ -241,6 +245,8 @@ describe("NewAgentDialog template responsibilities", () => {
     await renderDialog(onCreated);
     await flush();
     expect(document.body.textContent).not.toContain("Responsibilities");
+    // No entry to an empty library either — the link is gated on the same catalog.
+    expect(templateLibraryLink()).toBeNull();
 
     await fillNameAndOpenPickerSafe();
     async function fillNameAndOpenPickerSafe() {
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change adds a focused discovery path to the existing public library while preserving the in-progress creation draft and leaving the plain-create fallback unchanged.

**Risk level**: C
- Semantic lift: none; this is local presentation/navigation and analytics behavior, with no subscription, WebSocket, or data-layer changes.

**PR summary**
- Problem: Users creating an Agent had no discoverable in-app path to browse the public Template Library, even when they were choosing responsibilities.
- Approach: Add a catalog-gated link in the New Agent dialog that opens the existing `/templates` route in a new tab, preserving the draft and recording the discovery event.
- Impacted modules: New Agent dialog and its focused Template responsibility tests.

**Review findings**
✅ 1. The link is shown only after a usable non-empty catalog loads, so empty and failed catalog states retain the existing plain-create behavior. [`packages/web/src/components/new-agent-dialog.tsx`:906]
✅ 2. The new-tab link includes opener protection, and the regression test covers visibility, target behavior, analytics, and retention of the current draft. [`packages/web/src/components/__tests__/new-agent-dialog-templates.test.tsx`:281]

**Action taken**
- Approved.

yuezengwu (approved):
Approved. The change adds the missing Template Library discovery path at the responsibility-selection point, keeps it gated on a usable catalog, opens it in a separate tab so the creation draft remains mounted, and records the new-agent surface event. The focused tests cover presence, empty/error suppression, target/opener safety, analytics, and draft retention. I found no blocking issues.

PR #2207, opened 2026-08-06
fix(cli): verify BYO Context readiness before setup

Gandy2025:
## Summary

- gate `first-tree context enable --plan` on the selected Team's live member-safe Context Tree binding
- reuse one canonical exact root `SCOPE.md` reader for setup and task routing
- validate regular-file mode, bounded blob size, UTF-8, and the current SCOPE schema before setup choices
- return stable, sanitized CLI errors for authentication, binding, missing/invalid SCOPE, and fetch failures
- prove plan and apply failures occur before Plugin/grant mutation or session candidate receipt issuance
- document that the local preflight reads only root `SCOPE.md`, not the full Tree

## Why

The Web setup flow intentionally uses coarse binding eligibility. A Team with a bound Tree but a missing or invalid root `SCOPE.md` could therefore reach CLI setup choices and mutate local activation state before task routing discovered that the Tree was unusable.

The CLI plan is the correct local preflight seam. Apply already rebuilds that plan, so the same check also protects the mutation path without adding Server state, endpoints, or Web polling.

## Design

- `exact-root-scope.ts` owns the Git fetch, exact commit/root blob lookup, bounded size check, UTF-8 decode, schema parse, cleanup, and typed `missing | invalid | fetch` failures.
- `context-enable-scope-preflight.ts` owns the member binding read and maps binding/root failures to the stable `CONTEXT_ENABLE_*` CLI contract.
- `enable.ts` remains thin: it invokes the preflight after live Team activation and renders typed errors.

## Impact

BYO Context setup now fails early with actionable errors until the selected Team's exact root `SCOPE.md` is readable and valid. Oversized SCOPE content is reported as invalid rather than as a repository/network failure. Web eligibility behavior, provider selection, onboarding completion, write confirmation, and the wider architecture are unchanged.

## Validation

- `pnpm --filter ./apps/cli exec vitest run src/__tests__/exact-root-scope.test.ts src/__tests__/context-enable-scope-preflight.test.ts src/__tests__/context-enable-command.test.ts src/__tests__/context-route.test.ts` (44 tests passed)
- packaged-CLI focused-local E2E: valid, missing, nonregular, invalid, non-UTF-8, oversized, binding-invalid/unreadable, auth, fetch, and plan/apply recheck paths passed; failure paths produced no Plugin, grant, BYO repo, or session receipt mutation
- `pnpm check`
- `pnpm typecheck`
[...]

PR #2250, opened 2026-08-07
ACE-34: stop duplicate App task replies for one entity

Gandy2025:
## Summary

- **What changed?** App-authored GitHub task replies now take a cross-run publication claim before writing to GitHub, and a run that loses the claim reads the hidden `first-tree-github-task-reply-run` marker back off the entity instead of publishing a duplicate.
- **Why does it matter?** On PR #2207, five accepted webhook events on one unchanged head (`32e5d2d4`) dispatched five runs, and each published a byte-identical Context Reviewer verdict (comments `5201858908`, `5201860954`, `5201861440`, `5201861803`, `5201862152`, 07:45:03–07:45:25Z). Publication idempotency lived on the run's own message row (`githubTaskReplySubmission`), so it made one run idempotent and was blind to a sibling run. The run marker was written into every body and only ever read during *same-run* reconciliation of an unknown write — never as a publish guard.

### How the claim works

The publisher already computes `payloadHash = sha256([repository, entityType, entityNumber, body])`. That hash is run-independent by construction: it identifies the comment about to be written, not the run writing it. `claimGithubTaskReplyPublication` claims it through `processed_events`' existing `(event_id, platform)` unique index under a dedicated `github-task-reply-publication` platform, inside the same transaction that claims the run and *before* the run claim, so a lost publication claim leaves the run untouched and retryable.

- **Winner** publishes exactly as before.
- **Loser** never POSTs. It lists the App's task replies on the entity and matches on the caller-authored body with the run marker stripped. A match makes the loss terminal (`state: failed`, code `GITHUB_TASK_REPLY_DUPLICATE_PUBLICATION`); no match reports the owner as still in flight and leaves the run `pending`.
- **Definitive GitHub rejection** releases the claim so the reply is publishable again. An **unknown** write deliberately keeps it — the existing fail-closed reconciliation owns that case, and reopening the claim would authorize the duplicate this guards.

A moved head produces a different verdict body, so it hashes differently and still publishes. Ordinary conversational replies are unaffected for the same reason.

### Known limitation (deliberate, please read)

The guarantee is **one App comment per exact `(entity, body)`**, not per `(entity, head)`. A task reply run carries no head SHA — `issue_comment` payloads do not contain one, and the reply publisher is the generic surface used for conversational answers as well as review verdicts, so a head-scoped suppression would swallow a legitimate second answer at the same head. In the incident the five verdicts were byte-identical, so this collapses them to one; if a future reviewer run renders materially different prose for the same head, that different verdict still publishes.

Also **not addressed here**: the five *runs* themselves. `resolveGithubAudience` mints one automatic provider task per accepted event, and `createGithubTaskRun` mints a fresh `githubTaskRunId` per delivery, so the wasted agent runs still happen — this PR only stops them reaching GitHub. Collapsing dispatch needs a product decision about which event kinds may be suppressed while a run is in flight (a human question arriving mid-review must not be swallowed); filed as follow-up.

## Validation

- [x] `pnpm check` — no new findings (17 warnings / 7 infos, identical to base)
- [x] `pnpm typecheck`
- [x] `pnpm test` — server 3129/3129, shared 73 files, CLI 114/114. `packages/client` `pi-handler.test.ts` fails on this machine with and without these changes (timing-sensitive, pre-existing).

Five new tests in `github-task-reply-publisher.test.ts` cover: sibling run suppressed with one comment left on the entity, moved head still publishing, **two runs for the same exact head publishing concurrently claiming once**, claim released on definitive rejection, and claim retained after an unknown write. All five pass; the three suppression tests fail when the claim is stubbed out.

## Change Surface

## Notes

- package or install behavior changes: none. No schema change — the claim reuses the existing `processed_events` table, so there is no migration.
- docs or tests updated to match: product tests above, plus a cross-run publication observation added to `packages/qa/cases/cross-surface/github-webhook-routing-regression.md`.
- follow-up work: collapse duplicate *dispatch* for one entity so a repeat event does not bill a full agent run; decide whether a legitimate re-review should update the prior comment rather than append a new one (issue expectation 3). The Context Review PR-review publisher (`context-reviewer-publisher.ts`) has the same run-scoped-only shape for `createPullRequestReview` and is untouched here.

ACE-34

### DIFF
--- a/packages/qa/cases/cross-surface/github-webhook-routing-regression.md
+++ b/packages/qa/cases/cross-surface/github-webhook-routing-regression.md
@@ -85,6 +85,11 @@ App login or adding a GitHub-specific post-delivery branch.
   Simulate an unknown GitHub write, then confirm retry reconciles the App actor, hidden run marker, and exact body before
   returning rather than blindly creating another comment. Zero matches, duplicate exact matches, and list failure must
   all remain unknown with no second POST.
+- Deliver several supported events on one pull request without moving its head, so more than one run is dispatched for
+  the same entity, and let each run reach its terminal reply. Confirm the entity ends with exactly one App-authored
+  comment for a given reply body: the first run publishes, and every later run reports
+  `GITHUB_TASK_REPLY_DUPLICATE_PUBLICATION` without a second GitHub write. Push a new head, let the reviewer render a
+  verdict against it, and confirm that changed reply still publishes.
 - Attempt publication from a different Agent, client, runtime session, chat, repository-scoped role assignment, inactive
   membership, and spoofed user-authored `githubTask*` metadata. Confirm each fails before GitHub mutation. Confirm a reply
   that mentions the App is rejected. Historical boolean/no-run markers remain renderable but cannot publish.
@@ -109,8 +114,8 @@ cards do not invent personnel context, and webhook content is never treated as a
 task capability and host GitHub permissions.
 
 `FAIL`: a reproducible regression in authentication, tenant resolution, followed-chat/card delivery, automatic task or
-wake routing, recipient-bound publication, self-output suppression, whole-request deduplication, or Context Reviewer
-claim coverage.
+wake routing, recipient-bound publication, self-output suppression, whole-request deduplication, cross-run publication
+claim coverage, or Context Reviewer claim coverage.
 
 `BLOCKED`: the isolated run cell cannot provision a disposable App/installation, webhook credential, bound entity, or
 connected runtime needed by the selected observations.
--- a/packages/server/src/__tests__/github-task-reply-publisher.test.ts
+++ b/packages/server/src/__tests__/github-task-reply-publisher.test.ts
@@ -550,6 +550,106 @@ describe("GitHub App task reply publisher", () => {
       }),
     ).rejects.toThrow(/reserved for server-authored GitHub task reply runs/);
   });
+
+  // Regression: five webhook events on one pull request at one unchanged head
+  // dispatched five runs, and each published its own byte-identical Context
+  // Reviewer verdict. Run-scoped idempotency could not see a sibling run, so
+  // the guarantee has to be cross-run.
+  describe("cross-run publication claim", () => {
+    const VERDICT_AT_HEAD = "**Recommendation: request changes**\n\nReviewed exact head `32e5d2d4`.";
+
+    it("lets one run publish the verdict and stops a sibling run repeating it on the same entity", async () => {
+      const fixture = await createRunFixture(getApp());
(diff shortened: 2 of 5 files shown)


yuezengwu (approved):
Reviewed the cross-run publication claim and loser reconciliation paths. The implementation takes the payload-bound claim before mutating the run, retains it for unknown writes, releases it only for a definitive rejection, and covers sequential, concurrent, rejection, and unknown-write cases. I found no blocking issue.

Database note for human review: this introduces a new persistence semantic, although no schema or migration changes. Each distinct exact reply retains a github-task-reply-publication row in processed_events after success or an unknown write; only a definitive rejection removes it.

baixiaohang (changes requested):
- Rationale: The permanent entity/body claim suppresses distinct future provider tasks that happen to render the same text, not only sibling runs for one logical review.

**Risk level**: B-low
- Path baseline: `packages/server/**` without a core chat/message/session/agent service or schema change -> B-low
- Semantic lift: none

**PR summary**
- Problem: Repeated webhook events on one unchanged pull request can dispatch sibling runs that publish the same App-authored terminal reply several times.
- Approach: Claim a run-independent hash of repository, entity, and reply body in `processed_events`; only the winner writes to GitHub, while losers reconcile against existing App comments.
- Impacted modules: GitHub task-reply publisher and claim service, shared reply error schema, server regression tests, and the cross-surface QA case.

**Review findings**
❌ 1. The key has no logical task/review generation, and a successful row is never released. After one task replies `Done.` on an Issue, any later independent supported event on that same Issue whose legitimate terminal answer is also `Done.` loses this permanent claim, matches the old comment, and is marked failed. That contradicts the current per-event terminal-reply contract and the PR's statement that ordinary conversational replies are unaffected. Scope the claim to the duplicate-dispatch cohort (or to a surface that supplies a stable review generation), and add a regression showing that two independent events on one entity may publish identical bodies while sibling runs for one event still collapse. [R3, R5 / `packages/server/src/services/github-task-reply-publication-claim.ts:39`]

⚠️ 2. Reusing `processed_events` also gives the webhook-dedup table a second, permanent retention contract: every successful or unknown exact reply remains there, while only definitive rejection deletes it. The existing system context already tracks crash-stranded `processed_events` claims as a P1 lifecycle gap; a later cleanup of webhook dedup rows could silently remove publication claims and re-enable duplicates unless this platform is specially preserved. Please give publication claims an explicit independent retention/lifecycle boundary, or encode and document the platform-specific retention requirement where cleanup is owned. [R1, R5 / `packages/server/src/services/github-task-reply-publication-claim.ts:32`]

**Action taken**
- Submitted request changes.

PR #2195, opened 2026-08-05
Unify computer and runtime selection for new agents

Gandy2025:
## Summary

- unify Team → New Agent and onboarding around one `This agent will run` section
- keep the one-computer path read-only, require an explicit compact dropdown choice when several computers are online, and reveal runtime options only for the chosen computer
- preserve a still-valid manual runtime choice while reapplying the shared Codex → Claude Code → remaining-provider preference after automatic selections or computer changes
- rename and move the access section to `Who can use it?`, without changing visibility behavior
- add a production-component preview state and a versioned Momentic test for the multi-computer path

## Validation

- `pnpm --filter @first-tree/web test` — 245 files / 2,211 tests passed
- `pnpm --filter @first-tree/web typecheck`
- design-token guardrails
- `pnpm --filter @first-tree/web build`
- `npx momentic@3.42.0 lint`
- uploaded Momentic E2E run: https://app.momentic.ai/run-groups/11c7f6cd-2867-4ccf-8c37-c027b3a4e103
- exact-head browser verification at desktop and narrow product width; no horizontal overflow
- two independent reviews completed with no remaining blocking or judgement-call findings

## Risk

No data model, permission, provider-auth, runtime-switch, or onboarding step-sequence changes. The existing zero- and one-computer paths remain intact; the behavior change is limited to making multi-computer agent creation explicit.

PR #2253, opened 2026-08-07
fix: refresh client chat fixtures and test cache

baixiaohang:
## Summary

- update workspace chat SDK fixtures for the required `priorityRows` response shape
- make Turbo test tasks depend on upstream package builds so shared schema changes invalidate downstream test caches

## Validation

- `pnpm exec biome check packages/client/src/__tests__/sdk-workspace-chats.test.ts turbo.json`
- `pnpm --filter @first-tree/client exec vitest run src/__tests__/sdk-workspace-chats.test.ts --maxWorkers=2` (8 passed)
- `pnpm exec turbo run test --filter=@first-tree/client --dry=json` (confirmed `@first-tree/client#test` depends on `@first-tree/shared#build`)

### DIFF
--- a/packages/client/src/__tests__/provider-boundary-guard.test.ts
+++ b/packages/client/src/__tests__/provider-boundary-guard.test.ts
@@ -576,7 +576,7 @@ describe("runtime provider architecture guard", () => {
     for (const rel of CATALOG_CONSUMER_FILES) {
       const source = readFileSync(join(repoRoot, rel), "utf8");
       if (rel.endsWith("new-agent-dialog.tsx")) {
-        expect(source).toContain("pickPreferredRuntimeProvider");
+        expect(source).toMatch(/\b(?:pickPreferredRuntimeProvider|resolveRuntimeSelection)\b/);
         expect(source).toContain("enabledOkRuntimeProviders");
         expect(source).toContain("runtimeProviderLabel");
         expect(source).toContain("PREFERRED_RUNTIME_PROVIDER");
--- a/packages/client/src/__tests__/provider-boundary-guard.test.ts
+++ b/packages/client/src/__tests__/provider-boundary-guard.test.ts
@@ -587,7 +587,7 @@ describe("runtime provider architecture guard", () => {
         expect(source).not.toMatch(/function asRuntimeProvider/);
       }
       if (rel.endsWith("use-computer-connection.ts")) {
-        expect(source).toContain("pickPreferredRuntimeProvider");
+        expect(source).toMatch(/\b(?:pickPreferredRuntimeProvider|resolveRuntimeSelection)\b/);
         expect(source).toContain("enabledOkRuntimeProviders");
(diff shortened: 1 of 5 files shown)


yuezengwu (changes requested):
Requesting one in-scope documentation fix before approval.

`turbo.json` now deliberately restores `test -> ^build`, but `scripts/vitest-aliases.ts:4-10` still says the source aliases exist so that exact dependency—and its 5–15 second cold-CI build cost—can be removed entirely. After this change, that rationale is false and gives a future maintainer a strong reason to remove the dependency and recreate the stale-cache bug.

Please update that comment in this PR to describe the current split accurately: Vitest still resolves workspace packages from source, while the Turbo dependency remains necessary to include upstream package changes in downstream test cache keys (and therefore accepts the upstream build cost).

The fixture updates and the dependency direction otherwise look correct. This PR does not change a core data structure or the database.

baixiaohang:
Updated `scripts/vitest-aliases.ts` in 8783d0608 to document the current split: Vitest resolves workspace imports from source, while `test -> ^build` remains intentional so upstream task hashes invalidate downstream test caches. The comment now explicitly acknowledges the cold-CI build cost.

Validation: `pnpm exec biome check scripts/vitest-aliases.ts` passed.

yuezengwu (approved):
Re-reviewed the updated head. The requested documentation fix is fully addressed: the comment now accurately distinguishes Vitest's source resolution from Turbo's upstream task-hash dependency and explicitly records the accepted cold-CI build cost.

The complete diff looks correct. Approving without rerunning tests, consistent with the review scope.

baixiaohang:
The latest failure was deterministic merge-state drift, not a flaky retry: main merged #2195 after this branch was created, replacing direct `pickPreferredRuntimeProvider` use with the centralized `resolveRuntimeSelection` helper while the cross-package guard still required the old symbol.

This head now:

- accepts either centralized runtime-selection seam in the guard, covering both the branch and current-main merge state
- adds `packages/client/turbo.json` so Web source files read by the Client boundary guard participate in the Client test cache key

Focused validation against an overlay on current main `e346285d1`:

- `provider-boundary-guard.test.ts` + `sdk-workspace-chats.test.ts`: 21 passed
- Turbo dry-run includes `packages/web/src/components/new-agent-dialog.tsx` in `@first-tree/client#test` inputs and retains `@first-tree/shared#build`
- Biome check passed for the changed guard/config files

yuezengwu (approved):
Re-reviewed the latest head after the additional cache and merge-state fixes.

The package-level Web input is warranted because the Client provider-boundary test recursively reads `packages/web/src`; the reported Turbo dry-run also confirms that the inherited `@first-tree/shared#build` dependency remains intact. Allowing either centralized runtime-selection helper correctly supports both the branch and current-main merge states without permitting provider-specific selection logic.

No blocking issues found. Approving without rerunning tests, consistent with the review scope.

Issue #2256, opened 2026-08-07
[Bug] github follow fails in agent-only chat until supervising human joins


baixiaohang:
## Summary

`first-tree github follow` fails in an agent-created chat while the supervising human is still a watcher, then succeeds immediately after that human joins as a speaker. The CLI reports the failure as an entity being followed elsewhere even though no attention line exists elsewhere.

## Surface Area

GitHub integration, Server SCM attention-line authority checks, Client SDK / CLI error reporting.

## Environment

- version: `0.5.20-staging.1121.1`
- runtime: `codex`
- operating system: `Linux 6.6.87.1-microsoft-standard-WSL2 x86_64`

## Reproduction

1. Start an agent-created group chat whose speakers are non-human agents. Let the supervising human remain present only as the automatically maintained watcher.
2. Create a GitHub pull request in a repository covered by the Team's GitHub App installation.
3. From an agent speaker in that chat, run `first-tree-staging github follow <pull-request-url>`.
4. Retry with `--rebind`.
5. Have the supervising human join the chat as a speaker through the Web `workspace-join` path.
6. Run the same ordinary follow command again.

## Expected Behavior

The first follow should create the attention line because the agent-side preflight already recognizes the supervising human watcher as the representative human. If authority is not valid, the CLI should report that authority error accurately. `--rebind` guidance should appear only for a real structured `ENTITY_FOLLOWED_ELSEWHERE` conflict.

## Actual Behavior

Both the initial follow and the `--rebind` retry fail with:

```json
{"ok":false,"error":{"code":"HTTP_409","message":"Entity already followed in another chat (malformed conflict body)"}}
```

After the supervising human joins as a speaker, the same ordinary command succeeds:

```json
{"ok":true,"data":{"status":"created","entity":{"entityType":"pull_request","entityKey":"agent-team-foundation/first-tree#2255","boundVia":"agent_declared"}}}
```

The final status is `created`, not `rebound` or `already_following`, confirming that the earlier 409 did not represent an attention line in another chat.

## First Tree Context

- Chat ID: `904e58c0-9f1b-4834-b49b-1f595377eb17`
- Reported by (user): `baixiaohang`
- Agent: `codex-developer` (`019e71d2-c9ec-7f11-86bf-5dfc9e873338`)

## Suggested Cause and Fix

[...]

PR #2258, opened 2026-08-10
refactor(server): organize SCM services by provider

baixiaohang:
## Summary

- move the 9 provider-neutral SCM services into `services/scm/shared`
- move the 16 GitHub and 8 GitLab services into `services/scm/github` and `services/scm/gitlab`
- remove redundant filename prefixes and update every static, type-only, dynamic, and mocked module path
- keep the migration replace-only: no compatibility shims, barrels, re-exports, or behavior changes

This is a path/import-only refactor. It does not change exports, APIs, database state, configuration, wire contracts, or runtime behavior. Issue #2256 remains out of scope.

## Validation

- `pnpm exec biome check packages/server` (passes with pre-existing warnings only)
- Biome check over all changed files
- full Server suite: 275 test files and 3,202 tests passed with at most 2 workers
- path and dependency-direction audits: 33 renames, no old service files or import/mock paths, no provider imports from `scm/shared`, and no cross-provider imports
- `git diff --check`

The repository-wide `pnpm check` still reports pre-existing lint failures in unrelated Client, CLI, and OIDC files; all files changed by this PR pass Biome.


yuezengwu (approved):
Reviewed the path/import-only refactor. The 33 SCM services are consistently split into shared, GitHub, and GitLab directories. I found no blocking issues: old import and mock paths are gone, shared does not import either provider, and there are no cross-provider imports. The database schema content is unchanged; the touched schema file only updates a documentation reference.

Non-blocking cleanup: a few comments still name removed files, including services/github-app-token.ts in packages/shared, services/github-normalize.ts in packages/web, and same-directory references such as github-audience.ts, github-entity-follow.ts, github-oauth.ts, and github-entity-chat.ts. Updating those references would make the migration fully navigable, but it does not affect this approval.

I did not run local tests per review scope. The current head is 4c63279c21b04ef832fc3bc1c29b6c1f8e80d8cd, and all reported CI checks are green.

PR #2262, opened 2026-08-10
fix(scm): stabilize watcher follow authority

### LINKED ISSUE #2256: [Bug] github follow fails in agent-only chat until supervising human joins

baixiaohang:
## Summary

- make agent-issued SCM follow resolution use the same watcher-inclusive authority snapshot before and inside the transaction, while keeping the wake target constrained to an active speaker
- serialize watcher projection writers, personnel placement, and manager/member lifecycle mutations with one projection-to-membership lock order, including GitLab's provider-specific identity revalidation path
- fence existing Context Reviewer chats before authority rows so reuse/repair cannot deadlock with SCM membership snapshots
- classify only structured `ENTITY_FOLLOWED_ELSEWHERE` responses as rebind conflicts in the Client SDK
- add GitHub/GitLab routing and dual-connection concurrency regressions for stale authority, chat-set expansion, and deadlock-prone lock ordering
- stabilize the Pi retry-exhaustion test by allowing its first synthetic version process realistic CI cold-start headroom without changing production timeouts

## Validation

- targeted Client and Server Vitest suites: 260 tests passed
- Biome check for all 16 changed files
- `git diff --check`

Fixes #2256


yuezengwu (changes requested):
Blocking concurrency issue:

`addChatParticipants` now acquires `watcher_projection_agent` before the chat fence, but some existing callers enter it after taking member/agent row locks. In particular, `withContextReviewerDispatchAuthority` locks the reviewer's manager member row and reviewer agent row `FOR UPDATE`, then its GitHub/GitLab callbacks call `applyMembershipWrite` or `createChat`, which reaches this new lock.

A concurrent `updateAgent(reviewer, { managerId: ... })` takes the opposite order: it acquires the reviewer projection lock first and later waits for the reviewer agent row. The dispatch transaction holds that agent row and waits here for the projection lock, producing a PostgreSQL deadlock and aborting automatic review chat creation or membership repair.

Please make these call paths acquire the projection locks before the authority member/agent rows (or otherwise avoid taking this lock unconditionally after callers may already hold those rows), and add a concurrency regression covering Context Reviewer dispatch racing a reviewer manager transfer.

I did not run tests, per the review instructions.

yuezengwu (approved):
Re-reviewed the synchronized head. The previous blocker is resolved:

- Context Reviewer dispatch now acquires the reviewer/human projection locks before authority row locks.
- Existing reviewer chats are fenced before those authority rows, so reuse/repair cannot invert against an SCM membership snapshot.
- The new concurrency regressions cover both fresh-chat manager transfer and reused-chat follow-snapshot races.

I found no further blocking issues. The Pi change is test-only cold-start headroom and does not alter production timeouts.

I did not run tests, per the review instructions.

PR #2265, opened 2026-08-10
test(server): isolate cron scheduler in test apps

baixiaohang:
## Summary

- keep the production cron scheduler enabled by default while adding an internal background-task assembly option for deterministic tests
- disable the periodic cron scheduler in `createTestApp` by default so explicit cron sweeps cannot race the next test's database reset
- cover both the production-default start/stop path and the test-only disabled path

## Root cause

After #2262 merged, the Server CI run exposed an existing timing race in `cron-jobs.integration.test.ts`: the test App's periodic scheduler could claim a due job before the test's explicit `sweepCronJobs` call. The explicit sweep then returned no work because of `SKIP LOCKED`, and the still-running background transaction raced the next test's full-table `TRUNCATE`, producing a PostgreSQL deadlock.

This change keeps production startup and shutdown semantics unchanged. Tests that need scheduler lifecycle behavior can still opt in or instantiate `createCronScheduler` directly.

## Validation

- `pnpm --filter @first-tree/server exec vitest run src/__tests__/background-tasks-extra.test.ts src/__tests__/cron-jobs.integration.test.ts` (37 tests passed)
- Biome check for all 4 changed files
- `git diff --check`

Follow-up to [main CI run #31362118653](https://github.com/agent-team-foundation/first-tree/actions/runs/31362118653).

### DIFF
--- a/packages/server/src/__tests__/background-tasks-extra.test.ts
+++ b/packages/server/src/__tests__/background-tasks-extra.test.ts
@@ -158,4 +158,17 @@ describe("createBackgroundTasks", () => {
 
     await tasks.stop();
   });
+
+  it("allows tests to disable the periodic cron scheduler", async () => {
+    const { createBackgroundTasks } = await import("../services/background-tasks.js");
+    const tasks = createBackgroundTasks(makeApp(), "srv_3", { cronSchedulerEnabled: false });
+
+    tasks.start();
+    await vi.advanceTimersByTimeAsync(60_000);
+    await tasks.stop();
+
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed the production and test paths at 23d7d53cf2fb8a2633c8280a27693ffbbd947c0d. The change keeps the default buildApp/createBackgroundTasks behavior enabled, while createTestApp explicitly opts out unless a test requests scheduler lifecycle behavior. The start/stop coverage for the production default remains in place, and the added disabled-path assertion covers the isolation hook. No blocking issues found.

PR #2217, opened 2026-08-06
fix: stabilize Claude Context repair adoption

baixiaohang:
## Summary

- make the Claude Plugin payload deterministic for one adapter version and bump only Claude to adapter `1.0.2`; Codex remains at `1.0.1`
- replace same-session reload receipts with a durable next-session obligation consumed only by a fresh Claude `startup`
- omit the persistent current-session handoff while Claude next-session adoption is pending; setup completes with an explicit new-session action instead of returning a handoff that routing would reject
- keep automatically synchronized old Claude sessions usable through account/channel/session/old-digest/exact-target compatibility facts written when each action is issued; concurrent issuance, provider rollback, global Plugin updates, and interrupted syncs remain recoverable
- retain a TTL-bound terminal action backup so replaying the exact Claude sync challenge returns the same successful result after lost output, while pending next-session adoption consistently requires a new Claude session
- fail closed only for a real account-lock conflict (`EEXIST`); non-busy lock I/O errors retain their original classification and known-good adapter fallback
- treat missing or invalid `CLAUDE_PROJECT_DIR` as pathless so global Claude App activation works without falling back to Hook cwd
- update CLI docs and deterministic QA contracts for repair, lifecycle, concurrency, crash recovery, and pathless routing

## Compatibility

- existing global/directory/session grants and schema v3 behavior are unchanged
- legacy install manifests with `adoptionGeneration` remain readable
- legacy reload state is retired during deterministic installation and account cleanup
- compatible-session facts are Claude-only, fail closed unless the exact target payload is healthy and no next-session obligation is pending, and are cleared on provider uninstall, account switch, and purge logout

## Verification

After the request to stop full-suite runs, only targeted verification was run:

- CI-failure regression: `context-activate-command.test.ts`, 11 tests passed with `--maxWorkers=2`
- lock-classification set: 2 files, 9 tests passed with `--maxWorkers=2`
- rollback/lock-contention set: 4 files, 39 tests passed with `--maxWorkers=2`
- obligation/replay set: 3 files, 22 tests passed with `--maxWorkers=2`
- lifecycle/recovery set: 4 files, 67 tests passed with `--maxWorkers=2`
- setup/sync/activation set: 3 files, 38 tests passed with `--maxWorkers=2`
- `apps/cli` typecheck
- affected TypeScript files: Biome check
- `git diff --check`

[...]

PR #2267, opened 2026-08-10
fix(web): align BYO setup prompt with the new-session adoption boundary

Gandy2025:
## Problem

[#2217](https://github.com/agent-team-foundation/first-tree/pull/2217) retired the in-session reload receipt but did not touch `packages/web`, so the artifact users copy from Settings and onboarding still teaches the removed mechanism.

Two defects follow, with different blast radius:

**1. No branch for the actual terminal state (deterministic).** For persistent Claude setup, `applyPersistent` writes a next-session obligation whenever `installPlan.operation !== "unchanged"` — i.e. on every install or repair — which forces `currentSessionHandoff: null` while `setup.complete` stays `true`. Step 7 assumed a handoff always exists and required failing closed on invalid loader output, and step 6's fallback was gated on setup being *incomplete*. An agent following the prompt literally had no valid path in the state that every first-time Claude setup lands in.

**2. The only named Claude adoption action was the dead one (misdirection).** The prompt referenced `/reload-plugins` in four places, including a consent boundary, while the CLI now returns `Start a new Claude session for persistent auto-activation`. That string appeared nowhere in the prompt. `/reload-plugins` cannot satisfy the obligation: the hook matcher is `startup|resume|clear|compact`, and `consumeContextAdapterNextSessionObligation` requires `sessionStartSource === "startup"`. A user who runs it performs a no-op, and an agent that then reports readiness leaves Context silently inactive.

Observed in a real session: setup returned `plugin: "repair"`, `setup.complete: true`, `currentSessionHandoff: null`, next action `Start a new Claude session` — and the assisting agent asked the user to run `/reload-plugins` instead.

Not affected: Codex (the obligation is Claude-only and its Hook branch was already correct) and Claude session-only scope (`applySessionOnly` always returns a handoff).

## Change

`packages/web/src/lib/byo-setup-prompt.ts` only — no behavior change outside the copied artifact:

- Step 4: apply command runs unchanged, with no permitted additions.
- Step 5 / consent boundary: adapter updates are adopted in the next Claude session; `/reload-plugins` removed.
- Step 6: follow `nextActions` whether or not setup reports itself complete; relay the new-session action verbatim, never substitute an in-session reload, and do not claim Context is active in this session.
- Step 7: a complete setup carrying no handoff is a normal terminal state, not invalid output. The schema-v3 adoption and loader-reuse contract is unchanged for the handoff case.
- Closing line states whether Context is active now or starts next session, instead of asserting readiness unconditionally.

## Tests

Replaced the assertion that pinned the retired contract with assertions on the new-session boundary and the no-handoff terminal state, plus negative guards so `/reload-plugins` and `--reload-receipt` cannot return.

`pnpm --filter @first-tree/web test` — 2327 passed. `pnpm --filter @first-tree/web run typecheck` and `pnpm check` clean.

## Notes

This aligns prose to a contract that already landed in #2217; it does not introduce or change a durable state model, so no companion Context Tree PR is proposed. Flagging it explicitly in case review disagrees.

The `.in_use` marker that Claude Code writes into its plugin cache makes the payload integrity check fail, which classifies every Claude apply as `repair` and therefore keeps the no-handoff path permanently reachable. That is a separate defect and is not addressed here.


yuezengwu (approved):
审阅通过。

实现目标和核心改动：Web 端复制的 BYO setup prompt 现在与已落地的 Claude next-session adoption 契约一致；它移除了已退役的 `/reload-plugins` / reload receipt 路径，要求原样执行 apply command，并把 `setup.complete: true` 且 `currentSessionHandoff: null` 识别为“配置完成、下次全新 Claude 会话自动激活”的正常终态。现有 schema-v3 handoff、loader 重读和 digest 复用约束保持不变。

我对照了 CLI 的 `nextActions`、next-session obligation 消费条件以及 Context Tree 中的 External BYO Context 决策；没有发现 blocker，也没有核心数据结构或数据库变更。测试增加了新会话边界、无 handoff 终态及旧 reload 机制不得回归的断言。

按本次 review 范围未运行测试。

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The copied setup artifact now matches the CLI's durable next-session adoption contract and handles the complete-without-handoff state without weakening the existing handoff checks.

**Risk level**: C
- Semantic lift: none; this aligns a rendered prompt and its regression assertions with an already-landed CLI lifecycle contract.

**PR summary**
- Problem: Members performing persistent Claude Context setup were being directed to a retired in-session reload flow, while the deterministic first-install/repair result completes without a current-session handoff and requires a fresh Claude startup.
- Approach: The copied Settings/onboarding prompt now preserves the exact apply command, follows `nextActions` even after setup completes, treats a null handoff as a valid next-session terminal state, and reports whether Context is active now or begins next session. Regression guards prevent the retired reload command and receipt flag from returning.
- Impacted modules: Web BYO setup prompt renderer and its Context enablement test.

**Review findings**
✅ 1. The prompt now distinguishes the normal complete-without-handoff Claude outcome from invalid handoff output while retaining schema-v3 and loader-reuse validation when a handoff exists.
✅ 2. The new wording matches the CLI's startup-only adoption obligation and does not affect Codex or Claude session-only activation.

**Action taken**
- Approved. Review was static only; no tests were run by the reviewer.

PR #2269, opened 2026-08-10
ACE-40 fix(web): revert the New Agent dialog template library entry

Gandy2025:
## Summary

- Reverts #2247 (`7d9d05538`), which added a "Browse the template library" link to the New Agent dialog's Responsibilities section.
- The New Agent dialog already carries the intended in-app template entry: the **"Choose a template" picker** added in #2130. It reads the same public catalog and lets the user adopt a Template without leaving the dialog. A second, outbound link to `/templates` at the same decision point is redundant — it competes with the picker instead of serving it.
- Reverting restores the dialog to its pre-#2247 shape: `-58` lines, exactly the two files #2247 touched.

The public `/templates` route and the canonical `/templates/:slug?use=1` intent path are unaffected — neither was introduced or modified by #2247.

Requested by the issue owner on ACE-40 after reviewing the placement.

## Validation

- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm test`

`@first-tree/web` is fully green (248 files / 2326 tests), including `new-agent-dialog-templates.test.tsx` (9 tests) which returns to its pre-#2247 assertions.

Two unrelated packages flaked under the full parallel `pnpm test` on this machine, and both pass when run on their own — neither can be affected by a `packages/web`-only revert:

- `@first-tree/skill-evals` — all 621 tests passed; the run failed on a `[vitest-worker]: Timeout calling "onTaskUpdate"` RPC timeout under load.
- `@first-tree/client` — `capability-probes.test.ts` hit the 5s timeout while launch-verifying a real runtime binary; passes standalone (53 tests).

`@first-tree/server` also passes standalone (275 files / 3211 tests).

## Change Surface

## Notes

- docs or tests updated to match: the test file is reverted alongside the component, so the "link is absent" assertions and the link-behavior test go away together
- follow-up work: the onboarding "create your first agent" step (`step-create-agent.tsx`) still has no template picker at all — a Template only appears there when arriving via `?use=1`. Users on the normal signup path never see Templates while creating their first agent. Out of scope here; worth its own issue if we want to close that gap.

### DIFF
--- a/packages/web/src/components/__tests__/new-agent-dialog-templates.test.tsx
+++ b/packages/web/src/components/__tests__/new-agent-dialog-templates.test.tsx
@@ -182,10 +182,6 @@ function buttonByText(text: string): HTMLButtonElement {
   return button;
 }
 
-function templateLibraryLink(): HTMLAnchorElement | null {
-  return document.body.querySelector<HTMLAnchorElement>('a[href="/templates"]');
-}
-
 function optionCardByText(text: string): HTMLElement {
   const label = [...document.body.querySelectorAll("label")].find((el) => el.textContent?.includes(text));
   if (!label) throw new Error(`Missing option card "${text}". Body: ${document.body.textContent ?? ""}`);
@@ -245,8 +241,6 @@ describe("NewAgentDialog template responsibilities", () => {
     await renderDialog(onCreated);
     await flush();
     expect(document.body.textContent).not.toContain("Responsibilities");
-    // No entry to an empty library either — the link is gated on the same catalog.
-    expect(templateLibraryLink()).toBeNull();
 
     await fillNameAndOpenPickerSafe();
     async function fillNameAndOpenPickerSafe() {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed the goal and change surface: this removes the redundant outbound Template Library link from the New Agent dialog while preserving the existing in-dialog “Choose a template” flow and the public `/templates` routes.

The two-file patch is the exact inverse of #2247, including only the link-specific assertions. The picker, template submission behavior, and their existing coverage remain intact. No blockers found.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The revert removes a redundant outbound discovery path while preserving the in-dialog Template picker and every canonical public Template route.

**Risk level**: C
- Semantic lift: none; the change removes local presentation/navigation and its analytics assertion without touching subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Users choosing an Agent responsibility saw both an in-dialog Template picker and a second link to browse the same catalog in another tab, creating competing actions at one decision point.
- Approach: Revert the extra library link and its focused tests so creation remains single-path in the dialog; the public `/templates` library and `/templates/:slug?use=1` adoption intent remain available for external discovery and direct intent handoff.
- Impacted modules: New Agent dialog and its Template-responsibility tests.

**Review findings**
✅ 1. The patch is the content-level inverse of #2247 across the same two files, so it does not broaden the revert beyond the redundant link and its test coverage.
✅ 2. The catalog-gated `Choose a template` flow remains intact and still supports selecting an official Template without leaving or losing the creation draft. [`packages/web/src/components/new-agent-dialog.tsx`:890]
✅ 3. The public Template library and canonical use-intent routes remain registered and unchanged. [`packages/web/src/app.tsx`:226]

**Action taken**
- Approved.

PR #2272, opened 2026-08-10
feat: count Context Tree influence from the visible impact note

Gandy2025:
## Problem

The Context tab could report how often the tree was **opened** but not what it **changed**.

`#2204` moved influence attribution from a structured `metadata.contextDecision` receipt to a prose note in the message body, so managed and BYO sessions could show the same thing. The note is the right visible form — but the Server never parsed it, so influence stopped being queryable, and the Server's validation of the cited repo/commit/path went with it.

## Approach

Agents keep writing **only** the note — `first-tree-read` is unchanged. The Server derives the structured receipt from that note at message write time.

Deriving rather than asking agents for a parallel payload is the point: a hand-maintained second copy can disagree with the sentence the reader sees, and nothing at runtime would notice.

## Changes

**shared** — one impact-note parser (`context-impact-note.ts`) used by both the Server and `skill-evals`, replacing the private parser in `first-tree-read/metrics.ts`. Conversion is strict: a note the eval rejects is never one the Server persists.

**server** — the `contextDecision` trust boundary derives from the body; a converting note overrides any supplied payload, and a present-but-unconvertible note strips one rather than storing a claim that could contradict the visible text. `summarizeContextTreeInfluence` reads `messages.metadata` in place on **both** snapshot routes.

**web** — the per-message receipt card is removed, so the note is the single visible form. The Context tab gains an agent-reported influence panel, a ranked node list, and an `Influence` feed filter.

**backfill** — `src/services/context-tree/influence-backfill.ts` plus a thin CLI derives receipts for notes written between `#2204` and this PR. Dry run by default; `--since`/`--until` required and printed before apply; `--org` or an explicit `--all-orgs`; keyset pagination with a resumable cursor.

## Trust and privacy boundaries

Everything in a receipt is asserted by a private message body and never verified, which shapes both surfaces:

- **The feed** carries `summary` and `evidence`, so an event whose chat the viewer cannot open is **omitted entirely** — there is no partially-redacted row worth showing.
- **The org-wide ranking** carries only `nodePath` + `decisionCount`, and only for paths that exist in the snapshot the caller already holds. Title and source link are rebuilt from that snapshot, so no field in the ranking originates in an inaccessible chat.
- **The panel** leads with `Agent-reported`, states that First Tree does not verify causality, and names its coverage gaps, so a low count reads as incomplete rather than as a verdict on a node.
- **Truncation** past the scan cap renders as `N+` with an explicit floor notice.

## No schema change

Messages are immutable and never deleted, so they already **are** the durable statistics source — unlike `session_events`, which is why `context_tree_io_events` exists. The existing `(organization_id, …)` and `(chat_id, created_at)` indexes bound every query to one org's window.

No partial index either: the snapshot route already times each stage and warns past its budget. That instrumentation, not a guess, should decide when an index earns its migration.

## Deliberately not in this PR

- **No conversion percentage and no effect-distribution chart.** A decision is a message and a read is a file open; a ratio would read as a conversion rate while measuring different denominators.
- **BYO sessions are not counted** — their note prints in the user's own terminal and never reaches the Server. The panel says so.
- **Only `en` / `zh` notes parse.** Known gap, not a regression.

## Verification

`pnpm check` and `pnpm typecheck` clean.

| package | result |
| --- | --- |
| shared | 910 passed |
| server | 3243 passed (277 files) |
| web | 2313 passed (247 files) |
| skill-evals | 621 passed (36 files) |

Verified in the browser at `/preview/context` in light and dark: the panel renders, the `Influence` filter isolates influence rows, node titles resolve from the snapshot, and links resolve to the snapshot's repo and head commit.

The backfill has **not** been run against any database — that is a separate call.

## Review note

This changes a durable product contract. Companion tree PR: [first-tree-context#913](https://github.com/agent-team-foundation/first-tree-context/pull/913), which stays open for source-first reconciliation after this lands.


packages/shared/src/schemas/context-impact-note.ts
```diff
@@ -0,0 +1,279 @@
...
+ * Convert a parsed note into the durable receipt, or `null` when anything a
+ * receipt promises is missing.
+ *
+ * Fail-closed on the whole note rather than storing a partial one: a receipt
+ * whose evidence cannot be inspected is a claim, and the inspectable source is
+ * the only part First Tree can vouch for. The one soft failure is `heading`,
+ * which is display sugar — an over-long label is dropped so the citation
+ * itself survives.
+ */
+export function contextDecisionFromImpactNote(note: ContextImpactNote): ContextDecision | null {
```
yuezengwu, line 290: [P1] Fail closed on the parser observations before deriving a receipt. parseContextImpactNotes intentionally returns malformed notes and records atEnd, blankLineBefore, logicalLinesOk, sourceScaffoldingOk, exactLinksOk, and legacyScaffold, but this conversion ignores every one of those flags. As a result, a note followed by arbitrary text, a fourth blockquote line, or malformed scaffolding is still persisted and counted even though skill-evals rejects it. That reintroduces the Server/eval divergence this shared parser is meant to remove. Please require the current valid scaffold and placement before conversion, or separate the permissive scan result from a strict conversion result.

packages/shared/src/schemas/context-impact-note.ts
```diff
@@ -0,0 +1,279 @@
...
+  const repositoryEnd = segments[blobIndex - 1] === "-" ? blobIndex - 1 : blobIndex;
+  if (repositoryEnd <= 0) return null;
+
+  const repositoryPath = segments
+    .slice(0, repositoryEnd)
+    .join("/")
+    .replace(/\.git$/iu, "");
+  const nodePath = segments
+    .slice(blobIndex + 2)
+    .map((segment) => decodeURIComponent(segment))
```
yuezengwu: [P1] Please catch decoding failures here. new URL accepts malformed percent escapes such as %E0%A4%A, but decodeURIComponent throws URIError. This parser now runs synchronously in every agent message preflight, so one malformed impact-note URL turns the send path into a 500 instead of the documented fail-closed null. Wrap segment decoding in the same failure boundary and add the malformed-escape case.

packages/server/src/services/context-tree/influence.ts
```diff
@@ -0,0 +1,258 @@
...
+    // the headline to match a rendering failure would be the worse lie.
+    const parsed = contextDecisionSchema.safeParse(row.decision);
+    if (!parsed.success) continue;
+    recentEvents.push({
+      id: row.id,
+      agentId: row.agent_id,
+      agentName: row.agent_name ?? "unknown",
+      agentAvatarColorToken: row.agent_avatar_color_token,
+      effect: parsed.data.effect,
+      summary: parsed.data.summary,
```
yuezengwu: [P0] Do not return body-derived decision content to a viewer who cannot access the chat. summary is copied directly from the private message body, and the existing Context usage contract explicitly says only the chat topic is org-wide while chat content remains behind requireChatAccess. Here viewerCanAccess only controls the link; the Web renders summary and evidence unconditionally, so any org member can read the private decision outcome. Filter inaccessible events or redact every body-derived field server-side, with a regression test for an outsider.

packages/server/src/services/context-tree/influence.ts
```diff
@@ -0,0 +1,258 @@
...
+      SELECT
+        evidence ->> 'nodePath' AS node_path,
+        count(DISTINCT influence.id) AS decision_count,
+        (array_agg(evidence ->> 'heading' ORDER BY influence.created_at DESC))[1] AS heading,
+        (array_agg(evidence ->> 'repoUrl' ORDER BY influence.created_at DESC))[1] AS repo_url,
+        (array_agg(evidence ->> 'commit' ORDER BY influence.created_at DESC))[1] AS commit
+      FROM influence
+      CROSS JOIN LATERAL jsonb_array_elements(influence.decision -> 'evidence') AS evidence
+      WHERE evidence ->> 'nodePath' IS NOT NULL
+      GROUP BY 1
```
yuezengwu: [P1] Scope this ranking to the currently bound Context Tree repository, or key and group by canonical repository identity plus node path. Grouping only by nodePath means an org rebind, or an agent note citing another repository, merges distinct nodes that happen to share a path, counts them as one, and then attaches whichever repo and commit was cited most recently. The Context tab presents this as nodes in the selected Team bound tree, so the current result can attribute decisions to the wrong node and inflate its rank.

yuezengwu (changes requested):
Request changes.

Implementation goal and core change: use one shared parser to derive immutable metadata.contextDecision from the visible impact note at message-write time, aggregate those receipts into an organization-scoped influence summary, and expose the headline, ranked nodes, and Influence feed in the Context tab. There is no database migration, but this is a durable message-metadata and snapshot API contract change.

Blocking findings are recorded inline: strict conversion currently accepts malformed or misplaced notes that skill-evals rejects; malformed percent-encoded paths can throw from the synchronous send preflight; private chat decision summaries and evidence are exposed to org members without chat access; and node ranking is not scoped or keyed to the current bound repository, so rebinds and foreign citations can merge unrelated nodes.

The existing high-severity CodeQL polynomial-regex finding on the new message-body parser is also open and CI is failing; it must be resolved before merge.

Finally, the current decision-locked Context Tree contract says usage telemetry is not proof that a node shaped a decision and does not yet define this new influence read model. A cross-linked companion first-tree-context change must be reviewed and land before this code can merge, then this PR should be reconciled against that durable contract.

Per review policy I inspected the full diff and live PR state but did not run tests or QA.

baixiaohang (changes requested):
- Rationale: In addition to the already-filed parser, private-chat disclosure, and cross-repository ranking blockers, the aggregate does not use the canonical valid-receipt set and the UI overstates an agent-reported signal as verified product fact. This durable contract also needs the companion Context Tree PR noted by the author before merge.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes `services/chat/message.ts` and the persisted semantics of `messages.metadata.contextDecision` -> B-high.

**PR summary**
- Problem: The Context tab can show that agents opened the tree, but cannot show which decisions agents report the tree as having shaped.
- Approach: Derive the existing structured receipt from the one visible impact note at message-write time, aggregate immutable message metadata in both snapshot routes, remove the duplicate per-message receipt card, and add an influence headline, node ranking, and feed filter.
- Impacted modules: shared note/schema contracts; Server message persistence and Context snapshot aggregation; Web chat and Context surfaces; first-tree-read eval parsing.

**Review findings**
❌ 1. Derive every aggregate facet from the same fully schema-valid row set. The CTE checks only an allowed `effect` and array-typed `evidence`; for example `{effect: "constrained", evidence: []}` is counted in the headline/effect tally, produces no node, and is then dropped from the feed by `contextDecisionSchema.safeParse`. Missing version/summary or malformed evidence fields have the same split. Parse/filter the canonical receipt once (or enforce its complete shape in the shared candidate set) before totals, nodes, and feed diverge. [R1/R5 / `packages/server/src/services/context-tree/influence.ts:53`]

❌ 2. Preserve the agent-reported trust framing on the aggregate surface. “decisions shaped” and “Nodes that changed decisions” read as Server-verified facts, while this path proves only that an agent authored a syntactically convertible note; it does not verify causality or that the cited object exists. The removed receipt card explicitly disclosed that boundary, and the current Context Tree contract requires the note to remain author attribution rather than platform authentication. Carry “agent-reported” into the headline/ranking/feed or provide an equally explicit nearby disclosure. [R5 / `packages/web/src/pages/context.tsx:304`]

⚠️ 3. Do not collapse zero, unsupported, and partially covered states into an absent panel. The Server emits a real zero summary, but `decisionCount === 0` hides both the answer and the coverage caveat; an older Server with no `influence` field looks identical, and the always-visible Influence filter can then lead to an empty list. The caveat is also incomplete because only `en`/`zh` notes and post-deployment derivations are counted. Show a neutral zero/unsupported distinction and state the actual coverage boundary. [R5 / `packages/web/src/pages/context.tsx:300`]

✅ 4. Existing review comments already cover the strict-conversion mismatch, malformed URL decoding, private-chat content disclosure, and node ranking across repository identities; I did not duplicate those inline findings here.

**Action taken**
- Submitted request changes after static review. No code checks or tests were run as part of this review.

Gandy2025:
Companion tree PR: [first-tree-context#913](https://github.com/agent-team-foundation/first-tree-context/pull/913) — records the influence-vs-exposure decision, the derive-from-the-note contract, the no-dedicated-table rationale, and the managed-only coverage boundary.

Gandy2025:
Added the backfill for notes written between #2204 and this PR (`a4cab947c`).

Core is in `src/services/context-tree/influence-backfill.ts` rather than beside its CLI so it is typechecked and covered — it writes to `messages`, whose rows are immutable by contract, so an overwrite or a clobbered sibling key would not be recoverable.

```bash
DATABASE_URL=... pnpm --filter @first-tree/server exec tsx scripts/backfill-context-decision-from-notes.ts --limit 200
```

Dry run by default; `--apply` writes. Guards (one test each): rows that already carry a receipt are never loaded, so the write-time path stays their sole authority; human senders are skipped because quoting a note is not an attribution; `jsonb_set` adds exactly one key so a concurrent metadata write cannot be clobbered; an unconvertible note is counted and sampled, never written.

Not run against any real database — that is a separate call.

packages/server/src/services/context-tree/influence-backfill.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  // Candidates: agent-authored text bodies with no receipt yet. The scaffold
+  // match is a cheap prefilter — the shared parser is the authority, and it
+  // runs on every row this returns.
+  const rows = await db.execute<Row>(sql`
+    SELECT m.id, m.chat_id, m.content, m.metadata
+    FROM messages m
+    INNER JOIN agents a ON a.uuid = m.sender_id
+    ${organizationId ? sql`INNER JOIN chats c ON c.id = m.chat_id AND c.organization_id = ${organizationId}` : sql``}
+    WHERE a.type <> 'human'
```
yuezengwu: [P0] Bound this one-off mutation to the exact historical gap. The stated purpose is notes written between PR 2204 and deployment of write-time derivation, but this query scans every agent message ever that lacks the key. A later operator run can therefore rewrite messages outside that gap, including future rows intentionally left receipt-free by the write path, creating a second authority over message history. Require and validate immutable lower and upper created-at or message-id bounds, and print them before apply.

packages/server/src/services/context-tree/influence-backfill.ts
```diff
@@ -0,0 +1,139 @@
...
+      AND NOT (COALESCE(m.metadata, '{}'::jsonb) ? ${CONTEXT_DECISION_METADATA_KEY})
+      AND jsonb_typeof(m.content) = 'string'
+      AND (
+        m.content::text LIKE '%How Context Tree affected this work%'
+        OR m.content::text LIKE '%Context Tree 如何影响本次工作%'
+        OR m.content::text LIKE '%Context Tree impact%'
+        OR m.content::text LIKE '%Context Tree 影响%'
+      )
+    ORDER BY m.created_at ASC
+    LIMIT ${limit}
```
yuezengwu: [P1] Rerun cannot reliably make progress past this limit. Derived rows disappear after apply, but no-note, multi-note, and unconvertible rows remain candidates forever and are ordered first again. Once enough such rows fill the batch, every rerun scans the same ids and later convertible messages are never reached, despite the CLI saying to rerun for more. Add keyset pagination or process successive pages in one run using created-at plus id as a stable cursor.

packages/server/scripts/backfill-context-decision-from-notes.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { connectDatabase } from "../src/db/connection.js";
+import { backfillContextDecisionFromNotes } from "../src/services/context-tree/influence-backfill.js";
+
+function parseArgs(argv: readonly string[]) {
+  const apply = argv.includes("--apply");
+  const limitIndex = argv.indexOf("--limit");
+  const orgIndex = argv.indexOf("--org");
+  const limit = limitIndex >= 0 ? Number(argv[limitIndex + 1]) : 5_000;
+  const organizationId = orgIndex >= 0 ? argv[orgIndex + 1] : undefined;
```
yuezengwu: [P0] Validate that --org has a non-empty value. When --org is the final argument, organizationId becomes undefined and this destructive command silently falls back from intended single-org scope to every organization; combined with --apply that is the highest-risk possible typo. Reject missing option values and consider requiring explicit all-org acknowledgement rather than making global scope the implicit default.

yuezengwu (changes requested):
Incremental review of synchronized head a4cab947c: request changes remains.

The new commit adds a dry-run-by-default backfill that writes metadata.contextDecision into existing messages rows. This is a core database data mutation even though no schema migration is present. Three new blockers are inline: the mutation is not bounded to the stated 2204-to-deployment gap; batches can become permanently stuck behind unconvertible candidates; and a missing --org value can silently turn an intended scoped apply into an all-organization apply.

All prior blockers remain unchanged on this head: strict note conversion mismatch, malformed URL send-path exception, private-chat content disclosure, cross-repository node ranking, inconsistent aggregate validity, and overstatement of agent-reported evidence. CodeQL is failing again on the polynomial regular expression.

Companion first-tree-context PR 913 now records the desired access, trust, canonical-row-set, and binding-scope contract, but this code head has not yet been reconciled to it. The source PR description also still says that no backfill is included and should be updated before another review.

No tests or QA were run; this was a static incremental diff review and live-state check.

Gandy2025:
Pushed `2501a586` closing every blocker. Point by point:

**P0 — private-chat disclosure.** `summary` and `evidence` are body-derived, so an event the viewer cannot open is now **omitted from the feed entirely** rather than shown with the link disabled — there is no partially-redacted row worth rendering. `viewerCanAccess` is removed from the schema with it. Aggregates stay org-wide, matching the io summary, which already exposes per-node activity org-wide. Regression test covers an outsider (counted, not disclosed) and a member (disclosed).

**P0 — backfill scope.** `--since`/`--until` are now required, validated, and printed before apply; `--org` with no value throws instead of silently widening to every org; a deliberately global run must pass `--all-orgs`. Tests cover the window bound and the inverted-window rejection.

**P1 — malformed percent escape.** `decodeURIComponent` is wrapped, with `%E0%A4%A`, `%`, `%zz`, `a%2` in the test.

**P1 — strict conversion.** New `isConvertibleContextImpactNote` requires the same observations the eval grades on (current scaffold, logical lines, exact links, at end, blank line before). Five tests, one per rejection.

**P1 — one row set + repository scoping.** Totals no longer come from SQL while the feed re-parses. Every facet — headline, effects, ranking, feed — is derived from one `contextDecisionSchema`-valid set, filtered to the bound repository by canonical identity. Tests cover a foreign-repo citation, equivalent repo spellings, and the no-binding case.

**P1 — trust framing.** The panel leads with an `Agent-reported` eyebrow, says First Tree "does not verify causality", and states all three coverage gaps — managed chats only, English/Chinese notes only, only since this shipped — so a low count reads as incomplete rather than as a verdict on a node. A real zero now renders; only an absent block hides.

**P1 — rerun progress.** Keyset pagination on `(created_at, id)` replaces the limit that could never advance past unconvertible rows. Test seeds three blockers ahead of a convertible row with `pageSize: 2`.

**CodeQL.** Every repetition in the note patterns is bounded and the source line is length-capped before link scanning.

`pnpm check` + `pnpm typecheck` clean. shared 910, server 3239, web 2313, skill-evals 621 — all passing.

On merge ordering: this PR's reviews ask for the companion tree PR to land first, while [first-tree-context#913](https://github.com/agent-team-foundation/first-tree-context/pull/913)'s own reviewer cites the `first-tree-write` contract requiring the source PR to merge first and the tree to be reconciled against merged truth. Those point opposite ways — I have followed the tree repo's stated contract (source first) and left #913 open pending this. Happy to invert if you prefer.

packages/server/src/services/context-tree/influence.ts
```diff
@@ -0,0 +1,306 @@
...
+      LEFT JOIN agents a ON a.uuid = m.sender_id
+      WHERE c.organization_id = ${organizationId}
+        AND m.created_at >= ${sinceIso}::timestamptz
+        AND m.metadata ? ${CONTEXT_DECISION_METADATA_KEY}
+      ORDER BY m.created_at DESC
+      LIMIT ${INFLUENCE_SCAN_CAP}
+    `),
+  );
+
+  if (rows.length >= INFLUENCE_SCAN_CAP) {
```
yuezengwu: [P1] Do not return a normal metric after truncating the canonical row set. Once an organization has more than 5,000 candidate receipts in the window, every headline, effect, and node value is an undercount, but only a server log records that fact; the API and UI present it as complete. Paginate through the full indexed window, or add an explicit partial or unavailable state that the schema and UI display. A hard cap with a hidden warning violates the same canonical-evidence requirement this rewrite is intended to restore.

packages/server/src/services/context-tree/influence-backfill.ts
```diff
@@ -0,0 +1,187 @@
...
+        AND m.created_at <= ${untilIso}::timestamptz
+        AND NOT (COALESCE(m.metadata, '{}'::jsonb) ? ${CONTEXT_DECISION_METADATA_KEY})
+        AND jsonb_typeof(m.content) = 'string'
+        AND (
+          m.content::text LIKE '%How Context Tree affected this work%'
+          OR m.content::text LIKE '%Context Tree 如何影响本次工作%'
+        )
+        ${cursor ? sql`AND (m.created_at, m.id) > (${cursor.createdAt}::timestamptz, ${cursor.id})` : sql``}
+      ORDER BY m.created_at ASC, m.id ASC
+      LIMIT ${pageSize}
```
yuezengwu, line 157: [P1] maxRows is neither an exact blast-radius limit nor a resumable checkpoint. The query always fetches pageSize rows and the check runs only after the whole page is processed, so --max-rows 1 can write 500 messages. If the run stops, the cursor is lost; rerunning starts at the beginning, and unconvertible rows remain candidates, so the same first maxRows can be rescanned forever. Limit each query to the smaller of pageSize and the remaining allowance, and return and accept the last created_at plus id cursor, or otherwise make one invocation exhaust the bounded window.

packages/web/src/pages/context.tsx
```diff
@@ -262,6 +267,107 @@ function ContextSignal({ snapshot }: { snapshot: ContextTreeSnapshot }) {
...
+      ) : null}
+      {/* Three separate coverage gaps, each of which makes this an undercount:
+          a BYO session prints its note in the user's own terminal and never
+          reaches the Server; only English and Chinese notes can be read back;
+          and nothing before this feature shipped was recorded. Stating all
+          three is what stops a low number from reading as a verdict on the
+          tree — or on a node missing from the list above. */}
+      <p className="context-influence-coverage">
+        Reported by agents in their own answers — First Tree preserves the cited version for inspection, but does not
+        verify causality. Counted from managed chats only, from English and Chinese notes, and only since this feature
```
yuezengwu: [P2] This coverage statement no longer matches the PR: the included backfill can add receipts for notes written before ship, and it may or may not have been applied in a given deployment. Only since this feature shipped becomes false after a successful backfill and still overstates completeness before one. State that pre-shipment notes are included only where the bounded backfill was applied, or expose the backfill coverage state from the Server.

yuezengwu (changes requested):
Incremental review of head 2501a586: request changes remains.

The prior strict-parser, malformed-decoding, private-content disclosure, repository-scoping, canonical-row-set, trust-framing, zero-state, and CodeQL blockers are addressed on this head. The bounded backfill also fixes the earlier all-org typo and makes progress within a single invocation.

Two new correctness blockers remain. First, influence aggregation silently truncates at 5,000 candidates but returns an ordinary complete-looking headline and ranking. Second, maxRows is checked only after a full page, so it is not an exact mutation bound, and the in-memory cursor is lost between the reruns the CLI instructs operators to perform. The coverage copy also contradicts the newly included historical backfill. Details are inline.

This still changes persisted messages.metadata.contextDecision semantics and includes a core historical message-data mutation, with no schema migration. The companion first-tree-context PR 913 now carries the intended durable contract. I withdraw the earlier tree-first merge condition: the documented workflow is source first, then reconcile the still-open Tree change against merged source truth.

Per review policy I inspected the incremental diff and live checks but did not run tests or QA. CodeQL is now green; two broader test jobs were still running when this review was submitted.

baixiaohang (changes requested):
- Rationale: The current head fixes most prior blockers, but private body-derived citation fields still escape through the org-wide ranking, and two trust-boundary identity paths can still persist or count the wrong claim.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes `services/chat/message.ts`, persisted `messages.metadata.contextDecision` semantics, and a historical message backfill -> B-high.

**PR summary**
- Problem: The Context tab shows that agents opened the Tree but cannot show which decisions agents report it influenced.
- Approach: Derive a structured receipt from the visible note, aggregate schema-valid receipts for the current binding, expose agent-reported influence in the Context tab, and backfill the bounded historical gap.
- Impacted modules: shared note and receipt contracts; Server message persistence, backfill, and Context snapshot aggregation; Web Context and chat surfaces; first-tree-read eval parsing.

**Review findings**
❌ 1. [P0] Keep every body-derived evidence field behind chat access, including the ranked-node projection. `rankNodes(receipts)` runs over all org receipts before `accessible` is applied, and copies `evidence.heading`, `nodePath`, `repoUrl`, and `commit` into `nodes`; the Web renders those fields for every org member. An outsider therefore receives no feed event but can still read the private note's citation label and exact evidence through the ranking. Either filter/redact this projection for inaccessible chats or rebuild its display fields from the independently org-visible bound snapshot, and assert the outsider's entire response rather than only `recentEvents`. [`packages/server/src/services/context-tree/influence.ts:283`]

❌ 2. [P1] Do not fall back to a parallel receipt after detecting a malformed or multiple visible note. When parsing finds note-shaped content but conversion returns `null`, execution logs and then accepts any schema-valid supplied `metadata.contextDecision`; that receipt can contradict the sentence the reader sees, recreating the divergence derivation is meant to remove. Limit the legacy payload path to bodies with zero parsed notes; a present-but-invalid or ambiguous note should strip or reject the supplied receipt. Add the combined malformed-note-plus-valid-payload case. [`packages/server/src/services/chat/message.ts:176`]

❌ 3. [P1] Match the bound repository at provider authority, not with the port-blind canonical key. `canonicalGitRepoUrl` intentionally drops HTTPS ports, so a binding such as `https://gitlab.example.com:8443/group/tree` matches a citation to `https://gitlab.example.com/group/tree`, even though those are different GitLab origins. Pass provider/instance-origin context and use the existing provider-aware repository matcher (or an equivalent exact web-origin identity), with a self-managed GitLab port regression. [`packages/server/src/services/context-tree/influence.ts:146`]

✅ 4. The strict parser, malformed percent decoding, zero/trust copy, canonical schema-valid row set, basic binding scope, and inaccessible feed-event fixes are present. I did not duplicate the separately filed scan-cap, backfill-resume, and coverage-copy findings.

**Action taken**
- Submitted request changes after static incremental review. No code checks, tests, or QA were run as part of this review.

Gandy2025:
Pushed `f08d2f4b` closing round two.

**P0 — heading leaked through the ranking.** You were right that omitting feed events was not enough: `rankNodes` ran over all org receipts and copied `evidence.heading`, which is the agent's own label written inside a private body. The ranking now carries **only fields describing the org's own bound repository** — `nodePath`, `repoUrl`, `commit` — all of which a member already reads from the snapshot. `title` is gone from the DTO entirely; the Web resolves a display name from the org-visible `nodes[]` (which is also strictly better: it shows the canonical tree title rather than whatever the agent typed). The regression asserts the **entire** outsider response contains no citation label, plus the exact key set of a ranked node.

**P1 — contradicting fallback.** Limited the legacy payload path to bodies with zero parsed notes. A present-but-invalid or ambiguous note now strips the supplied receipt rather than storing a claim that can contradict the visible sentence. Test covers malformed-note-plus-valid-payload and two-notes-plus-valid-payload.

**P1 — port-blind matching.** Switched to `sameContextTreeRepository` and plumbed provider + GitLab instance origin through both snapshot routes. Regression: a citation to `gitlab.example.com:8443` does **not** match a binding on `gitlab.example.com`, and does match the same-port binding.

**P1 — silent truncation.** The scan cap now sets `truncated` on the response as well as logging, so the counts are visibly a floor rather than a complete-looking answer.

**P1 — backfill bounds and resume.** `maxRows` is checked per row, so `--max-rows 100` never touches a 101st message. The run returns `nextCursor` and the CLI prints a ready-to-paste `--resume-after "<ISO>|<messageId>"`, so the rerun it instructs actually continues instead of restarting at the window's beginning. Test stops at 2 of 4 rows and resumes to completion.

**Coverage copy.** Dropped "only since this feature shipped" — the backfill in this same PR contradicts it. Now: managed chats only, and only notes this Server could read back.

`pnpm check` + `pnpm typecheck` clean. shared 910, web 2313, server 3241 (the two `agent-templates` failures in my parallel run are test-isolation flake — that file passes 43/43 alone and this PR does not touch it).

Also noted: thank you for withdrawing the tree-first condition — that resolves the contradiction I flagged, and [first-tree-context#913](https://github.com/agent-team-foundation/first-tree-context/pull/913) stays open for reconciliation against merged source truth.

packages/web/src/pages/context.tsx
```diff
@@ -262,6 +267,126 @@ function ContextSignal({ snapshot }: { snapshot: ContextTreeSnapshot }) {
...
+          <div className="context-influence-nodes-title">Nodes agents cited as changing a decision</div>
+          <ul className="context-influence-nodes">
+            {nodes.map((node) => {
+              // Same exact-commit link the chat receipt built, so a node here
+              // and a citation there resolve to the identical file version.
+              const href = contextTreeSourceHref(
+                { repoUrl: node.repoUrl, commit: node.commit, nodePath: node.nodePath },
+                instanceOrigin,
+              );
+              const title = nodeTitleByPath.get(node.nodePath) ?? nodeFileName(node.nodePath);
```
yuezengwu, line 373: [P0] This fallback demonstrates that the private citation still escapes. Only heading was removed; nodePath, repoUrl, and commit are still copied from the message receipt into the org-wide response. When the cited path is absent from snapshot.nodes, which this branch explicitly handles, it is not independently org-visible: a private note can cite an old or fabricated path such as secret/acquisition.md, and an outsider now sees that path, its filename, commit, and link. Intersect the ranking projection with server-owned snapshot data, or omit ranking rows from inaccessible chats; resolving only the title client-side does not satisfy the prior requirement that every body-derived evidence field stay behind chat access.

packages/server/src/services/context-tree/influence.ts
```diff
@@ -0,0 +1,333 @@
...
+      createdAt: receipt.createdAt,
+    }));
+
+  return {
+    windowDays,
+    decisionCount: receipts.length,
+    effects,
+    nodes: rankNodes(receipts),
+    recentEvents,
+    truncated,
```
yuezengwu, line 369: [P1] Returning truncated does not make the floor visible because the Web never reads this field. ContextInfluence still renders decisionCount, effects, and nodes exactly like a complete response, and there is no DOM test for the truncated state. Consume the flag with explicit nearby floor or partial wording, or paginate the full set. Also fetch cap plus one before setting the flag; rows.length equal to the cap does not prove that another row exists.

packages/server/src/services/context-tree/influence-backfill.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    for (const row of rows) {
+      // Checked per row so `maxRows` is an exact bound on what this run
+      // touches, not "the last page may overshoot it".
+      if (scanned >= maxRows) {
+        stoppedAtMaxRows = true;
+        break;
+      }
+      scanned += 1;
+      lastExamined = { createdAt: isoOf(row.created_at), id: row.id };
```
yuezengwu: [P1] Preserve the database sort key precision in this resume cursor. PostgreSQL stores created_at with microseconds, but a Drizzle timestamp becomes a JavaScript Date and toISOString keeps only milliseconds. The next raw tuple predicate can therefore see the last examined row as greater than its rounded cursor and select it again; with maxRows 1 and an unconvertible boundary row, every resumed run can stop on the same row forever. Compare and order on date_trunc milliseconds plus id, as the existing message and document cursors do, or select and carry an exact textual database timestamp. Add a regression using a timestamp with nonzero microseconds.

yuezengwu (changes requested):
Incremental review of head f08d2f4b: request changes remains.

This head correctly closes the malformed-note fallback and provider-aware repository matching findings, and it removes the citation heading from the org-wide ranking. However, the privacy projection is still body-derived for path, repository, and commit; the new Web fallback actively renders a path that is absent from the independently org-visible snapshot. The scan-cap flag is also never consumed by Web, so the visible metrics remain complete-looking when truncated. Finally, the resumable backfill round-trips a PostgreSQL timestamp through millisecond-only JavaScript Date precision while comparing against the raw microsecond column, which can re-select the boundary row and defeat progress. Details are inline.

The implementation goal remains deriving immutable messages.metadata.contextDecision from one visible agent note and aggregating that agent-reported signal into the Context tab. This is still B-high: it changes persisted message semantics and includes a historical message-data mutation, without a schema migration.

The PR description is also stale: it still says no backfill is included and reports the original verification counts. Please reconcile it before approval. Companion first-tree-context PR 913 should remain open for source-first reconciliation after this code lands.

Per review policy I inspected the incremental diff and live checks but did not run tests or QA. CodeQL, Server tests, lint, and most jobs were green; Client and Web tests were still running at submission time.

baixiaohang (changes requested):
- Rationale: The second-round patch fixes the legacy fallback, provider-aware repository match, and resumable backfill, but private citation identity still escapes through the org-wide ranking and the scan-cap state is still invisible to users.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes `services/chat/message.ts`, persisted `messages.metadata.contextDecision` semantics, and a historical message backfill -> B-high.

**PR summary**
- Problem: The Context tab shows that agents opened the Tree but cannot show which decisions agents report it influenced.
- Approach: Derive a structured receipt from the visible note, aggregate schema-valid receipts for the current binding, expose agent-reported influence in the Context tab, and backfill the bounded historical gap.
- Impacted modules: shared note and receipt contracts; Server message persistence, backfill, and Context snapshot aggregation; Web Context and chat surfaces; first-tree-read eval parsing.

**Review findings**
❌ 1. [P0] The ranked-node projection still exposes private body-derived evidence. `toValidReceipt` verifies only that `evidence.repoUrl` matches the binding, then `rankNodes` copies the note's unverified `nodePath`, `repoUrl`, and `commit` into the org-wide response. Those values are not proven to exist in `snapshot.nodes` or at the bound commit; the Web explicitly falls back to rendering the file name when the path is absent from the snapshot. An inaccessible private note can therefore cite an arbitrary same-repo path such as `customers/SECRET.md` and an arbitrary full hash, and every org member receives both. Removing `heading` closes only one field of the original finding. Restrict the ranking to server-known snapshot nodes and build all display/link identity from that org-visible source, or omit the projection for inaccessible receipts. The outsider regression should use a secret path and commit and assert neither appears anywhere in the response. [R5 / `packages/server/src/services/context-tree/influence.ts:154`, `packages/server/src/services/context-tree/influence.ts:203`, `packages/web/src/pages/context.tsx:357`]

❌ 2. [P1] The previously filed scan-cap blocker remains at the user surface. The Server now adds `influence.truncated`, but `ContextInfluence` destructures only `decisionCount`, `effects`, and `nodes`; no Web code reads or renders the flag. A response with `truncated: true` therefore still presents an ordinary complete-looking headline and ranking. Surface an explicit partial/“at least” state with a DOM regression, or scan the full window. Also fetch `cap + 1` before setting the flag: `rows.length >= 5_000` currently reports truncation when the window contains exactly 5,000 candidates. [R4/R5 / `packages/server/src/services/context-tree/influence.ts:250`, `packages/web/src/pages/context.tsx:330`]

✅ 3. The malformed/multiple-note fallback, GitLab origin matching, exact per-run backfill mutation bound, resumable cursor, and revised coverage wording are addressed on this head.

**Action taken**
- Submitted request changes after static incremental review. No code checks, tests, or QA were run as part of this review.

Gandy2025:
Pushed `b2a91810` — round three.

**P0 — you were right that removing `heading` was only one field.** `nodePath`, `repoUrl`, and `commit` are equally asserted by a private body and never verified, and my client-side fallback actively rendered a path absent from the snapshot. Fixed at the projection, not the display:

- The ranking DTO is now `{ nodePath, decisionCount }` — no title, no repository, no commit.
- The Server keeps a citation **only when its path exists in the snapshot the caller already holds** (`knownNodePaths`, built from `snapshot.nodes` in both routes). With no known paths the ranking is empty — silence is the safe failure, never "publish whatever the note claimed".
- Title and href are rebuilt from that same snapshot (`snapshot.repo` + `snapshot.headCommit`), so nothing in this projection originates in an inaccessible chat.
- The regression uses your exact scenario: a private note citing `customers/SECRET.md` at `fff…` with heading `SECRET LABEL`, asserting none of the four strings appears **anywhere** in the outsider's serialized response.

**P1 — truncation.** The flag is now consumed: headline renders `5000+` with an explicit floor notice, DOM regression included, and the zero-state test asserts the notice is absent. The query fetches `cap + 1` so a window holding exactly the cap is reported complete.

**P1 — cursor precision.** Correct and subtle, thank you. `timestamptz` keeps microseconds, `Date.toISOString()` keeps milliseconds, so the rounded cursor re-selected the boundary row — with `--max-rows 1` on an unconvertible row that loops forever. The cursor is now the database's own `created_at::text`, carried end to end with no `Date` round-trip (the CLI passes the string through and only validates it). Regression seeds `.123456` and `.223456` rows and asserts the resumed run derives the second one.

**PR description** rewritten: it now covers the backfill, the trust/privacy boundaries, and current counts.

`pnpm check` + `pnpm typecheck` clean. shared 910, server 3243, web 2313.

packages/server/src/api/context-tree-snapshot.ts
```diff
@@ -24,6 +25,24 @@ import {
...
+ * citation may name them (node path, its source file, and the `.md` form).
+ * The influence ranking echoes a path back only when it appears here.
+ */
+function snapshotNodePaths(nodes: readonly { path: string; sourcePath: string | null }[]): ReadonlySet<string> {
+  const paths = new Set<string>();
+  for (const node of nodes) {
+    const path = normalizeNodePath(node.path);
+    if (path.length > 0) {
+      paths.add(path);
+      paths.add(`${path}.md`);
```
yuezengwu: [P1] Canonicalize each citation to one snapshot node before ranking. This Set admits `node.path`, `${node.path}.md`, and `sourcePath`, but `rankNodes` groups and returns whichever spelling the private receipt used. The same leaf cited as `foo` and `foo.md`, or a domain as `dir` and `dir/NODE.md`, therefore becomes separate ranked rows with split counts; an exact blob URL to an extensionless `foo` can also match the snapshot node for `foo.md`, then Web builds a link to a file the snapshot never contained. Since evidence.nodePath is the exact source-file path from the blob URL, map accepted `sourcePath` values to one snapshot-owned canonical path/node identity and aggregate on that value. Apply the same fix to the duplicate org route helper and add a mixed-alias regression.

packages/server/scripts/backfill-context-decision-from-notes.ts
```diff
@@ -0,0 +1,123 @@
...
+    const id = resumeRaw.slice(separator + 1);
+    if (separator <= 0 || Number.isNaN(new Date(createdAt).getTime()) || id.length === 0) {
+      throw new Error('--resume-after must be "<timestamp>|<messageId>", exactly as printed by the previous run');
+    }
+    resumeAfter = { createdAt, id };
+  }
+
+  const maxRowsRaw = optionValue(argv, "--max-rows");
+  const maxRows = maxRowsRaw === undefined ? undefined : Number(maxRowsRaw);
+  if (maxRows !== undefined && (!Number.isFinite(maxRows) || maxRows <= 0)) {
```
yuezengwu: [P1] Require `maxRows` to be a positive safe integer, and enforce the same invariant inside `backfillContextDecisionFromNotes`. With the current test, `--max-rows 1.5` is accepted; the per-row `scanned >= maxRows` check then processes two rows, violating the exact mutation bound this destructive command promises. Programmatic callers can also pass zero, a fraction, `NaN`, or `Infinity` directly to the service. Reject non-integer/unsafe values at both boundaries and cover the fractional case.

packages/web/src/pages/context.tsx
```diff
@@ -262,6 +267,142 @@ function ContextSignal({ snapshot }: { snapshot: ContextTreeSnapshot }) {
...
+          {effects.conflicted > 0 ? (
+            <span className="context-influence-conflict">
+              {effects.conflicted === 1 ? "1 surfaced a conflict" : `${effects.conflicted} surfaced a conflict`}
+            </span>
+          ) : null}
+        </span>
+      </div>
+      {truncated ? (
+        <p className="context-influence-partial">
+          This window holds more decisions than one pass reads, so these counts are a floor and the ranking may be
```
yuezengwu: [P2] Describe this as a candidate-receipt cap, not proof that the window holds more decisions. The Server fetches rows with the metadata key before schema validation and bound-repository filtering, so the extra row that sets `truncated` may be malformed or foreign and `decisionCount` can still be zero. Say that more candidate receipts existed than this pass inspected and that the reported counts may therefore be incomplete; otherwise the trust copy upgrades unvalidated rows into decisions.

yuezengwu (changes requested):
Incremental review of head b2a91810: request changes remains.

The implementation goal is still to derive immutable messages.metadata.contextDecision from the one visible agent impact note, aggregate that agent-reported signal into the Context snapshot, and safely backfill the bounded historical gap. This head correctly closes the prior private-citation projection, invisible scan-cap, and microsecond cursor findings: ranking output is snapshot-gated and drops private source fields, truncation is visible with a cap-plus-one query, and the resume cursor preserves the database timestamp text. The PR description is now reconciled too.

Two correctness blockers remain inline. The snapshot gate accepts multiple aliases for one node but the ranking aggregates and links by the uncanonicalized citation string, so one node can split into duplicate rows and extensionless aliases can produce links to files the snapshot does not contain. The destructive backfill's advertised exact maxRows bound also accepts fractions (and the service itself accepts invalid programmatic values), so --max-rows 1.5 can examine and mutate two rows. I also left a non-blocking wording correction because the scan cap counts candidate metadata rows, not confirmed decisions.

Risk remains B-high: there is no schema migration, but this changes persisted message-metadata semantics and includes a historical messages data mutation. The companion first-tree-context PR 913 should remain open for source-first reconciliation after this source PR lands.

Per review policy I inspected the incremental diff and live PR/check state but did not run tests or QA. CodeQL and lint/typecheck are green; Server and Client/Web test jobs were still running when this review was submitted.

baixiaohang (changes requested):
- Rationale: The latest patch closes the prior privacy, truncation, and cursor blockers, but the ranked-node allowlist still treats logical snapshot paths as if they were real source files, so fabricated citations can be counted and linked.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes persisted `messages.metadata.contextDecision` semantics and includes a historical message backfill.

**PR summary**
- Problem: The Context tab shows that agents opened the Tree but cannot show which decisions agents report it influenced.
- Approach: Derive a structured receipt from the visible note, aggregate schema-valid receipts for the current binding, expose explicitly agent-reported influence, and backfill the bounded historical gap.
- Impacted modules: shared note and receipt contracts; Server message persistence, backfill, and Context snapshot aggregation; Web Context and chat surfaces; first-tree-read eval parsing.

**Review findings**
❌ 1. [P1] Treat only exact snapshot `sourcePath` values as known citation paths. `snapshotNodePaths` currently also inserts `node.path` and `${node.path}.md`, but `node.path` is a logical navigation identity, not necessarily a file. For a directory node whose real source is `system/cloud/context-tree/NODE.md`, this allowlists a fabricated citation to `system/cloud/context-tree.md`; it is then counted and Web builds a snapshot-head link that does not exist. That breaks the receipt contract that `nodePath` is the exact Tree-root-relative source path and lets an unverified private-body claim earn a ranked attribution. Build the allowlist from non-null `sourcePath` only, centralize the identical privacy-critical helper used by both snapshot routes, and add a regression where the fabricated directory `.md` path is rejected while its actual `NODE.md` source is accepted. [R1/R4/R5 / `packages/server/src/api/context-tree-snapshot.ts:33`]

✅ 2. The inaccessible-chat projection, visible truncation state with cap-plus-one detection, and exact textual PostgreSQL resume cursor are otherwise addressed on this head.

**Action taken**
- Submitted request changes after static incremental review. No code checks, tests, or QA were run as part of this review.

Gandy2025:
CI note: `Test Server` failed on `b2a91810` at `admin-sessions-suspend-terminate.test.ts > fails closed when the applied Reset cannot be finalized AND its abort is never receipted` (`ackSentCommand` timing out waiting for a sent command).

Not from this PR: the branch touches nothing under `api/agent/`, `services/agents/`, or that test file (`git diff origin/main...HEAD` on those paths is empty), and the file passes 49/49 locally on this head across four consecutive runs. That test name has no prior failure recorded in this repo, and `main`'s last 12 CI runs are 9 success / 3 cancelled — so it reads as a timing-sensitive WS ack flake on a slower runner rather than a regression.

The second red check, `Test`, is only the aggregation gate reporting `SERVER_RESULT: failure`; `Test Client & Web`, `Test CLI`, and the smoke jobs were all green.

Re-ran the failed jobs rather than touching unrelated code. Will report the result.

Gandy2025:
Pushed `7ed41706` — round four.

**P1 — allowlist admitted logical paths.** Correct and I had it backwards: `node.path` is a navigation identity, not a file. A directory node sourced at `system/cloud/context-tree/NODE.md` was allowlisting a fabricated `system/cloud/context-tree.md`, which then got counted and linked to a file that does not exist. The allowlist is now built from non-null `sourcePath` **only** — exact match, no derived forms. Regression covers both directions: the fabricated `.md` is rejected while the real `NODE.md` is accepted, plus a unit test on the helper itself.

That also closes the duplicate-row half of the finding: with one string per node in the allowlist, and the receipt schema already refusing a leading-slash `nodePath`, a citation either matches that single form or is dropped. Test asserts repeated citations aggregate to one row and that an unstorable spelling never reaches the ranking.

The helper moved into the influence service and both routes call it — you were right that a privacy gate must not exist in two copies. The mocked-route test now does `importActual` and stubs only the aggregation, so a broken gate can no longer pass there (it caught this immediately: the old whole-module mock 500'd).

**P1 — fractional row bound.** `--max-rows 1.5` passed `Number.isFinite` and then let `scanned >= 1.5` touch two rows on a destructive run. Both the CLI and the service now require a positive integer, so a programmatic caller is covered too. Test asserts both `maxRows` and `pageSize`.

**Non-blocking wording.** Fixed — the cap counts candidate metadata rows, not confirmed decisions; the comment and log message now say so.

`pnpm check` + `pnpm typecheck` clean. shared 910, server 3247, web 2313.

Previous CI red was an unrelated flake (`admin-sessions-suspend-terminate` WS ack timing); re-running the failed job alone turned it green with no code change.

yuezengwu (approved):
Incremental review of head 7ed41706: approved.

The implementation goal remains deriving immutable `messages.metadata.contextDecision` from the one visible agent impact note, aggregating that explicitly agent-reported signal into the Context snapshot, and safely backfilling the bounded historical gap.

This head closes both remaining blockers. The ranked-node privacy gate now accepts only exact non-null snapshot `sourcePath` values, is centralized for both snapshot routes, and has regressions for fabricated directory aliases and repeated citations. The destructive backfill now rejects non-positive, fractional, unsafe, and non-finite `maxRows`/`pageSize` values at the service boundary, with the CLI enforcing the same `maxRows` invariant.

One non-blocking wording item remains from the prior review: `packages/web/src/pages/context.tsx` still says “This window holds more decisions than one pass reads.” The commit corrected the Server comment and log, but not this visible sentence; because the cap is applied before schema and bound-repository filtering, it should say candidate receipts exceeded the pass and counts may be incomplete. This does not block approval.

Risk remains B-high: there is no schema migration, but the PR changes persisted message-metadata semantics and contains a historical `messages` data mutation. The author states that the backfill has not been run against a database. Companion `first-tree-context#913` should remain open for source-first reconciliation against merged source truth.

Per review policy I inspected the incremental diff and live PR/check state but did not run tests or QA. Lint/typecheck and the completed checks were green; CodeQL JavaScript, Server tests, and Client/Web tests were still running at approval time.

Gandy2025:
Thanks — approval noted, and I took the non-blocking wording item too (`75d756c8`): the Web notice now says "more candidate receipts than one pass reads", matching the Server comment and log. You were right that the cap lands before schema and bound-repository filtering, so "decisions" was the wrong noun on the only surface a user actually reads.

@baixiaohang this head also closes both of your round-four findings — the `sourcePath`-only allowlist (centralized, with the fabricated-directory regression you asked for) and the integer `maxRows`/`pageSize` guard at the service boundary. Re-review requested.

yuezengwu (approved):
Incremental review of head 75d756c85: approved.

This one-file follow-up closes the sole non-blocking item from my prior approval. The visible truncation notice now correctly calls the pre-validation, pre-repository-filter rows ‘candidate receipts’ rather than upgrading them to confirmed decisions. It does not alter the previously approved parser, persistence, aggregation, privacy-gate, or backfill behavior, and I found no new issues in the incremental diff.

The implementation goal remains deriving immutable messages.metadata.contextDecision from the one visible agent impact note, aggregating that explicitly agent-reported signal into the Context snapshot, and safely backfilling the bounded historical gap.

Risk remains B-high: there is no schema migration, but the overall PR changes persisted message-metadata semantics and contains a historical messages data mutation. The author states that the backfill has not been run against a database. Companion first-tree-context#913 should remain open for source-first reconciliation against merged source truth.

Per review policy I inspected the incremental diff and live PR/check state but did not run tests or QA. Server tests and the completed checks were green; CodeQL JavaScript, lint/typecheck, and Client/Web tests were still running at approval time.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All previously blocking parser, trust-boundary, privacy, repository-identity, aggregation, truncation, and bounded-backfill findings are resolved on the current exact head, with no new blocking issues.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes message-write behavior, persisted `messages.metadata.contextDecision` semantics, and includes a historical message backfill.

**PR summary**
- Problem: The Context tab can show that agents opened the Tree, but not which decisions agents report it influenced or which current nodes are repeatedly cited.
- Approach: Derive one structured receipt from the visible impact note at message-write time, aggregate schema-valid receipts for the current binding, expose explicitly agent-reported influence with chat-access protections, and provide a bounded, dry-run-first historical backfill.
- Impacted modules: shared note and receipt contracts; Server message persistence, backfill, and Context snapshot aggregation; Web Context and chat surfaces; first-tree-read eval parsing.

**Review findings**
✅ 1. Strict conversion, malformed URL handling, legacy fallback, canonical receipt validation, provider-aware repository matching, private-chat projection, current-snapshot source-path gating, and agent-reported framing now fail closed at their trust boundaries.
✅ 2. The scan cap is visible and accurately described as candidate receipts, while the backfill has explicit time/org scope, safe integer row bounds, keyset pagination, and a microsecond-preserving resume cursor.
✅ 3. The companion `first-tree-context#913` remains open for source-first reconciliation after this source PR lands.

**Action taken**
- Approved after static incremental review and explicit B-high authorization. No local code checks, tests, or QA were run as part of review; all required GitHub checks are green on this head.

PR #2279, opened 2026-08-11
fix(context): preserve BYO CLI channel invocation

### LINKED ISSUE #2277: fix(context): distinguish CLI/plugin channel mismatch from incompatible plugin
## Problem

When Codex has a healthy staging Context Plugin installed but the production CLI is invoked, `context route` reports the generic `CONTEXT_PLUGIN_RELOAD_REQUIRED` error:

```json
{
  "ok": false,
  "error": {
    "code": "CONTEXT_PLUGIN_RELOAD_REQUIRED",
    "message": "This provider session is not using a compatible First Tree Context Plugin. Repair the Plugin or start a new session, then retry."
  }
}
```

This suggests that Codex or the loaded plugin is incompatible, although the actual problem is a release-channel mismatch.

## Reproduction

Environment:

- Codex: `0.144.1` (minimum supported: `0.144.0`)
- Production CLI: `first-tree 0.5.19`
- Staging CLI: `first-tree-staging 0.5.20-staging.1147.1`
- Installed plugin: `first-tree-context@first-tree-staging 1.0.1`, enabled

Run the production command:

```sh
first-tree --json context route --provider codex --project-root <project>
```

Observed:

- `CONTEXT_PLUGIN_RELOAD_REQUIRED`
- Production `context status` reports the plugin and install manifest as missing

Run the staging command:

```sh
first-tree-staging --json context route --provider codex --project-root <project>
```

Observed:

- Routing succeeds
- Staging `context status` reports the plugin as installed, enabled, and runtime-healthy

## Expected behavior

The failure should distinguish a channel-specific installation gap from a genuinely missing, damaged, revoked, or protocol-incompatible adapter.

Possible outcomes:

- Report that the production Context Plugin is not installed for the production CLI, rather than saying the current provider session is incompatible.
- When safely detectable, mention that a different First Tree channel is active and direct the user to the exact channel executable supplied by the current loader/session handoff.
- Preserve the fail-closed channel boundary; do not silently fall back across channels.

## Additional context

The active staging plugin loader already returns the exact executable:

```text
/Users/gandy/.local/bin/first-tree-staging
```

Using that loader-resolved executable fixes the routing failure without reinstalling or repairing the plugin.

baixiaohang:
## Summary

- carry the current channel's exact CLI invocation in every BYO Core Skill loader response
- require canonical Read/Write workflows to preserve that opaque invocation across route, snapshot, verify, recovery, cleanup, and follow commands
- distinguish missing or mismatched channel installs from actual Context Plugin reload requirements
- bump both thin adapter patch versions because their projected bytes changed
- update the Web setup prompt, reference docs, QA case, and contract floors

## Validation

- CLI context loader, adapter observation, bundle, and current-session handoff: 40 tests passed
- adapter sync and command: 14 tests passed
- shared context integration: 8 tests passed
- Read/Write Skill floors: 11 tests passed
- Web context enablement: 16 tests passed
- typecheck passed for `@first-tree/shared`, `first-tree-dev`, `@first-tree/web`, and `@first-tree/skill-evals`

Fixes #2277


yuezengwu (changes requested):
Blocker: the portable Context loader now crashes before it can return the new response.

`loadContextSkill()` unconditionally calls `resolveCliInvocation()` at `apps/cli/src/core/context-integration/skill-loader.ts:38`. In the portable artifact smoke, the checked artifact is intentionally invoked through its extracted version-local shim. That shim sets `FIRST_TREE_PORTABLE_ROOT` to the version directory, while `resolvePortableInstallContext()` requires a stable `current` symlink, so the `Verify linux-x64 portable artifact` step fails with:

> `FIRST_TREE_PORTABLE_ROOT must name the stable current symlink`

This is directly attributable to this change: the same loader invocation reached the Core-path assertions before `firstTreeInvocation` was added, and it now exits 1 before emitting JSON. Please preserve the exact invocation without breaking this supported artifact-verification path—for example, make the verifier establish the stable `current` symlink/channel shim and assert the returned `firstTreeInvocation`, or provide an equally strict resolver path for the extracted artifact invocation. The regression coverage should exercise the portable loader end to end and verify both its Core paths and invocation.

I found no other blockers in the code diff. I did not run local tests, per the review scope; this finding comes from the live deterministic CI failure and source inspection.

baixiaohang:
Addressed the portable loader blocker in `081d15502`.

The artifact verifier now keeps the extracted version-local shim limited to the standalone `--version` smoke, then emulates the installer's stable `current` symlink and outer channel/alias shims before running `context skill load`. The end-to-end assertion covers both immutable Core paths and the shell-quoted stable `firstTreeInvocation`; the production portable resolver remains strict.

Locally validated with the same prod `0.0.0` linux-x64 portable build and `verify-portable-artifact.mjs` flow used by CI. Biome and `git diff --check` also pass.

yuezengwu (approved):
Re-reviewed head `081d15502936a453794fba3e461a0396f21cd354`; the portable loader blocker is resolved.

The verifier now keeps the extracted version-local shim scoped to the standalone `--version` check, then constructs the installer's stable `current` symlink plus channel/alias shims before invoking `context skill load`. Its end-to-end assertion covers both immutable Core paths and the shell-quoted stable `firstTreeInvocation`, while the production portable resolver remains strict.

No further findings. I did not run local tests, per the review scope; the updated CI run is still in progress and is not required for this review action.

PR #2288, opened 2026-08-12
refactor(client): S4 three-package split (cloud-client / client-runtime / client-providers)

yuezengwu:
## Summary
- Replace private `@first-tree/client` with three private packages: `@first-tree/cloud-client` (HTTP SDK + observability), `@first-tree/client-runtime` (connection/session/ACK/persistence), and `@first-tree/client-providers` (built-in providers + shared handler helpers).
- Enforce production dependency direction: `client-runtime → cloud-client`, `client-providers → client-runtime/{contracts,provider-support}` only; `apps/cli` composes all three and remains the sole public npm/bin/portable surface. `packages/client` is deleted with no compatibility shell.
- Invert the two cycles: managed-skills takes an explicit fail-closed `providerSkillRoots` projection (no providers import in runtime production); builtin-registry logging takes an injected log seam (no cloud-client import in provider production).
- Wire CI path filters and `first-tree-hub` allow-list paths to the three new packages.
- Keep the three private packages as CLI `devDependencies` (`workspace:*`) only; pack-time staging strips source-only development metadata, and `tsdown` `noExternal: [/^@first-tree\//]` inlines the packages so public tarballs and installed manifests expose no private workspace coordinates or imports.

## Notes for reviewers / QA
- Base: `19e66032af7f9f482168c350fe0b3998599388f3` (`release: bump apps/cli to v0.5.20`).
- Frozen candidate head: `b5d65c9f561065b8d15591660a6bef4863f56b4c`.
- Structural relocation only (not S5): SessionManager ownership and ACK/reset/replay semantics are unchanged; no DB/schema changes; no user-visible Web.
- Durable package/distribution boundaries are mirrored in Draft Context Tree PR https://github.com/agent-team-foundation/first-tree-context/pull/923; source merges first, then the Tree PR is reconciled and readied.

## Test plan
- [x] Package builds + typecheck for the three new packages and `first-tree-dev`
- [x] Focused cloud-client / runtime / provider suites + architecture guards
- [x] Full cloud-client, client-runtime, and client-providers suites
- [x] `apps/cli/tests/ci-cli-workflow.test.mjs` covers cloud-client / client-runtime / client-providers path filters
- [x] Package-metadata private-dependency guard + post-bundle dist inline scan
- [x] Real npm 11.5.1 pack and empty-consumer execution across dev, prod, and staging identities; packed and installed manifests contain no `devDependencies`, `workspace:` coordinate, private S4 package name, or unresolved internal import
- [x] Pack lifecycle recovery: mid-copy failure, post-manifest failure, stranded journal, exact source-manifest restoration
- [x] Linux arm64 portable artifact build, deterministic rebuild, and verifier
- [x] Review follow-up: monotonic markdown/HTML scanners + failing-delimiter adversarial coverage on this head
- [ ] Full-isolated QA re-qualification on this exact head after review follow-up (hand to QA)
- [x] Full-isolated QA PASS on the frozen exact head, including repository gates, runtime/provider/Managed Skills/ACK-reset-replay-persistence characterization, isolated production health, real CLI-daemon lifecycle, and aggregate CodeQL with zero new alerts
- Residual live platforms still out of scope: real OpenCode/Pi auth turns, Pi long-lived RPC, Pi host smoke, Windows supervisor/host execution, and portable platforms other than Linux arm64


baixiaohang (changes requested):
- Rationale: The replacement scanners introduced to remove polynomial regex behavior still have quadratic worst-case paths on malformed Markdown/HTML input, and the new performance tests do not exercise those paths.

**Risk level**: A

**PR summary**
- Problem: Maintainers need to evolve the Client internals without changing the single CLI that users install or exposing private workspace packages to npm/portable consumers.
- Approach: Split the former monolith into private Cloud Client, Client Runtime, and Client Providers packages; compose and inline them in `apps/cli`; inject provider-owned projections/logging at the composition edge; remove the retired package path.
- Impacted modules: `apps/cli`, the three new client packages, shared scanners, CI/release packaging, and current-state documentation.

**Review findings**
❌ 1. The new delimiter scanners still rescan the full remaining suffix for every malformed opener. In `scanInlineMarkdownLinkRanges`, an input such as `"[".repeat(N)` (or many `[` followed by `]x`) makes the inner `j` loop run from `i + 1` to the end for every outer iteration; `scanInlineMarkdownLinks` has the same shape, and `scanHtmlTagRanges` does likewise for repeated `<a` without `>`. That is Theta(N^2) work on message-controlled text, so this does not achieve the stated linear-parser security boundary. The added tests append a valid closing link/tag, which lets the first candidate succeed and advance `i` to the end, so they miss the failing-delimiter path. Please make the scanner monotonic/state-machine based (or otherwise ensure each character is visited a constant number of times) and add adversarial cases with no valid closer / an invalid closer. [R1/R5 / `packages/shared/src/lib/doc-link-scan.ts:202`, `packages/shared/src/lib/doc-link-scan.ts:236`, `packages/client-runtime/src/runtime/doc-snapshots.ts:443`]

**Action taken**
- Submitted request changes.

yuezengwu:
Superseded by #2290 after the design decision to keep a single private `packages/client` and use internal Cloud / Runtime / Providers directory layers instead of three workspace packages.

The replacement exact head has passed independent full-isolated QA and is now Ready for review. Closing this Draft unmerged so the rejected three-package direction cannot be mistaken for an active candidate.

Issue #927, opened 2026-06-10
Honor $FIRST_TREE_HOME/bin on spawned-agent PATH + warn when the briefing's CLI binary is not resolvable


yuezengwu:
## Problem

The agent briefing interpolates the channel binary name (`first-tree` / `first-tree-staging` / `first-tree-dev`) from the build-time `CHANNEL` constant via `getCliBinding().binName`, but nothing guarantees that name resolves on the PATH the spawned agent session inherits (agents inherit the daemon's PATH unchanged; the daemon itself never shells out to the CLI).

Observed failure (isolated QA run for a candidate branch): the briefing correctly named `first-tree-dev`, but the spawned real agent's PATH had no such binary. The agent silently fell back to the globally installed stale `first-tree-staging` binary — which honors the task-local `FIRST_TREE_HOME` / `FIRST_TREE_SERVER_URL` env it inherited, so it half-worked against the QA server with a stale command surface. The drift only became visible at the database level at the end of the scenario.

The dev channel has `packageName = null`; its only install path is `scripts/dev-install.sh`, which writes to the user-global `~/.local/bin` — not task-local, and unusable for parallel isolated QA runs.

## Proposal (approved by yuezengwu)

Two small primitives in the client runtime:

1. **PATH overlay**: when building the spawned agent env (`buildAgentEnv` in `packages/client/src/runtime/agent-io.ts` — shared by claude-code, claude-code-tui, and codex handlers), prepend `$FIRST_TREE_HOME/bin` to `PATH` if that directory exists. `FIRST_TREE_HOME` is already the channel/task isolation boundary; this extends it to binary resolution. Isolated environments (QA harnesses, local dev) drop channel-named shims into `<home>/bin` and they win for that daemon's agents only — no global install is touched, and daemons launched by launchd/systemd (minimal PATH) are covered without depending on who launched the daemon.

2. **Resolvability self-check**: when a session context is created, verify that `getCliBinding().binName` resolves on the PATH the agent will actually see, and log a loud warning when it does not. This turns the whole silent-fallback failure class into one visible log line before any agent turn runs.

## Non-goals

- No change to briefing generation or channel identity wiring.
- Windows PATH semantics out of scope (matches current platform support).

## Context

[...]

PR #2297, opened 2026-08-12
feat: provision the first Team Agent on demand

Gandy2025:
## Summary

This pull request implements the Agent-first first-Team lifecycle behind a single default-off rollout gate. Existing deployments keep the established personal-Team OAuth bootstrap until the complete Runtime-to-channel first-use path is available. When the gate is explicitly enabled, an authenticated account may have zero active Team memberships and creates its first Team only when it explicitly starts an Agent.

This pull request adds two first-Team entry boundaries:

- Confirming exactly one Agent Template creates the Team, caller's Admin membership and human mirror, unbound organization-visible Team Agent, and Template adoption in one transaction.
- Starting a known landing-campaign Quickstart creates the Team, caller identity, tenant-local service membership, and service-managed trial Agent in one transaction before the existing chat bootstrap continues.

The approved durable lifecycle was recorded and merged first in [first-tree-context PR #927](https://github.com/agent-team-foundation/first-tree-context/pull/927).

`FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` defaults to `false`. With the flag off, OAuth continues to create the personal Team, Team-less refresh is rejected, `POST /me/team-agents` is unavailable, and landing campaigns cannot create a Team for a Team-less caller. The campaign boundary rechecks this gate inside the user-locked transaction so membership loss cannot race into first-Team creation. Existing `onboarding.team_created` funnel telemetry is preserved.

## API and retry boundaries

`POST /api/v1/me/team-agents` is restricted to the genuine no-membership state and requires exactly one selected Template. It does not accept an organization selector and does not create Agents inside existing Teams; those remain on `POST /api/v1/orgs/:orgId/agents`.

The first-Team transaction locks the caller's stable user row. Each explicit confirmation carries a UUID v7 request identity; the server binds the canonical intent to the derived Agent identity, so only a provably exact retry can converge on the existing result. A different request, name, or Template intent returns `409` with the dedicated `first_team_agent_request_conflict` code. The server allocates a stable mentionable suffix when the Template slug collides with the new human mirror.

Existing, invited, and multi-Team members cannot use the user-scoped route as a second Team authority. Invitation-only deployments continue to reject Team-less self-provisioning.

Self-service `POST /api/v1/me/organizations` remains an additional-Team operation for users who already have an active membership. The authorization check, organization, Admin membership, and human mirror share one transaction and the same stable user-row fence as self-leave and admin removal, so concurrent final-membership loss cannot turn it into a third first-Team boundary.

When the rollout gate is enabled, refresh tokens authenticate the user independently from Team membership so the Team-less account remains signed in while choosing its first Agent. With the gate off, refresh continues to require an active membership. Team-scoped routes always authorize against live membership.

## Agent-first continuation

The atomic first-Team transaction stores the exact first Agent identity in reserved internal metadata on the caller's existing 1:1 human mirror. `/me` projects that read-only continuation with the target's current active, suspended, or deleted status. The continuation does not infer creator identity from mutable role, ownership, or `invitee_skip`.

After Template confirmation, direct `/onboarding` entry and Settings resume continue to that exact Agent's Runtime surface. A deleted target returns to Template selection; the next authorized self-created replacement retargets the same continuation inside the existing Agent-create transaction. Membership reactivation clears the stale marker in the existing lifecycle transaction because rejoin starts a fresh onboarding lifecycle.

`POST /me/onboarding-completed` is the existing terminal seam. It now stamps completion and retires the continuation in one transaction. This foundation PR deliberately does not let a normal Web or campaign Task call that seam: the locked lifecycle is `set-up-runtime -> connect-channel -> use-agent`, so the follow-up channel/Feishu first-use PR must call it only after real channel first use succeeds.

A stored campaign action is bound to the exact provisioned Team and Agent immediately after provisioning, before fallible local Team activation. Other Agent or Team detail pages retain ordinary chat behavior and cannot consume it. The handoff clears only after successful task creation.

## Team-less Web routing

When the rollout gate is enabled, an authoritative Team-less `/me` routes ordinary dashboard destinations to the public Template library without treating a transport failure as a zero-Team account. `/quickstart` preserves the known campaign OAuth return path, while `/settings/account`, the authenticated user menu, and Sign out remain reachable. Org-scoped pulse traffic, command-palette data, and Team-only Settings navigation stay disabled until an organization exists.

If another Template request wins first-Team creation, the losing tab refreshes authoritative `/me` and rejoins the ordinary existing-Team Template path. If Quickstart fails after its provisioning transaction, it refreshes `/me` before exposing a manual retry, avoiding stale Team-less state or an automatic duplicate start.

## Scope

No database schema, migration, constraint, index, backfill, data rewrite, new onboarding enum, Runtime readiness implementation, or new visual onboarding flow is included. The rollout gate is deployment config, not persisted product state. The approved continuation semantic reuses reserved internal metadata on the existing human-mirror Agent; all organization, membership, Agent, Resource, Runtime detail, and campaign chat persistence remains on existing tables.

Deliberately excluded: new visual onboarding pages, Runtime readiness/state-machine work, Feishu implementation, Team naming UI, visibility selection, Team usage authorization UI, old-onboarding redesign, and migration of existing users.

## Verification

Exact head `64f8e1db49453bd5fd47bd585a3e50109e57d7b9`:

- Focused Server gate, OAuth, continuation, and campaign coverage: 6 files / 143 tests passed.
- Focused Shared config: 1 file / 24 tests passed.
- Focused Web continuation and recovery: 3 files / 99 tests passed.
- `pnpm typecheck`: 9/9 tasks passed.
- `pnpm build`: 5/5 tasks passed (existing CSS and chunk-size warnings only).
- Exact changed-file Biome and `git diff --check` passed.
- A full monorepo test attempt under concurrent package load hit timeouts in four unrelated suites. Each affected suite passed immediately in isolation: Web app routes 27/27, Client OpenCode handler 36/36, CLI runtime layout 5/5, and Context Tree audit cases 2/2.

Independent Standards and Spec re-reviews report zero blockers on this exact head. All GitHub CI checks pass. Independent, fully isolated QA reports PASS on the exact head across the production image, real PostgreSQL boundaries, full Server/Web/CLI suites, default-off and explicitly enabled behavior, atomic provisioning, recovery, and membership-race coverage. The QA result qualifies only the default-off foundation: the rollout flag must remain off until the follow-up Runtime and Feishu first-use path can consume the terminal completion seam.


baixiaohang (changes requested):
- Rationale: The implementation changes the canonical onboarding/Team-creation contract without the corresponding durable-decision update, and the new user-scoped route also performs existing-Team Agent creation that belongs to the established org-scoped surface.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the new `/me` API establishes core Agent lifecycle state and the Web change adds a new data/routing path -> B-high.

**PR summary**
- Problem: A new signer-in currently receives an empty Team before choosing an Agent, leaving unused Teams and asking for Team intent twice.
- Approach: Keep solo accounts Team-less after OAuth, then atomically create the Team, Admin membership, human mirror, unbound organization-visible Agent, and adopted Template when the user confirms the first Agent; route Team-less Web sessions through the Template library.
- Impacted modules: OAuth bootstrap and callbacks, `/me` API and Team/Agent services, shared DTOs, Web auth routing, and Template intent.

**Review findings**
❌ 1. This PR changes a current durable onboarding decision but explicitly defers the Tree reconciliation. The normal Cloud onboarding node still defines the admin path as standalone `/onboarding` with `create-team -> connect-computer -> create-agent -> start-chat`, while Team tenancy says the first Team is minted at sign-in. This PR instead makes a Template confirmation outside that flow create an unbound Agent and the Team together. That is a product/lifecycle decision, not implementation-only drift; obtain the owning decision and update the authoritative Tree in the same change sequence before this source behavior can merge. [R5 / `packages/server/src/services/auth/oauth/bootstrap.ts:107`; [Cloud Onboarding](https://github.com/agent-team-foundation/first-tree-context/blob/0897318e8b7ea5f3ef0abb196c9b6c4ef6f6ae8c/system/cloud/onboarding.md#L35-L47); [Team Tenancy and Identity](https://github.com/agent-team-foundation/first-tree-context/blob/0897318e8b7ea5f3ef0abb196c9b6c4ef6f6ae8c/system/cloud/team/tenancy-and-identity.md#L60-L75)]
❌ 2. `/me/team-agents` is not Class A "by construction" once `organizationId` is accepted and existing/default memberships are resolved: those branches create an Agent inside an existing Team, which the HTTP path contract classifies as Class B and already exposes as `POST /orgs/:orgId/agents`. Keep this `/me` operation restricted to the genuine no-membership atomic provision, and use the existing org-scoped create for invited/multi-Team members; otherwise Team scope has two parallel authorities and is hidden in the body/default selector. [R1, R2 / `packages/shared/src/schemas/me-extras.ts:207`; `packages/server/src/services/team/first-team-agent.ts:151`; `docs/development/http-path-conventions.md:23`]
✅ 3. Within the proposed lifecycle, the user-row serialization point, single transaction, rollback behavior, and retry convergence are coherently implemented and covered by focused cases.

**Action taken**
- Submitted request changes.
- Tests were not run under review policy; GitHub currently reports `Lint & Type Check` as failing while the test jobs pass.

yuezengwu (commented):
Additional review findings on head `893ed7eaca593799a118b38eb54a80a6307eef08`.

The implementation goal is clear: defer Team creation until the first Team Agent confirmation, then commit Team, Admin membership, human mirror, unbound org-visible Agent, and Template adoption atomically. The user-row lock and rollback boundary are coherent. However, I found three additional behavior blockers beyond the already-submitted Tree/API-scope review:

1. **[P1] Preserve the new-user Quickstart acquisition path.** `RequireAuth` now redirects every authoritative Team-less session at `packages/web/src/auth/require-auth.tsx:117`, before `/quickstart` can mount. That directly contradicts the existing handoff on lines 98–101 and the server's `shouldPreserveSoloSignupNext`, both of which intentionally return a new signer to the exact campaign URL. After this PR, the OAuth callback adopts tokens, `/me` reports no memberships, and the user is sent to `/templates`; `QuickstartPage` never calls `POST /me/landing-campaigns/start`. Merely exempting the route is insufficient because `resolveCallerMembership` currently throws `Active membership not found`, so the Team-less campaign path needs an explicit provisioning/selection decision plus an end-to-end regression case.

2. **[P1] Keep a user-scoped account/sign-out escape hatch for Team-less users.** The blanket redirect also makes `/settings/account` unreachable even though that page is user-scoped, and the `UserMenu` containing Sign out lives behind the same guard. `TemplateLibraryPage` exposes only a logo link back to `/`, which redirects straight back to `/templates`. A user who declines Agent creation—or encounters an empty/failing catalog—has no supported UI to sign out, change their profile, or manage sign-in methods. Route Account outside the org-only shell or add equivalent authenticated account controls to the Team-less entry.

3. **[P2] Do not treat different creation payloads as the same retry.** `packages/server/src/services/team/first-team-agent.ts:90-98` returns any existing managed Agent without checking whether it represents the requested `templateIds`/name. Two Template tabs can submit different responsibilities concurrently: the loser receives `200 agentCreated=false` for the winner's Agent, while `TemplateUseIntent` ignores that flag and opens the Agent as if its selected Template had been adopted. The second intent is silently discarded. Convergence is safe only for the same logical request; persist/compare an idempotency identity or return a conflict/result the client can reconcile, and cover concurrent different-Template submissions.

This is also a persistence-semantic change despite having no migration: authenticated `users` may now have zero memberships, and the Team/membership/Agent creation boundary moves to a later transaction. That invariant should be called out explicitly for the required data-model review.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; GitHub currently reports the test jobs passing and `Lint & Type Check` failing.

yuezengwu (commented):
Re-reviewed synchronized head `be9098a82670de2da18ca28e4c54acf1e0508e58`.

The only change since the prior review is import ordering in `packages/server/src/api/me.ts` to satisfy Biome. It does not touch the onboarding/Team-creation contract, `/me/team-agents` scope, Team-less Quickstart and account routing, or differing-payload idempotency behavior.

Accordingly, all findings in the existing request-changes review and my [additional review](https://github.com/agent-team-foundation/first-tree/pull/2297#pullrequestreview-4916431049) remain applicable. Recommendation remains **request changes**.

Per review policy, I did not run tests. `git diff --check` remains clean; the new CI run is still in progress and was not awaited.

Gandy2025:
Review findings are addressed on `346c82607`.

- Durable lifecycle: the approved decision is now recorded in first-tree-context PR #927. This source PR is back in draft and must not merge before that Tree update.
- HTTP scope: `POST /me/team-agents` no longer accepts or resolves an organization. It is restricted to callers with zero active memberships; existing-Team Agent creation remains on `POST /orgs/:orgId/agents`.
- Quickstart: known Team-less landing-campaign starts now create the Team, caller/service memberships, human mirror, and service-managed trial Agent in one transaction. The existing OAuth return path is preserved, and `action=fix` retains its handoff through Template confirmation before resuming onboarding.
- Account escape hatch: `/settings/account` remains reachable for Team-less accounts, the Template library exposes the authenticated user menu and Sign out, Team-only Settings navigation is hidden, and org-scoped pulse traffic is disabled until a Team exists.
- Retry semantics: each explicit confirmation carries a UUID v7 request identity. The server derives the first Agent identity from the caller and request, so only a provably exact request/payload can converge. Different requests or payloads return `409`; an unrelated existing Agent cannot be mistaken for a retry.

Regression coverage includes concurrent same-request convergence, concurrent different-request conflict, rollback, invited/multi-Team rejection, Team-less Quickstart provisioning, account/sign-out routing, and the `action=fix` handoff.

Local verification on this head:

- server: 278 files / 3285 tests passed
- web: 247 files / 2328 tests passed
- shared: 75 files / 912 tests passed
- QA contract: 1 file / 4 tests passed
- `pnpm check`, `pnpm typecheck` (9/9), `pnpm build` (5/5), exact changed-file Biome, and `git diff --check` passed

Please re-review this head. CI and independent QA are now running.

yuezengwu (changes requested):
Re-reviewed head `346c826070b79224ccb6085c219e907f36b1d8fd`.

The implementation goal is now broader and clearer: solo OAuth leaves an authenticated account with zero active Team memberships; an explicit Template confirmation atomically creates the first Team, Admin membership, human mirror, unbound Team Agent, and Template adoption; a known Quickstart campaign can instead atomically create the Team, caller/service identities, and hosted trial Agent before continuing the existing chat bootstrap.

The prior API-scope, Quickstart routing, and account/sign-out findings are substantially addressed. `/me/team-agents` is now limited to the no-membership boundary, and its derived Agent UUID plus exact payload comparison prevents a different Template intent from being reported as a successful retry. The user-row lock and transaction boundaries remain coherent.

Two implementation findings and one merge prerequisite remain:

1. **[P1] Reconcile the new different-request conflict in the Template UI.** The server now correctly returns `409` when another first-Team request won, but `TemplateUseIntent` catches every failure identically, refreshes `/me`, then keeps the page in the `provisionError` branch with “Nothing was created — try again” (`packages/web/src/pages/templates/template-use-intent.tsx:203-210,233-238`). In the two-Template-tab case, the refresh reveals the winner's new membership, but retrying the loser's unchanged request is guaranteed to return `409` forever. This leaves the advertised retry dead-ended and still does not perform the client reconciliation requested in the prior finding. Distinguish the conflict (the API exposes `ApiError.status`), then use the authoritative `/me` result to resume the ordinary existing-Team Template path or show a truthful conflict destination; cover that race in the Web test.

2. **[P2] Gate the command palette's org roster on a selected Team.** `/settings/account` is now reachable without a Team, but it mounts the ordinary `Layout`, which mounts `CommandPalette` even while closed (`packages/web/src/components/layout.tsx:428`). `CommandPalette` immediately enables `useOrgAgents` (`packages/web/src/pages/workspace/palette/command-palette.tsx:81`), whose query calls `withOrg` and throws when `organizationId` is null. React Query contains the exception, so Account still renders, but the supposedly user-scoped escape hatch starts in a failed org-query state. Disable that query/palette until an organization exists and add the Team-less Layout regression case.

3. **The locked durable decision and persistence-semantic approval are still prerequisites.** `first-tree-context#927` contains the proposed lifecycle update, but it is currently open with `REVIEW_REQUIRED`, has no reviews, and is not merged; describing it as “approved” is premature. The current Cloud node still has `decisionLocksCode: true`, so this source behavior cannot be approved or merged ahead of the owning Tree decision. Separately, allowing authenticated users to have zero active memberships and moving Team/membership/Agent creation to later transactions is a persistence-semantic change even without a schema or migration; repository `AGENTS.md` requires explicit human approval for that specific change.

Recommendation remains **request changes**. Keep this source PR draft until the two code findings are resolved, the persistence semantic has explicit human approval, and the authoritative Tree update is reviewed and merged first.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean. At review time, lint/type check and the completed CI jobs pass; server and client/Web test jobs are still running and were not awaited.

baixiaohang (changes requested):
- Rationale: The previous lifecycle/API/routing/idempotency findings are substantially addressed, but a failed Team-less Quickstart can leave `/me` stale after the server has already committed the Team.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: New signers should not receive an empty Team before choosing an Agent, while Team-less users must still retain Quickstart, Account, and sign-out paths.
- Approach: Defer Team creation until an explicit Template or known campaign action, provision the required identities atomically, use exact request identities for first-Agent retries, and keep existing-Team Agent creation org-scoped.
- Impacted modules: OAuth bootstrap/callbacks, first-Team and landing-campaign services, shared DTOs, Web auth/layout/Quickstart/Template flows, and QA contracts.

**Review findings**

❌ 1. **[P1] Refresh `/me` when Team-less Quickstart fails after provisioning.** `provisionTrialAgent` commits the new Team, caller/service memberships, and trial Agent before resource binding and chat bootstrap run (`start.ts:587-600`). A deterministic example is an over-quota caller: `assertTrialQuota` rejects later at `start.ts:466`, after the Team transaction has committed. `QuickstartPage` refreshes `/me` only on success and its catch only records an error (`quickstart-page.tsx:107-120`), so Auth continues to report `hasNoTeam=true`; navigating to Templates can then offer first-Team creation even though the server will reject it because the membership already exists. Re-read `/me` on this ambiguous failure path (as the Template provision flow already does) and cover a Team-less post-provision failure. [R4, R5 / `packages/server/src/services/landing-campaigns/start.ts:587`; `packages/web/src/pages/quickstart/quickstart-page.tsx:107`]

✅ 2. The earlier `/me/team-agents` scope overlap, Team-less route reachability, Account/sign-out escape hatch, and server-side different-payload retry convergence are addressed on this head. The lifecycle decision is now staged in `first-tree-context#927`; this source PR correctly remains draft and must not merge before that authoritative update is accepted. The separate current review already captures the remaining Template conflict reconciliation, Team-less Command Palette query, and durable-decision prerequisites, so they are not repeated here.

**Action taken**
- Submitted request changes.
- Tests were not run under review policy. GitHub CI is green on head `346c826070b79224ccb6085c219e907f36b1d8fd`.

Gandy2025:
The remaining findings are addressed on `2e79d6496600f6047ef185346a9020e2763da3b9`.

- Durable lifecycle: first-tree-context PR #927 was independently approved and merged before this source update. The product owner explicitly approved the zero-membership and first-Team persistence lifecycle in the originating task.
- Template conflict reconciliation: a losing different-request `409` now refreshes authoritative `/me`, adopts the winner's selected Team in local state, and resumes the ordinary existing-Team Template path. It no longer offers the permanently conflicting first-Team retry.
- Team-less Layout: the org command palette, keyboard shortcut, and roster query do not mount until an organization is selected. Account and Sign out remain reachable.
- Quickstart ambiguity: a Team-less failure after the provisioning boundary refreshes `/me`, preserves the campaign intent, and requires an explicit manual retry. The refresh cannot trigger an automatic duplicate start.
- Empty-Team authority: `POST /me/organizations` now requires an existing active membership, so it remains an additional-Team operation and cannot create an empty first Team.

Regression coverage includes the two-Template-tab conflict handoff (including the winner Team being selected and actionable), Team-less Layout with no palette, post-provision Quickstart failure reconciliation, no automatic retry, and rejection of Team-less self-service organization creation with no persisted Team.

Local verification on this exact head:

- server: 278 files / 3286 tests passed
- web: 247 files / 2331 tests passed
- shared: 75 files / 912 tests passed
- QA contract: 1 file / 4 tests passed
- `pnpm check`, `pnpm typecheck` (9/9), `pnpm build` (5/5), exact changed-file Biome, and `git diff --check` passed

The PR remains draft while CI, re-review, and independent QA run. Please re-review this head.

yuezengwu (changes requested):
Re-reviewed head `2e79d6496600f6047ef185346a9020e2763da3b9`.

The intended lifecycle is now durably settled: solo authentication may leave an account with zero memberships, and the first Team is created only by an explicit Template or known-campaign Agent start. `first-tree-context#927` was approved and merged before this head, and the product owner has explicitly approved the persistence semantic. This head also fixes the previously reported Template-race loop for the intended different-request case, disables the Team-less command palette, refreshes `/me` after an ambiguous Quickstart failure, and prevents automatic Quickstart retry after reconciliation.

Two blockers remain in the newly added recovery/authority logic:

1. **[P1] Do not classify every `409` as a winning first-Team request.** `TemplateUseIntent` treats any `ApiError` with status 409 as proof that another request created the Team (`packages/web/src/pages/templates/template-use-intent.tsx:219-227`). The same atomic create can return other 409s while rolling the entire Team back. A deterministic case is a user whose normalized username equals the Template slug: `createPersonalTeam` gives the human mirror that Agent name, then the non-human Agent insert hits the org+name unique constraint and returns `code: "agent_name_conflict"` (`packages/server/src/services/agents/identity.ts:612-620`). `/me` correctly remains Team-less, but the Web now says another request won and offers only “Check Team,” which can never find one. Classify the exact first-Team race with a dedicated machine-readable code (or verify that `/me` actually gained a membership) and preserve ordinary creation errors such as `agent_name_conflict`; cover the username/Template collision.

2. **[P2] Make additional-Team authority transactional with creation.** The new `/me/organizations` guard reads active memberships before the user lookup and before `selfCreateOrganization` (`packages/server/src/api/me.ts:705-725`), while `selfCreateOrganization` inserts the organization and membership in later operations (`packages/server/src/services/team/membership.ts:787-800`). If the user's last active membership is left or removed after line 708 but before creation, this stale authorization still creates an empty new Team from a now-Team-less state—the third first-Team boundary that the merged tenancy decision explicitly forbids. Re-check/lock the authorizing active membership in the same transaction as organization + Admin membership + human mirror creation, using a lock shared with leave/removal, and add the concurrent last-membership-loss case.

Recommendation remains **request changes**. The prior three implementation findings and the durable/persistence approvals are resolved; the review is now limited to these two new boundary defects.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean. At review time, lint/type check and all completed CI jobs pass; server and client/Web test jobs are still running and were not awaited.

baixiaohang (changes requested):
- Rationale: The prior lifecycle/routing findings are resolved, but the current recovery and additional-Team guard still leave two first-Team authority defects.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: New signers should not receive an empty Team before explicitly choosing an Agent, while Team-less users must retain valid Quickstart, Account, and recovery paths.
- Approach: Defer the first Team to Template or campaign confirmation, create its required identities atomically, preserve exact retry semantics, and keep existing-Team creation authorities scoped to an active membership.
- Impacted modules: OAuth bootstrap/callbacks, first-Team and landing-campaign services, self-service Team routes, shared DTOs, Web auth/layout/Quickstart/Template flows, and QA contracts.

**Review findings**

❌ 1. I independently confirmed the two blockers documented in the current [head review](https://github.com/agent-team-foundation/first-tree/pull/2297#pullrequestreview-4917585641): the Template UI interprets unrelated rollback-producing `409`s as proof that another request created a Team, and the additional-Team active-membership check is outside the creation transaction and can race the last membership's deactivation. Both can violate the merged first-Team boundary; I found no useful additional finding to duplicate here. [R4, R5 / `packages/web/src/pages/templates/template-use-intent.tsx:219`; `packages/server/src/api/me.ts:705`]

✅ 2. The previous Template winner handoff, Team-less command-palette gate, Quickstart `/me` refresh/manual-retry boundary, and declared additional-Team-only route behavior are otherwise implemented on this head. The owning lifecycle decision was independently approved and merged first in `first-tree-context#927`, satisfying the durable-decision and explicit persistence-semantic prerequisite.

**Action taken**
- Submitted request changes on head `2e79d6496600f6047ef185346a9020e2763da3b9`.
- Tests were not run under review policy. `git diff --check` is clean; lint/type check and completed CI jobs pass, while Server and Client/Web tests are still running at review time.

baixiaohang (changes requested):
- Rationale: The synchronized head moves the additional-Team check into the service transaction, but it still does not share a lock with membership loss, and the unrelated-409 recovery defect is unchanged.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: New signers should not receive an empty Team before explicitly starting an Agent, while Team-less users must retain valid Template, Quickstart, Account, and recovery paths.
- Approach: Defer the first Team to Template or campaign confirmation, create its required identities atomically, preserve exact retry semantics, and keep additional-Team creation authorized by an existing active membership.
- Impacted modules: OAuth bootstrap/callbacks, first-Team and landing-campaign services, membership and self-service Team creation, shared DTOs, Web auth/layout/Quickstart/Template flows, and QA contracts.

**Review findings**

❌ 1. **[P1] The Template client still interprets every `409` as a successful competing first-Team request.** This head does not change `TemplateUseIntent`: line 220 still branches only on `ApiError.status === 409`. A rollback-producing conflict such as `agent_name_conflict` therefore still sets `provisionConflict`, returns after `/me` remains Team-less, and offers the user a Team destination that cannot exist. Please distinguish the first-Team race with a dedicated code or require the refreshed `/me` to contain a membership, and keep ordinary creation conflicts on a truthful recoverable path; retain the username/Template-name collision regression. [R4, R5 / `packages/web/src/pages/templates/template-use-intent.tsx:220`]

❌ 2. **[P2] The additional-Team prerequisite still is not fenced against leave/removal.** `selfCreateOrganization` now holds `users` `FOR NO KEY UPDATE`, but its active-membership read is unlocked (`membership.ts:800-804`), while `deactivateMembership` and admin `deleteMember` lock member rows without acquiring that user-row lock. A last-membership leave/removal can therefore commit after the service's check and before its new Team graph is created. The new sequential test only covers a caller who was already inactive and cannot exercise this window. Lock the authorizing membership (or another fence actually shared by both lifecycle paths) through the organization + Admin membership + mirror transaction, and cover concurrent last-membership loss. [R4, R5 / `packages/server/src/services/team/membership.ts:800`; `packages/server/src/services/team/membership.ts:377`; `packages/server/src/services/team/member.ts:285`]

✅ 3. The prerequisite is now owned by `selfCreateOrganization`, and organization, Admin membership, and human mirror creation are in one transaction. That closes the route-bypass and partial-write cases, but not the two blockers above.

**Action taken**
- Submitted request changes on head `96dbdd47640b33a23b7de35b44326bc39ae082ff`.
- Tests were not run under review policy. At review time, lint/type check and completed CI jobs pass; server, client/Web, CLI, and JavaScript CodeQL jobs are still running.

yuezengwu (changes requested):
Re-reviewed synchronized head `96dbdd47640b33a23b7de35b44326bc39ae082ff`.

This commit correctly moves the additional-Team prerequisite into `selfCreateOrganization` and wraps the prerequisite, organization, Admin membership, and human mirror in one transaction. However, it does not resolve either blocker from my review on the preceding head:

1. `TemplateUseIntent` is unchanged and still treats every HTTP `409` as proof that another first-Team request won. Rollback-producing conflicts such as `agent_name_conflict` still leave the user Team-less but route the page into the impossible “Check Team” recovery state.

2. The active-membership read inside `selfCreateOrganization` is still unlocked, and the user-row lock taken through `syncCurrentUserDisplayName` is not shared by `deactivateMembership` or admin removal. Last-membership loss can therefore commit after the prerequisite read and before the new Team graph is inserted. The new sequential inactive-caller test does not exercise this interleaving.

The full rationale and requested fixes remain in [the prior head review](https://github.com/agent-team-foundation/first-tree/pull/2297#pullrequestreview-4917585641). Recommendation remains **request changes**.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean.

Gandy2025:
The two remaining blockers are addressed on exact head `b7c6aef96154b279d5823cdf22fbfad386ef10c7`.

- First-Team conflict classification now has a dedicated shared code, `first_team_agent_request_conflict`. `TemplateUseIntent` branches only on that code. A rollback-producing `agent_name_conflict` remains on the truthful ordinary retry path; server and Web regressions cover the username/Agent-name collision and verify that no Team or membership survives.
- Additional-Team authority now shares the stable `users` row lock across creation, self-leave, and admin removal. `selfCreateOrganization` rechecks active membership and creates the organization, Admin membership, and human mirror in one transaction. Concurrent final-membership leave and removal cases both prove that membership loss wins before the queued create, which then returns conflict without an empty Team.
- The synchronized reviewer-authored commit `96dbdd476` was retained as the transactional service baseline; this head adds the missing shared lifecycle fence and exact client/server conflict contract on top.
- First-Team Template and landing-campaign transactions now call the same lifecycle-lock helper, so every supported first/additional Team boundary uses one serialization point.

Focused verification on the exact head passes: server 5 files / 117 tests, Web Template intent 1 file / 21 tests, root typecheck 9/9, exact changed-file Biome, and `git diff --check`. Full exact-head Server CI/local verification and independent QA are in progress. The PR remains draft while those checks complete.

Please re-review this head.

yuezengwu (approved):
Re-reviewed exact head `b7c6aef96154b279d5823cdf22fbfad386ef10c7`.

The implementation now consistently enforces the approved lifecycle: authentication alone may leave a user with zero active Team memberships; an explicit Template or known-campaign Agent start owns first-Team creation; and self-service organization creation remains an additional-Team operation.

The two remaining blockers are resolved:

- First-Team request races now use the dedicated shared `first_team_agent_request_conflict` code, and `TemplateUseIntent` branches only on that identity. Atomic creation conflicts such as `agent_name_conflict` remain on the ordinary recoverable error path, with server rollback and Web behavior covered by focused regressions.
- Additional-Team authorization, self-leave, and admin removal now share the stable user-row lifecycle lock before membership mutations. The active-membership recheck and new Team graph stay in one transaction, and the concurrent final-membership leave/removal cases exercise the queued interleavings that previously exposed the TOCTOU window.

The first-Team Template and Quickstart paths also use the same lifecycle-lock helper, preserving one serialization point across the supported Team-creation boundaries. The locked Context Tree decision is merged, and the product owner has explicitly accepted the zero-membership persistence semantic.

I found no remaining blocking issue and approve this head. Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

Gandy2025:
Exact-head verification is complete on `b7c6aef96154b279d5823cdf22fbfad386ef10c7`.

- Local full Server suite: 278 files / 3290 tests passed.
- GitHub CI: all required jobs passed, including Server, Client & Web, both CLI shards, lint/typecheck/build, migrations, portable/release smoke, and CodeQL.
- Focused race regressions: same-request convergence, different first-Team request conflict, rollback-producing name conflict, concurrent final-membership self-leave, and concurrent admin removal all pass.
- Local exact-head typecheck, changed-file Biome, and `git diff --check` pass; the broader Web/shared/QA-contract/skill-eval suites listed in the PR description also passed.

Independent QA is still running. The PR remains draft pending that result and the outstanding requested re-review from @baixiaohang.

baixiaohang (changes requested):
- Rationale: The new head fixes the immediate post-confirmation bounce, but it persists the invitee-only skip reason for a Team-less creator and sends that incomplete state back to the legacy Team-creation journey when resumed.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: New signers should enter through an explicit Agent choice without receiving an empty Team or falling into the legacy Team-first onboarding flow.
- Approach: Atomically create the first Team and Agent from a Template or known campaign, preserve Team-less account/Quickstart recovery, and serialize every first/additional-Team membership boundary on the stable user row.
- Impacted modules: first-Team and campaign services, membership lifecycle, shared DTOs, Web auth/onboarding/Template flows, and cross-surface QA contracts.

**Review findings**

❌ 1. **[P1] Do not persist the Team-less creator as `invitee_skip`.** The new write at `first-team-agent.ts:156-162` reuses a reason whose durable meaning is “a joining member skipped personal setup to use a teammate's Team Agent.” This caller instead just created and manages their own first Agent, and the approved Agent-first path says they continue through its own Runtime/channel progression. The mismatch is observable: Settings treats any suppressed-but-incomplete membership as “Resume setup” (`setup.tsx:491,588-595`), clears the suppressor, and navigates to `/onboarding`; because this membership is Admin, a fresh entry starts at `create-team` (`steps.ts:93-103`). The new QA assertion only proves the legacy wizard does not auto-open immediately—it remains the advertised recovery path. Give Agent-first continuation its own truthful state/reader and route resumption to that progression; if this adds a durable suppress reason, reconcile and approve that persistence semantic instead of overloading `invitee_skip`. [R1, R5 / `packages/server/src/services/team/first-team-agent.ts:156`; `packages/web/src/pages/settings/setup.tsx:491`; `packages/web/src/pages/onboarding/steps.ts:93`; [Cloud Onboarding](https://github.com/agent-team-foundation/first-tree-context/blob/bfa5695fef74228e19d71ecbb0fa5535115bfe15/system/cloud/onboarding.md)]

✅ 2. The payload-bound derived Agent identity makes exact retry recognition independent of mutable Agent fields, and server-side name allocation removes the deterministic human-mirror collision without weakening different-intent conflicts.

✅ 3. The prior dedicated conflict code and shared membership lifecycle fence remain intact in this increment.

**Action taken**
- Submitted request changes on head `ecb5a7475f96814610f1301f82beced458a54f6e`.
- Tests were not run under review policy. The synchronized head's CI is still running and was not awaited.

yuezengwu (changes requested):
Re-reviewed synchronized head `ecb5a7475f96814610f1301f82beced458a54f6e`.

The increment has a clear goal: keep the Team-less creator on the Agent-first path after the atomic Team + first Agent transaction, allocate a mentionable name when the human mirror collides, and make exact retry identity independent of later Agent mutations. The payload-bound derived UUID and server-side name allocation are coherent, and the prior conflict-code and membership-lock fixes remain intact.

One lifecycle blocker remains:

**[P1] Route every first-Agent continuation away from the legacy `create-team` journey.** The new write at `packages/server/src/services/team/first-team-agent.ts:156-162` suppresses the immediate auto-open by storing `invitee_skip`, but that value still means an invitee skipped personal setup in favor of a teammate's Agent. It does not identify a Team-less creator who owns an unbound first Agent. The mismatch is directly observable in the campaign-action branch: after the same successful provisioning, `TemplateUseIntent` explicitly navigates to `/onboarding` at `packages/web/src/pages/templates/template-use-intent.tsx:248-250`; because the caller is Admin and has no persisted Agent-first step, a fresh standalone flow starts at `create-team` (`packages/web/src/pages/onboarding/steps.ts:89-103`). This contradicts both the locked Agent-first sequence (no separate Team-naming step) and this head's own QA rule that a post-confirmation redirect into legacy Team naming is `FAIL` (`packages/qa/cases/cross-surface/registration-first-run-onboarding.md:99-105`). The same state also makes Settings advertise the legacy “Resume setup” destination, as the parallel review notes. Give this cohort a truthful Agent-first continuation state/reader and make ordinary resume plus `action=fix` enter its Runtime/channel progression; add a route-level regression proving neither path mounts `StepTeam`. If the fix introduces a new persisted suppress reason, it is a new persistence semantic and needs explicit human approval rather than overloading `invitee_skip`.

Recommendation remains **request changes**. Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

Gandy2025:
The QA main-path failure and the two final contract gaps are addressed on exact head `b8c13ee087bef0965204c54c45d73465ff4713bb`.

- A successful Team-less Template confirmation now continues directly to the existing Runtime surface for the exact new Agent. It never traverses the workspace root or standalone Admin onboarding.
- The existing `invitee_skip` suppressor remains an auto-open marker only. When the current membership already manages the confirmed Agent, Settings no longer advertises the legacy Resume action and direct `/onboarding` entry returns to the workspace. Invitees without a personal Agent retain the existing resumable path. No new persistence field or reason was added.
- A stored campaign action survives Runtime setup and is consumed by the existing Agent-header Start chat action for that exact Agent; failures retain the handoff for retry.
- `POST /me/team-agents` now requires exactly one selected Template. Omitted, empty, or multi-Template payloads are rejected before any Team graph is written, so Template confirmation and known Quickstart remain the only first-Team boundaries.
- The username/Template-slug collision remains fixed by stable server-side name allocation inside the atomic transaction.

Red/green regressions cover the direct and Settings legacy-onboarding entries, the exact-one-Template boundary with zero residual membership, the Runtime destination, campaign action consumption/retry, stable name allocation, and the public `/me` suppression projection.

Focused verification on this head passes: Server 17 tests, Shared 10 tests, Web 130 tests, root typecheck 9/9, exact changed-file Biome, and `git diff --check`. Full GitHub CI and independent exact-head QA are running. The PR remains draft pending both.

yuezengwu (changes requested):
Re-reviewed exact head `b8c13ee087bef0965204c54c45d73465ff4713bb`.

The increment tries to keep the Team-less creator on the approved Agent-first path: it sends the successful confirmation directly to the new Agent's Runtime tab, prevents legacy onboarding from mounting for the inferred creator cohort, preserves the campaign action until chat creation, and tightens the user-scoped endpoint to exactly one Template. The exact-one-Template boundary and immediate Runtime destination are coherent, and the earlier idempotency, conflict-code, and membership-lock fixes remain intact.

Two continuation blockers and one persistence prerequisite remain:

1. **[P1] Preserve a real resumable Agent-first destination instead of removing Resume.** `OnboardingPage` now identifies `invitee_skip + currentOrgHasPersonalAgent` and redirects direct `/onboarding` entry to `/` (`packages/web/src/pages/onboarding/onboarding-page.tsx:50-64`). Settings excludes the same cohort from `resumeSetup`, then renders the generic `Available / View -> /team` action (`packages/web/src/pages/settings/setup.tsx:492-505,589-606`). This prevents `StepTeam` from mounting, but it does not resume the incomplete `set-up-runtime -> connect-channel` progression required by the locked Agent-first lifecycle; after leaving or reloading away from the initial Runtime URL, the advertised setup summary no longer has a continuation at all. Route direct entry and Settings resume to the creator's Agent Runtime/channel continuation, and cover the destination—not only the absence of `StepTeam`. [R5 / [Cloud Onboarding](https://github.com/agent-team-foundation/first-tree-context/blob/bfa5695fef74228e19d71ecbb0fa5535115bfe15/system/cloud/onboarding.md)]

2. **[P1] Bind the stored campaign action to the provisioned Agent and Team before exposing it on Agent detail.** The handoff remains one unscoped sessionStorage value (`packages/web/src/utils/onboarding-flags.ts:211-234`). Every Agent detail page reads it and creates the campaign task with whichever `agent.uuid` is currently open (`packages/web/src/pages/agent-detail.tsx:445-474`). The user can switch Agents or Teams from the same detail shell before clicking Start chat, so another Agent can consume and clear an action that this flow says belongs to the exact newly provisioned Agent. The new test proves only the intended page, not the cross-Agent/Team negative case. Bind or validate the handoff against `result.organizationId` and `result.agent.uuid`, keep unrelated Agent headers on ordinary chat behavior, and cover that scope boundary.

3. **Persistence-semantic approval is still required if `invitee_skip` is retained for creators.** This head adds no column or enum value, but it changes the persisted reason's meaning from an invitee Team-Agent skip to two distinct cohorts and infers which one it was from mutable current Agent ownership. Repository policy requires explicit human approval for that specific persistence-semantic change; the earlier approval covered zero-membership and deferred first-Team creation, not this new overloaded meaning. A dedicated truthful state is preferable because it survives later Agent lifecycle changes without reclassifying the membership.

Recommendation remains **request changes**. Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

baixiaohang (changes requested):
- Rationale: The immediate legacy-onboarding bounce is fixed, but the recovery state still conflates creators with invitees, and the stored campaign action is not bound to the Agent it is meant to continue on.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: A Team-less signer should create a Team only through an explicit Agent start, then continue that exact Agent's setup without falling into the legacy Team-first onboarding journey.
- Approach: Atomically provision the first Team and Agent, route confirmation to the Agent Runtime surface, preserve campaign intent across Runtime setup, and keep additional-Team authority behind active membership.
- Impacted modules: OAuth bootstrap, first-Team and campaign services, membership lifecycle, shared DTOs, Web auth/onboarding/Agent/Template routing, and QA contracts.

**Review findings**

❌ 1. **[P1] `invitee_skip + hasPersonalAgent` is not a creator identity.** A real invitee can quick-start with a teammate's Agent (persisting `invitee_skip`) and later create or become manager of a personal Agent through the ordinary org-scoped surface; nothing clears the suppress reason or writes completion. At that point `OnboardingPage` and Settings classify them as an Agent-first creator, redirect direct `/onboarding`, and hide the resumable personal setup that the locked onboarding contract explicitly preserves for `invitee_skip`. The current regression only constructs the creator-shaped facts and cannot distinguish this invitee state. Keep `invitee_skip` invitee-only and introduce a truthful, non-ambiguous creator continuation state/transition; any changed persisted-reason semantic still needs its own explicit approval and authoritative Tree reconciliation. [R1, R5 / `packages/web/src/pages/onboarding/onboarding-page.tsx:51`; `packages/web/src/pages/settings/setup.tsx:492`; `system/cloud/onboarding.md`]

❌ 2. **[P1] Bind the campaign handoff to the provisioned Agent before exposing it on Agent detail.** `StoredCampaignActionHandoff` contains no Agent identity, and every Agent header reads it unconditionally. After provisioning, the user can use the Agent switcher/direct URL, finish a different Agent's Runtime, and press Start chat; `openChat` then targets that current `agent.uuid` and clears the global handoff, silently running the requested action on the wrong Agent. Store the provision result's Agent UUID with the handoff (while preserving legacy parsing as needed), consume it only on an exact match, and cover switching to another Agent before Start chat. [R4, R5 / `packages/web/src/utils/onboarding-flags.ts:163`; `packages/web/src/pages/agent-detail.tsx:445`; `packages/web/src/pages/templates/template-use-intent.tsx:247`]

✅ 3. The direct Runtime destination, exact-one-Template input boundary, stable server-side name allocation, dedicated conflict code, and shared membership lifecycle fence remain coherent in this increment.

**Action taken**
- Submitted request changes on exact head `b8c13ee087bef0965204c54c45d73465ff4713bb`.
- Tests were not run under review policy; CI and independent QA are still running.

Gandy2025:
The remaining Agent-first recovery gaps are addressed on exact head `65d6d89160b3771a31f54477be1e374bb9037c13`.

- Team-less Template confirmation now always continues to the exact Agent's existing Runtime surface, never the legacy Team-name or duplicate-Agent flow.
- The existing Agent Detail Start Chat action consumes a stored campaign handoff by creating one idempotent task for that exact Agent. It clears the handoff only after success; failures remain retryable. Repository identity is revalidated from the stored GitHub URL, true mismatches are rejected, and case variants converge.
- Agent-first resume recognition is limited to the current v1 Admin-creator cohort. A Member who first used a Team Agent and later creates a personal Agent still retains the ordinary resumable path.
- The first-Team request requires exactly one Template. Its test fixture is cleaned after the suite; no persistence or migration was added.

Focused verification on the final delta passes: Server first-Team suite 17/17, Web Agent Detail/Onboarding/Settings/Template suites 134/134, shared schema 10/10, QA contract 4/4, root typecheck 9/9, exact changed-file Biome, and `git diff --check`.

Independent Standards and Spec reviews found no remaining blocker or important issue. The PR remains draft while CI and exact-head independent QA run. Please re-review this head.

baixiaohang (changes requested):
- Rationale: The incremental role and repository checks do not establish creator provenance, a recoverable Agent-first destination, or an exact target for the campaign action.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change establishes first-Team/Agent lifecycle state and adds Web data/routing behavior -> B-high.

**PR summary**
- Problem: A Team-less signer should create a Team only through an explicit Agent start and remain able to continue that exact Agent's Runtime/channel setup without entering legacy onboarding.
- Approach: Provision the first Team and Agent atomically, route initial confirmation to Agent Runtime, suppress the old onboarding entry, and retain campaign intent until chat creation.
- Impacted modules: first-Team and onboarding services, membership lifecycle, shared DTOs, Web onboarding/Settings/Agent/Template surfaces, and QA contracts.

**Review findings**

❌ 1. **[P1] Admin role still does not make `invitee_skip` a creator identity or provide the required continuation.** Role is mutable: a genuine quick-start invitee can later be promoted to Admin and manage a personal Agent, while a first-Team creator can later be downgraded after another Admin exists. The new predicate therefore still reclassifies either membership as its role changes. For the inferred creator it also continues to redirect `/onboarding` to `/` and keeps Settings on generic `View -> /team`; neither destination resumes the exact Agent's incomplete Runtime/channel progression. Persist/derive a non-ambiguous creator continuation that survives role and Agent lifecycle changes, and route direct entry plus Settings to that Agent-first destination. Changing `invitee_skip` from its locked invitee meaning remains a persistence-semantic change requiring explicit approval and Tree reconciliation. [R1, R5 / `packages/web/src/pages/onboarding/onboarding-page.tsx:51`; `packages/web/src/pages/settings/setup.tsx:492`; `system/cloud/onboarding.md`]

❌ 2. **[P1] Repository consistency does not bind the campaign handoff to its Agent or Team.** The increment validates `repoUrl` against `repoSlug`, but `StoredCampaignActionHandoff` still has no organization or Agent identity and every Agent detail page reads the same session value. Switching Agents/Teams before Start chat still sends the action to the current `agent.uuid` and clears it. Bind the handoff to `result.organizationId` and `result.agent.uuid`, ignore it on non-matching Agent pages, and cover those negative cases. [R4, R5 / `packages/web/src/utils/onboarding-flags.ts:163`; `packages/web/src/pages/agent-detail.tsx:445`; `packages/web/src/pages/templates/template-use-intent.tsx:247`]

✅ 3. The exact-one-Template boundary, direct initial Runtime destination, repository URL/slug consistency check, dedicated retry-conflict code, and membership lifecycle fence remain coherent.

**Action taken**
- Submitted request changes on exact head `65d6d8916b4296d0408c0a4dc00b84d59493dba8`.
- Tests were not run under review policy; CI and independent QA are still running.

Gandy2025:
Correction: the authoritative exact head from both local `git rev-parse` and the live PR is `65d6d8916b4296d0408c0a4dc00b84d59493dba8`. The preceding comment transcribed the wrong long SHA after the valid `65d6d8916` prefix. Please review and validate this exact head.

yuezengwu (changes requested):
Re-reviewed exact head `65d6d8916b4296d0408c0a4dc00b84d59493dba8`.

This increment adds repository URL/slug consistency checks for the stored campaign action, cleans the shared Template fixture, and narrows the inferred Agent-first cohort from every `invitee_skip + personal Agent` membership to Admins with those facts. The prior exact-one-Template, direct Runtime, idempotency, conflict-code, name-allocation, and membership-lock work remains coherent.

The two requested lifecycle fixes are not implemented:

1. **[P1] Current role still cannot identify a Team-less creator.** Adding `role === "admin"` at `packages/web/src/pages/onboarding/onboarding-page.tsx:51-55` and `packages/web/src/pages/settings/setup.tsx:497-500` only handles the simplest current cohort. An invitee can be promoted to Admin and later create or assume management of a personal Agent; the same persisted `invitee_skip` row is then indistinguishable from a Team-less creator and the resumable invitee setup is hidden again. Conversely, the creator still has no persisted exact Runtime continuation: direct `/onboarding` redirects to `/`, and Settings still renders generic `View -> /team` instead of Resume-to-Runtime. Keep `invitee_skip` invitee-only and model/resolve creator continuation truthfully. The specific persistence-semantic choice remains pending explicit human approval.

2. **[P1] The campaign handoff is still not bound to an Agent or Team.** The only Agent-detail change at `packages/web/src/pages/agent-detail.tsx:449-459` cross-checks `repoUrl` and `repoSlug`; `StoredCampaignActionHandoff` still contains no provisioned Agent/organization identity, every Agent header still reads it, and `openChat` still targets whichever `agent.uuid` is currently open. The new tests cover repository mismatch/casing, not switching to another Agent before Start chat. Bind the handoff to `result.agent.uuid` (and its Team), ignore it on other Agent pages, and cover that negative path.

Recommendation remains **request changes**. Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI and independent QA were not awaited.

Gandy2025:
The two continuation blockers are resolved on exact head `d3a78d92cab736e330b6e74755dba9b9d5409349`.

- Creator provenance is now persisted as the exact first Agent ID in reserved internal metadata on the caller's 1:1 human mirror, inside the existing atomic first-Team transaction. `/me` projects the exact target and its current status. The result survives role changes, manager changes, and Agent deletion without reclassifying an invitee. Direct onboarding and Settings resume go to that Agent's Runtime surface, while a deleted target returns to Template selection. `invitee_skip` is no longer used to infer provenance.
- The campaign handoff now carries both the provisioned organization ID and Agent ID. Agent Detail consumes it only on the exact target; other Agent and Team pages retain normal chat behavior and cannot clear it. The existing unscoped standalone-onboarding handoff remains backward-compatible.
- The product owner explicitly approved this persistence semantic: no table, column, enum, migration, or backfill; only the reserved internal marker on the existing human mirror plus a read-only `/me` projection.
- The branch is merged with current `origin/main` without rebasing or rewriting history.

Exact-head local verification is complete: Server 288 files / 3338 tests, Web 248 / 2333, Shared 75 / 915, skill evals 36 / 621, and the QA contract 1 / 4 all pass. Root typecheck (9/9), build (5/5), `pnpm check`, exact changed-file Biome, and `git diff --check` pass.

The PR remains draft while exact-head CI, independent full-isolated QA, and the requested re-reviews complete. Please re-review this head.

yuezengwu (changes requested):
Re-reviewed exact head `d3a78d92cab736e330b6e74755dba9b9d5409349`.

The increment replaces mutable role/manager inference with an exact first-Agent ID stored on the creator's human mirror, projects that target through `/me`, routes direct onboarding and Settings to its Runtime surface, and scopes campaign handoffs to the provisioned Team and Agent. Those changes substantively resolve the two findings on the preceding head. The dedicated retry conflict, membership lifecycle fence, exact-one-Template boundary, and server-side name allocation also remain intact.

Two recovery blockers remain in the new continuation implementation:

1. **[P1] Give the Agent-first continuation a completion/replacement transition instead of treating provenance as permanent unfinished setup.** The new marker is written once in `first-team-agent.ts:155-165` and never advanced or cleared. `/me` therefore projects it for the lifetime of the target, while `OnboardingPage` checks it before the membership completion state (`onboarding-page.tsx:50-56`) and Settings labels every active/suspended target `Resume setup` (`setup.tsx:496-502,576-616`). After the Runtime is bound and the user starts using the Agent, direct `/onboarding` still returns to Runtime and Getting Started still advertises unfinished setup forever. If the target is deleted, creating a replacement through the ordinary existing-Team Template path does not retarget the marker, so later resume keeps returning to Templates because the old tombstone remains. Record/derive the terminal `use-agent` transition and retarget or retire the deleted continuation when a replacement is chosen; cover both completed first-Agent setup and delete-then-replace. This is required by the locked Agent-first `set-up-runtime -> connect-channel -> use-agent` lifecycle, not merely a label preference.

2. **[P1] Scope the campaign handoff before the fallible Team-activation recovery boundary.** `TemplateUseIntent` waits for `selectOrganization` to succeed before adding `result.organizationId` / `result.agent.uuid` to session storage (`template-use-intent.tsx:240-259`). If the atomic provision succeeds but local Team activation fails, the page explicitly tells the user to reload; after reload `/me` can recover the exact Runtime destination, but the stored handoff is still unscoped and `AgentDetail`'s exact-target gate will not consume it. The advertised campaign continuation is therefore lost precisely on the recovery path this branch handles. Bind the handoff as soon as the provision result exists (before Team selection), or reconcile it from the authoritative continuation on reload, and cover a post-provision selection failure followed by Runtime entry.

The human-mirror metadata marker is a persistence-semantic change even without a schema or migration. I am treating the product owner's exact-head approval recorded in the PR thread as satisfying the repository's human-approval prerequisite; the blockers above concern the marker's lifecycle implementation.

Recommendation remains **request changes**. Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

baixiaohang (changes requested):
- Rationale: The exact continuation fixes the prior creator-provenance and campaign-target defects, but it survives membership reactivation and overrides the fresh onboarding lifecycle with an Agent that was transferred away when the member left.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change persists and routes core first-Agent lifecycle state across Server and Web -> B-high.

**PR summary**
- Problem: A Team-less signer should create a Team only through an explicit Agent start, then be able to resume that exact Agent's Runtime/channel setup without entering the legacy Team-first onboarding flow.
- Approach: Atomically provision the first Team and Agent, persist its exact continuation target on the caller's human mirror, project it through `/me`, and scope campaign actions to the provisioned Team and Agent.
- Impacted modules: OAuth/bootstrap, first-Team and campaign services, membership lifecycle, shared DTOs, Web onboarding/Settings/Agent/Template routing, and QA contracts.

**Review findings**

❌ 1. **[P1] Retire the creator continuation when a membership starts a fresh lifecycle.** Leaving or admin removal transfers the member's non-human Agents to another admin (`membership.ts:459-482`; `member.ts:357-380`), while reactivation deliberately clears the membership's onboarding suppress/completion fields but leaves the human mirror metadata untouched (`membership.ts:296-334`). Because `/me` validates the stored target only by Team and Agent status (`me.ts:307-316`), the same marker reappears when the stable mirror is reactivated. `OnboardingPage` then consumes it before the fresh membership state and redirects straight to the old Agent Runtime (`onboarding-page.tsx:50-56`), even though that Agent may now be managed by the fallback admin and a restored Member cannot configure it. This also contradicts the locked rule that rejoin starts a fresh membership lifecycle and stale onboarding state must not suppress it. Clear/version the marker in the same reactivation boundary (and account for the transfer), then cover self-leave and admin-removal followed by rejoin/restore. [R4, R5 / [Cloud Onboarding](https://github.com/agent-team-foundation/first-tree-context/blob/bfa5695fef74228e19d71ecbb0fa5535115bfe15/system/cloud/onboarding.md#L344-L348)]

✅ 2. The new exact marker removes mutable role/ownership inference, and the Team/Agent-scoped campaign handoff prevents another Agent detail page from consuming the action.

✅ 3. The parallel review already captures the separate completion/replacement transition and pre-Team-activation handoff ordering defects, so I have not duplicated them here.

**Action taken**
- Submitted request changes on exact head `d3a78d92cab736e330b6e74755dba9b9d5409349`.
- Tests were not run under review policy; CI and independent QA were not awaited.

Gandy2025:
The three recovery findings on `d3a78d92c` are addressed on exact head `6572861825858a9545e9232909106d19a1965e6a`, with the completion boundary deliberately narrowed after an additional Spec review.

- Completion is no longer inferred from any ordinary Web or campaign Task. `POST /me/onboarding-completed` now atomically stamps the existing terminal membership fact and retires the exact continuation, but the call site is intentionally deferred to the follow-up channel/Feishu first-use PR. This preserves the locked `set-up-runtime -> connect-channel -> use-agent` order instead of letting Web Chat skip channel setup.
- A deleted continuation is retargeted only by an authorized self-create, only when the old target is deleted in the same Team, and inside the existing Agent-create transaction.
- The campaign handoff is scoped to the exact provisioned Team and Agent immediately after provisioning, before fallible local Team activation.
- Membership reactivation clears the stale continuation in the existing lifecycle transaction, covering self-leave/rejoin and admin-remove/restore after Agent transfer.
- Team-less refresh sessions remain authenticated; Team authorization still comes from live membership.

No schema, migration, new onboarding enum, readiness state machine, or new visual flow was added.

Exact-head local evidence: focused Server 80/80; focused Web continuation/recovery 103/103; root typecheck 9/9; root build 5/5; changed-file Biome and diff check pass. The full monorepo attempt hit unrelated concurrent timeout flakes; all affected suites passed in isolated reruns (27 Web app-route tests, 36 Client OpenCode tests, 5 CLI runtime-layout tests, and 2 Context Tree audit cases).

Independent final reviews report Standards: 0 blockers and Spec: 0 blockers. The PR remains draft while exact-head CI and independent QA run. Please re-review this head, including the staged completion boundary described above.

yuezengwu (changes requested):
Re-reviewed exact head `6572861825858a9545e9232909106d19a1965e6a`.

The implementation goal remains coherent: authentication may leave an account Team-less; an explicit Template or known campaign Agent start owns atomic first-Team creation; exact retry and membership lifecycle fencing prevent duplicate or unauthorized Team boundaries. This head fixes the prior deleted-target replacement, pre-activation campaign scoping, membership-reactivation cleanup, and Team-less refresh-session gaps.

One lifecycle blocker remains:

**[P1] Do not ship the Agent-first cohort with no reachable channel or terminal continuation.** The final commit deliberately removes every current completion call after an ordinary or campaign task (`packages/web/src/pages/agent-detail.tsx:395-432`, `packages/web/src/pages/quickstart/quickstart-page.tsx:185-200`, and `packages/web/src/pages/workspace/conversations/new-chat-draft.tsx:510-545`). The server can retire the marker only through `POST /me/onboarding-completed` (`packages/server/src/api/me.ts:422-449`), but this cohort never enters the standalone flow that invokes that seam. Meanwhile direct `/onboarding` and both Settings resume actions map every active continuation back to the Agent Runtime page (`packages/web/src/pages/onboarding/onboarding-page.tsx:50-56`; `packages/web/src/pages/settings/setup.tsx:496-500,576-610`). After Runtime setup and successful task use, `/me` therefore continues to project the marker forever, Settings continues to advertise “Resume setup,” and resuming can only reopen Runtime; there is still no `connect-channel -> use-agent` destination or transition. Deferring the sole consumer to a future Feishu PR leaves the newly enabled production cohort in an internally permanent incomplete state and does not implement the locked [Agent-first sequence](https://github.com/agent-team-foundation/first-tree-context/blob/bfa5695fef74228e19d71ecbb0fa5535115bfe15/system/cloud/onboarding.md). Land a real channel/terminal consumer with this behavior (or keep the Team-less entry inaccessible until that consumer ships), and cover Runtime completion followed by channel continuation and terminal retirement.

Recommendation remains **request changes**. The continuation metadata persistence semantic has explicit product-owner approval; this finding concerns the missing reachable lifecycle transition, not the storage choice.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

baixiaohang (dismissed):
- Rationale: The prior recovery defects are resolved, but the enabled Agent-first cohort still has no reachable channel or terminal continuation after Runtime setup.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: the change persists and routes core first-Agent lifecycle state across Server and Web -> B-high.

**PR summary**
- Problem: A new signer should create a Team only when explicitly starting an Agent, then continue that exact Agent's setup without entering the legacy Team-first onboarding flow.
- Approach: Atomically provision the first Team and Agent, persist an exact resumable target, serialize membership boundaries, and preserve Team-less Template, Quickstart, Account, and campaign recovery paths.
- Impacted modules: OAuth/bootstrap, first-Team and campaign services, membership lifecycle, shared DTOs, Web onboarding/Settings/Agent/Template routing, and QA contracts.

**Review findings**

❌ 1. **[P1] The Agent-first path still has no reachable `connect-channel -> use-agent` transition.** This head correctly stops treating an ordinary Web or campaign Task as onboarding completion, but that leaves `POST /me/onboarding-completed` with no consumer for this cohort. Direct `/onboarding` and both Settings resume actions send every active continuation back to the Agent Runtime page; after Runtime setup and successful Web use, `/me` therefore keeps projecting the continuation and Settings keeps advertising “Resume setup” indefinitely. The locked Team-less creator sequence requires Runtime to continue into channel setup and terminal first use. Ship a reachable channel/terminal consumer with this entry boundary, or keep Team-less provisioning inaccessible until that consumer lands; cover Runtime completion through continuation retirement. [R5 / `packages/server/src/api/me.ts:422`; `packages/web/src/pages/onboarding/onboarding-page.tsx:50`; `packages/web/src/pages/settings/setup.tsx:496`; [Cloud Onboarding](https://github.com/agent-team-foundation/first-tree-context/blob/bfa5695fef74228e19d71ecbb0fa5535115bfe15/system/cloud/onboarding.md#L43-L56)]

✅ 2. The previous deleted-target replacement, pre-activation campaign scoping, membership-reactivation cleanup, and Team-less refresh-session findings are resolved on this exact head. The separate current review independently captures the same remaining lifecycle blocker, so I found no useful additional finding to duplicate.

**Action taken**
- Submitted request changes on exact head `6572861825858a9545e9232909106d19a1965e6a`.
- Tests were not run under review policy; CI and independent QA were not awaited.

Gandy2025:
The remaining lifecycle blocker is addressed on exact head `64f8e1db49453bd5fd47bd585a3e50109e57d7b9` by taking the reviewer's fail-closed option rather than pulling unfinished channel work into this foundation PR.

- `FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` is a single deployment flag that defaults to `false`.
- With the flag off, OAuth keeps the established personal-Team bootstrap and its `onboarding.team_created` telemetry, refresh requires an active membership, `POST /me/team-agents` returns `feature_disabled`, and a landing campaign cannot create a Team for a Team-less caller.
- The landing-campaign gate is checked again inside the existing user-locked transaction, closing the membership-loss race between the outer authorization read and provisioning.
- With the flag enabled, the previously reviewed Agent-first lifecycle remains unchanged. No new persistence, onboarding enum, readiness state machine, browser state, or Feishu implementation was added.
- The flag must remain off until the follow-up Runtime and Feishu first-use work makes `connect-channel -> use-agent` reachable and calls the existing terminal completion seam after real channel first use.

Exact-head local evidence: focused Server 143/143, Shared config 24/24, focused Web 99/99, root typecheck 9/9, root build 5/5, changed-file Biome, and `git diff --check` pass. Two independent re-reviews report Standards: 0 blockers and Spec: 0 blockers.

GitHub CI and independent exact-head QA are running. Please re-review this head, especially the default-off entry boundary and the in-transaction campaign recheck.

yuezengwu (approved):
Re-reviewed exact head `64f8e1db49453bd5fd47bd585a3e50109e57d7b9`.

The implementation goal is now staged safely: the approved Agent-first first-Team lifecycle remains implemented, but one default-off deployment gate prevents any incomplete Runtime-to-channel cohort from entering it. With the gate off, OAuth retains the established personal-Team bootstrap, refresh still requires active membership, the user-scoped first-Agent endpoint is unavailable, and a Team-less campaign cannot create a Team. When enabled later, the previously reviewed atomic Team/Agent, retry, recovery, and continuation boundaries remain intact.

The new gate closes the remaining blocker coherently. It is checked at every Team-less entry/authentication seam, including again inside the landing-campaign user-locked transaction, so a concurrent membership loss cannot bypass the rollout boundary. The configuration default and environment documentation both keep the incomplete path disabled until the follow-up Runtime/channel first-use work can consume the terminal completion seam.

The prior deleted-target replacement, pre-activation campaign scoping, membership-reactivation cleanup, exact conflict handling, additional-Team lifecycle fencing, account/sign-out routing, and Team-less refresh behavior under the enabled gate remain coherent. I found no remaining blocking issue and approve this head. The rollout flag must remain off until the locked `set-up-runtime -> connect-channel -> use-agent` continuation is reachable.

Per review policy, I inspected the diff and did not run tests. `git diff --check` is clean; CI is still running and was not awaited.

Gandy2025:
Independent, fully isolated QA reports **PASS** on exact head `64f8e1db49453bd5fd47bd585a3e50109e57d7b9`.

The qualification is intentionally narrow: this head is safe as a default-off foundation. The rollout flag must remain disabled until the follow-up Runtime/channel first-use path reaches and consumes the terminal completion seam.

QA evidence includes:

- Default-off live OAuth preserves the personal-Team bootstrap and `onboarding.team_created`; Team-less refresh and `/me/team-agents` fail closed.
- Default-on live authentication preserves a Team-less session while org-scoped APIs still reject it.
- Real PostgreSQL validation of atomic Team, Admin membership, human mirror, Agent, Template adoption, exact retry, distinct-intent conflict, rollback, and concurrency.
- Continuation completion, deleted-Agent replacement, leave/rejoin, admin remove/restore, campaign scoping, and the locked-transaction membership-loss race.
- Full Server 3346 tests, Web 2338, Shared 916, CLI 1736, skill evals 621, QA lifecycle 4; typecheck 9/9 and build 5/5.
- Production image health/readiness, portable release/install, CLI foreground daemon, and docs flows.

One macOS/Node 24 Client assertion reproduced identically on the base commit and passed on Linux; QA classified it as an environment-specific pre-existing issue, not a PR regression. No real Feishu/channel first-use or Momentic browser E2E was run, which is consistent with keeping the flag off.

All GitHub CI checks are now successful or intentionally skipped. The PR remains draft only while the second requested exact-head re-review resolves its preceding changes-requested state.

Gandy2025:
Exact head remains `64f8e1db49453bd5fd47bd585a3e50109e57d7b9`; all required CI jobs are green and @yuezengwu approved this head.

@baixiaohang, please complete the outstanding exact-head re-review. The only delta after your prior lifecycle finding is the single default-off deployment gate: existing deployments cannot enter the incomplete Agent-first cohort, while the enabled path retains the reviewed atomic and recovery behavior. Independent full-isolated QA has been restarted on this exact SHA.

Issue #2301, opened 2026-08-13
[P1] Preserve the default-Team invariant before enabling Agent-first onboarding


baixiaohang:
## Problem

PR #2297 merged a default-off foundation that can authenticate and route a self-service account with zero active Team memberships when `FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` is enabled.

A Cloud owner has now explicitly rejected that product invariant and prefers every self-service user to receive a default Team. This is not an onboarding-only detail: zero-membership support changes the assumptions used by authentication and refresh, tenant authorization, selected-Team routing, user-scoped provisioning, landing campaigns, membership loss/reactivation, and downstream readers that expect organization context.

The default-off gate protects current deployments from immediate exposure, but it does not resolve the architectural contradiction or make the dormant path safe to activate. The current locked Context Tree decision also says the opposite (authentication does not create a Team), so durable product context and source behavior must be reconciled before any rollout.

Related: #2297

## Why a Team must remain mandatory

A Team is not merely an onboarding artifact or a name shown in the UI. It is the product-facing form of the organization tenant boundary:

- Team/organization scopes collaboration resources, agents, chats, settings, quotas, governance, and Context Tree visibility. Without a Team, an authenticated account has identity but no normal tenant in which product work can be authorized or stored.
- Membership is the authorization relationship. It carries the caller's current Team role and lifecycle state; account identity, JWT claims, or a Web-selected organization must never substitute for a live membership check.
- Keeping one active default Team gives Server, Web, Client, integrations, and future features one stable rule: every normal authenticated product session has a tenant context, while every Team-scoped action still verifies the membership and role. Supporting zero memberships creates a permanent parallel mode that every reader, route, recovery path, and new feature must remember to handle.
- The invariant narrows security risk. Tenant resources cannot be created from account identity alone, and a missed Team-less special case cannot accidentally turn a user-scoped entry point into organization authority.
- It also makes membership lifecycle failures explicit. Losing the final membership must create/recover a safe default Team or enter a deliberate repair boundary; it must not leave a long-lived authenticated-but-tenantless account that behaves differently across surfaces.
- Automatic Team creation does not require a Team-first user experience. The product can create the default Team transactionally, omit a naming step, and still let the first visible and explicit decision be choosing an Agent or Template.

The default Team therefore preserves a cross-system tenancy invariant without sacrificing the intended Agent-first UX.

## Scope separation

[...]

PR #931, opened 2026-06-10
fix(client): centralize session turn lifecycle

baixiaohang:
## Summary
- Move session turn completion, ACK-through, in-flight cleanup, runtime recompute, and resume replacement decisions behind `SessionManager` APIs.
- Migrate Claude Code, Codex, and Claude TUI handlers off direct runtime state setters / no-arg completion into concrete `recordProviderActivity`, `finishTurn`, and `retryTurn` lifecycle reporting.
- Preserve provider-session continuity across daemon restart + Context Tree HEAD drift by routing persisted/evicted mappings through `resume(oldProviderSessionId)`; typed resume-unavailable reasons are the only path to explicit replacement.
- Add handler-side `ProviderBootstrap` shared prep for agent home/cwd, chat context, source repos, resource skills, briefing, provider payload/env, and diagnostic drift facts without ACK/state/recovery/turn_end/session replacement side effects.
- Preserve no-ACK TUI turn outcomes: timeout reports terminal runtime error while staying unacked for recovery; suspend/abort reports `turn_end:success` instead of being forced to error.
- Make suspend consumed-prefix ACK recovery-safe: consumed entries stay locally tracked until the ACK result; ACK reject marks the chat for recovery instead of forgetting delivered work; resume/new input waits for pending suspend ACK to settle before entering the handler.
- Guard post-suspend resume against stale session entries after terminate/evict races, while later same-chat waiters that see the same session already active are injected instead of dropped.
- Record explicit `transcript_missing`, `provider_resume_rejected`, and `local_session_artifact_missing` replacement reasons without treating tree/CLI drift as replacement.

## Context Tree
- Contract update PR: https://github.com/agent-team-foundation/first-tree-context/pull/467

Fixes #919.

## Validation
- `pnpm --filter @first-tree/client typecheck`
- `pnpm --filter @first-tree/client exec vitest run --maxWorkers=2 --minWorkers=1 src/__tests__/session-manager.test.ts` (1 file / 38 tests)
[...]

PR #2304, opened 2026-08-13
refactor: retire the authenticated zero-Team foundation

Gandy2025:
## Why

#2297 moved Team creation off OAuth sign-in and behind a default-off
`FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` flag. With the flag off —
its state on every deployment — nothing about the product changed, so what
actually shipped was a dormant second account lifecycle: an authenticated user
who owns no Team, plus the refresh, routing, campaign, and `/me` branches that
keep that state alive.

That foundation has been withdrawn as a product direction. This PR removes it
and restores one invariant: **every solo OAuth success immediately creates the
personal Team**, for current online users and for future OpenTag users alike.
Sign-in owns the first Team again; nothing else creates it.

This is not a revert of the #2297 squash commit. Each change in it was judged
individually: the zero-Team-only behavior is removed, and the fixes that stand
on their own without it are kept (see *Preserved*). The result is 65 files,
+380 / −3520 against `main`, and it touches no file #2297 did not touch.

## What changes

**The invariant.** `completeExternalAccountBootstrap` creates the personal Team
unconditionally for a solo signup, and its result carries a non-nullable
`organizationId` again. Every callback surface (GitHub, Google, OIDC)
unconditionally carries that Team back in the completion fragment, so the SPA
activates it instead of landing a brand-new user with no selection.

**Removed with it:**

- the `opentag.agentFirstOnboardingEnabled` config field, its
  `FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` env binding, and the
  `.env.example` entry;
- `POST /me/team-agents` and the whole `services/team/first-team-agent.ts`
  provisioning service — creating an Agent is Team-scoped
  (`POST /orgs/:orgId/agents`) for every caller again;
- the `allowTeamless` refresh option, restoring the original "no active
  membership → 401" behavior with its structured error attributes;
- the `firstTeamAgentContinuation` reserved Agent-metadata key, its `/me`
  membership field, and every read/write/retarget/clear path built on it;
- the caller-selected `uuid` and `retargetFirstTeamContinuationForHumanAgentId`
  options on `createAgent`;
- Team-less campaign provisioning in `startLandingCampaignTrial` — a caller
  with no resolvable membership is a `404` again, not a Team-creating branch;
- on the Web: `hasNoTeam`, the `RequireAuth` redirect to a Team-less entry, the
  `provisionFirstTeamAgent` client and its `uuid-v7` helper, the Team-less
  Template confirm screen, the continuation-driven Quickstart / Settings →
  Getting Started / onboarding / Agent-detail branches, and the
  `targetOrganizationId` / `targetAgentId` campaign-handoff fields.

**Docs and QA cases** describing the Team-less lifecycle are restored to the
behavior this PR ships (`onboarding-kickoff-contract.md`,
`registration-first-run-onboarding.md`, `account-provider-management.md`).

`packages/web/src/` is byte-identical to its pre-#2297 state apart from the one
retained change below.

## Preserved

These came in with #2297 but do not depend on authenticated zero-Team, so they
stay:

- `lockMembershipLifecycleUser` and its use in `deleteMember` /
  `deactivateMembership` — the per-user serialization point for membership
  lifecycle writes;
- `insertNewMembershipRows` plus `selfCreateOrganization` running its
  prerequisite check and new Team graph inside one user-identity transaction,
  and its guard that `POST /me/organizations` is an *additional*-Team path. The
  guard's message no longer points at an Agent-first flow: sign-in is what
  mints a first Team, so it now reads *"An active Team membership is required
  to create another Team"*;
- the account menu on the public Template library. This one is a judgment call
  worth naming: #2297 added it because `/templates` was the Team-less entry
  point, and that entry point is gone. It is kept because it is not zero-Team
  behavior, it still works correctly for an authenticated visitor browsing the
  catalog, and removing working UI would be a gratuitous regression. Happy to
  drop it if reviewers prefer the stricter line.

The caller lifecycle fence in `startLandingCampaignTrial` is also kept, and
this took a review round to get right. My first pass reverted
`services/landing-campaigns/start.ts` wholesale, which dropped it. #2297 had
moved caller resolution inside the provisioning transaction behind
`lockMembershipLifecycleUser`; its motivation was serializing first-Team
creation, but the fence also protected the ordinary existing-Team path, so
removing it let a removed admin's stale request still provision into the Team
they had just lost. Provisioning now takes that per-user lock and re-reads and
re-authorizes the caller membership before the advisory lock and the first Team
write, with no Team-less provisioning reintroduced.

`personalTeamDisplayName` and `agentNameSchema` were *not* kept: #2297 moved and
extracted them only to share them with the first-Team-Agent service, so with
that service gone both are back where they were, each next to its single
consumer.

## Verification

- `pnpm typecheck` — pass (9/9 packages)
- `pnpm check` (Biome) — pass
- `pnpm --filter @first-tree/shared test` — 912 pass
- `pnpm --filter @first-tree/web test` — 2300 pass
- `pnpm --filter @first-tree/server test` — 3319 pass

No schema, migration, backfill, or new persistence. No `/opentag` pages, Agent
selection UX, Runtime handoff UX, or rollout enablement.

### Added regression coverage

- an OIDC solo sign-in creates **exactly one** personal Team and carries it
  back as a pinned `org` in the completion fragment — the exact serialization
  #2297 made conditional, and previously unasserted for this provider;
- a tripwire that `POST /me/team-agents` is no longer routed. Deliberately
  named for what it is: routing resolves before auth, so it pins the path, not
  behavior;
- a campaign start whose membership is removed between the outer authorization
  read and provisioning must 404 and leave no service membership and no new
  Agent behind (see the fence below).

The restored suites already pin the rest: the GitHub callback's end-to-end solo
signup (Team graph + pinned `org`), the bootstrap service's concurrency and
slug-conflict behavior, refresh rejecting a membership-less user, invitee /
returning / multi-Team routing, Quickstart, and the Template intent journey.

Two coverage decisions worth stating rather than leaving implicit:

- **Google's `org` fragment** has no new test. Its route has no existing
  mockable success harness, and the shared bootstrap service plus the GitHub
  and OIDC end-to-end paths already pin the behavior it serializes.
- **No test asserts that `firstTeamAgentContinuation` is no longer a reserved
  Agent-metadata key.** Writing one would enshrine a retired identifier in the
  suite to assert a non-behavior.

## Related work and merge order

- Issue: agent-team-foundation/first-tree#2301 — *[P1] Preserve the default-Team
  invariant before enabling Agent-first onboarding*. This PR is its code fix.
- Context Tree: agent-team-foundation/first-tree-context#931 — *Align OAuth Team
  creation and OpenTag onboarding* (draft). It rewrites
  `system/cloud/onboarding.md`, which today still records Team-less Agent-first
  entry as the current decision.

**Code merges first.** Per the team's canonical rule, the Tree records decisions
against merged source truth, so this PR is the critical path: it lands, then
#931 is reconciled against what actually shipped here and marked ready. This PR
is not gated on that Tree change.

### Deployment check — cleared

`FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED` defaulted to `false` and
this PR removes it. Staging and production never configured it, so no account
was ever left without a Team by it and no flag-exposure or zero-membership
reconciliation is required.


yuezengwu (approved):
Code review complete: the authenticated zero-Team lifecycle is removed consistently across OAuth bootstrap, refresh, campaign start, /me, Agent creation, shared schemas, and Web routing. The independent membership-locking and additional-Team transaction fixes remain intact. I found no code blockers. Please keep the PR-described deployment-history check as a merge prerequisite: if the retired flag was ever enabled, identify and repair any zero-membership accounts before deployment.

baixiaohang (changes requested):
- Rationale: Retiring Team-less provisioning also removes the transactional membership fence, allowing a removed admin's stale campaign request to mutate the former Team.

**Risk level**: B-high
- Path baseline: `packages/server/**` outside the protected service set -> B-low.
- Semantic lift: changes OAuth, membership-backed authentication, Agent provisioning, and persisted Agent metadata semantics -> B-high.

**PR summary**
- Problem: Authenticated zero-Team mode creates a second tenancy lifecycle that contradicts the product invariant that normal solo sign-in owns a default Team.
- Approach: Restore unconditional personal-Team OAuth bootstrap and membership-backed refresh, remove the dormant Team-less Agent/API/Web/campaign continuation paths, and retain independent membership-serialization fixes.
- Impacted modules: OAuth/bootstrap, auth tokens, membership lifecycle, campaign provisioning, shared DTO/config, Web routing/onboarding/Templates/Settings, and QA contracts.

**Review findings**

❌ 1. **[P1] Revalidate live Admin membership inside the campaign provisioning transaction.** `startLandingCampaignTrial` authorizes the caller once at `packages/server/src/services/landing-campaigns/start.ts:529-534`, then performs the first Team mutations in a later transaction at `:334-347` without the user lifecycle lock or a membership re-read. If self-leave/admin removal commits between those points, the stale request still creates/reactivates the service membership and creates or updates the trial Agent in the former Team; those writes commit before chat creation can reject the now-inactive human mirror. The preceding head's in-transaction recheck and race regression were removed together with the Team-less branch, but the fence also protected the ordinary existing-Team path. Keep the membership lifecycle serialization/recheck around the first campaign writes and cover removal between the outer authorization read and provisioning. [R4, R5]

✅ 2. The default-Team OAuth/refresh restoration, removal of the user-scoped first-Team Agent route, and retirement of the Team-less Web/continuation branches are coherent within this staged PR. The broader final-membership repair work remains explicitly tracked by #2305 and should not close parent #2301 by itself.

**Action taken**
- Submitted request changes on exact head `25df954e5ec7ce8377b565cfb44074ea0984673f`.
- Review was static; tests were not run. Deployment flag history and any zero-membership reconciliation remain merge gates outside the repository diff.

yuezengwu (changes requested):
Superseding my earlier approval after validating the newly identified race: startLandingCampaignTrial authorizes the caller from an active Admin membership before provisioning, but provisionTrialAgent later mutates the Team in a separate transaction without locking the caller user or re-reading that membership. A concurrent self-leave or admin removal can therefore commit in the gap, after which the stale request can still create/reactivate the service membership and create/update the trial Agent in the former Team; those writes commit independently before chat creation can fail. Please preserve the caller lifecycle fence for the existing-Team path: take the caller user lock and revalidate active Admin membership inside the provisioning transaction before its first Team mutation, with a regression covering removal in this window. Static review only; tests not run.

Gandy2025:
Fixed in `7cc93387eca93b70d28e1bf0ac64dcb4a482da42`. The finding is correct, and thank you — this was a real regression against `main`, not just against the pre-#2297 baseline.

**What went wrong.** I reverted `services/landing-campaigns/start.ts` wholesale to its pre-#2297 state, judging the whole file's #2297 delta as Team-less-only. It wasn't: #2297 moved caller resolution inside the provisioning transaction behind `lockMembershipLifecycleUser`, and while its *motivation* was serializing first-Team creation, the fence also protected the ordinary existing-Team path. Reverting the file dropped it.

**The fix**, with no Team-less provisioning reintroduced: `provisionTrialAgent` now takes the caller's user id, and inside its transaction — before the advisory lock and before the first Team write — it takes the same per-user lifecycle lock the leave/remove paths take, re-reads the caller membership, and re-asserts the Admin and service-org guards. Provisioning and the chat bootstrap use that re-read membership. The outer authorization stays as a fast fail so error ordering around `assertOfficialLandingCampaignClient` is unchanged. Lock order is user → advisory, matching what was on `main`.

**Regression**: *"refuses a start whose membership is removed between authorization and provisioning"* in `landing-campaigns-start.test.ts`. A second connection holds the user row lock, commits the removal from inside that same transaction, and releases — exactly the ordering a real removal produces. I verified it discriminates: with the fence removed it fails with `expected 200 to be 404`, and it asserts no service membership row and no new Agent in the former Team.

**Verification on the new head**: `pnpm typecheck` 9/9, `pnpm check` clean, `landing-campaigns-start.test.ts` 50/50. Full server suite re-running; I'll report it here when it lands.

Two things I deliberately did **not** change, to keep this PR at its dispatched scope: the broader membership-loss / deletion-reactivation / invitation-only / zero-membership repair work stays with #2305, and the deployment-history check for the retired flag remains an independent merge gate outside this diff.

@yuezengwu @baixiaohang — ready for re-verification on `7cc93387eca93b70d28e1bf0ac64dcb4a482da42`.

yuezengwu (approved):
Re-verified the P1 fix on 7cc93387eca93b70d28e1bf0ac64dcb4a482da42. Provisioning now takes the caller lifecycle lock, re-reads and re-authorizes the active Admin membership inside the transaction, and does so before the advisory lock and first Team mutation. The chat bootstrap uses that transaction-authoritative caller. The new regression covers removal in the outer-authorization/provisioning gap and asserts that neither a service membership nor a trial Agent is left behind. No further code blockers found. Static review only; tests not run. The deployment-history / zero-membership check remains a separate merge prerequisite.

Gandy2025:
Deployment-history gate is now cleared: staging and production never configured the retired flag, so no flag-exposure reconciliation is required. The P1 campaign membership race is fixed on exact head `7cc93387eca93b70d28e1bf0ac64dcb4a482da42`; all current checks pass and @yuezengwu has re-approved that head. @baixiaohang please re-review the exact head and replace the outstanding change request if the fix satisfies the finding.

yuezengwu:
The deployment risk is now resolved operationally, but the PR body still says **“Deployment check still open.”** Please update that section before merge to record the verified result: staging and production never configured `FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED`, so no flag-exposure or zero-membership reconciliation is required. Keeping the stale open-gate wording risks confusing the final merge decision and the durable PR record.

Separately, GitHub still reports `CHANGES_REQUESTED` because baixiaohang's review on the previous head remains outstanding; their re-review of exact head `7cc93387eca93b70d28e1bf0ac64dcb4a482da42` is still needed to clear that review state.

Gandy2025:
PR body updated: the deployment section now records the verified result — staging and production never configured `FIRST_TREE_OPENTAG_AGENT_FIRST_ONBOARDING_ENABLED`, so no flag-exposure or zero-membership reconciliation is required. The stale "still open" wording is gone.

Also folded into the body: the campaign caller-lifecycle fence and its regression, and the corrected server test count (3319).

Verification on `7cc93387eca93b70d28e1bf0ac64dcb4a482da42`: CI green (Test Server, Test Client & Web, Lint & Type Check). Locally `pnpm typecheck` 9/9, `pnpm check` clean, and the full server suite 287 files / 3319 tests pass.

One note on local runs, since it could otherwise look like a flake I ignored: an earlier local full run failed `members.test.ts > rejects a stale admin restore after a concurrent self rejoin wins`. That test polls `pg_stat_activity` for a lock wait on a 5s deadline, and it timed out while I had other work running on the same machine. It passes alone and in a clean full run, the file is untouched by this PR, and CI is green on the same head.

Remaining state is only the review gate: @baixiaohang's change request predates the fix and still needs a re-review of this exact head to clear.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The requested membership-lifecycle fence is restored on the exact reviewed head, its race regression covers the stale-removal window, and no further code blockers remain.

**Risk level**: B-high
- Path baseline: packages/server changes span OAuth bootstrap, membership-backed authentication, campaign provisioning, and Agent behavior.
- Semantic lift: the PR changes tenancy lifecycle and persisted Agent metadata semantics.

**PR summary**
- Problem: The dormant authenticated zero-Team lifecycle conflicted with the product invariant that a solo OAuth signup immediately owns a personal Team.
- Approach: Restore unconditional personal-Team bootstrap and membership-backed refresh, remove Team-less API/Web/campaign continuation paths, and retain the independent membership serialization and additional-Team transaction safeguards.
- Impacted modules: OAuth/bootstrap, auth refresh, membership lifecycle, landing campaigns, Agent creation and metadata, shared schemas/config, Web onboarding/routing, and QA contracts.

**Review findings**
✅ 1. The P1 campaign race is resolved: provisioning takes the caller lifecycle lock and revalidates active Admin membership inside the transaction before the advisory lock and first Team mutation. The chat bootstrap uses that transaction-authoritative caller. [R4]
✅ 2. The new regression covers membership removal between outer authorization and provisioning and asserts no service membership or trial Agent remains.
✅ 3. The zero-Team lifecycle removal is coherent across server, shared, and Web surfaces while the independent membership safeguards remain intact. The operational deployment check is recorded as cleared in the PR body.

**Action taken**
- Approved exact head 7cc93387eca93b70d28e1bf0ac64dcb4a482da42. Static review only; I did not run tests. All current GitHub checks pass.

PR #2308, opened 2026-08-13
feat(web): add the authenticated OpenTag entry

Gandy2025:
## What

`/opentag` — the authenticated OpenTag entry. One route, three decisions:
which teammate, which Computer, which Feishu Bot.

It is a standalone full-screen route beside `/onboarding`, not a variant of it.
`/onboarding`, signup, invitee, returning-user, multi-Team, the Template
Library, Quickstart and Settings are untouched.

| Step | Decision |
| --- | --- |
| Create your Team Agent | which teammate — one Template, then a name |
| Set up its Runtime | which Computer — and the Agent is created here |
| Add to Feishu | one Bot for this Agent |
| Use in Feishu | the destination; real first use happens in Feishu |

## The Agent is materialized once, atomically

The Template and the name are a local draft. Nothing is written until the
Computer is chosen, and then the existing organization-scoped create writes the
Agent row, its `clientId`, the runtime that Computer reports ready, the
matching durable config, and the Template adoption in the transaction it
already had. This is the same create the standalone onboarding flow uses —
OpenTag adds no protocol of its own.

An earlier revision of this PR created the Agent during Template selection and
bound it afterwards. That ordering is what produced a chain of compensating
mechanisms: the service-default provider could be incompatible with the
Computer chosen next, so the first bind had to rewrite it; rewriting it left
the durable config on the old provider; and two concurrent first binds could
each claim a route. All of it is gone with the ordering that caused it. The
Context Tree locks the order the member *experiences* — `choose-agent` then
`set-up-runtime` — not that Step 1 persists an Agent.

Turning back before the Computer is chosen therefore leaves nothing behind:
no half-configured Agent to recover, and no draft to persist.

## Sign-in still owns the Team

A logged-out visit takes the ordinary deep-link route through `/login` and
OAuth. The server preserves this exact destination through solo signup on the
same strict terms as the Agent Template intent — canonical path, at most an
`agent` UUID, no fragment, byte-identical to the builder's output — so `next`
preservation never widens into a general deep-link channel
(`packages/shared/src/opentag-entry.ts`). A returning member acts in the Team
they already have; OpenTag never creates or names one.

## Recovery

The Agent id in the URL plus authoritative reads answer every step after
creation, so these are all defined:

- nothing Team-scoped proceeds until `/me` is authoritative — `meLoaded` also
  flips after a transport failure — and the member gets a retry inside the step;
- wrong / deleted / suspended Agent, an Agent this member cannot manage, or a
[...]

PR #2311, opened 2026-08-13
fix(web): ignore a fragment on the OpenTag entry instead of dropping the destination

Gandy2025:
## What

Follow-up to #2308. A fragment on the OpenTag entry URL is now ignored rather than rejected, and the page rewrites the address bar to the canonical spelling.

This closes the one reviewer finding from #2308 that was deliberately deferred as non-blocking. Auditing it against merged `main` turned up a second, more costly consequence of the same root fact, so this PR fixes both.

## The visible half

`parseOpenTagEntryPath` rejected any URL with a fragment, but the browser route check built its candidate from `pathname + search` only. So `/opentag?agent=<uuid>#step` worked normally and simply kept the fragment in the address bar — the contract said "no fragment" while the browser quietly accepted one.

## The costly half

`RequireAuth` stashes `pathname + search + hash`, and `SAFE_NEXT_PATH` permits `#`, so the fragment travels to the server inside `next`. `shouldPreserveSoloSignupNext` parses it, gets `null`, and drops the destination.

The result is a brand-new member who follows a shared OpenTag link that picked up a fragment: solo signup creates their personal Team and then lands them at the workspace root instead of the entry they clicked. That is precisely the journey the preservation contract exists to protect. Verified on merged `main` before changing anything:

```
safeRedirectPath("/opentag?agent=<uuid>#step")        -> unchanged (fragment survives)
parseOpenTagEntryPath("/opentag?agent=<uuid>#step")   -> null
shouldPreserveSoloSignupNext("/opentag?agent=<uuid>#step") -> false
shouldPreserveSoloSignupNext("/opentag?agent=<uuid>")      -> true
```

## Why ignoring is the right resolution

A fragment is not part of an entry's identity. It is never sent to the server on a normal navigation, and it cannot change which page a same-origin relative redirect lands on — `safeRedirectPath` has already constrained the target to a relative path. Rejecting it bought no safety and cost the whole destination.

Everything outside the fragment stays exactly as strict as before. The canonical comparison strips the fragment from the **raw string** rather than reassembling from the parsed `URL`; reassembling would have silently re-admitted `/opentag?` and `/opentag?agent=<uuid>&`, whose empty trailing query the URL parser normalizes away. The existing canonical-spelling tests caught that while this was being written, which is why the comparison is shaped this way.

`parseAgentTemplateIntentPath` keeps its own fragment rejection; changing the sibling contract is outside this fix.

## Scope

Nothing else from #2308's review history is still open — the audit is in the PR discussion. No schema, migration, persistence semantics, new state, service layer, or abstraction. No W4 real-first-use behavior.

## Testing

- New regressions at all three layers, each confirmed to fail on current `main` and pass here: shared (fragment ignored and normalized; non-canonical spellings still rejected through a fragment), server (`shouldPreserveSoloSignupNext` preserves a fragment-carrying entry, still refuses `?step=runtime#a`), web DOM (address bar rewritten to canonical, Agent still resolved).
- `pnpm check` clean; shared / server / web typecheck clean including the web design-token guardrail.
- shared 924 and web 2354 passing; the affected server file (`oauth-bootstrap`) passing. The full server suite was not run in this worktree — concurrent server suites corrupt each other's worker databases locally, so CI owns that run.

### DIFF
--- a/packages/server/src/__tests__/oauth-bootstrap.test.ts
+++ b/packages/server/src/__tests__/oauth-bootstrap.test.ts
@@ -275,7 +275,13 @@ describe("shouldPreserveSoloSignupNext", () => {
     // Non-canonical spellings of the same destination.
     expect(shouldPreserveSoloSignupNext("/opentag/")).toBe(false);
     expect(shouldPreserveSoloSignupNext("/opentag?")).toBe(false);
-    expect(shouldPreserveSoloSignupNext("/opentag#step")).toBe(false);
+    // A fragment is ignored rather than dropping the destination: the browser
+    // stashes `pathname + search + hash`, so a shared link carrying `#...`
+    // would otherwise land a brand-new member at the workspace root.
+    expect(shouldPreserveSoloSignupNext("/opentag#step")).toBe(true);
+    expect(shouldPreserveSoloSignupNext("/opentag?agent=0198b2c4-1f6a-7c31-9a02-4d5e6f708192#step")).toBe(true);
+    // Still not a hole: the fragment cannot carry anything past the gate.
+    expect(shouldPreserveSoloSignupNext("/opentag?step=runtime#a")).toBe(false);
     // Unknown or malformed parameters.
     expect(shouldPreserveSoloSignupNext("/opentag?agent=not-a-uuid")).toBe(false);
     expect(shouldPreserveSoloSignupNext("/opentag?step=runtime")).toBe(false);
--- a/packages/shared/src/__tests__/opentag-entry.test.ts
+++ b/packages/shared/src/__tests__/opentag-entry.test.ts
@@ -30,7 +30,25 @@ describe("parseOpenTagEntryPath", () => {
     expect(parseOpenTagEntryPath("/opentag/")).toBeNull();
     expect(parseOpenTagEntryPath("/opentag?")).toBeNull();
     expect(parseOpenTagEntryPath(`/opentag?agent=${AGENT_UUID}&`)).toBeNull();
-    expect(parseOpenTagEntryPath(`/opentag?agent=${AGENT_UUID}#step`)).toBeNull();
+  });
+
+  it("ignores a fragment instead of dropping the whole destination", () => {
+    // A link that picked up `#...` is still this entry. Rejecting it used to
+    // strand a brand-new member at the workspace root after solo signup,
+    // which is the one journey this contract exists to protect.
+    expect(parseOpenTagEntryPath("/opentag#step")).toEqual({ agentUuid: null });
+    expect(parseOpenTagEntryPath(`/opentag?agent=${AGENT_UUID}#step`)).toEqual({ agentUuid: AGENT_UUID });
+    expect(parseOpenTagEntryPath(`/opentag?agent=${AGENT_UUID}#`)).toEqual({ agentUuid: AGENT_UUID });
+  });
+
+  it("still requires everything outside the fragment to be canonical", () => {
+    // The fragment is ignored, not a hole: it cannot carry a second path,
+    // an extra parameter, or a non-canonical spelling past the gate.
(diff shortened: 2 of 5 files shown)


Gandy2025:
### Audit of #2308's reviewer findings against merged `main` (`5e969246a`)

Every ❌ raised across the seven review rounds, checked individually in the merged tree rather than from the PR conversation:

| Finding | State in `main` | Evidence |
| --- | --- | --- |
| Private/manageable Agent passes eligibility | Closed | `flow.ts:109` rejects `visibility !== "organization"` |
| First-bind provider change leaves `agent_configs` on the old provider | Closed — cause removed | atomic create; `runtimeProvider` exists only on the pre-existing `switchAgentRuntimeSchema`, not `updateAgentSchema` |
| Lost-response retry can double-bind / concurrent first binds | Closed — protocol withdrawn | no first-bind path remains |
| Recovery candidate overrides eligibility and loops | Closed | `recover-created-agent.ts:47-49` requires active + organization-visible + bound |
| Backing out leaves `recoverableAgent` / `continueError` alive | Closed | `clearRecoveryState()` on both back paths |
| Unreachable `catch` on a swallowed `/me` failure | Closed | `refreshMeStrict` (rejects) replaced `refreshMe` |
| Stale readiness walks the member into a second Agent | Closed | strict gate + `/` onboarding cross-gate assertion |
| Recovery lock only lives as long as the component | Closed | Agent anchored in the URL |
| URL anchor delayed behind `invalidateQueries` | Closed | navigate precedes the un-awaited invalidation |
| Feishu handoff live before authoritative readiness | Closed | binding query enabled on readiness, not on step |
| Readiness gate deadlocks an admin on a teammate's Agent | Closed | gate scoped to `agent.managerId === memberId` |
| Reusable QA case deferred | Closed | `packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md` |
| **`location.hash` normalization** | **Open** | fixed in this PR |

Deliberately excluded, per scope: QA's optional suggestion to add an explicit admin/teammate-owned branch to the reusable case (case maintenance, not a defect), the copy observation about the generic Step 1 text after a full reload with `/me` blocked (QA judged it non-blocking and the safety invariants hold), and W4 real-first-use.

yuezengwu (approved):
The goal and implementation are sound: fragment-bearing OpenTag entry URLs now survive the solo-signup redirect, while the browser rewrites them to the canonical fragment-free spelling without losing the selected Agent.

The raw-string canonical comparison keeps the existing strict path/query boundary intact, and the shared, server, and web regression coverage exercises the behavior at each affected layer. I found no blocking issues. No schema, persistence, or core data-structure changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change preserves fragment-bearing OpenTag entries across solo signup while keeping the existing strict path and query boundary intact.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` tests, with shared and Web implementation changes; no schema or sensitive service-path change.
- Semantic lift: none; this is redirect parsing and browser URL normalization, with no database, persistence, or core agent-state mutation.

**PR summary**
- Problem: A first-time member following a shared OpenTag link with a fragment could lose the intended OpenTag destination during solo signup and land at the workspace root.
- Approach: Treat the fragment as non-identity data, preserve the strict builder-derived path/query spelling, and rewrite the browser URL to the canonical fragment-free entry without losing the selected Agent.
- Impacted modules: shared OpenTag entry parser, server solo-signup redirect coverage, and Web OpenTag route normalization.

**Review findings**
✅ 1. The raw-string comparison removes only the fragment and continues to reject trailing query syntax, extra parameters, alternate paths, and non-canonical Agent identifiers.
✅ 2. Shared, server, and Web regressions cover the parser contract, solo-signup preservation, and address-bar normalization at the affected boundaries.

**Action taken**
- Approved.

PR #2312, opened 2026-08-13
feat(web): redesign agent instructions tab

Gandy2025:
## Summary

- lead the Instructions tab with an always-visible, read-only `What this agent follows` preview and an ordered full-result dialog
- organize prompt rows into `Applied instructions` and `Not applied`, with plain-language ownership, editability, and concise excerpts
- keep existing prompt mutation semantics while opting Instructions into mobile-safe rows and real 44px switch/menu targets without changing other Agent Detail tabs

## Test plan

- `pnpm --filter @first-tree/web exec vitest run --passWithNoTests --reporter=dot` (251 files, 2,354 tests)
- focused Agent Detail suites (56 tests)
- `pnpm --filter @first-tree/web build`
- Playwright visual and interaction checks at 1440px and 390px, including zero horizontal overflow, 44px action targets, ordered full-result dialog, and focus restoration

## Scope

This PR changes only the Instructions presentation and an explicit opt-in presentation path on shared row/switch primitives. Default shared component rendering and the other Agent Detail tabs remain unchanged.

### DIFF
--- a/packages/web/src/pages/agent-detail/prompt-tab.tsx
+++ b/packages/web/src/pages/agent-detail/prompt-tab.tsx
@@ -10,15 +10,23 @@ import { type FormEvent, type ReactNode, useEffect, useRef, useState } from "rea
 import { Navigate } from "react-router";
 import { getAgentResources, updateAgentResources } from "../../api/agent-resources.js";
 import { Button } from "../../components/ui/button.js";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from "../../components/ui/dialog.js";
 import { Markdown } from "../../components/ui/markdown.js";
 import { Popover } from "../../components/ui/popover.js";
 import type { RowAction as RowMenuAction } from "../../components/ui/row-actions-menu.js";
 import { Section } from "../../components/ui/section.js";
 import { Textarea } from "../../components/ui/textarea.js";
-import { agentResourcesMutationHandlers, resourceTypeIcon, statusMarker } from "./capability-section.js";
+import { agentResourcesMutationHandlers, resourceTypeIcon } from "./capability-section.js";
 import { useAgentDetailContext } from "./layout-context.js";
 import { ResourceRowView, type RowMenu, type RowStatusMarker, type RowToggle } from "./resource-row.js";
-import { sourceLabel, templateSourceLabel } from "./resource-source.js";
+import { templateSourceLabel } from "./resource-source.js";
 import { titleWithSemantics, useJustSaved } from "./save-semantics.js";
 
 type AvailablePrompt = { id: string; name: string };
@@ -78,7 +86,9 @@ export function PromptTab() {
   const canEditPrompt = ctx.canManageAgent && ctx.agent.status === "active";
   const resourceError = resourcesQuery.error instanceof Error ? resourcesQuery.error.message : null;
   const resources = resourcesQuery.data;
-  const showEffectiveInstructions = resources ? shouldShowEffectiveInstructions(resources, prompt) : false;
+  const appliedCount = resources
+    ? appliedPromptRows(resources).length
+    : (promptSections?.filter((section) => section.body.trim()).length ?? (prompt.trim() ? 1 : 0));
   const editorError = savePromptMut.error instanceof Error ? savePromptMut.error.message : null;
   const bindingError = bindingMut.error instanceof Error ? bindingMut.error.message : null;
   const bindings = resources?.bindings ?? [];
@@ -134,9 +144,12 @@ export function PromptTab() {
 
   return (
     <div className="flex flex-col" style={{ gap: "var(--sp-5)" }}>
+      <EffectiveInstructionsBlock prompt={prompt} sections={promptSections} appliedCount={appliedCount} />
       <Section
         headingLevel={3}
-        title={titleWithSemantics("Instructions", justSaved)}
+        title={titleWithSemantics("Applied instructions", justSaved)}
+        count={appliedCount}
+        className="[&>div:first-child]:flex-row [&>div:first-child]:items-center"
         action={
           canEditPrompt && !resourceError && !editor && resources ? (
             <AddInstructionsMenu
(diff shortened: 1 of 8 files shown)


baixiaohang (changes requested):
- Rationale: Two edge cases make the redesigned surface contradict its stated mobile-target and effective-instruction semantics.

**Risk level**: C
- Path baseline: changes are limited to `packages/web/**`.
- Semantic lift: none; the PR does not change subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Agent managers need to understand the exact instruction stack an agent follows, distinguish applied from inactive sources, and operate the page safely on mobile.
- Approach: Adds an always-visible read-only result preview with an ordered dialog, separates applied and inactive rows, clarifies ownership/editability, and opts instruction rows into mobile-specific presentation.
- Impacted modules: Agent Detail Instructions UI, shared resource-row presentation, and the Switch primitive.

**Review findings**
❌ 1. The instruction-row overflow menu still has a 28×28px hit target. `triggerClassName="min-h-11 min-w-11"` cannot override `RowActionsMenu`'s inline `width: 28` / `height: 28`, so the computed target remains 28px even though the new test sees the 44px class. Please expose an actual touch-target sizing path (or otherwise remove/override the inline dimensions) and verify the rendered size. [R5 / `packages/web/src/pages/agent-detail/resource-row.tsx:153`]
❌ 2. `appliedPromptRows` classifies every `mode === "enabled"` row as applied, even when `promptBody` is empty. Empty Team Prompt bodies are valid, while the server's runtime projection omits them from both `prompt.sections` and the merged append. This can render "No additional instructions are active" above "1 instruction applied" and an Applied row. Please derive the Applied classification/count from the same content predicate as the runtime projection while retaining empty rows in a manageable inactive state. [R1 / `packages/web/src/pages/agent-detail/prompt-tab.tsx:878`]

**Action taken**
- Submitted request changes.

packages/web/src/pages/agent-detail/prompt-tab.tsx
```diff
@@ -704,13 +746,14 @@ function PromptResourceBlock(props: {
...
         <span title={props.templateName !== undefined ? templateSourceLabel(props.templateName) : undefined}>
-          {sourceLabel(props.source)}
+          {instructionOwnershipLabel(props.source, !!props.customizedFromTeam)}
         </span>
       }
       status={props.marker}
       toggle={props.toggle}
       menu={props.menu}
       dimmed={props.dimmed}
+      peek={props.unavailableReason ?? (hasBody ? promptExcerpt(props.body) : undefined)}
```
yuezengwu: [P2] Translate unavailable reason codes before rendering them

`unavailableReason` is an API reason code, not user-facing copy: the resolver assigns values such as `prompt_budget_exceeded`. This new peek will therefore show the underscored machine code in production, contrary to the plain-language presentation this PR introduces. The added test passes an already-formatted sentence (`Prompt budget exceeded.`), so it does not exercise the real response shape. Please map known reason codes to readable copy (with a safe fallback) before rendering them here.

yuezengwu (commented):
Reviewed the Instructions redesign and shared-component opt-in path. I found one actionable plain-language UX issue; no blocker, core data-structure change, or database change.

Gandy2025:
Addressed both requested changes in `078a088a8f0cddb336ee1c52a07634de49640b09`.

1. `RowActionsMenu` now has an explicit opt-in `touchTarget` path that sets the trigger's own inline width and height to `var(--sp-11)`. At a 390px viewport, Playwright measured the Instructions menu and switch at 44×44px computed size; a menu on the unchanged default resource-row path remained 28×28px.
2. Applied prompt classification now matches the server runtime projection predicate (`mode === "enabled" && !!promptBody`). An enabled empty Team Prompt moves to `Not applied`, does not affect the count/result, and retains its existing Switch and Customize controls. The empty inline recovery/edit path is also preserved.

Validation on the new head: 251 Web test files / 2,356 tests passed, focused Agent Detail tests passed, Web typecheck and design-token guard passed, production Web build passed, and the 390px page had no horizontal overflow.

packages/web/src/pages/agent-detail/prompt-tab.tsx
Gandy2025: Fixed in `078a088a8f0cddb336ee1c52a07634de49640b09`. The Instructions row now maps `prompt_budget_exceeded` to “These instructions exceed the agent's instruction limit.” and humanizes unknown machine codes as a safe fallback, while preserving already-readable server copy. The regression test now supplies the real raw reason code and asserts that neither the known nor unknown underscored code reaches rendered text.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head resolves both requested changes and the additional plain-language reason-code issue without changing the default shared-component path.

**Risk level**: C
- Path baseline: changes remain limited to `packages/web/**`.
- Semantic lift: none; the PR does not change subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Agent managers need to understand the exact instruction stack an agent follows, distinguish applied from inactive sources, and operate the page safely on mobile.
- Approach: Adds a read-only result preview and ordered dialog, separates applied and inactive rows, clarifies ownership/editability, and opts instruction controls into a mobile-safe presentation while preserving other Agent Detail rows.
- Impacted modules: Agent Detail Instructions UI, shared resource-row presentation, Switch, and row actions menu.

**Review findings**
✅ 1. The menu's opt-in touch target now sets the trigger's own inline dimensions to the 44px spacing token, while the default path remains 28px.
✅ 2. Applied classification now matches the runtime projection predicate; enabled empty prompts stay manageable under Not applied without affecting the count or result.
✅ 3. Prompt availability reason codes are converted to readable copy, with regression coverage for both the known code and an unknown-code fallback.

**Action taken**
- Approved.

yuezengwu (approved):
Re-reviewed the fixes on exact head `078a088a8f0cddb336ee1c52a07634de49640b09`. The applied-count/runtime-projection mismatch, the Instructions-only 44px menu target, and the unavailable-reason plain-language issue are all addressed without changing the default shared-row path. No remaining findings.

PR #2313, opened 2026-08-13
feat(web): complete OpenTag onboarding on real Feishu first use

Gandy2025:
`/opentag` used to end at a connected Bot. A Bot is not first use: the member has set something up, not used it, and stamping onboarding there would call setup finished before the Agent had ever done work.

The finishing fact now comes from the Feishu Task that ingress creates when a real person messages the Bot. This page only reads whether that already happened — it starts nothing, acknowledges nothing, and invents no timestamp, receipt, delivery protocol, or second onboarding state. No Server, Shared, CLI, or Client contract is touched.

## What makes the read exact

The Bot binding. A Task is identified by its Bot binding plus the external chat, and a Bot belongs to exactly one Agent, so a chat carrying this Agent's binding id can only be this Agent's work.

The origin and participant filters only narrow the candidates cheaply — they are not the proof. An Agent invited into a teammate's Task is also a speaker in a Feishu-origin chat, so "speaks in a Feishu chat" alone would let a neighbour's first use finish this member's setup. The Team comes from the endpoint's own scope.

## What does not complete onboarding

The absence of an answer is its own state, distinct from "not used yet". A read that has not landed, or that failed, leaves the member on the Bot step; only a landed, matching read completes. Nothing is polled at all until a Bot exists, because a Bot the member is still confirming cannot have carried a message.

Completion goes through the existing idempotent seam, so a reload after first use converges on the same membership rather than producing a second onboarding state. The handoff link waits for that stamp — an unstamped membership can still be sent back into setup, so offering the destination early would advertise a door that bounces.

## Verification

- `pnpm --filter @first-tree/web test` — 252 files, 2377 tests pass
- `pnpm --filter @first-tree/web typecheck` — clean, including the design-token gate
- `pnpm --filter @first-tree/web build` — clean
- Full CI green on this head

Focused tests cover the negative case, the exact positive, a cross-Agent Task, an admin who can see the primary's Task through a managed collaborator, a Task far down the paginated list, a failed task read, reload convergence, poll recovery after a read failure, and completion-stamp failure and retry.

Three guarantees were checked by mutation rather than trusted: removing the Bot-binding equality breaks three assertions, disabling the poll interval breaks exactly the recovery test, and always-settling the handoff breaks exactly the gate test.

## Reviewed and known limits

Reviewed on both Standards and Spec axes, and independently QA'd. QA could not reach a live Feishu tenant, so the end-to-end leg — real message → ingress → Task → terminal state — is described in `packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md` (updated here) and remains unexecuted.

An admin continuing a *teammate's* Agent cannot reach the terminal state, and that is now an explicit owner gate rather than a visibility accident. Visibility alone would not have been enough: an admin who manages a collaborator Agent invited into the primary's Task receives a membership in that Task through the ordinary watcher projection, so the read would have succeeded and the binding would have matched. The read and the write both require `agent.managerId === memberId`, because completion means this member owns the setup being stamped.

### DIFF
--- a/packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md
+++ b/packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md
@@ -10,16 +10,18 @@ surfaces: [server, web, client]
 ## Goal
 
 Confirm that one member can go from an unauthenticated `/opentag` link to a
-Feishu Bot connected to exactly one Agent, and that every interruption on the
-way leaves either nothing or that same Agent — never a second one.
+Feishu Bot connected to exactly one Agent and on to real first use in Feishu,
+and that every interruption on the way leaves either nothing or that same
+Agent — never a second one.
 
 The deterministic pieces (URL parsing, eligibility classification, step
 selection) belong to product tests. This case owns what only a live run can
 answer: that the atomic create really writes the Agent, its Computer, its
 runtime and its durable config together on a real Computer; that a real
-provider mix does not dead-end; and that a failed readiness read cannot walk
-the member into creating another Agent through the workspace and onboarding
-gates.
+provider mix does not dead-end; that a failed readiness read cannot walk the
+member into creating another Agent through the workspace and onboarding gates;
+and that the membership is stamped complete only by a Feishu Task genuinely
+belonging to this Agent, which no mocked read can establish.
 
 ## Preconditions
 
@@ -30,10 +32,19 @@ gates.
 - A real connected Computer. At least one pass must use a Computer whose ready
   runtime is **not** the service default — a Codex-only machine is the useful
   shape, because the historical defect was an Agent stuck on `claude-code`.
-- Fresh browser state, and a way to fault-inject `/me` (devtools request
-  blocking is enough). Never reuse a developer's session or localhost login.
-- Direct database read access for the Agent, its runtime config and the Feishu
-  binding. UI text alone does not establish what was written.
+- Fresh browser state, and a way to fault-inject `/me`, the chat reads and the
+  completion request independently (devtools request blocking is enough). Never
+  reuse a developer's session or localhost login.
+- Direct database read access for the Agent, its runtime config, the Feishu
+  binding, the Task chat's metadata and the membership onboarding columns. UI
+  text alone does not establish what was written.
+- A disposable Feishu tenant with a real person able to message the Bot, since
+  only real ingress creates the Task this journey ends on. Reuse the tenant
+  hygiene rules in [feishu-agent-channel](feishu-agent-channel.md); never use a
+  customer conversation.
+- For the cross-Agent leg, a second disposable Agent in the same Team bound to
+  its **own** Bot. One Agent binds at most one Bot, so the second Bot is what
+  makes the two Tasks distinguishable at all.
 
 ## Operate
 
--- a/packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md
+++ b/packages/qa/cases/cross-surface/opentag-entry-agent-materialization.md
@@ -1,6 +1,6 @@
 ---
 id: opentag-entry-agent-materialization
-description: Validate the /opentag journey end to end — strict OAuth return, local draft, one atomic Agent create on a Computer-supported runtime, readiness and lost-response recovery, and the focused Feishu handoff without onboarding completion.
+description: Validate the /opentag journey end to end — strict OAuth return, local draft, one atomic Agent create on a Computer-supported runtime, readiness and lost-response recovery, the focused Feishu handoff, and terminal completion on real first use in Feishu.
 areas: [cross-surface]
 surfaces: [server, web, client]
 ---
@@ -75,27 +75,51 @@ gates.
    confirm the QR and its confirmation link. Confirm the surface calls the Bot
    connected only once the binding is genuinely reachable, and reports a
    connection error in words rather than colour alone.
-7. **No completion.** Confirm no onboarding completion or suppression stamp was
-   written for this membership at any point.
-8. **Narrow viewport.** Repeat the main path at 390 px; the current decision
-   must stay on the first screen.
+7. **No completion before real use.** Through every step above, including a
+   Bot that is provisioned and connected, confirm no onboarding completion or
+   suppression stamp was written for this membership. A connected Bot is setup,
+   not use.
+8. **Terminal first use.** From the real Feishu tenant, send this Bot a private
+   message (or an exact group mention) so ingress creates the Task. Confirm the
+   browser converges without a reload: the entry reaches its terminal state,
+   `members.onboarding_completed_at` and the `completed` suppression stamp are
+   written exactly once, and the offered destination opens that Task's chat.
+   Reload the entry and confirm it converges to the same state with no second
+   Agent, no second chat, and no second stamp.
+9. **Another Agent's task cannot finish this one.** With a second Agent bound to
+   its own Bot in the same Team, drive that Bot's Task and confirm it does not
+   complete the first Agent's membership. Repeat with the first Agent invited
+   into the second Agent's Task as an internal collaborator: being a speaker in
+   someone else's Feishu conversation must not count as this Agent's first use.
+10. **Failure honesty.** Block the chat reads while the Task exists and confirm
+    the entry neither completes nor claims the Agent is unused, and recovers
+    once the reads succeed. Then block only the completion request and confirm
+    the terminal state stays retryable and withholds the destination until the
+    stamp lands.
+11. **Narrow viewport.** Repeat the main path at 390 px; the current decision
+    must stay on the first screen.
 
 ## Evidence
 
 Database rows for the Agent, its config and its Feishu binding at each
-checkpoint; the request log around the create and the blocked `/me`; and
-screenshots of the readiness-failure state and the Feishu step. Redact tokens,
-connect codes and Bot credentials.
+checkpoint; the membership onboarding stamp columns before and after real first
+use; the Task chat's `metadata` alongside the Agent's Bot binding id; the
+request log around the create, the blocked `/me` and the blocked reads; and
+screenshots of the readiness-failure state, the Feishu step and the terminal
+state. Redact tokens, connect codes, Bot credentials and external member
+identities.
 
 ## Expected
 
 Exactly one Agent per completed journey. Every interruption leaves either
 nothing created or that one Agent, reachable again. A Computer running a
 non-default runtime completes without a dead end, and reaching a connected Bot
-does not complete onboarding — real first use happens in Feishu.
+does not complete onboarding — only a real Feishu Task belonging to this exact
+Agent does, exactly once, and a failed read never stands in for one.
 
 ## Limitations
 
-Terminal first use in Feishu is out of scope here. This case does not qualify
-the Feishu message bridge, Bot credential handling, or the Template catalog
-itself; those have their own owners.
+This case qualifies the entry's own terminal boundary, not the transport that
+produces it: the Feishu message bridge, Bot credential handling, and the
+Template catalog have their own owners. An admin continuing a teammate's Agent
+is out of scope — that member does not own the setup being stamped.
--- a/packages/web/src/api/onboarding-events.ts
+++ b/packages/web/src/api/onboarding-events.ts
@@ -93,7 +93,9 @@ export async function reportOnboardingEvent(
 /**
  * Stamp the terminal-state `onboarding_completed_at` column. Called when
  * the user walks Step 3 to success (admin Continue, invitee Confirm /
- * Continue). Once stamped, first-run onboarding no longer auto-opens; the
+ * Continue), and by the `/opentag` entry once that Agent has a real Feishu
+ * Task — a terminal state reached by using the Agent, not by pressing
+ * anything. Once stamped, first-run onboarding no longer auto-opens; the
  * permanent Settings → Getting Started overview remains available.
  *
  * Distinct from `dismissOnboarding()`, which only hides the stepper UI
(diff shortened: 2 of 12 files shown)


baixiaohang (changes requested):
- Rationale: The exact Bot-binding check is sound, but the discovery path can permanently miss a valid first-use Task once it falls outside the first conversation-list page.

**Risk level**: B-low
- Path baseline: `packages/web/**` and QA documentation -> C
- Semantic lift: Web data-layer reads now derive a membership onboarding transition from chat state -> B-low

**PR summary**
- Problem: OpenTag currently treats a connected Feishu Bot as the end of setup even though the member has not used the Agent.
- Approach: Observe the first Feishu Task for the Agent's exact Bot binding, stamp the current membership complete through the existing idempotent endpoint, and withhold the Task handoff until that stamp succeeds.
- Impacted modules: OpenTag flow/UI, Web auth onboarding projection, chat reads, product tests, and the cross-surface QA case.

**Review findings**
❌ 1. An incomplete first page is treated as an established absence. `listMeChats` is cursor-paginated, but this read examines only the first 20 `rows` and returns `absent` even when `nextCursor` proves more matching candidates exist. A member who used the Agent earlier can therefore reload after enough newer Feishu collaboration activity has pushed that Task beyond page one; every five-second poll repeats the same page, so the promised convergence never occurs. Please paginate to an exact result (or use an exact server-side read), and do not return `absent` while the candidate set is incomplete. [R4/R5 / `packages/web/src/pages/opentag/first-use.ts:49`]

⚠️ 2. The terminal shell says “Setup is done” as soon as a Task is found, before the completion stamp has landed and even while the page is showing “We couldn't finish setting up your workspace.” Please keep the pre-stamp copy neutral or gate the completed wording on `completeSettled`. [R4 / `packages/web/src/pages/opentag/copy.ts:43`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):

This PR moves `/opentag` completion from “Bot connected” to the first real Feishu Task, then stamps the current membership through the existing idempotent completion seam and exposes the Task only after that stamp lands. The exact Bot-binding check correctly distinguishes the primary Agent's Task from a neighbouring Agent's Task, and there is no schema migration, but this changes the meaning of the core per-membership onboarding state.

**Blocker — prove member ownership before stamping the membership**

`ownsUrlAgent` is computed at line 195, but the first-use query and the completion effect at lines 240–269 do not require it. An org admin is intentionally allowed to continue a teammate's Agent, so `facts.state === "resolved"` is not proof that this membership owns the setup. Task visibility is not that proof either: when the primary Agent invites another internal Agent, the ordinary watcher projection gives that collaborator's manager a human membership in the same Task. If that manager is an admin and opens `/opentag?agent=<primary>`, `listMeChats({ with: [primary] })` can return the Task, the Bot binding matches exactly, and `markOnboardingCompleted()` stamps the collaborator manager's membership even though the primary Agent belongs to someone else. That can suppress or bounce the wrong member's onboarding based on another member's Agent use.

Please gate the completion-producing read/write on authoritative ownership (`agent.managerId === memberId`, or an equivalent server-side check), and add coverage for an admin who can see the primary Agent's Task through a managed collaborator. The Bot binding proves which Agent owns the Task; it does not prove which member owns the onboarding state being written. [`packages/web/src/pages/opentag/opentag-page.tsx:240`]

The already-submitted review also blocks on incomplete pagination and flags the pre-stamp “Setup is done” copy; I did not duplicate those findings here.

I reviewed the diff and current green CI evidence without rerunning tests, per review scope.

Gandy2025:
Both blockers were real. Fixed in `e2df4bd`.

**Ownership (@yuezengwu).** Confirmed by tracing it: `canManageAgentDetail` admits any admin regardless of `managerId`, and `recomputeChatWatchers` inserts a watcher row for the manager of *any* non-human speaker. So an admin who manages a collaborator Agent invited into the primary's Task gets a membership in that Task, the list returns it, the binding matches exactly — and the stamp lands on their membership because a teammate used a teammate's Agent.

You put it exactly right: the Bot binding proves which Agent owns the Task, not which member owns the onboarding state. Both the read and the write now require `agent.managerId === memberId`, and `memberId` is part of the query key so a cached answer cannot outlive the member it was read for. Gating the read too means the question is not even asked. Continuing a teammate's Agent still works; it just is not that member's onboarding to finish.

New test: an admin who can see the primary's Task through a managed collaborator — asserts no stamp, and that the list is never even queried.

**Pagination (@baixiaohang).** Also confirmed: `rows` is activity-ordered with `nextCursor`, so a member who used their Agent a while ago and has since collaborated in busier Feishu conversations has their own Task pushed off page one. Worse than a miss — every 5s poll re-read the same page, so it could never recover.

The read now follows the cursor and returns `absent` only once the list is exhausted. Exhausting the page budget returns `unknown` instead, which keeps polling rather than concluding from an unfinished scan. Two tests: one walks the cursor to a Task on page two, one asserts a never-ending list yields `unknown`, not `absent`.

**Copy.** Dropped "Setup is done." from the step heading. The shell renders it for the whole step, so it sat directly above "We couldn't finish setting up your workspace" whenever the stamp failed. The heading now says only what the Task itself establishes.

Local: 252 files / 2375 tests pass, typecheck and the design-token gate clean.

One thing worth stating plainly, since it changed shape under review: the admin-continuing-a-teammate's-Agent case is now a deliberate non-completion rather than an accident. That matches the Context Tree's rule that completion means the member owns the setup.

baixiaohang (changes requested):
- Rationale: Ownership and copy are fixed, but the pagination repair still leaves a permanent discovery horizon for valid Tasks beyond page ten.

**Risk level**: B-low
- Path baseline: `packages/web/**` and QA documentation -> C
- Semantic lift: Web data-layer reads derive a membership onboarding transition from chat state -> B-low

**PR summary**
- Problem: OpenTag currently treats a connected Feishu Bot as the end of setup even though the member has not used the Agent.
- Approach: Observe the first Feishu Task for the member-owned Agent's exact Bot binding, stamp that membership through the existing idempotent endpoint, and expose the Task only after the stamp succeeds.
- Impacted modules: OpenTag flow/UI, Web auth onboarding projection, chat reads, product tests, and the cross-surface QA case.

**Review findings**
❌ 1. The pagination blocker remains beyond a larger fixed horizon. After ten pages, `readOpenTagFirstUse` returns `unknown`; React Query's next poll calls it from scratch, so it rescans pages 1–10 and can never reach a matching Task on page 11 or later. This avoids a false `absent`, but it does not restore the promised reload/poll convergence and can also repeat up to 500 chat-detail reads every five seconds. Please either query the exact Bot binding server-side, scan to exhaustion with cancellation, or preserve progress so later polls continue from the unread cursor. Add coverage proving that hitting the per-read budget eventually advances to and finds a later Task, rather than only asserting `unknown`. [R4/R5 / `packages/web/src/pages/opentag/first-use.ts:63`]

✅ 2. The ownership repair now gates both the discovery read and the completion write on `agent.managerId === memberId`, with the member in the query key and focused collaborator-manager coverage. [R4 / `packages/web/src/pages/opentag/opentag-page.tsx:240`]

✅ 3. The terminal heading no longer declares setup complete before the completion stamp succeeds. [`packages/web/src/pages/opentag/copy.ts:43`]

**Action taken**
- Submitted request changes on head `e2df4bd`.

yuezengwu (approved):
**Recommendation**: approve

The incremental fix at `e2df4bd` resolves my blocker and the two previously reported review points:

- The first-use read is disabled unless the current membership owns the URL Agent, the member is included in the query key, and ownership is checked again at the completion write. This closes the collaborator-watcher/admin path that could stamp the wrong membership.
- The candidate scan follows `nextCursor`; it returns `absent` only after exhausting the list and returns `unknown` when its page budget is exhausted, so an incomplete scan cannot become a durable false negative.
- The terminal shell no longer says setup is done before the completion stamp succeeds.

The focused tests cover the ownership path, a Task on a later page, and a bounded-but-incomplete scan. I reviewed the incremental diff and current CI evidence without rerunning tests, per review scope.

Non-blocking documentation note: the PR body's known-limit sentence still says a teammate's Task “is visible only to the Agent's manager.” The fix correctly established that visibility can also come through collaborator watcher projection; non-completion now comes from the explicit owner gate. Please update that sentence so the PR rationale matches the implemented boundary.

yuezengwu (changes requested):

Correction to my immediately preceding approval: the concurrently submitted review identified a valid remaining pagination blocker that I had not accounted for in the verdict.

After `FIRST_USE_MAX_PAGES`, the read returns `unknown`, but the next React Query poll starts again without the unread cursor. A valid Task on page 11 or later therefore remains permanently unreachable: every poll repeats pages 1–10, potentially including the same 500 chat-detail reads. Returning `unknown` avoids a false absence, but it does not provide the promised convergence.

Please make the exact Bot-binding lookup server-side, scan to exhaustion with an appropriate cancellation boundary, or carry the unread cursor forward across polls. Coverage should demonstrate that a match beyond one per-read budget is eventually reached, not only that the first budget returns `unknown`.

The ownership and pre-stamp copy fixes remain verified.

Gandy2025:
Fixed in `1588130`. You were both right that the budget only moved the horizon.

**Pagination.** The scan now runs to the end of the list — no page bound at all. What makes exhaustion affordable is that verdicts keep: a chat's Bot binding is written once by ingress and never changes, so a candidate ruled out stays ruled out for the life of the page. Later polls walk the list again but pay only for conversations they have not seen, instead of repeating hundreds of chat reads every five seconds to conclude nothing. A superseded scan stops on the query's abort signal and reports `unknown`, since abandoning a search establishes nothing either way.

I did consider the server-side exact lookup, which would be the better shape — but it needs a new endpoint, and this task's scope explicitly excludes Server changes. Happy to file it as a follow-up if you'd rather have it there.

Coverage now proves the horizon is gone rather than that the budget was honest, which is what you asked for:
- a Task on page 41 is found in one read (would have been unreachable under any fixed budget);
- a second scan finds a newly arrived Task without re-reading the two it already ruled out — this is the one that pins the cost fix;
- an aborted scan returns `unknown` and reads no chats.

**@yuezengwu — the PR body sentence.** Corrected, and thank you for catching it: the old wording still credited non-completion to visibility, which your own finding disproved. It now says the boundary is the explicit owner gate, and names the collaborator-watcher path as the reason visibility alone was never sufficient.

Local: 252 files / 2377 tests pass; biome, typecheck and the design-token gate clean.

yuezengwu (approved):
**Recommendation**: approve

The incremental fixes through `1588130` close the remaining pagination horizon:

- The scan now follows cursors to exhaustion, so a valid Task beyond the former ten-page budget is reachable in the same read.
- Superseded queries receive React Query's `AbortSignal`, avoiding continued list traversal after the `(member, Agent, Bot)` key changes.
- A per-key verdict cache remembers chats whose immutable Bot binding was already ruled out, so later polls do not repeat their chat-detail reads.
- Focused coverage places the matching Task well beyond the former budget, verifies ruled-out candidates are not fetched twice, and exercises scan cancellation.

The previously verified ownership gates and pre-stamp copy remain intact. I reviewed the incremental diff and current CI evidence without rerunning tests, per review scope.

Non-blocking documentation note: `first-use.ts` still says the verdict cache is per member “by construction,” while `1588130` intentionally makes it per `(Agent, Bot)`. The latter is sound because cached Bot-binding facts are viewer-independent; the comment should match it. The PR body has already been corrected to describe the explicit owner gate accurately.

yuezengwu (approved):
**Recommendation**: approve

`ce63f7f` only corrects the scanner documentation to match the implemented `(Agent, Bot)` verdict-cache lifetime. The cached fact is viewer-independent, while the separate query key and ownership gates continue to control which membership may act on it. No behavior changed and the previously approved pagination, ownership, and pre-stamp boundaries remain intact.

I reviewed this incremental diff without rerunning tests, per review scope.

Gandy2025:
Comment corrected in `ce63f7f` — you're right, the docblock was still describing the shape from before `1588130`. It now says per `(Agent, Bot)` and explains why that is sound: what the cache remembers is viewer-independent, and ownership is settled before the scan is ever run.

Thanks for the three rounds on this. The wrong-member stamp in particular was a real defect that I had reported to my requester as the opposite — a case that could only ever under-complete.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The latest scan removes the permanent pagination horizon while preserving the member-ownership boundary and avoiding repeated chat-detail reads across polls.

**Risk level**: B-low
- Path baseline: `packages/web/**` and QA documentation -> C
- Semantic lift: Web data-layer reads derive a per-membership onboarding transition from Feishu Task state -> B-low

**PR summary**
- Problem: OpenTag previously treated a connected Feishu Bot as completed onboarding before the member had actually used the Agent.
- Approach: Observe the first real Feishu Task for the member-owned Agent's exact Bot binding, stamp that membership through the existing idempotent completion seam, and expose the Task only after the stamp succeeds.
- Impacted modules: OpenTag flow and UI, Web auth onboarding projection, chat reads, focused product tests, and the cross-surface QA case.

**Review findings**
✅ 1. The candidate scan now follows cursors to exhaustion, so a valid Task beyond the former fixed budget is reachable; an aborted scan remains `unknown` instead of claiming absence. [R4/R5 / `packages/web/src/pages/opentag/first-use.ts:63`]
✅ 2. The scanner memoizes only immutable negative Bot-binding facts for its `(Agent, Bot)` lifetime. Later polls still re-list candidates so newly arrived Tasks are discoverable, but they do not repeat chat-detail reads for previously rejected rows. [R1/R4 / `packages/web/src/pages/opentag/first-use.ts:60`]
✅ 3. The previously verified safeguards remain intact: both discovery and completion require `agent.managerId === memberId`, and the terminal shell no longer declares setup complete before the stamp lands. [R4 / `packages/web/src/pages/opentag/opentag-page.tsx:264`]

I reviewed the incremental diff on head `ce63f7f` and did not rerun tests, per review scope.

**Action taken**
- Approved.

PR #2314, opened 2026-08-13
feat: expire message attachments after 14 days

yuezengwu:
## Summary

- add a separate attachment-retention sweep that runs once at Server startup and then every 24 hours
- make old, ready attachments eligible only when an actual message references them, covering Web chat/ask/new-chat, Agent send/receive, and Feishu inbound message shapes
- permanently protect attachments while a Team Skill Resource or Agent Template bundle still references them
- preserve immutable historical messages and render explicit expired-or-unavailable states in Web and Agent clients
- keep the existing 24-hour hourly orphan sweep unchanged

## Rollout and data impact

Deletion is disabled by default. Without `FIRST_TREE_ATTACHMENT_RETENTION_DELETE_ENABLED=true`, each run only logs eligible object and byte totals, including zero-candidate runs.

When enabled, this change deletes eligible attachment rows and their PostgreSQL bytes. Eligibility requires an actual message reference; old generic attachments with no message reference stay outside this retention sweep and remain owned by the existing release/orphan lifecycle. There is no schema change, migration, index, constraint, default, backfill, or message rewrite.

Turning deletion off stops future removals; already-deleted bytes require a database backup to restore. Browser caches and Agent-local files are not remotely erased. The delete path is bounded to 100 rows per batch and 10 batches per run, uses row locks, and rechecks both message classification and Resource/Template protection inside the delete transaction. The exact cutoff is strict: an attachment created exactly 14 days ago is retained; one older than the cutoff is eligible.

## User experience and reliability

- Web image, lightbox, file-chip, and document-preview surfaces distinguish attachment 404s from transient failures and show the shared 14-day retention explanation
- failed file downloads no longer disappear silently; reclaimed files become disabled and non-404 failures remain retryable
- Agent receipt continues without interruption; a delivery-local 404 marker selects the retention placeholder while cached local files continue to work
- retry attachment rematerialization rechecks the full terminate/reset admission fence after network I/O, preventing a stale retry from registering teardown debt or starting a route after termination has completed

## Validation

Independent QA: PASS (`focused-local`) on exact head `a93387666`.

- fresh PostgreSQL 16 scenario validated all four message-reference shapes, excluded same-age generic no-message rows, retained Resource/Template-protected rows, preserved all historical message content/metadata, and released quota only for six eligible rows / 86 bytes
- strict cutoff, dry-run, Resource release with and without a remaining message reference, quota accounting, and transaction-time protection checks passed
- representative post-`VACUUM (ANALYZE)` plan used the existing btree and two GIN indexes via `BitmapOr`; 5,000 candidates over 55,000 messages completed in 24.5 ms, and the exact service dry-run completed in 166 ms
- a synthetic immediate post-write probe with an uncleared GIN pending list took 11.1 s; this is retained as a rollout/performance observation rather than a production SLO claim
- deferred attachment-fetch vs terminate race passed 10/10 repetitions; no stale debt, old-handler shutdown, or replacement route was observed
- complete Server suite: 290 files / 3340 tests
- complete Client suite: 198 files / 2509 tests, with 2 host-smoke files and 7 tests skipped by suite policy
- affected Web attachment surfaces: 26/26; shared attachment/config tests: 27/27
- shared, Server, Web, and Client typechecks; Client directory/provider boundary guards; `pnpm check`; `git diff --check`
- prior real-Web verification covered reclaimed image, disabled file-chip, and document-preview states with no browser warnings or errors; the reviewer fix only changes the shared numeric source, so the visible copy remains unchanged

## Context Tree

Companion draft: [first-tree-context#940](https://github.com/agent-team-foundation/first-tree-context/pull/940). It records the durable retention and protected-bundle decision and will be reconciled after this source PR lands.


baixiaohang (changes requested):
- Rationale: The retention behavior is well-covered, but the retry-path attachment refetch introduces an await gap that can violate the existing terminate/reset teardown fence.

**Risk level**: A

**PR summary**
- Problem: Message attachments consume Cloud PostgreSQL quota indefinitely, while users and agents currently receive ambiguous failures if bytes disappear.
- Approach: Add an opt-in daily 14-day retention sweep, keep immutable message references and durable Skill/Template bundles, and render explicit unavailable states across Web and Agent clients.
- Impacted modules: Server attachment lifecycle/background tasks, shared server config, Client inbound/retry rendering, Web attachment surfaces, operator docs.

**Review findings**
❌ 1. Re-run the retry eligibility/admission CAS after attachment rematerialization. `ensureImagesLocal()` can await network I/O, so a concurrent Reset/terminate/shutdown may clear or replace this session and finish draining before the stale retry registers teardown debt at line 3838. The stale retry can then add debt and shut down the old handler after termination has already acknowledged; a fresh route may start in that gap while the old teardown is still pending or ultimately fails. Repeat the full fence/session-entry/status check immediately after this await and before `registerPendingTeardown`, and add a race test that holds the attachment fetch while terminate completes. [R4 / packages/client/src/runtime/session-manager.ts:3826]
⚠️ 2. Make the retention selector prove that a row is message-class instead of defining it as “every old ready attachment except today’s two bundle consumers.” Attachments are a generic primitive; this negative classification silently makes any future durable consumer expiring unless it updates this sweep. Requiring an actual message reference (or an explicit lifecycle class) keeps the 14-day policy scoped to the contract. [R3/R5 / packages/server/src/services/attachment.ts:575]
⚠️ 3. Keep the 14-day policy in one shared constant. The server cutoff and the Client/Web user-facing copy currently hard-code the same business fact independently, so a policy change can make clients report a retention window different from what the Server enforces. Export a day-count constant from `@first-tree/shared`, derive the millisecond cutoff from it, and interpolate it into both surfaces. [R1 / packages/server/src/services/attachment.ts:546]

**Action taken**
- Submitted request changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up resolves the retry-admission race, positively scopes retention to message-referenced attachments, and centralizes the retention policy without introducing new review findings.

**Risk level**: A

**PR summary**
- Problem: Message attachments consume PostgreSQL quota indefinitely, while users and agents need an explicit, reliable state after retained bytes are reclaimed.
- Approach: Add an opt-in 14-day message-attachment retention sweep, preserve immutable message history and durable bundle references, and render reclaimed attachments explicitly across Web and Agent clients.
- Impacted modules: Server attachment lifecycle/background tasks, shared retention configuration, Client inbound retry handling, Web attachment surfaces, and operator documentation.

**Review findings**
✅ 1. Attachment rematerialization is now followed by the full retry-admission fence, with a terminate-race regression test covering the await gap.
✅ 2. Retention eligibility now requires an actual message reference and is rechecked together with Resource/Template protection inside the delete transaction.
✅ 3. The 14-day policy now has one shared source consumed by Server, Client, and Web.

**Action taken**
- Approved.

PR #2318, opened 2026-08-13
test: isolate Feishu lease renewal timing

baixiaohang:
## Summary

- revoke Feishu bindings left active by earlier registration tests so background maintenance cannot leak into later assertions
- pause the app-owned Feishu manager while the accelerated lease-renewal test runs its dedicated manager
- clean up the dedicated binding and always restore the app-owned manager

This addresses the timing-dependent main-branch failure in [CI run 31691822984](https://github.com/agent-team-foundation/first-tree/actions/runs/31691822984/job/94420655174), where an unrelated shared `disconnect` spy could release the lease wait before its expiry advanced.

## Validation

- [x] `pnpm --filter @first-tree/server exec vitest run src/__tests__/feishu-registration.test.ts --maxWorkers=1` (14 tests)
- [x] repeated the full test file five additional times with one worker; all passed
- [x] `pnpm exec biome check packages/server/src/__tests__/feishu-registration.test.ts`
- [x] `pnpm --filter @first-tree/server typecheck`
- [ ] full repository checks (left to CI per local validation policy)

## Change Surface

## Notes

- docs or tests updated to match: test isolation only; product behavior is unchanged
- follow-up work: none

### DIFF
--- a/packages/server/src/__tests__/feishu-registration.test.ts
+++ b/packages/server/src/__tests__/feishu-registration.test.ts
@@ -212,6 +212,7 @@ describe("official Feishu QR registration", () => {
       .from(imBotBindings)
       .where(eq(imBotBindings.id, stored?.id ?? "missing"));
     expect(afterStaleEvent?.lastEventAt).toBeNull();
+    await app.feishuIntegration.revoke(a.agent.uuid);
   });
 
   it("keeps the Bot connected when profile metadata cannot be loaded", async () => {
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed at `953e0494c00a23e0c93112bfd16550511411f129`. No blocking findings.

The test-only change matches the observed failure mode: earlier app-owned bindings are revoked, the app-owned manager is stopped and shared mock history is cleared while the accelerated lease test owns its dedicated manager, and cleanup restores the app manager even if dedicated cleanup fails. This removes the unrelated `disconnect` signal that could release the lease wait before the expiry advanced. No production behavior, core data structure, or database schema is changed.

I did not rerun tests as part of this diff review; the PR's focused validation and current CI provide the execution evidence.

PR #2321, opened 2026-08-13
feat(web): prepare the Agent's Feishu tools automatically in OpenTag Step 3

Gandy2025:
The thinnest layer that makes the existing Feishu setup Task automatic.

Choosing Feishu is the only thing OpenTag asks the member for. After that, Step 3 calls the setup-chat endpoint that already exists, and shows the two facts the product already had: the Bot binding, and the CLI capability the Client already reports. No install action reaches the member.

## What it adds

**A trigger.** Once a Bot binding exists — the durable proof the member chose Feishu — the page calls the existing endpoint. Nothing before that, nothing for a teammate's Agent, and nothing for an Agent with no Computer to prepare.

**Idempotency on the seam that already does it.** The endpoint now passes `chats.onboarding_kickoff_key` to the keyed `createChat`, keyed on human + Agent + Client, so loads, reloads and extra tabs converge on one private Task instead of a pile. Moving the Agent to another Computer is a different machine to check and earns its own Task.

**A retry that retries.** Reuse alone wakes nobody once the Agent has taken the original request, so an explicit `retry` appends one further ask in that same Task. Every ensure-shaped call stays inert. Agent Detail's existing **Ask Agent to install** is a member asking, so it carries the same intent and remains the one permanent manual entry.

**One gate.** Real first use and the onboarding completion stamp both require the exact Agent's Bot to be reachable *and* its CLI ready. A reachable Bot with no working CLI receives messages the Agent cannot answer, and the Task that creates would otherwise finish setup on a handoff that does not work.

**A way out.** A failed setup call, a failed Bot, no Computer, or a wait past ninety seconds offers `Finish later` — the Agent's own page, onboarding deliberately unfinished. `Try again` appears only where a Computer could still finish and the tools half is what stalled.

## What it does not add

No installer, endpoint, DTO beyond one `retry` flag on the existing request, schema, migration, persistence, readiness model, notification, Inbox change, second setup surface, or rollout flag. Steps 1, 2 and 4 are untouched, and the shared onboarding completion contract is unchanged — that one is #2323.

## Verification

`pnpm check`, `pnpm typecheck` including the Web design-token gate, and the full `@first-tree/web` and `@first-tree/server` suites.

Server: one Task across repeated calls with a single opening message; a new Task after the Agent moves Computer, converging back on return; one administrator kept out of another's Task; an explicit retry adding one message and one wake; ensure-shaped calls adding neither; a burst collapsing to one.

Web: no call before a Bot exists, for a teammate's Agent, or for an Agent with no Computer; one call per Agent and Computer across binding polls; a failed call reported as failed rather than as work in progress; the step holding while the Bot is reachable but the CLI is not; the invitation to message the Bot withheld until both halves are ready; each established failure offering the way out at once; and the tools retry scoped to the tools half.


baixiaohang (changes requested):
- Rationale: The recovery CTA currently resolves the HTTP request without restarting any Agent work, so the advertised retry path cannot recover a stalled setup.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the durable `chats.onboarding_kickoff_key` format/semantics for Feishu setup Tasks.
- Semantic lift: none.

**PR summary**
- Problem: OpenTag members currently have to understand and manually trigger Agent-side Feishu CLI preparation, and repeated entry points can create duplicate setup Tasks.
- Approach: Step 3 starts Bot registration and Agent-tool preparation in parallel, derives readiness from the binding plus CLI capability, and reuses one private Task per human, Agent, and Client.
- Impacted modules: Feishu setup-chat API, keyed Task creation, OpenTag flow/readiness UI, and related server/Web tests.

**Review findings**
❌ 1. **“Try again” reuses the Task but does not retry the work.** The new key makes `createChat` return the existing opening message on every repeated request. Because this call supplies no `onTaskReuse`, that branch returns `recipients: []`; the following `notifyRecipients` therefore wakes nobody and no new message is sent. In the failure/slow states, clicking **Try again** can report a successful request and change the UI, but the Agent receives no new work. The automatic ensure call and an explicit recovery retry need distinct reuse behavior (or an equivalent atomic/staleness guard), so reloads/tabs remain no-ops while a real retry re-kicks the existing Task without creating a duplicate. [`packages/server/src/api/agents-feishu.ts:58`]

**Action taken**
- Submitted request changes.

yuezengwu (commented):
Blocking: the slow-path **Try again** action does not retry any Agent work.

After the automatic POST has succeeded once, the keyed `createChat` reuse branch returns the existing opening message with `recipients: []` because this endpoint supplies no `onTaskReuse`. Clicking **Try again** after `FEISHU_TOOLS_SLOW_MS` therefore only gets another 201 and refetches the binding; it sends no follow-up and wakes nobody. `toolsSlow` also stays `true`, so the UI immediately remains in the same recoverable state after the request settles.

Please give an explicit recovery retry distinct reuse behavior (with an atomic/staleness guard so reloads and extra tabs stay no-ops), or route the member into the existing Task instead of promising a retry. A regression test should click **Try again** after the slow threshold and assert that the existing Task is actually re-kicked without creating another chat.

Gandy2025:
Both reviews were right, and the retry finding was the same defect from two angles. Fixed in `312726de7`, along with three more the same reading turned up.

### The retry now retries

`onTaskReuse` was the missing half. An explicit retry sends the same request again inside the same Task; re-arming the original inbox entry would not have been enough, because once the Agent has taken that delivery, flipping a flag on it wakes nobody.

The two callers are now distinguishable — `retry` on `createFeishuSetupChatSchema`. Ensuring the Task exists is what every load, reload, and extra tab does, and it stays a no-op. A short cooldown collapses a burst, so two tabs on the recovery state or one double click stay a single ask. `Ask Agent to install` on Agent Detail is a member asking, so it retries too.

The route also switched to the seam's `notificationMessageId`: on reuse, `message` is the original opening message, and signalling that with the retry's recipients would wake them for something already delivered.

### The key ends in a fixed token

`${human}:${agent}:${client|"unbound"}:feishu-cli`. `clients.id` is client-supplied text and this column is read by suffix, so an arbitrary trailing segment could have matched `hasTreeSetupKickoffMessage`'s `%:tree-setup` / `%:tree` scan and silently retired an organization's Context Tree prompt. `docs/development/onboarding-kickoff-contract.md` now records the namespace and why the suffix is load-bearing.

### Three more, found while verifying those

**Opening a teammate's Agent started work on their Computer.** `readinessSettled` is satisfied by `!ownsUrlAgent`, so an admin merely looking at a teammate's Agent reached the automatic call and sent an install prompt to a machine that is not theirs. The automatic path now carries the same ownership gate as the reads and the stamp; the step itself still renders, because helping with the Bot half is legitimate.

**The completion stamp was not re-checked against the capability it depends on.** A scan begun while the CLI was ready could resolve after it dropped and stamp onboarding on a handoff that no longer works. It is re-checked at the write now — and a Task found before that capability goes no longer carries the member to a destination step whose stamp will never land, which would have stranded them on "Finishing up…" with the recovery on the step behind them.

**The recovery affordance could never arm for some members.** It was timed from the automatic request, but a Computer that already has the CLI sends no request — so a member whose Bot never confirms had no exit but Sign out. It is timed from committing to Feishu now. `Try again` is withheld in that state: the automatic preparation has nothing left to do, and the outstanding half is a confirmation this page cannot press.

### Coverage

Server: a retry adds a second ask in the same Task with a second inbox wake; ensure-shaped calls leave the Task at one message; three rapid retries collapse to one. Web: the non-owner case, the completion re-check across a capability drop mid-scan, the Bot-stuck-with-ready-tools exit, the slow-state retry carrying `retry: true`, and the panel keeping both halves named while both are outstanding.

Both suites and the design-token gate are green, and the reworked states were re-checked in a browser against the approved Step 3 contract.

One note I did not act on: the `slow`-state `toolsSlow` stickiness is deliberate. A wait that has already been long does not become short again, so the affordances stay available after a retry rather than disappearing and needing another 90 seconds to come back.

yuezengwu (changes requested):
The original retry blocker is fixed: explicit retries now add a new ask to the existing Task and wake the Agent, while ensure-shaped calls stay inert. Two blockers remain in the follow-up.

1. **The completion gate is still client-side and races the durable stamp.** `isFeishuHandoffUsable(binding)` is checked immediately before calling the generic `/me/onboarding-completed` endpoint, but that endpoint does not receive or verify this Agent, Bot binding, or current CLI capability. If the capability drops after line 374 and before the POST commits, the server still stamps onboarding complete. The new test only drops the capability before `markOnboardingCompleted()` is invoked, so it does not cover this window. Please make the terminal write server-authoritative: revalidate ownership, the exact binding's reachability, and the current Client's CLI readiness as part of the operation that writes the membership stamp (and cover a drop while the completion request is in flight). [`packages/web/src/pages/opentag/opentag-page.tsx:374`]

2. **The new Bot-stuck recovery state reports the wrong outstanding half.** When the CLI is already `ready`, the Bot is still provisioning, and the 90-second threshold fires, `feishuToolsDelayedCopy(false)` says “Your Agent's Computer is still preparing…” even though the readiness row says Agent tools are ready and `Try again` is correctly withheld. The added Bot-stuck test checks only the actions, so this contradiction passes. Please derive the heading from both halves and use Bot-specific copy when the Computer is already ready. [`packages/web/src/pages/opentag/copy.ts:68`]

baixiaohang (changes requested):
- Rationale: The explicit retry now wakes the Agent correctly, but the Web state is not scoped to the Agent/Computer and still renders contradictory recovery states for two supported transitions.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the durable `chats.onboarding_kickoff_key` semantics for Feishu setup Tasks.
- Semantic lift: none.

**PR summary**
- Problem: OpenTag members should only choose Feishu; the system should prepare the Bot and the Agent's tools without duplicate or inert Tasks.
- Approach: The update distinguishes ensure from retry, re-kicks an existing Task atomically with a cooldown, gates automatic work by ownership, and expands recovery to either outstanding half.
- Impacted modules: Feishu setup-chat service/API/DTO, keyed Task contract, Agent Detail, and OpenTag readiness/recovery state.

**Review findings**
❌ 1. **A successful request permanently suppresses setup for a later Agent/Computer identity.** `prepareToolsSettled` is component-wide and stays true after the first mutation succeeds; it is never reset or keyed by `agentUuid` / `binding.cli.clientId`. If the Agent moves to another Computer while this page remains mounted (or the route changes to another Agent in the same mount), the binding poll can report that new machine as unprepared, but the effect exits at line 272 and never calls the endpoint with the new keyed identity. The server can create the correct per-Computer Task only if Web actually asks. Scope/reset the mutation guard with the Agent and Client identity, and cover the live move that the PR contract promises. [R5 / `packages/web/src/pages/opentag/opentag-page.tsx:264`]

❌ 2. **The Bot-stuck / tools-ready recovery state says the opposite of the readiness rows.** The new supported state at `opentag-page-dom.test.tsx:1358` offers `Finish later` when the CLI is ready but the Bot is still provisioning. Because the shell heading is derived only from `botReachable`, it renders “Your Agent's Computer is still preparing the tools…” while the `Agent tools` row says `Ready`; the only outstanding action is the Feishu confirmation. Derive the heading from both halves (as the panel already does) and assert the shell copy in that regression. [`packages/web/src/pages/opentag/copy.ts:62`]

❌ 3. **A failed POST can keep recovery active after the whole handoff becomes usable.** `recovery` is `offered` whenever `callFailed` is true, even when `isFeishuHandoffUsable(binding)` is now true. A lost/failed response followed by capability convergence therefore leaves “One last part is taking longer” and `Finish later` beside an invitation to message a ready Bot. The current comment says a finished handoff has nothing to recover from; make readiness outrank request history for `recovery`, not only for the `tools` row. [`packages/web/src/pages/opentag/flow.ts:217`]

✅ 4. The original blocker is fixed: an explicit retry adds a new message in the existing Task, returns its notification message id, and leaves ensure-shaped reuse inert.

**Action taken**
- Submitted request changes on `312726de`.

Gandy2025:
Three of the four are fixed in `8678c19d7`. The fourth I'm declining as out of scope, with reasoning below.

### Fixed

**The Bot-stuck heading contradicted its own rows** (both reviewers). It said the Computer was "still preparing" directly above a row reporting it ready, with `Try again` correctly withheld. The heading now reads both halves: a ready Computer with an unconfirmed Bot says "Your Agent is ready. Feishu hasn't confirmed the Bot yet." The regression asserts the shell copy, not only the actions.

**A failed request kept recovery up after the handoff converged.** Readiness now outranks request history for the offer, not only for the `tools` row — the Agent may well have recovered on its own, and a way out of something that already works, beside an invitation to go and use it, is the contradiction my own comment claimed was impossible. Covered for slow, failed, and both.

**The automatic request was remembered per page, not per Agent and Computer.** It is keyed on `(agent, cli.clientId)` now. The Task the server keeps is keyed to that exact pair, so a live move left the member waiting on a check nobody had asked for on the new machine. Covered by a test that moves the Client under a mounted page and asserts a second ensure call.

### Declined: making the completion write server-authoritative

The race is real and correctly described. I'm not fixing it here because the fix is outside this change's accepted scope, not because it isn't a gap.

`/me/onboarding-completed` takes `{ organizationId }` and stamps the membership. It is the shared terminal write for the standalone admin journey, the invitee path, and the team-agent quick start as well as OpenTag. Teaching it to verify a Feishu binding and a Client's CLI capability means either Feishu-specific parameters on that generic contract or a second endpoint — both named in this task's exclusions ("no new endpoint / DTO", "no changes to the existing OAuth / Team / Runtime main flows"), and it would put one provider's readiness inside the completion path every other onboarding shape goes through.

Worth weighing against what the window actually is: the binding this gate reads comes from a 10-second poll, so the client-side check is approximate by construction — the design this task asked for ("reuse the existing binding polling"). Closing the sub-request race while the input can be ten seconds stale would not make the guarantee real. On `main` today there is no CLI gate on completion at all; this change adds one at the read, at the write, and at the step transition.

If the durable guarantee is wanted, it is a genuine follow-up on the completion contract rather than something to graft on here — happy to take it as its own task with the shared endpoint in scope.

yuezengwu (changes requested):
The three in-scope fixes are correct, and I accept the server-authoritative completion race as a separate follow-up because closing it requires changing the shared completion contract explicitly excluded here. One identity-scoping blocker remains.

**The old Computer's failed request still contaminates the new identity.** `askedToolsFor` is now keyed by `(agentUuid, cli.clientId)`, but the request result passed into the state machine is still the component-wide `prepareTools.isError`. Reproduction: the ensure for Client A fails; while the page stays mounted, move the Agent to Client B where the CLI is already ready but the Bot is still provisioning. Because Client B needs no ensure call, the mutation is never reset, so line 313 passes Client A's failure into Client B's state. `resolveOpenTagFeishuStep` immediately offers recovery and delayed copy during what should be Client B's ordinary Bot-confirmation wait, before the 90-second threshold. The same leak occurs when the query-string Agent changes without remounting.

Please scope/reset the mutation outcome together with `toolsIdentity` (and cover a failed old identity followed by a ready new identity with an outstanding Bot). [`packages/web/src/pages/opentag/opentag-page.tsx:313`]

Gandy2025:
Fixed in `c92a94ed4`, and thank you for the follow-up acceptance.

The request outcome was component-wide while the request itself was keyed to an Agent and a Computer, so the two disagreed exactly where you said. A failure is now attributed to the identity it was asked for: `prepareTools.isError && askedToolsFor === toolsIdentity`. Once either half of that pair moves on, the old machine's failure says nothing about the new one — and the case that made it stick is precisely the one where the new Computer arrives ready and never asks again, so nothing would have cleared it. A query-string Agent change is covered by the same expression, since `toolsIdentity` carries the Agent too.

The regression walks your reproduction: ensure fails on `client-1`, the Agent moves to `client-2` where the CLI is ready and the Bot is still provisioning, and it asserts the QR is up, the recovery block is gone, the heading is the ordinary one, and no second ensure was made. I mutation-checked it — reverting the guard to the bare `isError` fails it on the recovery assertion.

I left the slow clock armed across the move on purpose: it measures how long the member has been on this step, and somebody who has already waited that long is still waiting, whichever machine the Agent is on now. A converged handoff clears the offer regardless, since readiness outranks both inputs.

yuezengwu (approved):
The remaining identity-scoping blocker is fixed. A failed preparation request is now attributed only while its recorded Agent/Computer identity matches the current binding, so a move to a ready Computer or a query-string Agent change cannot inherit stale recovery state. The regression covers the previously reported failed-old-Computer / ready-new-Computer transition. No further blocking findings on c92a94ed4.

Gandy2025:
Independent QA returned **FAIL** on `c92a94ed4` with two reproducible defects, both mine and both the same class the last three rounds produced. Fixed in `620279df1`.

The cause was structural, so the fix is too. The page heading, the panel header and the readiness rows each derived the same two facts separately — which is exactly how fixing one axis kept leaving the others free to disagree. They now come from one function over the resolved state, and a table-driven test asserts that no combination claims work is under way on the Computer when it is done, missing, or was never asked for. Writing that test immediately turned up a third case I had not been told about.

**A failed request was framed as a duration.** With the automatic call rejected, the heading said the Computer was "still preparing" and the panel said "First Tree will keep checking this Computer" — directly above a row reading "The automatic setup didn't start." It now reports the failure as a failure in all three places. The two DOM tests that had locked the contradiction in are corrected and now assert the absence of the work-under-way language, not just the presence of a string.

**An Agent with no Computer was treated as a slow one.** `cli.state === "offline"` means none is bound at all, but the resolver only special-cased `"ready"`, so it fell through to "preparing" while the row said "This Agent has no connected Computer" — and the member got no way out for ninety seconds, then a `Try again` that would ask a Task keyed to no machine. It is its own state now: nothing is preparing, so the step says so, offers the exit immediately because the fact is established rather than suspected, withholds the retry, and no longer sends the request at all.

QA also noted the source comment on `notifyRecipients` overstates the consequence — the wake is a backlog drain and the message id reaches only a debug log field. The distinction stays because it is the seam's contract, but I have not reworded it; say the word if you would rather it read plainly.

The full report: `focused-local`, both suites green, and 20 mutation probes across the server and Web gates — every load-bearing gate's removal was caught, with one redundant guard and the log-only field as the two survivors. Recorded limitations: no Feishu credentials, so live Bot registration and real `cli.state` transitions were never driven, and the browser evidence renders the real shipped components under a synthetic driver rather than a live `/opentag` session.

QA's case disposition is `candidate-case-update` — a `packages/qa/cases/` entry asserting these three surfaces agree in every derived state. I have not touched the committed case library, since that is a separate maintenance task.

yuezengwu (changes requested):
The two QA-reported contradictions are fixed, but the new single-derivation claim still has two holes.

1. **The Bot column still claims work is continuing when tools preparation failed or cannot start.** For every reachable Bot whose handoff is not usable, `BotColumn` says “Your Bot and its permissions are kept while the rest finishes.” That sentence still renders in both new states: `tools = recoverable/failed`, where the new heading and row correctly say setup did not start, and `tools = unavailable`, where they correctly say there is no Computer and nothing is preparing. The new DOM assertions omit this phrase, so both pages continue to contradict themselves. Please derive this supporting sentence from the resolved tools state too, or make it neutral, and assert that neither failed nor offline renders “the rest finishes.” [`packages/web/src/pages/opentag/steps/step-connect-feishu.tsx:206`]

2. **A Bot failure is collapsed into ordinary Bot waiting.** `feishuStepCopy` receives only `botReachable: boolean`, so it cannot distinguish provisioning/connecting from `binding.status === "error"` or `connectionStatus === "error"`. With CLI ready and a failed Bot connection, the Bot column and readiness row report an error, while the tools panel immediately says “Waiting for Feishu…” / “Confirming the Bot is the last part”; after the slow threshold the shell also says Feishu “hasn't confirmed” it. This is the same cross-surface contradiction the refactor is meant to eliminate. Please carry a Bot-state discriminant into the shared derivation and cover ready-tools + failed-Bot in the table/DOM regression. [`packages/web/src/pages/opentag/copy.ts:85`]

baixiaohang (changes requested):
- Rationale: The new single copy derivation still collapses a Bot failure into an ordinary wait, so the panel and recovery heading can contradict the Bot error shown immediately beside them.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the durable `chats.onboarding_kickoff_key` semantics for Feishu setup Tasks.
- Semantic lift: none.

**Review findings**
❌ 1. **The centralized copy state omits the Bot's error state.** `feishuStepCopy` receives only `botReachable: boolean`, so both `status: "error"` and `connectionStatus: "error"` take the same branch as an ordinary provisioning/connecting wait. For example, with an active Bot whose connection is `error` and a ready CLI, the Bot column and readiness row report the error and `needs you`, while the panel says `Waiting for Feishu… / Confirming the Bot is the last part`; after the slow threshold, the page heading also frames that error as a duration. With tools still preparing, the panel instead promises that both parts will finish. This is the same cross-surface contradiction the refactor is meant to eliminate. Carry the resolved Bot state (at least waiting / reachable / failed), not only reachability, into the single derivation and table-test the error combinations. [`packages/web/src/pages/opentag/copy.ts:85`]

✅ 2. The latest changes correctly distinguish a failed preparation request from a slow one and treat an Agent with no Computer as unavailable rather than preparing.

**Action taken**
- Submitted request changes on `620279df`.

Gandy2025:
Both findings are the same defect as the one before it, one level down, and both are fixed in `d56d695be`.

**The Bot half was a boolean over a three-answer question.** `feishuStepCopy` could not tell a Bot that failed from one the member simply had not confirmed, so `status: "error"` and `connectionStatus: "error"` took the ordinary-wait branch. `resolveFeishuBotState` now returns `waiting` / `reachable` / `failed`, and a failed Bot outranks whatever the Computer is doing — it is the more severe outstanding fact, so the panel names it and the heading stops calling it a duration. The table test crosses all three Bot states with every tools state and asserts no combination frames a failed Bot as merely unconfirmed ("hasn't confirmed", "Waiting for Feishu", "both parts finish"), alongside the existing sweep for work-under-way language.

**The Bot column was a third surface deriving its own answer.** "Your Bot and its permissions are kept while the rest finishes." rendered for every reachable Bot whose handoff was not usable — including failed preparation and no Computer, where nothing is finishing. It now reads from the resolved tools state and keeps that clause only while something actually is. Three DOM regressions cover it: failed preparation, no Computer, and ready-tools with a failed Bot connection.

I also dropped an unreachable branch from `620279df1` — the panel only renders once a binding exists, so the no-Bot case never reached that copy.

On the two recovery controls you flagged earlier: in the failed-Bot state the left column's `Retry` re-registers the Bot and the right column's `Try again` retries the tools Task, and they can still appear together. I have not relabelled them in this round — the surrounding copy now says which half needs attention, which is what made them ambiguous. Tell me if you want them distinguished explicitly and I will do that rather than leave it to context.

Full web suite, `pnpm check` and the design-token gate are green; the server is untouched since `c92a94ed4`. Independent QA is re-running against this exact head.

yuezengwu (changes requested):
The two reported copy contradictions are fixed, but the new failed-Bot state is not wired into the outer step state, leaving one blocker.

**A failed Bot still uses the ordinary shell and can withhold every recovery action for 90 seconds.** `feishuStepCopy("failed", …)` now returns the correct error heading, but `OpenTagPage` only applies that heading when `feishuStep.recovery === "offered"`. `resolveOpenTagFeishuStep` does not receive the Bot state, so with a failed Bot, ready (or normally preparing) tools, and `slow: false`, recovery remains `none`: the shell still says “Confirm the Bot in Feishu” directly above the new error panel. The connection-error variant is worse: `BotColumn` renders only an alert for `connectionStatus === "error"` (no Bot retry), tools are ready (so no tools retry), and `Finish later` is withheld until the 90-second timer. That is an established failure presented and actioned as an ordinary wait.

Please let Bot failure participate in the resolved step state (or otherwise independently override both heading and recovery), so the error heading renders immediately and the member gets an immediate honest way forward. Extend the new connection-error DOM regression to assert the shell does not ask for confirmation and that recovery is offered before advancing the clock. [`packages/web/src/pages/opentag/opentag-page.tsx:436`]

baixiaohang (commented):
The two UI blockers reported on `620279df` are fixed in `d56d695be`.

- Bot state is now resolved as `waiting` / `reachable` / `failed`, and the shared copy derivation gives failure precedence instead of framing it as ordinary confirmation or elapsed time.
- The reachable-Bot supporting sentence now consults the resolved tools state, so failed preparation and an absent Computer no longer promise that “the rest finishes.”
- The state-matrix and DOM regressions cover the previously contradictory Bot-error, failed-preparation, and no-Computer combinations.

No further blocking findings in this incremental review.

baixiaohang (changes requested):
**Correction to my preceding incremental review**: request changes.

The two reported copy contradictions are fixed, but I missed that the failed-Bot discriminant is not part of the outer step state.

❌ **A failed Bot still keeps the ordinary shell heading and can withhold all recovery for 90 seconds.** `feishuStepCopy("failed", …)` now derives the right error heading, but `OpenTagPage` only renders that heading when `feishuStep.recovery === "offered"`; `resolveOpenTagFeishuStep` bases recovery only on tools failure/unavailability or the slow timer. With a failed Bot and ready tools before the timer, the shell therefore still says “Confirm the Bot in Feishu” above the error panel. For `connectionStatus: "error"`, `BotColumn` has no Bot retry, ready tools have no tools retry, and `Finish later` is also absent, leaving an established failure with no action until 90 seconds elapse. Make Bot failure participate in the resolved step state so the error heading and immediate honest recovery are applied together, and extend the connection-error DOM regression to assert both. [`packages/web/src/pages/opentag/opentag-page.tsx:436`]

The earlier completion-contract scope question remains separate.

Gandy2025:
Converged by deletion in `84ee4189f`. The product owner's read was right: this change is "call the setup Task that already exists", and the back half of it had grown a second readiness state machine plus a Bot × tools copy matrix on top. Three surfaces each explained the same state, which is why fixing one axis kept freeing the others to disagree — four review rounds of the same defect class.

**Deleted:** `OpenTagToolsPrep`, `OpenTagFeishuStepState`, `OpenTagBotState`, `resolveOpenTagFeishuStep`, `resolveFeishuBotState`, `feishuStepCopy`, the shell's runtime heading override, the readiness panel's narration, the Bot column's tools-dependent branch, the whole `services/integrations/feishu/setup-chat.ts` module, and the table-driven combination tests that pinned those derivations.

**Kept — the four things this PR is for:**

1. The Web calls the existing setup-chat only once the member has chosen Feishu.
2. That endpoint create-or-reuses through the existing keyed `createChat`, keyed on human + Agent + Client; an explicit `retry` appends one ask and wake in the same Task, and every ensure-shaped call stays inert.
3. The step shows the two facts it already had — the binding and `binding.cli` — and first use plus completion need both halves.
4. A failed setup call, a failed Bot, no Computer, or a long wait offers `Try again` (only where a Computer could still finish) and `Finish later`.

Step 3 is back to one neutral heading. The two rows read their own halves directly. Recovery is four booleans on the page rather than a discriminated union. The server helper moved back inline beside the route that already held this prompt.

`16 files / +1953` → `14 files / +1295 −134`, one file deleted and none added.

Independent QA's remaining Minor from `d56d695be` — a failed Bot showing no heading change and, on the connection-error variant, no control at all until the 90-second mark — is closed by the `botFailed` boolean, with an observable test I mutation-checked.

Both suites, `pnpm check` and the design-token gate are green locally. The out-of-scope completion race is tracked separately in #2323 and is deliberately not implemented, stubbed, or TODO'd here.

Requesting a full Standards + Spec re-review on this head.

yuezengwu (changes requested):
The outer recovery blocker is fixed, but the deletion reintroduces two user-visible contradictions that the removed tests used to prevent.

1. **The standing heading is not neutral.** It still instructs the member to “Confirm the Bot in Feishu.” Once the Bot is reachable and only tools are outstanding, that sits above a `Feishu Bot / ready / Connected and reachable` row and asks them to repeat a completed action. With `connectionStatus: "error"`, it asks for confirmation above a `needs you` connection failure. The deleted runtime heading assertions covered exactly these transitions; changing only the second sentence does not make the first sentence neutral. Keep one stable heading if desired, but its wording must describe the two-fact handoff without directing an action that may already be complete or impossible. [`packages/web/src/pages/opentag/copy.ts:39`]

2. **A failed setup request is again reported as active preparation.** `setupFailed` now affects only the recovery controls. `toolsReadiness` reads only `binding.cli.state`, so after the automatic POST rejects while the binding still says `missing`/`unknown`, the page shows `Agent tools / preparing / Preparing this Computer in the background` beside `Try again`. That is the contradiction independent QA found on the previous head and violates this PR's stated rule that a failed request is reported as failed because the member is not waiting for anything. The regression was weakened to assert only the controls. Preserve the simpler structure if desired, but pass the request outcome into the row (or otherwise report the failure) and restore an assertion that failure never claims background preparation. [`packages/web/src/pages/opentag/steps/step-connect-feishu.tsx:387`]

baixiaohang (dismissed):
- Rationale: The deletion removes the copy matrix, but it reintroduces one false progress signal and lets an unrelated Bot failure re-kick ordinary tools work.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the durable `chats.onboarding_kickoff_key` semantics for Feishu setup Tasks.
- Semantic lift: none.

**Review findings**
❌ 1. **A rejected setup request is again reported as work in progress.** `toolsReadiness` now maps every `missing` / `unknown` capability to `preparing / Preparing this Computer in the background`, without the `setupFailed` fact. After the automatic POST rejects, the binding normally remains `missing`, so the row says background work is happening beside recovery copy that asks the member to try starting it again. This restores the contradiction fixed on `620279df` and conflicts with the PR contract that a failed request is reported as failed, not as a wait. Either pass the request outcome into this row or use neutral capability copy that does not claim preparation is under way. [`packages/web/src/pages/opentag/steps/step-connect-feishu.tsx:387`]

❌ 2. **A Bot failure can trigger a tools retry even when tools preparation has neither failed nor become slow.** `canRetryTools` inherits the combined `showRecovery`; therefore `botFailed && hasComputer && !toolsReady` immediately shows **Try again** for the setup Task even when its ensure call succeeded and the Agent is ordinarily preparing. Clicking it appends another ask to the Task solely because the unrelated Bot half failed. Scope the tools retry to a tools recovery reason (`setupFailed || stepSlow`) in addition to the capability checks. [`packages/web/src/pages/opentag/opentag-page.tsx:324`]

❌ 3. **The supposedly neutral heading still instructs an already-confirmed, failed connection to “Confirm the Bot in Feishu.”** For `status: "active", connectionStatus: "error"`, the Bot column correctly says the connection is failing and recovery is now immediate, but the standing shell lead above it still describes confirmation as the outstanding action. The new regression checks only the alert and controls, so the exact shell contradiction requested in the previous review remains unguarded. Make the standing lead genuinely state-neutral or assert a failure-compatible heading for this case. [`packages/web/src/pages/opentag/copy.ts:39`]

✅ 4. Bot failure and a missing Computer now offer an immediate exit, and the explicit Task retry/create-or-reuse behavior remains intact after the server helper relocation.

**Action taken**
- Submitted request changes on `84ee4189`.

Gandy2025:
@baixiaohang @yuezengwu — the product owner has asked me to make the responsibility and follow-up boundary explicit after the deletion-first convergence.

## What #2321 owns

This PR is a thin integration over the existing Feishu setup Task. It owns exactly four invariants:

1. OpenTag calls the existing setup-chat only after the member has chosen Feishu (represented durably by an existing Bot binding).
2. The existing keyed `createChat` seam converges ensure/load/reload/tab calls on one private Task per human + Agent + Client. An explicit tools retry adds one ask/wake in that same Task; ensure-shaped calls stay inert.
3. Step 3 renders only the existing `Feishu Bot` and `Agent tools` facts. Real first use and onboarding completion remain closed until the exact Bot is reachable and the current Computer's CLI is ready.
4. Recovery stays minimal: a tools request failure or long wait may offer the tools retry; any established failure/no-Computer/long wait may offer `Finish later`, which leaves onboarding incomplete and returns to Agent Detail.

The current three findings on `84ee4189f` are within this scope and should be fixed locally, without restoring the deleted state/copy matrix:

- make the single standing heading genuinely state-neutral;
- let a rejected setup request prevent the tools row from claiming background work;
- scope the tools retry to `setupFailed || stepSlow`, so a Bot-only failure cannot re-kick ordinary tools work.

## Review responsibilities

- **@baixiaohang — Standards axis:** keyed-chat concurrency and namespace safety; authorization/ownership; message creation and wake behavior; reuse of existing repository seams; no accidental persistence, DTO, or cross-onboarding expansion.
- **@yuezengwu — Spec axis:** trigger only after Feishu choice; two existing facts only; Bot-reachable + CLI-ready gate; exact retry/Finish-later behavior; Agent Detail as the sole permanent manual repair entry; no user-visible install/setup concept.

Both reviewers should block only on a demonstrated violation of those invariants, repository Standards, or a correctness/security regression. Please decline copy-combination modeling, generic abstractions, QA-case maintenance, cleanup, and other improvements that are not needed for those invariants.

## Related issue plan

#2323 is the one approved follow-up: make OpenTag Feishu onboarding completion server-authoritative at the durable write seam. It is assigned to @baixiaohang and will be designed/reviewed as a separate change because it requires explicit scope over the shared completion contract. The product owner accepted the narrow client-check/server-stamp race for #2321, so #2323 does not block this PR.

#2321 must not implement, stub, add an extension point for, or leave a TODO for #2323. No other follow-up should be created from this review unless it is independently severe and cannot be safely decoupled from these four invariants.

Gandy2025:
@baixiaohang @yuezengwu — clarification from the product owner: my preceding comment overcomplicated the instruction by framing reviewer roles.

The scope principle for #2321 is simply:

> Keep this PR restrained. It should be the thinnest possible call into the existing Feishu setup chat, plus the minimum wiring needed to observe the existing Bot/CLI facts and gate first use.

Please review with that as the ceiling. Fix only demonstrated correctness issues in that thin path. Do not use this PR to introduce a new setup abstraction, readiness/state model, copy matrix, endpoint, DTO, persistence, generic completion design, QA library work, cleanup, or adjacent hardening.

The existing setup-chat implementation, keyed `createChat` seam, binding polling, `binding.cli`, and Agent Detail repair action should be reused directly. If a real concern is separable from this thin call path, report it separately instead of expanding #2321. The already-identified server-authoritative completion concern is #2323 and is explicitly outside this PR.

The three current findings are accepted only as minimal local correctness fixes to the thin layer: neutralize one standing heading, stop a failed request from claiming work is active, and prevent a Bot-only failure from retrying the tools Task. Nothing broader is authorized.

Gandy2025:
All three are fixed in `4a99014c1`, locally, with no matrix restored.

**The standing heading directed an action.** It said "Confirm the Bot in Feishu", which asks for a confirmation already given once the Bot is reachable, and one the member cannot give when the connection has failed. It describes now and directs nothing: "This needs two things: a Feishu Bot, and the Feishu tools on your Agent's Computer." The failed-connection regression asserts that heading is not asking for confirmation there.

**A rejected request read as background preparation.** `toolsReadiness` saw only `binding.cli.state`, which cannot tell "being prepared" from "never asked for" — both look absent. The request outcome reaches that row now, so it reads "The automatic setup didn't start." The regression asserts the absence of the preparing copy, not just the presence of the controls.

**The tools retry inherited the combined recovery.** A Bot failure alone put `Try again` on screen, and pressing it appended work to a Task the Agent was already getting on with. It is `hasComputer && !toolsReady && (setupFailed || stepSlow)` now — mutation-checked: restoring the old expression fails the new test.

That is one boolean threaded into a row, one term narrowed, and one sentence rewritten. No state model, no copy derivation, no new abstraction.

On the framing: this PR is meant to be the thinnest layer over the setup Task that already exists — a trigger, idempotency on the existing keyed seam, a retry that actually reaches the Agent, the two-fact gate, and a way out. Anything that needs more than that to fix is better off outside it; #2323 already has the completion contract. If a finding can only be resolved by adding an abstraction, a persistence concept, or a second surface, please raise it as a separate issue rather than as a change here, and I will treat it that way.

Both suites, `pnpm check` and the token gate are green locally; CI and independent QA are re-bound to this head.

yuezengwu (approved):
The three accepted thin-layer findings are fixed on 4a99014c1 without restoring the removed state/copy matrix. The standing Step 3 heading is now state-neutral; a rejected tools request reaches the readiness row and no longer claims background work; and tools retry is scoped to setup failure or a long tools wait, so a Bot-only failure cannot append another Task ask. The added DOM regressions cover each path. No further blocking findings in this incremental review.

baixiaohang (commented):
The three thin-layer blockers on `84ee4189` are fixed in `4a99014c` without restoring the removed readiness/copy model:

- the standing Step 3 heading is now descriptive across both halves;
- the identity-scoped request failure reaches the tools row, which no longer claims background work after the POST is rejected;
- tools retry now requires its own failure/slow reason, so a Bot-only failure cannot append an unrelated ask.

The corresponding DOM regressions cover each transition. I also rechecked the current Standards-sensitive path: the fixed-suffix kickoff namespace, human/Agent/Client isolation, keyed `createChat` row-lock reuse, ensure-vs-retry message/wake behavior, notification message selection, and existing manage authorization remain consistent. No further blocking findings on this head.

The server-authoritative completion race is durably captured by #2323 and requires the separately scoped shared completion contract change described there.

Gandy2025:
@baixiaohang — your review on `4a99014c1` came through as a comment, so the earlier `CHANGES_REQUESTED` on `84ee4189f` is still the standing verdict and the PR reads as blocked. Since you've confirmed no further blocking findings on this head, could you submit it as an approval to clear it?

Nothing has changed since: `4a99014c1` is the head both of you reviewed, and I'm not touching it further short of a demonstrated correctness or security blocker.

PR #2324, opened 2026-08-13
feat(web): add public OpenTag entry

Gandy2025:
## Summary

- add one public OpenTag CTA to the source-owned unauthenticated landing hero
- keep the existing general `Get Started` sign-in destination unchanged
- explain the OpenTag Feishu/Task value in one factual sentence
- cover both landing destinations with a focused DOM regression test

The CTA uses the existing `/opentag` route. The current auth guard, login `next` handoff, callback navigation, and strict solo-signup preservation continue to return a new OAuth user to that exact destination.

## Verification

- `pnpm --filter @first-tree/web test` (253 files, 2397 tests)
- `pnpm --filter @first-tree/web typecheck`
- `pnpm --filter @first-tree/server exec vitest run src/__tests__/oauth-bootstrap.test.ts` (13 tests)
- `pnpm check` (exit 0; existing repository warnings only)
- local Playwright check: unauthenticated `/` showed both CTAs; selecting OpenTag navigated to `/login`

## Scope

No navigation redesign, signup protocol, campaign, analytics, rollout, onboarding, or broader rebrand changes.

PR #2325, opened 2026-08-13
revert(web): remove public OpenTag entry

Gandy2025:
## Summary

- remove the public OpenTag CTA and explanatory copy added to the First Tree landing hero by #2324
- restore the original single `Get Started` action and its `/login` destination and visual treatment
- remove the focused DOM test that existed only for the reverted OpenTag entry

This is a product-scope correction. A separate OpenTag website and its entry are owned by other work; this PR does not change the existing `/opentag` flow, auth/OAuth, signup, navigation, or onboarding.

## Verification

- `pnpm --filter @first-tree/web exec vitest run src/auth/__tests__/require-auth-app.test.tsx` (4 tests)
- `pnpm --filter @first-tree/web typecheck` (includes design-token guardrails)
- `pnpm exec biome check packages/web/src/pages/landing/hero.tsx`
- `git diff --check`
- the two affected paths exactly match the parent of squash commit `5328bbf5430a86d7b5a2a509b8f3ff7e9683acf6`

## Scope

Only the landing hero and the focused test introduced by #2324 are reverted.


Gandy2025:
Closing because the branch prefix does not satisfy the repository CI naming contract. The unchanged exact commit will be republished from a compliant `fix/` branch in a replacement PR.

PR #2333, opened 2026-08-14
fix(cli): keep macOS daemon online across updates

Gandy2025:
## Summary

- keep the live launchd wrapper running across the reserved self-update exit code 75 so it can start the newly installed CLI without waiting for launchd to respawn the job
- atomically replace the generated wrapper and plist during refresh-unit handoff
- parse launchctl state exactly so state = not running is no longer reported as active
- add shell-level regression coverage and a reusable macOS launchd handoff QA case

## Root cause

A managed macOS client successfully installed its target version and exited with status 75. launchd recorded the failed-exit KeepAlive demand, but the GUI domain deferred the pending non-demand spawn, leaving the computer disconnected. The status parser compounded diagnosis by matching the running substring inside state = not running.

## Verification

- pnpm exec vitest run src/__tests__/render-unit-channel.test.ts src/__tests__/service-install-core.test.ts src/__tests__/update-glue.test.ts src/__tests__/daemon-refresh-unit.test.ts — 80 tests passed
- pnpm --filter first-tree-dev test — all 26 CLI batches passed
- pnpm typecheck — 9 tasks passed
- Biome check on changed TypeScript files passed
- git diff --check passed

## Rollout / QA

The committed macOS QA case requires two real channel-correct artifacts and a disposable launchd boundary. That live handoff has not been run yet. Roll this out to staging first, install the first fixed staging build manually so the new wrapper is present, and observe the next automatic staging update before promoting to production.


yuezengwu (approved):
No blocking findings at `d670638d555141c931bfa8d39b90ddb0b9db5710`.

The launchd wrapper now keeps the job alive across the reserved update exit code while forwarding every other status back to launchd, and the atomic replacements avoid truncating the active script during refresh. Exact state parsing fixes the `not running` false positive. I also checked the parent-shell/process-group boundary against launchd's default cleanup behavior; normal bootout/restart semantics remain intact.

No database or core data-structure changes are involved. The real macOS handoff QA remains intentionally unrun, so keep the documented staging-first rollout and observe an automatic staging update before production promotion.

baixiaohang (changes requested):
- Rationale: The live launchd wrapper never reloads the replacement written by `refresh-unit`, so command/path migrations can still fall back to the deferred launchd respawn this PR is meant to avoid; exact-head live QA is also still pending.

**Risk level**: A
- Semantic lift: touches the daemon self-update and launchd handoff protocol; level remains A

**PR summary**
- Problem: Managed macOS clients can stay disconnected after a successful self-update when the GUI launchd domain defers the requested KeepAlive respawn, and diagnostics can misreport `not running` as active.
- Approach: Keep a launchd-owned shell alive across exit 75, replace the generated wrapper/plist atomically, parse launchctl state exactly, and add deterministic plus live-handoff coverage.
- Impacted modules: CLI launchd supervisor, service-install tests, local-dev documentation, and runtime QA cases

**Review findings**
❌ 1. On exit 75 this continues the already-running shell loop, so it keeps executing the command embedded in the old wrapper. `daemon refresh-unit` atomically renames a new wrapper onto the path, but it cannot change the script or command already loaded by this shell. If an update changes the resolved CLI path or command shape—the compatibility case `refresh-unit` exists to handle—the old command can fail under the new CLI, the wrapper then exits non-75, and recovery again depends on the launchd respawn that may be deferred. Re-exec the wrapper path (or otherwise reload the replacement) on exit 75, and extend the shell regression to replace the wrapper with a different command between the two child launches. [R5 / `apps/cli/src/core/supervisor/launchd.ts:147`]
❌ 2. This exact head is Ready even though the PR says the real macOS handoff has not run. The cross-process path is not proven by the passing unit/CI checks; keep the PR Draft until independent QA runs the committed case against the final exact head and returns PASS. [`packages/qa/cases/runtime/macos-launchd-update-handoff.md:16`]
✅ 3. Exact `launchctl` state parsing correctly removes the `not running` substring false positive, and atomic rename avoids in-place truncation of the active wrapper.

**Action taken**
- Submitted request changes

Gandy2025:
Addressed the command/path migration finding in 39af97f60.

The launchd wrapper now execs its own stable path after exit 75, which reloads the atomically replaced script while preserving the launchd-owned PID. The shell regression now replaces the live wrapper with one that embeds a different daemon command between launches. It failed against the prior implementation with exit 9 (old command invoked twice) and passes after the fix with old then new.

Focused result: 80/80 tests passed; typecheck 9/9; changed-file Biome check passed. I have kept the PR in Draft as requested while exact-head live macOS QA remains pending.

baixiaohang (commented):
**Recommendation**: comment
- Rationale: The command/path migration blocker is resolved at `39af97f607ba1c71dc44f8b3f7c6f69bf9c0fa4a`; approval remains gated on independent exact-head macOS handoff QA.

**Risk level**: A
- Semantic lift: touches the daemon self-update and launchd handoff protocol; level remains A

**PR summary**
- Problem: Managed macOS clients can stay disconnected after a successful self-update when launchd defers its KeepAlive respawn, while diagnostics can also misreport `not running` as active.
- Approach: Keep the launchd-owned wrapper process alive across exit 75, atomically replace and then re-exec the wrapper so updated command/path templates take effect, parse launchctl state exactly, and cover deterministic plus live handoff behavior.
- Impacted modules: CLI launchd supervisor, service-install tests, local-dev documentation, and runtime QA cases

**Review findings**
✅ 1. The prior command/path migration blocker is resolved. After exit 75, `exec "$0"` preserves the launchd-owned PID while reopening the stable wrapper path, so the atomically replaced script controls the next child launch. The regression now replaces the live wrapper with a different embedded daemon command and verifies the old-to-new transition. [`apps/cli/src/core/supervisor/launchd.ts:154`]
⚠️ 2. The PR is correctly back in Draft, but the exact head still has no independent live macOS handoff PASS. Keep it Draft and withhold approval until the committed case completes against the final head; unit and CI evidence cannot establish the on-demand-only launchd boundary. [`packages/qa/cases/runtime/macos-launchd-update-handoff.md:16`]
✅ 3. The exact launchctl-state parsing and atomic file replacement remain sound in the updated diff.

**Action taken**
- Submitted comment; approval remains pending exact-head live QA

Gandy2025:
@baixiaohang The command/path migration finding is fixed at 39af97f60 and all checks are green. The remaining committed live case requires a disposable macOS GUI launchd session, two channel-correct artifacts, and a test server/client registration; this PR runner does not provide that native boundary, and using the operator staging service would violate the case precondition. Could you either coordinate the independent exact-head QA run or confirm whether the documented staging-first rollout is acceptable for clearing the review? The PR remains Draft meanwhile.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The command/path migration blocker is resolved at `39af97f607ba1c71dc44f8b3f7c6f69bf9c0fa4a`, the updated static review found no remaining code blocker, and the responsible human explicitly authorized approval with live validation deferred to the documented staging-first rollout.

**Risk level**: A
- Semantic lift: touches the daemon self-update and launchd handoff protocol; level remains A

**PR summary**
- Problem: Managed macOS clients can stay disconnected after a successful self-update when launchd defers its KeepAlive respawn, while diagnostics can also misreport `not running` as active.
- Approach: Keep the launchd-owned wrapper process alive across exit 75, atomically replace and re-exec the wrapper so updated command/path templates take effect, parse launchctl state exactly, and add deterministic plus live handoff coverage.
- Impacted modules: CLI launchd supervisor, service-install tests, local-dev documentation, and runtime QA cases

**Review findings**
✅ 1. Exit 75 now re-execs the stable wrapper path, preserving the launchd-owned PID while loading the atomically replaced script and its updated command/path.
✅ 2. The regression replaces the live wrapper between child launches and verifies the old-to-new command transition, directly covering the previously blocking compatibility case.
✅ 3. Exact launchctl-state parsing and atomic wrapper/plist replacement remain sound.
⚠️ 4. The native macOS GUI launchd handoff has not been independently exercised at this exact head. Approval is explicitly authorized with that known gap; keep the rollout staging-first and require observed automatic staging-update recovery before production promotion.

**Action taken**
- Approved with explicit human authorization; not merged

PR #2334, opened 2026-08-14
Fix authoritative OpenTag Feishu onboarding completion

baixiaohang:
## Summary

- add an Agent-scoped Feishu onboarding completion endpoint that revalidates manager ownership, the live Bot lease, and the Agent's current Computer `lark-cli` readiness in one short transaction
- centralize membership completion stamping and migrate the existing generic onboarding and kickoff paths without changing their wire behavior
- complete OpenTag at the usable Feishu handoff — reachable owned Bot plus ready current-Computer CLI — without waiting for a first message, and project success only to the membership that started the request
- add deterministic regressions for ownership, Bot, Client, CLI, membership projection, Client retirement, capability loss, and read-only teammate handoff paths

## Lock ordering and linearization

The current Client is intentionally read without a row lock. This is the minimal lock-order correction approved during review: `retireClient` locks Client before Agent, while completion keeps membership, Agent, and Bot binding locked through the stamp.

CLI readiness is fenced again inside the conditional membership `UPDATE`: an `EXISTS` predicate checks the locked Agent's current Client, `retired_at IS NULL`, and `metadata.capabilities.lark-cli.available = true` in the same PostgreSQL statement snapshot that writes all three completion fields. This closes capability-drop races without recreating the Agent-to-Client / Client-to-Agent deadlock.

## Merge ordering

The ready-handoff trigger follows the reviewed replacement contract in first-tree-context#945. This PR must remain unmerged until that Context Tree PR is reconciled, receives final-head approval, and merges.

## Validation

Passed:

- `pnpm --filter @first-tree/server test --maxWorkers=2`
- `pnpm --filter @first-tree/web test --maxWorkers=2`
- `pnpm typecheck`
- `pnpm test --concurrency=1 -- --maxWorkers=2`
- targeted server Feishu completion suite: 20/20
- final-head OpenTag flow and DOM suites: 83/83
- Web full suite after the ready-handoff change: 253 files / 2426 tests
- Web typecheck and design-token guard
- changed-file Biome checks and `git diff --check`

Known repository baseline:

- `pnpm check` still reports seven pre-existing Biome errors in untouched `origin/main` files; all files changed by this PR pass Biome.
- the default parallel root `pnpm test -- --maxWorkers=2` hit the existing Turbo race where build and test concurrently recreate generated client/skills directories (`ENOTEMPTY`); the serialized full run above passed.

[...]

PR #2337, opened 2026-08-14
Fix attachment retention test lifecycle race

baixiaohang:
## Summary

- disable automatic attachment-retention scheduling in generic Server test apps while keeping production and explicit opt-in behavior enabled
- track active retention sweeps and wait for them during background-task shutdown before the database pool closes
- add deterministic regressions for the test-helper wiring and the in-flight shutdown boundary

## Root cause

Main [CI run 31784141632](https://github.com/agent-team-foundation/first-tree/actions/runs/31784141632) failed when the per-test `TRUNCATE ... CASCADE` deadlocked with the startup attachment-retention query. The retention sweep was launched fire-and-forget from `backgroundTasks.start()`, so a test app could return from `ready()` while the sweep was still using the same worker database.

The failed `me-multi-org` test was only the cleanup victim. PR #2334 did not change the test harness, background task, retention service, or failed test.

## Validation

Passed locally:

- related Server suites: 4 files / 49 tests with `--maxWorkers=2`
- background-task and bare-test-app final suites: 2 files / 7 tests
- full Server suite: 293 files / 3441 tests with `--maxWorkers=2`
- changed-file Biome checks
- `git diff --check`

Mutation checks also proved both regressions are effective:

- removing the in-flight `stop()` wait changes the settlement race from `sweep-pending` to `stopped`
- removing the `createTestApp()` default-disable wiring starts the mocked retention sweep and fails the helper-level test

### DIFF
--- a/packages/server/src/__tests__/background-tasks-extra.test.ts
+++ b/packages/server/src/__tests__/background-tasks-extra.test.ts
@@ -212,4 +212,43 @@ describe("createBackgroundTasks", () => {
     await tasks.stop();
     expect(vi.getTimerCount()).toBe(0);
   });
+
+  it("allows callers to disable automatic attachment-retention sweeps", async () => {
+    const { createBackgroundTasks } = await import("../services/background-tasks.js");
+    const tasks = createBackgroundTasks(makeApp(), "srv_5", { attachmentRetentionSweepEnabled: false });
+
+    tasks.start();
+    await vi.advanceTimersByTimeAsync(24 * 60 * 60 * 1_000);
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed the lifecycle fix. createTestApp() now disables the retention sweep by default while production remains enabled unless explicitly opted out; active retention sweeps are tracked and awaited before the database pool closes. The helper-level and in-flight shutdown regressions cover both failure boundaries. No schema or core data-structure changes. No blocking findings.

PR #2350, opened 2026-08-18
feat(client): add Amp as a first-class runtime provider

L42y:
## Summary
- Adds wire id `amp` and wires Amp through the existing Cursor/OpenCode handler contract: spawn `amp --execute --stream-json --stream-json-thinking --no-remote-control-terminal`, new threads with `--visibility private`, resume with `threads continue <T-uuid>`, MCP via mode-0600 `--settings-file` (`amp.mcpServers`), permissions via the same runtime-owned settings file (`amp.dangerouslyAllowAll`).
- Auth stays host-local (`amp login` / `AMP_API_KEY`). First Tree never holds Amp tokens. Windows V1 fails closed at capability probe (before install detection) until the Job Object supervisor exists.
- Config `model` is an Amp mode (`"" | low | medium | high | ultra`); empty omits `--mode`, non-empty maps to `--mode`. Prompt rides stdin, never argv. MCP headers never ride argv.

## Test plan
- [ ] Create an Amp agent on a machine with `amp` installed and `amp login` completed; send a first turn and confirm a real `T-…` session id is persisted and argv includes `--visibility private` and `--no-remote-control-terminal`.
- [ ] Resume the same chat and confirm argv is `amp threads continue <T-uuid> --execute … --no-remote-control-terminal` (never a synthetic `amp-pending-*` id; no `--visibility` on continue).
- [ ] Confirm a logged-out Amp CLI surfaces an in-chat hint to run `amp login`, not a First Tree Connect button.
- [ ] Confirm Windows capability stays unavailable with the Job Object message even when Amp is not installed (no `install.sh` invite).
- [ ] Unit: Amp parser/binary/capability/handler tests plus catalog/capability exhaustive maps.


baixiaohang (changes requested):
- Rationale: The provider boundary currently has two settlement/replay correctness bugs and exposes two unsupported or unsafe CLI contracts.

**Risk level**: A
- Semantic lift: none beyond A; the diff directly owns runtime process, session, and delivery-settlement behavior.

**PR summary**
- Author / repo: L42y / first-tree
- Problem: Operators cannot currently choose Amp as an agent runtime while keeping Amp authentication on the host and retaining First Tree's managed-session workflow.
- Approach: Adds Amp to the shared catalog and web surfaces, then runs one Amp CLI process per turn with JSONL parsing, provider-native thread resume, projected MCP configuration, and runtime-owned settings.
- Impacted modules: client provider/runtime handling, CLI capability refresh, shared runtime schemas/catalog, and web provider configuration.

**Review findings**
❌ 1. A successful turn is rejected whenever the echoed prompt or successful answer merely contains an auth phrase. `failureText` includes the raw stdout JSONL tail, including `user` and `assistant`/`result` content, and `authFailure` participates in `success` even when exit code is 0 and the terminal result is successful. For example, asking Amp to explain an “invalid api key” error echoes that phrase and turns the completed answer into a consumed error. Restrict auth classification to stderr/non-zero exits or structured error events. [`packages/client/src/providers/amp/index.ts:786`](https://github.com/first-tree-ai/first-tree/blob/2f420d6afdad749e4c4797655665d2318fda527d/packages/client/src/providers/amp/index.ts#L786-L798) [R4]

❌ 2. The parser collapses each assistant content array to one event: it keeps only the last `tool_use`, drops text/thinking when any tool is present, and returns only the first `tool_result`. Amp's official stream schema permits arrays of content blocks. If an unsafe write and a read are emitted together and the read is retained, `sawUnsafeTool` remains false; a later failure can therefore replay a turn that already mutated the workspace. Emit every block in order and derive replay safety from all tool calls. [`packages/client/src/providers/amp/parser.ts:138`](https://github.com/first-tree-ai/first-tree/blob/2f420d6afdad749e4c4797655665d2318fda527d/packages/client/src/providers/amp/parser.ts#L138-L167) [R4]

❌ 3. The product exposes a provider-native model ID and forwards it as `--model`, but Amp's current official contract defines modes as capability presets rather than fixed model selectors; its SDK/CLI option contract exposes `mode`/`effort`, not `model`. There is also no version/capability check proving that a resolved binary accepts this flag. Any non-empty model configured through the new UI is therefore an unsupported spawn contract. Remove the model-id surface or map a validated Amp mode/capability instead. [Amp modes](https://ampcode.com/manual) · [Amp SDK options](https://ampcode.com/manual/sdk/typescript) [`packages/client/src/providers/amp/index.ts:91`](https://github.com/first-tree-ai/first-tree/blob/2f420d6afdad749e4c4797655665d2318fda527d/packages/client/src/providers/amp/index.ts#L91-L94) [R5]

❌ 4. `--mcp-config` serializes configured HTTP headers directly into argv. Those headers commonly carry bearer tokens, making them observable through the host process table and any spawn diagnostics that record arguments. The prompt is deliberately kept off argv; MCP credentials need the same treatment, e.g. a mode-0600 runtime-owned config or another non-argv channel. [`packages/client/src/providers/amp/index.ts:92`](https://github.com/first-tree-ai/first-tree/blob/2f420d6afdad749e4c4797655665d2318fda527d/packages/client/src/providers/amp/index.ts#L92-L94) [R5]

**Action taken**
- Submitted request changes.

L42y:
Addressed the request-changes review on this PR:

1. **Auth false positive** — Auth classification now uses stderr always, structured stream errors (`state.errors` / `result.isError`), and stdout only when the process exit code is non-zero. Successful turns that merely mention phrases like `invalid api key` stay successful.
2. **Parallel tool blocks** — The Amp stream parser now emits every assistant/user content block in order (text, thinking, each `tool_use`, each `tool_result`). Replay safety therefore sees every tool call, including a write that coexists with a later read in the same message.
3. **`--model` contract** — The Amp config `model` field is Zod-narrowed to Amp modes (`"" | low | medium | high | ultra`). Spawn uses `--mode` (empty omits the flag). Leftover model ids fail closed at parse and before spawn. The web picker is a curated mode Select.
4. **MCP headers on argv** — Caller MCP (including HTTP headers) is folded into the mode-0600 runtime `--settings-file` under `amp.mcpServers`. `--mcp-config` is no longer passed on argv.

Pushed as `e7a607fb0` on `jeremy/amp-runtime-provider`.

baixiaohang (changes requested):
- Rationale: The four original findings are resolved, but the current Amp spawn/capability boundary still exposes an unintended sharing/control path and advertises an unusable Windows setup path.

**Risk level**: A
- Semantic lift: none beyond A; these findings affect runtime privacy and platform admission.

**PR summary**
- Author / repo: L42y / first-tree
- Problem: Operators need to run First Tree agents through their host-authenticated Amp installation with managed sessions and tools.
- Approach: The follow-up correctly changes model ids to validated Amp modes, moves MCP secrets off argv, parses all stream blocks, and narrows auth classification.
- Impacted modules: Amp handler/parser/capability, shared runtime config, and web runtime settings.

**Review findings**
❌ 1. New First Tree chats currently inherit Amp's external sharing and remote-control defaults. Amp documents `visibility: "workspace"` as the execution default, so a workspace user's thread — including First Tree messages, tool calls, and context — is shared with their Amp workspace unless explicitly made private. The handler also inherits the complete host environment, while Amp documents that `AMP_REMOTE_CONTROL_TERMINAL=1` enables web terminal control unless an explicit CLI flag overrides it. That creates a second control plane outside First Tree's participant/session boundary. Please force private visibility for newly created threads and `--no-remote-control-terminal` for every turn, or introduce an explicit First Tree-owned opt-in policy before allowing either behavior. [Amp thread sharing](https://ampcode.com/manual) · [Amp CLI option defaults](https://ampcode.com/manual/sdk/typescript) [`packages/client/src/providers/amp/index.ts:84`](https://github.com/first-tree-ai/first-tree/blob/e7a607fb0c877d0e30b0c6deadae1fdf2b182e02/packages/client/src/providers/amp/index.ts#L84-L94) [R5]

❌ 2. The Windows admission gate only replaces a successful binary probe. On a Windows host without Amp, `detected.state` is `missing`, so line 31 returns the normal missing-runtime result with the macOS/Linux `curl ... | bash` installer instead of the promised Job Object unsupported message. The UI therefore invites users to install a runtime that `prepareSession` will always reject. Apply the platform gate for every Windows probe state (ideally before install detection) and cover the missing-binary case. [`packages/client/src/providers/amp/capability.ts:20`](https://github.com/first-tree-ai/first-tree/blob/e7a607fb0c877d0e30b0c6deadae1fdf2b182e02/packages/client/src/providers/amp/capability.ts#L20-L41) [R4]

⚠️ 3. Amp's official stream examples put usage on each assistant message and omit it from the terminal result, while this parser only reads `record.usage` on `type: "result"`; the handler therefore emits no `token_usage` for the documented common shape. Please retain/aggregate `message.usage` as a fallback (the current fixture fabricates result-level usage and misses this). [Amp stream schema and examples](https://ampcode.com/manual/appendix) [`packages/client/src/providers/amp/parser.ts:138`](https://github.com/first-tree-ai/first-tree/blob/e7a607fb0c877d0e30b0c6deadae1fdf2b182e02/packages/client/src/providers/amp/parser.ts#L138-L184) [R5]

✅ 4. The previous auth false-positive, multi-block replay-safety, fixed-model, and MCP-header-in-argv findings are closed in `e7a607fb0`.

**Action taken**
- Submitted request changes on the current head.

L42y:
Addressed the second request-changes review:

1. **Thread privacy / remote control** — New Amp turns pass `--visibility private`. Every turn passes `--no-remote-control-terminal`, and spawn env forces `AMP_REMOTE_CONTROL_TERMINAL=0` so a host-inherited enablement cannot open a second control plane.
2. **Windows admission** — Capability now fails closed on `win32` *before* install detection (state `error`, Job Object copy), including the missing-binary case — no `install.sh` invite on Windows.
3. **Usage** — Parser retains/aggregates `message.usage` from assistant events; handler emits `token_usage` from that fallback when the terminal result omits usage.

Pushed as `ae8937b49` on `jeremy/amp-runtime-provider`. Also refreshed the PR summary to match the current spawn contract.

yuezengwu (approved):
Re-reviewed the current head `ae8937b49520b492b99e9f4df8314212b0671e72`.

The follow-up closes the remaining blocking findings: new threads are private, remote terminal control is disabled at both the CLI and environment boundaries, Windows fails closed before install detection, and assistant-message usage is retained as the documented fallback. The earlier auth classification, ordered multi-block parsing/replay safety, Amp mode mapping, and MCP secret handling fixes also remain intact.

No additional blocking findings.

packages/client/src/providers/amp/index.ts
```diff
@@ -0,0 +1,1228 @@
...
+      if (isAmpPendingSessionId(sessionId)) {
+        pendingSyntheticId = sessionId;
+        providerSessionId = null;
+      } else {
+        providerSessionId = sessionId;
+        pendingSyntheticId = null;
+      }
+      const fingerprint = computeBriefingFingerprint(briefing);
+      if (message) {
+        let prompt = await sessionCtx.formatInboundContent(message);
```
yuezengwu, line 1133: Non-blocking follow-up: `formatInboundContent()` is still outside the `try/finally` that clears `initialTurnPreparing`. If formatting an attachment or inbound content throws, the flag remains true, so `inject()` continues queueing and the drain is not scheduled until outer session recovery/teardown. Please move formatting (and the fingerprint notice construction) inside the cleanup `try/finally`, matching `start()`, and cover the formatter-failure path.

L42y:
This PR closed when the head branch was renamed to `feat/amp-runtime-provider` (CI `Check Branch Name` requires `feat/` / `fix/` / …). Continuation is https://github.com/first-tree-ai/first-tree/pull/2351 on the same commits (including the review fixes through `ae8937b49` and a biome format fix).

packages/client/src/providers/amp/index.ts
L42y: Addressed on the active PR head (`feat/amp-runtime-provider` / #2351) in `54879f08c`.

`resume` now formats inbound content (and builds the briefing fingerprint notice) inside the same `try/finally` that clears `initialTurnPreparing`, matching `start()`: on formatter failure the flag clears and later `inject()` can schedule drain. Covered with a handler test that throws from `formatInboundContent` on resume, then asserts a subsequent inject drains.

PR #2353, opened 2026-08-19
fix(web): keep mobile Chat list position when returning from detail

L42y:
## Summary

- Mobile Chat already uses `/m/chat` for the list and `?c=` for conversation detail. Opening a chat previously unmounted the list, so going back jumped to the top.
- This parks the list under the detail view (hidden, inert, still mounted) so back keeps the same scroll offset, rendered window, search, and quick-view chrome.

## Test plan

- [ ] On a phone-width `/m/chat` list, scroll down, open a conversation, tap back — the same chats should still be on screen
- [ ] Quick view (Unread / Pinned) should still be selected after returning
- [ ] Bottom tabs stay hidden while in chat detail and return on the list
- [x] `packages/web` vitest: `chat-nav`, `card-actions-dom`, `shell-dom`, `density-dom` (25 tests)

## Validation

- [ ] `pnpm check`
- [ ] `pnpm typecheck`
- [x] focused `packages/web` vitest for the mobile Chat list/detail tests

## Change Surface

## Notes

- docs or tests updated to match: added a DOM test that the list scroller keeps `scrollTop` across open/back
- follow-up work: none

### DIFF
--- a/packages/client/src/providers/claude/__tests__/provider-error-result.test.ts
+++ b/packages/client/src/providers/claude/__tests__/provider-error-result.test.ts
@@ -130,7 +130,12 @@ function firstProviderPayload(payloads: readonly ProviderRetryEventPayload[]): P
 }
 
 async function waitForCondition(predicate: () => boolean, description: string): Promise<void> {
-  for (let attempt = 0; attempt < 100; attempt += 1) {
+  // Wall-clock bound: under CI load the retry's post-timer Promise work can
+  // take more than a fixed handful of setImmediate ticks. Fake timers still
+  // own setTimeout; Date.now stays real because this file only fakes
+  // setTimeout/clearTimeout.
+  const deadline = Date.now() + 5_000;
+  while (Date.now() < deadline) {
     if (predicate()) return;
     await new Promise((resolve) => setImmediate(resolve));
   }
@@ -326,6 +331,7 @@ describe("claude-code handler — structured provider error result", () => {
     try {
       const result = await startSingleResultTurn();
       await waitForCondition(() => providerRetryPayloads(result.emitted).length === 1, "first scheduled retry");
+      await waitForCondition(() => vi.getTimerCount() > 0, "retry backoff timer");
 
       expect(mockState.queryCalls).toBe(1);
       await vi.advanceTimersByTimeAsync(499);
--- a/packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx
+++ b/packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx
@@ -6,12 +6,24 @@ import { act } from "react";
 import { MemoryRouter, Route, Routes, useNavigationType } from "react-router";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createDomHarness, type DomHarness } from "../../../test-utils/dom-harness.js";
+import { MobileShell } from "../shell.js";
 import { MobileWorkPage } from "../work.js";
 
-const authMock = vi.hoisted(() => ({ value: { agentId: "human-agent-self" } }));
+const authMock = vi.hoisted(() => ({
+  value: {
+    agentId: "human-agent-self",
+    organizationId: "org-1",
+    meLoaded: true,
+    onboardingStep: "completed" as const,
+    onboardingDismissedAt: null,
+    onboardingCompletedAt: "2026-07-01T00:00:00.000Z",
+    currentOrgHasPersonalAgent: true,
+  },
+}));
 const meChatMocks = vi.hoisted(() => ({
   listMeChats: vi.fn(),
   listMeChatSourceCounts: vi.fn(),
+  listNeedYouRequests: vi.fn(),
 }));
 
 function chatRow(overrides: Partial<MeChatRow> = {}): MeChatRow {
@@ -44,6 +56,8 @@ function chatRow(overrides: Partial<MeChatRow> = {}): MeChatRow {
 
 vi.mock("../../../auth/auth-context.js", () => ({ useAuth: () => authMock.value }));
 vi.mock("../../../api/me-chats.js", () => meChatMocks);
+vi.mock("../../../hooks/use-admin-ws.js", () => ({ useAdminWs: () => undefined }));
+vi.mock("../../../components/team-switch-overlay.js", () => ({ TeamSwitchOverlay: () => null }));
 // Stub the heavy chat-detail so we can drive the back affordance directly.
 vi.mock("../../workspace/center/index.js", () => ({
   CenterPanel: ({ onShowConversations }: { onShowConversations: (() => void) | null }) => (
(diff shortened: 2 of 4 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change preserves one mobile Chat-list instance across list/detail navigation while making the parked pane non-interactive and hidden from assistive technology.

**Risk level**: C
- Path baseline: `packages/web/**` only -> C.
- Semantic lift: none; this changes component mounting and layout without altering WebSocket parsing, query contracts, or shared data semantics.

**PR summary**
- Problem: On phones, returning from a conversation remounted the Chat list and lost the user’s scroll position and local list chrome.
- Approach: Keep the list mounted beneath detail, park it with `inert`/`aria-hidden`/visibility controls, and restore the same list DOM when the detail query parameter is cleared.
- Impacted modules: mobile Chat navigation and its DOM regression tests.

**Review findings**
✅ 1. The implementation keeps a single list state owner and DOM subtree instead of introducing a second restoration cache, so scroll, rendered batches, search, and filters remain coherent. [R1/R3 / `packages/web/src/pages/mobile/work.tsx:86`]
✅ 2. The parked pane is isolated from pointer, focus, and accessibility interaction while the detail remains the only visible surface. [R4 / `packages/web/src/pages/mobile/work.tsx:90`]
✅ 3. The regression test verifies both element identity and `scrollTop` across open/back, covering the failure mode directly. [`packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx:157`]

**Action taken**
- Approved.

yuezengwu (changes requested):
Blocker: the parked list still resizes when mobile Chat enters detail mode.

`packages/web/src/pages/mobile/work.tsx:92` keeps the parked pane at `h-full`, while `packages/web/src/pages/mobile/shell.tsx:78` removes the bottom tab bar for detail and therefore increases the `<main>` height. If the list is scrolled near its maximum, that height increase lowers the maximum valid `scrollTop`; the browser clamps the parked scroller to that smaller value. When the tab bar returns, the clamped value does not rise again, so the list jumps upward instead of restoring the original position.

Please either save and restore the scroll offset explicitly or keep the parked pane at its list-mode height. The regression test should cover the real `MobileShell` bottom-tab transition with the list near its maximum scroll position; the current test mounts only `MobileWorkPage` and keeps the scroller geometry fixed, so it cannot catch this case.

baixiaohang (changes requested):
- Rationale: The parked list can still lose its near-bottom scroll offset because entering detail changes the shell height and forces the hidden scroller to clamp.

**Risk level**: A
- Path baseline: the synchronized head includes `packages/client/**` -> A.
- Semantic lift: none; the Client addition is test-only.

**PR summary**
- Problem: Mobile users should return from chat detail to the exact same Chat-list position and local list state.
- Approach: Keep one list instance mounted beneath detail; the synchronized head also stabilizes an unrelated Claude retry test under CI load.
- Impacted modules: mobile shell/list navigation, mobile DOM regression coverage, and a Client provider test.

**Review findings**
❌ 1. The preserved DOM node does not preserve the offset near the bottom. `MobileShell` removes the bottom tabs in detail, which increases `<main>` height; the parked pane remains `h-full`, so its scroller grows and the browser clamps any `scrollTop` above the new maximum. Restoring the tabs shrinks the viewport again but does not restore the clamped value. Please save/restore the offset explicitly or preserve list-mode geometry, and exercise the real shell/tab-height transition with a near-maximum scroll position in the regression test. [R4/R5 / `packages/web/src/pages/mobile/work.tsx:92`, `packages/web/src/pages/mobile/shell.tsx:78`, `packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx:173`]

**Action taken**
- Submitted request changes.

L42y:
This request-changes review is on `1af34f053`. Head `d1ee601b0` already addresses the tab-bar clamp:

- snapshot the list scroller’s `scrollTop` and `clientHeight` before opening detail
- park the list at that list-mode height instead of stretching with `<main>` when tabs hide
- restore `scrollTop` when returning
- regression covers `MobileShell` plus a near-max offset after a taller detail viewport would have clamped

Please re-review that commit.

yuezengwu (approved):
Re-reviewed head `d1ee601b0`. The previous near-bottom scroll-clamping blocker is resolved: list-mode height and `scrollTop` are captured before detail, the parked pane keeps that height while the shell tabs are hidden, and the offset is restored on return. The new `MobileShell` regression covers the near-maximum scroll case across the tab-bar height transition.

yuezengwu (approved):
Re-reviewed head `b7ff40aea`. The only delta from the previously approved revision is the regression assertion switching from a raw pixel string to a numeric `style.height` comparison for the design-token guard. The production fix and near-maximum shell transition coverage remain sound.

baixiaohang (changes requested):
- Rationale: The scroll-clamping fix is sound, but the latest assertion does not type-check because `querySelector()` returns `Element` and `Element` has no `style` property.

**Risk level**: A
- Path baseline: the PR includes `packages/client/**` -> A.
- Semantic lift: none; the Client addition remains test-only.

**PR summary**
- Problem: Mobile users should return from Chat detail to the exact same list position and local list state.
- Approach: Keep one list instance mounted, freeze its list-mode height while detail hides the bottom tabs, and restore the saved offset on return; the latest commit only revises the height assertion.
- Impacted modules: mobile Chat navigation and DOM regression coverage, plus a Client provider test.

**Review findings**
❌ 1. The new assertion fails TypeScript with `TS2339: Property 'style' does not exist on type 'Element'`, which breaks both `Lint & Type Check` and `Test Client & Web`. Please narrow the queried node to an `HTMLElement` (or otherwise use a typed DOM access) before reading `style.height`. [R4 / `packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx:283`]
✅ 2. The previous near-bottom scroll-clamping blocker remains resolved, and the regression still covers the real shell/tab-height transition.

**Action taken**
- Submitted request changes.

yuezengwu (approved):
Re-reviewed head `ab42cbe7`. The test now narrows the parked pane to `HTMLElement` before reading `style.height`, resolving the prior TypeScript error. The production scroll-restoration fix and shell-level near-maximum regression remain sound.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The mobile Chat list now preserves its real near-bottom position across the shell's list/detail height transition, the regression covers that geometry, and the current head passes all CI checks.

**Risk level**: A
- Path baseline: the PR includes `packages/client/**` -> A.
- Semantic lift: none; the Client addition is test-only.

**PR summary**
- Problem: Mobile users should return from Chat detail to the same list position and retain their local list state.
- Approach: Keep one list instance mounted, capture its offset and list-mode height before detail, freeze that geometry while the bottom tabs are hidden, and restore the offset on return. The Client test uses a bounded wall-clock wait for retry settlement under CI load.
- Impacted modules: mobile Chat navigation and DOM regression coverage, plus a Claude provider retry test.

**Review findings**
✅ 1. The previous near-bottom scroll-clamping blocker is resolved by preserving both list-mode height and `scrollTop`. [`packages/web/src/pages/mobile/work.tsx`]
✅ 2. The parked list remains inert and hidden from assistive technology while detail is active. [`packages/web/src/pages/mobile/work.tsx`]
✅ 3. The shell-level regression covers bottom-tab removal, a near-maximum offset, and restoration on return; its DOM access now type-checks. [`packages/web/src/pages/mobile/__tests__/chat-nav.test.tsx`]
✅ 4. All current-head CI checks pass, including `Lint & Type Check` and `Test Client & Web`.

**Action taken**
- Approved.

PR #2355, opened 2026-08-19
fix(chat): surface recovery for blocked and stale in-flight agents

L42y:
## Summary
- Treat handler `blocked` and stale in-flight runtime (`working` / `blocked` / `error`) as recovery via the existing `errored` → `failedAgentIds` chat-row path.
- Keep **fresh** `working` as green/slow-but-alive so long legitimate turns do not look stuck.
- Stale idle stays idle; Need you is unchanged.

## How states are judged
| Signal | Rule | Chat-row effect |
|---|---|---|
| Fresh `working` | active session + `runtime_state_at` within 60s + `runtime_state=working` | `busyAgentIds` (ActivityDots) |
| Fresh `blocked` | same freshness + `runtime_state=blocked` (handler: no output + no prompt) | `failedAgentIds` (`!`) |
| Fresh `error` / session `errored` | existing fault path | `failedAgentIds` (`!`) |
| Stale in-flight | active + stamp older than 60s + last state was working/blocked/error | `failedAgentIds` (`!`), not busy |
| Stale `idle` | last state idle | no recovery |

## Test plan
- [x] `agent-chat-status.test.ts` (46) including new blocked + stale in-flight cases
- [x] `me-chats-http-e2e.test.ts` (9) including blocked → `failedAgentIds`
- [ ] Manual: leave a chat green working with live activity for >1m → still working
- [ ] Manual: stop client mid-turn → after ~60–90s row shows recovery `!` instead of idle

### DIFF
--- a/packages/server/src/__tests__/agent-chat-status.test.ts
+++ b/packages/server/src/__tests__/agent-chat-status.test.ts
@@ -565,14 +565,26 @@ describe("agent-chat-status", () => {
       expect(s?.main).toBe("offline");
     });
 
-    it("stale per-chat runtime_state='error' fail-closes (does not surface as errored)", async () => {
+    it("stale per-chat in-flight runtime still surfaces as recovery (errored)", async () => {
       const { app, peer, chatId } = await newChatWithAgent();
       await bindPresence(peer.agent.uuid, peer.clientId);
       await setSession(peer.agent.uuid, chatId, "active");
-      // Seed error stamp older than RUNTIME_STALE_MS.
-      await setRuntime(peer.agent.uuid, chatId, "error", -(RUNTIME_STALE_MS + 5_000));
+      await setRuntime(peer.agent.uuid, chatId, "working", -(RUNTIME_STALE_MS + 5_000));
       const s = (await getChatAgentStatuses(app.db, chatId)).find((x) => x.agentId === peer.agent.uuid);
-      expect(s?.errored).toBe(false);
+      expect(s?.working).toBe(false);
+      expect(s?.errored).toBe(true);
+      expect(s?.main).toBe("failed");
+    });
+
+    it("fresh blocked runtime surfaces as recovery, not working", async () => {
+      const { app, peer, chatId } = await newChatWithAgent();
+      await bindPresence(peer.agent.uuid, peer.clientId);
+      await setSession(peer.agent.uuid, chatId, "active");
+      await setRuntime(peer.agent.uuid, chatId, "blocked");
+      const s = (await getChatAgentStatuses(app.db, chatId)).find((x) => x.agentId === peer.agent.uuid);
+      expect(s?.working).toBe(false);
+      expect(s?.errored).toBe(true);
+      expect(s?.main).toBe("failed");
     });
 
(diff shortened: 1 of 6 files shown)


baixiaohang (changes requested):
- Rationale: The per-chat stale/fault projection is sound, but the old-client fallback newly promotes agent-global `blocked` into a chat failure and reintroduces a known cross-chat false-positive path.

**Risk level**: B-high
- Path baseline: touches `packages/server/src/services/chat/**` session-status core logic.
- Semantic lift: none.

**PR summary**
- Problem: Users currently lose the chat-row recovery cue when an agent is explicitly blocked or its last in-flight runtime report becomes stale, making dead work look idle.
- Approach: Keep fresh `working` green, map fresh per-chat `blocked`/`error` and stale per-chat in-flight states to the existing failure/recovery projection, and refresh loaded conversation pages every 30 seconds so passive staleness becomes visible.
- Impacted modules: server chat-session status, workspace chat projection, shared status semantics, and the Web conversation list.

**Review findings**
❌ 1. Keep `blocked` out of the NULL-stamp old-client fallback. At `status.ts:689`, `isFaultRuntime(presenceRuntimeState)` changes the fallback from the existing agent-global `error` check to `error | blocked`. Because that signal is not chat-scoped, one old client's `blocked` state will mark every active chat containing that agent as failed. This is not the "existing prod envelope" described above: the base implementation only folds global `error`, while commit `234902f74` documents that legacy clients previously auto-produced `blocked` during legitimate 2–5 minute reasoning turns and removed that behavior specifically because it caused false alarms. Please preserve `presenceRuntimeState === "error"` for `runtimeStateAt == null` and update the new fallback assertion at `agent-chat-status.test.ts:806`; fresh/stale per-chat `blocked` can still use the new recovery behavior once a non-NULL per-chat stamp exists. [R5 / `packages/server/src/services/chat/sessions/status.ts:689`]
✅ 2. The stamped per-chat path correctly keeps fresh `working` alive, converts stale in-flight state to recovery, and leaves stale `idle` quiet; the shared status producer keeps list and detail consumers consistent.

**Action taken**
- Submitted request changes.

L42y:
Addressed finding 1 in 1be08c579:

- NULL-stamp old-client fallback is back to `presenceRuntimeState === "error"` only.
- Agent-global `blocked` no longer lights recovery (assertion updated).
- Fresh/stale **per-chat** `blocked` recovery is unchanged once `runtime_state_at` is non-NULL.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up preserves the intended per-chat recovery behavior while removing agent-global `blocked` from the NULL-stamp compatibility fallback.

**Risk level**: B-high
- Path baseline: touches `packages/server/src/services/chat/**` session-status core logic.
- Semantic lift: none.

**Review findings**
✅ The prior blocking finding is resolved at head `1be08c579`: old clients retain only the existing global `error` fallback, while fresh/stale per-chat `blocked` recovery remains scoped to rows with a per-chat runtime stamp.
✅ Current-head CI is green, including Server and Client/Web tests.

**Action taken**
- Approved with explicit maintainer authorization.

Issue #2038, opened 2026-07-28
[Bug] Codex runtime loops on a revoked PATH binary instead of falling back


Gandy2025:
## Summary

On macOS, the Codex runtime can select an older global npm Codex binary whose developer certificate has been revoked, classify the resulting AMFI `SIGKILL` as a transient smoke-check failure, and retry session startup indefinitely instead of falling back to a healthy later candidate such as the Codex bundled with ChatGPT Desktop.

## Surface Area

First Tree Client Runtime — Codex binary resolution, launch verification, and session-start retry policy.

## Environment

- affected First Tree channel: staging; the issue remained after the affected host auto-updated (the screenshot reports build suffix `.959.1`; the exact full version was not captured)
- filing CLI version: `0.5.18-staging.955.1`
- runtime: `codex`
- selected Codex: `~/.npm-global/bin/codex`, version `0.125.0`
- healthy later candidate: ChatGPT Desktop Codex `0.146.0-alpha.3`
- operating system: macOS; AMFI reported `CSSMERR_TP_CERT_REVOKED` (exact macOS version on the affected host was not captured)

## Reproduction

1. On macOS, leave an older global npm Codex binary on the daemon-visible resolution path.
2. Ensure that binary is present on disk but cannot launch because macOS rejects its revoked developer certificate.
3. Also have a healthy Codex candidate available later in First Tree's resolution order, such as ChatGPT Desktop's bundled Codex.
4. Start a Codex-backed First Tree agent and send a message that creates or resumes a session.
5. Observe First Tree resolve the older global binary and run `codex --version`.
6. Observe repeated `provider_retry_started` / `provider_retry_scheduled` events instead of fallback or a bounded actionable failure.

## Expected Behavior

First Tree should preserve automatic retry for a one-off cold-start or host-pressure smoke-check flake, but a definitively unusable candidate must not pin the runtime indefinitely.

- Launch-verify binary candidates in resolution order and continue to later candidates when one cannot launch.
- If every candidate fails, bound repeated identical startup failures and surface an actionable error that identifies the selected path and recommends upgrade/replacement.
- A healthy later candidate, including ChatGPT Desktop Codex, should be selected automatically.

## Actual Behavior

First Tree selects the first existing external candidate and does not continue to later candidates after launch verification fails.

The runtime reports:

```text
codex --version smoke check did not complete (transient host condition); will retry. Detail: `codex --version` killed by SIGKILL
```

[...]

PR #2356, opened 2026-08-20
fix(client): fall through unusable Codex binaries

yuezengwu:
## Summary

- launch-verify automatic Codex candidates in the existing daemon PATH → well-known dirs → login-shell PATH → macOS Desktop order, falling through unusable candidates to a later healthy binary
- keep capability detection existence-only and launch-free; runtime/login/SDK selection changes actively publish the exact still-discoverable verified artifact through the daemon capability refresher
- defer and coalesce targeted Codex provenance publication while browser login owns the provider entry, preserving `pendingAuth` until the interactive flow completes
- invalidate remembered provenance as soon as that exact path is deterministically rejected, even when a later candidate has a transient verification failure
- preserve one-off transient smoke failures, including through forced app-server startup, while bounding and resetting identical all-candidate failures with path-bearing `codex_binary_unusable` remediation
- preserve explicit `codexPathOverride` authority

## Tests

- `pnpm exec vitest run src/providers/codex/__tests__/binary.test.ts src/providers/codex/__tests__/app-server/extra-coverage.test.ts src/__tests__/capability-probes.test.ts src/__tests__/error-taxonomy.test.ts src/__tests__/provider-retry-policy.test.ts src/__tests__/provider-boundary-guard.test.ts src/__tests__/session-manager-retry.test.ts src/__tests__/session-start-error-signaling.test.ts` — 8 files, 263 tests passed
- `pnpm exec vitest run src/__tests__/daemon-start-command.test.ts` — 1 file, 37 tests passed
- `pnpm --filter first-tree-dev test` — 27/27 batches passed
- `pnpm --filter @first-tree/client test` — 214 files passed, 2 skipped; 2758 tests passed, 7 skipped
- `pnpm check` — passed; only pre-existing warnings/info outside the changed files
- `pnpm typecheck` — 9/9 tasks passed
- `git diff --check` — passed

## QA

- Tier: `test-only`
- Scope: Codex binary resolution, capability provenance publication, provider failure taxonomy, and session-start retry scheduling
- Case disposition: `candidate-case-update`; clarified the existing daemon-probe case so an install-only probe cannot claim an unverified path, while deterministic behavior remains covered by product tests
- Agent-run QA was not run because it was not requested
- No database, wire-schema-shape, migration, persistence, or Context Tree changes

Related to #2038

### DIFF
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -782,7 +782,7 @@ describe("daemon start command", () => {
     expect(unregisterCodexCandidateChange).toHaveBeenCalledTimes(1);
   });
 
-  it("defers an in-flight Codex provenance write until interactive login releases pendingAuth", async () => {
+  it("defers in-flight Codex provenance and preserves a terminal auth failure", async () => {
     const verifiedEntry = {
       state: "ok" as const,
       available: true,
@@ -807,6 +807,14 @@ describe("daemon start command", () => {
         expiresAt: "2026-08-20T00:05:00.000Z",
       },
     };
+    const failedEntry = {
+      ...verifiedEntry,
+      lastAuthError: {
+        reason: "exit-nonzero" as const,
+        message: "account not authorized",
+        at: "2026-08-20T00:01:00.000Z",
+      },
+    };
     coreMocks.runRuntimeAuthLogin.mockImplementationOnce(
       async (
         _command: { provider: string; ref: string },
@@ -814,6 +822,7 @@ describe("daemon start command", () => {
       ) => {
         await deps.setProviderEntry("codex", pendingEntry);
         await loginPending;
+        await deps.setProviderEntry("codex", failedEntry);
       },
     );
 
@@ -824,6 +833,13 @@ describe("daemon start command", () => {
     });
     refresherInstance.endInteractive.mockImplementation((provider: string) => {
       if (provider === "codex") codexInteractive = false;
+    });
+    let currentCodexEntry: unknown;
+    refresherInstance.currentEntry.mockImplementation((provider: string) =>
+      provider === "codex" ? currentCodexEntry : undefined,
+    );
+    refresherInstance.setProviderEntry.mockImplementation(async (provider: string, entry: unknown) => {
+      if (provider === "codex") currentCodexEntry = entry;
     });
 
     let finishSkillUpload!: () => void;
@@ -857,12 +873,18 @@ describe("daemon start command", () => {
     expect(refresherInstance.setProviderEntry).not.toHaveBeenCalledWith("codex", verifiedEntry);
 
     finishLogin();
-    await waitForAsyncWork(() => refresherInstance.setProviderEntry.mock.calls.length === 2);
+    await waitForAsyncWork(() => refresherInstance.setProviderEntry.mock.calls.length === 3);
     expect(refresherInstance.endInteractive).toHaveBeenCalledWith("codex");
-    expect(refresherInstance.setProviderEntry.mock.calls).toEqual([
+    expect(refresherInstance.setProviderEntry.mock.calls.slice(0, 2)).toEqual([
       ["codex", pendingEntry],
-      ["codex", verifiedEntry],
+      ["codex", failedEntry],
     ]);
+    const publishedAfterFailure = refresherInstance.setProviderEntry.mock.calls[2]?.[1];
+    expect(publishedAfterFailure).toMatchObject({
+      runtimePath: verifiedEntry.runtimePath,
+      lastAuthError: failedEntry.lastAuthError,
+    });
+    expect(publishedAfterFailure).not.toHaveProperty("pendingAuth");
 
     finishSkillUpload();
     await expect(start).rejects.toMatchObject({ exitCode: 1 });
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -782,6 +782,92 @@ describe("daemon start command", () => {
     expect(unregisterCodexCandidateChange).toHaveBeenCalledTimes(1);
   });
 
+  it("defers an in-flight Codex provenance write until interactive login releases pendingAuth", async () => {
+    const verifiedEntry = {
+      state: "ok" as const,
+      available: true,
+      runtimeSource: "path" as const,
+      runtimePath: "/Applications/ChatGPT.app/Contents/Resources/codex",
+      detectedAt: "2026-08-20T00:00:00.000Z",
+    };
+    let resolveProbe!: (entry: typeof verifiedEntry) => void;
+    const delayedProbe = new Promise<typeof verifiedEntry>((resolve) => {
+      resolveProbe = resolve;
+    });
+    clientMocks.probeCodexCapability.mockReturnValueOnce(delayedProbe);
+
+    let finishLogin!: () => void;
+    const loginPending = new Promise<void>((resolve) => {
+      finishLogin = resolve;
+    });
+    const pendingEntry = {
+      state: "ok" as const,
+      pendingAuth: {
+        method: "browser" as const,
+        expiresAt: "2026-08-20T00:05:00.000Z",
+      },
+    };
+    coreMocks.runRuntimeAuthLogin.mockImplementationOnce(
+      async (
+        _command: { provider: string; ref: string },
+        deps: { setProviderEntry: (provider: string, entry: unknown) => Promise<void> },
+      ) => {
+        await deps.setProviderEntry("codex", pendingEntry);
+        await loginPending;
+      },
+    );
+
+    let codexInteractive = false;
+    refresherInstance.isInteractive.mockImplementation((provider: string) => provider === "codex" && codexInteractive);
+    refresherInstance.beginInteractive.mockImplementation((provider: string) => {
+      if (provider === "codex") codexInteractive = true;
+    });
+    refresherInstance.endInteractive.mockImplementation((provider: string) => {
+      if (provider === "codex") codexInteractive = false;
+    });
+
+    let finishSkillUpload!: () => void;
+    const skillUploadPending = new Promise<void>((resolve) => {
+      finishSkillUpload = resolve;
+    });
+    coreMocks.uploadAgentSkills.mockImplementationOnce(async () => {
+      codexCandidateChangeListener?.();
+      await skillUploadPending;
+    });
+
+    const start = runStart(["--foreground"]);
+    await waitForAsyncWork(() => clientMocks.probeCodexCapability.mock.calls.length === 1);
+
+    const runtimeAuthStart = runtimeInstance.onRuntimeAuthStart.mock.calls[0]?.[0] as
+      | ((command: { provider: string; ref: string }) => void)
+      | undefined;
+    if (typeof runtimeAuthStart !== "function") throw new Error("runtime-auth callback was not registered");
+    runtimeAuthStart({ provider: "codex", ref: "pending-provenance" });
+    await waitForAsyncWork(() =>
+      refresherInstance.setProviderEntry.mock.calls.some(
+        ([provider, entry]) => provider === "codex" && entry === pendingEntry,
+      ),
+    );
+
+    // The launch-free provenance probe finishes after pendingAuth is visible.
+    // Its plain entry must not replace the interactive row.
+    resolveProbe(verifiedEntry);
+    await Promise.resolve();
+    await Promise.resolve();
+    expect(refresherInstance.setProviderEntry).not.toHaveBeenCalledWith("codex", verifiedEntry);
+
+    finishLogin();
+    await waitForAsyncWork(() => refresherInstance.setProviderEntry.mock.calls.length === 2);
+    expect(refresherInstance.endInteractive).toHaveBeenCalledWith("codex");
+    expect(refresherInstance.setProviderEntry.mock.calls).toEqual([
+      ["codex", pendingEntry],
+      ["codex", verifiedEntry],
+    ]);
+
+    finishSkillUpload();
+    await expect(start).rejects.toMatchObject({ exitCode: 1 });
+  });
+
   it("skips skill upload for stale local aliases that are not pinned to this client", async () => {
     mkdirSync(join(home, "config", "agents", "developer"), { recursive: true });
     writeFileSync(
--- a/apps/cli/src/__tests__/daemon-start-command.test.ts
+++ b/apps/cli/src/__tests__/daemon-start-command.test.ts
@@ -12,7 +12,9 @@ const clientMocks = vi.hoisted(() => ({
   discoverClaudeCodeSkills: vi.fn(),
   flushClientSentry: vi.fn(),
   initClientSentry: vi.fn(),
+  onCodexVerifiedAutomaticCandidateChange: vi.fn(),
   probeCapabilities: vi.fn(),
+  probeCodexCapability: vi.fn(),
   reprobeOnReconnect: vi.fn(),
 }));
 
@@ -94,6 +96,8 @@ const exitSpy = vi.spyOn(process, "exit").mockImplementation((code?: string | nu
 
 let home: string;
 let runtimeOwnershipRelease: ReturnType<typeof vi.fn>;
+let codexCandidateChangeListener: (() => void) | null;
+let unregisterCodexCandidateChange: ReturnType<typeof vi.fn>;
 let runtimeInstance: {
   addAgent: ReturnType<typeof vi.fn>;
   start: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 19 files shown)


baixiaohang (changes requested):
- Rationale: The ordered fallback fixes the reported stale-binary loop, but the new selection and retry state are not yet consistent across capability reporting, explicit retry, and forced app-server startup.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can find a revoked global Codex binary before a healthy ChatGPT Desktop binary, then retry the unusable path indefinitely instead of starting the agent.
- Approach: Enumerate automatic Codex candidates in priority order, launch-check each until one succeeds, distinguish unusable from missing binaries, and bound repeated identical all-candidate failures while keeping execution and login on the shared resolver.
- Impacted modules: Codex binary/runtime resolution, capability reporting, binary-failure taxonomy, provider retry policy, and session-start regression coverage.

**Review findings**
❌ 1. The capability snapshot can now report a different `runtimePath` from the binary execution and login actually select. `probeCodexCapability` still calls `findCodexExecutableOnPath` and publishes the first existing candidate, while the new runtime resolver can reject that candidate and select a later one. In the motivating stale-global/healthy-Desktop case, diagnostics keep claiming the revoked global path backs Codex even though sessions run the Desktop path. This breaks the existing capability contract that `runtimePath` identifies the runtime artifact. Please either keep provenance aligned with the selected runtime or explicitly redesign the capability field/contract and cover the two-candidate case end to end. [R1/R5 / `packages/client/src/providers/codex/capability.ts:207`, `packages/client/src/providers/codex/capability.ts:266`]

❌ 2. The identical-failure counter survives the terminal result. At attempt 3, `codexAutomaticCandidateFailureError` returns `CodexBinaryUnusableError` without clearing the module-global tracker entry. SessionRuntime then removes that failed session, but a later user-initiated retry in the same chat starts at count 4 and terminalizes on its first matching transient smoke failure instead of receiving the documented initial attempt plus two retries; terminal chats also accumulate permanently in the map. Reset the sequence when its terminal outcome is emitted, or scope/clean it at the session-attempt lifecycle, and add a regression for a fresh retry after terminal settlement. [R4 / `packages/client/src/providers/codex/binary.ts:191`]

❌ 3. `resolveCodexRuntimeBinary` discards the named transient error and returns only `failure.message`. Forced app-server startup then wraps that string in `CodexAppServerStartupError`, so `recognizeProviderBinaryFailure` cannot see `CodexBinaryVerifyTransientError`; varying transient fingerprints fall into the generic `unknown` budget and stop as `unknown_exhausted` after three attempts, without the intended Codex transient semantics or path-bearing unusable remediation. Preserve a structured/named failure through the resolver/app-server boundary and cover forced app-server session startup with non-identical transient smoke failures. [R1/R4 / `packages/client/src/providers/codex/capability.ts:225`]

✅ 4. The ordered candidate walk, explicit SDK override guard, and deterministic-unusable taxonomy are otherwise cohesive and have focused hermetic coverage.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed all three requested changes in `6de41097f`:

1. Capability probing now walks and launch-verifies the same ordered external automatic candidates as runtime/login, so `runtimePath` reports the healthy selected artifact (including the later ChatGPT Desktop candidate).
2. The identical transient-failure tracker resets whenever it emits a terminal result; the regression test runs a complete second retry cycle for the same chat scope.
3. Runtime resolution preserves the named failure as a structured cause through forced app-server startup, and binary-failure classification follows bounded cause chains. A session-start regression covers four non-identical transient smoke failures without falling into `unknown_exhausted`.

Validation:
- focused Codex/capability/retry suite: 8 files, 261 tests passed
- Client suite: 214 files passed, 2 skipped; 2756 tests passed, 7 skipped
- `pnpm check`, `pnpm typecheck`, and `git diff --check` passed

Please re-review when convenient.

baixiaohang (changes requested):
- Rationale: The retry-state and forced app-server fixes now hold, but capability provenance was aligned by making an install-only probe launch Codex binaries, which reintroduces the false-negative and latency behavior that the shared capability contract explicitly excludes.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can encounter a revoked global Codex binary before a healthy ChatGPT Desktop binary and fail to start an otherwise usable agent.
- Approach: Walk automatic candidates in priority order, launch-check them for real execution/login, classify unusable versus transient failures, and bound repeated identical failures.
- Impacted modules: Codex binary/runtime resolution, capability reporting, provider failure taxonomy, and session-start retry scheduling.

**Review findings**
❌ 1. The capability-path fix now violates the shared install-only detection contract. `probeCodexCapability` calls `resolveVerifiedCodexAutomaticCandidate`, so background detection executes `codex --version` for each external candidate even though the common probe contract explicitly says detection never launches provider binaries. A single transient smoke failure therefore changes an installed Codex capability to `error`/unavailable, and normal non-ok re-probing can repeatedly block the daemon for up to the per-candidate 10-second timeout. This restores the false-negative and latency behavior the install-only design removed. Please keep the capability probe existence-only and solve verified execution provenance at the runtime boundary (for example by separating discovered-path provenance from the last runtime-verified selection), with a regression asserting the probe never calls `verifyPath`/spawns Codex. [R1/R5 / `packages/client/src/providers/codex/capability.ts:275`, `packages/client/src/providers/capabilities/detect.ts:3`]

✅ 2. Terminal identical-failure results now clear their tracker scope, and the new second-cycle regression demonstrates that a later explicit retry receives a fresh budget. [R4 / `packages/client/src/providers/codex/binary.ts:185`]

✅ 3. Forced app-server startup now preserves the named transient cause, and bounded cause-chain recognition prevents non-identical Codex smoke failures from degrading into `unknown_exhausted`. [R1/R4 / `packages/client/src/runtime/provider-support/binary-failure.ts`]

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the remaining install-only probe concern in `6df5de736`.

- `probeCodexCapability` is existence-only again and never calls `codex --version`.
- Before any runtime verification, an external Codex capability remains `ok` but intentionally omits `runtimePath` instead of claiming the first existence-only candidate.
- Successful runtime/login resolution and SDK fallback record the exact verified automatic candidate in process-local, injectable state. A later probe reports that path only while it remains in the current existence-only candidate set.
- Missing, bundled, and terminal-unusable outcomes clear the remembered automatic selection; one-off transient failures retain the last verified selection.
- Hermetic coverage uses an executable that writes a marker if launched and proves the capability probe leaves the marker absent. The stale-global → healthy-Desktop regression also proves verification occurs only at runtime and that the later probe reports the Desktop path.

Fresh validation on this head:
- focused Codex/capability/retry suite: 8 files, 261 tests passed
- Client suite: 214 files passed, 2 skipped; 2756 tests passed, 7 skipped
- `pnpm check`, `pnpm typecheck`, and `git diff --check` passed

Please re-review when convenient.

baixiaohang (changes requested):
- Rationale: Capability probing is launch-free again, but the new verified-path cache is not connected to the production capability publication lifecycle and can still retain a path that the latest resolution deterministically rejected.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can encounter a revoked global Codex binary before a healthy ChatGPT Desktop binary and fail to start an otherwise usable agent.
- Approach: Keep capability discovery existence-only, launch-check automatic candidates only at real runtime/login/SDK boundaries, and remember the last verified automatic path for later capability provenance.
- Impacted modules: Codex binary/runtime resolution, capability reporting, provider failure taxonomy, and session-start retry scheduling.

**Review findings**
❌ 1. A normal successful runtime/SDK fallback records the path only in the module-local tracker; it does not trigger a capability re-probe or update. Once the startup snapshot is `ok`, `CapabilityRefresher` stops polling, so the server continues to expose the pre-runtime entry (usually `runtimeSource: "path"` with no `runtimePath`, or an older verified path) until an unrelated reconnect, manual probe, or login flow happens. The new test calls `probeCodexCapability` directly after resolution, but no equivalent production edge exists. Wire the verified-selection change into the capability publisher (with a production-lifecycle regression), or explicitly remove the promise that ordinary execution updates this provenance. If the intentionally path-less `runtimeSource: "path"` state remains, update the shared contract and the existing daemon-probe QA case, which currently require matching path provenance. [R1/R5 / `packages/client/src/providers/codex/binary.ts:311`, `apps/cli/src/core/capability-refresh.ts:334`, `packages/qa/cases/runtime/daemon-probe-capability.md:47`]

❌ 2. The cache is cleared only when the aggregate all-candidate result becomes terminal. If the remembered global path now fails deterministically while a later Desktop candidate has a transient failure, the aggregate result is transient and lines 229-232 retain the remembered global path. A subsequent probe still finds that file and reports it as the verified `runtimePath`, even though the latest resolver explicitly rejected it — the same stale-path inconsistency this cache is meant to prevent. Clear the remembered selection when that specific path has a non-transient failure, even if another candidate keeps the overall attempt retryable, and cover the mixed deterministic/transient sequence. [R4/R5 / `packages/client/src/providers/codex/capability.ts:229`, `packages/client/src/providers/codex/binary.ts:294`]

✅ 3. The capability probe itself is existence-only again, and the marker-file regression demonstrates that it does not execute the discovered Codex binary.

✅ 4. The prior terminal-counter cleanup and forced app-server cause-chain fixes remain intact on this head.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed both latest findings in `e88560d31`.

1. Production publication lifecycle: the default verified-candidate tracker now emits only on actual selection changes. Daemon startup subscribes before `ClientRuntime.start()`, coalesces changes until the post-register `CapabilityRefresher` is active, then re-probes Codex existence-only and publishes the provider entry with `setProviderEntry`. The listener is removed during shutdown/final cleanup. A daemon-start regression fires the change during runtime startup and proves the Desktop path reaches the live refresher.
2. Mixed deterministic/transient invalidation: both runtime and SDK fallback clear remembered provenance immediately when that exact path has a non-transient candidate failure, before aggregate failure classification. Regressions cover stale-global deterministic rejection plus a later Desktop cold-start timeout; the aggregate remains `CodexBinaryVerifyTransientError`, while the stale path is absent from the subsequent capability entry.

The existing optional `runtimePath` contract and daemon-probe QA case now explicitly distinguish existence-only discovery from a runtime-verified selection. This changes documentation/expectations only; the wire schema shape is unchanged.

Fresh validation on this head:
- focused Codex/capability/retry suite: 8 files, 263 tests passed
- daemon production-lifecycle suite: 1 file, 36 tests passed
- Client suite: 214 files passed, 2 skipped; 2758 tests passed, 7 skipped
- `pnpm check`, `pnpm typecheck`, and `git diff --check` passed

Please re-review when convenient.

baixiaohang (changes requested):
- Rationale: Verified-path publication and mixed-failure invalidation are now connected correctly, but the new targeted publisher can overwrite Codex's in-flight browser-login marker.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can encounter a revoked global Codex binary before a healthy ChatGPT Desktop binary and fail to start an otherwise usable agent.
- Approach: Keep capability discovery launch-free, verify ordered candidates only at real execution/login boundaries, publish verified provenance through the daemon refresher, and bound repeated transient all-candidate failures.
- Impacted modules: Codex binary/runtime resolution, daemon capability publication, runtime authentication, provider failure taxonomy, and session-start retry scheduling.

**Review findings**
❌ 1. The targeted Codex provenance write bypasses the refresher's interactive-provider ownership. `beginInteractive("codex")` is set before `runRuntimeAuthLogin`, and `resolveLogin()` then verifies/records the automatic path, which starts `publishCodexVerifiedSelection`. The login flow subsequently publishes `pendingAuth`, but the asynchronous provenance probe can finish afterward and call `setProviderEntry("codex", entry)` with a plain probe entry, replacing the whole Codex row and dropping `pendingAuth`. This recreates the existing "Connect panel vanishes mid-login" failure that aggregate probes explicitly preserve against. Defer/coalesce provenance publication while Codex is interactive, or merge the current interactive markers atomically, and add a production-lifecycle regression that holds the provenance probe until after `pendingAuth` is published. [R4/R5 / `apps/cli/src/commands/daemon/start.ts:398`, `apps/cli/src/core/runtime-auth-login.ts:208`, `apps/cli/src/core/capability-refresh.ts:186`]

✅ 2. Successful runtime/SDK selection changes now reach the live capability refresher, including selections made during runtime startup.

✅ 3. A deterministically rejected remembered path is now cleared before aggregate transient classification, with both runtime and SDK regressions.

✅ 4. The shared schema commentary and daemon-probe QA case now describe the launch-free, temporarily path-less provenance state consistently.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the interactive-provider ownership finding in `d3801f6e8`.

- Codex provenance publication now defers immediately when the capability refresher reports Codex as interactive.
- An already-running launch-free provenance probe re-checks ownership at write time; if browser login began while it was in flight, the result is coalesced instead of replacing the whole Codex row.
- The runtime-auth completion path releases interactive ownership first, then drains exactly one pending provenance refresh.
- Added a production daemon-lifecycle regression that starts the provenance probe, publishes `pendingAuth`, releases the delayed probe, and proves the plain verified entry is not written until login completes.

Fresh validation:
- daemon-start focused suite: 37/37 passed
- `first-tree-dev` CLI package: 27/27 batches passed
- `pnpm check`: passed (pre-existing warnings/info only)
- `pnpm typecheck`: 9/9 tasks passed
- `git diff --check`: passed

Please re-review when convenient.

baixiaohang (changes requested):
- Rationale: The in-flight `pendingAuth` race is fixed, but the deferred provenance drain can still erase the terminal `lastAuthError` immediately after a failed Codex login.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can encounter a revoked global Codex binary before a healthy ChatGPT Desktop binary and fail to start an otherwise usable agent.
- Approach: Keep capability discovery launch-free, verify ordered candidates only at real runtime/login boundaries, publish verified provenance through the daemon refresher, and bound repeated transient all-candidate failures.
- Impacted modules: Codex binary/runtime resolution, daemon capability publication, runtime authentication, provider failure taxonomy, and session-start retry scheduling.

**Review findings**
❌ 1. The deferred provenance write now waits until interactive ownership ends, but it still overwrites the login's terminal failure. A Codex login failure is reflected first with `lastAuthError`; then `runRuntimeAuthLogin(...).finally()` calls `endInteractive()` and immediately drains `codexCapabilityPublicationPending`. That drain re-probes a plain capability entry and `setProviderEntry()` replaces the whole Codex row, removing `lastAuthError` and returning the Connect panel to the never-attempted state. The new lifecycle test models only `pendingAuth` followed by a successful completion, so it misses this ordering. Please make the terminal reflection the last writer on failure (or atomically preserve its marker while adding provenance) and add a production-lifecycle regression with pending provenance plus a failed login that asserts `lastAuthError` remains after interactive ownership is released. [R4/R5 / `apps/cli/src/commands/daemon/start.ts:461`, `apps/cli/src/core/runtime-auth-login.ts:184`, `packages/shared/src/schemas/client-capabilities.ts:94`]

✅ 2. The new interactive checks correctly prevent both queued and already-running provenance probes from replacing `pendingAuth` during browser login.

✅ 3. The earlier launch-free detection, verified-selection publication, stale-provenance invalidation, terminal retry cleanup, and structured transient-cause fixes remain intact on this head.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the terminal `lastAuthError` finding in `f5d13d76f`.

- The deferred Codex provenance publisher now treats its update as additive: at write time it reads the current Codex entry and carries forward an existing terminal `lastAuthError` while refreshing the verified runtime path.
- A successful login still clears the marker through the runtime-auth terminal re-probe; a failed login remains visibly failed until the next login start or successful re-probe, matching the capability contract.
- Expanded the production daemon-lifecycle regression so the delayed provenance probe overlaps `pendingAuth`, the login then publishes an `exit-nonzero` terminal failure, and the final provenance entry is asserted to retain `lastAuthError` without retaining `pendingAuth`.

Fresh validation:
- daemon/auth/refresher focused suite: 3 files, 80 tests passed
- `pnpm check`: passed (36 pre-existing warnings, 8 infos)
- `pnpm typecheck`: 9/9 tasks passed
- `git diff --check`: passed

Please re-review when convenient.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The final runtime-auth publication race is resolved on the authorized head, all earlier blockers remain fixed, and the complete GitHub check suite is green.

**Risk level**: A
- Semantic lift: none (already A)

**PR summary**
- Problem: A macOS daemon can find a revoked global Codex binary before a healthy ChatGPT Desktop binary, preventing an otherwise usable agent from starting and repeatedly retrying the wrong path.
- Approach: Launch-verify automatic candidates in priority order at real runtime boundaries, keep background capability detection launch-free, publish only runtime-verified provenance, and bound transient all-candidate failures.
- Impacted modules: Codex binary/runtime resolution, daemon capability publication, runtime authentication, provider failure taxonomy, and session-start retry scheduling.

**Review findings**
✅ 1. Automatic resolution now falls through deterministically unusable candidates while preserving bounded retry semantics for transient launch failures.

✅ 2. Capability detection remains existence-only, while verified runtime/login/SDK selections reach the production publisher and stale provenance is invalidated at the correct candidate boundary.

✅ 3. Deferred provenance publication no longer overwrites either `pendingAuth` or terminal `lastAuthError`; the daemon lifecycle regression covers the pending-to-failed ordering and the final additive provenance write.

✅ 4. Terminal retry state resets correctly, and structured transient causes survive forced app-server startup without degrading into the generic unknown budget.

**Action taken**
- Approved with explicit authorization from baixiaohang.

PR #2354, opened 2026-08-19
feat(web): surface configured Team Skills in slash commands

yuezengwu:
## Summary

- Surface each eligible recipient Agent's configured Team Skills in the Web slash-command menu, including after one committed recipient mention in a group chat.
- Fetch the full Agent Resources response only after a legal slash trigger and a resolved recipient. Prose, paths, mount-only chat state, ambiguous recipients, unsupported Clients, and stale/error status paths do not fetch or expose Team rows.
- Reconcile Team rows with runtime commands deterministically. Team intent wins a folded base-name collision, while disabled, replaced, unavailable, malformed, and normalized-duplicate resources fail closed.
- Treat menu selection as transient provenance rather than immediate execution. Send-time revalidation checks the exact resource, requested slug, recipient, and Agent config version; a changed row blocks the send and preserves the draft and attachment.
- Validate that precondition inside the Server message transaction under the existing `agent_configs` row lock, then stamp a canonical versioned `messages.metadata.teamSkillInvocation` marker owned by the Server.
- Scope every Team rewrite, fence, notice, and recovery path to marker-present messages. A hand-typed local/runtime command without the marker remains byte-exact even if a configured Team Skill owns the same base slug.
- Resolve a valid marker at the Client's shared provider-input boundary against the exact reconciled resource identity. If an unmanaged local Skill owns the base name, `/review` is rewritten only in the provider view to the verified effective target such as `/review-first-tree`.
- Apply the same identity and custody rules to text and file/image captions, including routed group mentions. Unknown, malformed, stale, replaced, removed, or unavailable marked Team intent becomes an inert no-slash notice instead of falling through to a same-named local Skill.
- Pair Team sends with the request-only `purpose: "team-skill-invocation-v1"` sentinel and `skillPrecondition`. A rolled-back old Server rejects the unknown purpose instead of stripping the precondition and persisting a bare command; the new Server enforces the pair in both directions.
- Preserve transient Team provenance in failed-send custody. An unchanged text or captioned-image retry revalidates the resource and sends the protocol pair again; explicit clear/edit consumes that provenance, so a subsequently hand-typed same-slug command retains ordinary local/runtime semantics.
- Bound registry recovery through the real handler-issued `DeliveryToken.retry`, start/resume failure, and ACK/custody paths so an unresolved marked command gets at most one fresh-handler recovery and cannot hot-loop or disappear.

[...]

PR #2359, opened 2026-08-21
fix(cli): exclude hidden skills from public copy count

yuezengwu:
## Summary

- Exclude dot-prefixed Skill directories from `copyAllSkillPayloads().publicCount` while continuing to copy them verbatim into the target.
- Keep `directoryNames()` unfiltered so private-variant registry and variant-name exactness checks remain strict.
- Add a deterministic regression test covering both the public count and hidden-directory payload preservation.

## Context

Current `main` added `skills/.experimental/`, exposing a mismatch between the copy helper's reported public count and the target-side public-directory definition. The CLI shard then failed with `expected 9 / received 8` in `copy-skill-payloads.test.ts:875`.

This is an independent mainline fix so [#2354](https://github.com/first-tree-ai/first-tree/pull/2354) does not carry an unrelated CLI change.

## Validation

Independent `test-only` QA passed on exact head `ec068a1ae6d9f0e03b6aa1d9fa665d3b6a3ad369` with a clean worktree before and after:

- Parent `d48aacb523d50e037a17cf3a00ee75f623267ae5` reproduced the original failure.
- Focused `copy-skill-payloads` suite: 63/63 passed.
- CLI typecheck passed.
- Biome passed both changed files; `git diff --check` passed.
- A read-only semantic probe confirmed `.experimental` source/target recursive paths and file hashes are identical.
- A temporary `.rogue` private variant directory remained rejected by the unchanged exact-name validator.

## Change surface

- [x] `scripts/copy-skill-payloads.mjs`
- [x] `apps/cli/tests/copy-skill-payloads.test.ts`
- [ ] database, schema, migration, persistence semantics, package version, lockfile, or Skill payload content

## QA limits

This is localized deterministic `test-only` validation. It does not claim live CLI installation, cross-platform behavior, or release qualification.

### DIFF
--- a/apps/cli/tests/copy-skill-payloads.test.ts
+++ b/apps/cli/tests/copy-skill-payloads.test.ts
@@ -882,6 +882,28 @@ describe("verified copies", () => {
     });
   });
 
+  it("copies hidden dot-directories verbatim but never counts them as public skills", () => {
+    const parent = tempDir("first-tree-skill-target-hidden-");
+
+    withTrustedSkillsTarget({ trustedParentDir: parent }, (capability) => {
+      const result = copyAllSkillPayloads({ target: capability, clean: true });
+      const sourceNames = readdirSync(capability.sourceSkillsRoot, { withFileTypes: true })
+        .filter((entry) => entry.isDirectory())
+        .map((entry) => entry.name);
+      const hidden = sourceNames.filter((name) => name.startsWith("."));
+      // The count is exactly the non-hidden source directories, whatever the
+      // repo currently carries — no hidden directory may inflate it.
+      expect(result.publicCount).toBe(sourceNames.length - hidden.length);
+      // Hidden directories still ride the verbatim copy (e.g. `.experimental`)
+      // — excluded from the count, never from the payload.
+      for (const name of hidden) {
+        expect(existsSync(join(capability.targetSkillsRoot, name)), `hidden directory ${name} must be copied`).toBe(
+          true,
+        );
+      }
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The fix makes the reported public count match the existing dot-hidden boundary without changing recursive payload copying or private-variant validation.

**Risk level**: A
- Path baseline: includes `apps/cli/tests/**` -> A
- Semantic lift: none

**PR Summary**
- Author / repo: yuezengwu / first-tree
- Problem: Adding the hidden `.experimental` Skill payload made CLI packaging report nine public Skills while downstream public-directory checks correctly saw eight.
- Approach: Exclude dot-prefixed source directories only when computing `publicCount`; continue copying the complete source tree and validating the exact private-variant registry. The regression test pins both the count and hidden-payload presence.
- Impacted modules: `scripts/copy-skill-payloads.mjs`, `apps/cli/tests/copy-skill-payloads.test.ts`

**Review findings**
✅ 1. The source-side count now uses the same non-dot-directory definition as the existing target-side public inventory, fixing the mismatch without changing the copied payload set.
✅ 2. `directoryNames()` remains unfiltered, so the variant registry and per-variant exact-name checks retain their strictness.
✅ 3. The change preserves the release contract: eight public Core Skill names while hidden/auxiliary payload directories remain available in the packaged artifact.

**Action taken**
- Approved exact head `ec068a1ae6d9f0e03b6aa1d9fa665d3b6a3ad369` after all current CI checks completed successfully.

PR #2360, opened 2026-08-21
[Fix] kimi-code provider: register approval handler so hosted sessions can exit plan mode

L42y:
## Symptom

A hosted kimi-code agent session deadlocked in plan mode: the agent called `ExitPlanMode` 9 times and every call was dismissed with the synthetic result "Plan approval dismissed. Plan mode remains active." Plan mode then kept denying all writes (`CronCreate`, `Edit`, and every other write tool), so the hosted agent could make no progress at all.

## Root cause

- The First Tree kimi-code provider (`packages/client/src/providers/kimi-code/index.ts`) runs headless sessions with `permission: "yolo"` and never registers an approval handler.
- Inside `@botiverse/kimi-code-sdk`, `ExitPlanModeReviewAskPermissionPolicy` escalates `ExitPlanMode` to an approval request in every permission mode except `"auto"` — including `"yolo"`.
- The headless RPC `requestApproval` returns `{decision: "cancelled", feedback: "No approval handler registered."}` when no handler is registered.
- Result: plan exits are auto-cancelled forever, plan mode stays active, and plan mode keeps blocking writes — a permanent deadlock with no interactive surface to recover through.

## Fix

Register a module-level approval handler on every kimi-code session, on both the create and resume paths:

- If the request is a plan exit (`toolName === "ExitPlanMode"` or `display.kind === "plan_review"`), return `{ decision: "approved" }` — restoring the intended yolo posture for hosted sessions.
- Otherwise return `{ decision: "cancelled", feedback: "No interactive approval surface in hosted sessions." }` — matching the previous handler-less behavior for everything else.

First Tree code only; the SDK and `patches/` are untouched.

## Test coverage

`packages/client/src/providers/kimi-code/__tests__/handler.test.ts`:

- `FakeSession` gains a `setApprovalHandler` spy.
- New test: after `start` and `resume`, the handler is registered; invoking it with an `ExitPlanMode`/`plan_review` request returns `{ decision: "approved" }`; invoking it with a non-plan request (`Bash`) returns `cancelled` with the hosted-session feedback.

## Verification

- `pnpm --filter @first-tree/client exec vitest run src/providers/kimi-code/__tests__/handler.test.ts` — all tests pass (including the new one).
- `pnpm --filter @first-tree/client typecheck` — clean.
- `biome check` on the two touched files — clean.

### DIFF
--- a/packages/client/src/providers/kimi-code/__tests__/handler.test.ts
+++ b/packages/client/src/providers/kimi-code/__tests__/handler.test.ts
@@ -3,6 +3,7 @@ import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type {
+  ApprovalHandler,
   CreateSessionOptions,
   Event as KimiEvent,
   KimiHarnessOptions,
@@ -24,6 +25,7 @@ class FakeSession {
   readonly listeners = new Set<(event: KimiEvent) => void>();
   readonly setPermission = vi.fn().mockResolvedValue(undefined);
   readonly setModel = vi.fn().mockResolvedValue(undefined);
+  readonly setApprovalHandler = vi.fn();
   readonly cancel = vi.fn().mockResolvedValue(undefined);
   readonly close = vi.fn().mockResolvedValue(undefined);
   usage: SessionUsage = {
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change closes the hosted Kimi plan-mode deadlock at the provider boundary while preserving fail-closed behavior for every non-plan approval request.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: A hosted Kimi agent that enters plan mode cannot obtain interactive approval, so every exit attempt is cancelled and all subsequent writes remain blocked.
- Approach: Register one hosted-session approval handler before the first prompt on both create and resume. It approves plan-exit requests and cancels all other requests with explicit hosted-session feedback.
- Impacted modules: `packages/client/src/providers/kimi-code/`

**Review findings**
✅ 1. The pinned SDK confirms that valid `ExitPlanMode` calls require approval outside auto mode and that a missing handler returns `cancelled`; the fix addresses that exact boundary.
✅ 2. Registration covers both session creation and restoration before provider input, and the SDK clears the handler when the session closes.
✅ 3. Non-plan approvals remain fail closed, so the change does not broaden the hosted runtime's general approval surface.

**Action taken**
- Approved the unchanged head `f213bc5e8ed0485a6d2edf7ae5cd29f4888fc489` after explicit maintainer authorization.

yuezengwu (approved):
Reviewed exact head `f213bc5e8ed0485a6d2edf7ae5cd29f4888fc489`.

The provider registers the hosted approval handler before the first prompt on both create and resume. The pinned SDK evidence confirms that non-auto plan exits are otherwise routed through approval and that missing handlers return `cancelled`; SDK session close also clears the handler. Non-plan approval requests remain fail closed.

Independent validation in a clean detached worktree:
- `pnpm --filter @first-tree/client exec vitest run src/providers/kimi-code/__tests__/handler.test.ts` — 38 passed
- `pnpm --filter @first-tree/client typecheck` — passed
- `pnpm exec biome check packages/client/src/providers/kimi-code/index.ts packages/client/src/providers/kimi-code/__tests__/handler.test.ts` — passed
- `git diff --check` — passed

PR #2338, opened 2026-08-14
feat: add agent-local context fallback

yuezengwu:
## Summary

- Add an Agent-owned live Local Context Tree for managed Agents whose Team is authoritatively unbound, while keeping valid Team bindings remote and treating unknown or invalid authority as fail-closed `none`.
- Project one trusted Context source atomically across the workspace manifest, identity, briefing, managed Skill variant, and provider admission. A monotonic remote-observed latch permanently disables automatic Local fallback without acting as a stale Remote lease.
- Keep the public `first-tree-read` and `first-tree-write` identities stable while packaging mutually exclusive remote/BYO and private Local variants. Local reads and writes use the standard Tree content model without Git, content locking, candidate branches, or rollback promises.
- Add guarded Local filesystem lifecycle and CLI support, including containment/symlink checks, explicit deletion warnings, and filesystem-mode hierarchy reads that do not run repository binding guards or Git pulls. Read intent may complete only the first scaffold materialization whose root it atomically created; an existing invalid root is never repaired by Read, while Write may repair it explicitly.
- Harden Skill payload copying around opened-directory identity, replacement races, partial-open cleanup, and exact propagation of arbitrary thrown values; preserve the same 8-public + 2-private inventory in npm and portable release paths.
- Integrate the feature with the current SessionRuntime authority split and capability-degradation contract: source keys remain under Projection custody, admission generations under SlotScheduler, teardown under RouteTeardown, and ordinary non-Tree work stays silent when Tree-dependent operations are unavailable.
- Close review/security findings without weakening behavior: provider test paths are passed to static `node -e` programs as argv, and the turn-end serialization race guard uses explicit readiness plus failure-safe teardown instead of a fixed sleep.

## Validation

- [x] `pnpm check`
- [x] shared, client, CLI, and skill-evals typechecks
- [x] shared, client, and CLI builds
- [x] SessionRuntime authority/source/admission/retry/custody focused suites: 462/462
- [x] Local Context focused suites and independent immutable-read probe: 116/116
- [x] CodeQL-site provider handlers and architecture guard: 106/106
- [x] turn-end regression: 20 fresh serialized two-file processes (260/260), 20 fresh single-file processes, default-worker pair, and full Client all passed with no unhandled rejection or lock-cleanup leak
- [x] independent read-only code review: APPROVE on exact head `8d0495e8e8d78dfc00d60fe4653ab630cd3a565b`
[...]

PR #2363, opened 2026-08-23
fix(client): admit schema v1 Managed Skills state in publication preflight

liuchao-001:
## Problem

Since #2338, `assertRuntimeConfigCanPublish` in `packages/client/src/runtime/provider-support/preparation.ts` runs before Managed Skills reconcile and throws `ManagedSkillsUnsafeDiscoveryError` for any `managed.json` whose read result is not `kind: "current"`. That includes the schema v1 (`legacy`) state that `loadOrMigrateManagedState` already migrates in place.

Any agent workspace that was last projected before the v2 state shipped can therefore never start a session: every retry fails with

```
Managed Skills state is unreadable or from an unsupported version; preserving the existing projection
```

classified as `transient_transport` / `managed_skills_unsafe_discovery`, so the slot retries forever and the chat shows `Retrying provider · attempt N`.

Observed on staging (`0.5.22-staging.1215.1`): `liuchao-fable` hit this on the first message after 07‑23; four more agents on the same host (`liuchao-codex`, `fire-fable-dev`, `fire-fable-reviewer`, one unnamed) carried the same v1 file and would have failed on their next wake-up.

## Fix

Treat v1 state as a projection published at Team Resource fence `0` (which is what `emptyManagedState()` produces after migration) instead of unsafe. The existing `runtimeConfig === null && !hasCapturedPayload` and version-fence checks still apply; `future` / `invalid` state is still refused.

## Tests

- `prepare-managed-session.test.ts`: new case writes a v1 `managed.json` and asserts the session is admitted and reconcile runs (fails on `main` with the error above); new case writes a `schemaVersion: 3` file and asserts it is still refused before reconcile.
- `pnpm check`, `pnpm typecheck`, `pnpm --filter @first-tree/client test` green locally.

## Operator note

Workspaces already stuck in the retry loop recover on the next retry once this ships. Until then, moving the v1 `.first-tree-workspace/managed.json` aside has the same effect: with `skills: []` (the only shape the v1 writer produced for core-skill-only workspaces) the `missing` and `legacy` migration paths adopt the same symlink-pair-proven targets. That is how the five staging workspaces above were unblocked.

### DIFF
--- a/packages/client/src/__tests__/prepare-managed-session.test.ts
+++ b/packages/client/src/__tests__/prepare-managed-session.test.ts
@@ -3,6 +3,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { SessionContext } from "../runtime/contracts.js";
+import { MANAGED_STATE_REL } from "../runtime/managed-state.js";
 import type { PrepareManagedSessionParams } from "../runtime/provider-support/preparation.js";
 import { INIT_COMPLETE_SENTINEL_REL } from "../runtime/workspace.js";
 import { mockCtxPlumbing } from "./test-helpers.js";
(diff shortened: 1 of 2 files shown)


liuchao-001:
## Independent QA — PASS (tier `focused-local`, bound to head `7652496ae`)

*Posted by **code-reviewer** (First Tree QA/review agent) under @liuchao-001's account with his explicit authorization in chat. This is an independent QA report, **not** a GitHub review or approval — the author cannot sign an independent verdict for his own PR. The engineering verdict below is the reviewer's.*

**Verdict: HEALTHY — recommend merge.** Root cause correctly identified, fix is minimal and semantically right, and the regression coverage genuinely pins the behavior.

### Evidence boundary

Everything below was run at head `7652496ae` against the client package's real module seam and a real filesystem. **No live runtime, provider session, or cross-surface path was exercised** — this report does not claim release readiness beyond the Managed Skills admission/migration path.

### What was verified

| Check | Result |
|---|---|
| `prepare-managed-session` suite | ✅ 27/27 |
| `managed-skills` + `managed-state` + `managed-skills-local-variant` + `source-publication-lock` | ✅ 164/164 |
| `biome check .` (repo-wide) | ✅ exit 0 |
| `@first-tree/client` typecheck | ✅ exit 0 |
| CI at this head | ✅ all green (`Test Client & Web`, `Test Server`, `Test CLI`, `Lint & Type Check`) |

**The regression test really reproduces the outage.** Swapping `preparation.ts` back to its `main` contents makes `admits a schema v1 Managed Skills state so reconcile can migrate it` fail at `preparation.ts:276` with `Managed Skills state is unreadable or from an unsupported version` — character-for-character the error users were seeing.

**The full migration chain was proven with the real incident artifact.** The new preflight tests stub `reconcileManagedSkillsForConfig`, so they establish *"v1 is admitted"* but not *"an admitted v1 actually migrates"*; field recovery went through the `missing` branch, so the `legacy` branch had no live proof either. To close that gap, the byte-exact v1 file recovered from the stuck `liuchao-fable` workspace (`schemaVersion: 1`, `cliVersion: "0.5.17-staging.935.1"`, `updatedAt: "2026-07-23T23:24:32.393Z"`, `skills: []`) was driven through the **unmocked** `reconcileManagedSkills`: state migrated in place to v2, fence raised to the authoritative snapshot's 16, all Core skills installed.

**Mutation testing on the new coverage.**
- Reverting the `legacy → 0` branch to `null` kills **both** v1 tests. ✅
- Deleting the null-config guard entirely kills **exactly** `keeps the null-config guard for a schema v1 Managed Skills state` and nothing else in the suite. Worth noting: the pre-existing null-config coverage in `source-publication-lock.test.ts` builds a *current* v2 projection, so the legacy path was genuinely unpinned before this commit. ✅

### On the fix itself

Fence `0` is the right value rather than a convenient one: `loadOrMigrateManagedState` writes `emptyManagedState()` (= `resourceConfigVersion: 0`) on migration, so the preflight's assumption matches the migration outcome exactly. And v1 has no Team Resource concept at all (`skills` is a bare `string[]`), so "nothing fence-versioned has been published" can only be 0. `future` / `invalid` still refused, version-fence comparison unchanged.

**Context Tree constraint checked.** `agent-behavior-steering.md` constraint 5 requires that config unavailability preserve last-known-good rather than be read as an empty authoritative set. Verified by execution, not by reading: legacy state + `runtimeConfig: null` + no captured payload throws `Agent runtime config is unavailable while a managed projection already exists; preserving last-known-safe bytes` with reconcile never invoked. Because legacy now yields a non-null fence 0, the `highestPublishedVersion === null` early return no longer fires and the guard still engages. No conflict.

### Non-blocking notes

1. **Two sibling guards share this bug's shape but are not reachable.** `verifyManagedSkillsProjectionForAdmission` (`managed-skills.ts:196`) also rejects every non-`current` state, but only on the `existingPayload` in-process hot-switch path — and `existingPayload` implies a prior successful prepare in the same process, which would already have migrated the state. `recoverPendingJournal` (`managed-skills.ts:1601`) hard-fails on legacy state, but journals are a v2-era artifact, so "v1 state + journal file" is not a producible combination. **Correctly left alone in this PR.**
2. **The fence-0 trade-off, stated explicitly.** If a client were downgraded to a pre-v2 CLI that wrote v1 over a previously-v2 workspace, the recorded fence would be lost and one stale `runtimeConfig` could publish, where `main` fails closed. The exposure is a single reconcile (the authoritative snapshot then raises the fence again), and the fence is really lost by the downgrade itself — a pre-v2 client does not honor it in the first place. Against permanently bricked workspaces this is clearly the right trade.

### Scope limits

`PASS` covers the Managed Skills admission and v1→v2 migration path at this head only. Any new commit — including a docs- or test-only one — invalidates this report and requires a re-run at the new head.

yuezengwu (approved):
**Recommendation**: approve
- Rationale: the admission preflight now assigns schema-v1 Managed Skills state the same fence 0 that the reconciler writes during migration, while preserving the null-config guard and continuing to reject future/invalid state.

**Exact head reviewed**: 7652496aebb1ab19a939d57a6e0b6a476f91b979

**Findings**
- No blocking issues.
- The change is narrowly scoped to the pre-reconcile fence calculation. It does not weaken the last-known-good guard when runtime config is unavailable, and existing version comparisons still prevent a stale config from crossing a current v2 fence.
- Reverting the legacy to 0 branch reproduces the reported failure in the new regression test.

**Independent validation**
- prepare-managed-session.test.ts: 27/27
- managed-state.test.ts and managed-skills.test.ts: 119/119
- @first-tree/client typecheck: pass
- Biome on both changed files: pass
- GitHub checks at this exact head: all pass/skipping

**Context Tree alignment**
- This preserves the durable rule that config unavailability is not an empty authoritative snapshot: a legacy projection with no current config still fails closed before reconcile.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: schema-v1 Managed Skills state is assigned the same fence 0 produced by the existing conservative migration, while null-config, stale-version, future-schema, and invalid-state guards remain fail closed.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Author / repo: liuchao-001 / first-tree-ai/first-tree
- Problem: workspaces last projected with schema v1 can be trapped in an endless provider retry loop before reconciliation can migrate them.
- Approach: admit only parsed legacy state as fence 0 so the existing reconciler performs the in-place migration; retain all existing unsafe-state and last-known-good checks.
- Impacted modules: client managed-session preparation and its regression tests.

**Review findings**
- ✅ The fence value matches `emptyManagedState()` and does not duplicate migration logic.
- ✅ The highest fence across target and authority roots still prevents a legacy root from lowering a current v2 fence.
- ✅ Regression coverage pins legacy admission, null-config refusal, and future-schema refusal.
- ✅ Independent QA and all CI checks are green at exact head `7652496aebb1ab19a939d57a6e0b6a476f91b979`.

**Action taken**
- Approved the exact reviewed head.

Issue #2362, opened 2026-08-22
Claude Code session keeps a stale child process after credential failure; re-login never heals the chat


liuchao-001:
## Summary

When a Claude Code session's provider process hits a `credential` failure (e.g. local OAuth expired and refresh failed), the runtime settles the turn with `stop / needs_operator` but keeps the same child process alive. Claude Code caches its auth state in-process, so after the user signs in again (via the in-product Claude login) every subsequent turn on that chat still fails with `Not logged in · Please run /login`. Only brand-new chats (which spawn a new process) recover. The in-chat notice "sign in again and retry" can therefore never succeed on the affected chat.

## Reproduction (observed 2026-08-22 on staging, agent `opentag`, Feishu-bound chat `019ffa61-dca7-7285-a628-e0b26756298d`)

Local-time excerpts from `~/.first-tree-staging/logs/client.log` (client v0.5.22-staging.1215.1):

```
16:01:04 Resuming session (345a5510-…), cwd=…/workspaces/opentag
16:01:08 Claude SDK provider failure (credential/provider_credential_required): authentication_failed
         Failed to authenticate: OAuth session expired and could not be refreshed
16:11:28 Claude SDK provider failure (credential/provider_credential_required): authentication_failed
         Failed to authenticate: OAuth session expired and could not be refreshed
16:11:39 runtime-auth: starting claude auth login (method=browser)
16:12:12 runtime-auth: claude login complete               <-- keychain token refreshed here
16:12:43 Claude SDK provider failure (credential/provider_credential_required): authentication_failed
         Not logged in · Please run /login                  <-- same chat, same child process
16:15:24 [other chat] Session started (41ff1988-…)          <-- new process, works
16:16:04 / 16:16:09 / 16:18:44  same failure on 019ffa61    <-- still the 16:01 process
16:19:11 [other chat] Session started (a88853ad-…)          <-- new process, works
```

- The `claude --resume 345a5510-…` child (pid 1348, started 16:01:04) was still alive at 16:24.
- Env of the stuck process and a healthy one (HOME, PATH, ANTHROPIC_*, CLAUDE_CONFIG_DIR) was identical — this is purely in-process auth state, not a config-path difference.
- Side effect: while stuck, the slot logged `session idle threshold reached but work is still in flight — skipping suspend` every 10s, so idle eviction never reclaimed the process either.

[...]

PR #2364, opened 2026-08-23
fix(client): retire Claude query after credential failure

yuezengwu:
## Summary

- settle Claude credential failures as terminal `needs_operator` turns, publish the runtime failure event before the exact consumed-prefix settlement, and retire the failed SDK query/native generation
- preserve the logical chat and persisted Claude session id, then lazily create one fresh exact-resume query when a newly delivered message arrives
- project the latest non-model managed config before that recovery query, repeating projection if the cache advances so the applied snapshot cannot outrun the generated `CLAUDE.md`
- keep the settled failed input out of replay while returning only unentered tail messages through the existing FIFO recovery path
- tell the operator to sign in and send a message again; login completion itself does not restart affected chats
- keep non-credential terminal handling and transient automatic retry/replay paths unchanged

Relates to #2362. This PR intentionally does not auto-close the issue.

## Context Tree coordination

- Draft Tree PR: https://github.com/first-tree-ai/first-tree-context/pull/974
- The Tree PR records the durable generation-retirement and lazy exact-resume boundary and remains draft under the code-first flow.

## Verification

- `pnpm --filter @first-tree/client exec vitest run src/providers/claude/__tests__/provider-error-result.test.ts src/__tests__/runtime-notice.test.ts` — 22 tests passed
- adjacent Claude/retry/runtime-notice suite — 93 tests passed
- `pnpm --filter @first-tree/client typecheck` — passed
- `pnpm --filter @first-tree/client test` — 216 files passed, 2 skipped; 2,840 tests passed, 7 skipped
- `pnpm check` — passed with no errors (existing repository warnings/info only)
- `pnpm typecheck` — 9/9 tasks passed
- `pnpm test` — 10/10 tasks passed
- `git diff --check` — passed

## QA self-check

- Status: PASS
- Tier: `test-only`
- Scope: Claude handler lifecycle at the SDK query seam, exact terminal settlement ordering, lazy same-session resume, latest-config projection before recovery, failed-input exclusion, queued-tail FIFO, stale-query closure, duplicate-query prevention, and runtime notice wording
- Case disposition: `move-to-product-test` — the deterministic regressions are committed in the provider lifecycle and runtime-notice product suites
- Limitations: no live Anthropic login was performed; credential state and SDK/native lifecycle are controlled at the real handler seam

No database schema, migration, or persistence-semantic changes are included.

### DIFF
--- a/packages/client/src/__tests__/runtime-notice.test.ts
+++ b/packages/client/src/__tests__/runtime-notice.test.ts
@@ -102,9 +102,12 @@ describe("runtime notice formatting", () => {
         }),
       ),
     ).toContain("usually NOT a login problem");
-    expect(formatProviderFailureRuntimeNotice(payload({ provider: "claude-code", category: "credential" }))).toContain(
-      "Run `claude auth login`",
+    const credentialNotice = formatProviderFailureRuntimeNotice(
+      payload({ provider: "claude-code", category: "credential" }),
     );
+    expect(credentialNotice).toContain("Run `claude auth login`");
+    expect(credentialNotice).toContain("send your message again in this chat");
+    expect(credentialNotice).not.toContain("then retry");
     expect(
       formatProviderFailureRuntimeNotice(
         payload({ provider: "claude-code", category: "provider_capacity", reasonCode: "provider_billing_limit" }),
--- a/packages/client/src/providers/claude/__tests__/provider-error-result.test.ts
+++ b/packages/client/src/providers/claude/__tests__/provider-error-result.test.ts
@@ -1,4 +1,4 @@
-import { mkdtempSync, realpathSync, rmSync } from "node:fs";
+import { mkdtempSync, readFileSync, realpathSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { type ProviderRetryEventPayload, parseProviderRetryEventMessage, type SessionEvent } from "@first-tree/shared";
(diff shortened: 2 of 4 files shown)


baixiaohang (changes requested):
- Rationale: The credential-generation retirement is directionally correct, but lazy recovery can permanently skip a configuration briefing update made while the failed query is retired.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: After Claude credentials expire, signing in again does not heal the affected chat because the existing native process keeps stale authentication state.
- Approach: Settle the failed turn, close that provider generation, preserve the logical chat and Claude session identity, and lazily create an exact-resume query for the next newly delivered message without replaying the failed input.
- Impacted modules: Claude SDK handler lifecycle, inbox-tail recovery, and runtime failure notice wording.

**Review findings**
❌ 1. Lazy credential recovery marks a newer config as applied without publishing its briefing. [`packages/client/src/providers/claude/index.ts:1011`]

`spawnQuery()` reads the latest cache payload and immediately calls `recordAppliedPayload()`, but this new recovery branch bypasses the `projectManagedWorkspace()` work performed by `maybeSwitchConfig()` for non-model config changes. If the agent prompt or projected resources change while the query is retired (for example, while the operator completes login), the resumed query still reads the old project `CLAUDE.md`; then the first recovered input calls `maybeSwitchConfig()`, sees the new version already recorded as applied, and never repairs that stale briefing. Please prepare/project the latest config before the one recovery spawn (or otherwise preserve the old applied version until projection succeeds) and cover a config update during the credential-wait window.

✅ The terminal provider event is captured before `DeliveryToken.complete()`, the failed query is retired synchronously, and the added lifecycle test covers same-session exact resume, failed-input exclusion, tail FIFO, and duplicate-query prevention for the unchanged-config case.

**Action taken**
- Submitted request changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head resolves the credential-wait configuration gap while preserving the terminal settlement and lazy exact-resume lifecycle.

**Risk level**: A
- Semantic lift: none

**PR summary**
- Problem: After Claude credentials expire, signing in again does not heal an affected chat because its existing native query retains stale authentication state.
- Approach: Settle the failed turn as operator-required, retire only that provider generation, preserve the logical chat and Claude session identity, and create one fresh exact-resume query when a newly delivered message arrives. Before that spawn, project any newer managed non-model configuration and repeat if the cache advances, so the resumed process and workspace briefing share one current snapshot.
- Impacted modules: Claude SDK handler lifecycle, managed workspace projection, inbox-tail recovery, and runtime failure notice wording.

**Review findings**
✅ 1. The prior blocker is resolved: credential recovery projects the latest non-model payload before `spawnQuery()`, keeps the applied snapshot unchanged until both projection and query construction succeed, and retries projection if the cache advances.

✅ 2. Lifecycle guards still prevent a stale or duplicate recovery spawn, while the focused regression covers same-session resume, briefing refresh and re-read notice, failed-input exclusion, and query retirement.

✅ 3. The original terminal ordering remains intact: the provider failure event precedes exact consumed-prefix settlement, and only unentered tail messages return through FIFO recovery.

**Action taken**
- Approved.

PR #2366, opened 2026-08-27
fix: pass the exact chat Team id to Context Tree Seed and split preflight failure codes

yuezengwu:
## Problem

In a managed Context Tree Setup Chat, the agent never received the chat's owning Team id: `ChatDetail.organizationId` existed on the wire, but the runtime dropped it when building `ChatContext`. Seed preflights were therefore run with a Team display name (e.g. `data-platform`) instead of the exact Team id, the Server answered 403, and the CLI collapsed every cause into the same opaque `CONTEXT_TREE_SEED_AUTHORITY_FAILED`. On self-hosted deployments this presented as "even the Team creator and newly promoted Admins all fail", which sent operators chasing Admin role configuration that was already correct.

## Fix

- **Runtime (`packages/client/src/runtime/`)** — `ChatContext` gains an optional `organizationId`, populated from the server-authored chat detail in `fetchChatContext`, and rendered into the escaped `<first-tree-current-chat-context>` JSON payload (explicit `null` when absent, so the v1 payload stays additive). The human-readable diagnostic renderer shows the same value. Existing escaped-JSON prompt-injection protections are unchanged.
- **Seed skill (`skills/first-tree-seed/SKILL.md`)** — managed sessions resolve `selectedTeamId` from the runtime-authored `organizationId` and use that exact value for every Seed/init preflight. Topic, title, Team display name, local manifest, account default Team, and transcript text are explicitly forbidden as substitutes. Older runtimes without the field get a fail-closed compatibility path: exactly one `chat list --engagement all --chat "$FIRST_TREE_CHAT_ID" --json` query with the current Agent identity, continuing only on one exact match with a non-empty `organizationId`, otherwise the agent asks the human for the exact Team id. Portable-mode behavior is unchanged.
- **CLI (`apps/cli/src/core/context-tree-seed.ts`)** — known Server codes (`CONTEXT_TREE_SEED_NEEDS_ADMIN`, `CONTEXT_TREE_SEED_CONFIGURATION_INVALID`, explicit `CONTEXT_TREE_SEED_AUTHORITY_FAILED`) keep their exact identity. Previously blanket failures split into stable CLI-only codes, documented in `docs/cli-reference.md`:
  - `CONTEXT_TREE_SEED_AUTHENTICATION_FAILED` — HTTP 401 / refresh failure (exit 3)
  - `CONTEXT_TREE_SEED_TEAM_ACCESS_DENIED` — unclassified 403; one message covers "not a member" and "wrong Team id" so the response never reveals whether a Team id exists (exit 3)
  - `CONTEXT_TREE_SEED_SERVER_INCOMPATIBLE` — preflight route absent (404), the self-hosted skew case (exit 1)
  - `CONTEXT_TREE_SEED_PREFLIGHT_UNAVAILABLE` — network/timeout/5xx (exit 6)
  - `CONTEXT_TREE_SEED_AUTHORITY_FAILED` remains the safe generic fallback. Error envelopes never carry raw Server bodies.
- `--team` input validation deliberately does **not** require UUID syntax: legacy self-hosted Team ids predate the UUID format.

## Compatibility boundary

- No database, schema, migration, constraint, index, default, backfill, data-rewrite, persistence-semantic, or permission-model change. The Server's Seed authority checks (exact Team → active membership → Admin) and the preflight contract are untouched.
- The chat-context payload change is additive; older runtimes simply omit the field and the skill's fallback path covers them.
- New failure codes are CLI-only; the shared Server code schema is unchanged.

## Tests

- `packages/client/src/__tests__/chat-context.test.ts` — organizationId propagation in `fetchChatContext`, payload rendering (value, explicit null, escaping of a forged closing tag inside the id), and the diagnostic renderer line.
- `apps/cli/src/__tests__/context-tree-seed-core.test.ts` — every new mapping (401, refresh failure, unclassified 403 without body/existence leakage, 404, 5xx, timeout), preservation of known Server codes, no second authority request, and the generic fallback.
- `packages/skill-evals` floor suite — new no-model invariants pin the skill's managed Team-id resolution and fail-closed fallback.
- `packages/qa/cases/cross-surface/context-tree-seed-managed-team-selection.md` — live cross-surface QA case for the managed Setup Chat flow and the CLI failure ladder against a self-hosted Server.

Verification: `pnpm check` ✅, `pnpm typecheck` ✅ (client, cli, skill-evals), `@first-tree/client` suite 2842 ✅, `@first-tree/skill-evals` suite 749 ✅ + `eval:floor` 8/8 ✅, CLI seed core tests 13/13 ✅. The CLI full suite shows 17 failures in `update-detect-install-mode`, `context-byo-repository`, `context-project-resolver`, and `tree-init` environment-dependent tests; verified identical on the pristine base commit (environment-limited local run, unrelated to this change).

## QA

This change touches Runtime prompt injection plus CLI/skill behavior, so it ships with a focused non-executable prose case at `packages/qa/cases/cross-surface/context-tree-seed-managed-team-selection.md`. **Live QA needed after merge readiness**: a `focused-local` run of that case against a real self-hosted Server (managed Setup Chat Team-id sourcing on the exact head, older-runtime fallback, and the CLI failure ladder with two Teams of different roles). Deterministic payload/error/floor behavior is already covered by the product tests above; the case exists for the real runtime injection and self-hosted boundary that tests cannot reach.

### DIFF
--- a/apps/cli/src/__tests__/context-tree-seed-core.test.ts
+++ b/apps/cli/src/__tests__/context-tree-seed-core.test.ts
@@ -1,5 +1,6 @@
 import { SdkError } from "@first-tree/client";
 import { describe, expect, it, vi } from "vitest";
+import { AuthRefreshFailedError } from "../core/bootstrap.js";
 import { type ContextTreeSeedAuthorityReader, preflightContextTreeSeed } from "../core/context-tree-seed.js";
 
 function authorityReader(result: unknown): {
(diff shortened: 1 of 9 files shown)


baixiaohang (changes requested):
- Rationale: The product path and failure mapping are coherent, but the new live QA case contradicts the Server authority contract and cannot be executed as written.

**Risk level**: A

**PR summary**
- Problem: Managed Context Tree Setup agents can mistake a Team display name for its exact id, causing valid admins to receive an opaque 403 and misleading operators toward role troubleshooting.
- Approach: Inject the chat's server-authored Team id into current-chat context, make Seed consume only that exact id with a fail-closed older-runtime fallback, and classify previously opaque CLI preflight failures without exposing Server bodies.
- Impacted modules: client runtime chat context, Seed skill/evals, CLI Seed preflight, CLI docs, cross-surface QA

**Review findings**
❌ 1. Make the Team B and unclassified-403 steps executable and mutually consistent. The preconditions leave the Team B operator as an ordinary member, but the managed Setup Chat endpoint is admin-only, so that actor cannot create the “Team B Setup Chat” required at line 53. The case then requires the same ordinary-member condition to return both the Server-classified `CONTEXT_TREE_SEED_NEEDS_ADMIN` (lines 53–56) and the CLI-only `CONTEXT_TREE_SEED_TEAM_ACCESS_DENIED` (lines 59–60). On the current Server, an active ordinary member returns the former; the latter is for an unclassified 403 such as a non-member or wrong Team id. Please define a realizable setup (for example, create the chat while admin and demote before the run, or test the ordinary-member preflight directly) and use a non-member/wrong-id request for the unclassified-403 ladder step. [`packages/qa/cases/cross-surface/context-tree-seed-managed-team-selection.md:53`]

**Action taken**
- Submitted request changes.

yuezengwu:
Fixed in 9642035: the Team B actor now starts as Admin (managed Setup Chat creation is admin-only) and is demoted before the run, with NEEDS_ADMIN asserted through the direct preflight; the unclassified-403 ladder step now uses a wrong Team id or a Team the signed-in member does not belong to, and the case explicitly bars reusing the ordinary-member condition for it.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The runtime now carries the exact server-authored Team id into the managed prompt, Seed consumes that single authority with a fail-closed compatibility path, and CLI preflight failures are classified without leaking Server bodies; the prior QA blocker is resolved.

**Risk level**: A

**PR summary**
- Problem: Managed Context Tree Setup agents could substitute a Team display name for its exact id, causing valid admins to receive an opaque 403 and misleading operators toward role troubleshooting.
- Approach: Propagate `ChatDetail.organizationId` through runtime current-chat context, require Seed to use that exact value (with one fail-closed exact-chat lookup for older runtimes), and split previously opaque CLI failure classes while preserving known Server codes.
- Impacted modules: client runtime chat context, Seed skill/evals, CLI Seed preflight, CLI docs, cross-surface QA

**Review findings**
✅ 1. The exact chat Team id now has one server-authored path from chat detail to the escaped prompt payload, including explicit `null` compatibility and closing-tag escaping.
✅ 2. Seed rejects names/defaults/manifests/transcript text as authority and uses one exact-chat fallback query only when the runtime field is unavailable.
✅ 3. Known Server codes retain identity; unclassified authentication, access, compatibility, and availability failures receive stable CLI-only codes with safe messages and no raw-body leakage.
✅ 4. The revised QA case is executable: Team B's chat is created while Admin, the operator is demoted before the ordinary-member check, and the unclassified-403 step uses a non-member or wrong Team id rather than reusing the `NEEDS_ADMIN` condition.

**Action taken**
- Approved current head `964203509`.

PR #2373, opened 2026-08-31
fix(server): encode Unicode attachment filenames in response headers

roytrack:
# PR: Encode Unicode attachment filenames in response headers

## Suggested title

`fix(server): encode Unicode attachment filenames in response headers`

## Summary

- Percent-encode the complete attachment filename before placing it in the
  `Content-Disposition` response header.
- Preserve the existing quoted `filename="..."` wire shape and the current
  client-side `decodeURIComponent` behavior.
- Add an upload-to-download regression test covering Unicode, spaces, a
  literal percent sign, and full-width punctuation.
- Validate the generated header with Node's own `validateHeaderValue` API.

## Root cause

The attachment download route only percent-encoded CR, LF, quotes, and
backslashes. A filename containing characters outside Latin-1 was therefore
passed directly to Node's HTTP response implementation. Node rejects that
header value with `ERR_INVALID_CHAR` while Fastify is sending the attachment
stream, which can terminate the server process.

The upload path already transports filenames with `encodeURIComponent`, and
the client already decodes percent-encoded `filename="..."` values. Applying
the same encoding to the complete download filename fixes the invalid header
without changing the wire contract or requiring a client upgrade.

## Scope

The issue is not specific to attachment size or MIME type. It affects any
consumer of `GET /api/v1/attachments/:id` when the stored filename contains
non-ASCII characters. That includes chat files and images as well as other
features backed by the shared attachment store, such as Team Skill bundles.

## Verification

- `biome check` for the two changed files: passed.
- `tsc --noEmit -p packages/server/tsconfig.json`: passed.
- Server `tsdown` build: passed.
- `git diff --check`: passed.
- Node header reproduction: old implementation raises `ERR_INVALID_CHAR`;
  patched implementation is ASCII-only and round-trips the original filename.
- The equivalent built Server bundle was exercised against an isolated
  database clone: multiple Unicode filenames and an ASCII control file all
  returned HTTP 200 with matching decoded names, byte lengths, and SHA-256
  content hashes, with zero Server restarts or severe errors.
- The focused Vitest route test is included but was not executed locally
  because no Docker/PostgreSQL test runtime was available. It should run in CI.

### DIFF
--- a/packages/server/src/__tests__/attachments-route.test.ts
+++ b/packages/server/src/__tests__/attachments-route.test.ts
@@ -1,3 +1,4 @@
+import { validateHeaderValue } from "node:http";
 import { Readable } from "node:stream";
 import { ATTACHMENT_FILENAME_HEADER, ATTACHMENT_MIME_HEADER, MAX_ATTACHMENT_BYTES } from "@first-tree/shared";
 import { eq, sql } from "drizzle-orm";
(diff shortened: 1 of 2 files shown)


roytrack:
Superseded by a clean branch whose commit author is linked to the contributing GitHub account; the code diff is unchanged.

PR #2374, opened 2026-08-31
fix(server): encode Unicode attachment filenames in response headers

roytrack:
# PR: Encode Unicode attachment filenames in response headers

## Suggested title

`fix(server): encode Unicode attachment filenames in response headers`

## Summary

- Percent-encode the complete attachment filename before placing it in the
  `Content-Disposition` response header.
- Preserve the existing quoted `filename="..."` wire shape and the current
  client-side `decodeURIComponent` behavior.
- Bound canonical filenames to 255 UTF-8 bytes at the shared Client/Web and
  Server ingress contract, keeping worst-case percent-encoded values well
  below Node's default 16 KiB header limit.
- Normalize Client SDK and Web filenames once, then validate and encode that
  same trimmed value so transport cannot reintroduce unmeasured padding.
- Truncate Feishu resource filenames on UTF-8 code-point boundaries before
  the Server attachment trust boundary.
- Truncate legacy overlong database values on UTF-8 code-point boundaries at
  response time so pre-limit rows cannot overflow clients.
- Add an upload-to-download regression test covering Unicode, spaces, a
  literal percent sign, and full-width punctuation.
- Add real Node `fetch` boundary regressions for the maximum accepted filename
  and a historical 5,500-character filename.

## Root cause

The attachment download route only percent-encoded CR, LF, quotes, and
backslashes. A filename containing characters outside Latin-1 was therefore
passed directly to Node's HTTP response implementation. Node rejects that
header value with `ERR_INVALID_CHAR` while Fastify is sending the attachment
stream, which can terminate the server process.

Web uploads already transported filenames with `encodeURIComponent`, while
the Client SDK sent the raw value. Both paths now validate and percent-encode
the same canonical trimmed filename before transport; the Server decodes it
and applies the same shared validation before storage. The client already
decodes percent-encoded `filename="..."` values, so the download wire shape
remains compatible.

Complete percent-encoding can expand accepted ASCII punctuation by 3x. Without
an ingress bound, a long stored name can exceed Node's 16 KiB response-header
limit. The shared 255-byte UTF-8 limit caps the filename parameter at 765
characters in the worst case, and the response-time legacy bound covers rows
created before this validation existed. Feishu ingress uses the same UTF-8
truncation helper so long Unicode resource names remain available instead of
being rejected by the service-wide limit.

## Scope

The issue is not specific to attachment size or MIME type. It affects any
consumer of `GET /api/v1/attachments/:id` when the stored filename contains
non-ASCII characters. That includes chat files and images as well as other
features backed by the shared attachment store, such as Team Skill bundles.

## Verification

- `biome check` for all ten changed files: passed.
- Shared, Server, Client, and Web typechecks: passed.
- Shared 2/2, Client 15/15, Web 10/10, Server attachment route 29/29, and
  Feishu resource hydration 4/4 focused tests: passed.
- Shared, Server, Client, and Web production builds: passed.
- `git diff --check`: passed.
- Node header reproduction: old implementation raises `ERR_INVALID_CHAR`;
  patched implementation is ASCII-only and round-trips the original filename.
- The equivalent built Server bundle was exercised against an isolated
  database clone: multiple Unicode filenames and an ASCII control file all
  returned HTTP 200 with matching decoded names, byte lengths, and SHA-256
  content hashes, with zero Server restarts or severe errors.
- Server product regressions cover Unicode round-trip, the 255-byte boundary
  through real Node `fetch`, legacy overlong-row response bounding, padded
  Client SDK upload through a real Server, ingress rejection, and Feishu
  Unicode truncation. Server tests ran against an isolated PostgreSQL clone.


roytrack:

yuezengwu (dismissed):
Approved at exact head `3fcdc3012b1a362205118d956c4628689d6e0f2a`.

I reviewed the complete PR context, diff, upload/download contract, and current comments/checks. The change keeps the existing quoted `filename="..."` wire shape while encoding the stored semantic filename to ASCII. Unicode, spaces, literal `%`, quotes, backslashes, slashes, CR/LF, and control characters cannot escape the quoted header; the existing client performs one `decodeURIComponent`, so names round-trip without double encoding.

Exact-head verification:
- `attachments-route.test.ts`: 25/25
- client `sdk-comprehensive.test.ts`: 13/13
- full Server suite: 3488/3489; the unrelated rendezvous-bound test passed 10/10 in isolation
- `pnpm check`, `pnpm typecheck`, `pnpm build`, and `git diff --check`: pass
- mutation check: the new regression test fails on base with Node/Fastify `ERR_INVALID_CHAR` and passes on this head
- CLA passed; no unresolved review threads

No blocking correctness, header-injection, compatibility, identity, CLA, copyright, or license finding.

roytrack:
https://github.com/first-tree-ai/first-tree/issues/2291

yuezengwu:
## Blocking local validation finding: `Content-Disposition` expansion can exceed Node's header limit

Please continue revising this PR before merge. This finding applies to exact head `3fcdc3012b1a362205118d956c4628689d6e0f2a`.

The new `encodeURIComponent(name)` behavior can expand an accepted filename by 3x. The attachment input and database schema currently have no filename-length bound, while the official Client SDK sends `opts.filename` unchanged in `x-attachment-filename`.

I reproduced a regression with a filename consisting of 5,500 semicolons:

- SDK-style raw upload header: 5,500 bytes, accepted.
- Base response header: 5,519 bytes; Node `fetch` succeeds.
- PR response header: 16,519 bytes; Node's default `http.maxHeaderSize` is 16,384 bytes.
- PR download result: `fetch failed`, cause `UND_ERR_HEADERS_OVERFLOW`.

This is attributable to the PR: the base encoder leaves semicolons unchanged inside the quoted value, while this head encodes each one as `%3B`.

Before re-review, please:

1. Ensure every accepted/stored filename produces a response header within the supported client limit, or reject/normalize it at a defined ingress boundary.
2. Add a deterministic product regression covering accepted upload -> download through Node `fetch` around the header-size boundary.
3. Reconcile the PR description with the actual upload paths: Web percent-encodes filenames, but `packages/client/src/cloud/sdk.ts` currently sends them unchanged. Please also update the now-stale client decoder comment that says only CR, LF, quote, and backslash are encoded.

The intended Unicode fix itself is valid: raw Unicode reproduces `ERR_INVALID_CHAR`, and Unicode/control/injection-sensitive character round-trips are ASCII-safe on this head. Local targeted tests (server 25/25, client 13/13), full Server tests (3,489/3,489), check, typecheck, and build all passed, but none covers the overflow boundary above.

Please request a fresh exact-head validation after pushing the revision; prior findings and test results do not transfer to a new head.

roytrack:
Addressed the blocking header-overflow finding in commit `b403412c39ffab3b532e56a04e1206f8700ec38f` without rewriting the existing PR history.

Changes:
- Added a shared 255 UTF-8 byte filename limit used by Client SDK, Web, and the Server `createAttachment` trust boundary.
- Client SDK now percent-encodes upload filenames, matching Web and Server decode behavior.
- Bounded legacy overlong database filenames on UTF-8 code-point boundaries before response-header encoding.
- Added real Node `fetch` regressions at the accepted boundary and for a legacy 5,500-character value, plus HTTP/service/client/web rejection coverage.
- Updated the stale SDK decoder comment and corrected the PR upload-path description.

Local validation: changed-file Biome, four package typechecks, Shared 2/2, Client 14/14, Web 9/9, and Shared/Server/Client/Web builds passed. Please run a fresh exact-head validation; GitHub CI is the authoritative Server suite because the local Server global setup was blocked by the workstation pnpm dependency self-repair loop.

packages/client/src/cloud/sdk.ts
```diff
@@ -1045,20 +1046,24 @@ export class FirstTreeHubSDK {
...
+    const filenameError = getAttachmentFilenameError(opts.filename);
+    if (filenameError) throw new TypeError(filenameError);
     const json = await this.requestJson<unknown>(`/api/v1/orgs/${encodeURIComponent(opts.orgId)}/attachments`, {
       method: "POST",
       body: opts.bytes,
       headers: {
         "Content-Type": "application/octet-stream",
         [ATTACHMENT_MIME_HEADER]: opts.mimeType,
-        [ATTACHMENT_FILENAME_HEADER]: opts.filename,
+        [ATTACHMENT_FILENAME_HEADER]: encodeURIComponent(opts.filename),
```
yuezengwu: [P1] Send the filename that was actually validated

`getAttachmentFilenameError` measures `opts.filename.trim()`, but this line percent-encodes `opts.filename` itself. With `" ".repeat(6000) + "x"`, validation sees 1 byte while the wire value is 18,001 characters; a real default Node HTTP server returned 200 for the base raw-header path and 431 for this head. Please encode a canonical trimmed value, apply the same fix at `packages/web/src/api/attachments.ts:45`, and add a regression proving the padded request reaches the Server.

packages/server/src/services/attachment.ts
```diff
@@ -158,9 +162,8 @@ function validateCreateInput(input: CreateAttachmentInput): void {
   if (input.mimeType.trim().length === 0) {
     throw new BadRequestError("Attachment mime type is required");
   }
-  if (input.filename.trim().length === 0) {
-    throw new BadRequestError("Attachment filename is required");
-  }
+  const filenameError = getAttachmentFilenameError(input.filename);
+  if (filenameError) throw new BadRequestError(filenameError);
```
yuezengwu, line 166: [P1] Byte-bound Feishu names before this trust boundary

Applying the 255-byte validator here changes the existing Feishu path without updating its sanitizer: `chooseFilename` still slices to 255 JavaScript characters, so 86 Chinese characters survive there unchanged (258 UTF-8 bytes), then this throws and hydration is reported unavailable/`too_large`. Please truncate/normalize the Feishu ingress on a UTF-8 code-point boundary before it calls `createAttachment`, with regression coverage.

yuezengwu (changes requested):
Re-reviewed exact head `b403412c39ffab3b532e56a04e1206f8700ec38f` from scratch. The previous 5,500-semicolon download-header overflow is fixed, and all changed-package suites plus check/typecheck/build pass locally, but two in-scope regressions still need changes:

1. Client and Web validate `filename.trim()` but percent-encode the original filename. A name with 6,000 leading spaces plus `x` passes the one-byte validation, then expands to an 18,001-character header. Against a real default Node HTTP server, the base raw-header path returns 200 while this head's encoded path returns 431. Please encode the same canonical trimmed value that was validated in both upload clients and cover the boundary-padding case.
2. The new service-wide 255-byte limit makes Feishu files with long Unicode names unavailable because `chooseFilename` still truncates to 255 characters rather than 255 UTF-8 bytes. For example, 86 Chinese characters survive the sanitizer but total 258 bytes and are rejected by `createAttachment`. Please normalize/truncate the Feishu ingress on a UTF-8 code-point boundary and add coverage.

Local validation: `git diff --check`; Shared 955/955, Client 2,845/2,845 (7 skipped), Web 2,427/2,427, Server 3,492/3,492; `pnpm check`; `pnpm typecheck`; and `pnpm build` all passed. The review worktree remained clean.

packages/client/src/cloud/sdk.ts
roytrack: Addressed in `48ecc4b2f0a8ecc2676cf61193bc4543fff88138`. Client SDK and Web now normalize once, validate that canonical trimmed filename, and percent-encode the same value. Added Client/Web unit coverage plus a real `FirstTreeHubSDK.uploadAttachment()` -> Node HTTP -> Server regression with 6,000 leading spaces; the Server stores `padded.txt` and no oversized request header is sent.

packages/server/src/services/attachment.ts
roytrack: Addressed in `48ecc4b2f0a8ecc2676cf61193bc4543fff88138`. Added shared UTF-8 code-point-safe truncation and applied it in Feishu `chooseFilename` before `createAttachment`. The Feishu hydration regression uses 86 Chinese characters (258 bytes), verifies successful hydration, and confirms the stored filename is exactly 255 UTF-8 bytes. The same helper now also bounds legacy download response filenames.

roytrack:
@yuezengwu Both requested P1 changes are addressed at exact head `48ecc4b2f0a8ecc2676cf61193bc4543fff88138` and the two inline threads are resolved.

Fresh local validation:
- padded Client SDK -> real Node HTTP -> Server upload regression: pass
- attachment route suite: 29/29
- Feishu hydration suite: 4/4, including 86 Chinese characters truncated to exactly 255 UTF-8 bytes
- Shared: 2/2; Client: 15/15; Web: 10/10
- changed-file Biome, four package typechecks, git diff check, and Shared/Server/Client/Web builds: pass

Please re-review this exact head after the new CI run completes.

packages/shared/src/schemas/attachment.ts
```diff
@@ -11,6 +11,51 @@ import { z } from "zod";
...
+  }
+  return result;
+}
+
+/** Return the canonical filename validation error shared by every ingress. */
+export function getAttachmentFilenameError(value: string): string | null {
+  const filename = normalizeAttachmentFilename(value);
+  if (filename.length === 0) return "Attachment filename is required";
+  try {
+    encodeURIComponent(filename);
```
yuezengwu, line 49: [Non-blocking follow-up] Preserve Unicode validation in the built package

The build-output discrepancy remains real: after `pnpm build`, `getAttachmentFilenameError("\ud800")` from the built Shared package returns `null`, because the unused `encodeURIComponent` call and its catch branch are removed. Source-only tests do not catch this.

After checking the reachable paths, this is an accepted low-risk follow-up rather than a merge blocker. Normal Chinese/emoji filenames are unaffected. Client/Web still encode before transport (the built SDK rejects malformed input with URIError and makes no HTTP request), and the HTTP decoder does not turn malformed percent-encoded UTF-8 into lone surrogates. A malformed name entering through Feishu/internal calls may instead have its invalid character replaced during database-driver UTF-8 encoding, losing filename fidelity. No authorization bypass, injection, file-content corruption, or server crash has been demonstrated.

Please preserve this validation in optimized output and add built-artifact coverage for lone high/low surrogates as follow-up work. Resolving this review thread records risk acceptance, not a claim that the defect is fixed.

yuezengwu (changes requested):
Fresh review of exact head `48ecc4b2f0a8ecc2676cf61193bc4543fff88138`: both previous findings are fixed. The padded SDK -> real Node HTTP -> Server upload and the 86-Chinese-character Feishu hydration regressions pass. The maximum/legacy filename download regressions also pass.

However, one build-output defect still needs a change: **the Unicode-validation try/catch is removed from the shipped Shared/Client output**. After a successful `pnpm build`, importing `getAttachmentFilenameError` from `packages/shared/dist/index.mjs` and calling it with `"\ud800"` returns `null`, while importing the source returns the expected validation error. The generated function contains no `encodeURIComponent` call or catch branch. The built Server uses this shared export; a database-stub probe confirms that malformed Unicode reaches `createAttachment`'s database boundary instead of being rejected by validation.

Please make this validity check survive optimization and add a built-artifact regression for lone high/low surrogates (in addition to source tests). This is a demonstrated validation-contract defect; I am not claiming a reproduced remote crash.

Validation on this head:
- Shared 955 passed; Client 2,846 passed (7 skipped); Web 2,428 passed.
- Server after rebuild: 3,493/3,494 passed. The unchanged `clients-provider-models` rendezvous-bound test failed once; focused rerun passed 43/43 (attachments 29, Feishu 4, provider models 10).
- `pnpm check`, `pnpm typecheck`, `pnpm build`, and `git diff --check` passed.
- 8,000 UTF-8 truncation cases and three built-SDK/default-Node-HTTP transport probes passed. The built-artifact Unicode rejection assertion fails deterministically.

Non-blocking test-harness note: the new `@first-tree/client/cloud` test import has no source alias, so direct Server package tests can consume an old SDK `dist` directory. Rebuilding fixed the initial stale-artifact 431. The root/Turbo path already builds dependencies; adding the exact source alias would also make direct package runs reliable.

CI is still `action_required` (only CLA has passed); I did not authorize workflows, modify the contributor's source, or merge anything. Please request another exact-head review after the validation fix.

yuezengwu (approved):
Approved at exact head `48ecc4b2f0a8ecc2676cf61193bc4543fff88138`. This supersedes my earlier request-changes verdict.

The previous padded-filename transport and Feishu UTF-8 truncation regressions are fixed and their tests pass. Following the maintainer's explicit decision and a closer review of the reachable paths, the built Unicode-validator discrepancy is retained as a non-blocking, accepted-risk follow-up in the inline comment; it is not considered fixed.

Normal valid Unicode filenames are unaffected. The built Client rejects malformed input before transport, and ordinary HTTP upload decoding cannot introduce a lone surrogate. Residual risk is primarily malformed filename fidelity on internal/integration ingress, plus source/build validation inconsistency; no reproduced security exploit or file-content corruption has been found.

Local validation and the full-suite caveats are recorded in my preceding review. Required CI is still running and must pass independently. This approval does not merge the PR or bypass CI.

yuezengwu:
@roytrack, please fix the current CI failure on head `48ecc4b2f0a8ecc2676cf61193bc4543fff88138` before this PR is merged.

The failing [`Lint & Type Check` job](https://github.com/first-tree-ai/first-tree/actions/runs/33637803084/job/100510312089) reports:

```text
packages/client/src/providers/antigravity/__tests__/handler.test.ts(84,3): error TS2322
Property `noteTurnStart` is missing from the mocked `SessionContext`.
```

Please update that test fixture to satisfy the current `SessionContext` contract, run `pnpm --filter @first-tree/client typecheck` (and the repository CI checks), then push the fix so the required CI is green. The existing approval is the code-review decision for this exact head; it does not waive the failing CI gate.

roytrack:
@bestony @yuezengwu The latest CI workflow is currently awaiting maintainer approval. Could you please approve and run the workflow when you get a chance? Thanks!

bestony:
Approved the pending CI workflow on exact head `d0e78d5cd9458102ca919eaf029965c6e4f46aa4`. Required checks are now running: https://github.com/first-tree-ai/first-tree/actions/runs/34304067670

roytrack:
@bestony @yuezengwu I pushed `9b70a0ffbce71015e9d88037ce74a686b18f65e9` to fix the current Server CI failure.

The failing lost-wake fallback test used an 80 ms timer, so under full-suite database load the route could read the durable rendezvous before the test write committed and return 504. The test now commits the durable result first, then deterministically expires the pending reply to exercise the same fallback path without depending on runner timing.

Local checks passed: `pnpm check`, `pnpm --filter @first-tree/server typecheck`, and `git diff --check`. The new CI run is awaiting maintainer approval: https://github.com/first-tree-ai/first-tree/actions/runs/34852132370

PR #2378, opened 2026-09-01
fix(runtime): report working for a provider turn the inbox does not own

liuchao-001:
## What was wrong

An agent that was actively working showed as **Idle** on every chat surface.

Per-chat runtime state was projected from one thing only — whether an inbox delivery was currently being processed:

```ts
// session-projection-authority.ts
return this.deps.hasProcessingOwnedWork(chatId) ? "working" : "idle";
```

That equates "a turn is running" with "a message is being handled". A provider that re-invokes itself when a background task completes runs a full turn — tool calls, output, token spend — with **no owned inbox entry**. The client reported `idle` for that entire turn, the server's `computeWorking` therefore derived `working: false` → `main: "ready"`, and the sidebar, the compose bar and the chat row all rendered "Idle" while the agent ran tools. Because the composite carries `activity: working ? activity : null`, the "Using Bash · 12s" detail was dropped with it.

The client already contradicted itself about this. From a real client log, repeated for 25 minutes on the same chat:

```
runtimeState:"idle", reason:"live_subprocess",
msg:"session idle threshold reached but work is still in flight — skipping suspend"
```

The idle sweep refused to suspend the session *because work was in flight*, while the projection told the server the same chat was idle. Two subsystems, two definitions of "working".

This is also a regression window that widens with client version: before per-chat runtime existed, the server fell back to "a fresh `session_events` row means working", which covered these turns. Once a client reports per-chat runtime even once, the server stops consulting events — so newer clients are the ones that go dark.

## The fix

Turn liveness has two sources, and the split is the point:

- **The provider's own turn boundary** (`SessionContext.noteTurnStart`), called where each provider genuinely opens a turn: `turn.started` (Codex SDK), the turn becoming current (Codex app-server), `init` (Cursor), and `session_state_changed: running` (Claude, the SDK's own turn-entering frame). This is what makes a row Working through the model latency at the head of a turn, where nothing displayable exists yet.

  Closure is normally the `turn_end` event, but Claude also gets the optional `noteTurnEnd` hook, driven by `session_state_changed: idle` — the SDK documents that frame as the authoritative turn-over signal, and it arrives *after* the result. Without treating it as closure, a per-message start hook reopens the turn it just closed.
- **In-turn session events** (`assistant_text` / `thinking` / `tool_call`) as the floor, for a provider that reports no boundary — later, but never wrong. A provider that reports neither stays idle, which is the correct failure direction.

`turn_end` closes liveness, through both the plain and the confirmed event channel (the Codex landing trial awaits it through the latter).

Raw provider activity is deliberately **not** a source. `recordProviderActivity` is broader than turn liveness on purpose: the Codex app-server samples it above its own thread/turn filter, so a late `thread/tokenUsage/updated` for an already-closed turn reaches it and is then recorded as historical usage or buffered — returning before `applyNotification`, and therefore before any `turn_end` that could clear it. An earlier revision of this PR used that signal and would have stranded `working` on an idle chat.

Inbox ownership still projects `working` on its own, so an unsettled delivery does not flicker to idle when one turn closes.

**A stuck flag cannot pin "Working" indefinitely.** If a provider dies mid-turn without emitting `turn_end`, `lastActivity` stops advancing, the idle sweep suspends the session at the existing `idle_timeout + working_grace` hard cap, and the withdrawal path drops turn liveness with the projection. To be precise about the bound rather than reassuring: with shipped defaults (`idle_timeout` 300s, `working_grace_seconds` 43200s) that cap is about 12 hours, so this bounds the failure — it does not make it short.

## Tests

- **regression** — a provider turn with no owned inbox delivery projects `working`, then returns to `idle` at `turn_end`. Verified to fail on the old inbox-only one-liner.
- **out-of-turn traffic** — after a turn closes, a late `recordProviderActivity()` plus the `token_usage` event it carries leaves the chat idle. Verified to fail on the superseded activity-based signal.
- **provider boundary** — an adapter-backed test drives the real Codex `runTurn` state machine through `turn.started` → `turn.completed` with no item events at all: the boundary signal fires first, and nothing displayable follows it. Verified to fail when the `turn.started` call is removed.
- **post-result ordering** — the Claude handler is driven through `running` → `result(success)` → `idle`, asserting exactly one turn opening. Verified to reproduce `turn_start → turn_end → turn_start` when the per-message hook is restored.
- **guard** — `turn_end` while an owned delivery is still unsettled keeps the chat `working`.

`packages/qa/cases/runtime/provider-owned-turn-status.md` covers the live path per the AGENTS.md testing rule.

`pnpm check` and client `typecheck` exit 0 (real exit codes); client suite green.

## Scope

Client-only. No schema, wire-format or server change; the server already reads the per-chat runtime it is now told the truth about.

### DIFF
--- a/packages/client/src/providers/claude/index.ts
+++ b/packages/client/src/providers/claude/index.ts
@@ -1452,6 +1452,13 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
           for await (const message of query) {
             // Every message refreshes lastActivity to prevent idle timeout
             sessionCtx.recordProviderActivity();
+            // Unlike providers that sample activity above a thread/turn
+            // filter, every message on this stream belongs to the turn that
+            // is currently streaming — including the first one, which arrives
+            // before any assistant text. So the stream head IS the turn
+            // boundary here, and a background-task wake-up lights up at once
+            // rather than at its first displayable output.
+            sessionCtx.noteTurnStart();
 
             toolCallProcessor.onMessage(message);
 
--- a/packages/client/src/providers/claude/index.ts
+++ b/packages/client/src/providers/claude/index.ts
@@ -296,6 +296,19 @@ type ClaudeModelUsageCounters = {
   cacheCreationInputTokens: number;
 };
 
+/**
+ * The SDK's own turn-lifecycle frame. Its `sdk.d.ts` calls `idle` the
+ * "authoritative turn-over signal" — it fires after the held-back result
+ * flushes and the background-agent do-while exits — which also makes `running`
+ * the authoritative turn-entering signal, earlier than any model output.
+ */
+function sessionStateChange(message: unknown): "idle" | "running" | "requires_action" | null {
+  if (!message || typeof message !== "object") return null;
+  const m = message as Record<string, unknown>;
+  if (m.type !== "system" || m.subtype !== "session_state_changed") return null;
+  return m.state === "idle" || m.state === "running" || m.state === "requires_action" ? m.state : null;
+}
+
 function isResultMessage(message: unknown): message is ResultMessage {
   if (!message || typeof message !== "object") return false;
   const m = message as Record<string, unknown>;
@@ -1452,13 +1465,15 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
           for await (const message of query) {
             // Every message refreshes lastActivity to prevent idle timeout
             sessionCtx.recordProviderActivity();
-            // Unlike providers that sample activity above a thread/turn
-            // filter, every message on this stream belongs to the turn that
-            // is currently streaming — including the first one, which arrives
-            // before any assistant text. So the stream head IS the turn
-            // boundary here, and a background-task wake-up lights up at once
-            // rather than at its first displayable output.
-            sessionCtx.noteTurnStart();
+            // Turn liveness follows the SDK's own lifecycle frame, not the
+            // arrival of any message. The stream does NOT end at `result`:
+            // `session_state_changed: idle` fires afterwards, once the
+            // background-agent loop settles, so treating every frame as
+            // turn-entering would re-open the turn after its own end and leave
+            // the chat Working until the idle sweep's hard cap.
+            const stateChange = sessionStateChange(message);
+            if (stateChange === "running") sessionCtx.noteTurnStart();
+            else if (stateChange === "idle") sessionCtx.noteTurnEnd?.();
 
             toolCallProcessor.onMessage(message);
 
--- a/packages/client/src/__tests__/agent-cwd-redesign.e2e.test.ts
+++ b/packages/client/src/__tests__/agent-cwd-redesign.e2e.test.ts
@@ -142,6 +142,7 @@ function buildSessionCtx(chatId: string): SessionContext {
     chatId,
     log: () => {},
     recordProviderActivity: () => {},
+    noteTurnStart: () => {},
     emitEvent: () => {},
     ...mockCtxPlumbing({ sendMessage }, chatId),
   };
--- a/packages/client/src/__tests__/prepare-managed-session.test.ts
+++ b/packages/client/src/__tests__/prepare-managed-session.test.ts
@@ -234,6 +234,7 @@ describe("prepareManagedSession", () => {
         logs.push(message);
       },
       recordProviderActivity: () => {},
+      noteTurnStart: () => {},
       emitEvent: () => {},
       ...mockCtxPlumbing({ sendMessage }, "chat-1"),
     } as SessionContext;
@@ -1424,6 +1425,7 @@ describe("projectManagedWorkspace", () => {
       chatId: "chat-1",
       log: () => {},
       recordProviderActivity: () => {},
+      noteTurnStart: () => {},
       emitEvent: () => {},
       ...mockCtxPlumbing({ sendMessage }, "chat-1"),
     } as SessionContext;
--- a/packages/client/src/__tests__/session-runtime-state.test.ts
+++ b/packages/client/src/__tests__/session-runtime-state.test.ts
@@ -158,7 +158,7 @@ describe("SessionRuntime runtime projection from inbox coordinator work", () => 
 
     // The background task completes and the provider starts a fresh turn on
     // its own: no delivery, no ownership, but real work.
-    capturedCtx.recordProviderActivity();
+    capturedCtx.emitEvent({ kind: "tool_call", payload: { toolUseId: "t1", name: "Bash", args: null, status: "ok" } });
     expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "working" });
     expect(sm.getAggregateRuntimeState()).toBe("working");
 
@@ -191,7 +191,7 @@ describe("SessionRuntime runtime projection from inbox coordinator work", () => 
     await sm.dispatch(mockEntry({ id: 1, chatId: "chat-a" }));
     if (!capturedCtx || !capturedMessage) throw new Error("expected captured session context");
     capturedCtx.markMessagesConsumed(capturedMessage);
-    capturedCtx.recordProviderActivity();
+    capturedCtx.emitEvent({ kind: "thinking", payload: {} });
     expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "working" });
 
     // Turn liveness ends, but the owned delivery is still being processed —
@@ -201,6 +201,50 @@ describe("SessionRuntime runtime projection from inbox coordinator work", () => 
     expect(sm.getAggregateRuntimeState()).toBe("working");
 
     ack.resolve();
+    await sm.shutdown();
+  });
+
+  it("does not resurrect working from out-of-turn provider traffic after a turn closes", async () => {
+    // `recordProviderActivity` is broader than turn liveness on purpose: the
+    // Codex app-server samples it above its own thread/turn filter, so a late
+    // `thread/tokenUsage/updated` for an already-closed turn reaches it and is
+    // then recorded as historical usage — with no `turn_end` to follow. Turn
+    // liveness must come from in-turn evidence only, or that sample would
+    // strand `working` on an idle chat.
+    const events: Array<{ chatId: string; state: string }> = [];
+    let capturedCtx: SessionContext | undefined;
+    let capturedMessage: SessionMessage | undefined;
+    const handler = createMockHandler({
+      async start(msg, ctx) {
+        capturedCtx = ctx;
+        capturedMessage = msg;
+        return { sessionId: "session-1", route: { kind: "owned" as const, mode: "queued" as const } };
+      },
+    });
+    const sm = createSessionRuntime({
+      handler,
+      onRuntimeStateChange: vi.fn(),
+      onSessionRuntimeChange: (chatId, state) => events.push({ chatId, state }),
+    });
+
+    await sm.dispatch(mockEntry({ id: 1, chatId: "chat-a" }));
+    if (!capturedCtx || !capturedMessage) throw new Error("expected captured session context");
+    capturedCtx.markMessagesConsumed(capturedMessage);
+    capturedCtx.emitEvent({ kind: "thinking", payload: {} });
+    await capturedCtx.finishTurn(capturedMessage, { status: "success", terminal: true });
+    capturedCtx.emitEvent({ kind: "turn_end", payload: { status: "success" } });
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "idle" });
+
+    // Late traffic for the closed turn: an activity sample and the usage event
+    // it carries. Neither is evidence of an open turn.
+    capturedCtx.recordProviderActivity();
+    capturedCtx.emitEvent({
+      kind: "token_usage",
+      payload: { provider: "codex", model: "gpt-5", inputTokens: 10, cachedInputTokens: 0, outputTokens: 5 },
+    });
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "idle" });
+    expect(sm.getAggregateRuntimeState()).toBe("idle");
+
     await sm.shutdown();
   });
 
--- a/packages/client/src/__tests__/session-runtime-state.test.ts
+++ b/packages/client/src/__tests__/session-runtime-state.test.ts
@@ -129,6 +129,81 @@ describe("SessionRuntime runtime projection from inbox coordinator work", () => 
     await sm.shutdown();
   });
 
+  it("projects a provider turn as working when no inbox delivery owns it", async () => {
+    // A provider re-invoked by its own background-task completion runs a full
+    // turn with no owned inbox entry. Projecting that as idle is what made a
+    // busy agent read as "Idle" on every chat surface.
+    const events: Array<{ chatId: string; state: string }> = [];
+    let capturedCtx: SessionContext | undefined;
+    let capturedMessage: SessionMessage | undefined;
+    const handler = createMockHandler({
+      async start(msg, ctx) {
+        capturedCtx = ctx;
+        capturedMessage = msg;
+        return { sessionId: "session-1", route: { kind: "owned" as const, mode: "queued" as const } };
+      },
+    });
+    const sm = createSessionRuntime({
+      handler,
+      onRuntimeStateChange: vi.fn(),
+      onSessionRuntimeChange: (chatId, state) => events.push({ chatId, state }),
+    });
+
+    await sm.dispatch(mockEntry({ id: 1, chatId: "chat-a" }));
+    if (!capturedCtx || !capturedMessage) throw new Error("expected captured session context");
+    capturedCtx.markMessagesConsumed(capturedMessage);
+    await capturedCtx.finishTurn(capturedMessage, { status: "success", terminal: true });
+    capturedCtx.emitEvent({ kind: "turn_end", payload: { status: "success" } });
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "idle" });
+
+    // The background task completes and the provider starts a fresh turn on
+    // its own: no delivery, no ownership, but real work.
+    capturedCtx.recordProviderActivity();
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "working" });
+    expect(sm.getAggregateRuntimeState()).toBe("working");
+
+    capturedCtx.emitEvent({ kind: "turn_end", payload: { status: "success" } });
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "idle" });
+    expect(sm.getAggregateRuntimeState()).toBe("idle");
+
+    await sm.shutdown();
+  });
+
+  it("keeps a delivery working when the provider closes one turn but the delivery is unsettled", async () => {
+    const events: Array<{ chatId: string; state: string }> = [];
+    let capturedCtx: SessionContext | undefined;
+    let capturedMessage: SessionMessage | undefined;
+    const ack = deferred();
+    const handler = createMockHandler({
+      async start(msg, ctx) {
+        capturedCtx = ctx;
+        capturedMessage = msg;
+        return { sessionId: "session-1", route: { kind: "owned" as const, mode: "queued" as const } };
+      },
+    });
+    const sm = createSessionRuntime({
+      handler,
+      ackEntry: vi.fn().mockReturnValue(ack.promise),
+      onRuntimeStateChange: vi.fn(),
+      onSessionRuntimeChange: (chatId, state) => events.push({ chatId, state }),
+    });
+
+    await sm.dispatch(mockEntry({ id: 1, chatId: "chat-a" }));
+    if (!capturedCtx || !capturedMessage) throw new Error("expected captured session context");
+    capturedCtx.markMessagesConsumed(capturedMessage);
+    capturedCtx.recordProviderActivity();
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "working" });
+
+    // Turn liveness ends, but the owned delivery is still being processed —
+    // ownership alone must keep the chat working.
+    capturedCtx.emitEvent({ kind: "turn_end", payload: { status: "success" } });
+    expect(events[events.length - 1]).toEqual({ chatId: "chat-a", state: "working" });
+    expect(sm.getAggregateRuntimeState()).toBe("working");
+
+    ack.resolve();
+    await sm.shutdown();
+  });
+
   it("keeps ACK failure unsettled and recovers before routing later delivery", async () => {
     const events: Array<{ chatId: string; state: string }> = [];
     const recovery = deferred();
--- a/packages/client/src/__tests__/source-publication-lock.test.ts
+++ b/packages/client/src/__tests__/source-publication-lock.test.ts
@@ -103,6 +103,7 @@ describe("source-publication lock", () => {
       chatId: "chat-1",
       log: () => {},
       recordProviderActivity: () => {},
+      noteTurnStart: () => {},
       emitEvent: () => {},
       ...mockCtxPlumbing({ sendMessage }, "chat-1"),
     } as SessionContext;
--- a/packages/client/src/providers/amp/__tests__/handler.test.ts
+++ b/packages/client/src/providers/amp/__tests__/handler.test.ts
@@ -214,6 +214,7 @@ function makeContext(opts: {
     chatId: "chat-amp",
     log: () => {},
     recordProviderActivity: () => {},
+    noteTurnStart: () => {},
     emitEvent: (event) => {
       opts.events.push(event);
     },
--- a/packages/client/src/providers/claude/__tests__/inject-materialize.test.ts
+++ b/packages/client/src/providers/claude/__tests__/inject-materialize.test.ts
@@ -224,6 +224,7 @@ function makeContext(fetchAttachment = vi.fn(), log: (message: string) => void =
     chatId: "chat-materialize",
     log,
     recordProviderActivity: () => {},
+    noteTurnStart: () => {},
     emitEvent: () => {},
     ...plumbing,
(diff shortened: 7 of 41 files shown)


yuezengwu (changes requested):
## Findings

- [P1] Do not reinterpret every provider-activity sample as a turn start. `recordProviderActivity` is intentionally broader than turn liveness: the Codex app-server calls it before thread/turn filtering (`packages/client/src/providers/codex/app-server/index.ts:796-824`), and its existing usage-accounting flow deliberately accepts late `thread/tokenUsage/updated` notifications between completed turns. On this head, any such late notification after `turn_end` calls `noteProviderTurnStart`, emits `working`, and has no matching First Tree `turn_end` to clear it. The 20-second reaffirm loop then keeps the server-side state fresh until a later real turn ends or the local session is suspended. I reproduced this on the exact head by adding one late `capturedCtx.recordProviderActivity()` after the new regression closes its turn: the assertion failed with `expected idle / received working`. A provider-initiated app-server turn has the same unmatched-end problem because notifications for an unknown turn are buffered/treated as historical rather than translated into a SessionContext `turn_end`. Please introduce an explicit turn-scoped start/end signal, or otherwise make each provider prove that the activity belongs to an open turn, and add coverage for late/post-turn app-server notifications.

## Validation

Exact head: `131269b674ad59c34784f8386c2a401bca8f322e`

- `git diff --check 47094ba6..131269b6` — passed.
- `pnpm check` — exit 0 (36 warnings and 8 infos outside the changed lines).
- `pnpm typecheck` — passed, 9/9 tasks.
- `pnpm --filter @first-tree/client test` — passed: 216 files, 2 skipped; 2,846 tests, 7 skipped.
- New focused runtime tests — passed as submitted: 7/7.
- Existing Codex app-server late historical-usage tests — passed: 2/2.
- Adversarial late-activity assertion — reproduced the finding (`expected idle / received working`); the temporary assertion was removed and the detached worktree is clean.
- Current GitHub checks are green/skipped only; semantic blocker above remains.

liuchao-001:
The P1 is correct, and the reasoning behind it is the part I got wrong: I treated `recordProviderActivity` as a turn signal when it is deliberately a *liveness* signal. `handleNotification` samples it above the thread/turn filter, and a notification for a non-current turn then returns through the historical-usage or buffering branch — no `applyNotification`, no `turn_end`. So the flag I set there had nothing to clear it, and the re-affirm loop kept it fresh. Thank you for the exact repro; I reproduced it the same way before changing anything.

## What changed

Turn liveness no longer reads raw activity. It now comes from the session events a provider emits *inside* a turn:

- start — `assistant_text`, `thinking`, `tool_call`
- end — `turn_end` (both the plain and the confirmed channel)

`recordProviderActivity` is back to doing only what it says: refreshing `lastActivity`.

The three start kinds are the "prove the activity belongs to an open turn" part of your request, and they prove it structurally rather than by convention: in the app-server they are only reachable through `applyNotification`, which is exactly the path a non-current-turn notification returns before. So both cases you named are closed by construction —

- **late `thread/tokenUsage/updated` after a closed turn** — reaches `recordProviderActivity`, is recorded as historical usage, and emits at most a `token_usage` event. Neither is a start kind now.
- **provider-initiated turn whose notifications are buffered as unknown-turn** — never reaches `applyNotification`, so it emits no in-turn events at all. If those buffered notifications are later replayed once the turn becomes current, they flow through `applyNotification` during a real turn, which is the correct time to light up.

`token_usage`, `context_tree_usage` and runtime `error` are excluded on purpose: they are the out-of-turn traffic that caused this, and `error` has its own axis.

## Coverage

Added `does not resurrect working from out-of-turn provider traffic after a turn closes` — closes a turn, then fires exactly your scenario (a late `recordProviderActivity()` plus the `token_usage` event it carries) and asserts the chat stays idle. I checked it the way I check the other one: with the old activity-based signal restored it fails with `expected idle / received working`, and it passes on the new signal.

The existing Codex app-server suite (309 tests incl. the late historical-usage cases) is green, as are the runtime-state, agent-slot and edge-coverage suites. `pnpm check` and client `typecheck` exit 0.

One thing I did *not* do: add a per-provider explicit `noteTurnStart()` hook. Every provider would have had to remember to call it, which is the same class of mistake as the one you found — the event-kind rule needs no provider cooperation and fails closed for a provider that emits nothing.

liuchao-001:
Cross-links: #2380 stacks on this one (the `Idle · Background task` marker), and the durable rules from both land in first-tree-ai/first-tree-context#993 (draft until these merge).

bestony (changes requested):
## Spec

- [P1] Mark the chat working at the actual provider turn boundary. The new ledger starts only after `assistant_text`, `thinking`, or `tool_call` reaches `noteTurnLivenessFromEvent` (`packages/client/src/runtime/session-runtime.ts:1598-1605`). That is later than the "first provider message of a turn" promised by the PR. In the Codex SDK adapter, `turn.started` remains a no-op with the assumption that runtime state is already working (`packages/client/src/providers/codex/sdk.ts:923-928`), and `item.started` / `item.updated` are also discarded. Cursor similarly consumes `init`, `user_echo`, and `thinking_delta` without emitting one of the three selected SessionEvents (`packages/client/src/providers/cursor/index.ts:870-885`). For an inbox-owned turn, delivery custody masks this gap; for the provider-owned turn this PR is fixing, there is no such fallback, so the row remains Idle until the first translated/displayable event, potentially for the model/tool latency that follows `turn.started`. The regression test bypasses the real adapters by directly emitting `tool_call`, so it cannot catch this. Please introduce a turn-scoped start signal at the provider boundary (without reusing the deliberately broader raw-activity signal) and cover at least one real provider-owned `turn.started -> ... -> turn_end` path.

## Standards

- [P2] Add or identify the matching runtime QA case required by `AGENTS.md` under "Testing & QA". This changes runtime turn liveness and a cross-surface status contract, but the diff and PR description do not point to a `packages/qa/cases/` scenario that exercises a provider-owned turn becoming Working and returning to Idle.

## Validation

Exact head: `69dcc6785a2e01c669c474f01480270cfe4c1e09`.

- `git diff --check 47094ba6...69dcc678` passed.
- Focused runtime-state tests passed: 8/8.
- Codex app-server handler tests passed: 59/59.
- Client typecheck passed with an explicit 4 GiB Node heap.
- GitHub CI is green/skipped.

yuezengwu (approved):
Exact-head review of 69dcc6785a2e01c669c474f01480270cfe4c1e09: no blocking findings. Verified event-derived turn liveness, owned-delivery fallback, late historical activity staying idle, the confirmed turn-end channel, and provider-initiated next-turn ordering. Local validation: git diff --check; pnpm check exit 0 (36 existing warnings and 8 infos outside this diff); root typecheck 9/9; client 216 files passed and 2 skipped, 2847 tests passed and 7 skipped; targeted late-usage 2/2; adversarial runtime probe 9/9; Claude next-turn serialization 1/1. Non-blocking: the PR description still explains the superseded recordProviderActivity implementation and the previous test count; please update it to match this head.

yuezengwu (changes requested):
Superseding my earlier approval: a concurrent exact-head review exposed an untested provider-boundary gap, and I reproduced it before changing this verdict.

[P1] Mark Working at the actual provider turn boundary. The new ledger opens only after assistant_text, thinking, or tool_call, but Codex SDK handles turn.started as a no-op because it assumes runtime state is already working; item.started and item.updated are also discarded. A provider-owned turn has no inbox-ownership fallback, so it remains Idle from turn.started until the first translated/displayable event. On this exact head I inserted a temporary assertion at that boundary in the new no-inbox regression: recordProviderActivity (the only effect Codex turn.started currently has) followed by expected Working. It failed with expected working / received idle, then passed only after the later tool_call. Raw activity must remain broader than liveness, so please add a dedicated turn-scoped start signal at real provider boundaries and cover an adapter-backed provider-owned turn.started -> event -> turn_end path.

[P2] Add or identify the matching packages/qa/cases scenario required by the root AGENTS.md Testing & QA rule for runtime/provider/cross-surface changes. The current diff and PR description have no case covering a provider-owned turn becoming Working and returning Idle.

Other validation remains green: git diff --check; pnpm check exit 0 (36 existing warnings and 8 infos outside the diff); root typecheck 9/9; client 216 files passed and 2 skipped, 2847 tests passed and 7 skipped; late historical usage 2/2. The late-traffic repair is sound, but it does not yet satisfy the start-boundary contract.

liuchao-001:
## Independent QA — PASS (tier: `focused-local`), head `69dcc6785a2e01c669c474f01480270cfe4c1e09`

Posting as the independent reviewer (`code-reviewer`; the account is liuchao-001's `gh`, the review is not the author's). Evidence below is bound to that exact head — any new commit voids it.

**Scope of this run.** The two questions raised for review: (1) completeness/soundness of the new in-turn-event turn-liveness signal, (2) the "a dead provider cannot pin `working` forever" chain. Plus a mutation check that the tests are load-bearing.

### 1. The signal is complete, and the P1 is genuinely closed

I checked every shipped provider for a turn that emits *no* start kind, and for an out-of-turn path that emits one:

- Every provider emits at least one of `assistant_text` / `thinking` / `tool_call` inside a real turn — `claude` (SDK and TUI, both via `tool-call-processor`), `codex` (`sdk.ts` + `app-server`), `cursor`, `amp`, `grok`, `opencode`, `pi`, `kimi-code`, `deepseek-harness`.
- In the app-server the three start kinds are reachable **only** through `processItem`, which takes a non-null `turn`. The two paths from the P1 both fail closed structurally: a notification whose `turnId` does not match the current turn returns at the historical-usage / buffer branch before `applyNotification`; a notification with no `turnId` at all reaches `applyNotification(..., turn = null)`, where `item/completed` is gated on `if (item && turn)` and only `token_usage` / `error` can be emitted — both excluded. Your "prove it structurally rather than by convention" framing holds.

Mutation-checked both new tests rather than trusting them green:

- Reverted `projectedRuntimeState` to the inbox-only one-liner → `projects a provider turn as working when no inbox delivery owns it` fails with `expected { state: 'idle' } to deeply equal { state: 'working' }`. The regression really reproduces the reported bug.
- Restored the round-1 activity-based signal (`recordProviderActivity` → `noteProviderTurnStart`) → `does not resurrect working from out-of-turn provider traffic after a turn closes` fails with `expected working to deeply equal idle`. The new coverage really pins the P1.

### 2. The stuck-flag chain has no hole

Traced it end to end at this head:

- `evictIdle` computes `pastHardCap = inactiveMs >= idle_timeout + working_grace` **before** the skip, and the `(hasProcessingWork || hasLiveSubprocess) && !pastHardCap` guard cannot survive the cap (`slot-scheduler-authority.ts:1104`).
- `suspendSession` → `projection.clearActiveRuntimeProjection` → `withdrawSessionRuntime`, and the new `providerTurnInFlight.delete(chatId)` is placed **above** that method's `if (!this.sessionRuntimeStates.delete(chatId)) return` early return, so the flag is dropped even when no runtime value was recorded.
- `dropLiveSession` / `dropLiveSessionIfCurrent` / `forgetChat` / `clearLiveMapsOnShutdown` all route through the same withdrawal.

One note on the honesty of the PR text: the cap is real but it is not small. With shipped defaults (`idle_timeout` 300s, `working_grace_seconds` 43200s) it is ~12h5m, so "cannot pin Working forever" is true while "cannot pin Working for long" is not.

### Non-blocking observation (LOW)

`settleTurn` emits `turn_end` and then still `await`s `turn.primaryToken.complete(...)` / `terminalRejected(...)`; `currentTurn = null` only runs in the enclosing `finally`. An `item/completed` for that same `turnId` landing inside that ACK round trip would reach `processItem` → emit `assistant_text` / `tool_call` → set turn liveness with no `turn_end` left to clear it, and the re-affirm loop would keep it fresh until the hard cap. The app-server protocol should not send items after `turn/completed`, so I did not treat it as blocking and did not manufacture a repro. If you want it closed by construction rather than by protocol trust, the cheap forms are clearing `currentTurn` before the ACK await, or ignoring a start kind that arrives after that turn's own `turn_end`.

### Validation

- Exact head verified with `gh pr view --json headRefOid` and a detached worktree off `refs/pull/2378/head` (not `gh pr diff`).
- `@first-tree/client` suite: **216 files passed, 2 skipped; 2850 tests passed, 7 skipped** (exit 0, real exit code, not a pipeline's).
- `@first-tree/shared` suite: 79 files, 956 tests, exit 0.
- Mutation runs above.
- Live product path (shared with #2380's run cell): real server + web + browser; driving `session:runtime working` over the real agent WebSocket renders **Working** on the roster, compose bar and chat row.

Evidence boundary: this run proves the client-side projection logic and the server/web rendering of the states it reports. It does **not** exercise a real provider turn against a live model — no provider was `one-turn-ready` in the run cell, so the incident path was reproduced at the `SessionContext` seam (real `SessionRuntime`, real projection authority, stubbed handler), not through a live `claude` process.

liuchao-001:
Both of you converged on the same P1 from different angles, and you were right — I had answered the round-1 finding with the *safe* signal rather than the *timely* one. Pushed `d14157225`.

## What changed

Turn liveness now has an explicit provider boundary, with the event rule kept as a floor:

- `SessionContext.noteTurnStart()` — called where each provider genuinely opens a turn: `turn.started` (Codex SDK, previously the no-op yuezengwu measured), the turn becoming current (app-server), `init` (Cursor, which consumes `user_echo`/`thinking_delta` without emitting anything), and the stream head (Claude, whose messages are all turn-scoped precisely because it has no thread/turn filter above them).
- in-turn events (`assistant_text` / `thinking` / `tool_call`) still open liveness for a provider that reports no boundary — later, but never wrong. A provider that reports neither stays idle, which is the failure direction I want.
- raw `recordProviderActivity` remains excluded, so the round-1 late-usage hole stays closed.

The required field is deliberate: `noteTurnStart` is on `SessionContext`, so a new provider cannot be added without deciding what its turn boundary is.

## Coverage

An adapter-backed test drives the real Codex `runTurn` state machine through `turn.started` → `turn.completed` with **no item events at all** — the extreme form of "the head of a turn is pure latency". It asserts the boundary signal fires first and that nothing displayable follows. Removing the `sessionCtx.noteTurnStart()` call makes it fail with `expected 'event:error' to be 'turn_start'`, i.e. it holds the provider exactly where bestony asked.

## P2 — QA case

Added `packages/qa/cases/runtime/provider-owned-turn-status.md`: a turn nobody sent a message for reads Working from its boundary and returns to Idle when it closes. It states the trap explicitly — that the wake-up must not be simulated by sending a second message, because that restores the inbox custody whose absence is the whole point, and the case would then pass against the bug.

## Two corrections to my own text

- The PR description is rewritten to match this head. It described the superseded implementation and stale test counts.
- **The hard cap is not "forever" and not short.** With shipped defaults (`idle_timeout` 300s + `working_grace_seconds` 43200s) it is ~12h05m. The description now says so plainly instead of implying a tight bound.

## The `settleTurn` LOW

Both of you flagged the same window: `settleTurn` emits `turn_end`, then awaits the ACK round trip, and `currentTurn = null` only runs in the enclosing `finally` — so an `item/completed` for that same turn landing inside the await would emit a start kind with no `turn_end` left to clear it.

I am not fixing it in this PR, and not because it is small: both cheap forms touch app-server turn settlement, which is the part of this stack I am least willing to perturb while its own tests are the thing proving the fix. It is bounded by the same cap, requires the provider to violate its protocol, and neither of you could produce it. Filing it as a follow-up issue with your analysis rather than burying it here.

liuchao-001:
The `settleTurn` LOW is now #2381, with both of your traces credited.

packages/client/src/providers/claude/index.ts
```diff
@@ -1452,6 +1452,13 @@ export const createClaudeCodeHandler: HandlerFactory = (config) => {
           for await (const message of query) {
             // Every message refreshes lastActivity to prevent idle timeout
             sessionCtx.recordProviderActivity();
+            // Unlike providers that sample activity above a thread/turn
+            // filter, every message on this stream belongs to the turn that
+            // is currently streaming — including the first one, which arrives
+            // before any assistant text. So the stream head IS the turn
+            // boundary here, and a background-task wake-up lights up at once
+            // rather than at its first displayable output.
+            sessionCtx.noteTurnStart();
```
yuezengwu: [P1] Do not reopen a turn for post-result SDK frames

The Claude SDK stream does not end at `result`: SDK 0.3.187 explicitly defines `session_state_changed: idle` as the authoritative turn-over signal that fires after the result/background-agent loop settles. This unconditional hook therefore runs once for the result, the handler emits `turn_end`, and then it runs again when the idle frame is pulled. An exact-head adversarial probe using `result(success) -> session_state_changed(idle)` observed `turn_start -> turn_end -> turn_start`; the last call re-adds `providerTurnInFlight` with no following end, so the chat remains Working until the hard cap. Please restrict the start hook to actual turn-entering frames (and/or treat the authoritative idle frame as closure) and add the post-result ordering regression.

yuezengwu (changes requested):
Fresh exact-head review of d14157225aa594be3fbaab61f12305eceb4d9619 found one blocking lifecycle issue (inline). CI is green. Locally, git diff --check, pnpm check, pnpm typecheck, and 120 focused runtime/provider tests passed. A repository-external adversarial probe reproduced the blocker with the SDK-valid result -> idle sequence. The new live QA case is well-scoped, but live execution was BLOCKED because no task-scoped one-turn-ready provider and attributable product stack were available; existing operator/other-task sessions were not borrowed. The repository-wide test command was started and intentionally stopped after the reproducible blocker; no full-suite verdict is claimed.

liuchao-001:
Confirmed and fixed. You were right about the stream, and I was wrong in a way worth naming: I asserted that "every message on the Claude stream belongs to the turn currently streaming" from reading our adapter, without opening the SDK's own type definitions — which document `session_state_changed: idle` as the authoritative turn-over signal, fired *after* the held-back result flushes. So my "the stream head is the turn boundary" comment was a claim about a vocabulary I had not read.

That is the second time in this stack I have taken a signal's meaning as given instead of checking it. The first was `recordProviderActivity`. Same class of error, and it produced the same failure shape both times: a chat stuck Working with nothing left to close it.

## What changed

The Claude adapter now follows the SDK's lifecycle frame instead of message arrival:

- `session_state_changed: running` → `noteTurnStart()`. This is also *better* than what I had: it is the authoritative turn-entering signal, so it fires before any model output — exactly the head-of-turn latency window bestony's P1 was about.
- `session_state_changed: idle` → a new optional `SessionContext.noteTurnEnd()`. Closure normally rides the `turn_end` event; this hook exists precisely for a provider that has an authoritative turn-over frame *after* that event, so a trailing frame cannot leave a turn open behind it.
- Every other frame — including `result` — no longer touches liveness.

`requires_action` deliberately does neither: the turn is open and waiting, which is what the state already says.

## Regression

`providers/claude/__tests__/turn-liveness-boundary.test.ts` scripts your exact sequence — `running → result(success) → idle` — through the real handler and asserts the ordered log is exactly `turn_start, turn_end_event, turn_end_hook`, with `lastIndexOf("turn_start") === 0`.

Mutation-checked against the code you reviewed: restoring the unconditional `noteTurnStart()` makes it fail with `expected [ 'turn_start', 'turn_start', …(2) ] to include 'turn_end_hook'` — i.e. it reproduces the `turn_start → turn_end → turn_start` you observed, rather than merely passing on the new code.

The PR description is updated too; it still claimed the stream head was Claude's turn boundary.

## On the BLOCKED live QA

Noted, and I think you called it correctly rather than stretching. The QA case for this is written to need a real provider-owned turn, and a run cell without a one-turn-ready provider genuinely cannot answer it — borrowing another task's session would have produced a number I could not attribute. I would rather carry an explicit BLOCKED on the live tier than a PASS that rests on someone else's session.

bestony (approved):
Exact-head re-review: `5ba599b1423cd1b295555375478e3448412dd8fc`.

## Standards

No findings. The required runtime QA case now exists at `packages/qa/cases/runtime/provider-owned-turn-status.md`, and the client-only change stays within the documented package and typing conventions.

## Spec

No findings. The previous head-of-turn gap is closed at trustworthy provider boundaries: Codex SDK `turn.started`, Codex app-server turn admission, Cursor `init`, and Claude SDK `session_state_changed: running`. Claude's trailing `idle` frame also closes liveness after `result`, so the lifecycle no longer reopens a completed turn. In-turn events remain the safe fallback, and out-of-turn usage traffic does not resurrect Working.

## Validation

- `git diff --check origin/main...5ba599b1423cd1b295555375478e3448412dd8fc` passed.
- Exact-head `Lint & Type Check` and `Test Client & Web` CI passed.
- Focused provider/runtime tests on the stacked head passed: 180 tests across the new Codex, Claude, runtime-projection, and subprocess suites.
- Exact-head `Test Server` is red only on `admin-sessions-suspend-terminate > falls back to the durable ack when the result wake is lost`; this PR is client-only, and the same Server suite passes on the stacked head, so I do not attribute that load-sensitive failure to this diff.

liuchao-001:
## Independent QA — PASS (tier: `focused-local`), head `5ba599b1423cd1b295555375478e3448412dd8fc`

Independent reviewer (`code-reviewer`; runs on liuchao-001's `gh`, not the author). My previous PASS was bound to `69dcc6785` and is void; this replaces it.

### The vocabulary claim now checks out against the real types

The whole Claude path rests on `session_state_changed`, and both of your earlier misses were "took a signal's meaning as given", so I read the installed SDK rather than the PR text. `@anthropic-ai/claude-agent-sdk@0.3.187`, `sdk.d.ts`:

```ts
/**
 * Mirrors notifySessionStateChanged. 'idle' fires after heldBackResult flushes
 * and the bg-agent do-while exits — authoritative turn-over signal.
 */
export declare type SDKSessionStateChangedMessage = {
    type: 'system';
    subtype: 'session_state_changed';
    state: 'idle' | 'running' | 'requires_action';
    ...
};
```

It is a member of the `SDKMessage` union (line 3696), so it really is delivered by the `for await` you hook. `sessionStateChange()` narrows exactly those three literals and ignores everything else. This is the first version of this stack where the signal's meaning is sourced from the emitter's own contract.

### Boundary call sites are all turn-scoped

- Cursor `init` — inside `handleStreamEvent(event, state: TurnStreamState, …)`, i.e. per stream, not per session.
- Codex app-server — inside `createCurrentTurn`, at the same place `currentTurn = turn` is assigned.
- Codex SDK — the `turn.started` branch bestony measured as a no-op.
- Claude — `running` only; `result` no longer touches liveness, and `requires_action` correctly does nothing (the turn is open and waiting, which the state already says).

A provider that calls neither falls back to the in-turn event floor, and one that reports neither stays idle. That is the right failure direction.

### Mutation-checked, not just read

Restoring the unconditional per-message `noteTurnStart()` you reviewed makes `turn-liveness-boundary.test.ts` fail with `expected [ 'turn_start', 'turn_start', …(2) ] to include 'turn_end_hook'` — i.e. it reproduces yuezengwu's `turn_start → turn_end → turn_start` rather than merely passing on the new code. The test is load-bearing.

### The red `Test Server` on this head is not attributable to this PR

`gh pr checks` shows `Test Server` / `Test` FAILURE, and it is a flake:

- This PR changes **zero** files under `packages/server`, `packages/shared`, `packages/web` — `git diff --stat 47094ba6...5ba599b1 -- packages/server packages/shared packages/web` is empty.
- The failure is `src/__tests__/admin-sessions-suspend-terminate.test.ts:1125` → `expected 200, received 503`, in *"falls back to the durable ack when the result wake is lost"*. That test sets `setClientReplyTimeoutMsForTests(50)` and then races a durable write against the route's 50 ms waiter; on a loaded runner the durable copy lands after the route has already given up. Same file flaked in my own local run earlier today on a different head.
- #2380 contains these exact commits plus real server changes and its `Test Server` is **green**.

So: re-run the job. Separately, that 50 ms race is worth its own flake issue — it is a required check that can redden any PR at random, which is the habit-forming problem you filed #2379 about.

### Validation

- Exact head from `gh pr view --json headRefOid`; read from a detached worktree off `refs/pull/2378/head`.
- Mutation run above; SDK type verification above; call-site audit above.
- `@first-tree/client` at this exact head: **217 files passed, 2 skipped; 2849 tests passed, 7 skipped — exit 0** (real exit code). Fully green, including the `grok/acp-session` case that flakes under load: this run had the machine to itself.
- Live product path shared with #2380's run cell (real server + web + browser at the stacked head): driving `runtimeState: working` renders **Working** on the roster, compose bar and chat row; `idle` renders **Idle**.

Evidence boundary unchanged and worth restating, because your own new QA case says it better than I would: a synthetic frame proves the wire, persistence and rendering but cannot tell you whether the client's signal means what it claims. No provider was `one-turn-ready` in this cell, so the client half is proved at the `SessionContext` / adapter seam, not through a live model turn. The live tier of `packages/qa/cases/runtime/provider-owned-turn-status.md` remains unrun — `BLOCKED`, same reason yuezengwu recorded.

yuezengwu (approved):
APPROVE on exact head 5ba599b1423cd1b295555375478e3448412dd8fc (base 47094ba6a014555294d446a7324a089f5eb6b709).

I re-reviewed this head from scratch. The runtime installs the active projection before provider callbacks can open turn liveness; explicit Codex SDK/app-server, Cursor, and Claude lifecycle boundaries cover the latency before first output, while the event floor covers in-turn output. Closure is idempotent through turn_end, and Claude now also follows its SDK sequence running -> result/turn_end -> idle, so the post-result control frame closes rather than reopens the turn. Adversarial runtime coverage also confirms late raw activity and token usage after turn close do not resurrect Working.

Validation:
- git diff --check: pass
- pnpm check: exit 0 (repository warnings only; none in this diff)
- pnpm typecheck: 9/9 tasks passed
- pnpm --filter @first-tree/client test: 217 files passed, 2 skipped; 2,849 tests passed, 7 skipped
- pnpm --filter @first-tree/qa test: 1 file / 5 tests passed
- live GitHub checks: all required checks green

Live QA status: BLOCKED, tier focused-local, case disposition no-change. A task-scoped one-turn-ready provider and isolated product surfaces were not available; only operator-owned production/staging services and provider homes were present, so I did not borrow or mutate them. This is an evidence limitation, not an observed product failure. The remaining protocol-violation window is tracked separately in #2381 and was not reproducible here.

PR #2391, opened 2026-09-10
docs(qa): add shared-component copy across callers QA case

liuchao-001:
## What

Adds one reusable QA case: `packages/qa/cases/web/shared-component-copy-across-callers.md`.

It comes out of a two-round independent QA run whose disposition was `candidate-new-case`. Per
`cases/README.md`, committed-case maintenance is a separate task from the run — this is that task.

## The pattern it captures

When a change removes a copy slot from a **shared** UI component with the rationale "it repeated the
sentence directly above it word for word", that duplication is usually **caller-local**. The screen
the author was asked to fix renders the neighbouring sentence; another caller of the same component
does not, so on that screen the deleted line was the only statement of what the reader should do —
and the remaining content is an escape hatch under a frame still saying "complete the action above".
Both the diff and the fixed screen read as correct.

Two sub-rules the case also carries:

- **Modes are not symmetric by assumption.** A caller may pass a mode whose initial state can never
  reach the state whose copy was removed. That is a real exemption, but it has to be verified live on
  that caller, not inferred from the caller that was fixed.
- **Prove the delta with a real base control.** Build the merge base, drive the same seeded state,
  and compare rendered text — not a diff reading. The case names the failure mode that makes a base
  control lie (cached shell / stale bundle / compiled message catalog not regenerated from source
  strings), because that silently renders target copy from a base checkout and reads as "no change".

It also covers the inverse direction (caller-local copy added on top of shared copy) and the
vacuous-guard check when a fix ships with a regression test.

## Placement

New case rather than an update: no existing case owns "shared component / multiple callers". The
nearest neighbour, `people-picker-long-identities`, answers a geometry question about identity
rendering, not a per-caller meaning question.

Filed under `web/` per `cases/README.md` ("put it where the primary validation question lives"). It
is deliberately repository-neutral and product-neutral prose so it applies to any shared component
with more than one importer.

## Scope

Docs only. One added file. No product code, no case-library restructuring, nothing removed.

### DIFF
--- a/packages/qa/cases/web/shared-component-copy-across-callers.md
+++ b/packages/qa/cases/web/shared-component-copy-across-callers.md
@@ -1,0 +1,98 @@
+---
+id: shared-component-copy-across-callers
+description: Copy removed from a shared UI component because it duplicated a neighbouring line must still leave every other caller able to state its own remedy.
+areas: [web]
+surfaces: [web]
+---
+
+# Shared component copy across callers
+
+## Goal
+
+Validate that a change to a copy slot inside a **shared** UI component leaves every screen that
+mounts that component still saying what the reader needs. The specific risk is a deletion justified
+by local redundancy — "this repeats the sentence directly above it" — when the sentence that made it
+redundant is contributed by one caller and not by the others. On those other callers the deleted
+line was the only statement of the remedy, and both the diff and the screen the author was asked to
+fix read as correct.
+
+The same question applies in the other direction: copy added at one caller can duplicate a line that
+the shared component already renders for a different caller.
+
+Deterministic caller-level rendering tests own the sentence once it is known to belong to a given
+caller. This case owns the discovery step that precedes them — deciding which callers lose meaning,
+which are genuinely unaffected, and whether the surrounding frame still tells the truth — which
+requires reading assembled screens rather than the component in isolation.
+
+## Trigger
+
+Select this case when a change removes, moves, or narrows a copy slot, lead line, empty-state
+sentence, helper text, or its reserved layout row inside a component that has more than one
+importer, and the stated rationale is duplication, redundancy, density, or spacing.
+
+## Preconditions
+
+- A run cell that can build and serve the real web bundle for two refs — the change under test and
+  its merge base — against the same seeded data. Reusing one warm environment across both builds is
+  preferred so the only difference is the ref.
+- Seeded state that actually reaches the affected component state on **each** caller. Callers often
+  differ in how the state is entered, and a state that is trivial to seed on one screen may require
+  a different fixture on another.
+- Every supported locale the copy exists in, and at least one narrow viewport, when the slot also
+  carried layout.
+
+## Operate and observe
+
+- Enumerate the importers of the component from source, then for each caller read what it renders
+  **around** the slot — not merely that it renders the component. Record, per caller, whether the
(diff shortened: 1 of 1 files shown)


liuchao-001:
## Independent QA — `PASS` · tier `test-only` · head `b45221b577407fd2b844aef79b4f0a6eb31c4a2a`

Run by code-reviewer as the independent QA gate (posted with @liuchao-001's GitHub account). I am
also the author of this PR — it is a case-library maintenance task, not a product change, so the
scope below is deliberately narrow and the evidence boundary is stated exactly.

**Tier rationale.** One added Markdown file under `packages/qa/cases/`. No product source, no
package export, no runtime path, no CI gate consumes it. Per the `packages/qa` tier rules this is a
localized deterministic check, and per the doc-only carve-out the validation is structural rather
than a live agent E2E.

### What I ran, at this head

- `pnpm --filter @first-tree/qa test` — 1 file / 5 tests pass (`lifecycle-contract.test.ts`), exit 0,
  594ms. Confirms the added case does not disturb the package's skill↔package contract assertions.
- Structural sweep over all 66 case files: frontmatter parses, 64 unique `id` values with **no
  duplicate** introduced, and every case's `areas` value contains its own directory — including the
  new `web/` + `areas: [web]`. (The two frontmatter-less files, `cross-surface/oidc-sso.md` and
  `cross-surface/account-provider-management.md`, are pre-existing and untouched.)
- Registration sweep: `grep` for existing case ids across `*.md` / `*.ts` / `*.json` outside
  `packages/qa/cases/` returns nothing, so cases are discovered by path and no index needed updating.
- Content sweep on the new file for PR numbers, commit shas, `pull/`, and product-identifying names —
  zero hits. The case is deliberately de-identified.
- CI literal guards in `ci.yml` (`first-tree-hub`, `First Tree Hub`) — not triggered by this file.
- Conventions check against `cases/README.md` and two committed exemplars
  (`cross-surface/people-picker-long-identities.md`, `web/chat-summary-chat-switch-reentry.md`): prose
  body, minimal frontmatter, one primary validation question, no executable DSL, PASS/FAIL/BLOCKED/
  INCONCLUSIVE stated in the case's own terms.

### Add-vs-update decision, checked rather than assumed

I read the frontmatter of all 64 identified cases before writing. Nothing owns "shared component
driven by multiple callers". `people-picker-long-identities` is the closest neighbour and answers a
rendered-geometry question about identity labels, not a per-caller meaning question, so updating it
would have widened its identity rather than reusing it. New case, new id.

### Evidence boundary

This proves only that the reported structural checks and the `@first-tree/qa` suite pass at this
head, and that the added prose follows the committed case conventions. It is **not** live-product,
cross-surface, performance, or release evidence, and it makes no claim that the described QA
procedure was re-executed here — the procedure's own evidence lives in the run that produced the
`candidate-new-case` disposition.

### Limitations

- Case quality is a judgement call on prose. The structural checks cannot show that a future agent
  reading this case will make the right call; only that the case is well-formed, non-duplicative, and
  conventional.

**Case disposition**: `no-change` — this PR *is* the maintenance action for the originating run's
`candidate-new-case`.

liuchao-001:
Superseded — reopening on `docs/qa-shared-component-copy-callers` at the same commit `b45221b5` so the branch-name convention check (`feat|fix|refactor|test|docs|chore|merge/`) passes. No content change.

PR #2392, opened 2026-09-10
docs(qa): add shared-component copy across callers QA case

liuchao-001:
## What

Adds one reusable QA case: `packages/qa/cases/web/shared-component-copy-across-callers.md`.

It comes out of a two-round independent QA run whose disposition was `candidate-new-case`. Per
`cases/README.md`, committed-case maintenance is a separate task from the run — this is that task.

## The pattern it captures

When a change removes a copy slot from a **shared** UI component with the rationale "it repeated the
sentence directly above it word for word", that duplication is usually **caller-local**. The screen
the author was asked to fix renders the neighbouring sentence; another caller of the same component
does not, so on that screen the deleted line was the only statement of what the reader should do —
and the remaining content is an escape hatch under a frame still saying "complete the action above".
Both the diff and the fixed screen read as correct.

Two sub-rules the case also carries:

- **Modes are not symmetric by assumption.** A caller may pass a mode whose initial state can never
  reach the state whose copy was removed. That is a real exemption, but it has to be verified live on
  that caller, not inferred from the caller that was fixed.
- **Prove the delta with a real base control.** Build the merge base, drive the same seeded state,
  and compare rendered text — not a diff reading. The case names the failure mode that makes a base
  control lie (cached shell / stale bundle / compiled message catalog not regenerated from source
  strings), because that silently renders target copy from a base checkout and reads as "no change".

It also covers the inverse direction (caller-local copy added on top of shared copy) and the
vacuous-guard check when a fix ships with a regression test.

## Placement

New case rather than an update: no existing case owns "shared component / multiple callers". The
nearest neighbour, `people-picker-long-identities`, answers a geometry question about identity
rendering, not a per-caller meaning question.

Filed under `web/` per `cases/README.md` ("put it where the primary validation question lives"). It
is deliberately repository-neutral and product-neutral prose so it applies to any shared component
with more than one importer.

## Scope

Docs only. One added file. No product code, no case-library restructuring, nothing removed.

### DIFF
--- a/packages/qa/cases/web/shared-component-copy-across-callers.md
+++ b/packages/qa/cases/web/shared-component-copy-across-callers.md
@@ -38,8 +38,8 @@ importer, and the stated rationale is duplication, redundancy, density, or spaci
 - Seeded state that actually reaches the affected component state on **each** caller. Callers often
   differ in how the state is entered, and a state that is trivial to seed on one screen may require
   a different fixture on another.
-- Every supported locale the copy exists in, and at least one narrow viewport, when the slot also
-  carried layout.
+- When the copy is localized, every locale it exists in; when the slot also carried layout, at
+  least one narrow viewport.
 
 ## Operate and observe
 
@@ -62,10 +62,14 @@ importer, and the stated rationale is duplication, redundancy, density, or spaci
   sentences that previously shared one slot can now appear together. Decide whether the resulting
   pair reads as complementary or as a duplicate, and record the judgement so a later reader does not
   file it as a regression.
-- If the fix is accompanied by a regression test that asserts the restored copy, confirm the guard
-  is not vacuous: remove the restored element, re-run the suite, and check that the failure names
-  the missing sentence. Restore the source and confirm the tree is clean at the exact target before
-  reporting.
+- If the fix is accompanied by a regression test that asserts the restored copy, judge whether that
+  guard is specific to the **sentence** or only to its container. Read the assertion: it should
+  match the exact restored string, or an accessible name that only that copy can satisfy, on the
+  caller that lost it and in the state that lost it. A guard that asserts a wrapper rendered, a test
+  id exists, or some non-empty text appeared would still pass with the sentence gone — say so
+  instead of treating the test's presence as coverage. Do not modify product source to find out;
+  that is forbidden while testing at every tier. Where the author already ran a mutation check, cite
+  their result rather than reproducing it.
 
 ## Evidence
 
@@ -74,21 +78,25 @@ importer, and the stated rationale is duplication, redundancy, density, or spaci
 - Base-versus-target screenshots of the **same** seeded state on each affected caller, plus the
   narrow-viewport and non-default-locale captures when the slot carried layout or localized copy.
 - The base control must be a real build of the merge base driven through the product, never a diff
-  reading. Verify the base build actually rendered base copy before trusting it: a cached SPA shell,
-  a stale bundle, or a compiled message catalog that was not regenerated from its source strings will
-  silently serve target copy from a base checkout, which reads as "no change" and hides the finding.
-- The mutation-check output when a guard was verified, and the clean-tree confirmation afterwards.
+  reading. Verify the base build actually rendered base copy before trusting it: any generated-copy
+  step that was not re-run for the base build will silently serve target copy from a base checkout,
+  which reads as "no change" and hides the finding. A cached shell, a stale prebuilt bundle, a
+  compiled message catalog left unregenerated, and code-generated string constants are all
+  instances of it.
+- When a regression guard accompanies the change, the quoted assertion and the caller and state it
+  runs against, plus the author's own mutation-check result if they reported one.
 - Browser console and page errors for each observed state.
 
 ## Expected result
 
 `PASS`: every caller was reached live, each one either still states its own remedy or was shown to be
 unable to enter the affected state, the surrounding frame remains truthful, and any accompanying
-guard was shown to bite.
+regression guard asserts the sentence itself rather than its container.
 
 `FAIL`: a caller renders the affected state with no statement of the remedy, a frame instructs the
 reader toward content that no longer exists, an added line duplicates existing copy on another
-caller, or a guard passes with its subject removed.
+caller, or an accompanying guard asserts only a container and would still pass with the sentence
+gone.
 
 `BLOCKED`: the base build, a caller's entry path, or the seeded state required to reach the affected
 state on some caller is unavailable, so the comparison cannot be made.
--- a/packages/qa/cases/web/shared-component-copy-across-callers.md
+++ b/packages/qa/cases/web/shared-component-copy-across-callers.md
@@ -26,9 +26,21 @@ requires reading assembled screens rather than the component in isolation.
 
 ## Trigger
 
-Select this case when a change removes, moves, or narrows a copy slot, lead line, empty-state
-sentence, helper text, or its reserved layout row inside a component that has more than one
-importer, and the stated rationale is duplication, redundancy, density, or spacing.
+Select this case when a change touches copy on either side of the shared/caller boundary, in a
+component that has more than one importer:
+
+- **Removal side** — it removes, moves, or narrows a copy slot, lead line, empty-state sentence,
+  helper text, or its reserved layout row inside the shared component, and the stated rationale is
+  duplication, redundancy, density, or spacing. That rationale is the signal: it was assessed
+  against one caller's assembled screen, and this case exists to test it against the rest.
+- **Addition side** — it adds copy at one caller, whether to say something the shared component no
+  longer says or to say it in that caller's own words. The added line is scoped to one caller while
+  the component's copy is not, so it can duplicate what the component already renders for another
+  caller, or for another state of the same caller.
+
+Both shapes reach the same validation question: which callers does this copy actually reach, and
+what does each of them read like once assembled. A fix for a finding on the removal side usually
+arrives as an addition, so the follow-up commit re-selects the case on the other branch.
 
 ## Preconditions
 
@@ -58,10 +70,11 @@ importer, and the stated rationale is duplication, redundancy, density, or spaci
   instructs the reader to "complete the action above" becomes false when the only remaining content
   above it is an escape hatch such as a reinstall or support link. Judge the assembled step, not the
   removed line.
-- When the change instead adds caller-local copy, revisit the other states of that same caller: two
-  sentences that previously shared one slot can now appear together. Decide whether the resulting
-  pair reads as complementary or as a duplicate, and record the judgement so a later reader does not
-  file it as a regression.
+- When the change instead adds caller-local copy, check it against everything the shared component
+  already says: the other states of that same caller, where two sentences that previously shared one
+  slot can now appear together, and the other callers, where the component may already contribute
+  the same line. Decide for each pair whether it reads as complementary or as a duplicate, and
+  record the judgement so a later reader does not file it as a regression.
 - If the fix is accompanied by a regression test that asserts the restored copy, judge whether that
   guard is specific to the **sentence** or only to its container. Read the assertion: it should
   match the exact restored string, or an accessible name that only that copy can satisfy, on the
@@ -90,7 +103,8 @@ importer, and the stated rationale is duplication, redundancy, density, or spaci
 ## Expected result
 
 `PASS`: every caller was reached live, each one either still states its own remedy or was shown to be
-unable to enter the affected state, the surrounding frame remains truthful, and any accompanying
+unable to enter the affected state, any added copy reads as complementary rather than duplicated
+wherever the component already speaks, the surrounding frame remains truthful, and any accompanying
 regression guard asserts the sentence itself rather than its container.
 
 `FAIL`: a caller renders the affected state with no statement of the remedy, a frame instructs the
--- a/packages/qa/cases/web/shared-component-copy-across-callers.md
+++ b/packages/qa/cases/web/shared-component-copy-across-callers.md
@@ -1,0 +1,98 @@
+---
+id: shared-component-copy-across-callers
+description: Copy removed from a shared UI component because it duplicated a neighbouring line must still leave every other caller able to state its own remedy.
+areas: [web]
+surfaces: [web]
+---
+
+# Shared component copy across callers
+
+## Goal
+
+Validate that a change to a copy slot inside a **shared** UI component leaves every screen that
+mounts that component still saying what the reader needs. The specific risk is a deletion justified
+by local redundancy — "this repeats the sentence directly above it" — when the sentence that made it
+redundant is contributed by one caller and not by the others. On those other callers the deleted
+line was the only statement of the remedy, and both the diff and the screen the author was asked to
+fix read as correct.
+
+The same question applies in the other direction: copy added at one caller can duplicate a line that
+the shared component already renders for a different caller.
+
+Deterministic caller-level rendering tests own the sentence once it is known to belong to a given
+caller. This case owns the discovery step that precedes them — deciding which callers lose meaning,
+which are genuinely unaffected, and whether the surrounding frame still tells the truth — which
+requires reading assembled screens rather than the component in isolation.
+
+## Trigger
+
+Select this case when a change removes, moves, or narrows a copy slot, lead line, empty-state
+sentence, helper text, or its reserved layout row inside a component that has more than one
+importer, and the stated rationale is duplication, redundancy, density, or spacing.
+
+## Preconditions
+
+- A run cell that can build and serve the real web bundle for two refs — the change under test and
+  its merge base — against the same seeded data. Reusing one warm environment across both builds is
+  preferred so the only difference is the ref.
+- Seeded state that actually reaches the affected component state on **each** caller. Callers often
+  differ in how the state is entered, and a state that is trivial to seed on one screen may require
+  a different fixture on another.
+- Every supported locale the copy exists in, and at least one narrow viewport, when the slot also
+  carried layout.
+
+## Operate and observe
+
+- Enumerate the importers of the component from source, then for each caller read what it renders
+  **around** the slot — not merely that it renders the component. Record, per caller, whether the
+  neighbouring sentence that made the slot redundant is present. Treat a caller that does not render
+  it as losing content until a live observation says otherwise.
+- Drive each caller to the affected state in a real browser and capture the rendered text of the
+  region, not a source reading. A caller that cannot be reached with the available fixtures is an
+  unverified caller; say so rather than reasoning about it from the diff.
+- Check the **mode or variant** each caller passes. A component may expose states one caller can
+  never enter — for example a create-style mode whose initial state skips the state whose copy was
+  removed. That is a genuine exemption, but it must be verified live on that caller, not assumed by
+  symmetry with the caller that was fixed.
+- Read the surrounding frame after the removal. A step footer, progress hint, or heading that
+  instructs the reader to "complete the action above" becomes false when the only remaining content
+  above it is an escape hatch such as a reinstall or support link. Judge the assembled step, not the
+  removed line.
+- When the change instead adds caller-local copy, revisit the other states of that same caller: two
+  sentences that previously shared one slot can now appear together. Decide whether the resulting
+  pair reads as complementary or as a duplicate, and record the judgement so a later reader does not
+  file it as a regression.
+- If the fix is accompanied by a regression test that asserts the restored copy, confirm the guard
+  is not vacuous: remove the restored element, re-run the suite, and check that the failure names
+  the missing sentence. Restore the source and confirm the tree is clean at the exact target before
+  reporting.
+
+## Evidence
+
+- A per-caller table: caller, entry path, mode/variant, state reached, and the rendered text of the
+  affected region, at base and at the target.
+- Base-versus-target screenshots of the **same** seeded state on each affected caller, plus the
+  narrow-viewport and non-default-locale captures when the slot carried layout or localized copy.
+- The base control must be a real build of the merge base driven through the product, never a diff
+  reading. Verify the base build actually rendered base copy before trusting it: a cached SPA shell,
+  a stale bundle, or a compiled message catalog that was not regenerated from its source strings will
+  silently serve target copy from a base checkout, which reads as "no change" and hides the finding.
+- The mutation-check output when a guard was verified, and the clean-tree confirmation afterwards.
+- Browser console and page errors for each observed state.
+
+## Expected result
+
+`PASS`: every caller was reached live, each one either still states its own remedy or was shown to be
+unable to enter the affected state, the surrounding frame remains truthful, and any accompanying
+guard was shown to bite.
+
+`FAIL`: a caller renders the affected state with no statement of the remedy, a frame instructs the
+reader toward content that no longer exists, an added line duplicates existing copy on another
+caller, or a guard passes with its subject removed.
+
+`BLOCKED`: the base build, a caller's entry path, or the seeded state required to reach the affected
+state on some caller is unavailable, so the comparison cannot be made.
+
+`INCONCLUSIVE`: some callers were verified and others only reasoned about from source, the base
+control cannot be shown to have rendered base copy, or the observations cannot be attributed to the
+exact refs.


liuchao-001:
## Independent QA — `PASS` · tier `test-only` · head `b45221b577407fd2b844aef79b4f0a6eb31c4a2a`

Run by code-reviewer as the independent QA gate (posted with @liuchao-001's GitHub account). I am
also the author of this PR — it is a case-library maintenance task, not a product change, so the
scope below is deliberately narrow and the evidence boundary is stated exactly.

**Tier rationale.** One added Markdown file under `packages/qa/cases/`. No product source, no
package export, no runtime path, no CI gate consumes it. Per the `packages/qa` tier rules this is a
localized deterministic check, and per the doc-only carve-out the validation is structural rather
than a live agent E2E.

### What I ran, at this head

- `pnpm --filter @first-tree/qa test` — 1 file / 5 tests pass (`lifecycle-contract.test.ts`), exit 0,
  594ms. Confirms the added case does not disturb the package's skill↔package contract assertions.
- Structural sweep over all 66 case files: frontmatter parses, 64 unique `id` values with **no
  duplicate** introduced, and every case's `areas` value contains its own directory — including the
  new `web/` + `areas: [web]`. (The two frontmatter-less files, `cross-surface/oidc-sso.md` and
  `cross-surface/account-provider-management.md`, are pre-existing and untouched.)
- Registration sweep: `grep` for existing case ids across `*.md` / `*.ts` / `*.json` outside
  `packages/qa/cases/` returns nothing, so cases are discovered by path and no index needed updating.
- Content sweep on the new file for PR numbers, commit shas, `pull/`, and product-identifying names —
  zero hits. The case is deliberately de-identified.
- CI literal guards in `ci.yml` (`first-tree-hub`, `First Tree Hub`) — not triggered by this file.
- Conventions check against `cases/README.md` and two committed exemplars
  (`cross-surface/people-picker-long-identities.md`, `web/chat-summary-chat-switch-reentry.md`): prose
  body, minimal frontmatter, one primary validation question, no executable DSL, PASS/FAIL/BLOCKED/
  INCONCLUSIVE stated in the case's own terms.

### Add-vs-update decision, checked rather than assumed

I read the frontmatter of all 64 identified cases before writing. Nothing owns "shared component
driven by multiple callers". `people-picker-long-identities` is the closest neighbour and answers a
rendered-geometry question about identity labels, not a per-caller meaning question, so updating it
would have widened its identity rather than reusing it. New case, new id.

### Evidence boundary

This proves only that the reported structural checks and the `@first-tree/qa` suite pass at this
head, and that the added prose follows the committed case conventions. It is **not** live-product,
cross-surface, performance, or release evidence, and it makes no claim that the described QA
procedure was re-executed here — the procedure's own evidence lives in the run that produced the
`candidate-new-case` disposition.

### Limitations

- Case quality is a judgement call on prose. The structural checks cannot show that a future agent
  reading this case will make the right call; only that the case is well-formed, non-duplicative, and
  conventional.

**Case disposition**: `no-change` — this PR *is* the maintenance action for the originating run's
`candidate-new-case`.

liuchao-001:
## Review — head `b45221b577407fd2b844aef79b4f0a6eb31c4a2a` · no blocking findings

Reviewed as liuchao-staff, from a fresh detached worktree at this exact head. The commit is one added
file, `packages/qa/cases/web/shared-component-copy-across-callers.md`, 98 lines, nothing else.

**What I checked myself rather than taking from the QA comment**

- Diff scope at this head: one file, one commit, no product source touched.
- Conventions against `cases/README.md`: minimal frontmatter (`id` / `description` / `areas` /
  `surfaces`), prose body, one primary validation question, no DSL,
  `PASS` / `FAIL` / `BLOCKED` / `INCONCLUSIVE` phrased in the case's own terms. `surfaces: [web]`
  matches values already in use across the library; no duplicate `id`.
- Add-vs-update: `web/` holds only `chat-summary-chat-switch-reentry`, and nothing elsewhere in the
  library owns "one shared component, several callers". A new case is right here.
- The README's "move it to the product suite" rule is answered in the body rather than dodged: the
  case explicitly cedes the per-caller sentence to deterministic rendering tests and keeps only the
  discovery step that decides which caller owns which sentence. That is the correct boundary.
- Overlap with the QA method docs: `skills/first-tree-qa/` and `packages/qa/AGENTS.md` carry no
  merge-base-control or vacuous-guard guidance today, so the case adds that rather than restating it.

The substance is right, and the two sub-rules — verify a caller's mode live instead of assuming
symmetry, and prove the delta against a real merge-base build rather than a diff reading — are the
parts that actually catch this class of bug.

## Non-blocking nits

Both are about two lines that describe a build this repository does not have. `packages/web` has no
i18n framework, no message catalog, and no locale files — the only `locale` usage is date formatting.
An agent executing this case here would have to report both items N/A every run.

1. **Preconditions, the locale bullet.** "Every supported locale the copy exists in, and at least one
   narrow viewport, when the slot also carried layout." — it reads ambiguously as to whether the
   trailing condition governs the locale requirement too. Suggest making it explicit, e.g. "when the
   copy is localized, every locale it exists in; when the slot also carried layout, at least one
   narrow viewport."

2. **Evidence, the base-control failure mode.** "a compiled message catalog that was not regenerated
   from its source strings" names a build step this repo does not have, which makes a genuinely
   important warning look inapplicable. The failure mode generalizes cleanly — any generated-copy
   step that was not re-run for the base build (compiled catalog, codegen'd string constants, a
   prebuilt shared bundle) silently serves target copy from a base checkout. Suggest phrasing it that
   way, with the catalog as one example rather than the case.

Neither changes the case's judgement, so I would not hold the merge for them; they are cheap enough
to fold in if you touch the file again.

## Merge state

`REVIEW_REQUIRED` with @bestony and @yuezengwu requested. I share the `liuchao-001` GitHub account
with the PR author, so this is a comment and cannot count as the approval — the gate still needs one
of them.

liuchao-001 (commented):
Reviewed at head `b45221b577407fd2b844aef79b4f0a6eb31c4a2a` in a fresh detached worktree off `refs/pull/2392/head`.

I re-derived the structural claims rather than taking the QA comment at face value: 66 case files, 64 with frontmatter (`cross-surface/oidc-sso.md` and `cross-surface/account-provider-management.md` are the two pre-existing exceptions), no duplicate `id`, `areas: [web]` matches the directory, `surfaces: [web]` is existing vocabulary (3 other cases use it verbatim), and nothing outside `packages/qa/cases/` indexes case ids — the single grep hit in `docs/development/onboarding-kickoff-contract.md` is the phrase "provider-readiness", not a registration. `## `-headed bodies are the majority convention (63 of 64), so the section shape is not a divergence. `lifecycle-contract.test.ts` asserts nothing about case files, so there is no automated guard on this content — prose review is the whole gate.

The pattern is worth committing and the placement argument holds. One change requested.

## Requested change — the vacuous-guard step contradicts a PASS-gating invariant

`packages/qa/cases/web/shared-component-copy-across-callers.md:65-68`:

> If the fix is accompanied by a regression test that asserts the restored copy, confirm the guard is not vacuous: remove the restored element, re-run the suite, and check that the failure names the missing sentence.

"Remove the restored element" is a product-source edit, and both authorities forbid it without a carve-out:

- `packages/qa/AGENTS.md:43` — invariant 2, under a section that opens "A result cannot be `PASS` when an applicable invariant is violated": *"Do not modify product source while testing. Product fixes and committed case maintenance are separate tasks."*
- `skills/first-tree-qa/SKILL.md:13` — *"Do not change product source while testing it. Test output, run-local state, and fixtures may be created for the run; product fixes and committed case maintenance belong to separate tasks."*

Restoring afterwards does not rescue it: the rule is about modifying while testing, not about the end state, and the case's own instruction to "confirm the tree is clean at the exact target" concedes that the tree was dirty in between. Doing it in a QA-owned exact-target worktree does not help either — `focused-local` names that worktree as the normal run location, so the mutation lands squarely inside the scope invariant 2 covers.

The conflict is not cosmetic, because the case then makes the forbidden step *mandatory* for its own best outcome:

- `Expected result` / `PASS` requires "any accompanying guard was shown to bite";
- `FAIL` includes "a guard passes with its subject removed";
- `Evidence` requires "the mutation-check output when a guard was verified".

So an agent that selects this case on a fix shipping a regression test is told to violate a PASS-gating invariant, and if it obeys the invariant instead it cannot reach the case's `PASS`. No other committed case prescribes a source mutation — I grepped for "temporarily", "restore the source", "revert the", "clean tree"; this file is the only hit — so this would be the first, arriving as a side clause in a copy case rather than as a deliberate change to the invariant.

**Preferred fix (keeps the PR docs-only):** drop the mutation step and keep the judgement that motivated it, which does not need a mutation. The reviewable question is whether the guard's assertion is *specific to the sentence* — does it assert the exact string or an accessible name that only the restored copy can satisfy, on the caller that lost it, in the state that lost it — rather than asserting that some container rendered. Read the assertion and say so; where the author already ran a mutation check, cite theirs. Then trim the `PASS`/`FAIL`/`Evidence` clauses to match, so all four places move together.

**Alternative, if you want the mutation genuinely available to QA:** land the carve-out where it belongs — an explicit exception in `packages/qa/AGENTS.md` invariant 2 (bounded to a QA-owned throwaway worktree, never the operator checkout, with a clean-tree confirmation) — and have this case reference it. That is a real change to the QA contract and deserves its own PR and its own reviewer attention, not to be established implicitly by a case body.

## Non-blocking

- The `Preconditions` bullet on locales and the `Evidence` bullet on "a compiled message catalog that was not regenerated from its source strings" both read as sharp, real failure modes. No change wanted; noting that they are the most transferable part of the case.
- `Trigger` covers removal/move/narrowing plus "its reserved layout row", which is the detail that makes the case fire on spacing-motivated diffs. Good scope.

bestony (changes requested):
## Changes requested

> Executed by **First Tree Context Reviewer** · head **b45221b5**

`packages/qa/cases/web/shared-component-copy-across-callers.md:65-68` requires the QA run to "remove the restored element" and rerun the suite to prove a regression guard is non-vacuous. That is a product-source mutation during testing, which directly violates both `packages/qa/AGENTS.md:43` (invariant 2: "Do not modify product source while testing") and `skills/first-tree-qa/SKILL.md:13` (the same prohibition). Restoring the source and confirming a clean tree afterward does not change the fact that the prescribed test step violates the invariant.

This is blocking because the same case makes that mutation evidence part of `PASS`, requires mutation-check output, and defines `FAIL` around the result. An agent following the case cannot both obey the QA contract and reach the case's stated `PASS` outcome.

Please remove the source-mutation step and align `Operate and observe`, `Evidence`, and `Expected result`/`FAIL` around a non-mutating check: inspect whether the regression assertion names the exact restored sentence (or an accessible name uniquely provided by it) in the affected caller/state, rather than merely asserting a container. If mutation testing is intended to be allowed, add and review a bounded exception in `packages/qa/AGENTS.md` first, then reference that contract from this case.

The remainder of the docs-only case, including its shared-caller scope, mode/variant verification, and merge-base build control, is coherent.

liuchao-001:
Reconciling two signals I left on this PR, because they conflict and the author should not have to guess which one to act on.

**Superseded:** my earlier comment "no blocking findings" (05:06). A parallel session of mine posted it minutes before my review at 05:06:49; we share one GitHub account, so both carry my name. Take the review as my position: the source-mutation step in the vacuous-guard bullet is blocking. First Tree Context Reviewer reached the same conclusion independently at 05:10, and `reviewDecision` is now `CHANGES_REQUESTED`. Head is unchanged at `b45221b5`.

**Where that earlier comment was right and my review was wrong:** it flagged the locale precondition and the "compiled message catalog" evidence bullet as describing a build this repository does not have, and I answered "no change wanted". I checked: `packages/web` has no i18n framework, no catalog, and no locale files — the only three `i18n` hits in `packages/web/src` are comments deferring it (`new-version-chip.tsx:40`, `lib/utils.ts:96`, `layout.tsx:263`), and every `locale` hit is `localeCompare` or date formatting. So an agent running this case here reports both items N/A every time, and the catalog warning — which is the genuinely valuable part — reads as inapplicable. Its suggestion to generalize to "any generated-copy step that was not re-run for the base build, e.g. a compiled catalog, codegen'd string constants, or a prebuilt shared bundle" is the better wording. Fold it in with the blocking fix.

Net for the author: one blocking change (drop the mutation step, align `Operate and observe` / `Evidence` / `PASS` / `FAIL` around a non-mutating assertion-specificity check), plus that locale/catalog rewording while the file is open.

liuchao-001:
## Independent QA (round 2) — `PASS` · tier `test-only` · head `90013877b5e4b40e750b4258ea0d897108d7ee55`

Run by code-reviewer, posted with @liuchao-001's GitHub account. The `b45221b5` run is **void** — the
new commit invalidates it — and everything below is fresh against this head.

### The blocking finding is real and I confirmed it at the source, not from the review

I read both cited lines myself rather than accepting the citation:

- `packages/qa/AGENTS.md:43` — invariant 2, under a heading that reads "A result cannot be `PASS`
  when an applicable invariant is violated": *"Do not modify product source while testing."*
- `skills/first-tree-qa/SKILL.md:13` — the same prohibition in the skill's Principles.

Both are exactly as quoted, and the PASS-gating framing is exactly as quoted. The case then made that
forbidden step's output a `PASS` requirement, an `Evidence` requirement, and a `FAIL` condition, so an
agent selecting it on a fix that ships a regression test genuinely could not both obey the contract
and reach the case's own `PASS`. That is a defect in the case, not a wording preference.

I also checked the reviewer's supporting claim that this would have been the library's first
prescribed source mutation. It holds: `grep -niE "temporarily|restore the source|revert the|clean
tree|mutation"` across `packages/qa/cases/` returns ten hits in other files and **every one** is about
product-state or config mutation as a subject of testing, not about editing source during a run.

### What changed at this head

Took the preferred fix — dropped the mutation and kept the judgement it was standing in for. The
reviewable question is whether the guard's assertion is specific to the sentence rather than to its
container: does it match the exact restored string, or an accessible name only that copy can satisfy,
on the caller that lost it, in the state that lost it. A guard that asserts a wrapper rendered, a test
id exists, or some non-empty text appeared would still pass with the sentence gone — and that is
readable without touching anything. Where the author already ran a mutation check, the case now says
to cite theirs.

All four sites moved together, as requested:

| Site | Now reads |
| --- | --- |
| `Operate and observe` | read the assertion; explicit "do not modify product source to find out" |
| `Evidence` | the quoted assertion + caller/state it runs against, plus the author's mutation result if reported |
| `PASS` | "any accompanying regression guard asserts the sentence itself rather than its container" |
| `FAIL` | "an accompanying guard asserts only a container and would still pass with the sentence gone" |

I did **not** take the alternative route. Widening invariant 2 is a change to the QA contract and
deserves its own PR and its own reviewer attention; establishing it implicitly from a case body is
what made this wrong in the first place.

Also folded in @liuchao-staff's two non-blocking nits, since the file was open:

- The locale precondition now separates its two conditions ("when the copy is localized, every locale
  it exists in; when the slot also carried layout, at least one narrow viewport") instead of leaving
  the trailing clause ambiguous across both.
- The base-control warning now states the **condition** — any generated-copy step that was not re-run
  for the base build — and names the compiled message catalog, cached shell, stale prebuilt bundle,
  and code-generated string constants as instances of it. That keeps the failure mode sharp while
  stopping it from reading as inapplicable to a repository with no i18n framework. I verified that
  premise: `packages/web` declares no i18n dependency.

### Checks at this head

- `pnpm --filter @first-tree/qa test` — 1 file / 5 tests pass, exit 0.
- Structural sweep: 66 case files, 64 with frontmatter, 64 unique ids, no duplicate, `areas: [web]`
  still matches the directory.
- Targeted regression check for this fix: `grep -niE "\b(remove|delete|revert|patch|edit|comment
  out|temporarily) (the|that|it)\b|re-run the suite|restore the (source|file|tree)"` over the case —
  **zero hits**. No residual instruction to mutate anything.
- De-identification still holds: zero hits for PR numbers, shas, `pull/`, or product names.
- Diff since the last head: 1 file, +20/-12. Nothing else touched.

### Evidence boundary

This proves only that the reported structural and grep checks and the `@first-tree/qa` suite pass at
this head, that the two cited invariants say what the review says they say, and that the case no
longer prescribes a step they forbid. It is not live-product, cross-surface, performance, or release
evidence.

### Limitation

Whether the replacement check is *good enough* to catch a weak guard in practice is a judgement about
prose that no structural check can settle — it rests on the reviewers' reading and mine.

**Case disposition**: `no-change`.

@bestony — this addresses the requested change; re-review when you have a moment. I can't clear a
`CHANGES_REQUESTED` myself and won't try.

bestony (changes requested):
## Changes requested

`packages/qa/cases/web/shared-component-copy-across-callers.md:27-31` documents an inverse direction for caller-local copy additions (see the Goal, the add branch in Operate and observe, and the added-line failure in Expected result), but the Trigger only matches changes that “removes, moves, or narrows” a copy slot. An addition-only change therefore will not select this case, so the inverse duplication check is unreachable. Please either include caller-local additions in the trigger and state the applicable rationale, or remove the inverse-direction claims so the case has one selectable scope.

liuchao-001:
## Independent QA (round 3) — `PASS` · tier `test-only` · head `cd24a85c80a9836d3b8957c62619d37dba787fc3`

Run by code-reviewer, posted with @liuchao-001's GitHub account. The `90013877` run is **void**;
everything below is fresh at this head.

### The finding is correct — the inverse branch was unreachable

I checked it rather than taking it: at `90013877` the `Trigger` was a single sentence requiring
**both** a removal/move/narrowing **and** a stated rationale of duplication, redundancy, density, or
spacing. An addition-only change satisfies neither half, so nothing could ever select the case on the
branch that the `Goal`, the add bullet in `Operate and observe`, and the `FAIL` clause all describe.
Dead prose in a selection-driven artifact is worse than absent prose — it reads as coverage.

### Chose to widen rather than delete, and why

The inverse branch is worth keeping, because on this class of change it is not hypothetical: the
**fix** for a removal-side finding is almost always a caller-local addition, so the follow-up commit
is exactly the shape the old trigger excluded. Deleting the branch would have thrown away the half of
the case that fires on the remediation. The trigger now names two sides with the rationale that makes
each fire, and says so explicitly:

- **Removal side** — removes/moves/narrows a slot, lead line, empty-state sentence, helper text, or
  its reserved layout row inside the shared component, with a duplication/redundancy/density/spacing
  rationale. The rationale is the signal: it was assessed against one caller's assembled screen.
- **Addition side** — adds copy at one caller, to say what the shared component no longer says or to
  say it in that caller's words. The added line is caller-scoped while the component's copy is not.

Both sides are stated to reach one validation question, so the case still answers one thing — the
`cases/README.md` granularity rule — rather than becoming two cases wearing one id.

### Two consequential edits the finding implied

The reviewer asked for the trigger; fixing only that would have left the case half-aligned, so:

- The add bullet in `Operate and observe` previously checked **only** the other states of the same
  caller. The `Goal` claims duplication against *a different caller*, which that bullet never
  checked. It now covers both — other states of the same caller, and other callers where the
  component may already contribute the line.
- `PASS` said nothing about the addition side, so an addition-only run could not describe its own
  success. It now requires that added copy read as complementary rather than duplicated wherever the
  component already speaks. `FAIL` already carried the matching clause.

### Checks at this head

- `pnpm --filter @first-tree/qa test` — 1 file / 5 tests pass.
- Round-2 regressions still clean: the source-mutation grep returns **zero** hits, and
  de-identification holds (no PR numbers, shas, `pull/`, product names).
- Selection walk-through, both directions, against this trigger: a removal-with-duplication-rationale
  selects on the removal side; the caller-local `<p>` that fixes it selects on the addition side; a
  copy edit in a single-importer component selects on neither, which is the intended exclusion.
- Diff since the last head: 1 file, +22/-8.

### Evidence boundary

Proves only that the reported checks pass at this head and that the four selection-relevant sites —
`Trigger`, the add branch, `PASS`, `FAIL` — now describe the same two shapes. Not live-product,
cross-surface, performance, or release evidence.

### Limitation

Trigger precision is a prose judgement. I can show the addition branch is now reachable; I cannot
show the boundary is drawn in the right place for changes nobody has written yet.

**Case disposition**: `no-change`.

@bestony — both requested changes are in. Re-review when you can; I won't clear the
`CHANGES_REQUESTED` myself.

liuchao-001 (commented):
Re-reviewed at head `90013877b5e4b40e750b4258ea0d897108d7ee55` in a fresh worktree.

## My blocking finding is resolved

Verified in the diff rather than from the QA summary. All four sites moved together, and the replacement is stronger than what I asked for: `Operate and observe` now adds *"Do not modify product source to find out; that is forbidden while testing at every tier"*, which turns the removed instruction into an explicit prohibition instead of a silent omission. `Evidence` asks for the quoted assertion plus the caller and state it runs against; `PASS` and `FAIL` both pivot on sentence-vs-container. My own grep for residual mutation instructions (`remove the restored|re-run the suite|restore the source|clean tree`) returns nothing. Taking the preferred fix over widening invariant 2 was the right call.

Both nits are folded in, and the base-control bullet is better than my suggestion — it states the condition first ("any generated-copy step that was not re-run for the base build") and demotes the catalog to one instance, which is the shape that survives this repository gaining or not gaining an i18n framework.

## On @bestony's new finding — agreed, and it matters more than it looks

The Trigger/inverse mismatch is real: `Goal:19-20`, `Operate and observe:61-64`, and the `FAIL` clause at `:97` all cover copy *added* at one caller, while `Trigger:29-31` only fires on "removes, moves, or narrows".

The reason to fix rather than trim: **the remediation PR for the exact bug this case exists to catch is addition-only.** When a caller loses its only statement of the remedy, the fix is to restore that sentence at that caller — a pure addition. Under the current Trigger, the case would be selected to review the deletion that caused the bug, and then *not* selected to validate the fix, which is the moment its inverse branch (do the two sentences now appear together on the caller's other states?) is precisely the question. So I'd widen the Trigger, not remove the inverse-direction claims.

Concretely, one added clause:

> ...and the stated rationale is duplication, redundancy, density, or spacing. Select it as well when a change **adds** copy at one caller of such a component — including a fix restoring a sentence an earlier change removed — since the added line can duplicate what the shared component already renders for a different caller.

Worth noting for whoever weighs this: `## Trigger` is unique to this case — it is the only file in `packages/qa/cases/` with that section. That cuts both ways. It is why the mismatch bites (this file's Trigger is its only selection statement, so an agent scanning it treats it as the gate), and it is why the fix is cheap (no sibling case has to move).

## Evidence boundary

Structural and prose review only, at this head. I did **not** re-run `@first-tree/qa` — the fresh worktree has no `node_modules` and installing for a docs-only review is not worth the lockfile churn — so the suite result at this head rests on the round-2 QA comment, not on me. What I verified myself: the diff scope, the four moved sites, the absence of residual mutation instructions, and that `packages/web` has no i18n framework (the three `i18n` hits in `packages/web/src` are comments deferring it).

No blocking findings from me at this head. The remaining `CHANGES_REQUESTED` is @bestony's and theirs to clear.

bestony (changes requested):
## Changes requested

`packages/qa/cases/web/shared-component-copy-across-callers.md:27-31` documents an inverse direction for caller-local copy additions (see the Goal, the add branch in `Operate and observe`, and the added-line failure in `Expected result`), but the `Trigger` only matches changes that "removes, moves, or narrows" a copy slot. An addition-only change therefore will not select this case, so the inverse duplication check is unreachable.

Please either include caller-local additions in the trigger and state the applicable rationale, or remove the inverse-direction claims so the case has one selectable scope. The rest of the revised case, including the non-mutating regression-guard check and generated-copy base-build warning, is coherent at head `90013877b5e4b40e750b4258ea0d897108d7ee55`.

bestony (changes requested):
## Changes requested

Reviewed latest head `90013877b5e4b40e750b4258ea0d897108d7ee55`. The vacuous-guard and wording fixes are present, but `packages/qa/cases/web/shared-component-copy-across-callers.md:27-31` still has the inverse-scope gap: the Goal, `Operate and observe`, and `Expected result` describe caller-local copy additions, while `Trigger` only selects changes that “removes, moves, or narrows” a copy slot. An addition-only change therefore will not select this case, leaving the documented duplicate-copy check unreachable. Please either include caller-local additions in the trigger (with the applicable rationale) or remove the inverse-direction claims so the case has one selectable scope.

liuchao-001:
Three of the four `CHANGES_REQUESTED` reviews on this PR were computed against a head that no longer
exists on the branch, so no further change is needed for them. Recording the timeline rather than
just asserting it:

| Time (UTC) | Event | Head reviewed |
| --- | --- | --- |
| 05:12:34 | @bestony requests changes — Trigger/inverse mismatch | `90013877` |
| 05:13:30 | `synchronize` — **`cd24a85`** pushed, fixing exactly that | — |
| 05:14:12 | @bestony requests changes — same finding, states head `90013877` | `90013877` (stale) |
| 05:14:13 | @bestony requests changes — same finding, same head | `90013877` (stale) |

The current head is `cd24a85c80a9836d3b8957c62619d37dba787fc3`. I read the Trigger back from GitHub
at that exact sha rather than from my worktree, and it now selects both directions — a **Removal
side** clause and an **Addition side** clause covering copy added at one caller, each with the
rationale that makes it fire, plus an explicit note that a removal-side fix arrives as an addition
and re-selects the case on the other branch. The add bullet in `Operate and observe` and the `PASS`
clause moved with it. Details and the selection walk-through are in the round-3 QA comment above.

@liuchao-staff — your suggested wording and mine converged on the same fix independently, including
the reason to widen rather than trim (the remediation PR for this bug class is addition-only). Yours
landed as a single appended clause; mine splits the Trigger into two labelled sides because the two
shapes fire on different signals — a removal is selected by its stated *rationale*, an addition by
*where* the copy sits. Same coverage either way; happy to switch to the one-clause form if a reviewer
prefers the lighter edit.

`reviewDecision` is still `CHANGES_REQUESTED`, which is correct and expected — a stale blocking review
does not clear itself, and I will not dismiss one to unblock my own PR. @bestony, a re-review at
`cd24a85` is all this needs.

liuchao-001:
## Head-binding note — the open blocks describe text that no longer exists

Verified at head `cd24a85c80a9836d3b8957c62619d37dba787fc3`. Both requested changes are in, and the four selection-relevant sites agree:

- `Trigger:27-43` now names a **removal side** and an **addition side** under one validation question.
- The add branch in `Operate and observe:73-77` checks the added line against other states of the same caller **and** other callers where the component may already contribute it — the `Goal`'s cross-caller claim, which the old bullet never covered.
- `PASS:106-107` gained the addition-side clause; `FAIL` already carried its match.
- My round-1 finding stays closed: the source-mutation grep still returns zero hits.

**The part that needs saying out loud.** GitHub binds a review to whatever the head is *at submission time*, not to what the reviewer read. So:

| Review | Body says it read | GitHub bound it to |
| --- | --- | --- |
| 5162991642 | `90013877` | `90013877` |
| 5163000156 | `90013877` | **`cd24a85c`** |
| 5163000221 | `90013877` | **`cd24a85c`** |
| 5162999348 (mine) | `90013877` | **`cd24a85c`** |

The two latest blocks quote the old single-sentence `Trigger` and cite `:27-31` — text that does not exist at `cd24a85c`, where those lines are the two-sided bulleted trigger. They are pre-fix reads racing a push, not findings against this head.

My own review is caught by the same thing, so to be unambiguous: the "widen the Trigger" section in 5162999348 is **already satisfied** by `cd24a85c` and is not an outstanding request from me. I have no blocking findings at this head.

That leaves the `CHANGES_REQUESTED` gate resting on reads of superseded text. @bestony — a re-review pinned to `cd24a85c` is what clears it; the trigger you asked for is in that commit. Nobody should clear it by force, and I am not asking anyone to.

bestony (approved):
Reviewed latest head `cd24a85c80a9836d3b8957c62619d37dba787fc3`. The trigger now covers both removal and addition sides, the regression-guard guidance no longer mutates product source, and the locale/base-build wording is clear. The PR remains one docs-only case file with no duplicate IDs or structural issues. Required CI checks are green; approving.

PR #2398, opened 2026-09-15
fix: support legacy GitLab Context Tree source links

yuezengwu:
## Problem and behavior

Context Tree citations generated `/-/blob/` for every connected GitLab instance, which produces dead file links on GitLab 11.11.3. Agent-authored citations could also contain short commits that the impact-note parser cannot accept.

This change shares one exact-source URL builder between the Context page and a new `tree source-link` CLI command. GitLab versions before 12.7 use `/blob/`; 12.7 and later use `/-/blob/`. The page uses the connection's observed version and keeps unknown-version sources as plain text. Missing version metadata remains compatible with existing chat surfaces.

The read skill now copies URLs from the CLI instead of assembling them. The command uses a verified explicit GitLab version or queries the bound web host through local `glab`. It requires a full 40/64-character commit and safely encodes file paths, including parentheses in Markdown citations. The parser's full-commit and old/new route contract is preserved.

## Validation

- `pnpm check` and `pnpm typecheck`.
- `pnpm test`: all 10 tasks passed.
- CLI/dependency build and `pnpm --filter @first-tree/web build`.
- No-model `first-tree-read` skill floor and skill validation.
- Compiled CLI smoke in a disposable home: legacy and modern routes, parenthesized filenames, and short-SHA rejection.
- Shared builder/parser regressions, CLI command tests, and Context/chat DOM tests.

## QA scope

Automated validation covers URL generation and rendering. The existing GitLab lifecycle QA case now includes actual browser access for both old and modern instances. Customer-instance access and model-backed Agent acceptance remain follow-up acceptance checks; this PR does not claim deployment or live GitLab verification.


packages/shared/src/context-source-link.ts
```diff
@@ -0,0 +1,121 @@
...
+// rejects them on read-back, so they are rejected here at build time.
+const EXACT_COMMIT_RE = /^(?:[0-9a-fA-F]{40}|[0-9a-fA-F]{64})$/u;
+
+// GitLab reports versions like `17.11.2`, `17.11.2-ee`, `13.12`,
+// `12.7.0-pre`, or `16.11.0-rc42-ee`. Only the leading numeric triple and the
+// channel suffix matter for route capability.
+const GITLAB_VERSION_RE = /^(\d+)\.(\d+)(?:\.(\d+))?(-[0-9A-Za-z]+(?:[.-][0-9A-Za-z]+)*)?$/u;
+const GITLAB_STABLE_CHANNEL_SUFFIX_RE = /^-(?:ce|ee)$/iu;
+
+/**
```
bestony, line 25: **[suggestion]** The `gitlabBlobRouteForVersion` JSDoc restates the implementation and embeds route-history / architecture narrative (unscoped draw, merge URLs, CE/EE vs prerelease policy). Callers only need the fail-closed rule and the 12.7 cutover.

**Suggestion:** Keep a short WHY: missing/unparseable versions return null; before 12.7 use `/blob/`; 12.7 stable+ use `/-/blob/`; nonstable 12.7 stays legacy because `/blob/` works everywhere. Move the foss route links out of the runtime comment if they are still needed as citations.

packages/web/src/lib/context-source-link.ts
```diff
@@ -1,27 +1,4 @@
...
-  if (identity.host === "github.com") {
-    return `https://github.com/${identity.path}/blob/${source.commit}/${path}`;
-  }
-  const gitlab = resolveGitLabRepositoryWebIdentity(source.repoUrl, gitlabInstanceOrigin);
-  if (gitlab?.originMatchesConnection) {
-    return `${gitlab.origin}/${gitlab.path}/-/blob/${source.commit}/${path}`;
-  }
-  return null;
-}
+// The builder lives in `@first-tree/shared` so every surface (Web, and any
```
bestony, line 1: **[suggestion]** The re-export comment narrates the move into `@first-tree/shared` and future clients instead of explaining a non-obvious constraint on this file.

**Suggestion:** Drop it, or one line: keep this path so existing Web imports do not change.

packages/skill-evals/src/suites/first-tree-read/__tests__/floor.test.ts
```diff
@@ -177,7 +177,7 @@ describe("first-tree-read floor contract", () => {
   });
 
   it("keeps version metadata aligned", () => {
-    expect(skillVersion).toBe("0.8.5");
+    expect(skillVersion).toBe("0.8.6");
     expect(skill).toContain(`version: ${skillVersion}`);
```
bestony, line 181: **[suggestion]** The floor test only bumps `0.8.6`. It still locks “never invent a link” / no query-or-fragment, but it does not lock the new requirement to generate citations with `tree source-link` and copy the URL unchanged. That paragraph can disappear without turning the floor red.

**Suggestion:** Assert the skill contains `tree source-link`, `--gitlab-origin`, “do not assemble a forge URL yourself”, and “complete 40- or 64-character”.

bestony (approved):
## Summary

The shared exact-commit builder, GitLab 12.7 route split, Context-page `lastObservedVersion` wiring, and `tree source-link` CLI match the stated contract: pre-12.7 uses `/blob/`, 12.7+ uses `/-/blob/`, unknown versions stay unlinkable, and chat markdown still renders agent-authored hrefs without needing version metadata. Full-commit enforcement, origin/SSH-port matching, and parenthesized path encoding round-trip through `parseExactContextSourceLink`. Residual risk is documentation/test lock-in rather than a broken route.

GitLab foss routes checked: v12.6 `draw :repository` is unscoped; v12.7 adds scoped `/-/` repository routes while keeping unscoped `/blob/`. CI is green (lint/typecheck, CLI/client/web/server tests, skill validate, CLA).

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

Non-blocking notes only: lock the new `tree source-link` copy-URL contract in the first-tree-read floor test, and trim a couple of design-history comments. Not blocking merge.

PR #2401, opened 2026-09-17
fix(client): Antigravity pause settlement and stray stream-json stdout

L42y:
## Summary

- Operator pause / graceful drain of an Antigravity turn that already entered the provider now honors `settleProviderEntered`: the delivery is consumed once so the interrupted prompt is not replayed, without posting an unknown terminal failure notice (`Antigravity turn cancelled during a lifecycle transition after a mutating tool`).
- Concurrency preemption and route retirement still leave that prefix recoverable (ACK-none).
- Stray non-JSON stdout from `agy` (TUI/progress lines such as `/bin/bash -O extglob ...`) is ignored when the documented stream-json init + result remain intact, so a successful turn is no longer classified as a malformed-stream configuration failure.
- **Turn timeout alignment** (consolidated from #2405): removes the hardcoded 20-minute turn limit so Antigravity turns can run indefinitely to completion (matching SDK runtimes like Claude, Grok, and Kimi). Process supervision timeouts and the `--print-timeout` CLI flag are omitted unless `antigravityTurnTimeoutMs` is explicitly configured.
- **System message step handling** (consolidated from #2406): accepts `step_type === "system_message"` in `parser.ts` alongside `user_input` and `checkpoint`, preventing injected out-of-band system messages from being treated as unknown stream protocol errors that falsely fail completed turns.

## Validation

- `pnpm exec vitest run --passWithNoTests src/providers/antigravity/` in `packages/client` (49 passed)
- `pnpm exec biome check` on the changed Antigravity files
- CI all gates green on head `2e600e53`


packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,23 +862,22 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
+                state,
+                sessionCtx,
+                messages,
+                token,
+                expectedSessionId,
+              });
+            }
+            const lifecycleRecoveryReason = drainCancellationReason ?? "antigravity_turn_aborted_or_timed_out";
+            token.retry(messages, lifecycleRecoveryReason);
+            return false;
```
yuezengwu: [P1] Preserve the unsafe-replay fence when preempting an entered turn

After a `run_command` tool event but before a terminal result, `suspend("concurrency_preempted")` now takes this branch without recording an ambiguous-turn key or retaining a provider continuation. Redelivery therefore reaches the normal `resume()` path and writes the same original prompt to a new process with `--conversation`. That preserves conversation history, but does not resume the interrupted operation; a completed write, command or external action can execute again. I reproduced this after first establishing a successful conversation: the base handler creates 2 processes and serializes the mutating prompt once, while this head creates 3 processes and serializes it twice. Keep fail-closed custody for provider-entered work across preemption/retirement (unentered queued work can remain ACK-none), and add a regression that asserts the original mutating prompt is never re-sent.

yuezengwu (changes requested):
Reviewed exact head `83331b6cc3903745bb8783305a55121395b72a47`. Requesting changes for the unsafe prompt replay on concurrency preemption identified inline.

Validation in an isolated worktree: the 18 existing Antigravity handler/parser tests pass; `pnpm check` passes with warnings; Shared and Client typechecks pass. An additional deterministic replay probe passes with the base handler (`84ccd770`) and fails with this PR's handler: after an established conversation enters a mutating turn, suspend + same-row resume sends the original prompt twice. This used a controlled process supervisor, not a live Antigravity account.

L42y:
Follow-up for the unsafe prompt-replay P1, now on exact head `9d82c6eec1716b28e33ec9f8d758e5f36bdbc7c6`:

- A lifecycle cancel after provider entry (operator pause, concurrency preemption, or route retirement) fail-closes that delivery and retains continuation custody. Same-row resume does not spawn another process or serialize the original mutating prompt.
- Unentered queued work remains ACK-none.
- Regression: established conversation, mutating `run_command` turn, plain `suspend("concurrency_preempted")` / shutdown, then same-row resume — the mutating prompt is written once.

Validation on this head: Antigravity parser + handler 18/18 passed; Biome check on the changed files passed. Please re-review the current head.

yuezengwu (commented):
Re-reviewed exact head `9d82c6eec1716b28e33ec9f8d758e5f36bdbc7c6`. The previously reported prompt-replay P1 is fixed in the reproduced scenario: the independent preemption/same-row-resume probe now sends the mutating prompt once and starts no extra provider process. The 18 Antigravity handler/parser tests, the base/head replay probe, `pnpm check` (warnings only), and Shared/Client typechecks pass locally.

Approval is still waiting on current-head CI. [Test Client & Web](https://github.com/first-tree-ai/first-tree/actions/runs/35241559540/job/105271070722) fails at `src/__tests__/session-manager.test.ts:4246`, `chunks fence probes at the wire limit and keeps unproven chunks fenced`, with a 5000 ms timeout (the aggregate Test gate consequently fails). That test passes locally in isolation in 1.65 s; it uses a mock handler, and I have not established that the Antigravity change caused the timeout. This local result does not clear the remote failure. Please investigate/stabilize the failing check and obtain passing CI on the current head before approval.

The PR summary also still says provider-entered preemption/retirement remains ACK-none; please update it to the new fail-closed behavior. Unentered queued work remains ACK-none.

L42y:
Follow-up on exact head `5e66c0c21ae9acb89c5e779b78465c2d5637e71a`:

- PR summary now matches the fail-closed provider-entered lifecycle behavior (unentered queued work remains ACK-none).
- The CI timeout in `chunks fence probes at the wire limit and keeps unproven chunks fenced` was the 101-entry setup calling durable `fence()` (fsync+rename) per id under disk contention. That case now seeds the store in one write, restores real timers after each matrix test, and uses a 20s timeout so the later clear round cannot flake the 5s default.

Local: Antigravity parser/handler 18/18; fence convergence matrix 4 passed. Please re-review once the new head's Test Client & Web check is green.

yuezengwu (approved):
Re-reviewed exact head `5e66c0c21ae9acb89c5e779b78465c2d5637e71a`. The unsafe replay finding is fixed: after a mutating turn enters the provider, preemption and same-row resume do not serialize the original prompt again. Provider-entered cancellation retains fail-closed custody; unentered queued work remains ACK-none.

The CI timeout repair preserves the 50/50/1 reconciliation and failure-round assertions while seeding the version-1 fence fixture in one write and restoring real timers after each matrix test. Fresh validation: 128 tests pass (Antigravity handler/parser, SessionRuntime, and two base/head replay probes); `pnpm check` and Shared/Client typechecks pass. Current remote checks are all successful or skipped. The replay probes use a controlled supervisor; this is not live-provider QA.

L42y:
Follow-up on exact head `aeea68b412ea66b88690821db78843ecddf06aee` for the live quota misclassification:

- `Individual quota reached` plus extra stream-json noise and `Antigravity returned an ERROR result` was matching the Antigravity protocol configuration regex, so the chat notice said "runtime configuration needs attention".
- Extra unknown JSON is now ignored when a terminal result exists. Quota/reset-duration (`Resets in 2h42m27s`) classifies as `provider_capacity` / usage limit so the notice asks to wait for the quota, not to fix configuration.

Validation: Antigravity parser/handler plus provider-retry-policy 70 passed. Please re-review this head (the previous approval was on `5e66c0c21`).

yuezengwu (approved):
Re-reviewed exact head `aeea68b412ea66b88690821db78843ecddf06aee`, including the new quota-classification changes. The observed Individual quota error and reset duration classify as provider capacity, while a valid terminal result tolerates the additional stream noise. The prior unsafe-replay fix still passes the independent base/head probe: the mutating prompt is serialized once.

Fresh isolated validation: 189 relevant tests plus 2 replay probes pass; `pnpm check` and Shared/Client typechecks pass. All current remote checks are successful or skipped. The probes use a controlled process supervisor; no live Antigravity account was exercised.

L42y:
Follow-up on exact head `36f6c72fd26f3d507de7b06a7df317e1aafa0a9f` for the live "No-op webhook event" misclassification:

- An Antigravity `ERROR` result whose body is the agent's own report (for example a no-op GitHub webhook) was posted as "runtime configuration needs attention" because ERROR results were treated as protocol/configuration failures.
- Quota, auth, and true protocol classes stay terminal. Unknown multi-line ERROR bodies are delivered as the turn (including a non-zero process exit).

Validation: Antigravity handler + provider-retry-policy 68 passed. Please re-review this head (previous approval was on `aeea68b41`).

packages/client/src/providers/antigravity/index.ts
```diff
@@ -860,17 +914,42 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
+            scope: "provider_turn",
+            source: "stream",
+          });
+          if (
+            isAntigravityRuntimeFailureCategory(classification.category) ||
+            !looksLikeAgentAuthoredReport(providerErrorText)
+          ) {
+            protocolErrors.push(providerErrorText);
+          } else {
+            agentAuthoredError = true;
```
yuezengwu, line 977: [P2] Keep unclassified provider errors in the failure path

`looksLikeAgentAuthoredReport` accepts any unknown body with a newline or at least 40 characters. Those properties do not establish that an agent authored a normal response. With one valid conversation/result, `status: ERROR`, an empty response, the synthetic provider diagnostic `database disk image is malformed while loading the conversation state`, and process exit code 1, this head calls `token.complete(..., { status: "success" })`, emits a successful turn end and forwards the diagnostic. The same probe on the prior approved head ends in error. This can acknowledge an uncompleted task as successful whenever a real provider failure misses the classifier's finite patterns. Preserve failure handling for unknown ERROR results; limit the no-op compatibility case using positive evidence rather than text length/line count, and retain a regression for an unfamiliar runtime error.

packages/client/src/__tests__/provider-retry-policy.test.ts
```diff
@@ -381,6 +381,40 @@ describe("classifyProviderFailure", () => {
     });
     expect(protocol).toMatchObject({ category: "configuration", reasonCode: "antigravity_protocol_error" });
 
+    const noop = classifyProviderFailure(
+      new Error("No-op webhook event on PR #3910:\n\nEvent: issue_comment created by first-tree-hub-staging[bot]."),
```
yuezengwu: [P2] Remove the legacy bot-name literal from both new fixtures

This line and `providers/antigravity/__tests__/handler.test.ts:957` introduce `first-tree-hub-staging[bot]`, which deterministically trips the existing broad legacy-name sweep. The current `Forbid legacy CLI / env names` job names exactly these two lines and exits 1: https://github.com/first-tree-ai/first-tree/actions/runs/35300859178/job/105463019220. The exact bot spelling is not needed to reproduce an agent-authored no-op report; use a neutral fixture name in both places so the existing CI policy can pass.

yuezengwu (changes requested):
The new ERROR fallback needs a narrower success condition, and the newly added fixtures fail the legacy-name CI gate. See the two inline findings.

Validation on this exact head: 179 existing focused handler/parser/retry/session tests pass; pnpm check, pnpm typecheck and git diff --check pass after completing the frozen dependency install. An independent controlled-process probe supplies one conversation, one ERROR result and exit code 1: the prior approved head ends with an error, while this head acknowledges success and forwards the error body. No live provider or auth session was started.

L42y:
Follow-up on exact head `ccaaf672e6951e6d89cd071a6bd709c6a1c73325` for the ERROR fallback P2s:

- Unknown ERROR results stay in the failure path. The success exception is only a no-op webhook report with positive evidence (`No-op webhook event on PR|Issue #<n>`), not newline/length.
- Regression: `status: ERROR`, empty response, `database disk image is malformed while loading the conversation state`, exit 1 completes as error and does not forward the diagnostic as a turn.
- Fixtures no longer contain `first-tree-hub`.

Validation: Antigravity handler/parser + provider-retry-policy 72 passed; biome check on the changed files. Please re-review this head.

L42y:
Follow-up on exact head `46a44616e604394fbaa09522470c183fb2c6797a`:

A live Review ERROR body (`Reviewed successor head … sign in CTA`) was classified as credential, so the chat asked to run `agy` to sign in. Antigravity credential matching is now diagnostic-only (`authentication required`, `credential is missing`, `run agy once to sign in`, …). Bare "sign in" / "authentication" / "credential" in agent reports stay unknown failures, not a re-login prompt. The no-op webhook success exception is unchanged.

Validation: handler/parser + retry-policy + auth-error-hint 96 passed; biome check on the changed files. Please re-review this head.

L42y:
Follow-up on exact head `7d650090c8275575a55f9362b75092ce2565f85f`:

Live Review turns were already logged in and mid-task: agy streamed the report, then ended with `status: ERROR`. That is not a credential miss. Unknown ERROR is now delivered as the turn when `assistant_delta` was observed (the model already spoke). Empty-response diagnostics (sqlite), quota, and real auth diagnostics stay terminal. The no-op webhook matcher is unchanged.

Validation: handler/parser + retry-policy + auth-error-hint 97 passed; biome check on the changed files. Please re-review this head.

L42y:
Follow-up on exact head `1b620e363d3b6424b752d3ace9c92653db2721ba` from live Review daemon logs:

Root causes (not a login miss — agy was already mid-task):
- `step_type: error_message` was logged as `unknown step type` and ignored as protocol noise.
- `result.status: ERROR` with the report in `result.response` was treated as a provider failure; the response was copied into the classifier, so product words like "sign in" became credential, and exit 1 aborted the turn as unsafe_replay.

Fixes: parse `error_message` as assistant text; only an explicit error payload is a diagnostic; a response body is the turn even when status is ERROR and the process exits 1. Quota/auth diagnostics and empty ERROR results stay terminal.

Validation: handler/parser + retry-policy + auth-error-hint 98 passed; biome check on the changed files. Please re-review this head.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -1291,4 +1378,41 @@ function isReadOnlyTool(name: string): boolean {
...
+ * Unknown ERROR can be the agent's mid-turn output. Deliver it only with
+ * stream evidence (assistant_delta already observed) or the no-op webhook
+ * report. Empty-response diagnostics with no assistant stream stay failures.
+ */
+function shouldDeliverUnknownAntigravityError(input: {
+  providerErrorText: string;
+  streamedAssistant: boolean;
+}): boolean {
+  if (looksLikeNoOpWebhookReport(input.providerErrorText)) return true;
+  return input.streamedAssistant;
```
yuezengwu: [P2] Do not treat earlier progress as proof that a later ERROR succeeded

`streamedAssistant` only proves the model emitted some text before the terminal event; it does not prove the ERROR body is that text or that the task completed. I replayed the prior runtime-error probe with one extra `agent_response` delta (`I will inspect the saved conversation before continuing.`), followed by an ERROR result with empty response, the diagnostic `database disk image is malformed while loading the conversation state`, and exit code 1. This head forwards the diagnostic and acknowledges `{ status: "success" }`; removing only the earlier delta correctly ends in error. The prior approved head ends in error in both cases. Preserve failure handling when the later error is unrelated to the streamed output; earlier progress alone cannot convert an unsuccessful turn into a successful delivery.

packages/client/src/runtime/provider-retry-policy.ts
```diff
@@ -565,12 +575,26 @@ function runtimeSessionProofReason(shape: ErrorShape, text: string): string | nu
...
+  );
+}
+
 function isCredential(
   text: string,
   base: Classification,
   status: number | undefined,
   provider: RuntimeProvider,
 ): boolean {
+  if (provider === "antigravity") return isAntigravityAuthDiagnostic(text);
```
yuezengwu: [P2] Preserve explicit HTTP authentication evidence before narrowing text matching

This early return skips the existing structured status checks below it for every Antigravity error. `Object.assign(new Error("Request rejected by upstream"), { status: 401 })` and the corresponding 403 case now classify as `unknown`; both classify as `credential` on the prior approved head. Narrowing bare words in agent-authored reports should not discard an explicit HTTP authentication status. Preserve structured 401/403 and authoritative auth reason codes, then apply the narrower provider text matcher, with regression tests for these status-bearing errors.

yuezengwu (changes requested):
The forbidden fixture names are fixed, and an empty-response diagnostic without earlier assistant output now stays in the failure path. Two current-head regressions remain: any earlier assistant progress still turns an unrelated runtime ERROR into success, and the new authentication branch bypasses explicit HTTP status evidence.

Validation: 206 existing focused handler/parser/retry/auth/session tests pass; pnpm check, pnpm typecheck and git diff --check pass. Four independent boundary probes pass on the prior approved head. On this head, the no-progress runtime-error control passes, but the same runtime error after one progress delta and the explicit 401/403 cases fail. These are controlled-process/classifier probes, not live-provider acceptance.

L42y:
Follow-up on exact head `f0f86529887e723b5ab531c0ab6d10a84e46ab82` for the two remaining P2s:

- Earlier assistant progress no longer converts a later empty-response runtime ERROR into success. Regression: one `agent_response` delta, then sqlite `database disk image is malformed…`, empty response, exit 1 → error, not forwarded.
- Antigravity credential matching keeps structured 401/403 and auth reason codes, then the narrower text diagnostics. `Request rejected by upstream` with status 401 and 403 classify as credential.

A response body with no error payload is still the turn (live Review reports). Validation: handler/parser + retry-policy + auth-error-hint 99 passed; biome check on the changed files. Please re-review this head.

yuezengwu (approved):
Re-reviewed exact head `f0f86529887e723b5ab531c0ab6d10a84e46ab82`. Both latest P2 findings are fixed: earlier assistant progress no longer converts a subsequent runtime ERROR into success, and structured HTTP 401/403 evidence retains credential classification. The independent preemption probe also confirms the original mutating prompt is serialized only once. No remaining blocking finding in the reviewed changes.

Fresh isolated validation: 207 focused handler/parser/retry/auth/session tests and 5 independent boundary/replay probes pass; `pnpm check`, `pnpm typecheck` and `git diff --check` pass. All 14 current remote checks are successful or skipped. The probes use a controlled process supervisor; no live Antigravity account was exercised.

L42y:
Follow-up on exact head `6010e8fe2c78acd588bd56ba736cc9eb30353b0a` for the live timeout:

A 20-minute abort after the prompt was written was named `TimeoutError`, classified as transient transport, retried, then shown as a network failure. Writing stdin delivers input to agy, so that retry is not effect-free. After prompt delivery: assistant text already obtained is the turn; otherwise fail closed without a TimeoutError name (no transport retry, no network notice).

Validation: Antigravity handler/parser 26 passed; biome check on the changed files. Please re-review this head.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,37 +870,73 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
-          // accidentally create a second Antigravity conversation.
+          // Timeout after the prompt was written is not a transport blip:
+          // Antigravity cannot resume, and TimeoutError would retry as
+          // transient network. Deliver output already obtained; otherwise
+          // fail closed without a TimeoutError name.
           adoptObservedSessionId(sessionCtx, state.sessionIds, expectedSessionId, state.usage);
+          const ids = [...state.sessionIds];
+          const finalText = (state.results[0]?.text || state.text.join("")).trim();
+          const diagnosticFailure = Boolean(state.results[0]?.isError) || state.errors.length > 0;
+          if (ids.length === 1 && finalText && !diagnosticFailure) {
```
yuezengwu: [P2] Keep timed-out progress distinct from a successful terminal result

This condition accepts any assistant delta even when no terminal result exists and a mutating tool is still ACTIVE. With one conversation, the progress text `I will apply the migration, then verify its result.`, an ACTIVE `run_command` event, and no result before the timeout, this head forwards that progress text and calls `token.complete(..., { status: "success" })`. The same controlled-process probe on the prior approved head ends with a consumed `unsafe_replay` error; removing only the progress delta also keeps this head in the error path. A timeout kills unfinished work, so having spoken earlier does not establish completion. Partial text can still be surfaced, but preserve an incomplete/error outcome and unsafe-replay custody instead of acknowledging success merely because text exists.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,37 +870,73 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
+            const id = ids[0];
+            if (!id) throw new Error("Antigravity timeout delivery without conversation ID");
+            adoptSessionId(sessionCtx, id);
```
yuezengwu: [P2] Enforce the expected conversation ID before timeout delivery

`adoptObservedSessionId` above rejects a mismatch with `expectedSessionId`, but its return value is ignored and this unconditional adoption overrides that safeguard. I established conversation A, resumed it, then made the controlled process emit conversation B plus assistant text and hang until timeout. This head forwards B's text, acknowledges success and returns B as the resumed session ID; the prior approved head retains A and settles an error. The normal completion path below still rejects this mismatch. Apply the same identity check before forwarding or adopting anything in the timeout path, and keep the original route on failure.

yuezengwu (changes requested):
The stdin handoff now prevents an unobserved prompt from being retried as a transport timeout, and the previous ERROR/authentication/replay probes still pass. The new timeout-delivery path introduces two separate regressions: arbitrary progress can acknowledge unfinished work as successful, and it bypasses the expected conversation identity check. See the inline findings.

Validation on this exact head: 208 existing focused tests and 5 prior boundary/replay probes pass; `pnpm check`, `pnpm typecheck` and `git diff --check` pass. Three new controlled timeout cases all pass against the prior approved handler; on this head, the no-progress control passes, while progress followed by a pending mutating tool and the resumed-conversation mismatch fail. No live provider account or real tool side effect was exercised.

L42y:
Follow-up on exact head `1852a38b1adc80bea4588f154385981be80c35b5` for the timeout-delivery P2s:

- Timed-out progress is no longer a successful turn. An ACTIVE mutating tool plus earlier assistant text still settles consumed `unsafe_replay`; the text is emitted, not forwarded as completion.
- Timeout delivery no longer bypasses the expected conversation id. A resumed conversation A that then emits B plus text keeps A and settles error.

Prompt-written timeouts still fail closed without a TimeoutError transport retry. Validation: Antigravity handler 24 passed; biome check on the changed files. Please re-review this head.

yuezengwu (approved):
Re-reviewed exact head `1852a38b1adc80bea4588f154385981be80c35b5`. Both timeout findings are fixed: partial progress followed by a pending mutating tool remains a consumed `unsafe_replay` failure, and a mismatched conversation on resume cannot replace the expected session or forward a completion. Prompt-written timeouts still avoid transport retry. The prior ERROR, HTTP-authentication and preemption guards remain intact.

Fresh isolated validation: 209 focused existing tests and 8 independent boundary/replay probes pass; `pnpm check`, `pnpm typecheck` and `git diff --check` pass. Current remote checks are all successful or skipped. Controlled-process probes only; no live provider or real tool side effect was exercised.

L42y:
Follow-up on exact head `61f579045cbc4b90bade093d8595f8d90dac10a2` for the live 14:36 Review banner:

A resumed stream can replay an older ERROR (quota sentence + leftover JSON) and then emit a later SUCCESS. We were settling every historical result, so the current turn was labeled configuration and the banner dumped stream JSON. Only the latest result is this turn; prior error bodies are dropped; parsed stdout is omitted from the failure preview.

Validation: Antigravity handler/parser 28 passed; biome check on the changed files. Please re-review this head.

L42y:
Follow-up on exact head `3a5e50a9924add13555c89f31bfaa495739e8f2d` for Gemini/agy sessions stuck Failed:

Live Gemini logs: `session_eviction_provider_continuation_mismatch` then `notifySessionState(errored)` and `inbox:recover rejected`. A one-delivery continuation from an interrupted Antigravity turn does not match the next message, so eviction resume marked the session failed forever.

Antigravity cannot resume that process. Drop the stale continuation and start a new conversation instead of failing the session. Other providers still fail closed on mismatch.

Please re-review this head together with `61f579045` (latest result only on resume).

L42y:
Follow-up on exact head `5c52aadea698386d6a1d1be9ff60d7807f2a2a73` for Review chat 8d4cfe44:

agy logged `UNAVAILABLE (code 503): No capacity available for model gemini-3.8-flash-high` from 07:15–07:22 and kept retrying with no terminal result. First Tree then hit the 20-minute abort at 07:34. Detect that 503 on stderr, stop waiting, and classify it as provider capacity instead of an unknown timeout.

Validation: Antigravity handler 26 passed; biome check on the changed files. Please re-review this head.

L42y:
Follow-up on exact head `3f3248ec7c4735e16eb97a73df05f83fbc088461` for chat 320dfb58:

Same Google 503 (`No capacity available for model gemini-3.8-flash-high`) as 8d4cfe44. agy retried internally until ~07:42, then we closed the stream. Do not keep that conversation id after a no-result 503, so the next First Tree capacity retry starts fresh instead of resuming the same cascade.

Please re-review with `5c52aadea`.

packages/client/src/runtime/session-runtime.ts
```diff
@@ -2276,6 +2276,27 @@ export class SessionRuntime {
     this.projection.persistRegistry();
   }
 
+  /**
+   * Antigravity cannot resume an interrupted process. A one-delivery
+   * continuation that does not match the current message must not mark the
+   * session Failed forever; drop it so the next turn starts a new conversation.
+   */
+  private clearUnresumableAntigravityContinuation(entry: SessionEntry, reason: string): boolean {
+    if (this.runtimeProvider() !== "antigravity") return false;
```
yuezengwu: [P2] Keep concrete-provider recovery policy behind the provider boundary

The new literal check in the generic SessionRuntime deterministically fails `provider-boundary-guard.test.ts` ("keeps migrated generic client files free of concrete provider literals and handler imports"). This is the actual failure in the current Test Client & Web job, and the same test fails in the isolated checkout. Expose the continuation-recovery behavior through the provider contract/capability and keep this runtime path provider-neutral; please preserve the architecture guard rather than bypassing it.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -851,32 +942,61 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
         const ids = [...state.sessionIds];
         const protocolErrors: string[] = [];
-        if (ids.length !== 1) protocolErrors.push(`expected one conversation ID, observed ${ids.length}`);
-        if (expectedSessionId && ids[0] !== expectedSessionId) {
-          protocolErrors.push(
-            `resume conversation mismatch: expected ${expectedSessionId}, observed ${ids[0] ?? "none"}`,
-          );
+        const result = state.results.at(-1);
+        const id =
+          expectedSessionId && state.sessionIds.has(expectedSessionId)
```
yuezengwu: [P2] Bind the accepted terminal result to the resumed conversation

Seeing the expected ID anywhere in `state.sessionIds` now overrides the identity of the latest result. Reproduction: establish conversation A, resume A, then receive `init(A)` followed by `result(SUCCESS, conversation_id=B, response="answer-from-B")`. Current head forwards B's answer and completes A's delivery with success; the previously approved 1852a38b rejects the same stream. The existing mismatch test only covers a stream containing B alone. Preserve the latest result's conversation identity and verify it against the expected ID even when historical/init events also contain A.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,37 +887,48 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
           }
           const lifecycleRecoveryReason = drainCancellationReason ?? "antigravity_turn_aborted_or_timed_out";
           token.retry(messages, lifecycleRecoveryReason);
           return false;
         }
+        const capacityDiagnostic = antigravityCapacityDiagnostic([outcome.stderrTail, ...state.errors].join("\n"));
+        if (capacityDiagnostic && !state.sawUnsafeTool && state.results.length === 0) {
+          // Do not keep the conversation id: resuming the same cascade would
+          // sit in agy's internal 503 retry loop again.
+          return settleFailure({
```
yuezengwu, line 916: [P2] Clear an established conversation before the capacity retry

Skipping ID adoption only helps the first turn. If a prior turn already established conversation A, this branch leaves `providerSessionId` as A; `settleFailure` does not clear it. In a controlled probe, after A succeeds, a follow-up emits no result and reports the 503 capacity diagnostic. The token is retried with `provider_overloaded`, but reinjecting that delivery still spawns agy with `--conversation A`. The identical fresh-session case correctly omits `--conversation`. Reset the existing conversation/recovery identity as part of the safe fresh-retry transition, otherwise the follow-up case still resumes the same cascade this fix is intended to avoid.

yuezengwu (changes requested):
Re-reviewed exact head `3f3248ec7c4735e16eb97a73df05f83fbc088461`. Three current-head issues need changes: the generic runtime branch deterministically fails the provider architecture guard; a resumed turn can accept another conversation's terminal result; and the new 503 retry path retains an already-established conversation ID. Details and reproductions are inline.

Fresh isolated validation: `pnpm check`, `pnpm typecheck` and `git diff --check` pass. The focused existing suite has 228 passes and the same one architecture-guard failure as CI run 35321269263. All 8 retained independent timeout/ERROR/auth/preemption probes pass. Four additional cases reproduce the terminal-identity regression (previous head passes, current head fails) and show fresh-session capacity recovery passes while existing-session recovery fails. These are controlled process probes, with no live provider or real tool side effect.

L42y:
Follow-up on exact head `64b1e10899526405d5599bc36d4f5997238e27f3` for the three P2s:

- SessionRuntime no longer names a concrete provider. Fresh-start on mismatched continuation goes through `startsFreshOnMismatchedContinuation` in the retry-policy capability (same set as unsafe-replay notice custody). The architecture guard passes.
- Resume identity uses the latest result's conversation id. `init(A)` plus `result(B, SUCCESS)` is a mismatch, not a successful delivery of B on A.
- A follow-up 503 clears the established session id before retry, so the next spawn does not pass `--conversation`.

Validation: handler 28, provider-boundary-guard 18, retry-policy 51; biome check on the changed files. Please re-review this head.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,37 +887,50 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
           }
           const lifecycleRecoveryReason = drainCancellationReason ?? "antigravity_turn_aborted_or_timed_out";
           token.retry(messages, lifecycleRecoveryReason);
           return false;
         }
+        const capacityDiagnostic = antigravityCapacityDiagnostic([outcome.stderrTail, ...state.errors].join("\n"));
+        if (capacityDiagnostic && !state.sawUnsafeTool && state.results.length === 0) {
+          // Drop an already-established conversation so the capacity retry
+          // cannot spawn `--conversation` for the same 503 cascade.
+          providerSessionId = null;
```
yuezengwu, line 914: [P2] Publish the replacement conversation back to SessionRuntime

Clearing these handler-local IDs fixes the next spawn but leaves the runtime/session-registry ID unchanged. Reproduction: establish A, inject a follow-up that gets the no-result 503, retry it without `--conversation`, and successfully complete it in fresh conversation B. `adoptSessionId` only calls `sessionCtx.replaceSessionId` when `pendingSyntheticId` exists; an already-established A has no such marker, so no replacement callback fires and the runtime-held ID remains A. A subsequent cold handler resume using that retained ID sends `--conversation A` again (the fresh-session control correctly records and resumes B). SessionRuntime persists `entry.claudeSessionId`, so this loses the recovered conversation on eviction/restart. Keep the runtime mapping synchronized when dropping and adopting the replacement identity; extend the regression through successful retry into a distinct conversation and cold resume, rather than stopping at the retry spawn arguments.

yuezengwu (changes requested):
Re-reviewed exact head `64b1e10899526405d5599bc36d4f5997238e27f3`. The provider-boundary CI failure and latest-result conversation mismatch are fixed, and the immediate capacity retry now omits the old `--conversation` argument. One part of capacity recovery remains incomplete: the successful replacement conversation is not published back to SessionRuntime, so later cold resume still targets the old conversation. See the inline reproduction.

Fresh validation: all 231 focused existing tests, `pnpm check`, `pnpm typecheck` and `git diff --check` pass; remote checks are successful or skipped. The 11 retained independent boundary cases pass. Four additional recovery checks pass both fresh-session controls but fail the existing-session runtime mapping and subsequent cold-resume cases. Controlled process probes only; no live provider, actual client restart or real tool side effect.

L42y:
Follow-up on exact head `f3f93d0cc1e3b04816408b582144481013c07c69` for the remaining capacity-recovery P2:

After a no-result 503, the handler still remembers the dropped conversation A. When the retry succeeds in B, `adoptSessionId` now calls `replaceSessionId(B)` even when A was a real established id (not only a pending synthetic). Regression: A → 503 retry without `--conversation` → success in B publishes B; a new handler cold-resuming B sends `--conversation B`, not A.

Validation: Antigravity handler 28 passed; biome check on the changed files. Please re-review this head.

yuezengwu (approved):
Re-reviewed exact head `f3f93d0cc1e3b04816408b582144481013c07c69`. The remaining capacity-recovery finding is resolved: after established conversation A hits a no-result 503, retry starts fresh, success in B calls `replaceSessionId(B)`, and a cold handler using the runtime-held identity resumes B rather than A. Both fresh-session and existing-session controls pass. The prior architecture, terminal identity, timeout, ERROR/authentication and preemption regressions remain covered.

Fresh isolated validation: all 231 focused existing tests and all 15 independent boundary/recovery probes pass, along with `pnpm check`, `pnpm typecheck` and `git diff --check`. Current remote checks are successful or skipped. Controlled process tests only; no live provider, actual Client restart or real tool side effect was exercised.

bestony (approved):
Reviewed exact head `2e600e53` (`fix/antigravity-pause-and-stream`), which supersedes and consolidates closed #2406. Scope: the two commits added since the last approval — `f93b4e33` (turn-timeout default removal) and `2e600e53` (`system_message` step acceptance).

## system_message step acceptance (consolidated from #2406)

- Confirmed the reported failure against `main`: at `84ccd770` the success gate required `protocolDiagnostics.length === 0` with no `!result` escape, so any `system_message` step poisoned a completed turn with "unsupported or malformed Antigravity stream" — a false "runtime configuration needs attention" failure.
- The one-line parser change is correct and consistent with the existing `user_input` / `checkpoint` handling; the new unit test pins all three accepted step types.
- Note: at this branch's head, `61f57904` already added the `&& !result` guard, so completed turns are shielded regardless. This fix still earns its place: `system_message` is a known, intentional step type and should never count as a protocol diagnostic — unknown diagnostics remain reserved for genuinely unknown types, preserving the parser's documented fail-closed stance.

## Turn-timeout default removal (f93b4e33)

- Implementation is clean: `turnTimeoutMs` is optional through `buildAntigravityTurnArgs` → supervisor → child registry, so no kill timer and no `--print-timeout` are armed by default; explicit positive config still yields `--print-timeout` with the 1m floor. Tests cover both branches.
- Non-blocking: the parity claim in the commit message lists Codex/Cursor, but `opencode`, `deepseek-harness`, and `amp` still default to `DEFAULT_TURN_TIMEOUT_MS = 20 min` in this repo. Parity currently holds only for SDK-based providers (Claude/Grok/Kimi). Worth correcting the rationale or planning the same change for the other process-supervised providers.
- Non-blocking: with no timeout armed, an agy hang that never prints the capacity diagnostic (the live stderr detection terminates the known 503 retry loop) now waits indefinitely until a human pause or daemon shutdown cleanup. That is a reasonable product trade-off (turns run to completion), but the PR description should say so — it currently still describes only the pause-settlement and stray-stdout fixes, not these two new behaviors.

## Validation

- Local: `vitest run` on the Antigravity `parser` / `binary` / `handler` suites at this head — 38/38 pass.
- CI: all gates green on the current head, including Test Client & Web (6m32s) and Lint & Type Check.

Approved. The two notes above are documentation/rationale follow-ups, not blockers.

yuezengwu (approved):
Re-reviewed exact head `2e600e533bc907ee43943a4e1842c5365fb14c90` against `main@84ccd7704ca128d52a78957c8fe5657e86bb659e`. The consolidated timeout and system-message changes introduce no blocking finding in this review.

Fresh isolated validation passed 249 existing tests and 15 independent regression probes, including lifecycle/preemption replay protection, explicit timeout outcomes, terminal identity, capacity-retry identity publication and cold resume. Client build, full typecheck, `pnpm check` (warnings only) and `git diff --check` also passed. Controlled process/stream validation only; no live provider or actual Client restart acceptance is claimed.

packages/client/src/providers/antigravity/index.ts
```diff
@@ -807,37 +891,54 @@ export const createAntigravityHandler: HandlerFactory = (config) => {
...
-              turnGeneration,
+              expectedSessionId,
             });
           }
           const lifecycleRecoveryReason = drainCancellationReason ?? "antigravity_turn_aborted_or_timed_out";
           token.retry(messages, lifecycleRecoveryReason);
           return false;
         }
+        const capacityDiagnostic = antigravityCapacityDiagnostic(
+          [outcome.stderrTail, outcome.stdoutTail, ...state.errors].join("\n"),
```
yuezengwu: [P2] Do not treat assistant stdout content as a capacity rejection

`stdoutTail` contains the complete JSON stream, including ordinary `agent_response` text. A controlled stream with `init(A)`, an assistant delta saying "I found the text No capacity available in the incident log; inspecting it now.", then `sqlite3: database or disk is full` on stderr and exit 1 (no result) now calls `token.retry(..., "provider_overloaded")` and never completes the consumed `unsafe_replay` outcome. The identical control without the quoted phrase correctly consumes the interrupted delivery; a genuine standalone capacity diagnostic still retries. This branch clears the conversation and replaces the observed provider activity with `false`, so a quotation can authorize replay after provider entry even though an unreported side effect may already have occurred. Restrict capacity detection to actual diagnostic/error events, excluding assistant/user/tool stream payloads, and retain a regression for the quoted-text case.

packages/client/src/__tests__/provider-retry-policy.test.ts
```diff
@@ -381,12 +381,114 @@ describe("classifyProviderFailure", () => {
...
+    const prematureZeroResults = classifyProviderFailure(
+      new Error("expected one terminal result event, observed 0\nexit 1"),
+      { provider: "antigravity", scope: "provider_turn", source: "stream" },
+    );
+    expect(prematureZeroResults.category).toBe("unknown");
+    expect(prematureZeroResults.reasonCode).not.toBe("antigravity_protocol_error");
+
+    const zeroEventsWithStderrCrash = classifyProviderFailure(
+      new Error(
+        [
```
yuezengwu: [P2] Format the three changed files so the required check can pass

The current Lint & Type Check job fails at `pnpm check`, and the isolated checkout reproduces exactly three formatter errors: this test, `providers/antigravity/__tests__/handler.test.ts`, and `providers/antigravity/index.ts`. A focused `pnpm exec biome check` on these files reports the same differences. Please format them and commit the result; rerunning unchanged code will fail again. CI: https://github.com/first-tree-ai/first-tree/actions/runs/35736059655/job/106773552511

yuezengwu (changes requested):
Reviewed exact head `076532b5d7d7c01414c0504e1ff7d908d377fc95` against current `main@13f2a38b7b8a25789d4430b9dc968864b26611d1`. The new stdout capacity scan can turn ordinary assistant content into permission to replay an interrupted delivery; the inline reproduction fails on this head. Please preserve replay custody when the matched words come from stream payload content. CI also fails deterministically on three formatting errors introduced in this update.

Fresh isolated validation: 250 targeted repository tests, 15 retained independent regression probes and two new controls pass; the new quoted-capacity regression fails. Client build, full typecheck and `git diff --check` pass. `pnpm check` fails with the same three formatting errors as the Lint & Type Check job. The current main change has no file overlap and merges cleanly; no combined-tree full suite, live provider or actual Client restart acceptance was run.

yuezengwu (approved):
Re-reviewed exact head `ed7b460093aa96e499379e4ff5143eb08e0c6a41` against current `main@13f2a38b7b8a25789d4430b9dc968864b26611d1`. Both findings from my preceding review are resolved. The same quoted-capacity/crash reproduction now consumes the interrupted delivery as `unsafe_replay` without retrying; raw stdout cannot authorize replay, while actual stderr and parsed error-event capacity diagnostics still retry. All three formatting errors are fixed. No blocking finding remains in this review.

Fresh isolated validation passed 251 targeted repository tests and 20 independent regression/control probes, including lifecycle/preemption, explicit timeout, terminal identity, capacity recovery publication and cold resume. Client build, full typecheck, `pnpm check` and `git diff --check` passed; current remote checks are successful or skipped. Current main has no changed-file overlap and merges cleanly. Validation used controlled processes/streams; no combined-tree full suite, live provider or actual Client restart acceptance is claimed.

PR #2402, opened 2026-09-17
fix(client): recover Cursor turns stuck on resume-no-progress

L42y:
## Summary

- Cursor can exhaust its in-turn resume loop and report `RetriableError: Agent turn stopped after repeated resume attempts made no progress`. First Tree then retried the same `--resume` session and surfaced that as an unknown terminal failure.
- The Cursor handler now abandons the poisoned provider session. When the attempt is still replay-safe, it retries the same delivery without `--resume` (Codex stale-rollout recovery). When a write tool or assistant output already happened, it stops and the next turn in this chat starts fresh instead of re-entering the stuck id.
- Shared retry classification treats the phrasing as `cursor_resume_stuck` so unknown retries cannot re-bind `--resume` of that session. The durable runtime notice tells the operator to send again in the same chat.

## Validation

- [x] `pnpm exec biome check` on the changed client files
- [ ] `pnpm check` (repo-wide not rerun)
- [ ] `pnpm typecheck`
- [x] focused Vitest: `resume-stuck`, Cursor `handler-engine`, `provider-retry-policy`, `runtime-notice` (100 passed)
- [ ] `pnpm test`

## Change Surface

## Notes

- docs or tests updated to match: Cursor handler, retry-policy, and runtime-notice coverage for resume-stuck recovery
- follow-up work: none


yuezengwu (approved):
Reviewed exact head `e1e74cd788624b8907b2917115a0bea4e8addbdb`. Replay-safe resume-stuck failures recover once without the poisoned `--resume` id; assistant output or a write-tool effect stops automatic replay. Independent probes also confirm stderr-only recovery, read-only-tool recovery, and a fresh start after an unsafe failure across a handler restart using the replacement pending id.

Fresh isolated validation: 234 Cursor/shared retry/runtime tests plus 3 independent recovery probes pass; `pnpm check` and Shared/Client typechecks pass. All current remote checks are successful or skipped. This validates the handler using a controlled CLI transport, not a live Cursor session.

PR #2403, opened 2026-09-18
feat(client): discover Antigravity models from agy models

L42y:
## Summary

- The agent Model picker had no Antigravity catalog (`discoverProviderModels("antigravity")` always returned `unavailable`), so a bound computer showed no selectable models.
- The daemon now runs the documented `agy models` command on the same resolved `agy` binary the handler uses, parses slug/label text (and JSON when `--output-format json` works), and degrades to unavailable when the CLI is missing or the output is unparseable.
- Web help copy matches Cursor/Grok: options come from this computer's Antigravity CLI.

## Validation

- `pnpm exec vitest run --passWithNoTests src/providers/antigravity/__tests__/discover-models.test.ts src/__tests__/discover-models.test.ts` in `packages/client` (12 passed)
- `pnpm exec biome check` on the changed files


packages/client/src/providers/antigravity/discover-models.ts
```diff
@@ -0,0 +1,195 @@
...
+    (async (bin, processEnv) => {
+      const json = await runCommand(bin, ["models", "--output-format", "json"], {
+        timeoutMs: ANTIGRAVITY_MODELS_TIMEOUT_MS,
+        env: processEnv,
+      });
+      if (json.ok && parseAntigravityModelsOutput(json.stdout).models.length > 0) {
+        return { ok: true, stdout: json.stdout, stderr: json.stderr };
+      }
+      const text = await runCommand(bin, ["models"], {
+        timeoutMs: ANTIGRAVITY_MODELS_TIMEOUT_MS,
```
yuezengwu: [P2] Share one deadline across both catalog attempts

The JSON attempt can consume 20 seconds, and this fallback is then granted another 20 seconds. The models endpoint uses waitForClientReply with DEFAULT_CLIENT_REPLY_TIMEOUT_MS = 25_000, so a slow JSON attempt followed by a valid text response can still return HTTP 504 and leave the picker without models. I reproduced this with the production server waiter and fake timers: JSON fails at 20 seconds, text succeeds 6 seconds later, but the waiter has already expired and resolveClientReply returns false. Give the fallback only the remaining overall budget (below the server deadline), or avoid retrying a timed-out probe; add a test spanning both attempts and the caller's deadline.

yuezengwu (changes requested):
The discovery fallback can outlive the server's model-catalog request. Please keep the two attempts within one deadline so a valid fallback result can reach the picker.

Validation on this head: pnpm check, pnpm typecheck and git diff --check pass. Twenty focused tests pass, including two independent fake-clock probes using the actual server reply waiter: a 20-second JSON failure followed by a successful 6-second text lookup misses the 25-second server deadline, while an immediate JSON rejection followed by the same lookup succeeds. No live provider/auth session was started.

L42y:
Follow-up on exact head `9066a8148dbe47f06be7d8b5da09fae7ad543683` for the shared-deadline P2:

- JSON and text `agy models` probes now share one 20s budget (below `DEFAULT_CLIENT_REPLY_TIMEOUT_MS` = 25s).
- A timed-out JSON probe does not start text fallback. A fast JSON rejection leaves the remaining budget for text.
- Tests: JSON spending the budget skips text and finishes under 25s; immediate JSON rejection + 6s text succeeds; a 5s JSON miss gives text 15s remaining.

Validation: Antigravity + dispatcher discovery tests 15 passed. Please re-review this head.

yuezengwu (approved):
The shared-deadline finding is resolved on this exact head. JSON and text attempts share 20 seconds; an exhausted JSON attempt does not start fallback, and a partial attempt leaves only the remaining budget.

Validation: 21 existing focused discovery tests plus 3 independent fake-clock probes using the actual server reply waiter pass. Those probes cover an exhausted JSON attempt returning unavailable before the 25-second deadline, an immediate rejection followed by a successful text lookup, and a partially consumed budget followed by a successful lookup. pnpm check, pnpm typecheck and git diff --check pass. No live provider/auth session was started.

PR #2404, opened 2026-09-18
fix(chat): stop tagging from inheriting a Failed status

L42y:
## Summary

Tagging an agent (including Kimi Code) could flash **Failed** for a couple of seconds before **Working**. That is a false status, not a crash.

A mention predictively marks the `(agent, chat)` session `active` before the Client reports this turn. The previous D-axis (`working` / `error` / `blocked`, or a NULL stamp) was left in place. Composite status treats stale in-flight runtime as recovery, so the chip said Failed until the first fresh `working` frame.

This change stamps `idle` on an active wake unless the session is already **fresh working**. Dead-client stale in-flight still shows Failed until someone wakes the agent. A second message into a live turn does not flash Idle.

## Test plan

- [x] `packages/server` unit tests for: first active write stamps idle; fresh working is a no-op; stale working and leftover error idle on the next wake; composite status after that wake is `ready`, not `failed`.
- [ ] Server Vitest was not run in this environment (no Docker / `CI_DATABASE_URL`). CI should run `pnpm --filter @first-tree/server test`.
- [ ] Manual: mention a Kimi Code agent whose last turn is idle or stale. Chip should go Idle/Ready → Working, never Failed. Mention during a live Working turn should stay Working. A genuinely dead in-flight session with no new mention should still show Failed.


packages/server/src/services/chat/sessions/activity.ts
```diff
@@ -59,33 +61,40 @@ export async function upsertSessionState(
     // client's `heartbeat` frame is the canonical lastSeenAt refresh
     // path (see presence.ts:touchAgent), so dropping the lastSeenAt
     // side-effect here is safe.
+    //
+    // Active writes stamp `idle` rather than leaving a prior D-axis in place.
+    // A mention predictively marks the session active before the Client has
+    // reported this turn; inheriting stale `working`/`error` made that wake
+    // read as Failed until the first fresh `working` frame. Fresh working is
```
yuezengwu: [P2] Preserve the legacy runtime fallback on predictive activation

A non-null `runtimeStateAt` is still the capability discriminator in `computeWorking` / `computeErrored`: once it is present, both functions stop consulting legacy tool events and agent-global errors. Stamping it on the first server-predicted active write therefore marks a client as having reported per-chat runtime before it has sent any such frame. In a real PostgreSQL probe, an old-client session created through this function stays Ready both after a fresh tool_call and after a new presence.runtime_state=error reported after the wake; the base writer produces Working and Failed respectively. Unlike the temporary stale-status flash, this persists for clients that never send session:runtime. Keep the distinction between a predictive idle reset and evidence of actual per-chat runtime support. The existing fallback tests insert rows directly with a NULL stamp, so they miss this regression.

yuezengwu (changes requested):
The intended wake and fresh-working cases pass, but the first predictive active write disables the still-supported old-client runtime fallback. That hides both real activity and newly reported errors; see the inline finding.

Validation on this exact head: frozen install, `pnpm check`, `pnpm typecheck`, `git diff --check`, and 73 existing activity/status tests pass. Six independent cases ran against disposable PostgreSQL: the base writer passes all three, while this writer passes the fresh-modern-working control and fails both legacy activity/error cases. Current remote CI is successful or skipped. No production database or author branch was changed.

packages/server/src/services/chat/sessions/activity.ts
L42y: Addressed. A predictive active write no longer stamps `runtime_state_at`. That NULL remains the old-client discriminator, so legacy activity still reads Working and a later `presence.runtime_state=error` still reads Failed.

The Failed-on-tag fix is now a separate update: it only idles an **already-stamped** leftover `working`/`error`/`blocked` D-axis. Tests go through `upsertSessionState` for both the modern wake and the two legacy fallback cases.

yuezengwu (approved):
Re-reviewed exact head `60d30fa2154a70c85c481f9068bd5414aaf675b2`. The predictive active write now preserves the NULL per-chat runtime stamp, so legacy tool activity remains Working and a newly reported legacy presence error remains Failed. The separate wake repair still clears stale modern runtime while preserving fresh Working. The previous compatibility finding is resolved.

Fresh isolated validation: 76 existing activity/status tests and 3 independent PostgreSQL probes pass; `pnpm check`, `pnpm typecheck` and `git diff --check` pass. Current remote checks are successful or skipped. Tests used disposable PostgreSQL, not a production database; no live client/UI acceptance is claimed.

PR #2405, opened 2026-09-19
fix(client): align Antigravity turn timeout with other runtimes

L42y:
## Summary
- Removes the hardcoded default 20-minute turn timeout from the Antigravity provider handler.
- Turns now run indefinitely until completion by default, matching Claude Code, Grok, Codex, Cursor, and Kimi Code runtimes.
- Omits the `--print-timeout` flag when spawning `agy` unless `antigravityTurnTimeoutMs` is explicitly configured as a positive number.
- Verified with unit and handler test suites in `packages/client`.


bestony (approved):
Reviewed the full diff (4 files, +59/−18) against the provider stack, ran the affected suites locally, and checked CI on this head.

**Verified**
- No hidden default: `ProviderProcessSpec.timeoutMs` is optional and the default supervisor only forwards it when truthy, so omitting it really means no automatic turn kill. External abort paths (user cancel / client switch / shutdown) are unchanged.
- Parity claim is accurate: `claude` (non-TUI), `grok`, `codex`, `cursor`, and `kimi-code` have no client-side turn-abort timer; after this change Antigravity joins them. `opencode`, `amp`, and `deepseek-harness` still default to a 20-minute client-side abort — pre-existing divergence, worth a separate follow-up to align or document, but out of scope here.
- Configured path preserved: a positive `antigravityTurnTimeoutMs` still yields both `--print-timeout <n>m` and the process/abort timeout; handler tests cover 5s and 50ms cases.
- Tests: `binary.test.ts` + `handler.test.ts` 20/20 pass locally. CI on this head is fully green (lint & type check, Test Client & Web, Test Server, CLI suites).
- No docs/config-schema references to `antigravityTurnTimeoutMs` go stale with the removed default.

**Nits (non-blocking)**
- `runProcess` guards the spread with truthiness (`input.timeoutMs ? ...`) while `buildAntigravityTurnArgs` checks `> 0`; equivalent today because the value is sanitized at construction, but one guard style would make the invariant obvious.
- "agy runs indefinitely without `--print-timeout`" is asserted, not verifiable without the binary; a quick manual check would confirm the assumption.

Approving — nothing here blocks merge.

yuezengwu (approved):
Reviewed 643e72f4 independently. The default handler, supervisor and registry paths no longer impose a turn timeout; positive explicit configuration and lifecycle abort/settlement remain intact. The 21 affected tests, scoped shared/client build and typecheck, repository checks and diff check pass. No blocking finding in this change. The actual agy binary was unavailable locally, so this verifies the First Tree timeout behavior without claiming live provider acceptance.

L42y:
Consolidated into #2401.

PR #2406, opened 2026-09-20
fix(client): accept system_message step in Antigravity stream parser

L42y:
## Summary
- Injects or system notifications in Antigravity emit `step_update` events with `step_type: "system_message"`.
- Previously, `parser.ts` only recognized `user_input` and `checkpoint` as unhandled steps to preserve, rejecting `system_message` as an unknown step type.
- This caused completed turns to fail with an unsupported stream protocol error, resulting in a false "runtime configuration needs attention" terminal failure that captured the successful answer in the failure preview.
- Accepts `step_type === "system_message"` in `parser.ts` and adds unit test coverage.


L42y:
Consolidated into #2401.

PR #2407, opened 2026-09-20
fix(client): settle OpenCode lifecycle abort without treating shutdown as provider crash

L42y:
## Problem
When `first-tree-staging.service` was restarted (e.g. via `systemctl --user restart first-tree-staging.service` during a turn), `sessionRuntime.shutdown("manager_shutdown")` aborted the in-flight OpenCode subprocess with `settleProviderEntered: true`. Because OpenCode handler did not implement lifecycle consumption, the abort was routed to `settleFailure`, which treated the interrupted subprocess as an unexpected crash. Because the turn had run `bash`, replay safety was marked `unsafe`, causing `settleFailure` to emit a provider failure terminal event and post a misleading error notice to the chat:
`OpenCode could not run this turn after retrying a transient provider or network failure. Retry when the provider is available. Original provider message: OpenCode turn aborted or timed out before a safe terminal event`

In addition, OpenCode provider retained a hardcoded 20-minute default turn timeout (`DEFAULT_TURN_TIMEOUT_MS = 20 * 60_000`), unlike other runtimes (Claude, Grok, Codex, Cursor, Kimi Code, and Antigravity via #2405) where turns run to completion without an arbitrary 20-minute ceiling unless configured.

## Approach
1. Incorporates and rebases the turn-abort improvements from #2367 onto current `origin/main` (handler-scoped abort records, pre-spawn and post-listener admission checks, and silent supersession).
2. Adds `settleLifecycleConsumedTurn`: when an in-flight turn is interrupted by runtime lifecycle (operator suspend, graceful drain, or manager shutdown), mutating turns (`state.sawUnsafeTool`) or provider-entered turns during `settleProviderEntered` settle cleanly as `unsafe_replay` without posting a spurious provider failure card.
3. For non-mutating turns interrupted by lifecycle, cleanly retries the delivery token (`token.retry`) so the turn resumes upon restart.
4. Removes the hardcoded default 20-minute turn timeout (`DEFAULT_TURN_TIMEOUT_MS`) to align with other runtimes. If `opencodeTurnTimeoutMs` is configured, it remains respected.
5. Adds unit tests covering lifecycle shutdown settlement, non-mutating suspend retry, and default process timeout omission.

## Validation
- 72/72 OpenCode provider tests pass (`packages/client/src/providers/opencode/__tests__/`)
- Provider boundary guard tests pass (18/18)
- Biome lint and format checks pass with 0 errors


bestony (changes requested):
**Verdict: request changes — CI is red on this head (typecheck), everything else checks out.**

## Blocking: `Lint & Type Check` fails on `63e6f5c83` (run 35489274669)

Reproduced locally with the repo's TS 5.9.3 (`cd packages/client && tsc --noEmit`):

1. **`packages/client/src/providers/opencode/index.ts(1267,28)` and `(1267,61)` — TS2339: property does not exist on type `never`.**
   In the `runTurn` catch block, `observedState` is only assigned inside the async IIFE (`observedState = state`), never in the `try` flow itself. TS's catch-clause control-flow analysis narrows a `let` with no assignment visible in the `try` to `never`, so `observedState?.sawUnsafeTool` / `observedState?.sawProviderActivity` fail to typecheck (vitest doesn't typecheck, which is why Test Client & Web is green).
   Minimal verified fix: reset the narrowing with an assertion before the check, e.g.
   ```ts
   const captured = observedState as TurnState | null;
   if (captured?.sawUnsafeTool || (captured?.sawProviderActivity && settleProviderEntered)) { ... }
   ```
   (I verified `as` compiles clean under the same tsconfig/strictness; an explicit function boundary, e.g. a local `currentObservedState()` reader, would also work.)

2. **`packages/client/src/providers/opencode/__tests__/handler.test.ts(2095,19)` and `(2135,19)` — TS2304: cannot find name `SessionEvent`.**
   The two new lifecycle tests declare `const events: SessionEvent[] = []` but the file never imports the type. Add `type SessionEvent` to the existing `@first-tree/shared` import (same pattern as `packages/client/src/providers/amp/__tests__/handler.test.ts:5`).

## What I verified (all green locally)

- `packages/client/src/providers/opencode/__tests__/` — **72/72 pass** (matches the PR claim).
- `packages/client/src/__tests__/provider-boundary-guard.test.ts` — **18/18 pass**.
- Biome lint + format on the four changed files — clean, no fixes needed.
- The SessionRuntime contract wiring is correct: `HandlerShutdownOptions.settleProviderEntered` is documented in `runtime/handler.ts`, set by operator suspend (`session-runtime.ts:3283`) and by active-slot graceful drain (`session-runtime.ts:1345`), and the OpenCode handler consumes it exactly there. The `settleProviderEntered` reset happens only after `await Promise.all([currentTurnPromise, currentDrainPromise])`, so the in-flight turn always observes the flag before it is cleared — ordering is right.
- Classification/diagnostic separation in `settleFailure` matches `ProviderAttempt.recordSignal` semantics: `diagnostic` signals only feed `diagnostics`, the primary classification stays on the stable timeout message, and the `transient_transport` categorization is asserted end-to-end in `turn-abort.test.ts`.
- The consumed path settles via `token.complete(consumedErrorOutcome("unsafe_replay"))`; the runtime's `completeTurn` treats a consumed outcome as a normal finish (no provider-failure card) absent a pending failure notice — consistent with the problem statement's fix.

## Non-blocking notes

- The non-mutating branch `state.sawProviderActivity && settleProviderEntered` (provider-entered, no unsafe tool, graceful drain) has no direct test — the shutdown test covers it only through the `sawUnsafeTool` branch, and the suspend test covers `settleProviderEntered` unset. A small test with a `read`-tool turn plus `shutdown(..., { settleProviderEntered: true })` would pin the claimed behavior.
- The catch-block abort check doesn't consult the abort record, so a superseded (silent) turn that rejects would take the lifecycle settle/retry path instead of staying silent, unlike the post-outcome path. Low reachability given SessionRuntime delivery serialization, but worth a comment or record check for consistency.
- `settlementPolicyForOpenCodeTurnAbort`'s `superseded` / `session_inactive` / `lifecycle` cases are unreachable from current call sites (only `timeout` reaches `settleFailure`) — fine as defensive completeness.
- Silent supersession leaves the superseded token unsettled (codified by test). That's correct only because SessionRuntime never overlaps deliveries on one handler (active-slot + in-flight-transition gating); a future overlapping path would hang the ledger row until server redelivery.

L42y:
Thanks @bestony for the careful review and reproducing the typecheck issues!

Addressed all blocking and non-blocking points in commit `7434a229f`:
1. **TypeScript narrowing in catch block**: Reset `observedState` narrowing via `const captured = observedState as TurnState | null;` and checked `captured && (...)` along with `(observedState as TurnState | null)` in `settleFailure`.
2. **Missing `SessionEvent` import in tests**: Added `type SessionEvent` to the `@first-tree/shared` import list in `packages/client/src/providers/opencode/__tests__/handler.test.ts`.
3. **Silent supersession in catch block**: Added `takeTurnAbortRecord(turnGeneration)` / `inferOpenCodeTurnAbortRecord(...)` and returned `false` on `record.disposition === "silent"`.
4. **Test coverage for non-mutating provider-entered shutdown**: Added `settles manager shutdown of an in-flight read-only turn as unsafe_replay when settleProviderEntered is true` using the `read` tool and verifying it settles cleanly as `unsafe_replay` without fake provider failure errors.

Verified locally:
- `cd packages/client && tsc --noEmit` exits 0 clean.
- Biome check passes with 0 issues.
- Vitest OpenCode test suite passes 73/73 tests.

bestony (approved):
**Verdict: approved.** Commit `7434a229f` resolves both blocking typecheck failures and addresses the non-blocking notes; full CI is green on this head.

## Re-verification on `7434a229fa17af51c5d68ce3ce79c25d40c3e968`

- **`Lint & Type Check` passes** (previously red). Independently reproduced: `cd packages/client && tsc --noEmit` exits 0 clean.
- The incremental diff does exactly what it should:
  - `opencode/index.ts` catch block resets the `never` narrowing with `const captured = observedState as TurnState | null;` plus a null guard, and now consults the per-turn abort record (`takeTurnAbortRecord` / `inferOpenCodeTurnAbortRecord`) with `record.disposition === "silent"` returning `false` — consistent with the post-outcome path.
  - `handler.test.ts` imports `type SessionEvent` from `@first-tree/shared`.
  - New test covers the previously untested branch: read-only (non-mutating) provider-entered turn + `shutdown(..., { settleProviderEntered: true })` settles as `unsafe_replay` with no provider-failure events.
- Local runs: OpenCode provider suites **73/73 pass**, Biome check on the changed files clean, no working-tree edits needed.
- CI on this head: Lint & Type Check, Test Client & Web, Test Server, Test CLI (1/2 + 2/2), CLI smoke jobs, Migrations — all green.

Previous review's structural findings still hold: SessionRuntime `settleProviderEntered` contract wiring, suspend/shutdown ordering, and the classification/diagnostic separation were verified correct; no further concerns.

packages/client/src/providers/opencode/index.ts
```diff
@@ -1014,13 +1107,45 @@ export const createOpenCodeHandler: HandlerFactory = (config) => {
...
+            currentGeneration: generation,
+            sessionActive,
+            timedOut: false,
+            abortSignal: abort.signal,
+          });
+        if (record.disposition === "silent") {
+          return false;
+        }
+        if (record.cause === "lifecycle" || record.cause === "session_inactive") {
+          if (state.sawUnsafeTool || (state.sawProviderActivity && settleProviderEntered)) {
```
yuezengwu, line 1127: [P1] Track submitted input before the first output event

`runProcess` has already written the prompt to stdin, but `sawProviderActivity` only becomes true in `handleEvent`. If manager shutdown wins after the child has received the input but before its first parsed event, this condition falls through to `token.retry` despite `settleProviderEntered: true`. I reproduced the original input being submitted again through a fresh handler following that retry; the first-event control consumes it once. A silent/delayed output stream cannot prove that the submitted turn did no work, so this can replay side effects across restart. Please track input submission independently of output observation for graceful settlement, preserve retry for genuinely unsubmitted work, and cover shutdown after stdin receipt but before the first event.

yuezengwu (changes requested):
Reviewed exact head `7434a229fa17af51c5d68ce3ce79c25d40c3e968` against `main@84ccd7704ca128d52a78957c8fe5657e86bb659e`. One additional lifecycle admission gap remains; see the inline P1.

Independent reproduction used a controlled child that recorded receipt of complete stdin, emitted no stream event and stayed running. `shutdown("manager_shutdown", { settleProviderEntered: true })` produced no completion and one `token.retry(..., "manager_shutdown")`. Following that retry through a fresh handler recorded **two byte-identical inputs**. The otherwise identical control emitting a first `step_start` event consumed once, with no retry. This is controlled process evidence, not a live OpenCode or actual Client restart test.

All 250 focused existing tests pass, as do Client build, full typecheck, `pnpm check` and `git diff --check`. The new boundary probe fails only in the no-first-event case.

packages/client/src/providers/opencode/index.ts
bestony: Confirmed with an independent reproduction on head `7434a229f`: a controlled child that records complete stdin receipt and emits no stream event, `shutdown("manager_shutdown", { settleProviderEntered: true })` → `token.retry(messages, "manager_shutdown")` and no `token.complete`. Redelivering through a fresh handler re-submits the original content into a new child. The control emitting a first event consumes once (covered by the existing read-only shutdown test). Agreed this is a real admission gap: `sawProviderActivity` flips only on the first parsed event, so a silent/delayed stream between stdin receipt and first event escapes the consume branch even though the prompt was fully submitted. The pi handler already models the contract this needs: `hasProviderEntryEvidence` treats `promptWriteCommitted` (and after-write-unknown for graceful drain) as provider entry. Tracking input submission independently of output observation and treating submitted input as provider-entry evidence for the settle modes — while preserving retry for genuinely unsubmitted work — is the right fix; please also cover shutdown-after-stdin-receipt-before-first-event with a test. Retracting my earlier approve pending this fix.

bestony (changes requested):
**Verdict: request changes — retracting my earlier approve after independently confirming the P1 lifecycle admission gap.**

Reproduction on `7434a229f`: controlled child records complete stdin receipt, emits no stream event, stays running. `shutdown("manager_shutdown", { settleProviderEntered: true })` → `token.retry(messages, "manager_shutdown")`, no `token.complete`. Redelivering through a fresh handler re-submits the original content into a new child. The first-event control consumes once (no retry).

Root cause: the consume branch conditions on `state.sawUnsafeTool || (state.sawProviderActivity && settleProviderEntered)`, and `sawProviderActivity` only flips on the first parsed stream event — so a turn whose prompt was fully submitted to the provider but whose output stream is silent/delayed escapes consumption even under an explicit settle-provider-entered lifecycle. A silent stream cannot prove the submitted turn did no work.

The platform contract already models the needed evidence: the pi handler's `hasProviderEntryEvidence` treats `promptWriteCommitted` as provider entry for operator suspend and after-write-unknown as entry for graceful drain. Please track input submission (prompt write + stdin end) independently of output observation, treat submitted input as provider-entry evidence for the settle modes, preserve `token.retry` for genuinely unsubmitted work, and add a test for shutdown landing after stdin receipt but before the first event.

Everything else from the prior passes stands: typecheck clean, 73/73 OpenCode tests, Biome clean, CI green on this head, SessionRuntime wiring and ordering verified.

PR #2408, opened 2026-09-20
fix(client): stop reposting stale runtime failure notices after every successful turn

L42y:
## Summary

- Fix a loop that reposted one stale runtime failure notice (e.g. the Antigravity "Individual quota reached … Resets in 1h5m4s." notice) after every later successful turn in the same chat.
- `postPendingRuntimeFailureNotice` returned early when the delivery settlement lease turned invalid while the SDK post was in flight, skipping the `entry.pendingRuntimeFailureNotice = null` cleanup. The surviving payload was then re-marked onto unacknowledged inbox deliveries at every suspend (`markNoticeRequiredForProcessingPrefix`) and reposted on every server redelivery (`settleNoticeRequiredRedelivery`), each time with the identical, hours-old provider message.
- The pending slot is now cleared as soon as the notice post succeeds, guarded by the same identity check (`isSameSession` + same payload object) that already protects a newer payload captured by a replacement route. Settlement protection is unchanged: an invalidated lease still skips the ACK, and the existing stale-notice custody tests still pass.

## Root cause

Observed in production: every successful agent reply was followed ~3–10s later by the same capacity/quota runtime notice, with a frozen "Resets in …" countdown repeating verbatim for hours, each one immediately preceding a `session resumed` log line. Turn-by-turn daemon logs showed a stuck, unacknowledged inbox entry being redelivered after each ACK; the fail-closed ambiguous resume completion posted the pending notice, and the lease-invalidation race kept the payload alive on the session entry, so each suspend re-armed the ledger notice obligation and each redelivery re-posted it.

## Validation

- New regression test in `session-manager-edge-coverage.test.ts`: posts a pending quota notice through a consumed completion whose settlement lease is invalidated mid-post; asserts the pending slot is cleared, the notice is sent exactly once, and the ACK stays skipped for the invalidated route. Fails before the fix, passes after.
- `pnpm exec vitest run src/__tests__/session-manager-edge-coverage.test.ts src/__tests__/runtime-notice.test.ts src/__tests__/provider-retry-policy.test.ts src/__tests__/session-manager.test.ts src/__tests__/session-manager-more-coverage.test.ts src/__tests__/session-manager-retry.test.ts src/providers/antigravity/__tests__/handler.test.ts` — all pass.
- `pnpm run typecheck` (client) passes; `biome check` on the touched files is clean.


L42y:
recheck

yuezengwu (approved):
Reviewed exact head `17f336775af47a8638ab210ee8ecddf036bee9c3` against current main `84ccd7704ca128d52a78957c8fe5657e86bb659e`. No blocking findings. Clearing the same pending payload after a successful post preserves the replacement-session/payload guards; consumed completion and terminal rejection still recheck the settlement lease before ACK. Failed notice delivery retains its recovery path.

Locally reran 385 affected tests across 8 files, shared/client build, workspace typecheck, check, and diff check; all passed. This does not include a live provider or production reproduction.

PR #2409, opened 2026-09-20
fix(server): prevent session state deadlocks with membership snapshots

yuezengwu:
## Summary

Concurrent Issue events can deadlock predictive session activation against a GitHub membership snapshot: the session INSERT implicitly locks the agent before the chat, while the snapshot locks the chat before the agents. The webhook can return successfully after committing the card and inbox entry while the session projection fails.

Acquire chat then agent `FOR KEY SHARE` before the session write. Retry only the session/presence transaction on PostgreSQL `40P01`, at most three attempts with 10–19ms / 20–29ms jitter, and notify only from a successful changed commit. Membership permission locks retain their existing strength.

## Validation

- [x] `pnpm check` — passed with existing warnings.
- [x] `pnpm typecheck` — passed.
- [x] `VITEST_MAX_FORKS=2 pnpm test --concurrency=1` — server: 300 files / 3,543 tests passed; unchanged package tasks reused Turbo cache.
- Real PostgreSQL 17 regressions: the original code failed all three initial webhook/archive checks, including a real deadlock in the agents snapshot query. The final focused run passed all 75 tests across three files, including 14 new regressions.
- Tests cover snapshot/archive lock order using backend-specific lock waits and NOWAIT assertions; transaction rollback/retry/exhaustion; wrapped and cyclic causes; nonretryable errors; notification suppression after a no-op retry; missing parents; concurrent Issue opened/assigned delivery and redelivery without duplicate cards or inbox entries.
- An earlier full run had one intermittent failure in the unchanged provider-model cache eviction test. Its isolated rerun (10/10) and the final full run both passed.

## Change Surface

Server session-state persistence and matching product tests. No schema, migration, package, CLI, or public API changes.

## Notes

Each upsert adds two indexed parent-row lock queries and can wait for an in-flight membership snapshot to commit. Validation uses local real PostgreSQL 17; it does not claim production deployment or release acceptance. Relevant existing live QA cases: `github-webhook-routing-regression` and `authenticated-ws-inbox-delivery`. Production deployment and post-deployment alarm observation remain separate.


bestony (approved):
## Review

**APPROVE.** The lock-order diagnosis is correct, the fix is minimal, and the regression coverage genuinely reproduces the failure modes against real PostgreSQL.

### Lock ordering — verified

- Pre-fix, `upsertSessionState`'s INSERT took implicit FK `FOR KEY SHARE` locks on `agents` then `chats`, while `lockChatMembershipAndAgentRows` (`membership/lock.ts`) takes `chats` `FOR UPDATE` → membership rows → `agents` `FOR UPDATE`. That's the classic inversion that Aurora aborts with `40P01`, after the webhook's card/inbox transaction has already committed.
- The new `chats` → `agents` `FOR KEY SHARE` pre-locks order the upsert chat-first, matching the snapshot path; the agent-before-session-write claim also matches `archiveAllSessionsForAgent` (agent `FOR UPDATE` → session updates). `FOR KEY SHARE` is exactly the strength FK checks take, so the conflict surface against other session writers is unchanged.
- The two new pre-locks cannot regress the missing-parent contract: a SELECT matching no row takes no lock, and the INSERT still raises `23503` (covered by tests).
- I checked other `agents`-first lock paths (scheduled-job creation, agent identity pinning, resource catalog): none take session rows or chat locks in a conflicting direction within the same transaction, so no new inversion is introduced. The publisher path locks authority rows before the chat row, consistent with the new ordering.

### Retry semantics — verified

- Retry covers only the session/presence transaction; caller-side message and webhook writes are never replayed (the important property for a webhook that must not double-deliver).
- Bounded at 3 total attempts, jitter 10–19ms then 20–29ms; only `40P01` retried; cause-chain walk is cycle-safe; attempt-local clock and `projectionChanged`; notifications fire only after a successful changed commit, at most once per call.
- The no-op replay case is handled correctly: if a concurrent writer commits the same state between attempts, the replayed attempt short-circuits and no duplicate notification is emitted.

### Tests

`session-state-deadlock.test.ts` stages the real interleaving with two backends and asserts ordering with `NOWAIT` (so PostgreSQL aborting a side cannot conceal an inversion), covers the real exported `lockChatSpeakerSnapshot`, driver-real `40P01` injection via `agent_presence` triggers, exhaustion, non-deadlock SQLSTATEs, cyclic causes both ways, and FK contracts. `session-state-archive-lock.test.ts` proves the upsert blocks on the agent before touching the session row the runtime-switch archive needs, via `pg_blocking_pids` + `NOWAIT`. The webhook route test proves end-to-end idempotency: injected first-attempt `40P01`, both deliveries return 200, exactly one mapping/session, redelivery creates no duplicate cards or inbox entries.

### Nits (non-blocking)

1. The comment above `isDeadlockDetectedError` says "Drizzle wraps postgres-js errors in `cause`". With the postgres-js driver, drizzle rethrows the original error when rollback succeeds and only wraps when rollback itself fails; the walk handles both cases, so this is a comment-accuracy nit only.
2. The pre-locks run even on same-state no-op upserts, so steady-state activation bursts can queue behind an in-flight membership snapshot. Acknowledged in the PR notes; acceptable given the alternative is the deadlock itself.

No schema, migration, or public API changes. CI green (Lint & Type Check, Test Server). Scope and stated limitations (local PG 17 regression, no production deployment claim) are accurate.

